PR #9599, opened 2021-06-16
ovsx-client: add special handling for external vscode builtins

vince-fugnitto:
#### What it does

The pull-request adds additional logic to `@theia/ovsx-client` in order to handle `ms-vscode` builtin extensions which are included with vscode at build time (referenced in [vscode/product.json](https://github.com/microsoft/vscode/blob/1.57.0/product.json#L34-L126).  For such extensions, we want to handle them as specifically as regular vscode builtins meaning they should be resolved at their compatible versions. Since `ms-vscode` is not a reserved namespace (other extensions may use the namespace) we need to maintain a list of extensions we need to handle specifically.

#### How to test

1. ensure a clean working state by removing the `plugins` folder and tmp files:
    - `rm -rf plugins/`
    - `rm -rf /tmp/vscode-unpacked`
    - `rm -rf ~/.theia/extensions`
2. remove builtin `ms-vscode` extensions from the [root package.json](https://github.com/eclipse-theia/theia/blob/c92e822830e356c77d3186fda73b47f91da272e9/package.json#L85):
    - `vscode-builtin-js-debug`
    - `vscode-builtin-node-debug`
    - `vscode-builtin-node-debug2`
    - `vscode-references-view`
3. start the application, and open the `extensions-view`
4. confirm that searching, opening extensions's readme, installing extensions work correctly
5. search for `js-debug` - extension should be at version `1.50.0` (current api version)
6. search for `js-debug-companion` - extension version should be latest compatible api version
7. repeat step 1 - ensure that no builtins are present
8. download the `ms-vscode` extension pack (change from `.zip` to `.vsix`) - ([pack](https://github.com/eclipse-theia/theia/files/6658996/ms-vscode.zip)
9. install the `.vsix` from the `install from vsix` command
10. ensure that extensions in the pack are resolve to version `1.50.0` it if exists, else the latest compatible version.

#### Review checklist

#### Reminder for reviewers

Co-authored-by: Alvaro Sanchez-Leon <alvaro.sanchez-leon@ericsson.com>

### DIFF
--- a/dev-packages/ovsx-client/src/ovsx-client.ts
+++ b/dev-packages/ovsx-client/src/ovsx-client.ts
@@ -34,6 +34,14 @@ const postJson = bent('POST', {
     'Accept': 'application/json'
 }, 'json', 200);
 
+/**
+ * The list of external builtins included with vscode at build time.
+ * // https://github.com/microsoft/vscode/blob/1.57.0/product.json#L34-L126
+ */
+export const VSCODE_EXTERNAL_BUILTINS = [
+    'ms-vscode.node-debug', 'ms-vscode.node-debug2', 'ms-vscode.references-view', 'ms-vscode.js-debug-companion', 'ms-vscode.js-debug', 'ms-vscode.vscode-js-profile-table'
+];
+
 export interface OVSXClientOptions {
     apiVersion: string
     apiUrl: string
@@ -134,8 +142,7 @@ export class OVSXClient {
             return undefined;
         }
 
-        const namespace = extensions[0].namespace.toLowerCase();
-        if (this.isBuiltinNamespace(namespace)) {
+        if (this.isBuiltinExtension(id)) {
             const apiVersion = this.options!.apiVersion;
             for (const extension of extensions) {
                 if (this.isVersionLTE(extension.version, apiVersion)) {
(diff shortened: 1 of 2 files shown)


alvsan09 (commented):
It looks good to me! Thanks!

dev-packages/ovsx-client/src/ovsx-client.ts
```diff
@@ -160,7 +167,7 @@ export class OVSXClient {
      */
     getLatestCompatibleVersion(entry: VSXSearchEntry): VSXAllVersions | undefined {
         const extensions = entry.allVersions;
-        if (this.isBuiltinNamespace(entry.namespace)) {
+        if (this.isBuiltinExtension(`${entry.namespace}.${entry.name}`)) {
```
alvsan09, line 170: In the initial change, I did considered this name but I thought it would be confusing,
as at a first impression the reader may think it's referring to the builtin extensions of the specific application being handled.  but this is more a type of extension.
may be we can consider `isBuiltinType`?

vince-fugnitto: We never used the terminology `builtin type` before so it might add more confusion.
I've also made sure to properly document the method to reference what it is used for.

The check is to determine if a vscode-managed builtin is present (vscode and eclipse-theia namespace, with external vscode builtins). If you think it might be more descriptive, I could rename the method to:

```ts
protected isVscodeBuiltinExtension(id: string): boolean {
```

vince-fugnitto:
Closing for the moment, more investigation/discussion needed.

Issue #9326, opened 2021-04-10
Pin editor in tab-bar


DukeNgn:
### Feature Description:
+ As a user, sometimes I would like to pin editors in the tab-bar so I don't accidentally close it later.
+ Once an editor was pinned, triggering the keyboard shortcut <kbd>cmd</kbd>+<kbd>w</kbd> won't close the editor. To close the editor, users need to explicitly toggle back to 'unpin' mode by either clicking on the tab editor or using a keyboard shortcut.

### Additional info:
+ vscode uses the keyboard shortcut: <kbd>cmd+k</kbd><kbd>shift+enter</kbd> to toggle pin mode of an editor.
+ pinned editor (with a pin icon) vs normal editor in vscode:
![image](https://user-images.githubusercontent.com/43587865/114272799-90c01500-99e5-11eb-8a34-e16604043e4b.png)

PR #9614, opened 2021-06-20
Add ability to pin tabs

sh7dm:
#### What it does

Commands for pinning and unpinning tabs. Addresses #9326.

#### How to test

Open multiple editors, pin and unpin some of those

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (commented):
@sh7dm are you planning on updating the UI for pinned tabs as well, to visually represent that they are pinned? At the moment there is no distinction between pinned tabs and regular tabs.

For example, in the issue and in vscode we have (first tab is pinned):

image

sh7dm:
@vince-fugnitto hello, I plan for sure. Is [`title.className`](http://phosphorjs.github.io/phosphor/api/widgets/classes/title.html#classname) the correct way to add extra classes to Phosphor tabs&

vince-fugnitto:
@sh7dm I believe it would be a fine solution, for example it is used when setting a dirty decoration in editors:

https://github.com/eclipse-theia/theia/blob/4c9b8b80a4b4a5dfc969737ac8314c8a82f2cb1d/packages/core/src/browser/saveable.ts#L207-L217

sh7dm:
One more feature: keyboard shortcut. Is it that simple like registering both on the same shortcuts since those commands are mutually exclusive?

vince-fugnitto:
> One more feature: keyboard shortcut. Is it that simple like registering both on the same shortcuts since those commands are mutually exclusive?

It should be, you can try (and set it to the same value found in vscode).

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -281,6 +291,8 @@ export const supportPaste = browser.isNative || (!browser.isChrome && document.q
 
 export const RECENT_COMMANDS_STORAGE_KEY = 'commands';
 
+const PINNED_CLASS = 'theia-mod-pinned';
```
vince-fugnitto, line 294: @sh7dm something to try, rather than include the `svg` for each icon, we can simply use `codicons` directly:

```
codicon codicon-pinned
```

We have other examples in the codebase:

https://github.com/eclipse-theia/theia/search?q=codicon

sh7dm: What about integration with dirty classes and positioning?

vince-fugnitto: Same idea as the following (using the [unicode](https://github.com/microsoft/vscode-codicons/blob/main/dist/codicon.csv)):


```css
.theia-search-button.codicon-filter:not(.filter-active):before {
    content: "\eb85";
}

.theia-search-button.codicon-filter.filter-active:before {
    content: "\eb83";
}
```

It is something to at least try and see if it works :)

sh7dm:
> > One more feature: keyboard shortcut. Is it that simple like registering both on the same shortcuts since those commands are mutually exclusive?
> 
> It should be, you can try (and set it to the same value found in vscode).

No, that didn't work :( . I changed unpin to `ctrlcmd+k ctrlcmd+shift+enter` and both worked. We can do it as a single `pin-unpin` command

packages/core/src/browser/common-frontend-contribution.ts
sh7dm: Weird, those codicons don't work, maybe codicon package is out-of-date?
image

vince-fugnitto: > Weird, those codicons don't work, maybe codicon package is out-of-date?

It looks like `pin` is supported, but `pin-dirty` would work once we update monaco (https://github.com/eclipse-theia/theia/pull/9154) which contributes the codicons.

sh7dm: So, what can we do? I can keep SVGs and add a TODO to them about that update.f

vince-fugnitto: > So, what can we do? I can keep SVGs and add a TODO to them about that update.

I think that would be fine, the monaco merge should happen fairly soon but doing a refactor from svg to codicons can be done at a later time 👍

sh7dm:
So, I think I can merge those two commands into one to save space in menu + give users a single keybinding?

vince-fugnitto:
> So, I think I can merge those two commands into one to save space in menu + give users a single keybinding?

@sh7dm I'm not sure I follow? In vscode there are two commands present:

```ts
export const PIN_EDITOR_COMMAND_ID = 'workbench.action.pinEditor';
export const UNPIN_EDITOR_COMMAND_ID = 'workbench.action.unpinEditor';
```

I do not think we should merge the commands, rather their enablement should be controlled by if the editor is currently pinned or not.

sh7dm:
Sadly, those couldn't coexist with the same keybinding (see [my comment](https://github.com/eclipse-theia/theia/pull/9614#issuecomment-865064855))

vince-fugnitto:
> Sadly, those couldn't coexist with the same keybinding (see [my comment](https://github.com/eclipse-theia/theia/pull/9614#issuecomment-865064855))

They can, just need a proper `when` context:

Screen Shot 2021-06-21 at 2 22 04 PM

Screen Shot 2021-06-21 at 2 22 53 PM

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -350,6 +364,10 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
         this.contextKeyService.createKey<boolean>('isWindows', OS.type() === OS.Type.Windows);
         this.contextKeyService.createKey<boolean>('isWeb', !this.isElectron());
 
+        this.pinnedKey = this.contextKeyService.createKey<boolean>('activeEditorIsPinned', false);
```
vince-fugnitto, line 367: Shouldn't the functionality be specific to `editors` only, what does it mean to have a pinned `widget` (ex: scm).
If it's the case, the functionality should not be added to `@theia/core`.

vince-fugnitto (commented):
A couple of observations with the pull-request:

- Should the functionality be specific to editors-widgets only?
- What behavior do we expect for pinned editors in the sidebar (not main-area), they are currently not properly decorated.
- Clicking the pin icon of a pinned editor does not unpin it like in VS Code, any plans to align?
- Pinning multiple views in the sidebar re-arranges them in an order that it is unexpected.

In general, perhaps we can simplify to only pinning editors, and only pinning them in the main-area.

packages/core/src/browser/common-frontend-contribution.ts
sh7dm: 1. Common frontend contribution seems to be the place where all editor-tabs commands reside. Well, pinning widgets might help those who accidentally middle-click them although that doesn't quite make sense typically.
2. Nice catch, should be decorated I guess, but need to choose how should it look.
3. That's close button of tab, which is managed by Phosphor (not sure if I'm right), need to somehow intercept event and do unpin instead of close.
4. It does... I didn't really think about sidebar, arrangement code is specific to editor tabs probably

sh7dm: I think we can mitigate those problems (1, 2, 4) by removing pin/unpin commands for non-editor widgets. 3 needs investigation on opportunities

vince-fugnitto: @sh7dm initially I think we can implement pinning of editors only (`@theia/editor` extension), and likely only in the main-area for simplicity.

colin-grant-work: > Shouldn't the functionality be specific to `editors` only, what does it mean to have a pinned `widget` (ex: scm).
> If it's the case, the functionality should not be added to `@theia/core`.

This is always a tricky terminological gap to bridge between Theia and VSCode, since in VSCode, 'editor' means 'editor or webview,' (see screenshot for a pinned webview) but also implies 'any main-area widget'. Since Theia is much more flexible with what goes in what area, allowing editors and webviews to appear outside the main area and other widgets to appear in it, I'm not sure restricting the pinning functionality to editors makes much sense. I would be inclined to extend the ability to any closeable widget on any tabbar that allows 'close all' or 'close other' functionality (all of them, I think). Otherwise the logic to achieve strictly VSCode-like behavior when so much else is different can get torturous.

image

sh7dm: So we can fix pinned widgets placement on non-main views and decorate those, I think that would be really better and help avoiding edge cases @colin-grant-work described.

`activeEditorPinned` name was taken from VSCode keyboard shortcuts and should really be `activeWidgetPinned` since it works for everything.

vince-fugnitto: Let's go ahead with that :+1: Let's make it available for any closeable widget, and keep it generic (ex: `activeWidgetPinned`) like you've proposed would be a fine `contextKey`. The only issue would be determining how to properly decorate views not in the main-area (likely only sidebar has issues at the moment), and solving the issue of views re-arranging oddly in the sidebars.

colin-grant-work: Re: the context key name, we should probably either keep `activeEditorPinned` or implement both `activeEditorPinned` and `activeWidgetPinned` since plugin `when` clauses will expect `activeEditorPinned`.

vince-fugnitto: My issue was having editor references in `@theia/core` but you are right about plugins using the same `contextKey` so it should be implemented.

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -215,6 +215,16 @@ export namespace CommonCommands {
...
+    export const PIN_TAB: Command = {
+        id: 'core.pin.tab',
+        category: VIEW_CATEGORY,
+        label: 'Pin Tab'
+    };
+    export const UNPIN_TAB: Command = {
+        id: 'core.unpin.tab',
+        category: VIEW_CATEGORY,
+        label: 'Unpin Tab'
+    };
```
colin-grant-work, line 227: It looks like the VSCode ID's for these are `workbench.action.pinEditor`  and `workbench.action.unpinEditor`. We should either use those names or [map those names](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts) to this so that this command can be run by plugins.

sh7dm:
I did not really notice weird ordering in sidebar. @vince-fugnitto could you please describe reproduction steps?

vince-fugnitto:
> I did not really notice weird ordering in sidebar. @vince-fugnitto could you please describe reproduction steps?

I could no longer reproduce it, I believe part of the confusion is due to side-panels not being decorated when pinned.

sh7dm:
So how should we decorate pinned sidebar widgets? There seems to be quite a lot of free space in the corner where SVG pin icon can be placed. Top-right should be great. What do you think on that (at least for this PR)?
image

vince-fugnitto:
> So how should we decorate pinned sidebar widgets? 

@sh7dm I'm not really sure how it should look, we already have some issues in sidepanels for editors when changing file-icon themes. I would not hold back the pull-request if the initial changes does not add decorations. @colin-grant-work any thoughts?

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -837,6 +931,23 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
         return environment.electron.is();
     }
 
+    private isPinned(title: Title<Widget>): boolean {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        return (title as any).pinned;
```
colin-grant-work, line 936: I'd prefer a real type to this `as any` cast, something like
```typescript
interface PinnableTitle extends (Phosphor)Title { pinned?: boolean }
```
or (probably in `widget.ts` or somewhere)
```typescript
interface TheiaTitle extends Title {
    pinned?: boolean;
    // ... who knows what else in the future?
}
```
That way that information could be used elsewhere, for example if we want to override the rendering of tabbars [here](https://github.com/eclipse-theia/theia/blob/master/packages/core/src/browser/shell/tab-bars.ts#L133-L158).

colin-grant-work (commented):
This is looking pretty good to me. One small point is that in VSCode, when you pin an editor preview widget, it converts to a normal widget, but that isn't happening here.

If you'd like to implement that for editor previews, you can modify the `initializePreview` method in `editor-preview-widget.ts` to look like this:
```
    initializePreview(): void {
        const oneTimeListeners = new DisposableCollection();
        this._isPreview = true;
        this.title.className += ` ${PREVIEW_TITLE_CLASS}`;
        const oneTimeDirtyChangeListener = this.saveable.onDirtyChanged(() => {
            this.convertToNonPreview();
            oneTimeListeners.dispose();
        });
        oneTimeListeners.push(oneTimeDirtyChangeListener);
        const oneTimeTitleChangeHandler = () => {
            if ((this.title as any).pinned) {
                this.convertToNonPreview();
                oneTimeListeners.dispose();
            }
        };
        this.title.changed.connect(oneTimeTitleChangeHandler);
        oneTimeListeners.push(Disposable.create(() => this.title.changed.disconnect(oneTimeTitleChangeHandler)));
        this.toDispose.push(oneTimeListeners);
    }
```

If you're interested in implementing the unpin-on-click behavior, one way to do it would be to modify the rendering [here](https://github.com/eclipse-theia/theia/blob/master/packages/core/src/browser/shell/tab-bars.ts#L133-L158) to react to pinned / unpinned status. Another strategy would be do something like this: 

https://github.com/eclipse-theia/theia/blob/74fbeb7c2c70cab215f3cbc238d71918c6ceebb6/packages/editor-preview/src/browser/editor-preview-manager.ts#L70

Where you check whether the click is on an element with the class name you're interested in and, if so, execute the unpin command. But I'm fine leaving that out, for now.

colin-grant-work:
> > So how should we decorate pinned sidebar widgets?
> 
> @sh7dm I'm not really sure how it should look, we already have some issues in sidepanels for editors when changing file-icon themes. I would not hold back the pull-request if the initial changes does not add decorations. @colin-grant-work any thoughts?

I agree with this. Figuring out how to handle the decoration on sidebars will be a bit tricky, but I also think it isn't urgent because pinning in that context is going to be much less common than pinning in the main area, so we don't have to resolve it immediately.

sh7dm:
Build failure is not caused by changes, there's some network or server error. https://github.com/eclipse-theia/theia/runs/2902667623#step:5:1446

colin-grant-work:
@sh7dm this is working very well for me. I'm not seeing any problems on the main tabbar or other tabbars. Personally, I don't really like the tab rearrangement behavior, but that is how VSCode is doing it, and this code implements it correctly.

On the code side, I think I'm going to backpedal, with apologies, and suggest that the `TheiaTitle` type, while necessary in the long run, should maybe be created in a separate refactor where we apply it throughout the application. As long as we have to do a bunch of casts, it doesn't much matter whether it's to `any` or `TheiaTitle`, and introducing it here has actually increased the number of casts, rather than reducing them. Sorry to change directions there!

Otherwise the code looks good to me. @vince-fugnitto, are you comfortable with the changes to the tabbar code introduced by [this commit](https://github.com/eclipse-theia/theia/pull/9614/commits/279b37b315a786b4a4bf5a6602273992c7091691)? They look reasonable to me, but it's digging pretty deep into `core`.

sh7dm:
> Personally, I don't really like the tab rearrangement behavior, but that is how VSCode is doing it, and this code implements it correctly.

Why clone VSCode behavior in terms of UX? Theia is a different project with its own UX, we can implement better behavior I think. If possible, suggest your ideas for rearrangement.

> As long as we have to do a bunch of casts, it doesn't much matter whether it's to any or TheiaTitle, and introducing it here has actually increased the number of casts, rather than reducing them. Sorry to change directions there!

`any` cast is a linter error we have to suppress, so type is better, but I'm fine with casting to `any` to limit scope of the PR.

colin-grant-work:
> Why clone VSCode behavior in terms of UX? Theia is a different project with its own UX, we can implement better behavior I think. If possible, suggest your ideas for rearrangement.

A fair question. The answer (not a very good one) is that it's at least a baseline to target, and users familiar with VSCode will have a lower learning / surprise curve if we stick closely to the patterns VSCode establishes. I suppose the arguments run as follows:
## For rearranging
 - If you pin an editor, you probably want to keep it visible, and moving it to the left sort of achieves that.
 > But... probably isn't definitely, and 'left' doesn't mean 'visible' unless you fix the scroll behavior somehow so that you can't scroll them out of view.
## Against rearranging
 - It's visually a bit disorienting to see your tab zoom away, and if you have a lot tabs and decide to pin one that's far to the right, you can actually zoom it _out_ of view. 
 - In VSCode, it looks like pinning and location are actually mutually dependent. If you move a pinned tab to the right of an unpinned tab, it unpins. That's pretty unintuitive to me, though I'm sure I could get used to it.
> That's a specific quirk that doesn't have to do with the question of rearrangement in general.

My preference would be to just decouple pinning and arrangement, since they're conceptually separate. You can pin a tab if you don't want it to close (I'm a habitual `ctrl+k+w` user, so that appeals to me), but you can then put that pinned tab wherever you want (why not?). Alternatively, we could implement an application actual preference 'application.pinnedTabsLeft' or something that determines whether to rearrange or not. Then everyone (except you, the developer of the feature 😉) is happy.

> any cast is a linter error we have to suppress, so type is better, but I'm fine with casting to any to limit scope of the PR.

True, and I won't insist on reverting to `any` if you prefer to start the move to a `TheiaTitle` type (and @vince-fugnitto agrees?), though I would probably make it generic (with a default, since `Widget` is by far our most common `Title` type) `TheiaTitle<T = Widget> extends Title<T>`.

sh7dm:
Agreed with everything :+1:

colin-grant-work (approved):
This is looking good to me. Thanks for the UI changes - it feels like there are a lot fewer unexpected side effects.

colin-grant-work:
@vince-fugnitto

> Otherwise the code looks good to me. @vince-fugnitto, are you comfortable with the changes to the tabbar code introduced by this commit? They look reasonable to me, but it's digging pretty deep into core.

How do you feel about this change to the tabbars? It seems reasonable to me, but I think (looking ahead to breadcrumbs, also), like we'll probably want a better abstraction over the various things that want to decorate tabbars or affect their behavior.

vince-fugnitto:
@sh7dm can you please rebase and squash your pull-request, there are currently some conflicts and merge commits. I'll take a final look at the changes after.

@colin-grant-work I agree, the changes do not need to be done now, but there should likely be in place what you described :+1:

sh7dm:
I squashed and rebased everything. CI looks fine (Ubuntu test fail looks like not related). @vince-fugnitto PTAL when you have time

vince-fugnitto (commented):
It looks like the commands `pin tab` and `unpin tab` are not executable from the quick-command palette (<kbd>F1</kbd>).

sh7dm:
> It looks like the commands `pin tab` and `unpin tab` are not executable from the quick-command palette (F1).

Probably it's because they're functionally-enabled based on event. So, when event is not from tabBar itself we can get `this.shell.findTabBar` and find title there I think.

vince-fugnitto (commented):
I confirmed that the behavior works as expected :+1: 
- confirmed the commands for `pin` and `unpin` work
- confirmed the menu items for `pin` and `unpin` work
- confirmed that the keybindings work
- confirmed the when context works correctly
- confirmed the dirty pin works correctly

---

I did notice a bug however:

- pin the explorer
- move it to the main-area
- the explorer displays a pinned decoration
- attempt to unpin the widget, the context-menu displays pin


https://user-images.githubusercontent.com/40359487/125301750-478ba400-e2f9-11eb-9154-f17124af8ebb.mp4

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -334,6 +347,9 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
     @inject(AuthenticationService)
     protected readonly authenticationService: AuthenticationService;
 
+    protected pinnedKey: ContextKey<boolean>;
+    protected pinnedKeyCompat: ContextKey<boolean>;
```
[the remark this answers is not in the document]
vince-fugnitto: Yes I understand, the only issue is we now have two context-keys that extensions can use which are similar or duplicated. I don't think it hurts an extension to use the `activeEditorIsPinned` context-key, nor would they really re-use a context-key.

cc @colin-grant-work any thoughts?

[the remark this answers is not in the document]
sh7dm: I agree, that wording should not matter. Moreover, devs familiar with VSCode will use the same key as there.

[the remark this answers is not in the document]
colin-grant-work: I think it should be fine to only have the VSCode key. In VSCode, it applies to all main-area widgets, not just editors, so any references to that key still need to check what kind of widget it really is if they want to act on it.

vince-fugnitto:
@sh7dm were you able to reproduce the bug I mentioned in https://github.com/eclipse-theia/theia/pull/9614#pullrequestreview-704147174 with the help of the video?

sh7dm:
I was, thanks. Looking for a fix. Seems like that's because we need to reset that class when drop event happens.

sh7dm:
This might need a bit of discussion: the problem is because additional props don't get passed when Phosphor copies data. I have two fix options:
1. Check pinned state by class name
2. Avoid setting class name for sidebar editors.

@vince-fugnitto which would you prefer?

vince-fugnitto:
> 1. Check pinned state by class name

@sh7dm I think the first approach is reasonable :+1:

vince-fugnitto (commented):
@sh7dm do we plan on supporting the following use-case? (supported in vscode)

1. pin an editor
2. reload the application
3. the editor is restored, but not its pinned state

sh7dm:
I'd like to do so. How to include some extra data to save of application state?

vince-fugnitto:
> I'd like to do so. How to include some extra data to save of application state?

In order to store and restore state, [`StatefulWidget`](https://eclipse-theia.github.io/theia/docs/next/modules/core.statefulwidget-2.html) needs to be implemented.
You'll likely need to look at implementations and attempt to restore the pinned state.

colin-grant-work:
> In order to store and restore state, [`StatefulWidget`](https://eclipse-theia.github.io/theia/docs/next/modules/core.statefulwidget-2.html) needs to be implemented.

Perhaps this is a case where the `ApplicationShell` should do the work, since all of the pinning logic resides there, rather than on the widget? Otherwise all pinnable widgets (all widgets) would need to become `Stateful`, and that would wreck a lot of existing `StatefulWidget` code.

vince-fugnitto:
> > In order to store and restore state, [`StatefulWidget`](https://eclipse-theia.github.io/theia/docs/next/modules/core.statefulwidget-2.html) needs to be implemented.
> 
> Perhaps this is a case where the `ApplicationShell` should do the work, since all of the pinning logic resides there, rather than on the widget? Otherwise all pinnable widgets (all widgets) would need to become `Stateful`, and that could wreck a lot of existing `StatefulWidget` code.

You're right, we should save layout and not widget state. So we should take a look at `setLayoutData`, and `restoreLayout`.

sh7dm:
Thanks, will do!

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -561,9 +561,16 @@ export class ApplicationShell extends Widget {
      * to widgets; these need to be transformed before the layout can be serialized.
      */
     getLayoutData(): ApplicationShell.LayoutData {
+        const pinned: boolean[] = [];
+
+        toArray(this.mainPanel.widgets()).forEach((a, i) => {
+            pinned[i] = a.title.className.indexOf('theia-mod-pinned') >= 0;
+        });
```
vince-fugnitto, line 568: Is the goal only to restore the pinned state for the main-area? I think we should aim to restore the pinned state for the entire layout as the current behavior is confusing for users.

Also, do you mind extracting this functionality to a method, it will make it easier for downstream apps or extensions to customize without needing to re-implement `getLayoutData` entirely.

vince-fugnitto (commented):
Please be sure to clean up the commits, it seems the ECA check is failing.

packages/core/src/browser/shell/application-shell.ts
sh7dm: Will extract that. Side panels get saved&restored as well.

sh7dm:
I fixed ECA by going to the website, now it's all-ok.

packages/core/src/browser/shell/application-shell.ts
vince-fugnitto: I'd think all panels, for example the bottom panel looks a lot like the `main` area and it would be confusing for users to pin widgets there and not see them restored.

sh7dm: Ah, forgot about that one, sorry!

vince-fugnitto (commented):
I'm not sure about how you're restoring the pinned state on reload, mainly about why its only being done to main and bottom and not in a general way, I'll need some time to think about it.

@colin-grant-work I'd appreciate if you can take a look and provide any feedback if you have any.

sh7dm:
@vince-fugnitto Side panels are done in their generic code, but I couldn't find a good generic place for it in main/bottom.

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -561,9 +561,16 @@ export class ApplicationShell extends Widget {
      * to widgets; these need to be transformed before the layout can be serialized.
      */
     getLayoutData(): ApplicationShell.LayoutData {
+        const pinned: boolean[] = [];
+
+        toArray(this.mainPanel.widgets()).forEach((a, i) => {
+            pinned[i] = a.title.className.indexOf('theia-mod-pinned') >= 0
+        });
```
colin-grant-work: maybe just this:
```typescript
const pinned = toArray(this.mainPanel.widgets()).map(widget => widget.title.className.includes('theia-mod-pinned');
```

Alternatively, since there's some uncertainty below about whether the number of widgets examined here will be the same as the number of widgets eventually restored, you could use something like `WidgetManager.getDescription()` to get a (probably) unique string(ifiable) identifier for each widget, use that as the key in a `Record<string, boolean>` and then match that up with the restored widgets on the other side of the restoration.

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -638,6 +645,14 @@ export class ApplicationShell extends Widget {
         if (mainPanel) {
             this.mainPanel.restoreLayout(mainPanel);
             this.registerWithFocusTracker(mainPanel.main);
+            const widgets = toArray(this.mainPanel.widgets());
+            if (mainPanelPinned && mainPanelPinned.length === widgets.length) {
```
colin-grant-work, line 675: See above for a way to handle the uncertainty of whether the widgets restored will be the same as the widgets whose state was saved.

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -587,6 +588,17 @@ export class ApplicationShell extends Widget {
...
+    // Get an array corresponding to bottom panel widgets' pinned state.
+    getPinnedBottomWidgets(): boolean[] {
+        const pinned: boolean[] = [];
+
+        toArray(this.bottomPanel.widgets()).forEach((a, i) => {
+            pinned[i] = a.title.className.indexOf('theia-mod-pinned') >= 0;
+        });
+
+        return pinned;
+    }
```
colin-grant-work, line 600: Not sure that this warrants its own function. Maybe `getPinnedWidgetsForArea(area)` that can handle both of them?

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -643,6 +655,15 @@ export class ApplicationShell extends Widget {
             } else {
                 this.collapseBottomPanel();
             }
+            const widgets = toArray(this.bottomPanel.widgets());
+            if (bottomPanel.pinned && bottomPanel.pinned.length === widgets.length) {
```
colin-grant-work, line 659: Same comment re: handling uncertainty of restoration.

colin-grant-work:
The restoration on sidebars is a little unreliable for me. The following scenario seems to fail consistently.

1. Using the Electron app (don't know if it matters, but I was)
2. Open the application and `Reset workbench layout`
3. Close application and reopen.
4. Pin extensions view.
5. Close application and reopen.
6. Extensions view is not pinned.

If I move it into the main area and back, then it seems to retain its pinned state even on the sidebar. But if I close and reopen, leave it pinned, close and reopen again, it will have lost its pinned state. Since the only check for pinning is the classname, I'd guess the sidebar is wiping out the classname at some point. On the other hand, an editor on the sidebar seems to retain its pinned state more reliably, so there may be more to it - something related to `ViewContainer` or maybe `AbstractViewContribution`? Since the `SearchInWorkspace` widget shows the same bug, the `ViewContribution` is maybe a better bet? Or do the sidebars not store their state if they think nothing has happened? When I moved a few widgets out of the left but left the SCM widget there, the SCM widget retained its pinned state, but the Outline view, on the right side, which I hadn't touched, didn't restore its state.

Anyway - tl;dr is that I didn't have any problems in main or bottom areas, but the side panels were a little funky.

sh7dm:
Actually, I once noticed view self-unpinning after opening and closing it (switching side panel views). Maybe they are recreated on some conditions?

sh7dm:
Btw pinning side panel views is not really used feature, typically they do not need to be protected from being accidentally closing.

colin-grant-work:
@sh7dm, are you interested in further work on this feature?

sh7dm:
Not really sure, we should plan required changes

colin-grant-work:
@sh7dm, I think (apart from the conflicts that have arisen in the meantime), the main thing that was holding it back was the side-panel state restoration behavior. I agree with you that that isn't a very common use case, but we should decide whether to just disable it completely in side panels, or we should figure out what's making it buggy. If you're willing to investigate, I'd be happy to take a look at what you find; if you don't have time, I'd be happy to pick up the PR and do some investigation.

sh7dm:
Most likely I won't be able to properly debug this as I couldn't do that. Sorry for this problem. You can probably disable it for initial merge and open an issue for that.

PR #9620, opened 2021-06-21
monaco: use fetch instead of old AJAX

paul-marechal:
Update the code to use `fetch` to get Onigasm's wasm code.

Update our bindings such as Onigasm is lazily loaded instead of
hardcoded and downloaded on module loading.

#### How to test

Textmate syntax coloring should still be working.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
I confirmed that textmate syntax coloring still works correctly with the latest updates.

PR #9623, opened 2021-06-21
application-package: require deepmerge cjs

paul-marechal:
The line `const merge = require('deepmerge')` actually implies we want
the CJS version of deepmerge. It makes it easier to bundle using
Webpack.

#### How to test

N/A

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
Confirmed that it builds and there are no regressions 👍

Should we also align the `application-props` or is there a reason we use `umd` over `cjs`:

https://github.com/eclipse-theia/theia/blob/31f78efffccc739e8497e43e2473aaf55e7fccd8/dev-packages/application-package/src/application-props.ts#L20

paul-marechal:
I previously used `umd` because application props are common to browser/node, but with hindsight we should align to CJS:

1. Webpack doesn't always play nice with UMD modules (e.g. `jsonc-parser`).
2. We output CJS for our browser implementations anyway.

I'll update the commit.

vince-fugnitto (approved):

Issue #9434, opened 2021-05-04
Theia extensions: declare entrypoints


marechal-p:
Related: https://github.com/eclipse-theia/theia-blueprint/issues/74#issuecomment-832251519

Theia's backend uses a good deal of sub-processes, and some are spawned from entrypoints implemented within extensions. In order to bundle backend applications, we need Webpack to be aware of those entrypoints.

Theia extensions should declare their entrypoints so that Webpack can more easily be run.

Issue #9432, opened 2021-05-04
replace dynamic `require` calls to be customizable


paul-marechal:
Related: https://github.com/eclipse-theia/theia-blueprint/issues/74#issuecomment-832251519

Currently we use dynamic `require` calls in various backend places. This causes issues when trying to Webpack the backend.

To fix that issue, I think the best approach would be to expose a common script re-exposing `require`.  This way we can replace imports of to that module with a custom one using Webpack in order to replace the actual method used.

PR #9624, opened 2021-06-22
limit use of dynamic require

paul-marechal:
Using Node's builtin require function to dynamically import scripts
makes bundling Node apps difficult when using tools like Webpack.

Add a new package `@theia/dynamic-require`: This is a minimal package
that re-export a function wrapping Node's require. The purpose of this
is to factorize dynamic requires into a single package to make bundling
easier.

Remove as many occurences of Node require and replace the few required
ones with `@theia/dynamic-require`.

Add a new eslint rule `import/no-dynamic-require` to help detect places
that would benefit from `@theia/dynamic-require`.

Add `entryPoints` declarations to Theia extensions.

Add `extensionEntryPoints` property to `ApplicationPackage` that lists
all declared entry points.

Closes https://github.com/eclipse-theia/theia/issues/9434
Closes https://github.com/eclipse-theia/theia/issues/9432

#### How to test

Everything should work as before, this is a refactoring.

As long as the new test in `dev-packages/application-package` passes it means the added API works as well.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-package/src/application-package.ts
+++ b/dev-packages/application-package/src/application-package.ts
@@ -22,8 +22,6 @@ import { ExtensionPackageCollector } from './extension-package-collector';
 import { ApplicationProps } from './application-props';
 const merge = require('deepmerge/dist/cjs');
 
-// tslint:disable:no-implicit-dependencies
-
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type ApplicationLog = (message?: any, ...optionalParams: any[]) => void;
 export class ApplicationPackageOptions {
@@ -99,12 +97,13 @@ export class ApplicationPackage {
         return this._pck = readJsonFile(this.packagePath);
     }
 
-    protected _frontendModules: Map<string, string> | undefined;
-    protected _frontendElectronModules: Map<string, string> | undefined;
-    protected _backendModules: Map<string, string> | undefined;
-    protected _backendElectronModules: Map<string, string> | undefined;
-    protected _electronMainModules: Map<string, string> | undefined;
-    protected _extensionPackages: ReadonlyArray<ExtensionPackage> | undefined;
+    protected _frontendModules?: Map<string, string>;
+    protected _frontendElectronModules?: Map<string, string>;
+    protected _backendModules?: Map<string, string>;
+    protected _backendElectronModules?: Map<string, string>;
+    protected _electronMainModules?: Map<string, string>;
+    protected _extensionPackages?: ReadonlyArray<ExtensionPackage>;
+    protected _extensionEntryPoints?: Map<string, string>;
 
     /**
      * Extension packages in the topological order.
@@ -118,6 +117,27 @@ export class ApplicationPackage {
             this._extensionPackages = collector.collect(this.pck);
         }
         return this._extensionPackages;
+    }
+
+    /**
+     * All entry points defined by the installed Theia extensions.
+     *
+     * Map of entry point name to script import path.
+     */
+    get extensionEntryPoints(): ReadonlyMap<string, string> {
+        if (!this._extensionEntryPoints) {
+            this._extensionEntryPoints = new Map();
+            for (const extensionPackage of this.extensionPackages) {
+                for (const theiaExtension of extensionPackage.theiaExtensions) {
+                    if (typeof theiaExtension.entryPoints === 'object') {
+                        for (const [key, value] of Object.entries(theiaExtension.entryPoints)) {
+                            this._extensionEntryPoints.set(key, `${extensionPackage.name}/${value}`);
+                        }
+                    }
+                }
+            }
+        }
+        return this._extensionEntryPoints;
     }
 
     getExtensionPackage(extension: string): ExtensionPackage | undefined {
--- a/dev-packages/application-package/src/application-package.ts
+++ b/dev-packages/application-package/src/application-package.ts
@@ -122,7 +122,7 @@ export class ApplicationPackage {
     /**
      * All entry points defined by the installed Theia extensions.
      *
-     * Map of entry point name to script import path.
+     * Map of entry point name to script absolute path.
      */
     get extensionEntryPoints(): ReadonlyMap<string, string> {
         if (!this._extensionEntryPoints) {
@@ -131,7 +131,7 @@ export class ApplicationPackage {
                 for (const theiaExtension of extensionPackage.theiaExtensions) {
                     if (typeof theiaExtension.entryPoints === 'object') {
                         for (const [key, value] of Object.entries(theiaExtension.entryPoints)) {
-                            this._extensionEntryPoints.set(key, `${extensionPackage.name}/${value}`);
+                            this._extensionEntryPoints.set(`${extensionPackage.name}/${key}`, require.resolve(`${extensionPackage.name}/${value}`));
                         }
                     }
                 }
--- a/dev-packages/application-package/src/extension-package.ts
+++ b/dev-packages/application-package/src/extension-package.ts
@@ -25,6 +25,10 @@ export interface Extension {
     backend?: string;
     backendElectron?: string;
     electronMain?: string;
+    /**
+     * Map of script name to script relative path
+     */
+    entryPoints?: Record<string, string>;
 }
 
 export class ExtensionPackage {
--- a/dev-packages/application-package/src/extension-package.ts
+++ b/dev-packages/application-package/src/extension-package.ts
@@ -26,7 +26,7 @@ export interface Extension {
     backendElectron?: string;
     electronMain?: string;
     /**
-     * Map of script name to script relative path
+     * Map of entry point id to js script path relative to its package root.
      */
     entryPoints?: Record<string, string>;
 }
--- a/dev-packages/dynamic-require/index.js
+++ b/dev-packages/dynamic-require/index.js
@@ -1,0 +1,26 @@
+"use strict";
+/********************************************************************************
+ * Copyright (C) 2021 Ericsson and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+module.exports = function dynamicRequire(id) {
+    if (typeof id !== 'string') {
+        throw new TypeError('module id must be a string');
+    }
+    if (id.startsWith('.')) {
+        throw new Error(`module id cannot be a relative path, id: "${id}"`);
+    }
+    return require(/* webpackIgnore: true */ id);
+}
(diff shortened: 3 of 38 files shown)


paul-marechal:
This PR is part of the work to bundle/package the browser version of Theia Blueprint.

@eclipse-theia/core I added a new package `@theia/dynamic-require` which is meant to be used whenever one wants to dynamically require a script from an absolute path. This impacts mostly plugin support and core IPC. I really wanted to avoid making a new package, but it seems like the best way to go for me. Just a heads up in case anyone would have concerns about this.

packages/plugin-ext/src/hosted/node/metadata-scanner.ts
```diff
@@ -40,24 +40,23 @@ export class MetadataScanner {
...
+        let scanner: PluginScanner | undefined;
         if (plugin && plugin.engines) {
             const scanners = Object.keys(plugin.engines)
-                .filter((engineName: string) => this.scanners.has(engineName))
-                .map((engineName: string) => this.scanners.get(engineName));
-
+                .filter(engineName => this.scanners.has(engineName))
+                .map(engineName => this.scanners.get(engineName)!);
             // get the first suitable scanner from the list
             scanner = scanners[0];
```
tsmaeder, line 55: What if _no_ engines are found here?

dev-packages/application-package/src/application-package.ts
```diff
@@ -120,6 +119,27 @@ export class ApplicationPackage {
...
+     *
+     * Map of entry point name to script import path.
+     */
+    get extensionEntryPoints(): ReadonlyMap<string, string> {
+        if (!this._extensionEntryPoints) {
+            this._extensionEntryPoints = new Map();
+            for (const extensionPackage of this.extensionPackages) {
+                for (const theiaExtension of extensionPackage.theiaExtensions) {
+                    if (typeof theiaExtension.entryPoints === 'object') {
+                        for (const [key, value] of Object.entries(theiaExtension.entryPoints)) {
```
tsmaeder, line 133: This code here requires entrypoint id's to be globally unique. I don't think this is a necessary or desirable requirement.

packages/core/package.json
```diff
@@ -114,6 +115,11 @@
     ]
   },
   "theiaExtensions": [
+    {
+      "entryPoints": {
```
tsmaeder, line 119: Why not just make this an array: I don't see the need to have an id for each entry point.

dev-packages/application-package/src/application-package.ts
```diff
@@ -120,6 +119,27 @@ export class ApplicationPackage {
         return this._extensionPackages;
     }
 
+    /**
+     * All entry points defined by the installed Theia extensions.
+     *
+     * Map of entry point name to script import path.
+     */
+    get extensionEntryPoints(): ReadonlyMap<string, string> {
```
tsmaeder, line 127: I searched through the PR, but I don't see the "extensionEntryPoints" instance var used anywhere 🤷

dev-packages/dynamic-require/index.js
```diff
@@ -0,0 +1,26 @@
+"use strict";
```
tsmaeder, line 1: Could you explain why we need a *.d.ts and a *.js file instead of just writing typescript?

dev-packages/application-package/src/application-package.ts
```diff
@@ -120,6 +119,27 @@ export class ApplicationPackage {
...
+        if (!this._extensionEntryPoints) {
```
tsmaeder, line 128: I typically frown upon lazy initialization for two reasons:

1. If the init is fast, we should just do it at object construction
2. If it is slow, we should make that explicit by returning a promise. A method that is fast or slow arbitrarily is usually not what we want.
3. If you don't control when the init is done, the init might be invoked before it's prerequisites are fulfilled.

dev-packages/application-package/src/extension-package.ts
```diff
@@ -25,6 +25,10 @@ export interface Extension {
     backend?: string;
     backendElectron?: string;
     electronMain?: string;
+    /**
+     * Map of script name to script relative path
+     */
+    entryPoints?: Record<string, string>;
```
tsmaeder, line 31: As mentioned elsewhere, I don't think we need an id for the entry points: if the _values_ are not unique, we should coalesce them anyway. (no need to load /foo/bar/zoz twice when it is mentioned twice.).

dev-packages/application-package/src/extension-package.ts
```diff
@@ -25,6 +25,10 @@ export interface Extension {
     backend?: string;
     backendElectron?: string;
     electronMain?: string;
+    /**
+     * Map of script name to script relative path
```
tsmaeder, line 29: We need doc describing what kind of value should be put here. What does "script-relative-path" mean? Relative to the package.json? Should it be './foobar.js', do we reference the *.ts file or do we reference the compiled version?

tsmaeder (changes requested):
Did a quick code review (no testing). If I understand it correctly, this PR:

1. Enforces that everyone uses the "dynamicRequire" function instead of "require".
2. Turns off webpack reference detection on "require", when `dynamicRequire` is used
3. Should add a root in the reachability graph for webpack to include the modules referenced as "entryPoints" in package.json. (I haven't seen the code that does that, though).

dev-packages/dynamic-require/index.js
paul-marechal: It felt simplier to setup for such a trivial package. Tell me if you'd rather want compiled Typescript.

dev-packages/application-package/src/application-package.ts
paul-marechal: Ok, I just aligned with what was already there. VS Code have a `@memoize` decorator for that kind of stuff, so idk really. I'd keep the code that way for now, until we decide we should change the rest too?

dev-packages/application-package/src/extension-package.ts
paul-marechal: So you suggest logic to de-duplicate entry points based on their value(/path)?

dev-packages/application-package/src/extension-package.ts
paul-marechal: Will change to

> Map of entry point id to js script path relative to its package root.

dev-packages/application-package/src/extension-package.ts
tsmaeder: If we're just using them as roots in the reachability graph, yes. If this was Java, I'd use a Set<String>.

paul-marechal:
>     1. Enforces that everyone uses the "dynamicRequire" function instead of "require".

Only when doing dynamic requires from Node.

>     2\. Turns off webpack reference detection on "require", when `dynamicRequire` is used

Yes, as Webpack gets confused and doesn't know what to include since everything could be required, and it replaces that call to something that always throws with `MODULE_NOT_FOUND`.

>     3\. Should add a root in the reachability graph for webpack to include the modules referenced as "entryPoints" in package.json. (I haven't seen the code that does that, though).

`entryPoints` is necessary for our child process spawning/forking to work. Webpack doesn't have the notion of what is being spawned as a different process and hence should be a separate script on disk for Node to run. This information will be used by clients bundling the backend, as I did for my [browser Blueprint PoC](https://github.com/eclipse-theia/theia-blueprint/blob/dbe08f1a1156d7cc4fd30d9a6a92d970315a4d1b/applications/browser/node-webpack.config.js#L19-L52).

packages/plugin-ext/src/hosted/node/metadata-scanner.ts
paul-marechal: `scanner` will be `undefined`, which is checked after:

```ts
        if (!scanner) {
            throw new Error('There is no suitable scanner found for ' + plugin.name);
        }
```

paul-marechal:
@tsmaeder I added two commits:

1. Adds a `EntryPointRegistry`: the goal is to query entry points from the name given when declaring them using the key/value map in the `package.json`. To avoid collisions, I added the name of package that contributes the entry point to the entry point name. See `getEntryPoint` callsites to see it in action.
2. Adds some API sugar using inversify named injections. TBH I think we should keep it as I really like the code with it, but I placed it in a separate commit so that it can be easily removed.

Some more information about this design: I plan on overriding the `EntryPointRegistry` in Theia Blueprint in order to craft valid paths when bundling. Once bundled, the original file structure is lost, so the idea will be to rebind the registry to do the proper path mapping.

dev-packages/application-package/src/extension-package.ts
paul-marechal: I did not bother trying to de-duplicate yet. My reasoning is that there's a low chance of duplication here.

paul-marechal:
CI is flaky and fails with HTTP error 503/502 when downloading plugins.

dev-packages/application-package/src/extension-package.ts
tsmaeder: The point is not the deduplication, the main point is that we're requiring a unique id when we don't need one.

packages/core/src/node/entry-point.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const EntryPointsRegistry = Symbol('EntryPointsRegistry');
+export interface EntryPointsRegistry {
+    getEntryPoint(name: string): string
+}
+
+@injectable()
+export class EntryPointsRegistryImpl implements EntryPointsRegistry {
+
+    @inject(ApplicationPackage)
```
tsmaeder, line 43: "ApplicationPackage" comes from the dev-packages folder. I don't think we should require it at runtime. Wouldn't the correct way be to generate some code during `theia build`?

dev-packages/application-package/src/application-package.ts
tsmaeder: > it only works when bundling browser Blueprint because everywhere we happen to spawn paths written as path.join(__dirname, 'entry-point-name')

I'm not sure I follow your argument here: in one of our extensions in Che we _do_ have a require of the form: `@eclipse-che/....`. Why would that not work?

tsmaeder:
I'm not sure if the current solution is not a bit of overkill. What we're trying to achieve is this:

1. We want to know what to write as "entry-points" into the webpack config
2. We want those references to be relative to some module in order to be independent of the concrete layout of the project

Maybe it's enough to simply declare entry-points in the package.json and have a `dynamicRequire` that takes a module id and a module-relative id or path.

It's very hard to automatically tell what entrypoints need to be included with webpack anyway: If you look at `NsfwFileSystemWatcherServiceProcessOptions`, the entrypoint can be overridden by a Theia extensions: so if we're just automatically adding all the declared entrypoints from the extensions, we'll include both the original and the override entrypoints. So maybe just having an easy way to discover them for manual inclusion is sufficient.

dev-packages/application-package/src/application-package.ts
paul-marechal: When bundling using Webpack all bundles will live in the same folder, so everyone is only reachable by doing `path.join(__dirname, 'entry-point-name')`. If paths were something like `path.join(__dirname, '..', 'some-folder', 'entry-point-name')` it would cause issues once bundled because it would walk out of the bundle folder. From this point the second pass done with `pkg` will most likely fail too because of the erroneous Webpack bundling.

paul-marechal:
> It's very hard to automatically tell what entrypoints need to be included with webpack anyway

Given how we currently handle spawning and overriding entry points yes.

> If you look at `NsfwFileSystemWatcherServiceProcessOptions`, the entrypoint can be overridden by a Theia extensions [...]

Overriding via Theia extensions/inversify is a mean to an end. If we want to truly automatically enumerate entry points I think we would need to go full declarative and deprecate the inversify bindings for those.

> I'm not sure if the current solution is not a bit of overkill.

Understandable, for now I'll put this use case on ice. There's much more value in just replacing the dynamic require calls, so I'll close this PR in favor of a simplier one: https://github.com/eclipse-theia/theia/pull/9660

PR #9631, opened 2021-06-22
terminal: normalize envs before merging

paul-marechal:
Environment variables are case insensitive on Windows so we'll align
everything to use uppercase when on Windows.

#### How to test

Not really sure. My `PATH` environment variable on Windows is named `Path` in `process.env` for some reason. This caused issues when running the `@theia/terminal` test suite as at some point we expect to fetch `PATH` from an object that should look like `process.env` but failed as the name was `Path` instead of `PATH` in that new js object.

#### Review checklist

#### Reminder for reviewers


paul-marechal:
cc @alvsan09

alvsan09:
> Case should not be important, so we'll align everything to
> use uppercase.

My understanding is that the applications that use the variable may rely on the original casing, it's out of our control.
if that's the case, we should preserve the casing. Would you agree?

paul-marechal:
The general wisdom is that environment variables should be uppercased. When it comes to `process.env`, it is a special object which will ignore the case when accessing members, e.g. `process.env.PATH` is equivalent to `process.env.Path` or `process.env.path`. Since we create new objects which are not special like `process.env` is, we have to pick a case. In this case, it's uppercase.

packages/terminal/src/node/shell-process.ts
```diff
@@ -123,3 +124,23 @@ export class ShellProcess extends TerminalProcess {
...
+ * Normalize an environment map for a given OS.
+ *
+ * On Windows it will uppercase all keys.
+ *
+ * @param env Environment variables map to normalize.
+ * @param platform Platform to normalize for.
+ * @returns New object with normalized environment variables.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function uppercaseEnv<T extends Record<string, any>>(env: T, platform: NodeJS.Platform = process.platform): T {
```
alvsan09, line 138:
```suggestion
export function normalizeCaseEnv<T extends Record<string, any>>(env: T, platform: NodeJS.Platform = process.platform): T {
```

packages/terminal/src/node/shell-process.ts
```diff
@@ -123,3 +124,23 @@ export class ShellProcess extends TerminalProcess {
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const normalized: any = {};
+    for (const [key, value] of Object.entries(env)) {
+        const normalizedKey = platform === 'win32' ? key.toUpperCase() : key;
```
alvsan09, line 142:
```suggestion

```
we can add a top if statement instead, e.g.
if ( platform !== 'win32') { 
   return env;
}

vince-fugnitto (commented):
@paul-marechal is there a need to add tests to `shell-process.spec.ts` to ensure the method works successfully under different conditions, and for different operating systems?

alvsan09 (commented):
just a couple of suggestions.

packages/terminal/src/node/shell-process.ts
paul-marechal: I'll slightly change that to always return a new object.

paul-marechal:
I renamed `uppercaseEnv` to `normalizeEnv` and added tests for it.

packages/terminal/src/node/shell-process.ts
```diff
@@ -103,17 +103,22 @@ export class ShellProcess extends TerminalProcess {
     }
 }
 
+export interface MergeProcessEnvOptions {
```
alvsan09, line 106: Why are we defining two interfaces with the same signature?  i.e. `NormalizeEnvOptions`, and `MergeProcessEnvOptions`,

paul-marechal: It sounds better API-wise: The functions happen to share a common option field now, but it doesn't mean they actually depend on the same options. Hence why I kept it separated here.

alvsan09: Thanks for the explanation !!

packages/core/src/node/environment-utils.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { injectable } from 'inversify';
+
+export interface PlatfomOptions {
```
alvsan09, line 19:
```suggestion
export interface PlatformOptions {
```

alvsan09 (commented):
It looks good to me, 
Just a typo, see inline comment!

alvsan09 (approved):
Looks go to me !! :+1:

packages/core/src/node/environment-utils.ts
```diff
@@ -0,0 +1,74 @@
...
+    normalizeEnv<T>(env: Record<string, T>, options: NormalizeEnvOptions = {}): Record<string, T> {
+        const {
+            platform = process.platform
+        } = options;
+        if (platform !== 'win32') {
+            return { ...env };
+        }
+        const normalized: Record<string, T> = {};
+        for (const [key, value] of Object.entries(env)) {
+            normalized[key.toUpperCase()] = value;
```
colin-grant-work, line 70: Apparently you'd have to do some fancy configuration and perhaps compile Node yourself to make it register a locale on the backend, but perhaps this should be `toLocaleUppercase()` in case anyone goes to such trouble?

packages/core/src/node/environment-utils.ts
```diff
@@ -0,0 +1,74 @@
...
+     */
+    mergeProcessEnv(env: Record<string, string | null>, options: MergeProcessEnvOptions = {}): Record<string, string> {
+        env = this.normalizeEnv(env, { platform: options.platform });
+        // eslint-disable-next-line no-null/no-null
+        const mergedEnv: Record<string, string> = Object.create(null);
+        for (const [key, value] of Object.entries(this.normalizeEnv(process.env, { platform: options.platform }))) {
+            // Ignore keys from `process.env` that are overridden in `env`. Accept only non-empty strings.
+            if (!(key in env) && value) { mergedEnv[key] = value; }
+        }
+        for (const [key, value] of Object.entries(env)) {
```
colin-grant-work, line 45: Are we happy not normalizing these, since they come from the user in any case?

colin-grant-work (approved):
Looks reasonable to me. One minor comment and a question, but nothing to hold this up.

packages/core/src/node/environment-utils.ts
paul-marechal: I normalized `env` first thing in `mergeProcessEnv`.

paul-marechal:
CI fails with HTTP error 503 service unavailable when downloading plugins...

Issue #9578, opened 2021-06-10
Preserve the collapse state of a tree-view


alvsan09:
### Feature Description:
Storing the collapse state of view is a feature available in `vscode` as can be seen in 'gif' below, 
this shall be possible to implement in `Theia` as well.
The implementation may be similar to #9539.

![preserving_collapse_state](https://user-images.githubusercontent.com/76971376/121598729-19702700-ca10-11eb-8b7d-ef40fbf9d437.gif)

PR #9636, opened 2021-06-23
Keep the collapse state of a tree-view

alvsan09:
The 'collapse' state of a tree view is preserved, however setting the
`hidden` state to `false` was incorrectly cascading to `collapse`
`false`.

NOTE: The issue can be reproduced using a `tree-view` provided by a plugin
which uses a `when` clause to open it.

Fixes: #9578 

#### What it does

Removes a condition which incorrectly caused this types of views to expand when the `hidden` state was set to `false`

#### How to test

1) Use the following [tree view extension example](https://github.com/microsoft/vscode-extension-samples/blob/main/tree-view-sample/README.md) or the 'npm' scripts view.
The extension provides a tree-view called `JSON OUTLINE` which is activated when the open file in editor is a `json` file i.e. It uses  a `when` clause. 
2) Open a 'json'  file in editor to make sure the 'JSON OUTLINE' opens.
3) Follow the sequence as shown in the following `.gif` files to reproduce and then verify the fix.
Keep-collapsed-fails

keep_collapses_works

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/view-container.ts
+++ b/packages/core/src/browser/view-container.ts
@@ -756,9 +756,6 @@ export class ViewContainerPart extends BaseWidget {
             return;
         }
         super.setHidden(hidden);
-        if (!this.isHidden) {
-            this.collapsed = false;
-        }
     }
 
     get canHide(): boolean {


msujew (approved):
@alvsan09 Thank you for your contribution.

I can confirm that the issue is present on master and is fixed by your changes:

Closing the `npm` container and reloading the application will re-open the container on `master`. Repeating the same with your changes keeps the `npm` container closed after reloading.

I wasn't able to find any regressions with this approach.

vince-fugnitto (approved):
I confirmed that the _tree-view part_ collapsed state is preserved on reloads under different conditions:
- confirmed that collapsing all parts are preserved on reload.
- confirmed under different conditions that parts are preserved on reload (collpased + uncollapsed).
- confirmed that the `reset workbench layout` properly resets the layout, and parts explicitly collapsed by users are reset.

PR #9644, opened 2021-06-27
Fix #9203: Drag and drop sections between views

### LINKED ISSUE #9203: Drag and drop sections between views
### Feature Description:

In VS Code it is possible to drag and drop sections between views. In theia it is currently possible only in the same view.

Scenario:

1. install Docker vscode extension
2. Open Docker view 
3. Try to drag images section into the explorer view - it can be done in vscode but not in theia

Example:

view

EstherPerelman:

This PR enable to group multiple views at the same container without the need to navigate between the containers

#### What it does
Fix #9203 

#### How to test
1. Install Docker extension vsix from https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker
2. Try Drag & Drop views between containers:
view
 
3. Attentions: 
   * d&d for `Extensions` view is disabled.
   * dragging is allowed only inside the container for: `File explorer` view part, `Search` view part, `SCM` view part (you still able to drop views from other containers on those parts)

#### Review checklist

#### Reminder for reviewers


EstherPerelman:
This PR is not completed yet, In order to enable d&d for the `Search` view I wrapped it with `ViewContainer` widget but after this changes the search results not appear, Any one can help? I think I'm missing something on [the bindings file](https://github.com/EstherPerelman/theia/blob/438c2a12045edb892e962a11722d24897a3d4193/packages/search-in-workspace/src/browser/search-in-workspace-frontend-module.ts)
Any help will be appreciated!

colin-grant-work:
@EstherPerelman, are you still having trouble with the Search-in-Workspace widget? When I pull your code, I'm able to search and get results. However, I'm not able to drop anything in that widget. Have you removed the code that placed it in a ViewContainer?

EstherPerelman:
> @EstherPerelman, are you still having trouble with the Search-in-Workspace widget? When I pull your code, I'm able to search and get results. However, I'm not able to drop anything in that widget. Have you removed the code that placed it in a ViewContainer?

@colin-grant-work Thank for trying to help! I'm still having this trouble, It's very strange I haven't removed any code - for me it still enables dropping into it but not showing the search results, Do the other containers enable d&d for you?

EstherPerelman:
Hi all, If there's no solution for fixing the problem with `search` container - is it possible to push it without the ability to use the `dnd` inside `search` container?
cc @colin-grant-work

packages/core/src/browser/view-container.ts
```diff
@@ -653,6 +830,32 @@ export namespace ViewContainer {
         }
         return 'vertical';
     }
+
+    export function isSupportedDND(event: DragEvent): boolean {
+        const { dataTransfer } = event;
+        return !!dataTransfer && dataTransfer.types.indexOf(DataTransfers.DND) > -1;
```
colin-grant-work, line 836: Here and below, please use `includes()` rather than `indexOf() ... > -1`.

colin-grant-work (commented):
With minor exceptions (one immediately below, and one in the comments), both of them apparently related to the special status of the debug session widget, this is working pretty well for me. The logic for identifying and moving parts is pretty involved, but I don't see any easy simplifications. I will see if I can work out what's going on with the search-in-workspace widget - I suspect that it's another special case - but I have no objection in principle to merging this and leaving the SIW as a later to-do.

I got somewhat inconsistent behavior with the highlighting for drop. In a plugin view container (GitLens, in my case), and the internal SCM widget, I got highlighting when dropping parts from other containers. However, in the debug view container, after collapsing all of the parts, there was no highlighting when I dragged parts from other containers. The drop functionality still worked fine, though.

packages/core/src/browser/view-container.ts
```diff
@@ -246,22 +268,31 @@ export class ViewContainer extends BaseWidget implements StatefulWidget, Applica
...
-        const description = this.widgetManager.getDescription(widget);
-        const partId = description ? JSON.stringify(description) : widget.id;
-        const newPart = this.createPart(widget, partId, options);
+        this.toRemoveWidgets.set(newPart.wrapped.id, toRemoveWidget);
+        toRemoveWidget.push(Disposable.create(() => this.toRemoveWidgets.delete(newPart.wrapped.id)));
         this.registerPart(newPart);
-        if (newPart.options && newPart.options.order !== undefined) {
-            const index = this.getParts().findIndex(part => part.options.order === undefined || part.options.order > newPart.options.order!);
+        if (insertIndex !== undefined || (newPart.options && newPart.options.order !== undefined)) {
+            const index = insertIndex !== undefined ? insertIndex
```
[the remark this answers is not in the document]
EstherPerelman: But in case `insertIndex` is equals to `0` it won't return it, Am I wrong?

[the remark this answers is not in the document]
colin-grant-work: Unlike `||`, `??` (a relatively new operator in TypeScript) only returns the second value if the first is value is 'nullish' (`undefined` and `null`). If the first value is any other value, including a 'falsey' value, [`??` returns the first value](https://www.typescriptlang.org/play?#code/MYewdgziA2CmB00QHMAUBXMATWAzAlmLFgAQD8ZJAjAJQDcJA9I9QFCiQwJJpjrTRylWg2ZsOUOIhSoA5LKHV6TFvPbhJ3GQAZFIlSW3rOUnqlwBDaBFh7lYy9dhA).

image

EstherPerelman:
@colin-grant-work Thank you so much for the help and review! setting the height of `search-in-workspace` to 100% did solve the problem! 
I added more fixes and solved according to you comments... 
Thanks👍
(Responds late because I haven't been at work for a week...)

packages/core/src/browser/style/tabs.css
```diff
@@ -320,6 +320,14 @@ body.theia-editor-highlightModifiedTabs
...
 }
 
+.p-TabBar-tab.drop-target-top {
+  border-top: 2px solid white !important;
+}
+
+.p-TabBar-tab.drop-target-bottom {
+  border-bottom: 2px solid white !important;
+}
+
```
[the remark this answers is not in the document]
colin-grant-work: I think the problem still exists. I think we probably need to set the `box-sizing` property for these tabs to be `border-box` (hopefully that will fix it - it may be more complex) so that they retain the same total size regardless of the presence or absence of the border.

To test, grab a view container part, and then drag up and down the sidebar over the various other tabs. The borders for the tabs should appear and disappear, but the icons should remain in the same position.

packages/core/src/browser/view-container.ts
```diff
@@ -586,36 +704,115 @@ export class ViewContainer extends BaseWidget implements StatefulWidget, Applica
...
+
+    protected isDebugWidget(widget: Widget): boolean {
+        return 'sessionWidget' in widget;
+    }
+
+    protected getDebugWidgetId(): string | undefined {
+        if (this.isDebugContainer()) {
+            return this.parent!.parent!.id;
+        }
+    }
```
colin-grant-work, line 814: If possible, we should avoid anything debug related in `core`. Perhaps define an interface here that the debug widget could implement that would get around these problems?

colin-grant-work:
This is working very nicely for me. Now only the comments about the shifting icons and references to `debug` in `core` remain. 👍

EstherPerelman:
> This is working very nicely for me. Now only the comments about the shifting icons and references to `debug` in `core` remain. 👍

Great! I'm still working on this because the restore after refresh(f5) not working well, I still didn't succeed to see it working perfect, Maybe adding a `ViewContainerManager ` will solve the problem...

packages/core/src/browser/view-container.ts
EstherPerelman: @colin-grant-work 

1. should I create an interface with the name `debug` (referring only the name)?
2. Which file is the right place to put the interface in?

colin-grant-work: What I have in mind is something like
In view-container.ts:
```typescript
export interface ViewContainerContainer extends Widget {
    getContainedViewContainer(): ViewContainer
}

export namespace ViewContainerContianer {
    export const is = (candidate: Widget): candidate is ViewContainerContainer => {
       const maybeViewContainerContainer = candidate as ViewContainerContainer;
       return 'getContainedViewContainer' in maybeViewContainerContainer && typeof maybeViewContainerContainer.getContainedViewContainer === 'function';
};
```

Then in `debug-widget.ts`
```typescript
export class DebugWidget extends BaseWidget implements StatefulWidget, ApplicationShell.TrackableWidgetProvider, ViewContainerContainer {

getContainedViewContainer(): ViewContainer {
    return this.sessioWidget.viewContainer; 
    // this is protected, so it would either have to made public, or SessionWidget would implement ViewContainerContainer and you could return this.sessionWidget.getContainedViewContainer();
}
```

That way, the `debug` package would know about `core`, but `core` wouldn't have to know anything about `debug` - not even the name.

EstherPerelman:
@colin-grant-work 

- [ ] **A:** 

> If possible, we should avoid referring to anything related to debug in core. Perhaps define an interface here that the debug widget could implement that would get around these problems?

I have started to implement as you wrote but saw that it doesn't fit all places(where I have only the `ViewContainer` widget without information about the parent parent ...), Anyway I have found a solution by creating `draggingPartManager` which make things easier...

- [ ] **B:**

> I think the problem still exists. I think we probably need to set the `box-sizing` property for these tabs to be `border-box` (hopefully that will fix it - it may be more complex) 

I have spent time on this but `box-sizing` setting and others just didn't solve it (while setting `outline` instead of `border` the icon didn't move but the appearance was not uniform)

- [ ] **C:**

Restore after refresh suppose to work well now...

EstherPerelman:
@colin-grant-work 
I found a solution for the issue with the border 
**So - the PR is full ready now for another review**
Thanks for all advices and comments

colin-grant-work:
Another VSCode feature that doesn't necessarily need to be implemented here, but that you may want to consider is the draggability of parts that are the sole child of their view container. E.g. in VSCode, if the Search-in-Workspace widget is the sole child of the search container, you can start a drag on the `ViewContainer` title that behaves the same as dragging the `ViewContainerPart`'s title.

packages/core/src/browser/style/tabs.css
```diff
@@ -38,6 +40,25 @@
...
+  border-top: var(--theia-dragover-tab-border-width) solid transparent !important;
+  border-bottom: var(--theia-dragover-tab-border-width) solid transparent !important;
+}
+
+.p-TabBar[data-orientation='vertical'] .p-TabBar-tab.drop-target-top {
+  border-top-color: white !important;
+}
+
+.p-TabBar[data-orientation='vertical'] .p-TabBar-tab.drop-target-bottom {
+  border-bottom-color: white !important;
```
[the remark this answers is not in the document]
EstherPerelman: @colin-grant-work Could you please explain more on how to do it, Taking an existing variable for `white` or just create a new one? does it need to be sync in some places where applying a new theme?

packages/core/src/browser/view-container.ts
```diff
@@ -276,22 +351,29 @@ export class ViewContainer extends BaseWidget implements StatefulWidget, Applica
 
     protected readonly toRemoveWidgets = new Map<string, DisposableCollection>();
 
-    addWidget(widget: Widget, options?: ViewContainer.Factory.WidgetOptions): Disposable {
+    createPartId(widget: Widget): string {
+        const description = this.widgetManager.getDescription(widget);
+        return widget.id || JSON.stringify(description);
```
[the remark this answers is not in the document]
EstherPerelman: @colin-grant-work the `id` was created that way (with the `|| description`) before this PR changes. I have just moved it into a function because I use it in 2 places...

packages/debug/src/browser/view/debug-session-widget.ts
```diff
@@ -82,17 +88,18 @@ export class DebugSessionWidget extends BaseWidget implements StatefulWidget, Ap
...
-        this.viewContainer.addWidget(this.stackFrames, { weight: 20 });
-        this.viewContainer.addWidget(this.variables, { weight: 10 });
-        this.viewContainer.addWidget(this.watch, { weight: 10 });
-        this.viewContainer.addWidget(this.breakpoints, { weight: 10 });
+        this.viewContainer.setTitleOptions(DEBUG_VIEW_CONTAINER_TITLE_OPTIONS);
+        this.viewContainer.addWidget(this.threads, { weight: 30, disableDraggingToOtherContainers: true });
+        this.viewContainer.addWidget(this.stackFrames, { weight: 20, disableDraggingToOtherContainers: true });
+        this.viewContainer.addWidget(this.variables, { weight: 10, disableDraggingToOtherContainers: true });
+        this.viewContainer.addWidget(this.watch, { weight: 10, disableDraggingToOtherContainers: true });
+        this.viewContainer.addWidget(this.breakpoints, { weight: 10, disableDraggingToOtherContainers: true });
```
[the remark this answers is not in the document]
EstherPerelman: @colin-grant-work I disabled dragging `debug` widgets because there's a problem with restoring them to the last container they were dragged to after refresh, Do you think it still worth to enable dragging theme although they won't get restored after refresh (but return to the `debug` container?
About the icon - it's very strange because I do show this icon when trying to drag such views...

EstherPerelman:
> Another VSCode feature that doesn't necessarily need to be implemented here, but that you may want to consider is the draggability of parts that are the sole child of their view container. E.g. in VSCode, if the Search-in-Workspace widget is the sole child of the search container, you can start a drag on the `ViewContainer` title that behaves the same as dragging the `ViewContainerPart`'s title.

@colin-grant-work I know about this, it also available vscode to drop the part header on the sidebar and a new tab will be created,
I'm afraid of the mess with creating new viewContainers, I'll need to create a new id to a duplicate container - then to get it by this id in other places... don't know how it can be managed...

packages/navigator/src/browser/navigator-widget-factory.ts
```diff
@@ -50,7 +50,8 @@ export class NavigatorWidgetFactory implements WidgetFactory {
         order: 1,
         canHide: false,
         initiallyCollapsed: false,
-        weight: 80
+        weight: 80,
+        disableDraggingToOtherContainers: true
```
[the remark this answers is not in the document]
EstherPerelman: > About the icon - it's very strange because I do show this icon when trying to drag such views

packages/core/src/browser/view-container.ts
```diff
@@ -531,6 +631,38 @@ export class ViewContainer extends BaseWidget implements StatefulWidget, Applica
...
+            return;
+        }
+        const newPart = this.createNewPartFromExisting(part);
+        this.attachNewPart(newPart, index);
+        part.fireOnDidRecreate(newPart, this);
+    }
+
+    protected moveBetweenContainers(fromIndex: number, toIndex: number, fromContainer: ViewContainer): void {
+        const partToMove = fromContainer.getParts()[fromIndex];
+        fromContainer.removeWidget(partToMove);
```
msujew, line 646: I am currently in the process of finding out what causes the weird layouting behavior after reloading.

One thing I noticed during debugging is that this `fromContainer.removeWidget(partToMove);` call never does anything. This is due to [this line](https://github.com/eclipse-theia/theia/blob/b6b0f1be2375fd454fefbbbf28895bdffe06fc69/packages/core/src/browser/view-container.ts#L372) which uses the id of the wrapped element. But here we call remove on the wrapper instead, which is not part of the `toRemoveWidgets` map.

I'll continue investigating what's wrong here.

packages/core/src/browser/style/tabs.css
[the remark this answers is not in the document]
colin-grant-work: @EstherPerelman, I believe the color of the drag-drop border in VSCode is the same as the active border, so use this as your guide:

https://github.com/eclipse-theia/theia/blob/e579c515614c3d15a7e2b529d69b66a1d47a8065/packages/core/src/browser/style/sidepanel.css#L84-L87

packages/debug/src/browser/view/debug-session-widget.ts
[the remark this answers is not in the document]
colin-grant-work: I think this can also be omitted for now if it's going to be difficult to implement - the general functionality is present. Is there a reason the approach taken to restoring the state of other view containers that contain migratory view container parts doesn't apply here?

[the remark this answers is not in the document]
EstherPerelman: Only for the `debug` widgets [this property] (https://github.com/EstherPerelman/theia/blob/b6b0f1be2375fd454fefbbbf28895bdffe06fc69/packages/core/src/browser/view-container.ts#L499) is correctly save on storing but is coming `undefined` on [restore](https://github.com/EstherPerelman/theia/blob/b6b0f1be2375fd454fefbbbf28895bdffe06fc69/packages/core/src/browser/view-container.ts#L519) which make it impossible to create it in the right place, I can guess it has something with the different behavior of the `debug` session widget.

packages/core/src/browser/view-container.ts
EstherPerelman: @msujew thanks for helping, I really missed that `removeWidget` was never called, I have fixed it but the issue still exist I'm trying my best..., Have you discovered some more hints?

packages/core/src/browser/style/tabs.css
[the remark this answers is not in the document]
EstherPerelman: @colin-grant-work absolutely right! thank you!

packages/core/src/browser/view-container.ts
msujew: Yeah, I spent an hour trying to fix it but to no avail as well. 

Looking at the html source, I noticed that the last `ViewContainerPart` isn't rendered correctly. Somehow, phosphor does not correctly render its top and height style properties. That's why it shows up at the top and basically shadows every other element in the `ViewContainer`. I couldn't figure out why this happens though. I'll take another look on monday.

EstherPerelman: @msujew Thank you, I have also noticed that it has no `top` style and its handle is hidden (might not right in all cases)
Actually if there will be no solution we are able to completely disable drag & drop for the `debugger` container  (which is the only one who makes these troubles) as done for `Extensions`, Don't we?
cc @colin-grant-work

EstherPerelman:
> Actually if there will be no solution we are able to completely disable drag & drop for the debugger container (which is the only one who makes these troubles) as done for Extensions, WDYS? 
> (I have already fixed all other issues)

@msujew @colin-grant-work

packages/debug/src/browser/view/debug-session-widget.ts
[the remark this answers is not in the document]
colin-grant-work: So it looks like what's happening here is this. When the `ShellLayoutRestorer` is storing the layout, if it encounters a `Widget` it tries to convert that widget to a description:

https://github.com/EstherPerelman/theia/blob/84bca9b36ac6a2029a57d9406d8c13f695aa0397/packages/core/src/browser/shell/shell-layout-restorer.ts#L205-L221

To do that, it asks the `WidgetManager` to get the description, and that only works _if the widget was created through the `WidgetManager`_

https://github.com/EstherPerelman/theia/blob/84bca9b36ac6a2029a57d9406d8c13f695aa0397/packages/core/src/browser/widget-manager.ts#L237-L250

But, the `DebugSessionWidget` parts are created without going through the `WidgetManager`

https://github.com/EstherPerelman/theia/blob/84bca9b36ac6a2029a57d9406d8c13f695aa0397/packages/debug/src/browser/view/debug-session-widget.ts#L41-L45

So they always wind up with no description. I think that that is a bad design on the part of the `debug` package, because it means that it expects to create a whole set of widgets for each session and then hide and reveal them depending on which session is selected. Instead, I think it would be better to have the same widgets able to hook up to whatever session is selected, but that's a major refactor of the `debug` package, so I don't think it should be undertaken as part of this PR. It seems like disabling drag-drop there is probably best for now, then.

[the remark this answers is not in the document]
EstherPerelman: Yea, all the issues with the `debug` container are related to the missing creation by `widgetManager`, During the tries to solve this I thought to add it manually to the widgets list (not by factory creation), but it seemed to be incorrect according to the `widgetManager` behavior, 
Actually I didn't catch that resorting `debug` widgets issue is also related to this, thank you for the clear explanation!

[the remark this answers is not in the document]
msujew: I'm also in favor of disabling dnd for debug widgets and proceeding with this PR 👍

packages/navigator/src/browser/navigator-widget-factory.ts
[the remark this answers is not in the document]
colin-grant-work: When I was testing yesterday, I still never saw the icon for forbidden dropping. Only an arrow. Not quite sure what the intent is.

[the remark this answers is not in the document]
EstherPerelman: @colin-grant-work is that what you are seeing?
forbidden dropping in `vscode`:

https://user-images.githubusercontent.com/71729183/129927499-cbc67545-af52-4bb2-8f71-1c2472bc84df.mp4

forbidden drooping in `theia`:

https://user-images.githubusercontent.com/71729183/129927717-3a1804d5-6a06-48ea-988d-bd251588863e.mp4

EstherPerelman:
> Let me take today to look into this in some detail. I think the bug likely points to something that needs improvement anyway - either in the handling of the dragging or in the debug area specifically - and I'd rather find that now than later. If it turns out to be very hard to diagnose or fix, though, then we can likely move forward with the work-around.

@colin-grant-work did you find something or should I move forward with the workaround?

colin-grant-work:
> @colin-grant-work did you find something or should I move forward with the workaround?

Go ahead with the work-around - I don't think it's practical to do the necessary refactor as part of this PR.

packages/navigator/src/browser/navigator-widget-factory.ts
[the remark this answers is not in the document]
EstherPerelman: @colin-grant-work what about this?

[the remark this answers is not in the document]
colin-grant-work: This is what I see in Electron:

bad-icon

The arrow icon disappears, but the forbidden icon never appears.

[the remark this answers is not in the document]
EstherPerelman: Are you on `Windows`?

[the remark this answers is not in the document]
colin-grant-work: That's in a Linux (RHEL7) environment

EstherPerelman:
@colin-grant-work 
I uploaded the work-around for the `debug` widget
I have implemented `dropeffect = none` for all the non droppable targets, I have confirmed it both on Windows `chrome` & `edge` & `firefox` browsers and Electron `chrome`. It looks like it's not working in your environment what can I do with that? `dnd` is very well supported on modern browsers...

colin-grant-work:
> @colin-grant-work
> I uploaded the work-around for the `debug` widget
> I have implemented `dropeffect = none` for all the non droppable targets, I have confirmed it both on Windows `chrome` & `edge` & `firefox` browsers and Electron `chrome`. It looks like it's not working in your environment what can I do with that? `dnd` is very well supported on modern browsers... what to do about the old ones?

It seems to be something about the OS rather than the browser. The forbidden icon doesn't seem to show up in any browser, no matter how new, or in Electron. So that doesn't seem to be our problem to solve, and I do see the forbidden icon in GitPod. 👍

EstherPerelman:
> It seems to be something about the OS rather than the browser. The forbidden icon doesn't seem to show up in any browser, no matter how new, or in Electron. So that doesn't seem to be our problem to solve, and I do see the forbidden icon in GitPod. 👍

Great, so is this PR prepared for merge?

packages/search-in-workspace/src/browser/search-in-workspace-factory.ts
```diff
@@ -0,0 +1,60 @@
...
+export class SearchInWorkspaceFactory implements WidgetFactory {
+
+    static ID = SEARCH_VIEW_CONTAINER_ID;
+
+    readonly id = SearchInWorkspaceFactory.ID;
+
+    protected searchWidgetOptions: ViewContainer.Factory.WidgetOptions = {
+        canHide: false,
+        initiallyCollapsed: false,
+        disableDraggingToOtherContainers: true
```
colin-grant-work, line 43: I believe this widget is dragable in VSCode. Is there a reason to make it non-dragable here?

colin-grant-work:
@EstherPerelman, I believe this PR is essentially complete. Just one last question about the SIW widget.

packages/search-in-workspace/src/browser/search-in-workspace-factory.ts
EstherPerelman: You are right, things were changed since I needed to disable it, I pushed the fix and rebased, could you please merge🤝?

colin-grant-work (approved):
This looks good to me, and the problem cases seem to have been worked out or worked around. @msujew, would you mind doing another run through before I merge, just in case I've missed anything?

EstherPerelman:
A note: for me when moving from older version to this PR version I must run `f1` -> `View: Reset workbench layout` to build the layout correctly, So it might affect also the users/companies which adopt the next version, They need to be informed to run the `f1` -> `View: Reset workbench layout` command before starting, or maybe you do it automatically?

msujew:
@esterw1109 This should be gracefully handled by an `ApplicationShellLayoutMigration`. I saw that you added one but for version `3`. Currently we are on version `4`, so the migration should be using version `5` (probably `6`, as #9864 adds a migration to version 5 and hopefully will be merged before the Thursday release)

EstherPerelman:
> @esterw1109 This should be gracefully handled by an `ApplicationShellLayoutMigration`. I saw that you added one but for version `3`. Currently we are on version `4`, so the migration should be using version `5` (probably `6`, as #9864 adds a migration to version 5 and hopefully will be merged before the Thursday release)

@msujew Meantime I updated to version 5, I really hope you will merge it before the Thursday release, Is there a reason to postpone it? 
cc @colin-grant-work

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
...
 - [vsx-registry] `VSXExtensionsContribution` no longer implements `TabBarToolbarContribution` and is not bound as such. Extensions of the class that expect such behavior should reimplement it with caution. See caveats in PR. [#9798](https://github.com/eclipse-theia/theia/pull/9798)
 - [core] `handleExpansionToggleDblClickEvent` in `TreeWidget` can no longer be overridden. Instead, `doHandleExpansionToggleDblClickEvent` can be overridden. [#9877](https://github.com/eclipse-theia/theia/pull/9877)
+- [view-container] `ViewContainerPart` constructor takes new 2 parameters: `originalContainerId` and `originalContainerTitle`, The existing `viewContainerId` parameter has been renamed to `currentContainerId` to enable drag & drop views. [#9644](https://github.com/eclipse-theia/theia/pull/9644)
```
vince-fugnitto, line 14:
```suggestion
- [view-container] `ViewContainerPart` constructor takes new 2 parameters: `originalContainerId` and `originalContainerTitle`. The existing `viewContainerId` parameter has been renamed to `currentContainerId` to enable drag & drop views. [#9644](https://github.com/eclipse-theia/theia/pull/9644)
```

packages/plugin-ext/src/main/browser/view/plugin-view-registry.ts
```diff
@@ -466,34 +470,72 @@ export class PluginViewRegistry implements FrontendApplicationContribution {
         }
         for (const viewId of this.getContainerViews(viewContainerId)) {
             const identifier = this.toPluginViewWidgetIdentifier(viewId);
+            // Keep exising widget in its current container and reregister its part to the plugin view widget events.
```
vince-fugnitto, line 473: minor: 

```suggestion
            // Keep existing widget in its current container and reregister its part to the plugin view widget events.
```

colin-grant-work:
@msujew, apart from the migration question, which @EstherPerelman seems to have addressed, are you satisfied with the behavior of this PR? If so, I'll merge later today.

msujew (approved):
Looks good to me 👍 

- [x] Dragging extensions view parts between most view containers is possible
- [x] Reloading keeps the widget in a correct state
- [x] Cannot drag into invalid view containers which is marked with the appropriate cursor

vince-fugnitto (commented):
I've been getting some odd UI behaviors with the changes, perhaps something to test further:
- `dnd` (blue highlighting) when attempting to drag-and-drop between views does not always appear (ex: tested between scm and explorer)
- I ended up in a state where view-parts were overlapped:

image

- I also ended up in a state with overlap + npm scripts never being populated

image

I was able to reproduce after performing multiple dnd.

image

colin-grant-work:
I was able to reproduce @vince-fugnitto's results, though I'm not exactly sure what the conditions were. I was dragging various widgets back and forth, and then switched to the Explorer view _without_ dragging anything to it, and its layout was broken:

image

EstherPerelman:
Hi @vince-fugnitto @colin-grant-work @msujew 
After deep search - All the bugs with the broken layout suppose to be fixed by the code  I added [here](https://github.com/EstherPerelman/theia/blob/c40d348f68aad947cdab0f634a4068e6b0cd619b/packages/core/src/browser/view-container.ts#L1291-L1294) and [here](https://github.com/EstherPerelman/theia/blob/c40d348f68aad947cdab0f634a4068e6b0cd619b/packages/core/src/browser/view-container.ts#L1277-L1286)
I have also changed the code to use `phosphor` drag and drop support which reduce the amount of code and make things clearer...
I have updated the description - have a look at the updated `Attentions`
I'll be happy with your reviews!

EstherPerelman:
@colin-grant-work @msujew can you please approve last changes?

vince-fugnitto:
> @colin-grant-work @msujew can you please approve last changes?

@EstherPerelman there are a lot of conflicting files that would need to be addressed first.

EstherPerelman:
> > @colin-grant-work @msujew can you please approve last changes?
> 
> @EstherPerelman there are a lot of conflicting files that would need to be addressed first.

@vince-fugnitto the conflicts are resolved now..
cc @colin-grant-work @msujew

packages/search-in-workspace/src/browser/search-layout-migrations.ts
```diff
@@ -0,0 +1,53 @@
...
+import { ApplicationShellLayoutMigration, WidgetDescription, ApplicationShellLayoutMigrationContext } from '@theia/core/lib/browser/shell/shell-layout-restorer';
+import { SearchInWorkspaceWidget } from './search-in-workspace-widget';
+import { SEARCH_VIEW_CONTAINER_ID, SEARCH_VIEW_CONTAINER_TITLE_OPTIONS } from './search-in-workspace-factory';
+
+@injectable()
+export class SearchLayoutVersion3Migration implements ApplicationShellLayoutMigration {
+    readonly layoutVersion = 6.0;
+    onWillInflateWidget(desc: WidgetDescription, { parent }: ApplicationShellLayoutMigrationContext): WidgetDescription | undefined {
+        if (desc.constructionOptions.factoryId === SearchInWorkspaceWidget.ID && !parent) {
+            return {
```
msujew, line 27: There seems to be an issue related to the new codicon icon for the search widget:

image

Simply add the following in between the `if` and `return`:

```ts
if (typeof desc.innerWidgetState === 'string') {
    desc.innerWidgetState = desc.innerWidgetState.replace(/search-in-workspace-tab-icon/g, codicon('search'));
}
```

EstherPerelman: Hi @msujew, why not to change [this line](https://github.com/EstherPerelman/theia/blob/94f37b3812e961d076a4369fab0b07a968630cd1/packages/search-in-workspace/src/browser/search-in-workspace-factory.ts#L29)
to ` iconClass: codicon('search')`

msujew: Right, that probably makes more sense 👍

colin-grant-work:
I'm seeing one odd behavior. It looks like the git decorations for modified files remain attached to the Source Control View Container even if you move the source control widget out of that container:

Source control view container with only NPM scripts in it, but decorated with modified file count:  
image

Source control widget in the Search view container, showing two modified files:  
image

The problem here seems to be in these lines:
https://github.com/eclipse-theia/theia/blob/10d5275207879334d0052a2e8a4d85c82fe30245/packages/scm/src/browser/decorations/scm-tab-bar-decorator.ts#L51-L58

which explicitly target the SCM_VIEW_CONTAINER_ID. In VSCode, the decorations follow the SCM widget, so you'd need some code that does basically this: 

https://github.com/eclipse-theia/theia/blob/f776392b6b616af48d222fbeb035e08c2c1f2865/packages/scm/src/browser/scm-contribution.ts#L153-L161

Another minor oddity:
1. Move another view into the Search container
2. Collapse the Search-in-Workspace widget
3. Use `ctrl+shift+f` to activate the Search-in-Workspace widget.
4. Observe that its title bar is highlighted, but the widget remains collapsed.

I think that calls for some logic in the `ViewContainerPart` that expands the part if `this.wrapped` is activated. Though strictly speaking it isn't related to this PR and could be deferred.

colin-grant-work:
I'm unable to reproduce the overlapping widget behavior that was previously observed, so I have no objections to merging this PR, if the search icon issue is addressed. The issues I've observed could be addressed in followups, if desired.

EstherPerelman:
> I'm unable to reproduce the overlapping widget behavior that was previously observed, so I have no objections to merging this PR, if the search icon issue is addressed. The issues I've observed could be addressed in followups, if desired.

Great, I've addressed the search icon issue, 
About the other issues:  
1. fixed:
> * Use `ctrl+shift+f` to activate the Search-in-Workspace widget.

2. it seems to be some more work than just change the `if` condition, 
First - when having for example both `git` and `open editors` parts on the same container - need to calculate the summary of them but for now it takes only one value (either the dirty open editors or the dirty git files) , 
Second - the `DebugWidget` container not supports decoration, 
Third - need to find a way to fire `fireDidChangeDecorations` or something else to update decorators on part moved.

EstherPerelman:
@colin-grant-work I have addressed the issue with `search` icon and the next issues:
> I'm seeing one odd behavior. It looks like the git decorations for modified files remain attached to the Source Control View Container even if you move the source control widget out of that container:

I have pushed a commit that supports the decorations moves, But It doesn't work for the widgets that move into the `debug` container.

> Another minor oddity:
> 
> 1. Move another view into the Search container
> 2. Collapse the Search-in-Workspace widget
> 3. Use `ctrl+shift+f` to activate the Search-in-Workspace widget.
> 4. Observe that its title bar is highlighted, but the widget remains collapsed.

I have added a code that opens the part on activating the widget by `ctrl+shift+g` / `ctrl+shift+g` / etc...

colin-grant-work:
Both of the new behaviors are working for me 👍. @msujew, @vince-fugnitto, would you mind confirming that the bad behaviors have been addressed and there are no new regressions?

One thing I realized recently is that the highlight behavior in VSCode (adding a white bar above or below the icon) isn't actually connected to the dragging between containers, but to general drag-drop repositioning behavior, since in VSCode you can drop a `ViewContainerPart` on the main tabbar, and it will create a new `ViewContainer`. In Phosphor, the general drag-drop is handled by shifting the icons around to open space for a repositioning. In this context, that's an undesirable behavior, because we want to drop _in_ a container, not between container icons. We may need to rethink one or the other of these drag behaviors if we implement VSCode's full set of interactive possibilities.

packages/core/src/browser/shell/tab-bars.ts
```diff
@@ -934,4 +955,78 @@ export class SideTabBar extends ScrollableTabBar {
         }
     }
 
+    toCancelViewContainerDND = new DisposableCollection();
+    protected cancelViewContainerDND = () => {
+        this.toCancelViewContainerDND.dispose();
+    };
+
+    /**
+     * handles `viewContainerPart` drag enter.
```
vince-fugnitto, line 964:
```suggestion
     * Handles `viewContainerPart` drag enter.
```

packages/core/src/browser/shell/tab-bar-decorator.ts
```diff
@@ -57,7 +57,7 @@ export class TabBarDecoratorService implements FrontendApplicationContribution {
         this.contributions.getContributions().map(decorator => decorator.onDidChangeDecorations(this.fireDidChangeDecorations));
     }
 
-    protected fireDidChangeDecorations = debounce(() => this.onDidChangeDecorationsEmitter.fire(undefined), 150);
+    fireDidChangeDecorations = debounce(() => this.onDidChangeDecorationsEmitter.fire(undefined), 150);
```
vince-fugnitto, line 60: I am wondering if we should expose the internals of firing the event from outside callers.

EstherPerelman: So how do you suggest to inform the decorations on parts movements? (Currently do it: `this.onDidChangeTrackableWidgets(() => this.decoratorService.fireDidChangeDecorations())`)
@vince-fugnitto

packages/search-in-workspace/src/browser/search-in-workspace-factory.ts
```diff
@@ -0,0 +1,60 @@
...
+        initiallyCollapsed: false
+    };
+
+    @inject(ViewContainer.Factory)
+    protected readonly viewContainerFactory: ViewContainer.Factory;
+    @inject(WidgetManager) protected readonly widgetManager: WidgetManager;
+
+    async createWidget(): Promise<ViewContainer> {
+        const viewContainer = this.viewContainerFactory({
+            id: SEARCH_VIEW_CONTAINER_ID,
```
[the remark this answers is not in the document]
EstherPerelman: @vince-fugnitto I fixed it, It just the same as done in [NavigatorWidgetFactory](https://github.com/EstherPerelman/theia/blob/5426df722169c145a3407409f63ff0ea2d44c204/packages/navigator/src/browser/navigator-widget-factory.ts#L38-L40)...

packages/search-in-workspace/src/browser/search-in-workspace-factory.ts
```diff
@@ -0,0 +1,60 @@
...
+export const SEARCH_VIEW_CONTAINER_TITLE_OPTIONS: ViewContainerTitleOptions = {
+    label: 'Search',
+    iconClass: codicon('search'),
+    closeable: true
+};
+
+@injectable()
+export class SearchInWorkspaceFactory implements WidgetFactory {
+
+    static ID = SEARCH_VIEW_CONTAINER_ID;
```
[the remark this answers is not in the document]
EstherPerelman: @vince-fugnitto I fixed it, It just the same as done in [NavigatorWidgetFactory](https://github.com/EstherPerelman/theia/blob/5426df722169c145a3407409f63ff0ea2d44c204/packages/navigator/src/browser/navigator-widget-factory.ts#L38-L40)...

EstherPerelman:
> One thing I realized recently is that the highlight behavior in VSCode (adding a white bar above or below the icon) isn't actually connected to the dragging between containers, but to general drag-drop repositioning behavior, since in VSCode you can drop a `ViewContainerPart` on the main tabbar, and it will create a new `ViewContainer`. In Phosphor, the repositioning drag-drop is handled by shifting the icons around to open space for a repositioning. In this context, that's an undesirable behavior, because we want to drop _in_ a container, not between container icons. We may need to rethink one or the other of these drag behaviors if we implement VSCode's full set of interactive possibilities.

@vince-fugnitto absolutely right! I still think these bars gives the user more indication on the dragging but I can remove it if you want to...

colin-grant-work:
> @vince-fugnitto absolutely right! I still think these bars gives the user more indication on the dragging but I can remove it if you want to...

No, I think the bars are the more future-proof version of the drag-drop marker. Eventually, we will want to be able to drop both inside of and next to, and only the bars - not the shifting behavior - will allow that. Just something for the person who implements the next round to think about 🙂.

packages/core/src/browser/shell/tab-bar-decorator.ts
colin-grant-work: The `onDidChangeTrackableWidgets` event is currently received in the `ApplicationShell`. One option would be to fire an event there (or see if the current logic allows for listeners), and then `@inject` the `ApplicationShell` into the decorator service or the decorators that need to respond to that event. It's a bit more roundabout, but it encapsulates the events better.

EstherPerelman: @colin-grant-work thank for the idea but It doesn't work for me because it causes circular dependency, I tried also to inject `shell` into `statusBarImpl` ctor but it throws errors too... Do you have something in mind to solve it?
This is the circular error:
image

colin-grant-work: > @colin-grant-work thank you for the idea but It doesn't work for me because it causes circular dependency, I tried also to inject `shell` into `statusBarImpl` ctor but it throws errors too... Do you have something in mind to solve it?
 
I have a solution, but it isn't a clean one. We've encountered a number of errors like this downstream when we've modified code around the shell. Basically something the shell creates in its constructor also injects the shell, so you end up  with `inversify` chasing its tail trying to construct the shell. One solution is to create something like this:
```typescript
export delayedShellInjectionFactory = Symbol('delayedShellInjectionFactory');
export interface delayedShellInjectionFactory {() => ApplicationShell}
bind(delayedShellInjectionFactory).toFactory(({container}) => container.get(ApplicationShell));
```
And then instead of this:
```typescript
@inject(ApplicationShell) protected readonly applicationShell: ApplicationShell;
```
you write this:
```typescript
@injectable()
export MyOtherClass {
protected applicationShell: ApplicationShell;
@inject(delayedShellInjectionFactory) protected readonly shellInjector: delayedShellInjectionFactory;

@postConstruct()
protected init(): void {
    this.applicationShell = this.shellInjector();
    // ... use the shell now
}
}
```

That way, `inversify` doesn't need the `ApplicationShell` to have already been constructed when `MyOtherClass` (on which it turns out `ApplicationShell`  depends) is constructed. That breaks the construction-time circular dependency, and you shouldn't see the stack overflow error.

But let me look at the dependency tree and see if there's a better way to handle this case.

EstherPerelman: @colin-grant-work this solution did the work, I already pushed it because I missed this line you wrote:
> But let me look at the dependency tree and see if there's a better way to handle this case.

So, If you find a better way I'll update accordingly, Thank you.

packages/core/src/browser/shell/tab-bar-decorator.ts
```diff
@@ -53,8 +55,16 @@ export class TabBarDecoratorService implements FrontendApplicationContribution {
...
+    protected shell: ApplicationShell;
+    @inject(delayedShellInjectionFactory) protected readonly shellInjector: delayedShellInjectionFactory;
+
     initialize(): void {
         this.contributions.getContributions().map(decorator => decorator.onDidChangeDecorations(this.fireDidChangeDecorations));
+        // Avoid circular dependency caused by the direct injection of the ApplicationShell
+        setTimeout(() => {
+            this.shell = this.shellInjector();
+            this.shell.onDidChangeTrackableWidgets(() => this.fireDidChangeDecorations());
+        });
```
vince-fugnitto, line 67: This does not look good unfortunately.

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -59,14 +59,20 @@ export type ApplicationShellLayoutVersion =
...
+    /** added the ability to drag and drop view parts between view containers */
+    6.0;
 
 /**
  * When a version is increased, make sure to introduce a migration (ApplicationShellLayoutMigration) to this version.
  */
 export const applicationShellLayoutVersion: ApplicationShellLayoutVersion = 5.0;
 
 export const ApplicationShellOptions = Symbol('ApplicationShellOptions');
+export const delayedShellInjectionFactory = Symbol('delayedShellInjectionFactory');
```
vince-fugnitto, line 72:
```suggestion
export const DelayedShellInjectionFactory = Symbol('DelayedShellInjectionFactory');
```

vince-fugnitto (changes requested):

packages/core/src/browser/shell/tab-bar-decorator.ts
paul-marechal: I didn't try but from the looks of it it should work without the `setTimeout`?

colin-grant-work: It's worth a try, but I believe that `inversify` runs `@postConstruct()`-decorated methods before it returns the instance from the 'constructor', so that it will still fail without the `setTimeout` explicitly rescheduling the 'injection' of the `ApplicationShell` until after the `TabBarDecoratorService` is returned from its `inversify` instantiation routine.

EstherPerelman: @colin-grant-work I reverted the commit, Is the PR ready to merge now?

msujew (commented):
It seems to work well in Chrome now, but there's some strange behavior in Firefox (92.0):

https://user-images.githubusercontent.com/4377073/134135084-b8a8c7b1-a160-410e-96fe-e52d5574ea84.mp4

As soon as I let the mouse button go, it starts dragging correctly. But before letting go, the dragged container stays at the same place.

EstherPerelman:
>It seems to work well in Chrome now, but there's some strange behavior in Firefox (92.0):

@msujew thanks for finding that, I called the event `preventDefault` which seems to solve it. (sorry for the late answer, from now on I am available) 
Can you merge it for tomorrow release? 
cc @colin-grant-work @vince-fugnitto 

By the way: I took an holiday for 2 weeks

msujew (approved):
Alright, looks good to me 👍

vince-fugnitto:
> **Can you merge it for tomorrow release?** cc @colin-grant-work @vince-fugnitto

@EstherPerelman last release we had a couple of major regressions after merging big pull-requests late in the month (resulted in two patch releases). Should we instead wait till right after the release so we have the month to discover any possible regressions? In any case I'll still have to take the time to re-review the code and functionality before approving.

EstherPerelman:
> > **Can you merge it for tomorrow release?** cc @colin-grant-work @vince-fugnitto
> 
> @EstherPerelman last release we had a couple of major regressions after merging big pull-requests late in the month (resulted in two patch releases). Should we instead wait till right after the release so we have the month to discover any possible regressions? In any case I'll still have to take the time to re-review the code and functionality before approving.

@vince-fugnitto you are right, I also think it's better to wait after the release in case of regressions, Thank you!

EstherPerelman:
@colin-grant-work / @vince-fugnitto, Please notify me before merging so I'll resolve conflicts..

EstherPerelman:
@vince-fugnitto why don't you merge it? I don't want to wait till the end of the month...

vince-fugnitto:
> @vince-fugnitto why don't you merge it? I don't want to wait till the end of the month...

@EstherPerelman I set time aside to be able to test the pr again first (I had never approved) https://github.com/eclipse-theia/theia/pull/9644#issuecomment-930130720.

> In any case I'll still have to take the time to re-review the code and functionality before approving.

Issue #9646, opened 2021-06-28
Supporting deep comparison of custom properties while distinguishing tasks.


magician-margatroid:
### Bug Description:
Hi there, in this issue a task with **reference value (array/object) as custom property in definition** is needed. 

While this task is running, we manually trigger a new execution of this task with the same type + scope + properties (same primitives but different references) + etc. In Theia, it will result in duplicated new executions of this task.

### Steps to Reproduce:

1. Initialize the Theia dev-env, clone [vscode-extension-samples](https://github.com/microsoft/vscode-extension-samples.git).
2. In the samples above, use [task-provider-sample](https://github.com/microsoft/vscode-extension-samples/tree/main/task-provider-sample), and modify [this line](https://github.com/microsoft/vscode-extension-samples/blob/main/task-provider-sample/src/customTaskProvider.ts#L132) into a longer time, e.g., 1000 * 60 for convenience.
3. Compile the task-provider-sample, package it if needed and make it loaded to Theia in the way you want.
4. Run or debug this plugin.
5. Run the task: Command Palette -> Tasks: run Task -> `custombuildscript: 32 watch incremental`
6. A pseudo terminal was spawn, this task runs successfully, and leave it untouched.
7. Re-run this task before former exec ends: Command Palette -> Tasks: run Task -> `custombuildscript: 32 watch incremental`
8. A new pseudo terminal with the same name was spawn again. This might not be expected.

### Analysis

The direct reason of the phenomenon above is, [in this line](https://github.com/eclipse-theia/theia/blob/master/packages/task/src/browser/task-definition-registry.ts#L122), the comparison between tasks with custom properties are shallow comparison. So if there's a reference type within them, they will never be the same even though they "want to be" the same, unless corresponding value share the same reference.

In VSCode, the 8th step will prompt user **this task is already running, to terminate it or restart it**.

In VSCode's implementation, one of the identity of a task is based on stringified enumerable own properties of task property, which ensures the equality of tasks "want to be equal", and is similar to deep comparison. 

E.g., by debugging, with the sample above, VSCode will yeild `"flags,0,watch,1,incremental,,flavor,32,type,custombuildscript,"`, which is a matter of string comparison.

I added several unit-test cases for "task-definition-registry.ts".

* By shallow comparison, it ends up with 33 passing and 2 failing.
* By deep comparison, it ends up with 35 passing.

### Additional Information

- Operating System: Ubuntu 20.04 LTS
- Node.js: v12.19.1 LTS
[...]

PR #9647, opened 2021-06-28
Enhancement: Supporting deep comparison of tasks with reference value in custom properties

magician-margatroid:
#### What it does

Fixes: #9646

[Issue link](https://github.com/eclipse-theia/theia/issues/9646): in short, tasks with same content but different reference value as custom properties are treated as different tasks, e.g., `[]`, `{prop: 'value'}`. This can lead to duplicate execution of same task.

This PR includes 2 commits:

1. Add several unit test cases for `task-definition-registry.ts` to ensure the logic of distinguishing.
2. Change shallow comparison into deep comparison of properties in `compareTasks` method.

#### How to test

- Unit test:
    - I've added some cases in the 1st commit.
    - Before the changes in 2nd commit, tests for `@theia/task` will ends up with 33 passing and 2 failing.
    - After the changes in 2nd commit, tests for `@theia/task` will ends up with 35 passing and 0 failing.

- Manually integration test:
    - Just follow the exact steps in the issue linked above.
    - Before changes, the task can be run for several time no need former execution ends.
    - After changes, "Task is already running" will be prompted for user if former execution doesn't end. (Which is also in consistency with VSCode.)

#### Review checklist

#### Reminder for reviewers


packages/task/src/browser/task-definition-registry.spec.ts
```diff
@@ -15,6 +15,7 @@
  ********************************************************************************/
 
 import { expect } from 'chai';
+import { foobarTaskFixture } from '../node/test/test-helper';
```
vince-fugnitto, line 18: @magician-margatroid `browser` (frontend) should not import anything from `node` (backend), even if it is simply for test purposes: [code organization](https://github.com/eclipse-theia/theia/wiki/Code-Organization). Given that the `foobarTaskFixture` is simply for test purposes, you can include the contents directly in the `task-definition-registry.spec.ts` file.

magician-margatroid: Thank you very much @vince-fugnitto . Excuse me and I'll improve it.

magician-margatroid:
Hi @vince-fugnitto , when you have the time, could you pls help to have a look at the build check result?

As it indicates, it seems the **Build / ubuntu-18.04, Node.js v12.x (pull_request)** failed after 14m — ubuntu-18.04, Node.js v12.x.

I found `root ERROR [nsfw-watcher: 9379] NSFW service error on "/home/runner/work/theia/theia/examples/browser": [Error: Service shutdown unexpectedly]` and several other errors in the log, but still not very sure about the root cause.

Personally I think this new commit doesn't have side effects, and also executed unit-tests locally with all-green.

magician-margatroid:
> Hi @vince-fugnitto , when you have the time, could you pls help to have a look at the build check result?
> 
> As it indicates, it seems the **Build / ubuntu-18.04, Node.js v12.x (pull_request)** failed after 14m — ubuntu-18.04, Node.js v12.x.
> 
> I found `root ERROR [nsfw-watcher: 9379] NSFW service error on "/home/runner/work/theia/theia/examples/browser": [Error: Service shutdown unexpectedly]` and several other errors in the log, but still not very sure about the root cause.
> 
> Personally I think this new commit doesn't have side effects, and also executed unit-tests locally with all-green.

Thanks. This time all that succeeded.

packages/task/src/browser/task-definition-registry.spec.ts
```diff
@@ -54,6 +55,52 @@ describe('TaskDefinitionRegistry', () => {
...
+    };
+    const defaultPresentation = {
+        clear: false,
+        echo: true,
+        focus: false,
+        panel: PanelKind.Shared,
+        reveal: RevealKind.Always,
+        showReuseMessage: true,
+    };
+    const foobarTaskFixture = {
```
vince-fugnitto, line 70: minor: what does a `fixture` represent, it is the first time I heard of this terminology for `tasks`.

packages/task/src/browser/task-definition-registry.spec.ts
```diff
@@ -105,4 +152,54 @@ describe('TaskDefinitionRegistry', () => {
...
+    describe('compareTasks function', () => {
+
+        beforeEach(() => registry.register(foobarTaskFixture.def));
+
+        it('should return false if given 2 task configurations with different type', () => {
+            const res = registry.compareTasks(
+                foobarTaskFixture.conf('id_1', 'type_1'),
+                foobarTaskFixture.conf('id_2', 'type_2'),
+            );
+            expect(res).to.be.not.ok;
```
vince-fugnitto, line 165: minor: rather than `okayness`, we can simply do truthy checks.

```ts
expect(res).to.be.true;
```

or


```ts
expect(res).equal(true);
```

packages/task/src/browser/task-definition-registry.spec.ts
```diff
@@ -105,4 +152,54 @@ describe('TaskDefinitionRegistry', () => {
...
+        });
+
+        it('should return true if given 2 same task configurations with empty arrays (different by reference) as custom property', () => {
+            const res = registry.compareTasks(
```
vince-fugnitto, line 169: I see the use of `res` a lot in the tests, what does it represnt? Perhaps a clearer name should be used?

packages/task/src/browser/task-definition-registry.spec.ts
magician-margatroid: Nice point. Well, frankly I didn't aim to use it exclusively for `tasks`, but I mean to use "test fixtures" here. In my understanding `fixture` can be something stable across runs of tasks. E.g., a file loaded with sample data or a script setting up stable env for task. 

Perhaps here I need to change it into mock or fake.

packages/task/src/browser/task-definition-registry.spec.ts
magician-margatroid: Good suggestion. I'll change it in this way.

packages/task/src/browser/task-definition-registry.spec.ts
magician-margatroid: Nice suggestion. I agree and I meant "result" here. I'll change them.

alvsan09 (commented):
I have tested this change and it works as expected :+1: 

The code looks good to me as well!
Thanks @magician-margatroid !

magician-margatroid:
> I have tested this change and it works as expected 
> 
> The code looks good to me as well!
> Thanks @magician-margatroid !

Ah, good to know it, @alvsan09 thx for your testing. It's my pleasure.

vince-fugnitto (approved):
The changes work well for me, I verified with the described steps in #9646 :+1: 
Thank you for the contribution, and looking forward to many more in the future!

RomanNikitenko:
I tried to test according to `steps to reproduce`. 
It's strange, but I don't see `custombuildscript: 32 watch incremental` task
custom_pr

For the `master` I do see it
custom_master

Also, the order is different.
Could someone confirm that the problem is on my side only...
thanks in advance!

p/s 
I installed the extension as vsix
I used multi-root workspace for testing
I'm on macos

RomanNikitenko:
I found what's the problem...
It looks like I selected `Never scan the task output` at running the task.
Then the task was converted to `configured` with another label.

https://user-images.githubusercontent.com/5676062/125966425-86a506e8-e7e4-4276-9f4d-09c11e683a1b.mp4

RomanNikitenko (approved):
Tested according to the `steps to reproduce` section of the issue, it works well for me.
I think it makes sense to squash the changes in 1 or 2 commits before merging.

vince-fugnitto:
I used the new `squash and merge` option to merge the commit :+1:

magician-margatroid:
Hi everyone, thanks a lot for all your work. It's my pleasure and appreciate your help and analysis all the way.

PR #9658, opened 2021-06-28
Modify next-node selection logic for PreferenceTreeModel

### LINKED ISSUE #9655: Arrows should select only visible preferences
### Bug Description:

When navigating the preference tree using the arrow keys, and when there is a search filter hiding some of the preferences, it appears that arrow keys also select the hidden preferences. They should select only visible preferences.

### Steps to Reproduce:

1. Open the Preferences widget.
2. Search for "selection", or other text of your choosing.
3. Click on "Text Editor".
image
4. Press the down arrow and notice that it takes 4x before selecting"Find", and then another 9x before selecting "Suggest".
5. Clear the search filter and expand the "Text Editor" category, and you see the 4x & 9x above suggest the arrows were selecting the hidden entries.
image

### Additional Information

- Operating System: RHEL 7, Firefox 89.0
- Theia Version: 1.14.0, browser example

colin-grant-work:
#### What it does

Fixes #9655 by changing the logic for choosing the 'next selectable node' in preferences tree. For extensibility, I've made that check a protected method in the base `TreeModel` class and overridden it in the `PreferenceTreeModel`

#### How to test

1. Open the Preferences widget.
2. Search for "selection", or other text of your choosing.
3. Click on "Text Editor".
4. Press the down arrow and notice that it takes 1x before selecting "Find", and then another 1x before selecting "Suggest".
5. Perform the same test on an unfiltered tree and see that each keyboard press moves you 1 item.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/tree/tree-model.ts
+++ b/packages/core/src/browser/tree/tree-model.ts
@@ -306,7 +306,7 @@ export class TreeModelImpl implements TreeModel, SelectionProvider<ReadonlyArray
         // Skip the first item. // TODO: clean this up, and skip the first item in a different way without loading everything.
         iterator.next();
         let result = iterator.next();
-        while (!result.done && !SelectableTreeNode.isVisible(result.value)) {
+        while (!result.done && !this.isVisibleSelectableNode(result.value)) {
             result = iterator.next();
         }
         const node = result.value;
@@ -314,6 +314,10 @@ export class TreeModelImpl implements TreeModel, SelectionProvider<ReadonlyArray
             return node;
         }
         return undefined;
+    }
+
+    protected isVisibleSelectableNode(node: TreeNode): node is SelectableTreeNode {
+        return SelectableTreeNode.isVisible(node);
     }
 
     protected createBackwardIterator(node: TreeNode | undefined): TreeIterator | undefined {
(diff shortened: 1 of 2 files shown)


packages/core/src/browser/tree/tree-model.ts
```diff
@@ -302,21 +302,22 @@ export class TreeModelImpl implements TreeModel, SelectionProvider<ReadonlyArray
         return iterator && this.doGetNextNode(iterator);
     }
 
-    protected doGetNextNode(iterator: TreeIterator): SelectableTreeNode | undefined {
+    protected doGetNextNode(iterator: IterableIterator<TreeNode>): SelectableTreeNode | undefined {
```
colin-grant-work, line 305: I'd rather just change this to be `extends IterableIterator<TreeNode>`, but that might be slightly more breaking. Anyone who's implemented an iterator that isn't iterable would probably be grateful in the long run, though. Who wants to type `.next()` all the time?

https://github.com/eclipse-theia/theia/blob/210fa4ac6875716f23ac70f94fe6f0ffc693bfe5/packages/core/src/browser/tree/tree-iterator.ts#L20

vince-fugnitto (approved):
I confirmed the presence of the bug on master and that the pull-request resolves the issue 👍

PR #9660, opened 2021-06-28
limit use of dynamic require

paul-marechal:
Using Node's builtin require function to dynamically import scripts
makes bundling Node apps difficult when using tools like Webpack.

Remove as many occurences of Node require and replace the few required
ones with `@theia/core/lib/node/dynamic-require#dynamic-require()`.

Add a new eslint rule `import/no-dynamic-require` to help detect places
that would benefit from `@theia/core/lib/node/dynamic-require`.

Closes https://github.com/eclipse-theia/theia/issues/9432

#### How to test

Pure refactoring. Systems that spawn child processes should work like before (file watching, plugins, git).

#### Review checklist

#### Reminder for reviewers


paul-marechal:
@eclipse-theia/core any opinions about that new `@theia/dynamic-require` package? Its purpose is to factorize all dynamic requires calls across the application, to make bundling easier.

If everyone's fine with it, we would like to get this PR merged and included in 1.15.0, if possible.

dev-packages/dynamic-require/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+<h2>ECLIPSE THEIA - DYNAMIC REQUIRE</h2>
+
+<hr />
+
+</div>
+
+## Description
+
+We need to factorize dynamic require calls for when bundling. This module only re-exports its own `require` function for that purpose.
```
vince-fugnitto, line 15:
```suggestion
We need to factorize dynamic require calls when bundling. This module only re-exports its own `require` function for that purpose.
```

vince-fugnitto (commented):
The changes look good to me at a high-level, I haven't tested them out thoroughly yet.

paul-marechal:
Note that I confirmed that this PR helps with bundling Theia Blueprint for the browser. See [branch](https://github.com/eclipse-theia/theia-blueprint/tree/mp/browser-2).

packages/plugin-ext/src/hosted/node/plugin-host-rpc.ts
```diff
@@ -197,7 +199,8 @@ export class PluginHostRPC {
                 for (const api of extApi) {
                     if (api.backendInitPath) {
                         try {
-                            const extApiInit = require(api.backendInitPath);
+                            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                            const extApiInit: any = dynamicRequire(api.backendInitPath);
```
msujew, line 203: Nitpick: we could use `{ provideApi: ExtPluginApiBackendInitializationFn }` or `dynamicRequire<{ provideApi: ExtPluginApiBackendInitializationFn }>` instead of `any` here.

packages/debug/src/node/vscode/vscode-debug-adapter-contribution.ts
```diff
@@ -98,13 +98,11 @@ export abstract class AbstractVSCodeDebugAdapterContribution implements DebugAda
     }
 
     protected async parse(): Promise<VSCodeExtensionPackage> {
-        let text = (await fs.readFile(this.pckPath)).toString();
+        let text: string = await fs.readFile(this.pckPath, 'utf8');
```
msujew, line 101: Nitpick: The `string` type is not necessary here, TS should infer that.

packages/git/src/node/git-locator/git-locator-client.ts
```diff
@@ -36,7 +36,7 @@ export class GitLocatorClient implements GitLocator {
         return new Promise((resolve, reject) => {
             const toStop = this.ipcConnectionProvider.listen({
                 serverName: 'git-locator',
-                entryPoint: paths.resolve(__dirname, 'git-locator-host')
+                entryPoint: paths.join(__dirname, 'git-locator-host'),
```
msujew, line 39: Nitpick: redundant `,` at EOL. Afaik we don't enforce a member delimiter style, so feel free to ignore this.

packages/plugin-ext-vscode/src/node/scanner-vscode.ts
```diff
@@ -79,7 +81,7 @@ export class VsCodePluginScanner extends TheiaPluginScanner implements PluginSca
             startMethod: 'activate',
             stopMethod: 'deactivate',
 
-            backendInitPath: __dirname + '/plugin-vscode-init.js'
+            backendInitPath: path.join(__dirname, 'plugin-vscode-init'),
```
msujew, line 84: Nitpick: redundant `,` at EOL. Afaik we don't enforce a member delimiter style, so feel free to ignore this.

msujew (approved):
LGTM, I ran a few manual tests for cases which used to use the default require:

* Debugging adapter ✔️
* Installing extensions ✔️
* Plugin host in general ✔️
* TextMate grammars ✔️ 
* Testrunner ✔️
* Electron ✔️

Added a few nitpicks though.

packages/debug/src/node/vscode/vscode-debug-adapter-contribution.ts
paul-marechal: It is to enforce `readFile` to return `string`. i.e. if I remove the `utf8` argument it returns `Buffer`...  I usually prefer to let TypeScript infer too, except when a function has overloads with incompatible return types like this. Tell me if you'd still prefer to just rely on inference.

msujew: Right, but as long as the `encoding` argument is there, `readFile` will always be enforced to return `string`, right? At least that's what the [typescript definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1fc1033f3a50e169b088729cbd691f60108c0262/types/fs-extra/index.d.ts#L217) says:

```ts
export function readFile(file: PathLike | number, encoding: string): Promise<string>;
```

So I don't see a reason to keep the `string` type for `text` there.

paul-marechal:
Addressed Mark's comments and moved `@theia/dynamic-require` from `dev-packages` to `package` since this will be a runtime dependency.

tsconfig.json
```diff
@@ -183,6 +183,9 @@
       ],
       "@theia/ovsx-client/lib/*": [
         "dev-packages/ovsx-client/src/*"
+      ],
+      "@theia/dynamic-require/*": [
+        "dev-packages/dynamic-require/*"
```
msujew, line 188: I think this path isn't correct. You moved it to a normal package.

packages/dynamic-require/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@theia/dynamic-require",
+  "version": "1.14.0",
```
msujew, line 3: The package states `1.14.0` but the `core` package requires `1.16.0`. Is this the source of the build error?

paul-marechal:
From the start I didn't like the idea of adding a new package which was an implementation detail (and publishing it alone).

I just remove that package and exported the dynamic require function into `@theia/core/src/node/dynamic-require`.

packages/plugin-ext/src/hosted/node/plugin-host-rpc.ts
```diff
@@ -14,13 +14,14 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
+import { dynamicRequire } from '@theia/core/lib/node/dynamic-require';
```
paul-marechal, line 17: @eclipse-theia/plugin-system does that `@theia/core` import cause any issue with isolating the host process or anything else I would be missing?

tsmaeder: I don't think so: as long as the module is available on the back end, we should be fine.

tsconfig.json
```diff
@@ -183,6 +183,9 @@
       ],
       "@theia/ovsx-client/lib/*": [
         "dev-packages/ovsx-client/src/*"
+      ],
+      "@theia/dynamic-require/*": [
```
tsmaeder, line 187: I think you merged this package into core, right?

paul-marechal: Good catch, thought the reference generator would take care of these, but apparently not.

vince-fugnitto (approved):

I did not notice any regressions:
- builds and starts successfully
- plugins are correctly loaded and started
- executing debug sessions works correctly
- executing tasks works correctly

vince-fugnitto:
@paul-marechal closes https://github.com/eclipse-theia/theia/issues/9432?

PR #9671, opened 2021-06-30
Fix #9670: Custom editor plugin activation called 3 times when double click quickly

### LINKED ISSUE #9670: Custom editor plugin activation called 3 times when double click quickly
### Bug Description:
When pressed custom editor file several times (quickly) the plugin activation was called 3 timed

### Steps to Reproduce:

1. Unzip & upload next extension:
[custom-editor-sample-with-output.zip](https://github.com/eclipse-theia/theia/files/6739706/custom-editor-sample-with-output.zip)

2. Unzip & open the example.catscratch file in Theia: 
[example file.zip](https://github.com/eclipse-theia/theia/files/6739737/example.file.zip)

3. Click first time to open `example.catscratch` and then close
4. Press several times quickly to open `example.catscratch` (I used touchpad to press quickly but it also reproduced with double click the file on electron)
5. See the output is showing 3 timed in a row (each activation adding a row) like here:

https://user-images.githubusercontent.com/71729183/123939248-9d2e8b00-d9a0-11eb-8eae-1309b6730ebf.mp4

### Additional Information

- Theia Version: master

EstherPerelman:

#### What it does

Fixes #9670

1. by disable creating new `customEditorWidget` for a specific url if we already pending to create a widget for this url (added `pendingGetWidgets`)
2. by moving the `onDidOpenCustomEditorEmitter` fire event inside the `if(!widget)` statement to prevent firing in case widget created but still pending open...

These 2 changes are required to solve this issue.

#### How to test
Follow #9670  Steps to Reproduce, See the output not showing three times

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/custom-editors/custom-editor-opener.tsx
+++ b/packages/plugin-ext/src/main/browser/custom-editors/custom-editor-opener.tsx
@@ -57,6 +57,7 @@ export class CustomEditorOpener implements OpenHandler {
         }
     }
 
+    protected readonly pendingWidgetPromises = new Map<string, Promise<CustomEditorWidget>>();
     async open(uri: URI, options?: OpenerOptions): Promise<Widget | undefined> {
         let widget: CustomEditorWidget | undefined;
         const widgets = this.widgetManager.getWidgets(CustomEditorWidget.FACTORY_ID) as CustomEditorWidget[];
@@ -69,13 +70,20 @@ export class CustomEditorOpener implements OpenHandler {
             return this.shell.activateWidget(widget.id);
         }
         if (!widget) {
-            const id = v4();
-            widget = await this.widgetManager.getOrCreateWidget<CustomEditorWidget>(CustomEditorWidget.FACTORY_ID, { id });
-            widget.viewType = this.editor.viewType;
-            widget.resource = uri;
+            const uriString = uri.toString();
+            let widgetPromise = this.pendingWidgetPromises.get(uriString);
+            if (!widgetPromise) {
+                const id = v4();
+                widgetPromise = this.widgetManager.getOrCreateWidget<CustomEditorWidget>(CustomEditorWidget.FACTORY_ID, { id });
+                this.pendingWidgetPromises.set(uriString, widgetPromise);
+                widget = await widgetPromise;
+                this.pendingWidgetPromises.delete(uriString);
+                widget.viewType = this.editor.viewType;
+                widget.resource = uri;
+                this.onDidOpenCustomEditorEmitter.fire(widget);
+            }
         }
-
-        this.onDidOpenCustomEditorEmitter.fire(widget);
+        return widget;
     }
 
     matches(selectors: CustomEditorSelector[], resource: URI): boolean {


packages/plugin-ext/src/main/browser/custom-editors/custom-editor-opener.tsx
```diff
@@ -68,14 +69,21 @@ export class CustomEditorOpener implements OpenHandler {
...
         }
+        const stringUri = uri.toString();
+        if (!widget && this.pendingGetWidgets.has(stringUri)) {
+            return;
+        }
         if (!widget) {
             const id = v4();
+            this.pendingGetWidgets.set(stringUri, Disposable.create(() => {
+                this.pendingGetWidgets.delete(stringUri);
+            }));
```
msujew, line 80: I'd argue creating a `Disposable` in here is a bit over-engineered. Would a `() => void` suffice in this place?

packages/plugin-ext/src/main/browser/custom-editors/custom-editor-opener.tsx
```diff
@@ -68,14 +69,19 @@ export class CustomEditorOpener implements OpenHandler {
         if (widget?.isAttached) {
             return this.shell.activateWidget(widget.id);
         }
+        const stringUri = uri.toString();
+        if (!widget && this.pendingGetWidgets.has(stringUri)) {
+            return;
```
tsmaeder, line 74: This resolves to "undefined" if a widget is already being opened. Wouldn't it be better if we returned a promise that resolves to the widget that is already being opened? I mean something like this:

```suggestion
        if (!widget) {
            const uriString = uri.toString();
            let creation = this.pendingCreations.get(uriString);
            if (!creation) {
                const id = v4();
                creation = this.widgetManager.getOrCreateWidget<CustomEditorWidget>(CustomEditorWidget.FACTORY_ID, { id }).then(w => {
                    this.pendingCreations.delete(uriString);
                    w.viewType = this.editor.viewType;
                    w.resource = uri;
                    this.onDidOpenCustomEditorEmitter.fire(w);
                    return w;
                });
                this.pendingCreations.set(uriString, creation!);
            }
            return creation;
        }
```

EstherPerelman: Much better, Thanks! 
Pushed the fix..

tsmaeder: I didn't test it though 🦆

EstherPerelman: My fault, I didn't `yarn`... now it's working

EstherPerelman:
@msujew @tsmaeder Can you approve?

msujew (approved):
I can confirm that the issue can be reproduced on the current `master` and this change addresses the issue nicely.

The activation event is only fired once every time the editor is opened.

EstherPerelman:
> I can confirm that the issue can be reproduced on the current `master` and this change addresses the issue nicely.
> 
> The activation event is only fired once every time the editor is opened.

Thank you @msujew Who can merge it?
cc @tsmaeder

tsmaeder:
There is a build failure on Windows, but there is simply no imaginable way the failure could be caused by this PR.

tsmaeder (approved):

EstherPerelman:
@tsmaeder Thank you, Are you able to merge?

EstherPerelman:
@tsmaeder Thanks 😊

PR #8334, opened 2020-08-10
fix signatures of vscode.open and vscode.diff

akosyakov:
#### What it does

- fix #8333: fix signatures of vscode.open and vscode.diff. Internal `TextDocumentShowOptions` was used instead of `vscode.TextDocumentShowOptions` leading to ignoring the selection all the time.

#### How to test

- Call `vscode.open` or `vscode.diff` and pass selection. The selection should be properly revealed.

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #9681, opened 2021-06-30
Implement missing functionality for call hierarchy API

### LINKED ISSUE #7065: showTextDocument is not compatible to vscode and assumes non zero-based range
### Description
showTextDocument is not compatible to vscode and assumes non zero-based range

### Reproduction Steps
See: https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/main/browser/documents-main.ts#L225
it converts to zero-based although TextDocumentShowOptions has selection with type Range with zero-based positions.
it is not compatible with vscode.

See the options received is already zero-based and should be kept like that.
image

**OS and Theia version:**

**Diagnostics:**

### LINKED ISSUE #8333: vscode.open does not accept vscode.Selection as a selection option
VS Code allow to pass any subclass of Range to vscode.open. In Theia subclasses gets corrupted leading to bogus open options.

### LINKED ISSUE #9612: Calls: Show Call Hierarchy command doesn't work
### Bug Description:
I get an error `Cannot use 'in' operator to search for 'uri' in undefined` at using `Calls: Show Call Hierarchy` command.
Also the corresponding panel is empty - it doesn't display the result of command execution.   

### Steps to Reproduce:

Try to use `F1`=> `Calls: Show Call Hierarchy`  command.
Please see videos.

### Additional Information

- Operating System: Fedora
- Theia Version: 55b55eec6af8b50ae61fc21bc0cd46a6c52d3272

call_hierarchy

Actual behavior:

https://user-images.githubusercontent.com/5676062/122606341-78254880-d081-11eb-8409-702cae4a3309.mp4

Expected behavior:

https://user-images.githubusercontent.com/5676062/122606609-e8cc6500-d081-11eb-8085-a2f9ddc7a80f.mp4

colin-grant-work:
#### What it does

Fixes #9612 by implementing the commands used by the vscode-references-view plugin to retrieve call hierarchy information.
It also:
 - Deprecates the `CurrentEditorAccess` class in the `call-hierarchy` package in favor of the one implemented in the `editor` package.
 - Fixes a bunch of off-by-one translation errors between RPC Model Ranges and VSCode Ranges. That is an absolute mess that deserves separate treatment. The translation is implemented separately at least 5 times.
    - Fixes #7065
    - Fixes #8333
 - Adds a check to Timeline module to ensure that a selection exists to avoid an error log that appeared in the original issue report but wasn't actually the cause of the failure of the call hierarchy code.

#### How to test

1. You should be able to use the `Calls: Show Call Hierarchy Command`
2. You should be able to get both incoming (default) and outgoing calls.
3. When you click on a node, it should reveal the location of that function. Confirm that your cursor is at the beginning of the function.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/command-registry-main.ts
+++ b/packages/plugin-ext/src/main/browser/command-registry-main.ts
@@ -77,7 +77,7 @@ export class CommandRegistryMainImpl implements CommandRegistryMain, Disposable 
             throw new Error(`Command with id '${id}' is not registered.`);
         }
         try {
-            return await this.delegate.executeCommand(id, ...args);
+            return await this.delegate.executeCommand<T>(id, ...args);
         } catch (e) {
             // Command handler may be not active at the moment so the error must be caught. See https://github.com/eclipse-theia/theia/pull/6687#discussion_r354810079
             if ('code' in e && e['code'] === 'NO_ACTIVE_HANDLER') {
--- a/packages/plugin-ext/src/plugin/documents.ts
+++ b/packages/plugin-ext/src/plugin/documents.ts
@@ -255,10 +255,10 @@ export class DocumentsExtImpl implements DocumentsExt {
             if (options.selection) {
                 const { start, end } = options.selection;
                 selection = {
-                    startLineNumber: start.line,
-                    startColumn: start.character,
-                    endLineNumber: end.line,
-                    endColumn: end.character
+                    startLineNumber: start.line + 1,
+                    startColumn: start.character + 1,
+                    endLineNumber: end.line + 1,
+                    endColumn: end.character + 1
                 };
             }
             documentOptions = {
--- a/packages/plugin-ext/src/plugin/known-commands.ts
+++ b/packages/plugin-ext/src/plugin/known-commands.ts
@@ -17,13 +17,13 @@
 import { Range as R, Position as P, Location as L } from '@theia/core/shared/vscode-languageserver-types';
 import * as theia from '@theia/plugin';
 import { cloneAndChange } from '../common/objects';
-import { Position, Range, Location, CallHierarchyItem, URI } from './types-impl';
+import { Position, Range, Location, CallHierarchyItem, URI, TextDocumentShowOptions } from './types-impl';
 import {
     fromPosition, fromRange, fromLocation,
     isModelLocation, toLocation,
     isModelCallHierarchyItem, fromCallHierarchyItem, toCallHierarchyItem,
     isModelCallHierarchyIncomingCall, toCallHierarchyIncomingCall,
-    isModelCallHierarchyOutgoingCall, toCallHierarchyOutgoingCall
+    isModelCallHierarchyOutgoingCall, toCallHierarchyOutgoingCall, fromTextDocumentShowOptions
 } from './type-converters';
 
 // Here is a mapping of VSCode commands to monaco commands with their conversions
@@ -294,6 +294,8 @@ export namespace KnownCommands {
     mappings['vscode.prepareCallHierarchy'] = ['vscode.prepareCallHierarchy', CONVERT_VSCODE_TO_MONACO, CONVERT_MONACO_TO_VSCODE];
     mappings['vscode.provideIncomingCalls'] = ['vscode.provideIncomingCalls', CONVERT_VSCODE_TO_MONACO, CONVERT_MONACO_TO_VSCODE];
     mappings['vscode.provideOutgoingCalls'] = ['vscode.provideOutgoingCalls', CONVERT_VSCODE_TO_MONACO, CONVERT_MONACO_TO_VSCODE];
+    mappings['vscode.open'] = ['vscode.open', CONVERT_VSCODE_TO_MONACO];
+    mappings['vscode.diff'] = ['vscode.diff', CONVERT_VSCODE_TO_MONACO];
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     export function map<T>(id: string, args: any[] | undefined, toDo: (mappedId: string, mappedArgs: any[] | undefined, mappedResult: ConversionFunction | undefined) => T): T {
@@ -346,6 +348,12 @@ export namespace KnownCommands {
     function vscodeToMonacoArgsConverter(args: any[]) {
         // tslint:disable-next-line:typedef
         return cloneAndChange(args, function (value) {
+            if (CallHierarchyItem.isCallHierarchyItem(value)) {
+                return fromCallHierarchyItem(value);
+            }
+            if (TextDocumentShowOptions.isTextDocumentShowOptions(value)) {
+                return fromTextDocumentShowOptions(value);
+            }
             if (Position.isPosition(value)) {
                 return fromPosition(value);
             }
@@ -354,9 +362,6 @@ export namespace KnownCommands {
             }
             if (Location.isLocation(value)) {
                 return fromLocation(value);
-            }
-            if (CallHierarchyItem.isCallHierarchyItem(value)) {
-                return fromCallHierarchyItem(value);
             }
             if (!Array.isArray(value)) {
                 return value;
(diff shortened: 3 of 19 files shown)


packages/plugin-ext/src/plugin/known-commands.ts
```diff
@@ -294,6 +294,7 @@ export namespace KnownCommands {
     mappings['vscode.prepareCallHierarchy'] = ['vscode.prepareCallHierarchy', CONVERT_VSCODE_TO_MONACO, CONVERT_MONACO_TO_VSCODE];
     mappings['vscode.provideIncomingCalls'] = ['vscode.provideIncomingCalls', CONVERT_VSCODE_TO_MONACO, CONVERT_MONACO_TO_VSCODE];
     mappings['vscode.provideOutgoingCalls'] = ['vscode.provideOutgoingCalls', CONVERT_VSCODE_TO_MONACO, CONVERT_MONACO_TO_VSCODE];
+    mappings['vscode.open'] = ['vscode.open', CONVERT_VSCODE_TO_MONACO];
```
colin-grant-work, line 297: Strictly speaking, this isn't necessary to get the version of the plugin downloaded by our scripts to work because it uses `vscode.window.showTextDocument` rather than this command. However, later versions of the plugin use the `vscode.open` command, and we have an off-by-one issue when handling `TextDocumentShowOptions` passed by plugins into that command.

RomanNikitenko:
@colin-grant-work 
I see that you picked up the latest changes from the `master` - just fyi - some tests will fail due to https://github.com/eclipse-theia/theia/issues/9692

colin-grant-work:
> ... some tests will fail due to #9692, so such failure is not related to your changes

Thanks for the heads up. Always nice to have someone else to blame 🙂

RomanNikitenko:
@colin-grant-work 
I'm going to test/review the changes tomorrow.
I'm sorry for the delay...

packages/plugin-ext/src/main/browser/command-registry-main.ts
```diff
@@ -77,7 +77,7 @@ export class CommandRegistryMainImpl implements CommandRegistryMain, Disposable
             throw new Error(`Command with id '${id}' is not registered.`);
         }
         try {
-            return await this.delegate.executeCommand(id, ...args);
+            return this.delegate.executeCommand<T>(id, ...args);
```
RomanNikitenko, line 80: does try/catch still work after these changes?
there is a comment that it was added for some use case...

colin-grant-work: You're right. I think I removed it reflexively, but `try...catch` is the context where `return await` actually has a function....

RomanNikitenko:
I tested a little the changes and can confirm that in general it works well for me!

I noticed that the source is not displayed for items:

VS Code:

vscode_source

Theia:

theia_source

Another difference which I found:
VS Code moves cursor to the corresponding method when I use double click, but Theia does nothing for me.

VS Code
double_click_vscode

Theia
double_click_theia

@colin-grant-work 
Does it work for you?

I'm not requesting a fix - I think we can improve it in a separate PR - just sharing what I detected within review the PR...

RomanNikitenko:
@colin-grant-work 
> Confirm that your cursor is at the beginning of the function.

What do you mean?
I don't see that my cursors moves to a function. Should it work at double click?

colin-grant-work:
> I noticed that the source is not displayed for items

Good find. I'm not sure what's going on here - I'll check how that data is specified in the tree items created by the plugin and how we're handling them.
 
> Another difference which I found:
> VS Code moves cursor to the corresponding method when I use double click, but Theia does nothing for me.
> 
> @colin-grant-work
> Does it work for you?

It does work... sort of. I think the difference here is that VSCode transfers focus to the editor on double click, so you can see your cursor, but Theia transfers focus back to the reference view, so the editor cursor disappears. If you click the editor tab to focus the editor, the cursor _should_ be in the same place in Theia as it is in VSCode. I recall seeing code with something like `preserveFocus: true`, so I'll check whether that's on our side or the plugin's, and whether it's an appropriate choice for all situations.

colin-grant-work:
@RomanNikitenko, I think that what you're seeing in VSCode may be a fairly recent addition or depend on other functionality. Running VSCode 1.50 in the same environment as I run Theia, here's what I see:

image

And when double clicking to open the reference, focus stays in the references widget (orange/brown is the focus color for the tree, and no cursor is present in the editor):

image

The version of the plugin that we're using is quite old (0.0.47 vs current 0.0.80), so I think we shouldn't expect to see the exact same behavior as in the latest VSCode.

packages/plugin-ext/src/main/browser/command-registry-main.ts
colin-grant-work: I've restored the `await` here.

packages/plugin-ext/src/plugin/documents.ts
```diff
@@ -255,10 +255,10 @@ export class DocumentsExtImpl implements DocumentsExt {
...
                 const { start, end } = options.selection;
                 selection = {
-                    startLineNumber: start.line,
-                    startColumn: start.character,
-                    endLineNumber: end.line,
-                    endColumn: end.character
+                    startLineNumber: start.line + 1,
+                    startColumn: start.character + 1,
+                    endLineNumber: end.line + 1,
+                    endColumn: end.character + 1
```
RomanNikitenko, line 261: It looks like you have to add these `+1` because `theia` has `-1` [here](https://github.com/eclipse-theia/theia/blob/6125352297da5c0431819ac172e69830670c0c78/packages/plugin-ext/src/main/browser/documents-main.ts#L242).

So, maybe it's true way to remove those `-1` instead of adding these `+1`.
Could you check it?

RomanNikitenko:
@colin-grant-work 
> @RomanNikitenko, I think that what you're seeing in VSCode may be a fairly recent addition or depend on other functionality.

thank you for the investigation and detailed explanation!

packages/plugin-ext/src/plugin/documents.ts
colin-grant-work: Both are, unfortunately, correct. The problem is we have two different types named `Range` and they have different indexing. 

1-based indexing for this type:
https://github.com/eclipse-theia/theia/blob/12f3389a6fcfed7143c1dc5aef5a2c507d7bc38f/packages/plugin-ext/src/common/plugin-api-rpc-model.ts#L53-L70

And 0-based indexing for this type: 
https://github.com/eclipse-theia/theia/blob/12f3389a6fcfed7143c1dc5aef5a2c507d7bc38f/packages/plugin/src/theia.d.ts#L343-L352
because this type has 0-based indexing:
https://github.com/eclipse-theia/theia/blob/12f3389a6fcfed7143c1dc5aef5a2c507d7bc38f/packages/plugin/src/theia.d.ts#L230-L243

It seems like we translate between those two types of `Range`s more often than necessary, but every time we do, we have to be sure to add or subtract one, depending on the direction.

As I say in my PR description:
> Fixes a bunch of off-by-one translation errors between RPC Model Ranges and VSCode Ranges. **_That is an absolute mess_** that deserves separate treatment. The translation is implemented separately at least 5 times.

There are functions specialized in the translation between the two types, but they aren't used everywhere, and it didn't seem like this PR was the right place to make that refactoring everywhere where the translation is currently implemented.

colin-grant-work: @RomanNikitenko, I've added an issue for this here: https://github.com/eclipse-theia/theia/issues/9739 .

RomanNikitenko: > The problem is we have two different types named Range and they have different indexing.

I see...
thanks!

RomanNikitenko:
@colin-grant-work 
btw: your PR fixes https://github.com/eclipse-theia/theia/issues/8333
at least `vscode.open` command

RomanNikitenko (approved):

colin-grant-work:
> @colin-grant-work
> btw: your PR fixes #8333
> at least `vscode.open` command

Thanks for pointing that out. I'll check the diff command, and if it's as straightforward to fix as `vscode.open`, I'll add it to this PR.

colin-grant-work:
> Thanks for pointing that out. I'll check the diff command, and if it's as straightforward to fix as `vscode.open`, I'll add it to this PR.

Adding `vscode.diff` to the list of known commands resulted in the same behavior as #8334. I've created an issue (#9746) for the odd behavior of the diff widget on opening, which I believe to be a separate problem.

colin-grant-work:
I'll merge this tomorrow if there are no objections, to leave some time before the release.

PR #9684, opened 2021-06-30
Support multiselection in 'Add Folder to Workspace...' dialog

### LINKED ISSUE #9665: `Add folder to workspace...` dialog should support multiselection
The dialog that opens when you select the 'Add folder to workspace...' menu item should support multi-selection to facilitate adding multiple folders to a workspace simultaneously. The corresponding VSCode dialog does support multiple selection.

### Feature Description:

The command is registered here:
https://github.com/eclipse-theia/theia/blob/0ebdec5311cf01450de43acd08bfecfe72a1df7c/packages/workspace/src/browser/workspace-commands.ts#L306-L332

I tried a naive implementation just setting `canSelectMany` to true and setting up a loop over the selected URI's, but that didn't work immediately, so some additional modification of the workspace adding code is likely necessary.

colin-grant-work:

#### What it does

Fixes #9665 by allowing multiple selection in the workspace dialog and ensuring that the public methods 'addRoot' and 'removeRoots' (and new method `addRoots`) don't return before the `._roots` field of the `WorkspaceService` has been updated. Previously, changes to the workspace would emit an event that would trigger an `updateWorkspace` cycle, but since that method is also asynchronous, it was guaranteed not to have finished its work in the tick after `addRoot` and `removeRoots` resolved, meaning that even `await`ing those methods didn't guarantee that the `WorkspaceService` was ready for another call.

#### How to test

1. Open a workspace.
2. Use the 'Add folder to workspace' command (available in the `...` context menu in the explorer).
3. Select several folders.
4. Observe that your workspace updates correctly to include all of the folders added.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/workspace/src/browser/workspace-commands.ts
+++ b/packages/workspace/src/browser/workspace-commands.ts
@@ -307,16 +307,18 @@ export class WorkspaceCommandContribution implements CommandContribution {
                 isEnabled: () => this.workspaceService.isMultiRootWorkspaceEnabled,
                 isVisible: () => this.workspaceService.isMultiRootWorkspaceEnabled,
                 execute: async () => {
-                    const uri = await this.fileDialogService.showOpenDialog({
+                    const selection = await this.fileDialogService.showOpenDialog({
                         title: WorkspaceCommands.ADD_FOLDER.label!,
                         canSelectFiles: false,
-                        canSelectFolders: true
+                        canSelectFolders: true,
+                        canSelectMany: true,
                     });
-                    if (!uri) {
+                    if (!selection) {
                         return;
                     }
+                    const uris = Array.isArray(selection) ? selection : [selection];
                     const workspaceSavedBeforeAdding = this.workspaceService.saved;
-                    await this.addFolderToWorkspace(uri);
+                    await this.addFolderToWorkspace(...uris);
                     if (!workspaceSavedBeforeAdding) {
                         const saveCommand = registry.getCommand(WorkspaceCommands.SAVE_WORKSPACE_AS.id);
                         if (saveCommand && await new ConfirmDialog({
@@ -426,13 +428,17 @@ export class WorkspaceCommandContribution implements CommandContribution {
         }
     }
 
-    protected async addFolderToWorkspace(uri: URI | undefined): Promise<void> {
-        if (uri) {
+    protected async addFolderToWorkspace(...uris: URI[]): Promise<void> {
+        if (uris.length) {
+            const foldersToAdd = [];
             try {
-                const stat = await this.fileService.resolve(uri);
-                if (stat.isDirectory) {
-                    await this.workspaceService.addRoot(uri);
+                for (const uri of uris) {
+                    const stat = await this.fileService.resolve(uri);
+                    if (stat.isDirectory) {
+                        foldersToAdd.push(uri);
+                    }
                 }
+                await this.workspaceService.addRoot(foldersToAdd);
             } catch { }
         }
     }
--- a/packages/workspace/src/browser/workspace-service.ts
+++ b/packages/workspace/src/browser/workspace-service.ts
@@ -366,11 +366,12 @@ export class WorkspaceService implements FrontendApplicationContribution {
     }
 
     /**
-     * Adds a root folder to the workspace
-     * @param uri URI of the root folder being added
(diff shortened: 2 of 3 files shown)


packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -426,13 +428,17 @@ export class WorkspaceCommandContribution implements CommandContribution {
         }
     }
 
-    protected async addFolderToWorkspace(uri: URI | undefined): Promise<void> {
-        if (uri) {
+    protected async addFolderToWorkspace(...uris: URI[]): Promise<void> {
```
msujew, line 431: `undefined` is no longer a valid input type for this method. Does this qualify as a breaking change?

colin-grant-work: Yes, I think it does, even though we weren't availing ourselves of that option previously.

msujew: Right, I also don't think anyone is really affected by this change in any way. Should we still add it as a breaking change to the changelog?

msujew (approved):
Looks good to me. I can confirm that the changes allow to select multiple folders in the browser and Electron (Ubuntu) environment.

packages/workspace/src/browser/workspace-service.ts
```diff
@@ -373,6 +373,10 @@ export class WorkspaceService implements FrontendApplicationContribution {
         await this.spliceRoots(this._roots.length, 0, uri);
     }
 
+    async addRoots(...uris: URI[]): Promise<void> {
+        await this.spliceRoots(this._roots.length, 0, ...uris);
+    }
```
vince-fugnitto, line 378: Should we re-use functionality from `addRoot` rather than copy it's implementation?
Alternatively we can:

```ts
async addRoot(uri: URI | URI[]): Promise<void> {
```

vince-fugnitto (approved):
I confirmed that the behavior works as intended:
- verified without a `*.theia-workspace` file present and adding multiple roots
- verified with a previous `*.theia-workspace` file present and adding multiple roots
- adding a single workspace root works as before

packages/workspace/src/browser/workspace-service.ts
colin-grant-work: I'm happy to go either way. I decided to reimplement it because I could imagine someone having overridden `addRoot` to add additional arguments, and so that the plural would be parallel to `removeRoots`, but it is a bit silly to have both `addRoot` and `addRoots`, so if you vote for just modifying `addRoot` to accept multiple roots, I'm happy with that.

vince-fugnitto: @vince-fugnitto I'm also fine with either approach :+1: 
I think it would be fine to allow `addRoot` to accept multiple roots as an argument instead of implementing it as it's own method.

colin-grant-work: > ```ts
> async addRoot(uri: URI | URI[]): Promise<void> {
> ```

@vince-fugnitto, I've implemented this version ^^^ :-)

packages/workspace/src/browser/workspace-service.ts
```diff
@@ -366,11 +366,12 @@ export class WorkspaceService implements FrontendApplicationContribution {
...
-     * @param uri URI of the root folder being added
+     * Adds root folder(s) to the workspace
+     * @param uris URI or URIs of the root folder(s) to add
      */
-    async addRoot(uri: URI): Promise<void> {
-        await this.spliceRoots(this._roots.length, 0, uri);
+    async addRoot(uris: URI[] | URI): Promise<void> {
+        const toAdd = Array.isArray(uris) ? uris : [uris];
+        await this.spliceRoots(this._roots.length, 0, ...toAdd);
     }
```
vince-fugnitto, line 375: I believe it can be simplified to:

```ts
async addRoot(...uris: URI[]): Promise<void> {
    await this.spliceRoots(this._roots.length, 0, ...uris);
}
```

Calls like the following are allowed:

```ts
await this.workspaceService.addRoot(...workspaceFolders);
await this.workspaceService.addRoot(uri);
```

But I won't nitpick :)

vince-fugnitto (approved):
The changes work for me :+1:

PR #9695, opened 2021-07-02
Enable file links when task terminals complete

### LINKED ISSUE #9677: Link detection for files not working in task terminals
### Bug Description:

When output is generated in a terminal which has been started by a task, file link detection does not seem to work which is inconvenient when you want a quick way to open files mentioned in the terminal.

### Steps to Reproduce:

**Good Behavior**
1. Open VSCode to an empty folder and create a `.vscode/tasks.json` file as well as an empty file named `test.txt`.
2. Open a VSCode terminal and run the command `echo $PWD/test.txt`, hover over the terminal output and you should be able to `ctrl+click` on the file name
3. Populate the `tasks.json` file with the content below and run that task, step 2 should still work:

```
{
    "version": "2.0.0",
    "tasks": [
        {
            "command": "echo $PWD/test.txt",
            "label": "Test Task",
            "type": "shell"
        }
    ]
}
```

**Bad Behavior**
Repeat the above steps in Theia, but notice that the output generated from the task is not clickable

### Additional Information

- Operating System: Linux
- Theia Version: Master

kenneth-marut-work:
Signed-off-by: Kenneth Marut <kenneth.marut@ericsson.com>

#### What it does

Fixes #9677 by ensuring that the `terminal-linkmatcher-files`'s `getValidate` method does not fail when trying to retrieve the `TerminalWidget`'s `cwd` field when a task is terminated. This is done by adding a `lastCwd` field on the terminalWidget which is set every time a taskTerminal is started or reused. 

I've added this fix on the frontend although the underlying issue is due to the `base-terminal-server`'s call to `this.processManager.get(id)` being prone to failure when tasks end (i.e. 'terminal "id" does not exist'). 

#### How to test

- Before pulling this branch, open Theia's master branch and build. 
- Open Theia to a workspace with some files in it
- Add a task in your workspace that looks something like:

```
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "list filepaths",
      "command": "find $(pwd)",
      "type": "shell"
    }
  ]
}
```
- Run that task and observe that you cannot ctrl+click on the file paths
- Run the command `find $(pwd)` manually in a separate Theia terminal and observe that you *can* click the paths

- Exit Theia, and build this branch
- Run through the same steps and observe that paths are now clickable even after the task ends.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/terminal/src/browser/terminal-widget-impl.ts
+++ b/packages/terminal/src/browser/terminal-widget-impl.ts
@@ -61,6 +61,7 @@ export class TerminalWidgetImpl extends TerminalWidget implements StatefulWidget
     protected hoverMessage: HTMLDivElement;
     protected lastTouchEnd: TouchEvent | undefined;
     protected isAttachedCloseListener: boolean = false;
+    lastCwd: URI;
 
     @inject(WorkspaceService) protected readonly workspaceService: WorkspaceService;
     @inject(WebSocketConnectionProvider) protected readonly webSocketConnectionProvider: WebSocketConnectionProvider;
@@ -298,7 +299,10 @@ export class TerminalWidgetImpl extends TerminalWidget implements StatefulWidget
         }
         if (this.terminalService.getById(this.id)) {
             return this.shellTerminalServer.getCwdURI(this.terminalId)
-                .then(cwdUrl => new URI(cwdUrl));
+                .then(cwdUrl => {
+                    this.lastCwd = new URI(cwdUrl);
+                    return this.lastCwd;
+                }).catch(() => this.lastCwd);
         }
         return Promise.resolve(new URI());
     }
--- a/packages/task/src/browser/task-terminal-widget-manager.ts
+++ b/packages/task/src/browser/task-terminal-widget-manager.ts
@@ -23,6 +23,7 @@ import { PanelKind, TaskConfiguration, TaskWatcher, TaskExitedEvent, TaskServer,
 import { ProcessTaskInfo } from '../common/process/task-protocol';
 import { TaskDefinitionRegistry } from './task-definition-registry';
 import { WorkspaceService } from '@theia/workspace/lib/browser/workspace-service';
+import URI from '@theia/core/lib/common/uri';
 
 export interface TaskTerminalWidget extends TerminalWidget {
     readonly kind: 'task';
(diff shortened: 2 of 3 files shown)


packages/terminal/src/browser/terminal-widget-impl.ts
```diff
@@ -292,12 +293,12 @@ export class TerminalWidgetImpl extends TerminalWidget implements StatefulWidget
...
     }
 
-    get cwd(): Promise<URI> {
+    get cwd(): MaybePromise<URI> {
         if (!IBaseTerminalServer.validateId(this.terminalId)) {
             return Promise.reject(new Error('terminal is not started'));
         }
         if (this.terminalService.getById(this.id)) {
-            return this.shellTerminalServer.getCwdURI(this.terminalId)
+            return this.lastCwd || this.shellTerminalServer.getCwdURI(this.terminalId)
```
colin-grant-work, line 301: In case the backend knows something we don't shall we try that first, and then fall back to last known if that fails (and maybe set it, if it succeeds)

rcla:
@kenneth-marut-work

With the initial changes proposed by you, solve this problem that I have had: (Thanks)
[Firing up a Task](https://community.theia-ide.org/t/firing-up-a-task-from-an-extension/1342/19)

However with the last proposed change "switch ordering of fallback"
The same problem appears again

vince-fugnitto (approved):
I confirmed that it fixes the issue with file links in terminals when executing tasks :+1:

kenneth-marut-work:
> @kenneth-marut-work
> 
> With the initial changes proposed by you, solve this problem that I have had: (Thanks)
> [Firing up a Task](https://community.theia-ide.org/t/firing-up-a-task-from-an-extension/1342/19)
> 
> However with the last proposed change "switch ordering of fallback"
> The same problem appears again

Thanks for bringing attention to this issue. The behavior you've described *does* make sense with the ordering change since prior to the last commit, the backend method was never being called and thus never triggering the error (which is just hiding the underlying issue). Unfortunately, though the original commit did prevent the error from being thrown, it's probably not in the best interest to revert that change since I'm relying on the error being thrown to allow for the fallback inside the `catch` block. 

Maybe a potential fix for the future is to make these errors less invasive and to use warnings instead, however I think it may be out of the scope of this MR since several features of the task/terminal system likely rely on these errors.

colin-grant-work (approved):

colin-grant-work:
Thanks for addressing my comments. I agree that this likely isn't the place to modify the behavior of the backend error logs: the faulty logic to which they point still exists. The code here allows the frontend to work around the possibility of backend failure, and that's a step forward.

rcla:
@kenneth-marut-work @colin-grant-work

Thanks for the explanations, now I understand what concerns the backend errors for the future in another PR.

PR #9700, opened 2021-07-05
Remove important keyword in focus opacity style

xcariba:
Signed-off-by: Alexander Kozinko <xcariba@gmail.com>

#### What it does

Allows client to override opacity in focused elements without !important keyword.
I've tried to set existing style to my extension but there is a problem with opacity. There is no way to override it anywhere without marking every opacity change in css. I found [this conversation](https://community.theia-ide.org/t/important-keyword-in-css-for-opacity-in-focus/1786) and fixed it. I've checked existing styling and nothing changed. I also looked in all `:focus` overrides in this repository and found that it has a few overrides with this keyword. I think that it will be better if this base property can be changed without additional marks.

#### How to test

* Open Preferences, set focus in `Search Settings`, `Editor: Font Size`, `Editor: Render Whitespace` and check that focus is visible. Set focus on `Editor: Insert Spaces` Checkbox using Tab key and check that focus border is visible.
* Open debug view and use Tab to move focus on buttons, tree view headers and configuration selector, check that focus border is visible.
* Open `Find Command` view (Ctrl+Shift+P) and check that focus border is visible.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/style/index.css
+++ b/packages/core/src/browser/style/index.css
@@ -146,7 +146,7 @@ blockquote {
     outline-width: 1px;
     outline-style: solid;
     outline-offset: -1px;
-    opacity: 1 !important;
+    opacity: 1;
     outline-color: var(--theia-focusBorder);
 }
 


vince-fugnitto:
@xcariba please fill out the `how to test` section for reviewers.

xcariba:
> @xcariba please fill out the `how to test` section for reviewers.

Added steps to check that existing focus selector is visible. Please, let me know if I need to add anything or check to ensure that nothing is broken.

msujew (approved):
I couldn't find any regressions - i.e. no focused element which was missing its borders - due to the removed `!important` statement.

So LGTM then.

vince-fugnitto (approved):
I'm fine with the changes, I did not notice any regression during testing :+1:

PR #9707, opened 2021-07-06
Fix/document some of `keytar` issues

paul-marechal:
#### What it does

See commits.

Update documentation to provide guidelines when dealing with `keytar` on RHEL7-like distros.

Add `keytar` to the list of native packages to rebuild when doing `yarn rebuild:*`.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/doc/Developing.md
+++ b/doc/Developing.md
@@ -48,6 +48,7 @@ For Windows instructions [click here](#building-on-windows).
      - [Linux](#linux)
      - [Windows](#windows)
      - [macOS](#macos)
+     - [node-keytar](#keytar)
      - [Root privileges errors](#root-privileges-errors)
 
 ## Prerequisites
@@ -64,10 +65,13 @@ Some additional tools and libraries are needed depending on your platform:
   - [gcc](https://gcc.gnu.org/) (or another compiling toolchain)
   - [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/)
   - build-essential: `sudo apt-get install build-essential`
-  - Dependencies for `native-keymap` node native extension:
-    - Debian-based: `sudo apt-get install libx11-dev libxkbfile-dev`
-    - Red Hat-based: `sudo yum install libX11-devel.x86_64 libxkbfile-devel.x86_64 # or .i686`
-    - FreeBSD: `sudo pkg install libX11`
+  - Dependencies for node native extensions:
+    - `native-keymap`:
+      - Debian-based: `sudo apt-get install libx11-dev libxkbfile-dev`
+      - Red Hat-based: `sudo yum install libX11-devel.x86_64 libxkbfile-devel.x86_64 # or .i686`
+      - FreeBSD: `sudo pkg install libX11`
+    - `node-keytar`:
+      - Debian-based: `sudo apt-get install libsecret-1-dev`
 
 - Linux/MacOS
   - [nvm](https://github.com/nvm-sh/nvm) is recommended to easily switch between Node.js versions.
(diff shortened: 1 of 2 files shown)


vince-fugnitto (approved):
The changes look good to me 👍 
I believe @colin-grant-work confirmed it works on RHEL7 with the updates as well.

doc/Developing.md
```diff
@@ -47,7 +47,8 @@ For Windows instructions [click here](#building-on-windows).
  - [**Troubleshooting**](#troubleshooting)
      - [Linux](#linux)
      - [Windows](#windows)
-     - [macOS](#macos)
+     - [macOS](#macOS)
+     - [RHEL7](#RHEL7)
```
colin-grant-work, line 51: Some of the people reporting similar errors online were using CentOS, so I might go with RHEL7 / CentOS7 here.

doc/Developing.md
```diff
@@ -470,6 +487,16 @@ If you already have Xcode installed, but you see the `xcode-select: error: tool
 
 The solution is the same if you have updated to `10.14` (Mojave) and you can see the `gyp: No Xcode or CLT version detected!` error. More details [here](https://github.com/nodejs/node-gyp#on-macos).
 
+### RHEL7
+
+See prerequisites for [`node-keytar`](#prerequisites)
+
+You might need to set the following environment variable before doing `yarn`:
```
colin-grant-work, line 494: It might be useful to describe the problem this addresses. Something like `At runtime, you may encounter an error with the message <error message from issue>. This can be fixed by forcing pre-built dependencies to rebuild from source by setting the following environment variable before `yarn`:` (you may also need to clean, since that script gets run on `install`, so if you try setting the environment variable when you already have the package, it may not get run.)

doc/Developing.md
```diff
@@ -470,6 +487,18 @@ If you already have Xcode installed, but you see the `xcode-select: error: tool
 
 The solution is the same if you have updated to `10.14` (Mojave) and you can see the `gyp: No Xcode or CLT version detected!` error. More details [here](https://github.com/nodejs/node-gyp#on-macos).
 
+### Keytar
+
+See prerequisites for [`node-keytar`](#prerequisites)
+
+On some distributions (e.g. RHEL7, CentOS7) the prebuilt binaries for `node-keytar` might crash at runtime because of some missing shared libraries (libc?)
```
vince-fugnitto, line 494: Minor: offline discussion to modify the following as it looks incomplete as is:

_Current_: `(libc?)`

_To_: `(ex: libc)` or `(libc, etc.)`

vince-fugnitto (approved):

vince-fugnitto (approved):

paul-marechal:
There is a build error with keytar when rebuilding for Electron, I'll fix it first.

paul-marechal:
More details on the issue I am seeing:

Without this PR, rebuilding modules for Electron doesn't rebuild `keytar` because it is missing from our hardcoded list.

But when adding `keytar` to that list, it fails for me when trying to rebuild `keytar`. What happens is:

- I am using Node 12.22.3 which supports NAPI 8.
- Our target Electron version is built on Node 12.14.1 which supports NAPI 5.
- When running `electron-rebuild` which runs `node-gyp` under the hood, it uses my Node to run the cli scripts.
- `node-gyp` runs `<current-node> -p process.versions.napi` to define the GYP variable `napi_build_version`.
- `napi_build_version` is set to `8`.
- C++ build runs, and since we're building for Electron it will use Electron's header files for the C NAPI.
- `keytar` is made using C++ which has its own header files wrapping the C NAPI.
- Those C++ headers have a `#if NAPI_VERSION > 5` that controls if it will try to refer to more recent NAPI APIs.
- Because `NAPI_VERSION` is set to 8 and Electron only support 5, it leads to "missing symbol" errors.

When going into my `node_modules` and hardcoding that `NODE_VERSION` GYP variable to 5, it works all well.
Alternatively: when installing Node 12.14.1 using nvm, it works too.

I have yet to find a way to force both `electron-rebuild` and `node-gyp` to not use the current Node executable to determine what version of NAPI to expect when rebuilding for Electron...

This build issue is most likely a problem with `node-gyp` and `electron-rebuild` :/

colin-grant-work:
> More details on the issue I am seeing:
> 
> * I am using Node 12.22.3 which supports NAPI 8.
> * Our target Electron version is built on Node 12.14.1 which supports NAPI 5.

Would upgrading our Electron (not a trivial task, I know) or narrowing our supported Node range (annoying, but feasible) address these problems?

paul-marechal:
@colin-grant-work yes those are valid avenues.

paul-marechal:
On the other hand, I wonder if we even need to rebuild `keytar` for Electron. The whole point of using NAPI is to be ABI compatible across Node versions. Not sure how that translates to Electron, but I would assume it should work there as well?

paul-marechal:
Closing for now.

PR #9709, opened 2021-07-07
Workspace: Rename files/folder is case sensitive

### LINKED ISSUE #9663: Rename of files and folder are case insensitive
### Bug Description:
On rename files or folders the registry is not taking into count.
The rename should be case-sensitive.

### Steps to Reproduce:

1. Create a new file.
2. Change some characters of the file name to a different registry.
3. Wrong error appears about the file already exists.

image

### Additional Information

- Theia Version: 1.14.0

OmarSdt-EC:
#### What it does

Fixes #9663
This commit fixes the rename command for windows users by making possible
the case sensitive rename of files and folders. Also, it adds the test file
`workspace-commands.spec.ts` which verifies that the implementation works as expected.

#### How to test

If your operating system is Windows, you can test this commit with the following steps in the browser or electron target :
1. Create a new folder : `foo`
2. Create two files in it : `a.txt` and `b.txt`
3. Rename `a.txt` to `A.txt`
4. It should let you do the rename
5. Rename `b.txt` to `a.txt`
6. It should not let you do the rename and the message : "A file or folder 'a.txt' already exists at this location" should appear

You can also test the commit by running : 
` npx run build @theia/workspace && npx run test @theia/workspace`

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/workspace/src/browser/workspace-commands.ts
+++ b/packages/workspace/src/browser/workspace-commands.ts
@@ -14,9 +14,10 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-import { inject, injectable } from '@theia/core/shared/inversify';
+import { inject, injectable, postConstruct } from '@theia/core/shared/inversify';
 import URI from '@theia/core/lib/common/uri';
 import { SelectionService } from '@theia/core/lib/common/selection-service';
+import { ApplicationServer } from '@theia/core/lib/common/application-protocol';
 import { Command, CommandContribution, CommandRegistry } from '@theia/core/lib/common/command';
 import { MenuContribution, MenuModelRegistry } from '@theia/core/lib/common/menu';
 import { CommonMenus } from '@theia/core/lib/browser/common-frontend-contribution';
@@ -34,7 +35,7 @@ import { WorkspaceCompareHandler } from './workspace-compare-handler';
 import { FileDownloadCommands } from '@theia/filesystem/lib/browser/download/file-download-command-contribution';
 import { FileSystemCommands } from '@theia/filesystem/lib/browser/filesystem-frontend-contribution';
 import { WorkspaceInputDialog } from './workspace-input-dialog';
-import { Emitter, Event } from '@theia/core/lib/common';
+import { Emitter, Event, OS } from '@theia/core/lib/common';
 import { FileService } from '@theia/filesystem/lib/browser/file-service';
 import { FileStat } from '@theia/filesystem/lib/common/files';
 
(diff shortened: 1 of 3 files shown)


packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -394,8 +394,19 @@ export class WorkspaceCommandContribution implements CommandContribution {
...
+            if (initialValue) {
+                const actualUri = parent.resource.resolve(initialValue);
+                if (actualUri.isEqual(childUri, false)) {
+                    return '';
+                };
+            }
+            // Determine if the case-insensitive version already exists at the location.
+            if (parent.children?.some(child => child.resource.isEqual(childUri, false))) {
+                return `A file or folder "${this.trimFileName(name)}" already exists at this location.`;
+            }
```
msujew, line 409: I don't believe that this check is actually necessary. Assuming `exists` is already true on windows, we don't need to determine again whether a case-insensitive version already exists, since that's exactly what `exists` represents.

OmarSdt-EC: That's true, I removed it thank you !

msujew (approved):
I tested the changes on Windows and Ubuntu. The behavior on Ubuntu didn't change, while it fixed the issue on Windows. I tried:

* Renaming a file to the same name but with a different case.
* Renaming a file to create a name collision with an existing file with the **same case** as the other file.
* Renaming a file to create a name collision with an existing file with a **different case**.

LGTM 👍

packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -368,19 +380,26 @@ export class WorkspaceCommandContribution implements CommandContribution {
...
+        }
+        return this.validateFileName(name, parent, false);
+    }
+
     /**
      * Returns an error message if the file name is invalid. Otherwise, an empty string.
      *
      * @param name the simple file name of the file to validate.
      * @param parent the parent directory's file stat.
      * @param recursive allow file or folder creation using recursive path
```
vinokurig, line 395: This has to be updated since it is renamed in the method

vinokurig:
The code is OK for me but I can't approve it because I don't have a Windows workstation to test it.

packages/workspace/src/browser/workspace-commands.ts
OmarSdt-EC: Hi, unfortunately, I didn't understand what you meant, can you explain it to me ?

msujew: @OmarSdt-EC You renamed the `recursive` argument to `allowNested`, but didn't change the documentation to reflect that. At least I think that's what he meant.

msujew (approved):
I tested everything again, as the code changed a lot since the last time I took a look at it. However, everything still performs as expected.

packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -196,10 +197,18 @@ export class WorkspaceCommandContribution implements CommandContribution {
...
     @inject(WorkspaceCompareHandler) protected readonly compareHandler: WorkspaceCompareHandler;
+    @inject(ApplicationServer) protected readonly applicationServer: ApplicationServer;
 
     private readonly onDidCreateNewFileEmitter = new Emitter<DidCreateNewResourceEvent>();
     private readonly onDidCreateNewFolderEmitter = new Emitter<DidCreateNewResourceEvent>();
 
+    protected _backendOS: Promise<OS.Type>;
+
+    @postConstruct()
+    async onStart(): Promise<void> {
```
paul-marechal, line 208: Please rename to either `async postConstruct()` or `async init()`.

packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -196,10 +197,18 @@ export class WorkspaceCommandContribution implements CommandContribution {
     @inject(WorkspaceDeleteHandler) protected readonly deleteHandler: WorkspaceDeleteHandler;
     @inject(WorkspaceDuplicateHandler) protected readonly duplicateHandler: WorkspaceDuplicateHandler;
     @inject(WorkspaceCompareHandler) protected readonly compareHandler: WorkspaceCompareHandler;
+    @inject(ApplicationServer) protected readonly applicationServer: ApplicationServer;
 
     private readonly onDidCreateNewFileEmitter = new Emitter<DidCreateNewResourceEvent>();
     private readonly onDidCreateNewFolderEmitter = new Emitter<DidCreateNewResourceEvent>();
 
+    protected _backendOS: Promise<OS.Type>;
```
paul-marechal, line 205: No need for the leading underscore AFAIK?

msujew:
@paul-marechal Anything blocking you from approving/retesting?

paul-marechal (approved):
Code LGTM, and I confirm that I am now able to rename files by just changing the case on Windows.

Issue #7275, opened 2020-03-04
proposal: update scripts to not lint when building


vince-fugnitto:
**Description**

As a developer, I'd like to be able to `build` the repository (and extensions) without the need to `lint`.
The linting process is quite long and is not necessary during the build. I'd like there to be a clear distinction between commands like `build`, `lint` and `test`. We'd need to update the `prepare` script to exclude the call to `lint`.

**Pros:**
- The initial build will be much quicker
- Linting an entire project is excessive, linting should happen for extensions that are updated by developers during a pull-request
- The CI already handles linting the entire project so there is no need to add the overhead when simply building

Issue #9153, opened 2021-03-05
A Script to Scaffold Applications around Extensions


colin-grant-work:
> The basic goal here is to make the process of developing, consuming, and forking extensions developed outside the main Theia repo smooth, preferably with a workflow very similar to working on Theia itself. The suggestion below is one idea, but certainly not the only approach.

At the moment, the process of developing and consuming Theia packages from outside this repository is not very smooth. Two that I regularly work with, the [Theia Trace Extension](https://github.com/theia-ide/theia-trace-extension) and [Theia CPP Extensions](https://github.com/eclipse-theia/theia-cpp-extensions) both come wrapped in their own example applications. That can allow for some customization, but comes with the overhead of ensuring that the wrapper is up to date with Theia as well as the risk of getting far enough away from Theia that the developer experience working on the extension is quite different from working on a package inside Theia - making for a steeper learning curve and unnecessary friction.

One solution would be a Theia extension script that could scaffold a full Theia application around an extension for demos / testing and then unwrap the extension again.

### Feature Description:

Something like this:

1. Before wrapping:
```
└── myExtensionRepo
    ├── .gitignore
    ├── package.json
    └── src
        ├── browser
        ├── common
        └── node
```
2. execute `yarn wrap = 'theia-ext-scripts wrap'`
3. See this:
```
└── myExtensionRepo
    ├── .gitignore
    ├── package.json
    ├── src
    │   ├── browser
    │   ├── common
    │   └── node
    └── wrapper
        ├── examples
        │   ├── browser # With some @theia packages specified by myExtensionRepo/package.json
        │   └── electron
        └── packages
            └── my-extension
                └── ... # Content here copied or symlinked from myExtensionRepo/...
```
Where the `wrapper` is `.gitignore`d, and everything necessary for the package to operate in the example applications is copied - or, better, symlinked - from its location in the upper folder.

This would have a number of benefits:
1. Working on a standalone Theia extension would look a lot more like working on a VSCode extension - just the source code of the extension.
2. Developers who wanted to start a Theia extension could take an existing package from Theia as their model, and just take queues for any changes in config from Theia, reducing integration friction.
3. Ensuring that their extension worked with latest Theia would only mean keeping their ext-scripts dependency up to date and generating the wrapper.
[...]

PR #9710, opened 2021-07-07
rework build system

paul-marechal:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/7275.

Simplify `scripts/compile-references.js`. TypeScript Server now properly
understand TypeScript references for navigation, so we can simplify our
`tsconfig` setup.

Rewrite some `npm/yarn` scripts to do less on install (`yarn` or `yarn
install`). It will only run the required `prepare` scripts across our
packages, wire `tsconfig.json` files and build the TypeScript sources.

To build the example applications you need to run one of `yarn browser
build`, `yarn electron build` or `yarn build:examples`.

Linting is done as part of CI, but it won't automatically run locally
anymore. You can run `yarn lint` to lint your code. For the best
experience you should also enable the ESLint extension in your IDE.

#### How to test

Installation:

- `yarn` alone should not build the example applications anymore, nor should it lint.

Building:

- `yarn build` should build TypeScript and both apps.
- `yarn compile` should build TypeScript.
- `yarn browser build` should build both TypeScript and the browser app.
- `yarn electron build` should build both TypeScript and the electron app.
- `yarn build:examples` should build both apps.

Watching:

- `yarn watch` should watch TypeScript and both example apps.
- `yarn watch:ts` should watch only TypeScript.
- `yarn browser watch` should watch both TypeScript and the browser app.
- `yarn electron watch` should watch both TypeScript and the electron app.

Linting:

- `yarn lint` should lint TypeScript sources.

#### Review checklist

#### Reminder for reviewers


paul-marechal:
This is especially useful when using the main repo as a submodule for a different repository such as a Theia extension:

My approach was to clone Theia as a submodule in the extension repo, then add `submodules/theia/packages/*` and `submodules/theia/dev-packages/*` to the parent repo yarn workspaces. Finally, I would run the TypeScript compilation in Theia to simply build the TypeScript packages, as this is all that is required for dependents to consume.

`yarn link` doesn't play nicely with our Theia building tools for some reason, hence why this alternative approach is interesting.

This PR also makes developing a bit more comfortable: only the minimal amount of work required is done, if you need more you need to explicitly trigger it. E.g. building a given app, linting, etc...

doc/Developing.md
```diff
@@ -145,22 +163,45 @@ installing
...
 ```
+
+To learn more and understand precisely what's going on, please look at scripts in [package.json](../package.json).
+
+## Build Everything
+
+```sh
+cd $THEIA
+yarn build
+```
```
colin-grant-work, line 193: The only thing that gives me (slight) pause here is the change in behavior of the plain `yarn` command. If we want to have more options in how we build, there are basically two options: reduce `yarn`'s functionality and create a new command that has its old functionality (this PR), or leave `yarn` as was and create a new script with the reduced functionality of installing and transpiling. I don't think it matters much either way. Either we tell new people 'just `yarn` and then start' or we tell new people 'just `yarn build` and then start.' I might incline toward leaving `yarn` alone so that casual users' workflows don't change, but there's a certain iconicity to the shorter command (`yarn`) do less work than the longer command (`yarn build`).

paul-marechal: My biggest pain is that `yarn` alone is a shortcut to `yarn install` meaning it is meant to deal with dependencies. At best it is common practice to prepare the packages by transpiling sources, but linting and building two applications is overkill. This is especially annoying to work with when we need to work with dependencies when you have to keep an eye at the `yarn` output to see when to `ctrl+c` it.

With this PR, doing `yarn` will at least put the project in a state that can be published/consumed (we don't need to build the example apps to publish).

paul-marechal: More importantly: `yarn (install)` is the first command you HAVE to run in order to work with any npm package. Making that command do everything means there is no way to do less when preferable to do so.

colin-grant-work: Sounds fair to me. 👍

colin-grant-work:
> This is especially useful when using the main repo as a submodule for a different repository such as a Theia extension

This sounds it goes some way to addressing some of the pain points described in #9153, as well.

tsmaeder:
> yarn link doesn't play nicely with our Theia building tools for some reason, hence why this alternative approach is interesting.

Could you detail in what way it does not play well with `yarn link`? it seems to work for me, but I fear I'm missing something.

scripts/compile-references.js
```diff
@@ -25,25 +25,20 @@
...
 const ROOT = path.join(__dirname, '..');
 
-const DRY_RUN = popFlag(process.argv, '--dry-run');
-
-const FORCE_REWRITE = popFlag(process.argv, '--force-rewrite') && !DRY_RUN;
+const FORCE_REWRITE = process.argv.includes('--force-rewrite');
 
 /** @type {{ [packageName: string]: YarnWorkspace }} */
-const YARN_WORKSPACES = JSON.parse(cp.execSync('yarn --silent workspaces info').toString());
+const YARN_WORKSPACES = JSON.parse(cp.execSync('yarn --silent workspaces info', { cwd: ROOT }).toString());
```
tsmaeder, line 41: I don't quite understand why we need to rewrite references as part of the build at all: aren't they static and inferred by the usage relation between packages?

package.json
```diff
@@ -30,52 +30,45 @@
...
-    "test:references": "node scripts/compile-references --dry-run",
+    "docs": "rimraf gh-pages/docs/next && typedoc --tsconfig configs/typedoc-tsconfig.json --options configs/typedoc.json .",
+    "test": "yarn test:theia && yarn test:electron && yarn test:browser",
     "test:theia": "run test \"@theia/!(example-)*\" --stream --concurrency=1",
-    "test:browser": "yarn rebuild:browser && run test \"@theia/example-browser\"",
-    "test:electron": "yarn rebuild:electron && run test \"@theia/example-electron\"",
+    "test:browser": "yarn browser test",
+    "test:electron": "yarn electron test",
     "rebuild:clean": "rimraf .browser_modules",
-    "rebuild:browser": "theia rebuild:browser",
```
tsmaeder, line 67: Where did these go?

package.json
```diff
@@ -30,52 +30,45 @@
...
     "postinstall": "node scripts/post-install.js",
-    "prepare": "yarn prepare:references && yarn --cwd packages/core prepare && yarn prepare:build && yarn prepare:hoisting && yarn download:plugins",
-    "prepare:references": "node scripts/compile-references.js",
-    "prepare:build": "yarn build && run lint && run build \"@theia/example-*\" --stream --parallel",
-    "prepare:hoisting": "theia check:hoisted -s",
-    "clean": "yarn lint:clean && node scripts/run-reverse-topo.js yarn clean",
-    "build": "tsc -b configs/root-compilation.tsconfig.json",
-    "watch": "tsc -b configs/root-compilation.tsconfig.json -w",
-    "lint": "run lint",
+    "prepare": "yarn tsreferences && lerna run prepare && yarn build:ts",
```
tsmaeder, line 49: I kind of regret that we don't have a single command anymore that will "build everyting" after checking out the source. As I understand it, we'll have to `yarn install && yarn build` now, right?

doc/Developing.md
```diff
@@ -177,59 +218,101 @@ From the package:
...
 In the root directory run:
 
-    yarn run build
+```sh
+yarn run build
+```
 
 ## Watching
 
+### Watch the TypeScript packages
```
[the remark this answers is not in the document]
paul-marechal: Then there's nothing to watch?

Do we have such non-TypeScript package?

examples/browser/package.json
```diff
@@ -56,13 +56,14 @@
...
-    "watch": "yarn build --watch",
-    "start": "theia start --plugins=local-dir:../../plugins",
+    "rebuild": "theia rebuild:browser",
+    "build": "tsc -b && theia build --mode development",
+    "watch": "concurrently --kill-others -n tsc,bundle -c red,yellow \"tsc -b -w --preserveWatchOutput\" \"yarn watch:bundle\"",
+    "watch:bundle": "theia build --watch --mode development",
+    "start": "yarn rebuild && theia start --plugins=local-dir:../../plugins",
     "start:debug": "yarn start --log-level=debug",
-    "test": "theia test . --plugins=local-dir:../../plugins --test-spec=../api-tests/**/*.spec.js",
+    "test": "yarn rebuild && theia test . --plugins=local-dir:../../plugins --test-spec=../api-tests/**/*.spec.js",
```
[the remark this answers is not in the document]
paul-marechal: Good catch: it is a typo and should be `rebuild:browser`, it is to ensure native dependencies are built for Node and not Electron.

scripts/compile-references.js
```diff
@@ -54,66 +49,36 @@ for (const [packageName, yarnWorkspace] of Object.entries(YARN_WORKSPACES)) {
...
-        }
-    }
-
-} catch (error) {
+main().catch(error => {
     console.error(error);
     process.exitCode = 1;
+})
+
+async function main() {
```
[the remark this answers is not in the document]
paul-marechal: `main` is usually a function entry point, like C/C++/C#/Java etc... If I could I would use top-level await but our supported Node version doesn't supports this yet.

package.json
paul-marechal: I can add a `yarn everything` or something similiar, otherwise yes you have to do what you said.

package.json
```diff
@@ -154,5 +147,12 @@
     "vscode-editorconfig": "https://open-vsx.org/api/EditorConfig/EditorConfig/0.14.4/file/EditorConfig.EditorConfig-0.14.4.vsix",
     "vscode-eslint": "https://open-vsx.org/api/dbaeumer/vscode-eslint/2.1.1/file/dbaeumer.vscode-eslint-2.1.1.vsix",
     "vscode-references-view": "https://open-vsx.org/api/ms-vscode/references-view/0.0.47/file/ms-vscode.references-view-0.0.47.vsix"
+  },
+  "dependencies": {
+    "@theia/node-pty": "0.9.0-theia.6",
```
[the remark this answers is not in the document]
paul-marechal: Good catch, I think yarn did that because I sure didn't, I'll remove those.

doc/Developing.md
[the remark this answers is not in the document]
paul-marechal: This mostly refers to TypeScript compilation which happens before bundling, so if we have a pure JS extension then there will be nothing to watch.

scripts/compile-references.js
paul-marechal: They are static and inferred. The reason we do it on build is to make sure everything is in sync since building relies on it.

scripts/compile-references.js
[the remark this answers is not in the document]
colin-grant-work: Maybe just name it the same thing as the module, in camelCase (`compileReferences`), the same way we (often) name modules the same thing as the main class they export?

paul-marechal:
> Could you detail in what way it does not play well with `yarn link`? it seems to work for me, but I fear I'm missing something.

All the times I remember doing `yarn link` in each Theia package for consumption in third-party Theia extension projects, it ended up with Webpack getting confused and duplicating packages in the output bundles for some reason, leading to runtime errors.

Since we migrated to Webpack 5 maybe this won't be an issue anymore...

So independently from that, this PR still simplifies the `compile-references.js` script, allows one to manage dependencies without triggering a full build, and cleans up scripts.

packages/callhierarchy/tsconfig.json
```diff
@@ -10,10 +10,13 @@
...
     {
-      "path": "../core/compile.tsconfig.json"
+      "path": "../core"
     },
     {
-      "path": "../editor/compile.tsconfig.json"
+      "path": "../editor"
+    },
+    {
+      "path": "../monaco"
```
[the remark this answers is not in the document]
paul-marechal: `callhierarchy` depends on `monaco` as of master already:

https://github.com/eclipse-theia/theia/blob/faae393f66c779c9640c77b0d23c60a4d0d10f62/packages/callhierarchy/package.json#L8
https://github.com/eclipse-theia/theia/blob/faae393f66c779c9640c77b0d23c60a4d0d10f62/packages/callhierarchy/compile.tsconfig.json#L19

scripts/compile-references.js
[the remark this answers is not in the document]
paul-marechal: And yes, making I/O stuff concurrent always yields better performance. Everything was sync before, and as part of simplifying the script I made it async where it made sense.

scripts/compile-references.js
paul-marechal: Note that it won't write anything on disk if everything is already synced.

package.json
paul-marechal: In their respective example applications. I just tweaked the `rebuild` scripts to better handle the use case when it is invoked from non-root locations.

examples/browser/package.json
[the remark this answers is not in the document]
paul-marechal: Ah not a typo, there's a `rebuild` script in the same `package.json` now.

package.json
paul-marechal: Added a `yarn all` command that does everything again.

package.json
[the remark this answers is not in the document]
paul-marechal: Turns out this was due to an aborted `theia rebuild:electron` command: it used to modify the current `package.json` to pass arguments to `electron-rebuild`... Fixed it to not do that anymore, and rather use CLI options.

doc/Developing.md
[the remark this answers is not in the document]
paul-marechal: I understand `TypeScript package` as a package written using TypeScript and requiring transpilation.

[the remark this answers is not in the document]
tsmaeder: So it's really "typescript files", because some packages may container both *.js and *.ts files, right? Also, not sure how that works, exactly, but don't we have to reload the code when *.js files change, as well?

doc/Developing.md
```diff
@@ -145,91 +163,153 @@ installing
...
 
-## Watching
+```sh
+yarn watch:ts
+```
 
 ### Watch the core and extension packages
 
-To rebuild each time a change is detected run:
+To rebuild _everything_ each time a change is detected run:
```
tsmaeder, line 268: "Everything" being both the regular packages and the example applications?

examples/browser/package.json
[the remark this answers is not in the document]
tsmaeder: But was _not_ doing some kind of "rebuild:browser" a mistake in the old code?

package.json
tsmaeder: So it would be `yarn browser rebuild` now, right.

packages/callhierarchy/tsconfig.json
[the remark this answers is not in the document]
tsmaeder: I think the dependency is there by mistake, but I agree it's not your problem in this PR.

scripts/compile-references.js
tsmaeder: So I _did_ understand that right. It's a pity the tsc compiler can't infer those if we tell it what the set of source packages are 🤷  Fair enough.

scripts/compile-references.js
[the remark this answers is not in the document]
tsmaeder: > And yes, making I/O stuff concurrent always yields better performance

Well, that's only true if you can do something in the meanwhile that is not I/O. But anyway...async code _is_ more complex to understand at run time, which is a cost to be measured against the gains. `await` makes it _look_ simple, but it isn't.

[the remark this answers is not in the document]
tsmaeder: I like @colin-grant-work's idea about the "script-named" top functions.

doc/Developing.md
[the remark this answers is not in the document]
paul-marechal: In Theia the watch scripts are related to code transformations. There's basically two steps: TypeScript compilation and Webpack bundling. This documentation refers to the former. Javascript sources don't require any change during the first step as well, but bundling still depend on them indeed.

examples/browser/package.json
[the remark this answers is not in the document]
paul-marechal: Not so much a mistake as something that I though I could improve here :)

package.json
paul-marechal: Yes.

scripts/compile-references.js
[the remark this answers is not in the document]
paul-marechal: I prefer `main` for this kind of use case, but I still renamed it to `compileTypeScriptReferences`.

msujew:
@paul-marechal I'll review this, if you're still interested in getting this merged. Would you mind rebasing and fixing merge conflicts?

paul-marechal:
@msujew thanks for proposing! I just rebased. Note that one of the test lifecycle keeps failing and I don't know why...

paul-marechal:
@msujew thanks for the idea! I finally addressed your comments.

I tried improving the GitHub workflows too but I'll keep that for another PR.

.gitpod.yml
```diff
@@ -12,7 +12,7 @@ ports:
   - port: 9339 # Node.js debug port
     onOpen: ignore
 tasks:
-  - init: yarn --network-timeout 100000
+  - init: yarn --network-timeout 100000 && yarn build:examples
```
msujew, line 15: I was wondering why my Theia installation in gitpod didn't run any extensions. I think we should download extensions here as well:

```suggestion
  - init: yarn --network-timeout 100000 && yarn build:examples && yarn download:plugins
```

paul-marechal: Ah good catch!

msujew (approved):
I like the change. I could not find any build-regressions and the new watch options improve my personal development experience 👍

paul-marechal:
There seems to be an issue when rebuilding native dependencies for Electron where `electron-rebuild` doesn't rebuild `@theia/node-pty` for some reason, will fix before merging.

paul-marechal:
We decided to update `electron-rebuild` now while we are improving the rebuild scripts, but this ended up pulling a new dependency requiring a CQ: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23694

tsmaeder:
@paul-marechal That's why I usually push back on suggestions to do "one more useful thing" in my PR's. Small, incremental changes that make the code base better is my ideal. As long as the change in the PR stands up on its own, pushing back on change requests is fine by me. Do followups, I trust you to do them :-)

paul-marechal:
@tsmaeder upgrading `electron-rebuild` is something we'll want to do anyway. But I can extract this is its own PR for sure.

tsmaeder:
@paul-marechal are you aware the tests are failing?

paul-marechal:
@tsmaeder painfully yes, I am yet to find why...

CHANGELOG.md
```diff
@@ -81,15 +82,10 @@
...
-  - `SidebarBottomMenu` interface is renamed `SidebarMenu` and handles not only bottom menu's.
-  - Changed style class name from `theia-sidebar-bottom-menu` to `theia-sidebar-menu`
-  - `TheiaDockPanel` constructor takes a new parameter `preferences`
+- [core] registering toolbar items for commands that explicitly target a `ViewContainer` rather than a child widget may not behave as expected. Such registrations should be made in the `ViewContainer` by overriding the `updateToolbarItems` method and using the `registerToolbarItem` utility. See the modifications to the `scm` and `vsx-registry` packages in the PR for examples. [#9798](https://github.com/eclipse-theia/theia/pull/9798)
+- [vsx-registry] `VSXExtensionsContribution` no longer implements `TabBarToolbarContribution` and is not bound as such. Extensions of the class that expect such behavior should reimplement it with caution. See caveats in PR. [#9798](https://github.com/eclipse-theia/theia/pull/9798)
+- [core] `handleExpansionToggleDblClickEvent` in `TreeWidget` can no longer be overridden. Instead, `doHandleExpansionToggleDblClickEvent` can be overridden. [#9877](https://github.com/eclipse-theia/theia/pull/9877)
+
```
vince-fugnitto, line 88: Revert?

doc/Developing.md
```diff
@@ -86,32 +86,45 @@ Some additional tools and libraries are needed depending on your platform:
...
 
-    git clone https://github.com/eclipse-theia/theia \
+```sh
+git clone https://github.com/eclipse-theia/theia \
     && cd theia \
     && yarn \
-    && cd examples/browser \
-    && yarn run start
+    && yarn browser build \
+    && yarn browser start
```
vince-fugnitto, line 94: Given that it does not automatically happen anymore, do we need to mention to download plugins?

vince-fugnitto (commented):
I confirmed the behavior works well:
- building
- starting both browser and electron apps
- rebuilding native modules works well on Linux

vince-fugnitto (approved):
Don't forget to squash the 17 commits on merge ;)

paul-marechal:
@tsmaeder @msujew last heads up before merging. If you guys are still fine with the changes then let's break everyone's muscle memory... I'll open and pin an issue to mention the new commands to try and help with the new workflows.

msujew (approved):
@paul-marechal Still fine with merging, everything looks good 👍

paul-marechal:
I'll just wait on CI to turn green...

thegecko:
Really miss the easy `yarn start:browser` command in the root to ensure native modules are built correctly and start the browser example.

msujew:
@thegecko I think you should be able to run `yarn browser start` to do exactly the same as the old `yarn start:browser`.

thegecko:
> I think you should be able to run `yarn browser start`

Confirmed, thanks!

Issue #6601, opened 2019-11-21
Frontend plugins are broken


mmorhun:
### Description
Theia cannot start frontend plugins any more. Despite frontend plugins are displayed in `Plugins` panel they aren't loaded correctly, so there is no contributions from them.

### Reproduction Steps

1. Generate a hello world frontend plugin.
2. Start Theia with it (use `THEIA_PLUGINS` or `HOSTED_PLUGIN` environment variable)

**OS and Theia version:**
Fedora 31, Theia `7a77a3b36416fca1319e66048feb1a1d8264401c` (master at the moment of this issue creation).

**Diagnostics:** 

<details>
<summary>Theia log</summary>

```
root INFO Theia app listening on http://0.0.0.0:3000.
root INFO PluginTheiaDirectoryHandler: accepting plugin with path /home/mmorhun/projects/theia/plugins/2222
root INFO Deploying frontend plugin "2222@0.0.1" from "dist/2222-frontend.js"
root INFO Detected keyboard layout from browser API: US (PC)
root WARN Collided keybinding is ignored;  {"command":"editor.action.toggleTabFocusMode","keybinding":"ctrl+m"}  collided with  {"command":"core.toggleMaximized","keybinding":"ctrl+m"}
root WARN Could not register keybinding:
  {"command":"editor.action.toggleTabFocusMode","keybinding":"ctrl+m"}
Error: "ctrl+m" is in collision with something else [scope:0]
root WARN Collided keybinding is ignored;  {"command":"passthrough","keybinding":"ctrl+k","context":"terminalActive"}  collided with  {"command":"terminal:clear","keybinding":"ctrlcmd+k","context":"terminalActive"}
root WARN Could not register keybinding:
  {"command":"passthrough","keybinding":"ctrl+k","context":"terminalActive"}
Error: "ctrl+k" is in collision with something else [scope:0]
root INFO Using Git [2.23.0] from the PATH. (/usr/bin/git)
root INFO [nsfw-watcher: 84352] Started watching: /home/mmorhun/.theia
root WARN EditorNavigationContribution.onStart is slow, took: 558.870000000752 ms
root INFO Changed application state from 'init' to 'started_contributions'.
root INFO Changed application state from 'started_contributions' to 'attached_shell'.
root INFO >>> Restoring the layout state...
root INFO Checking whether '--no-optional-locks' can be used with the current Git executable. Minimum required version is '2.15.0'.
[...]

PR #9715, opened 2021-07-08
frontend-plugin: fix creating a webworker for frontend plugin (browser only)

vitaliy-guliy:
Signed-off-by: Vitaliy Gulyy <vgulyy@redhat.com>

#### What it does

After moving to webpack 5, frontend plugins were not working for some time. Now it's not necessary to create a worker-loader.
It's enough to create a worker, pointing on javascript file.
https://webpack.js.org/guides/web-workers/

It's still not possible to create a worker for electron browser. Frontend plugins were turned off and the appropriate check for the browser was added to the `hosted-plugin.ts`.

Partially solves https://github.com/eclipse-theia/theia/issues/6601

#### How to test

- Clone and build Theia
- Create a hello-world theia frontend plugin
- Run browser version of Theia with the plugin 

#### Review checklist

#### Reminder for reviewers


packages/plugin-ext/src/hosted/browser/plugin-worker.ts
```diff
@@ -16,26 +16,41 @@
...
     constructor() {
         const emitter = new Emitter<string>();
-        this.worker = new (require('../../hosted/browser/worker/worker-main'));
-        this.worker.onmessage = message => {
-            emitter.fire(message.data);
-        };
-        this.worker.onerror = e => console.error(e);
+
+        if (this.isElectron()) {
+            console.log('PluginWorker : Frontend plugins are temporary not supported for Electron');
```
virtuakazib, line 32:
```suggestion
            console.log('PluginWorker : Frontend plugins are temporarily not supported for Electron');
```

vitaliy-guliy: @virtuakazib Thanks!

packages/plugin-ext/src/plugin/path.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import * as chai from 'chai';
+
+import { normalize, join } from './path';
+
+const expect = chai.expect;
+
+describe('Path implementation:', () => {
+
+    it('Test normalize /foo/bar//baz/asdf/quux/..', () => {
```
vince-fugnitto, line 25: minor: 

```suggestion
    it('should normalize \'/foo/bar//baz/asdf/quux/..\'', () => {
```

packages/plugin-ext/src/plugin/path.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+        const result = normalize('/foo/bar//baz/asdf/quux/..');
+        expect(result).to.be.equal('/foo/bar/baz/asdf');
+    });
+
+    it('Test normalize /foo/bar//baz/asdf/../quux/', () => {
```
vince-fugnitto, line 30: minor:

```suggestion
    it('should normalize \'/foo/bar//baz/asdf/../quux/\'', () => {
```

packages/plugin-ext/src/plugin/path.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+        const result = normalize('/foo/bar//baz/asdf/../quux/');
+        expect(result).to.be.equal('/foo/bar/baz/quux/');
+    });
+
+    it('Test join(/foo, bar, baz/asdf, quux, ..)', () => {
```
vince-fugnitto, line 35:
```suggestion
    it('should join(/foo, bar, baz/asdf, quux, ..)', () => {
```

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
```diff
@@ -16,26 +16,41 @@
...
+
     constructor() {
         const emitter = new Emitter<string>();
-        this.worker = new (require('../../hosted/browser/worker/worker-main'));
-        this.worker.onmessage = message => {
-            emitter.fire(message.data);
-        };
-        this.worker.onerror = e => console.error(e);
+
+        if (this.isElectron()) {
```
vince-fugnitto, line 31: I confirmed that the behavior works for the `browser` use-case, but if a user adds a theia frontend plugin to an electron use-case all other plugins fail to start (ex: `typescript-language-features`).

vince-fugnitto (commented):
I confirmed with the `hello-world` theia plugin that frontend plugins work again for the browser use-case :+1:

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
tsmaeder: That is a well-known and long-standing problem: https://github.com/eclipse-theia/theia/issues/6601

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
```diff
@@ -16,26 +16,41 @@
 import { injectable } from '@theia/core/shared/inversify';
```
tsmaeder, line 16: What's the reason for moving this file?

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
tsmaeder: If we want to disable loading plugins in the electron case (which is better than running into errors, I agree), we should do it in hosted-plugin.ts#startPlugins() instead of constructing a rpc endoint that doesn't do anything.

packages/plugin-ext/src/plugin/plugin-manager.ts
```diff
@@ -30,7 +30,7 @@ import {
 } from '../common/plugin-api-rpc';
 import { PluginMetadata, PluginJsonValidationContribution } from '../common/plugin-protocol';
 import * as theia from '@theia/plugin';
-import { join } from 'path';
+import { join } from './path';
```
tsmaeder, line 33: There is a file `env.ts` in this directory, but no `path.ts`

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
```diff
@@ -16,26 +16,41 @@
...
 import { Emitter } from '@theia/core/lib/common/event';
 import { RPCProtocol, RPCProtocolImpl } from '../../common/rpc-protocol';
+import { environment } from '@theia/core';
```
tsmaeder, line 19: Shouldn't we consume this as 
```
import { environment } from '@theia/core/shared/@theia/application-package/lib/environment';
```
?

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
```diff
@@ -16,26 +16,41 @@
...
-        this.worker = new (require('../../hosted/browser/worker/worker-main'));
-        this.worker.onmessage = message => {
-            emitter.fire(message.data);
-        };
-        this.worker.onerror = e => console.error(e);
+
+        if (this.isElectron()) {
+            console.log('PluginWorker : Frontend plugins are temporarily not supported for Electron');
+        } else {
+            require('./worker/worker-main');
```
tsmaeder, line 34: Why are we requiring this without using the result as before? Could you explain why you needed to change loading the worker?

tsmaeder (changes requested):

tsmaeder (dismissed):
I think we should turn off front-end plugins more cleanly in the electron case. Beyond that, I there are a couple of changes I just don't understand. It would help if you described hat the fix is a bit in the PR description.

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
vince-fugnitto: > That is a well-known and long-standing problem: #6601

@tsmaeder I know about #6601, the problem I wanted to highlight wasn't about theia frontend plugins but the fact that if a a theia frontend plugin is added in the electron application it will cause vscode plugins such as the typescript-language-features to not start. I was hopping for a more elegant way to handle the frontend theia plugins in electron as to not fail other valid plugins.

tsmaeder: Ah, so https://github.com/eclipse-theia/theia/pull/9715#discussion_r669315368 is what you want, right?

vince-fugnitto: Right, if it means not failing other valid plugins I'm all for it :)

vitaliy-guliy: Reworked

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
vitaliy-guliy: I think it's reasonable to have this file next to `hosted-plugin.ts`, where it's used.
I have a feeling, that before this file was in a wrong place.

packages/plugin-ext/src/plugin/plugin-manager.ts
vitaliy-guliy: `path.ts` is a new file with necessary `join` and `normalize` implementations (like in `path` package).
We need then to be able to run the script on the browser side.

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
vitaliy-guliy: Reworked

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
vitaliy-guliy: If the file is required somewhere, the webpack generates `examples/browser/lib/worker-ext.js` and `examples/electron/lib/worker-ext.js` files.
If we don't require './worker/worker-main', the 'worker-ext.js' is missing.

If somebody know how to force the webpack to generate .js file without adding that 'require', it would be helpful.

paul-marechal:
@vitaliy-guliy why should we avoid using `worker-loader`?

Modifying the generated Webpack configuration like you did will not work if we don't include `@theia/plugin-ext` right?

vitaliy-guliy:
> @vitaliy-guliy why should we avoid using `worker-loader`?

With Webpack 5 we can use webworker without worker loader https://webpack.js.org/guides/web-workers/

I've tried to use it with loader and create  a worker using a `require` as it [was before](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/main/browser/plugin-worker.ts#L27), but it does not work.
Nothing happens as well if I run worker-loader
```
new Worker(worker-loader script)
```

> Modifying the generated Webpack configuration like you did will not work if we don't include `@theia/plugin-ext` right?

Yes, it will fail, plugin-ext must be included. I will think how to improve it and make the build successful without plugin-ext.

dev-packages/application-manager/src/generator/webpack-generator.ts
```diff
@@ -114,7 +114,7 @@ module.exports = {
                 test: /worker-main\\.js$/,
                 loader: 'worker-loader',
                 options: {
-                    filename: 'worker-ext.[fullhash].js'
+                    filename: 'worker-ext.js'
```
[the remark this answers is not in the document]
vitaliy-guliy: Nothing is impossible, but since we don't need a worker-loader we need to keep somewhere the file name.
Before it was inside generated worker-loader script
```
  return new Worker(__webpack_require__.p + "worker-ext.5f6918e41dd2b417d716.js");
```

For now we need to keep somewhere a proper name of the file and fetch it before running a worker.

Btw, other javascript files don't have hash in their names. Do we really need to have it only for this specific file?

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
vitaliy-guliy: It was reworked. After adding a webpack configuration, the javascript file is taken from the @threia/plugin-ext directly.

vitaliy-guliy:
@paul-marechal I played with dependencies a bit, but was easily compile the browsrer/electron.
I think it's because plugin-ext is still physically present on the file system and could be found when running webpack.

paul-marechal:
> I think it's because plugin-ext is still physically present on the file system and could be found when running webpack.

So it only works in the Theia repo but will most likely fail in a downstream Theia app.

> With Webpack 5 we can use webworker without worker loader https://webpack.js.org/guides/web-workers/
> 
> I've tried to use it with loader and create a worker using a `require` as it [was before](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/main/browser/plugin-worker.ts#L27), but it does not work.
> Nothing happens as well if I run worker-loader
> 
> ```
> new Worker(worker-loader script)
> ```

I played around with what the Webpack's doc was saying: https://github.com/eclipse-theia/theia/compare/mp/frontend-plugins#diff-6ccaef7f7e0ffcd2911ccf76fd97fbe9f2e75f1009ddda5c9024d9807bb093a3R30-R31

It worked for me without the changes to the application-manager generators.

vitaliy-guliy:
> > I think it's because plugin-ext is still physically present on the file system and could be found when running webpack.
> 
> So it only works in the Theia repo but will most likely fail in a downstream Theia app.
> 
> > With Webpack 5 we can use webworker without worker loader https://webpack.js.org/guides/web-workers/
> > I've tried to use it with loader and create a worker using a `require` as it [was before](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/main/browser/plugin-worker.ts#L27), but it does not work.
> > Nothing happens as well if I run worker-loader
> > ```
> > new Worker(worker-loader script)
> > ```
> 
> I played around with what the Webpack doc was saying: https://github.com/eclipse-theia/theia/compare/mp/frontend-plugins#diff-6ccaef7f7e0ffcd2911ccf76fd97fbe9f2e75f1009ddda5c9024d9807bb093a3R30-R35
> 
> It worked for me without the changes to the application-manager generators.

How did you build it?
I always get an error, when I try to create an URL with `import.meta.url`
```
packages/plugin-ext/src/hosted/browser/plugin-worker.ts:30:52 - error TS1343: The 'import.meta' meta-property is only allowed when the '--module' option is 'esnext' or 'system'.

30         const workerURI = new URL('./worker-main', import.meta.url);
                                                      ~~~~~~~~~~~
```
That's why I used `location.href` instead.

vitaliy-guliy:
@paul-marechal I've updated the PR, based on your branch. Seems it works fine.
Thanks!

paul-marechal (approved):
LGTM!

paul-marechal:
> That's why I used `location.href` instead.
> But seems `// @ts-expect-error (TS1343)` helps to compile ignoring the error.

Did it worked with `location.href`? I only used `import.meta.url` because I thought it was a pattern Webpack is looking for to do the proper transformations, but maybe something else would be detected by Webpack? I'm fine with the current changes anyway.

vitaliy-guliy:
> > That's why I used `location.href` instead.
> > But seems `// @ts-expect-error (TS1343)` helps to compile ignoring the error.
> 
> Did it work with `location.href`? I only used `import.meta.url` because I thought it is a pattern Webpack is looking for to do the proper transformations, but maybe something else would also be detected by Webpack? I'm fine with the current changes anyway.

It works with `location.href`, but I have to create an URL referring on compiled javascript, not the typescript file.

paul-marechal:
Out of curiosity I tried the following:

```ts
this.worker = new Worker(new URL('./worker/worker-main', location.href));
```

It didn't work at all... So `import.meta.url` it is then!

PR #9719, opened 2021-07-08
Don't modify settings.json when adding debug configurations

### LINKED ISSUE #9712: Upgrade 1.14.0 is causing problem with settings.json
### Bug Description:

- Upgrade 1.14.0 is causing a problem.
- We have settings.json file stored in `/home/ide/.theia/settings.json` and it does contains IDE preferences. These settings are working in version 1.13.0 but when we try to upgrade Theia to version 1.14.0 it start causing a problem.
- When we try to open a settings.json from FILE->Settings->Open Preferences->Edit in Settings.json it does open the settings.json file and I can see the copied preferences.
- But when we try to open the same via Debug->Open launch.json it does override the `settings.json` file first. If we click once again it opens `launch.json` file.
- For a workaround I tried to add configurations in settings.json at parent node and after that it stopped overriding the settings.json file but since configurations is unknown parent property for settings.json it does create a problem in Theia problems tab.
- FYI - We don't have launch.json file by default in our IDE as we didn't needed it. I tried to add default launch.json file but still it was overriding the settings.json file.

Attached is the Video for reference.

- Broken Configurations.
- In the below video you will be able to find that debug launch.json file is overriding the settings.json file.

https://user-images.githubusercontent.com/73994076/124892610-a3071a80-dff7-11eb-95ef-b831bad87229.mov

- Workaround I tried to add configurations in settings.json on parent node. It does resolve the problem of overriding settings.json but it doesn't seems to be right solution to tackle this problem.

https://user-images.githubusercontent.com/73994076/124893166-2163bc80-dff8-11eb-97ed-b60668022986.mov

### Steps to Reproduce:

1. Create IDE using Theia version 1.14.0 and Node v12.22.1. Make sure to keep settings.json with preferences before running the IDE. Attached is the sample settings.txt file - Rename the extension and keep it under /home/ide/.theia/settings.json
[settings.txt](https://github.com/eclipse-theia/theia/files/6783162/settings.txt)

2. Keep all the preferences in settings.json.
3. Open the Debug from left panel of the IDE
4. Click on either settings icon or try to add a new configuration. Also try to refer the first video - Broken configurations

PS - This issue happen only for the first time when you open and IDE and click on Debug settings icon. Once it overrides the settings.json file and create a new launch.json file it does not repeat it.

- I feel this issue have links with the earlier bug - https://github.com/eclipse-theia/theia/issues/9310 which got fixed in version 1.14.0. We never faced this challenge in version 1.13.0.

### Additional Information

- Operating System: Mac-OS & Linux 18.04
- Theia Version: 1.14.0

colin-grant-work:
#### What it does

Fixes #9712 by checking whether the file found is a default config file (`settings.json`) before modifying it when adding debug configurations.

#### How to test

1. In an otherwise empty folder, create a `.theia` folder and populate it with a `settings.json` that includes a `"launch" field. Something like this:
```
{
  "launch": {
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    "version": "0.2.0",
    "configurations": []
  },
  "editor.fontSize": 20
}
```
2. Open that folder in Theia.
3. Open the debug view, open the dropdown and select 'Add configurations'
4. It should open your `settings.json` and put you in the middle of the array, but it should _not_ remove your "editor.fontSize" setting.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-configuration-manager.ts
+++ b/packages/debug/src/browser/debug-configuration-manager.ts
@@ -265,7 +265,11 @@ export class DebugConfigurationManager {
         if (!uri) { // Since we are requesting information about a known workspace folder, this should never happen.
             throw new Error('PreferenceService.getConfigUri has returned undefined when a URI was expected.');
         }
-        await this.ensureContent(uri, model);
+        const settingsUri = this.preferences.getConfigUri(PreferenceScope.Folder, model.workspaceFolderUri);
+        // Users may have placed their debug configurations in a `settings.json`, in which case we shouldn't modify the file.
+        if (settingsUri && !uri.isEqual(settingsUri)) {
+            await this.ensureContent(uri, model);
+        }
         return uri;
     }
 


vince-fugnitto (approved):
I confirmed that it fixes the reported issue https://github.com/eclipse-theia/theia/issues/9712 :+1:

PR #8731, opened 2020-11-06
preferences:  Support 'Close On File Delete' preference

DukeNgn:
#### What it does
Fixes #8730 
+ Changes `deletedSuffix` to `(deleted)`.
+ Adds new preference `workbench.editor.closeOnFileDelete` (`false` by default) to control whether the editor should be closed if other processes deleted or renamed the file.

#### How to test
+ Start the application and open a file (ex: readme.md).
+ Delete the file on the filesystem.

**Observe**:
**Note**: Regardless of the setting, dirty editors should not be closed when other processes `deleted` or `renamed` the file.

+ If `closeOnFileDelete` is true => the editor is closed.
+ If `closeOnFileDelete` is false => the editor is kept alive with the change in the title (`readme.md (deleted)`)
image

#### Review checklist

- [ ] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

Signed-off-by: Duc Nguyen <duc.a.nguyen@ericsson.com>

PR #9720, opened 2021-07-08
Modify closeOnFileDelete default to agree with VSCode

colin-grant-work:
#### What it does

Modifies the default value of `workbench.editor.closeOnFileDelete` to agree with VSCode. We have it set to `true`, they have it set to `false`. `false` is a better default, I think - closing fewer editors automatically is safer.

https://github.com/microsoft/vscode/blob/40b2219e394fd4721934b4007ae0b2890b6b086a/src/vs/workbench/browser/workbench.contribution.ts#L163-L167

#### How to test

1. Open a workspace with no modifications to the `workbench.editor.closeOnFileDelete` preference.
2. Check the preferences UI and observe that the preference is set to false, and that setting it to true gives you a blue modified border.
3. Open a file, then delete it, and observe that the editor doesn't close, but does mark itself as `(deleted)`.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
The changes look good to me, I believe it was a typo to have it `true` by default as the original pull-request mentioned `false` was the default (https://github.com/eclipse-theia/theia/pull/8731).

The disabling of the preference also makes sense to me, and aligns with vscode as files deleted outside the application are no longer automatically closed within the application :+1:

Issue #8557, opened 2020-09-24
Support deprecated strikethrough directly in the editor


DukeNgn:
### Feature Description:
Would be nice if Theia highlights deprecated calls with strikethrough in the editor

+ Similar to VSCode:
![strikethrough](https://user-images.githubusercontent.com/43587865/94155516-064f0400-fe4d-11ea-86cd-5edf4490d778.png)

vince-fugnitto:
I believe it will require an update to **monaco v0.21.0** in order to make use of the `showDeprecated` flag:

https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.ieditoroptions.html#showdeprecated

vince-fugnitto:
The following should now be possible with the latest monaco upgrade.
The preference `editor.showDeprecated` is not yet supported at the moment (it does not display deprecations in the editor as is).

https://github.com/eclipse-theia/theia/blob/f5d042f00eaf973d4c3156d161ab1252b616b257/packages/editor/src/browser/editor-preferences.ts#L1078-L1082

PR #9721, opened 2021-07-09
plugin: add `showDeprecated` support

vince-fugnitto:
#### What it does

Fixes: #8557 

The pull-request adds support for the `editor.showDeprecated` preference which styles deprecated code in editors with a `strike-through`. The diagnostic is contributed by language-servers:

image

#### How to test

1. update the [vscode-builtin-typescript-language-features](https://open-vsx.org/api/vscode/typescript-language-features/1.57.1/file/vscode.typescript-language-features-1.57.1.vsix) extension to `1.57.1` in the `package.json`.
2. start the application with the changes, and use `theia` as a workspace folder.
3. open `keymaps-service.ts` and navigate to line `140`
4. uses of `context` (deprecated) should have the deprecated strike-through.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/protocol-to-monaco-converter.ts
+++ b/packages/monaco/src/browser/protocol-to-monaco-converter.ts
@@ -131,7 +131,8 @@ export class ProtocolToMonacoConverter {
             startColumn: diagnostic.range.start.character + 1,
             endLineNumber: diagnostic.range.end.line + 1,
             endColumn: diagnostic.range.end.character + 1,
-            relatedInformation: this.asRelatedInformations(diagnostic.relatedInformation)
+            relatedInformation: this.asRelatedInformations(diagnostic.relatedInformation),
+            tags: diagnostic.tags
         };
     }
 
--- a/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
+++ b/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
@@ -219,6 +219,7 @@ export enum MarkerSeverity {
 
 export enum MarkerTag {
     Unnecessary = 1,
+    Deprecated = 2,
 }
 
 export interface ParameterInformation {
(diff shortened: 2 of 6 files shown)


alvsan09 (approved):
I can confirm that it works as expected, there is some latency to apply the markings but that is something to investigate separately.  :+1:

image

packages/monaco/src/browser/protocol-to-monaco-converter.ts
```diff
@@ -131,7 +131,8 @@ export class ProtocolToMonacoConverter {
             startColumn: diagnostic.range.start.character + 1,
             endLineNumber: diagnostic.range.end.line + 1,
             endColumn: diagnostic.range.end.character + 1,
-            relatedInformation: this.asRelatedInformations(diagnostic.relatedInformation)
+            relatedInformation: this.asRelatedInformations(diagnostic.relatedInformation),
+            tags: diagnostic.tags
```
tsmaeder, line 135: Looks like we did not handle tags _at all_ before. Is there an instance of a "unnecessary" tag we could use to test?

packages/plugin-ext/src/plugin/languages/diagnostics.ts
```diff
@@ -223,7 +223,8 @@ export class DiagnosticCollection implements theia.DiagnosticCollection {
                                         startLineNumber: lastMarker.startLineNumber,
                                         startColumn: lastMarker.startColumn,
                                         endLineNumber: lastMarker.endLineNumber,
-                                        endColumn: lastMarker.endColumn
+                                        endColumn: lastMarker.endColumn,
+                                        tags: lastMarker.tags,
```
tsmaeder, line 227: This is an artificial marker. I don't think we should don't think we should apply the tags for the marker it replaces.

packages/monaco/src/browser/protocol-to-monaco-converter.ts
msujew: I think @alvsan09 tested it already in his comment. See the `myUndefined` variable and its 'fading' effect:

image

packages/plugin-ext/src/plugin/languages/diagnostics.ts
vince-fugnitto: @tsmaeder thank you for clarifying it's use-case!

tsmaeder (approved):

Issue #9477, opened 2021-05-13
file-search: support 'go to symbol' directly when searching for files


vince-fugnitto:
#### Feature Description:

Related https://github.com/eclipse-theia/theia/issues/8878.

The goal is to support performing a `go to symbol` directly when searching for a file in the `quick-file-open`.

https://user-images.githubusercontent.com/40359487/118142029-fc781200-b3d7-11eb-880a-c73a8db25ea1.mp4

PR #9727, opened 2021-07-13
Refactor to hide VS Code interfaces behind core services

tsmaeder:
Signed-off-by: Thomas Mäder <tmader@redhat.com>

#### What it does
This PR refactors the `QuickInput` and `QuickAccess` services introduced in https://github.com/eclipse-theia/theia/pull/9154 in order to completely hide the types imported from VS Code. 
In particular, this allows to write the all the related contributions in terms of a single implementation of quick access and quick input. Also, the quick access registry implementation allows to use dependency injection in the regular Theia way.
This refactoring was deferred from the original PR in order to facilitate merging it earlier.

#### How to test
We should do the same testing as for https://github.com/eclipse-theia/theia/pull/9154

#### Review checklist

#### Reminder for reviewers


examples/api-tests/src/file-search.spec.js
```diff
@@ -14,6 +14,8 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
+const { QuickPickItem } = require('@theia/core/lib/browser');
```
vince-fugnitto, line 17: @tsmaeder I believe it should be inside the `describe('file-search')`, line 26.

packages/core/src/browser/quick-input/quick-access.ts
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * A context key that will be set automatically when this quick access is being shown
+     */
+    readonly contextKey?: string;
+}
+
+export const QuickAccessRegistry = Symbol('QuickAccessRegistry');
+
+/**
+ * A registry for quick access providers.
```
vince-fugnitto, line 68: There are trailing spaces errors here which fail the build.

packages/file-search/src/browser/quick-file-open.ts
```diff
@@ -27,24 +27,26 @@ import * as fuzzy from '@theia/core/shared/fuzzy';
...
     range?: Range;
 }
 
 // Supports patterns of <path><#|:><line><#|:|,><col?>
 const LINE_COLON_PATTERN = /\s?[#:\(](?:line )?(\d*)(?:[#:,](\d*))?\)?\s*$/;
+type FileQuickPickItem = QuickPickItem & { uri: URI };
 
 @injectable()
-export class QuickFileOpenService implements monaco.quickInput.IQuickAccessDataService {
+export class QuickFileOpenService implements QuickAccessProvider {
```
alvsan09, line 47: I am contemplating the possibility to support file search by text and symbol e.g. 'filex@functiony'
and similarly to `vscode` select a file via the drop down and follow the selection with the `@` symbol.
 
In order to achieve that we would need to use a `provider` interface like `IQuickAccessProvider` or extend `PickerQuickAccessProvider` so we can get a handle to the picker instance `IQuickPickItem` e.g. from the method `provide`. 
Having access to the picker can then make it possible to keep track of the `active` items i.e. the selected item entries in the drop down and therefore be the foundation to navigate to e.g. `filex@symboly`.

What do you think about replacing this interface for one of the options mentioned above?

tsmaeder: The point of this PR is to hide the implementation details (aka everything imported from VS Code), so no, exposing IQuickAccessProvider or other VS Code API is a no-no, IMO. But nothing prevents us from exposing more functionality in the QuickInput/QuickAccess interfaces. Why don't you open an issue describing the feature?

packages/core/src/browser/quick-input/quick-access.ts
```diff
@@ -0,0 +1,75 @@
+/********************************************************************************
+ * Copyright (c) 2021 SAP SE or an SAP affiliate company and others.
```
RomanNikitenko, line 2: RedHat?
or it was moved from another place?

tsmaeder: Yes, this started as "quick-access-contribution.ts". Let's not start a copyright war here.

RomanNikitenko: I didn't aware about `quick-access-contribution.ts`-  asked just in case - maybe you just missed that...  
anyway - it's not starting a copyright war :-)

tsmaeder: It's just that I had the same discussion with the SAP folks on the original monaco update ("don't change copyright on existing/renamed files"). :-)

packages/core/src/browser/quick-input/quick-command-service.ts
```diff
@@ -42,12 +45,93 @@ export class QuickCommandService extends QuickInputService implements QuickAcces
     @inject(CorePreferences)
     protected readonly corePreferences: CorePreferences;
 
+    @inject(QuickAccessRegistry)
+    protected readonly quickAccessRegistry: QuickAccessRegistry;
+
+    constructor(@inject(KeybindingRegistry) protected readonly keybindingRegistry: KeybindingRegistry) {
+    }
```
RomanNikitenko, line 52: do we really need an empty constructor?
can we use [the property injection](https://github.com/eclipse-theia/theia/wiki/Coding-Guidelines#dependency-injection) here instead?

packages/file-search/src/browser/quick-file-open.ts
alvsan09: It's fine if we handle the whole feature in a separate PR.
The issue describing the feature already exists, see:
https://github.com/eclipse-theia/theia/issues/9477

tsmaeder: I believe what you want is already possible: see how the quick pick is manipulated in plugin-ext-deploy-command.ts

packages/monaco/src/browser/monaco-quick-input-service.ts
```diff
@@ -15,138 +15,175 @@
...
-    get onShow(): Event<void> { return this.controller.onShow; }
-    get onHide(): Event<void> { return this.controller.onHide; }
+    @inject(KeybindingRegistry)
+    protected readonly keybindingRegistry: KeybindingRegistry;
 
     constructor() {
-        super();
-        this.initContainer();
-        this.initController();
+        this.monacoService = new MonacoQuickInputImplementation(this.contextKeyService);
```
alvsan09, line 37: Shouldn't we use inversify to inject the instance instead?

packages/file-search/src/browser/quick-file-open.ts
alvsan09: The reference you provided is nice, although it does not apply to `quick-file-open` where we handle PREFIX='',
for this case and looking at `vscode` as a reference, the implementation relies on registering a provider instance of `IQuickAccessProvider` via `AnythingQuickAccessProvider` so the `provide` method may nicely override the one in the parent `PickerQuickAccessProvider` and therefore gain access to the picker, and when calling  `super.provide` which reaches `PickerQuickAccessProvider` it will then call `getPicks` which can be overridden as well to provide the customized behavior..  

We can leave this refactoring for later although it would be great if we can get your expertise in the area to prepare the foundation for it..

packages/monaco/src/browser/monaco-quick-input-service.ts
paul-marechal: If so you'd need to bind a factory to wrap the call to `new MonacoQuickInputImplementation`.

tsmaeder: The idea was that this is an implementation detail of the MonacoQuickInputService, but since we're using the same class over in MonacoEditorProvider, that is not true anyway.
The VS Code code can handle multiple instances no problem, but we don't need more than one instance, so using the one from the container seems the right thing.

packages/core/src/browser/quick-input/quick-command-service.ts
tsmaeder: No we don't.

tsmaeder:
Addressed review comments and rebased. Now (desperately ;-)) seeking for approval. 🥺

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -762,7 +759,7 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
         });
 
         commandRegistry.registerCommand(CommonCommands.OPEN_VIEW, {
-            execute: () => this.quickInputService?.open(this.quickView?.prefix)
+            execute: () => this.quickInputService?.open(QuickViewService.PREFIX)
```
vince-fugnitto, line 762: The following command does not work for me, likely related to changes in `quick-view-service`.
Triggering the command prompts you to choose a view from the list, but upon selection the view is never opened.

packages/core/src/browser/quick-input/quick-command-service.ts
```diff
@@ -31,7 +33,8 @@ export const CLEAR_COMMAND_HISTORY: Command = {
 };
 
 @injectable()
-export class QuickCommandService extends QuickInputService implements QuickAccessContribution {
+export class QuickCommandService implements QuickAccessContribution, QuickAccessProvider {
+    static PREFIX = '>';
```
vince-fugnitto, line 37: Changes in this file do not seem to work correctly.
The list of recently used commands are not preserved when re-triggering the quick-command.

The list of recently used commands only appear upon reloading the application, performing `clear command history` also does not update the list until the app is reloaded.

packages/core/src/browser/quick-input/quick-access.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export interface QuickAccessProviderHelp {
+    prefix?: string;
+    description: string;
+    needsEditor: boolean;
+}
+
+export interface QuickAccessProviderDescriptor {
+    /**
+     * return an instance of QuickAccssProvider. Implementers are free to return that same instance multiple times
```
vince-fugnitto, line 42:
```suggestion
     * return an instance of QuickAccessProvider. Implementers are free to return that same instance multiple times
```

packages/core/src/browser/quick-input/quick-command-service.ts
msujew: I can't reproduce this issue in the browser (Firefox) environment:

https://user-images.githubusercontent.com/4377073/127165072-d4797bde-4b8d-4ef4-8b97-af0caf3a0c88.mp4

vince-fugnitto: @msujew it does not work for me in the browser (chrome + firefox), and electron. The commands are not present in the recently used list until the app is reloaded.

packages/core/src/browser/common-frontend-contribution.ts
tsmaeder: I don't think it's the command: it also doesn't work when typing "view " in the `F1` quick access directly.

packages/core/src/browser/quick-input/quick-command-service.ts
tsmaeder: This seems to work just fine for me, as well (both the "recently used" commands as well as the "clear history" part. Could you maybe give more detailed steps to reproduce?

packages/core/src/browser/common-frontend-contribution.ts
tsmaeder: Found and fixed.

packages/core/src/browser/quick-input/quick-command-service.ts
vince-fugnitto: @tsmaeder I'm using Ubuntu. it's likely its my environment since Mark, Alvaro and yourself all don't experience the issue while I do even when building multiple times. 

If interested, to reproduce I do the following to have a clean environment:
- perform `rm -rf ~/.theia` && `rm -rf ~/.config/Theia\ Electron\ Example` (to remove any config or cache)
- start the electron app
- execute `reset workbench layout`
- execute `clear command history`
- close the app through the window `x`
- restart the app
- perform the tests, see if executing commands adds them to the recently used list and so on

Note that prior to the changes the recently used list did work for me.

vince-fugnitto: Just to mention, I built on a completely different machine (macos) and the problem persists for me.

tsmaeder: I don't know what to do here: this works perfectly for me and I've inspected the code and my changes just don't look like they could cause the behaviour you mention. Could you give the exact steps you're doing, including

1. The name of the commands you're invoking
2. How you're invoking them (menu, etc.)
3. Open editors, views, etc?

/Thomas

vince-fugnitto: @tsmaeder I'm not sure either, I've rebuilt the pull-request a couple of times and on different machines but I can retry:
1. the commands are mainly random, only to see if they are populated in the recent list and they only are after reload.
2. invoking the commands through the `command palette` itself.
3. the view is the default since I performed `reset workbench layout` to have a clean reproducible state.

tsmaeder: And by `command palette`, you mean the "quick access" invoked by hitting F1?

tsmaeder: Last resort: make a video 🤷 ?

vince-fugnitto: @tsmaeder yes correct, the quick-command-service through <kbd>F1</kbd>. 
I managed to do a fresh clone in a new directory, and rebuild your changes and they seem to be working fine now, I'm really sorry about that.

tsmaeder: Javascript build systems for teh win! Personally, I like `git clean -xfd` ;-)

vince-fugnitto: I always do it as part of my workflow so not really sure what happened (the fact it even happened on other machines as well). Anyways please proceed :)

vince-fugnitto (commented):
@tsmaeder do you mind rebasing the pull-request, I'd be happy to perform another review so we can benefit from the cleanup following the upgrade :+1:

packages/core/src/browser/quick-input/quick-view-service.ts
```diff
@@ -14,24 +14,74 @@
...
+        const items = this.items.filter(item =>
+            (item.when === undefined || this.contextKexService.match(item.when)) &&
+            (!this.hiddenItemLabels.has(item.label))
+        );
+        return filterItems(items, filter);
     }
-    hideItem(label: string): void { }
-    showItem(label: string): void { }
+
+    readonly prefix = 'view ';
```
vince-fugnitto, line 85: Leftover? We have `QuickViewService.PREFIX`.

vince-fugnitto (approved):
The changes look good to me, I verified the different quick-input menus we have (seen with the `?` prefix).

msujew (approved):
Looks good to me as well:

* Every quick input prefix is working correctly ✔️
* Command history is working correctly ✔️ 
* Quick input icons are correctly displayed ✔️

PR #9731, opened 2021-07-13
Implement http fallback for unavailable websockets

### LINKED ISSUE #3053: Question: Using Theia behind a corporate proxy that blocks websocket
Hi,
I just wanted to know if if there is a way to use theia without websocket access. Our corporate proxy blocks all websockets and that really s*****. Maybe some one else is in the same position. HTTP/S tunneling would be possible over the proxy... but no clue how to achive that without local admin rights :-)
Any input welcome.
I am coming from ice ide, and would like to switch to theia... but that isnt possible so far.

msujew:
#### What it does
Closes #3053

Adds a polling based http fallback mode for contexts in which websockets aren't available like behind certain firewall or proxy configurations. By default this fallback mode is activated once the websocket connection fails two times in a row.

This basically simulates a websocket connection via http: Frontend messages are pushed to the backend on request and backend messages are retrieved by the frontend with a polling mechanism that is able to receive multiple messages with a single request.

I noticed that the file upload does not work, since that doesn't use the default messaging service and instead creates its own websocket connection. We could fix that in another PR, since its quite independent of my changes. Everything else should behave as usual.

I also thought about using server-sent events instead of the naive long polling. WDYT?

#### How to test

All steps after 5 require you to run Theia on a remote machine (like Gitpod).

1. Open Firefox and visit the `about:config` page.
2. Set the `network.websocket.max-connections` setting to `0`.
3. Start Theia and observe two websocket error messages (in the console).
4. Afterwards the application should start and display a warning, indicating it is running in the fallback mode.
5. The application should run and behave as usual. (manual smoke tests)
6. Disconnect from the internet.
7. The status bar should indicate the offline status.
8. Upon reconnecting to the internet the status should be back to online and the application should work again.

Let me know if you find other ways to nicely simulate the unavailability of websockets.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/messaging/ws-connection-provider.ts
+++ b/packages/core/src/browser/messaging/ws-connection-provider.ts
@@ -14,12 +14,13 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-import { injectable, interfaces, decorate, unmanaged } from 'inversify';
-import { JsonRpcProxyFactory, JsonRpcProxy, Emitter, Event } from '../../common';
+import { injectable, interfaces, decorate, unmanaged, inject } from 'inversify';
+import { JsonRpcProxyFactory, JsonRpcProxy, Emitter, Event, MessageService, MessageServiceFactory } from '../../common';
 import { WebSocketChannel } from '../../common/messaging/web-socket-channel';
 import { Endpoint } from '../endpoint';
 import ReconnectingWebSocket from 'reconnecting-websocket';
 import { AbstractConnectionProvider } from '../../common/messaging/abstract-connection-provider';
+import { v4 as uuid } from 'uuid';
 
 decorate(injectable(), JsonRpcProxyFactory);
 decorate(unmanaged(), JsonRpcProxyFactory, 0);
@@ -31,6 +32,29 @@ export interface WebSocketOptions {
     reconnecting?: boolean;
 }
 
+export const HttpFallbackOptions = Symbol('HttpFallbackOptions');
+
+export interface HttpFallbackOptions {
+    /** Determines whether Theia is allowed to use the http fallback. True by default. */
+    allowed: boolean;
+    /** Number of failed websocket connection attempts before the fallback is triggered. 2 by default. */
+    maxAttempts: number;
+    /** The maximum duration (in ms) after which the http request should timeout. 5000 by default. */
+    pollingTimeout: number;
+    /** The timeout duration (in ms) after a request was answered with an error code. 5000 by default. */
+    errorTimeout: number;
+    /** The minimum timeout duration (in ms) between two http requests. 0 by default. */
+    requestTimeout: number;
+}
+
+export const DEFAULT_HTTP_FALLBACK_OPTIONS: HttpFallbackOptions = {
+    allowed: true,
+    maxAttempts: 2,
+    errorTimeout: 5000,
+    pollingTimeout: 5000,
+    requestTimeout: 0
+};
+
 @injectable()
 export class WebSocketConnectionProvider extends AbstractConnectionProvider<WebSocketOptions> {
 
@@ -40,17 +64,27 @@ export class WebSocketConnectionProvider extends AbstractConnectionProvider<WebS
     protected readonly onSocketDidCloseEmitter: Emitter<void> = new Emitter();
     readonly onSocketDidClose: Event<void> = this.onSocketDidCloseEmitter.event;
 
+    protected readonly onHttpFallbackDidActivateEmitter: Emitter<void> = new Emitter();
+    readonly onHttpFallbackDidActivate: Event<void> = this.onHttpFallbackDidActivateEmitter.event;
+
     static createProxy<T extends object>(container: interfaces.Container, path: string, arg?: object): JsonRpcProxy<T> {
         return container.get(WebSocketConnectionProvider).createProxy<T>(path, arg);
     }
 
+    @inject(MessageServiceFactory) protected readonly messageService: () => MessageService;
+    @inject(HttpFallbackOptions) protected readonly httpFallbackOptions: HttpFallbackOptions;
+
     protected readonly socket: ReconnectingWebSocket;
+    protected useHttpFallback = false;
+    protected websocketErrorCounter = 0;
+    protected httpFallbackId = uuid();
+    protected httpFallbackDisconnected = true;
 
     constructor() {
         super();
         const url = this.createWebSocketUrl(WebSocketChannel.wsPath);
         const socket = this.createWebSocket(url);
-        socket.onerror = console.error;
+        socket.onerror = event => this.handleSocketError(event);
         socket.onopen = () => {
             this.fireSocketDidOpen();
         };
@@ -61,6 +95,7 @@ export class WebSocketConnectionProvider extends AbstractConnectionProvider<WebS
             this.fireSocketDidClose();
         };
         socket.onmessage = ({ data }) => {
+            this.websocketErrorCounter = 0;
             this.handleIncomingRawMessage(data);
         };
         this.socket = socket;
@@ -68,8 +103,62 @@ export class WebSocketConnectionProvider extends AbstractConnectionProvider<WebS
         window.addEventListener('online', () => this.tryReconnect());
     }
 
+    handleSocketError(event: unknown): void {
+        this.websocketErrorCounter += 1;
+        if (this.httpFallbackOptions.allowed && this.websocketErrorCounter >= this.httpFallbackOptions.maxAttempts) {
+            this.useHttpFallback = true;
+            this.socket.close();
+            const httpUrl = this.createHttpWebSocketUrl(WebSocketChannel.wsPath);
+            this.onHttpFallbackDidActivateEmitter.fire(undefined);
+            this.doLongPolling(httpUrl);
+            this.messageService().warn(
+                'Could not establish a websocket connection. The application will be using the HTTP fallback mode. This may affect performance and the behavior of some features.'
+            );
+        }
+        console.error(event);
+    }
+
+    async doLongPolling(url: string): Promise<void> {
+        let timeoutDuration = this.httpFallbackOptions.requestTimeout;
+        const controller = new AbortController();
+        const pollingId = window.setTimeout(() => controller.abort(), this.httpFallbackOptions.pollingTimeout);
+        try {
+            const response = await fetch(url, {
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/json'
+                },
+                signal: controller.signal,
+                keepalive: true,
+                body: JSON.stringify({ id: this.httpFallbackId, polling: true })
+            });
+            if (response.status === 200) {
+                window.clearTimeout(pollingId);
+                if (this.httpFallbackDisconnected) {
+                    this.fireSocketDidOpen();
+                }
+                const json: string[] = await response.json();
+                if (Array.isArray(json)) {
+                    for (const item of json) {
+                        this.handleIncomingRawMessage(item);
+                    }
+                } else {
+                    throw new Error('Received invalid long polling response.');
+                }
+            } else {
+                timeoutDuration = this.httpFallbackOptions.errorTimeout;
+                this.httpFallbackDisconnected = true;
+                this.fireSocketDidClose();
+                throw new Error('Response has error code: ' + response.status);
+            }
+        } catch (e) {
+            console.error('Error occured during long polling', e);
+        }
+        setTimeout(() => this.doLongPolling(url), timeoutDuration);
+    }
+
     openChannel(path: string, handler: (channel: WebSocketChannel) => void, options?: WebSocketOptions): void {
-        if (this.socket.readyState === WebSocket.OPEN) {
+        if (this.useHttpFallback || this.socket.readyState === WebSocket.OPEN) {
             super.openChannel(path, handler, options);
         } else {
             const openChannel = () => {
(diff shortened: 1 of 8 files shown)


paul-marechal:
> I noticed that the file upload does not work, since that doesn't use the default messaging service and instead creates its own websocket connection. We could fix that in another PR, since its quite independent of my changes.

I'll work on making the file upload functionality always use bare HTTP (PUT), is that fine with you? I am looking at improvements to the messaging infrastructure to make it easier to deal with, this would be part of that.

msujew:
> I'll work on making the file upload functionality always use bare HTTP (PUT), is that fine with you?

Sure, sounds very good to me!

> I am looking at improvements to the messaging infrastructure to make it easier to deal with

I thought about implementing an abstraction layer between the `MessagingContribution` and the web sockets, but decided to create a `ws` compatible http API instead (namely the `HttpFallbackAdapter` class). WDYT about that? Would that help your plans for improving the infrastructure?

paul-marechal:
> I thought about implementing an abstraction layer between the `MessagingContribution` and the web sockets [...]

This is precisely what I had in mind. The HTTP polling would be just one implementation in that middle layer, next to the WebSockets rather than within it as a fallback.

I experimented with that idea a bit already. I can share a very-wip draft branch on Monday that should illustrate things.

Until I clean it up a bit, what I had looked like this:

- `Connection` objects representing bidirectional _message streams_.
- `ConnectionServer` singleton at the middle of everything on the backend (maybe should just be the messaging service).
- `ConnectionServerContribution` that handle connections (WebSocket, HTTP long poll, etc) and call `ConnectionServer.accept(openConnectionParams)` to know if a `ConnectionHandler` can accept the incoming `Connection` or refuse it otherwise.
- `ConnectionProvider` is a `Connection` factory in the frontend that dispatches to either `WebSocketConnectionProvider` to connect via WebSockets, or `HTTPLongPollingConnectionProvider` in case the former fails to establish a connection too many times.
- I am missing adapters to hook a `Connection` to a `vscode-jsonrpc.MessageConnection` but the goal is to have one.
- Multiplexing/channel system that implements the `Connection` interface while working over a single `Connection`, making the channels independent from the WebSockets (it is currently quite coupled).

Some things will be breaking, but I think that's for the best. Maybe this would be better tracked as its own issue?

I trust you that the current change works, but I definitely would like to rework things to remove ad-hoc implementations. It could be a following PR, meaning it doesn't have to prevent merging this one.

packages/core/src/browser/messaging/ws-connection-provider.ts
```diff
@@ -31,6 +32,19 @@ export interface WebSocketOptions {
     reconnecting?: boolean;
 }
 
+export interface HttpFallbackOptions {
+    /** Determines whether Theia is allowed to use the http fallback. True by default. */
+    allowed: boolean;
+    /** Amount of failed websocket connections attempts before the fallback is triggered. 2 by default. */
```
colin-grant-work, line 38:
```suggestion
    /** Number of failed websocket connection attempts before the fallback is triggered. 2 by default. */
```

packages/core/src/browser/messaging/ws-connection-provider.ts
```diff
@@ -31,6 +32,19 @@ export interface WebSocketOptions {
...
+    trigger: number;
```
colin-grant-work, line 39: This name is a bit odd. Maybe `maxAttempts` or `maxErrors` or `maxFailures`?

packages/core/src/browser/messaging/ws-connection-provider.ts
```diff
@@ -40,17 +54,33 @@ export class WebSocketConnectionProvider extends AbstractConnectionProvider<WebS
...
+
     static createProxy<T extends object>(container: interfaces.Container, path: string, arg?: object): JsonRpcProxy<T> {
         return container.get(WebSocketConnectionProvider).createProxy<T>(path, arg);
     }
 
+    @inject(MessageServiceFactory) protected readonly messageService: () => MessageService;
+
     protected readonly socket: ReconnectingWebSocket;
+    protected useHttpFallback = false;
+    protected httpFallbackCounter = 0;
```
colin-grant-work, line 68:
```suggestion
    protected websocketErrorCounter = 0;
```

packages/core/src/browser/messaging/ws-connection-provider.ts
```diff
@@ -61,28 +91,93 @@ export class WebSocketConnectionProvider extends AbstractConnectionProvider<WebS
...
+                        this.handleIncomingRawMessage(item);
+                    }
+                } else {
+                    throw new Error('Received invalid long polling response.');
+                }
+            } else if (response.status >= 400) {
+                timeoutDuration = this.httpFallbackOptions.errorTimeout;
+                this.httpFallbackDisconnected = true;
+                this.fireSocketDidClose();
+                throw new Error('Response has no valid status code: ' + response.status);
```
colin-grant-work, line 148:
```suggestion
                throw new Error('Response has error code: ' + response.status);
```

'no valid status code' / 'invalid status code' makes it sound like it's not adhering to HTTP status codes, rather than that it sent a code that signifies an error.

packages/core/src/browser/messaging/ws-connection-provider.ts
```diff
@@ -61,28 +91,93 @@ export class WebSocketConnectionProvider extends AbstractConnectionProvider<WebS
...
+            this.onHttpFallbackDidActivateEmitter.fire(undefined);
+            this.longPolling(httpUrl);
+            this.messageService().warn(
+                'Could not establish a websocket connection. The application will be using the HTTP fallback mode. This may affect performance and the behavior of some features.'
+            );
+        }
+        console.error(event);
+    }
+
+    async longPolling(url: string): Promise<void> {
```
colin-grant-work, line 117:
```suggestion
    async initiatePolling(url: string): Promise<void> {
```
Or some other verb?

packages/core/src/node/messaging/http-websocket-adapter.ts
```diff
@@ -0,0 +1,83 @@
...
+                this.pendingMessagesResolve = undefined;
+                resolve(messages);
+                this.pendingMessages = [];
+            } else {
+                this.pendingMessagesResolve = resolve;
+                this.pendingTimeout = setTimeout(() => {
+                    resolve([]);
+                    this.pendingMessagesResolve = undefined;
+                    this.pendingTimeout = undefined;
+                }, 4000);
```
colin-grant-work, line 48: Perhaps make this timeout configurable with options passed to a factory?

packages/core/src/node/messaging/http-websocket-adapter.ts
```diff
@@ -0,0 +1,83 @@
...
+    readyState: number = ws.OPEN;
+    alive: boolean = true;
+
+    protected pendingTimeout: NodeJS.Timeout | undefined;
+    protected pendingMessages: unknown[] = [];
+    protected pendingMessagesResolve?: (messages: unknown[]) => void;
+
+    getPendingMessages(): Promise<unknown[]> {
+        this.alive = true;
+        return new Promise(resolve => {
```
colin-grant-work, line 36: I might use a `Deferred` here, rather than creating a `Promise` and conditionally assigning its `render` callback to the class. The effect would be the same, but I think the intent might be clearer.

packages/core/src/node/messaging/http-websocket-adapter.ts
```diff
@@ -0,0 +1,83 @@
...
+
+@injectable()
+export class HttpWebsocketAdapter {
+
+    readyState: number = ws.OPEN;
+    alive: boolean = true;
+
+    protected pendingTimeout: NodeJS.Timeout | undefined;
+    protected pendingMessages: unknown[] = [];
+    protected pendingMessagesResolve?: (messages: unknown[]) => void;
```
colin-grant-work, line 32: If you go with a `Deferred`, then something like `DeferredMessageHandler` might work here. If you stick with `resolve` approach, go with a verb here: `resolvePendingMessages`.

packages/core/src/node/messaging/http-websocket-adapter.ts
```diff
@@ -0,0 +1,83 @@
...
+                    this.pendingMessagesResolve = undefined;
+                    this.pendingTimeout = undefined;
+                }, 4000);
+            }
+        });
+    }
+
+    onerror: (event: { error: any, message: string, type: string, target: WebSocket }) => void;
+    onclose: (event: { wasClean: boolean; code: number; reason: string; target: WebSocket }) => void;
+    onmessage: (data: string) => void;
```
colin-grant-work, line 55: Are these guaranteed to be assigned? I see below that they get assigned if people subscribe to those events, but ensuring that that actually happens seems to be taken care of elsewhere?

packages/core/src/node/messaging/messaging-contribution.ts
```diff
@@ -115,9 +122,47 @@ export class MessagingContribution implements BackendApplicationContribution, Me
                 socket.alive = false;
                 socket.ping();
             });
+            this.httpWebsocketAdapters.forEach((adapter, id) => {
+                if (adapter.alive === false) {
+                    adapter.onclose({ code: 404, reason: 'Connection was closed', target: adapter as unknown as WebSocket, wasClean: true });
```
colin-grant-work, line 127: I'm a bit confused by the typing here. Is there a downside to modifying the relevant declarations to `WebSocket / ws | HttpWebsocketAdapter`?

colin-grant-work (commented):
I haven't tested the functionality, but the idea looks reasonable. I have some comments mainly about naming and typing.

packages/core/src/browser/messaging/ws-connection-provider.ts
msujew: I used `doLongPolling` now, since every call to this method only sets up a single long polling call which repeats itself using a timeout.

packages/core/src/node/messaging/http-websocket-adapter.ts
msujew: Thanks for the hint in the `Deferred` direction. The code looks way cleaner now. I had to add state to the `Deferred` class though, to find out whether the promise is already resolved. Let me now if you find a better solution to that.

packages/core/src/node/messaging/http-websocket-adapter.ts
```diff
@@ -0,0 +1,84 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
colin-grant-work, line 17: Minor: You've mixed `any` and `unknown` in this file. It looks like you could use `unknown` for all remaining `any`s except line 74, and disable the lint rule for that line only.

packages/core/src/node/messaging/messaging-backend-module.ts
```diff
@@ -34,4 +35,7 @@ export const messagingBackendModule = new ContainerModule(bind => {
     bind(BackendApplicationContribution).toService(MessagingContribution);
     bind(MessagingListener).toSelf().inSingletonScope();
     bindContributionProvider(bind, MessagingListenerContribution);
+    bind(HttpWebsocketAdapter).toSelf().inRequestScope();
```
colin-grant-work, line 38: Just a question: why `inRequestScope()` here? Do you anticipate more than one `@inject`ion of this class during a single `.get()` for a given container?

packages/core/src/node/messaging/messaging-contribution.ts
```diff
@@ -54,9 +57,13 @@ export class MessagingContribution implements BackendApplicationContribution, Me
     @inject(MessagingListener)
     protected readonly messagingListener: MessagingListener;
 
+    @inject(HttpWebsocketAdapterFactory)
+    protected readonly httpWebsocketAdapterFactory: () => HttpWebsocketAdapter;
+
     protected webSocketServer: ws.Server | undefined;
     protected readonly wsHandlers = new MessagingContribution.ConnectionHandlers<ws>();
```
colin-grant-work, line 64:
```suggestion
    protected readonly wsHandlers = new MessagingContribution.ConnectionHandlers<ws | HttpWebsocketAdapter>();
```

Same on l. 195.

packages/core/src/node/messaging/messaging-contribution.ts
```diff
@@ -115,9 +122,47 @@ export class MessagingContribution implements BackendApplicationContribution, Me
...
+        }
+        app.use(json({ limit: '50mb' }));
+        app.use(async (req, res, next) => {
+            const body: HttpRequestBody = req.body;
+            if (body && typeof body.id === 'string') {
+                let adapter = this.httpWebsocketAdapters.get(body.id);
+                if (!adapter) {
+                    adapter = this.httpWebsocketAdapterFactory();
+                    this.httpWebsocketAdapters.set(body.id, adapter);
+                    this.handleConnection(adapter as unknown as ws, req);
```
colin-grant-work, line 150:
```suggestion
                    this.handleConnection(adapter, req);
```

colin-grant-work (commented):
This seems to be working, but with a significant caveat. I was able to open the application, see the errors described, and go about my business (modifying a few files). However, when I switched to a different workspace using File > Open Recent Workspace and reloaded, I saw the message again, used Run > Start Debugging to set up a debug configuration for a baby JS file, I wound up in a weird state. I could not save the `launch.json` with the new configuration, and an endless series of errors showed up in the console:

image

On the backend, I got a `JavasScript heap out of memory` error

image

I was able to reproduce by following the same steps: open a workspace, modify some files, then open another workspace, start trying to modify files ---> backend soon crashes outof memory.

msujew:
> open a workspace, modify some files, then open another workspace, start trying to modify files ---> backend soon crashes outof memory.

I can't reproduce this in Gitpod or locally. Does anyone else also experience this issue?

msujew:
@colin-grant-work I'm not 100% sure I fixed the cause of your exact out of memory error, because I couldn't reproduce the issue using your method. Would you mind retesting this? My recent changes fixed at least the oom error that I found.

colin-grant-work (approved):
This looks good to me. I can no longer reproduce the OOM crash that occurred before, and all of the functionality I tested worked well, though in some cases slightly more slowly (as expected).

msujew:
@paul-marechal Any more feedback? If not, I would merge this and #9820 as well later today/tomorrow.

paul-marechal:
@msujew no more feedback, sounds good to me :)

vince-fugnitto:
@msujew the changes have broken electron unfortunately, we should try and fix it or revert:

image

PR #7359, opened 2020-03-18
Update styling for description when a tree node is focused

kaiyue0329:
#### What it does
Fixes: #7189 

Updated the styling for tree nodes in the following places so that the node description can be seen when it is focused under the Light Color Theme:
1. Search in workspace
2. Debug sidebar
3. Problems View

#### How to test
1. Search in workspace
- Change the color theme by pressing f1 or Ctrl+Shift+P and setting it to light(theia)\
- Click the Search Icon or press ctrl+shift+f
- Search for something (eg. background color)
-  The description for the tree node should be easily seen when the node is focused, loses focus or unselected

2. Debug sidebar
- Go to the 'View' on the top menu and click Debug
- Select 'Launch with Node' from the dropdown and start the debugger
- The description for the node under BREAKPOINTS should be easily seen when the node is focused, loses focus or unselected

3. Problems View
- Go to the 'View' on the top menu and click Problems
- The description for the node should be easily seen when the node is focused, loses focus or unselected

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #9742, opened 2021-07-15
Improve problem marker readability in light mode

### LINKED ISSUE #9741: Highlighting makes parts of the problem message unreadable
When using problem view to display the error message. After highlighting one of the problems, highlighted text is unfortunately no longer readable! like following:
th3
Theia version is 1.14.0
How to reproduce:
1. On terminal create a new node project with npm init
2. Use this package.json which has an error:
[package2.zip](https://github.com/eclipse-theia/theia/files/6821382/package2.zip)
3. Open Problem view.
4. Select a problem row

msujew:
#### What it does
Closes #9741

Aligns our marker CSS with vscode. Allows for better readability in light mode.

Problem markers on current `master`. Note the hard to read light theme:

image

image

With the change in place:

image

image

Compared to vscode:

image

image

#### How to test

1. Open a project with problem markers in Theia.
2. Focus/Select the problem marker. Assess readability, look and feel.
3. Change between the different themes (dark/light/solarized, etc.) and repeat step 2.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto:
@msujew I'll take a look at the changes :+1: In the past we tried to fix the problem, but it was true for multiple trees:
- https://github.com/eclipse-theia/theia/pull/7359

msujew:
@vince-fugnitto Thanks for the input. I applied the same fix to the other trees as well.

vince-fugnitto:
I was hoping there could be a general solution in `@theia/core` for all trees, and possibly trees in the future but since this is low-effort enough I'd be fine to go with this solution for the moment :+1:

vince-fugnitto (commented):
In a multi-root workspace, we should also think of handling marker info nodes:

image

msujew:
> In a multi-root workspace, we should also think of handling marker info nodes

Forgot about those, done.

> I was hoping there could be a general solution in @theia/core for all trees

What would you propose? We would basically need some kind of common class for all marker info elements, right? I could try to implement a more general solution instead of the current approach.

vince-fugnitto:
> > I was hoping there could be a general solution in @theia/core for all trees
> 
> What would you propose? We would basically need some kind of common class for all marker info elements, right? I could try to implement a more general solution instead of the current approach.

Perhaps it is out of scope for the moment (I'd be fine with this approach and a refactoring after). I was thinking to have a generic class for node descriptions which can be handled in `core` and refactor some trees not to include their custom styling:

https://github.com/eclipse-theia/theia/blob/8dd8401c8125ab679de727f6576d76cea4d89d2c/packages/core/src/browser/tree/tree-widget.tsx#L45-L56

msujew:
> I'd be fine with this approach and a refactoring after

I'd usually agree, but one of my meetings got canceled and I got an hour of free time ;)

> I was thinking to have a generic class for node descriptions

Yeah, that's what I thought about as well. I implemented it that way, WDYT?

vince-fugnitto (approved):
The changes look much better to me :+1: and work correctly as well!
Do you mind updating the commit message and description to reflect the changes better, I believe the scope changed since the initial implementation.

msujew:
@vince-fugnitto FYI, I updated the severity marker as well, so that it aligns with vscode.

image

vince-fugnitto:
> @vince-fugnitto FYI, I updated the severity marker as well, so that it aligns with vscode.

Looks great! In the future we can do the same for the `file-icons` in the trees (ex: explorer), and also https://github.com/eclipse-theia/theia/issues/9744.

PR #9754, opened 2021-07-16
Pass `EditorWidgetFactory` options to `TextEditorProvider`

### LINKED ISSUE #9746: Diff Editor always puts cursor on last changed line when opened
When a new diff editor is opened, it always moves the cursor to the line of the last change in the file. It appears that any selection behavior specified in the opener options take effect but then are wiped out. If the same opener options are used again once the widget is open, the selection takes effect and the cursor is _not_ moved to the last modified line.

_Originally posted by @RomanNikitenko in https://github.com/eclipse-theia/theia/issues/8334#issuecomment-727033747_

Steps:
- Clone https://github.com/RomanNikitenko/vscode-test-extension.git
- Build the extension using yarn and copy it to `theia/plugins` folder.
- Start `Theia` and open `Theia` project as a workspace folder - it's required to have some files for testing.
- `F1` => run `Test 'vscode.open' command`
 `Test 'vscode.open' command` uses `vscode.open` command which opens `package.json` file and selects `engines` section. 
- `F1` => run `Create file-copy`
It creates copy of the `package.json` file - `package_copy.json` and provides some difference in the content related to node version.
- `F1` => run `Test 'vscode.diff' command` 
This one uses `vscode.diff` which opens the provided resources in the diff editor to compare their contents. So, it should open the diff editor for `package.json` and `package_copy.json` files, `engines` section should be selected for the right editor.

Also, you can test the same extension for `VS Code` and compare the behavior:
- go to `vscode-test-extension` folder
- use `vsce package` command to prepare `vsix`.
- go to `VS Code` and add `vsix` to extensions
- use `Steps` section above to see `VS Code` behavior for the same commands.

colin-grant-work:
#### What it does

This PR fixes #9746 by giving `DiffEditor`s knowledge of the options used to create them. This allows them to set the correct value for `alwaysRevealFirst` in the options passed to the `DiffNavigator`. It may also make it possible to fix some of the lifecycle difficulties we previously had with `EditorManager.revealSelection` in a manner more consistent with VSCode's / Monaco's approach.

#### How to test

1. Please follow the steps outlined in #9746.
2. You should observe that, when you open the diff editor for the first time, the correct selection is revealed and the editor does not scroll away to the next change in the file.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-editor-provider.ts
+++ b/packages/monaco/src/browser/monaco-editor-provider.ts
@@ -16,7 +16,7 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import URI from '@theia/core/lib/common/uri';
-import { EditorPreferenceChange, EditorPreferences, TextEditor, DiffNavigator } from '@theia/editor/lib/browser';
+import { EditorPreferenceChange, EditorPreferences, TextEditor, DiffNavigator, EditorFactoryOptions } from '@theia/editor/lib/browser';
 import { DiffUris } from '@theia/core/lib/browser/diff-uris';
 import { inject, injectable, named } from '@theia/core/shared/inversify';
 import { DisposableCollection, deepClone, Disposable, } from '@theia/core/lib/common';
@@ -122,9 +122,9 @@ export class MonacoEditorProvider {
         return reference.object;
     }
 
-    async get(uri: URI): Promise<MonacoEditor> {
+    async get(uri: URI, options?: EditorFactoryOptions): Promise<MonacoEditor> {
         await this.editorPreferences.ready;
-        return this.doCreateEditor(uri, (override, toDispose) => this.createEditor(uri, override, toDispose));
+        return this.doCreateEditor(uri, (override, toDispose) => this.createEditor(uri, override, toDispose, options));
     }
 
     protected async doCreateEditor(uri: URI, factory: (override: IEditorOverrideServices, toDispose: DisposableCollection) => Promise<MonacoEditor>): Promise<MonacoEditor> {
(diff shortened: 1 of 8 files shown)


colin-grant-work:
I want to add some caveats / discussion points here:
 - The bug is caused by a quirk in how VSCode treats one setting (`alwaysRevealFirst`) for diff editors, and it may be reasonable to conclude that that behavior is a bug, so it may not be ours to fix.
 - This solution involves some dubious work to get around the workings of the `WidgetManager`, which probably isn't safe since callers shouldn't know exactly how the `WidgetManager` is implemented.
 - The fact that I want to get around certain aspects of how the `WidgetManager` is implemented may indicate that we've conflated two things that ought to be kept distinct: the options that identify a widget 'sufficiently' for some purpose and the options that might be useful _in case_ the `WidgetManager` creates a new widget rather than returning an old one. Specifically, for the `EditorManager`, if an editor with a given set of {type, uri, counter} options already exists, that editor is fine to pass to any caller who asks for those options. However, if such an editor _doesn't_ exist, it may be useful to pass extra information to the `EditorWidgetFactory`, but at the moment, we can't do that legitimately: the `WidgetManager` implementation implies that if the options are different (or stringify differently), a new widget should be created, and no distinction is made between taking one 'off the shelf' and creating a new one. It would probably be good if there weren't situations where those things diverged - i.e. where creating a new widget with the minimum identifying information was wrong - but we can't always guarantee that if we don't have total control of the widget's initialization routine (as we don't control the Monaco classes).

packages/monaco/src/browser/monaco-editor-provider.ts
```diff
@@ -341,10 +346,11 @@ export class MonacoEditorProvider {
         toDispose.push(editor.onLanguageChanged(() => this.updateMonacoDiffEditorOptions(editor)));
         return editor;
     }
-    protected createMonacoDiffEditorOptions(original: MonacoEditorModel, modified: MonacoEditorModel): MonacoDiffEditor.IOptions {
+    protected createMonacoDiffEditorOptions(original: MonacoEditorModel, modified: MonacoEditorModel, factoryOptions?: EditorFactoryOptions): MonacoDiffEditor.IOptions {
         const options = this.createOptions(this.diffPreferencePrefixes, modified.uri, modified.languageId);
         options.originalEditable = !original.readOnly;
         options.readOnly = modified.readOnly;
+        options.alwaysRevealFirst = Boolean(factoryOptions?.selection);
```
colin-grant-work, line 353: This line is the actual fix. It needs to be known when the `DiffNavigator` is created because it's supposed to be `private` and inaccessible thereafter.

msujew (approved):
I can confirm that the issue exists on `master` and is fixed by this change:

* Using the testing steps provided, the diff view is opened with the first change select ✔️ 
* Other diff views (scm/compare opened editors) are continue working as expected ✔️ 

Using the `Create file-copy` command I noticed that the copied file is opened twice in the same area. Though this is not related to this change and also happens on `master`.

colin-grant-work:
Closing for now. It does fix the problem, and may be a reasonable solution, but only if the underlying changes solve more problems than just this one.

PR #9755, opened 2021-07-17
upgrade TypeScript to latest

paul-marechal:
Upgrade TypeScript version and fix new compilation errors.

#### How to test

Everything should work like before.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/source-tree/source-tree.ts
+++ b/packages/core/src/browser/source-tree/source-tree.ts
@@ -41,50 +41,56 @@ export class SourceTree extends TreeImpl {
         return parent.element.getElements();
     }
 
-    protected toNode(element: TreeElement, index: number, parent: TreeElementNodeParent): TreeElementNode {
-        const id = element.id ? String(element.id) : (parent.id + ':' + index);
-        const name = id;
-        const existing = this.getNode(id);
-        const updated = existing && <TreeElementNode>Object.assign(existing, { element, parent });
+    /**
+     * Convert a parent's elements into `TreeElementNode` or `CompositeTreeElementNode`.
+     *
+     * If a node already exists in this tree, we'll mutate it rather than create a new updated one.
+     */
+    protected toNode(element: TreeElement, index: number, parent: TreeElementNodeParent): TreeElementNode | CompositeTreeElementNode {
+        const id: string = element.id === undefined ? `${parent.id}:${index}` : String(element.id);
+        // The base `TreeImpl` class expects to store simple `TreeNode` instances,
+        // but we know for a fact that we handle `TreeElementNode` in this `SourceTree` class.
+        // TODO: `Tree` should be generic and require something like `T extends TreeNode`?
+        const existing = this.getNode(id) as TreeElementNode | CompositeTreeElementNode | undefined;
+        if (existing) {
+            existing.element = element;
+            existing.parent = parent;
+        }
         if (CompositeTreeElement.hasElements(element)) {
-            if (updated) {
-                if (!ExpandableTreeNode.is(updated)) {
-                    Object.assign(updated, { expanded: false });
+            if (existing) {
+                if (!ExpandableTreeNode.is(existing)) {
+                    (existing as Partial<ExpandableTreeNode>).expanded = false;
                 }
-                if (!CompositeTreeNode.is(updated)) {
-                    Object.assign(updated, { children: [] });
+                if (!CompositeTreeNode.is(existing)) {
+                    (existing as Partial<CompositeTreeNode>).children = [];
                 }
-                return updated;
+                return existing;
             }
             return {
                 element,
                 parent,
                 id,
-                name,
+                name: id,
                 selected: false,
                 expanded: false,
                 children: []
-            } as TreeElementNode;
+            };
         }
-        if (CompositeTreeElementNode.is(updated)) {
-            delete updated.expanded;
-            delete updated.children;
-        }
-        if (updated) {
-            if (ExpandableTreeNode.is(updated)) {
-                delete updated.expanded;
+        if (existing) {
+            if (ExpandableTreeNode.is(existing)) {
+                delete (existing as Partial<ExpandableTreeNode>).expanded;
             }
-            if (CompositeTreeNode.is(updated)) {
-                delete updated.children;
+            if (CompositeTreeNode.is(existing)) {
+                delete (existing as Partial<CompositeTreeNode>).children;
             }
-            return updated;
+            return existing;
         }
         return {
             element,
             parent,
             id,
-            name,
-            selected: false
+            name: id,
+            selected: false,
         };
     }
 
@@ -97,7 +103,7 @@ export interface TreeElementNode extends TreeNode, SelectableTreeNode {
     parent: TreeElementNodeParent
 }
 export namespace TreeElementNode {
-    export function is(node: TreeNode | undefined): node is TreeElementNode {
+    export function is(node?: TreeNode): node is TreeElementNode {
         return SelectableTreeNode.is(node) && 'element' in node;
     }
 }
@@ -108,7 +114,7 @@ export interface CompositeTreeElementNode extends TreeElementNode, CompositeTree
     parent: TreeElementNodeParent
 }
 export namespace CompositeTreeElementNode {
-    export function is(node: TreeNode | undefined): node is CompositeTreeElementNode {
+    export function is(node?: TreeNode): node is CompositeTreeElementNode {
         return TreeElementNode.is(node) && CompositeTreeNode.is(node) && ExpandableTreeNode.is(node) && !!node.visible;
     }
 }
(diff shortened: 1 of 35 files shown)


packages/core/src/browser/source-tree/source-tree.ts
```diff
@@ -41,50 +43,56 @@ export class SourceTree extends TreeImpl {
...
+            existing.element = element;
+            existing.parent = parent;
+        }
         if (CompositeTreeElement.hasElements(element)) {
-            if (updated) {
-                if (!ExpandableTreeNode.is(updated)) {
-                    Object.assign(updated, { expanded: false });
+            if (existing) {
+                if (!ExpandableTreeNode.is(existing)) {
+                    (existing as any).expanded = false;
```
colin-grant-work: TS seems happy with these `Object.assign` calls. Why do you prefer this style?

packages/core/src/browser/source-tree/source-tree.ts
```diff
@@ -41,50 +43,56 @@ export class SourceTree extends TreeImpl {
...
-            delete updated.expanded;
-            delete updated.children;
+            };
         }
-        if (updated) {
-            if (ExpandableTreeNode.is(updated)) {
-                delete updated.expanded;
+        if (existing) {
+            if (ExpandableTreeNode.is(existing)) {
+                delete (existing as any).expanded;
```
colin-grant-work: Rather than an `any` cast here, you can use `Partial<ExpandableTreeNode>` (or `CompositeTreeNode`, where appropriate). I'd prefer to avoid file-scoped disabling of `no-explicit-any` where possible.

packages/plugin-ext/src/main/browser/plugin-contribution-handler.ts
```diff
@@ -300,7 +300,7 @@ export class PluginContributionHandler {
         }
 
         if (contributions.colors) {
-            pushContribution('colors', () => this.colors.register(...contributions.colors));
+            pushContribution('colors', () => this.colors.register(...contributions.colors!));
```
colin-grant-work, line 303: Would it be safer to do this:
```typescript
const {colors} = contributions;
pushContribution('colors', () => this.colors.register(...colors));
```
Then TS is happy that colors will exist without the non-null assertion.

packages/typehierarchy/src/browser/tree/typehierarchy-tree.ts
```diff
@@ -31,8 +33,8 @@ export class TypeHierarchyTree extends TreeImpl {
         if (TypeHierarchyTree.Node.is(parent)) {
             await this.ensureResolved(parent);
             if (parent.children.length === 0) {
-                delete parent.children;
-                delete parent.expanded;
+                delete (parent as any).children;
```
colin-grant-work, line 36: Looks like some `as any` TreeNode casts are still here.

packages/core/src/browser/source-tree/source-tree.ts
paul-marechal: Because TS checks nada when doing `Object.assign`... It's worse that using `any` and relies on useless reflection.

packages/plugin-ext/src/main/browser/plugin-contribution-handler.ts
paul-marechal: I feel like it is OK with the way it currently is, not a fan of binding consts just to please typing...

colin-grant-work: It isn't _really_ just to please typing in this case. It seems reasonable to believe that nothing will delete the `colors` between this call and when the callback is called, but TS is reminding us that `delete` is a thing, and the rest of this PR shows we're not afraid to use it.

packages/core/src/browser/source-tree/source-tree.ts
colin-grant-work: In that case, I think the `Partial<ExpandableTreeNode>` trick will work here, as well, avoiding the `any`.

packages/plugin-ext/src/main/browser/plugin-contribution-handler.ts
paul-marechal: > It seems reasonable to believe that nothing will delete the colors between this call and when the callback is called [...]

It's the goal of the current implementation.

> [...] but TS is reminding us that `delete` is a thing, and the rest of this PR shows we're not afraid to use it.

Agreed, but once again in this specific instance, adding the `non-null` coercion or capturing the value out of the `contribution` object wouldn't help much should a `delete` be introduced for any reason. In one case it will fail because of an undefined value that was supposed to be defined, and in the other case it will keep working despite the value being deleted.

TL;DR: whenever we have code that isn't specifically tailored like this one, we could do as you say and capture properties to ensure they don't disappear from the object. Although ultimately removing properties mid-flight sounds like a foot gun.

packages/core/src/browser/source-tree/source-tree.ts
paul-marechal: I updated the code with `Partial` like you suggested. We're still kinda lying to the compiler, but without `any`s now.

packages/plugin-ext/src/hosted/node/plugin-reader.ts
```diff
@@ -58,7 +58,8 @@ export class HostedPluginReader implements BackendApplicationContribution {
                         // the request was already closed
                         return;
                     }
-                    if ('code' in e && e['code'] === 'ENOENT') {
+                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                    if ((e as any).code === 'ENOENT') {
```
vince-fugnitto, line 62: If we properly type `e` based on what we might expect then we would not need to cast as `any`?

```ts
res.sendFile(filePath, { root: localPath }, e => {
```

to

```ts
res.sendFile(filePath, { root: localPath }, (e: Error & { code?: string }) => {
```

packages/core/src/browser/source-tree/tree-source.ts
```diff
@@ -37,12 +39,12 @@ export interface CompositeTreeElement extends TreeElement {
     getElements(): MaybePromise<IterableIterator<TreeElement>>
 }
 export namespace CompositeTreeElement {
-    /* eslint-disable @typescript-eslint/no-explicit-any */
-    export function is(element: CompositeTreeElement | any): element is CompositeTreeElement {
-        return !!element && 'getElements' in element;
+    export function is(element?: any): element is CompositeTreeElement {
```
vince-fugnitto, line 42: Since we do a `typeof` check, we can simplify it to `unknown`?

packages/core/src/browser/tree/tree-expansion.ts
```diff
@@ -65,15 +67,16 @@ export interface ExpandableTreeNode extends CompositeTreeNode {
 }
 
 export namespace ExpandableTreeNode {
-    export function is(node: Object | undefined): node is ExpandableTreeNode {
-        return !!node && CompositeTreeNode.is(node) && 'expanded' in node;
+    export function is(node?: any): node is ExpandableTreeNode {
```
vince-fugnitto, line 70: Can you clarify why we want to now accept `any` given we try to avoid using it project wide?
Why for example in `ExpandableTreeNode` are we not doing like `SelectableTreeNode` and not accepting the parent `TreeNode` as a type?

packages/plugin-ext/src/hosted/node/plugin-reader.ts
paul-marechal: `any` is more convenient here, we still end up lying to the compiler (can make a typo either in the `{ code?: string }` or `(e as any).code`).

packages/core/src/browser/source-tree/tree-source.ts
paul-marechal: Not really since TypeScript makes `object` type guards pretty painful to handle. `any` is fine here.

packages/core/src/browser/tree/tree-expansion.ts
paul-marechal: Because it doesn't really matter. The whole point of those functions is to tell TypeScript how to coerce values into specific types: `node is ExpandableTreeNode`.

packages/plugin-ext/src/hosted/node/plugin-reader.ts
paul-marechal: Found `NodeJS.ErrnoException` that also inherits `Error`, will use that instead.

paul-marechal:
I am contemplating dropping this TypeScript upgrade.

My original idea was that we would be able to use more recent language constructs (named tuple elements, etc...)

The problem I now realize this can cause is that if our typings use those newer constructs, it might break dependents that rely on older versions of TypeScript.

In short, let's keep our old compiler to be as easy to consume as possible?

tsmaeder:
@paul-marechal yet another upgrade policy we should define to be a reliable partner to adopters. The pains of growing up as a project, I guess.

PR #9759, opened 2021-07-21
Align code-lens css with vscode

msujew:
#### What it does

Aligns the code-lens css with vscode. Currently code-lenses always look like they're highlighted, because they contain the `a` tag:

grafik

With this change in place, the code-lenses in Theia will look like this:

https://user-images.githubusercontent.com/4377073/126474427-8b625bd9-309c-44d6-a249-6a66da607ab7.mp4

#### How to test

1. Ensure the `vscode-npm` extension is installed in Theia.
2. Open a `package.json` with the `scripts` property.
3. Observe the color of the start button and the `Debug` text.
4. Hover over the element and observe the change in color.
5. This can also be repeated for the debugger extension for java for example. See the screencap above.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/style/index.css
+++ b/packages/monaco/src/browser/style/index.css
@@ -71,6 +71,14 @@
   height: var(--theia-scrollbar-width) !important;
 }
 
+.monaco-editor .codicon.codicon-debug-start {
+  color: var(--theia-debugIcon-startForeground) !important;
+}
+
+.monaco-editor .codelens-decoration a {
+  color: inherit !important;
+}
+
 .monaco-editor .reference-zone-widget .ref-tree .referenceMatch .highlight {
   color: unset !important;
 }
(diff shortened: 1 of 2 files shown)


packages/monaco/src/browser/style/index.css
```diff
@@ -71,6 +71,14 @@
   height: var(--theia-scrollbar-width) !important;
 }
 
+.monaco-editor .codicon.codicon-debug-start {
+  color: #89d185 !important;
+}
```
vince-fugnitto, line 76: @msujew I'm not sure about hardcoding values for the icon, can you clarify why it is necessary and where the color comes from?

msujew: Yeah, totally not necessary. I've added a new variable (`theia-start-debug-color`) to the `debug` css files. I took the color from [here](https://github.com/microsoft/vscode/blob/81626fc25e2ca00cce51f61f6880cfb7c9ff27c5/src/vs/workbench/contrib/debug/browser/debugColors.ts#L24-L28).

I thought about adding `bright` and `dark` css files to monaco, but as the debug package is necessary to display code-lenses, I added the variable in there.

msujew: I see, thank you! And I kept wondering, where those magic css variables where coming from...

I changed my code to align with the coding guidelines.

vince-fugnitto (approved):

Issue #9722, opened 2021-07-12
Add performance measurement


sgraband:
### Feature Description:
We would like to contribute a mechanism to measure/monitor the system performance, more precisely the startup time of Theia to avoid regressions and as a benchmark for possible improvements.

We currently have a script (using puppeteer) that uses the Google DevTools performance tracing to measure the largest contentful paint (LCP) metric. The script is parameterized and can run the measurement multiple times, if neccessary. The script starts a performance tracing, opens theia and stops the tracing again. This generates a profile file, that contains all events that were captured during the recording. This file can also be imported into the Google DevTools to see a timeline of all events. Then the LCP metric is parsed from the file and written to the console. If there is more than 1 run a mean and standard deviation are calculated and logged as well.

We believe this script is already useful as a stand-lone tool as it allows measuring performance effects/improvements in a consistent way.

As useful extensions, we could also integrate this into the nightly build and in PR builds.
However, in our opinion, hardcoded limits for the measurements should be avoided, as this will lead to a lot of failed builds.
One possible solution to integrate the script into the build could be to run the startup measurement multiple times during the nightly build and keep a history of the results. These numbers can then be used to compare the results of PR builds to see if the performance is affected. For example, startup times that take 20% longer than the mean of the nightly builds could be flagged with a warning.

Any opinions or suggestions? We will first contribute the script and then potential integrations if wanted.

@JonasHelming

tsmaeder:
Having a single metric that tells you whether your PR makes Theia slower or faster sounds like a very good thing. However, I have a couple of questions:

1. Is LSP a good metric? Can the user start doing work at the the point of of the LCP? Theia is not a traditional website and probably many things are started lazily or in the background.
2. What state is the IDE started to? Are there editors open, etc.?
3. Is back-end startup included in the number?
4. While browser load time is important, why are we only measuring startup time? Is browser load time a good measure for most PRs?
5. If running on CI, how do we ensure the results are not skewed by other tasks being run on the same shared infrastructure?
6. Having a number is useful, but if we get a "yellow card" on our PR, what are the expectations on the developer and how does the developer find out what he needs to do? If we have bad numbers, we should at least provide the trace file.

I'm wondering where you folks are coming from: is startup time a problem in your work? Is VS Code considerably faster, for example?

One thing that would be really cool IMO is to have a suite of common tasks that we time for every release that could be part of the release process.

sgraband:
[...]

PR #9777, opened 2021-07-26
Add performance measurement

sgraband:
#### What it does

This commit adds a script for performance measurement using the Chrome DevTools tracing.
Currently the startup time of the browser-app is being measured.
The script uses Puppeteer so other processes can be measured in the future as well.
For more information see the `README.md` in the scripts/performance directory.

This PR is intended as the first step to an eventual performance test suite which could also be integrated into the CI tooling.
See https://github.com/eclipse-theia/theia/issues/9722 for more information about this topic.

Example output:
```
[Performance][Startup][1] Largest Contentful Paint (LCP): 2.95 seconds
[Performance][Startup][2] Largest Contentful Paint (LCP): 2.823 seconds
[Performance][Startup][3] Largest Contentful Paint (LCP): 2.829 seconds
[Performance][Startup][4] Largest Contentful Paint (LCP): 2.835 seconds
[Performance][Startup][5] Largest Contentful Paint (LCP): 2.778 seconds
[Performance][Startup][6] Largest Contentful Paint (LCP): 2.795 seconds
[Performance][Startup][7] Largest Contentful Paint (LCP): 2.772 seconds
[Performance][Startup][8] Largest Contentful Paint (LCP): 2.672 seconds
[Performance][Startup][9] Largest Contentful Paint (LCP): 2.706 seconds
[Performance][Startup][10] Largest Contentful Paint (LCP): 2.804 seconds
[Performance][Startup][MEAN] Largest Contentful Paint (LCP): 2.796 seconds
[Performance][Startup][STDEV] Largest Contentful Paint (LCP): 0.072 seconds
```

Signed-off-by: sgraband@eclipsesource.com
#### How to test

See the README on how you can run the script manually or run the `performance:startup` node script in the root directory to measure the startup time. You can also adjust the parameters to for example execute more iterations or to run the test headful.
Additionally, you can import the generated files in `scripts/performance/profiles` in the performance tab of the Chrome DevTools.
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/scripts/performance/measure-performance.js
+++ b/scripts/performance/measure-performance.js
@@ -1,0 +1,182 @@
+/********************************************************************************
+ * Copyright (C) 2021 STMicroelectronics and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+// @ts-check
+const puppeteer = require('puppeteer');
+const fs = require('fs');
+const fsExtra = require('fs-extra');
+const resolve = require('path').resolve;
+const workspacePath = resolve('./workspace');
+const profilesPath = './profiles/';
+
+const lcp = 'Largest Contentful Paint (LCP)';
+const performanceTag = braceText('Performance');
+
+let name = 'StartupPerformance';
+let url = 'http://localhost:3000/#' + workspacePath;
+let folder = 'profile';
+let headless = true;
+let runs = 10;
+
+(async () => {
+    let defaultUrl = true;
+
+    const args = require('yargs/yargs')(process.argv.slice(2)).argv;
+    if (args.name) {
+        name = args.name.toString();
+    }
+    if (args.url) {
+        url = args.url.toString();
+        defaultUrl = false;
+    }
+    if (args.folder) {
+        folder = args.folder.toString();
+    }
+    if (args.runs) {
+        runs = parseInt(args.runs.toString());
+    }
+    if (args.headless) {
+        if (args.headless.toString() === 'false') {
+            headless = false;
+        }
+    }
+
+    if (defaultUrl) { fsExtra.ensureDirSync(workspacePath); }
+    fsExtra.ensureDirSync(profilesPath);
+    const folderPath = profilesPath + folder;
+    fsExtra.ensureDirSync(folderPath);
+
+    const deployed = await waitForDeployed(url, 10, 500);
+    if (deployed == false) {
+        console.error('Could not connect to application.')
+    } else {
+        await measurePerformance(name, url, folderPath, headless, runs);
+    }
+})();
+
+async function measurePerformance(name, url, folder, headless, runs) {
+    const durations = [];
+    for (let i = 0; i < runs; i++) {
+        const runNr = i + 1;
+        const browser = await puppeteer.launch({ headless: headless });
+        const page = await browser.newPage();
+
+        const file = folder + '/' + runNr + '.json';
+        await page.tracing.start({ path: file, screenshots: true });
+        await page.goto(url);
+        // This selector is for the theia application, which is exposed when the loading indicator vanishes
+        await page.waitForSelector('.theia-ApplicationShell', { visible: true });
+        // Prevent tracing from stopping too soon and skipping a LCP candidate
+        await delay(1000);
+
+        await page.tracing.stop();
+
+        await browser.close();
+
+        const time = await analyzeStartup(file)
+        durations.push(time);
+        logDuration(name, runNr, lcp, time.toFixed(3), runs > 1);
+    }
+
+    if (runs > 1) {
+        const mean = calculateMean(durations);
+        logDuration(name, 'MEAN', lcp, mean);
+        logDuration(name, 'STDEV', lcp, calculateStandardDeviation(mean, durations));
+    }
+}
+
+async function analyzeStartup(profilePath) {
+    let startEvent;
+    const tracing = JSON.parse(fs.readFileSync('./' + profilePath, 'utf8'));
+    const lcpEvents = tracing.traceEvents.filter(x => {
+        if (isStart(x)) {
+            startEvent = x;
+            return false;
+        }
+        return isLCP(x);
+    });
+
+    if (startEvent !== undefined) {
+        return duration(lcpEvents[lcpEvents.length - 1], startEvent);
+    }
+    throw new Error('Could not analyze startup');
+}
+
+function isLCP(x) {
+    return x.name === 'largestContentfulPaint::Candidate';
+}
+
+function isStart(x) {
+    return x.name === 'TracingStartedInBrowser';
+}
+
+function duration(event, startEvent) {
+    return (event.ts - startEvent.ts) / 1000000;
+}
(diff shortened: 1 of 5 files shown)


scripts/performance/measure-performance.js
```diff
@@ -0,0 +1,195 @@
...
+                if (args[index] === '--runs') {
+                    runs = parseInt(next);
+                    index++;
+                    continue;
+                }
+            }
+        }
+    }
+
+    if (defaultUrl) getOrCreateFolder(path);
```
tsmaeder, line 69: Curly braces, please (see https://github.com/eclipse-theia/theia/wiki/Coding-Guidelines#style). Also, what do we do if the path the user passes in does not exist?

scripts/performance/measure-performance.js
```diff
@@ -0,0 +1,195 @@
...
+        const browser = await puppeteer.launch({ headless: headless });
+        const page = await browser.newPage();
+
+        const file = folder + '/' + runNr + '.json';
+        await page.tracing.start({ path: file, screenshots: true });
+        await page.goto(url);
+        // This selector is for the problems indicator in the status bar, as this is one of the last elements to be rendered.
+        await page.waitForSelector('.theia-ApplicationShell', { visible: true });
+        // Prevent tracing from stopping to soon and skipping a LCP candidate
+        await delay(1000);
```
tsmaeder, line 96: This needs explanation: is there no true stop condition we can wait for?

scripts/performance/measure-performance.js
```diff
@@ -0,0 +1,195 @@
...
+    browser.close();
+
+    await measurePerformance(name, url, folderPath, headless, runs);
+})();
+
+async function measurePerformance(name, url, folder, headless, runs) {
+    const multipleRuns = runs > 1 ? true : false;
+    let runNr = 1;
+    let durations = [];
+    while (runs > 0) {
```
tsmaeder, line 86: This sounds like a classic for loop: `for (i = 0; i < runs; i++)`. Also, if you won't need the `multipleRuns` variable if you leave the `runs` parameter alone.

scripts/performance/measure-performance.js
```diff
@@ -0,0 +1,195 @@
...
+    const browser = await puppeteer.launch({ headless: true });
+    const page = await browser.newPage();
+    await retryConnection(() => page.goto(url), 1000);
+    browser.close();
+
+    await measurePerformance(name, url, folderPath, headless, runs);
+})();
+
+async function measurePerformance(name, url, folder, headless, runs) {
+    const multipleRuns = runs > 1 ? true : false;
```
tsmaeder, line 83: This does not need a conditional, I believe: `const multipleRuns = runs > 1;

scripts/performance/measure-performance.js
```diff
@@ -0,0 +1,195 @@
...
+    let runNr = 1;
+    let durations = [];
+    while (runs > 0) {
+        const browser = await puppeteer.launch({ headless: headless });
+        const page = await browser.newPage();
+
+        const file = folder + '/' + runNr + '.json';
+        await page.tracing.start({ path: file, screenshots: true });
```
tsmaeder, line 91: Do we need to write the file out or can we just use what we get back from page.tracing.stop()?

scripts/performance/measure-performance.js
```diff
@@ -0,0 +1,195 @@
...
+
+function getOrCreateFolder(path) {
+    if (!fs.existsSync(path)) {
+        fs.mkdirSync(path);
+    }
+}
+
+const retryConnection = (fn, ms) => new Promise(url => {
+    fn()
+        .then(url)
```
tsmaeder, line 183: Isn't `url` really the `resolve` function?

scripts/performance/measure-performance.js
```diff
@@ -0,0 +1,195 @@
...
+    return '[' + text + ']';
+}
+
+function getOrCreateFolder(path) {
+    if (!fs.existsSync(path)) {
+        fs.mkdirSync(path);
+    }
+}
+
+const retryConnection = (fn, ms) => new Promise(url => {
```
tsmaeder, line 181: Dont' we need maximum number of retries until failure?

scripts/performance/measure-performance.js
```diff
@@ -0,0 +1,195 @@
...
+
+function isLCP(x) {
+    return x.name === 'largestContentfulPaint::Candidate';
+}
+
+function isStart(x) {
+    return x.name === 'TracingStartedInBrowser';
+}
+
+function duration(event, startEvent) {
```
tsmaeder, line 142: I can understand it makes sense to restrict precision when printing durations, etc, but I don't think we should do it when calculating stuff. Same goes for "mean", etc.

tsmaeder:
@sgraband thanks for this contribution: it's important that we keep track of startup performance. I have a bunch of suggestions/comments which are mostly code style issues.

scripts/performance/measure-performance.js
```diff
@@ -0,0 +1,195 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+// @ts-check
+const puppeteer = require('puppeteer');
```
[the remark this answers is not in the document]
sgraband: I moved the `puppeteer` dependencies as well as the `yargs` dependency to the root package.json.

scripts/performance/measure-performance.js
sgraband: Since we need to programatically decide when to watch the tracing, the difficulty is to choose an event, where the last LCP Candidate was already recorded. This is close to the time, where the `-theia-ApplicationShell` selector is visible. However, due to the loading indicator fading away it takes a little bit extra time for the tracing to realize that there is nothing more being painted. Hence, we need to wait a little bit longer. Originally i thought about waiting for a selector which comes later in the DOM, but we cannot be really sure, if this selector exists for other custom theia applications, which should also be able to use the script.

scripts/performance/measure-performance.js
sgraband: The advantage of writing the file out is, that you can then also use that file to have a detailed look of the tracing inside of the Google DevTools using this file. This can give additional information as well as a good overview over the applications startup.

.npmrc
```diff
@@ -0,0 +1 @@
+scripts-prepend-node-path=true
```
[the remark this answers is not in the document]
sgraband: Without this option the script showed a warning:
```
The node binary used for scripts is /tmp/yarn--1627025300882-0.06251617126804576/node but npm is using
/home/simon/.nvm/versions/node/v12.22.1/bin/node itself. Use the --scripts-prepend-node-path option to include the
path for the node binary npm was
executed with.
```
However, right now i cannot reproduce it. Should i remove it?

sgraband:
@tsmaeder I rebased the changes to master. Could you have another look?

colin-grant-work:
@sgraband, it looks like the builds are failing here because running `yarn` on this commit changes the `yarn.lock`. You may need to commit those changes so that the builds are satisfied that our `yarn.lock` is up to date.

sgraband:
@colin-grant-work rebasing seemed to do the trick.

colin-grant-work (approved):
This script is performing correctly for me and seems like a reasonable basis for judging changes in startup performance for the application. @tsmaeder, do you have any additional comments?

.npmrc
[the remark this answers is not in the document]
tsmaeder: https://github.com/yarnpkg/yarn/issues/6685 I haven't traced trough all that, but I think yarn is just making copies 🤷 I think we should remove it (or understand why it happens).

tsmaeder:
@tsmaeder nothing to prevent merging.

.npmrc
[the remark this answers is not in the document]
sgraband: Removed it.

sgraband:
Rebased and removed the .npmrc file.

Issue #9795, opened 2021-07-28
Unexpected Identifier ffunction


Infuser:
Creating this ticket in hope it will be useful for anyone else if they get this issue, we started getting this issue today when running a new theia build.

Build runs fine but when opening the browser we where getting unexpected identifier causing theia to completely fail looking at the source of the error it became likely it was an issue with packaging as we could see code in the bundle like ffunction instead of function, notice the double ff

Adding this to package.json fixed the issue:

```json
  "resolutions": {
    "webpack-sources": "2.3.1"
  },
```

vince-fugnitto:
I confirmed it fixes the recent CI failures on [`theia-apps`](https://github.com/theia-ide/theia-apps):
- https://github.com/theia-ide/theia-apps/pull/487

We will need to try and fix it upstream (this repo) since it likely affects all packaging.
cc @paul-marechal

Infuser:
We noticed webpack sources is coming from 2 locations compression-webpack-plugin and webpack itself and the versions dont match

paul-marechal:
A better fix is to put a resolution as:

```json
"resolutions": {
  "**/webpack": "5.46.0"
}
```

`webpack@^5.47.0` is the one depending on `webpack-sources@^3.0.1`, so just forcing the latter to resolve to `webpack-sources@2` is not the best idea when the dependent package expects the next major version.

PR #9799, opened 2021-07-29
Fix webpack build issues in production mode

paul-marechal:
There is a bug with a new dependency from `webpack`: Using
`webpack-sources@3` produces bogus bundles, failing our application
build process.

We will need to update the `webpack` range again once this issue is
fixed in whatever package is causing this build issue on our end.

#### What it does

Fixes https://github.com/eclipse-theia/theia/issues/9795 by constraining our `webpack` range to avoid the bogus version.

See https://github.com/webpack/webpack/issues/13887 for failure details.

Once this issue will be fixed by `webpack` and/or `webpack-sources` we should be able to revert the constrained range to something more "open" (like before).

#### How to test

To see the failure:

- Checkout master
- Run `yarn upgrade`
- Run `yarn`
- Run `yarn --cwd examples/browser theia build`

This should lead to an error coming out of `webpack`/`Terser`.

To make sure this issue is fixed:

- Checkout this PR
- Run `yarn`
- Run `yarn --cwd examples/browser theia build`

`yarn upgrade` is already part of those changes.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
  - [workspace] added support for multiple selections in 'Add folder to workspace' dialog. [#9684](https://github.com/eclipse-theia/theia/pull/9684)
  - [debug] added support for managing debug sessions for extensions from debug panel (previously only possible using `Hosted Plugin` commands) [#8706](https://github.com/eclipse-theia/theia/pull/8706)
 
+<a name="notable_changes_1.16.0">[Notable Changes:](#notable_changes_1.16.0)</a>
+
+- [application-manager] defines a new range for `webpack` (`^5.36.2 <5.47.0`). `webpack@5.47.0` depends on `webpack-sources@^3.0.1` but this new version produces bogus bundles in Theia applications. The fix works by constraining the `webpack` version range to not pull newer versions for Theia v1.16.0 meaning clients creating Theia applications will not be affected by the bundling failures caused by the new dependency. The bogus library will most likely be fixed before next release (v1.17.0) so we'll need to update the `webpack` range back to pull newer versions again (bug/performance/security updates).
+
 <a name="breaking_changes_1.16.0">[Breaking Changes:](#breaking_changes_1.16.0)</a>
 
 - [debug] `DebugSession` and `PluginDebugSession` constructors accept a `parentSession` of type `DebugSession | undefined` as their 3rd parameter, offsetting every subsequent parameter by one. [#9613](https://github.com/eclipse-theia/theia/pull/9613)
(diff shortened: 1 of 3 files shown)


paul-marechal:
Few notes about the failure: This is pretty much out of our hands as a framework. One of our transitive dependencies was updated but bogus. We could just wait it out until the bug is fixed and published, and the next time someone pulls our dependencies or do a `yarn upgrade` the fix will be included.

Given this, this PR is a stop gap to prevent that issue for dependents that wouldn't want to deal with Theia's deep dependencies as this is a quite nasty bug. Once the fix is available, we shall revert the `webpack` range to keep pulling newer versions again.

vince-fugnitto (approved):
I confirm that the `webpack` error we were experiencing is now resolved, and we can think of updating when they provide a fix upstream.

marcdumais-work:
> Maybe a shorter version of your explanation in your comment above, Paul.

something like your commit message is what I had in mind.

marcdumais-work (approved):
LGTM, thanks Paul! Vince will add a note in CHANGELOG, answering my PR comment.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
  - [workspace] added support for multiple selections in 'Add folder to workspace' dialog. [#9684](https://github.com/eclipse-theia/theia/pull/9684)
  - [debug] added support for managing debug sessions for extensions from debug panel (previously only possible using `Hosted Plugin` commands) [#8706](https://github.com/eclipse-theia/theia/pull/8706)
 
+<a name="notable_changes_1.16.0">[Notable Changes:](#notable_changes_1.16.0)</a>
+
+- [application-manager] defines a new range for Webpack such as: `^5.36.2 <5.47.0`. `webpack@5.47.0` updated to `webpack-sources@^3.0.1` but this new version produces bogus bundles in Theia applications. The issue being fixed by constraining `webpack` to not pull newer versions for Theia v1.16.0 is so that clients making Theia applications will not be affected by the bundling failures caused by the new dependency. The bogus library will most likely be fixed before next release (v1.17.0) so we'll need to update the `webpack` range back to pull newer versions again (bug/performance/security updates).
```
vince-fugnitto, line 12: minor: (I can also update in my `changelog` pull-request)

```suggestion
- [application-manager] defines a new range for `webpack` (`^5.36.2 <5.47.0`). `webpack@5.47.0` depends on `webpack-sources@^3.0.1` but this new version produces bogus bundles in Theia applications. The fix works by constraining the `webpack` version range to not pull newer versions for Theia v1.16.0 meaning clients creating Theia applications will not be affected by the bundling failures caused by the new dependency. The bogus library will most likely be fixed before next release (v1.17.0) so we'll need to update the `webpack` range back to pull newer versions again (bug/performance/security updates).
```

PR #9805, opened 2021-07-30
terminal: add mouse support of GUI terminal application , such as vim, mc

linhanyu:
#### What it does

vim use binary message to indicate mouse event but not utf-8 data. In xterm.js, it be emitted by onBinary. Simplely connect it to send Data and resolved. Test environment: MacOS

#### How to test

Open terminal in macos, open vim and type :set mouse=a. 
Then Type some word and use mouse to click, normally, cursor would move for.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (commented):
@linhanyu thank you for your first contribution :+1: 

Please be sure to sign the [eclipse contributor agreement (eca)](https://github.com/eclipse-theia/theia/blob/master/CONTRIBUTING.md#eclipse-contributor-agreement) with the same email as your authorship so we can accept the changes (it is the reason for the failed CI check).

linhanyu:
oh, I'm not sure what to do?
I have force pushed, change author and sign-off

vince-fugnitto:
The `ECA` sign-up remains unsuccessful for `305991071@qq.com`:

image

> There is no valid ECA on file for 305991071@qq.com.

linhanyu:
so complicated~

> The `ECA` sign-up remains unsuccessful for `305991071@qq.com`:
> 
> image
> 
> > There is no valid ECA on file for [305991071@qq.com](mailto:305991071@qq.com).

thanks~ I got it

msujew (approved):
I can confirm that the issue (vim not being able to receive mouse events) exists on `master` and is addressed by your changes nicely. Code LGTM as well.

packages/terminal/src/browser/terminal-widget-impl.ts
```diff
@@ -499,7 +503,10 @@ export class TerminalWidgetImpl extends TerminalWidget implements StatefulWidget
                     }
                 };
 
-                const disposable = this.term.onData(sendData);
+                const disposable = new DisposableCollection;
+                disposable.push(this.term.onData(sendData));
+                disposable.push(this.term.onBinary(sendData));
```
paul-marechal, line 508: It might work for now, but I'd like someone to understand the implications of doing this?

- [Xterm.js mentions how this binary data should be consumed](https://xtermjs.org/docs/guides/encoding/#output).
- [VS Code seems to follow it](https://github.com/microsoft/vscode/blob/ee1655a82ebdfd38bf8792088a6602c69f7bbd94/src/vs/platform/terminal/node/terminalProcess.ts#L413).

Either we can actually just shove it that way, same as the regular `onData` event, or we're missing some processing?

packages/terminal/src/browser/terminal-widget-impl.ts
```diff
@@ -499,7 +503,10 @@ export class TerminalWidgetImpl extends TerminalWidget implements StatefulWidget
                     }
                 };
 
-                const disposable = this.term.onData(sendData);
+                const disposable = new DisposableCollection;
```
paul-marechal, line 506: I am surprised our linter didn't flag this `new` call without parenthesis?

Please add the parenthesis for the sake of consistency with the rest of the code base, thanks :)

packages/terminal/src/browser/terminal-widget-impl.ts
linhanyu: In theia cases, both node-pty in Macos and Windows use Nodejs-PipeSocket as ipc. And PipeSocket were initialized by encoding=undefined, which mean socket treat input flow as utf-8 string. String data would be convert to Buffer after write to Nodejs-Socket by Buffer.from(chunk,  'utf-8').
Binary string is equal to latin string. Every code unit tread as a latin character (0 ~ 255).So It's safe apply  Buffer.from(chunk, 'utf-8') to binary-string because latin code point was included by UTF-8 charset.
The problem of padding for convert latin string to UTF-8 string, actually happend in convert a Buffer to string. That's problem of LANG or LC_ALL environment variable of shell. So you can temporary ignore these case.

linhanyu: The second argument of Buffer.from is meaning "How Buffer treat every string code point and save them in UInt8Array?".Both latin string and utf-8 string use 1 byte.

paul-marechal (approved):
Thanks for the explanation, it looks good for now.

PR #9820, opened 2021-08-03
filesystem: upload via HTTP

paul-marechal:
Refactor the code to upload files via HTTP rather than WS. This
simplifies the code by removing the ad-hoc WS upload protocol and should
make it easier to configure proxies as it will be using good ol' HTTP.

Note 1: The file upload is done over HTTP POST `multipart/form-data`.

Note 2: I had to use the older XMLHttpRequest (XHR/AJAX) API in order to
monitor the upload progress. The Fetch API is not there yet.

Add new preference `files.maxConcurrentUploads` that defaults to 1 as it
seems to the be fastest on my test environment. Can be user-customized.

Add new class `AsyncQueue` that runs a fixed amount of async tasks
concurrently. We cannot use `p-queue` in the browser, hence this minimal
implementation.

This PR goes hand-in-hand with https://github.com/eclipse-theia/theia/pull/9731

#### How to test

- Upload files by dnd over the file explorer, many should work.
- Upload files using the contextual command (right click in the file navigator), many should work.
- Progress notification should be somewhat accurate in both cases.

#### Review checklist

#### Reminder for reviewers


paul-marechal:
cc @msujew

packages/filesystem/src/browser/filesystem-preferences.ts
```diff
@@ -80,24 +80,30 @@ These have precedence over the default associations of the languages installed.'
...
-            'scope': 'language-overridable'
+            type: 'boolean',
+            default: false,
+            description: 'When enabled, will trim trailing whitespace when saving a file.',
+            scope: 'language-overridable'
+        },
+        'files.maxConcurrentUploads': {
+            type: 'integer',
+            default: 1,
+            description: 'Maximum number of concurrent files to upload when uploading multiple files.',
```
msujew, line 91: Setting this preference to something <1 leads to this unexpected behavior when trying to upload a file:

1. The 'Uploading Files' notification pops up for a few frames.
2. The notification disappears and an error is logged, since the `AsyncQueue` will throw.

I think we should handle this more gracefully. Does Theia support the notion of using only positive integers as preferences? Alternatively check for a value <1 in the `FileUploadService` and either 1. Behave if it was simply set to `1` or 2. Display a warning and ignore the dd event.

paul-marechal: Ah yes, forgot to gracefully handle that case.

> Does Theia support the notion of using only positive integers as preferences?

Not AFAIK. I made it so it will fall back on its feet no matter what, using the default if invalid.

packages/core/src/common/async-queue.ts
```diff
@@ -0,0 +1,110 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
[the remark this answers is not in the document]
paul-marechal: Sound good to me, thanks for the suggestion :)

msujew (approved):
I can confirm that this change allows to upload files without a websocket connection:

* Just to be sure, I disabled websockets and tested this change without the ability to establish websocket connections ✔️ 
* Uploading single files works as expected ✔️
* Uploading multiple files with `files.maxConcurrentUploads = 1` works as expected ✔️ 
* Uploading multiple files with `files.maxConcurrentUploads > 1` works as expected ✔️
* Uploading files with `files.maxConcurrentUploads < 1` works just like `files.maxConcurrentUploads = 1` ✔️
* Progress bar updates correctly as well with different configurations ✔️

packages/core/src/common/async-queue.ts
```diff
@@ -0,0 +1,130 @@
...
+import { Disposable } from './disposable';
+import { Deferred } from './promise-util';
+import { MaybePromise } from './types';
+
+export interface AsyncQueueOptions {
+
+    /**
+     * Maximum amount of async tasks to run concurrently.
+     *
+     * Default to 1.
```
vince-fugnitto, line 27: minor:

```suggestion
     * Defaults to 1.
```

packages/core/src/common/async-queue.ts
```diff
@@ -0,0 +1,130 @@
...
+        }
+        this.concurrency = concurrency;
+    }
+
+    get closed(): boolean {
+        return this._closed;
+    }
+
+    /**
+     * Schedule async tasks. The task will be ran as soon as it can.
```
vince-fugnitto, line 67: minor (either `run` or `execute`):

```suggestion
     * Schedule async tasks. The task will run as soon as it can.
```

packages/filesystem/package.json
```diff
@@ -5,12 +5,14 @@
...
     "@theia/core": "1.15.0",
     "@types/body-parser": "^1.17.0",
+    "@types/multer": "^1.4.7",
     "@types/rimraf": "^2.0.2",
     "@types/tar-fs": "^1.16.1",
     "@types/uuid": "^7.0.3",
     "body-parser": "^1.18.3",
     "http-status-codes": "^1.3.0",
     "minimatch": "^3.0.4",
+    "multer": "^1.4.2",
```
vince-fugnitto, line 15: I confirmed the added dependencies are license compatible :+1:

vince-fugnitto (commented):
I confirmed that uploading works as expected with different use-cases and preference values :+1:

msujew:
@paul-marechal Could you please resolve the merge conflict? You/I can merge this afterwards.

PR #9827, opened 2021-08-03
fix vscode debug source map problem

linhanyu:
#### What it does

Fix source map issue that made debug in electron frontend process unavailable in vscode, which made us cannot set breakpoint

#### How to test

1. attach to frontend process, set breakpoint

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -151,7 +151,12 @@
       "type": "chrome",
       "request": "attach",
       "name": "Attach to Electron Frontend",
-      "port": 9222
+      "port": 9222,
+      "webRoot": "${workspaceRoot}",
+      "sourceMapPathOverrides": {
+        "webpack://*": "${workspaceRoot}/*",
+        "webpack://@theia/example-electron/../*": "${workspaceRoot}/*"
+      }
     },
     {
       "name": "Launch VS Code Tests",


alvsan09:
I have managed to reproduce the issue, 
the problem  seems to have been introduced with: 
* b2e679163e (HEAD) application-manager: pull newer version of Webpack

However this fix did not solve the problem for me.

@paul-marechal, could you also take a look at this issue, as you are more familiar with  the are in the commit mentioned above.

.vscode/launch.json
```diff
@@ -151,7 +151,12 @@
       "type": "chrome",
       "request": "attach",
       "name": "Attach to Electron Frontend",
-      "port": 9222
+      "port": 9222,
+      "webRoot": "${workspaceRoot}",
```
thegecko, line 155: Please use `workspaceFolder` rather than `workspaceRoot` as it is [deprecated](https://code.visualstudio.com/docs/editor/variables-reference#_why-isnt-workspaceroot-documented)

thegecko, line 155: In our tests, we've found `webRoot` isn't required to fix this issue.

.vscode/launch.json
```diff
@@ -151,7 +151,12 @@
...
+      "sourceMapPathOverrides": {
+        "webpack://*": "${workspaceRoot}/*",
+        "webpack://@theia/example-electron/../*": "${workspaceRoot}/*"
+      }
```
thegecko, line 159: I _think_ this can be written more generically as:

```suggestion
      "sourceMapPathOverrides": {
        "webpack:////src/*": "${workspaceFolder}/*",
        "webpack:////*": "/*"
      }
```

Which is how we have this working in Mbed Studio & Keil Studio

thegecko (changes requested):
Minor suggestions, as yet untested

paul-marechal:
@alvsan09 I opened an alternative PR with a slightly simplier config, can you tell me if it fixes the issue?

https://github.com/eclipse-theia/theia/pull/10101

PR #9830, opened 2021-08-03
Fix #9779: add support for window.menuBarVisibility

### LINKED ISSUE #9779: add support for window.menuBarVisibility
The menu bar is not very useful for many users who uses keyboard shortcuts. It would be greate if Theia supported "window.menuBarVisibility":"compact/hiden" as VSCode.

[See example](https://www.youtube.com/watch?v=bpXmCdh-FzM)

ChayaLau:
Signed-off-by: Esther Perelman <esther.perelman@sap.com>, Chaya Lau <chaya.lau@sap.com>

#### What it does
Fix #9779 add support for window.menuBarVisibility 
The pull-request adds the menu Bar Visibility setting to window setting, this allows compact and hidden layout for menu bar.

#### How to test

click on settings -> preferences ->
tap menu bar visibility in the search
menu bar visibility section should be displayed and drop down should appear with the options : 
classic - menu bar should be displayed on the top - as original.
visible - behaves like classic.
hidden - menu bar does not appear anywhere.
custom - Menu is displayed as a compact button in the sidebar.
 
#### Review checklist

#### Reminder for reviewers


vince-fugnitto (commented):
@ChayaLau thank you for the contribution :+1: 

Please ensure that:
- the breaking changes introduced in this pull-request are properly documented in the changelog - if we can avoid breakage please go that route.
- the default layout is broken when starting the application in the browser.
- please confirm the behavior when testing the changes in `electron` for the different settings.

ChayaLau:
> @ChayaLau thank you for the contribution 👍
> 
> Please ensure that:
> 
> * the breaking changes introduced in this pull-request are properly documented in the changelog - if we can avoid breakage please go that route.
> * the default layout is broken when starting the application in the browser.
> * please confirm the behavior when testing the changes in `electron` for the different settings.

hi, 
Thanks for the reply
I fixed the PR according to your comments - point 2 and 3 - thanks!
about point 1 - I am new in this repo, and I did not fully understand what you meant about the breakage  changes.
should i add a line in the changelog describing this PR? or did you mean something more than that?

 in addition i wanted to mention this setting in VS CODE - has one additional option to this - toggle which i still did not develop, and planning on adding it soon. will this be OK adding it in like this and adding the option in the future, or will it be merged only if we add the toggle now?

Thanks,
Chaya

vince-fugnitto (changes requested):
@ChayaLau the main-menu is broken in the `electron` target with these changes, please address the issue before we proceed with an in-depth review.

The following is with `classic` (default), no main-menu exists:

image

ChayaLau:
Thanks @vince-fugnitto for this important fix, i fixed the broken electron menu

packages/core/src/browser/shell/side-panel-handler.ts
```diff
@@ -385,12 +393,30 @@ export class SidePanelHandler {
...
+        this.topMenu.removeMenu(menuId);
+    }
+
     /**
      * Add a menu to the sidebar bottom.
      *
      * If the menu is already added, it will be ignored.
      */
-    addMenu(menu: SidebarBottomMenu): void {
+    addBottomMenu(menu: SidebarMenu): void {
```
vince-fugnitto, line 419: This is a breaking change for anyone calling `addMenu` and should be documented in the `changelog`.

packages/core/src/browser/shell/side-panel-handler.ts
```diff
@@ -399,7 +425,7 @@ export class SidePanelHandler {
      *
      * @param menuId id of the menu to remove
      */
-    removeMenu(menuId: string): void {
+    removeBottomMenu(menuId: string): void {
```
vince-fugnitto, line 428: This is a breaking change for anyone calling `removeBottomMenu` and should be documented in the `changelog`.

packages/core/src/browser/shell/sidebar-menu-widget.tsx
```diff
@@ -20,9 +20,9 @@ import { ReactWidget } from '../widgets';
 import { ContextMenuRenderer } from '../context-menu-renderer';
 import { MenuPath } from '../../common/menu';
 
-export const SidebarBottomMenuWidgetFactory = Symbol('SidebarBottomMenuWidgetFactory');
+export const SidebarMenuWidgetFactory = Symbol('SidebarMenuWidgetFactory');
 
-export interface SidebarBottomMenu {
+export interface SidebarMenu {
```
vince-fugnitto, line 25: Renaming `SidebarBottomMenu` is a breaking change, and should be documented as such.

packages/core/src/browser/shell/sidebar-menu-widget.tsx
```diff
@@ -20,9 +20,9 @@ import { ReactWidget } from '../widgets';
 import { ContextMenuRenderer } from '../context-menu-renderer';
```
vince-fugnitto, line 20: Renaming the file is a breaking change, and should be documented as such.

packages/core/src/browser/shell/sidebar-menu-widget.tsx
```diff
@@ -31,11 +31,11 @@ export interface SidebarBottomMenu {
 }
 
 /**
- * The menu widget placed on the bottom of the sidebar.
+ * The menu widget placed on the bottom or top of the sidebar.
  */
 @injectable()
-export class SidebarBottomMenuWidget extends ReactWidget {
-  protected readonly menus: SidebarBottomMenu[];
+export class SidebarMenuWidget extends ReactWidget {
```
vince-fugnitto, line 37: Renaming the class is a breaking change.

packages/core/src/browser/style/sidepanel.css
```diff
@@ -207,10 +207,14 @@
   font-size: var(--theia-private-sidebar-icon-size);
 }
 
-.theia-sidebar-bottom-menu i:hover {
+.theia-sidebar-menu i:hover {
   color: var(--theia-activityBar-foreground);
 }
 
+.theia-sidebar-menu > i.codicon-menu{
```
vince-fugnitto, line 214:
```suggestion
.theia-sidebar-menu > i.codicon-menu {
```

packages/core/src/electron-browser/menu/electron-main-menu-factory.ts
```diff
@@ -71,33 +71,49 @@ export class ElectronMainMenuFactory {
...
-            template.unshift(this.createOSXMenu());
+            electron.remote.Menu.setApplicationMenu(createdMenuBar);
+        } else {
+            electron.remote.getCurrentWindow().setMenu(createdMenuBar);
+        }
+    }
+
+    createMenuBar(): Electron.Menu | null {
+        let preference = this.preferencesService.get<string>('window.menuBarVisibility');
+        preference = preference ? preference : 'classic';
```
vince-fugnitto, line 103: When would `window.menuBarVisiblity` be `undefined` given it has a default value?

```suggestion
        const preference = this.preferencesService.get<string>('window.menuBarVisibility') ||  'classic';
```

packages/core/src/electron-browser/menu/electron-main-menu-factory.ts
```diff
@@ -71,33 +71,49 @@ export class ElectronMainMenuFactory {
...
+        if (preference && ['classic', 'visible'].includes(preference)) {
```
vince-fugnitto, line 104: Given the above, `preference` will always be set, so no need?

```suggestion
        if ( ['classic', 'visible'].includes(preference)) {
```

packages/core/src/electron-browser/menu/electron-main-menu-factory.ts
```diff
@@ -71,33 +71,49 @@ export class ElectronMainMenuFactory {
...
+            const menu = electron.remote.Menu.buildFromTemplate(template);
+            this._menu = menu;
+            return this._menu;
         }
-        const menu = electron.remote.Menu.buildFromTemplate(template);
-        this._menu = menu;
-        return menu;
+        this._menu = undefined;
+        // eslint-disable-next-line no-null/no-null
+        return null;
```
vince-fugnitto, line 116: We should avoid using `null` and use `undefined` instead. Same in other areas.

vince-fugnitto, line 116: We should avoid using `null` and instead use `undefined`. Same is true in other areas.

packages/core/src/browser/menu/browser-menu-plugin.ts
```diff
@@ -348,6 +368,9 @@ export class BrowserMenuBarContribution implements FrontendApplicationContributi
     @inject(ApplicationShell)
     protected readonly shell: ApplicationShell;
 
+    @inject(CorePreferences)
+    protected readonly corePreferences: CorePreferences;
```
vince-fugnitto, line 372: Leftover?

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -354,16 +354,38 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
...
+                            iconClass: 'codicon codicon-menu',
+                            title: 'Application Menu',
+                            menuPath: ['menubar'],
+                            order: 0,
+                        });
+                    } else {
+                        app.shell.leftPanelHandler.removeTopMenu(mainMenuId);
+                    }
+                    break;
+                }
```
vince-fugnitto, line 382: We should refactor and create a `onPreferenceChange` handler method so it will be easier for downstream applications to override if necessary (ex: if they want to handle the preference change event differently).

packages/core/src/browser/core-preferences.ts
```diff
@@ -97,6 +97,20 @@ export const corePreferenceSchema: PreferenceSchema = {
...
             description: 'Whether to interpret keypresses by the `code` of the physical key, or by the `keyCode` provided by the OS.'
         },
+        'window.menuBarVisibility': {
+            type: 'string',
+            enum: ['classic', 'visible', 'hidden', 'compact'],
+            markdownEnumDescriptions: [
+                'Menu is displayed at the top of the window and only hidden in full screen mode.',
+                'Menu is always visible at the top of the window even in full screen mode.',
+                'Menu is always hidden.',
+                'Menu is displayed as a compact button in the sidebar. This value is ignored when `#window.titleBarStyle#` is `native`.'
```
vince-fugnitto, line 107: I believe this enum description is incorrect given that ``#window.titleBarStyle#` is unsupported, and `native` menus in electron are always true but compact still displays the side-menu.

packages/core/src/electron-browser/menu/electron-main-menu-factory.ts
ChayaLau: I looked and saw the menu in electron accepts types : electron.Menu | null (and not undefined)
`    private setMenu(menu: electron.Menu | null = this.factory.createMenuBar(), electronWindow: electron.BrowserWindow = electron.remote.getCurrentWindow()): void {
        if (isOSX) {
            electron.remote.Menu.setApplicationMenu(menu);
        } else {
            // Unix/Windows: Set the per-window menus
            electronWindow.setMenu(menu);
        }
    }`
we can change the return value of the function, but then we will have to handle the undefined value and replace with null when using the menu functions. (there are few more places calling the functions after create)

should we do that?

packages/core/src/browser/common-frontend-contribution.ts
ChayaLau: I separated it to another protected function, is that what you meant?

packages/core/src/electron-browser/menu/electron-main-menu-factory.ts
vince-fugnitto: I believe this comment is not addressed, do we need to provide the `default` value?
If we need to provide the default, we can:

```ts
const preference = this.preferencesService.get<string>('window.menuBarVisibility', 'classic');
```

packages/core/src/electron-browser/menu/electron-main-menu-factory.ts
vince-fugnitto: I see, we can accept `null` then:

- https://github.com/electron/electron/blob/main/docs/api/browser-window.md#winsetmenumenu-linux-windows
- https://github.com/electron/electron/blob/main/docs/api/browser-window.md#winremovemenu-linux-windows
 
We'll need to confirm this actually works on `macOS`.

packages/core/src/browser/common-frontend-contribution.ts
vince-fugnitto: Correct, it makes it easier to extend if needed, and also cleans it up.

colin-grant-work:
What is the intended behavior of the `hidden` setting in the browser? When I set to hidden, I get this:

image

The menu has been removed from the top section, but the top section remains, so the user doesn't get back any vertical space by hiding the menu. In Electron, the bar disappears completely, but it also doesn't contain the branding icon that the browser does.

vince-fugnitto:
> The menu has been removed from the top section, but the top section remains, so the user doesn't get back any vertical space by hiding the menu. In Electron, the bar disappears completely, but it also doesn't contain the branding icon that the browser does.

I agree, the whole top-menu should be hidden else there is no value for end-users.

packages/core/src/browser/shell/sidebar-menu-widget.tsx
```diff
@@ -55,14 +55,14 @@ export class SidebarBottomMenuWidget extends ReactWidget {
...
+    if (menu) {
+      const index = this.menus.indexOf(menu);
+      if (index !== -1) {
+        this.menus.splice(index, 1);
+        this.update();
       }
+    }
   }
 
   protected onClick(e: React.MouseEvent<HTMLElement, MouseEvent>, menuPath: MenuPath): void {
```
colin-grant-work, line 68: Minor, and related only to old code: this method uses the click as the anchor for the menu. In VSCode, the menu is always anchored to the right of the element, so it likely uses the element's coordinates instead.

packages/core/src/browser/shell/side-panel-handler.ts
```diff
@@ -187,9 +194,9 @@ export class SidePanelHandler {
         return toolbar;
     }
 
-    protected createSidebarBottomMenu(): SidebarBottomMenuWidget {
-        const bottomMenu = this.sidebarBottomWidgetFactory();
-        bottomMenu.addClass('theia-sidebar-bottom-menu');
+    protected createSidebarMenu(): SidebarMenuWidget {
+        const bottomMenu = this.sidebarWidgetFactory();
+        bottomMenu.addClass('theia-sidebar-menu');
```
colin-grant-work, line 199: In addition to the breaking changes Vince has noted, changing this classname is also breaking, as it requires adopters to modify their CSS (as you've modified the internal CSS) if they have any rules targeting these items. I understand the impetus to rename things according to their new behavior, but I'm not sure that it adds enough value here to justify the number of breakages it introduces.

colin-grant-work:
The description for `visible` indicates that the menu should always be visible, but setting to `visible` and then double-clicking a tab hides the menu.

image

colin-grant-work:
Because of the way MacOS handles menu bars, this preference doesn't have any effect in Theia - the menu bar can be made compact, but it always appears in the OS menu bar anyway. The preference doesn't appear to be present in the MacOS version of VSCode, so perhaps we should also conditionally bind it so that it doesn't appear as a preference in Electron if we detect that we're running on MacOS.

ChayaLau:
> The menu has been removed from the top section, but the top section remains, so the user doesn't get back any vertical space by hiding the menu. In Electron, the bar disappears completely, but it also doesn't contain the branding icon that the browser does.

i saw this is the behavior in VS code, are we sure we want the whole line to disappear not like VS Code?

packages/core/src/browser/shell/side-panel-handler.ts
ChayaLau: > In addition to the breaking changes Vince has noted, changing this classname is also breaking, as it requires adopters to modify their CSS (as you've modified the internal CSS) if they have any rules targeting these items. I understand the impetus to rename things according to their new behavior, but I'm not sure that it adds enough value here to justify the number of breakages it introduces.

can i add this to the breaking changes, or you mean we need to refactor this all commit in order to avoid the breaking changes?

colin-grant-work:
> I saw this is the behavior in VS code, are we sure we want the whole line to disappear not like VS Code?

If it's the intended behavior, then it is working as intended 👍. I would agree with @vince-fugnitto that that's somewhat odd behavior, since there doesn't then seem much point in hiding the menu, but I guess it reduces visual noise a bit.

packages/core/src/browser/shell/side-panel-handler.ts
colin-grant-work: At your discretion, but do try to avoid breaking changes unless they're necessary for or contribute substantially to enhancing the functionality delivered. On that basis, I would probably stick with the old classname.

ChayaLau:
> The description for `visible` indicates that the menu should always be visible, but setting to `visible` and then double-clicking a tab hides the menu.
> 
> image

I looked at this behavior, but i think visible in VS Code refers only to the top menu, and not the sidebar, i saw the `visible` description is VS is : "Menu is always visible at the top of the window." - i dont see anything about the sidebar, and in general always behaves as classic, should we handle the double clicking differently ?

ChayaLau:
> Because of the way MacOS handles menu bars, this preference doesn't have any effect in Theia - the menu bar can be made compact, but it always appears in the OS menu bar anyway. The preference doesn't appear to be present in the MacOS version of VSCode, so perhaps we should also conditionally bind it so that it doesn't appear as a preference in Electron if we detect that we're running on MacOS.

@colin-grant-work  - i will add the `included` property to `PreferenceDataProperty,` i saw this property is still not supported in theia (although it is used in several places).

packages/core/src/browser/shell/side-panel-handler.ts
ChayaLau: @colin-grant-work - will you recommend creating a base shared class and deriving the bottom and top from it to avoid the bottom breakage (although for now the behavior is same and derived class will be empty).?

colin-grant-work:
> @colin-grant-work - I looked at this behavior, i think visible in VS Code refers only to the top menu, and not the sidebar. 

Definitely - the sidebar is not visible here because I hid it, and that's fine. But the top bar should be visible.

> In order to refer to menuBarVisibility preference i tried to inject `corePreferences` to `TheiaDockPanel` class, but corePreferences is undefined, since I am new, i am not sure if i need to inject it in a different way or how to avoid the corePreference to be undefined?

That's a little surprising - I'm happy to take a look at this to see what's going on.

PS: I originally clicked 'Edit' on your message rather than 'Quote Reply' by accident. Sorry if I restored the content wrong!

ChayaLau:
@colin-grant-work  - i added the injection and handled the maximized mode.
i added the included prop for the preference (in order to not include this on mac OS)
i added the class name to the changelog - but will be happy if you have a different advice how to prevent the breaking changes and handle this in a prettier way.
thanks for all the great advices and help!

colin-grant-work:
I'll take a more careful look tomorrow, but I'm still seeing some oddities with the `visible` option:

image

The menu is still visible, but the maximized tab isn't truly maximized on the vertical axis, and the shell is still visible behind it.

ChayaLau:
> I'll take a more careful look tomorrow, but I'm still seeing some oddities with the `visible` option:

@colin-grant-work  - you are totally right.
i fixed this and will mark down to always test on both carefully.
thanks!

ChayaLau:
> If it's the intended behavior, then it is working as intended 👍. I would agree with @vince-fugnitto that that's somewhat odd behavior, since there doesn't then seem much point in hiding the menu, but I guess it reduces visual noise a bit.

@colin-grant-work -  i checked in VS code online edition and saw they did remove the top panel when using a browser for `compact` and `hidden` - so i added this to the commit too.

packages/core/src/browser/shell/sidebar-menu-widget.tsx
EstherPerelman: Reference for VSCode open menu code: https://github.com/microsoft/vscode/blob/99f71a032ea932670744d14d6364903a00b99ae3/src/vs/base/browser/ui/menu/menubar.ts#L957

packages/core/src/browser/shell/theia-dock-panel.ts
```diff
@@ -148,13 +178,21 @@ export class TheiaDockPanel extends DockPanel {
         }
         maximizedElement.style.display = 'block';
         this.addClass(MAXIMIZED_CLASS);
+        console.log(this.preferences);
```
EstherPerelman, line 181: forgot?

packages/core/src/browser/shell/theia-dock-panel.ts
```diff
@@ -148,13 +178,21 @@ export class TheiaDockPanel extends DockPanel {
...
+        const preference = this.preferences?.get('window.menuBarVisibility') || 'classic';
```
EstherPerelman, line 182: It seems that 'classic' is not in use (the if statement refers only to 'visible' value and will ignore undefined value
```suggestion
        const preference = this.preferences?.get('window.menuBarVisibility');
```

EstherPerelman (commented):
Hi, just found some minor mistakes

packages/core/src/browser/style/index.css
```diff
@@ -78,6 +78,10 @@ blockquote {
     background: var(--theia-editor-background);
 }
 
+.theia-visible-menu-maximized{
+  top: 32px !important;
```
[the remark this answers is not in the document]
ChayaLau: this was already fixed

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -493,6 +496,21 @@ export class ApplicationShell extends Widget {
     protected createTopPanel(): Panel {
         const topPanel = new Panel();
         topPanel.id = 'theia-top-panel';
+
+        // hide top panel when `menuBarVisibility` preference change
```
colin-grant-work, line 500:
```suggestion
        // hide top panel when `menuBarVisibility` preference changes
```

packages/core/src/browser/shell/sidebar-bottom-menu-widget.tsx
```diff
@@ -1,88 +1,24 @@
...
+
+    protected onClick(e: React.MouseEvent<HTMLElement, MouseEvent>, menuPath: MenuPath): void {
+        const botton = (e.target as HTMLButtonElement).getBoundingClientRect();
+        this.contextMenuRenderer.render({
+            menuPath,
+            anchor: {
+                x: botton.left + botton.width,
+                y: botton.top + botton.height,
+            }
+        });
```
colin-grant-work, line 21:
```suggestion
        const button = (e.target as HTMLButtonElement).getBoundingClientRect();
        this.contextMenuRenderer.render({
            menuPath,
            anchor: {
                x: button.left + button.width,
                y: button.top + button.height,
            }
        });
```

packages/core/src/browser/shell/sidebar-bottom-menu-widget.tsx
```diff
@@ -1,88 +1,24 @@
...
-
-export interface SidebarBottomMenu {
-  id: string;
-  iconClass: string;
-  title: string;
-  menuPath: MenuPath;
-  order: number; // smaller one place lower
-}
-
+/* eslint-disable @typescript-eslint/tslint/config */
```
colin-grant-work, line 1: I don't believe that this rule should be disabled here. Please restore the original header.

packages/core/src/browser/shell/sidebar-menu-widget.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  protected onClick(e: React.MouseEvent<HTMLElement, MouseEvent>, menuPath: MenuPath): void {
+    const botton = e.currentTarget.getBoundingClientRect();
+    this.contextMenuRenderer.render({
+      menuPath,
+      anchor: {
+        x: botton.left + botton.width,
+        y: botton.top,
+      }
+    });
```
colin-grant-work, line 77:
```suggestion
    const button = e.currentTarget.getBoundingClientRect();
    this.contextMenuRenderer.render({
      menuPath,
      anchor: {
        x: button.left + button.width,
        y: button.top,
      }
    });
```

packages/core/src/browser/shell/sidebar-top-menu-widget.tsx
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable @typescript-eslint/tslint/config */
```
colin-grant-work, line 1: I don't believe that this rule should be disabled here. Please add a header.

packages/core/src/browser/shell/theia-dock-panel.ts
```diff
@@ -44,7 +48,9 @@ export class TheiaDockPanel extends DockPanel {
      */
     readonly widgetRemoved = new Signal<this, Widget>(this);
 
-    constructor(options?: DockPanel.IOptions) {
+    constructor(options?: DockPanel.IOptions,
+        @inject(CorePreferences) protected readonly preferences?: CorePreferences
```
colin-grant-work, line 52: Adding a constructor arg is a breaking change. Please add it to the changelog.

colin-grant-work (commented):
This is working very nicely for the most part. I observe only one minor difference in behavior between Electron and browser with `classic` + maximized view:

Electron: Menu visible  
image

Browser: Menu not visible  
image

VSCode Zen mode: Menu not visible
image

Otherwise, just a few comments about breaking changes and linting.

ChayaLau:
@colin-grant-work - thanks for all the comments, i changed commit accordingly.
please if you can take a look again

colin-grant-work (approved):
@ChayaLau, this looks good to me - thanks for your hard work fixing it up. Before it's merged, please move the change log entries under the new 1.17.0 heading.

It also looks like some conflicts have crept in from a recent merge - please rebase and address those.

CHANGELOG.md
```diff
@@ -59,6 +59,11 @@
 - [debug] `DebugSession` and `PluginDebugSession` constructors accept a `parentSession` of type `DebugSession | undefined` as their 3rd parameter, offsetting every subsequent parameter by one [#9613](https://github.com/eclipse-theia/theia/pull/9613)
 - [monaco] upgraded to monaco 0.23.0 including replacement of `quickOpen` API (0.20.x) with `quickInput` API (0.23.x) [#9154](https://github.com/eclipse-theia/theia/pull/9154)
 - [workspace] `WorkspaceCommandContribution.addFolderToWorkspace` no longer accepts `undefined`. `WorkspaceService.addRoot` now accepts a `URI` or a `URI[]` [#9684](https://github.com/eclipse-theia/theia/pull/9684)
+-[core] `SidePanelHandler.addMenu` and `SidePanelHandler.removeMenu` no longer exists, instead added `addBottomMenu` and `addTopMenu` for adding menu, `removeTopMenu` and `removeBottomMenu` for removing menu. 
+`SidebarBottomMenu` interface is renamed `SidebarMenu` and handles not only bottom menu's. 
+changed style class name from `theia-sidebar-bottom-menu` to `theia-sidebar-menu`
+-[preferences] `TheiaDockPanel` constructor takes a new parameter `preferences`.
+ [#9830](https://github.com/eclipse-theia/theia/pull/9830)
```
vince-fugnitto, line 66: As @colin-grant-work mentioned, these entries will need to move under `1.17.0`:

```suggestion
- [core] `SidePanelHandler.addMenu` and `SidePanelHandler.removeMenu` no longer exists, instead added `addBottomMenu` and `addTopMenu` for adding menu, `removeTopMenu` and `removeBottomMenu` for removing menu. 
  - `SidebarBottomMenu` interface is renamed `SidebarMenu` and handles not only bottom menu's. 
  - Changed style class name from `theia-sidebar-bottom-menu` to `theia-sidebar-menu`
- [preferences] `TheiaDockPanel` constructor takes a new parameter `preferences`.
 [#9830](https://github.com/eclipse-theia/theia/pull/9830)
```

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -420,6 +414,36 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
...
+            case 'workbench.colorTheme':
+            case 'workbench.iconTheme': {
+                this.updateThemeFromPreference(e.preferenceName);
+                break;
+            }
+            case 'window.menuBarVisibility': {
+                const { newValue } = e;
+                const mainMenuId = 'main-menu';
+                if (newValue === 'compact') {
+                    app.shell.leftPanelHandler.addTopMenu({
```
vince-fugnitto, line 432: @ChayaLau we have access to `shell` here, why do we need to pass `app: FrontendApplication`?

```suggestion
                    this.shell.leftPanelHandler.addTopMenu({
```

packages/core/src/browser/shell/sidebar-top-menu-widget.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { SidebarMenuWidget } from './sidebar-menu-widget';
+import { injectable } from 'inversify';
+
+/**
+ * The menu widget placed on the top of the sidebar.
+ */
+
```
vince-fugnitto, line 23: nitpick: remove `newline`.

packages/core/src/browser/style/index.css
```diff
@@ -78,6 +78,10 @@ blockquote {
     background: var(--theia-editor-background);
 }
 
+.theia-visible-menu-maximized{
```
vince-fugnitto, line 81:
```suggestion
.theia-visible-menu-maximized {
```

ChayaLau:
@vince-fugnitto - thanks for the comments i fixed all according to your recommendations.

colin-grant-work:
@ChayaLau, it looks like there are still conflicts - please rebase and resolve those.

packages/core/src/common/preferences/preference-schema.ts
```diff
@@ -88,6 +88,7 @@ export interface PreferenceDataProperty extends PreferenceItem {
     description?: string;
     markdownDescription?: string;
     scope?: PreferenceScope;
+    included?: boolean;
```
vince-fugnitto, line 91: Leftover?

vince-fugnitto (commented):
I noticed an odd bug, but it might indicate issues with other menu items as well:
- set `menuBarVisibility` to `compact`
- attempt to open the `preferences-view` through the menu


https://user-images.githubusercontent.com/40359487/130083644-dce98078-17f3-44a8-a035-df428b3ef02f.mp4

I was only able to reproduce through `compact`, the default menu worked fine for the same use-case.

Also, when no workspace is present and `menuBarVisibility` is set to something other than `classic`, the menu appears as default then is updated to the correct value. It might be something to look into but is not super critical.


https://user-images.githubusercontent.com/40359487/130084158-f5e81311-602c-47d9-8ab4-80bc7c1aa500.mp4

vince-fugnitto:
@msujew I noticed the bug https://github.com/eclipse-theia/theia/issues/9931 as part of this pull-request, I assume you've noticed it on master?

vince-fugnitto:
@ChayaLau disregard the comment about the `preferences`, I was able to confirm it is a bug on `master` by using the bottom sidebar context-menu :+1: Mark was kind enough to already provide a fix https://github.com/eclipse-theia/theia/pull/9932.

ChayaLau:
@vince-fugnitto - thanks for the comments, i fixed accordingly.
i saw the issue you were talking about, the top panel is hidden when the preference is set either to `hidden` or `compact`.
for some reason, which i still don't understand when the workspace is not set, the preferences are set only after the top menu is build.
i can - as a default, open the window without the top panel and show it when preference is auto populated, but maybe that will be worse, because most users will probably still use the classic menu, and they will experienced their top menu delay

colin-grant-work:
> for some reason, which i still don't understand when the workspace is not set, the preferences are set only after the top menu is build.

One way to handle this might be to delay the initial population of the top panel until after `PreferenceService.ready` has resolved, and then check the value before proceeding. At the moment, the default top-bar creation routine runs, and then the actual preference setting takes effect through `PreferenceService.onPreferenceChanged` listeners, but always a moment after the bar is initially created.

colin-grant-work:
@ChayaLau, you seem to have made a change to handle the flashing menu, but your change has had the consequence that now the menu is never shown if the user has not set the preference:

image

It is not safe to rely on `PreferenceChangeEvent`s alone, because the defaults are set during the construction phase of the `PreferenceContribution`, and if no other setting is detected, no event will ever fire. The correct way to wait for preferences to be read and then act on the settings is this:

```typescript
await this.preferenceService.ready
const myValue = this.preferenceService.get('my-preference');
```

And that should be done wherever the menu is constructed so that it can be created according to whatever values is found when the `PreferenceService` is initialized, even if no event fires.

ChayaLau:
@colin-grant-work  - i tried to test before, and because i saw the preference change was firing - i thought that is enough.
according to your comment - i did change the code and registered the `preference.ready.then()` to set the top panel visibility

colin-grant-work:
@ChayaLau, I also notice that you're doing checks for whether the application is running in Electron, but in fact a parallel problem occors in Electron. 

electron-menu-bounce

Here I've set the menu bar visibility to 'compact,' but the menu appears for a moment before disappearing when the preference change event files.

ChayaLau:
> @ChayaLau, I also notice that you're doing checks for whether the application is running in Electron, but in fact a parallel problem occors in Electron.
@colin-grant-work - thanks for this point, i fixed this by - first i hide the menu, then i run the `setMenu` function which applies the preference value after the preference is ready

ChayaLau:
@colin-grant-work , @vince-fugnitto - is there any chance merging this into the next branch? 
i know this is more than a small issue to review and test, but we will be grateful if you can try pushing it in.
thanks

PR #9833, opened 2021-08-03
Update DebugRequestTypes to reflect DAP changes

colin-grant-work:
#### What it does

This PR updates the Theia types used to reflect Debug Adapter Protocol types. There have been some updates around which UI may need to be built, but in the meantime, this at least allows adopters / extension writers to write code for the new requests and events without running afoul of our type declarations. At the moment something like `session.sendRequest('readMemory', ...)` is treated as an error because we just haven't declared that possibility.

#### How to test

If it builds, then we're referring to real things in the DAP.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
I confirmed it looks correct based on the `DAP` specifications:
- https://microsoft.github.io/debug-adapter-protocol/specification

PR #8909, opened 2020-12-29
fix #8903 Encode workspacePath before being used as URL hash value. 

shuyaqian:
#### What it does

fixes [#8903](https://github.com/eclipse-theia/theia/issues/8903)
Encode workspacePath before being used as URL hash value when openNewWindow.

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #9850, opened 2021-08-05
URI:folders with '%20' in their name can be opened

OmarSdt-EC:
The commit adds the encoding of URIs. After this update, we can now
open workspaces with reserved symbols like '%' in their name.

#### What it does

FIxes #8909 

#### How to test

1. Try to open a workspace with '%' in the name
2. Check if it opens correctly
#### Review checklist

#### Reminder for reviewers


msujew (commented):
I noticed some unexpected behavior on browser reload:

1. Open a workspace with `%20` in its directory name
2. Observe that the fragment in the url does not contain the escaped sequence for `%20` but instead a space character.
3. Reloading the browser window closes the workspace. (No workspace is selected upon reload)

OmarSdt-EC:
> I noticed some unexpected behavior on browser reload:
> 
> 1. Open a workspace with `%20` in its directory name
> 2. Observe that the fragment in the url does not contain the escaped sequence for `%20` but instead a space character.
> 3. Reloading the browser window closes the workspace. (No workspace is selected upon reload)

Thank you @msujew , I'll take a look at that

msujew (approved):
I can confirm that the issue exists on `master` and is addressed by this change nicely. I tested this on both browser and Electron targets (Windows):

* Opening a directory with `%20` in its name works correctly ✔️ 
* The encoded URL appears in the browser bar ✔️ 
* Reloading the application correctly brings one back to the previously opened project ✔️

Issue #9806, opened 2021-07-30
TaskService.runTask(...) not safe for parallel Execution


tsmaeder:
### Bug Description:
When invoking `TaskService.runTask(...)` twice in a row for the same task, you can run into a race condition: the second execution of the method will not detect that the task is already running and therefore we end up executing the task twice instead of asking the user to cancel/restart the task.

### Steps to Reproduce:
I don't have an upstream reproducing scenario right now, but in Che, it reproduces since we have a tree view that you can just click on to run tasks. Double-clicking reproduces the issue: https://github.com/eclipse/che/issues/19821

### Additional Information

- Operating System: linux
- Theia Version: master

tsmaeder:
The run task method looks something like this:
```
    async runTask(task: TaskConfiguration, option?: RunTaskOption): Promise<TaskInfo | undefined> {
        const runningTasksInfo: TaskInfo[] = await this.getRunningTasks();

        if (findRunningTask(runningTaskInfo, task) { // the task is active
            const selectedAction = await this.messageService.info(`The task '${taskName}' is already active`, 'Terminate Task', 'Restart Task');
            if (selectedAction === 'Terminate Task') {
                await this.terminateTask(matchedRunningTaskInfo);
            } else if (selectedAction === 'Restart Task') {
                return this.restartTask(matchedRunningTaskInfo, option);
            }
        } else { // run task as the task is not active
            return this.doRunTask(task, option);
        }
    }
```

vince-fugnitto:
@tsmaeder do you know if `che` is using the `latest` version, I believe it might be fixed by the update to the way we `compare`:

- https://github.com/eclipse-theia/theia/pull/9647

tsmaeder:
@vince-fugnitto I don't believe comparisons have anything to do with it, otherwise the problem would not be timing-dependent.

tsmaeder:
I haven't been able to reproduce with a pure plugin approach so far: however, I have no trouble when I change TestService.runTask(...) like so:

```
    protected delay<T>(ms: number): (value: T) => Promise<T> {
        return value => new Promise((resolve, reject) => { setTimeout(() => resolve(value), ms); });
    };

    async runTask(task: TaskConfiguration, option?: RunTaskOption): Promise<TaskInfo | undefined> {
        const runningTasksInfo: TaskInfo[] = await this.getRunningTasks().then(this.delay(700));
        console.info('got running tasks, size is ' + runningTasksInfo.length);

```

[...]

PR #9858, opened 2021-08-06
Introduce lock to prevent parallel task execution

tsmaeder:
Signed-off-by: Thomas Mäder <tmader@redhat.com>

#### What it does
Fixes https://github.com/eclipse-theia/theia/issues/9806

The idea is to have a lock that prevents interleaved execution of the check whether a task is already running with the actual starting of the task. 
The approach is to delay later execution of tasks instead of ignoring them, because `runTask()` is used in various places that treat failure to start a task as an error.

#### How to test
to reproduce first: 

* Install the extension from the issue: https://github.com/eclipse-theia/theia/files/6922291/prefs-explorer-0.0.1.vsix.zip
* Change the theia source like described here: https://github.com/eclipse-theia/theia/issues/9806#issuecomment-891618281

You should be able to start multiple instances of the same task by clicking on the tasks view.

to verify it's fixed, keep the above changes, but check out the PR branch

You should not be able to create multiple instances of the same task.

#### Review checklist

#### Reminder for reviewers


tsmaeder:
Grrrr...changed code to use injection and didn't rerun the tests. Stand by...

packages/core/src/common/promise-util.ts
```diff
@@ -53,9 +53,29 @@ export async function retry<T>(task: () => Promise<T>, delay: number, retries: n
...
 
     throw lastError;
 }
+
+/**
+ * A function to allow a promise resolution to be delayed by a number of milliseconds. Usage is like so:
+ *
+ * const stringValue = await myPromise.then(delay(600)).then(value => value.toString());
+ *
+ * @param ms the number of millisecond os dealy
```
msujew, line 68:
```suggestion
 * @param ms the number of milliseconds to delay
```

packages/core/src/common/promise-util.ts
```diff
@@ -53,9 +53,29 @@ export async function retry<T>(task: () => Promise<T>, delay: number, retries: n
...
         }
     }
 
     throw lastError;
 }
+
+/**
+ * A function to allow a promise resolution to be delayed by a number of milliseconds. Usage is like so:
+ *
+ * const stringValue = await myPromise.then(delay(600)).then(value => value.toString());
```
msujew, line 66:
```suggestion
 * A function to allow a promise resolution to be delayed by a number of milliseconds. Usage is as follows:
 *
 * `const stringValue = await myPromise.then(delay(600)).then(value => value.toString());`
```

packages/task/src/browser/task-service.ts
```diff
@@ -727,34 +730,48 @@ export class TaskService implements TaskConfigurationClient {
...
+                const selectedAction = await this.messageService.info(`The task '${taskName}' is already active`, 'Terminate Task', 'Restart Task');
+                if (selectedAction === 'Terminate Task') {
+                    await this.terminateTask(matchedRunningTaskInfo);
+                } else if (selectedAction === 'Restart Task') {
+                    return this.restartTask(matchedRunningTaskInfo, option);
+                }
+            } else { // run task as the task is not active
+                console.debug('task about to start');
+                const taskInfo = await this.doRunTask(task, option);
+                this.taskStartingLock.release(lockToken);
```
msujew, line 769: I think you should be able to move `this.taskStartingLock.release(lockToken);` into a `finally` block instead of duplicating the call.

packages/core/src/common/lock.ts
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * function parallelWork() {
+ *   const token = await lock.acquire();
+ *   try {
+ *    await somePromise;
+ *     await someOtherPromise;
+ *   } finally {
+ *     lock.release(token);
+ *   }
+ * }
```
msujew, line 49: You should probably wrap this in a code block ` ```ts``` `

msujew (commented):
I can confirm that the issue exists on `master` and is resolved by these changes. Starting a task multiple times is no longer possible.

packages/task/src/browser/task-service.ts
tsmaeder: I need an early release in the other branch, so I thought it cleaner to only do a fallback release when an exception occurs.

msujew: Ah, yes I didn't see the release in the other branch. Makes sense 👍

packages/task/src/browser/task-service.ts
```diff
@@ -727,34 +730,48 @@ export class TaskService implements TaskConfigurationClient {
     }
 
     async runTask(task: TaskConfiguration, option?: RunTaskOption): Promise<TaskInfo | undefined> {
-        const runningTasksInfo: TaskInfo[] = await this.getRunningTasks();
+        console.debug('entering runTask');
+        const lockToken: number = await this.taskStartingLock.acquire();
```
paul-marechal, line 734: Wall of code incoming. For locks like these, we usually use a different but equivalent pattern in this case.

Since the lock cannot be configured and only chains tasks one after the other, we can use promise chaining:

```ts
    // move this line to the top of the class with the other fields
    protected taskExecutionQueue: Promise<void> = Promise.resolve();
    async runTask(task: TaskConfiguration, option?: RunTaskOption): Promise<TaskInfo | undefined> {
        // backup the current task execution queue
        const currenTaskExecutionQueue = this.taskExecutionQueue;
        // create the lock representing this task
        const lock = new Deferred<void>();
        // update the queue to chain ourselves
        this.taskExecutionQueue = currenTaskExecutionQueue.then(() => lock.promise);
        // wait for our turn to execute
        await currenTaskExecutionQueue;
        try {
            const runningTasksInfo: TaskInfo[] = await this.getRunningTasks();

            // check if the task is active
            const matchedRunningTaskInfo = runningTasksInfo.find(taskInfo => {
                const taskConfig = taskInfo.config;
                return this.taskDefinitionRegistry.compareTasks(taskConfig, task);
            });
            if (matchedRunningTaskInfo) { // the task is active
                lock.resolve();
                const taskName = this.taskNameResolver.resolve(task);
                const terminalId = matchedRunningTaskInfo.terminalId;
                if (terminalId) {
                    const terminal = this.terminalService.getByTerminalId(terminalId);
                    if (terminal) {
                        if (TaskOutputPresentation.shouldSetFocusToTerminal(task)) { // assign focus to the terminal if presentation.focus is true
                            this.terminalService.open(terminal, { mode: 'activate' });
                        } else if (TaskOutputPresentation.shouldAlwaysRevealTerminal(task)) { // show the terminal but not assign focus
                            this.terminalService.open(terminal, { mode: 'reveal' });
                        }
                    }
                }
                const selectedAction = await this.messageService.info(`The task '${taskName}' is already active`, 'Terminate Task', 'Restart Task');
                if (selectedAction === 'Terminate Task') {
                    await this.terminateTask(matchedRunningTaskInfo);
                } else if (selectedAction === 'Restart Task') {
                    return this.restartTask(matchedRunningTaskInfo, option);
                }
            } else { // run task as the task is not active
                return this.doRunTask(task, option);
            }
        } catch (error) {
            console.error(error);
        } finally {
            lock.resolve();
        }
    }
```

So far we avoided creating such a construct, but if you really want to introduce a new `Lock` class then I might have to look closer at the current implementation.

packages/task/src/browser/task-service.ts
```diff
@@ -727,34 +730,48 @@ export class TaskService implements TaskConfigurationClient {
...
+                            this.terminalService.open(terminal, { mode: 'reveal' });
+                        }
                     }
                 }
+                const selectedAction = await this.messageService.info(`The task '${taskName}' is already active`, 'Terminate Task', 'Restart Task');
+                if (selectedAction === 'Terminate Task') {
+                    await this.terminateTask(matchedRunningTaskInfo);
+                } else if (selectedAction === 'Restart Task') {
+                    return this.restartTask(matchedRunningTaskInfo, option);
+                }
```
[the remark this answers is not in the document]
tsmaeder: Yes, that would be possible. The behavior has not changed in that respect. As I've mentioned in the description, some call sites consider it an error if no task is started, so it's safer to preserve that behavior. 
Because otherwise, the operation is stopped until the user clicks on the dialog, which might be much later. Starting a task at that point would be more surprising than stacked dialogs.

packages/task/src/browser/task-service.ts
tsmaeder: I prefer my approach because it replaces a pattern with a piece of code that can be reused. Is there a reason we use that pattern instead a lock class?

colin-grant-work:
Although in general I'm not a fan of the question 'why not use an existing library?' I do think it deserves to be posed. There seem to be a few libraries that implement lock-like functionality with enough downloads to make it plausible that they're useful:
[async-mutex](https://www.npmjs.com/package/async-mutex)
[async-lock](https://www.npmjs.com/package/async-lock)

paul-marechal:
@colin-grant-work [async-mutex](https://www.npmjs.com/package/async-mutex) looks great.

paul-marechal:
@tsmaeder what do you think about [async-mutex](https://www.npmjs.com/package/async-mutex)? In your case you might be interested by the `Lock` class. I am interested by the `Semaphore` class for [one of my PRs](https://github.com/eclipse-theia/theia/pull/9820) (in order to limit concurrency).

packages/task/src/browser/task-service.ts
paul-marechal: I guess in this specific scenario, the promise chain pattern is not the most adequate since we need a mechanism to release early. Agreed that a dedicated class sounds good here. See https://github.com/eclipse-theia/theia/pull/9858#issuecomment-897840612.

packages/core/src/common/lock.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import { expect } from 'chai';
+import { Lock } from './lock';
+import { wait } from './promise-util';
+
+describe('Lock', () => {
+
+    beforeEach(() => {
+
+    });
```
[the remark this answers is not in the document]
tsmaeder: 🤦

tsmaeder:
@paul-marechal I'm not sure: I'm relying on the fact that I can multi-release the same Lock multiple times without it causing an error. The doc of async-mutex does not mention that. Out of curiosity, how does a semaphore (aka n parallel tasks) make sense in the single-threaded environment?

paul-marechal:
> I'm relying on the fact that I can multi-release the same Lock multiple times without it causing an error.

I don't see that being relied upon within the code? Perhaps I am missing it? But from what I understand calling `async-lock`'s acquired `release` function multiple times only releases once. We can always open a PR to add that missing documentation upstream.

> [...] how does a semaphore (aka n parallel tasks) make sense in the single-threaded environment?

Semaphore apply to any concurrent system. Note that concurrency can be achieved without parallelism. Node's async tasks are very much concurrent, that's the whole selling point of Node :) See [this](https://nodejs.org/en/docs/guides/blocking-vs-non-blocking/#concurrency-and-throughput) or [that](https://stackoverflow.com/questions/21438207/can-node-js-code-result-in-race-conditions).

In my case I would spawn X amount of upload promises, and the semaphore would limit the amount uploaded by those concurrent tasks so that only a fewer amount Y is actually uploading at any given time. The semaphore would guard the "upload budget count" shared resource.

tsmaeder:
> In my case I would spawn X amount of upload promises, and the semaphore would limit the amount uploaded by those concurrent tasks so that only a fewer amount Y is actually uploading at any given time.

But why: doesn't the I/O of the upload block the promises from proceeding just as effectively as waiting on the semaphore?

paul-marechal:
The I/O is happening in parallel in my case, so the semaphore would help throttle the logic. Right now I also had to implement my own class to handle only X tasks at a given time, but semaphores from `async-mutex` might be more adequate than re-implementing something equivalent in Theia.

<details>

<summary>A few lines of code might be worth a thousand words:</summary>

---

I invite you to open a new empty tab, open the dev tools, open the network tab and/or the console and paste the following snippets. They each do 16 requests to some random API, but one is doing the requests in parallel and the other not.

```ts
// parallel
(async function() {
    console.log('start');
    const start = Date.now();
    const jokes = await Promise.all(new Array(16).fill().map(async (_, i) => {
        const response = await fetch('https://v2.jokeapi.dev/joke/Any?type=single');
        const { joke = 'something went wrong' } = await response.json();
        return joke;
    }));
    const end = Date.now();
    console.log('jokes:', jokes);
    console.log(`end (took: ${end - start}ms)`);
})();
```

```ts
// sequential
(async function() {
    console.log('start');
    const start = Date.now();
    const jokes = [];
    for (let i = 0; i < 16; i++) {
        const response = await fetch('https://v2.jokeapi.dev/joke/Any?type=single');
        const { joke = 'something went wrong' } = await response.json();
        jokes.push(joke);
    }
    const end = Date.now();
    console.log('jokes:', jokes);
    console.log(`end (took: ${end - start}ms)`);
})();
```

Promises are handles to some underlying operation. Said operation happens "outside of JS" meaning it can happen in parallel. JS code attached to a promise won't run in parallel of another JS handler code, but we still benefit from having the underlying process doing things in the background in parallel.

</details>

tsmaeder:
@paul-marechal but why throttle uploads? And if we _are_ trottling, what is the resource we're trying to conserve? I'm not questioning Semaphores in general, I'm asking about your specific case.

tsmaeder:
@colin-grant-work I'm kinda split down the middle about using a library: async-lock is a no-go for me since it's a not being maintained. Async mutex seems weird: it has two interface classes that basically do the same thing (a mutex is really just a semaphore with n=1).  I'm just no sure there is enough meat there to not write the utility ourselves.

colin-grant-work:
@tsmaeder, I'm certainly not wedded to either of those, and this is a simple-enough utility that I don't think we need to worry too much about missing subtleties. It's one I've been tempted to write [in the past](https://github.com/colin-grant-work/theia/commit/0dd1ba1643b40d861d47eb5bdbc9da24a9aa9fc2) to handle synchronous calls to update preferences, and I think your implementation handles that case, as well, and with the added feature of safe multiple-release.

paul-marechal:
@tsmaeder in my case I noticed that not throttling uploads on my environment takes more time than when throttled. But from what I've seen it's a common thing to do to not hammer down servers with too many concurrent requests.

tsmaeder:
I've changed the code to use async-mutex.

packages/core/src/common/promise-util.ts
```diff
@@ -53,9 +53,29 @@ export async function retry<T>(task: () => Promise<T>, delay: number, retries: n
...
+    return value => new Promise((resolve, reject) => { setTimeout(() => resolve(value), ms); });
+}
+
+/**
+ * Constructs a promise that will resolve after a given delay.
+ * @param ms the number of milliseconds to wait
+ */
+export async function wait(ms: number): Promise<void> {
+    await delay(ms)(undefined);
+}
```
paul-marechal, line 81: Do we still need this?

paul-marechal (approved):
I confirm that I am unable to start the same task multiple times anymore, following the "How to test" instructions.

Code LGTM.

packages/core/src/common/promise-util.ts
tsmaeder: Not strictly, but I find it a useful debugging aid and therefore the couple of lines of code.

Issue #9768, opened 2021-07-22
Codicon spin animation


vincejo:
### Feature Description: 
When we do `vscode.ThemeIcon('loading~spin')`, we noticed that the icon does not spin. In VSCode the animation modifier works. It would be great to have in Theia as well. We are in Theia 1.14.0.

vince-fugnitto:
@vincentjocodes thank you for the issue, is it something you’re interested in contributing?

vince-fugnitto:
I believe this is fixed.

PR #9436, opened 2021-05-05
Transpile theia to es6

sgraband:
#### What it does

This is is a work-in-progress PR for transpiling Theia to es6. To continue supporting es5 based VSCode extensions, the `@es5ClassCompat` tag was used.

The tag was also applied to be used with Theia es5 plugins. This is however yet to be tested.

We would like to gather feedback while tackling the remaining issues, so feel free to comment on this PR!

Open issues:
- [ ] The `@es5ClassCompat` tag is not working for all of the classes in `types-impl.ts`.
The tag is failing on `Location` and `DiagnosticRelatedInformation` because of an `Cannot access Location before its initialization` error
Applying the tag to `CodeLens`, `TreeItem`, `ProgressExecution`, `ShellExecution` and `DebugAdapterExecutable` will result in a `Can't resolve @theia/plugin ERROR`.
- [ ] In eclipse-theia#5902 it was mentioned that there could be issues with `phosphorjs` and `inversifyjs`, however we could not find any regressions while testing. This may need some further research.
- [ ] Test support for es5 Theia plugins.

Contributed on behalf of STMicroelectronics
Signed-off-by: Simon Graband <sgraband@eclipsesource.com>
#### How to test

To test the support for es5 VSCode extensions, you can try out one of the extensions described here: https://github.com/microsoft/vscode/issues/68698.

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #9864, opened 2021-08-09
Replace icons with codicons

### LINKED ISSUE #6115: Update VSCode Icons
**Description**

The repository [vscode-icons](https://github.com/microsoft/vscode-icons) has recently been created where we can re-use the same icons present in VSCode. Is it something we'd like to align to? If we use the outdated icons it's quite possible we will get a mismatch in our own applications when using plugins (which bring the new icons in their views).

msujew:
#### What it does

Closes #6115

Does a lot of css work:

* Replaces all internally used icons (fontawesome, custom svg files) with codicons
* Adds a new utility method for codicons, namely `codicon` and `codiconArray`.
* Aligns button-icon css with vscode: Buttons in toolbars, etc. are now highlighted with a backdrop

I decided to leave the old svg icons in the codebase for backwards compatibility. I also couldn't notice any breaking change, but please keep an eye out for them.

#### How to test

Before starting to test I would recommend to start a second Theia instance to compare visual changes. The following testing steps are only the most important parts that have changed. As every icon in the app has changed, please refer to the code to gather a list of everything that has changed. Also, change the icon theme to `seti` at appropriate times to check for tab-bar items, etc.

* Check toolbar icons, every icon should show a backdrop on hover.
* Check tabbar close icons, every close icon should show a backdrop on hover.
* Check status bar icons
* Check notification widget icons
* Check debug view widget icons
* Check scm view widget icons
* Check problems view widget icons
* Check extensions view widget icons

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/callhierarchy/src/browser/callhierarchy-tree/callhierarchy-tree-widget.tsx
+++ b/packages/callhierarchy/src/browser/callhierarchy-tree/callhierarchy-tree-widget.tsx
@@ -17,7 +17,7 @@
 import { injectable, inject } from '@theia/core/shared/inversify';
 import {
     ContextMenuRenderer, TreeWidget, NodeProps, TreeProps, TreeNode,
-    TreeModel, DockPanel
+    TreeModel, DockPanel, codicon
 } from '@theia/core/lib/browser';
 import { LabelProvider } from '@theia/core/lib/browser/label-provider';
 import { DefinitionNode, CallerNode } from './callhierarchy-tree';
@@ -47,7 +47,7 @@ export class CallHierarchyTreeWidget extends TreeWidget {
         this.id = CALLHIERARCHY_ID;
         this.title.label = 'Call Hierarchy';
         this.title.caption = 'Call Hierarchy';
-        this.title.iconClass = 'fa call-hierarchy-tab-icon';
+        this.title.iconClass = codicon('references');
         this.title.closable = true;
         this.addClass(HIERARCHY_TREE_CLASS);
         this.toDispose.push(this.model.onSelectionChanged(selection => {
--- a/packages/core/src/browser/label-provider.ts
+++ b/packages/core/src/browser/label-provider.ts
@@ -23,15 +23,16 @@ import { Event, Emitter, Disposable, Path } from '../common';
 import { FrontendApplicationContribution } from './frontend-application';
 import { EnvVariablesServer } from '../common/env-variables/env-variables-protocol';
 import { ResourceLabelFormatter, ResourceLabelFormatting } from '../common/label-protocol';
+import { codicon } from './widgets';
 
 /**
  * @internal don't export it, use `LabelProvider.folderIcon` instead.
  */
-const DEFAULT_FOLDER_ICON = 'fa fa-folder';
+const DEFAULT_FOLDER_ICON = `${codicon('folder')} default-folder-icon`;
 /**
  * @internal don't export it, use `LabelProvider.fileIcon` instead.
  */
-const DEFAULT_FILE_ICON = 'fa fa-file';
+const DEFAULT_FILE_ICON = `${codicon('file')} default-file-icon`;
 
 /**
  * Internal folder icon class for the default (File Icons) theme.
--- a/packages/core/src/browser/progress-status-bar-item.ts
+++ b/packages/core/src/browser/progress-status-bar-item.ts
@@ -73,7 +73,7 @@ export class ProgressStatusBarItem implements ProgressClient {
             this.statusBar.removeElement(this.id);
             return;
         }
-        const text = `$(refresh~spin) ${message}`;
+        const text = `$(codicon-sync~spin) ${message}`;
         this.statusBar.setElement(this.id, {
             text,
             alignment: StatusBarAlignment.LEFT,
--- a/packages/plugin-ext/src/main/browser/dialogs/modal-notification.ts
+++ b/packages/plugin-ext/src/main/browser/dialogs/modal-notification.ts
@@ -15,7 +15,7 @@
  ********************************************************************************/
 import { injectable } from '@theia/core/shared/inversify';
 import { Message } from '@theia/core/shared/@phosphor/messaging';
-import { Key } from '@theia/core/lib/browser';
+import { codiconArray, Key } from '@theia/core/lib/browser';
 import { AbstractDialog } from '@theia/core/lib/browser/dialogs';
 import '../../../../src/main/browser/dialogs/style/modal-notification.css';
 import { MainMessageItem } from '../../../common/plugin-api-rpc';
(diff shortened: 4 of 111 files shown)


vince-fugnitto (commented):
@msujew I believe we should also update the widget icons themselves (ex: `explorer`, `scm`, ...), any reason we wanted to omit it initially?

msujew:
> any reason we wanted to omit it initially?

Actually no, I just stopped thinking about it, once I noticed that the explorer icon is not in the codicon collection. But I just saw that every other icon is in there. So I'll replace those. What is your opinion on replacing the default folder/file icon in the explorer? Currently it still uses the fontawesome icons.

vince-fugnitto:
> Actually no, I just stopped thinking about it, once I noticed that the explorer icon is not in the codicon collection. But I just saw that every other icon is in there. So I'll replace those. What is your opinion on replacing the default folder/file icon in the explorer? Currently it still uses the fontawesome icons.

I think the idea for consistency with vscode and what plugins use we should aim to update all icons (including font-awesome icons) to their codicon counterparts. This includes widgets when docked to sidepanels. It will create a uniform look and feel across the framework.

packages/callhierarchy/src/browser/callhierarchy-tree/callhierarchy-tree-widget.tsx
```diff
@@ -177,7 +177,8 @@ export class CallHierarchyTreeWidget extends TreeWidget {
 
     private doOpenEditor(uri: DocumentUri, range: Range, keepFocus: boolean): void {
         this.editorManager.open(
-            new URI(uri), {
+            new URI(uri),
+            {
```
vince-fugnitto, line 181: minor: unrelated change.

packages/core/src/browser/label-provider.ts
```diff
@@ -27,7 +27,7 @@ import { ResourceLabelFormatter, ResourceLabelFormatting } from '../common/label
 /**
  * @internal don't export it, use `LabelProvider.folderIcon` instead.
  */
-const DEFAULT_FOLDER_ICON = 'fa fa-folder';
+const DEFAULT_FOLDER_ICON = '';
```
vince-fugnitto, line 30: Can you confirm why we want the default folder icon to be non-existent?

packages/core/src/browser/progress-status-bar-item.ts
```diff
@@ -73,7 +73,7 @@ export class ProgressStatusBarItem implements ProgressClient {
             this.statusBar.removeElement(this.id);
             return;
         }
-        const text = `$(refresh~spin) ${message}`;
+        const text = `$(codicon-sync~spin) ${message}`;
```
vince-fugnitto, line 76: This reminds me of the following issue https://github.com/eclipse-theia/theia/issues/9768, perhaps it's something we can easily implement in the future.

packages/plugin-ext/src/main/browser/dialogs/modal-notification.ts
```diff
@@ -60,7 +60,7 @@ export class ModalNotification extends AbstractDialog<string | undefined> {
         const iconContainer = messageNode.appendChild(document.createElement('div'));
         iconContainer.classList.add(ICON);
         const iconElement = iconContainer.appendChild(document.createElement('i'));
-        iconElement.classList.add('fa', this.toIconClass(messageType), 'fa-fw', messageType.toString());
+        iconElement.classList.add(... this.toIconClass(messageType), 'fa-fw', messageType.toString());
```
vince-fugnitto, line 63: nitpick: the `...` formatting looks a bit off.

packages/core/src/browser/label-provider.ts
msujew: Because that's the way it works in vscode (default icon theme). I wanted to align with that.

vince-fugnitto (commented):
If we previously had layout data saved (ex: local-storage), then starting the app may cause icons to be non-existent:

image


We may want to introduce a layout-migration:

https://github.com/eclipse-theia/theia/blob/1dba01446fd7aef464bcaef5f853d6f595f96bbd/packages/core/src/browser/shell/application-shell.ts#L51-L62

https://github.com/eclipse-theia/theia/blob/1dba01446fd7aef464bcaef5f853d6f595f96bbd/packages/core/src/browser/shell/shell-layout-restorer.ts#L87-L110

packages/core/src/browser/label-provider.ts
vince-fugnitto: I believe this is for our custom `file-icons` theia theme, not the vscode theme:

_before_:

image

_after_:

image


I don't think we should break, if users want the vscode experience they would use `seti` (default in vscode):

image

msujew: I see. Would you recommend to keep the old `fa` icon then? Or rather replace it with the codicon as well?

vince-fugnitto: Up to you, I think `codicon folder` would look nice :+1:, and perhaps in the future we can think of deprecating font-awesome in the framework altogther.

msujew:
> We may want to introduce a layout-migration

Alright, done. It's a bit crude, but I found the string replacement method to be sufficient.

packages/core/src/browser/label-provider.ts
msujew: Alright, replaced them with codicons 👍

vince-fugnitto (commented):
@msujew do you mind rebasing the pull-request when you get a chance? I'd be happy to take a look at it afterwards :)

vince-fugnitto (commented):
I really love the new icons and behavior for toolbar items :+1: 
I noticed two things when testing:
- I believe the bottom sidebar (settings icon) may need to be adjusted to have the same `font-size` as the rest of the sidebar
- I believe the search-in-workspace widget needs to be updated to take into account the new chevron (left is the pr, and right is master), the input fields are no longer aligned:

siw-diff

- dirty editors are no longer marked as dirty:

image

msujew:
@vince-fugnitto Thanks for the feedback, I think I addressed everything 👍

vince-fugnitto (commented):
I'd be fine to merge and fix issues throughout the month if necessary as I do not want to hold back the pull-request :+1: 


I still notice some UI issues, perhaps one of the bigger ones (that may be caused by a bigger issue) is that the plugin-views sometimes do not display their proper icons:

image

I also noticed that the alignment of toolbar items contributed by plugins are not necessarily aligned properly as well (compared to builtin views like the explorer):

image


A very minor bug is how the scm arrow is not aligned properly (for `CHANGES`) compared to other trees:

image

msujew:
> I still notice some UI issues, perhaps one of the bigger ones (that may be caused by a bigger issue) is that the plugin-views sometimes do not display their proper icons

This seems to be unrelated to my changes, see the current `master`:

image

I noticed that the `instanceof` check in the plugin host fails:

https://github.com/eclipse-theia/theia/blob/7442150b335a9280a314dbf30f6a9227305b6b04/packages/plugin-ext/src/plugin/tree/tree-views.ts#L372

I assume this is due to the ES2017 upgrade and the `@es5Compatibility` annotation (#9436). `instanceof` checks are not a valid operation on es5 classes as far as I can tell.

I'll add an `is` check for `ThemeIcon` users instead.

vince-fugnitto:
@msujew sorry I had thought I confirmed against master that it was a regression (turns out the code I confirmed against was older) but you're right that its not related to your changes and we can omit the fix for now :+1:

vince-fugnitto (approved):

I think there is a leftover file:

image

msujew:
I'll merge this Tuesday or Wednesday, if no one else has any issues with my changes.

PR #9878, opened 2021-08-10
doc: runtime support policy

paul-marechal:
Document a new policy about our Node.js and Electron version support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,15 +27,15 @@
 
 </div>
 
-- [**Website**](#website)
-- [**Scope**](#scope)
-- [**Roadmap**](#roadmap)
-- [**Getting Started**](#getting-started)
-- [**Contributing**](#contributing)
-- [**Feedback**](#feedback)
-- [**Documentation**](#documentation)
-- [**License**](#license)
-- [**Trademark**](#trademark)
+- [Website](#website)
+- [Scope](#scope)
+- [Roadmap](#roadmap)
+- [Getting Started](#getting-started)
+- [Contributing](#contributing)
+- [Feedback](#feedback)
+- [Documentation](#documentation)
+- [License](#license)
+- [Trademark](#trademark)
 
 ## Website
 
@@ -82,6 +82,7 @@ Read below how to engage with Theia community:
 
 ## Documentation
 
+- [Runtime Support Policy (Node/Electron)](./doc/runtime-policies.md)
 - [API Documentation](https://eclipse-theia.github.io/theia/docs/next/index.html)
 - Useful Links:
   - [Developing](doc/Developing.md)
@@ -100,4 +101,3 @@ Read below how to engage with Theia community:
 ## Trademark
 "Theia" is a trademark of the Eclipse Foundation
 https://www.eclipse.org/theia
-
--- a/doc/runtime-support-policy.md
+++ b/doc/runtime-support-policy.md
@@ -1,0 +1,42 @@
+# Node.js
+
+## Version Support Policy
+
+We try to support two (2) Node.js versions up to the current _Active LTS_ version.
+
+> For example:
+> - Node.js 14 is _Active LTS_: we'll support both 14 and 12 and drop 10.
+> - Node.js 16 is _Active LTS_: we'll support both 16 and 14 and drop 12.
+
+See https://nodejs.org/en/about/releases/ to see the status of Node.js versions.
+
+We recommend setting up your environments to run Theia using the Node.js _Active LTS_, but any supported version should work as well. Fill an issue otherwise: https://github.com/eclipse-theia/theia/issues/new/choose.
+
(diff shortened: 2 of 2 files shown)


paul-marechal:
@JonasHelming you proposed supporting up to two Node.js versions at a time, but I was thinking we could instead expand the support from our current Node.js version (12) and extend it up only.

Doing so might sound tricky as we might write JS code compatible with 12 which will also run in newer versions, but we have to keep an eye out on our dependencies because one of those could start requiring >=14 or something else.

In practice, unless there's something we are itching to use from newer Node.js versions, we could keep the baseline at Node.js 12 and only move the upper bound by updating the CI to run on newer Node.js versions.

The main benefit I see doing so is when you are making applications that depend on the framework _and_ other packages: By making Theia compliant with "older" versions and up, it makes it easier to consume other packages that may have stricter constraints. In such cases Theia should not be the bottleneck and allow you to choose the Node.js version adequate for your dependency tree and environment.

Of course there will be issues if someone tries to use a package that requires <12, but this is to be expected.

README.md
```diff
@@ -27,15 +27,15 @@
...
-- [**Trademark**](#trademark)
+- [Website](#website)
+- [Scope](#scope)
+- [Roadmap](#roadmap)
+- [Getting Started](#getting-started)
+- [Contributing](#contributing)
+- [Feedback](#feedback)
+- [Documentation](#documentation)
+- [License](#license)
+- [Trademark](#trademark)
```
paul-marechal, line 38: Auto-format doesn't like the bold links, fine to simplify it this way or should I revert?

vince-fugnitto: I think it's fine :+1:

README.md
```diff
@@ -82,6 +82,7 @@ Read below how to engage with Theia community:
 
 ## Documentation
 
+- [Runtime Support Policy (Node/Electron)](./doc/runtime-policies.md)
```
vince-fugnitto, line 85: Not sure if it should be here or under `useful links`, I'll let you decide:

```suggestion
- [Runtime Support Policy (Node/Electron)](doc/runtime-policies.md)
```

doc/runtime-support-policy.md
```diff
@@ -0,0 +1,42 @@
...
+
+We try to support two (2) Node.js versions up to the current _Active LTS_ version.
+
+> For example:
+> - Node.js 14 is _Active LTS_: we'll support both 14 and 12 and drop 10.
+> - Node.js 16 is _Active LTS_: we'll support both 16 and 14 and drop 12.
+
+See https://nodejs.org/en/about/releases/ to see the status of Node.js versions.
+
+We recommend setting up your environments to run Theia using the Node.js _Active LTS_, but any supported version should work as well. Fill an issue otherwise: https://github.com/eclipse-theia/theia/issues/new/choose.
```
vince-fugnitto, line 13:
```suggestion
We recommend setting up your environment to run Theia using the Node.js _Active LTS_, but any supported version should work as well. Fill an issue otherwise: https://github.com/eclipse-theia/theia/issues/new/choose.
```

vince-fugnitto, line 13: nitpick, either:

```
Please submit an issue...
```

```
Please fill out an issue...
```

```
Please file an issue...
```

doc/runtime-support-policy.md
```diff
@@ -0,0 +1,42 @@
...
+
+Note that the Node.js version you should use depends on your own project's dependencies: packages others than Theia might have their own requirements, so we try to support a reasonable range for adopters to be able to satisfy such constraints.
```
vince-fugnitto, line 15:
```suggestion
Note that the Node.js version you should use depends on your own project's dependencies: packages other than Theia might have their own requirements, so we try to support a reasonable range for adopters to be able to satisfy such constraints.
```

doc/runtime-support-policy.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Update Process
+
+- Follow Node.js LTS cadence and initiate the update when a new Node.js version becomes _Active LTS_.
+- Use `@types/node` for the oldest supported Node version (backward compatibility).
+- Update the CI matrix to include the new Node.js version to support.
+- Update the CHANGELOG.
```
vince-fugnitto, line 22: Do we need to mention what we should do when deprecating a node runtime?
Ex: updating the `@types/node`, `breaking change`, removing old node in matrix.

doc/runtime-support-policy.md
```diff
@@ -0,0 +1,42 @@
...
+
+# Electron
+
+## Version Support Policy
+
+We aim at using Electron's latest _Stable Release_.
```
vince-fugnitto, line 28: nitpick:

```suggestion
We aim to use Electron's latest _Stable Release_.
```

Not sure if we should mention `use`, `target`, or `support`. Whichever you feel is best.

doc/runtime-support-policy.md
paul-marechal: I'll wait for more feedback on what's currently presented.

PR #8626, opened 2020-10-13
Allow extenders to modify preferences

westbury:
What it does

These changes add the ability for packages to modify the preference schema for preferences added by others.  This allows extenders to make modifications to the schema for preferences added either by other packages or by plugins.  See https://community.theia-ide.org/t/hidden-settings/821 .

An additional property has been added to the preference schema: 'hidden'.  This allows extenders to remove a preference altogether from the UI.  It won't show in the preference view nor will it be included in the counts shown in that view.

Note that when a preference is created, the code reading that preference is going to assume that the preference exists and that the value meets the constraints specified in the schema.  Therefore the preference is still there, even if hidden.  Furthermore changes cannot violate the original schema.  The following changes are allowed:

- the description can be changed.
- if an enum, the enum list can be replaced by a subset.  However extra values cannot be added to the enum because that might break readers of the preference.
- if a string preference then an enum can be set to limit the string to certain values
- a minimum value can be set for numbers.  However if the preference already has a minimum then the new minimum cannot be less.
- the 'hidden' flag can be turned on or off.
- the default value can be changed (note that this can also be changed in packages.json)

#### How to test

Checkout branch modifying-prefs-test.  This branch contains an extra commit with various preferences modified.

Some of these modifications are to preferences in the git built-in plugin.  Start up without the git plugin, then install it.  You will see the plugin's preferences but with the modifications.  Uninstall the plugin and its preferences disappear.  Install the plugin and they come back again with the modifications.

Modifications always override the original preference schema, even if there is no package dependency.  However package dependency does prioritize appropriately where multiple packages are modifying the same preference.
  
#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #9883, opened 2021-08-11
preferences: make schemas extensible

vince-fugnitto:
#### What it does

The pull-request introduces a new a new abstract class `PreferencesSchemaProvider` which is used to provide preference schema definitions. The class is ultimately used to bind preference schemas in an extensible way instead of to constant values.

This allows extenders to easily modify preference schemas as illustrated in the commit https://github.com/eclipse-theia/theia/commit/c5a7d712d21a7f89eb03c03244ac1b91f38d02a6.

#### How to test

1. ensure that the application successfully builds and starts.
2. start the application, and navigate to the preferences-view - preferences should be correctly listed.
3. search for `workspace.preserveWindow`, the preference definition should not be listed as it was removed in https://github.com/eclipse-theia/theia/commit/c5a7d712d21a7f89eb03c03244ac1b91f38d02a6 for test purposes.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/scm/src/browser/scm-preferences.ts
+++ b/packages/scm/src/browser/scm-preferences.ts
@@ -43,18 +43,20 @@ export interface ScmConfiguration {
     'scm.defaultViewMode': 'tree' | 'list'
 }
 
+export const ScmPreferenceContribution = Symbol('ScmPreferenceContribution');
 export const ScmPreferences = Symbol('ScmPreferences');
 export type ScmPreferences = PreferenceProxy<ScmConfiguration>;
 
-export function createScmPreferences(preferences: PreferenceService): ScmPreferences {
-    return createPreferenceProxy(preferences, scmPreferenceSchema);
+export function createScmPreferences(preferences: PreferenceService, schema: PreferenceSchema = scmPreferenceSchema): ScmPreferences {
+    return createPreferenceProxy(preferences, schema);
 }
 
 export function bindScmPreferences(bind: interfaces.Bind): void {
     bind(ScmPreferences).toDynamicValue((ctx: interfaces.Context) => {
         const preferences = ctx.container.get<PreferenceService>(PreferenceService);
-        return createScmPreferences(preferences);
+        const contribution = ctx.container.get<PreferenceContribution>(ScmPreferenceContribution);
+        return createScmPreferences(preferences, contribution.schema);
     }).inSingletonScope();
-
-    bind(PreferenceContribution).toConstantValue({ schema: scmPreferenceSchema });
+    bind(ScmPreferenceContribution).toConstantValue({ schema: scmPreferenceSchema });
+    bind(PreferenceContribution).toService(ScmPreferenceContribution);
 }
--- a/packages/core/src/browser/core-preferences.ts
+++ b/packages/core/src/browser/core-preferences.ts
@@ -114,17 +114,20 @@ export interface CoreConfiguration {
     'workbench.tree.renderIndentGuides': 'onHover' | 'none' | 'always';
 }
 
+export const CorePreferenceContribution = Symbol('CorePreferenceContribution');
 export const CorePreferences = Symbol('CorePreferences');
 export type CorePreferences = PreferenceProxy<CoreConfiguration>;
 
-export function createCorePreferences(preferences: PreferenceService): CorePreferences {
-    return createPreferenceProxy(preferences, corePreferenceSchema);
+export function createCorePreferences(preferences: PreferenceService, schema: PreferenceSchema = corePreferenceSchema): CorePreferences {
+    return createPreferenceProxy(preferences, schema);
 }
 
 export function bindCorePreferences(bind: interfaces.Bind): void {
     bind(CorePreferences).toDynamicValue(ctx => {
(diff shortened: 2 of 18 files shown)


examples/api-samples/src/browser/api-samples-frontend-module.ts
```diff
@@ -24,13 +24,28 @@ import { bindSampleFileWatching } from './file-watching/sample-file-watching-con
...
+@injectable()
+class TestWorkspacePreferencesSchemaProvider extends WorkspacePreferencesSchemaProvider {
+
+    get schema(): PreferenceSchema {
+        const schema = workspacePreferenceSchema;
+        delete schema.properties['workspace.preserveWindow'];
+        return schema;
+    }
+
+}
```
vince-fugnitto, line 51: Note: the following commit https://github.com/eclipse-theia/theia/commit/c5a7d712d21a7f89eb03c03244ac1b91f38d02a6 will be dropped (which includes this change). The change is present to illustrate that we can now extend/modify schemas.

paul-marechal: Hmm... If you are going to delete a field on an imported object like that you don't even need to rebind anything. That's basically monkey-patching: since you are mutating a shared reference it would update all other references as well.

packages/scm/src/browser/scm-preferences.ts
```diff
@@ -46,15 +46,21 @@ export interface ScmConfiguration {
...
+
+export function createScmPreferences(preferences: PreferenceService, schema: PreferenceSchema = scmPreferenceSchema): ScmPreferences {
+    return createPreferenceProxy(preferences, schema);
 }
 
 export function bindScmPreferences(bind: interfaces.Bind): void {
     bind(ScmPreferences).toDynamicValue((ctx: interfaces.Context) => {
         const preferences = ctx.container.get<PreferenceService>(PreferenceService);
-        return createScmPreferences(preferences);
+        const container = ctx.container.get<PreferenceContribution>(ScmPreferenceContribution);
```
paul-marechal, line 61: `contribution`

vince-fugnitto:
@paul-marechal @colin-grant-work I refactored and simplified the pull-request to use `Symbol` instead when binding :+1:

westbury:
@vince-fugnitto thanks for resurrecting this issue about overriding preferences.  We've have been using a patched version of Theia for quite some time now so that we can overwrite preferences.

This is a good and simple way of allowing extenders to override preferences and would support a lot of our use cases.  My main concerns are:
1. I don't think this will support the override of preferences that are provided by plugins.
2. This puts responsibility on the extender to ensure that modifications to the preferences do not break the requirements of the code.  If changes to the preference schema result in preferences that do not match what is in the interface (eg WorkspaceConfiguration) then we could have problems.  For example, the code will assume that a preference exists and there must be a default value that is returned even if the preference has been 'removed'.  Perhaps this is ok.
3. We cannot remove a preference from the UI but still allow it to be changed programmatically, unless I am missing something. 

If you prefer this approach to the approach in https://github.com/eclipse-theia/theia/pull/8626 that is fine but I will then need to amend that PR so that we can make our changes to preferences in plugins.

vince-fugnitto:
@westbury thank you for the discussion, in the end I believe the change would be something like the two pull-requests combined (easier to rebind, possibility to hide). 

> 1. I don't think this will support the override of preferences that are provided by plugins.

Correct, the pull-request aims at making the rebinding of individual `PreferenceContribution` more extensible from `@theia` extensions. There was a question in our community regarding the difficulties of rebinding a schema which also required additional workarounds to rebind the `PreferenceContribution`:

- https://github.com/eclipse-theia/theia/discussions/9880
- https://github.com/eclipse-theia/theia/discussions/9144





> 2. This puts responsibility on the extender to ensure that modifications to the preferences do not break the requirements of the code.  If changes to the preference schema result in preferences that do not match what is in the interface (eg WorkspaceConfiguration) then we could have problems.  For example, the code will assume that a preference exists and there must be a default value that is returned even if the preference has been 'removed'.  Perhaps this is ok.

I believe references to preferences should be fine, the `preference-service` will handle getting preference values which do not exist, returning `undefined`. If there happens to be issues, I believe it would be the responsibility of the application or extension developer to handle them.

> 3. We cannot remove a preference from the UI but still allow it to be changed programmatically, unless I am missing something.

The possibility to `hide` a preference would be achieved by your pull-request I believe (only UI). I add the possibility of modifying the schema altogether in a downstream extension. If a preference no longer exists it will of course not be visible, or editable by end-users.

paul-marechal (approved):
The refactoring is minimal and seems to make sense.

colin-grant-work:
> The possibility to `hide` a preference would be achieved by your pull-request I believe (only UI). I add the possibility of modifying the schema altogether in a downstream extension. If a preference no longer exists it will of course not be visible, or editable by end-users.

@westbury is correct that his PR has this feature which this PR does not - keeping a preference alive for programmatic manipulation but (somewhat) hidden from users - but I don't think that's an argument against this PR. That is a separate feature that can be implemented on top of this PR.

vince-fugnitto:
I dropped the test commit (which performs the rebind for test-purposes). I'll merge once CI is green :+1:

PR #9890, opened 2021-08-11
Added support for yarn aliases

### LINKED ISSUE #9856: Handling of yarn aliases in the application package generator
### Feature Description:
[`yarn` offers the possibility of adding a package with an alias](https://classic.yarnpkg.com/en/docs/cli/add/#toc-yarn-add-alias). When this is used a dependency entry in `package.json` is create with the following format:

```json
    "<alias>": "npm:<original_name>@<version>",
```

If this feature is used, the source generator in `@theia/application_package` breaks because `<original_name>` is used in place of `<alias` in the generated files. This happens because the information inside each package's `package.json` is used and the alias is ignored. This breaks every place where an import or a require like `import * from <alias>` is used.

Are you interested in a PR to fix this issue? I can work on it myself.

federicobozzini:
#### What it does

This PR fixes #9856 . It is done by changing the application package extension to use as a name for the packages (and more specifically theia extensions) the name used in the list of dependencies in `package.json` in place of the package name. In most cases these two names should match, but when yarn aliases are used they differ.

#### How to test

I've tested this PR by adding a Theia extension using a yarn alias. After that it should be possible to build Theia and import the extension exported objects using its alias. As a comparison, if a yarn alias is used without this PR the building phase of the examples fails.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-package/src/application-package.ts
+++ b/dev-packages/application-package/src/application-package.ts
@@ -112,7 +112,7 @@ export class ApplicationPackage {
     get extensionPackages(): ReadonlyArray<ExtensionPackage> {
         if (!this._extensionPackages) {
             const collector = new ExtensionPackageCollector(
-                raw => this.newExtensionPackage(raw),
+                (raw: PublishedNodePackage, alias: string) => this.newExtensionPackage(raw, alias),
                 this.resolveModule
             );
             this._extensionPackages = collector.collect(this.pck);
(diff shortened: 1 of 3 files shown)


federicobozzini:
It looks to me like in `dev-packages/application-package/src/extension-package.ts` and `dev-packages/application-package/src/application-package.ts` there are several exported method there were only used by the `extension-manager` extension prior to version `0.15.0`. Would it make sense to delete or deprecate those?

mcgordonite:
I've tested this with our browser Theia application. It generates the same server.js and index.js files as before without aliases, and uses the aliased name when an alias is added. Thanks @federicobozzini!

dev-packages/application-package/src/application-package.ts
```diff
@@ -130,11 +130,11 @@ export class ApplicationPackage {
 
     async resolveExtensionPackage(extension: string): Promise<ExtensionPackage | undefined> {
```
paul-marechal, line 131: Add a comment to define `extension` as `the name of the extension's package as defined in "dependencies" (might be aliased)`.

dev-packages/application-package/src/application-package.ts
```diff
@@ -130,11 +130,11 @@ export class ApplicationPackage {
...
         const raw = await RawExtensionPackage.view(this.registry, extension);
-        return raw ? this.newExtensionPackage(raw) : undefined;
+        return raw ? this.newExtensionPackage(raw, extension) : undefined;
     }
 
-    protected newExtensionPackage(raw: PublishedNodePackage): ExtensionPackage {
-        return new ExtensionPackage(raw, this.registry);
+    protected newExtensionPackage(raw: PublishedNodePackage, name: string): ExtensionPackage {
+        return new ExtensionPackage(raw, this.registry, name);
```
paul-marechal, line 137: Rename `name` to `alias`? It can be confusing why we need to pass this string around instead of doing `raw.name`.

We could update the other locations in the same way.

paul-marechal (commented):
The idea makes sense and code mostly LGTM, just a few comments.

federicobozzini:
@paul-marechal Thanks for the suggestions. I've made some changes and added some comments to clarify the intent of the code.

paul-marechal:
I wasn't sure if it was ok with adding new required parameters to functions that could be imported by dependents. So I went ahead and pushed a commit to make it not breaking, just in case... If it still works for you then I'll merge, thanks!

federicobozzini:
I've tested this change with this code and works flawlessly :+1:

paul-marechal (approved):

PR #9909, opened 2021-08-16
Add support for sub heading in extension views

### LINKED ISSUE #9903: Implement vscode TreeView description
### Feature Description:

I would like to implement TreeView.description from the vscode api. This appears as a small subheading next to the panel title.

https://code.visualstudio.com/api/references/vscode-api#TreeView%3CT%3E

aaronrhodes:
#### What it does

Adds the description field to TreeView from the vscode API.

resolves #9903 

#### How to test

Add a plugin that utilises the TreeView description field.

There is a modified version of the vscode tree-view-sample available in the following repo:

https://github.com/aaronrhodes/vscode-extension-samples

build, and copy or symlink this example into the plugins directory and select a file using the "file explorer" panel in the explorer tab.

Screenshot 2021-08-16 at 11 03 57

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/style/view-container.css
+++ b/packages/core/src/browser/style/view-container.css
@@ -75,10 +75,19 @@
 }
 
 .theia-view-container .part > .header .label {
+    flex: 0;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+}
+
+.theia-view-container .part > .header .description {
     flex: 1;
     overflow: hidden;
     white-space: nowrap;
     text-overflow: ellipsis;
+    padding-left: var(--theia-ui-padding);
+    text-transform: none;
+    opacity: 0.6;
 }
 
 .theia-view-container .part > .body {
--- a/packages/core/src/browser/view-container.ts
+++ b/packages/core/src/browser/view-container.ts
@@ -47,6 +47,17 @@ export class ViewContainerIdentifier {
     progressLocationId?: string;
 }
 
+export interface DescriptionWidget {
+    description: string;
+    onDidChangeDescription: Emitter<void>;
+}
+
+export namespace DescriptionWidget {
+    export function is(arg: Object | undefined): arg is DescriptionWidget {
+        return !!arg && typeof arg === 'object' && 'onDidChangeDescription' in arg;
+    }
+}
+
 /**
  * A view container holds an arbitrary number of widgets inside a split panel.
  * Each widget is wrapped in a _part_ that displays the widget title and toolbar
@@ -673,6 +684,8 @@ export class ViewContainerPart extends BaseWidget {
     readonly onTitleChanged = this.onTitleChangedEmitter.event;
     protected readonly onDidFocusEmitter = new Emitter<this>();
     readonly onDidFocus = this.onDidFocusEmitter.event;
+    protected readonly onDidChangeDescriptionEmitter = new Emitter<void>();
+    readonly onDidChangeDescription = this.onDidChangeDescriptionEmitter.event;
 
     protected _collapsed: boolean;
 
@@ -699,6 +712,11 @@ export class ViewContainerPart extends BaseWidget {
         this.wrapped.title.changed.connect(fireTitleChanged);
         this.toDispose.push(Disposable.create(() => this.wrapped.title.changed.disconnect(fireTitleChanged)));
 
+        if (DescriptionWidget.is(this.wrapped)) {
+            const fireDescriptionChanged = () => this.onDidChangeDescriptionEmitter.fire(undefined);
+            this.toDispose.push(this.wrapped?.onDidChangeDescription.event(fireDescriptionChanged));
+        }
+
         const { header, body, disposable } = this.createContent();
         this.header = header;
         this.body = body;
@@ -709,6 +727,7 @@ export class ViewContainerPart extends BaseWidget {
             this.collapsedEmitter,
             this.contextMenuEmitter,
             this.onTitleChangedEmitter,
+            this.onDidChangeDescriptionEmitter,
             this.registerContextMenu(),
             this.onDidFocusEmitter,
             // focus event does not bubble, capture it
@@ -853,15 +872,29 @@ export class ViewContainerPart extends BaseWidget {
 
         const title = document.createElement('span');
         title.classList.add('label', 'noselect');
+
+        const description = document.createElement('span');
+        description.classList.add('description');
+
         const updateTitle = () => title.innerText = this.wrapped.title.label;
         const updateCaption = () => title.title = this.wrapped.title.caption || this.wrapped.title.label;
+        const updateDescription = () => {
+            description.innerText = DescriptionWidget.is(this.wrapped) && !this.collapsed && this.wrapped.description || '';
+        };
+
         updateTitle();
         updateCaption();
+        updateDescription();
+
         disposable.pushAll([
             this.onTitleChanged(updateTitle),
-            this.onTitleChanged(updateCaption)
+            this.onTitleChanged(updateCaption),
+            this.onDidChangeDescription(updateDescription),
+            this.onCollapsed(updateDescription),
         ]);
         header.appendChild(title);
+        header.appendChild(description);
+
         return {
             header,
             disposable
@@ -1002,6 +1035,7 @@ export namespace ViewContainerPart {
         collapsed: boolean;
         hidden: boolean;
         relativeSize?: number;
+        description?: string;
     }
 
     export function closestPart(element: Element | EventTarget | null, selector: string = 'div.part'): Element | undefined {
--- a/packages/plugin-ext/src/main/browser/view/plugin-view-widget.ts
+++ b/packages/plugin-ext/src/main/browser/view/plugin-view-widget.ts
@@ -22,6 +22,8 @@ import { ViewContextKeyService } from './view-context-key-service';
 import { StatefulWidget } from '@theia/core/lib/browser/shell/shell-layout-restorer';
 import { Message } from '@theia/core/shared/@phosphor/messaging';
 import { TreeViewWidget } from './tree-view-widget';
+import { DescriptionWidget } from '@theia/core/lib/browser/view-container';
+import { Emitter } from '@theia/core/lib/common';
 
 @injectable()
 export class PluginViewWidgetIdentifier {
@@ -30,7 +32,7 @@ export class PluginViewWidgetIdentifier {
 }
 
 @injectable()
-export class PluginViewWidget extends Panel implements StatefulWidget {
+export class PluginViewWidget extends Panel implements StatefulWidget, DescriptionWidget {
 
     @inject(MenuModelRegistry)
     protected readonly menus: MenuModelRegistry;
@@ -49,6 +51,8 @@ export class PluginViewWidget extends Panel implements StatefulWidget {
         this.node.tabIndex = -1;
         this.node.style.height = '100%';
     }
+
+    public onDidChangeDescription: Emitter<void> = new Emitter<void>();
 
     @postConstruct()
     protected init(): void {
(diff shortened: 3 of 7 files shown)


packages/core/src/browser/shell/description-widget.ts
```diff
@@ -0,0 +1,29 @@
+/********************************************************************************
+ * Copyright (C) 2017 TypeFox and others.
```
msujew, line 2: As you created this file yourself, please change the copyright notice to you/your company.

packages/core/src/browser/shell/description-widget.ts
```diff
@@ -0,0 +1,29 @@
...
+export interface DescriptionWidget {
+    description: string;
+    onDescriptionChange: Emitter<void>;
+}
+
+export namespace DescriptionWidget {
+    /* eslint-disable @typescript-eslint/no-explicit-any */
+    export function is(arg: any): arg is DescriptionWidget {
+        return arg !== undefined && !!arg['onDescriptionChange'];
+    }
```
msujew, line 28: Suggestion: We generally try not to use `any`, even in `is` methods. Change this to `Object | undefined` or `unknown`.

msujew (commented):
Thank you for your contribution, the changes look good to me so far. The description property does its job. 

FYI the title of your PR is a bit misleading, as you're adding support for the description property for views contributed by extensions, not only for outlines.

I have a few remarks before approving.

packages/core/src/browser/shell/description-widget.ts
vince-fugnitto, line 28: I wasn't so sure the file was necessary, and instead it could be directly added to the `tree-view`.
Adding it under `shell/` is also a bit awkward.

packages/plugin-ext/src/main/browser/view/plugin-view-widget.ts
```diff
@@ -112,6 +116,17 @@ export class PluginViewWidget extends Panel implements StatefulWidget {
...
     }
 
+    private _description: string = '';
+    get description(): string {
+        return this._description;
+    }
+
+    set description(description: string) {
+        this._description = description;
+        this.onDescriptionChange.fire();
```
vince-fugnitto, line 126: Should be moved under `this.updateWidgetDescription()`.

packages/plugin-ext/src/main/browser/view/plugin-view-widget.ts
```diff
@@ -121,14 +136,25 @@ export class PluginViewWidget extends Panel implements StatefulWidget {
...
     }
 
+    private updateWidgetDescription(): void {
+        const widget = this.widgets[0];
+        if (widget) {
+            if (DescriptionWidget.is(widget)) {
+                widget.description = this._description;
+            }
+        }
+    }
```
vince-fugnitto, line 146:
```suggestion
    private updateWidgetDescription(): void {
        const widget = this.widgets[0];
        if (widget) {
            if (DescriptionWidget.is(widget)) {
                widget.description = this._description;
                widget.onDescriptionChange.fire();
            }
        }
    }
```

packages/plugin-ext/src/main/browser/view/plugin-view-widget.ts
```diff
@@ -50,6 +52,8 @@ export class PluginViewWidget extends Panel implements StatefulWidget {
         this.node.style.height = '100%';
     }
 
+    public onDescriptionChange: Emitter<void> = new Emitter<void>();
```
vince-fugnitto, line 55: Can be removed, no need to duplicate the event and instead use the widget directly.

vince-fugnitto (commented):
I noticed a couple of things:
- the commit message needs to be updated, we mention `outline-view` but I believe we mean all view-container-parts.
- the description is only visible if the view-container-part is currently expanded in vscode - I believe we should align.

packages/core/src/browser/shell/description-widget.ts
aaronrhodes: I copied this implementation from a few examples in the rest of the code and did initially try and use `unknown`, but the compiler wouldn't let me use the string object accessor. Will update to `Object | undefined`.

msujew: For `unknown` you'll have to do this:

```ts
return typeof arg === 'object' && arg && !!arg['onDescriptionChange'];
```

aaronrhodes: I still get compiler errors about any having an implicit any type whichever way I try this..

aaronrhodes: I'm now moved this file into `view-container.ts` as it only really applies to the container widgets

packages/plugin-ext/src/main/browser/view/plugin-view-widget.ts
aaronrhodes: Have now removed this block of code as it was redundant - I use the `setWidgetMessage` flow as an example when first writing this but it's now obvious that description is more of an upward flowing property and has no utility or path to be passed down.

packages/plugin-ext/src/main/browser/view/plugin-view-widget.ts
aaronrhodes: As mentioned above, the wrapped widget actually doesn't have the `DescriptionWidget` interface implemented, but `plugin-view-widget` does so is needed

aaronrhodes:
> I noticed a couple of things:
> 
> * the commit message needs to be updated, we mention `outline-view` but I believe we mean all view-container-parts.
> * the description is only visible if the view-container-part is currently expanded in vscode - I believe we should align.

thanks for the feedback! Have updated the commit message and the subheading now hides and shows when the panel is collapsed/expanded respectively.

packages/core/src/browser/shell/description-widget.ts
vince-fugnitto: @aaronrhodes you can do the following to avoid using `any`:

```ts
export namespace DescriptionWidget {
    export function is(arg: Object | undefined): arg is DescriptionWidget {
        return !!arg && typeof arg === 'object' && 'onDidDescriptionChange' in arg;
    }
}
```

aaronrhodes: 👍

packages/core/src/browser/view-container.ts
```diff
@@ -853,15 +873,33 @@ export class ViewContainerPart extends BaseWidget {
...
+        subHeading.classList.add('subheading');
+
         const updateTitle = () => title.innerText = this.wrapped.title.label;
         const updateCaption = () => title.title = this.wrapped.title.caption || this.wrapped.title.label;
+        const updateSubHeading = () => {
+            if (DescriptionWidget.is(this.wrapped) && !this.collapsed) {
+                subHeading.innerText = !!this.wrapped.description ? this.wrapped.description : '';
+            } else {
+                subHeading.innerText = '';
+            }
```
msujew, line 887: Nitpick: We can simplify this

```suggestion
            subHeading.innerText = DescriptionWidget.is(this.wrapped) && !this.collapsed && this.wrapped.description || '';
```

packages/core/src/browser/view-container.ts
```diff
@@ -699,6 +712,12 @@ export class ViewContainerPart extends BaseWidget {
         this.wrapped.title.changed.connect(fireTitleChanged);
         this.toDispose.push(Disposable.create(() => this.wrapped.title.changed.disconnect(fireTitleChanged)));
 
+        if (DescriptionWidget.is(this.wrapped)) {
+            const fireSubHeadingChanged = () => this.onSubHeadingChangedEmitter.fire(undefined);
+            const sub = (this.wrapped as DescriptionWidget)?.onDescriptionChange.event(fireSubHeadingChanged);
```
vince-fugnitto, line 717: Update to a more meaningful name than `sub`?

packages/core/src/browser/style/view-container.css
```diff
@@ -75,10 +75,19 @@
...
+    white-space: nowrap;
+    text-overflow: ellipsis;
+}
+
+.theia-view-container .part > .header .subheading {
     flex: 1;
     overflow: hidden;
     white-space: nowrap;
     text-overflow: ellipsis;
+    padding-left: 12px;
```
vince-fugnitto, line 88: We should compute this value I believe rather than hardcoding.
Example:

https://github.com/eclipse-theia/theia/blob/a6422c39d6c09095f2bdc8cb4af4427e6bc78082/packages/debug/src/browser/style/index.css#L142

packages/core/src/browser/view-container.ts
aaronrhodes: 👍

packages/core/src/browser/view-container.ts
```diff
@@ -673,6 +684,8 @@ export class ViewContainerPart extends BaseWidget {
     readonly onTitleChanged = this.onTitleChangedEmitter.event;
     protected readonly onDidFocusEmitter = new Emitter<this>();
     readonly onDidFocus = this.onDidFocusEmitter.event;
+    protected readonly onSubHeadingChangedEmitter = new Emitter<void>();
+    readonly onSubHeadingChanged = this.onSubHeadingChangedEmitter.event;
```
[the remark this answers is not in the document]
aaronrhodes: 👍

packages/core/src/browser/style/view-container.css
aaronrhodes: Great suggestion, thanks!

packages/core/src/browser/view-container.ts
```diff
@@ -673,6 +684,8 @@ export class ViewContainerPart extends BaseWidget {
     readonly onTitleChanged = this.onTitleChangedEmitter.event;
     protected readonly onDidFocusEmitter = new Emitter<this>();
     readonly onDidFocus = this.onDidFocusEmitter.event;
+    protected readonly onDidSubHeadingChangedEmitter = new Emitter<void>();
+    readonly onDidSubHeadingChanged = this.onDidSubHeadingChangedEmitter.event;
```
colin-grant-work, line 688: I think `onDidChangeSubHeading(Emitter)` would conform more closely to the guidelines on [event naming](https://github.com/eclipse-theia/theia/wiki/Coding-Guidelines#event_names).

packages/core/src/browser/view-container.ts
```diff
@@ -699,6 +712,11 @@ export class ViewContainerPart extends BaseWidget {
         this.wrapped.title.changed.connect(fireTitleChanged);
         this.toDispose.push(Disposable.create(() => this.wrapped.title.changed.disconnect(fireTitleChanged)));
 
+        if (DescriptionWidget.is(this.wrapped)) {
+            const fireSubHeadingChanged = () => this.onDidSubHeadingChangedEmitter.fire(undefined);
+            this.toDispose.push((this.wrapped as DescriptionWidget)?.onDidDescriptionChange.event(fireSubHeadingChanged));
```
colin-grant-work, line 717: Since you've performed a type check already, you shouldn't need the `as` cast or the `?.` conditional accessor.

Why the change in terminology here from `onDidChangeDescription` to `onDidChangeSubHeading`? I think I'd prefer that we use the `description`-type names throughout and leave the fact that they're rendered as a subheading as an implementation detail.

packages/plugin-ext/src/main/browser/view/plugin-view-widget.ts
```diff
@@ -50,6 +52,8 @@ export class PluginViewWidget extends Panel implements StatefulWidget {
         this.node.style.height = '100%';
     }
 
+    public onDidDescriptionChange: Emitter<void> = new Emitter<void>();
```
colin-grant-work, line 55:
```suggestion
    onDidChangeDescription: Emitter<void> = new Emitter<void>();
```

Issue #6731, opened 2019-12-10
[notification] I want the notification promise to resolve when it timeouts


kittaakos:
### Description
 
The notification closes,
 - when a `timeout` was specified and not actions were provided and the notification closes without user interaction, or
 - when actions were provided, and the user dismisses it or performs one of the actions.

The notification promise resolves only in the latter case, I want the promise to resolve when it timeouts without interactions.

So I can write something like
```ts
this.messageService.warn(`Reconnecting in ${x} seconds`, { timeout: x * 1000 }).then(()=> this.reconnect());
```

### Reproduction Steps

**OS and Theia version:**

**Diagnostics:**

AlexTugarev:
`timeout` in message options was used to hide the message from the beginning. that could  break clients. 

in general, cancellable messages (like progress messages) would be a really nice idea, which was asked for already. cancellation after some timeout would do the trick. the only reason I didn't include that into the revision of the notification center was to not break API, because it required to change the type of the promise.

now, in examples like ☝️, actually, I suggest to use progress messages like so:

```ts
const progress = await this.messageService.showProgress({ text: `Reconnecting in ${x} seconds`,  });
timeoutAfter(4).then(() => progress.cancel);
progress.result.then(() => this.reconnect());
```

kittaakos:
> I suggest to use progress messages like

Nice! Thank you!

> cancellable messages (like progress messages) would be a really nice idea

I agree.

> which was asked for already

Assuming it someone has already requested it in another issue, and you gave me a working solution, there's not much to do here. I am closing it.

PR #9920, opened 2021-08-17
Add breadcrumbs for editors

### LINKED ISSUE #5475: Add Breadcrumb to Theia
### Description
Add breadcrumb in Theia between the menubar and the editor. VScode already have it available

### Reproduction Steps
breadcrumbTopEditor

**OS and Theia version:**
Theia 0.7.0
UIbuntu 16.04

colin-grant-work:
This PR updates #6731 and completes the remaining to-do's from that PR.

From that description:
> This commit adds a breadcrumbs bar to the editor widget. It shows the path to the current file and outline information as breadcrumbs. A click of breadcrumbs allows to switch to siblings.

> This fixes #5475.

> How to test:
> Open an editor shows the breadcrumbs on top of the editor. Folders with files are clickable. A popup allows to select a file that will be opened immediately. Outline information is added to the breadcrumbs as well.

image

A few more details:
1. You should see breadcrumbs for `NavigatableWidget`s (mainly editors) that are docked in an area with a horizontal tabbar. Those breadcrumbs should contain at least the path to the file.
2. For editors with language support, you should get the content of the outline view for the location of your cursor.
  > Try refreshing when you have a TS file open. The layout should be restored. There were changes to the Outline view that required changes to its state restoration.
3. If you open a non-navigatable widget, the breadcrumbs bar should disappear.
4. There is a preference to disable breadcrumbs. If you disable the breadcrumbs, they should disappear, and they shouldn't reappear anywhere.
  > Watch out for any odd styling behavior in the transition from breadcrumbs <-> no breadcrumbs. I think I've handled it, but there may still be some quirks.
5. If you click a breadcrumb, it should open a popup. The popup should show other possibilities at the same level of organization (sibling folders, object fields, etc.).
6. You should be able to navigate the tree

#### Review checklist

#### Reminder for reviewers


kenneth-marut-work (commented):
Looks really nice so far, and really nice functionality. Things mostly seem to be behaving as expecting but I have a few observations after doing an initial feature test:

---
-  Toggling the preference to show hide doesn't seem to resize the base editor, so you get a empty bar: 

breadrumbs-showhide
---
After waiting several minutes, I was able to get the correct behavior, but it seems to be very finicky. For example in in this GIF, you'll see the breadcrumbs toggle ON, but when I click on the editor they disappear again:

breadcrumbsnotshowingup

---
- In VSCode, clicking on a path segment reveals the corresponding segment (and *selected*) and its siblings at the top level vs. in this PR, we see the path segment's parent at the top level (and the segment is *unselected*):

**VSCode**
image

**This PR**
image

---
The treeview seems to have a minimum height that looks a bit odd when there are not enough elements to populate the view. In VSCode the view seems to only take up the minimum height it needs (and changes height when nodes are expanded -- up to a fixed maximum height):

image

---
Very minor: the breadcrumbs is a 24px height but the viewContainer toolbar is 22px. In VSCode they are both 22px (I only noticed it via the broken toggle mentioned above 🙂) 

image

--- 
In VSCode, if I click on a path segment in the breadcrumbs and then click on a file inside that crumb, a new editor opens with that file in *preview* mode. In this PR, the file opens in normal editor mode.

**This PR**
new-editor-open

**VSCode**
vscode-open

colin-grant-work:
@kenneth-marut-work, thanks for taking a look. I believe I've addressed your comments.
 - the odd toggling was a consequence of not hooking into Phosphor's fit logic; I've added code that asks the tabbar's parent to fit everything to size.
 - added code that selects the node corresponding to the crumb clicked on.
 - changed the trees to non-virtualized so that they can have a definite height smaller than 200px if there's less content to render.
 - changed the height of the breadcrumbs area to 22px. You'll have to check in your fancy theme whether everything lines up now :-)
 - changed the opener to use `{preview: true}`

msujew (commented):
I managed to notice some weird behavior as well. When opening the file tree via the breadcrumbs and opening a folder in there, clicking outside of the popup does not close the popup. This leads to things like these:

grafik

colin-grant-work:
> I managed to notice some weird behavior as well. When opening the file tree via the breadcrumbs and opening a folder in there, clicking outside of the popup does not close the popup.

@msujew, thanks for pointing this out. It pointed the way to some overly elaborate focus handling logic that conflicted with the expectations of the widgets being rendered in the popups. I've simplified that and used `focusout` events rather than `blur` to make things a little more flexible.

packages/core/src/browser/widgets/react-renderer.tsx
```diff
@@ -19,6 +19,9 @@ import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import { Disposable } from '../../common';
 
+export type RendererHost = HTMLElement;
```
kenneth-marut-work, line 22: Looks like these declarations might make more sense in the `breadcrumb-popup-container` file?

packages/outline-view/src/browser/outline-view-widget.tsx
```diff
@@ -27,9 +27,11 @@ import {
 } from '@theia/core/lib/browser';
 import { OutlineViewTreeModel } from './outline-view-tree-model';
 import { Message } from '@theia/core/shared/@phosphor/messaging';
-import { Emitter } from '@theia/core';
+import { Emitter, Mutable, UriSelection } from '@theia/core';
 import { CompositeTreeNode } from '@theia/core/lib/browser';
 import * as React from '@theia/core/shared/react';
+import { Range } from '@theia/editor/lib/browser/editor';
```
kenneth-marut-work, line 33: I think this is the only import from `editor` in this package if I'm not mistaken? Any way you can decouple editor from this package and get this type another way so you don't have to add the additional dependency?

kenneth-marut-work (commented):
A couple minor comments, also as we discussed offline the WorkspaceBreadcrumbsContribution doesn't seem to be doing what it should (prefixing the breadcrumbs with WS name in multiroot). But otherwise the functionality is very impressive and is working great for me, and thanks for addressing my previous comments as well 👍

packages/core/src/browser/breadcrumbs/breadcrumbs-renderer.tsx
```diff
@@ -0,0 +1,233 @@
...
+import { BreadcrumbRenderer } from './breadcrumb-renderer';
+import PerfectScrollbar from 'perfect-scrollbar';
+import URI from '../../common/uri';
+import { Emitter, Event } from '../../common';
+import { BreadcrumbPopupContainer } from './breadcrumb-popup-container';
+import { DisposableCollection } from '../../common/disposable';
+import { CorePreferences } from '../core-preferences';
+import { Coordinate } from '../context-menu-renderer';
+
+interface Cancelable {
```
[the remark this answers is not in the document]
colin-grant-work: This needs to be an object so that there can be a shared reference to it on the class and in the method. If it were just a boolean, setting the member to `false` wouldn't affect the variable referred to be the previous method call that needs to be cancelled.

msujew (approved):
Looks good to me, great work 👍 

One minor nitpick: vscode uses a larger padding on the left side, which I think looks a bit nicer:

vscode:

image

Theia:

image

packages/core/src/browser/breadcrumbs/breadcrumbs-constants.ts
```diff
@@ -0,0 +1,79 @@
...
+    /** The text that will be rendered as label. */
+    readonly label: string;
+
+    /** A longer text that will be used as tooltip text. */
+    readonly longLabel: string;
+
+    /** A CSS class for the icon. */
+    readonly iconClass?: string;
+
+    /** CSS classes for the container */
```
vince-fugnitto, line 47:
```suggestion
    /** CSS classes for the container. */
```

packages/core/src/browser/breadcrumbs/breadcrumbs-service.ts
```diff
@@ -0,0 +1,108 @@
...
+    protected createOverlayContainer(): void {
+        this.popupsOverlayContainer = window.document.createElement('div');
+        this.popupsOverlayContainer.id = Styles.BREADCRUMB_POPUP_OVERLAY_CONTAINER;
+        if (window.document.body) {
+            window.document.body.appendChild(this.popupsOverlayContainer);
+        }
+    }
+
+    /**
+     * Subscribe to this event emitter to be notifed when the breadcrumbs have changed.
```
vince-fugnitto, line 52: minor typo:

```suggestion
     * Subscribe to this event emitter to be notified when the breadcrumbs have changed.
```

packages/core/src/browser/breadcrumbs/breadcrumbs-service.ts
```diff
@@ -0,0 +1,108 @@
...
+     * The URI is the URI of the editor the breadcrumbs have changed for.
+     */
+    get onDidChangeBreadcrumbs(): Event<URI> {
+        // This lazy subscription is to address problems in iversify's instantiation routine
```
vince-fugnitto, line 56: minor typo:

```suggestion
        // This lazy subscription is to address problems in inversify's instantiation routine
```

packages/filesystem/src/browser/breadcrumbs/filepath-breadcrumbs-contribution.ts
```diff
@@ -0,0 +1,129 @@
...
+import { Breadcrumb, BreadcrumbsContribution, CompositeTreeNode, LabelProvider, SelectableTreeNode, Widget } from '@theia/core/lib/browser';
+import { FilepathBreadcrumb } from './filepath-breadcrumb';
+import { BreadcrumbsFileTreeWidget } from './filepath-breadcrumbs-container';
+import { DirNode } from '../file-tree';
+import { FileService } from '../file-service';
+import { FileStat } from '../../common/files';
+
+export const FilepathBreadcrumbType = Symbol('FilepathBreadcrumb');
+
+export interface FilepathBreadccrumbClassNameFactory {
```
vince-fugnitto, line 29: typo:

```suggestion
export interface FilepathBreadcrumbClassNameFactory {
```

packages/monaco/src/browser/monaco-outline-contribution.ts
```diff
@@ -42,20 +41,17 @@ export class MonacoOutlineContribution implements FrontendApplicationContributio
...
-                    debounce(() => this.handleCurrentEditorChanged(), 50)
-                ));
-                this.handleCurrentEditorChanged();
-            } else {
-                this.toDisposeOnClose.dispose();
-            }
-        });
+
+        // updateOutline and handleCurrentEditorChanged need to be called even when the outline view widget is closed
+        // in order to udpate breadcrumbs.
```
vince-fugnitto, line 46: minor typo:

```suggestion
        // in order to update breadcrumbs.
```

packages/outline-view/src/browser/outline-view-service.ts
```diff
@@ -61,8 +61,10 @@ export class OutlineViewService implements WidgetFactory {
     publish(roots: OutlineSymbolInformationNode[]): void {
         if (this.widget) {
             this.widget.setOutlineTree(roots);
-            this.onDidChangeOutlineEmitter.fire(roots);
         }
+        // onDidChangeOutline needs to be fired even when the outline view widget is closed
+        // in order to udpate breadcrumbs.
```
vince-fugnitto, line 66: minor typo:

```suggestion
        // in order to update breadcrumbs.
```

packages/core/src/browser/breadcrumbs/breadcrumbs-service.ts
```diff
@@ -0,0 +1,108 @@
...
+        // related to use of the IconThemeService by different components instantiated by the
+        // ContributionProvider.
+        if (!this.hasSubscribed) {
+            this.subscribeToContributions();
+        }
+        return this.onDidChangeBreadcrumbsEmitter.event;
+    }
+
+    /**
+     * Subscribes to the onDidChangeBreadcrumbs events for all contributions
```
vince-fugnitto, line 66:
```suggestion
     * Subscribes to the onDidChangeBreadcrumbs events for all contributions.
```

vince-fugnitto:
I noticed the `perfect-scrollbar` styling might be off for the breadcrumbs compared to the rest of the application (and vscode).
I'm not sure if it is something we want handled initially by the pull-request but I was able to reproduce on `electron`:


https://user-images.githubusercontent.com/40359487/131505872-403d7af1-6ed5-4844-867e-dec2fcf8e309.mp4

colin-grant-work:
> I noticed the `perfect-scrollbar` styling might be off for the breadcrumbs compared to the rest of the application (and vscode).
> I'm not sure if it is something we want handled initially by the pull-request but I was able to reproduce on `electron`:

@vince-fugnitto The easiest (and, I think, most correct) way to solve that would be to remove the scroll bar from the breadcrumb popup container and require breacrumb providers to provide their own scroll if necessary - both of the current ones show tree widgets, so they come with their own scrollbars anyway. That also avoids duplication of scroll bars in case the rendered UI happens to include a scroll bar, as all widgets would. Does that sound reasonable?

vince-fugnitto:
> @vince-fugnitto The easiest (and, I think, most correct) way to solve that would be to remove the scroll bar from the breadcrumb popup container and require breacrumb providers to provide their own scroll if necessary - both of the current ones show tree widgets, so they come with their own scrollbars anyway. That also avoids duplication of scroll bars in case the rendered UI happens to include a scroll bar, as all widgets would. Does that sound reasonable?

I think that sounds reasonable yes :)

colin-grant-work:
> @vince-fugnitto The easiest (and, I think, most correct) way to solve that would be to remove the scroll bar from the breadcrumb popup container and require breacrumb providers to provide their own scroll if necessary - both of the current ones show tree widgets, so they come with their own scrollbars anyway. That also avoids duplication of scroll bars in case the rendered UI happens to include a scroll bar, as all widgets would. Does that sound reasonable?

Turns out I couldn't do that. I'd hoped the good styles were attached to widgets, but in fact they're attached to the shell, and dialogues, and now also breadcrumb popups. I think we could do that better, since I think we really just mean 'everywhere in the application,' but someone decided that that should be classed down to the shell. If you agree, @vince-fugnitto, we can probably simplify the perfect scrollbar CSS section by eliminating all of the #-referencing prefixes, but that might have unintended consequences (maybe the monaco overlays use PS and want different styles, e.g.? Perhaps best to put that off for another PR.

colin-grant-work:
If CI passes and no one objects, I'll merge this Wednesday or Thursday.

Issue #9919, opened 2021-08-17
monaco: `recently used commands` not displaying


vince-fugnitto:
#### Bug Description:

I was finally able to identify the path that lead to the bug I noticed in https://github.com/eclipse-theia/theia/pull/9727#discussion_r677449756.

When we trigger the `palette` and type `>` (to trigger the quick-commands), we will not get the same `recently used` results as if we trigger <kbd>F1</kbd> directly.

https://user-images.githubusercontent.com/40359487/129747246-a0f191cf-d7b2-46a7-93d5-cbdf713345ea.mp4

### Steps to Reproduce:

1. ensure that you have a clean state
    - `rm -rf ~/.theia`
    - `reset workbench layout`
    - `clear command history`
    - close the application and restart 
2. trigger the quick-open (<kbd>ctrlcmd</kbd>+<kbd>p</kbd>) and type `>`
3. execute a command, repeat for multiple commands
4. triggering the quick-open with <kbd>ctrlcmd</kbd>+<kbd>p</kbd> will not yield recently-used results
5. trigger <kbd>F1</kbd> - the correct results will now appear

### Additional Information

- Operating System: Ubuntu 18.04
- Theia Version: https://github.com/eclipse-theia/theia/commit/7d637a175bcd653389ce363e08564a7b41417e2b

PR #9921, opened 2021-08-17
core: fix `quick-commands` to properly display recently used commands

vince-fugnitto:
#### What it does

Fixes: #9919 

The commit fixes the `quick-command` behavior which would not display the list of recently used commands properly when users would use the `quick-command` + type the prefix value, instead of triggering the command directly. The change refactors the code to perform the `reset` when we trigger the picks so we pick up the latest commands available.

#### How to test

1. execute commands in order to add them to recently used list.
2. using either paths to trigger the `quick-command-service` should work and properly display recent commands:
    - <kbd>F1</kbd> directly.
    - <kbd>ctrlcmd</kbd>+<kbd>p</kbd> then typing `>` to trigger the prefixed `quick-commands`.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/quick-input/quick-command-frontend-contribution.ts
+++ b/packages/core/src/browser/quick-input/quick-command-frontend-contribution.ts
@@ -32,7 +32,6 @@ export class QuickCommandFrontendContribution implements CommandContribution, Ke
     registerCommands(commands: CommandRegistry): void {
         commands.registerCommand(quickCommand, {
             execute: () => {
-                this.quickCommandService?.reset();
                 this.quickInputService?.open('>');
             }
         });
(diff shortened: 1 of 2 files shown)


alvsan09 (approved):
It looks good to me! :+1: 
I have tested that after this change and following the steps to reproduce the initial issue, 
the following options to display the command history list the entries as expected: 

* ctrlcmd + p and then >
* ctrl+shft+p
* F1

msujew (approved):
I can confirm that the issue exists on `master` and is fixed by this change ✔️

tsmaeder:
We should probably move the call of the optional "reset()" to the MonacoPickerAccessProvider.getPicks() implementation in monaco-quick-access-provider.ts, but since this change improves things from the current state, so merging.

PR #9932, opened 2021-08-19
preferences: Fix search term on open via menu

### LINKED ISSUE #9931: Opening the preferences menu via menu automatically sets a search term
### Bug Description:

When opening the preferences menu via the menu action like this, Theia automatically places `[object Object]` into the search bar, resulting in no preferences being found. This was probably introduced by #9785.

### Steps to Reproduce:

1. Open the preferences menu via the menu action: 
grafik
2. Observe the opened preferences view:
grafik

### Additional Information

- Operating System: Windows
- Theia Version: a6422c3

msujew:
#### What it does
Closes #9931 

Fixes the issue by simply checking for the string type before passing the argument into the search term.

#### How to test

1. Open the preferences menu via the settings menu in the bottom left.
2. Observe an empty search bar

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/browser/preferences-contribution.ts
+++ b/packages/preferences/src/browser/preferences-contribution.ts
@@ -67,7 +67,7 @@ export class PreferencesContribution extends AbstractViewContribution<Preference
         commands.registerCommand(CommonCommands.OPEN_PREFERENCES, {
             execute: async (query?: string) => {
                 const widget = await this.openView({ activate: true });
-                if (query) {
+                if (typeof query === 'string') {
                     widget.setSearchTerm(query);
                 }
             },


packages/core/src/browser/context-menu-renderer.ts
```diff
@@ -94,15 +94,14 @@ export interface RenderContextMenuOptions {
...
+        let args: any[] | undefined;
         if (Array.isArray(menuPathOrOptions)) {
             menuPath = menuPathOrOptions;
-            args = [anchor!];
         } else {
             menuPath = menuPathOrOptions.menuPath;
             anchor = menuPathOrOptions.anchor;
             onHide = menuPathOrOptions.onHide;
-            args = menuPathOrOptions.args ? [...menuPathOrOptions.args, anchor] : [anchor];
+            args = menuPathOrOptions.args;
```
colin-grant-work, line 104: I think this change is correct, but I happen to know that I've relied on the existing behavior downstream to distinguish command activations from different sources, and I may not be alone in that. A more limited fix to the specific problem might be to check whether the type of the arg is string before setting it in the search bar of the preferences view.

msujew: That was my first thought as well, but that sounds like a workaround, as this issue will repeat itself for any command with a single optional argument. But I guess I can change it that way, if you rely on it. Maybe there's a cleaner solution for that problem?

colin-grant-work: I'm OK with removing it - I know where I've relied on it and how to fix it - but it would definitely be a breaking change worth highlighting in the changelog, and I'm not sure that there aren't legitimate uses for it that we'd be cutting off. It's worth being cautious about, because it isn't something that people may realize was questionable behavior, and they won't get much help from TypeScript about it, because command handlers are supposed to expect `any[]`, so changes in what will _actually_ be passed in aren't detected by the type system.

vince-fugnitto (commented):
@msujew I confirmed that it fixes the issue with the `preferences`, however I believe it introduces a regression in the following use-case:

_pull-request_:

https://user-images.githubusercontent.com/40359487/130087610-5074c5fe-6b45-46df-854c-93f5c5c4f0da.mp4

_master_:

https://user-images.githubusercontent.com/40359487/130087778-cd10ccee-42e3-4eab-8a5a-dc5c746eb375.mp4

colin-grant-work:
> @msujew I confirmed that it fixes the issue with the `preferences`, however I believe it introduces a regression in the following use-case:

Very timely: there's the legitimate use case for it (maybe - I'm not sure that's how the menu should work...).

msujew:
> however I believe it introduces a regression

Ah, as I thought we rely somewhere on that undocumented behavior. :/

Alright, I'll look for a more localized fix for this issue.

colin-grant-work (approved):
This solves the problem nicely, and definitely looks like a safer approach. 👍

vince-fugnitto (approved):
I confirmed that the change fixes the issue on `master` :+1: 
It is also self-contained and does not cause any regressions.

PR #8967, opened 2021-01-18
#8966: Update `ViewContainer` Toolbars

colin-grant-work:
Signed-off-by: Colin Grant <colin.grant@ericsson.com>

#### What it does

Fixes #8966 by adding code to update a `ViewContainer` part toolbar when a change event is emitted by the `TabbarToolbarRegistry`.

##### Toolbar Part
1. Register a toolbar item that includes a `didChange` event and should appear in a `ViewContainer` part (for example, one that is attached to part of the debug panel or one that is attached to the file explorer when a plugin contribution (e.g. NPM) is also present).
1. Hover to reveal the toolbar and fire the command to trigger the `didChange` event.
1. Observe that the event *does* affect the `ViewContainer` part toolbar.

After the fix

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #9935, opened 2021-08-20
Use CSS to control ViewContainerPart toolbar visibility

colin-grant-work:
#### What it does

> This PR supersedes #8967, as it contains alternative logic for updating the toolbar.

At the moment, we have quite an elaborate routine for managing the visibility of the toolbars on `ViewContainerPart`s. The results differ from VSCode in a couple of ways:
 - The toolbar disappears if a context-menu is opened using the `...` more icon.
 - The toolbar isn't always visible if the widget is focused.

#### How to test

1. Test different states of widget focus and hovering.
   - The toolbar should never be visible if the `ViewContainerPart` is collapsed
   - The toolbar should be visible if the widget is focused
   - The toolbar should be visible if you hover over the widget
   - The toolbar should be visible if a context menu is opened from the toolbar

#### Review checklist

#### Reminder for reviewers


msujew (approved):
Looks good me:

* Focusing the container part always shows the toolbar ✔️ 
* Hovering inside the container part always shows the toolbar ✔️ 
* A collapsed container part never shows the toolbar ✔️ 
* Using the ellipsis menu still shows the toolbar ✔️ 
* Everything works as expected when there's only a single container part inside the view ✔️

vince-fugnitto (approved):
The changes look good and work well for me :+1: 

- [x] the toolbar should never be visible if the ViewContainerPart is collapsed
- [x] the toolbar should be visible if the widget is focused
- [x]  the toolbar should be visible if you hover over the widget
- [x] the toolbar should be visible if a context menu is opened from the toolbar
- [x] confirmed that executing toolbar items, and menu items in the `...` (more) context-menu works well

Issue #9860, opened 2021-08-06
Electron File Dialog Can't Open Read-only Files


colin-grant-work:
### Bug Description:

Before returning from `.showOpenDialog()`, the `ElectronFileDialogService` checks whether the selected files are both readable and writable for the current user. It seems the check for writability makes sense for the `.showSaveDialog()` method, but not necessarily for the `.showOpenDialog()` case, since 'opening' doesn't necessarily imply writing.

https://github.com/eclipse-theia/theia/blob/7e09e93a7a741137a9581125e8df39bcfeb05ee5/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts#L47-L61

https://github.com/eclipse-theia/theia/blob/7e09e93a7a741137a9581125e8df39bcfeb05ee5/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts#L83-L91

### Steps to Reproduce:

1. On a multi-user system, find or create a file that certain users can read but not write
2. As a user who can read but not write to a given file, try to 'open' the file using the file dialog
3. See an error toast `Cannot access resource at <path>.`

- Operating System: RHEL7
- Theia Version: 7e09e93a7a741137a9581125e8df39bcfeb05ee5

tsmaeder:
Sounds like a low hanging fruit for a new contributor.

kejsitake:
Hi! I submitted a PR. I would be happy to make any other changes if necessary.

PR #9950, opened 2021-08-22
Enables opening read-only files in electron.

kejsitake:
Signed-off-by: Kejsi Take <kejsitake@hotmail.com>

Addresses issue #9860
Makes sure that a user can open a new file in the ElectronFileDialog when it has Read permissions (and not Write).

#### What it does

- Adds a new canRead() method to check if the user has read permissions to the file.
- Calls the method from .showOpenDialog() 

#### How to test

After making the above-mentioned changes, I ran the Electron example and verified that I could open a file that had only read permissions.

#### Review checklist

- [ X] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

### DIFF
--- a/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
+++ b/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
@@ -53,7 +53,7 @@ export class ElectronFileDialogService extends DefaultFileDialogService {
             }
 
             const uris = filePaths.map(path => FileUri.create(path));
-            const canAccess = await this.canReadWrite(uris);
+            const canAccess = await this.canRead(uris);
             const result = canAccess ? uris.length === 1 ? uris[0] : uris : undefined;
             return result;
         }
@@ -74,8 +74,8 @@ export class ElectronFileDialogService extends DefaultFileDialogService {
                 return uri;
             }
 
-            const canAccess = await this.canReadWrite(uri);
-            return canAccess ? uri : undefined;
+            const canWrite = await this.canReadWrite(uri);
+            return canWrite ? uri : undefined;
         }
         return undefined;
     }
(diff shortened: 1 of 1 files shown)


tsmaeder (commented):
The code looks good and I've tested that I can open a read-only file on Windows. There is a lot of weird behaviour around reading/writing files into read-only directories on Windows, but the PR fixes the particular problem the issue is about.

tsmaeder:
@colin-grant-work could you test on Linux?

packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
```diff
@@ -90,6 +90,16 @@ export class ElectronFileDialogService extends DefaultFileDialogService {
         return true;
     }
 
+    protected async canRead(uris: MaybeArray<URI>): Promise<boolean> {
+        for (const uri of Array.isArray(uris) ? uris : [uris]) {
+            if (!(await this.fileService.access(uri, FileAccess.Constants.R_OK))) {
+                this.messageService.error(`Cannot read resource at ${uri.path}.`);
+                return false;
+            }
+        }
```
msujew, line 99: I could imagine that this could impact performance for large arrays of URIs because every `access` call is sequential. We can parallelize this fairly well like this:
```suggestion
        const filePaths = await Promise.all((Array.isArray(uris) ? uris : [uris]).map(uri => {
            return !await this.fileService.access(uri, FileAccess.Constants.R_OK) && uri.path || '';
        }).filter(e => e);
        if (filePaths.length) {
            this.messageService.error(`Cannot read ${filePaths.length} resources: ${filePaths.join(', ')}`);
        }
        return !!filePaths.length;
```

msujew:
Hi @kejsitake, thank you for your first contribution and taking care of this issue 👍 

Please be sure to sign the [eclipse contributor agreement (eca)](https://github.com/eclipse-theia/theia/blob/master/CONTRIBUTING.md#eclipse-contributor-agreement) with the same email as your authorship in order for us to accept the changes.

kejsitake:
I think I got the changes done and signed the eca. I am happy to look at this again if needed.

PR #9957, opened 2021-08-24
electron: fix problematic `HttpFallbackOptions` injection

vince-fugnitto:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/9958
Reference: https://github.com/eclipse-theia/theia/pull/9731#issuecomment-904818785

The commit fixes the problematic `HttpFallbackOptions` injection which should not be available on `electron`. The fix updates the injection making it `optional`.

#### How to test

- confirm that the `electron` target builds and starts

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/messaging/ws-connection-provider.ts
+++ b/packages/core/src/browser/messaging/ws-connection-provider.ts
@@ -14,7 +14,7 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-import { injectable, interfaces, decorate, unmanaged, inject } from 'inversify';
+import { injectable, interfaces, decorate, unmanaged, inject, optional } from 'inversify';
 import { JsonRpcProxyFactory, JsonRpcProxy, Emitter, Event, MessageService, MessageServiceFactory } from '../../common';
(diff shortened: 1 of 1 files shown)


msujew (approved):
I can confirm that the changes fix the electron injection issue. Even if the dependency is not bound, the app still works correctly.

The code looks reasonable to me as well 👍

vince-fugnitto:
@msujew an alternative would be to make the binding available on `electron` but I don't believe it makes sense to have the http fallback in the desktop use-case correct?

```ts
export const messagingFrontendModule = new ContainerModule(bind => {
    bind(ElectronWebSocketConnectionProvider).toSelf().inSingletonScope();
    bind(FrontendApplicationContribution).toService(ElectronWebSocketConnectionProvider);
    bind(WebSocketConnectionProvider).toService(ElectronWebSocketConnectionProvider);
    bind(ElectronIpcConnectionProvider).toSelf().inSingletonScope();
    bind(HttpFallbackOptions).toConstantValue(DEFAULT_HTTP_FALLBACK_OPTIONS);
});
```

msujew:
@vince-fugnitto Yes, I thought about that as well. I think the current fix is fine as it is 👍

PR #9993, opened 2021-08-30
electron: fix re-opening of electron window

vince-fugnitto:
#### What it does

Fixes: #9990.

The commit restores the previous window state (position and size) when re-opening an electron window, and also preserves the fix that avoids the electron window overlap. The idea is to use the previous `windowstate` when opening the window when a single window is found.

#### How to test

1. start the example-electron application.
2. resize the window, and move the window position.
3. close the window, and re-open it - the size and position should be restored.
4. open a new window (or folder) - there should be a proper overlap (not exactly on top of the other window(s)).

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

Signed-off-by: vince-fugnitto <vincent.fugnitto@ericsson.com>

Issue #9990, opened 2021-08-30
Electron version does not restore window size


ccorn:
### Bug Description:

The electron build of Theia v1.17.0 no longer restores window size and position on startup.
Even after manually resizing and quitting, the next start still does not restore window size and position. Theia v1.16.0 does not have this issue.

Reverting 19ace46f5dadc3345cc14a7edf9f85a9a025a9b0 solves this issue.
That revert required manual conflict resolution; see the diff below for my result.

### Steps to Reproduce:

 1. `git checkout v1.17.0`
 2. `yarn install`
 3. `yarn build`
 4. `yarn start:electron`
 5. Observe that window size is not restored; resize and quit
 6. `yarn start:electron`
 7. Observe that window size is still not restored; quit
 8. `git switch -c revert-nowinoverlap`
 9. `git revert 19ace46f5dadc3345cc14a7edf9f85a9a025a9b0`
    with conflict resolution as demonstrated in the attached (resulting) patch
10. `git commit -m 'Revert 19ace46 "Enhancement : Avoid electron windows to overlap"'`
11. `yarn build`
12. `yarn start:electron`
13. Observe that window size gets restored; resize and quit
14. `yarn start:electron`
15. Observe that window size gets restored; quit

### Additional Information

- Operating System: Arch Linux `5.10.52-1-lts x86_64`
- Theia Version: 1.17.0
- NodeJS: LTS-Erbium 12.22.4

The resulting diff (should be applicable with `-p1`) from reverting with my conflict resolution:

```diff
--- a/packages/core/src/electron-main/electron-main-application.ts
+++ b/packages/core/src/electron-main/electron-main-application.ts
@@ -38,7 +38,6 @@ const createYargs: (argv?: string[], cwd?: string) => Argv = require('yargs/yarg
  */
 export interface TheiaBrowserWindowOptions extends BrowserWindowConstructorOptions {
     isMaximized?: boolean;
-    isFullScreen?: boolean;
 }
 
 /**
@@ -216,9 +215,8 @@ export class ElectronMainApplication {
      *
      * @param options
      */
-    async createWindow(asyncOptions: MaybePromise<TheiaBrowserWindowOptions> = this.getDefaultTheiaWindowOptions()): Promise<BrowserWindow> {
-        let options = await asyncOptions;
-        options = this.avoidOverlap(options);
[...]

PR #9995, opened 2021-08-30
electron: restore previous workspace

vince-fugnitto:
#### What it does

Fixes: #9990.
Fixes: #9997.
Closes: #9993

The commit fixes logic when determining whether to open a default window or existing workspace on electron startup. The update now correctly restores the workspace which would previously never restore properly. As a consequence, the fix also resolves the window-state not being properly restored.

#### How to test

_Workspace Restoration_:

1. start the `example-electron` application, and select a workspace folder.
2. close the application.
3. the application should re-open with the previous workspace folder (master would reload to the default window).
4. execute the command `new window` - a new window should open with no workspace present.
5. start the application and specify a workspace folder (ex: `(cd examples/electron/ && yarn start ~/workspaces/theia)`) - the application should start and open with the specified workspace.

_Test Size + Position_:

1. start the `example-electron` application
2. resize the window, move the position
3. close the application, and re-open
4. confirm the window size and position is restored
5. opening new folders (which spawns new windows) should not overlap eachother

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/window/window-service.ts
+++ b/packages/core/src/browser/window/window-service.ts
@@ -26,9 +26,9 @@ export interface NewWindowOptions {
 export const WindowService = Symbol('WindowService');
 
 /**
- *  The window hash value that is used to spawn a new default window.
+ * The window hash value that is used to spawn a new default window.
  */
-export const DEFAULT_WINDOW_HASH: string = '#!empty';
+export const DEFAULT_WINDOW_HASH: string = '!empty';
 
 export interface WindowService {
 
--- a/packages/core/src/electron-main/electron-main-application.ts
+++ b/packages/core/src/electron-main/electron-main-application.ts
@@ -290,7 +290,9 @@ export class ElectronMainApplication {
     }
 
     protected async handleMainCommand(params: ElectronMainExecutionParams, options: ElectronMainCommandOptions): Promise<void> {
-        if (options.file === undefined) {
+        if (params.secondInstance === false) {
+            await this.openWindowWithWorkspace(''); // restore previous workspace.
+        } else if (options.file === undefined) {
             await this.openDefaultWindow();
         } else {
             let workspacePath: string | undefined;
--- a/packages/workspace/src/browser/workspace-service.ts
+++ b/packages/workspace/src/browser/workspace-service.ts
@@ -130,7 +130,8 @@ export class WorkspaceService implements FrontendApplicationContribution {
     protected async doGetDefaultWorkspaceUri(): Promise<string | undefined> {
 
         // If an empty window is explicitly requested do not restore a previous workspace.
-        if (window.location.hash === DEFAULT_WINDOW_HASH) {
+        // Note: `window.location.hash` includes leading "#" if non-empty.
+        if (window.location.hash === `#${DEFAULT_WINDOW_HASH}`) {
             window.location.hash = '';
             return undefined;
         }
@@ -213,6 +214,7 @@ export class WorkspaceService implements FrontendApplicationContribution {
             this.setURLFragment('');
         }
         this.updateTitle();
+        await this.server.setMostRecentlyUsedWorkspace(this._workspace ? this._workspace.resource.toString() : '');
         await this.updateWorkspace();
     }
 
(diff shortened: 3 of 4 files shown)


msujew (approved):
I can confirm that the issue exists on `master` and is addressed by this change nicely.

ccorn:
I hereby confirm that `git cherry-pick origin/vf/electron-restore` solves #9990 for me as well as immediately opening the previous workspace instead of showing a welcome page. Well done!

vince-fugnitto:
One limitation with the approach is if the default window (no workspace) was the last opened window, then we will still attempt to re-open the last workspace. For comparison, vscode restores the default window successfully in such a case.

ccorn:
Question: This would be the first non-maintenance commit to `master` since v1.17.0. Could we have a release tag (e.g. `v1.17.1`) so packagers do not need to cherry-pick?

vince-fugnitto:
> Question: This would be the first non-maintenance commit to `master` since v1.17.0. Could we have a release tag (e.g. `v1.17.1`) so packagers do not need to cherry-pick?

@ccorn still confirming if the patch is proper, or if additional work is required to handle the other use-cases. As for creating a patch release, we generally do not do it for each non-maintenance  commits to master but only if a major bug is resolved (we can think about if this warrants a patch release). Else, you can always use a `next` version of the `@theia` dependencies once the commit is merged for your application.

ccorn:
> One limitation with the approach is if the default window (no workspace) was the last opened window, then we will still attempt to re-open the last workspace. For comparison, vscode restores the default window successfully in such a case.

In that case, opening the welcome page should clear Theia's idea of *latest workspace* (without clearing the list of recent workspaces, of course), or there should be a distinction between *latest window* and *latest workspace*, I think.

ccorn:
> One limitation with the approach is if the default window (no workspace) was the last opened window, then we will still attempt to re-open the last workspace. For comparison, vscode restores the default window successfully in such a case.

Interesting. With this patch:

 1. `yarn start:electron` opens previous workspace with previous window coordinates (as expected)
 2. *New Window* opens the welcome page with new window coordinates (as expected)
 3. Close the workspace window, leaving only the welcome page, then close that as well.
 4. `yarn start:electron` now opens previous workspace with latest welcome page's coordinates. Hmmm. Not expected, but not such a show stopper as the behavior without this patch.

paul-marechal (approved):
Code LGTM.

Using this patch I am able to restore the last workspace as well as the case where the workspace is not set (the window re-opens without a workspace).

paul-marechal:
@ccorn can you also confirm that the behavior is fixed?

ccorn:
> @ccorn can you also confirm that the behavior is fixed?

The slightly odd behavior described [above](https://github.com/eclipse-theia/theia/pull/9995#issuecomment-908501389) remains.
@paul-marechal: How have you got rid of workspaces so the next start shows the welcome page instead (apart from deleting persistent storage)?

paul-marechal:
@ccorn I followed the same procedure as you mentioned.

Vincent force-pushed, maybe you are still using the old changes? `git show` should show commit c4fff9ecbf2fa17c6fe31b01a10455b0d3acd653.

ccorn:
> @ccorn I followed the same procedure as you mentioned.
> 
> Vincent force-pushed, maybe you are still using the old changes? `git show` should show commit [c4fff9e](https://github.com/eclipse-theia/theia/commit/c4fff9ecbf2fa17c6fe31b01a10455b0d3acd653).

I basically do a `git fetch origin vf/electron-restore` (which has got the forced update) and apply that on a branch starting from `v1.17.0` using `git cherry-pick FETCH_HEAD`. The latest such action reported `4 files changed, 9 insertions(+), 15 deletions(-)`, and `git show FETCH_HEAD` showed `commit c4fff9ecbf2fa17c6fe31b01a10455b0d3acd653`. That should be it.

ccorn:
Update: A clean rebuild fixed it. I can confirm that closing the welcome page at the end will get it reopened at the next start, otherwise the latest workspace gets restored.

Excellent issue handling! I cannot comment on the code, but I ran across a Linter warning for one of the affected files:

```log
@theia/workspace: /home/ccorn/BUILD/theia/packages/workspace/src/browser/workspace-service.ts
@theia/workspace:   434:16  warning  'getTemporaryWorkspaceFileUri' is deprecated. since 1.4.0 - because of https://github.com/eclipse-theia/theia/wiki/Coding-Guidelines#di-function-export, use `WorkspaceService.getUntitledWorkspace` instead  deprecation/deprecation
```

paul-marechal:
We'll release tomorrow, just need to give a heads up during the dev meeting as there's at least one other commit I'd like to include.

Issue #10014, opened 2021-08-31
task: handle possibly missing `required` property


vince-fugnitto:
#### Bug Description:

There are errors thrown if an extension provides a `TaskDefinition` without the `required` property.

Based on the [json schema validation](https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.5.3), the following is documented for the `required` keyword:

> 6.5.3. required
> The value of this keyword MUST be an array. Elements of this array, if any, MUST be strings, and MUST be unique.
>
> An object instance is valid against this keyword if every item in the array is the name of a property in the instance.
> 
> Omitting this keyword has the same behavior as an empty array.

_Error_:

```yaml
Uncaught (in promise) TypeError: Cannot read property 'find' of undefined
    at task-schema-updater.ts:172
    at Array.forEach (<anonymous>)
    at task-schema-updater.ts:171
    at Array.forEach (<anonymous>)
    at TaskSchemaUpdater.updateSchemasForRegisteredTasks (task-schema-updater.ts:158)
    at TaskSchemaUpdater.updateSupportedTaskTypes (task-schema-updater.ts:219)
    at task-schema-updater.ts:75
    at event.ts:117
    at CallbackList.invoke (event.ts:125)
    at Emitter.fire (event.ts:267)
```

```yaml
Uncaught (in promise) TypeError: can't access property "find", def.properties.required is undefined
```

We should therefore be more robust when handling the `required` property, and if omitted treat it as an empty array.

#### Steps to Reproduce:

1. update the `theiaPlugins` to the following:
```json
"theiaPlugins": {
  "vscode-optional-required": "https://github.com/vince-fugnitto/vscode-optional-required/releases/download/v0.0.1/task-provider-samples-0.0.1.vsix"
}
```

2. start the `example-browser` application - notice the errors in the console

PR #10015, opened 2021-08-31
task: make `TaskDefinition` `required` optional

vince-fugnitto:
#### What it does

Fixes: #10014 

The commit makes `TaskDefinition.properties.required` optional based on the json schema validation definition. Previously, if an extension defines a schema that omits the `required` property the application will throw errors and fail to display the menu. The change treats the missing `required` property as an empty array as per the spec.

#### How to test

- Follow the steps as described in #10014

#### Review checklist

#### Reminder for reviewers

image

### DIFF
--- a/packages/task/src/browser/task-configurations.ts
+++ b/packages/task/src/browser/task-configurations.ts
@@ -249,10 +249,9 @@ export class TaskConfigurations implements Disposable {
         if (hasCustomization) {
             const taskDefinition = this.taskDefinitionRegistry.getDefinition(taskConfig);
             if (taskDefinition) {
-                const cus = customizationByType.filter(customization =>
-                    taskDefinition.properties.required.every(rp => customization[rp] === taskConfig[rp])
-                )[0]; // Only support having one customization per task
-                return cus;
+                const required = taskDefinition.properties.required || [];
+                // Only support having one customization per task.
+                return customizationByType.find(customization => required.every(property => customization[property] === taskConfig[property]));
             }
         }
         return undefined;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 <a name="breaking_changes_1.18.0">[Breaking Changes:](#breaking_changes_1.18.0)</a>
 
 - [core] added `BreadcrumbsRendererFactory` to constructor arguments of `DockPanelRenderer` and `ToolbarAwareTabBar`. [#9920](https://github.com/eclipse-theia/theia/pull/9920)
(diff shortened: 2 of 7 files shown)


paul-marechal (approved):

packages/task/src/browser/task-configurations.ts
```diff
@@ -250,7 +250,7 @@ export class TaskConfigurations implements Disposable {
             const taskDefinition = this.taskDefinitionRegistry.getDefinition(taskConfig);
             if (taskDefinition) {
                 const cus = customizationByType.filter(customization =>
-                    taskDefinition.properties.required.every(rp => customization[rp] === taskConfig[rp])
+                    taskDefinition.properties.required?.every(rp => customization[rp] === taskConfig[rp])
```
colin-grant-work, line 253: If no `required` is specified, should this pass or fail? This code will filter out anything where `required` is not defined, but if having no required is the same as having `[]`, then the `every` check would pass.

Could do `(taskDefinition.properties.required ?? [])...` or `taskDefinition.properties.required?.every(...) ?? true`

paul-marechal: Ah because `[].every(...)` returns `true`, makes sense.

colin-grant-work (approved):

vince-fugnitto:
@tsmaeder any objections?

tsmaeder:
@vince-fugnitto which schema are we using? https://code.visualstudio.com/api/references/contribution-points#contributes.taskDefinitions seems to indicate that the field is required. Do we have real cases of extensions that fail because of this?

Not objecting to be more lenient than the json schema in what we accept, though.

vince-fugnitto:
> @vince-fugnitto which schema are we using? https://code.visualstudio.com/api/references/contribution-points#contributes.taskDefinitions seems to indicate that the field is required. Do we have real cases of extensions that fail because of this?

@tsmaeder I initially started working on the issue after noticing errors in the console when using `vscode-docker`:

image

I believe a real-world example is `ms-azuretools.vscode-docker`:

- https://github.com/microsoft/vscode-docker/blob/9372dad847505cebac176dd21e8213a5cdf0be14/package.json#L983

In addition, vscode declares the schema like so, so I thought we should be more robust as not to have such errors:

- https://github.com/microsoft/vscode/blob/7392504f85d49732003f1e031c75e63fafae4de2/src/vs/workbench/contrib/tasks/common/taskDefinitionRegistry.ts#L50-L55

tsmaeder:
> In addition, vscode declares the schema like so

So the documentation is lying ;-) Merge away!

paul-marechal (approved):
LGTM code-wise.

Issue #9988, opened 2021-08-30
Add `CancellationError`


tsmaeder:
### Feature Description:
Need to add a `CancellationError` class to the plugin API. https://code.visualstudio.com/api/references/vscode-api#CancellationError

Issue #9136, opened 2021-02-27
Support Rust-Analyzer Plugin


RodogInfinite:
### Bug Description:
No rust-analyzer functionality. Appears in settings and commands, but doesn't work.

**Terminal output:**

> ERROR [hosted-plugin: 32685] Activating extension 'rust-analyzer' failed: Error: Text Content Document Provider for scheme 'rust-analyzer' is already registered
>     at WorkspaceExtImpl.registerTextDocumentContentProvider (/node_modules/@theia/plugin-ext/lib/plugin/workspace.js:274:19)
>     at Object.registerTextDocumentContentProvider (/node_modules/@theia/plugin-ext/lib/plugin/plugin-context.js:488:37)
>     at expandMacro (/plugins/rust-analyzer/extension/out/src/main.js:15024:58)
>     at Ctx.registerCommand (/plugins/rust-analyzer/extension/out/src/main.js:15541:21)
>     at tryActivate (/plugins/rust-analyzer/extension/out/src/main.js:17612:11)
>     at async activate (/plugins/rust-analyzer/extension/out/src/main.js:17540:5)

**Notifications:**
![image](https://user-images.githubusercontent.com/22568404/109397284-02477680-7904-11eb-9f2b-bc28c9dcadf0.png)

### Steps to Reproduce:

1. Add the following line to the theiaPlugins in the package.json
`"rust-analyzer":"https://github.com/rust-analyzer/rust-analyzer/releases/download/2020-11-16/rust-analyzer.vsix"`

- This is the latest version that will install without failing on VS Code version conflict. Example of the failure https://github.com/rust-analyzer/rust-analyzer/releases/download/2020-11-23/rust-analyzer.vsix notifies:
![image](https://user-images.githubusercontent.com/22568404/109397458-a5988b80-7904-11eb-8a4a-0ed1f41cedaa.png)
- The current fix is a bit hacky for this, see: https://github.com/eclipse-theia/theia/issues/8828

2. Build and Start Theia
3. Install Language server version 0.2.384 for rust-analyzer from notification prompt
![image](https://user-images.githubusercontent.com/22568404/109397276-f3f95a80-7903-11eb-9d3a-5c53b3cf81da.png)

### Additional Information

- Operating System: Debian Buster 
- Theia Version: 1.12.0-next.8d5e12ac
- Node Version: v12.14.1-linux-x64
- Yarn Version: v1.22.10
- "@theia/monaco" : "next" in package.json dependencies
[...]

PR #10035, opened 2021-09-01
Add CancellationError

mtsmfm:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/9988.

Add CancellationError to be compatible with VSCode.

https://github.com/eclipse-theia/theia/issues/9136#issuecomment-809660980

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/cancellation.ts
+++ b/packages/core/src/common/cancellation.ts
@@ -48,6 +48,13 @@ export namespace CancellationToken {
     });
 }
 
+export class CancellationError extends Error {
+    constructor() {
+        super('Canceled');
+        this.name = this.message;
+    }
+}
+
 class MutableToken implements CancellationToken {
 
     private _isCancelled: boolean = false;
--- a/packages/plugin-ext/src/plugin/plugin-context.ts
+++ b/packages/plugin-ext/src/plugin/plugin-context.ts
@@ -149,7 +149,7 @@ import { LanguagesExtImpl } from './languages';
 import { fromDocumentSelector, pluginToPluginInfo, fromGlobPattern } from './type-converters';
 import { DialogsExtImpl } from './dialogs';
 import { NotificationExtImpl } from './notification';
-import { CancellationToken } from '@theia/core/lib/common/cancellation';
+import { CancellationToken, CancellationError } from '@theia/core/lib/common/cancellation';
 import { score } from '@theia/callhierarchy/lib/common/language-selector';
 import { MarkdownString } from './markdown-string';
 import { TreeViewsExtImpl } from './tree/tree-views';
@@ -952,7 +952,8 @@ export function createAPIFactory(
             SemanticTokensEdit,
             ColorThemeKind,
             SourceControlInputBoxValidationType,
-            FileDecoration
+            FileDecoration,
+            CancellationError
         };
     };
 }
--- a/packages/plugin/src/theia.d.ts
+++ b/packages/plugin/src/theia.d.ts
@@ -2018,6 +2018,10 @@ declare module '@theia/plugin' {
         onDidDelete: Event<Uri>;
     }
 
+    export class CancellationError extends Error {
+        constructor();
+    }
+
     /**
      * A cancellation token used to request cancellation on long running
      * or asynchronous task.


vince-fugnitto:
@mtsmfm fixes https://github.com/eclipse-theia/theia/issues/9988?

mtsmfm:
@vince-fugnitto Sorry I overlooked that issue. Yes, this PR fixes that.

tsmaeder:
@mtsmfm thanks for the PR. How can I test it? Why I'm asking is this: what happens when a plugin actually throws "CancelledError"? I think we should handle that more gracefully than to just display an error. What does VS Code do in this instance?

paul-marechal:
@tsmaeder VS Code made a function to intercept some calls and log or not some of the errors:

https://github.com/microsoft/vscode/blob/1.60.1/src/vs/workbench/api/common/extHostLanguageFeatures.ts#L1595-L1624

It is then extensively used, but they parameterize it in a few places to ignore logging `CancellationError`:

- https://github.com/microsoft/vscode/blob/1.60.1/src/vs/workbench/api/common/extHostLanguageFeatures.ts#L1927
- https://github.com/microsoft/vscode/blob/1.60.1/src/vs/workbench/api/common/extHostLanguageFeatures.ts#L1941

Since this is just related to logging, I think it is fine to do this as a follow-up.

paul-marechal:
@mtsmfm overall the changes looks ok to me, but do you know which plugin/version I could use to compare the state of master vs your PR? I tried the latest release of `rust-analyzer` but I got a nasty path bug error on start, which I assume comes from our implementations...

tsmaeder:
@mtsmfm Theia is handling errors on remote calls generally here: https://github.com/eclipse-theia/theia/blob/96930a2edcc4c706d41ab8486500de7979d01283/packages/core/src/common/messaging/proxy-factory.ts#L247

I agree we should not delay merging this PR because we're not "properly" handling `CancellationError` because of the following argument: for a plugin that does not use `CancellationError`, nothing changes. For a plugin that _does_ use cancellation error, we now would get a different error each time it tries to throw `CancellationError`. So in both cases, we end up with a log full of bogus error messages. So as far as I'm concerned, it passes the "the code is no worse than before" test.

tsmaeder:
The code looks good to me, but I can't provoke the error from https://github.com/eclipse-theia/theia/issues/9136#issuecomment-809660980 on master. @vince-fugnitto you had analyzed the problem, right?

vince-fugnitto:
> The code looks good to me, but I can't provoke the error from [#9136 (comment)](https://github.com/eclipse-theia/theia/issues/9136#issuecomment-809660980) on master. @vince-fugnitto you had analyzed the problem, right?

@tsmaeder I don't really remember, in the issue all I did was help identify where `CancellationError` came from, and how someone might implement it.

mtsmfm:
@paul-marechal @tsmaeder Currently rust-analyzer compiles its extension for node 14, not 12. So latest rust-analyzer doesn't work because theia only supports 12 until https://github.com/eclipse-theia/theia/pull/9936

Can you try my fork? https://github.com/mtsmfm/rust-analyzer/releases/tag/nightly

It applies a patch to change target

https://github.com/mtsmfm/rust-analyzer/pull/1

---

I tried che-theia with my patch and with my fork (https://gist.github.com/mtsmfm/a9b34a4004d20c38caf8e06f7929feb4) and I confirmed `Class extends value undefined is not a constructor or null` is not appeared.
I got another error now though.

> Activating extension 'rust-analyzer (nightly)' failed: The language client requires VS Code version ^1.53.0 but received version 1.50.0

image

For che-theia, rust-analyzer still doesn't work since its dependency requires vscode ^1.53.0.

https://github.com/rust-analyzer/rust-analyzer/blob/f1d7f98ed07b9934286b9c4809dd4d7a47537879/editors/code/package-lock.json#L3764
https://github.com/microsoft/vscode-languageserver-node/blob/release/client/7.1.0-next.5/client/src/node/main.ts#L25

I think if you can use latest theia, it should work since theia's default is 1.53.2 now.
https://github.com/eclipse-theia/theia/pull/9959

For che-theia, I think this patch is needed
https://github.com/eclipse-che/che-theia/pull/1213

mtsmfm:
Or perhaps this release can work without patch since it's the last release which supports node 12

https://github.com/rust-analyzer/rust-analyzer/releases/tag/2021-08-09

tsmaeder:
Luckily, the newest vscode-java uses `CancellationError`.

packages/plugin-ext/src/plugin/quick-open.ts
```diff
@@ -40,6 +40,13 @@ export function isPromiseCanceledError(error: any): boolean {
     return error instanceof Error && error.name === canceledName && error.message === canceledName;
 }
 
+export class CancellationError extends Error {
```
tsmaeder, line 43: I just noticed that this is in quick-open.ts. However, this is a general class and should be moved to `types-impl.ts`

tsmaeder (changes requested):
I was able to test with vscode-java v.0.82.0, but we should move the type declaration.

packages/plugin-ext/src/plugin/quick-open.ts
tsmaeder: Let me correct that: we're consuming cancellation-related stuff from core/common/cancellation.ts. We probably put the implementation of the class there, as well.

mtsmfm:
@tsmaeder I've moved class into core/common/cancellation.ts. Can you review again?

tsmaeder (approved):
lgtm now

PR #9560, opened 2021-06-07
Enhancement : Avoid electron windows to overlap

OmarSdt-EC:
With this enhancement, an electron window, when it's newly opened, has
different x and y coordinates than all the previous ones. Therefore,
there's no overlap between all the opened electron windows.

#### What it does

Fixes #7431

With this enhancement, an electron window, when it's newly opened, has
different x and y coordinates than all the previous ones. Therefore,
there's no overlap between all the opened electron windows.

#### How to test

1. Open a new electron window (by opening a folder, a file or a workspace for example)
2. The new window has normally different x and y coordinates than the previous one and there's no overlap.

#### Review checklist

- [ ] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #9186, opened 2021-03-11
Initial contribution of @theia/external-terminal

DukeNgn:
## What it does

+ Introduces a new extension: `@theia/external-terminal`. The extension adds support for spawning external terminals from Electron applications. In the future, the extension can be extended to support features such as spawning external terminals during debug sessions.
(**Browser applications should not pull this extension**).

#### Checklist for merging:
- [x] [CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23140) registered and **approved**
- [x] #9169 needs to be merged first since this PR depends on the ability for late binding in preference proxy.

#### Command:
+ Closes #8712 : Supports opening native terminals by calling `Open New External Terminal` from the command palette. 

#### Preferences:
+ The PR contributes the following preferences:
	- `terminal.external.windowsExec`
	- `terminal.external.osxExec`
	- `terminal.external.linuxExec`
to allow users to customize the desired terminal application to run on their current platform. 
+ On startup, the backend determines the default terminal on the host machine, then sets the default field in the preference schema accordingly.
+ Changing the `Exec` preference of other OS than the current one won't have any effect. (i.e: Changing `osxExec` while running on Windows doesn't have any effect).

## How to test:
+ Start **Theia Electron application**
+ Verify the default `exec` of host OS is displaying the correct terminal.

Calling command `Open New External Terminal` has different behaviors, depending on the current workspace status:

#### No workspaces opened and no current active editor -> Spawn the terminal at the user home directory.

https://user-images.githubusercontent.com/43587865/110700880-76183780-81be-11eb-9a89-a85e73fe18e3.mp4

#### No workspaces opened but has an active editor -> Spawn the terminal at the parent folder of the active editor file.

https://user-images.githubusercontent.com/43587865/110700923-83352680-81be-11eb-9f74-2bedbf28c081.mp4

#### Only one workspace opened -> Spawn the terminal at the root of the current workspace.

https://user-images.githubusercontent.com/43587865/110700788-5bde5980-81be-11eb-99c8-2eacd73f0a72.mp4

#### Multi-root workspace opened -> Displays a quick pick to let users choose which workspace to spawn the terminal.

https://user-images.githubusercontent.com/43587865/110700703-40734e80-81be-11eb-8423-d6e72130b811.mp4

## Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

[...]

PR #10044, opened 2021-09-03
Support vscode's `titleBarStyle`

msujew:
#### What it does

Adds support for vscode's `window.titleBarStyle` setting that allows to use a browser-like menu in the electron environment:

https://user-images.githubusercontent.com/4377073/131932222-3654effd-86ef-4ed1-aed6-d05da5c0df25.mp4

Note that this is only supported in **Windows** and **Linux** distributions. I actually have no idea how it works on MacOS.

#### How to test

1. Start Electron Theia on either Windows or Linux.
2. On Windows, Theia will use by default the `custom` (browser-like) title bar. On Linux the `native` one.
3. Assert that switching between the preference values is possible. The application has to be restarted **manually** afterwards.
4. Assert that the window functions are working correctly. This includes the newly added buttons, the draggable menu bar, and the resizable window.
4. Assert that both `custom` and `native` state work correctly with the different configurations of the `window.menuBarVisibility` setting.

#### Review checklist

#### Reminder for reviewers


packages/core/src/browser/core-preferences.ts
```diff
@@ -72,6 +72,18 @@ export const corePreferenceSchema: PreferenceSchema = {
...
+            enum: ['native', 'custom'],
+            markdownEnumDescriptions: [
+                'Native title bar is displayed.',
+                'Custom title bar is displayed.'
+            ],
+            default: isWindows ? 'custom' : 'native',
+            scope: 'application',
+            markdownDescription: 'Adjust the appearance of the window title bar. Changes require a full restart to apply',
+            included: !isOSX
+        },
```
vince-fugnitto, line 86: @msujew we probably only want to support this preference in the desktop (electron) use-case? Should it be moved to an electron-only preference-schema?

packages/core/src/browser/core-preferences.ts
```diff
@@ -72,6 +72,18 @@ export const corePreferenceSchema: PreferenceSchema = {
             A setting of 'compact' will move the menu into the sidebar.`,
             included: !isOSX
         },
+        'window.titleBarStyle': {
```
vince-fugnitto, line 75: In vscode, the preference also controls the display of context-menus, meaning we can have browser-like menus in electron instead of native. That's why when I first thought of the issue I was thinking we can control the binding we do to not bind native electron menus and instead use that of the browser:

https://github.com/eclipse-theia/theia/blob/068e78d34f88f274c9c7f9ddd64299fd7796589d/packages/core/package.json#L123-L124

With a bit more extra work, we can simply use the browser menus in electron (main menu, and context menus) by modifying the way we bind.

vince-fugnitto (commented):
Starting the application on Ubuntu I am getting some odd behavior with my app (no electron top-menu):

image

msujew:
> Starting the application on Ubuntu I am getting some odd behavior with my app (no electron top-menu)

Should be fixed now 👍

packages/core/src/browser/core-preferences.ts
msujew: Yes, that makes sense 👍

packages/core/src/browser/core-preferences.ts
msujew: Do we want to have browser-like menus in Electron always? I'm not quite sure how we would be able to accomplish that during the binding phase, as we need the preference service to identify which title bar style is selected.

vince-fugnitto: @msujew I believe dynamic binding might be possible, and is something we likely did in https://github.com/eclipse-theia/theia/pull/9186.

As for the behavior, I would expect that:
- `native` displays the native menus (main-menu and context-menus) similarly to vscode
- `custom` displays the browser-like menus (main-menu and context-menus) similarly to vscode

You're right that there might be issues especially on startup, but perhaps it's something to try.
cc @paul-marechal any thoughts?

msujew: You reminded me that the context menus where not displayed in the browser styling if `custom` is used. I addressed that.

> I believe dynamic binding might be possible, and is something we likely did in #9186.

I only saw conditional binding based on `os` based constants. That works fairly well. However, during binding we have no access to preferences as they are 1. a service in themselves 2. we would have to wait for the async `ready` state.

vince-fugnitto (commented):
@msujew when updating the preference, we probably want to ask users to confirm a restart of the application similarly to vscode (confirmation dialog)?

I was able to get the following state:
- update preference from `native` to `custom`
- execute the command `reload application`

image

---

> You reminded me that the context menus where not displayed in the browser styling if custom is used. I addressed that.

The custom context-menus (with `custom` enabled) do not appear for me on Ubuntu.

msujew:
> when updating the preference, we probably want to ask users to confirm a restart of the application similarly to vscode (confirmation dialog)?

Yes, I also thought about that. I tried implementing it, but got discouraged by the electron app restarting without starting the backend server. I'll take another look at it 👍 

> The custom context-menus (with custom enabled) do not appear for me on Ubuntu.

Weird, it works fine for me on Windows:

image

Time to bring out my old Ubuntu vm...

vince-fugnitto:
> Weird, it works fine for me on Windows:

Hopefully the video helps :)

https://user-images.githubusercontent.com/40359487/132030147-e8c8ad1a-8f96-426c-8129-0b663ecbcc1b.mp4

msujew:
> Hopefully the video helps :)

I got that part, but do have a video of the native context menus as well? ;)

vince-fugnitto:
> I got that part, but do have a video of the native context menus (on `custom`) as well? ;)

Sure! 

https://user-images.githubusercontent.com/40359487/132031159-846ef168-471f-4c90-b427-caf3eab425b2.mp4

msujew:
> when updating the preference, we probably want to ask users to confirm a restart of the application similarly to vscode (confirmation dialog)?

I managed to implement that without the whole application crashing. However, the `onPreferenceChanged` event handler fires instantly on startup on all preferences (and therefore always shows a restart dialog). Is this a feature or a bug?

colin-grant-work (commented):
> Is this a feature or a bug?

Both :-/. Ideally, the initial setting of preferences _wouldn't_ fire an event, and they would just take on those values and then the preference service would resolve its `.ready` promise and everyone would see the initial values and work from there. In practice, because there's a (reasonably) predictable flood of events (under most conditions), there are a number of places where code has been written in such a way to depend on those events, in the framework and almost certainly downstream. The bug earlier this week re: the browser menu bar was an example of the consequences of such code, but it would almost certainly be a pain in adopters' butts to remove that flood of events, however unsafe it is to rely on it in general.

msujew:
> this particular case, you should be able to do something like setting a flag when you first create the menu bar and not showing the dialog before that

Thank you for the explanation, I implemented something similiar that ignores the first wave of preference changes 👍 

However, the issue Vince experienced (changing preference -> ignoring the dialog -> resetting the view) still persists, I'll try to find a solution for that as well.

Also, one other weird thing: Changing between the preference > 1 time leads to exponential growth in windows, I don't know what's going on. 😕

msujew:
@vince-fugnitto I can't reproduce your context-menu issue using Ubuntu 20.04 (running emulated on Windows):

image

Do you mind retesting it? I think I addressed every other issue.

vince-fugnitto (commented):
I think the behavior I was noticing was something in cache, since if I switch between two different electron instances (ex: master and this pr) I end up with odd behaviors.

One thing I did notice is the following:

1. start the application, update the preference to `custom`
2. accept the restart prompt (the app restarts with the new menu)
3. close the application
4. modify the preference in the user's home (remove the `custom` titleBarStyle preference)
5. when restarting the application I get back the `custom` menu

image

It's only when I remove the electron cache do I see the proper menu again:

```bash
$ rm -rf ~/.config/Theia\ Electron\ Example/
```

I'm not sure if there is anything we can improve in this aspect, but I don't think it should block the pull-request, just a bug end-users might experience.

packages/core/src/electron-browser/menu/electron-main-menu-factory.ts
```diff
@@ -92,15 +86,16 @@ export class ElectronMainMenuFactory {
...
             electron.remote.Menu.setApplicationMenu(createdMenuBar);
-        } else {
+        } else if (this.preferencesService.get('window.titleBarStyle') === 'native') {
+            const createdMenuBar = this.createElectronMenuBar();
             electron.remote.getCurrentWindow().setMenu(createdMenuBar);
         }
     }
 
-    createMenuBar(): Electron.Menu | null {
+    createElectronMenuBar(): Electron.Menu | null {
```
vince-fugnitto, line 98: Should we mark as a breaking change?
The same for:
- `createContextMenu`
- `handleDefault`

packages/core/src/electron-browser/window/electron-window-preferences.ts
```diff
@@ -38,11 +39,24 @@ export const electronWindowPreferencesSchema: PreferenceSchema = {
...
+        'window.titleBarStyle': {
+            type: 'string',
+            enum: ['native', 'custom'],
+            markdownEnumDescriptions: [
+                'Native title bar is displayed.',
+                'Custom title bar is displayed.'
+            ],
+            default: isWindows ? 'custom' : 'native',
+            scope: 'application',
+            markdownDescription: 'Adjust the appearance of the window title bar. Changes require a full restart to apply',
```
vince-fugnitto, line 51: We might want to align here, and reference context-menus which it also updates:

```suggestion
            markdownDescription: 'Adjust the appearance of the window title bar. On Linux and Windows, this setting also affects the application and context menu appearances. Changes require a full restart to apply.',
```

msujew:
> I'm not sure if there is anything we can improve in this aspect

I don't think so either. We don't have access to the users preferences during startup, which would be needed in order to display the correct window title. Instead, everything is saved in the `windowStore`.

I also just noticed that the context menu uses the default electron menu if you set the preference from `custom` to `native` (without restarting) and reset the view. I fixed that as well 👍

vince-fugnitto:
> I don't think so either. We don't have access to the users preferences during startup, which would be needed in order to display the correct window title. Instead, everything is saved in the `windowStore`.

@msujew unfortunately, the `window.titleBarStyle` preference does not ultimately behave like other preferences in the framework for that reason. It means that downstream applications cannot set the default value for the preference (in the `package.json`) as it will not apply like others.

The following for instance does not work for me:

```json
"theia": {
    "target": "electron",
    "frontend": {
      "config": {
        "applicationName": "Theia Electron Example",
        "preferences": {
          "window.titleBarStyle": "custom"
        }
      }
    }
  },
```

The app starts with `custom` but it is never applied (even on subsequent starts).

msujew:
> unfortunately, the window.titleBarStyle preference does not ultimately behave like other preferences in the framework for that reason.

Actually, that gave me the insight I needed on that matter: I tried fitting it into the usual way of working with preferences. But as it doesn't work like any other preference, instead of using the `PreferenceService` as the source of truth, I now use the electron application as the source of truth and override the preference if necessary. Thank you :)

Also I addressed the issue related to the config in the `package.json`. We have access to that in our electron app, so we can set the title bar according to the config.

packages/core/src/electron-browser/menu/electron-menu-contribution.ts
```diff
@@ -119,6 +120,31 @@ export class ElectronMenuContribution implements FrontendApplicationContribution
...
+                this.preferenceService.set('window.titleBarStyle', this.titleBarStyle, PreferenceScope.User);
+            });
+        });
+        electron.ipcRenderer.send('request-titleBarStyle');
+        this.preferenceService.ready.then(() => {
+            this.setMenu(app);
+            electron.remote.getCurrentWindow().setMenuBarVisibility(true);
+            setTimeout(() => {
+                this.titleBarStyleChangeFlag = true;
+            }, 1000);
```
colin-grant-work, line 137: This `setTimeout` is a little odd. What's the significance of 1 second in this context?

msujew: I removed the timeout. It was no longer necessary. We can set the flag after the first change in preferences happens.

vince-fugnitto (commented):
The changes work well for me, there are some minor problems still with caching but I don't believe this is a preference that would be updated frequently. 

I also confirmed that if an application sets the `titleBarStyle` as a default preference in the `package.json` then it would be applied correctly :+1:

msujew:
@vince-fugnitto Is this an approval? :)  
@colin-grant-work Do you still plan on reviewing it?

It would be great, if we were able to include it in this month's release. Although, I don't have an issue with delaying it.

colin-grant-work:
@msujew, Sorry, got distracted. I'll take a look later today.

packages/core/src/electron-browser/menu/electron-menu-style.css
```diff
@@ -0,0 +1,87 @@
...
+
+#minimize-button {
+  grid-column: 1;
+}
+#maximize-button, #restore-button {
+  grid-column: 2;
+}
+#close-button {
+  grid-column: 3;
+}
```
colin-grant-work, line 59: Minor

```suggestion
#minimize-button {
  grid-column: 1;
}

#maximize-button, #restore-button {
  grid-column: 2;
}

#close-button {
  grid-column: 3;
}
```

packages/core/src/electron-browser/menu/electron-menu-style.css
```diff
@@ -0,0 +1,87 @@
...
+  color: white;
+}
+
+#restore-button {
+  display: none !important;
+}
+
+.maximized #restore-button {
+  display: flex !important;
+}
```
colin-grant-work, line 83: Are these `!important` specifications necessary? If possible I'd prefer something like `.not-maximized #restore-button { display: none; }` vs. `.maximized #restore-button { display: flex; }`, for example.

packages/core/src/electron-main/electron-main-application.ts
```diff
@@ -182,6 +183,10 @@ export class ElectronMainApplication {
     readonly backendPort = this._backendPort.promise;
 
     protected _config: FrontendApplicationConfig | undefined;
+    protected frame: boolean = true;
```
colin-grant-work, line 186: Another case where I'll dare to disagree with the established terminology. I think this would be clearer as something like `useNativeWindowFrame`

packages/core/src/electron-main/electron-main-application.ts
```diff
@@ -182,6 +183,10 @@ export class ElectronMainApplication {
...
+    protected originalFrames = new Map<number, boolean>();
```
colin-grant-work, line 187: Maybe `didUseNativeWindowFrameOnStart`? Very prolix, but clear, I hope

packages/core/src/electron-main/electron-main-application.ts
```diff
@@ -477,6 +508,19 @@ export class ElectronMainApplication {
...
+            this.frame = titleBarStyle === 'native';
+            this.saveWindowState(BrowserWindow.fromId(sender.id));
+        });
+
+        ipcMain.on('restart', ({ sender }) => {
+            this.restart(sender.id);
+        });
+
+        ipcMain.on('request-titleBarStyle', ({ sender }) => {
+            sender.send('original-titleBarStyle', this.originalFrames.get(sender.id) ? 'native' : 'custom');
```
colin-grant-work, line 522: consider extracting `request-titleBarStyle` and `original-titleBarStyle` (or `titleBarStyleAtStartup`) to constants to be used here and in the code that makes the request from the `electron-browser` code.

packages/core/src/electron-main/electron-main-application.ts
```diff
@@ -495,7 +539,22 @@ export class ElectronMainApplication {
...
+        this.restarting = true;
+        BrowserWindow.fromId(id).close();
+        setTimeout(async () => {
+            await this.launch({
+                secondInstance: false,
+                argv: this.processArgv.getProcessArgvWithoutBin(process.argv),
+                cwd: process.cwd()
+            });
+            this.restarting = false;
+        }, 500);
```
colin-grant-work, line 557: What's the significance of the `setTimeout` here? Could it be extracted into some sort of one-time event listener?

colin-grant-work (commented):
This seems to be working well for me. I was able to switch between modes without issue, and the behavior was correct.

Re: styling, the Theia icon certainly, and I think the other menu bar icons less obviously, are hovering a little higher than they should. It looks like about 10px below and only 6px above.

image

packages/core/src/electron-main/electron-main-application.ts
```diff
@@ -348,31 +373,37 @@ export class ElectronMainApplication {
...
+            this.originalFrames.delete(electronWindow.id);
+        });
         electronWindow.on('resize', saveWindowStateDelayed);
         electronWindow.on('move', saveWindowStateDelayed);
+        this.originalFrames.set(electronWindow.id, this.frame);
+    }
+
+    protected saveWindowState(electronWindow: BrowserWindow): void {
+        try {
+            const bounds = electronWindow.getBounds();
```
colin-grant-work, line 394: I did run into an error during one of my first runs through, but I couldn't reproduce it:
```
Error while saving window state: TypeError: Cannot read property 'getBounds' of null
```

packages/core/src/electron-main/electron-main-application.ts
msujew: The `BrowserWindow.close` call isn't instant and therefore we have two windows open for a really short duration. This isn't a problem for the user usually, since it looks like the old window swaps with the new one. However, due to #9560, having two windows on the same spot isn't allowed. So Theia will instead move the second window to a new location. If we wait a bit, the first window will successfully close and the new window will appear at the original position.

But you're right, this can be successfully moved to the `on('closed')` event listener.

packages/core/src/electron-main/electron-main-application.ts
msujew: It seems like `BrowserWindow.fromId()` can sometimes return `null`. I added a guard for that 👍

msujew:
> Re: styling, the Theia icon certainly, and I think the other menu bar icons less obviously, are hovering a little higher than they should. It looks like about 10px below and only 6px above.

@colin-grant-work I agree, but this also appears in the browser version (`master`):

image

I think this is unrelated to my changes and should be addressed in a separate PR.

colin-grant-work:
> I think this is unrelated to my changes and should be addressed in a separate PR.

Sounds fair.

> But that is kind of related to Theia's top panel being 32px high, while vscode's is only 30px high. Should I change this to 30px in Theia as well?

I don't have strong feelings here. I certainly have no objection in principle to `30px`, but I don't think the `1px` shift is objectionable either, and there is always the chance that downstream (or other upstream) code / CSS expects `32px`. For now, I'd probably opt not to worry about changing it.

packages/core/src/electron-browser/menu/electron-menu-style.css
colin-grant-work: Removing `!important` has rendered the `#close-button:hover` (ID + pseudoclass (0,1,1,0) specificity) rule inoperative, because it's competing with `#window-controls .control-button:hover` (ID + class + pseudoclass (0,1,2,0)). You could change it to `#window-controls #close-button:hover` (0,2,1,0) or change `window-controls` to a class rather than an ID, since `.window-controls .control-button:hover` would be have 0,0,3,0 specificity. Our CSS isn't very standardized or especially well organized, so you can choose how you'd like to resolve it. In general, I shy away from ID-scoped styling and `!important`, if they can be avoided.

colin-grant-work: @msujew, just this minor thing to fix, and then I think this is good to go.

msujew: @colin-grant-work Thanks, didn't notice that :)

I think that should be everything.

colin-grant-work (approved):
This looks good to me. Good work finding a way to get electron to play nicely.

colin-grant-work:
@msujew, @vince-fugnitto, this is working well for me, and the code looks good. Depending on your feelings about its complexity / likelihood of bugs, it might be worth putting off merging until after tomorrow's release?

msujew:
> It might be worth putting off merging until after tomorrow's release?

Sure, fine with me 👍

orange-pig:
> Starting the application on Ubuntu I am getting some odd behavior with my app (no electron top-menu):
> 
> image

I get the same problem now, when I try upgrade theia to 1.23.0. from 1.18.0. 

After build the electron and start, I got it. No titlebar & top menu, only the content be display.

And I try to set window: Title bar style custom -> native, it will be have a windows tile bar. (other: The restart window look like do not show always.) 

So, the window title bar style be default custom in new version, but it can't work in my PC, and then build electron installer running in other PC still so.

And, I try to delete .theia/ in user folder, still so.
And, I try to delete electron/ in AppData folder, still so.

I am using:
node.js: 14.18.1
typescript: 4.5.5

kittaakos:
@msujew, was there a technical obstacle to allowing the custom title bar on macOS?

https://github.com/eclipse-theia/theia/blob/52d6e877371204045d9b047b39f3d54672387258/packages/core/src/electron-browser/window/electron-window-preferences.ts#L50

VS Code supports it. Thank you!

msujew:
@kittaakos No real technical reason. I just don't have a macOS machine to test it. So we should still support it sometime on macOS.

kittaakos:
Excellent, thank you! We must implement it downstream, and if it works, we will contribute it back.

PR #10055, opened 2021-09-06
fix updating items in the quick-pick menu

vinokurig:
Signed-off-by: Igor Vinokur <ivinokur@redhat.com>

#### What it does

Fixes a bug when quick pick items are not updated on the fly.

When a quick open widget is shown, set an instance of the monaco quick-input to the [sessions map](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/main/browser/quick-open-main.ts#L195).
When plugin request's a quick-input items update, they are set to the monaco quick-input instance from the map:
https://github.com/eclipse-theia/theia/blob/d3501165bb4e87c3612a1a02c34a1d16ab81802c/packages/plugin-ext/src/main/browser/quick-open-main.ts#L257-L262

#### How to test

1. Clone and compile the test plugin: https://github.com/microsoft/vscode-extension-samples/tree/main/quickinput-sample
2. Start hosted plugin
3. Execute the `Quick Input Samples` from the command palette.
4. Choose `quickOpen` and type something to the input box.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/quick-input/quick-input-service.ts
+++ b/packages/core/src/browser/quick-input/quick-input-service.ts
@@ -206,6 +206,7 @@ export interface QuickInputService {
         Promise<(O extends { canPickMany: true } ? T[] : T) | undefined>;
     showQuickPick<T extends QuickPickItem>(items: Array<T>, options?: QuickPickOptions<T>): Promise<T>;
     hide(): void;
+    getCurrentInput(): QuickInput | undefined;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


tsmaeder:
@vinokurig can you describe what the fix does in the code? I.e. what is that basic idea of the code changes in the fix?

vinokurig:
@tsmaeder 
> @vinokurig can you describe what the fix does in the code? I.e. what is that basic idea of the code changes in the fix? Also, what was the cause of the bug? Remember to "write shit down" (TM)

Updated the description

tsmaeder:
@vinokurig I think your change takes the code in the wrong direction: instead of adding "currentInput" state into the `QuickInputService`, we should add a "createQuickPick" method to the interface and use that in the `quick-open-main.ts#$showCustomQuickPick`.  The current change fails when someone calls `QuickInputService#showQuickPick` twice in a row. There have been a number of things wrong with this area of the code for a long time which are not your fault, but we should refrain from making the situation worse.

tsmaeder (changes requested):
See comment on the main page for the reasons.

vinokurig:
Opened a PR with different approach https://github.com/eclipse-theia/theia/pull/10065

PR #10073, opened 2021-09-09
`CodeAction` and `CodeActionProviderMetaData` fields added

### LINKED ISSUE #9989: Implement missing `CodeAction` fields
### Feature Description:
* Add the `isPreferred` and `disabled` fields in `CodeAction` 
* Add the `documentation` field in `CodeActionMetadataProvider `

(https://code.visualstudio.com/api/references/vscode-api#CodeAction). AndWith a bit of luck, the semantics are already implemented through the editor. In particular menu enablement might not be.

Archie27376:
#### What it does

Fixes #9989

- added support for the `disabled` and `isPreferred` fields to the `CodeAction` interface
- added support for the `documentation` field to `CodeActionProviderMetaData` interface

#### How to test

1. include the [vscode-code-action-test](https://github.com/vince-fugnitto/vscode-code-action-test/releases/download/0.0.1/code-actions-sample-0.0.2.vsix) test plugin
2. start the application and open a markdown file
3. type `:)` in the file
4. the quick-fix should display entries - the cat emoji entry is explicitly hidden (will be present on master)
   - [reference](https://github.com/vince-fugnitto/vscode-code-action-test/blob/54dd7898b5f0834eb70bca56c6a267b67a71d2a9/src/extension.ts#L48)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin/src/theia.d.ts
+++ b/packages/plugin/src/theia.d.ts
@@ -7523,6 +7523,16 @@ declare module '@theia/plugin' {
         kind?: CodeActionKind;
 
         /**
+         * Marks that the code action cannot currently be applied.
+         */
+        disabled?: { reason: string };
+
+        /**
+         * Marks this as a preferred action.
+         */
+        isPreferred?: boolean;
+
+        /**
          * Creates a new code action.
          *
          * A code action must have at least a [title](#CodeAction.title) and [edits](#CodeAction.edit)
@@ -7571,6 +7581,13 @@ declare module '@theia/plugin' {
          * may list our every specific kind they provide, such as `CodeActionKind.Refactor.Extract.append('function`)`
          */
         readonly providedCodeActionKinds?: ReadonlyArray<CodeActionKind>;
+
+        /**
+         * Documentation from the provider is shown in the code actions menu
+         *
+         * At most one documentation entry will be shown per provider.
+         */
+        documentation?: ReadonlyArray<{ command: Command, kind: CodeActionKind }>
     }
 
     /**
--- a/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
+++ b/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
@@ -343,6 +343,8 @@ export interface CodeAction {
     edit?: WorkspaceEdit;
     diagnostics?: MarkerData[];
     kind?: string;
+    disabled?: { reason: string };
+    isPreferred?: boolean;
 }
 
 export interface CodeActionContext {
--- a/packages/plugin-ext/src/main/browser/languages-main.ts
+++ b/packages/plugin-ext/src/main/browser/languages-main.ts
@@ -1013,6 +1013,7 @@ function toMonacoAction(action: CodeAction): monaco.languages.CodeAction {
     return {
         ...action,
         diagnostics: action.diagnostics ? action.diagnostics.map(m => toMonacoMarkerData(m)) : undefined,
+        disabled: action.disabled?.reason,
         edit: action.edit ? toMonacoWorkspaceEdit(action.edit) : undefined
     };
 }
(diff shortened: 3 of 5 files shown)


msujew:
@tsmaeder Do you have an example extension at hand with which one can test these changes?

vince-fugnitto (commented):
@Archie27376 using the test plugin [vscode-code-action](https://github.com/vince-fugnitto/vscode-code-action-test/releases/download/0.0.1/code-actions-sample-0.0.2.vsix) I was able to confirm at least for the `CodeAction` changes that there is still work to do. The `cat` result in the quick-fix should not be displayed (similarly to vscode) as it is explicitly disabled. The fix is quite simple however.

vince-fugnitto (approved):
I confirmed that the behavior works correctly with the test plugin :+1: 

The `cat` result is explicitly disabled and not visible similarly to vscode:

https://user-images.githubusercontent.com/40359487/133617453-bc79216c-b3ae-4239-9b52-ebf8c49bf1c2.mp4

packages/plugin/src/theia.d.ts
```diff
@@ -7571,6 +7581,13 @@ declare module '@theia/plugin' {
          * may list our every specific kind they provide, such as `CodeActionKind.Refactor.Extract.append('function`)`
          */
         readonly providedCodeActionKinds?: ReadonlyArray<CodeActionKind>;
+
+        /**
+         * Documentation from the provider is shown in the code actions menu
+         *
+         * At most one documentation entry will be shown per provider.
+         */
+        documentation?: ReadonlyArray<{ command: Command, kind: CodeActionKind }>
```
tsmaeder, line 7590: I don't see how the documentation field is transferred to the main side and ultimately the monaco editor.

tsmaeder (changes requested):

packages/plugin/src/theia.d.ts
vince-fugnitto: I think we can omit support for `documentation` for now, it is used when registering `registerCodeActionsProvider` but our implementation differs from vscode in many ways:

https://github.com/eclipse-theia/theia/blob/0f7dec1acc4eb738111ed33d41bb5bd47e9bf897/packages/plugin-ext/src/plugin/languages.ts#L441-L456

- https://github.com/microsoft/vscode/blob/abfaf30a04ffd25f9b3a835479379329ab947495/src/vs/workbench/api/common/extHostLanguageFeatures.ts#L1785-L1797

tsmaeder: But ultimately, we're registering the code action provider to the editor, which _does_ have a documentation field. We should pass this info to the editor in Theia, as well.

JonasHelming: @tsmaeder

tsmaeder: Our definition of modes.CodeActionProvider is _way_ out of date. We need to update our definition in monaco.d.ts and pass the necessary information to the front end through "registerQuickFixProvider". The current definition in monaco actually looks like this: https://github.com/microsoft/vscode/blob/2586299c42490a8ed6dff96ed731d44fcbe5b20e/src/vs/editor/common/languages.ts#L777

tsmaeder: I looks like in the branch we currently use (0.23) of vs code, the definition lives here: (https://github.com/microsoft/vscode/blob/b882940dc6d5b2acb096969b9e43385248719df7/src/vs/editor/common/modes.ts#L740). 

I believe it would work to extend the definition of CodeActionProvider in `monaco.d.ts` with the needed bits.

JonasHelming: @Archie27376 : Please have a look at the hints that Thomas provided and let us know if you need further information!

vince-fugnitto: @JonasHelming Dante's (@Archie27376) internship has terminated, someone else would need to pick up the pull-request.

PR #9538, opened 2021-06-01
Implement internationalization support

msujew:
#### What it does

Closes #1327

Adds internationalization support for Theia. The following features are supported:

- [x] Support for [vscode language packs](https://code.visualstudio.com/docs/getstarted/locales) to localize Theia itself.
- [x] Support for localized vscode extensions (both `package.json` and runtime translations via `nls.localize()`)
- [x] Support for custom Theia localizations through a `LocalizationContribution` interface.
- [x] Support for localization of the monaco editor (search widget, context menu and so on).
- [x] Showing the original text under the localized value (for commands).

Things that are out of scope for this PR (and will probably be addressed in another PR):

- Support for vscode language packs to localize other vscode extensions (like the builtins)

#### How it works

1. The `core/browser` package contains a `nls` namespace with a `localize` function similiar to vscode. It accepts a translation key, a default value (usually the original english value) and additional args. The translations are loaded before any other imports so that they are available during load-time of any other code.
2. The backend application loads vscode language packs and Theia's `LocalizationContribution`s and places them in a `LocalizationProvider`. Additionally, the backend exposes a `/i18n/<locale>` HTTP endpoint to retrieve localizations for any locale.
3. The vscode plugin host starts the vscode extension process with the locale that is currently used by the user. This allows vscode extensions to translate themselves.
4. The `monaco-loader.ts` uses the frontend locale as well to configure monaco to use the currently active language.

#### How to test

0. Test that the `Configure display language` command can be invoked but only displays `en` (even though german translations exist in the backend, but no language pack is installed)
1. Download & install the german language pack (v1.50, search for `german` in the Theia extension menu)
2. Download & install [the i18n sample extension in german](https://github.com/eclipse-theia/theia/files/6658579/i18n-sample-0.1.0.zip)
3. Execute the `Hello` and `Bye` command contributed by the i18n sample extension. Those will simply show `Hello` and `Bye` respectively in a notification.
4. Execute the `Configure Display Language` command and choose the german locale `de`.
5. The application will be reloaded (you will maybe have to reload it yourself) and partially translated into german. You will see this in the main menu as well as the workspace and some common commands. *(Test for vscode localization packs)*
6. Execute the `Hallo` (Hello) and `Auf Wiedersehen` (Bye) command, which will now display german text as a notification. *(Test for vscode extensions to correctly localize themselves and for Theia to read the translated `package.json` correctly)*
[...]

PR #10084, opened 2021-09-10
Enable localization of monaco-editor

msujew:
#### What it does

Fixes one of the open issues from #9538: Enables the localization of the monaco-editor using vscode language packs.

Monaco is using a quite complicated localization mechanic. We are able to inject our custom translation into monaco using the `loadBundle` function. The callback receives a localization bundle in the form of `Record<string, string[]>` which is build from the language pack (which uses the form of `Record<string, Record<string, string>>`). The bulk of the work translating between these formats is done using a script (`translate-monaco.js`).

The script tries to match the indices of the currently used monaco translation (array based) to the provided language pack. This works mostly fine, aside from two special cases (`keybindingLabels` and `findWidget`). There is a special duplicating logic inside of the script, that needs to be used due to duplicated entries inside of the monaco `nls.js` files.

There would be an alternative approach to translating monaco: Update the `@theia/monaco-editor-core` package to natively include the fitting translations. That would alleviate the need for all of this. However, the drawback of that method would be that only selected localizations would be available. For example, the `monaco-editor` package on npm does not provide a turkish translation. Using the turkish language pack with the current proposed approach translates monaco into turkish perfectly. We are able to provide arbitrary translations for monaco using language packs this way.

The current proposal comes with another drawback: The provided `monaco-nls.json` has to be updated with every monaco update. The scripts should be able to deal with that, but there may be need for small adjustments in the script itself. Another possible drawback is that future language pack versions might not be compatible with the old monaco-editor translations. I couldn't observe any issues with that using a language version from vscode `1.60.4`, so we don't have to worry about that for quite some time.

#### How to test

1. Download a language pack of your choice from the ovsx-registry (version `1.53.2` or newer).
2. Use the configure language command to change to that language.
3. Confirm that the monaco related parts of Theia have been correctly translated:
    - The editor context menu
    - The find widget
    - The `Selection` main menu entry and its associated commands
4. You can also try the script itself using the instructions provided in the script itself.

#### Review checklist

#### Reminder for reviewers


packages/monaco/data/monaco-nls.json
```diff
@@ -0,0 +1,1379 @@
...
+        "_duplicate/15": "Could not redo '{0}' across all files. {1}",
+        "cannotWorkspaceRedoDueToChanges": "Could not redo '{0}' across all files because changes were made to {1}",
+        "cannotWorkspaceRedoDueToInProgressUndoRedo": "Could not redo '{0}' across all files because there is already an undo or redo operation running on {1}",
+        "cannotWorkspaceRedoDueToInMeantimeUndoRedo": "Could not redo '{0}' across all files because an undo or redo operation occurred in the meantime",
+        "cannotResourceRedoDueToInProgressUndoRedo": "Could not redo '{0}' because there is already an undo or redo operation running."
+    },
+    "vs/platform/workspaces/common/workspaces": {
+        "codeWorkspace": "Code Workspace"
+    }
+}
```
vince-fugnitto, line 1379: minor: newline

packages/monaco/data/monaco-nls.json
```diff
@@ -0,0 +1,1379 @@
+{
```
vince-fugnitto, line 1: @msujew I assume the file was copied from vscode? Do you know if a past CQ covered the file, else I believe we should register one:
- https://github.com/eclipse-theia/theia/wiki/Registering-CQs#codecontent-is-copied-from-other-projectsrepos

msujew: No, this file is generated by the `translate-monaco` script that I wrote. It combines the index based approach from monaco with the key-value pair based approach from the language packs, in order to localize monaco. This file is basically just a generated bridge to connect them both.

vince-fugnitto: Thank you for clarifying!

vince-fugnitto (approved):
I confirmed that I see localizations (french extension-pack) in the following places:
- editor context-menu items (those contributed by monaco)
- peek, and find widgets
- commands contributed by monaco in the quick-commands (<kbd>F1</kbd>)

PR #10087, opened 2021-09-12
Localize plugins using language packs

msujew:
#### What it does

Addresses one of the open issues of #9538: Enables Theia to localize vscode extensions/plugins using language packs.

Basically addresses two problems regarding vscode extensions:

First off, it changes how the plugin package is read. Placeholders are now replaced when the frontend loads and not when the plugin is deployed. 

Secondly, through reverse engineering of the [`vscode-nls`](https://github.com/microsoft/vscode-nls/blob/main/src/node/main.ts) module, we are able to create `translationConfig` files that point to the correct localization file for each vscode extension. These files are cached in the `localization-cache` folder in the Theia home directory.

#### How to test

1. Install the `vscode.npm` builtin extension and a language pack of your liking.
2. Verify that the `npm Scripts` view is visible and in english.
3. Open a `package.json` file that contains a `scripts` section. Hover over one of the scripts and observe the `Run Script` popup: 
image
4. Use the `Configure Display Language` command to change to a different language.
5. The application reloads and the `npm Scripts` view should now have a different name. (*Tests that the `package.json` is still correctly translated*)
6. Hover over one of the scripts from the `package.json` file and observe a changed popup. (*Tests that the plugin host runtime is using the correct localization*) 
image

#### Review checklist

#### Reminder for reviewers


tsmaeder:
@msujew one of the problems I see is that we're assuming that plugins are deployed at the back end, not the plugin host. I think this prevents us from doing stuff like remote workspaces, as VS Code does. I think just using the `packageUri` field instead of `packagePath` and using the FileService instead of node file system should get us there.

tsmaeder:
Also: could you open an epic with problems you know that need solving with NLS? Would be good to be able to track the state of the implementation.

msujew:
> Could you open an epic with problems you know that need solving with NLS?

Sure, done https://github.com/eclipse-theia/theia/issues/10188

> One of the problems I see is that we're assuming that plugins are deployed at the back end, not the plugin host.

I'm not sure I can follow. Can the plugin host be deployed to other resources (which do not have access to the file system of the backend) using remote workspaces? The issue is that the extensions/plugins internally try (using the `vscode-nls` package) to read the translation files using the same file system where the language pack was installed.

Issue #10094, opened 2021-09-14
`QuickInputService.input()` does not respect `options`


safisa:
### Bug Description:
Hi,
I the quickInputService.input() method with the options prompt, placeHolder and value, all is working fine in Theia 1.16, once I updated to the latest 1.17.2, the input does not work as expected (does not show the prompt, placeHolder, and value).
Note: I did the upgrade as mentioned in the migration guide.

### Steps to Reproduce:

1. try this code: 
2. let val = await this.quickInputService.input({
            prompt: "prompt msg",
            placeHolder: "a place holder",
            value: 'initial value'
        });
3. the input will show up, but will not show the properties in the options above

### Additional Information

- Operating System: Window 10
- Theia Version: 1.17.2

vince-fugnitto:
It looks like the `options` are not being properly passed:

https://github.com/eclipse-theia/theia/blob/8df74a70d59136d15802ac6c55bdad0d9dc1b5f5/packages/monaco/src/browser/monaco-quick-input-service.ts#L184-L186

PR #10096, opened 2021-09-14
monaco: properly pass `options` to `input`

vince-fugnitto:
#### What it does

Fixes: #10094 

The pull-request addresses the issue where the `options` parameter used in `QuickInputService.input` was not being passed properly. This meant that configurations provided by callers were always disregarded.

#### How to test

1. start the application
2. use <kbd>F1</kbd> to execute `Quick Input: Test Positive Integer`
3. confirm that options such as `placholder` and `validateInput` works correctly (positive integer allowed only)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-quick-input-service.ts
+++ b/packages/monaco/src/browser/monaco-quick-input-service.ts
@@ -182,8 +182,17 @@ export class MonacoQuickInputService implements QuickInputService {
     }
 
     input(options?: InputOptions, token?: CancellationToken): Promise<string | undefined> {
-        return this.monacoService.input();
+        let inputOptions: monaco.quickInput.IInputOptions | undefined;
+        if (options) {
+            const { validateInput, ...props } = options;
+            inputOptions = { ...props };
+            if (validateInput) {
+                inputOptions.validateInput = async input => validateInput(input);
+            }
+        }
+        return this.monacoService.input(inputOptions, token);
     }
+
     pick<T extends QuickPickItem, O extends PickOptions<T>>(picks: T[] | Promise<T[]>, options?: O, token?: CancellationToken):
         Promise<(O extends { canPickMany: true; } ? T[] : T) | undefined> {
         return this.monacoService.pick(picks, options, token);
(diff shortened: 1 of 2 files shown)


packages/monaco/src/browser/monaco-quick-input-service.ts
```diff
@@ -182,8 +182,23 @@ export class MonacoQuickInputService implements QuickInputService {
...
+        let inputOptions: monaco.quickInput.IInputOptions | undefined = undefined;
+        if (options) {
+            inputOptions = {
+                ignoreFocusLost: options?.ignoreFocusLost,
+                password: options?.password,
+                placeHolder: options?.placeHolder,
+                prompt: options?.prompt,
+                value: options?.value,
+                valueSelection: options?.valueSelection
+            };
```
msujew, line 194: Why not use the spread operator on the options here like `inputOptions = { ...options }`? Is there a benefit in manually assembling the options? Is there a special reason why we don't pass the options directly?

msujew (commented):
The changes look good to me, the options are passed correct

packages/monaco/src/browser/monaco-quick-input-service.ts
vince-fugnitto: @msujew I could not use the spread since the `validateInput` typings are different for `InputOptions` and `IInputOptions`, and we'd get a compilation error.

I could however do the following, or perhaps align the typings:

```ts
input(options?: InputOptions, token?: CancellationToken): Promise<string | undefined> {
    let inputOptions: monaco.quickInput.IInputOptions | undefined = undefined;
    if (options) {
        let validateInput = undefined;
        if (options.validateInput) {
            validateInput = (input: string) => options!.validateInput!(input)!;
        }
        inputOptions = {...options, validateInput};
    }
    return this.monacoService.input(inputOptions, token);
}
```

msujew: Alright I see, I didn't notice the difference in typing previously. That's alright with me 👍

vince-fugnitto:
The changes should be ready for an additional round of review when there's time :+1: 
- updated the api-sample to handle negative numbers
- updated to use a `spread` for simplicity

msujew (approved):
I can confirm that the issue exists on `master` and is addressed by the changes nicely:

- [x] Options are correctly passed to the input
- [x] Input validation is running successfully

One could argue that `0` is an element of the set of positive integers, but I don't think we need this discussion here 😄

PR #10108, opened 2021-09-15
Add markdown renderer for plugin tree tooltips

thegecko:

#### What it does

The plugin-ext support for tree views coming from VSCode plugins doesn't support rendering markdown in tooltips.
This PR introduces functionality to render a rich tooltip when a plugin supplies markdown for that purpose.

It checks that the string passed is markdown before rendering, so shouldn't interfere with normal titles and doesn't instantiate the tooltip renderer until required.

This PR introduces a new direct dependency to [react-tooltip](https://www.npmjs.com/package/react-tooltip) (MIT), so let's see what the new `License check` GitHub CI flow says.

The tooltip has been made available to the entire application and can render html. As this has to be set programmatically, it shouldn't pose a security risk.

Tooltip follows the theming of Theia (see below)

#### How to test

Use theia with a plugin which offers known markdown tooltips, e.g. the SVD peripheral viewer of [cortex-debug](https://open-vsx.org/extension/marus25/cortex-debug)

Dark:
Screenshot 2021-09-15 at 16 25 18

Light:
Screenshot 2021-09-15 at 16 25 35

#### Review checklist

#### Reminder for reviewers


packages/core/src/browser/tooltip-renderer.tsx
```diff
@@ -0,0 +1,41 @@
...
+ ********************************************************************************/
+
+import * as React from 'react';
+import ReactTooltip from 'react-tooltip';
+import { ReactRenderer } from './widgets/react-renderer';
+import { v4 } from 'uuid';
+
+export const TOOLTIP_ID = v4();
+
+class TooltipRenderer extends ReactRenderer {
```
vince-fugnitto, line 24: Should the renderer be `injectable` so it can more easily be extended?

packages/core/src/browser/style/tooltip.css
```diff
@@ -0,0 +1,28 @@
...
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+.theia-tooltip {
+    color: var(--theia-list-hoverForeground) !important;
+    background: var(--theia-list-hoverBackground) !important;
+    border: 1px solid !important;
+    border-color: var(--theia-list-hoverForeground) !important;
```
vince-fugnitto, line 21: We already make use of the shorthand `border`, we can simplify to:

```suggestion
    border: 1px solid var(--theia-list-hoverForeground) !important;
```

vince-fugnitto (commented):
@thegecko given the fact we don't have an example where the new tooltip renderer is actually used in the codebase (meaning it is hard to test especially regressions), can we not add a example in the `api-samples`?

packages/core/src/browser/style/tooltip.css
thegecko: That approach didn't work! For some reason the variable didn't expand properly and splitting the css rules was the only way I got it to work :/

packages/core/src/browser/tooltip-renderer.tsx
thegecko: sure

thegecko:
> can we not add a example in the `api-samples`

Depends on the effort involved...

Is `api-samples` a package or repo?

vince-fugnitto:
> Is `api-samples` a package or repo?

It's an extension that is part of our repo and included in the examples (browser and electron) but not published:
- https://github.com/eclipse-theia/theia/tree/master/examples/api-samples

The idea is to at least have one example in the repo for test purposes, and given nothing at the moment uses the new tooltip renderer it is hard to test, add features to, and verify for any possible regressions.

packages/core/src/browser/style/tooltip.css
thegecko: Just tried again and it won't work :/

Screenshot 2021-09-15 at 18 18 27

packages/core/src/browser/tooltip-renderer.tsx
thegecko: done

packages/core/src/browser/tooltip-renderer.tsx
```diff
@@ -0,0 +1,43 @@
...
+ ********************************************************************************/
+
+import { injectable } from 'inversify';
+import * as React from 'react';
+import ReactTooltip from 'react-tooltip';
+import { ReactRenderer } from './widgets/react-renderer';
+import { v4 } from 'uuid';
+
+export const TOOLTIP_ID = v4();
+
```
colin-grant-work, line 24: To make it a little clearer how to use this, it might be desirable to add something like this:

```typescript
/**
 * Attributes to be added to an HTML element to enable
 * rich HTML tooltip rendering
 */
export interface TooltipAttributes {
    /**
     * HTML to render in the tooltip.
     */
    'data-tip': string;
    /**
     * The ID of the tooltip renderer. Should be TOOLTIP_ID.
     */
    'data-for': string;
}
```

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
```diff
@@ -274,13 +279,32 @@ export class TreeViewWidget extends TreeViewWelcomeWidget {
...
         const className = classes.join(' ');
-        const title = node.tooltip ||
-            (node.resourceUri && this.labelProvider.getLongName(new CoreURI(node.resourceUri)))
-            || this.toNodeName(node);
-        const attrs = this.decorateCaption(node, {
-            className, id: node.id,
-            title
-        });
+
+        let attrs: {} = {
```
colin-grant-work, line 283: If you export the `TooltipAttributes` interface I suggested above, this could be typed like this:

```typescript
let attrs: React.HTMLAttributes<HTMLElement> & Partial<TooltipAttributes> = ...
```

thegecko:
Still working on an api-example, please bear with me.

I believe all other comments/requests have been resolved.

vince-fugnitto:
A good place to use it in the future would be the following like in vscode :) 

image

thegecko:
@vince-fugnitto I spent some time trying to devise an `api-example` for this, but it got too complicated and was taking too long.

I found it far easier to simply implement the vsx-extension tooltips you suggested (see added bonus in PR description).

Would this suffice as a usage example?

thegecko:
> I’ll likely do so on Tuesday

No problem!

thegecko:
Version and preference suggestions implemented.

Issue #9987, opened 2021-08-30
Add `tags` field to `CallHierarchyItem`


tsmaeder:
### Feature Description:
Need to add the field in the API, converters and make sure it's used in the appropriate places.

PR #10114, opened 2021-09-15
call-hierarchy: add support for `tags`

Archie27376:
#### What it does

Fixes: #9987

Add support for `tags` in `CallHierarchyItem`, and includes updates to the rendering of the call-hierarchy-tree in order to display deprecated tags as deprecated (with a line-through) similarly to vscode.

https://user-images.githubusercontent.com/40359487/133489281-f9d66c14-ffee-4b31-850b-e458a16c582f.mp4

#### How to test

 - add [vscode-callhierarchy-test](https://github.com/vince-fugnitto/vscode-callhierarchy-test/releases/download/0.0.1/call-hierarchy-sample-0.0.1.vsix) to the 'plugins' folder
 - start the application
 - follow the steps in the plugin's [readme](https://github.com/vince-fugnitto/vscode-callhierarchy-test/blob/main/README.md)
    - the plugin should open a file from which you can trigger the `call-hierachy` for
    - the nodes should be marked as deprecated (for test purposes)
- confirm that the standard `call-hierarchy` works correctly (ex: typescript files)
 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/callhierarchy/src/browser/callhierarchy-tree/callhierarchy-tree-widget.tsx
+++ b/packages/callhierarchy/src/browser/callhierarchy-tree/callhierarchy-tree-widget.tsx
@@ -24,7 +24,7 @@ import { DefinitionNode, CallerNode } from './callhierarchy-tree';
 import { CallHierarchyTreeModel } from './callhierarchy-tree-model';
 import { CALLHIERARCHY_ID, Definition, Caller } from '../callhierarchy';
 import URI from '@theia/core/lib/common/uri';
-import { Location, Range, SymbolKind, DocumentUri } from '@theia/core/shared/vscode-languageserver-types';
+import { Location, Range, SymbolKind, DocumentUri, SymbolTag } from '@theia/core/shared/vscode-languageserver-types';
 import { EditorManager } from '@theia/editor/lib/browser';
 import * as React from '@theia/core/shared/react';
 
@@ -103,7 +103,13 @@ export class CallHierarchyTreeWidget extends TreeWidget {
         const symbol = definition.symbolName;
         const location = this.labelProvider.getName(new URI(definition.location.uri));
         const container = (containerName) ? containerName + ' — ' + location : location;
-        return <div className='definitionNode'>
+        const isDeprecated = definition.tags?.includes(SymbolTag.Deprecated);
+        const classNames = ['definitionNode'];
+        if (isDeprecated) {
+            classNames.push('deprecatedDefinition');
+        }
+
+        return <div className={classNames.join(' ')}>
             <div className={'symbol-icon ' + this.toIconClass(definition.symbolKind)}></div>
             <div className='definitionNode-content'>
                 <span className='symbol'>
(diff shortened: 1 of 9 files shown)


packages/callhierarchy/src/browser/callhierarchy-tree/callhierarchy-tree-widget.tsx
```diff
@@ -178,9 +190,9 @@ export class CallHierarchyTreeWidget extends TreeWidget {
     private doOpenEditor(uri: DocumentUri, range: Range, keepFocus: boolean): void {
         this.editorManager.open(
             new URI(uri), {
-                mode: keepFocus ? 'reveal' : 'activate',
-                selection: range
-            }
+            mode: keepFocus ? 'reveal' : 'activate',
+            selection: range
+        }
```
msujew, line 195: Nitpick: The formatting seems off to me. I think the opening brace belongs to the next line for TypeScript to format this correctly.

packages/plugin-ext/src/plugin/type-converters.ts
```diff
@@ -169,9 +169,9 @@ export function fromRangeOrRangeWithMessage(ranges: theia.Range[] | theia.Decora
         });
     } else {
         return ranges.map((r): DecorationOptions =>
```
msujew, line 171: Nitpick: Formatting looks off as well here

```suggestion
        return ranges.map(r => ({ range: fromRange(r) }));
```

msujew (commented):
@Archie27376 thank you for your contribution. The feature itself is working quite well, I have a few remarks on some formatting issues.

packages/callhierarchy/src/browser/callhierarchy-tree/callhierarchy-tree-widget.tsx
Archie27376: I have made the appropriate fix.

packages/plugin-ext/src/plugin/type-converters.ts
Archie27376: I have made the appropriate fix.

colin-grant-work:
Shouold we expect that this change will also affect how items are displayed in the `vscode-references-view` plugin? I tried with versions `0.0.47` (as specified in our `package.json`) and `0.0.81` (latest), and it didn't appear that items in the `vscode-references-view` tree were being decorated in particular way:

image

That could very be a separate deficiency in our `TreeView`, though.

vince-fugnitto:
> Should we expect that this change will also affect how items are displayed in the `vscode-references-view` plugin? 

@colin-grant-work it looks like it is not implemented by the `references-view` so I don't think there's anything we should do in this case. The following is the same test-case in vscode:


image

colin-grant-work:
> @colin-grant-work it looks like it is not implemented by the `references-view` so I don't think there's anything we should do in this case. 

Guess we're ahead of the curve on this feature, then :-).

colin-grant-work (approved):
This look good to me, and puts our Call Hierarchy View a step ahead of its VSCode analog.

vince-fugnitto:
I'll merge tomorrow if there are no objections :+1:

PR #10116, opened 2021-09-16
Add additional startup logging

sgraband:
#### What it does

- Add log message with timestamp of finished applicaiton loading
- Add timestamp from frontend start to sync, load and start plugin durations
- Split sync duration in waiting for backend deployment and sync
- This prevents the sync time to be longer because the backend is not yet ready

Signed-off-by: Simon Graband <sgraband@eclipsesource.com>
#### How to test

Start the frontend and inspect the log messages.
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## v1.19.0 - 10/28/2021
+
+[1.19.0 Milestone](https://github.com/eclipse-theia/theia/milestone/25)
+
+- [plugin-ext] add additional startup logging for plugin starting and application loading [#10116](https://github.com/eclipse-theia/theia/pull/10116) - Contributed on behalf of STMicroelectronics
+
 ## v1.18.0 - 9/30/2021
 
 [1.18.0 Milestone](https://github.com/eclipse-theia/theia/milestone/24)
--- a/packages/core/src/browser/frontend-application.ts
+++ b/packages/core/src/browser/frontend-application.ts
@@ -231,6 +231,7 @@ export class FrontendApplication {
             return new Promise(resolve => {
                 window.requestAnimationFrame(() => {
                     startupElem.classList.add('theia-hidden');
+                    console.log(`Finished loading frontend application after ${(performance.now() / 1000).toFixed(3)} seconds`);
                     const preloadStyle = window.getComputedStyle(startupElem);
                     const transitionDuration = parseCssTime(preloadStyle.transitionDuration, 0);
                     window.setTimeout(() => {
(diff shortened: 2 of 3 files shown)


colin-grant-work:
What do you think of the relative merits of adding the performance logs you've added in this PR as opposed to adding some sort of general timestamp to the generic logger here?

https://github.com/eclipse-theia/theia/blob/86181832c33b10e54117490b3f5a5bf9d13e142a/packages/core/src/common/logger-protocol.ts#L109-L113

sgraband:
IMHO the log messages are already relatively long and there are -depending on the number of plugins- a lot of them. Therefore, one goal was to minimize the amount of lines (and length of lines) added. I am not sure if every log message would really benefit from a timestamp. 
Maybe we should discuss this in a follow-up issue? What do you think about that?

colin-grant-work:
> IMHO the log messages are already relatively long and there are -depending on the number of plugins- a lot of them. 

Certainly can't argue with that! Perhaps we could downgrade some of the messages to `debug`-level logging, rather than `info` level. 

> Maybe we should discuss this in a follow-up issue? What do you think about that?

That sounds reasonable. Perhaps it can be part of the discussion of @tsmaeder's logging epic.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - [git] removed exports from namespace `defaultGutterStyles`, `maxWidth`, `continuationStyle`, `highlightStyle`). [#9999](https://github.com/eclipse-theia/theia/pull/9999)
 - [task] `TaskDefinition.properties.required` is now optional to align with the specification [#10015](https://github.com/eclipse-theia/theia/pull/10015)
 - [core] `setTopPanelVisibily` renamed to `setTopPanelVisibility` [#10020](https://github.com/eclipse-theia/theia/pull/10020)
+- [core] add additional startup logging for plugin starting and application loading [#10116](https://github.com/eclipse-theia/theia/pull/10116) - Contributed on behalf of STMicroelectronics
```
colin-grant-work, line 15: Since the majority of the changes are in the `plugin-ext` package, it might make sense to add that to the changelog?

colin-grant-work (approved):
The changes look reasonable, and the log messages appear as expected:

image

image

image

etc.

CHANGELOG.md
sgraband: I rebased and changed the changelog message from `[core]` to `[plugin-ext]`. Or would you expect two different entries? One for the `plugin-ext` and one for the `core` changes?

sgraband:
@colin-grant-work Can this be merged? I would then create the follow-up issue as discussed above.

Issue #5873, opened 2019-08-06
tslint rule to enforce project organization


akosyakov:
All code should be aligned with: https://github.com/theia-ide/theia/wiki/Code-Organization. It is critical since avoiding it increases amount of code which should be loaded and bundled. But it is very tedious to check import each time and new contributors very often miss it.

Maybe there is configurable tslin rule to check such things. Or we can write one based on regex.

cc @vince-fugnitto @marcdumais-work

vince-fugnitto:
I think it'd be good, I don't see however a rule that can accomplish it so it'd most likely need to be a custom rule.

akosyakov:
vscode guys seems to have such rules: https://github.com/microsoft/vscode/blob/5cad6c4879e40eb6212ea6df8d8e024958566221/tslint.json#L38-L602

Maybe we can copy and use them. Again need a CQ 🙈

vince-fugnitto:
> Maybe we can copy and use them. Again need a CQ

I guess we will need a CQ then :see_no_evil:

akosyakov:
@vince-fugnitto Do you think you can have a look into it? or you busy with something else?

vince-fugnitto:
> @vince-fugnitto Do you think you can have a look into it? or you busy with something else?

I'd like to resolve my existing task PR first then I can take a look into it for you :)

PR #10124, opened 2021-09-16
repo: add `runtime-import-check` eslint rule

vince-fugnitto:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/5873

The pull-request adds a custom **eslint** rule meant to check for invalid runtime imports which do not comply to our [code organization guidelines](https://github.com/eclipse-theia/theia/wiki/Code-Organization). The rule will help us identify uses of invalid imports which cause downstream adopters and extenders runtime issues.

#### How to test

The new rule should report errors when we use invalid imports based on different runtimes.

The pull-request includes a _baseline_ commit meant to suppress the errors currently in the codebase so we can track and fix them, while new errors from contributions should fail linting.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/eslint-plugin/index.js
+++ b/dev-packages/eslint-plugin/index.js
@@ -17,5 +17,6 @@
 
 /** @type {{[ruleId: string]: import('eslint').Rule.RuleModule}} */
 exports.rules = {
-    "shared-dependencies": require('./rules/shared-dependencies'),
+    "runtime-import-check": require('./rules/runtime-import-check'),
+    "shared-dependencies": require('./rules/shared-dependencies')
 };
--- a/dev-packages/eslint-plugin/rules/runtime-import-check.js
+++ b/dev-packages/eslint-plugin/rules/runtime-import-check.js
@@ -1,0 +1,119 @@
+// @ts-check
+/********************************************************************************
+ * Copyright (C) 2021 Ericsson and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+/* eslint-disable max-len */
+
+const path = require('path');
+
+/**
+ * Runtime-specific folders according to our coding guidelines.
+ */
+const folders = {
+    common: 'common',
+    browser: 'browser',
+    node: 'node',
+    electronCommon: 'electron-common',
+    electronBrowser: 'electron-browser',
+    electronNode: 'electron-node',
+    electronMain: 'electron-main',
+};
+
+/**
+ * @typedef {object} ImportRule
+ * @property {string[]} allowed
+ * @property {string[]} restricted
+ */
+
+/**
+ * @param {string} src
+ * @param {string[]} allowedFolders
+ * @returns {[string, ImportRule]}
+ */
+function allow(src, allowedFolders) {
+    const allowed = [src, ...allowedFolders];
+    const restricted = Object.values(folders).filter(folder => !allowed.includes(folder));
+    return [src, { allowed, restricted }];
+}
+
+/**
+ * Mapping of folders to the list of allowed/restricted folders to import from.
+ * @type {[string, ImportRule][]}
+ */
+const importRuleMapping = [
+    allow(folders.common, []),
+    allow(folders.browser, [folders.common]),
+    allow(folders.node, [folders.common]),
+    allow(folders.electronCommon, [folders.common]),
+    allow(folders.electronBrowser, [folders.electronCommon, folders.browser, folders.common]),
+    allow(folders.electronNode, [folders.electronCommon, folders.node, folders.common]),
+    allow(folders.electronMain, [folders.electronCommon, folders.node, folders.common]),
+];
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+    meta: {
+        type: 'problem',
+        fixable: 'code',
+        docs: {
+            description: 'prevent imports from folders meant for incompatible runtimes.',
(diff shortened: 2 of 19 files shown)


dev-packages/eslint-plugin/rules/runtime-import-check.js
```diff
@@ -0,0 +1,100 @@
...
+        docs: {
+            description: 'prevent imports from folders meant for incompatible runtimes.',
+            url: 'https://github.com/eclipse-theia/theia/wiki/Code-Organization'
+        },
+    },
+    create(context) {
+        const relativeFilePath = path.relative(context.getCwd(), path.resolve(context.getFilename()));
+
+        /**
+         * Check if the import is valid according to out code organization guidelines:
```
alvsan09, line 64: Just a nitpick, 
typo on 'out' -> 'our', or probably to 'the'

alvsan09 (commented):
I have tested it by removing some of the lint disabling lines and it works as expected :+1: 

The code looks good to me, with only one typo mentioned as an inline comment.

alvsan09:
@vince-fugnitto , it would be great to raise an issue and make a reference to it in this PR i.e. to track fixes for the deviations disabled on this baseline.

alvsan09 (approved):
The additional changes look good to me, 
I have only verified the functionality under Linux.

dev-packages/eslint-plugin/index.js
```diff
@@ -17,5 +17,6 @@
 
 /** @type {{[ruleId: string]: import('eslint').Rule.RuleModule}} */
 exports.rules = {
-    "shared-dependencies": require('./rules/shared-dependencies'),
+    "runtime-import-check": require('./rules/runtime-import-check'),
```
colin-grant-work, line 20: This name could be a little bit misleading, since it could sound like a check on imports made at runtime (I'm reminded of Paul's recent PR re: dynamic `require`). Maybe something like `code-partition`, `code-partition-check` or `code-partition-integrity`?

dev-packages/eslint-plugin/rules/runtime-import-check.js
```diff
@@ -0,0 +1,105 @@
...
+            ImportDeclaration(node) {
+                checkModuleImport(node.source);
+            },
+            TSExternalModuleReference(node) {
+                checkModuleImport(node.expression);
+            },
+        }
+        function checkModuleImport(node) {
+            const module = /** @type {string} */(node.value);
+            if (restrictedFolders.some(restricted => module.includes(restricted))) {
```
colin-grant-work, line 97: Would it be worth it to use a `Set` rather than an array to avoid iterating over all restricted folders for every import statement?

packages/core/src/common/quick-pick-service.ts
```diff
@@ -14,6 +14,9 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 import { Event } from './event';
+
+// TODO: fix the import.
+// eslint-disable-next-line @theia/runtime-import-check
```
colin-grant-work, line 19: Rather than adding these `TODO` annotations and then turning off the linter, might it make sense to set the level for the rule to `warning` for now, make a PR (or several) to fix these, and then switch it to `error`?

dev-packages/eslint-plugin/index.js
paul-marechal: No no, it really depends on the runtime in the sense of Node runtime? Browser runtime? Electron runtime? Runtime means execution engine/environment? WDYT?

dev-packages/eslint-plugin/rules/runtime-import-check.js
paul-marechal: We can't since we must match one of the paths (e.g. `/browser/`) to see if it is part of the path (the `module` const).

We are doing a substring search, so can't use a set here.

packages/core/src/common/quick-pick-service.ts
paul-marechal: The idea is to shoot down any new code that would mis-import code, and to do this we really need it to be an error. It also doesn't help that we already swim in a sea of warnings, so new bogus imports would get lost in the noise...

I think we can just open a followup issue inviting whoever will look at those odd imports to search for this comment.

dev-packages/eslint-plugin/index.js
colin-grant-work: Yes, I understand what sense of runtime is intended, but I think it might be better to choose a name without the ambiguity of that term.

packages/core/src/common/quick-pick-service.ts
colin-grant-work: We're also swimming in a sea of `TODO` comments, but I take your meaning :-). It depends on how quickly we _actually_ intend to do the todo's. If it's something we intend to address soon, then I don't think we need them. If it's not something we intend to address soon, then it doesn't really sound like a `TODO` at all, and anyone interested in addressing them could search for the disable comments.

dev-packages/eslint-plugin/index.js
vince-fugnitto: @colin-grant-work I'm not really found of the use of `partition` here, I think it only adds additional confusion (it's not intuitive for me, and I can imagine for downstream and newcomers it won't be either). Originally I had `code-organization`, but agreed with Paul that we are really verifying that we are importing correctly for the different runtimes we support.

Perhaps it'd be clearer like:

```
check-import-for-runtimes
```

or the more generic rule name:

```
no-invalid-import
```

colin-grant-work: Yes, I liked the `code-organization` name better :-). If there are more votes for `runtime` or some variation, I won't press too insistently. Maybe expand the message to something like "<Module> cannot be imported into <matched folder> to avoid incompatibility with target runtime." ?

vince-fugnitto:
@colin-grant-work the error message was updated to include the list of allowed folder as requested offline.

_Example_:

```
/home/evinfug/workspaces/theia/packages/core/src/electron-common/electron-main-window-service.ts
  17:34  error  '../browser/window/window-service' cannot be imported in 'electron-common', only 'electron-common, common' is allowed  @theia/runtime-import-check

/home/evinfug/workspaces/theia/packages/core/src/electron-main/electron-main-application.ts
  33:37  error  '../browser/window/window-service' cannot be imported in 'electron-main', only 'electron-main, electron-common, node, common' is allowed  @theia/runtime-import-check

/home/evinfug/workspaces/theia/packages/core/src/electron-main/electron-main-window-service-impl.ts
  23:34  error  '../browser/window/window-service' cannot be imported in 'electron-main', only 'electron-main, electron-common, node, common' is allowed  @theia/runtime-import-check
  ```

Issue #6645, opened 2019-11-27
Remove explicit dependency on terminal


thegecko:
### Description

There are many cases when the terminal functionality isn't required. It gives full access to the system being run in a web context and should be easily turned off. e.g. https://github.com/eclipse-theia/theia/issues/5663

The problem is, many extensions have a direct or indirect dependency on terminal (e.g. plugin-ext, debug, cpp), so taking any of these means the terminal functionality appears.

As discussed in the [dev meeting](https://github.com/eclipse-theia/theia/wiki/Dev-Meetings), there are a few ways this could be resolved:

- Split each extension requiring terminal into two parts; a base part and the part needing the terminal for extended functionality
- Implement a dummy implementation of terminal in `theia/core` which will satisfy all the plugins, but not register any commands or menus. Including the terminal extension would override this dummy.

I think the last option here makes most sense, but as both may require large changes, I want to get some input from others first.

@marcdumais-work @marechal-p @akosyakov @svenefftinge Thoughts?

akosyakov:
> Split each extension requiring terminal into two parts; a base part and the part needing the terminal for extended functionality

I was thinking along this lines:
- `@theia/terminal-api` extension which only defines no-op implementations with interfaces, i.e. like we do for `QuickPickService`.
- `@theia/terminal` is an implementation of `@theia/terminal-api`

Other extensions can depend only on `@theia/terminal-api`. `@theia/terminal` should be added to concrete product to get terminal actually to work if needed.  Some APIs can be marked as optional with `?`  that clients can react if API is not implemented, i.e. error reporting.

I think it is the standard way how such issues are resolved for OSGI bundles.

thegecko:
We discussed that approach in the meeting (similar to a header file and concrete implementation split), but @svenefftinge wasn't keen.

marechal-p:
>     * `@theia/terminal-api` extension which only defines no-op implementations with interfaces, i.e. like we do for `QuickPickService`.

I am not so sure about no-ops, thinking about why debug depends on the terminals. In a perfect world it would be nice to be able to not include the terminals, at all, and have the debug adapter client in Theia report that `runInTerminal` reverse requests are not supported.

If we want to keep the package count minimal, maybe we can make each Theia package contribute multiple Inversify modules? (the `theiaExtensions` array the generators look up to build apps).

Then all we lack is a way to precisely pick which modules we want to checkout.

[...]

PR #10147, opened 2021-09-22
fix toggle bottom panel to align to vscode

shuyaqian:
Signed-off-by: shuyaqian 717749594@qq.com

#### What it does

When there is no bottom panel,  we must create a terminal through the top menu or shortcut keys. However, in vscode, the terminal icon on the bottom status bar is always there. If there is no bottom panel, we can click the icon to quickly create a terminal.

#### How to test

1. click the ‘toggle bottom panel’ icon in status bar
2. if there is no bottom panel, it will create a terminal and expand

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/common-frontend-contribution.ts
+++ b/packages/core/src/browser/common-frontend-contribution.ts
@@ -762,9 +762,10 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
             }
         });
         commandRegistry.registerCommand(CommonCommands.TOGGLE_BOTTOM_PANEL, {
-            isEnabled: () => this.shell.getWidgets('bottom').length > 0,
             execute: () => {
-                if (this.shell.isExpanded('bottom')) {
+                if (this.shell.getWidgets('bottom').length === 0) {
+                    commandRegistry.executeCommand('terminal:new');
+                } else if (this.shell.isExpanded('bottom')) {
                     this.shell.collapsePanel('bottom');
                 } else {
                     this.shell.expandPanel('bottom');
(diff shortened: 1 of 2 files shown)


packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -762,9 +762,10 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
             }
         });
         commandRegistry.registerCommand(CommonCommands.TOGGLE_BOTTOM_PANEL, {
-            isEnabled: () => this.shell.getWidgets('bottom').length > 0,
             execute: () => {
-                if (this.shell.isExpanded('bottom')) {
+                if (this.shell.getWidgets('bottom').length === 0) {
+                    commandRegistry.executeCommand('terminal:new');
```
vince-fugnitto, line 767: @shuyaqian I do not believe that `@theia/core` should know anything about terminals.
It is possible that applications will also choose to omit the use of terminals (https://github.com/eclipse-theia/theia/issues/6645).

vince-fugnitto (commented):
> However, in vscode, the terminal icon on the bottom status bar is always there.

I was unable to see the bottom statusbar item to toggle the bottom panel in vscode, can you clarify?

> If there is no bottom panel, we can click the icon to quickly create a terminal.

Given we are a framework we should instead think about providing an API for adopters and extenders to contribute default views to the bottom-panel and open them if the bottom-panel does not exist rather than hardcoding that the terminal should be opened.

shuyaqian:
> the bottom panel in vscode, can you clarify?

Sorry, I found out that I installed a terminal plugin in vscode. The status bar icon is registered with this plugin. But I have to say, it's really convenient. Intellij has the icon that allow you to quickly create a terminal.

vince-fugnitto:
> Sorry, I found out that I installed a terminal plugin in vscode.

@shuyaqian if it’s the case then you can include the plugin by default in your custom application as a built-in? I don’t think there’s a need to support it directly in the framework like you’ve done. If it’s necessary you can also create a custom theia extension for your product that contributes the functionality as well.

Issue #10150, opened 2021-09-22
Theia should offer to delete files permanently when moving to Trash failed


martin-fleck-at:
I have a use case where the moving files to Trash fails as there are some restrictions on the file system. I know that there is already a preference to completely disable using the Trash but I think it might be a good idea to offer the user to permanently delete the file instead when moving files to Trash causes an error.

VS Code already does something similar [here](https://github.com/microsoft/vscode/blob/d4c222488de817d491f7c5686cf1161e9cceb7d4/src/vs/workbench/contrib/files/browser/fileActions.ts#L203).

I suggest to simply extend the `WorkspaceDeleteHandler` with this functionality. I can also provide a PR that does that.

PR #10151, opened 2021-09-22
Offer permanent deletion if Trash deletion fails

martin-fleck-at:
Fixes https://github.com/eclipse-theia/theia/issues/10150

#### What it does
If there is a failure during file deletion using the Trash option, we will offer to delete the file permanently instead.

dialog

#### How to test
Option 1 (You have a file system that does not support moving to Trash)
- Ensure that you enable the preference 'Enable Trash'
- Delete a file. This will cause an error since your file system does not support it and the new permanent deletion dialog should occur.

Option 2 (We need to throw an error ourselves)
- Ensure that you enable the preference 'Enable Trash'
- The only way I found to throw an error is to add the following snippet into the new `deleteFile` method before we call the deletion with the file service. This should make us move to the catch block and retry the delete without the useTrash option.
```
if (options.useTrash) {
   throw new Error('Error: Unknown Issue');
}
```

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/workspace/src/browser/workspace-delete-handler.ts
+++ b/packages/workspace/src/browser/workspace-delete-handler.ts
@@ -168,10 +168,10 @@ export class WorkspaceDeleteHandler implements UriCommandHandler<URI[]> {
         try {
             this.fileService.delete(uri, { ...options, useTrash: true });
         } catch (error) {
+            console.error('Error deleting with trash:', error);
             if (await this.confirmDeletePermanently(uri)) {
                 return this.deleteFilePermanently(uri, options);
             }
-            throw error;
         }
     }
 
@@ -181,7 +181,7 @@ export class WorkspaceDeleteHandler implements UriCommandHandler<URI[]> {
      * @param uri URI of selected resource.
      */
     protected async confirmDeletePermanently(uri: URI): Promise<boolean> {
-        const title = 'Move File to Trash';
+        const title = 'Error deleting file';
 
         const msg = document.createElement('div');
 
@@ -190,7 +190,7 @@ export class WorkspaceDeleteHandler implements UriCommandHandler<URI[]> {
         msg.append(question);
 
         const info = document.createElement('p');
-        info.textContent = 'You can disable the usage of Trash in the preferences.';
+        info.textContent = 'You can disable the use of Trash in the preferences.';
         msg.append(info);
 
         const response = await new ConfirmDialog({ title, msg }).open();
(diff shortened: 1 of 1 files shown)


packages/workspace/src/browser/workspace-delete-handler.ts
```diff
@@ -152,13 +152,53 @@ export class WorkspaceDeleteHandler implements UriCommandHandler<URI[]> {
...
         }
     }
 
+    /**
+     * Performs deletion of a given URI.
+     *
+     * @param uri URI of selected resource.
+     * @param options deletion options
+     */
+    protected async deleteFile(uri: URI, options: FileDeleteOptions): Promise<void> {
```
vince-fugnitto, line 168: The method has the same documentation as `delete`, perhaps elaborate on the differences between the two methods?

We can either:
- modify `delete` to perform the `catch` handling like you've done here
- rename the method to `deleteFilePermanently` and it will handle only the trash aspect.

vince-fugnitto (commented):
Thank you for your first contribution :+1: 

Do you mind updating the `how to test` so that reviewers can more easily test the changes (the section is for the author to describe test steps for reviewers).

packages/workspace/src/browser/workspace-delete-handler.ts
martin-fleck-at: You are right that having the same documentation is a bit confusing. I think adapting the documentation of the `delete` method is more reasonable as it actually does two things: close without saving and deleting the file.

I think there is still value in having a separate `closeWithoutSaving` and `deleteFile` method whereas the new permanent deletion is only relevant for the `deleteFile`. For instance, if we move the error handling into the `delete` method the error could theoretically also come from the `closeWithoutSaving` (due to Promise.all) in which case the dialog would not make any sense. But maybe I misunderstood your suggestion.

I just wanted to raise my thoughts but I am not feeling too strongly about it, so if you want I can also change the PR into one of the other two options you provided.

martin-fleck-at:
@vince-fugnitto Thank you very much for your fast review! I extended the 'How to test' with how I managed to test it even without an unsupportive file system. Unfortunately, there are code modifications necessary for this. I hope this is alright.

packages/workspace/src/browser/workspace-delete-handler.ts
vince-fugnitto: I see what you might mean, maybe we can document the method better and possibly rename it to make it clearer.

martin-fleck-at:
@colin-grant-work I rebased the change and localized the messages. If there is anything else that needs to be done, please let me know.

colin-grant-work (approved):
The latest changes look good to me. Thanks for the contribution 👍

martin-fleck-at:
@colin-grant-work Is there anything else I need to do to get this merged? :-)

Issue #10121, opened 2021-09-16
localization: move `nls` to `common`


vince-fugnitto:
#### Bug Description:

Given that `nls` is used outside of `browser`, we should move it under `common` so it respects our [code organization guidelines](https://github.com/eclipse-theia/theia/wiki/Code-Organization).

I don't believe there is any browser-specific code that prevents us to moving it under the common folder (to be more easily re-used):

https://github.com/eclipse-theia/theia/blob/master/packages/core/src/browser/nls.ts

_Invalid References_:

https://github.com/eclipse-theia/theia/blob/86181832c33b10e54117490b3f5a5bf9d13e142a/packages/monaco/src/browser/monaco-loader.ts#L19

https://github.com/eclipse-theia/theia/blob/86181832c33b10e54117490b3f5a5bf9d13e142a/packages/core/src/browser/common-frontend-contribution.ts#L57

vince-fugnitto:
@msujew any thoughts? Should we move `nls` to common, or should we localize commands perhaps later (when registered)?

msujew:
@vince-fugnitto We should probably move it to common. Although it does access some browser/window API, everything should be safe to run in common

paul-marechal:
@msujew Only `loadTranslations` is browser-specific right? Everything else can indeed be moved to `common`.

PR #10153, opened 2021-09-22
Move `nls` localization namespace to `common`

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/10121

Moves the `nls` namespace to the `common` directory. Adds a `nls-loader.ts` in the browser directory that actually loads the translation.

#### How to test

1. The disabled eslint rule `@theia/runtime-import-check` should no longer be needed
2. The application runs as expected

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 <a name="breaking_changes_1.19.0">[Breaking Changes:](#breaking_changes_1.19.0)</a>
 
+- [core] moved `nls` localization namespace from `browser` to `common`. [#10153](https://github.com/eclipse-theia/theia/pull/10153)
 - [output] moved `output-channel` from `common` to `browser` [#10154](https://github.com/eclipse-theia/theia/pull/10154)
 - [output] moved `output-preferences` from `common` to `browser` [#10154](https://github.com/eclipse-theia/theia/pull/10154)
 - [view-container] `ViewContainerPart` constructor takes new 2 parameters: `originalContainerId` and `originalContainerTitle`. The existing `viewContainerId` parameter has been renamed to `currentContainerId` to enable drag & drop views. [#9644](https://github.com/eclipse-theia/theia/pull/9644)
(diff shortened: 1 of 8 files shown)


paul-marechal:
@msujew both you and @vince-fugnitto are doing refactorings at the same time and you both are hitting the same issue with TypeScript not showing errors on your machine after moving files around. Know that this is not an issue with the build scripts rework, but an inherent problem with TypeScript build mode and refactorings...

What happened is that you had JS files built before doing your changes. When you created new files, the old ones were still in the output folder, so packages relying on those files kept working because they were still on disk as JS, even though you might have removed the original TS file.

I'll make a "smart cleanup" script that will remove output JS files without a corresponding TS source.

msujew:
@paul-marechal Ah, thanks for the explanation. I was pretty confused that TypeScript didn't bother showing errors on a build :)

vince-fugnitto (commented):

Two things that can be updated (unfortunately not well displayed on GitHub):

Update `nls.ts`:

```ts
import { Localization } from './i18n/localization';
```

and 

```ts
export const locale = typeof window === 'object' ? window.localStorage.getItem('localeId') : undefined;
```

msujew:
> `export const locale = typeof window === 'object' ? window.localStorage.getItem('localeId') : undefined;`

@vince-fugnitto This will result in `locale` being of type `string | null | undefined`, since `getItem` returns `string | null`. I kept it the current way for now.

tsmaeder:
Sorry to butt in here, but I don't believe we should move endpoint.ts to `common`. `nls.ts` consists of one function that is used everywhere and `loadTranslatations` which is browser-specific. We should split the file (we can even keep the name `nls.ts` in both places.

msujew:
> Sorry to butt in here

You're always welcome to butt in :)

Anyway, I've split the files accordingly. Let me know, if you see any other Issues with the current approach.

PR #10156, opened 2021-09-23
add ts-clean script

paul-marechal:
When doing refactorings, one might end up renaming/deleting TypeScript
source files. But if a build was run previously, the outputted
JavaScript files will still be in the build output folder. Then
extensions that depend on the refactored package will not fail to build
because the old path still resolves to the lingering JavaScript files.

Add `ts-clean` script that deletes all generated files without a
corresponding source file.

Add the missing `compile` script to most packages.

#### How to test

- Build those changes.
- Rename a TypeScript file to something new.
- Run `npx ts-clean dev-packages/* packages/* --dry`.
- The tool should prompt about a missing source file and deleting generated files.

The tool supports a `--watch (-w)` mode, use `--dry` to only get logs without touching the fs.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (commented):
I believe the CI error is simply because `@theia/timeline` incorrectly does not have a `package.spec.ts` so it fails to execute it's tests.

paul-marechal:
@vince-fugnitto CI is fixed \o/

dev-packages/ext-scripts/theia-ts-clean.js
```diff
@@ -0,0 +1,210 @@
+#!/usr/bin/env node
+
+/********************************************************************************
+ * Copyright (C) 2017 TypeFox and others.
```
msujew, line 4: Nitpick: `2021 Ericsson`?

msujew (approved):
The changes look good to me, removing/moving a file now correctly leads to typescript compilation errors due to the removed `.js` files.

dev-packages/ext-scripts/theia-ts-clean.js
paul-marechal: Good old copy/paste, good catch.

Issue #10006, opened 2021-08-31
Implement Inline and Named Pipe Debug Adapters


tsmaeder:
### Feature Description:
The type `DebugAdapterDescriptor` is missing the types `DebugAdapterNamedPipeServer` and `DebugAdapterInlineImplementation`. We'll probably have to refactor some of the debugger implementation in order to accomodate the inline debug adapter.

PR #10134, opened 2021-09-17
Add dynamic debug configurations API

alvsan09:
#### What it does

Fixes: #8652

This initial implementation exposes plugin provided dynamic debug configurations via the quick pick command palette.
e.g.
`ctrlcmd+p -> type 'debug ' (Note the space at the end).`

or alternatively: 

`ctrl+shft+p -> type 'Select and start debugging'`

If an extension provider of dynamic debug configurations is loaded, these entries are listed and available for selection.

#### How to test

#Alternative 1
1) Assuming the extension `ms-vscode.js-debug` is installed as a builtin dependency
Open a `javascript` or `typescript` project, build it  and open in editor a `.js` or `.ts` file ready to be debugged.
> ctrlcmd+p -> type 'debug ' (Note the space at the end).
2) You will notice two new entries under the `pwa-node` debug type, 
Select the one called `Run current file`
3) Verify that the debug session starts and that debug options work (e.g. stop at break points, etc.).

run_current_file

NOTE: The dynamic configuration **'Create JavaScript debug terminal',** does not work properly, this will be tracked in the following issue #9978.

#Alternative 2 (For a more in depth testing and development)

The following URL points to a fork repository of `vscode-mock-debug`, and the `theia-dynamic-debug` branch is provided for test purposes of this functionality.

* from your `theia` development repository install the `vscode-mock-debug` extension inside the `plugins` folder via the following commands:
> git clone https://github.com/alvsan09/vscode-mock-debug.git; cd vscode-mock-debug; git checkout  theia-dynamic-debug
> yarn
> yarn compile
> yarn run vscode:prepublish

* open the following file (from the extension above) in the editor: 
 > plugins/vscode-mock-debug/sampleWorkspace/readme.md

* From the command palette type 'debug ' or use the shortcut keys below:
> ctrlcmd+p -> type 'debug ' (Note the space at the end).

* select one of the provided dynamic configurations under `mock`:
image

* Example:
dynamic_mock_debug

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

Signed-off-by: Alvaro Sanchez-Leon <alvaro.sanchez-leon@ericsson.com>

PR #10163, opened 2021-09-24
Add DebugAdapterNamedPipeServer and DebugAdapterInlineImplementation

tsmaeder:
#### What it does
Adds new debug adapter descriptor types to the plugin API
Fixes https://github.com/eclipse-theia/theia/issues/10006

#### How to test
Check out the mock debug extension (https://github.com/Microsoft/vscode-mock-debug) and build it (yarn compile & yarn package) in various debug adapter mode (see extension.ts#21). Also, you need to remove some stuff that we don't support yet from activateMockDebug.ts, here's the changed file:

[activateMockDebug.ts.txt](https://github.com/eclipse-theia/theia/files/7224622/activateMockDebug.ts.txt)

The extension allows to add a launch configuration that will step through any *.md file. You just have to point the config at a concrete file instead of using the "select file" command.

#### Review checklist

#### Reminder for reviewers


packages/debug/src/node/debug-adapter-session.ts
```diff
@@ -34,25 +34,23 @@ import { DisposableCollection, Disposable } from '@theia/core/lib/common/disposa
...
     ) {
-        this.contentLength = -1;
-        this.buffer = Buffer.alloc(0);
         this.toDispose.pushAll([
             this.communicationProvider,
             Disposable.create(() => this.write(JSON.stringify({ seq: -1, type: 'request', command: 'disconnect' }))),
             Disposable.create(() => this.write(JSON.stringify({ seq: -1, type: 'request', command: 'terminate' })))
         ]);
+
+        this.communicationProvider.onMessageReceived((message: string) => this.send(message));
```
msujew, line 50: Nitpick:
```suggestion
        this.communicationProvider.onMessageReceived(message => this.send(message));
```

packages/debug/src/node/debug-model.ts
```diff
@@ -87,6 +87,11 @@ export interface DebugAdapterForkExecutable {
  */
 export type DebugAdapterExecutable = DebugAdapterSpawnExecutable | DebugAdapterForkExecutable;
 
+export interface DebugProtocolMessage extends Object {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    [properties: string]: any;
```
msujew, line 92: Is this `any` necessary to align with a vscode type? Alternatively I would change this to `unknown`.

msujew (commented):
The changes look quite good to me, I was able to successfully debug with all four different debug flags.

I had a few issues testing any flags that relied on the external `debugAdapter.js` process, since that isn't built by the webpack config. I'll include all compiled vsix (including the `debugAdapter.js)` in a single zip here for other reviewers. Be aware that after installing a different version of the extension the backend needs to be restarted, as new debug adapters will not override old ones:

[mock-debug-0.46.7.zip](https://github.com/eclipse-theia/theia/files/7235892/mock-debug-0.46.7.zip)

packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -2125,6 +2131,52 @@ export class DebugAdapterServer {
...
+    export function is(adapter: theia.DebugAdapterDescriptor | undefined): adapter is DebugAdapterNamedPipeServer {
+        return !!adapter && 'path' in adapter;
+    }
+}
+
+/**
+ * A debug adapter descriptor for an inline implementation.
+ */
+@es5ClassCompat
+export class DebugAdapterInlineImplementation implements DebugAdapterInlineImplementation {
```
alvsan09, line 2161: Shouldn't this be: 
```suggestion
export class DebugAdapterInlineImplementation {
```

alvsan09 (commented):
I have tested it using the `vscode-mock-debug` extension and it works fine :+1: 

I have just noticed a possible copy/paste issue as per inline comment.

One comment:
I ran into some additional adjustments to be able to compile and re-run the `vscode-mock-debug' extension, I had to use this extension in the PR mentioned below and created a branch to make it easier for reviewers. may be something to consider.
  Hopefully we can have a fully working branch after few more fixes in the area.
https://github.com/eclipse-theia/theia/pull/10134

packages/debug/src/node/stream-communication-provider.ts
```diff
@@ -0,0 +1,92 @@
...
+export class StreamCommunicationProvider extends DisposableCollection implements CommunicationProvider {
+    private messageReceivedEmitter = new Emitter<string>();
+    onMessageReceived: Event<string> = this.messageReceivedEmitter.event;
+    private errorEmitter = new Emitter<Error>();
+    onError: Event<Error> = this.errorEmitter.event;
+    private closeEmitter = new Emitter<void>();
+    onClose: Event<void> = this.closeEmitter.event;
+
+    private static TWO_CRLF = '\r\n\r\n';
+    private static CONTENT_LENGTH = 'Content-Length';
```
alvsan09, line 31: It would be great to add a reference to the specification that describes the header or protocol.

packages/debug/src/node/debug-model.ts
tsmaeder: It's my attempt to convince Typescript that I want an object type, not a string. Any suggestions on how to express that better would be welcome.

packages/debug/src/node/debug-adapter-session.ts
tsmaeder: I kinda like to repeat the types when they are not obvious from the context. And in this code, we're stringifying/parsing messages all the time, so it matters to know the types.

packages/debug/src/node/stream-communication-provider.ts
tsmaeder: https://microsoft.github.io/debug-adapter-protocol/overview#header-part

packages/plugin-ext/src/plugin/types-impl.ts
tsmaeder: In formal logic, you'd probably call it a tautology 🤦

packages/debug/src/node/debug-adapter-session.ts
msujew: Sure, alright with me 👍

packages/debug/src/node/debug-model.ts
msujew: I would usually use `Record<string, unknown>` or something similar for that. `any` always accepts any type, not only objects. The record should probably be what you're looking for.

packages/debug/src/node/debug-model.ts
```diff
@@ -22,12 +22,12 @@
 // Some entities copied and modified from https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts
```
thegecko, line 22: This file should have been kept in the common folder. The move has blocked browser-based adapters from working

thegecko:
This file `debug/lib/node/debug-model` should have been kept in the common folder. The move has blocked browser-based adapters from working.

tsmaeder:
@thegecko you're right, of course, it just looked like an implementation detail of the back end. Can you open an issue with what you need in the common folder so we can track it?

thegecko:
> @thegecko you're right, of course, it just looked like an implementation detail of the back end. Can you open an issue with what you need in the common folder so we can track it?

I think this work and your efforts in #10333 allows me to upstream in-browser debug adapter support more easily, so I could make these changes when doing that.

tsmaeder:
@thegecko currently, the default implementation of the `DebugService` lives in the back end. Moving that to the front end would more easily allow for contributors to implement debug adapter connections in the front or back end. It's a bit bass ackwards, right now, if you ask me.

PR #10165, opened 2021-09-24
Add possibility of DebugVariables navigation

federicobozzini:
#### What it does

In some cases it might be useful to allow navigating `DebugVariable`s, making it possible to understand things like what is their parent in the Debug panel or in which scope they are defined.

An example might be [the `theia-cpp-extension`](https://github.com/eclipse-theia/theia-cpp-extensions/blob/master/packages/cpp-debug/src/browser/cpp-debug-frontend-contribution.ts#L92) where the context of a variable is important. Another example is adding different menus on C modules if they are shown as sub-elements of a scope.

#### How to test

To test this some new code needs to be defined. It would be possible to just add a new command that as an example prints the name of the scope of a variable in `debug-frontend-contribution`:

```javascript
        registry.registerCommand(DebugCommands.PRINT_SCOPE, {
            execute: () => {
                if (this.selectedVariable) {
                    let v: ExpressionContainer = this.selectedVariable;
                    while (v instanceof DebugVariable) {
                        v = v.parent;
                    }
                    if (v instanceof DebugScope) {
                        console.log(v.name);
                    }
                }
            },
            isEnabled: () => true,
            isVisible: () => true,
        });
```

with the necessary code on top of it. If a local variable is selected and the command `PRINT_SCOPE` is executed `Local` will be printed.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/console/debug-console-items.tsx
+++ b/packages/debug/src/browser/console/debug-console-items.tsx
@@ -137,7 +137,7 @@ export class DebugVariable extends ExpressionContainer {
     constructor(
         session: DebugSessionProvider,
         protected readonly variable: DebugProtocol.Variable,
-        protected readonly parent: ExpressionContainer
+        readonly parent: ExpressionContainer
     ) {
         super({
             session,
@@ -359,7 +359,7 @@ export class DebugScope extends ExpressionContainer {
     }
 
     render(): React.ReactNode {
-        return this.raw.name;
+        return this.name;
     }
 
     get expensive(): boolean {
@@ -372,6 +372,10 @@ export class DebugScope extends ExpressionContainer {
             return new monaco.Range(line, column, endLine, endColumn);
         }
         return undefined;
+    }
+
+    get name(): string {
+        return this.raw.name;
     }
 
 }


packages/debug/src/browser/console/debug-console-items.tsx
```diff
@@ -137,7 +137,7 @@ export class DebugVariable extends ExpressionContainer {
     constructor(
         session: DebugSessionProvider,
         protected readonly variable: DebugProtocol.Variable,
-        protected readonly parent: ExpressionContainer
+        public readonly parent: ExpressionContainer
```
colin-grant-work, line 140: In general, we try to avoid writing `public`, as it's the default.

colin-grant-work (approved):
This looks good to me. Since these are effectively tree items, it makes sense that they should satisfy more of the `TreeItem` interface, including a public `parent` field.

packages/debug/src/browser/console/debug-console-items.tsx
federicobozzini: I've changed this.

PR #10175, opened 2021-09-27
fix next publishing and temporarly disable typedocs

paul-marechal:


marcdumais-work (approved):
LGTM - merging will be the real test, to confirm we are again able to  publish `next` packages!

vince-fugnitto (approved):
:champagne:

marcdumais-work:
Publish failed, because of an authentication issue. I'll try re-setting the `npm token` and restart the CI, when I have a minute.

marcdumais-work:
This might be the issue: https://github.com/lerna/lerna/issues/2788

I'll try setting a normal token instead of an automation token. If confirmed, it means that we can't enable 2 FA on the npm account.

PR #10177, opened 2021-09-27
fix publish:next

paul-marechal:
`lerna@4` follows lifecycle scripts when running publish:
https://github.com/lerna/lerna/tree/main/commands/publish#lifecycle-scripts

We never relied on those, and it also causes issues with our setup.

Add `--ignore-scripts` to prevent script recursion.


vince-fugnitto (approved):
👍

marcdumais-work:
We might also need to add an option for lerna to go ahead and publish despite the uncommitted change to the various package.json files: https://github.com/eclipse-theia/theia/pull/10175#issuecomment-928498635

marcdumais-work:
Maybe this: https://github.com/lerna/lerna/issues/1700#issuecomment-451198263

marcdumais-work (approved):
Not sure it will fix all our issues, but I noticed the extra scripts being called and it's a good step forward to get rid of them.

marcdumais-work:
I used this PR in a local registry (verdaccio) and was able to successfully publish. Let's merge and see if it also works in CI!

PR #10179, opened 2021-09-28
filter duplicate items between namedVariables and indexedVariables in…

yacosdad:
#### What it does
filter duplicate items between namedVariables and indexedVariables in same result sets, but not just in namedVariables or indexedVariable

==== BEFORE changed：
**It will show two same item in Debug Variables Widget when a breakpoint hits**
dap adapter: lldb-vscode
debuggee: ios app
variables-1

because it will fetch both namedVariables and indexedVariables for a non-zero vairableReference, and merge the results in one sets without removing duplicates.

variables-2
variables-3

== Data streams below:

f054e813 theia -> adapter: {"seq":31,"type":"request","command":"variables","arguments":{"variablesReference":1,"filter":"named"}}
f054e813 theia <- adapter: {"body":{"variables":[{"evaluateName":"self","id":0,"name":"self","type":"UIKitCatalog.ButtonViewController","value":"0x000000010280aa70","variablesReference":4}]},"command":"variables","request_seq":31,"seq":0,"success":true,"type":"response"}

f054e813 theia -> adapter: {"seq":34,"type":"request","command":"variables","arguments":{"variablesReference":1,"filter":"indexed","start":0}}
f054e813 theia <- adapter: {"body":{"variables":[{"evaluateName":"self","id":0,"name":"self","type":"UIKitCatalog.ButtonViewController","value":"0x000000010280aa70","variablesReference":4}]},"command":"variables","request_seq":34,"seq":0,"success":true,"type":"response"}

==== AFTER changed：
**This submit will keep the variable names of same result set to names set to filter duplicate item.**

variables-4

#### How to test
1. hits a breakpoint
2. lookup locals variables in debug widget
3. check the list of local variables to find duplicate items

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/console/debug-console-items.tsx
+++ b/packages/debug/src/browser/console/debug-console-items.tsx
@@ -103,6 +103,10 @@ export class ExpressionContainer implements CompositeConsoleItem {
             const response = await this.session!.sendRequest('variables', { variablesReference, filter, start, count });
             const { variables } = response.body;
             const names = new Set<string>();
+            for (const item of result) {
+                const variable = item as DebugVariable
+                names.add(variable.name)
+            }
             for (const variable of variables) {
                 if (!names.has(variable.name)) {
                     result.push(new DebugVariable(this.sessionProvider, variable, this));


packages/debug/src/browser/console/debug-console-items.tsx
```diff
@@ -103,6 +103,10 @@ export class ExpressionContainer implements CompositeConsoleItem {
             const response = await this.session!.sendRequest('variables', { variablesReference, filter, start, count });
             const { variables } = response.body;
             const names = new Set<string>();
+            for (const item of result) {
+                const variable = item as DebugVariable
+                names.add(variable.name)
```
msujew, line 108: These two statements are missing their semicolons. Please activate the ESLint extension in the Theia repo to catch these issues beforehand.

packages/debug/src/browser/console/debug-console-items.tsx
```diff
@@ -103,6 +103,10 @@ export class ExpressionContainer implements CompositeConsoleItem {
             const response = await this.session!.sendRequest('variables', { variablesReference, filter, start, count });
             const { variables } = response.body;
             const names = new Set<string>();
+            for (const item of result) {
+                const variable = item as DebugVariable
```
msujew, line 107: Directly casting from `ConsoleItem` to `DebugVariable` seems a bit unsafe to me. Since `DebugVariable` is a class, you should be able to use the `item instanceof DebugVariable` check to assert that it definitely is a `DebugVariable`. Alternatively, you could first assert that `item` contains a `name` property and then cast to something like `{ name: string }`.

msujew (commented):
Hi @yacosdad, thank you for your first contribution to Theia 👍 

The changes seem reasonable to fix the issue. However, I have no means to actually test these changes. Do you (or someone else) have alternative instructions to reproduce the issue a bit more easily?

colin-grant-work:
It's a slightly larger intervention in the code, but it would be more efficient to swap out 

```typescript
fetch(result: ConsoleItem[] /* Other stuff */)
```

for 

```typescript
fetch(result: Map<string, ConsoleItem> /* Other stuff */)
```

and make the subsidiary adjustments necessary to make that work. That means that we wouldn't need to separately set up a `Set`, because we'd get that functionality from the `Map`.

JonasHelming:
@yacosdad Are you interested in addressing the review comments?

msujew:
I will close this PR. Anyone who is interested in contributing this change can use the code changes as well as Colin's and my comments to create a new PR.

PR #10183, opened 2021-09-28
ci: increase Node max memory for docs

paul-marechal:
TypeDoc requires a decent amount of memory with our repo.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -109,6 +109,8 @@ jobs:
       - name: Pre-docs-Publish
         run: |
           yarn docs
+        env:
+          NODE_OPTIONS: --max_old_space_size=8196
 
       - name: Publish GH Pages
         uses: peaceiris/actions-gh-pages@v3


vince-fugnitto (approved):
🤞

.github/workflows/build.yml
```diff
@@ -109,6 +109,8 @@ jobs:
       - name: Pre-docs-Publish
         run: |
           yarn docs
+        env:
+          NODE_OPTIONS: --max_old_space_size=4096
```
marcdumais-work, line 113: should we merge-back together steps Pre-npm-Publish and Pre-docs-Publish? Also step max_old_space_size to 8192 (4096 failed locally for me)

paul-marechal: I was afraid of being too much, but that's only an upper bound and it will use as "little" as it can anyway. Will do.

paul-marechal: I'll keep both separated as it "isolates" the CI secrets a bit more...

marcdumais-work (commented):
LGTM

PR #10185, opened 2021-09-28
electron: define "files" in package.json

paul-marechal:
`lerna@4` seems to include almost EVERYTHING unless told otherwise when
publishing. `dev-packages/electron` is the only published package we
have that didn't define a `files` field and `lerna` tried to publish
symlinked binaries...

Add `files` field to `dev-packages/electron/package.json`.

#### How to test

Not sure how to test beside CI passing.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/electron/package.json
+++ b/dev-packages/electron/package.json
@@ -14,6 +14,14 @@
     "url": "https://github.com/eclipse-theia/theia/issues"
   },
   "homepage": "https://github.com/eclipse-theia/theia",
+  "files": [
+    "native/src",
+    "native/binding.gyp",
+    "scripts",
+    "*.js",
+    "*.d.ts",
+    "!.eslintrc.js"
+  ],
   "bin": {
     "electron": "electron-cli.js",
     "electron-codecs-test": "electron-codecs-test.js",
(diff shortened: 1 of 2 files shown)


dev-packages/electron/package.json
```diff
@@ -14,6 +14,12 @@
     "url": "https://github.com/eclipse-theia/theia/issues"
   },
   "homepage": "https://github.com/eclipse-theia/theia",
+  "files": [
+    "native",
```
marcdumais-work, line 18: The file causing issues is under this folder and still present in the resulting `.tar.gz` archive. Do we need to exclude it from here or create a `.npmignore` file?

marcdumais-work: > create a .npmignore file

I've pushed a small commit on top that does that

marcdumais-work: there were additional test commits I did not intend to push - corrected with force-push.

marcdumais-work: That did not work - `native/build` is still present in the resulting archive. My guess is that "files" has precedence (.npmignore on its own worked). Feel free to get rid of my extra commit if you have another idea that might work.

paul-marechal: Ahh I forgot that in this case it would build into `native/build`, usually it's at the package's root...

marcdumais-work (approved):
LGTM - tested locally and works!

PR #10193, opened 2021-09-29
Add closeEditor command

### LINKED ISSUE #10138: editor: add `close editor` command
#### Feature Description:

We should add support for the `close editor` command similarly to vscode.

This includes:
- declaring the command
- registering the command
- registering the menu item for the command (in the `file` main-menu)
- registering the keybinding for the command

smart-bo:
#### What it does

Fixes #10138

**editor**
The command `Close Editor` is added on the menu item (in the `File` main-menu), in order to close tabs present in the main-area.

**plugin**
Updates the `workbench.action.closeActiveEditor` mapping present in the plugin-system to use the 
already existing `CLOSE_MAIN_TAB` command. Previously, the mapping would only close actual editors
and not necessarily in the main-area which is a difference in behavior to what vscode supports. Since
vscode treats all content in the main-area as "editors" we want to do the same (since it is what plugins expect).

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/editor/src/browser/editor-menu.ts
+++ b/packages/editor/src/browser/editor-menu.ts
@@ -102,6 +102,11 @@ export class EditorMenuContribution implements MenuContribution {
             label: 'Render Whitespace',
             order: '2'
         });
+        registry.registerMenuAction(CommonMenus.FILE_CLOSE, {
+            commandId: CommonCommands.CLOSE_MAIN_TAB.id,
+            label: 'Close Editor',
+            order: '0'
+        });
     }
 
 }
--- a/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
+++ b/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
@@ -333,19 +333,7 @@ export class PluginVscodeCommandsContribution implements CommandContribution {
             execute: () => this.shell.saveAll()
         });
         commands.registerCommand({ id: 'workbench.action.closeActiveEditor' }, {
-            execute: async (uri?: monaco.Uri) => {
-                let widget = this.editorManager.currentEditor || this.shell.currentWidget;
-                if (uri) {
-                    const uriString = uri.toString();
-                    widget = this.shell.widgets.find(w => {
-                        const resourceUri = NavigatableWidget.is(w) && w.getResourceUri();
-                        return (resourceUri && resourceUri.toString()) === uriString;
-                    });
-                }
-                if (this.codeEditorWidgetUtil.is(widget)) {
-                    await this.shell.closeWidget(widget.id);
-                }
-            }
+            execute: () => commands.executeCommand(CommonCommands.CLOSE_MAIN_TAB.id)
         });
         commands.registerCommand({ id: 'workbench.action.closeOtherEditors' }, {
             execute: async (uri?: monaco.Uri) => {


packages/editor/src/browser/editor-command.ts
```diff
@@ -139,6 +139,14 @@ export namespace EditorCommands {
         category: VIEW_CATEGORY,
         label: 'Toggle Word Wrap'
     };
+    /**
+     * Command that close editor.
```
vince-fugnitto, line 143:
```suggestion
     * Command that closes the current editor.
```

packages/editor/src/browser/editor-command.ts
```diff
@@ -288,6 +299,12 @@ export class EditorCommandContribution implements CommandContribution {
         }
     }
 
+    protected closeCurrentEditor(): void {
+        const editor = this.editorManager.currentEditor;
+        if (editor) {
+            editor.close();
+        }
+    }
```
vince-fugnitto, line 307: Please add a `newline` following this method.

packages/editor/src/browser/editor-keybinding.ts
```diff
@@ -44,6 +45,12 @@ export class EditorKeybindingContribution implements KeybindingContribution {
             {
                 command: EditorCommands.REOPEN_CLOSED_EDITOR.id,
                 keybinding: this.isElectron() ? 'ctrlcmd+shift+t' : 'alt+shift+t'
+            },
+
```
vince-fugnitto, line 49: Please remove the extra `newline`.

packages/editor/src/browser/editor-keybinding-contexts.ts
```diff
@@ -36,6 +36,12 @@ export namespace EditorKeybindingContexts {
      * Unique identifier of a keybinding context that is enabled if the active editor has the focus but it does not have any overlaying widgets, such as the content assist widget.
      */
     export const strictEditorTextFocus = 'strictEditorTextFocus';
+
+    /**
+     * Unique identifier of a keybinding context that is enabled if the active editor exist.
+     */
+    export const editorExist = 'editorExist';
```
vince-fugnitto, line 43:
```suggestion
    /**
     * Unique identifier of a keybinding context that is enabled if the active editor exists.
     */
    export const editorExists = 'editorExists';
```

vince-fugnitto (commented):
I'll have to think about the actual `context` in this case, @colin-grant-work do you have any thoughts?

@smart-bo please update your commit title, there are currently typos:

> Add closeEditor command with regist and keybinding

colin-grant-work:
> I'll have to think about the actual `context` in this case, @colin-grant-work do you have any thoughts?

As I understand it, the crux of the problem is that we have a command that closes the active widget in the main area (`CommonCommands.CLOSE_MAIN_TAB`), and VSCode has a command that it labels 'Close editor'. That's a terminological difference, but it isn't really a practical difference: the VSCode command will close any widget in the main area (e.g. Getting Started or Settings), not just editors. To me that means that if we want to support VSCode's `'workbench.action.closeEditor'`, then we should map that ID it to the existing command, and put the existing command in the file menu, rather than creating a new command whose behavior actually differs further from the VSCode command's behavior.

smart-bo:
> 
> 
> > I'll have to think about the actual `context` in this case, @colin-grant-work do you have any thoughts?
> 
> As I understand it, the crux of the problem is that we have a command that closes the active widget in the main area (`CommonCommands.CLOSE_MAIN_TAB`), and VSCode has a command that it labels 'Close editor'. That's a terminological difference, but it isn't really a practical difference: the VSCode command will close any widget in the main area (e.g. Getting Started or Settings), not just editors. That means that the behavior of the 'Close editor' command is actually basically the same as our `CLOSE_MAIN_TAB` command. To me that means that if we want to support VSCode's `'workbench.action.closeEditor'`, then we should map that ID it to the existing command, and put the existing command in the file menu, rather than creating a new command whose behavior actually differs further from the VSCode command's behavior.

Do you mean `workbench.action.closeActiveEditor` in VSCode?  I saw that we already have `workbench.action.closeActiveEditor`  registed in `plugin-vscode-commands-contribution.ts` file.

colin-grant-work:
> Do you mean `workbench.action.closeActiveEditor` in VSCode? I saw that we already have `workbench.action.closeActiveEditor` registed in `plugin-vscode-commands-contribution.ts` file.

Hm... it looks like you're right about that. It looks like that implementation is wrong in the sense that, unlike VSCode, it implements a command that only targets editors. I think this is how I would go about addressing #10138.

 - Modify the code in `plugin-vscode-commands-contribution` to directly map `workbench.action.closeActiveEditor` to `CommonCommands.CLOSE_MAIN_TAB`, given that that's a more faithful implementation of VSCode's command.
 - Add `CommonCommands.CLOSE_MAIN_TAB` to the file menu.

That would satisfy both plugins and users examining the file menu.

colin-grant-work (approved):
The behavior of the new menu item matches that of the corresponding VSCode item, and the behavior of the `workbench.action.closeActiveEditor` command is also closer to VSCode than it previously was. 👍

Issue #10077, opened 2021-09-10
Menu is not shown when Electron on Linux is switched to full-screen mode


westbury:
### Steps to Reproduce:

1. Run the desktop example on Linux
2. Press F11 to switch to full-screen
3. Note that the menu is missing

### Additional Information

- Operating System: Linux
- Theia Version: 1.17

For reference the PR that introduced the full-screen feature is https://github.com/eclipse-theia/theia/pull/9399

vince-fugnitto:
@westbury I believe it is the same behavior as in vscode.
I believe the issue is that the menus are still hidden when `window.menuBarVisibility = visible` and we toggle full screen.

vince-fugnitto:
The more appropriate pull-request would be https://github.com/eclipse-theia/theia/pull/9830.

westbury:
> I believe it is the same behavior as in vscode

When using vscode, tapping the alt key shows the menu (I'm using Ubuntu 20.04).  In Theia this does not appear to work.

> The more appropriate pull-request would be #9830

Thanks for pointing out that PR, that was useful though it does not appear to affect this issue.

PR #10213, opened 2021-10-03
Ensure menu remains visible when fullscreen on Linux

westbury:
Fixes: https://github.com/eclipse-theia/theia/issues/10077

#### What it does
On Linux Electron, when `window.menuBarVisibility` is set to visible, switch to fullscreen (F11).  Previously the menu would disappear.  This PR fixes that, so the menu now remains.

#### How to test
Check that the menu now operates correctly in Electron on Linux.  Check also that the menu still works correctly on Windows and Mac.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-browser/menu/electron-menu-contribution.ts
+++ b/packages/core/src/electron-browser/menu/electron-menu-contribution.ts
@@ -207,7 +207,10 @@ export class ElectronMenuContribution implements FrontendApplicationContribution
         registry.registerCommand(ElectronCommands.TOGGLE_FULL_SCREEN, {
             isEnabled: () => currentWindow.isFullScreenable(),
             isVisible: () => currentWindow.isFullScreenable(),
-            execute: () => currentWindow.setFullScreen(!currentWindow.isFullScreen())
+            execute: () => {
+                currentWindow.setFullScreen(!currentWindow.isFullScreen());
+                currentWindow.menuBarVisible = true; // Ensure menu remains visible on Linux
+            }
         });
     }
 


packages/core/src/electron-browser/menu/electron-menu-contribution.ts
```diff
@@ -207,7 +207,11 @@ export class ElectronMenuContribution implements FrontendApplicationContribution
         registry.registerCommand(ElectronCommands.TOGGLE_FULL_SCREEN, {
             isEnabled: () => currentWindow.isFullScreenable(),
             isVisible: () => currentWindow.isFullScreenable(),
-            execute: () => currentWindow.setFullScreen(!currentWindow.isFullScreen())
+            execute: () => {
+                currentWindow.setFullScreen(!currentWindow.isFullScreen());
+                currentWindow.menuBarVisible = true; // Ensure menu remains visible on Linux
+
```
msujew, line 213: Nitpick: Unnecessary new line.

msujew (approved):
I was able to confirm that the issue exists on `master` and the changes address the issue (tested on Ubuntu and Windows):

- [x] Toggling full screen shows the menu bar correctly when it should be visible.
- [x] Toggling full screen continues to hide the menu bar correctly when it shouldn't be visible (using `menuBarVisibility` hidden or compact)

packages/core/src/electron-browser/menu/electron-menu-contribution.ts
```diff
@@ -207,7 +207,11 @@ export class ElectronMenuContribution implements FrontendApplicationContribution
         registry.registerCommand(ElectronCommands.TOGGLE_FULL_SCREEN, {
             isEnabled: () => currentWindow.isFullScreenable(),
             isVisible: () => currentWindow.isFullScreenable(),
-            execute: () => currentWindow.setFullScreen(!currentWindow.isFullScreen())
+            execute: () => {
+                currentWindow.setFullScreen(!currentWindow.isFullScreen());
+                currentWindow.menuBarVisible = true; // Ensure menu remains visible on Linux
```
vince-fugnitto, line 212: @westbury as I mentioned in the issue itself, the behavior is controllable in vscode through the `window.menuBarVisibility` preference:

- `window.menuBarVisibility: classic`: will hide the menu-bar on fullscreen
- `window.menuBarVisibility: visible`: will keep the menu-bar visible on fullscreen

You can confirm by testing with vscode.

colin-grant-work:
```suggestion
                currentWindow.menuBarVisible = this.preferenceService.get('window.menuBarVisibility', 'classic') === 'visible';
```

colin-grant-work (commented):
I think the other place that would need to be modified to match is here:

https://github.com/eclipse-theia/theia/blob/9ed915b80c587379a04b01f3e3c4486bd73d4d1a/packages/core/src/electron-browser/menu/electron-main-menu-factory.ts#L103-L119

Currently we check if a single pane is maximized (Zen mode), but we also need to check whether the whole window is maximized.

colin-grant-work:
@westbury, sorry, I lost track of this ticket. I will take another look. I've also been working on menu handling logic [here](https://github.com/eclipse-theia/theia/pull/10643), and in that context I've encountered and (I hope already or soon to have) fixed this issue.

colin-grant-work (changes requested):
@westbury, I believe that my snippet and Vince's comments go in the same direction. With the current code, if you maximize the window, the menu is visible if `window.menuBarVisibility` is set to either `visible` or `classic`. It should only be visible if `window.menuBarVisibility` is set to `visible`. In the meantime, with the implementation of the `window.titleBarStyle` preference, some additional adjustments to the code may be needed to accommodate both `native` and `custom` menu bars.

Issue #10225, opened 2021-10-05
Output/terminal linkmatcher does not match windows-style-paths resulting in no clickable links.


emilhammarstedtst:
### Bug Description:
When getting terminal output any windows-style-paths produced are not underlined when hovered or clickable due to TerminalLinkmatcherFiles-class passing string paths to Path-class that are not "valid".

not the same as issue #10224
### Steps to Reproduce:

Run any terminal command that produces a file reference in output. eg. build error message from GCC
`
c:\code\my_test_prj\main.c:6:30: error: expected ';' before '}' token
`

The above output will not be hover-able and show as an underlined and be ctrl-clickable

### Additional Information

- Operating System: Windows
- Theia Version: 1.18

PR #10226, opened 2021-10-05
Normalize and prefix windows paths with Path.separator

emilhammarstedtst:
#### What it does

Fixes: #10225 

- Prefix path with / if on the form of "c:/" into "/c:/"
- Normalize windows path-separator paths to use / instead of \\

#### How to test

Run any command that produces an output of windows-style-path in terminal and it should be clickable.

#### Review checklist

#### Reminder for reviewers

Signed-off-by: Emil HAMMARSTEDT <emil.hammarstedt@st.com>

### DIFF
--- a/packages/core/src/common/path.spec.ts
+++ b/packages/core/src/common/path.spec.ts
@@ -257,6 +257,20 @@ describe('Path', () => {
         });
     }
 
+    describe('Normalize path separator', () => {
+        it('should handle windows styled paths', async () => {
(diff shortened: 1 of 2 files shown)


vince-fugnitto (commented):
@emilhammarstedtst thank you for the contribution :+1: 
Would you mind adding additional tests to [path.spec.ts](https://github.com/eclipse-theia/theia/blob/master/packages/core/src/common/path.spec.ts) to cover the new use-cases?

Issue #10221, opened 2021-10-04
[vscode built-in extensions pack] resolved at build-time, does not fetch latest compatible extensions


marcdumais-work:
### Bug Description:
Using the `v1.53.2` version of the [vscode built-in extension pack](https://open-vsx.org/api/eclipse-theia/builtin-extension-pack/1.53.2/file/eclipse-theia.builtin-extension-pack-1.53.2.vsix) did not result in the latest compatible set of extensions being fetched. Rather they seem to remain at the previous API level: `v1.50.0`. If I instead install the pack at runtime, then extensions of the expected version are fetched: `v1.53.2`. 

I do not know if this is a generic bug we have with packs or if it's due to special processing we do for the vscode built-ins extensions.

### Steps to Reproduce:

1. Replace individual vscode built-in extensions by the `v1.53.2` pack (see pack URL above), in root `package.json` 
2. Build and run example application. Open the plugins view and confirm the version of the build-in extensions is `v1.50.0`
3. Retry, but this time remove the pack from `package.json`, rebuild, and instead install the pack from the extensions view (search for `builtin pack`)- `v1.53.2` should be proposed
4. Confirm that v1.53.2 of misc built-in vscode extensions are installed.

### Additional Information

- Operating System:
- Theia Version: 1.18.0

From the following traces, when the pack is revolved, it looks like it's only considering `v1.50.0`, and not finding some of the pack's extensions at that version:
```
Skipping: built-in extension "vscode.testing-editor-contributions" at version "1.50.0" does not exist.
Skipping: built-in extension "vscode.simple-browser" at version "1.50.0" does not exist.
```

marcdumais-work:
cc: @vince-fugnitto

marcdumais-work:
I think this may be due to the value being hardcoded here:
https://github.com/eclipse-theia/theia/blob/master/dev-packages/cli/src/download-plugins.ts#L76

```typescript
    const {
        packed = false,
        ignoreErrors = false,
        apiVersion = '1.50.0',
        apiUrl = 'https://open-vsx.org/api'
    } = options;
```

vince-fugnitto:
@marcdumais-work I think you might be right, when we bumped the api-version we had to also bump the version we download. I did not see a way to not have the two variables.

PR #10229, opened 2021-10-05
ovsx-client: update api version

vince-fugnitto:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/10221.

The commit updates the default **api-version** used when resolving extension-packs through our download script.
Since https://github.com/eclipse-theia/theia/pull/9959 bumped the default api, we had to also update the script defaults to properly respect the new default the framework supports.

#### How to test

- update the `package.json` to use the [builtin-extension-pack](https://open-vsx.org/api/eclipse-theia/builtin-extension-pack/1.50.1/file/eclipse-theia.builtin-extension-pack-1.50.1.vsi)
- delete the `plugins` folder if it exists (ex: `rm -rf plugins`)
- perform `yarn download:plugins`
- confirm that extensions are downloaded and resolved to their `1.53.2` version if available

#### Review checklist

#### Reminder for reviewers


dev-packages/cli/src/download-plugins.ts
```diff
@@ -73,7 +73,7 @@ export default async function downloadPlugins(options: DownloadPluginsOptions =
     const {
         packed = false,
         ignoreErrors = false,
-        apiVersion = '1.50.0',
+        apiVersion = '1.53.2', // should align with the framework.
```
colin-grant-work, line 76: Could you add a little more detail to this comment about which things need to agree for this specification and the framework to be aligned? And maybe add a comment at each place so that anyone who changes one location knows they need to change others.

vince-fugnitto: @colin-grant-work I updated the docs to give additional information, please let me know if there was anything else you had in mind.

colin-grant-work:
The changes look good. Is there any way we could extract the `VSCODE_DEFAULT_API_VERSION` declaration to a place where it could be used in all of these locations, rather than having to synchronize the three of them? I don't really think they share any dependencies, so it may not be possible at the moment, but it would be nice to have somewhere to put such global values.

colin-grant-work (approved):

alvsan09 (approved):
I can confirm the change works as expected and the downloaded plugins are at Version: 1.53.2 :+1:

alvsan09 (approved):
I have tested the latest commit and it works as expected :+1: 

Nice code update to be able to update the API version from a single location :+1:

colin-grant-work (approved):

Issue #10231, opened 2021-10-05
core: regression when using `iconClass`


vince-fugnitto:
### Discussed in https://github.com/eclipse-theia/theia/discussions/10223

<div type='discussions-op-text'>

<sup>Originally posted by **wassimkha** October  5, 2021</sup>
Hi,

I recently upgraded my theia from `1.14` to `1.18` and my theia extension logo broke. 

Before, to set the logo, I created my own css class:
```css
.icon-class {
  -webkit-mask: url('logo-no-text.svg');
  mask: url('logo-no-text.svg');
}
```

then I would set the `iconClass` class of that module to be equal to `"icon-class"`. 

After upgrading theia to the `1.18` version, this seems to not work anymore. Is there something new that was added recently that could have broken this?
</div>

msujew:
I'll take care of this 👍

PR #10232, opened 2021-10-05
core: Fix invisible svg icons regression

msujew:
#### What it does
Closes https://github.com/eclipse-theia/theia/issues/10231

Adds additional CSS rules that previously allowed to display svg based icons.

#### How to test

1. Open the unclosable sample view
2. Assert that the icon is correctly displayed in the sidebar
3. Move the widget to the main area
4. Assert that the icon is still correctly displayed
5. Assert that other codicons/fa icons are still correctly displayed

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/api-samples/src/browser/style/branding.css
+++ b/examples/api-samples/src/browser/style/branding.css
@@ -27,3 +27,8 @@
     background-repeat: no-repeat;
     background-size: 15%;
 }
+
+.unclosable-window-icon {
+    -webkit-mask: url('window-icon.svg');
+    mask: url('window-icon.svg');
+}
--- a/examples/api-samples/src/browser/style/window-icon.svg
+++ b/examples/api-samples/src/browser/style/window-icon.svg
@@ -1,0 +1,4 @@
+<!--Copyright (c) Microsoft Corporation. All rights reserved.-->
+<!--Copyright (C) 2019 TypeFox and others.-->
+<!--Licensed under the MIT License. See License.txt in the project root for license information.-->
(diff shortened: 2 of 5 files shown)


packages/vsx-registry/src/browser/vsx-extensions-view-container.ts
```diff
@@ -43,7 +43,7 @@ export class VSXExtensionsViewContainer extends ViewContainer {
 
         this.setTitleOptions({
             label: VSXExtensionsViewContainer.LABEL,
-            iconClass: codicon('extensions'),
+            iconClass: 'theia-vsx-extensions-icon',
```
vince-fugnitto, line 46: @msujew any reason we want to revert the icon for the `extensions-view`?

If we wanted to add an example in the codebase for the use of a non-codicon `iconClass` we can use the following sample:

https://github.com/eclipse-theia/theia/blob/1738469b045934201215c1752965d4d45f4eba1d/examples/api-samples/src/browser/view/sample-unclosable-view.tsx#L33

msujew: This is just a test commit, though using the old window icon works as well for a more permanent regression test 👍

msujew: Hm, looking through it, it uses a FontAwesome icon, which won't show a regression. Only svg based icons were affected by the regression.

vince-fugnitto: > Hm, looking through it, it uses a FontAwesome icon, which won't show a regression. Only svg based icons were affected by the regression.

I just mean we can modify the one in the sample to now use an svg icon to test for backwards compatibility, not a codicon or font-awesome one :+1:

msujew: Sure, makes sense

vince-fugnitto (approved):
I confirmed with the sample that the svg icon is correctly displayed :+1: 
I verified that:
- [x] it is correctly displayed when docked in the side-panels (left and right)
- [x] it is correctly displayed as a tab (main and bottom area) 
- [x] I did not notice any regressions with codicons

PR #10244, opened 2021-10-07
Refactor Theia to use the vscode-uri package internally

### LINKED ISSUE #9610: Use vscode URI instead of theia URI
### Feature Description:
At the moment [theia implementation of URI](https://github.com/eclipse-theia/theia/blob/55b55eec6af8b50ae61fc21bc0cd46a6c52d3272/packages/plugin-ext/src/plugin/types-impl.ts#L55) is used instead of vscode URI.

Upgrade `Monaco` version is in progress at the moment.
One of the benefit from the upgrade `Monaco` version is: using vscode URI utils functions, like https://github.com/eclipse-theia/theia/issues/8752

I see that with [the PR changes](https://github.com/eclipse-theia/theia/pull/9154) there is still `theia` URI usage instead of `vscode` URI.
Also I don't see the reason to support current theia implementation.

So, I think we should:
- revert https://github.com/eclipse-theia/theia/pull/9422, 
- test `vscode` URI usage with the new version of `Monaco`
- apply it.

If we current theia implementation is used to speed up the process of [the PR](https://github.com/eclipse-theia/theia/pull/9154) merging then my propose is - we should consider the current issue directly after merging the `Monaco` upgrade.

msujew:
#### What it does
Closes #9610.

Refactors Theia to use the `vscode-uri` package internally everywhere. Since the package does not support all class methods that were supported by the original `Uri` and `Path` implementations, appropriate functions where added to the according namespaces.

By the way, does anyone know why it takes literally 5 minutes for the code-lense test to start?

Also, how do we deal with breaking changes in this one? There are literally dozens of them across all packages.

#### How to test

1. Assert that all tests pass
2. Manual regression testing throughout the whole application

#### Review checklist

#### Reminder for reviewers


colin-grant-work:
> Also, how do we deal with breaking changes in this one? There are literally dozens of them across all packages.

I suppose there's the question on the one hand of whether it's a good idea to use the VSCode URI directly (it is) and on the other of whether it's good to _switch_ to using the VSCode URI now that we haven't been. It's hard to imagine any downstream code this wouldn't break, and while fixing the code would be fairly straightforward, but it wouldn't be completely trivial. Perhaps we need to enumerate more exactly the concrete benefits that this change would bring so that we can weight them against the costs.

colin-grant-work:
> * We can solely rely on the battle-tested implementation of vscode's uri implementation (i.e. drive/scheme detection, normalization, etc.)
> * Simplifies the plugin-ext as it doesn't have to convert between Theia's uri and vscode's uri anymore.

I'm in favor of simplicity, but I'm not sure how highly to rate these against the bother caused by removing / deprecating `Theia.URI`. Since `Theia.URI` uses `VSCode.URI` internally, we _should_ get all the benefits of VSCode's implementation - it's always there, and anything we want from it we can get via pass through. If there are things that we're doing on top of VSCode's `URI` that seem error-prone, we should look at improving or deprecating those parts of the wrapper. Eventually that might make the interface of `Theia.URI` and `VSCode.URI` converge 🙂. Regarding conversion, it's a bit of a pain, but among the conversions made by the plugin package, it's one of the simplest.

msujew:
I will close this, since I don't plan on continuing on this work for now.

Issue #9514, opened 2021-05-25
Performance of vscode.workspace.fs.readDirectory(Uri) is worse by *100 compared to vscode


idantr:
### Bug Description:
Update the description after comments by @vince-fugnitto : 
When calling to vscode.workspace.fs.readFile(Uri)  the amount of time it takes differ significally from the times of the same code in vscode:
Running the our code in vscode locally we get response times of ~15 ms 
![VSCODE Local](https://user-images.githubusercontent.com/10183179/120463379-85b1b180-c3a4-11eb-92c2-7f1d744240dc.JPG)

When running the code in Theia via electron locally  n ~597ms range on the same file
![Theia Electron Local](https://user-images.githubusercontent.com/10183179/120463770-e93bdf00-c3a4-11eb-8842-9db4bb9a9419.JPG)

** also we saw that in vscode that consecutive calls are even faster

### Steps to Reproduce:

1. Create a very simple vscode plugin using the following code:
```
import * as vscode from 'vscode';
import * as _ from 'lodash';
import { promises as fs } from 'fs';
import * as path from 'path';
 
export async function activate(context: vscode.ExtensionContext) {
    context.subscriptions.push(vscode.commands.registerCommand('idan-test.readDirs', async () => {
        const baseUri: vscode.Uri = _.get(vscode.workspace.workspaceFolders, "[0].uri");
        const plUri: vscode.Uri = vscode.Uri.file(path.join(baseUri.fsPath, "package-lock.json"));
 
        const before = Date.now();
        const contentVscode: Uint8Array = await vscode.workspace.fs.readFile(plUri);
        const after = Date.now();
        vscode.window.showInformationMessage(`Result VSCODE - ${after - before}`);
        
        const beforeNode = Date.now();
        const contentNode = await fs.readFile(path.resolve(plUri.fsPath));
        const afterNode = Date.now();
        vscode.window.showInformationMessage(`Result NODE - ${afterNode - beforeNode}`);
    }));
}
```
2. in you `projects` folder insert the following file
[package-lock.json.txt](https://github.com/eclipse-theia/theia/files/6583849/package-lock.json.txt)
(remove the ".txt" in the end)
package-lock.json  486k

### Additional Information

- Operating System:  WIN10
- vscode 1.56.0
- theia latest master code revision 583220b2ecae38bf16f89403417c68e4c191d1dc

vince-fugnitto:
[...]

PR #10265, opened 2021-10-12
Update JSON-RPC

### LINKED ISSUE #10264: Update JSON-RPC and Related Dependencies
### Feature Description:

We use `vscode-ws-jsonrpc` a [wrapper around `vscode-jsonrpc`](https://github.com/TypeFox/vscode-ws-jsonrpc) developed by TypeFox that TypeFox has stopped supporting. That package currently only supports `vscode-jsonrpc` < 6 as `vscode-jsonrpc@6` introduced significant changes, and that affects the versions of the language server protocol that we can use. A [fork](https://github.com/CodinGame/monaco-jsonrpc) has been made of the `vscode-ws-jsonrpc` that has updated to use `vscode-jsonrpc` >= 6, and we should consider updating to use that package instead or dump the wrapper and use `vscode-jsonrpc` directly.

colin-grant-work:
#### What it does

Fixes #10264 by switching to use a fork of `vscode-ws-jsonrpc` that has been updated to support `vscode-jsonrpc@6`. 

I could use some help, though, as I'm not very familiar with all of the details of the JSON RPC system. I'll make some comments pointing to the places where I'm uncertain of the right move.

#### How to test

1. Since this modifies the core of our communication with the backend, thoroughly test... everything, I guess.
2. In particular, check that interactions with plugins work as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -62,10 +62,8 @@
     "route-parser": "^0.0.5",
     "safer-buffer": "^2.1.2",
     "uuid": "^8.3.2",
-    "vscode-languageserver-protocol": "~3.15.3",
-    "vscode-languageserver-types": "^3.15.1",
+    "vscode-languageserver-protocol": "^3.16.0",
     "vscode-uri": "^2.1.1",
-    "vscode-ws-jsonrpc": "^0.2.0",
     "ws": "^7.1.2",
     "yargs": "^15.3.1"
   },
@@ -101,9 +99,7 @@
       "react-dom",
       "react-virtualized",
       "vscode-languageserver-protocol",
-      "vscode-languageserver-types",
-      "vscode-uri",
-      "vscode-ws-jsonrpc"
+      "vscode-uri"
     ],
     "export =": [
       "dompurify as DOMPurify",
--- a/packages/core/src/common/messaging/proxy-factory.ts
+++ b/packages/core/src/common/messaging/proxy-factory.ts
@@ -16,7 +16,7 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { MessageConnection, ResponseError } from 'vscode-ws-jsonrpc';
+import { MessageConnection, ResponseError, ParameterStructures, CancellationToken } from 'vscode-languageserver-protocol';
 import { ApplicationError } from '../application-error';
 import { Event, Emitter } from '../event';
 import { Disposable } from '../disposable';
@@ -114,14 +114,8 @@ export class JsonRpcProxyFactory<T extends object> implements ProxyHandler<T> {
     }
 
     protected waitForConnection(): void {
-        this.connectionPromise = new Promise(resolve =>
-            this.connectionPromiseResolve = resolve
-        );
-        this.connectionPromise.then(connection => {
-            connection.onClose(() =>
-                this.onDidCloseConnectionEmitter.fire(undefined)
-            );
-            this.onDidOpenConnectionEmitter.fire(undefined);
+        this.connectionPromise = new Promise(resolve => {
+            this.connectionPromiseResolve = resolve;
         });
     }
 
@@ -132,10 +126,33 @@ export class JsonRpcProxyFactory<T extends object> implements ProxyHandler<T> {
      * response.
      */
     listen(connection: MessageConnection): void {
-        connection.onRequest((prop, ...args) => this.onRequest(prop, ...args));
-        connection.onNotification((prop, ...args) => this.onNotification(prop, ...args));
+        // When a method is called without arguments then `params` is `undefined`.
+        // `token` seems to always be set by `vscode-jsonrpc` on the receiving side.
+        connection.onRequest((method: string, params: object | unknown[] | undefined, token: CancellationToken) => {
+            if (params === undefined) {
+                return this.onRequest(method, token);
+            }
+            if (!Array.isArray(params)) {
+                throw new Error('Unexpected JSON-RPC params format');
+            }
+            return this.onRequest(method, ...params, token);
+        });
+        connection.onNotification((method: string, params: object | unknown[] | undefined) => {
+            if (params === undefined) {
+                return this.onNotification(method);
+            }
+            if (!Array.isArray(params)) {
+                throw new Error('Unexpected JSON-RPC params format');
+            }
+            return this.onNotification(method, ...params);
+        });
         connection.onDispose(() => this.waitForConnection());
+        connection.onClose(() => {
+            this.waitForConnection();
+            this.onDidCloseConnectionEmitter.fire();
+        });
         connection.listen();
+        this.onDidOpenConnectionEmitter.fire();
         this.connectionPromiseResolve(connection);
     }
 
@@ -151,10 +168,10 @@ export class JsonRpcProxyFactory<T extends object> implements ProxyHandler<T> {
      *
      * @returns A promise of the method call completion.
      */
-    protected async onRequest(method: string, ...args: any[]): Promise<any> {
+    protected async onRequest(method: string, ...params: any[]): Promise<any> {
         try {
             if (this.target) {
-                return await this.target[method](...args);
+                return await this.target[method](...params);
             } else {
                 throw new Error(`no target was set to handle ${method}`);
             }
@@ -176,9 +193,9 @@ export class JsonRpcProxyFactory<T extends object> implements ProxyHandler<T> {
      * Same as [[onRequest]], but called on incoming notifications rather than
      * methods calls.
      */
-    protected onNotification(method: string, ...args: any[]): void {
+    protected onNotification(method: string, ...params: any[]): void {
         if (this.target) {
-            this.target[method](...args);
+            this.target[method](...params);
         }
     }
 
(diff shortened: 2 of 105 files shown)


tsmaeder:
T.b.h, I would tend toward in-sourcing this code: I believe the newest version of vscode-jsonrpc actually _has_ support for sockets/ipc, etc. We would be better served maintaining the bit of glue code needed ourselves.

paul-marechal:
@msujew good catch, this is now fixed. There is still an error being thrown when stopping the session but it already happens on master so that makes it out of scope for this PR.

colin-grant-work:
@tsmaeder, it seems like there's some overlap between the work here and the work you've been doing on the debug system. Would you mind taking a look at this so we can merge if appropriate, and you can then use some of the new messaging equipment in your PR?

packages/plugin-ext/src/common/connection.ts
```diff
@@ -79,3 +87,10 @@ export class PluginWebSocketChannel implements IWebSocket {
         this.connection.dispose();
     }
 }
+
+/**
+ * Use `PluginChannel` instead.
+ *
+ * @deprecated since 1.19.0
+ */
+export const PluginWebSocketChannel = PluginChannel;
```
tsmaeder, line 96: I'm not sure we need a deprecation cycle here: it's not "official" API and I doubt someone actually used this implementation detai.

packages/plugin-ext/src/common/connection.ts
```diff
@@ -52,23 +55,28 @@ export class PluginConnection implements Connection {
 }
 
 /**
- * [IWebSocket](#IWebSocket) implementation over RPC.
+ * Wrapper around a [PluginConnection](#PluginConnection) to match the [Channel](#Channel) interface.
  */
-export class PluginWebSocketChannel implements IWebSocket {
-    constructor(protected readonly connection: PluginConnection) { }
+export class PluginChannel implements Channel<string> {
```
tsmaeder, line 60: We're wrapping a string channel around a json rcp connection which wraps a string channel. That does not really make sense, IMO. I got rid of that all in https://github.com/eclipse-theia/theia/pull/10333. So I wouldn't sweat the small stuff here.

packages/plugin-ext/src/common/plugin-message-writer.ts
```diff
@@ -56,10 +56,15 @@ export class PluginMessageWriter extends AbstractMessageWriter implements Messag
...
+        const content = JSON.stringify(message);
+        try {
+            this.proxy.$sendMessage(this.id, content);
+            return Promise.resolve();
+        } catch (e) {
+            return Promise.reject(e);
+        }
     }
+
+    end(): void { }
```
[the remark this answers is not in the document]
tsmaeder: @colin-grant-work like what?

packages/core/src/node/messaging/messaging-contribution.ts
```diff
@@ -75,15 +74,16 @@ export class MessagingContribution implements BackendApplicationContribution, Me
...
     listen(spec: string, callback: (params: MessagingService.PathParams, connection: MessageConnection) => void): void {
         this.wsChannel(spec, (params, channel) => {
-            const connection = createWebSocketConnection(channel, new ConsoleLogger());
+            const connection = Channel.createMessageConnection(channel, new ConsoleLogger());
             callback(params, connection);
         });
     }
 
-    forward(spec: string, callback: (params: MessagingService.PathParams, connection: IConnection) => void): void {
+    forward(spec: string, callback: (params: MessagingService.PathParams, connection: TheiaMessageConnection) => void): void {
```
tsmaeder, line 82: Is this even ever used? I didn't find any usages.

tsmaeder (dismissed):
I really wish we could have done the changing of the imports to 'vscode-languageserver-protocol'  in a separate PR: 88 files is very big when "real changes" are insterspersed with just changed imports.

The only real change I'm requesting is the renaming of `TheiaMessagingConnection`: I just find the name very confusing. There are many things in this area of code that I disapprove of, but they were there before so 🤷 .

In the long run, I would like to get rid of jsonrpc anyway: we already have a remote proxy invocation framework in our code and I don't think we need two: the only thing I can see missing is the ability to read partial messages, but that could quite simply be retrofitted.

packages/plugin-ext/src/common/plugin-message-writer.ts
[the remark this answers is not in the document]
colin-grant-work: If you don't know, and @paul-marechal doesn't know, then I don't know. The method is required for the interface, but it doesn't seem we have anything we want to do in this case.

colin-grant-work:
> @colin-grant-work do we need more testing on this PR?

@tsmaeder, more testing never hurts. It seems that @msujew was happy with his last run through, and I haven't noticed any problems, but you may be aware of use cases that we're not hitting.

packages/core/package.json
```diff
@@ -61,10 +61,9 @@
     "route-parser": "^0.0.5",
     "safer-buffer": "^2.1.2",
     "uuid": "^8.3.2",
-    "vscode-languageserver-protocol": "~3.15.3",
-    "vscode-languageserver-types": "^3.15.1",
+    "vscode-languageserver-protocol": "^3.16.0",
```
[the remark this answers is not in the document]
paul-marechal: It's a struggle.

packages/plugin-ext/src/common/connection.ts
paul-marechal: I did what I must to get it to compile as before, I'd be more than happy to merging #10333 after this!

packages/core/src/node/messaging/messaging-contribution.ts
paul-marechal: Not anymore but despite the amount of changes in this PR I tried to only touch what was broken/bad.

This method is not bothering anyone so I'd keep it for now :)

packages/plugin-ext/src/common/connection.ts
paul-marechal: I'm very fine with that, will remove.

packages/core/src/common/messaging/proxy-factory.ts
```diff
@@ -212,46 +237,51 @@ export class JsonRpcProxyFactory<T extends object> implements ProxyHandler<T> {
...
+            if (CancellationToken.is(args[args.length - 1])) {
+                token = args.pop();
+            }
+            // Convert the list of arguments into a structure that allows us to conserve undefined values.
+            const params: any[] = this.createJsonRpcParameters(args);
+            // Push back the token if there was one.
+            // `vscode-jsonrpc` will remove the `CancellationToken` out of `params` when is is the last element.
+            if (token) {
+                params.push(token);
+            }
```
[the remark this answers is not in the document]
paul-marechal: Like I mentioned, `undefined` values are transformed into `null` without the type system knowing.

https://github.com/eclipse-theia/theia/blob/8bf15dc1d9b186b11c1f954b4c51500e611d3072/packages/plugin-ext/src/main/common/plugin-paths-protocol.ts#L24-L25

Here `workspaceUri` is actually `string | null` when used over RPC. We got "lucky" here as the only implementation checks for truthiness to detect the type rather than some `workspaceUri === undefined` which would not work.

One solution would be for interfaces meant to be used over RPC to almost never use `undefined` in their API.

packages/core/src/common/messaging/theia-message-connection.ts
```diff
@@ -0,0 +1,57 @@
...
+import { MessageReader, MessageWriter, Message } from 'vscode-languageserver-protocol';
+import { Disposable, DisposableCollection } from '../../common';
+
+/**
+ * Abstraction that can be used to route JSON-RPC messages around.
+ *
+ * This is close to `vscode_jsonrpc.MessageConnection` but without all the JSON-RPC
+ * specific methods such as `sendRequest`, `sendNotification`, etc.
+ */
+export interface TheiaMessageConnection extends Disposable {
```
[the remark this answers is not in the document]
colin-grant-work: Since its job is to forward messages, how about `RPCMessageRelay`?

tsmaeder:
Btw: does the new version help with https://github.com/eclipse-theia/theia/issues/9514?

colin-grant-work:
> Btw: does the new version help with #9514?

@tsmaeder, I haven't checked, but it would surprise me if it had much effect on the performance - I don't think there was that much extraneous code running before, and I don't think there's much less now. We still have to JSON-ify everything and send it over the wire (short or long) to the other end. There's a little bit of argument processing on our side and in the package, and that certainly adds a little overhead, but I don't think that's changed much. Someone please bring in a performance tracer and correct me :-).

packages/core/src/common/messaging/theia-message-connection.ts
```diff
@@ -0,0 +1,57 @@
...
+import { MessageReader, MessageWriter, Message } from 'vscode-languageserver-protocol';
+import { Disposable, DisposableCollection } from '../../common';
+
+/**
+ * Abstraction that can be used to route JSON-RPC messages around.
+ *
+ * This is close to `vscode_jsonrpc.MessageConnection` but without all the JSON-RPC
+ * specific methods such as `sendRequest`, `sendNotification`, etc.
+ */
+export interface RpcMessageRelay extends Disposable {
```
tsmaeder, line 26: I guess the file name does not make sense now.

msujew:
@colin-grant-work @paul-marechal I assume a rebase is in order due to #10333. I would retest this afterwards and approve if possible. We'd have a few weeks after merging to find possible regressions before the next release.

colin-grant-work:
I'm going to close this for now, since it looks like there's more interest in rewriting our message layer than in updating this package.

PR #10268, opened 2021-10-13
Fix #4783: Support flatten/compress folders in the explorer

### LINKED ISSUE #4783: [explorer] flatten/compress folders in the explorer tree when possible
One enhancement to the current `Explorer` is to be able to flatten the tree when possible (whenever a parent directory contains a single child directory).

Such an improvement will allow for numerous benefits and improve the display of large workspaces. One such example is a large Java project with long package names.

In essence, the following tree:

```
+ com
  + domain
    + package
      + task 
        Task.java
...
```

can be optimized to the following:

```
+ com/domain/package/task
  Task.java
...
```
Here is the `VSCode` issue highlighting the change: https://github.com/Microsoft/vscode/issues/41627

**Benefits:**
1. Improves navigation of large projects (easier to navigate to the correct folder)
2. Improves display (easier to parse the tree visually)
3. Improves use of vertical space (more content can be displayed on the screen at once)
4. Removes useless information and manual intervention (takes longer to individually click each node that has no children in order to get to desired content)
5. Potentially speed up the display of the tree (no need to add unnecessary nodes with icons, labels, decorators)

**Limitations:**
1. Find a mechanism or way to add content in the middle, for example `domain` in the following package name `com.domain.package.task`
2. Users may not want to display a flattened `Explorer`, therefore a preference may be suitable, or simply flatten with the use of a `Toggle`.

**References:**
- [vscode flatten](https://github.com/Microsoft/vscode/issues/41627)

EstherPerelman:

#### What it does

Fix #4783

#### How to test

**A.**

 1. Go to preferences
 2. Search for `compact folders`  
 3. Toggle the setting
 4. See that folders are compressed/uncompressed accordingly...

**B.** 

- Add/Remove folders 
- Right click on compressed element
- Search any compressed folder (stand on a folder in the tree and type the letters to search)

compressible sample — Theia Browser Example - Google Chrome 2021-10-13 14-52-08

#### Review checklist

#### Reminder for reviewers


EstherPerelman:
> I haven't looked at the code yet, but I was quickly able to produce a few bugs when testing this manually:
Thanks for your reports,
> 1. Folders in empty workspace are not visible. Reproducable by opening an empty workspace and adding folders. This leads to even more unintended behavior when adding files into the mix:
 2021-10-13.15-05-28.mp4 

Fixed (I treated it before but later missed it during a refactor)
> When having a stack of folders with a depth > 2, closing the folders and opening it by clicking on the first folder (a in the example), does not actually expand the tree-node in the navigator:
 2021-10-13.15-07-43.mp4 
There also seems to be an issue related to react keys:

I couldn't reproduce it (tested after the fixes, can you confirm?)
> root ERROR Warning: Each child in a list should have a unique "key" prop. See https://fb.me/react-warning-keys for more information.

Fixed

EstherPerelman:
@msujew thanks for your fixes
in vscode the highlight not so clear for me look at this:
image
its taking the first letter from the first and skipping to the next caption, not so clear what's going here...

packages/core/src/browser/tree/tree-compression.ts
```diff
@@ -0,0 +1,122 @@
...
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { injectable } from 'inversify';
+import { ExpandableTreeNode } from './tree-expansion';
+import { TreeNode, CompositeTreeNode } from './tree';
+
+export class CompressionStack {
```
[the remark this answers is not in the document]
EstherPerelman: @colin-grant-work it looks much better, I'm going to try it... thank you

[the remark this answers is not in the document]
EstherPerelman: > ```ts
>  export function isCompressionParent(node?: TreeNode): node is CompositeTreeNode {
>         return CompositeTreeNode.is(node) && node.children.length === 1 && CompositeTreeNode.is(node.children[0]);
>     }
> ```

@colin-grant-work where will be the check for `DirNode`? only on the `file tree`? (a single file shouldn't be compressed...)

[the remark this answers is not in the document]
EstherPerelman: @colin-grant-work 
where will be all the conditions from here:
` protected shouldCompressNode(node: TreeNode): boolean {
        if (!CompositeTreeNode.is(node)) {
            return false;
        }
        const root = this.getRootNode(node);
        // The tree root is always `WorkspaceNode`, its child is the main project folder - shouldn't be compressed,
        // its grandchildren are the direct children of the main project folder - shouldn't be compressed either.
        if (node.parent && (node.parent.id === root.id || (node.parent.parent && node.parent.parent.id === root.id))) {
            return false;
        }
        return DirNode.is(node) && this.isSingleChild(node);
    }

    protected isSingleChild(node: CompositeTreeNode): boolean {
        return !!node.parent && node.parent.children.length === 1;
    }`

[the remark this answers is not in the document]
EstherPerelman: @colin-grant-work one more thing is problematic - each place that uses for example `isCompressionChild` function - need to ensure that compression is enabled, otherwise - the actions will be executed also when unselect compact settings...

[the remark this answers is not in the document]
colin-grant-work: > where will be the check for DirNode? only on the file tree? (a single file shouldn't be compressed...)

One option would be to put these functions not in a namespace, but on the `Tree` or `TreeWidget` itself. Then it could be overridden by the `FileTree`. But would the check for `CompositeTreeNode.is()` not cover the distinction between directories and files?

---

> where will be all the conditions from here:

In the `shouldCompressNode` function, only the `CompositeTreeNode.is(node)` refers to something that wouldn't be available with my suggested code, and that could be replaced with something like `(this.isCompressionEnabled() && CompositeTreeNode.isCompressionParticipant(node)`

---

> one more thing is problematic - each place that uses for example isCompressionChild function - need to ensure that compression is enabled, otherwise - the actions will be executed also when unselect compact settings... As I see it - the node must have the compressed property in order to indicate the enablement of the compression... I'm waiting forward to see you answer...

There are a couple of possibilities here. If the functions are left in the namespace, then they could all accept a second argument `applyCompression` (as `getCompressedItems` does in my suggestion), and only return true if that's true. Alternatively, if the functions are moved into the `Tree` system, then they can easily check `this.isCompressionEnabled`. The trickiest part is the change to the `TreeIterator` system. There, I think the most straightforward approach would be to add a third option to `TreeIterator.Options`, something like this:

```typescript
export interface Options {
        /**
         * Whether to ignore siblings of the starting node.
         */
        readonly pruneSiblings: boolean;
        /**
         * Whether to ignore nodes whose parents are not expanded.
         */
        readonly pruneCollapsed: boolean;
        /**
         * Whether to include composite nodes that are the only children of their parents
         * even if the parent is collapsed.
         * Ignored of `pruneCollapsed` is `false`.
         */
        readonly applyCompression: boolean;
    }
```

The general question is whether we want `CompressibleTreeNode` to be a new entity, or whether we want compression to be a new type of tree behavior that depends on other state / flags. I think it's predictable enough to be a form of behavior, but if it turns out that we need more entities to track the state accurately.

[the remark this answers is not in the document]
EstherPerelman: @colin-grant-work  thank you for the clear answers, 
> The trickiest part is the change to the `TreeIterator` system

Beside the `TreeIterator` - the `TreeExpansionServiceImpl` also uses the compression... 
I'm able to put all the compression functions in the `CompressibleTree` and check in `TreeExpansionServiceImpl`: ``if('isCompressionParticipiant' in this.tree){return this.tree as any.isCompressionParticipiant}`` ...

Should I do this or leave all and stay with `CompressibleTreeNode` (without the `compressibleTreeService`), Pay attention that it might be that in the future will be more places that need to use it (e.g in `selectionTree` creating `TreeIterator` without the compression childs because the `applyCompression` info won't be available to him)

EstherPerelman:
@colin-grant-work meantime (waiting for your response in the conversation above) I removed `TreeCompressionService` and rearranged the compression function to be like your suggestion...
can you merge it or reply to the conversation?

colin-grant-work (commented):
This looks good. Dragging and dropping appears to be fixed, and the approach to tracking compression is leaner. There was only one minor issue I encountered:
 - When navigating the file tree using they keyboard, pressing up or down when on a multi-node (compressed) row would jump to the next multi-node (compressed) row. If none was available in the relevant direction, then the selection was stuck in that row, even if there were many non-compressed nodes available to select.

EstherPerelman:
> This looks good. Dragging and dropping appears to be fixed, and the approach to tracking compression is leaner. There was only one minor issue I encountered:
> 
> * When navigating the file tree using they keyboard, pressing up or down when on a multi-node (compressed) row would jump to the next multi-node (compressed) row. If none was available in the relevant direction, then the selection was stuck in that row, even if there were many non-compressed nodes available to select.

@colin-grant-work thanks, I fixed it, Can you merge it?

colin-grant-work:
@EstherPerelman, please don't ask whether something can be merged. It will be merged if it is reviewed and approved, so please just request a review or that someone take another look (I will).

colin-grant-work (commented):
With the latest changes, up and down are working well, but now, if a folder is selected and you use the right arrow button, the selection moves down, rather than expanding the node - there appears to be no way to expand nodes using the keyboard.

EstherPerelman:
> With the latest changes, up and down are working well, but now, if a folder is selected and you use the right arrow button, the selection moves down, rather than expanding the node - there appears to be no way to expand nodes using the keyboard.

@colin-grant-work fixed.

colin-grant-work (approved):
This is working well for me:
- Keyboard navigation
  - [x] up-down moves around the rows as expected
  - [x] left-right moves between collapsed nodes and expands / collapses as expected
- Mouse actions
  - [x] Clicking a node expands / collapses as expected
  - [x] Drag actions highlight and target the correct node
  - [x] Context-menu actions target the correct node
  - [x] Usual tree selection and actions still working correctly

I suggest we merge this shortly after the release on Thursday.

EstherPerelman:
@colin-grant-work, thanks for everything,
It is very important to us at SAP that it will be for the next version because it might be the latest version for us (we had a lot of changes upgrading our product to version 1.18 and prefer to maintain stability)
Is there any chance you consider this?

colin-grant-work:
> @colin-grant-work, thanks for everything, It is very important to us at SAP that it will be for the next version because it might be the latest version for us (we had a lot of changes upgrading our product to version 1.18 and prefer to maintain stability) Is there any chance you consider this?

@EstherPerelman, it's a pretty big change to merge just a few days before the release. I *believe* everything is working well, but let me ask @msujew and @vince-fugnitto to have a look, and depending on how it looks in their testing, it may make it in.

packages/filesystem/src/browser/compressible-tree/compressible-tree.ts
```diff
@@ -0,0 +1,40 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { injectable } from '@theia/core/shared/inversify';
+import { TreeNode, CompressibleTreeNode, TreeImpl } from '@theia/core/lib/browser';
+
+@injectable()
+export abstract class CompressibleTree extends TreeImpl {
+
+    protected abstract isCompressionEnabled: () => boolean;
```
vince-fugnitto, line 23: Why do we require a callback instead of just:

```ts
protected abstract isCompressionEnabled(): boolean;
```

packages/filesystem/src/browser/compressible-tree/compressible-tree-container.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    child.unbind(TreeModelImpl);
+    child.bind(CompressibleTreeModel).toSelf();
+    child.rebind(TreeModel).toService(CompressibleTreeModel);
+
+    child.unbind(TreeWidget);
+    child.bind(CompressibleTreeWidget).toSelf();
+
+    return child;
+}
```
vince-fugnitto, line 38: I'm not sure why `compression` exists in `filesystem`, should it not be in `@theia/core`?

packages/filesystem/src/browser/compressible-tree/compressible-tree-widget.tsx
```diff
@@ -0,0 +1,301 @@
...
+                if (changedNode.expanded && !item.expanded) {
+                    return !!this.model.expandNode(item);
+                } else if (!changedNode.expanded && item.expanded) {
+                    return !!this.model.collapseNode(item);
+                }
+            }
+            return false;
+        };
+        // Searching for a compressed child that is not updated with the `node` expansion state and update it,
+        // The update will cause the `onExpansionChanged` event to fire with the updated child and search again for a compressed child that need to be upadted,
```
vince-fugnitto, line 53: typo:

```suggestion
        // The update will cause the `onExpansionChanged` event to fire with the updated child and search again for a compressed child that need to be updated,
```

packages/core/src/browser/tree/tree-widget.tsx
```diff
@@ -1088,12 +1104,16 @@ export class TreeWidget extends ReactWidget implements StatefulWidget {
         this.addEventListener(this.node, 'focus', () => this.doFocus());
     }
 
+    protected isMutiSelectMaskEvent(event: KeyboardEvent): boolean {
```
vince-fugnitto, line 1107: typo:

```suggestion
    protected isMultiSelectMaskEvent(event: KeyboardEvent): boolean {
```

packages/filesystem/src/browser/compressible-tree/compressible-tree-widget.tsx
```diff
@@ -0,0 +1,301 @@
...
+        const compressedNode: React.ReactNode[] = [];
+        if (compressedItem) {
+            compressedNode.push(this.getSeparatorElement(index));
+        }
+        compressedNode.push(this.getCompressibleElement(node, compressedItem, index, children));
+        return compressedNode;
+    }
+
+    protected getSeparatorElement(index: number): React.ReactNode {
+        return <span className={COMPRESSED_CAPTION_SEPARATOR_CLASS} key={`seperator-${index}`}>/</span>;
```
vince-fugnitto, line 282: typo:

```suggestion
        return <span className={COMPRESSED_CAPTION_SEPARATOR_CLASS} key={`separator-${index}`}>/</span>;
```

packages/filesystem/src/browser/compressible-tree/compressible-tree-widget.tsx
```diff
@@ -0,0 +1,301 @@
...
+        // In case user clicked on the area outside the label of the tree raw, the context menu event need to be handled for the selected compressed item.
+        if (!isCompressionPartEvent && CompressibleTreeNode.isCompressionHead(node)) {
+            const items = this.getCompressedItems(node);
+            selectedNode = items.find(item => SelectableTreeNode.isSelected(item));
+        }
+        super.handleContextMenuEvent(selectedNode || node, event);
+    }
+
+    protected needsActiveIndentGuideline(node: TreeNode): boolean {
+        // Commpressed node has no indent for itself.
```
vince-fugnitto, line 191: typo:

```suggestion
        // Compressed node has no indent for itself.
```

packages/filesystem/src/browser/compressible-tree/compressible-tree-widget.tsx
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    protected renderCaptionNode(node: TreeNode): React.ReactNode[] {
+        const captionNode: React.ReactNode[] = [];
+        captionNode.push(
+            this.toCompressibleHighlightNode(0, node)
+        );
+        if (CompressibleTreeNode.isCompressionHead(node)) {
+            const comppressedItems = this.getCompressedItems(node);
+            comppressedItems.forEach((item, idx) => {
```
vince-fugnitto, line 239: typos:

```suggestion
            const compressedItems = this.getCompressedItems(node);
            compressedItems.forEach((item, idx) => {
```

packages/navigator/src/browser/navigator-preferences.ts
```diff
@@ -24,12 +24,19 @@ export const FileNavigatorConfigSchema: PreferenceSchema = {
             type: 'boolean',
             description: 'Selects file under editing in the explorer.',
             default: true
+        },
+        'explorer.compactFolders': {
+            type: 'boolean',
+            description: `Controls whether the explorer should render folders in a compact form. In such a form, single child folders will be compressed in a combined tree element.
```
vince-fugnitto, line 30: Please localize the following preference description:

```suggestion
            // eslint-disable-next-line max-len
            description: nls.localizeByDefault('Controls whether the explorer should render folders in a compact form. In such a form, single child folders will be compressed in a combined tree element. Useful for Java package structures, for example.'),
```

packages/core/src/browser/tree/tree-compression.ts
```diff
@@ -0,0 +1,79 @@
...
+        return isCompressionParent(node) && !isCompressionChild(node);
+    }
+
+    export function isCompressionTail(node?: TreeNode): node is CompressibleTreeNode {
+        return isCompressionChild(node) && !isCompressionParent(node);
+    }
+
+    export function isCompressionParticipant(node?: TreeNode): node is CompositeTreeNode {
+        return isCompressionParent(node) || isCompressionChild(node);
+    }
```
vince-fugnitto, line 58: Can we document the terminology here? For example what is a compression `head` and `tail`. How are they different than children. We should probably document each of these functions.

packages/filesystem/src/browser/compressible-tree/compressible-tree.ts
vince-fugnitto: No strong opinion about having the callback instead of a field, just wanted to know as to why we went for this approach, no need to update :)

packages/core/src/browser/tree/tree-compression.ts
EstherPerelman: I added some comments but I'm not sure it's clear enough, I found it little hard for me to find the right wording, I'll be happy with your suggestions

packages/filesystem/src/browser/compressible-tree/compressible-tree.ts
EstherPerelman: I had no particular intention, prefered to update...

EstherPerelman:
Hi all, Who can tell me what the tests are failing on?

colin-grant-work:
> Hi all, Who can tell me what the tests are failing on?

@EstherPerelman, it looks like lint errors, at the moment. Try running `yarn lint --fix` from the repo root.

packages/core/src/browser/compressible-tree/tree-compression.ts
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Represents a tree node that is able to be compressed into
+ * its parent node depending on its `compressed` state.
+ */
+export interface CompressibleTreeNode extends ExpandableTreeNode {
+    /**
+     * Indicates whether the tree node can be compressed into its parent node
+     */
+    compressible: boolean;
+}
```
colin-grant-work, line 29: As far as I can tell, this interface is no longer really used. There's code in `compressible-tree` that manipulates this field, but no code that checks it elsewhere. I think the interface can be removed, and the code below that returns `node is CompressibleTreeNode` or the like can be replaced with `node is CompositeTreeNode`.

packages/core/src/browser/compressible-tree/tree-compression.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export namespace CompressibleTreeNode {
+    export function is(node: Object | undefined): node is CompressibleTreeNode {
+        return !!node && CompositeTreeNode.is(node) && 'compressible' in node;
+    }
```
colin-grant-work, line 34: This can be removed.

packages/core/src/browser/compressible-tree/compressible-tree.ts
```diff
@@ -0,0 +1,41 @@
...
+                const uncompressedParent = CompressibleTreeNode.getUncompressedParent(node);
+                if (uncompressedParent) {
+                    node.compressible = true;
+                }
+            } else {
+                node.compressible = false;
+            }
+        }
+        super.addNode(node);
+    }
```
colin-grant-work, line 40: If the `CompressibleTreeNode` interface is removed, this method can be removed.

packages/core/src/browser/compressible-tree/tree-compression.ts
EstherPerelman: @colin-grant-work I use it here:
 ``` `/** Determines whether the node is a compressed child */
    export function isCompressionChild(node: Object | undefined): node is CompressibleTreeNode {
        return is(node) && node.compressible && isCompressibleParent(node.parent);
    }` ```
([A link to the code](https://github.com/EstherPerelman/theia/blob/f44787127aacbfea20510f9097093f4d4b935253/packages/core/src/browser/compressible-tree/tree-compression.ts#L46-L49))
This what does all the work of the compressible node...

EstherPerelman:
@colin-grant-work I answered you [here](https://github.com/eclipse-theia/theia/pull/10268#discussion_r756643587)

packages/core/src/browser/compressible-tree/tree-compression.ts
colin-grant-work: As I've [said before](https://github.com/eclipse-theia/theia/pull/10268#discussion_r738710469), everything you need to know about whether a node is compressible can be determined from its features and those of its parent; you don't need the `compressible` field. A node is a compressed child (i.e. a non-initial node in a compressed row) under the following conditions:
 - its parent is a composite tree node
 - its parent has a single child (the node itself)
 - the node itself is a composite tree node

colin-grant-work:
@EstherPerelman, are you still interested in working on this feature?

EstherPerelman:
> @EstherPerelman, are you still interested in working on this feature?

Actually we are not upgrading our theia version any more and I'm busy with other tasks, 
I don't care you finish it (it will be grate), I'll be here if needed
thanks

Issue #10164, opened 2021-09-24
debug: clean up debug session management


tsmaeder:
### Bug Description:
Whe only remove debug sessions when the debuggee sends an "exited" event. However, when we press the "stop" button in the UI, we're sending a 'terminate' request to the debuggee...this just asks the debuggee to terminate the debug session not to exit the process. So we end up with a stale debug session we cannot get rid of.

### Steps to Reproduce:

1. Install the mock debug extension (probably attaching to a long-running java process works, as well) (see https://github.com/eclipse-theia/theia/pull/10163 for instructions)
2. Launch a moch debug sessions
3. Press the "stop" button on the debug session
4. Observe. the session remains.

I believe we should get rid of the debug session upon 'terminated', not 'exited'.

### Additional Information

- Operating System:
- Theia Version: 1.18.0 master

svor:
Can see this problem when running Theia with **redhat.java-0.82.0.vsix** and **vscode-java-debug-0.35.0.vsix**

https://user-images.githubusercontent.com/1271546/135986824-09694a0b-1eb6-40e2-9dad-47aa71433f69.mp4

alvsan09:
> @alvsan09 are you going to work on this, because I was going to.

I started looking into it but I have been a bit distracted from it, 
It's a good time to take it over if you like :+1: 
Thanks for reaching out !

tsmaeder:
I'm currently working on a fix, but t.b.h it's not obvious how to track the lifecycle of nested debug sessions, debug adapters and debuggees. Still seeing lots of surprises.

PR #10333, opened 2021-10-27
Clean up debug session lifecycle

tsmaeder:
#### What it does

Reworks the lifecycle of debug session related objects to be more reliable for edge cases. Fixes https://github.com/eclipse-theia/theia/issues/10164

The fix consists of multiple changes:

1. Do not rely on fake events to signal debug session lifecycle events ("Exited"), etc. Instead, rely on closing comms channels.
2. The owner of all debug session-related objects is the `DebugAdapterSession` object. Once that object goes away, all related objects in the back-end and front-end go away.
3. Simplified the "PluginConnection" service:
  * shared implementation for main/ext side
  * simplify structure of connection objects

#### How to test

I've tested these changes by using the newest VS Code Java/java debug extensions, the jsdebug typescript debugger and the mock debug extension from https://github.com/eclipse-theia/theia/pull/10163

It would be interesting to extend the testing to other debuggers I'm less familiar with, in particular C++ or python, for example. The relevant scenarios are stop and restart actions, as well as killing debuggees or debug adapters.

#### Review checklist

#### Reminder for reviewers


alvsan09:
I have tested it using the `vscode-mock-debug` extension using all available run modes and the sessions are now terminated :+1: .
However, using the chrome developer tools (Memory) to take snapshots of heap memory before and after a debug session we can detect debug session objects that are not garbage collected i.e.
- DebugSessionConnection
- DebugSource

These objects increase in proportion to the number of executions. e.g. 
the following image show the number of object instances remaining after the first run:
image

The next image shows the number of object instances after 5 additional runs:
image

tsmaeder:
Great catch, @alvsan09. Does the tool show which references retain the leaked objects?

alvsan09:
> Great catch, @alvsan09. Does the tool show which references retain the leaked objects?

The tool itself does not provide you with the facility to move up to the containing references (as far as I can see), 
however the references in the leaked objects may help us do that some times e.g 
in the picture below we can see that debug source has a reference to the debug session pointing to an instance of  PluginDebugSession, so that help us for this case.

image

alvsan09:
> > Great catch, @alvsan09. Does the tool show which references retain the leaked objects?
> 
I was wrong about the facilities of the tool, look at the button of the page inside Memory "Retainers", 
you can then see the whole tree :-)

image

tsmaeder:
It looks like the debug console session never disposes the completion provider it registers with monaco. It looks like this is is not a new bug, but should be trivial to fix.

tsmaeder:
@msujew while cleaning up sessions, I'm getting stuck at console-session.ts#118 ff. 
```
            // Don't delete the last session by overriding it with 'undefined'
            if (session) {
```

but we _have to_ let go of the session once it is disposed.

colin-grant-work:
> @msujew while cleaning up sessions, I'm getting stuck at console-session.ts#118 ff.
> 
> ```
>             // Don't delete the last session by overriding it with 'undefined'
>             if (session) {
> ```
> 
> but we _have to_ let go of the session once it is disposed.

`console-widget.ts`, I think.

msujew:
@tsmaeder I see, that makes sense. I'm not totally against removing the last session if it closes. I implemented it this way, since that's how it seemed to work in vscode. When ending the last active debug session, the console would still display the latest session's text. However, when removing the last debug session console, Theia loses the text as well.

I'm alright with you removing the last session as well, I'm sure we'll find another way to persist the last session's text in the debug console somehow.

packages/debug/src/node/stream-debug-adapter.ts
```diff
@@ -16,10 +16,13 @@
...
 import { Emitter, Event } from '@theia/core/lib/common/event';
+import { Deferred } from '@theia/core/lib/common/promise-util';
+import { ChildProcess } from 'child_process';
 import * as stream from 'stream';
-import { CommunicationProvider } from './debug-model';
+import * as net from 'net';
+import { DebugAdapter } from './debug-model';
 
-export class StreamCommunicationProvider extends DisposableCollection implements CommunicationProvider {
+abstract class StreamDebugAdapter extends DisposableCollection {
```
thegecko, line 25: It would be nice to see this adapter pulled into a common area with the node specific implementations below existing in the `node` folder. This would help abstract this from any browser implementations ;)

packages/plugin-ext/src/common/connection.ts
```diff
@@ -13,69 +13,126 @@
...
-import { PluginMessageReader } from './plugin-message-reader';
-import { PluginMessageWriter } from './plugin-message-writer';
-import { IWebSocket, MessageReader, MessageWriter, Message } from '@theia/core/shared/vscode-ws-jsonrpc';
+import { Channel } from '@theia/debug/lib/common/debug-service';
+import { ConnectionExt, ConnectionMain } from './plugin-api-rpc';
+import { Emitter } from '@theia/core/lib/common/event';
 
 /**
- * The interface for describing the connection between plugins and main side.
+ * [IWebSocket](#IWebSocket) implementation over RPC.
```
thegecko, line 21: nit: IWebSocket isn't used any more here

packages/debug/src/common/debug-service.ts
```diff
@@ -116,3 +116,16 @@ export namespace DebugError {
         data: { type }
     }));
 }
+
```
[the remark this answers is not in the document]
thegecko: Is it simply to avoid the `IDisposable` interface?

thegecko (changes requested):
Still pending testing with our debug adapter, but have added a few minor comments.

thegecko:
@tsmaeder where possible, please ensure common code (which doesn't use node features) is kept in the common folder. See https://github.com/eclipse-theia/theia/pull/10163#issuecomment-958656286

packages/debug/src/browser/debug-session.tsx
```diff
@@ -63,9 +64,10 @@ export class DebugSession implements CompositeTreeElement {
     }
 
     protected readonly childSessions = new Map<string, DebugSession>();
-
     protected readonly toDispose = new DisposableCollection();
 
+    private isStopping: boolean = false;
```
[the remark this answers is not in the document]
tsmaeder: Good question...need to think about that.

packages/debug/src/browser/debug-session.tsx
```diff
@@ -308,59 +312,44 @@ export class DebugSession implements CompositeTreeElement {
...
-        if (!this.terminated && this.capabilities.supportsTerminateRequest && this.configuration.request === 'launch') {
-            this.terminated = true;
-            await this.connection.sendRequest('terminate', { restart });
-            if (!await this.exited(1000)) {
-                await this.disconnect(restart);
-            }
-        } else {
-            await this.disconnect(restart);
-        }
+    canTerminate(): boolean {
```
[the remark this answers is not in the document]
tsmaeder: Why would you want it to be a property?

packages/debug/src/browser/debug-session.tsx
```diff
@@ -308,59 +312,44 @@ export class DebugSession implements CompositeTreeElement {
...
-                listener.dispose();
-                resolve(false);
-            }, timeout);
-        });
+    async stop(isRestart: boolean, callback: () => void): Promise<void> {
+        if (!this.isStopping) {
+            this.isStopping = true;
+            if (this.canTerminate()) {
+                const terminated = this.waitFor('terminated', 1000);
+                await this.connection.sendRequest('terminate', { restart: isRestart });
```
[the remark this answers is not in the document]
tsmaeder: The "restart" is just a flag indicating to the debug adapter that the request is part of a restart sequence.

packages/debug/src/common/debug-service.ts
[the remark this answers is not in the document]
tsmaeder: Well...first, depending vscode-ws-jsonrpc seems iffy: debugger connections are not related to jsonrpc, but use their own wire format.
Second, yes, I think we're very much overusing the "dispose()" concept: it suggests a single lifecycle when objects may be retained by different referents for different amounts of time. I think "disconnect" and "destroy" would be more helpful abstractions.
Third: there is an additional "close" method.

packages/debug/src/node/debug-adapter-session.ts
```diff
@@ -92,10 +85,18 @@ export class DebugAdapterSessionImpl implements DebugAdapterSession {
     }
 
     protected write(message: string): void {
-        this.communicationProvider.send(message);
+        if (!this.isClosed) {
```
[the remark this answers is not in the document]
tsmaeder: Since the debug adapter might be stupid things at any time, we need to guard against the unforseen 🤷 .

packages/debug/src/node/stream-debug-adapter.ts
tsmaeder: I agree this implementation might be interesting to reuse for browser-side debug adapters (if they exist). However, I see that as a different PR, not part of "cleaning up the lifecycle".

packages/debug/src/browser/debug-session.tsx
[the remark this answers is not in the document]
thegecko: No strong reason, just neater as `if (this.canTerminate) ...`

packages/debug/src/common/debug-service.ts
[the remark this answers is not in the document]
thegecko: Ack, missed the `close` (but that could be an extended interface)

packages/debug/src/node/stream-debug-adapter.ts
thegecko: We have a browser debug adapter for Keil Studio today, currently reworking it to test your PR which is where these comments are coming from.

packages/debug/src/node/debug-model.ts
```diff
@@ -88,18 +89,35 @@ export interface DebugAdapterForkExecutable {
...
- * procedure, however it is also not disallowed.
- *
- * TODO: the better name is DebugStreamConnection + handling on error and close
+ * Implementers stand for the various types of debug adapters the system can talk to.
+ * Creation of debug adapters is not covered in this interface, but handling communication
+ * and the end of life is.
  */
-export interface CommunicationProvider extends Disposable {
+
+export interface DebugAdapter {
```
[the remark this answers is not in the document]
tsmaeder: I can see that the interfaces are very similar. However, I don't think we need to strive for minimalism where interfaces are concerned: with duck typing, object having the same signature are still type-compatible.

So the question we should ask is this: when we change the `Channel` interface, do we want the `DebugAdatper` interface, as well? Or in other words, _IS A_ debug adapter a special kind of channel? I don't feel that is the intention, in this case.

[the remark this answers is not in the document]
tsmaeder: For "IWebSocket", that is a firm no: debug adapters typically don't use websockets, but just regular TCP sockets.

tsmaeder:
> @tsmaeder I see, that makes sense. I'm not totally against removing the last session if it closes. I implemented it this way, since that's how it seemed to work in vscode. When ending the last active debug session, the console would still display the latest session's text. However, when removing the last debug session console, Theia loses the text as well.
> 

As far as I can tell, VS code keeps debug sessions after they are closed. However, I don't think the debug console session would react properly if we tried to evaluate an expression on a closed debug session, for example. 

> I'm alright with you removing the last session as well, I'm sure we'll find another way to persist the last session's text in the debug console somehow later (in a separate PR).

In general, what's the lifecycle of console sessions? Do they stick around after the process they are bound to terminates? I think the relationship between the console session and the counterpart that it gets output from is similar.

packages/debug/src/browser/debug-session.tsx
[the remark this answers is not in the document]
tsmaeder: I don't think they can: "isStopping" basically means that the session is going away: when we do a restart without creating a new session, we just send a "restart" message to the debug adapter and we should not hit this code path.

packages/debug/src/browser/debug-session.tsx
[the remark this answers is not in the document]
tsmaeder: Personally, I'm not a big fan of casual use of property getters and setters for two reasons:

1. It doesn't play well with JSON serialisation
2. It may hide arbitrary complexity behind something that looks simple like `foo.bar = true`. 

For those reasons writing property getters and setters need justification, IMO, for example to ensure internal consistency of an object or implementing some "magic" that can't be achieved otherwise. Not having to type `()` is not sufficient reason, IMO, but then, I don't share the industry's obsession with typing less ;-)

packages/debug/src/browser/debug-session-connection.ts
```diff
@@ -78,7 +78,7 @@ export interface DebugEventTypes {
     'breakpoint': DebugProtocol.BreakpointEvent
     'capabilities': DebugProtocol.CapabilitiesEvent
     'continued': DebugProtocol.ContinuedEvent
-    'exited': DebugExitEvent
+    'exited': DebugProtocol.ExitedEvent,
```
thegecko, line 81: Why was this event changed? Using the DebugProtocol event loses the `reason` field exposed by `DebugExitEvent`.

If this change is needed, I believe the `DebugExitEvent` interface above is no longer used.

tsmaeder: Because there already is an "ExitedEvent" in the debug adapter protocol that is sent when the debug adapter learns that the debuggee exited. 
The `DebugExitEvent` is a fake event that was sent to signal that something went wrong with the connection to the debug adapter (not the debuggee). The decision to remove fake (or "generated" events) was made in order to make it simpler to understand the debug session lifecycle to make it more robust. If we need to communicate this error to the debug session, we need to establish a different channel, IMO.

tsmaeder: You're right that it's no longer used.

packages/debug/src/browser/debug-session-manager.ts
```diff
@@ -278,51 +278,69 @@ export class DebugSessionManager {
...
         });
-        session.on('exited', () => this.destroy(session.id));
-        session.start().then(() => this.onDidStartDebugSessionEmitter.fire(session));
+
+        session.on('exited', async event => {
+            await session.disconnect(false, () => this.debug.terminateDebugSession(session.id));
+        });
+
+        session.onDispose(() => this.cleanup(session));
+        session.start().then(() => this.onDidStartDebugSessionEmitter.fire(session)).catch(e => session.stop(false, () => this.debug.terminateDebugSession(session.id)));
```
msujew, line 293: Catching the start error and trying to terminate it afterwards will lead to unexpected behavior. `session.stop` awaits a `sendRequest` call just before executing the callback. When the session errors on startup, the `sendRequest` call will never complete, and we are left with a dangling debug session. You test this by installing the current version of `cortex-debug` from OVSX and running this debug config:

```json
{
  "cwd": "${workspaceRoot}",
  "name": "Cortex-Debug",
  "request": "launch",
  "type": "cortex-debug",
  "executable": "",
  "servertype": "openocd",
  "serverpath": "",
  "armToolchainPath": "",
  "configFiles": [
    ""
  ]
}
```

`session.stop` Is called and will stop it's execution at the following line, which never completes:
https://github.com/eclipse-theia/theia/blob/bbba07d45ffa63b7c4edb767f66f2487d6b7d0d2/packages/debug/src/browser/debug-session.tsx#L341

packages/debug/src/browser/debug-session-connection.ts
thegecko: OK, if this is being removed, please be aware it breaks an established API (we rely on detecting errors through this event).

packages/debug/src/node/stream-debug-adapter.ts
thegecko: > I see that as a different PR, not part of "cleaning up the lifecycle".
Ack

thegecko (approved):
Successfully tested with our debug adapter and happy for this to be merged when you are ready.
Thanks for the changes, should make a future PR for upstreaming browser debug support easier.

packages/debug/src/browser/debug-session-manager.ts
tsmaeder: I would suspect the problem is that we don't know the state of the connection to the debug adapter if the session start fails. My approach would be to make sending messages fail when the connection to the debugger is closed.

tsmaeder: Hmh...debugging through the scenario with cortex-debug, the creation of the session does _not_ fail. However, sending the "disconnect" request never resolves nor rejects.

tsmaeder: Here's the sequenced I get when runng the debug config: 

```
06b9e88e theia -> adapter: {"seq":1,"type":"request","command":"initialize","arguments":{"clientID":"Theia","clientName":"Theia IDE","adapterID":"cortex-debug","locale":"en-US","linesStartAt1":true,"columnsStartAt1":true,"pathFormat":"path","supportsVariableType":false,"supportsVariablePaging":false,"supportsRunInTerminalRequest":true}}
06b9e88e theia -> adapter: {"seq":2,"type":"request","command":"get-arguments"}
06b9e88e theia <- adapter: {"seq":1,"type":"response","request_seq":1,"command":"initialize","success":true,"body":{"supportsHitConditionalBreakpoints":true,"supportsConfigurationDoneRequest":true,"supportsConditionalBreakpoints":true,"supportsFunctionBreakpoints":true,"supportsEvaluateForHovers":true,"supportsSetVariable":true,"supportsRestartRequest":true,"supportsGotoTargetsRequest":true,"supportSuspendDebuggee":true,"supportTerminateDebuggee":true}}
06b9e88e theia -> adapter: {"seq":3,"type":"request","command":"launch","arguments":{"cwd":"c:\\Users\\thomas\\code\\chedev\\theia","name":"Cortex-Debug","request":"launch","type":"cortex-debug","executable":"","servertype":"openocd","serverpath":null,"armToolchainPath":"","configFiles":[""],"noDebug":false,"gdbServerConsolePort":55878,"debuggerArgs":[],"demangle":true,"swoConfig":{"enabled":false,"decoders":[],"cpuFrequency":0,"swoFrequency":0,"source":"probe"},"rttConfig":{"enabled":false,"decoders":[]},"graphConfig":[],"preLaunchCommands":[],"postLaunchCommands":[],"preAttachCommands":[],"postAttachCommands":[],"preRestartCommands":[],"postRestartCommands":[],"searchDir":[],"toolchainPath":null,"toolchainPrefix":"arm-none-eabi","gdbPath":null,"extensionPath":"c:/Users/thomas/.theia/extensions/marus25.cortex-debug-0.4.7/extension","flattenAnonymous":false,"registerUseNaturalFormat":true,"variableUseNaturalFormat":true}}
06b9e88e theia <- adapter: {"seq":2,"type":"response","request_seq":2,"command":"get-arguments","success":false,"message":"TypeError: Cannot read property 'customRequest' of undefined\n    at q.customRequest (c:\\Users\\thomas\\.theia\\extensions\\marus25.cortex-debug-0.4.7\\extension\\dist\\debugadapter.js:9:31666)\n    at q.dispatchRequest (c:\\Users\\thomas\\.theia\\extensions\\marus25.cortex-debug-0.4.7\\extension\\dist\\debugadapter.js:1:18100)\n    at q.handleMessage (c:\\Users\\thomas\\.theia\\extensions\\marus25.cortex-debug-0.4.7\\extension\\dist\\debugadapter.js:9:61573)\n    at q._handleData (c:\\Users\\thomas\\.theia\\extensions\\marus25.cortex-debug-0.4.7\\extension\\dist\\debugadapter.js:9:63297)\n    at Socket.<anonymous> (c:\\Users\\thomas\\.theia\\extensions\\marus25.cortex-debug-0.4.7\\extension\\dist\\debugadapter.js:9:61907)\n    at Socket.emit (events.js:315:20)\n    at addChunk (_stream_readable.js:295:12)\n    at readableAddChunk (_stream_readable.js:271:9)\n    at Socket.Readable.push (_stream_readable.js:212:10)\n    at Pipe.onStreamRead (internal/stream_base_commons.js:186:23)","body":{"error":{"id":1104,"format":"{_stack}","variables":{"_exception":"Cannot read property 'customRequest' of undefined","_stack":"TypeError: Cannot read property 'customRequest' of undefined\n    at q.customRequest (c:\\Users\\thomas\\.theia\\extensions\\marus25.cortex-debug-0.4.7\\extension\\dist\\debugadapter.js:9:31666)\n    at q.dispatchRequest (c:\\Users\\thomas\\.theia\\extensions\\marus25.cortex-debug-0.4.7\\extension\\dist\\debugadapter.js:1:18100)\n    at q.handleMessage (c:\\Users\\thomas\\.theia\\extensions\\marus25.cortex-debug-0.4.7\\extension\\dist\\debugadapter.js:9:61573)\n    at q._handleData (c:\\Users\\thomas\\.theia\\extensions\\marus25.cortex-debug-0.4.7\\extension\\dist\\debugadapter.js:9:63297)\n    at Socket.<anonymous> (c:\\Users\\thomas\\.theia\\extensions\\marus25.cortex-debug-0.4.7\\extension\\dist\\debugadapter.js:9:61907)\n    at Socket.emit (events.js:315:20)\n    at addChunk (_stream_readable.js:295:12)\n    at readableAddChunk (_stream_readable.js:271:9)\n    at Socket.Readable.push (_stream_readable.js:212:10)\n    at Pipe.onStreamRead (internal/stream_base_commons.js:186:23)"},"sendTelemetry":true}}}
06b9e88e theia <- adapter: {"seq":3,"type":"response","request_seq":3,"command":"launch","success":false,"message":"Unable to find executable file at .","body":{"error":{"id":103,"format":"Unable to find executable file at .","showUser":true}}}
06b9e88e theia -> adapter: {"seq":4,"type":"request","command":"disconnect","arguments":{"restart":false}}
```

To me, this looks like a misbehaviour of the debug adapter: it remains running, but does not send an answer to the disconnect request. I noticed that we're not reacting to the failed launch request, but even if we detect that, we'd still have to disconnect from the debug adapter somehow, so the problem would be the same.

msujew: Hm, interesting. I wonder how vscode is dealing with this. It fails with the same error, but seems to resolve the issue (at least in the frontend) gracefully. I don't know what actually happens in the backend though, and whether everything is terminating there correctly.

thegecko: I've seen this before,  vscode has a lot more guards against misbehaving adapters

tsmaeder: I've introduced a timeout when sending the "disconnect" event. That should shield us from the debugger not responding to the event.

msujew (approved):
Looks good to me 👍 

- [x] Common debug adapters (js, markdown-debug) work well with the changes
- [x] Malfunctioning debug adapters are closed when a timeout is reached

@thegecko Did you test the changes with a C debug adapter/cortex-debug? I wasn't able to get mine to work correctly.

thegecko:
> @thegecko Did you test the changes with a C debug adapter/cortex-debug? I wasn't able to get mine to work correctly.

No, I tested a proprietary adapter we are developing

packages/debug/src/node/debug-adapter-session.ts
```diff
@@ -92,10 +85,18 @@ export class DebugAdapterSessionImpl implements DebugAdapterSession {
...
     async stop(): Promise<void> {
-        this.toDispose.dispose();
+        console.debug(`stopping debug adapter session: '${this.id}'`);
+
+        if (!this.isClosed) {
+            await this.debugAdapter.stop();
+        }
+        this.channel?.close();
+        this.channel = undefined;
     }
```
paul-marechal, line 101: The logic looks weird here: Shouldn't the channel track its own opened/closed state rather than storing it on the side like this (`this.isClosed`)?

packages/debug/src/node/stream-debug-adapter.ts
```diff
@@ -73,21 +80,48 @@ export class StreamCommunicationProvider extends DisposableCollection implements
...
+    constructor(socket: net.Socket) {
+        super(socket, socket);
+        this.socket = socket;
+    }
+
+    stop(): Promise<void> {
+        const deferred = new Deferred<void>();
+        this.socket.end(() => deferred.resolve());
+        return deferred.promise;
+    }
```
paul-marechal, line 126:
```ts
return new Promise(resolve => {
    this.socket.end(() => resolve());
});
```

paul-marechal (commented):
I think we would benefit from merging https://github.com/eclipse-theia/theia/pull/10265 before this one?

tsmaeder:
> > @thegecko Did you test the changes with a C debug adapter/cortex-debug? I wasn't able to get mine to work correctly.
> 
> No, I tested a proprietary adapter we are developing

@thegecko did you test after my latest changes?

packages/debug/src/browser/debug-session-connection.ts
```diff
@@ -186,44 +194,53 @@ export class DebugSessionConnection implements Disposable {
     sendCustomRequest<T extends DebugProtocol.Response>(command: string, args?: any): Promise<T> {
         return this.doSendRequest<T>(command, args);
     }
-    protected async doSendRequest<K extends DebugProtocol.Response>(command: string, args?: any): Promise<K> {
+
+    protected cancelPendingRequests(): void {
+        this.pendingRequests.forEach((deferred, requestId) => {
```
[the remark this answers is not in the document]
thegecko: Hmm, forEach is meant to be faster, but I tend to avoid it when dealing with promises.

thegecko:
> @thegecko did you test after my latest changes?

I have now and it still works

packages/debug/src/browser/debug-session-connection.ts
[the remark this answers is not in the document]
tsmaeder: IMO, both versions are perfectly acceptable 🤷

packages/debug/src/node/debug-adapter-session.ts
tsmaeder: Weird it may be, but I'd rather not change state management, because that would require retesting.

packages/debug/src/browser/debug-session-connection.ts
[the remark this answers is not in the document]
paul-marechal: @thegecko faster on what front?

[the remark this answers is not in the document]
paul-marechal: @tsmaeder right, you may take this comment with a grain of salt as I am nitpicking. But I don't like how the `forEach` method gives us elements in (value, key) order rather than (key, value) order, and also I don't understand why we would want to define anonymous functions to process elements when the language itself gives us `for` loops from the get-go. Feel free to leave as is.

packages/core/src/common/promise-util.ts
```diff
@@ -90,3 +92,17 @@ export function delay<T>(ms: number): (value: T) => Promise<T> {
...
+    const deferred = new Deferred<T>();
+    const disposable = event(() => {
+        disposable.dispose();
+        deferred.resolve(undefined);
+    }, thisArgs, disposables);
+    setTimeout(() => {
+        disposable.dispose();
+        deferred.reject(new CancellationError());
+    }, ms);
+    return deferred.promise;
```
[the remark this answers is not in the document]
paul-marechal: I just merged typings improvements to `Deferred` which should fail your line that resolves with `undefined`, you may want to update that.

packages/debug/src/browser/debug-session-connection.ts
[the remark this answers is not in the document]
paul-marechal: Following @thegecko comment I just tried comparing `for of` vs `.forEach` for maps, and `.forEach` seems to win everytime, call me enlightened... Why are we using TS/JS again? /jk

With that in mind, please keep the code as is.

paul-marechal:
@tsmaeder CI is red you may need to remove an import.

Also, do you feel this should land in 1.20.0 being released today or can this wait for 1.21.0?

packages/debug/src/node/debug-adapter-session.ts
paul-marechal: As a follow up then.

Issue #10354, opened 2021-10-29
TypeScript's `Find All References` and other requests are slow when used in this repository


paul-marechal:
For some reason TypeScript has been taking _much_ more time when responding to requests such as `Find All References`.

My current guess is that this is due to `tsserver@3.9.10` itself being slow, maybe upgrading to `@4` will fix it?

Issue #6501, opened 2019-11-06
remove deps to phosphorjs


akosyakov:
phosphor.js was archived: https://github.com/phosphorjs/phosphor#phosphorjs

We don't really use much from there except widgets. Also it's causing issues since phosphor.js comes with own concepts for events (signals), does not support disposables and does not live in DI context.

Instead of maintaining our own fork, it would be better to copy relevant bits, refactor them to use our events, clean up its dependencies, put in DI context and then clean up the shell. It would reduce amount of dependencies and size of bundled code as well.

kittaakos:
> It would reduce amount of dependencies and size of bundled code as well.

It sounds great, but would it worth the effort? The `@phosphor` folder is 1.2 MB compared to the ~500 MB size of the final electron application. Even if we do the refactoring and API adjustments, we still wouldn't be able to use async API for menus, for instance, and factories for `DockPanel`.

akosyakov:
> The @phosphor folder is 1.2 MB compared to the ~500 MB size of the final electron application.

For the browser context, we don't have electron and there less code we bundle less time a user waits to get an IDE.

> Even if we do the refactoring and API adjustments, we still wouldn't be able to use async API for menus, for instance, and factories for DockPanel.

Not sure what you mean. If we own code, we can change it as we like.

BroKun:
I totally agree with you that we are much better at events and DI than Phosphor. If we want to achieve a better experience than vscode, we need to make a lot of improvements, such as more flexible split-panel. The existing implementation of Phosphor may become a constraint.

akosyakov:
I think it is also better instead of using https://github.com/jupyterlab/lumino, since it will make it possible to integrate JLab widgets in our widget without version conflicts.

mikael-desharnais:
Hi
this can be quite an issue at the moment. 
Should we go on developing extensions that use the Phosphor lib & API ? or should we add a new library that would provide these features ?
Could you give us some hints on the solution that is the most likely to be followed ?
Regards,

akosyakov:
@mikael-desharnais you can use phorpshor.js for now, even if we merge it we will keep API the same just change namespaces.

thegecko:
It's not clear the effort involved switching to lumino, but assuming it's a lot of work, other options could be:

- Hard fork the repo and maintain our own widget framework we control ourselves (as suggested in the [README](https://github.com/phosphorjs/phosphor#phosphorjs))
[...]

PR #10355, opened 2021-10-29
repo: upgrade to TypeScript 4

paul-marechal:
@tsmaeder I remember saying we should avoid upgrading TypeScript, but I changed my mind.

I want to upgrade in order to build slightly faster and hope to fix an issue with our current version of `tsserver` being quite slow on our repo. Switching to TypeScript 4.4 I couldn't notice the lag anymore.

Upgrading the TS version we use to build this repo might lead to breakage in downstream applications in the following case:

- We start using a new TypeScript 4 construct. e.g. Named tuple elements (`type SomeTuple = [a: string, b: number]`)
- Downstream consumers of Theia use TypeScript 3 and it fails to parse this definition.

The solution would be for them to also upgrade their version of TypeScript in such a case. I think this is manageable.

As @msujew mentioned: TypeScript doesn't follow semantic versioning. But patch versions shouldn't bring breaking changes.

Closes https://github.com/eclipse-theia/theia/issues/10354.

#### How to test

- Building this repo should still work.
- Once everything is built, try running a few LS commands and get a feel for the performance. It shouldn't take to long to complete. It's quite subjective...

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,7 +4,6 @@
 // - Linux: $HOME/.config/Code/User/settings.json
 // - Mac: $HOME/Library/Application Support/Code/User/settings.json
 {
-  "tslint.enable": true,
   "editor.formatOnSave": true,
   "search.exclude": {
     "**/node_modules": true,
@@ -55,7 +54,8 @@
   "typescript.tsdk": "node_modules/typescript/lib",
   "files.insertFinalNewline": true,
   "clang-format.language.typescript.enable": false,
+  // ESLint `max-len` rule.
   "editor.rulers": [
     180
-  ], // ESLint `max-len` rule.
+  ],
 }
--- a/dev-packages/localization-manager/package.json
+++ b/dev-packages/localization-manager/package.json
@@ -33,7 +33,7 @@
     "deepmerge": "^4.2.2",
     "fs-extra": "^4.0.2",
     "glob": "^7.2.0",
-    "typescript": "^4.4.3"
+    "typescript": "~4.5.5"
   },
   "devDependencies": {
     "@theia/ext-scripts": "1.22.1"
--- a/license-check-baseline.json
+++ b/license-check-baseline.json
@@ -1,6 +1,5 @@
 {
   "npm/npmjs/-/eslint-plugin-deprecation/1.2.1": "Approved as 'works-with': https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22573",
   "npm/npmjs/-/jschardet/2.3.0": "Approved for Eclipse Theia: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22481",
-  "npm/npmjs/-/jsdom/11.12.0": "Approved as 'works-with': https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23640https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23640",
-  "npm/npmjs/-/node-pty/0.11.0-beta17": "Manually checked using ClearlyDefined"
+  "npm/npmjs/-/jsdom/11.12.0": "Approved as 'works-with': https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23640https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23640"
 }
--- a/packages/core/src/browser/shell/tab-bars.ts
+++ b/packages/core/src/browser/shell/tab-bars.ts
@@ -640,6 +640,7 @@ export class ToolbarAwareTabBar extends ScrollableTabBar {
     /**
      * Overrides the `contentNode` property getter in PhosphorJS' TabBar.
      */
+    // @ts-expect-error TS2611 `TabBar<T>.contentNode` is declared as `readonly contentNode` but is implemented as a getter.
     get contentNode(): HTMLUListElement {
         return this.tabBarContainer.getElementsByClassName(ToolbarAwareTabBar.Styles.TAB_BAR_CONTENT)[0] as HTMLUListElement;
     }
--- a/packages/core/src/browser/view-container.ts
+++ b/packages/core/src/browser/view-container.ts
@@ -1307,6 +1307,7 @@ export class ViewContainerLayout extends SplitLayout {
         return map(this.items, item => item.widget);
     }
 
+    // @ts-expect-error TS2611 `SplitLayout.widgets` is declared as `readonly widgets` but is implemented as a getter.
     get widgets(): ViewContainerPart[] {
         return toArray(this.iter());
     }
--- a/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
+++ b/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
@@ -511,11 +511,6 @@ export interface RenameLocation {
     text: string;
 }
 
-export interface RenameProvider {
-    provideRenameEdits(model: monaco.editor.ITextModel, position: Position, newName: string): PromiseLike<WorkspaceEdit & Rejection>;
-    resolveRenameLocation?(model: monaco.editor.ITextModel, position: Position): PromiseLike<RenameLocation & Rejection>;
-}
-
 export interface CallHierarchyDefinition {
     name: string;
     kind: SymbolKind;
--- a/packages/plugin-ext/src/main/node/paths/plugin-paths-service.ts
+++ b/packages/plugin-ext/src/main/node/paths/plugin-paths-service.ts
@@ -142,7 +142,7 @@ export class PluginPathsServiceImpl implements PluginPathsService {
 
     private async cleanupOldLogs(parentLogsDir: string): Promise<void> {
         // @ts-ignore - fs-extra types (Even latest version) is not updated with the `withFileTypes` option.
-        const dirEntries = await readdir(parentLogsDir, { withFileTypes: true });
+        const dirEntries = await readdir(parentLogsDir, { withFileTypes: true }) as string[];
         // `Dirent` type is defined in @types/node since 10.10.0
         // However, upgrading the @types/node in theia to 10.11 (as defined in engine field)
         // Causes other packages to break in compilation, so we are using the infamous `any` type...
--- a/doc/Migration.md
+++ b/doc/Migration.md
@@ -19,9 +19,16 @@ For example:
 }
 ```
 
+### v1.23.0
+
+#### TypeScript 4.5.5
+
+If you are using TypeScript <= 4.5.5 and you encounter issues when building your Theia application because your compiler fails to parse our type definitions,
+then you should upgrade to TypeScript >= 4.5.5.
+
 ### v1.22.0
 
-### Electron Update
+#### Electron Update
 
 Electron got updated from 9 to 15, this might involve some modifications in your code based on the new APIs.
 
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "tslint": "^5.12.0",
     "typedoc": "0.20.36",
     "typedoc-plugin-external-module-map": "1.2.1",
-    "typescript": "^3.9.2",
+    "typescript": "~4.5.5",
     "uuid": "^8.0.0",
     "yargs": "^15.3.1"
   },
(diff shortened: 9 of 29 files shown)


packages/plugin-ext/src/common/plugin-api-rpc-model.ts
```diff
@@ -512,8 +512,8 @@ export interface RenameLocation {
 }
 
 export interface RenameProvider {
-    provideRenameEdits(model: monaco.editor.ITextModel, position: Position, newName: string): PromiseLike<WorkspaceEdit & Rejection>;
-    resolveRenameLocation?(model: monaco.editor.ITextModel, position: Position): PromiseLike<RenameLocation & Rejection>;
+    provideRenameEdits(model: monaco.editor.ITextModel, position: theia.Position, newName: string): PromiseLike<WorkspaceEdit & Rejection>;
+    resolveRenameLocation?(model: monaco.editor.ITextModel, position: theia.Position): PromiseLike<RenameLocation & Rejection>;
```
paul-marechal, line 516: @eclipse-theia/plugin-system In a different PR there was a problem with this `Position` type too.

TypeScript's old DOM typings used to define some `Position` interface available globally, and I believe that this is what was wrongfully referenced here.

I replaced it with `theia.Position` because new TypeScript DOM typings don't define it anymore, but I don't understand if that's what's expected here.

paul-marechal: Actually there's no references to this interface, I'll just remove it.

license-check-baseline.json
```diff
@@ -4,5 +4,5 @@
   "npm/npmjs/-/jsdom/11.12.0": "Approved as 'works-with': https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23640https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23640",
   "npm/npmjs/-/jsmin/1.0.1": "Approved as 'works-with': https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23640https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23640",
   "npm/npmjs/-/seek-bzip/1.0.6": "Approved by CQ: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=17349#c7",
-  "npm/npmjs/-/typescript/4.4.3": "Approved manually: https://github.com/eclipse-theia/theia/pull/10247"
+  "npm/npmjs/-/typescript/4.4.4": "From ClearlyDefined: Apache-2.0 with a score of 78"
```
vince-fugnitto, line 7: We might be able to remove this entry altogether:

```
WARN: Some entries in the baseline did not match anything from dash-licenses output:
> npm/npmjs/-/seek-bzip/1.0.6
npm/npmjs/-/seek-bzip/1.0.6: Approved by CQ: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=17349#c7
> npm/npmjs/-/typescript/4.4.4
npm/npmjs/-/typescript/4.4.4: From ClearlyDefined: Apache-2.0 with a score of 78
```

tsmaeder:
@paul-marechal I'm not against keeping up with typescript versions, but I'm getting a bit worried with the rate of changes we have in our build system. We don't want adopters to have to catch up every release. Would it be possible to update the tooling (which is what we want) and keep the syntax checking at 3.9? 
It would be nice if we could be more predictable with such changes, such as announcing breaking updates to tooling a release in advance.

paul-marechal:
> We don't want adopters to have to catch up every release.

The build system shouldn't impact adopters that are consuming the npm packages we are publishing, like at all?

Beside the potentially new type definition, which won't cause any issue until the moment someone makes a PR using new things from TypeScript.

For those that build the repo themselves, doing `yarn` alone should do everything up until the point where all packages are built and ready for use. Unless the wiring is deeply plugged into how Theia is built?

paul-marechal:
> and keep the syntax checking at 3.9?

Not AFAIK, we'd have to keep an eye out on what goes in but that doesn't sound super robust.

tsmaeder:
> The build system shouldn't impact adopters that are consuming the npm packages we are publishing, like at all?

Since we have no way to ensure that no > 4.0 constructs are used in Theia, they will effectively have to move to 4.0, as well eventually. Looking at the Typescript changelog, that might be a breaking change.

packages/core/src/browser/shell/tab-bars.ts
```diff
@@ -615,6 +615,7 @@ export class ToolbarAwareTabBar extends ScrollableTabBar {
     /**
      * Overrides the `contentNode` property getter in PhosphorJS' TabBar.
      */
+    // @ts-expect-error TS2611 because Phosphor.js implements a getter for `TabBar<T>.contentNode` but it typed it as `readonly contentNode`.
```
colin-grant-work, line 618: See comment below on the nature of the error.

packages/core/src/browser/view-container.ts
```diff
@@ -1277,6 +1277,7 @@ export class ViewContainerLayout extends SplitLayout {
         return map(this.items, item => item.widget);
     }
 
+    // @ts-expect-error TS2611 because Phosphor.js implements a getter for `SplitLayout.widgets` but typed it as `readonly widgets`.
```
colin-grant-work, line 1280: I don't think comment is quite right. I think that older versions of TS, like the one Phosphor was most recently compiled with, didn't distinguish between properties and getters, so the `.d.ts` file shows this like it's a property even though it isn't, and then TS is looking at the `.d.ts` file and seeing a property, even though the class actually has a getter.

packages/debug/src/browser/debug-session-connection.ts
```diff
@@ -290,6 +290,8 @@ export class DebugSessionConnection implements Disposable {
                 this.onDidCustomEventEmitter.fire(event);
             }
         } else {
+            // TODO: Does this branch even make sense?
+            // @ts-expect-error TS2559
```
colin-grant-work, line 294: This one should go away with @tsmaeder's work on the debug system.

colin-grant-work (commented):
Looks fine to me. Minor comments on the `@ts-expect-error` comments. Probably if we compiled Phosphor ourselves with a later TS version, we could get rid of those, but that doesn't seem worth it. I don't love the number of new `any`'s that are introduced, but they point to places where we're not being as clear about our intent as we should be, and I'm fine with them for now. 

The upgrade didn't seem to improve my build or code browsing experience, but I agree with @paul-marechal that these are chores that we're stuck with - we can't really plan to lag behind indefinitely, so we'll have to do this at some point. Certainly open to discussion about the schedule on which we should upgrade our tools, but this doesn't seem hasty to me.

msujew (commented):
By the way, I can confirm that running `Find All References` with TypeScript 4.4.4 runs way faster in Theia now, compared to 3.9. However, the current default for the TypeScript extension seems to be 3.8, so I had to manually change it to the workspace version to verify this.

paul-marechal:
> the current default for the TypeScript extension seems to be 3.8

How? I thought the extension would use what VS Code bundles I thought it would be using 4.4?

We also have the TypeScript version set to `node_modules/typescript/sdk` in our `.vscode/settings.json` file, so it should pick up the locally installed one?

packages/debug/src/browser/debug-session-connection.ts
paul-marechal: Good to know, but I still need to keep this here for now.

msujew:
> How? I thought the extension would use what VS Code bundles I thought it would be using 4.4?

@paul-marechal I was confused as well, but we currently use the `1.45.1` version of the `typescript-language-features` extension in the Theia repo, which comes bundled with TypeScript `3.8.3`:

image

And you're right as well that we added the appropriate setting to use the `node_modules/typescript/lib` library, but there's a huge catch hidden in the [docs of the extension](https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-the-workspace-version-of-typescript):

> The `typescript.tsdk` workspace setting only tells VS Code that a workspace version of TypeScript exists. To actually start using the workspace version for IntelliSense, you must run the TypeScript: Select TypeScript Version command and select the workspace version.

So the setting only tells the extension which workspace version should be used, but it doesn't tell it to use that one from the start. The user has switch manually. I just learned about that one as well.

paul-marechal:
Thanks for pointing all of this out @msujew.

> VS Code will automatically detect workspace versions of TypeScript that are installed under `node_modules` in the root of your workspace.

This means that the `typescript.tsdk` setting we had is useless now, so I removed it.

packages/core/src/browser/view-container.ts
paul-marechal: I tried to be the most concise with what is happening without going into why, but you are totally right.

But it stays true that the typings describe this field as a readonly prop and we try to redefine it as a getter, causing TS2611. I just don't think it is super useful explaining the historical reason in this specific comment.

We could indeed avoid this issue by compiling Phosphor.js using a newer TypeScript, but it isn't worth it.

FYI: https://github.com/eclipse-theia/theia/issues/6501

colin-grant-work: How about `the Phosphor field is declared as a property`. The passive avoids the attribution to the Phosphor authors that 'implements a getter for `SplitLayout.widgets` but typed it as `readonly widgets`' implies.

paul-marechal: Sounds good to me! Updated.

.vscode/settings.json
```diff
@@ -52,10 +51,10 @@
...
     "editor.defaultFormatter": "vscode.json-language-features",
   },
-  "typescript.tsdk": "node_modules/typescript/lib",
   "files.insertFinalNewline": true,
   "clang-format.language.typescript.enable": false,
   "editor.rulers": [
     180
-  ], // ESLint `max-len` rule.
+  ],
+  "typescript.tsdk": "node_modules\\typescript\\lib", // ESLint `max-len` rule.
```
paul-marechal, line 59: Well I guess the extension forcefully writes where it found it... So much for removing it.

paul-marechal:
I rebased this PR on master and also pushed TypeScript to 4.5.

JonasHelming:
@tsmaeder

dev-packages/localization-manager/package.json
```diff
@@ -33,7 +33,7 @@
     "deepmerge": "^4.2.2",
     "fs-extra": "^4.0.2",
     "glob": "^7.2.0",
-    "typescript": "^4.4.3"
+    "typescript": "~4.4.4"
```
msujew, line 36: Minor: Please update this to `~4.5.5` as well.

msujew (approved):
Thanks Paul! Looks good to me, I have a small remark, but I'll approve this already 👍

dev-packages/localization-manager/package.json
paul-marechal: Good catch.

packages/plugin-ext/src/main/node/paths/plugin-paths-service.ts
```diff
@@ -142,7 +142,7 @@ export class PluginPathsServiceImpl implements PluginPathsService {
 
     private async cleanupOldLogs(parentLogsDir: string): Promise<void> {
         // @ts-ignore - fs-extra types (Even latest version) is not updated with the `withFileTypes` option.
-        const dirEntries = await readdir(parentLogsDir, { withFileTypes: true });
+        const dirEntries = await readdir(parentLogsDir, { withFileTypes: true }) as string[];
```
vince-fugnitto, line 145: I believe the `await` is unnecessary?

vince-fugnitto (approved):
I confirmed that 
- editing in vscode still shows markers for tslint
- typescript functionality works (and feels quicker)
- the application builds and starts

packages/plugin-ext/src/main/node/paths/plugin-paths-service.ts
paul-marechal: `readdir` returns `Promise<string[]>` here, TypeScript is confused because we don't respect the typings (see comment above).

`as string[]` applies to the result of the `(await ...)` expression.

colin-grant-work (approved):
LGTM and seems to work at least as well. Excited to get better type check handling, and some day I may find the time to address the new `any`'s :-).

JonasHelming:
@paul-marechal : Let's merge this!

msujew:
@paul-marechal Mergeable for this release or postpone until after the release?

PR #10357, opened 2021-10-29
Implement multiroot Open Folder command

### LINKED ISSUE #10334: Permit multiselection in "Open Folder"
Currently the "Open Folder" command only allows one selection. However, in VSCode multiple folders can be selected at once, this has the advantage of allowing a user to create a new multiroot workspace in a single step. The behavior is VSCode is as follows:

**Starting with no workspace:**
1. Select "Open Folder" and select multiple folders
1. Select "Ok"
1. A multiroot workspace with all of the selected folders is created in the same instance

**Start with an existing workspace:**
- Repeat steps 1-2 above
- If the workspace has been saved then a *new* workspace will be opened with the selected folders in the same instance
- If the workspace has not been saved, user will be prompted to save their current workspace before a new workspace is opened in the same instance

### LINKED ISSUE #10366: All folder entries in workspace file (.theia-workspace) are treated as relative
### Bug Description:

If a workspace file contains folder entries with absolute paths, those paths are still resolved relative to the workspace configuration file.

### Steps to Reproduce:

1. Create a workspace configuration file that specifies its roots using absolute paths.
```jsonc
{ // in /a/b/c/good-workspace.theia-workspace or /a/b/c/good-workspace.code-workspace
  "folders": [
    { "path": "/d/e/f" }
  ]
}
```
2. Open that file using `Open Workspace...`
3. Open the navigator widget and hover over the root folder.
4. Observe that the path given is `/a/b/c//d/e/f`. Note double slash between `c` and `d`.

Expected: should treat absolute paths as absolute and resolve the folder to `/d/e/f`

### Additional Information

- Operating System: RHEL7 (Electron)
- Theia Version: bca6eda5a46d485a98f9eec96a5ce7c45d9262c1

colin-grant-work:
#### What it does

Fixes #10334 by adding handling for multiselection to the `Open Folder...` command.
Fixes #10366 by implementing `resolveToAbsolute`  for URI's.

#### How to test

1. Use the 'Open Folder...' command.
2. Select multiple folders.
3. Observe that a new multi-root workspace is created and opened.
4. It should work from either an existing workspace or a no-workspace state.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/workspace/src/browser/workspace-commands.ts
+++ b/packages/workspace/src/browser/workspace-commands.ts
@@ -354,16 +354,7 @@ export class WorkspaceCommandContribution implements CommandContribution {
                     const workspaceSavedBeforeAdding = this.workspaceService.saved;
                     await this.addFolderToWorkspace(...uris);
                     if (!workspaceSavedBeforeAdding) {
-                        const saveCommand = registry.getCommand(WorkspaceCommands.SAVE_WORKSPACE_AS.id);
-                        if (saveCommand && await new ConfirmDialog({
-                            title: nls.localize('theia/workspace/workspaceFolderAddedTitle', 'Folder added to Workspace'),
-                            msg: nls.localize('theia/workspace/workspaceFolderAdded',
-                                'A workspace with multiple roots was created. Do you want to save your workspace configuration as a file?'),
-                            ok: Dialog.YES,
-                            cancel: Dialog.NO
-                        }).open()) {
-                            registry.executeCommand(saveCommand.id);
-                        }
+                        this.saveWorkspaceWithPrompt(registry);
                     }
                 }
             });
@@ -539,6 +530,19 @@ export class WorkspaceCommandContribution implements CommandContribution {
             if (await dialog.open()) {
                 await this.workspaceService.removeRoots(toRemove);
             }
+        }
+    }
+
+    async saveWorkspaceWithPrompt(registry: CommandRegistry): Promise<void> {
+        const saveCommand = registry.getCommand(WorkspaceCommands.SAVE_WORKSPACE_AS.id);
+        if (saveCommand && await new ConfirmDialog({
+            title: nls.localize('theia/workspace/workspaceFolderAddedTitle', 'Folder added to Workspace'),
+            msg: nls.localize('theia/workspace/workspaceFolderAdded',
+                'A workspace with multiple roots was created. Do you want to save your workspace configuration as a file?'),
+            ok: Dialog.YES,
+            cancel: Dialog.NO
+        }).open()) {
+            return registry.executeCommand(saveCommand.id);
         }
     }
 
--- a/packages/workspace/src/browser/workspace-service.ts
+++ b/packages/workspace/src/browser/workspace-service.ts
@@ -24,7 +24,7 @@ import {
 } from '@theia/core/lib/browser';
 import { Deferred } from '@theia/core/lib/common/promise-util';
 import { EnvVariablesServer } from '@theia/core/lib/common/env-variables';
-import { ILogger, Disposable, DisposableCollection, Emitter, Event, MaybePromise, MessageService } from '@theia/core';
+import { ILogger, Disposable, DisposableCollection, Emitter, Event, MaybePromise, MessageService, nls } from '@theia/core';
 import { WorkspacePreferences } from './workspace-preferences';
 import * as jsoncparser from 'jsonc-parser';
 import * as Ajv from 'ajv';
@@ -298,7 +298,7 @@ export class WorkspaceService implements FrontendApplicationContribution {
         if (this._workspace) {
             const displayName = this._workspace.name;
             if (this.isWorkspaceFile(this._workspace)) {
-                title = this.isUntitledWorkspace(this._workspace.resource) ? 'Untitled (Workspace)' : displayName.slice(0, displayName.lastIndexOf('.'));
+                title = this.isUntitledWorkspace(this._workspace.resource) ? nls.localizeByDefault('Untitled (Workspace)') : displayName.slice(0, displayName.lastIndexOf('.'));
             } else {
                 title = displayName;
             }
@@ -515,7 +515,7 @@ export class WorkspaceService implements FrontendApplicationContribution {
             this.setURLFragment('');
         }
 
-        window.location.reload(true);
+        this.windowService.reload();
     }
 
     protected openNewWindow(workspacePath: string): void {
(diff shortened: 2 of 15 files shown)


kenneth-marut-work:
Looks great, though I found a case where I'm getting some odd behavior

- Start with an empty workspace (close your existing workspace if you need to)
- `Open Folder` and select multiple folders (this will create the untitledWorkspace)
- Don't save the workspace
- `Open Folder` and select some new folders
- Observe that a second window opens with a workspace consisting of the new folders, but the first window also changes its workspace to the second window's workspace

I believe that VSCode doesn't actually launch a second instance. You're first prompted to save your untitled workspace then it just swaps out the first workspace for the second

colin-grant-work:
> Looks great, though I found a case where I'm getting some odd behavior
> 
> * Start with an empty workspace (close your existing workspace if you need to)
> * `Open Folder` and select multiple folders (this will create the untitledWorkspace)
> * Don't save the workspace
> * `Open Folder` and select some new folders
> * Observe that a second window opens with a workspace consisting of the new folders, but the first window also changes its workspace to the second window's workspace
> 
> I believe that VSCode doesn't actually launch a second instance. You're first prompted to save your untitled workspace then it just swaps out the first workspace for the second

Thanks for checking it out. I have pushed a patch that should handle this better.

colin-grant-work:
@vince-fugnitto, @paul-marechal, to fix #10366 I've added an implementation of `Path.resolve` to our `Path` class, which seems in the spirit of that class: it's probably mostly fine, but it's a bit weird that we have it.

Alternatively, we could follow [VSCode](https://github.com/microsoft/vscode/blob/main/src/vs/base/common/path.ts) and just include the NodeJS `Path` package in our browser packages.

Let me know what you think.

msujew (commented):
> Thanks for checking it out. I have pushed a patch that should handle this better.

I just retested it but was still able to reproduce the issue in another configuration:

1. Open a multi-root workspace
2. Open a new window
3. Open another multi-root workspace using `Open Folder` in that new window
4. The old window now uses the new multi-root workspace as well

colin-grant-work:
> > Thanks for checking it out. I have pushed a patch that should handle this better.
> 
> I just retested it but was still able to reproduce the issue in another configuration:
> 
> 1. Open a multi-root workspace
> 2. Open a new window
> 3. Open another multi-root workspace using `Open Folder` in that new window
> 4. The old window now uses the new multi-root workspace as well

Fair enough. The problem is that we always generate the same untitled workspace. VSCode uses a time-stamped system so that every new untitled workspace is distinct, presumably with some cleanup when you leave the workspace behind. I will implement something similar.

packages/workspace/src/browser/workspace-service.ts
```diff
@@ -437,7 +435,9 @@ export class WorkspaceService implements FrontendApplicationContribution {
     }
 
     async getUntitledWorkspace(): Promise<URI> {
-        return getTemporaryWorkspaceFileUri(this.envVariableServer);
+        const configDirURI = await this.envVariableServer.getConfigDirUri();
+        const semiRandomID = (Date.now() + Math.floor(Math.random() * 1000)).toString();
+        return new URI(configDirURI).resolve(`workspaces/Untitled-${semiRandomID}.${THEIA_EXT}`);
```
msujew, line 440: Suggestion: Instead of generating a random ID, we could first generate an URI for `Untitled.theia`, check whether that already exists, increment, generate `Untitled-1`, etc. until we arrive at a number, which does not exist. I think that would be preferable, but takes a bit more time, especially if we have a large amount of untitled workspaces. (unlikely?) What do you think?

msujew (approved):
The changes look good to me:

- [x] Enables selecting multiple folders when running `Open Folder`
- [x] Automatically creates an untitled workspace
- [x] Displays a dialog asking to save untitled workspaces

The untitled workspace always containing a suffix (in the file explorer) isn't optimal, but I don't see a better way around that right now to prevent overwriting other untitled workspaces. I do have one suggestion, but that doesn't prevent this from getting merged.

packages/workspace/src/browser/workspace-service.ts
colin-grant-work: That's doable. The main risk is in the browser, where we can't guarantee that we'll actually get to run our cleanup code. In that context, people could generate a lot of untitled workspaces without realizing it. But in any case, whether the suffix is random or ordinal, I think it's a good idea to check for existence before saving in any case. Even a small chance of clobbering someone's active workspace should be avoided.

colin-grant-work: I decided to leave the random suffix, because the ordinal approach is guaranteed to get linearly worse if the files don't get cleaned up, whereas the random approach will get only probabilistically worse with each file, but I have added code that checks for existence before returning the URI. I've also added code that cleans up a now-unused untitled file when the 'save workspace as' command is run.

vince-fugnitto (commented):
@colin-grant-work I noticed something:

- if no workspace is present, and we perform the `open folder...` with multiple folders, the dialog appears (which occurs on unload), selecting `don't save` will close the dialog and restart the application without a workspace present. In vscode this same use-case will open the untitled workspace which was my expectation as a user.


https://user-images.githubusercontent.com/40359487/148396894-17beaed3-71bf-4576-98d7-f75e8b323547.mp4

colin-grant-work:
@vince-fugnitto, thanks for the review and good catch. I've updated the code not to delete an untitled workspace file on exit, and instead added code to the backend WorkspaceServer that deletes untitled workspaces files that aren't in the top 10 most recent workspaces used. It's not foolproof: if you've kept an untitled workspace open while you've opened a bunch of other workspaces, and then your backend goes down and you have to restart, then the logic would delete that file even though it's still in use. If you have any suggestions for how to perform _some_ cleanup without the risk of deleting a file in use, I'm happy to hear them.

packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -542,6 +533,19 @@ export class WorkspaceCommandContribution implements CommandContribution {
...
     }
 
+    async saveWorkspaceWithPrompt(registry: CommandRegistry): Promise<void> {
+        const saveCommand = registry.getCommand(WorkspaceCommands.SAVE_WORKSPACE_AS.id);
+        if (saveCommand && await new ConfirmDialog({
+            title: nls.localize('theia/workspace/workspaceFolderAddedTitle', 'Folder added to Workspace'),
+            msg: nls.localize('theia/workspace/workspaceFolderAdded',
+                'A workspace with multiple roots was created. Do you want to save your workspace configuration as a file?'),
+            ok: Dialog.YES,
+            cancel: Dialog.NO
```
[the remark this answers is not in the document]
msujew: They are localized, they are defined like this:

https://github.com/eclipse-theia/theia/blob/450d257d5425e6c2893b5abd3bdc2a41641a2c4d/packages/core/src/browser/dialogs.ts#L69-L74

Issue #7590, opened 2020-04-16
Warn before closing a terminal


sh7dm:
### Description
Show a warning if user attempts to close a terminal which has running processes in it.

### Reproduction Steps
1. Run a command (e.g. `top`) in Theia terminal and any other one.
2. Close both terminals

**OS and Theia version:**
OS: Ubuntu 19.10
Theia: `master`

**Diagnostics:**

akosyakov:
I am not sure whether it is possible. Terminals are stream based, we don't really know what a user is doing. We could have something like that for tasks.

cc @elaihau @RomanNikitenko @marechal-p ?

marechal-p:
I think a new internal method was added to test if a shell has a subprocess running (meaning the shell is in use). See https://github.com/eclipse-theia/theia/blob/master/packages/terminal/src/node/shell-terminal-server.ts#L66-L88

elaihau:
> I think a new internal method was added to test if a shell has a subprocess running (meaning the shell is in use). See https://github.com/eclipse-theia/theia/blob/master/packages/terminal/src/node/shell-terminal-server.ts#L66-L88

+1

akosyakov:
We should check though it does not break VS Code extensions. If they can close terminal programmatically then there should not be such dialogs. Maybe also make it configurable via preferences.

sh7dm:
@akosyakov right, it should work only when clicking `X` button manually, not by commands. Preference should be added as well, but its default must prevent closing.

PR #10374, opened 2021-11-02
terminal: add `confirmOnExit` preference

vince-fugnitto:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/7590

The pull-request adds support for `terminal.integrated.confirmOnExit` which is used to prevent closing the application if there are any terminals present.

At the moment, the following is supported:
- `never`: never warn when closing the application if terminals are present (default)
- `always`: always warn when closing the application if terminals are present
- `hasChildProcesses` - currently treated like `always` due to `async` limitations

#### How to test

Verify the following use-cases:

| Use-Case | Preference Value | Terminals | Expectation |
|:---|:---|:---|:---|
| 1 | `never` | no terminal | should not warn when closing the app |
| 2 | `never` | one or more terminals | should not warn when closing the app |
| 3 | `always` or `hasChildProcesses` | no terminal | should not warn when closing the app |
| 4 | `always` or `hasChildProcesses` | one or more terminals | should warn when closing the app |

#### Review checklist

#### Reminder for reviewers

--- 

Note: since `onWillStop` is synchronous, and `hasChildProcesses` is async, the preference was treated like `always` for the time being. If anyone has suggestions on how to better support the actual preference it is most welcome. One idea would maybe be to poll the backend for `hasChildProcesses` on all open terminals every so often, but there were concerns over performance.

### DIFF
--- a/packages/terminal/src/browser/terminal-frontend-contribution.ts
+++ b/packages/terminal/src/browser/terminal-frontend-contribution.ts
@@ -29,7 +29,7 @@ import {
 } from '@theia/core/lib/common';
 import {
     ApplicationShell, KeybindingContribution, KeyCode, Key, WidgetManager,
-    KeybindingRegistry, Widget, LabelProvider, WidgetOpenerOptions, StorageService, QuickInputService, codicon, CommonCommands
+    KeybindingRegistry, Widget, LabelProvider, WidgetOpenerOptions, StorageService, QuickInputService, codicon, CommonCommands, FrontendApplicationContribution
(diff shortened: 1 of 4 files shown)


colin-grant-work (approved):
Behavior is as described in both Electron and browser, the code looks good, and the preference is correctly localized. #10379 would let us get over the limitation on `async` code checking for child processes in Electron as well as show a dialog tailored to the preference, but this can be merged first and the necessary changes made in #10379.

PR #10375, opened 2021-11-03
Fix #10369: icons of custom tree view row are not visible on mouse hover

### LINKED ISSUE #10369: the action icon of tree-view extensions with menu icon contribution doesn't show up on hover
### Bug Description:
when contributing a tree view menu item - the icon is showing up on hover as to vs code.
i tested on theia 15 - the icon appears when hovering.
but in theia 18 the icon appears only after the node is selected 

### Steps to Reproduce:

1. add a tree view extension with menu contribution
2. open the tree view to explore its content
3. hover on a node
expected - the icon should show up 
current - the icon shows up only after line is selected.

icons-bug

### Additional Information

- Operating System:
- Theia Version:

EstherPerelman:

#### What it does
Fix #10369 by replacing `update` to `forceUpdateGrid` call on `mouseover` event since the `update` request not activate the rendering of the grid by react (it doesn't pass the internal React.Grid `_calculateChildrenToRender` function because of old caches)

#### How to test (also for buggy result)

1. install `tree-view-sample` extension of vscode-extensions-samples on Theia (you can easily download my [custom-view-sample.zip](https://github.com/eclipse-theia/theia/files/7465409/custom-view-sample.zip) )
2. from the top menu select `View` -> `Open View` -> `Packege Explorer`
3. hover with the mouse on the rows
4. **good result**: see edit icon for each row on hover, **bad result**: see edit icon only on selecting row 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/style/tree.css
+++ b/packages/plugin-ext/src/main/browser/style/tree.css
@@ -41,6 +41,6 @@
     margin-left: 17px;
 }
 
-.theia-TreeNode:not(:hover):not(.theia-mod-selected) .theia-tree-view-inline-action {
+.theia-tree-view .theia-TreeNode:not(:hover):not(.theia-mod-selected) .theia-tree-view-inline-action {
     display: none;
 }
--- a/packages/plugin-ext/src/main/browser/style/tree.css
+++ b/packages/plugin-ext/src/main/browser/style/tree.css
@@ -40,3 +40,7 @@
     margin-bottom: 10px;
     margin-left: 17px;
 }
+
+.theia-TreeNode:not(:hover):not(.theia-mod-selected) .theia-tree-view-inline-action {
+    display: none;
+}
--- a/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
+++ b/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
@@ -45,6 +45,7 @@ import CoreURI from '@theia/core/lib/common/uri';
 import { ContextKeyService } from '@theia/core/lib/browser/context-key-service';
 import * as markdownit from 'markdown-it';
 import { isMarkdownString } from '../../../plugin/markdown-string';
+import debounce = require('p-debounce');
 
 export const TREE_NODE_HYPERLINK = 'theia-TreeNodeHyperlink';
 export const VIEW_ITEM_CONTEXT_MENU: MenuPath = ['view-item-context-menu'];
@@ -359,8 +360,10 @@ export class TreeViewWidget extends TreeViewWelcomeWidget {
     protected hoverNodeId: string | undefined;
     protected setHoverNodeId(hoverNodeId: string | undefined): void {
         this.hoverNodeId = hoverNodeId;
-        this.update();
+        this.forceUpdateGrid();
     }
+
+    protected forceUpdateGrid = debounce(() => this.view?.list?.forceUpdateGrid(), 1);
 
     protected createNodeAttributes(node: TreeNode, props: NodeProps): React.Attributes & React.HTMLAttributes<HTMLElement> {
         return {
--- a/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
+++ b/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
@@ -37,7 +37,7 @@ import { MenuPath, MenuModelRegistry, ActionMenuNode } from '@theia/core/lib/com
 import * as React from '@theia/core/shared/react';
 import { PluginSharedStyle } from '../plugin-shared-style';
 import { ViewContextKeyService } from './view-context-key-service';
-import { Widget } from '@theia/core/lib/browser/widgets/widget';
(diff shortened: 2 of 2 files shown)


packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
```diff
@@ -359,9 +360,15 @@ export class TreeViewWidget extends TreeViewWelcomeWidget {
...
-        this.update();
+        this.forceUpdateGrid();
     }
 
+    protected forceUpdateGrid = debounce(() => {
+        if (this.view && this.view.list) {
+            this.view.list.forceUpdateGrid();
+        }
+    }, 10);
+
```
msujew, line 371:
```suggestion
        this.view?.list?.forceUpdateGrid();
    }
```

msujew (commented):
Thanks @EstherPerelman for taking care of this.

While this works on a surface level, I noticed some issues with debouncing the forced update here. Especially when swiftly moving over the tree nodes (where the hover marker will lag behind a few frames). I tried the change below and it seems to fix the issue as well. Are there any issues with this approach that I didn't notice?

msujew (commented):
@EstherPerelman I still feel like debouncing here is not necessary and only leads to artifacts when swiftly moving the cursor over the tree nodes. Notice how the icon always lags behind the cursor, while the background highlighting correctly moves with the cursor:

https://user-images.githubusercontent.com/4377073/140354152-ba51034d-c980-48bb-be0e-5aa480664623.mp4

When removing the debounce, the feature continues to work as expected, without any of the artifacts, at least in my tests. What speaks in favor of debouncing the forced update?

EstherPerelman:
> What speaks in favor of debouncing the forced update?

@msujew I added the debounce because in my test(without the debounce) the `edit` icon wasn't clickable - didn't show the information message as happens now,
Also the icon not always appears `onMouseOver` event as you can see in the video:

https://user-images.githubusercontent.com/71729183/140431705-e8d6c48b-85ac-411b-bb85-6bb64457395a.mp4

msujew:
@EstherPerelman Alright, thank you for the explanation. That makes me believe that forcing an update here is the wrong direction. I have another idea:

1. Add this CSS to the [`tree.css`](https://github.com/eclipse-theia/theia/blob/6a376d3ce5dd66a3750887f05087c387ef49ab6b/packages/plugin-ext/src/main/browser/style/tree.css) file:
```css
.theia-TreeNode:not(:hover):not(.theia-mod-selected) .theia-tree-view-inline-action {
    display: none;
}
```
2. Always render the inline commands by removing these lines of code:
https://github.com/eclipse-theia/theia/blob/6a376d3ce5dd66a3750887f05087c387ef49ab6b/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx#L331-L333
3. Add the `ACTION_ITEM` class constant (imported from `@theia/core/lib/browser/widgets/widget`) to the inline commands:
https://github.com/eclipse-theia/theia/blob/6a376d3ce5dd66a3750887f05087c387ef49ab6b/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx#L353
4. Remove the forced update on the `setHoverNodeId` method.

That way we always render the inline commands to html, but don't display them if it's not necessary. What do you think about that?

EstherPerelman:
> That way we always render the inline commands to html, but don't display them if it's not necessary. What do you think about that?

@msujew, in the first time I looked at this code area I wondered why the `hover` was not implemented with css... but preferred not to touch the existing code, with your suggestion it's working perfect and the code is more reasonable

msujew:
@EstherPerelman Yeah, it seems to be some ancient code that's been there since the first version of the inline commands and has never been touched since. The new approach looks way better.

@colin-grant-work @vince-fugnitto Any possible regressions that you might see due to the CSS approach compared to using `setHoverNodeId`?

packages/plugin-ext/src/main/browser/style/tree.css
```diff
@@ -40,3 +40,7 @@
     margin-bottom: 10px;
     margin-left: 17px;
 }
+
+.theia-TreeNode:not(:hover):not(.theia-mod-selected) .theia-tree-view-inline-action {
```
vince-fugnitto, line 44: @msujew should we prefix the entry with `.theia-tree-view` as not to potentially affect other views and keep the styling specific to the plugin-system?

```suggestion
.theia-tree-view .theia-TreeNode:not(:hover):not(.theia-mod-selected) .theia-tree-view-inline-action {
```

vince-fugnitto (commented):
I confirmed that the behavior works well on hover :+1:

packages/plugin-ext/src/main/browser/style/tree.css
msujew: Sure, sounds like a good idea 👍

EstherPerelman:
Hi @vince-fugnitto @msujew I pushed the change, Can you merge it soonish?

msujew (approved):
LGTM! I just reconfirmed that everything works as expected.

I will probably merge this tomorrow, if there are no objections.

PR #10378, opened 2021-11-03
Remove from recently opened workspaces list

### LINKED ISSUE #4884: [workspaces] support removing workspaces from 'recently opened workspaces'
Currently, there is no way to easily remove `workspaces` from the `Open Recent Workspace...` quick-open menu. The only way to remove workspace items from the list, is to remove the `recentworkspace.json` completely from the user's home under `.theia`.

VSCode supports directly removing workspaces from the quick-open, and I suggest we do the same for end-users.

alvsan09:
#### What it does
Fixes #4884

Added the ability to remove Recent Workspaces from the Open Recent
Workspace... quick-open command palette.

Users can now directly remove items from the list by selecting an
associated remove button (X).

#### How to test

1) Open the command palette (Ctrl+Shft+P)
2) Type or select the command "File: Open recent workspace"
3) Using the mouse, hover over the different entries in the list and make sure to locate a `close` icon 'X' at the end of the item in focus.
4) Click the 'X' icon to remove the item from the list
5) Refresh the browser and make sure the workspace removed in the previous step does not appear

Other instances of quick access should not be impacted, try few e.g. from the main menu Terminal -> Run Task, and any other run command, etc.
remove_from_recent_opened_workspaces

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-quick-input-service.ts
+++ b/packages/monaco/src/browser/monaco-quick-input-service.ts
@@ -257,16 +257,8 @@ export class MonacoQuickInputService implements QuickInputService {
                             {
                                 ...evt,
                                 removeItem: () => {
-                                    const index = wrapped.items.indexOf(evt.item);
-                                    if (index !== -1) {
-                                        const filteredItems = wrapped.items.slice();
-                                        const removed = filteredItems.splice(index, 1);
-                                        const activeFilteredItems = wrapped.activeItems.filter(item => item !== removed[0]);
-                                        wrapped.items = filteredItems;
-                                        if (activeFilteredItems) {
-                                            wrapped.activeItems = activeFilteredItems;
-                                        }
-                                    }
+                                    wrapped.items = wrapped.items.filter(item => item !== evt.item);
+                                    wrapped.activeItems = wrapped.activeItems.filter(item => item !== evt.item);
                                 }
                             });
                     }
--- a/packages/monaco/src/browser/monaco-quick-input-service.ts
+++ b/packages/monaco/src/browser/monaco-quick-input-service.ts
@@ -261,7 +261,7 @@ export class MonacoQuickInputService implements QuickInputService {
                                     if (index !== -1) {
                                         const filteredItems = wrapped.items.slice();
                                         const removed = filteredItems.splice(index, 1);
-                                        const activeFilteredItems = wrapped.activeItems.filter(ai => ai !== removed[0]);
+                                        const activeFilteredItems = wrapped.activeItems.filter(item => item !== removed[0]);
                                         wrapped.items = filteredItems;
                                         if (activeFilteredItems) {
                                             wrapped.activeItems = activeFilteredItems;
--- a/packages/monaco/src/browser/monaco-quick-input-service.ts
+++ b/packages/monaco/src/browser/monaco-quick-input-service.ts
@@ -17,7 +17,8 @@
 
 import {
     InputBox, InputOptions, KeybindingRegistry, PickOptions,
-    QuickInputButton, QuickInputService, QuickPick, QuickPickItem, QuickPickItemButtonEvent, QuickPickItemHighlights, QuickPickOptions, QuickPickSeparator
+    QuickInputButton, QuickInputService, QuickPick, QuickPickItem,
+    QuickPickItemButtonEvent, QuickPickItemHighlights, QuickPickOptions, QuickPickSeparator
 } from '@theia/core/lib/browser';
 import { CancellationToken, Event } from '@theia/core/lib/common';
 import { injectable, inject } from '@theia/core/shared/inversify';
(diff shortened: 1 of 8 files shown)


packages/workspace/src/browser/quick-open-workspace.ts
```diff
@@ -25,9 +25,13 @@ import { FileService } from '@theia/filesystem/lib/browser/file-service';
 import { FileStat } from '@theia/filesystem/lib/common/files';
 import { Path } from '@theia/core/lib/common';
 
+interface IRecentlyOpenedPick extends QuickPickItem {
```
vince-fugnitto, line 28:
```suggestion
interface RecentlyOpenedPick extends QuickPickItem {
```

vince-fugnitto (commented):
The change reminds me that a couple of things are incorrect about the quick-input:
- the close button theming (should not be styled as the separator color (blue))
- the superfluous separators between each workspace
- we can probably ditch the "last modified" as it is not part of vscode's implementation, and since we now support removing items directly

packages/monaco/src/browser/monaco-quick-input-service.ts
```diff
@@ -251,7 +252,23 @@ export class MonacoQuickInputService implements QuickInputService {
...
+                        // https://github.com/theia-ide/vscode/blob/standalone/0.23.x/src/vs/base/parts/quickinput/browser/quickInput.ts#L1387
+                        options.onDidTriggerItemButton(
+                            {
+                                ...evt,
+                                removeItem: () => {
+                                    const index = wrapped.items.indexOf(evt.item);
+                                    if (index !== -1) {
+                                        const filteredItems = wrapped.items.slice();
+                                        const removed = filteredItems.splice(index, 1);
+                                        const activeFilteredItems = wrapped.activeItems.filter(ai => ai !== removed[0]);
```
vince-fugnitto, line 264:
```suggestion
                                        const activeFilteredItems = wrapped.activeItems.filter(item => item !== removed[0]);
```

packages/workspace/src/browser/quick-open-workspace.ts
```diff
@@ -47,9 +55,13 @@ export class QuickOpenWorkspace {
         await this.preferences.ready;
         if (!workspaces.length) {
             this.items.push({
-                label: 'No Recent Workspaces'
+                label: nls.localize('theia/windowActions/noRecentWorkSpaces', 'No Recent Workspaces')
             });
```
vince-fugnitto, line 59: We can potentially remove this item completely like in vscode if you prefer.

packages/workspace/src/browser/quick-open-workspace.ts
```diff
@@ -81,7 +93,18 @@ export class QuickOpenWorkspace {
                 },
             });
         }
-        this.quickInputService?.showQuickPick(this.items, { placeholder: 'Type the name of the workspace you want to open' });
+        this.quickInputService?.showQuickPick(this.items, {
+            placeholder: nls.localize(
+                'theia/windowActions/openRecentPlaceholder',
+                'Type the name of the workspace you want to open'),
```
vince-fugnitto, line 99: Do you think it's worth using what vscode uses here as a placeholder (which will then be localized)?

packages/workspace/src/node/default-workspace-server.ts
```diff
@@ -97,6 +97,17 @@ export class DefaultWorkspaceServer implements WorkspaceServer {
...
 
+    async removeRecentWorkspace(uri: string): Promise<void> {
+        const recentRoots = await this.getRecentWorkspaces();
+        const index = recentRoots.indexOf(uri);
+        if (index !== -1) {
+            recentRoots.splice(index, 1);
+        }
+        this.writeToUserHome({
+            recentRoots
+        });
```
vince-fugnitto, line 108: Shouldn't we only write to the user home if the `index !== -1`?

packages/workspace/src/browser/quick-open-workspace.ts
alvsan09: I agree, it's best to have all strings on this class localized as per the current solution.
Done!

packages/workspace/src/browser/quick-open-workspace.ts
alvsan09: Good suggestion, as the string would not be localized.
Done!

packages/workspace/src/node/default-workspace-server.ts
alvsan09: Done

vince-fugnitto (commented):
- [x] code looks good after the feedback
- [x] the recent workspace list is correctly displayed
- [x] removing workspaces works well, the menu remains opened similarly to vscode
- [x] performing a filter (search) and removing items works well

vince-fugnitto:
We can squash and merge tomorrow if there are no objections :)

msujew (approved):
The changes look good to me!

- [x] Clicking on the button removes the entry from the quick-input
- [x] No unnecessary separators anymore
- [x] Code is correctly localized

PR #10379, opened 2021-11-03
Improve handling of close and reload events

### LINKED ISSUE #10090: Dirty Files + Reset Workbench Layout ---> Exit Rather than Reload
### Bug Description:

The dirty file exit handler doesn't play nicely with a reload initiated by 'Reset workbench layout'

### Steps to Reproduce:

1. Open Electron application
2. Dirty some files
3. Run `Reset workbench layout`
4. See prompt about exiting with dirty files.
5. Answer Yes, you'd like to exit.
6. Observe that application exits but doesn't restart.

### Additional Information

- Operating System: RHEL7
- Theia Version: `master`

colin-grant-work:
#### What it does

- Fixes #10090 by implementing messaging between the Electron main application and Electron windows that runs instead of normal `unload` lifecycle, which allows Electron to resume the operation that was blocked (reload or close). 
- Also modifies `WindowService` to allow for safe ready checks before programmatic close or reload.

Will help with resolving issues in #10357 and will help make it possible to show a custom message in #10374.

##### In Electron
1. Dirty an editor
2. Try to close / reload / reset workbench layout.
3. See a confirmation dialog.
  - If you say you're sure you want to exit, the operation should continue and close or reload appropriately.
  - If you say you're not sure, nothing should happen.

##### In Browser
1. Dirty an editor
2. Try to close / reload
3. See no in-app dialog, but see a browser confirmation dialog.
4. Try to reset workbench layout.
5. See same behavior as in Electron.

#### Review checklist

#### Reminder for reviewers


msujew (commented):
The confirmation seems to be triggered twice in my tests, both in the browser and in electron:

Browser:

https://user-images.githubusercontent.com/4377073/140369321-cd50eda0-455f-47f0-89d1-a1659979cdb5.mp4

Electron (it's kind of difficult to see):


https://user-images.githubusercontent.com/4377073/140369758-78a134ec-7a7f-4a12-9407-9e5bf9ab2a90.mp4

colin-grant-work:
@msujew good point. In the Browser, for the reset workbench layout, there's a check ahead of time, but it's still triggering the original check as well. I'm not sure I do see it in Electron mode - were you clicking on the same dialogue twice? I don't have an explanation for why it would happen in Electron, since the original check has been removed, but I'll do some further checking.

msujew:
> I'm not sure I do see it in Electron mode - were you clicking on the same dialogue twice?

Yes, my screencaping program doesn't capture it well (as it captures on 30fps) but I clicked on the same dialogue twice, and after a frame of seeing plain Theia the same dialogue appears again. Confirming the second one actually closes the application. Perhaps it only happens on my machine. ¯\\\_(ツ)\_/¯

> For the window title bar preference, you've implemented a check that says that the application needs to be restarted, but you haven't done so for changing the configuration language, where VSCode also has a check.

Right, I didn't implement that back then because a simple window reload was enough for Theia to apply a different locale. VSCode needs to restart the whole application (including its backend to apply different locales). Although thinking about it now, we probably should include a warning in there as well, as unsaved data can be lost during the reload process. The difference to the title bar styling is that it required a whole electron window restart to apply.

> so we want to tell the user we're going to restart and give them a chance to save their work, but that makes it tricky. There are at least three options

I like the 3rd proposal the most, as it is the most intuitive and user friendliest. Though, it is of course the most complex solution as well. However, I think the first option is good as well.

colin-grant-work:
@msujew, I've pushed up changes that should remove the double prompts, adds confirmation to the language change restart, and adds the ability to tell the window service to ignore additional checks, so that one confirmation for restart for language change or window title bar style change is enough to ensure restart.

packages/core/src/browser/dialogs.ts
```diff
@@ -385,7 +385,16 @@ export class ConfirmDialog extends AbstractDialog<boolean> {
...
         return msg;
     }
+}
 
+export async function confirmExit(): Promise<boolean> {
+    const safeToExit = await new ConfirmDialog({
+        title: 'Are you sure you want to quit?',
+        msg: 'Any unsaved changes will not be saved.',
+        ok: 'Yes',
+        cancel: 'No',
```
msujew, line 395:
```suggestion
        title: nls.localize('theia/core/quitTitle', 'Are you sure you want to quit?'),
        msg: nls.localize('theia/core/quitMessage', 'Any unsaved changes will not be saved.'),
        ok: Dialog.YES,
        cancel: Dialog.NO,
```

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -1055,6 +1063,18 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
             });
     }
 
+    protected async confirmRestart(): Promise<boolean> {
+        const shouldRestart = await new ConfirmDialog({
+            title: nls.localize('vscode/localizationsActions/relaunchDisplayLanguageMessage', 'A restart is required for the change in display language to take effect.'),
+            msg: nls.localize(
+                'vscode/localizationsActions/relaunchDisplayLanguageDetail', 'Press the restart button to restart {0} and change the display language.', FrontendApplicationConfigProvider.get().applicationName
+            ),
+            ok: nls.localize('vscode/localizationsActions/restart', 'Restart'),
```
msujew, line 1072:
```suggestion
            title: nls.localizeByDefault('A restart is required for the change in display language to take effect.'),
            msg: nls.localizeByDefault('Press the restart button to restart {0} and change the display language.', FrontendApplicationConfigProvider.get().applicationName),
            ok: nls.localizeByDefault('Restart'),
```

msujew (commented):
The changes look good to me:

- [x] Reloading the browser window with dirty editors prompts the dialog once
- [x] Changing the display language (with and without dirty edits) prompts the dialog once
- [x] Changing the titleBarStyle prompts the dialog once
- [x] The dialog behaves as expected on browser/electron apps.

I have some nitpicks in regards to localization, but everything else looks good.

colin-grant-work:
@msujew, thanks for the comments. I've made the changes you suggested.

@vince-fugnitto, I converted your terminal code to use the new system: in Electron, it should only prevent exit if some terminal widget reports that it _does_ have a child process. Browser should behave the same way as before.

colin-grant-work:
@vince-fugnitto (forgot to implement the dialog for terminals; one moment)

colin-grant-work:
@msujew, when creating the terminal dialog, I lifted the strings straight from VSCode, but I got these logs:

image

I suppose that means that they aren't in the set of translated strings / weren't in it at the version on which our translation set is based? In that case, what's the ideal procedure, since they likely _will_ be in the set, eventually?

msujew:
@colin-grant-work Hm, difficult one. As they aren't in the current set of translations, they won't be translated with the 1.53.2 versions of the vscode language packs (only newer ones). I would argue that we should use the usual `nls.localize` function for these cases, but add a *todo* comment (?) that these should be replaced once the appropriate version has been reached. The eslint rule should take care of this automatically.

packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -205,11 +206,38 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
...
+    }
+
+    protected async confirmExitWithActiveTerminals(activeTerminalCount: number): Promise<boolean> {
+        const msg = activeTerminalCount === 1
+            ? nls.localizeByDefault('Do you want to terminate the active terminal session?')
+            : nls.localizeByDefault('Do you want to terminate the {0} active terminal sessions?', activeTerminalCount);
+        const safeToExit = await new ConfirmDialog({
+            title: '',
+            msg,
+            ok: nls.localizeByDefault('Terminate'),
```
msujew, line 237: I mentioned earlier how I would like to see this addressed. Afterwards, I think this would be good to go 👍 

```suggestion
            ? nls.localize('theia/terminal/terminateActive', 'Do you want to terminate the active terminal session?')
            : nls.localize('theia/terminal/terminateActiveMultiple', 'Do you want to terminate the {0} active terminal sessions?', activeTerminalCount);
        const safeToExit = await new ConfirmDialog({
            title: '',
            msg,
            ok: nls.localize('theia/terminal/terminate', 'Terminate'),
```

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -55,6 +55,10 @@ import { FormatType } from './saveable';
 import { QuickInputService, QuickPick, QuickPickItem } from './quick-input';
 import { AsyncLocalizationProvider } from '../common/i18n/localization';
 import { nls } from '../common/nls';
+import { confirmExit } from './dialogs';
+import { WindowService } from './window/window-service';
+import { ConfirmDialog, Dialog } from '.';
```
msujew, line 60: Minor: We shouldn't import from the index.

msujew (approved):
Great, looks good to me!

- [x] Editing a file without saving and closing the browser tab/electron window will display an appropriate message.
- [x] Changing the display language or the title bar style will also show a message, which shadows the "unsaved files" message.
- [x] The different values of `terminal.integrated.confirmOnExit` behave as expected and work well together with the other exiting criteria.

packages/core/src/browser/common-frontend-contribution.ts
colin-grant-work: Thanks, good catch. Between `src` imports and `.` imports, auto-import functionality is almost more trouble than it's worth. 🤦‍♂️

colin-grant-work:
@tsmaeder, @JonasHelming, before we merge this, I just want to check with y'all whether you're OK with the changes to the `WindowService` interface and behavior. Basically the changes are these:
 - remove `canUnload(): boolean`- it's replaced by `isSafeToShutDown(): Promise<boolean>` to allow asynchronous handling in Electron.
 - add `isSafeToShutDown() - it replaces `canUnload()`, because it seemed safer to change the name so that code that expected a `boolean` return didn't keep running on `Promise<boolean>` (always truthy).
 - add `setSafeToShutDown()` - handles user-triggered shutdowns that have their own confirmation systems so that we avoid *also* triggering the default shutdown handlers.
 - add `reload()` - so that Electron and browser `WindowService`s can handle reloading differently.

tsmaeder:
@colin-grant-work If I understand correctly, the breaking change is not really necessary in order to implement the fix to the missing reload on electron. As a general rule, I think it makes sense to separate bugfixes from new features in PR's. I don't think it's such a big deal in this case, though.

What I don't really understand is in what scenarios this is a breaking change for clients: I don't expect too many folks to call "canUnload", so the breakage would mostly for folks overriding `WindowService`, right?

colin-grant-work:
> @colin-grant-work If I understand correctly, the breaking change is not really necessary in order to implement the fix to the missing reload on electron. As a general rule, I think it makes sense to separate bugfixes from new features in PR's. I don't think it's such a big deal in this case, though.

If it sets your mind at ease, consider this an implementation of the feature of allowing variable (per contribution) and asynchronous handling of application shutdown. It only just happens to fix the reload bug because that was a symptom of the same underlying deficiency previously preventing us from doing asynchronous shutdown.

> What I don't really understand is in what scenarios this is a breaking change for clients: I don't expect too many folks to call "canUnload", so the breakage would mostly for folks overriding `WindowService`, right?

Yes, the changes to the `WindowService` interface only affect adopters who override or otherwise interact with the `WindowService`.

tsmaeder:
lgtm, on a high level :-)

PR #10382, opened 2021-11-04
Use `QuickPick` for opening/saving files in the browser environment

msujew:
#### What it does

Aligns Theia's behavior when opening files to the browser version of vscode.

Introduces the `QuickPickResourceService` which is the new default implementation for the `FileDialogService`. It uses monaco's `QuickPickService` to provide a new user interface for picking files and folders. One caveat of the new UI is that it does not allow to select multiple files/folders (vscode has the same limitation).

This only influences the browser environment, using the `FileDialogService` in electron still uses the native file dialogs. The old browser file dialog is still available trough the `legacy` methods in the `FileDialogService`.

Additionally adds a few properties to the `QuickPick` that were previously missing. Also fixes a selection related bug in the `QuickPick` (programmatically selecting an element was previously not possible).

#### How to test

1. Start the browser version of Theia.
2. Run the `Open` command and assert that you can open files and folders..
3. Run the `Save As` command and assert that you can save files to new locations.
4. Run the electron version and assert that the native file dialogs work as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/quick-input/quick-input-service.ts
+++ b/packages/core/src/browser/quick-input/quick-input-service.ts
@@ -110,13 +110,24 @@ export interface InputBox extends QuickInput {
 
 export interface QuickPick<T extends QuickPickItem> extends QuickInput {
     value: string;
+    valueSelection: readonly [number, number] | undefined;
     placeholder: string | undefined;
+    validationMessage: string | undefined;
     items: ReadonlyArray<T | QuickPickSeparator>;
     activeItems: ReadonlyArray<T>;
     selectedItems: ReadonlyArray<T>;
     canSelectMany: boolean;
     matchOnDescription: boolean;
     matchOnDetail: boolean;
+    matchOnLabel: boolean;
+    sortByLabel: boolean;
+    autoFocusOnList: boolean;
+    ok: boolean | 'default';
+    readonly onDidCustom: Event<void>;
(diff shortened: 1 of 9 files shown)


vince-fugnitto (commented):
@msujew for my own understanding, is there a strong reason for performing this change?
I think it's one aspect we probably have a nicer implementation than what vscode provides in the browser, and might be an overuse of the quick-pick on their part. Our browser file-dialog more closely resembles the native one and would be more familiar and easier for downstream to extend.

msujew:
@vince-fugnitto @colin-grant-work Yeah, that makes sense.

I am fine with closing this PR, just a quick question beforehand: Would it the interesting to have the `QuickPickResourceService` included in general in Theia (just not as the default implementation for opening/saving files)? I'd be happy to revert the changed behavior in the `FileDialogService`.

msujew:
@colin-grant-work Yeah, it would probably weird to have it hanging around without real necessity. There's no real use case for me, as I only build this to align to vscode, rather than my own aesthetic feeling. Also, it'd be only more maintenance surface, so I'll close this for now. Anyone who actually wants this can probably just pick up my changes, as they're quite isolated.

Though I'll probably separate out some of my changes into a new PR, as they fix some selection related logic in the `QuickPick` class.

PR #10394, opened 2021-11-05
Improve tabbar context menu

colin-grant-work:
#### What it does

This PR improves the state of the tabbar context menu, adding a number of missing commands.

Top left, before; top right, after; right, VSCode

image

#### Review checklist
1. Run through each of the new commands see that it does what you expect
> New commands are: `Close saved` (only new implementation), `Copy Path`, `Copy Relative Path`, `Keep Open` (previews only), `Split Up/Down/Left/Right`.

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/current-widget-command-adapter.ts
+++ b/packages/core/src/browser/shell/current-widget-command-adapter.ts
@@ -18,21 +18,24 @@ import { CommandHandler } from '../../common';
 import { TabBar, Title, Widget } from '../widgets';
 import { ApplicationShell } from './application-shell';
 
-type TabBarContextMenuCommandAdapterBooleanCheck = undefined | ((event: Event) => boolean);
-type TabBarContextMenuCommandHandlerBooleanCheck = undefined | ((widget: Title<Widget> | undefined, tabbar: TabBar<Widget> | undefined, event: Event) => boolean);
+type CurrentWidgetCommandAdapterBooleanCheck = (event: Event) => boolean;
+type CurrentWidgetCommandHandlerBooleanCheck = (title: Title<Widget> | undefined, tabbar: TabBar<Widget> | undefined, event: Event) => boolean;
 
 export interface TabBarContextMenuCommandHandler extends CommandHandler {
-    execute(widget: Title<Widget> | undefined, tabbar: TabBar<Widget> | undefined, event: Event): unknown;
-    isEnabled: TabBarContextMenuCommandHandlerBooleanCheck;
-    isVisible: TabBarContextMenuCommandHandlerBooleanCheck;
-    isToggled: TabBarContextMenuCommandHandlerBooleanCheck;
+    execute(title: Title<Widget> | undefined, tabbar: TabBar<Widget> | undefined, event: Event): unknown;
+    isEnabled?: CurrentWidgetCommandHandlerBooleanCheck;
+    isVisible?: CurrentWidgetCommandHandlerBooleanCheck;
+    isToggled?: CurrentWidgetCommandHandlerBooleanCheck;
 }
 
-export class TabBarContextMenuCommandAdapter implements CommandHandler {
+/**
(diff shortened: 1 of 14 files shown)


vince-fugnitto (commented):
I believe one issue I noticed in the past with the context-menu for tabs is how they are sometimes behave incorrectly, for example with this pull-request the following occurs:

https://user-images.githubusercontent.com/40359487/140758812-35e89cc3-8d9f-4ecd-b343-93b4539ffd48.mp4

colin-grant-work:
@vince-fugnitto, thanks for pointing this out. This is a pretty messy area of the command and menu systems, or the interface between them. There are a few things that cause things like this:
 - Treating only text editors as editors (so that URI-selection related context isn't refreshed / cleared when moving to a non-text-editor widget)
 - Missing context keys (VSCode has something like `isFileSystemResource` to cover URI's that aren't exactly file references)
 - Fallbacks in the `findTab` and `findTitle` methods of the `ApplicationShell` that return the current tab or current title in order to facilitate keyboard and command palette access.

In the case you demoed, it's the first: the SIW widget isn't `navigatable`, so we don't update the `SelectionService`, and then the whole process proceeds as though you were targeting the last editor you had open, or the last tree node you'd selected in the navigator. Deciding how to fix / workaround the existing system will take some thought. Easiest would be just  adding redundant commands that use more specific checks, but that's probably not the best idea.

colin-grant-work:
@vince-fugnitto, I believe I've addressed the problems you identified, would you mind having another look?

vince-fugnitto (commented):

- [x] the new commands work correctly
- [x] issues related to https://github.com/eclipse-theia/theia/pull/10394#pullrequestreview-800183789 are now resolved, and commands are respect the current widget
- [x] closing widgets that are not active in the sidebar still works
- [x] no visible regressions

colin-grant-work:
@msujew, would you mind giving it a once over just to make sure everything's in order?

msujew (commented):
Most of the behavior looks really good 👍. I only noticed one minor issue: When splitting (using the `Split <direction>` command) repeatedly, splitting seems to happen on the focused tab, instead on the one whose context menu was opened:

https://user-images.githubusercontent.com/4377073/144324720-2c732822-7d87-4487-91f0-959f32e01c4c.mp4

I think the expected behavior would be that `Split Down` would create a vertically split editor on the left side and not on the right. That's also the behavior of vscode.

colin-grant-work:
> Most of the behavior looks really good 👍. I only noticed one minor issue: When splitting (using the `Split <direction>` command) repeatedly, splitting seems to happen on the focused tab, instead on the one whose context menu was opened:

@msujew, good catch. I've added code that makes the new widget open relative to the clicked widget rather than the default, which is the active widget.

packages/core/src/browser/shell/current-widget-command-adapter.ts
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    protected transformArguments(shell: ApplicationShell, event: Event): [Title<Widget> | undefined, TabBar<Widget> | undefined, Event] {
+        const tabBar = shell.findTabBar(event);
+        if (tabBar) {
+            const title = shell.findTitle(tabBar, event);
+            if (title) {
+                return [title, tabBar, event];
+            }
```
msujew, line 58: Nitpick: I think this can be simplified.
```suggestion
            return [title, tabBar, even];
```

msujew (approved):
Great work! I can confirm that everything works well:

- [x] Better populated tabbar context menus
- [x] The displayed menus depend on the type of tab item (e.g. editors have more available commands than others)
- [x] The commands themselves work well

packages/core/src/browser/shell/current-widget-command-adapter.ts
colin-grant-work: Fair point. ✔️

PR #10395, opened 2021-11-08
Disable http fallback after successful websocket connection

### LINKED ISSUE #10261: Automatic websocket reconnection after transient offline events is broken
#### Context

Before Theia 1.13.0, when the browser is temporarily offline (e.g. toggling Wifi on and off), Theia reconnects automatically *except* for some Chromium versions under Linux (see https://github.com/eclipse-theia/theia/issues/9298).

After Theia 1.13.0, this issue with some Chromium versions under Linux was fixed in https://github.com/eclipse-theia/theia/pull/9299 so that when the browser is temporarily offline, Theia reconnects automatically all the time.

#### Description

In Theia 1.17.0, when the browser is temporarily offline Theia doesn't reconnect properly afterwards.

This takes a slightly different form in some Chromium versions for Linux:
- in the general case, on disconnection the bottom bar becomes yellow as expected, but even though it goes back to normal as expected when internet reconnects, Theia seems to not actually have reconnected and local changes cannot be saved.
- in the case of specific Chromium versions under Linux (presumably exactly the same as those that were affected by https://github.com/eclipse-theia/theia/issues/9298), the bottom bar doesn't go back to normal and instead stays yellow forever.

In both cases, any unsaved changes at the time of the disconnection can never be saved afterwards; the only way to get back to a working state is to reload the page, which loses all unsaved changes.

This might be related to the new HTTP fallback feature: https://github.com/eclipse-theia/theia/pull/9731: at the moment of disconnection there's also a notification signaling that Theia switched to HTTP fallback mode because Websockets are unavailable (but actually it's internet as a whole that's unavailable). Maybe the HTTP fallback feature interferes with Theia reconnecting.

Also possibly the HTTP fallback notification appears too soon: clearly HTTP fallback cannot be working either when the browser is offline.

### Steps to Reproduce:

1. Create a file and make some unsaved changes in the editor
2. Disconnect internet, e.g. by manually toggling the Wifi off
3. The bottom bar becomes yellow and a HTTP fallback notification pops up
4. Reconnect internet
5. Try to save the previous changes to the file

Below are some screen videos of this with Theia 1.17.0 and Theia 1.13.0 on Firefox.

#### Theia 1.17.0 on Firefox

https://user-images.githubusercontent.com/48094549/136992071-556254cf-5fc2-4916-8a6a-b8f2f6ed680b.mp4

https://user-images.githubusercontent.com/48094549/136992094-141e31c6-5b65-48f5-8e50-008b2f7721ac.mp4

#### Theia 1.13.0 on Firefox

https://user-images.githubusercontent.com/48094549/136992278-f34ea854-5e55-49aa-98cd-880afa99bec4.mp4

### Additional Information

- Operating System: Ubuntu, NayuOS (a ChromiumOS variant)
- Theia Version: 1.17.0, I did not test with newer versions yet

msujew:
#### What it does

Closes #10261

Simply disables the fallback by setting the `allowed` property on the fallback options to `false` once a websocket connection has been established.

#### How to test

1. Start Theia (Browser) on a remote server
2. Open the Theia window in your browser
3. Disconnect from the remote server (unplug your ethernet)
4. The http fallback should not activate
5. After re-establishing network connection, Theia should re-establish its connection as well.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/messaging/ws-connection-provider.ts
+++ b/packages/core/src/browser/messaging/ws-connection-provider.ts
@@ -98,7 +98,6 @@ export class WebSocketConnectionProvider extends AbstractConnectionProvider<WebS
             this.fireSocketDidClose();
         };
         socket.onmessage = ({ data }) => {
-            this.websocketErrorCounter = 0;
             this.handleIncomingRawMessage(data);
         };
         this.socket = socket;
@@ -218,6 +217,10 @@ export class WebSocketConnectionProvider extends AbstractConnectionProvider<WebS
     }
 
     protected fireSocketDidOpen(): void {
+        // Once a websocket connection has opened, disable the http fallback
+        if (this.httpFallbackOptions?.allowed) {
+            this.httpFallbackOptions.allowed = false;
+        }
         this.onSocketDidOpenEmitter.fire(undefined);
     }
 


paul-marechal:
I feel like this HTTP fallback depends on where/how a Theia instance is deployed, and as such I'm wondering if it's a good idea to always fallback to it? Assuming a WebSocket connection is supposed to work, is it a good idea to try to use HTTP if that fails?

I'm thinking that we could have an option like `connectionType: 'single-websocket' | 'http-long-polling'` to have Theia explicitly operate in one mode or the other?

msujew:
> I feel like this HTTP fallback depends on where/how a Theia instance is deployed

@paul-marechal Well yes, but also no. It depends more on the network configuration of the user. The main use case for the http fallback is for users in heavily firewalled corporate environments where websockets are blocked. The ones responsible for the deployment don't really know, whether their users will be restricted to http only.

> I'm thinking that we could have an option like connectionType: 'single-websocket' | 'http-long-polling' to have Theia explicitly operate in one mode or the other?

I'm not really in favor of that, since the long polling option is considerably slower in a lot of cases, where a lot of data is send between the clients.

Instead, I would change the default implementation to prevent using the http fallback solution. Downstream users would have to explicitly enable the option. WDYT?

paul-marechal:
I understand now, you are right. We can keep the HTTP fallback and forget about making it configurable.

paul-marechal (approved):
Code LGTM

paul-marechal:
But this got me thinking about what we do with this WebSocket multiplexing and HTTP fallback, is there any advantage to doing it ourselves over using https://socket.io/ ?

msujew:
> is there any advantage to doing it ourselves over using https://socket.io/ ?

There isn't, I just didn't know about socket.io so far. I will create an issue for that and try integrating it into Theia when I have some time for that.

PR #10187, opened 2021-09-29
Add localization cli command

msujew:
#### What it does
Closes #9708

Uses the machine translation API of DeepL to automatically translate json files.

- [x] [CQ for DeepL usage](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23832)

#### How to test

Testing this PR requires a DeepL API key. While 500k characters/month are free for developers, even a free account requires a credit card (claiming to reduce the amount of multi-accounts that abuse the free dev-tier). If you want to review this PR but don't have access to an account already, or are hesitant to use your own credit card there, you can send me a mail (linked in my github profile) and I'll send you an API key.

1. Create a sample file that you want to translate:
```json
{
  "test": "This is a value",
  "nested": {
    "hi": "Hello",
    "deeper": {
      "why": "Why"
    }
  }
}
```
2. Run the cli:
```sh
yarn theia localize -f test.json -k <api-key> (--free-api) <language codes like de, it, fr>
```
3. Assert that the output files are correctly translated (they will be in the same directory as the source file)
4. Make use of different arguments, or play around with the `help` of the localize command
5. Change the sample file and assert that only new entries which can't be found in the already translated files are actually translated (the cli will output the amount of translated entries)

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #10397, opened 2021-11-08
Add eslint rule for localization issues

msujew:
#### What it does

Adds a new eslint rule that helps developers of Theia to correctly localize the application:

1. Flags calls to `nls.localizeByDefault` which use a default value which does not exist inside of the `nls.metadata.json` file. Once such a call is detected, the rule tries to find the best fitting existing default value using the levenshtein distance.
2. Flags calls to `nls.localize` which could be replaced by `nls.localizeByDefault`.

#### How to test

1. Assert that `yarn lint` does not yield any errors (CI is green)
2. Changing the default value of an existing `localizeByDefault` call slightly should lead to an eslint error and a quickfix is provided.
3. Changing an existing `localizeByDefault(default)` call to `localize(key, default)` should lead to an eslint error and a quickfix is provided here as well.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/private-eslint-plugin/rules/localization-check.js
+++ b/dev-packages/private-eslint-plugin/rules/localization-check.js
@@ -1,0 +1,109 @@
+// @ts-check
+/********************************************************************************
+ * Copyright (C) 2021 TypeFox and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+const levenshtein = require('js-levenshtein');
+
+const metadata = require('@theia/core/src/common/i18n/nls.metadata.json');
+const messages = new Set(Object.values(metadata.messages)
+    .reduceRight((prev, curr) => prev.concat(curr), [])
+    .map(e => e.replace(/&&/g, '')));
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+    meta: {
+        type: 'problem',
+        fixable: 'code',
+        docs: {
+            description: 'prevent incorrect use of \'nls.localize\'.',
+        },
+    },
+    create(context) {
+        return {
+            CallExpression(node) {
+                const callee = node.callee;
+                if (callee.type === 'Super') {
+                    return;
+                }
+                const { value, byDefault, node: localizeNode } = evaluateLocalize(node);
+                if (value !== undefined) {
+                    if (byDefault && !messages.has(value)) {
+                        let lowestDistance = Number.MAX_VALUE;
+                        let lowestMessage = '';
+                        for (const message of messages) {
+                            const distance = levenshtein(value, message);
+                            if (distance < lowestDistance) {
+                                lowestDistance = distance;
+                                lowestMessage = message;
+                            }
+                        }
+                        if (lowestMessage) {
+                            context.report({
+                                node: localizeNode,
+                                message: `'${value}' is not a valid default value. Did you mean '${lowestMessage}'?`,
+                                fix: function (fixer) {
+                                    const updatedCall = `'${lowestMessage.replace(/'/g, "\\'")}'`;
+                                    return fixer.replaceText(localizeNode, updatedCall);
+                                }
+                            });
+                        } else {
+                            context.report({
+                                node: localizeNode,
+                                message: `'${value}' is not a valid default value.`
+                            });
+                        }
+                    } else if (!byDefault && messages.has(value)) {
+                        context.report({
+                            node,
+                            message: `'${value}' can be translated using the 'nls.localizeByDefault' function.`,
+                            fix: function (fixer) {
+                                const code = context.getSourceCode();
+                                const args = node.arguments.slice(1);
+                                const argsCode = args.map(e => code.getText(e)).join(', ');
+                                const updatedCall = `nls.localizeByDefault(${argsCode})`;
+                                return fixer.replaceText(node, updatedCall);
+                            }
+                        });
+                    }
+                }
+            }
+        };
+        function evaluateLocalize(/** @type {import('estree').CallExpression} */ node) {
+            const callee = node.callee;
+            if ('object' in callee && 'name' in callee.object && 'property' in callee && 'name' in callee.property && callee.object.name === 'nls') {
+                if (callee.property.name === 'localize') {
+                    const defaultTextNode = node.arguments[1]; // The default text node is the second argument for `nls.localize`
+                    if (defaultTextNode && defaultTextNode.type === 'Literal' && typeof defaultTextNode.value === 'string') {
+                        return {
+                            value: defaultTextNode.value,
+                            byDefault: false
+                        };
+                    }
+                } else if (callee.property.name === 'localizeByDefault') {
+                    const defaultTextNode = node.arguments[0]; // The default text node is the first argument for ``nls.localizeByDefault`
+                    if (defaultTextNode && defaultTextNode.type === 'Literal' && typeof defaultTextNode.value === 'string') {
+                        return {
+                            node: defaultTextNode,
+                            value: defaultTextNode.value,
+                            byDefault: true
+                        };
+                    }
+                }
+            }
+            return {};
+        }
+    }
+};
--- a/packages/debug/src/node/vscode/vscode-debug-adapter-contribution.ts
+++ b/packages/debug/src/node/vscode/vscode-debug-adapter-contribution.ts
@@ -14,6 +14,8 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
+/* eslint-disable @theia/localization-check */
+
 import * as fs from '@theia/core/shared/fs-extra';
 import * as path from 'path';
 import { DebugAdapterExecutable, DebugAdapterContribution } from '../debug-model';
--- a/packages/git/src/common/git-model.ts
+++ b/packages/git/src/common/git-model.ts
@@ -108,6 +108,7 @@ export namespace GitFileStatus {
             case GitFileStatus.New: return !!staged ? nls.localize('theia/git/added', 'Added') : nls.localize('theia/git/unstaged', 'Unstaged');
             case GitFileStatus.Renamed: return nls.localize('theia/git/renamed', 'Renamed');
             case GitFileStatus.Copied: return nls.localize('theia/git/copied', 'Copied');
+            // eslint-disable-next-line @theia/localization-check
             case GitFileStatus.Modified: return nls.localize('vscode.git/repository/modified', 'Modified');
             case GitFileStatus.Deleted: return nls.localize('vscode.git/repository/deleted', 'Deleted');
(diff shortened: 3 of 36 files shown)


dev-packages/eslint-plugin/rules/localization-check.js
```diff
@@ -0,0 +1,110 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+const path = require('path');
```
vince-fugnitto, line 18: I believe `path` is unused.

vince-fugnitto (commented):
The changes overall look good to me, I'll let @colin-grant-work review as well and take a deeper look into the code:

- [x] warns when `localizeByDefault` is used incorrectly (no value exists)
- [x] warns when there is a typo, for instance a string does not exist but is close
- [x] warns when `localize` is used when `localizeByDefault` can be used

While potentially difficult, will there be plans to have user facing strings checked if they can also be localized, potentially in the future?

msujew:
> While potentially difficult, will there be plans to have user facing strings checked if they can also be localized, potentially in the future?

@vince-fugnitto No plans so far, due to the difficulty. Perhaps there's some good idea on how to accomplish that, but I have no idea how one would do that without flagging a ton of false positives. I would be glad to contribute that, if we come up with a good solution for that.

vince-fugnitto:
> @vince-fugnitto No plans so far, due to the difficulty. Perhaps there's some good idea on how to accomplish that, but I have no idea how one would do that without flagging a ton of false positives. I would be glad to contribute that, if we come up with a good solution for that.

@msujew I don't really have a good solution other than creating "hint" or "info" markers in the frontend for all strings as reminders for devs if they want to/should localize. The issue is even as a hint devs might find it is too verbose or distracting.

colin-grant-work:
One question that this raises for me is how to handle new strings that don't appear in VSCode. There are a few places where we differ in irreconcilable ways, and I've seen some examples of `.localize` with a key containing `theia` rather than VSCode, and you've mentioned machine translating new strings in the past. Is the idea that we should then modify the `nls.metadata.json` file directly, or how would those be managed? Basically: how do you fix this lint error if you're introducing a genuinely new string?

msujew:
> One question that this raises for me is how to handle new strings that don't appear in VSCode.

@colin-grant-work I think I went over most of this during this weeks dev-meeting. New strings will be extracted using the `theia nls-extract` cli and translated using https://github.com/eclipse-theia/theia/pull/10187. I'm currently thinking that these commands will run just before the release and any mis-translations will be fixed during the month after the release (probably with the help/issues raised by the community).

> Basically: how do you fix this lint error if you're introducing a genuinely new string?

Genuinely new strings should just use the `nls.localize` function and introduce a new key. The proposed eslint rule just catches any issues that arise due to incorrect use of `nls.localizeByDefault` and helps users not introduce values that already exist inside of the `nls.metadata.json` file.

packages/debug/src/node/vscode/vscode-debug-adapter-contribution.ts
```diff
@@ -14,6 +14,8 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
+/* eslint-disable @theia/localization-check */
```
colin-grant-work, line 17: Perhaps add a second commit fixing these? it looks like it's mainly suggestions to use `byDefault` in place of the keyed version. For that violation, it might make sense to add a fixer to the plugin.

dev-packages/eslint-plugin/rules/localization-check.js
```diff
@@ -0,0 +1,109 @@
...
+        fixable: 'code',
+        docs: {
+            description: 'prevent incorrect use of \'nls.localize\'.',
+        },
+    },
+    create(context) {
+        const metadata = require('../data/nls.metadata.json');
+        const messages = new Set(Object.values(metadata.messages)
+            .reduceRight((prev, curr) => prev.concat(curr), [])
+            .map(e => e.replace(/&&/g, '')));
```
colin-grant-work, line 33: Could this code be moved out of the `create` function? It looks like that function is run quite frequently (once per file, perhaps?), so it might be better to do this work somewhere it can be globally accessible.

packages/git/src/common/git-model.ts
```diff
@@ -109,6 +109,7 @@ export namespace GitFileStatus {
             case GitFileStatus.New: return !!staged ? nls.localize('theia/git/added', 'Added') : nls.localize('theia/git/unstaged', 'Unstaged');
             case GitFileStatus.Renamed: return nls.localize('theia/git/renamed', 'Renamed');
             case GitFileStatus.Copied: return nls.localize('theia/git/copied', 'Copied');
+            // eslint-disable-next-line @theia/localization-check
```
colin-grant-work, line 112: Is this the solution for things that _are_ VSCode but _aren't_ in our JSON (yet)?

dev-packages/eslint-plugin/package.json
```diff
@@ -4,6 +4,13 @@
...
   "dependencies": {
-    "@theia/core": "1.19.0"
+    "@theia/core": "1.19.0",
+    "js-levenshtein": "^1.1.6"
+  },
+  "devDependencies": {
+    "copyfiles": "^2.4.1"
+  },
+  "scripts": {
+    "prepare": "copyfiles ../../packages/core/src/common/i18n/nls.metadata.json ./data -f"
```
colin-grant-work, line 14: Why is this necessary?

packages/git/src/common/git-model.ts
msujew: Well, in this particular case the `Modified` value seems to exist somewhere in the `nls.metadata.json` (our version). However, I worry about semantics here, since the `Modified` that exists inside of vscode is used for modified preferences. I assume that this particular context does not translate correctly into other language when used for git related modifications.

Also you perhaps have noticed the key to `vscode.git`, which is the builtin extension, that is also translated in the language pack. As this isn't localized text from the actual vscode codebase, it doesn't appear inside of the `nls.metadata.json`. Now that we use `localizeByDefault` basically everywhere, I would be fine with using `theia/*` keys for these instead of relying on the extension translations. There are only around ~20 instances of these cases in Theia. WDYT?

dev-packages/eslint-plugin/package.json
msujew: Actually, I don't think that this is necessary. I thought that we would also release the `eslint-plugin` to npm. For this case we would need to copy the json file. However, that doesn't seem to be the case, so we can rely on the exact path of the json file in the repo.

packages/debug/src/node/vscode/vscode-debug-adapter-contribution.ts
msujew: Well, this file is a weird case anyway. This file uses it's own "implementation" (if one can say so), since it's in the `node` directory and has no actual localizations available:

https://github.com/eclipse-theia/theia/blob/89b8dd1ab7daa5dbead588ae30a02fc8f15ad595/packages/debug/src/node/vscode/vscode-debug-adapter-contribution.ts#L27-L31

This code has been there for quite a long time, or at least way before I started working on i18n in Theia. Actually, looking further into it, nothing that's exported in this file is referenced somewhere else. It seems to be dead code from before the vscode-extension time of Theia. I would remove this in a separate PR.

packages/git/src/common/git-model.ts
colin-grant-work: I guess my target would be a one-tiered world: everything uses `localizeByDefault`. If we find that the thing translated doesn't exist in the JSON, there are two possible causes: it's a typo / rephrase, or it hasn't been translated by language packs. In the former case, we should correct it, because then it will be translated correctly. In the latter case, rewriting `localizeByDefault` to `localize('some/theia/key')` doesn't actually get us any closer to getting the thing translated, so I don't see the point. Instead, whenever we are able to create new translations ourselves, we can traverse the code base, find any uses of `localizeByDefault` that aren't currently working, translate them, and add it to the JSON (creating an automatic key, if that's necessary to make things work, something like `theia/<package>/<file>/<first word or two as necessary>`). If we want a system that requires investigating the actual origin of a particular string to determine its key and checking the JSON to see if that source is covered, then we'll need a lot more documentation about when and how to do that.

msujew: Interesting approach. I like the simplicity (for the user, hidden behind a bit of technical complexity). I'll create a draft for this, and add it as an additional commit to this PR.

msujew:
@colin-grant-work It took me a while to get back to this, since I was busy with other stuff and waiting for #10187 to get merged. So, I think having a one-tiered approach to this (only using `localizeByDefault` in Theia) will lead to too much implicit behavior. I can imagine multiple occasions where we run into the following issues:

1. Using `localizeByDefault` everywhere makes this eslint rule obsolete. As a consequence, developers will be uncertain whether they actually use a "default" translation (using vscode) or introduce a new translation.
2. Given that translation ids will be generated using this approach from the text, minor changes in the original English text might lead to stability issues down the line, i.e. English text changes -> id changes -> translations for other languages are suddenly missing.

Therefore, I believe the two-tiered approach makes it clear which method to use, especially when the eslint rule shows errors on misused `localizeByDefault` and `localize` calls. Does my line of thinking make sense here? In short: I like the explicit nature of `localize(key, default)` combined with the safety of the eslint rule more than implicit `localizeByDefault` behavior which might break translations on minor changes.

Do you mind looking into my rebased changes? :)

colin-grant-work:
> @colin-grant-work It took me a while to get back to this, since I was busy with other stuff and waiting for #10187 to get merged. So, I think having a one-tiered approach to this (only using `localizeByDefault` in Theia) will lead to too much implicit behavior...

@msujew, now that there are real translations of the non-VSCode strings in Theia (or at least the potential to create them, and an associated procedure), I think it makes sense to distinguish between the two systems. I'll take a look at the code again in the next day or two.

dev-packages/private-eslint-plugin/rules/localization-check.js
```diff
@@ -0,0 +1,109 @@
...
+                    }
+                } else if (callee.property.name === 'localizeByDefault') {
+                    const firstArgument = node.arguments[0];
+                    if (firstArgument && firstArgument.type === 'Literal' && typeof firstArgument.value === 'string') {
+                        return {
+                            node: firstArgument,
+                            value: firstArgument.value,
+                            byDefault: true
+                        };
+                    }
```
colin-grant-work, line 103: Same question re: `firstArgument` :-).

colin-grant-work (approved):
The behavior looks correct to me, and the check is quite quick.
 - Detected bad use of `byDefault` and offered a suggestion.
 - Automatically inserted the suggestion when triggered to do so.
 - Detected use of `.localize` when `byDefault` was possible.
 - Correctly rewrote call to `byDefault` when triggered.

Minor comment re: variable names. It also doesn't trigger on `nls.localizeByDefault('')` (empty string). I'm not sure whether that's because the empty string is in the `metadata.json`, and I'm not sure whether it's desirable to trigger in that case, but it does seem like an odd input.

dev-packages/private-eslint-plugin/rules/localization-check.js
```diff
@@ -0,0 +1,109 @@
...
+        };
+        function evaluateLocalize(/** @type {import('estree').CallExpression} */ node) {
+            const callee = node.callee;
+            if ('object' in callee && 'name' in callee.object && 'property' in callee && 'name' in callee.property && callee.object.name === 'nls') {
+                if (callee.property.name === 'localize') {
+                    const secondArgument = node.arguments[1];
+                    if (secondArgument && secondArgument.type === 'Literal' && typeof secondArgument.value === 'string') {
+                        return {
+                            value: secondArgument.value,
+                            byDefault: false
```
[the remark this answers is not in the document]
msujew: Yes, there is ;)

PR #10407, opened 2021-11-10
Add start-up timing logs in frontend and backend

### LINKED ISSUE #10406: Refine start-up performance logs
### Feature Description:

An earlier contribution, PR #10116, added log messages with timings of frontend application loading, especially as it pertains to initialization of plugins. It would be nice to do similarly in the backend, to add log messages showing the timing of start-up tasks on that end that are subject to potential performance problems. As much as possible, the look-and-feel of the timing log API and messages on the frontend and backend should be aligned, preferably a single API that has different implementations on each end if appropriate.

cdamus:
#### What it does

Refactors various functions in a few different places in the frontend into a new abstract `Stopwatch` class that provides measurement of tasks and logs their completion with messages similar to what we had before. The frontend injects a concrete implementation based on the browser performance API used previously and the backend injects a concrete implementation based on the similar, but different, node.js performance API.

Additionally, as the performance of the frontend and backend applications is largely independent but also coupled at a few synchronization points, a `BackendStopwatch` RPC service is added that publishes limited access to the backend’s stopwatch, letting the frontend log timing messages in the timeframe of the backend application. This is used to log how long it took the frontend to fully start up from the backend’s perspective, which is especially useful in the Electron deployment but also makes sense in most Browser use cases.

Resolves #10406.

No new package dependencies are added.

#### How to test

Start the Electron or Browser example and see the timing log messages from both the frontend and the backend emitted on the console.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (commented):

cdamus:
> Thank you for your contribution 👍
> 
> In order to accept your changes please be sure to sure to sign the [eclipse contributor agreement (eca)](https://github.com/eclipse-theia/theia/blob/master/CONTRIBUTING.md#eclipse-contributor-agreement) with the same email as your authorship.

Thanks, @vince-fugnitto ! I had already signed the ECA (being a committer on other projects) but used a different e-mail address as commit author than what is registered at Eclipse. I have fixed that.

msujew (commented):
The approach already looks quite good. I have a few suggestions, regarding APIs and optionality.

packages/core/src/browser/frontend-application.ts
```diff
@@ -121,6 +121,13 @@ export class FrontendApplication {
     @inject(TooltipService)
     protected readonly tooltipService: TooltipService;
 
+    @inject(Stopwatch)
+    protected readonly stopwatch: Stopwatch;
+
+    @optional()
+    @inject(BackendStopwatch)
+    protected readonly backendStopwatch?: BackendStopwatch;
```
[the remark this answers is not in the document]
cdamus: I'll look for an example elsewhere in the codebase to mimic.

packages/core/src/common/performance/measurement-protocol.ts
```diff
@@ -0,0 +1,87 @@
...
+@injectable()
+export class DefaultBackendStopwatch {
+
+    readonly measurements = new Map<number, Measurement>();
+
+    protected idSequence: number = 0;
+
+    constructor(@inject(new LazyServiceIdentifer(() => Stopwatch)) protected readonly stopwatch: Stopwatch) {
+        // Nothing else to initialize
+    }
```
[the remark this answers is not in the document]
cdamus: Good point!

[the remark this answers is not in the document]
cdamus: Actually, I had originally tried field injection, but the `LazyServiceIdentifer` [sic] for function-based injection to break dependency cycles doesn't work for property injection[^1]. It has to be used in a class constructor. And this case is a dependency cycle, so I'm not sure what else I can do. I could try `@lazyInject` but that requires adding a new package dependency that we haven't used to date in Theia.

[^1]: https://github.com/inversify/InversifyJS/issues/944

cdamus:
Thanks for the review, @msujew! I've pushed commit 28c7094955f088b4be864fe8f0548dc7b460273c that rebases to resolve merge conflicts and addresses some of your comments:

- the back-end stopwatch RPC service is now optional via injectable options. The only option so far is `disabled`. If disabled, an empty stub is supplied instead of the default implementation (on either frontend or backend or both). I tested this _ad hoc_ with option injection on the frontend and the backend separately to confirm that the setup doesn't break and no logs actually happen
- private methods are made protected in the `Stopwatch` class
- I've added an explanatory comment to the constructor injection in the `DefaultBackendStopwatch` class. As I replied in-line above, constructor injection is required with the packages currently available. Better suggestions are most welcome!

packages/core/src/browser/performance/frontend-stopwatch.ts
```diff
@@ -0,0 +1,62 @@
+/********************************************************************************
+* Copyright (c) 2019, 2021 TypeFox, STMicroelectronics and others.
```
colin-grant-work, line 2: Where does the 2019 date come from here?

packages/core/src/common/performance/measurement-protocol.ts
```diff
@@ -0,0 +1,138 @@
...
+ * Injectable options for configuration of the (remote) back-end stopwatch service.
+ */
+export interface BackendStopwatchOptions {
+    /**
+     * Wether to disable the back-end stopwatch service. When `disabled`, the service interface
+     * is still fulfilled, but it just doesn't do anything. The default `disabled` state is
+     * `false` (being enabled).
+     */
+    disabled?: boolean;
+}
```
colin-grant-work, line 44: I'm not sure I see the purpose of this interface. Without it, adopters who don't want a BackendStopwatch can do something like

```
rebind(BackendStopwatch).to(NullBackendStopwatch).inSingletonScope();
```
which seems at least as straightforward as
```
bind(BackendStopwatchOptions).toConstantValue({disabled: true});
```

packages/core/src/common/performance/stopwatch.ts
```diff
@@ -0,0 +1,157 @@
...
+    protected atLevel(logOptions: LogOptions, levelOverride?: LogLevel, optionalArgs?: any[]): LogOptions {
+        return { ...logOptions, levelOverride, arguments: optionalArgs };
+    }
+
+    protected logLevel(elapsed: number, options?: Partial<LogOptions>): LogLevel {
+        if (options?.levelOverride) {
+            return options?.levelOverride;
+        }
+
+        return options?.defaultLogLevel || DEFAULT_LOG_LEVEL;
```
colin-grant-work, line 132:
```suggestion
        return options?.defaultLogLevel ?? this.defaultOptions.defaultLogLevel ?? DEFAULT_LOG_LEVEL;
```
This method is likely expected to be called on options have gone through `mergeOptions`, but it isn't guaranteed, so it's like worth inserting the local default.

packages/core/src/common/performance/stopwatch.ts
```diff
@@ -0,0 +1,157 @@
...
+        return result;
+    }
+
+    protected atLevel(logOptions: LogOptions, levelOverride?: LogLevel, optionalArgs?: any[]): LogOptions {
+        return { ...logOptions, levelOverride, arguments: optionalArgs };
+    }
+
+    protected logLevel(elapsed: number, options?: Partial<LogOptions>): LogLevel {
+        if (options?.levelOverride) {
+            return options?.levelOverride;
```
colin-grant-work, line 129:
```suggestion
            return options.levelOverride;
```

packages/core/src/node/backend-application.ts
```diff
@@ -152,12 +151,11 @@ export class BackendApplication {
...
     protected readonly processUtils: ProcessUtils;
 
-    private readonly _performanceObserver: PerformanceObserver;
-
     constructor(
         @inject(ContributionProvider) @named(BackendApplicationContribution)
         protected readonly contributionsProvider: ContributionProvider<BackendApplicationContribution>,
-        @inject(BackendApplicationCliContribution) protected readonly cliParams: BackendApplicationCliContribution
+        @inject(BackendApplicationCliContribution) protected readonly cliParams: BackendApplicationCliContribution,
+        @inject(Stopwatch) protected readonly stopwatch: Stopwatch
```
colin-grant-work, line 158: This can be added as a property injection.

packages/core/src/node/performance/node-stopwatch.ts
```diff
@@ -0,0 +1,40 @@
+/********************************************************************************
+* Copyright (c) 2019, 2021 TypeFox, STMicroelectronics and others.
```
colin-grant-work, line 2: See previous comment on  header.

packages/core/src/common/performance/measurement-protocol.ts
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * A service that exposes the back-end's {@link Stopwatch} to clients
+ * via the remote API.
+ */
+export interface BackendStopwatch extends JsonRpcServer<unknown> {
```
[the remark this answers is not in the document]
cdamus: I see no use cases for exposing those APIs, so yes I would prefer to leave off the inherited interface. Didn't know I could do that!

packages/core/src/common/performance/measurement-protocol.ts
```diff
@@ -0,0 +1,138 @@
...
+                    return () => Promise.resolve(target.start());
+                }
+                if (p === 'stop') {
+                    return () => Promise.resolve();
+                }
+            }
+        });
+
+    }
+}
```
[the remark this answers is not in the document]
cdamus: Yes, this complexity was only there to maintain the RPC protocol. Glad to drop it.

packages/core/src/browser/performance/frontend-stopwatch.ts
cdamus: Much of the content of this file was factored out of another source file that was copyrighted 2019 by TypeFox and others. So it seemed prudent to maintain that copyright statement. The same applies to the backend file `node-stopwatch.ts`.

packages/core/src/common/performance/measurement-protocol.ts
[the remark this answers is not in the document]
cdamus: > > And this case is a dependency cycle
> 
> I'm not sure what you mean here. `DefaultBackendStopwatch` depends on `Stopwatch`, and on the back end, `Stopwatch` is bound to `NodeStopwatch`, which does not depend on `DefaultBackendStopwatch` in any way, so I don't see any dependency cycle, and I was able to get this code working with simple property injection `@inject(Stopwatch) protected readonly stopwatch: Stopwatch;`. What circularity do you see or foresee here?

I got an injection failure exception complaining of unsatisfied dependency when I tried that initially. I tracked it down to a dependency cycle in debugging, which is why I resorted to the lazy service identifier. Perhaps subsequent changes have broken that cycle. I'll have another look.

packages/core/src/node/backend-application.ts
cdamus: The constructor calls `initialize`, which needs to use the stopwatch to measure the initializations of contributions. But this happens before property injection, so constructor injection seemed to be the way to go. This then would also have to be added to the breaking API changes in the changelog.

I would have moved the `initialize` step to the `@postConstruct configure` method, but I figured there must have been a reason to call it from the constructor and so I wasn't comfortable making that change.

packages/core/src/common/performance/measurement-protocol.ts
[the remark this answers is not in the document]
cdamus: Confirmed. The dependency cycle that we once had here is evidently no longer. So, that's better all around.

cdamus:
Thanks @colin-grant-work for the review. I think this has helped to clean things up considerably!

I've rebased to resolve merge conflict in the changelog and made changes to address your comments. I've not got a solution to the matter of changing the constructor of the `BackendApplication`. Suggestions what to do about that are welcome. I'm nervous about moving the `this.initialize()` call to the `configure()` method because I expect it was called by the constructor for a good reason. But as it's an async method I don't see that it could be such a problem to make that change.

Personally, I'd rather see that initialization step also happen in the post construction because it doesn't feel like an object constructor responsibility and it would mean that we can avoid the API-breaking changing to the constructor signature. WDYT?

colin-grant-work:
> Personally, I'd rather see that initialization step also happen in the post construction because it doesn't feel like an object constructor responsibility and it would mean that we can avoid the API-breaking changing to the constructor signature. WDYT?

I agree here - [in general](https://github.com/eclipse-theia/theia/wiki/Coding-Guidelines#post-construct) we want initialization to happen in decorated `postConstruct` methods. It's still likely worth mentioning as a breaking change, though, since extenders may have code in their own constructors that assumes that `super` has already run the `initialize` method, but if we have to break API - and we do, either way - I'd rather break it in a direction that moves us closer to our guidelines.

cdamus:
I've pushed an update that moves the initialization as discussed, to maintain the existing constructor signature. Testing the example deployment shows no outward change in behaviour. In particular, initialization and configuration of contributions are still interleaved as before (the `initialize` method still being async). Presumably contributions that provide both methods are already expected to handle them being invoked in either order, so this should not be a change that impacts on contributions.

Because now the constructor does not trigger initialization, any derivative applications that don't get the `BackendApplication` from the IoC container may now find that `initialize` is not called and so could break. Hence the mention in the changelog.

msujew (approved):
The changes are looking really good to me now. I have a minor remark, but except for that I don't have anything to add 👍

packages/core/src/browser/frontend-application.ts
```diff
@@ -161,13 +169,15 @@ export class FrontendApplication {
...
-        await this.initializeLayout();
+        await this.measure('initializeLayout', () => this.initializeLayout(), 'Initialize the workbench layout', false);
         this.stateService.state = 'initialized_layout';
         await this.fireOnDidInitializeLayout();
 
         await this.revealShell(host);
         this.registerEventListeners();
         this.stateService.state = 'ready';
+
+        startup.then(idToken => this.backendStopwatch.stop(idToken, 'Frontend application start', []));
```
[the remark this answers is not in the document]
cdamus: Thanks, Mark. If we wait for the back-end logging round-trip to complete here, then a caller of `start` that awaits its result would be waiting for that back-end call to finish. But a client of the `start` API would only plausibly care about it actually finishing its start-up flow, not also the logging. Does that make sense?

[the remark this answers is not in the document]
msujew: Alright, I assumed something in this direction. Feel free to ignore this 👍

colin-grant-work:
@cdamus, apologies for the delay. This is first thing on my agenda for Monday.

packages/core/src/browser/frontend-application.ts
```diff
@@ -263,10 +273,9 @@ export class FrontendApplication {
     protected revealShell(host: HTMLElement): Promise<void> {
         const startupElem = this.getStartupIndicator(host);
         if (startupElem) {
-            return new Promise(resolve => {
+            const doRevealShell = () => new Promise<void>(resolve => {
```
colin-grant-work, line 276: Since we're already adding several measurements to the `start` method, which is where this is called, it seems more consistent (and easier) to add the measurement there? I'm not sure how often we enter the `else` here - in principle, it shouldn't be often.

colin-grant-work (approved):
Looks good to me, and it's good to have a more unified style of measurement.

packages/core/src/browser/frontend-application.ts
cdamus: Indeed, this was too cute by half. Thanks for the insight about the else case!

cdamus:
> @cdamus, apologies for the delay. I'll take a look now.

No apologies needed @colin-grant-work. I'm very glad to have your keen eyes on this, and as you can tell I don't have a great deal of bandwidth for this, either!

I've pushed an update that resolves the merge conflict and simplifies the frontend application start measurement per your suggestion.

PR #10408, opened 2021-11-10
The Active LTS version of NodeJS does not work

Unifex:
Standing up the project with the latest Active LTS of NodeJS just doesn't work.  Rolling back through the Maintenance LTS versions I found v12 was the only one that worked.

#### What it does
Corrects documentation for the prerequisites

#### How to test
Create the package.json as described here: https://theia-ide.org/docs/composing_applications#consuming-vs-code-extensions
Using what ever node version manager you use set node to the latest Active LTS and run yarn.
Run `yarn`
Errors ensue and it fails very early.
Switch node to the v12 LTS ([12.22.7 ](https://nodejs.org/download/release/latest-v12.x/) as of this writing).
Run `yarn`
There's still errors, but it gets a lot further and they don't seem related to the version of NodeJS.

### DIFF
--- a/doc/Developing.md
+++ b/doc/Developing.md
@@ -53,7 +53,7 @@ For Windows instructions [click here](#building-on-windows).
 ## Prerequisites
 
  - Node.js `>= 12.14.1`.
-   - Preferably, **use** Node's [Active LTS version](https://nodejs.org/en/about/releases/).
+   - Preferably, **use** Node's [v12 Maintenance LTS version](https://nodejs.org/en/about/releases/).
    - If you are interested in Theia's VS Code Extension support then you should use a Node version at least compatible with the one included in the version of Electron used by [VS Code](https://github.com/microsoft/vscode).
  - [Yarn package manager](https://yarnpkg.com/en/docs/install)  `>= 1.7.0` **AND** `< 2.x.x`.
  - git (If you would like to use the Git-extension too, you will need to have git version 2.11.0 or higher.)


msujew:
Hi @Unifex, using newer versions of Node is indeed possible (our CI for example is using Node v12+v14). However, due to a capped `node` engines version in one of our dependencies, yarn will warn you about using an unsupported version of node. You can simply run yarn with the `--ignore-engines` flag and successfully run Theia under newer versions of node.

vince-fugnitto (commented):
I believe this pull-request can be closed, there should not be a need to recommend node `v12`:
- [x] successfully built with node `v16` (using `yarn --ignore-engines` for `dugite`)
- [x] node `v12` and `v14` is tested during CI:

https://github.com/eclipse-theia/theia/blob/bb5a4366f5abac1e44df53d2ad961027d2ae0316/.github/workflows/build.yml#L23

paul-marechal (approved):
I think this change is for the best. Yes Theia itself works on Node 16 but people are getting confused because of `dugite-extra` having the annoying range.

I think there is an issue building `@theia/node-pty` on Mac with Node 16, we have a fix but didn't release yet as we are mostly waiting for the Electron upgrade to finally finally fully support it without quirks.

Sorry for the inconvenience. I'll update the doc to mention Node's LTS again once we get around the Electron upgrade.

paul-marechal:
Ah you don't have an ECA. I'll open another PR to merge quickly.

paul-marechal:
https://github.com/eclipse-theia/theia/pull/10411

PR #10413, opened 2021-11-12
search-in-workspace: allow search to follow symlinks

### LINKED ISSUE #10390: Support preference for Search to follow symbolic links
### Feature Description:
This was originally discussed at:
https://community.theia-ide.org/t/follow-symlinks-when-searching/1668/2

Vscode supports this preference as shown below:
image

This will track a similar implementation in theia.

The search feature uses  `ripgrep` to delegate the search and the command has a switch to follow symbolic links:

>~/git/theia6/node_modules/vscode-ripgrep/bin$ rg --help |grep follow
            will be provided. Namely, the following is equivalent to the above:
            For example, the following command will change the match color to magenta and
            For example, the following command will change the match background color to
    -L, --follow                            
            When this flag is enabled, ripgrep will follow symbolic links while traversing
            This flag can be disabled with --no-follow.
            Whenever a file path is printed, follow it with a NUL byte. This includes
            preprocessing. For example, given the following shell script, 'pre-pdftotext':
            will be provided. Namely, the following is equivalent to the above:
            present, ripgrep will print the following stats to stdout at the end of the

So the preference would be read and if activated if will add this switch as part of the command arguments

Archie27376:
#### What it does
Fixes #10390
This commit provides the user to enable a preference to follow symlinks while searching using the `Follow Symlinks` preference, similar to that of vscode. 

#### How to test
- Have a symbolic linked file within a folder
- Open either the electron app or browser and select the folder with the symlinked file as a workspace
- Access `Preferences` and make sure that `Follow Symlinks` is enabled
- Search text within the file in the search widget and make sure results are given

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/search-in-workspace/src/browser/search-in-workspace-preferences.ts
+++ b/packages/search-in-workspace/src/browser/search-in-workspace-preferences.ts
@@ -53,6 +53,11 @@ export const searchInWorkspacePreferencesSchema: PreferenceSchema = {
             description: nls.localizeByDefault('Search case-insensitively if the pattern is all lowercase, otherwise, search case-sensitively.'),
             default: false,
             type: 'boolean',
+        },
+        'search.followSymlinks': {
+            description: nls.localizeByDefault('Controls whether to follow symlinks while searching.'),
+            default: true,
+            type: 'boolean',
         }
     }
 };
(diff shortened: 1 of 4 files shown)


vince-fugnitto (approved):
I confirmed that the behavior works as expected with a workspace which contains symlinks:

_preference is on_:

image


_preference is off_:

image

msujew (approved):
The changes look good to me as well:

- [x] With the preference enabled, text in symlinked files are correctly displayed
- [x] With the preference disabled, symlinks are ignored

alvsan09 (approved):
I have tested this PR and it works as expected :+1: 

The code looks good as well !!

alicezhangyuan:
there is another issue, for quick-file-open(ctrl+p), it is not available for synlink files. @Archie27376

Issue #9676, opened 2021-06-30
vsx-registry: display incompatible extensions as disabled


vince-fugnitto:
#### Feature Description:

Rather than completely hide incompatible extensions (non api-compatible) in the `extensions-view`, we can display them as disabled. At the moment there may be confusion for end users when they see extensions available from [open-vsx](https://open-vsx.org/) and not in the application.

We can mark them as disabled in the tree:
- visually disabled (such as greyed out),
- hide the `install` functionality from them.

Related conversation: https://github.com/eclipse-theia/theia-blueprint/issues/96

Jeeppler:
Yes, I 100% agree.

vince-fugnitto:
In VS Code they introduced the idea of [deprecated extensions](https://code.visualstudio.com/updates/v1_68#_deprecated-extensions) which would clash with this approach (marking them as disabled). If anyone is interested in still rendering extensions which are incompatible they might need to find a new way to visually mark them as such.

PR #9330, opened 2021-04-12
VSX: Add 'Install Another Version...' Command

seantan22:
#### What it does

- Supports the ability to install any compatible version of an user-installed extension, provided that the extension is available in the Open VSX Registry.

https://user-images.githubusercontent.com/46289281/114419822-c3067980-9b81-11eb-9155-6556d10b7919.mp4

#### How to test

1. Open the extensions view and locate the `Installed` view of all user-installed extensions.
2. Pick an extension --> Click the `manage` gear icon --> Click `Install Another Version...` command. Note that the command is disabled for any extension not available in the Open VSX Registry.
3. Select a compatible version to install from the Quick Pick dropdown.
4. Observe that the previous version is uninstalled and replaced with the newly selected version.

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

Signed-off-by: seantan22 <sean.a.tan@ericsson.com>

PR #10424, opened 2021-11-16
vsx-registry: display incompatible extensions

vince-fugnitto:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/9676.

The commit adds the functionality to display incompatible extensions when performing a search. The changes will now display incompatible extensions but mark them as disabled, and non-installable. The behavior is also controllable by a new preference so users can decide if they want the potential extra noise when searching.

_Example: using an api-version of 1.30.0_:

incompatible-plugins

#### How to test

1. start the application
2. open the extensions-view
3. perform a search - if an extension has no compatible version then it should be marked as disabled and non-installable
  3.1. hovering over the disabled extension should correctly display the tooltip
  3.2 the plugin should not resolve, the readme won't be displayed
4. updating the preference `'extensions.displayIncompatible'` and re-searching should not yield incompatible results
5. repeat the steps by starting with `yarn start --vscode-api-version=1.30.0` for additional incompatible results

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/vsx-registry/src/browser/vsx-extension.tsx
+++ b/packages/vsx-registry/src/browser/vsx-extension.tsx
@@ -144,6 +144,11 @@ export class VSXExtension implements VSXExtensionData, TreeElement {
         return type === PluginType.System;
     }
 
+    get compatible(): boolean {
+        // An extension is not compatible if it has no compatible version.
+        return !!this.version;
+    }
+
     update(data: Partial<VSXExtensionData>): void {
         for (const key of VSXExtensionData.KEYS) {
             if (key in data) {
@@ -260,7 +265,7 @@ export class VSXExtension implements VSXExtensionData, TreeElement {
     }
 
     get tooltip(): string {
-        let md = `__${this.displayName}__ ${this.version}\n\n${this.description}\n_____\n\nPublisher: ${this.publisher}`;
+        let md = `__${this.displayName}__ ${this.version !== undefined ? this.version : 'No Compatible Version'}\n\n${this.description}\n_____\n\nPublisher: ${this.publisher}`;
 
         if (this.license) {
             md += `  \rLicense: ${this.license}`;
@@ -384,7 +389,7 @@ export abstract class AbstractVSXExtensionComponent extends React.Component<Abst
 
     protected renderAction(): React.ReactNode {
         const extension = this.props.extension;
-        const { builtin, busy, installed } = extension;
+        const { builtin, busy, installed, compatible } = extension;
         if (builtin) {
             return <div className="codicon codicon-settings-gear action" onClick={this.manage}></div>;
         }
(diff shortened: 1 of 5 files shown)


packages/vsx-registry/src/browser/vsx-extension.tsx
```diff
@@ -414,16 +421,16 @@ const downloadCompactFormatter = new Intl.NumberFormat(undefined, { notation: 'c
...
-        return <div className='theia-vsx-extension noselect' data-for={tooltipId} data-tip={tooltip}>
+        return <div className={`theia-vsx-extension noselect ${compatible ? '' : 'incompatible'}`} data-for={tooltipId} data-tip={tooltip}>
             {iconUrl ?
                 <img className='theia-vsx-extension-icon' src={iconUrl} /> :
                 <div className='theia-vsx-extension-icon placeholder' />}
             <div className='theia-vsx-extension-content'>
                 <div className='title'>
                     <div className='noWrapInfo'>
-                        <span className='name'>{displayName}</span> <span className='version'>{version}</span>
+                        <span className='name'>{displayName}</span> <span className='version'>{version !== undefined ? version : 'No Compatible Version'}</span>
```
msujew, line 433: Suggestion: please use the nls.localize function for the `No Compatible Version` string.

msujew (commented):
The behavior looks good to me:

- [x] Incompatible version (tested this with API version 1.53.2 and 1.30.0) are shown but not installable.
- [x] Switching the preference on/off correctly starts/stops displaying extensions with incompatible versions
- [x] The visual changes work well with different themes.

I have a few suggestions regarding i18n though.

packages/vsx-registry/src/browser/vsx-extension.tsx
```diff
@@ -398,7 +403,9 @@ export abstract class AbstractVSXExtensionComponent extends React.Component<Abst
             return <div><button className="theia-button action" onClick={this.uninstall}>Uninstall</button>
                 <div className="codicon codicon-settings-gear action" onClick={this.manage}></div></div>;
         }
-        return <button className="theia-button prominent action" onClick={this.install}>Install</button>;
```
vince-fugnitto, line 401: @msujew your previous comment (https://github.com/eclipse-theia/theia/pull/10424/files#r751129871) reminds me, shouldn't we localize the buttons?

msujew: We should, yeah. `nls.localizeByDefault` should do the trick for both of them 👍

colin-grant-work (commented):
A few comments:
### Functionality

- If you search for plugins and then change the preference, the current search result set doesn't change (the search isn't rerun). Particularly when moving from not showing -> showing, I think it would be better if the display did update.
- It seems that if a plugin is incompatible, you can't open an extension view widget for it to read its readme. That was unexpected, and seems undesirable.

### Experience

Displaying a plugin that you can't interact with at all seems odd. Certainly it should be possible to open the extension view for it to see the readme, but I think it might also make sense to allow the user to install an incompatible plugin after a warning dialog that points out that the plugin may not work. Perhaps beyond the scope of this PR, but an excellent addition to it, if you're willing, would be to take the logic from #9330 to give the user the chance to select the oldest available version to install, in case that works better than the latest.

vince-fugnitto:
Converted the pull-request to a draft in favor of completing #9330 first to get a better behavior in this pull-request :+1:

vince-fugnitto:
Moved the draft to my fork in the meantime https://github.com/vince-fugnitto/theia/pull/9. 
The changes should be revisited once the following are completed:
- https://github.com/eclipse-theia/theia/pull/9330
- https://github.com/eclipse-theia/theia/pull/11084

PR #10429, opened 2021-11-17
workaround `electron-rebuild` bug

paul-marechal:
See https://github.com/electron/electron-rebuild/pull/888

The way we invoke `electron-rebuild` should allow Theia application
developers to run `theia rebuild:*` from the root of their monorepos,
but because of a bug in `electron-rebuild` it will only build
dependencies listed in the `package.json` file located in `cwd`.

Implement a workaround that modifies or creates a `package.json` file to
list the dependencies to rebuild.

Add exit hooks to cleanup in case users do `Ctrl+C` while theia rebuild
is running so that we don't pollute user workspaces.

#### How to test

- Run `npx theia rebuild:electron` and `npx theia rebuild:browser` from Theia's monorepo root. It should properly rebuild modules for the specified target.
- Doing `Ctrl+C` while `theia rebuild:electron` is running should revert the modules back to the browser version.
- Doing `Ctrl+C` while `theia rebuild:browser` is running should not prevent the process from finishing.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-manager/src/rebuild.ts
+++ b/dev-packages/application-manager/src/rebuild.ts
@@ -17,8 +17,16 @@
 import cp = require('child_process');
 import fs = require('fs-extra');
 import path = require('path');
+import os = require('os');
 
 export type RebuildTarget = 'electron' | 'browser';
+
+const EXIT_SIGNALS: NodeJS.Signals[] = ['SIGINT', 'SIGTERM'];
+
+interface ExitToken {
+    getLastSignal(): NodeJS.Signals | undefined
+    onSignal(callback: (signal: NodeJS.Signals) => void): void
+}
 
 export const DEFAULT_MODULES = [
     '@theia/node-pty',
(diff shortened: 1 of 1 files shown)


paul-marechal:
Just updated the signals to listen to: I mixed up `SIGKILL` and `SIGTERM`. The former cannot be prevented.

vince-fugnitto (approved):
I confirmed that the changes work well:
- [x] `yarn browser rebuild`
- [x] `yarn electron rebuild`
- [x] `npx theia rebuild:electron` (root) works well
- [x] `npx theia rebuild:browser` (root) works well  
- [x] starting the browser app after the browser rebuild works
- [x] starting the electron app after the electron rebuild works
- [x] performing `ctrl+c` will revert the rebuild properly for `yarn {app} rebuild`
- [x] performing `ctrl+c` will revert the rebuild properly for `npx theia rebuild:{app}`

paul-marechal:
@vince-fugnitto the value of this PR lies in being able to do `npx theia rebuild:browser` and `npx theia rebuild:electron` from the root of the repo. `yarn browser rebuild` and `yarn electron rebuild` change the cwd so that the bug doesn't happen in the first place.

vince-fugnitto:
> Have you tried the former or just the latter?

@paul-marechal both :) I updated my comment https://github.com/eclipse-theia/theia/pull/10429#pullrequestreview-809078653.

colin-grant-work (commented):
Only very minor punctuation comments on the code.

More substantively, I'm able to run `npx theia rebuild:*` from the Theia repo root on `master` (29ebe4c1cc1f8608d23ad92220c6a615c775a27c), so if that's the main objective of this PR, I'm not sure it's necessary. Is there something I'm missing there?

colin-grant-work (approved):
My comments re: punctuation stand, though they don't really need to be addressed. My comment re: ability to run `npx theia rebuild:*` in the current state of `master` appears to have been erroneous: that runs afoul of the electron rebuild bug, and a module fails to self register.

PR #10431, opened 2021-11-18
preferences: Render markdown descriptions

### LINKED ISSUE #7742: preferences: render 'markdownDescription' as markdown
#### Prerequisite:

Following #7105

#### Feature Description:

It is possible to define a `markdownDescription` property for preferences in the schema, which is ultimately used as a description of the preference including the tooltip (when editing JSON), and in the user-interface. In the user-interface, the `markdownDescription` should be properly rendered as markdown similarly to VS Code:

_Theia_:

Screen Shot 2020-05-05 at 9 17 33 AM

_VS Code_:

Screen Shot 2020-05-05 at 9 23 04 AM

msujew:
#### What it does

Closes #7742

Renders the `markdownDescription` property into html instead of just setting the `textContent` to the `markdownDescription` value. As there are some special semantics related to the rendered markdown like links that have to be opened in a new window or linked preferences that have to be rendered as links instead of `code`, I added a new `MarkdownRenderer` that allows easy post-processing on rendered markdown. Perhaps this is a bit overkill, but I couldn't find any other method to overwrite the `code` rule for `markdown-it`.

Also updates some preferences that were using `description` instead of `markdownDescription`.

Previously:

image

Now:

image

Compared to vscode:

image

#### How to test

1. Open the `Preferences` widget.
2. Assert that preferences with a `markdownDescription` property are rendered correctly to html.
3. Assert that preferences with links to other preferences are correctly rendered and clicking the links brings the preference into view
4. Assert that actual http links open a new tab/window when clicking on them.
3. Assert that no preference exists which should use `markdownDescription` instead of `description`. (i.e. no markdown syntax exists in `description` properties)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/markdown-renderer.ts
+++ b/packages/core/src/browser/markdown-renderer.ts
@@ -1,0 +1,76 @@
+/********************************************************************************
+ * Copyright (C) 2021 TypeFox and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import * as DOMPurify from 'dompurify';
+import * as markdownit from 'markdown-it';
+
+export class MarkdownRenderer {
+
+    protected engine: markdownit;
+    protected callbacks = new Map<string, ((element: Element) => Element | void)[]>();
+
+    constructor(engine?: markdownit) {
+        this.engine = engine ?? markdownit();
+    }
+
+    /**
+     * Adds a modification callback that is applied to every element with the specified tag after rendering to HTML.
+     *
+     * @param tag The tag that this modification applies to.
+     * @param callback The modification to apply on every selected rendered element. Can either modify the element in place or return a new element.
+     */
+    modify<K extends keyof HTMLElementTagNameMap>(tag: K, callback: (element: HTMLElementTagNameMap[K]) => Element | void): MarkdownRenderer {
+        if (this.callbacks.has(tag)) {
+            this.callbacks.get(tag)!.push(callback);
+        } else {
+            this.callbacks.set(tag, [callback]);
+        }
+        return this;
+    }
+
+    render(markdown: string): HTMLElement {
+        return this.renderInternal(this.engine.render(markdown));
+    }
+
+    renderInline(markdown: string): HTMLElement {
+        return this.renderInternal(this.engine.renderInline(markdown));
+    }
+
+    protected renderInternal(renderedHtml: string): HTMLElement {
+        const div = this.sanitizeHtml(renderedHtml);
+        for (const [tag, calls] of this.callbacks) {
+            for (const callback of calls) {
+                const elements = Array.from(div.getElementsByTagName(tag));
+                for (const element of elements) {
+                    const result = callback(element);
+                    if (result) {
+                        const parent = element.parentElement;
+                        if (parent) {
+                            parent.replaceChild(result, element);
+                        }
+                    }
+                }
+            }
+        }
+        return div;
+    }
+
+    protected sanitizeHtml(html: string): HTMLElement {
+        const div = document.createElement('div');
+        div.innerHTML = DOMPurify.sanitize(html);
+        return div;
+    }
+}
--- a/packages/preferences/src/browser/preference-tree-model.ts
+++ b/packages/preferences/src/browser/preference-tree-model.ts
@@ -227,6 +227,11 @@ export class PreferenceTreeModel extends TreeModelImpl {
         }
     }
 
+    getNodeFromPreferenceId(id: string): Preference.TreeNode | undefined {
+        const node = this.getNode(this.treeGenerator.getNodeId(id));
+        return node && Preference.TreeNode.is(node) ? node : undefined;
+    }
+
     /**
      * @returns true if selection changed, false otherwise
      */
--- a/packages/preferences/src/browser/util/preference-types.ts
+++ b/packages/preferences/src/browser/util/preference-types.ts
@@ -64,6 +64,7 @@ export namespace Preference {
     export interface LeafNode extends BaseTreeNode {
         depth: number;
         preference: { data: PreferenceDataProperty };
+        preferenceId: string;
     }
 
     export namespace LeafNode {
--- a/packages/preferences/src/browser/views/components/preference-node-renderer.ts
+++ b/packages/preferences/src/browser/views/components/preference-node-renderer.ts
@@ -15,15 +15,19 @@
  ********************************************************************************/
 
 import { injectable, inject, postConstruct } from '@theia/core/shared/inversify';
-import { PreferenceService, ContextMenuRenderer, PreferenceInspection, PreferenceScope, PreferenceProvider } from '@theia/core/lib/browser';
+import {
+    PreferenceService, ContextMenuRenderer, PreferenceInspection,
+    PreferenceScope, PreferenceProvider, MarkdownRenderer, codicon, animationFrame
+} from '@theia/core/lib/browser';
 import { Preference, PreferenceMenus } from '../../util/preference-types';
 import { PreferenceTreeLabelProvider } from '../../util/preference-tree-label-provider';
 import { PreferencesScopeTabBar } from '../preference-scope-tabbar-widget';
-import { Disposable } from '@theia/core/lib/common';
+import { Disposable, nls } from '@theia/core/lib/common';
 import { JSONValue } from '@theia/core/shared/@phosphor/coreutils';
 import debounce = require('@theia/core/shared/lodash.debounce');
 import { PreferenceTreeModel } from '../../preference-tree-model';
 import { PreferencesSearchbarWidget } from '../preference-searchbar-widget';
+import { WindowService } from '@theia/core/lib/browser/window/window-service';
 
 export const PreferenceNodeRendererFactory = Symbol('PreferenceNodeRendererFactory');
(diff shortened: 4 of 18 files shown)


vince-fugnitto (commented):
The changes look good to me overall, believe it fixes https://github.com/eclipse-theia/theia/issues/7742 if you'd like to link it 👍 

I have a couple of suggestions for improvements, but they can be tackled in separate pull-requests:
- ensuring that links (ex: `read more`) open in an external window and do not replace the app.
- ensuring that preferences are anchors, so we can reference them when clicking the following `#search.searchOnType#`.

msujew:
> believe it fixes #7742 if you'd like to link it 👍

Ah, good to know 👍 

> I have a couple of suggestions for improvements, but they can be tackled in separate pull-requests:

I see, I will address them in this PR 👍

vince-fugnitto (commented):
Thank you for taking care of the updates, I can confirm that links open in new windows, and we can reference other preference ids through the markdown as anchors.

There are some instances where preferences are not anchorable:

image

Is this an issue with `localization` in general? We might reference invalid preference ids because we want to use strings present in `nls.metadata.json` for translation. For an end-user it might be confusing when we use a preference description and refer to an invalid id.

packages/core/src/browser/markdown-renderer.ts
```diff
@@ -0,0 +1,83 @@
...
+            // As the `taggedElements` collection may change while calling `replaceChild`,
+            // we have to create a copy of the collection beforehand.
+            const taggedElements = div.getElementsByTagName(tag);
+            const elements: Element[] = [];
+            for (let i = 0; i < taggedElements.length; i++) {
+                const element = taggedElements.item(i);
+                if (element) {
+                    elements.push(element);
+                }
+            }
```
colin-grant-work, line 65: Could this just be `Array.from(taggedElements)`?

packages/core/src/browser/markdown-renderer.ts
```diff
@@ -0,0 +1,83 @@
...
+            for (const callback of calls) {
+                for (const element of elements) {
+                    const result = callback(element);
+                    if (result) {
+                        div.replaceChild(result, element);
+                    }
+                }
+            }
```
colin-grant-work, line 73: This looks like it could easily fail. If there are multiple callbacks, and one or more of them return a new element in place of the original element, then the `div.replaceChild` call will subsequently fail because `element` is no longer a child of `div`: it will have been replaced by a previously-run callback.

This would be the scenario:

image

packages/preferences/src/browser/preference-tree-model.ts
```diff
@@ -227,6 +227,17 @@ export class PreferenceTreeModel extends TreeModelImpl {
...
+        if (this.root) {
+            for (const node of new TopDownTreeIterator(this.root)) {
+                if (Preference.LeafNode.is(node) && node.preferenceId === id) {
+                    return node;
+                }
+            }
+        }
+        return undefined;
+    }
+
```
colin-grant-work, line 240: I believe that this can be made simpler if you make the `getGroupName` method of `PreferenceTreeGenerator` public and then do this:

```suggestion
    getNodeFromPreferenceId(preferenceId: string): Preference.LeafNode | undefined {
         const expectedGroup = this.treeGenerator.getGroupName(preferenceId.split('.'));
         const expectedId = `${expectedGroup}@${preferenceId}` // could combine this line and the previous as a utility on the PreferenceTreeGenerator: 'getNodeId(preferencId: string): string'
         return this.getNode(expectedId);
    }
```

The only reason ID's can't be computed completely programmatically is that an arbitrary set of preferences are in two groups: `Commonly Used` and whatever group they belong to 'naturally', but this procedure should always generate the correct ID pointing to the 'natural' location of a preference.

packages/preferences/src/browser/views/components/preference-node-renderer.ts
```diff
@@ -170,6 +175,65 @@ export abstract class PreferenceLeafNodeRenderer<ValueType extends JSONValue, In
...
+
+    protected async selectPreference(event: MouseEvent, preferenceId: string): Promise<void> {
+        event.preventDefault();
+        event.stopPropagation();
+        // Exclude right click
+        if (event.which < 3) {
+            const selector = `li[data-pref-id="${preferenceId}"]`;
+            const element = document.querySelector(selector);
+            if (element) {
+                if (element.clientHeight === 0 || element.clientWidth === 0) {
```
colin-grant-work, line 227: There should be a more elegant check for this - we should be able to get the renderer and check whether it's `hidden`, but there isn't, so this works alright. Mea culpa, since I wrote all this :-/.

packages/preferences/src/browser/views/components/preference-node-renderer.ts
```diff
@@ -170,6 +175,65 @@ export abstract class PreferenceLeafNodeRenderer<ValueType extends JSONValue, In
...
+                }
+                return code;
+            });
+    }
+
+    protected openLink(event: MouseEvent, href: string): void {
+        event.preventDefault();
+        event.stopPropagation();
+        // Exclude right click
+        if (event.which < 3) {
```
colin-grant-work, line 214: looks like `event.button` is the preferred way to figure out what button was clicked. Is there a reason not to use it here? (Same at l. 223)

msujew: I got the inspiration from here: https://github.com/microsoft/vscode/blob/7be48d527a1a7fdf5fbaf4fc7af52195e6fa0bac/src/vs/code/electron-sandbox/issue/issueReporterMain.ts#L1190-L1197

However, `event.button` seems to be the better solution 👍

msujew:
@vince-fugnitto Thanks for the feedback, I removed the references to non-existing preferences and used custom localization keys 👍

packages/core/src/browser/markdown-renderer.ts
msujew: I restructured this loop, as using multiple callbacks for a single `tag` was able to violate the contract. For example, if someone has two callbacks for `a` elements and the first callback builds `div` from them, the second callback would operate on these `div` elements, instead of re-selecting all `a` elements inside our rendered markdown.

This shouldn't fail anymore, as we re-select all tagged elements in each callback loop.

packages/preferences/src/browser/views/components/preference-node-renderer.ts
msujew: Yeah, I wasn't too happy with the solution as well, since it directly operates on the DOM.

I made the check a bit more sophisticated, using the `hidden` class and only selecting preference nodes which don't appear in the commonly-used group. Are you fine with this approach (at least for this PR)?

colin-grant-work: Checking for the `hidden` class seems safer than checking the size. 👍

packages/core/src/browser/markdown-renderer.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    protected renderInternal(renderedHtml: string): HTMLElement {
+        const div = this.sanitizeHtml(renderedHtml);
+        for (const [tag, calls] of this.callbacks) {
+            for (const callback of calls) {
+                const elements: Element[] = Array.from(div.getElementsByTagName(tag));
+                for (const element of elements) {
+                    const result = callback(element);
+                    if (result) {
+                        div.replaceChild(result, element);
```
colin-grant-work, line 61: It's still pretty easy to get this to fail. `.getElementsByTagName` retrieves all descendants, but (in Chrome, at least) `replaceChild` only works if the element to be replaced is an immediate child of the element on which `replaceChild` is called, so if the original rendering or a previous callback creates a nested structure that includes the tag of interest, the next callback targeting that tag will fail. Since `markdown-it` already supports plugins, I wouldn't make this hard on yourself trying to invent a new plugin system for it: just apply the modifications you want to apply to the HTML.

msujew: I would like to use the plugin system of `markdown-it`, but it's just not powerful enough. The main issue is that it always renders to text, eliminating every listener I place on the html elements in the process. The main purpose of the `MarkdownRenderer` class is to operate directly on the html elements. (I've tried implementing it using purely `markdown-it` at the start, but couldn't get around this limitation)

Anyway, changed this to only select direct children of the `div`.

packages/preferences/src/browser/util/preference-types.ts
```diff
@@ -64,6 +64,7 @@ export namespace Preference {
     export interface LeafNode extends BaseTreeNode {
         depth: number;
         preference: { data: PreferenceDataProperty };
+        preferenceId: string;
```
colin-grant-work, line 67: I'm Ok with adding this, but there is also a utility in the `Preference.TreeNode` namespace (`getGroupAndIdFromNodeId`) to get the `group` and `id` (i.e. preferenceId) from the ID of the node.

packages/core/src/browser/markdown-renderer.ts
colin-grant-work: Thanks for adding the tests. I don't see an obvious way to break the current implementation (though a triply nested for-loop is a little scary...) 👍. I'm still not sure whether it's worth adding that machinery when the only current use case is a fairly simple one (running one callback on each of two different tags). Adding it puts us on the hook for its correctness in the general case when that level of complexity isn't (yet) needed. But it is working, and the tests suggest it should handle the general case, so I'll leave it up to you.

colin-grant-work (commented):
This is behaving correctly and the implementation looks sound. The one outstanding issue that I'd like to discuss - it may or may not need to be addressed - is that in Electron, clicking an external link opens a new Electron window rather than external browser window as VSCode does. That can make Electron behave oddly, because events from that window are treated the same as events from 'real' Theia windows, which can be undesirable (I think it would mess with the code to close windows in #10379, for example, since the new window wouldn't have the code to confirm exit). FWIW, I think it would be better for it to open an external window, but the fact that it's so easy to open a non-IDE Electron window means that #10379 needs to be more careful about which windows it tries to ask for close confirmation.

msujew:
@colin-grant-work Thanks, I only tested this in the browser environment. I wasn't aware of the whole `WindowService.openNewWindow` stuff. It should work as expected in electron now :)

vince-fugnitto (approved):
I confirmed that:
- [x] markdown descriptions of preferences are correctly styled
- [x] clicking the anchor links of preferences correctly navigates to the corresponding preference
- [x] opening external links works correctly for the browser and electron
- [x] no visible regressions
  - searching works
  - scrolling works
  - layout restoration works

colin-grant-work (approved):
Looking good and working well for me. 👍

Issue #10417, opened 2021-11-15
enable TypeScript strict checks


paul-marechal:
We currently only have `strictNullChecks` enabled, which is the least we can do when using TypeScript. It already guards against assigning `null` or `undefined` to objects:

```ts
interface Foo {
  a: string
  b: number
}

const a: Foo = null; // Error thanks to strictNullChecks.
```

But there are other checks that are currently not enabled and that let some _potential_ issues slip into the code base, e.g.:

```ts

interface Bar {
  method(a: Meh | number): string
}

class Baz implements Bar {
  // Here TypeScript should fail but doesn't because strictFunctionTypes is not enabled.
  method(a: string): string {
     return a;
  }
}

const bar: Bar = new Baz();
const result: string = bar.method({ a: 'meh', b: 2}); // `result` isn't a string here, but TypeScript doesn't know.
```

There's also `strictPropertyInitialization` that tries to enforce proper initialization of class fields. If not done at construction time, then the field must be declared as potentially undefined, because it will be undefined until some later point in time. This check has to be the most annoying one in our code base as we are using Inversify's property injectors:

```ts
@injectable()
class Qux {
  @inject(Something)
  protected something: Something; // TypeScript will complain here because not set at construction time.
}
```

This issue can be avoided by telling TypeScript to ignore checking those fields by adding a `!`:

```ts
@injectable()
class Qux {
  @inject(Something)
  protected something!: Something; // TypeScript is happy.
}
```

Enabling this check will incur a LOT of changes but going forward it will help catch mistakes where a field might be left undefined but is not typed as such.

tsmaeder:
Is that list of checks you're proposing final or are there others that might get turned on?

paul-marechal:
I want to run `tsc --strict` which turns on every strict check listed here: https://www.typescriptlang.org/docs/handbook/compiler-options.html#compiler-options

<html><body>


flag | type | default
-- | -- | --
`--strict` | boolean | false
Enable all strict type-checking options.
`--strictBindCallApply` | boolean | true if strict, false otherwise.
Check that the arguments for bind, call, and apply methods match the original function.
`--strictFunctionTypes` | boolean | true if strict, false otherwise.
When assigning functions, check to ensure parameters and the return values are subtype-compatible.
`--strictNullChecks` | boolean | true if strict, false otherwise.
When type checking, take into account null and undefined.
`--strictPropertyInitialization` | boolean | true if strict, false otherwise.
[...]

PR #10435, opened 2021-11-19
enable TS strict mode

paul-marechal:
#### What it does

Enable TS strict checks and fixes related problems in the code base.

I used a few strategies:

- Any injected field is marked as not null with `!`
- Unless there is a `@optional()` decorator in which case I mark the field as nullable with `?`
- Initialize boolean values with `false` because `undefined` is falsy (to conserve the current logic)
- Initialize some fields as early as possible by slightly changing the logic
- When not possible to initialize fields, I added checks at use time to fail early if something is `undefined` when it shouldn't

Then there's another special case with some classes like:

```ts
@injectable()
export class SomeOptions {
  readonly option1: string
  readonly option2: number
}
```

My guess is that this was more convenient than writing:

```ts
export const SomeOptions = Symbol('SomeOptions');
export interface SomeOptions {
  readonly option1: string
  readonly option2: number
}
```

The issue is that in the `class` version, the field are declared as not nullable, but they aren't initialized either. This means that doing `new SomeOptions()` will give back a bogus object. Furthermore those classes aren't ever instantiated: they're always used as a type and some regular JSON object is passed instead. To keep as much semantic, I fixed the issue by making both the class and its non nullable fields abstract.

Alternatively I could convert those classes into symbol+interface if we think it's a better idea.

Close https://github.com/eclipse-theia/theia/issues/10417

#### How to test

Pure refactoring, everything should keep working like before.

#### Review checklist

#### Reminder for reviewers


packages/core/src/browser/tree/tree.spec.ts
```diff
@@ -221,8 +221,8 @@ describe('Tree', () => {
     }
 
     function assertTreeNode(expectation: string, node: TreeNode): void {
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        assert.deepStrictEqual(expectation, JSON.stringify(node, (key: keyof CompositeTreeNode, value: any) => {
+        assert.deepStrictEqual(expectation, JSON.stringify(node, (key, value) => {
+            // eslint-disable-next-line no-null/no-null
```
colin-grant-work, line 225: Doesn't look like this `no-null/no-null` is necessary?

packages/core/src/browser/widgets/widget.ts
```diff
@@ -254,9 +254,10 @@ export type EventListenerOrEventListenerObject<K extends keyof HTMLElementEventM
 export function addEventListener<K extends keyof HTMLElementEventMap>(
     element: HTMLElement, type: K, listener: EventListenerOrEventListenerObject<K>, useCapture?: boolean
 ): Disposable {
-    element.addEventListener(type, listener, useCapture);
+    // The DOM API typings don't seem to take a specific listener as described in this function.
+    element.addEventListener(type, listener as EventListenerOrEventListenerObject<any>, useCapture);
```
colin-grant-work, line 258: Changing the type of `EventListenerOrEventListenerObject<K>` to `(this: HTMLElement, ev: HTMLElementEventMap[K]) => unknown` (derived from the DOM `element.addEventListener` typing) appears to fix this without introducing any new type errors.

packages/core/src/node/backend-application.ts
```diff
@@ -101,20 +101,21 @@ export interface BackendApplicationContribution {
...
+    certkey?: string;
 
     configure(conf: yargs.Argv): void {
         conf.option('port', { alias: 'p', description: 'The port the backend server listens on.', type: 'number', default: DEFAULT_PORT });
         conf.option('hostname', { alias: 'h', description: 'The allowed hostname for connections.', type: 'string', default: DEFAULT_HOST });
         conf.option('ssl', { description: 'Use SSL (HTTPS), cert and certkey must also be set', type: 'boolean', default: DEFAULT_SSL });
         conf.option('cert', { description: 'Path to SSL certificate.', type: 'string' });
         conf.option('certkey', { description: 'Path to SSL certificate key.', type: 'string' });
-        conf.option(APP_PROJECT_PATH, { description: 'Sets the application project directory', default: this.appProjectPath() });
+        conf.option(APP_PROJECT_PATH, { description: 'Sets the application project directory' });
```
colin-grant-work, line 118: Why remove this `default`?

colin-grant-work (commented):
First round...

examples/api-samples/src/browser/file-watching/sample-file-watching-contribution.ts
```diff
@@ -50,19 +50,19 @@ const FileWatchingPreferencesSchema: PreferenceSchema = {
 @injectable()
 class SampleFileWatchingContribution implements FrontendApplicationContribution {
 
-    protected verbose: boolean;
+    protected verbose: boolean = true;
```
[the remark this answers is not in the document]
paul-marechal: Good catch

dev-packages/application-package/src/extension-package-collector.ts
```diff
@@ -73,7 +73,7 @@ export class ExtensionPackageCollector {
         if (RawExtensionPackage.is(pck)) {
             const parent = this.parent;
             const version = pck.version;
-            const transitive = !(name in this.root.dependencies!);
+            const transitive = !(name in this.root!.dependencies!);
```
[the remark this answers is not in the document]
paul-marechal: My guess is that if we had the strict checks enabled before, this method wouldn't have been implemented without this oddity in the first place. So not sure what to think about this code indeed. What I like right now is that it properly describes how `root` can be `undefined` but that this place expects it not to be.

packages/core/src/node/backend-application.ts
paul-marechal: I had understood that `BackendApplicationCliContribution.configure/setArguments` wasn't called when this component was used, but turns out it might because the CLI is initialized in the generated app. Will revert this part now that I better understand how it works. I got confused.

packages/core/src/browser/widgets/widget.ts
paul-marechal: It doesn't show an error in the editor but it fails at build time, will leave as is for now.

JonasHelming:
@paul-marechal : Are you ready to trigger a re-review from colin?

paul-marechal:
@JonasHelming I will need to rebase first. I'd also like to have other opinions about this. I believe this is for the better despite the additional strictness, but I know some prefer it more lax...

JonasHelming:
@paul-marechal : I think we should not let this go stale, could you bring it up in the dev call again, please?

paul-marechal:
@JonasHelming if https://github.com/eclipse-theia/theia/pull/10355 gets merged, there's even one more flag I'd like to turn on which would cause yet another big PR to align the code base with it: https://www.typescriptlang.org/tsconfig#noImplicitOverride

I'll bring it up at the next dev meeting.

dev-packages/application-package/src/extension-package-collector.ts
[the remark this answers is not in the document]
msujew: In the rare case that this could become undefined in the future, I believe that having some sort of guard for that would be great. Something as simple as `if (!this.root) { return; }`

paul-marechal:
So many conflicts... I'll close for now and come back to enable this flag in bits later.

PR #10441, opened 2021-11-19
Add sash option for widget resize

### LINKED ISSUE #10416: Add window resize "sash" options
For some users (including those working through a remote desktop client with latency) it can be quite hard to resize windows & view containers as the mouse target size for resize handles is quite small. VSCode has a feature called workbench "sash" that applies focus to borders as you drag over them. VSCode also implements 2 preferences to customize this behavior (hover delay and size):

sash

This would be a really nice feature to have

kaiyue0329:
#### What it does
Fixes #10416

https://user-images.githubusercontent.com/23107734/143316416-621fb965-b0b1-4c0a-8d72-6fc51cc38cf6.mov

- Apply focus to a widget handle when you hover over it
- `sash.hoverDelay` controls the hover feedback delay in milliseconds
- `sash.size` controls the feedback area size in pixels for the dragging area
- Add preference to set workbench.sash.hoverDelay (value must be in the range of 0 to 2000, default is 0ms)
- Add preference to set workbench.sash.size (value must be in the range of 1 to 20, default is 4px)
- In the left panel, a sash should only appear when the following two conditions are met:
            1. the active `horizontal` container is expanded
            2. the container that is above/below the active `horizontal` container is also expanded
- Move the utilities for creating and manipulating dynamic stylesheets from `editor-decoration-style.ts` to `decoration-style.ts` in `core` so that: 
    - Each namespace now has its indepednent style sheet
    - Only one rule should exist for a given selector in the provided stylesheet

| Senario  | Demo | Sash State |
| ------------- | ------------- | ------------- | 
| Active container expanded, Container below collapsed |  container-below-collapsed | Hidden  |
| Active container expanded, Container below expanded | container-below-expanded  | **Shown***  |
| Active container expanded, Container above collapsed | container-above-collapsed | Hidden  |
| Active container expanded, Container above expanded | container-above-expanded  | **Shown***  |
| Active container collapsed  | see first video below | Hidden  |
| No active container, All containers collapsed  | see second video below  | Hidden |

https://user-images.githubusercontent.com/23107734/144339788-7d484181-2bb0-4417-990d-9dbd9390061f.mov

https://user-images.githubusercontent.com/23107734/144339924-8d186a8b-03b6-4f96-ace8-60a2631e72d3.mov

#### How to test
1.	Start the application
2.	Hover the cursor over the horizontal and vertical handles for different view containers in Theia and drag 
**(please refer to the screencast above for more details)**
3.	Search for 'workbench sash' in Preferences (File > Preferences > Open Settings (UI) )
4.	Update the `sash.hoverDelay` and `sash.size` preferences to valid values
5.	Repeat Step 2 and verify that the hover delay and feedback area size for the border/sash are updated successfull
6. Verify the sash behaviors for the horizontal containers in the left panel based on the info provided in the table above

#### Review checklist

#### Reminder for reviewers

Signed-off-by: Kaiyue Pan <kaiyue.pan@ericsson.com>

### DIFF
--- a/packages/core/src/browser/common-frontend-contribution.ts
+++ b/packages/core/src/browser/common-frontend-contribution.ts
@@ -58,6 +58,7 @@ import { CurrentWidgetCommandAdapter } from './shell/current-widget-command-adap
 import { ConfirmDialog, confirmExit, Dialog } from './dialogs';
 import { WindowService } from './window/window-service';
 import { FrontendApplicationConfigProvider } from './frontend-application-config-provider';
+import { DecorationStyle } from './decoration-style';
 
 export namespace CommonMenus {
 
@@ -308,6 +309,8 @@ export const RECENT_COMMANDS_STORAGE_KEY = 'commands';
 @injectable()
 export class CommonFrontendContribution implements FrontendApplicationContribution, MenuContribution, CommandContribution, KeybindingContribution, ColorContribution {
 
+    protected commonDecorationsStyleSheet: CSSStyleSheet = DecorationStyle.createStyleSheet('coreCommonDecorationsStyle');
+
     constructor(
         @inject(ApplicationShell) protected readonly shell: ApplicationShell,
         @inject(SelectionService) protected readonly selectionService: SelectionService,
@@ -378,6 +381,7 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
         this.updateStyles();
         this.updateThemeFromPreference('workbench.colorTheme');
         this.updateThemeFromPreference('workbench.iconTheme');
+        this.preferences.ready.then(() => this.setSashProperties());
         this.preferences.onPreferenceChanged(e => this.handlePreferenceChange(e, app));
         this.themeService.onDidColorThemeChange(() => this.updateThemePreference('workbench.colorTheme'));
         this.iconThemes.onDidChangeCurrent(() => this.updateThemePreference('workbench.iconTheme'));
@@ -466,7 +470,22 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
                 }
                 break;
             }
+            case 'workbench.sash.hoverDelay':
+            case 'workbench.sash.size': {
+                this.setSashProperties();
+                break;
+            }
         }
+    }
+
+    protected setSashProperties(): void {
+        const sashRule = `:root {
+            --theia-sash-hoverDelay: ${this.preferences['workbench.sash.hoverDelay']}ms;
+            --theia-sash-width: ${this.preferences['workbench.sash.size']}px;
+        }`;
+
+        DecorationStyle.deleteStyleRule(':root', this.commonDecorationsStyleSheet);
+        this.commonDecorationsStyleSheet.insertRule(sashRule);
     }
 
     onStart(): void {
@@ -1100,7 +1119,8 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
             // if not yet contributed by Monaco, check runtime css variables to learn
             { id: 'selection.background', defaults: { dark: '#217daf', light: '#c0dbf1' }, description: 'Overall border color for focused elements. This color is only used if not overridden by a component.' },
             { id: 'icon.foreground', defaults: { dark: '#C5C5C5', light: '#424242', hc: '#FFFFFF' }, description: 'The default color for icons in the workbench.' },
-
+            { id: 'sash.hoverBorder', defaults: { dark: Color.transparent('focusBorder', 0.99), light: Color.transparent('focusBorder', 0.99), hc: Color.transparent('focusBorder', 0.99) }, description: 'The hover border color for draggable sashes.' },
+            { id: 'sash.activeBorder', defaults: { dark: 'focusBorder', light: 'focusBorder', hc: 'focusBorder' }, description: 'The active border color for draggable sashes.' },
             // Window border colors should be aligned with https://code.visualstudio.com/api/references/theme-color#window-border
             {
                 id: 'window.activeBorder', defaults: {
--- a/packages/core/src/browser/core-preferences.ts
+++ b/packages/core/src/browser/core-preferences.ts
@@ -135,6 +135,20 @@ export const corePreferenceSchema: PreferenceSchema = {
             default: isOSX ? 1500 : 500,
             description: nls.localizeByDefault('Controls the delay in milliseconds after which the hover is shown.')
         },
+        'workbench.sash.hoverDelay': {
+            type: 'number',
+            default: 300,
+            minimum: 0,
+            maximum: 2000,
+            description: nls.localizeByDefault('Controls the hover feedback delay in milliseconds of the dragging area in between views/editors.')
+        },
+        'workbench.sash.size': {
+            type: 'number',
+            default: 4,
+            minimum: 1,
+            maximum: 20,
+            description: nls.localizeByDefault('Controls the feedback area size in pixels of the dragging area in between views/editors. Set it to a larger value if needed.')
+        },
     }
 };
 
@@ -154,6 +168,8 @@ export interface CoreConfiguration {
     'workbench.statusBar.visible': boolean;
     'workbench.tree.renderIndentGuides': 'onHover' | 'none' | 'always';
     'workbench.hover.delay': number;
+    'workbench.sash.hoverDelay': number;
+    'workbench.sash.size': number;
 }
 
 export const CorePreferenceContribution = Symbol('CorePreferenceContribution');
--- a/packages/core/src/browser/decoration-style.ts
+++ b/packages/core/src/browser/decoration-style.ts
@@ -1,0 +1,61 @@
+/********************************************************************************
+ * Copyright (C) 2021 Ericsson and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+export namespace DecorationStyle {
+
+    export function createStyleSheet(styleId: string, container: HTMLElement = document.getElementsByTagName('head')[0]): CSSStyleSheet {
+        const style = document.createElement('style');
+        style.id = styleId;
+        style.type = 'text/css';
+        style.media = 'screen';
+        style.appendChild(document.createTextNode('')); // trick for webkit
+        container.appendChild(style);
+        return <CSSStyleSheet>style.sheet;
+    }
+
+    function getRuleIndex(selector: string, styleSheet: CSSStyleSheet): number {
+        return Array.from(styleSheet.cssRules || styleSheet.rules).findIndex(rule => rule.type === CSSRule.STYLE_RULE && (<CSSStyleRule>rule).selectorText === selector);
+    }
+
+    export function getOrCreateStyleRule(selector: string, styleSheet: CSSStyleSheet): CSSStyleRule {
+        let index = getRuleIndex(selector, styleSheet);
+        if (index === -1) {
+            // The given selector does not exist in the provided independent style sheet, rule index = 0
+            index = styleSheet.insertRule(selector + '{}', 0);
+        }
+
+        const rules = styleSheet.cssRules || styleSheet.rules;
+        const rule = rules[index];
+        if (rule && rule.type === CSSRule.STYLE_RULE) {
+            return rule as CSSStyleRule;
+        }
+        styleSheet.deleteRule(index);
+        throw new Error('This function is only for CSS style rules. Other types of CSS rules are not allowed.');
+    }
+
+    export function deleteStyleRule(selector: string, styleSheet: CSSStyleSheet): void {
+        if (!styleSheet) {
+            return;
+        }
+
+        // In general, only one rule exists for a given selector in the provided independent style sheet
+        const index = getRuleIndex(selector, styleSheet);
+        if (index !== -1) {
+            styleSheet.deleteRule(index);
+        }
+    }
+}
+
--- a/packages/core/src/browser/style/view-container.css
+++ b/packages/core/src/browser/style/view-container.css
@@ -36,14 +36,39 @@
     min-height: var(--theia-content-line-height);
 }
 
-.theia-view-container > .p-SplitPanel > .p-SplitPanel-handle:after {
-    background-color: var(--theia-sideBarSectionHeader-border);
+.theia-view-container > .p-SplitPanel > .p-SplitPanel-handle::after {
     min-height: 2px;
-    min-width: 2px
+    min-width: 2px;
 }
 
 .theia-view-container > .p-SplitPanel > .p-SplitPanel-handle {
     background-color: var(--theia-sideBarSectionHeader-border);
+}
+
+.p-SplitPanel > .p-SplitPanel-handle:hover::after {
+    background-color: var(--theia-sash-hoverBorder);
+    transition-delay: var(--theia-sash-hoverDelay);
+}
+
+.p-SplitPanel > .p-SplitPanel-handle:active::after {
+    background-color: var(--theia-sash-activeBorder);
+    transition-delay: 0s !important;
+}
+
+.p-SplitPanel[data-orientation='horizontal'] > .p-SplitPanel-handle::after {
+    min-width: var(--theia-sash-width);
+}
+
+.p-SplitPanel[data-orientation='vertical'] > .p-SplitPanel-handle::after {
+    min-height: var(--theia-sash-width);
+}
+
+.p-SplitPanel[data-orientation='vertical'] > .p-SplitPanel-handle.sash-hidden {
+    visibility: hidden;
+}
+
+.p-SplitPanel[data-orientation='vertical'] > .p-SplitPanel-handle.sash-hidden::after {
+    min-height: 0px;
 }
 
 .theia-view-container .part {
--- a/packages/core/src/browser/view-container.ts
+++ b/packages/core/src/browser/view-container.ts
@@ -46,6 +46,11 @@ export interface ViewContainerTitleOptions {
     closeable?: boolean;
 }
 
+enum SashType {
+    TopSash = 'top-sash',
+    BottomSash = 'bottom-sash'
+}
+
 @injectable()
 export class ViewContainerIdentifier {
     id: string;
@@ -994,6 +999,11 @@ export class ViewContainerPart extends BaseWidget {
         }
         this._collapsed = collapsed;
         this.node.classList.toggle('collapsed', collapsed);
+
+        // Update the sashes for the active `horizontal` container when the container is collapsed/expanded
+        this.updateSashState(SashType.TopSash, this.node.previousElementSibling, collapsed);
+        this.updateSashState(SashType.BottomSash, this.node.nextElementSibling, collapsed);
+
         if (collapsed && this.wrapped.node.contains(document.activeElement)) {
             this.header.focus();
         }
@@ -1230,6 +1240,26 @@ export class ViewContainerPart extends BaseWidget {
         }
     }
 
+    protected updateSashState(sashType: SashType, sashElement: Element | null, activeContainerCollapsed: boolean): void {
+        if (sashElement && sashElement.className.includes('p-SplitPanel-handle')) {
+            // Hide the sash when the active `horizontal` container in the left panel is collapsed
+            sashElement.classList.toggle('sash-hidden', activeContainerCollapsed);
+
+            let adjacentContainer: Element | null;
+            if (sashType === SashType.TopSash) {
+                adjacentContainer = sashElement.previousElementSibling;
+            } else {
+                adjacentContainer = sashElement.nextElementSibling;
+            }
+
+            // Sash should only appear when the following two conditions are met:
+            //    1.  the active `horizontal` container is expanded
+            //    2.  the container that is above/below the active `horizontal` container is also expanded
+            if (!activeContainerCollapsed && adjacentContainer) {
+                sashElement.classList.toggle('sash-hidden', adjacentContainer.className.includes('collapsed'));
+            }
+        }
+    }
 }
 
 export namespace ViewContainerPart {
--- a/packages/editor/src/browser/decorations/editor-decoration-style.ts
+++ b/packages/editor/src/browser/decorations/editor-decoration-style.ts
@@ -15,14 +15,19 @@
  ********************************************************************************/
 
 import { Disposable } from '@theia/core';
+import { DecorationStyle } from '@theia/core/lib/browser';
 
 export class EditorDecorationStyle implements Disposable {
 
     constructor(
         readonly selector: string,
         styleProvider: (style: CSSStyleDeclaration) => void,
+        protected decorationsStyleSheet: CSSStyleSheet
     ) {
-        EditorDecorationStyle.createRule(selector, styleProvider);
+        const styleRule = DecorationStyle.getOrCreateStyleRule(selector, decorationsStyleSheet);
+        if (styleRule) {
+            styleProvider(styleRule.style);
+        }
     }
 
     get className(): string {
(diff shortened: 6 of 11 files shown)


packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -1140,6 +1140,7 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
             // if not yet contributed by Monaco, check runtime css variables to learn
             { id: 'selection.background', defaults: { dark: '#217daf', light: '#c0dbf1' }, description: 'Overall border color for focused elements. This color is only used if not overridden by a component.' },
             { id: 'icon.foreground', defaults: { dark: '#C5C5C5', light: '#424242', hc: '#FFFFFF' }, description: 'The default color for icons in the workbench.' },
+            { id: 'sash.hoverBorder', defaults: { dark: '#007acc', light: '#007acc', hc: '#007acc' }, description: 'The hover border color for draggable sashes.' },
```
vince-fugnitto, line 1143: minor - I believe we can re-use `focusBorder`:

```suggestion
            { id: 'sash.hoverBorder', defaults: { dark: 'focusBorder', light: 'focusBorder', hc: 'focusBorder' }, description: 'The hover border color for draggable sashes.' },
```

packages/core/src/browser/core-preferences.ts
```diff
@@ -136,6 +136,20 @@ export const corePreferenceSchema: PreferenceSchema = {
...
+            maximum: 2000,
+            description: nls.localizeByDefault('Controls the hover feedback delay in milliseconds of the dragging area in between views/editors.')
+        },
+        'workbench.sash.size': {
+            type: 'number',
+            default: 7,
+            minimum: 1,
+            maximum: 20,
+            description: nls.localizeByDefault('Controls the feedback area size in pixels of the dragging area in between views/editors. Set it to a larger value if needed.')
+        },
```
vince-fugnitto, line 152: Any reason we do not want to use the same values and defaults as in vscode?

```ts
'workbench.sash.size': {
	type: 'number',
	default: isIOS ? 20 : 4,
	minimum: 1,
	maximum: 20,
	description: localize('sashSize', "Controls the feedback area size in pixels of the dragging area in between views/editors. Set it to a larger value if you feel it's hard to resize views using the mouse.")
},
'workbench.sash.hoverDelay': {
	type: 'number',
	default: 300,
	minimum: 0,
	maximum: 2000,
	description: localize('sashHoverDelay', "Controls the hover feedback delay in milliseconds of the dragging area in between views/editors.")
},
```

packages/core/src/browser/view-container.ts
```diff
@@ -46,6 +47,9 @@ export interface ViewContainerTitleOptions {
     closeable?: boolean;
 }
 
+const SASH_DELAY_PREFERENCE = 'workbench.sash.hoverDelay';
+const SASH_SIZE_PREFERENCE = 'workbench.sash.size';
```
vince-fugnitto, line 51: I don't believe the variables are necessary, the `CorePreferences` should already know about them.

packages/core/src/browser/view-container.ts
```diff
@@ -837,6 +857,20 @@ export class ViewContainer extends BaseWidget implements StatefulWidget, Applica
         return false;
     }
 
+    protected getSashHoverDelay(): void {
+        const sashHoverDelay = this.corePreferences.get(SASH_DELAY_PREFERENCE);
```
vince-fugnitto, line 861: The variables would not be necessary if we update the following to:

```suggestion
        const sashHoverDelay = this.corePreferences['workbench.sash.hoverDelay'];
```

```ts
const SASH_DELAY_PREFERENCE = 'workbench.sash.hoverDelay';
```

vince-fugnitto (commented):
I noticed that there is a flicker when resizing with the `sash`, something which is not present in vscode and can likely be improved upon:

_theia_:

https://user-images.githubusercontent.com/40359487/142873666-2a419f20-3773-409a-9007-8bd593268160.mp4

_vscode_:

https://user-images.githubusercontent.com/40359487/142873709-5729b096-2750-403f-9563-65fd1ffd0227.mp4

Also, I noticed on startup that the sash is quite small (I never updated the preference value), and then it becomes bigger later, perhaps something to try out for yourself.

packages/core/src/browser/view-container.ts
```diff
@@ -172,6 +179,19 @@ export class ViewContainer extends BaseWidget implements StatefulWidget, Applica
         if (this.options.progressLocationId) {
             this.toDispose.push(this.progressBarFactory({ container: this.node, insertMode: 'prepend', locationId: this.options.progressLocationId }));
         }
+
+        this.getSashHoverDelay();
```
kenneth-marut-work, line 183: it might be a good idea to wait for the `corePreferences.ready` promise before attempting to retrieve these preferences on start

packages/core/src/browser/style/view-container.css
```diff
@@ -38,14 +38,24 @@
...
+    min-width: 2px;
 }
 
 .theia-view-container > .p-SplitPanel > .p-SplitPanel-handle {
     background-color: var(--theia-sideBarSectionHeader-border);
 }
 
+.p-SplitPanel > .p-SplitPanel-handle:hover:after {
+    background-color: var(--theia-sash-hoverBorder);
+    transition-delay: var(--theia-sash-hover-delay);
```
kenneth-marut-work, line 51:
```suggestion
    transition-delay: var(--theia-sash-hoverDelay);
```
for consistency

packages/core/src/browser/view-container.ts
```diff
@@ -837,6 +857,20 @@ export class ViewContainer extends BaseWidget implements StatefulWidget, Applica
...
+            handle.style.setProperty('--theia-sash-hover-delay', `${sashHoverDelay}ms`);
+        });
+    }
+
+    protected getSashSize(): void {
+        const sashSize = this.corePreferences.get(SASH_SIZE_PREFERENCE);
+        document.querySelectorAll<HTMLElement>('.p-SplitPanel-handle').forEach(handle => {
+            handle.style.setProperty('--theia-sash-width', `${sashSize}px`);
+        });
+    }
```
colin-grant-work, line 872: It should be possible to find where this value is defined in the style in the header and set it once there, rather than setting it separately for all the handles. Alternatively, the rules that set this width could be managed in a dynamically created `style` element, and not written into a CSS file at all.

packages/core/src/browser/style/view-container.css
```diff
@@ -38,14 +38,24 @@
...
+    min-width: var(--theia-sash-width);
```
kenneth-marut-work, line 52: Minor: there's a very short fade out transition when hovering: The style in VSCode looks like this:
`transition: background-color .1s ease-out`

kaiyue0329:
@vince-fugnitto Thanks for the feedback! I've pushed changes that addresses your comments.

> I noticed that there is a flicker when resizing with the `sash`, something which is not present in vscode and can likely be improved upon

I've pushed up changes that should remove the flicker when resizing the `sash`.

https://user-images.githubusercontent.com/23107734/143143842-ad82375f-e418-4dec-9f92-522ea0998271.mov



> Also, I noticed on startup that the sash is quite small (I never updated the preference value), and then it becomes bigger later, perhaps something to try out for yourself.

After some testings, I noticed a bug with my previous change. When an editor widget is focused, the sash size becomes smaller by a certain ratio (compared to when the left panel is in focused). When the left panel gains focus again, the sash size increases to the value defined in the preference. This bug is fixed in the latest commit.

packages/core/src/browser/core-preferences.ts
kaiyue0329: @vince-fugnitto I have updated the values and defaults to align with the behavior present in vscode.

I kept the default value for `sash.hoverDelay` to be 0 because it solves resizing latency issue mentioned in #10416.

packages/core/src/browser/view-container.ts
kaiyue0329: @colin-grant-work Thanks for the feedback! Since the `sash`/ handles are defined as a pseudo element (**::after**) in CSS, I was only able to change their UI through TypeScript using CSS Custom Properties. I was wondering if there is a better approach for making UI changes to the pseudo elements.

packages/core/src/browser/view-container.ts
kaiyue0329: @kenneth-marut-work Thanks for the feedback! I have pushed changes that addresses your comments.

colin-grant-work (commented):
@kaiyue0329, for some reason GitHub isn't letting me reply to our existing conversation, but one way to manage a dynamic style element would be like [this](https://github.com/colin-grant-work/theia/commit/5cc4715e4acd514dcdec7f2c65f248c6614c99f5). That would give you a central location to update the value, rather than requiring you to iterate through all of the individual elements.

I also notice that in VSCode, the hover shows in between groups if you have multiple editor groups, but that doesn't appear to be the case with this code:

image

kaiyue0329:
> @kaiyue0329, for some reason GitHub isn't letting me reply to our existing conversation, but one way to manage a dynamic style element would be like [this](https://github.com/colin-grant-work/theia/commit/5cc4715e4acd514dcdec7f2c65f248c6614c99f5). That would give you a central location to update the value, rather than requiring you to iterate through all of the individual elements.
> 
> I also notice that in VSCode, the hover shows in between groups if you have multiple editor groups, but that doesn't appear to be the case with this code:

Thank you @colin-grant-work! I have pushed changes that address your comments. I also updated the screencast in the PR description to include the procedure for testing `sash` for muliple edior groups.

msujew (commented):
It seems like the flickering Vince experienced on Chrome is fixed. However, it still exists on Firefox:

https://user-images.githubusercontent.com/4377073/143436200-df731dfd-27ab-435b-b103-08adf04631c8.mp4

kenneth-marut-work:
@kaiyue0329 looking really nice so far. I noticed one small functionality issue: 
- In a VSCode view container, the sash does not appear if the viewcontainer is collapsed (or not-draggable). This is really more of an issue if the sash is set to its maximum size since it can actually occlude the mouse target for the expand/collapse

| VSCode | This Branch |
| ----------- | ----------- |
| collapsed-vscode | theia-blocking |

- This also has the side effect of causing 2 sashes to appear if you attempt to click and drag while the view container is collapsed:
drag-and-hover

kaiyue0329:
> It seems like the flickering Vince experienced on Chrome is fixed. However, it still exists on Firefox:

@msujew Thank you for the feedback. I have pushed up changes that should remove the flickering in firefox.

kaiyue0329:
> * In a VSCode view container, the sash does not appear if the viewcontainer is collapsed (or not-draggable). This is really more of an issue if the sash is set to its maximum size since it can actually occlude the mouse target for the expand/collapse
> * This also has the side effect of causing 2 sashes to appear if you attempt to click and drag while the view container is collapsed:

@kenneth-marut-work Thank you for the feedback. I have made changes to the code so that a sash in the left panel only appears under certain conditions (please see the updated PR description for more details).

kaiyue0329:
@vince-fugnitto @colin-grant-work This PR is ready for another round of review.

colin-grant-work:
@vince-fugnitto points out that there are already some reasonable utilities for creating and manipulating dynamic style sheets [here](https://github.com/eclipse-theia/theia/blob/master/packages/editor/src/browser/decorations/editor-decoration-style.ts#L38-L91). Would you be interested in taking that code and moving it to `core` (and perhaps refactoring a bit to make it more generic) where it can be used in the different different places that currently use their own custom solution to this problem?

kaiyue0329:
> @vince-fugnitto points out that there are already some reasonable utilities for creating and manipulating dynamic style sheets [here](https://github.com/eclipse-theia/theia/blob/master/packages/editor/src/browser/decorations/editor-decoration-style.ts#L38-L91). Would you be interested in taking that code and moving it to `core` (and perhaps refactoring a bit to make it more generic) where it can be used in the different different places that currently use their own custom solution to this problem?

@colin-grant-work Yes, I will do that!

packages/core/src/browser/core-preferences.ts
```diff
@@ -136,6 +136,20 @@ export const corePreferenceSchema: PreferenceSchema = {
             default: isOSX ? 1500 : 500,
             description: nls.localizeByDefault('Controls the delay in milliseconds after which the hover is shown.')
         },
+        'workbench.sash.hoverDelay': {
+            type: 'number',
+            default: 0,
```
msujew, line 141: Looking at the [VSCode codebase](https://github.com/microsoft/vscode/blob/2085b19a51fcab47f4991f51ae8f9eda7923c1d7/src/vs/workbench/contrib/sash/browser/sash.contribution.ts#L35), I noticed that they use a delay of `300`. I think we should align to that.

packages/core/src/browser/decoration-style.ts
```diff
@@ -0,0 +1,61 @@
...
+        if (!styleSheet) {
+            return;
+        }
+        const rules = styleSheet.cssRules || styleSheet.rules;
+        for (let i = 0; i < rules.length; i++) {
+            if (rules[i].type === CSSRule.STYLE_RULE) {
+                if ((rules[i] as CSSStyleRule).selectorText === selector) {
+                    styleSheet.removeRule(i);
+                }
+            }
```
colin-grant-work, line 57: Combine conditions and replace `removeRule` (deprecated) with `deleteRule`.
```suggestion
            if (rules[i].type === CSSRule.STYLE_RULE && (<CSSStyleRule>rules[i]).selectorText === selector) {
                styleSheet.deleteRule(i);
            }
```

packages/core/src/browser/decoration-style.ts
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        const index = styleSheet.insertRule(selector + '{}', 0);
+        const rules = styleSheet.cssRules || styleSheet.rules;
+        const rule = rules[index];
+        if (rule && rule.type === CSSRule.STYLE_RULE) {
+            const styleRule = rule as CSSStyleRule;
+            styleProvider(styleRule.style);
+        }
+    }
```
colin-grant-work, line 45: A few comments on this method:
 - If you insert rules at index 0, they have the lowest precedence, so adding tow rules for the same selector and rule would result in the earlier rule overriding the later, which probably isn't intended. There are two ways out:
   - figure out the next index and insert at the end of the list.
   - search the list for rules with the same selector and work with the existing rule. This would simplify the `deleteRule` logic a bit, since you would know that (unless the caller did something weird with their style sheet), there would be only one rule for a given selector, and you could bail if/when you found it.
 - The `styleProvider` callback is weird. I would probably just return `CSSStyleRule` and let the caller do what they want with it rather than having them pass in a callback.

kaiyue0329: Thank you @colin-grant-work for the very detailed explanation. I updated the code for `createRule` (renamed to `createStyleRule`) and `deleteRule` (renamed to `deleteStyleRule` to avoid confusion with the other `deleteRule` that exists in the code) so that there is only one rule for a given selector.

I also updated `editor-decoration-style.ts` and removed the `styleProvider` callback based on your comment.

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -288,6 +289,8 @@ export namespace CommonCommands {
 
 }
 
+const commonDecorationsStyleSheet: CSSStyleSheet | undefined = DecorationStyle.createStyleSheet();
```
colin-grant-work, line 292: Rather than placing this in module scope, I think it would be better to have a field on the `FrontendContribution`, instead. That will allow adopters to manipulate with extensions to this class.

packages/core/src/browser/decoration-style.ts
```diff
@@ -0,0 +1,72 @@
...
+ ********************************************************************************/
+
+export namespace DecorationStyle {
+
+    export function createStyleSheet(container: HTMLElement = document.getElementsByTagName('head')[0]): CSSStyleSheet | undefined {
+        if (!container) {
+            return undefined;
+        }
+        const style = document.createElement('style');
+        style.id = 'editorDecorationsStyle';
```
colin-grant-work, line 24: This should probably be an argument, since multiple contributions will want to make their own style sheets and they shouldn't all have the same ID.

packages/core/src/browser/decoration-style.ts
```diff
@@ -0,0 +1,72 @@
...
+        return <CSSStyleSheet>style.sheet;
+    }
+
+    function getRuleIndex(selector: string, styleSheet: CSSStyleSheet): number | undefined {
+        const rules = styleSheet.cssRules || styleSheet.rules;
+        for (let i = 0; i < rules.length; i++) {
+            if (rules[i].type === CSSRule.STYLE_RULE && (<CSSStyleRule>rules[i]).selectorText === selector) {
+                return i;
+            }
+        }
```
colin-grant-work, line 38:
```suggestion
        return Array.from(styleSheet.cssRules || styleSheet.rules).findIndex(rule => rule.type === CSSRule.STYLE_RULE && (<CSSStyleRule>rule).selectorText === selector);
```

packages/core/src/browser/decoration-style.ts
```diff
@@ -0,0 +1,72 @@
...
+        if (!index) {
+            // The given selector does not exist in the provided independent style sheet, rule index = 0
+            index = styleSheet.insertRule(selector + '{}', 0);
+        }
+
+        const rules = styleSheet.cssRules || styleSheet.rules;
+        const rule = rules[index];
+        if (rule.type === CSSRule.STYLE_RULE) {
+            const styleRule = rule as CSSStyleRule;
+            return styleRule;
```
colin-grant-work, line 56:
```suggestion
            return rule as CSSStyleRule;
```

packages/core/src/browser/decoration-style.ts
```diff
@@ -0,0 +1,72 @@
...
+    function getRuleIndex(selector: string, styleSheet: CSSStyleSheet): number | undefined {
+        const rules = styleSheet.cssRules || styleSheet.rules;
+        for (let i = 0; i < rules.length; i++) {
+            if (rules[i].type === CSSRule.STYLE_RULE && (<CSSStyleRule>rules[i]).selectorText === selector) {
+                return i;
+            }
+        }
+    }
+
+    export function createStyleRule(selector: string, styleSheet: CSSStyleSheet | undefined): CSSStyleRule | void {
```
colin-grant-work, line 41:
```suggestion
    export function getOrCreateStyleRule(selector: string, styleSheet: CSSStyleSheet | undefined): CSSStyleRule | undefined {
```

packages/editor/src/browser/decorations/editor-decoration-style.ts
```diff
@@ -15,77 +15,27 @@
...
 
 import { Disposable } from '@theia/core';
+import { DecorationStyle } from '@theia/core/lib/browser';
 
 export class EditorDecorationStyle implements Disposable {
 
     constructor(
         readonly selector: string,
         styleProvider: (style: CSSStyleDeclaration) => void,
+        private decorationsStyleSheet: CSSStyleSheet | undefined
```
colin-grant-work, line 25:
```suggestion
        protected decorationsStyleSheet: CSSStyleSheet | undefined
```

packages/git/src/browser/blame/blame-decorator.ts
```diff
@@ -20,6 +20,9 @@ import { GitFileBlame } from '../../common';
 import { Disposable, DisposableCollection } from '@theia/core';
 import * as moment from 'moment';
 import URI from '@theia/core/lib/common/uri';
+import { DecorationStyle } from '@theia/core/lib/browser';
+
+const blameDecorationsStyleSheet: CSSStyleSheet | undefined = DecorationStyle.createStyleSheet();
```
colin-grant-work, line 25: Here as well, please make this a field on the class so that it can be accessed by extenders.

packages/core/src/browser/decoration-style.ts
```diff
@@ -0,0 +1,72 @@
...
+        if (!styleSheet) {
+            return;
+        }
+
+        let index = getRuleIndex(selector, styleSheet);
+        if (!index) {
```
colin-grant-work, line 47: If  you adopt the `Array.from` suggestion above, this should be `index !== -1`. If not, this should be `index !== undefined` since `!0` would return `true`, but `0` is a real index, so we wouldn't need to create a new rule.

packages/core/src/browser/view-container.ts
```diff
@@ -46,6 +46,11 @@ export interface ViewContainerTitleOptions {
     closeable?: boolean;
 }
 
+enum SashType {
+    topSash = 'top-sash',
+    bottomSash = 'bottom-sash'
```
kenneth-marut-work, line 51: minor, [use PascalCase](https://github.com/eclipse-theia/theia/wiki/Coding-Guidelines#pascalcase-enum) for enum vals

kenneth-marut-work (commented):
@kaiyue0329 once you've addressed the remaining comments I'll give this another test to make sure the functionality is in place and there are no issues. I think the `EditorDecorationStyle`  class & `DecorationStyle` namespace could benefit from some refactoring after this is merged since it's a bit hard to follow. But that's outside the scope of this PR.

packages/core/src/browser/decoration-style.ts
```diff
@@ -0,0 +1,66 @@
...
+        if (index === -1) {
+            // The given selector does not exist in the provided independent style sheet, rule index = 0
+            index = styleSheet.insertRule(selector + '{}', 0);
+        }
+
+        const rules = styleSheet.cssRules || styleSheet.rules;
+        const rule = rules[index];
+        if (rule && rule.type === CSSRule.STYLE_RULE) {
+            return rule as CSSStyleRule;
+        }
```
colin-grant-work, line 51: Rather than having `undefined` as an option for the `return`, let's `throw` an error if this check fails.

packages/core/src/browser/decoration-style.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    function getRuleIndex(selector: string, styleSheet: CSSStyleSheet): number {
+        return Array.from(styleSheet.cssRules || styleSheet.rules).findIndex(rule => rule.type === CSSRule.STYLE_RULE && (<CSSStyleRule>rule).selectorText === selector);
+    }
+
+    export function getOrCreateStyleRule(selector: string, styleSheet: CSSStyleSheet | undefined): CSSStyleRule | undefined {
+        if (!styleSheet) {
+            return;
+        }
```
colin-grant-work, line 39:
```suggestion
    export function getOrCreateStyleRule(selector: string, styleSheet: CSSStyleSheet): CSSStyleRule {
```

packages/core/src/browser/decoration-style.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+export namespace DecorationStyle {
+
+    export function createStyleSheet(styleId: string, container: HTMLElement = document.getElementsByTagName('head')[0]): CSSStyleSheet | undefined {
+        if (!container) {
+            return undefined;
+        }
```
colin-grant-work, line 22: The `container` should never be `undefined` given the current fallback.
```suggestion
    export function createStyleSheet(styleId: string, container: HTMLElement = document.getElementsByTagName('head')[0]): CSSStyleSheet {
```

packages/git/src/browser/blame/blame-decorator.ts
```diff
@@ -20,10 +20,27 @@ import { GitFileBlame } from '../../common';
...
+        }, this.blameDecorationsStyleSheet);
+
+        new EditorDecorationStyle('git-blame-continuation-line::before', style => {
+            style.content = "'\u2007'"; // blank
+        }, this.blameDecorationsStyleSheet);
+
+        new EditorDecorationStyle('git-blame-continuation-line::after', style => {
+            style.content = "'\u2007'";
+        }, this.blameDecorationsStyleSheet);
+    }
```
colin-grant-work, line 42: If you remove the `undefined` from the return options of `DecorationStyle.createStyleSheet` and `getOrCreateStyleRule`, you can shorten this constructor to this:
```typescript
    constructor(
        protected blameDecorationsStyleSheet: CSSStyleSheet = DecorationStyle.createStyleSheet('gitBlameDecorationsStyle')
    ) {
        DecorationStyle.getOrCreateStyleRule(`.${BlameDecorator.GIT_BLAME_HIGHLIGHT}`, this.blameDecorationsStyleSheet).style.backgroundColor = 'var(--theia-gitlens-lineHighlightBackgroundColor)';
        DecorationStyle.getOrCreateStyleRule(`.${BlameDecorator.GIT_BLAME_CONTINUATION_LINE}::before`, this.blameDecorationsStyleSheet).style.content = "'\u2007'"; // blank
        DecorationStyle.getOrCreateStyleRule(`.${BlameDecorator.GIT_BLAME_CONTINUATION_LINE}::after`, this.blameDecorationsStyleSheet).style.content = "'\u2007'"; // blank;
    }
```
Since we don't retain a reference to these `EditorDecorationStyle`s instances, there's no point in creating them - we might as well just insert the rules.

colin-grant-work:
A couple of visual and functional oddities:
In `Dark (Visual Studio)` theme, the sash appears as a grey bar between ViewContainer parts when not hovered on:  
image
It may be that the color variable that's being applied to the `::after` element is not quite right.

(That's at 20 pixels)

It also appears that the Git Blame annotation styling is not correct:

Master:  
image

Now:  
image

colin-grant-work (approved):
The sash functionality is working well in the contexts where it's expected to appear, and the problems with the git decorations and odd sash appearance have been resolved.

kenneth-marut-work (approved):
I've tested this again and confirm that it's working as expected and confirmed that the blame CSS regressions have been fixed as well as the theming issue

PR #10455, opened 2021-11-24
core: improve Deferred typings

paul-marechal:
For some reason `Deferred` objects allow you to resolve them with
`undefined` even though the corresponding promise only mentions `T`.

Remove the optional `value` when resolving.

Default to `void` if `T` is unspecified.

#### How to test

Everything should build and work as before.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,14 @@
 # Change Log
 
-## v1.19.0 - 11/25/2021
+## v1.20.0 - 11/25/2021
 
-[1.20.0 Milestone](https://github.com/eclipse-theia/theia/milestone/26)
+[1.20.0 Milestone](https://github.com/eclipse-theia/theia/milestone/28)
 
 <a name="breaking_changes_1.20.0">[Breaking Changes:](#breaking_changes_1.20.0)</a>
 
-- [plugindev] Renamed HostedPlugin(Server|Client) to PluginDev(Server|Client)
-- [core] `value` is now not optional in `Deferred<T>.resolve(value: T)`
-- [core] `T` defaults to `void` if not specified when defining a `Deferred`
+- [plugindev] Renamed HostedPlugin(Server|Client) to PluginDev(Server|Client) [#10352](https://github.com/eclipse-theia/theia/pull/10352)
+- [core] `value` is now not optional in `Deferred<T>.resolve(value: T)` [#10455](https://github.com/eclipse-theia/theia/pull/10455)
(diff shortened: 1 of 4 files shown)


paul-marechal:
I noticed this oddity while reviewing https://github.com/eclipse-theia/theia/pull/10333#discussion_r755460572

We shouldn't be able to resolve with `undefined` if `T` doesn't include it explicitly.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
+
+## v1.19.0 - 11/25/2021
+
+[1.20.0 Milestone](https://github.com/eclipse-theia/theia/milestone/26)
+
+<a name="breaking_changes_1.20.0">[Breaking Changes:](#breaking_changes_1.20.0)</a>
+
 - [plugindev] Renamed HostedPlugin(Server|Client) to PluginDev(Server|Client)
+- [core] `value` is now not optional in `Deferred<T>.resolve(value: T)`
+- [core] `T` defaults to `void` if not specified when defining a `Deferred`
```
colin-grant-work, line 11: I don't _think_ this should count as a breaking change, since previously you _had_ to specify the type in a generic, and now you don't if you mean `void`, but no one's existing generics should be invalidated by that change?

colin-grant-work (approved):
Definitely doesn't make any sense to allow `undefined`.

CHANGELOG.md
vince-fugnitto, line 11: minor - I can also update before the release:

```suggestion
## v1.20.0 - 11/25/2021

[1.20.0 Milestone](https://github.com/eclipse-theia/theia/milestone/28)

<a name="breaking_changes_1.20.0">[Breaking Changes:](#breaking_changes_1.20.0)</a>

- [plugindev] Renamed HostedPlugin(Server|Client) to PluginDev(Server|Client) [#10352](https://github.com/eclipse-theia/theia/pull/10352)
- [core] `value` is now not optional in `Deferred<T>.resolve(value: T)` [#10455](https://github.com/eclipse-theia/theia/pull/10455)
- [core] `T` defaults to `void` if not specified when defining a `Deferred` [#10455](https://github.com/eclipse-theia/theia/pull/10455)
```

paul-marechal: I noticed a few places where `T` was omitted but TS default used to be `any` in that case, now it is `unknown`, and only by having `void` can we call `.resolve()` without a value... So I thought it would be worth mentioning that we default to void now.

colin-grant-work: Never hurts to provide more information. After I mentioned it, I was working in some Deferred-heavy areas and noticed the same thing. 👍

Issue #10468, opened 2021-11-26
Completion not properly working for Debug Console Sessions


martin-fleck-at:
### Bug Description:
For each debug console session we register a completion item provider that may query the debug adapter for completion items. However, the item provider is never disposed and always active for all sessions. The result is that you get an unlimited 'Loading' indicator as we are waiting for suggestions from sessions which may no longer be active. Furthermore, even if we had multiple sessions where one session uses the editor suggestions and the other does not, we would always get the set of all suggestions.

### Steps to Reproduce:

1. Install Java Debugger
2. Create two simple Java programs with one variable to test the completion capabilities
3. Run both debugging sessions at once and try to use completion in the Debug Console

![java_notLoading](https://user-images.githubusercontent.com/19170971/143573252-73f13bfb-e5c3-4963-be38-c13ded3c81bf.gif)

### Additional Information

- Operating System: Linux
- Theia Version: 1.20

PR #9613, opened 2021-06-19
Add vscode API `DebugSessionOptions`

msujew:
#### What it does
Closes #8518 

Adds the `vscode.debug.DebugSessionOptions` API to Theia. This enables the following features:

- Show child debug sessions as a tree under the parent sessions (Theia used to show them as siblings)
    - The new `compact` property will simplify the label and tree structure if possible
- Different debug sessions no longer share the same `DebugConsoleSession` and therefore are displayed independently
    - Child debug sessions with the `consoleMode` property set to `MergeWithParent` will continue to share their console session with each other.
    - You can switch between console sessions with a selector, assuming more than one session is currently running.

The changes to `DebugConsoleSession` are slightly breaking. While it's still possible to simply inject the `DebugConsoleSession` directly, its tight coupling to `DebugSession` makes that impractical. Instead of injecting the `DebugConsoleSession` directly, a `DebugConsoleSessionFactory` allows to create a `DebugConsoleSession` with a provided `DebugSession`.

#### How to test

1. Clone [this test repo](https://github.com/msujew/debug-session-options-example) into Theia and make sure the `vscode-js-debug` extension is installed in Theia.
2. Start a debugging session with the contained `Launch Program` configuration.
3. Child processes will now be shown as a tree in the threads view.
4. Start a second debugging session with the same configuration
5. Another tree will appear in the threads view and you can switch between the different debug consoles.
6. You can add `configuration.consoleMode = undefined;` and/or `configuration.compact = undefined;` after [this line](https://github.com/msujew/theia/blob/e26d3bff923e51f013aece29ed7a9db8eb27efce/packages/plugin-ext/src/main/browser/debug/debug-main.ts#L273), and restart the server to test how these properties influence the debug console/view.

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #10469, opened 2021-11-26
Properly handle debug console completion

### LINKED ISSUE #10648: console: unable to execute expressions during a debug session
#### Bug Description:

Executing expressions during a debug session in the debug console does not work.
The following logs are present:

```
root ERROR [hosted-plugin: 7831] Unknown request: completions
root ERROR [hosted-plugin: 7831] Unknown request: completions
root ERROR [hosted-plugin: 7831] Unknown request: evaluate
root ERROR [hosted-plugin: 7831] Unknown request: completions
root ERROR [hosted-plugin: 7831] Unknown request: completions
root ERROR [hosted-plugin: 7831] Unknown request: evaluate
```

_theia_: executing expressions does not work

https://user-images.githubusercontent.com/40359487/149970416-a91af77a-902a-4164-8154-bea4bedf3f30.mp4

_vscode_:

https://user-images.githubusercontent.com/40359487/149970522-117b88ed-e0ef-46ad-9891-15967adf43ab.mp4

#### Steps to Reproduce:

1. start a debug session (ex: debugging mocha tests) and hit a breakpoint
2. attempt to execute the expression `console.log("hello")` in the debug console - nothing happens

martin-fleck-at:
#### What it does
- Ensure console session is disposed when debug session is destroyed
- Remove  monaco completion item provider on disposal
- Ensure we only provide completion items for our own session
- Sync debug console with the currently selected session

java_fix

Fixes https://github.com/eclipse-theia/theia/issues/10468
Fixes #10648 

#### How to test
1. Install Java Debugger
2. Create two simple Java programs with one variable to test the completion capabilities
3. Run both debugging sessions at once and try to use completion in the Debug Console
4. Switch between debug session consoles to see if completion items are available and only provided for the selected session

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/console/debug-console-contribution.tsx
+++ b/packages/debug/src/browser/console/debug-console-contribution.tsx
@@ -24,9 +24,8 @@ import { Command, CommandRegistry } from '@theia/core/lib/common/command';
 import { Severity } from '@theia/core/lib/common/severity';
 import { inject, injectable, interfaces, postConstruct } from '@theia/core/shared/inversify';
 import * as React from '@theia/core/shared/react';
-import { DebugConsoleMode } from '../../common/debug-configuration';
 import { DebugSession } from '../debug-session';
-import { DebugSessionManager } from '../debug-session-manager';
+import { DebugSessionManager, DidChangeActiveDebugSession } from '../debug-session-manager';
 import { DebugConsoleSession, DebugConsoleSessionFactory } from './debug-console-session';
 
 export type InDebugReplContextKey = ContextKey<boolean>;
@@ -71,9 +70,9 @@ export class DebugConsoleContribution extends AbstractViewContribution<ConsoleWi
     @postConstruct()
     protected init(): void {
         this.debugSessionManager.onDidCreateDebugSession(session => {
-            const topParent = this.findParentSession(session);
-            if (topParent) {
-                const parentConsoleSession = this.consoleSessionManager.get(topParent.id);
+            const consoleParent = session.findConsoleParent();
+            if (consoleParent) {
+                const parentConsoleSession = this.consoleSessionManager.get(consoleParent.id);
                 if (parentConsoleSession instanceof DebugConsoleSession) {
                     session.on('output', event => parentConsoleSession.logOutput(parentConsoleSession.debugSession, event));
                 }
@@ -83,20 +82,24 @@ export class DebugConsoleContribution extends AbstractViewContribution<ConsoleWi
                 session.on('output', event => consoleSession.logOutput(session, event));
             }
         });
+        this.debugSessionManager.onDidChangeActiveDebugSession(event => this.handleActiveDebugSessionChanged(event));
         this.debugSessionManager.onDidDestroyDebugSession(session => {
             this.consoleSessionManager.delete(session.id);
         });
     }
 
-    protected findParentSession(session: DebugSession): DebugSession | undefined {
-        if (session.configuration.consoleMode !== DebugConsoleMode.MergeWithParent) {
-            return undefined;
+    protected handleActiveDebugSessionChanged(event: DidChangeActiveDebugSession): void {
+        if (!event.current) {
+            this.consoleSessionManager.selectedSession = undefined;
+        } else {
+            const topSession = event.current.findConsoleParent() || event.current;
+            const consoleSession = topSession ? this.consoleSessionManager.get(topSession.id) : undefined;
+            this.consoleSessionManager.selectedSession = consoleSession;
+            const consoleSelector = document.getElementById('debugConsoleSelector');
+            if (consoleSession && consoleSelector instanceof HTMLSelectElement) {
+                consoleSelector.value = consoleSession.id;
+            }
         }
-        let debugSession: DebugSession | undefined = session;
-        do {
-            debugSession = debugSession.parentSession;
-        } while (debugSession?.parentSession && debugSession.configuration.consoleMode === DebugConsoleMode.MergeWithParent);
-        return debugSession;
     }
 
     registerCommands(commands: CommandRegistry): void {
--- a/packages/debug/src/browser/console/debug-console-session.ts
+++ b/packages/debug/src/browser/console/debug-console-session.ts
@@ -22,7 +22,8 @@ import { DebugSession } from '../debug-session';
 import URI from '@theia/core/lib/common/uri';
 import { ExpressionContainer, ExpressionItem } from './debug-console-items';
 import { Severity } from '@theia/core/lib/common/severity';
-import { injectable, postConstruct } from '@theia/core/shared/inversify';
+import { inject, injectable, postConstruct } from '@theia/core/shared/inversify';
+import { DebugSessionManager } from '../debug-session-manager';
 
 export const DebugConsoleSessionFactory = Symbol('DebugConsoleSessionFactory');
 
@@ -32,6 +33,8 @@ export type DebugConsoleSessionFactory = (debugSession: DebugSession) => DebugCo
 export class DebugConsoleSession extends ConsoleSession {
 
     static uri = new URI().withScheme('debugconsole');
+
+    @inject(DebugSessionManager) protected readonly sessionManager: DebugSessionManager;
 
     protected items: ConsoleItem[] = [];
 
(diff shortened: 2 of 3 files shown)


packages/debug/src/browser/console/debug-console-session.ts
```diff
@@ -23,10 +23,11 @@ import URI from '@theia/core/lib/common/uri';
 import { ExpressionContainer, ExpressionItem } from './debug-console-items';
 import { Severity } from '@theia/core/lib/common/severity';
 import { injectable, postConstruct } from '@theia/core/shared/inversify';
+import { ConsoleSessionManager } from '@theia/console/src/browser/console-session-manager';
```
vince-fugnitto, line 26: @martin-fleck-at importing from `src/` will cause the build to fail:
```suggestion
import { ConsoleSessionManager } from '@theia/console/lib/browser/console-session-manager';
```

martin-fleck-at: Ah, quick fix import from vscode got me again! Thanks!

msujew (commented):
> Sync debug console with the currently selected session

It seems to me like this isn't working correctly. When clicking on another thread in the threads view (using the JS debugger), the select element gets blanked out. Only when clicking on the actual root debug session the correct option is selected:

https://user-images.githubusercontent.com/4377073/143893470-d4fa1ff4-08a3-4ca6-8457-ed53a18fab62.mp4

The issue is probably related to the changes I did in https://github.com/eclipse-theia/theia/pull/9613. Due to these changes, some debug sessions share the same console session while others don't. This has to be reflected in the selection logic.

packages/debug/src/browser/console/debug-console-contribution.tsx
```diff
@@ -78,13 +78,22 @@ export class DebugConsoleContribution extends AbstractViewContribution<ConsoleWi
...
+            this.consoleSessionManager.selectedSession = consoleSession;
+            const consoleSelector = document.getElementById('debugConsoleSelector');
+            if (session.current && consoleSelector) {
+                (consoleSelector as HTMLSelectElement).value = session.current.id;
+            }
+        });
         this.debugSessionManager.onDidDestroyDebugSession(session => {
-            this.consoleSessionManager.delete(session.id);
+            const consoleSession = this.consoleSessionManager.delete(session.id);
+            consoleSession?.dispose();
```
msujew, line 96: Suggestion: `consoleSessionManager` should call `dispose` on deleted console sessions. I can't envision a case where we want to delete a console session without disposing of it as well. We can revert the changes in this callback afterwards.

packages/debug/src/browser/console/debug-console-session.ts
```diff
@@ -42,6 +43,8 @@ export class DebugConsoleSession extends ConsoleSession {
 
     protected readonly completionKinds = new Map<DebugProtocol.CompletionItemType | undefined, monaco.languages.CompletionItemKind>();
 
+    sessionManager: ConsoleSessionManager;
```
colin-grant-work, line 46: You should be able to change this to `@inject(ConsoleSessionManager) protected readonly sessionManager: ConsoleSessionManager;` and then not change the interface for the `DebugConsoleSessionFactory` - the `ConsoleSessionManager` will be provided automatically by the call to `context.container.get(DebugConsoleSession)`.

packages/debug/src/browser/console/debug-console-contribution.tsx
martin-fleck-at: You are right, I did not properly consider the parent relationship between sessions. I adapted the code to reflect that, thank you!

packages/debug/src/browser/console/debug-console-session.ts
martin-fleck-at: Good catch! That was an unnecessary change indeed, I fixed it, thank you!

packages/debug/src/browser/console/debug-console-contribution.tsx
martin-fleck-at: Update: PR https://github.com/eclipse-theia/theia/pull/10333 actually did something similar so that is now resolved on master already.

packages/debug/src/browser/console/debug-console-contribution.tsx
```diff
@@ -237,3 +241,14 @@ export class DebugConsoleContribution extends AbstractViewContribution<ConsoleWi
...
+
+export function findParentSession(session: DebugSession): DebugSession | undefined {
+    if (session.configuration.consoleMode !== DebugConsoleMode.MergeWithParent) {
+        return undefined;
+    }
+    let debugSession: DebugSession | undefined = session;
+    do {
+        debugSession = debugSession.parentSession;
+    } while (debugSession?.parentSession && debugSession.configuration.consoleMode === DebugConsoleMode.MergeWithParent);
+    return debugSession;
```
colin-grant-work, line 253: I know this logic is unchanged, but it's a little confusing to me. In the first check, if we find `consoleMode !== DebugConsoleMode.MergeWithParent`, we return `undefined`, but in the `while` loop, as soon as we find `consoleMode !== DebugConsoleMode.MergeWithParent`, we return the session.

Are we looking for the first session that doesn't merge with the parent, or are we looking for a session without a parent?

colin-grant-work (commented):
Functionally this seems to be working for me, though perhaps @msujew can test some of the intricacies of parent and child sessions, since he's worked on that recently.

Aside from that, a few comments on the implementation.

packages/debug/src/browser/console/debug-console-contribution.tsx
msujew: My take on this, since I wrote the original algorithm for finding the parent to merge with:

Basically we look for the top-most parent of the current session which still allows console merging, assuming the debug session in question actually wants to merge with it's parent (l. 246).

> Are we looking for the first session that doesn't merge with the parent, or are we looking for a session without a parent?

Both :) if the session has no parent, it has its own console session, which we merge our output into. If some debug session in the parent chain does not want to merge its output into one of its parents, it creates its own console session in which merge our output into instead.

colin-grant-work: So I guess my question is why we return `undefined` in the first case (l. 247). If it doesn't want to merge into its parent, isn't it effectively its own top-most parent, so shouldn't we return `session` there?

msujew: IIRC I structured the code in such a way that `undefined` is the signal to construct its own console session, instead of attaching to an existing one. It could just as well return `session`, but the calling code would have to differentiate whether the result is the input session - meaning it would have to construct a new session - or another, in which case we attach to its existing console session.

colin-grant-work: Sounds fair - thanks for the explanation.

packages/debug/src/browser/console/debug-console-contribution.tsx
```diff
@@ -237,3 +241,14 @@ export class DebugConsoleContribution extends AbstractViewContribution<ConsoleWi
...
+}
```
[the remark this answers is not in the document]
martin-fleck-at: I moved the method into the debug session and renamed it to `findConsoleParent()`.

packages/debug/src/browser/console/debug-console-contribution.tsx
```diff
@@ -83,20 +83,24 @@ export class DebugConsoleContribution extends AbstractViewContribution<ConsoleWi
...
+    protected activeDebugSessionChanged(event: DidChangeActiveDebugSession): void {
+        if (!event.current) {
+            this.consoleSessionManager.selectedSession = undefined;
+        } else {
+            const topSession = event.current.parentSession ? findParentSession(event.current) : event.current;
+            const consoleSession = topSession ? this.consoleSessionManager.get(topSession.id) : undefined;
+            this.consoleSessionManager.selectedSession = consoleSession;
+            const consoleSelector = document.getElementById('debugConsoleSelector');
+            if (consoleSession && consoleSelector) {
+                (consoleSelector as HTMLSelectElement).value = consoleSession.id;
```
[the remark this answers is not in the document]
martin-fleck-at: Done

packages/debug/src/browser/console/debug-console-contribution.tsx
martin-fleck-at: Thank you for elaborating on this!

martin-fleck-at:
@colin-grant-work Thank you very much for your feedback! I changed the things you mentioned and hopefully everything is now up to par.

colin-grant-work (commented):
The new changes look good to me, and it is now possible to get auto-completion and execution in the debug console. From that perspective, I would consider the PR complete. However, the functionality of the debug console still seems fairly broken, as described [here](https://github.com/eclipse-theia/theia/issues/10648). This PR addresses one of the logs mentioned there: `Unknown request: completions` is no longer logged, but I still see `Unknown request: evaluate` when entering code in a Mocha debugging session in the Theia repo. I would guess that the cause is similar to the problem here: we're not sending the request to the right session. Would you be interested in addressing that problem as part of this PR, or would you prefer to limit this PR to the completion problem and address execution separately?

martin-fleck-at:
@colin-grant-work Ah, I see what you mean and I had a quick look at it. You are right that we also used the wrong session for the request for the `evaluate` request. I pushed a new commit on top which should fix that.

colin-grant-work (approved):
With the latest changes, both completion and execution work as expected. 👍

JonasHelming:
@vince-fugnitto @msujew : Any final remarks, otherwise I would merge this

msujew (commented):
When using the node debugger, the completion does not to work correctly when manually triggering the completion (Ctrl+Space). I'm not completely sure what's wrong there:

https://user-images.githubusercontent.com/4377073/151803852-761bbfb1-7fd3-44e3-ad55-0e73933dfde9.mp4

While typing the completion seems to work correctly, but when triggering it manually, an empty completion box appears (visible at the end of the video). Also, when pressing Ctrl+Space without any input, only the name of the debug session appears.

martin-fleck-at:
@msujew Thank you for your fast response! I had a look at this and the box with the debug session name without any input is the expected behavior if you compare it with VS Code. As far as I can see, this is actually what the debug adapter returns.

As for the "empty" box, it turns out that this is a message box saying 'No suggestions.' that is just placed "badly". The VS Code suggestion widget shows suggestions above the input but messages like 'Loading' and 'No suggestions.' below. The big difference is that in VS Code the message box is actually properly rendered ABOVE the status bar at the bottom, so you can actually read the message. You can test the suggestion widget in Theia in any editor that supports suggestions. You'll see how the different things are rendered. So the problem we have is that there is not enough space at the bottom of our console views.

I tried looking for a pure CSS solution but unfortunately I could not find anything satisfying. The problem is that the `suggest-widget` provides the surrounding container (with border, background colors, sash handles etc.) while either showing the `.tree` child with the suggestions or the `.message` child with messages. We can query which one of those is visible but we cannot move the parent suggest widget based on that (like `margin: -50px`) as we do not have a proper parent selector.

Do you, by any chance, have any idea how we can achieve this without restructuring the DOM? Otherwise I'd probably open up a separate issue as it is unrelated to the actual functionality of completion. What do you think?

msujew (approved):
@martin-fleck-at Ha, interesting. You're right, VSCode also exhibits the same behavior. Seems to me like the node-debugger extension does not work as expected, but that's not on us to fix ;)

I don't have too much of an issue with the weirdly placed 'No suggestions' box, it just added to my already existing confusion.

So in that regard everything else works as expected and the code looks good to me 👍

JonasHelming:
@martin-fleck-at : Could you squash, I will merge then

martin-fleck-at:
@JonasHelming Thank you! I squashed and rebased the change. Now we only need to wait for the CI to run through.

Issue #10470, opened 2021-11-26
Notification expand/collapse is no longer shown


martin-fleck-at:
### Bug Description:

For messages longer than 500 characters the expand/collapse indicator is shown to now blow up the screen. However, there seems to be an issue with the class name asignment so at the moment it is there but invisible:

### Steps to Reproduce:

1. Produce a message over 500 characters (for example through a sample menu contribution)
2. See that the expand/collapse action is "missing"

![expandCollapse](https://user-images.githubusercontent.com/19170971/143581078-de1d9b9d-dfa3-46e4-93fd-00bc94e59cf1.png)

### Additional Information

- Operating System: Linux
- Theia Version: 1.20

PR #10078, opened 2021-09-10
Aligns icons of alerts and notifications to the text

msujew:
#### What it does

Closes #10066

Aligns the icons to the text and vscode's styling.

vscode notifications:

image

Theia's notifications:

image

Alert messages:

image

#### How to test

1. Delete all plugins/open a folder without a git repository
2. Observe the alerts
3. Observe Theia's notifications

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #10471, opened 2021-11-26
Properly render 'expand' and 'collapse' action on notification

martin-fleck-at:
#### What it does
- Fix assignment of class names
- Scale icons to better fit the 'close icon'
-- Remove hover background as due to the scaling it looks weird
-- Use scaling as indicator instead

fix_expandCollapse

Fixes https://github.com/eclipse-theia/theia/issues/10470

#### How to test

1. Produce a message over 500 characters (for example through a sample menu contribution)
2. See that the expand/collapse action is there and working

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/messages/src/browser/style/notifications.css
+++ b/packages/messages/src/browser/style/notifications.css
@@ -195,10 +195,6 @@
     transform: rotate(180deg);
 }
 
-.theia-notification-actions > .collapse {
-    transform: rotate(0deg);
-}
-
 .theia-notification-list-item-content-bottom {
     display: flex;
     flex-direction: row;
--- a/packages/messages/src/browser/style/notifications.css
+++ b/packages/messages/src/browser/style/notifications.css
@@ -192,11 +192,11 @@
 }
 
 .theia-notification-actions > .expand {
-    transform: rotate(0deg);
+    transform: rotate(180deg);
 }
 
 .theia-notification-actions > .collapse {
-    transform: rotate(-90deg);
+    transform: rotate(0deg);
 }
 
 .theia-notification-list-item-content-bottom {
--- a/packages/messages/src/browser/style/notifications.css
+++ b/packages/messages/src/browser/style/notifications.css
@@ -191,27 +191,17 @@
     margin-left: 4px;
 }
 
-.theia-notification-actions > li:hover {
-    /* remove default background change as due to the different scaling of the items it might look odd */ 
-    background-color: transparent;
-    transform: scale(1.5);
-}
-
 .theia-notification-actions > .expand {
     transform: rotate(0deg) scale(1.35);
 }
-
-.theia-notification-actions > .expand:active,
-.theia-notification-actions > li.expand:hover {
+.theia-notification-actions > .expand:active {
     transform: rotate(0deg) scale(1.8);
 }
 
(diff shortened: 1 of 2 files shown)


packages/messages/src/browser/style/notifications.css
```diff
@@ -191,17 +191,27 @@
     margin-left: 4px;
 }
 
+.theia-notification-actions > li:hover {
+    /* remove default background change as due to the different scaling of the items it might look odd */ 
+    background-color: transparent;
+    transform: scale(1.5);
+}
```
vince-fugnitto, line 198: @martin-fleck-at we should neither remove the background, or change the scaling of the icons when hovering as it is the old way of rendering icons, and does not align with the rest of the icon buttons or toolbar items present in the framework. 

It currently breaks the updates from:
- https://github.com/eclipse-theia/theia/pull/9864
- https://github.com/eclipse-theia/theia/pull/10078

vince-fugnitto (changes requested):

martin-fleck-at:
@vince-fugnitto Thank you for your fast review! I removed the adaptation again and just reduced the scaling a bit to make the icon size match the close icon a bit better. If you want I can remove that as well though:

fix_expandCollapse2

vince-fugnitto:
@martin-fleck-at I haven't tested your latest changes yet, but based on the gif you sent something still looks incorrect. The behavior should be like the `close` icon, and align with vscode:

https://user-images.githubusercontent.com/40359487/143596448-fd6a7b55-659a-461e-9c1e-04587a7a6af2.mp4

martin-fleck-at:
@vince-fugnitto Got it! I updated the change but I removed the scaling now because otherwise it looks weird:

fix_noScale
(without scaling and submitted)

fix_scale
(with the scaling defined in master)

msujew (commented):
Great, thanks for taking care of that.

Something didn't seem quite right about the direction of the chevron both in expanded and collapsed form. I took a look at how VSCode handles this. Seems like my intuition proved correct :)

Collapsed:

image

Expanded:

image

Do you mind adjusting the rotation of the chevron to align with VSCode? Afterwards, this looks good to go.

martin-fleck-at:
@msujew Very good catch! Thank you! I aligned the chevron direction accordingly:

arrow

msujew (approved):
Thanks, looks great!

- [x] The icon is visible again after your changes. (compared to master)
- [x] The chevron does not increase its size on hover/click.
- [x] The chevron rotation is aligned with vscode for expansion/collapse/hover.

packages/messages/src/browser/style/notifications.css
```diff
@@ -192,17 +192,11 @@
...
+    transform: rotate(180deg);
 }
 
 .theia-notification-actions > .collapse {
-    transform: rotate(-90deg) scale(1.5);
-}
-.theia-notification-actions > .collapse:active {
-    transform: rotate(-90deg) scale(1.8);
+    transform: rotate(0deg);
 }
```
msujew, line 200: Minor: We don't need that rule anymore (rotating by 0deg does nothing) .

martin-fleck-at: :man_facepalming: I removed the rule, thanks!

Issue #10472, opened 2021-11-26
Workspace Trust API to support newer VS Code Git plugins


martin-fleck-at:
When creating an electron application with the latest VS Code built-in Git plugins:

```
  "theiaPlugins": {
    "vscode-builtin-git": "https://open-vsx.org/api/vscode/git/1.62.3/file/vscode.git-1.62.3.vsix",
    "vscode-builtin-github": "https://open-vsx.org/api/vscode/github/1.62.3/file/vscode.github-1.62.3.vsix",
    "vscode-builtin-github-authentication": "https://open-vsx.org/api/vscode/github-authentication/1.62.3/file/vscode.github-authentication-1.62.3.vsix",
    ...
  }
```

I get the following error while trying to add a remote repository with Source Control View > Remote > Add Remote...:

![git_repositories](https://user-images.githubusercontent.com/19170971/143603413-eb220442-9b4c-4678-820a-2dbefb8e8f17.png)

I believe the reason for that is that the Git plugin uses some API that Theia does not provide: Trusted workspace API:
```
// Don't allow auto-fetch in untrusted workspaces
if (workspace.isTrusted) {
   this.disposables.push(new AutoFetcher(this, globalState));
} else {
   const trustDisposable = workspace.onDidGrantWorkspaceTrust(() => {
      trustDisposable.dispose();
      this.disposables.push(new AutoFetcher(this, globalState));
   });
   this.disposables.push(trustDisposable);
}
```
https://github.com/microsoft/vscode/blob/bfad20be9d75fa83c1d3e31fb6aefbca71aeaddc/extensions/git/src/repository.ts#L942

From what I can see, the `isTrusted` is false and the `onDidGrantWorkspaceTrust` does not exist causing the error.

### Feature Description:

Implement (a very light-weight version of) the Workspace Trust API:
- At least provide the `isTrusted` flag to plugins
- Support the `onDidGrantWorkspaceTrust` event to which plugins might be registered
- Add a dialog to query the user about workspace trust to fire the event based on the users answer

The full functionality is described in https://code.visualstudio.com/docs/editor/workspace-trust

PR #10473, opened 2021-11-26
Implement light-weight Workspace Trust API for plugins

martin-fleck-at:
#### What it does
- Add trust-relevant methods and interfaces to the plugin workspace API
- Add request for workspace trust to workspace service
- Allow application setting to auto-respond to trust request

https://github.com/eclipse-theia/theia/issues/10472

**Note: I share this PR to showcase how the issue might be solved as I tried to double-check the issue I was facing. While it is working and hopefully not doing any harm, further discussions and improvements might be needed for a real support of the workspace trust API.**

#### How to test
- Create a Theia Electron application as described in https://github.com/eclipse-theia/theia/issues/10472
- Trigger the remote pickup through: Source Control View > Remote > Add Remote...

fix_git
(working Remote query)

git-trust
(user trust query)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/workspace-main.ts
+++ b/packages/plugin-ext/src/main/browser/workspace-main.ts
@@ -22,7 +22,7 @@ import { URI as Uri } from '@theia/core/shared/vscode-uri';
 import { UriComponents } from '../../common/uri-components';
 import { FileSearchService } from '@theia/file-search/lib/common/file-search-service';
 import URI from '@theia/core/lib/common/uri';
-import { WorkspaceService } from '@theia/workspace/lib/browser';
+import { WorkspaceService, WorkspaceTrustService } from '@theia/workspace/lib/browser';
 import { Resource } from '@theia/core/lib/common/resource';
 import { Disposable, DisposableCollection } from '@theia/core/lib/common/disposable';
 import { Emitter, Event, ResourceResolver, CancellationToken } from '@theia/core';
@@ -52,6 +52,8 @@ export class WorkspaceMainImpl implements WorkspaceMain, Disposable {
 
     private workspaceService: WorkspaceService;
 
+    private workspaceTrustService: WorkspaceTrustService;
+
     private fsPreferences: FileSystemPreferences;
 
     protected readonly toDispose = new DisposableCollection();
@@ -67,6 +69,7 @@ export class WorkspaceMainImpl implements WorkspaceMain, Disposable {
         this.resourceResolver = container.get(TextContentResourceResolver);
         this.pluginServer = container.get(PluginServer);
         this.workspaceService = container.get(WorkspaceService);
+        this.workspaceTrustService = container.get(WorkspaceTrustService);
         this.fsPreferences = container.get(FileSystemPreferences);
 
         this.processWorkspaceFoldersChanged(this.workspaceService.tryGetRoots().map(root => root.resource.toString()));
@@ -76,6 +79,8 @@ export class WorkspaceMainImpl implements WorkspaceMain, Disposable {
         this.toDispose.push(this.workspaceService.onWorkspaceLocationChanged(stat => {
             this.proxy.$onWorkspaceLocationChanged(stat);
         }));
+
+        this.workspaceTrustService.getWorkspaceTrust().then(trust => this.proxy.$onWorkspaceTrustChanged(trust));
     }
 
     dispose(): void {
@@ -269,6 +274,9 @@ export class WorkspaceMainImpl implements WorkspaceMain, Disposable {
         await this.workspaceService.spliceRoots(start, deleteCount, ...rootsToAdd.map(root => new URI(root)));
     }
 
+    async $requestWorkspaceTrust(_options?: theia.WorkspaceTrustRequestOptions): Promise<boolean | undefined> {
+        return this.workspaceTrustService.requestWorkspaceTrust();
+    }
 }
 
 /**
--- a/packages/plugin-ext/src/plugin/workspace.ts
+++ b/packages/plugin-ext/src/plugin/workspace.ts
@@ -56,6 +56,10 @@ export class WorkspaceExtImpl implements WorkspaceExt {
     private searchInWorkspaceEmitter: Emitter<{ result?: theia.TextSearchResult, searchId: number }> = new Emitter<{ result?: theia.TextSearchResult, searchId: number }>();
     protected workspaceSearchSequence: number = 0;
 
+    private _trusted?: boolean = undefined;
+    private didGrantWorkspaceTrustEmitter = new Emitter<void>();
+    public readonly onDidGrantWorkspaceTrust: Event<void> = this.didGrantWorkspaceTrustEmitter.event;
+
     constructor(rpc: RPCProtocol,
         private editorsAndDocuments: EditorsAndDocumentsExtImpl,
         private messageService: MessageRegistryExt) {
@@ -418,4 +422,23 @@ export class WorkspaceExtImpl implements WorkspaceExt {
             this.workspaceFileUri = URI.parse(workspace.resource.toString());
         }
     }
+
+    get trusted(): boolean {
+        if (this._trusted === undefined) {
+            this.requestWorkspaceTrust();
+        }
+        return !!this._trusted;
+    }
+
+    requestWorkspaceTrust(options?: theia.WorkspaceTrustRequestOptions): Promise<boolean | undefined> {
+        return this.proxy.$requestWorkspaceTrust(options);
+    }
+
+    $onWorkspaceTrustChanged(trust: boolean | undefined): void {
+        if (!this._trusted && trust) {
+            this._trusted = trust;
+            this.didGrantWorkspaceTrustEmitter.fire();
+        }
+    }
+
 }
--- a/packages/workspace/src/browser/workspace-trust-preferences.ts
+++ b/packages/workspace/src/browser/workspace-trust-preferences.ts
@@ -1,0 +1,76 @@
+/********************************************************************************
+ * Copyright (C) 2021 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import {
+    createPreferenceProxy, PreferenceContribution, PreferenceProxy, PreferenceSchema, PreferenceService
+} from '@theia/core/lib/browser/preferences';
+import { nls } from '@theia/core/lib/common/nls';
+import { interfaces } from '@theia/core/shared/inversify';
+
+export const WORKSPACE_TRUST_ENABLED = 'security.workspace.trust.enabled';
+export const WORKSPACE_TRUST_STARTUP_PROMPT = 'security.workspace.trust.startupPrompt';
+export const WORKSPACE_TRUST_EMPTY_WINDOW = 'security.workspace.trust.emptyWindow';
+
+export enum WorkspaceTrustPrompt {
+    ALWAYS = 'always',
+    ONCE = 'once',
+    NEVER = 'never'
+}
+
+export const workspaceTrustPreferenceSchema: PreferenceSchema = {
+    type: 'object',
+    properties: {
+        [WORKSPACE_TRUST_ENABLED]: {
+            description: nls.localize('theia/workspace/trustEnabled', 'Controls whether or not workspace trust is enabled. If disabled, all workspaces are trusted.'),
+            type: 'boolean',
+            defaultValue: true
+        },
+        [WORKSPACE_TRUST_STARTUP_PROMPT]: {
+            description: nls.localizeByDefault('Controls when the startup prompt to trust a workspace is shown.'),
+            enum: Object.values(WorkspaceTrustPrompt),
+            defaultValue: WorkspaceTrustPrompt.ALWAYS
+        },
+        [WORKSPACE_TRUST_EMPTY_WINDOW]: {
+            description: nls.localize('theia/workspace/trustEmptyWindow', 'Controls whether or not the empty workspace is trusted by default.'),
+            type: 'boolean',
+            defaultValue: true
+        }
+    }
+};
+
+export interface WorkspaceTrustConfiguration {
+    [WORKSPACE_TRUST_ENABLED]: boolean,
+    [WORKSPACE_TRUST_STARTUP_PROMPT]: WorkspaceTrustPrompt;
+    [WORKSPACE_TRUST_EMPTY_WINDOW]: boolean;
+}
+
+export const WorkspaceTrustPreferenceContribution = Symbol('WorkspaceTrustPreferenceContribution');
+export const WorkspaceTrustPreferences = Symbol('WorkspaceTrustPreferences');
+export type WorkspaceTrustPreferences = PreferenceProxy<WorkspaceTrustConfiguration>;
+
+export function createWorkspaceTrustPreferences(preferences: PreferenceService, schema: PreferenceSchema = workspaceTrustPreferenceSchema): WorkspaceTrustPreferences {
+    return createPreferenceProxy(preferences, schema);
+}
+
+export function bindWorkspaceTrustPreferences(bind: interfaces.Bind): void {
+    bind(WorkspaceTrustPreferences).toDynamicValue(ctx => {
+        const preferences = ctx.container.get<PreferenceService>(PreferenceService);
+        const contribution = ctx.container.get<PreferenceContribution>(WorkspaceTrustPreferenceContribution);
+        return createWorkspaceTrustPreferences(preferences, contribution.schema);
+    }).inSingletonScope();
+    bind(WorkspaceTrustPreferenceContribution).toConstantValue({ schema: workspaceTrustPreferenceSchema });
+    bind(PreferenceContribution).toService(WorkspaceTrustPreferenceContribution);
+}
--- a/packages/workspace/src/browser/workspace-trust-service.ts
+++ b/packages/workspace/src/browser/workspace-trust-service.ts
@@ -1,0 +1,143 @@
+/********************************************************************************
+ * Copyright (C) 2021 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { ConfirmDialog, Dialog, PreferenceChange, StorageService } from '@theia/core/lib/browser';
+import { PreferenceService } from '@theia/core/lib/browser/preferences/preference-service';
+import { MessageService } from '@theia/core/lib/common/message-service';
+import { nls } from '@theia/core/lib/common/nls';
+import { Deferred } from '@theia/core/lib/common/promise-util';
+import { inject, injectable, postConstruct } from '@theia/core/shared/inversify';
+import { WindowService } from '@theia/core/lib/browser/window/window-service';
+import { WorkspaceService } from '.';
+import {
+    WorkspaceTrustPreferences, WORKSPACE_TRUST_EMPTY_WINDOW, WORKSPACE_TRUST_ENABLED, WORKSPACE_TRUST_STARTUP_PROMPT, WorkspaceTrustPrompt
+} from './workspace-trust-preferences';
+import { FrontendApplicationConfigProvider } from '@theia/core/lib/browser/frontend-application-config-provider';
+
+const STORAGE_TRUSTED = 'trusted';
+
+@injectable()
+export class WorkspaceTrustService {
+    @inject(WorkspaceService)
+    protected readonly workspaceService: WorkspaceService;
+
+    @inject(PreferenceService)
+    protected readonly preferences: PreferenceService;
+
+    @inject(StorageService)
+    protected readonly storage: StorageService;
+
+    @inject(MessageService)
+    protected readonly messageService: MessageService;
+
+    @inject(WorkspaceTrustPreferences)
+    protected readonly workspaceTrustPref: WorkspaceTrustPreferences;
+
+    @inject(WindowService)
+    protected readonly windowService: WindowService;
+
+    protected workspaceTrust = new Deferred<boolean>();
+
+    @postConstruct()
+    protected async init(): Promise<void> {
+        await this.workspaceService.ready;
+        await this.resolveWorkspaceTrust();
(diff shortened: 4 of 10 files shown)


packages/plugin-ext/src/main/browser/workspace-main.ts
```diff
@@ -14,23 +14,23 @@
...
-import { SearchInWorkspaceService } from '@theia/search-in-workspace/lib/browser/search-in-workspace-service';
 import { FileStat } from '@theia/filesystem/lib/common/files';
 import { MonacoQuickInputService } from '@theia/monaco/lib/browser/monaco-quick-input-service';
+import * as theia from '@theia/plugin';
+import { SearchInWorkspaceService } from '@theia/search-in-workspace/lib/browser/search-in-workspace-service';
+import { WorkspaceService } from '@theia/workspace/lib/browser';
+import { MAIN_RPC_CONTEXT, StorageExt, WorkspaceExt, WorkspaceFolderPickOptionsMain, WorkspaceMain } from '../../common/plugin-api-rpc';
+import { PluginServer } from '../../common/plugin-protocol';
+import { RPCProtocol } from '../../common/rpc-protocol';
+import { UriComponents } from '../../common/uri-components';
```
vince-fugnitto, line 33: @martin-fleck-at do you mind reverting these unnecessary changes, it makes the pull-request harder to review.

packages/plugin-ext/src/plugin/workspace.ts
```diff
@@ -56,6 +56,10 @@ export class WorkspaceExtImpl implements WorkspaceExt {
     private searchInWorkspaceEmitter: Emitter<{ result?: theia.TextSearchResult, searchId: number }> = new Emitter<{ result?: theia.TextSearchResult, searchId: number }>();
     protected workspaceSearchSequence: number = 0;
 
+    private _trusted: boolean = false;
```
vince-fugnitto, line 59: I believe we should persist the choice from the user, unfortunately on each startup the application will ask users which can become annoying.

packages/workspace/src/browser/workspace-service.ts
```diff
@@ -111,6 +113,11 @@ export class WorkspaceService implements FrontendApplicationContribution {
                 this.refreshRootWatchers();
             }
         });
+
+        const autoTrust = FrontendApplicationConfigProvider.get()['autoTrust'];
+        if (typeof autoTrust === 'boolean') {
+            this.trustRequest.resolve(autoTrust);
+        }
```
colin-grant-work, line 120: I know the headline says 'light-weight,' but I'd prefer to also implement some of the [preferences](https://code.visualstudio.com/docs/editor/workspace-trust#_settings) associated with the feature so that users have some control.

packages/plugin-ext/src/main/browser/workspace-main.ts
martin-fleck-at: Ah yes, sorry, I will push an update for that soon.

packages/plugin-ext/src/plugin/workspace.ts
martin-fleck-at: Great idea, I store the value now if the preference is set to ask only 'once'.

packages/workspace/src/browser/workspace-service.ts
martin-fleck-at: Good idea! I have implemented some of the settings in a follow-up change and rebased my branch.

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -0,0 +1,115 @@
...
+        this.trusted.promise.then(trusted => this._isTrusted = trusted);
+        const implicitTrust = await this.calculateWorkspaceTrust();
+        if (implicitTrust !== undefined) {
+            this.trusted.resolve(implicitTrust);
+        }
+    }
+
+    get isTrusted(): boolean | undefined {
+        return this._isTrusted;
+    }
```
colin-grant-work, line 65: I don't see any references to this field elsewhere?

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    get onTrusted(): Promise<boolean> {
+        return this.trusted.promise;
+    }
```
colin-grant-work, line 69: The `on__` name convention is usually reserved for events. The way this is used has some similarity to an event (fire once, check often), but I'd prefer either just changing this to `get isTrusted` (since that isn't currently used) or something like
```typescript
getWorkspaceTrust(): Promise<Boolean> {
    return this.trusted.promise;
}
}

File: packages/workspace/src/browser/workspace-trust-service.ts [hunk 6]
```diff
@@ -0,0 +1,115 @@
...
+    protected _isTrusted: boolean | undefined;
+
+    @postConstruct()
+    protected async init(): Promise<void> {
+        await this.workspaceService.ready;
+        this.trusted.promise.then(trusted => this._isTrusted = trusted);
+        const implicitTrust = await this.calculateWorkspaceTrust();
+        if (implicitTrust !== undefined) {
+            this.trusted.resolve(implicitTrust);
+        }
```
colin-grant-work, line 60: Does this need to be run when this class is initialized, or could we wait until the first call to `requestWorkspaceTrust`?

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -0,0 +1,115 @@
...
+            const trusted = queryTrusted === Dialog.YES;
+
+            const trustPrompt = this.workspaceTrustPref[WORKSPACE_TRUST_STARTUP_PROMPT];
+            if (trustPrompt === WORKSPACE_TRUST_STARTUP_PROMPT_ONCE) {
+                await this.storage.setData(STORAGE_TRUSTED, trusted);
+            }
+            this.trusted.resolve(trusted);
+        }
+        return this.trusted.promise;
+    }
```
colin-grant-work, line 114: If it's reasonable to wait to run `calculateWorkspaceTrust` until the first call to `requestWorkspaceTrust`, it would be better to have the query as the final fallback if we arrive at the end of `calculateWorkspaceTrust` without having returned a `boolean` value.

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -0,0 +1,115 @@
+/********************************************************************************
+ * Copyright (C) 2017 TypeFox and others.
```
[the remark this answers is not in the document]
martin-fleck-at: Thank you, done!

packages/workspace/src/browser/workspace-trust-preferences.ts
```diff
@@ -0,0 +1,77 @@
+/********************************************************************************
+ * Copyright (C) 2018 Ericsson and others.
```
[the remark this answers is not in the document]
martin-fleck-at: Thank you, done!

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -0,0 +1,115 @@
...
+        return this._isTrusted;
+    }
+
+    get onTrusted(): Promise<boolean> {
+        return this.trusted.promise;
+    }
+
+    protected async calculateWorkspaceTrust(): Promise<boolean | undefined> {
+        await this.preferences.ready;
+        const autoTrust = FrontendApplicationConfigProvider.get()[CONFIG_AUTO_TRUST];
```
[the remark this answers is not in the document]
martin-fleck-at: I was not aware of this approach, I adapted the code accordingly. Thank you for pointing that out!

packages/workspace/src/browser/workspace-trust-service.ts
martin-fleck-at: You are right, I made a mistake there. Fixed it and went with the getWorkspaceTrust approach.

packages/workspace/src/browser/workspace-trust-service.ts
martin-fleck-at: Ideally yes, since in the plugin API the 'onDidGrantWorkspaceTrust' which we trigger after resolving this sets the trusted flag on the workspace (see plugin-context.ts and workspace.ts in plugin-ext). The Git plugin, for instances, has code like this:
```
if (workspace.isTrusted) {
   this.disposables.push(new AutoFetcher(this, globalState));
} else {
   const trustDisposable = workspace.onDidGrantWorkspaceTrust(() => {
      trustDisposable.dispose();
      this.disposables.push(new AutoFetcher(this, globalState));
   });
   this.disposables.push(trustDisposable);
}
```
So they query the flag early on and afterwards listen to any potential changes. Therefore we should set and propagate that flag as early as possible if we already know what the value will be. What do you think?

packages/workspace/src/browser/workspace-trust-service.ts
martin-fleck-at: I agree but I think we need to calculate the flag early on if possible, see my previous answer.

packages/workspace/src/browser/workspace-trust-service.ts
martin-fleck-at: Right, it was a leftover of some experiment, sorry for that!

packages/workspace/src/browser/workspace-trust-service.ts
colin-grant-work: Sounds fair to me - I'm not very familiar with the plugin side of things, but if they're expecting that it might be resolved before they're activated, then we should make the effort to satisfy that expectation. 👍

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -0,0 +1,101 @@
...
+            return true;
+        }
+
+        const trustPrompt = this.workspaceTrustPref[WORKSPACE_TRUST_STARTUP_PROMPT];
+        if (trustPrompt === WORKSPACE_TRUST_STARTUP_PROMPT_ONCE) {
+            return this.storage.getData<boolean>(STORAGE_TRUSTED);
+        }
+
+        // reset stored trusted setting
+        this.storage.setData(STORAGE_TRUSTED, undefined);
```
colin-grant-work, line 79: The handling of the storage is a bit odd. Basically, we only care about storing the value if the user has set `PROMPT_ONCE` - and we want to be able to save the value if the user changes that preference value after we run `requestWorkspaceTrust`, so perhaps the saving / deleting should be handled in a preference change listener, as well?

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -0,0 +1,101 @@
...
+        if (implicitTrust !== undefined) {
+            this.trusted.resolve(implicitTrust);
+        }
+    }
+
+    getWorkspaceTrust(): Promise<boolean> {
+        return this.trusted.promise;
+    }
+
+    protected async calculateWorkspaceTrust(): Promise<boolean | undefined> {
```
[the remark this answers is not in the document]
martin-fleck-at: Yes, I would like to keep that for customization purposes if that is alright. If not I can change it back.

packages/plugin-ext/src/plugin/workspace.ts
```diff
@@ -418,4 +422,23 @@ export class WorkspaceExtImpl implements WorkspaceExt {
...
+    get trusted(): boolean {
+        if (!this._trusted) {
+            this.requestWorkspaceTrust();
+        }
+        return this._trusted;
+    }
+
+    requestWorkspaceTrust(options?: theia.WorkspaceTrustRequestOptions): Promise<boolean | undefined> {
+        return this.proxy.$requestWorkspaceTrust(options);
+    }
```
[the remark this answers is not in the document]
martin-fleck-at: You are right, that this is a bit odd. The main 'problem' is that we do not have a Promise here to return, so we cannot actually await the result of the request. In fact, now that I think about it, I don't think we should to do the request here at all. I double checked and [VS Code](https://github.com/microsoft/vscode/blob/a0a448f344bc8c068deb1202708dfc1e7c17ef89/src/vs/workbench/api/common/extHostWorkspace.ts#L561) is also not doing it implicitly, so I removed this call for now. That means that plugins need to explicitly call the 'requestWorkspaceTrust' method if they require it.

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -0,0 +1,101 @@
...
+        if (!this.workspaceTrustPref[WORKSPACE_TRUST_ENABLED]) {
```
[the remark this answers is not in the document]
martin-fleck-at: Yeah, I think you are right about the setting. I will add a change that listens to preference changes and updates the trust only if the value has not already been set to avoid such unexpected behavior. However, I don't think we should do that if the trust has already been set since that might be even more confusing, what do you think?

packages/workspace/src/browser/workspace-trust-service.ts
martin-fleck-at: I will implement that in a follow-up change as well, I think that might be a good idea, thank you!

packages/workspace/src/browser/workspace-trust-preferences.ts
```diff
@@ -0,0 +1,84 @@
...
+        [WORKSPACE_TRUST_ENABLED]: {
+            description: nls.localize('theia/workspace/trustEnabled', 'Controls whether or not workspace trust is enabled. If disabled, all workspaces are trusted.'),
+            type: 'boolean',
+            default: FrontendApplicationConfigProvider.get()[CONFIG_AUTO_TRUST] === false
+
+        },
+        [WORKSPACE_TRUST_STARTUP_PROMPT]: {
+            description: nls.localizeByDefault('Controls when the startup prompt to trust a workspace is shown.'),
+            enum: Object.values(WorkspaceTrustPrompt),
+            default: typeof FrontendApplicationConfigProvider.get()[CONFIG_AUTO_TRUST] === 'boolean' ? WorkspaceTrustPrompt.NEVER : WorkspaceTrustPrompt.ALWAYS
```
colin-grant-work, line 52: Here the norm would be to implement a default in the `preferences` section of the application configuration rather than introducing a separate field. If you'd like to implement a separate field, you probably want to check the value of the field, not just its type.

packages/workspace/src/browser/workspace-trust-preferences.ts
```diff
@@ -0,0 +1,84 @@
...
+
+const CONFIG_AUTO_TRUST = 'autoTrust';
+
+export const workspaceTrustPreferenceSchema: PreferenceSchema = {
+    type: 'object',
+    properties: {
+        [WORKSPACE_TRUST_ENABLED]: {
+            description: nls.localize('theia/workspace/trustEnabled', 'Controls whether or not workspace trust is enabled. If disabled, all workspaces are trusted.'),
+            type: 'boolean',
+            default: FrontendApplicationConfigProvider.get()[CONFIG_AUTO_TRUST] === false
```
colin-grant-work, line 46: Here, too.

packages/workspace/src/browser/workspace-trust-preferences.ts
```diff
@@ -0,0 +1,84 @@
...
+        },
+        [WORKSPACE_TRUST_STARTUP_PROMPT]: {
+            description: nls.localizeByDefault('Controls when the startup prompt to trust a workspace is shown.'),
+            enum: Object.values(WorkspaceTrustPrompt),
+            default: typeof FrontendApplicationConfigProvider.get()[CONFIG_AUTO_TRUST] === 'boolean' ? WorkspaceTrustPrompt.NEVER : WorkspaceTrustPrompt.ALWAYS
+        },
+        [WORKSPACE_TRUST_EMPTY_WINDOW]: {
+            description: nls.localize('theia/workspace/trustEmptyWindow', 'Controls whether or not the empty workspace is trusted by default.'),
+            type: 'boolean',
+            default: FrontendApplicationConfigProvider.get()[CONFIG_AUTO_TRUST] === true
```
colin-grant-work, line 57: Here, too, use the default from VSCode and let application developers set their own defaults in the `preferences` section if desired.

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -0,0 +1,119 @@
...
+    @inject(WorkspaceTrustPreferences)
+    protected readonly workspaceTrustPref: WorkspaceTrustPreferences;
+
+    protected workspaceTrust = new Deferred<boolean>();
+
+    @postConstruct()
+    protected async init(): Promise<void> {
+        await this.workspaceService.ready;
+        await this.resolveWorkspaceTrust();
+        this.preferences.onPreferenceChanged(change => this.preferenceChanged(change));
```
colin-grant-work, line 53:
```suggestion
        this.preferences.onPreferenceChanged(change => this.handlePreferenceChange(change));
```

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    getWorkspaceTrust(): Promise<boolean> {
+        return this.workspaceTrust.promise;
+    }
+
+    protected async resolveWorkspaceTrust(givenTrust?: boolean): Promise<void> {
+        if (this.workspaceTrust.state === 'unresolved') {
```
colin-grant-work, line 61: If the user switches the preference value from trust enabled `false` -> trust enabled `true`, what should happen? With the current code, we would have resolved the trust to `true` at startup, but now the user might be requesting that we rethink that. It seems that VSCode requires a restart if you change whether workspace trust is enabled or not, presumably so that it can rerun the checks that it would run at startup.

packages/plugin-ext/src/plugin/workspace.ts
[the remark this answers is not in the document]
martin-fleck-at: We now only return the trusted value and skip the async request as discussed above.

packages/workspace/src/browser/workspace-trust-preferences.ts
martin-fleck-at: You are right, we can simply override the preferences in the application configuration so there is no real need for a dedicated field. I changed it to "normal" default values.

packages/workspace/src/browser/workspace-trust-preferences.ts
martin-fleck-at: Done

packages/workspace/src/browser/workspace-trust-service.ts
martin-fleck-at: Method was renamed

packages/workspace/src/browser/workspace-trust-service.ts
martin-fleck-at: Great suggestion! You are right that this is a bit confusing. I now ask the user if they want to restart the application if they change the trust enablement, similar to what VS Code is doing. I think with this the implication of that preference should be much clearer.

martin-fleck-at:
@colin-grant-work I hope you had a great start into the new year and thank you very much for your extensive review! It took me some time to come back to this but I updated the PR as you suggested and everything should now be working as expected.

packages/workspace/src/browser/workspace-trust-preferences.ts
```diff
@@ -0,0 +1,76 @@
...
+    NEVER = 'never'
+}
+
+export const workspaceTrustPreferenceSchema: PreferenceSchema = {
+    type: 'object',
+    properties: {
+        [WORKSPACE_TRUST_ENABLED]: {
+            description: nls.localize('theia/workspace/trustEnabled', 'Controls whether or not workspace trust is enabled. If disabled, all workspaces are trusted.'),
+            type: 'boolean',
+            defaultValue: false
```
colin-grant-work, line 39: Is it your intent to make this disabled by default (i.e. make all workspaces trusted?)

colin-grant-work (commented):
Something seems to be malfunctioning. I modified my frontend config to enable workspace trust (see comment below), confirmed that, on startup, workspace trust was not resolved, and then ran the `Add remote...` command.

When I selected the `Add remote from GitHub` option, I got this notice:

image

That looks like a problem with the authentication API, not the workspace trust API (probably).

When I just put in the URL of the Theia repo, I was prompted to name the remote, and the process concluded. I was never prompted to confirm workspace trust, but I think I should have been? Or should the request only have come when using GitHub as an authentication provider?

martin-fleck-at:
@colin-grant-work That behavior is expected now that we no longer explicitly call `requestWorkspaceTrust` whenever a plugin queries the `isTrusted` flag. I had that originally but I think it was removed as part of the review https://github.com/eclipse-theia/theia/pull/10473#discussion_r761244478. So we either need to find a good trigger point to request the workspace trust, e.g., when an API registers a listener for it with `onDidGrantWorkspaceTrust` in which case the asynchronous nature does not hurt us or we expect the plugins to call that behavior explicitly which may be a bad idea seeing that the vscode git-plugin already does not do that. I suggest to add it on the `onDidGrantWorkspaceTrust` if the trusted flag is too risky, what do you think?

packages/workspace/src/browser/workspace-trust-preferences.ts
martin-fleck-at: Yes, that was the intent. I believe VS Code does it similarly, see https://github.com/microsoft/vscode/blob/f7bfe5ca6682f27bc0ef19b412a4888078265ecc/src/vs/workbench/services/workspaces/common/workspaceTrust.ts#L291

colin-grant-work:
@martin-fleck-at, that all sounds reasonable, and it explains why, the first time I ran through the test steps with workspace trust automatically enabled, I got an additional prompt asking me if I wanted to do automatic fetches, which is the feature that's checking workspace trust.

So as you say, the question is when to prompt, I guess. I find VSCode's prompt-on-start pretty annoying, so maybe your original idea of requesting trust when anyone checks for it is the best idea. The GitHub plugin immediately creates a listener if it finds the workspace is untrusted, and that seems like the only good way for a plugin to handle finding `isTrusted` === `false`. That has the benefit of not requesting trust if no one depends on it and requesting trust in a timely manner if a feature activates that does depend on it. In that case, though, we probably want to have a system that caches a definite negative response so that we don't pester the user with multiple requests if they've already denied it.

colin-grant-work:
I've implemented a tiny VSCode plugin [here](https://github.com/colin-grant-work/vscode-extension-samples/tree/demo/workspace-trust/workspace-api-test) with commands to exercise the workspace API.

martin-fleck-at:
@colin-grant-work Thank you very much for the plugin, it was very useful to test the different scenarios! As you suggested, we now request the workspace trust whenever the trust is queried. In addition to that, we report any workspace trust state back to the plugin (`onWorkspaceTrustChanged`) even when we determine that it is false. In general, the user should not be pestered with multiple requests anyway as once we resolve the workspace trust no further dialogs are shown. This is also the reason that we trigger a restart when the respective preference was changed, similar to VS Code. 

Thank you for your review and effort!

colin-grant-work (commented):
I'm happy with the code and the current behavior. The only outstanding question is whether to enable or disable the workspace trust feature by default, and I'd like to give others a chance to weigh in.

JonasHelming:
I think we should enable the feature by default as it provides a more defensive strategy when it comes to security. Adopters can still choose to disable the feature in their application default preferences if they accept any risk that may come with it. However, if they do not do that it is probably better to inform the user that a plugin requests workspace trust.

martin-fleck-at:
@JonasHelming @colin-grant-work Thank you for your feedback! Thinking about it a bit more, I fully agree with Jonas that if no plugin requests the trust, the user is not bothered anyway but if a plugin needs it, it may be better to inform the user unless some adopters explicitly choose not to do so. As such, I changed the default enablement preference to `true` and squashed and rebased the change. If there are any more changes necessary before this can be merged, please let me know.

JonasHelming:
@colin-grant-work : I would merge this now, OK?

msujew (commented):
@colin-grant-work I don't really have a strong opinion on that. I think both would be fine, but I'm slightly leaning towards not enabling it by default.

In the end, downstream users of Theia can always choose to have it enabled by default in their product.

JonasHelming:
Just to mention, I also do not have a strong opinion about it. So @colin-grant-work Your vote will decide on this, no option is "wrong"

colin-grant-work:
@msujew, @JonasHelming, @martin-fleck-at, thanks for your input. As annoying as I find VSCode's modal at start up, I think enabling security by default is better than disabling it by default - adopters and users are free to follow whatever policies their risk managers will let them get away with :-). So default on sounds good to me, and I'll merge.

colin-grant-work (approved):
The latest code looks good to me, and I think we've settled on enabling the feature by default.

PR #10480, opened 2021-11-26
webpack: fix the frontend source map names

paul-marechal:
Webpack 5 changed the default for source map names to something like
`@theia/example-browser/<path/to/file>` which confused the JS debugger.

Use `output.devtoolModuleFilenameTemplate` to define a "simpler" source
map filename.

https://webpack.js.org/configuration/output/#outputdevtoolmodulefilenametemplate

#### How to test

- You should be able to put breakpoints in frontend code from your IDE and have them bound and hit.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -20,18 +20,18 @@
       "type": "node",
       "request": "launch",
       "name": "Launch Electron Backend",
-      "runtimeExecutable": "${workspaceRoot}/node_modules/.bin/electron",
+      "runtimeExecutable": "${workspaceFolder}/node_modules/.bin/electron",
       "windows": {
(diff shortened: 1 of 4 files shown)


alvsan09:
@paul-marechal ,
I have tested the change successfully using the browser, 
however it does not seem to work for me for Electron.

alvsan09:
> @paul-marechal , I have tested the change successfully using the browser, however it does not seem to work for me for Electron.

I can confirm the fix in the PR works fine in Electron as well, as @paul-marechal showed me, I must have missed an step while building the application.

alvsan09 (approved):
I can confirm that this fixes the setting of breakpoints both in the browser as well as in electron :+1: 
The code looks good to me as well.

Issue #10478, opened 2021-11-26
Mini browser preview URL shows open source icon


aditya211935:
### Bug Description:
Upon opening a url on preview mode via mini browser, "Open Source" icon is shown. Since a remote url opened in mini browser has no source on disk, this shouldn't happen. There's already a check for this inside `getSourceUri` function located in `packages/mini-browser/src/browser/mini-browser-open-handler.ts`, but it checks for URI's scheme to be "http" or "https". This is not true, since this is Theia's internal URI, and as such has a scheme `__minibrowser__preview__` that's defined inside 
`MiniBrowserOpenHandler`'s static variable called `PREVIEW_URI`. For internal files that are opened in mini browser, their URI corresponds to location of the file on disk. Thus, the uri passed to `getSourceUri` will never have a http or https scheme.

### Steps to Reproduce:

1. Press Command + Shift + P to open up command palette. Type "Open URL" and press enter.
2. Enter any URL that you want to open.
3. Mini browser will open the specified url in preview mode, and will be docked to the right side panel. Look at its top right side, and an "Open Source" icon will be visible. Upon clicking this icon, nothing will happen.

<img width="1029" alt="preview-icon" src="https://user-images.githubusercontent.com/23213686/143612901-0b325b8e-6ca7-4479-a65a-ad3cf0eef9a3.png">

### Additional Information

- Operating System: MacOS Monterey
- Theia Version: 1.20.0

PR #10481, opened 2021-11-26
fixed mini browser getSource uri

aditya211935:
Signed-off-by: Aditya Sharma <aditya211935@gmail.com>

#### What it does
Fixes: https://github.com/eclipse-theia/theia/issues/10478

Since the URI passed to getSourceUri function doesn't have "http" or "https" scheme, and since a remote URL opened in preview mode doesn't have a a source URI on disk, a simple check is performed to check if the passed URI is equal to MiniBrowserOpenHandler's PREVIEW_URI static variable.

#### How to test

1. Press Command + Shift + P to open up command palette. Type "Open URL" and press enter.
2. Enter any URL that you want to open.
3. Mini browser will open the specified url in preview mode, and will be docked to the right side panel. Look at its top right side, and an "Open Source" icon will not be visible.
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/mini-browser/src/browser/mini-browser-open-handler.ts
+++ b/packages/mini-browser/src/browser/mini-browser-open-handler.ts
@@ -273,7 +273,7 @@ export class MiniBrowserOpenHandler extends NavigatableWidgetOpenHandler<MiniBro
 
     protected getSourceUri(ref?: Widget): URI | undefined {
         const uri = ref instanceof MiniBrowser && ref.getResourceUri() || undefined;
-        if (!uri || uri.isEqual(MiniBrowserOpenHandler.PREVIEW_URI)) {
+        if (!uri || uri.scheme === 'http' || uri.scheme === 'https' || uri.isEqual(MiniBrowserOpenHandler.PREVIEW_URI)) {
             return undefined;
         }
         return uri;
--- a/packages/mini-browser/src/browser/mini-browser-open-handler.ts
+++ b/packages/mini-browser/src/browser/mini-browser-open-handler.ts
@@ -273,7 +273,7 @@ export class MiniBrowserOpenHandler extends NavigatableWidgetOpenHandler<MiniBro
 
     protected getSourceUri(ref?: Widget): URI | undefined {
         const uri = ref instanceof MiniBrowser && ref.getResourceUri() || undefined;
-        if (!uri || uri.scheme === 'http' || uri.scheme === 'https') {
+        if (!uri || uri.isEqual(MiniBrowserOpenHandler.PREVIEW_URI)) {
             return undefined;
         }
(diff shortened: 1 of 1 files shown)


aditya211935:
@vince-fugnitto I've accepted the terms.

aditya211935:
@vince-fugnitto I can see in GitHub workflow a green tick related to signing ECA


image


but on Eclipse's website, I see a red cross.

image

vince-fugnitto:
@aditya211935 thank you for signing the eca, everything looks in order :+1:

aditya211935:
@vince-fugnitto Any update on when it'll be merged? I can see that all workflow checks have passed.

msujew (approved):
@aditya211935 Thank you for your contribution.

I confirmed that the issue exists on `master` and is addressed by your changes. Very clean solution 👍 

- [x] Opening the mini browser using the `Open URL` command does not show the `Open Source` toolbar button.
- [x] Opening the mini browser by previewing a local html files shows the toolbar button and clicking it opens the original html file correctly.

aditya211935:
@msujew Thanks! 👍

packages/mini-browser/src/browser/mini-browser-open-handler.ts
```diff
@@ -273,7 +273,7 @@ export class MiniBrowserOpenHandler extends NavigatableWidgetOpenHandler<MiniBro
 
     protected getSourceUri(ref?: Widget): URI | undefined {
         const uri = ref instanceof MiniBrowser && ref.getResourceUri() || undefined;
-        if (!uri || uri.scheme === 'http' || uri.scheme === 'https') {
```
paul-marechal, line 276: Do we really want to remove the checks for the HTTP(S) scheme though? Should we rather add the new check on top of the old ones?

paul-marechal: IIRC the mini-browser can still display HTTP(S) content and now it will show the "open source" button, no?

msujew: You're right, we should probably add this new check in addition to the old ones.

aditya211935: If you follow the chain of uri, you'll see that it comes from `widgetOptions`. This URI is only internal to Theia, and doesn't contain any http, or https scheme. Probably what you're referring to is `startPage`. Mini browser can display http(s) content, but only when it's opened in preview mode. Also, the http(s) content's link will be inside `startPage` MiniBrowserProps, and not URI. Hope that clears things up!

paul-marechal (approved):
Thanks for keeping the old checks.

PR #10484, opened 2021-11-27
feat: add timestamp for DAP trace and formate output using `JSON.stringify(message, null, 4)`

DanboDuan:
#### ADD timestamp for DAP trace output
DAP trace output does not contain timestamp. it looks like this

```
fabf8f58 theia <- adapter: {"body":{"reason":"exited","threadId":2305159},"event":"thread","seq":0,"type":"event"}
fabf8f58 theia <- adapter: {"body":{"reason":"exited","threadId":2305187},"event":"thread","seq":0,"type":"event"}
fabf8f58 theia <- adapter: {"body":{"reason":"exited","threadId":2305194},"event":"thread","seq":0,"type":"event"}
fabf8f58 theia <- adapter: {"body":{"reason":"exited","threadId":2305286},"event":"thread","seq":0,"type":"event"}
fabf8f58 theia <- adapter: {"body":{"reason":"exited","threadId":2305343},"event":"thread","seq":0,"type":"event"}
fabf8f58 theia <- adapter: {"body":{"reason":"exited","threadId":2305191},"event":"thread","seq":0,"type":"event"}
fabf8f58 theia <- adapter: {"body":{"reason":"exited","threadId":2305281},"event":"thread","seq":0,"type":"event"}
fabf8f58 theia -> adapter: {"seq":18,"type":"request","command":"threads","arguments":{}}
```
It's not easy to get the performance with those trace output. It's better to output a timestamp so you will be easy to get a context.
#### How to test

check the dap trace output is now with timestamp like this

```
fabf8f58 theia <- adapter: {"body":{"reason":"exited","threadId":2305159},"event":"thread","seq":0,"type":"event"}
```

after the changes submit, check the dap trace output is now with timestamp like this

```
fabf8f58 2021-11-27T02:40:06.428Z theia <- adapter: {"body":{"reason":"exited","threadId":2305159},"event":"thread","seq":0,"type":"event"}
```
#### Review checklist

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with the review guidelines

### DIFF
--- a/packages/debug/src/browser/debug-session-connection.ts
+++ b/packages/debug/src/browser/debug-session-connection.ts
@@ -243,10 +243,9 @@ export class DebugSessionConnection implements Disposable {
         const connection = await this.connectionPromise;
         const messageStr = JSON.stringify(message);
         if (this.traceOutputChannel) {
+            // keep getMilliseconds
             const now = new Date();
-            /// offset in milliseconds
-            const tzoffset = now.getTimezoneOffset() * 60 * 1000;
-            const dateStr = new Date(now.getTime() - tzoffset).toISOString();
+            const dateStr = `${now.toLocaleString(undefined, { hour12: false })}.${now.getMilliseconds()}`;
             this.traceOutputChannel.appendLine(`${this.sessionId.substring(0, 8)} ${dateStr} theia -> adapter: ${messageStr}`);
         }
         connection.send(messageStr);
@@ -254,10 +253,9 @@ export class DebugSessionConnection implements Disposable {
 
     protected handleMessage(data: string): void {
         if (this.traceOutputChannel) {
+            // keep getMilliseconds
             const now = new Date();
-            /// offset in milliseconds
-            const tzoffset = now.getTimezoneOffset() * 60 * 1000;
-            const dateStr = new Date(now.getTime() - tzoffset).toISOString();
+            const dateStr = `${now.toLocaleString(undefined, { hour12: false })}.${now.getMilliseconds()}`;
             this.traceOutputChannel.appendLine(`${this.sessionId.substring(0, 8)} ${dateStr} theia <- adapter: ${data}`);
         }
         const message: DebugProtocol.ProtocolMessage = JSON.parse(data);
--- a/packages/debug/src/browser/debug-session-connection.ts
+++ b/packages/debug/src/browser/debug-session-connection.ts
@@ -243,14 +243,22 @@ export class DebugSessionConnection implements Disposable {
         const connection = await this.connectionPromise;
         const messageStr = JSON.stringify(message);
         if (this.traceOutputChannel) {
-            this.traceOutputChannel.appendLine(`${this.sessionId.substring(0, 8)} theia -> adapter: ${messageStr}`);
+            const now = new Date();
+            /// offset in milliseconds
+            const tzoffset = now.getTimezoneOffset() * 60 * 1000;
+            const dateStr = new Date(now.getTime() - tzoffset).toISOString();
+            this.traceOutputChannel.appendLine(`${this.sessionId.substring(0, 8)} ${dateStr} theia -> adapter: ${messageStr}`);
         }
         connection.send(messageStr);
     }
 
     protected handleMessage(data: string): void {
         if (this.traceOutputChannel) {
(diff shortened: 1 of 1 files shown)


packages/debug/src/browser/debug-session-connection.ts
```diff
@@ -243,14 +243,22 @@ export class DebugSessionConnection implements Disposable {
         const connection = await this.connectionPromise;
         const messageStr = JSON.stringify(message);
         if (this.traceOutputChannel) {
-            this.traceOutputChannel.appendLine(`${this.sessionId.substring(0, 8)} theia -> adapter: ${messageStr}`);
+            const now = new Date();
+            /// offset in milliseconds
+            const tzoffset = now.getTimezoneOffset() * 60 * 1000;
+            const dateStr = new Date(now.getTime() - tzoffset).toISOString();
```
msujew, line 249: Suggestion: As far as I know, `toISOString` will always convert the input time into UTC. So we can simplify this by using the following code. The same applies to the other change.
```suggestion
            const dateStr = new Date().toISOString();
```

msujew (commented):
Hi @DanboDuan, thank you for your first contribution.

the improvement to logging looks good. I have a suggestion to simplify the change.

packages/debug/src/browser/debug-session-connection.ts
DanboDuan: minus the timezone offset for the output to be the same as local time. If not, the time string should be `2021-11-27T02:40:06.428Z` for a UTC+9 which should be `2021-11-27 11:40:06.428` as local. It is good, but it looks strange. So I make a new date minus the timezone offset to create a data string

DanboDuan: I update to use `toLocaleString`

packages/debug/src/browser/debug-session-connection.ts
```diff
@@ -243,16 +243,22 @@ export class DebugSessionConnection implements Disposable {
         const connection = await this.connectionPromise;
         const messageStr = JSON.stringify(message);
         if (this.traceOutputChannel) {
-            this.traceOutputChannel.appendLine(`${this.sessionId.substring(0, 8)} theia -> adapter: ${messageStr}`);
+            // keep getMilliseconds
+            const now = new Date();
+            const dateStr = `${now.toLocaleString(undefined, { hour12: false })}.${now.getMilliseconds()}`;
+            this.traceOutputChannel.appendLine(`${this.sessionId.substring(0, 8)} ${dateStr} theia -> adapter: ${JSON.stringify(message, null, 4)}`);
```
vince-fugnitto, line 249: @DanboDuan if you don't need `null` explicitly, please use `undefined` instead, it fails the build unless the line is suppressed:

```suggestion
            this.traceOutputChannel.appendLine(`${this.sessionId.substring(0, 8)} ${dateStr} theia -> adapter: ${JSON.stringify(message, undefined, 4)}`);
```

The same is similar for `handleMessage`.

DanboDuan: Thx, done.

packages/debug/src/browser/debug-session-connection.ts
```diff
@@ -243,16 +243,22 @@ export class DebugSessionConnection implements Disposable {
         const connection = await this.connectionPromise;
         const messageStr = JSON.stringify(message);
         if (this.traceOutputChannel) {
-            this.traceOutputChannel.appendLine(`${this.sessionId.substring(0, 8)} theia -> adapter: ${messageStr}`);
+            // keep getMilliseconds
```
vince-fugnitto, line 246: I believe we can remove the comment, it does not add value (the same goes for `handleMessage`).

vince-fugnitto (approved):
I confirmed that the changes work well, and the timestamp is now added to the traces of the debug adapter. Example:

```
ab3a895f 12/1/2021, 08:50:30.582 theia -> adapter: {
    "seq": 34,
    "type": "request",
    "command": "scopes",
    "arguments": {
        "frameId": 34
    }
}
```

packages/debug/src/browser/debug-session-connection.ts
DanboDuan: ok

vince-fugnitto:
@msujew any objections? :)

msujew (approved):
@vince-fugnitto None, looks good to me 👍

Issue #10383, opened 2021-11-04
Support `editor.wrapTabs` preference


colin-grant-work:
### Feature Description:

VSCode has a preference that allows users to choose to display editor tabs with wrapping rather than with scrolling:

![image](https://user-images.githubusercontent.com/62660806/140340746-a4c03d27-6852-4036-81fe-39f3ee264034.png)

We currently don't support this preference.

PR #10499, opened 2021-12-03
Add preferences wraptabs

smart-bo:
Add `wraptabs` preferences at setting menu, and modify the coresponding css styles to make whole display fit the wrapped tabs.

#### What it does

fix: #10383
#### How to test

1. Open settings
2. Search `tab`
image
3. Check `Workbench › Editor: Wraptabs`
4. Confirm can choose to display editor tabs with wrapping rather than with scrolling
5. Uncheck this preference
6. Confirm display editor tabs with scroll bar.
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/common-frontend-contribution.ts
+++ b/packages/core/src/browser/common-frontend-contribution.ts
@@ -370,6 +370,7 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
         this.registerCtrlWHandling();
 
         this.updateStyles();
+        this.wrapTabs();
         this.updateThemeFromPreference('workbench.colorTheme');
         this.updateThemeFromPreference('workbench.iconTheme');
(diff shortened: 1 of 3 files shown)


colin-grant-work (commented):
There are a few issues that will need to be investigated:
 - Dragging tabs: at the moment, if you have multiple tab rows and try to drag a tab all tabs at approximately the same horizontal position move out of the way, and the tab can only be placed on the top row.
 - Splitting the dock panel I got very strange behavior with this:
 
wrap-tab-bug

- Differences from VSCode: not necessarily problems, but worth thinking about.

VSCode:  
image

Theia:  
image

- In VSCode, the toolbar appears in the last row; in Theia it's vertically centered
- In VSCode the last tab in a row extends to the end of the row; in Theia the toolbar takes up space in every row.

smart-bo:
This PR lead to many display issues.

PR #10509, opened 2021-12-06
tasks: Use valid column when creating range for problem matcher

msujew:
#### What it does

When calling [`Range.create`](https://github.com/microsoft/vscode-languageserver-node/blob/2794f5bd7436d2545e542ac86a291bc4bbe6d2af/types/src/main.ts#L166) with a value larger than > [`2^32-1`](https://github.com/microsoft/vscode-languageserver-node/blob/2794f5bd7436d2545e542ac86a291bc4bbe6d2af/types/src/main.ts#L4137), the function will throw an error. We do this when building ranges parsed from matched problems where we don't receive start/end columns:

https://github.com/eclipse-theia/theia/blob/afd980a4d4ff6b48d70fc0ce1e459f6496c62624/packages/task/src/node/task-abstract-line-matcher.ts#L204

`Number.MAX_VALUE` is something like 10^308. Since `Range.create` always throws in this case, we never receive any matched problems from the server, and any errors are swallowed by a `catch { return undefined }`.

#### How to test

0. Use this reproducible setup:

A file named `test.js` in the project root:
```js
console.log('[DUMMY] WARNING test.js:1 This is a dummy warning')
console.log('[DUMMY] ERROR test.js:2 This is a dummy error')
```
A task definition like this:
```json
{
  "label": "Test Matcher",
  "type": "shell",
  "command": "node",
  "args": [
    "test.js"
  ],
  "problemMatcher": {
    "name": "test-matcher",
    "owner": "test-matcher",
    "fileLocation": [
      "relative",
      "${workspaceFolder}"
    ],
    "pattern": {
      "regexp": "^\\[[A-Z]+\\] (INFO|WARNING|ERROR) (.*):(\\d+) (.*)$",
      "severity": 1,
      "file": 2,
      "line": 3,
      "message": 4
    }
  }
}
```

1. Run the `Test Matcher` task
2. Open `test.js` and assert that the first line is marked with a warning
3. Assert that the second line is marked with an error

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/task/src/node/task-abstract-line-matcher.ts
+++ b/packages/task/src/node/task-abstract-line-matcher.ts
@@ -29,6 +29,7 @@ import URI from '@theia/core/lib/common/uri';
 // TODO use only URI from '@theia/core'
 import { URI as vscodeURI } from '@theia/core/shared/vscode-uri';
 import { Severity } from '@theia/core/lib/common/severity';
+import { MAX_SAFE_INTEGER } from '@theia/core/lib/common/numbers';
 
 const endOfLine: string = isWindows ? '\r\n' : '\n';
 
@@ -201,7 +202,7 @@ export abstract class AbstractLineMatcher {
                 range = Range.create(startLine, startColumn, startLine, startColumn);
             }
         } else {
-            range = Range.create(startLine, 1, startLine, Number.MAX_VALUE);
+            range = Range.create(startLine, 1, startLine, MAX_SAFE_INTEGER);
         }
 
         // range indexes should be zero-based
--- a/packages/core/src/common/index.ts
+++ b/packages/core/src/common/index.ts
@@ -39,6 +39,7 @@ export * from './application-error';
 export * from './lsp-types';
 export * from './contribution-filter';
 export * from './nls';
+export * from './numbers';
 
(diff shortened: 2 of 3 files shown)


packages/task/src/node/task-abstract-line-matcher.ts
```diff
@@ -31,6 +31,7 @@ import { URI as vscodeURI } from '@theia/core/shared/vscode-uri';
 import { Severity } from '@theia/core/lib/common/severity';
 
 const endOfLine: string = isWindows ? '\r\n' : '\n';
+const MAX_RANGE_COLUMN = 2147483647;
```
colin-grant-work, line 34: Might it make sense to export something like `MAX_SAFE_INTEGER` from somewhere in `core`?

colin-grant-work (approved):
I confirm the presence of the bug on `master` and the efficacy of this fix. 👍

packages/task/src/node/task-abstract-line-matcher.ts
msujew: Yeah, makes sense. I added a `numbers.ts` in `core/common`.

Issue #10428, opened 2021-11-17
VSCode Semantic Tokens Sample Extension Does Not Work


colin-grant-work:
### Feature Description:

On current `master` (29ebe4c1cc1f8608d23ad92220c6a615c775a27c), it appears that the VSCode (semantic tokens sample extension)[https://github.com/microsoft/vscode-extension-samples/tree/main/semantic-tokens-sample] does not work. It appears to activate successfully, but the code to `provideDocumentSemanticTokens` is never called. This is the case in all themes, even with the preference `editor.SemanticHighlightingEnabled` set to `true`.

VSCode:

![image](https://user-images.githubusercontent.com/62660806/142246485-898d2f6d-c12c-4b2d-b001-ea96ec02593f.png)

Theia:

![image](https://user-images.githubusercontent.com/62660806/142245802-7673ac41-049a-45e0-b3e6-1665ad43a972.png)

OS: RHEL7
Platform: Electron
Version: 29ebe4c1cc1f8608d23ad92220c6a615c775a27c

colin-grant-work:
After fixing a bug with the prefrence setting (#10511) and setting `editor.semanticHighlighting.enabled` to `true`, I can get to this screenshot:

![image](https://user-images.githubusercontent.com/62660806/145064130-65257943-0394-4390-96e5-1377d670c5cb.png)

There seem to be two problems remaining:
 - Theia's builtin themes do not have `semanticHighlighting` set to `true`, so the `'configuredByTheme'` setting mostly equates to `false`, unlike in VSCode.
 - In VSCode, if you force semantic highlighting on a theme that doesn't support it, it appears that the editor is able to fall back to colors provided by the grammar for tokens whose semantic specification it doesn't understand. Theia fails and falls back to the default color (plain white, e.g.).

colin-grant-work:
> In VSCode, if you force semantic highlighting on a theme that doesn't support it, it appears that the editor is able to fall back to colors provided by the grammar for tokens whose semantic specification it doesn't understand. Theia fails and falls back to the default color (plain white, e.g.).

This seems to be a consequence of the way we consume Monaco. We use the `IStandaloneThemeService(Impl)`:

https://github.com/eclipse-theia/theia/blob/d149f5166063412e40350b6cad16104175d6fa4f/packages/monaco/src/typings/monaco/index.d.ts#L937-L954

Which implements a `getTokenStyleMetadata()` method:

[...]

PR #10511, opened 2021-12-06
Handle non-string enum values in preference schemata

colin-grant-work:
#### What it does

Addresses part of #10428 by handling non-string enum values in the preferences UI.

Our current preference UI system assumes that all enum values will be strings, but it turns out that's not the case. For example, the `editor.semanticHighlighting.enabled` preference is defined like this:

https://github.com/eclipse-theia/theia/blob/c4c6cab037230fc05e39c7b124a681630c2f6a33/packages/editor/src/browser/editor-preferences.ts#L124

#### How to test

1. Open the preferences UI.
2. Open the JSON file corresponding to the scope you're currently examining so that you can view both at once.
3. In the UI, search for the `editor.semanticHighlighting.enabled` preference.
4. Set it to its various values and observe that the JSON receives the correct value: an actual `boolean` if you selected one, or the string `'configuredByTheme'`.
5. Try some other enums (editor.wordWrap, editor.acceptSuggestionOnEnter)
6. Observe that they also set the values correctly.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/browser/views/components/preference-select-input.ts
+++ b/packages/preferences/src/browser/views/components/preference-select-input.ts
@@ -16,22 +16,28 @@
 
 import { PreferenceLeafNodeRenderer } from './preference-node-renderer';
 import { injectable } from '@theia/core/shared/inversify';
+import { JSONValue } from '@theia/core/shared/@phosphor/coreutils';
 
 @injectable()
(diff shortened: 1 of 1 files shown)


vince-fugnitto (approved):

I verified using the `editor.semanticHighlighting.enabled` preference that the following values are correctly populated from the select input:
- [x] `true`
- [x] `false`
- [x] `"configuredByTheme"`

I've also confirmed that other select inputs still work correctly.

PR #10524, opened 2021-12-09
Monkeypatch Monaco to improve semantic token resolution

colin-grant-work:
#### What it does

Fixes most of #10428. Read the comments there for why this involves monkey-patching Monaco. It's not elegant, so I'm open to any suggestions.

#### How to test

1. Set the `editor.semanticHighlighting.enabled` preference to `true`
2. Use the `Dark+` theme
3. Open the file `output-channel-registry-main.ts`
4. Observe that `commandService` in `this.commandService` is a darker blue than a function parameter (e.g.), and that `OutputCommands` is colored green.
5. If you'd like, build the [semantic highlighting sample extension](https://github.com/microsoft/vscode-extension-samples/tree/main/semantic-tokens-sample) and open the `sample.semanticLanguage` file in the `sample` directory.
6. Observe that you get something like:

image

> Update: I've added a second commit that adds a crude version of [VSCode's TokenClassificationRegistry](https://github.com/theia-ide/vscode/blob/standalone/0.23.x/src/vs/workbench/services/themes/common/colorExtensionPoint.ts). With that commit, the behavior of the semantic highlighting example is almost identical to VSCode:  
image

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-frontend-module.ts
+++ b/packages/monaco/src/browser/monaco-frontend-module.ts
@@ -15,6 +15,7 @@
 // *****************************************************************************
 
 import * as MonacoNls from '@theia/monaco-editor-core/esm/vs/nls';
+import './monaco-monkey-patches';
 import { nls } from '@theia/core/lib/common/nls';
 import { FormatType, Localization } from '@theia/core/lib/common/i18n/localization';
 
--- a/packages/monaco/src/browser/monaco-monkey-patches.ts
+++ b/packages/monaco/src/browser/monaco-monkey-patches.ts
@@ -1,0 +1,127 @@
+/********************************************************************************
+ * Copyright (C) 2022 Ericsson and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { ColorId, FontStyle } from '@theia/monaco-editor-core/esm/vs/editor/common/languages';
+import {
+    ColorMap, ParsedTokenThemeRule, strcmp,
+    ThemeTrieElement, ThemeTrieElementRule, TokenTheme
+} from '@theia/monaco-editor-core/esm/vs/editor/common/languages/supports/tokenization';
+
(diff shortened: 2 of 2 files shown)


msujew:
@colin-grant-work Do have an idea whether this will still be necessary after the `monaco-editor` update? I would gladly review this, assuming it is still relevant after the update.

colin-grant-work:
> @colin-grant-work Do have an idea whether this will still be necessary after the `monaco-editor` update? I would gladly review this, assuming it is still relevant after the update.

@msujew, unfortunately, I think it still will be necessary. The VSCode system that this aims to simulate is the `TokenClassificationRegistry`, and that is not part of the `monaco-editor-core` build, so if we continue to consume that package as intended, we won't get it. In the past, we've modified the build recipe for the `monaco-editor-core` package slightly to include some components that aren't in the base build, and that may be an option. I'm in the process of updating our API usage at the moment, but once that's done, I can check on whether it would be possible to include the tokenization system without bringing along too much else that we might not want.

packages/monaco/src/browser/monaco-frontend-module.ts
```diff
@@ -15,6 +15,7 @@
 // *****************************************************************************
 
 import * as MonacoNls from '@theia/monaco-editor-core/esm/vs/nls';
+import './monaco-monkey-patches';
```
msujew, line 18: The reason why the `MonacoNls` override stuff happens before any other monaco code is loaded, is that the `nls.localize` method is often times called during load-time of code.

I'm not sure whether `vs/editor/common/languages` or `vs/editor/common/languages/supports/tokenization` use localizations at load-time, but just to be sure, please move this import after the `Object.assign` call.

packages/monaco/src/browser/monaco-monkey-patches.ts
```diff
@@ -0,0 +1,127 @@
...
+
+import { ColorId, FontStyle } from '@theia/monaco-editor-core/esm/vs/editor/common/languages';
+import {
+    ColorMap, ParsedTokenThemeRule, strcmp,
+    ThemeTrieElement, ThemeTrieElementRule, TokenTheme
+} from '@theia/monaco-editor-core/esm/vs/editor/common/languages/supports/tokenization';
+
+// Cf. https://github.com/theia-ide/vscode/blob/e930e4240ee604757efbd7fd621b77b75568f95d/src/vs/platform/theme/common/tokenClassificationRegistry.ts#L508-L535
+// TODO: Remove in favor of full implementation of the token type contribution points.
+// https://github.com/theia-ide/vscode/blob/standalone/0.23.x/src/vs/workbench/services/themes/common/colorExtensionPoint.ts
```
msujew, line 25: I guess some of these links should be updated?

msujew (commented):
This looks quite good. I can confirm that this change improves handling of semantic and syntactic token highlighting.

The comment below is just a minor thing. I would like to see the comment above addressed before approving.

colin-grant-work:
@msujew, thanks for taking a look. I do think this is a step in the right direction, but I'd like to implement the plugin side of semantic token contributions as well, so I've marked this as a draft. Now that I know more about Monaco, I should be able to make the necessary modifications as a true override rather than as a monkey patch, but I wanted to update it for the Monaco uplift as a version of the code is in use in our downstream and needed to be adjusted to use the new import styles.

colin-grant-work:
Closing for now - I have WiP to replace it with real support, but it isn't very close to completion.

PR #10527, opened 2021-12-09
fix #8210: make validation message of existing resources more specific

### LINKED ISSUE #8210: [dialog][workspace] Enhance the validation message when creating a new resource
It would be great to know whether it is a file or a directory:

https://github.com/eclipse-theia/theia/blob/4c63f08a6283f19f6f28727727378f7c1e38e316/packages/workspace/src/browser/workspace-commands.ts#L386

Let's improve the validation message.

xpomul:
Signed-off-by: Stefan Winkler <stefan@winklerweb.net>

#### What it does
Fixes #8210

Instead of providing a generic validation message in `validateFileName()`, if the resource exists, we resolve the resource and decide whether it is a folder or not. Depending on the outcome, we use the correct type (file or folder) in the validation message.

#### How to test
I have added respective test cases to the unit tests (-> yarn test).
To verify manually, open Theia with any workspace containg at least a file and a folder.
- Execute the command „New File“ and enter the name of the existing file.
- Execute the command „New Folder“ and enter the name of the existing file.
=> the validation message should say that A file „XXX“ already exists at this location

- Execute the command „New File“ and enter the name of the existing folder.
- Execute the command „New Folder“ and enter the name of the existing folder.
=> the validation message should say that A folder „XXX“ already exists at this location

#### Review checklist

#### Reminder for reviewers


packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -442,7 +442,13 @@ export class WorkspaceCommandContribution implements CommandContribution {
         const childUri = parent.resource.resolve(name);
         const exists = await this.fileService.exists(childUri);
         if (exists) {
-            return nls.localizeByDefault('A file or folder "{0}" already exists at this location.', this.trimFileName(name));
+            // check the nature of the child item so we can return a specific validation message
+            const existingChild = await this.fileService.resolve(childUri);
+            if (existingChild.isDirectory) {
+                return nls.localizeByDefault('A folder "{0}" already exists at this location.', this.trimFileName(name));
+            } else {
+                return nls.localizeByDefault('A file "{0}" already exists at this location.', this.trimFileName(name));
```
msujew, line 450: As per the [i18n section in our coding guidelines](https://github.com/eclipse-theia/theia/wiki/Coding-Guidelines#internationalizationlocalization), the `nls.localizeByDefault` function should only be called when an appropriate value exists in vscode's language packs (represented by `nls.metadata.json` file). This file should not be modified, since it won't match strings that exist inside of vscode's language packs anymore.

Instead you should use `nls.localize` and introduce new keys like `theia/workspace/folderExists` and `theia/workspace/fileExists`.

packages/core/src/common/i18n/nls.metadata.json
```diff
@@ -16277,7 +16277,8 @@
 			"Open a file first to open in new window",
 			"A file or folder name must be provided.",
 			"A file or folder name cannot start with a slash.",
-			"A file or folder **{0}** already exists at this location. Please choose a different name.",
+			"A folder **{0}** already exists at this location. Please choose a different name.",
+			"A file **{0}** already exists at this location. Please choose a different name.",
```
msujew, line 16281: Please revert these changes.

msujew (commented):
Hi @xpomul, and thank you for your first contribution 👍 

While the behavior looks good at first glance, there is an issue regarding the localization, see below.

packages/workspace/src/browser/workspace-commands.ts
xpomul: Thanks. I misunderstood that part...
Changed now to `nls.localize`.

packages/core/src/common/i18n/nls.metadata.json
xpomul: reverted in the recent commit.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - [scripts] added Electron frontend start-up performance measurement script [#10442](https://github.com/eclipse-theia/theia/pull/10442) - Contributed on behalf of STMicroelectronics
 - [core, editor, editor-preview] additional commands added to tabbar context menu for editor widgets. [#10394](https://github.com/eclipse-theia/theia/pull/10394)
 - [preferences] Updated `AbstractResourcePreferenceProvider` to handle multiple preference settings in the same tick and handle open preference files. It will save the file exactly once, and prompt the user if the file is dirty when a programmatic setting is attempted. [#7775](https://github.com/eclipse-theia/theia/pull/7775)
+- [workspace]]The validation message when creating a file or folder is now more specific if a file or folder with the same name already exists in the respective location. [#10527](https://github.com/eclipse-theia/theia/pull/10527)
```
msujew, line 10: Minor: 

```suggestion
- [workspace] The validation message when creating a file or folder is now more specific if a file or folder with the same name already exists in the respective location. [#10527](https://github.com/eclipse-theia/theia/pull/10527)
```

msujew (commented):
Upon deeper inspection, I'm not so sure about the distinction between files and folders anymore: When creating a new file with the exact same name as an existing folder, we receive the message `A folder "X" already exists at this location.`.

I feel like this could lead to confusion to someone not-so-familiar with file-systems and their peculiarities. Of course, having a file and a folder with the same name within the same directory doesn't make a lot of sense. However, I'm not creating a folder with the same name, but a file. So why would Theia tell me that this is an illegal operation just because a folder with the same name already exists? Leaving the distinction open by using "A file __or__ folder" clarifies that having a file and folder occupying the same path isn't possible. I imagine this is the reason why VSCode originally uses the "A file or folder" message.

Perhaps I'm also overthinking this, I was just a bit perplexed by the message mentioning a folder when creating a file. @vince-fugnitto @colin-grant-work Your opinions on this?

colin-grant-work:
> Upon deeper inspection, I'm not so sure about the distinction between files and folders anymore: When creating a new file with the exact same name as an existing folder, we receive the message `A folder "X" already exists at this location.`.
> 
> I feel like this could lead to confusion to someone not-so-familiar with file-systems and their peculiarities. Of course, having a file and a folder with the same name within the same directory doesn't make a lot of sense. However, I'm not creating a folder with the same name, but a file. So why would Theia tell me that this is an illegal operation just because a folder with the same name already exists? Leaving the distinction open by using "A file **or** folder" clarifies that having a file and folder occupying the same path isn't possible. I imagine this is the reason why VSCode originally uses the "A file or folder" message.

I'm not sure how likely it is to cause confusion - I suspect most Theia users are familiar with the distinction between files and folders - but I also don't see much point in distinguishing between files and folders in this context. If either a file or a folder with the same name exists, we won't let you create a new file or a new folder, and the remedy is the same in either case: remove the conflict by deleting or changing the name of the existing item, or change the name you're trying to give the new item. Given that, I'm happy with one message that applies to all cases.

vince-fugnitto:
Like @colin-grant-work I'd be fine with keeping it as a generic message (which is already present in the `nls.metadata.json`):

https://github.com/eclipse-theia/theia/blob/7f954ac2561e73f9cc1e4c2df1e347ba9431b2c4/packages/core/src/common/i18n/nls.metadata.json#L16280

xpomul:
Ok, so what now? I close this pull-request and the issue?

And, what could I have done better? Should I always ask whether an issue is actually recognized as a valuable change before starting to work on it?

colin-grant-work:
@xpomul, I wouldn't say you've done anything wrong. We could have done better validating the issue and pruning it if it didn't fit with the plan, but the particular factors that make it desirable to keep the current wording - agreement with what's available in the existing localization setup - are a relatively recent addition to the code base. If in doubt, you can post a message on the issue checking that it's still relevant.

xpomul:
@colin-grant-work ah, thanks for clarifying. Ok, I’ll close this PR and the issue.

PR #10540, opened 2021-12-15
fix default Anchor param of context menu

shuyaqian:
Signed-off-by: shuyaqian 717749594@qq.com

#### What it does

Fix default Anchor param of context menu.

#### How to test
1. click 'Start Debugging' menu
image
2.  you will get error
image
3. because the param `debugSessionOptions` received is an anchor
image

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/context-menu-renderer.ts
+++ b/packages/core/src/browser/context-menu-renderer.ts
@@ -29,6 +29,13 @@ export function toAnchor(anchor: HTMLElement | Coordinate): Anchor {
     return anchor instanceof HTMLElement ? { x: anchor.offsetLeft, y: anchor.offsetTop } : anchor;
 }
 
+export function isAnchor(arg: unknown): boolean {
+    if (arg && typeof arg === 'object' && 'x' in arg && 'y' in arg && Object.keys(arg).length === 2) {
+        return true;
+    }
+    return false;
+}
+
 export function coordinateFromAnchor(anchor: Anchor): Coordinate {
     const { x, y } = anchor instanceof MouseEvent ? { x: anchor.clientX, y: anchor.clientY } : anchor;
     return { x, y };
--- a/packages/debug/src/browser/debug-frontend-application-contribution.ts
+++ b/packages/debug/src/browser/debug-frontend-application-contribution.ts
@@ -14,7 +14,7 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-import { AbstractViewContribution, ApplicationShell, KeybindingRegistry, Widget, CompositeTreeNode, LabelProvider, codicon } from '@theia/core/lib/browser';
+import { AbstractViewContribution, ApplicationShell, KeybindingRegistry, Widget, CompositeTreeNode, LabelProvider, codicon, isAnchor } from '@theia/core/lib/browser';
 import { injectable, inject } from '@theia/core/shared/inversify';
 import { ThemeService } from '@theia/core/lib/browser/theming';
 import { MenuModelRegistry, CommandRegistry, MAIN_MENU_BAR, Command, Emitter, Mutable } from '@theia/core/lib/common';
@@ -640,10 +640,10 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
     registerCommands(registry: CommandRegistry): void {
         super.registerCommands(registry);
         registry.registerCommand(DebugCommands.START, {
-            execute: (config?: DebugSessionOptions) => this.start(false, config)
+            execute: (config?: DebugSessionOptions) => this.start(false, isAnchor(config) ? undefined : config)
         });
         registry.registerCommand(DebugCommands.START_NO_DEBUG, {
-            execute: (config?: DebugSessionOptions) => this.start(true, config)
+            execute: (config?: DebugSessionOptions) => this.start(true, isAnchor(config) ? undefined : config)
         });
         registry.registerCommand(DebugCommands.STOP, {
             execute: () => this.manager.terminateSessions(),


colin-grant-work:
A similar change has been proposed previously, but it [was rejected](https://github.com/eclipse-theia/theia/pull/9932#pullrequestreview-734070992) because there was code that relied on the current behavior. I believe that [this recent change](https://github.com/eclipse-theia/theia/pull/10394/files#diff-bb62d77ca908eb463355fefc1920763f61f7880c48235f559da8d9cd91f8e753R460) should reduce our dependence on that behavior upstream, but there may be quite a lot of downstream code that still relies on it.

Given that, there are other possibilities for working around this problem in the current context.
 - Modify the handler for the command that is failing to account for the possibility of an `anchor` as the first argument.
 - Modify the code that opens that menu to explicitly pass `undefined` as the first argument (or first several arguments)

shuyaqian:
> A similar change has been proposed previously, but it [was rejected](https://github.com/eclipse-theia/theia/pull/9932#pullrequestreview-734070992) because there was code that relied on the current behavior. I believe that [this recent change](https://github.com/eclipse-theia/theia/pull/10394/files#diff-bb62d77ca908eb463355fefc1920763f61f7880c48235f559da8d9cd91f8e753R460) should reduce our dependence on that behavior upstream, but there may be quite a lot of downstream code that still relies on it.
> 
> Given that, there are other possibilities for working around this problem in the current context.
> 
> * Modify the handler for the command that is failing to account for the possibility of an `anchor` as the first argument.
> * Modify the code that opens that menu to explicitly pass `undefined` as the first argument (or first several arguments)

I defined the `isAnchor` function,  and verify the first argument.

packages/core/src/browser/context-menu-renderer.ts
```diff
@@ -29,6 +29,14 @@ export function toAnchor(anchor: HTMLElement | Coordinate): Anchor {
     return anchor instanceof HTMLElement ? { x: anchor.offsetLeft, y: anchor.offsetTop } : anchor;
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function isAnchor(arg: any): boolean {
+    if (arg && 'x' in arg && 'y' in arg && Object.keys(arg).length === 2) {
```
msujew, line 34: Suggestion: We try to limit our `any` usage in the framework.
```suggestion
export function isAnchor(arg: unknown): boolean {
    if (typeof arg === 'object' && arg && 'x' in arg && 'y' in arg && Object.keys(arg).length === 2) {
```

packages/debug/src/browser/debug-frontend-application-contribution.ts
```diff
@@ -1197,7 +1197,7 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
     }
 
     async start(noDebug?: boolean, debugSessionOptions?: DebugSessionOptions): Promise<void> {
-        let current = debugSessionOptions ? debugSessionOptions : this.configurations.current;
+        let current = debugSessionOptions && !isAnchor(debugSessionOptions) ? debugSessionOptions : this.configurations.current;
```
colin-grant-work, line 1200: It would be better to check this condition in the command handler (around line 640), since the possibility of receiving of anchor is limited to that context.

shuyaqian: > It would be better to check this condition in the command handler (around line 640), since the possibility of receiving of anchor is limited to that context.

ok， done

packages/core/src/browser/context-menu-renderer.ts
```diff
@@ -29,6 +29,13 @@ export function toAnchor(anchor: HTMLElement | Coordinate): Anchor {
     return anchor instanceof HTMLElement ? { x: anchor.offsetLeft, y: anchor.offsetTop } : anchor;
 }
 
+export function isAnchor(arg?: Anchor): boolean {
+    if (arg && 'x' in arg && 'y' in arg && Object.keys(arg).length === 2) {
+        return true;
+    }
+    return false;
+}
```
colin-grant-work, line 37: Is this correct and sufficient? The `Anchor` type also includes `MouseEvent`s (l. 26).

```typescript
export interface Coordinate { x: number; y: number; }
export const Coordinate = Symbol('Coordinate');
export type Anchor = MouseEvent | Coordinate;
```

shuyaqian: I changed the type to unknown.

packages/debug/src/browser/debug-frontend-application-contribution.ts
```diff
@@ -640,10 +640,10 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
     registerCommands(registry: CommandRegistry): void {
         super.registerCommands(registry);
         registry.registerCommand(DebugCommands.START, {
-            execute: (config?: DebugSessionOptions) => this.start(false, config)
+            execute: (config?: DebugSessionOptions) => this.start(false, isAnchor(config) ? undefined : config)
         });
         registry.registerCommand(DebugCommands.START_NO_DEBUG, {
-            execute: (config?: DebugSessionOptions) => this.start(true, config)
+            execute: (config?: DebugSessionOptions) => this.start(true, isAnchor(config) ? undefined : config)
```
colin-grant-work, line 646: Sorry to change my mind here, but even though the cause of the problem is that an Anchor is being passed in, but what we really want to guarantee is that the `config` is an instance of `DebugSessionOptions` - the problem with the `CommandHandler` system is that we're allowed to say `config?: DebugSessionOption`, but in fact anything can be passed in - Anchors, or anything else..

So rather than adding a check for whether this is an Anchor, it would be more correct in this context to implement a check that something is a DebugSessionOptions. Something like.

```typescript
export namespace DebugSessionOptions {
    export function is(candidate: unknown): candidate is DebugSessionOptions {
        const options = candidate as DebugSessionOptions;
        return typeof options === 'object' && !!candidate && 'configuration' in candidate && DebugConfiguration.is(options.candidate);
    }
}
```
Again, sorry for the change of direction!

colin-grant-work: I would say hold off on this pending a general resolution of the problems introduced by `window.menuBarVisibility` `compact`. As I say [here](https://github.com/eclipse-theia/theia/issues/10571#issuecomment-999769286), the correct, non-breaking solution is probably just to fix the behavior of the menu introduced by that PR, rather than fix the commands one by one or change the existing behavior of real context menus.

PR #9080, opened 2021-02-17
9075-Fix fetching of values from workspace config

balajiv113:
#### What it does

The changes fixes #9075. 
The root cause of the issue is resourceUri is set as false if resource is not present in overrides.
https://github.com/eclipse-theia/theia/blob/f3a15db2e364549922a843f195371f5a19a1a403/packages/monaco/src/browser/monaco-frontend-module.ts#L158
Due to this preference proxy is returning the default value which is flat.
#### How to test

**Setup**
Add the change from #8967 for reloading toolbar based on ToolbarItem.onDidChange.

Test 1 - To test if toolbar items are reloaded on configuration change 
Follow the steps mentioned under the issue #9075

Test 2 - To test if toolbar items are reloaded on context change
Follow the steps mentioned under the issue #8757 

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)
- [x] Update workspace config using preference UI and notice icon is getting updated
- [x] Clear `maven.view` from workspace config and update `maven.view` in user config and notice the behaviour is working fine

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #10546, opened 2021-12-15
Improve plugin-view toolbar lifecycle

### LINKED ISSUE #9075: Icons (and Their Matching Commands ) of view/title Do not Update Upon Changes in Workspace settings.json
### Bug Description:

The Maven for Java extension uses 2 icons in its view/title menu:
https://github.com/microsoft/vscode-maven/blob/master/package.json#L290-L298

flat

hierarchical

The "flat" and "hierarchical" icons should replace one another upon clicking on them:

When the user clicks on the icon, a change is made to the Workspace Configuration:
https://github.com/microsoft/vscode-maven/blob/master/src/Settings.ts#L20-L26

The icon at view/title does not change after clicking (and so is the matching command it exposes).

When removing the matching configuration "maven.view" from the Workspace settings.json, and pasting it in the User settings.json, upon manual change of the value (between "flat" and "hierarchical"), the icon changes accordingly and as expected:

flat

hierarchical

Bottom line - It looks like the "when" clause in the package.json that uses "config":

`"when": "view == mavenProjects && config.maven.view != flat",`
(from https://github.com/microsoft/vscode-maven/blob/master/package.json#L291)

Listens to changes of the User configurations and not the Workspace configurations.

### Steps to Reproduce:

1. Open https://gitpod.io/#https://github.com/gitpod-io/spring-petclinic
2. Install the "Maven for Java" Extension.
3. In the Maven menu (in Explorer Tab), click on the icon "Switch to hierarchical view" - NOTHING HAPPENS.

4. Open File -> Settings -> Preferences.
5. In Preferences tab -> Workspaces -> Open settings.json file.
6. Remove `"maven.view": "hierarchical"`
7. In Preferences tab -> User-> Open settings.json file.
8. Add `"maven.view": "hierarchical"`. See Icon. Change value from "hierarchical" to "flat". See how the icon changes.

- Operating System: gitpod
- Theia Version: gitpod

colin-grant-work:
#### What it does

Fixes #9075.
Closes #9080.
 This PR builds on #9080. That PR addressed a bug in the fetching of values from workspace scope in the `MonacoConfigurationService`, but did not fix the problem with missing events for updating toolbars in plugin-contributed views. This PR makes several changes to address the latter problem:
 - Removes the `.with` utility from the `ViewContextKeyService`. That utility was designed to check whether a `when` expression's conditions would be satisfied given certain assumptions, but it caused two or four change events to be fired by the `ContextKeyService` for every check. It has been replaced by a `.with` utility on the `MonacoContextKeyService` class that uses Monaco utilities to create subcontexts so that the events don't bubble to the global context.
 - Adds a local context to plugin view widgets so that tabbar toolbar checks no longer need to use the `.with` utility at all.
 - Separates the `ContextKeyService` interface from the mostly useless dummy implementation in `core`. This is a breaking change (extenders can no longer `extends ContextKeyService`), but it allows us to document the functionality of the methods. Previously each method was just annotated with `Should be implemented elsewhere`.

#### How to test

- Add the Maven plugin.
- Click on the flat / hierarchical item.
- Observe that when you click, it changes to the other item consistently.
- Check that other tabbar toolbar items and context-dependent items are working as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,9 @@
 
 <a name="breaking_changes_1.22.0">[Breaking Changes:](#breaking_changes_1.22.0)</a>
 - [core] Removed `MarkdownRenderer` class [#10589](https://github.com/eclipse-theia/theia/pull/10589)
-- [plugin] Removed deprecated fields `id` and `label` from `theia.Command`
+- [core] `ContextKeyService` is now an interface. Extenders should extend `ContextKeyServiceDummyImpl` [#10546](https://github.com/eclipse-theia/theia/pull/10546)
+- [plugin] Removed deprecated fields `id` and `label` from `theia.Command` [#10512](https://github.com/eclipse-theia/theia/pull/10512)
+- [plugin-ext] `ViewContextKeyService#with` method removed. Use `ContextKeyService#with` instead. `PluginViewWidget` and `PluginTreeWidget` inject the `ContextKeyService` rather than `ViewContextKeyService`. [#10546](https://github.com/eclipse-theia/theia/pull/10546)
 
 
 ## v1.21.0 - 12/16/2021
--- a/packages/core/src/browser/context-key-service.ts
+++ b/packages/core/src/browser/context-key-service.ts
@@ -15,7 +15,8 @@
  ********************************************************************************/
 
 import { injectable } from 'inversify';
-import { Emitter } from '../common/event';
+import { Disposable } from '../common';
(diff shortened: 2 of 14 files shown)


msujew (approved):
I can confirm that the issue exists on `master` and is addressed by this change.

- [x] The file tree/list icon correctly changes using the maven extension
- [x] Other buttons remain working as expected (no regressions)

The code looks good to me 👍

PR #10554, opened 2021-12-16
use lerna lifecycle scripts

paul-marechal:
When publishing, the version of packages are changed but some generated
files were left out of date (e.g. core/README.md shared packages).

Remove `publish` script, we can do without it.

Add `version` script in core to ensure the README.md is up to date.


vince-fugnitto (approved):
LGTM :+1:

PR #10570, opened 2021-12-21
Fix memory leak in ApplicationShell#activateWidget

colin-grant-work:
#### What it does

The `ApplicationShell#activateWidget` method `await`s a `Promise` returned by `ApplicationShell#waitForActivation`, but that `Promise` is not guaranteed to be settled. In particular, if the widget in question fails to accept focus within 2 seconds, it will not be set as the `ApplicationShell`'s `activeWidget`. Because of that, a call to `activateWidget` can hang indefinitely, and the `Promise` handlers retain references to the widget that was to be activated even if that widget is later closed.

This code ensures that the `Promise` returned by `waitForActivation` will settle so that `activateWidget` will run to completion.

#### How to test

1. Comment out the implementation of `onActivateRequest` in `vsx-extensions-view-container.tsx`.
2. Run the application and open the extensions view using the `View` menu.
3. In the console, see a message about the widget's failure to accept focus in a timely fashion.
4. Close the extensions view widget.
5. Open the dev tools and take a memory snapshot. Search for `VSXExtensionsViewContainer`
6. You should _not_ see any retained references, and in particular, you should _not_ see this as a retainer:

image

That chain indicates that the view container is being retained (in part) because of the promise in `ApplicationShell#activateWidget`.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/application-shell.ts
+++ b/packages/core/src/browser/shell/application-shell.ts
@@ -37,6 +37,7 @@ import { Emitter } from '../../common/event';
 import { waitForRevealed, waitForClosed } from '../widgets';
 import { CorePreferences } from '../core-preferences';
 import { BreadcrumbsRendererFactory } from '../breadcrumbs/breadcrumbs-renderer';
+import { Deferred } from '../../common/promise-util';
 
 /** The class name added to ApplicationShell instances. */
 const APPLICATION_SHELL_CLASS = 'theia-ApplicationShell';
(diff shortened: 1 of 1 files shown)


packages/core/src/browser/shell/application-shell.ts
```diff
@@ -1076,25 +1077,27 @@ export class ApplicationShell extends Widget {
...
+        const success = this.onDidChangeActiveWidget(() => {
+            if (this.activeWidget && this.activeWidget.id === id) {
+                activation.resolve();
+            }
+        });
+        const failure = setTimeout(() => activation.reject(new Error(`Widget with id '${id}' failed to activate.`)), this.activationTimeout + 250);
+        return activation.promise.finally(() => {
+            success.dispose();
+            clearTimeout(failure);
         });
```
paul-marechal, line 1101:
```suggestion
        return new Promise((resolve, reject) => {
            let listener, timeout;
            listener = this.onDidChangeActiveWidget(() => {
                if (this.activeWidget && this.activeWidget.id === id) {
                    clearTimeout(timeout);
                    listener.dispose();
                    resolve();
                }
            });
            timeout = setTimeout(() => {
                listener.dispose();
                reject(new Error(`Widget with id '${id}' failed to activate.`))
            }, this.activationTimeout + 250);
        });
```

I rarely see the use for `Deferred` objects...

colin-grant-work: I don't really see any advantage to rewriting it like this. The main advantages of the `Deferred` in this case to my mind are that (mainly) you can refer to the things you need to dispose of in a `finally` rather than having to write the cleanup as part of the resolution or rejection and (much less importantly), it saves one level of indentation and mental effort tracking what's available in what scope.

paul-marechal (approved):

Issue #10572, opened 2021-12-22
showOpenDialog fails to fall back to workspace root


mcgordonite:
### Bug Description:

The `showOpenDialog` plugin API accepts an optional `defaultUri`, the directory to first show in the file dialog. If not provided, it should fall back to the workspace root. However, this fall back fails due to a bug in the implementation, leading to the file system root being shown instead.

This also applies to `showSaveDialog` and `showUploadDialog`.

### Steps to Reproduce:

This is easy to reproduce using the browser example and the vscode-xml extension from Open-VSX:
1. Start the browser example and open a folder.
2. Use the Extensions widget to install Red Hat's XML extension from Open VSX.
3. Create an XML file in the workspace, e.g. example.xml:
    ```xml
    <?xml version="1.0" encoding="UTF-8" standalone="no" ?>
    <root></root>
    ```
4. With the file focused, run the "XML: Bind to grammar/schema file" command.
5. Select file association from the quick picker.

Expected behaviour: a file dialog should open at the root of the workspace. Instead, the dialog opens at the root of the file system.

The XML extension calls `showOpenDialog` without a `defaultUri`, see the implementation: https://github.com/redhat-developer/vscode-xml/blob/7f3a13c013754fab8b8f1a6e47acaf32330a647b/src/commands/registerCommands.ts#L181.

### Additional Information

- Operating System: Windows 10
- Theia Version: 1.21.0

Issue #7451, opened 2020-03-30
[showSaveDialog] different behavior from VSCode


veredcon:
### Description
showSaveDialog API acts differently from VSCode in the following scenario:
In theia it basically does not allow to send URI that does not yet exist. If the API is called on a file URI that does not yet exist, the fallback would be according to this logic: https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/main/browser/dialogs-main.ts#L45-L62 (workspace root folder etc.)

However in VSCode, it allows to send URI (defaultUri) that does not exist and it will open the dialog according to that.

### Reproduction Steps
I created a simple VSCode Extension to demonstrate this issue and a simple project:
https://github.com/veredcon/vscodetests/blob/master/src/extension.ts#L19-L33 

1. Open theia locally (v0.16.1 but can be reproduced also in later versions)
2. Use this simple project attached: 
[simpleproject.zip](https://github.com/eclipse-theia/theia/files/4404378/simpleproject.zip)
3. Add the following simple extension vsix:
[vscode-ext-sample-0.0.2.zip](https://github.com/eclipse-theia/theia/files/4404390/vscode-ext-sample-0.0.2.zip)
4. Click on file1.txt --> "Show Save Dialog" 
![image](https://user-images.githubusercontent.com/9718939/77942322-1d5cd100-72c4-11ea-8811-8b0b535beeab.png)
You can see the dialog is opened like this:
![image](https://user-images.githubusercontent.com/9718939/77942386-32396480-72c4-11ea-870c-5c5c5a5b759f.png)

but in VSCode it looks better because it navigates to the parent folder:

![image](https://user-images.githubusercontent.com/9718939/77942728-bab80500-72c4-11ea-8f50-976bbed75dc5.png)

5. Click on file2.json --> "Show Save Dialog". In this sample extension, json suffix is replaced with ".txt" so the URI file2.txt does not exist and you can see the dialog is opened like this:
![image](https://user-images.githubusercontent.com/9718939/77942440-48dfbb80-72c4-11ea-9d4c-1cdb82d271de.png)

but in VSCode it is opened properly:
![image](https://user-images.githubusercontent.com/9718939/77942823-e5a25900-72c4-11ea-9ec6-7cd13e06de1e.png)

**OS and Theia version:**
OS: windows/linux (does not matter)
Theia: 0.16.1 (and after..)

**Diagnostics:**

vince-fugnitto:
[...]

PR #10573, opened 2021-12-22
Fix showOpenDialog fallback

mcgordonite:
#### What it does

Fixes https://github.com/eclipse-theia/theia/issues/10572 by correcting logic in the plugin extension. The `showOpenDialog` and other file dialog extension APIs now correctly fall back to the workspace root when the `defaultUri` is not specified.

#### How to test

Fallback to the workspace root when a `defaultUri` is not specified can be tested with the vscode-xml extension as described in https://github.com/eclipse-theia/theia/issues/10572.

The behaviour when a `defaultUri` is present can be tested as described in https://github.com/eclipse-theia/theia/issues/7451. In this case, the file dialog should open at the parent folder when a file is passed as the `defaultUri`.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/dialogs-main.ts
+++ b/packages/plugin-ext/src/main/browser/dialogs-main.ts
@@ -52,14 +52,14 @@ export class DialogsMainImpl implements DialogsMain {
             } catch {
                 rootStat = undefined;
             }
-        }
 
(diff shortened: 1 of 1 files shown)


colin-grant-work (approved):
I confirm the bug on master and that the fix prevents early fallback to the filesystem root.

PR #10580, opened 2021-12-23
Fix the widget options when opening a VSCode Custom Editor

### LINKED ISSUE #10583: vscode.openWith command does not propagate widgetOptions for Custom Editors
### Bug Description:

When using the *[vscode.openWith](https://code.visualstudio.com/api/references/commands)* with a columnOrOptions argument in a VSCode extension, the Custom Editor in Theia will not open in the right area.

I created a sample extension from official samples @ https://github.com/debovema/vscode-extension-samples/tree/custom-editor-column/custom-editor-sample

Especially I added the vscode.ViewColumn.Beside argument @ https://github.com/debovema/vscode-extension-samples/blob/custom-editor-column/custom-editor-sample/src/pawDrawEditor.ts#L212

A packaged version is available @ https://github.com/debovema/vscode-extension-samples/blob/custom-editor-column/custom-editor-sample/cat-customs-0.0.1.vsix

### Steps to Reproduce:

1. Clone my fork : ```git clone https://github.com/debovema/theia.git && cd theia```
2. Use latest release (1.21.0) : ```git checkout v1.21.0```
3. Download the cat-customs-0.0.1.vsix sample (see above) to plugins directory: ```mkdir -p plugins && wget https://github.com/debovema/vscode-extension-samples/raw/custom-editor-column/custom-editor-sample/cat-customs-0.0.1.vsix -O plugins/cat-customs-0.0.1.vsix```
4. Build and start: ```yarn && yarn browser build && yarn browser start```
5. In Theia create two Paw Draw documents (press F1 then ">Paw Draw: Create new Paw Draw Document") => the two Custom Editors are created in the main area

image
image

### Steps to Resolve:
1. Still on my fork, checkout the fix branch: ```git checkout fix-custom-editor-widget-options```
2. Build and start: ```yarn && yarn browser build && yarn browser start```
3. In Theia create two Paw Draw documents (press F1 then ">Paw Draw: Create new Paw Draw Document") => the second Custom Editor is created to the right of the first one

image

### Additional Information

- Operating System: Ubuntu 20.04.3 LTS (Focal Fossa)
- Theia Version: 1.21.0

Proposed fixed : #10580

debovema:
#### What it does
Fixes #10583

#### How to test
See #10583

#### Review checklist

### DIFF
--- a/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
+++ b/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
@@ -179,12 +179,7 @@ export class PluginVscodeCommandsContribution implements CommandContribution {
         }
 
         if (openHandler) {
-            const widget = await openHandler.open(uri, editorOptions);
-            if (widget instanceof Widget) {
-                if (!widget.isAttached) {
-                    await this.shell.addWidget(widget, editorOptions?.widgetOptions);
-                }
-            }
+            await openHandler.open(uri, editorOptions);
             return true;
         }
 
--- a/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
+++ b/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
@@ -179,7 +179,12 @@ export class PluginVscodeCommandsContribution implements CommandContribution {
         }
 
         if (openHandler) {
-            await openHandler.open(uri, editorOptions);
+            const widget = await openHandler.open(uri, editorOptions);
+            if (widget instanceof Widget) {
+                if (!widget.isAttached) {
+                    await this.shell.addWidget(widget, editorOptions?.widgetOptions);
+                }
+            }
             return true;
         }
 
--- a/packages/plugin-ext/src/common/plugin-api-rpc.ts
+++ b/packages/plugin-ext/src/common/plugin-api-rpc.ts
@@ -93,7 +93,7 @@ import type {
 import { SerializableEnvironmentVariableCollection } from '@theia/terminal/lib/common/base-terminal-protocol';
 import { ThemeType } from '@theia/core/lib/common/theme';
 import { Disposable } from '@theia/core/lib/common/disposable';
-import { PickOptions, QuickInputButtonHandle, QuickPickItem } from '@theia/core/lib/browser';
+import { PickOptions, QuickInputButtonHandle, QuickPickItem, WidgetOpenerOptions } from '@theia/core/lib/browser';
 
 export interface PreferenceData {
     [scope: number]: any;
@@ -1555,7 +1555,7 @@ export interface CustomEditorsExt {
         newWebviewHandle: string,
         viewType: string,
         title: string,
-        position: number,
+        widgetOpenerOptions: WidgetOpenerOptions | undefined,
         options: theia.WebviewPanelOptions,
         cancellation: CancellationToken): Promise<void>;
     $createCustomDocument(resource: UriComponents, viewType: string, backupId: string | undefined, cancellation: CancellationToken): Promise<{ editable: boolean }>;
@@ -1578,7 +1578,7 @@ export interface CustomEditorsMain {
     $registerTextEditorProvider(viewType: string, options: theia.WebviewPanelOptions, capabilities: CustomTextEditorCapabilities): void;
     $registerCustomEditorProvider(viewType: string, options: theia.WebviewPanelOptions, supportsMultipleEditorsPerDocument: boolean): void;
     $unregisterEditorProvider(viewType: string): void;
-    $createCustomEditorPanel(handle: string, title: string, viewColumn: theia.ViewColumn | undefined, options: theia.WebviewPanelOptions & theia.WebviewOptions): Promise<void>;
+    $createCustomEditorPanel(handle: string, title: string, widgetOpenerOptions: WidgetOpenerOptions | undefined, options: theia.WebviewPanelOptions & theia.WebviewOptions): Promise<void>;
     $onDidEdit(resource: UriComponents, viewType: string, editId: number, label: string | undefined): void;
     $onContentChange(resource: UriComponents, viewType: string): void;
 }
--- a/packages/plugin-ext/src/main/browser/custom-editors/custom-editor-opener.tsx
+++ b/packages/plugin-ext/src/main/browser/custom-editors/custom-editor-opener.tsx
@@ -28,7 +28,7 @@ export class CustomEditorOpener implements OpenHandler {
     readonly id: string;
     readonly label: string;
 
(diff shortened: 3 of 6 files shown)


packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
```diff
@@ -179,7 +179,12 @@ export class PluginVscodeCommandsContribution implements CommandContribution {
         }
 
         if (openHandler) {
-            await openHandler.open(uri, editorOptions);
+            const widget = await openHandler.open(uri, editorOptions);
+            if (widget instanceof Widget) {
+                if (!widget.isAttached) {
+                    await this.shell.addWidget(widget, editorOptions?.widgetOptions);
+                }
```
colin-grant-work, line 186: @debovema, thanks for your contribution! I believe the code is good, but should be moved elsewhere. See below.

It looks like this is should be the responsibility of the open handler for the widget. For example, here's the default `WidgetOpenHandler`:

https://github.com/eclipse-theia/theia/blob/17b419fdf0d1015f2269ffcb64a095b305a8ec2d/packages/core/src/browser/widget-open-handler.ts#L98-L100

But the same is not done when a new custom editor is created:

https://github.com/eclipse-theia/theia/blob/17b419fdf0d1015f2269ffcb64a095b305a8ec2d/packages/plugin-ext/src/main/browser/custom-editors/custom-editor-opener.tsx#L83-L87

I think that for consistency's sake, that's the place to handle this issue.

colin-grant-work: Alternatively, it looks like this is where the attachment is handled, at the moment: 

https://github.com/eclipse-theia/theia/blob/17b419fdf0d1015f2269ffcb64a095b305a8ec2d/packages/plugin-ext/src/main/browser/custom-editors/custom-editors-main.ts#L234-L240

colin-grant-work:
The other thing to consider here is that VSCode opener options, which is what is expected from plugins, differ from Theia's opener options, so there may be opposition to applying the Theia logic to a case where it is not expected to apply. It should at least be avoided that the Theia logic is applied when no `widgetOptions` field is present: in that case, the existing handler should be used.

packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
debovema: Hi @colin-grant-work 

Thank you for your review.

The issue with your suggestion to move the logic in the Custom Editor Opener (*custom-editor-opener.tsx*) is that we don't have the info provided by VSCode extension being opened in this location.

The other place in ```$createCustomEditorPanel``` method in *custom-editors-main.ts* does receive a viewColumn parameter but AFAICS this parameter is hardcoded to ```EditorPosition.ONE```, because of this:

https://github.com/eclipse-theia/theia/blob/17b419fdf0d1015f2269ffcb64a095b305a8ec2d/packages/plugin-ext/src/main/browser/custom-editors/custom-editors-main.ts#L152

If I understand well, the call to ```await openHandler.open(uri, editorOptions);``` actually delegates to the ```CustomEditorOpener``` class (this class is not injectable with inversify so I assume it is created here : https://github.com/eclipse-theia/theia/blob/17b419fdf0d1015f2269ffcb64a095b305a8ec2d/packages/plugin-ext/src/main/browser/custom-editors/plugin-custom-editor-registry.ts#L84

So it might be the simplest solution to add the behaviour in my original place, don't you think ?

colin-grant-work: > The issue with your suggestion to move the logic in the Custom Editor Opener (custom-editor-opener.tsx) is that we don't have the info provided by VSCode extension being opened in this location.

I'm not sure what you mean by this. As you say

> the call to await openHandler.open(uri, editorOptions); actually delegates to the CustomEditorOpener

the call to `openHandler.open` calls into `CustomEditorOpener.open`, so if we have the information we need where you've added it, then we have that very same information in `custom-editor-opener.tsx`, because we're passing that information to the `CustomEditorOpener`, and we have access to the `ApplicationShell` there, as well, so there's nothing in the code you've added that we couldn't do in the `CustomEditorOpener` class.

Regarding the hardcoded `EditorPosition.ONE` and the comment  there, I don't think I understand what they mean. Placing a widget in a particular column doesn't necessarily have anything to do with whether the widget can be split, but I haven't worked much with custom editors, myself, so there may be some logic there that I'm missing. To my mind, the problem has nothing to do with the splittability of editors but rather with whether Theia correctly handles the `viewColumn` field, and there is code that appears to try to do so:

https://github.com/eclipse-theia/theia/blob/d7cdb85c1793ae5e4af8d05da249a2dcea606da2/packages/plugin-ext/src/main/browser/webviews-main.ts#L108-L114

So it may be worth simply swapping out that hardcoded value with whatever is contextually appropriate to see if that addresses the problem.

debovema: > the call to `openHandler.open` calls into `CustomEditorOpener.open`, so if we have the information we need where you've added it, then we have that very same information in `custom-editor-opener.tsx`, because we're passing that information to the `CustomEditorOpener`, and we have access to the `ApplicationShell` there, as well, so there's nothing in the code you've added that we couldn't do in the `CustomEditorOpener` class.

I pushed another commit moving the logic in the CustomEditorOpener.

Please notice that the options parameter has been changed from ```OpenerOptions``` to ```WidgetOpenerOptions```.
I hope there are no impact elsewhere.

The general behaviour is fine but there are some errors in the console worth investigating :
image

debovema: @colin-grant-work : eventually I fully reworked the propagation of WidgetOpenerOptions from CustomEditorOpener to CustomEditorsMainImpl

colin-grant-work: @debovema, 

> @colin-grant-work : eventually I fully reworked the propagation of WidgetOpenerOptions from CustomEditorOpener to CustomEditorsMainImpl

Does the latest code cover functionality apart from what's described in #10583, or do the testing steps remain the same?

debovema: No it covers the exact same original issue.
But the implementation is more consistent.

colin-grant-work (approved):
The new code looks good to me and handles options passed by the plugin consistently and correctly.

debovema:
Hi,

Is this PR in the waiting line for being merged ?

msujew (approved):
@debovema We sometimes wait for a few days before merging to get input from any willing other reviewers as well :)

Anyway, this looks good to me as well, so merge at your leisure @colin-grant-work 👍

debovema:
Thank you @msujew for the follow-up and @colin-grant-work for the review !
👍 🎉

PR #10574, opened 2021-12-22
Fix handling of localhost links in the ExternalUriService

AlexandraBuzila:
#### What it does
In the ExternalUriService, don't try to resolve localhost URIs when running in electron.

#### How to test
1. Start the electron example app
2. Print a link to localhost in the terminal, e.g. `localhost:3000`
3. Ctrl+click on the link
4. The link should open in an external browser.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #10590, opened 2022-01-03
core: properly handle localhost uris on Electron

paul-marechal:
When running Electron `window.location` is not a standard HTTP(S) URL,
so we cannot use it to determine the host where the backend is running.

Resolve the remote host to `localhost` when running Electron.

Closes #10574

#### How to test

1. Start the electron example app
2. Print a link to localhost in the terminal, e.g. localhost:3000
3. Ctrl+click on the link
4. The link should open in an external browser.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/external-uri-service.ts
+++ b/packages/core/src/browser/external-uri-service.ts
@@ -14,10 +14,16 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
+import { environment } from '@theia/application-package/lib/environment';
 import { injectable } from 'inversify';
-import URI from '../common/uri';
 import { MaybePromise } from '../common/types';
+import URI from '../common/uri';
 import { Endpoint } from './endpoint';
+
+export interface AddressPort {
+    address: string
+    port: number
+}
 
 @injectable()
 export class ExternalUriService {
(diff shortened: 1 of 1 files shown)


packages/core/src/browser/external-uri-service.ts
```diff
@@ -31,34 +37,43 @@ export class ExternalUriService {
...
+    protected toRemoteUrl(uri: URI, address: AddressPort): URI {
+        return new Endpoint({ host: this.toRemoteHost(address) })
+            .getRestUrl()
+            .withPath(uri.path)
+            .withFragment(uri.fragment)
+            .withQuery(uri.query);
+    }
+
+    protected toRemoteHost(address: AddressPort): string {
+        return `${this.getRemoteHost()}:${address.port}`;
```
colin-grant-work, line 70: Will there always be a port here?

colin-grant-work:
This looks pretty reasonable to me. It leaves a little bit more room for later customization than #10574.

packages/core/src/browser/external-uri-service.ts
paul-marechal: Given how `AddressPort` works/is defined: yes.

colin-grant-work: Yes, I suppose there likely will be, since this is only called for `localhost`.

colin-grant-work (approved):
This looks good to me. @AlexandraBuzila, does this address your concerns with the current handling of `localhost`?

vince-fugnitto (approved):

On `master`, using `echo localhost:3000` in the terminal and attempting to click the link results in the following error:

```
cat: standard output: Bad file descriptor
sh: echo: I/O error
```

The same use-case works well with this change (the link is opened in an external browser).

AlexandraBuzila:
> This looks good to me. @AlexandraBuzila, does this address your concerns with the current handling of `localhost`?

Thank you for the changes, they work for me!

Issue #10552, opened 2021-12-16
Strange problem with in conditional operator


bd82:
I have a strange problem where I am trying to test the support for the "when clause in conditional operator"
using this small sample I've created:
- https://github.com/bd82/vscode-extension-samples/tree/main/in-conditional-operator-sample
  - The example includes documentation and relevant scripts to create a vsix to deploy.

The **strangeness** of the problem is that this works successfully when I build Theia browser example on this repo. (master / 1.20 / 1.19).
But does not work on both:
- gitpod (does gitpod still use Theia?).
- SAP BAS (uses Theia 1.19 when tested).

Does anyone have any thoughts on what could be causing this issue?
Could a 3rd party extension somehow break the custom context support?

Thanks in advance.
Shahar.

bd82:
Could it be that the `uri.fsPath` value works on Windows systems in combination with the setContext command and when operators.

- https://github.com/bd82/vscode-extension-samples/blob/main/in-conditional-operator-sample/src/contexts.ts#L34

But not on unix systems?

ChayaLau:
> * gitpod (does gitpod still use Theia?).

gitpod is using vscode, but when we run theia on gitpod the extension also does not work

colin-grant-work:
The menu items are appearing correctly in my Linux environment in both the Electron application and in the browser. Changes to the `package.json` files are also successfully picked up.

msujew:
@colin-grant-work Interestingly enough, it works correctly on Ubuntu (at least in my VM):

![image](https://user-images.githubusercontent.com/4377073/146613838-cd553599-016a-4deb-8749-69b5d86aa3c3.png)

But not on Windows (Electron & Browser):

![image](https://user-images.githubusercontent.com/4377073/146613885-8d5679c5-a7bf-4a27-8d6a-2e157a01b2fd.png)

I assume that @bd82 is correct in his assumption that it's somehow related to windows file path calculation of the `vscode-uri` package.

tsmaeder:
There is an open PR related to uri's and windows: https://github.com/eclipse-theia/theia/pull/10434 Maybe it's related.

bd82:
Thanks everyone for the feedback.
I will continue investigating it tomorrow and update if I have any new information

ChayaLau:
> There is an open PR related to uri's and windows: #10434 Maybe it's related.

hi, 
I checked this fix, but it wasn't enough.
[...]

PR #10591, opened 2022-01-04
Path conversion for windows and posix file systems

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/10552

Enables to convert URIs in the frontend to a file system path independent of the frontend OS (unlike `vscode-uri.fsPath`). This can be used to display backend paths, such as the file paths of opened editors in the `title` property.

#### How to test

This one is a bit difficult to test. The easiest setup would be to run Theia on a Unix machine and access the browser version on Windows.

---

1. Open a file in the editor. The `title` (hover display) of the file node should always display the correct backend path. I.e. `C:\file\to\path.txt` on windows and `/path/to/file.txt` on Unix.

---

0. Start Theia on a different OS than your frontend OS.
1. Download and install [this extension](https://github.com/eclipse-theia/theia/files/7808965/dynamic-context-menus-0.0.1.zip) ([source](https://github.com/bd82/vscode-extension-samples/tree/main/in-conditional-operator-sample))
2. Open the context-menu for `package.json` files which contain a `dependencies` section. The `Install Dependencies` menu entry should be visible.
3. Confirm that other configurations (electron/windows/unix) work as expected

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/resource-context-key.ts
+++ b/packages/core/src/browser/resource-context-key.ts
@@ -19,6 +19,7 @@ import URI from '../common/uri';
 import { URI as Uri } from 'vscode-uri';
 import { ContextKeyService, ContextKey } from './context-key-service';
 import { LanguageService } from './language-service';
+import { ApplicationServer } from '../common/application-protocol';
 
 @injectable()
 export class ResourceContextKey {
@@ -29,6 +30,9 @@ export class ResourceContextKey {
     @inject(ContextKeyService)
     protected readonly contextKeyService: ContextKeyService;
 
+    @inject(ApplicationServer)
+    protected readonly applicationService: ApplicationServer;
+
     protected resource: ContextKey<Uri>;
     protected resourceSchemeKey: ContextKey<string>;
     protected resourceFileName: ContextKey<string>;
@@ -38,7 +42,7 @@ export class ResourceContextKey {
     protected resourcePath: ContextKey<string>;
 
     @postConstruct()
-    protected init(): void {
+    protected async init(): Promise<void> {
         this.resource = this.contextKeyService.createKey<Uri>('resource', undefined);
         this.resourceSchemeKey = this.contextKeyService.createKey<string>('resourceScheme', undefined);
         this.resourceFileName = this.contextKeyService.createKey<string>('resourceFilename', undefined);
@@ -59,8 +63,8 @@ export class ResourceContextKey {
         this.resourceFileName.set(resourceUri && resourceUri.path.base);
         this.resourceExtname.set(resourceUri && resourceUri.path.ext);
         this.resourceLangId.set(resourceUri && this.getLanguageId(resourceUri));
-        this.resourceDirName.set(resourceUri && Uri.parse(resourceUri.path.dir.toString()).fsPath);
-        this.resourcePath.set(resourceUri && resourceUri['codeUri'].fsPath);
+        this.resourceDirName.set(resourceUri && resourceUri.path.dir.fsPath());
+        this.resourcePath.set(resourceUri && resourceUri.path.fsPath());
     }
 
     protected getLanguageId(uri: URI | undefined): string | undefined {
--- a/packages/core/src/browser/resource-context-key.ts
+++ b/packages/core/src/browser/resource-context-key.ts
@@ -19,7 +19,6 @@ import URI from '../common/uri';
 import { URI as Uri } from 'vscode-uri';
 import { ContextKeyService, ContextKey } from './context-key-service';
 import { LanguageService } from './language-service';
-import { ApplicationServer } from '../common/application-protocol';
 
 @injectable()
 export class ResourceContextKey {
@@ -30,9 +29,6 @@ export class ResourceContextKey {
     @inject(ContextKeyService)
     protected readonly contextKeyService: ContextKeyService;
 
-    @inject(ApplicationServer)
-    protected readonly applicationService: ApplicationServer;
-
     protected resource: ContextKey<Uri>;
     protected resourceSchemeKey: ContextKey<string>;
     protected resourceFileName: ContextKey<string>;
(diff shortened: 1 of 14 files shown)


bd82:
Thanks @msujew 

At the moment we have a workaround and also had the internal feature that needed the fix/workaround reduced in priority
So I am unsure when we will get around to testing it 😢

msujew:
@bd82 Not a problem, I actually tackled this issue due to something different that I noticed, it just also fixes the issue you were experiencing ;)

packages/core/src/browser/resource-context-key.ts
```diff
@@ -29,23 +31,28 @@ export class ResourceContextKey {
...
-    protected init(): void {
+    protected async init(): Promise<void> {
         this.resource = this.contextKeyService.createKey<Uri>('resource', undefined);
         this.resourceSchemeKey = this.contextKeyService.createKey<string>('resourceScheme', undefined);
         this.resourceFileName = this.contextKeyService.createKey<string>('resourceFilename', undefined);
         this.resourceExtname = this.contextKeyService.createKey<string>('resourceExtname', undefined);
         this.resourceLangId = this.contextKeyService.createKey<string>('resourceLangId', undefined);
         this.resourceDirName = this.contextKeyService.createKey<string>('resourceDirName', undefined);
         this.resourcePath = this.contextKeyService.createKey<string>('resourcePath', undefined);
+        this.isPosix = await this.applicationService.getBackendOS() !== OS.Type.Windows;
```
colin-grant-work, line 55: @msujew, it looks like you could import `isWindows` from `core/src/common/os.ts` rather than relying on the ApplicationServer. That would give you what you need synchronously rather than waiting on a `Promise`.

paul-marechal: Except `isWindows` will give the OS of the running browser here, and not the remote backend? Not sure which one makes more sense here, but they aren't the same.

msujew: @colin-grant-work Exactly as @paul-marechal is saying, in this case we need the backend OS, since it specifies the string format of all file paths (forward- vs. backward-slashes, drive letters in front, etc.). The whole reason why this bug exists, is that the frontend and backend might convert the same URI into different strings, which would then result in a mismatch of the `when` context property of commands.

In particular, when we have something like `resourcePath in ext:pkgJsonWithTestScript` as a context, the monaco frontend on windows will convert the current `resourcePath` into something like `\user\settings.json`, while the posix backend path is actually using `/user/settings.json`. Monaco will then perform an `in` check on the `ext:pkgJsonWithTestScript` object, which will fail.

msujew:
@colin-grant-work Care to take another look at this? I also opened https://github.com/eclipse-theia/theia/issues/10830 to keep track of all misusage of `URI#path.toString()`.

colin-grant-work:
@msujew, I've got a brand-new Windows system I should be able to test this on, so I'll take another look today :-)

packages/core/src/common/path.ts
```diff
@@ -70,6 +70,35 @@ export class Path {
...
+            const driveLetter = path.charAt(offset).toUpperCase();
+            const substring = path.substring(offset + 2).replace(/\//g, '\\');
+            return `${driveLetter}:${substring || '\\'}`;
+        }
+        return path.replace(/\//g, '\\');
+    }
+
+    /**
+     * Creates a posix path from the given path string.
+     * A posix path always starts with a forward slash and contains only forward slashes.
```
paul-marechal, line 91: > A posix path always starts with a forward slash and contains only forward slashes.

This is only true for absolute paths no?

packages/core/src/browser/resource-context-key.ts
colin-grant-work: I suppose then I wonder whether we shouldn't make retrieving information about the backend OS part of the pre-load startup routine so that we can then access it synchronously from inside the application? That would alleviate some of my anxiety about race conditions.

colin-grant-work (commented):
The changes work for me in various combinations of Linux and Windows servers and browsers. My only comment is about the asynch retrieval of the backend OS, which is probably fine but could potentially go wrong. If it's impractical to make the check synchronous while the application is running, then I'm fine with the current implementation.

colin-grant-work:
@msujew, I missed this review request - sorry; I'll take a look today.

packages/core/src/browser/resource-context-key.ts
```diff
@@ -38,7 +42,7 @@ export class ResourceContextKey {
     protected resourcePath: ContextKey<string>;
 
     @postConstruct()
-    protected init(): void {
+    protected async init(): Promise<void> {
```
colin-grant-work, line 45: With the refactor of the backend OS access, I don't see any reason for this to be `async` anymore?

packages/core/src/common/path.ts
```diff
@@ -230,6 +249,20 @@ export class Path {
         return this.raw;
     }
 
+    /**
+     * Converts the current path into a file system path.
+     * @param posix Indicates posix or windows file path.
+     * If `undefined`, the format will be determined by the `OS.backend.type` value.
+     * @returns A file system path.
+     */
+    fsPath(posix?: boolean): string {
```
colin-grant-work, line 258: What would you think of an enum here? A bit more typing, but more explicit than the boolean.

packages/monaco/src/browser/monaco-text-model-service.ts
```diff
@@ -47,11 +45,12 @@ export interface MonacoEditorModelFactory {
...
     declare readonly _serviceBrand: undefined;
 
-    protected readonly _ready = new Deferred<void>();
     /**
      * This component does some asynchronous work before being fully initialized.
+     *
+     * @deprecated since 1.25.0. Is instantly resolved.
      */
-    readonly ready: Promise<void> = this._ready.promise;
+    readonly ready: Promise<void> = Promise.resolve();
```
colin-grant-work, line 53: 👍 Good catch.

colin-grant-work (approved):
The functionality continues to work, and the move to a `preloader` makes this code cleaner, cleans up a few other areas, and opens up a lot of possibilities (e.g. preferences) so that's a definite 👍. One or two minor things that you can address or leave, as you choose.

packages/core/src/common/path.ts
msujew: I agree, I introduced an enum for this 👍

PR #10597, opened 2022-01-05
electron: Fix path comparison for exit confirmation

msujew:
#### What it does

Closes #10596 

Before testing for `includes`, we replace all backwards slashes with forward slashes.

#### How to test

1. Start the electron App on Windows.
2. Set the `application.confirmExit` preference to `always`.
3. Exit the application and assert that a confirmation dialog appears.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #10600, opened 2022-01-05
Encapsulate Electron Window Logic - Allow Exit during Load

### LINKED ISSUE #10586: Unable to close the window of theia electron version while loading
### Bug Description:
unable to close the window of theia electron version while loading

### Steps to Reproduce:

1.run the electron version of theia
2.click the close button during theia loading and the window cannot be closed

### Additional Information

- Operating System: macOS Big Sur 11.2
- Theia Version: 1.21.0

colin-grant-work:
#### What it does

Fixes #10586
This PR is a pretty big intervention into the logic of the ElectronMainApplication. The narrow goal is to track whether a window on the frontend is in a state to handle a close request so that windows that aren't ready to handle the close request can be closed immediately. The larger goal is to begin to extract the logic of handling Theia windows so that we can develop an abstraction that will work for both Electron and browser with an eye to doing things like running cleanup when the last window is closed. There are much smaller interventions that can address the narrow goal, so if there isn't an appetite for this size of refactor, that's fine, and I'm happy to reduce it to the minimum necessary to ensure that close gets handled correctly.

#### How to test

> It would be nice to write automated tests for some of this logic, but I haven't interacted with our Electron-specific test system much. Any pointers are appreciated.
1. Add some code that will prevent the frontend from ever achieving a `ready` state.
> e.g. add a `return` to `FrontendApplication#start` that terminates it early.
2. Run the application and try to close the window.
3. It should close.

---
1. Let the application get ready normally.
2. Dirty a file, or do something else that would generate a close prompt.
3. Observe that the prompt is displayed correctly and the application closes or doesn't close correctly
4. Change the `window.titleBarStyle` (if not on Mac) preference and observe that the application restarts correctly.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (commented):
@colin-grant-work I noticed something:

- if no workspace is present, and we perform the `open folder...` with multiple folders, the dialog appears (which occurs on unload), selecting `don't save` will close the dialog and restart the application without a workspace present. In vscode this same use-case will open the untitled workspace which was my expectation as a user.


https://user-images.githubusercontent.com/40359487/148396894-17beaed3-71bf-4576-98d7-f75e8b323547.mp4

msujew (approved):
The changes are looking really good to me! After applying the fix from #10597 I was able to successfully execute all testing steps:

- [x] Shutting off the application before it reaches its `ready` state is possible
- [x] Different configurations of the `application.confirmOnExit` preference, dirty editors, etc. work correctly when trying to exit the application
- [x] Commands such as `Configure Display Language` or `Reset Workbench Layout` still ask for confirmation for a reload
- [x] Switching between different `window.titleBarStyle` on different windows and reloading the windows continues to work correctly

colin-grant-work:
Thanks, @msujew. I'll merge this after the release.

PR #6803, opened 2019-12-31
Fix 'window.showTextDocument' to open resources with 'untitled' schema

ShimonBenYair:
Fixes eclipse-theia#6565

Signed-off-by: Shimon Ben Yair <shimon.ben.yair@sap.com>

#### What it does
Currently "window.showTextDocument" Does Not open Resources With "untitled" Scheme.
i.e. Using the following code : 
`const fileUri = Uri.parse(`untitled:${uri}`).with({ scheme });
    await window.showTextDocument(fileUri);`
does not open a file in Theia an raises an error.
With changes in this pull request, when calling the above API a file is created  and the user can modify and save the file with no problems.
This pull request Fixes eclipse-theia#6565 bug.

#### How to test

1. Add this  SQLTools extension (current version 0.21.5) under your plugins folder in Theia. 
[sqltools-0.21.5.zip](https://github.com/eclipse-theia/theia/files/4012222/sqltools-0.21.5.zip)

2. Create in workspace/theia a new file called "tom.db".
3.Add to User preferences the following values:
{
    "sqltools.useNodeRuntime": true,
    "sqltools.dependencyManager": {
        "autoAccept": true
    },
    "sqltools.connections": [
        {
            "name": "sqlite",
            "dialect": "SQLite",
            "database": "workspace/theia/tom.db"
        }
    ]
}`
`
image

3. Click on SQLTools extension.
4. In "sqlite" linem, click on connect:
image

When clicking on the "connect" button , code similar to the following is executed :
`const uri = vscode.Uri.file(os.homedir + '/newnew.txt');
const fileUri = vscode.Uri.parse(`untitled:${uri}`).with({ scheme: 'untitled' });
vscode.window.showTextDocument(fileUri);`

5. Verify a file called "sqlite Session.sql" was  created and  opened.
6. Modify this file (write something  and save) and verify there are no exceptions and errors.

showTextDocument_bug_fix

I have found some more issues in this code but as their fix is not required for this feature, it is out of the scope of this pull request. 

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

Issue #3354, opened 2018-10-31
Allow creating new files without a workspace


vince-fugnitto:
#### Expected Behaviour

As a user, I'd like the ability to **create new files** without a workspace or folder necessarily opened.
This will allow for a wider scale or use-cases for end-users who might want to edit files temporarily, use Theia as a scratchpad, or even do single-file editing such as a python script or markdown file.

#### Actual Behaviour
Theia does not currently support creating files without an explicit workspace opened which limits itself from more use cases by end-users.

#### Implementation Proposal
Allow files to be created regardless if a workspace is present or not.
- Create `untitled` files which can be saved in a temporary location by Theia until they are explicitly saved by end-users.
- Enable the menu option / command `New File` when no workspace is present (currently disabled)

PR #10608, opened 2022-01-10
Fix 'Save As' when saving a document with 'untitled' scheme

westbury:
#### What it does

This allows the user to save documents with 'untitled' scheme.  These documents are often opened by plugins that want to show content that is not in underlying storage (typically dynamically generated)

#### How to test

To test this you will need a plugin that creates documents with 'untitled' scheme.  The AWS toolkit extension does this (https://open-vsx.org/extension/amazonwebservices/aws-toolkit-vscode). Having configured an Access Key ID and a Secret Access Key and connected to your AWS account, in the AWS Explorer you can either:
- view a policy under IoT / Policies, right-click on a policy version and View...
- view a schema under Schemas / aws.events, right-click and View Schema.

Either of these should open an editor with a document with 'untitled' scheme.

Without this fix you will not be able so use the 'Save As...' menu to save untitled documents. It silently fails. (The FileSystem copy fails because 'untitled' is not a supported scheme).

Note on the fix: It might have been simpler to create an empty file in all cases. However I decided to leave it as is when the FileSystem supports the scheme because I assume there are reasons the file was being copied. This could be to preserve file permissions for example. The FileSystem implementation does support copying across from one provider to another.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/common-frontend-contribution.ts
+++ b/packages/core/src/browser/common-frontend-contribution.ts
@@ -50,7 +50,7 @@ import { EncodingRegistry } from './encoding-registry';
 import { UTF8 } from '../common/encodings';
 import { EnvVariablesServer } from '../common/env-variables';
 import { AuthenticationService } from './authentication-service';
-import { FormatType, Saveable } from './saveable';
+import { FormatType, Saveable, SaveOptions } from './saveable';
 import { QuickInputService, QuickPick, QuickPickItem } from './quick-input';
 import { AsyncLocalizationProvider } from '../common/i18n/localization';
 import { nls } from '../common/nls';
@@ -60,6 +60,7 @@ import { WindowService } from './window/window-service';
 import { FrontendApplicationConfigProvider } from './frontend-application-config-provider';
 import { DecorationStyle } from './decoration-style';
 import { isPinned, Title, togglePinned, Widget } from './widgets';
+import { SaveResourceService } from './save-resource-service';
 
 export namespace CommonMenus {
 
@@ -338,7 +339,8 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
         @inject(MessageService) protected readonly messageService: MessageService,
         @inject(OpenerService) protected readonly openerService: OpenerService,
         @inject(AboutDialog) protected readonly aboutDialog: AboutDialog,
-        @inject(AsyncLocalizationProvider) protected readonly localizationProvider: AsyncLocalizationProvider
+        @inject(AsyncLocalizationProvider) protected readonly localizationProvider: AsyncLocalizationProvider,
+        @inject(SaveResourceService) protected readonly saveResourceService: SaveResourceService,
     ) { }
 
     @inject(ContextKeyService)
@@ -889,10 +891,10 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
         });
 
         commandRegistry.registerCommand(CommonCommands.SAVE, {
-            execute: () => this.shell.save({ formatType: FormatType.ON })
+            execute: () => this.save({ formatType: FormatType.ON })
         });
         commandRegistry.registerCommand(CommonCommands.SAVE_WITHOUT_FORMATTING, {
-            execute: () => this.shell.save({ formatType: FormatType.OFF })
+            execute: () => this.save({ formatType: FormatType.OFF })
         });
         commandRegistry.registerCommand(CommonCommands.SAVE_ALL, {
             execute: () => this.shell.saveAll({ formatType: FormatType.DIRTY })
@@ -1056,6 +1058,11 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
                 when: 'activeEditorIsPinned'
             }
         );
+    }
+
+    protected async save(options?: SaveOptions): Promise<void> {
+        const widget = this.shell.currentWidget;
+        this.saveResourceService.save(widget, options);
     }
 
     protected async openAbout(): Promise<void> {
--- a/packages/core/src/browser/frontend-application-module.ts
+++ b/packages/core/src/browser/frontend-application-module.ts
@@ -120,6 +120,7 @@ import {
 import { RendererHost } from './widgets';
 import { TooltipService, TooltipServiceImpl } from './tooltip-service';
 import { bindFrontendStopwatch, bindBackendStopwatch } from './performance';
+import { SaveResourceService } from './save-resource-service';
 
 export { bindResourceProvider, bindMessageService, bindPreferenceService };
 
@@ -395,4 +396,6 @@ export const frontendApplicationModule = new ContainerModule((bind, unbind, isBo
 
     bindFrontendStopwatch(bind);
     bindBackendStopwatch(bind);
+
+    bind(SaveResourceService).toSelf().inSingletonScope();
 });
--- a/packages/core/src/browser/save-resource-service.ts
+++ b/packages/core/src/browser/save-resource-service.ts
@@ -1,0 +1,45 @@
+/********************************************************************************
+ * Copyright (C) 2022 Arm and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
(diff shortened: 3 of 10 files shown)


msujew (commented):
For anyone interested in testing this, but unwilling to go through the AWS setup, I created a [small sample extension](https://github.com/eclipse-theia/theia/files/7838915/dynamic-editor-0.0.1.zip) for this. Run the `Show test editor` command to create a simple untitled file.

For the change itself: While this works for the `Save As` command, simply saving the document using `Ctrl+S` will throw an error: `NoPermissions (FileSystemError): Error: EPERM: operation not permitted, open 'C:\Untitled-0'`.

For feature completeness and to better align with vscode, I would recommend that we automatically delegate to the `Save As` command when saving a file with the `untitled` scheme. What do you think @westbury?

westbury:
> For feature completeness and to better align with vscode, I would recommend that we automatically delegate to the `Save As` command when saving a file with the `untitled` scheme.

Sounds good. It is certainly common for applications to open up the 'Save As' dialog when saving an untitled file. I just need to think about a good way to do that because 'Save' is in core package and 'Save As' is in workspace package. 'Save As' can and does call 'Save', see https://github.com/eclipse-theia/theia/blob/52a652922c431bb26b171636004536921e46e908/packages/workspace/src/browser/workspace-frontend-contribution.ts#L497 but the other way around is more problematic. I'll find a clean way of doing it.

westbury:
> For feature completeness and to better align with vscode, I would recommend that we automatically delegate to the `Save As` command when saving a file with the `untitled` scheme.

This is far harder than it sounds. See the comments in https://github.com/eclipse-theia/theia/pull/6803 in which Anton requested this in another PR and accepted it was not straightforward. The problem is that 'save' support is in the core package and baked into the core and monaco packages. Implementations can be placed in the Resource or ResourceResolver but that is really too late to switch from 'save' to 'save as'.

Also note that 'save' should work for 'untitled', see https://github.com/eclipse-theia/theia/blob/52a652922c431bb26b171636004536921e46e908/packages/plugin-ext/src/main/browser/editor/untitled-resource.ts#L94-L102

However, even if it is fixed, the file is saved silently in a location where users are not likely to find it. So I would be inclined to remove the optional `saveContents` method, leaving untitled documents as read-only. This would be consistent, and would mean users are not prompted to save their untitled documents when they 'save all'. If we want the alternative, to align with vscode, then we need to be prepared to accept more significant architectural changes.

I think this needs to be handled in a separate PR.

msujew:
@westbury I see. My idea would be to perhaps create a `SaveAsService`, which is defined at `core` level, but only bound in the `workspace` package. If the workspace package isn't used (which is unlikely) we keep the current behavior. What do you think about that? I don't see to many architectural changes in that, since it would just move the `saveAs` implementation in its own service.

westbury:
@msujew I've pushed a second commit which adds support for the 'Save' action on untitled resources, invoking the 'Save As' code to prompt the user for a file name. In addition to the 'save' service, there are also a few changes so we can determine early if a widget contains an untitled resource.

This replaces most of the changes added by https://github.com/eclipse-theia/theia/pull/6803 . The way saving of untitled documents was handled by the Resource itself was very limiting. It was not possible to have user interactions such as prompts for file names. Furthermore there is good code in the 'Save As' that seamlessly replaces the 'untitled' editor with the 'file' editor, which also could not work when saving is done by the Resource. I therefore removed all the saving code from untitled-resource.ts. I did leave a stub implementation of `saveContents`. This was necessary to ensure the untitled documents remain editable.

It would be logical to make untitled documents read-only. The user would then have to save to a file before starting to make edits. However it may not be intuitive to the user that they need to do this in order to start editing the contents. Also it would not be aligned with vscode.

'Save All' does not save untitled documents. This is as it currently works and it may be confusing if the user is prompted for lots of file names when they 'Save All' with lots of untitled documents.

msujew:
@westbury I agree with Colin that untitled resources should always be marked as dirty. I think this would probably alleviate most of the issues encountered here.

colin-grant-work:
@westbury, are you interested in continuing work on this feature? I was thinking of moving some of the machinery for creating untitled resources out of the plugin package so that it could be used by the 'New File' command to create an empty editor rather than requiring the user to immediately create a new file, and this PR significantly improves our handling untitled resources; if you are busy with other work, I'd be happy to pick it up address the remaining comments.

westbury:
@colin-grant-work The problem is that marking all untitled documents as 'unsaved' would not be the correct thing to do in our case. In VSCode, when one creates a new file using File, New, the file is not marked as dirty and one can close it without getting a prompt. In the case of the AWS plugin we definitely would not want the untitled document to be marked as dirty. It should be marked as dirty only if the user changes the text.

The 'untitled' is used in Theia not for a new file but when text data is to shown and that text data comes from a source that is not in a file system that can provide a file name. If the user doesn't change the data then there is no unsaved data.

Sorry, I meant to respond earlier but I was not sure how to resolve these different expectations.

colin-grant-work:
VSCode has this interesting comment about opening untitled editors:

https://github.com/microsoft/vscode/blob/ab49f4f310a014c9e03839f37c1ca3c48196aeca/src/vs/workbench/common/editor.ts#L417-L433

So it seems that the decision about whether to mark the editor as dirty has in part to do with way it's originally opened and the point at which a URI is created. At the moment, we don't have distinct handling for different untitled cases (untitled because new, untitled because just a temporary reference to some text data, etc.), but we would need to implement some to match VSCode's behavior in this area.

colin-grant-work:
@westbury, @msujew, what do you think of proceeding with the PR as it currently is, and then iterating on it as part of a separate PR aimed at addressing https://github.com/eclipse-theia/theia/issues/3354?

msujew:
@colin-grant-work I see, yeah, I'd be fine with that 👍

colin-grant-work (approved):
This is a significant improvement over the current code and does not produce any noticeable regressions. The behavior can be further refined in later PR's.

colin-grant-work:
@westbury, would you mind rebasing the PR so that it can be merged? I'm not sure exactly what's going on with the 3PP check (@vince-fugnitto) - I don't see any changes in dependencies in the PR - but perhaps that will resolve itself with another push.

westbury:
@colin-grant-work Thanks for approving this. I've rebased and retested with the AWS plugin and it works as expected. I have not made all untitled as dirty, as discussed.

Also https://github.com/eclipse-theia/theia/pull/10608#discussion_r786063316 is still outstanding because it still calls Saveable.save and saves the current resource so is not really the same.

colin-grant-work:
> In the case of the AWS plugin we definitely would not want the untitled document to be marked as dirty.

@westbury, have you confirmed this behavior? It appears that if you use the `View Schema` command, you end up in this code:

https://github.com/aws/aws-toolkit-vscode/blob/3ea63e0952b322b6f24e811a341c02ef99cac0c3/src/eventSchemas/commands/viewSchemaItem.ts#L44-L54

When I mimic that behavior in a VSCode plugin command:
```typescript
	context.subscriptions.push(vscode.commands.registerCommand('untitled-editors.language.lateContent', async () => {
		const newDoc = await vscode.workspace.openTextDocument({
			language: 'json',
		})
		const editor = await vscode.window.showTextDocument(newDoc, vscode.ViewColumn.One, false)
		await editor.edit(edit => edit.insert(new vscode.Position(/*line*/ 0, /*character*/ 0), '{"content": "for example"}'))
	}));
```

The resulting editor is marked as dirty in VSCode. So far, it appears that any untitled editor that is not completely empty is marked as dirty, which seems correct to me: if there's any text that doesn't match the backing store (none, in this case), then the editor is dirty.

westbury:
@colin-grant-work Here is a recording of what I see. I view a couple of policies but I don't edit the text. When I 'close all' they are both closed without prompt, as I would expect. I then make a change. Now you see the dirty indicator and when I close the editor, I am prompted to save, as I would expect. Also if I select 'Save' then I am prompted for a file as though I had selected 'Save As'.

https://user-images.githubusercontent.com/87310/157766784-ab611eba-9708-4e57-bb29-b6f26916bca2.mp4

In your test code it appears you are creating the documents as empty documents. This is not the usual case, as when you view a textual representation of some data it would not normally be empty.

colin-grant-work:
@westbury, thanks for the recording. It looks like you're demonstrating what happens in Theia, where we know that they're not marked as dirty, but in order to match the expectations of VSCode plugins, we would want to mark the editors dirty as VSCode does - do you know if the editors created by the AWS plugin show up as dirty in VSCode?

westbury:
@colin-grant-work Sorry I misunderstood your comment. You are correct that untitled documents show immediately as dirty in VSCode. Which means if users view lots of AWS policies, schemas etc. then they have to click 'don't save' for each one.

PR #10641, opened 2022-01-17
Group views in view menu by location

westbury:
#### What it does
Updates the 'Open View...' menu so the views are now grouped like they are in vscode.

So without the changes in this PR, the open view menu looks something like:

Screenshot from 2022-01-17 12-33-55

and with the changes it now looks like:

Screenshot from 2022-01-17 12-28-40

Note how the AWS plugin contributes a view also called 'Explorer'. The two 'Explorer' views are indistinguishable in Theia but appear in different groups in vscode.
 
#### How to test
Try out the view menu with various packages and plugins. Check that the groupings are correct and there is nothing that will confuse users.
- drag views to various places in the shell
- drag view container parts to other view containers
- check that the menu shows dragged views correctly after a refresh
- 
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/quick-input/quick-view-service.ts
+++ b/packages/core/src/browser/quick-input/quick-view-service.ts
@@ -24,6 +24,7 @@ import { ApplicationShell, WidgetArea } from '../shell';
 export interface QuickViewItem {
     readonly label: string;
     readonly viewId: string;
+    readonly widgetId: string;
     readonly location: string;
     readonly when?: string;
     readonly open: () => void;
@@ -34,14 +35,15 @@ export class QuickViewService implements QuickAccessContribution, QuickAccessPro
     static PREFIX = 'view ';
 
     protected readonly items: (QuickPickItem & {
-        viewId: string,
+        widgetId: string,
         location: string,
         defaultLocation: string,
         viewContainerTitle?: string,
         when?: string
     })[] = [];
-    protected readonly containers = new Map<string, string>();
+    protected readonly containers = new Map<string, { location: string, title: string }>();
     private hiddenItemLabels = new Set<string | undefined>();
+    protected locationsToRestore = new Map<string, string>();
 
     @inject(ApplicationShell)
     protected readonly shell: ApplicationShell;
@@ -55,33 +57,75 @@ export class QuickViewService implements QuickAccessContribution, QuickAccessPro
     @postConstruct()
     init(): void {
         this.shell.onDidAddWidget(widget => {
-            const item = this.items.find(i => i.viewId === widget.id);
+            const viewId = widget.id.startsWith('plugin-view-container:')
+                ? `plugin-view:${widget.id.substring('plugin-view-container:'.length)}`
+                : widget.id;
+            const item = this.items.find(i => i.widgetId === viewId);
             if (item) {
                 item.location = widget[WidgetArea];
+            }
+
+            if (this.containers.has(widget.id)) {
+                const containerData = this.containers.get(widget.id)!;
+                containerData.location = widget[WidgetArea];
+            } else {
+                this.containers.set(widget.id, { location: widget[WidgetArea], title: '' });
             }
         });
 
         this.shell.onDidRemoveWidget(widget => {
-            const item = this.items.find(i => i.viewId === widget.id);
+            const item = this.items.find(i => i.widgetId === widget.id);
             if (item) {
                 item.location = item.defaultLocation;
+            } else {
+                if (this.containers.has(widget.id)) {
+                    const containerData = this.containers.get(widget.id)!;
+                    containerData.location = '';
+                }
+            }
+        });
+
+        this.shell.onDidMoveContainerPart(info => {
+            const item = this.items.find(i => i.widgetId === info.widget.id);
+            if (item) {
+                item.location = info.destinationContainerId;
+            } else {
+                // Views from plugins will be restored from state before being registered, so create
+                // an entry with the location, to be filled in when the widget is registered.
+                this.locationsToRestore.set(info.widget.id, info.destinationContainerId);
             }
         });
     }
 
     registerContainer(id: string, location: string, title: string): Disposable {
-        const label = this.getLocationLabel(location) + ' / ' + title;
-        this.containers.set(id, label);
+        const containerData = this.containers.get(id);
+        if (containerData) {
+            // Exists, so use actual location, not location from contributor
+            containerData.title = title;
+        } else {
+            this.containers.set(id, { location, title });
+        }
         return Disposable.create(() => {
             this.containers.delete(id);
         });
     }
 
     registerItem(item: QuickViewItem): Disposable {
+        const containerWidgetId = item.widgetId.startsWith('plugin-view:')
+            ? `plugin-view-container:${item.widgetId.substring('plugin-view:'.length)}`
+            : item.widgetId;
+
+        const containerData = this.containers.get(containerWidgetId);
+        const locationToRestore = this.locationsToRestore.get(item.widgetId);
+        const location = containerData
+            ? containerData.location
+            : (locationToRestore ?? item.location);
+
         const quickOpenItem = {
             label: item.label,
             viewId: item.viewId,
-            location: item.location,
+            widgetId: item.widgetId,
+            location,
             defaultLocation: item.location,
             execute: () => item.open(),
             when: item.when
@@ -148,7 +192,10 @@ export class QuickViewService implements QuickAccessContribution, QuickAccessPro
             case 'right': return 'Right';
             case 'main': return 'Editor';
             default:
-                return this.containers.get(location) ?? location;
+                const containerData = this.containers.get(location);
+                return containerData
+                    ? this.getLocationLabel(containerData.location) + ' / ' + containerData.title
+                    : location;
         }
     }
 }
--- a/packages/core/src/browser/quick-input/quick-view-service.ts
+++ b/packages/core/src/browser/quick-input/quick-view-service.ts
@@ -14,14 +14,17 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 // *****************************************************************************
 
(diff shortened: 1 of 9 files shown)


packages/core/src/browser/shell/view-contribution.ts
```diff
@@ -142,6 +142,7 @@ export abstract class AbstractViewContribution<T extends Widget> implements Comm
         }
         this.quickView?.registerItem({
             label: this.viewLabel,
+            location: this.options.viewContainerId ?? this.options.defaultWidgetOptions.area ?? "",
```
vince-fugnitto, line 145: The following is causing the build to fail (singlequotes):

```suggestion
            location: this.options.viewContainerId ?? this.options.defaultWidgetOptions.area ?? '',
```

vince-fugnitto (commented):
I've noticed some odd behaviors:

1. the location is not always correctly computed, for example with `npm scripts` we have a value of (lowercase) 'explorer' while it should be `Side Panel / Explorer` similarly to vscode:

https://user-images.githubusercontent.com/40359487/149781558-5e795460-352e-4a37-b92d-a9e306136eaf.mp4


2. the _open view_ does not respect the location of a view if it is moved (ex: moving `outline` from right to left. The views in vscode cannot be moved but we should think of handling it since we are more flexible, and giving an incorrect location is confusing for end-users:

https://user-images.githubusercontent.com/40359487/149781866-06fe9e01-3b30-41a3-a84c-3185467dbf28.mp4

westbury:
@vince-fugnitto I have updated the PR so the menu now shows the actual location of views. If the view is closed then the location where it would open is shown.

The changes are more extensive than I had hoped because I changed `onDidAddWidget` and `onDidRemoveWidget` to send both the widget and the area. I tried using `shell.getAreaFor(widget)` but it was too fragile being dependent on stuff happening in a precise order. By having the emitter send the area, we can be sure we are always getting the correct area.

To test, check the following:
- when a view is moved, the new location is used in the menu
- when the browser is refreshed, the menu continues to show the actual location
- when a view is closed, the menu reverts to showing the default location
- 
I think that covers all review comments.

msujew (commented):
While I can confirm that your recent changes address the comment brought up by Vince, I believe we can accomplish this with less breaking changes:

Instead of introducing a new type for the `onDidAddWidget` and `onDidRemoveWidget` event, I propose a new interface like `AreaWidget` which looks something this:

```ts
export const WidgetArea = Symbol('WidgetArea');

export interface AreaWidget {
  [WidgetArea]: ApplicationShell.Area;
}
```

```ts
protected fireDidAddWidget(widget: Widget, area: ApplicationShell.Area): void {
  const areaWidget = widget as AreaWidget;
  areaWidget[WidgetArea] = area;
  this.onDidAddWidgetEmitter.fire(areaWidget);
}
```

(The symbol is used to prevent conflicts with possible existing extenders of `Widget`)

The `AreaWidget` is then used instead, which completely prevents a breaking change, since it is only an extension of the existing `Widget` type.

westbury:
> I propose a new interface like `AreaWidget`

I have pushed some changes as per your suggestion. I have kept them as a separate commit for now because I am not entirely happy with the changes. I did make one change, though. The event is now the intersection of `Widget` and the 'area' property, not just `Widget`. This is important because otherwise we are breaking the types. Consider an extender that extends `ApplicationShell` and fires onDidAddWidgetEmitter directly. If a widget was passed without an area then the code would be broken. It will compile ok but the view menu will fail due to the missing area. By typing the emitter correctly, typescript will point out the missing area. This also makes it known to listeners that the area is available.

As I said, I am not entirely convinced that these changes are better. I understand trying to avoid breaking changes, but so is keeping the code clean and readable.

When performing Theia upgrades, breakages in the API are the simplest to fix. I can always fix API changes in minutes. Compared to the other issues that come up during a Theia upgrade, some of which can takes days to resolve, API breakages are just simply not an issue. Perhaps I am biased because I am the one that has been doing our Theia upgrades, but I really appreciate seeing Typescript errors showing me where our extensions need to be changed; that is far better than dealing with problems that don't show up until testing or later.

westbury:
Actually I'm fine with the changes so I have gone ahead and squashed the commit.

vince-fugnitto:
@westbury are there plans to support the tracking of tree-view parts which can be moved to different containers (ex: `explorer` to `scm`) - likely not critical initially:

image

In this case the `open view `command still shows `explorer` as the location while in vscode it will show up correctly under `scm`.

westbury:
> are there plans to support the tracking of tree-view parts which can be moved to different containers

Yes, I can add support to track those too. If a view is already in another container then users are more likely to use the menu to bring the containing view into view, but this would make it consistent across all views.

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -659,6 +677,14 @@ export class ApplicationShell extends Widget {
         }
     }
 
+    protected registerViewPositions(area: ApplicationShell.Area, widgets: (Widget | undefined)[]): void {
+        for (const widget of widgets) {
+            if (widget) {
+                this.fireDidAddWidget(widget, area);
+            }
+        }
+    }
```
colin-grant-work, line 686: I think this step is only necessary for the `main` and `bottom` areas. The `DockPanel.restoreLayout` method doesn't produce any `widgetAdded` events, but our implementation of `setLayoutData` for the side panels does already produce events.

packages/core/src/browser/quick-input/quick-view-service.ts
```diff
@@ -30,18 +34,99 @@ export interface QuickViewItem {
...
 
     @inject(QuickAccessRegistry)
     protected readonly quickAccessRegistry: QuickAccessRegistry;
 
     @inject(ContextKeyService)
-    protected readonly contextKexService: ContextKeyService;
+    protected readonly contextKeyService: ContextKeyService;
+
+    @postConstruct()
+    init(): void {
```
colin-grant-work, line 58: A lot of state management has been added here, but I'm not sure it's necessary. When this quick pick is opened, we're dealing with two sets of things: widgets that are already in the shell and widgets that are not already in the shell, but have been registered as openable.

For widgets that are already in the shell, we can just compute everything we need to know about them using utilities on the `ApplicationShell` class.

For widgets that aren't already in the shell, we know where they _would_ go if they were opened, and we can just use that information. We don't need any live state tracking there, either.

My mental model may be a bit naïve, though, so please correct me if I'm missing something.

colin-grant-work (changes requested):
This is promising, but there some issues that need to be addressed before it is ready to be merged.

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -188,16 +194,20 @@ export class ApplicationShell extends Widget {
...
 
-    protected readonly onDidRemoveWidgetEmitter = new Emitter<Widget>();
+    protected readonly onDidRemoveWidgetEmitter = new Emitter<AreaWidget>();
     readonly onDidRemoveWidget = this.onDidRemoveWidgetEmitter.event;
-    protected fireDidRemoveWidget(widget: Widget): void {
-        this.onDidRemoveWidgetEmitter.fire(widget);
+    protected fireDidRemoveWidget(widget: Widget, area: ApplicationShell.Area): void {
+        const areaWidget = widget as AreaWidget;
+        areaWidget[WidgetArea] = area;
+        this.onDidRemoveWidgetEmitter.fire(areaWidget);
```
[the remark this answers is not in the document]
westbury: It was changed from a tuple to a Widget with an extra property to increase the chances that extenders' code will build without modification. However you are right that using a tuple is cleaner so I have changed to use a tuple (and added an entry to the changelog).

packages/core/src/browser/quick-input/quick-view-service.ts
westbury: Yes, that would certainly be a nicer solution. However we can't currently get the information we need from `ApplicationShell`. (See also https://github.com/eclipse-theia/theia/pull/10641#issuecomment-1025644454). It would be possible to refactor `ApplicationShell` so it contained more of a model, so `getAreaFor` would get from this model instead of looking through the widget layout structures. I was trying to keep the changes to `ApplicationShell` to a minimum but I am happy to make these changes if you feel that would be better.

packages/core/src/browser/quick-input/quick-view-service.ts
```diff
@@ -75,9 +160,42 @@ export class QuickViewService implements QuickAccessContribution, QuickAccessPro
...
     getPicks(filter: string, token: CancellationToken): QuickPicks {
         const items = this.items.filter(item =>
-            (item.when === undefined || this.contextKexService.match(item.when)) &&
+            (item.when === undefined || this.contextKeyService.match(item.when)) &&
             (!this.hiddenItemLabels.has(item.label))
         );
-        return filterItems(items, filter);
+
+        // Reversing is a trick to ensure any others (indexOf = -1) appear at the end of the sorted array.
+        const locationOrder = [ 'left', 'bottom', 'right', 'main' ].reverse();
```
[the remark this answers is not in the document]
westbury: I thought it was more readable to list the areas in the expected order. If performance or reducing code size is more important then I can change.

colin-grant-work (commented):
There are some odd behaviors. For example, if you close the `SCM` view container, then open the `view` quick open, the label for SCM widgets is missing the `Side Bar` prefix:

image

It doesn't appear to pick up parts in the explorer correctly. Note that Open Editors and NPM Scripts are both in the view, but not the quick pick:

image

Alternatively, it may not be picking NPM scripts at all, since it also doesn't track it when it is moved to the search container:

image

Not a regression, but undesirable behavior:

 - Move another view container part into the search view container
 - Move the search widget into the SCM view container
 - Close SCM
 - `view` -> select search (on `master` just as `Search`, with this code under the `/ Source Control` section.
 - Nothing happens (probably because the search view container exists?)

vince-fugnitto:
Closing the pull-request to keep the list of opened pull-requests more manageable, and the pull-request still has unaddressed feedback for over 1 year. We can always re-open the pull-request once we update the branch again.

PR #10643, opened 2022-01-17
Reimplement `window.menuBarVisibility` preference

### LINKED ISSUE #10571: 'CUT' not work in context menu
### Steps to Reproduce:
1.  select something
2.  click Edit—>Cut
3.  `cut` not work

### Additional Information

- Operating System: linux
- Theia Version: 1.19.0

colin-grant-work:
#### What it does

> I'm leaving this in 'Draft' status while I do some more [general](#General) testing...

Fixes #10571 by ensuring that the same logic is used for the main menu when `window.menuBarVisibility` is `compact` as when it is not.

The heart of this PR is a move to use the same menu widget for the (browser-style) main menu regardless of the value of the `window.menuBarVisibility` preference, rather than using a context menu in `compact` mode. That ensures that we get the same behavior as we do with the normal menu bar, and that any fixes apply equally to both. In particular, it ensures that (1) commands invoked from the main menu do _not_ receive an `anchor` argument and (2) the logic for preserving the focus context to use when a command is run is consistent between the menu bar and compact menu.

While working on that, I noticed several other features of the menu implementation that could be improved:
 - the `window.menuBarVisibility` preference handling was scattered across a number of different classes. This PR moves (almost) all references to the preference into the `MainMenuContribution` classes.
 - event handling for menus was also split between the `MainMenuFactory` and `MainMenuContribution` classes. This PR moves all event handling into the `MainMenuContribution` classes, with the factories only responsible for creating the widget.

##### #10571
1. Set the `window.menuBarVisibility` preference to `'compact'` (non-OSX only)
2. Open an editor
3. Select some text
4. Open the main menu > Edit > Cut
5. The text you selected should be removed and available in your clipboard.

##### <a name="General">General</a>
1. In browser, test various transitions between and startup in different `window.menuBarVisibility` states.
2. In Electron, test various combinations of `window.titleBarStyle` and `window.menuBarVisibility` states and transitions.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/api-tests/src/menus.spec.js
+++ b/examples/api-tests/src/menus.spec.js
@@ -45,7 +45,7 @@ describe('Menus', function () {
     const container = window.theia.container;
     const shell = container.get(ApplicationShell);
     const menuBarContribution = container.get(BrowserMenuBarContribution);
-    const menuBar = /** @type {import('@theia/core/lib/browser/menu/browser-menu-plugin').MenuBarWidget} */ (menuBarContribution.menuBar);
+    const menuBar = /** @type {import('@theia/core/lib/browser/menu/browser-menu-plugin').MenuBarWidget} */ (menuBarContribution.browserMenu);
     const pluginService = container.get(HostedPluginSupport);
     const menus = container.get(MenuModelRegistry);
     const commands = container.get(CommandRegistry);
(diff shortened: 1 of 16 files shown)


colin-grant-work:
Closing for now - seems better just to implement a focus restoration for the compact menu widget.

Issue #10033, opened 2021-09-01
Add `ExtensionContext.logUri`


tsmaeder:
### Feature Description:
We already have `logPath`, but we have to do the same as for the other fields like `storagePath|Uri`. Should be trivial. See https://code.visualstudio.com/api/references/vscode-api#ExtensionContext

PR #10650, opened 2022-01-19
Support 'logUri' and 'extension' (partially) to PluginContext

martin-fleck-at:
- Ensure 'logUri' and 'extension' property are present in context
- 'extension' currently has always theia.Plugin type, even in vscode

Fixes https://github.com/eclipse-theia/theia/issues/10033

#### What it does
- Add support for 'logUri' and 'extension' property to the PluginContext

#### How to test
- I pushed a branch https://github.com/martin-fleck-at/theia/tree/issues/10033-test which adds a [plugin](https://github.com/martin-fleck-at/theia/blob/issues/10033-test/packages/test-plugin/src/extension.ts) to Theia that evaluates the Plugin context to produce error dialogs for missing properties such as:
image

#### Review checklist

#### Reminder for reviewers

---

** **To Note:** ** This PR adds the 'extension' property but it is currently ALWAYS of type TheiaPlugin even if the plugin is a VS Code extension. There is already a conversion available in the VS Code package for the 'vscode.extensions.getExtension' API. However, I did not find the proper way to use that functionality in the plugin manager as it probably requires some RPC calls. So if someone could point me in the right direction, I could also have a look at this. Until then, the custom VS Code extension properties (extensionPath, extensionUri and extensionKind) are not available through the context. However, they are available if you get the extension through vscode.extensions.getExtension.

image


colin-grant-work (commented):
Could you clarify your question about the plugin type a bit? Am I correct that you'd like to use the `asVsCodeExtension` converter in the `plugin-manager.ts` file when you build the context for the plugin, if the plugin is in fact a VSCode plugin?

My inclination would be to just add the fields that `asVsCodeExtension` adds anyway, since they're unlikely to do any harm to a plugin that isn't expecting them, but perhaps others have better ideas.

packages/plugin-ext-vscode/src/node/plugin-vscode-extension.ts
```diff
@@ -0,0 +1,42 @@
+/********************************************************************************
+ * Copyright (C) 2022 Red Hat, Inc.
```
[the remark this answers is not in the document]
martin-fleck-at: I am not but since I only moved classes to a new file I thought I should keep the original contributor company. However, with the latest changes this no longer applies since I moved the vscode extension code to the plugin anyway. If that should change, I will keep that in mind next time, sorry for that!

martin-fleck-at:
@colin-grant-work Thank you for your fast reply! Yes, exactly, the plugin manager also needs to be aware of the additional properties that were added to the Theia plugin so that it can also be used as a VSCode extension. Previously this was only done for the vscode API calls (vscode.extensions.all, vscode.extensions.getExtension) but we need the same conversion for the 'extension' property used in the PluginContext (or ExtensionContext in VSCode). As you suggested, I now simply moved the additional properties directly into the class so that it automatically covers all aspects of a Theia plugin and a VSCode extension. This also made the conversion within the vscode.extension.* calls unnecessary. With that everything seems to work for me:
image

packages/plugin-ext/src/plugin/plugin-context.ts
```diff
@@ -972,19 +972,33 @@ export function createAPIFactory(
     };
 }
 
-class Plugin<T> implements theia.Plugin<T> {
+enum ExtensionKind {
+    UI = 1,
+    Workspace = 2
+}
+
+export class PluginInstance<T> implements theia.Plugin<T> {
```
colin-grant-work, line 980: Is the renaming of this class necessary? Changing the name of an export is a breaking change, so if it isn't strictly necessary, it's better to avoid it.

packages/plugin-ext/src/plugin/plugin-context.ts
```diff
@@ -972,19 +972,33 @@ export function createAPIFactory(
...
         this.packageJSON = plugin.rawModel;
         this.pluginType = plugin.model.entryPoint.frontend ? 'frontend' : 'backend';
+
+        // Support VSCode Extension properties even if not explicitly exposed through interface
+        // as they are expected in the vscode.extension.* API calls and the ExtensionContext of the activate method
+        Object.assign(this, {
+            extensionPath: plugin.pluginPath,
+            extensionUri: plugin.pluginUri,
+            extensionKind: ExtensionKind.UI // stub as a local VS Code extension (not running on a remote workspace)
+        });
```
colin-grant-work, line 1001: It seems a bit odd to call `Object.assign` in the constructor. In this case, it's currently necessary because the fields added here don't appear in the class or interface declaration. There are a couple of options for addressing this:
 - Add these fields to the `theia.Plugin` interface.
 - Declare another interface that extends the `theia.Plugin` interface and have this class implement that interface.
 - Declare another interface and another class, but use that class by default rather than the previous `Plugin` class.

The last option is the most elaborate but causes the least breakage to downstream adopters who may have their own extensions of the `theia.Plugin` interface, but it may just defer problems if we eventually want to change our call signatures to require the new fields.

packages/plugin-ext/src/plugin/plugin-context.ts
martin-fleck-at: Good point! I changed it back to the original class name though the original class was not exported so I guess it never was officially API. I now export the "base" Plugin class as well as the extended PluginExt class so that adopters could even base their own implementation on top of that if they need to.

packages/plugin-ext/src/plugin/plugin-context.ts
martin-fleck-at: You are right, I was just worried about the external visibility as I did not actually want to make it API. But I agree with you, it is probably inevitable and if we already have it we may as well define it in a well-structured manner. I followed your advice and implemented the third option to have a "base" Plugin class as before but an extension interface + implementation for the VS code-specific properties.

colin-grant-work (approved):
This looks good to me and better satisfies the expectations of VSCode plugins.

Issue #10653, opened 2022-01-19
debug: console is cleared after a debug session


vince-fugnitto:
#### Bug Description:

There is a regression with the **debug console** as it is cleared after a debug session is over.
The output should instead be preserved.

https://user-images.githubusercontent.com/40359487/150209279-c988d7da-21b2-42f7-9c55-a209d0192729.mp4

### Steps to Reproduce:

1. start the application using `theia` as a workspace
2. set a breakpoint in a `*.spec.ts` file (the bug is also reproducible without any breakpoints)
3. start the `run mocha tests` debug session
4. complete the debug session
5. notice that the debug console output is cleared

vince-fugnitto:
Confirmed that the regression was introduced in https://github.com/eclipse-theia/theia/pull/10333.

PR #10671, opened 2022-01-26
console: do not clear output when a session ends

vince-fugnitto:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/10653.

The pull-request fixes an issue which caused the debug console output to clear when a session was completed. 
The change now preserves the output so end-users can view the output of their debug session.

It namely reverts a change proposed in https://github.com/eclipse-theia/theia/pull/10333.

#### How to test

1. start a debug session (with and without debug breakpoints).
2. when the session is terminated, the console output should be preserved.
3. confirm that subsequent sessions also work correctly.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/console/src/browser/console-widget.ts
+++ b/packages/console/src/browser/console-widget.ts
@@ -115,7 +115,10 @@ export class ConsoleWidget extends BaseWidget implements StatefulWidget {
 
         this.session = this.sessionManager.selectedSession;
         this.toDispose.push(this.sessionManager.onDidChangeSelectedSession(session => {
-            this.session = session;
+            // Do not clear the session output when `undefined`.
+            if (session) {
+                this.session = session;
+            }
         }));
 
         this.updateFont();


vince-fugnitto:
@msujew do you mind reviewing whenever you get a chance? I reverted the logic from https://github.com/eclipse-theia/theia/pull/10333 which broke the output, and am unsure if no longer setting the cached session to `undefined` will cause any side-effects.

colin-grant-work:
I believe that the motivation behind removing the reference to the debug session was to ensure that it would be garbage-collected when the session was terminated. With this change, that doesn't happen:

image

The `ConsoleWidget` isn't the only retainer of old sessions: of the four in that screenshot, two are retained by something in `debug-hover-source.ts`, one by something in `debug-stack-frame.tsx`, and it appears that the reference retained by the console is released when the session is replaced by another console session, but it seems that there ought to be some way to release the debug session without deleting the content.

colin-grant-work:
But if push comes to shove, I think retaining a reference to the session is better than deleting the output, so if there's no way to preserve the output without retaining the reference for now, that's OK.

colin-grant-work (approved):
Better than the alternative for now, though it may be worth making an issue to allow for retention of output when session is set to undefined.

msujew (approved):
Looks good to me as well. I agree with Colin, we can fix the dangling debug session reference sometime else.

vince-fugnitto:
I'll merge for the moment since it improves the behavior for end-users, and open a follow-up issue to track the cleanup.
Thanks guys!

Issue #6589, opened 2019-11-20
Complete terminal(pseudoTerminal) api


AndrienkoAleksandr:
### Description
VSCode has new pseudo-terminal api https://github.com/microsoft/vscode/blob/master/src/vs/vscode.d.ts#L7518 

We need implement it.

**OS and Theia version:**

**Diagnostics:**

akosyakov:
Are there known VS Code samples, extensions to test again?

AndrienkoAleksandr:
> Are there known VS Code samples, extensions to test again?

I will take a look.

PEZ:
I am about to switch to this for Calva (a Clojure extension for VS Code). Here's the PR: https://github.com/BetterThanTomorrow/calva/pull/654

It's a bit ironic, because the reason I started with it was to see if it would help Calva start working in Theia. Something that is partly do to Theia loosing track of the terminal used for hosting the orphaned Task I am currently using.

I elaborated some more on how I use the API here: https://community.theia-ide.org/t/can-i-use-vscode-extensionterminal-in-theia/24

Also throwing in some more of my struggles here: https://community.gitpod.io/t/is-there-an-easy-way-to-debug-a-vscode-extension-inside-gitpod/596/68?u=pez (I wanted to keep reporting in that thread, but am to new to the community to be allowed to 😄 ).

akosyakov:
@PEZ I don't think anyone is looking into it, right now. PRs are welcomed.

datou0412:
this sample can be used to test [extension-terminal-sample](https://github.com/microsoft/vscode-extension-samples/tree/master/extension-terminal-sample)

tsmaeder:
Closing in favor of https://github.com/eclipse-theia/theia/issues/12074

PR #10683, opened 2022-01-28
Align window.createTerminal API with VSCode

sgraband:
#### What it does

Deprecated PseudoTerminalOptions and created ExternalTerminalOptions to align with VSCode API.

Please let me know if this alternation from the VSCode API was done on purpose. I couldn't find a reason to not align with the VSCode API.

VSCode API for reference: https://code.visualstudio.com/api/references/vscode-api#window.

This ticket works toward resolving #6589.

#### How to test

This is rather a renaming, rather than a full change, but to ensure this still works you can use [this](https://github.com/microsoft/vscode-extension-samples/tree/main/extension-terminal-sample) sample extension.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin/src/theia.d.ts
+++ b/packages/plugin/src/theia.d.ts
@@ -2745,8 +2745,25 @@ export module '@theia/plugin' {
 
     /**
      * Options a virtual process terminal.
+     * @deprecated Use [ExtensionTerminalOptions](#ExtensionTerminalOptions) instead.
      */
     export interface PseudoTerminalOptions {
+        /**
+         * The name of the terminal.
+         */
+        name: string;
+
+        /**
+         * An implementation of [Pseudoterminal](#Pseudoterminal) where an extension can
+         * control it.
+         */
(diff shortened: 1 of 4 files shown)


packages/plugin/src/theia.d.ts
```diff
@@ -2745,6 +2745,7 @@ export module '@theia/plugin' {
 
     /**
      * Options a virtual process terminal.
+     * @deprecated Use [ExtensionTerminalOptions](#ExtensionTerminalOptions) instead.
```
colin-grant-work, line 2748: We usually include `@since <version>` with our deprecations to help decide when to remove a deprecated feature.

sgraband: I looked through other deprecated annotations and it seems like they all give the since information without the `@since` annotation but rather a plain `since` string. So i stayed with that pattern. Let me now if you would rather have it with the annotation.

colin-grant-work (approved):
Looks good to me 👍

PR #10685, opened 2022-01-28
fix: return focus to least recently active tab #10533

### LINKED ISSUE #10533: Closing a split editor tab, focus is not returned to the right tab
### Bug Description:
When closing an editor tab when we have split editors, the focus doesn't return to the tab that previously had the focus.

### Steps to Reproduce:
1. Open a few files ( at least 3) so that multiple tabs are opened. 
2. Drag and drop one of the tabs to create a split editor on the right
3. On the left side of the split screen, select one of the tabs that is not the right most tab
4. Close the editor tab on the right side of the split screen.
5. The focus is moved to the editor tab on the right ( instead of the one that was previously opened before)

### Additional Information
- Theia Version: 1.19

alxflam:
#### What it does
Fixes #10533: When the last tab of a TabBar is closed, always the last tab of the previous TabBar is activated.
This behavior differs from VSCode - the previously opened tab stays open.

https://user-images.githubusercontent.com/23746778/151578650-a7f0bd0a-004d-47af-a99b-996d0f512e16.mp4

#### How to test
- Open three files
- Move one file into a separate TabBar to the right
- Activate the leftmost tab in the first TabBar, then then single tab in the right TabBar
- Then close the single tab in the right TabBar
- Expected: The least recently activated tab is activated (displayed) in the first TabBar

https://user-images.githubusercontent.com/23746778/151578645-53facb79-cce9-42c7-9710-7609497ab05d.mp4

#### Review checklist

#### Reminder for reviewers

I couldn't spot any existing tests regarding the TabBar behavior - if i have missed them, please let me know. Locally my tests are passing for the core package.


packages/core/src/browser/shell/application-shell.ts
```diff
@@ -1744,8 +1744,6 @@ export class ApplicationShell extends Widget {
         if (!prevBar) {
             return false;
         }
-        const len = prevBar.titles.length;
-        prevBar.currentIndex = len - 1;
```
colin-grant-work, line 1748: I like the elegance of the solution :-). Perhaps, rather than removing this code completely, it should be put in a guard:

```typescript
if (!prevBar.currentTitle) {
    prevBar.currentIndex = prevBar.title.length - 1;
}
```
I'm not certain under what conditions `currentTitle` would be undefined, but this way we ensure that some widget on the previous bar is activated, but we don't change which widget is active if one is already marked so.

alxflam: sounds good, added the mentioned guard :+1:

colin-grant-work (approved):
The code looks good to me, and now the active tab is not changed when reverting to a previously active tabbar. 👍

PR #10688, opened 2022-01-29
skip `TypeScript: run reference code lens`

paul-marechal:
We'll need to un-skip this test sooner than later.


vince-fugnitto (approved):
The test has become increasingly problematic, I believe skipping it for the moment would be best 👍

PR #10694, opened 2022-01-31
Fix sorting problem in QuickFileOpenService

colin-grant-work:
#### What it does

There were a couple of problems in the sort function for the file quick pick and in the tests for them:

A positive returns means left [_after_](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description) right, so these values are reversed.

https://github.com/eclipse-theia/theia/blob/be4eba9d5aeb08e29f5a7dcf646c6defb25c28d4/examples/api-tests/src/file-search.spec.js#L41-L42

I've replaced the return expectations with actual passes through a sort function so that we're guaranteed to mirror the actual behavior of the function in its intended context of use.

---

To test whether L before R, these should also be reversed (but since the test expectations were reversed, the tests passed)

https://github.com/eclipse-theia/theia/blob/be4eba9d5aeb08e29f5a7dcf646c6defb25c28d4/packages/file-search/src/browser/quick-file-open.ts#L272

---

Fuzzy scores can be greater than `Scores.max` (1000), so imperfect matches could get a lead over perfect matches.

https://github.com/eclipse-theia/theia/blob/be4eba9d5aeb08e29f5a7dcf646c6defb25c28d4/packages/file-search/src/browser/quick-file-open.ts#L254

#### How to test

1. Mostly see if the tests pass.
2. Create a workspace with files with the filenames used in the new test.
3. Search for the perfect match.
4. Observe that it is ordered above the very good match.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/api-tests/src/file-search.spec.js
+++ b/examples/api-tests/src/file-search.spec.js
@@ -31,6 +31,9 @@ describe('file-search', function () {
 
         describe('#compareItems', () => {
 
+            /** @type import ('@theia/file-search/lib/browser/quick-file-open').QuickFileOpenService['compareItems']*/
+            const sortByCompareItems = (a, b) => quickFileOpenService['compareItems'](a, b);
+
             it('should compare two quick-open-items by `label`', () => {
 
                 /** @type import ('@theia/file-search/lib/browser/quick-file-open').FileQuickPickItem*/
@@ -38,8 +41,8 @@ describe('file-search', function () {
                 /** @type import ('@theia/file-search/lib/browser/quick-file-open').FileQuickPickItem*/
                 const b = { label: 'b', uri: new Uri.default('a') };
 
-                assert.equal(quickFileOpenService['compareItems'](a, b), 1, 'a should be before b');
-                assert.equal(quickFileOpenService['compareItems'](b, a), -1, 'a should be before b');
+                assert.deepEqual([a, b].sort(sortByCompareItems), [a, b], 'a should be before b');
+                assert.deepEqual([b, a].sort(sortByCompareItems), [a, b], 'a should be before b');
                 assert.equal(quickFileOpenService['compareItems'](a, a), 0, 'items should be equal');
             });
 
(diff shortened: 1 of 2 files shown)


vince-fugnitto (commented):
@colin-grant-work unfortunately the changes do not work for me, the file-search is broken when I tried in the browser and electron. The following error is reported multiple times:

```
quick-file-open.ts:247 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'score')
    at score (/home/evinfug/workspaces/theia/examples/electron/lib/packages_file-search_lib_browser_file-search-frontend-module_js.bundle.js:346)
    at Array.compareByLabelScore (/home/evinfug/workspaces/theia/examples/electron/lib/packages_file-search_lib_browser_file-search-frontend-module_js.bundle.js:355)
    at compareWithDiscriminators (/home/evinfug/workspaces/theia/examples/electron/lib/packages_file-search_lib_browser_file-search-frontend-module_js.bundle.js:523)
    at QuickFileOpenService.compareItems (/home/evinfug/workspaces/theia/examples/electron/lib/packages_file-search_lib_browser_file-search-frontend-module_js.bundle.js:361)
    at /home/evinfug/workspaces/theia/examples/electron/lib/packages_file-search_lib_browser_file-search-frontend-module_js.bundle.js:305
    at Array.sort (<anonymous>)
    at handler (/home/evinfug/workspaces/theia/examples/electron/lib/packages_file-search_lib_browser_file-search-frontend-module_js.bundle.js:305)
```

**Steps to Reproduce**:
1. start the application with `theia` as a workspace
2. search for `pack`
3. the file-search is broken

colin-grant-work:
> 3. the file-search is broken

@vince-fugnitto, good catch. The perils of only doing searches that work in a small workspace (and of trusting the declared types...)! I've pushed a commit that handles the `null` returns from `fuzzy`, and I also noticed that we were mishandling the `file results` (not `recently opened`) separator and fixed that.

vince-fugnitto (approved):
I confirm that the changes work as expected following the update, and the tests successfully pass :+1: 
I verified using a large workspace with exact, good and invalid results and it works fine, even with recent results.

JonasHelming (approved):
Looks good to me, I did a smoke test as well.

Issue #10337, opened 2021-10-28
Proposal: Playwright-based Framework for End-to-end Testing


planger:
Alongside unit and integration tests, it is often very useful to have end-to-end tests, which interact with the full application just like a user would do, e.g. for smoke testing the most relevant user flows. This entails that the end-to-end tests typically interact with the user interface and also only observe the results via the user interface (and not necessarily the internal state), just like a user.

### Proposal

To simplify writing end-to-end tests for Theia-based applications, we have developed a [Playwright](https://github.com/microsoft/playwright)-based [page object](https://martinfowler.com/bliki/PageObject.html) framework, which introduces an abstraction layer over Theia's user interfaces, encapsulating the details of the user interface interactions, wait conditions, etc., to help keeping end-to-end tests more concise, maintainable, and stable. This framework covers the generic user interface interactions of Theia applications, such as dealing with Theia views or editors (opening, closing, querying dirty state), file explorer, text editor, main menus, context menus, status bar elements, etc. Moreover, the framework is extensible so you can add dedicated page objects for custom Theia components, such as custom views, editors, status bar items, etc.

We'd be happy to contribute the page object framework to Theia!

### Example

Below, we show an example of how the framework's page objects `TheiaWorkspace`, `TheiaApp`, `TheiaTextEditor` etc. are used in a test.

#### Screencast of debugging the example
![debug-example](https://user-images.githubusercontent.com/588090/139237343-def01402-0a04-4a3f-852f-faf5ffc292b2.gif)

#### Code of the example test

```typescript
it("should undo and redo text changes and correctly update the dirty state", async () => {
  // 1. set up workspace contents and open Theia app
  const ws = new TheiaWorkspace(["tests/resources/sample-files1"]);
  const app = await TheiaApp.load(page, "http://localhost:3000", ws);

  // 2. open Theia text editor
  const sampleTextEditor = await app.openEditor("sample.txt", TheiaTextEditor);

  // 3. make a change and verify contents and dirty
  await sampleTextEditor.replaceLineWithLineNumber("change", 1);
  expect(await sampleTextEditor.textContentOfLineByLineNumber(1)).toBe(
    "change"
  );
  expect(await sampleTextEditor.isDirty()).toBe(true);

  // 4. undo and verify contents and dirty state
  await sampleTextEditor.undo(2);
  expect(await sampleTextEditor.textContentOfLineByLineNumber(1)).toBe(
    "this is just a sample file"
  );
[...]

PR #10494, opened 2021-11-30
testing: Add Theia Playwright framework (#10337)

planger:
Fixes https://github.com/eclipse-theia/theia/issues/10337

#### What it does

This change adds a page object framework based on Playwright that enables writing system tests (or end-to-end tests) of Theia-based applications more efficiently by introducing an abstraction layer over Theia's user interfaces, encapsulating the details of the user interface interactions, wait conditions, etc. This helps keeping your tests more concise, maintainable, and stable.

As part of this change, we also add tests for the introduced page objects, which run against the Theia example browser app in this repo. Currently, the framework is added as a separate package, which in a next step should be published and made consumable via npm, which we are happy to work on once this is merged.

#### How to test

With this change we added some initial documentation in the form of MD files on how to build, run and test the page object framework. Basically you'll have to run the following commands in the theia-playwright-testframework folder:

    yarn
    yarn theia:start &
    yarn ui-tests

There are other scripts for running the tests headfully or only run specific tests (see documentation).

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #10708, opened 2022-02-03
Run playwright tests in CI

planger:
#### What it does

  * Add script `yarn test:playwright` for conveniently running [playwright tests](https://github.com/eclipse-theia/theia/issues/10337) in the CI
  * Include invocation of above-mentioned script in `build.yml` 
    * Playwright tests are executed only once in `ubuntu-18.04, Node.js v16.x`

If you prefer, we could also run the playwright tests in a separate job on every PR / master commit.
Alternatively, we could only run the playwright tests periodically in a separate action.

I tend to prefer running the tests on every PR and master commit (within build.yml) either as part of the build job (as in this PR) or as a separate build job, as this raises breakages immediately when a change either broke expected behavior or broke the page objects. However, I understand that this would potentially add responsibilities to the committers. So alternatively, we could run those tests in a nightly job and we batch-fix the page objects periodically.

Please let me know what you think! Thanks!

#### How to test

  * To only run playwright tests, run `yarn && yarn build && yarn test:playwright`
  * Check CI build result of `Build / ubuntu-18.04, Node.js v16.x`

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,0 +1,48 @@
+name: Playwright Tests
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - master
+  schedule:
+    - cron: "0 4 * * *" # Runs every day at 4am: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule
+
+jobs:
+  build-and-test-playwright:
+    name: ubuntu-18.04,  Node.js 16.x
+
+    runs-on: ubuntu-18.04
+    timeout-minutes: 60
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Use Node.js "16.x"
+        uses: actions/setup-node@v1
+        with:
+          node-version: "16.x"
+          registry-url: "https://registry.npmjs.org"
+
+      - name: Use Python 3.x
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+
+      - name: Build
+        shell: bash
+        run: |
+          yarn --skip-integrity-check --network-timeout 100000
+          yarn browser build
+        env:
+          NODE_OPTIONS: --max_old_space_size=4096
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # https://github.com/microsoft/vscode-ripgrep/issues/9
+
+      - name: Test (playwright)
+        uses: GabrielBB/xvfb-action@v1
+        with:
+          run: yarn test:playwright
--- a/examples/playwright/configs/playwright.ci.config.ts
+++ b/examples/playwright/configs/playwright.ci.config.ts
@@ -1,0 +1,26 @@
+/********************************************************************************
+ * Copyright (C) 2022 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { PlaywrightTestConfig } from '@playwright/test';
+import baseConfig from './playwright.config';
+
+const ciConfig: PlaywrightTestConfig = {
+    ...baseConfig,
+    workers: 1,
+    retries: 1
+};
+
+export default ciConfig;
--- a/examples/playwright/package.json
+++ b/examples/playwright/package.json
@@ -19,6 +19,7 @@
     "lint": "eslint -c ./.eslintrc.js --ext .ts ./src ./tests",
     "lint:fix": "eslint -c ./.eslintrc.js --ext .ts ./src ./tests --fix",
     "ui-tests": "yarn && playwright test --config=./configs/playwright.config.ts",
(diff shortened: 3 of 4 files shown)


planger:
@vince-fugnitto Thanks again for reviewing https://github.com/eclipse-theia/theia/pull/10494!

I'd be very grateful if you could look at this PR too (as you already have the context) and let me know how whether you are fine with the way how the playwright build/verification is integrated.

Thanks!

vince-fugnitto:
> I tend to prefer running the tests on every PR and master commit (within build.yml) either as part of the build job (as in this PR) or as a separate build job, as this raises breakages immediately when a change either broke expected behavior or broke the page objects. However, I understand that this would potentially add responsibilities to the committers. So alternatively, we could run those tests in a nightly job and we batch-fix the page objects periodically.

@planger @JonasHelming should we make the decision in the next dev meeting? For example, who is expected to maintain these page object and how frequently do we update them (when they break vs before a release).

JonasHelming:
@vince-fugnitto : Yes, let's at least point to this discussion in the meeting. I added it to the agenda.
My POV: I would also run them on every PR and also directly maintain them if they break. I expect maintenance to be very very minimal. Only if a core UI element is drastically changes, the mapping in the page object model has to be adapted, which in the current state almost never happens. If it happens, it will be much easier to fix it immediatly, because you know what you changed. In the time period this PR was open, ~70 other PRs were merged without any effect. If we experience an overhead with this, we can still change this process, i.e. extract it into a separate build and do a separate maintainace step. We are happy to observe this, support committers when something breaks and adapt the process if required.

planger:
As discussed in the dev meeting, we decided to start with running the playwright tests on every PR & master commit and show the results in the list of checks, but don't make them blocking a merge if they happen to fail. This way we can observe their stability over a certain time period and decide later whether we make them a mandatory check or keep them just as information.

Thus, I moved the playwright test run into an own job in a separate workflow `.github/workflows/playwright.yml`, so the test run shows up as a separate check (see `Playwright Tests / ubuntu-18.04, Node.js 16.x`). Unfortunately, I can't see how this repo is configured regarding its branch rules due to missing permissions.
Are all checks currently configured to be mandatory? If yes, we will have to disable the `Playwright Tests` workflow in this list.

Also, I re-triggered the tests a couple of times to check their behavior on this infrastructure with this config and so far they look fine.

Please let me know if that meets the expectation as discussed in the dev meeting! Thanks a lot!

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Use Python 3.x
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+
+      - name: Build
+        shell: bash
+        run: |
+          yarn --ignore-engines --skip-integrity-check --network-timeout 100000
```
vince-fugnitto, line 39: `--ignore-engines` is no longer needed thanks to https://github.com/eclipse-theia/theia/commit/37ed897594a709d7e0a10d3db4c3bba85f00878a.

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,48 @@
...
+          yarn build:examples
```
vince-fugnitto, line 40: I don't think building both examples is necessary, the tests do not work on `example-electron` afaik.

examples/playwright/configs/playwright.ci.config.ts
```diff
@@ -0,0 +1,26 @@
+/********************************************************************************
+ * Copyright (C) 2021 EclipseSource and others.
```
vince-fugnitto, line 2:
```suggestion
 * Copyright (C) 2022 EclipseSource and others.
```

examples/playwright/configs/playwright.ci.config.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+import { PlaywrightTestConfig } from '@playwright/test';
+
+import baseConfig from './playwright.config';
```
vince-fugnitto, line 18: minor: newline and fixing the imports.

```suggestion
 ********************************************************************************/

import { PlaywrightTestConfig } from '@playwright/test';
import baseConfig from './playwright.config';
```

.github/workflows/playwright.yml
planger: Right, that makes the build a bit faster. I now use:
```
yarn --skip-integrity-check --network-timeout 100000
yarn browser build
```

planger:
Thanks a lot for your review @vince-fugnitto!

Hm, the 3PP License Check fail seems unrelated on a first glance, as we didn't any dependency in this change.
Or am I missing something?

```
WARN: Command [
  "java",
  "-jar",
  "/home/runner/work/theia/theia/scripts/download/dash-licenses.jar",
  "yarn.lock",
  "-batch",
  "50",
  "-timeout",
  "240",
  "-summary",
  "/home/runner/work/theia/theia/license-check-summary.txt"
] exited with code: 5
INFO: Checking results against the baseline...
X npm/npmjs/-/doctrine/2.1.0, Apache-2.0
ERROR: Found results that aren't part of the baseline!
error Command failed with exit code 1.
```

vince-fugnitto:
@planger I believe there might have been an error on dash-licenses yesterday which resulted in a few deps being marked as false positives. The dependency is approved (twice actually):
- https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/1987
- https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/1981

I do have the pull-request https://github.com/eclipse-theia/theia/pull/10725 which we can merge first if you'd like, it includes an upgrade and adds the approved deps to our baseline (until dash-licenses updates itself).

vince-fugnitto:
@planger #10725 was merged so if you rebase the CI should be all green :)

planger:
Thanks, rebased and keeping my fingers crossed :)

vince-fugnitto (approved):
LGTM :+1:

PR #9498, opened 2021-05-19
feat: upgrade authentication API to latest vscode

PhilippeVienne:
#### What it does
It implements the new authentication provider API from VSCode (the one moved from proposed to stable)

See #9345 for more details.

#### How to test
Use any vscode plugin depending on authentication like:
https://github.com/redhat-developer/vscode-redhat-account

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

Issue #9345, opened 2021-04-13
Authentication API in Theia does not match the latest vscode version


PhilippeVienne:
### Bug Description

VSCode has moved Authentication API from the proposed stage to the stable stage and this causes bugs on the method:

https://github.com/eclipse-theia/theia/blob/148e70c0e6baaec02c781cc31b56df5eb555d76f/packages/plugin/src/theia-proposed.d.ts#L117

has become

https://github.com/microsoft/vscode/blob/83f13c70ff500abe2d07938fedffbe45e5b6d172/src/vs/vscode.d.ts#L12689

### Steps to Reproduce:

1. Create an authentication Provider inside a VSCode plugin
2. Register the authentication provider using the latest VSCode api:
```
disposable = vscode.authentication.registerAuthenticationProvider(
    "rtezjigrezjklnjkgnrejginpzreger",
    "Smartfire",
    smartfireAuthenticationProvider);
context.subscriptions.push(disposable);
```
3. When theia register the provider, it contains only the name of the provider and not the provider itself, see debug output:
![image (5)](https://user-images.githubusercontent.com/1990262/114577010-f4637000-9c7b-11eb-9c17-9452f58a6af9.png)

### Additional Information

- Operating System: ubuntu
- Theia Version: latest

Issue #7243, opened 2020-02-28
Elm Extension won't start because it requires VSCode 1.42


JesterOrNot:
### Description
Language server fails to start due to an outdated version of VSCode

### Reproduction Steps
![image](https://user-images.githubusercontent.com/50386709/75571622-2cc7d080-5a1f-11ea-9f26-e908be14e03b.png)
Open https://gitpod.io/#https://github.com/jesterornot/gitpod-elm/tree/bug
The relevant extension is https://marketplace.visualstudio.com/items?itemName=Elmtooling.elm-ls-vscode

**OS and Theia version:**
Gitpod

**Diagnostics:** 
```
The language client requires VS Code version ^1.42.0 but received version 1.41.1
```

vince-fugnitto:
@JesterOrNot you can specify a `VSCODE_API_VERSION` when starting the application if you'd like, else I'd suggest using an older version of the `elm` extension. We are still working towards supporting newer versions of the API (and ultimately extensions). (related https://github.com/eclipse-theia/theia/issues/6804)

For the moment I think we can hold-off creating issues for each extension we encounter which requires the newer API as it will become noisy and instead work towards supporting the new API.
In the meantime it is acceptable to use older versions of extensions (it is actually what we do with the `builtins`).

akosyakov:
We should bump up VSCODE_API_VERSION. It's again `vscode-languageclient` defaults to 1.42. Any extension which is using this version regardless whether it actually uses unsupported by us features won't work.

akosyakov:
Yes, just checked that they are using progress enhancements from new `vscode-languageclient`: https://github.com/elm-tooling/elm-language-client-vscode/commit/b6516fb2ed38d5b070e47c48c69600d601af3382#diff-bc3c0f2fe3aca9ddbe96347ffb00641e

It is already supported by us.

akosyakov:
@JesterOrNot it can be bumped up here: https://github.com/eclipse-theia/theia/blob/764b74045fc1d8e1f4db2870e5517b5dd44693de/packages/plugin-ext-vscode/src/node/plugin-vscode-init.ts#L22

JesterOrNot:
On it!

sh7dm:
Vetur is also affected, v0.24.0 doesn't work without VS Code API v1.42.x

akosyakov:
@sh7dm Could you send a PR bumping up the version and check whether vetur works with it?

sh7dm:
[...]

PR #10709, opened 2022-02-03
Add support for newer authentication API matching VS Code 1.63.1

martin-fleck-at:
#### What it does

Fixes https://github.com/eclipse-theia/theia/issues/9345
- Stay compatible with newer and previous authentication API
-- Keep prev authentication API matching VS Code 1.53.0-next.ea1b3f27db
-- Add type files for newer authentication API and implement interfaces
-- Bridge API version gap in plugin context and with merged types

- Add support for 'onAuthenticationRequest' activation event
-- Allow dedicated trigger through 'ensureProvider' call
-- Trigger activation if session of provider is requested

**Note:** This PR is based on the work of @PhilippeVienne done in this PR: https://github.com/eclipse-theia/theia/pull/9498

#### How to test
I created two authentication provider plugins based on [the sample app](https://github.com/clovis-dugue/theia-authenticationapi-pr) [mentioned](https://github.com/eclipse-theia/theia/pull/9498#issuecomment-869734727) in the previous PR: https://github.com/martin-fleck-at/theia/tree/issue-9345-test

- One plugin is using the older VS Code API
- One plugin is using the newer VS Code API
- Provide commands to log in and log out of an account, we can see notifications and the proper Theia UI integration

authentication

- To test the `onAuthenticationRequest` activation, add `github-authentication` plugin

authentication-github

#### Review checklist

#### Reminder for reviewers


msujew:
As far as I understood it, this PR supersedes https://github.com/eclipse-theia/theia/pull/9498, right? Would you add a `Closes #9498` to your PR in that case? This will automatically close the old PR once your's is merged.

msujew:
You mentioned two extensions (with old/new vscode extension API) in your testing steps, but only link to [this one](https://github.com/clovis-dugue/theia-authenticationapi-pr). Can you provide both extensions somewhere, so we can successfully run through all the testing steps?

martin-fleck-at:
@msujew Absolutely, is it enough to simply add it to the description or do I need to add it in the commit message?

The link to the tests plugin is in the description but hidden among many others it seems ;-) https://github.com/martin-fleck-at/theia/tree/issue-9345-test. Specifically, the second commit adds the plugins (the first one being the PR): https://github.com/martin-fleck-at/theia/commit/6404ff90eed23feee213949f0f399975eac6f2eb

packages/plugin/src/theia.v1.63.authentication.d.ts
```diff
@@ -0,0 +1,191 @@
+/********************************************************************************
```
vince-fugnitto, line 1: Is this file really needed, or can we simply add it to `theia.d.ts` (is there a lot of incompatibilities between 1.53.2 and 1.63.0).

vince-fugnitto (commented):
I'm not sure that I follow the practicality of the pull-request, perhaps you can clarify.

Given how the framework officially supports the API `1.53.2` (with a certain compatibility), how would we also support extensions that make use of the authentication api of `1.63`. If a user were to bump the API in their product they would still be at a greater chance that the extension does not work for them given that all other `1.63.1` APIs are not implemented.

I'd probably be more fine if the changes were cleaner rather than versioning our files.

martin-fleck-at:
@vince-fugnitto Thank you for your fast response!

The goal of this PR is to make Theia compatible with newer versions of the authentication API. The main reason being that both the authentication provider interface and the registration of an authentication provider has changed over time and make some plugins, such as the github-authentication plugin, unusable. 

Maybe just to give a quick overview of the history:

- Theia - 2020: Implementation of the [authentication API in Theia](https://github.com/eclipse-theia/theia/pull/8402) from [API state](https://github.com/microsoft/vscode/blob/ab42ffc44c709aded523b15399a070ae8c724824/src/vs/vscode.proposed.d.ts#L98-L131) from April 23, 2020.
- VSCode - Janurary 11, 2021: The authentication provider registration is [revised](https://github.com/microsoft/vscode/blob/c852976/src/vs/vscode.proposed.d.ts#L129).
- VSCode - February 11, 2021: Release of v1.53.2: https://github.com/microsoft/vscode/tree/1.53.2
- VSCode - February 11, 2021: Further changes were made to the authentication API: [change event](https://github.com/microsoft/vscode/blob/9118a3461c52972f52a979349397268b81198ca1/src/vs/vscode.proposed.d.ts#L39-L53), [scopes](https://github.com/microsoft/vscode/blob/650906c3697b1ce2b9e6406a5a9bafdb93ee4771/src/vs/vscode.proposed.d.ts#L79), [method rename](https://github.com/microsoft/vscode/blob/eceff53351511e75b3c87f1600cebdd7f2df0737/src/vs/vscode.proposed.d.ts#L85-L92).
- VSCode - Feburary 23, 2021: API was moved from its proposed state to the stable VS Code API: https://github.com/microsoft/vscode/commit/67ded11020014caf9f9627f3756d398adaeae848.
- VSCode - March 2, 2021: Release of v1.54.0 containing the stable API: https://github.com/microsoft/vscode/tree/1.54.0.
- VSCode - up to now: Additional options for session retrieval (forceNewSession,  silent, clearSessionPreference) were added as optional fields to the API since then, not breaking any existing functionality afaik.

So, as it turns out we do not actually fully support the v1.53.2 since it would need to have the correct registration method but not the renamed methods, updated events or scope parameters. You can also observe this if you try the github-authentication plugin in version 1.53.2 which will not work, you would need to use a very specific commit, i.e., https://open-vsx.org/api/vscode/github-authentication/1.53.0-next.ea1b3f27db/file/vscode.github-authentication-1.53.0-next.ea1b3f27db.vsix.

Luckily, all those changes are made in a way that allows us to be backwards compatible. Specifically what we try to do in this commit is to bridge the difference between the APIs in the plugin context used by the plugins of the different versions. However, there are some changes (like changed types in the event) which need to be handled by merging the types and then querying the type, for instance in the sessions changed event in the authentication-main. In general, we aimed to make that change as least intrusive as possible. This is also the main reason, I moved the typings into a dedicated file instead of merging them into the Theia types (theia.d.ts), i.e., to make it obvious for the developer and to allow for easier transition if we ever were to add new versions, drop old ones, etc. Functionally, it does not make any difference at all so I can merge all types into one file if you prefer.

vince-fugnitto:
@martin-fleck-at thank you for the details :+1: 

> Luckily, all those changes are made in a way that allows us to be backwards compatible.

Given that the typings are backwards compatible (which should be the case in vscode since the API is finalized) I would merged them together with `theia.d.ts` and `theia-proposed.d.ts` respectfully. I'd avoid trying to version the files since it'll only get messier going forward, it might also be the case that we currently support newer APIs.

---

My main concern was that I was under the impression that in order to support newer extensions which are incompatible by definition (targeting a higher version than our default) was that it will cause issues on activation similarly to vscode (if it does not today I believe it might be a theia bug) - extensions from the vsx-registry view only install based on our default supported API :

image

In order to actually use an extension that targets a newer API an application developer would need to override the default value, or pass a value through the environment value, and as a consequence will start targetting a whole new set of extensions that are not compatible (going forward he might always pull, install, search extensions at `1.63` which is not implemented).

martin-fleck-at:
@vince-fugnitto I followed your advice and merged the two type definitions. I still left the part of the old API that was previously in the proposed types in it's place but moved everything that is part of the official VS Code API to the normal Theia types.

As far as I know, Theia does not check whether the installed plugins fulfill a certain version. You are correct that in the [OVSXClient](https://github.com/eclipse-theia/theia/blob/bf00fd0fcf5b8153178b9bab05212da1ea9ceee8/packages/vsx-registry/src/common/ovsx-client-provider.ts#L23) we use the version to retrieve the latest compatible version but other than that no constraints are made. This allows us actually to simply install plugins via the `theiaPlugins` in the `package.json`. I would hope that this does not get removed and mostly results in a warning or something similar that indicates that the API might not be compatbile.

vince-fugnitto:
> As far as I know, Theia does not check whether the installed plugins fulfill a certain version. 

@martin-fleck-at I believe we do, similarly to the following [issue](https://github.com/eclipse-theia/theia/issues/7243). I believe it might be a bug that we do not check the `engines.vscode` in this case and should fix it similarly to vscode. I don't think we want to arbitrarily allow any incompatible extension to be installed unless the application itself declares it supports it?

@msujew any thoughts?

martin-fleck-at:
@vince-fugnitto I see what you mean. Yes, currently the `engines.vscode` does not get checked for sure. Still, I hope that if we fix that, we still make it configurable for adopters. I understand that it makes sense to set a default VS Code version that Theia is (mostly) compatible with to restrict the search in the VSX registry. And I see that this version can be configured. However,  I can imagine a scenario where we want the Open VSX integration to work with a certain VS Code version (the one Theia is most compatible with) but still allow to add some additional plugins in the `theiaPlugins` section of the plugin.xml for a given product. For sure, the adopter has to take the responsibility to make sure the plugins work as expected but I think it is better than to absolutely forbid the installation.

vince-fugnitto:
@martin-fleck-at right, by default we should align and do the check since it is the most basic verification we have not to pull extensions which have a greater chance of incompatibility with the current version of the framework. Downstream they can override this behavior or update the default API, but of course with the risk of using extensions which may be incompatible or use a different node runtime (for the most part they will want to stick with the framework). 

If the changes work with `1.53.2` I'm fine with it, I was expecting however that the newer extension fails immediately due to `engines.vscode` so it's why I wanted to share my concern regarding the practicality of why we would support such a bump in API.

msujew:
@vince-fugnitto We definitely don't check the `engines.vscode` version when installing/loading extensions, I would've been pretty annoyed by that by now while testing PRs :D

Fixing this behavior would probably be a good idea anyway, but we should probably make this configurable using a build-config entry or preference. I can imagine a lot of downstream-users accidentally using extensions with "invalid" `engines.vscode` fields which still run perfectly fine inside of Theia.

In general, as long as this PR does not break the existing "old" auth API but also adds support for the new API I'm completely fine with it. I will take a closer look at the changes later today 👍

packages/plugin-ext/src/main/browser/plugin-authentication-service.ts
```diff
@@ -0,0 +1,70 @@
...
+                        resolve(provider);
+                    } else {
+                        throw new Error(`No authentication provider '${providerId}' is currently registered.`);
+                    }
+                }
+            });
+        });
+
+        const didTimeout: Promise<AuthenticationProvider> =
+            new Promise((_, reject) => setTimeout(() => reject('Timed out waiting for authentication provider to register'), 5000));
```
msujew, line 66: Suggestion: We already have a `timeout` function in the `promise-util` which works quite similar. But instead of rejecting the promise, the `timeout` function resolves it. What do you think about adding a `timeoutReject` method which looks something like this?

```ts
export function timeoutReject(ms: number, message?: string): Promise<void> {
    const deferred = new Deferred<void>();
    setTimeout(() => deferred.reject(new Error(message)), ms);
    return deferred.promise;
}
```

This could also be used [here](https://github.com/eclipse-theia/theia/blob/37ed897594a709d7e0a10d3db4c3bba85f00878a/packages/monaco/src/browser/monaco-editor-provider.ts#L307).

I also thought about an additional override for `timeout`, but switching whether the promise resolves or rejects depending on an optional argument seems to me like a weird change of semantics.

Your opinion on that?

packages/plugin-ext/src/plugin/plugin-context.ts
```diff
@@ -215,14 +215,40 @@ export function createAPIFactory(
 
     return function (plugin: InternalPlugin): typeof theia {
         const authentication: typeof theia.authentication = {
-            registerAuthenticationProvider(provider: theia.AuthenticationProvider): theia.Disposable {
-                return authenticationExt.registerAuthenticationProvider(provider);
+            // support older (< 1.53.0) and newer version of authentication provider registration
+            registerAuthenticationProvider(
+                id: string | theia.AuthenticationProvider, label?: string, provider?: theia.AuthenticationProvider, options?: theia.AuthenticationProviderOptions):
+                theia.Disposable {
```
msujew, line 221: Suggestion: This implementation seems a bit convoluted to me. Perhaps we could use something like this?

```ts
registerAuthenticationProvider(id: string | theia.AuthenticationProvider, label?: string, provider?: theia.AuthenticationProvider, options?: theia.AuthenticationProviderOptions): theia.Disposable {
    // Add some comment here about why we need the merged definition
    type MergedAuthProvider = theia.AuthenticationProvider & {
        createSession(scoped: string[]): Thenable<theia.AuthenticationSession>
        removeSession(sessionId: string): Thenable<void>
        getSessionsWithoutScopes(): Thenable<ReadonlyArray<theia.AuthenticationSession>>
    };
    const isNew = typeof id === 'string';
    const providerId = isNew ? id : id.id;
    const mergedProvider = (isNew ? provider : id) as MergedAuthProvider;

    if (!mergedProvider.createSession) {
         mergedProvider.createSession = (scopes: string[]) => mergedProvider.login(scopes);
    }
    if (!mergedProvider.removeSession) {
        mergedProvider.removeSession = (sessionId: string) => mergedProvider.logout(sessionId);
    }
    if (!mergedProvider.getSessionsWithoutScopes) {
         mergedProvider.getSessionsWithoutScopes = mergedProvider.getSessions;
    }
    return authenticationExt.registerAuthenticationProvider(providerId, label!, mergedProvider, isNew ? options : {
          supportsMultipleAccounts: mergedProvider.supportsMultipleAccounts
    });
},
```

As far as I can tell, we don't need the following bit:

```ts
newProvider['getSessions'] = (_scopes?: string[]) => (newProvider as any)['getSessionsWithoutScopes']();
```

msujew (commented):
@martin-fleck-at This is looking really good already. I have a few remarks about the code, but the functionality seems to work well already 👍

packages/plugin-ext/src/main/browser/authentication-main.ts
```diff
@@ -196,20 +208,39 @@ export class AuthenticationMainImpl implements AuthenticationMain {
         return allow;
     }
 
-    protected async loginPrompt(providerName: string, extensionName: string): Promise<boolean> {
-        const choice = await this.messageService.info(`The extension '${extensionName}' wants to sign in using ${providerName}.`, 'Allow', 'Cancel');
+    protected async loginPrompt(providerName: string, extensionName: string, recreatingSession: boolean, _detail?: string): Promise<boolean> {
+        const message = recreatingSession
+            ? nls.localize('confirmRelogin', "The extension '{0}' wants you to sign in again using {1}.", extensionName, providerName)
+            : nls.localize('confirmLogin', "The extension '{0}' wants to sign in using {1}.", extensionName, providerName);
```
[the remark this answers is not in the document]
martin-fleck-at: Thank you very much for the explanation, I updated it as you suggested!

packages/plugin-ext/src/main/browser/plugin-authentication-service.ts
martin-fleck-at: I think that is a great idea! I was actually looking for a function like that but then simply decided to do it like in the vscode code. I took the method that you proposed since I am also not a huge fan of a Boolean parameter to switch between resolution and rejection since I know that I always have to double and triple check Boolean parameters when reading/writing to ensure correct usage ;-)

martin-fleck-at:
@msujew Thank you very much for your thorough feedback! You were correct in all points and I adapted the code accordingly - I think it looks much cleaner now. I also adapted the [test commit](https://github.com/martin-fleck-at/theia/commits/issue-9345-test) to have the old provider register without any of the new API to see if the delegation and everything still works as expected.

If you have any further comments, please let me know.

packages/plugin-ext/src/plugin/plugin-context.ts
msujew: I see, I didn't look into it too much and thought `getSessionsWithoutScopes` was part of the API. Yeah, JS runtimes don't care about parameters :)

packages/plugin-ext/src/main/browser/plugin-authentication-service.ts
```diff
@@ -39,32 +37,35 @@ export class PluginAuthenticationServiceImpl extends AuthenticationServiceImpl i
...
+    protected async waitForProviderRegistration(providerId: string): Promise<AuthenticationProvider> {
+        const waitForRegistration = new Deferred<AuthenticationProvider>();
+        const registration = this.onDidRegisterAuthenticationProvider(info => {
+            if (info.id === providerId) {
+                registration.dispose();
+                const provider = this.authenticationProviders.get(providerId);
+                if (provider) {
+                    waitForRegistration.resolve(provider);
+                } else {
+                    throw new Error(`No authentication provider '${providerId}' is currently registered.`);
```
msujew, line 65: Suggestion: I don't believe we should throw this error here, but instead call `waitForRegistration.reject` with it? Looking at it, this seems to neither reject nor resolve the promise when no authentication provider with the given id is registered. Luckily, we always fail after 5 seconds using the timeout in that case, but I believe we should handle this more gracefully. (I know that this is kind of "old code")

msujew (approved):
Great, the code is looking good for me, and I could not find any issues in the functionality of this change. I have a minor suggestion though 👍

- [x] Old API continues to work correctly
- [x] New API is working correctly
- [x] Using both APIs is possible and works correctly as well

martin-fleck-at:
@msujew Thank you very, very much for your continuous and thorough feedback, I really appreciate that! I fixed the last issue and squashed the commits so that they can be merged if the build runs through.

packages/core/src/browser/authentication-service.ts
```diff
@@ -99,72 +130,84 @@ export interface AuthenticationService {
...
+        // register new command and menu entry for the sessions account
+        const disposables = new DisposableCollection();
+        disposables.push(this.commands.registerCommand({ id: providerAccountId }, {
+            execute: async () => {
+                this.signOutOfAccount(providerId, session.account.label);
+            }
+        }));
+        this.menus.registerSubmenu(providerAccountSubmenu, `${session.account.label} (${providerLabel})`);
+        disposables.push(this.menus.registerMenuAction(providerAccountSubmenu, {
+            label: 'Sign Out',
```
msujew, line 205: One last thing, this should be localized:
```suggestion
            label: nls.localizeByDefault('Sign Out'),
```

martin-fleck-at: Done! I actually found a few more places I internationalized. Hopefully it is fine now.

packages/plugin-ext/src/main/browser/authentication-main.ts
```diff
@@ -86,75 +88,85 @@ export class AuthenticationMainImpl implements AuthenticationMain {
...
             }));
             items.push({
-                label: 'Sign in to another account',
+                label: nls.localizeByDefault('Sign in to another account'),
                 value: { session: undefined }
             });
             const selected = await this.quickPickService.show(items,
                 {
-                    title: `The extension '${extensionName}' wants to access a ${providerName} account`,
+                    title: nls.localize('confirmAuthenticationAccess', "The extension '{0}' wants to access a {1} account", extensionName, providerName),
```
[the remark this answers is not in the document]
martin-fleck-at: Done!

PR #10719, opened 2022-02-07
Align preference and notification styling with VSCode

msujew:
#### What it does

Updates our hovering and focus behavior for preferences and notifications to behave like the current version of VSCode.

#### How to test

Both of these tests should be performed for different light/dark themes. I changed the color slightly for preferences in the light theme compared to VSCode, since hovering and focus color were completely transparent. In my opinion this didn't really look good.

---

1. Open the example app, hover over and click on one of the browser-endpoint notifications.
2. Open the notification center, again, checking hover and click/focus behavior
3. Assert that it aligns to VSCode's behavior

---

1. Open the preference widget
2. Hover over different preferences and click on them (focus). The `More Actions...` gear should only appear when a preference is focused.
3. Change a preference so that it deviates from its original default value. Assert that the appearing vertical bar looks correct.
4. When clicking on a referenced preference, the target preference should be displayed in the center of the widget and should automatically gain the focus.
5. Assert that is aligns to VSCode's behavior

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/browser/style/index.css
+++ b/packages/preferences/src/browser/style/index.css
@@ -224,20 +224,28 @@
 
 .theia-settings-container li.single-pref {
     list-style-type: none;
-    margin: 12px 0 18px 0;
+    padding: 12px 14px 18px;
     width: 100%;
     display: flex;
     flex-direction: column;
     box-sizing: border-box;
-    padding-left: calc(2 * var(--theia-ui-padding));
     position: relative;
+}
+
+.theia-settings-container li.single-pref:hover:not(:focus) {
+    background-color: var(--theia-settings-rowHoverBackground);
+}
+
+.theia-settings-container li.single-pref:focus {
+    background-color: var(--theia-settings-focusedRowBackground);
+    outline: 1px solid var(--theia-settings-focusedRowBorder);
 }
 
 .theia-settings-container li.single-pref .pref-context-gutter {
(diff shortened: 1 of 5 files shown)


packages/preferences/src/browser/style/index.css
```diff
@@ -247,8 +255,6 @@
 }
 
 .theia-settings-container .single-pref:focus-within .pref-context-gutter .settings-context-menu-btn,
-.theia-settings-container .pref-name:hover+.pref-context-gutter .settings-context-menu-btn,
-.theia-settings-container .pref-context-gutter:hover .settings-context-menu-btn,
```
colin-grant-work, line 251: With these changes (I think), it's now possible to hover over where the gear button should be and not reveal it if the preference line isn't focused - it remains opacity 0.

colin-grant-work (commented):
Apart from the bug mentioned below, the changes look good. One other thing I noticed is that we currently allow selection of the text of the preference name and description where VSCode does not, but that doesn't have to be addressed here.

packages/preferences/src/browser/style/index.css
msujew: Good catch, I noticed that these changes weren't needed at all :)

colin-grant-work (approved):
Looks good to me, in various themes. 👍

PR #10737, opened 2022-02-10
ts: enable `noImplicitOverride`

paul-marechal:
This feature helps with quickly identifying overridden methods, as well
as notify us of potential breakage whenever APIs are removed and some
classes still try to override those missing methods.

It seems like this change wouldn't even force dependents to update their TS version as the `override` keyword doesn't make its way into the generated `.d.ts` files.

#### How to test

Try to remove one of the `override` and build: it should fail.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/editor/debug-hover-widget.ts
+++ b/packages/debug/src/browser/editor/debug-hover-widget.ts
@@ -56,7 +56,8 @@ export function createDebugHoverWidgetContainer(parent: interfaces.Container, ed
 @injectable()
 export class DebugHoverWidget extends SourceTreeWidget implements monaco.editor.IContentWidget {
 
-    protected readonly toDispose = new DisposableCollection();
+    // TODO: Is this still needed?
+    // protected override toDispose = new DisposableCollection();
 
     @inject(DebugEditor)
     protected readonly editor: DebugEditor;
@@ -85,7 +86,7 @@ export class DebugHoverWidget extends SourceTreeWidget implements monaco.editor.
     }
 
     @postConstruct()
-    protected init(): void {
+    protected override init(): void {
         super.init();
         this.domNode.className = 'theia-debug-hover';
         this.titleNode.className = 'theia-debug-hover-title';
@@ -107,15 +108,15 @@ export class DebugHoverWidget extends SourceTreeWidget implements monaco.editor.
         ]);
     }
 
-    dispose(): void {
+    override dispose(): void {
         this.toDispose.dispose();
     }
 
-    show(options?: ShowDebugHoverOptions): void {
+    override show(options?: ShowDebugHoverOptions): void {
         this.schedule(() => this.doShow(options), options && options.immediate);
     }
 
-    hide(options?: HideDebugHoverOptions): void {
+    override hide(options?: HideDebugHoverOptions): void {
         this.schedule(() => this.doHide(), options && options.immediate);
     }
 
@@ -228,7 +229,7 @@ export class DebugHoverWidget extends SourceTreeWidget implements monaco.editor.
         } : undefined!;
     }
 
-    protected onUpdateRequest(msg: Message): void {
+    protected override onUpdateRequest(msg: Message): void {
         super.onUpdateRequest(msg);
         const { expression } = this.hoverSource;
         const value = expression && expression.value || '';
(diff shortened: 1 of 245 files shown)


paul-marechal:
Doing this made me notice how many `readonly` we have, I don't actually think those provide much value in most cases...

Worth a PR to remove those too? I'd be in favor.

JonasHelming:
How have you created these override annotations? Was there an automated way?

paul-marechal:
@JonasHelming I didn't find an automated way to do it than ctrl-clicking on TS build errors and pasting the missing keywords. I see this as a small time investment to set this all up :)

paul-marechal:
@colin-grant-work I think we could wholesale remove the `readonly` from all injected properties.

For one for property injection to work something needs to write from the exterior into those fields. So far Inversify is doing it because it doesn't care about TS `readonly` (and can't).

It would also remove some noise:

```ts
// from
@inject(Something) protected readonly something: Something;
@inject(Something) protected readonly something: Something;
@inject(Something) protected readonly something: Something;
@inject(Other) protected readonly other: Other;

// to
@inject(Something) protected override something: Something;
@inject(Other) protected other: Other;
```

paul-marechal:
@colin-grant-work I think we could wholesale remove the `readonly` from all injected properties.

For one for property injection to work something needs to write from the exterior into those fields. So far Inversify is doing it because it doesn't care about TS `readonly` (and can't). I just find this modifier to be noisy in this case. Moreover it should be common practice to not just re-assign protected fields without knowing what you are doing.

colin-grant-work:
> ...it should be common practice to not just re-assign protected fields without knowing what you are doing.

Many things are good practice that we nevertheless ask our tools to help us do :-).

packages/debug/src/browser/editor/debug-hover-widget.ts
```diff
@@ -56,7 +56,8 @@ export function createDebugHoverWidgetContainer(parent: interfaces.Container, ed
 @injectable()
 export class DebugHoverWidget extends SourceTreeWidget implements monaco.editor.IContentWidget {
 
-    protected readonly toDispose = new DisposableCollection();
+    // TODO: Is this still needed?
+    // protected override toDispose = new DisposableCollection();
```
colin-grant-work, line 60: Do you want to address this as part of this PR?

colin-grant-work (commented):
Looks good to me. I checked the cases where you've deleted injections, and they all look correct. One potential hanging comment, but if that's intentional, then it's good to go.

packages/debug/src/browser/editor/debug-hover-widget.ts
paul-marechal: Honestly if we catch bugs by turning this flag on then yes:

If we keep this line, it will override the previous declaration of the same field. Except that the previous declaration holds references to several disposables. Is that a bug?

paul-marechal: We have `DebugHoverWidget#toDispose > SourceTreeWidget > TreeWidget > ReactWidget > Widget#toDispose`.

And it looks like a mistake to override `toDispose` in `DebugHoverWidget`.

paul-marechal:
@colin-grant-work Agreed, just like what is being done in this PR :)

The difference I would make is that `readonly` semantics depend on our implementation's design, whereas `override` is systematic though we benefit from it with easier understanding of where methods come from.

And I don't like our uses of `readonly` as it feel it got used systematically rather than trying to properly design our API...

packages/debug/src/browser/editor/debug-hover-widget.ts
colin-grant-work: I agree, it seems as though the `DisposableCollection` from the super class should not be clobbered here. The first bug fixed by the `override` marker already. :-)

colin-grant-work:
> And I don't like our uses of `readonly` as it feels like it got used mechanically rather than trying to properly design our APIs...

I don't disagree with you about the substance of your claim, but you may take my argument as a warning that a PR removing a lot of `readonly`'s probably isn't a good idea, because it would require a lot more care on the part of the reviewer than a PR like this one, which is almost purely mechanical, for not much benefit (but perhaps not _no_ benefit), so you might not find eager reviewers.

vince-fugnitto (approved):
The changes look good to me :+1: CI is also successful (build, tests)

colin-grant-work (approved):
I'll add my endorsement as well. 👍

PR #10745, opened 2022-02-13
Fix frontend use of Uri.joinPath with path-browserify

thegecko:

#### What it does

Using `vscode.Uri.joinPath()` or `theia.Uri.joinPath()` in frontend modules (Theia frontend plugins or VS Code Web Extensions) was failing because `path` was not found:

```
TypeError: Cannot read properties of undefined (reading 'join')
```

This PR adds the `path-browserify` polyfill to fix the issue.

Note, a new dependency has been added which uses the MIT license, does this need to be tracked or signed off?

[path-browserify](https://github.com/browserify/path-browserify)

#### How to test

Create a frontend module which calls `Uri.joinPath()`

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-manager/package.json
+++ b/dev-packages/application-manager/package.json
@@ -48,6 +48,7 @@
     "ignore-loader": "^0.1.2",
     "less": "^3.0.3",
     "node-abi": "*",
+    "path-browserify": "^1.0.1",
     "semver": "^7.3.5",
     "setimmediate": "^1.0.5",
     "source-map": "^0.6.1",
--- a/dev-packages/application-manager/src/generator/webpack-generator.ts
+++ b/dev-packages/application-manager/src/generator/webpack-generator.ts
@@ -195,7 +195,7 @@ module.exports = {
             'child_process': false,
             'crypto': false,
             'net': false,
-            'path': false,
+            'path': require.resolve('path-browserify'),
             'process': false,
             'os': false,
             'timers': false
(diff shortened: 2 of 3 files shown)


vince-fugnitto:
> This PR adds the path-browserify polyfill to fix the issue.
> Note, a new dependency has been added which uses the MIT license, does this need to be tracked or signed off?

@thegecko the automated _IP Due Diligence_ check passed with the new dependency, it should be good to use :+1: 
- https://github.com/eclipse-theia/theia/runs/5174618930?check_suite_focus=true

vince-fugnitto (approved):
Infact it looks like it might be necessary based on the [webpack documentation](https://webpack.js.org/configuration/resolve/#resolvefallback).
I confirmed that the `joinPath` works in web extensions:

image

paul-marechal:
@thegecko I remembered that we had a small implementation of `path` in the plugin code: https://github.com/eclipse-theia/theia/blob/443130396f7b1118eaf02e9309be555c938409ae/packages/plugin-ext/src/plugin/path.ts

Would that suffice or is it better to use the full npm polyfill? Just trying to figure out if we have redundancy here.

thegecko:
> Is the change still necessary, is it not handled by webpack?

Yeah, webpack 5 no longer bundles the polyfills

> I remembered that we had a small implementation of path in the plugin code

Hmm, this may work for the 2 functions defined, but there may be other path functions/cases in future. Personally I'd rather rely on the recommended polyfill.

Will merge for now, but happy to revisit using the leaner path implementation in a future PR.

PR #7513, opened 2020-04-07
Update loading animation

sh7dm:
#### What it does

Fixes #6463.

#### How to test

Visually, inspect SVG file of the image.

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

Issue #6463, opened 2019-10-29
Improve default loading animation


vince-fugnitto:
**Description**

I believe the default loading animation should be improved as it is quite pixelated on certain types of screens. See https://github.com/eclipse-theia/theia/issues/4102#issuecomment-465586438. The same is true for Firefox.

<div align='center'>

![image](https://user-images.githubusercontent.com/40359487/67765651-0703ce00-fa23-11e9-83aa-7490f39a91fe.png)

</div>

**Additional Information**

A previous attempt was made in the past to update the animation (https://github.com/eclipse-theia/theia/pull/3156), but a more generic one should be present so it can be used by adopters without requiring the need to update it.

PR #10761, opened 2022-02-16
core: update default loading animation

vince-fugnitto:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/6463.
Closes: https://github.com/eclipse-theia/theia/pull/7513.

The pull-request updates the default loading animation on startup to instead use the existing `codicon` icon for consistency with the rest of the framework. 

The pull-request contains the minimal changes from #7513 without the need to replace the existing icon, or preload template not to break downstream applications.

https://user-images.githubusercontent.com/40359487/154357336-be912ec7-0e24-4013-a574-adde067a65ee.mov

#### How to test

1. build the example-browser or example-electron application
2. start the application - the new codicon icon should be used
3. verify the changes with different themes - both light and dark

#### Review checklist

#### Reminder for reviewers

Co-authored-by: Dmitry Sharshakov <d3dx12.xx@gmail.com>

### DIFF
--- a/packages/core/src/browser/style/index.css
+++ b/packages/core/src/browser/style/index.css
@@ -99,12 +99,30 @@ blockquote {
   bottom: 0;
   z-index: 50000;
   background: var(--theia-editor-background);
-  background-image: var(--theia-preloader);
   background-size: 60px 60px;
   background-repeat: no-repeat;
   background-attachment: fixed;
   background-position: center;
   transition: opacity 0.8s;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.theia-preload::after {
+  animation: 1s theia-preload-rotate infinite;
+  color: #777; /* color works on both light and dark themes */
+  content: "\eb19"; /* codicon-load */
+  font: normal normal normal 72px/1 codicon;
+}
+
+@keyframes theia-preload-rotate {
+  0% {
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
 }
 
 .theia-preload.theia-hidden {


packages/core/src/browser/style/index.css
```diff
@@ -99,12 +99,30 @@ blockquote {
...
   background-position: center;
   transition: opacity 0.8s;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.theia-preload::after {
+  animation: 1s theia-preload-rotate infinite;
+  color: #777; /* color works on both light and dark themes */
```
vince-fugnitto, line 114: Unfortunately we cannot re-use a variable here like `var(theia-foreground` since it is not ready and would cause a flicker. Instead we can use a color which should work for most themes.

JonasHelming (approved):
Looks nice, thanks for taking care of this!

msujew (commented):
I believe using the `load` codicon (`\eb19`) would be a better fit here perhaps? At least to me, it looks more visually pleasing for this use case.

vince-fugnitto:
> I believe using the `load` codicon (`\eb19`) would be a better fit here perhaps? At least to me, it looks more visually pleasing for this use case.

Sounds good! I updated it and agree it looks better in this case:


https://user-images.githubusercontent.com/40359487/154357336-be912ec7-0e24-4013-a574-adde067a65ee.mov

msujew (approved):
Thanks, looks great!

kittaakos:
@vince-fugnitto, thanks for the change. It looks excellent 😍  However, I want to keep the old spinner for a while but use the latest Theia. How can I customize it? In my extension, I can override the `.theia-preload` selector, but this is hardcoded here:

https://github.com/eclipse-theia/theia/blob/70466bb4b7283187b019defec095f8dea42d47c6/dev-packages/application-manager/src/generator/frontend-generator.ts#L56

So at app startup, I can see the new Theia spinner for a second or two if I add these selectors to my styles:

```css
/* restore the old Theia spinner */
.theia-preload {
  background-image: var(--theia-preloader) !important;
  display: unset;
  justify-content: unset;
  align-items: unset;
}

.theia-preload::after {
  animation: none;
  color: unset;
  content: none;
  font: unset;
}
```


https://user-images.githubusercontent.com/1405703/169263672-0f379388-7085-4b07-aab6-d7c270ff726a.mp4


I also tried to customize it on the code level [here](https://github.com/eclipse-theia/theia/blob/70466bb4b7283187b019defec095f8dea42d47c6/packages/core/src/browser/frontend-application.ts#L202-L205) add replaced the CSS selectors on the DOM, but it's too late. How does one change the app spinner? Thank you!


I also noticed that the old spinner is still in the app and used by the mini-browser? Why is this difference in the Theia UI?


I appreciate your support.

vince-fugnitto:
@kittaakos you can provide your own custom preload element similar to https://github.com/eclipse-theia/theia-blueprint/pull/73 (the actual splash screen can be whatever you'd like, and could be the old spinner if you require it).

> I also noticed that the old spinner is still in the app and used by the mini-browser? Why is this difference in the Theia UI?

At the time I wanted to keep the changes minimal to only the splash screen and also make that loading animation different than regular loading within the app. We can of course update so that both are consistent if we want.

kittaakos:
Thank you so much for your time and help, @vince-fugnitto. 🙏 I was unaware of the custom preload option.

Finally, I ended up patching the generated `index.html` after the app build.


----

> I wanted to keep the changes minima

I see. It makes sense.

PR #10767, opened 2022-02-17
Copy/paste not working within folder in Explorer

### LINKED ISSUE #9923: Copy/paste not working within folder in Explorer
### Bug Description:

Copying and pasting files within a single folder does not work in the Explorer. Neither Cmd+C and Cmd+V, nor Opt+Drag of the mouse will create a copy of a file in the same folder.

### Steps to Reproduce:

1. Select a file in some folder (call it `src`) in the Explorer view.
2. Copy the file with Cmd+C.
3. Attempt to paste into the same folder `src` with Cmd+V.

Expected result:

1. A new copy of the file is created with a distinct name. The usual pattern in macOS is that Xyz.ext is pasted as "Xyz copy.ext" or "Xyz copy 1.ext" if "Xyz copy.ext" already exists, etc.

Actual result:

1. A sequence of messages are shown in the status bar, too fleeting to read, and no new file is created.

The same occurs when trying to drag and drop the file using the mouse pointer with Option key modifier.

Note that pasting a copied file into a different folder than the original works (at least, it does if the destination folder doesn't already have a file with that name).

### Additional Information

- Operating System: macOS Catalina 10.15.7 (19H15)
- Theia Version: Theia Blueprint Version 1.14.0 (1.14.0.17)

AlexandraBuzila:
#### What it does

Fixes copy/pasting of a file or folder within the same parent folder in the Explorer, by renaming the target file if the source and target URIs are the same. 
Also changes the naming scheme to use spaces instead of underscores when creating a new file, to align with VS Code and user expectations (e.g. on macOS).

Fixes #9923 

##### Testing copy/paste in the same parent folder:
1. Copy a file/folder in the Explorer view
2. Paste in the same folder
3. A copy of the file/folder should be created (its name should be in the form `originalName copy.ext` or 
`originalName copy index.ext`)

##### Testing the name generation of new files:
Several commands were affected by the change in naming for new files/folders. New files/folders should now have spaces instead of underscores in the generated name portion:

1. the **New File** / **New Folder** commands: from the _File_ menu, select _New File_: the name suggestion in the dialog that opens should use spaces, e.g. `Untitled 2.txt`
2. the **Duplicate** command: select a file in the Explorer and select Duplicate from its context menu: a copy of the file is created and its name has ` copy <index>` appended to the original file name
3. copy a file in the Explorer view and paste in into a folder that already contains a file with the same name: a copy of the file is created and its name has ` copy <index>` appended to the original file name

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/browser/file-tree/file-tree-model.ts
+++ b/packages/filesystem/src/browser/file-tree/file-tree-model.ts
@@ -151,13 +151,16 @@ export class FileTreeModel extends CompressedTreeModel implements LocationServic
     }
 
     async copy(source: URI, target: Readonly<FileStatNode>): Promise<URI> {
-        let targetUri = target.uri.resolve(source.path.base);
+        /** If the target is a file or if the target is a directory, but is the same as the source, use the parent of the target as a destination. */
+        const parentNode = (target.fileStat.isFile || target.uri.isEqual(source)) ? target.parent : target;
+        if (!FileStatNode.is(parentNode)) {
+            throw new Error('Parent of file has to be a FileStatNode');
+        }
+        let targetUri = parentNode.uri.resolve(source.path.base);
         try {
-            if (source.path.toString() === target.uri.path.toString()) {
-                const parent = await this.fileService.resolve(source.parent);
-                const name = source.path.name + '_copy';
-                targetUri = FileSystemUtils.generateUniqueResourceURI(source.parent, parent, name, source.path.ext);
-            }
+            const parent = await this.fileService.resolve(parentNode.uri);
+            const sourceFileStat = await this.fileService.resolve(source);
+            targetUri = FileSystemUtils.generateUniqueResourceURI(parent, targetUri, sourceFileStat.isDirectory, 'copy');
             await this.fileService.copy(source, targetUri);
         } catch (e) {
             this.messageService.error(e.message);
--- a/packages/filesystem/src/common/filesystem-utils.ts
+++ b/packages/filesystem/src/common/filesystem-utils.ts
@@ -18,24 +18,47 @@ import { FileStat } from '../common/files';
 import URI from '@theia/core/lib/common/uri';
 
 export namespace FileSystemUtils {
+    export const FILE_NAME_SEPARATOR = ' ';
 
     /**
      * Generate unique URI for a given parent which does not collide
      *
-     * @param parentUri the `URI` of the parent
      * @param parent the `FileStat` of the parent
-     * @param name the resource name
-     * @param ext the resource extension
+     * @param targetUri the initial URI
+     * @param isDirectory indicates whether the given targetUri represents a directory
+     * @param suffix an optional string to append to the file name, in case of collision (e.g. `copy`)
      */
-    export function generateUniqueResourceURI(parentUri: URI, parent: FileStat, name: string, ext: string = ''): URI {
+    export function generateUniqueResourceURI(parent: FileStat, targetUri: URI, isDirectory: boolean, suffix?: string): URI {
         const children = !parent.children ? [] : parent.children!.map(child => child.resource);
+        let name = targetUri.path.name;
+        let extension = targetUri.path.ext;
+        if (!name) {
+            // special case for dotfiles (e.g. '.foobar'): use the extension as the name
+            name = targetUri.path.ext;
+            extension = '';
+        }
+        // we want the path base for directories with the source path `foo.bar` to be generated as `foo.bar copy` and not `foo copy.bar` as we do for files
+        if (isDirectory) {
+            name = name + extension;
+            extension = '';
+        }
 
-        let index = 1;
-        let base = name + ext;
+        let base = name + extension;
+        // test if the name already contains the suffix or the suffix + index, so we don't add it again
+        const nameRegex = RegExp(`.*${FileSystemUtils.FILE_NAME_SEPARATOR}${suffix}(${FileSystemUtils.FILE_NAME_SEPARATOR}[0-9]*)?$`);
+        if (suffix && !nameRegex.test(name) && children.some(child => child.path.base === base)) {
+            name = name + FILE_NAME_SEPARATOR + suffix;
+            base = name + extension;
+        }
+        if (suffix && nameRegex.test(name)) {
+            // remove the existing index from the name, so we can generate a new one
+            name = name.replace(RegExp(`${FILE_NAME_SEPARATOR}[0-9]*$`), '');
+        }
+        let index = 0;
         while (children.some(child => child.path.base === base)) {
             index = index + 1;
-            base = name + '_' + index + ext;
+            base = name + FILE_NAME_SEPARATOR + index + extension;
         }
-        return parentUri.resolve(base);
+        return parent.resource.resolve(base);
     }
 }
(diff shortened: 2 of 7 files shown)


packages/filesystem/src/browser/file-tree/file-tree-model.ts
```diff
@@ -153,9 +153,9 @@ export class FileTreeModel extends CompressedTreeModel implements LocationServic
     async copy(source: URI, target: Readonly<FileStatNode>): Promise<URI> {
         let targetUri = target.uri.resolve(source.path.base);
         try {
-            if (source.path.toString() === target.uri.path.toString()) {
+            if (source.path.toString() === target.uri.path.toString() || source.path.toString() === targetUri.path.toString()) {
```
msujew, line 156: Suggestion: This additional check feels like an unclean solution to the problem. Instead, I would first try to identify potential parent node of the target node and then perform a check. The full method should look something like this:

```ts
async copy(source: URI, target: Readonly<FileStatNode>): Promise<URI> {
  const dirNode = target.fileStat.isFile ? target.parent : target;
  if (!FileStatNode.is(dirNode)) {
    throw new Error('Parent of file has to be a FileStatNode');
  }
  let targetUri = dirNode.uri.resolve(source.path.base);
  try {
    if (source.path.toString() === targetUri.path.toString()) {
      const parent = await this.fileService.resolve(source.parent);
      const name = source.path.name + FileSystemUtils.FILE_NAME_SEPARATOR + 'copy';
      targetUri = FileSystemUtils.generateUniqueResourceURI(source.parent, parent, name, source.path.ext);
    }
    await this.fileService.copy(source, targetUri);
  } catch (e) {
    this.messageService.error(e.message);
  }
  return targetUri;
}
```

msujew (commented):
Looking mostly good already. However, I noticed that copying a folder leads to some weird behavior. Instead of creating `<original-name> copy`, the copied folder is called `copy<original-name>` instead.

packages/filesystem/src/browser/file-tree/file-tree-model.ts
AlexandraBuzila: I applied your suggestion in the new commit. There's one change in how the `targetUri` is computed: we need to append the source base  to the `dirNode` with `uri.resolve` only for directories.

AlexandraBuzila:
> However, I noticed that copying a folder leads to some weird behavior. Instead of creating `<original-name> copy`, the copied folder is called `copy<original-name>` instead.

I could reproduce the behavior you mention for hidden folders and files, whose names begin with a `.`. The problem occurs because the uri [Path](https://github.com/eclipse-theia/theia/blob/master/packages/core/src/common/path.ts) for the resource is returning an empty name and treats the name as an extension. For example for the `.theia` folder, the `name` is `''` and the `ext` is `.theia`. I'm not sure if hidden files are a special use case that we should consider when renaming or rather a bug in Path.

packages/filesystem/src/browser/file-tree/file-tree-model.ts
msujew: Perhaps my understanding of certain file systems is wrong, but isn't the `parent` node of a file node always a directory node? In what case can `dirNode` be a file?

msujew:
@AlexandraBuzila I see, I simply used the `.theia` folder in the repo to test this. It isn't hidden, but that doesn't matter in that case anyway.

I wouldn't say it's in error in the path implementation, files like `.gitignore` are correct in having an empty `name` property and an `ext` value of `.gitignore`. I would try to align this VSCode in that regard:

1. Files are copied as `<name> copy.<ext>`
2. Files with an empty `name` are copied as `.<ext> copy`
3. Folders are always copied as `<basename> copy`

packages/filesystem/src/browser/file-tree/file-tree-model.ts
AlexandraBuzila: I'm sorry, I misspoke. What I meant to say is that when initially computing the targetUri (now line 158 in [file-tree-model.ts](https://github.com/eclipse-theia/theia/pull/10767/files#diff-a25c896256c15be09f9da9d145e14a4f92697ba18ac8612dcd2e1418be759f61R158)), we need to append the source path base to the dirNode only when the target is a _file_ (not a directory, as I initially wrote).

AlexandraBuzila:
I rebased and updated the code to handle files and folders that start with a `.`

msujew (commented):
Looking better, but copying a copied file results in `<filename> copy copy.<ext>`. In VSCode copying `<filename> copy.<ext>` leads to creating `<filename> copy 2.<ext>` instead. I believe we should align to VSCode's behavior.

AlexandraBuzila:
> copying a copied file results in `<filename> copy copy.<ext>`. In VSCode copying `<filename> copy.<ext>` leads to creating `<filename> copy 2.<ext>` instead. I believe we should align to VSCode's behavior.

Should be fixed now

packages/filesystem/src/browser/file-tree/file-tree-model.ts
msujew: Ah, I see. That makes sense 👍

msujew (commented):
Almost working correctly now. I believe this point hasn't been resolved correctly:

> Folders are always copied as `<basename> copy`

Because naming a folder `test.theia` and copying it, leads to the new file name `test copy.theia`. In VSCode this will be `test.theia copy` instead.

AlexandraBuzila:
>  I believe this point hasn't been resolved correctly:
> 
> > Folders are always copied as `<basename> copy`

You're right, I missed that. It should be fixed now. I also added some test cases.

packages/filesystem/src/common/filesystem-utils.ts
```diff
@@ -18,24 +18,46 @@ import { FileStat } from '../common/files';
...
 
     /**
      * Generate unique URI for a given parent which does not collide
      *
-     * @param parentUri the `URI` of the parent
      * @param parent the `FileStat` of the parent
-     * @param name the resource name
-     * @param ext the resource extension
+     * @param targetUri the initial URI
+     * @param suffix an optional string to append to the file name, in case of collision (e.g. `copy`)
```
msujew, line 28: Missing entry `@param` for `isDirectory`.

packages/filesystem/src/common/filesystem-utils.ts
```diff
@@ -18,24 +18,46 @@ import { FileStat } from '../common/files';
...
      */
-    export function generateUniqueResourceURI(parentUri: URI, parent: FileStat, name: string, ext: string = ''): URI {
+    export function generateUniqueResourceURI(parent: FileStat, targetUri: URI, isDirectory: boolean, suffix?: string): URI {
```
msujew, line 30: Please add a breaking change entry in the changelog for this.

msujew (commented):
Looking good now!

- [x] Copying files (with and without names) works as in VSCode
- [x] Copying folders works as in VSCode 

Please address my documentation related suggestions and I'll merge before the release!

packages/filesystem/src/common/filesystem-utils.ts
AlexandraBuzila: I updated the changelog

msujew (commented):
Oh well, I didn't notice the test failure on Windows. Some test in `navigator.spec.js` is failing on Windows. Do you mind addressing it?

If you don't have access to a Windows system to reproduce the issue, I will look into it next week.

AlexandraBuzila:
> Oh well, I didn't notice the test failure on Windows. Some test in `navigator.spec.js` is failing on Windows. Do you mind addressing it?
I can have a look at it at the beginning of next week.

msujew (approved):
Thanks, looking well now! The behavior is perfectly aligned to VSCode now:

- [x] Copying a normal file suffixes ` copy (<number>)` to its **file**name.
- [x] Copying a file without a filename, suffixes ` copy (<number>)` to its **full**name.
- [x] Copying a folder suffixes the fullname as well.
- [x] Copying a folder into itself will result in a copied sibling folder.

msujew:
@AlexandraBuzila There seem to be some conflicts (although GitHub isn't willing to show me what it is - I suspect the changelog). Do you mind resolving them? I'll merge this afterwards.

AlexandraBuzila:
> @AlexandraBuzila There seem to be some conflicts (although GitHub isn't willing to show me what it is - I suspect the changelog). Do you mind resolving them? I'll merge this afterwards.

I rebased the changes and fixed the conflict (it was indeed the changelog).

Issue #10768, opened 2022-02-17
electron: Open dialogs are not modal


AlexandraBuzila:
### Bug Description:
The Open dialog in the electron application is not modal. Multiple ones can be opened at the same time. This affects the Open File, Open Folder, Open Workspace commands.

### Steps to Reproduce:

1. Go to the File menu, select the Open File... menu entry. A dialog to select the file should open.
2. Click outside the dialog, e.g. in the main window
3. The dialog goes into the background and a new dialog can be opened via the Open File... command, while the first one is still open.

The expectation would be that the dialog is modal and remains on top of the main window until it's dismissed.

vince-fugnitto:
Closed thanks to #10769.

PR #10769, opened 2022-02-17
electron: Open dialogs are not modal (#10768)

AlexandraBuzila:
#### What it does

The Open dialog in the electron application is not modal. Multiple ones can be opened at the same time. This affects the Open File, Open Folder, Open Workspace commands. 

Pass the electron BrowserWindow to the dialog.showOpenDialog command, to make the dialog modal.

#### How to test

1. Start the electron example app
2. Go to the _File_ menu, select the _Open File..._ menu entry. A dialog to select the file should open.
3. Click outside the dialog, e.g. in the main window
4. Nothing should happen, as the dialog is modal.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
+++ b/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
@@ -49,7 +49,9 @@ export class ElectronFileDialogService extends DefaultFileDialogService {
     override async showOpenDialog(props: OpenFileDialogProps, folder?: FileStat): Promise<MaybeArray<URI> | undefined> {
         const rootNode = await this.getRootNode(folder);
         if (rootNode) {
-            const { filePaths } = await electronRemote.dialog.showOpenDialog(this.toOpenDialogOptions(rootNode.uri, props));
+            const { filePaths } = props.modal !== false ?
+                await electronRemote.dialog.showOpenDialog(electronRemote.getCurrentWindow(), this.toOpenDialogOptions(rootNode.uri, props)) :
+                await electronRemote.dialog.showOpenDialog(this.toOpenDialogOptions(rootNode.uri, props));
             if (filePaths.length === 0) {
                 return undefined;
             }
@@ -65,7 +67,9 @@ export class ElectronFileDialogService extends DefaultFileDialogService {
     override async showSaveDialog(props: SaveFileDialogProps, folder?: FileStat): Promise<URI | undefined> {
         const rootNode = await this.getRootNode(folder);
         if (rootNode) {
-            const { filePath } = await electronRemote.dialog.showSaveDialog(this.toSaveDialogOptions(rootNode.uri, props));
+            const { filePath } = props.modal !== false ?
+                await electronRemote.dialog.showSaveDialog(electronRemote.getCurrentWindow(), this.toSaveDialogOptions(rootNode.uri, props)) :
+                await electronRemote.dialog.showSaveDialog(this.toSaveDialogOptions(rootNode.uri, props));
             if (!filePath) {
                 return undefined;
             }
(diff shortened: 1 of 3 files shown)


vince-fugnitto (commented):
I think we might want to take it a step further and also apply it for save dialogs?

In addition, what do you think of supporting an optional `modal` prop (true by default if you prefer) for `FileDialogProps` to control the behavior of whether the dialog should be a modal or not?

AlexandraBuzila:
> I think we might want to take it a step further and also apply it for save dialogs?
> 
> In addition, what do you think of supporting an optional `modal` prop (true by default if you prefer) for `FileDialogProps` to control the behavior of whether the dialog should be a modal or not?

Makes sense, thank you for the review! I amended the commit with the changes you suggested.

packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
```diff
@@ -49,7 +49,11 @@ export class ElectronFileDialogService extends DefaultFileDialogService {
     override async showOpenDialog(props: OpenFileDialogProps, folder?: FileStat): Promise<MaybeArray<URI> | undefined> {
         const rootNode = await this.getRootNode(folder);
         if (rootNode) {
-            const { filePaths } = await electronRemote.dialog.showOpenDialog(this.toOpenDialogOptions(rootNode.uri, props));
+            const electronWindow = electronRemote.getCurrentWindow();
+            const isModal = props.modal !== false;
+            const { filePaths } = isModal ?
+                await electronRemote.dialog.showOpenDialog(electronWindow, this.toOpenDialogOptions(rootNode.uri, props)) :
+                await electronRemote.dialog.showOpenDialog(this.toOpenDialogOptions(rootNode.uri, props));
```
vince-fugnitto, line 56: minor: we can simplify the expression to only get the current window if necessary:

```suggestion
            const { filePaths } = props.modal !== false ?
                await electronRemote.dialog.showOpenDialog(electronRemote.getCurrentWindow(), this.toOpenDialogOptions(rootNode.uri, props)) :
                await electronRemote.dialog.showOpenDialog(this.toOpenDialogOptions(rootNode.uri, props));
```

packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
```diff
@@ -65,7 +69,10 @@ export class ElectronFileDialogService extends DefaultFileDialogService {
     override async showSaveDialog(props: SaveFileDialogProps, folder?: FileStat): Promise<URI | undefined> {
         const rootNode = await this.getRootNode(folder);
         if (rootNode) {
-            const { filePath } = await electronRemote.dialog.showSaveDialog(this.toSaveDialogOptions(rootNode.uri, props));
+            const electronWindow = electronRemote.getCurrentWindow();
+            const isModal = props.modal !== false;
+            const { filePath } = isModal ? await electronRemote.dialog.showSaveDialog(electronWindow, this.toSaveDialogOptions(rootNode.uri, props)) :
+                await electronRemote.dialog.showSaveDialog(this.toSaveDialogOptions(rootNode.uri, props));
```
vince-fugnitto, line 75: same is true as above comment, with additional formatting:

```suggestion
            const { filePath } = props.modal !== false ? 
                await electronRemote.dialog.showSaveDialog(electronRemote.getCurrentWindow(), this.toSaveDialogOptions(rootNode.uri, props)) :
                await electronRemote.dialog.showSaveDialog(this.toSaveDialogOptions(rootNode.uri, props));
```

AlexandraBuzila:
I applied your suggestions, they look good.

vince-fugnitto:
@msujew any objection? :)

msujew (approved):
Looks good to me as well 👍

PR #10782, opened 2022-02-21
Add machine translations for 12 languages

msujew:
#### What it does

Adds support for 12 languages supported by official VSCode language packs. There are some still missing (Turkish, Korean, traditional Chinese), since these aren't translatable by DeepL yet.

Also fixes another round of issues surrounding missing translations. Also makes the `nls-extract` and `nls-translate` commands a bit nicer to work with.

While I'm sure some translations are erroneous or at least sound weird to native speakers, I would like to fix this in separate PRs, if necessary.

#### How to test

1. Install a language pack of your choice (except for the missing ones mentioned above)
2. Use the `Configure Display Language` command to change to that language
3. Inspect the app: Except for extensions, everything should be translated in the app

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
The changes look good to me and it works well :+1: I confirmed with the french language-pack that theia-specific labels are translated. 

I did notice however that we might need to scan the repo for other labels in views which render content but does not attempt to localize, for instance:

https://github.com/eclipse-theia/theia/blob/363e5216dac77e4c3fa419f201cc161955084c43/packages/scm-extra/src/browser/history/scm-history-widget.tsx#L362-L371

The view will not translate `repository`, `files` or `commits`:

image

I assume there is likely other cases like the following scattered across the codebase, but it is not related to this specific pull-request.

dev-packages/localization-manager/src/deepl-api.ts
```diff
@@ -40,9 +41,38 @@ export async function deepl(
...
 
+/**
+ * Coerces the target language into a form expected by Deepl.
+ *
+ * Currently only replaces `ZH-CN` with `ZH`
+ */
+function coerceLanguage(parameters: DeeplParameters): void {
+    if (parameters.target_lang === 'ZH-CN') {
+        parameters.target_lang = 'ZH';
+    }
```
colin-grant-work, line 58: Does DeepL allow other `ZH` variants (`ZH-TW`, e.g.), or does it expect all `ZH` to be only `ZH`?

Issue #10005, opened 2021-08-31
Implement `CustomDocumentOpenContext`


tsmaeder:
### Feature Description:
Implement the functionality in `CustomDocumentOpenContext`. While we already have the `backupId` field defined, Theia is never passing it to any CustomEditorProvider. The `untitledDocumentData` field is not present in `theia.d.ts`, but since it's optional in VS Code, any plugin written against `vscode.dt.ts` will work anyway.

tsmaeder:
To achieve source compatibility, it would be enough to just declare the field in `theia.d.ts`.

vince-fugnitto:
@tsmaeder what should the scope be for initial support of the feature? If we are to add support for custom-editor backups it'd be quite involving, especially for a beginner (`good first issue` label).

tsmaeder:
> To achieve source compatibility, it would be enough to just declare the field in theia.d.ts

No real implementation needed.

PR #10784, opened 2022-02-22
Add missing property to CustomDocumentOpenContext

planger:
#### What it does

VSCode defines the optional property `untitledDocumentData` in `CustomDocumentOpenContext`. In `theia.d.ts`, this property is however missing. As it is optional and is not used from the plugin main implementation, this change is merely achieving source compatibility, but prepares its usage in the ext implementation of the plugin API.

Fixes https://github.com/eclipse-theia/theia/issues/10005

#### How to test
As this change doesn't introduce new functionality per se, we just need to make sure existing functionality, especially with respect to custom editors contributed by VS Code extensions still work. Therefore, I tested with `cat-customs.cat-customs` and `vscode.simple-browser`, which both seem to be still working fine.

image

#### Review checklist

#### Reminder for reviewers


msujew (approved):
Looking good. I can confirm that this change aligns the API to VSCode and introduces no regressions on existing custom document providers.

PR #10796, opened 2022-02-24
fix license headers

paul-marechal:
Do a pass over recently merged files to use the new license header format.


vince-fugnitto (approved):
LGTM :rocket:

PR #10801, opened 2022-02-24
Improve key binding rendering

### LINKED ISSUE #5377: [keybindings] display keys instead of string representation in the keybindings widget
**Description**

An improvement to the `keybindings` widget is to display the actual keys instead of their string representation. 

_Example:_ <kbd>cmd</kbd>+<kbd>A</kbd> would become

<div align='center'>

Screen Shot 2019-06-06 at 8 30 30 AM

</div>

_Other examples:_

<div align='center'>

Screen Shot 2019-06-06 at 8 38 52 AM

</div>

### LINKED ISSUE #10789: Display 'ctrlcmd' keybindings as 'ctrl' or 'cmd' depending on OS
### Feature Description:
Some keyboard shortcuts are set to `ctrlcmd`, which equals Cmd on Mac and Ctrl on other systems.

MicrosoftTeams-image

The meaning is not completely obvious, so the display could be improved. Instead of displaying `ctrlcmd` as above, Theia should detect whether the system is Mac or other, and display `cmd` or `ctrl` as suited to the system.

colin-grant-work:
#### What it does

Fixes #10789 and fixes #5377 by using utilities from the keybinding registry to determine the shape of the rendered item. It also refactors the code to frontload the handling of fuzzy matching and not rebuild the whole list of keybinding candidates on every search. This entailed modifying the interfaces in the file so that the data maintained on the widget was a better input into the render cycle.

MacOS:

image

Ubuntu:

image

#### How to test

1. Open the keybinding widget.
2. Observe that the keys are rendered nicely (symbols on MacOS, capitalized keys without any `ctrlcmd` elsewhere). 
3. Search for keybindings.
4. Observe that the results are as you expect, given our fuzzy implementation. In particular:
    - Observe that keys are only matched on a perfect matched
    - Observe that both `+` and whitespace can be used as term separators for keys

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/keybinding.ts
+++ b/packages/core/src/browser/keybinding.ts
@@ -337,23 +337,28 @@ export class KeybindingRegistry {
      * @param asciiOnly if `true`, no special characters will be substituted into the string returned. Ensures correct keyboard shortcuts in Electron menus.
      */
     acceleratorForKeyCode(keyCode: KeyCode, separator: string = ' ', asciiOnly = false): string {
+        return this.componentsForKeyCode(keyCode, asciiOnly).join(separator);
+    }
+
+    componentsForKeyCode(keyCode: KeyCode, asciiOnly = false): string[] {
         const keyCodeResult = [];
+        const useSymbols = isOSX && !asciiOnly;
         if (keyCode.meta && isOSX) {
-            keyCodeResult.push('Cmd');
+            keyCodeResult.push(useSymbols ? '⌘' : 'Cmd');
         }
         if (keyCode.ctrl) {
-            keyCodeResult.push('Ctrl');
+            keyCodeResult.push(useSymbols ? '⌃' : 'Ctrl');
         }
         if (keyCode.alt) {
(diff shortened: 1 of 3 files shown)


vince-fugnitto (commented):
I confirmed that the feature works for the most part, especially on macos :+1: 
- [x] confirmed that on macos the key symbols are displayed but not their string values
- [x] confirmed that updating keybinding will properly render afterwards
- [x] confirmed that fuzzy matching still works

---

I did notice an issue on Linux when using electron (the browser seems fine), the <kbd>Escape</kbd> keybinding rendering is broken:

keybinding-error

colin-grant-work:
@vince-fugnitto, good catch re: special characters in Electron. Esc, Delete, and Backspace appear to have special unicode characters that come from `nativekeymap`. I've added special handling in the method that generates the text representations for those values.

vince-fugnitto (approved):
I confirmed that my previous feedback has been resolved.

PR #10802, opened 2022-02-25
doc: split up `changelog` and archive old content

vince-fugnitto:
#### What it does

When preparing for yesterday's release and doing the bulk [changelog](https://github.com/eclipse-theia/theia/pull/10798) update I had noticed that editing the file takes a considerable amount of time, and some extensions fail to properly activate or work when doing so (using vscode). The idea is to archive some old content of the changelog into their respective files under `docs/` to make the editing experience much better, especially for new contributions.

The changes include:
- adding a new `changelogs/` subdirectory under `docs/` to archive old changelog content.
- adding a link to the directory from the existing changelog.
- splitting up content from the changelog into their respective files.

#### How to test

- confirm that the changes look correct.
- **note**: once merged the link to https://github.com/eclipse-theia/theia/tree/master/doc/changelogs/ will work.

#### Review checklist

#### Reminder for reviewers


colin-grant-work (approved):
I think this is a great idea, making it easier to maintain the file and easier to track what happened in particular periods. It's still easy to track particular package threads using search capabilities. To further facilitate that, should the current changelog be in the same folder as the archive?

vince-fugnitto:
> I think this is a great idea, making it easier to maintain the file and easier to track what happened in particular periods. It's still easy to track particular package threads using search capabilities. To further facilitate that, should the current changelog be in the same folder as the archive?

We could if you think it adds value to have them together, I didn't want to pollute the root with multiple changelogs however :) The only thing I wanted to avoid was breaking links to the original changelog.

colin-grant-work:
> We could if you think it adds value to have them together, I didn't want to pollute the root with multiple changelogs however :) The only thing I wanted to avoid was breaking links to the original changelog.

I think the way you've set it up is a reasonable compromise between those two objectives. With the link from the main changelog to the others, people shouldn't have a hard time finding their way around.

JonasHelming (approved):
Very good idea and I agree to the result of the discussion above.

msujew (approved):
The change looks good to me as well 👍

PR #10781, opened 2022-02-21
Protoype of a faster RPC protocol

tsmaeder:
Signed-off-by: Thomas Mäder <tmader@redhat.com>

#### What it does
This PR is purely for legal reasons: I had prototyped the code in my personal repo and now need to officially contribute it to the project so that the good folks at EclipseSource can use it in their own contribution to Theia.

PR #10809, opened 2022-02-28
[WIP] Integrate new message-rpc prototype into core messaging API (extensions)

### LINKED ISSUE #10684: Improve performance of JSON-RPC communication
Theia relies heavily on JSON-RPC for exchanging messages between frontend and backend processes. Messages are often routed multiple times through different processes (Theia frontend, Theia backend, plugin processes) therefore it is essential
that the communication layer (i.e. JSON-RPC protocols) is implemented in an efficient and performant fashion.

Currently there is a lot of momentum in this topic to significantly improve the communication layer in both architecture and performance. We collected and analyzed the currently available information which is distributed over numerous issues, pull requests and POC branches. Based on that information we distilled a 5 step development plan to coordinate the overall changes.

## Analysis

Recently it has been reported that the performance of the Filesystem API for Theia plugins is significantly worse than the performance of its VS Code counterpart (https://github.com/eclipse-theia/theia/issues/9514). According to the issue creator, in this particular case the loading times increase by the factor 100.  After investigations led by @tsmaeder it was concluded that the performance issues are not directly related to the Filesystem API. The performance hit really occurs in the JSON-RPC communication layer when encoding large chunks of binary data (i.e. files).

Currently Theia has two different JSON-RPC message protocols in place. For Theia backend-frontend services `vscode-ws-jsonrpc` is used and for plugin communication a custom protocol implementation is used.  Both encode the message data as strings and additionally wrap them into dedicated message objects. This is problematic because encoding binary data to strings is very inefficient, for example the message size alone already increases by a factor of 5. In addition, messages are often wrapped and encoded multiple times on the way to their destination.

To increase performance the json-rpc protocols need to be refactored to:
- Directly transfer binary data over websocket instead of always using encoded strings
- Use a  0-copy/write-through approach. Instead of wrapping messages at lower levels we could directly write the encoded values to a message buffer that is managed by the lower level protocol.

The RPC protocols have been recently refactored to be independent of plugin concerns (https://github.com/eclipse-theia/theia/pull/8972) which means the code base is now in a good state for further improvements to the RPC protocol infrastructure.

In this process we should probably get rid of any usage of `vscode-ws-jsonrpc`. This library is a custom fork of `vscode-jsonrpc` to add web socket support and is no longer maintained. It has been concluded that the way to go forward is to replace it with an in-repo alternative instead of consuming an updated version or an alternative library (see also https://github.com/eclipse-theia/theia/pull/10265).

@tsmaeder has already worked on a first attempt to refactor the Theia Messaging RPC protocol to a write-through RPC framework (https://github.com/tsmaeder/message-rpc). This could serve as a base for any future efforts.

Note that another recent effort has been made to improve the communication layer by switching from simple websockets to `socket.io` (https://github.com/eclipse-theia/theia/pull/10514). However, strictly speaking this change only touches the transportation layer and does not directly affect the json-rpc communication layer on top.

## Development plan

Based on the analysis we have identified the following tasks to increase the overall performance of messaging via json-rpc.

- [x] **Setup infrastructure to measure the performance RPC calls** 
Setup test cases that send large chunks of binary data via websocket (e.g. reading a file from the filesystem) for all different scenarios (Theia Extension (browser, electron), Theia plugin (browser,electron)) and record execution times. This could probably be integrated as part into the CI Performance Test Suite (https://github.com/eclipse-theia/theia/issues/10443).

- [x] **Prepare  0-copy RPC framework prototype for integration into Theia**
 Review the rpc-framework proposal of @tsmaeder (https://github.com/tsmaeder/message-rpc) and clean up any potential rough edges. In addition, the new framework should be properly unit tested.

- [x] **Integrate new PRC framework with Theia message API and replace `vscode-ws-jsonrpc`**
Once the new rpc framework is ready for contribution it has to be integrated into the existing Theia messaging API and should serve as a replacement for `vscode-ws-jsonrpc`. Once this task is finished there should be no further dependency to `vscode-ws-jsonrpc` in the core framework.

- [ ] **Align plugin RPC protocol with Theia messaging RPC protocol**
Currently Theia uses a different RPC protocol for plugin main/ext remote calls. It would be easier if we could use the same infrastructure (i.e. the new JSON-RPC protocol) here.

- [ ] **Evaluate core framework to identify inefficient transfer of binary data**
Depending on the situations (local/remote processes, etc.) it should be evaluated what the most efficient way to transfer binary values is. For example, ChildProcess.send() between the back end and the plugin host seems to be wasteful because it internally converts the data to string.

tortmayr:
#### What it does

- Refactors and improves the prototype of a faster JSON-RPC protocol intially contributed by @tsmaeder  (See also #10781).
  The encoding approach used in the initial POC has some performance drawbacks when encoding plain JSON objects. We refactored the protocol to improve the performance for JSON objects whilst maintaining the excellent performance for encoding objects that contain binary data. In addition, the error handling and lifecycle management is  improved.
- Integrates the new message RPC protocol into Theia's core messaging API and therfore replaces the previously used `vscode-ws-jsonrpc` protocol. This has major impacts and the Messaging API as we no longer expose a  `Connection` object (which was provided by vscode-ws-jsonrpc) and directly rely on the generic `Channel` implementation instead. 
- Adapt the connection providers for the different connection types (Websocket, IPC, custom connections like TerminalWidget-> underlying process) to efficiently use the new RPC protocol. Efficiently in this case means that for some providers adoptions are necessary so that they can send messages as binary data instead of strings.

A performance test suite has been executed to mesaure  the improvements gained by this change (See https://github.com/eclipse-theia/theia/issues/10684#issuecomment-1086160205 for more details).

Note: This is a WIP PR that is not ready to be reviewed yet and contains a lot of temporary hacks and workarounds (e.g. commenting out of *.spec.ts* files that have compilation errors). It is currently only working in the browser-context as the electron-ipc connection provider is not fully adopted to the new protocol yet.
 It's intention is to present the current state of #10684 to interested parties.

Closes #10684

#### Review checklist

#### Reminder for reviewers


packages/core/src/common/message-rpc/websocket-client-channel.ts
```diff
@@ -0,0 +1,229 @@
...
+        };
+        this.socket = socket;
+        window.addEventListener('offline', () => this.tryReconnect());
+        window.addEventListener('online', () => this.tryReconnect());
+    }
+
+    getWriteBuffer(): WriteBuffer {
+        const result = new ArrayBufferWriteBuffer();
+        const httpUrl = this.createHttpWebSocketUrl('/services');
+        if (this.useHttpFallback) {
```
paul-marechal, line 107: The whole HTTP fallback shouldn't be required anymore?

We're using `socket.io` which should handle this natively.

tortmayr: Good catch. Yes this file was a remainder from the initial message-rpc PoC code drop. It has been removed as part of the update that I pushed.

tsmaeder:
@tortmayr still WIP?

tortmayr:
@tsmaeder Yes, still a little bit of code cleanup to do and testing to ensure that everything works as expected electron. 
We will provide an update soon that should make the PR ready-to-review

paul-marechal:
@tortmayr do you mind joining the [technical meeting on Wednesday](https://github.com/eclipse-theia/theia/wiki/Technical-Meeting) in order to sync up our respective WIPs?

I have been re-writing a few core APIs related to messaging, and I'd like to go over some of those changes with you to see if anything needs to be changed either here or on my side.

tortmayr:
> @tortmayr do you mind joining the [technical meeting on Wednesday](https://github.com/eclipse-theia/theia/wiki/Technical-Meeting) in order to sync up our respective WIPs?
> 
> I have been re-writing a few core APIs related to messaging, and I'd like to go over some of those changes with you to see if anything needs to be changed either here or on my side.
No at all, I'm happy to join the technical meeting on Wednesday. Could provide me with a link to your WIP so that I can have a look at it beforehand?

paul-marechal:
@tortmayr here's the [_very_ wip branch](https://github.com/eclipse-theia/theia/tree/mp/electron-ipc-refact-rebased). It builds but doesn't run yet.

I rewrote our proxies to hook themselves to some `RpcConnection` interface which abstracts away what it takes to do a RPC call:

https://github.com/eclipse-theia/theia/commit/e8b5605b7d670e12d9949131ed1766b75537f7c9#diff-0e651b8d4ece117f58140576d68cb6a86b4ed1cb2c52057850b8dae52045c935R28-R34

Whenever you do `proxy.someMethod(arg1, arg2)` it will actually go through `rpcConnection.sendRequest('someMethod', arg1, arg2)`. I was wondering if your new RPC implementation could fit in that abstraction, or if something prevents doing so.

I want to go over some other details during the call and see if some of the other abstractions will benefit you as well or not.

tortmayr:
This WIP PR has served its purpose as  I now have opened the actual PR: https://github.com/eclipse-theia/theia/pull/11011

PR #10817, opened 2022-02-28
Add Pinned Tabs

### LINKED ISSUE #9326: Pin editor in tab-bar

colin-grant-work:
#### What it does

Commands for pinning and unpinning tabs. Fixes #9326. Closes #9614.

This is based on #9614 with a bug fix and a switch to codicons for pinned tab icons.

#### How to test

1. Use the tabbar context menu or command palette to pin and unpin tabs. 
   > In the main or bottom areas, they should receive a pin icon.
   If dirty, an editor should receive pin-with-dot icon.
2. Try to close pinned editors. They should not close, but should convert to unpinned editors.
3. Reload the page / restart the application. Pinned state should be restored.

#### Review checklist

#### Reminder for reviewers


packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -241,6 +242,16 @@ export namespace CommonCommands {
...
+    export const PIN_TAB = Command.toDefaultLocalizedCommand({
+        id: 'workbench.action.pinEditor',
+        category: VIEW_CATEGORY,
+        label: 'Pin Tab'
+    });
+    export const UNPIN_TAB = Command.toDefaultLocalizedCommand({
+        id: 'workbench.action.unpinEditor',
+        category: VIEW_CATEGORY,
+        label: 'Unpin Tab'
+    });
```
msujew, line 254: The eslint rule doesn't really work for `toDefaultLocalizedCommand` (I should fix this perhaps)

`Pin` and `Unpin` are the labels used by VSCode and can both be found in the `nls.metadata.json` file.

msujew (commented):
I noticed some minor issues in this feature:

1. Pinning a side bar tab prevents it from closing. Moving this pinned tab to the main area leaves it in an invalid state. The usual close icon is displayed and works as expected. However, the context menu entry for `Pin` is disabled and the one for `Unpin` is enabled.
2. Pinning a focused tab removes the focus from the tab. VSCode refocuses the target tab of the context menu after it closes. This one might not have to do with your PR and is probably more of a general issue of how context menus and tab focus work.

packages/core/src/browser/shell/tab-bars.ts
```diff
@@ -162,7 +163,10 @@ export class TabBarRenderer extends TabBar.Renderer {
                 this.renderLabel(data, isInSidePanel),
                 this.renderBadge(data, isInSidePanel)
             ),
-            this.renderCloseIcon(data)
+            h.div({
+                className: 'p-TabBar-tabCloseIcon',
+                onclick: this.handleCloseClickEvent
+            })
```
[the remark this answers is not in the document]
colin-grant-work: Thanks. I think the original PR predated your work in that area.

packages/core/src/browser/common-frontend-contribution.ts
colin-grant-work: I did wonder about whether the linting would catch it for the other NLS utilities. I'll update the text.

colin-grant-work:
@msujew, I think the problem was probably code in `ViewContainer`, and I've adjusted that, and adjusted the code checking enablement of the commands to consistently check the `closeable` field, as well, so that `closeable: true` is always treated as `pinned: false` even if the classname is present.

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -636,6 +658,16 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
             label: nls.localizeByDefault('Toggle Menu Bar'),
             order: '0'
         });
+        registry.registerMenuAction(SHELL_TABBAR_CONTEXT_PIN, {
+            commandId: CommonCommands.PIN_TAB.id,
+            label: 'Pin',
```
msujew, line 663:
```suggestion
            label: nls.localizeByDefault('Pin'),
```

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -636,6 +658,16 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
...
+            order: '7'
+        });
+        registry.registerMenuAction(SHELL_TABBAR_CONTEXT_PIN, {
+            commandId: CommonCommands.UNPIN_TAB.id,
+            label: 'Unpin',
```
msujew, line 668:
```suggestion
            label: nls.localizeByDefault('Unpin'),
```

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -996,6 +1048,16 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
...
                 keybinding: 'ctrlcmd+k ctrlcmd+t'
+            },
+            {
+                command: CommonCommands.PIN_TAB.id,
+                keybinding: 'ctrlcmd+k shift+enter',
+                when: '!activeWidgetIsPinned'
+            },
+            {
+                command: CommonCommands.UNPIN_TAB.id,
+                keybinding: 'ctrlcmd+k shift+enter',
```
msujew, line 1059: I noticed that pinning does work using the specified keyboard shortcut. However, unpinning does not work for some reason.

packages/core/src/browser/common-frontend-contribution.ts
msujew: I actually just noticed that VSCode is using `Pin editor`/`Unpin editor`. Since we do not only allow pinning on editors, we should maybe introduce new translations here? Or is `Pin editor` accurate enough for the command label?

colin-grant-work: I think editors are going to be the most common targets, in any case, so I'll use '(un)pin editor' for now.

packages/core/src/browser/common-frontend-contribution.ts
colin-grant-work: It turns out this was also a consequence of the 'editor' vs. 'widget' question. An earlier version of the PR had either had two context keys or only `activeWidgetIsPinned` and it was switched to `editor` to agree with VSCode, but the keybindings still referred to `widget`.

colin-grant-work:
> Alright, I noticed a minor issue with the keyboard shortcut: When an editor is unfocused (for example by clicking on the tab-bar), using the shortcut to pin works, but using it to unpin does not. It works in every other case that I've tested. Since this is quite minor, I will approve this PR nonetheless, as everything else is working.

I appreciate the thorough testing! I will check on this scenario.

vince-fugnitto (commented):
I confirmed the following :+1: 

- [x] pinning/unpinning works
- [x] pinning/unpinning tabs in the main-area properly decorate the tab with the pin icon
- [x] pinning a editor-preview works, the editor becomes persistent
- [x] pinning dirty editors works, the tab is decorated with the dirty pin icon 
- [x] pinning tabs in the sidepanels work well
- [x] pinned tabs are uncloseable unless unpinned first
- [x] pinned tabs are persistent on reloads
- [x] pinning with the keyboard shortcut works well
- [ ] unpinning with the keyboard shortcut does not seem to work for some widgets (ex: reproducible with search-in-workspace) (likely what Mark noticed too)

colin-grant-work:
@msujew, @vince-fugnitto, thanks again for the thorough reviews. I believe I tracked down the problem with unpinning unfocused widgets: the context key setting and the command enablement were looking at different conditions. I have adjusted the code so that the context key is always looking at the same widget as the command will act on.

vince-fugnitto (approved):
I confirmed that the outstanding issue regarding the keyboard shortcut has been fixed, the last issue in https://github.com/eclipse-theia/theia/pull/10817#pullrequestreview-902060656 now works 👍

msujew (approved):
Thanks Colin, that last change seems to have done it 👍

Issue #10828, opened 2022-03-02
Uninstall of extensions does not refresh the view due to logging error


luettmaSICKAG:
### Bug Description:
When trying to uninstall an extension, it is still listed to the user in the extension view until the next reload.
I traced it down to this line where `location` is not defined:
https://github.com/eclipse-theia/theia/blob/v1.23.0/packages/plugin-ext/src/hosted/node/hosted-plugin-deployer-handler.ts#L181

Log output:
```
root ERROR Request undeploy failed with error: location is not defined ReferenceError: location is not defined
    at HostedPluginDeployerHandler.undeployPlugin (C:\development\MY_THEIA_PROJECT_DIRECTORY\node_modules\@theia\plugin-ext\lib\hosted\node\hosted-plugin-deployer-handler.js:164:67)
    at async PluginDeployerImpl.undeploy (C:\development\MY_THEIA_PROJECT_DIRECTORY\node_modules\@theia\plugin-ext\lib\main\node\plugin-deployer-impl.js:95:13)
    at async JsonRpcProxyFactory.onRequest (C:\development\MY_THEIA_PROJECT_DIRECTORY\node_modules\@theia\core\lib\common\messaging\proxy-factory.js:127:24)
```

### Steps to Reproduce:

1. Install any extension
2. Click "Uninstall" in the extensions view
→ Extension will be uninstalled, but still displayed (maybe other side effects)
3. Reload the window
→ Extension finally disappeared

### Additional Information

- Operating System: Windows 10
- Theia Version: 1.23.0 - Electron

vince-fugnitto:
@luettmaSICKAG thank you for reporting the issue, I can confirm the problem on my side.

@cdamus it looks like the bug was introduced by https://github.com/eclipse-theia/theia/pull/10407.

PR #10829, opened 2022-03-02
Fix undefined reference error when uninstalling extensions

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/10828

The previous code referenced the global variable `location`, which was `undefined`. The new code references the loop variable.

#### How to test

1. Install an extension and uninstall it directly after that
2. The UI should refresh
3. The backend should have logged a message about the uninstalled location

#### Review checklist

#### Reminder for reviewers


cdamus (approved):
Thanks, @msujew! Best to keep the code simpler, after all.

vince-fugnitto (approved):
I confirm that the changes work well :+1: 
- [x] undeploying extensions (ex: vsx-registry view) works well.
- [x] the logs regarding undeployment work well, and the approach is simpler.

PR #10833, opened 2022-03-02
vsx-registry: prevent search with no query present

vince-fugnitto:
#### What it does

The pull-request prevents a search from the vsx-registry view to open-vsx when no query is present.
Previously, a search would happen when no query is present which caused us to call the api and it would return the list of popular extensions. This change should effectively prevent extraneous searches from happening like an initial search when a user types, and when the search term is cleared.

https://user-images.githubusercontent.com/40359487/156428739-76f8f7ca-c006-4628-bfc5-f600c398d147.mp4

#### How to test

1. start the application, and open the extensions-view
2. perform searches - no intermediate results should appear unrelated to the query, listing the popular extensions
3. clearing the search should not cause an api call 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/vsx-registry/src/browser/vsx-extensions-model.ts
+++ b/packages/vsx-registry/src/browser/vsx-extensions-model.ts
@@ -166,12 +166,16 @@ export class VSXExtensionsModel {
     }, 500);
     protected doUpdateSearchResult(param: VSXSearchParam, token: CancellationToken): Promise<void> {
         return this.doChange(async () => {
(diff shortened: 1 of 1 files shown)


colin-grant-work (approved):
Looks good to me. Reduces calls to Open VSX and ensures that old results are cleared when the user clears the search bar.

msujew (approved):
Looking good to me as well! 👍

PR #10840, opened 2022-03-04
Align "Add Folder" visibility with VSCode

colin-grant-work:
#### What it does

This PR aligns the visibility of the context menu command to add a folder to a workspace with the corresponding command in VSCode.

#### How to test

On `master`, the command is visible if:
 - You right-click in empty space in the Navigator widget in an unsaved single-root workspace.
 - You right-click in on an existing root in the Navigator widget in a saved or multi-root workspace.

With these changes the command is visible if:
 - You right-click in empty space in the navigator widget in any workspace.
> On `master`, there may not be any empty space in the Navigator widget. This PR adds a margin to a React virtualized scroll container to ensure that there is always some empty space at the bottom of the list, as there is in VSCode.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/style/tree.css
+++ b/packages/core/src/browser/style/tree.css
@@ -42,6 +42,10 @@
     border: none;
 }
 
+.theia-TreeContainer .ReactVirtualized__Grid__innerScrollContainer {
+    margin-bottom: calc(var(--theia-ui-padding) * 3);
+}
+
 .theia-TreeContainer {
     height: 100%;
 }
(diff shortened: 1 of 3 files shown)


packages/core/src/browser/style/tree.css
```diff
@@ -42,6 +42,10 @@
     border: none;
 }
 
+.theia-TreeContainer .ReactVirtualized__Grid__innerScrollContainer {
+    margin-bottom: calc(var(--theia-ui-padding) * 3);
+}
```
vince-fugnitto, line 47: I confirmed that the extra space works well, do we want to add it to all trees or just the explorer?

vince-fugnitto (commented):
I confirmed that:
- [x] the command is available from the _more_ toolbar item
- [x] the command is available from the empty space in the explorer
- [x] the command from the command palette works well, and is properly enabled

packages/core/src/browser/style/tree.css
colin-grant-work: I think we should have it for all trees, since all trees have (by default, anyway) listeners for clicks targeting their nodes but no particular tree node, and that is only guaranteed to work if we ensure that all trees have some empty space. But I don't feel very strongly, since the Navigator is the only place I'm certain it matters.

vince-fugnitto (approved):
Confirming approval :+1:

PR #10848, opened 2022-03-06
Add iframe allow attribute fow access to web-* APIs

thegecko:

#### What it does

Adds an `allow` attribute to the webview iframe to allow contributed frontend webviews to request access to web-* APIs (webusb, webhid and webserial)

Similar to this PR in VS Code: https://github.com/microsoft/vscode/pull/117786

#### How to test

Create a VS Code web extension or frontend theia plugin which allows the user to request a device. e.g.:

```typescript
const device = await navigator.usb.requestDevice({ filters: [] });
```

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/webview/webview.ts
+++ b/packages/plugin-ext/src/main/browser/webview/webview.ts
@@ -44,6 +44,7 @@ import { OutputChannelManager } from '@theia/output/lib/browser/output-channel';
 import { WebviewPreferences } from './webview-preferences';
 import { WebviewResourceCache } from './webview-resource-cache';
 import { Endpoint } from '@theia/core/lib/browser/endpoint';
+import { isFirefox } from '@theia/core/lib/browser/browser';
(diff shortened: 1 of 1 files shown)


vince-fugnitto (commented):
I don't have any strong opinions on the change, I did try to follow as to why vscode was not eager yet to allow the changes on their side.

I'm fine with being more permissive than them, I just would like to have feedback for others if they have any concerns.

msujew (approved):
I agree with @vince-fugnitto here. As far as I understand the argument from the vscode team, they don't want users to be forced into using webviews to access `usb` and similar APIs. Since users of Theia usually have quite a lot of choices on how they want to implement this or similar features, I don't see any reason why we should prevent `web-*` API access from iframes.

With that being said, the code looks good to me 👍

thegecko:
Thanks both, I'm still working on the VS Code side. There seemed to eb a misunderstanding that a webview was required to access devices when in fact it's only needed to allow the user to authorise this access.

Issue #10815, opened 2022-02-28
ViewContainerParts have same styling as tree nodes (almost)


colin-grant-work:
### Bug Description:

If multiple tree views are open in the same view container and collapsed, there is almost nothing to distinguish them from a group of tree nodes. The `ViewContainerPart` header styling should be more distinct, including at least a divider and font style change.

Theia: 

![image](https://user-images.githubusercontent.com/62660806/156043020-41b488c3-2e1c-46fa-8d5c-5c3e8157a2eb.png)

VSCode (note dividers and bold font style):

![image](https://user-images.githubusercontent.com/62660806/156042661-f14fbe48-065d-41e4-b7b2-5e0ee6ffb552.png)

### Steps to Reproduce:

1. Open the application with Theia as workspace
2. Open the Navigator view
3. Collapse all views

### Additional Information

- Operating System: MacOS (Electron)
- Theia Version: `master`(ebd0fb8)

PR #10854, opened 2022-03-07
Align view container styling with VSCode

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/10815

#### How to test

1. Compare the view container styling with VSCode (in particular title font weight and border styling)
3. Try this with different themes.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/style/index.css
+++ b/packages/core/src/browser/style/index.css
@@ -143,7 +143,7 @@ blockquote {
 .theia-header {
   text-transform: uppercase;
   font-size: var(--theia-ui-font-size0);
-  font-weight: 450;
+  font-weight: 700;
 }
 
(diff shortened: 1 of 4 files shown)


msujew:
@vince-fugnitto You're right, I can reproduce that as well. This cannot be easily handled just with CSS, because the `:nth-child(of <selector>` syntax is only supported by Safari :/

I'll see how this can be addressed.

msujew:
@vince-fugnitto The final solution is not as clean as previously, but there is no nice way to filter for the first visible view container part using purely CSS.

vince-fugnitto (approved):
The changes work well for me, and I confirmed the bug I noticed earlier is fixed :+1:

Issue #10528, opened 2021-12-09
Deprecate Call Hierarchy View and Types


colin-grant-work:
### Feature Description:

Our call hierarchy package is in a strange state. Since Theia's non-plugin language capabilities were removed, the package exists almost exclusively to interact with plugins, but its design complicates that interaction in a couple of ways:

- It declares its own `CallHierarchyItem` type. That type corresponds closely to VSCode's types but doesn't distinguish between ingoing and outgoing calls, as VSCode's types do. The Theia `CallHierarchyItem` type introduces the need for type conversion and also introduces information loss.
- We contribute a view and a command to open that view, but the view's functionality is basically a strict subset of the functionality available from the `vscode-references-view` plugin, which is part of our default build: our view only supports incoming calls, not outgoing.

The upshot is: you can't really use call hierarchy in Theia without plugins, and if you have plugins, you're better off using `vscode-references-view`. Given that, deprecating and removing the Theia call hierarchy view and `CallHierarchyItem` type would simplify our code without reducing usability. That would leave the `CallHierarchyService` in the package, or it could be moved to the `plugin-ext` package where the providers are registered.

vince-fugnitto:
I believe we should remove both the `@theia/callhierarchy` and `@theia/typehierarchy` (thanks to https://github.com/eclipse-theia/theia/pull/11694) extensions. The `@theia/typehierachy` extension is notoriously difficult to get working since it relied on experimental changes on top of the LSP at the time.

colin-grant-work:
At the moment, the execution of call hierarchy commands depends on services defined in the `@theia/callhierarchy` package, but we could move those over to the `plugin-ext` package without too much difficulty.

PR #10857, opened 2022-03-07
Cache instances from call hierarchy provider

colin-grant-work:
#### What it does

As part of #10310, it was found that Theia does not satisfy the expectations of VSCode CLangD (or any other language client based on `vscode-languageclient`) because we do not guarantee that the plugin will receive the same instances of `CallHierarchyItem`s that it creates when calls are made.

This PR addresses that problem by caching call hierarchy items received from plugins and retrieving them when passing them back into the plugin.

In addition, it moves some way to addressing #10528, but with a twist. Since call hierarchy items are now being cached, there's need for a signal to clear the cache, and the return from `CallHierarchyAdapter.prepareCallHierarchy` is now a `CallHierarchySession` that includes a disposable to dispose of the cache. That isn't part of the API available to plugins, so the VSCode references view never disposes of its sessions (as far as I can tell, `CallHierarchySession`s are never disposed of in VSCode), but our internal view _does_ know when to dispose of its sessions, so I have left the view in, but adjusted the types to eliminate our purely internal types in favor of close mirrors of the plugin types.

#### How to test

1. Install the latest CLangD plugin
> To install the CLangD plugin, add this line to the `.theiaPlugins` field of the repo's `package.json`:
```
"vscode-clangd": "https://open-vsx.org/api/llvm-vs-code-extensions/vscode-clangd/0.1.13/file/llvm-vs-code-extensions.vscode-clangd-0.1.13.vsix"
```
2. Open the `examples/cpp-debug-workspace` from the [Theia CPP Extensions](https://github.com/eclipse-theia/theia-cpp-extensions) repo
3. Open `a.cpp`
4. Use `Calls: Show Call Hierarchy` on one of the functions there other than `main`.
5. You should get results, and you should be able to expand those results.
6. Call hierarchy for languages like TS, etc. should continue to work as before in both references view and internal view.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
I confirmed that it works well :+1: 
- [x] the code looks good.
- [x] confirmed that `call-hierarchy` from out custom view works well using `clangd`. 
- [x] confirmed that `show call hierarchy` (plugin side) works well as well.

Issue #10860, opened 2022-03-08
Application shutdown should stop asking for vetoes after first rejection


msujew:
### Feature Description:

When having multiple `FrontendApplicationContributions` which override `onWillStop`, all `action` functions will be executed, even if one has already returned `true`. This can be reproduced as follows in the example app:

1. Create a new multi-root workspace, but don't save it
2. Edit a file, but don't save it
3. Try to exit the application
4. Clicking `Cancel` on the first dialog will cancel the shutdown, but a second dialog will still appear. The app won't shutdown independent of what you select on the second dialog

PR #10861, opened 2022-03-08
Execute app shutdown vetoes in sequence

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/10860

Since we prioritize actions anyway, it didn't make a lot of sense to execute the actions in parallel. For example, if we have the following two `OnWillStop` results, the second one returns first, even though its priority is lower:

```ts
[{
  reason: 'a',
  priority: 10,
  action: async () => {
    await timeout(10);
    return ...;
  }
},{
  reason: 'b',
  priority: 0,
  action: async () => {
    return ...;
  }
}]
```

#### How to test

1. Create a new multi-root workspace and __don't save it__
2. Edit a file in your workspace and __don't save it__
3. Exit the app
4. Any shutdown action returning `false` will prevent execution of subsequent actions

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/frontend-application.ts
+++ b/packages/core/src/browser/frontend-application.ts
@@ -55,7 +55,7 @@ export interface FrontendApplicationContribution {
      * Return `true` or an OnWillStopAction in order to prevent exit.
      * Note: No async code allowed, this function has to run on one tick.
      */
-    onWillStop?(app: FrontendApplication): boolean | undefined | OnWillStopAction;
+    onWillStop?(app: FrontendApplication): boolean | undefined | OnWillStopAction<unknown>;
 
     /**
      * Called when an application is stopped or unloaded.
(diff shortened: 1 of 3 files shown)


msujew:
I liked the parallel approach initially as well, however I noticed one issue in particular with it: When we have multiple `FrontendApplicationContributions` which have their own dialog to ensure that the user definitely wants to quit the app - even though it would be wise not to - every dialog after the first canceled dialog is completely useless and has no effect on the shutdown procedure. This is kind of bad UX in my opinion.

The reproduction/testing steps show this issue fairly well with the base app. This becomes more exaggerated in downstream Theia apps, where there are more conditions which might prevent the app from shutting down gracefully.

colin-grant-work:
Ah - that clarifies the problem for me. It definitely makes sense to bail after the first rejected option. Your code definitely does that - let me think a moment about other possibilities.

msujew:
@colin-grant-work I have an idea which should at least alleviate the issue of "data that needs some time to be prepared". What if we change the `OnWillStopAction` to be:

```ts
interface OnWillStopAction<T = undefined> {
  reason: string
  priority?: number
  prepare?(): T
  action(prepared: T): boolean
}
```

We could run `OnWillStopAction#prepare` in parallel, while passing the prepared return values in sequence into the `action` method. WDYT?

colin-grant-work:
> ... WDYT?

@msujew, I think I like it :-).

colin-grant-work (approved):
This looks good to me. It should not break any implementations oriented to the older behavior, but it does provide flexibility and cancel any irrelevant dialogues.

PR #10863, opened 2022-03-08
Refactor Menu Contribution Handler to define variables before use

colin-grant-work:
#### What it does

The code in the plugin `MenuContributionHandler` was organized so that certain callbacks referred to variables that would be defined (textually) later in the code. That had the side effect that if an error occurred in the intervening code, the callback could be called with references to variables that were never defined. This light refactor moves the declaration of function-scoped variables before the definition of any callback that refers to them to ensure that the callbacks always refer to defined variables.

#### How to test

The issue is intermittent, so it's difficult to reproduce. This PR doesn't change the logic of the code - only the order of statements.

1. Ensure that plugin-contributed menu items are present as expected.
> E.g. check that the `Call hierarchy` and `Find all references` commands from the references view are present in the editor context menu. Install something like GitLens and ensure that all of the expected icons (i.e. menu items) are present in the tree and have the expected effects.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/menus/menus-contribution-handler.ts
+++ b/packages/plugin-ext/src/main/browser/menus/menus-contribution-handler.ts
@@ -518,13 +518,28 @@ export class MenusContributionPointHandler {
         }
         const toDispose = new DisposableCollection();
         const commandId = this.createSyntheticCommandId(menu.command, { prefix: '__plugin.menu.action.' });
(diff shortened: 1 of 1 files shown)


gbodeen:
I requested this change as it's something I've encountered:
image

It's not clear to us *how* we're getting the event handler called before the `const inline` was initialized, but on the other hand there's no need here to rely on the intricacies of the hoisting mechanism, so I think changes like in this PR are justified.

paul-marechal (approved):
The refactoring looks good, I couldn't spot a difference in the logic.

PR #10868, opened 2022-03-10
New File opens untitled editor 

### LINKED ISSUE #3354: Allow creating new files without a workspace

colin-grant-work:
#### What it does

Fixes #3354 by building on #10608, and moving the code necessary to handle untitled resources into `core` rather than `plugin-ext`.

#### How to test

1. The New file keyboard shortcut and menu item should open an untitled editor without prompting immediately for a filename.
2. You should be able to use that editor to type whatever you'd like.
3. If you try to close the widget after adding content, it should prompt you to save
4. If you try to save the widget, it should open a file picker with a location determined by the following considerations (at the time you created the widget)
     - If you had another editor open at the time, it should try to place the new file in the same folder.
     - If you had a URI selected in the navigator, it should try to place the file in that location (same folder if file, in the folder if a folder is selected)
     - If you have a workspace open, it should try to place the file in the (first) root
     - It should fall back to your home directory

#### Review checklist

#### Reminder for reviewers


packages/filesystem/src/browser/filesystem-save-resource-service.ts
```diff
@@ -0,0 +1,130 @@
...
+        let exist: boolean = false;
+        let overwrite: boolean = false;
+        let selected: URI | undefined;
+        const canSave = this.canSaveNotSaveAs(sourceWidget);
+        const uri: URI = sourceWidget.getResourceUri()!;
+        let stat;
+        if (uri.scheme === 'file') {
+            stat = await this.fileService.resolve(uri).catch(() => undefined);
+        } else if (uri.scheme === UNTITLED_SCHEME) {
+            stat = await this.fileService.resolve(uri.withScheme('file').parent).catch(() => undefined);
```
msujew, line 60: I noticed that this usually leads to an undefined `stat` variable. At least it does that in my test extension.

Suggestion: Use the `UntitledFileLocationProvider#getParent()` method, if the uri parent returns a root path (`uri.withScheme('file').parent.path === '/'` or something similar), to get a better starting point for the file dialog.

packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -89,6 +89,7 @@ export namespace WorkspaceCommands {
         category: WORKSPACE_CATEGORY,
         label: 'Close Workspace'
     });
+    /** @deprecated @since 1.24 Use CommonCommands.NEW_FILE instead. */
     export const NEW_FILE = Command.toDefaultLocalizedCommand({
```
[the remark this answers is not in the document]
colin-grant-work: I left this one for the context menu in the navigator, since we still don't have the ability to do input fields in trees as we should for that command.

packages/filesystem/src/browser/filesystem-save-resource-service.ts
colin-grant-work: I've eliminated the check for a `stat` here in favor of just letting the `FileDialogService` use the `UntitledfileLocationProvider#getParent` logic (now `UserWorkingDirectoryProvider).

msujew (approved):
Looking good 👍

vince-fugnitto (approved):
I confirmed that the behavior works well :+1: 

- [x] creating untitled documents works
- [x] creating multiple untitled documents 
- [x] the untitled document is closeable without a prompt if not dirty
- [x] the untitled document produces a prompt if dirty using `save`
- [x] the untitled document produces a prompt if dirty and we close the editor 
- [x] the `save all` command does not apply for untitled documents (similarly to vscode)
- [x] performing `save all` will properly save workspace documents, but not untitled documents
- [x] toggling the `auto save` does not apply for untitled documents (does not produce a save)


I looked at the code at a high level and it seems fine to me as well.

Tesbarmix1:
Need a mobile vation

colin-grant-work:
@Tesbarmix1, I'm not sure what you mean; could you clarify?

PR #8066, opened 2020-06-20
support multi-root for 'Add Configuration'

elaihau:
- If a user is using a multi-root workspace, the command "Debug: Add Configuration" adds a configuration directly in the first workspace root instead of prompting users to select the correct root. With this change the command prompts users with a quick-pick which allows them to select their desired workspace root.

- resolves #5166

Signed-off-by: Liang Huang <lhuang4@ualberta.ca>

#### How to test

- in a workspace that has only one root folder, the command "Debug: Add Configuration" should work the way it used ot

- in a multi root workspace, we should see a prompt that asks users to select which root folder the configuration should be added to.

Peek 2020-06-20 15-58

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

Issue #5166, opened 2019-05-16
[debug] support multi-root for 'Add Configuration'


vince-fugnitto:
**Description**

The command **`Debug: Add Configuration`** does not work nicely in a multi-root context.

If a user is using a multi-root workspace, the command adds a configuration directly in the first workspace root instead of prompting users to select the correct root.

I propose that when in a multi-root workspace, the command should prompt users with a `quick-pick` which allows them to select their desired workspace root.

PR #10875, opened 2022-03-11
debug: add multi-root support for add configurations

vince-fugnitto:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/5166.
Closes: https://github.com/eclipse-theia/theia/pull/8066.

The commit adds support when executing `add configuration...` in a multi-root workspace where the user will first be prompted to choose the location for their config (choose the root) before adding. Previously, the first root was always chosen.

#### How to test

I verified the changes using multiple different setups:

1. confirm in a single-root workspace, that the changes properly add configurations (verify with `.vscode` and `.theia`)
2. confirm in a multi-root workspace that users will be prompted to choose a root before adding a configuration
3. confirm step 2 using combinations of workspaces without/with configuration folders, mix-and-match of `.vscode` and `.theia`. 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-configuration-manager.ts
+++ b/packages/debug/src/browser/debug-configuration-manager.ts
@@ -26,7 +26,7 @@ import URI from '@theia/core/lib/common/uri';
 import { Emitter, Event, WaitUntilEvent } from '@theia/core/lib/common/event';
 import { EditorManager, EditorWidget } from '@theia/editor/lib/browser';
 import { MonacoEditor } from '@theia/monaco/lib/browser/monaco-editor';
-import { PreferenceScope, PreferenceService, QuickPickValue, StorageService } from '@theia/core/lib/browser';
+import { LabelProvider, PreferenceScope, PreferenceService, QuickPickValue, StorageService } from '@theia/core/lib/browser';
 import { QuickPickService } from '@theia/core/lib/common/quick-pick-service';
 import { WorkspaceService } from '@theia/workspace/lib/browser/workspace-service';
 import { DebugConfigurationModel } from './debug-configuration-model';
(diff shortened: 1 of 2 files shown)


packages/debug/src/browser/debug-configuration-manager.ts
```diff
@@ -191,14 +195,23 @@ export class DebugConfigurationManager {
...
+        const model = this.getModel();
         if (model) {
             await this.doOpen(model);
         }
     }
 
     async addConfiguration(): Promise<void> {
-        const { model } = this;
+        let rootUri: URI | undefined = undefined;
+        if (this.workspaceService.isMultiRootWorkspaceOpened) {
```
colin-grant-work, line 206: This field doesn't actually guarantee that there are multiple roots, only that the the workspace is defined in a workspace file, so the `selectRootUri` should have special handling (not show the quick pick) in case of only 1 root (or 0).

colin-grant-work (approved):
- [x] No workspace -> No op (we should perhaps hide the option in that case, but that's out of scope here)
 - [x] Unsaved workspace no launch -> creates and opens `.theia/launch.json`
 - [x] Unsaved workspace .vscode/launch -> opens `.vscode/launch.json`
 - [x] Unsaved workspace .theia/launch -> opens `.vscode/launch.json`
 - [x] Unsaved workspace both/launch -> opens `.theia/launch.json`
 - [x] Saved empty workspace -> No op (see above)
 - [x] Saved single-root no launch -> creates and opens `.theia/launch.json`
 - [x] Saved single-root .vscode/launch -> opens `.vscode/launch.json`
 - [x] Saved single-root .theia/launch -> opens `.theia/launch.json`
 - [x] Saved single-root both/launch -> opens `.theia/launch.json`
 - [x] Saved multi-root select with no launch -> quick pick -> creates and opens `.theia/launch.json`
 - [x] Saved multi-root select with .vscode/launch -> quick pick -> opens `.vscode/launch.json`
 - [x] Saved multi-root select with .theia/launch -> quick pick -> opens `.theia/launch.json`
 - [x] Saved multi-root select with both/launch -> quick pick -> `.theia/launch.json`

PR #10881, opened 2022-03-14
Update type check for `TreeContainerProps`

colin-grant-work:
#### What it does

The previous implementation of the check to distinguish `TreeProps` from `TreeContainerProps` could be confused by the fact that both had a `search` key. This PR introduces special handling of that field to ensure that the check returns the correct value.

#### How to test

1. Assert that the new test introduced in this PR passes.
2. use `createTreeContainer` with `TreeProps` that include a `search` field. Observe that the tree view is successfully constructed and functions correctly.

e.g. in `outline-view-frontend-module.ts` replace this:

```typescript
    const child = createTreeContainer(parent, {
         props: { expandOnlyOnExpansionToggleClick: true, search: true },
         widget: OutlineViewWidget,
         model: OutlineViewTreeModel,
         decoratorService: OutlineDecoratorService,
     });
```

with this:
```typescript
    const child = createTreeContainer(parent, { ...defaultTreeProps, expandOnlyOnExpansionToggleClick: true, search: true });


     child.unbind(TreeWidget);
     child.bind(OutlineViewWidget).toSelf();

     child.unbind(TreeModelImpl);
     child.bind(OutlineViewTreeModel).toSelf();
     child.rebind(TreeModel).toService(OutlineViewTreeModel);

     child.bind(OutlineDecoratorService).toSelf().inSingletonScope();
     child.rebind(TreeDecoratorService).toDynamicValue(ctx => ctx.container.get(OutlineDecoratorService)).inSingletonScope();
```

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/tree/tree-container.spec.ts
+++ b/packages/core/src/browser/tree/tree-container.spec.ts
@@ -13,6 +13,7 @@
 //
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 // *****************************************************************************
+
 import * as assert from 'assert';
 import { Container } from 'inversify';
 import { createTreeContainer, isTreeServices } from './tree-container';
--- a/packages/core/src/browser/tree/tree-container.spec.ts
+++ b/packages/core/src/browser/tree/tree-container.spec.ts
@@ -18,11 +18,15 @@ import { isTreeServices } from './tree-container';
 import { TreeSearch } from './tree-search';
 import { defaultTreeProps } from './tree-widget';
 
-describe('IsTreeServices should accurately distinguish TreeProps from TreeContainerProps', () => {
-    it('should not be confused by the fact that both have a `search` key', () => {
-        assert(isTreeServices({
-            ...defaultTreeProps, search: true, multiSelect: true, globalSelection: true, contextMenuPath: ['so-contextual']
-        }) === false, 'search:boolean -> false');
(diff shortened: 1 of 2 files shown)


packages/core/src/browser/tree/tree-container.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+// *****************************************************************************
+import * as assert from 'assert';
+import { isTreeServices } from './tree-container';
+import { TreeSearch } from './tree-search';
+import { defaultTreeProps } from './tree-widget';
+
+describe('IsTreeServices should accurately distinguish TreeProps from TreeContainerProps', () => {
```
vince-fugnitto, line 21: My only suggestion would be to have a top level `describe` to describe the whole file before we have individual test-cases:

```ts
describe('tree-container', () => {
...
});
```

It helps with the output of the tests, and helps group additional tests later for the same file.

vince-fugnitto (approved):
I confirmed with the changes that the `outline-view` can successfully be constructed and viewed (the `search` does not work, but it does without the changes to the `outline-view` itself).

vince-fugnitto (approved):
I confirmed that the new code works well, the outline-view works including the snippet, and so does searching in the trees.

PR #10887, opened 2022-03-16
No validation warning for undefined preference value

colin-grant-work:
#### What it does

Reduces the number of warnings logged during preference validation by not warning if the initial value is `undefined`. Since `undefined` is not a JSON value, it doesn't reflect a bad user setting, so the warning is unnecessary.

#### How to test

1. Start the application, open an editor, and open the dev tools.
2. On master, you would see something like this:
image

3. Observe that you do not see the warnings for `undefined` values.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
LGTM 👍

PR #10900, opened 2022-03-18
Check application state before showing file resource dialog

### LINKED ISSUE #9843: Layout cannot restored when a large binary file is opened.
### Bug Description:
Layout cannot restored when a large binary file is opened.

### Steps to Reproduce:

1. open a large binary file
2. refresh page
3. page cannot be restored.
image

The output error:
image

I find that the error is reported here. Does the judgment of object equality need to be modified?
image

### Additional Information

- Operating System: linux
- Theia Version: 1.13.0

colin-grant-work:
#### What it does

Fixes #9843 by checking application state before trying to show a file-resource-related dialog.

I think ideally, it wouldn't be the resource that was responsible for this dialog: different contexts should be allowed to handle it differently. But this fix does address the specific problem described in the bug with a minimum change in the architecture. I'm open to other suggestions (for example, maybe if we're running remotely, we should automatically reject rather than automatically approve the request?).

#### How to test

1. Open the application.
2. Open a binary file or large file of some kind.
3. Observe that you get a dialog requesting confirmation that you want to open the binary file as text.
4. Close and reopen the application.
5. The layout should be restored, and the file should be opened as text without a dialog prompt.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/browser/file-resource.ts
+++ b/packages/filesystem/src/browser/file-resource.ts
@@ -22,7 +22,7 @@ import { Readable, ReadableStream } from '@theia/core/lib/common/stream';
 import URI from '@theia/core/lib/common/uri';
 import { FileOperation, FileOperationError, FileOperationResult, ETAG_DISABLED, FileSystemProviderCapabilities, FileReadStreamOptions, BinarySize } from '../common/files';
 import { FileService, TextFileOperationError, TextFileOperationResult } from './file-service';
-import { ConfirmDialog } from '@theia/core/lib/browser/dialogs';
+import { ConfirmDialog, Dialog } from '@theia/core/lib/browser/dialogs';
 import { LabelProvider } from '@theia/core/lib/browser/label-provider';
 import { GENERAL_MAX_FILE_SIZE_MB } from './filesystem-preferences';
 import { FrontendApplicationStateService } from '@theia/core/lib/browser/frontend-application-state';
@@ -327,8 +327,8 @@ export class FileResourceResolver implements ResourceResolver {
             title: nls.localize('theia/filesystem/fileResource/overwriteTitle', "The file '{0}' has been changed on the file system.", this.labelProvider.getName(uri)),
             msg: nls.localize('theia/fileSystem/fileResource/overWriteBody',
                 "Do you want to overwrite the changes made to '{0}' on the file system?", this.labelProvider.getLongName(uri)),
-            ok: 'Yes',
-            cancel: 'No'
+            ok: Dialog.YES,
+            cancel: Dialog.NO,
         });
         return !!await dialog.open();
     }
@@ -345,8 +345,8 @@ export class FileResourceResolver implements ResourceResolver {
                     msg: nls.localize('theia/filesystem/fileResource/binaryFileQuery',
                         "Opening it might take some time and might make the IDE unresponsive. Do you want to open '{0}' anyway?", this.labelProvider.getLongName(uri)
                     ),
-                    ok: 'Yes',
-                    cancel: 'No'
+                    ok: Dialog.YES,
+                    cancel: Dialog.NO,
                 });
                 return !!await dialog.open();
             }
--- a/packages/filesystem/src/browser/file-resource.ts
+++ b/packages/filesystem/src/browser/file-resource.ts
@@ -26,6 +26,7 @@ import { ConfirmDialog } from '@theia/core/lib/browser/dialogs';
 import { LabelProvider } from '@theia/core/lib/browser/label-provider';
 import { GENERAL_MAX_FILE_SIZE_MB } from './filesystem-preferences';
 import { FrontendApplicationStateService } from '@theia/core/lib/browser/frontend-application-state';
+import { nls } from '@theia/core';
 
 export interface FileResourceVersion extends ResourceVersion {
     readonly encoding: string;
@@ -110,14 +111,15 @@ export class FileResource implements Resource {
             return stat.value;
         } catch (e) {
             if (e instanceof TextFileOperationError && e.textFileOperationResult === TextFileOperationResult.FILE_IS_BINARY) {
-                if (await this.shouldOpenAsText('The file is either binary or uses an unsupported text encoding.')) {
+                if (await this.shouldOpenAsText(nls.localize('theia/filesystem/fileResource/binaryTitle', 'The file is either binary or uses an unsupported text encoding.'))) {
                     this.acceptTextOnly = false;
                     return this.readContents(options);
                 }
             } else if (e instanceof FileOperationError && e.fileOperationResult === FileOperationResult.FILE_TOO_LARGE) {
                 const stat = await this.fileService.resolve(this.uri, { resolveMetadata: true });
                 const maxFileSize = GENERAL_MAX_FILE_SIZE_MB * 1024 * 1024;
-                if (this.limits?.size !== maxFileSize && await this.shouldOpenAsText(`The file is too large (${BinarySize.formatSize(stat.size)}).`)) {
+                if (this.limits?.size !== maxFileSize && await this.shouldOpenAsText(nls.localize(
+                    'theia/filesystem/fileResource/largeFileTitle', 'The file is too large ({0}).', BinarySize.formatSize(stat.size)))) {
                     this.limits = {
                         size: maxFileSize
                     };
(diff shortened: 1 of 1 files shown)


packages/filesystem/src/browser/file-resource.ts
```diff
@@ -326,13 +330,21 @@ export class FileResourceResolver implements ResourceResolver {
...
-        return !!await dialog.open();
+        switch (this.applicationState.state) {
+            case 'init':
+            case 'started_contributions':
+            case 'attached_shell':
+                return true; // We're restoring state - assume that we should open files that were previously open.
+            default: {
+                const dialog = new ConfirmDialog({
+                    title: error,
+                    msg: `Opening it might take some time and might make the IDE unresponsive. Do you want to open '${this.labelProvider.getLongName(uri)}' anyway?`,
```
vince-fugnitto, line 341: @msujew this makes me wonder if we are somehow missing some localizations, perhaps we need to enhance the custom eslint rule. 

The same is likely true here (unless it is done automatically that I am not aware of):

https://github.com/eclipse-theia/theia/blob/432602946323bc1500c585de4432021b6d6a4a0a/packages/filesystem/src/browser/file-resource.ts#L320-L321

How should we go about tracking what might be missing?

msujew: Yeah, I noticed that as well. I mostly looked for widgets, commands, menus etc. but missed a lot of dialogs. I'm still unsure how we write such a eslint rule, without generating tons of false-positives. Only for strings inside of certain dialog constructors (`new ConfirmDialog(...)` etc)?

vince-fugnitto: That's true, it might produce too many false-positives, even if we were to only apply the rule to `browser/`. It might be worthwhile however to have a way to run the check manually periodically so we can see if we missed anything, but we would need to also manually verify if they are valid.

packages/filesystem/src/browser/file-resource.ts
```diff
@@ -153,14 +156,15 @@ export class FileResource implements Resource {
...
+                if (await this.shouldOpenAsText(nls.localize('theia/filesystem/fileResource/binaryTitle', 'The file is either binary or uses an unsupported text encoding.'))) {
                     this.acceptTextOnly = false;
                     return this.readStream(options);
                 }
             } else if (e instanceof FileOperationError && e.fileOperationResult === FileOperationResult.FILE_TOO_LARGE) {
                 const stat = await this.fileService.resolve(this.uri, { resolveMetadata: true });
                 const maxFileSize = GENERAL_MAX_FILE_SIZE_MB * 1024 * 1024;
-                if (this.limits?.size !== maxFileSize && await this.shouldOpenAsText(`The file is too large (${BinarySize.formatSize(stat.size)}).`)) {
+                if (this.limits?.size !== maxFileSize && await this.shouldOpenAsText(nls.localize(
+                    'theia/filesystem/fileResource/largeFileTitle', 'The file is too large ({0}).', BinarySize.formatSize(stat.size)))) {
```
msujew, line 167:
```suggestion
                if (this.limits?.size !== maxFileSize && await this.shouldOpenAsText(nls.localizeByDefault(
                    'The file is not displayed in the editor because it is too large ({0}).', BinarySize.formatSize(stat.size)))) {
```

packages/filesystem/src/browser/file-resource.ts
```diff
@@ -153,14 +156,15 @@ export class FileResource implements Resource {
             return stat.value;
         } catch (e) {
             if (e instanceof TextFileOperationError && e.textFileOperationResult === TextFileOperationResult.FILE_IS_BINARY) {
-                if (await this.shouldOpenAsText('The file is either binary or uses an unsupported text encoding.')) {
+                if (await this.shouldOpenAsText(nls.localize('theia/filesystem/fileResource/binaryTitle', 'The file is either binary or uses an unsupported text encoding.'))) {
```
msujew, line 159:
```suggestion
                if (await this.shouldOpenAsText(nls.localizeByDefault('The file is not displayed in the editor because it is either binary or uses an unsupported text encoding.'))) {
```

packages/filesystem/src/browser/file-resource.ts
```diff
@@ -109,14 +111,15 @@ export class FileResource implements Resource {
...
```
msujew, line 114:
```suggestion
                if (await this.shouldOpenAsText(nls.localizeByDefault('The file is not displayed in the editor because it is either binary or uses an unsupported text encoding.'))) {
```

packages/filesystem/src/browser/file-resource.ts
```diff
@@ -109,14 +111,15 @@ export class FileResource implements Resource {
...
                     this.acceptTextOnly = false;
                     return this.readContents(options);
                 }
             } else if (e instanceof FileOperationError && e.fileOperationResult === FileOperationResult.FILE_TOO_LARGE) {
                 const stat = await this.fileService.resolve(this.uri, { resolveMetadata: true });
                 const maxFileSize = GENERAL_MAX_FILE_SIZE_MB * 1024 * 1024;
-                if (this.limits?.size !== maxFileSize && await this.shouldOpenAsText(`The file is too large (${BinarySize.formatSize(stat.size)}).`)) {
+                if (this.limits?.size !== maxFileSize && await this.shouldOpenAsText(nls.localize(
+                    'theia/filesystem/fileResource/largeFileTitle', 'The file is too large ({0}).', BinarySize.formatSize(stat.size)))) {
```
msujew, line 122:
```suggestion
                if (this.limits?.size !== maxFileSize && await this.shouldOpenAsText(nls.localizeByDefault(
                    'The file is not displayed in the editor because it is too large ({0}).', BinarySize.formatSize(stat.size)))) {
```

packages/filesystem/src/browser/file-resource.ts
```diff
@@ -317,22 +324,33 @@ export class FileResourceResolver implements ResourceResolver {
 
     protected async shouldOverwrite(uri: URI): Promise<boolean> {
         const dialog = new ConfirmDialog({
-            title: `The file '${this.labelProvider.getName(uri)}' has been changed on the file system.`,
-            msg: `Do you want to overwrite the changes made to '${this.labelProvider.getLongName(uri)}' on the file system?`,
+            title: nls.localize('theia/filesystem/fileResource/overwriteTitle', "The file '{0}' has been changed on the file system.", this.labelProvider.getName(uri)),
+            msg: nls.localize('theia/fileSystem/fileResource/overWriteBody',
+                "Do you want to overwrite the changes made to '{0}' on the file system?", this.labelProvider.getLongName(uri)),
             ok: 'Yes',
             cancel: 'No'
```
msujew, line 331:
```suggestion
            ok: Dialog.YES,
            cancel: Dialog.NO
```

packages/filesystem/src/browser/file-resource.ts
```diff
@@ -317,22 +324,33 @@ export class FileResourceResolver implements ResourceResolver {
...
+            case 'attached_shell':
+                return true; // We're restoring state - assume that we should open files that were previously open.
+            default: {
+                const dialog = new ConfirmDialog({
+                    title: error,
+                    msg: nls.localize('theia/filesystem/fileResource/binaryFileQuery',
+                        "Opening it might take some time and might make the IDE unresponsive. Do you want to open '{0}' anyway?", this.labelProvider.getLongName(uri)
+                    ),
+                    ok: 'Yes',
+                    cancel: 'No'
```
msujew, line 349:
```suggestion
                    ok: Dialog.YES,
                    cancel: Dialog.NO
```

msujew (commented):
Just a minor comment: A few of those texts already exist in the `nls.metadata.json` file.

packages/filesystem/src/browser/file-resource.ts
colin-grant-work: Unfortunately, we can't use the one from the `metadata.json` because this message isn't restricted to the editor package: any file operation that tries to access a binary file or a file that's too large will get these messages. I think that that's the wrong approach, but it's the way it's currently handled, and changing it would require a non-trivial reworking of the code that loads such files. I was tempted, though :-).

packages/filesystem/src/browser/file-resource.ts
colin-grant-work: Same as above.

packages/filesystem/src/browser/file-resource.ts
msujew: Ah, I see, then I'll resolve those comments, thanks for the explanation.

vince-fugnitto (approved):
I confirm that the changes work well for me :+1: 
I verified by:
1. starting the application
2. open a `.png` with the code editor - prompt is shown
3. reload the application
4. the app starts and properly restores the editor - on master the loading animation would go on forever

PR #10901, opened 2022-03-18
Draft: Add a watermark widget

### LINKED ISSUE #10804: Helpful Banner on Empty Window
### Feature Description:

In VSCode, when no tabs are present, the VSCode logo is shown, together with a banner showing some of the most important keyboard shortcuts in the application:

image

Since it can otherwise be difficult (as a new user) to discover features like the global command palette, this banner is very helpful, and we should consider implementing something similar.

### LINKED ISSUE #10878: Getting Started Widget doesn't accept focus
### Steps to Reproduce:

1. Open the application with no workspace.
2. Try to focus the Getting Started Widget.
3. Observe that the tab is never marked active (brighter text) and you see a log:
```
root WARN Widget was activated, but did not accept focus after 2000ms: getting.started.widget
```

in order for a widget to accept focus it needs to implement onActivateRequest and needs something to focus on. The same was done in blueprint for the same view:

https://github.com/eclipse-theia/theia-blueprint/pull/201

### Additional Information

- Theia Version: `master` (e62fd33ce46055d1633d781c08bf5852535f53c5)

alxflam:
#### What it does
Adds a watermark widget which, similar to the one in VSCode, displays useful commands and their keybindings.
Closes #10804.

Marked this as a draft for now due to some open topics:
- Is the implementation approach taken fine? (adding a new contribution point for watermark commands and moving existing keybinding rendering logic from the keymaps package into core to enable a watermark widget already in core instead of in a separate package - see #10804 for earlier discussion)
-  Which commands should be shown as watermarks? (for now, i only used a subset of the same commands VSCode shows for demonstration - i'd simply use the same commands VSCode displays)

watermark

#### How to test
- Verify watermark widget is shown when there's no open editor

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/frontend-application-module.ts
+++ b/packages/core/src/browser/frontend-application-module.ts
@@ -122,6 +122,10 @@ import { TooltipService, TooltipServiceImpl } from './tooltip-service';
 import { bindFrontendStopwatch, bindBackendStopwatch } from './performance';
 import { SaveResourceService } from './save-resource-service';
 import { UserWorkingDirectoryProvider } from './user-working-directory-provider';
+import { WatermarkWidget } from './watermark/watermark-widget';
+import { WatermarkCommandRegistry } from './watermark/watermark-command-registry';
+import { KeybindingUtil } from './keybinding/keybinding-util';
+import { WatermarkCommandApplicationContribution, WatermarkCommandContribution } from './watermark';
 
 export { bindResourceProvider, bindMessageService, bindPreferenceService };
 
@@ -224,6 +228,7 @@ export const frontendApplicationModule = new ContainerModule((bind, unbind, isBo
     });
     bind(CommandService).toService(CommandRegistry);
     bindContributionProvider(bind, CommandContribution);
+    bind(KeybindingUtil).toSelf().inSingletonScope();
 
     bind(ContextKeyService).to(ContextKeyServiceDummyImpl).inSingletonScope();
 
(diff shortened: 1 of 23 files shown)


vince-fugnitto (commented):
@alxflam thank you for your contribution and interest in implementing the watermark :+1: 

I noticed a couple of issues with a high-level review:

1. It might be overkill to move all the rendering logic of keybindings to core, all you really need is the CSS styling. The watermark should not need to know about doing fuzzy matching (highlights when searching) or any specifics of the keybindings-widget. The changeset should then be much smaller.
2. The keybindings-widget no longer works from what I tested, when searching no results exist:
keybindings-search-bug
3. On electron I noticed that a command does not show the `label`, instead we get the `id`:

image

alxflam:
@vince-fugnitto thanks for your feedback. I removed the code in KeybindingUtil to the necessary minimum and moved everything back to the keymaps package (also solves the issue in the keybindings widget). The 'Open File' command does not have a label, only a dialogLabel - seems this caused the issue on electron (for now i added a fallback to dialogLabel in case label is not set).

Current state is still rather rough, so no in-depth review required.

vince-fugnitto:
@alxflam I think the general idea is fine if you want to continue, cleanup and finalize the pull-request. We can perform a more in depth review of the changes when it is finalized (it will likely go through further iterations so don't worry about converting it from a draft).

alxflam:
@vince-fugnitto thanks, i'll provide an update

JonasHelming:
@alxflam : What is the current state of this PR, is this ready for another review? If so, could you resolve the conflicts, please?

chakraborty-slx:
@alxflam are you looking into this?

alxflam:
@JonasHelming @chakraborty-slx It might take a while until i can get another look at this, if someone else wants to take over feel free to do so and reuse my commits.

vince-fugnitto:
@alxflam friendly note that we might close the pull-request soon for the moment to keep the list of open pull-requests more manageable but we can always re-open and re-visit the pull-request in the future when needed and when updated.

Issue #10907, opened 2022-03-22
monaco: menus should not overlay the main-menu


vince-fugnitto:
#### Bug Description:

The monaco menus (`quick-open`, `quick-input`) should not overlay the main-menu but instead be attached to the main area (similarly to vscode).

![monaco-menu](https://user-images.githubusercontent.com/40359487/159518115-45793552-8938-45c0-9476-d5ab47ca16b2.gif)

#### Steps to Reproduce:

1. start the browser or electron application with the custom titlebar
2. press <kbd>F1</kbd>
3. notice that the monaco menu is overlaying the main-menu

### Additional Information

- Theia Version: https://github.com/eclipse-theia/theia/commit/6d67d664bd3f679fca68be51d06d360a115decb2

PR #10909, opened 2022-03-22
monaco: adjust where the monaco menu appears

vince-fugnitto:
#### What it does

Fixes: #10907 

The pull-request adjusts the monaco menu (quick-input) to appear after the main-menu and not overlay it.

https://user-images.githubusercontent.com/40359487/159699337-5d0f5480-6831-483b-9db9-9d38711f66e6.mp4

#### How to test

1. start the browser application
2. ensure that the different menus work well (quick-file open, quick-commands, etc.)
3. start the electron application with the `native` titlebar style
4. repeat step 2
5. start the electron application with the `custom` titlebar style
6. repeat step 2
7. the change should also work nicely with the toolbar toggled 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/style/index.css
+++ b/packages/monaco/src/browser/style/index.css
@@ -4,6 +4,12 @@
   font-size: inherit !important;
 }
 
+#quick-input-container {
+  position: fixed;
+  right: 50%;
+  z-index: 1000000;
+}
+
 /*
  * set z-index to 0, so tabs are not above overlay widgets
  */
--- a/packages/monaco/src/browser/monaco-quick-input-service.ts
+++ b/packages/monaco/src/browser/monaco-quick-input-service.ts
@@ -15,11 +15,12 @@
 // *****************************************************************************
 
 import {
+    ApplicationShell,
     InputBox, InputOptions, KeybindingRegistry, PickOptions,
     QuickInputButton, QuickInputHideReason, QuickInputService, QuickPick, QuickPickItem,
     QuickPickItemButtonEvent, QuickPickItemHighlights, QuickPickOptions, QuickPickSeparator
 } from '@theia/core/lib/browser';
-import { injectable, inject } from '@theia/core/shared/inversify';
+import { injectable, inject, postConstruct } from '@theia/core/shared/inversify';
 import {
     IInputBox, IInputOptions, IKeyMods, IPickOptions, IQuickInput, IQuickInputButton,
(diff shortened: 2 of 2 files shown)


packages/monaco/src/browser/style/index.css
```diff
@@ -4,6 +4,13 @@
   font-size: inherit !important;
 }
 
+#quick-input-container {
+  position: absolute;
+  top: 0px;
+  right: 50%;
+  z-index: 1000000;
+}
+
```
msujew, line 13: I'm wondering whether we need this change. It doesn't seem to do anything, and removing it doesn't change the behavior.

msujew (commented):
While the solution is quite clean, I'm wondering whether this is the right approach. When attaching the div to the main panel of Theia, it will be offset from the center of the app. See this:

image

vince-fugnitto:
> While the solution is quite clean, I'm wondering whether this is the right approach. When attaching the div to the main panel of Theia, it will be offset from the center of the app. See this:

@msujew ahh you're right, I see the odd behavior when I modify the layout. I'm not checking what might be a better choice to append to, but I'm also worried if downstream they modify their `shell` or layout. Do you have any suggestions?

msujew:
Yeah, the downstream issue is difficult. I have two ideas:

1. Disregard downstream apps that modify their layout and simply append to the html element with the `#theia-left-right-split-panel` id. That way we always stay in the center of the app with the quick input.
2. Keep the quick input attached to the body of the document and simply change the `top` style attribute to consider the top-panel height. I assume we would need to change the `--theia-private-menubar-height` css variable to `0px` if we hide the top panel in any way (electron native mode/compact menu bar mode). It's probably a more complicated approach.

vince-fugnitto:
I think I need to adjust how the overlay works, currently the menu is hidden behind other parts of the layout like the sidepanels.

vince-fugnitto:
@msujew I modified the code slightly to use a fixed position and it seems to have resolved the issue, is there an issue with this approach?

colin-grant-work (approved):
LGTM. The quick input no longer obscures the menu bar.

colin-grant-work:
It looks like appending the QP container to the main area has negative side effects when side panels are expanded:

image

Perhaps we should revert to attaching the QP to the body, but calculate its `top` value to match the top of `main` when showing the quick pick?

Issue #7076, opened 2020-02-04
[electron] Electron not showing the active editor in the window title as in VScode


lmcbout:
### Description
 
Theia electron version does not show the active editor in the window title as in VSCode

### Reproduction Steps

Open a file in Theia 
--> observe the window title, it does not show the active file editor
Open the same workspace with VSCode
--> Observe, the window title is upsdated including the active file editor 

**OS and Theia version:**
OS: Ubuntu 18.04
Theia : latest commit cc0da4e69f8141122496c119bb72f4df46584e76

**Diagnostics:** 
Theia electron:
![ElectronTitleNotshowingActiveeditor](https://user-images.githubusercontent.com/419051/73766818-94eb0580-4744-11ea-8b54-303fce2d8d2d.gif)

VSCode electron:
![VSCodeShowPreview](https://user-images.githubusercontent.com/419051/73766853-a2a08b00-4744-11ea-8848-3cb7e91e1498.gif)

vince-fugnitto:
I don't believe it is possible at the moment, but perhaps after #4869 (at least for electron)
I believe it is also a small enhancement, for instance the `browser` target also does not update based on the active file.

akosyakov:
It is not fix, we have quite different concept of what is active compare to VS Code.

vince-fugnitto:
@lmcbout I've removed the `electron` label since the enhancement is not specific to electron (the browser target also does not display the active editor). As Anton mentioned, our concept of **active** is quite different and due to limitations it might not be easy to update the title.

I see the enhancement as a low-priority if we ever do decide to work on it.

lmcbout:
Low enhancement priority is fine with me
Thanks

PR #10916, opened 2022-03-23
Add a title for custom title bar style

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/7076

When selecting the `custom` title bar style, the application currently shows no title. This change adresses the issue by trying to replicate what VSCode does with its title bars. Also aligns the title behavior of Theia to VSCode which shows the currently opened workspace and active main area widget in the title. For this, some more title-based infrastructure has been introduced to facilitate different contributions to the title.

#### How to test

1. Start the electron app on Windows or Linux and select the `custom` title bar style in the preferences
2. Assert that the title bar behaves like VSCode when resizing the window.
3. Open a workspace and start opening files or other widgets which appear in the main area. The currently opened workspace and the currently selected main area widget appear in the title bar.
4. Closing all workspaces/main area widgets removes these contributions from the title bar.
5. Assert that these changes also work correctly when using the `native` title bar style and in the browser.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/window/window-title-service.ts
+++ b/packages/core/src/browser/window/window-title-service.ts
@@ -1,0 +1,107 @@
+// *****************************************************************************
+// Copyright (C) 2022 TypeFox and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { inject, injectable, postConstruct } from 'inversify';
+import { escapeRegExpCharacters } from '../../common/strings';
+import { Emitter, Event } from '../../common/event';
+import { CorePreferences } from '../core-preferences';
+import { FrontendApplicationConfigProvider } from '../frontend-application-config-provider';
+
+export const InitialWindowTitleParts = {
+    activeEditorShort: undefined,
+    activeEditorMedium: undefined,
+    activeEditorLong: undefined,
+    activeFolderShort: undefined,
+    activeFolderMedium: undefined,
+    activeFolderLong: undefined,
+    folderName: undefined,
+    folderPath: undefined,
+    rootName: undefined,
+    rootPath: undefined,
+    appName: FrontendApplicationConfigProvider.get().applicationName,
+    remoteName: undefined,
+    dirty: undefined,
+    developmentHost: undefined
+};
+
+@injectable()
+export class WindowTitleService {
+
+    @inject(CorePreferences)
+    protected readonly preferences: CorePreferences;
+
+    protected _title = '';
+    protected titleTemplate?: string;
+
+    protected onDidChangeTitleEmitter = new Emitter<string>();
+    protected titleParts = new Map<string, string | undefined>(Object.entries(InitialWindowTitleParts));
+    protected separator = ' - ';
+
+    @postConstruct()
+    protected init(): void {
+        this.titleTemplate = this.preferences['window.title'];
+        this.separator = this.preferences['window.titleSeparator'];
+        this.updateTitle();
+        this.preferences.onPreferenceChanged(e => {
+            if (e.preferenceName === 'window.title') {
+                this.titleTemplate = e.newValue;
+                this.updateTitle();
+            } else if (e.preferenceName === 'window.titleSeparator') {
+                this.separator = e.newValue;
+                this.updateTitle();
+            }
+        });
+    }
+
+    get onDidChangeTitle(): Event<string> {
+        return this.onDidChangeTitleEmitter.event;
+    }
+
+    get title(): string {
+        return this._title;
+    }
+
+    update(parts: Record<string, string | undefined>): void {
+        for (const [key, value] of Object.entries(parts)) {
+            this.titleParts.set(key, value);
+        }
+        this.updateTitle();
+    }
+
+    protected updateTitle(): void {
+        if (!this.titleTemplate) {
+            this._title = '';
+        } else {
+            let title = this.titleTemplate;
+            for (const [key, value] of this.titleParts.entries()) {
+                if (key !== 'developmentHost') {
+                    const label = `$\{${key}\}`;
+                    const regex = new RegExp(escapeRegExpCharacters(label), 'g');
+                    title = title.replace(regex, value ?? '');
+                }
+            }
+            const separatedTitle = title.split('${separator}').filter(e => e.length > 0);
+            this._title = separatedTitle.join(this.separator);
+        }
+        const developmentHost = this.titleParts.get('developmentHost');
+        if (developmentHost) {
+            this._title = developmentHost + this.separator + this._title;
+        }
+        document.title = this._title;
+        this.onDidChangeTitleEmitter.fire(this._title);
+    }
+
+}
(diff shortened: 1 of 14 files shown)


vince-fugnitto:
I haven't looked at the code yet but I believe by reading the description it may fix https://github.com/eclipse-theia/theia/issues/7076.

packages/core/src/browser/window/window-title-service.ts
```diff
@@ -0,0 +1,68 @@
...
+// *****************************************************************************
+
+import { injectable } from 'inversify';
+import { Emitter, Event } from '../../common/event';
+
+export interface WindowTitlePart {
+    id?: string
+    value?: string
+    priority?: number
+    event?: Event<string | undefined>
```
colin-grant-work, line 24: No strong feelings here, but in `TabBarToolbarItem`s, the corresponding functionality is under a field `onDidChange`.

colin-grant-work (commented):
Functionally, looking good, and only minor comments on the code.

packages/core/src/browser/window/window-title-service.ts
```diff
@@ -0,0 +1,68 @@
...
+            this.titleParts.splice(index, 1);
+            this.updateTitle();
+        }
+    }
+
+    protected updateTitle(): void {
+        const sortedParts = this.titleParts.sort((a, b) => (b.priority ?? 100) - (a.priority ?? 100));
+        this._title = sortedParts.map(e => e.value).filter(e => e).join(' - ');
+        this.onDidChangeTitleEmitter.fire(this._title);
+    }
```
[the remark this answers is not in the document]
msujew: Ah, I didn't know about this. I think that's a better approach than the title parts. I'll rewrite the feature to accommodate for it :)

msujew:
@colin-grant-work I refactored the feature to use the `window.title` preference. Care to take another look?

packages/core/src/browser/window/window-title-service.ts
```diff
@@ -0,0 +1,100 @@
...
+                const label = `$\{${key}\}`;
+                const value = this.titleParts.get(key) ?? '';
+                title = title.replace(label, value);
+            }
+            const separatedTitle = title.split('${separator}').filter(e => e.length > 0);
+            this._title = separatedTitle.join(this.separator);
+        }
+        if (this.titleParts.get('developmentHost')) {
+            this._title = this.titleParts.get('developmentHost') + this.separator + this._title;
+        }
```
colin-grant-work, line 95:
```suggestion
        const developmentHost = this.titleParts.get('developmentHost');
        if (developmentHost) {
            this._title = developmentHost + this.separator + this._title;
        }
```

packages/core/src/browser/window/window-title-service.ts
```diff
@@ -0,0 +1,100 @@
...
+    'activeFolderLong': true,
+    'folderName': true,
+    'folderPath': true,
+    'rootName': true,
+    'rootPath': true,
+    'appName': true,
+    'remoteName': true,
+    'dirty': true,
+    'developmentHost': true
+};
```
colin-grant-work, line 36: I don't mean to make life difficult, but would it be possible to get the best of your original implementation (dynamic contributions) and support (most of) VSCode's options? I can imagine downstream IDE's being interested in adding contributions here, maybe variations on 'project type' fields or other custom status markers?

packages/core/src/browser/window/window-title-updater.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    protected updateTitleWidget(widget?: Widget): void {
+        let activeEditorLong: string | undefined;
+        let activeEditorMedium: string | undefined;
+        let activeEditorShort: string | undefined;
+        let activeFolderLong: string | undefined;
+        let activeFolderMedium: string | undefined;
+        let activeFolderShort: string | undefined;
+        if (Navigatable.is(widget) && widget.getResourceUri()) {
+            const uri = widget.getResourceUri()!;
```
colin-grant-work, line 47:
```suggestion
        const uri = NavigatableWidget.getUri(widget);
        if (uri) {
```

packages/core/src/browser/window/window-title-service.ts
msujew: I've made the `WindowTitleService` a bit more lenient in what it accepts. Downstream apps can now define custom parts 👍

colin-grant-work (commented):
This looks good to me. There are just two things I wanted to check on, and then a couple of odd behaviors associated with them.

- It looks like for the `activeEditor` fields, all widgets are eligible, main area and otherwise. Is that intentional, or did you want to restrict to editors or main-area widgets (this is trickier for us than VSCode, so I'm happy with whatever policy, as long as the behavior is as expected).
- In Electron, when the window is no longer the active window (e.g. alt-tab to a browser window), `activeEditor` fields disappear - I guess that the `activeWidget` is set to `undefined` in that case.
- Because `activeEditor` fields reflect all widgets, there are a number of quirks.
    - In the Theia workspace, the title for the navigator tree is `theia` so you can end up with this:

`${activeEditorLong}${separator}${activeEditorShort}${separator}${rootName}`  
image

msujew:
@colin-grant-work Thanks for the feedback! I refactored some stuff, so that only the `main` area is considered for title changes. This should result in less unexpected behavior. Also correctly handles the case when the electron window loses focus.

vince-fugnitto (commented):
I haven't tested thoroughly yet but I noticed that the menu does not update properly when a view is opened and closed.

image

For example, in this use-case I:
- opened the preferences-view
- the menu displays `Preferences - theia - Theia Electron Example`
- closed the preferences-view
- the menu persists with `Preferences - theia - Theia Electron Example`

msujew:
@vince-fugnitto It seems like `onDidRemoveWidget` fires a few frames before the last widget of the main area is actually removed, which is why it stays in the title. I added special handling for that case 👍

colin-grant-work:
Another odd sequence:

 - Open Preferences -> Preferences in title 👍 
 - Open Explore -> Preferences in title 👍 
 - Single-click a file to open preview -> Preferences still in title ❓ 
 - Change OS-window focus (for example, switch to a browser to write this comment :-)) -> Editor in title 👍 ❓

colin-grant-work:
> Instead of trying to rely on unreliable events, I've changed the code to query the current main area widget every 33ms.

I'll take a look, but I'm a little nervous about this premise. In general, the cure for unreliable events shouldn't be timeouts but fixing the events so that they're reliable.

msujew:
@colin-grant-work I looked into it again (and rebased everything) and I might need to clarify my previous statement:

The events that are currently available on the `ApplicationShell` (`onDidAddWidget`/`onDidRemoveWidget`/`onDidChangeActiveWidget`) are reliable in the way they are usually used. I.e. identifying if anything has changed. However, we don't try to identify all changes, but only the ones that are relevant to the `main` area of the app. These events don't allow us to filter on that (as in: we can't identify whether the widget which was changed/removed was previously in the main area) and so they are not usable for this feature. The only reliable way to actually identify what the current main area widget is, is to use `setInterval` in a tight loop.

So with that said, do you mind giving this another review?

colin-grant-work:
@msujew, I will take a look at this by the end of the week.

colin-grant-work:
> The only reliable way to actually identify what the current main area widget is, is to use `setInterval` in a tight loop.

I can't embrace this approach. I don't really think it's going to be a significant performance drain - the execution should be quite quick - but if it's true that that's the only option, it does mean that our event system is not as informative as it needs to be for quite a straightforward use case. I will do some looking at whether I see an alternative.

colin-grant-work:
It looks like code like [this](https://github.com/colin-grant-work/theia/commit/d686c0ac250c5132f8b87a7f9d7a904aa9efc503) handles the edge cases pretty well. See what you think and use it if it seems useful.

msujew:
@colin-grant-work Thanks a lot for the clue to look into the dock panel events. That works perfectly even without the interval 👍 

Note that I still had to use the `Saveable.isDirty` check when the main widget changes. Without it, the title wouldn't pick up on switches to an already dirty widget.

colin-grant-work (approved):
Consistently updates the title according to the user's preference. 👍

PR #10917, opened 2022-03-23
Keep current configuration when executing a dynamic configuration

alvsan09:
#### What it does

Starting a session from a dynamic debug configuration left it selected as `current` in the `DebugConfigurationManager`, however this selection was not reflected in the debug view since dynamic configurations are not found under the `launch.json` file.

Therefore before this change, starting a debug session by clicking the `start` button under the debug view may start a previously run dynamic configuration instead of the one reflected in the GUI.

debug_started_wrong_config

This change fixes the above problem by saving the `current` configuration before the start of a dynamic debug session and
restores it right after the session starts.

#### How to test

- Having two `.ts` program files e.g. A and B
  - Define a corresponding launch entry (under the `launch.json`) for each of them (the configurations are arbitrary as long as you have two file programs that are ready for debugging).
    - Example [launch.json](https://gist.github.com/alvsan09/a040d742c4fbf82e485328a87b140940)
  - Start B, make sure the debug session starts properly
  - Start A, make sure the debug session starts properly
  - Select the Editor for B
  - Using the command palette call the command `debug ', NOTE: The last space is important to trigger the command
  - Select the entry dynamic configuration entry: 'Run Current File' 
  - Make sure a debug session is started for B and then stop the session
  - The debug view shall still indicate A as the currently selected configuration
  - From the debug view click the `start` button and make sure that A is started (and not B as before the fix)
  

debug_started_wrong_config_fix

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-prefix-configuration.ts
+++ b/packages/debug/src/browser/debug-prefix-configuration.ts
@@ -135,7 +135,7 @@ export class DebugPrefixConfiguration implements CommandContribution, CommandHan
             for (const configuration of dynamicConfigurations) {
                 items.push({
                     label: configuration.name,
-                    execute: () => this.runConfiguration({ configuration })
+                    execute: () => this.runDynamicConfiguration({ configuration })
                 });
             }
         }
@@ -151,6 +151,14 @@ export class DebugPrefixConfiguration implements CommandContribution, CommandHan
     protected runConfiguration(configuration: DebugSessionOptions): void {
         this.debugConfigurationManager.current = { ...configuration };
         this.commandRegistry.executeCommand(DebugCommands.START.id);
+    }
+
+    /**
+     * Execute the debug start command without affecting the current debug configuration
+     * @param configuration the `DebugSessionOptions`.
+     */
+    protected runDynamicConfiguration(configuration: DebugSessionOptions): void {
+        this.commandRegistry.executeCommand(DebugCommands.START.id, configuration);
     }
 
     /**


colin-grant-work:
@alvsan09, could you add some detail to what you mean by this?
> define a corresponding launch entry for each of them.

What should the launch entry look like?

alvsan09:
@colin-grant-work  for the question.
> What should the launch entry look like?

The configurations are not really special, as long as you have two file programs ready to be debugged from an entry in your `launch.json`, but just in case here goes an example:
[launch.json](https://gist.github.com/alvsan09/a040d742c4fbf82e485328a87b140940)

I have updated the instructions to point to this gist file as well !

packages/debug/src/browser/debug-prefix-configuration.ts
```diff
@@ -148,9 +155,9 @@ export class DebugPrefixConfiguration implements CommandContribution, CommandHan
      *
      * @param configuration the `DebugSessionOptions`.
      */
-    protected runConfiguration(configuration: DebugSessionOptions): void {
+    protected async runConfiguration(configuration: DebugSessionOptions): Promise<void> {
         this.debugConfigurationManager.current = { ...configuration };
-        this.commandRegistry.executeCommand(DebugCommands.START.id);
+        await this.commandRegistry.executeCommand(DebugCommands.START.id);
```
colin-grant-work, line 160: It looks like a more direct strategy might be to modify this code. The `DebugCommands.Start` command accepts an optional argument `DebugSessionOptions` to be run, and falls back to `debugConfigurationManager.current` if none is supplied. Given that, I believe that passing `configuration` as a second argument to `executeCommand` should successfully run the configuration without modifying the `current` field of the `debugConfigurationManager`.

alvsan09: Good catch Colin!
Since there are two different behaviours when exectuing 'dynamic' vs 'non dynamic' configurations, 
I created a new function for 'dynamic', I could inline this call rather than creating a new function although I think it's clearer this way.
Comments are very welcome !

colin-grant-work (approved):
Looks good to me. Running a dynamic configuration no longer affects the selection in the debug widget, but does successfully run the config. 👍

alvsan09:
Re-based to latest master

vince-fugnitto (approved):
I confirmed that both the debug-view, and statusbar reflect the last proper debug configuration executed.

Issue #10922, opened 2022-03-24
monaco: adjust styling of the peek widget


vince-fugnitto:
#### Bug Description:

Following https://github.com/eclipse-theia/theia/pull/10736 we should adjust the styling of the peek widget.
The close button in general is the issue I notice following the upgrade:

![image](https://user-images.githubusercontent.com/40359487/159929218-6b7920b9-a33b-469b-a58d-d4fd29783b15.png)

### Steps to Reproduce:

1. start the application using theia as a workspace
2. open a typescript file
3. trigger `peek > peek definition`
4. notice the styling

PR #10544, opened 2021-12-15
quick-input: alter styling of codicon to resemble vscode

Archie27376:
#### What it does
Fixes #10389

This commit alters the appearance of the codicons that show in the `Run Task` menu to be similar to vscode.
The color has been changed from blue to one that follows the present color theme.
The background of the icon when hovering over is now a square and centered around the icon to be more aesthetically pleasing.

#### How to test
- Start the application using theia as a workspace
- Execute the command `Run Task` using F1
- Hover over each task to notice the tail icon colored correctly with the theme
- Hover over the tail icon for the hover background

#### Review checklist

- [X] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #10923, opened 2022-03-24
monaco: fix quick-input-list styling

vince-fugnitto:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/10922#issue-1179539726.

The pull-request removes superfluous styling from #10544 which caused issues in monaco after the latest upgrade:

1. fixes the peek widget styling

image

2. fixes the styling of codicons as tail icons in the quick-input
 
https://user-images.githubusercontent.com/40359487/159934055-a4b172aa-f6b1-430c-82b0-ea9b87dd8c27.mp4

#### How to test

1. confirm that the `peek` widget styling is correct - (ex: `peek definition`)
3. confirm that codicons in the quick-input menu work well (ex: `run task...`)

#### Review checklist

#### Reminder for reviewers


msujew (commented):
I noticed that VSCode applies the padding of the `.action-item` class differently than we do. The `action-item` padding is actually applied to the `action-label` class, so we apply padding to codicons inside of monaco twice. We should probably add another rule to account for this:

```
.action-item .action-label {
  padding: none;
}
```

Notice how the icon has no rounded border due to the additional padding:

image

vince-fugnitto:
@msujew thanks! It seems like the hover of the codicons in the quick-input menu have bigger issues so I'll investigate what is needed there.

vince-fugnitto:
@msujew I'm struggling a bit with the css in the quick-input for the codicons (`action-item`), I'm wondering if you would have more luck especially since you nicely improved our tabbars with the new styling.

msujew:
@vince-fugnitto I can see why you're struggling, using `action-item` to implement the hover behavior was incorrect in the first place, so I guess that's on me :)

[Here's a patch](https://github.com/eclipse-theia/theia/files/8346496/action-label.txt) you can apply to fix the behavior. It changes our usage of `action-item` to be `action-label`. That automatically fixes the issues in monaco.

vince-fugnitto:
@msujew thank you for your help! It definitely looks much better, I had confirmed the peek, toolbar, global toolbar and the quick-input menus with your changes.

msujew (commented):
I noticed a minor issue with codicons in the quick-input. Current:

image

How it should look like:

image

Notice the top and left padding which is set to 0px due to [this css rule](https://github.com/eclipse-theia/theia/blob/b1bd7bd528c257652f29451d168b750840c22e45/packages/monaco/src/browser/style/index.css#L166-L169). I think after removing it, there's another, additional rule needed:

```
.quick-input-list .quick-input-list-entry-action-bar .action-label.codicon {
  padding: 2px; /* monaco's own css rules use 0px 2px 2px 2px here which removes the left padding */
}
```

Everything else looks good 👍

vince-fugnitto:
@msujew thanks, did you notice that the top-padding for the hover is still off after the changes?

msujew:
@vince-fugnitto Yeah, seems like the monaco rule has higher precedence. I tested it using this rule and this seems to work:

```css
.quick-input-list .quick-input-list-entry-action-bar .action-label.codicon {
  color: var(--theia-foreground) !important;
  padding: 2px !important;
}
```

And replaced [this rule](https://github.com/eclipse-theia/theia/blob/b1bd7bd528c257652f29451d168b750840c22e45/packages/monaco/src/browser/style/index.css#L166-L169).

msujew (approved):
Thanks, looks good to me!

- [x] Quick input items (`Task: Run task...`, `Open recent workspace`)+
- [x] Quick peek window
- [x] The other tab bar icons still work correctly

PR #10930, opened 2022-03-25
Attach Open Editors after Navigator

### LINKED ISSUE #10928: InitiallyCollasped property isn't working on Open Editors widget
### Steps to Reproduce:

1. Open the file navigator widget
2. Observe that the Open Editors "subwidget" is expanded by default

This occurs even though the `initiallyCollapsed` option is set to true when creating the widget.
https://github.com/eclipse-theia/theia/blob/b1bd7bd528c257652f29451d168b750840c22e45/packages/navigator/src/browser/navigator-widget-factory.ts#L46

### Additional Information

The Open Editors widget has been a UX pain since it takes up half of the navigator widget when the IDE starts and it's pretty confusing to first-time users.

It would be great if this can be controlled through a preference (whether the widget is initially collapsed, hidden or expanded) and I would happily submit a PR to resolve this (with some guidance).

I tried to work around this with some container rebinding, but with no luck and I would appreciate some insight into why the `initiallyCollapsed` property does not work for this widget.

- Operating System: Mac
- Theia Version: 1.23.0

colin-grant-work:
#### What it does

Fixes #10928 by adding the open editors widget after the navigator. The `order` field should ensure that the open editors widget appears before the navigator, and adding it second ensures that it is not automatically expanded as the sole child of the `ViewContainer`.

#### How to test

1. Run `Reset Workbench Layout`
2. Open the Navigator view container
3. Observe that the open editors widget is collapsed and above the navigator tree.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/view-container.ts
+++ b/packages/core/src/browser/view-container.ts
@@ -1036,7 +1036,7 @@ export class ViewContainerPart extends BaseWidget {
     set collapsed(collapsed: boolean) {
         // Cannot collapse/expand if the orientation of the container is `horizontal`.
         const orientation = ViewContainer.getOrientation(this.node);
-        if (this._collapsed === collapsed || orientation === 'horizontal' && collapsed) {
+        if (this._collapsed === collapsed || (orientation === 'horizontal' && collapsed)) {
             return;
         }
         this._collapsed = collapsed;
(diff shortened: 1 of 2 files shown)


vince-fugnitto (approved):

msujew (approved):
Yep, also works well for me 👍

Issue #10944, opened 2022-03-29
Support indeterminate progress notifications in message service


planger:
Clients can [report progress via the message service](https://theia-ide.org/docs/message_service/#progress-reporting), which then would show up as follows:
![message-service-progress-reporting](https://user-images.githubusercontent.com/588090/160600304-cdb6f948-b20d-46b2-8b06-914dfabb120f.gif)

However, also as can be seen in the first notification state "Doing something", the notification doesn't show any progress indication if there is no `work` progress being specified. Thus, the notification can't be distinguished from plain info notifications. However, reporting progress with indeterminate progress is often useful, e.g. when dealing with external processes.

### Feature Description:

I suggest to show an indeterminate progress indication if clients report progress without specifying a work progress. This is also in line with VS Code, which shows an indeterminate progress indicator if clients don't specify progress.

The following should show indeterminate progress:
![Peek 2022-03-29 13-22](https://user-images.githubusercontent.com/588090/160600723-5ca8b5eb-3182-4512-9061-b7689bd2fb2e.gif)
```typescript
this.messageService
    .showProgress({
        text: 'Starting to report indeterminate progress',
    })
    .then(progress => {
        // do something
        progress.report({
            message: 'First step completed',
        });
        // do something
        progress.report({
            message: 'Next step completed',
        });
        // do something
        progress.report({
            message: 'Complete',
        });
        // do something
        progress.cancel();
    });
```

And the following should report actual progress:
![Peek 2022-03-29 13-21](https://user-images.githubusercontent.com/588090/160600576-1474ca62-7ee7-48c2-81d9-e62b3b1a7eac.gif)
```typescript
this.messageService
    .showProgress({
        text: 'Starting to report progress',
    })
    .then(progress => {
        // do something
        progress.report({
            message: 'First step completed',
            work: { done: 25, total: 100 }
        });
        // do something
        progress.report({
            message: 'Next step completed',
            work: { done: 60, total: 100 }
        });
        // do something
        progress.report({
            message: 'Complete',
            work: { done: 100, total: 100 }
        });
        // do something
        progress.cancel();
    });
```

planger:
I'll submit a PR for that soon.

PR #10945, opened 2022-03-29
Support indeterminate progress notifications

planger:
#### What it does

  * Default to `progress` notification type if `showProgress()` is used instead of overwriting it with `info`
  * Pass `progress` notification type through to view to make it aware that this is a progress without having to specify a value for the progress
  * Show indeterminate progress animation if notification type is `progress` but no numeric `progress` value is specified
  * Use `info` icon by default in view for `progress` notifications (this was implicitly done before too because we overwrote the notification type with `info` if the client didn't specify one
  * Add sample menu contributions to document and test progress notifications with determinate and indeterminate progress

Fixes https://github.com/eclipse-theia/theia/issues/10944

#### How to test

Use the added sample commands "Sample Command With Indeterminate Progress Message" and "Sample Command With Progress Message". See also their handler implementation in `sample-menu-contribution.ts`, in particular how one reports no progress (indeterminate) and the other one does.

With progress reporting:
Peek 2022-03-29 13-21

Without progress reporting:
Peek 2022-03-29 13-22

#### Other Observations

  * Theoretically you can specify a message type with `messageService.showProgress({ type: MessageType.Error, ...})` even though the argument type `ProgressMessage` dictates the optional `type` parameter to be of type `MessageType.Progress`. If clients still set another type, such as `MessageType.Error` this would affect the icon being shown in the notification. This PR doesn't change that behavior, but we only show an indeterminate progress, if the message type is `MessageType.Progress`.
  * Reporting a progress `{ done: 0, total: 100 }` is equal to specifying no progress, which is after this PR `indeterminate`. This might be surprising, but I felt it is fine as no progress bar would otherwise be visible anyway. This is because of the way how a progress update is handled in `NotificationManager#toPlainProgress()`.
  * Clients can't switch for existing notifications from a determinate progress to an indeterminate progress, because `NotificationManager#reportProgress(...)` falls back to previous progress message, if the new one is undefined. I guess this behavior is fine, because otherwise clients would be forced to always provide a `work` in their update, even if they just want to set a new message.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/message-service.ts
+++ b/packages/core/src/common/message-service.ts
@@ -195,12 +195,13 @@ export class MessageService {
         const report = (update: ProgressUpdate) => {
             this.client.reportProgress(id, update, message, cancellationSource.token);
         };
+        const type = message.type ?? MessageType.Progress;
         const actions = new Set<string>(message.actions);
         if (ProgressMessage.isCancelable(message)) {
             actions.delete(ProgressMessage.Cancel);
             actions.add(ProgressMessage.Cancel);
         }
-        const clientMessage = { ...message, actions: Array.from(actions) };
+        const clientMessage = { ...message, type, actions: Array.from(actions) };
         const result = this.client.showProgress(id, clientMessage, cancellationSource.token);
         if (ProgressMessage.isCancelable(message) && typeof onDidCancel === 'function') {
             result.then(value => {
--- a/examples/api-samples/src/browser/menu/sample-menu-contribution.ts
+++ b/examples/api-samples/src/browser/menu/sample-menu-contribution.ts
@@ -29,6 +29,14 @@ const SampleCommand2: Command = {
     id: 'sample-command2',
     label: 'Sample Command2'
 };
+const SampleCommandWithProgressMessage: Command = {
+    id: 'sample-command-with-progress',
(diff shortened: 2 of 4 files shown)


packages/core/src/common/message-service.ts
```diff
@@ -195,12 +195,13 @@ export class MessageService {
         const report = (update: ProgressUpdate) => {
             this.client.reportProgress(id, update, message, cancellationSource.token);
         };
+        const type = message.type ? message.type : MessageType.Progress;
```
msujew, line 198: Nitpick:
```suggestion
        const type = message.type ?? MessageType.Progress;
```

msujew (approved):
LGTM!

- [x] Normal progress notifications with `progress` are displayed as usual
- [x] Progress notifications without a `progress` property are displayed as in VSCode
- [x] Normal notifications continue to work as expected

packages/core/src/common/message-service.ts
planger: Thanks for the fast feedback and your suggestion!
I applied this change and also used the ?? operator in `notification-component.tsx` with the newly force-pushed commit 9f7c090.

PR #10951, opened 2022-03-29
Use same tempdir for user and system compressed plugin files

colin-grant-work:
#### What it does

On present `master`, if you put a zipped `.vsix` file in your user plugin directory (`~/.theia/extensions`), that plugin will not be loaded successfully. This is due to the behavior of our `PluginDeployerFileHandler` implementations, which attempt to decompress into the user plugin folder for user plugins but back off when they find the existing (compressed) file. This PR modifies that behavior to use the same temporary folder for all plugin decompressions.

#### How to test

1. Put a compressed `.vsix` file in the user plugin area. (e.g. [quick-pick-align-0.0.1.zip](https://github.com/eclipse-theia/theia/files/8365758/quick-pick-align-0.0.1.zip))
2. Start the application.
3. On `master`, you see messages like this and the plugin is not loaded (e.g. no `Show quick pick` command):
```
root ERROR Failed to load plugin dependencies from '.../.theia/extensions/quick-pick-align-0.0.1.vsix' path Error: ENOTDIR: not a directory, open '.../.theia/extensions/quick-pick-align-0.0.1.vsix/package.json'
```

4. With this code, you should eventually see a message like the one below and the plugin should work in the application
```
root INFO Resolved "<plugin-file-name>" to a VS Code extension "<plugin-id>@0.0.1" with engines: { vscode: '^1.65.0' }
```

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext-vscode/src/node/plugin-vscode-file-handler.ts
+++ b/packages/plugin-ext-vscode/src/node/plugin-vscode-file-handler.ts
@@ -54,11 +54,7 @@ export class PluginVsCodeFileHandler implements PluginDeployerFileHandler {
     }
 
     protected async getExtensionDir(context: PluginDeployerFileHandlerContext): Promise<string> {
-        let extensionsDirUri = this.systemExtensionsDirUri;
-        if (context.pluginEntry().type === PluginType.User) {
-            extensionsDirUri = await this.environment.getExtensionsDirUri();
-        }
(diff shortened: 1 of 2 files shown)


vince-fugnitto (approved):
I confirmed that the `.vsix` added to `~/.theia/extensions` works now with the changes, and also confirmed that other extensions such as installed through the view or packs continue to work as expected.

PR #10955, opened 2022-03-30
Add fragment to scheme uris to align to vscode api

westbury:
#### What it does
This PR fixes a problem where the codelens link for showing schema source does not show the schema. The problem is that the https URL of the schema is expected to be specified as the fragment of the schema uri. This change brings Theia into line with vscode and adds the fragment.

#### How to test
- Install a plugin with schema validation, for example Redhat's YAML plugin.
- Open a yml file and configure a schema.
- Click on the codelens link at the top of the yml file.
expected: the schema should open in an editor.
actual without this fix: an error is shown because an attempt is made to fetch the schema contents using json-schema rather than https.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/workspace.ts
+++ b/packages/plugin-ext/src/plugin/workspace.ts
@@ -277,7 +277,11 @@ export class WorkspaceExtImpl implements WorkspaceExt {
     }
 
     async $provideTextDocumentContent(documentURI: string): Promise<string | undefined> {
-        const uri = URI.parse(documentURI);
+        let uri = URI.parse(documentURI);
+        if (uri.scheme !== 'http' && uri.scheme !== 'https') {
+            const httpScheme = URI.parse(documentURI).with({ scheme: 'https' });
+            uri = uri.with({ fragment: httpScheme.toString() });
+        }
         const provider = this.documentContentProviders.get(uri.scheme);
         if (provider) {
             return provider.provideTextDocumentContent(uri, CancellationToken.None);


colin-grant-work:
For interested reviewers, the testing step above 'configure a schema' can be completed by adding something like this at the top of a yaml file:
```yaml
# yaml-language-server: $schema=https://www.graphql-code-generator.com/config.schema.json
```

colin-grant-work:
@westbury, could you point to where VSCode performs this transformation? Looking through the `provideDocumentTextContent` cycle, I didn't see anything that looked quite like this. I do see that the YAML language server [creates a `URI` in this style](https://github.com/redhat-developer/yaml-language-server/blob/cb688a6a7ff8612d78c85a083515c1f6b2a0522b/src/languageservice/services/yamlCommands.ts#L19-L24). Could it be that somehow Theia is failing to respect that URI's integrity as it's passed around from language server to command executor?

colin-grant-work:
@westbury, I think I've tracked down the source of the problem. The URI that the plugin passes in does have the expected form (e.g. `json-schema://www.graphql-code-generator.com/config.schema.json#https%3A%2F%2Fwww.graphql-code-generator.com%2Fconfig.schema.json`), but Theia code removes it as part of the editor opening process:

https://github.com/eclipse-theia/theia/blob/51443155c8c860c7d517e2f692336e149a8b4988/packages/core/src/browser/navigatable.ts#L31-L37

I believe the correct solution would be to remove the `withoutFragment` calls here and adjust our code as necessary. I believe that the original intent was to ensure that the URI could be used as an identifier of the document without being affected by fragments intended to indicate selection range, for example. It would likely be possible to keep the original URI alive while still achieving that goal.

westbury:
@colin-grant-work Undoubtedly the correct way to fix this. I will make the changes.

colin-grant-work:
Interesting, this behavior (treating the fragment as an identifier) is counter to the specification of URI's - but VSCode's stuck with it.

https://github.com/microsoft/vscode/issues/87332

PR #10956, opened 2022-03-30
Allow to configure default locale

msujew:
#### What it does

I originally became aware of this due to [this community post](https://community.theia-ide.org/t/how-to-set-default-locale/2312).

Since we don't have access to the `FrontendApplicationConfigProvider` in the `nls.ts` file, we need to set the `nls.locale` during the `nls-loader` routine, which requires the `Object.assign` workaround.

#### How to test

0. Delete the `localeId` value from your local storage.
1. Download and install a language pack of your choice.
2. Set the `defaultLocale` of the frontend config to the language code of your language pack (i.e. `fr` for french)
3. Confirm that the application starts in the configured locale without having it selected beforehand using the `Configure Display Language` command.
4. Configuring the language manually (with another language pack) should override the default locale.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-package/src/application-props.ts
+++ b/dev-packages/application-package/src/application-props.ts
@@ -64,7 +64,8 @@ export namespace FrontendApplicationConfig {
         applicationName: 'Eclipse Theia',
         defaultTheme: 'dark',
         defaultIconTheme: 'none',
-        electron: ElectronFrontendApplicationConfig.DEFAULT
+        electron: ElectronFrontendApplicationConfig.DEFAULT,
+        defaultLocale: ''
     };
     export interface Partial extends ApplicationConfig {
 
@@ -95,6 +96,13 @@ export namespace FrontendApplicationConfig {
          * Defaults to `ElectronFrontendApplicationConfig.DEFAULT`.
          */
         readonly electron?: ElectronFrontendApplicationConfig.Partial;
+
+        /**
+         * The default locale for the application.
+         *
+         * Defaults to "".
+         */
+        readonly defaultLocale?: string;
     }
 }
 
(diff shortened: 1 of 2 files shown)


dev-packages/application-package/src/application-props.ts
```diff
@@ -64,7 +64,8 @@ export namespace FrontendApplicationConfig {
         applicationName: 'Eclipse Theia',
         defaultTheme: 'dark',
         defaultIconTheme: 'none',
-        electron: ElectronFrontendApplicationConfig.DEFAULT
+        electron: ElectronFrontendApplicationConfig.DEFAULT,
+        defaultLocale: ''
```
vince-fugnitto, line 68: @msujew I'm a bit confused by this change, isn't something like `defaultLocale` specific to the end-user? 
I'm not sure I follow as to why application would set the `defaultLocale` for all their users as an application prop.

msujew: As outlined in the community post, I can imagine it to be useful when deploying your app in a certain region, such as China, were you can assume that your user base is more likely to understand Chinese than English. Also, special installers might modify the `defaultLocale` (as in: Install this Theia app in English, German, French, etc.), were this change comes in handy. 

But otherwise I agree, it should be left to the choice of the end-user.

vince-fugnitto (approved):
I confirmed that :+1: 
- [x] the `defaultLocale` of `''` behaves like today
- [x] setting the `defaultLocale` to `'fr'` works, and the app starts in french
- [x] setting an invalid `defaultLocale` will default to `''`
- [x] configuring a language after setting a `defaultLocale` works well

Issue #7797, opened 2020-05-12
Theia Electron Application - Proxy settings


davide-sergi:
I noticed that in vscode there are a number of predefined settings to act on (please refer to [this file](https://github.com/microsoft/vscode/blob/03a7223ee2a086bbcf028801da9370d5deff5093/src/vs/platform/request/common/request.ts#L76-L106) ), so it's possible to have something like that for theia.

### Feature Description:
In my mind, Theia should be able to get proxy configurations from its settings.json file or at least directly from OS proxy configuration (as happen for Postman electron app - please refer to [this link](https://learning.postman.com/docs/postman/sending-api-requests/proxy/) ).

r10r:
@davide-sergi  Hi, did you manage to configure the http proxy ?

I'm trying to install an extension within theia which is running in kubernetes behind a proxy and I'm getting the following error:

> ERROR Failed to resolve plugins from 'vscode:extension/golang.go' Error: connect ECONNREFUSED 198.41.30.195:443

The http proxy environment variables are properly set.

luettmaSICKAG:
We need Proxy-Support as well. The setting to use the system certificate is also important.

I think it would make sense to:
- also to support a `NO_PROXY` settings (or use the environment variable)  
  → No proxy required for requests to `localhost` or services in the company network
- Apply the same proxy settings to the Theia Extension and Theia plugins/VSCode extensions


Microsoft has a library to handle the proxy settings and patch the http client: https://github.com/Microsoft/vscode-proxy-agent  
I couldn't get it running in a Theia extension, though. There were some NAPI issues (possibly due to an incompatible node version).

PR #10698, opened 2022-02-02
Implement backend to electron IPC channels

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/2162

Uses a similar architecture to the existing Electron Main <-> Frontend communication layer over `ipcRenderer` and `ipcMain`. Contains additional changes to deal with the `--no-cluster` argument by using a special pipe object.

For testing purposes I included a `TestConnection` interface which returns the titles of all opened electron windows on startup to the backend server.

In a second version of this we could mirror the proposed classes/APIs here to also enable electron -> backend requests (this PR only enables backend -> electron requests), since we create only a single backend server for each electron main.

_Note:_ My use case for this change is that I plan to add proxy support for all requests in Theia. In particular, requests from the backend running in Electron should try to resolve to the Chromium proxy (which automatically picks up the OS proxy) in case no proxy is specified using the apps' preferences. For this, we need some kind of way to communicate with the electron process from the backend server process.

#### How to test

1. Start the Electron app.
2. Search for `Titles:` in the console. The console should contain the `title` of the newly opened Electron window.
3. Repeat with the `--no-cluster` argument enabled.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #10958, opened 2022-03-30
Support http proxies

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/7797

Adds general proxy support to the whole application (backend+frontend) by using a `RequestService` which comes with automatic proxy support. Other services and downstream applications are encouraged to use that service whenever requests to non-origin or non-localhost web-services are performed.

Also patches the `http` and `https` node.js imports for the extension host. This enables extensions to perform requests through proxies as well.

For electron there's currently a large TODO: Automatic OS proxy detection doesn't work yet, since that requires access to the electron-main process (see also https://github.com/eclipse-theia/theia/pull/10698).

#### How to test

Generally, testing this PR requires setting up firewall rules which block all outgoing traffic and configuring a local proxy rule. I used [Fiddler Classic](https://www.telerik.com/fiddler/fiddler-classic) (Windows only?) which sets itself up at `http://localhost:8888`. I assume Mac and Linux users have a simpler time configuring such a proxy.

Also, these tests should preferably performed using the Electron version. The browser version will automatically use the OS proxy for stuff like fetching vsx search results, whereas the electron version will fail if the proxy is not configured.

1. Open the vsx registry and search for `java`. Nothing should appear if outbound traffic is blocked.
2. Search in the preferences for `proxy` and set the `Http: Proxy` preference to your local proxy url.
3. Confirm that re-searching displays results (frontend request) and allows to install the extension (backend request).
4. Install the [`proxy-test` extension](https://github.com/eclipse-theia/theia/files/8380275/proxy-test-0.0.1.zip).
5. Execute the `Hello World Request` command. It will send a request to `https://www.google.com` and display the error message/success code.
6. Change the `Http: Proxy support` preference. `override`, `fallback` and `on` should all provide the same success result, while `off` should display an error, since the extension proxy support has been turned off.

---

There are now proxy settings for fetching packages from ovsx registries using the `theia download:plugins` command.

0. Delete your `plugins` folder.
1. Executing `theia download:plugins` should result in an error, since the outgoing download requests are blocked
2. Adding the `--proxy-url <url>` parameter should enable successful download of these extensions.
 
#### Review checklist

#### Reminder for reviewers


dev-packages/cli/package.json
```diff
@@ -32,6 +32,7 @@
   "dependencies": {
     "@theia/application-manager": "1.23.0",
     "@theia/application-package": "1.23.0",
+    "@theia/core": "1.23.0",
```
msujew, line 35: I'm not happy with this dependency in our `cli`. It shouldn't really influence anything, since any project which uses `@theia/cli` probably also has `@theia/core` in there, but its still quite nasty.

We need this currently to use the `BackendRequestService` in the `ovsx` and `cli` package, but I could refactor this into its own `dev-package`. WDYT?

vince-fugnitto: > I'm not happy with this dependency in our cli

@msujew I agree, if possible I would like to avoid coupling `core` into `cli`. The refactor you mentioned seems good to me.

msujew: Thanks for the Input Vince, I've added `@theia/request-service` which is used by `@theia/ovsx-client`, `@theia/cli` and `@theia/core`.

packages/core/src/browser/request/browser-request-service.ts
```diff
@@ -0,0 +1,167 @@
...
+            outer: for (const k of Object.keys(options.headers)) {
+                switch (k) {
+                    case 'User-Agent':
+                    case 'Accept-Encoding':
+                    case 'Content-Length':
+                        // unsafe headers
+                        continue outer;
+                }
+                xhr.setRequestHeader(k, options.headers[k]);
+            }
```
colin-grant-work, line 153: Is the `outer` label necessary? Since `continue` doesn't apply to `switch` statements, I think that it will work without the label.

packages/core/src/electron-node/request/electron-backend-request-service.ts
```diff
@@ -0,0 +1,58 @@
...
+                return this.buildProxyUrl(url, proxyHost);
+            }
+        } catch (e) {
+            console.error('Could not resolve electron proxy.', e);
+        }
+        return super.getProxyUrl(url);
+    }
+
+    override async resolveProxy(url: string): Promise<string | undefined> {
+        // TODO: Implement IPC to the backend to access the Electron proxy resolver
```
colin-grant-work, line 43: Will this be possible once #10698 (or something like it) is in place?

packages/preferences/src/browser/abstract-resource-preference-provider.ts
```diff
@@ -157,18 +157,18 @@ export abstract class AbstractResourcePreferenceProvider extends PreferenceProvi
                 this.fileExists = false;
                 return { value: '' };
             });
-        this.readPreferencesFromContent(content.value);
+        await this.readPreferencesFromContent(content.value);
```
colin-grant-work, line 160: There used to be stern warnings in this file that most of its work had to be done synchronously. I believe I've eliminated most of the reasons that used to be true - and written tests to ensure everything is alright in any case (I hope) - but adding so many `async-awaits` still makes me a little nervous. What is the problem that the new `awaits` aim to solve?

packages/plugin-ext/src/hosted/node/plugin-host-proxy.ts
```diff
@@ -0,0 +1,92 @@
...
+    node_module._load = function (request: string): typeof tls | typeof http | typeof https {
+        if (request === 'tls') {
+            return lookup.tls;
+        }
+
+        if (request !== 'http' && request !== 'https') {
+            return original.apply(this, arguments);
+        }
+
+        // Create a shallow copy of the http(s) module to workaround extensions which apply changes to the modules
```
colin-grant-work, line 88:
```suggestion
        // Create a shallow copy of the http(s) module to work around extensions that apply changes to the modules
```

colin-grant-work (commented):
Just a quick glance through the code so far. I'll take a closer look and test it after the release.

packages/preferences/src/browser/abstract-resource-preference-provider.ts
msujew: Yes, I know about that warning, and I was quite hesitant adding all the `awaits` here. However, it didn't seem to break anything and fixed my usecase:

https://github.com/eclipse-theia/theia/blob/68bb5157fc25c946a34f65833758b2f96ba6f744/packages/core/src/browser/request/browser-request-service.ts#L68-L69

In there, we wait for the preference service to be ready before sending out our first http-request. The request might fail otherwise, if the backend's `http.proxy` setting hasn't been configured yet. On Theia `master` code, this series of events would unfold:

1. `await preferenceService.ready` (this starts the async user preference reading process __without awaiting it__)
2. Sending the http-request __without having configured proxy settings__ (missing user preferences).
3. `preferenceService.onPreferenceChanged` is triggered, since the async user preference reading process has finished. This then triggers changes to the proxy settings, which is too late.

colin-grant-work: In general, I'm in favor of the new code, especially if it doesn't break anything - and I've added tests of various parts so that if they don't break, the system _should_ be working as expected. But there's one point I'm not sure of:

> 1. `await preferenceService.ready` (this starts the async user preference reading process __without awaiting it__)

There are two `async` parts to reading preferences: one is reading the actual content, and the other is waiting for a debounced event to fire. In the current code, the `AbstractResourcePreferenceProvider` _does_ await the first part before declaring itself ready:

https://github.com/eclipse-theia/theia/blob/f89e8265ef1c59427bc8160ffcb3b2e5f2fe7d15/packages/preferences/src/browser/abstract-resource-preference-provider.ts#L60-L61

https://github.com/eclipse-theia/theia/blob/f89e8265ef1c59427bc8160ffcb3b2e5f2fe7d15/packages/preferences/src/browser/abstract-resource-preference-provider.ts#L149-L159

The last line of the second snippet there, the call to `readPreferencesFromContent` is not awaited (and isn't currently `async`) so `ready` _is_ resolved before the first event emission. That's deliberate, if undesirable, because there are a number of cases even in the framework where the emission of initial values was used to initialize values in other services. But, even though the event emission is not `await`ed, it should be possible to access accurate values as soon as the `ready` promise resolves: the internal state of the provider is already up to date before the event is emitted.

https://github.com/eclipse-theia/theia/blob/f89e8265ef1c59427bc8160ffcb3b2e5f2fe7d15/packages/preferences/src/browser/abstract-resource-preference-provider.spec.ts#L80-L97

Given that, for your case, I think it should be sufficient to `await` the `ready` promise and then access the values, rather than adding additional `await`s for the events. Or have you observed that the configured values are not available when the `ready` promise resolves?

packages/core/src/electron-node/request/electron-backend-request-service.ts
msujew: Yes, that's what https://github.com/eclipse-theia/theia/pull/10698 was initially meant for.

packages/preferences/src/browser/abstract-resource-preference-provider.ts
msujew: I see, that makes sense. I reverted the preference changes and use the `ready` promise to retrieve the preference values instead of waiting for the `onPreferenceChanged` event 👍

PR #10961, opened 2022-03-30
Add Accessibility API

### LINKED ISSUE #9965: Implement Accessibility API
### Feature Description:
Implement the Accessibility VS Code API, in particular in StatusBarItem and TreeItem. The idea is to set the `aria-label` and and 'role' attributes on the rendered HTML.

see https://code.visualstudio.com/api/references/vscode-api#AccessibilityInformation

sgraband:
#### What it does

Add AccessibilityInformation interface.
Add optional accessibilityInformation to StatusBarItem and TreeItem.
Also added the type to TimelineItem, but will create a follow-up ticket for loading the information into the DOM.
Proposed API:
Add StatusBarItemOptions (containing accessibilityInformation).
Provide window.createStatusBarItem(StatusBarItemOptions).

Update createStatusBarItem handler to handle StatusBarItems.
Added aria-labels for default StatusBarItems.

Resolves #9965.

#### How to test

StatusBarItems:
Check one of the `StatusBarItems`, like the `Toggle Bottom Panel`, in the Accessibilty tab of the Chrome DevTools. The `aria-label` should be recognized correctly.

TreeItems:
To simplify the testing the [tree-view-sample](https://github.com/microsoft/vscode-extension-samples/tree/main/tree-view-sample) vscode-extension could be used. Just add:

```
treeItem.accessibilityInformation = {
	label: 'somValue',
	role: 'someValue'
};
```
[here](https://github.com/microsoft/vscode-extension-samples/blob/92b34c5733e2d1640ad877e9ed4605da04709124/tree-view-sample/src/jsonOutline.ts#L125). And add the extension to Theia. The jsonOutline view tree should then also display the correct label and role.
#### Review checklist

#### Reminder for reviewers


packages/plugin/src/theia-proposed.d.ts
```diff
@@ -734,4 +739,57 @@ export module '@theia/plugin' {
...
+
+        /**
+         * Creates a status bar [item](#StatusBarItem).
+         *
+         * @param options The options of the item. If not provided, some default values
+         * will be assumed. For example, the `StatusBarItemOptions.id` will be the id
+         * of the extension and the `StatusBarItemOptions.name` will be the extension name.
+         * @return A new status bar item.
+         */
+        export function createStatusBarItem(options?: StatusBarItemOptions): StatusBarItem;
```
vince-fugnitto, line 791: @sgraband similar to above, where does this api come from?

packages/plugin/src/theia-proposed.d.ts
```diff
@@ -734,4 +739,57 @@ export module '@theia/plugin' {
...
+         * The alignment of the status bar item.
+         */
+        alignment?: StatusBarAlignment;
+
+        /**
+         * The priority of the status bar item. Higher value means the item should
+         * be shown more to the left.
+         */
+        priority?: number;
+    }
```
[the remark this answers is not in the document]
sgraband: It was part of `1.55.2` see [here](https://github.com/microsoft/vscode/blob/3c4e3df9e89829dce27b7b5c24508306b151f30d/src/vs/vscode.proposed.d.ts#L844-L875). But apparently it was removed again afterwards. How should we handle this?

packages/plugin/src/theia-proposed.d.ts
sgraband: Similar to above. See [here](https://github.com/microsoft/vscode/blob/3c4e3df9e89829dce27b7b5c24508306b151f30d/src/vs/vscode.proposed.d.ts#L877-L888).

packages/plugin/src/theia-proposed.d.ts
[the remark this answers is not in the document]
msujew: I would say that we shouldn't support temporary `vscode.proposed.d.ts` stuff, which hasn't moved into the stable API. Especially if only has been there for a single version and is now unsupported in vscode itself. We can assume that no vscode extension makes use of that API and there's no need to support it.

msujew (commented):
Looking like a good start already :) I was planning on doing some work on a11y as well, but the less for me to do there the better 👍 

I have some minor code suggestions, in addition to what I already said regarding the proposed API stuff (see [comment](https://github.com/eclipse-theia/theia/pull/10961#discussion_r842643631)). I haven't tested the changes yet, will do so later this week.

packages/plugin-ext/src/plugin/plugin-context.ts
```diff
@@ -416,22 +416,35 @@ export function createAPIFactory(
             showInputBox(options?: theia.InputBoxOptions, token?: theia.CancellationToken): PromiseLike<string | undefined> {
                 return quickOpenExt.showInput(options, token);
             },
-            createStatusBarItem(alignmentOrId?: theia.StatusBarAlignment | string, priorityOrAlignment?: number | theia.StatusBarAlignment,
+            createStatusBarItem(optionsOrAlignmentOrId?: theia.StatusBarItemOptions | theia.StatusBarAlignment | string, priorityOrAlignment?: number | theia.StatusBarAlignment,
                 priorityArg?: number): theia.StatusBarItem {
                 let id: string | undefined;
                 let alignment: number | undefined;
                 let priority: number | undefined;
+                let accessibilityInformation: theia.AccessibilityInformation | undefined = undefined;
```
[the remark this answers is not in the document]
sgraband: If we remove the proposed API this method will never receive `AccessibilityInformation` as it is not part of the API.

packages/core/src/browser/status-bar/status-bar.tsx
```diff
@@ -55,6 +57,7 @@ export interface StatusBarEntryAttributes {
     className?: string;
     title?: string;
     style?: object;
+    role?: string;
     onClick?: (e: MouseEvent) => void;
 }
```
vince-fugnitto, line 62: I believe this interface can be removed completely, I'll open a separate pull-request to do a cleanup.

vince-fugnitto (commented):
@sgraband it seems as though the `ariaLabel` is not respected when a plugin defines it when creating the `StatusBarItem`. It always ends up using the tooltip for such a case.

You can confirm with [status-item-0.0.1.zip](https://github.com/eclipse-theia/theia/files/8428513/status-item-0.0.1.zip), I create the item like so (when triggering the `StatusBar Item: Create` command):

```ts
let disposable = vscode.commands.registerCommand('status-item.create', () => {
    const item = vscode.window.createStatusBarItem();
    item.name = 'Test Item';
    item.text = 'Test Text';
    item.accessibilityInformation = { label: 'Test Label', role: 'button' };
    item.show();
});
```

The same is also true for `role`.

vince-fugnitto:
@sgraband I have the following commit that handles the `accessibilityInformation` for the `StatusBarItem` that you can take a look at. It does not require a change to `createStatusBarItem` as it follows the design of other attributes.

- https://github.com/eclipse-theia/theia/commit/26ff86bff349c143467703d32cf5048ca1187550

I also confirmed it works with different use-cases using [status-item-0.0.1.zip](https://github.com/eclipse-theia/theia/files/8428914/status-item-0.0.1.zip)

sgraband:
Good catch and thank you for the commit @vince-fugnitto! I also tested it with the extension you provided and everything seems to work like intended now.

packages/core/src/browser/status-bar/status-bar.tsx
```diff
@@ -221,7 +225,7 @@ export class StatusBarImpl extends ReactWidget implements StatusBar {
         });
         const elementInnerDiv = <React.Fragment>{children}</React.Fragment>;
 
-        return React.createElement('div', { key: id, ...this.createAttributes(entry) }, elementInnerDiv);
+        return React.createElement('div', { key: id, ...this.createAttributes(entry), 'aria-label': entry.ariaLabel ?? entry.tooltip }, elementInnerDiv);
```
vince-fugnitto, line 228: We can revert this line as it is handled already by `createAttributes`.

packages/core/src/browser/status-bar/status-bar.tsx
```diff
@@ -36,6 +36,8 @@ export interface StatusBarEntry {
      * http://fontawesome.io/examples/#animated
      */
     text: string;
+    ariaLabel?: string;
+    role?: string;
```
vince-fugnitto, line 40: Can be removed, we now have `accessibilityInformation`.

packages/plugin/src/theia.d.ts
```diff
@@ -2600,6 +2623,24 @@ export module '@theia/plugin' {
...
+        label: string;
+
+        /**
+         * Role of the widget which defines how a screen reader interacts with it.
+         * The role should be set in special cases when for example a tree-like element behaves like a checkbox.
+         * If role is not specified VS Code will pick the appropriate role automatically.
+         * More about aria roles can be found here https://w3c.github.io/aria/#widget_roles
+         */
+        role?: string;
+    }
```
vince-fugnitto, line 2642: We can remove the duplicate interface, we already have `AccessibilityInformation`  (duplicated at line `2405`).

packages/core/src/browser/status-bar/status-bar.tsx
```diff
@@ -44,20 +44,17 @@ export interface StatusBarEntry {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     arguments?: any[];
     priority?: number;
+    accessibilityInformation?: {
+        label: string,
+        role?: string;
```
vince-fugnitto, line 49: I think we can probably create an `accessibility.ts` file in `@theia/core/src/common` so it can be re-used (eventually also by `timeline`).

vince-fugnitto (commented):
I'll let @msujew complete the review since part of the changes are from the commit I proposed :+1:

msujew (approved):
LGTM!

- [x] Tree elements display the correct accessibility information
- [x] Status items display the correct accessibility information

sgraband:
Thanks for the review @msujew. I added two `accessibilityInformation`s to the `Notification` and `Toggle Bottom panel` status item. I also added the `AccessibilityInformation` type to the common package like @vince-fugnitto suggested.

packages/core/src/common/accessibility.ts
```diff
@@ -0,0 +1,33 @@
+// *****************************************************************************
+// Copyright (C) 2021 STMicroelectronics and others.
```
msujew, line 2: Minor:
```suggestion
// Copyright (C) 2022 STMicroelectronics and others.
```

sgraband: Updated the year.

PR #10977, opened 2022-04-01
Add correct extension mode to plugin context

### LINKED ISSUE #10201: Implement `ExtensionMode`
### Feature Description:
Add support for the `extensionMode` field in the `ExtensionContext` interface. We currently don't support running tests (as far as I know), but we should set "production" and "development" correctly when we run a hosted instance of Theia for a plugin under development.

eneufeld:
Fix #10201

Signed-off-by: Eugen Neufeld <eneufeld@eclipsesource.com>

#### What it does
The plugin context was hardcoded to have the production extensionMode.
Now a flag is passed to mark a plugin as under development when
deploying it. This flag is stored in the plugin metadata and later set in
the plugin. When creating the context this flag is then converted to the
correct extensionMode.

#### How to test
Download this plugin project: [theia-hello-world-plugin.zip](https://github.com/eclipse-theia/theia/files/8407558/theia-hello-world-plugin.zip) .
It contains a built version of the plugin which you should add to your plugins folder. This add a new command which opens a message with `Hello World! Mode: ${mode}` where the mode is the extensionMode. For the deployed plugin the mode should be '1' (Production).
Also add the project to your workspace so that you can start the plugin in a hosted version, by calling `Hosted mode: start instance`. When you trigger the command now in the hosted mode, the message should be `Hello World! Mode: 2` (Development).

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/hosted/node/hosted-plugin-deployer-handler.ts
+++ b/packages/plugin-ext/src/hosted/node/hosted-plugin-deployer-handler.ts
@@ -99,17 +99,17 @@ export class HostedPluginDeployerHandler implements PluginDeployerHandler {
         }
     }
 
-    async deployFrontendPlugins(frontendPlugins: PluginDeployerEntry[]): Promise<void> {
+    async deployFrontendPlugins(frontendPlugins: PluginDeployerEntry[], isUnderDevelopment?: boolean): Promise<void> {
         for (const plugin of frontendPlugins) {
-            await this.deployPlugin(plugin, 'frontend');
+            await this.deployPlugin(plugin, 'frontend', isUnderDevelopment);
         }
         // resolve on first deploy
         this.frontendPluginsMetadataDeferred.resolve(undefined);
     }
 
-    async deployBackendPlugins(backendPlugins: PluginDeployerEntry[]): Promise<void> {
+    async deployBackendPlugins(backendPlugins: PluginDeployerEntry[], isUnderDevelopment?: boolean): Promise<void> {
         for (const plugin of backendPlugins) {
-            await this.deployPlugin(plugin, 'backend');
+            await this.deployPlugin(plugin, 'backend', isUnderDevelopment);
         }
         // rebuild translation config after deployment
         this.localizationService.buildTranslationConfig([...this.deployedBackendPlugins.values()]);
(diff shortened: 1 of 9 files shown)


vince-fugnitto (commented):
@eneufeld thank you for the contribution, please be sure to properly fill out the pull-request template including steps on how to test for reviewers (ideally with an example plugin).

eneufeld:
@vince-fugnitto sorry, I was to fast creating the PR. I updated the description now.

packages/plugin-ext/src/hosted/node/hosted-plugin-deployer-handler.ts
```diff
@@ -131,6 +131,7 @@ export class HostedPluginDeployerHandler implements PluginDeployerHandler {
             }
 
             const metadata = this.reader.readMetadata(manifest);
+            metadata.isUnderDevelopment = isUnderDevelopment;
```
colin-grant-work, line 134: I think that this line suggests that the approach taken here is a little bit against the grain of how the plugin system should work: ideally the plugin reader system should set the metadata, but here `isUnderDelevopment` is grafted onto the metadata after the readers have had their say. Given that, I think the correct approach would be to ensure that whichever `HostedPluginReader` is responsible for setting the metadata of the plugin under development correctly sets this field, rather than passing it through the deployer system.

eneufeld: Thank you for the review and the hints to clean up the code. 
Unfortunately the scanner never has access to the `PluginDeployerEntry`.
I now set a marker inside the `PluginDeployerEntryImpl` based on the `originId` and instead of passing the `isUnderDelevopment` through the deployment system, the value is read from the `PluginDeployerEntry`.
What do you think, is this a viable approach?

packages/plugin-ext/src/main/node/plugin-deployer-entry-impl.ts
```diff
@@ -47,6 +47,7 @@ export class PluginDeployerEntryImpl implements PluginDeployerEntry {
         } else {
             this.resolved = false;
         }
+        this.storeValue('isUnderDevelopment', originId === 'Hosted Plugin');
```
colin-grant-work, line 50: I don't think this needs to be done in the `PluginDeployerEntryImpl`. The `HostedPluginReader`(packages/plugin-dev/src/node/hosted-plugin-reader.ts) can call `storeValue` on the plugin deployer entry that it creates.

packages/plugin-ext/src/hosted/node/hosted-plugin-deployer-handler.ts
colin-grant-work: Another alternative - a bit more attractive, though along the same lines as the separate service approach - would be to implement a `MetadataProcessor` for the development host case. Then, at this stage:

https://github.com/eclipse-theia/theia/blob/9b8bc9eb61573ad7e663587af2ef2043371b3321/packages/plugin-ext/src/hosted/node/plugin-reader.ts#L109-L117

The development host metadata processor could add the notation that the plugin in question is under development.

colin-grant-work (approved):
The changes are sound and correctly set the plugin mode 👍

PR #10981, opened 2022-04-01
Support contributions to the shell layout

msujew:
#### What it does

Allows downstream apps to contribute widgets to the layout without extending the `ApplicationShell` itself. I had the idea while reviewing https://github.com/eclipse-theia/theia/pull/10731, where overriding the `ApplicationShell` to contribute a new widget to the layout seemed a bit overengineered.

So far it only supports adding panels under the top panel, although we could extend this for all areas of the app. Although, I'm not totally sure that's even useful.

Aside from the toolbar, I can imagine that we reuse this API to also add support for banners in the future like in VSCode:

image

#### How to test

The test case for the introduced `ShellLayoutContribution` is the toolbar. Simply confirm that it behaves as usual.

#### Review checklist

#### Reminder for reviewers


msujew:
I'll maybe come back to this later. For now, I don't really have an approach on how to fix the underlying issue, i.e. the rigid shell structure.

Issue #8153, opened 2020-07-08
Have the Default preference settings more visible when selecting a value from the enumeration


lmcbout:
I noticed than the default value for an enumeration preference setting is not well visible to the end-user, I think it would be nice to have something like in VSCode: the display "Default" beside the preference settings for the enumation selection.

### Feature Description:
To see what is like in VSCode
Open VScode
Open the preference setting
Look for: " Debug: Internal Console Options"
There will be 3 choices and the last one has "Default" information (openOnFirstSessionStart    Default)

vince-fugnitto:
I don't see the big benefit of marking a preference enum option as the `default` since I don't think users will care what the default is (they will only care about the current value they set). 

The default value will always be the value displayed unless modified, and a modified value is marked differently for easier visual feedback. In addition, we would need a custom solution for `electron` as we use native menus (for a native feel) unlike vscode so we cannot compare the two at the moment.

PR #10991, opened 2022-04-05
Introduce custom select component

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/8153

Introduces a custom react component for rendering `select` elements. It mirrors mostly how a native html `select` element works and allows for some additional features like element descriptions, details, separators, etc.

Mostly get's used for preference enum rendering, but I've also replaced some other `select` elements throughout the app.

#### How to test

In general, test that the component works just like the VSCode select element for enum preferences, launch debug config selection, etc.

1. Open the preference widget and play around with the different enum preferences. Also try to open preferences which are near the bottom of the app and observe that the dropdown gets inverted.
2. The component should react as expected when clicking on other elements or scrolling
3. The debug toolbar also uses the component. Confirm that every option is always rendered on a single line, expanding the dropdown width as necessary.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/browser.ts
+++ b/packages/core/src/browser/browser.ts
@@ -156,3 +156,42 @@ export function preventNavigation(event: WheelEvent): void {
     event.preventDefault();
     event.stopPropagation();
 }
+
+export function measureTextWidth(text: string | string[]): number {
+    const measureElement = getMeasurementElement();
+    text = Array.isArray(text) ? text : [text];
+    let width = 0;
+    for (const item of text) {
+        measureElement.textContent = item;
+        width = Math.max(measureElement.getBoundingClientRect().width, width);
+    }
+    return width;
+}
+
+export function measureTextHeight(text: string | string[], maxWidth?: number): number {
+    const measureElement = getMeasurementElement();
+    if (maxWidth !== undefined) {
+        measureElement.style.maxWidth = `${Math.ceil(maxWidth)}px`;
+    }
+    text = Array.isArray(text) ? text : [text];
+    let height = 0;
+    for (const item of text) {
+        measureElement.textContent = item;
+        height = Math.max(measureElement.getBoundingClientRect().height, height);
+    }
+    measureElement.style.maxWidth = 'none';
+    return height;
+}
+
+function getMeasurementElement(): HTMLElement {
+    let measureElement = document.getElementById('measure');
+    if (!measureElement) {
+        measureElement = document.createElement('span');
+        measureElement.id = 'measure';
+        measureElement.style.fontFamily = 'var(--theia-ui-font-family)';
+        measureElement.style.fontSize = 'var(--theia-ui-font-size1)';
+        measureElement.style.visibility = 'hidden';
+        document.body.appendChild(measureElement);
+    }
+    return measureElement;
+}
--- a/packages/core/src/browser/browser.ts
+++ b/packages/core/src/browser/browser.ts
@@ -157,8 +157,18 @@ export function preventNavigation(event: WheelEvent): void {
     event.stopPropagation();
 }
 
-export function measureTextWidth(text: string | string[]): number {
-    const measureElement = getMeasurementElement();
+export type PartialCSSStyle = Omit<Partial<CSSStyleDeclaration>,
+    'visibility' |
+    'display' |
+    'parentRule' |
+    'getPropertyPriority' |
+    'getPropertyValue' |
+    'item' |
+    'removeProperty' |
+    'setProperty'>;
+
+export function measureTextWidth(text: string | string[], style?: PartialCSSStyle): number {
+    const measureElement = getMeasurementElement(style);
     text = Array.isArray(text) ? text : [text];
     let width = 0;
     for (const item of text) {
@@ -168,30 +178,43 @@ export function measureTextWidth(text: string | string[]): number {
     return width;
 }
 
-export function measureTextHeight(text: string | string[], maxWidth?: number): number {
-    const measureElement = getMeasurementElement();
-    if (maxWidth !== undefined) {
-        measureElement.style.maxWidth = `${Math.ceil(maxWidth)}px`;
-    }
+export function measureTextHeight(text: string | string[], style?: PartialCSSStyle): number {
+    const measureElement = getMeasurementElement(style);
     text = Array.isArray(text) ? text : [text];
     let height = 0;
     for (const item of text) {
         measureElement.textContent = item;
         height = Math.max(measureElement.getBoundingClientRect().height, height);
     }
-    measureElement.style.maxWidth = 'none';
     return height;
 }
 
-function getMeasurementElement(): HTMLElement {
+const defaultStyle = document.createElement('div').style;
+defaultStyle.fontFamily = 'var(--theia-ui-font-family)';
+defaultStyle.fontSize = 'var(--theia-ui-font-size1)';
+defaultStyle.visibility = 'hidden';
+
+function getMeasurementElement(style?: PartialCSSStyle): HTMLElement {
     let measureElement = document.getElementById('measure');
     if (!measureElement) {
         measureElement = document.createElement('span');
         measureElement.id = 'measure';
-        measureElement.style.fontFamily = 'var(--theia-ui-font-family)';
-        measureElement.style.fontSize = 'var(--theia-ui-font-size1)';
-        measureElement.style.visibility = 'hidden';
+        measureElement.style.fontFamily = defaultStyle.fontFamily;
+        measureElement.style.fontSize = defaultStyle.fontSize;
+        measureElement.style.visibility = defaultStyle.visibility;
         document.body.appendChild(measureElement);
+    }
+    const measureStyle = measureElement.style;
+    // Reset styling first
+    for (let i = 0; i < measureStyle.length; i++) {
+        const property = measureStyle[i];
+        measureStyle.setProperty(property, defaultStyle.getPropertyValue(property));
+    }
+    // Apply new styling
+    if (style) {
+        for (const [key, value] of Object.entries(style)) {
+            measureStyle.setProperty(key, value as string);
+        }
     }
     return measureElement;
 }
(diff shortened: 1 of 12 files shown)


packages/core/src/browser/widgets/select-component.tsx
```diff
@@ -0,0 +1,301 @@
...
+        const selected = this.state.selected;
+        return (
+            <div
+                key={index}
+                className={`theia-select-component-option${index === selected ? ' selected' : ''}`}
+                onMouseOver={() => {
+                    this.setState({
+                        selected: index
+                    });
+                }}
```
colin-grant-work, line 286: It looks like this updates the the top-line selected item every time the user moves the mouse. In VSCode, the description changes `onMouseOver`, but the top line only changes on selection by click or keyboard action.

packages/core/src/browser/widgets/select-component.tsx
```diff
@@ -0,0 +1,301 @@
...
+            } else {
+                descriptionNode = <div key="description" className={className}>
+                    {description}
+                </div>;
+            }
+            if (invert) {
+                items.unshift(descriptionNode);
+            } else {
+                items.push(descriptionNode);
+            }
```
colin-grant-work, line 260: It would be slightly more efficient to wait till all items had been added and then `.reverse()` if necessary.

packages/core/src/browser/widgets/select-component.tsx
```diff
@@ -0,0 +1,301 @@
...
+                onKeyDown={e => this.handleKeypress(e)}
+            >
+                <div key="label" className="theia-select-component-label">{selectedItemLabel}</div>
+                <div key="icon" className={`theia-select-component-chevron ${codicon('chevron-down')}`} />
+            </div>
+            {ReactDOM.createPortal(this.renderDropdown(), this.dropdownElement)}
+        </>;
+    }
+
+    protected handleKeypress(ev: React.KeyboardEvent<HTMLDivElement>): void {
```
colin-grant-work, line 161: With this code, <kbd>Up</kbd> and <kbd>Down</kbd> change the selection but don't reveal the select options. In VSCode, the first stroke opens the options, and subsequent <kbd>Up</kbd>s or <kbd>Down</kbd>s change selection.

packages/core/src/browser/widgets/select-component.tsx
```diff
@@ -0,0 +1,301 @@
...
+    detail?: string
+    description?: string
+    markdown?: boolean
+}
+
+export interface SelectComponentProps {
+    options: SelectOption[]
+    selected: number
+    onChange?: (option: SelectOption, index: number) => void
+    onDidSelectedChange?: TheiaEvent<number>
```
colin-grant-work, line 40:
```suggestion
    onDidChangeSelection?: TheiaEvent<number>
```

packages/preferences/src/browser/views/components/preference-select-input.ts
```diff
@@ -17,29 +17,60 @@
...
 
 @injectable()
-export class PreferenceSelectInputRenderer extends PreferenceLeafNodeRenderer<JSONValue, HTMLSelectElement> {
+export class PreferenceSelectInputRenderer extends PreferenceLeafNodeRenderer<JSONValue, HTMLDivElement> {
+
+    protected readonly onDidChangeEmitter = new Emitter<number>();
+
+    protected get onDidSelectedChange(): Event<number> {
+        return this.onDidChangeEmitter.event;
+    }
```
colin-grant-work, line 34: Are these used?

packages/preferences/src/browser/views/components/preference-select-input.ts
```diff
@@ -17,29 +17,60 @@
...
         return this.preferenceNode.preference.data.enum!;
     }
 
+    protected get selectComponentOptions(): SelectOption[] {
+        const items: SelectOption[] = [];
+        const values = this.enumValues;
+        const defaultValue = this.preferenceNode.preference.data.default;
+        for (let i = 0; i < values.length; i++) {
+            const value = `${values[i]}`;
+            const detail = defaultValue === value ? 'default' : undefined;
```
colin-grant-work, line 46: Since `enum` values can be objects, I think this should be a deep equality check (`PreferenceProvider.deepEqual`).

packages/preferences/src/browser/views/components/preference-select-input.ts
```diff
@@ -48,26 +79,25 @@ export class PreferenceSelectInputRenderer extends PreferenceLeafNodeRenderer<JS
...
         }
     }
 
     /**
      * Returns the stringified index corresponding to the currently selected value.
      */
-    protected getDataValue(): string {
+    protected getDataValue(): number {
         const currentValue = this.getValue();
         const selected = this.enumValues.findIndex(value => value === currentValue);
```
colin-grant-work, line 95: To myself a few months ago: since enum values can be objects, this should probably be a deep equality check.

colin-grant-work (commented):
Overall, looking pretty awesome. This is something I've been looking forward to / meaning to do for a while, so I'm really excited to see it!

(PS: everything I say about VSCode's behavior applies in the preference case - I haven't checked the other use cases as thoroughly)

packages/core/src/browser/widgets/select-component.tsx
msujew: It might be, but that would affect the order in which the selection options appear too. They should stay the same independent of whether we are in inverted mode or not. Only the description should change it's place.

packages/core/src/browser/widgets/select-component.tsx
msujew: In my VSCode instance, using `Up` or `Down` with the select field focused only iterates through the preferences (i.e. previous or next preference) and does not open the dropdown field. Is it different for you?

packages/preferences/src/browser/views/components/preference-select-input.ts
msujew: They are used to notify the select component if the preference has been changed from the outside (i.e. by modifying the JSON). See:

https://github.com/eclipse-theia/theia/blob/26e29077154b8864a4aabf0b0a8c746d6fca2464/packages/preferences/src/browser/views/components/preference-select-input.ts#L66-L71

packages/core/src/browser/browser.ts
```diff
@@ -156,3 +156,42 @@ export function preventNavigation(event: WheelEvent): void {
...
+    return height;
+}
+
+function getMeasurementElement(): HTMLElement {
+    let measureElement = document.getElementById('measure');
+    if (!measureElement) {
+        measureElement = document.createElement('span');
+        measureElement.id = 'measure';
+        measureElement.style.fontFamily = 'var(--theia-ui-font-family)';
+        measureElement.style.fontSize = 'var(--theia-ui-font-size1)';
```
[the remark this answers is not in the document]
msujew: Yep, added a configurable `style` argument which allows to pass arbitrary styles 👍

They are cleaned up before the next measurement is performed.

packages/core/src/browser/style/select-component.css
```diff
@@ -0,0 +1,90 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+.theia-select-component {
+    background-color: var(--theia-dropdown-background);
+    outline: var(--theia-dropdown-border) solid 1px;
+    outline-offset: -1px;
+    min-height: 23px;
+    min-width: 90px;
+    padding: 0px 8px;
```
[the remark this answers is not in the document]
msujew: Some of these values were sourced from [vscode's source](https://github.com/microsoft/vscode/blob/main/src/vs/base/browser/ui/selectBox/selectBoxCustom.css), but most of them were simply eyeballed. VSCode's custom select box has a slightly different structure, which is why I couldn't just use their stylesheets.

packages/core/src/browser/widgets/select-component.tsx
colin-grant-work: Oops, fair point - I thought this was applying to every option, not just the description.

packages/core/src/browser/widgets/select-component.tsx
colin-grant-work: That's odd. Here's what I see:

select-screencast

packages/preferences/src/browser/views/components/preference-select-input.ts
colin-grant-work: I had mistakenly reversed the flow: I thought the `SelectComponent`'s `onDidSelectionChange` was to notify external parties, not to receive notifications. In principle, I'd probably prefer a more DOM-like API (i.e. the ability to set the value), but with React, that would mean tracking a ref anyway, so I'm not sure it would end up any cleaner. What do you think?

vince-fugnitto (commented):
The changes look really great and are a very nice improvement to the framework 👍 

I do not have any feedback for the code itself at the time, I mostly tested the functionality and each dropdown, especially how the changes enhance the UX of the preferences-view and now display `enumDescriptions`.

The only minor UI improvement I could see at the time is to reduce the separator height, which can be seen with debug configurations dropdown:

Screen Shot 2022-04-07 at 9 44 50 AM

It is debatable however if you believe the thicker border looks better than the behavior we see in vscode.

packages/core/src/browser/widgets/select-component.tsx
msujew: Alright, I will keep it in mind for a future contribution, but for now I would keep it as it is 👍

packages/preferences/src/browser/views/components/preference-select-input.ts
msujew: Since this is the only case where we need to programmatically change the selection value (so far), I would actually go with the ref idea.

msujew:
@vince-fugnitto I think you're right. I reduce the separator thickness to `1px` and reduced the margin as well to make everything a bit more compact 👍

vince-fugnitto (approved):
The changes look good, and work well for me :+1:

msujew:
@colin-grant-work I would merge this within the next few days. Any other comments from your side?

colin-grant-work (approved):
Looks great. Very nice work @msujew!

PR #10994, opened 2022-04-06
refactor: rename ConnectionHandlers.push param in packages/core/src/node/messaging/messaging-contribution.ts

harbin1053020115:
#### What it does

Rename message-contirbution.ts(packages/core/src/node/messaging/messaging-contribution.ts) ConnectionHandlers.push param, which is changing channel to socket.

#### How to test

no need to test

#### Review checklist

#### Reminder for reviewers


harbin1053020115:
In MessageContribution
```typescript
push(spec: string, callback: (params: MessagingService.PathParams, connection: T) => void): void {
      const route = new Route(spec);
      this.handlers.push((path, channel) => {
          const params = route.match(path);
          if (!params) {
              return false;
          }
          callback(params, channel);
          return route.reverse(params);
      });
  }
```

PR #10997, opened 2022-04-06
For calls to vscode.tasks.executeTask that use shell tasks, treat command as we treat the arguments in terms of the use of ShellQuotedString.

### LINKED ISSUE #10805: Add ability to pass a QuotedString as a command to tasks.executeTask
The motivation behind this request is to make [makefile-tools extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.makefile-tools) work under Theia. Currently the extension fails to execute the make commands. Any guidance on how to do a PR to allow commands to be `ShellQuotedString`s would be greatly appreciated.

### Feature Description:

This is because [this code](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/type-converters.ts#L894) specifically disallows passing a command that is not a `string`. The error message shows that `ShellQuotedStrings`s are not handled. And the makefile-tools extension [specifically](https://github.com/microsoft/vscode-makefile-tools/blob/main/src/make.ts#L328) passes the command as a `ShellQuotedStrings`

As an experiment, I changed [this code](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/type-converters.ts#L894) to allow a `ShellQuotedString` to be handled:
```
    const command = execution.command;
    if (typeof command === 'string' || (command && typeof command.value === 'string')) {
        taskDto.command = command;
        taskDto.args = getShellArgs(execution.args);
        return taskDto;
    } else {
        throw new Error('Converting ShellQuotedString command is not implemented');
    }
```
And I saw that the underlying task handler correctly ran the command (in this case `command.value='make', 
command.quoting=vscode.ShellQuoting.Strong`).

jdmchp:
#### What it does
This pull request fixes #10805 

It basically removes the [limitation](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/type-converters.ts#L889) of a command having to be a string.
It does this by treating the command the same way as the arguments are currently treated: for `ShellQuotedString`s keep the value of the command, and then later on in the backend, quote strongly.

The reason for this fix is to allow the [ms-vscode.makefile-tools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.makefile-tools) extension to work with Theia. This extension runs make and uses a `ShellQuotedString` as a command.

#### How to test
Try to use the ms-vscode.makefile-tools extension, it will not work. Apply the patch, it will start working.

Before the patch, any call to `vscode.tasks.executeTask` where a command is passed as a `ShellQuotedString` would fail. For example:

```ts
let myTaskOptions: vscode.ShellExecutionOptions = {env: process.env as { [key: string]: string }, cwd: "/home/jose/tmp/foo"};
const quotingStlye: vscode.ShellQuoting = vscode.ShellQuoting.Strong;
let myTaskCommand: vscode.ShellQuotedString = {value: "make with spaces", quoting: quotingStlye};
let shellExec: vscode.ShellExecution = new vscode.ShellExecution("make with spaces", myTaskOptions);
let myTask: vscode.Task = new vscode.Task({type: "shell", group: "build", label: "my label"},
vscode.TaskScope.Workspace, "makefileBuildTaskName", "makefile", shellExec);

myTask.problemMatchers = [];
myTask.presentationOptions.showReuseMessage = true;

await vscode.tasks.executeTask(myTask);
```

would fail with:

```
throw new Error('Converting ShellQuotedString command is not implemented');
```
#### Review checklist

- [X] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/eclipse-theia/theia/blob/master/doc/pull-requests.md#requesting-a-review)
Commands with spaces, for example, now can be passed to `executeTask` and they will work OK in windows or *nix.

### DIFF
--- a/packages/plugin-ext/src/plugin/type-converters.spec.ts
+++ b/packages/plugin-ext/src/plugin/type-converters.spec.ts
@@ -235,6 +235,19 @@ describe('Type converters:', () => {
             }
         };
 
+        const shellPluginTaskWithQuotedCommand: theia.Task = {
+            ...shellPluginTask, execution: {
+                command: {
+                    quoting: types.ShellQuoting.Strong,
+                    value: 'yarn'
+                },
+                args,
+                options: {
+                    cwd
+                }
+            }
+        };
+
         const pluginTaskWithCommandLine: theia.Task = {
             name: label,
             source,
@@ -278,9 +291,17 @@ describe('Type converters:', () => {
             }
         };
 
-        it('should convert to task dto', () => {
+        it('shell task with string command should convert to task dto', () => {
             // when
             const result: TaskDto | undefined = Converter.fromTask(shellPluginTask);
+
+            // then
+            assert.notStrictEqual(result, undefined);
+            assert.deepStrictEqual(result, shellTaskDto);
+        });
+
+        it('shell task with ShellStringQuoted command should convert to task dto', () => {
+            const result: TaskDto | undefined = Converter.fromTask(shellPluginTaskWithQuotedCommand);
 
             // then
             assert.notStrictEqual(result, undefined);
--- a/packages/plugin-ext/src/plugin/type-converters.spec.ts
+++ b/packages/plugin-ext/src/plugin/type-converters.spec.ts
@@ -291,7 +291,7 @@ describe('Type converters:', () => {
             }
         };
 
-        it('shell task with string command should convert to task dto', () => {
+        it('should convert to task dto with string command', () => {
             // when
             const result: TaskDto | undefined = Converter.fromTask(shellPluginTask);
 
@@ -300,7 +300,7 @@ describe('Type converters:', () => {
             assert.deepStrictEqual(result, shellTaskDto);
         });
 
-        it('shell task with ShellStringQuoted command should convert to task dto', () => {
+        it('should convert to task dto with ShellStringQuoted', () => {
             const result: TaskDto | undefined = Converter.fromTask(shellPluginTaskWithQuotedCommand);
 
             // then
--- a/packages/plugin-ext/src/plugin/type-converters.spec.ts
+++ b/packages/plugin-ext/src/plugin/type-converters.spec.ts
@@ -302,8 +302,6 @@ describe('Type converters:', () => {
 
         it('should convert to task dto with ShellStringQuoted', () => {
             const result: TaskDto | undefined = Converter.fromTask(shellPluginTaskWithQuotedCommand);
-
-            // then
             assert.notStrictEqual(result, undefined);
             assert.deepStrictEqual(result, shellTaskDto);
         });
--- a/packages/plugin-ext/src/plugin/type-converters.ts
+++ b/packages/plugin-ext/src/plugin/type-converters.ts
@@ -885,13 +885,12 @@ export function fromShellExecution(execution: theia.ShellExecution, taskDto: Tas
         return taskDto;
     }
 
-    const command = execution.command;
-    if (typeof command === 'string') {
-        taskDto.command = command;
+    if (execution.command) {
+        taskDto.command = getCommand(execution.command);
         taskDto.args = getShellArgs(execution.args);
         return taskDto;
     } else {
-        throw new Error('Converting ShellQuotedString command is not implemented');
+        throw new Error('Command is undefined');
     }
 }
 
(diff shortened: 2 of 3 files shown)


vince-fugnitto (commented):
@jdmchp thank you for your first contribution to the project! 

Please be sure to address the test failures introduced with the changes and we will review shortly afterwards :)

packages/plugin-ext/src/plugin/type-converters.spec.ts
```diff
@@ -278,7 +291,7 @@ describe('Type converters:', () => {
             }
         };
 
-        it('should convert to task dto', () => {
+        it('shell task with string command should convert to task dto', () => {
```
vince-fugnitto, line 294: minor: generally we follow titles like `should...` so we can update to:

```suggestion
        it('should convert to task dto with with string command', () => {
```

packages/plugin-ext/src/plugin/type-converters.spec.ts
```diff
@@ -287,6 +300,14 @@ describe('Type converters:', () => {
             assert.deepStrictEqual(result, shellTaskDto);
         });
 
+        it('shell task with ShellStringQuoted command should convert to task dto', () => {
```
vince-fugnitto, line 303: minor: similarly to above:

```suggestion
        it('should convert to task dto with `ShellStringQuoted` command', () => {
```

packages/plugin-ext/src/plugin/type-converters.spec.ts
```diff
@@ -287,6 +300,14 @@ describe('Type converters:', () => {
...
+            const result: TaskDto | undefined = Converter.fromTask(shellPluginTaskWithQuotedCommand);
+
+            // then
```
vince-fugnitto, line 306: minor: unecessary.

packages/plugin-ext/src/plugin/type-converters.ts
```diff
@@ -949,6 +948,15 @@ export function getShellArgs(args: undefined | (string | theia.ShellQuotedString
...
 }
 
+function getCommand(command: string | theia.ShellQuotedString): string {
+    if (typeof command === 'string') {
+        return command;
+    }
+    const shellQuotedCommand = command as theia.ShellQuotedString;
+
+    return shellQuotedCommand.value;
+}
```
vince-fugnitto, line 958: minor: can be simplified to:

```suggestion
function getCommand(command: string | theia.ShellQuotedString): string {
    return typeof command === 'string' ? command : command.value;
}
```

or 

```suggestion
function getCommand(command: string | theia.ShellQuotedString): string {
    if (typeof command === 'string') {
        return command;
    }
    return shellQuotedCommand.value;
}
```

vince-fugnitto (commented):
The changes work well for me, I confirmed that it works when using my custom extension 
([shell-quote-0.0.1.zip](https://github.com/eclipse-theia/theia/files/8473190/shell-quote-0.0.1.zip)). I only have minor commands about the code.

packages/plugin-ext/src/plugin/type-converters.ts
jdmchp: @vince-fugnitto thanks for the quick feedback. I will apply your comments.

packages/plugin-ext/src/plugin/type-converters.spec.ts
jdmchp: I did not get which part was unnecessary @vince-fugnitto , do you mean the new ''should convert to task dto with ShellStringQuoted'' is unnecessary?

msujew: @jdmchp I believe Vince meant the comment, which doesn't really convey any content IMO.

jdmchp:
Hello @vince-fugnitto , I pushed changes using the feedback you gave me. 
Please let me know if more work is needed. Thanks.

packages/plugin-ext/src/plugin/type-converters.spec.ts
```diff
@@ -287,6 +300,14 @@ describe('Type converters:', () => {
             assert.deepStrictEqual(result, shellTaskDto);
         });
 
+        it('should convert to task dto with ShellStringQuoted', () => {
+            const result: TaskDto | undefined = Converter.fromTask(shellPluginTaskWithQuotedCommand);
+
+            // then
+            assert.notStrictEqual(result, undefined);
+            assert.deepStrictEqual(result, shellTaskDto);
+        });
```
vince-fugnitto, line 309: minor: I believe this is still unaddressed from the previous review (https://github.com/eclipse-theia/theia/pull/10997#discussion_r848420705)

```suggestion
        it('should convert to task dto with ShellStringQuoted', () => {
            const result: TaskDto | undefined = Converter.fromTask(shellPluginTaskWithQuotedCommand);
            assert.notStrictEqual(result, undefined);
            assert.deepStrictEqual(result, shellTaskDto);
        });
```

jdmchp: Ah... yes. You are right. Just pushed the fix.
Thanks for the quick response.

vince-fugnitto (approved):
@jdmchp thank you for your patience and first contribution to the project! :)

Issue #10998, opened 2022-04-06
resolveDebugConfiguration returning 'undefined'  or 'null' does not comply to vscode API


alvsan09:
### Bug Description:
The vscode plugin `ms-vscode.js-debug` provides a dynamic configuration called `Create Java Debug Terminal`
which does not actually require a session, it simply opens a terminal as a side effect of starting it, and when the extension
provider is called with `resolveDebugConfiguration` the extension simply returns `undefined` to indicate that no further processing  is required.

This is expected as specified:  https://code.visualstudio.com/api/references/vscode-api#DebugConfigurationProvider

Theia currently ignores `undefined` or `null`, and starts the session using the configuration in place before an `undefined` or `null` is returned.

### Steps to Reproduce:
As it can be seen in the `gif` below this case starts a session with a corresponding terminal in addition to the terminal triggered by the extension.
1. Open the command palette and type `debug `, not the ending space is important
2. Select `Create JavaScript Debug terminal`  from the list of available configurations
3. See that a Session and corresponding terminal is started in addition to a `JavaScript Debug Terminal`

![js-debug-two-terminals](https://user-images.githubusercontent.com/76971376/162032238-94ed537a-dec3-4c1d-ae23-232b665da8e6.gif)

### Additional Information

- Operating System: Ubuntu
- Theia Version: 1.24.0

PR #10999, opened 2022-04-06
Honor debug stop from configuration providers

alvsan09:
#### What it does

Fixes: #10998 

Debug configuration providers can indicate to stop debugging by
returning `undefined` or `null` on either of the functions
- `resolveDebugConfiguration` or
- `resolveDebugConfigurationWithSubstitutedVariables`

The vscode API
https://code.visualstudio.com/api/references/vscode-api#DebugConfigurationProvider
specifies different handling for `undefined` and `null` as follows:

"Returning the value 'undefined' prevents the debug session from
starting. Returning the value 'null' prevents the debug session from
starting and opens the underlying debug configuration instead."

This implementation focuses on passing the `undefined` or `null`
response from the extension to the debug session manager where the
indication from the extension is handled.

#### How to test

1. Download the following extension and place it under your theia/plugins folder 
https://github.com/alvsan09/debug-activation-events/blob/master/on-debug-resolve-null.vsix
2. Start theia 
3. Testing behavior for an extension returning `undefined`: 
From the command palette (Ctrl + shft + P) , type `debug `, note trailing space is needed.
Select configuration named `Create JavaScript Debug Terminal`
Verify that no debug session is started and that a single `JavaScript Debug Terminal` instance is created.
This verifies the fix for issue: #10998
6. Testing behavior for an extension returning `null`:
From the command palette (Ctrl + shft + P) , type `debug `, note trailing space is needed.
Select configuration named `Resolve config to null` (this configuration is provided by the test extension loaded in step 1)
Verify that no debug session is started and that the `launch.json` file is opened.
 
config_resolve_undefined_null

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-session-manager.ts
+++ b/packages/debug/src/browser/debug-session-manager.ts
@@ -193,6 +193,19 @@ export class DebugSessionManager {
                 await this.fireWillStartDebugSession();
                 const resolved = await this.resolveConfiguration(options);
 
+                if (!resolved) {
+                    // As per vscode API: https://code.visualstudio.com/api/references/vscode-api#DebugConfigurationProvider
+                    // "Returning the value 'undefined' prevents the debug session from starting.
+                    // Returning the value 'null' prevents the debug session from starting and opens the
+                    // underlying debug configuration instead."
+
+                    // eslint-disable-next-line no-null/no-null
+                    if (resolved === null) {
+                        this.debugConfigurationManager.openConfiguration();
+                    }
+                    return undefined;
+                }
+
                 // preLaunchTask isn't run in case of auto restart as well as postDebugTask
                 if (!options.configuration.__restart) {
                     const taskRun = await this.runTask(options.workspaceFolderUri, resolved.configuration.preLaunchTask, true);
@@ -221,17 +234,31 @@ export class DebugSessionManager {
     }
 
     protected configurationIds = new Map<string, number>();
-    protected async resolveConfiguration(options: Readonly<DebugSessionOptions>): Promise<InternalDebugSessionOptions> {
+    protected async resolveConfiguration(
+        options: Readonly<DebugSessionOptions>
+    ): Promise<InternalDebugSessionOptions | undefined | null> {
         if (InternalDebugSessionOptions.is(options)) {
             return options;
         }
         const { workspaceFolderUri } = options;
         let configuration = await this.resolveDebugConfiguration(options.configuration, workspaceFolderUri);
-        configuration = await this.variableResolver.resolve(configuration, {
-            context: options.workspaceFolderUri ? new URI(options.workspaceFolderUri) : undefined,
-            configurationSection: 'launch'
-        });
-        configuration = await this.resolveDebugConfigurationWithSubstitutedVariables(configuration, workspaceFolderUri);
+
+        if (configuration) {
+            configuration = await this.variableResolver.resolve(configuration, {
+                context: options.workspaceFolderUri ? new URI(options.workspaceFolderUri) : undefined,
+                configurationSection: 'launch',
+            });
+
+            configuration = await this.resolveDebugConfigurationWithSubstitutedVariables(
+                configuration,
+                workspaceFolderUri
+            );
+        }
+
+        if (!configuration) {
+            return configuration;
+        }
+
         const key = configuration.name + workspaceFolderUri;
         const id = this.configurationIds.has(key) ? this.configurationIds.get(key)! + 1 : 0;
         this.configurationIds.set(key, id);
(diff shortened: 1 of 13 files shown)


packages/plugin/src/theia.d.ts
```diff
@@ -9691,7 +9698,11 @@ export module '@theia/plugin' {
          * @param token A cancellation token.
          * @return The resolved debug configuration or undefined or null.
          */
-        resolveDebugConfigurationWithSubstitutedVariables?(folder: WorkspaceFolder | undefined, debugConfiguration: DebugConfiguration, token?: CancellationToken): ProviderResult<DebugConfiguration>;
+        resolveDebugConfigurationWithSubstitutedVariables?(
+            folder: WorkspaceFolder | undefined,
+            debugConfiguration: DebugConfiguration,
+            token?: CancellationToken
+        ): ProviderResult<DebugConfiguration | undefined | null>;
```
vince-fugnitto, line 9705: @alvsan09 I don't believe this api needs to be update since it is already aligned with [vscode](https://github.com/microsoft/vscode/blob/a3d53f76790d8c88f835afe85225ff6442703409/src/vscode-dts/vscode.d.ts#L13655-L13699) (and doing so will likely mark our api as incompatible in the compatibility report). Rather, the [`ProviderResult`](https://github.com/microsoft/vscode/blob/a3d53f76790d8c88f835afe85225ff6442703409/src/vscode-dts/vscode.d.ts#L2135) should be updated to accept `null` like in vscode:

```ts
export type ProviderResult<T> = T | undefined | null | Thenable<T | undefined | null>;
```

packages/plugin-ext/src/plugin/node/debug/debug.ts
```diff
@@ -394,6 +406,9 @@ export class DebugExtImpl implements DebugExt {
         return result;
     }
 
+    /**
+     * @deprecated since 1.25.0. Use $registerDebugConfigurationProvider with $resolveDebugConfigurationByHandle instead.
+     */
     async $resolveDebugConfigurations(debugConfiguration: theia.DebugConfiguration, workspaceFolderUri: string | undefined): Promise<theia.DebugConfiguration | undefined> {
```
vince-fugnitto, line 412: Given how the method is pretty internal to the plugin-system (and not easily extended), it would be fine to simply break and not deprecate.

packages/plugin-ext/src/plugin/node/debug/debug.ts
```diff
@@ -419,6 +434,9 @@ export class DebugExtImpl implements DebugExt {
         return current;
     }
 
+    /**
+     * @deprecated since 1.25.0. Use $registerDebugConfigurationProvider with $resolveDebugConfigurationWithSubstitutedVariablesByHandle instead.
+     */
     async $resolveDebugConfigurationWithSubstitutedVariables(debugConfiguration: theia.DebugConfiguration, workspaceFolderUri: string | undefined):
```
vince-fugnitto, line 440: Same as the comment for `$resolveDebugConfigurations`.

packages/plugin/src/theia.d.ts
alvsan09: Nice catch Vnce! Thanks!
It is addressed in today's commit.

colin-grant-work (approved):
The code looks good and the provided tests work. @vince-fugnitto mentioned VSCode's `ProviderResult` type, and I think it might make sense to introduce something similar in our `packages/core/src/common/types.ts` file to reduce the number of times we have to write `| undefined | null`?

alvsan09:
> I think it might make sense to introduce something similar in our `packages/core/src/common/types.ts` file to reduce the number of times we have to write `| undefined | null`?
Hi @colin-grant-work, 
I introduced a new type under `common` so the result can be passed across all the way to the browser, 

The new type is under a new commit so we can discuss this particular update separately!

I have updated all the related returns accordingly,  let me know what you think !

packages/core/src/common/types.ts
```diff
@@ -130,3 +130,5 @@ export namespace ArrayUtils {
 export function unreachable(_never: never, message: string = 'unhandled case'): never {
     throw new Error(message);
 }
+
+export type ResultPromise<T> = Promise<T | undefined | null>;
```
vince-fugnitto, line 134: @alvsan09 I would probably keep the same name `ProviderResult` unless you have a strong reason to deviate? I don't think `ResultPromise` would be a good name for this type.

paul-marechal: I don't think we benefit from this type at all at them moment, I would remove it.

packages/debug/src/browser/debug-session-manager.ts
```diff
@@ -221,17 +235,31 @@ export class DebugSessionManager {
     }
 
     protected configurationIds = new Map<string, number>();
-    protected async resolveConfiguration(options: Readonly<DebugSessionOptions>): Promise<InternalDebugSessionOptions> {
+    protected async resolveConfiguration(
+        options: Readonly<DebugSessionOptions>
+    ): ResultPromise<InternalDebugSessionOptions> {
```
paul-marechal, line 240: Those methods are implementations. As long as the interface to implement defines return types as `ProviderResult<T>` which really means `T | undefined | null | Thenable<T | undefined  | null>` we can retype implementations to something more specific/convenient. The interface says "one of those can be returned" while the implementation will only return a subset of that.

It's fine to type as `Promise<InternalDebugSessionOptions | undefined>` here.

packages/core/src/common/types.ts
paul-marechal: Specifically I don't think it's super useful to use for implementations, which is its main use right now.

Looking at VS Code they have a `ProviderResult` used in definitions, but they never mention it in implementations.

alvsan09:
@colin-grant-work, @paul-marechal, @vince-fugnitto,
After offline discussions with Paul and Vince, there seems to be a preference to not introduce the new type,
so I have reset the tip of the branch to the previous commit (before the introduction of the new type), 

Colin had already approved that commit, @vince-fugnitto  and @paul-marechal any pending items before approving this PR?

alvsan09:
Re-based to latest master

colin-grant-work:
My approval stands, with or without the additional type. 👍

vince-fugnitto (approved):
LGTM!

PR #11018, opened 2022-04-12
Allow to override language metadata

### LINKED ISSUE #11010: Get more detailed language info
### Feature Description:
When asking the backend which are the languages available for localization, it's currently possible to do so calling [getAvailableLanguages()](https://github.com/eclipse-theia/theia/blob/master/packages/core/src/node/i18n/localization-provider.ts#L50-L58) in the LocalizationProvider.

This method returns only an array of _language ids_. I would really like this method to return a more detailed object, containing at least the language name (and maybe the localized language name too).

msujew:
#### What it does

Closes #11010 

1. Fixes a bug in the `LocalizationProvider` which should've allowed to merge localization. It only merged the `translations` property and nothing else, leaving the `languageName` empty.
2. Returns a bit more metadata in the `getAvailableLanguages()` method.

#### How to test

1. Change the language and assert that the app works as expected.
2. Set a breakpoint in [this line](https://github.com/eclipse-theia/theia/blob/07dcfbd071c5a184a32efe72915e09506c927bdb/packages/core/src/browser/common-frontend-contribution.ts#L1145) and assert that the `languageName` and `localizedLanguageName` property is correctly set in the `availableLanguages` array (is `undefined` on `master`).

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 <a name="breaking_changes_1.25.0">[Breaking Changes:](#breaking_changes_1.25.0)</a>
 
+- [core] changed return type of `(Async)LocalizationProvider#getAvailableLanguages` from `string[]` to `LanguageInfo[]` [#11018](https://github.com/eclipse-theia/theia/pull/11018)
+
 ## v1.24.0 - 3/31/2022
 
 [1.24.0 Milestone](https://github.com/eclipse-theia/theia/milestone/32)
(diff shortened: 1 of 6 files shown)


vince-fugnitto (commented):
I confirmed that `getAvailableLanguages` now contains additional information:

localization

However there is an error thrown when executing `configure display language` with a language-pack installed:

image

msujew:
@vince-fugnitto I cannot reproduce the issue you experience (the error message in the console), but I've made the code a bit more error-safe (it now checks that the `languageId` is actually a string). Does it still happen for you?

vince-fugnitto (approved):
It works well for me now :+1: 

For reference I had:
- installed the french language pack
- triggered the `configure display language` command
- the quick-pick was empty, and the error was thrown

PR #11023, opened 2022-04-13
Export page objects directly under `lib`

planger:
#### What it does
Having `src` in the exported paths of the `lib` folder is sort of unexpected and weird.
Better to export it in a way so that imports for consumers look like
```typescript
import { TheiaApp } from '@theia/playwright';
```

Therefore, we define `src` as `rootDir` to get rid of the `src` in the `lib` path, add an index re-exporting all files in the root and define this index as `main` of this package.

For simplicity we also moved the tests into the `src` folder to be able to compile them consistently.

#### How to test

After a full build, you can add `@theia/playwright` as a dependency of another package in this repo and add an import to any source file of the package as follows:
```typescript
import { TheiaApp } from '@theia/playwright';
```
This import should be successfully resolved during a build.

#### Review checklist

#### Reminder for reviewers


examples/playwright/package.json
```diff
@@ -13,11 +13,11 @@
   "homepage": "https://github.com/eclipse-theia/theia",
   "scripts": {
     "prepare": "yarn clean && yarn build",
-    "clean": "rimraf lib",
+    "clean": "rimraf lib *.tsbuildinfo",
     "build": "tsc --incremental && yarn lint && npx playwright install chromium",
```
colin-grant-work, line 17: For another time: as part of https://github.com/eclipse-theia/theia/pull/9710, linting was removed from the `prepare` stage of our builds, but it still runs for Playwright. Since linting can be time consuming, it would be preferable to move it to its own command.

colin-grant-work (approved):
Playwright tests continue to run, and the exports from the package appear to be accessible externally. 👍

examples/playwright/package.json
planger: Ah, thanks for the hint! I appended a change to remove linting from the `yarn build` script.
As I see `yarn lint` will be explicitly invoked during the CI build. So we can safely remove it from the `yarn build` script.
The `playwright` package has an explicit lint script already, so that should already be fine.

planger:
@vince-fugnitto If you are fine with this change (and build is green), can you please also merge it? (I can't do it myself and would have to wait :)) Thanks a lot for your review and effort!

colin-grant-work:
@planger, I'm happy to merge. At the moment, with several commits with descriptive commit messages, I'm not sure exactly what you'd like the final commit message to say - would you mind squashing the commits and adjusting the message how you'd like?

planger:
@colin-grant-work Thanks a lot! I've squashed and "merged" the commit messages into a single one.

PR #11027, opened 2022-04-14
Implement CancellationToken logic for withProgress API

matt-dot:
#### What it does

- Unlocks functionality of `token.onCancellationRequested` & `token.isCancellationRequested` (from the VS Code Extension API)
- Fixes https://github.com/eclipse-theia/theia/issues/11030 
#### How to test

- Clone this [VS Code Example](https://github.com/microsoft/vscode-extension-samples/tree/main/progress-sample) and open in Theia
- Run the Launch Extension launch configuration
- Add a `console.log` within the `setTimeout` callbacks [here](https://github.com/microsoft/vscode-extension-samples/blob/2556c82cb333cf65d372bd01ac30c35ea1898a0e/progress-sample/src/extension.ts#L21-L31) to check the value of `token.isCancellationRequested`
- Run the extension's `showProgress` command (using the command pallette) and press the cancel button contained in the notification
- Pre-changes
    - No console output from within `token.onCancellationRequested`
    - `token.isCancellationRequested` remains false indefinitely
 - Post-changes
    - Console output from within `token.onCancellationRequested`
    - `token.isCancellationRequested` changes based on button press

#### Review checklist

#### Reminder for reviewers

Signed-off-by: Matthew Langley <matthew.langley@arm.com>


packages/plugin-ext/src/common/plugin-api-rpc.ts
```diff
@@ -1858,6 +1861,7 @@ export const MAIN_RPC_CONTEXT = {
     WEBVIEW_VIEWS_EXT: createProxyIdentifier<WebviewViewsExt>('WebviewViewsExt'),
     STORAGE_EXT: createProxyIdentifier<StorageExt>('StorageExt'),
     TASKS_EXT: createProxyIdentifier<TasksExt>('TasksExt'),
+    TASKS_MAIN: createProxyIdentifier<TasksMain>('TasksMain'),
```
colin-grant-work, line 1864: I don't see any references to this new field in the code, and I don't think a `*MAIN` entry should appear here?

packages/plugin-ext/src/plugin/notification.ts
```diff
@@ -32,13 +33,20 @@ export class NotificationExtImpl implements NotificationExt {
         options: ProgressOptions,
         task: (progress: Progress<{ message?: string; increment?: number }>, token: CancellationToken) => PromiseLike<R>
     ): Promise<R> {
+        const source = new CancellationTokenSource();
         const id = new Deferred<string>();
```
colin-grant-work, line 37: The handling of this ID - from the old code - is very odd to me. I don't think the Ext side should let the Main side tell it what local handle to use; it should tell the Main side what handle to use to communicate with it. Similarly, the Main side waits for the `ProgressService` to provide it with a handle, when it should use its own handle and map it to what the progress service returns. But those existing oddities don't need to be fixed here.

colin-grant-work (commented):
I believe this is working well. The extraneous `TASKS_MAIN` should be removed (I think), and it would be helpful to update the testing instructions:

> Add a console.log checking the value of token.isCancellationRequested

Add a `console.log` where? In my testing it, I added it to the `setTimeout` callbacks [here](https://github.com/microsoft/vscode-extension-samples/blob/2556c82cb333cf65d372bd01ac30c35ea1898a0e/progress-sample/src/extension.ts#L21-L31), but the instructions should provide more detail.

> Run and press the cancel button contained in the notification

Run what? I assume the extension's 'Show Progress' command.

matt-dot:
> 

Yes your assumptions are correct. I'll be sure to be more detailed and specific in the future 👍

packages/plugin-ext/src/plugin/notification.ts
matt-dot: Yes I agree. I noticed this while making my changes as I was comparing with the VS Code implementation. It seemed that the changes required to align with VS Code would be more complex and should be a separate PR, so I chose to work within the current interface.

packages/plugin-ext/src/common/plugin-api-rpc.ts
matt-dot: Removed.

colin-grant-work: It looks like you've removed TASKS_MAIN from the main RPC object and left it here - it should be vice versa :-).

matt-dot: Ah yes thanks. I have made the changes.

packages/plugin-ext/src/common/plugin-api-rpc.ts
```diff
@@ -1843,6 +1845,7 @@ export const MAIN_RPC_CONTEXT = {
     QUICK_OPEN_EXT: createProxyIdentifier<QuickOpenExt>('QuickOpenExt'),
     WINDOW_STATE_EXT: createProxyIdentifier<WindowStateExt>('WindowStateExt'),
     NOTIFICATION_EXT: createProxyIdentifier<NotificationExt>('NotificationExt'),
+    NOTIFICATION_MAIN: createProxyIdentifier<NotificationMain>('NotificationMain'),
```
colin-grant-work, line 1848: Sorry for making life difficult, but at a second glance, it looks like neither of these additions is necessary, either: `NOTIFICATION_MAIN` is already present in the object that contains all of the `MAIN` identifiers, and `NOTIFICATION_EXT` is already present in the object with all of the `EXT` identifiers. I believe that wherever you need to refer to those identifiers, you should do so by referring to object in which they already exist.

matt-dot: Not at all, I appreciate the thorough review!

colin-grant-work (approved):
LGTM, thanks for the contribution!

Issue #10987, opened 2022-04-04
monaco: quick-input list styling needs to be adjusted


vince-fugnitto:
### Bug Description:

Unfortunately, despite the changes from https://github.com/eclipse-theia/theia/pull/10962, the styling of entries in the quick-input menu needs to be adjusted further. The changes caused the entries with the `theia-file-icon` icon-theme to be cut off and misaligned:

![image](https://user-images.githubusercontent.com/40359487/161544353-9b6c5e25-ce45-48eb-a52d-4428edace225.png)

### Steps to Reproduce:

1. start the application
2. trigger `preferences: file icon theme` and select `file icons (theia)`
3. trigger the quick file open (notice the incorrect styling)

PR #10962, opened 2022-03-30
monaco: fix styling of toggled commands

vince-fugnitto:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/10960.

The commit fixes the styling of the toggled commands in the quick-command menu. Previously, the font-family was incorrect and led to an inconsistent styling with the rest of the commands.

#### How to test

1. start the application
2. trigger the quick-command menu <kbd>F1</kbd>
3. confirm the styling of the commands are correct, and that toggle commands (with checkmarks) are correct

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

Signed-off-by: vince-fugnitto <vincent.fugnitto@ericsson.com>

PR #11029, opened 2022-04-14
Fix more quick input styling issues

msujew:
#### What it does

1. Closes https://github.com/eclipse-theia/theia/issues/10987 by fixing the actual source of the misalignment. A `codicon` css rule sets the `line-height` of the whole row to `13px` which messes up all of the sub-layouting
2. Fixes a minor font issue for `label-description` parts of the quick input row, which used a serif font. See also https://github.com/eclipse-theia/theia/pull/10962

#### How to test

1. Open the command palette and enter `view: toggle`
2. Assert that quick input items with checkmarks are correctly layouted
3. Run the `Open Recent Workspaces` command (The original issue only seems to appear on Windows+Electron, don't know why)
4. Assert that the `label-description` font is using the default app font

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/style/index.css
+++ b/packages/monaco/src/browser/style/index.css
@@ -197,6 +197,7 @@
   .monaco-icon-label
   .monaco-icon-description-container
   .label-description {
+    font-family: var(--theia-ui-font-family);
     font-size: calc(var(--theia-ui-font-size1) * 0.95) !important;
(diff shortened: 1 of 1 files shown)


vince-fugnitto (approved):
Thank you for taking a look into fixing the problem :+1: 
I confirmed that it works well for, I also confirmed using different file-icon themes.

Issue #11031, opened 2022-04-14
editor: 'save as' does not work correctly


vince-fugnitto:
#### Bug Description:

There is an issue with the **'save as** functionality where:
- performing `save as` on an editor does not close the original editor
- performing `save` for untitled editors (ultimately a save as) the original untitled editor is preserved

https://user-images.githubusercontent.com/40359487/163440958-ff2d9669-b447-428b-a660-c09a848164fc.mp4

```
root WARN TypeError: Cannot read properties of undefined (reading 'create')
    at createTextBuffer (file:///home/evinfug/workspaces/theia/examples/electron/lib/vendors-node_modules_theia_monaco-editor-core_esm_vs_editor_editor_main_js.js:93904:20)
    at TextModel.setValue (file:///home/evinfug/workspaces/theia/examples/electron/lib/vendors-node_modules_theia_monaco-editor-core_esm_vs_editor_editor_main_js.js:94130:44)
    at MonacoEditorModel.applySnapshot (file:///home/evinfug/workspaces/theia/examples/electron/lib/node_modules_theia_monaco-editor-core_esm_vs_base_common_sync_recursive-packages_monaco_lib_b-6d7964.js:565:20)
    at FilesystemSaveResourceService.copyAndSave (file:///home/evinfug/workspaces/theia/examples/electron/lib/packages_filesystem_lib_browser_breadcrumbs_filepath-breadcrumb_js-packages_filesystem_lib_br-eff824.js:842:28)
    at async FilesystemSaveResourceService.saveAs (file:///home/evinfug/workspaces/theia/examples/electron/lib/packages_filesystem_lib_browser_breadcrumbs_filepath-breadcrumb_js-packages_filesystem_lib_br-eff824.js:816:17)
    at async FilesystemSaveResourceService.save (file:///home/evinfug/workspaces/the
```

#### Steps to Reproduce:

1. start the application
2. execute the command `new file` - an untitled editor should open
3. add content to the untitled editor, and `save` - notice the editor is kept open and dirty with a new editor opened
4. execute the command `save as` on a regular editor
5. after performing the `save as` the original editor is not replaced with the new one

PR #11032, opened 2022-04-14
Narrow typing of Saveable.createSnapshot

colin-grant-work:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/11031.

In the Monaco uplift, the return of `MonacoEditorModel.createSnapshot` was changed, but its effect on the `Saveable` system was not detected because the return of `Saveable.createSnapshot` was typed as `object`. This PR narrows that typing so that callers of the function have some idea what to expect, and deviations from those expectations can be caught.

#### How to test

1. Open two editors, one with an existing file and one Untitled.
1. Use `Save as` on both editors
1. In both cases, the content should be copied and the old editor closed.
1. No errors should be logged about failures to create text buffers.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
I confirmed that the changes successfully fix https://github.com/eclipse-theia/theia/issues/11031 :+1:

Issue #10267, opened 2021-10-13
yarn download:plugins disregard proxy settings


Niklas-Dahlquist:
#### Bug Description:
Latest changes to download-plugins.ts doesn't follow proxy settings

#### Steps to Reproduce:

Do a yarn download:plugins behind a proxy that requires login in [Theia Blueprint](https://github.com/eclipse-theia/theia-blueprint), is it caused by Blueprint now since a vscode-builtin bundle was introduced, where the resolving kicks-in?

Fails during this step: [download-plugins.ts:125 console.warn('--- resolving extension-packs ---');](https://github.com/eclipse-theia/theia/blame/master/dev-packages/cli/src/download-plugins.ts#L125l)

```
--- resolving extension-packs ---
Error: connect ETIMEDOUT 198.41.30.195:443
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1141:16) {
  errno: 'ETIMEDOUT',
  code: 'ETIMEDOUT',
  syscall: 'connect',
  address: '198.41.30.195',
  port: 443
}
```

- Theia Version: 1.18.0

gbodeen:
We're now running into almost the same problem in our CI. The trouble appears to start with the `bent` library used in `dev-packages/ovsx-client/src/ovsx-client.ts` that isn't configurable to use proxy settings, from where it is used in the `getLatestCompatibleExtensionVersion` method, which is in turned called from within `dev-packages/cli/src/download-plugins.ts` when the `yarn download:plugins` script is run.

vince-fugnitto:
@gbodeen that's right, we would need to refactor `bent` to similarly use what we have in the `download:plugins` script which supports the proxy settings: https://github.com/eclipse-theia/theia/pull/7747.

ballcoach12:
FWIW, I am seeing the same problem.

msujew:
Just to give some context how this is fixed: The `download:plugins` command will automatically pick up the `http_proxy` and `https_proxy` env variables. Also, you can manually set the `--proxy-url <some http address>` argument, which will then take precedence over any existing env variables.

PR #11043, opened 2022-04-19
Support http proxies

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/7797
Closes https://github.com/eclipse-theia/theia/issues/10267

For some reason GitHub force-closed my old PR while force-pushing and doesn't allow me to reopen it, so I had to create a new PR. See https://github.com/eclipse-theia/theia/pull/10958 for previous discussion.

Adds general proxy support to the whole application (backend+frontend) by using a `RequestService` which comes with automatic proxy support. Other services and downstream applications are encouraged to use that service whenever requests to non-origin or non-localhost web-services are performed.

Also patches the `http` and `https` node.js imports for the extension host. This enables extensions to perform requests through proxies as well.

For electron there's currently a large TODO: Automatic OS proxy detection doesn't work yet, since that requires access to the electron-main process (see also https://github.com/eclipse-theia/theia/pull/10698).

#### How to test

Generally, testing this PR requires setting up firewall rules which block all outgoing traffic and configuring a local proxy rule. I used [Fiddler Classic](https://www.telerik.com/fiddler/fiddler-classic) (Windows only?) which sets itself up at `http://localhost:8888`. I assume Mac and Linux users have a simpler time configuring such a proxy.

Also, these tests should preferably performed using the Electron version. The browser version will automatically use the OS proxy for stuff like fetching vsx search results, whereas the electron version will fail if the proxy is not configured.

1. Open the vsx registry and search for `java`. Nothing should appear if outbound traffic is blocked.
2. Search in the preferences for `proxy` and set the `Http: Proxy` preference to your local proxy url.
3. Confirm that re-searching displays results (frontend request) and allows to install the extension (backend request).
4. Install the [`proxy-test` extension](https://github.com/eclipse-theia/theia/files/8380275/proxy-test-0.0.1.zip).
5. Execute the `Hello World Request` command. It will send a request to `https://www.google.com` and display the error message/success code.
6. Change the `Http: Proxy support` preference. `override`, `fallback` and `on` should all provide the same success result, while `off` should display an error, since the extension proxy support has been turned off.

---

There are now proxy settings for fetching packages from ovsx registries using the `theia download:plugins` command.

0. Delete your `plugins` folder.
1. Executing `theia download:plugins` should result in an error, since the outgoing download requests are blocked
2. Adding the `--proxy-url <url>` parameter should enable successful download of these extensions.
 
#### Review checklist

#### Reminder for reviewers


packages/core/src/node/request/backend-request-facade.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    configure(config: RequestConfiguration): Promise<void> {
+        return this.requestService.configure(config);
+    }
+
+    async request(options: RequestOptions, token?: CancellationToken): Promise<RequestContext> {
+        const context = await this.requestService.request(options, token);
+        // Convert the buffer to base64 before sending it to the frontend
+        // This reduces the amount of JSON data transferred massively
+        const base64Data = Buffer.from(context.buffer).toString('base64');
```
colin-grant-work, line 35: We actually have access to this API on the frontend for decoding, as well, although we shouldn't, in theory.

packages/core/src/node/request/proxy.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    return undefined;
+}
+
+export interface ProxySettings {
+    proxyUrl?: string;
+    strictSSL?: boolean;
+}
+
+export function getProxyAgent(rawRequestURL: string, env: typeof process.env, options: ProxySettings = {}): ProxyAgent | undefined {
```
colin-grant-work, line 38: It looks like this function is an exact copy of the one in `dev-packages` - can we import that one, instead?

packages/plugin-ext/src/hosted/node/plugin-host-proxy.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+
+    return {
+        http: Object.assign(http, createHttpPatch(http, resolveProxy, proxySetting, certSetting, true)),
+        https: Object.assign(https, createHttpPatch(https, resolveProxy, proxySetting, certSetting, true)),
+        tls: Object.assign(tls, createTlsPatch(tls))
+    };
+}
+
+function configureModuleLoading(lookup: PatchedModules): void {
```
colin-grant-work, line 64: We already have code that overrides `_load` in `plugin-vscode-init` and `backend-init-theia`. It should be safe, because whichever one happens to go first will become the other's `original` and whichever goes second should delegate to that and both should get to do their work, but it still makes me a bit queasy. Would it be possible to do this load override in the same place as the other?

packages/preferences/src/browser/util/preference-tree-generator.ts
```diff
@@ -58,6 +58,7 @@ export class PreferenceTreeGenerator {
         ['files', 'editor'],
         ['hosted-plugin', 'features'],
         ['keyboard', 'application'],
+        ['http', 'application'],
```
colin-grant-work, line 61: Alphabetization:
```suggestion
        ['hosted-plugin', 'features'],
        ['http', 'application'],
        ['keyboard', 'application'],
```

colin-grant-work (commented):
Comments on the code after a more thorough read through. I am setting up my proxy settings now to test the functionality...

colin-grant-work:
Functionally, I believe that this is working more or less as advertised. I'm not an expert on proxies or firewalls, but having set both up naïvely, traffic for Electron appears to be running through the configured proxy and making it through the firewall. @paul-marechal, I know you have some experience setting up proxies / throttlers on Linux, so if you'd be willing to give it a test, I'd appreciate it.

packages/core/src/node/request/proxy.ts
msujew: Actually, that was a leftover of when I refactored the request-service into its own package. I removed the file.

dev-packages/ovsx-client/src/ovsx-client.ts
```diff
@@ -37,11 +34,19 @@ export interface OVSXClientOptions {
 
 export class OVSXClient {
 
-    constructor(readonly options: OVSXClientOptions) { }
+    constructor(readonly options: OVSXClientOptions, readonly request: RequestService) { }
```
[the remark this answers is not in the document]
msujew: Nope, made it `protected`.

packages/core/src/browser/request/browser-request-service.ts
```diff
@@ -0,0 +1,177 @@
...
+        return this.backendRequestService.configure(config);
+    }
+
+    resolveProxy(url: string): Promise<string | undefined> {
+        return this.backendRequestService.resolveProxy(url);
+    }
+
+    protected transformBackendResponse(context: RequestContext): RequestContext {
+        // In the `backend-request-facade` we transform the binary buffer into a base64 string to save space
+        // We need to tranform it back into a binary buffer here
```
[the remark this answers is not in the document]
msujew: I added a `RequestContext.compress` and `decompress` method instead of the type. That way things should be a bit clearer.

packages/core/src/electron-browser/request/electron-browser-request-module.ts
```diff
@@ -0,0 +1,26 @@
...
+ ********************************************************************************/
+
+import { ContainerModule } from 'inversify';
+import { DefaultBrowserRequestService } from '../../browser/request/browser-request-service';
+import { RequestService } from '@theia/request-service';
+
+export default new ContainerModule(bind => {
+    // We bind the non-xhr request service here. This will always proxy every request through the backend.
+    // We do this since the backend currently cannot automatically resolve proxies, but the frontend can.
+    // We try to avoid confusion with this where some (frontend) requests successfully go through the proxy, but some others (backend) don't.
```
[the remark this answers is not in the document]
msujew: Yes, this won't be a problem once we can use the automatic proxy resolution through electron. We currently have to do this to not run into confusing issues, where parts of the app can go through the proxy (frontend) and others don't (backend).

packages/core/src/node/request/backend-request-facade.ts
msujew: Is the `Buffer` class actually available in the frontend? I thought it was just a stub.

packages/plugin-ext/src/hosted/node/plugin-host-proxy.ts
msujew: I think this will lead to convoluted code, due to two issues, mainly because we don't have the `PreferenceRegistryExtImpl` easily available in the scope the `plugin-vscode-init` and `backend-init-theia` functions are running in. Trying to configure the proxy support at the place the `vscode`/`theia` load gets overriden seems to be therefore quite difficult.

I get the queasyness, but I think that this is the least convoluted way to achieve this.

colin-grant-work: Sounds reasonable. Perhaps for a later refactoring of the plugin initialization system.

packages/core/src/node/request/backend-request-facade.ts
colin-grant-work: I believe it's fully functional. It works [in some extensions](https://github.com/eclipse-theia/theia-cpp-extensions/blob/d537453bdc6c08fea728c54c2f9c43d7428ebb19/packages/cpp-debug/src/browser/editable-widget/memory-editable-table-widget.tsx#L206). But what you've done above with `atob` is no doubt the correct way to handle things in the browser.

dev-packages/request-service/src/common-request-service.ts
```diff
@@ -0,0 +1,120 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { CancellationToken } from 'vscode-languageserver-protocol';
```
paul-marechal, line 17: I'm not sure we should use `CancellationToken` from `vscode-languageserver-protocol` here... I know we can't depend on `@theia/core` either.

Can we provide a minimal implementation with this package?

I was looking at [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) but it seems to only be supported on Node starting with 15.0.0.

dev-packages/request-service/src/common-request-service.ts
```diff
@@ -0,0 +1,120 @@
...
+     * Convert the buffer to base64 before sending it to the frontend.
+     * This reduces the amount of JSON data transferred massively.
+     * Does nothing if the buffer is already compressed.
+     */
+    export function compress(context: RequestContext): RequestContext {
+        if (context.buffer instanceof Uint8Array && Buffer !== undefined) {
+            const base64Data = Buffer.from(context.buffer).toString('base64');
+            Object.assign(context, {
+                buffer: base64Data
+            });
```
paul-marechal, line 89: I assume you need `Object.assign` to by-pass the type error from assigning a `string` to a `Uint8Array`?

Not sure what's the idea here.

paul-marechal (commented):
It looks mostly good to me, just a few oddities.

dev-packages/request-service/src/common-request-service.ts
msujew: Yeah, that's the idea. We could also use `as unknown as Uint8Array`, but I'm not too fond of that. Any recommendations on how to improve this?

dev-packages/request-service/src/common-request-service.ts
msujew: Makes sense 👍 I copied the `CancellationToken` interface, so that users of Theia can use the `CancellationToken` implementation from the core package.

dev-packages/request-service/src/common-request-service.ts
msujew: Uhm, you're right. `RequestContext.buffer` is now `Uint8Array | string` 👍

paul-marechal (approved):
Looks good as first iteration. We can take care of the missing bits in Electron later with IPC/RPC.

vince-fugnitto:
@msujew I believe the proxy part of `download:plugins` should actually fix https://github.com/eclipse-theia/theia/issues/10267 😃

Issue #10895, opened 2022-03-17
Dock panel's `currentTitle` is not reset to `undefined` when the last widget is moved elsewhere


lucas-koehler:
### Bug Description:
After the last tab of a `TheiaDockPanel` in the shell (e.g. main or bottom) was moved to another area (e.g. main to bottom), the `currentTitle` property is not set to `undefined` but still points to the moved tab. It seems that the panel's `markAsCurrent` method is never called to reset to `undefined`.
This does not happen if the last widget is closed because the `currentTitle` is properly reset then.

The bug can be observed via the `Close Tab in Main Area` command. It is still active after the last widget was moved from the main to another area. Executing the command closes the moved widget despite it not being in the main area.

### Steps to Reproduce:

1. Start with empty main area
2. Open editor
3. Move editor to bottom area
4. Execute command `Close Tab in Main Area`
5. The editor will be closed in the bottom area

### Additional Information

https://user-images.githubusercontent.com/6959840/158841248-be1bf395-4b34-49ff-b50b-7897821128f6.mp4

- Operating System: Ubuntu 20.04
- Theia Version: 1.23.0, commit 441136570fe7ddb3a99408df0acbef9f31a9bdd7

Issue #10526, opened 2021-12-09
Multi Window support (Floating windows)


sdirix:
# Summary

We would like to have fully featured multi window, a.k.a. floating window support in Theia. This issue proposes a specific approach and may serve as the `Epic` to implement this feature in full.

We evaluated the approach with a PoC and created an initial list of development tasks. Also we added a rough risk assessment for each of the determined challenges.

We are looking for feedback in regards to the suggested approach and the overall assessments. Feel free to add suggestions on how to solve the determined challenges and please notify us when we missed a whole concept or even only a minor detail.

# Goal

* Views shall be movable into secondary windows.
* Views shall behave the same no matter whether they are contained in the main or in a secondary window

## Status Quo

Theia offers a ["New Window" action](https://github.com/eclipse-theia/theia/blob/c4c6cab037230fc05e39c7b124a681630c2f6a33/packages/core/src/browser/window-contribution.ts#L26-L30) which opens an additional window. In this window a complete new Theia frontend is instantiated, including, among others, the inversify container, frontend application and application shell.

![MultipleFrontendsDetail](https://user-images.githubusercontent.com/8998368/145390519-a77fe7ab-f75a-497c-9a3d-36e80a4968a8.png)

The diagram above shows this case in a simplified way. The result is two completely separated frontend instances which do not share the same "UX context", e.g. they don't share a common selection or debugging state. They don't even share the same plugin process on the backend. They _almost_ behave like two completely different instances of the same Theia application, with the difference that they share some singletons via the backend. For example the "running task" list is synchronized between these two frontends via the backend.

### Downsides of Status Quo

* Views can not be dragged and dropped between the windows
* The same file or view can be opened in both windows, but they belong to their respective frontend instance. So for example the errors view will always be scoped to the corresponding window.

Overall it resembles a multi instance rather than a multi window approach

### Comparison to other frameworks

A good implementation of multi window support can be found in Eclipse RCP.

![EclipseRCPExample](https://user-images.githubusercontent.com/8998368/145390553-974d0110-eeca-4b41-b54e-dd6cae20f02a.png)

Here the user is able to detach windows from the main window into (one or more) secondary windows. These share the same context as the main window. In the screenshot above you can see the detached console and debug view which share the same state with the main window. Ideally we would like to see similar behavior in Theia.

Looking at VS Code, [multi window support](https://github.com/Microsoft/vscode/issues/10121) is one of the most requested features for years.

# Web technology

[...]

PR #11048, opened 2022-04-20
Multi window support for web views

lucas-koehler:
#### What it does

This implements the multi window MVP proposal desribed in https://github.com/eclipse-theia/theia/issues/10526#issuecomment-1033950569

Support for moving webview-based views into a secondary window or tab.
For webview-based views a new button becomes available in the toolbar of the view to move the view to a secondary window.
This is only supported for webview-based views and only one view can be moved into the secondary window.
There can be multiple secondary windows though.

Primary code changes:
- Add concept of extractable widgets. Only widgets implementing the interface can be extracted.
- Add SecondaryWindowHandler that encapsulates logic to move widgets to new windows.
- Only webviews can be extracted
- Configure opened secondary windows in electron
  - Hide electron menu
  - Always use the native window frame for secondary windows to get window controls
  - Do not show secondary window icon if main window uses a custom title bar
- Contribute widget extraction button in a separate new extension `widget-extraction-ui`
- Extend application shell areas with a `secondaryWindow` area that contains all extracted widgets
- Extend frontend and webpack generators to generate the base html for secondary windows and copy it to the lib folder
- Bridge plugin communication securely via secure messaging between external webview and Theia:
  Webviews only accept messages from its direct parent window. This is necessary to avoid Cross Site Scripting (XSS).
  To make the messaging work in secondary windows, messages are sent to the external widget and then delegated to the webview's iframe.
  Thereby, the secondary window only accepts messages from its opener which is the theia main window.
  To achieve this, a webview knows the secondary window it is in (if any). It then sends messages to this window instead of directly to the webview iframe.
- Patch phosphor library during install via a postinstall hook. Remove check that widget attachment target must be in same DOM.

Contributed on behalf of ST Microelectronics and Ericsson and by ARM and EclipseSource.

Co-authored-by: Stefan Dirix <sdirix@eclipsesource.com>
Co-authored-by: robmor01 <Rob.Moran@arm.com>
Signed-off-by: Lucas Koehler <lkoehler@eclipsesource.com>

https://user-images.githubusercontent.com/6959840/164273386-8c032eac-72a3-4f97-ad7d-3e7966285bb7.mp4

#### Limitations

- Electron is unsupported. Generally, the functinality is existent but there are some issues with close handling of secondary windows.
- Only web views are supported. Plugin views don’t use webviews/iframes and thus cannot be extracted as part of this MVP.
- Widgets/editors that do not support moving from one area to another (e.g. main to bottom) also do not support moving to secondary windows
- If the last widget of an area is moved to a secondary window, it is still tracked as the former area’s current widget. This is caused by #10895 and also happens when moving widgets inside the main window
- When a webview in a secondary window opens a file picker - or possibly another dialog - it is opened in the main window instead of the secondary one.
- While an extracted webview resides in a secondary window, no contributed toolbar items contributed by the plugin are shown above the webview. This is caused by the absence of subsidiary UI (e.g. tabs, toolbar) in secondary windows. Fortunately, that isn't a very common use case since  webviews can integrate any additional UI necessary.
- External views cannot be restored to the main window, yet. To get a view back to the main window it needs to be closed and re-opened
- Windows are not restored on reload because they are not part of the stored application layout.
- extracted views do not respect "autoSave" (ex: for custom editors)

#### Screenshots
<details>
<summary><strong>Ubuntu (click to expand)</strong></summary>

titlebar-ubuntu-custom-native
titlebar-ubuntu-native-native
</details>

<details>

<summary><strong>Windows 11 (click to expand)</strong></summary>

titlebar-win11-custom-native-dark-no-icon
titlebar-win11-native-native

</details>

<details>
<summary><strong>Mac (click to expand)</strong></summary>

mac
</details>

#### How to test

- Get one or more VS Code extensions that contribute web views and drop them in the `plugins` folder. For instance:
  - [Protein viewer](https://marketplace.visualstudio.com/items?itemName=ArianJamasb.protein-viewer), you can use protein `2btf` as example input
  - [Draw.io integration](https://marketplace.visualstudio.com/items?itemName=hediet.vscode-drawio)
  - [Mermaid Editor](https://marketplace.visualstudio.com/items?itemName=tomoyukim.vscode-mermaid-editor)
  - Build the webview sample providing a cat coding view from [here](https://github.com/microsoft/vscode-extension-samples/tree/main/webview-sample)
- If you use other extensions, make sure they support being moved around in the shell (e.g. from main to bottom) because webviews that don't support that won't support being moved to a new window either
- Open a web view and click the window button in its tool bar:
how-to-move

#### Review checklist

#### Reminder for reviewers


colin-grant-work (commented):
With the specified limitations, this appears to be working as advertised. There are some jarring bits, depending on the behavior of different webviews. For example, the Java plugin opens a webview to allow the user to configure their JDK settings, and that webview can open a file picker, but the file picker appears in the main window rather than secondary window. That's expected at the moment, but isn't optimal UI, by any means.

Otherwise, minor comments only.

packages/secondary-window-ui/src/browser/secondary-window-ui-frontend-contribution.ts
```diff
@@ -0,0 +1,60 @@
...
+import { inject, injectable } from '@theia/core/shared/inversify';
+import { ApplicationShell } from '@theia/core/lib/browser/shell';
+import { CommandRegistry, CommandContribution, Command } from '@theia/core/lib/common/command';
+import { codicon, ExtractableWidget } from '@theia/core/lib/browser/widgets';
+import { TabBarToolbarContribution, TabBarToolbarRegistry } from '@theia/core/lib/browser/shell/tab-bar-toolbar';
+
+export const EXTRACT_WIDGET = Command.toDefaultLocalizedCommand({
+    id: 'extract-widget',
+    label: 'Move View to Secondary Window'
+});
```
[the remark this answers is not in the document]
msujew: We have another method for normal localized commands, usage:

```ts
export const EXTRACT_WIDGET = Command.toLocalizedCommand({
    id: 'extract-widget',
    label: 'Move View to Secondary Window'
}, 'theia/secondary-window-ui/extract-widget');
```

lucas-koehler:
@colin-grant-work Thank you very much for the quick review. I pushed two commits to address the issues and add machine translations for the new command.
I added the file picker issue to the limitations so we remember it and can open a follow up issue. Is that alright with you?

colin-grant-work:
Another UI limitation of the current iteration - also related to the absence of subsidiary UI - is the absence of any toolbar items contributed by the plugin to be shown above its views. That isn't a very common use case since the webview can integrate any additional UI necessary, but it's worth noting along with the file-picker.

lucas-koehler:
@colin-grant-work That is a good point, thanks. I added the following limitation to the PR description:
> - While an extracted webview resides in a secondary window, no contributed toolbar items contributed by the plugin are shown above the webview. This is caused by the absence of subsidiary UI (e.g. tabs, toolbar) in secondary windows. Fortunately, that isn't a very common use case since  webviews can integrate any additional UI necessary.

koegel:
@colin-grant-work Do you feel comfortable in approving this change or is there open issues to be fixed in your opinion or should we get an additional pair of eyes? Thanks!

colin-grant-work:
@koegel, I'll give it another run through tomorrow to make sure there aren't any regressions that I've missed, and if there aren't then I'll approve.

koegel:
Perfect, thanks!

packages/core/src/browser/secondary-window-handler.ts
```diff
@@ -0,0 +1,249 @@
...
+    readonly onDidAddWidget = this.onDidAddWidgetEmitter.event;
+
+    protected readonly onDidRemoveWidgetEmitter = new Emitter<Widget>();
+    /** Subscribe to get notified when a widget is removed from this handler, i.e. the widget's window was closed or the widget was disposed. */
+    readonly onDidRemoveWidget = this.onDidRemoveWidgetEmitter.event;
+
+    constructor(
+        @inject(MessageService) protected readonly messageService: MessageService,
+        @inject(WindowService) protected readonly windowService: WindowService
+    ) { }
```
vince-fugnitto, line 78: I believe we should use property injection here?

- https://github.com/eclipse-theia/theia/wiki/Coding-Guidelines#property-injection

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -219,6 +220,7 @@ export class ApplicationShell extends Widget {
         @inject(ApplicationShellOptions) @optional() options: RecursivePartial<ApplicationShell.Options> = {},
         @inject(CorePreferences) protected readonly corePreferences: CorePreferences,
         @inject(SaveResourceService) protected readonly saveResourceService: SaveResourceService,
+        @inject(SecondaryWindowHandler) protected readonly secondaryWindowHandler: SecondaryWindowHandler,
```
vince-fugnitto, line 223: I believe this is a breaking change that should be documented:
- https://github.com/eclipse-theia/theia/wiki/Coding-Guidelines#property-injection

dev-packages/application-manager/src/generator/frontend-generator.ts
```diff
@@ -197,4 +198,48 @@ module.exports = Promise.resolve()${this.compileElectronMainModuleImports(electr
...
+        if (e.source === window.opener) {
+            // Delegate message to iframe
+            document.getElementsByTagName('iframe').item(0).contentWindow.postMessage({ ...e.data }, '*');
+        }
+    });
+    </script>
+</head>
+
+<body>
+    <div id="pwidget"></div>
```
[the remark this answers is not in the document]
lucas-koehler: I renamed this to `widget-host` as PhoshporJS describes the element to attach a widget to as host. I hope that's alright with you. Otherwise, I can rename it again.

lucas-koehler:
Hi @vince-fugnitto , thank you for your review :)
I applied your comment and added a new 1.26.0 section to the changelog

koegel:
@vince-fugnitto Thank you for taking the time to review this. I think the three missing features you point out are all valid and important.

We have aimed to rather release an MVP of the MDI earlier with less features than a feature-complete version of it later. In addition the functionality is opt-in, meaning you will have to add the extension to your product to make use of it, so you will be aware of its limitations. Therefore we have stripped down this PR to the bare minimum. In this sense "Restore View" and "Restore Window on Reload" are imho features that do not have to be part of the MVP. In my opinion the functionality is still valuable without them and is far more than you can achieve with VS Code today.

Some more reasons why I believe we should rather release this earlier with less features than later with more features:
- Might drive adoption of Theia: This is a textbook case of where Theia is more powerful than VS Code. This feature is ranked high by users of VS Code but there seems to be no intent to get it implemented for VS Code.
- Might attract contributions: Once this features becomes visible there is a good chance we might see contributions by others that improve and enhance it.
- Prioritization of additional features might change once this feature is in use: For example I could imagine that users would rather like to have support for the terminal in secondary windows than being able to restore view on reload and both might have similar effort. 

I would therefore propose the following: 
- We will take a look whether "SecondaryWindowHandler Improvement" can be fixed without additional changes to Phosphor (which is to be avoided) and with reasonable additional effort.
- For the other two features we would also do a ball-park estimation and open up follow-up issues to extract them from this PR.
- We amend the information about limitations of this feature with your comments so adopters know exactly what they get.

What do you think?

vince-fugnitto:
> I would therefore propose the following:
> 
> - We will take a look whether "SecondaryWindowHandler Improvement" can be fixed without additional changes to Phosphor (which is to be avoided) and with reasonable additional effort.
> - For the other two features we would also do a ball-park estimation and open up follow-up issues to extract them from this PR.
> - We amend the information about limitations of this feature with your comments so adopters know exactly what they get.
>
> What do you think?


@lucas-koehler I don't mean to hold back the pull-request because the basic features are missing, just wanted to highlight them and give my first impressions as a user which others might experience as well. 

I'm fine with the way forward, we still have the month to polish the mvp and possibly implement these features which I do think are generic and basic and will benefit all future views. The mvp already only handles a subset of all use-cases (webviews), which others may not find all that useful, so I thought that making sure the features are present will only help adoption.

lucas-koehler:
Hi @vince-fugnitto , thank you very much for your input.

One question regarding "SecondaryWindowHandler Improvement": Did you observe this in electron or the browser. For electron this is already known but in the browser bringing up the minimized window works. If in the browser, which browser did you use?

For the other two points, I added the following limitations to the PR description:
>- External views cannot be restored to the main window, yet. To get a view back to the main window it needs to be closed and re-opened
>- Windows are not restored on reload because they are not part of the stored application layout.

vince-fugnitto:
> One question regarding "SecondaryWindowHandler Improvement": Did you observe this in electron or the browser? For electron this is already known but in the browser bringing up the minimized window works for me. If in the browser, which browser did you use?

@lucas-koehler I observed it for electron, have you investigated as to why it does not work in electron but does in the browser?

lucas-koehler:
> @lucas-koehler I observed it for electron, have you investigated as to why it does not work in electron but does in the browser?

@vince-fugnitto When an external widget is revealed by the application shell, the secondary window handler calls `focus` on the widget's external `Window`. In the browser this is sufficient to bring the window to the front. However, I do not know why this does not work for electron.

vince-fugnitto:
> @vince-fugnitto When an external widget is revealed by the application shell, the secondary window handler calls `focus` on the widget's external `Window`. In the browser this is sufficient to bring the window to the front. However, I do not know why this does not work for electron.

@lucas-koehler I believe the bug should be fixed for the mvp, it is likely due to a design problem of how we deal with windows in browser vs electron. For instance, we have the `window-service` today which properly handles each target separately that we may want to enhance on how to create new windows, and focus them.

lucas-koehler:
> @lucas-koehler I believe the bug should be fixed for the mvp, it is likely due to a design problem of how we deal with windows in browser vs electron. For instance, we have the `window-service` today which properly handles each target separately that we may want to enhance on how to create new windows, and focus them.

@vince-fugnitto Seems reasonable. I will investigate this later this week.

lucas-koehler:
Hello @vince-fugnitto ,
I solved the electron focus issue by adding a new service `SecondaryWindowService` that handles creating and focussing the secondary windows. I preferred a separate service because the secondary window creation has not much in common with opening other windows. But if you prefer, I can also integrate this in the existing `WindowService`.

I squashed the existing commits from the older review comments and added a separate commit for the new service to hopefully make it easier to review :)

For the other two issues, I'd stick with opening follow up issues along with the other limitations after this was merged.

vince-fugnitto (commented):
The changes work well for me, we can open up the follow-up issues.
I have some minor comments, and I would like others to take a look at the changes as well.

lucas-koehler:
@vince-fugnitto I fixed the minor comments. I'll address the phosphor comment here for better visibility.

> I believe this change will actually not work downstream, and means that the feature is unusable for them?
We are only patching for ourselves, what do we expect downstream to do?

That is a very good catch, thanks! Besides actually forking PhosphorJS - which was determined not to be viable for this MVP (see [here](https://github.com/eclipsesource/theia/issues/29#issuecomment-1060519887) for reasons) - I see two possibilities to solve this:

1. Provide a quick guide for users to apply the given patch in their Theia instance (adding the script and adding the postinstall to their root package.json)
2. Adapt the webpack file generated in `webpack-generator.ts` to remove the problematic line from PhosphorJS during the build.

Initially we went with the separate script because it is easier to extend and see what exactly is happening. However, requiring users to manually adapt their package.json might not be the best way to go.

IMO, option number 2 is the way to go. Downstream users don't need to do anything and no files are altered on the user's file system. This is also safe against `postinstall` scripts potentially not being run by default in the future. As PhosphorJS is archived, the webpack adaption should also be stable because Phoshpor does not change.

Does that sound like an acceptable solution to you?

package.json
```diff
@@ -69,6 +69,7 @@
     "lint:clean": "rimraf .eslintcache",
     "lint:oneshot": "node --max-old-space-size=4096 node_modules/eslint/bin/eslint.js --cache=true \"{dev-packages,packages,examples}/**/*.{ts,tsx}\"",
     "preinstall": "node-gyp install",
+    "postinstall": "node scripts/patch-libraries.js",
```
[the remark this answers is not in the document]
lucas-koehler: See https://github.com/eclipse-theia/theia/pull/11048#issuecomment-1128633471

vince-fugnitto:
> Does that sound like an acceptable solution to you?

@lucas-koehler if the second option works then I'd be fine with it :) Initially I had thought the best solution would be to fork the repo, even if in the long run we want to move away from it. Alternatively, like 1 we could have the script under `cli` but I also find that undesirable for downstream apps. 

Let's go with your second proposal :+1:

lucas-koehler:
@vince-fugnitto I implemented the PhosphorJS by extending the genrated webpack config for applications. I tested this on Ubuntu and in a Windows VM and the patch works for me. I also removed the postinstall hook and the script again as we no longer need it.

dev-packages/application-manager/src/generator/webpack-generator.ts
```diff
@@ -104,6 +111,16 @@ module.exports = {
     cache: staticCompression,
     module: {
         rules: [
+            {
+                // Removes the host check in PhosporJS to enable moving widgets to secondary windows.
```
vince-fugnitto, line 115:
```suggestion
                // Removes the host check in PhosphorJS to enable moving widgets to secondary windows.
```

packages/secondary-window-ui/src/browser/secondary-window-ui-frontend-contribution.ts
```diff
@@ -0,0 +1,60 @@
...
+    registerCommands(commands: CommandRegistry): void {
+        commands.registerCommand(EXTRACT_WIDGET, {
+            execute: async widget => {
+
+                // sanity check
+                if (!ExtractableWidget.is(widget)) {
+                    // command executed with a non-extractable widget
+                    console.error('Invalid attempt to move non-extractable widget to secondary window.');
+                    return;
+                }
```
vince-fugnitto, line 44: Is the check necessary? Both `isVisible` and `isEnabled` work only on `ExtractableWidget`, and `moveWidgetToSecondaryWindow` itself also checks for extractability.

vince-fugnitto (commented):
I confirmed that the changes work well for both the browser and electron.
As I mentioned earlier I would like the review from others for the changeset in case they have feedback or notice issues with the implementation.

lucas-koehler:
@vince-fugnitto Thank you for reviewing and testing this again. I fixed your two small remarks :)
I'll rebase and solve the `CHANGELOG.md` after the other reviews as I suspect we won't merge this for this milestone anymore and thus will adapt the changes again anyway.

lucas-koehler:
I rebased the changes on the current master.

JonasHelming:
@vince-fugnitto @colin-grant-work @msujew can we merge this? We should probably merge this some days before the release.

colin-grant-work (approved):
From my perspective, it is ok to merge this. I've spent some time using / developing on this branch, and I haven't seen any regressions. I agree with @vince-fugnitto that moving widgets back into the main window is an important use case and should be addressed as this feature moves along.

JonasHelming:
@msujew : Would you mind looking at this soon, I do not want to merge this too late in the dev cycle.

msujew (commented):
I noticed an issue when trying to reopen a closed secondary window:

https://user-images.githubusercontent.com/4377073/173611166-3bd6aac4-7aa2-4115-a820-08231ab738c8.mp4

The error displayed in the console is:

```
2022-06-14T15:02:39.610Z root WARN There is no known secondary window ''. Thus, the window could not be focussed.
2022-06-14T15:02:41.632Z root WARN Widget was activated, but did not accept focus after 2000ms: plugin-webview:d1016297-4a62-446a-b7fc-5e94977eb97c
```

lucas-koehler:
> I noticed an issue when trying to open a webview after closing the secondary window:
> 
>  2022-06-14.17-02-06.mp4 
> The error displayed in the console is:
> 
> ```
> 2022-06-14T15:02:39.610Z root WARN There is no known secondary window ''. Thus, the window could not be focussed.
> 2022-06-14T15:02:41.632Z root WARN Widget was activated, but did not accept focus after 2000ms: plugin-webview:d1016297-4a62-446a-b7fc-5e94977eb97c
> ```

@msujew Thank you very much for testing and the showcase video. I can reproduce the problem. It seems to be a problem with window listeners not being invoked. I am going to investigate further in the beginning of next week.

lucas-koehler:
Update: I rebased the changes locally and fixed the behavior on Electron. However, there are now/still some issues with the browser version regarding the re-opening of web views.

lucas-koehler:
@msujew I finally got to investigating this further and could fix the electron and browser issues. For me everything now seemed to work as expected on Ubuntu 20.04 and a Windows 11 VM.

lucas-koehler:
Rebased to fix merge conflicts with master branch.

msujew:
@lucas-koehler It works for browser flawlessly now, but I'm getting another error on Electron (Windows 11) now: When closing the external webview window, I get an `Uncaught illegal access` error in the main window console (without a stack trace). Afterwards I cannot open any webviews any more. Can you reproduce that?

lucas-koehler:
> @lucas-koehler It works for browser flawlessly now, but I'm getting another error on Electron (Windows 11) now: When closing the external webview window, I get an `Uncaught illegal access` error in the main window console (without a stack trace). Afterwards I cannot open any webviews any more. Can you reproduce that?

@msujew I could not at first but after a few retries I could reproduce this. Strangely, for me the `Uncaught illegal access` does not occur every time and even if it occurs, it does not always block opening web views. Could you also observe this, too, or does it happen consistently for you?

msujew:
@lucas-koehler Interesting, it's 100% reproducible for me. It happens every time I open the webview in a secondary window 🤔

lucas-koehler:
> @lucas-koehler Interesting, it's 100% reproducible for me. It happens every time I open the webview in a secondary window thinking

@msujew Ok, thanks. One last question because I think that might be relevant: Which node version do you use and which electron version is installed in node_modules\electron\package.json for you? TIA!

msujew:
> Which node version

`node -v` returns `16.16.0` for me

> which electron version

`15.5.7` which is referenced in our `yarn.lock` file.

koegel:
I have a suggestion: This feature is not active by default in a Theia-based product since you need to install the extension. In addition currently it is limited to web views and does not work with Monaco-based editors for example. So I propose to also limit it to web browser-only and postpone the support in electron for this PR. This would help to make the feature finally available and open it to more feedback from the community. For support in electron (and for different views) we will continue our work on a new issue with a high priority. 
To implement this suggestion, we would add a note to the Readme of the extension to make people aware that this extension is currently experimental, only targeting the browser and does only support web views. In addition we would rename the package `secondary-windows-ui-browser-only-experimental` to highlight to potential consumers what they can expect. (The extension will also **not** be available in the open VSX marketplace.)
However the extension would be available in the example app but only in the browser build.
What do you think?

koegel:
> I'm confused, why would it ever be in Open VSX Marketplace? The `secondary-windows-ui` is a theia extension (npm dependency) and not a vscode plugin.
You are right, it will *never* be in the Open VSX marketplace :). I was trying to rule out how potential users could come in contact with the feature.

koegel:
> I don't think a rename is necessary, only a note in the readme for the time being about what the extension supports/provides. The unfortunate consequence is a lot of changes were already necessary in the framework to support only the use-case of webviews, and now we won't support Electron to start.

Yes, but we will eventually make use of the code for Electron. We still intend to work on the electron implementation with a high priority.

msujew:
@koegel @lucas-koehler I'm fine with that. Getting the bulk of the work in now and fixing anything that's not working in this version in a later PR sounds like a good idea to me.

lucas-koehler:
I rebased the change onto master, removed the `secondary-window-ui` extension from the electron example app and added the limitation to the extension's readme.
I created issue https://github.com/eclipse-theia/theia/issues/11574 as a follow up to fix the electron behavior.

lucas-koehler:
Rebased changes onto 1.29.0

vince-fugnitto (commented):
@lucas-koehler @koegel given that the **electron** use-case is not yet supported, should we remove the electron-specific changes from this pull-request and add them when it is supported? It feels odd to me to include electron-specific changes when it does not work.

I do wonder if `@theia/secondary-window-ui` should be made more generic (possibly `@theia/secondary-window` as we might add non-UI functionality in that package later on) I also wonder if a lot of the secondary window changes can actually be moved to that package instead of in `core` given that all the code is actually dormant unless the package is consumed.

Am I correct that these are the latest limitations (we may want to update the pr description with the latest state):

- [ ] electron is unsupported
- [ ] only **webviews** (contributed by plugins) are supported
- [ ] only extraction is supported, does not allow views to return to the main window
- [ ] does not support possible tabbar or menu support
- [ ] secondary windows does not restore on startup
- [ ] possible actions from the webviews appear in the wrong window (ex: dialogs)
- [ ] secondary windows are incorrectly tracked if the last widget (ex: "close all tabs in the main area" can potentially close secondary windows)
- [ ] extracted views do not respect "autoSave" (ex: for custom editors)

I also noticed that the `open editors` tree-view should be updated with a better name than `secondarywindow` (all one word) for end-users:

Screen Shot 2022-08-31 at 12 32 05 PM

I've also noticed that secondary windows do not display the workspace and application name the same way the main window does. If it is intended it unfortunately makes it harder to distinguish which secondary window belongs to which workspace.

lucas-koehler:
Hello @vince-fugnitto ,
thank you for your extensive feedback :)
I updated the limitations section in the PR. I also squashed the commits before your review and rebased onto master.

As I believe you discussed yesterday with @koegel, I incorporated parts of your suggestions:

- The extension was renamed to `secondary-window`
- The `open editors` tree view now shows `IN SECONDARY WINDOW` instead of `SECONDARYWINDOW` (see screenshot below)
- The external windows now append the workspace and application name (i.e. the main window's title) to their title (see screenshot below)

> I also wonder if a lot of the secondary window changes can actually be moved to that package instead of in core given that all the code is actually dormant unless the package is consumed.

This was omitted because the application shell uses the secondary window handler in multiple places. If we move it to the other extension and make it optional we need to check and handle its absence in various places.

> It feels odd to me to include electron-specific changes when it does not work.

They were left in to save the effort of extracting them just to re-add them again soon. However, the `moveWidgetToSecondaryWindow` method was removed from the `application-shell` and the secondary window handler was marked as experimental. This should avoid adopters using this functionality "by accident".

I hope this works for you and thanks again for your thoughts :)

UI update:
image

packages/secondary-window/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@theia/secondary-window",
+  "version": "1.29.0",
+  "description": "Theia - Secondary window UI contributions",
```
vince-fugnitto, line 4: minor:

```suggestion
  "description": "Theia - Secondary Window Extension",
```

vince-fugnitto (approved):
The changes look good to me given the known limitations. Once merged we should open issues to track them.

lucas-koehler:
@vince-fugnitto Thank you for the fast review. I squashed the commits and added your suggestion for the extension title

vince-fugnitto:
@lucas-koehler I've opened a couple of follow-up issues to track the known limitations and bugs. If you want to do the same that would be appreciated.

lucas-koehler:
@vince-fugnitto Thank you for your effort and the reminder :)
I believe I added the remaining issues based on the limitations listed in the PR.

PR #11053, opened 2022-04-21
Ensure `Disposable.NULL` always returns a new `Disposable`

colin-grant-work:
#### What it does

Because of the way `DisposableCollections` handle disposal of their members, it is unsafe to add `Disposable.NULL` to any disposable collection: doing so risks disposal of one collection marking another collection as disposed inappropriately. This PR adds a test showing the risk of `Disposable.NULL` and replaces `Disposable.NULL` throughout the framework with a new method `Disposable.createNull`, with a test to demonstrate that disposables created in that way are safe for use in `DisposableCollections`.

#### How to test

1. Assert that the tests in [`disposable.spec.ts`](https://github.com/eclipse-theia/theia/pull/11053/files#diff-c94e7d7c6a8101807f20839ba79577f5565e362b980f004ff2d5f04bb0ff3484) pass (and look correct)
2. Check that changes other than `disposable.ts` and `dispospable.spec.ts` just replace `Disposable.NULL` with `Disposable.createNull`
3. Check that no references to `Disposable.NULL` remain in the framework.
4. Check that the application behaves as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/disposable.ts
+++ b/packages/core/src/common/disposable.ts
@@ -30,9 +30,14 @@ export namespace Disposable {
     export function create(func: () => void): Disposable {
         return { dispose: func };
     }
+    /** Always provides a reference to a new disposable. */
     export declare const NULL: Disposable;
 }
 
+/**
+ * Ensures that every reference to {@link Disposable.NULL} returns a new object,
+ * as sharing a disposable between multiple {@link DisposableCollection} can have unexpected side effects
+ */
 Object.defineProperty(Disposable, 'NULL', {
     configurable: false,
     enumerable: true,
--- a/packages/core/src/common/disposable.ts
+++ b/packages/core/src/common/disposable.ts
@@ -28,19 +28,18 @@ export namespace Disposable {
         return !!arg && typeof arg === 'object' && 'dispose' in arg && typeof arg['dispose'] === 'function';
     }
     export function create(func: () => void): Disposable {
-        return {
-            dispose: func
-        };
+        return { dispose: func };
     }
-    /**
-     * @deprecated since 1.25 Disposable.NULL should not be used because adding it to DisposableCollections may produce side effects in other collections
-     * Use Disposable.createNull instead.
-     */
-    export const NULL = create(() => { });
-    export function createNull(): Disposable {
-        return create(() => { });
-    }
+    export declare const NULL: Disposable;
 }
+
+Object.defineProperty(Disposable, 'NULL', {
+    configurable: false,
+    enumerable: true,
+    get(): Disposable {
+        return { dispose: () => { } };
+    }
+});
 
 export class DisposableCollection implements Disposable {
 
--- a/packages/core/src/common/disposable.ts
+++ b/packages/core/src/common/disposable.ts
@@ -32,7 +32,14 @@ export namespace Disposable {
             dispose: func
         };
     }
+    /**
+     * @deprecated since 1.25 Disposable.NULL should not be used because adding it to DisposableCollections may produce side effects in other collections
+     * Use Disposable.createNull instead.
+     */
     export const NULL = create(() => { });
+    export function createNull(): Disposable {
+        return create(() => { });
+    }
 }
 
 export class DisposableCollection implements Disposable {
--- a/packages/core/src/browser/color-registry.ts
+++ b/packages/core/src/browser/color-registry.ts
@@ -59,7 +59,7 @@ export class ColorRegistry {
     }
 
     protected doRegister(definition: ColorDefinition): Disposable {
-        return Disposable.createNull();
+        return Disposable.NULL;
     }
 
 }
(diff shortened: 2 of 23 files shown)


paul-marechal:
I honestly dislike what's being done to disposable references in `DisposableCollection`: it mutates instances with a new `dispose` function, and then tries to revert it somehow. Instead we should consider the instances as managed by the collection and assume `dispose` will only be called once by the collection and no one else.

Anyway, I would propose two alternative changes:

1. Make `Disposable.NULL` return new instances on each access: see [playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCLAM4AcD2G4BGANigN4BQyVyAJprhhABQCUAXMgG47A3kC+5ciDgBbCNkQp02PIRLIK1ZBAAeuKGGQIcIDFow5xABSg4syALzIAjAG4B5GhARE4UFCPGSkaenOIySmo1DS0dPS0AOQBVABk4jhwCACsXMEcAeVT0gDpnGFAIU3NoMABPJhkGeQgAGmQAcliExoalEJAAV3EoWo4wKC764KoIwoBzLr7AjngiRjrR5AmIMFZFZeUPMGmQTeVD6jpZRg3SQSOqS+pBfhYhCP1kOCs-U9rcwxMzLAcnrQEN7VAIkXItOL-XTPBDA-z4QLg+KQx7QnBgog4CZMRpwNovFhQvToiC5THYxpAyzU5AEfFUmkEB5PElkrE4hnWBD0qw0hAPIA).
2. Make `DisposableCollection` aware of `Disposable.NULL` and ignore it when pushed to it.

Those alternatives both have the benefit of requiring no changes for downstream projects. WDYT?

colin-grant-work:
> Is this related to the fact that `Disposable.NULL` always returns the same disposable instance, and `DisposableCollection` is messing with the reference of disposables added to it?

Yes, exactly. 

> I honestly dislike what's being done to disposable references in `DisposableCollection`.

Yeah, it's certainly not ideal. I'm not exactly sure what system to put in its place: if the point of disposables is to avoid leaks, then having some mechanism for removing references to them that occur outside of the collection currently disposing of itself is probably bad, but it probably doesn't make sense to have two different collections that can both dispose of something the holder of the other collection might be depending on. So it would probably make sense to guarantee that no disposable is ever put in two disposable collections, but I'm not sure how frequently that's happening in the wild or how much fixing it would take to ensure it doesn't happen.

> Anyway, I would propose two alternative changes:
> 
> 1. Make `Disposable.NULL` return new instances on each access: see [playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCLAM4AcD2G4BGANigN4BQyVyAJprhhABQCUAXMgG47A3kC+5ciDgBbCNkQp02PIRLIK1ZBAAeuKGGQIcIDFow5xABSg4syALzIAjAG4B5GhARE4UFCPGSkaenOIySmo1DS0dPS0AOQBVABk4jhkGeQhHAHkCACsXMAA6ZxhQCFNzaDAATyZkgJIAGmQAcliExoalagiigHMAVyhUjngiRjrgqggQXvEBwI4wKF6IMeVuiDBWRXHlZA8wfpAtnePaf0ZN0kET5CvqQX4WIQj9ZDgrP1l8QLzDEzMsBzPLQEd41L4kPItOKA3QvBCgs6pSHxaFPWE4CFEHDdJiNOBtV4sGF6DEQPJYnGNEGWGnIAgE6m0giPZ6k8nY3GM6wIBlWWkIR5AA).
> 2. Make `DisposableCollection` aware of `Disposable.NULL` and ignore it when pushed to it.
> 
> Those alternatives both have the benefit of requiring no changes for downstream projects. WDYT?

I'm in favor of 1, though in general I'm not very fond of manual prototype manipulation. But as you say, it would avoid having adopters need to do anything.

paul-marechal:
> So it would probably make sense to guarantee that no disposable is ever put in two disposable collections [...]

I don't see how to enforce that nicely, I have better hope with being careful with our references and not sharing them to multiple dependants that all will try to dispose the same instance. To me it sounds like a "double free" kind of issue. Where is Rust's borrow checker when we need it? :)

Let's go with my first proposed change, I don't like it either but you can add a comment mentioning how each access will return a new disposable instance (the jsdoc should go on the `declare` line).

packages/core/src/common/disposable.ts
```diff
@@ -28,13 +28,19 @@ export namespace Disposable {
...
-        return {
-            dispose: func
-        };
+        return { dispose: func };
     }
-    export const NULL = create(() => { });
+    export declare const NULL: Disposable;
 }
 
+Object.defineProperty(Disposable, 'NULL', {
```
paul-marechal, line 36: I would add a comment for the next person reading this oddity:

```suggestion
// There are issues with sharing the same disposable reference around, so we
// need to make `Disposable.NULL` return new references for each access.
Object.defineProperty(Disposable, 'NULL', {
```

paul-marechal (approved):
LGTM

PR #11055, opened 2022-04-21
fix quick-open: multi-step input (#8777)

### LINKED ISSUE #8777: quick-open: multi-step input does not work well
### Bug Description:
The [PostgreSQL](https://open-vsx.org/extension/ckolkman/vscode-postgres) plugin has a UI issue with the command palette. It wants a series of inputs put into the palette, but pressing "enter" causes the palette to disappear. Thus making it impossible to enter all the fields and connect a database.

### Steps to Reproduce:

1. Install the plugin from OpenVSX.
Install
2. Go to the new Postgres menu and click the little + in the top right corner.
Screenshot_2020-11-24 blackbirdcode app - stage
3. You should see the command palette open with the add database dialogs.
Screenshot_2020-11-24 blackbirdcode app - stage(1)

4. Enter a value for the hostname, e.g. "localhost" and press enter.

**Expected:** The next input appears in the command palette.
**Result:** The command palette disappears without error or confirmation.

### Additional Information

- Operating System: Gitpod
- Theia Version: Not sure how to tell

caseyflynn-google:
#### What it does
Fixes #8777 multiStep quick input flows that call createInputBox 

Previously, when invoking show() a quick input box would be initially created by calling update({visible: true}) in super(), which in turn would invoke $createOrUpdate in quick-open-main and create an InputBox and corresponding session. After the session is created, show() then calls $showInputBox (An api method introduced by Theia, which is not supported by the VS Code API) which creates another InputBox, destroying the previously created session and returning no value to the caller. This leaves an orpaned InputBox on screen.

To correct this, we simply enable show to invoke $createOrUpdate which handles the quick input flow correctly, and set the proper value when received from the quick-open-main.

#### How to test
1. Validate existing multi-step work flows by using the [QuickInput](https://github.com/microsoft/vscode-extension-samples/tree/master/quickinput-sample) [sample](https://github.com/vince-fugnitto/quick-open-samples/releases/download/1.0.0/quickinput-sample-0.0.1.vsix)
2. Validate quick input works in multi-step flows used by [VS Code Solution Explorer](https://open-vsx.org/extension/fernandoescolar/vscode-solution-explorer) using the the [reproduction steps provided](https://github.com/eclipse-theia/theia/issues/8777#issuecomment-1045213722)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/quick-open.ts
+++ b/packages/plugin-ext/src/plugin/quick-open.ts
@@ -451,6 +451,7 @@ export class QuickInputExt implements QuickInput {
     }
 
     _fireChangedValue(changedValue: string): void {
+        this._value = changedValue;
         this._onDidChangeValueEmitter.fire(changedValue);
     }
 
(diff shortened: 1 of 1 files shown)


colin-grant-work (approved):
I confirm that the workflows in the sample extension and the Postgress extension involving multi-step inputs now work as expected, and other plugin quick input use cases are also working well. Always nice to have a solution by subtraction :-).

vince-fugnitto (approved):
LGTM!

PR #11057, opened 2022-04-21
Revert "Use same tempdir for user and system compressed plugin files"

colin-grant-work:
## What it does:

Reverts #10951, which modified the behavior of the `PluginVscodeFileHandler` and `PluginTheiaFileHandler`. It turns out that several different scenarios depended indirectly on the old behavior according to which user plugins were decompressed into the `~/.theia/extensions` folder, so these changes should be reverted, pending a new PR that implements better handling of those scenarios without an implicit dependency on the behavior of the `PluginFileHandlers`.

## How to test
1. Using the plugin view, download a plugin from Open VSX.
2. Observe that it is decompressed into your `<home>/.theia/extensions` directory.
3. Download a compressed plugin from `open-vsx.org` and Install it using the  'Install from VSX' command.
4. Observe that it is decompressed into your `<home>/.theia/extensions` directory.


vince-fugnitto (approved):
LGTM :+1: Extensions that are fetched from open-vsx, and installed by vsix end up under `~/.theia/extensions`.

PR #10846, opened 2022-03-04
Decorate Plugin Trees

colin-grant-work:
#### What it does

Fixes #9642, closes #9643

This PR

 - enables Problem and SCM decoration in plugin `TreeView`s. The code for that is based roughly on #9643.
 - introduces an `AbstractFileTreeDecorator` class with common methods for decorating tree views and refactors the `ScmNavigatorDecorator` and `ProblemDecorator` to inherit from that class.
 - moves special-cased code for the open editors widget decoration from the `ProblemDecorator` into the `NavigatorOpenEditorsWidget` class
 - moves special handling of 'dot decorations' (indicating decorations propagated to a parent from a child) from the navigator into the `TreeWidget` class.

#### How to test

1. Follow the steps outlined in #9643
2. Modify a file in your workspace and save it.
3. Observe that git decoration colors are carried over to the caption in the navigator, and to the caption and the caption suffix in the Open Editors widget.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #11066, opened 2022-04-22
[Decorations]: Emit old and new URI's to ensure ProvideWrapper.data correctly updated

colin-grant-work:
#### What it does

In #10846, I misunderstood the behavior of the [DecorationProviderWrapper](https://github.com/eclipse-theia/theia/blob/08abc2f830708e7a6acfe0e9fbcc8da7293321f9/packages/core/src/browser/decorations-service.ts#L61) class and believed that it would respond correctly to a `DecorationsProvider` emitting only the currently decorated URI's. In fact, it needs to receive a list of both removed and changed URIs. 

#### How to test

1. Open a file.
2. Create some problems.
3. Problem decorations should appear in the file tree.
4. Save the file.
5. Git decorations should appear in the file tree. (assuming you're in a tracked repository)
6. Revert the file to its old state.
7. The git decorations should disappear from the file tree, and if there are no problems left in the file, the problem decorations should disappear from the file tree.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/git/src/browser/git-decoration-provider.ts
+++ b/packages/git/src/browser/git-decoration-provider.ts
@@ -43,8 +43,9 @@ export class GitDecorationProvider implements DecorationsProvider {
         const newDecorations = new Map<string, Decoration>();
         this.collectDecorationData(event.status.changes, newDecorations);
 
+        const uris = new Set([...this.decorations.keys()].concat([...newDecorations.keys()]));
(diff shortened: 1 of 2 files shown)


vince-fugnitto (approved):
I confirm that the changes fix the issues present on master where the decorations do not update :+1:

PR #11074, opened 2022-04-25
Make preference node renderers more robust against `null` values

colin-grant-work:
#### What it does

A number of preferences use `null` as a default even if that is not among their declared values. As a consequence, it is possible for a preference renderer to encounter a null value when another value type is expected.

#### How to test

1. Start the application.
2. Open the preferences UI
3. On `master`, you will see a message about inability to read `toString` of `null` and the layout may fail to be restored.
4. With this code, you will see no message about inability to read `toString` of `null` and the layout should successfully be restored. The preference `html.format.maxPreserveNewLines` should show an empty input box, unless you have set a numerical value for it.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
 
The error on `master` is no longer present, and input fields such as `html.format.maxPreserveNewLines` now properly render by default (`null` value).

PR #11076, opened 2022-04-26
refactor core messaging and RPC (electron-only so far)

paul-marechal:
- Add new reusable components to do communication and RPC in the framework.
- Refactor `electron-main` to `electron-browser` to use those new RPC components.
- Refactor the sample updater from `dev-packages/sample-api` to use the new system.
- Add `electron-main` to `backend` IPC communication.

I open this PR as a draft because **I still need to add much needed documentation and tests**.

I expect from this PR to get feedback on the proposed new APIs and go forward with it. I also hope for https://github.com/eclipse-theia/theia/pull/11011 to align with it.

This PR only covers Electron-specific communication, I left out the `backend` to `frontend` part for now. We can discuss if it should be done in this PR too or not.

Here's a temporary rundown of the proposed new messaging and RPC system, until I document it properly:

---

The goal of the new APIs is to simplify usage by extenders, internal wiring should hold all the complexity but stay reasonable.

`Connection<T>` are the heart of any communication. `T` can be an object, serialization may happen eventually. Connections are often API adapters: Each and every communication library has its own API to send and receive messages but for most relevant use cases only `sendMessage` and `onMessage` are required. You are only required to implement this abstraction if you wish to feed your own details into the other reusable messaging components.

`RpcConnection` are the highest-level abstraction to be able to setup JavaScript proxies. Note that instances _are scoped_ to their respective remote instances, hence why there is no notion of `id` in this API. However you establish a remote connection to a specific instance, or whatever protocol you actually use, the `RpcConnection` is abstracting it.

`ProxyProvider` are named bindings to allow binding a proxy as a regular Inversify service. Example:

```ts
bind(SomeRemoteService)
    .toDynamicValue(ctx => ctx.container.getNamed(ProxyProvider, 'backendOrFrontendOr...').getProxy(remoteServicePath))
    .inSingletonScope();
```

The API is meant to hide how proxies are setup: connection, protocol, etc... You just want a proxy to some remote runtime.

Proxies no longer have a `setClient` method: if you want to react to notifications you can add a listener on methods named like `onSomething...`. If you want a remote object to make requests and get responses, you need a new proxy.

`ServiceContribution` is the new API to expose remote services for proxies to be created. They are either a map of `servicePath` to a callback to get the actual instance to serve over RPC, or directly a callback that will handle the request (`serviceId, serviceParams`).

Inversify bindings might be more verbose now, but this is the best trade-off I was able to find to have small components that make sense. Implementation details are kept to the binding of internals. Theia extenders will most likely never have to write as much wiring code just to use what I documented so far.

Lots of implementation details.

Noteworthy:

- I kept using JSON-RPC as protocol for now, my main concern is redesigning our APIs/abstractions to make JSON-RPC -or any other protocol- a replaceable implementation detail.
- The current JSON-RPC implementation relies on a generic connection multiplexer to serve different remote services over the main connection used for messaging.
- I don't explicitly serialize anything and instead I rely on the low-level transport's implicit serialization. We can serialize ourselves at any point, should we require it.
- Implementation of a simple and generic `ConnectionMultiplexer` that can run over most `Connection<any>`. Supports bi-directional channel opening.
- Inversify only resolves bindings as needed/required. You might see some places where I `get` a binding without doing anything with it, this is because there are side-effects when resolving it (e.g. serve RPC calls).
- We often have an issue with Inversify where some bindings _have_ to be resolved synchronously, but they are inherently constructed asynchronously. This is especially true with some proxy connections that can only be setup once something happens. This is why I rely on `DeferredConnection` in various places: It will buffer messages until the underlying connection is ready, but I can start using it right-away to get a RPC proxy out of it synchronously.
- I added ["tagged strings and symbols"](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#type-aliases) to allow TypeScript to understand our typings when binding using Inversify. See `packages/core/src/common/types.ts#serviceIdentifier`.
- Reacting to the frontend loading and reloading itself in Electron `BrowserWindow`s is a pain...

#### How to test

Run the Electron app, IPC-related features should work:

- You should see a log in the console output with the title of all opened windows.
- You should be able to execute the `Open New Window` command.

#### Review checklist

#### Reminder for reviewers


dev-packages/private-re-exports/src/package-re-exports.ts
```diff
@@ -17,8 +17,7 @@
 import cp = require('child_process');
 import fs = require('fs');
 import path = require('path');
-import { parseModule } from '.';
-import { PackageJson, ReExportJson } from './utility';
```
paul-marechal, line 21: @tsmaeder if by drive-by changes you mean ones like this, this is actually to prevent some circular dependencies that prevented me from building.

packages/core/src/electron-common/electron-main-window-service.ts
```diff
@@ -14,11 +14,12 @@
...
 
-export const electronMainWindowServicePath = '/services/electron-window';
-export const ElectronMainWindowService = Symbol('ElectronMainWindowService');
+export const electronMainWindowServicePath = servicePath<ElectronMainWindowService>('/services/electron-window');
+export const ElectronMainWindowService = serviceIdentifier<ElectronMainWindowService>('ElectronMainWindowService');
 export interface ElectronMainWindowService {
-    openNewWindow(url: string, options?: NewWindowOptions): undefined;
-    openNewDefaultWindow(): void;
+    openNewWindow(url: string, options?: NewWindowOptions): MaybePromise<void>;
+    openNewDefaultWindow(): MaybePromise<void>;
```
paul-marechal, line 24: This is another seemingly drive-by change that allowed me to actually type check the bindings for it.

Without this TypeScript wasn't giving me any type checking...

packages/core/src/common/service-provider.ts
```diff
@@ -0,0 +1,128 @@
...
+ *     .inSingletonScope()
+ *     .whenTargetNamed(YourServiceNamespace);
+ * ```
+ */
+export const ServiceContribution = serviceIdentifier<ServiceContribution>('ServiceContribution');
+export type ServiceContribution = ServiceContributionFunction | ServiceContributionRecord;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ServiceContributionFunction = (serviceId: string, params?: any) => any;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export interface ServiceContributionRecord { [serviceId: string]: (params?: any) => any };
```
paul-marechal, line 92: Service contributions are one of the new layers: When you want to do RPC rather than explicitly binding services to work over one protocol or the other, you can use this contribution point to provide the instance to expose via RPC, and connection handlers may choose any RPC implementation that will fetch the instance to serve from the contributed services.

e.g. Frontend requests for a proxy at `/jsonrpc/some/service/path` via socket.io. Socket.io receives the new connection and routes it to the `/jsonrpc/*serviceId` connection handler which runs JSON-RPC over the incoming connection. This JSON-RPC connection handler then looks up the service for `serviceId` through `ServiceContribution` bindings, if it find an instance it will wire it to answer JSON-RPC requests over the incoming connection. Done.

packages/core/src/common/common-module.ts
```diff
@@ -0,0 +1,59 @@
...
+        .toDynamicValue(ctx => promise => {
+            // eslint-disable-next-line no-null/no-null
+            const nullObject = Object.freeze(Object.create(null));
+            const reflection = ctx.container.get(Reflection);
+            const proxyHandler = new LazyProxyHandler(promise, reflection);
+            return new Proxy(nullObject, proxyHandler);
+        })
+        .inSingletonScope();
+    // #endregion
+});
```
paul-marechal, line 59: This container module contains just contains new re-usable components to facilitate wiring using the new actual APIs. One of the things that doesn't help is that communication between electron-main/electron-backend is quite different from the way it works between frontend/backend...

examples/api-samples/src/electron-browser/updater/sample-updater-frontend-contribution.ts
```diff
@@ -68,18 +67,6 @@ export namespace SampleUpdaterMenu {
...
-export class SampleUpdaterClientImpl implements SampleUpdaterClient {
-
-    protected readonly onReadyToInstallEmitter = new Emitter<void>();
-    readonly onReadyToInstall = this.onReadyToInstallEmitter.event;
-
-    notifyReadyToInstall(): void {
-        this.onReadyToInstallEmitter.fire();
-    }
-
-}
```
paul-marechal, line 81: This might be the most important change with the new RPC/proxy API: we no longer setup "clients" to get notifications back from the remote: we can use `Event` fields directly instead.

examples/api-samples/src/electron-browser/updater/sample-updater-frontend-contribution.ts
```diff
@@ -113,14 +100,11 @@ export class SampleUpdaterFrontendContribution implements CommandContribution, M
...
 
-    @inject(SampleUpdaterClientImpl)
-    protected readonly updaterClient: SampleUpdaterClientImpl;
-
     protected readyToUpdate = false;
 
     @postConstruct()
     protected init(): void {
-        this.updaterClient.onReadyToInstall(async () => {
+        this.updater.onReadyToInstall(async () => {
```
paul-marechal, line 107: Here's an example of directly using `Event` fields directly on the proxy. Less boilerplate code than with clients.

packages/core/src/common/disposable-proxy.ts
```diff
@@ -0,0 +1,103 @@
...
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { Disposable } from './disposable';
+
+export function createDisposableProxy<T extends Disposable>(target: T): T {
+    return new Proxy(target, new DisposableProxyHandler());
+}
+
+export class DisposableProxyHandler<T extends Disposable> implements ProxyHandler<T> {
```
paul-marechal, line 25: This file is not used anymore, but it gives a tool to wrap disposables so that we get warnings (without throwing) whenever we use a disposed instance because that's a bug.

packages/core/src/common/rpc.ts
```diff
@@ -0,0 +1,140 @@
...
+ * remote service it represents.
+ */
+export const RpcConnection = serviceIdentifier<RpcConnection>('RpcConnection');
+export interface RpcConnection {
+    onClose: Event<void>
+    onRequest(handler: (method: string, params: any[], token: CancellationToken) => any): void
+    onNotification(handler: (method: string, params: any[]) => void): void
+    sendRequest<T>(method: string, params: any[]): Promise<T>
+    sendNotification(method: string, params: any[]): void
+}
```
paul-marechal, line 39: In my opinion this is one of the most important new abstractions in this change.

No matter how your RPC protocol works, or how you establish the connection, as long as you respect this interface we can create and serve JavaScript proxies.

Instances that implement this interface should be scoped to a service.

packages/core/src/common/connection.ts
```diff
@@ -0,0 +1,270 @@
...
+    onClose: Event<void>
+    /**
+     * Emitted when something goes wrong with the underlying transport.
+     * Typically when a message fails to be sent or received.
+     */
+    onError: Event<Error>
+    onMessage: Event<T>
+    sendMessage(message: T): void
+    close(): void
+}
```
paul-marechal, line 65: Second most important abstraction: a normalized interface to send/receive messages.

We handle so many APIs to do communications (sockets, node IPC, electron IPC, custom multiplexed channels, etc) and none really has the same API. We use them all in the same way: to send and receive messages.

This type wouldn't be required if it weren't for the other components I created that are used to help with setting up communication stacks.

e.g. I need to do multiplexing over different IPC APIs for Electron but I only want one multiplexer implementation. By using the `Connection` type I can just write adapters for each API and use both in the same `DefaultConnectionMultiplexer` implementation.

Another important thing to keep in mind with connections: They may not only send strings or buffers. Most APIs support automatic serialization, and some of those implementations do a much better job than any other `JSON.stringify` we would call ourselves.

e.g. Socket.io will optimize the transmission of buffers sent through it. By thinking we should just send strings we called `JSON.stringify` on our structures before passing them to Socket.io, preventing it from properly sending our data because we passed it as a raw string.

The idea with connections then is to understand what the underlying transport support sending and stick to that, unless we have more specific needs. You may see more `Connection<any>` than `Connection<string>` in this PR.

paul-marechal:
Note that I am in the process of replacing the frontend/backend API with the new one. I initially thought it would be better to merge only the Electron bits but I don't think so anymore...

I'll mark the PR ready for review once I'm finished with frontend/backend. This should not block https://github.com/eclipse-theia/theia/pull/11011.

In the meantime the high-level abstractions and ideas won't change, so feedback is still welcome on the new approach!

e.g. Named `ProxyProvider`, named `ServiceContribution`, `RpcConnection`, `Connection<T>`.

paul-marechal:
I don't have an ETA on the backend-frontend refactoring part, but I believe that I passed the mid-point.

Mostly fighting with previous code oddities.

packages/core/src/common/rpc.ts
tsmaeder: IMO, we need to think about what we're abstracting here: is it transport (socket, electron, ipc) or is it lifecycle? I think the concepts we should deal with are service objects (proxies and servers) and transports (or channels in the related PR).  Everything in between should be an implementation detail.

examples/api-samples/src/electron-browser/updater/sample-updater-frontend-contribution.ts
tsmaeder: I'm not sure this is actually an improvement. How is this better than the old way of setting up service objects on both ends?

packages/core/src/electron-node/electron-node-backend-module.ts
```diff
@@ -0,0 +1,114 @@
...
+        bind(ProxyProvider)
+            .toDynamicValue(ctx => {
+                const multiplexer = ctx.container.getNamed(ConnectionMultiplexer, ElectronMainAndBackend);
+                const jsonRpcConnectionFactory = ctx.container.get(JsonRpcConnectionFactory);
+                return ctx.container.get(DefaultRpcProxyProvider).initialize(
+                    (serviceId, serviceParams) => jsonRpcConnectionFactory(multiplexer.open({ serviceId, serviceParams }))
+                );
+            })
+            .inSingletonScope()
+            .whenTargetNamed(ElectronMainAndBackend);
```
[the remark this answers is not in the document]
tsmaeder: Maybe I'm not understanding the concepts here yet, but it just seem I need to understand too many interlocking concepts at once here: If I think about what we're trying to achieve, it's not that complicated:

1. The system establishes a number of well know point-to-point connections at startup, for example front end <=> back end or front end <=> plugin host or back end <=> file system change watcher
2. Some clients want to bind service objects whose lifecycle it bound to those connections when the connections are established. 
3. Some clients bind service proxies to those connections. 

When we establish a connection, we establish a service scope for those services. The connection being established is put inside this scope and allow any listeners to react to the connection.
We can bind a rpc protocol factory that will create an rpc protocol instance in this new scope. it will be bound to the connection. 
We need a contribution point that allows clients to bind service object factories and client proxy factories inside this scope. There needs to be an initialization protocol between clients and services that is two-phase: only after full initialization are remote calls allowed.
The services are bound to a named endpoints, for example: "BackEndFrontEnd" or "PluginHost". Note that multiple instances of the same endpoint may be established (such as with plugin hosts).

examples/api-samples/src/electron-browser/updater/sample-updater-frontend-contribution.ts
tsmaeder: I fear this might make initialization of service objects more complicated.

examples/api-samples/src/electron-browser/updater/sample-updater-frontend-contribution.ts
tsmaeder: I generally don't mind boilerplate code if it's simpler to understand than the alternative. Typing is not the problem in software construction.

dev-packages/application-manager/src/generator/backend-generator.ts
```diff
@@ -91,7 +91,7 @@ const serverModule = require('./server');
 const serverAddress = main.start(serverModule());
 
 serverAddress.then(({ port, address }) => {
-    if (process && process.send) {
+    if (process.send) {
```
tsmaeder, line 94: This is what I mean by "drive by fix". Yes, it's a good fix, but unrelated to the problem at hand.

packages/core/src/common/types.ts
```diff
@@ -130,3 +132,25 @@ export namespace ArrayUtils {
...
+
+/**
+ * @internal
+ */
+export type SymbolIdentifier<T> = symbol & interfaces.Abstract<EmptyMeansAny<T>>;
+
+/**
+ * Use this to create "typed symbols" for Inversify typings to understand our bindings.
+ */
+export function serviceIdentifier<T>(name: string): SymbolIdentifier<T> {
```
tsmaeder, line 153: Changing service identifiers from symbols to their own types is one change that could be done in a separate PR to make this one smaller. It affects a lot of files that need to be reviewed.

packages/core/src/common/connection.ts
```diff
@@ -0,0 +1,270 @@
...
+export type AnyConnection = Connection<any>;
+
+/**
+ * A `Connection` allows you to listen for messages and send messages back.
+ *
+ * Messages must arrive integrally and in the order they are sent!
+ *
+ * Most implementations are going to be API adapters.
+ */
+export const Connection = serviceIdentifier<Connection<any>>('Connection');
```
tsmaeder, line 52: Why is there a service identifier here?

packages/core/src/common/connection.ts
```diff
@@ -0,0 +1,270 @@
...
+export interface Connection<T> {
```
tsmaeder, line 53: What's the concrete benefit of connections being typed?

dev-packages/application-manager/src/generator/backend-generator.ts
paul-marechal: Fair enough on this one.

packages/core/src/common/connection.ts
paul-marechal: This is a way to advertise what the transport supports. Most connections are `Connection<any>` but as soon as you see `Connection<string>` or `Connection<Buffer>` you know you will need to serialize your data.

packages/core/src/common/connection.ts
paul-marechal: I used to bind named connections as part of internal wiring, having the identifier defined here helped.

packages/core/src/common/types.ts
paul-marechal: I'll see to extract this once I get back to a functional state with this PR.

Right now I only used `serviceIdentifier` where I happened to work, but I could try to make a PR that replaces everything instead.

packages/core/src/common/rpc.ts
paul-marechal: This is an implementation detail, but _remote_ procedure calls inherently carry the idea that there's some underlying "connection" to a remote. The connection related event `onClose` is only there to propagate to users of this API that the link is no longer valid and further calls won't work anymore. If the link never breaks, `onClose` will never be fired. Note that this API doesn't leak more than that to users, only notify proxies that RPC won't work anymore with the current instance.

examples/api-samples/src/electron-browser/updater/sample-updater-frontend-contribution.ts
paul-marechal: Typing prevents mistakes. This apart, the previous required boilerplate code for the `xxxClient` API was often misused which I assumed was a symptom of a design problem.

The new API allows server to expose their events through proxies, so it almost remove the needs for clients like before. The last feature of client were requests going the other way from the main RPC setup. To do that now you can just create a new dedicated proxy.

Just as an example, the most egregious implementation I've seen using clients must be this one:

https://github.com/eclipse-theia/theia/blob/fd14c0bb423ece9add7825919cc3454875897625/packages/terminal/src/browser/terminal-frontend-contribution.ts#L199-L206

Here `TerminalWatcher` is a client to `TerminalServer`. And instead of having the remote server call a method to fetch collections from its clients and use the result, it does this roundabout procedure where it sends an event and expects the receiving side to initiate subsequent RPC calls (plural) with `this.shellTerminalServer.setCollection(...)`.

In my latest changes I replaced the idea of the client on the server side by a proxy from the frontend implementing storage methods. The server just does `environmentStore.getEnvironmentVariables()` or `environmentStore.saveEnvironmentVariables(vars)`, it is transparently handled by the frontend.

examples/api-samples/src/electron-browser/updater/sample-updater-frontend-contribution.ts
paul-marechal: I don't see how this makes things more complicated. I keep deleting code to use this new API.

dev-packages/private-re-exports/src/package-re-exports.ts
paul-marechal: Those imports are ticking bombs waiting to blow up in someone's face.

As it blew up in mine, I didn't try to untangle the cyclic import error and just nuked the issue.

I'll try to make a PR just for that if that helps.

paul-marechal:
Closing for now, I'll come back to it eventually.

PR #4110, opened 2019-01-18
Replace files.eol preferences with LF and CRLF

uniibu:
This is in relation to https://github.com/theia-ide/theia/issues/4087 Currently, the preferences widget parses the `\n` and `\r\n` which is why only `auto` is available from the choices. 

Primarily, i thought of simply escaping both values (`\\n`, `\\r\\n`) which technically will work and will show up as `\n` and `\r\n` from the choices.  However, choosing any of the two, it gets stored on the settings file as `"files.eol": "\\n"` or `"files.eol": "\\r\\n"` . So I decided to just use LF and CRLF and just converts them to their corresponding values during file creation.

Signed-off-by: Uni Sayo <unibtc@gmail.com>

PR #11079, opened 2022-04-27
fix files.eol preferences cannot display correctly

gaomin:
#### What it does

https://github.com/eclipse-theia/theia/pull/4110
https://community.theia-ide.org/t/preference-files-eol-cannot-display-correctly/2368

#### How to test

click menu: Preferences => Open Settings(UI) => Files Eol => edit & show options

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/browser/views/components/preference-select-input.ts
+++ b/packages/preferences/src/browser/views/components/preference-select-input.ts
@@ -23,6 +23,7 @@ import { Preference } from '../../util/preference-types';
 import { PreferenceLeafNodeRendererContribution } from './preference-node-renderer-creator';
 import * as React from '@theia/core/shared/react';
 import * as ReactDOM from '@theia/core/shared/react-dom';
+import { escapeInvisibleChars } from '@theia/core/lib/common/strings';
 
 @injectable()
 export class PreferenceSelectInputRenderer extends PreferenceLeafNodeRenderer<JSONValue, HTMLDivElement> {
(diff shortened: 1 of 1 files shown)


packages/preferences/src/browser/views/components/preference-select-input.ts
```diff
@@ -38,7 +39,7 @@ export class PreferenceSelectInputRenderer extends PreferenceLeafNodeRenderer<JS
         const values = this.enumValues;
         const defaultValue = this.preferenceNode.preference.data.default;
         for (let i = 0; i < values.length; i++) {
-            const value = `${values[i]}`;
+            const value = escapeInvisibleChars(`${values[i]}`);
             const detail = PreferenceProvider.deepEqual(defaultValue, value) ? 'default' : undefined;
```
msujew, line 43: Suggestion: I believe that this messes with the `deepEqual` check in case `value !== escapeInvisibleChars(value)`. I would recommend the follwing:

```suggestion
            const value = values[i];
            const label = escapeInvisibleChars(`${value}`);
            const detail = PreferenceProvider.deepEqual(defaultValue, value) ? 'default' : undefined;
```

And then use the `label` later at the `items.push` invocation.

msujew (commented):
@gaomin Thanks for your first contribution 👍 

This addresses the issue at hand, but might introduce another issue. See the comment below.

packages/preferences/src/browser/views/components/preference-select-input.ts
gaomin: Thanks for you suggestion, I modified it.

packages/preferences/src/browser/views/components/preference-select-input.ts
```diff
@@ -39,6 +40,7 @@ export class PreferenceSelectInputRenderer extends PreferenceLeafNodeRenderer<JS
         const defaultValue = this.preferenceNode.preference.data.default;
         for (let i = 0; i < values.length; i++) {
             const value = `${values[i]}`;
+            const label = escapeInvisibleChars(`${values[i]}`);
```
msujew, line 43: Suggestion: I believe this might actually be an error in the original implementation (my fault). Using the stringified version in the `deepEqual` check is incorrect (but usually works). It should probably look like this:

```suggestion
            const value = values[i];
            const stringValue = `${value}`;
            const label = escapeInvisibleChars(stringValue);
```

And later use `value: stringValue` instead. Do you mind making this change as well?

gaomin: I got it, modified.

msujew (approved):
Thanks, looks good to me 👍

PR #11082, opened 2022-04-27
Theia launched from Theia hangs on Windows [Electron] #11081

### LINKED ISSUE #11081: Theia launched from Theia hangs on Windows [Electron]
### Bug Description:
See discussion: https://github.com/eclipse-theia/theia/discussions/11052

Launching a Theia Electron App from the terminal of a Theia Electron App on windows does not work. 

### Steps to Reproduce:

1. Launch the Theia Electron Example on windows
2. Inside the running example, open a workspace with a second clone of the Theia sources, build examples, and launch the Theia Electron Example
3. The second electron application will hang on the preload animation

### Additional Information

- Operating System: Windows
- Theia Version: Master

jfaltermeier:
#### What it does
* Check existing environment for variants of the ElectronSecurityToken key on windows

fixes #11081

#### How to test
* Launch the Theia Electron Example on windows
* Inside the running example, open a workspace with a second clone of the Theia sources, build examples, and launch the Theia Electron Example
* The second electron application should start

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-common/electron-token.ts
+++ b/packages/core/src/electron-common/electron-token.ts
@@ -15,12 +15,13 @@
 // *****************************************************************************
 
 /**
- * This token is unique the the current running instance. It is used by the backend
+ * This token is unique to the current running instance. It is used by the backend
  * to make sure it is an electron browser window that is connecting to its services.
  *
- * The identifier is a string, which makes it usable as a key for cookies or similar.
+ * The identifier is a string, which makes it usable as a key for cookies, environments, etc.
  */
-export const ElectronSecurityToken = 'x-theia-electron-token';
+// Note that it needs to be uppercase for it to work properly in Windows environments.
+export const ElectronSecurityToken: string = 'THEIA_ELECTRON_TOKEN';
 export interface ElectronSecurityToken {
     value: string;
 };


packages/core/src/electron-main/electron-main-application.ts
```diff
@@ -479,14 +479,24 @@ export class ElectronMainApplication {
     }
 
     protected async getForkOptions(): Promise<ForkOptions> {
+        const env: NodeJS.ProcessEnv = {
+            ...process.env,
+            [ElectronSecurityToken]: JSON.stringify(this.electronSecurityToken),
+        };
+
+        // when the ElectronSecurityToken was present on the parent process already, we need to check for variants of the key
+        if (process.platform === 'win32') {
```
colin-grant-work, line 488: Is this a transformation that Windows does automatically - it translates all env variables to upper case?

paul-marechal: I just tried quickly and if I set a env var in lowercase I still see it lowercase in `process.env`.

The weird behaviour on Windows is that `process.env.TEST` is the same as `process.env.test`:

```batch
> set test=1 && node -p "'TEST' in process.env && 'test' in process.env"
true
```

paul-marechal (commented):
It seems that setting the key to uppercase fixes the issue on Windows, according to your change.

Ultimately it would be nice if we could understand why using uppercase works and lowercase doesn't, but let's just use uppercase.

https://github.com/eclipse-theia/theia/blob/c6ac473868c08557a8f343e35ea1bc9604fd91f3/packages/core/src/electron-common/electron-token.ts#L23

To:

```ts

/**
 * This token is unique the the current running instance. It is used by the backend
 * to make sure it is an electron browser window that is connecting to its services.
 *
 * The identifier is a string, which makes it usable as a key for cookies, environments, etc.
 */
// Note that it needs to be uppercase for it to work properly in Windows environments.
export const ElectronSecurityToken: string = 'THEIA_ELECTRON_TOKEN';
```

(I put the note as an inline comment to not pollute the public jsdoc with details)

jfaltermeier:
Thanks, I will change it to upper case and update the PR. The information I found was this: https://nodejs.org/api/child_process.html

---

On Windows, environment variables are case-insensitive. Node.js lexicographically sorts the env keys and uses the first one that case-insensitively matches. Only first (in lexicographic order) entry will be passed to the subprocess. This might lead to issues on Windows when passing objects to the env option that have multiple variants of the same key, such as PATH and Path.

---
I am not sure if there is a guarantee that windows always upper cases the keys, however nodejs will chose the upper-cased one in case there are multiple ones. So this should work.

paul-marechal (approved):

PR #11090, opened 2022-04-29
Introduce DebugSession#workspaceFolder

### LINKED ISSUE #10023: Introduce `DebugSession#workspaceFolder`
### Feature Description:
Introduce `DebugSession#workspaceFolder` 
See https://code.visualstudio.com/api/references/vscode-api#DebugSession

The workspace folder should be taken into account when running pre-launch/debug tasks, for example, but also in labels where it makes sense. Digging through the source code VS Code might be necessary to identify usages.

jfaltermeier:
#### What it does
Fixes #10023

* Adds the missing `workspaceFolderUri` to the `DebugSession` API
* Adjust code to create the session including `workspaceFolderUri`
* The existing `DebugSessionManager` was already honoring the workspace folder as far as I could see (-> e.g. when executing pre launch tasks: https://github.com/eclipse-theia/theia/blob/7f5047fcb21c1415d4ecf9b5ba34f802615e1c05/packages/debug/src/browser/debug-session-manager.ts#L211) so only the session object creation itself was adjusted
* Adds the workspacefolder to the DebugSession's UI label, in case we have a multi-root workspace (and multiple running debug sessions, as otherwise the ui node is hidden)

_VS-Code UI (esm-i was the set workspace-folder. This is not shown in the Top, because this is a launch without a set ws folder):_
ksnip_20220426-092056

_Theia UI (workspace folder was called java):_
ksnip_20220503-101753

Signed-off-by: Johannes Faltermeier <jfaltermeier@eclipsesource.com>

#### How to test
I've created a small VSCode extension that tracks the creation of debug sessions and logs the workspace folder of a session:
https://github.com/jfaltermeier/vscode-debug-playground
https://github.com/jfaltermeier/vscode-debug-playground/blob/main/src/extension.ts
You may get this test extension from https://github.com/jfaltermeier/vscode-debug-playground/releases/tag/0.0.1

* start Theia Browser Example with above extension
* debug any application (I was testing with a java application)

This should create a log message similar to this one:

```
root INFO [hosted-plugin: 67589] ES : DebugAdapterTracker: DebugSession created with workspaceFolder: file:///home/johannes/git/java-sample
```

Without the changes, the worspaceFolder was undefined all the time
```
root INFO [hosted-plugin: 70803] ES : DebugAdapterTracker: DebugSession created with workspaceFolder: undef
```

* For testing the UI label, create a multi root workspace and start multiple debug sessions that stop at a breakpoint. 
* The session label will show the wsfolder in brackets

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-session-manager.ts
+++ b/packages/debug/src/browser/debug-session-manager.ts
@@ -214,7 +214,7 @@ export class DebugSessionManager {
                     }
                 }
 
-                const sessionId = await this.debug.createDebugSession(resolved.configuration);
+                const sessionId = await this.debug.createDebugSession(resolved.configuration, options.workspaceFolderUri);
                 return this.doStart(sessionId, resolved);
             } catch (e) {
                 if (DebugError.NotFound.is(e)) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - [Previous Changelogs](https://github.com/eclipse-theia/theia/tree/master/doc/changelogs/)
 
+## v.1.26.0
+
+- [plugin] Introduce `DebugSession#workspaceFolder` [#11090](https://github.com/eclipse-theia/theia/pull/11090) - Contributed on behalf of STMicroelectronics
+
 ## v1.25.0 - 4/28/2022
 
 [1.25.0 Milestone](https://github.com/eclipse-theia/theia/milestone/35)
--- a/packages/debug/src/browser/debug-session-contribution.ts
+++ b/packages/debug/src/browser/debug-session-contribution.ts
@@ -30,6 +30,7 @@ import { Channel, DebugAdapterPath } from '../common/debug-service';
 import { ContributionProvider } from '@theia/core/lib/common/contribution-provider';
 import { FileService } from '@theia/filesystem/lib/browser/file-service';
 import { DebugContribution } from './debug-contribution';
+import { WorkspaceService } from '@theia/workspace/lib/browser';
 
 /**
  * DebugSessionContribution symbol for DI.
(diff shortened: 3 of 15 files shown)


packages/debug/src/browser/debug-session-manager.ts
```diff
@@ -214,7 +214,7 @@ export class DebugSessionManager {
                     }
                 }
 
-                const sessionId = await this.debug.createDebugSession(resolved.configuration);
+                const sessionId = await this.debug.createDebugSession(resolved.configuration, options.workspaceFolderUri);
```
colin-grant-work, line 217: It does not appear that the implementation of the `createDebugSession` method on the `DebugServiceImpl` has been updated to handle the `workspaceFolderUri`. Is that intentional?

colin-grant-work, line 217: It does not appear that the implementation of the `createDebugSession` method on the `DebugServiceImpl` has been updated to handle the `workspaceFolderUri`. Is that intentional?

jfaltermeier: Thank you. No, this was an oversight, since this value is not needed there.
In the plain Theia API the ws folder is taken from the `DebugSessionOptions` and we only need it for vs code compatibility.
I've added the parameter as `_workspaceFolderUri?`

colin-grant-work (commented):
My fault, but in my suggested snippet, I omitted a space between `this.configuration.name` and `(${suffixes.join(' - ')})`. Apart from that minor UI issue, this looks good to me, so if you fix that, I'll approve.

jfaltermeier:
I've added the space between configuration name and the joined suffixes and rebased on current master. 
I haven't changed the logic for checking whether we have a multiroot-workspace. I also think that it is fine to display the suffix in a multi-root environment with just one root.

PR #11091, opened 2022-04-29
Upgrade playwright to 1.21.1

planger:
#### What it does
Upgrade playwright dependency to `1.21.1`.

#### How to test
`yarn test:playwright` should still run without failures locally and in the CI.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/playwright/package.json
+++ b/examples/playwright/package.json
@@ -29,7 +29,7 @@
     "src"
   ],
   "dependencies": {
-    "@playwright/test": "1.17.1",
+    "@playwright/test": "1.21.1",
     "fs-extra": "^9.0.8"
   },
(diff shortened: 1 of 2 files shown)


vince-fugnitto (commented):
@planger do you still plan on upgrading playwright to `1.21.1`?

planger:
@vince-fugnitto Thanks for asking! This change was mostly to test whether the upgrade works also in the CI and not only on my machine.

I don't think there is a particular feature that we'd like to make use of in the new playwright releases. So from our point of view it doesn't justify the overhead with updating the used licenses etc.

So unless anyone in the community would like to see this update, I'd hold off for a few more month and do an upgrade then.

Issue #10778, opened 2022-02-18
viewsContainers should be hidden when empty


thegecko:
### Bug Description:

To mimic functionality in VS Code, any viewsContainer which ends up being empty (e.g. all views it contains are hidden with when clauses) should itself be hidden.

In Theia, if all child views are hidden, the parent viewsContainer remains visible

### Steps to Reproduce:

1. Load a VS Code extension which provides views in a custom viewsContainer
2. Hide all views in this viewsContainer using `when` clauses
3. Expect the viewsContainer to be hidden, but it remains visible

### Additional Information

This can be observed when running the `Theia IDE` example in this repo:

https://github.com/thegecko/vscode-theia-extension

- Operating System: Browser
- Theia Version: v1.22

PR #11095, opened 2022-05-02
Prevent MonacoEditorService from instantiating ContextKeyService thro…

colin-grant-work:
…ugh Monaco

#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/11096.

If we wish to override Monaco services, we must do so before the same services are instantiated by Monaco by other means. At the moment, one of our desired overrides, the `MonacoEditorService` is instantiating a `ContextKeyService` using Monaco methods, but we want to override the `ContextKeyService` as well. This PR changes the way the `MonacoEditorService` accesses the `ContextKeyService` to allow the override to take effect.

#### How to test

1. Install [this plugin](https://github.com/vince-fugnitto/enablement-test/releases/tag/0.0.1)
2. Use the command palette to run the `Enablement: Enable` and `Enablement: Disable` commands.
3. Observe that the `Show Enablement Issue` widget appears and disappears from its view container. (But due to https://github.com/eclipse-theia/theia/issues/10778 , the view container will not close)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "all": "yarn -s install && yarn -s lint && yarn -s build",
     "browser": "yarn -s --cwd examples/browser",
     "build": "yarn -s compile && yarn -s build:examples",
+    "build-p": "yarn -s compile && lerna run --scope=\"@theia/example-*\" bundle --parallel",
     "build:examples": "yarn -s download:plugins && lerna run --scope=\"@theia/example-*\" bundle --parallel",
     "clean": "yarn -s rebuild:clean && yarn -s lint:clean && node scripts/run-reverse-topo.js yarn -s clean",
     "compile": "echo Compiling TypeScript sources... && yarn -s compile:clean && yarn -s compile:tsc",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -54,6 +54,7 @@
     "all": "yarn -s install && yarn -s lint && yarn -s build",
     "browser": "yarn -s --cwd examples/browser",
     "build": "yarn -s compile && yarn -s build:examples",
+    "build-p": "yarn -s compile && lerna run --scope=\"@theia/example-*\" bundle --parallel",
```
vince-fugnitto, line 57: Do we mean to include this script as part of the change?

vince-fugnitto (approved):
I confirm that the changes work as expected, we can now handle `setContext` calls properly from plugins in order to support [custom clause contexts](https://code.visualstudio.com/api/extension-guides/command#using-a-custom-when-clause-context). I also confirmed that existing when clauses work as expected causing no regressions.

PR #8776, opened 2020-11-24
8769-user-keybinding: propagating Theia overriden keybindings into VS Code editor

danarad05:
Signed-off-by: Dan Arad <dan.arad@sap.com>

#### What it does
Fixes #8769  
VS Code/monaco editor uses default vscode keybindings if no dynamicKeybindings are set. Theia keybindings overriden by user were not propagated into VS Code because they weren't added to dynamicKeybindings array using addDynamicKeybinding method.

On VS Code' side,  **_getResolver** is responsible for merging default keybindings with overriden ones ( = dynamicKeybindings)
image

#### How to test
After PR [8589](https://github.com/eclipse-theia/theia/pull/8589) is merged:
a. Update Keyboard shortcuts for "acceptRenameInput" & "acceptRenameInputWithPreview":
image
b. click F2 to rename a function - rename input field is displayed with updated keyboard settings
image

#### Review checklist

- [X] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #11102, opened 2022-05-03
Update keybinding widget when keybindings change

colin-grant-work:
#### What it does

In the review of #8776, [it was noticed](https://github.com/eclipse-theia/theia/pull/8776#pullrequestreview-795287467) that the keybinding widget does not update in response to changes to keybindings, so that at startup, it can get stuck displaying default values before user or workspace values are processed. 

#### How to test

1. Modify a keybinding.
2. With the keybinding widget open, restart the application.
3. Find the modified keybinding.
4. Observe that modified keybinding is shown rather than the default.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
I confirmed that the changes work as expected :+1: 

- [x] updated keybindings are properly reflected on reload
- [x] the `query` is restored on reload, the view correctly restores the state and re-searches

PR #10525, opened 2021-12-09
Update Themes

colin-grant-work:
#### What it does

Updates the specification of our default themes to match recent VSCode.

#### How to test

1. Open a TS file.
2. Move through the built-in themes.
3. Observe that they behave as expected.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

CQ: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23884

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #11111, opened 2022-05-06
repo: give proper `id` for builtin themes

vince-fugnitto:
#### What it does

The pull-request updates the root `package.json` to use the proper `id` for builtin themes so they are not re-downloaded by the `extension-pack`.

#### How to test

1. perform `yarn`
2. perform `yarn download:plugins`
3. confirm that themes are not installed twice under `plugins/` and you should see the following logs for each plugin when attempting to resolve the `extension-pack`:

    ```
    - vscode.theme-monokai: already downloaded - skipping
    ````

#### Review checklist

#### Reminder for reviewers


colin-grant-work (approved):
The changes are reasonable and prevent duplicate downloads of theme extensions included in the `builtins` extension pack. 👍

ccorn:
Note that there is also the case of `ms-vscode.references-view` in the [pack](https://open-vsx.org/api/eclipse-theia/builtin-extension-pack/1.50.1) being labelled `vscode-references-view` in `theiaPlugins`.

My recommendation: If you go that way, simply use it for *all* entries of `theiaPlugins`. Exceptions just invite trouble.

Totally off-topic: What about `vscode.theme-solarized-light`? The pack includes it, the version-upgrading theme entries in `package.json` do not...

vince-fugnitto:
> Totally off-topic: What about vscode.theme-solarized-light? The pack includes it, the version-upgrading theme entries in package.json do not...

@ccorn we can add it as well, but is there a need to use the newer version of the theme in the repo?

ccorn:
> we can add it as well, but is there a need to use the newer version of the theme in the repo?

I have no interests there, and no issues to report, just trying get a consistent package.

The reason for all those theme upgrades might be given in #10525 which seems to have been part of an effort on semantic colorization. So, if `vscode.theme-solarized-light` needs custom code to support that effort, this is not a simple matter anyway. Though upgrading that theme would probably still be OK, I guess. (But that is really just a guess.)

PR #11112, opened 2022-05-06
Add --parallel/--no-parallel option to download:plugins

### LINKED ISSUE #1110: [extension-manager] There are a couple of broken test cases when running tests locally
See the end of the console output after running `yarn run test`:

```
@theia/extension-manager:     5) list with search
@theia/extension-manager: Stopped watching. 
@theia/extension-manager:   4 passing (2m)
@theia/extension-manager:   5 failing
@theia/extension-manager:   1) node-extension-server install:
@theia/extension-manager:      Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
@theia/extension-manager:   
@theia/extension-manager:   2) node-extension-server outdated:
@theia/extension-manager:      Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
@theia/extension-manager:   
@theia/extension-manager:   3) node-extension-server update:
@theia/extension-manager:      Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
@theia/extension-manager:   
@theia/extension-manager:   4) node-extension-server list:
@theia/extension-manager:      Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
@theia/extension-manager:   
@theia/extension-manager:   5) node-extension-server list with search:
@theia/extension-manager:      Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
@theia/extension-manager:   
@theia/extension-manager: npm ERR! code ELIFECYCLE
@theia/extension-manager: npm ERR! errno 5
@theia/extension-manager: npm ERR! @theia/extension-manager@0.3.2 test: `theiaext test`
@theia/extension-manager: npm ERR! Exit status 5
@theia/extension-manager: npm ERR! 
@theia/extension-manager: npm ERR! Failed at the @theia/extension-manager@0.3.2 test script.
@theia/extension-manager: npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
@theia/extension-manager: npm ERR! A complete log of this run can be found in:
@theia/extension-manager: npm ERR!     /Users/akos.kitta/.npm/_logs/2018-01-23T13_19_34_711Z-debug.log
(node:38771) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 6): Error: Command failed: npm run test
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @theia/application-package@0.3.2 test: `theiaext test`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @theia/application-package@0.3.2 test script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/akos.kitta/.npm/_logs/2018-01-23T13_17_48_376Z-debug.log


> @theia/application-package@0.3.2 test /Users/akos.kitta/Desktop/theia/dev-packages/application-package
> theiaext test



  extension-package
    isOutdated
      ✓ @theia/core@next (3179ms)
      ✓ @theia/core@latest (1250ms)
      1) @theia/core@0.1.0


  2 passing (14s)
  1 failing

  1) extension-package isOutdated @theia/core@0.1.0:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

Running the test again, I got different test failures:
```
lerna ERR! 
lerna ERR! 
lerna ERR!   8 passing (1m)
lerna ERR!   1 failing
lerna ERR! 
lerna ERR!   1) node-extension-server list with search:
lerna ERR!      Error: Timeout of 30000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
lerna ERR!   
lerna ERR! 
lerna ERR! 

```

xoriath:
The default behaviour is unchanged and is still --parallel,
however this change adds --no-parallel or --parallel false
as an option to not download every listed plugin in parallel.

We are seeing some issues both with ECONNRESET and ZlibBuffer
intermittently (probably due to network errors/throttling)

This fixes #1110

Signed-off-by: Morten Engelhardt Olsen <MortenEngelhardt.Olsen@microchip.com>

#### What it does
Add new argument --parallel (default: true) which will either download plugins in serial or in parallel

#### How to test
There is no real user visible changes, the only difference is how quickly the target folder is created and the plugins extracted (parallel vs serial)

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (commented):

xoriath:
@vince-fugnitto I could have sworn... gimme a moment :)

xoriath:
hmmmm, I'm failing spectacularly at getting the ECA to validate...

vince-fugnitto:
> hmmmm, I'm failing spectacularly at getting the ECA to validate...

I do see that there is a valid eca account for your email 

image

And your authorship and sign-off match:

```
From f8fd03abbc2a953eb7919a3524bcdbb6058fb15a Mon Sep 17 00:00:00 2001
From: Morten Engelhardt Olsen <MortenEngelhardt.Olsen@microchip.com>
Date: Fri, 6 May 2022 08:36:39 -0700
Subject: [PATCH] Add --parallel/--no-parallel option to download:plugins

The default behaviour is unchanged and is still --parallel,
however this change adds --no-parallel or --parallel false
as an option to not download every listed plugin in parallel.

We are seeing some issues both with ECONNRESET and ZlibBuffer
intermittently (probably due to network errors/throttling)

This fixes #11110

Signed-off-by: Morten Engelhardt Olsen <MortenEngelhardt.Olsen@microchip.com>
```

So perhaps it's just the tool taking time to update.

xoriath:
I sent an email to licensing@eclipse.com.. we'll see :)

dev-packages/cli/src/download-plugins.ts
```diff
@@ -103,7 +109,12 @@ export default async function downloadPlugins(options: DownloadPluginsOptions =
             if (typeof pluginUrl !== 'string') {
                 continue;
             }
-            downloads.push(downloadPluginAsync(failures, plugin, pluginUrl, pluginsDir, packed));
+            const download = downloadPluginAsync(failures, plugin, pluginUrl, pluginsDir, packed);
+            if (parallel) {
+                downloads.push(download);
+            } else {
+                await download;
+            }
```
colin-grant-work, line 117: This rewrite is a bit awkward in a context where we're still always declaring `downloads` and always `await`ing `Promise.all(downloads)`. I think it would be preferable to rewrite the implementation at a high level to respect the distinction between parallel and non-parallel downloads.

dev-packages/cli/src/theia.ts
```diff
@@ -223,10 +227,16 @@ async function theiaCli(): Promise<void> {
...
+                'parallel': {
+                    describe: 'Download in parallel',
+                    boolean: true,
+                    default: true
                 }
             },
-            handler: async ({ packed }) => {
-                await downloadPlugins({ packed });
+            handler: async ({ packed, ignoreErrors, apiVersion, apiUrl, parallel }) => {
+                console.warn({ parallel });
```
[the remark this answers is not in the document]
xoriath: ... that was not supposed to be there :)

dev-packages/cli/src/download-plugins.ts
xoriath: I was wondering if we really want to do a full change just for this, as Promise.all([]) does return a already resolved promise so this is ... correct but looks a bit weird... 

I sort of don't like diverging a parallel vs serial download paths too much, as serial is just a special case of parallel :)

Do you have a better vision to make this code less clunky?

colin-grant-work: My idea would be something like this:
```typescript
        const downloadPlugin = ([plugin, url]: [string, string]) => downloadPluginAsync(failures, plugin, url, pluginsDir, packed);
        const toDownload = Object.entries(pck.theiaPlugins).filter((entry: [string, unknown]): entry is [string, string] => typeof entry[1] === 'string');
        if (parallel) {
            await Promise.all(toDownload.map(downloadPlugin));
        } else {
            for (const entry of toDownload) {
                await downloadPlugin(entry);
            }
        }
```

However you'd like to do it, it will need to be done also for extension packs, since the code that handles those currently also does a parallel `Promise.all`.

xoriath:
ECA check fixed...

dev-packages/cli/src/download-plugins.ts
xoriath: Yeah, I see the extension pack path now... got confused by the nested await but you are correct...

ccorn: I am not fluent with typescript , but given the need for the downloads to actually succeed, I would unconditionally chunk the downloads array into bins of size (say) 5 (or 1 if --no-parallel) and await one chunk after the other.

xoriath: ... let's start with a simple yes/no and see how it turns out :) If you want to add chunking into it then feel free to try

colin-grant-work: I'm not sure I see the benefit to this approach, since I don't think it's desirable (in general) to bail early. If I'm the developer, I'd like to get as many plugins as I can successfully get, and then address any failures. If we bail early, it's possible we'll discover additional failures the next time around that we could have tried to address after the first pass. That isn't exactly what the current code on `master` does, since it does stop at the first failure, but it will have initiated and actually execute other requests as well, so it comes close to what I think is the desired behavior.

ccorn: What I mean is: with chunking, the only effect of `--no-parallel` is on the chunk size. No other branches necessary.

colin-grant-work: It's true no branching would be necessary, but all of the current code would have to be rewritten, since chunking requires serialization of a (variable number of) parallelized operations. If you have a concrete use case for chunks of sizes other than 1 (`--no-parallel` in this PR) or infinity (`--parallel` in this PR), please open an issue or contribute a PR to implement it.

xoriath: Rewrote the download parts into a couple of helper functions, and an interface type to carry the necessary plugin fields in a bit more structured form. 

This also makes the parallel/no-parallel decision in one function, if someone wants to add chunking later :)

dev-packages/cli/src/download-plugins.ts
```diff
@@ -114,13 +135,17 @@ export default async function downloadPlugins(options: DownloadPluginsOptions =
             const client = new OVSXClient({ apiVersion, apiUrl });
             // De-duplicate extension ids to only download each once:
             const ids = new Set<string>(Array.from(extensionPacks.values()).flat());
-            await Promise.all(Array.from(ids, async id => {
+            const extensionPacksToDownload = await Promise.all(Array.from(ids, async id => {
                 const extension = await client.getLatestCompatibleExtensionVersion(id);
+                const version = extension?.version;
```
colin-grant-work, line 140: It looks like this code is still making some network calls in parallel to determine the latest compatible version of the given plugin. Would you like to deparallelize that, as well?

xoriath: I have not seen any issues with the ovsx client api requests... so I dont have any reason to run those in serial...

ccorn:
For testing, I have locally merged #11113 and this branch at 547e2d80edbddf6f70cb7e990c6019f6a42124cd.

My test command:

    rm -rf plugins; yarn download:plugins --no-parallel

Improvement: This no longer silently ignores extensions for which no compatible version can be found, but throws:

```
--- resolving 1 extension-packs ---
Error: No download url for extension pack ms-vscode.js-debug-companion (undefined)
    at /home/ccorn/BUILD/theia/dev-packages/cli/lib/download-plugins.js:83:27
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Promise.all (index 72)
    at async downloadPlugins (/home/ccorn/BUILD/theia/dev-packages/cli/lib/download-plugins.js:75:46)
    at async Object.handler (/home/ccorn/BUILD/theia/dev-packages/cli/lib/theia.js:206:13)
```

Which is actually true: The [earliest version](https://open-vsx.org/api/ms-vscode/js-debug-companion/1.0.13) has `engine.vscode == "^1.55.0"`, which is later than `VSCODE_DEFAULT_API_VERSION = 1.53.2`.

Therefore I have added `ms-vscode.js-debug-companion` to `theiaPluginsExcludeIds`.

The bad news: I still get lots of parallel connections and `ECONNRESET`:

```
--- resolving 1 extension-packs ---
Error: socket hang up
    at connResetException (internal/errors.js:609:14)
    at TLSSocket.socketOnEnd (_http_client.js:458:23)
    at TLSSocket.emit (events.js:326:22)
    at endReadableNT (_stream_readable.js:1241:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21) {
  code: 'ECONNRESET'
}
```

Network traffic continues long after that, and even after network traffic has ended,
the command does not terminate, with `lsof -i` listing between 3 and 7 remaining open `node` connections. This behavior seems to be the same as of the v1.25.0 release.

Worse: After manually aborting, the `plugins` directory only contains the extensions listed in `theiaPlugins`, nothing more. Where have the pack-bundled extensions gone? I do not recall similar behavior of the v1.25.0 release.

My workaround for the v1.25.0 release was to replace the extension pack with an explicit list in `theiaPlugins`, as in previous releases.

So, no success yet, but improvement in error checking.
The problem (still) is in resolving the extension packs.

vince-fugnitto:
@ccorn I'm not sure I follow, the behavior should be the following:

1. download all plugins listed under `theiaPlugins`.
2. if an extension is an `extensionPack`, then resolve the extensions it declares:
    - the resolvement should download a version of the extension (listed by id) that satisfies the framework's declared supported api.
    - the resolvement of an extension should skip an extension if already downloaded from step 1.
    - the resolvement should not download an extension that is part of the exclusion list.

If you were to abort before the resolvement of the `extensionPack` then of course you would not see the extension under your folder.

> My workaround for the v1.25.0 release was to replace the extension pack with an explicit list in theiaPlugins, as in previous releases.

As I mentioned previously, the `theiaPlugins` we define in this repo are for our purposes only. There is nothing different in the `1.25.0` release, the `download:plugins` script has not been updated for over [2 months now](https://github.com/eclipse-theia/theia/tree/master/dev-packages/cli/src). This repo simply updated our own builtins for test purposes to make use of an `extensionPack` instead of listing each individually, but it does not affect you downstream.

ccorn:
> If you were to abort before the resolvement of the extensionPack then of course you would not see the extension under your folder

Just tried it again, with #11113 but without this merge request. This time it's a `ZBufError`:

```
--- resolving 1 extension-packs ---
Error: ZBufError
    at Gunzip.<anonymous> (/home/ccorn/BUILD/theia/node_modules/bent/src/nodejs.js:36:57)
    at Gunzip.emit (events.js:314:20)
    at Zlib.zlibOnError [as onerror] (zlib.js:185:8)
```

And I do have to abort as well, but without #11112 at least *some* of the bundled plugins show up after the abort. With 547e2d80edbddf6f70cb7e990c6019f6a42124cd, *none* show up.

> This repo simply updated our own builtins for test purposes to make use of an `extensionPack` instead of listing each individually

That is the point: Resolving extension packs uses a new code path and/or a new scale of connection parallelism which turns out to be problematic. Well, even the usual downloading of `theiaPlugins` sometimes failed, but resolving of extension packs has *never* succeeded here, which is why I have reverted to an all-explicit plugin list, as in v1.24.0 (but with ID-based names where the extension pack would have created those).

ccorn:
More experimenting: I have listed all bundled plugins explicitly in `theiaPlugins`, *but kept the extension pack listed as well*. Thus, resolving the extension pack would be attempted, but should find all extensions downloaded already. Would the script succeed in this case?

**No.** Both without and with this merge request (547e2d80edbddf6f70cb7e990c6019f6a42124cd), the steps logged as `--- downloading plugins ---` and `--- collecting extension-packs ---` went well (and the `plugins` directory populated accordingly); then in `--- resolving 1 extension-packs ---` I still got a `ZBufError` or `ECONNRESET` as reported before. Again, I noticed lots of network traffic, and after that traffic had died down, I still had to abort the script.

Without 547e2d80edbddf6f70cb7e990c6019f6a42124cd, many (but not all) messages like
`- [extension ID]: already downloaded - skipping` were output.
With 547e2d80edbddf6f70cb7e990c6019f6a42124cd, no such messages appeared.

dev-packages/cli/src/download-plugins.ts
ccorn: By now, I suspect that the problematic stage is indeed where compatible versions are determined. The `includeAllVersions` flag probably leads to a significant response size, multiplied by about 80 requests in parallel.

ccorn: Interesting fact: The API response sends an `extensions` array whose entries each have an `allVersions` map. That is, response size grows asymptotically with the square of the number of versions.

ccorn: The median version count currently seems to be at 457. The query for `vscode.bat` comes with 487 versions, and the JSON response is 20,377,701 bytes. With line separators (as prettified by `jq .`), 236680 out of 258585 lines contain version-like crossref URLs (characterized by containing exactly 6 slashes). That's 92% of lines and 96% by byte count after prettifying.

ccorn: This approach does not scale. I'd suggest omitting the `allVersions` maps from the `extensions` entries if `includeAllVersions` is true. But that's an issue for Open-VSX. Here I suppose we should simply resort to query versions one extension at a time (with `--no-parallel` at least).

ccorn: Actually, since the `extensions` array lists most recent versions first, and the logic of `getLatestCompatibleExtensionVersion` is to pick the first compatible one, and if you have, and use, a streamable JSON decoder, one optimization would be to cancel the version query download once a suitable `extensions` array entry has been found...

ccorn: Trying to get the total byte count of the query responses for the bundled extensions. Might take an hour or so.

vince-fugnitto: @ccorn I believe that the recent comments are adding unecessary noise to the pull-request for the author, if you want to investigate further optimizations you can comment instead on the issue. I think the optimizations you recently commented on are likely out-of-scope for the pull-request at this time. Implement parallel downloads should be the full scope, modifications to how we handle the `allVersions` would be out-of-scope. I don't think we should perform multiple open-vsx calls per extension to determine the compatible version.

xoriath:
So, I don't have a problem by just blindly doing the talking to open-vsx in serial... let me add a commit you can try for that...

ccorn:
> So, I don't have a problem by just blindly doing the talking to open-vsx in serial... let me add a commit you can try for that...

Thanks, I will try it! But be prepared for feedback delays measured in hours, in case your changes succeed in making the `includeAllVersion` queries run to completion here.

ccorn:
OK, I locally merged #11112 (8b22b76c6fa57b920965cfa8e56bf535b610399c) with #11113 (51a7899173201132cb1768048ac55557759628ce), and `time yarn download:plugins --no-parallel` succeeded, running to completion without errors.

Not only did it succeed, but it did so within under 30 minutes.
This is remarkable because when I (earlier) used a sequence of `curl` commands to simulate the API requests with `includeAllVersions` and store the responses, that took 186 minutes (yes, 3h) for 73 non-excluded plugins with a total response size of 1,323,992,876 bytes.
(I now suspect that Theia's code uses compression for transfers, which in this case is ideally suited because of the high redundancy in the version info.)

**I solemnly declare that this fix works for me.**
You probably want to polish this, so I'll keep watching.

(Of course, the gross amount of data transferred for version-checking is still good material for a precious rant, but not within the scope of this pull request.)

colin-grant-work:
> (Of course, the gross amount of data transferred for version-checking is still good material for a precious rant, but not within the scope of this pull request.)

I'd encourage taking the rant to the [Open VSX](https://github.com/eclipse/openvsx) repository, since it seems that they have structured their responses to include such a massive amount of redundant information.

xoriath:
yeah, that's for OpenVSX :) However it does look like doing the queries against OpenVSX in parallel does not really gain much w.r.t time, so I don't actually see much issue doing this resolving stage in serial... Not sure if you have any differing opinion, @colin-grant-work  ?

colin-grant-work:
No, I agree - if the user chooses to do thing serially, we should try to do everything that we can serially, including the queries to Open VSX about versions.

xoriath:
But what if he chooses parallel... In my mind resolving this could also go in serial then just to make the code easier to follow...

xoriath:
@vince-fugnitto  / @colin-grant-work  do you guys think more is needed on this PR?

colin-grant-work:
@xoriath, apologies for the delay. I've been slowly chewing on the question of whether to run the Open VSX requests in parallel or sequence. Let me take another look this afternoon and I'll get back to you.

xoriath:
No rush, @colin-grant-work , I've just ended up that changing the queries back to parallel (or make them actually heed the new switch) is... a bit more meh... So at least in my mind I would like to put this in and see if we alleviate the issues that we have. If resolving against open-vsx is too slow, then deal with that in round 2.

However it looks like [eclipse/openvsx#379 ](https://github.com/eclipse/openvsx/issues/379) is discussing the data response format in itself.

ccorn:
The latest commit (f1f66c18b4c7a02d80f93b8e38927d5bc1174b93) in this PR, while needed for testing, would be duplicating part of #11113, so I think you should merge 8b22b76c6fa57b920965cfa8e56bf535b610399c instead or update #11113 accordingly.

I also think it is useful to keep the two essential commits (`--no-parallel` and serial version query) separate (i. e. no squashing). Because if OpenVSX succeeds in reducing their response size by a factor of 30, then the (reduced) total time needed could be dominated by connection buildup latencies, and then you might want to retry parallel version queries by tentatively reverting only 8b22b76c6fa57b920965cfa8e56bf535b610399c and keeping the `--no-parallel` commit.

vince-fugnitto:
I think we should either fix the implementation or revert the commit for now since we only respect the `parallel` flag for the initial download, but we do not respect it for the resolution of packs and dependencies which are now handled sequentially.

ccorn:
@colin-grant-work: Shouldn't the revert take place on the master branch? To me this looks as if the master branch has the changes, whereas this PR branch has them reverted, and I suppose it should be the other way raound.

colin-grant-work:
My PR is not a reversion of this PR - just an adjustment.

PR #11113, opened 2022-05-06
repo: update builtins

vince-fugnitto:
#### What it does

The commit updates additional builtins to use their proper `id` so they are not re-downloaded when resolving packs.

#### How to test

- confirm that the `yarn download:plugins` script passes and skips any plugins that are defined by `id` and included in the pack
- confirm that CI passes
- confirm that the `references-view` works well (now uses the version from the pack)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
   ],
   "theiaPluginsDir": "plugins",
   "theiaPlugins": {
+    "eclipse-theia.builtin-extension-pack": "https://open-vsx.org/api/eclipse-theia/builtin-extension-pack/1.50.1/file/eclipse-theia.builtin-extension-pack-1.50.1.vsix",
     "vscode.theme-abyss": "https://open-vsx.org/api/vscode/theme-abyss/1.62.3/file/vscode.theme-abyss-1.62.3.vsix",
     "vscode.theme-defaults": "https://open-vsx.org/api/vscode/theme-defaults/1.62.3/file/vscode.theme-defaults-1.62.3.vsix",
     "vscode.theme-kimbie-dark": "https://open-vsx.org/api/vscode/theme-kimbie-dark/1.62.3/file/vscode.theme-kimbie-dark-1.62.3.vsix",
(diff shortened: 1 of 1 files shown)


vince-fugnitto:
cc @ccorn

ccorn:
Suggestion: Just out of principle, I'd apply the same naming-by-ID to the extension pack.
Loosely related: Testing #11112, it seems that `ms-vscode.js-debug-companion` should be added to `theiaPluginsExcludeIds`.

ccorn (approved):
From my point of view, this changeset in its current state is complete and immediately useful e.g. for testing the first two commits of #11112.

ccorn:
With #11112 merged into master,  `ms-vscode.js-debug-companion` has now been added to the end of `theiaPluginsExcludeIds`. This PR adds it to the beginning, so this might not be flagged as a textual conflict, but semantically that would be redundant by now.

ccorn:
I wonder what the holdup here is. I find myself routinely merging this into whatever recent Theia checkout I have to investigate, just for my own sanity. I simply do not feel comfortable with the idea of having two (differing) versions of the `references-view` plugin installed.

vince-fugnitto:
@ccorn 

> I wonder what the holdup here is.

Someone needs to review, given how it does not affect downstream applications and only the examples present in this repo that we test against it is not critical.

> I find myself routinely merging this into whatever recent Theia checkout I have to investigate, just for my own sanity.

I'm curious to know how you consume the Theia framework, are you forking this repo and making an application (which is not ideal)? The plugins we define under our monorepo [package.json](https://github.com/eclipse-theia/theia/blob/aebf3fd0149d415e836d5674be9b7674ffa4fa0a/package.json#L93) do not affect downstream apps.

>  I simply do not feel comfortable with the idea of having two (differing) versions of the `references-view` plugin installed.

There is only one plugin actually installed, the duplication is only on disk. We do not install two different versions of a plugin with the same `id`.

colin-grant-work (approved):
LGTM. 👍

PR #11142, opened 2022-05-11
Fix `'.'` and `'..'` imports

colin-grant-work:
#### What it does

Imports from `index.ts` files in the same directory as a file or a higher level almost always entail circular dependencies. That doesn't have to be a catastrophe, but it's not a good idea. This PR adds a lint rule forbidding such imports and fixes those cases that currently exist. This reduces the number of warnings from `webpack` from 95 to 81.

#### How to test

1. Assert that there are no more imports `from '.'` or `from '..'` in our code.
2. Add such an import to a file.
3. Observe that the import is treated as a lint error.
4. Check that the application builds and runs as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/configs/errors.eslintrc.json
+++ b/configs/errors.eslintrc.json
@@ -139,7 +139,14 @@
     "@theia/runtime-import-check": "error",
     "@theia/shared-dependencies": "error",
     "import/no-extraneous-dependencies": "error",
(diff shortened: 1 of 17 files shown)


paul-marechal (approved):
LGTM, the ESLint rule is like icing on the cake ;)

PR #11150, opened 2022-05-12
add timestamp string to console log

DanboDuan:
Signed-off-by: bob <bob170731@gmail.com>

#### What it does

add timestamp string to console log

#### How to test

see the backend or frontend console log, it should be like:
```
5/15/2022, 08:17:56.269 root INFO xxxx
```

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/logger-protocol.ts
+++ b/packages/core/src/common/logger-protocol.ts
@@ -110,7 +110,7 @@ export namespace ConsoleLogger {
         const console = consoles.get(logLevel) || originalConsoleLog;
         const severity = (LogLevel.strings.get(logLevel) || 'unknown').toUpperCase();
         const now = new Date();
-        const timestamp = `${now.toLocaleString(undefined, { hour12: false })}.${now.getMilliseconds()}`;
+        const timestamp = `${now.toLocaleString('en-GB')}.${now.getMilliseconds()}`;
         console(`${timestamp} ${name} ${severity} ${message}`, ...params);
     }
 }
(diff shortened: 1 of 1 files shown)


DanboDuan:
could you please help merging this pr @msujew ? This is also for logging output. Timestamp information should be helpful when tracing some important performance issue.

packages/core/src/common/logger-protocol.ts
```diff
@@ -109,6 +109,8 @@ export namespace ConsoleLogger {
     export function log(name: string, logLevel: number, message: string, params: any[]): void {
         const console = consoles.get(logLevel) || originalConsoleLog;
         const severity = (LogLevel.strings.get(logLevel) || 'unknown').toUpperCase();
-        console(`${name} ${severity} ${message}`, ...params);
+        const now = new Date();
+        const timestamp = `${now.toLocaleString(undefined, { hour12: false })}.${now.getMilliseconds()}`;
```
msujew, line 113: Leaving `locale` at undefined will lead to unexpected behavior on different systems. `en-GB` seems to be the most sensible, as it uses `day/month/year` with a 24-hour cycle (no `AM/PM`) by default.

```suggestion
        const timestamp = `${now.toLocaleString('en-GB')}.${now.getMilliseconds()}`;
```

msujew (commented):
Works well, but I'm wondering whether we should use a better machine readable format such as `toISOString()`. Using ISO time would also be better in case the logs get sorted for some reason. @vince-fugnitto WDYT?

vince-fugnitto:
> using ISO time would also be better in case the logs get sorted for some reason. @vince-fugnitto WDYT?

@msujew sounds good to me, we can use `toISOString()`.

DanboDuan:
> > using ISO time would also be better in case the logs get sorted for some reason. @vince-fugnitto WDYT?
> 
> @msujew sounds good to me, we can use `toISOString()`.

OK, updated to use `toISOString()`

DanboDuan:
@msujew @vince-fugnitto so could you please help merge this ?

Issue #11158, opened 2022-05-16
Support traversing the editor navigation history through mouse buttons


martin-fleck-at:
### Feature Description
In VS Code it is possible to navigate the editor navigation history through commands, keyboard shortcuts and the mouse. In Theia, only the first two options are supported. It would be great if we could also utilize the mouse for navigation through the history, button 4 for 'Back' and button 5 for 'Forward' to align with VS Code and provide better usability.

Please note that in VS Code this functionality is guarded through a dedicated preference `workbench.editor.mouseBackForwardToNavigate` so we should do the same. Furthermore, this feature is a little bit more useful in the electron case as in the browser case the mouse 'Back' button is also used to navigate within the windows history. This is also true for the web-based VS Code, as can be tested by opening a [GitHub Dev]( https://github.dev/microsoft/vscode/blob/c23f0305dbf82b2319b198f4dbf3c5d5bc522f15/src/vs/workbench/services/history/browser/historyService.ts#L114) web editor and navigating around a bit.

martin-fleck-at:
I'll take this task on myself and will provide a PR for it.

PR #11163, opened 2022-05-17
Support traversing the editor navigation history through mouse buttons

martin-fleck-at:
#### What it does
- Adds support for traversing the editor navigation history through mouse buttons 4 and 5, similar to the [VS Code History Service](https://github.dev/microsoft/vscode/blob/c23f0305dbf82b2319b198f4dbf3c5d5bc522f15/src/vs/workbench/services/history/browser/historyService.ts#L114).
- Fixes https://github.com/eclipse-theia/theia/issues/11158

#### How to test
- Open application in browser or electron. If you use the browser you can open your application through a link so you don't have any history to begin with, otherwise the window history navigation will navigate you away from your Theia app. This is also the behavior in the GitHub web editor.
- Navigate within a file or between files.
- Click your mouse button 4 and 5 to navigate between your previous navigations.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/core-preferences.ts
+++ b/packages/core/src/browser/core-preferences.ts
@@ -97,6 +97,12 @@ export const corePreferenceSchema: PreferenceSchema = {
             'description': nls.localizeByDefault('Controls whether editors showing a file that was opened during the session should close automatically when getting deleted or renamed by some other process. Disabling this will keep the editor open  on such an event. Note that deleting from within the application will always close the editor and that dirty files will never close to preserve your data.'),
             'default': false
         },
+        'workbench.editor.mouseBackForwardToNavigate': {
+            'type': 'boolean',
+            // nls-todo: Will be available with newer VSCode API
+            'description': nls.localize('theia/core/mouseBackForwardToNavigate', "Enables the use of mouse buttons four and five for commands 'Go Back' and 'Go Forward'."),
+            'default': true
+        },
         'workbench.commandPalette.history': {
             type: 'number',
             default: 50,
@@ -166,6 +172,7 @@ export interface CoreConfiguration {
     'workbench.list.openMode': 'singleClick' | 'doubleClick';
     'workbench.commandPalette.history': number;
     'workbench.editor.highlightModifiedTabs': boolean;
+    'workbench.editor.mouseBackForwardToNavigate': boolean;
     'workbench.editor.closeOnFileDelete': boolean;
     'workbench.colorTheme': string;
     'workbench.iconTheme': string | null;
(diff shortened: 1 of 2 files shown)


packages/core/src/browser/core-preferences.ts
```diff
@@ -97,6 +97,12 @@ export const corePreferenceSchema: PreferenceSchema = {
             'description': nls.localizeByDefault('Controls whether editors showing a file that was opened during the session should close automatically when getting deleted or renamed by some other process. Disabling this will keep the editor open  on such an event. Note that deleting from within the application will always close the editor and that dirty files will never close to preserve your data.'),
             'default': false
         },
+        'workbench.editor.mouseBackForwardToNavigate': {
+            'type': 'boolean',
+            // nls-todo: Will be available with newer VSCode API
```
colin-grant-work, line 102: When it's available, this should be caught automatically by the linter (using `localizeByDefault` is an error if the string isn't available; using just `localize` is an error if the exact string is available).

packages/editor/src/browser/editor-navigation-contribution.ts
```diff
@@ -102,6 +103,39 @@ export class EditorNavigationContribution implements Disposable, FrontendApplica
...
+            if (change.preferenceName === EditorNavigationContribution.MOUSE_NAVIGATION_PREFERENCE) {
+                updateMouseNavigationListener();
+            }
+        }));
+        updateMouseNavigationListener();
+        this.toDispose.push(mouseNavigationSupport);
+    }
+
+    protected async onMouseDown(event: MouseEvent): Promise<void> {
+        // Support to navigate in history when mouse buttons 4/5 are pressed
```
colin-grant-work, line 128:
```suggestion
        // Support navigation in history when mouse buttons 4/5 are pressed
```

colin-grant-work (commented):
I don't have a mouse that would let me test this as it's meant to function, but some minor comments below.

packages/core/src/browser/core-preferences.ts
martin-fleck-at: Thank you for the clarification! I noticed that I got the error with `localizeByDefault` and checked what it is you are doing in the `localization-check.js`, it's very nice!

msujew (approved):
Looks good to me. Although I'm having a bit of trouble getting this to work in the browser (as my browsers try to leave the back when back-navigating), it works perfectly in Electron.

- [x] Back-navigation works correctly using button 3
- [x] Front-navigation works correctly using button 4
- [x] Other mouse buttons continue to work as expected

JonasHelming (approved):

Issue #11146, opened 2022-05-12
[vscode] Support `TextEditor#show()` and `TextEditor#hide()`


planger:
### Feature Description:
Add support for showing and hiding [text editors](https://code.visualstudio.com/api/references/vscode-api#TextEditor) with `show(column?: ViewColumn)` and `hide()`, see https://code.visualstudio.com/api/references/vscode-api#TerminalOptions.

jfaltermeier:
I will have a look at this

PR #11168, opened 2022-05-17
[vscode] Support TextEditor#show() and TextEditor#hide() #11146

jfaltermeier:
#### What it does
Add support for TextEditor#show() and TextEditor#hide()

VSCode-behaviour:
When using TextEditor.show(ViewColumn) VSCode opens the editor again in preview-mode. 
When using a numbered ViewColumn it brings the editor to the front, if existing, or opens it in the respective column. If the column is not existing it will be created.
TextEditor.hide() closes the editor.

Peek 2022-05-17 15-13

#### How to test
I've created a small vscode-extension here: https://github.com/jfaltermeier/vscode-playground/blob/f290f4b2f00cc3460e45b525eb822fe6824d986d/text-editor-show-hide/src/extension.ts
The vsix may be downloaded from here: https://github.com/jfaltermeier/vscode-playground/releases/tag/text-editor-show-hide-0.0.1

Start Theia with this extension and try the "Test show editor beside", "Test show editor in column one", "Test show editor in column two", "Test show editor in column three", and "Test hide editor" commands. 
These work on the `vscode.window.activeTextEditor`. 

#### Review checklist

#### Reminder for reviewers


packages/plugin-ext/src/main/browser/editors-and-documents-main.ts
```diff
@@ -168,6 +169,19 @@ export class EditorsAndDocumentsMain implements Disposable {
     saveAll(includeUntitled?: boolean): Promise<boolean> {
         return this.modelService.saveAll(includeUntitled);
     }
+
+    hideEditor(id: string): Promise<void> {
+        for (const editorWidget of this.editorService.all) {
```
jfaltermeier, line 174: Before trying to remove this loop, is there an existing way to get the EditorWidget for a MonacoEditor that I have missed?

vince-fugnitto:
@jfaltermeier 

> 1. Currently it seems that only one Preview editor can be opened for the same document in Theia. Is this correct? When I use "Split right" on a preview editor, the new editor is not opened in preview mode. We could decide to open in non-preview mode when using `TextEditor#show() ` as a mitigation.

As far as I can tell this is also the same behavior in VS Code. Splitting a `preview` editor will open the new editor normally and not as a preview.

> 2. When using the numbered ViewColumn it seems that this leads to a split-right in Theia. Is this a known issue?

How would we test this?

> 3. When using numbered ViewColumns, opening the document in an existing column does not seem to work. Is this a known issue?

How would we test this?

packages/editor/src/browser/editor-manager.ts
```diff
@@ -187,6 +187,20 @@ export class EditorManager extends NavigatableWidgetOpenHandler<EditorWidget> {
...
+        /* check whether we have a split request without a counter */
+        if (options && options.counter === undefined && options.widgetOptions?.mode) {
+            switch (options.widgetOptions.mode) {
+                case 'split-right':
+                case 'split-bottom':
+                case 'split-top':
+                case 'split-left':
+                    const counter = this.createCounterForUri(uri);
+                    const splitOptions: EditorOpenerOptions = { counter, ...options };
+                    return super.open(uri, splitOptions);
```
colin-grant-work, line 200:
```suggestion
                    return this.openToSide(uri, options);
```

packages/plugin-ext/src/common/plugin-api-rpc.ts
```diff
@@ -1157,11 +1157,11 @@ export interface DecorationOptions {
 }
 
 export interface TextEditorsMain {
-    // $tryShowTextDocument(resource: UriComponents, options: TextDocumentShowOptions): Promise<string>;
+    $tryShowTextDocument(uri: UriComponents, options?: TextDocumentShowOptions): Promise<void>;
     $registerTextEditorDecorationType(key: string, options: DecorationRenderOptions): void;
     $removeTextEditorDecorationType(key: string): void;
     // $tryShowEditor(id: string, position: EditorPosition): Promise<void>;
```
colin-grant-work, line 1163: Should we implement / delete this variant?

packages/plugin-ext/src/main/browser/editors-and-documents-main.ts
colin-grant-work: Since the editor data reported to the plugin side has gone through the `EditorSnapshot` system, I don't really see a way recover the editor meant other than to repeat the process.

colin-grant-work:
> 1. Currently it seems that only one Preview editor can be opened for the same document in Theia. Is this correct?
>    When I use "Split right" on a preview editor, the new editor is not opened in preview mode.

Only one preview can be open at all. If any other editor is opened as a preview, the old preview will be closed and the new preview will replace it.

> 2. When using the numbered ViewColumn it seems that this leads to a split-right in Theia. Is this a known issue?

That appears to be a consequence of the implementation of `toEditorOpenerOptions` [here](https://github.com/eclipse-theia/theia/blob/7d1f7241fa30f033a04e025501719d6dcdf9e549/packages/plugin-ext/src/main/browser/documents-main.ts#L235-L267). The initial default is `split-right`, and if the column requested is greater than the number of columns currently available, that default remains in place.

> 3. When using numbered ViewColumns, opening the document in an existing column does not seem to work. Is this a known issue?

Based on the code in the method referred to above, this _looks_ like it should work: if the column exists, the active editor in that column should be used as the ref for opening the new widget.

colin-grant-work:
I've pushed a commit [here](https://github.com/colin-grant-work/theia/commit/9ae44ec37f06288d38665a1f405577f5bd48feca) that should get the behavior a closer to VSCode.

- Editor previews are now limited to one per tab bar, not one per application.
- `EditorManager.open` figures out where the new editor would go and only reuses an existing editor if the widget would be opened on that widget's tabbar. Otherwise, a new editor is opened.

packages/plugin-ext/src/common/plugin-api-rpc.ts
jfaltermeier: I will remove it. Looking at the arguments the idea might have been to show/split an existing editor to a tab bar. 
I think this can be achieved with `tryShowTextDocument` as well

colin-grant-work:
> Can I cherry-pick your change to this branch?

Yes, by all means, feel free.

vince-fugnitto (commented):
@jfaltermeier @JonasHelming I'm not sure about implementing the `deprecated` api, can't we just stub it? Extensions really shouldn't be using this API and it has been deprecated in VS Code from what it seems like the start (when it was public). I traced the deprecation back 7 years:

image

packages/editor-preview/src/browser/editor-preview-manager.ts
```diff
@@ -44,17 +42,19 @@ export class EditorPreviewManager extends EditorManager {
...
             }
         });
 
         this.preferences.onPreferenceChanged(change => {
             if (!change.newValue) {
-                this.currentPreview?.convertToNonPreview();
+                this.all.forEach((editor: EditorPreviewWidget) => {
+                    if (editor.isPreview) {
+                        editor.convertToNonPreview();
+                    }
```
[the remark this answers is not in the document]
jfaltermeier: Thanks, I've added the check

jfaltermeier:
@vince-fugnitto I agree that we should not spent much effort into fully implementing deprecated API. Here the actual implementation of the deprecated api is not a big code change however (see the first commit): https://github.com/eclipse-theia/theia/pull/11168/commits/fda59378729934493df0f5d3065e2d62a97255b6
`hide` loops over all widgets of the factory to find the editor to hide, which is a stub-implementaton. And `show` delegates to existing API. 
All the other changes are related to differences in behaviour between Theia and VSCode, but those could also be triggered by using the proposed non-deprecated API `window.showTextDocument`
Should I open a further issue for the changelog explaining those differences? Then this PR could close two issues.

vince-fugnitto:
@jfaltermeier @colin-grant-work if you still feel like the deprecated methods should in fact be supported then please disregard my comments :) I just thought it was a little odd to actually support the api given it was deprecated from the start.

colin-grant-work:
I think I'm inclined to merge the implementation. It's strange that VSCode included deprecated API in their first public commit, and even stranger that they haven't removed it in the intervening 7 years, but as long as we've committed to supporting the VSCode plugin API, we should not reject a sound implementation, even of deprecated API.

vince-fugnitto:
@jfaltermeier for the `show editor beside` command I noticed a difference between our behavior and vscode, it seems that vscode checks first if the requested editor already exists in that location and focuses it, rather than open a new editor:

_theia_:

show-editor-beside

_vscode_:

vscode-show-editor-beside

##### Steps:

1. open an editor
2. execute `test show editor beside`
3. focus original editor
4. re-do 2nd step

jfaltermeier:
I'm on vacation until June 7th. I will pick this up again when I'm back.

vince-fugnitto (commented):
I confirm that the changes work as expected, I did not notice any regressions :+1:

vince-fugnitto (approved):

JonasHelming:
@colin-grant-work : Fine if I merge this?

colin-grant-work:
Ok with me. 👍

JonasHelming:
@jfaltermeier : Please resolve conflict and merge ;-)

PR #11181, opened 2022-05-20
feat: add message batcher for terminal process

zhaomenghuan:
#### What it does

Batch the output of the terminal, combine the output content within 16ms and send it to the front end in batches, so as to avoid blocking RPC communication when the terminal executes a lot of output commands. Inspired by [opensumi#304](https://github.com/opensumi/core/pull/304) and [hyper 3](https://hyper.is/blog).

#### How to test

- Click the new terminal menu to create a new terminal
- Enter the `find ~` command, then abort the command, the terminal stops immediately

#### Review checklist

#### Reminder for reviewers

Signed-off-by: zhaomenghuan <1028317108@qq.com>

### DIFF
--- a/packages/process/src/node/terminal-process.ts
+++ b/packages/process/src/node/terminal-process.ts
@@ -45,6 +45,12 @@ export enum NodePtyErrors {
     ENOENT = 'No such file or directory'
 }
 
+// Max size of a data batch. Note that this value can be exceeded by ~4k (chunk sizes seem to be 4k at the most)
+export const BATCH_MAX_SIZE = 200 * 1024;
+
+// Max duration to batch data before sending it to the renderer process.
+export const BATCH_DURATION_MS = 16;
+
 /**
  * Run arbitrary processes inside pseudo-terminals (PTY).
  *
@@ -54,6 +60,8 @@ export enum NodePtyErrors {
 export class TerminalProcess extends Process {
 
     protected readonly terminal: IPty | undefined;
+    protected batchedData: string;
+    protected timeout: NodeJS.Timeout | undefined;
 
     readonly outputStream = this.createOutputStream();
     readonly errorStream = new DevNullStream({ autoDestroy: true });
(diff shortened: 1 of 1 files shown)


vince-fugnitto (commented):
@zhaomenghuan it seems like the macos tests are failing with this change.

zhaomenghuan:
> @zhaomenghuan it seems like the macos tests are failing with this change.

@vince-fugnitto The code has been adjusted, and the test case error has been fixed.

packages/process/src/node/terminal-process.ts
```diff
@@ -45,6 +45,12 @@ export enum NodePtyErrors {
     ENOENT = 'No such file or directory'
 }
 
+// Max size of a data batch. Note that this value can be exceeded by ~4k (chunk sizes seem to be 4k at the most)
+export const BATCH_MAX_SIZE = 200 * 1024;
+
+// Max duration to batch data before sending it to the renderer process.
+export const BATCH_DURATION_MS = 16;
```
msujew, line 52: While testing this, I've noticed that 16 ms is way too short of a delay. Every other action was heavily delayed, often until the command running in the terminal has ended. Changing this to `64` has improved this behavior dramatically, basically enabling normal app interaction.

msujew (commented):
Looks quite good. However with the `16 ms` duration the app was still unusable (web mode). Increasing the duration improved responsiveness without noticeably impacting the terminal output.

msujew (approved):
Actually, I'll dismiss my previous comment, as that seems to be related to an issue I experienced in Gitpod only. I can't reproduce that anymore, so I'll approve it.

- [x] The changes allow for app interaction even when faced with a lot of terminal output
- [x] I couldn't find any regressions using the changes.

@vince-fugnitto Do you mind testing this as well? It seems like this could be influenced by the environment Theia runs in.

paul-marechal:
I am not sold on putting those changes at the `TerminalProcess` level: Ultimately we want to control the size and timing of the RPC requests we make to the frontend. We also use independent ring buffer readers that may not benefit from this change.

May I propose an alternative change? https://github.com/eclipse-theia/theia/pull/11449

PR #11182, opened 2022-05-20
feat(core): Fix Chinese localization translation

zhaomenghuan:
#### Review checklist

#### Reminder for reviewers

Signed-off-by: zhaomenghuan <1028317108@qq.com>


packages/core/i18n/nls.zh-cn.json
```diff
@@ -1,25 +1,25 @@
...
+        "closeAll": "全部关闭",
+        "closeAllTabMain": "关闭主区域的所有标签页",
+        "closeOtherTabMain": "关闭主区域的其他标签页",
+        "closeOthers": "关闭其他标签页",
+        "closeRight": "关闭右侧标签页",
+        "closeTab": "关闭标签页",
+        "closeTabMain": "关闭主区域标签页",
+        "collapseAllTabs": "折叠所有标签页",
+        "collapseBottomPanel": "折叠底部面板",
+        "collapseTab": "折叠标签页",
```
colin-grant-work, line 17: Although the key is 'collapseTab', the English here is 'Collapse Side Panel'. Would it make sense to use a translation similar to 'Collapse Bottom Panel' above, something with '面板' rather than '标签页'?

packages/core/i18n/nls.zh-cn.json
```diff
@@ -160,9 +160,9 @@
       "compareWith": "对比...",
       "compareWithBranchOrTag": "挑选一个分支或标签,与当前活动的{0}分支进行比较。",
       "dirtyDiffLinesLimit": "如果编辑器的行数超过这个限制,则不显示肮脏的差异装饰。",
-      "dropStashMessage": "藏品成功删除。",
-      "editorDecorationsEnabled": "在编辑器中显示git的装饰。",
-      "fetchPickRemote": "挑选一个遥控器来取物。",
+      "dropStashMessage": "存储删除成功。",
+      "editorDecorationsEnabled": "控制 Git 是否在打开编辑器视图中添加颜色和小标",
```
colin-grant-work, line 164: This appears to be a translation of the description for the VSCode Git preference `git.decorations.enabled`, the English text of which is 'Controls whether Git contributes colors and badges to the Explorer and the Open Editors view.' However, the Theia text to be translated here is 'Show git decorations in the editor.' In particular, the Theia text does not mention 'open editors' (打开编辑器视图) or  'badges' (小标).

colin-grant-work (commented):
To the extent I can judge, this is a vast improvement. Many thanks for the contribution! I have only a couple of comments below.

@msujew, in the case of the editor preference description translations, this touches translations of strings taken from Monaco and likely covered by newer versions of VSCode's translations. Even if we don't update our base `nls.metadata.json` when we update Monaco, might there be a way to incorporate a check of VSCode's translation repo when doing the automated translations so we don't end up with bad translations that we could have avoided?

packages/core/i18n/nls.zh-cn.json
zhaomenghuan: > This appears to be a translation of the description for the VSCode Git preference `git.decorations.enabled`, the English text of which is 'Controls whether Git contributes colors and badges to the Explorer and the Open Editors view.' However, the Theia text to be translated here is 'Show git decorations in the editor.' In particular, the Theia text does not mention 'open editors' (打开编辑器视图) or 'badges' (小标).

Original translation restored

msujew:
> might there be a way to incorporate a check of VSCode's translation repo when doing the automated translations so we don't end up with bad translations that we could have avoided?

@colin-grant-work Hm, maybe. I'm not too sure how difficult this would be. While we can always access the [`vscode-loc`](https://github.com/microsoft/vscode-loc) repo for this, it doesn't contain the English version of the text. So it becomes quite difficult to find the correct translation, since we don't have the correct keys (we create our own) and we can't search by the English text.

colin-grant-work (approved):
Thanks for the contribution! 非常感谢你的贡献！

PR #11195, opened 2022-05-25
core: bind `UntitledResourceResolver` in core

vince-fugnitto:
#### What it does

The commit updates the binding of `UntitledResourceResolver` so that it is bound in `core` as it was moved around previously to be accessible outside of the plugin-system. Since the old binding was not updated the untitled functionality would break for apps that did not depend on plugin dependencies ([forum discussion](https://community.theia-ide.org/t/error-child-node-open-editors-root-does-not-belong-to-this-open-editors-root-tree/2407))

#### How to test

1. confirm that the `untitled` behavior still works well (`new editor` in the application).
2. a little trickier to test but the idea is if an application does not have `plugin` dependencies then the `new editor` functionality should still work.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/frontend-application-module.ts
+++ b/packages/core/src/browser/frontend-application-module.ts
@@ -30,7 +30,8 @@ import {
     MessageClient,
     InMemoryResources,
     messageServicePath,
-    InMemoryTextResourceResolver
+    InMemoryTextResourceResolver,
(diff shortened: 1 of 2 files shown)


colin-grant-work (approved):
Correctly allows the creation and handling of untitled files without `plugin-ext`.

PR #11207, opened 2022-05-26
Add electron security token to backend request service

msujew:
#### What it does

Just after merging https://github.com/eclipse-theia/theia/pull/11043, I noticed a minor issue with requests to the localhost in Electron. When trying to read the README.md of an already installed extension, the backend rejects the request (although it's coming from the backend server), since it does not use the electron security token. This change addresses this issue by adding the security token if we perform a request to the `localhost`.

#### How to test

1. Download and install an extension
2. Open the extension overview/readme
3. It should render the readme correctly
4. Other requests to the backend localhost address should also run as expected

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-node/request/electron-backend-request-service.ts
+++ b/packages/core/src/electron-node/request/electron-backend-request-service.ts
@@ -15,7 +15,8 @@
  ********************************************************************************/
 
 import { decorate, injectable } from 'inversify';
-import { NodeRequestService } from '@theia/request/lib/node-request-service';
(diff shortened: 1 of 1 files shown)


msujew:
@paul-marechal Alternatively, we can change it so Electron requests aren't tunneled through the backend. This removes the need to add special backend handling for this, at the cost of some confusion. I.e. some requests getting through the proxy (XHR which uses the browser's proxy resolution), while some others don't (backend requests).

PR #11219, opened 2022-05-30
[electron]: Moved the menu creation/update to electron-main.

kittaakos:
#### What it does

This PR moves the electron menu creation/update to the electron-main process.

OK but why?
 - Electron `remote` has been deprecated and will be moved outside of electron, (https://github.com/electron/electron/issues/21408)
 - `getCurrentWindow` is a synchronous call and blocks the thread. This is problematic when Theia (or its customization) opens with several browser windows at startup,
 - This will help to remove the rest of the Electron `remote` code from Theia and use IPC between the main and the renderers.

One of my top priorities during the development was to reduce the amount of breaking changes. Please chime in and propose changes. Feedback is welcome. Thank you! 🧁

#### How to test
 - Start the Electron app in debug mode, and check the console how the context menu items are cleaned up on the main process,
 - TBD,

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-common/messaging/electron-messages.ts
+++ b/packages/core/src/electron-common/messaging/electron-messages.ts
@@ -14,6 +14,8 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 // *****************************************************************************
 
+import type { MenuItemConstructorOptions } from '../menu';
+
 export const RequestTitleBarStyle = 'requestTitleBarStyle';
 export const TitleBarStyleChanged = 'titleBarStyleChanged';
 export const TitleBarStyleAtStartup = 'titleBarStyleAtStartup';
@@ -50,4 +52,71 @@ export interface CloseRequestArguments {
     confirmChannel: string;
     cancelChannel: string;
     reason: StopReason;
+}
+
+/**
+ * Request is sent by the render to the main to rebuild and set the menu.
+ */
+export namespace SetMenu {
+    export const Signal = 'setMenu';
+    export interface Params {
+        readonly template: MenuItemConstructorOptions[] | undefined;
+    }
+}
+/**
+ * Request sent by the render to the main to build and show the context menu.
+ */
+export namespace ShowContextMenu {
+    export const Signal = 'showContextMenu';
+    export interface Params {
+        readonly template: MenuItemConstructorOptions[];
+        /** Must be an integer. */
+        readonly x: number;
+        /** Must be an integer. */
+        readonly y: number;
+        readonly contextMenuId: string;
+    }
+};
+/**
+ * Sent by the renderer to main to programmatically close an open context menu identified by the unique `contextMenuId`.
+ * This request might be ignored by the recipient if no open context menu exists with the ID. For example, the context menu was closed by user clicking away.
+ */
+export namespace CloseContextMenu {
+    export const Signal = 'closeContextMenu';
+    export interface Params {
+        readonly contextMenuId: string;
+    }
+}
+/**
+ * Event sent by the main to the render when a context menu was closed.
+ */
+export namespace ContextMenuDidClose {
+    export const Signal = 'contextMenuDidClose';
+    export interface Params {
+        readonly contextMenuId: string;
+    }
+}
+/**
+ * Event sent by the main to the renderer when one of the menu items was clicked.
+ */
+export namespace MenuItemDidClick {
+    export const Signal = 'menuItemDidClick';
+    export interface Params {
+        readonly commandId: string;
+        /**
+         * The arguments will be serialized with the [_Structured Clone Algorithm_](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).
+         * Sending Functions, Promises, Symbols, DOM or special electron objects, WeakMaps, or WeakSets will throw an exception.
+         */
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        readonly args?: any[];
+    }
+}
+/**
+ * Sent by the renderer to the main to updated the `checked` state of the menu item IDs.
+ */
+export namespace UpdateMenuItems {
+    export const Signal = 'updateMenuItems';
+    export interface Params {
+        readonly menuItems: { id: string, checked: boolean }[];
+    }
 }
--- a/packages/core/src/electron-main/electron-main-menu.ts
+++ b/packages/core/src/electron-main/electron-main-menu.ts
@@ -15,42 +15,38 @@
 // *****************************************************************************
 
 import { injectable } from 'inversify';
-import { app, BrowserWindow, ipcMain, WebContents } from '../../electron-shared/electron';
+import { app, BrowserWindow, ipcMain, IpcMainEvent, WebContents } from '../../electron-shared/electron';
 import { Menu } from '../../electron-shared/electron/index';
 import { isOSX } from '../common';
 import { MenuItemConstructorOptions } from '../electron-common/menu';
-import {
-    CloseContextMenu,
-    ContextMenuDidClose, MenuItemDidClick, SetMenu, ShowContextMenu, UpdateMenuItems
-} from '../electron-common/messaging/electron-messages';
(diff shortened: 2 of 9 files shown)


packages/core/src/electron-common/messaging/electron-messages.ts
```diff
@@ -51,3 +53,70 @@ export interface CloseRequestArguments {
     cancelChannel: string;
     reason: StopReason;
 }
+
+/**
+ * Request is sent by the render to the main to rebuild and set the menu.
```
colin-grant-work, line 58:
```suggestion
 * Request is sent by the renderer to the main to rebuild and set the menu.
```

packages/core/src/electron-common/messaging/electron-messages.ts
```diff
@@ -51,3 +53,70 @@ export interface CloseRequestArguments {
...
+ */
+export namespace SetMenu {
+    export const Signal = 'setMenu';
+    export interface Params {
+        readonly template: MenuItemConstructorOptions[] | undefined;
+    }
+}
+/**
+ * Request sent by the render to the main to build and show the context menu.
```
colin-grant-work, line 67:
```suggestion
 * Request sent by the renderer to the main to build and show the context menu.
```

colin-grant-work (commented):
Planning to do a more thorough review later today, but noticed these when I was skimming the code.

packages/core/src/electron-main/electron-main-menu.ts
```diff
@@ -0,0 +1,238 @@
...
+import { isOSX } from '../common';
+import { MenuItemConstructorOptions } from '../electron-common/menu';
+import {
+    CloseContextMenu,
+    ContextMenuDidClose, MenuItemDidClick, SetMenu, ShowContextMenu, UpdateMenuItems
+} from '../electron-common/messaging/electron-messages';
+import { ElectronMainApplicationContribution } from './electron-main-application';
+
+@injectable()
+export class ElectronMainMenu implements ElectronMainApplicationContribution {
```
colin-grant-work, line 29: `ElectronMainMenu` is a bit ambiguous, esp. since this also handles context menus. Perhaps `ElectronMenuHandler` on `ElectronApplicationMenuHandler`?

packages/core/src/electron-main/electron-main-menu.ts
```diff
@@ -0,0 +1,238 @@
+// *****************************************************************************
```
colin-grant-work, line 1: Since this file doesn't use the logging utilities defined in the browser and backend, `console.debug` is just an alias of `console.log`, and all of the debug logs show up in the terminal. They should be removed or muted.

colin-grant-work (commented):
With these changes, I'm not seeing any context menus on tabbar items, either in the side panels or the main area tabbar (I use MacOS).

When I right clicked on a `ViewContainerPart` (the NPM view, collapsed), I got a log about failure to clone an object:

image

packages/core/src/electron-common/messaging/electron-messages.ts
```diff
@@ -51,3 +53,70 @@ export interface CloseRequestArguments {
...
+    export interface Params {
+        readonly contextMenuId: string;
+    }
+}
+/**
+ * Event sent by the main to the renderer when one of the menu items was clicked.
+ */
+export namespace MenuItemDidClick {
+    export const Signal = 'menuItemDidClick';
+    export interface Params {
```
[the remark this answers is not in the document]
kittaakos: I do not understand the question. Please expand. Thank you!

packages/core/src/electron-main/electron-main-menu.ts
kittaakos: Thanks for the remark. Sure. They won't be in the final version, at least as-is now.

packages/core/src/electron-main/electron-main-menu.ts
kittaakos: I see your point; however, appending `Handler`, `Manager`, or `Service`, .etc does not clarify much. Please pick a name, and I will use that in the next PR update. Thanks!

packages/core/src/electron-main/electron-main-menu.ts
```diff
@@ -0,0 +1,238 @@
...
+    // keys are the sender IDS, values are the current menu per window.
+    protected readonly menus = new Map<number, Menu | null>();
+    // keys are sender IDs, values are the open context menus per sender.
+    protected readonly openContextMenus = new Map<number, Map<string, Menu>>();
+
+    onStart(): void {
+        ipcMain.on(SetMenu.Signal, ({ sender }, params: SetMenu.Params) => this.setMenu({ sender }, params));
+        ipcMain.on(UpdateMenuItems.Signal, ({ sender }, params: UpdateMenuItems.Params) => this.updateMenuItems({ sender }, params));
+        ipcMain.on(ShowContextMenu.Signal, ({ sender }, params: ShowContextMenu.Params) => this.showContextMenu({ sender }, params));
+        ipcMain.on(CloseContextMenu.Signal, ({ sender }, params: CloseContextMenu.Params) => this.closeContextMenu({ sender }, params));
```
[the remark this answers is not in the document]
kittaakos: Because the `sender` is sufficient in the implementation.

kittaakos:
Thank you for your time for the review, @colin-grant-work. I will do the updates in the next couple of days.

packages/core/src/electron-main/electron-main-menu.ts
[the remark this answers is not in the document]
colin-grant-work: Destructuring out the sender value is less performant than simply passing the pointer to the event (it requires both the destructuring itself and the allocation of the object containing only `{sender}` - trivial, but true) and limits the customization possible in the handlers, since they lack access to the rest of the fields on the event. Unless you foresee a use case where we need to perform these actions and _only_ have access to the `BrowserWindow`, it would be better - more extensible - to just pass around the event.

packages/core/src/electron-common/messaging/electron-messages.ts
```diff
@@ -51,3 +53,70 @@ export interface CloseRequestArguments {
...
+        readonly commandId: string;
+        /**
+         * The arguments will be serialized with the [_Structured Clone Algorithm_](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).
+         * Sending Functions, Promises, Symbols, DOM or special electron objects, WeakMaps, or WeakSets will throw an exception.
+         */
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        readonly args?: any[];
```
colin-grant-work, line 111: The renderer should always already know the arguments that should be passed to command handlers, so I'm not sure we ever need to send complex data between the renderer and the main process. The main process only needs to know enough to render the menu - the renderer has all the information it needs to know how to handle a click when the main process informs it that one has occurred.

kittaakos: I agree.

kittaakos:
I do not know why GH does not allow me to make inline comments:

> If that's important - ideally, it shouldn't be, because only one menu should be open at once

Ah, I see what you mean. Thanks! So the `commandId` and the id of the MenuItem are the same:

https://github.com/eclipse-theia/theia/blob/5f887e22825c7284b16104b517678ca0a2824319/packages/core/src/common/menu.ts#L422-L424 

https://github.com/eclipse-theia/theia/blob/5f887e22825c7284b16104b517678ca0a2824319/packages/core/src/electron-browser/menu/electron-main-menu-factory.ts#L193

I do not think we should send `id` and `commandId` with the same values, but we can clarify somehow on the API that they're the same. What do you think?

> Destructuring out the sender value is less performant than simply passing the pointer to the event (it requires both the destructuring itself and the allocation of the object containing only {sender} - trivial, but true) and limits the customization possible in the handlers, since they lack access to the rest of the fields on the event. Unless you foresee a use case where we need to perform these actions and only have access to the BrowserWindow, it would be better - more extensible - to just pass around the event.

I really do not want to get into premature optimization here. I want to solve a blocking, sync call when building the menus repeatedly. Regarding the APIs and further customizations, I appreciate you keeping the extensibility in mind; however, a Theia version update is always so tricky for downstreams. I do not want to make an API from everything, unlike you.

paul-marechal:
> I want to solve a blocking, sync call when building the menus repeatedly.

The end doesn't justify the means.

I think most of the issues with `electron-main-menu.ts` had to do with the really weird code style.

I pushed a commit to make it easier to read.

PR #11227, opened 2022-05-31
Update Developing.md

fipro78:
Fix indentation

Signed-off-by: Dirk Fauth <dirk.fauth@googlemail.com>

### DIFF
--- a/doc/Developing.md
+++ b/doc/Developing.md
@@ -515,7 +515,7 @@ etc.) by opening `packages/<package name>/coverage/index.html`.
  - If you need to install `windows-build-tools`, see [`Installing Windows Build Tools`](#installing-windows-build-tools).
  - If you run into problems with installing the required build tools, the `node-gyp` documentation offers a useful [guide](https://github.com/nodejs/node-gyp#on-windows) how to install the dependencies manually. The versions required for building Theia are:
    - Python 3.6 or higher
-- Visual Studio [build tools](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2022) 17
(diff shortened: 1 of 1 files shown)


JonasHelming:
@fipro78 : Hi Dirk and thank you for the contribution! Could you please check whether your ECA is signed with the same E-Mail address used in the commit? Otherwise the ECA validation will fail (it currently fails)

fipro78:
@JonasHelming Well the Signed-off-by email is my Eclipse account email address ... but as I did a simple modification in the GitHub browser UI it seems the author email is not set accordingly. Any idea how I can change this?

Otherwise it is a very simple formatting fix as the list was not correct and confused me a bit. If there is too much hassle to get things working, feel free to abandon this change and fix it directly please. :)

I think I need to get my private GitHub account configured to make it work with Eclipse GitHub projects. Not sure how but I will find out.

msujew:
@fipro78 You can do something like this:

```sh
git commit --amend --author="John Doe <john@doe.org>"
```

To change the name and email address of your last commit.

fipro78:
@msujew Thanks, I am aware of the command line options. :) 

The problem is that I am behind a corporate firewall and can not push via ssh or https to GitHub. So I would need to start my private laptop, checkout the sources, modify and push. And my hope was that with the editing on GitHub in the browser I could skip that. At the end that is the major benefit of a cloud based editor, to skip the local stuff for minimal changes.

fipro78:
I created a new PR #11234 with the correct email address as author. The local checkout variant would have surely worked, but I wanted to verify a working process in the web based user interface. Hope that this is ok for you.

Issue #11203, opened 2022-05-25
New IPC does not handle debugging gracefully


colin-grant-work:
### Bug Description:

When debugging in the browser and after sitting on a breakpoint for some time, resuming the application produces tens of thousands of logs and crashes the browser.

### Steps to Reproduce:

0. In the browser application (may also apply to Electron)
1. Set a break point in the developer tools and hit it.
2. Wait a few minutes.
3. Remove your breakpoint and hit continue.
4. Observe tens of thousands of logs about failure to send a message. I took the screenshot below a few seconds into the process - eventually the counter got over 250,000 before the browser killed the page as unresponsive.

@tortmayr, @JonasHelming, @paul-marechal

<img width="624" alt="image" src="https://user-images.githubusercontent.com/62660806/170374360-9b2e45d8-9056-4829-be49-abcad7da1c60.png">

### Additional Information

- Operating System: OS X, Chrome 101
- Theia Version: 9a10b6b5daab5e046f28fde8803b3612e84f6ff4

colin-grant-work:
For me, this is a showstopper - if we don't fix this, we shouldn't release with the new IPC.

JonasHelming:
@colin-grant-work : Just to be sure I get this right: You run a Theia-based application (e.g. the example application) and then you set a break point in this Theia-based application using the developer tools, correct?

colin-grant-work:
> @colin-grant-work : Just to be sure I get this right: You run a Theia-based application (e.g. the example application) and then you set a break point in this Theia-based application using the developer tools, correct?

Exactly. This was in the example application on `master` in the browser, and I set a breakpoint (`@theia/monaco-editor-core/src/vs/editor/common/languages.ts:143`, though I don't think that particular detail matters), browsed some source code for a while, and when I resumed the application after removing my breakpoint, the console flooded and it crashed. I've reproduced it a couple of times.

JonasHelming:
> > @colin-grant-work : Just to be sure I get this right: You run a Theia-based application (e.g. the example application) and then you set a break point in this Theia-based application using the developer tools, correct?
> 
> Exactly. This was in the example application on `master` in the browser, and I set a breakpoint (`@theia/monaco-editor-core/src/vs/editor/common/languages.ts:143`, though I don't think that particular detail matters), browsed some source code for a while, and when I resumed the application after removing my breakpoint, the console flooded and it crashed. I've reproduced it a couple of times.

Might be even related to #11199

[...]

Issue #11199, opened 2022-05-25
No Yellow Status Bar on Disconnect


colin-grant-work:
### Bug Description:

In the browser application, the status bar no longer changes to a yellow color when the connection is lost.

### Steps to Reproduce:

1. Start the browser application and connect in a browser.
2. Kill the backend.
3. Observe that the application status bar never changes color to indicate the lost connection.

Likely to do with recent changes to the IPC protocol. Possibly intentional?

CC: @tortmayr, @JonasHelming 

### Additional Information

- Operating System: OS X
- Theia Version: 1a6c7366ddb1c90f5f1b39687b5502a9a59b01a5

PR #11228, opened 2022-05-31
Reapply changes for new message rpc

tortmayr:
#### What it does

Reapplies the changes for the binary message rpc, originally introduced with #11011 and fixes the critical issues which lead to reverting the original change. 

In particular the following issues are addressed:

- #11199  Ensure that the new message rpc properly handles backend disconnects/reconnects.
- #11203 Ensure that debugging is handled gracefully (implicitly fixed with the fix for #11199)

These bugs were caused by a combination of  underlying issues:
- Trying to send RPC calls on a already disposed/disconnected websocket channel resulted in an endless loop due to using `console.warn`. Using console logging in the browser indirectly also triggers an RPC call to the `LoggerServer`. Naturally this call also fails and due to the console logging triggers another rpc call => we end up in and endless loop. Fixed by removing the console.warn statement in `websocket-channel.ts`.
- On socket connection loss (i.e. backend disconnect) channels and `RpcProtocol` instances were only partly disposed. This means once the backend reconnected did end up with multiple `RpcProtocl` instances for the same proxy which lead to weird issues e.g. Sending a request from one instance but the reply for the request is then handled by the other instance.
This is now fixed by ensuring that all components involved in the rpc communication (websocket main channel, multiplexer, multiplexed service channels, `RpcProtocol` instances) are fully disposed on disconnect and re-instantiated on reconnect.
- On reconnect the `connection-service` was informed too early i.e. before all previously open channels have been restored. As a consequence some RPC calls (e.g. ping requests) were triggered before the responsible service channel was available.
Fixed in `WebSocketConnectionProvider` by ensuring that the `onSocketDidOpen` event is fired after all previously openend channels have been restored.
- The `onIncomingMessgeActivity` event of the `WebSocketConnectionProvider` was never fired.
### How to test

To test the fix for #11199:

- Start the browser application and connect in a browser.
- Kill the backend.
- Application status should change color to indicate the lost connection.
- Restart backend
- Application should reconnect and work as before the connection loss

To test the fix for #11203 
- In the browser application (may also apply to Electron)
- Set a break point in the developer tools and hit it.
- Wait a few minutes.
- Remove your breakpoint and hit continue.
- Debugging process should continue without any issues and application should be responsive. No errors about failures to send messages should be logged.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,13 @@
   - Added the following method to the interface `DebugService`: `fetchDynamicDebugConfiguration` as well as the property `onDidChangedDebugConfigurationProviders`.
   - Removed method `DebugPrefixConfiguration#runDynamicConfiguration`
   - [core] The interface `SelectComponentProps` was updated to rename a property from `value` to `defaultValue`
+- [core] Refactored the core messaging API. Replaced `vscode-ws-jsonrpc` with a custom RPC protocol that is better suited for handling binary data and enables message tunneling.
+  This impacts all main concepts of the messaging API. The API no longer exposes a `Connection` object and uses a generic `Channel` implementation instead.
+  - Replaces usage of `vscode-json-rpc`'s `Connection` with the new generic `Channel`. Affects `AbstractConnectionProvider`, `MessagingService`, `IPCConnectionProvider`, `ElectronMessagingService`
+  - `MessagingService`: No longer offers the `listen` and `forward` method. Use `wsChannel` instead.
+  - `RemoteFileSystemServer`: Use `UInt8Array` instead of plain number arrays for all arguments and return type that store binary data
+  - `DebugAdapter`: Replaced the debug-service internal `Channel` implementation with the newly introduced generic `Channel`.
+  [#11228](https://github.com/eclipse-theia/theia/pull/11228) - Contributed on behalf of STMicroelectronics.
 
 ## v1.26.0 - 5/26/2022
 
--- a/package.json
+++ b/package.json
@@ -10,6 +10,8 @@
     "**/@types/node": "12"
   },
   "devDependencies": {
+    "@types/chai": "4.3.0",
+    "@types/chai-spies": "1.0.3",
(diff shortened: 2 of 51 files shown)


colin-grant-work (approved):
Both of the problems I had previously identified have been addressed, and I believe we should merge this to give us as much time as possible to identify any other issues before the next release.


I did run into one odd log from the backend. Using Theia as a workspace and navigating around a few files with go-to-definition, I saw this log:

image

That appears to come from some of the `nls` machinery (@msujew) that is a common dependency of the built-in extensions, but I haven't seen it before, so it may be worth checking on.

```typescript
function createScopedLocalizeFunction(messages) {
    return function (key, message) {
        var args = [];
        for (var _i = 2; _i < arguments.length; _i++) {
            args[_i - 2] = arguments[_i];
        }
        if (isNumber(key)) {
            if (key >= messages.length) {
                console.error("Broken localize call found. Index out of bounds. Stacktrace is\n: " + new Error('').stack);
                return;
            }
            return common_1.format(messages[key], args);
        }
        else {
            if (isString(message)) {
                console.warn("Message " + message + " didn't get externalized correctly.");
                return common_1.format(message, args);
            }
            else {
                console.error("Broken localize call found. Stacktrace is\n: " + new Error('').stack);
            }
        }
    };
}
```

colin-grant-work:
> I did run into one odd log from the backend. Using Theia as a workspace and navigating around a few files with go-to-definition, I saw this log:
> 
> image

I saw this later when working on `master` as well, so it likely isn't related to this PR at all.

tortmayr:
@colin-grant-work  Thanks for the fast review. I agree that we should merge this as fast as possible to give us enough time to react to potential bugs and issues.

tsmaeder:
@tortmayr if you want an in-depth code review, of the latest changes, it might take until end of week. I'm fine with merging, though: I'm confident we can find/fix outstanding issues before the next release.

PR #11232, opened 2022-05-31
Use Number.MAX_SAFE_INTEGER in file quick search

colin-grant-work:
#### What it does

Despite #10694, it is still possible to get good matches ranked ahead of perfect matches, because the `MAX_SAFE_INTEGER` constant used was not necessarily the actual maximum safe integer available in the browser, and fuzzy search could still return larger numbers.

#### How to test

1. Tests should pass.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/api-tests/src/file-search.spec.js
+++ b/examples/api-tests/src/file-search.spec.js
@@ -58,8 +58,8 @@ describe('file-search', function () {
             });
 
             it('should not place very good matches above exact matches', () => {
-                const exactMatch = 'aa_bb_cc_dd.file';
-                const veryGoodMatch = 'aa_bb_cc_ed_fd.file';
+                const exactMatch = 'almost_absurdly_long_file_name_with_many_parts.file';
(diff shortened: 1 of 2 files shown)


colin-grant-work:
@msujew, should we just use `Number.MAX_SAFE_INTEGER` everywhere, you think?

msujew (commented):
@colin-grant-work Yes and no. There are two kinds of `MAX_SAFE_INTEGER` cases:

1. We actually want the larges possible integer available in the runtime (2^53 - 1)
2. We want the largest possible integer that monaco accepts (2^32 - 1)

So maybe rename the exported `MAX_SAFE_INTEGER` to `MAX_SAFE_MONACO_INTEGER`? As far as I can tell, the change that you performed is the only instance where we use the exported `MAX_SAFE_INTEGER` variable unintentionally.

vince-fugnitto (approved):
LGTM :+1:

Issue #11233, opened 2022-05-31
download:plugins much slower after merging HTTP proxy support


ccorn:
### Context

After your awesome efforts resulting in commits 8ba808a0008714fdbbc9a596b8408b5c602eff43 (#11112) and 9a10b6b5daab5e046f28fde8803b3612e84f6ff4 (#11201), there was a brief phase during which I could enjoy reliable `download:plugins` functionality with a slow (1Mbps) connection, at least when using the `--no-parallel` option.

As you may remember, when resolving extension packs, a lot of metadata gets downloaded (about 1,3 GB for 73 plugins). Without compression that would take 3 hours over a 1Mbps connection. But apparently the transfer uses compression, which speeds up the transfer by a factor of 6, reducing the time to 30 minutes.

### Issue

Alas, with the master branch at v1.26.0, `download:plugins --no-parallel` keeps a 1Mbps download connection maxed out for 3 hours. Therefore I suspect that compression has been inhibited.

The culprit seems to be

  - d276e5aa1174c0806d5c225afdad4165bca46f4f Support http proxies (#11043)

And no, I do not have any proxy setup.

### How to reproduce

 1. Checkout and build 116a40c10233a404216131283ee4a1a3de751fb4
 2. `rm -rf plugins; time yarn download:plugins --no-parallel`
 3. Checkout and build the successor d276e5aa1174c0806d5c225afdad4165bca46f4f
 4. `rm -rf plugins; time yarn download:plugins --no-parallel`
 5. Note that the second duration is about 6 times longer, consistent with lack of compression.

### Requests

 1. I have not examined the actual API request headers, so further investigation to identify the cause would be prudent.
 2. To catch such unintended side effects in the future, an automated test case for the CI would be helpful. Not about the time taken, but about the presence of compression request headers (or whatever the cause turns out to be).
 3. And, yes, a bug fix would be nice. Particularly because a factor 6 increase in traffic seems [not desirable](https://github.com/eclipse/openvsx/issues/379) for open-vsx.org right now.

ccorn:
> Even if it has been downloaded, it is still very slow to skip it when `theia download:plugins` is executed again.

That is because the metadata for all versions is requested even if the extension has been downloaded already. Ideally, the skip test would occur before the metadata request (and prevent the latter if it turns out to be unnecessary.

ccorn:
[...]

PR #11236, opened 2022-06-01
Allow gzip encoding for request services

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/11233

Allow `gzip` encoding by default for all requests coming from the `NodeRequestService`. Sends the respective `Accept-Encoding` headers and respects the `content-encoding` response header.

#### How to test

1. Confirm (debug) that the headers are present on request and response headers.
2. `download:plugins` cli command should still work as expected (maybe a bit faster)

#### Review checklist

#### Reminder for reviewers


msujew:
cc @ccorn Can you try whether this alleviates the performance regressions you were experiencing?

ccorn:
If `zstd` were available as well, that [could reduce the huge transfer sizes](https://github.com/eclipse/openvsx/issues/379#issuecomment-1125263546) of some responses by a factor of 80 in comparison to `gzip`. Plus, it's faster. So, if the libs provide that feature already, using it would be beneficial. If (and only if) that is a low-hanging fruit, I'd recommend plucking it.

ccorn:
Looking at the code of this PR, it seems that it re-implements request-response analysis. But the OVSX client had compression supported previously. Which leads me to the question: Isn't there a simple pass-through for the no-proxy case available? That would be more robust, not just with respect to compression headers but to all other conceivable specifics, I think.

msujew:
@ccorn Thanks for testing it!

Regarding the `zstd` change: I would be all-in for it, if Node would provide a built-in lib for that (like they do for `gzip`). But as it stands I would wait until ovsx supports it.

In general regarding using a more sophisticated http-library: We could do that, but lose some of the fine-grained nature of using the plain http package. I've seen that `node-fetch` now actually supports proxy agents, so it might worthwhile to investigate. For now, I would stick with the low-level approach.

dev-packages/request/src/node-request-service.ts
```diff
@@ -112,11 +119,17 @@ export class NodeRequestService implements RequestService {
                 } else {
                     const chunks: Uint8Array[] = [];
 
-                    res.on('data', chunk => {
+                    let stream: Stream = res;
+
+                    if (res.headers['content-encoding'] === 'gzip') {
+                        stream = stream.pipe(createGunzip());
+                    }
```
colin-grant-work, line 126: I may be a bit too big a fan of ternaries, but here it seems reasonable, unless we anticipate other cases to handle?
```suggestion
                    const stream = res.headers['content-encoding'] === 'gzip' ? res.pipe(createGunzip()) : res;
```

colin-grant-work (approved):
Code looks reasonable and no problems on my end. It did speed up the plugin download noticeably, even on a pretty fast network.

PR #11243, opened 2022-06-02
Support ThemeIcon's `color` property of the VS Code API

### LINKED ISSUE #11128: [vscode] Support optional property `ThemeIcon#color`
### Feature Description:

Support plugins referring to a `ThemeIcon` with a specific `ThemeColor` via constructor or property `color` of a `ThemeIcon`, see https://code.visualstudio.com/api/references/vscode-api#ThemeIcon

lucas-koehler:
#### What it does
Fixes #11128

* Add optional `color` property to `ThemeIcon` of the VS Code API
* Hand over the `ThemeIcon` instead of just its id from plugin Ext to Main
* Apply the color in the `PluginTree`
* Adapt `TreeViewNode.is` to also recognize sub types such as `CompositeTreeViewNode`

Signed-off-by: Lucas Koehler <lkoehler@eclipsesource.com>

#### How to test

* Download the [test extension](https://github.com/lucas-koehler/vscode-extension-samples/releases/download/theme-icon-color-test-0.0.1/custom-view-samples-0.0.1.vsix). [Source](https://github.com/lucas-koehler/vscode-extension-samples/tree/theme-icon-color-test/tree-view-sample) for reference.
* Launch and open the `TEST VIEW' in the Explorer:
image

* Expand the tree nodes. See that the tree nodes use file icons in the editor's warning color (usually some kind of yellow or orange)
* See that file icons in other views (e.g. the File Explorer) still have their original color
* Switch themes to see that the icons in the Test View use the warning color of the selected theme (select high contrast theme to see a noticeable difference)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/view/plugin-tree-view-node-label-provider.ts
+++ b/packages/plugin-ext/src/main/browser/view/plugin-tree-view-node-label-provider.ts
@@ -33,7 +33,7 @@ export class PluginTreeViewNodeLabelProvider implements LabelProviderContributio
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     canHandle(element: TreeViewNode | any): number {
-        if (TreeNode.is(element) && ('resourceUri' in element || 'themeIconId' in element)) {
+        if (TreeNode.is(element) && ('resourceUri' in element || 'themeIcon' in element)) {
             return this.treeLabelProvider.canHandle(element) + 1;
         }
         return 0;
@@ -43,12 +43,14 @@ export class PluginTreeViewNodeLabelProvider implements LabelProviderContributio
         if (node.icon) {
             return node.icon;
         }
-        if (node.themeIconId) {
-            if (node.themeIconId === 'file' || node.themeIconId === 'folder') {
+        if (node.themeIcon) {
+            if (node.themeIcon.id === 'file' || node.themeIcon.id === 'folder') {
                 const uri = node.resourceUri && new URI(node.resourceUri) || undefined;
-                return this.labelProvider.getIcon(URIIconReference.create(node.themeIconId, uri));
+                if (uri) {
+                    return this.labelProvider.getIcon(URIIconReference.create(node.themeIcon.id, uri));
+                }
             }
-            return ThemeIcon.asClassName({ id: node.themeIconId });
+            return ThemeIcon.asClassName(node.themeIcon);
         }
         if (node.resourceUri) {
             return this.labelProvider.getIcon(new URI(node.resourceUri));
--- a/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
+++ b/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
@@ -16,7 +16,7 @@
 
 import { URI } from '@theia/core/shared/vscode-uri';
 import { injectable, inject, postConstruct } from '@theia/core/shared/inversify';
-import { TreeViewsExt, TreeViewItemCollapsibleState, TreeViewItem, TreeViewSelection } from '../../../common/plugin-api-rpc';
+import { TreeViewsExt, TreeViewItemCollapsibleState, TreeViewItem, TreeViewSelection, ThemeIcon } from '../../../common/plugin-api-rpc';
 import { Command } from '../../../common/plugin-api-rpc-model';
 import {
     TreeNode,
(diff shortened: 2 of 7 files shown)


packages/plugin/src/theia.d.ts
```diff
@@ -2652,7 +2652,7 @@ export module '@theia/plugin' {
          */
         static readonly Folder: ThemeIcon;
 
-        private constructor(public id: string);
+        private constructor(public id: string, public color?: ThemeColor);
```
vince-fugnitto, line 2655: minor: do you mind adding the documentation here?

```suggestion
		/**
		 * Creates a reference to a theme icon.
		 * @param id id of the icon. The available icons are listed in https://code.visualstudio.com/api/references/icons-in-labels#icon-listing.
		 * @param color optional `ThemeColor` for the icon. The color is currently only used in {@link TreeItem}.
		 */
        private constructor(public id: string, public color?: ThemeColor);
```

vince-fugnitto, line 2655: I believe you are missing the optional `color` in `ThemeIcon` (which will still report as missing in the compatibility report):

```ts
/**
 * The optional ThemeColor of the icon. The color is currently only used in {@link TreeItem}.
 */
readonly color?: ThemeColor | undefined;
```

lucas-koehler: Are you sure we need this? There is no explicit property for `id` either and it is recognized in the report. Or is this different for optional properties? In this case I'll add it. I would then also add the `id` prop to make it consistent with the VS Code API which defines both explicitly.

packages/plugin-ext/src/main/browser/view/plugin-tree-view-node-label-provider.ts
```diff
@@ -43,12 +43,14 @@ export class PluginTreeViewNodeLabelProvider implements LabelProviderContributio
         if (node.icon) {
             return node.icon;
         }
-        if (node.themeIconId) {
-            if (node.themeIconId === 'file' || node.themeIconId === 'folder') {
+        if (node.themeIcon) {
+            if (node.themeIcon.id === 'file' || node.themeIcon.id === 'folder') {
```
vince-fugnitto, line 47: I just wanted to highlight that this only allows us a subset of icons to be used which is different than the API actually allows. VS Code allows all icons listed here (https://code.visualstudio.com/api/references/icons-in-labels#icon-listing) which is documented in `id`:

```ts
/**
 * The id of the icon. The available icons are listed in https://code.visualstudio.com/api/references/icons-in-labels#icon-listing.
 */
readonly id: string;
```

With some modifications to the extension provided in VS Code I can see (while in Theia it does not work):


image

vince-fugnitto (commented):
I confirm that the `color` is working well with the sample extension.

I made a comment about actually properly supporting all options for icons however that we can possibly tackle later if you don't want to update.

lucas-koehler:
@vince-fugnitto
>I made a comment about actually properly supporting all options for icons however that we can possibly tackle later if you don't want to update.

I would prefer to have that as a follow up because it does not seem trivial to support all these icons. For instance, I investigated a bit with the example you provided. The `accounts-view-bar-icon` theme icon is contributed in VSCode as part of its `ActivitybarPart`. It basically maps this identifier to the account codicon. I guess this is not even known to Theia (and possibly many other icons made available in such a way in the VS Code workbench).

vince-fugnitto (approved):
The changes look good to me, I confirmed that the color is properly applied :+1: 
I did not notice any regressions.

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
```diff
@@ -60,15 +61,15 @@ export interface TreeViewNode extends SelectableTreeNode {
...
-    themeIconId?: string | 'folder' | 'file';
+    themeIcon?: ThemeIcon;
     tooltip?: string;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     description?: string | boolean | any;
     accessibilityInformation?: AccessibilityInformation;
 }
 export namespace TreeViewNode {
     export function is(arg: TreeNode | undefined): arg is TreeViewNode {
-        return !!arg && SelectableTreeNode.is(arg) && !ExpandableTreeNode.is(arg) && !CompositeTreeNode.is(arg);
```
vince-fugnitto, line 71: @lucas-koehler can you clarify why this change is necessary?

PR #11248, opened 2022-06-02
[VSX] Display plugins counts

alvsan09:
#### What it does

Adds a plugin count badge to the title of each instance of
`VSXExtensionsWidget` except for `SEARCH_RESULT` as the count may be
partial.

The solution extends the functionality under `view-container` by adding a `BadgeWidget` interface that can be
implemented by widgets wrapped in a `ViewContainerPart`, this is the reference implementation for it.
This solution does not show the badge for view parts that contain a toolbar populated with more than one item.

#### How to test

- Add an `extensions.json` file with the following content to your workspace `.theia` folder:
```ts
{
    "recommendations": [
        "ms-azuretools.vscode-docker",
        "exiasr.hadolint",
        "timonwong.shellcheck",
        "coenraads.bracket-pair-colorizer-2",
        "scalameta.metals"
    ]
}
```
- Start `theia` and open the `EXTENSIONS` view
- Install few extensions
- Verify the counter values for each widget i.e. `INSTALLED`, `BUILT-IN`, `RECOMMENDED`
- As per the listed extensions above the expected count for `RECOMMENDED` is **5**
- Make sure other view parts don't display a badge e.g. `WATCH`  under the `DEBUG` view.

vsx_plugin_count gif

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/vsx-registry/src/browser/vsx-extensions-widget.ts
+++ b/packages/vsx-registry/src/browser/vsx-extensions-widget.ts
@@ -19,6 +19,8 @@ import { TreeNode } from '@theia/core/lib/browser';
 import { SourceTreeWidget } from '@theia/core/lib/browser/source-tree';
 import { VSXExtensionsSource, VSXExtensionsSourceOptions } from './vsx-extensions-source';
 import { nls } from '@theia/core/lib/common/nls';
+import { BadgeWidget } from '@theia/core/lib/browser/view-container';
+import { Emitter } from '@theia/core/lib/common';
 
 @injectable()
 export class VSXExtensionsWidgetOptions extends VSXExtensionsSourceOptions {
@@ -28,7 +30,7 @@ export class VSXExtensionsWidgetOptions extends VSXExtensionsSourceOptions {
 export const generateExtensionWidgetId = (widgetId: string): string => VSXExtensionsWidget.ID + ':' + widgetId;
 
 @injectable()
-export class VSXExtensionsWidget extends SourceTreeWidget {
+export class VSXExtensionsWidget extends SourceTreeWidget implements BadgeWidget {
 
     static ID = 'vsx-extensions';
 
@@ -57,13 +59,30 @@ export class VSXExtensionsWidget extends SourceTreeWidget {
         this.addClass('theia-vsx-extensions');
 
         this.id = generateExtensionWidgetId(this.options.id);
+
+        this.toDispose.push(this.extensionsSource);
+        this.source = this.extensionsSource;
+
         const title = this.options.title ?? this.computeTitle();
         this.title.label = title;
         this.title.caption = title;
 
-        this.toDispose.push(this.extensionsSource);
-        this.source = this.extensionsSource;
+        this.toDispose.push(this.source.onDidChange(async () => {
+            this.badge = await this.resolveCount();
+        }));
     }
+
+    private _badge: number | undefined = undefined;
+    get badge(): number | undefined {
+        return this._badge;
+    }
+
+    set badge(count: number | undefined) {
+        this._badge = count;
+        this.onDidChangeBadge.fire();
+    }
+
+    public onDidChangeBadge: Emitter<void> = new Emitter<void>();
 
     protected computeTitle(): string {
         switch (this.options.id) {
@@ -78,6 +97,14 @@ export class VSXExtensionsWidget extends SourceTreeWidget {
             default:
                 return '';
         }
+    }
+
+    protected async resolveCount(): Promise<number | undefined> {
+        if (this.options.id !== VSXExtensionsSourceOptions.SEARCH_RESULT) {
+            const elements = await this.source?.getElements() || [];
+            return [...elements].length;
+        }
+        return undefined;
     }
 
     protected override tapNode(node?: TreeNode): void {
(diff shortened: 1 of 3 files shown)


msujew (commented):
Looks good from a surface viewpoint. Haven't looked into the code yet. I noticed some minor layouting issues, where the badge clips into the focus outline. I believe we need a larger margin there:

grafik

VSCode:

grafik

alvsan09:
> Looks good from a surface viewpoint. Haven't looked into the code yet. I noticed some minor layouting issues, where the badge clips into the focus outline. I believe we need a larger margin there:
> 
> grafik
> 
> VSCode:
> 
> grafik

I also received some comments off-line from @vince-fugnitto , I am in the process to change the code, i.e.
* Give the responsibility to render the badge to the `ViewContainerPart` rather than from the `VSXExtensionsWidget`
* Make the `BadgeWidget` interface more specific to contain a `badge` of type `number | undefined` rather that `node | undefined`
* Make sure that if a part contains a `toolbar` the widget is not displayed.

I will add a right margin to it as you have pointed out as well, 
Additional comments are very welcome !
Thanks @msujew !

packages/core/src/browser/style/notification.css
```diff
@@ -34,6 +34,10 @@
     text-align: center;
 }
 
+.notification-count-container-margin {
```
vince-fugnitto, line 37: We should drop `-margin` from the classname.

vince-fugnitto, line 37: I believe this file is strictly for notifications styling, shouldn't we put our styling under `view-container.css`?
The use of `notifications` for badge styling in `view-container` seems a little odd.

packages/vsx-registry/src/browser/vsx-extensions-widget.ts
```diff
@@ -57,27 +59,60 @@ export class VSXExtensionsWidget extends SourceTreeWidget {
...
             default:
-                return '';
+                label = '';
+        }
+        const title = `${label}`;
+        this.title.label = title;
+        this.title.caption = title;
+
+        this.badge = await this.resolveCount();
+    }
```
vince-fugnitto, line 108: I believe this method is doing too much work and merges two concepts together (which is confusing), the `title` of the view-part and handling the `badge`.

packages/vsx-registry/src/browser/vsx-extensions-widget.ts
```diff
@@ -57,27 +59,60 @@ export class VSXExtensionsWidget extends SourceTreeWidget {
...
+
+    protected async resolveCount(): Promise<number | undefined> {
+        if (this.options.id !== VSXExtensionsSourceOptions.SEARCH_RESULT) {
+            const elements = await this.source?.getElements() || [];
+            return [...elements].length;
         }
+        return undefined;
```
vince-fugnitto, line 115: The changes do not properly handle a `0` count:

image

In this case I would still expect `0` to be displayed for recommended and installed.

alvsan09: I have submitted a new commit to address these comments !!
Thanks @vince-fugnitto !!

packages/vsx-registry/src/browser/vsx-extensions-widget.ts
```diff
@@ -52,32 +54,65 @@ export class VSXExtensionsWidget extends SourceTreeWidget {
     protected readonly extensionsSource: VSXExtensionsSource;
 
     @postConstruct()
-    protected override init(): void {
+    protected override async init(): Promise<void> {
```
paul-marechal, line 57: Don't make the post-construct method async: Inversify 5 doesn't wait for it.

Please use the `Promise.then` API instead.

packages/vsx-registry/src/browser/vsx-extensions-widget.ts
```diff
@@ -57,14 +59,31 @@ export class VSXExtensionsWidget extends SourceTreeWidget {
...
         this.title.caption = title;
 
-        this.toDispose.push(this.extensionsSource);
-        this.source = this.extensionsSource;
+        this.toDispose.push(this.source.onDidChange(async () => {
+            this.badge = await this.resolveCount();
+        }));
+    }
+
+    private _badgeCount: number | undefined = undefined;
```
vince-fugnitto, line 75: minor: we should rename to `_badge` due to the getter and setter, or just use `badgeCount` (dropping the unnecessary `_`). I'm in favor of using `_badge` however.

- https://github.com/eclipse-theia/theia/wiki/Coding-Guidelines#no_underscore_private

vince-fugnitto (commented):
I confirm that the changes work as expected:

- [x] a badge of `0` is supported and displayed
- [x] the 'builtin', 'recommended' and 'installed' tree-view-parts have badges
- [x] the badges correctly display the extension count for each part
- [x] the badges are correctly updated when the count is updated

packages/vsx-registry/src/browser/vsx-extensions-widget.ts
alvsan09: I agree with you, `_badge` is more appropriate, 
It has been updated in the latest commit, 
I have also squashed the commits and re-based it to later master.

Thanks @vince-fugnitto !

vince-fugnitto (approved):
LGTM :+1:

alvsan09:
@msujew, @vince-fugnitto, Thanks !! :+1:

PR #10965, opened 2022-03-30
Ensure that old values aren't cached inappropriately

colin-grant-work:
#### What it does

Fixes https://github.com/eclipse-theia/theia/issues/10959 
Fixes https://github.com/eclipse-theia/theia/issues/10865

Previously, the caching mechanism in `ValidatedPreferenceProxy` could mistakenly retain cached values for language overrides. This PR ensures that the cache is maintained only once the proxy subscribes to preference events and that it is updated for all overrides in case of a preference change.

#### How to test

1. Activate and deactivate the `editor.minimap.enabled` preference. It should immediately take effect in open editors (in all languages) and apply correctly to new editors.
2. Assert that the new tests pass.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #11257, opened 2022-06-03
Add validations to Monaco

### LINKED ISSUE #11241: Editor preferences ignoring scope precedence
### Bug Description:

Editor preferences are ignoring the precedence of preference scopes and accepting the most recent change to preference values.

### Steps to Reproduce:

1. Open a workspace and open a text file in that workspace.
2. In that workspace, set e.g. `editor.fontSize` to some non-default value.
3. Observe that the new setting takes effect in your open file.
4. In user scope, set the same preference to some other non-default value.
5. Observe that that new setting also takes effect in your open file, even though it should be ignored because it is overridden by the  workspace value.

### Additional Information

- Theia Version: 5f887e22825c7284b16104b517678ca0a2824319

colin-grant-work:
#### What it does

Fixes #11241 by removing the buggy `ValidatedPreferenceProxy` implementation in favor of adding validations for a few key preferences to Monaco's own validation system for editor preferences.

#### How to test

1. Open a workspace and open a text file in that workspace.
2. In that workspace, set e.g. `editor.fontSize` to some non-default value.
3. Observe that the new setting takes effect in your open file.
4. In user scope, set the same preference to some other non-default value.
5. Observe that that new setting does _not_ take effect in your open file because it is overridden in a higher-priority scope.

---

1. Also confirm that the testing steps for #10607 and #10965 still go as expected.

#### Review checklist

#### Reminder for reviewers


colin-grant-work:
> * the browser tests fail on CI

I'll check into this.

> * there is no validation message when setting an invalid preference (even with [#11025](https://github.com/eclipse-theia/theia/pull/11025) I understood that it should happen for manual updates by the user)

This is expected: Monaco doesn't provide any indication that the user has done something wrong; it just uses a different value. With this code, the `PreferenceValidatioService`, which issued those warnings, is no longer in use, but I haven't removed it because I think it could be a useful way to provide feedback in the preference UI, with tweaks to provide detailed information. Have to compare it with just using AJV, though.

msujew (approved):
The changes are looking good to me 👍 

- [x] Preferences scopes seem to use the correct precedence now
- [x] Monaco preferences (minimap, rendering whitespace) continues to work correctly
- [x] Setting a preference manually to an invalid value shows a warning and the default value is applied

PR #11259, opened 2022-06-06
Replace non-breaking with normal spaces

colin-grant-work:
#### What it does

This PR replaces a number of instances of non-breaking spaces (U+00A0) with normal spaces (U+0020). VSCode now highlights characters that look like but are not identical to more common characters, and non-breaking spaces fall into that category. This just cleans them up so they don't produce visual distraction.

#### How to test

1. No effective changes - CI should pass.
2. Search for non-breaking spaces (`" "` <-- you can copy that).
3. There should only be two in `packages/core/src/common/keyboard/layouts/fr-Canadian_French-mac.json`, where it looked like they might be intentional.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-manager/src/generator/frontend-generator.ts
+++ b/dev-packages/application-manager/src/generator/frontend-generator.ts
@@ -122,7 +122,7 @@ module.exports = preloader.preload().then(() => {
     }
 
     function start() {
-        (window['theia'] = window['theia'] || {}).container = container;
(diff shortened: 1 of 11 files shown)


vince-fugnitto (approved):
I confirm the changes update all occurrences of `U+00A0`.
The CI failures are due to open-vsx being down.

Issue #11249, opened 2022-06-02
Application freeze at startup with many errors in RPC system


colin-grant-work:
### Bug Description:

Starting the application in the browser, I got stuck at the loading screen for 2-3 minutes with numerous errors logged in various parts of the new IPC system. (@tortmayr, @JonasHelming)

### Steps to Reproduce:

1. Introduce a `console.log` that produces numerous rapid logs (ca. 100 or so in a tick), possibly involving non-stringifiable types (`Map`, e.g.).

### Additional Information

![image](https://user-images.githubusercontent.com/62660806/171732124-4f487e62-bbd9-41d8-bb76-68a2e70e3231.png)

![image](https://user-images.githubusercontent.com/62660806/171732193-6e063deb-0f1d-4de9-b4e9-4426e3b11f00.png)

![image](https://user-images.githubusercontent.com/62660806/171732248-92af908e-3b71-4663-bf71-a3b56b2be9dc.png)

- Operating System: MacOS, Chrome browser
- Theia Version: 9b1f414a77884c2ab120af26f7232e0541a60612

colin-grant-work:
Likely a symptom of similar issues:

1. Add this log to `DebugSessionManager` `set currentSession`:
```typescript
console.log('SENTINEL FOR SETTING THE CURRENT SESSION', current);
```
2. Start a debug session inside the application.
3. Observe that the frontend (browser) logs log the message as expected.
4. Observe that the backend logs do _not_ show the message and the front end logs show this error:

![image](https://user-images.githubusercontent.com/62660806/172253953-607bf387-9c12-487d-9258-b6412022776f.png)

colin-grant-work:
@tortmayr, any plans to tackle this one? It seems that logging almost any object produces the `Cannot read properties of undefined (reading 'length')`, and it isn't hard to get the maximum stack frame error for frequent logs with complex arguments (e.g. an `isVisible` check for a command where a widget will be the argument.)

tortmayr:
@colin-grant-work  I'm looking into this right now. At first glance it looks like we have to tackle two issues here:

[...]

Issue #11159, opened 2022-05-16
Improve new message-rpc API


tortmayr:
With https://github.com/eclipse-theia/theia/pull/11011 a new message rpc protocl will be integrated into Theia's core messaging API.
During the review a couple of issues and potential improvments where identified.
 1. Channels are directly receiving and sending binary buffers. This rather low-level API is not convenient for adopters to interact with.
 Ideally we can make the binary encoding/decoding an implementation detail of the channel implementation and the Channel itself only sends/receives messages. It has to be investigated weather this is feasible especially with the Plugin RPC API in mind. A solution is required that still allows to directly forwards received messages in the `hosted-plugin-process` and `plugin-host` process without having to decode the entire messages. (See also https://github.com/eclipse-theia/theia/pull/11011#issuecomment-1119333821)
2. Currently a custom binary message codec is in place. Maybe we can offload the encoding/decoding to a specialized library (e.g. [msgpack](https://msgpack.org/index.html). It has to be evaluated how performant these third party libraries are compared to our custom solution. (see also https://github.com/eclipse-theia/theia/pull/11011#issuecomment-1121270230)
- Nice-to-have feature: Serialization of circular structures ( see also https://github.com/eclipse-theia/theia/issues/11249)

We (EclipseSource) plan to contribute this on behalf of STMicroelectronics

tortmayr:
Point 1 (object-based-channel) is addressed with this PR: https://github.com/eclipsesource/theia/commit/042c40ad96bc95661e04a157eaccb99be4a1fec6)

tortmayr:
For Point 2 I also did a little evaluation/benchmarking of different msgpack implementations:

## Introduction

Currently we are using a custom binary codec for encoding RPC messages that are sent between the Theia frontend and backend processes.Compared to plain JSON-RPC this codec offers efficient encoding of binary data (buffers). However, it turns out the performance for large (buffer-free) objects is considerably worse with the new message codec.
`Msgpack` is a standard protocol for encoding objects into a binary format and should be more performant than our custom codec. In addition, building ontop of a well established standard probably makes more sense than trying to re-implement the wheel. 
In order to switch to a `msgpack` library the following requirements need to be fulfilled:
- The msgpack implementation needs be quicker (or at least en par with) than the current custom binary codec
- The msgpack implementation needs to support an extension mechanism that allows us to register custom value encoders/decoders
- The project should be active and we need to trust the maintainer
- Compatible OS License

Nice to have features:
[...]

PR #11261, opened 2022-06-06
Use binary message RPC protocol for plugin API

### LINKED ISSUE #9514: Performance of vscode.workspace.fs.readDirectory(Uri) is worse by *100 compared to vscode

tortmayr:
#### What it does
Refactors the plugin RPC protocol to make use of the new message-rpc introduced with #11011/#11228.
- Refactor plugin-ext RpcProtocol API to reuse the new message-rpc protocol
  - Remove custom RPC message encoding and handling reuse message-rpc
  - Implement `QueuingChannelMultiplexer` that queues messages and sends them accumulated on the next process.tick (replaces the old Multiplexer 
    implementation)
   - Refactors proxy handlers and remote target handlers
   - Use `Channel` instead of `MessageConnection` for creating new instances of RPCProtocol
   - Refactor `RpcMessageEncoder`/`RpcMessageDecoder` to enable overwritting of already registered value encoders/decoders. 
   - Add mode property to  base `RpcProtocol` to enable switching from a bidirectional RPC protocol to a client-only or server-only variant.
- Implement special message encoders and decoders for the plugin communication. (Replacement for the old `ObjectTransferrer` JSON replacers/revivers)
- Adapt `HostedPluginServer` and `HostedPluginClient` API to send/receive messages in binary format instead of strings. This enables direct writethrough of the binary messages received from the hosted plugin process.
- Adapt `hosted-plugin-process` and `plugin-host` to directly send binary messages via  `IpcChannel`/`BinaryMessagePipe`

- Remove incorrect (and unused) notification proxy identifiers and instantiation
  - NotificationExt was instantiated in the main context
  - There were unused notification proxy identifiers for main and ext in the wrong contexts

Part of #10684
Fixes #9514

Co-authored-by: Lucas Koehler <lkoehler@eclipsesource.com>

#### How to test
For regression testing, any plugin loaded in the example applications can be used. Everything (i.e. interacting with and using the plugin should work as before.

For performance testing see comment below

#### Review checklist

#### Reminder for reviewers


tsmaeder:
Every once in a while, completion just seems to either fail or take a long time. I can't repeat nor do I have a handle on what's wrong yet.

tortmayr:
Thanks for testing this change @tsmaeder.

>I've done a little testing, and I noticed that I get a ton of cancellation error stack traces in the console
>
>Details
Monaco silently accepts cancellation if it detects that it is a cancellation. I think the proper layering would be to catch our own cancellation error and to translate it into a VS Code Cancellation Error at some appropriate place. see https://github.com/microsoft/vscode/blob/f52952332d14de7517cb22b002ef1352ecdee889/src/vs/base/common/errors.ts#L149

Good catch! I double checked with the "old" plugin rpc protocol implementation and it indeed throws a VS Code Cancellation Error whereas the new protocol didn't: https://github.com/eclipse-theia/theia/blob/53a272e113890174c8b3049b8d1a74999dee0cd4/packages/plugin-ext/src/common/rpc-protocol.ts#L166

I have adapted the new protocol accordingly and we now should throw the expected Cancellation Error. In addition, I reverted another "over-optimiatization" that prematurely rejected the response of  a canceled request.

I did a little testing and did not encounter cancellation error stack traces anymore.

>Every once in a while, completion just seems to either fail or take a long time. I can't repeat nor do I have a handle on what's wrong yet.
>
>Turns out, the completion doesn't really fail, but it kind of looks like it's waiting for something: when I just let the cursor sit there, the completion does not complete. As soon as I move the cursor around, I get the completion.

I could not observe this behavior anymore. Maybe it got fixed with the cancellation error fixes? Could you please retest this? Thanks!

packages/core/src/common/message-rpc/rpc-protocol.ts
```diff
@@ -145,21 +144,20 @@ export class RpcProtocol {
...
         }
 
+        const id = this.nextMessageId++;
+        const reply = new Deferred<T>();
+
         if (cancellationToken) {
             args.push(RpcProtocol.CANCELLATION_TOKEN_KEY);
             cancellationToken.onCancellationRequested(() => {
                 this.sendCancel(id);
-                this.pendingRequests.get(id)?.reject(this.cancelError());
```
tsmaeder, line 143: So the idea is that the "other side" will cancel the call, and send a "replyCancel"?

packages/core/src/common/message-rpc/rpc-protocol.ts
```diff
@@ -191,8 +189,8 @@ export class RpcProtocol {
     }
 
     cancelError(): Error {
-        const error = new Error('"Request has already been canceled by the sender"');
-        error.name = 'Cancel';
+        const error = new Error('Canceled');
```
tsmaeder, line 192: If we rely on these values to be the same as in VS Code, we really should have a comment explaining the fact and pointing to the VS Code code. Otherwise someone _will_ "fix" the message at some point.

tortmayr: Actually I just discovered that a corresponding `CancellationError` is already available (introduced with https://github.com/eclipse-theia/theia/pull/10035) so I removed the custom error handling and we throw a new `CancellationError` instead.

tsmaeder:
There _is_ a problem with the completion in the PR: I've made two videos, first the behavior in master:

https://youtu.be/53T4Hvcw1rM

and now in the PR

https://youtu.be/sj8GVgzBC2k

Both videos are mad in master, on the source code of Theia and with the browser example, using Firefox. Completion seems slower and erratic in timing: sometimes leaving the cursor where you clicked before completion seems to block the completion somehow, but eventually the completion appears. To me it feels as if we were "crossing the streams". Is it possible that cancellations and multiple requests might be blocking or interfering with each other? If I look at what's changed with respect to master, what I'm seeing is that we're using a different multiplexing strategy for routing to multiple services and message batching has changed, as well. I think it would be good to have some configurable tracing in the code to find out what's going on. We just need to make sure we're not sending it over the wire with the remote log infrastructure again ;-)

packages/core/src/common/message-rpc/rpc-protocol.ts
tortmayr: To my understanding the idea here is that the request sender (client) just informs he request handler (server) that it has cancelled the request and is no longer interested in the result. How the server reacts to this cancel notification is up to the actual method implementation. The implementation can either 
- Not listen to the cancellation notification at all and just complete normally
- React to the cancel notification and throw an error (Ideally but not necessarily  a `CancellationError`). 

At least that's the way it was implemented before this PR :wink:

tortmayr:
Some of this issues might be related to the behavior we have encountered in https://github.com/eclipse-theia/theia/issues/11249.
Also the potential improvements discussed in https://github.com/eclipse-theia/theia/issues/11159.
I will work on these two issues and then continue with cleaning up this PR.

paul-marechal:
I wanted to try out this PR to run benchmarks, but plugins don't seem to work anymore. I am running Windows 10.

tsmaeder:
@paul-marechal how did you resolve the conflicts? @tortmayr could you, please? ;-)

tortmayr:
Sure I can resolve the conflicts and retest it under Windows.

paul-marechal:
@tsmaeder I just pulled the branch as is and built it, I didn't think I had to rebase myself?

packages/plugin-ext/src/hosted/node/hosted-plugin-process.ts
```diff
@@ -156,9 +159,11 @@ export class HostedPluginProcess implements ServerPluginRunner {
             logger: this.logger,
             args: []
         });
-        this.childProcess.on('message', message => {
+
+        this.messagePipe = new BinaryMessagePipe(this.childProcess.stdio[4] as Duplex);
```
paul-marechal, line 163: fyi: The `BinaryMessagePipe` class currently uses a `net.Socket` to wrap the file descriptor, but this doesn't seem to work properly on Windows and leads to communication issues.

packages/plugin-ext/src/hosted/node/hosted-plugin-process.ts
```diff
@@ -184,7 +189,7 @@ export class HostedPluginProcess implements ServerPluginRunner {
             silent: true,
             env: env,
             execArgv: [],
-            stdio: ['pipe', 'pipe', 'pipe', 'ipc']
+            stdio: ['pipe', 'pipe', 'pipe', 'ipc', 'pipe']
```
paul-marechal, line 192: For this change to have a chance to work on Windows:

```suggestion
            stdio: ['pipe', 'pipe', 'pipe', 'ipc', 'overlapped']
```

lucas-koehler:
@paul-marechal Because @tortmayr is on vacation until the end of this week, I committed your suggestion. Thanks for that :)

tortmayr:
@tsmaeder  I have rebased the PR and adapted to the latest changes (msgpack).

> There _is_ a problem with the completion in the PR: I've made two videos, first the behavior in master:
> 
> https://youtu.be/53T4Hvcw1rM
> 
> and now in the PR
> 
> https://youtu.be/sj8GVgzBC2k
> 
> Both videos are mad in master, on the source code of Theia and with the browser example, using Firefox. Completion seems slower and erratic in timing: sometimes leaving the cursor where you clicked before completion seems to block the completion somehow, but eventually the completion appears. To me it feels as if we were "crossing the streams". Is it possible that cancellations and multiple requests might be blocking or interfering with each other? If I look at what's changed with respect to master, what I'm seeing is that we're using a different multiplexing strategy for routing to multiple services and message batching has changed, as well. I think it would be good to have some configurable tracing in the code to find out what's going on. We just need to make sure we're not sending it over the wire with the remote log infrastructure again ;-)

The recent change to msgpack and other bugfixes seem to have resolved this issue. I can no longer reproduce it


@paul-marechal This PR now also includes your fix to ensure that the BinaryMessagePipe is working when using Windows.

packages/plugin-ext/src/hosted/browser/hosted-plugin.ts
```diff
@@ -534,18 +535,37 @@ export class HostedPluginSupport {
...
+                const writer = new Uint8ArrayWriteBuffer();
+                writer.onCommit(buffer => {
+                    this.server.onMessage(pluginHostId, buffer);
+                });
+                return writer;
+            },
+            onClose: onCloseEmitter.event,
+            onError: onErrorEmitter.event,
+            onMessage: onMessageEmitter.event
+        });
```
kittaakos, line 560: This could deserve a dedicated function. What do you think?

The same logic is also in `packages/plugin-ext/src/hosted/browser/plugin-worker.ts` and `packages/plugin-ext/src/hosted/browser/worker/worker-main.ts`.

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
```diff
@@ -25,22 +27,41 @@ export class PluginWorker {
...
+        });
+
+        // eslint-disable-next-line arrow-body-style
+        this.worker.onmessage = buffer => onMessageEmitter.fire(() => {
+            return new Uint8ArrayReadBuffer(buffer.data);
         });
+
+        this.worker.onerror = e => onErrorEmitter.fire(e);
+
+        this.worker.onerror = e => console.error(e);
```
kittaakos, line 64: One of them is probably not needed.

packages/plugin-ext/src/common/rpc-protocol.ts
```diff
@@ -131,416 +117,154 @@ export class RPCProtocolImpl implements RPCProtocol {
...
-
-    private messagesToSend: string[];
-
-    private readonly messageEmitter = new Emitter<string>();
-    private readonly toDispose = new DisposableCollection();
-
-    constructor(connection: MessageConnection) {
-        this.connection = connection;
-        this.sendAccumulatedBound = this.sendAccumulated.bind(this);
+export class QueuingChannelMultiplexer extends ChannelMultiplexer {
```
tsmaeder, line 161: I really feel we should keep the concerns of "batching messages" and "multiplexing channels" separate. The old `RPCMultiplexer` name was always wrong, since it never did multiplex, it batched.

tsmaeder: So I would really have a `BatchingChannel` that takes care of batching and use a stock `ChannelMultiplexer` on top of that.

tsmaeder (commented):
Mostly questions.

tsmaeder:
In testing I get the following error message in the log (with electron)
```
2
node.cjs:963 Uncaught (in promise) Error: unknown document
    at currentExtensions.<computed> (node.cjs:963)
    at read (node.cjs:403)
    at Array.readObject [as read] (node.cjs:496)
    at recordDefinition (node.cjs:956)
    at read (node.cjs:392)
    at checkedRead (node.cjs:193)
    at Packr.unpack (node.cjs:103)
    at Packr.decode (node.cjs:175)
    at MsgPackMessageDecoder.decode (rpc-message-encoder.ts:162)
    at MsgPackMessageDecoder.parse (rpc-message-encoder.ts:166)
```

packages/core/src/common/message-rpc/channel.ts
```diff
@@ -263,10 +267,14 @@ export class ChannelMultiplexer implements Disposable {
     }
 
     open(id: string): Promise<Channel> {
+        const existingChannel = this.getOpenChannel(id);
```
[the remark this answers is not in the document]
tortmayr: Yes I think you are right. I think this is a remainder of a previous implementation iteration where both  sides (proxy & invocation handler) could potentially open a channel. This is now been reworked and channels are always opened from the proxy side. I have to test it but I think throwing an error should work and is probably the better option here.

packages/core/src/common/message-rpc/msg-pack-extension-manager.ts
```diff
@@ -0,0 +1,87 @@
...
+import { ResponseError } from './rpc-message-encoder';
+
+/**
+ * Handles the global registration of custom MsgPackR extensions
+ * required for the default RPC communication. MsgPackR extensions
+ * are installed globally on both ends of the communication channel.
+ * (frontend-backend, pluginExt-pluginMain).
+ * Is implemented as singleton as it is  also used in plugin child processes which have no access to inversify.
+ */
+export class MsgPackExtensionManager {
```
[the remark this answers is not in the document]
tortmayr: I agree, the global extension mechanism is not ideal but sufficient for now. Maybe we should raise an issue for that in the msgpackR Repo. Development there seems to pretty active and responsive to community feedback. So we might be able to get a better extension mechanism with future versions.

Here is my reasoning for introducing this class:
- Better error handling. MsgpackR simply overrides previous registered extensions if you add a new one with the same tag. This might make error tracing really difficult (which is already complicated enough in the rpc context). With the dedicated manager we can throw errors for extensions with duplicate or invalid (out-of-range) tags.
- MsgPackR provides no possibility to retrieve already registered extensions.
- Adding and extension would require an direct import from "msgPackR". The dependency is currently not exported via @theia/core/shared so this opens up the possibility for version resolution issues. All we need is basically one function and one type definition, so wrapping and reexporting this in Theia core via `MsgPackExtensionManager` might be cleaner.
- Adopters don't have to interact directly with a third party dependency
- The msgpackR source (especially the typings) is not always well documented. With this approach we can keep documentation for the extension mechanism in repo.
However, I have no strong opinion on that and I'm also fine to use msgPackR directly.

packages/core/src/common/message-rpc/rpc-protocol.ts
```diff
@@ -68,37 +73,50 @@ export class RpcProtocol {
...
+    constructor(public readonly channel: Channel, public readonly requestHandler: RequestHandler | undefined, options: RpcProtocolOptions = {}) {
         this.encoder = options.encoder ?? new MsgPackMessageEncoder();
         this.decoder = options.decoder ?? new MsgPackMessageDecoder();
         this.toDispose.push(this.onNotificationEmitter);
-        this.toDispose.push(channel.onMessage(readBuffer => this.handleMessage(this.decoder.parse(readBuffer()))));
         channel.onClose(() => this.toDispose.dispose());
+        this.toDispose.push(channel.onMessage(readBuffer => this.handleMessage(this.decoder.parse(readBuffer()))));
+        this.mode = options.mode ?? 'default';
+
+        if (this.mode !== 'clientOnly' && requestHandler === undefined) {
```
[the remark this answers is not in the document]
tortmayr: Good point. I will refactor this

packages/core/src/common/message-rpc/rpc-protocol.ts
```diff
@@ -185,14 +209,14 @@ export class RpcProtocol {
         }
 
         try {
-            const result = await this.requestHandler(method, args);
+            const result = await this.requestHandler!(method, args);
             this.cancellationTokenSources.delete(id);
             this.encoder.replyOK(output, id, result);
             output.commit();
```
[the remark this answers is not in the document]
tortmayr: I'm assuming that this comment is referring to the explicit dispose() in the catch clause, right?

Yes the dispose() is implicit in a commit(). WriteBuffers are for one-time use and once committed they cannot be reused.
However, we have to consider a special case here. Assuming an error happens in `this.encoder.replyOK()`. Then we never actually commit the buffer but still have to make sure that is disposed and not reused for sending the reply error.

packages/plugin-ext/src/common/proxy-handler.ts
```diff
@@ -0,0 +1,116 @@
...
+ * A proxy handler that will send any method invocation on the proxied object
+ * as a rcp protocol message over a channel.
+ */
+export class ClientProxyHandler<T extends object> implements ProxyHandler<T> {
+    private rpcDeferred: Deferred<RpcProtocol> = new Deferred();
+    private isRpcInitialized = false;
+
+    constructor(protected readonly id: string, protected readonly codec: RpcMessageCodec, protected readonly channelProvider: () => MaybePromise<Channel>) { }
+
+    private async initializeRpc(): Promise<void> {
```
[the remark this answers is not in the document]
tortmayr: We have to handle the `ChannelProvider` promise inside of the `initializeRpc` method. I personally prefer the async-await syntax but if you want I can also refactor this to the classical promise.then syntax.

packages/plugin-ext/src/common/proxy-handler.ts
```diff
@@ -0,0 +1,116 @@
...
+        const clientOptions: RpcProtocolOptions = { ...this.codec, mode: 'clientOnly' };
+        const channel = await this.channelProvider();
+        const rpc = new RpcProtocol(channel, undefined, clientOptions);
+        this.rpcDeferred.resolve(rpc);
+        this.isRpcInitialized = true;
+    }
+
+    get(target: any, name: string, receiver: any): any {
+        if (!this.isRpcInitialized) {
```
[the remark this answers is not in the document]
tortmayr: Yes, starting the initialization from the constructor might result in timing issues.The `HostedPluginWatcher` might receive the first batch for backend messages to early. i.e. before the `HostedPluginSupport` as subscribed to the `onPostMessageEvent`. So the first message batch might get lost in the ether and plugin initialization fails.
By delaying the initialization until the first request this timing issues are resolved.

packages/plugin-ext/src/common/rpc-protocol.ts
tortmayr: Yes you are absolutely right. A batching channel is definitely the cleaner solution.

packages/core/src/common/message-rpc/rpc-protocol.ts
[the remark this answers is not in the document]
tortmayr: On second thought: Without a flag it is not possible to distinguish between the default bi-directional mode and serverOnly mode. Because the requestHadler has to be present for both cases.

packages/plugin-ext/src/common/proxy-handler.ts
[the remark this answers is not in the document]
tsmaeder: My thinking is that if the async/await syntax is not able to express the intent of the API, we should not use it, but it's a bit of a philosophical argument ;-)

packages/plugin-ext/src/common/rpc-protocol.ts
tortmayr: Fixed with https://github.com/eclipse-theia/theia/commit/2b4016e8da50466f271bce2b2f808bf0ed371a10

packages/plugin-ext/src/common/proxy-handler.ts
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        if (target[name] || name.charCodeAt(0) !== 36 /* CharCode.DollarSign */) {
+            return target[name];
+        }
+        const isNotify = this.isNotification(name);
+        return (...args: any[]) => {
+            const method = name.toString();
+            return this.rpcDeferred.promise.then((connection: RpcProtocol) =>
+                new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
tortmayr: Fixed with https://github.com/eclipse-theia/theia/commit/2b4016e8da50466f271bce2b2f808bf0ed371a10

packages/plugin-ext/src/common/proxy-handler.ts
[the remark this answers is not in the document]
tortmayr: Fixed with https://github.com/eclipse-theia/theia/commit/2b4016e8da50466f271bce2b2f808bf0ed371a10

packages/core/src/common/message-rpc/channel.ts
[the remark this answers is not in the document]
tortmayr: Fixed with https://github.com/eclipse-theia/theia/commit/2b4016e8da50466f271bce2b2f808bf0ed371a10

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
tortmayr: Fixed with https://github.com/eclipse-theia/theia/commit/2b4016e8da50466f271bce2b2f808bf0ed371a10

packages/plugin-ext/src/hosted/browser/hosted-plugin.ts
tortmayr: Introduced a new "BasicChannel" class with  https://github.com/eclipse-theia/theia/commit/2b4016e8da50466f271bce2b2f808bf0ed371a10 to reduce the need for code duplication.

tortmayr:
@tsmaeder Thanks for your fast review. I have addressed most of your comments with https://github.com/eclipse-theia/theia/commit/2b4016e8da50466f271bce2b2f808bf0ed371a10.

I'm now starting to investigate how to fix the confusing mgspackR error output

packages/core/src/common/message-rpc/msg-pack-extension-manager.ts
```diff
@@ -0,0 +1,87 @@
...
+    private static readonly INSTANCE = new MsgPackExtensionManager();
+    public static getInstance(): MsgPackExtensionManager {
+        return this.INSTANCE;
+    }
```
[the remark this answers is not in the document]
tortmayr: >First I think singleton instances/APIs should be declared outside of the class describing the singleton: Where we can use Inversify then the singleton instance can be constructed there. For runtimes that aren't using Inversify, then a singleton instance can be constructed/defined specifically for them.

Sure we can do that. However, I just wanted to notice that the  current architecture of the core Theia messaging proxies doesn't really rely on inversify. Everything starting from the ProxyFactory, over to the RPCProtocol and the underlying channels is directly constructed. So even though we have access to inversify we probably end up with constructing a singleton instance.

>Second, with the way things are done here, the extension for ResponseError will only be registered if this script happens to be loaded/imported somehow. As I don't see anyone importing this script yet, I'm quite curious to understand how this can be working?

Good catch, its probably just workig because its indirectly imported in the index.ts of the plugin-ext package. I will have a look at this.

>Overall I don't think we should rely on the global packer/unpacker but rather have components depending on those to have it injected, with or without Inversify. Customization should then be done before injection.

In fact we don't rely on the global packr/unpackr and [already construct](https://github.com/eclipse-theia/theia/blob/d5d67db470a88738fdac9593fd6e0393a206b2c8/packages/core/src/common/message-rpc/rpc-message-encoder.ts#L127) dedicated packr/unpackr isntances. However the extension mechanism in msgpackR still only allows global registration. So you cannot configure extensions per packr/unpackR instance at the moment.

tsmaeder:
@tortmayr I didn't see anything about the comments in https://github.com/eclipse-theia/theia/pull/11261#issuecomment-1303491276. What's the story there?

packages/core/src/common/message-rpc/msg-pack-extension-manager.ts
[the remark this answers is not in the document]
tortmayr: @paul-marechal Are you fine with the way this extension mechanism is currently implemented (even though it is not ideal) or do you have another solution in mind?

tortmayr:
> @tortmayr I didn't see anything about the comments in [#11261 (comment)](https://github.com/eclipse-theia/theia/pull/11261#issuecomment-1303491276). What's the story there?

Ah sorry, I forgot to add a comment regarding this. We need msgpackR extension for error encoding to keep the original stacktrace information (i.e. the pendant to https://github.com/eclipse-theia/theia/blob/32deea205d784831d385825e7e68f02f43cd2b7a/packages/plugin-ext/src/common/rpc-protocol.ts#L519 for the new RPC protocol)
I'm going to add this once the discussion regarding the msgpackR extension mechanism is resolved (https://github.com/eclipse-theia/theia/pull/11261#discussion_r1023214572)

tortmayr:
@tsmaeder I pushed an update that improves the error encoding via msgpack and preserves the stack trace.
Error output from plugins should now look as expected again.

e.g. trying to execute a non-existing command in a plugin would previously result in this message:
```console
2022-11-29T08:18:18.440Z root ERROR [hosted-plugin: 449972] Promise rejection not handled in one second: Error: Command with id 'non.existing.command' is not registered. , reason: Error: Command with id 'non.existing.command' is not registered.
2022-11-29T08:18:18.444Z root ERROR [hosted-plugin: 449972] With stack trace: Error: Command with id 'non.existing.command' is not registered.
    at /home/tobias/Git/OpenSource/theia/theia/node_modules/msgpackr/dist/node.cjs:971:39
    at read (/home/tobias/Git/OpenSource/theia/theia/node_modules/msgpackr/dist/node.cjs:411:15)
    at Array.readObject [as read] (/home/tobias/Git/OpenSource/theia/theia/node_modules/msgpackr/dist/node.cjs:504:18)
    at recordDefinition (/home/tobias/Git/OpenSource/theia/theia/node_modules/msgpackr/dist/node.cjs:964:19)
    at read (/home/tobias/Git/OpenSource/theia/theia/node_modules/msgpackr/dist/node.cjs:400:13)
    at checkedRead (/home/tobias/Git/OpenSource/theia/theia/node_modules/msgpackr/dist/node.cjs:195:13)
    at Packr.unpack (/home/tobias/Git/OpenSource/theia/theia/node_modules/msgpackr/dist/node.cjs:105:12)
    at Packr.decode (/home/tobias/Git/OpenSource/theia/theia/node_modules/msgpackr/dist/node.cjs:177:15)
    at MsgPackMessageDecoder.decode (/home/tobias/Git/OpenSource/theia/theia/packages/core/lib/common/message-rpc/rpc-message-encoder.js:84:29)
    at MsgPackMessageDecoder.parse (/home/tobias/Git/OpenSource/theia/theia/packages/core/lib/common/message-rpc/rpc-message-encoder.js:87:21)
 ```

and now correctly preserves the stacktrace of the original error:

```console
2022-11-29T08:23:58.196Z root ERROR [hosted-plugin: 455858] Promise rejection not handled in one second: Error: Command with id 'non.existing.command' is not registered. , reason: Error: Command with id 'non.existing.command' is not registered.
2022-11-29T08:23:58.197Z root ERROR [hosted-plugin: 455858] With stack trace: Error: Command with id 'non.existing.command' is not registered.
    at CommandRegistryMainImpl.$executeCommand (http://localhost:3000/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:1436:19)
    at http://localhost:3000/packages_plugin-ext_lib_common_plugin-api-rpc_js.js:955:71
    at async RpcProtocol.handleRequest (http://localhost:3000/bundle.js:142581:28)
 ```

tsmaeder:
@tortmayr just a hint: I consider it good practice to keep updates to PR's in separate commits until ready to merge: it makes it possible just review the delta.

tortmayr:
> @tortmayr just a hint: I consider it good practice to keep updates to PR's in separate commits until ready to merge: it makes it possible just review the delta.

Noted, thanks. Apparently I have interacted with too many gerrit based systems recently and developed the bad habit of rebasing/force pushing all changes into a single commit 😉

tsmaeder:
> Apparently I have interacted with too many gerrit based systems recently

It's a bit early for a flamewar ;-)

tsmaeder (approved):
Some more testing did not reveals anything untoward for me. I'm fine with the code as it stands now.

tsmaeder:
@paul-marechal it looks to me your like your comments have been addressed and we're ready to merge this one?

tortmayr:
For anyone interested I did some benchmarking for plugins and frontend plugins to compare the file reading performance starting from the 1.26 Release:

## Benchmarks

Test scenario: Read a file of size x from the workspace using the FileSystem plugin API.
Median of 50 test runs.

- Theia 1.26.0 -> Old json-rpc based protocol
- Theia 1.32.0 -> Current release
- PR -> latest state of this PR branch

## Browser

| Test context             | 100K       | 1000K     | 5000K       | 10000K     |
| ------------------------ | ---------- | --------- | ----------- | ---------- |
| Plugin (1.26.0)          | 49 ms      | 347 ms    | 1964 ms     | 4090 ms    |
| Plugin (1.32.0)          | 27 ms      | 165 ms    | 940 ms      | 2021 ms    |
| **Plugin (PR)**          | **11 ms**  | **18 ms** | **68.5 ms** | **120 ms** |
|                          |            |           |             |            |
| Frontend Plugin (1.32.0) | 20 ms      | 140 ms    | 741 ms      | 1642.5 ms  |
| **Frontend Plugin(PR)**  | **6.5 ms** | **10 ms** | **41 ms**   | **90 ms**  |

## Electron

| Test context    | 100K        | 1000K     | 5000K     | 10000K     |
| --------------- | ----------- | --------- | --------- | ---------- |
| Plugin (1.26.0) | 49 ms       | 351 ms    | 2042 ms   | 4189 ms    |
| Plugin (1.32.0) | 40 ms       | 175.5 ms  | 969 ms    | 2018 ms    |
| **Plugin (PR)** | **13.5 ms** | **20 ms** | **77 ms** | **132 ms** |

## VS Code

| Test context                 | 100K  | 1000K | 5000K | 10000K  |
| ---------------------------- | ----- | ----- | ----- | ------- |
| Extension (1.73.1 local)     | 12 ms | 15 ms | 40 ms | 87 ms   |
| Web Extension (1.73.1 local) | 6 ms  | 11 ms | 34 ms | 53.5 ms |

paul-marechal:
@tsmaeder @tortmayr I had trouble formulating what was bothering me but I think I got it:

I don't understand the whole rationale behind the msgpackr extension API:

1. It stinks that the mspackr API for extensions is global to the whole process, because it means we cannot define sets of extensions to use per connection either. Already it makes me not want to rely on extensions, but I understand that it's the way it currently works.
2. I don't understand what's special with `ResponseError` that requires a global extension to be installed. If the rationale is that it's a commonly used type, then what about other recurring types like `URI`? What's the difference between those types, where one requires an extension and not the other? I have the feeling we could do without msgpackr extensions completely.

Obligatory "it works right now" so if you are in a hurry and want to merge feel free to do so, but this is my only outstanding concern.

paul-marechal:
Lastly I also am not a fan of the class + static member way of creating JS singletons, why not just exposing a global function like msgpackr does?

tortmayr:
> 1. It stinks that the mspackr API for extensions is global to the whole process, because it means we cannot define sets of extensions to use per connection either. Already it makes me not want to rely on extensions, but I understand that it's the way it currently works.

Totally agreed. This global extension mechanism is less than optimal. However, it's currently the only way to customize the encoding of certain types. We can try and raise this issue in the msgpackR repo. From what I could see up until now maintainers are quite open for feedback and suggestions so we might be able to get a fix for this in the upstream repo. 
We should tackle this with a follow up issue.



> 2\. I don't understand what's special with `ResponseError` that requires a global extension to be installed. If the rationale is that it's a commonly used type, then what about other recurring types like `URI`? What's the difference between those types, where one requires an extension and not the other? I have the feeling we could do without msgpackr extensions completel

The `ResponseError` is currently used in the [proxy-factory](https://github.com/eclipse-theia/theia/blob/2ec7d87d85cdd863b135a54b92c95b3eb0ee4415/packages/core/src/common/messaging/proxy-factory.ts#L282) to serialize and derserialize `Application Errors`. Some services have a special handling for these Application Errors (debug service, remote file system provider, task protocol ...). So we have to make sure the the type information for `ResponseErrors` doesn't get lost when sending the over the wire. Otherwise all services that rely on ApplicationErrors no longer work as expected.

I guess this dependency on ResponseErrors is a remainder from the original vscode-jsonrpc protocol and we should probably refactor the code to remove it. IMO this is also something that can easily be handled in a follow-up task. Nevertheless, I will take a look on it. If it can be removed rather easily without any side effects I will do it right away.

>If the rationale is that it's a commonly used type, then what about other recurring types like `URI`? What's the difference between those types, where one requires an extension and not the other?

They do require a custom extension. We currently register custom extension at two places:
https://github.com/eclipsesource/theia/blob/7927546f39d411dd2e01a9af99be7add2d9a6859/packages/core/src/common/message-rpc/rpc-message-encoder.ts#L173 and https://github.com/eclipsesource/theia/blob/7927546f39d411dd2e01a9af99be7add2d9a6859/packages/plugin-ext/src/common/rpc-protocol.ts#L224

In a nutshell, we have three topics that require custom encoding/decoding which can currently only be done (efficiently) with msgpackr extensions:
- Preserving error stack traces when sending them over the wire (see also https://github.com/eclipse-theia/theia/pull/11261#issuecomment-1330265268)
- Preserve type information for `RepsonseErrors` to ensure that the error handling derived from the old json-rpc protocol doesn't break
- The old plugin-rpc implementation had JSON replacers/revivers in place to handle the serialization of the common types you just mentioned (uri,vscode uri, range, BinaryBuffer). This is now also handled via msgpackR extensions.

>Lastly I also am not a fan of the class + static member way of creating JS singletons as it's more of a Java way. Why not just exposing a global function like msgpackr does?

I agree, a global function is probably the lesser evil here.

tsmaeder:
IMO, we're ready to merge this one: 
1. msgpackr only supports global extensions. That sucks, but what can we do except fix msgpackr? https://github.com/kriszyp/msgpackr/issues/93

2. The rest of the objections seem purely style issues, and those should not hold up merging of a PR. It's up to @tortmayr, though. (https://github.com/eclipse-theia/theia/blob/master/doc/pull-requests.md#requesting-changes)

tortmayr:
I would be in favor of merging this asap. This will give us the chance to fix potential bugs before the next release.
The rest of the objections can be easily addressed with a follow-up PR.

tortmayr:
@tsmaeder @paul-marechal @JonasHelming 
I have now squashed the commit history and updated the changelog. PR should be ready to merge now.
I will tackle the issues that Paul raised in a follow up PR (https://github.com/eclipse-theia/theia/issues/11941).

PR #11272, opened 2022-06-08
Add tooltip for tab close button

msujew:
#### What it does

Adds a tooltip/title for the tab close button. Previously it showed the tooltip of the tab, now either "Close" or "Unpin", depending on whether it is closable or pinned.

Also cleans up some `className` manipulation sorrounding the pin feature.

#### How to test

1. Hover over the close/unpin button
2. Assert that the tooltip is updated correctly based on the state
3. Assert that the pinning/unpinning feature hasn't experienced any regressions

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/tab-bars.ts
+++ b/packages/core/src/browser/shell/tab-bars.ts
@@ -17,7 +17,7 @@
 import PerfectScrollbar from 'perfect-scrollbar';
 import { TabBar, Title, Widget } from '@phosphor/widgets';
 import { VirtualElement, h, VirtualDOM, ElementInlineStyle } from '@phosphor/virtualdom';
-import { Disposable, DisposableCollection, MenuPath, notEmpty, SelectionService, CommandService } from '../../common';
+import { Disposable, DisposableCollection, MenuPath, notEmpty, SelectionService, CommandService, nls } from '../../common';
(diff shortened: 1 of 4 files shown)


packages/core/src/browser/shell/tab-bars.ts
```diff
@@ -148,6 +148,9 @@ export class TabBarRenderer extends TabBar.Renderer {
         const style = this.createTabStyle(data);
         const className = this.createTabClass(data);
         const dataset = this.createTabDataset(data);
+        const closeIconTitle = data.title.className.includes(PINNED_CLASS)
+            ? nls.localizeByDefault('Unpin')
```
vince-fugnitto, line 152: In VS Code I noticed that his is `Unpin Editor`, but given we can pin more than just editors I'm fine with the changes :+1:

vince-fugnitto (approved):
LGTM :+1: 
I confirm that the tooltip is correctly displayed, and there are no visible regressions.

packages/core/src/browser/shell/tab-bars.ts
msujew: I agree, I had this discussion with Colin over in the original PR: https://github.com/eclipse-theia/theia/pull/10817#discussion_r816709293, We decided in favor of just using `Unpin`, since we can theoretically pin anything.

PR #11274, opened 2022-06-08
test: fix spec files so they are debuggable

vince-fugnitto:
#### What it does

The pull-request fixes the `*.spec.ts` files which were previously not debuggable through the `run mocha tests` debug configuration.

The main issue was the following error:

```
The configuration is not set. Did you call FrontendApplicationConfigProvider#set?
```

#### How to test

1. CI should pass (build and especially unit tests)
2. For the three `*.spec.ts` files, it should be possible to run the debug configuration `run mocha tests` in vscode, and even set breakpoints (not possible on master). 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/keybinding.spec.ts
+++ b/packages/core/src/browser/keybinding.spec.ts
@@ -14,8 +14,14 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { enableJSDOM } from '../browser/test/jsdom';
+import { enableJSDOM } from './test/jsdom';
 let disableJSDOM = enableJSDOM();
+
+import { FrontendApplicationConfigProvider } from './frontend-application-config-provider';
+import { ApplicationProps } from '@theia/application-package/lib/application-props';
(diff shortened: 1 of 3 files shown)


packages/core/src/browser/keybinding.spec.ts
```diff
@@ -14,9 +14,15 @@
...
 
-import { enableJSDOM } from '../browser/test/jsdom';
+import { enableJSDOM } from './test/jsdom';
 let disableJSDOM = enableJSDOM();
 
+import { FrontendApplicationConfigProvider } from './frontend-application-config-provider';
+import { ApplicationProps } from '@theia/application-package/lib/application-props';
+FrontendApplicationConfigProvider.set({
+    ...ApplicationProps.DEFAULT.frontend.config
+});
```
msujew, line 24: Suggestion: Would it make sense to extract this into a dedicated method? I'm open for the current style, but it's just a lot of redundancy.

vince-fugnitto: @msujew I cleaned up the code, and existing usages of `FrontendApplicationConfigProvider.set` since it already handles populating the default configuration values. Is it the idea you had in mind?

PR #11277, opened 2022-06-08
Reuse same DebugSessionWidget for all debug sessions.

### LINKED ISSUE #9930: Debug Session Widget Creation Routine Prevents Drag-Drop
As part of #9644, it was found that the way `DebugSessionWidget`s are created and managed makes it difficult to include its subwidgets in the Drag-Drop functionality implemented in that PR. 

Part of the reason is that those widgets are inacessible to the shell layout restoration routine. When the `ShellLayoutRestorer` is storing the layout, if it encounters a `Widget` it tries to convert that widget to a description:

https://github.com/EstherPerelman/theia/blob/84bca9b36ac6a2029a57d9406d8c13f695aa0397/packages/core/src/browser/shell/shell-layout-restorer.ts#L205-L221

To do that, it asks the `WidgetManager` to get the description, and that only works _if the widget was created through the `WidgetManager`_

https://github.com/EstherPerelman/theia/blob/84bca9b36ac6a2029a57d9406d8c13f695aa0397/packages/core/src/browser/widget-manager.ts#L237-L250

But, the `DebugSessionWidget` parts are created without going through the `WidgetManager`

https://github.com/EstherPerelman/theia/blob/84bca9b36ac6a2029a57d9406d8c13f695aa0397/packages/debug/src/browser/view/debug-session-widget.ts#L41-L45

So they always wind up with no description. 

The deeper cause of the mismatch is that we maintain one `DebugSessionWidget` for every `DebugSession`, rather than allowing the same set of widgets to hook up to multiple sessions, depending on the user's selection. That means that whereas in VSCode there is _the_ variables that can be dragged anywhere and still hook up to whatever session is active / selected, we have as many threads widgets as we have sessions.

_Originally posted by @colin-grant-work in https://github.com/eclipse-theia/theia/pull/9644#discussion_r690621050_

colin-grant-work:
#### What it does

- [x] Depends on #11268

Fixes #9930 by using the same `DebugSessionWidget` for all debug sessions. It removes a fair bit of code that assumes that we will open many `DebugSessionWidget`s, which could break downstream applications that exploit that assumption.

#### How to test

1. Start more than one debug session (e.g. start one using the `Run Mocha test` configuration, and another on a basic JS file using the dynamic session)
2. Switch between sessions.
3. Assert that switching between sessions updates widgets appropriately (or at least as on master) and that the debug toolbar controls the currently selected session.
4. Move some of the children of the `DebugSessionWidget` into other view containers, and assert that (3) is still true.
5. After moving some of the children of the `DebugSessionWidget`, restart the application.
6. Assert that the moved children remain where they were before the restart.

#### Review checklist

#### Reminder for reviewers


packages/debug/src/browser/view/debug-view-model.ts
```diff
@@ -62,62 +57,36 @@ export class DebugViewModel implements Disposable {
...
-    protected readonly options: DebugViewOptions;
-
     @inject(DebugSessionManager)
     protected readonly manager: DebugSessionManager;
 
     @inject(DebugWatchManager)
     protected readonly watch: DebugWatchManager;
 
-    protected readonly _sessions = new Set<DebugSession>();
+    // protected readonly _sessions = new Set<DebugSession>();
```
msujew, line 66: Minor: I don't think it makes sense to preserver this line for historic reasons. Leftover?

packages/debug/src/browser/debug-frontend-application-contribution.ts
```diff
@@ -721,31 +717,9 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
         });
         registry.registerCommand(DebugSessionContextCommands.REVEAL, {
             execute: () => this.selectedSession && this.revealSession(this.selectedSession),
-            isEnabled: () => this.hasSessionWidget,
-            isVisible: () => !this.selectedThread && this.hasSessionWidget
-        });
-        registry.registerCommand(DebugSessionContextCommands.OPEN_LEFT, {
```
msujew, line 727: Suggestion: Should we mark `OPEN_LEFT`, `OPEN_RIGHT` and `OPEN_BOTTOM` as deprecated? They don't have an implementation anymore. They should probably also be removed from this code:

https://github.com/eclipse-theia/theia/blob/0512a5f4ede37547a9949d514936daf8c972fcb8/packages/debug/src/browser/debug-frontend-application-contribution.ts#L559-L564

msujew (commented):
That's a lot of removed code :D

Anyway, I couldn't notice any regressions from master and the improved behavior works correctly

- [x] Moving a debug view part to another view container works as expected
- [x] Changing to another debug session (in one container) updates the call stack view (in another container) correctly
- [x] Moved view parts stay where they are supposed to be after reloading the app

A minor thing which also doesn't work on master currently is that switching between debug sessions doesn't update the select component (but correctly updates the debug repl):

image

This is likely out of scope for this PR, but I just wanted to write it down somewhere.

colin-grant-work:
@thegecko, I know that debugging is a topic of interest to ARM. Is the ability to open multiple sessions simultaneously in different areas important to you?

msujew (approved):
Alright, looks good 👍

thegecko:
> @thegecko, I know that debugging is a topic of interest to ARM. Is the ability to open multiple sessions simultaneously in different areas important to you?

Thanks for the heads up. At the moment, we don't require support for running multiple debuggers with separate UI panels.
As long as this works (as confirmed by others above) and this mimics the VS Code workflow then I'm happy to get this merged.

In fact it makes my open PR a little simpler (although it will need rewriting). I'll revisit that when this is merged:

https://github.com/eclipse-theia/theia/pull/10748

thegecko (approved):
Code looks good and tested the debugger still works as expected

Issue #5757, opened 2019-07-19
[editor-preview] double-clicking editor tab maximizes view instead of docking


vince-fugnitto:
**Description**

In order to persist an `editor-preview` (dock the editor), the tab of the editor can be double-clicked similarly to vscode. However, after the change #4744 this is no longer the case since double-clicking any editor results in maximizing the view. Should the `editor-preview`'s behavior be updated in this case to first persist the tab (first double-click) and any subsequent double-clicks maximize the view (since it is now treated as a regular `editor`)?

**How to test**

1. start the application, confirm that the preferences `editor.enablePreview` is enabled and `list.openMode` is set to `singleClick`
2. open an editor by single-clicking a file in the explorer
3. double-click the editor's tab (expect the editor to be docked but instead the view is maximized)

akosyakov:
Can we handle it somehow with DOM mouse click events? i.e. editor-preview should stop even propagation if it pins an editor.

vince-fugnitto:
> Can we handle it somehow with DOM mouse click events? i.e. editor-preview should stop even propagation if it pins an editor.

It might be possible, if I have the time I can try it out :)

PR #11279, opened 2022-06-09
core: remove `maximize` with double-click

vince-fugnitto:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/5757.

The pull-request removes the problematic and often annoying behavior of when a tabbar is double-click we trigger a `maximized` state.  The behavior was previously problematic as it does not align with vscode, and often we are attempting to dock panels and not maximize them with double-clicks. 

The maximized functionality is still accessible by the command, and keybinding.

#### How to test

1. start the application and open files
2. confirm that double-clicking the tabbar for files, or widgets in the main-area or bottom panel does not maximize them
3. enable the preference `editor.enablePreview`
4. open a file as preview, double clicking the tabbar should dock the file (turn it into a regular editor)
5. confirm the command and keybinding for `view: toggle maximized` works well

#### Review checklist

#### Reminder for reviewers


vince-fugnitto:
Alternatively (as Colin mentioned offline), we could additionally control this behavior using a preference so we leave it up to the user to decide if they want to maximize using double-click or not. In order to align with the default behavior in VS Code I would make the preference off by default.

colin-grant-work:
Although I find the current behavior annoying, it is an easily discoverable feature (or bug) in the application. We may want to accommodate users who have come to rely on the behavior by gating the current behavior behind a preference or making it easy to modify the behavior in downstream applications. If we opt for the latter course, I think it would make sense to make the default implementation of the double click handler a no-op - in the current code, it's easier to override the double click handler than the method that adds the handler.

colin-grant-work (approved):
This looks good to me. With the preference set to `false`, tabs do not maximize on double click, which is refreshing. With preference `true`, they do, so users who rely on that behavior can still activate it if they wish.

CHANGELOG.md
```diff
@@ -53,6 +53,8 @@
 - [workspace] removed deprecated `getDefaultWorkspacePath` [#11185](https://github.com/eclipse-theia/theia/pull/11185)
 - [vsx-registry] removed deprecated `VSXExtensionsCommands` re-export [#11185](https://github.com/eclipse-theia/theia/pull/11185)
 - [plugin] Remove `TreeItem2` from proposed plugin API. `TreeItem` can be used instead. [#11288](https://github.com/eclipse-theia/theia/pull/11288) - Contributed on behalf of STMicroelectronics
+- [core] removed handling that maximized tabbars when performing a double-click [#11279](https://github.com/eclipse-theia/theia/pull/11279)
+  - `handleDblClickEvent` in `tab-bars.ts` is deprecated.
```
msujew, line 57: Minor: I don't think this changelog entry applies anymore.

msujew (approved):
Also looks good to me 👍

Issue #11063, opened 2022-04-22
Allow fine-grained CSS control for theming


msujew:
### Feature Description:

VSCode allows fine-grained CSS control using the `registerThemingParticipant((theme: IColorTheme, collector: ICssStyleCollector) => void)` method. In particular, it allows to set/change CSS rules depending on the currently selected theme colors.

I believe that this is a missing piece in our framework to implement https://github.com/eclipse-theia/theia/issues/6743. In a recent PR I encountered an CSS `outline` property, which would only work if the currently selected theme is a high contrast theme. In every other case it would lead to unexpected effects. So we definitely have a need to be able to control CSS rules in a more fine-grained way than currently possible.

In the spirit of DI and extensibility, my proposal would be to have something like a `ThemingParticipant` contribution provider which fulfills the same purpose as VSCode's `registerThemingParticipant`. However, I would like to hear the opinions of other contributors on that matter.

colin-grant-work:
I'm strongly in favor. I believe that we could implement high-contrast theming without this, but doing so would require creating separate CSS files for all HC theme effects and dynamically loading and unloading them (we have a similar mechanism in place in our `variables-light.useable.css` and `variables-dark.useable.css`), but I believe that this style of contribution would be much more flexible and transparent.

Issue #6743, opened 2019-12-12
high contrast theming


akosyakov:
Review how VS Code applies high contrast colors https://code.visualstudio.com/api/references/theme-color#contrast-colors and apply them similarly in Theia.

We leave it out from the initial theming.

kittaakos:
Some issues are also listed here: https://github.com/eclipse-theia/theia/issues/7353
 - selection in the `Explorer`,
 - border of the buttons in the dialogs,
 - tbd

kittaakos:
The context menu, in the browser example, is also broken:

![Screen Shot 2020-03-18 at 09 08 05](https://user-images.githubusercontent.com/1405703/76939349-81f95280-68f8-11ea-94f5-f25f7dde73eb.jpg)

kittaakos:
[nitpick]: I think the border of the left side-bar is not aligned with the main menu bar:
![Screen Shot 2020-03-18 at 09 13 25](https://user-images.githubusercontent.com/1405703/76939552-def50880-68f8-11ea-9b5d-d23925f8ee1b.jpg)

kaiyue0329:
In the Problems View, there is no border around the selected node.

In VS code,
![image](https://user-images.githubusercontent.com/23107734/77187786-62565b80-6aab-11ea-8c07-b137cdd2784d.png)

In Theia,
![image](https://user-images.githubusercontent.com/23107734/77188083-d264e180-6aab-11ea-8b0e-fc319683d29c.png)
(Arrow pointing to the selected node)

kittaakos:
Editor and Explorer decorators always white with the HC theme. I would expect red and yellowish colors.
![Screen Shot 2020-04-02 at 16 13 43](https://user-images.githubusercontent.com/1405703/78259439-01903f80-74fd-11ea-9afc-3148a4f3b3f8.jpg)

kittaakos:
Also: https://github.com/eclipse-theia/theia/issues/7489

kittaakos:
Support for hover. VS Code:
![screencast 2020-04-02 17-21-32](https://user-images.githubusercontent.com/1405703/78269368-c6483d80-7509-11ea-9ba1-cd9dfe5eee21.gif)

TODO: Revisit whether we want to support it for dialog buttons or not. VS Code does not have any dialogs but they do not support it in the Explorer:

[...]

PR #11280, opened 2022-06-09
Enable fine-grained styling in code

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/11063
Closes https://github.com/eclipse-theia/theia/issues/6743

Adds a new `StylingService` (and an associated `StylingParticipant`) which allows us to dynamically change styles using code. This has the advantage that we can set styles based on the existence of a color in a color theme.

As a simple example, we can take the notification hover color, which isn't set for some themes (hc-black). Hovering over a notification when the color isn't set will make the background transparent, while it should usually just keep the existing background color.

I used this feature to also implement the missing capabilities for high contrast theming and clean up some more (unused) css files (such as the `variables-dark.useable.css`/ `variables-bright.useable.css` files).

#### How to test

1. Open VSCode and Theia side by side
2. Select the dark high contrast theme for both of them
3. Assert that they behave (close to) the same with the high contrast theme
4. Assert that no visual regressions appear on other themes

#### Review checklist

#### Reminder for reviewers


packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -2238,4 +2245,164 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
             }
         );
     }
+
+    registerThemeStyle(theme: ColorTheme, collector: CssStyleCollector): void {
```
colin-grant-work, line 2249: This aggregates a lot of styling that it might be nice to do more locally, both for legibility and overridability (since overriding this method would likely mean copying it, though I guess you could also just call `super` and then add the rules you wanted later in the same sheet). One easy to implement but harder to override idea might be something like a static `registerThemeStyle` method on, say, `TreeWidget` and then a registration `bind(StylingParticipant).toConstantValue(TreeWidget);`; a heavier but easier to override alternative would be adding a little `TreeWidgetStylingParticipant` service (and `StatusBarStylingParticipant` etc. etc.). What do you think about moving some of the styling closer to the affected elements, and what the best way to do that would be?

packages/core/src/browser/styling-service.ts
```diff
@@ -0,0 +1,82 @@
...
+            addRule: rule => rules.push(rule)
+        };
+        for (const themingParticipant of this.themingParticipants.getContributions()) {
+            themingParticipant.registerThemeStyle(colorTheme, styleCollector);
+        }
+        const fullCss = rules.join('\n');
+        this.cssElement.innerText = fullCss;
+    }
+
+    protected createStyleElement(className?: string): HTMLStyleElement {
```
colin-grant-work, line 76: There are some utilities for this in `packages/core/src/browser/decoration-style.ts` (could bear to be renamed), that it may be desirable to reuse.

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -2238,4 +2245,164 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
...
+        if (isHighContrast(theme.type) && focusBorder) {
+            collector.addRule(`
+                #theia-statusBar .area .element.hasCommand:hover {
+                    outline: 1px dashed ${focusBorder};
+                    cursor: pointer;
+                }
+                #theia-statusBar .area .element.hasCommand:active {
+                    outline: 1px solid ${focusBorder};
+                    cursor: pointer;
+                }
```
colin-grant-work, line 2382: Do the `cursor` settings need to be here, or could they be in static CSS?

colin-grant-work (commented):
Looks super good 👍, and you've made it look easy! I thought it would take a lot more code than this to implement the feature. I haven't noticed any regressions in a quick look, but I'll leave an HC theme in place for a bit to see if anything comes up. A few comments on the code.

msujew:
@colin-grant-work Thanks, I split the styling participants related to a certain area of the app into one class each 👍

colin-grant-work (approved):
With the modularization, it should be easier for downstream applications to customize or add to particular parts. 👍

PR #11287, opened 2022-06-13
Add imports for describe and it

Niklas-Dahlquist:

Signed-off-by: Niklas DAHLQUIST <niklas.dahlquist@st.com>

#### What it does

When relying on implicit imports, wrong classes can be imported since other test frameworks define describe and it

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-manager/src/package.spec.ts
+++ b/dev-packages/application-manager/src/package.spec.ts
@@ -21,6 +21,7 @@
    showing 0% coverage, instead of no report.
    This file can be removed once we have real unit
    tests in place. */
+import { describe, it } from 'mocha';
 
 describe('application-manager package', () => {
 


vince-fugnitto (commented):
@Niklas-Dahlquist any reason why only `application-manager` was updated, I see that all root spec files for each extension does not include the imports.

Niklas-Dahlquist:
A build of Theia included jest and global type definitions gave problem in application-manager. Sorry I didn't check if there are more packages missing imports. Also noted that test code is included in normal build and that is a bit unfortunate and could be potential for change to introduce a test folder in each package.

paul-marechal:
I am not sure why you had issues with this spec file specifically?

If I understand correctly you are running Jest and it tried running this file? If so it would be a configuration problem on your side?

Niklas-Dahlquist:
Could be since the only two dev-packages used are cli which doesn't have a spec file. The reason is since both jest and mocha do global type definitions, the spec file included the definition of jest and that could be prevented by having an explicit import. So two real resons why are spec files included in build and then caused by not declaring imports. But if a bigger overhaul how spec files should be organized and declaring the specific imports in more spec files is a bigger PR than this one  was trying to address.

paul-marechal:
Closing because of staleness, feel free to re-open with a comment.

PR #11289, opened 2022-06-13
Update `Configure Display Language` command

msujew:
#### What it does

Updates the UX of the `Configure Display Language` command to align to VSCode version 1.68 (see [changelogs](https://code.visualstudio.com/updates/v1_68#_configure-display-language-improvements)).

VSCode:

image

Theia (new):

image

#### How to test

1. Run the `Configure Display Language` command
2. Assert that the new functionality (displaying available languages from ovsx) works as it does in vscode
    * Selecting one of the "Available" languages should install it and ask to confirm the language change
    * Selecting one of the "Installed" language should show no regressions

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/i18n/language-quick-pick-service.ts
+++ b/packages/core/src/browser/i18n/language-quick-pick-service.ts
@@ -33,9 +33,9 @@ export class LanguageQuickPickService {
     @inject(WindowService) protected readonly windowService: WindowService;
 
     async pickDisplayLanguage(): Promise<string | undefined> {
-        const quickInput = this.quickInputService.createQuickPick();
+        const quickInput = this.quickInputService.createQuickPick<LanguageQuickPickItem>();
         const installedItems = await this.getInstalledLanguages();
-        const quickInputItems: (QuickPickItem | QuickPickSeparator)[] = [
+        const quickInputItems: (LanguageQuickPickItem | QuickPickSeparator)[] = [
             {
                 type: 'separator',
                 label: nls.localize('theia/core/installedLanguages', 'Installed languages')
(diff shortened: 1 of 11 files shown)


packages/core/src/browser/i18n/language-quick-pick-service.ts
```diff
@@ -0,0 +1,124 @@
...
+        const installedItems = await this.getInstalledLanguages();
+        const quickInputItems: (QuickPickItem | QuickPickSeparator)[] = [
+            {
+                type: 'separator',
+                label: nls.localize('theia/core/installedLanguages', 'Installed languages')
+            },
+            ...installedItems
+        ];
+        quickInput.items = quickInputItems;
+        quickInput.busy = true;
```
colin-grant-work, line 46: Do we have any visual indicator that reflects this state? When I tested this, the experience was a bit jarring: a moment or two of just the installed languages with the available languages suddenly popping up - I couldn't see any indication that anything was happening in the background. VSCode isn't much better, but it would be nice to have something.

packages/core/src/browser/i18n/language-quick-pick-service.ts
```diff
@@ -0,0 +1,124 @@
...
+            quickInput.busy = false;
+        });
+
+        return new Promise(resolve => {
+            quickInput.onDidAccept(async () => {
+                const selectedItem = quickInput.selectedItems[0] as LanguageQuickPickItem;
+                // Some language quick pick items want to install additional languages
+                // We have to await that before returning the selected locale
+                await selectedItem.execute?.();
+                resolve(selectedItem.languageId);
```
colin-grant-work, line 79: Best to guard with `if (selectedItem)` or the like. It _should_ always  exist, but in the past, there have been errors related to similar behavior in the color and icon theme quick picks.

colin-grant-work (commented):
Functionally, this worked well for me. A few comments on the code.

packages/core/src/browser/i18n/language-quick-pick-service.ts
msujew: It seems like the busy indicator for quick inputs in VSCode has been broken for two years (starting in v1.44), see https://github.com/microsoft/vscode/issues/128966.

However, the actual html element is correctly added to the DOM, so with some CSS magic we're able to correctly display the busy indicator :)

colin-grant-work (approved):
The new functionality is working well, and it's nice to have the progress marker 👍

PR #11294, opened 2022-06-14
Improve RPC message encoding

tortmayr:
#### What it does

- Refactor `RPCMessageEncoder` to enable detection of circular object structures. Circular structures can not be serialized. The detection enables us to fail early instead of indirectly failing once the maximum stack frame limit is reached.
- Add a custom encoder for functions. This ensures that direct function entries in arrays can be serialized.
- Add custom encoder for common collection types (maps, sets). Enables direct use of collection types as rpc call arguments (e.g. logging statements)
- Restore behavior of the old protocol if a log RPC call can not be sent to the backend due to an encoding error 
   - Add custom `EncodingError` type to explicitly catch these errors
   - Refactor `loggerFrontendModule` to catch encoding errors and continue execution normally. i.e. log call is just logged in frontend console and not sent to backend

Fixes https://github.com/eclipse-theia/theia/issues/11249
 
#### How to test

The bug described in https://github.com/eclipse-theia/theia/issues/11249#issuecomment-1147961021 should be fixed. i.e.:

1. Add this log to DebugSessionManager set currentSession:
 ```bash
 console.log('SENTINEL FOR SETTING THE CURRENT SESSION', current);
```

2. Start a debug session inside the application.
3. Observe that the frontend (browser) logs log the message as expected.
4. Observe that no error logged in the frontend console
5. Observe that the backend logs do not show the message
( Same behavior as with the old RPC protocol)
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/message-rpc/rpc-message-encoder.ts
+++ b/packages/core/src/common/message-rpc/rpc-message-encoder.ts
@@ -382,28 +382,19 @@ export class RpcMessageEncoder {
         this.registerEncoder(ObjectType.Object, {
             is: value => typeof value === 'object',
             write: (buf, object, visitedObjects, recursiveEncode) => {
-                try {
-                    if (visitedObjects.has(object)) {
-                        throw new EncodingError('Object to encode contains circular references!');
+                const properties = Object.keys(object);
+                const relevant = [];
+                for (const property of properties) {
+                    const value = object[property];
+                    if (typeof value !== 'function') {
+                        relevant.push([property, value]);
                     }
-                    visitedObjects.add(object);
+                }
 
-                    const properties = Object.keys(object);
-                    const relevant = [];
-                    for (const property of properties) {
-                        const value = object[property];
-                        if (typeof value !== 'function') {
-                            relevant.push([property, value]);
-                        }
-                    }
-
-                    buf.writeLength(relevant.length);
-                    for (const [property, value] of relevant) {
-                        buf.writeString(property);
-                        recursiveEncode?.(buf, value, visitedObjects);
-                    }
-                } finally {
-                    visitedObjects.delete(object);
+                buf.writeLength(relevant.length);
+                for (const [property, value] of relevant) {
+                    buf.writeString(property);
+                    recursiveEncode?.(buf, value, visitedObjects);
                 }
             }
         });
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by colin-grant-work — APPROVED** [review]
I've only glanced over the code, but functionally this appears to be in order. Circular objects (tested on all Widgets that pass through the `WidgetManager`) are logged to the front end console without error and nothing appears in the backend log output. Logging many logs also appears much snappier than before the change.

File: packages/core/src/common/message-rpc/rpc-message-encoder.ts [hunk 1]
```diff
@@ -388,6 +428,16 @@ export class RpcMessageEncoder {
             write: (buf, value) => buf.writeString(JSON.stringify(value))
         });
 
+        this.registerEncoder(ObjectType.Map, {
+            is: value => value instanceof Map,
```
tsmaeder, line 432: Don't we have to treat anything that can hold a reference the same as the `typeof x === object`? I mean I can do

```
const x= new Set();
const y= new Set();
x.add(y);
y.add(x);
```
Bingo, circular reference 🤷

packages/core/src/common/message-rpc/rpc-message-encoder.ts
```diff
@@ -348,22 +374,36 @@ export class RpcMessageEncoder {
...
+                    if (visitedObjects.has(object)) {
+                        throw new EncodingError('Object to encode contains circular references!');
+                    }
+                    visitedObjects.add(object);
+
+                    const properties = Object.keys(object);
+                    const relevant = [];
+                    for (const property of properties) {
+                        const value = object[property];
+                        if (typeof value !== 'function') {
```
tsmaeder, line 395: If this line filters out functions, why do we need an encoder for functions?

packages/core/src/common/message-rpc/rpc-message-encoder.ts
```diff
@@ -476,22 +526,23 @@ export class RpcMessageEncoder {
...
+                    this.writeTypedValue(innerBuffer, innerValue, _visitedObjects);
                 });
                 return;
             }
         }
+        throw new EncodingError(`No suitable value encoder found for ${value}`);
     }
 
-    writeArray(buf: WriteBuffer, value: any[]): void {
+    writeArray(buf: WriteBuffer, value: any[], visitedObjects = new WeakSet()): void {
```
tsmaeder, line 542: I would feel more comfortable not using a default argument here: without it, we would catch cases where the "visitedObjects" are not forwarded.

packages/core/src/common/message-rpc/rpc-message-encoder.ts
```diff
@@ -119,11 +122,12 @@ export interface ValueEncoder {
      * Write the given value to the buffer. Will only be called if {@link is(value)} returns true.
      * @param buf The buffer to write to
      * @param value The value to be written
+     * @param visitedObjects The collection of already visited (i.e. encoded) objects. Used to detect circular references
      * @param recursiveEncode A function that will use the encoders registered on the {@link MessageEncoder}
      * to write a value to the underlying buffer. This is used mostly to write structures like an array
      * without having to know how to encode the values in the array
      */
-    write(buf: WriteBuffer, value: any, recursiveEncode?: (buf: WriteBuffer, value: any) => void): void;
+    write(buf: WriteBuffer, value: any, visitedObjects: WeakSet<object>, recursiveEncode?: (buf: WriteBuffer, value: any, visitedObjects: WeakSet<object>) => void): void;
```
tsmaeder, line 130: Doesn't a regular set work just as well? We're never holding on to a reference to it anyway, so retention is not a problem.

tsmaeder (changes requested):
I think circular structures can still slip through, see comments.

packages/core/src/common/message-rpc/rpc-message-encoder.ts
colin-grant-work: `Set`s aren't a problem here, but JavaScript is:

```javascript
const a = new Set();
// undefined
typeof a
// 'object'
const b = 'hello'
// undefined
const c = 'other string'
// undefined
 b.c = c;
// 'other string'
 c.b = b
// 'hello'
typeof b
// 'string'

File: packages/core/src/common/message-rpc/rpc-message-encoder.ts (hunk 4 above)
Comment (tortmayr) [reply packages/core/src/common/message-rpc/rpc-message-encoder.ts:130]: It would work as well. However, ciruclar reference dection is the prime usecase for weak sets and its even mentioned in the [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet#use_case_detecting_circular_references)

>The number of objects or their traversal order is immaterial, so a WeakSet is more suitable (and performant) than a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) for tracking object references, especially if a very large number of objects is involved.

So this is mainly a performance optimization. It probably has little to no effect but I don't see any harm in using it.

File: packages/core/src/common/message-rpc/rpc-message-encoder.ts (hunk 2 above)
Comment (tortmayr) [reply packages/core/src/common/message-rpc/rpc-message-encoder.ts:395]: This line only filters out object properties that are functions. However, we also have to consider that functions can occur as elements of an array. (e.g. 
```ts
[()=> { // do something}, ()=>{// do something else}];
```
)

Of course we could also just filter out functions in the `writeArray` method but I opted for the same approach that JSON.stringify/parse uses and decided to encode this functions as emtpy objects.

I have no hard opinion on this. If you think it's better to filter out functions entirely I'm happy to apply the changes.

packages/core/src/common/message-rpc/rpc-message-encoder.ts
tortmayr: Good catch. I have refactored the code and now all object types (other than null/undefined) are going through the reference check. I have also added a test case for circular references based on  the example that @tsmaeder provided.

tortmayr:
Thanks for the review @tsmaeder  & @colin-grant-work. I have pushed an update that should address the review comments.

tsmaeder (approved):

tsmaeder:
The code looks correct to me now and cursory testing revealed no obvious problems.

Issue #11283, opened 2022-06-10
too much `viewWelcomeNodes`


meng-jpg:
### Bug Description:
When the state of workspace is `no folder opened`, close the `Explorer` panel and open it from the `View` menu again.
Then one more `visibleItem` of the `TreeViewWelcomeWidget` will be displayed.
This may be caused by the response mode of the `onDidCreateWidget` event and the storage form of the `viewWelcome` nodes.

### Steps to Reproduce:

1. run theia electron
2. right click on the explorer icon to call the context menu
3. choose close
4. reopen the  panel from `View->Explorer`
<img width="616" alt="截屏2022-06-10 下午2 15 32" src="https://user-images.githubusercontent.com/74631407/173003947-d64bb0df-57a4-41ad-909a-e6c9dd2a0898.png">

### Additional Information

- Operating System: macOS Big Sur 11.2
- Theia Version: 1.25

inlann:
Theia 1.26 still has this bug

yiliang114:
![image](https://user-images.githubusercontent.com/11473889/174540407-fa52ca73-2c8c-4b26-ae8b-20b70cbf6d48.png)

Same error in Theia Web (examples/browser)1.26

PR #11312, opened 2022-06-20
Dispose explorer view welcome correctly

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/11283

As we call `registerViewWelcome` each time we recreate the `FileNavigatorWidget`, we need to clean up the disposable that we create, once the widget is disposed.

Also localizes the view welcome text.

#### How to test

1. Close your workspace
2. Close the file explorer (using the `Close` context menu entry)
3. Open the file explorer
4. Assert that only a single view welcome entry is displayed

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/view/plugin-view-registry.ts
+++ b/packages/plugin-ext/src/main/browser/view/plugin-view-registry.ts
@@ -403,6 +403,9 @@ export class PluginViewRegistry implements FrontendApplicationContribution {
         const toDispose = new DisposableCollection();
 
         const viewsWelcome = this.viewsWelcome.get(viewWelcome.view) || [];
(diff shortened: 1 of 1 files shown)


msujew:
@vince-fugnitto Thanks for the quick review. I agree, registering multiple elements with the same content should be prevented. I've created a commit for that.

But we should dispose of the created view welcome here anyway, since it creates entries in the `viewClauseContexts`, which should be deleted when the widget is diposed.

vince-fugnitto (approved):
LGTM :+1: I confirm that it fixes the issue reported in https://github.com/eclipse-theia/theia/issues/11283. The only difference in behavior is that vscode does not only check for `content` as duplicates but I think it should be fine.

PR #11323, opened 2022-06-21
Mimic VS Code Experimental devices

thegecko:
#### What it does

VS Code has just added experimental support for requesting access to `WebUSB`, `WebSerial` and `WebHID` devices as outlined in https://github.com/microsoft/vscode/issues/116761

The implementation is [here](https://github.com/microsoft/vscode/blob/main/src/vs/base/browser/deviceAccess.ts) and exposed as commands [here](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/browser/window.ts#L247).

This PR reflects these additions by exposing the same functionality in Theia via the commands:

- `workbench.experimental.requestUsbDevice`
- `workbench.experimental.requestSerialPort`
- `workbench.experimental.requestHidDevice`

These are added so that extensions utilising the experimental commands will work in Theia and VS Code alike.
I haven't seen experimental commands exposed in Theia anywhere else, so keen to understand if this is OK.

#### How to test

Execute the commands to request access to a device.

#### Review checklist

#### Reminder for reviewers


colin-grant-work (commented):
> I haven't seen experimental commands exposed in Theia anywhere else, so keen to understand if this is OK.

Usually we lag far enough behind that the question of keeping up with experimental API rarely arises :-). In general I don't see any problem with it, but we'll have to watch for changes on the VSCode side to make sure that we adjust if they decide to change names before they finalize the API. 

Regarding the code, one question I have is whether we should expose the interfaces somewhere (similar to the separate file in VSCode) in case we want to use them somewhere.

thegecko:
> we'll have to watch for changes on the VSCode side to make sure that we adjust if they decide to change names before they finalize the API.

Don't worry, I'll be watching these like a hawk!

> one question I have is whether we should expose the interfaces somewhere (similar to the separate file in VSCode) in case we want to use them somewhere.

I've updated the PR to expose these interfaces, but I'd probably recommend anyone using this functionality uses the "official" types (e.g. [@types/w3c-web-usb](https://www.npmjs.com/package/@types/w3c-web-usb)).

Users would also need to explicitly cast the return value of any executed command to the correct type due to the nature of how commands are executed.

Issue #11321, opened 2022-06-21
Files without extension are filtered in `ElectronFileDialogService` even though no filter is provided


planger:
### Bug Description:

The file selection dialog of Electron hides files without extensions also if no specific file filters are specified when calling `FileDialogService#showOpenDialog` in Electron with `OpenFileDialogProps#canSelectFiles = true` and `OpenFileDialogProps#canSelectFolders = false`.

`ElectronFileDialogService` adds an _All Files_ filter with an extension pattern of `'*'` in that case:

https://github.com/eclipse-theia/theia/blob/ce0128453165d9a9493a7a958fcfaa4b0a2c6e9e/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts#L128

At least for Ubuntu / GTK, this results in hiding all files that have no extension due to Electron issue https://github.com/electron/electron/issues/34492. This issue has been fixed, but a backport of this fix only exists for Electron 18.x upwards.

I'm wondering why we are adding this _All Files_ filter in the first place.
If callers of the `FileDialogService` don't provide any filters, then all files should be shown, which is also accomplished if we don't add another _All Files_ filter. Not adding any filter by-passes the aforementioned Electron bug and thus also files without any extension would show up.
If callers of the `FileDialogService` provide filters, why do we want to "overrule" them by adding another _All Files_ filter?

Would you agree to remove the line linked above or maybe just adding it, if callers specify a filter?
According to your feedback, I'm happy to open a PR.

Thanks!

### Steps to Reproduce:

1. Use Ubuntu Linux
2. Start Electron Example App
3. _Open Settings (UI)_
4. Navigate to _Terminal › External: Windows Exec_
5. Click _Browse_
6. Navigate to a folder that contains a file without an extension
7. Observe how it doesn't show in the dialog even though it exists

### Additional Information

- Operating System: Linux
- Theia Version: 1.26.0

vince-fugnitto:
@planger I believe the discussions from https://github.com/eclipse-theia/theia/pull/9659 might be interesting as to why we include the "all filters" filter.

planger:
@vince-fugnitto Thanks for the feedback! The argument in #9659 certainly makes sense. How about only adding the _All Files_ filter if there is a filter specified by the caller? This would address the requirement in #9659, but still bypass the electron bug when no file filter is specified in the first place. If a file filter is specified, we'll run into the electron bug anyway.

vince-fugnitto:
[...]

PR #11325, opened 2022-06-22
fix: Omit _All Files_ filter in Electron on Linux

planger:
#### What it does
Omits adding the _All Files_ filter, if we are on Linux and no other filter is specified. Otherwise, we run into https://github.com/eclipse-theia/theia/issues/11321 which causes an extension filter `['*']` to hide files without extensions. Once we are on Electron >18, we should be able to add the _All Files_ filter again in all cases, as this Electron bug seems to be resolved in Electron above >18.

Fixes https://github.com/eclipse-theia/theia/issues/11321

Change-Id: Iec3120a6260f20b7f0d42c091aa00bde33fa915a
Signed-off-by: Philip Langer <planger@eclipsesource.com>

#### How to test
1. Use Ubuntu Linux
2. Start Electron Example App
3. _Open Settings (UI)_
4. Navigate to _Terminal › External: Windows Exec_
5. Click _Browse_
6. Navigate to a folder that contains a file without an extension
7. Observe how the file without extension shows up in the dialog

We should also ensure that other OSs aren't affected (OSX and Windows), as well as that the _All Files_ filter still shows up when a specific filter is specified. Unfortunately, I could only test the latter myself (when specific filters are specified) but not the former, because I don't have a Windows or OSX machine.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
+++ b/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
@@ -135,11 +135,8 @@ export class ElectronFileDialogService extends DefaultFileDialogService {
     /**
      * Specifies whether an _All Files_ filter should be added to the dialog.
      *
-     * We shouldn't add an _All Files_ filter, if we are on Linux and no other filter is specified.
-     * Otherwise, we run into a [bug](https://github.com/eclipse-theia/theia/issues/11321) in Electron
-     * which causes an extension filter `['*']` to hide files without extensions. Once we are on Electron >18,
-     * we should be able to add the _All Files_ filter again in all cases, as this bug seems to be resolved in
-     * Electron above >18.
+     * On Linux, the _All Files_ filter [hides](https://github.com/eclipse-theia/theia/issues/11321) files without an extension.
+     * The bug is resolved in Electron >=18.
      */
     protected shouldAddAllFilesFilter(electronProps: electron.FileDialogProps): boolean {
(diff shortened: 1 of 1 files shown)


packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
```diff
@@ -125,11 +125,28 @@ export class ElectronFileDialogService extends DefaultFileDialogService {
...
 
+    /**
+     * Specifies whether an _All Files_ filter should be added to the dialog.
+     *
+     * We shouldn't add an _All Files_ filter, if we are on Linux and no other filter is specified.
+     * Otherwise, we run into a [bug](https://github.com/eclipse-theia/theia/issues/11321) in Electron
+     * which causes an extension filter `['*']` to hide files without extensions. Once we are on Electron >18,
+     * we should be able to add the _All Files_ filter again in all cases, as this bug seems to be resolved in
+     * Electron above >18.
+     */
```
colin-grant-work, line 143:
```suggestion
    /**
     * Specifies whether an _All Files_ filter should be added to the dialog.
     *
     * On Linux, the _All Files_ filter [hides](https://github.com/eclipse-theia/theia/issues/11321) files without an extension.
     * The bug is resolved in Electron >=18.
     */
```

colin-grant-work (approved):
The fix appears sound. 👍 Optionally, the comment could be more concise.

PR #11332, opened 2022-06-23
Fix debug launch failures from variable resolution

alvsan09:
#### What it does

Fix debug launch failures from variable resolution

The debug session manager relies on the variable resolving services
to replace variables present in the launch configuration.

Some of these variables are resolved by interactive commands towards
the user, however if the user press escape or ignores a prompt, the
variable resolution will not take place and if the flow continues,
errors will be displayed to the user (unnecessarily).

The solution before this fix was relying on a check on the successful
resolution of all the variables i.e. non are 'undefined', however
there are cases were this happens and the flow shall continue.

This fix, replaces the check for the presence of `escape` entries in
an array, this array is populated with the command ids that were not
successfully resolved i.e. giving a result of 'null'.

#### How to test

1) Download and build the following test project 
https://github.com/alvsan09/hello_express

2) Open the project in theia

3) From the debug view launch the configuration `Attach to node process`, 
- Verify that a list of process is present
- Click outside the list area or the escape key
- Make sure that not Error dialogs pop up

4) Launch the configuration `Launch Backend`
5) Launch the configuration `Launch Frontend`
- Verify that a chrome browser opens successfully with the `Hello World` message.

launch_with_undefined_variables

#### Review checklist

#### Reminder for reviewers


jfaltermeier:
Hi @alvsan09, I've had a short look at the code and ran the test project. It looks good imo. 
I can confirm that escaping the process list does not lead to an error popup. 

If I understand the fix correctly the `Variable#resolve` methods may now decide whether a resolvement to `undefined` is ok or not, by pushing a value into the `trackEscape`-array. 
Currently only the `command`-Variable uses this by pushing the commandid. 

I am wondering if the `Variable` itself can decide this in all cases however. Couldn't there be a command with a legit `null` result?
So the caller of `VariableResolverService#resolve` might need a way to influence this as well. Maybe rather than passing an empty `trackEscape: []` we might pass the values of failed resolvements that we accept. So the `VariableResolverService.Context#hasEscaped()` would not only check whether there are values in the `trackEscape`-Array but also if those could be ignored/are overridden. 
Does this make sense or can we ignore this use case?

alvsan09:
Hello @jfaltermeier,
Thanks for your remarks!, 

For the question:
> Couldn't there be a command with a legit `null` result?

Checking the solution in `vscode` 
https://github.com/microsoft/vscode/blob/main/src/vs/workbench/services/configurationResolver/browser/baseConfigurationResolverService.ts#L128

we can see three resulting options from an interaction:
1) mappings is 'undefined' -> stop processing,
2) mappings size is greater than zero, proceed to variable substitution
3) mappings size equals zero, then it returns the original configuration

option 1) is the interesting one to our case, as we can see in the link below, a command execution for variable resolution resulting in 'null' will be returning 'undefined' for mappings.

https://github.com/microsoft/vscode/blob/main/src/vs/workbench/services/configurationResolver/browser/baseConfigurationResolverService.ts#L236

In `theia`, the structure is different and does not report mappings, however the command resulting in 'null' should provide an equivalent behavior to option 1 above.

The original commit was doing so, however not very intuitively, so I have change the solution to use an object holding a state
rather than an array of `ids`.

For the second part, the proposal is not quite clear to me. i.e.
How would `theia` get to know which 'id' values are OK to accept as `escaped`?
`vscode` does not seem to have any registry for this purpose, if you could please expand on the idea..

jfaltermeier:
Thank you for the clarification. 

Regarding the second part:
I had the use case of a Theia adopter that wants to use the `VariableResolverService` in their business logic in mind. 

At the moment a variable may set the `interactionsState` to `NOK`. There may also be multiple variables that do so. 
As a result the `VariableResolverService#resolve` method returns `undefined`. 
The caller of `VariableResolverService#resolve` has no further information what went wrong. 

Depending on the use case, the caller of `VariableResolverService#resolve` may want to handle this. 
(E.g. the failed variable could actually be some optional property, so removing it and showing a warning could be enough in their business logic)

For this the caller probably needs to know which variables failed to resolve and might want to have access to the "incomplete" resolved object rather than `undefined`. 

So my (slightly adjusted) proposal would be something like this:

The command-Variable could add "command:PickProcess" as failed resolvement to the interaction state rather than calling `setNOK`. 
`isNOK()` returns `true` if there was at least one failed resolvement. 
When `context.isInteractionsNOK()` is `true`, the `VariableResolverService` could add the partly resolved object to the `interactionsState` as an intermediate result and then return undefined. 

Now the caller would have access to the failed variables and and the intermediate result via the `InteractionsAggregatedState` object. 


For the current Theia use cases this would not be needed, so I would also be fine with just merging the current state. 
I just thought if we change the API anyway, we could also enable this use case, if I'm not the only one that finds it useful.

alvsan09:
@jfaltermeier, thanks for expanding on the proposal.

I have my concerns around it though,
i.e.
- When registering unresolved id's in the `interactionsState`, the logic have already parsed the variable string so it points to a specific variable instance e.g. `command`, the call back function under `Variable#resolve` does not have a knowledge of which variable it belongs to, so in order to build a useful id we would need to reconstruct this  possibly by hard-coding the `variable`'s name and joining it with `:`, or add an extra argument to the `resolve` functions so they become aware of the encapsulating `variable`'s name.
- The `#resolve` function receives `commandIdVariables` which are provided by extensions, and the resolution of a command may depend of a previous translation from the received `argument` to an entry under `commandIdVariables`, so which 'id` would be useful to register in the `interactionsState`, the one before or after translation ?

We can implement the above making some assumptions, but considering that `vscode` does not currently have this mechanism I would prefer to wait until the actual need arises. 

Please let me know what you think !

jfaltermeier (approved):
Ok, I see, thanks. Changes look good to me. I've tested with the hello_express test project and no errors where shown using "Attach to node process". Other launch configs worked as expected as well.

alvsan09:
The two initial commits were now squashed and then re-based to latest master!
Thanks @jfaltermeier!

packages/variable-resolver/src/common/variable-types.ts
```diff
@@ -21,3 +21,19 @@
...
+/**
+ * Holds a common state among variables being resolved by interactions e.g. commands.
+ * A `NOK` state indicates that at least one of the variables did not get resolved successfully.
+ */
+export class InteractionsAggregatedState {
+    protected state: 'NOK' | 'OK' = 'OK';
+
+    setNOK(state: 'NOK'): void {
+        this.state = state;
+    }
```
vince-fugnitto, line 34: Can be simplified to:

```suggestion
    setNOK(): void {
        this.state = 'NOK';
    }
```

packages/variable-resolver/src/browser/variable-resolver-service.ts
```diff
@@ -150,6 +150,10 @@ export namespace VariableResolverService {
             return true;
         }
 
+        isInteractionsNOK(): boolean {
+            return this.options.interactionsState ? this.options.interactionsState.isNOK() : false;
+        }
```
vince-fugnitto, line 155: Can be simplified to:

```suggestion
        isInteractionsNOK(): boolean {
            return this.options.interactionsState?.isNOK() ?? false;
        }
```

or 

```suggestion
        isInteractionsNOK(): boolean {
            return !!this.options.interactionsState?.isNOK()
        }
```

Issue #11316, opened 2022-06-20
Implement Internationalized Text Interpolation for Plugin Descriptions


colin-grant-work:
### Feature Description:

Some extensions have localized descriptions, but it appears that we don't handle those correctly when generating the content of their entries in the Open VSX tree or Open VSX editor views:

![image](https://user-images.githubusercontent.com/62660806/174633314-1f7ad03b-6f43-4895-a3e2-539515182bf9.png)

![image](https://user-images.githubusercontent.com/62660806/174633358-8827bf71-4f5a-4d76-962b-0dc92e7733a3.png)

Note the text `%extension.description%` indicating that the plugin expects the appropriate localized text to be interpolated.

PR #11334, opened 2022-06-24
Localize full plugin data

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/11316

Instead of only localizing the `contributes` part of plugins (such as menus, commands, etc), we should localize everything, including the name and description property of the plugin.

#### How to test

1. Assert that the `ms-vscode.node-debug2` extension exists as a built-in extension.
2. Open the vsx-registry view and scroll down to the extension. The description should not be `%extension.description%` but an actual text value.
3. Delete the extension from the `plugins` directory and restart the server/app.
4. Search for the `ms-vscode.node-debug2` extension in the vsx-registry and install it. The description should not change to `%extension.description%` but continue saying `Node.js debugging support`.
6. Uninstall the extension again, and change the local to another language. I used `zh-cn`.
7. Installing the extension again should change the description to a localized value.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/hosted/node/hosted-plugin-localization-service.ts
+++ b/packages/plugin-ext/src/hosted/node/hosted-plugin-localization-service.ts
@@ -19,7 +19,7 @@ import * as fs from '@theia/core/shared/fs-extra';
 import { LocalizationProvider } from '@theia/core/lib/node/i18n/localization-provider';
 import { Localization } from '@theia/core/lib/common/i18n/localization';
 import { inject, injectable } from '@theia/core/shared/inversify';
-import { DeployedPlugin, Localization as PluginLocalization, PluginContribution } from '../../common';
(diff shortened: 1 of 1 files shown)


colin-grant-work (commented):
I'm not entirely sure how the data for plugins that have not yet been installed is handled. Searching for `node-debug2` while using the application in Chinese, I get this: 

image

And once I install, I get this:

image

Otherwise, works very well, so if there's no convenient way to fix the content before we download, this is a significant improvement.

msujew:
@colin-grant-work The open-vsx API does not allow us to set the locale when searching for an extension and only replaces what it can find in the english `package.nls.json`. We don't have access to the actual localization files while searching for extensions, so that's the best we can do for now.

colin-grant-work (approved):
This fixes what's in our power to fix, and I see no regressions. 👍

Issue #11335, opened 2022-06-24
Changing locale does not preserve the recently used commands


msujew:
### Bug Description:

When changing the locale, opening the command palette displays no recently used commands.

### Steps to Reproduce:

1. Execute the `Configure Display Language` command
2. Open the command palette. The list of recently used commands is empty, but should actually contain the `Configure Display Language` command.

### Additional Information

- Operating System: Windows
- Theia Version: 1.26.0

PR #11336, opened 2022-06-24
Preserve recent commands for different locales

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/11335

Makes the internal list of recently used commands a simple `string` array containing the IDs of these commands. This allows us to have them independent from the selected locale.

#### How to test

1. Run the `Configure Display Language` command
2. Select another locale
3. Open the command palette, the `Configure Display Language` command should still be in the list of recently used commands.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/quick-input/quick-command-service.ts
+++ b/packages/core/src/browser/quick-input/quick-command-service.ts
@@ -16,7 +16,7 @@
 
 import { inject, injectable } from 'inversify';
 import { KeybindingRegistry } from '../keybinding';
-import { Disposable, Command, CommandRegistry, CancellationToken } from '../../common';
+import { Disposable, Command, CommandRegistry, CancellationToken, nls } from '../../common';
 import { ContextKeyService } from '../context-key-service';
 import { CorePreferences } from '../core-preferences';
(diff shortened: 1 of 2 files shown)


colin-grant-work (commented):
This worked for me, but with the caveat that the first time I ran it, I wound up with duplicates of several of my recent commands:

image

Note that `Color Theme` and `Configure Display Language` each appear twice.

msujew:
@colin-grant-work I couldn't reproduce the issue, but I'm using an intermediate `Set` now to avoid duplicates. Does that fix the issue?

colin-grant-work (approved):
Is working, and is now more robust against duplicates that have snuck into the state restoration system on application close.

colin-grant-work:
I think we should really not use `Command.equals` anywhere. The only source of truth about command identity is the `id` field, so checking anything else, for example here,

https://github.com/eclipse-theia/theia/blob/4061027c98af84900d496e65e66553d426b9801e/packages/core/src/browser/quick-input/quick-command-service.ts#L198-L209

is just misleading: it can only return false negatives and introduce duplicates.

Issue #11044, opened 2022-04-19
nls: cannot uninstall language-pack due to cache


vince-fugnitto:
#### Bug Description:

The **language-pack** extensions cannot be uninstalled since they are cached under `~/.theia/localization-cache`.
Despite an pack being uninstalled by the user, the cache means that it will always be displayed.

#### Steps to Reproduce:

1. start the application
2. install the `french language-pack`
3. execute `configure display language` - notice that both `en` and `fr` are present
4. uninstall the `french language-pack`
5. execute `configure display language` - notice that `fr` persists
6. restart the application
7. execute `configure display language` - notice that `fr` still persists
8. notice that `fr` exists under `~/.theia/localization-cache`

PR #11338, opened 2022-06-24
Unload language pack correctly

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/11044

Refactors the `LocalizationProvider` to only merge localizations when requesting them. This facilitates easier removal later.

#### How to test

1. Install a language pack and change to that language
2. Uninstall it again and reload the application
3. The app should be completely in English (`nls.locale` being `undefined`)
4. Running the `Configure Display Language` command should not show the language as being installed

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/hosted/node/hosted-plugin-localization-service.ts
+++ b/packages/plugin-ext/src/hosted/node/hosted-plugin-localization-service.ts
@@ -19,9 +19,12 @@ import * as fs from '@theia/core/shared/fs-extra';
 import { LocalizationProvider } from '@theia/core/lib/node/i18n/localization-provider';
 import { Localization } from '@theia/core/lib/common/i18n/localization';
 import { inject, injectable } from '@theia/core/shared/inversify';
-import { DeployedPlugin, Localization as PluginLocalization } from '../../common';
+import { DeployedPlugin, Localization as PluginLocalization, PluginIdentifiers } from '../../common';
 import { URI } from '@theia/core/shared/vscode-uri';
 import { EnvVariablesServer } from '@theia/core/lib/common/env-variables';
+import { BackendApplicationContribution } from '@theia/core/lib/node';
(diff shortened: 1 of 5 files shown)


packages/plugin-ext/src/hosted/node/hosted-plugin-localization-service.ts
```diff
@@ -34,22 +36,38 @@ export interface VSCodeNlsConfig {
...
 
     @inject(EnvVariablesServer)
     protected readonly envVariables: EnvVariablesServer;
 
+    protected localizationDisposeMap = new Map<string, Disposable>();
     protected translationConfigFiles: Map<string, string> = new Map();
 
+    async initialize(): Promise<void> {
+        const cacheDir = await this.getLocalizationCacheDir();
+        await fs.emptyDir(cacheDir);
```
colin-grant-work, line 52: Is it desirable to guarantee that this code completes before we start populating the cache directory in `buildLocalizations` / `deployLocalizations`? It seems like it always _should_, but it might still be a good idea to be sure.

colin-grant-work (approved):
The code looks good and the results are as described. 👍  Just one minor comment.

packages/plugin-ext/src/hosted/node/hosted-plugin-localization-service.ts
msujew: Actually, I just noticed that the backend server never awaits the `initialize` method, so we have to defer the execution of `buildTranslationConfig` until is has actually finished `emptyDir`. I added a small change which adds a `ready` promise to the service.

msujew:
@kittaakos I believe that would be fine as well, do you want to contribute that change?

kittaakos:
> @kittaakos I believe that would be fine as well, do you want to contribute that change?

Thank you for the prompt reply. Yes, I will open a PR soon.

PR #11340, opened 2022-06-24
Cancel Tree.refresh when new `root` set

colin-grant-work:
#### What it does

On application startup, these errors are logged by the `scm` package:

image

This PR addresses the error by modifying the logic for handling `set root` on the `TreeImpl` so that it cancels a `refresh` cycle when a new root is set. Previously, `refresh` would continue running and eventually discover that the `root` with which it began no longer corresponded to the current `root` of the tree, producing the error.

#### How to test

1. Start the application.
2. Assert that the errors noted above no longer occur.
3. Trees should continue to work normally: since `refresh` is called every time `set root` runs, the final `refresh` should still run to completion.

#### Review checklist

#### Reminder for reviewers


msujew (approved):
Looks good to me 👍 I can confirm that the issue exists on master and is addressed by this change.

vince-fugnitto (approved):
LGTM :+1:

PR #11347, opened 2022-06-26
fix(commad): close #11346, use available number to label when create …

### LINKED ISSUE #11346: The Resource Sequence Index is not start from available number when create untitled file
### Steps to Reproduce:

1. yarn & yarn browser build & yarn browser start.
2. Open `http://localhost:3000/` and open any workspace.
3. Use `alt + n` to call  `NEW_FILE` command some times, we can see tab named `Untitled-${index}`  one by one. And we can notice the number in label is start from `0` rather then `1`(vscode is started from 1).
4. Call  `NEW_FILE` command again after close some older Untitled file, it should used the number which is  closed just now like vscode. 

Refer to the following screenshots:

vscode behavior:

QQ20220626-180642

theia behavior:

QQ20220626-181132

### Additional Information

- Operating System: macos 
- Theia Version: 1.26.0  the latest commit id is 0e05d0ad8b91e3

yiliang114:
#### What it does

close #11346

after change:

QQ20220626-180740

#### How to test

1. yarn & yarn browser build & yarn browser start.
2. Open `http://localhost:3000/` and open any workspace.
3. Use `alt + n` to call `NEW_FILE` command some times, we can see tab named `Untitled-${index}` one by one. The number in label is start from `1` like vscode.
4. Call `NEW_FILE` command again after close some older Untitled file, it will use the number which is closed just now like vscode.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/resource.ts
+++ b/packages/core/src/common/resource.ts
@@ -320,6 +320,8 @@ export class InMemoryTextResourceResolver implements ResourceResolver {
 
 export const UNTITLED_SCHEME = 'untitled';
 
+let untitledResourceSequenceIndex = 0;
+
 @injectable()
 export class UntitledResourceResolver implements ResourceResolver {
 
@@ -409,4 +411,15 @@ export class UntitledResource implements Resource {
     get encoding(): string | undefined {
         return undefined;
     }
+}
+
+/**
+ * @deprecated Since 1.27.0. Please use `UntitledResourceResolver.createUntitledURI` instead.
+ */
+export function createUntitledURI(extension?: string, parent?: URI): URI {
+    const name = `Untitled-${untitledResourceSequenceIndex++}${extension ?? ''}`;
+    if (parent) {
+        return parent.resolve(name).withScheme(UNTITLED_SCHEME);
+    }
+    return new URI().resolve(name).withScheme(UNTITLED_SCHEME);
 }
--- a/packages/core/src/common/resource.ts
+++ b/packages/core/src/common/resource.ts
@@ -327,6 +327,14 @@ export class UntitledResourceResolver implements ResourceResolver {
 
     protected readonly resources = new Map<string, UntitledResource>();
 
+    has(uri: URI): boolean {
+        if (uri.scheme !== UNTITLED_SCHEME) {
+            throw new Error('The given uri is not untitled file uri: ' + uri);
+        } else {
+            return this.resources.has(uri.toString());
+        }
+    }
+
     async resolve(uri: URI): Promise<UntitledResource> {
         if (uri.scheme !== UNTITLED_SCHEME) {
             throw new Error('The given uri is not untitled file uri: ' + uri);
@@ -341,8 +349,14 @@ export class UntitledResourceResolver implements ResourceResolver {
     }
 
     async createUntitledResource(content?: string, extension?: string, uri?: URI): Promise<UntitledResource> {
-        return new UntitledResource(this.resources, uri ? uri : new URI().withScheme(UNTITLED_SCHEME).withPath(`/Untitled-${untitledResourceSequenceIndex++}${extension ?? ''}`),
-            content);
+        if (!uri) {
+            let counter: number = 1; // vscode is started from 1
+            do {
+                uri = new URI().withScheme(UNTITLED_SCHEME).withPath(`/Untitled-${counter}${extension ?? ''}`);
+                counter++;
+            } while (this.has(uri));
+        }
+        return new UntitledResource(this.resources, uri!, content);
     }
 }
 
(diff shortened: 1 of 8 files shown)


packages/core/src/common/resource.ts
```diff
@@ -389,8 +403,8 @@ export class UntitledResource implements Resource {
     }
 }
 
-export function createUntitledURI(extension?: string, parent?: URI): URI {
-    const name = `Untitled-${untitledResourceSequenceIndex++}${extension ?? ''}`;
+export function createUntitledURI(extension?: string, parent?: URI, sequenceIndex?: number): URI {
```
msujew, line 406: Instead of expanding on this `createUntitledURI` function, we should move it to the `UntitledResourceResolver` class (deprecate the existing function). We could then move the `do ... while` loop in here to remove the need for it in the `createUntitledResource` method and the execute function for `CommonCommands.NEW_FILE`.

msujew (commented):
I believe we can greatly simplify this change and cut down on the logic duplication. See below on how we could accomplish this.

packages/core/src/common/resource.ts
yiliang114: Thank you for reviewing at my code and your professional advice. I have already made changes to simplify.

packages/plugin-ext/src/plugin/document-data.ts
```diff
@@ -127,7 +128,7 @@ export class DocumentDataExt {
 
         // Insert new lines & store lengths
         const newLengths = new Uint32Array(insertLines.length - 1);
-        for (let i = 1; i < insertLines.length; i++) {
+        for (let i = 1;i < insertLines.length;i++) {
```
colin-grant-work, line 131: I believe that when CI runs, this will cause a lint error. Please restore the original spacing, here and elsewhere in this file.

yiliang114: Sorry, it was my negligence. I has reverted the code here. Thank you.

packages/core/src/common/resource.ts
```diff
@@ -388,11 +410,3 @@ export class UntitledResource implements Resource {
...
     }
 }
-
-export function createUntitledURI(extension?: string, parent?: URI): URI {
-    const name = `Untitled-${untitledResourceSequenceIndex++}${extension ?? ''}`;
-    if (parent) {
-        return parent.resolve(name).withScheme(UNTITLED_SCHEME);
-    }
-    return new URI().resolve(name).withScheme(UNTITLED_SCHEME);
-}
```
msujew, line 398: Suggestion: Instead of deleting the function completely, we might want to deprecate it instead, to give downstream users a clear way for migration:

```ts
/**
 * @deprecated Since 1.27.0. Please use `UntitledResourceResolver.createUntitledURI` instead.
 */
export function createUntitledURI(extension?: string, parent?: URI): URI {
...
```

packages/core/src/common/resource.ts
```diff
@@ -341,8 +347,24 @@ export class UntitledResourceResolver implements ResourceResolver {
...
-        return new UntitledResource(this.resources, uri ? uri : new URI().withScheme(UNTITLED_SCHEME).withPath(`/Untitled-${untitledResourceSequenceIndex++}${extension ?? ''}`),
-            content);
+        if (!uri) {
+            uri = this.createUntitledURI(extension);
+        }
+        return new UntitledResource(this.resources, uri, content);
+    }
+
+    createUntitledURI(extension?: string, parent?: URI): URI {
+        let counter: number = 1; // vscode is started from 1
```
msujew, line 357: Minor: We can let TypeScript infer the type.
```suggestion
        let counter = 1; // vscode starts at 1
```

yiliang114:
What did I ignore caused two CI run failed? I try to find errors in log, get nothing useful but large empty section below limited content. 

image

packages/plugin-ext/src/plugin/document-data.ts
```diff
@@ -80,7 +81,7 @@ export class DocumentDataExt {
             this._document = {
                 get uri(): theia.Uri { return that.uri; },
                 get fileName(): string { return that.uri.fsPath; },
-                get isUntitled(): boolean { return that.uri.scheme === 'untitled'; },
+                get isUntitled(): boolean { return that.uri.scheme === UNTITLED_SCHEME; },
```
msujew, line 84: Same issue as above. You can use `Schemes.untitled` instead (imported from `uri-components`).

msujew (commented):
The code looks good already, however the changes to the plugin host code crash the runtime. See below for how to address this.

packages/plugin-ext/src/common/uri-components.ts
```diff
@@ -18,6 +18,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { UNTITLED_SCHEME } from '@theia/core/lib/common';
```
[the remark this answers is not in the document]
yiliang114: Yes, I think so. I has reverted these codes about `UNTITLED_SCHEME`. And try to run CI in my own theia repo,  it can run successful 

- [action1](https://github.com/yiliang114/theia/actions/runs/2575979658)
- [action2](https://github.com/yiliang114/theia/actions/runs/2575939827)

packages/plugin-ext/src/plugin/document-data.ts
yiliang114: In order to avoid unnecessary influence, roll back the code for the time being. Thank you again for your help @msujew @colin-grant-work

msujew (approved):
Thank you for your contribution, the changes look good to me 👍

PR #11353, opened 2022-06-28
Add `Thenable` type to `theia.d.ts`

### LINKED ISSUE #11214: Use Thenable instead of PromiseLike in API
### Feature Description:
Currently most of the Theia API uses `PromiseLike` and VSCode exports its own type `Thenable` and uses it. However, there a some usages of `Thenable` in the Theia API. For example [here](https://github.com/eclipse-theia/theia/blob/dbc574288c95108581d15849df498625d0d81c18/packages/plugin/src/theia.d.ts#L7003). IMHO the API should only use types, that are defined within the file (next to typescript native types obviously). On top of that this leads to problems in the `theia-vscode-comparator`.

We could either redefine the type in the API itself or just use the `PromiseLike` type consistently. I would prefer the second option, as having two of the "same" type seems weird. Would there be any objections or preferences to this?

sgraband:
#### What it does

As discussed in #11214 the `theia.d.ts` should look as close as possible
to the `vscode.d.ts`.
It is currently missing the `Thenable` type and using `PromiseLike` instead.
The goal of this PR is to unify this all to `Thenable`.
Adding the type will also fix and issue in the `vscode-theia-comparator`,
where the Thenable type could not be resolved.

Resolves #11214

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - [core, monaco] refactored theme initialization to occur within application lifecycle rather than at import time. [#11213](https://github.com/eclipse-theia/theia/pull/11213)
 - [plugin] Support optional property `TaskPresentationOptions#clear` [#11298](https://github.com/eclipse-theia/theia/pull/11298) - Contributed on behalf of STMicroelectronics
 - [plugin] added support for debuggers running in the frontend [#10748](https://github.com/eclipse-theia/theia/pull/10748)
+- [plugin] Added `Thenable` type to API and replaced `PromiseLike` with `Thenable` [#11352](https://github.com/eclipse-theia/theia/pull/11352) - Contributed on behalf of STMicroelectronics
 
 <a name="breaking_changes_1.27.0">[Breaking Changes:](#breaking_changes_1.27.0)</a>
 
--- a/packages/plugin/src/theia-proposed.d.ts
+++ b/packages/plugin/src/theia-proposed.d.ts
@@ -161,7 +161,7 @@ export module '@theia/plugin' {
          * Get the keybindings associated to commandId.
          * @param commandId The ID of the command for which we are looking for keybindings.
          */
-        export function getKeyBinding(commandId: string): PromiseLike<CommandKeyBinding[] | undefined>;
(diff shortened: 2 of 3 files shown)


colin-grant-work (approved):
Looks good to me, and should remove a number of hidden `any`'s that had snuck into `theia.d.ts` I notice that there are still a few instances of `PromiseLike` in `theia-proposed.d.ts` that we might consider converting to `Thenable`, I'm not sure how downstream applications would consume that - I don't believe it's part of our current comparator - so it is likely low priority.

packages/plugin-ext/src/plugin/plugin-context.ts
```diff
@@ -284,7 +284,7 @@ export function createAPIFactory(
                 return commandRegistry.registerCommand(command, handler, thisArg);
             },
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            executeCommand<T>(commandId: string, ...args: any[]): PromiseLike<T | undefined> {
+            executeCommand<T>(commandId: string, ...args: any[]): Thenable<T | undefined> {
```
paul-marechal, line 287: I don't think we should use `Thenable` in this file. I had trouble understanding where it even came from, because it isn't imported from `theia.d.ts`. Turns out it's a side effect of using `vscode-jsonrpc` and somehow the symbol is available here.

Long story short, all those sites are implementations so I think we can/should use `Promise` here.

paul-marechal (commented):
Comment apart, this is good change!

packages/plugin-ext/src/plugin/plugin-context.ts
sgraband: Ah yes i must have overlooked this! Will change! Do you have an opinion on the usage of `PromiseLike`/`Thenable` in the `theia-proposed.d.ts`? Then i will just change this as well, if needed.

sgraband: Reverted the changes for the `plugin-context.ts` file and changed the usages of `PromiseLike` to `Thenable` in `theia-proposed.d.ts`.

Issue #10885, opened 2022-03-15
Regular translation update


msujew:
I want to use this issue to have discussions on how to perform regular updates to our translation files.

### Proposal

My current idea is to use the CI to create a commit on a separate branch for every commit to `master`. This branch is always rebased to the latest `master` and runs the `theia nls-extract` and `theia nls-localize` commands to produce updated localization files. These can be merged/cherry-picked into the release-branch and will be then available for every release.

msujew:
@vince-fugnitto regarding API-keys: I would simply store my API-key as a secret in the GitHub repo, which is then used on every GitHub actions workflow. So, from that perspective, it doesn't matter whether it's on a per-PR or per-release basis, since they use the stored API-key anyway. Contributors to Theia shouldn't have to touch that anyway.

vince-fugnitto:
@msujew since we have the free API afaik and we were to do the per pr-basis we might need to check that we do not hit the limit for the month. I assume we should only translate the delta and not the whole codebase again?

msujew:
@vince-fugnitto the command always works on a delta, since I anticipated this ;)

msujew:
@vince-fugnitto Yeah, I will create a PR for this. Didn't have time to take care of this issue yet.

PR #11354, opened 2022-06-28
nls: localize `no-Info` messages

vince-fugnitto:
#### What it does

The commit sets up the localizations for `no-Info` messages which are used by views when they do not provide any content but instead display a default message. Previously the label was hardcoded and not set up for localization.

#### How to test

The code sets up localization which should work when we switch languages after we run a translation of our code-base (https://github.com/eclipse-theia/theia/issues/10885).

1. start the application
2. switch language using `configure display language`
3. confirm that the three views are not localized but should be after https://github.com/eclipse-theia/theia/issues/10885.

#### Review checklist

#### Reminder for reviewers


msujew (approved):
LGTM 👍

PR #11377, opened 2022-06-30
Update Developing.md

fipro78:
Update npm version in Building on Windows section

### DIFF
--- a/doc/Developing.md
+++ b/doc/Developing.md
@@ -509,7 +509,7 @@ etc.) by opening `packages/<package name>/coverage/index.html`.
 
  - Install [`scoop`](https://github.com/lukesampson/scoop#installation).
  - Install [`nvm`](https://github.com/coreybutler/nvm-windows) with scoop: `scoop install nvm`.
- - Install Node.js with `nvm`: `nvm install 12.14.1`, then use it: `nvm use 12.14.1`. You can list all available Node.js versions with `nvm list available` if you want to pick another version.
+ - Install Node.js with `nvm`: `nvm install 16.15.1`, then use it: `nvm use 16.15.1`. You can list all available Node.js versions with `nvm list available` if you want to pick another version.
  - Install `yarn`: `scoop install yarn`.
  - If you need to install `windows-build-tools`, see [`Installing Windows Build Tools`](#installing-windows-build-tools).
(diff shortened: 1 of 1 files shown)


vince-fugnitto (commented):
@fipro78 thank you for your contribution :+1: 

I've updated the pull-request to respect the template, but can you also update the commit, it seems as though you've mentioned `npm version` and not `node version` which is misleading:

```
From a1b971c867f6ec8510a8acb661e0a5e9a5d61f57 Mon Sep 17 00:00:00 2001
From: Dirk Fauth <dirk.fauth@googlemail.com>
Date: Thu, 30 Jun 2022 10:17:40 +0200
Subject: [PATCH] Update Developing.md

Update npm version in Building on Windows section
```

fipro78:
@vince-fugnitto Sorry but I don't understand your comment. What do you mean with `node version`? And where should I change something? It was simply the wrong version in the command.

fipro78:
@vince-fugnitto 
Ah, the commit message. Thanks. But I suppose to do this I have to checkout my branch locally and do an amend to update the commit message, right? That actually introduces quite some work for a minimal change.

Don't misunderstand my comment, I don't want to blame anybody or anything. I just learn that the "cool" feature of using an online IDE only for minimal changes breaks in case of updates on review comments. While I simply edited in the browser and created the PR automatically, I now have to checkout, amend and push via commandline from a local computer. And for this I need to use a different computer. So the online-only workflow for simple contributions has some flaws. It looks like it is easier to close this one and create a new PR. Which would actually not a very good way to work IMHO. :(

vince-fugnitto:
@fipro78 I can do it when I perform `squash and merge` no worries.
I'm not exactly sure what you're referring to regarding the online IDE you mentioned, or your setup so I can't really comment on that.

vince-fugnitto (approved):
LGTM :+1:

fipro78:
@vince-fugnitto 
Thank you very much! I meant, I simply clicked "edit" on the documentation page. This automatically created a local branch from which I was able to create the PR. So for me it was a very easy and quick way to make a simple contribution for which I do not need a local checked out branch. Every action on my side was online on GitHub. This is what I referred to as an online IDE. I did not need anything locally, which is actually the use case for going into the direction of VS Code or Theia in combination with a workspace management. 

As I said, I didn't wanted to be offending in any way. I just noticed this flaw in the process. If I would have done everything correctly from the beginning, this would not have been an issue.

Next time I will try the local process to be able to react on review comments.

PR #11331, opened 2022-06-23
Uplift Monaco to VSCode 1.67.2

colin-grant-work:
#### What it does

Uplifts `@theia/monaco-editor-core` to the state of VSCode@1.67.2 and makes a few adjustments. In particular, it updates the `vscode-textmate` dependency to `7.0.1` to match VSCode in order to provide the accustomed theme experience, and it replaces `onigasm` with `vscode-oniguruma` for the same reason. The former change required some reorganization of code to account for breaking changes in `vscode-textmate`.

#### How to test

1. Test your accustomed workflows in the application and make sure they're still working.
2. Test bracket pair colorization in various languages.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #11393, opened 2022-07-07
Improve Plugin Preference Access

### LINKED ISSUE #10004: Introduce Type `ConfigurationScope`
### Feature Description:
Add the `ConfigurationScope` type and use it in `vscode.workspace.getConfiguration(...)` and `ConfigurationChangedEvent#affectsConfiguration()`.

colin-grant-work:
#### What it does

Fixes #10004
Depends on #11331
This PR updates various aspects of preference access and setting from plugins.
 - Adjusts PreferenceRegistry to use Monaco code for `Configuration` and `ConfiguratioModel`.
 - Enables access and setting of language-overridden preferences from plugins according to VSCode API (#10004)
 - Updates `affectsConfiguration` check to respect override options (language and URI)

#### How to test

0. Assert that the tests pass - most of the functionality related to preference access and `affectsConfiguration` has new tests.
0. Install [this plugin](https://github.com/eclipse-theia/theia/files/9067216/extension-preference-access-0.0.1.zip) ([source](https://github.com/colin-grant-work/demo-extensions/tree/main/extension-preference-access))
1. Open a single-root workspace.
2. Run the command `Listen for config changes`
3. Add and remove tasks, observe that a toast informs you that you have done so. 
4. Delete your `tasks.json` file.
5. Observe that a toast registers for changes to `tasks` and `tasks.tasks`
> On `master`, the deletion of the file would not be registered by plugins as a change in the section.
6. Modify the value of `editor.fontSize` in your workspace; modify it for language overrides, including `[rust]`
7. Observe that your changes are accurately reported.
8. Clear the values of `editor.fontSize` in your workspace.
> Run the command 'Stop listening for config changes' now if you want to see fewer toasts in the next step.
9. Run the command 'Run config test (no language override)'
10. It should modify the value for `editor.fontSize` in your workspace and report that the effective value has changed for non-overridden and for `rust`.
11. Run the command 'Run config test (with language override)'
12. Same as 10., but with different numerical values.
13. Run the command 'Run config test (with forced language override)'
14. It should create an overridden value for `editor.fontSize` for the `rust` language, and the report should show that only `rust` and not un-overridden `editor.fontSize` was affected
15. Run the command 'Run config test (with language override)' again.
16. This time, because an override already exists for `rust` after (13), you should see a toast similar to 14 - only the value for `rust` should be modified.

#### Review checklist

#### Reminder for reviewers


colin-grant-work:
@tsmaeder, in the past, I believe you expressed concern about using Monaco code in the plugin host. This PR does so: it uses the Monaco versions of `Configuration` and `ConfigurationModel` rather than versions that we had copied when the plugin system was first written. Those classes reside in `common` folders of Monaco and they, or extensions of them, are also used in the VSCode plugin host. Do you believe that this use is reasonable, or that we should avoid importing directly from Monaco and copy / reimplement these classes?

tsmaeder:
@colin-grant-work <sigh> like all interesting questions, this one does not have a straightforward answer. I don't mind having VS Code stuff in the plugin host per se. The problem is that as we use more code outside the editor proper, the larger the change becomes when we uplift Monaco to a new version. The reason this is more of a problem with Monaco than with other libraries is that we have to update _all_ Monaco modules at the same time. Is this the direction we want to move the code in? The upside is that we can reuse stuff from VS Code.
As for this PR, I would propose to keep the copied classes unless that's a problem. Now that we have more experience with the improved uplift procedures (with the new way we consume VS Code modules), we should discuss whether we think it's proper to use stuff from VS code in our code base more broadly with the community. If we do, the two classes can be easily nuked in a follow-up PR.

packages/core/src/browser/preferences/preference-service.ts
```diff
@@ -459,20 +467,20 @@ export class PreferenceServiceImpl implements PreferenceService {
         return Number(value);
     }
 
-    inspect<T>(preferenceName: string, resourceUri?: string): PreferenceInspection<T> | undefined {
-        const defaultValue = this.inspectInScope<T>(preferenceName, PreferenceScope.Default, resourceUri);
-        const globalValue = this.inspectInScope<T>(preferenceName, PreferenceScope.User, resourceUri);
-        const workspaceValue = this.inspectInScope<T>(preferenceName, PreferenceScope.Workspace, resourceUri);
-        const workspaceFolderValue = this.inspectInScope<T>(preferenceName, PreferenceScope.Folder, resourceUri);
+    inspect<T>(preferenceName: string, resourceUri?: string, forceLanguageOverride?: boolean): PreferenceInspection<T> | undefined {
```
tsmaeder, line 470: The name `forceLanguageOverride` confused me a bit at first. What it means is "don't return the general value if the key is language specific", right? I haven't really thought of a good name, though.

packages/monaco/README.md
```diff
@@ -19,6 +19,62 @@ This includes:
...
+process, the steps for undertaking a Monaco uplift are outlined here.
+
+### Setting up the VSCode side
+
+1. Clone the VSCode repo and make sure you have the following remotes:
+ - https://github.com/microsoft/vscode.git - the official VSCode repo.
+ - https://github.com/theia-ide/vscode.git - Theia's fork.
+2. Find the latest release tag in the official VSCode repo, and the most recent uplift branch in the Theia fork.
+ > At the time of this writing the latest release tag is `1.67.2`, and the uplift branch is `monaco-uplift-2022-6`
+3. Check out the release tag, cherry pick the tip of the uplift branch, and resolve any conflicts.
```
tsmaeder, line 35: Usually, one cherry-picks commits, not branches. Which commits do you mean, concretely?

packages/monaco/src/browser/textmate/monaco-textmate-frontend-bindings.ts
```diff
@@ -40,6 +41,35 @@ export default (bind: interfaces.Bind, unbind: interfaces.Unbind, isBound: inter
     bindContributionProvider(bind, LanguageGrammarDefinitionContribution);
     bind(TextmateRegistry).toSelf().inSingletonScope();
     bind(MonacoThemeRegistry).toSelf().inSingletonScope();
+    bind(TextmateRegistryFactory).toFactory(({ container }) => (theme?: ThemeMix) => {
```
tsmaeder, line 44: What's behind this additional binding?

tsmaeder: Ah, I see: moved from below.

tsmaeder (commented):
Just some comments after a superficial look at the change.

packages/core/src/browser/preferences/preference-service.ts
```diff
@@ -171,10 +171,18 @@ export interface PreferenceService extends Disposable {
      *
      * @param preferenceName the preference identifier.
      * @param resourceUri the uri of the resource for which the preference is stored.
+     * @param forceLanguageOverride if `true` and `preferenceName` is a language override, only values for the specified override will be returned.
+     * Otherwise, values for the override will be returned where defined, and values from the base preference will be returned otherwise.
      *
      * @return an object containing the value of the given preference for all scopes.
      */
-    inspect<T extends JSONValue>(preferenceName: string, resourceUri?: string): PreferenceInspection<T> | undefined;
+    inspect<T extends JSONValue>(preferenceName: string, resourceUri?: string, forceLanguageOverride?: boolean): PreferenceInspection<T> | undefined;
```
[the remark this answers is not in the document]
colin-grant-work: `forceLanguageOverride` is referred to everywhere this method delegates to `inspectInScope`:

```typescript
const defaultValue = this.inspectInScope<T>(preferenceName, PreferenceScope.Default, resourceUri, forceLanguageOverride);
         const globalValue = this.inspectInScope<T>(preferenceName, PreferenceScope.User, resourceUri, forceLanguageOverride);
         const workspaceValue = this.inspectInScope<T>(preferenceName, PreferenceScope.Workspace, resourceUri, forceLanguageOverride);
         const workspaceFolderValue = this.inspectInScope<T>(preferenceName, PreferenceScope.Folder, resourceUri, forceLanguageOverride);
```

And it's used in `inspectInScope` to decide whether to fall back to a non-overriden value here:

```typescript
if (value === undefined && !forceLanguageOverride) {
             const overridden = this.overriddenPreferenceName(preferenceName);
             if (overridden) {
                 return this.doInspectInScope(overridden.preferenceName, scope, resourceUri);
```

packages/core/src/browser/preferences/preference-service.ts
colin-grant-work: I could reverse the polarity and rename it something like `acceptDefault`. There's no exactly antonym for `language overridden`, though...

packages/monaco/README.md
colin-grant-work: In the git CLI, `git cherry-pick <branch name>` is equivalent to cherry-picking the tip of that branch, which is the intended effect of this step. See [this example](https://git-scm.com/docs/git-cherry-pick#Documentation/git-cherry-pick.txt-codegitcherry-pickmastercode).

JonasHelming:
@colin-grant-work Is this ready for a re-review?

colin-grant-work:
@JonasHelming, yes - I believe I have addressed Thomas' comments, apart from the naming, and that review did not check the functionality, as far as I'm aware.

msujew (approved):
I'm happy with the changes, good job 👍 

- [x] Language scoped preferences work as expected (i.e. doesn't bleed into the "normal" preferences, are applied correctly and instantly)
- [x] Adding a `tasks.json` and changes within are correctly transmitted to the plugin
- [x] Running the different plugin commands for (non)-language-override preferences works as expected

Issue #11392, opened 2022-07-07
Incorrect conversions from `null` to `undefined` in new RPC


colin-grant-work:
### Bug Description:

In certain cases, `null` values from the backend are appearing on the frontend as `undefined`.

### Steps to Reproduce:

1. Install and activate the `vscode-lldb` plugin.
2. See error mentioned [here](https://github.com/eclipse-theia/theia/issues/9136#issuecomment-1175400641): Failure to activate because of problem in a method `startRpcServer`.
> The error does not occur in `1.26` (without new RPC) but does in `1.27` with new RPC.

---

You can also use [this plugin](https://github.com/eclipse-theia/theia/files/9067216/extension-preference-access-0.0.1.zip) ([source](https://github.com/colin-grant-work/demo-extensions/tree/main/extension-preference-access)). Assuming you have not set the preference `peaSoup.secretIngredients`, running the command `Access a (probably) non-existent preference` should show a toast with the default value `null`, but on master it shows `[]`.

### Additional Information

The relevant preference is defined [here](https://github.com/vadimcn/vscode-lldb/blob/54d4fbb2dead83e98bd44d07057b67efc4305531/package.json#L455-L463) with a default of `null`. However, when the plugin's data is transmitted to the frontend, that `null` becomes `undefined`, and we enter this code:

https://github.com/eclipse-theia/theia/blob/84b317ae4b6610b8e23ae5b87d845c4abee78116/packages/core/src/browser/preferences/preference-contribution.ts#L248-L271

Because the preference is typed as `["object", "null"]`, that method returns a default of `{}` rather than the configured default of `null`.

I have confirmed that `null` is present [here](https://github.com/eclipse-theia/theia/blob/84b317ae4b6610b8e23ae5b87d845c4abee78116/packages/plugin-ext/src/hosted/node/plugin-service.ts#L161), but becomes `undefined` when received [here](https://github.com/eclipse-theia/theia/blob/84b317ae4b6610b8e23ae5b87d845c4abee78116/packages/plugin-ext/src/hosted/browser/hosted-plugin.ts#L344).

- Operating System: MacOS 12.4
- Theia Version: 1.27.0

colin-grant-work:
Attn: @JonasHelming, @sgraband

JonasHelming:
@tortmayr

Issue #10841, opened 2022-03-04
Support memory-related features introduced in VS Code


thegecko:
### Feature Description:

In the January 2022 release, VS Code introduced support for [memory related features](https://code.visualstudio.com/updates/v1_64#_debugger-extension-authoring), specifically [viewing and editing binary data for variables](https://code.visualstudio.com/updates/v1_64#_viewing-and-editing-binary-data).

During debug, if the adapter supports `readMemoryRequest`, `writeMemoryRequest` and a variable contains the `memoryReference` property, then an inline button appears to allow viewing the memory.

In VS Code this directs the user to install and use the [hex editor extension](https://open-vsx.org/extension/ms-vscode/hexeditor).

As we have similar functionality in the form of [memory inspection](https://github.com/eclipse-theia/theia-cpp-extensions/tree/master/packages/cpp-debug/src/browser/memory-widget), I propose Theia mimics this functionality, but allows the developer to choose which memory viewer extension to contribute for the UI

PR #11394, opened 2022-07-07
Integrate Memory Inspector in Theia repository

colin-grant-work:
#### What it does

This PR transfers the Memory Inspector functionality from [the Theia CPP Extensions](https://github.com/eclipse-theia/theia-cpp-extensions) to the Theia repository. Relative to the current state of that repository, this PR makes some minor changes necessary to operate with Theia@1.27.0 and transfers some functionality that is not fully determined by the Debug Adapter Protocol into the `MemoryProviderService`, where it can be more easily adjusted for adapters with specific known properties.

#### How to test

1. With the CDT-GDB adapter, the instructions from [this PR](https://github.com/eclipse-theia/theia-cpp-extensions/pull/119) still apply
2. To demonstrate the functionality with another debug adapter: (less convenient than GDB, because it supports fewer reference options)
3. Create and open a new Rust project or use [this one](https://github.com/colin-grant-work/demo-extensions/tree/main/struct_memory)
4. In your workspace, make sure that the preference `lldb.rpcServer` is set to `null` to work around #11392
5. Install `Rust Analyzer` and `CodeLLDB`
6. (If using the example project linked above) open `main.rs` and put a breakpoint on line 12.
7. Use the code lenses provided by Rust Analyzer to start a debug session from the testson lines 16-25.
8. When you hit the breakpoint, the terminal in which the program is running will have printed two addresses.
9. Open the memory inspector from the right panel and input one of the addresses into the address selector and hit `Go`. The memory location should load, and it should match the data of the struct whose address you entered.

#### Review checklist

#### Reminder for reviewers


JonasHelming:
Just out of interest: Did you consider moving this to CDT.cloud instead of the Theia main repo?

colin-grant-work:
@JonasHelming, I've made this PR for two reasons: I believe that having the code in this repo will make it easier to ensure its compatibility with latest Theia, and I believe that it's advantageous with respect to the features requested in #10841. What would be the benefit of placing it in CDT.cloud?

JonasHelming:
Basically to keep the core Theia project as slim as possible. Most components would benefit from beeing moved to the core project/repo in terms of ensuring compatibility. Are the people maintaining this component the same team that are core Theia developers? Again: I am not opposed to move this to Theia core, I just wanted to raise the question as we had a tendency to try to keep the core repo as small as possible.

One example that you could aim at when hosting it at CDT.cloud would be to ensure compatibility for community releases only.
Again, up to you!

JonasHelming:
Thinking about this again: the memory inspector is not necessarily bound to C/C++, correct? This would be an argument in favor to keep it in Theia.

colin-grant-work:
@JonasHelming, exactly. I've been waiting till I could get it working on at least one non-C(++) debug adapter before trying to move it here, and I was finally able to get it working for Rust. :-)

packages/memory-inspector/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "compile": "theiaext compile",
+    "lint": "theiaext lint",
+    "test": "theiaext test",
+    "watch": "theiaext watch"
+  },
+  "dependencies": {
+    "@theia/core": "1.27.0",
+    "@theia/debug": "1.27.0",
+    "long": "^4.0.0",
+    "vscode-debugprotocol": "^1.48.0"
```
vince-fugnitto, line 33: Do we need to also update `vscode-debugprotocol` in `@theia/debug` and `@theia/plugin-ext` to follow this version:

https://github.com/eclipse-theia/theia/blob/b000f9603107cf20e46cb989505b1e2dbb546808/packages/debug/package.json#L25

https://github.com/eclipse-theia/theia/blob/b000f9603107cf20e46cb989505b1e2dbb546808/packages/plugin-ext/package.json#L45

packages/memory-inspector/src/browser/memory-inspector-frontend-contribution.ts
```diff
@@ -0,0 +1,301 @@
...
+            {
+                id: 'editor.foreground.light',
+                defaults: {
+                    dark: Color.transparent('editor.foreground', ONE_HALF_OPACITY),
+                    light: Color.transparent('editor.foreground', ONE_HALF_OPACITY),
+                },
+                description: 'A lighter version of the text in the editor.',
+            },
+        );
+    }
```
vince-fugnitto, line 300: We should maybe revisit these colors to see if they are still needed, some of them mention other components such as `diff` and `editor` and perhaps they belong there instead. Else we can rename them to be more meaningful for `memory-inspector`.

packages/memory-inspector/src/browser/wrapper-widgets/memory-dockpanel-placeholder-widget.tsx
```diff
@@ -0,0 +1,46 @@
...
+                <i className='memory-view-icon toolbar' />
+                {' '}
+                icon to add a new memory view or the
+                {' '}
+                <i className='register-view-icon toolbar' />
+                {' '}
+                icon to add a register view.
+            </div>
+        );
+    }
```
vince-fugnitto, line 45: We can probably clean up this method and localize if possible.

vince-fugnitto (commented):
I added a second commit to localize everything I could find, and cleaned up the code slightly.
I still have some comments however.

packages/memory-inspector/package.json
colin-grant-work: Because they're both caret versions, we're currently getting a high-enough version (1.51) in our `yarn.lock`, but it's probably best to be explicit.

msujew (commented):
I haven't tested it yet, but here are my obligatory i18n comments :)

Also please rebase on latest `master` to get access to automatic translations for your changes.

packages/memory-inspector/src/browser/memory-provider/cdt-gdb-memory-provider.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    override supportsVariableReferenceSyntax(session: DebugSession, currentLevel?: DebugVariable): boolean {
+        if (this.canHandle(session)) {
+            if (!currentLevel) {
+                return false;
+            }
+            while (currentLevel.parent instanceof DebugVariable) {
+                currentLevel = currentLevel.parent;
+            }
+            return currentLevel.parent instanceof DebugScope && currentLevel.parent['raw'].name === 'Local';
```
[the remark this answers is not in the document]
colin-grant-work: That's a fair observation. This class is intended to be tightly coupled to CDT-GDB adapter, and at the moment, it does not localize any of its text, and adapter returns are mapped fairly directly into the display in the Variables widget, so this should work, but it's certainly not as robust as might be ideal. To my knowledge, there aren't hard conventions about the naming of variable scopes - an adapter could return any labels it wants - so I don't see a very good way around this, at the moment.

JonasHelming:
@colin-grant-work Is this ready for a re-review?

packages/memory-inspector/src/browser/memory-widget/memory-options-widget.tsx
```diff
@@ -0,0 +1,726 @@
...
+        if (element) {
+            this.headerInputField = element;
+            element.focus();
+        }
+    };
+
+    protected renderToolbar(): React.ReactNode {
+        return (
+            <div className='memory-widget-toolbar'>
+                {this.memoryWidgetOptions.dynamic !== false && (
```
martin-fleck-at, line 485: It would be really great if this could also be moved into a separate method to support easier overriding also allow the user to add something to the toolbar before the lock.

packages/memory-inspector/src/browser/memory-widget/memory-options-widget.tsx
```diff
@@ -0,0 +1,726 @@
...
+                            }
+                            onClick={this.toggleAutoUpdate}
+                            onKeyDown={this.toggleAutoUpdate}
+                            role='button'
+                            tabIndex={0}
+                        />
+                    </div>
+                )}
+                {this.renderEditableTitleField()}
+                <div className='toggle-settings-container'>
```
martin-fleck-at, line 502: It would be really great if this could also be moved into a separate method to support easier overriding also allow the user to add something to the toolbar after the settings.

packages/memory-inspector/src/browser/memory-widget/memory-options-widget.tsx
```diff
@@ -0,0 +1,726 @@
...
+    };
+
+    protected saveHeaderInputValue = (e: React.MouseEvent<HTMLDivElement> | React.KeyboardEvent<HTMLDivElement>): void => {
+        const isMouseDown = !('key' in e);
+        const isSaveKey = 'key' in e && e.key === 'Enter';
+        const isCancelKey = 'key' in e && e.key === 'Escape';
+        e.stopPropagation();
+        if (isMouseDown || isSaveKey || isCancelKey) {
+            if (!isCancelKey && this.headerInputField) {
+                this.title.label = this.headerInputField.value;
```
martin-fleck-at, line 466: Would it be possible to move this into a separate protected method (like `updateHeader`) so that it is easier to hook into that for some custom behavior?

packages/memory-inspector/src/browser/memory-widget/memory-options-widget.tsx
```diff
@@ -0,0 +1,726 @@
...
+            this.update();
+        }
+    }
+
+    protected getUserError(err: unknown): string {
+        return err instanceof Error ? err.message : nls.localize('theia/memory-inspector/memory/userError', 'There was an error fetching memory.');
+    }
+
+    protected async getMemory(memoryReference: string, count: number, offset: number): Promise<Interfaces.MemoryReadResult> {
+        const result = await this.memoryProvider.readMemory({ memoryReference, count, offset });
```
martin-fleck-at, line 672: In our customization we had to extend the read memory arguments so it would have made things a bit easier if that call were in a separate protected method.

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
```diff
@@ -0,0 +1,324 @@
...
+            try {
+                await this.memoryProvider.writeMemory(edit);
+            } catch (e) {
+                console.log('Problem writing memory with arguments', edit, '\n', e);
+            }
+        }
+        this.memoryEditsCompleted.resolve();
+    };
+
+    private createUniqueEdits(): Array<DebugProtocol.WriteMemoryArguments> {
```
martin-fleck-at, line 201: Could you please make this method protected?

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
```diff
@@ -0,0 +1,324 @@
...
+            ? super.getWrapperHandlers()
+            : {
+                onClick: this.handleTableClick,
+                onContextMenu: this.handleTableRightClick,
+                onKeyDown: this.handleTableInput,
+                onMouseMove: this.handleTableMouseMove,
+            };
+    }
+
+    private handleTableClick = (event: React.MouseEvent): void => {
```
martin-fleck-at, line 228: Could you please make this field protected?

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
```diff
@@ -0,0 +1,324 @@
...
+    protected override doHandleTableRightClick(event: React.MouseEvent): void {
+        const target = event.target as HTMLElement;
+        if (target.classList?.contains('eight-bits')) {
+            this.highlightedField = hexStrToUnsignedLong(target.getAttribute('data-id') ?? '-0x1');
+        }
+        super.doHandleTableRightClick(event);
+    }
+
+    // eslint-disable-next-line max-lines-per-function,complexity
+    private handleTableInput = (event: React.KeyboardEvent): void => {
```
martin-fleck-at, line 246: Could you please make this field protected?

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
```diff
@@ -0,0 +1,324 @@
...
+        const valueWasChanged = newValue !== oldValue;
+        if (valueWasChanged) {
+            this.pendingMemoryEdits.set(this.highlightedField.toString(), newValue);
+        }
+        if (valueWasChanged || !this.highlightedField.equals(initialHighlight)) {
+            this.update();
+        }
+    };
+
+    private isInBounds(candidateAddress: Long): boolean {
```
martin-fleck-at, line 319: Could you please make this method protected?

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
```diff
@@ -0,0 +1,324 @@
...
+    protected submitMemoryEdits = async (): Promise<void> => {
+        this.memoryEditsCompleted = new Deferred();
+        for (const edit of this.createUniqueEdits()) {
+            try {
+                await this.memoryProvider.writeMemory(edit);
+            } catch (e) {
+                console.log('Problem writing memory with arguments', edit, '\n', e);
+            }
+        }
+        this.memoryEditsCompleted.resolve();
```
martin-fleck-at, line 198: I noticed that when I submit edits there is no update on the view by default. We do resolve the edits which means that the next time the memory is changed it will update accordingly and the pending edits are cleared (`handleMemoryChange`). But this requires a specific user action. At the moment, the pending changes just remain in the UI. Should we trigger an explicit update from here or what is the expected flow after the user submits the edits?

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
```diff
@@ -0,0 +1,324 @@
...
+        this.update();
+    }
+
+    protected submitMemoryEdits = async (): Promise<void> => {
+        this.memoryEditsCompleted = new Deferred();
+        for (const edit of this.createUniqueEdits()) {
+            try {
+                await this.memoryProvider.writeMemory(edit);
+            } catch (e) {
+                console.log('Problem writing memory with arguments', edit, '\n', e);
```
martin-fleck-at, line 195: I think it would be great if we could show the errors below the pending edits within the table footer - similar to how read errors are shown in the options widget. What do you think?

martin-fleck-at (commented):
Hi @colin-grant-work and everyone, thank you for the great work!

I had a look at this PR for an adopter where we customized the memory view to stick on a particular thread. The design of this component is really thought-out and it was easy to do most of the necessary customizations. Still, I thought I'd share some of my observations with you - mostly about making things a bit easier to override. Please note that my comments are just some minor things that should not prevent this PR from being merged.

martin-fleck-at:
@vince-fugnitto @colin-grant-work: Thanks again for the great work! Is there any chance this PR can make it in the upcoming release? If there is any way I can help out, please let me know.

colin-grant-work:
@martin-fleck-at, I'll devote some time to this today to address the outstanding comments.

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
colin-grant-work: This is a bit tricky because the debug session doesn't emit any events when memory changes - we just get a response. When we implemented this internally, we just had the debug adapter send a `StoppedEvent` after successfully updating memory, but that was a bit of a workaround, and I didn't include it when implementing the write memory request [upstream](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/d8c6ab55a03025f6facf917cca8eaed69b19fe76/src/GDBDebugSession.ts#L1373-L1408) in the CDT-GDB adapter. I think your suggestion of triggering a refresh cycle is probably the best way to handle it.

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
colin-grant-work: Done.

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
colin-grant-work: I've modified the code here so that if a write succeeds we refresh the memory, but we don't clear any pending edits.

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
colin-grant-work: Done.

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
colin-grant-work: Done.

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
colin-grant-work: Done.

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
colin-grant-work: Done.

packages/memory-inspector/src/browser/memory-widget/memory-options-widget.tsx
colin-grant-work: Done.

packages/memory-inspector/src/browser/memory-widget/memory-options-widget.tsx
colin-grant-work: Done.

packages/memory-inspector/src/browser/utils/memory-widget-manager.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    protected setCanCompare(): void {
+        this._canCompare = this.availableWidgets.filter(widget => !RegisterWidget.is(widget) && !MemoryDiffWidget.is(widget)).length > 1;
+    }
+
+    async createNewMemoryWidget<T extends MemoryWidget>(kind: 'register' | 'memory' = 'memory'): Promise<T> {
+        this.widgetDisplayId = this._availableWidgets.size !== 0 ? this.widgetDisplayId + 1 : 1;
+        const options: MemoryWidgetOptions = { identifier: this.createdWidgetCount += 1, displayId: this.widgetDisplayId };
+        const widgetId = kind === 'memory'
```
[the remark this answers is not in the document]
colin-grant-work: This is reasonable, but doing it right would require a bit of a refactor - I'll leave this one out for now.

packages/memory-inspector/src/browser/memory-widget/memory-options-widget.tsx
colin-grant-work: Done.

packages/memory-inspector/src/browser/memory-widget/memory-options-widget.tsx
colin-grant-work: Done.

colin-grant-work:
@vince-fugnitto, @martin-fleck-at, I've addressed most of y'all's comments - since there's some interest in getting this into the release, would you mind taking another look?

packages/memory-inspector/src/browser/wrapper-widgets/memory-dockpanel-placeholder-widget.tsx
colin-grant-work: Cleaned up, but not localized. Not quite sure what to do with the fact that it's broken up by the icons - the pieces in between aren't meaningful enough to pass to an automated translation tool. @msujew, any suggestions?

packages/memory-inspector/src/browser/memory-inspector-frontend-contribution.ts
colin-grant-work: Renamed.

packages/memory-inspector/package.json
colin-grant-work: I've updated the entries in our various `package.json`. They've deprecated this package name, so I've updated that, as well, with the unfortunate consequence that the new package name isn't part of the 3PP baseline, apparently.

packages/memory-inspector/src/browser/utils/memory-widget-manager.ts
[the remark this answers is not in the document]
martin-fleck-at: Sure, you can keep the effort as low as possible. I just thought something simple along the lines of:
```
protected mapKindToWidgetId(kind: 'register' | 'memory' = 'memory') {
   return kind === 'memory' ? MemoryWidget.ID : RegisterWidget.ID;
}
```

So in case you want the editable memory widget you could simply do:

```
protected override mapKindToWidgetId(kind: 'register' | 'memory' = 'memory') {
   return kind === 'memory' ? EditableMemoryWidget.ID : RegisterWidget.ID;
}
```

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
```diff
@@ -0,0 +1,353 @@
...
+                    </button>}
+                    {showButtons && <button
+                        className='theia-button main'
+                        onClick={this.submitMemoryEdits}
+                        type='submit'
+                        title={nls.localize('theia/memory-inspector/editable/apply', 'Apply Changes')}
+                    >
+                        {nls.localize('theia/memory-inspector/editable/apply', 'Apply Changes')}
+                    </button>}
+                    {!!this.writeErrorInfo && <div className='memory-edit-error'>
```
martin-fleck-at, line 99: The way this element is styled causes some unexpected rendering for me:

image

Due to the absolute positioning, the error div only "hides" the buttons but also overlays the 'Load...' message for me.  Do you see the same behavior? I think we have two options here:
- Show the error simply below the buttons
- Only render the buttons when there is no error or fix the overlay by ensuring that it does not overlap with the 'Load...' message. The easiest option for me was to change the style from `bottom: 0` to `top: 0`.
 
But maybe this is just caused by my setup and it can be easily fixed with CSS so it should not be a blocker if it looks good for you.

packages/memory-inspector/src/browser/style/index.css
```diff
@@ -0,0 +1,753 @@
...
+}
+
+.memory-edit-button-container .memory-edit-error {
+    position: absolute;
+    width: 100%;
+    bottom: 0;
+    box-sizing: border-box;
+    white-space: normal;
+    padding: var(--theia-ui-padding);
+    background-color: var(--theia-editor-background);
```
martin-fleck-at, line 446: The fetch error has a different error color and is centered:
```
.t-mv-memory-fetch-error {
    margin: var(--theia-ui-padding);
    text-align: center;
    color: var(--theia-errorForeground);
}
```

Is it a conscious choice to have it different for write errors? I am fine with it, just want to make sure.

martin-fleck-at (approved):
Hi Colin, I retested our customization with the new commits and everything still works great and I definitely saved a couple of lines of copied code ;-) I really appreciate the time you are taking for this, so thank you very much!

I found a few minors in the new code that again do not a affect the functionality so if you have some time you could fix them but otherwise I'd rather see this change merged than delayed for the next release.

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
```diff
@@ -0,0 +1,357 @@
...
+        }
+        this.update();
+    }
+
+    protected submitMemoryEdits = async (): Promise<void> => {
+        this.memoryEditsCompleted = new Deferred();
+        let didUpdateMemory = false;
+        for (const [key, edit] of this.createUniqueEdits()) {
+            try {
+                await this.memoryProvider.writeMemory(edit);
```
martin-fleck-at, line 200: Thank you for the update, could you use the `doWriteMemory` method here as well?

colin-grant-work: Yes, sorry - that was an intermediate state getting transmitted from a machine without GDB to one with GDB so I can actually test what I'm doing :-)

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
colin-grant-work: I've adjusted this a bit. It will still appear to occlude the 'load more memory' area (because the error takes up more room, so the bottom of the widget is moved up), but now it will be possible to scroll down a little further to reveal the 'load more memory' option again.

packages/memory-inspector/src/browser/style/index.css
colin-grant-work: I've aligned them with each other in favor of the styling on the write error. Prose text is better aligned left, I think.

colin-grant-work:
@martin-fleck-at, I've pushed commits that I hope should address your concerns and requests.

martin-fleck-at (approved):
Oh wow, everything has been done, thank you very much, Colin! I really appreciate it!

I re-tested the latest version with our customizations and everything works and looks great. It is really an awesome feature, thank you to everyone involved.

packages/memory-inspector/src/browser/style/index.css
martin-fleck-at: Sure, as long as it is consistent, I am fine with it, thank you!

packages/memory-inspector/src/browser/editable-widget/memory-editable-table-widget.tsx
martin-fleck-at: Looks good to me, thank you for taking another look at it.

martin-fleck-at:
@vince-fugnitto Do you want to have another look at it or do you think we can merge this?

PR #11396, opened 2022-07-08
Ensure proper null-value encoding in new RPC

### LINKED ISSUE #11392: Incorrect conversions from `null` to `undefined` in new RPC

tortmayr:
#### What it does

Fixes the bug described in #11392 by ensuring that null-values are properly encoded/decoded when sending them over the wire with the new RPC protocol.
Previously null-values were converted to `undefined` which could result in incorrect/faulty behavior. (see #11392)

Fixes #11392

#### How to test

Follow the `Steps to reproduce` in the description of #11392 and check that no error occurs/ the correct toast message is displayed.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/message-rpc/rpc-message-encoder.ts
+++ b/packages/core/src/common/message-rpc/rpc-message-encoder.ts
@@ -92,17 +92,18 @@ export enum ObjectType {
     JSON = 0,
     ByteArray = 1,
     ObjectArray = 2,
-    Undefined = 3,
-    Object = 4,
-    String = 5,
-    Boolean = 6,
-    Number = 7,
+    Null = 3,
+    Undefined = 4,
+    Object = 5,
+    String = 6,
+    Boolean = 7,
+    Number = 8,
     // eslint-disable-next-line @typescript-eslint/no-shadow
-    ResponseError = 8,
-    Error = 9,
-    Map = 10,
-    Set = 11,
-    Function = 12
+    ResponseError = 9,
+    Error = 10,
+    Map = 11,
+    Set = 12,
+    Function = 13
 
 }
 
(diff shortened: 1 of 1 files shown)


msujew (approved):
I can confirm that the issue exists on `master` (command shows `[]`) and is addressed by this change nicely (command shows `null`).

PR #11400, opened 2022-07-08
file-search: fix test case

vince-fugnitto:
#### What it does

The commit fixes the test-case which was not producing results but was passing as a false positive.
The previous implementation was relying on where `theia` was cloned and meant that the test could fail.

The test was updated to harcode the `searchPattern` rather than use the `path` to determine it.

#### How to test

- confirm that the `file-search` tests pass.

#### Review checklist

#### Reminder for reviewers


colin-grant-work (approved):
CI passes, locally and automatically, and the change makes the test more transparent.

PR #11423, opened 2022-07-14
Allow @ in deployed plugin identifiers

### LINKED ISSUE #11419: PluginDeployer won't work correctly if plugin package name has @scope
### Bug Description:

If plugins folder contains two or more plugins with `@scope` in the package name, PluginDeployer will confuse it for version and _silently_ uninstall these plugins.

### Steps to Reproduce:

1. Plugins folder has 2 plugins, package names: @my-org/pluginA, @my-org/pluginB, same publisher id.
2. On Theia startup, plugins are deployed successfully and marked as installed here: https://github.com/eclipse-theia/theia/blob/3bf8244332e57d5870c2aef1d0a7a1cdb021c7e9/packages/plugin-ext/src/hosted/node/hosted-plugin-deployer-handler.ts#L209
3. Their id and version will be evaluated here: https://github.com/eclipse-theia/theia/blob/3bf8244332e57d5870c2aef1d0a7a1cdb021c7e9/packages/plugin-ext/src/common/plugin-identifiers.ts#L76
4. And come out as: 
```
{
   id: "publisher-name."
   version: "my-org/pluginA@x.x.x"
}
```
5. Then these plugins will be uninstalled, as they have same id and different versions, here: https://github.com/eclipse-theia/theia/blob/3bf8244332e57d5870c2aef1d0a7a1cdb021c7e9/packages/plugin-ext/src/hosted/node/hosted-plugin-deployer-handler.ts#L216

### Additional Information

- Operating System: macos 12.2.1
- Theia Version: since 1.27, won't reproduce on 1.26

colin-grant-work:
#### What it does

Fixes #11419

As noted [here](https://github.com/eclipse-theia/theia/issues/11419#issuecomment-1184529919), `@` is not permitted in official plugin identifiers, but previously it was nevertheless possible to deploy plugins that used `@`, and this PR restores that ability without otherwise affecting plugin deployment.

#### How to test

0. Build this branch and download builtin plugins.
1. Move two plugins from the `plugins` folder to your user plugin directory (`~/.theia/extensions`)
2. Open the JSON manifests of those two plugins and add the same `@experiment/` scope to the `name` field of the JSON.
3. Start the example application.
4. Both plugins should be shown as installed in the VSX Registry view, and both should remain in the `~/.theia/extensions` folder.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/common/plugin-identifiers.ts
+++ b/packages/plugin-ext/src/common/plugin-identifiers.ts
@@ -56,7 +56,7 @@ export namespace PluginIdentifiers {
      * @returns a string in the format `<publisher>.<name>`.
      */
     export function unversionedFromVersioned(id: VersionedId): UnversionedId {
-        const endOfId = id.indexOf('@');
+        const endOfId = id.lastIndexOf('@');
         return id.slice(0, endOfId) as UnversionedId;
     }
     /**
@@ -64,7 +64,7 @@ export namespace PluginIdentifiers {
      */
     export function identifiersFromVersionedId(probablyId: string): Components | undefined {
         const endOfPublisher = probablyId.indexOf('.');
-        const endOfName = probablyId.indexOf('@', endOfPublisher);
+        const endOfName = probablyId.lastIndexOf('@');
         if (endOfPublisher === -1 || endOfName === -1) {
             return undefined;
         }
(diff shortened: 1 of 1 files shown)


colin-grant-work:
Closing as this is not a use case we intend to support.

Issue #11302, opened 2022-06-15
debug: support `compounds` launch configurations


vince-fugnitto:
#### Feature Description:

The goal is to support [compound](https://code.visualstudio.com/docs/editor/debugging#_compound-launch-configurations) for debug launch configurations. Currently the compound launch configurations are not present in the dropdown, or supported in the `launch.json` schema:

![image](https://user-images.githubusercontent.com/40359487/173908623-1388bc63-7fa1-4591-aa79-1dc437d2b1d6.png)

#### Steps to Reproduce

1. start the application using `theia` as a workspace
2. open the `debug` view and select the dropdown
3. notice that both compound launch configurations are missing

    https://github.com/eclipse-theia/theia/blob/f6c0689dde8b207726deab75bc3cb5cfdae6acfb/.vscode/launch.json#L196-L216

5. notice that there is no context assistance for the compounds in the `.vscode/launch.json` file

martin-fleck-at:
@vince-fugnitto Do you know if anyone is already working on this? If not I might have a look at it one of these days.

vince-fugnitto:
@martin-fleck-at I'm not aware of anyone looking into so please go ahead :)

PR #11444, opened 2022-07-20
Add support for compound launches

martin-fleck-at:
#### What it does
Fixes https://github.com/eclipse-theia/theia/issues/11302

- Extend debug model for compound launches
-- Support pre-launch task, stopAll, list of configurations
-- Properly parse compounds into configuration model
-- Support mapping to custom session option type
-- Support search of options by configuration, compound, and name

- Show compound launches in configuration dropdown
-- Query all available options
-- Use JSON (de-)serialization to allow multiple options with same name
--- Previously custom string was used for identification

- Extend session manager to start compound session options
- Ensure compound sessions can also be found from DebugMain

I tried to keep backwards compatibility with interfaces and parameters as good as possible.

#### How to test
- If you start Theia with Theia itself as a workspace, you should be able to see the already present compound launches `Launch Electron Backend & Frontend` and `Launch Browser Backend & Frontend`. But you can also add your own compound launches and play around with the options (stopAll, preLaunchTask).
- To test the call from within a plugin, I added a small plugin that has a few commands to trigger the existing launch configurations (with or without workspace) or provide a custom launch configuration: https://github.com/martin-fleck-at/theia/tree/issue-11302-test

#### Review checklist

#### Reminder for reviewers


JonasHelming:
@colin-grant-work It would help us a lot, if this could make it to the upcoming release. So if you have some cycles, we would be very thankful for a review. Thanks!

packages/debug/src/browser/debug-configuration-manager.ts
```diff
@@ -470,7 +513,7 @@ export class DebugConfigurationManager {
         // Between versions v1.26 and v1.27, the expected format of the data changed so that old stored data
         // may not contain the configuration key.
         if (data.current && 'configuration' in data.current) {
-            this.current = this.find(data.current.configuration, data.current.workspaceFolderUri, data.current.providerType);
+            this.current = this.find(data.current.configuration!, data.current.workspaceFolderUri, data.current.providerType);
         }
```
msujew, line 517: Minor: When checking `data.current.configuration` instead of `'configuration' in data.current`, you can remove the non-null assertion again.

packages/debug/src/browser/debug-configuration-model.ts
```diff
@@ -75,9 +80,20 @@ export class DebugConfigurationModel implements Disposable {
...
+        const compounds: DebugCompound[] = [];
+        if (value && typeof value === 'object' && 'compounds' in value) {
+            if (Array.isArray(value.compounds)) {
+                for (const compound of value.compounds) {
+                    if (DebugCompound.is(compound)) {
+                        compounds.push(compound);
+                    }
+                }
+            }
+        }
```
msujew, line 92: Minor: This can be simplified to
```suggestion
        if (value && typeof value === 'object' && Array.isArray(value.compounds)) {
            for (const compound of value.compounds) {
                if (DebugCompound.is(compound)) {
                    compounds.push(compound);
                }
            }
        }
```

packages/debug/src/browser/debug-session-options.ts
```diff
@@ -14,44 +14,88 @@
...
 import { DebugConfiguration } from '../common/debug-common';
+import { DebugCompound } from '../common/debug-compound';
+
+export class DebugCompoundRoot {
+    private stopped = false;
+    private stopEmitter = new Emitter<void>();
+    onDidSessionStop = this.stopEmitter.event;
+
+    sessionStopped(): void {
+        if (!this.stopped) { // avoid sending extranous terminate events
```
msujew, line 27: Minor: typo

```suggestion
        if (!this.stopped) { // avoid sending extraneous terminate events
```

packages/debug/src/browser/debug-session-manager.ts
```diff
@@ -236,13 +254,70 @@ export class DebugSessionManager {
...
+            throw new Error(nls.localizeByDefault('Compound must have "configurations" attribute set in order to start multiple configurations.'));
+        }
+
+        const compoundRoot = compound.stopAll ? new DebugCompoundRoot() : undefined;
+        const configurations: DebugConfigurationSessionOptions[] = [];
+        for (const configData of compound.configurations) {
+            const name = typeof configData === 'string' ? configData : configData.name;
+            if (name === compound.name) {
+                throw new Error(nls.localize("Launch configuration '{0}' contains a circle with itself", name));
+            }
```
msujew, line 291: Suggestion: As we already check whether a launch config calls itself in a compound config, would it make sense to perform this check recursively as well?

msujew (commented):
Functionality-wise this looks already quite good. I have a few comments on the code though.

packages/debug/src/common/debug-compound.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export interface DebugCompound {
+    name: string;
+    stopAll?: boolean;
+    preLaunchTask?: string | TaskIdentifier;
+    configurations: (string | { name: string, folder: string })[];
+}
+
+export namespace DebugCompound {
+    export function is(arg: DebugCompound | any): arg is DebugCompound {
```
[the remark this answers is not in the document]
martin-fleck-at: I agree with you 100%. I just had a look at the other type guards but it seems that using `unknown` is really the better option here, thank you!

packages/debug/src/common/debug-compound.ts
```diff
@@ -0,0 +1,33 @@
...
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+// *****************************************************************************
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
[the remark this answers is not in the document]
martin-fleck-at: Done!

packages/debug/src/browser/view/debug-configuration-select.tsx
```diff
@@ -75,7 +75,7 @@ export class DebugConfigurationSelect extends React.Component<DebugConfiguration
 
     protected get currentValue(): string {
         const { current } = this.manager;
-        return current ? InternalDebugSessionOptions.toValue(current) : DebugConfigurationSelect.NO_CONFIGURATION;
+        return current ? JSON.stringify(current) : DebugConfigurationSelect.NO_CONFIGURATION;
```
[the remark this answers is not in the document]
martin-fleck-at: See my comments above, we would break the signature as we don't use the custom type that was introduced with those functions.

packages/debug/src/browser/debug-configuration-manager.ts
```diff
@@ -248,24 +246,61 @@ export class DebugConfigurationManager {
     /**
      * Find / Resolve DebugSessionOptions from a given target debug configuration
      */
-    find(targetConfiguration: DebugConfiguration, workspaceFolderUri?: string, providerType?: string): DebugSessionOptions | undefined;
-    find(nameOrTargetConfiguration: string | DebugConfiguration, workspaceFolderUri?: string, providerType?: string): DebugSessionOptions | undefined {
+    find(compound: DebugCompound, workspaceFolderUri?: string): DebugSessionOptions | undefined;
+    find(configuration: DebugConfiguration, workspaceFolderUri?: string, providerType?: string): DebugSessionOptions | undefined;
+    find(name: string, workspaceFolderUri?: string, providerType?: string): DebugSessionOptions | undefined;
+    find(nameOrConfigurationOrCompound: string | DebugConfiguration | DebugCompound, workspaceFolderUri?: string, providerType?: string): DebugSessionOptions | undefined;
```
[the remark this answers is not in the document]
martin-fleck-at: You are absolutely right, I guess it is a remnant from some other calls I had to do but no longer need. Thank you for catching that!

packages/debug/src/browser/debug-configuration-manager.ts
martin-fleck-at: Great suggestion!

packages/debug/src/browser/debug-configuration-model.ts
martin-fleck-at: Agreed, I blindly copied the code for the configurations handling. I updated both loops with the nicer code.

packages/debug/src/browser/debug-session-options.ts
martin-fleck-at: Thank you for catching that!

packages/debug/src/browser/debug-session-manager.ts
martin-fleck-at: As far as I can tell, compound launches cannot contain other compound launches so there is no need to have a recursive check here.

martin-fleck-at:
@msujew Thank you very much for your fast and detailed review! I added another commit to it so you can see the changes but I can also squash them if you feel everything is taken care of.

packages/debug/src/browser/debug-session-options.ts
```diff
@@ -14,44 +14,88 @@
...
         return 'id' in options;
     }
 
-    export function toValue(debugSessionOptions: DebugSessionOptions): string {
-        return debugSessionOptions.configuration.name + SEPARATOR +
-            debugSessionOptions.configuration.type + SEPARATOR +
-            debugSessionOptions.configuration.request + SEPARATOR +
-            debugSessionOptions.workspaceFolderUri + SEPARATOR +
-            debugSessionOptions.providerType;
+    export function toValue(options: DebugSessionOptions): string {
```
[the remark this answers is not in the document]
msujew: Yep, I'm happy with deprecating them.

packages/debug/src/browser/debug-session-manager.ts
msujew: Alright, that's good to know 👍

packages/debug/src/common/debug-compound.ts
```diff
@@ -0,0 +1,32 @@
...
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { TaskIdentifier } from '@theia/task/lib/common';
+
+export const defaultCompound: DebugCompound = { name: 'Compound', configurations: [] };
```
colin-grant-work, line 19: Appears to be unused.

packages/debug/src/browser/debug-session-manager.ts
```diff
@@ -236,13 +248,70 @@ export class DebugSessionManager {
...
+        if (!compound.configurations) {
+            throw new Error(nls.localizeByDefault('Compound must have "configurations" attribute set in order to start multiple configurations.'));
+        }
+
+        const compoundRoot = compound.stopAll ? new DebugCompoundRoot() : undefined;
+        const configurations: DebugConfigurationSessionOptions[] = [];
+        for (const configData of compound.configurations) {
+            const name = typeof configData === 'string' ? configData : configData.name;
+            if (name === compound.name) {
+                throw new Error(nls.localize("Launch configuration '{0}' contains a circle with itself", name));
```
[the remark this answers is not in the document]
martin-fleck-at: Ah, good catch! I added the missing key to it, thank you!

packages/debug/src/common/debug-compound.ts
martin-fleck-at: True, I defined it and forgot to use it in the schema updater but now it is used there.

martin-fleck-at:
@colin-grant-work Thank you very much for your review! I pushed it as a separate commit so you can see the changes at a glance. If need be, I can squash the changes and push the whole thing again.

vince-fugnitto (commented):
@martin-fleck-at something to try but I was unable to successfully start the `Launch Electron Backend & Frontend` debug configuration. The electron window did spawn but from the main window (where the config was launched from) I was not able to interact with the debug session (ex: stop it) since it did not seem to exist.

alvsan09 (commented):
It's looking good !!

I am not sure if the following is related but 
When I start a debug session (Browser Backend and Frontend), I am getting the following errors (TextEditor is disposed!) as per the image.
  My setup is a `theia` electron app. debugging `theia` on browser running a 'hello_express' project). The two `theia` workspace folders are different.

image

martin-fleck-at:
@vince-fugnitto @alvsan09  Thank you very much for your input!

I tested it a few times with `Launch Electron Backend & Frontend` and for me the debug sessions do seem to stay interactive, especially when I cancel the debug session early on:

debug-session-cancel

The only one that sometimes causes problems is the `Attach to Plugin Host` which sometimes remains in the debug view. This can happen when I try to stop the debug sessions by stopping any of the debug sessions but also happens when I simply close the started Electron window. I can also see the following warning in the output sometimes: `Did not receive terminated event in time.`. @vince-fugnitto Is this what you are also experiencing? Do you experience the same problems when quickly starting one debug session after another manually?

For simple cases everything seems to work as expected, e.g.,
```
{
  "configurations": [{
      "command": "top",
      "name": "Run top",
      "request": "launch",
      "type": "node-terminal"
  }],
  "compounds": [{
      "name": "Multi-Run",
      "configurations": [
        "Run top",
        "Run top"
      ],
      "stopAll": false
    }]
}
```

I think it may has to do with how the debug adapters and their environment are set up or maybe it is a timing issue and some launch configurations need to wait for others to be completely started? I checked the [VSCode code](https://github.com/microsoft/vscode/blob/3aa87adc4365c37d46d03b3f3d6ce166c7c6356b/src/vs/workbench/contrib/debug/browser/debugService.ts#L359) and there they also don't have any special waiting mechanism, they simply use `Promise.all` like we do.

@alvsan09 Unfortunately, I cannot reproduce the `TextEditor is disposed` message but I think my setup might not be exactly the same. Does it cause any failures on using any of the text editors?  Do you experience the same problems when quickly starting one debug session after another manually?

martin-fleck-at:
@colin-grant-work Could you please have a look if you believe the reported issues are connected to this PR or some other setup issue that is only revealed through this PR? I already tested it with the pre-built Gitpod but I am not sure how else I can proceed here so any help would be greatly appreciated.

alvsan09 (commented):
I have tested running the updated code base to run a different `Theia` repository, and there is one combination as show below that causes errors.
I don't think it's related to this change but it would be great if all combinations below could work:

I have also changed the configurations to run 'pwa-chrome' instead of 'chrome'

Running Theia updated on browser,
        then spawning Theia on electron works fine.
        === WORKS FINE ====

Running Theia updated on electron,
        Changing electron launch backend config to: --remote-debugging-port=9223
        Changing to port: 9223 on Attach to Electron Frontend
        then spawning Theia on electron, gives different errors and ultimately crashes.
        === FAILURES ====

Running Theia updated on electron,
        then spawning Thein on browser
        === WORKS FINE ===

Running Theia updated on browser,
       Changing ports on backend / frontend launch configs.
       then spawning Theia on browser
        === WORKS FINE ===

martin-fleck-at:
@alvsan09 Thank you for your in-depth analysis! I agree that starting another Theia instance from Electron seems to cause a problem with the plugin host. 

When I start the debug session I can see that the new Electron window opens normally and everything seems fine but after some time the 'Attach to Plugin Host' reports the following problem in the original window: `Debug session initialization failed. See console for details.`. Since we have the `stopAll` option set to true that stops all other debug sessions as well and the whole started Electron application closes/crashes. However, if we set `stopAll` to false, we can see that the window remains open as the other two launch configurations `Launch Electron Backend` and `Attach to Electron Frontend` do not have a problem. I know to little about the Plugin host to actually know how the attachment works and how that problem could be fixed if it has some deep underlying issue. I tried changing the port from 9339 to 9340 but that didn't solve the issue.

Still I would argue that this is a problem specifically to the plugin host, so maybe this could be considered as a separate bug?

packages/debug/src/browser/debug-schema-updater.ts
```diff
@@ -91,6 +92,56 @@ const launchSchema: IJSONSchema = {
                 oneOf: []
             }
         },
+        compounds: {
+            type: 'array',
+            description: nls.localizeByDefault('List of compounds. Each compound references multiple configurations which will get launched together.'),
```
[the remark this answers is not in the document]
martin-fleck-at: Good catch, I also updated the other fields!

packages/debug/src/browser/debug-configuration-manager.ts
```diff
@@ -248,24 +246,61 @@ export class DebugConfigurationManager {
...
-        if (typeof nameOrTargetConfiguration === 'object' && providerType) {
-            return {
-                configuration: nameOrTargetConfiguration,
-                providerType
-            };
+        if (typeof nameOrConfigurationOrCompound === 'object' && providerType) {
+            if (DebugConfiguration.is(nameOrConfigurationOrCompound)) {
+                return this.configurationToOptions(nameOrConfigurationOrCompound, workspaceFolderUri, providerType);
+            }
+            return this.compoundToOptions(nameOrConfigurationOrCompound, workspaceFolderUri);
```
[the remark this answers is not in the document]
martin-fleck-at: No, that is not possible as far as I can tell.

packages/debug/src/browser/debug-session-options.ts
```diff
@@ -14,44 +14,88 @@
...
+        if (!this.stopped) { // avoid sending extranous terminate events
+            this.stopped = true;
+            this.stopEmitter.fire();
+        }
+    }
+}
 
 export interface DebugSessionOptionsBase {
     workspaceFolderUri?: string,
     providerType?: string // Applicable to dynamic configurations
```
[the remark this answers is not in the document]
martin-fleck-at: Good idea! I moved it to the configuration options and also created a convenience type for it.

martin-fleck-at:
I just checked and the problem with 'Attach to Plugin Host' also happens when I start all three launch configurations manually. I therefore created a new bug for it here: https://github.com/eclipse-theia/theia/issues/11475. I think I addressed all other comments so far so is there any chance we can still get this merged before the release? An adopter would really appreciate that.

alvsan09:
Hi @martin-fleck-at ,
Thanks for writing the issue above.

I just saw the following problem, when toggling `stopAll` in the editor,
in the following, the `Auto save` option is on, but changes to it does not always take effect on the next launch. 
 
https://github.com/alvsan09/hello_express

martin-fleck-at:
@alvsan09 Oh wow, great catch, thank you! Seems like I removed something when updating the current configuration and so changed properties were not correctly applied.

alvsan09:
@martin-fleck-at, 
The latest change is breaking non compound configs, see: 

single_launch_fails

martin-fleck-at:
@alvsan09 Oh wow, you are right! I am very sorry for missing that! Seems like the typeguards were not good enough to detect a dynamic configuration vs a regular configuration. I just checked whether the property was there but not if that property was set.

alvsan09:
@martin-fleck-at ,
I saw one more problem,
we are no longer able to select dynamic configurations, see `gif` below:

unable_to_select_dynamic_configs

martin-fleck-at:
@alvsan09 I updated the search for the current configuration again, I think now all cases should be covered. Thank you very much for taking the time!

alvsan09:
@martin-fleck-at , 
The latest change fixes the previous issues, 
unfortunately, I found new ones: 

1) When starting a workspace used in a previous version of `theia` where that workspace had registered dynamic configurations in local storage, the configurations will be restored with a label of `undefined`.
restoring_dynamic_configs_issue

2) From the debug view, selecting `Add Configuration` has no effect
3) When reloading the browser, `sometimes` the selection box in the debug view go out of bounds, 
I tried reproducing this one on master but I couldn't, with the new change I had to try two or three times and it will happen.

See the `gif` below:
  
debug_view_selection

martin-fleck-at:
@alvsan09 Thank you again for your feedback! 

1. I fixed the naming issue, there was indeed a change as we now use a different serialization so the name was not properly recognized.
2. For me 'Add Configuration...' opens the launch.json file as expected both in Chrome and in Firefox.
3. I also cannot reproduce the select rendering issue. I tried several times in Chrome and in Firefox but it always renders to the bottom.

debug-select

Could you please give it another try? I don't know what could have caused this but I cannot reproduce it at all. What browser are you using?

alvsan09:
@martin-fleck-at ,
I tried the correction for 'undefined' showing in the names on restored dynamic configs and it works fine now, :+1: 

> Could you please give it another try? I don't know what could have caused this but I cannot reproduce it at all. What browser are you using?
For the two remaining issues, I managed to reproduce them using a mult-root workspace, and the amount of entries in the configuration is larger e.g.. all the ones provided by theia + all the ones provided by the `hello_express` example mentioned above.

The following clip shows the difference when having a folder vs a workspace project.

https://user-images.githubusercontent.com/76971376/181771509-74032bfb-10bf-40c5-9154-9f2e3d6230e1.mp4

martin-fleck-at:
@alvsan09 With the multi-root workspace service I can indeed reproduce the problem! Seems like when we need to select a workspace the update of the configurations dropdown takes focus and automatically closes the workspace select, I now keep it open even on focus lost.

As for the rendering "issue": This really depends on how much space is available at the top or bottom. Previously, if there was not enough space at the bottom, it was rendered to the top. I changed that so that it prefers bottom rendering unless more content can be shown when rendering to the top.

Hopefully this will resolve the remaining issues. In any case, I really appreciate you sticking with me on this ;-)

alvsan09:
@martin-fleck-at 
I have done a regression test and almost everything seems to work as far as I can tell, 
Except the following issues (related to recent dynamic configurations from old format)

1)
The video starts showing a set of dynamic launch configurations which were created on a previous session with `Theia` master,
then we launch each of them and duplicates can sometimes be created.

https://user-images.githubusercontent.com/76971376/182202236-6ca35688-2631-45f5-9b64-d356e77788a2.mp4

2)
I saw a strange behaviour once before where a non dynamic config. managed to get inserted among the recent dynamic configurations, I am not sure how to reproduce it, although  I managed to get a screen shot of the serialized data.

image

alvsan09:
> As for the rendering "issue": This really depends on how much space is available at the top or bottom. Previously, if there was not enough space at the bottom, it was rendered to the top. I changed that so that it prefers bottom rendering unless more content can be shown when rendering to the top.

@msujew, Do you agree with the change of approach affecting the custom `Select Component` widget ?

msujew:
@alvsan09 @martin-fleck-at Yeah, the changes look reasonable to me. Everything in regards to the select component still works as expected or even better 👍

martin-fleck-at:
@alvsan09 I hope I managed to remove the last issues now. When restoring the old dynamic configurations I now ensure that the options name is properly derived if it is not set already. I also filter for dynamic options to ensure nothing else gets into that list (even though I was not able to reproduce that locally).

Interestingly, the fix showed an issue which probably was the reason for the initial custom serialization: When we set the current value in the debug session manager, the value we have in current does not necessarily match exactly one of the options we have on the configuration drop down, mostly due to the fact that properties are in different order. So I now aim to match the current value to the current option before when updating the select. As far as I can tell, this now works for all option types and updates in the launch.json are also properly considered.

packages/debug/src/browser/debug-configuration-manager.ts
```diff
@@ -527,7 +529,12 @@ export class DebugConfigurationManager {
             return;
         }
 
-        this.recentDynamicOptionsTracker = options;
+        // ensure options name is reflected from old configurations data
+        const dynamicOptions = options.map(option => {
+            option.name = option.name ?? option.configuration.name;
+            return option;
+        }).filter(DebugSessionOptions.isDynamic);
```
alvsan09, line 536: is it really necessary to filter for `isDynamic` ?
non dynamic configs should not be available under the received argument:
`options?: DynamicDebugConfigurationSessionOptions[]`, right?

alvsan09 (commented):
Hi @martin-fleck-at , 
I have done a full regression and it looks almost there !

I found the issue shown on the video clip below, where dynamic configurations are not restored  are not restored from local storage.

I have also added a question as an inline comment.

https://user-images.githubusercontent.com/76971376/182635356-108bf9ff-7fb9-4e45-927c-1ac4db695b33.mp4

packages/debug/src/browser/debug-configuration-manager.ts
martin-fleck-at: You are absolutely correct that it should not be necessary. I added this guard because you had the case where there was a regular configuration in the 'recentDynamicOptions'. I could not reproduce the case but during runtime we already have a type guard before updating recently used dynamic options so I reasoned that it probably was caused during loading some old data. Since the recent options are limited to 3, I thought the extra safety outweighs any potential performance cost. What do you think?

alvsan09: I am leaning towards removing it, as the case may have been caused while switching among different versions of this change.
  If we ever see it again, it's not too bad for the user to ignore it or replace it by making a new selection, and then we could try to fix the actual source of the problem.

alvsan09:
Using the last commit I have verified that the functionality works as expected ! :+1: 

@vince-fugnitto and @colin-grant-work, any pending items on your side ?

I have successfully run the tests below:
* Storing / Restoring configurations as `current` for (dynamic, not dynamic & compound)
* Storing / Restoring `recently used` dynamic configurations
* Restoring configurations from previous `theia` version
* Launching configuration by 'debug ' command
* Launching compound configurations
        - `stopAll` [true, false]
        - toggling `stopAll` takes effect at the next launch
        - with a `preLaunchTask`
* Launching not compound configurations
* "Add configuration" works
* The above works for multi-root as well as for project folders

vince-fugnitto (approved):
I don't have any objections to the changes 👍 
I believe Alvaro reviewed thoroughly https://github.com/eclipse-theia/theia/pull/11444#issuecomment-1205280054.

martin-fleck-at:
@colin-grant-work @msujew Any objections to merging this?

colin-grant-work:
> Any objections?

None from me.

Issue #10662, opened 2022-01-24
Implement `env` namespace VSCode API


lucas-koehler:
### Feature Description:

Add the `env` namespace of the VSCode API to the plugin API.

- appHost
- isNewAppInstall
- isTelemetryEnabled
- onDidChangeTelemetryEnabled
- remoteName

See the VSCode API for more information on the listed properties/event: https://code.visualstudio.com/api/references/vscode-api#env

paul-marechal:
fyi: We have someone looking at this.

PR #11446, opened 2022-07-20
Add missing vscode api in env namespace

yiningwang11:
#### What it does

Relevant issues: https://github.com/eclipse-theia/theia/issues/10662

Added appHost, isNewAppInstall, isTelemetryEnabled, onDidChangeTelemetryEnabled, and remoteName APIs to our missing vscode API implementation.

#### How to test

Using the electron app, notifications indicating the value of those APIs would pop up.
The appHost is set to 'desktop', the remoteName is set to 'undefined'. For isNewAppInstall, we use the creation date of a random theia file as the installation date. Moreover, since theia doesn't support telemetry, we implemented stubs for the API related to telemetry. The isTelemetryEnabled API is set to false. The onDidChangeTelemetryEnabled always fires an empty event.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/hosted/browser/worker/worker-env-ext.ts
+++ b/packages/plugin-ext/src/hosted/browser/worker/worker-env-ext.ts
@@ -34,4 +34,8 @@ export class WorkerEnvExtImpl extends EnvExtImpl {
         throw new Error('There is no app root in worker context');
     }
 
+    get isNewAppInstall(): boolean {
+        throw new Error('Cannot determine the installation date in worker context');
+    }
+
 }
--- a/packages/plugin-ext/src/plugin/env.ts
+++ b/packages/plugin-ext/src/plugin/env.ts
@@ -19,6 +19,7 @@ import { RPCProtocol } from '../common/rpc-protocol';
 import { EnvMain, PLUGIN_RPC_CONTEXT } from '../common/plugin-api-rpc';
 import { QueryParameters } from '../common/env';
 import { v4 } from 'uuid';
+import { Emitter, Event } from '@theia/core/lib/common/event';
 
 export abstract class EnvExtImpl {
     private proxy: EnvMain;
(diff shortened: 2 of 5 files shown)


packages/plugin-ext/src/hosted/browser/worker/worker-env-ext.ts
```diff
@@ -34,4 +34,8 @@ export class WorkerEnvExtImpl extends EnvExtImpl {
         throw new Error('There is no app root in worker context');
     }
 
+    get isNewAppInstall(): boolean {
+        throw new Error('Cannot determine the installation date in worker context');
```
msujew, line 38: Suggestion: Since [vscode](https://github.com/microsoft/vscode/blob/edc585edf6ec34d121196a3eda229c1c0fed676e/src/vs/workbench/api/common/extHost.api.impl.ts#L325-L328) never throws here, we should probably just return `false` to keep better compatibility with vscode.

packages/plugin-ext/src/plugin/env.ts
```diff
@@ -29,11 +30,20 @@ export abstract class EnvExtImpl {
...
+    private host: string;
+    private _isTelemetryEnabled: boolean;
+    private _remoteName: string | undefined;
+    private onDidChangeTelemetryEnabledEmitter: Emitter<boolean>;
 
     constructor(rpc: RPCProtocol) {
         this.proxy = rpc.getProxy(PLUGIN_RPC_CONTEXT.ENV_MAIN);
         this.envSessionId = v4();
         this.envMachineId = v4();
+        this.host = 'desktop';
```
msujew, line 42: Suggestion: It looks to me like [vscode](https://github.com/microsoft/vscode/blob/edc585edf6ec34d121196a3eda229c1c0fed676e/src/vs/workbench/services/extensions/browser/webWorkerExtensionHost.ts#L280) uses `web` here when the `uiKind` is also `web`. Since we set `uiKind` to `web` when running in the browser mode, we should probably also set `host` to `web` in that case.

msujew (commented):
@yiningwang11 Thank you for your first contribution!

Please make sure to sign the [ECA](https://www.eclipse.org/legal/ECA.php) with the same email as your commit in order for us to accept your changes.

packages/plugin-ext/src/plugin/env.ts
paul-marechal: I understood `web` as in the extension host would be running in a browser? And `desktop` would be more similar to how VS Code desktop runs: the plugin host is a regular node process. Not sure what's the actual semantic for this value.

msujew: I see, makes sense.

JonasHelming:
@yiningwang11  Is this ready for a re-review? I see that the ECA succeeds now.

vince-fugnitto (approved):
The changes look good to me and work well :+1: 
I confirm for both the browser and electron the values are correct.

JonasHelming (approved):

JonasHelming:
@vince-fugnitto : Can we merge this for the release still?

paul-marechal:
@JonasHelming the changes were low risk enough to be merged in our opinion, yes :)

PR #11449, opened 2022-07-21
terminal: buffer the output before sending

paul-marechal:
In order to not overload the websocket channel opened between the
frontend and the backend we should chunk and buffer the output of
terminals.

Add `BufferingStream` that will buffer data for some time before
emitting it.

Closes https://github.com/eclipse-theia/theia/pull/11181

#### How to test

Terminals should still work.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/terminal/src/node/buffering-stream.ts
+++ b/packages/terminal/src/node/buffering-stream.ts
@@ -18,12 +18,12 @@ import { Emitter, Event } from '@theia/core/lib/common/event';
 
 export interface BufferingStreamOptions {
     /**
-     * Max size of the chunks being emitted.
+     * Max size in bytes of the chunks being emitted.
      */
     maxChunkSize?: number
     /**
-     * Amount of time to wait between the moment we start buffering data
-     * and when we emit the buffered chunk.
+     * Amount of time in milliseconds to wait between the moment we start
+     * buffering data and when we emit the buffered chunk.
      */
     emitInterval?: number
 }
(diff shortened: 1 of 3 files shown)


packages/terminal/src/node/buffering-stream.ts
```diff
@@ -0,0 +1,78 @@
...
+    protected buffer?: Buffer;
+    protected timeout?: NodeJS.Timeout;
+    protected maxChunkSize: number;
+    protected emitInterval: number;
+
+    protected onDataEmitter = new Emitter<Buffer>();
+
+    constructor(options?: BufferingStreamOptions) {
+        this.emitInterval = options?.emitInterval ?? 16; // ms
+        this.maxChunkSize = options?.maxChunkSize ?? 16384; // bytes
```
msujew, line 47: I feel like 16kb is leads to too many messages being send without pauses in between. At least when trying this out (using `find ~`) on Gitpod, it still leads to the connection starvation experienced on `master`. Increasing this to 256kb does improve responsiveness massively.

msujew (commented):
Looks way cleaner than the previous approach. However, I still notice resource starvation when running on non-local servers.

packages/terminal/src/node/buffering-stream.ts
paul-marechal: I wonder if could be automatically tested, I don't know how right now.

msujew (approved):
Great, looks good to me and cleaner than the other proposed approach 👍

PR #11450, opened 2022-07-21
Ensure WidgetManager creates same key for objects that are deep equal

### LINKED ISSUE #11309: [editor] Widget manager should compare widget keys with deep equals instead of string equal
### Bug Description:

When opening editors, the opener `options` is used to generate an [identity key](https://github.com/eclipse-theia/theia/blob/024d8714393cc0d8f821d6350cba71c6df4a63ad/packages/core/src/browser/widget-manager.ts#L258-L260) for the widget. This is clear and works as expected. However, the property order of the opener `options` also matters when trying to [find an already opened widget](https://github.com/eclipse-theia/theia/blob/024d8714393cc0d8f821d6350cba71c6df4a63ad/packages/core/src/browser/widget-manager.ts#L195).

The following code causes unexpected behavior:

```ts
const uri = 'file:///path/to/an/existing/resource.txt';
const options = {
    counter: 0,
    mode: 'reveal',
    preview: false
} as EditorOpenerOptions;
await this.editorManager.open(uri, options);

const sameOptions = {
    ...options,
    preview: options.preview,
    mode: options.mode // change the property order
};
await this.editorManager.open(widget.editor.uri, sameOptions);
```

My code is here: https://github.com/eclipse-theia/theia/compare/master...kittaakos:opener-options
Steps to reproduce:
 - Open an editor,
 - Wait 2 sec

See in action:

https://user-images.githubusercontent.com/1405703/174308827-9fa9c652-caea-44d5-8365-c8f5f36576de.mp4

Do I misuse the APIs? if not, I think Theia should compare keys with deepEqual and try to find the existing widget instead doing `Map.get` by the JSON options.

### Steps to Reproduce:

1.
2.
3.

### Additional Information

- Operating System: macOS 12.3.1
- Theia Version: 462376a7993

colin-grant-work:
#### What it does

Fixes #11309 by using a library that produces stable stringifications for deep-equal objects.

#### How to test

1. A test command is included in the first commit of this PR (to be dropped)
> Label: `Try opening the same file with different options!`
3. Running that command without the second commit should produce a log that says that it is `false` that the two widgets are equal.
4. Running that command with the second commit of this PR should produce a log that says that it is `true` that the two widgets are equal.
5. CI should pass.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/widget-manager.ts
+++ b/packages/core/src/browser/widget-manager.ts
@@ -17,6 +17,7 @@
 import { inject, named, injectable } from 'inversify';
 import { Widget } from '@phosphor/widgets';
 import { ILogger, Emitter, Event, ContributionProvider, MaybePromise, WaitUntilEvent } from '../common';
+import stableJsonStringify = require('fast-json-stable-stringify');
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 export const WidgetFactory = Symbol('WidgetFactory');
@@ -256,7 +257,7 @@ export class WidgetManager {
      * @returns the widget construction options represented as a string.
      */
     protected toKey(options: WidgetConstructionOptions): string {
-        return JSON.stringify(options);
+        return stableJsonStringify(options);
     }
 
     /**
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -41,6 +41,7 @@
     "drivelist": "^9.0.2",
     "es6-promise": "^4.2.4",
     "express": "^4.16.3",
+    "fast-json-stable-stringify": "^2.1.0",
     "file-icons-js": "~1.0.3",
     "font-awesome": "^4.7.0",
     "fs-extra": "^4.0.2",
--- a/packages/core/src/browser/widget-manager.spec.ts
+++ b/packages/core/src/browser/widget-manager.spec.ts
@@ -63,4 +63,18 @@ describe('widget-manager', () => {
         expect(wA).equals(await widgetManager.getOrCreateWidget('test', 'widgetA'));
     });
 
+    it('produces the same widget key regardless of object key order', () => {
+        const options1 = {
+            factoryId: 'a',
+            key1: 1,
+            key2: 2,
+        };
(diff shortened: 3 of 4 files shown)


packages/core/src/browser/widget-manager.ts
```diff
@@ -256,7 +257,7 @@ export class WidgetManager {
      * @returns the widget construction options represented as a string.
      */
     protected toKey(options: WidgetConstructionOptions): string {
-        return JSON.stringify(options);
+        return JSONstringify(options);
```
paul-marechal, line 260: nitpick: rename to `stableJsonStringify`?

packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
```diff
@@ -298,6 +300,25 @@ export class PluginVscodeCommandsContribution implements CommandContribution {
...
+                const secondEditor = await this.widgetManager.getOrCreateWidget('code-editor-opener', {
+                    'kind': 'navigatable',
+                    'uri': workspaceRoot.resource.resolve('package.json').toString(),
+                    'counter': 0,
+                });
+                console.log('SENTINEL FOR WHETHER THE EDITORS ARE EQUAL...', firstEditor === secondEditor);
+                firstEditor.dispose();
+                secondEditor.dispose();
+            }
+        });
```
paul-marechal, line 320: Should we have a place where we can put those commands without polluting random extensions? Maybe `@theia/api-samples` would be a better place for that next time?

paul-marechal (approved):
Code+test LGTM and the 3PP check is green.

packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
colin-grant-work: If it's going to be dropped, I just put it in whatever package seems to have already imported most of what I need 😄.

PR #11454, opened 2022-07-22
Add errors handling for plugins downloading

### LINKED ISSUE #11260: Handle unexpected Open VSX responses more gracefully
### Feature Description:

Today the Open VSX registry backend was down, and queries from the `download:plugins` script were failing. It seems that Open VSX was returning responses that had HTML rather than JSON in their body, and that caused unexpected, build-crashing errors when resolving dependencies

image

We should perhaps make the download script more robust against such errors so that the build can continue if the user wishes it to.

pluralia:
#### What it does
Closes #11260.
Catches and collects errors that occur when loading plugins. This allows to try to load all plugins and not terminate the load after the first fail. Does not interact with the user (what was suggested in the issue), because the loading of plugins goes in parallel and, as a consequence, it is difficult to stop loading if the user wants to.

#### How to test
Run `yarn download:plugins --api-url https://google.com` -- plugins downloading is failed. Before the change you get an error after loading the first plugin. After the change, you should see attempts to load all plugins.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/cli/src/download-plugins.ts
+++ b/dev-packages/cli/src/download-plugins.ts
@@ -147,13 +147,17 @@ export default async function downloadPlugins(options: DownloadPluginsOptions = 
             // De-duplicate extension ids to only download each once:
             const ids = new Set<string>(dependencies.flat());
             await parallelOrSequence(...Array.from(ids, id => async () => {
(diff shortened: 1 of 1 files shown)


vince-fugnitto (approved):
I confirm that the changes work as expected 👍

PR #11455, opened 2022-07-22
Update React to v18

AlexandraBuzila:
#### What it does

This PR updates Theia to React v18. We adapted the invocations of React where necessary and made sure that Theia does not behave differently. With these changes in place we could not observe any regressions.

The major change in React v18 affecting the Theia codebase is the deprecation of the `ReactDOM.render` API. Although it is still available, we adapted all invocations of React to use the new `createRoot` API, enabling new React v18 features like concurrent rendering throughout the code base. What we avoided is to explicitly use any of the new features to improve the Theia UX. This should be done separately.

With the removal of the `render` API we lost the option to use its render callbacks to trigger functionality in Theia. In consequence, the `onRender` field from `ReactDialog` and `ReactWidget` was removed. Users of the render callbacks are adapted to use the equivalent `ref` callbacks or `useEffect` patterns instead.

Other minor updates:
- the `children` prop now needs to be explicitly specified when needed
- React state updates are now batched  (`SearchInWorkspaceInput` was using an intermediate state, this is fixed now)

Consumer code will most likely continue to work, especially when using `ReactDOM.render` for now instead of the new API, as the new features of React are then not used. In extreme cases, React 17 could be additionally loaded and used for their own code.

If a jump directly to v18 is not desired at the moment, we also prepared a branch which updates React to v17: https://github.com/eclipsesource/theia/tree/react-v17-update 

#### Review checklist

#### Reminder for reviewers

Signed-off-by: Alexandra Buzila <abuzila@eclipsesource.com>


dependency-check-baseline.json
```diff
@@ -3,5 +3,6 @@
   "npm/npmjs/-/jschardet/2.3.0": "Approved for Eclipse Theia: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22481",
   "npm/npmjs/-/jsdom/11.12.0": "Approved as 'works-with': https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23640https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23640",
   "npm/npmjs/-/lzma-native/8.0.6": "Approved as 'works-with': https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/1850",
-  "npm/npmjs/-/playwright-core/1.22.2": "Approved as 'works-with': https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/2734"
+  "npm/npmjs/-/playwright-core/1.22.2": "Approved as 'works-with': https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/2734",
+  "npm/npmjs/-/request-promise-core/1.1.4": "Approved as 'works-with': https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/3033"
```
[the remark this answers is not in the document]
AlexandraBuzila: Without it the license check failed for me. I'll remove it to see if it's still the case.

package.json
```diff
@@ -7,7 +7,9 @@
     "node": ">=14"
   },
   "resolutions": {
-    "**/@types/node": "14"
+    "**/@types/node": "14",
+    "@types/react": "18.0.15",
+    "@types/react-dom": "18.0.6"
```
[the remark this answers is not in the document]
vince-fugnitto: I'm not a huge fan of resolutions since it does not solve the same problem for downstream extensions and applications. We should probably think of using [react-window](https://github.com/bvaughn/react-window) instead of `react-virtualized` which is written by the same author and is a replacement for the latter ([related](https://github.com/bvaughn/react-window#how-is-react-window-different-from-react-virtualized)). We can tackle it later if needed but we should probably mention something in the migration guide?

dependency-check-baseline.json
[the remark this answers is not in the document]
AlexandraBuzila: It seems to work now without it, not sure why it didn't before.

package.json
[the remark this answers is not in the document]
AlexandraBuzila: I added a note in the migration guide.

AlexandraBuzila:
I rebased the changes on the latest master and fixed the conflicts.

packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
```diff
@@ -50,6 +50,7 @@ export const COMMENT_TITLE: MenuPath = ['comment-title-menu'];
 export class CommentThreadWidget extends BaseWidget {
 
     protected readonly zoneWidget: MonacoEditorZoneWidget;
+    protected containerNodeRoot: Root;
```
msujew, line 53: Minor: I'm wondering whether it makes sense to also declare this `readonly`, as it's only set in the constructor, right?

packages/preferences/src/browser/views/preference-searchbar-widget.tsx
```diff
@@ -139,7 +143,7 @@ export class PreferencesSearchbarWidget extends ReactWidget implements StatefulW
         const optionContainer = this.renderOptionContainer();
         return (
             <div className='settings-header'>
-                <div className="settings-search-container">
+                <div className="settings-search-container" ref={this.focus}>
```
msujew, line 146: Question: Is there a meaningful difference between `ref={this.focus}` + `this.focus = this.focus.bind(this);` and `ref={() => this.focus}`? If not, we should always prefer to use the latter. The former is quite unintuitive and overcomplicates the code in my opinion.

packages/search-in-workspace/src/browser/search-in-workspace-widget.tsx
```diff
@@ -100,6 +100,8 @@ export class SearchInWorkspaceWidget extends BaseWidget implements StatefulWidge
     @inject(SearchInWorkspacePreferences)
     protected readonly searchInWorkspacePreferences: SearchInWorkspacePreferences;
 
+    protected searchFormContainerRoot: Root;
```
msujew, line 103: Minor: Again, what about making this readonly?

msujew (commented):
I'd generally try to merge this relatively early in this month's dev cycle to find possible regressions and also address the outstanding comment from Vince (replacing the `react-virtualized` package).

packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
AlexandraBuzila: I changed it to `readonly`.

packages/preferences/src/browser/views/preference-searchbar-widget.tsx
AlexandraBuzila: Yes, unfortunately there's a difference. When using an arrow function like this `ref={() => this.focus}`, it will have a new identity on each re-render. The `ref` will be recreated each time, triggering two calls to the callback: first when the `ref` will be set to null and again when the function will be assigned to the `ref`. So on every render (e.g. on each key press in the search bar), the `focus` function will be called twice.

This does not happen when using a stable callback, like it is the case currently in the code.

packages/search-in-workspace/src/browser/search-in-workspace-widget.tsx
AlexandraBuzila: It does not work here, because the `searchFormContainerRoot` is created in the post construct callback. I guess the code could be moved to the constructor, the class does not currently have one. Would you prefer that?

msujew: It doesn't really bother me too much, it's fine 👍

doc/Migration.md
```diff
@@ -19,6 +19,26 @@ For example:
 }
 ```
 
+### v1.28.0
```
msujew, line 22: Please update this to `1.29.0`.

msujew (approved):
I'm fine with the changes as they are and I can't see any regressions in the application. After merging this, I'll take care of replacing `react-virtualized` with `react-window`.

doc/Migration.md
AlexandraBuzila: Thank you for the review, @msujew! I updated the version.

AlexandraBuzila:
I rebased the branch on the lastest master and fixed a conflict.

msujew:
@vince-fugnitto Any objections to merging this? I'll be merging this later today.

vince-fugnitto:
@msujew I must have missed the ping from https://github.com/eclipse-theia/theia/pull/11455#pullrequestreview-1058775044, I'm fine with merging and we can do the update we mentioned afterwards.

PR #11457, opened 2022-07-22
vsx: display error message when failing to fetch

vince-fugnitto:
#### What it does

The pull-request updates the **Extensions** view to display an error alert message when we fail to fetch extensions from the marketplace, like we had during the week. The changes should help with the user-experience so they have feedback when something goes wrong.

_Open-VSX being down:_

Screen Shot 2022-07-22 at 2 12 09 PM

#### How to test

1. set the `VSX_REGISTRY_URL` environment variable to some fake url (ex: `https://open-vsx.foo`)
2. start the application.
3. open the **Extensions** view.
4. confirm the error message is visible when searching and failing to fetch extensions.

#### Review checklist

#### Reminder for reviewers


msujew (approved):
The changes look good to me 👍 Please rebase on `master` before merging.

vince-fugnitto:
@msujew I'm not sure why but with the auto translation update we don't get the additional checks for CI.
 I tried twice and it didn't work unfortunately.

msujew:
> It does appear on the original commit however, just not the overall pull-request:

Yeah, I noticed as well already. Should we try to change it so it also runs on the CI commit?

vince-fugnitto:
> Yeah, I noticed as well already. Should we try to change it so it also runs on the CI commit?

@msujew I think so, else the final checks are misleading 👍

msujew:
@vince-fugnitto I've looked into this, but it seems like this is a [deliberate decision of GitHub](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow). When pushing using the provided github token, no github actions are triggered. There seem to be [ways around that](https://github.com/stefanzweifel/git-auto-commit-action#commits-made-by-this-action-do-not-trigger-new-workflow-runs) though, which I'll try out.

PR #11472, opened 2022-07-27
Wait for plugins to deploy before restoring the current language

### LINKED ISSUE #11471: Restoring current language might fail at start-up
### Bug Description:
It might happen that, when setting a language different from `en`, the language is not restored when quitting and reopening the IDE.

Doing some research, I noticed the root of the problem is here:
https://github.com/eclipse-theia/theia/blob/f4327c2f0c0f47f960a302590a51698879357841/packages/core/src/node/i18n/localization-backend-contribution.ts#L38-L39
`req.params.locale` is always restored correctly, but `this.localizationProvider.getAvailableLanguages()` doesn't return all the available languages. This happens because `getAvailableLanguages()` returns all the registered localizations with `languagePack` set to `true` (unless we call `getAvailableLanguages(true)`, in that case, it would always return all the languages)
https://github.com/eclipse-theia/theia/blob/f4327c2f0c0f47f960a302590a51698879357841/packages/core/src/node/i18n/localization-provider.ts#L50

The weird thing is that this happens also with _some_ languages registered with a language pack (so they'll eventually have `languagePack` set to `true`). I say _some_ languages because _some other_ work as expected. I strongly believe this is happening because of a race condition happening because `getAvailableLanguages` is called before the language packs are finished loading.

### Steps to Reproduce:
You can use this bug in the Arduino IDE as a reference to reproduce this bug: https://github.com/arduino/arduino-ide/issues/1219

### Additional Information
Passing `true` to `getAvailableLanguages` fixes the issue for me.

- Operating System: MacOS
- Theia Version: 1.25.0

AlbyIanna:
#### What it does
Wait for the plugins to be deployed before restoring the current language.

Fixes #11471

#### How to test
- Open IDE select `de` as the default
 - IDE reloads, it shows `de`
 - Quit IDE
 - Start IDE, it still shows `de` ✅
 - Select `zh-cn` as the default language
 - IDE reloads and shows `zh-cn` ✅
 - Quit IDE
 - Open IDE
 - The language is still `zh-cn` ✅

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/node/plugin-ext-backend-module.ts
+++ b/packages/plugin-ext/src/main/node/plugin-ext-backend-module.ts
@@ -40,7 +40,7 @@ import { PluginTheiaEnvironment } from '../common/plugin-theia-environment';
 import { PluginTheiaDeployerParticipant } from './plugin-theia-deployer-participant';
 import { WebviewBackendSecurityWarnings } from './webview-backend-security-warnings';
 import { PluginUninstallationManager } from './plugin-uninstallation-manager';
-import { LocalizationBackendContribution } from '@theia/core/src/node/i18n/localization-backend-contribution';
+import { LocalizationBackendContribution } from '@theia/core/lib/node/i18n/localization-backend-contribution';
 import { PluginLocalizationBackendContribution } from './plugin-localization-backend-contribution';
 
 export function bindMainBackend(bind: interfaces.Bind, unbind: interfaces.Unbind, isBound: interfaces.IsBound, rebind: interfaces.Rebind): void {
--- a/packages/plugin-ext/src/main/node/plugin-ext-backend-module.ts
+++ b/packages/plugin-ext/src/main/node/plugin-ext-backend-module.ts
@@ -40,8 +40,10 @@ import { PluginTheiaEnvironment } from '../common/plugin-theia-environment';
 import { PluginTheiaDeployerParticipant } from './plugin-theia-deployer-participant';
 import { WebviewBackendSecurityWarnings } from './webview-backend-security-warnings';
 import { PluginUninstallationManager } from './plugin-uninstallation-manager';
+import { LocalizationBackendContribution } from '@theia/core/src/node/i18n/localization-backend-contribution';
+import { PluginLocalizationBackendContribution } from './plugin-localization-backend-contribution';
 
-export function bindMainBackend(bind: interfaces.Bind): void {
+export function bindMainBackend(bind: interfaces.Bind, unbind: interfaces.Unbind, isBound: interfaces.IsBound, rebind: interfaces.Rebind): void {
     bind(PluginApiContribution).toSelf().inSingletonScope();
     bind(BackendApplicationContribution).toService(PluginApiContribution);
     bind(WsRequestValidatorContribution).toService(PluginApiContribution);
@@ -87,4 +89,7 @@ export function bindMainBackend(bind: interfaces.Bind): void {
 
     bind(WebviewBackendSecurityWarnings).toSelf().inSingletonScope();
     bind(BackendApplicationContribution).toService(WebviewBackendSecurityWarnings);
+
+    rebind(LocalizationBackendContribution).to(PluginLocalizationBackendContribution).inSingletonScope();
+
 }
--- a/packages/plugin-ext/src/main/node/plugin-localization-backend-contribution.ts
+++ b/packages/plugin-ext/src/main/node/plugin-localization-backend-contribution.ts
@@ -1,21 +1,42 @@
+// *****************************************************************************
+// Copyright (C) 2021 TypeFox and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+// *****************************************************************************
+
 import { inject, injectable } from '@theia/core/shared/inversify';
 import { LocalizationBackendContribution } from '@theia/core/lib/node/i18n/localization-backend-contribution';
 import { PluginDeployer } from '../../common/plugin-protocol';
 import { PluginDeployerImpl } from './plugin-deployer-impl';
-import { Deferred } from '@theia/core/src/common/promise-util';
+import { Deferred } from '@theia/core/lib/common/promise-util';
 
 @injectable()
 export class PluginLocalizationBackendContribution extends LocalizationBackendContribution {
     @inject(PluginDeployer)
     protected readonly pluginDeployer: PluginDeployerImpl;
+    protected readonly pluginsDeployed = new Deferred();
 
     override async initialize(): Promise<void> {
-        const pluginsDeployed = new Deferred();
         this.pluginDeployer.onDidDeploy(() => {
-            pluginsDeployed.resolve();
+            this.pluginsDeployed.resolve();
         });
+        await super.initialize();
+    }
 
-        await Promise.all([this.localizationRegistry.initialize(), pluginsDeployed.promise]);
-        this.initialized.resolve();
+    override async waitForInitialization(): Promise<void> {
+        await Promise.all([
+            super.waitForInitialization(),
+            this.pluginsDeployed.promise,
+        ]);
     }
 }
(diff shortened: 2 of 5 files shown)


vince-fugnitto (commented):

Please also be sure to properly fill out the original pull-request template.

msujew (commented):
I'm not sure that this is the right way to go about this, since it doesn't actually resolve the issue. When getting all available languages, including those which don't have language packs loaded, we will definitely have to deal with the problem of 90% of Theia not being translated.

As far as I can tell this is a race condition, right? So instead, the express endpoint should await the plugin loading process, where all language packs are loaded into memory. This should be done fairly quickly without impacting the load time on Electron.

I can probably contribute a fix for this within the next few days.

AlbyIanna:
> we will definitely have to deal with the problem of 90% of Theia not being translated.


This would be true only during the loading time, right?

But yes, I would go for the other solution too, I just couldn't find a way to await for the languages loading process. Probably I've just didn't dig deep enough (and also didn't know how waiting for it would affect the Theia loading time)

msujew:
> This would be true only during the loading time, right?

Yes, but that's the issue. We already load translations during loading time. The `/i18n/:locale` endpoint not only returns metadata on languages, but all of the translations as well, since they need to be available at the time of loading the JS code.

AlbyIanna:
@msujew I see. Thanks for the tip, I think I've found a better solution [here](https://github.com/arduino/arduino-ide/pull/1261/files#diff-19d891f1ffb8ce2979b6bee0d166700e20a6b965d1be5f72b3e331d0442eba35)

if you like it, I'd gladly update this PR with that code.

msujew:
@AlbyIanna Sure, something similar should work for the Theia codebase as well 👍

AlbyIanna:
@msujew done 👌

msujew (commented):
@AlbyIanna With "something similar" I didn't mean the exact same approach ;) Given our architecture, you can't simply import code from the `plugin-ext` package into the `core` package. You'll have to create a dedicated override in the `plugin-ext` package for the `LocalizationBackendContribution` and bind it there.

AlbyIanna:
> You'll have to create a dedicated override in the plugin-ext package for the LocalizationBackendContribution and bind it there.


@msujew I'm not sure how to proceed here. Could you please provide me with an example of how to do that? 

> Also we can't accept your changes when you haven't signed the [ECA](https://www.eclipse.org/legal/ECA.php). Please do so.

I should have done it now. Can you please confirm I did it right?

msujew:
> I'm not sure how to proceed here. Could you please provide me with an example of how to do that?

Sure, I recently did something similar: There is a `core` [`LanguageQuickPickService`](https://github.com/eclipse-theia/theia/blob/master/packages/core/src/browser/i18n/language-quick-pick-service.ts), which is [overriden](https://github.com/eclipse-theia/theia/blob/0c0f8c7a7aedd67930bf65d6f63d827721707d60/packages/vsx-registry/src/browser/vsx-registry-frontend-module.ts#L105) in the `vsx-registry` package as [`VSXLanguageQuickPickService`](https://github.com/eclipse-theia/theia/blob/master/packages/vsx-registry/src/browser/vsx-language-quick-pick-service.ts) as the services in there aren't available in `core`.

> I should have done it now. Can you please confirm I did it right?

Are you sure you did it with the mail that's in the commit? It doesn't seem to validate.

image

AlbyIanna:
> Are you sure you did it with the mail that's in the commit? It doesn't seem to validate.

Yes, I'm sure about it because I received the confirmation e-mail after the registration.


This is what I see in my Eclipse Account:
image

msujew:
> Yes, I'm sure about it because I received the confirmation e-mail after the registration.

Alright, cool. It seems to work now. No idea why it didn't work previously.

packages/plugin-ext/src/main/node/plugin-localization-backend-contribution.ts
```diff
@@ -0,0 +1,35 @@
...
+  protected readonly initialized = new Deferred<void>();
+
+  override async initialize(): Promise<void> {
+    this.pluginDeployer.onDidDeploy(() => {
+      this.initialized.resolve();
+    });
+    return super.initialize();
+  }
+
+  override configure(app: express.Application): void {
```
msujew, line 22: Suggestion: To cut down on code duplication in this part, it might make sense to introduce a method in `LocalizationBackendContribution` called `waitForInitialization` that ensures that all initialization has already resolved. You can await that instead in the base class, removing the need to reimplement the express endpoint again.

packages/plugin-ext/src/main/node/plugin-localization-backend-contribution.ts
```diff
@@ -0,0 +1,35 @@
+import * as express from 'express';
+import { inject, injectable } from '@theia/core/shared/inversify';
+import { LocalizationBackendContribution } from '@theia/core/lib/node/i18n/localization-backend-contribution';
+import { Deferred } from '@theia/core/lib/common/promise-util';
+import { PluginDeployer } from '../../common/plugin-protocol';
+import { PluginDeployerImpl } from './plugin-deployer-impl';
+
+@injectable()
+export class PluginLocalizationBackendContribution extends LocalizationBackendContribution {
+  @inject(PluginDeployer)
```
msujew, line 10: Minor: We use a tab width of 2 spaces across Theia. Please align this file to this guideline.

msujew (commented):
There are a few linting errors across the changes. You should probably enable the `ESLint` vscode extension so you can catch those during development.

AlbyIanna:
@msujew thanks for the feedback. In the last commit I've made some changes according to your requests 👍

packages/plugin-ext/src/main/node/plugin-ext-backend-module.ts
```diff
@@ -40,8 +40,10 @@ import { PluginTheiaEnvironment } from '../common/plugin-theia-environment';
 import { PluginTheiaDeployerParticipant } from './plugin-theia-deployer-participant';
 import { WebviewBackendSecurityWarnings } from './webview-backend-security-warnings';
 import { PluginUninstallationManager } from './plugin-uninstallation-manager';
+import { LocalizationBackendContribution } from '@theia/core/src/node/i18n/localization-backend-contribution';
```
msujew, line 43: Minor: Please don't import from `src`.
```suggestion
import { LocalizationBackendContribution } from '@theia/core/lib/node/i18n/localization-backend-contribution';
```

packages/plugin-ext/src/main/node/plugin-localization-backend-contribution.ts
```diff
@@ -0,0 +1,21 @@
+import { inject, injectable } from '@theia/core/shared/inversify';
+import { LocalizationBackendContribution } from '@theia/core/lib/node/i18n/localization-backend-contribution';
+import { PluginDeployer } from '../../common/plugin-protocol';
+import { PluginDeployerImpl } from './plugin-deployer-impl';
+import { Deferred } from '@theia/core/src/common/promise-util';
```
msujew, line 5: Minor: Please don't import from `src`.

```suggestion
import { Deferred } from '@theia/core/lib/common/promise-util';
```

packages/plugin-ext/src/main/node/plugin-localization-backend-contribution.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    override async initialize(): Promise<void> {
+        const pluginsDeployed = new Deferred();
+        this.pluginDeployer.onDidDeploy(() => {
+            pluginsDeployed.resolve();
+        });
+
+        await Promise.all([this.localizationRegistry.initialize(), pluginsDeployed.promise]);
+        this.initialized.resolve();
+    }
```
msujew, line 20: Suggestion: While this works quite well, the idea behind introducing `waitForInitialization` was that we would be able to hide the default behavior of `super.initialize()`:
```suggestion
    protected readonly pluginsDeployed = new Deferred();

    override async initialize(): Promise<void> {
        this.pluginDeployer.onDidDeploy(() => {
            this.pluginsDeployed.resolve();
        });
        await super.initialize();
    }
    
    override waitForInitialization(): Promise<void> {
       return Promise.all([super.waitForInitialization(), this.pluginsDeployed.promise]);
    }
```

msujew (commented):
Great, just one change and this should be good to go 👍

Also, just for completeness' sake, do you mind filling out the PR template?

msujew:
@AlbyIanna Are you still interested in contributing the changes? I still have a few outstanding comments before approving this.

AlbyIanna:
Hi @msujew, thanks for the review. Sorry for the delay in answering, I must have missed the notification 🙏 
I will address your remarks soon.

msujew (approved):
The changes look good to me 👍 

Please address the linting issues (see below) and then I'll merge this.

packages/plugin-ext/src/main/node/plugin-localization-backend-contribution.ts
```diff
@@ -0,0 +1,26 @@
+import { inject, injectable } from "@theia/core/shared/inversify";
+import { LocalizationBackendContribution } from "@theia/core/lib/node/i18n/localization-backend-contribution";
+import { PluginDeployer } from "../../common/plugin-protocol";
+import { PluginDeployerImpl } from "./plugin-deployer-impl";
+import { Deferred } from "@theia/core/lib/common/promise-util";
```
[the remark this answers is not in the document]
AlbyIanna: Done

PR #10938, opened 2022-03-28
Ensure default icon theme is properly applied even if registered late

martin-fleck-at:
#### What it does
Ensures that the default icon theme is applied properly even if registered late by storing it in the local cache on initialization.

Fixes https://github.com/eclipse-theia/theia/issues/10937

#### How to test
0. Setup available on this branch as well: https://github.com/martin-fleck-at/theia/tree/issue-10937-case
1. Install any icon theme from vsx into your product, e.g., https://open-vsx.org/extension/PKief/material-icon-theme
2. Set the default icon theme for your application to that theme, e.g., "defaultIconTheme": "material-icon-theme"
3. Start up the (browser or electron) application and see if the theme is applied on a clean state. You can get to a clean state by clearing the browser data or removing the respective electron directories.

https://user-images.githubusercontent.com/19170971/160380088-f17c5dad-9859-405e-aa73-0e8d2a379520.mp4

#### Review checklist
- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers
- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #11473, opened 2022-07-27
Modify theme services to use preference service

colin-grant-work:
#### What it does

This PR's primary objective is to modify the `ThemeService` and `IconThemeService` to interact directly with the `PreferenceService` as their primary source of truth. Previously, neither service was `injectable`, and so neither had access to the `PreferenceService` directly. As a consequence, their interactions with the `PreferenceService` were mediated by the `CommonFrontendContribution` and both services relied heavily on local storage for persistence.

#### How to test

1. Open your user `settings.json` and execute one of the commands `Preferences: Color Theme` or `Preferences: Icon Theme`.
1. Move the selector to preview themes.
1. Note that the `settings.json` file is _not_ modified.
1. Select a theme from the quick pick.
1. Note that your `settings.json` _is_ modified.
1. In the `settings.json` file, set a nonsensical value (`"Definitely not any kind of theme!"`) for the preference.
1. Note that the theme you selected remains in place.
1. Reload the page.
1. Note that the theme you selected (if it was a color theme - we don't store assets of icon themes) still remains in place.
> This is different from VSCode's behavior where the theme is reset to default on restart. It would be possible to implement this behavior, if preferable.
---
Check for #10938
1. Remove the Material Icon Theme from your plugins, if you have it.
1. Open the application, and in your user `settings.json`, set `workbench.iconTheme` to `"material-icon-theme"`.
1. No change should be immediately observable.
1. Install the Material Icon Theme from Open VSX.
1. The theme should immediately take effect.
---
1. Set your theme to `Red` or something else with a distinctive editor background color.
1. Reload the application.
1. There should be no theme flashes.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 - [Previous Changelogs](https://github.com/eclipse-theia/theia/tree/master/doc/changelogs/)
 
+## v1.29.0 - unreleased
+
+<a name="breaking_changes_1.29.0">[Breaking Changes:](#breaking_changes_1.29.0)</a>
+
+- [core] `updateThemePreference` and `updateThemeFromPreference` removed from `CommonFrontendContribution`. Corresponding functionality as been moved to the respective theme service. `load` removed from `IconThemeService` [#11473](https://github.com/eclipse-theia/theia/issues/11473)
+
 ## v1.28.0 - 7/28/2022
 
 - [cli] improved error handling when interacting with the API [#11454](https://github.com/eclipse-theia/theia/issues/11454)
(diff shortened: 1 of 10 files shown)


msujew (approved):
Looks good to me 👍 

- [x] Selected themes (color/icon) are persisted across reloads
  - [x] When changing the theme and reloading the app, the background color is correctly set to the last theme applied.
  - [x] Even if the user enters an incorrect/nonsensical entry in the user settings
- [x] Setting the icon theme and only loading the corresponding vscode extension afterwards correctly applies the icon theme to the application.

307god:
```
let theiaDidInitialize = false;
const originalInitialize = StandaloneServices.initialize;
StandaloneServices.initialize = overrides => {
    if (!theiaDidInitialize) {
        console.warn('Monaco was initialized before overrides were installed by Theia\'s initialization.'
            + ' Please check the lifecycle of services that use Monaco and ensure that Monaco entities are not instantiated before Theia is initialized.', new Error());
    }
    return originalInitialize(overrides);
};
```
this contaminate @injectable() MonacoFrontendApplicationContribution
if i want rebind MonacoFrontendApplicationContribution, i can`t change theiaDidInitialize and StandaloneServices.initialize,it always tell me 'Monaco was initialized before overrides were installed by Theia\'s initialization.'
what can i do?
@colin-grant-work @msujew

PR #11474, opened 2022-07-28
Fix: Display failed when text document content provider return empty string ('')

hyy215:
#### What it does

Fix：Display failed when text document content provider return empty string ('')
image

#### How to test

1. install [text-document-demo-0.0.1.vsix](https://github.com/hyy215/vsx-demo/blob/master/text-document-demo/text-document-demo-0.0.1.vsix)
2. run 'Show Custom Text Document' command
3. show empty string in editor
image

#### Review checklist

#### Reminder for reviewers

- As a reviewer, I agree to review in accordance with [the review guidelines](https://github.com/eclipse-theia/theia/blob/master/doc/pull-requests.md#reviewing)

### DIFF
--- a/packages/plugin-ext/src/main/browser/workspace-main.ts
+++ b/packages/plugin-ext/src/main/browser/workspace-main.ts
@@ -367,7 +367,7 @@ export class TextContentResource implements Resource {
             return content;
         } else {
             const content = await this.proxy.$provideTextDocumentContent(this.uri.toString());
-            if (content !== undefined && content !== null) {
+            if (typeof content === 'string') {
                 return content;
             }
         }
--- a/packages/plugin-ext/src/main/browser/workspace-main.ts
+++ b/packages/plugin-ext/src/main/browser/workspace-main.ts
@@ -367,9 +367,7 @@ export class TextContentResource implements Resource {
             return content;
         } else {
             const content = await this.proxy.$provideTextDocumentContent(this.uri.toString());
-            if (typeof content === 'string') {
-                return content;
-            }
+            return content ?? '';
         }
 
         return Promise.reject(new Error(`Unable to get content for '${this.uri.toString()}'`));
--- a/packages/plugin-ext/src/main/browser/workspace-main.ts
+++ b/packages/plugin-ext/src/main/browser/workspace-main.ts
@@ -367,7 +367,7 @@ export class TextContentResource implements Resource {
             return content;
         } else {
             const content = await this.proxy.$provideTextDocumentContent(this.uri.toString());
-            if (content) {
+            if (content !== undefined && content !== null) {
                 return content;
             }
         }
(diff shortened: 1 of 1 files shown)


packages/plugin-ext/src/main/browser/workspace-main.ts
```diff
@@ -367,7 +367,7 @@ export class TextContentResource implements Resource {
             return content;
         } else {
             const content = await this.proxy.$provideTextDocumentContent(this.uri.toString());
-            if (content) {
+            if (content !== undefined && content !== null) {
```
msujew, line 370: Minor: It might be better to simply check whether the content is a string instead of checking what it isn't.

```suggestion
            if (typeof content === 'string') {
```

msujew (commented):
It seems like my changes lead to the ECA check failing. Do you mind squashing the commits so that only your original email appears in the commit log?

hyy215:
>

hyy215:
> It seems like my changes lead to the ECA check failing (they were authored using `574477988@qq.com`, which isn't covered by the ECA). Do you mind squashing the commits so that only your original email appears in the commit log?

I amend the commit email.

packages/plugin-ext/src/main/browser/workspace-main.ts
```diff
@@ -367,7 +367,7 @@ export class TextContentResource implements Resource {
             return content;
         } else {
             const content = await this.proxy.$provideTextDocumentContent(this.uri.toString());
-            if (content) {
+            if (typeof content === 'string') {
                 return content;
```
vince-fugnitto, line 371: Should we not simply return a default value if `content` is `undefined` or `null` like:

```suggestion
            return content ?? '';
```

msujew: Yeah, that's probably the better solution for that issue.

vince-fugnitto (commented):
@hyy215 please take a look at the build errors,  The `throw` is also unreachable so we can safely remove it.

packages/plugin-ext/src/main/browser/workspace-main.ts
```diff
@@ -367,9 +367,7 @@ export class TextContentResource implements Resource {
             return content;
         } else {
             const content = await this.proxy.$provideTextDocumentContent(this.uri.toString());
-            if (content) {
-                return content;
-            }
+            return content ?? '';
         }
 
         return Promise.reject(new Error(`Unable to get content for '${this.uri.toString()}'`));
```
msujew, line 373: Minor: This `return` statement is unreachable, so we should remove it.

hyy215: @msujew @vince-fugnitto Thank you for your review. I have resolve it.

msujew (approved):
Thanks, I can confirm that the issue exists on `master` and is addressed by these changes nicely 👍

Issue #11482, opened 2022-07-29
cli: Enhance `theia` CLI to identify builds pulling multiple versions of Theia


planger:
### Problem

In a Theia-based application, if `yarn install` is not able to resolve all direct and indirect dependencies of `@theia/*` packages to a unique Theia version, the resulting build will likely lead to a broken application, because the dependency injection modules contributed by those extensions may point to different keys when binding their contributions or requesting injections. It is easy to run into such a situation, especially if you depend on third-party Theia extensions. But they can also happen if you forgot to upgrade the version (ranges, tags) of direct dependencies to Theia packages in an extension.

Such scenarios are usually only detected at runtime: often the Theia front-end app doesn't start at all with inversify errors in the console or, if you are unlucky, only certain functionality isn't working right, which is even nastier as this may detected only very late in end-to-end tests, etc. This is especially confusing for new adopters of Theia, as they are not yet sensitive to these very specific inversify errors.

### Feature Description

Theia builds should fail at compile time in such cases and avoid having adopters to face those issues only at runtime without an easy-to-understand error message for new adopters. Thus, I suggest to introduce a CLI tooling that detects scenarios in which dependencies have been resolved to a mix of different `@theia/*` package versions (or other relevant packages, which adopters should be able to configure). Those dependency issues can already be detected after `yarn install`, so adopters could integrate this CLI check after running `yarn install` in their build pipeline.

Adopters should be able to configure the check in some aspects:
* which local workspaces to include in the check (maybe use `yarn.workspaces` by default)
* which dependencies to include and exclude (e.g. all @theia/**) in the check
* which checks to perform: hoisting, unique version
* whether to fail the build or just print warnings

### Related: `theia check:hoisted`

There is already a CLI command `theia check:hoisted` which is capable of detecting many, but not all of the problematic cases. However, for the use case above, it has the following limitations:

* It uses hard-coded paths and expects all Theia extensions to be located in the folder `packages`
* It doesn't look at application packages, which often reside somewhere else (e.g. `examples` in the Theia repo)
* It raises an issue if *any* dependency wasn't hoisted (i.e. couldn't be resolved to a unique version), including dev-dependencies. While it is surely desirable to avoid non-hoisted dependencies, it often is hard to achieve (e.g. rimraf), especially with third-party extensions, and in many cases unproblematic.
[...]

PR #11483, opened 2022-07-29
cli: Enhance tooling for checking dependencies

planger:
#### What it does

Enhances the `theia` CLI with dedicated checks regarding pulled dependencies as detailed below.

##### `theia check:theia-version`

*Check that all dependencies have been resolved to the same Theia version.*

It'll obtain all package workspaces of the monorepo from `package.json/workspaces` and run through their `node_modules` as well as the root `node_modules` in order to collect all dependencies to `@theia/**` packages and test whether they all have the same version. It'll exclude `@theia/monaco-editor-core` though, as it follows the version of Monaco, which differs from the common Theia version.

Parameters:
* `--suppress` or `-s`: Suppress exiting with failure code

##### `theia check:hoisted`

*Check that all dependencies are hoisted.*

This command behaves as before, but uses the same enhanced implementation as the command above and below. Also the output is a bit improved with respect to readability in comparison to the current implementation.

##### `theia check:dependencies`

*Check uniqueness of dependency versions or whether they are hoisted.*

This command is the basis for the two commands above, but exposes the full configurability to adjust the dependency checks to specific needs.

Parameters:
* ` --workspaces` or `-w`: Glob patterns of workspaces to analyze, relative to `cwd`. [array] [default: All glob patterns listed in the package.json's workspaces]
* `--include` or `-i`: Glob pattern of dependencies' package names to be included, e.g. `-i "@theia/**"`. [array] [default: ["**"]]
* `--exclude` or `-e`: Glob pattern of dependencies' package names to be excluded. [array] [default: None]
* `--skip-hoisted` or `-h`: Skip checking whether dependencies are hoisted. [boolean] [default: false]
* `--skip-uniqueness` or `-u`: Skip checking whether all dependencies are resolved to a unique version. [boolean] [default: false]
* `--skip-single-theia-version` or `-t`: Skip checking whether all @theia/** dependencies are resolved to a single version. [boolean] [default: false]
* `--suppress` or `-s`: Suppress exiting with failure code. [boolean] [default: false]

See https://github.com/eclipse-theia/theia/issues/11482 for the problem this PR aims to address.

#### How to test

There are three commands (see above) to test.
For `theia check:hoisted` it should be tested whether it prints the same amount of information as before. However, the content may look slightly different now:

image

To test `yarn check:theia-version` you could proceed as follows:

* Check out this PR and build (`yarn`)
* Go to `dev-packages/cli` and run `yarn link`
* Create a sample Theia application, e.g. with the theia generator, and run `yarn link "@theia/cli"` in its root

* Running `yarn && yarn theia check:theia-version` should give no errors: 
image

* To test different Theia versions of non-overlapping dependencies, you could remove `@theia/terminal` from the dependencies of `browser-app` and `electron-app`, set all their dependencies to `1.28.0`, but add `@theia/terminal` as a dependency _with a `1.26.0`_ in a Theia extension of your project. Note how no `@theia/*` dependencies are hoisted. However, in the root `node_modules` you'll find a mix of versions in the `@theia/*` dependencies (terminal vs the others). Now run `yarn && yarn theia check:theia-version` again and it should print: 
image

* You can now re-add `@theia/terminal` with version `1.28.0` to `browser-app`, which will also lead to non-hoisted dependencies and re-run `yarn && yarn theia check:theia-version -s` (suppress). You'll get now one error (the same error as above), but also a warning that there is a dependency that isn't hoisted. Also note how the exit code differs because of the `-s`: 
image

To test `yarn check:dependencies` you could proceed as above, but now play around with the parameters. Implementation-wise it is the same as the two commands above. Try e.g.
`yarn theia check:dependencies --workspaces "packages/**" --exclude "rimraf/**" --skip-single-theia-version`
which will find non-hoisted dependencies, only the packages of the folder `packages`, but will not show an error for `rimraf`.

Don't forget to `yarn unlink` in `dev-packages/cli` after testing.

#### Review checklist

#### Reminder for reviewers


planger:
Just wanted to ping if someone would be available to give feedback to this proposal? :relaxed: Thanks a lot!

planger:
@msujew Did you have a chance to look at this PR? I'm looking forward to your feedback. Thanks a lot!

dev-packages/cli/src/check-dependencies.ts
```diff
@@ -0,0 +1,306 @@
...
+    const remainingIndent = indent - issueNumber.toString().length;
+    const indentString = ' '.repeat(remainingIndent + 1);
+    console.log(issueTitle(issue, issueNumber, indentString));
+    console.log(issueDetails(issue, '   ' + ' '.repeat(indent)));
+    console.log();
+}
+
+function issueTitle(issue: DependencyIssue, issueNumber: number, indent: string): string {
+    const dependent = issue.package.dependent ? ` in ${chalk.blueBright(issue.package.dependent ?? 'unknown')}` : '';
+    return chalk.bgWhiteBright.bold(`#${issueNumber}${indent}`) + ' ' + chalk.cyanBright(issue.package.name)
```
vince-fugnitto, line 269: I'm not sure if it is my setup or terminal preferences but I do not see the text (white on white) for the issue number:

Screen Shot 2022-08-23 at 1 59 49 PM

dev-packages/cli/src/check-dependencies.ts
```diff
@@ -0,0 +1,306 @@
...
+    for (const include of options.include) {
+        matchingPackageJsons.push(
+            ...glob.sync(include + '/' + PACKAGE_JSON, {
+                cwd: nodeModulesDir,
+                ignore: [`**/${NODE_MODULES}/**`, ...options.exclude]
+            })
+        );
+    }
+    return matchingPackageJsons.map(packageJson => toDependency(packageJson, nodeModulesDir, dependent));
+}
```
paul-marechal, line 126: Please consider the following code that avoids iterating twice over everything:

```suggestion
function findDependencies(workspace: string, options: CheckDependenciesOptions): Package[] {
    const dependent = getPackageName(path.join(process.cwd(), workspace, PACKAGE_JSON));
    const nodeModulesDir = path.join(workspace, NODE_MODULES);
    const matchingPackageJsons: Package[] = [];
    options.include.forEach(include => {
        glob.sync(`${include}/${PACKAGE_JSON}`, {
            cwd: nodeModulesDir,
            ignore: [`**/${NODE_MODULES}/**`, ...options.exclude]
        })
        .forEach(packageJson => {
            matchingPackageJsons.push(toDependency(packageJson, nodeModulesDir, dependent));
        });
    });
    return matchingPackageJsons;
}
```

dev-packages/cli/src/check-dependencies.ts
```diff
@@ -0,0 +1,306 @@
...
+
+function deriveWorkspaces(options: CheckDependenciesOptions): string[] {
+    const wsGlobs = options.workspaces ?? readWorkspaceGlobsFromPackageJson();
+    const workspaces: string[] = [];
+    for (const wsGlob of wsGlobs) {
+        workspaces.push(...glob.sync(wsGlob + '/'));
+    }
+    return workspaces;
+}
```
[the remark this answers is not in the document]
paul-marechal: I believe that the use of `workspaces` downstream should be optional indeed.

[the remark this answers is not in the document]
planger: Great suggestion, thanks! I'll make it optional.
I indeed only tested a project without `workspaces` in combination with the option `--workspaces`.

dev-packages/cli/src/check-dependencies.ts
planger: Hm, interesting. I'll amend a change to set the font to black explicitly. Can you please test with your theme and terminal preferences, whether this works for you?

dev-packages/cli/src/check-dependencies.ts
planger: Sure, thanks, that is probably more efficient!

planger:
@vince-fugnitto @paul-marechal Thanks a lot for your feedback! I pushed an update which includes your suggestions.

vince-fugnitto (commented):
I confirm that the changes work better 👍 
I'll give @paul-marechal a chance to perform a final review as well.

dev-packages/cli/package.json
```diff
@@ -47,7 +47,9 @@
     "puppeteer": "^2.0.0",
     "puppeteer-to-istanbul": "^1.2.2",
     "temp": "^0.9.1",
-    "yargs": "^15.3.1"
+    "yargs": "^15.3.1",
+    "glob": "^8.0.3",
+    "log-update": "^4.0.0"
```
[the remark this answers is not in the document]
planger: Sure, no problem! Done with https://github.com/eclipse-theia/theia/commit/94ac97460ee55464ed9c642995c10d573725cf7b

paul-marechal:
Using the following `package.json`

```json
{
  "name": "theia-app-example",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "@theia/core": "^1.28.0",
    "@theia/plugin-ext": "^1.28.0",
    "@theia/plugin-ext-vscode": "^1.28.0"
  },
  "devDependencies": {
    "@theia/cli": "^1.28.0"
  }
}
```

Running `theia check:dependencies` seems to find errors that I believe shouldn't be picked up:

image

Notice what the paths look like: They aren't nested dependencies... As a rule of thumb I think we should only consider packages like `node_modules/package.json`.

paul-marechal:
Otherwise I tried polluting my example app by adding `@theia/workspace@1.25.0` to force a duplication and it was correctly picked up. I think it should be good to merge once my previous issue is fixed.

planger:
Thanks again for your feedback!
@paul-marechal You're right, we should filter package.json's deeply nested in dependencies. I've now added these two ignore globs:
```javascript
`[^@]*/*/**/${PACKAGE_JSON}`, // package.json that isn't at the package root (and not in an @org)
`@*/*/*/**/${PACKAGE_JSON}`, // package.json that isn't at the package root (and in an @org)
```

I confirmed that the dependencies you mentioned aren't showing up, while other expected ones still show up.
Can you confirm this works for your example too?

paul-marechal (approved):
Successfully picks up duplicated Theia package version, this will be helpful!

PR #11484, opened 2022-07-29
#10027 Support EvaluatableExpressions

### LINKED ISSUE #10027: Support `EvaluatableExpression`s
### Feature Description:
Implement support for plugins providing evalutable epressions. see https://code.visualstudio.com/api/references/vscode-api#EvaluatableExpression

Currently, the the debugger scans the current line to guess what expressions to evaluate and show in the hover widget. Instead, the debuggers should forward that work to plugins. It may be necessary to keep the current logic as a fallback when no expressions are provided for a particular location.

ndoschek:
#### What it does

- Implement support for plugins providing evalutable epressions
- Update the debug hover widget to consume evaluatable expressions from the registered providers. Keep the former implementation of guessing the expression from the current line as fallback.

Signed-off-by: Nina Doschek <ndoschek@eclipsesource.com>

Fixes #10027

#### How to test

To test, I created a simple evalutable expression provider for the `java` language, available via the custom vscode extension: [evaluatable-expression-extension-10027-0.0.1.zip](https://github.com/eclipse-theia/theia/files/9219279/evaluatable-expression-extension-10027-0.0.1.zip)

1. Install the extensions `Language Support for Java(TM) by Red Hat` (identifier: `redhat.java`) and `Debugger for Java` (identifier: `vscjava.vscode-java-debug`) from the Open VSX Registry (`Extensions` view) in your Theia test workspace.
On startup, it will show an information message that the provider was registered.
2. Download and install the test extension `terminal-message-extension-11145.vsix` in your Theia test workspace (via `Extensions` view - top menu bar `More actions...` - Install from VSIX).
3. Create a simple Java application to debug
<details><summary>Simple Test application</summary>

```java
package myproject;

public class Test {

    public static void main(String[] args) {
        for (int i = 0; i < 2; i++) {
            String logMessage = "logging a message for i=" + i;
            System.out.println(logMessage);
        }
    }
}
```
</details>

4. Set a breakpoint in line 8 (sysout) and launch the file debugging
<details><summary>Launch config</summary>

```json
 {
   "type": "java",
   "name": "Launch Current File",
   "request": "launch",
   "mainClass": "${file}"
 }
```
</details>

5. Once the breakpoint is hit, hover over the local variable `logMessage`, the extension shows an information message with the cursor position and the provided word range of the expression.
e.g. `[TestEvaluatableExpressionExtension] provide word range for position { "line": 6, "character": 26 } > [ { "line": 6, "character": 19   },   { "line": 6, "character": 29 } ]`

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ## v1.29.0 - unreleased
 
+- [plugin] added support for `EvaluatableExpressions` [#11484](https://github.com/eclipse-theia/theia/pull/11484) - Contributed on behalf of STMicroelectronics
+
 <a name="breaking_changes_1.29.0">[Breaking Changes:](#breaking_changes_1.29.0)</a>
 
 - [core] `updateThemePreference` and `updateThemeFromPreference` removed from `CommonFrontendContribution`. Corresponding functionality as been moved to the respective theme service. `load` removed from `IconThemeService` [#11473](https://github.com/eclipse-theia/theia/issues/11473)
--- a/packages/core/src/common/types.ts
+++ b/packages/core/src/common/types.ts
@@ -162,6 +162,13 @@ export namespace ArrayUtils {
         }
         return -(low + 1);
     }
+
+    /**
+     * @returns New array with all falsy values removed. The original array IS NOT modified.
+     */
+    export function coalesce<T>(array: ReadonlyArray<T | undefined | null>): T[] {
+        return <T[]>array.filter(e => !!e);
+    }
 }
 
 /**
--- a/packages/debug/src/browser/editor/debug-hover-widget.ts
+++ b/packages/debug/src/browser/editor/debug-hover-widget.ts
@@ -28,6 +28,11 @@ import { DebugExpressionProvider } from './debug-expression-provider';
 import { DebugHoverSource } from './debug-hover-source';
 import { DebugVariable } from '../console/debug-console-items';
 import * as monaco from '@theia/monaco-editor-core';
+import { StandaloneServices } from '@theia/monaco-editor-core/esm/vs/editor/standalone/browser/standaloneServices';
+import { ILanguageFeaturesService } from '@theia/monaco-editor-core/esm/vs/editor/common/services/languageFeatures';
+import { CancellationTokenSource } from '@theia/monaco-editor-core/esm/vs/base/common/cancellation';
+import { Position } from '@theia/monaco-editor-core/esm/vs/editor/common/core/position';
+import { ArrayUtils } from '@theia/core';
 
 export interface ShowDebugHoverOptions {
     selection: monaco.Range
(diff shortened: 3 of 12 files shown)


colin-grant-work (commented):
I believe the implementation is correct, but while testing, I discovered a problem with the implementation of our `DebugHoverWidget`. I tested this PR using [this plugin](https://github.com/colin-grant-work/demo-extensions/tree/main/evaluatable-expression-demo) and [this debugger](https://github.com/microsoft/vscode-mock-debug) with the `InlineValueProvider` code removed to allow it to activate.

As far as I can tell, all of the EvaluatableExpression code ran as expected - the plugin code was called, the returns were correct - but the DebugHoverWidget never appeared. It seems that when this code runs

https://github.com/eclipse-theia/theia/blob/0c0f8c7a7aedd67930bf65d6f63d827721707d60/packages/debug/src/browser/editor/debug-hover-source.tsx#L49-L56

`elements` is undefined, and so there is nothing to render when the `DebugHoverWidget` requests elements

https://github.com/eclipse-theia/theia/blob/0c0f8c7a7aedd67930bf65d6f63d827721707d60/packages/debug/src/browser/editor/debug-hover-source.tsx#L35-L37

and so the `DebugHoverWidget` does not appear. In VSCode, by contrast, the widget does appear and renders the text as returned by the debugger.

ndoschek:
Thanks for your comments @colin-grant-work!
Regarding the minor ones, I updated my commit accordingly.

About the `mock debug` extension, I have a more general problem, that it does not work as properly as in VSCode for me, i.e. it does not hit breakpoints and a debug session cannot be exited/terminated. It is stuck on the last line of any markdown file after hitting `Continue` and does not respond to the `Stop` button. Stepping however works, and in that case I can confirm your observation that the debug widget does not appear.
I used the current master of the mock debug extension, and omitted the `provideInlineValues` command as you described.
Is there something I'm missing?

As I mentioned in the description, for the Java case, the debugging session and the hover widget are working fine for me.

colin-grant-work:
@ndoschek, I see the problems you mention - I'm not sure what's going on there, and it should be investigated elsewhere - but it looks like it's still possible to step though, so it's possible to stop on any line you'd like, even without a breakpoint. The inability to exit the session is a bummer. I tried moving back to commit `553fbf61c7d42d9c77ead3fdcb8bc134fef11677` (and removing the `InlineValueProvider` code), and the same problem with breakpoints is present, but the session at least doesn't seem to crash and be unkillable at the end of the file.

ndoschek:
Thanks @colin-grant-work, that's way better now, I'll have another look at the mentioned issue!

colin-grant-work:
@ndoschek, I definitely think that the problems with the hover widget ought to be fixed, and the problems with the mock debug extension investigated, but in principle both are separate from the work you've done. If you'd prefer to pursue the fix for the hover widget in a separate PR, I think the plugin side of this PR is correct and complete.

ndoschek:
Hi @colin-grant-work,
Thanks for your input, I had another look at the causes of the issues with the mock debug extension, and it is not yet fully clear to me what the actual cause is. Hence, I would prefer to separate those two concerns, as the support of EvaluatableExpressions via plugins is available now with this PR.
I created a separate issue for the formerly discussed issue here: #11500

colin-grant-work (approved):
The implementation of the `EvaluatableExpressionProvider` API is complete and correct. 👍

PR #11487, opened 2022-07-29
plugin: add `keys` support for `Memento`

vince-fugnitto:
#### What it does

The pull-request adds support for `keys()` in the `Memento` VS Code API.

#### How to test

1. include the following test plugin ([vscode-memento-keys-0.0.1.vsix.zip](https://github.com/eclipse-theia/theia/files/9221698/vscode-memento-keys-0.0.1.vsix.zip)).
2. start the application with a workspace.
3. execute the command `Memento: Workspace Keys` - a notification should appear with workspace keys (denoted by `w-` prefix.
4. execute the command `Memento: Global Keys` - a notification should appear with global keys (workspace keys should also appear).

The result should be the same as in VS Code:

Screen Shot 2022-07-29 at 12 33 19 PM

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/plugin-storage.ts
+++ b/packages/plugin-ext/src/plugin/plugin-storage.ts
@@ -38,6 +38,10 @@ export class Memento implements theia.Memento {
         }
     }
 
+    keys(): string[] {
+        return Object.entries(this.cache).filter(([, value]) => value !== undefined).map(([key]) => key);
+    }
+
     get<T>(key: string): T | undefined;
     get<T>(key: string, defaultValue: T): T;
     get<T>(key: string, defaultValue?: T): T | undefined {
--- a/packages/plugin/src/theia.d.ts
+++ b/packages/plugin/src/theia.d.ts
@@ -3470,6 +3470,13 @@ export module '@theia/plugin' {
     export interface Memento {
 
         /**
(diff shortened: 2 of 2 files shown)


packages/plugin-ext/src/plugin/plugin-storage.ts
```diff
@@ -38,6 +38,10 @@ export class Memento implements theia.Memento {
         }
     }
 
+    keys(): string[] {
+        return Object.entries(this.cache ?? {}).filter(([, value]) => value !== undefined).map(([key]) => key);
```
msujew, line 42: Question: I'm not too familiar with this code, but it seems like `this.cache` can never be undefined, right? Or are there cases in which the `?? {}` actually does something? If so, we should probably align `get<T>` and `update` so they also correctly deal with `cache` being undefined.

msujew (commented):
The change looks good to me. I have a question though, see below.

packages/plugin-ext/src/plugin/plugin-storage.ts
vince-fugnitto: You're right, we always make sure to properly initialize the cache so it is never `undefined`. 
I've updated the pull-request.

msujew (approved):
LGTM 👍

PR #11490, opened 2022-08-01
Clean up `any` type uses

msujew:
#### What it does

Removes a lot of `any` and `Object/object` references, cleans up the associated `eslint-disable` comment and replaces each instance with `unknown`, making the functions more robust against possible type-related issues.

#### How to test

1. CI is green and the application continues to work as expected

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/types-impl.ts
+++ b/packages/plugin-ext/src/plugin/types-impl.ts
@@ -20,7 +20,6 @@
 *--------------------------------------------------------------------------------------------*/
 
 /* eslint-disable no-null/no-null */
-/* eslint-disable @typescript-eslint/no-explicit-any */
 
 import { UUID } from '@theia/core/shared/@phosphor/coreutils';
 import { illegalArgument } from '../common/errors';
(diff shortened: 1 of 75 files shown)


packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -423,7 +423,7 @@ export class Position {
         return false;
     }
 
-    toJSON(): any {
+    toJSON(): unknown {
```
paul-marechal, line 426: `object` would make more sense here.

```suggestion
    toJSON(): object {
```

IIUC the original author didn't want to bother declaring the object being returned, hence the `any`, but using `unknown` instead feels odd.

packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -546,19 +546,17 @@ export class Range {
...
-        return Position.isPosition((<Range>thing).start)
-            && Position.isPosition((<Range>thing).end);
+        const range = thing as theia.Range;
+        return !!thing && typeof thing === 'object'
+            && Position.isPosition(range.start)
+            && Position.isPosition(range.end);
     }
 
-    toJSON(): any {
+    toJSON(): unknown {
```
paul-marechal, line 559:
```suggestion
    toJSON(): object {
```

paul-marechal (approved):
Changes look mostly good to me, see comments.

I find using `any` fine as we can often hit [TypeScript's type narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html)'s limits... But beside that I don't have very strong opinions about diligently using `unknown` instead of `any`.

Thanks for the work!

PR #10274, opened 2021-10-14
typedoc: fix docs for `@theia/plugin`

vince-fugnitto:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/10227

The pull-request fixes the generated typedoc documentation for the `@theia/plugin` extension including `theia.d.ts` and `theia-proposed.d.ts` following a [breaking change](https://github.com/TypeStrong/typedoc/releases/tag/v0.20.0) in typedoc regarding `modules`:

> Version 0.20 completely reworks how documentation is generated by TypeDoc. In previous versions, there was --mode file and --mode modules, which documented files according to their content on the filesystem. This worked reasonably well before ES modules were commonly used, but was insufficient for the modern ecosystem. In 0.20, TypeDoc documents your project according to what you export.

#### How to test

1, confirm that plugins still work correctly 
2. confirm that typings for `theia.d.ts` and `theia-proposed.d.ts` works correctly
3. performing a `yarn docs` should now generate the typedoc correctly for `@theia/plugins` (both `theia.d.ts` and `theia-proposed.d.ts` combined) - note `export NODE_OPTIONS=--max-old-space-size=8196` may be necessary:

image

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

Signed-off-by: vince-fugnitto <vincent.fugnitto@ericsson.com>

PR #11493, opened 2022-08-01
Clean Up `theia.d.ts`

colin-grant-work:
Depends on #11564.

#### What it does

There are a number of syntactic (misplaced `readonly` modifiers) and semantic (use of undeclared variables) problems in our `theia.d.ts` file. This cleans them up.

We can detect these problems by setting `skipLibCheck: false` in `packages/plugin/tsconfig.json`, but that makes it impossible to compile the application because we have a lot of conflicting declarations in our node modules.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin/src/theia-proposed.d.ts
+++ b/packages/plugin/src/theia-proposed.d.ts
@@ -20,7 +20,7 @@
  * This is the place for API experiments and proposals.
  * These API are NOT stable and subject to change. Use it on own risk.
  */
-export module '@theia/plugin' {
+declare module '@theia/plugin' {
     /**
      * The contiguous set of modified lines in a diff.
      */
@@ -333,6 +333,23 @@ export module '@theia/plugin' {
     }
 
     /**
+     * Options to specify the size of the result text preview.
+     * These options don't affect the size of the match itself, just the amount of preview text.
+     */
+    export interface TextSearchPreviewOptions {
+        /**
+         * The maximum number of lines in the preview.
+         * Only search providers that support multiline search will ever return more than one line in the match.
+         */
+        matchLines: number;
+
+        /**
+         * The maximum number of characters included per line.
+         */
+        charsPerLine: number;
+    }
+
+    /**
      * A line of context surrounding a TextSearchMatch.
      */
     export interface TextSearchContext {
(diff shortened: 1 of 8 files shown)


packages/plugin-ext/src/main/browser/authentication-main.ts
```diff
@@ -273,7 +254,7 @@ interface AccountUsage {
 
 export class AuthenticationProviderImpl implements AuthenticationProvider {
     private accounts = new Map<string, string[]>(); // Map account name to session ids
-    private sessions = new Map<string, string>(); // Map account id to name
+    private sessions = new Map<string, string>();   // Map account id to name
```
paul-marechal, line 257: nitpick: unnecessary change.

packages/plugin/src/theia-proposed.d.ts
```diff
@@ -705,3 +610,21 @@ export module '@theia/plugin' {
...
     // #endregion
 }
+
+/**
+ * Thenable is a common denominator between ES6 promises, Q, jquery.Deferred, WinJS.Promise,
+ * and others. This API makes no assumption about what promise library is being used which
+ * enables reusing existing code without migrating to a specific promise implementation. Still,
+ * we recommend the use of native promises which are available in this editor.
+ */
+interface Thenable<T> {
```
paul-marechal, line 620: Isn't this type already defined in `@theia/plugin`? Is it necessary to re-declare it here?

paul-marechal (commented):
Looks mostly good, thanks for the cleanup!

packages/plugin/src/theia-proposed.d.ts
colin-grant-work: I'm open to suggestions. It could be declared only in `theia.proposed.d.ts` and then picked up automatically in `theia.d.ts` but it's good for it to be in `theia.d.ts` for the use of the comparator.

packages/plugin/src/theia.d.ts
```diff
@@ -25,7 +25,7 @@ import './theia-proposed';
 /* eslint-disable @typescript-eslint/no-explicit-any */
 /* eslint-disable max-len */
 
-export module '@theia/plugin' {
+declare module '@theia/plugin' {
```
vince-fugnitto, line 28: This update unfortunately breaks our typedocs generation for `@theia/plugin` (fixed by https://github.com/eclipse-theia/theia/pull/10274). We may need to revisit the typedocs fix since it now produces a report with these changes like:

typedocs

vince-fugnitto (commented):
The changes overall look good to me 👍 The only thing I noticed was how our typedocs generation would fail to produce any meaningful content now for the `@theia/plugin` extension.

vince-fugnitto (approved):
LGTM 👍 It'll definitely help with all the false positives we used to see and would now be correctly identified with the comparator updates.

Issue #11271, opened 2022-06-08
Message Connection Stability Improvement


DanboDuan:
Should we(Theia) provide a stable connection between server and browser ?

Here is the situation :

Theia would close the message connection at once when networking is disconnected, e.g. the socket.io emit the `disconnect` event. Then the browser would reconnect and is treated as a new browser connected. Some doing job would be stopped . 
For most situation there would be no problem, but for some case, this should be annoy. For example, if we are debugging some program and the network does not work well just for some moment, we will have to restart the job.

If we could provide a stable connection between server and browser, and do some job to manage the connections and re-connections, like [open-vscode-server](https://github.com/gitpod-io/openvscode-server), this could be better.

It not a cost or hard job. We could just add a `PersistentConnection` base on `WebSocketChannel` and do some management, and it should work fine

kittaakos:
> For example, if we are debugging some program and the network does not work well just for some moment, we will have to restart the job.

I had a similar issue, @DanboDuan. Do you need this for dev only or production too? What's your use case? For me, it was a ping timeout in dev mode. I logged it here: https://github.com/eclipse-theia/theia/discussions/11543. Please chime in if you have the same problem.


> to manage the connections and re-connections

I develop electron apps, so there are no real connection issues between the backend and frontend, but even in dev mode, the reconnection happens automatically after a disconnect. In electron, there is nothing to manage. Theia does it. Do you experience no reconnect in web env?

DanboDuan:
> > For example, if we are debugging some program and the network does not work well just for some moment, we will have to restart the job.
> 
> I had a similar issue, @DanboDuan. Do you need this for dev only or production too? What's your use case? For me, it was a ping timeout in dev mode. I logged it here: #11543. Please chime in if you have the same problem.
> 
> > to manage the connections and re-connections
> 
> I develop electron apps, so there are no real connection issues between the backend and frontend, but even in dev mode, the reconnection happens automatically after a disconnect. In electron, there is nothing to manage. Theia does it. Do you experience no reconnect in web env?

Well, I develop browser apps. So it would be much easy to meet the issues.

tsmaeder:
I believe this request is acomplished with https://github.com/eclipse-theia/theia/pull/13082

PR #11499, opened 2022-08-03
support persistent websocket

DanboDuan:
Signed-off-by: bob <bob170731@gmail.com>

#### What it does

Support persistent websocket connection, which allow network to be weak and will not fire close event for shortly disconnected.

#### How to test

just open an ide in browser and fake network weakness by turning the computer with network-off mode, then turn on the network and the ide should continue to work fine.

#### Review checklist

#### Reminder for reviewers


DanboDuan:
for [#11271](https://github.com/eclipse-theia/theia/issues/11271)

DanboDuan:
would you like to help review this @msujew @paul-marechal

DanboDuan:
> I think these kind of changes require a much more thorough explanation of how it works.
> 
> I don't want to reverse engineer code to guess things like: How are connections identified? How are "backend sessions" persisted? How long may a session persist? Is there anything worth mentioning regarding the approach taken?
> 
> I agree that the connection infrastructure is in need of more robustness, but please help us understand your solution.


The reason for the unreliable message is that the message layer close the socket once the `socket.io` disconnects.
```ts
    protected toIWebSocket(socket: Socket): IWebSocket {
        return {
            close: () => {
                socket.removeAllListeners('disconnect');
                socket.removeAllListeners('error');
                socket.removeAllListeners('message');
                socket.disconnect();
            },
            isConnected: () => socket.connected,
            onClose: cb => socket.on('disconnect', reason => cb(reason)),
            onError: cb => socket.on('error', error => cb(error)),
            onMessage: cb => socket.on('message', data => cb(data)),
            send: message => socket.emit('message', message)
        };
    }
```

so if we do not close the socket at once and try to reconnect from frontend to backend, we can keep the message reliable.

So we do as following:

1. Generate a unique `persistent_key` for `socket.io` URL
2. The backend keep a map of `socket.io`. once a socket connected, check if it's reconnection or first connecting using the `persistent_key`
3. Once the `socket.io` disconnected, wait for a certain timeout to emit the close event while the frontend might re-connect


So we build a `PersistentWebSocket` to hold the underlying socket.

packages/core/src/common/messaging/abstract-connection-provider.ts
```diff
@@ -78,6 +78,9 @@ export abstract class AbstractConnectionProvider<AbstractOptions extends object>
     protected reconnectChannelOpeners: Array<() => Promise<void>> = [];
 
     protected initializeMultiplexer(): void {
+        if (!!this.channelMultiplexer) {
+            return;
+        }
```
paul-marechal, line 83:
```suggestion
        if (this.channelMultiplexer) {
            return;
        }
```

packages/core/src/common/messaging/web-socket-channel.ts
```diff
@@ -112,5 +113,130 @@ export interface IWebSocket {
...
+        this.underlyingSocketConnected = false;
+        this._onCloseReason = reason;
+        // The socket has closed, let's give the renderer a certain amount of time to reconnect
+        if (!this._disconnectRunner.isScheduled()) {
+            this._disconnectRunner.schedule();
+        }
+        this._onSocketDisconnect.fire();
+    }
+
+    protected onReConnect(): void {
```
paul-marechal, line 193:
```suggestion
    protected handleReconnect(): void {
```

packages/core/src/common/messaging/web-socket-channel.ts
```diff
@@ -112,5 +113,130 @@ export interface IWebSocket {
...
+     */
+    onConnect(cb: () => void): void;
 }
 
+/**
+ * An persist websocket implementation base on the `{@link IWebSocket} .
+ * For the client side, it will always try reconnect to remote .
+ * For the sever side, it will fire close state after the `ReconnectionGraceTime` time
+ */
+export class PersistentWebSocket implements IWebSocket {
```
paul-marechal, line 128:
```suggestion
/**
 * A persistant WebSocket implementation based on the `{@link IWebSocket}.
 * For the client side, it will always try to reconnect to the remote.
 * For the sever side, it will close itself after the `ReconnectionGraceTime` time.
 */
export class PersistentWebSocket implements IWebSocket {
```

packages/core/src/common/scheduler.ts
```diff
@@ -0,0 +1,87 @@
...
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+// *****************************************************************************
+
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export class ProcessTimeRunOnceScheduler {
```
paul-marechal, line 22: I'm not convinced we need this copied code.

packages/core/src/node/messaging/messaging-contribution.ts
```diff
@@ -107,7 +109,26 @@ export class MessagingContribution implements BackendApplicationContribution, Me
     }
 
     protected handleChannels(socket: Socket): void {
-        const socketChannel = new WebSocketChannel(this.toIWebSocket(socket));
+        const reconnectionKey = new URL(socket.request.url!, 'http://localhost/').searchParams.get(PersistentWebSocket.ReconnectionKey) || '';
+        const toClose = new Array<PersistentWebSocket>();
```
paul-marechal, line 113:
```suggestion
        const toClose: PersistentWebSocket[] = [];
```

packages/core/src/node/messaging/messaging-contribution.ts
```diff
@@ -107,7 +109,26 @@ export class MessagingContribution implements BackendApplicationContribution, Me
...
+        this.persistentConnections.forEach((value, key) => {
```
paul-marechal, line 114:
```suggestion
        this.persistentConnections.forEach((persistentConnection, connectionKey) => {
```

packages/core/src/node/messaging/messaging-contribution.ts
```diff
@@ -107,7 +109,26 @@ export class MessagingContribution implements BackendApplicationContribution, Me
...
+            if (key !== reconnectionKey && !value.underlyingSocketConnected) {
+                toClose.push(value);
+            }
+        });
+        // close other non-connected connections when there is a connection in
+        toClose.forEach(value => value.fireClose());
```
paul-marechal, line 120: I don't understand this logic:

1. We handle 1 channel reconnection with some key
2. For all opened connections that don't match that key we fire the close event on those??
3. Looking at the implementation for `fireClose` it seems to shortcut `PersistentWebSocket`'s internal state, as in it doesn't do anything but just fires the event?

paul-marechal:
Given the current `ReconnectionGraceTime` set at 10 min, do I understand correctly that we will leak backend-scoped inversify containers that will live for 10 min each time the user reloads the browser? This means any connection-scoped process like the plugin host process, language server processes, etc? If so I'm not a fan.

packages/core/src/node/messaging/messaging-contribution.ts
paul-marechal: Here's what I understand:

1. I have 3 Theia tabs opened to the same backend on different workspaces
2. I lose connection to the backend in all 3 tabs
3. One tab reconnects before the others
4. The 2 other slower tabs get their connection closed?

packages/core/src/node/messaging/messaging-contribution.ts
```diff
@@ -107,7 +109,26 @@ export class MessagingContribution implements BackendApplicationContribution, Me
     }
 
     protected handleChannels(socket: Socket): void {
-        const socketChannel = new WebSocketChannel(this.toIWebSocket(socket));
+        const reconnectionKey = new URL(socket.request.url!, 'http://localhost/').searchParams.get(PersistentWebSocket.ReconnectionKey) || '';
```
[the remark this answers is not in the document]
DanboDuan: just any host is ok, we only need it to be valid url so that we could get query keys & values. Or we could us `qs` to parse query data

packages/core/src/node/messaging/messaging-contribution.ts
DanboDuan: we can change the behavior, to wait for every connection for the same timeout

DanboDuan:
> Given the current `ReconnectionGraceTime` set at 10 min, do I understand correctly that we will leak backend-scoped inversify containers that will live for 10 min each time the user reloads the browser? This means any connection-scoped process like the plugin host process, language server processes, etc? If so I'm not a fan.

that is why we could close other no-connected connections while some connection re-connect to stop waiting

safisa:
Hi,
Any updates on this PR?

This change could help us a lot with disconnections issues since today after reconnecting all the plugins are reloaded/reactivated again and this takes a lot of time (e.g. RedHat Java plugins) also we have a vscode plugin that uses many webviews and custom editors that are fully reloaded cause bad UX.

tortmayr:
I'm happy to take another look at this PR.
@DanboDuan I know that this PR has been stale for a while, nevertheless I hope that you are still able/willing to work on this. If so could you please rebase this PR to resolve the conflicts?

>Given the current ReconnectionGraceTime set at 10 min, do I understand correctly that we will leak backend-scoped inversify containers that will live for 10 min each time the user reloads the browser? This means any connection-scoped process like the plugin host process, language server processes, etc? If so I'm not a fan.

I have to take a deeper look into this but I'm pretty confident that we can find a way to let us distinguish between graceful/intended websocket connection closes and actual connection loses. This way we could avoid leaking of backend-scoped containers on reload.

@safisa 
Could you please eloborate on the disconnection issues that you are encoutering recently? This PR is definitely a big step into the right direction but I also want to make sure that there are no additional underlying problems that should be addressed.

DanboDuan:
> I'm happy to take another look at this PR. @DanboDuan I know that this PR has been stale for a while, nevertheless I hope that you are still able/willing to work on this. If so could you please rebase this PR to resolve the conflicts?
> 
> > Given the current ReconnectionGraceTime set at 10 min, do I understand correctly that we will leak backend-scoped inversify containers that will live for 10 min each time the user reloads the browser? This means any connection-scoped process like the plugin host process, language server processes, etc? If so I'm not a fan.
> 
> I have to take a deeper look into this but I'm pretty confident that we can find a way to let us distinguish between graceful/intended websocket connection closes and actual connection loses. This way we could avoid leaking of backend-scoped containers on reload.
> 
> @safisa Could you please eloborate on the disconnection issues that you are encoutering recently? This PR is definitely a big step into the right direction but I also want to make sure that there are no additional underlying problems that should be addressed.

@tortmayr I'm now away. But @FinnChen would continue to help it.

safisa:
Hi @tortmayr 

This disconnection can cause several issues, but I am not sure about all of them. I will try to mention the most annoying issues we faced:

1. biggest issue we faced, is the web views and custom editors we are using, those web views are reloaded (on online), and this is a very bad UX.
2. vscode plugins are activated again on online. Some of the plugins have issues after this re-activation. E.g. we have a log-viewer (berublan.vscode-log-viewer-0.13.1), that stopped working after re-connection.
3. re-activations of vscode plugins take time and resources (bad UX), e.g try the RedHat java plugin. it is a heavy one and takes some time to be loaded (not sure about the side effects of this re-activation on its functionality).

Thanks

safisa:
Hi @DanboDuan,

First thank you for this work in the PR.
I want to know if you want to continue working on it. We are wishing to have support for persistent websockets in Theia to help us resolve many issues like the ones mentioned above.

Thanks in advance

msujew:
I am not totally sure whether this is the right approach to resolving the issues that have been brought up. I agree that something needs fixing and I appreciate the work that went into this PR, but keeping the connection alive (and everything associated with it, like the plugin host) seems a bit heavy handed for me.

I would much rather see a clean reconnection (including a restart of the plugin host) than trying to keep everything alive. I won't prevent other committers from reviewing/approving this PR, but I'm more inclined to close this to find another solution to the issues. Unless there are objections, I'll close this in the near future.

Issue #11521, opened 2022-08-03
[vscode] Implement `window.registerTerminalLinkProvider`


planger:
The API exists, but it doesn't seem to be implemented actually:

https://github.com/eclipse-theia/theia/blob/7e177f92c6d3f32dbc876d25ec380bd403e13858/packages/plugin-ext/src/plugin/plugin-context.ts#L503-L505

The type `TerminalLink` already exists, however it is a class with constructor in VS Code, see https://github.com/eclipse-theia/theia/issues/11507

planger:
I'll look into providing the actual functionality for `window.registerTerminalLinkProvider`. Also, in a first glance, I think it makes sense to harmonize this implementation with [`AbstractCmdClickTerminalContribution`](https://github.com/eclipse-theia/theia/blob/master/packages/terminal/src/browser/terminal-linkmatcher.ts#L25) and try to get rid of the usage of the deprecated xterm `registerLinkMatcher` API.

PR #11522, opened 2022-08-03
vscode: Make `TerminalLink` compatible to VS Code 1.64.2

### LINKED ISSUE #11507: [vscode] `TerminalLink` turned into a class with constructor
See [`TerminalLink`](https://code.visualstudio.com/api/references/vscode-api#TerminalLink)

planger:
#### What it does
Replaces the out-dated definition of the `TerminalLink` interface with the current class that also provides a constructor.

In the course of fixing this, I discovered #11521. So it looks like Theia doesn't support registering `TerminalLinkProvider` at the moment. Anyway, making `TerminalLink` compatible to VS Code will help anyway, but only really gets effective once #11521 is fixed.

Fixes #11507

#### How to test

There is nothing really to test as `TerminalLink` isn't used in the Theia code base anyway and now its definition is at least equivalent to how it is defined in VS Code `1.64.2`.
Also I tested that this change makes the check in the [`vscode-theia-comparator`](https://github.com/eclipse-theia/vscode-theia-comparator) happy.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin/src/theia.d.ts
+++ b/packages/plugin/src/theia.d.ts
@@ -3097,7 +3097,7 @@ export module '@theia/plugin' {
     /**
      * A link on a terminal line.
      */
-    export interface TerminalLink {
+    export class TerminalLink {
         /**
          * The start index of the link on [TerminalLinkContext.line](#TerminalLinkContext.line].
          */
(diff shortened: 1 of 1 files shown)


vince-fugnitto (commented):
@planger should we not update `TerminalLink` once https://github.com/eclipse-theia/theia/issues/11521 is fixed? I don't see a strong reason to update them separately and would rather have full support.

planger:
@vince-fugnitto Sure, we can wait, if you prefer. Since terminal link providers of plugins that register a provider will never be called by Theia, I guess it doesn't make a difference whether `TerminalLink` that they would create is already 100 % compatible or not. Just thought, it doesn't take effort to already adjust this right away. But as you prefer... Should I close this PR and link #11507 in #11521?

vince-fugnitto:
@planger I'd prefer if we link the two yes, it gives a better representation that it is actually supported and they are tied together.

planger:
Closing this to wait for an actual implementation of the `TerminalLinkProvider` integration and registration, tracked by https://github.com/eclipse-theia/theia/issues/11521.

planger:
This change is now part of https://github.com/eclipse-theia/theia/pull/11552

Issue #11525, opened 2022-08-04
Automatic translations not available for external contributions


msujew:
### Bug Description:

When someone creates a PR based on a fork of Theia, the `Automatic translation` workflow doesn't run, leaving their work untranslated. This also leads to further issues down the line as soon as someone else tries to create a PR, as it will pick up the untranslated changes.

### Steps to Reproduce:

1. Fork Theia and create a new feature branch in the fork
2. Add a new `nls.localize` key
3. Create a PR, the `Automatic translation` workflow doesn't run

### Additional Information

- Operating System: GitHub
- Theia Version: 1.28.0

PR #11526, opened 2022-08-04
Run automatic translation on `workflow_dispatch`

msujew:
#### What it does

Fixes https://github.com/eclipse-theia/theia/issues/11525

Instead of running on `push` (which only really works in our own repo), we instead create a workflow, which we'll manually start before creating a new release.

#### How to test

After merging this, a new workflow dispatch should be available, which automatically performs the translation update and creates a new PR based on the results. See https://github.com/msujew/theia/pull/7 for an example of how this would look like.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.github/workflows/translation.yml
+++ b/.github/workflows/translation.yml
@@ -42,11 +42,20 @@ jobs:
         env:
           DEEPL_API_TOKEN: ${{ secrets.DEEPL_API_TOKEN }}
 
+      - name: Get Actor User Data
+        uses: octokit/request-action@v2.x
+        id: actor_user_data
+        with:
+          route: GET /users/{user}
+          user: ${{ github.actor }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v4
         with:
-          commiter: Mark Sujew <mark.sujew@typefox.io>
-          author: Mark Sujew <mark.sujew@typefox.io>
+          commiter: ${{ github.actor }} <${{ fromJson(steps.actor_user_data.outputs.data).email }}>
+          author: ${{ github.actor }} <${{ fromJson(steps.actor_user_data.outputs.data).email }}>
           branch: bot/translation-update
           title: Translation update for version ${{ env.NEXT_VERSION_NUMBER }}
           commit-message: Translation update for version ${{ env.NEXT_VERSION_NUMBER }}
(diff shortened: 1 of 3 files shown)


vince-fugnitto:
@msujew It might not even be possible from forks  as the `GITHUB_TOKEN` only has `read` permissions for that special case:

> https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

msujew:
@vince-fugnitto Yeah, I noticed multiple issues already, forks also (unsurprisingly) can't read from our secrets. I'm wondering whether my other idea would be better suited, as it only runs on our repo. The idea was to push to another branch when pushing to `master` (or merging into it) and merge this one when performing a release.

Anyway, I'm first looking into whether we can force push from Actions to a protected branch, since that would allow us to perform automatic translations after merging and simply amend it to the squashed commit, thereby preserving the history of why translations were added a bit better.

vince-fugnitto:
@msujew if we wanted, we can add a `workflow_dispatch` (so we can manually trigger the CI from the actions page) and do so on release as you mentioned (just need to document it in the release page). We can try to make the CI create pull-request with the updates. I'll leave the final decision up to you however :)

msujew:
@vince-fugnitto Thanks for the hint! Seems like we also can't push, so automatically creating a PR sounds like the best approach. Works quite well, see https://github.com/msujew/theia/pull/5.

Quick question: Do you have an idea whether there's some Eclipse bot/fake email that's covered by the ECA? I'm using my own for now, but that seems a bit weird to me.

vince-fugnitto:
> Quick question: Do you have an idea whether there's some Eclipse bot/fake email that's covered by the ECA? I'm using my own for now, but that seems a bit weird to me.

I'm not aware of any bot account :( Is it possible to use the `actor`?

vince-fugnitto:
@JonasHelming are you aware of any bot account in the foundation (covered by the ECA) that we can use to perform the automatic translation updates?

msujew:
> Is it possible to use the actor?

Unfortunately, `github.actor` only gives us the user name. However, in theory we should be able to get the email (if set) based on the user name by querying the github API. I'll see what I can do.

Anyway, I've also updated the `publishing.md` in my last push, this is ready for a review.

msujew:
@vince-fugnitto Running Github API requests inside of actions is surprisingly easy, the action now will always commit using the email of the user that triggered the workflow.

msujew:
@vince-fugnitto Thanks for the quick review. Please let me know if you find any issues once you run this. Maybe we should run this a day before the release to fix any potential issues?

PR #11534, opened 2022-08-04
use `Generator<Yield, Return, Next>`

paul-marechal:
Replace occurences of `IterableIterator<T>` as return type for generator
functions with `Generator<Yield, Return, Next>` as it better describes
what we are dealing with.

#### How to test

Static change only, it should build as before.

#### Review checklist

#### Reminder for reviewers


paul-marechal:
I happened to write a generator function and noticed that this type exists, figured we could use it in the relevant places.

packages/core/src/browser/tree/tree-model.ts
```diff
@@ -260,7 +260,7 @@ export class TreeModelImpl implements TreeModel, SelectionProvider<ReadonlyArray
         return undefined;
     }
 
-    protected *getExpansionCandidates(raw?: Readonly<TreeNode>): IterableIterator<TreeNode | undefined> {
+    protected *getExpansionCandidates(raw?: Readonly<TreeNode>): Generator<TreeNode | undefined, void, undefined> {
```
colin-grant-work, line 263: This suggests that this change is not necessarily desirable: that is, the `Generator` type is more specific than the `IterableIterator`, and rewriting `IterableIterator<T>` as `Generator<T>` as is done in most places in this PR is potentially breaking: `Array<T>.values()` (or `*yield someArray<T>`) satisfies `IterableIterator<T>` but not `Generator<T>`. See [this playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBApgDwIYFsAOAbOMC8MxwDuMAggE5lICeAPGAK4oBGcZAfABQCUA3AFB9QkWADMkASwwQAXDADicApSggydRi3a54ydFgB0ANyQZ6cCN36Dw0GBHrBgcOABMZMAJJRWSJli8+KmoMzKxs2oiomHBGJmYWvHwwMEA) for a simple demonstration.

colin-grant-work (changes requested):
In general, I'm opposed to this change. It's technically breaking in every case, and I don't see a need for it unless the greater flexibility of the `Generator` type, which is just the same as `IterableIterator` with the additional type metadata of `Iterator` exposed, is actually being used.

packages/core/src/browser/tree/tree-model.ts
paul-marechal: The issue with your playground is that we would never try to assign `IterableIterator<T>` to a `Generator<...>` type.

[But `Generator` is assignable to `IterableIterator`](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAKgOYFMzoE4EMpzYAUAlAFyIDimO+hAPGCALYBGOAfIgN4BQiAxAE8Y6ADYATRAEZeAXwDcvXhAQBnKIjUgIEdOglqKASSi1WY9KdoFsjFu2xcAvIgxY8t0kqA).

I guess the most annoying bit from `Generator` is how they typed `TNext = unknown` as it makes it incompatible with `IterableIterator` as it has `TNext = undefined`... But occurrences of this problem are low, and can be fixed by specifying the templates types.

colin-grant-work: > [But `Generator` is assignable to `IterableIterator`](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAKgOYFMzoE4EMpzYAUAlAFyIDimO+hAPGCALYBGOAfIgN4BQiAxAE8Y6ADYATRAEZeAXwDcvXhAQBnKIjUgIEdOglqKASSi1WY9KdoFsjFu2xcAvIgxY8t0kqA).

Exactly - so we should stick with `IterableIterator` :-). I do agree that it's a pain that they typed them in a way that means they aren't compatible by default when they didn't really have to do that, but given that they have, we're safer sticking with `IterableIterator` where we've been using it, though I'm not opposed to using `Generator` where appropriate in the future.

paul-marechal: > though I'm not opposed to using Generator where appropriate in the future.

Will do then :)

I'd still prefer to changes previous occurrences, but this isn't really a hill I'm willing to die on.

PR #11537, opened 2022-08-05
First cut of VS Code usage policy

tsmaeder:
#### What it does
Draft PR to allow discussion of usage policy for code from VS Code

### DIFF
--- a/doc/vscode-usage.md
+++ b/doc/vscode-usage.md
@@ -1,0 +1,23 @@
+## Using code from VS Code
+
+Since its inception, Theia has used the "Monaco" editor component from VS Code. With the recent move to using ECMAScript modules, consuming code from the VS Code project has become much easier and safer. But while reusing code saves us work, there is also a downside to it. Monaco has a relatively stable external API because Microsoft also releases it as a stand-alone editor component. But other parts of the code base may change more frequently and in unexpected ways. We always use the same version of all modules making up VS Code. So when we update VS Code, often to provide a new feature in Monaco to our adopters, we will have to deal with all the API changes at that same time. As an example: Theia used the quick-input component from VS Code directly to implement it's own quick-input component. Because the component was not encapsulated in any way, the updating Monaco to a new version became difficult and time-consuming. 
+
+So while we don't  prohibit the use of code from VS Code (other than the Monaco editor API), we have the following goals:
+
+* Updating Monaco should not impact adopters
+
+* Adoption of a new Monaco version should generally be a straightforward, quick process (< 1 week of work)
+
+In order to achieve those goals, follow these simple rules:
+
+* Never export a type, function or variable from an internal VS Code API from a theia package
+
+* Don't use code from VS Code that you could not easily copy into the Theia codebase if the need arises. 
+
+The first rule shields our adopters from having to change their code in response to updating Monaco. If they cannot see an object or type, they cannot rely on its existence or API. Note that this includes functions, supertype relationships or parameter and return types. If you need to export functionality, export an interface in Theia and import that interface using imported code from VS Code. This way, we can build adapters to shield against API changes. The rule also prevents spreading dependencies on VS Code in our code without our being aware of them.  While it's not technically possible to enforce non-export of stuff in our current build system, we should make sure we're not exporting tainted code through `index.ts` or similar mechanisms. At the very least, we should not require package users to rely on VS Code stuff.
+
+The second rule  ensures that we do not rely on VS Code stuff that is deeply coupled with other parts of VS Code that we don't want to import. It gives us the escape hatch of just copying the old version of the code and filing a CQ with the Eclipse foundation if updating to the version of stuff from VS Code is not what we want (because it doesn't fit our needs or takes to long).
+
+Tip: you can find locations where VS code is used by searching for import statements from `@theia/monaco-editor-core/esm/vs`.
+
+


doc/vscode-usage.md
```diff
@@ -0,0 +1,23 @@
...
+
+The second rule  ensures that we do not rely on VS Code stuff that is deeply coupled with other parts of VS Code which we don't want to import. It gives us the escape hatch of: if updating to the version of stuff from VS Code is not what we want (because it doesn't fit our needs or takes to long), we can always just copy the old version of the code and file a CQ with the Eclipse foundation.
```
colin-grant-work, line 19: > deeply coupled with other parts of VS Code which we don't want to import.

I don't know if this really carries weight. We're starting with a tree-shaken (gently - admittedly we could be [more aggressive](https://github.com/theia-ide/vscode/commit/a16ce8a4e8aab5ee151140e9ed16bff00562c2c4#diff-ede8d9dd0ad352e9aa350bcf8c008e46b188464250fced658358867125c1ea3a)), heavily reduced version of VSCode in `monaco-editor-core`, and we import the public API (as well some internal stuff), which is the basis for the tree shaking. That means that anyone who depends on `monaco-editor-core` through Theia will be getting at least every file in that package - if they tree-shake the application more thoroughly, it's possible they don't get every line of code, but they get at least every file - so there aren't really any big parts of VSCode that are in the package the user isn't _already_ importing.

Given that, I think the only real question is stability / maintenance. I absolutely agree that internal types from Monaco shouldn't be exported unadapted to adopters, but I think the only other real criterion to consider is the estimated cost of maintaining whatever API (adapted, reimplemented, or copied) that we do present to the world.

colin-grant-work (commented):
Nice work. I've left some copy editing comments and some thoughts.

doc/vscode-usage.md
```diff
@@ -0,0 +1,23 @@
+## Using code from VS Code
+
+Since its inception, Theia has used the "Monaco" editor component from VS Code. With the recent move to using ECMAScript modules, consuming code from the VS Code project has become much easier and safer. But while reusing code saves us work, there is also a down-side to it. Monaco has a relatively stable external API because Microsoft also releases it as a stand-alone editor component. But other parts of the code base may change more frequently and in unexpected ways. We always use the same version of all modules making up VS Code. So when we update VS Code, often to provide a new feature in Monaco to our adopters, we will have to deal with all the API changes at that same time. As an example: Theia used the quick-input component from VS Code directly to implement it's own quick-input component. Because the component was not encapsulated in any way, the updating Monaco to a new version became difficult and time-consuming. 
+
+So while we don't  prohibit the use of code from VS Code (outside of Monaco), we have the following goals:
```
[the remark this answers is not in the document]
tsmaeder: I think "(other than the Monaco editor API)" better captures the intent.

doc/vscode-usage.md
```diff
@@ -0,0 +1,23 @@
...
+
+* Updating Monaco should not impact adopters
+
+* Adoption of a new Monaco version should generally be a straightforward, quick process (< 1 week of work)
+
+In order to achieve those goals, follow these simple rules:
+
+* Never export a type, function or variable from VS Code from a theia package
```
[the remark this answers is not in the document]
tsmaeder: Some of the Code I VS Code has nothing to do with the Monaco Editor, see for example the quick-input component I mentioned in the intro .

doc/vscode-usage.md
```diff
@@ -0,0 +1,23 @@
...
+
+* Don't use code from VS Code that you could not easily copy into the Theia codebase if the need arises.
```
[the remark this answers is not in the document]
tsmaeder: The idea is to not _increase_ coupling to the VS Code internals. Updating to a new VS version should be a trivial task we can do frequently without undue burden. If we use large parts of unstable VS code (unstable in the sense that the API changes), we will have to update our usages each time we update to a new VS Code version. The "escape hatch" from this coupling (if the code in VS goes into a direction we don't want to follow). Therefore, "I could easily copy this over if I need to" seems to be a good heuristic to me. .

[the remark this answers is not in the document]
tsmaeder: If we had copies of utilities in the code we wrote ourselves, why would we not have copies of utilities when we import them from VS? The problem seems to be that people are not aware a thing already exists. I would expect folks to be _more_ familiar with the Theia code base than with the VS codebase, thus creating _less_ copies.

doc/vscode-usage.md
tsmaeder: Dead code elimination is fine as far as it goes, but even just importing a simple feature (again: quick-input), might use large swathes of code in it's implementation. We just don't know just by looking at the API. We _do_ know with the Monaco editor, because the VS Code team themselves want to make sure it has relatively compact dependencies, since the publish it as a distinct package.

MatthewKhouzam (approved):
I like the clarifications here. It makes it clear how we are to work.

JonasHelming (approved):

doc/vscode-usage.md
```diff
@@ -0,0 +1,23 @@
+## Using code from VS Code
+
+Since its inception, Theia has used the "Monaco" editor component from VS Code. With the recent move to using ECMAScript modules, consuming code from the VS Code project has become much easier and safer. But while reusing code saves us work, there is also a down-side to it. Monaco has a relatively stable external API because Microsoft also releases it as a stand-alone editor component. But other parts of the code base may change more frequently and in unexpected ways. We always use the same version of all modules making up VS Code. So when we update VS Code, often to provide a new feature in Monaco to our adopters, we will have to deal with all the API changes at that same time. As an example: Theia used the quick-input component from VS Code directly to implement it's own quick-input component. Because the component was not encapsulated in any way, the updating Monaco to a new version became difficult and time-consuming.
```
vince-fugnitto, line 3: minor: 
```suggestion
Since its inception, Theia has used the "Monaco" editor component from VS Code. With the recent move to using ECMAScript modules, consuming code from the VS Code project has become much easier and safer. But while reusing code saves us work, there is also a downside to it. Monaco has a relatively stable external API because Microsoft also releases it as a standalone editor component. But other parts of the code base may change more frequently and in unexpected ways. We always use the same version of all modules making up VS Code. So when we update VS Code, often to provide a new feature in Monaco to our adopters, we will have to deal with all the API changes at that same time. As an example: Theia used the quick-input component from VS Code directly to implement it's own quick-input component. Because the component was not encapsulated in any way, the updating Monaco to a new version became difficult and time-consuming. 
```

doc/vscode-usage.md
```diff
@@ -0,0 +1,23 @@
...
+
+* Adoption of a new Monaco version should generally be a straightforward, quick process (< 1 week of work)
+
+In order to achieve those goals, follow these simple rules:
+
+* Never export a type, function or variable from VS Code from a theia package
+
+* Don't use code from VS Code that you could not easily copy into the Theia codebase if the need arises. 
+
+The first rule shields our adopters from having to change their code in response to updating Monaco. If they cannot see an object or type, they cannot rely on its existence or API. Note that this includes functions, supertype relationships or parameter and return types. If you need to export functionality, export an interface in Theia and import that interface using imported code from VS Code. This way, we can build adapters to shield against API changes. The rule also prevents spreading dependencies on VS Code in our code without our being aware of them.  While it's not technically possible to enforce non-export of stuff in our current build system, we should make sure we're not exporting tainted code through `index.ts` or similar mechnisms. At the very least, we should not require package users to rely on VS Code stuff.
```
vince-fugnitto, line 17: minor:

```suggestion
The first rule shields our adopters from having to change their code in response to updating Monaco. If they cannot see an object or type, they cannot rely on its existence or API. Note that this includes functions, supertype relationships or parameter and return types. If you need to export functionality, export an interface in Theia and import that interface using imported code from VS Code. This way, we can build adapters to shield against API changes. The rule also prevents spreading dependencies on VS Code in our code without our being aware of them.  While it's not technically possible to enforce non-export of stuff in our current build system, we should make sure we're not exporting tainted code through `index.ts` or similar mechanisms. At the very least, we should not require package users to rely on VS Code stuff.
```

vince-fugnitto (commented):
I believe the ECA should pass if you squash and force-push, it's likely the `no-reply` email that is causing problems:

```
From: =?UTF-8?q?Thomas=20M=C3=A4der?= <tsmaeder@users.noreply.github.com>
```

doc/vscode-usage.md
tsmaeder: "downside" yes, but "stand-alone" ist still more common according to the internet.

PR #11546, opened 2022-08-08
Add confirmation message for debug exit

colin-grant-work:
#### What it does

At the moment, if the preferences are set to prevent exit with running debug sessions, we have no message specifically for that case. This PR adds such a message.

#### How to test

1. Start a debug session.
2. Set the preference `debug.confirmOnExit` to `always`
3. Try to exit with the debug session running.
4. You should get a confirmation that mentions the running session.
5. Select an option. If you choose to cancel, the application should not exit and the debug session should not be terminated; if you choose to stop debugging, the application should exit.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-frontend-application-contribution.ts
+++ b/packages/debug/src/browser/debug-frontend-application-contribution.ts
@@ -14,7 +14,9 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { AbstractViewContribution, KeybindingRegistry, Widget, CompositeTreeNode, LabelProvider, codicon } from '@theia/core/lib/browser';
+import {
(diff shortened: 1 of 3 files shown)


jfaltermeier (commented):
Hi, I've tested the changes with the browser app and electron and it is working as expected as far as I can see. 

I've noticed one issue with the Close Workspace action however. The reproducer looks like this:
Browser App: MenuBar -> File -> Close Workspace -> OK in Dialog whether to close workspace -> Cancel on Dialog about active debug sessions -> Close Workspace Action in Menu stays disabled
In the Electron App the menu entry stays enabled, but clicking it does not try to close the workspace anymore.

We probably have to show the dialog about active debug sessions before the close workspace dialog?

colin-grant-work:
@jfaltermeier, I've modified the code that handles `WorkspaceService.close()` to make sure that it doesn't perform any of the actions involved in closing the workspace until it has checked whether it is safe to do so. That should address the problem you encountered.

jfaltermeier (approved):
Thanks, looks good to me!

PR #11547, opened 2022-08-09
core: add support for typed `Command`

paul-marechal:
There is currently no typings for commands, which means users that wish
to execute commands must know/guess what the command arguments and
return value are without much telling them what those are precisely.

Add a new type `CommandId<Arguments, ReturnType>`.

Have support optional generics `Command<Arguments, ReturnType>`.

Update `CommandService` and `CommandRegistry` to deal with this new
information.

#### How to test

TBD

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/callhierarchy/src/browser/callhierarchy-contribution.ts
+++ b/packages/callhierarchy/src/browser/callhierarchy-contribution.ts
@@ -28,7 +28,7 @@ export const CALL_HIERARCHY_TOGGLE_COMMAND_ID = 'callhierarchy:toggle';
 export const CALL_HIERARCHY_LABEL = nls.localizeByDefault('Call Hierarchy');
 
 export namespace CallHierarchyCommands {
-    export const OPEN = Command.toLocalizedCommand<[], CallHierarchyTreeWidget>({
+    export const OPEN = Command.toLocalizedCommand<() => CallHierarchyTreeWidget>({
         id: 'callhierarchy:open',
         label: 'Open Call Hierarchy'
     }, 'theia/callhierarchy/open');
--- a/packages/callhierarchy/src/browser/callhierarchy-contribution.ts
+++ b/packages/callhierarchy/src/browser/callhierarchy-contribution.ts
@@ -28,7 +28,7 @@ export const CALL_HIERARCHY_TOGGLE_COMMAND_ID = 'callhierarchy:toggle';
 export const CALL_HIERARCHY_LABEL = nls.localizeByDefault('Call Hierarchy');
 
 export namespace CallHierarchyCommands {
-    export const OPEN = Command.toLocalizedCommand({
+    export const OPEN = Command.toLocalizedCommand<[], CallHierarchyTreeWidget>({
         id: 'callhierarchy:open',
         label: 'Open Call Hierarchy'
     }, 'theia/callhierarchy/open');
--- a/packages/core/src/browser/common-frontend-contribution.ts
+++ b/packages/core/src/browser/common-frontend-contribution.ts
@@ -148,13 +148,13 @@ export namespace CommonCommands {
         label: 'Select All'
     });
 
-    // ???
-    export const FIND = Command.toDefaultLocalizedCommand<[], void>({
+    // TODO: type command
+    export const FIND = Command.toDefaultLocalizedCommand({
         id: 'core.find',
         label: 'Find'
     });
-    // ???
-    export const REPLACE = Command.toDefaultLocalizedCommand<[], void>({
+    // TODO: type command
+    export const REPLACE = Command.toDefaultLocalizedCommand({
         id: 'core.replace',
         label: 'Replace'
     });
--- a/packages/core/src/browser/common-frontend-contribution.ts
+++ b/packages/core/src/browser/common-frontend-contribution.ts
@@ -117,216 +117,218 @@ export namespace CommonCommands {
         id: 'core.open',
     };
 
-    export const CUT = Command.toDefaultLocalizedCommand({
+    export const CUT = Command.toDefaultLocalizedCommand<[], void>({
         id: 'core.cut',
         label: 'Cut'
     });
-    export const COPY = Command.toDefaultLocalizedCommand({
+    export const COPY = Command.toDefaultLocalizedCommand<[], void>({
         id: 'core.copy',
         label: 'Copy'
     });
-    export const PASTE = Command.toDefaultLocalizedCommand({
+    export const PASTE = Command.toDefaultLocalizedCommand<[], void>({
         id: 'core.paste',
         label: 'Paste'
     });
 
-    export const COPY_PATH = Command.toDefaultLocalizedCommand({
+    export const COPY_PATH = Command.toDefaultLocalizedCommand<[uris: URI[]], void>({
         id: 'core.copy.path',
         label: 'Copy Path'
     });
 
-    export const UNDO = Command.toDefaultLocalizedCommand({
+    export const UNDO = Command.toDefaultLocalizedCommand<[], void>({
         id: 'core.undo',
         label: 'Undo'
     });
-    export const REDO = Command.toDefaultLocalizedCommand({
+    export const REDO = Command.toDefaultLocalizedCommand<[], void>({
         id: 'core.redo',
         label: 'Redo'
     });
-    export const SELECT_ALL = Command.toDefaultLocalizedCommand({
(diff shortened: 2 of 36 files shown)


packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -117,216 +117,218 @@ export namespace CommonCommands {
...
+    export const FIND = TypedCommand.toDefaultLocalizedCommand<[], void>({
         id: 'core.find',
         label: 'Find'
     });
-    export const REPLACE = Command.toDefaultLocalizedCommand({
+    // ???
+    export const REPLACE = TypedCommand.toDefaultLocalizedCommand<[], void>({
         id: 'core.replace',
         label: 'Replace'
     });
```
paul-marechal, line 160: I could not figure out if those have handlers?

vince-fugnitto: I believe we rely on monaco itself for the implementation:

https://github.com/eclipse-theia/theia/blob/1626004f4d35f462cc8f57ec33f252f70d5b9f3f/packages/monaco/src/browser/monaco-command.ts#L37-L43

paul-marechal:
Managed to find a way to do without the new `TypedCommand` type and keep the changes backward compatible.

packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
```diff
@@ -735,7 +735,7 @@ export class PluginVscodeCommandsContribution implements CommandContribution {
         commands.registerCommand({
             id: 'copyFilePath'
         }, {
-            execute: () => commands.executeCommand(CommonCommands.COPY_PATH.id)
+            execute: () => commands.executeCommand(CommonCommands.COPY_PATH.id, [])
```
paul-marechal, line 738: If I typed `COPY_PATH` properly, then this is the first potential bug caught by this change.

vince-fugnitto:
@msujew before adding additional changes what do you think of such a change to improve our command typings, and potentially pick up bugs (a couple have already been fixed thanks to the pull-request).

packages/callhierarchy/src/browser/callhierarchy-contribution.ts
```diff
@@ -28,7 +28,7 @@ export const CALL_HIERARCHY_TOGGLE_COMMAND_ID = 'callhierarchy:toggle';
 export const CALL_HIERARCHY_LABEL = nls.localizeByDefault('Call Hierarchy');
 
 export namespace CallHierarchyCommands {
-    export const OPEN = Command.toLocalizedCommand({
+    export const OPEN = Command.toLocalizedCommand<[], CallHierarchyTreeWidget>({
```
colin-grant-work, line 31: In cases like this, it might be extension-friendlier to say the args are `any[]` or `unknown[]` so that if adopters want to add the ability to pass an argument that affects (e.g.) what happens after the widget opens or something, they can register a new handler that accepts an argument without breaking the typing?

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -117,216 +117,218 @@ export namespace CommonCommands {
...
         label: 'Switch to Next Tab Group'
     }, 'theia/core/common/showNextTabGroup', VIEW_CATEGORY_KEY);
-    export const PREVIOUS_TAB_GROUP = Command.toLocalizedCommand({
+    export const PREVIOUS_TAB_GROUP = Command.toLocalizedCommand<[], void>({
         id: 'core.previousTabBar',
         category: VIEW_CATEGORY,
         label: 'Switch to Previous Tab Group'
     }, 'theia/core/common/showPreviousTabGroup', VIEW_CATEGORY_KEY);
-    export const CLOSE_TAB = Command.toLocalizedCommand({
+    export const CLOSE_TAB = Command.toLocalizedCommand<[event: Event], void>({
```
colin-grant-work, line 192: A number of labeled tuples in this PR, which we know could be breaking for downstream projects who don't keep up to date with TS. Are they really adding value?

packages/core/src/common/command.ts
```diff
@@ -301,24 +355,21 @@ export class CommandRegistry implements CommandService {
...
+    isVisible<Id extends string>(command: Id, ...args: Command.Arguments<Id>): boolean {
         return typeof this.getVisibleHandler(command, ...args) !== 'undefined';
     }
 
     /**
      * Test whether there is a toggled handler for the given command.
      */
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    isToggled(command: string, ...args: any[]): boolean {
+    isToggled<Id extends string>(command: Id, ...args: Command.Arguments<Id>): boolean {
```
colin-grant-work, line 372: I wouldn't mind seeing the returns here typed `unknown` and letting truthiness take care of itself (as is done for `Array.prototype.filter`), but requiring `boolean`s isn't wrong either.

packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
colin-grant-work: I don't believe this should be called with any arguments. `UriAwareCommandHandlers` (multi, in this case), can be called with an array of URI's or with _anything else_, and in the case of anything else, the selection service will be used to generate the URI list. In this case, calling the command with an empty array will result in nothing ever being copied to the clipboard, because an empty array passes this test:

https://github.com/eclipse-theia/theia/blob/c26f12e54ccb5c860e8b6e3d0810ddd40caad6e2/packages/core/src/common/uri-command-handler.ts#L62

so the empty array will be passed to the handler, and nothing will happen.

packages/core/src/common/uri-command-handler.ts
```diff
@@ -35,24 +35,24 @@ export interface UriCommandHandler<T extends MaybeArray<URI>> extends CommandHan
...
 /**
  * Handler for multiple URIs.
  */
-export interface MultiUriCommandHandler extends UriCommandHandler<URI[]> {
+export interface MultiUriCommandHandler extends UriCommandHandler<URI[], any> {
 
 }
 
-export class UriAwareCommandHandler<T extends MaybeArray<URI>> implements UriCommandHandler<T> {
+export class UriAwareCommandHandler<T extends MaybeArray<URI>, R = any> implements UriCommandHandler<T, R | undefined> {
```
colin-grant-work, line 49: See my comment on the `plugin-vscode-commands` change. These handlers can be called with anything. If it's a URI / URI[], then it will be used as the first arg. Otherwise a new first arg will be supplied by the command service and the supplied args will be shifted.

colin-grant-work (changes requested):
A few comments. The one re: `plugin-vscode-commands` definitely need fixing; the rest, only consideration.

packages/core/src/browser/common-frontend-contribution.ts
paul-marechal: Named tuples are already in the code base (found at least one occurrence):

https://github.com/eclipse-theia/theia/blame/a58e2f474c168d322a950fcef26bc4c95472f226/packages/core/src/common/menu/menu-adapter.ts#L22

packages/callhierarchy/src/browser/callhierarchy-contribution.ts
paul-marechal: Defining a handler with different arguments than `[]` should already be possible:

```ts
execute = (newArg1?: A, newArg2?: B, newArg3?: C) => {}
```

All new arguments must be optional as the framework will expect calling the handler with none, as per the current typings.

packages/core/src/common/command.ts
paul-marechal: `unknown` would be more ergonomic and `boolean` is more explicit. But if we document checking for truthiness then `unknown` should do the trick indeed.

packages/core/src/common/uri-command-handler.ts
paul-marechal: Honestly this `uri-command-handler` file is a pain to type... What was missing was a clear understanding of the expected behavior, thanks for clearing this out. If you see a way to properly type this then feel free to push a commit on top of this branch.

vince-fugnitto:
After today's dev meeting we decided on closing the nice-to-have pull-request for the time being due to other priorities and to keep the list of open pull-requests more manageable.

PR #11555, opened 2022-08-10
Remove `widgetPromises` from `WidgetManager`

### LINKED ISSUE #11554: `WidgetManager.tryGetWidget()` always fails for widgets with `async` factory
### Steps to Reproduce:

1. Use `tryGetWidget()` in the `OutlineViewContribution` class or call `WidgetManager.tryGetWidget()` with any of the following ID's: `'outline-view'`, `'search-view-container'`, `'scm-view-container'`, `'explorer-view-container'`, `'vsx-extensions-view-container'`, 
2. Observe that you always get `undefined`, even if the widget has obviously been successfully created.

### Additional Information

The problems are here:

https://github.com/eclipse-theia/theia/blob/1c0253adc71fa59d6ca7aef0797d3ba0c7d7d36c/packages/core/src/browser/widget-manager.ts#L221-L225

This is the only place we ever `.set()` on the `widgetPromises` `Map`, so if the factory returned a `Promise`, that map will always contain a `Promise`, although that `Promise` will always already have been settled at that point. From `tryGetWidget()` we then return what we find under the `key` in `widgetPromises`, and that will never be `instanceof Widget` if it was not so when returned from the factory.

https://github.com/eclipse-theia/theia/blob/1c0253adc71fa59d6ca7aef0797d3ba0c7d7d36c/packages/core/src/browser/widget-manager.ts#L160-L167
 
- Theia Version: 1c0253adc71fa59d6ca7aef0797d3ba0c7d7d36c

colin-grant-work:
#### What it does

Fixes #11554

#### How to test

0. CI should pass
1. Use `tryGetWidget()` in the `OutlineViewContribution` class or call `WidgetManager.tryGetWidget()` with any of the following ID's: `'outline-view'`, `'search-view-container'`, `'scm-view-container'`, `'explorer-view-container'`, `'vsx-extensions-view-container'`, 
2. If the widget has been created, you should get the widget. If it has not been created, you should get undefined.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/widget-manager.ts
+++ b/packages/core/src/browser/widget-manager.ts
@@ -114,7 +114,7 @@ export class WidgetManager {
 
     protected _cachedFactories: Map<string, WidgetFactory>;
     protected readonly widgets = new Map<string, Widget>();
-    protected readonly widgetPromises = new Map<string, MaybePromise<Widget>>();
+    // protected readonly widgetPromises = new Map<string, MaybePromise<Widget>>();
     protected readonly pendingWidgetPromises = new Map<string, MaybePromise<Widget>>();
 
     @inject(ContributionProvider) @named(WidgetFactory)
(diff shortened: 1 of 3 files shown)


packages/core/src/browser/widget-manager.ts
```diff
@@ -114,7 +114,7 @@ export class WidgetManager {
 
     protected _cachedFactories: Map<string, WidgetFactory>;
     protected readonly widgets = new Map<string, Widget>();
-    protected readonly widgetPromises = new Map<string, MaybePromise<Widget>>();
+    // protected readonly widgetPromises = new Map<string, MaybePromise<Widget>>();
```
msujew, line 117: Minor: I don't believe this serves any real purpose, right?

msujew (approved):
Looks good to me 👍 

- [x] Using the `WidgetManager#tryGetWidget` with previously troublesome IDs returns the widget
- [x] Couldn't identify any regressions with widget creation

packages/core/src/browser/widget-manager.ts
colin-grant-work: It didn't before - apart from causing errors - and it certainly doesn't now :-).

PR #11530, opened 2022-08-04
11442: Use named pipes for inter-process RPC communication

tortmayr:
#### What it does

Refactors `IPCChannel` to use named pipe communication via sockets. Previously we used an extra process pipe on top of the standard pipes (stdin,stdout,sdterr,ipc) which does not work reliably under windows. Switching to named pipe communication ensures that inter process RPC communication works reliably on all platforms.

Fixes #11442
Fixes #11390

Contributed on behalf of STMicroelectronics

#### How to test

Check that the bugs descriped in https://github.com/eclipse-theia/theia/issues/11390 and  https://github.com/eclipse-theia/theia/issues/11442#issue-1309856089  no longer occur .
**Important**: In order to launch the file watcher in a separate process the Theia application needs to be started without the `--no-cluster` flag . 

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #11561, opened 2022-08-11
revert IPC server change + fix pipe communication on Windows

paul-marechal:
#### What it does

Revert https://github.com/eclipse-theia/theia/pull/11530

Use `overlapped` instead of `pipe` when opening a process pipe for the bidirectional communication to work properly on all platforms.

#### How to test

File watching should work on every platform: Try renaming files, etc.

#### Review checklist

#### Reminder for reviewers


paul-marechal:
cc @tortmayr

.vscode/launch.json
```diff
@@ -59,6 +59,7 @@
       "args": [
         "--hostname=0.0.0.0",
         "--port=3000",
+        "--no-cluster",
```
paul-marechal, line 62: Do we still want rid of that?

packages/core/src/node/messaging/ipc-channel.ts
```diff
@@ -48,13 +49,12 @@ export class IPCChannel implements Channel {
 
     protected ipcErrorListener: (error: Error) => void = error => this.onErrorEmitter.fire(error);
 
-    constructor(pipe: Duplex, childProcess?: cp.ChildProcess) {
```
paul-marechal, line 51: Do we still want to pass the duplex to listen to as argument?

tortmayr: I don't think that is necessary.

.vscode/launch.json
tortmayr: No, that was an oversight on my side. Thanks for readding that.

tortmayr:
@paul-marechal Thanks for following up on that. Using  `overlapped` instead of `pipe` seems to fix all Windows issues. Definitely much cleaner than the named pipe setup introduced with  https://github.com/eclipse-theia/theia/pull/11530. :+1:

msujew (approved):
The changes look good to me 👍 I can confirm that the application behaves as expected on Windows.

.vscode/launch.json
paul-marechal: The revert did it, not me :)

PR #11563, opened 2022-08-11
plugin: Support Cancellation token in file events

### LINKED ISSUE #11508: [vscode] Support `CancellationToken` in file events
Support the property `token` taking a [`CancellationToken`](https://code.visualstudio.com/api/references/vscode-api#CancellationToken) in

- [ ] [`FileWillCreateEvent`](https://code.visualstudio.com/api/references/vscode-api#FileWillCreateEvent)
- [ ] [`FileWillDeleteEvent`](https://code.visualstudio.com/api/references/vscode-api#FileWillDeleteEvent)
- [ ] [`FileWillRenameEvent`](https://code.visualstudio.com/api/references/vscode-api#FileWillRenameEvent)

xoriath:
This closes #11508

Signed-off-by: Morten Engelhardt Olsen <MortenEngelhardt.Olsen@microchip.com>

#### What it does
Adds `token` to the `Will` file events, in accordance with https://code.visualstudio.com/api/references/vscode-api#FileWillCreateEvent

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/file-system-event-service-ext-impl.ts
+++ b/packages/plugin-ext/src/plugin/file-system-event-service-ext-impl.ts
@@ -209,13 +209,13 @@ export class ExtHostFileSystemEventService implements ExtHostFileSystemEventServ
     async $onWillRunFileOperation(operation: FileOperation, target: UriComponents, source: UriComponents | undefined, timeout: number, token: CancellationToken): Promise<any> {
         switch (operation) {
             case FileOperation.MOVE:
-                await this._fireWillEvent(this._onWillRenameFile, { files: [{ oldUri: URI.revive(source!), newUri: URI.revive(target) }] }, timeout, token);
+                await this._fireWillEvent(this._onWillRenameFile, { files: [{ oldUri: URI.revive(source!), newUri: URI.revive(target) }], token: token }, timeout, token);
                 break;
             case FileOperation.DELETE:
-                await this._fireWillEvent(this._onWillDeleteFile, { files: [URI.revive(target)] }, timeout, token);
+                await this._fireWillEvent(this._onWillDeleteFile, { files: [URI.revive(target)], token: token }, timeout, token);
                 break;
             case FileOperation.CREATE:
-                await this._fireWillEvent(this._onWillCreateFile, { files: [URI.revive(target)] }, timeout, token);
+                await this._fireWillEvent(this._onWillCreateFile, { files: [URI.revive(target)], token: token }, timeout, token);
                 break;
             default:
             // ignore, dont send
--- a/packages/plugin/src/theia.d.ts
+++ b/packages/plugin/src/theia.d.ts
@@ -1696,6 +1696,11 @@ export module '@theia/plugin' {
     export interface FileWillCreateEvent {
 
         /**
+         * A cancellation token.
+         */
+        readonly token: CancellationToken;
+
+        /**
          * The files that are going to be created.
          */
         readonly files: ReadonlyArray<Uri>;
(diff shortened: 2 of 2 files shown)


colin-grant-work:
@xoriath, the implementation looks correct to me, but typically for enhancements to the plugin API, the author provides a testing procedure using an existing plugin or one written for the purpose of testing the new functionality. Do you have a testing scenario in mind?

xoriath:
Oh, I do not really have a testing scenario in mind...

vince-fugnitto:
> Oh, I do not really have a testing scenario in mind...

@xoriath generally, when adding support for new VS Code APIs, we create a [test extension](https://code.visualstudio.com/api/get-started/your-first-extension) to test against to confirm that the behavior is indeed correct and the changes in fact work. For example [this](https://github.com/eclipse-theia/theia/pull/11487) pull-request includes one such extension to test against.

xoriath:
I see, let me have a stab at that as well :)

xoriath:
For my own notes: https://github.com/microsoft/vscode/commit/3d13b9aeb482c09330c8872b3bc79de5750d265c

xoriath:
Hmmmm, so I'm actually drawing a blank on how to test this... 

I went through all uses of this API on github, and I can't find any that uses the `CancellationToken` on the even. So, I tried to just write up a blocking handler and see what vscode did to it:

```ts
import * as vscode from 'vscode';

export function activate(context: vscode.ExtensionContext) {

	vscode.workspace.onWillCreateFiles(e => {
		vscode.window.showInformationMessage(`onWillCreateFiles called: ${e.files}`);

		e.token.onCancellationRequested(_ => {
			vscode.window.showInformationMessage(`onWillCreateFiles cancelled: ${e.files}`);
		});

		while(!e.token.isCancellationRequested) {
			;
		}

		vscode.window.showInformationMessage(`onWillCreateFiles isCancellationRequested: ${e.files}`);
	});
}

export function deactivate() {}
```

The token is never cancelled, so I guess the timeout part of the code path is what is actually being used. I can also not really see in where the `CancellationTokenSource` is triggered from, so not sure what the logic is for when to trigger it...

msujew:
Hey @xoriath, your code exhibits some non-obvious behavior, which prevents the cancellation token from ever changing state. You run this in the extension host:

```ts
while(!e.token.isCancellationRequested) { }
```

Since node.js uses an event loop instead of a multi-thread architecture (JS runtimes in general), there is no way that the `e.token.isCancellationRequested` can ever change, since no other code can be executed while in the `while` loop. You're basically blocking the whole process with that loop. Instead you should use async-await to create a non-blocking loop:

```ts
function wait() {
  return new Promise(resolve => {
    setTimeout(resolve, 10);
  });
}

export function activate(context: vscode.ExtensionContext) {

  vscode.workspace.onWillCreateFiles(async e => {
    vscode.window.showInformationMessage(`onWillCreateFiles called: ${e.files}`);

    e.token.onCancellationRequested(_ => {
      vscode.window.showInformationMessage(`onWillCreateFiles cancelled: ${e.files}`);
    });

    while(!e.token.isCancellationRequested) {
      await wait(); // non-blocking wait
    }

    vscode.window.showInformationMessage(`onWillCreateFiles isCancellationRequested: ${e.files}`);
  });
}
```

msujew:
@colin-grant-work @vince-fugnitto @xoriath I've tried building a test extension myself, but noticed that Theia doesn't support cancellation at all during file operations, i.e. copying a large folder isn't cancellable, even though the operations might take a few seconds on the system. This makes this change impossible to test, since the cancellation token that we do pass into the participant here is never actually cancelled:

https://github.com/eclipse-theia/theia/blob/62ce7862ef820d91dbb7ef731c58c6e5eef246a5/packages/filesystem/src/browser/file-service.ts#L1700-L1725

Issue #11556, opened 2022-08-10
Conflicting declarations in `theia.d.ts` and `theia.proposed.d.ts`


colin-grant-work:
### Steps to Reproduce:

1. In `packages/plugin/tsconfig.json` set `compilerOptions.skipLibCheck: false`.
2. Open `theia.d.ts`
3. Find `AuthenticationProviderAuthenticationSessionsChangeEvent` and observe that TS now points out that the interface is declared with different types in `theia.proposed.d.ts`

### Additional Information

Various other problems are found with `skipLibCheck: false`, and straightforward fixes result in build errors in the code.

colin-grant-work:
@martin-fleck-at, since it looks like you added the `authentication` API to `theia.d.ts`, would you be willing to clean up `theia.proposed.d.ts` where to two conflict / where the latter is out of date with VSCode's proposals?

martin-fleck-at:
@colin-grant-work Sure, I can have a look!

martin-fleck-at:
@colin-grant-work I had a look at the API and I do not think there is a good strategy to get the types proposed API and the stable API compatible. I therefore suggest to get rid of the proposed API since we only were compatible with a very specific commit of the proposed API in VS Code 1.53.2 anyway and therefore never fully compatible with 1.53.2 (see https://github.com/eclipse-theia/theia/pull/10709#issuecomment-1029910285 and the PR). Our current stable API matches the stable API from VS Code 1.54.0 and higher which I hope will be enough, what do you think?

colin-grant-work:
@martin-fleck-at, I agree. I don't believe VSCode does, and I don't believe we can or should guarantee ongoing compatibility with outdated proposal stages, so if the API has moved on, we should as well.

martin-fleck-at:
@colin-grant-work Perfect, I already implemented this in a PR here: https://github.com/eclipse-theia/theia/pull/11564 ;-)

colin-grant-work:
Closing this as complete via #11564 and #11684

PR #11564, opened 2022-08-11
Cleanup deprecated and out-dated authentication API

martin-fleck-at:
#### What it does
- Remove very out-dated backwards compatibility to proposed VS Code API to ensure we can have proper compatible types and no conflicts between the deprecated proposed API and the stable API 
  - Get rid of provider id-based API
  - Remove deprecated 'hasSession' from API
  - Deprecate login/logout (now: createSession/removeSession) in Theia

Previously we were compatible with a very specific commit of the proposed authentication API of VS Code 1.53.2. The API was moved from its proposed state to stable with 1.54.0 to which we are fully compatible even after this commit.

For a list of how the current compatibility came to be, please see https://github.com/eclipse-theia/theia/pull/10709#issuecomment-1029910285

Relates to https://github.com/eclipse-theia/theia/issues/11556

#### How to test
Check out commit https://github.com/martin-fleck-at/theia/tree/issue-1156-test which adds an authentication provider based on the stable API. We have commands that showcase the log in and log out of an account. This is basically an updated version of the testing strategy here: https://github.com/eclipse-theia/theia/pull/10709

#### Review checklist

#### Reminder for reviewers


packages/plugin-ext/src/main/browser/authentication-main.ts
```diff
@@ -273,7 +254,7 @@ interface AccountUsage {
 
 export class AuthenticationProviderImpl implements AuthenticationProvider {
     private accounts = new Map<string, string[]>(); // Map account name to session ids
-    private sessions = new Map<string, string>(); // Map account id to name
+    private sessions = new Map<string, string>();   // Map account id to name
```
colin-grant-work, line 257: Maybe as a JSDoc so that it appears on hover for anyone working here in the future?

packages/plugin-ext/src/main/browser/authentication-main.ts
```diff
@@ -347,29 +328,22 @@ export class AuthenticationProviderImpl implements AuthenticationProvider {
...
-
-    // utility method to be backwards compatible with the old AuthenticationProviderAuthenticationSessionsChangeEvent containing only the session id string
-    private getSessionId(obj: string | theia.AuthenticationSession | undefined): string | undefined {
-        if (!obj || typeof obj === 'string') {
-            return obj;
-        }
-        return obj.id;
+        return this.proxy.$removeSession(this.id, sessionId)
+            .then(() => this.messageService.info('Successfully signed out.'))
+            .then(() => { });
```
colin-grant-work, line 346:
```suggestion
            .then(() => { this.messageService.info('Successfully signed out.'); })
```

packages/plugin-ext/src/plugin/authentication-ext.ts
```diff
@@ -163,30 +119,8 @@ export class AuthenticationExtImpl implements AuthenticationExt {
         throw new Error(`Unable to find authentication provider with handle: ${providerId}`);
     }
 
-    $onDidChangeAuthenticationSessions(id: string, label: string): Promise<void> {
-        this.onDidChangeSessionsEmitter.fire({ provider: { id, label } });
+    async $onDidChangeAuthenticationSessions(provider: theia.AuthenticationProviderInformation): Promise<void> {
+        this.onDidChangeSessionsEmitter.fire({ provider });
         return Promise.resolve();
```
colin-grant-work, line 124: Can be deleted if the function is declared `async`.

colin-grant-work (approved):
Looks good to me. I don't see any way we can guarantee compatibility with proposed API stages, and the entire purpose of declaring the API a proposal only is that it should _not_ be relied upon indefinitely.

packages/plugin-ext/src/main/browser/authentication-main.ts
martin-fleck-at: Much better, thanks!

packages/plugin-ext/src/plugin/authentication-ext.ts
martin-fleck-at: True, thank you!

martin-fleck-at:
@colin-grant-work Thank you very much! I fixed the issues you mentioned and pushed an update. Should we merge this then or do you (or someone else) want to perform additional tests?

colin-grant-work:
@martin-fleck-at, I'm fine with merging, but if you'd like to request additional reviews, that is of course perfectly fine.

PR #11565, opened 2022-08-11
terminal: allow `terminal clear` without it being active

vince-fugnitto:
#### What it does

The pull-request updates the command handler for `Terminal: Clear` so it does not rely on the terminal actually being the `activeWidget`. Instead, the handler will clear the `currentTerminal`.

_**Note**_: if the terminal is hidden the output will not be cleared (I'm not sure I agree with VS Code's decision to clear).

#### How to test

1. start the application
2. spawn multiple terminals and produce output in them (ex: `ls -al`)
3. open an editor or focus another widget
4. <kbd>F1</kbd> > `Terminal: Close` should be enabled, and it should clear the last active terminal (`currentTerminal`).

#### Review checklist

#### Reminder for reviewers


alvsan09 (approved):
I have tested it according to the test instructions and it works fine :+1: 

The code looks good to me as well !!
Thanks Vince !!

PR #11568, opened 2022-08-12
eslint: disable `no-unused-expressions` for tests

vince-fugnitto:
#### What it does

The eslint rule `no-unused-expressions` does not work well with mocha and is often disabled per file. Rather than continuously disabling the rule we now turn it off for all test files.

The rule for instance did not like `expect(true).to.not.be.undefined`.

#### How to test

- CI should pass (most notably `yarn lint`).

#### Review checklist

#### Reminder for reviewers


colin-grant-work (approved):
Any change to make writing tests more convenient is good in my book.

Issue #11583, opened 2022-08-18
macos: `window.titleBarStyle` removes top level menu


vince-fugnitto:
#### Bug Description:

There is a significant UI/UX bug in **macOS** regarding the `window.titleBarStyle` (which should only apply for Linux and Windows) which when set to `custom` will completely remove the menu and make it difficult to make it return.

_native_:

![Screen Shot 2022-08-18 at 8 12 14 AM](https://user-images.githubusercontent.com/40359487/185392331-bb3b2e4c-af72-4a33-8500-a6f8bf6c31eb.png)

_custom_ (missing window controls):

![Screen Shot 2022-08-18 at 8 12 29 AM](https://user-images.githubusercontent.com/40359487/185392367-eedd80e3-052f-4144-8ebb-5b7f7f730be3.png)

#### Steps to Reproduce:

1. start the application on macOS
2. open the preferences-view
3. search for `titleBarStyle` and change from `native` to `custom` (accept the prompt)
4. on reload the top level menu will disappear

_Note_: in order to bring back the menu I had to remove my user preferences, and cached configs:
- `rm -rf ~/.theia`
- `rm -rf ~/Library/ApplicationSupport/Theia\ Electron\ Example/`

PR #11584, opened 2022-08-18
macos: fix `window.titleBarStyle` on macOS

vince-fugnitto:
#### What it does

Fixes: #11583 

The commit fixes the handling when setting `window.titleBarStyle` to `custom` on macOS (which does not apply) as it has some negative side effects.

#### How to test

1. start the application on macOS
2. open the preferences-view
3. search for `titleBarStyle` and change from `native` to `custom` 
4. the preference change will not take effect

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-main/electron-main-application.ts
+++ b/packages/core/src/electron-main/electron-main-application.ts
@@ -500,7 +500,7 @@ export class ElectronMainApplication {
         app.on('window-all-closed', this.onWindowAllClosed.bind(this));
 
         ipcMain.on(TitleBarStyleChanged, ({ sender }, titleBarStyle: string) => {
-            this.useNativeWindowFrame = titleBarStyle === 'native';
+            this.useNativeWindowFrame = isOSX || titleBarStyle === 'native';
             const browserWindow = BrowserWindow.fromId(sender.id);
             if (browserWindow) {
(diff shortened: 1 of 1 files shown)


vince-fugnitto:
We should also think of validating preference against `included` which seems to be ignored at the moment:

https://github.com/eclipse-theia/theia/blob/d27ebe7496c574ac92886b41e04a414b9fa8b69b/packages/core/src/electron-browser/window/electron-window-preferences.ts#L43-L51

msujew:
> Should we apply the extra macOS handling at the preference level (and also benefit from not having to do a reload) or change the following line:

@vince-fugnitto I believe we should also perform the proposed change in the `ElectronMainApplication` just to make absolutely sure that the window is never created without a frame. Keeping the current change would be preferable as well.

vince-fugnitto:
@msujew thank you for the feedback, I made the update and now handle it in both places 👍

msujew (approved):
I don't have a Mac at hand, but the changes looks reasonable to me 👍 

We should probably fix the `included` preference issue sometime.

msujew:
@vince-fugnitto The latest change look better to me. I believe that we should be rather safe than sorry in this regard. Checking for `isOSX` seems perfectly reasonable.

Issue #11587, opened 2022-08-19
Preference widget does not respect `included`


msujew:
### Bug Description:

Preferences which have set their `included` property set to `false` should not be visible in the preference widget. However, this property is currently ignored during rendering. This was noticed in https://github.com/eclipse-theia/theia/issues/11583.

### Steps to Reproduce:

1. Set the `included` property to `false` on one of our preferences (or see the `window.menuBarVisibility` preference on Mac).
2. Open the preference widget and assert that the preference is still visible.

### Additional Information

- Operating System: Windows
- Theia Version: 1.28.0

PR #11588, opened 2022-08-19
Respect `included` property on preferences

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/11587. Tree nodes belonging to a not included preference are no longer visible.

#### How to test

1. Start the application on Mac (or set one of the preferences `included` properties to `false`)
2. Assert that the preference is not rendered/does not appear in the search

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/preferences/preference-contribution.ts
+++ b/packages/core/src/browser/preferences/preference-contribution.ts
@@ -201,11 +201,11 @@ export class PreferenceSchemaProvider extends PreferenceProvider {
         const changes: PreferenceProviderDataChange[] = [];
         const defaultScope = PreferenceSchema.getDefaultScope(schema);
         const overridable = schema.overridable || false;
-        for (const preferenceName of Object.keys(schema.properties)) {
+        for (const [preferenceName, rawSchemaProps] of Object.entries(schema.properties)) {
             if (this.combinedSchema.properties[preferenceName]) {
                 console.error('Preference name collision detected in the schema for property: ' + preferenceName);
-            } else {
-                const schemaProps = PreferenceDataProperty.fromPreferenceSchemaProperty(schema.properties[preferenceName], defaultScope);
+            } else if (!rawSchemaProps.hasOwnProperty('included') || rawSchemaProps.included) {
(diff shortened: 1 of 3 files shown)


vince-fugnitto (commented):
@msujew I confirmed that the preference is hidden from the UI but I believe we need more.
The preference still exists as a suggestion in the `settings.json` and setting the preference still takes effect which can cause unwanted side effects.

In my example I also set `workbench.statusBar.visible` to `include: false` and it still shows up, and takes effect:

Screen Shot 2022-08-19 at 9 42 44 AM

msujew:
@vince-fugnitto Right, I'll look into that 👍

colin-grant-work (approved):
This is the correct approach - preventing the preference from ever being registered with the schema at all. In testing, this prevents undesired / inapplicable preferences from appearing in any context. 👍

vince-fugnitto:
@msujew should we still go ahead with https://github.com/eclipse-theia/theia/pull/11584 after your changes?

vince-fugnitto:
The changes unfortunately include a regression in the browser application on macOS where the top-level disappears after the application is loaded. I'll need to see why that happens:

Screen Shot 2022-08-24 at 11 09 51 AM

colin-grant-work:
```typescript
        'window.menuBarVisibility': {
            type: 'string',
            enum: ['classic', 'visible', 'hidden', 'compact'],
            markdownEnumDescriptions: [
                nls.localizeByDefault('Menu is only hidden in full screen mode.'),
                nls.localizeByDefault('Menu is always visible even in full screen mode.'),
                nls.localizeByDefault('Menu is always hidden.'),
                nls.localizeByDefault('Menu is displayed as a compact button in the sidebar. This value is ignored when `#window.titleBarStyle#` is `native`.')
            ],
            default: 'classic',
            scope: 'application',
            // eslint-disable-next-line max-len
            markdownDescription: nls.localizeByDefault("Control the visibility of the menu bar. A setting of 'toggle' means that the menu bar is hidden and a single press of the Alt key will show it. By default, the menu bar will be visible, unless the window is full screen."),
            included: !isOSX
        },
```
The `included` should check whether we're in Electron or not. If we're in the browser, the preference should be present, even on MacOS.

Issue #11460, opened 2022-07-25
Auto Save on Binary Custom Editor does not work correctly


safisa:
### Bug Description:
Hi,
When turning the auto save on, a custom editor will start doing save operations recursively after the first change.

### Steps to Reproduce:

1. Changed the auto-save preference to afterDelay
2. open a custom editor, and do a change.
3. the auto-save is triggered after the delayed time
4. I have a console.log on save operation in my custom editor, and it seems to be triggered recursively in each delay time

This could be in the custom-editors-main.ts file.
The change method keeps calling save that is calling change again:

```
private change(makeEdit: () => void): void {
        const wasDirty = this.dirty;
        makeEdit();

        if (this.dirty !== wasDirty) {
            this.onDirtyChangedEmitter.fire();
        }

        if (this.autoSave !== 'off') {
            const handle = window.setTimeout(() => {
                this.save();
                window.clearTimeout(handle);
            }, this.autoSaveDelay);
        }
    }

async save(options?: SaveOptions): Promise<void> {
        await this.saveCustomEditor(options);
    }

async saveCustomEditor(options?: SaveOptions): Promise<void> {
        if (!this.editable) {
            return;
        }

        const cancelable = new CancellationTokenSource();
        const savePromise = this.proxy.$onSave(this.resource, this.viewType, cancelable.token);
        this.ongoingSave?.cancel();
        this.ongoingSave = cancelable;

        try {
            await savePromise;

            if (this.ongoingSave === cancelable) { // Make sure we are still doing the same save
                this.change(() => {
                    this.isDirtyFromContentChange = false;
                    this.savePoint = this.currentEditIndex;
                });
            }
        } finally {
            if (this.ongoingSave === cancelable) { // Make sure we are still doing the same save
                this.ongoingSave = undefined;
            }
        }
    }

```

### Additional Information

- Operating System: win 10
- Theia Version: 1.26

PR #11599, opened 2022-08-22
custom-editors: Fix the issue: Auto Save on Binary Custom Editor does not work correctly

safisa:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/11460.
Fix an issue where auto save on custom editor does not work correctly.

#### How to test

1. enable the auto-save
2. Use the vscode cats example and open the binary file example.pawDraw
3. do a change, and make sure that the file was saved
4. make sure that the save is not called recursively, you can see it by adding a breakpoint in the custom-editors-main.ts file using dev-tools, on the method change of the MainCustomEditorModel class.

#### Review checklist

#### Reminder for reviewers

Signed-off-by: Safi Seid-Ahmad <safi@k2view.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 - [vsx-registry] improved styling of the `Extensions` view [#11494](https://github.com/eclipse-theia/theia/pull/11494)
 - [vsx-registry] removed localization for `Open VSX Registry` [#11523](https://github.com/eclipse-theia/theia/pull/11523)
 - [vsx-registry] updated extension editor rendering [#11605](https://github.com/eclipse-theia/theia/pull/11605)
+- [custom-editor] Auto Save on Binary Custom Editor does not work correctly [#11460](https://github.com/eclipse-theia/theia/issues/11460)
 
 <a name="breaking_changes_1.29.0">[Breaking Changes:](#breaking_changes_1.29.0)</a>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,6 @@
 - [vsx-registry] improved styling of the `Extensions` view [#11494](https://github.com/eclipse-theia/theia/pull/11494)
 - [vsx-registry] removed localization for `Open VSX Registry` [#11523](https://github.com/eclipse-theia/theia/pull/11523)
 - [vsx-registry] updated extension editor rendering [#11605](https://github.com/eclipse-theia/theia/pull/11605)
-- [custom-editor] Auto Save on Binary Custom Editor does not work correctly [#11460](https://github.com/eclipse-theia/theia/issues/11460)
 
 <a name="breaking_changes_1.29.0">[Breaking Changes:](#breaking_changes_1.29.0)</a>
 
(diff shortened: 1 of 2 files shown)


safisa:
Hi @vince-fugnitto  @msujew 
Can you review this PR?

vince-fugnitto:
> Hi @vince-fugnitto @msujew Can you review this PR?

@safisa do you mind sharing the `custom-editor-sample` you used, I see the following locally:

Screen Shot 2022-08-24 at 1 42 05 PM

safisa:
Hi @vince-fugnitto

First, thank you for the review. 
Actually, I am using a custom editor that I can't share here, since it requires many other dependencies. I also checked the vscode cats pawDraw example (this uses the binary custom editor), and there is an issue in the example since it uses API that is not supported in Theia (vscode.workspace.fs.isWritableFileSystem - there is already an opened issue on this).
I did some changes in the example by removing the isWritableFileSystem API, but still had an issue in loading the image in the pawDraw.js file...
Anyway, now you can use the attached updated example. When opening the example.pawDraw file, you will get a blank canvas with no image (image load issue!), and then start drawing, you will see that the file becomes dirty, and on save you will see that the file on disk is updated (check file timestamp or size), but when reopened you will start from blank again because of the image load issue.

If you enable the auto-save, and open and edit the file as above, before this PR, the save event will keep being invoked infinitely. you can check it by testing the timestamp of the file on disk, or putting a breakpoint in the pawDraw.js file inside the message event (getFileData) or in the Theia custom-editors-main.ts file, on the method change of the MainCustomEditorModel class.
After this PR fix the auto-save works as expected.
I need this fix since we are suing many custom editors in our App.

Thanks in advance.

[cat-customs-0.0.1.vsix.zip](https://github.com/eclipse-theia/theia/files/9423258/cat-customs-0.0.1.vsix.zip)

vince-fugnitto:
@safisa can you please fix the git history (seems to be issues after rebasing) and ping me when it's ready for another review?

safisa:
Hi @vince-fugnitto
Fixed the git history, now there is only the 2 files I have changed

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@
 - [vsx-registry] improved styling of the `Extensions` view [#11494](https://github.com/eclipse-theia/theia/pull/11494)
 - [vsx-registry] removed localization for `Open VSX Registry` [#11523](https://github.com/eclipse-theia/theia/pull/11523)
 - [vsx-registry] updated extension editor rendering [#11605](https://github.com/eclipse-theia/theia/pull/11605)
+- [custom-editor] Auto Save on Binary Custom Editor does not work correctly [#11460](https://github.com/eclipse-theia/theia/issues/11460)
```
vince-fugnitto, line 45: minor: `1.29.0` was released last week, we'd need to add it under a new `1.30.0` entry. 
You can remove it for now, I generally update the changelog prior to the release anyways.

vince-fugnitto (approved):
I confirmed the bug on master and that it is correctly fixed by the updates 👍

safisa:
Hi @vince-fugnitto 
This PR is already approved, is it going to be merged before the next Theia version?

PR #11603, opened 2022-08-24
[workspace] Exported `RecentWorkspacePathsData` type, fixed typo in field name

kittaakos:
#### What it does

 - Exports the `RecentWorkspacePathsData` type so that extenders can subclass the `protected` methods of the `DefaultWorkspaceServer` class with proper typing.
 - Fixes the typo in the field name.

#### How to test

No behavioral changes are expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/workspace/src/node/default-workspace-server.ts
+++ b/packages/workspace/src/node/default-workspace-server.ts
@@ -89,7 +89,7 @@ export class DefaultWorkspaceServer implements WorkspaceServer, BackendApplicati
      * Untitled workspaces that are not among the most recent N workspaces will be deleted on start. Increase this number to keep older files,
      * lower it to delete stale untitled workspaces more aggressively.
      */
-    protected untitledWorkspaceStaleThreshhold = 10;
+    protected untitledWorkspaceStaleThreshold = 10;
 
     @inject(WorkspaceCliContribution)
     protected readonly cliParams: WorkspaceCliContribution;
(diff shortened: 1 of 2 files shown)


packages/workspace/src/node/default-workspace-server.ts
```diff
@@ -89,7 +89,7 @@ export class DefaultWorkspaceServer implements WorkspaceServer, BackendApplicati
      * Untitled workspaces that are not among the most recent N workspaces will be deleted on start. Increase this number to keep older files,
      * lower it to delete stale untitled workspaces more aggressively.
      */
-    protected untitledWorkspaceStaleThreshhold = 10;
+    protected untitledWorkspaceStaleThreshold = 10;
```
colin-grant-work, line 92: The type is unfortunate, but perhaps not worth fixing if it counts as a breaking change?

colin-grant-work (commented):
Exporting the types makes sense. I'm not sure whether the typo needs to be fixed.

packages/workspace/src/node/default-workspace-server.ts
kittaakos: > The type is unfortunate, but perhaps not worth fixing if it counts as a breaking change?

 - I can drop the commit. Probably the fastest. I initially wanted to open the PR for the type `export` only, then noticed the typo.
 - I can deprecate the field and introduce a new one. Initially, I did that and wanted to avoid the _why do not you rename it?_ questions, so I changed my mind and renamed it. 
 - But I am fine also fine leaving the PR as is. There are breaking API changes between patch releases, and this change is documented. I wonder how many extenders subclass the `DefaultWorkspaceServer` to override this field 😊

Let me know. Thanks!

paul-marechal: @kittaakos like you said there's a low chance of someone using this API, so fixing the typo now should be fine.

paul-marechal (approved):
Code LGTM, breaking change is documented.

kittaakos:
Thank you for the review 🙏

Issue #10787, opened 2022-02-23
custom-editor: restore editor as part of layout


vince-fugnitto:
#### Bug Description:

The framework currently does not support restoring `custom-editors` as part of the layout. If a given custom-editor is opened and the application is restarted (reload the window for example) the editor will not be preserved.

https://user-images.githubusercontent.com/40359487/155335495-776c73c1-774e-4d67-af58-aa2d7a87e147.mp4

#### Steps to Reproduce:

1. use the [custom-editor-sample](https://github.com/eclipse-theia/theia/files/8124915/cat-customs-0.0.1.zip) extension.
2. start the application using [exampleFiles](https://github.com/lucas-koehler/vscode-extension-samples/tree/main/custom-editor-sample/exampleFiles) as a workspace.
3. open `example.cscratch` - the custom editor should open.
4. reload the application - confirm that the editor is not preserved.

vince-fugnitto:
> Or, can you help on how to fix it using Theia extension?

@safisa I haven't worked on the issue, if you are interested please don't hesitate on contributing a fix. You'll need to debug as to why the `custom-editors` are not being properly stored or restored in the layout.

Some interesting places to look:

https://github.com/eclipse-theia/theia/blob/432602946323bc1500c585de4432021b6d6a4a0a/packages/plugin-ext/src/main/browser/custom-editors/custom-editor-widget.ts#L86-L97

https://github.com/eclipse-theia/theia/blob/432602946323bc1500c585de4432021b6d6a4a0a/packages/plugin-ext/src/main/browser/custom-editors/plugin-custom-editor-registry.ts#L56-L72

The layout is stored in the `localStorage` of the browser or electron.

safisa:
To resolve this issue, we need to registerWebviewPanelSerializer in the vscode extension code. This was missing in my implementation, once I did it (in the following example) the custom editor layout was restored on reload. BUT there was a small issue on restore, getting undefined error in Theia CustomEditorWidget class, the _modelRef is undefined (we got there from a call to the setLayoutData in the application shell):

```
get saveable(): Saveable {
        return this._modelRef.object;
    }
```
Once overriding this in my Theia extension: 
```
@injectable()
export class MyCustomEditorWidget extends CustomEditorWidget {

    // override saveable() to fix the undefined _modelRef issue when reviving (restoring state) after reload (when using the registerWebviewPanelSerializer in the fabric vscode extesion)
    get saveable(): Saveable {
[...]

PR #11619, opened 2022-08-30
[custom-editor] fix restore layout issue when using the vscode registerWebviewPanelSerializer API

safisa:
#### What it does

Fix this issue: https://github.com/eclipse-theia/theia/issues/10787

#### How to test

1. use the attached updated vscode custom editor cats sample
2. open the example.cscratch file - will open the cutom editor
3. reload the page (F5)
4. the custom editor should be restored

[cat-customs-0.0.2.vsix.zip](https://github.com/eclipse-theia/theia/files/9454709/cat-customs-0.0.2.vsix.zip)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/custom-editors/custom-editor-widget.ts
+++ b/packages/plugin-ext/src/main/browser/custom-editors/custom-editor-widget.ts
@@ -45,7 +45,7 @@ export class CustomEditorWidget extends WebviewWidget implements SaveableSource,
         );
     }
     get saveable(): Saveable {
-        return this._modelRef.object;
+        return this._modelRef?.object;
     }
 
     @inject(UndoRedoService)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 - [Previous Changelogs](https://github.com/eclipse-theia/theia/tree/master/doc/changelogs/)
 
+## v1.30.0
+
+- [custom-editor] fix the restore editor as part of layout when using the vscode registerWebviewPanelSerializer API [#10787](https://github.com/eclipse-theia/theia/issues/10787)
+
+
 ## v1.29.0 - 8/25/2022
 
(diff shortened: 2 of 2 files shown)


packages/plugin-ext/src/main/browser/custom-editors/custom-editor-widget.ts
```diff
@@ -45,7 +45,7 @@ export class CustomEditorWidget extends WebviewWidget implements SaveableSource,
         );
     }
     get saveable(): Saveable {
-        return this._modelRef.object;
+        return this._modelRef?.object;
```
vince-fugnitto, line 48: @safisa given that `_modelRef` can be `undefined` (if `setRef` is not called) I think we should properly type it, do you mind making the updates?

```ts
protected _modelRef?: Reference<CustomEditorModel>;
```

paul-marechal: More importantly it doesn't fix but moves the problem downstream: `get saveable` is typed as returning `Saveable`, not `Saveable | undefined`. With this change `undefined` may be returned here, causing problems later down.

safisa: @vince-fugnitto, @paul-marechal, I agree this change may cause problems later down...

Actually, the fix I did in my Theia extension (``this._modelRef?.object``) is working for me and after reloading the page my custom editors keep working as expected, I can do changes and save. But I retested it again on the attached vscode cats example, the custom editor is restored after the reload but it is not functional, I can't do edits!!!

I will keep my override fix in my extension, but I will close this PR change.
WDYT?

paul-marechal: > I will keep my override fix in my extension, but I will close this PR change.

Up to you. I don't think we'll merge this change in its current form as it would be more interesting to understand why `_modelRef` isn't set before being accessed.

Issue #11144, opened 2022-05-12
[vscode] Support optional property `TerminalOptions#hideFromUser`


planger:
### Feature Description:
Add support for `TerminalOptions#hideFromUser` to allow specifying whether the terminal should not be surfaced to the user until `Terminal.show` is called, see https://code.visualstudio.com/api/references/vscode-api#TerminalOptions.

planger:
I'll take a look at this issue.

PR #11630, opened 2022-09-02
vscode: Support TerminalOptions.hideFromUser

planger:
#### What it does

Adds support for VS Code's `TerminalOptions.hideFromUser`. From the [VS Code API doc](https://code.visualstudio.com/api/references/vscode-api#TerminalOptions):

> When enabled the terminal will run the process as normal but not be surfaced to the user until `Terminal.show` is called. The typical usage for this is when you need to run something that may need interactivity but only want to tell the user about it when interaction is needed. Note that the terminals will still be exposed to all extensions as normal.

In particular, this change does:

* Add TerminalOptions.hideFromUser in `theia.d.ts` and pass it on from ext to main
* Add TerminalWidget.hiddenFromUser and respect it when showing terminals in the UI
* Fix `cwd` for created terminals with undefined `cwd` (default: workspace)<br>Strictly speaking, this change is unrelated but makes the terminal's `cwd` consistent with VS Code, when no `cwd` is explicitly specified

If `TerminalOptions.hideFromUser` is set to `true`, the `TerminalWidget` will be hidden from any user-facing terminal lists until it is first explicitly shown to the user (see `TerminalWidget.hidden`).

As we don't open terminal widgets by default on `TerminalService.newTerminal` anyway and as we don't have a global terminal view (in contrast to VS Code), we only need to hide hidden terminals in the `TerminalQuickOpenService` really. Or am I missing something?

Change-Id: Ic1fdf2d5515d5b2fa8feb58b72218b93eeabda4d
Fixes https://github.com/eclipse-theia/theia/issues/11144

#### How to test

* Install test VS Code extension [terminal-creation-options-0.0.6.vsix](https://github.com/planger/vscode-playground/raw/planger/issues/11144/terminal-creation-options/terminal-creation-options-0.0.6.vsix) (code is pushed to https://github.com/planger/vscode-playground/tree/planger/issues/11144/terminal-creation-options)
* Execute task "Create default terminal"
* Run Quick Palette `term` and observe how it is listed
* Execute task "Create hidden terminal"
* Run Quick Palette `term` and observe how the default terminal is listed, but not the hidden terminal
* Execute task "Show hidden terminal"
* Observe how the hidden terminal is shown to the user
* Run Quick Palette `term` and observe how the default AND the hidden terminals are both listed now

Here is how this extension behaves in VS Code:

https://user-images.githubusercontent.com/588090/188132106-f557f035-671d-4b0e-93f1-d6db65dcffa9.mp4

Here is how this extension behaves in Theia (with this PR):

https://user-images.githubusercontent.com/588090/188132359-4a72421c-f905-4671-8d61-fc6e2756201a.mp4

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/terminal/src/browser/base/terminal-widget.ts
+++ b/packages/terminal/src/browser/base/terminal-widget.ts
@@ -51,6 +51,9 @@ export abstract class TerminalWidget extends BaseWidget {
 
     abstract readonly exitStatus: TerminalExitStatus | undefined;
 
+    /** Terminal widget can be hidden from users until explicitly shown once. */
+    abstract readonly hiddenFromUser: boolean;
+
     /** The last CWD assigned to the terminal, useful when attempting getCwdURI on a task terminal fails */
     lastCwd: URI;
 
@@ -202,4 +205,9 @@ export interface TerminalWidgetOptions {
      * Terminal kind that indicates whether a terminal is created by a user or by some extension for a user
      */
     readonly kind?: 'user' | string;
+
+    /**
+     * When enabled the terminal will run the process as normal but not be surfaced to the user until `Terminal.show` is called.
+     */
+    readonly hideFromUser?: boolean;
 }
--- a/packages/plugin-ext/src/main/browser/terminal-main.ts
+++ b/packages/plugin-ext/src/main/browser/terminal-main.ts
@@ -123,12 +123,13 @@ export class TerminalServiceMainImpl implements TerminalServiceMain, Disposable 
                 title: options.name,
                 shellPath: options.shellPath,
                 shellArgs: options.shellArgs,
-                cwd: new URI(options.cwd),
+                cwd: options.cwd ? new URI(options.cwd) : undefined,
                 env: options.env,
                 strictEnv: options.strictEnv,
                 destroyTermOnClose: true,
                 useServerTitle: false,
                 attributes: options.attributes,
+                hideFromUser: options.hideFromUser,
                 isPseudoTerminal
             });
             if (options.message) {
--- a/packages/plugin/src/theia.d.ts
+++ b/packages/plugin/src/theia.d.ts
@@ -3007,6 +3007,15 @@ export module '@theia/plugin' {
         strictEnv?: boolean;
 
         /**
+         * When enabled the terminal will run the process as normal but not be surfaced to the user
+         * until `Terminal.show` is called. The typical usage for this is when you need to run
+         * something that may need interactivity but only want to tell the user about it when
+         * interaction is needed. Note that the terminals will still be exposed to all extensions
+         * as normal.
+         */
+        hideFromUser?: boolean;
+
+        /**
          * A message to write to the terminal on first launch. Note that this is not sent to the
          * process, but rather written directly to the terminal. This supports escape sequences such
          * as setting text style.
--- a/packages/terminal/src/browser/terminal-quick-open-service.ts
+++ b/packages/terminal/src/browser/terminal-quick-open-service.ts
@@ -51,8 +51,8 @@ export class TerminalQuickOpenService implements QuickAccessProvider {
     async getPicks(filter: string, token: CancellationToken): Promise<QuickPicks> {
         const items: QuickPickItem[] = [];
 
-        // Get the sorted list of currently opened terminal widgets
-        const widgets: TerminalWidget[] = this.terminalService.all
+        // Get the sorted list of currently opened terminal widgets that aren't hidden from users
+        const widgets: TerminalWidget[] = this.terminalService.all.filter(widget => !widget.hiddenFromUser)
             .sort((a: TerminalWidget, b: TerminalWidget) => this.compareItems(a, b));
 
         for (const widget of widgets) {
--- a/packages/terminal/src/browser/terminal-widget-impl.ts
+++ b/packages/terminal/src/browser/terminal-widget-impl.ts
@@ -64,6 +64,7 @@ export class TerminalWidgetImpl extends TerminalWidget implements StatefulWidget
     protected hoverMessage: HTMLDivElement;
     protected lastTouchEnd: TouchEvent | undefined;
     protected isAttachedCloseListener: boolean = false;
+    protected shown = false;
     override lastCwd = new URI();
 
     @inject(WorkspaceService) protected readonly workspaceService: WorkspaceService;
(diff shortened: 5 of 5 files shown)


packages/terminal/src/browser/base/terminal-widget.ts
```diff
@@ -51,6 +51,9 @@ export abstract class TerminalWidget extends BaseWidget {
 
     abstract readonly exitStatus: TerminalExitStatus | undefined;
 
+    /** Terminal widget can be hidden from users until explicitly shown once. */
+    abstract readonly hidden: boolean;
```
colin-grant-work, line 55: I believe all Phosphor widgets have an `isHidden` property already.

packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -321,6 +321,10 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
         return this.widgetManager.getWidgets(TERMINAL_WIDGET_FACTORY_ID) as TerminalWidget[];
     }
 
+    get allVisible(): TerminalWidget[] {
+        return this.all.filter(widget => !widget.hidden);
```
colin-grant-work, line 325: Since widgets can be programmatically hidden in other ways, it may make sense to use the `isHidden` property here, as well. In fact, it might have made sense from the start, to offer only those widgets in the quick pick :-).

colin-grant-work (commented):
> As we don't open terminal widgets by default on TerminalService.newTerminal anyway and as we don't have a global terminal view (in contrast to VS Code), we only need to hide hidden terminals in the TerminalQuickOpenService really. Or am I missing something?

Is this behavior correct? Should the default be not showing the terminal, or should we in fact be showing terminals that are opened with `newTerminal` unless they specify `hideFromUser`?

planger:
@colin-grant-work Thank you for your fast reply!

> Is this behavior correct? Should the default be not showing the terminal, or should we in fact be showing terminals that are opened with `newTerminal` unless they specify `hideFromUser` / or if `hideFromUser` is `false`, should we reveal the terminal automatically?

Yeah, I had the same thought initially. I even changed this behavior in my original draft change. However, looking at the behavior in VS Code in more detail, I think Theia actually matches the behavior of VS Code correctly:
`vscode.window.createTerminal` does not show the terminal directly in VS Code, neither if the terminal view is toggled away, nor does it bring the terminal view to the front if it is not toggled away. It doesn't even switch the current terminal. All of the above only happens if an extension explicitly invokes `terminal.show` (see also VS Code video above at the end). So I think that's ok in Theia.

The real difference between Theia and VS Code is that we don't have a singleton terminal view in Theia that lists all "visible" terminals in VS Code (i.e. terminals that have a `hideFromUser !=== true` && never been explicitly shown once). Instead each terminal is an own widget in Theia and, afaik, we can only list all terminals with `term` in the quick command palette to actually show all that have previously been created. Also the command "Toggle Terminal" doesn't bring up all terminals, but just the current terminal.

Once an extension invokes `terminal.show` on a terminal with `hideFromUser === true` in VS Code, the terminal essentially becomes a normal terminal that is visible in the terminal view as well as in the quick command palette list shown after `term` is entered.

This is also why I'm not sure whether `Widget.isHidden` is a good match to capture this state. The `TerminalWidget.hidden` property I introduced exactly just tracks the state from initial creation (if `hideFromUser === true`) to the first time it is shown. After a terminal is shown once, it becomes a normal terminal and should show up in all lists in the UI (`TerminalWidget.hidden` becomes `false` and thus behaves the same as if it would have been created with `hideFromUser = false`. Afais, `Widget.isHidden` doesn't really match that meaning. `TerminalWidget.hidden` doesn't seem to be a great name then. :thinking: 

I might be missing something though. WDYT?

planger:
I've rebased this change to resolve the conflict with #11641.

@colin-grant-work I've written my answers to your comments [above](https://github.com/eclipse-theia/theia/pull/11630#issuecomment-1235840659). It'd be great if you could let me know what you think. I'm then happy to adjust this change, if needed, depending on the outcome.

Thanks!

vince-fugnitto (commented):
I believe the changes work as expected given our differences with vscode (the single view for all terminals):

- [x] terminals are not shown unless `show` is called
- [x] terminals with `hideFromUser: false` will make terminals visible by the `term:` quick-input but not the UI
- [x] terminals with `hideFromUser: true` will make terminals invisible by the `term:` quick-input but not the UI 
- [x] terminals with `hideFromUser: true` are only visible after `show` is called

planger:
Thanks a lot @vince-fugnitto for testing the change!
@colin-grant-work Do you have further feedback?

JonasHelming:
@vince-fugnitto Is this approved from your PoV?

packages/terminal/src/browser/terminal-frontend-contribution.ts
vince-fugnitto: @planger I believe this comment is not addressed?

There are two types of **hidden** at play here which can be confusing for downstream adopters. A widget can be hidden manually through the UI, or may not have been shown yet (`hideFromUser`). For adopters they may be using `allVisible` and wondering why they are given terminals which are hidden by phosphor.

packages/terminal/src/browser/base/terminal-service.ts
```diff
@@ -33,6 +33,11 @@ export interface TerminalService {
 
     readonly all: TerminalWidget[];
 
+    /**
+     * All terminal widgets that aren't hidden (see `TerminalWidget.hidden`).
+     */
+    readonly allVisible: TerminalWidget[];
```
vince-fugnitto, line 39: @planger I'm not a huge fan of adding this new property when we already have `all` and it's used once, wdyt?

planger: Right, I haven't addressed the issue with the name `hidden` yet. I suggest to flesh out this name a bit, maybe `hiddenFromUser`?
Regarding the `allVisible` property: My thinking was that if ever someone needed to show a list of terminals in the UI, they should show the list without those `hiddenFromUser`. Thus I preemptively added a property for that to avoid forgetting about that in the future. However, I agree it might be unnecessary for now, as we only use this property once at the moment.

I'll update the change with the rename and remove the new property and do the filtering in the frontend contribution instead. Please let me know what you think.

planger:
@vince-fugnitto Thanks for the review! I pushed another update to address your feedback, see also [my comment](https://github.com/eclipse-theia/theia/pull/11630#discussion_r994356609) above.

vince-fugnitto (approved):
The latest changes look better to me, and are simpler 👍 

- [x] terminals are not visible unless explicitly shown
- [x] terminals with `hideFromUser: false` will make terminals visible by the term: quick-input but not the UI
- [x] terminals with `hideFromUser: true` will make terminals invisible by the term: quick-input but not the UI
- [x] terminals with `hideFromUser: true` are only visible after show is called

PR #11641, opened 2022-09-07
vscode: Support `TerminalOptions.strictEnv`

### LINKED ISSUE #11143: [vscode] Support optional property `TerminalOptions#strictEnv`
### Feature Description:
Add support for `TerminalOptions#strictEnv` to allow specifying whether the terminal should be exactly as provided in `TerminalOptions.env`, see https://code.visualstudio.com/api/references/vscode-api#TerminalOptions.

planger:
#### What it does

Support optional property `vscode.TerminalOptions.strictEnv` when used in `createTerminal(options: TerminalOptions)` in Theia. In particular, the doc defines the property `strictEnv` of [`TerminalOptions`](https://code.visualstudio.com/api/references/vscode-api#TerminalOptions) as follows:

> Whether the terminal process environment should be exactly as provided in `TerminalOptions.env`. When this is false (default), the environment will be based on the window's environment and also apply configured platform settings like `terminal.integrated.windows.env` on top. When this is true, the complete environment must be provided as nothing will be inherited from the process or any configuration.

Fixes #11143 

PS: This change might cause conflicts with https://github.com/eclipse-theia/theia/pull/11630, which are however very easy to fix in `theia.d.ts` once either of those PRs are merged. I didn't combine them to ensure a more focused and simple test scenario.

#### How to test

* Download test extension [terminal-creation-options-0.0.7.vsix](https://github.com/planger/vscode-playground/raw/planger/issues/11143/terminal-creation-options/terminal-creation-options-0.0.7.vsix) and copy it to the `plugins` folder
* Start Theia with an environment variable: e.g. `TESTVAR=TESTVALUE yarn browser start`
* This test extension registers two commands (see [code of this extension](https://github.com/planger/vscode-playground/tree/planger/issues/11143/terminal-creation-options))
  * *Create default terminal*:<br>Creates a terminal without using `strictEnv` but defining an environment variable `hello: world`.
  * *Create strict env terminal*:<br>Creates a terminal with `strictEnv` and an environment variable `hello: world`.
* Run both of these commands in Theia
* Show both terminals: e.g., by running the command "View: Show all opened terminals" in the command palette and select one terminal and run command again and select the other terminal
* Run `printenv` in both of the terminals
* Compare the printed environment and check that the one with strict environment shows no environment variables of the Theia process (e.g. `TESTVAR*` added above but also others, etc.).

I also compared the resulting environments in Ubuntu between VS Code and Theia and the environments are equivalent for the `strictEnv` case.
We should also verify with Windows that no transformations are applied to the specified environment variables, as indicated by the documentation in VS Code.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/terminal-main.ts
+++ b/packages/plugin-ext/src/main/browser/terminal-main.ts
@@ -125,6 +125,7 @@ export class TerminalServiceMainImpl implements TerminalServiceMain, Disposable 
                 shellArgs: options.shellArgs,
                 cwd: new URI(options.cwd),
                 env: options.env,
+                strictEnv: options.strictEnv,
                 destroyTermOnClose: true,
                 useServerTitle: false,
                 attributes: options.attributes,
--- a/packages/plugin/src/theia.d.ts
+++ b/packages/plugin/src/theia.d.ts
@@ -2988,6 +2988,15 @@ export module '@theia/plugin' {
         env?: { [key: string]: string | null };
 
         /**
+         * Whether the terminal process environment should be exactly as provided in
+         * `TerminalOptions.env`. When this is false (default), the environment will be based on the
+         * window's environment and also apply configured platform settings like
+         * `terminal.integrated.windows.env` on top. When this is true, the complete environment
+         * must be provided as nothing will be inherited from the process or any configuration.
+         */
+        strictEnv?: boolean;
+
+        /**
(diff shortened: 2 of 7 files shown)


msujew (approved):
The change look good to me. I can confirm that the `strictEnv` option works as expected:

- [x] Passing `false/undefined` keeps the full env variables from the Theia process.
- [x] Passing `true` only retains the explicitly passed env variables.

PR #11658, opened 2022-09-12
Support cancellation on plugin file events

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/11508
Closes https://github.com/eclipse-theia/theia/pull/11563

Supersedes the PR above by adapting the approach taken by vscode. vscode [always passes the token into the async event](https://github.com/microsoft/vscode/blob/da34e253ca55a22c0272708dcfe622a6f8fb33fc/src/vs/base/common/event.ts#L880-L883), so we don't have to do that ourselves for each event.

Additionally enables to actually cancel the file events using a notification similar to vscode.

Also updates a few missing localizable strings in the `file-service.ts` file.

#### How to test
1. Install the test extension [cancellation-token-0.0.1.vsix](https://github.com/eclipse-theia/theia/files/9548623/cancellation-token-0.0.1.vsix.zip).
2. Create a new file in the explorer.
3. A notification should appear that a file is in progress of being created (both from Theia and from the extension)
4. Cancelling the operation explicitly or by waiting (5 seconds by default) will cancel the token and another notification will appear that the operation has been cancelled. The file will then be successfully created.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/event.ts
+++ b/packages/core/src/common/event.ts
@@ -336,8 +336,7 @@ export namespace WaitUntilEvent {
         token = CancellationToken.None
     ): Promise<void> {
         const waitables: Promise<void>[] = [];
-        const asyncEvent = {
-            ...event,
+        const asyncEvent = Object.assign(event, {
             token,
             waitUntil: (thenable: Promise<any>) => {
                 if (Object.isFrozen(waitables)) {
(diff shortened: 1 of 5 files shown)


vince-fugnitto (approved):
I confirm that the changes look good, and work as expected 👍

PR #8911, opened 2020-12-29
Apply FileDecoration provider API

vinokurig:
#### What it does

Align the FileDecoration vscode Plugin API with the latest vscode version. This API was changed and it is used in the vscode git plugin, so to be able to use the latest vscode git plugin it is needed to apply the changes in the Theia's FileDecoration plugin API.

related CQ: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22921
#### How to test

1. Remove `"@theia/git": "^1.9.0"` from `examples/browser/package.json` file.
2. Download [vscode-git-1.52.1](https://open-vsx.org/api/vscode/git/1.52.1/file/vscode.git-1.52.1.vsix) and copy it to the `plugins` folder.
3. Start Theia and  open a project cloned from GitHub.
4. Make some changes in the project and check them in the SCM panel:
screenshot-localhost_3000-2020 12 29-17_06_55

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #11674, opened 2022-09-16
git: re-add support for git decoration preferences

vince-fugnitto:
#### What it does

The pull-request re-adds support for `git.decorations.enabled` and `git.decorations.colors` which were previously not referenced and supported after a change in https://github.com/eclipse-theia/theia/pull/8911 (v1.14.0).

https://user-images.githubusercontent.com/40359487/190656031-c124f559-ca5d-4e43-b052-32005dc0a11f.mov

#### How to test

1. start the application with a workspace that is under version control
2. perform some changes in the workspace, (create new files, modifications, deletions)
3. open the preferences-view
4. search for `git decorations`
5. confirm that toggling `git.decorations.enabled` hides/shows decorations in the navigator for the resources
6. confirm that toggling `git.decorations.colors` hides/displays colors for the decorations in the navigator for the resources
7. confirm the preferences are respected on startup

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/git/src/browser/git-decoration-provider.ts
+++ b/packages/git/src/browser/git-decoration-provider.ts
@@ -14,46 +14,66 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { inject, injectable } from '@theia/core/shared/inversify';
+import { inject, injectable, postConstruct } from '@theia/core/shared/inversify';
 import { GitFileChange, GitFileStatus, GitStatusChangeEvent } from '../common';
 import { CancellationToken, Emitter, Event } from '@theia/core/lib/common';
 import { Decoration, DecorationsProvider } from '@theia/core/lib/browser/decorations-service';
 import { GitRepositoryTracker } from './git-repository-tracker';
 import URI from '@theia/core/lib/common/uri';
+import { GitConfiguration, GitPreferences } from './git-preferences';
+import { PreferenceChangeEvent } from '@theia/core/lib/browser';
 
 @injectable()
 export class GitDecorationProvider implements DecorationsProvider {
 
+    @inject(GitPreferences) protected readonly preferences: GitPreferences;
+    @inject(GitRepositoryTracker) protected readonly gitRepositoryTracker: GitRepositoryTracker;
+
+    protected decorationsEnabled: boolean;
+    protected colorsEnabled: boolean;
+
+    protected decorations = new Map<string, Decoration>();
+    protected uris: Set<string> = new Set<string>();
+
+    /**
+     * Cached change event for re-rendering decorations.
+     */
+    protected changeEvent: GitStatusChangeEvent | undefined;
+
     private readonly onDidChangeDecorationsEmitter = new Emitter<URI[]>();
     readonly onDidChange: Event<URI[]> = this.onDidChangeDecorationsEmitter.event;
 
-    private decorations = new Map<string, Decoration>();
-
-    constructor(@inject(GitRepositoryTracker) protected readonly gitRepositoryTracker: GitRepositoryTracker) {
-        this.gitRepositoryTracker.onGitEvent((event: GitStatusChangeEvent | undefined) => {
-            this.onGitEvent(event);
-        });
+    @postConstruct()
+    protected init(): void {
+        this.decorationsEnabled = this.preferences['git.decorations.enabled'];
+        this.colorsEnabled = this.preferences['git.decorations.colors'];
+        this.gitRepositoryTracker.onGitEvent((event: GitStatusChangeEvent | undefined) => this.handleGitEvent(event));
+        this.preferences.onPreferenceChanged(event => this.handlePreferenceChange(event));
     }
 
-    private async onGitEvent(event: GitStatusChangeEvent | undefined): Promise<void> {
-        if (!event) {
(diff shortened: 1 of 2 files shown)


packages/git/src/browser/git-decoration-provider.ts
```diff
@@ -14,54 +14,103 @@
...
+
     private readonly onDidChangeDecorationsEmitter = new Emitter<URI[]>();
     readonly onDidChange: Event<URI[]> = this.onDidChangeDecorationsEmitter.event;
 
-    private decorations = new Map<string, Decoration>();
+    @postConstruct()
+    protected init(): void {
+        this.decorationsEnabled = this.preferences['git.decorations.enabled'];
+        this.colorsEnabled = this.preferences['git.decorations.colors'];
+        this.gitRepositoryTracker.onGitEvent((event: GitStatusChangeEvent | undefined) => this.handleGitEvent(event));
```
msujew, line 50: Minor: Do we need to be this explicit?
```suggestion
        this.gitRepositoryTracker.onGitEvent(event => this.handleGitEvent(event));
```

packages/git/src/browser/git-decoration-provider.ts
```diff
@@ -14,54 +14,103 @@
...
 export class GitDecorationProvider implements DecorationsProvider {
 
+    @inject(GitPreferences) protected readonly preferences: GitPreferences;
+    @inject(GitRepositoryTracker) protected readonly gitRepositoryTracker: GitRepositoryTracker;
+
+    protected decorationsEnabled: boolean;
+    protected colorsEnabled: boolean;
+
+    protected decorations = new Map<string, Decoration>();
+    protected uris: Set<string> = new Set<string>();
```
msujew, line 36: Minor:
```suggestion
    protected uris = new Set<string>();
```

packages/git/src/browser/git-decoration-provider.ts
```diff
@@ -14,54 +14,103 @@
...
+
+    /**
+     * Cached change event for re-rendering decorations.
+     */
+    protected changeEvent: GitStatusChangeEvent | undefined;
```
msujew, line 41: Suggestion: I'm not sure whether we need to store the `changeEvent` if we already store the `uris` and `decorations`. We can simply recompute `uris` and `decorations` when a `GitStatusChangeEvent` occurs and reuse these results.

packages/git/src/browser/git-decoration-provider.ts
```diff
@@ -14,54 +14,103 @@
...
+                }
+            }
+            if (preferenceName === 'git.decorations.colors') {
+                const colorsEnabled = !!newValue;
+                if (this.colorsEnabled !== colorsEnabled) {
+                    this.colorsEnabled = colorsEnabled;
+                }
+            }
+            this.updateDecorations();
+        }
```
msujew, line 105: Suggestion: Since all these comparisons can quickly become difficult to maintain we should maybe opt for something like this. It effectively cuts the amount of string comparisons in half.
```suggestion
        let updateDecorations = false;
        if (preferenceName === 'git.decorations.enabled') {
            updateDecorations = true;
            const decorationsEnabled = !!newValue;
            if (this.decorationsEnabled !== decorationsEnabled) {
                this.decorationsEnabled = decorationsEnabled;
            }
        }
        if (preferenceName === 'git.decorations.colors') {
            updateDecorations = true;
            const colorsEnabled = !!newValue;
            if (this.colorsEnabled !== colorsEnabled) {
                this.colorsEnabled = colorsEnabled;
            }
        }
        if (updateDecorations) {
            this.updateDecorations();
        }
```

packages/git/src/browser/git-decoration-provider.ts
```diff
@@ -14,54 +14,103 @@
...
                 letter: GitFileStatus.toAbbreviation(change.status, change.staged)
             });
         });
     }
 
     provideDecorations(uri: URI, token: CancellationToken): Decoration | Promise<Decoration | undefined> | undefined {
-        return this.decorations.get(uri.toString());
+        if (this.decorationsEnabled) {
+            return this.decorations.get(uri.toString());
+        }
```
msujew, line 86: Minor: We should be more explicit about what we return when decorations are disabled:
```suggestion
        if (this.decorationsEnabled) {
            return this.decorations.get(uri.toString());
        }
        return undefined;
```

msujew (commented):
I can confirm that the changes work as expected. I have some comments though, see below.

packages/git/src/browser/git-decoration-provider.ts
msujew: I've got to admit, I didn't think about that. I've provided a patch that shows a working solution: [decorations-patch.txt](https://github.com/eclipse-theia/theia/files/9585579/decorations.txt)

I'm not sure whether that's better though, since it introduces some complexity into `provideDecorations` (but removes some complexity from `collectDecorationData`).

vince-fugnitto:
@msujew thank you for your help, I like it better than having to cache the event itself.

msujew (approved):
The changes look good to me 👍

- [x] Decorations are applied/colored correctly when both preferences are enabled
- [x] Decorations can be completely disabled
- [x] Decoration colors can be disabled independently of that
- [x] Changes these preferences is applied correctly and instantly

alvsan09 (approved):
I have verified the changes as per instructions and with  a different theme and it works as expected ! :+1: 

The code looks good to me as well !!

Issue #11672, opened 2022-09-16
Settings for color theme and icon theme in preferences should probably be a drop-down box


inlann:
### Feature Description:

Hello:

Settings for color theme and icon theme in preferences is a text input field,

![](https://community.theia-ide.org/uploads/default/original/2X/1/13b82689b5225b810f45d91a06c0ba02cf61b4eb.jpeg)

which would be better if it were a drop-down box like VSCode:

![](https://community.theia-ide.org/uploads/default/original/2X/7/7c228ea037c6f6e98bb700ba4c2e47a10440b238.png)

Issue #11676, opened 2022-09-16
Cannot uninstall color theme plugins


msujew:
### Bug Description:

I didn't dive too deep into this issue, but noticed it while working on https://github.com/eclipse-theia/theia/issues/11672. When uninstalling a color theme, the frontend contributions from that plugin aren't disposed at all. When reloading the application, the monaco theming service still thinks these color themes exist, and stores them in the indexed DB. After reloading the plugin has been disposed, but since the indexed DB still contains the themes, the theming service loads them from the indexed DB, making it effectively impossible to uninstall color themes.

### Steps to Reproduce:

1. Install a color theme
2. Uninstall it
3. Confirm that the color themes are still available for selection, even after reloading the application

### Additional Information

- Operating System: Windows
- Theia Version: 1.29.0

vince-fugnitto:
I meant to open this a while ago but forgot, thanks!

PR #11678, opened 2022-09-16
Improve theme preference selection

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/11672

Changes the preferences for color theme and icon theme to use an enum. In order to facilitate label displaying and changing the associated preference schemas during runtime some more changes had to be performed:

1. Support for `enumItemLabels` in `PreferenceSelectInput`
2. Add `updateSchemaProperty` method to `PreferenceSchemaProvider` that triggers the associated change event.
3. Creating and storing keys for each preference while rendering. Allows us to identify which preferences have changed. The previous method only could tell whether a preference was removed/added.

#### How to test

1. Open the preferences and search for the file/color theme preference
2. It should be a select component and display all currently displayed themes
3. Changing to another file/color theme applies the change as expected
4. Installing a new file/color theme using plugins updates the preference widget and allows to select the newly installed theme

Note that uninstalling color themes doesn't work correctly due to https://github.com/eclipse-theia/theia/issues/11676.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/core-preferences.ts
+++ b/packages/core/src/browser/core-preferences.ts
@@ -152,11 +152,15 @@ export const corePreferenceSchema: PreferenceSchema = {
         },
         'workbench.colorTheme': {
             type: 'string',
+            enum: ['dark', 'light', 'hc-theia'],
+            enumItemLabels: ['Dark (Theia)', 'Light (Theia)', 'High Contrast (Theia)'],
             default: DefaultTheme.defaultForOSTheme(FrontendApplicationConfigProvider.get().defaultTheme),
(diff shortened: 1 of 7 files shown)


vince-fugnitto (approved):
LGTM 👍 

- [x] both the color and icon theme use a select component for theming similarly to vscode
- [x] changing color and icon themes work as expected
- [x] the list is correctly sorted, and the `default` value is identified
- [x] resetting the color and icon themes work as expected (the list is correctly updated)
- [x] installing new themes are listed in the dropdown

PR #11684, opened 2022-09-19
vscode: introduce `theia-extra.d.ts`

vince-fugnitto:
#### What it does

The pull-request introduces the `theia-extra.d.ts` file which contains plugin APIs which are supported by the framework thanks to our flexibility but not necessarily available in the VS Code API. The APIs are also different than `proposed` so a new file was needed to reflect additional APIs Theia supports but VS Code does not.

One such case is `WebviewPanel.reveal` (which is currently marked as partially supported).

#### How to test

There is no change in functionality:
- CI should pass.
- Confirm that `reveal` in `theia.d.ts` matches VS Code.
- The comparator should mark the `WebviewPanel.reveal` as supported.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin/src/theia-extra.d.ts
+++ b/packages/plugin/src/theia-extra.d.ts
@@ -1,0 +1,36 @@
+// *****************************************************************************
+// Copyright (C) 2022 Ericsson and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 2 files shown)


msujew (approved):
I can confirm that the change marks the `WebviewPanel.reveal` method as fully supported. I like the idea of adding the `theia-extra.d.ts` file.

colin-grant-work (approved):
I like the idea 👍

PR #11693, opened 2022-09-22
11685: Restore cancellation token behavior

### LINKED ISSUE #11685: New RPC changes cancellation token behavior
### Bug Description:

Prior to the changes to the RPC architecture, cancelling an RPC using a cancellation token would not automatically result in a rejection on the calling side. Instead, the cancellation would be passed to the receiving side, and the callee could handle the cancellation itself. In the new RPC, any cancellation automatically cancels the request on the sending side.

https://github.com/eclipse-theia/theia/blob/99edac2e68cc8886c30865842b895cd93dc331b3/packages/core/src/common/message-rpc/rpc-protocol.ts#L135-L146

Arguments can be made for either approach (if the frontend cancels, it probably shouldn't expect the response; on the other hand, perhaps it would be happy to accept a partial result on cancellation), but it is a definite breakage.

### Steps to Reproduce:

1. Implement a frontend service that calls a backend service, provide a cancellation token, and cancel the action immediately.
2. Implement logic on the backend to handle cancelation and return some kind of default / partial result.
3. Before 1.27 (tested on 1.24), the partial return from the backend would be received on the frontend; since 1.27, the call will be rejected.

### Additional Information

- Theia Version: 1.28

tortmayr:
#### What it does

Ensure that cancelling an RPC call does not result in an automatic rejection on the calling side. Instead the token is passed to the receiving side and handed there.  This restores the cancellation strategy that was used with the old json-rpc architecture (prior to 1.28)

Fixes #11685
#### How to test

- Implement a frontend service that calls a backend service, provide a cancellation token, and cancel the action immediately.
- Implement logic on the backend to handle cancelation and return some kind of default / partial result.
- The (partial) return from the backend should be received on the frontend. The call should not get rejected.

The simple service that I used for testing is available here: 
https://github.com/eclipsesource/theia/tree/cancelation-token-test

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/message-rpc/rpc-protocol.ts
+++ b/packages/core/src/common/message-rpc/rpc-protocol.ts
@@ -132,23 +132,23 @@ export class RpcProtocol {
         // The last element of the request args might be a cancellation token. As these tokens are not serializable we have to remove it from the
         // args array and the `CANCELLATION_TOKEN_KEY` string instead.
         const cancellationToken: CancellationToken | undefined = args.length && CancellationToken.is(args[args.length - 1]) ? args.pop() : undefined;
-        if (cancellationToken && cancellationToken.isCancellationRequested) {
-            return Promise.reject(this.cancelError());
-        }
 
         if (cancellationToken) {
             args.push(RpcProtocol.CANCELLATION_TOKEN_KEY);
-            cancellationToken.onCancellationRequested(() => {
-                this.sendCancel(id);
-                this.pendingRequests.get(id)?.reject(this.cancelError());
-            }
-            );
         }
+
         this.pendingRequests.set(id, reply);
 
         const output = this.channel.getWriteBuffer();
         this.encoder.request(output, id, method, args);
         output.commit();
+
+        if (cancellationToken?.isCancellationRequested) {
+            this.sendCancel(id);
+        } else {
+            cancellationToken?.onCancellationRequested(() => this.sendCancel(id));
+        }
+
         return reply.promise;
     }
 
@@ -162,12 +162,6 @@ export class RpcProtocol {
         const output = this.channel.getWriteBuffer();
         this.encoder.cancel(output, requestId);
         output.commit();
-    }
-
-    cancelError(): Error {
-        const error = new Error('"Request has already been canceled by the sender"');
-        error.name = 'Cancel';
-        return error;
     }
 
     protected handleCancel(id: number): void {


packages/core/src/common/message-rpc/rpc-protocol.ts
```diff
@@ -132,15 +132,11 @@ export class RpcProtocol {
         // The last element of the request args might be a cancellation token. As these tokens are not serializable we have to remove it from the
         // args array and the `CANCELLATION_TOKEN_KEY` string instead.
         const cancellationToken: CancellationToken | undefined = args.length && CancellationToken.is(args[args.length - 1]) ? args.pop() : undefined;
-        if (cancellationToken && cancellationToken.isCancellationRequested) {
-            return Promise.reject(this.cancelError());
-        }
```
colin-grant-work, line 137: From the comment above about subbing in a magic string for the cancellation token itself, it looks like we don't pass the state of token - whether it's already been cancelled or not - to the receiving end. If that's the case, then by removing this block, we remove handling the case in which the token has already been cancelled when we get here, since I don't believe that it will fire the `onCancellationRequested` event again. In that case, should we send the cancellation notice immediately after sending the request itself?

tortmayr: Good point. Yes, sending the cancellation notice directly after sending the request sounds reasonable to me.

tortmayr: @colin-grant-work After investigating and testing this it turns out that both main `CancellationToken` implementations (Theia and vscode-languageserver) already have a special shortcut event in place that will trigger newly registered cancellation listeners on already canceled tokens immediately. So the cancellation notice for immediately canceled tokens is already passed correctly to the receiving side.

However, just to be on the safe side (and since in theory also custom cancellation token implementations could be used)    I still added the discussed behavior. Directly after sending the request we check whether the token is already canceled.
If so, send the cancellation notice, otherwise listen on the `onCancellationRequested` event.

colin-grant-work (approved):
This restores the old behavior: the backend receives the cancellation as expected, and the frontend receives the response generated by the backend in case of cancellation. 👍 Thanks!

colin-grant-work:
@JonasHelming, I have a high level of confidence in this solution, and I believe there's little chance of regression. But I wouldn't insist on its inclusion in the community release if there are objections or concerns.

JonasHelming:
OK, I talked to Tobi and agree. We do not have a formal process, but @paul-marechal or @tsmaeder would you mind to either give your OK, too or veto?

tsmaeder:
The solution looks good to me, as for whether to include this in the community release: does the behaviour we fix in this PR cause problems "in the wild" and how bad are those problems? https://github.com/eclipse-theia/theia/issues/11685 doesn't really have info about the real world impact. If it has significant impact, let's include it, if it doesn't, it's an automatic "no" from me ;-)

JonasHelming:
Yes, that is a valid point, @colin-grant-work could you comment on this? I think I remember you mentioned it to be a "corner case"?

colin-grant-work:
The impact on us was non-trivial - a whole widget ended up hanging in an indeterminate state because we hadn't expected the Promise generated by the backend call to ever get rejected - either it would succeed or be canceled with partial return. But it took a number of steps to get there: we had to know we wanted to time out at some point and issue the cancelation but also expect cancellation to work in a fairly timely manner and without rejection. Those assumptions had been correct before, but I'm not sure how many teams will have gotten themselves in the same situation.

colin-grant-work:
@JonasHelming, since we uplifted to 1.28 and encountered this problem, we've adjusted our code to handle the cancellation rejection on the frontend, so it isn't urgent for _us_ that it make it into the community release, but if there are products waiting to uplift until they can uplift to the community release, then it might still be worth it to get this fix in so that they don't run into the same problem when they do.

JonasHelming:
Sorry, I rephrase my question: Will you uplift your project to the community release soon? If that is the case, you would most likely identify any potential regressions we find with this fix before we announce the release in two weeks from now. Even if the risk is low, this would be another safety net.

JonasHelming:
In any case, my vote is a "yes" on this one based on what you described. @paul-marechal ?

colin-grant-work:
@JonasHelming, I think we will start an uplift to 1.30, yes, so we should catch any regressions in practical code.

JonasHelming:
OK, we will include this.

tsmaeder:
@colin-grant-work given the real-world impact, I'm fine with the inclusion, as well.

JonasHelming:
This is now part of the release (1.29.1) Thanks @colin-grant-work for bringing this up!

PR #11729, opened 2022-09-30
#10028 Support InlineValue feature

### LINKED ISSUE #10028: Support `InlineValue` feature
### Feature Description:
Support plugins contributing `InlineValue` objects to the debugger editor, see https://code.visualstudio.com/api/references/vscode-api#InlineValuesProvider

`debug-editor-model.ts` already has support for inline value decorations, but we would have to call the plugin API to add to or replace the one's we're currently showing.

Depends partially on #10027

ndoschek:
#### What it does

- Implement support for plugins providing inline values

Signed-off-by: Nina Doschek <ndoschek@eclipsesource.com>

Resolves #10028

#### How to test

1. Download the VSCode extension [`Mock Debug`](https://marketplace.visualstudio.com/items?itemName=ms-vscode.mock-debug) (current latest version `0.51.0`) and install via Theia's `Extensions` view (Menu `View > Extensions`; Toolbar `... > Install from VSIX...`).
2. [Use the extension](https://github.com/microsoft/vscode-mock-debug#using-mock-debug) to debug a `readme.md` file (e.g. the proposed [`readme.md`](https://github.com/microsoft/vscode-mock-debug/blob/main/sampleWorkspace/readme.md) from the extension repo)
3. The inline values for the variables in the `Variables` section will show their values as inline decorations.

#### Review checklist

#### Reminder for reviewers


packages/debug/src/browser/editor/debug-inline-value-decorator.ts
```diff
@@ -240,3 +374,13 @@ export class DebugInlineValueDecorator implements FrontendApplicationContributio
...
+/**
+ * @returns New array with all falsy values removed. The original array IS NOT modified.
+ */
+export function coalesce<T>(array: ReadonlyArray<T | undefined | null>): T[] {
+    return <T[]>array.filter(e => !!e);
+}
+
+export function flatten<T>(arr: T[][]): T[] {
+    return (<T[]>[]).concat(...arr);
+}
```
colin-grant-work, line 386: `coalesce` is already available in the types utility file:

https://github.com/eclipse-theia/theia/blob/d3f6359934d025781810c467e2f3b42b55f0ff0d/packages/core/src/common/types.ts#L166-L171

and `flatten` should also be added there. Or, since `debug` depends on `monaco-editor-core`, they could be imported from there.

colin-grant-work (commented):
Just reading over the code which looks reasonable, for the most part. Regarding testing, could you clarify which versions of the plugin and the readme file the PR should be tested against? The first step says

> Install the VSCode extension [Mock Debug](https://marketplace.visualstudio.com/items?itemName=ms-vscode.mock-debug) via the Open VSX registry.

with a link that points not to Open VSX but to the VSCode store. The versions available in the two locations are quite different, and the versions of the model readme file are also quite different.

packages/debug/src/browser/editor/debug-inline-value-decorator.ts
ndoschek: Thanks for catching that, I forgot to remove those two functions again, as they are not needed with the current state anymore.

colin-grant-work (approved):
The code looks good and the Mock Debug example behaves as it should. 👍

PR #11752, opened 2022-10-11
Improved support for VsCode frontend extensions

federicobozzini:
#### What it does

Looking at the logs it looks like web vs-code extensions are still classified as backend extension. Shouldn't they be treated the same as [theia plugins where there is a clear discrimination between backend and frontend plugins](https://github.com/ARMmbed/theia/blob/master/packages/plugin-ext/src/main/node/handlers/plugin-theia-directory-handler.ts#L62)?

#### How to test

Installing a web vs-code extension should show a difference in the logs. There are more differences in how backend and frontend plugins are treated but this PR is not meant to address any specific issue, but more to make vs-code extension support future proof.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext-vscode/src/node/plugin-vscode-directory-handler.ts
+++ b/packages/plugin-ext-vscode/src/node/plugin-vscode-directory-handler.ts
@@ -49,7 +49,15 @@ export class PluginVsCodeDirectoryHandler implements PluginDeployerDirectoryHand
 
     async handle(context: PluginDeployerDirectoryHandlerContext): Promise<void> {
         await this.copyDirectory(context);
-        context.pluginEntry().accept(PluginDeployerEntryType.BACKEND);
+        const types: PluginDeployerEntryType[] = [];
+        const packageJson: PluginPackage = context.pluginEntry().getValue('package.json');
+        if (packageJson.main) {
+            types.push(PluginDeployerEntryType.BACKEND);
+        }
+        if (packageJson.browser) {
+            types.push(PluginDeployerEntryType.FRONTEND);
+        }
+        context.pluginEntry().accept(...types);
     }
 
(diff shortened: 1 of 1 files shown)


packages/plugin-ext-vscode/src/node/plugin-vscode-directory-handler.ts
```diff
@@ -49,7 +49,17 @@ export class PluginVsCodeDirectoryHandler implements PluginDeployerDirectoryHand
...
+        const packageJson: PluginPackage = context.pluginEntry().getValue('package.json');
+        if (packageJson.main) {
+            types.push(PluginDeployerEntryType.BACKEND);
+        }
+        if (packageJson.browser) {
+            types.push(PluginDeployerEntryType.FRONTEND);
+        }
+
+        context.pluginEntry().accept(...types);
+        return Promise.resolve();
```
paul-marechal, line 62: Unnecessary white space and return value:

```suggestion
        context.pluginEntry().accept(...types);
```

paul-marechal (commented):
LGTM on principle

packages/plugin-ext-vscode/src/node/plugin-vscode-directory-handler.ts
federicobozzini: Changed.

colin-grant-work (approved):
This is correct and ensures that plugins with a browser component are properly deployed.

vince-fugnitto:
@federicobozzini I believe this pull-request should be reverted or fixed, it breaks the loading of most plugins (ex: out of 76 builtins only 23 are actually loaded).

federicobozzini:
@vince-fugnitto I didn't notice any error on my local Theia setup but I'll quickly investigate this. If I will not find any good fix I'll open a PR to revert this change.

PR #11753, opened 2022-10-11
Plugin might contain @ in its name

federicobozzini:
#### What it does

This PR fixes a problem coming from plugins containing the `@` in their name. If a plugin (or a vs-code extension) contains `@` in its name (eg: `@acme/hello-world`) at the moment the name is processed incorrectly. This is especially relevant for private plugins where this naming convention is frequently used.

Using `lastIndexOf` guarantees we stop reading the name where the version (eg: `@2.4.5`) begins.

#### How to test

Two plugins need to be installed, using the same pattern (eg: `@acme/hello-world` and `@acme/hello-theia`). One of the two will be removed because of a incorrect processing of their name.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/common/plugin-identifiers.ts
+++ b/packages/plugin-ext/src/common/plugin-identifiers.ts
@@ -14,6 +14,10 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 // *****************************************************************************
 
+// From https://semver.org/#does-semver-have-a-size-limit-on-the-version-string
+// eslint-disable-next-line max-len
+const semverRegex = /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/;
+
 export namespace PluginIdentifiers {
     export interface Components {
         publisher?: string;
@@ -56,7 +60,7 @@ export namespace PluginIdentifiers {
      * @returns a string in the format `<publisher>.<name>`.
      */
     export function unversionedFromVersioned(id: VersionedId): UnversionedId {
-        const endOfId = id.indexOf('@');
+        const endOfId = id.lastIndexOf('@');
         return id.slice(0, endOfId) as UnversionedId;
     }
     /**
@@ -64,21 +68,34 @@ export namespace PluginIdentifiers {
      */
     export function identifiersFromVersionedId(probablyId: string): Components | undefined {
         const endOfPublisher = probablyId.indexOf('.');
-        const endOfName = probablyId.indexOf('@', endOfPublisher);
+        const endOfName = probablyId.lastIndexOf('@');
         if (endOfPublisher === -1 || endOfName === -1) {
             return undefined;
         }
-        return { publisher: probablyId.slice(0, endOfPublisher), name: probablyId.slice(endOfPublisher + 1, endOfName), version: probablyId.slice(endOfName + 1) };
+        const version = probablyId.slice(endOfName + 1);
+        if (!isValidVersion(version)) {
+            return undefined;
+        }
+        return { publisher: probablyId.slice(0, endOfPublisher), name: probablyId.slice(endOfPublisher + 1, endOfName), version };
     }
     /**
      * @returns `undefined` if it looks like the string passed in does not have the format returned by {@link PluginIdentifiers.toVersionedId}.
      */
     export function idAndVersionFromVersionedId(probablyId: string): IdAndVersion | undefined {
         const endOfPublisher = probablyId.indexOf('.');
-        const endOfName = probablyId.indexOf('@', endOfPublisher);
+        const endOfName = probablyId.lastIndexOf('@');
         if (endOfPublisher === -1 || endOfName === -1) {
             return undefined;
         }
-        return { id: probablyId.slice(0, endOfName) as UnversionedId, version: probablyId.slice(endOfName + 1) };
+        const version = probablyId.slice(endOfName + 1);
+        if (!isValidVersion(version)) {
+            return undefined;
+        }
+        return { id: probablyId.slice(0, endOfName) as UnversionedId, version };
+    }
+
+    function isValidVersion(probablyVersion: string): boolean {
+        // eslint-disable-next-line max-len
+        return semverRegex.test(probablyVersion);
     }
 }


packages/plugin-ext/src/common/plugin-identifiers.ts
```diff
@@ -56,15 +56,15 @@ export namespace PluginIdentifiers {
...
+        const endOfId = id.lastIndexOf('@');
         return id.slice(0, endOfId) as UnversionedId;
     }
     /**
      * @returns `undefined` if it looks like the string passed in does not have the format returned by {@link PluginIdentifiers.toVersionedId}.
      */
     export function identifiersFromVersionedId(probablyId: string): Components | undefined {
         const endOfPublisher = probablyId.indexOf('.');
-        const endOfName = probablyId.indexOf('@', endOfPublisher);
+        const endOfName = probablyId.lastIndexOf('@');
```
paul-marechal, line 67: I can imagine this fail on versions like `@acme/publisher.plugin`.

Maybe a quick `.spec.ts` file testing these functions could help here.

federicobozzini: The description of this method says `if it looks like the string passed in does not have the format returned by {@link PluginIdentifiers.toVersionedId}.`. [As far as I know the format it refers to is `${string}.${string}@${string}`](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/common/plugin-identifiers.ts#L29) so it an unversioned id name is passed it should not be supported. With the change coming from this PR if an unversion id is passed the method would not return undefined and this might indeed be a problem.

As far as I understand with scoped packages the organization name can be pretty much any identifier (even only numbers and dots) so discriminating between that and and a version would require to look at what characters are allowed in package names and in versions. [The website `semver.org` offers a regex to test if a version is valid](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string), so it might make sense to use that to test whether a version is valid.

Following this idea:

- `publisher.package-name@3.5.6` => valid
- `publisher.@3.5.6/package-name@3.5.6` => valid
- `publisher.@3.5.6` => valid (name is empty, are we happy with that? ...probably not)
- `publisher.package-name@3.5.6/name` => not valid (version cannot contain `/`)

If you are happy with this idea I'm happy to improve the information extraction according to this logic.

vince-fugnitto (changes requested):
@federicobozzini we discussed including `@` for scoped plugins and decided on not supporting the use-case for various reasons https://github.com/eclipse-theia/theia/issues/11419.

federicobozzini:
@vince-fugnitto Would there be any issue in addition to what I described and tried to fix in this PR when adding support for scoped plugins?

At least at first glance adding support for them doesn't look overly complicated.

vince-fugnitto:
@federicobozzini perhaps @colin-grant-work can expand further but it was not something we intended on supporting as the scoped plugins are really an npm concept that is not really needed for plugins.

colin-grant-work:
> @federicobozzini perhaps @colin-grant-work can expand further but it was not something we intended on supporting as the scoped plugins are really an npm concept that is not really needed for plugins.

At least as far as the VSCode API is concerned, `@` is not a legal character in either the `name` or `publisher` fields of the `package.json`. VSCE, for example, will throw an error if it encounters those characters. It is possible to end up with plugins with `@` in the descriptor if they are unpublished and unpackaged or packaged by other means, and it is true that it would be possible to support `@` and parse names differently, and I'm open to discussing it, but I think a case has to be made that allowing `@` in plugin descriptors is an important use case if we plan to depart from VSCode's standards when handling VSCode extensions. For example, I don't believe such a plugin could currently be published to either the VSCode marketplace or Open VSX (though it's possible that Open VSX has looser rules).

colin-grant-work:
> vs-code extensions might not be published at all but just part of the local monorepo. 

Certainly it is important to support local plugins - the question is whether it is important that the `name` or `publisher` fields of local plugins contain `@` signs.

> Not supporting scoped npm packages containing Theia plugins.

This is mixing two concepts: is the project in question a plugin, or is it an NPM package? Plugins aren't typically published as NPM packages or consumed as NPM packages, so conventions for the meaning of `@` in package descriptors in NPM don't really apply to plugins.

> Taking a step back, I wonder if potentially another solution to the main problem (plugins getting uninstalled because of unexpected name conflicts) would be to avoid trying to uninstall system plugins. Is this necessary?

The use case here is to allow users to install newer versions of the same plugins that are built in. If the system finds a newer version of a given plugin in the user area, it needs to mark the built-in version as not intended for use in the current session.

federicobozzini:
> Certainly it is important to support local plugins - the question is whether it is important that the `name` or `publisher` fields of local plugins contain `@` signs.

I think it's mostly a matter of keeping things consistent with Lerna scopes.

> This is mixing two concepts: is the project in question a plugin, or is it an NPM package? Plugins aren't typically published as NPM packages or consumed as NPM packages, so conventions for the meaning of `@` in package descriptors in NPM don't really apply to plugins.

You are right. I'm only referring to naming convention for the scoped NPM packages. But the fact that they are (or are not) NPM packages is irrelevant.  

> The use case here is to allow users to install newer versions of the same plugins that are built in. If the system finds a newer version of a given plugin in the user area, it needs to mark the built-in version as not intended for use in the current session.

Sounds like it's necessary then. :+1: 

One possibility in order to keep things as they are but allow some easier overriding of this behavior on the extension side would be to transform `PluginIdentifiers` into a injectable class. Unfortunately it would be a breaking change for any dependency using the current implementation. What do you think of that?

colin-grant-work:
> I think it's mostly a matter of keeping things consistent with Lerna scopes.

I believe that Lerna supports glob patterns, so if you'd like to have `@something/my-package` but also `something/my-plugin` (though of course `/` is also illegal in plugin names - but one can get away with it), `--scope *something/*` would cover both conveniently. I don't believe that `@` itself has any special significance at all to Lerna, though I could be mistaken.

> But the fact that they are (or are not) NPM packages is irrelevant.

Inasmuch as NPM packages have certain conventions for their names and plugins have other conventions for their names, and in the code in question we are dealing exclusively with plugins and never with NPM packages, the difference seems relevant.

> One possibility in order to keep things as they are but allow some easier overriding of this behavior on the extension side would be to transform PluginIdentifiers into a injectable class. Unfortunately it would be a breaking change for any dependency using the current implementation. What do you think of that?

There are a few possible solutions: 
- Switch to overridable class - definitely breaks anyone who refers to the current implementations.
- Since the code that expects `@` is internal and not easily overridable, we could use some other character, perhaps one not in the ASCII range as the version separator. That would also be breaking, but only if adopters are manually composing strings rather than using the utilities to construct versioned ID's
- Parse the current output with code that anticipates that `@` may occur in plugin identifiers.

But since the characters that should compose a plugin identifier are well known, I only feel strongly that we should be clear about whatever restrictions we impose. For example, if someone used `.` in either the `name` or `publisher` fields, the system would definitely break down, and we should reject any such plugin early. Similarly, if `@` (or any character or character sequence) will break the system, we should reject plugin identifiers that contain that / those characters. There must be some convention.

federicobozzini:
> Similarly, if @ (or any character or character sequence) will break the system, we should reject plugin identifiers that contain that / those characters. There must be some convention.

Fundamentally I agree with this.

PR #11759, opened 2022-10-12
Add `task#runOptions` to API.

### LINKED ISSUE #11132: [vscode] Support `Task#runOptions`
### Feature Description:
Support specifying [`RunOptions`](https://code.visualstudio.com/api/references/vscode-api#RunOptions) for a [`Task`](https://code.visualstudio.com/api/references/vscode-api#Task). Both is missing, the type `RunOptions` and support for the property `Task#runOptions`.

sgraband:
#### What it does

Additionally added functionality to support `reevaluateOnRerun`. If this flag is set to false no variables will be resolved again. Therefore lastTask can now be re-triggered without resolving all variables. For this the lastTask object was updated to contain all resolved information.

Resolves #11132
#### How to test

To test you can use the follwing `tasks.json`:
```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "Node Command",
      "type": "shell",
      "command": "node",
      "args": [
        "${input:node command}"
      ],
      "isBackground": true,
      "runOptions": {
        "reevaluateOnRerun": false
      }
    },
  ],
  "inputs": [
    {
      "id": "node command",
      "description": "node ${input:node command}",
      "default": "--version",
      "type": "promptString"
    },
  ]
}
```

When you run the node command you will be asked, which command you want to run. If you enter `--version` for example and then rerun the last task. You will not be asked and `node --version` is automatically executed.
When you change `reevaluateOnRerun` to true, run and rerun the task, you will need to reenter the command every time.

To test the API you can use this [extension](https://github.com/eclipse-theia/theia/files/9764231/task-provider-samples-0.0.1.zip). It provides a task (custombuildscript: 32) with `reevaluateOnRerun: true`. When you run this task and debug, for example at `task-service.ts#808` the `task` object should have the correct `runOptions` set.
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@
 
 - [Previous Changelogs](https://github.com/eclipse-theia/theia/tree/master/doc/changelogs/)
 
+## v1.32.0
+
+- [plugin] added `Task#runOptions` field and `RunOptions` interface [#11759](https://github.com/eclipse-theia/theia/pull/11759) - Contributed on behalf of STMicroelectronics
+- [tasks] added support for `reevaluateOnRerun` run option [#11759](https://github.com/eclipse-theia/theia/pull/11759) - Contributed on behalf of STMicroelectronics
+
+<a name="breaking_changes_1.32.0">[Breaking Changes:](#breaking_changes_1.32.0)</a>
+
+- [tasks] if the variables of a task should be reevaluated on a rerun (this was the behavior until now) the `reevaluateOnRerun` run option in the task description needs to be set to `true` from now on [#11759](https://github.com/eclipse-theia/theia/pull/11759) - Contributed on behalf of STMicroelectronics
+
 ## v1.31.0 - 10/27/2022
 
 - [debug] added confirmation message for debug exit [#11546](https://github.com/eclipse-theia/theia/pull/11546)
--- a/packages/plugin-ext/src/plugin/type-converters.ts
+++ b/packages/plugin-ext/src/plugin/type-converters.ts
@@ -816,6 +816,8 @@ export function fromTask(task: theia.Task): TaskDto | undefined {
     taskDto.label = task.name;
     taskDto.source = task.source;
 
+    taskDto.runOptions = task.runOptions;
+
     if ((task as types.Task).hasProblemMatchers) {
         taskDto.problemMatcher = task.problemMatchers;
     }
@@ -877,10 +879,11 @@ export function toTask(taskDto: TaskDto): theia.Task {
         throw new Error('Task should be provided for converting');
     }
 
-    const { type, taskType, label, source, scope, problemMatcher, detail, command, args, options, group, presentation, ...properties } = taskDto;
+    const { type, taskType, label, source, scope, problemMatcher, detail, command, args, options, group, presentation, runOptions, ...properties } = taskDto;
     const result = {} as theia.Task;
     result.name = label;
     result.source = source;
+    result.runOptions = runOptions ?? {};
     if (detail) {
         result.detail = detail;
     }
--- a/packages/plugin-ext/src/plugin/type-converters.ts
+++ b/packages/plugin-ext/src/plugin/type-converters.ts
@@ -816,7 +816,7 @@ export function fromTask(task: theia.Task): TaskDto | undefined {
     taskDto.label = task.name;
     taskDto.source = task.source;
 
-    taskDto.runOptions = task.runOptions;
+    taskDto.runOptions = { reevaluateOnRerun: task.runOptions.reevaluateOnRerun };
 
     if ((task as types.Task).hasProblemMatchers) {
         taskDto.problemMatcher = task.problemMatchers;
--- a/packages/plugin-ext/src/plugin/types-impl.ts
+++ b/packages/plugin-ext/src/plugin/types-impl.ts
@@ -2057,7 +2057,7 @@ export class Task {
     private taskSource: string;
     private taskGroup: TaskGroup | undefined;
     private taskPresentationOptions: theia.TaskPresentationOptions;
-    private _runOptions: theia.RunOptions;
+    private taskRunOptions: theia.RunOptions;
     constructor(
         taskDefinition: theia.TaskDefinition,
         scope: theia.WorkspaceFolder | theia.TaskScope.Global | theia.TaskScope.Workspace,
@@ -2232,14 +2232,14 @@ export class Task {
     }
 
     get runOptions(): theia.RunOptions {
-        return this._runOptions;
+        return this.taskRunOptions;
     }
 
     set runOptions(value: theia.RunOptions) {
         if (value === null || value === undefined) {
             value = Object.create(null);
         }
-        this._runOptions = value;
+        this.taskRunOptions = value;
     }
 }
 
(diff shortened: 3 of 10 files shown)


packages/task/src/browser/task-service.ts
```diff
@@ -496,11 +501,14 @@ export class TaskService implements TaskConfigurationClient {
      * @param token  The cache token for the user interaction in progress
      */
     async runLastTask(token: number): Promise<TaskInfo | undefined> {
-        if (!this.lastTask) {
+        if (!this.lastTask || !this.lastTask.resolvedTask) {
             return;
         }
-        const { source, taskLabel, scope } = this.lastTask;
-        return this.run(token, source, taskLabel, scope);
+        if (!this.lastTask.resolvedTask.runOptions?.reevaluateOnRerun) {
```
colin-grant-work, line 507: > If this flag is set to false no variables will be resolved again. Therefore lastTask can now be re-triggered without resolving all variables.

It appears that `reevaluateOnRerun` is being treated as the default. That's a change in the expectations for internal tasks, which in the past would always have been resolved when rerun. Is there any value in treating internal and external tasks differently in this respect?

packages/task/src/browser/task-service.ts
```diff
@@ -496,11 +501,14 @@ export class TaskService implements TaskConfigurationClient {
      * @param token  The cache token for the user interaction in progress
      */
     async runLastTask(token: number): Promise<TaskInfo | undefined> {
-        if (!this.lastTask) {
+        if (!this.lastTask || !this.lastTask.resolvedTask) {
```
colin-grant-work, line 504:
```suggestion
        if (!this.lastTask?.resolvedTask) {
```

packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -2034,6 +2034,7 @@ export class Task {
     private taskSource: string;
     private taskGroup: TaskGroup | undefined;
     private taskPresentationOptions: theia.TaskPresentationOptions;
+    private _runOptions: theia.RunOptions;
```
tsmaeder, line 2037: While would prefer that we keep the naming of the fields consistent inside the same class: let's fix all of the names to follow the `_` convention or call the field `taskRunOptions`.

packages/plugin-ext/src/plugin/type-converters.ts
```diff
@@ -783,6 +783,8 @@ export function fromTask(task: theia.Task): TaskDto | undefined {
     taskDto.label = task.name;
     taskDto.source = task.source;
 
+    taskDto.runOptions = task.runOptions;
```
tsmaeder, line 786: I am aware that this pattern is used elsewhere, but I'm not sure this is a safe assignment: this interface can be implemented by the user. Imagine the user did it like this:

``` 
export class RunOptionsImpl implements theia.RunOptions {
   private _reevaluateOnRerun;

   get reevaluateOnRerun(): boolean {
      return this._reevaluateOnRerun;
   }

   set reevaluateOnRerun(boolean doIt) {
      this._reevaluateOnRerun= doit;
   }
}
```

if we assign the object to a field in a TaskDTO and then proceed to `JSON.stringify()` the task, we'll end up with an object on the other side of the remote service which does _not_ implement `theia.RunOptions`. As far as I remember, `stringify()` does not respect property accessors.

packages/task/src/browser/task-service.ts
tsmaeder: @colin-grant-work by "internal", do you mean tasks defined in `tasks.json` or contributed by non-plugin API? If we need to treat tasks different, I would much prefer putting the difference where we _create_ the tasks (for example by setting a flag in the contructor) and not where we process the tasks. Locality of reference for teh win!

packages/plugin-ext/src/plugin/type-converters.ts
sgraband: Good point! What should we do to mitigate that? Assign an object in this line? Because
```
    taskDto.runOptions = { reevaluateOnRerun: task.runOptions.reevaluateOnRerun };
```
should mitigate this problem, right?

packages/task/src/browser/task-service.ts
colin-grant-work: Re: `tasks.json`, we should likely follow what VSCode does, though there's scope for making our own choice there. Mostly I mean other tasks created internally, but I'm not very familiar with the behavior or UI of 'rerun' tasks (as opposed to running the same task again, but not as a rerun), so I'm not sure how disruptive it would be to change from resolving to not resolving variables in the task description.

 One solution might be to modify the plugin system to supply `false` by default for tasks originating from plugins, and then have a flag in the task service for whether to use `false` or `true` as the default value there so that application implementers can easily toggle it if they have a use case where `true` is appropriate. On the other hand, it sounds like @tsmaeder favors something on the task description itself (in which case maybe setting `false` in the plugin system and using `true` as a default otherwise would be the best course?). In any case, I don't know if it really warrants too much concern.

tsmaeder: I thought about this some more: what I _really_ would like to avoid is a "if tasksource = plugin, do x, else do y". I'm also not a fan of setting variables that are client API in order to simulate the "right" default behaviour: if anything, we should set a flag : `task.defaultRerunOnReevaluate= false` that is _not_ part of the API in the plugin task provider implementation and default to true in clients to preserve the current behaviour. 
But t.b.h, I would just make the change to the VS Code behavior: it's a breaking change, and while we try not to frivolously break API, we also should not unduly complicate the system to keep the API stable.
My vote is: put the code in as it is and make a note of it in the "breaking changes". Clients can trivially update their code if they rely on the behaviour.

packages/plugin-ext/src/plugin/type-converters.ts
tsmaeder: I think that should do the trick.

packages/task/src/browser/task-service.ts
colin-grant-work: I'm fine with this approach.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 - [plugin] added support for the `InlineValues` feature [#11729](https://github.com/eclipse-theia/theia/pull/11729) - Contributed on behalf of STMicroelectronics
 - [plugin] Added support for `resolveTreeItem` of `TreeDataProvider` [#11708](https://github.com/eclipse-theia/theia/pull/11708) - Contributed on behalf of STMicroelectronics
+- [plugin] added `Task#runOptions` field and `RunOptions` interface [#11759](https://github.com/eclipse-theia/theia/pull/11759) - Contributed on behalf of STMicroelectronics
+- [tasks] added support for `reevaluateOnRerun` run option [#11759](https://github.com/eclipse-theia/theia/pull/11759) - Contributed on behalf of STMicroelectronics
```
colin-grant-work, line 12: As suggested by @tsmaeder, please add a note about the change in behavior for tasks in the breaking changes area.

sgraband: Done. Could you recheck?

colin-grant-work (approved):
The behavior looks good, and adopters should be adequately warned re: any effect on their current setup.

PR #11762, opened 2022-10-13
Added unit tests for the functions in object.

EstFoisy:
#### What it does

The pull-request adds unit-tests for methods in `object.ts`.

It test functions from object file.

#### How to test

The CI should pass for the newly added tests.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/objects.spec.ts
+++ b/packages/core/src/common/objects.spec.ts
@@ -1,0 +1,112 @@
+// *****************************************************************************
+// Copyright (C) 2022 Ericsson and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { expect } from 'chai';
+import { deepClone, deepFreeze, isEmpty, notEmpty } from './objects';
+
+describe('Objects', () => {
+
+    describe('#deepClone', () => {
+
+        describe('primitives', () => {
+            it('should deep clone numbers', () => {
+                expect(1).to.equal(deepClone(1));
+            });
+            it('should deep clone strings', () => {
+                expect('foo').to.equal(deepClone('foo'));
+                expect('').to.equal(deepClone(''));
+            });
+            it('should deep clone booleans', () => {
+                expect(true).to.equal(deepClone(true));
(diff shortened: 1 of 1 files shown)


packages/core/src/common/object.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+    describe('#isEmpty', () => {
+        it('should return "true" when it is Empty', () => {
+            const z = {};
+            expect(isEmpty(z)).to.equal(true);
+        });
+        it('should return "false" when not Empty', () => {
+            const z = { d: 5 };
+            expect(isEmpty(z)).to.equal(false);
+        });
+    });
```
colin-grant-work, line 109: `isEmpty` checks two features of an object: that it has keys, and that its constructor is `Object`. These tests only confirm the behavior of one check. How can we check the `Object` constructor feature of `isEmpty`?

EstFoisy: Hi @colin-grant-work , 
Thanks a lot for advising about the above: I have revised my units tests according to your comment.

packages/core/src/common/object.spec.ts
```diff
@@ -0,0 +1,111 @@
+// *****************************************************************************
```
vince-fugnitto, line 1: Given that the file is `objects.ts` the file should be renamed to `objects.spec.ts` (notice the missing `s`).

packages/core/src/common/object.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+            expect(notEmpty({ a: 1 })).to.equal(true);
+        });
+        it('should return "false" when empty', () => {
+            expect(notEmpty(undefined)).to.equal(false);
+        });
+    });
+
+    describe('#isEmpty', () => {
+        it('should return "true" when it is Empty', () => {
+            const obj = Object.create({});
```
vince-fugnitto, line 102: I don't believe this tests the non-object `constructor`.


```bash
$ node
Welcome to Node.js v16.14.0.
Type ".help" for more information.
> a = Object.create({})
{}
> a.constructor
[Function: Object]
> a.constructor === Object
true

```

packages/core/src/common/object.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+            expect(notEmpty([])).to.equal(true);
+            expect(notEmpty({ a: 1 })).to.equal(true);
+        });
+        it('should return "false" when empty', () => {
+            expect(notEmpty(undefined)).to.equal(false);
+        });
+    });
+
+    describe('#isEmpty', () => {
+        it('should return "true" when it is Empty', () => {
```
vince-fugnitto, line 101:
```suggestion
        it('should return "true" when it is empty', () => {
```

packages/core/src/common/object.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+            const obj = Object.create({});
+            expect(isEmpty(obj)).to.equal(true);
+        });
+        it('should return "false" when not Empty', () => {
```
vince-fugnitto, line 105:
```suggestion
        it('should return "false" when not empty', () => {
```

vince-fugnitto (approved):
LGTM :+1:

Issue #11659, opened 2022-09-12
vscode: add full support for `Diagnostic.code`


vince-fugnitto:
#### Feature Description:

The framework currently only has partial support for `Diagnostic.code`:

https://github.com/eclipse-theia/theia/blob/62ce7862ef820d91dbb7ef731c58c6e5eef246a5/packages/plugin/src/theia.d.ts#L8070-L8075

We should update to have full support according to the VS Code API:

```ts
/**
 * A code or identifier for this diagnostic.
 * Should be used for later processing, e.g. when providing {@link CodeActionContext code actions}.
 */
code?: string | number | {
    /**
     * A code or identifier for this diagnostic.
     * Should be used for later processing, e.g. when providing {@link CodeActionContext code actions}.
     */
    value: string | number;

    /**
     * A target URI to open with more information about the diagnostic error.
     */
    target: Uri;
};
```

PR #11765, opened 2022-10-13
Add full support for Diagnostic.code

EstFoisy:
What it does
Add full support to Diagnostic.code according to the VS Code API.

closes https://github.com/eclipse-theia/theia/issues/11659

How to test
CI should pass

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/type-converters.spec.ts
+++ b/packages/plugin-ext/src/plugin/type-converters.spec.ts
@@ -437,4 +437,24 @@ describe('Type converters:', () => {
             assert.deepStrictEqual(result, showOptions);
         });
     });
+
+    describe('#convertCode', () => {
+        it('should convert a "code" of type "string"', () => {
+            assert.strictEqual(Converter.convertCode('string'), 'string');
+        });
+        it('should convert a "code" of type "number"', () => {
+            assert.strictEqual(Converter.convertCode(4), '4');
+        });
+        it('should convert an undefined "code"', () => {
+            assert.strictEqual(Converter.convertCode(undefined), undefined);
+        });
+        it('should convert a "code" of type "{ value: number, target: Uri }"', () => {
+            const uri = types.URI.parse('foo://example.com:8042/over/there?name=ferret#nose');
+            assert.strictEqual(Converter.convertCode({ value: 4, target: uri }), '4');
+        });
+        it('should convert a "code" of type "{ value: number, target: Uri }"', () => {
+            const uri = types.URI.parse('foo://example.com:8042/over/there?name=ferret#nose');
+            assert.strictEqual(Converter.convertCode({ value: 'string', target: uri }), 'string');
+        });
+    });
 });
--- a/packages/plugin-ext/src/plugin/type-converters.ts
+++ b/packages/plugin-ext/src/plugin/type-converters.ts
@@ -332,12 +332,15 @@ export function convertDiagnosticToMarkerData(diagnostic: theia.Diagnostic): mod
     };
 }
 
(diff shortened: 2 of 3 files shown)


packages/plugin-ext/src/plugin/type-converters.spec.ts
```diff
@@ -437,4 +437,24 @@ describe('Type converters:', () => {
...
+        it('should convert an undefined "code"', () => {
+            assert.strictEqual(Converter.convertCode(undefined), undefined);
+        });
+        it('should convert a "code" of type "{ value: number, target: Uri }"', () => {
+            const uri = types.URI.parse('foo://example.com:8042/over/there?name=ferret#nose');
+            assert.strictEqual(Converter.convertCode({ value: 4, target: uri }), '4');
+        });
+        it('should convert a "code" of type "{ value: number, target: Uri }"', () => {
+            const uri = types.URI.parse('foo://example.com:8042/over/there?name=ferret#nose');
+            assert.strictEqual(Converter.convertCode({ value: 'string', target: uri }), 'string');
```
colin-grant-work, line 457: It looks like there's loss of information here. Is there a reason we don't want to include the `target` field in the `code` when we convert it? I notice that the corresponding code in VSCode does preserve the target:

https://github.com/microsoft/vscode/blob/65a9097aa2da4d55318073ca23bc678a2885c461/src/vs/workbench/api/common/extHostTypeConverters.ts#L198-L207

vince-fugnitto: @colin-grant-work I believe it was done this way due to the conversion of `vst.Diagnostic` (`vscode-languageserver-types`) which did not support the new type:

https://github.com/eclipse-theia/theia/blob/e96bb5b435808b2a2a2148527b71d3d1cb0ba531/packages/plugin-ext/src/main/browser/languages-main.ts#L1072-L1080

```ts
/**
 * Represents a diagnostic, such as a compiler error or warning. Diagnostic objects
 * are only valid in the scope of a resource.
 */
export interface Diagnostic {
    /**
     * The range at which the message applies
     */
    range: Range;
    /**
     * The diagnostic's severity. Can be omitted. If omitted it is up to the
     * client to interpret diagnostics as error, warning, info or hint.
     */
    severity?: DiagnosticSeverity;
    /**
     * The diagnostic's code, which usually appear in the user interface.
     */
    code?: number | string;
    /**
     * A human-readable string describing the source of this
     * diagnostic, e.g. 'typescript' or 'super lint'. It usually
     * appears in the user interface.
     */
    source?: string;
    /**
     * The diagnostic's message. It usually appears in the user interface
     */
    message: string;
    /**
     * Additional metadata about the diagnostic.
     */
    tags?: DiagnosticTag[];
    /**
     * An array of related diagnostic information, e.g. when symbol-names within
     * a scope collide all definitions can be marked via this property.
     */
    relatedInformation?: DiagnosticRelatedInformation[];
}
```

We could from the `types-converter` convert fully, and only when it comes time to `reviveMarker` we update to ensure the types match.

JonasHelming:
@EstFoisy Do you plan to look at the review comments?

EstFoisy:
Hi @JonasHelming  ,
Many thanks for advising. I was working with Vince for this PR and, from my understanding of the discussion I had with him, everything was ok and ready for this PR.  I will look into it, try to fix the code to the best of my understanding, and see with him when he come back from vacation if all is good or if there is any further adjustment needed.

JonasHelming:
@vince-fugnitto @EstFoisy : Maybe we could get this into the January release?

PR #11771, opened 2022-10-14
Fixed issue with language server extensions

federicobozzini:
#### What it does

It fixes the error discussed at https://github.com/eclipse-theia/theia/pull/11752 . Language server extensions were not loaded properly. With this PR they are considered backend extension and loaded correctly.

#### How to test

Start Vanilla theia and observe the logs. All 76 builtin are loaded.

#### Review checklist

#### Reminder for reviewers


federicobozzini:
@vince-fugnitto  @colin-grant-work

packages/plugin-ext-vscode/src/node/plugin-vscode-directory-handler.ts
```diff
@@ -58,6 +58,10 @@ export class PluginVsCodeDirectoryHandler implements PluginDeployerDirectoryHand
             types.push(PluginDeployerEntryType.FRONTEND);
         }
         context.pluginEntry().accept(...types);
+        if (types.length === 0) {
+            // Language server extension
+            context.pluginEntry().accept(PluginDeployerEntryType.BACKEND);
+        }
```
vince-fugnitto, line 64: Rather than have to `accept` twice we can implement a fallback like so?

```ts
async handle(context: PluginDeployerDirectoryHandlerContext): Promise<void> {
    await this.copyDirectory(context);
    const types: PluginDeployerEntryType[] = [];
    const packageJson: PluginPackage = context.pluginEntry().getValue('package.json');
    if (packageJson.browser) {
        types.push(PluginDeployerEntryType.FRONTEND);
    }
    if (packageJson.main || !packageJson.browser) {
        types.push(PluginDeployerEntryType.BACKEND);
    }
    context.pluginEntry().accept(...types);
}
```

federicobozzini: I like it :+1:

vince-fugnitto (commented):
@federicobozzini I confirmed that plugins are now properly loaded in both the browser and electron applications :+1: 

---
_Side Note_


I'm yet to use web extensions but I believe they should only work in the remote or browser context correct? It seems as though we still load some content but not the command handler for instance. I verified by creating a hello-world web extension from the vscode generator:
- [hello-0.0.1.zip](https://github.com/eclipse-theia/theia/files/9788534/hello-0.0.1.zip)

The browser will correctly execute the `hello world` command while electron will throw an error:

image

vince-fugnitto (approved):
In any case, the changes fix the bug where plugins were not loaded properly :+1:

Issue #10194, opened 2021-09-29
Improve Theia extension inter-dependencies


paul-marechal:
# Current State

Theia extensions currently depend on each other as hard dependencies. This means that when you install something like `@theia/plugin-ext@1.17.2` your package manager will pull all of its dependencies with the same exact version.

We had to depend on exact versions when publishing to avoid some pain when Theia extension consumers, because having different ranges and whatnot led to duplicate installations of the same Theia packages.

But this starts to cause issues when trying to use third-party extensions that might not be updated as often, and then including those in your applications can lead to duplicated Theia core packages.

The whole issue revolves around defining inter-Theia extension dependencies as regular dependencies. Doing so means package managers make no guaranty that a package will only be installed once.

# Peer Dependencies

Conceptually Theia extensions are build-time plugins that must be installed only once with a specific version, and they rely on other extensions to be installed alongside them. We even mount them together to create one big Inversify container.

A better way of expressing this relationship would be to use peer dependencies to link Theia extensions together.

Pros:
(a) Package managers should not try to pull different versions of the same Theia package.
(b) Ranges can be larger again, what's important is to target a minimal version.
(c) The larger ranges mean that it will be easier to consume third party Theia extensions.

Cons:
(d) Package managers will most likely not pull anything anymore.
(e) Dependencies on Theia extensions all must be explicitly declared in the application's package.

The biggest reason I bring this issue up again is because of (c): Maintaining and consuming third-party extensions is a bit hellish.

Fortunately (d) and (e) can be fixed by adding the logic in `@theia/cli` to properly update/check an application's package so that everything is in working order (e.g. add/ignore missing extensions, ensure versions match, etc.)

# Proposal

- Rewrite Theia extensions dependencies to work with peer dependencies (only for Theia extension packages).
- Use broader ranges for those peer-dependencies (making it easier to consume extensions not updated as often as core).
- Make `@theia/cli` ensure applications are still built correctly.

tsmaeder:
Good idea, however, I'm not sure we need to be more lenient with the peer dependencies among theia packages: as I understand it, we always update all version numbers in parallel: if third-party packages depend on different version of theia, their dependency requirements are either compatible or not compatible: if they are not, I would consider that an error. It's up to third-party extensions to be lenient, not core Theia.

paul-marechal:
[...]

PR #5939, opened 2019-08-13
POC: Use peer dependencies for @theia dependencies

marechal-p:
#### What it does

It moves all `@theia` dependencies to "peer dependencies". This should allow the dependency tree to only look for one extension version instead of duplicating it. See https://nodejs.org/es/blog/npm/peer-dependencies/

Only quirk is that in order for lerna to understand the package topology, we need to declare the packages as dev dependencies.

#### How to test

Build and run (might be broken for now, since this is a draft/poc)

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #11808, opened 2022-10-27
Support for declaring extensions as peer dependencies

tortmayr:
#### What it does

Ensures that the ExtensionPackageCollector also correctly picks up packages that are declared as peer dependencies.

This change allows extension providers to declare their dependencies to Theia core extensions as peer dependencies. This can simplify version management and eliminates the need for yarn resolutions (or a similar feature) to ensure that only one version of Theia extensions is resolved. 

This topics has already been discussed multiple times:
-  https://github.com/eclipse-theia/theia/discussions/10791
- https://github.com/eclipse-theia/theia/issues/10194
- https://github.com/eclipse-theia/theia/pull/5939

The question whether we actually want to change inter dependencies between Theia core packages to peer dependencies is still up for discussion. 
Nevertheless, ensuring the the application builder also supports and correctly loads Theia extensions that are declared as peer dependencies is a good idea.

This is an opt-in feature that does not break any existing applications and gives third party extension providers the opportunity to use peer dependencies if they want to.

#### How to test

Can for instance be tested with the "@theia/api-samples" package by simply switching all dependencies to peerDependencies:
```diff
- "dependencies": {
+ "peerDependencies": {
    "@theia/core": "1.30.0",
    "@theia/file-search": "1.30.0",
    "@theia/filesystem": "1.30.0",
    "@theia/monaco": "1.30.0",
    "@theia/monaco-editor-core": "1.67.2",
    "@theia/output": "1.30.0",
    "@theia/search-in-workspace": "1.30.0",
    "@theia/toolbar": "1.30.0",
    "@theia/vsx-registry": "1.30.0",
    "@theia/workspace": "1.30.0"
  },
```

1. Building this change on master results in a corrupt application. I.e. the browser application no longer loads and the following error is logged:
Screenshot from 2022-10-27 18-15-20
In the generated frontend index.js file you can see that the api-samples frontend module gets loaded before the required core modules which causes the error.

2. Building this change on the PR branch causes no errors. The browser application loads as expected and the module loading is in correct order.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-package/src/extension-package-collector.ts
+++ b/dev-packages/application-package/src/extension-package-collector.ts
@@ -36,13 +36,11 @@ export class ExtensionPackageCollector {
     }
 
     protected collectPackages(pck: NodePackage): void {
-        if (!pck.dependencies) {
-            return;
-        }
-        // eslint-disable-next-line guard-for-in
-        for (const dependency in pck.dependencies) {
-            const versionRange = pck.dependencies[dependency]!;
-            this.collectPackage(dependency, versionRange);
+        for (const [dependency, versionRange] of [
+            ...Object.entries(pck.dependencies ?? {}),
+            ...Object.entries(pck.peerDependencies ?? {})
+        ]) {
+            this.collectPackage(dependency, versionRange!);
         }
     }
 
--- a/dev-packages/application-package/src/npm-registry.ts
+++ b/dev-packages/application-package/src/npm-registry.ts
@@ -47,6 +47,7 @@ export interface NodePackage {
     maintainers?: Maintainer[];
     keywords?: string[];
     dependencies?: Dependencies;
+    peerDependencies?: Dependencies;
     [property: string]: any;
 }
 


dev-packages/application-package/src/extension-package-collector.ts
```diff
@@ -36,13 +36,20 @@ export class ExtensionPackageCollector {
     }
 
     protected collectPackages(pck: NodePackage): void {
-        if (!pck.dependencies) {
-            return;
+        if (pck.dependencies) {
```
msujew, line 39: Suggestion: I believe we can cut down on the eslint-disable stuff as well as the duplicated for-loop by replacing the method body with something like this:

```ts
for (const [dependency, versionRange] of [
    ...Object.entries(pck.dependencies ?? {}),
    ...Object.entries(pck.peerDependencies ?? {})
]) {
    this.collectPackage(dependency, versionRange!);
}
```

tortmayr: Thanks! That's a lot cleaner and works like charm. I have updated the PR accordingly.

JonasHelming:
@msujew Could you do a re review? Would be great to have this in the December release (candidate for the community release)

msujew (approved):
The change looks good to me 👍

I would like a second opinion on that though, @paul-marechal WDYT?

paul-marechal (approved):
Simple enough, looks good to me.

23444577gyghuyhhg:
sylvan-chess-371100

23444577gyghuyhhg:
sylvan-chess-371100

PR #11831, opened 2022-11-03
Simplify type checking for objects

msujew:
#### What it does

A follow up on https://github.com/eclipse-theia/theia/pull/11490 (cc @paul-marechal)

Introduces an `Is` namespace with helpful methods to identify types. Uses the newly introduced methods to simplify type checking for objects. I didn't use the `Is.string`, `Is.number` etc functions as that would make the PR even larger and I wasn't sure whether it would make sense to use the functions that way anyway.

For some reason, this change required an additional call to `reflect-metadata` in the plugin host. Otherwise it would crash as the `ContributionFilterRegistryImpl` would try to annotate itself without the `Reflect.hasOwnMetadata` function available.

#### How to test

Nothing should've changed

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (commented):
@msujew should we maybe update [packages/core/src/common/types.ts](https://github.com/eclipse-theia/theia/blob/master/packages/core/src/common/types.ts) to include similar methods present in [vscode](https://github.com/microsoft/vscode/blob/main/src/vs/base/common/types.ts) (copying them over)?

As trivial as the checks might be I think unit-testing them might be good too.

paul-marechal:
I pushed a commit with an alternative writing: The reasoning is that the function names should better describe what it does (`isObject` vs `object`) and if users want to use the functions under a namespace then can import as:

```ts
import * as types from '@theia/core/lib/common/types';
// ...
types.isObject(...);
```

paul-marechal (approved):
I like these new functions as they greatly help when dealing with `unknown`-typed variables!

paul-marechal:
@msujew feel free to merge if you agree with the changes suggested in the added commit, otherwise drop it before.

msujew:
@paul-marechal Thanks, I didn't have time recently to update this PR :)

Your changes look good to me 👍

PR #11432, opened 2022-07-15
Fix circular dependencies in packages with few

colin-grant-work:
#### What it does

When we run do a build or start a watch, we are informed that WebPack has many warnings for us, most of them to do with circular dependencies. This PR goes some way to fixing circular dependencies, although many still remain. It uses the ESLint rule  import/no-cycle as its diagnostic, and eventually it would be desirable to activate that rule for the repository. However, since both `core` and `plugin-ext` have quite a number of difficult to untangle circularities, that goal is still some distance away.

#### How to test

1. CI should pass.
2. The number of warnings displayed when watching should be four fewer than on `master`.
> Before: webpack 5.73.0 compiled with 80 warnings in 16221 ms
> After: webpack 5.73.0 compiled with 72 warnings in 2270 ms

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

Issue #7751, opened 2020-05-05
Blank square instead of plugins icons in sidebar when theia starts


DoroNahari:
### Bug Description:
Blank square instead of plugins icons in sidebar when theia starts.

### Steps to Reproduce:

1. Deploy a vscode extension that have a side pane such as gitlen. 
2. See that new pane added to the left pane
3. Refresh browser and see that for about 1 second there's blank square instead of the plugin icon.

screen recording of my local Theia:
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/24614792/81097133-21f25600-8f10-11ea-94ab-24fe65c391d5.gif)

### Additional Information

- Operating System: Mac Os Catalina
- Theia Version: master 1.1.0

PR #11835, opened 2022-11-04
Fix some webpack warnings

msujew:
#### What it does

Related to https://github.com/eclipse-theia/theia/pull/11432

This change fixes/circumvents 3 kinds of webpack warnings:
1. Missing source maps. We ignore those, as these come from dependencies and it doesn't really harm us too much
2. Weird `require` usage in monaco on a path which is never executed. We can ignore that as well
3. Unable to inline `url('')` css. I've replaced the URL with a gradient which serves the same result

#### How to test

Only the third point can actually be tested. It's a regression test against https://github.com/eclipse-theia/theia/issues/7751. The plugin icons should be empty instead of a grey box.

For everything else: The related webpack warnings shouldn't appear anymore when running `theia build`.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-manager/src/generator/webpack-generator.ts
+++ b/dev-packages/application-manager/src/generator/webpack-generator.ts
@@ -226,7 +226,16 @@ module.exports = [{
     stats: {
         warnings: true,
         children: true
-    }
+    },
+    ignoreWarnings: [
+        // Some packages do not have source maps, that's ok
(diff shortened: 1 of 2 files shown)


vince-fugnitto (approved):

- [x] confirmed that source map warnings are no longer present
- [x] custom icons from plugins do not display the grey square on reload

PR #11853, opened 2022-11-09
Plugin localization service does not await on init

kittaakos:
#### What it does

 - This PR changes the initialization (`BackendApplicationContribution#initialize`) of the localization service for plugins. With the proposed changes, the backend initialization does not wait for the cleaning of the localization cache folder.
 - The PR also changes the return type of the `initialize` method from `Promise<void>` to `MaybePromise<void>` not to force downstream projects to use `aync`/`Promise` when doing the customization.

Why?

To speed up the backend startup when the Theia-based app is stated with multiple languages.

Originally from [here](https://github.com/eclipse-theia/theia/pull/11338#issuecomment-1308653707):

> I am seeing slow startup time with multiple languages. It's more than a sec.
> 
> ```
> 2022-11-09T11:46:27.180Z root WARN Backend HostedPluginLocalizationService.initialize took longer than the expected maximum 50 milliseconds: 1070.7 ms [Finished 4.488 s after backend start]
> ```
> 
> Could something like this also work in Theia?
> 
> ```ts
> initialize(): void {
> 	this.getLocalizationCacheDir().then((cacheDir) => fs.emptyDir(cacheDir)).then(() => this._ready.resolve())
> }
> ```
> 
> With the proposed change, the initialization is significantly less.
> 
> ```
> 2022-11-09T12:05:46.544Z root INFO Backend HostedPluginLocalizationService.initialize: 3.0 ms [Finished 2.557 s after backend start]
> ```
> 
> Thank you!

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/hosted/node/hosted-plugin-localization-service.ts
+++ b/packages/plugin-ext/src/hosted/node/hosted-plugin-localization-service.ts
@@ -23,7 +23,7 @@ import { DeployedPlugin, Localization as PluginLocalization, PluginIdentifiers }
 import { URI } from '@theia/core/shared/vscode-uri';
 import { EnvVariablesServer } from '@theia/core/lib/common/env-variables';
 import { BackendApplicationContribution } from '@theia/core/lib/node';
-import { Disposable } from '@theia/core';
+import { Disposable, MaybePromise } from '@theia/core';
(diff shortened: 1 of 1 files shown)


msujew (approved):
Alright, that looks good to me 👍

PR #11857, opened 2022-11-10
Fix potential race condition when copying files and directories

erezmus:
#### What it does
This fixes a potential race condition when copying files with directories. The call to `mkdirp` was not awaited, meaning that subsequent copying of files within the copied directory would occasionally fail as the directory creation had not finished.

#### How to test
We found this issue while duplicating a project in ARM mbed-ide, so presumably, copying a large git project will suffice.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/node/disk-file-system-provider.ts
+++ b/packages/filesystem/src/node/disk-file-system-provider.ts
@@ -708,7 +708,7 @@ export class DiskFileSystemProvider implements Disposable,
         copiedSources[source] = true; // remember as copied
 
         // Create folder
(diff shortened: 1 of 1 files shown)


thegecko (approved):
Sensible change :+1:

msujew (approved):
I can confirm that the issue exists on `master` and is addressed by this change 👍

thegecko:
I'm merging this even though the Node 16 CI is failing. I tried poking it again, but also see this has sporadic failures on master.

PR #11860, opened 2022-11-10
ci: make plugin download sequential

vince-fugnitto:
#### What it does

The pull-request updates the `download:plugins` script to use `sequential` downloads as the new default (rather than `parallel`).

The changes include:
- the default for `download:plugins` is now set to `parallel=false` (sequential)
- the `build` step during CI is broken down into their respective parts (Install, Download Plugins and Build)
- we no longer unnecessarily call `download;plugins` for MacOS and Windows as we do not run integration tests there

#### How to test

Errors related to `download:plugins` (429) should occur less during CI.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -73,11 +73,24 @@ jobs:
         with:
           python-version: '3.x'
 
-      - name: Install and Build
+      - name: Install
         shell: bash
         run: |
           yarn --skip-integrity-check --network-timeout 100000
           ./scripts/check_git_status.sh
+        env:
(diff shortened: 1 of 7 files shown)


msujew (commented):
We should also respect this change in the gitpod config:
```
yarn --network-timeout 100000 && yarn build:examples
```
to
```
yarn --network-timeout 100000 && yarn build:examples && yarn download:plugins --parallel=false
```

msujew (approved):
Looks good to me 👍 

I believe we should get this change into `master` before the release.

Issue #11827, opened 2022-11-03
vscode: update typings for `provideHover`


vince-fugnitto:
#### Feature Description:

The typings for `HoverProvider.provideHover` should be updated to align with the VS Code API:

https://github.com/eclipse-theia/theia/blob/1cdfae21899f1efacb2a8caec7f2147129518c8d/packages/plugin/src/theia.d.ts#L10070

The changes should be (drop `undefined` from `token`):

```diff
- provideHover(document: TextDocument, position: Position, token: CancellationToken | undefined): ProviderResult<Hover>;
+ provideHover(document: TextDocument, position: Position, token: CancellationToken): ProviderResult<Hover>;
```

PR #11862, opened 2022-11-11
vscode: align typings for provideHover

planger:
#### What it does

Aligns the typings of `HoverProvider.provideHover` with the API of VS Code.
This change also updates the documentation to include the token parameter in the signature of the hover provider examples. Technically, this isn't necessary, but for the sake of completeness, I thought it makes sense.

Fixes https://github.com/eclipse-theia/theia/issues/11827

#### How to test

This isn't really a breaking change and only impacts the external extension API, which isn't used directly (I assume) anyway, as people most likely implement against the VS Code API anyway. Also, this method is never invoked directly from within Theia too. So there is nothing really to test, imho.

I locally confirmed that this change makes the [report](https://eclipse-theia.github.io/vscode-theia-comparator/status.html) happy.

#### Review checklist

#### Reminder for reviewers


msujew (approved):
I can confirm that this change correctly aligns the typing for the `provideHover` method, as indicated by the vscode comparator report.

PR #11864, opened 2022-11-14
Remove circular dependency check

msujew:
#### What it does

See https://github.com/eclipse-theia/theia/discussions/11863

Removes circular dependency checking from the generated webpack build. Downstream users still have the option to add the check themselves

#### How to test

Build the application. No circular dependency warnings should appear. In fact, this should remove all warnings that are left in the webpack build.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-manager/package.json
+++ b/dev-packages/application-manager/package.json
@@ -39,7 +39,6 @@
     "@types/semver": "^7.3.8",
     "babel-loader": "^8.2.2",
     "buffer": "^6.0.3",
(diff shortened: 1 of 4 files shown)


vince-fugnitto:
> @vince-fugnitto Do you think this warrants an entry in the migration guide?

@msujew if we can I think it would be good to document, it might not be super clear on how downstream adopters can include their own custom plugins to webpack.

vince-fugnitto (approved):
LGTM :+1:

Issue #11594, opened 2022-08-22
New React error in TreeWidget after `react-virtuoso` change


colin-grant-work:
### Steps to Reproduce:

1. Start the application
2. (Maybe open the plugin view?)
3. See an error like this: 

![image](https://user-images.githubusercontent.com/62660806/185945957-7515fc01-1a8c-46b0-9bca-a279e3fcc6db.png)

### Additional Information

- Operating System: Windows 10, Edge browser
- Theia Version: 3e664e7adcef4e19321ddfe209ec17796fa96bf4

colin-grant-work:
@msujew

msujew:
@colin-grant-work Thanks, I'll look into it 👍

Issue #10867, opened 2022-03-09
React Tooltips can get stuck


colin-grant-work:
### Bug Description:

There are conditions under which a React tooltip will get stuck even though it no longer applies.

### Steps to Reproduce:

I've noticed this in a few cases, but here's one that's reproducible:

1. Have a non-builtin plugin installed.
2. Hover over the plugin so that the tooltip shows up.
3. Click `Uninstall` so that the tree node for the plugin disappears.
4. Observe that the tooltip stays.

![image](https://user-images.githubusercontent.com/62660806/157498383-1dab70b1-7633-4676-82d9-f3d98b31f092.png)

### Additional Information

- Operating System: Ubuntu 20 (Firefox)
- Theia Version: e62fd33ce46055d1633d781c08bf5852535f53c5

PR #11708, opened 2022-09-28
vscode: Add support for `resolveTreeItem` to `TreeDataProvider`

lucas-koehler:
#### What it does

* Extend `theia.d.ts` with optional method `resolveTreeItem` in `TreeDataProvider`
* Add infrastructure on ext side to resolve tree items
* Add resolvable tree view node classes
* Add resolvement for tooltips and commands
* Align TreeViewItem and TreeViewNode tooltip property type to include `MarkdownString`

Fixes #11147

Contributed on behalf of STMicroelectronics

Signed-off-by: Lucas Koehler <lkoehler@eclipsesource.com>

#### How to test

- Download the test extension [here](https://github.com/lucas-koehler/vscode-extension-samples/releases/download/TreeDataProvider-resolveTreeItem-0.0.1/tree-view-sample-resolve-items-0.0.1.vsix)
  - See this [release](https://github.com/lucas-koehler/vscode-extension-samples/releases/tag/tree-item-label-0.0.1) and find the code that implements TreeDataProvider.resolveTreeItem [here](https://github.com/lucas-koehler/vscode-extension-samples/blob/TreeDataProvider-resolveTreeItem-0.0.1/tree-view-sample/src/testView.ts)
- The extension adds a view `TEST VIEW` to the explorer. The `a` element subtree resolves tooltips and commands while the `b` element subtree has both given as part of the original tree items.
image
- Hover over `a` and `b` sub(elements): Tooltips should be shown for both.
- Click elements to show a notification including the element id. Same for using space or Enter on selected entries

https://user-images.githubusercontent.com/6959840/192825732-81f55877-6911-481e-98fb-01294e96fb5c.mp4

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #11869, opened 2022-11-15
Generic hover service support

msujew:
#### What it does

Fixes https://github.com/eclipse-theia/theia/issues/11594
Fixes https://github.com/eclipse-theia/theia/issues/10867
Also addresses a comment I've left in another PR https://github.com/eclipse-theia/theia/pull/11083#discussion_r873871011.

Uses the `StatusBarHoverManager` (now `HoverService`) as a base for building a generic hover service for markdown/string/html display as hovers/tooltips. This retires the `TooltipService` (introduced in https://github.com/eclipse-theia/theia/pull/10108), which was the source of the error messages in https://github.com/eclipse-theia/theia/issues/11594.

One feature that is missing here is that the hover could appear at the cursor position, right now it's bound to the position of the target element. I plan to add that in a separate PR later on.

#### How to test

1. Confirm that hovers still appear correctly on status bars (see also https://github.com/eclipse-theia/theia/pull/11083)
2. Confirm that hovers appear correctly on icons of widgets on the side bar panels (navigator, debug, extension, preferences, etc.)
3. Confirm that hovers appear correctly on tree nodes (see also https://github.com/eclipse-theia/theia/pull/11708)
4. Confirm that hovers appear correctly in the vsx extension view (see also https://github.com/eclipse-theia/theia/pull/10108)

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (commented):
I can confirm that the tooltips work really well with the new hover service :+1: 

I might actually open an issue in vscode for them to standardize using these tooltips everywhere for a more consistent look and feel (ex: explorer file paths, tree-view hover, toolbar item hover).

One improvement would be to adjust the `hr` rule for the `vsx-extensions` view as the horizontal rule is not styled similarly to vscode since we have a padding on each side:

https://user-images.githubusercontent.com/40359487/202214506-6204f5c0-134f-4ca8-8a75-ca89e3ffafb6.mp4

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
```diff
@@ -409,8 +409,8 @@ export class TreeViewWidget extends TreeViewWelcomeWidget {
     @inject(ContextKeyService)
     protected readonly contextKeyService: ContextKeyService;
 
-    @inject(TooltipService)
-    protected readonly tooltipService: TooltipService;
+    @inject(HoverService)
```
colin-grant-work, line 412: I believe that with the use of the HoverService here, we are no longer making use of the markdownIt field of this widget, and it could likely be at least deprecated, or more probably removed?

msujew: I believe we can safely remove it, good catch 👍

vince-fugnitto (approved):
I confirmed that my previous comment about the styling was addressed.

packages/core/src/browser/hover-service.ts
```diff
@@ -0,0 +1,189 @@
...
+            this.cancelHover();
+            this.pendingTimeout = disposableTimeout(() => this.renderHover(request), this.getHoverDelay());
+        }
+    }
+
+    protected getHoverDelay(): number {
+        return Date.now() - this.lastHidHover < 200
+            ? 0
+            : this.preferences.get('workbench.hover.delay', isOSX ? 1500 : 500);
+    }
```
colin-grant-work, line 100: @msujew, I notice that in Theia with this change, if _any_ hover has recently been rendered, then any other hover, of the same or different kind, will be rendered immediately, but in VSCode, it appears to differ by source. So e.g. if you hover over an extension tree node, get the hover, then hover over a  sidebar tabbar icon, the tabbar icon's title hover will appear immediately, but in VSCode, it will only hover after a delay.

packages/core/src/browser/frontend-application-module.ts
```diff
@@ -135,6 +135,7 @@ import { bindStatusBar } from './status-bar';
 import { MarkdownRenderer, MarkdownRendererFactory, MarkdownRendererImpl } from './markdown-rendering/markdown-renderer';
 import { StylingParticipant, StylingService } from './styling-service';
 import { bindCommonStylingParticipants } from './common-styling-participants';
+import { HoverService } from './hover-service';
```
Perea17, line 138: _

Perea17 (approved):

colin-grant-work:
@msujew, I was looking at the asynchronous tooltip content provision part of https://github.com/eclipse-theia/theia/issues/16280 and came across code for the old `TooltipService` from https://github.com/eclipse-theia/theia/pull/10108. In my view, this PR rendered that service obsolete, and it's no longer used by this repo - we just attach it to the DOM and never use it again. Should we perhaps delete / deprecate that service in favor of this?

msujew:
@colin-grant-work Yeah, makes sense.

PR #11888, opened 2022-11-19
Use environment-safe platform check in preferences

thegecko:

#### What it does

The PR: https://github.com/eclipse-theia/theia/pull/11393

Introduced a dependency on the node `os` module in common plugin source. This had the effect that the preferences system doesn't load for web-based extensions.

This PR fixes the issue by using the environment-safe checks instead.

#### How to test

Load a web extension which accesses user configuration and check it is possible.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/preference-registry.ts
+++ b/packages/plugin-ext/src/plugin/preference-registry.ts
@@ -17,13 +17,13 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
 import { Emitter, Event } from '@theia/core/lib/common/event';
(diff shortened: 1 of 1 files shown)


msujew (approved):
Looks reasonable to me 👍

I'm wondering whether it makes sense to grab the backend OS type instead since we are dealing with URIs. I saw that we never set the backend OS type in the web-based plugin host, so it will always be populated with the frontend OS type. We might want to address this in a follow-up PR.

colin-grant-work:
Apologies for introducing the bad dependency, and thanks for the fix!

Issue #10877, opened 2022-03-11
Reduce usage of `ContributionProvider`s in favor of Registries


colin-grant-work:
### Feature Description:

At the moment, we have numerous contribution points and contribution providers that give access to contributions bound to a particular identifier (e.g. `LabelProviderContribution`, `CommandContribution`, `MenuContribution`, `FrontendApplicationContribution`). We also have numerous registries that allow dynamic collection of implementations providing a particular service. The distinction has a number of consequences, the most important of which is that anything that is registered with a contribution provided must be known at build time, is typically instantiated at application startup, and is never disposed of. That makes contribution points less open to extension and enforces a less flexible lifecycle on their contributions, and so, where possible, we should favor registries, and it may be sensible to gradually migrate existing contribution points to registries.

A concrete example of the drawbacks of contribution points is the `LabelProvider` system and `TreeWidgets`. By default, `TreeWidgets` use the `LabelProvider` extensively, and that means that the correct way to change the way e.g. the caption of a novel `TreeWidget` is rendered is to contribute a `LabelProviderContribution` with an appropriate `getName` implementation. However, that `LabelProviderContribution` can't follow the same lifecycle as the widget: it will be instantiated at startup, and will never be destroyed even when the widget is closed. That means that even if the user never opens the widget and the `LabelProviderContribution` never provides a single label, we will have instantiated it and considered it as a candidate label provider for every label in the application. There are of course workarounds for particular situations: `TreeWidget`s could be refactored to consider a local `LabelProviderContribution` and only fall back to the global `LabelProvider` only if the local contribution fails. Nevertheless, registries are likely to be preferred where practical.

colin-grant-work:
Another point against `ContributionProviders`: in the current code, they are bound in the following way:

https://github.com/eclipse-theia/theia/blob/06b848e7d4f60ce4209fc17079fc40cab24312b5/packages/core/src/common/contribution-provider.ts#L77-L82

It appears that the dynamic binding means that they retain a reference to the first container in which they are instantiated. As a consequence, references to instances of identifiers bound in that container may be retained because the contribution `ContributionProvider` is not released.

PR #11889, opened 2022-11-19
Fix treeview item icon

thegecko:

#### What it does

The [label-provider](https://github.com/eclipse-theia/theia/blob/master/packages/core/src/browser/label-provider.ts#L369) class seems to iterate contributions for specified functionality until something is returned.

Both the [icon-theme-service](https://github.com/eclipse-theia/theia/blob/master/packages/core/src/browser/icon-theme-service.ts#L73) and [plugin-tree-view-node-label-provider](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/main/browser/view/plugin-tree-view-node-label-provider.ts#L42) contribution offer a `getIcon()` function.

Multiple contributions offering features seems to be mitigated by checking the return value is not `undefined`, however `icon-theme-service.getIcon()` returns an empty string instead and seems to get called first. This means all treeview icons are always empty.

This PR adds the additional check for an empty string which fixes the issue and the correct contribution is called, but I can't help feel this architecture is a messy approach and needs some further attention.

#### How to test

Load an extension with a treeview which has items with icons.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/label-provider.ts
+++ b/packages/core/src/browser/label-provider.ts
@@ -369,7 +369,7 @@ export class LabelProvider implements FrontendApplicationContribution {
     protected handleRequest(element: object, method: keyof Omit<LabelProviderContribution, 'canHandle' | 'onDidChange' | 'affects'>): string | undefined {
         for (const contribution of this.findContribution(element, method)) {
             const value = contribution[method]?.(element);
-            if (value !== undefined) {
+            if (!!value) {
                 return value;
             }
         }


packages/core/src/browser/label-provider.ts
```diff
@@ -369,7 +369,7 @@ export class LabelProvider implements FrontendApplicationContribution {
     protected handleRequest(element: object, method: keyof Omit<LabelProviderContribution, 'canHandle' | 'onDidChange' | 'affects'>): string | undefined {
         for (const contribution of this.findContribution(element, method)) {
             const value = contribution[method]?.(element);
-            if (value !== undefined) {
+            if (value !== undefined && value !== '') {
```
colin-grant-work, line 372: Shall we just go to `!value` and exclude `null`, as well?

thegecko: Yup, done

colin-grant-work (approved):
Looks good to me; it is correct to treat an empty string as a non-return in these cases.

vince-fugnitto:
@thegecko @colin-grant-work I think we might need to revert the changes for now, I noticed a regression (and possibly others) where the file-icon themes are not properly applied (ex: the theme is set to `none`):


_before_:

icons-before

_after_:

themes-after

thegecko:
Recommend we investigate a proper fix rather than swapping one issue for another.

thegecko:
I'll investigate, it looks like the file icon theme is being applied to all treeviews, not just the explorer.

vince-fugnitto:
@thegecko what was the initial pull-request trying to achieve? I believe that `''` was a valid icon in terms that we did not want an icon to be displayed (perhaps we can use something like `'none'`. Anyone that called `this.labelProvider.getIcon()` would be returned an icon for the contribution that can best handle the element.

colin-grant-work:
The problem (if it is one) is the entire system of [`ContributionProvider`s](https://github.com/eclipse-theia/theia/issues/10877) and unranked contributions. Anything we try to do here will be ad hoc, because there's no general way to order unordered contributions to make sure that we don't bail before getting the return we really want. Arguably the correct approach downstream is to inject providers you're interested in directly into the label provider and then check them in the order you want them checked before falling back to the default implementation, but that doesn't fix the fact that the default implementation implicitly relies on a specific ordering and undocumented sentinel values.

vince-fugnitto:
@thegecko if that's the case then we should revert, tree-views should respect the file-icon theme, and in this case it was set to `none`. It should be the same behavior present in vscode as well.

vince-fugnitto:
@thegecko what is the behavior you noticed in vscode when setting the theme to `none`?
I can't really test your extension, or take a look the sources.

marcdumais-work:
@thegecko could you provide an example FOSS vscode extension that can be used to test this PR?

According the the project's [PR rules and expectations](https://github.com/eclipse-theia/theia/blob/master/doc/pull-requests.md#reverting), regressions introduced in a PR need to be addressed within 2 days or the PR can be reverted. Ideally we come to an agreement before tomorrow's release or risk delaying it.

Reading the comments above, it seems that a more generic fix could be ideal, and that in the meantime your product could be customized to achieve the result you want (we can assist if you want to explore that path).

thegecko:
Without this patch, the default 'none' file tree theme doesn't show icons on any treeview:

Screenshot 2022-11-23 at 16 36 56

Switching the file tree theme then shows the treeview icons:

Screenshot 2022-11-23 at 16 37 11

While I agree this patch isn't the correct fix, I also believe we should only be affecting the file explorer treeview with the theme changes.

> I can't really test your extension

To recreate this, you can download the [device-manager vsix](https://open-vsx.org/api/arm/device-manager/1.2.5/file/Arm.device-manager-1.2.5.vsix) and put it in the `plugins` folder in theia and start the browser app.

You may need to modify the `Extensions > Device Manager > Manage All Devices` setting to show more USB devices.

> could you provide an example FOSS vscode extension that can be used to test this PR?

The extension above should be enough, it simply uses the `vscode.ThemeIcon` to set a treeview icon:

```typescript
    public getTreeItem(element: Node): vscode.TreeItem | Promise<vscode.TreeItem> {
        return {
            label: element.name,
            iconPath: new vscode.ThemeIcon('pass-filled')
        };
    }
```

> your product could be customized to achieve the result you want...

It would be quicker to implement a fix in Theia IMO

colin-grant-work:
To supplement, here's the extension in VSCode with `none` theme set, with an icon still showing in this tree view:

image

colin-grant-work:
The problem here seems to be that we don't respect icons on tree view elements at all:

https://github.com/eclipse-theia/theia/blob/909f4106e8c15c5c2c320401da4f48f8c6080734/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx#L432-L445

where `toNodeIcon` is the default implementation from the TreeWidget:

https://github.com/eclipse-theia/theia/blob/909f4106e8c15c5c2c320401da4f48f8c6080734/packages/core/src/browser/tree/tree-widget.tsx#L1411-L1413

colin-grant-work:
@thegecko, I've proposed an alternative solution in #11914. What do you think of that?

PR #11928, opened 2022-11-28
If evaluate request fails, ignore variable

colin-grant-work:
#### What it does

This PR addresses a problem where failure to evaluate a variable, for example because that variable has been optimized out, causes the entire sequence of memory retrieval to fail.

#### How to test

... to be continued

#### Review checklist

#### Reminder for reviewers


paul-marechal (approved):
I don't have a way to test this code but the logic is reasonable.

PR #11887, opened 2022-11-19
Implement VS Code extensions namespace

thegecko:
Signed-off-by: thegecko <rob.moran@arm.com>

#### What it does

This PR implements the extensions namespace in the VS Code API which is currently marked as not implemented in [the comparison coverage report](https://eclipse-theia.github.io/vscode-theia-comparator/status.html).

The functionality was already implemented, but exposed under the `plugins` export.

#### How to test

You can test the extensions `onDidChange` event by hooking into it in a VS Code extension.
You can test the `getExtension` function by calling it from a VS Code extension. You can then activate this extension and access the functionality on it using this pattern:

https://github.com/Marus/cortex-debug-db-stm32f4/blob/master/src/extension.ts#L15

Note: access to plugins is currently only possible within the same extensionHost (this matches VS Code functionality)

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #11931, opened 2022-11-29
Remove plugins namespace and extension 'hack'

thegecko:
#### What it does

As mentioned in https://github.com/eclipse-theia/theia/pull/11887#issuecomment-1330743397 there is a `hacked in` extensions namespace for the node extension host. Thi si snow surplus to requirements so can be removed.

I also took the liberty of removing the plugins namespace as this doesn't align with the VS Code API and mimics the extensions namespace. 
However, this can potentially break the Theia API, so would be happy to revert it depending on what people think.

#### How to test

Hopefully nothing should have changed! Testing is the same as #11887 

#### Review checklist

#### Reminder for reviewers


vince-fugnitto:
@thegecko I don't have any strong opinion to keep the old way, I believe it was originally implemented by RedHat and they required it.

thegecko:
Thanks @vince-fugnitto , perhaps someone from RedHat can confirm if this is OK?

packages/plugin-ext/src/plugin/plugin-context.ts
```diff
@@ -991,7 +973,6 @@ export function createAPIFactory(
             env,
             extensions,
             languages,
-            plugins,
```
msujew, line 994: Question: Doesn't this change result in all extensions which use `@theia/plugin` and the respective `plugins` namespace to break?

thegecko: Potentially. The same functionality is possible by using the `extensions` export instead. I seem to remember there is a move to deprecate the Theia API and focus on the VS Code one, but may be wrong.

Options are to include this breaking change and continue tidying towards the VS Code API only or keep this here, perhaps marking it as deprecated?

msujew: I would keep it for now, and deprecate the whole `@theia/plugin` API in one swoop later in a separate PR. Removing just one part of it sounds like inviting trouble. There might be no one actually using it, but I like to be [better safe than sorry](https://xkcd.com/1172/).

thegecko: Have now removed this removal.

msujew (approved):
Alright, seems to work well 👍 There's just one more thing that we can remove, see below.

packages/plugin-ext-vscode/src/node/plugin-vscode-init.ts
```diff
@@ -47,19 +47,6 @@ export const doInitialization: BackendInitializationFn = (apiFactory: PluginAPIF
 function createVSCodeAPI(apiFactory: PluginAPIFactory, plugin: Plugin): typeof theia {
     const vscode = Object.assign(apiFactory(plugin), { ExtensionKind });
```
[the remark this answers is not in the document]
thegecko: Have removed this along with the redundant interface

Issue #11950, opened 2022-12-05
It seems Theia can't show debug evaluate errors, because of incorrect DAP error handling


EvilBeaver:
### Bug Description:

When trying to evaluate expression in debugger, debug adapter can return success: false and error description. This is common thing that expressions can be misprinted and contain errors.

But, it seems Theia doesn't honor success flag of response https://github.com/eclipse-theia/theia/blob/master/packages/debug/src/browser/console/debug-console-items.tsx#L318 and shows error UI only in case of exception, which is not thrown in case of evaluation error.

### Steps to Reproduce:

1. Start debug session of some language and stop it on breakpoint
2. Print something wrong and incorrect in "watch" widget
3. Error is not shown, instead we see "not available" message

Expected, that real error is shown in an "error" manner, with Severity.Error

![image](https://user-images.githubusercontent.com/2657561/205612613-bc9295d6-db82-44db-8e10-9837c94343d0.png)

Reason: Theia tries to show a successful evaluation resultm which is "not available" because debug adapter returned error, not a result.

### Additional Information

- Operating System: any
- Theia Version: Version 1.31.1 (Beta)

PR #11953, opened 2022-12-05
debug: fix `watch` expression errors

vince-fugnitto:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/11950.

The pull-request updates handling for `watch` expressions to respect the error message rather than hardcode one.
The changes also align the styling with VS Code.

watch-fix

#### How to test

1. start the application with theia as a workspace
2. open a `*.spec.ts` file
3. set a breakpoint in the file and trigger the `run mocha tests` debug configuration
4. attempt to create `watch` expressions with variables that are known and unknown
5. known expressions should work
6. unknown expressions should display a proper error message

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/view/debug-watch-expression.tsx
+++ b/packages/debug/src/browser/view/debug-watch-expression.tsx
@@ -24,6 +24,7 @@ import { nls } from '@theia/core';
 export class DebugWatchExpression extends ExpressionItem {
 
     readonly id: number;
+    protected isError: boolean;
 
     constructor(protected readonly options: {
         id: number,
@@ -40,9 +41,12 @@ export class DebugWatchExpression extends ExpressionItem {
         await super.evaluate('watch');
     }
 
-    protected override setResult(body?: DebugProtocol.EvaluateResponse['body']): void {
-        // overridden to ignore error
-        super.setResult(body);
+    protected override setResult(body?: DebugProtocol.EvaluateResponse['body'], error?: string): void {
+        if (!this.options.session()) {
+            return;
+        }
+        super.setResult(body, error);
+        this.isError = !!error;
         this.options.onDidChange();
     }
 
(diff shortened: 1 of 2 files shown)


packages/debug/src/browser/view/debug-watch-expression.tsx
```diff
@@ -37,16 +38,19 @@ export class DebugWatchExpression extends ExpressionItem {
         await super.evaluate('watch');
     }
 
-    protected override setResult(body?: DebugProtocol.EvaluateResponse['body']): void {
-        // overridden to ignore error
-        super.setResult(body);
+    protected override setResult(body?: DebugProtocol.EvaluateResponse['body'], error?: string): void {
+        if (!this.options.session()) {
+            return;
+        }
```
msujew, line 44: I've noticed that vscode yields this (generic but fitting) result when no session is active:

image

Does that make sense for us as well? Or is that out of scope?

vince-fugnitto: @msujew I believe we do too with these changes? I did notice that we do not update `watch expressions` properly which I wanted to file a bug for, basically if a new debug session is execute the existing `watch expressions` are not updated.

vince-fugnitto: Right, I think it's the overall issue on master where our watch expressions do not seem to be properly updated.

For example:

image

If I do not have an active debug session and set the `value` watch expression, hitting the breakpoint when I do start a session does not update the expression like it does in vscode.

image

msujew: I see, alright. Then it's out of scope 👍

msujew (approved):
Alright, I can confirm that this change fixes the issue at hand 👍

PR #11956, opened 2022-12-05
Allow to remove watch expressions one by one

msujew:
#### What it does

I noticed in https://github.com/eclipse-theia/theia/pull/11953 that watch expressions can only be removed in bulk. This change aligns the UX to vscode, where users are able to remove watched expressions one by one.

Also fixes an existing issue where watched expressions which evaluate to an object result in an overflow:

image

#### How to test

1. Start debugging a Test in Theia
2. Add multiple watched expressions
3. Confirm that they can be removed one by one

image

I believe the `theia-tree-element-node` css class is reused in the debug console to display evaluated expressions. Assert that no regressions appeared due to my changes.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/style/tree.css
+++ b/packages/core/src/browser/style/tree.css
@@ -171,10 +171,6 @@
     opacity: var(--theia-mod-disabled-opacity);
 }
 
-.theia-tree-element-node {
-    width: 100%
-}
-
 .theia-tree-node-indent {
     position: absolute;
     height: var(--theia-content-line-height);
--- a/packages/core/src/browser/source-tree/source-tree-widget.tsx
+++ b/packages/core/src/browser/source-tree/source-tree-widget.tsx
@@ -17,7 +17,7 @@
 import * as React from 'react';
 import { injectable, postConstruct, interfaces, Container } from 'inversify';
 import { DisposableCollection } from '../../common/disposable';
-import { TreeWidget, TreeNode, createTreeContainer, TreeProps, TreeModel } from '../tree';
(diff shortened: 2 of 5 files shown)


vince-fugnitto (commented):
@msujew I believe we might want to improve the styling by:
- only showing the remove expression icon on hover (similarly to vscode)
- aligning the icon with the row

image

msujew:
@vince-fugnitto Thanks, didn't notice these two differences. Funnily enough, vscode's button also isn't correctly vertically centered with the row :D

image

packages/core/src/browser/style/tree.css
```diff
@@ -171,10 +171,6 @@
     opacity: var(--theia-mod-disabled-opacity);
 }
 
-.theia-tree-element-node {
-    width: 100%
-}
```
vince-fugnitto, line 176: I'm not sure if there is any regression with trees with this change since it is affects all trees but I haven't encountered any.

vince-fugnitto (approved):
The changes look good to me, and I confirm that we can now discard individual watch expressions :+1:

packages/core/src/browser/style/tree.css
msujew: I believe this class in particular is exclusively used in the [`SourceTreeWidget`](https://github.com/eclipse-theia/theia/blob/1dcddc630ab0beb5d66ec885d207d3e6529fb692/packages/core/src/browser/source-tree/source-tree-widget.tsx#L89), which is only used by the expression watching stuff (i.e. the `watch` debug tab and the debug console). In both cases it didn't do anything other than accidentally overflowing the tree node (see picture above).

PR #11962, opened 2022-12-06
Use client side rate limiting for plugin download

msujew:
#### What it does

Related to https://github.com/eclipse-theia/theia/pull/11860. The publicly managed open-vsx instance is using a rate limiting of 15 requests per second (yielding a `429` status code on too many requests). This change introduces a rate limiter that limits the amount of requests that we send per second to 15, while also handling the `429` status code a bit better.

It also reinstates the parallel-by-default behavior of the plugin download command, which was changed in https://github.com/eclipse-theia/theia/pull/11860. Additionally provides a CLI argument to change the rate limit.

#### How to test

0. The CI should not fail due to 429 errors even though parallel plugin downloads are enabled
1. Delete your `plugins` folder.
2. Run `yarn download:plugins --rate-limit=x` with different values for `x`
3. Assert that the rate limit is applied and the command doesn't return with the 429 http error

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/cli/package.json
+++ b/dev-packages/cli/package.json
@@ -44,6 +44,7 @@
     "chalk": "4.0.0",
     "decompress": "^4.2.1",
     "glob": "^8.0.3",
+    "limiter": "^2.1.0",
     "log-update": "^4.0.0",
     "mocha": "^10.1.0",
     "puppeteer": "^2.0.0",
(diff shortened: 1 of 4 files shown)


paul-marechal (approved):
I tried with 15 and 1 as a rate limit and it worked.

Thanks for making it fast again!

Issue #11509, opened 2022-08-03
[vscode] Support querying whether file system and file stats are readonly


planger:
The VS Code API can be used to query whether the file system or single file stats are readonly:

 - [ ] [`FileStat.permissions`](https://code.visualstudio.com/api/references/vscode-api#FileStat), which requires enum type [`FilePermission`](https://code.visualstudio.com/api/references/vscode-api#FilePermission)
 - [ ] [`FileSystem.isWritableFileSystem(scheme: string)`](https://code.visualstudio.com/api/references/vscode-api#FileSystem)

JonasHelming:
@msujew volunteered to take this one

msujew:
Note that this also requires support for https://github.com/eclipse-theia/theia/issues/199, as we currently have no way of actually defining that a file/editor is readonly.

Issue #199, opened 2017-06-13
Handle readonly files


akosyakov:
- [ ] `FileStat` should have access flags. There should be utility methods to check flags on read, write and execute access
- [ ] `FileTreeModel` should mark file stat nodes without read access as invisible
- [ ] `Resource` should have an explicit optional `isReadonly` property with default based on the presence of `Resource.save` method. There should be a utility function to check it.
- [ ] `FileResource` should use `FileStat` flags to implement `isReadonly` property
- [ ] editor providers should use `Resource.isReadonly(resource)` to mark editors readonly or not

hexa00:
I disagree that some files should be invisible you have the right to see them... but an error should be thrown when you try to open ?

akosyakov:
I meant if the user doesn't have rights to see them he should not see them.

hexa00:
Then it's the directory that does not have read access and thus you won't be able to get to the file at all...

Otherwise I can see a no permission file with ls without any problems:
----------  1 x x        0 Jun 13 08:22 testfi

akosyakov:
PRs are welcomed.

epatpol:
Would you rather have the fs.modes for the filestat (i.e bitmasks) or easier API like `Filestat.isReadable: boolean`, `isWriteable`, `isExecutable` and so on? Like so:

```
export interface FileStat {

    /**
     * The uri of the file.
     */
    uri: string;

    /**
     * File access
     */
    isReadable: boolean;
    isWriteable: boolean;
    isExecutable: boolean;

    /**
     * The last modification of this file.
     */
    lastModification: number;

    /**
     * The resource is a directory. Iff {{true}}
     * {{encoding}} has no meaning.
     */
    isDirectory: boolean;

    /**
     * The children of the file stat.
     * If it is undefined and isDirectory is true, then this file stat is unresolved.
     */
    children?: FileStat[];

    /**
     * The size of the file if known.
     */
    size?: number;
}
```

akosyakov:
@epatpol Could you check how VS Code handle it? They don't seem to expose such apis, maybe we are overthinking it.

aaqaishtyaq:
@epatpol @akosyakov Hey is there any way currently to now allow editing of readOnly files.

dpetroff:
In theia 1.36, I set some files to read only on the file system with chmod, and all the resource, filestat, etc. properties all still say that it's not read-only.

[...]

PR #11965, opened 2022-12-07
Handle file system permissions

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/11509
Closes https://github.com/eclipse-theia/theia/issues/199

This was more work than initially expected. It seems like we didn't handle readonly editors at all previously (see https://github.com/eclipse-theia/theia/issues/199). So this change not only adds the file system permissions extension API, but also enables us to pass permission information all the way to the monaco model and phosphor tabbars.

#### How to test

1. Download and install [this test extension](https://github.com/eclipse-theia/theia/files/10171750/fs-readonly-0.0.1.zip)
2. Run the 4 contributed `Open file from file system` commands. They should all result in read only editors except for the one that opens the `mutable-in-mutable` file.
3. Readonly editors should be marked as such (a lock icon) and should behave that way (i.e. the document is not editable).
4. Run the `Check file system isWritableFileSystem` command. Input different URI schemes which then displays a notification containing the result of the `isWritableFileSystem` call. The extension contributes the `readonly` and `mutable` file system schemes, which should return `false`/`true` respectively. Calls to `file` should also return `true`. Everything else should return `undefined`, unless contributed by an extension or plugin.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/tab-bars.ts
+++ b/packages/core/src/browser/shell/tab-bars.ts
@@ -31,7 +31,7 @@ import { IconThemeService } from '../icon-theme-service';
 import { BreadcrumbsRenderer, BreadcrumbsRendererFactory } from '../breadcrumbs/breadcrumbs-renderer';
 import { NavigatableWidget } from '../navigatable-types';
 import { IDragEvent } from '@phosphor/dragdrop';
-import { PINNED_CLASS } from '../widgets/widget';
+import { LOCKED_CLASS, PINNED_CLASS } from '../widgets/widget';
 import { CorePreferences } from '../core-preferences';
 import { HoverService } from '../hover-service';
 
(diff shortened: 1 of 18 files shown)


vince-fugnitto (approved):
The changes look good to me and work well :+1: 

- [x] confirmed the commands contributed by the plugin work as expected
- [x] confirmed that readonly editors display the `lock` decoration as a caption
- [x] confirmed that readonly editors are readonly from a monaco standpoint (with the default message as a user attempts to edit the file)

JonasHelming:
@colin-grant-work Do you want to have a look or can we merge this?

colin-grant-work:
It looks reasonable to me, though I haven't tested it. I'm fine with merging.

Issue #11730, opened 2022-09-30
builtin-git: add support for missing menu paths


vince-fugnitto:
#### Bug Description:

There are issues when using the [vscode.git](https://open-vsx.org/extension/vscode/git) and [vscode.git-ui](https://open-vsx.org/extension/vscode/git-ui) builtins where it is unable to register toolbar items for the `scm` view due to missing menu paths:

- `scm/change/title`
- `scm/sourceControl`

#### Steps to Reproduce:

1. remove `vscode.git` and `vscode.git-ui` from the exclusion list

https://github.com/eclipse-theia/theia/blob/978a3a0d504f5f54425175d88afee2cdf1617d02/package.json#L119-L120

3. remove `@theia/git` and `@theia/scm-extra` from the example applications
4. rebuild, and confirm that the `download:script` pulls the git builtins
5. notice the errors related to not being able to set the toolbar items

Example:

```
2022-09-30T13:10:45.625Z root WARN Failed to register a menu item for plugin vscode.git contributed to scm/change/title {
  command: 'git.revertChange',
  submenu: undefined,
  alt: undefined,
  group: undefined,
  when: 'config.git.enabled && !git.missing && originalResourceScheme == git'
} Error: Could not find submenu with id scm/change/title
```

```
2022-09-30T13:10:45.625Z root WARN Failed to register a menu item for plugin vscode.git contributed to scm/sourceControl {
  command: 'git.close',
  submenu: undefined,
  alt: undefined,
  group: 'navigation',
  when: 'scmProvider == git'
} Error: Could not find submenu with id scm/sourceControl
```

safisa:
Hi @vince-fugnitto, @colin-grant-work 
I have a similar or maybe the same issue. I am getting this error when trying to press on the 3 dots toolbar menu on the git/source control:

```
Uncaught Error: 'git.tags' is not a menu group.
    at MenuModelRegistry.findSubMenu (menu-model-registry.ts:250:19)
    at MenuModelRegistry.findGroup (menu-model-registry.ts:235:32)
    at MenuModelRegistry.getMenuNode (menu-model-registry.ts:112:25)
    at MenuModelRegistry.linkSubmenu (menu-model-registry.ts:170:28)
    at TabBarToolbar.renderMoreContextMenu (tab-bar-toolbar.tsx:183:53)
    at TabBarToolbar.showMoreContextMenu (tab-bar-toolbar.tsx:166:14)
[...]

PR #12007, opened 2022-12-19
Adding timestamp to comment. fixes #11702

### LINKED ISSUE #11702: [vscode] Support optional property Comment#timestamp
see https://code.visualstudio.com/api/references/vscode-api#Comment.

rschnekenbu:
#### What it does

_Add Comment#timestamp optional property_
This adds the timeStamp optional property from Comment to improve vscode api coverage. It also displays the timestamp of each comments in the Comment Thread widget

Fixes #11702 

#### How to test 
Note: there are currently some issues with menus, since theia 1.28.0. This is due to #11290, integrated in 1.28.0. Prior to this change, the comment sample works on theia 1.27.0. So to test this change, I cherry-picked my commit on top of 1.27.0 tag. Please see also #11730 for this menu issues.
1. Install forked extension [vscode-comment-api-example-0.0.1.zip](https://github.com/eclipse-theia/theia/files/10267130/vscode-comment-api-example-0.0.1.zip)
 vsix file
2. Create a comment using the side bar while comparing 2 test files. 
addCommentTimeStamp

4. The timestamp should be displayed along the user name.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
+++ b/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
@@ -660,6 +660,8 @@ export interface Comment {
     readonly contextValue?: string;
     readonly label?: string;
     readonly mode?: CommentMode;
+    /** Timestamp serialized as ISO date string via Date.prototype.toISOString */
+    readonly timestamp?: string;
 }
 
 export enum CommentThreadCollapsibleState {
--- a/packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
+++ b/packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
@@ -479,6 +479,7 @@ export class ReviewComment<P extends ReviewComment.Props = ReviewComment.Props> 
             <div className={'review-comment-contents'}>
                 <div className={'comment-title monaco-mouse-cursor-text'}>
                     <strong className={'author'}>{comment.userName}</strong>
+                    <small className={'timestamp'}>{this.localeDate(comment.timestamp)}</small>
                     <span className={'isPending'}>{comment.label}</span>
                     <div className={'theia-comments-inline-actions-container'}>
                         <div className={'theia-comments-inline-actions'} role={'toolbar'}>
@@ -497,6 +498,16 @@ export class ReviewComment<P extends ReviewComment.Props = ReviewComment.Props> 
                     commands={commands} />
             </div>
         </div>;
+    }
+    protected localeDate(timestamp: string | undefined): string {
+        if (timestamp === undefined) {
+            return '';
+        }
+        const date = new Date(timestamp);
+        if (!isNaN(date.getTime())) {
+            return date.toLocaleString();
+        }
+        return '';
     }
 }
 
--- a/packages/plugin/src/theia.d.ts
+++ b/packages/plugin/src/theia.d.ts
@@ -12084,6 +12084,11 @@ export module '@theia/plugin' {
          * Label will be rendered next to authorName if exists.
          */
         label?: string;
+
+        /**
+         * Optional timestamp.
+         */
+        timestamp?: Date;
     }
 
     /**
--- a/packages/plugin-ext/src/main/browser/style/comments.css
+++ b/packages/plugin-ext/src/main/browser/style/comments.css
@@ -149,6 +149,12 @@
     line-height: var(--theia-content-line-height);
 }
 
+.monaco-editor .review-widget .body .review-comment .review-comment-contents .timestamp {
+    line-height: var(--theia-content-line-height);
+    margin: 0 5px 0 5px;
+    padding: 0 2px 0 2px;
+}
+
 .monaco-editor .review-widget .body .review-comment .review-comment-contents .isPending {
     margin: 0 5px 0 5px;
     padding: 0 2px 0 2px;
--- a/packages/plugin-ext/src/plugin/comments.ts
+++ b/packages/plugin-ext/src/plugin/comments.ts
@@ -491,6 +491,7 @@ function convertToModeComment(thread: ExtHostCommentThread, commentController: C
     }
 
     const iconPath = theiaComment.author && theiaComment.author.iconPath ? theiaComment.author.iconPath.toString() : undefined;
+    const date = theiaComment.timestamp ? theiaComment.timestamp.toISOString() : undefined;
 
     return {
         mode: theiaComment.mode,
@@ -500,6 +501,7 @@ function convertToModeComment(thread: ExtHostCommentThread, commentController: C
         userName: theiaComment.author.name,
         userIconPath: iconPath,
         label: theiaComment.label,
+        timestamp: date,
     };
 }
 


packages/plugin/src/theia.d.ts
```diff
@@ -12084,6 +12084,11 @@ export module '@theia/plugin' {
          * Label will be rendered next to authorName if exists.
          */
         label?: string;
+
+        /**
+         * Optional timestamp.
+         */
+        timeStamp?: Date;
```
kittaakos, line 12091: [VS Code uses `timestamp`](https://github.com/microsoft/vscode/blob/2c8b54bbbf0167e52ef4602a9ee14e1223fb0d7c/src/vscode-dts/vscode.d.ts#L15384). Can Theia use the same?

`timeStamp` -> `timestamp`?

packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
```diff
@@ -498,6 +499,16 @@ export class ReviewComment<P extends ReviewComment.Props = ReviewComment.Props>
             </div>
         </div>;
     }
+    localeDate(stamp: string | undefined): string {
+        if (stamp === undefined) {
+            return '';
+        }
+        const date = Date.parse(stamp);
+        if (date && !Number.isNaN(date)) {
+            return new Date(date).toLocaleString();
```
kittaakos, line 508: Hi, why is it need to create two `Date` instances for one string? Thanks!

packages/plugin-ext/src/common/plugin-api-rpc-model.ts
```diff
@@ -660,6 +660,8 @@ export interface Comment {
     readonly contextValue?: string;
     readonly label?: string;
     readonly mode?: CommentMode;
+    /** Timestamp serialized as ISO date string via Date.prototype.toISOString */
+    readonly timeStamp?: string;
```
vince-fugnitto, line 664: minor: `timeStamp` > `timestamp`

```suggestion
    readonly timestamp?: string;
```

packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
```diff
@@ -498,6 +499,16 @@ export class ReviewComment<P extends ReviewComment.Props = ReviewComment.Props>
             </div>
         </div>;
     }
+    localeDate(stamp: string | undefined): string {
```
vince-fugnitto, line 502: minor: no need to abbreviate

```suggestion
    localeDate(timestamp: string | undefined): string {
```

packages/plugin/src/theia.d.ts
vince-fugnitto: Right, @rschnekenbu if you run the [vscode-theia-comparator](https://github.com/eclipse-theia/vscode-theia-comparator) against this branch you'll see that `timestamp` is still marked as **unsupported**:

image

packages/plugin-ext/src/common/plugin-api-rpc-model.ts
rschnekenbu: This is solved with commit 376dbca. time**s**tamp is now used.

packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
rschnekenbu: I now use timestamp rather than stamp in commit 376dbca.

packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
rschnekenbu: Hi, thanks for your very fast review @kittaakos! 
I am not sure to understand your comment: I only create 1 Date instance line 508, result of line 506 is a number.

rschnekenbu:
@vince-fugnitto, with the new version, I get a green flag on the vscode-theia-comparator:
timestamp

Note: I also updated the comment vsix extension sample archive, which has the same typo issue.

packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
kittaakos: You're right, but you parse the string twice: `parse` and the `Date` constructor. Doing it once would be sufficient. What do you think?

From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date):

> When parsing date strings with the Date constructor (and Date.parse, they are equivalent)

```ts
parse(s: string | undefined): string {
  if (s) {
    const d = new Date(s);
    if (!Number.isNaN(d.getTime())) {
      return d.toLocaleString();
    }
  }
  return '';
}
```

packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
```diff
@@ -498,6 +499,16 @@ export class ReviewComment<P extends ReviewComment.Props = ReviewComment.Props>
             </div>
         </div>;
     }
+    localeDate(timestamp: string | undefined): string {
```
kittaakos, line 502: Does this method require `public` visibility, or can be reduced to `protected`?

rschnekenbu: Visibility reduced to protected, no need to expose.

packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
rschnekenbu: @kittaakos, thanks a lot for your explanation! I did simplify indeed the code with 5d615cd.

kittaakos: Thank you so much!

kittaakos: I appreciate your taking the time to update your PR after my comment. The code is correct, but I could not resist adding that the `date` check in the ` if (date && !isNaN(date.getTime())) {` condition is superfluous.

A constructor function [can return with `undefined`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/constructor):

> The constructor method may have a return value. While the base class may return anything from its constructor, the derived class must return an object or undefined, or a [TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) will be thrown.

But you will never get `undefined` after successfully calling it with `new` at runtime. See [here](https://github.com/microsoft/TypeScript/issues/29893#issuecomment-463409560). So you can eliminate the `date` check, which will always be `true`, and simplify the code if you wish.

I am happy to receive a correction if I am proven wrong on this. Thanks!

packages/plugin/src/theia.d.ts
rschnekenbu: the timestamp typo has been fixed, as all last comments

packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
rschnekenbu: Thanks @kittaakos, you are absolutely right! The code has been simplified in 6895086.

kittaakos (approved):
The code looks great. Thank you!

msujew (approved):
Looks good to me as well 👍

JonasHelming:
@vince-fugnitto Are you fine with merging this?

paul-marechal (approved):
LGTM

PR #8868, opened 2020-12-15
yarn upgrade: avoid vscode-jsonrpc@6

marechal-p:
#### What it does

Closes: #8862

`vscode-languageserver-protocol@3.16.x` depends on `vscode-jsonrpc@6`
which breaks our build. We expect to have `vscode-jsonrpc@5`.

Signed-off-by: Paul <paul.marechal@ericsson.com>

#### How to test

- Do `yarn upgrade`
- `yarn why vscode-jsonrpc` should not show a version different from `vscode-jsonrpc@5.x.x`

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #12012, opened 2022-12-20
Update vscode-languageserver-protocol version

tortmayr:
#### What it does
Updates the shared `vscode-language-server-protocol` dependency to the latest version. The previous version (3.15.x) is quite old and pulls in an outdated version of `vscode-jsonrpc` (5.x). This can cause conflicts with adopters that want to use a more recent version.

Previously this update was blocked due to an incompatibility with `vscode-ws-jsonrpc` (see https://github.com/eclipse-theia/theia/pull/8868). However, after the recent  RPC communication refactoring we no longer use `vscode-ws-jsonrpc` and can update without issues.

#### How to test

There is nothing to explicitly test. Everything should work as before.
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -98,7 +98,7 @@ export class SomeClass {
     - `react-dom` (from [`react-dom@^18.2.0`](https://www.npmjs.com/package/react-dom))
     - `react-dom/client` (from [`react-dom@^18.2.0`](https://www.npmjs.com/package/react-dom))
     - `react-virtuoso` (from [`react-virtuoso@^2.17.0`](https://www.npmjs.com/package/react-virtuoso))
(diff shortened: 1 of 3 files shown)


msujew (approved):
LGTM 👍 

@vince-fugnitto Should we include this in the upcoming release or wait?

vince-fugnitto:
> @vince-fugnitto Should we include this in the upcoming release or wait?

@msujew I would maybe wait, but what do you think?

msujew:
I don't really have an opinion on this, but also no idea how urgent this is from @tortmayr's perspective.

tortmayr:
@vince-fugnitto  @msujew  It's not urgent at all, no need to squeeze this into the upcoming release.

Issue #11991, opened 2022-12-13
Many debug configurations make it impossible to reach 'Add Configurations'


colin-grant-work:
### Bug Description:

If enough debug configurations are available to fill the screen, it is impossible to see the bottom 'Add Configuration' option. This is likely a general problem with the `ReactSelectComponent`

### Steps to Reproduce:

1. Open a workspace with many debug configurations.
2. Open the debug view and open the debug config dropdown.
3. Observe that you cannot see or scroll to the bottom of the list.

![image](https://user-images.githubusercontent.com/62660806/207459530-2cc17e19-561e-4828-8d90-a255508a0791.png)

### Additional Information

- Operating System: Windows 10, Electron
- Theia Version: ac7a9ff51afa71d26673a7a8f7968c8e9a3f51ff

PR #12038, opened 2023-01-02
added a max height and overflow to select-dropdown

jonah-iden:
auto to the select-component-dropdown,
so that the dropdown can't get bigger than the

#### What it does

Closes: #11991 .
Added a default "overflow: auto" and a max-height to the select-component dropdown. The max-height is calculated so that the dropdown is never outside of the visible area (clientRect).

#### How to test

Best way to test is probably adding a lot of debug configurations to the launch.json. Then in the debug window the configuration dropdown should on go to the bottom of the page and have a scrollbar. 
For the inverted state i testet it thorugh the preferences (Terminal › Integrated: Font Weight Bold is a good select-box for that) by making the browser very small. 
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/style/select-component.css
+++ b/packages/core/src/browser/style/select-component.css
@@ -47,6 +47,7 @@
     outline: var(--theia-focusBorder) solid 1px;
     outline-offset: -1px;
     user-select: none;
+    overflow: auto;
 }
 
 .theia-select-component-dropdown .theia-select-component-option {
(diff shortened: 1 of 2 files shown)


msujew (commented):
Hey @jonah-iden, thanks for the first time contribution.

Please make sure to sign the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php) with the same email that is associated with this commit for us to accept the changes.

msujew (commented):
When scrolling using the mouse wheel, the select component will automatically close due to [this code](https://github.com/eclipse-theia/theia/blob/8110ddab61c92c9bc7e8465f34a36c4d18fb281b/packages/core/src/browser/widgets/select-component.tsx#L130-L131). The `hide` function needs to be updated to detect whether the `target` element of the event is actually inside of the select component. If it is outside the component, it should close the dropdown and keep it open if the target is inside.

msujew (approved):
The changes look good to me, thanks 👍

PR #12046, opened 2023-01-06
focus out possible again for debug configuration

### LINKED ISSUE #11992: Workspace Selection QuickPick for Debug 'Add Configuration' ignores focus out
### Bug Description:

In #11444, [`ignoreFocusOut`](https://github.com/eclipse-theia/theia/blob/ac7a9ff51afa71d26673a7a8f7968c8e9a3f51ff/packages/debug/src/browser/debug-configuration-manager.ts#L391) was added to the options for the workspace selection quick pick for adding debug configurations. That produces a bad UX because no other quickpicks inside the application ignore focus out.

The code was added to addresses a bug that was present prior to that change: clicking 'Add Configuration' in the DebugConfigurationSelect would open and then immediately close the quick pick. That is a problem with the `ReactSelectComponent` - because it removes the rendered container, it triggers a blur event with `relatedTarget: null`. The problem should be addressed there rather than worked around in the `debug` package.

### Steps to Reproduce:

1. Open a multi-root workspace.
2. Open the debug view and use the dropdown to select 'Add Configuration'
3. See a quickpick.
4. Try to get out of the quick pick by focusing other elements.
5. Observe that the quick pick remains.
---
1. Remove the `ignoreFocusOut: true` from the options in the `DebugConfigurationManager`
2. Follow steps 1-2 above.
3. Observe that the quickpick appears and then immediately disappears.

### Additional Information

- Operating System: Windows 10, Electron
- Theia Version: ac7a9ff51afa71d26673a7a8f7968c8e9a3f51ff

jonah-iden:
#### What it does
fixes #11992

this makes the call to configurationManager.addConfiguration() execute after react has closed the dropdown. Through this it's possible to remove the prevous ignoreFocusOut fix and make it possible again to click outside of the quick select to cancel it.

#### How to test
1. have at least 2 roots in your workspace
2. go to the debug view
3. open debug select and "Add Configuration"
4. quick select for which root to choose should open (and not close instantly again)
5. click outside and see it cancels the action.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/view/debug-configuration-select.tsx
+++ b/packages/debug/src/browser/view/debug-configuration-select.tsx
@@ -103,7 +103,7 @@ export class DebugConfigurationSelect extends React.Component<DebugConfiguration
         if (!value) {
             return false;
         } else if (value === DebugConfigurationSelect.ADD_CONFIGURATION) {
-            this.manager.addConfiguration();
+            setTimeout(() => this.manager.addConfiguration());
         } else if (value.startsWith(DebugConfigurationSelect.PICK)) {
             const providerType = this.parsePickValue(value);
             this.selectDynamicConfigFromQuickPick(providerType);
--- a/packages/debug/src/browser/debug-configuration-manager.ts
+++ b/packages/debug/src/browser/debug-configuration-manager.ts
@@ -388,7 +388,6 @@ export class DebugConfigurationManager {
         }
         const root = await this.quickPickService.show(items, {
             placeholder: nls.localize('theia/debug/addConfigurationPlaceholder', 'Select workspace root to add configuration to'),
-            ignoreFocusOut: true
         });
         return root?.value;
     }


paul-marechal:
Why do I feel like this change replaces a hack with another hack?

Is there a React API to listen for the dropdown closing rather than using `setTimeout`?

jonah-iden:
yeah it really is a little bit hacky. I searched a little, but could't find anything on closing the dropdown and defocussing the selectComponent seems to also not fix the issue.

tsmaeder:
@martin-fleck-at we should decide if the "hack" is a step forward and either merge or close the PR accordingly.

packages/debug/src/browser/view/debug-configuration-select.tsx
```diff
@@ -103,7 +103,7 @@ export class DebugConfigurationSelect extends React.Component<DebugConfiguration
         if (!value) {
             return false;
         } else if (value === DebugConfigurationSelect.ADD_CONFIGURATION) {
-            this.manager.addConfiguration();
+            setTimeout(() => this.manager.addConfiguration());
```
martin-fleck-at, line 106: I agree that this is a bit 'hacky' and I tested `animationFrame().then(() => this.manager.addConfiguration());` which also seems to work but I don't see any additional benefit from using that instead. I also did not find any alternative solution that solves that issue but I also don't think there are going to be any side effects from this, so in my opinion this could be merged.

The only thing that would be nice is a short comment explaining why this particular call needs to be handled differently though in theory that could be derived from the Git history. @tsmaeder How do you usually handle these situations with regard to documentation?

martin-fleck-at (approved):

JonasHelming:
@msujew Do you want to merge this?

Issue #12022, opened 2022-12-22
[vscode] Support SnippetTextEdit


rschnekenbu:
see [SnippetTextEdit](https://code.visualstudio.com/api/references/vscode-api#SnippetTextEdit)
since vscode 1.72

see also https://code.visualstudio.com/updates/v1_72#_improved-workspace-editing-apis

jonah-iden:
Hi, I would like to try working on this task as a good, bit more complex, next step in getting to know the theia vscode plugin integration

PR #12047, opened 2023-01-06
Add support for `SnippetTextEdit` vscode API

jonah-iden:
#### What it does

Closes: #12022

as described [here](https://code.visualstudio.com/updates/v1_72#_improved-workspace-editing-apis) this PR implements SnippetTextEdits the same way they are implemented in vscode. This adds the api to set them on Workspaceedits and sends them as TextEdits with the insertAsSnippet flag to the monaco editor. 
Through the snippetController2 class those Edits can be applied to the current editor. 
Should  SnippetTextEdits be done on a file which editor is not currently active, they are transformed to just the text and applied as normal TextEdits. 

It sadly seems both in vscode and theia this snippetTextEdit/SnippetController2 only supports snippets with a single tabstop. So tabbing through seems to not work yet in both IDEs.

#### How to test
I built a tiny vscode extension which is downloadable from here
[snippet-text-edit-test-0.0.1.zip](https://github.com/eclipse-theia/theia/files/10361208/snippet-text-edit-test-0.0.1.zip)
or buildable from here https://github.com/jonah-iden/snippet-text-edit-test.

This just inserts a test snippetTextEdit at the start of the currently open file and a file called file2.json in the workspace root (this file needs to be created manually first).

you should see the snippet is inserted correctly at the start of the file and the cursor is moved to the tabstop and in file2.json its also inserted at the start of the file without moving the cursor.

#### Review checklist

#### Reminder for reviewers


msujew (commented):
The CI fails with the current changes.

packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -1279,6 +1279,36 @@ export class NotebookRange implements theia.NotebookRange {
 
 }
 
+export class SnippetTextEdit implements theia.SnippetTextEdit {
+    range: Range;
+    snippet: SnippetString;
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    static isSnippetTextEdit(thing: any): thing is SnippetTextEdit {
```
msujew, line 1287: Suggestion: We usually try to use `unknown` in place of `any` where appropriate. Even though they use `any` in the original vscode implementation...

You can use the `isObject` utility function to easily convert `thing` to an instance of `SnippetTextEdit`.

packages/monaco/src/browser/monaco-workspace.ts
```diff
@@ -364,4 +373,20 @@ export class MonacoWorkspace {
             }
         }
     }
+
+    private async performSnippetEdits(edits: MonacoResourceTextEdit[]): Promise<void> {
```
msujew, line 377: Minor: To keep it as simple as possible to override services, Theia never uses `private`, but keeps everything `protected`.

packages/monaco/src/browser/monaco-workspace.ts
```diff
@@ -288,7 +297,7 @@ export class MonacoWorkspace {
                 const uri = monaco.Uri.parse(key);
                 let eol: EndOfLineSequence | undefined;
                 const editOperations: monaco.editor.IIdentifiedSingleEditOperation[] = [];
-                const minimalEdits = await StandaloneServices.get(IEditorWorkerService).computeMoreMinimalEdits(uri, value.map(v => v.textEdit));
+                const minimalEdits = await StandaloneServices.get(IEditorWorkerService).computeMoreMinimalEdits(uri, value.map(this.transformSnippetStringToInsertText));
```
msujew, line 300: Suggestion: Since the `transformSnippetStringToInsertText` function isn't bound to anything, references to `this` in there will fail. To prevent such an error in the future, either use a lambda or transform `transformSnippetStringToInsertText` to a field:

```ts
private transformSnippetStringToInsertText = (resourceEdit: MonacoResourceTextEdit): TextEdit & { insertAsSnippet?: boolean } => {
    ...
}  
```

msujew (commented):
The test extension seems to work well 👍 Just a few comments, see below.

packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -1284,14 +1284,9 @@ export class SnippetTextEdit implements theia.SnippetTextEdit {
     snippet: SnippetString;
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
msujew, line 1286: Minor: Disabling the ESLint rule should no longer be necessary.

packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -1284,14 +1284,9 @@ export class SnippetTextEdit implements theia.SnippetTextEdit {
...
-    static isSnippetTextEdit(thing: any): thing is SnippetTextEdit {
-        if (thing instanceof SnippetTextEdit) {
-            return true;
-        }
-        if (!thing) {
-            return false;
-        }
-        return Range.isRange((<SnippetTextEdit>thing).range)
+    static isSnippetTextEdit(thing: unknown): thing is SnippetTextEdit {
+        return isObject<SnippetTextEdit>(thing)
```
msujew, line 1288: Suggestion: I believe it still makes sense to have the shortcut here:
```suggestion
        return thing instanceof SnippetTextEdit || isObject<SnippetTextEdit>(thing)
```

msujew (approved):
Looks good to me 👍

JonasHelming:
@msujew Do you want to merge this?

Issue #12016, opened 2022-12-22
[vscode] Support InputBox#valueSelection


rschnekenbu:
see https://code.visualstudio.com/api/references/vscode-api#InputBox

- valueSelection: readonly [number, number]
_Selection range in the input value. Defined as tuple of two number where the first is the inclusive start index and the second the exclusive end index. When undefined the whole pre-filled value will be selected, when empty (start equals end) only the cursor will be set, otherwise the defined range will be selected.
This property does not get updated when the user types or makes a selection, but it can be updated by the extension._

jonah-iden:
Hi, i would try working on this issue

PR #12050, opened 2023-01-09
vscode: add support for `valueSelection` in `InputBox`

jonah-iden:

#### What it does
Fixes: #12016.

it adds a new field to the InputBox with which a specific range of text from the InputBox can be slected

#### How to test
i build a small vscode extension for testing found here: https://github.com/jonah-iden/inputBox-ValueSelection-test
and as vsix  [value-selection-test-0.0.1.zip](https://github.com/eclipse-theia/theia/files/10373489/value-selection-test-0.0.1.zip)

it just opens up a input box which selects all text until the firs space character when the text is changed

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/quick-open.ts
+++ b/packages/plugin-ext/src/plugin/quick-open.ts
@@ -582,7 +582,7 @@ export class InputBoxExt extends QuickInputExt implements theia.InputBox {
     }
 
     set valueSelection(valueSelection: readonly [number, number] | undefined) {
-        this._valueSelection = valueSelection ?? [0, this.value.length];
+        this._valueSelection = valueSelection;
         this.update({ valueSelection });
     }
 
--- a/packages/plugin-ext/src/plugin/quick-open.ts
+++ b/packages/plugin-ext/src/plugin/quick-open.ts
@@ -543,7 +543,7 @@ export class InputBoxExt extends QuickInputExt implements theia.InputBox {
 
     private _password: boolean;
     private _prompt: string | undefined;
-    private _valueSelection: readonly [number, number]
+    private _valueSelection: readonly [number, number];
     private _validationMessage: string | undefined;
 
     constructor(
--- a/packages/plugin-ext/src/plugin/quick-open.ts
+++ b/packages/plugin-ext/src/plugin/quick-open.ts
@@ -582,8 +582,8 @@ export class InputBoxExt extends QuickInputExt implements theia.InputBox {
     }
 
     set valueSelection(valueSelection: readonly [number, number] | undefined) {
-        this._valueSelection = valueSelection ? valueSelection : [0, this.value.length]
-        this.update({ valueSelection })
+        this._valueSelection = valueSelection ? valueSelection : [0, this.value.length];
+        this.update({ valueSelection });
     }
 
     get validationMessage(): string | undefined {
--- a/packages/plugin-ext/src/plugin/quick-open.ts
+++ b/packages/plugin-ext/src/plugin/quick-open.ts
@@ -582,7 +582,7 @@ export class InputBoxExt extends QuickInputExt implements theia.InputBox {
     }
 
     set valueSelection(valueSelection: readonly [number, number] | undefined) {
-        this._valueSelection = valueSelection ? valueSelection : [0, this.value.length];
+        this._valueSelection = valueSelection ?? [0, this.value.length];
         this.update({ valueSelection });
     }
 
--- a/packages/plugin-ext/src/plugin/quick-open.ts
+++ b/packages/plugin-ext/src/plugin/quick-open.ts
@@ -543,7 +543,7 @@ export class InputBoxExt extends QuickInputExt implements theia.InputBox {
 
     private _password: boolean;
     private _prompt: string | undefined;
-    private _valueSelection: readonly [number, number];
+    private _valueSelection: readonly [number, number] | undefined;
     private _validationMessage: string | undefined;
 
     constructor(
@@ -577,7 +577,7 @@ export class InputBoxExt extends QuickInputExt implements theia.InputBox {
         this.update({ prompt });
     }
 
-    get valueSelection(): readonly [number, number] {
+    get valueSelection(): readonly [number, number] | undefined {
         return this._valueSelection;
     }
 
--- a/packages/plugin-ext/src/plugin/quick-open.ts
+++ b/packages/plugin-ext/src/plugin/quick-open.ts
@@ -543,6 +543,7 @@ export class InputBoxExt extends QuickInputExt implements theia.InputBox {
 
     private _password: boolean;
     private _prompt: string | undefined;
+    private _valueSelection: readonly [number, number]
     private _validationMessage: string | undefined;
 
     constructor(
@@ -574,6 +575,15 @@ export class InputBoxExt extends QuickInputExt implements theia.InputBox {
     set prompt(prompt: string | undefined) {
         this._prompt = prompt;
         this.update({ prompt });
+    }
+
+    get valueSelection(): readonly [number, number] {
+        return this._valueSelection;
+    }
+
+    set valueSelection(valueSelection: readonly [number, number] | undefined) {
+        this._valueSelection = valueSelection ? valueSelection : [0, this.value.length]
+        this.update({ valueSelection })
     }
 
     get validationMessage(): string | undefined {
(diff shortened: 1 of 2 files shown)


jonah-iden:
I sadly have no idea where the extra changes in theia.d.ts are comming from. I can't see them locally, if someone has any idea how to remove them please tell me

packages/plugin-ext/src/plugin/quick-open.ts
```diff
@@ -543,6 +543,7 @@ export class InputBoxExt extends QuickInputExt implements theia.InputBox {
 
     private _password: boolean;
     private _prompt: string | undefined;
+    private _valueSelection: readonly [number, number]
```
msujew, line 546: Minor:
```suggestion
    private _valueSelection: readonly [number, number];
```

packages/plugin/src/theia.d.ts
```diff
@@ -5323,6 +5323,19 @@ export module '@theia/plugin' {
          */
         password: boolean;
 
+
```
msujew, line 5326: Minor: Our linter complains here that there is more than one consecutive empty line here (below `valueSelection` as well).

packages/plugin-ext/src/plugin/quick-open.ts
jonah-iden: fixed

packages/plugin/src/theia.d.ts
jonah-iden: removed the lines

packages/plugin/src/theia.d.ts
```diff
@@ -5323,6 +5323,28 @@ export module '@theia/plugin' {
...
+        /**
+         * Selection range in the input value. Defined as tuple of two number where the
+         * first is the inclusive start index and the second the exclusive end index. When `undefined` the whole
+         * pre-filled value will be selected, when empty (start equals end) only the cursor will be set,
+         * otherwise the defined range will be selected.
+         *
+         * This property does not get updated when the user types or makes a selection,
+         * but it can be updated by the extension.
+         */
+        valueSelection: readonly [number, number];
```
msujew, line 5335: This seems to be duplicated.

msujew (commented):
Do you mind rebasing your changes to the latest `master`, just to have the `theia.d.ts` styling fixes out of the way.

packages/plugin/src/theia.d.ts
jonah-iden: removed the duplicate

vince-fugnitto (commented):
@jonah-iden please take a look at the linting errors due to missing semicolons: https://github.com/eclipse-theia/theia/actions/runs/3884942760/jobs/6628169768#step:6:832.

If you haven't already I'd suggest installing the `dbaeumer.vscode-eslint` extension which highlights eslint errors and warnings directly in the editor.

packages/plugin-ext/src/plugin/quick-open.ts
```diff
@@ -576,6 +577,15 @@ export class InputBoxExt extends QuickInputExt implements theia.InputBox {
         this.update({ prompt });
     }
 
+    get valueSelection(): readonly [number, number] {
+        return this._valueSelection;
+    }
+
+    set valueSelection(valueSelection: readonly [number, number] | undefined) {
+        this._valueSelection = valueSelection ? valueSelection : [0, this.value.length];
```
paul-marechal, line 585:
```suggestion
        this._valueSelection = valueSelection ?? [0, this.value.length];
```

paul-marechal:
Despite what the typings for the getter say, I believe that `valueSelection` can be `undefined` until someone explicitly sets it somehow. I'm also confused with how the setter accepts `undefined` wen the getter says it's always defined??

Please make sure it's all consistent.

jonah-iden:
your totally right though. ValueSelection can be undefined until the setter is called. I think it was this way in the vs code code

packages/plugin/src/theia.d.ts
```diff
@@ -5377,6 +5377,17 @@ export module '@theia/plugin' {
...
```
msujew, line 5389: This doesn't seem to align with the [vscode definition of `valueSelection`](https://github.com/microsoft/vscode/blob/a606c86868801bdd6a3f4ddd3ba6253d068d0ed5/src/vscode-dts/vscode.d.ts#L11544), which has an additional ` | undefined`.

packages/plugin-ext/src/plugin/quick-open.ts
```diff
@@ -576,6 +577,15 @@ export class InputBoxExt extends QuickInputExt implements theia.InputBox {
...
     }
 
+    get valueSelection(): readonly [number, number] {
+        return this._valueSelection;
+    }
+
+    set valueSelection(valueSelection: readonly [number, number] | undefined) {
+        this._valueSelection = valueSelection ?? [0, this.value.length];
+        this.update({ valueSelection });
+    }
```
msujew, line 587: With the above comment in mind, we don't need to set a default value for `_valueSelection`. Especially since `this.value` can change without `this._valueSelection` being updated.

packages/plugin-ext/src/plugin/quick-open.ts
```diff
@@ -576,6 +577,15 @@ export class InputBoxExt extends QuickInputExt implements theia.InputBox {
         this.update({ prompt });
     }
 
+    get valueSelection(): readonly [number, number] | undefined {
+        return this._valueSelection;
+    }
+
+    set valueSelection(valueSelection: readonly [number, number] | undefined) {
+        this._valueSelection = valueSelection ?? [0, this.value.length];
```
msujew, line 585: I don't believe we need to set a default value for this anymore, right?

jonah-iden: this is specified in the vs code api docs that setting undefined is supposed to do this

jonah-iden: _When undefined the whole pre-filled value will be selected_

msujew: I believe the underlying `IQuickInput` element will take care of this by itself. We don't need to explicitly set the value, especially since the following will yield unexpected results:

```
quickOpen.value = '';
quickOpen.valueSelection = undefined;
quickOpen.value= 'x';
```

jonah-iden: ok I'll change and test it again

msujew (approved):
Looks good to me 👍 

- [x] The API is marked as supported
- [x] The extension works as expected

jonah-iden:
if this PR is okay someone else would have to merge it since i don't have write access rights yet

packages/plugin/src/theia.d.ts
```diff
@@ -5377,6 +5377,17 @@ export module '@theia/plugin' {
...
+        /**
+         * Selection range in the input value. Defined as tuple of two number where the
+         * first is the inclusive start index and the second the exclusive end index. When `undefined` the whole
+         * pre-filled value will be selected, when empty (start equals end) only the cursor will be set,
+         * otherwise the defined range will be selected.
+         *
+         * This property does not get updated when the user types or makes a selection,
+         * but it can be updated by the extension.
+         */
+        valueSelection: readonly [number, number] | undefined;
```
paul-marechal, line 5389: Looking at the VS Code API `valueSelection` shouldn't be undefined. I think the solution is to ensure the field is always defined, rather than allowing it to be undefined.

paul-marechal: Note that there is a field named `valueSelection` that can be undefined, but it's from `InputBoxOptions` and not `InputBox` as is the case here.

jonah-iden: but if you look into the vscode.d.ts in the vs code source code there the field is defined as `valueSelection: readonly [number, number] | undefined;` for `InputBox` and `valueSelection?: [number, number]; for `InputBoxOptions`
`

jonah-iden: so this is exactly how its implementd in vscode

paul-marechal (changes requested):
~~`InputBox#valueSelection` shouldn't be undefined according to the VS Code API.~~

edit: Turns out `vscode.d.ts` allows `undefined` but not the web documentation...

paul-marechal (approved):

packages/plugin/src/theia.d.ts
paul-marechal: You are right, I just double checked in `vscode.d.ts` and it seems to be allowed... I was solely looking at the web documentation here: https://code.visualstudio.com/api/references/vscode-api.

It seems to be an issue with their web doc, your changes are good.

Issue #11985, opened 2022-12-13
dialog: `ConfirmDialog` should respect line-breaks for `msg`


vince-fugnitto:
#### Feature Description:

The `ConfirmDialog` does not currently respect line breaks for `msg` content.

Sample test code:
```ts
const dialog = new ConfirmDialog({
    title: 'Sample Dialog',
    msg: 'A\r\nB\r\nC\r\n'
});
return dialog.open();
```
You can add the code [here](https://github.com/eclipse-theia/theia/blob/master/examples/api-samples/src/browser/menu/sample-menu-contribution.ts#L58), rebuild the example application and trigger the menu using `Sample Command` like so:
![image](https://user-images.githubusercontent.com/25749063/207399170-518810b1-9c4f-4c55-8288-30dd075f79c0.png)

Will produce a dialog of:

![image](https://user-images.githubusercontent.com/40359487/207383409-caa8eba3-7f58-4bd1-a89e-cc20cc3732ae.png)

Instead of:

![image](https://user-images.githubusercontent.com/40359487/207383568-369bf85b-870a-40c0-8a1f-ae3a0d00229b.png)

The fix is to include the following CSS to respect the content passed as a `msg` ([dialogContent](https://github.com/eclipse-theia/theia/blob/ac7a9ff51afa71d26673a7a8f7968c8e9a3f51ff/packages/core/src/browser/style/dialog.css#L60-L68)):

```patch
.p-Widget.dialogOverlay .dialogContent {
    display: flex;
    flex-grow: 2;
    flex-direction: column;
    justify-content: space-around;
    align-items: stretch;
    position: relative;
    padding: calc(var(--theia-ui-padding)*2);
+   white-space: pre;
}
```

jonah-iden:
Hi, as a new contributor would it be ok for me to work on this ticket even though it has the hackathon label or should i rather look for a different one?

vince-fugnitto:
> Hi, as a new contributor would it be ok for me to work on this ticket even though it has the hackathon label or should i rather look for a different one?

@jonah-iden please go ahead :)

PR #12052, opened 2023-01-09
core: fix displaying for large `ConfirmDialog` dialogs

jonah-iden:

#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/11985.

smaller css fixes for the confirm dialog so that it respects linebreaks and is scrollable in case of too much text.
I used pre-line for automatic line breaks so the dialog only needs vertical scroll though it could be nice if stacktraces or simmilar long lines are supposed to be displayed in it, to also give it horizontal scrolling

#### How to test
in `sample-menu-contributions.ts` in the sample command instead of returning an alert we can create a ConfirmDialog.
something like this
```
dialog = new ConfirmDialog({
    title: 'Sample Dialog',
    msg: 'A\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\nA\r\n'
});
return dialog.open();
```
clicking on sampleCommand in theia should then open the dialog with scrollable multiline message.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/style/dialog.css
+++ b/packages/core/src/browser/style/dialog.css
@@ -64,10 +64,14 @@
     justify-content: space-around;
     align-items: stretch;
     position: relative;
-    margin: calc(var(--theia-ui-padding)*2);
+    padding: calc(var(--theia-ui-padding)*2);
     white-space: pre-line;
     max-height: 50vh;
     overflow-y: auto;
+}
+
+.p-Widget.dialogOverlay .dialogContent > * {
+    height: 100%;
 }
 
 .p-Widget.dialogOverlay .dialogControl {
--- a/packages/core/src/browser/style/dialog.css
+++ b/packages/core/src/browser/style/dialog.css
@@ -76,7 +76,7 @@
 
 .p-Widget.dialogOverlay .dialogControl {
     padding: calc(var(--theia-ui-padding)*2);
-    padding-top: 0px;
+    padding-top: var(--theia-ui-padding);
     display: flex;
     flex-direction: row;
     align-content: right;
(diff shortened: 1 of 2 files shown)


paul-marechal:
I pushed a commit adding a sample command to try your feature and it seems like there is a bug: The text seems to be truncated where if the text is big enough it only displays the 100 last lines or so?

jonah-iden:
oh yeah your totally right. Seems to be because of the display: flex. I'll look into it

packages/core/src/browser/style/dialog.css
```diff
@@ -76,7 +76,7 @@
 
 .p-Widget.dialogOverlay .dialogControl {
     padding: calc(var(--theia-ui-padding)*2);
-    padding-top: 0px;
+    padding-top: var(--theia-ui-padding);
```
paul-marechal, line 79: I don't even understand why `padding-top` was set to 0 before, but if we're going to add padding again why not remove that rule completely? It looks better to me this way in our current scenarios.

cc @vince-fugnitto?

jonah-iden: for me personally it looked better with just the single padding instead of the double. But sure i can change it

jonah-iden: "When undefined the whole pre-filled value will be selected"

msujew (approved):
Looks good to me 👍 I can confirm that this addresses the overflow and the newline issue.

vince-fugnitto:
@jonah-iden @msujew I believe we should revert the changes or fix the regressions since the changes break dialogs like "open folder":

image

jonah-iden:
thats not good at all. Im really sorry for that. 
I may have a fix for that though. I'll link it here as soon as i created the PR

PR #12060, opened 2023-01-11
[vscode] Add iconPath and stub color optional properties to TerminalOptions and ExtensionTerminalOptions

### LINKED ISSUE #11504: [vscode] Support optional properties `iconPath` and `color` in `TerminalOptions` and `ExtensionTerminalOptions`
The type [`TerminalOptions`](https://code.visualstudio.com/api/references/vscode-api#TerminalOptions) and [`ExtensionTerminalOptions`](https://code.visualstudio.com/api/references/vscode-api#ExtensionTerminalOptions) miss the optional properties:

- [ ] `iconPath`: The icon path or `ThemeIcon` for the terminal
- [ ] `color`: `ThemeColor` for the terminal

Related VSCode API issues:
- #11503
- #11138
- #11506

rschnekenbu:
#### What it does

_Add iconPath optional property_ to TerminalOptions and ExtensionTerminalOptions
_Stub color optional property_ to TerminalOptions and ExtensionTerminalOptions
This allows creation of terminals from extension with different icons.

Fixes #11504 

#### How to test 
1. Install extension [terminal-creation-options-0.0.13.zip](https://github.com/eclipse-theia/theia/files/10394158/terminal-creation-options-0.0.13.zip)
. This extension provides several actions to open a new terminal with a different icon in the title and several color options (not supported yet)
2. Trigger in the command Palette different commands:
* 'Create terminal with green Theme Icon'. a new terminal with a file icon shall open.
* 'Create  terminal with red URI Icon'. A terminal with default terminal icon shall open as the URIs iconPath is not yet supported.
* 'Create new terminal' should create a terminal with standard terminal icon. 

terminal-colorandicon

#### Known limitations
1. Color is not supported but stubbed, so the extension is still supported. The support is strange in vscode, as having a specific color changes also the icon to a kind of cube (at least when I am testing). The main technical reason is that underlying representation Title<Widget>, coming from phosphorjs library, does not directly support styling. The phosphorjs library is deprecated now, so this may be hard to have a proper fix. 
2. Only ThemeIcon is supported currently in iconPath. URI(s) based icons are not supported, but extension does not fail. Same reason for the color management applies here. The limitation comes from Title<Widget> only accepting iconClass string. I have seen some similar work on webview title, but this includes some dependencies to classes in plugin-ext package (PluginSharedStyle, and some other elements). In Terminal-Widget-impl, we are in terminal package. Terminal package  can not depend on plugin-ext as it would introduce circular dependencies. So there maybe some other way to implement, but with either some code duplication or a larger refactoring required. This will be part of a following issue once this PR is merged.

#### Review checklist

#### Reminder for reviewers

- [x] As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

### DIFF
--- a/packages/plugin-ext/src/main/browser/terminal-main.ts
+++ b/packages/plugin-ext/src/main/browser/terminal-main.ts
@@ -27,6 +27,7 @@ import { SerializableEnvironmentVariableCollection } from '@theia/terminal/lib/c
 import { ShellTerminalServerProxy } from '@theia/terminal/lib/common/shell-terminal-protocol';
 import { TerminalLink, TerminalLinkProvider } from '@theia/terminal/lib/browser/terminal-link-provider';
 import { URI } from '@theia/core/lib/common/uri';
+import { getIconClass } from '../../plugin/terminal-ext';
 
 /**
  * Plugin api service allows working with terminal emulator.
@@ -127,6 +128,7 @@ export class TerminalServiceMainImpl implements TerminalServiceMain, TerminalLin
             const terminal = await this.terminals.newTerminal({
                 id,
                 title: options.name,
+                iconClass: getIconClass(options),
                 shellPath: options.shellPath,
                 shellArgs: options.shellArgs,
                 cwd: options.cwd ? new URI(options.cwd) : undefined,
--- a/packages/plugin-ext/src/plugin/terminal-ext.ts
+++ b/packages/plugin-ext/src/plugin/terminal-ext.ts
@@ -20,9 +20,25 @@ import { RPCProtocol } from '../common/rpc-protocol';
 import { Event, Emitter } from '@theia/core/lib/common/event';
 import { Deferred } from '@theia/core/lib/common/promise-util';
 import * as theia from '@theia/plugin';
-import { Disposable, EnvironmentVariableMutatorType } from './types-impl';
+import { Disposable, EnvironmentVariableMutatorType, ThemeIcon } from './types-impl';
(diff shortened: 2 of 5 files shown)


vince-fugnitto:
@rschnekenbu the APIs added are already optional and shouldn't affect the extension from loading or starting. Given that, why should we stub and not just implement the API properly when ready?

rschnekenbu:
@vince-fugnitto, I think this PR still makes sense for 2 points: the support for the icon is already be partially availlable for the incoming release and the API is clearly marked as stubbed, not simply ignored. Second point would mean we could step up the vscode API compatibility version with some remaining work clearly tracked. 
Thanks in any case to have commented this PR!

vince-fugnitto:
@rschnekenbu If we want to stub I'm mainly wondering if we should instead mark both `iconPath` and `color` as stubbed given they are implemented partially or not at alll. Currently we have (compatibility report with this branch):

image

image

It is misleading to mark `iconPath` as fully compatible, perhaps `partial` or `stubbed` is more representative. The `color` property for `ExtensionTerminalOptions` should be marked as `stubbed` as well.

rschnekenbu:
@vince-fugnitto, indeed, that can be misleading for iconPath. I updated the contribution removing the URIs on iconPath.
I also added 'stubbed' marker to ExtensionTerminalOptions.color.

Here is the report on TerminalOptions and ExtensionTerminalOptions now:
TerminalOptions
ExtensionTerminalOptions

vince-fugnitto (approved):
LGTM :+1: I think it's better now that we represent the actual support, and we can track compatibility.

Issue #12025, opened 2022-12-22
[vscode] Support DocumentDropEditProvider and DocumentDropEdit


rschnekenbu:
see [DocumentDropEditProvider](https://code.visualstudio.com/api/references/vscode-api#DocumentDropEditProvider) and [DocumentDropEdit](https://code.visualstudio.com/api/references/vscode-api#DocumentDropEdit)
since vscode 1.70

namespace [languages](https://code.visualstudio.com/api/references/vscode-api#languages) should probably be updated also.

jonah-iden:
Hi, i would like to work on this issue. could someone assign me?

jonah-iden:
just realized this depends on #11776. So I'll have to wait for #12065 to be merged

Issue #6623, opened 2019-11-25
Implement SelectionRange and SelectionRangeProvider


tolusha:
### Description
`css-language-features`, `html-language-features`, `json-language-features` builtin extensions failed due to unimplemented VS Code API.

https://github.com/Microsoft/vscode/blob/8795a9889db74563ddd43eb0a897a2384129a619/src/vs/vscode.d.ts#L3842
https://github.com/Microsoft/vscode/blob/8795a9889db74563ddd43eb0a897a2384129a619/src/vs/vscode.d.ts#L3863

akosyakov:
Right now html and css coming from monaco. It's backed in the generator: https://github.com/eclipse-theia/theia/blob/b8c5e18530b93c5a8b7a1890b68f5bcc202e5817/dev-packages/application-manager/src/generator/webpack-generator.ts#L56-L57

We should make it optional somehow. Otherwise there would be duplicate proposals from monaco and vs code built-in extensions.

vzhukovskii:
Implemented via https://github.com/eclipse-theia/theia/pull/7534

PR #12065, opened 2023-01-12
Support tree view drag and drop. 

### LINKED ISSUE #11776: [vscode] Support DataTransfer
A [DataTransfer](https://code.visualstudio.com/api/references/vscode-api#DataTransfer) is a map containing a mapping of the mime type of the corresponding transferred data.
API is available since VSCode 1.66.0

The API consists of:
- [ ] [DataTransfer](https://code.visualstudio.com/api/references/vscode-api#DataTransfer) 
- [ ] [DataTransferFile](https://code.visualstudio.com/api/references/vscode-api#DataTransferFile)
- [ ] [DataTransferItem](https://code.visualstudio.com/api/references/vscode-api#DataTransferItem)

### LINKED ISSUE #11778: [vscode] Support TreeDragAndDropController and optional property TreeViewOptions.dragAndDropController
[TreeDragAndDropController](https://code.visualstudio.com/api/references/vscode-api#TreeDragAndDropController) API was added in version 1.66.0, along the optional property dragAndDropController in [TreeViewOptions](https://code.visualstudio.com/api/references/vscode-api#TreeViewOptions)

It provides support for drag and drop in [TreeView](https://code.visualstudio.com/api/references/vscode-api#TreeView%3CT%3E).

tsmaeder:

#### What it does
Adds support for contributing drag and drop controller in the tree view plugin API. The code tries to mimick the behaviour described in https://code.visualstudio.com/api/references/vscode-api#TreeDragAndDropController&lt;T&gt; as well as possible. I have also refactored some of the "drag into editor area" code to work with the new code (and fixed dragging external files into electron on windows as a "drive-by")

Fixes #11778, Fixes #11776

#### How to test
I have adapted the VS Code tree view sample to exercise the code added in the PR. The extension & source is attached to this PR.
The extension contributes a view called "Test View Drag and Drop", which you need to open.

The example implement multiple features:
1. Drag and drop within the tree
The dropped element should be moved under the element where it was dropped.
Note that there seems to be a bug in our tree-view refresh code where the UI goes into an endless loop when you reparent the elements of the tree: It seems the parent field of the tree items is not properly updated when an item is reparented. I am pretty convinced this is not related to the changes in this PR.
2. Drag an element to the editor area:
This should open a *.js file from the extension itself: note that the cursor will be placed at line 10, column 6.
3. Exercise the current drag and drop to the editor area. Drag from the file navigator and from the OS file explorer
4. Exercise dropping files to the example view: the extension will show the first couple of bytes of the file content.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/application-shell.ts
+++ b/packages/core/src/browser/shell/application-shell.ts
@@ -41,6 +41,8 @@ import { Deferred } from '../../common/promise-util';
 import { SaveResourceService } from '../save-resource-service';
 import { nls } from '../../common/nls';
 import { SecondaryWindowHandler } from '../secondary-window-handler';
+import URI from '../../common/uri';
+import { OpenerService } from '../opener-service';
 
 /** The class name added to ApplicationShell instances. */
 const APPLICATION_SHELL_CLASS = 'theia-ApplicationShell';
@@ -190,9 +192,13 @@ export class ApplicationShell extends Widget {
 
     private readonly tracker = new FocusTracker<Widget>();
     private dragState?: WidgetDragState;
+    additionalDraggedUris: URI[] | undefined;
 
     @inject(ContextKeyService)
     protected readonly contextKeyService: ContextKeyService;
+
+    @inject(OpenerService)
+    protected readonly openerService: OpenerService;
 
     protected readonly onDidAddWidgetEmitter = new Emitter<Widget>();
     readonly onDidAddWidget = this.onDidAddWidgetEmitter.event;
@@ -498,7 +504,66 @@ export class ApplicationShell extends Widget {
         dockPanel.id = MAIN_AREA_ID;
         dockPanel.widgetAdded.connect((_, widget) => this.fireDidAddWidget(widget));
         dockPanel.widgetRemoved.connect((_, widget) => this.fireDidRemoveWidget(widget));
+
+        const openUri = async (fileUri: URI) => {
+            try {
+                const opener = await this.openerService.getOpener(fileUri);
+                opener.open(fileUri);
+            } catch (e) {
+                console.info(`no opener found for '${fileUri}'`);
+            }
+        };
+
+        dockPanel.node.addEventListener('drop', event => {
+            if (event.dataTransfer) {
+                const uris = this.additionalDraggedUris || ApplicationShell.getDraggedEditorUris(event.dataTransfer);
+                if (uris.length > 0) {
+                    uris.forEach(openUri);
+                } else if (event.dataTransfer.files?.length > 0) {
+                    // the files were dragged from the outside the workspace
+                    Array.from(event.dataTransfer.files).forEach(async file => {
+                        if (file.path) {
+                            const fileUri = URI.fromComponents({
+                                scheme: 'file',
+                                path: file.path,
+                                authority: '',
+                                query: '',
+                                fragment: ''
+                            });
+                            openUri(fileUri);
+                        }
+                    });
+                }
+            }
+        });
+        const handler = (e: DragEvent) => {
+            if (e.dataTransfer) {
+                e.dataTransfer.dropEffect = 'link';
+                e.preventDefault();
+                e.stopPropagation();
+            }
+        };
+        dockPanel.node.addEventListener('dragover', handler);
+        dockPanel.node.addEventListener('dragenter', handler);
+
         return dockPanel;
+    }
+
+    addAdditionalDraggedEditorUris(uris: URI[]): void {
+        this.additionalDraggedUris = uris;
+    }
+
+    clearAdditionalDraggedEditorUris(): void {
+        this.additionalDraggedUris = undefined;
+    }
+
+    static getDraggedEditorUris(dataTransfer: DataTransfer): URI[] {
+        const data = dataTransfer.getData('theia-editor-dnd');
+        return data ? data.split('\n').map(entry => new URI(entry)) : [];
+    }
+
+    static setDraggedEditorUris(dataTransfer: DataTransfer, uris: URI[]): void {
+        dataTransfer.setData('theia-editor-dnd', uris.map(uri => uri.toString()).join('\r\n'));
     }
 
     /**
--- a/packages/core/src/browser/shell/application-shell.ts
+++ b/packages/core/src/browser/shell/application-shell.ts
@@ -521,7 +521,7 @@ export class ApplicationShell extends Widget {
                     uris.forEach(openUri);
                 } else if (event.dataTransfer.files?.length > 0) {
                     // the files were dragged from the outside the workspace
-                    Array.from(event.dataTransfer.files).forEach(async file => {
+                    Array.from(event.dataTransfer.files).forEach(file => {
                         if (file.path) {
                             const fileUri = URI.fromComponents({
                                 scheme: 'file',
--- a/packages/core/src/browser/shell/application-shell.ts
+++ b/packages/core/src/browser/shell/application-shell.ts
@@ -557,13 +557,13 @@ export class ApplicationShell extends Widget {
         this.additionalDraggedUris = undefined;
     }
 
-    static getDraggedEditorUris(dataTransfer: DataTransfer): URI[] {
+    protected static getDraggedEditorUris(dataTransfer: DataTransfer): URI[] {
         const data = dataTransfer.getData('theia-editor-dnd');
         return data ? data.split('\n').map(entry => new URI(entry)) : [];
     }
 
     static setDraggedEditorUris(dataTransfer: DataTransfer, uris: URI[]): void {
-        dataTransfer.setData('theia-editor-dnd', uris.map(uri => uri.toString()).join('\r\n'));
+        dataTransfer.setData('theia-editor-dnd', uris.map(uri => uri.toString()).join('\n'));
     }
 
     /**
--- a/packages/filesystem/src/browser/file-tree/file-tree-widget.tsx
+++ b/packages/filesystem/src/browser/file-tree/file-tree-widget.tsx
@@ -25,6 +25,7 @@ import { FileUploadService } from '../file-upload-service';
 import { DirNode, FileStatNode, FileStatNodeData } from './file-tree';
 import { FileTreeModel } from './file-tree-model';
 import { IconThemeService } from '@theia/core/lib/browser/icon-theme-service';
+import { ApplicationShell } from '@theia/core/lib/browser/shell';
 import { FileStat, FileType } from '../../common/files';
 import { isOSX } from '@theia/core';
 
@@ -119,14 +120,18 @@ export class FileTreeWidget extends CompressedTreeWidget {
 
     protected handleDragStartEvent(node: TreeNode, event: React.DragEvent): void {
         event.stopPropagation();
-        let selectedNodes;
-        if (this.model.selectedNodes.find(selected => TreeNode.equals(selected, node))) {
-            selectedNodes = [...this.model.selectedNodes];
-        } else {
-            selectedNodes = [node];
-        }
-        this.setSelectedTreeNodesAsData(event.dataTransfer, node, selectedNodes);
         if (event.dataTransfer) {
+            let selectedNodes;
+            if (this.model.selectedNodes.find(selected => TreeNode.equals(selected, node))) {
+                selectedNodes = [...this.model.selectedNodes];
+            } else {
+                selectedNodes = [node];
+            }
+            this.setSelectedTreeNodesAsData(event.dataTransfer, node, selectedNodes);
+            const uris = selectedNodes.filter(n => FileStatNode.is(n)).map(n => (n as FileStatNode).fileStat.resource);
+            if (uris.length > 0) {
+                ApplicationShell.setDraggedEditorUris(event.dataTransfer, uris);
+            }
             let label: string;
             if (selectedNodes.length === 1) {
                 label = this.toNodeName(node);
--- a/packages/filesystem/src/browser/file-tree/file-tree-widget.tsx
+++ b/packages/filesystem/src/browser/file-tree/file-tree-widget.tsx
@@ -128,7 +128,7 @@ export class FileTreeWidget extends CompressedTreeWidget {
                 selectedNodes = [node];
             }
             this.setSelectedTreeNodesAsData(event.dataTransfer, node, selectedNodes);
-            const uris = selectedNodes.filter(n => FileStatNode.is(n)).map(n => (n as FileStatNode).fileStat.resource);
+            const uris = selectedNodes.filter(FileStatNode.is).map(n => n.fileStat.resource);
             if (uris.length > 0) {
                 ApplicationShell.setDraggedEditorUris(event.dataTransfer, uris);
             }
--- a/packages/plugin-ext/src/main/browser/view/dnd-file-content-store.ts
+++ b/packages/plugin-ext/src/main/browser/view/dnd-file-content-store.ts
@@ -1,0 +1,41 @@
+// *****************************************************************************
+// Copyright (C) 2022 ST Microelectronics and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { injectable } from '@theia/core/shared/inversify';
+
+@injectable()
+export class DnDFileContentStore {
+    static id: number = 0;
+    private files: Map<string, File> = new Map();
+    addFile(f: File): string {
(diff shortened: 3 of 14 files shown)


tsmaeder:
Here's the source:

[tree-view-sample.zip](https://github.com/eclipse-theia/theia/files/10402183/tree-view-sample.zip)

and here's the built extension

[custom-view-samples-0.0.1.zip](https://github.com/eclipse-theia/theia/files/10402189/custom-view-samples-0.0.1.zip)

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
```diff
@@ -546,6 +559,143 @@ export class TreeViewWidget extends TreeViewWelcomeWidget {
...
+
+    handleDragEnd(node: TreeViewNode, event: React.DragEvent<HTMLElement>): void {
+        this.applicationShell.clearAdditionalDraggedEditorUris();
+    }
+
+    handleDragOver(event: React.DragEvent<HTMLElement>): void {
+        if (event.dataTransfer) {
+            const canDrop = event.dataTransfer.types.some(type => event.dataTransfer.types.indexOf(type) >= 0) ||
+                event.dataTransfer.types.indexOf(this.treeDragType) > 0 ||
+                this.options.dropMimeTypes!.indexOf('files') > 0 && event.dataTransfer.files.length > 0;
```
colin-grant-work, line 640: `.includes()` rather than `.indexOf() >= 0` would make this rather long condition easier to read. But if you prefer `indexOf()`, you've mixed `> 0` and `>= 0`. Change the former to the latter, unless there's special significance.

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
```diff
@@ -546,6 +559,143 @@ export class TreeViewWidget extends TreeViewWelcomeWidget {
...
+                            this.dndFileContentStore.removeFile(file);
+                        }
+                    });
+                    files = [];
+                }
+            } finally {
+                for (const file of files) {
+                    this.dndFileContentStore.removeFile(file);
+                }
+            }
```
colin-grant-work, line 695: I'm a little confused about the logic here. It looks like _if_ we've gotten items, or we ran into an error before line 680, then we want to wait before removing the files from the store; otherwise, we want to remove them immediately?

If the call to `.$drop` throws an error, we will never clear the files, because we will have run 689, but we'll never get to 685-688.

If I'm right about the goal, then we could achieve the same end by `awaiting` at line 684 and not bothering with `filesCopy`. Then failure of the `.$drop()` call would be covered by the `try` and we'd always end up in the `finally`, but only after we'd either succeeded or failed in the `.$drop()`, or skipped it because there was nothing to do.

packages/plugin-ext/src/main/browser/view/tree-views-main.ts
```diff
@@ -48,15 +51,22 @@ export class TreeViewsMainImpl implements TreeViewsMain, Disposable {
...
         this.widgetManager = this.container.get(WidgetManager);
+        this.fileContentStore = this.container.get(DnDFileContentStore);
     }
 
     dispose(): void {
         this.toDispose.dispose();
     }
 
-    async $registerTreeDataProvider(treeViewId: string): Promise<void> {
+    async $registerTreeDataProvider(treeViewId: string, dragMimetypes: string[] | undefined, dropMimetypes: string[] | undefined): Promise<void> {
```
colin-grant-work, line 61:
```suggestion
    async $registerTreeDataProvider(treeViewId: string, dragMimeTypes: string[] | undefined, dropMimeTypes: string[] | undefined): Promise<void> {
```

Then you can use shorthand assignment below, if you'd like.

colin-grant-work (commented):
Some initial thoughts on the code

packages/plugin-ext/src/main/browser/view/dnd-file-content-store.ts
```diff
@@ -0,0 +1,41 @@
...
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { injectable } from '@theia/core/shared/inversify';
+
+@injectable()
+export class DnDFileContentStore {
+    static id: number = 0;
```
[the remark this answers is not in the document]
tsmaeder: No, it shouldn't.

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
```diff
@@ -171,8 +174,8 @@ export class PluginTree extends TreeImpl {
     @inject(PluginSharedStyle)
     protected readonly sharedStyle: PluginSharedStyle;
 
-    @inject(TreeViewWidgetIdentifier)
-    protected readonly identifier: TreeViewWidgetIdentifier;
+    @inject(TreeViewWidgetOptions)
+    protected readonly options: TreeViewWidgetOptions;
```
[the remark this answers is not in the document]
tsmaeder: But in reality a fix ;-)

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -498,9 +504,68 @@ export class ApplicationShell extends Widget {
...
+    }
+
+    static getDraggedEditorUris(dataTransfer: DataTransfer): URI[] {
+        const data = dataTransfer.getData('theia-editor-dnd');
+        return data ? data.split('\n').map(entry => new URI(entry)) : [];
+    }
+
+    static setDraggedEditorUris(dataTransfer: DataTransfer, uris: URI[]): void {
+        dataTransfer.setData('theia-editor-dnd', uris.map(uri => uri.toString()).join('\r\n'));
+    }
```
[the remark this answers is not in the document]
tsmaeder: I don't think so: the code only serves to being able to drag URI's like `file:///foo/bar/zoz.java#L3,5` to the editor area. These URI's are not available at the time the drag is initiated. The code has no use outside of the application shell and therefore IMO should be here.

packages/plugin-ext/src/main/browser/view/dnd-file-content-store.ts
```diff
@@ -0,0 +1,41 @@
...
+    private files: Map<string, File> = new Map();
```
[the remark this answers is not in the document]
tsmaeder: Because I don't want this implementation detail to escape the class.

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
```diff
@@ -41,7 +41,7 @@ import { ACTION_ITEM, Widget } from '@theia/core/lib/browser/widgets/widget';
 import { Emitter, Event } from '@theia/core/lib/common/event';
 import { MessageService } from '@theia/core/lib/common/message-service';
 import { View } from '../../../common/plugin-protocol';
-import CoreURI from '@theia/core/lib/common/uri';
+import CoreURI, { URI } from '@theia/core/lib/common/uri';
```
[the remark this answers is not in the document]
tsmaeder: Sorry, but I don't get it 🤷

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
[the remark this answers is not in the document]
tsmaeder: I'll add a note to the changelog.

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
tsmaeder: I'm reluctant to make this an async method, as it's an event handler. It's just to easy to confuse what's happening now and what's later. But you're right about the logic being broken. I'll try to find a more appropriate form.

packages/plugin-ext/src/main/browser/view/tree-views-main.ts
tsmaeder: It's more consistent naming anyway.

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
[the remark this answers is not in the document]
paul-marechal: Here `CoreURI` is the exact same thing as `URI`.

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
```diff
@@ -677,21 +677,21 @@ export class TreeViewWidget extends TreeViewWelcomeWidget {
...
                         }
                     });
                     files = [];
                 }
-            } finally {
+            } catch (e) {
                 for (const file of files) {
                     this.dndFileContentStore.removeFile(file);
                 }
+                throw e;
```
colin-grant-work, line 694: The new logic looks more solid. Now the only slightly (and only slightly) troubling bit is that the connection between `items` and `files` is implicit. According to the current logic, there will only be `files` if there are `items`, so we always clean `files` up if there are any, because we always clean them up if there are `items`. But that connection isn't expressed in the code, so it would be possible in the future to break it (by adding a stricter filter to 'itemness' for example), and then we might not hit either 685 or 692. But that's a strictly hypothetical issue.

tsmaeder: I don't really understand what you mean here: we need to forget every file we remember after the drop operation is over, either by an sync exception or when the `$drop` promise finishes (resolve or reject). There never will be any file in the `dndFileContentStore` when no drop operation is going on.

colin-grant-work: My only point is that the connection between the `items` array that determines whether we get into the `$drop` branch or not and the `files` array that we clear if we don't get into that branch are only implicitly connected. In the current logic, everything should be alright, but if there were every a disconnect between the `items` and the `files`, then it would be possible to add things to the store and not clear them.

tsmaeder: I just don't see how that could happen: we add the ids of the files we put into the store into the `files` array immediately after we put them into the store (#6623), so every remembered file will be in the `files` array and the files array is what we use to release the files again. The items just never figure into the equation. If you're still worried, you'll have to explain it to me like im five ;-)

colin-grant-work: You're right about the current logic - it's solid. But because `items` is one thing and `files` is another, it would be possible for their current synchronicity to be broken in the future. The solution I can see would be not to have a separate `files` array, but add the file handle to the item itself - just here, not as part of the interface - and then use the handle stored on the item when clearing the cache. That would make explicit the 1:1 mapping of items and files. But I'm not really worried about it.

tsmaeder: But there isn't a 1:1 mapping: some items have files, and others don't, so it would be a 1:[0, 1]. I don't think the spec guarantees that all items being dropped are either all files or all not files.

packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -1754,6 +1754,55 @@ export class WorkspaceEdit implements theia.WorkspaceEdit {
...
+
+/**
+ * A map containing a mapping of the mime type of the corresponding transferred data.
+ *
+ * Drag and drop controllers that implement {@link TreeDragAndDropController.handleDrag `handleDrag`} can add additional mime types to the
+ * data transfer. These additional mime types will only be included in the `handleDrop` when the the drag was initiated from
+ * an element in the same drag and drop controller.
+ */
+@es5ClassCompat
+export class DataTransfer implements Iterable<[mimeType: string, item: DataTransferItem]> {
```
jonah-iden, line 1779: Hi, is there a specific reason why the constructor which exists in vscode for this class is not is not implemented here?

jonah-iden: also why is the list implemented as `Map<string, DataTransferItem> `when in vscode its `Map<string, DataTransferItem[]>`

tsmaeder: [`vscode.d.ts`](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.d.ts)  does not specify a constructor. Neither does it mandate any particular implementation.

jonah-iden: thats true. Though while i'm currently working on #12025 i saw that i probably will need to convert from a VSDataTransfer object to our implementation which is implemented with Map<string, IDataTransferItem[]>

jonah-iden: im just not totally sure what the Impact of that could be. But i would guess it could get more difficult to support drag and dop of multiple files

tsmaeder: Wouldn't those be dragged as a `uri-list` anyway? Native DnD does not support multiple items per type, see https://html.spec.whatwg.org/multipage/dnd.html or https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/setData. That said: it's a good test case: what happens when dragging multiple file of the same type into the test view?

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -498,9 +504,68 @@ export class ApplicationShell extends Widget {
...
+        };
+
+        dockPanel.node.addEventListener('drop', event => {
+            if (event.dataTransfer) {
+                const uris = this.additionalDraggedUris || ApplicationShell.getDraggedEditorUris(event.dataTransfer);
+                if (uris.length > 0) {
+                    uris.forEach(openUri);
+                } else if (event.dataTransfer.files?.length > 0) {
+                    // the files were dragged from the outside the workspace
+                    Array.from(event.dataTransfer.files).forEach(async file => {
```
msujew, line 524: Minor: Any reason to make this function `async` if we don't await anything in there?

msujew (commented):
> 3. Exercise the current drag and drop to the editor area. Drag from the file navigator and from the OS file explorer
> 4. Exercise dropping files to the example view: the extension will show the first couple of bytes of the file content.

I noticed that these only work on Electron. Is it out of scope to make these work in the browser version as well?

Also, when dropping a file into an already open editor, even though it does open the file, it also pastes the file path into the existing editor. Can we change that somehow?

packages/core/src/browser/shell/application-shell.ts
tsmaeder: It still should compile to a promise though and thus not block the main loop, right?

msujew: It doesn't matter whether it's a promise or not when it comes to blocking the event loop. It will still be blocking, except if it calls an explicitly non-blocking function (such as `fs.promises`, `setTimeout` or `fetch`). As an example execute the following in any JS shell:

```
[0, 1, 2, 3].forEach(async (e) => { console.log(e) });
console.log('end')
```

It will print `0 1 2 3 end` as the `forEach` callback is instantly executed and blocking the event loop.

tsmaeder: So this doesn't compile to 
```
[0, 1, 2, 3].forEach(async (e) => { new Promise((resolve => {
   console.log(e);
   resolve();
})});
```
?

msujew: My snippet is pure JS, no compilation step needed.

tsmaeder: Under the hood, it must convert to a promise, but I guess it does not push the leading code onto the task queue. So more likely it looks like:
```javascript
async (e) => {
   console.log(e);
   return Promise.resolve();
}
```

tsmaeder: Wat?!

msujew: Anyway, having the `async` in there does not yield any benefits, as it doesn't affect the execution order. JS runtimes will always try to execute as much code as possible until they hit a non-blocking, awaited function. Enough argument to remove the `async`?

packages/filesystem/src/browser/file-tree/file-tree-widget.tsx
```diff
@@ -119,14 +120,18 @@ export class FileTreeWidget extends CompressedTreeWidget {
...
-        this.setSelectedTreeNodesAsData(event.dataTransfer, node, selectedNodes);
         if (event.dataTransfer) {
+            let selectedNodes;
+            if (this.model.selectedNodes.find(selected => TreeNode.equals(selected, node))) {
+                selectedNodes = [...this.model.selectedNodes];
+            } else {
+                selectedNodes = [node];
+            }
+            this.setSelectedTreeNodesAsData(event.dataTransfer, node, selectedNodes);
+            const uris = selectedNodes.filter(n => FileStatNode.is(n)).map(n => (n as FileStatNode).fileStat.resource);
```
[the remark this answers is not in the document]
tsmaeder: Strange that the compiler picks up the typing for `filter(FileStatNode.is)`, but not `filter(node => FileStatNode.is(node)`, but it's true 🤷

packages/plugin-ext/src/plugin/types-impl.ts
tsmaeder: Anyway: since the _interface_ of `DataTransferItem` only supports a single item per MIME type, I don't really see the utility of having a differing implementation.

msujew (approved):
I'm happy with the changes, just a few very minor comment below.

* [x] Dragging a file from the file system explorer works and opens the file in the editor area (on Electron)
* [x] The extension behaves as indicated

Issue #11503, opened 2022-08-03
[vscode] Support `TerminalProfileProvider`


planger:
A [`TerminalProfileProvider`](https://code.visualstudio.com/api/references/vscode-api#TerminalProfileProvider) can be registered via `window.registerTerminalProfileProvider(id, provider)` and eventually provides a [`TerminalProfile`](https://code.visualstudio.com/api/references/vscode-api#TerminalProfile) defining how a terminal will be launched. All those types and their support in Theia are currently missing.

The types of the options with which a terminal can be provided already exist:

- [`TerminalOptions`](https://code.visualstudio.com/api/references/vscode-api#TerminalOptions)
- [`ExtensionTerminalOptions`](https://code.visualstudio.com/api/references/vscode-api#ExtensionTerminalOptions)

Related VSCode API issues:
- #11504
- #11138
- #11506

tsmaeder:
Theia seems to have no support for terminal profiles at all at the moment. Implementing the API correctly would also entail support for configuration in the preferences and a UI to select profiles.

tsmaeder:
My plan of attack for this issue looks like this:

1. Introcuce a `ProfileManagerService` in the front end
2. Read the "legacy" preference values to add default profiles to the profile manager
4. Make TerminalService use the default profiles 
5. Add schema info for terminal profiles in `settings.json`
6. Add code to merge profiles from `setttings.json` into the profile manager
7. Add UI to select profiles and default profiles when creating a terminal
8. Add API support for contributing profiles
9. Add support for `contributes` terminal profiles in `package.json`

Because we might want to declare VS Code `1.64.2` support (even if we lack full functionality), I'm starting with 1. and 8. and proceed from there.

marcdumais-work:
@tsmaeder You can probably test using extension `ms-vscode.js-debug`, installable from the `extensions` view. 


2022-12-16T16:45:06.372Z root ERROR [hosted-plugin: 1993873] Activating extension 'JavaScript Debugger' failed: TypeError: a.window.registerTerminalProfileProvider is not a function
    at t.registerDebugTerminalUI (/tmp/vscode-unpacked/ms-vscode.js-debug-1.65.0.vsix/extension/src/extension.js:2:1279548)
    at t.activate (/tmp/vscode-unpacked/ms-vscode.js-debug-1.65.0.vsix/extension/src/extension.js:2:1073785)
[...]

PR #12066, opened 2023-01-13
Support for terminal profiles. 

tsmaeder:
#### What it does
Adds support for terminal profiles. 

Fixes [#11503](https://github.com/eclipse-theia/theia/issues/11503)

* UI and service to manage terminal profiles
* Handle profiles in preferences according to VS Code schema
* API and contribution markup for contributing profiles and activation event handling

#### How to test
This PR adds two new items under the "Terminal" menu: "New Terminal...", which allows to select a profile and use it to open a terminal and "Choose Default Profile", which selects the profile to be used when selecting "New Terminal" from the menu.
Some profiles are added by default ("cmd" by default on windows), but they can also be contributed by plugins or defined in the preferences. 
Here are two extensions that can be used to test contributed terminals:

[plugins.zip](https://github.com/eclipse-theia/theia/files/10409008/plugins.zip)

One contributes a shell terminal profile ("JavaScript Debug Terminal") and the other a pseudo-pty terminal ("My Extension Terminal"). 

Adding profiles via terminals is described here:

https://code.visualstudio.com/docs/terminal/profiles

Note that terminal "source"s have not been implemented. Theia will accept the syntax, but the resulting profiles will not work.

#### Review checklist

#### Reminder for reviewers


packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -74,6 +80,16 @@ export namespace TerminalCommands {
...
+    export const PROFILE_NEW = Command.toLocalizedCommand({
+        id: 'terminal:new:profile',
+        category: TERMINAL_CATEGORY,
+        label: 'Create New Integrated Terminal from a Profile'
+    });
+    export const PROFILE_DEFAULT = Command.toLocalizedCommand({
+        id: 'terminal:profile:default',
+        category: TERMINAL_CATEGORY,
+        label: 'Choose the default Terminal Profile'
+    });
```
vince-fugnitto, line 92: These commands do not seem to be present in VS Code, and are not added to the `terminal` main-menu.
I do see the command `'Create New Terminal (With Profile)'` however.

packages/terminal/src/browser/terminal-preferences.ts
```diff
@@ -17,9 +17,58 @@
...
+        },
+        markdownDescription: nls.localize('theia/terminal/profileEnv', 'An object with environment variables that will be added to the terminal profile process. Set to `null` to delete environment variables from the base environment.'),
+    },
+    overrideName: {
+        type: 'boolean',
+        description: nls.localizeByDefault('Controls whether or not the profile name overrides the auto detected one.')
+    },
+    icon: {
+        type: 'string',
+        markdownDescription: nls.localize('theia/terminal/profileIcon', 'A codicon ID to associate with the terminal icon.  \nterminal-tmux:"$(terminal-tmux)"')
```
vince-fugnitto, line 41: minor:

```suggestion
        markdownDescription: nls.localize('theia/terminal/profileIcon', 'A codicon ID to associate with the terminal icon. \nterminal-tmux:"$(terminal-tmux)"')
```

vince-fugnitto (commented):
It seems that there are conflicts, do you mind rebasing?

packages/core/i18n/nls.de.json
```diff
@@ -447,12 +447,24 @@
...
+      "defaultProfile": "Das Standardprofil unter {0}",
+      "profiles": "Die Profile welche zur Erzeugung eines Terminals verwendet werden können. Setzen Sie den Pfad von Hand mit optionalen Parametern.\n\nSezen Sie ein Profile auf `null` um es zu verbergen, z.B.: `{0}: null`.",
+      "profilePath": "Der Pfad der Shell, den dieses Profil benutzt.",
+      "profileSource": "Eine Profilquelle, die Shellpfade automatisch erkennt. Unübliche Installationsorte werden nicht unterstützt und müssen manuell erfasst werden",
+      "profileArgs": "Die Shellparameter, welche dieses Profil verwendet.",
+      "profileEnv": "Ein Objekt mit Umgebungsvariablen die zum Terminalprozess hinzugefügt werden. Setzen Sie Variablen auf `null` um sie aus der Basisumgebung zu löschen",
+      "profileIcon": "Eine codicon ID zur Verwendung mit diesem Terminal.  \nterminal-tmux:\"$(terminal-tmux)\"",
+      "profileColor": "ID einer Terminal-Themenfarbe zur Verwendung mit diesem Terminal.",
```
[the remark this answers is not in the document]
tsmaeder: Would you prefer translations by a German speaker or machine translations ;-)? As for the original nls.json: I assume the "defaults" would be taken from the code, right? If so, why do we even have the file?

packages/terminal/src/browser/terminal-frontend-contribution.ts
tsmaeder: terminal-frontend-contribution.ts L 653-668?

packages/terminal/src/browser/terminal-preferences.ts
tsmaeder: What's the difference? Looks identical to me 🤷

packages/core/i18n/nls.de.json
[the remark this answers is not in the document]
msujew: Keeping the german translation in there is reasonable, however you should remove the entries from the `nls.json`. The automatic translation update will not be performed in case the `nls.json` doesn't change after performing the extraction CLI command.

packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -212,6 +245,145 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
...
+            if (e.preferenceName.startsWith('terminal.integrated.')) {
+                this.mergePreferencesPromise = this.mergePreferencesPromise.finally(() => this.mergePreferences());
+            }
+        });
+        this.mergePreferencesPromise = this.mergePreferencesPromise.finally(() => this.mergePreferences());
+
+        this.profileService.onAdded(id => {
+            // extension contributions get read after this point: need to set the default profile if necessary
+            let defaultProfileId;
+            switch (OS.type()) {
```
msujew, line 261: Suggestion: Given that the terminal runs on the backend, it makes more sense to decide on the preference by using the backend OS, instead of the frontend OS.

```suggestion
            switch (OS.backend.type()) {
```

packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -212,6 +245,145 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
...
+                    defaultProfileId = this.terminalPreferences['terminal.integrated.defaultProfile.osx'];
+                    break;
+                }
+            }
+            this.profileService.setDefaultProfile(defaultProfileId);
+        });
+    }
+
+    async contributeDefaultProfiles(): Promise<void> {
+        if (isWindows) {
```
msujew, line 280: Suggestion: Use `backend.isWindows` instead.

packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -212,6 +245,145 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
...
+        // contribute default profiles based on legacy preferences
+    }
+
+    protected async mergePreferences(): Promise<void> {
+        let profiles: Profiles;
+        let defaultProfile: string;
+        let legacyShellPath: string | undefined;
+        let legacyShellArgs: string[] | undefined;
+        const removed = new Set(this.userProfileStore.all.map(([id, profile]) => id));
+        switch (OS.type()) {
```
msujew, line 303: Suggestion:

```suggestion
        switch (OS.backend.type()) {
```

packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -724,11 +934,43 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
...
+        this.open(termWidget, { widgetOptions: options });
+    }
+
+    protected async chooseDefaultProfile(): Promise<void> {
+        const result = await this.selectTerminalProfile(nls.localizeByDefault('Select your default terminal profile'));
+        if (!result) {
+            return;
+        }
+
+        this.preferenceService.set(`terminal.integrated.defaultProfile.${OS.type().toLowerCase()}`, result[0], PreferenceScope.User);
```
msujew, line 971: Suggestion:

```suggestion
        this.preferenceService.set(`terminal.integrated.defaultProfile.${OS.backend.type().toLowerCase()}`, result[0], PreferenceScope.User);
```

packages/terminal/src/browser/terminal-widget-impl.ts
```diff
@@ -487,9 +487,8 @@ export class TerminalWidgetImpl extends TerminalWidget implements StatefulWidget
         const { cols, rows } = this.term;
 
         const terminalId = await this.shellTerminalServer.create({
-            shellPreferences: this.shellPreferences,
-            shell: this.options.shellPath,
-            args: this.options.shellArgs,
+            shell: this.options.shellPath || this.shellPreferences.shell[OS.type()],
+            args: this.options.shellArgs || this.shellPreferences.shellArgs[OS.type()],
```
msujew, line 491:
```suggestion
            shell: this.options.shellPath || this.shellPreferences.shell[OS.backend.type()],
            args: this.options.shellArgs || this.shellPreferences.shellArgs[OS.backend.type()],
```

msujew (commented):
I noticed a few issues when the frontend OS isn't the same as the backend OS. The preferences don't match up well.

packages/terminal/src/browser/terminal-preferences.ts
vince-fugnitto: No problem, the extra whitespace.

packages/terminal/src/browser/terminal-frontend-contribution.ts
vince-fugnitto: I mean, these commands do not exist in VS Code, are they necessary?
Instead we should probably implement `Create New Terminal (With Profile)` which is present in VS Code and likely achieves what we are attempting here.

tsmaeder: Yes, the UI (in particular choosing the default profile) is very different from what VS Code does. As for the "Create New Terminal (With Profile)", I would argue that my solution is superior: the "..." at the end of "New Profile" is a well established UI pattern indicating that the user will face a dialog to do some configuration or selection before running the action.
Remember that "do the same as VS Code" is not a strategic goal of the project anymore.

packages/terminal/src/browser/terminal-frontend-contribution.ts
tsmaeder: That's not a suggestion: it's a bug.

packages/core/i18n/nls.de.json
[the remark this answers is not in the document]
tsmaeder: Thanks for the insight.

packages/terminal/src/browser/terminal-frontend-contribution.ts
tsmaeder: I don't really understand why "New Terminal" should not exist in VS Code: it exists and is under the "terminal" main menu. If you're saying we should use the same command ids, I'm not sure about that: are the command id's supposed to be API in VS Code?

packages/terminal/src/browser/terminal-preferences.ts
tsmaeder: Ah, I see.

packages/terminal/src/browser/terminal-preferences.ts
```diff
@@ -17,9 +17,58 @@
...
+
+const pathProperty: IJSONSchema = {
+    description: nls.localize('theia/terminal/profilePath', 'The path of the shell that this profile uses.'),
+    ...stringOrStringArray
+};
+
+function shellArgsDeprecatedMessage(type: OS.Type): string {
+    return nls.localize('theia/terminal/shell.deprecated', 'This is deprecated, the new recommended way to configure your default shell is by creating a '
+        + 'terminal profile in \'terminal.integrated.profiles.{0}\' and setting its profile name as the default in '
+        + '\'terminal.integrated.defaultProfile.{0}.\'', type.toString().toLowerCase());
```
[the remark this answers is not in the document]
tsmaeder: I understand that the default localization cannot handle a computed _key_, but shouldn't it just treat the default value as a random string?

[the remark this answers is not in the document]
tsmaeder: I mean: your statement would imply that we cannot default-localize any messages with parameters, right?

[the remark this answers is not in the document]
tsmaeder: Ah, I get it.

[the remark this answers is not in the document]
tsmaeder: Ah...you mean the `+`?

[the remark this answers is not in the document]
msujew: > Ah...you mean the `+`?

Yes, exactly :)

packages/terminal/src/browser/terminal-frontend-contribution.ts
vince-fugnitto: However in this case it adds entries in the `terminal` menu that downstream applications might not be expecting since in general we do follow the default implementation of VS Code. Moreover, plugins may attempt to trigger the `workbench.action.terminal.newWithProfile` command but it does not exist.

If we want to keep the commands that you've added on top of VS Code I'd maybe suggest we move them out of the `terminal` menu.

tsmaeder: > However in this case it adds entries in the terminal menu that downstream applications might not be expecting

Having the same menu structure as VS Code is not a promise I think we are making to our adopters. If downstream apps rely on that, we have a problem. Our UI with terminals and profiles is very different (no tabbed terminals, for example) and I feel it's a bit of a kludge in VS Code. It would be wrong to strive for visual compatibility in this respect, IMO.

> Moreover, plugins may attempt to trigger the workbench.action.terminal.newWithProfile command but it does not exist.

That's the question I'm asking: is this command id VS Code published API we should support?

tsmaeder: These are, though: https://code.visualstudio.com/api/references/commands

vince-fugnitto: > That's the question I'm asking: is this command id VS Code published API we should support? I don't think it is.

I believe it is the main reason that command ids follow VS Code, since they can be ultimately executed from plugins.
In the case that we do not match their id, we have a mapping performed in the plugin system, for example:
- https://github.com/eclipse-theia/theia/blob/d383fe25ee6dd2d0a3296222235b73e425666e2f/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts#L373-L375

I understand you might be referring to [builtin commands](https://code.visualstudio.com/api/references/commands), but the page also describes "common commands" and such commands are also accessible by plugins.

tsmaeder: Having a look at the broader picture here: all the other terminal-related commands do _not_ have the same command ids as in VS Code (e.g. `terminal:new` instead of `workbench.action.terminal.new`. So having just the two new one's aligned seems odd. In this PR, the command ids are consistent with the rest of the terminal commands. If we want to align the command ids with VS Code, it should be done in a separate PR that covers all terminal-related command ids.

tsmaeder: I've opened an issue: https://github.com/eclipse-theia/theia/issues/12084

msujew (approved):
The changes look good to me. I would wait to merge until @vince-fugnitto has given his approval as well 👍

vince-fugnitto (commented):
The changes work well for me when I verified the functionality, I only have a minor comment regarding the new menu item.

vince-fugnitto (approved):
The latest updates look good to me :+1: 
I confirmed that the functionality works well and as expected, and the new API is marked as supported.

packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -724,11 +933,43 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
...
         const cwd = await this.selectTerminalCwd();
-        const termWidget = await this.newTerminal({ cwd });
-        termWidget.start();
+        let profile = this.profileService.defaultProfile;
+
+        if (!profile) {
+            throw new Error('There are not profiles registered');
+        }
+        if (profile instanceof ShellTerminalProfile) {
+            profile = profile.modify({ rootURI: cwd });
```
colin-grant-work, line 942: @tsmaeder, I believe this should have just been `modify({cwd})`. `rootURI` is not a recognized field when these options are eventually passed into the TerminalWidgetFactory.

packages/terminal/src/browser/shell-terminal-profile.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Makes a copy of this profile modified with the options given
+     * as an argument.
+     * @param options the options to override
+     * @returns a modified copy of this profile
+     */
+    modify(options: IShellTerminalServerOptions): TerminalProfile {
+        return new ShellTerminalProfile(this.terminalService, { ...this.options, ...options });
+    }
```
colin-grant-work, line 39: This operation is mostly bogus, because `IShellTerminalServerOptions` (the function arg) and `TerminalWidgetOptions` (the constructor argument type) have very little in common.

tsmaeder:
I think this is a leftover from prototype I did where I separated the code paths for shell terminals and other terminals (`Pseudoterminal`).

PR #12067, opened 2023-01-13
Toolbars: ToolbarItem.when field is evaluated and respected

### LINKED ISSUE #11705: Toolbar items should be enabled/visible according to when-clause-contexts
### Feature Description:

The `TabBarToolbarItem` interface for items for the the main toolbar offers an optional `when` property, that is supposed to accept [when-clause-contexts](https://code.visualstudio.com/docs/getstarted/keybindings#_when-clause-contexts).
This can be set, but it has no impact on the toolbar item.
Also, editing the user's `toolbar.json` directly in the application does not allow to set such a clause:
image

I would like to see items to be enabled/visible only if the `when` clause returns `true` (e.g. only show item if the workbench has a valid workspace: `when: 'workbenchState != empty'`).

jonah-iden:
ToolbarItem.when field is evaluated and respected
by toolbar and tab-bar-toolbar +
when clause can be set in toolbar.json

#### What it does
Implements/fixes #11705

This PR implements the funcitionality for toolbars to evaluate and respect the when field/clause of `ConditionalToolbarItem`.
Also allows setting the when field in `toolbar.json`.

#### How to test
easiest way to test is probably through the toolbar.json file:

1. press Alt+T to open the toolbar 
2. right click it and click Customize Toolbar (open JSON)
3. either add a new toolbar item or edit an existing one. Add "when": "inDebugMode" and save the file
4. see the icon gets grayed out and does not react to clicks
5. start a debug session
6. see icon is white again and reacts to clicks
7. stop debug session
8. icon disabled again

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
+++ b/packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
@@ -246,7 +246,7 @@ export class TabBarToolbar extends ReactWidget {
         return this.commands.isToggled(command, this.current);
     }
 
-    protected evalualteWhenClause(whenClause: string | undefined): boolean {
+    protected evaluateWhenClause(whenClause: string | undefined): boolean {
         return whenClause ? this.contextKeyService.match(whenClause) : true;
     }
 
--- a/packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
+++ b/packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
@@ -65,7 +65,7 @@ export class TabBarToolbar extends ReactWidget {
         this.inline.clear();
         this.more.clear();
 
-        const contextKeys: Set<string> = new Set();
+        const contextKeys = new Set<string>();
         for (const item of items.sort(TabBarToolbarItem.PRIORITY_COMPARATOR).reverse()) {
             if ('render' in item || item.group === undefined || item.group === 'navigation') {
                 this.inline.set(item.id, item);
@@ -149,7 +149,7 @@ export class TabBarToolbar extends ReactWidget {
         }
         const tooltip = item.tooltip || (command && command.label);
 
-        const toolbarItemClassNames = this.getToolbarItemClassNames(command?.id ?? item.command, this.evaluateWhenClause(item.when));
+        const toolbarItemClassNames = this.getToolbarItemClassNames(item);
         if (item.menuPath && !item.command) { toolbarItemClassNames.push('enabled'); }
         return <div key={item.id}
             ref={this.onRender}
@@ -164,13 +164,13 @@ export class TabBarToolbar extends ReactWidget {
         </div>;
     }
 
-    protected getToolbarItemClassNames(commandId: string | undefined, whenClauseResult: boolean): string[] {
+    protected getToolbarItemClassNames(item: AnyToolbarItem): string[] {
         const classNames = [TabBarToolbar.Styles.TAB_BAR_TOOLBAR_ITEM];
-        if (commandId) {
-            if (this.commandIsEnabled(commandId) && whenClauseResult) {
+        if (item.command) {
+            if (this.commandIsEnabled(item.command) && this.evaluateWhenClause(item.when)) {
                 classNames.push('enabled');
             }
-            if (this.commandIsToggled(commandId)) {
+            if (this.commandIsToggled(item.command)) {
                 classNames.push('toggled');
             }
         }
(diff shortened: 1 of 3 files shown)


packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
```diff
@@ -221,17 +246,23 @@ export class TabBarToolbar extends ReactWidget {
         return this.commands.isToggled(command, this.current);
     }
 
+    protected evalualteWhenClause(whenClause: string | undefined): boolean {
```
msujew, line 249: Typo:

```suggestion
    protected evaluateWhenClause(whenClause: string | undefined): boolean {
```

packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
```diff
@@ -60,13 +64,22 @@ export class TabBarToolbar extends ReactWidget {
     updateItems(items: Array<TabBarToolbarItem | ReactTabBarToolbarItem>, current: Widget | undefined): void {
         this.inline.clear();
         this.more.clear();
+
+        const contextKeys: Set<string> = new Set();
```
vince-fugnitto, line 68: minor: we can simplify and infer the type

```suggestion
        const contextKeys = new Set<string>();
```

packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
```diff
@@ -139,10 +164,10 @@ export class TabBarToolbar extends ReactWidget {
         </div>;
     }
 
-    protected getToolbarItemClassNames(commandId: string | undefined): string[] {
+    protected getToolbarItemClassNames(commandId: string | undefined, whenClauseResult: boolean): string[] {
```
vince-fugnitto, line 167: We should probably pass `item: AnyToolBarItem` instead of `commandId` and `whenClauseResult`.

packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
```diff
@@ -221,17 +246,23 @@ export class TabBarToolbar extends ReactWidget {
...
-        } else if (item?.menuPath) {
-            this.renderMoreContextMenu(this.toAnchor(e), item.menuPath);
+        if (this.evaluateWhenClause(item?.when)) {
+            if (item?.command && item.menuPath) {
+                this.menuCommandExecutor.executeCommand(item.menuPath, item.command, this.current);
+            } else if (item?.command) {
+                this.commands.executeCommand(item.command, this.current);
+            } else if (item?.menuPath) {
+                this.renderMoreContextMenu(this.toAnchor(e), item.menuPath);
+            }
```
vince-fugnitto, line 265: We can avoid all these changes, and the nested if statement by performing the opposite logic which is also clearer:

```ts
if (!this.evaluateWhenClause(item?.when)) {
    return;
}

...
```

packages/toolbar/src/browser/toolbar.tsx
```diff
@@ -77,13 +77,20 @@ export class ToolbarImpl extends TabBarToolbar {
     protected updateInlineItems(): void {
         this.inline.clear();
         const { items } = this.model.toolbarItems;
+
+        const contextKeys: Set<string> = new Set();
```
vince-fugnitto, line 81: minor:

```suggestion
        const contextKeys = new Set<string>();
```

packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
jonah-iden: but if we pass in just the item wouldn't we have to do the command extraction at line 144 again?

vince-fugnitto: > but if we pass in just the item wouldn't we have to do the command extraction at line 144 again?

No, we would just use `item.command` and `item.when`?

jonah-iden: oh ok. Im sorry i didn't realize that `item.command` and `command.id` would be the same value. I'll do that. thanks for clarifiying

msujew (approved):
The changes look good to me 👍 

- [x] Registering a `when` context correctly updates the toolbar item (disables/enables it)
- [x] When the context changes, the item is also updated as expected
- [x] A disabled toolbar item will not trigger its command when pressed (though it still allows to remove/move the item)
- [x] The `string` type for the `when` property is correctly reflected when editing the JSON file

vince-fugnitto:
For transparency the pull-request has caused regressions with other toolbar items such as https://github.com/eclipse-theia/theia/issues/12314.

alvsan09:
Another regression is described at: 
https://github.com/eclipse-theia/theia/issues/12314#issuecomment-1474224067

Issue #11898, opened 2022-11-21
Use "forEach" instead of "map" in search-in-workspace-result-tree-widget#updateRightResults


JonasHelming:
**Blocked for Hackathon**

The "map" function is used to iterate over an array, but its return value is not used. In this case, it is faster and also less confusing to use "forEach".

See https://github.com/eclipse-theia/theia/blob/700c25ca2cbbae5e10dd240a33e447b57d9af0c5/packages/search-in-workspace/src/browser/search-in-workspace-result-tree-widget.tsx#L845

Operating System: All
Theia Version: Latest

Steps to test the behavior:

1. Open an example file with the same search String appearing multiple times in one row, e.g. "test foobar test"
2. Use the Theia search view to search for the test String occurring in the opened file
3. Enter a replace value
4. In the search results, select to replace the first occurrence of the search String only
5. Check that the second hit is correctly highlighted in the editor after the replacement
6. Replace the second hit in the same line

vince-fugnitto:
Closed thanks to https://github.com/eclipse-theia/theia/pull/12071.

Issue #11897, opened 2022-11-21
Use "forEach" instead of "map" in search-in-workspace-result-tree-widget#updateCurrentEditorDecorations


JonasHelming:
**Blocked for Hackathon**

The "map" function is used to iterate over an array, but its return value is not used. In this case, it is faster and also less confusing to use "forEach".

See https://github.com/eclipse-theia/theia/blob/700c25ca2cbbae5e10dd240a33e447b57d9af0c5/packages/search-in-workspace/src/browser/search-in-workspace-result-tree-widget.tsx#L633

- Operating System: All
- Theia Version: Latest

Steps to test the behavior:

1. Open two examples file with occurrences of the same test String
2. Search for this String in the Theia search view
3. Check that the search String is highlighted in both editors

There should be no functional change when replacing map with forEach

vince-fugnitto:
Closed thanks to https://github.com/eclipse-theia/theia/pull/12071.

PR #12071, opened 2023-01-13
searchInWorkspaceResult: `.map` replaced for `.forEach`

FernandoAscencio:
This commit replaces the the `.map` in several lines for a `.forEach`.

Signed-off-by: FernandoAscencio <fernando.ascencio.cama@ericsson.com>

#### What it does

Implements the suggested change in #11897 and #11898

#### How to test
For #11897:
1. Open two examples file with occurrences of the same test String
2. Search for this String in the Theia search view
3. Check that the search String is highlighted in both editors
For #11898:
1. Open an example file with the same search String appearing multiple times in one row, e.g. "test foobar test"
2. Use the Theia search view to search for the test String occurring in the opened file
3. Enter a replace value
4. In the search results, select to replace the first occurrence of the search String only
5. Check that the second hit is correctly highlighted in the editor after the replacement
6. Replace the second hit in the same line

As mentioned, there should be no difference in behavior from before the changes.

#### Review checklist

#### Reminder for reviewers

- [x] As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)


vince-fugnitto (approved):
LGTM :+1:

PR #12088, opened 2023-01-19
Support multi-selection of tree items in tree views

### LINKED ISSUE #9074: Support multi-selection in Tree View
Support multi-selection of tree items in Tree View.

### Feature Description:
Feature should allow selecting more than 1 item in tree view using Ctrl + Click on the relevant items. 

Can be added as a property of **TreeViewOptions** like in vscode api: 

**canSelectMany?: boolean**  

Whether the tree supports multi-select. When the tree supports multi-select and a command is executed from the tree, the first argument to the command is the tree item that the command was executed on and the second argument is an array containing all selected tree items.

tsmaeder:
#### What it does
Adds multi-selection support to tree views contributed via VS Code API

Fixes #9074

#### How to test
The attached extension contributes a tree view that has multi-select enabled. There is a command "Print Selection" which prints out the selected items on the extension side. Make sure the behavior is consistent with VS Code.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
+++ b/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
@@ -718,7 +718,7 @@ export class TreeViewWidget extends TreeViewWelcomeWidget {
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     protected renderInlineCommand(actionMenuNode: ActionMenuNode, index: number, tabbable: boolean, args: any[]): React.ReactNode {
-        if (!actionMenuNode.icon || !this.commands.isVisible(actionMenuNode.command, args) || !actionMenuNode.when || !this.contextKeys.match(actionMenuNode.when)) {
+        if (!actionMenuNode.icon || !this.commands.isVisible(actionMenuNode.command, ...args) || !actionMenuNode.when || !this.contextKeys.match(actionMenuNode.when)) {
             return false;
         }
         const className = [TREE_NODE_SEGMENT_CLASS, TREE_NODE_TAIL_CLASS, actionMenuNode.icon, ACTION_ITEM, 'theia-tree-view-inline-action'].join(' ');
--- a/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
+++ b/packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
@@ -15,7 +15,7 @@
 // *****************************************************************************
 
 import { injectable, inject, postConstruct } from '@theia/core/shared/inversify';
-import { TreeViewsExt, TreeViewItemCollapsibleState, TreeViewItem, TreeViewSelection, ThemeIcon, DataTransferFileDTO } from '../../../common/plugin-api-rpc';
+import { TreeViewsExt, TreeViewItemCollapsibleState, TreeViewItem, TreeViewItemReference, ThemeIcon, DataTransferFileDTO } from '../../../common/plugin-api-rpc';
 import { Command } from '../../../common/plugin-api-rpc-model';
 import {
     TreeNode,
@@ -164,6 +164,7 @@ export namespace CompositeTreeViewNode {
 @injectable()
 export class TreeViewWidgetOptions {
     id: string;
+    multiSelect: boolean | undefined;
     dragMimeTypes: string[] | undefined;
     dropMimeTypes: string[] | undefined;
 }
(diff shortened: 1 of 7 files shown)


tsmaeder:
Here's the extension:

[custom-view-samples-0.0.1.zip](https://github.com/eclipse-theia/theia/files/10458482/custom-view-samples-0.0.1.zip)

And here's the source:

[tree-view-sample.zip](https://github.com/eclipse-theia/theia/files/10458485/tree-view-sample.zip)

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
```diff
@@ -549,38 +550,42 @@ export class TreeViewWidget extends TreeViewWelcomeWidget {
...
 
-    protected override toContextMenuArgs(node: SelectableTreeNode): [TreeViewSelection] {
-        return [this.toTreeViewSelection(node)];
+    protected override toContextMenuArgs(target: SelectableTreeNode): [TreeViewItemReference, TreeViewItemReference[]] | [TreeViewItemReference] {
+        if (this.options.multiSelect) {
+            return [this.toTreeViewItemReference(target), this.model.selectedNodes.map(node => this.toTreeViewItemReference(node))];
+        } else {
+            return [this.toTreeViewItemReference(target)];
+        }
     }
```
paul-marechal, line 589: I was unable to understand which handler would consume the returned values... Ultimately they seem to be forwarded to `this.commands.executeCommand(node.command, ...args);` line 579 of the same file but that doesn't help me either.

It doesn't seem like your sample VS Code Extension contributes any inline command for us to test this code, right?

tsmaeder: Command arguments are described [here](https://code.visualstudio.com/api/references/vscode-api#TreeViewOptions&lt;T&gt;). The "Print Selection" from the extension command spews out the command arguments to the console.

paul-marechal: I don't see the documentation about said arguments in the link you shared...

I finally managed to find the equivalent code in VS Code that may explain what context menu items expect to invoke commands with: https://github.com/microsoft/vscode/blob/d5379b5e34e4bb0f5ddb33688da91f787ebc159f/src/vs/workbench/browser/parts/views/treeView.ts#L1192-L1210

paul-marechal (approved):
The changes work as described when using the provided extension, although I made some modifications to make sure that inline actions also work properly:

I added the following menu entry in the extension's `package.json`:

```json
        {
          "command": "treeViewDnD.printSelection",
          "when": "view == testViewDragAndDrop && viewItem == aaaaaa",
          "group": "inline"
        }
```

Along with editing the following code:

```ts
// file: out/testViewDragAndDrop.js
// class: TestViewDragAndDrop

    _getTreeItem(key) {
        const treeElement = this._getTreeElement(key);
        // An example of how to use codicons in a MarkdownString in a tree item tooltip.
        const tooltip = new vscode.MarkdownString(`$(zap) Tooltip for ${key}`, true);
        return {
            // required for `viewItem == X` to work:
            contextValue: key,
            label: /**vscode.TreeItemLabel**/ { label: key, highlights: key.length > 1 ? [[key.length - 2, key.length - 1]] : void 0 },
            tooltip,
            collapsibleState: treeElement && Object.keys(treeElement).length ? vscode.TreeItemCollapsibleState.Collapsed : vscode.TreeItemCollapsibleState.None,
            resourceUri: vscode.Uri.parse(`/tmp/${key}`),
        };
    }
```

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
tsmaeder: TreeViewOptions.canSelectMany doc says: 

>Whether the tree supports multi-select. When the tree supports multi-select and a command is executed from the tree, the first argument to the command is the tree item that the command was executed on and the second argument is an array containing all selected tree items.

tsmaeder: I'm not sure what you mean by "inline commands".

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
```diff
@@ -696,41 +697,44 @@ export class TreeViewWidget extends TreeViewWelcomeWidget {
         }
     }
 
-    protected override renderTailDecorations(node: TreeViewNode, props: NodeProps): React.ReactNode {
-        return this.contextKeys.with({ view: this.id, viewItem: node.contextValue }, () => {
+    protected override renderTailDecorations(treeViewNode: TreeViewNode, props: NodeProps): React.ReactNode {
```
tsmaeder, line 700: It's common practice to encode the type of a variable in dynamically typed languages like Smalltalk, wher you would call a parameter `aString`. IMO, it's just redundant information in a language like Typescript. I wouldn't want to make it a pattern.

tsmaeder:
@paul-marechal since you changed some code and I assume you tested it, I'll let you merge the changes.

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
paul-marechal: I understand, but it genuinely got confusing when reading the code where multiple types of nodes are being handled in the same general area. I think future readers will benefit from it in this instance.

paul-marechal:
@tsmaeder most of my changes were trivial, though there was one potential bug fixed by https://github.com/eclipse-theia/theia/pull/12088/commits/b610af8f6403d66c48068ab1454ff47d869cef4c.

I confirm that I manually tested the changes and it still works as expected.

PR #12094, opened 2023-01-20
repo: update builtins to `1.64.2`, `1.66.2`

vince-fugnitto:
#### What it does

The pull-request updates the repository's **builtins** to their latest version (`1.64.2`).
Note:  work to publish and use `1.68.1` is still underway.

#### How to test

- Confirm that CI passes

#### Review checklist

#### Reminder for reviewers


colin-grant-work:
I have misgivings about lifting the default supported API to a level that includes code we know to be stubbed. As an application developer, I would not want users downloading plugins when I don't know how stubbed API is actually being used. Much of the stubbed API is fairly self-contained - Test API, Notebook API - so there's a good chance that those stubs won't interfere with functionality the user is accustomed to, but other pieces are not so isolated. For example, if a plugin switches from the `visibleEditors` API to a `TabGroup`-based approach, it will just fail, and the user won't have much an idea why. I don't want to have to debug those cases and support them as users encounter them one by one (issues * users), so I would (and will) just set the compatibility level to whatever I can be reasonably confident is actually supported.

So I wonder whether, with the advent of stubbing, we should introduce different concepts of compatibility. In the context of Open VSX, the lower level - real support - could be used for the default download and the higher level - subbed support - could be used for the 'Install another version command,' so that users can be held responsible for their decision there.

marcdumais-work (approved):
LGTM, thanks Vince!

PR #12120, opened 2023-01-27
Use tree item path for tree node id. Fixes #12111

### LINKED ISSUE #12111: TreeView `refresh` Is Broken When Nodes Are Reparented
### Bug Description:
The refresh in TreeView contributed by extensions is broken in the case of node being reparented. In order to reproduce this, you can use the test extension attached to https://github.com/eclipse-theia/theia/pull/12088

### Steps to Reproduce:

1. Open the Test DnD View as per the lined PR
2. Drag "bb" on "aaa"
3. Drag "bb" on "b"
4. Observe: Theia goes into an endless loop.

### Additional Information

- Operating System:
- Theia Version:

tsmaeder:
- Since our front end tree is not safe for reparenting nodes, use the path of tree item ids as the tree node id.
- Rename TreeViewExt.getTreeItem to getElement, since that is what it does.

#### What it does
Fixes #12111
Since the Theia front end tree cannot handle reparenting tree nodes with the same id, this PR fixes the problem for tree views contributed via the VS Code API. The idea is to concatenate the id of the parent `TreeItem` with the id of the tree item itself to use as the id of the `TreeViewItem` that is sent to the front end. Thus, the id of the tree node in the front end will be different when being reparented, thus working around the problem.

#### How to test
Install the extension from https://github.com/eclipse-theia/theia/pull/12065#issuecomment-1380392900 and test that elements can be properly reparented and that selection and expansion behaviour is as expected. Note that, since the id changes when reparenting, an item that is dragged to a different folder will not be selected after the move. I consider this an acceptable trade-off considering the current state may lead to an endless loop. I was not able to think of a fix that would avoid this without thoroughly refactoring the tree subsystem.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/tree/tree-views.ts
+++ b/packages/plugin-ext/src/plugin/tree/tree-views.ts
@@ -315,13 +315,9 @@ class TreeViewExtImpl<T> implements Disposable {
             // parent is inconsistent
             return undefined;
         }
-        let possibleIndex = children.length;
-        // find the right element id by searching all possible id names in the cache
-        while (possibleIndex-- > 0) {
-            const candidateId = this.buildTreeItemId(parentId, treeItem);
-            if (this.nodes.has(candidateId)) {
-                return chain.concat(candidateId);
-            }
+        const candidateId = this.buildTreeItemId(parentId, treeItem);
+        if (this.nodes.has(candidateId)) {
+            return chain.concat(candidateId);
         }
         // couldn't calculate consistent parent chain and id
         return undefined;
--- a/packages/plugin-ext/src/plugin/tree/tree-views.ts
+++ b/packages/plugin-ext/src/plugin/tree/tree-views.ts
@@ -34,7 +34,6 @@ import { URI } from '@theia/core/shared/vscode-uri';
 import { UriComponents } from '@theia/core/lib/common/uri';
 
 export class TreeViewsExtImpl implements TreeViewsExt {
-
     private proxy: TreeViewsMain;
 
     private readonly treeViews = new Map<string, TreeViewExtImpl<any>>();
@@ -45,9 +44,9 @@ export class TreeViewsExtImpl implements TreeViewsExt {
         commandRegistry.registerArgumentProcessor({
             processArgument: arg => {
                 if (TreeViewItemReference.is(arg)) {
-                    return this.toTreeItem(arg);
+                    return this.toTreeElement(arg);
                 } else if (Array.isArray(arg)) {
-                    return arg.map(param => TreeViewItemReference.is(param) ? this.toTreeItem(param) : param);
+                    return arg.map(param => TreeViewItemReference.is(param) ? this.toTreeElement(param) : param);
                 } else {
                     return arg;
                 }
@@ -62,8 +61,8 @@ export class TreeViewsExtImpl implements TreeViewsExt {
         return this.getTreeView(treeViewId).handleDrop!(treeItemId, dataTransferItems, token);
     }
 
-    protected toTreeItem(treeViewItemRef: TreeViewItemReference): any {
-        return this.treeViews.get(treeViewItemRef.viewId)?.getTreeItem(treeViewItemRef.itemId);
+    protected toTreeElement(treeViewItemRef: TreeViewItemReference): any {
+        return this.treeViews.get(treeViewItemRef.viewId)?.getElement(treeViewItemRef.itemId);
     }
 
     registerTreeDataProvider<T>(plugin: Plugin, treeViewId: string, treeDataProvider: TreeDataProvider<T>): PluginDisposable {
@@ -186,6 +185,10 @@ interface TreeExtNode<T> extends Disposable {
 }
 
 class TreeViewExtImpl<T> implements Disposable {
+    private static readonly ID_COMPUTED = 'c';
+    private static readonly ID_ITEM = 'i';
+
+    private nextItemId: 0;
 
     private readonly onDidExpandElementEmitter = new Emitter<TreeViewExpansionEvent<T>>();
     readonly onDidExpandElement = this.onDidExpandElementEmitter.event;
@@ -274,7 +277,7 @@ class TreeViewExtImpl<T> implements Disposable {
         this.proxy.$setDescription(this.treeViewId, this._description);
     }
 
-    getTreeItem(treeItemId: string): T | undefined {
+    getElement(treeItemId: string): T | undefined {
         return this.nodes.get(treeItemId)?.value;
     }
 
@@ -312,11 +315,10 @@ class TreeViewExtImpl<T> implements Disposable {
             // parent is inconsistent
             return undefined;
         }
-        const idLabel = this.getTreeItemIdLabel(treeItem);
         let possibleIndex = children.length;
         // find the right element id by searching all possible id names in the cache
         while (possibleIndex-- > 0) {
-            const candidateId = this.buildTreeItemId(parentId, possibleIndex, idLabel);
+            const candidateId = this.buildTreeItemId(parentId, treeItem);
             if (this.nodes.has(candidateId)) {
                 return chain.concat(candidateId);
             }
@@ -325,7 +327,7 @@ class TreeViewExtImpl<T> implements Disposable {
         return undefined;
     }
 
-    private getTreeItemLabel(treeItem: TreeItem): string | undefined {
+    private getLabelString(treeItem: TreeItem): string | undefined {
         const treeItemLabel: string | TreeItemLabel | undefined = treeItem.label;
         return typeof treeItemLabel === 'object' ? treeItemLabel.label : treeItemLabel;
     }
@@ -335,8 +337,8 @@ class TreeViewExtImpl<T> implements Disposable {
         return typeof treeItemLabel === 'object' ? treeItemLabel.highlights : undefined;
     }
 
-    private getTreeItemIdLabel(treeItem: TreeItem): string | undefined {
-        let idLabel = this.getTreeItemLabel(treeItem);
+    private getItemLabel(treeItem: TreeItem): string | undefined {
+        let idLabel = this.getLabelString(treeItem);
         // Use resource URI if label is not set
         if (idLabel === undefined && treeItem.resourceUri) {
             idLabel = treeItem.resourceUri.path.toString();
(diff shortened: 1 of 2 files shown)


packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -312,20 +315,15 @@ class TreeViewExtImpl<T> implements Disposable {
...
-        let possibleIndex = children.length;
-        // find the right element id by searching all possible id names in the cache
-        while (possibleIndex-- > 0) {
-            const candidateId = this.buildTreeItemId(parentId, possibleIndex, idLabel);
-            if (this.nodes.has(candidateId)) {
-                return chain.concat(candidateId);
-            }
+        const candidateId = this.buildTreeItemId(parentId, treeItem);
+        if (this.nodes.has(candidateId)) {
+            return chain.concat(candidateId);
```
paul-marechal, line 320: Does this new code mean we'll never hit the cache for tree items for which we compute an ID?

Before the logic seemed to assume we could find items from some `possibleIndex`, but now there's a case where `buildTreeItemId` generates a new unique ID which means that those items will never be found here.

tsmaeder: Only if the tree item does not have an id set. This is correct, in my opinion, because if it doesn't have an id, there's no way to tell it's the same tree item, so any state we might have for the tree item at index n might not apply to the new item at index n.

colin-grant-work:
> Since the Theia front end tree cannot handle reparenting tree nodes with the same id...

This seems unfortunate. Can you give an idea of the code to reparent a node that you think should work (in an ideal world), but doesn't?

tsmaeder:
> > Since the Theia front end tree cannot handle reparenting tree nodes with the same id...
> 
> This seems unfortunate. Can you give an idea of the code to reparent a node that you think should work (in an ideal world), but doesn't?

Any time you use the same id with a different parent, think dragging a file to a different folder (the "real" file explorer uses the full pah as a key, which is what I'm doing in this PR). The problem is described in https://github.com/eclipse-theia/theia/issues/12111

tsmaeder:
@paul-marechal @marcdumais ping?

packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -62,8 +61,8 @@ export class TreeViewsExtImpl implements TreeViewsExt {
         return this.getTreeView(treeViewId).handleDrop!(treeItemId, dataTransferItems, token);
     }
 
-    protected toTreeItem(treeViewItemRef: TreeViewItemReference): any {
-        return this.treeViews.get(treeViewItemRef.viewId)?.getTreeItem(treeViewItemRef.itemId);
+    protected toTreeElement(treeViewItemRef: TreeViewItemReference): any {
```
colin-grant-work, line 64: Is there a reason for this name change? In principle, it's breaking, so if it isn't necessary, may be best to stick with the `item` name.

packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -274,7 +277,7 @@ class TreeViewExtImpl<T> implements Disposable {
         this.proxy.$setDescription(this.treeViewId, this._description);
     }
 
-    getTreeItem(treeItemId: string): T | undefined {
+    getElement(treeItemId: string): T | undefined {
```
colin-grant-work, line 280: Same here re: naming.

packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -312,20 +315,15 @@ class TreeViewExtImpl<T> implements Disposable {
...
+        const candidateId = this.buildTreeItemId(parentId, treeItem);
+        if (this.nodes.has(candidateId)) {
+            return chain.concat(candidateId);
         }
         // couldn't calculate consistent parent chain and id
         return undefined;
     }
 
-    private getTreeItemLabel(treeItem: TreeItem): string | undefined {
+    private getLabelString(treeItem: TreeItem): string | undefined {
```
colin-grant-work, line 326: The `String` part of this name is redundant, given the typing.

packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -348,8 +346,18 @@ class TreeViewExtImpl<T> implements Disposable {
         return idLabel;
     }
 
-    private buildTreeItemId(parentId: string, index: number, idLabel: string | undefined): string {
-        return `${parentId}/${index}:${idLabel}`;
+    private buildTreeItemId(parentId: string, item: TreeItem): string {
+        // build tree id according to https://code.visualstudio.com/api/references/vscode-api#TreeItem
```
colin-grant-work, line 350: Do those docs prescribe or describe any particular form for the ID or any algorithm for its derivation? I only see that they ID should be a string.

packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -348,8 +346,18 @@ class TreeViewExtImpl<T> implements Disposable {
...
+        // note: the front end tree implementation cannot handle reparenting items, hence the id is set to the "path" of individual ids
+        let id = typeof item.id === 'string' ? item.id : this.getItemLabel(item);
+        if (id) {
+            // we use '' as the id of the root, we don't consider that a valid id
+            // since '' is falsy, we'll never get '' in this branch
+            id = TreeViewExtImpl.ID_ITEM + id;
+        } else {
+            id = TreeViewExtImpl.ID_COMPUTED + this.nextItemId++;
+        }
+        return `${parentId}/${id}`;
```
colin-grant-work, line 360: Since we're always adding additional material, wether we find an existing ID or not, if we have a long-lived tree, will it gradually accrete infinitely long ID's, or is there some guarantee that we'll start fresh at some point?

colin-grant-work (commented):
Functionally, things appear to be working. The only missing element there, not related to this PR, is that we get no drag-over highlight on tree nodes when dragging over them.

I have some other commends and questions about the code.

packages/plugin-ext/src/plugin/tree/tree-views.ts
tsmaeder: Yes, the method is simply misnamed. It does not convert to a TreeItem, it converts to the underlying domain element. I don't really see this class as part of the Theia API: first, it's part of the VS Code API implementation, which is determined by external factors, second, it's not something you can replace with inversify. So for me, the correct name trumps any compatibility concerns.

packages/plugin-ext/src/plugin/tree/tree-views.ts
tsmaeder: And same answer ;-)

packages/plugin-ext/src/plugin/tree/tree-views.ts
tsmaeder: Also, the documentation of `TreeItem` says:

>If not provided, an id is generated using the tree item's label. Note that when labels change, ids will change and that selection and expansion state cannot be kept stable anymore.

If you look at the way id's are constructed, the id's will be the same for items that have the same label.

packages/plugin-ext/src/plugin/tree/tree-views.ts
tsmaeder: The algorithm for computing the label is given here: https://code.visualstudio.com/api/references/vscode-api#TreeItem (label)
If there is a label string given in the item, it is used, otherwise try computing something based on the URI. The "String" is to differentiate the method from `getItemLabel`, not to indicate typing.

packages/plugin-ext/src/plugin/tree/tree-views.ts
tsmaeder: Yes, if you look at the doc for the id field. 

>  Optional id for the tree item that has to be unique across tree. The id is used to preserve the selection and expansion state of the tree item.
If not provided, an id is generated using the tree item's label. Note that when labels change, ids will change and that selection and expansion state cannot be kept stable anymore.

packages/plugin-ext/src/plugin/tree/tree-views.ts
tsmaeder: A 64 bit number maxes out at around 20 digits, so the individual id will be max 21 chars. Since we're using paths, the average path length will by something like 21 * log(n) (n= number of items). With 100000 items _visible_ in the tree, we get 1000000*5*21 which is somewhere around 10mb. Should be fine.

Note that the id here is not the id of of the TreeItem, so the id's don't get longer.

packages/plugin-ext/src/plugin/tree/tree-views.ts
colin-grant-work: In that case, please add a notice to the changelog indicating the potential breakage.

packages/plugin-ext/src/plugin/tree/tree-views.ts
colin-grant-work: I see it now - the color they use for the expansion toggles was light enough that I believed that it was a 'disabled' color.

packages/plugin-ext/src/plugin/tree/tree-views.ts
colin-grant-work: Ah - the initial ID is of the API item, not a known TreeNode.

packages/plugin-ext/src/plugin/tree/tree-views.ts
colin-grant-work: If the aim is to disambiguate those two methods, then the names used should reflect the different purposes. Both return `string | undefined`, so `getLabelString` doesn't help in that regard.

tsmaeder: We'll have to disagree that this change makes the code clearer. If you don't mind, I'll just undo my rename of the method.

colin-grant-work (approved):
Allows nodes to be moved among parents smoothly 👍

packages/plugin-ext/src/plugin/tree/tree-views.ts
colin-grant-work, line 360: @tsmaeder, this turns out to be a dangerous way to build the ID. If `getItemLabel` returns the same thing for several children of the same item - we've seen this downstream in the references view when the lines where something is referred to are identical - then this produces children with the exact same ID's, which isn't permitted in our core `Tree` implementation. Should be able to reproduce in Theia in a moment.

PR #12122, opened 2023-01-29
[scripts] Fixed the comparison of package references

inlann:
#### What it does

If there are some dependencies in the package.json but developer forgot to add them into the `references` array of the tsconfig.json, the `compile:references` will not add these dependencies into the tsconfig.json. This can lead to build failures due to errors in the compile order of the modules at build time.

image

I think we should add a check instead of requiring developers to ensure references in tsconfig.json are written correctly. So, I added `expectedReferences.some((reference, index) => currentReferences[index] !== reference)` to fix the generation of the tsconfig.json

#### How to test

Run `yarn compile:references`

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/scripts/compile-references.js
+++ b/scripts/compile-references.js
@@ -126,7 +126,10 @@ async function configureTypeScriptReferences(targetPackage, expectedReferences) 
     /** @type {string[]} */
     const currentReferences = (tsconfigJson['references'] || []).map(reference => reference.path);
     // Compare both arrays: if an element is not the same we need to rewrite.
-    needRewrite = needRewrite || currentReferences.some((reference, index) => expectedReferences[index] !== reference);
+    needRewrite = needRewrite
+        || currentReferences.some((reference, index) => expectedReferences[index] !== reference)
+        || expectedReferences.some((reference, index) => currentReferences[index] !== reference);
+
     if (needRewrite) {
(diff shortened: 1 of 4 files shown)


scripts/compile-references.js
```diff
@@ -126,7 +126,10 @@ async function configureTypeScriptReferences(targetPackage, expectedReferences)
     /** @type {string[]} */
     const currentReferences = (tsconfigJson['references'] || []).map(reference => reference.path);
     // Compare both arrays: if an element is not the same we need to rewrite.
-    needRewrite = needRewrite || currentReferences.some((reference, index) => expectedReferences[index] !== reference);
+    needRewrite = needRewrite
+        || currentReferences.some((reference, index) => expectedReferences[index] !== reference)
+        || expectedReferences.some((reference, index) => currentReferences[index] !== reference);
+
```
paul-marechal, line 132: Nice catch, the previous logic didn't properly compare if the two arrays were equal.

Though I would suggest using the following logic instead:

```suggestion
    needRewrite = needRewrite
        || currentReferences.length !== expectedReferences.length
        || currentReferences.some((reference, index) => expectedReferences[index] !== reference);
```

inlann:
@paul-marechal Thanks for your review 😉 👍🏼

paul-marechal (approved):
LGTM

PR #12144, opened 2023-02-03
Fix leak in File System Watcher

### LINKED ISSUE #12001: Potential filewatcher memory leak
### Bug Description:
The memory consumption of a Theia application steadily grows over time even if the application is in idle state.
We first discovered this issue with a deployed Theia-blueprint instance. After opening the workspace we kept the application in idle state for ~10 hours (apart from occasionally executing "top" in the terminal).
Over this timeperiod the memory consumption linearly increased from 0.18 Gib t o 0.72 Gib:

demo

The issue can also be reproduced with the Theia example application.
After further investigating this it turns out that the culprit is the Theia filewatcher process. All of the memory overhead is allocated by this process.

It has to be tested whether the issue also occurs if the file watcher is not running in a child process (--no-cluster).

### Steps to Reproduce:

1. Start the browser example application via terminal (or explicitly set the --no-cluster option)
2. Note down the memory consumption of the Theia processes with special focus to the file watcher process.
   (On linux i did this with `top -bc |grep nsfwOptions` )
3. Let the application idling but don't close the browser
4. After some time measure the memory consumption again. Should have increased slowly but steadily.

### Additional Information

- Operating System: Linux
- Theia Version: 1.32.0

tsmaeder:

#### What it does
The code was calling Promise.race() repeatedly while waiting for a file to come into existence. This added a `PromiseReaction` object each time, which never was collected.

Fixes #12001

#### How to test
Make sure file system watchers are still started correctly: in particular for

1. Open editors
2. Preferences, Launch configs, etc.
3. Workspace folders

Since the fix concerns file that don't exist in the beginning, it makes sense cases like when I add a new file `.theia/launch.json` and expect the launch configs to show up and changes to the file to be tracked.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/node/nsfw-watcher/nsfw-filesystem-service.ts
+++ b/packages/filesystem/src/node/nsfw-watcher/nsfw-filesystem-service.ts
@@ -197,24 +197,32 @@ export class NsfwWatcher {
     }
 
     /**
+     * @throws with {@link WatcherDisposal} if this instance is disposed.
+     */
+    protected assertNotDisposed(): void {
+        if (this.disposed) {
+            throw WatcherDisposal;
+        }
+    }
+
+    /**
      * When starting a watcher, we'll first check and wait for the path to exists
      * before running an NSFW watcher.
      */
     protected async start(): Promise<void> {
-        while (!this.disposed && await fsp.stat(this.fsPath).then(() => false, () => true)) {
+        while (await fsp.stat(this.fsPath).then(() => false, () => true)) {
             await timeout(500);
+            this.assertNotDisposed();
         }
+        this.assertNotDisposed();
         const watcher = await this.createNsfw();
-        if (this.disposed) {
-            return;
-        }
-
+        this.assertNotDisposed();
         await watcher.start();
         this.debug('STARTED', `disposed=${this.disposed}`);
         // The watcher could be disposed while it was starting, make sure to check for this:
         if (this.disposed) {
             await this.stopNsfw(watcher);
-            return;
+            throw WatcherDisposal;
         }
         this.nsfw = watcher;
     }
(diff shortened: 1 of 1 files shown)


paul-marechal:
The old code was very unpalatable and the new code looked much better, except it was missing the part where cancellation should reject the promise returned by `start` with `WatcherDisposal` that acts as some flag value.

I was also not understanding why the previous code caused a leaked, and I still don't precisely understand the mechanism by which it happened. But I've found discussions about `Promise.race` causing leaks when used in certain patterns: https://github.com/nodejs/node/issues/17469#issuecomment-685216777

tsmaeder:
> I was also not understanding why the previous code caused a leak, and I still don't precisely understand the mechanism by which it happened.

When you say `promiseA.then(promiseB)`, you need a pointer from `promiseA` to `promiseB`, so that when `promiseA` settles, you can start evaluating `promiseB`. This seems to be implemented via a `PromiseReaction` object in node (at least, that's what shows up in the heap dump). The same applies to `race()`: there needs to be some bookkeeping about which promise depende on which other promises. That bookkeeping consumes heap memory.

paul-marechal:
My point is that some garbage collection should be happening and it's not. People on the `Promise.race` leak thread seem to suggest this is an issue with V8's `Promise.race` implementation. They went as far as to create a [garbage-collector-friendly JS implementation](https://www.npmjs.com/package/race-as-promised)...

What we need to remember when using `Promise.race` from now on is that nothing will get collected for as long as one of the promises that's part of the race hasn't completed. Even if we discard the race promise.

paul-marechal (approved):
Code LGTM, file watching still works for me and CI is not showing new failures.

PR #12145, opened 2023-02-03
Implement workbench.editor.revealIfOpen preference

colin-grant-work:
#### What it does

Implements the `workbench.editor.revealIfOpen` preference.

#### How to test

1. By default the preference should be `false`. That means that if you open a file, it should by default open on the active tabbar even if you already have an editor open for that file on another tabbar.
2. If you set the preference to `true`, an existing editor for a file should be revealed if it is present, even if it is present on another tab than the current active tab.
> Note that it should be *revealed*; it may not be focused, so if you e.g. click from the explorer to open the file, the explorer may remain focused unless you double click. That appears to be consistent with VCode's behavior, though I'm open to defaulting to activating in that case.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/core-preferences.ts
+++ b/packages/core/src/browser/core-preferences.ts
@@ -23,6 +23,7 @@ import { isOSX } from '../common/os';
 import { nls } from '../common/nls';
 import { DefaultTheme } from '@theia/application-package/lib/application-props';
 
+/* eslint-disable max-len */
 const windowTitleDescription = [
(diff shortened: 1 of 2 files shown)


vince-fugnitto (approved):
LGTM :+1: 

- [x] if the preference is `false` the editor will open in the active tabbar if it can
- [x] if the preference is `true` the editor will be revealed if it is already open in a tabbar

PR #12098, opened 2023-01-23
Revert "Fix starting pseudoterminal (#10780)"

thegecko:
Signed-off-by: robmor01 <Rob.Moran@arm.com>

This reverts commit 363e5216dac77e4c3fa419f201cc161955084c43.

This reverts the previous fix for opening pseudoterminals as they now work as expected.

Supersedes: #12005
Fixes: #11858

Tested in browser and electron examples. Test steps from #12005 can be used (thanks @zvzuola)

cc @arekzaluski @zvzuola

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #12005, opened 2022-12-18
Fix the problem that the pseudo-terminal open event is not triggered

zvzuola:
#### What it does
Fix the problem that the pseudo-terminal open event of the custom task is not triggered

#### How to test
1. install the following test extension [task-provider-samples-0.0.1.vsix.zip](https://github.com/eclipse-theia/theia/files/10253713/task-provider-samples-0.0.1.vsix.zip) (build form [task-provider-sample](https://github.com/microsoft/vscode-extension-samples/blob/main/task-provider-sample/README.md))
2. Open any workspace and create a Rakefile empty file
3. Run task, and select "custombuildscript: 32 incremental" task to run
4. The pseudo-terminal open event is not triggered and the task is not running

#### Review checklist

- [ ] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #12146, opened 2023-02-03
Fix PseudoTerminal events

thegecko:

#### What it does

Well, this is embarrassing. The [fix merged for opening terminals](https://github.com/eclipse-theia/theia/pull/10780) and [the subsequent revert](https://github.com/eclipse-theia/theia/pull/12098) have both failed to sufficiently fix the terminal opening issues.

It transpires that opening custom terminals via tasks and by manually creating them both have different code paths ending up with the pseudoterminal map tracking the terminals being keyed with different types of IDs.

PR https://github.com/eclipse-theia/theia/pull/12005 actually partially fixed this and led me to this PR (thanks @zvzuola), however the resize events and close events weren't fixed.

This PR allows process-based terminals (used by tasks and keyed by process ID) to be re-keyed by the tracked ID (from the widget ID), so all events correctly work for both types of terminal.

#### How to test

I've created a test extension over at https://github.com/thegecko/web-extension-tasks and the built VSIX can be found in this zip: [web-extension-tasks-1.0.0.vsix.zip](https://github.com/eclipse-theia/theia/files/10583241/web-extension-tasks-1.0.0.vsix.zip)

It exposes two commands:

- `Text: Run task` which [executes a task](https://github.com/thegecko/web-extension-tasks/blob/main/src/commands.ts#L12) using a custom terminal and writes out some text
- `Text: Show text` which [creates a custom terminal](https://github.com/thegecko/web-extension-tasks/blob/main/src/commands.ts#L38) and writes out some text

The extension also shows messages when the custom terminal is resized, closed or text is input.

To test, run both these commands and observe the text being displayed and events being raised when resized/closed.

Without this PR, only one command works and not all events are raised.

The issue has been confirmed and fixed on both the electron and browser examples.

#### Review checklist

#### Reminder for reviewers

cc @arekzaluski


tsmaeder:
Hi Rob, chiming in here, since I've recently dabbled in terminals.

> This PR allows process-based terminals (used by tasks and keyed by process ID) to be re-keyed by the tracked ID (from the widget ID), so all events correctly work for both types of terminal.

The progress around this issue worries me a bit. We have "terminal id", "id" and "widget id" and "process id", but to me it's not clear what entities are designated by the various ids and what their lifecycle is.  Could you explain your thinking around this in a paragraph or two? When we have to redefine an "id", there's usually a problem with the concepts, not the implementation, in my experience.

thegecko:
> to me it's not clear what entities are designated by the various ids and what their lifecycle is. Could you explain your thinking around this in a paragraph or two?

It's a mess. In the backend (e.g. tasks) it seems the terminal ID comes from the process ID (and I think they remain identical for tasks). As the terminal is created before the widget, widget ID can't be used at this point, but is used later for keying the pseudoterminal. One is a number, one a string.
The custom task terminal is created with a UUID in the frontend and persists this ID instead (as it should IMO). The fix here is to perhaps use a UUID instead of the process ID for tasks, but is still gets created too early and would have to replace the widget ID at som point (with unknown side effects). With limited time, I traced the first point/event where we known about the terminal ID and (widget)ID in the frontend and resolved the relationship.

Happy to see this as a short-term fix with further refactoring required.

tsmaeder:
I think a lot of the confusion stems from the fact that terminal processes and pseudoterminals are both handled via the shell terminal server: there are numerous places where the front end or back end just does nothing when a pseudoterminal (from a custom task execution or from extension terminal options) is involved. In the end, the terminal widget is only interested in reading and writing text from/to some entity (and the lifecycle of the underlying entity, like a process or task execution). IMO having different back ends for process terminals and pseudo terminals and an abstraction of the two in the front end would clarify things a lot: the id namespaces for back-end entities would be separate for processes and pseudos.

thegecko:
In the interests of continuous improvement, shall we consider this fix now and create an issue for refactoring terminal support?

tsmaeder:
@thegecko I have a bit of a work queue. I won't be able to look at this before late this week/early next.

packages/plugin-ext/src/plugin/terminal-ext.ts
```diff
@@ -189,7 +189,18 @@ export class TerminalServiceExtImpl implements TerminalServiceExt {
             terminal.deferredProcessId = new Deferred<number>();
             terminal.deferredProcessId.resolve(processId);
         }
-        const pseudoTerminal = this._pseudoTerminals.get(terminalId.toString());
+
+        // Switch the pseudoterminal keyed by terminalId to be keyed by terminal ID
+        const tId = terminalId.toString();
+        if (this._pseudoTerminals.has(tId)) {
+            const pseudo = this._pseudoTerminals.get(tId);
+            if (pseudo) {
```
tsmaeder, line 197: Does it ever make sense that the pseudo-terminal is `falsy` if the map `has` the key in question?

packages/plugin-ext/src/main/browser/terminal-main.ts
```diff
@@ -95,20 +95,22 @@ export class TerminalServiceMainImpl implements TerminalServiceMain, TerminalLin
...
-        terminal.title.changed.connect(updateTitle);
-        this.toDispose.push(Disposable.create(() => terminal.title.changed.disconnect(updateTitle)));
 
         const updateProcessId = () => terminal.processId.then(
             processId => this.extProxy.$terminalOpened(terminal.id, processId, terminal.terminalId, terminal.dimensions.cols, terminal.dimensions.rows),
             () => {/* no-op */ }
         );
+
         updateProcessId();
+        terminal.title.changed.connect(updateTitle);
```
tsmaeder, line 112: This looks like some code was moved around with no functional changes. Am I missing something here?

packages/plugin-ext/src/plugin/terminal-ext.ts
thegecko: No, but `Map.get()` returns `thing | undefined`, so won't compile without that guard IIRC. The only change I think can be made here is to remove the outer `has()` check. Want me to do that?

packages/plugin-ext/src/main/browser/terminal-main.ts
thegecko: The ordering here was important. Its been a while since I was in this code, but I think `updateProcessId()` needed to be called first to open the terminal and guarantee the IDs. With the changed event earlier in the code a race condition existed.

tsmaeder: But we're not even waiting for anything here: so while this might change the timing, it doesn't seem to guarantee any ordering of events, as far as I can tell. Since the `terminalOpened` waits for `terminal.processId`, there is no guarantee that `updateTitle` is not called before before `updateProcessId`. If the ordering dependencies are that subtle, we should at least add a comment, lest the next person reorders the code.

packages/plugin-ext/src/plugin/terminal-ext.ts
thegecko: OK

packages/plugin-ext/src/main/browser/terminal-main.ts
thegecko: I re-tested this and it seems this wasn't needed. Perhaps I just preferred the code that way!. Have removed the changes in this file.

tsmaeder (approved):
Looks good to me and seems to do the job in test.

Issue #12153, opened 2023-02-06
Duplicated launch configs


marcdumais-work:
### Bug Description:
I've been working to improve the `browser test suite` and when I updated my baseline to include the content of the most recent `master` branch, over 100 tests from `launch-preferences.spec.js` suddenly started consistently failing. 

This test file is a bit obscure, but I think the fundamental reason for the failures is that we end-up with duplicated launch configurations, whenever the preferences machinery evaluates them. I have traced the issue to PR #12126, which makes a change in `PreferenceProvider`. 

Note: the `browser test suite` has been very flaky lately, so it's no surprise its failures ends-up being not noticed during PR reviews. Additionally, the new failures related to the launch preferences generate so many error traces that I think the GitHub runner gets overwhelmed and  ends-up not showing the detailed errors at the end of execution, which makes it very hard to notice that there are additional new failures (vs the usual ones). 

To run the failing test file locally and get access to the full output log: 

1. build the Theia repo from sources: `yarn && yarn browser build && yarn download:plugins`
2. `cd examples/browser`
3. `yarn -s rebuild && npx theia test . --plugins=local-dir:../../plugins --test-spec=../api-tests/src/launch-preferences.spec.js`

**update: If the test failure traces are still not present in the output, try running in non-headless mode:**
```yarn -s rebuild && npx theia test . --plugins=local-dir:../../plugins --test-inspect --test-spec=../api-tests/src/launch-preferences.spec.js```

### Steps to Reproduce:

1. Build the Theia repo from sources
5. Use the example Theia application to open folder `examples/browser` as your workspace
6. Switch to the debug view. 
7. Trigger the `launch` drop-down and select `Add Configuration` and select a type of launch (e.g.: "Node.js: Launch Program")
8. Trigger the `launch` drop-down again and confirm that two launches are listed, under the name of the new added launch (e.g. "Launch Program"

Alternatively, open the Theia repo root folder as a workspace and confirm that all launch are duplicated (some maybe more than one time). 

The same duplication happens if one adds a launch to the `settings.json` file. 

See this video capture: 
![duplicated launches](https://user-images.githubusercontent.com/25749063/217039194-84463df1-199d-46b5-b4ca-caae5fd214ff.gif)

### Additional Information
Not sure if it might help, but while debugging I have added some printouts to `preferences-provider.ts` and captured the following traces, when adding a launch configuration, using the reproduce steps above:

```typescript diff
[...]

PR #12165, opened 2023-02-09
preferences: fix duplicated resolution

paul-marechal:
The `WorkspacePreferenceProvider` can do one of two things: It may return preferences from the current workspace folder if the workspace is a single root workspace, or it may return preferences as defined in the workspace file if the workspace is a multi-root workspace.

Whenever we want to resolve a preference, the following scopes are queried: `Default`, `User`, `Workspace`, and `Folder`.

When a single root workspace is opened the following happens:

- Query `Default` scope [...]
- Query `User` scope [...]
- Query `Workspace` scope: use `WorkspacePreferenceProvider` which delegates to `FoldersPreferenceProvider`.
- Query `Folder` scope: use `FoldersPreferenceProvider`.

Conclusion: The `FoldersPreferenceProvider` is queried twice.

In order to fix that issue this commit introduces a new component: `TogglePreferenceProvider` which extends `PreferenceProvider` and wraps another provider so that it can be enabled/disabled.

Whenever we are running a single root workspace, we'll enable the `FoldersPreferenceProvider` used by `WorkspacePreferenceProvider` and disable the one bound to the `Folder` scope.

Whenever we are running a multi root workspace, we'll disable the `FoldersPreferenceProvider` used by `WorkspacePreferenceProvider` and enable the one bound to the `Folder` scope.

Fixes https://github.com/eclipse-theia/theia/issues/12153

#### How to test

See instructions in https://github.com/eclipse-theia/theia/issues/12153

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/api-tests/src/launch-preferences.spec.js
+++ b/examples/api-tests/src/launch-preferences.spec.js
@@ -37,6 +37,7 @@ describe('Launch Preferences', function () {
     const { FileResourceResolver } = require('@theia/filesystem/lib/browser/file-resource');
     const { AbstractResourcePreferenceProvider } = require('@theia/preferences/lib/browser/abstract-resource-preference-provider');
     const { waitForEvent } = require('@theia/core/lib/common/promise-util');
+    const { FoldersPreferencesProvider } = require('@theia/preferences/lib/browser/folders-preferences-provider');
 
     const container = window.theia.container;
     /** @type {import('@theia/core/lib/browser/preferences/preference-service').PreferenceService} */
@@ -46,7 +47,7 @@ describe('Launch Preferences', function () {
     /** @type {import('@theia/preferences/lib/browser/workspace-preference-provider').WorkspacePreferenceProvider} */
     const workspacePreferences = container.getNamed(PreferenceProvider, PreferenceScope.Workspace);
     /** @type {import('@theia/preferences/lib/browser/folders-preferences-provider').FoldersPreferencesProvider} */
-    const folderPreferences = container.getNamed(PreferenceProvider, PreferenceScope.Folder);
+    const folderPreferences = container.get(FoldersPreferencesProvider);
     const workspaceService = container.get(WorkspaceService);
     const fileService = container.get(FileService);
     const fileResourceResolver = container.get(FileResourceResolver);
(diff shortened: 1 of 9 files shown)


paul-marechal:
cc @AlexandraBuzila

paul-marechal:
I didn't know what to do with our current preference infrastructure. It's a mess, and this patch doesn't make things really nicer... In the meantime, this patch seems to work (on my machine) and it's not super invasive. If anyone has a better idea please open an alternative PR, or push directly to this one if you can.

AlexandraBuzila:
Hi @paul-marechal 

I also found a bug in the PreferenceProvider while investigating this issue: the merge of array values was producing duplicates. I pushed my changes here https://github.com/eclipsesource/theia/commit/eb8d246ca7a09a8c8da89ecdd7daf0f6188e7211
If you agree with them, it would be fine for me if you were to integrate them in your PR.

With my changes alone, some tests from  `examples/api-tests/src/launch-preferences.spec.js` were still failling. While investigating them, I saw that preferences were merged twice, so your changes might have fixed the issues I was seeing when running the tests. I couldn't verify this ATM since the tests won't run now anymore, I think they might need to be adapted to account for the new provider.

planger:
@paul-marechal I saw that this PR impacts the Playwright page objects. On a first glance, it looks like the selector of the modification indicator in the gutter needs to be adapted. Please let me know if you'd like us to assist with that.

paul-marechal:
@AlexandraBuzila my problem is that from the code alone:

https://github.com/eclipse-theia/theia/blob/371ab02dda5c6df2c1381ab68b2f5ae341f7d258/packages/core/src/browser/preferences/preference-provider.ts#L225-L246

I can't see why values would be duplicated, _unless_ we pass two arrays containing the same data together.

If you read the PR description I highlighted the root cause of the duplication, that is the preference system calls twice the same handler. Since the previous logic didn't properly merge arrays, that didn't seem to cause an issue.

With that in mind, this PR addresses this problem of `FoldersPreferencesProvider` being called twice, and so far this seemed to have fixed the duplicated config issue.

paul-marechal:
@planger I don't think this PR should require updating Playwright object models, let me see if I can fix the issues that appeared in CI first.

paul-marechal:
Closing in favor of https://github.com/eclipse-theia/theia/pull/12174

Issue #10669, opened 2022-01-26
Add `tests` namespace from VSCode API


sgraband:
### Feature Description:
Add function from the `tests` namespace:

- [ ] createTestController

Add required variables, functions and fields from `root` namespace:

- [ ] TestController
- [ ] TestItem
- [ ] TestItemCollection
- [ ] TestMessage
- [ ] TestRun
- [ ] TestRunProfile
- [ ] TestRunProfileKind
- [ ] TestRunRequest
- [ ] TestTag

VSCode API for reference: https://code.visualstudio.com/api/references/vscode-api#tests.

fyudanov:
Would be nice to have, e.g. for integrating https://github.com/robocorp/robotframework-lsp with Test Explorer view in Theia.

planger:
Note that the test API [has been stubbed](https://github.com/eclipse-theia/theia/pull/11717) for now. So extensions using this API aren't blocked from running, but usages of the test API won't have any effect.

KevinClapperton:
Hello, I speak on behalf of the student project team from Polytechnique Montreal (Kevin, Théo, Mathieu and Zakaria). This is our first time contributing and we would like to work on this API. We are still going through some tutorials and documentation to make sure we understand how to implement this feature (we might have a slow start). 

Here is what we are doing right now in order to understand how to implement the Test API: 
- Reading [Plugin-API.md](https://github.com/eclipse-theia/theia/blob/master/doc/Plugin-API.md) and [how to add the test namespace](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md) 
- Looking how the [Test API was stubbed](https://github.com/eclipse-theia/theia/pull/11717) 
- Looking how other VSCode API are implemented (path: [theia/packages/plugin-ext/src/plugin/](https://github.com/eclipse-theia/theia/tree/master/packages/plugin-ext/src/plugin))
- Checking [VSCode extension sample for the Test API](https://github.com/microsoft/vscode-extension-samples/blob/main/test-provider-sample/src/extension.ts) and [the guide](https://code.visualstudio.com/api/extension-guides/testing) plus the [official documentation](https://code.visualstudio.com/api/references/vscode-api#tests)
- Looking at a [partial implementation of the Test API in Haxe](https://github.com/vshaxe/vscode-extern/tree/master/src/vscode) 

Question: 
[...]

PR #12175, opened 2023-02-12
Tests API implementation

D-Zaq:
#### What it does

Initial setup for the VSCode tests API implementation by the students team from Polytechnique Montreal.

#### Issue adressed

Feature description of the issue: [ Add tests namespace from VSCode API #10669 ](https://github.com/eclipse-theia/theia/issues/10669)

#### Review checklist

#### Reminder for reviewers


paul-marechal:
If you look at the CI output you have compilation errors, please fix.

packages/plugin-ext/src/common/plugin-api-rpc.ts
```diff
@@ -114,6 +114,8 @@ import { isString, isObject, PickOptions, QuickInputButtonHandle } from '@theia/
 import { Severity } from '@theia/core/lib/common/severity';
 import { DebugConfiguration, DebugSessionOptions } from '@theia/debug/lib/common/debug-configuration';
 
+import { CoverageDetails, IFileCoverage, ISerializedTestResults, RunTestForControllerRequest, TestsDiffOp, ITestRunProfile, ResolvedTestRunRequest, ITestItem, TestResultState, ITestMessage, ITestRunTask, ExtensionRunTestsRequest } from '@theia/testing/lib/common/test-types';
```
colin-grant-work, line 117: It looks like you may need to add the `testing` package to the `package.json` for `plugin-ext` to make these items available in this package.

paul-marechal: The `tsconfig` should be updated automatically when running `yarn compile`.

mathblin:
Thank you for the feedback. Like you both said, it was related to the package file in plugin-ext.
We were able to make it compile locally from a fresh clone.
We'll rebase the history for the conflicts soon.

JonasHelming:
@tsmaeder

mathblin:
Can anybody confirm that this kind of import is ok in Theia? (see in [packages/plugin-ext/src/plugin/type-converters.ts](https://github.com/eclipse-theia/theia/pull/12175/commits/1bfe5b8b9c366607dd4474c843f3d118ad1e2bc6#diff-433c140c704e6be24792fa08539863d9b41465fbee75a3856bd524181b8d3bb9)
`
import { parse } from '@theia/monaco-editor-core/esm/vs/base/common/marshalling';
`
@colin-grant-work , @paul-marechal 
Also, after doing the `git clean -xffd` command, we could compile without errors. However, we saw there was another commit on master today. We will rebase later on.

colin-grant-work:
> Can anybody confirm that this kind of import is ok in Theia?
> `import { parse } from '@theia/monaco-editor-core/esm/vs/base/common/marshalling';`

Yes, that should work fine, assuming the corresponding function is in fact available in the `monaco-editor-core` code.

paul-marechal:
You currently have errors related to some import: `D:\a\theia\theia\node_modules\@theia\monaco-editor-core\src\vs\base\common\marked\marked.js` (from the CI build failures).

Note the `/src/` part in the path: you should not import anything from there as it is where the TypeScript source files are located, and you want to import JavaScript files instead. You may want to replace `src` by `esm`.

packages/plugin-ext/src/plugin/type-converters.ts
```diff
@@ -29,6 +29,16 @@ import { UriComponents } from '../common/uri-components';
...
+import * as editorRange from '@theia/monaco-editor-core/esm/vs/editor/editor.api';
+import { denamespaceTestTag, ITestErrorMessage, ITestItem, ITestTag, namespaceTestTag, TestMessageType } from '@theia/testing/lib/common/test-types';
+import { TestId } from '@theia/testing/lib/common/test-id';
+import * as htmlContent from '@theia/monaco-editor-core/esm/vs/base/common/htmlContent';
+import { getPrivateApiFor } from './testing-private-api';
+import { parse } from '@theia/monaco-editor-core/esm/vs/base/common/marshalling';
+import { cloneAndChange } from '@theia/monaco-editor-core/esm/vs/base/common/objects';
+import * as languages from '@theia/monaco-editor-core/esm/vs/editor/common/languages';
+import { Location } from '../common/plugin-api-rpc-model';
+import * as marked from '@theia/monaco-editor-core/src/vs/base/common/marked/marked';
```
paul-marechal, line 41:
```suggestion
import * as marked from '@theia/monaco-editor-core/esm/vs/base/common/marked/marked';
```

D-Zaq: Hi @paul-marechal, we have in fact tried importing `marked` with the JavaScript file from `esm` instead of the Typescript from `src` like the other imports but that generates compilation errors (as shown in the screenshot below) suggesting to add the marked.d.ts file in the _marked_ directory where `marked.js` is found following the `esm` path which is a temporary fix as it would result in the same error with a fresh build. We don't know the reason of this and we still didn't find a permanent solution :

image

colin-grant-work: The reason is that the `marked` files in `esm` have no sourcemap because `marked` isn't compiled from source. [`marked` is a library](https://www.npmjs.com/package/marked) that VSCode has in-sourced by adding the compiled JS from a particular commit of `marked`. That means there's no source map in `esm`, so TS can't find its way back to the `d.ts` in the sources. There may be some TS mechanism to import from `esm` (definitely the correct idea) and still point to the `d.ts` in the `src` directory. though.

mathblin:
Could it be possible to launch the pipeline again. We just rebased origin/master and it is working locally for `yarn­` and `yarn lint`.

JonasHelming:
sure, running!

mathblin:
> > Can anybody confirm that this kind of import is ok in Theia?
> > `import { parse } from '@theia/monaco-editor-core/esm/vs/base/common/marshalling';`
> 
> Yes, that should work fine, assuming the corresponding function is in fact available in the `monaco-editor-core` code.

Thanks for the reply. It is always appreciated and most responses lead us to helping solve the problems that we face.

However, as discussed in the dev-meeting as of 2023-03-21, some tests now fails due to imports like this from `monaco-editor-core` code. Maybe it is linked to browser code as in code that is not specific to browser maybe?

At this point, we are not sure if we should replace all `monaco-editor-core` code imports by real code lines in `Theia` or at least in appropriate parts. We would also like to know if anyone could help us define these appropriate parts. Could it be that these imports from `monaco-editor-core` are reserved for files that are created in or used by main implementations. For example, files created in or used in this [repository](https://github.com/eclipse-theia/theia/tree/master/packages/plugin-ext/src/main/browser).

By the way, once Theia compiled, whether it is with `Electron` or `Browser`, the plugins crashes in runtime. We think this crash is related to the same issue that makes the tests fail.
image

paul-marechal:
About the plugin host crash, I get the following error in my console when trying this PR:

```
root ERROR [hosted-plugin: 39724] E:\sources\theia.banana\node_modules\@theia\monaco-editor-core\esm\vs\editor\standalone\browser\standalone-tokens.css:8
.monaco-editor {
^

SyntaxError: Unexpected token '.'
    at Object.compileFunction (node:vm:352:18)
    at wrapSafe (node:internal/modules/cjs/loader:1032:15)
    at Module._compile (node:internal/modules/cjs/loader:1067:27)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1157:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (E:\sources\theia.banana\node_modules\@theia\monaco-editor-core\esm\vs\editor\standalone\browser\standaloneEditor.js:8:1)
    at Module._compile (node:internal/modules/cjs/loader:1103:14)
```

Something is trying to import a CSS file from the plugin host and it is causing this problem.

Note that Theia frontend code can import `.css` files because these imports are converted to something that works by Webpack upon bundling. Beside the frontend we should never somehow import CSS files in TS/JS scripts.

paul-marechal:
@colin-grant-work do I guess correctly that `@theia/monaco-editor-core` is a browser package? As in we'd have issues trying to import it from node runtimes?

paul-marechal:
@mathblin I put a breakpoint in `node_modules\@theia\monaco-editor-core\esm\vs\editor\standalone\browser\standaloneEditor.js` on the `require("./standalone-tokens.css");` line and here's the stack I got:

`packages\testing\src\common\test-types.ts` imports symbols from `@theia/monaco-editor-core/esm/vs/editor/editor.api` which in turn imports things from these browser APIs.

The fix would be to not import symbols from `@theia/monaco-editor-core/esm/vs/editor/editor.api` because of its side effects.

colin-grant-work:
> @colin-grant-work do I guess correctly that @theia/monaco-editor-core is a browser package? As in we'd have issues trying to import it from node runtimes?

Not exclusively. It includes a number of folders from the Code OSS repo in `common` areas, and those should be safe to use in Node contexts.

mathblin:
Hello @paul-marechal @colin-grant-work @marcdumais-work ,

I was curious to know if anyone could help us fix this issue with `inversify`:
Since we started implementing the tests-main.ts in `packages/plugin-ext/src/main/browser/`, we had problems with binding of a class/interface which `inversify` says it does not exist or has not been binded to what we understand is the container module.
In short, we can compile without errors with the following commands: `yarn` and `yarn browser build`. The problem is at run time reproducible with the command `yarn browser start` and opening the browser at `localhost:3000/`. We did not test with electron directly, but we expect a similar behaviour.
Here is the screenshot of the problem:
image
As we can see at the end of the first line, there is a binding problem with the error: `Error: No matching bindings found for serviceIdentifier: TestService`

The error is also reproductible with a test file. One can go in `packages/testing/` folder and use command `yarn test`. Here is a similar screenshot:
image
The test is made with the same class: 
```typescript
export class TestService ...
```
*Important*: You'll have to use this fork of eclipse theia : [https://github.com/D-Zaq/theia](https://github.com/D-Zaq/theia)
We did not push it in the PR because the PR compiles and doesn't break any code.

Well, any help would be appreciated.

paul-marechal:
@mathblin I only quickly looked over why the `tests-frontend-module.spec.ts` would be failing and the reason is that `TestService` has a lot of dependencies that require binding in the Inversify container for an instance of `TestService` to be resolvable.

From [`test-service-impl.ts`](https://github.com/D-Zaq/theia/blob/b0826453c95b7bdca0a3def9ebace737eec1b92f/packages/testing/src/common/test-service-impl.ts#L146-L156):

```ts
    constructor(
        @inject(IContextKeyService) contextKeyService: IContextKeyService,
        @inject(IInstantiationService) instantiationService: IInstantiationService,
        @inject(IStorageService) private readonly storage: IStorageService,
        // @IEditorService private readonly editorService: IEditorService,
        @inject(ITestProfileService) private readonly testProfiles: ITestProfileService,
        @inject(INotificationService) private readonly notificationService: INotificationService,
        // @IConfigurationService private readonly configurationService: IConfigurationService,
        @inject(ITestResultService) private readonly testResults: ITestResultService,
        @inject(IWorkspaceTrustRequestService) private readonly workspaceTrustRequestService: IWorkspaceTrustRequestService,
    ) {
```

This is all the dependencies of your components, and all need to have a valid binding for `TestService` to be instantiable.

paul-marechal:
Now I'm also a bit queasy about some of the service identifiers you are using: `IContextKeyService` seems to come from `@theia/monaco-editor-core` and is a VS Code specific abstraction. Inversify seems ok with its value being used as a service identifier because it happens to be a function, but something tells me you should be using Theia's bindings and abstractions over things coming straight from the monaco editor? I'm just not really sure about directly using these symbols.

In this specific case we have a `ContextKeyService` component from `@theia/core` that you should be able to use.

paul-marechal:
Tried looking at the `TestService` dependencies and where to look:

- I don't think you should need `IInstantiationService`, or convert what it is used for by using Inversify instead?
- I found a `StorageService` API within Theia but it doesn't seem to do everything you want it to?
- It seems like you have an `@injectable()` implementation for `ITestProfileService`, so you should use that.
- I think instead of `INotificationService` Theia has a `MessageService` API that fulfills the same role?
- There seems to be an `@injectable()` implementation for `ITestResultService`, use that.
- I don't think Theia supports an API similar to what `IWorkspaceTrustRequestService` does yet.

For better or for worse, we define and implement our own APIs/abstractions instead of directly using what third-party libraries might use (e.g. `@theia/monaco-editor-core` is a third-party library).

I think it's especially important when creating whole new components like `TestService`. If I understand correctly the current dependencies are directly inspired from the VS Code implementation? Then it makes sense why you'd try to find the same abstractions from `@theia/monaco-editor-core`, but Theia offers a different internal API and we must use that instead.

colin-grant-work:
To supplement what @paul-marechal has said, it is possible to integrate services available from `@theia/monaco-editor-core` into the `inversify` dependency injection system, but it is tricky. You can see a couple of places where we've done it by searching for `decorate(injectable()` in the `monaco` package, but you have to be very careful, because, with dependency injection, instantiating one thing may instantiate a bunch of other things you didn't realize were necessary, and you have to take their lifecycles (and whether you - or anyone else - might want to override them) into account. So trying to use as many things as possible that are implemented in Theia rather than in `monaco-editor-core` is likely the right way to go, or delegating, but carefully.

mathblin:
Thank you all for the quick responses. We'll look into them in the next few days and we'll also try to answer your comments (which are helpful) as fast as possible.

mathblin:
> Now I'm also a bit queasy about some of the service identifiers you are using: `IContextKeyService` seems to come from `@theia/monaco-editor-core` and is a VS Code specific abstraction. Inversify seems ok with its value being used as a service identifier because it happens to be a function, but something tells me you should be using Theia's bindings and abstractions over things coming straight from the monaco editor? I'm just not really sure about directly using these symbols.
> 
> edit: Note that even if you try to inject using these symbols, unless you have them bound in this runtime's Inversify container then nothing will happen and Inversify will fail to resolve the injections and instantiations.
> 
> In this specific case we have a `ContextKeyService` component from `@theia/core` that you should be able to use.

I think we can use the `ContextKeyService` as you suggested. It does not make me any compile errors if I use it instead of `IInstantiationService` (with some internal changes in Tests API of course). However, I am not sure of the behaviour thereafter. To be honest, even if we succeed in making all these changes, there will probably be a lot to modify after. We'll do our best anyway.

mathblin:
> are implemented in Theia rather than in `monaco`

I think you are absolutely right about avoiding the use of `monaco` and focus more on what's already implemented in Theia. At first, we used it because we saw some examples of it (that we didn't fully understand). However, it gave us more problems than expected. We are still figuring out ways to avoid it, which in part would mean importing a lot of new lines of code or changing what we already imported. 

For the first point, I don't think we should try the `decorate(injectable()` solution unless we have no other choice. We'll keep it in mind though. Maybe we'll make something out of it.

JonasHelming:
I triggered

D-Zaq:
@JonasHelming can we re-triggre again 🙏🏼 ?

KevinClapperton:
There was an update to the function declaration called `createRunProfile`. The current code doesn't migrates the new implementation that goes with the updated declaration. I will see what I can do on Tuesday

KevinClapperton:
I restored the commit before the changes from the new [Test API stub](https://github.com/eclipse-theia/theia/pull/12456), but keep the correction of headers, comments and links to imported code. Also, made sure to double check the build, tests and lint.

tsmaeder:
Incorporated and superseded by https://github.com/eclipse-theia/theia/pull/12935

Issue #11878, opened 2022-11-17
Multiple `DebugEditorModel`s for same URI lead to stack overflow


colin-grant-work:
### Bug Description:

If two editors are open for the same file, a number of issues arise related to breakpoint management due to the fact that both editors have their own `DebugEditorModel`s.

### Steps to Reproduce:

1. Open an editor.
2. Open a second instance (`Ctrl+\`)
3. Add a breakpoint.
4. Right click on the breakpoint decoration in the first editor and say 'Edit Breakpoint'
5. Observe that the breakpoint editor opens in the second editor.
6. Move the breakpoint up or down a few lines by adding or removing content from the second editor.
7. Observe hundreds of warnings about calling `deltaDecorations` recursively, as well as, possibly, a freeze and numerous stack overflow warnings.

### Additional Information

![image](https://user-images.githubusercontent.com/62660806/202559699-0afee09f-c1c1-4372-82ab-a6b34d9a9d71.png)

There are two related issues. A `DebugEditorModel` is supposed to ignore events that originate from itself, which is the import of the check for `this.updatingDecorations` here:

https://github.com/eclipse-theia/theia/blob/fe8e4c1276854baad22a2ffcde6acfb1e8e50926/packages/debug/src/browser/editor/debug-editor-model.ts#L308-L320

But if another editor is active for the same URI, then each will trigger the other to check its breakpoints and eventually emit a new change event.

Relatedly, in the `DebugEditorService`, a new model is created for every editor, regardless of whether a model exists for that URI. This may also lead to undisposed `DebugEditorModels`.

https://github.com/eclipse-theia/theia/blob/fe8e4c1276854baad22a2ffcde6acfb1e8e50926/packages/debug/src/browser/editor/debug-editor-service.ts#L61-L73

- Operating System: RHEL 7
- Theia Version: fe8e4c1276854baad22a2ffcde6acfb1e8e50926

colin-grant-work:
The behavior seems to be quite bad in Firefox; it's hard to reproduce steps 6-7 at all in Chrome.

PR #12183, opened 2023-02-15
Fire breakpoint events only if breakpoint metadata changes

colin-grant-work:
#### What it does

Fixes https://github.com/eclipse-theia/theia/issues/11878.

This PR fixes the stack overflow component of #11878 by ensuring that debug editor models provide the ID of breakpoints they believe have moved rather than retrieving the modified breakpoint from the breakpoint manager (whose state may have been updated in the meantime by a different model for the same editor) using location data and by ensuring that the breakpoint manager doesn't fire events when `setMarkers` is called but no marker - breakpoint - metadata changes.

#### How to test

1. Add a breakpoint to an editor.
2. Split that editor so that two (or more) editor instances are opened for the file.
3. Perform a text edit in one of the editor instances that moves the breakpoint up or down.
4. The breakpoint should move, the UI should remain responsive, and the console should not show any stack overflow errors.
5. Other breakpoint actions (adding, removing, modifying with conditions, etc.) should work as before.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/package.json
+++ b/packages/debug/package.json
@@ -17,6 +17,7 @@
     "@theia/variable-resolver": "1.34.0",
     "@theia/workspace": "1.34.0",
     "@vscode/debugprotocol": "^1.51.0",
+    "fast-deep-equal": "^3.1.3",
     "jsonc-parser": "^2.2.0",
     "p-debounce": "^2.1.0"
   },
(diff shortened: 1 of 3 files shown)


vince-fugnitto (approved):
I can confirm that the changes work as expected, and the errors no longer exist when a split editor is modified with breakpoints :+1:

Issue #11829, opened 2022-11-03
vscode: add full support for `SourceControlResourceThemableDecorations.iconPath`


vince-fugnitto:
#### Feature Description:

The `SourceControlResourceThemableDecorations.iconPath` should be updated to accept `ThemeIcon` similarly to the [VS Code API](https://code.visualstudio.com/api/references/vscode-api#SourceControlResourceThemableDecorations).

https://github.com/eclipse-theia/theia/blob/1cdfae21899f1efacb2a8caec7f2147129518c8d/packages/plugin/src/theia.d.ts#L10352-L10358

PR #12187, opened 2023-02-15
plugin: ThemeIcon Support in SCM

FernandoAscencio:
Finishes Estelle's changes to support Themables.

#### What it does
Closes: #11829 
Allows `SourceControlResourceThemableDecorations.iconPath` to accept `ThemeIcon` and modifies the SCM supporting architecture to accept the change.

TheiaTest

#### How to test
1. Run Theia
2. Check that the status symbol is an M.
3. Modify the git extension files to pass a `ThemeIcon` (URI or codicon) as a decoration
4. Run Theia
5. Verify that the status symbol has changed from an M to the `ThemeIcon`.

Tests can be done with modified `git-base` and `git` files [here](https://github.com/eclipse-theia/theia/files/10744547/vsix-test-files.zip)
If you do, you will need to delete `"@theia/git": "1.34.0"` from the `examples\browser\package.json` and `examples\electron\package.json` files. This package currently only tests for 'Light' and "Dark" theme changes as well as `URI` and `ThemeIcon` inputs.

Possible tests:
- [ ] Check for different icons for different Status (Untracked, Modified, Staged) if possible.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/scm-main.ts
+++ b/packages/plugin-ext/src/main/browser/scm-main.ts
@@ -40,6 +40,8 @@ import { URI as vscodeURI } from '@theia/core/shared/vscode-uri';
 import { Splice } from '../../common/arrays';
 import { UriComponents } from '../../common/uri-components';
 import { ColorRegistry } from '@theia/core/lib/browser/color-registry';
+import { PluginSharedStyle } from './plugin-shared-style';
+import { IconUrl } from '../../common';
 
 export class PluginScmResourceGroup implements ScmResourceGroup {
 
@@ -147,6 +149,7 @@ export class PluginScmProvider implements ScmProvider {
     constructor(
         private readonly proxy: ScmExt,
         private readonly colors: ColorRegistry,
+        private readonly style: PluginSharedStyle,
         private readonly _handle: number,
         private readonly _contextValue: string,
         private readonly _label: string,
@@ -222,13 +225,13 @@ export class PluginScmProvider implements ScmProvider {
                 const { start, deleteCount, rawResources } = groupSlice;
                 const resources = rawResources.map(rawResource => {
                     const { handle, sourceUri, icons, tooltip, strikeThrough, faded, contextValue, command } = rawResource;
-                    const icon = icons[0];
-                    const iconDark = icons[1] || icon;
+                    const icon = this.toIconClass(icons, true);
+                    const iconDark = this.toIconClass(icons, false);
                     // eslint-disable-next-line @typescript-eslint/no-explicit-any
                     const colorVariable = (rawResource as any).colorId && this.colors.toCssVariableName((rawResource as any).colorId);
                     const decorations = {
-                        icon: icon ? vscodeURI.revive(icon) : undefined,
-                        iconDark: iconDark ? vscodeURI.revive(iconDark) : undefined,
+                        icon,
+                        iconDark,
                         tooltip,
                         strikeThrough,
                         // TODO remove the letter and colorId fields when the FileDecorationProvider is applied, see https://github.com/eclipse-theia/theia/pull/8911
@@ -258,6 +261,12 @@ export class PluginScmProvider implements ScmProvider {
         this.onDidChangeResourcesEmitter.fire();
     }
 
+    private toIconClass(icons: IconUrl, isLightTheme: boolean): string {
+        let icon = typeof icons === 'object' ? (isLightTheme ? icons.light : icons.dark) : icons;
+        icon = icon.includes('codicon') ? icon : this.style.toIconClass(icon).object.iconClass;
+        return icon;
+    }
+
     unregisterGroup(handle: number): void {
         const group = this.groupsByHandle[handle];
 
@@ -280,11 +289,13 @@ export class ScmMainImpl implements ScmMain {
     private repositoryDisposables = new Map<number, DisposableCollection>();
     private readonly disposables = new DisposableCollection();
     private readonly colors: ColorRegistry;
+    private readonly style: PluginSharedStyle;
 
     constructor(rpc: RPCProtocol, container: interfaces.Container) {
         this.proxy = rpc.getProxy(MAIN_RPC_CONTEXT.SCM_EXT);
         this.scmService = container.get(ScmService);
         this.colors = container.get(ColorRegistry);
+        this.style = container.get(PluginSharedStyle);
     }
 
     dispose(): void {
@@ -298,7 +309,7 @@ export class ScmMainImpl implements ScmMain {
     }
 
     async $registerSourceControl(handle: number, id: string, label: string, rootUri: UriComponents | undefined): Promise<void> {
-        const provider = new PluginScmProvider(this.proxy, this.colors, handle, id, label, rootUri ? vscodeURI.revive(rootUri) : undefined);
+        const provider = new PluginScmProvider(this.proxy, this.colors, this.style, handle, id, label, rootUri ? vscodeURI.revive(rootUri) : undefined);
         const repository = this.scmService.registerScmProvider(provider, {
             input: {
                 validator: async value => {
--- a/packages/plugin-ext/src/plugin/plugin-icon-path.ts
+++ b/packages/plugin-ext/src/plugin/plugin-icon-path.ts
@@ -18,10 +18,11 @@ import * as path from 'path';
 import { URI } from './types-impl';
 import { IconUrl, PluginPackage } from '../common/plugin-protocol';
 import { Plugin } from '../common/plugin-api-rpc';
+import { ThemeIcon } from '@theia/monaco-editor-core/esm/vs/platform/theme/common/themeService';
 
-export type PluginIconPath = string | URI | {
-    light: string | URI,
-    dark: string | URI
+export type PluginIconPath = string | URI | ThemeIcon | {
+    light: string | URI | ThemeIcon,
+    dark: string | URI | ThemeIcon
 };
 export namespace PluginIconPath {
     export function toUrl(iconPath: PluginIconPath | undefined, plugin: Plugin): IconUrl | undefined {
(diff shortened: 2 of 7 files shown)


packages/plugin-ext/src/plugin/plugin-icon-path.ts
```diff
@@ -18,10 +18,11 @@ import * as path from 'path';
 import { URI } from './types-impl';
 import { IconUrl, PluginPackage } from '../common/plugin-protocol';
 import { Plugin } from '../common/plugin-api-rpc';
+import { ThemeIcon } from '@theia/monaco-editor-core/esm/vs/platform/theme/common/themeService';
 
-export type PluginIconPath = string | URI | {
-    light: string | URI,
-    dark: string | URI
+export type PluginIconPath = string | URI | ThemeIcon | {
+    light: string | URI | ThemeIcon,
```
tsmaeder, line 24: Since the ThemeIcon comes from a Theme, which is already light or dark, does it make sense the have a "light ThemeIcon" and a "dark ThemeIcon"?

packages/plugin-ext/src/plugin/scm.ts
```diff
@@ -38,17 +39,23 @@ import { RPCProtocol } from '../common/rpc-protocol';
...
     if (!decorations) {
         return undefined;
     } else if (typeof decorations.iconPath === 'string') {
         return URI.file(decorations.iconPath);
-    } else {
+    } else if (URI.isUri(decorations.iconPath)) {
+        return decorations.iconPath;
+    } else if (ThemeIcon.isThemeIcon(decorations.iconPath)) {
         return decorations.iconPath;
+    } else {
```
tsmaeder, line 57: Isn't this an error condition that should be logged as a warning?

packages/plugin-ext/src/plugin/scm.ts
```diff
@@ -511,6 +508,14 @@ class SsmResourceGroupImpl implements theia.SourceControlResourceGroup {
         return rawResourceSplices;
     }
 
+    private getIcons(decorations: theia.SourceControlResourceDecorations | undefined): IconUrl | undefined {
```
tsmaeder, line 511: Again, I would prefer `getIcon` singular here.

packages/plugin-ext/src/plugin/scm.ts
```diff
@@ -635,13 +642,14 @@ class SourceControlImpl implements theia.SourceControl {
...
         this.proxy.$registerSourceControl(this.handle, _id, _label, _rootUri);
+        this.plugin = plugin;
     }
 
     private createdResourceGroups = new Map<SsmResourceGroupImpl, Disposable>();
     private updatedResourceGroups = new Set<SsmResourceGroupImpl>();
 
     createResourceGroup(id: string, label: string): SsmResourceGroupImpl {
-        const group = new SsmResourceGroupImpl(this.proxy, this.commands, this.handle, id, label);
+        const group = new SsmResourceGroupImpl(this.proxy, this.commands, this.handle, this.plugin, id, label);
```
tsmaeder, line 652: Not in scope for this PR, but shouldn't it be `ScmResourceGroupImpl`?

packages/scm/src/browser/scm-provider.ts
```diff
@@ -58,6 +58,8 @@ export interface ScmResource {
 }
 
 export interface ScmResourceDecorations {
+    icon?: string;
```
tsmaeder, line 61: Not sure whether it would be a better idea to keep the conditional structure (either a ThemeIcon or a { dark: ..., light...} structure until we render the whole thing in the UI? It kinda doesn't make sense to have a dark and light ThemeIcon, as mentioned elsewhere, IMO.

tsmaeder:
>3. Modify the git extension files to pass a ThemeIcon (URI or codicon) as a decoration

@FernandoAscencio what do you mean by that?

packages/plugin-ext/src/plugin/scm.ts
```diff
@@ -111,8 +118,8 @@ function compareResourceThemableDecorations(a: theia.SourceControlResourceThemab
         return 1;
     }
 
-    const aPath = typeof a.iconPath === 'string' ? a.iconPath : a.iconPath.fsPath;
-    const bPath = typeof b.iconPath === 'string' ? b.iconPath : b.iconPath.fsPath;
+    const aPath = typeof a.iconPath === 'string' ? a.iconPath : URI.isUri(a.iconPath) ? a.iconPath.fsPath : (a.iconPath as theia.ThemeIcon).id;
```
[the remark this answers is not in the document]
FernandoAscencio: Yes, this can end up comparing a Theme Icon ID to a File System Path. This part was copied from [here](https://github.com/microsoft/vscode/blob/d1aa00acef8b2cfa88621c448b8fd8cd034f60a9/src/vs/workbench/api/common/extHostSCM.ts#LL55C1-L56C1)

As far as I am aware the Theme Icon IDs are codicon IDs. Thus, they should all be in lowercase.

packages/plugin-ext/src/plugin/scm.ts
FernandoAscencio: Thanks!
Will do.

packages/plugin-ext/src/plugin/scm.ts
FernandoAscencio: Yeah, looks like a typo.
I can add it to this PR.

packages/plugin-ext/src/main/browser/scm-main.ts
```diff
@@ -258,6 +261,12 @@ export class PluginScmProvider implements ScmProvider {
         this.onDidChangeResourcesEmitter.fire();
     }
 
+    private toIconClass(icons: IconUrl, isLightTheme: boolean): string {
+        let icon = typeof icons === 'object' ? (isLightTheme ? icons.light : icons.dark) : icons;
+        icon = icon.includes('codicon') ? icon : this.style.toIconClass(icon).object.iconClass;
```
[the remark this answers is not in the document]
FernandoAscencio: Yes. I modified it so now it adds it to disposables.

FernandoAscencio:
> > 3. Modify the git extension files to pass a ThemeIcon (URI or codicon) as a decoration
> 
> @FernandoAscencio what do you mean by that?

It has been a while so I have no idea. I'll change the steps so they make more sense.

tsmaeder:
@FernandoAscencio having done a bit more thinking about the general approach of this PR. In `plugin-icon-path.ts`, you import from VS Code in order to be able to resolve `ThemeIcon`s in the plugin host process. Using stuff from VS Code via import in the plugin host or the back-end in general is unusual. I think the right way to to this is to transfer the `ThemeIcon` as a typed object to the front end and to do the conversion to CSS there. As we do, for example, for tree items in `tree-views.ts`.

tsmaeder:
@FernandoAscencio about the "how to test": what are the changes you did to the git and git-base extensions? We already could have dark and light icons before, how does the testing procedure test the use of `ThemeIcon`?

packages/plugin-ext/src/plugin/scm.ts
FernandoAscencio: Currently there are two possible arguments for the `decorations` are of type `undefined` or `SourceControlResourceThemableDecorations`.

Then the function mostly cares about the `iconPath` attribute that can be `undefined` or of type `ThemeIcon`, `URI` or `string`. 

The case where `decorations` is not defined is covered by the first `if` statement. This also covers the case where `iconPath` is not defined within `decorations`. This is used mainly to determine if `decorations` has a `light` and `dark` icons. The other cases for `iconPath` being a URI, ThemeIcon or string are also accounted for, That last `else` is there for unexpected situations. I don't expect the warning to be called at all.

However, if a proper warning message should be placed then I think it is probably better to go for the value type instead of the argument. I am not sure how to do that though. For now, I'll add the value of `iconPath` to the message.

packages/plugin-ext/src/plugin/plugin-icon-path.ts
```diff
@@ -36,7 +37,7 @@ export namespace PluginIconPath {
         }
         return asString(iconPath, plugin);
     }
-    export function asString(arg: string | URI, plugin: Plugin): string {
+    export function asString(arg: string | URI | ThemeIcon, plugin: Plugin): string {
         arg = arg instanceof URI && arg.scheme === 'file' ? arg.fsPath : arg;
         if (typeof arg !== 'string') {
             return arg.toString(true);
```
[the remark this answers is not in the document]
FernandoAscencio: Thanks for the catch.
It should not be accepting `ThemeIcon` at all as it is filtered out before being used.

packages/plugin-ext/src/plugin/scm.ts
```diff
@@ -511,6 +513,15 @@ class SsmResourceGroupImpl implements theia.SourceControlResourceGroup {
         return rawResourceSplices;
     }
 
+    private getThemableIcon(icon: UriComponents | ThemeIcon | undefined): IconUrl | ThemeIcon | undefined {
```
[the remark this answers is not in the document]
FernandoAscencio: I agree.

packages/scm/src/browser/scm-tree-widget.tsx
```diff
@@ -567,7 +585,7 @@ export class ScmResourceComponent extends ScmElement<ScmResourceComponent.Props>
                 model,
                 treeNode
             }}>
-                <div title={tooltip} className='status' style={{ color }}>
+                <div title={tooltip} className={classNames.join(' ')} style={{ color }}>
```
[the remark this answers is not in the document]
FernandoAscencio: Yes.
This is mainly due to how the `status` class style is defined. If included, codicon icons will display smaller, URI icons will display normally. Maybe a restructuring for the `status` class style is required?

FernandoAscencio:
@tsmaeder 
Yes, `dark` and `light` icons were available before. This test makes it possible to test for the added ThemeIcon which is displayed as a globe codicon. URI is displayed as the blue M icon. These do not display in the current master when using the vsix files.

I will add for testing using current master.

FernandoAscencio:
@tsmaeder 
I think that is it for this PR.

Issue #12189, opened 2023-02-17
DebugSession.getInstructionBreakpoints does not return instruction breakpoints


martin-fleck-at:
### Bug Description:
The DebugSession does not return the instruction breakpoints with `getInstructionBreakpoints` but instead the function breakpoints.

### Steps to Reproduce:
1. Start debug session and add instruction breakpoints through the Disassembly view.
2. Alternatively: It is very clear from the code that it was simply an oversight that the wrong URI is used.

### Additional Information

- Operating System: all
- Theia Version: master

PR #12190, opened 2023-02-17
Properly return instruction breakpoints in DebugSession

martin-fleck-at:
Fixes https://github.com/eclipse-theia/theia/issues/12189

#### What it does
Fixes https://github.com/eclipse-theia/theia/issues/12189

#### How to test
Start a debug session, open the Disassembly view and add an instruction breakpoint. The debug view should be updated with the instruction breakpoint but it does not as the session does not return it properly.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-session.tsx
+++ b/packages/debug/src/browser/debug-session.tsx
@@ -1,5 +1,5 @@
 // *****************************************************************************
-// Copyright (C) 2018 Red Hat, Inc. and others.
+// Copyright (C) 2018-2023 Red Hat, Inc. and others.
 //
 // This program and the accompanying materials are made available under the
 // terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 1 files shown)


colin-grant-work (approved):
That explains - and fixes - that problem. 👍

packages/debug/src/browser/debug-session.tsx
```diff
@@ -1,5 +1,5 @@
 // *****************************************************************************
-// Copyright (C) 2018 Red Hat, Inc. and others.
+// Copyright (C) 2018-2023 Red Hat, Inc. and others.
```
vince-fugnitto, line 2: minor: I don't think we generally update the license headers when a file is modified.

vince-fugnitto (approved):
LGTM :+1:

packages/debug/src/browser/debug-session.tsx
martin-fleck-at: Ah, force of habit, thanks!

vince-fugnitto (approved):

Issue #12119, opened 2023-01-27
terminal: error thrown when attempting to set the default profile to `''`


vince-fugnitto:
#### Bug Description:

Following #12066 we have an error thrown on startup related to the default terminal-profile:

```
2023-01-27T13:29:59.848Z root ERROR Error: Cannot set default to unknown profile '' 
    at DefaultTerminalProfileService.setDefaultProfile (file:///home/foobar/workspaces/theia/examples/electron/lib/packages_terminal_lib_browser_terminal-profile-service_js-packages_terminal_lib_browser_termi-8fdc06.js:809:19)
    at file:///home/foobar/workspaces/theia/examples/electron/lib/packages_terminal_lib_browser_terminal-frontend-contribution_js.js:264:33
    at file:///home/foobar/workspaces/theia/examples/electron/lib/bundle.js:136667:69
    at CallbackList.invoke (file:///home/foobar/workspaces/theia/examples/electron/lib/bundle.js:136673:26)
    at Emitter.fire (file:///home/foobar/workspaces/theia/examples/electron/lib/bundle.js:136788:29)
    at DefaultTerminalProfileService.handleAdded (file:///home/foobar/workspaces/theia/examples/electron/lib/packages_terminal_lib_browser_terminal-profile-service_js-packages_terminal_lib_browser_termi-8fdc06.js:796:33)
    at file:///home/foobar/workspaces/theia/examples/electron/lib/packages_terminal_lib_browser_terminal-profile-service_js-packages_terminal_lib_browser_termi-8fdc06.js:769:26
    at file:///home/foobar/workspaces/theia/examples/electron/lib/bundle.js:136667:69
```

#### Steps to Reproduce:

You can view the error during [CI](https://github.com/eclipse-theia/theia/actions/runs/4018619820/jobs/6905186854#step:9:329)

1.start the application with no previous default profile data
2. notice the error in the backend logs on startup

PR #12191, opened 2023-02-17
Handle the case where no default terminal profile is set no startup

tsmaeder:
#### What it does
We are handling the `contributes.terminal` extension point later than we are initializing terminal profiles from the preferences. So the default profile in the preferences might not be present yet in the terminal profile registry when we read the preferences. So we try to set the default whenever a new profile is added. In this endeavor, we were not handling the case that no default profile is set in the preferences.

Fixes https://github.com/eclipse-theia/theia/issues/12119

#### How to test
1. Make sure you have not default profile setting in the Theia preferences
2. Make sure you have the default extensions downloaded
3. Start the electron version of Theia
4. Observe: you do not get the exception from the issue
5. Now set a default profile to "Javascript Debug Terminal"
6. Restart Theia
7. Observe: the default terminal is correctly set.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/terminal/src/browser/terminal-frontend-contribution.ts
+++ b/packages/terminal/src/browser/terminal-frontend-contribution.ts
@@ -256,7 +256,7 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
 
         // extension contributions get read after this point: need to set the default profile if necessary
         this.profileService.onAdded(id => {
-            let defaultProfileId = undefined;
+            let defaultProfileId: string | undefined;
             switch (OS.backend.type()) {
                 case OS.Type.Windows: {
                     defaultProfileId = this.terminalPreferences['terminal.integrated.defaultProfile.windows'];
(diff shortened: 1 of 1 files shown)


packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -254,9 +254,9 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
         });
         this.mergePreferencesPromise = this.mergePreferencesPromise.finally(() => this.mergePreferences());
 
+        // extension contributions get read after this point: need to set the default profile if necessary
         this.profileService.onAdded(id => {
-            // extension contributions get read after this point: need to set the default profile if necessary
-            let defaultProfileId;
+            let defaultProfileId = undefined;
```
msujew, line 259: Minor: This gets implicitly typed as `any`
```suggestion
            let defaultProfileId: string | undefined;
```

msujew (approved):
Looks good to me. I can't reproduce the issue anymore 👍

packages/terminal/src/browser/terminal-frontend-contribution.ts
tsmaeder: Strange: it seems trivial to infer the type here.

msujew: TypeScript doesn't seem to do it any ¯\_(ツ)_/¯

PR #12196, opened 2023-02-17
Improve transition to/from "offline" state

### LINKED ISSUE #12194: ConnectionStatusService is slow to recognize "offline" when backend processes are consuming resources
### Bug Description:

If a backend process is hogging resources, the `ConnectionStatusService`'s `this.performPingRequest()` call will be blocked, preventing the `offline` status bar warning from showing up. This can cause issues since no frontend<->backend requests can be reliably fulfilled during that time and the frontend gives no warning that the connection is broken/interrupted

### Steps to Reproduce:

1. Build this experimental [branch](https://github.com/kenneth-marut-work/theia/tree/hang-backend) which adds a command "hang backend".
2. Start Theia for browser and run the "hang backend" command
3. Try to interact with the IDE and notice that you can't
4. After about 1 minute the IDE will go "offline", even though it should really report this status much sooner
5. After 3 minutes of being stuck in the "hang backend" loop the IDE will recover

### Additional Information

- Operating System:
Linux
- Theia Version:
master

### LINKED ISSUE #12195: Dirty editors cannot be saved after application recovers from "offline" mode
### Bug Description:

If a backend process hogs resources and blocks frontend<->backend communication the application will go "offline" and eventually recover. During this time, the frontend can be slow to recognize that connection is interrupted as per #12194 and a user may continue to make edits to a file, putting it into a dirty state. When the application finally recovers from "offline" mode, it is impossible to save any new changes to the existing editor. The editor must be closed and reopened.

### Steps to Reproduce:

1. Build this [branch](https://github.com/kenneth-marut-work/theia/tree/hang-backend) which adds a command to hang the backend for 3 minutes.
2. Start Theia for browser and open an editor (`FileA.txt`)
3. Run the command "hang backend"
4. Start making edits to `FileA.txt` and try to save them (they will fail because the backend is blocked, notice also that there is no indication that there is a failure)
5. Eventually (about 1 min) the status bar will turn yellow (#12194), after 3 minutes the "hang backend" command will complete and  the status bar will return to blue. The application should be interactive again
6. notice that you still cannot save your edits you made to `FileA.txt`

### Additional Information

- Operating System:
- Linux
- Theia Version:
Master

kenneth-marut-work:
#### What it does
Fixes #12194 by adding a timeout to `ConnectionStatusService`'s `ping` service to make sure frontend is aware of network interruptions *sooner* if backend is hanging.

Fixes #12195 by cancelling save/sync in `monaco-editor-model` when `ConnectionStatusService`'s `onStatusChanged` event fires preventing scheduled saves/syncs from never resolving when backend goes offline
 
The `hang-backend` package and command will be removed before merging this PR.

**Breaking Changes**
Modifies constructor args for `MonacoEditorModel` to pass in `ConnectionStatusService`
#### How to test

**To test #12194:**

1. Build this branch
2. Start Theia in browser or electron mode
3. run command "hang backend" to kick off a 30second backend loop
4. After 10s, a connection interruption should be detected by the frontend (2 * the usual ping timeout)
5. After the 30 second loop completes, the frontend will recover
6. Compare with the usual bad behavior described [here](https://github.com/eclipse-theia/theia/issues/12194)

**To test #12195:**

1. Open an editor under normal conditions
2. make some changes to that file and try to save them, observe normal behavior
3. Now execute "hang backend" command
4. Try to save that file, observe that it doesn't work
5. Frontend will recognize interruption after 10s, observe that saving still does not work
6. After 30 seconds, frontend will recover, observe that saving *does* work now
7. Compare with bad behavior described [here](https://github.com/eclipse-theia/theia/issues/12195)
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-editor-model.ts
+++ b/packages/monaco/src/browser/monaco-editor-model.ts
@@ -16,6 +16,7 @@
 
 import { Position, Range, TextDocumentSaveReason, TextDocumentContentChangeEvent } from '@theia/core/shared/vscode-languageserver-protocol';
 import { TextEditorDocument, EncodingMode, FindMatchesOptions, FindMatch, EditorPreferences } from '@theia/editor/lib/browser';
+import { ConnectionStatusService } from '@theia/core/lib/browser/connection-status-service';
 import { DisposableCollection, Disposable } from '@theia/core/lib/common/disposable';
 import { Emitter, Event } from '@theia/core/lib/common/event';
 import { CancellationTokenSource, CancellationToken } from '@theia/core/lib/common/cancellation';
@@ -90,6 +91,7 @@ export class MonacoEditorModel implements IResolvedTextEditorModel, TextEditorDo
         protected readonly resource: Resource,
         protected readonly m2p: MonacoToProtocolConverter,
         protected readonly p2m: ProtocolToMonacoConverter,
+        protected readonly connectionStatusService: ConnectionStatusService,
         protected readonly logger?: ILogger,
         protected readonly editorPreferences?: EditorPreferences
     ) {
--- a/packages/monaco/src/browser/monaco-editor-model.ts
+++ b/packages/monaco/src/browser/monaco-editor-model.ts
@@ -107,6 +107,10 @@ export class MonacoEditorModel implements IResolvedTextEditorModel, TextEditorDo
         this.resolveModel = this.readContents().then(
             content => this.initialize(content || '')
         );
+        this.toDispose.push(this.connectionStatusService.onStatusChange(() => {
+            this.cancelSave();
+            this.cancelSync();
+        }));
     }
 
     dispose(): void {
--- a/examples/browser/package.json
+++ b/examples/browser/package.json
@@ -26,6 +26,7 @@
     "@theia/filesystem": "1.34.0",
     "@theia/getting-started": "1.34.0",
     "@theia/git": "1.34.0",
+    "@theia/hang-backend": "1.34.0",
     "@theia/keymaps": "1.34.0",
     "@theia/markers": "1.34.0",
     "@theia/memory-inspector": "1.34.0",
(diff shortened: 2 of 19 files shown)


packages/monaco/src/browser/monaco-editor-model.ts
```diff
@@ -90,6 +91,7 @@ export class MonacoEditorModel implements IResolvedTextEditorModel, TextEditorDo
         protected readonly resource: Resource,
         protected readonly m2p: MonacoToProtocolConverter,
         protected readonly p2m: ProtocolToMonacoConverter,
+        protected readonly connectionStatusService: ConnectionStatusService,
```
colin-grant-work, line 94: This should get a 'breaking changes' mention.

packages/monaco/src/browser/monaco-editor-model.ts
```diff
@@ -105,6 +107,10 @@ export class MonacoEditorModel implements IResolvedTextEditorModel, TextEditorDo
         this.resolveModel = this.readContents().then(
             content => this.initialize(content || '')
         );
+        this.toDispose.push(this.connectionStatusService.onStatusChange(() => {
+            this.cancelSave();
+            this.cancelSync();
+        }));
```
colin-grant-work, line 113: The this is where the problem manifested itself, but I wonder if the the problem shouldn't be fixed in the `FileResource` code, since that's where the saving happens, and it's the connection to the backend file system that breaks saving / loading new content?

colin-grant-work (commented):
This fixes the clearest manifestation of the problem, but I wonder if the fix is in the right place. See my comment on the changes to the `monaco-editor-model` file.

vince-fugnitto:
I believe we can close the pull-request for the time being, and revisit as necessary.
Both Kenneth and Colin are busy with other projects.

PR #12220, opened 2023-02-24
Add boolean suppression DebugSessionOptions

### LINKED ISSUE #12015: [vscode] Support DebugSessionOptions boolean options for suppression
see https://code.visualstudio.com/api/references/vscode-api#DebugSessionOptions

4 new options since 1.73:

- suppressDebugStatusbar?: boolean
_When true, the window statusbar color will not be changed for this session._

- suppressDebugToolbar?: boolean
_When true, the debug toolbar will not be shown for this session._

- suppressDebugView?: boolean
_When true, the debug viewlet will not be automatically revealed for this session._

- suppressSaveBeforeStart?: boolean
_When true, a save will not be triggered for open editors when starting a debug session, regardless of the value of the debug.saveBeforeStart setting._

sgraband:
#### What it does

Add options to suppress the debugging statusbar, toolbar and view. 
As well as the saving before starting a debugging session.

Fixes #12015 

_Note:_
Parent/child sessions are currently handled as followed:
- For the saving supression only the root session is taken into account as this could have unintended side effects (saving files when not wanted), if a child session specifies something else than the parent session.
- For the other three supressions (debug view, toolbar and statusbar) the value of the current session is evaluated. If there is no value the nearest parent session will be evaluated.

I mostly went with two different approaches here due to the significance of the actions (saving vs graphical-only), but your input is welcome and i will gladly unify the approaches, when the "better" solution is identified.

#### How to test

- Install this extension: [helloworld-sample-0.0.1.zip](https://github.com/eclipse-theia/theia/files/10811670/helloworld-sample-0.0.1.zip)
It registers a few commands that start the "Launch Browser Backend" debug config with different options set.
- Open theia and the theia repository as the workspace.
- Add the following to your setting.json file:

>   "debug.saveBeforeStart": "allEditorsInActiveGroup",
>   "debug.openDebug": "openOnSessionStart",
>   "files.autoSave": "off",

- Trigger one of the `Start Debugging` commands. It says in parenthesis which options are enabled.
- See if the behavior is correct:

> suppressDebugStatusbar?: boolean
> When true, the window statusbar color will not be changed for this session.
> 
> suppressDebugToolbar?: boolean
> When true, the debug toolbar will not be shown for this session.
> 
> suppressDebugView?: boolean
> When true, the debug viewlet will not be automatically revealed for this session.
> 
> suppressSaveBeforeStart?: boolean
> When true, a save will not be triggered for open editors when starting a debug session, regardless of the value of the debug.saveBeforeStart setting.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-frontend-application-contribution.ts
+++ b/packages/debug/src/browser/debug-frontend-application-contribution.ts
@@ -54,6 +54,7 @@ import { DebugFunctionBreakpoint } from './model/debug-function-breakpoint';
 import { DebugBreakpoint } from './model/debug-breakpoint';
 import { nls } from '@theia/core/lib/common/nls';
 import { DebugInstructionBreakpoint } from './model/debug-instruction-breakpoint';
+import { DebugConfiguration } from '../common/debug-configuration';
 
 export namespace DebugMenus {
     export const DEBUG = [...MAIN_MENU_BAR, '6_debug'];
@@ -451,14 +452,14 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
             }
             const shouldOpenDebug = openDebug === 'openOnSessionStart' || (openDebug === 'openOnFirstSessionStart' && this.firstSessionStart);
             // Do not open debug view when suppressed via configuration
-            if (!noDebug && !this.shouldViewBeSuppressed(session) && shouldOpenDebug) {
+            if (!noDebug && !this.getOption(session, 'suppressDebugView') && shouldOpenDebug) {
                 this.openSession(session);
             }
             this.firstSessionStart = false;
         });
         this.manager.onDidStopDebugSession(session => {
             const { openDebug } = session.configuration;
-            if (!this.shouldViewBeSuppressed(session) && openDebug === 'openOnDebugBreak') {
+            if (!this.getOption(session, 'suppressDebugView') && openDebug === 'openOnDebugBreak') {
                 this.openSession(session);
             }
         });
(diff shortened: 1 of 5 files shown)


packages/debug/src/browser/debug-session-manager.ts
```diff
@@ -210,7 +210,8 @@ export class DebugSessionManager {
     protected async startConfiguration(options: DebugConfigurationSessionOptions): Promise<DebugSession | undefined> {
         return this.progressService.withProgress('Start...', 'debug', async () => {
             try {
-                if (!await this.saveAll()) {
+                // If a parent session is available saving should be handled by the parent
+                if (!options.configuration.parentSessionId && !options.configuration.suppressSaveBeforeStart && !await this.saveAll()) {
```
tsmaeder, line 214: This seems to be a drive-by fix. Was our behaviour broken before? If so, what was the bug?

packages/debug/src/common/debug-configuration.ts
```diff
@@ -72,6 +72,21 @@ export interface DebugConfiguration {
...
     /** Task to run after debug session ends */
     postDebugTask?: string | TaskIdentifier;
+
+    /**
+     * When true, a save will not be triggered for open editors when starting a debug session,
+     * regardless of the value of the `debug.saveBeforeStart` setting.
+     */
+    suppressSaveBeforeStart?: boolean;
+
+    /** When true, the debug toolbar will not be shown for this session. */
```
tsmaeder, line 82: As mentioned above, I don't think this flag makes sense in Theia. If Theia does not use it, It shouldn't appear in Theia code.

tsmaeder (commented):
A couple of questions/suggestions

packages/debug/src/browser/debug-frontend-application-contribution.ts
```diff
@@ -1494,10 +1496,36 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
...
+            } else {
+                break;
+            }
+        }
+        return result;
+    }
+
+    protected shouldViewBeSuppressed(session: DebugSession): boolean | undefined {
+            // Only suppress the view if both the debug view and the toolbar are suppressed, as the toolbar is always docked.
+            return this.getOption(session, 'suppressDebugView') && this.getOption(session, 'suppressDebugToolbar');
```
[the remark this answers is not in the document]
sgraband: Mhm.. I see your point. My reasoning behind this was that if the extensions sets `suppressDebugView: true` but does not set `suppressDebugToolbar` they might expect to see the toolbar. 

But you are probably right, that we should just ignore the `suppressDebugToolbar` all together. I will change this.

packages/debug/src/browser/debug-session-manager.ts
sgraband: No this is basically just what i described here:

> Note:
Parent/child sessions are currently handled as followed:

> For the saving supression only the root session is taken into account as this could have unintended side effects (saving files when not wanted), if a child session specifies something else than the parent session.

I wanted to make sure, that if the config that is executed specifies `suppressSaveBeforeStart: true` the save is never executed, e.g. if child configs do not specify the flag. WDYT?

tsmaeder:
There does not seem to be an option in the settings UI for `debug.saveBeforeStart`. Is this maybe missing in the preference schema?

tsmaeder (changes requested):
I'm fine with the changes, but we need to decide whether the support for the `saveOnStart` preference should be in this PR or not. If not, could you please open an issue so we can track the problem?

sgraband:
I created #12319 for the follow up.

tsmaeder (approved):
Looks good to me now.

PR #12222, opened 2023-02-24
[browser tests] update Puppeteer

marcdumais-work:
#### What it does

The version of Puppeteer (and few related dependencies) we had is quite old and uses a version of Chromium that does not support some of the DOM API functions we use in the current code base. e.g. HTMLElement.replaceChildren() \[1\], that we use in hover-service.ts

The latest version seems to work well, and supports the example above,  so I went with that. Note that Puppeteer now bundles its typing file, and that some of the APIs have changed slightly, requiring a small adaptation on our end.

\[1\]: https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceChildren#browser_compatibility

#### How to test

- The browser test should still pass in CI
- you may also try running them locally: `yarn && yarn download:plugins && yarn browser build && yarn browser test`
- you may also confirm that it's still possible to run the suite to gather coverage information: 
``` bash
cd example/browser
yarn -s rebuild && npx theia test . --plugins=local-dir:../../plugins --test-coverage  --test-spec=../api-tests/src/*.spec.js
npx nyc report --reporter=lcov --reporter=text
```
note: for the code coverage, source-map is not currently configured, so the results will be against the browser app "compiled .js files", not the original Theia sources. Anyway, the immediate goal is just to confirm that the code coverage infrastructure still works after the update.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
LGTM :+1: 

- [x] CI successfully passes for playwright
- [x] CI successfully passes for the API integration tests
- [x] `yarn browser test` passes locally

vince-fugnitto (approved):

PR #12224, opened 2023-02-24
Reduce plugin logging to debug level

colin-grant-work:
#### What it does

At the moment, on startup and sometimes on shutdown, we log a separate message for various lifecycle events for every single plugin in the application. That's a lot of logs, most of which are completely uninformative. This PR converts them to `debug` level logs so that they are still available if they're of interest but won't be displayed routinely.

#### How to test

1. Start the application.
2. Everything should work as normal.
3. There should be a couple hundred fewer lines in your log.

#### Review checklist

#### Reminder for reviewers


msujew (approved):
Looks good to me 👍

Issue #12217, opened 2023-02-23
Newlines are doubled in HTML


gbodeen:
### Bug Description:

Hitting enter in an HTML file causes two newlines to be inserted.

![image](https://user-images.githubusercontent.com/20653241/221026277-04611485-5c75-4270-a78f-4cd1aa2dbeda.png)

### Steps to Reproduce:

1. Use the Theia repo as the workspace. Open `examples/electron/src-gen/frontend/index.html`, place the cursor at the end of a line, and hit Enter.

### Additional Information

- Operating System: RHEL7
- Theia Version: 1.34.0

PR #12212, opened 2023-02-23
repo: pin `ts` plugins

vince-fugnitto:
#### What it does

The commit pins the version of typescript plugins used as builtins in the repository due to an issue when modifying contents of the file. The source of the bug still needs to be investigated further when we upgrade again.

The issue results in newlines (<kbd>enter<kbd>) producing odd output:

https://user-images.githubusercontent.com/40359487/220960512-07510f6b-450a-4ca9-b054-9f5e20ff3f74.mp4

#### How to test

1. start the application and open a ts file
2. language features, syntax highlighting should work as expected
3. add newlines to a ts file, the content should be as expected

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #12228, opened 2023-02-24
Handle OnEnterRule patterns

gbodeen:
#### What it does
Fixes https://github.com/eclipse-theia/theia/issues/12217 and the issue raised in https://github.com/eclipse-theia/theia/pull/12212 .

The format of `OnEnterRule` had been defined with `beforeText`, `afterText`, and `previousLineText` as only strings. However they are often quasi-RegExp objects with `pattern` and `flags` string fields.

#### How to test

1. With TypeScript plugins 1.66.2 or later, in a JavaScript or TypeScript file, if you hit Enter to create a new line, do you still get ` * ` appended in appropriately?
2. Similarly in HTML files, if you hit Enter to create a new line, do you still get two new lines inappropriately?
3. Do other insertions (e.g. for JsDoc strings) still work properly?

#### Review checklist

Note: I've only tested on RHEL7, electron & browser modes, and a limited set of circumstances involving insertions of newlines.

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/common/plugin-protocol.ts
+++ b/packages/plugin-ext/src/common/plugin-protocol.ts
@@ -726,10 +726,15 @@ export interface FoldingRules {
     markers?: FoldingMarkers;
 }
 
+export interface OnEnterRuleRegExp {
+    pattern: string;
+    flags?: string;
+}
+
 export interface OnEnterRule {
-    beforeText: string;
-    afterText?: string;
-    previousLineText?: string;
+    beforeText: string | OnEnterRuleRegExp;
+    afterText?: string | OnEnterRuleRegExp;
+    previousLineText?: string | OnEnterRuleRegExp;
     action: EnterAction;
 }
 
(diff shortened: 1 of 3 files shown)


packages/plugin-ext/src/common/plugin-protocol.ts
```diff
@@ -726,10 +726,15 @@ export interface FoldingRules {
     markers?: FoldingMarkers;
 }
 
+export interface OnEnterRuleRegExp {
+    pattern: string;
+    flags?: string;
+}
```
colin-grant-work, line 732: Looking at [this file](https://github.com/microsoft/vscode/blob/d912e6c92514b3dd524fe6ff1332a06c5f1d5534/src/vs/workbench/contrib/codeEditor/browser/languageConfigurationExtensionPoint.ts#L22-L25) in VSCode, it looks like there are a number of language configuration options that refer to this interface; shall we update all of the ones we know of - IndentationRule and FoldingRule - here?

gbodeen: Oh yes, I suppose so.

gbodeen: OK, I've adjusted the plugin-protocol interfaces to match. They all rely on the same createRegex method, so I don't think any additional runtime changes are necessary to handle those.

colin-grant-work (approved):
These changes address the problems previously observed with recent language feature plugins. 👍

vince-fugnitto (approved):
LGTM :+1: 
I confirmed that:
- [x] the `onEnterRule` are properly applied (ex: comments)
- [x] we no longer get the odd behavior for HTML, or TypeScript for newlines

ccorn:
This means that 2f287df2050fa8e3cad1cc030edbc268607372f4 can be reverted, right?

vince-fugnitto:
@ccorn the changes already remove the pinned builtins:

https://github.com/eclipse-theia/theia/pull/12228/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L106-R106

PR #12236, opened 2023-02-27
plugin-ext: Change `PluginsKeyValueStorage` to not continuously access the disk

### LINKED ISSUE #8384: [python] plugin-storage/global-state.json: Unexpected end of JSON input
Python extension is trying to read storage files and fails:
```
root ERROR Failed to parse data from " /home/gitpod/.theia/plugin-storage/global-state.json ". Reason: SyntaxError: /home/gitpod/.theia/plugin-storage/global-state.json: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at /workspace/theia/node_modules/jsonfile/index.js:33:18
    at /workspace/theia/node_modules/graceful-fs/graceful-fs.js:115:16
    at FSReqCallback.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:61:3)
````

### PR DESCRIPTION
**Pull request by marcdumais-work**
#### What it does

Fixes #8384

Added the use of a semaphore to lock access to the global plugin storage file, to avoid it being accessed in a disorderly fashion, causing potential corruption.

As per the issue discussion, it's possible that there is also a potential race condition caused by node's write function, not flushing its buffers to disk, but in this specific case, it was obvious that parallel calls to set() and get() were the main cause of the problems.

#### How to test

One way is to add ENTER/EXIT debug traces [1] to `packages/plugin-ext/src/main/node/plugins-key-value-storage.ts`, in `set()` and `get()` and then rebuild , run the browser tests ('yarn browser test`), and confirm that the accesses are now matched. i.e. "ENTER get()" should always be followed by "EXIT get()", and "ENTER set()" always followed by "EXIT set(). 

\[1\]
```  diff
diff --git a/packages/plugin-ext/src/main/node/plugins-key-value-storage.ts b/packages/plugin-ext/src/main/node/plugins-key-value-storage.ts
index 7200ce2e049..c0a7e91a325 100644
--- a/packages/plugin-ext/src/main/node/plugins-key-value-storage.ts
+++ b/packages/plugin-ext/src/main/node/plugins-key-value-storage.ts
@@ -57,6 +57,7 @@ export class PluginsKeyValueStorage {
     }
 
     async set(key: string, value: KeysToAnyValues, kind: PluginStorageKind): Promise<boolean> {
+        console.log('ENTER set()');
         try {
             await lock.acquire();
             const dataPath = await this.getDataPath(kind);
@@ -77,10 +78,12 @@ export class PluginsKeyValueStorage {
             return true;
         } finally {
             lock.release();
+            console.log('EXIT set()');
         };
     }
 
     async get(key: string, kind: PluginStorageKind): Promise<KeysToAnyValues> {
+        console.log('ENTER get()');
         try {
             await lock.acquire();
             const dataPath = await this.getDataPath(kind);
@@ -91,10 +94,12 @@ export class PluginsKeyValueStorage {
             return data[key];
         } finally {
             lock.release();
+            console.log('EXIT get()');
         };
     }
 
     async getAll(kind: PluginStorageKind): Promise<KeysToKeysToAnyValue> {
+        console.log('ENTER getAll()');
         try {
             await lock.acquire();
             const dataPath = await this.getDataPath(kind);
@@ -104,6 +109,7 @@ export class PluginsKeyValueStorage {
             return this.readFromFile(dataPath);
         } finally {
             lock.release();
+            console.log('EXIT getAll()');
         }
     }

```

#### Review checklist

#### Reminder for reviewers


paul-marechal:
I pushed changes to use another semaphore library that is already used by Theia: `async-mutex`.

I also cleaned up the class to get rid of the asynchronous `postConstruct` as this is a mistake when using Inversify 4.

Lastly, is there a way to write browser tests for this fix? From what I understand this class is quite nested in the plugin system infrastructure, so if it's too difficult writing integration tests please write unit tests then.

Here's an example for a test I think could be good:

```ts
async function increment() {
    const sum = await storage.get('sum');
    await storage.set('sum', sum + 1);
}
const tasks = [];
for (let i = 0; i < 1000; i++) {
    tasks.push(increment());
}
await Promise.all(tasks);
assert.deepStrictEqual(
    await storage.get('sum'),
    1000,
    'the sum should have been incremented 1000x'
);
```

paul-marechal:
The last commit I pushed adds logic to use unique mutexes based on the file path being written to. It seems to be something that can happen based on the `kind` parameter passed to several methods. This is better than locking access to unrelated files than using the same global lock.

colin-grant-work:
One thing to consider, at least re: writing: there's already something doing a somewhat similar job for our preference files on the frontend in `packages/preferences/src/browser/preference-transaction-manager.ts`. The only thing missing for it to work on the backend is a backend equivalent to the `MonacoJSONCEditor` in `packages/preferences/src/browser/monaco-jsonc-editor.ts`

paul-marechal:
@colin-grant-work is it me or the component you linked seems hyper-specific to preferences?

The `FileSystemLocking` service introduced here is mostly a generic mutex bag and doesn't make any assumption about what you might do while having exclusive access (use the node APIs or don't)

colin-grant-work (commented):
I'm a bit confused about the lifecycle of this system: it seems to me we only need to read from the file once at startup, and after that, all changes can be kept in memory and written to the underlying storage only at the end of the session (or periodically to avoid breakdown). That should mean that we never have a problem with concurrent reading and writing.

colin-grant-work:
> @colin-grant-work is it me or the component you linked seems hyper-specific to preferences?

Yes, but it would do the job that we need to do for the `globalState` mechanism, since that also just manages a JSON file.

packages/core/src/node/filesystem-locking.ts
```diff
@@ -0,0 +1,89 @@
...
+     * @param lockPath The path to request exclusive access to.
+     * @param transaction The job to do while having exclusive access.
+     * @param thisArg `this` argument used when calling `transaction`.
+     */
+    lockPath<T>(lockPath: string | URI, transaction: (lockPath: string) => T | Promise<T>, thisArg?: unknown): Promise<T>;
+    /**
+     * Check if someone already has exclusive access to a file.
+     * @param lockPath The path to test if access is restricted.
+     */
+    isPathLocked(lockPath: string | URI): boolean;
```
colin-grant-work, line 39: It doesn't seem that this is every referred to, and it seems that the intended API is to lock the path and then perform whatever operation (reading or writing), that needs to be done?

paul-marechal: This is a left over, will remove.

colin-grant-work:
I also think that @tsmaeder's point about this being an OS-level problem merits serious consideration, because two Theia instances running on the same machine will both want access to the same files, and as long as we're not using the single-instance lock to ensure that we only have one backend, we've got no guarantee that they won't be operating on the file at the same item. It also increases the urgency of keeping the file up to date with regular writes, because the second instance would want to have an up-to-date version of the data when it starts up. But that means that all changes need to be written promptly to disk, and that means that the probability of OS-level concurrent access is non-negligible.

paul-marechal:
Looking at "system file locking" was inconclusive, and only the `proper-lockfile` package looks usable which would prevent two Theia instances from conflicting. Other processes can still mess up with our files, but at this point it's out of scope.

It then depends on what kind of behaviour we want: The current approach assumes any update is directly written on disk and visible by subsequent accesses. I'm not sure of the benefits from this approach.

paul-marechal:
@colin-grant-work I used the strategy you mentioned where we use an in-memory cache first for the values, and only sync with the disk every so often. The sync delay is long enough with a bit of randomness as to avoid collisions with other Theia instances. It might not be fool-proof but it's arguably better than the previous implementation. I think we'll have a long and happy life before we'll get to see problems with this new implementation, but I might be wrong.

marcdumais-work:
> Well, `proper-lockfile` doesn't seem to work when the file doesn't yet exist.

Have you tried to disable the [realPath option](https://github.com/moxystudio/node-proper-lockfile#lockfile-options)? Else creating an empty file first, if/when it doesn't yet exists, looks like a viable option (the file itself is not used as the lock, so this should not introduce race conditions).

marcdumais-work:
The scope of the PR has changed quite a bit, so probably the title/description should be updated accordingly.

paul-marechal:
To @msujew, @JonasHelming, can someone from either TypeFox or EclipseSource review this patch?

vince-fugnitto (approved):
LGTM 👍 

- [x] confirmed that the code looks good
- [x] confirmed that the unit-tests pass successfully 
- [x] confirmed that the `global-state.json` error no longer occurs

msujew (approved):
The changes look good to me as well 👍

PR #12239, opened 2023-02-28
monaco: quick input regression fix

FernandoAscencio:
#### What it does
Fixes a regression  introduced in #12095 

#### How to test
Currently the commit in #12095 introduced an regression with some menus did not display properly.

12095Regression

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #12247, opened 2023-03-01
Align quickpick labels

gbodeen:
#### What it does

This is a very minor change. In the quickpick menu when searching for a file, some results have a generic icon image. For those results, the path is vertically offset.

image

With the change:

image

A more perfect change would refactor the CSS to avoid the unnecessary hard-coded pixel values in `monaco-editor-core` that lead to this. However, that would also have complex effects. The advantage of this PR's change is that it's only within a highly specific CSS selector, so should not interfere anywhere else.

#### How to test

1. Open the quickpick to search for a file (`Ctrl + P`). Search for various strings and observe the icons in the list of results. For only the generic file icon, Theia on `main` has misaligned path labels.
2. On this branch, the path labels should be correctly aligned.

#### Review checklist

#### Reminder for reviewers


gbodeen:
I'm going to address this problem in a different manner that also handles a few related problems.

gbodeen:
I see a problem with the underscores. I'll aim to fix that soon.

gbodeen:
OK, now this also fixes the debug quickpick similarly as in https://github.com/eclipse-theia/theia/pull/12239 👍 

image

colin-grant-work (dismissed):
All the listed scenarios, including those from #12239, look good to me.

vince-fugnitto:
@gbodeen I've noticed issues with other file-icon themes such as `seti`:

image

gbodeen:
seti icon theme:

image

theia icon theme:

image

debug menu still good:

image

...so empirically this draft is looking better.  I think it's also theoretically better, as the current approach is:
* make the `codicon default-file-icon` pseudoelements have essentially all the same properties as the `theia-file-icons-js file-icon` pseudoelements
* don't allow the wrong font & font-size to leak out of the `codicon default-file-icon` elements
* still unset the flex property that was inherited from elsewhere and creating the debug problem

FernandoAscencio:
Found an issue with the check marks next to toggle commands.
Theia browser Microsoft Edge: 110.0.1587.57

alignmentBrowserEdge

gbodeen:
> @gbodeen do you mind explaining the change from [f41e71b](https://github.com/eclipse-theia/theia/commit/f41e71b56722fc4ee57e6e04d7cf9a470cb79a1b)? 

Perhaps I misunderstood. I thought the checkboxes were never supposed to be visible. My second draft, as discovered by Fernando above, had an unexpected side effect of making them visible under certain circumstances where they had not previously been visible.

What is the intended behavior? (I'll be able to touch this again later this week.)

FernandoAscencio:
>I thought the checkboxes were never supposed to be visible.

Those are not checkboxes but check codicons that should appear when a toggle command is "on".
The issue is that they are not aligned with the rest of the text.

gbodeen:
Interesting. I'd never seen them before and presumed they were only used in the code, not something that should be visible to users. 

In any case, it now follows the same approach as above, applying all the same properties as used for icons in other areas of the code, and the result is well aligned:

image

vince-fugnitto:
@msujew I still noticed an issue and possibly others where the icons are not properly aligned.
For example:

https://user-images.githubusercontent.com/40359487/232528684-7b0d074b-4f6c-4f2b-a254-da1c62bb3a17.mp4

vince-fugnitto:
Closing in favor of the updates from https://github.com/eclipse-theia/theia/pull/12239.

PR #12248, opened 2023-03-01
playwright: cleanup build process

vince-fugnitto:
#### What it does

The pull-request cleans up the build process which previously would automatically install playwright dependencies such as chromium when performing `yarn` on the repo (due to the `prepare` script). The change cleans up the build so that playwright can be built only when needed (ex: when we need to run the suite) and not unnecessarily.

#### How to test

- [x] confirm that CI successfully passes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -33,7 +33,7 @@ jobs:
         with:
           python-version: "3.x"
 
-      - name: Build
+      - name: Build Browser
         shell: bash
         run: |
           yarn --skip-integrity-check --network-timeout 100000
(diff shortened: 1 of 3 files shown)


msujew (approved):
LGTM 👍

vince-fugnitto:
@planger do you have any objections to the following approach?

planger:
Thanks for pinging me @vince-fugnitto! I don't have any objections. Good idea to minify if Playwright isn't needed.
I just think we should also update `examples/playwright/docs/DEVELOPING.md` and mention that users should run `yarn build` in `examples/playwright` now, am I right?

PR #12249, opened 2023-03-01
Improve Breakpoint Decoration Rendering

colin-grant-work:
#### What it does

Fixes another part of #11878 by setting decorations on the editor class rather than editor model class. In the latter case, all decorations were shared by all editors referring to the model. With this code, each editor owns a different set of decorations that refer to common state, where appropriate.

#### How to test

1. Follow instructions from https://github.com/eclipse-theia/theia/pull/12183.
2. Hover over the breakpoint gutter for one of the split editors, but don't add a breakpoint
3. You should see a lighter / less opaque breakpoint marking, but _only_ in the editor whose gutter you're hovering over.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
I confirm that the changes work as expected :+1: Setting breakpoints, and debugging still works.

PR #12251, opened 2023-03-02
Add `isObject` check to ProblemMarker

### LINKED ISSUE #12250: ProblemsView context menu only shows 'Collapse All'
### Bug Description:
After the update to Theia 1.34 the problems view only shows one context entry: Collapse All
ProblemsView

Furthermore the following stack trace is logged:
```
2023-03-02T11:59:52.043Z root ERROR TypeError: Cannot use 'in' operator to search for 'kind' in undefined
    at Object.is (http://localhost:3000/packages_markers_lib_browser_problem_problem-manager_js.js:296:23)
    at Object.is (http://localhost:3000/packages_markers_lib_browser_problem_problem-widget_js.js:409:78)
    at CommandHandler.toSelection (http://localhost:3000/packages_markers_lib_browser_problem_problem-widget_js.js:414:61)
    at CommandHandler.getSelection (http://localhost:3000/packages_filesystem_lib_browser_breadcrumbs_filepath-breadcrumb_js-packages_filesystem_lib_br-eff824.js:50:52)
    at CommandHandler.isVisible (http://localhost:3000/packages_filesystem_lib_browser_breadcrumbs_filepath-breadcrumb_js-packages_filesystem_lib_br-eff824.js:39:32)
    at CommandRegistry.getVisibleHandler (http://localhost:3000/bundle.js:139171:55)
    at CommandRegistry.isVisible (http://localhost:3000/bundle.js:139133:28)
    at MenuCommandExecutorImpl.delegate (http://localhost:3000/bundle.js:142270:43)
    at MenuCommandExecutorImpl.isVisible (http://localhost:3000/bundle.js:142258:21)
    at SampleMenuCommandRegistry.registerCommand (http://localhost:3000/packages_core_lib_browser_menu_browser-menu-plugin_js.js:435:41)

```

The reason for this is most likely this change: https://github.com/eclipse-theia/theia/commit/83b726eaf4ceb14e6d2f64ddbf63665a11baf552#diff-af21beb4f671b6cf6cd72962b340942cd675026547f91ccc93c4e96b732fb2c5R28

The problem-selection.ts does not check anymore that a marker property exists. The problem-marker.ts needs to have an `isObject` check.

### Steps to Reproduce:

1. Create a file that produces an error, eg a javascript file with: 
```
const a = () {
    let b = 'x';
};
```
2. Open the problems view
3. right click on any entry
4. By default there should be 3 entries: `Copy`, `Copy Message` and `Collapse All`

### Additional Information

- Operating System: Linux
- Theia Version: 1.34 tested on commit `f7ef6b56471926d30809e56de946fe61dddefb64`

eneufeld:
#### What it does

Use isObject check in problem-marker class instead of `'kind' in node` check to not get type errors which occur on non marker. These occur because a previously existing check was removed in problem-selection.ts . The error prevents the context menu from showing all entries.

Fixes #12250

#### How to test

1. Open problems view
2. Create a problem marker eg by creating an invalid js file
3. right click on the problem in the problems view
4. See that without the fix there is only one entry: `Collapse All` , with the fix 3 entries are shown: `Copy`, `Copy Message` and `Collapse All`

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/markers/src/common/problem-marker.ts
+++ b/packages/markers/src/common/problem-marker.ts
@@ -16,6 +16,7 @@
 
 import { Marker } from './marker';
 import { Diagnostic } from '@theia/core/shared/vscode-languageserver-protocol';
+import { isObject } from '@theia/core/lib/common';
 
 export const PROBLEM_KIND = 'problem';
 
@@ -24,7 +25,7 @@ export interface ProblemMarker extends Marker<Diagnostic> {
 }
 
 export namespace ProblemMarker {
-    export function is(node: Marker<object>): node is ProblemMarker {
-        return 'kind' in node && node.kind === PROBLEM_KIND;
+    export function is(node: unknown): node is ProblemMarker {
+        return isObject<Marker<object>>(node) && node.kind === PROBLEM_KIND;
     }
 }


packages/markers/src/common/problem-marker.ts
```diff
@@ -25,6 +26,6 @@ export interface ProblemMarker extends Marker<Diagnostic> {
 
 export namespace ProblemMarker {
     export function is(node: Marker<object>): node is ProblemMarker {
-        return 'kind' in node && node.kind === PROBLEM_KIND;
+        return isObject<ProblemMarker>(node) && node.kind === PROBLEM_KIND;
     }
```
msujew, line 30: The fact that we have to use `isObject` here to determine that the `node` is an object implies that the `Marker<object>` contract doesn't hold up as expected. We should use `node: unknown` instead.

packages/markers/src/common/problem-marker.ts
```diff
@@ -24,7 +25,7 @@ export interface ProblemMarker extends Marker<Diagnostic> {
 }
 
 export namespace ProblemMarker {
-    export function is(node: Marker<object>): node is ProblemMarker {
-        return 'kind' in node && node.kind === PROBLEM_KIND;
+    export function is(node: unknown): node is ProblemMarker {
+        return isObject<Marker<object>>(node) && node.kind === PROBLEM_KIND;
```
tsmaeder, line 29: I'm not sure this is correct: shouldn't we assert it's a marker first? Right now an object like this wold fit the bill: 
```
{
   kind: 'problem'
}
```

would fit the bill, even though it's not a marker. But asserting `ProblemMarker.is(foo)` implies that it's a marker, typing-wise.

eneufeld: @tsmaeder I agree, the code is very optimistic. But the previous code did the same assumption.
So I would suggest to tackle this improved marker assertion in a separate ticket.

msujew: Well, during the runtime, yes. But from a statical perspective, it assumed that the `node` already is a `Marker<object>`. So we should confirm that as well.

tsmaeder: I think the static typing problem stems from https://github.com/eclipse-theia/theia/issues/12253. That is a considerably bigger fish to fry, though, and it would be asking to much from a PR to fix a concrete problem. @msujew what are you suggesting, concretely?

eneufeld:
Closing as a better fix will be provided by https://github.com/eclipse-theia/theia/pull/12259

Issue #12253, opened 2023-03-02
`isObject()` asserts untrue types


tsmaeder:
### Bug Description:
in core/common/types.ts, we have the following code: 

```
export function isObject<T = Record<string | number | symbol, unknown>>(value: unknown): value is T {
    // eslint-disable-next-line no-null/no-null
    return typeof value === 'object' && value !== null;
}

```
Let's say we have an interface like so:

```typescript
interface MyInterface {
   title: string;
}
```

Now if we do this:
```
function foo(someObject: unknown) {
   if (isObject<MyInterface>.is(someObject) {
      console.log(someObject.title.toString());
   }
}

foo({});
```

Note that we don't get a compile problem, since inside the if, `someObject` is typed as a `MyInterface`, but we get a runtime error, since the VM frowns upon calling `toString()` on `undefined`.

tsmaeder:
@msujew since you brought in this change in https://github.com/eclipse-theia/theia/pull/11831, I'd be interested in your thinking.

tsmaeder:
This issue is following up on https://github.com/eclipse-theia/theia/issues/12250

msujew:
It's imo obvious that the function cannot assert that the input type is actually of type used for the generic parameter. The parameter is there to keep the code more concise and help devs with code completion compared to just `any`.

tsmaeder:
Hmh...that's sacrificing type safety for convenience. Not something I'm in favor of, in general. Could we assert `Partial<T>` or something like that? That way, we might get the help from the tooling without asserting non-thruths.

paul-marechal:
`Partial<T>` is not really better as it's still possible to get in a situation such as:

```ts
interface MyType {
    foo: string
}

const unknownValue: unknown = { foo: 2 };

if (isObject<MyType>(unknownValue)) {
    // We made it to this branch, and `unknownValue` now looks like:
    // `{ foo?: string }` but `unknownValue.foo` is currently a number.
}
```

PR #12259, opened 2023-03-04
core: refine typing of `isObject<T>`

paul-marechal:
The template parameter for `isObject` is currently used to force cast the value passed to the function, although there is no guarantee nor check done to ensure that this is really the case.

This commit updates the typings so that the type passed to `isObject` is used to guide TypeScript during subsequent field type checking, assuming each field value is `unknown` and must also be type checked.

Example:

```ts
interface MyType {
    foo: string
}

const unknownValue: unknown = { foo: 2 };

if (isObject<MyType>(unknownValue)) {
    // We enter this branch because `unknownValue` is indeed an object,
    // and TypeScript now knows we might want to access the `foo` field
    // from our interface. But we can't yet assume to know the type of
    // `foo`, so it is still typed as `unknown` and we must check:
    if (isString(unknownValue.foo)) {
        // We won't get in this branch. But if we did, then `foo` is
        // now properly typed (and checked) as `string`.
    }
}
```

Fix https://github.com/eclipse-theia/theia/issues/12253
Fix https://github.com/eclipse-theia/theia/issues/12250

Close https://github.com/eclipse-theia/theia/pull/12251

#### How to test

Should compile.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/types.ts
+++ b/packages/core/src/common/types.ts
@@ -17,7 +17,7 @@
 export { ArrayUtils } from './array-utils';
 export { Prioritizeable } from './prioritizeable';
 
-type UnknownObject = Record<string | number | symbol, unknown>;
+type UnknownObject<T extends object> = Record<string | number | symbol, unknown> & { [K in keyof T]: unknown };
 
 export type Deferred<T> = { [P in keyof T]: Promise<T[P]> };
 export type MaybeArray<T> = T | T[];
@@ -56,7 +56,7 @@ export function isFunction<T extends (...args: unknown[]) => unknown>(value: unk
     return typeof value === 'function';
 }
 
-export function isObject<T extends object = UnknownObject>(value: unknown): value is UnknownObject & { [K in keyof T]: unknown } {
+export function isObject<T extends object>(value: unknown): value is UnknownObject<T> {
     // eslint-disable-next-line no-null/no-null
     return typeof value === 'object' && value !== null;
 }
--- a/packages/core/src/common/types.ts
+++ b/packages/core/src/common/types.ts
@@ -17,6 +17,8 @@
 export { ArrayUtils } from './array-utils';
 export { Prioritizeable } from './prioritizeable';
 
+type UnknownObject = Record<string | number | symbol, unknown>;
+
 export type Deferred<T> = { [P in keyof T]: Promise<T[P]> };
 export type MaybeArray<T> = T | T[];
 export type MaybeNull<T> = { [P in keyof T]: T[P] | null };
@@ -54,7 +56,7 @@ export function isFunction<T extends (...args: unknown[]) => unknown>(value: unk
     return typeof value === 'function';
 }
 
-export function isObject<T = Record<string | number | symbol, unknown>>(value: unknown): value is T {
+export function isObject<T extends object = UnknownObject>(value: unknown): value is UnknownObject & { [K in keyof T]: unknown } {
     // eslint-disable-next-line no-null/no-null
     return typeof value === 'object' && value !== null;
 }
--- a/packages/markers/src/common/marker.ts
+++ b/packages/markers/src/common/marker.ts
@@ -14,6 +14,8 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
 // *****************************************************************************
 
+import { isObject, isString } from '@theia/core/lib/common/types';
+
 /*
 * A marker represents meta information for a given uri
 */
@@ -36,4 +38,16 @@ export interface Marker<T> {
      * marker kind specific data
      */
     data: T;
+}
+export namespace Marker {
+    export function is(value: unknown): value is Marker<object>;
+    export function is<T>(value: unknown, subTypeCheck: (value: unknown) => value is T): value is Marker<T>;
+    export function is(value: unknown, subTypeCheck?: (value: unknown) => boolean): boolean {
+        subTypeCheck ??= isObject;
+        return isObject<Marker<object>>(value)
+            && !Array.isArray(value)
+            && subTypeCheck(value.data)
+            && isString(value.uri)
+            && isString(value.owner);
+    }
 }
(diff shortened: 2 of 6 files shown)


eneufeld:
@paul-marechal ,
I think this makes https://github.com/eclipse-theia/theia/pull/12251 obsolete, do you agree?

packages/core/src/common/types.ts
```diff
@@ -17,6 +17,8 @@
 export { ArrayUtils } from './array-utils';
 export { Prioritizeable } from './prioritizeable';
 
+type UnknownObject = Record<string | number | symbol, unknown>;
```
msujew, line 20: Minor: Improvements to the typing of `UnknownObject` and `isObject`. I feel like this would be a more intuitive way of understanding the return type of `isObject`.

```suggestion
type UnknownObject<T extends object> = Record<string | number | symbol, unknown> & { [K in keyof T]: unknown };
```

packages/core/src/common/types.ts
```diff
@@ -54,7 +56,7 @@ export function isFunction<T extends (...args: unknown[]) => unknown>(value: unk
     return typeof value === 'function';
 }
 
-export function isObject<T = Record<string | number | symbol, unknown>>(value: unknown): value is T {
+export function isObject<T extends object = UnknownObject>(value: unknown): value is UnknownObject & { [K in keyof T]: unknown } {
```
msujew, line 59:
```suggestion
export function isObject<T extends object = {}>(value: unknown): value is UnknownObject<T> {
```

packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -546,6 +546,8 @@ export class Range {
             return true;
         }
         return isObject<theia.Range>(arg)
+            && isObject(arg.start)
```
tsmaeder, line 549: Why not fix `Position.is()` to accept `unknown` instead?

tsmaeder (commented):
I've checked that we're finding the bug @eneufeld detected with this change.

packages/plugin-ext/src/plugin/types-impl.ts
paul-marechal: Because I didn't look at `Position.is` and only fixed the new failure here. Do you want to change `Position.is`?

paul-marechal: If you meant to update `Position.isPosition` then this looks like copied code and I'm not changing copied code.

packages/core/src/common/types.ts
paul-marechal: It does look better, thanks for the suggestion.

paul-marechal:
> I think this makes #12251 obsolete, do you agree?

Yes and no. While this PR might fix the issue you found, I still think it would be for the best to update the `ProblemMarker.is` function such as:

```ts
export namespace ProblemMarker {
    export function is(node: unknown): node is ProblemMarker {
        return Marker.is(node) && node.kind === PROBLEM_KIND;
    }
}
```

Just for good measure.

packages/markers/src/common/marker.ts
```diff
@@ -37,3 +39,15 @@ export interface Marker<T> {
      */
     data: T;
 }
+export namespace Marker {
+    export function is(value: unknown): value is Marker<object>;
+    export function is<T>(value: unknown, subTypeCheck: (value: unknown) => value is T): value is Marker<T>;
```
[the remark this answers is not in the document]
tsmaeder: I'm more worried about the function parameter, which is unusual.

packages/plugin-ext/src/plugin/types-impl.ts
tsmaeder: Yes, I mean `isPosition`. It fixes the problem and it's the better fix. What do you mean by "copied code"?

packages/markers/src/common/marker.ts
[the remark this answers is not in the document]
paul-marechal: What worries you?

packages/plugin-ext/src/plugin/types-impl.ts
paul-marechal: It's code copied from VS Code: https://github.com/microsoft/vscode/blob/60f2446458c6806be54eb1d1c877bc42b38a1cd6/src/vs/workbench/api/node/extHostTypes.ts#L68-L80

paul-marechal: Updated `Position.isPosition` because I don't understand why they changed the param type when copying.

packages/markers/src/common/marker.ts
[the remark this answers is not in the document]
tsmaeder: Why add an overload that is not used? It makes the code more complex and currently adds nothing.

msujew (approved):
I'm happy with the changes 👍

packages/markers/src/common/marker.ts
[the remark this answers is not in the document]
paul-marechal: Why: it is the proper way to test this generic type. The `Marker<object>` is just a more common occurrence, and the signature line 43 is here for convenience. If anything, that same signature would be the one that's extra.

Otherwise we have a type `Marker<T>` that could be `Marker<Potato>`, wouldn't it be weird if the only `Marker.is` function only checks for `Marker<object>`?

It would make more sense for `Marker.is` to only check for `Marker<object>` if `T` was constrained to `object` too, but that's currently not the case.

Tell me if you want it removed, otherwise I'll merge this PR.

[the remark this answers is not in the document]
tsmaeder: I want to convince you we should not add stuff that is not needed. Since it's unused, it's not needed. Nobody is passing in the `subTypeCheck` parameter, as far as I can tell.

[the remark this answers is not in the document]
paul-marechal: I introduced a new change, tell me if it's ok with you.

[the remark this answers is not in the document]
paul-marechal: > I want to convince you we should not add stuff that is not needed.

I don't think I can be convinced about taking shortcuts. The implementation with the `subTypeCheck` function is the correct implementation to type guard this type. Anything else is a specialization and might be optimized subsequently without getting rid of the original proper implementation.

PR #12265, opened 2023-03-07
core: fallback to `applicationName`

vince-fugnitto:
#### What it does

The commit adds handling to fallback to the `applicationName` when setting the window title which would previously display either `index.html` in electron  or `localhost:3000` in the browser.

_Before_:

Screen Shot 2023-03-07 at 3 40 20 PM

_After_:

Screen Shot 2023-03-07 at 3 38 43 PM

#### How to test

1. for both the **browser** and **electron** example applications start the app without a workspace
2. confirm that with views open the tab or window title relects the current widget (ex: getting started, preferences, keyboard shortcuts)
3. close all open widgets
4. the title should fallback to the `applicationName`
5. confirm the behavior is correct with a workspace

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/window/window-title-service.ts
+++ b/packages/core/src/browser/window/window-title-service.ts
@@ -100,7 +100,7 @@ export class WindowTitleService {
         if (developmentHost) {
             this._title = developmentHost + this.separator + this._title;
         }
-        document.title = this._title;
+        document.title = this._title || FrontendApplicationConfigProvider.get().applicationName;
         this.onDidChangeTitleEmitter.fire(this._title);
     }
 


packages/core/src/browser/window/window-title-service.ts
```diff
@@ -100,7 +100,7 @@ export class WindowTitleService {
         if (developmentHost) {
             this._title = developmentHost + this.separator + this._title;
         }
-        document.title = this._title;
+        document.title = !!this._title ? this._title : FrontendApplicationConfigProvider.get().applicationName;
```
msujew, line 103: I believe this is effectively the same as:
```suggestion
        document.title = this._title || FrontendApplicationConfigProvider.get().applicationName;
```

msujew (approved):
Looks good to me 👍

PR #12268, opened 2023-03-08
Replace node-keytar with @napi-rs/keyring

Brooooooklyn:
#### What it does

Replace the https://github.com/atom/node-keytar with https://github.com/Brooooooklyn/keyring-node since the `node-keytar` is archived.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/doc/Developing.md
+++ b/doc/Developing.md
@@ -70,12 +70,6 @@ Some additional tools and libraries are needed depending on your platform:
     - Debian-based: `sudo apt-get install libx11-dev libxkbfile-dev`
     - Red Hat-based: `sudo yum install libX11-devel.x86_64 libxkbfile-devel.x86_64 # or .i686`
     - FreeBSD: `sudo pkg install libX11`
-  <a name="prerequisite_keytar"></a>
-  - [`keytar`](#prerequisite_keytar) native node module dependencies ([reference](https://github.com/atom/node-keytar#on-linux)):
-    - Debian/Ubuntu: `sudo apt-get install libsecret-1-dev`
-    - Red Hat-based: `sudo yum install libsecret-devel`
-    - Arch Linux: `sudo pacman -S libsecret`
-    - Alpine: `apk add libsecret-dev`
 
 - Linux/MacOS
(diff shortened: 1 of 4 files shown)


vince-fugnitto (commented):
@Brooooooklyn thank you for the contribution :+1: Unfortunately if we are to replace `keytar` with a new library we should probably choose one which is widely adopted, and well tested with multiple contributors. It seems that `@napi-rs/keyring` might be your personal project?

Brooooooklyn:
@vince-fugnitto I understand your concerns. The npm package `@napi-rs/keyring` is indeed a personal project that I just created. However, its dependencies, https://github.com/hwchen/keyring-rs and https://github.com/hwchen/secret-service-rs, have been widely used in various projects in the Rust ecosystem. My `@napi-rs/keyring` package is only a simple and transparent binding layer that allows Node.js to call the functions of these libraries in the Rust ecosystem, nothing more.

However, `@napi-rs/keyring` also has its advantages: 
1. It does not depend on toolchains like node-gyp and node-gype-prebuild, and can be fully installed with npm/yarn. 
2. It provides many precompiled binaries for different platforms, so downstream users can directly download and install it on any platform. 
3. It has very good binary compatibility on Linux, and can run normally on widely-used but old systems like `centos:7`. 
4. Its upstream dependencies are still actively maintained and developed, while `node-keytar` has already been archived and many existing bugs cannot be followed up and resolved in the future.

t1m0thyj:
> @Brooooooklyn I found a little bit more info, `keytar` is not dead, instead due to atom sunsetting it will move on to a new home [microsoft/vscode#169503 (comment)](https://github.com/microsoft/vscode/issues/169503#issuecomment-1358172058).

There seems to already be a new home for it here: https://github.com/microsoft/keytar
Currently the repo is just a template though - not sure when the source code will be uploaded 🙂

vince-fugnitto:
I believe we can close the pull-request for the time being to make the list of opened pull-requests more manageable.
We can revisit the changes in the future if needed. Thank you for the contribution anyways!

PR #12286, opened 2023-03-10
[CI] Increase memory for failing Pre-docs-Publish workflow step

marcdumais-work:
#### What it does

This PR increases the node memory available for CI workflow step "Pre-docs-Publish". Recently that step, run when a PR is merged, has been failing for apparently running out of memory:

"FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory"
"FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory"

https://github.com/eclipse-theia/theia/actions/runs/4387152336/jobs/7684278705#step:6:479
https://github.com/eclipse-theia/theia/actions/runs/4373388883/jobs/7651868481#step:6:319
https://github.com/eclipse-theia/theia/actions/runs/4365448185/jobs/7635150562#step:6:491

We have in the past increased the limit for that Workflow step, but as the code base grows, it seems to have caught-up with the value we set and probably requires more.

#### How to test

Unfortunately, since the affected workflow step only runs after a PR is merged, I think we will need to merge this PR to validate that it resolves the problem. It seems likely that it will do so. 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -150,7 +150,7 @@ jobs:
         run: |
           yarn docs
         env:
-          NODE_OPTIONS: --max_old_space_size=8196
+          NODE_OPTIONS: --max_old_space_size=9216
 
       - name: Publish GH Pages
         uses: peaceiris/actions-gh-pages@v3


vince-fugnitto (approved):
LGTM, hopefully it fixes the problem :+1:

marcdumais-work:
Ok, let's give it a try.

PR #12296, opened 2023-03-13
[CI] [browser tests] Fix find-replace test suite

marcdumais-work:
#### What it does

I have noticed that the browser test suite has had failures recently. A relatively common one \[1\]  is related to the `find-replace` test suite, and has the following symptom:
`Uncaught Error: Uncaught Error: There is no document for file:///home/runner/work/theia/theia/examples/browser/webpack.config.js`

In a previous PR where I fixed several problems with the browser tests, I could not get to the bottom of this issue and attempted to mitigate by adding pauses. It helped some but did not fix the issue completely.

Looking at it again, I noticed that opening the `.js` file in an editor would trigger the Typescript Language Server, which would then start initializing and loading the whole project. The test script would not wait for it to be done before closing the editor and explorer view container. Some async processing, started by the LS would be interrupted.  The LS initialization would then happen from scratch, when the editor was re-opened. Doing this several times, something would sometimes go wrong, causing an uncaught exception and failing one of the tests. The solution: When opening an editor, wait until the LS has finished its initialization before proceeding. This causes a small delay the first time, but is instantaneous the following times.

Also, added a new browser test file: explorer-open-close.spec.js. I suspected that there was some flakiness in the find-replace tests, potentially caused by the mix of opening and closing both the explorer and an editor. I was not able to pinpoint anything in that regard, but it's possible that something will be caught eventually by the new tests.

\[1\]:
https://github.com/eclipse-theia/theia/actions/runs/4376851067/jobs/7659545233#step:9:1449
https://github.com/eclipse-theia/theia/actions/runs/4379086248/jobs/7664624127#step:9:1466
https://github.com/eclipse-theia/theia/actions/runs/4385465720/jobs/7678225181#step:9:1448

#### How to test

This is bit tricky to test - the original problem is intermittent, having happened in CI a few times in the last several weeks. But at least make sure that CI passes. If you want to compare vs before this PR, when running the version on `master`, disable the calls to `pause()`, so that the issue will be easier to reproduce (this new version no longer has `pause()`). 

During development, I frequently restarted the CI step that runs the browser tests, to create opportunities for failures to happen. Doing so, I have gained increased confidence that the failures above look to be gone. 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/api-tests/src/explorer-open-close.spec.js
+++ b/examples/api-tests/src/explorer-open-close.spec.js
@@ -1,5 +1,5 @@
 // *****************************************************************************
-// Copyright (C) 2020 TypeFox and others.
+// Copyright (C) 2023 Ericsson and others.
 //
 // This program and the accompanying materials are made available under the
 // terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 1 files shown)


vince-fugnitto (approved):
LGTM :+1: 
I confirm that the tests pass successfully during CI, and when I run them locally.

PR #12309, opened 2023-03-15
Improved selection  menu multi-language support

fanyipin:
#### What it does
Improved selection  menu multi-language support
#### How to test
open the selection menu
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-menu.ts
+++ b/packages/monaco/src/browser/monaco-menu.ts
@@ -131,7 +131,8 @@ export class MonacoEditorMenuContribution implements MenuContribution {
 
     protected buildMenuAction(commandId: string, item: IMenuItem): MenuAction {
         const title = typeof item.command.title === 'string' ? item.command.title : item.command.title.value;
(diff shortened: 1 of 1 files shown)


msujew:
Hey @fanyipin, thank you for your first time contributing!

Can you explain a bit better what the issue is about? Maybe even create a corresponding issue first, that you can link in here. The selection menu seems to be localized correctly on `master` (see below), so I'm wondering what you're trying to solve.

image

fanyipin:
Oh, this is really strange, I integrated theia into the ide and then used the Chinese language package, only to find that the submenus below the selection menu are all in English.
image
I traced the code and found that the submenu registration was not handled using the nls multilingual module, resulting in an abnormal display
image
I have tried to remove the custom extension and the result is the same.

My version of theia is 1.35

fanyipin:
I can see why, in order for the ide to start up faster, I adjusted the loading time of the plugin in my extension, which caused the  problem。hope to become one of theia's contributors in the future

PR #12312, opened 2023-03-15
keybindings: search improvement

### LINKED ISSUE #11873: Keyboard Shortcut Widget Search returns too many results
### Steps to Reproduce:

1. Open the keybinding widget
2. Search for a key combination, e.g. Ctrl+U
3. Observe that all bindings that include Ctrl are shown, not just those that have U.

image

### Additional Information

- Operating System:
- Theia Version: 1.32

FernandoAscencio:
#### What it does
Closes #11873

This commit aims to improve the keybinding search results in the Keyboard Shortcut Widget.

#### How to test
Similar to the linked issue.
1. Open the keybinding widget
2. Search for a key combination, e.g. Ctrl+U
3. Check that the proper keybindings are shown.
##### Notes
Ctrl+U should show keybindings with `Ctrl` and `U`, with the following being examples that may or may not exist already.
- `Ctrl`+`U`
- `Ctrl`+`Alt`+`U`
- `Ctrl`+`T` `U`
- `Ctrl`+`K` `Ctrl`+`U`.

Other search terms that should work for finding `Ctrl`+`U` are 'Ctrl U', 'U', 'Ctrl' and 'Ctrl+'. These should include `Ctrl`+`U` in the results.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto:
@FernandoAscencio do you mind rebasing the pull-request against the latest master?

FernandoAscencio:
@vince-fugnitto 
Rebase is done.

FernandoAscencio:
@colin-grant-work 
Does this seem good to you?

packages/keymaps/src/browser/keybindings-widget.tsx
```diff
@@ -211,7 +211,7 @@ export class KeybindingWidget extends ReactWidget implements StatefulWidget {
 
     protected formatAndMatchKeybinding(item: KeybindingItem, queryItems: string[]): boolean {
         if (item.keybinding) {
-            let match = false;
+            const searchTerms = [...queryItems].filter(Boolean);
```
colin-grant-work, line 214: Since `filter` always creates a new array, there's no need to spread here.

```suggestion
            const searchTerms = queryItems.filter(Boolean);
```

packages/keymaps/src/browser/keybindings-widget.tsx
```diff
@@ -224,14 +224,14 @@ export class KeybindingWidget extends ReactWidget implements StatefulWidget {
                     if (chunkIndex !== 0) {
                         collection.push({ value: '+', match: false, key: false });
                     }
-                    const chunkMatches = queryItems.includes(matchChunks[chunkIndex].toLocaleLowerCase());
-                    match ||= chunkMatches;
+                    const indexOfTerm = searchTerms.indexOf(matchChunks[chunkIndex].toLocaleLowerCase());
+                    const chunkMatches = indexOfTerm > -1 ? !!searchTerms.splice(indexOfTerm, 1) : false;
```
colin-grant-work, line 228: This line is a bit confusing. I don't believe that `!!anything.slice(any, arguments)` can ever be false, so it seems this is just for the side effect of deletion.

FernandoAscencio: Yes, it can neve be false in this case. Should I split the deletion from setting `chunkMatches`?

colin-grant-work: Yeah, I think that would be a little clearer - two lines for two concepts (match -> true + don't match same thing twice. You could even call `searchTerms` something like `unmatchedSearchTerms` to make it clearer that once something has been matched, it doesn't belong.).

packages/keymaps/src/browser/keybindings-widget.tsx
```diff
@@ -224,14 +224,14 @@ export class KeybindingWidget extends ReactWidget implements StatefulWidget {
...
                     collection.push({ value: chunk, match: chunkMatches, key: true });
                 });
                 return collection;
             }, []);
             item.labels.keybinding = { value: item.labels.keybinding.value, segments };
-            return match;
+            return !searchTerms.length;
```
colin-grant-work: This now only returns true if _all_ chunks were matched (and so all removed from `searchTerms`, leaving none); previously it returned true if _any_ chunk matched. Is that deliberate?

packages/keymaps/src/browser/keybindings-widget.tsx
FernandoAscencio: Thanks!

packages/keymaps/src/browser/keybindings-widget.tsx
FernandoAscencio: Yes
Part of the reason as to why searching for `Ctrl+U` returned more than the expected entries was because if only `Ctrl` matched then it would return that entry. Thus you would end with results for `Ctrl+K` and `Ctrl+T` when searching `Ctrl+U`.

vince-fugnitto (approved):
LGTM :+1: 

- [x] the code looks good 
- [x] confirmed that searching works as expected and is aligned with the behavior present in VS Code

PR #12326, opened 2023-03-20
terminal: add context menu

### LINKED ISSUE #7632: terminal: add context-menu
### Description

We should include a terminal context-menu to include the following items:
- **Copy**
- **Paste**
- **Select All**
- **Clear** (clears the terminal output)
- **Split Terminal**
- **New Terminal**
- **Kill Terminal** (close terminal)

**Additional Info**
- 7d56c34

planger:
#### What it does
Adds a context menu to the terminal widget that covers roughly the same menu items as VS Code.

* New Terminal
* Split Terminal
* Copy
* Paste
* Select All
* Clear
* Kill Terminal

Therefore, this change also creates the two new commands *Select All* and *Kill Terminal*.

Fixes #7632
Change-Id: Icb23f6fbdc44cdc5e018277e44ed5fe46c8ac26a

#### How to test
Open a terminal and try out all context menu items (list see above).
It may be worth making sure that the commands also work when triggered via the command palette.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/terminal/src/browser/terminal-frontend-contribution.ts
+++ b/packages/terminal/src/browser/terminal-frontend-contribution.ts
@@ -31,8 +31,8 @@ import {
 } from '@theia/core/lib/common';
 import {
     ApplicationShell, KeybindingContribution, KeyCode, Key, WidgetManager, PreferenceService,
-    KeybindingRegistry, Widget, LabelProvider, WidgetOpenerOptions, StorageService,
-    QuickInputService, codicon, CommonCommands, FrontendApplicationContribution, OnWillStopAction, Dialog, ConfirmDialog, FrontendApplication, PreferenceScope
+    KeybindingRegistry, LabelProvider, WidgetOpenerOptions, StorageService, QuickInputService,
+    codicon, CommonCommands, FrontendApplicationContribution, OnWillStopAction, Dialog, ConfirmDialog, FrontendApplication, PreferenceScope
 } from '@theia/core/lib/browser';
 import { TabBarToolbarContribution, TabBarToolbarRegistry } from '@theia/core/lib/browser/shell/tab-bar-toolbar';
 import { TERMINAL_WIDGET_FACTORY_ID, TerminalWidgetFactoryOptions, TerminalWidgetImpl } from './terminal-widget-impl';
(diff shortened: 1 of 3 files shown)


packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -70,6 +70,8 @@ export namespace TerminalMenus {
     export const TERMINAL_TASKS_CONFIG = [...TERMINAL_TASKS, '4_terminal'];
     export const TERMINAL_NAVIGATOR_CONTEXT_MENU = ['navigator-context-menu', 'navigation'];
     export const TERMINAL_OPEN_EDITORS_CONTEXT_MENU = ['open-editors-context-menu', 'navigation'];
+
+    export const TERMINAL_CONTEXT_MENU = ['terminal'];
```
vince-fugnitto, line 74: minor: we should add `-context-menu` so it is a little clearer and has less change for collisions:

```suggestion
    export const TERMINAL_CONTEXT_MENU = ['terminal-context-menu'];
```

packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -826,6 +867,11 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
             command: TerminalCommands.TOGGLE_TERMINAL.id,
             keybinding: 'ctrl+`',
         });
+        keybindings.registerKeybinding({
+            command: TerminalCommands.SELECT_ALL.id,
+            keybinding: 'ctrlcmd+a',
+            context: TerminalKeybindingContexts.terminalActive
```
vince-fugnitto, line 873: minor: `context` is deprecated, we should probably make use of `when` clauses instead.

vince-fugnitto (commented):
I confirmed that the functionality works as expected :+1: 
I only have some very minor comments regarding the code.

packages/terminal/src/browser/terminal-frontend-contribution.ts
planger: Right, I didn't use the `when` to maintain consistency, but you're right, better to start at some point. As far as I could tell, the context key `terminalFocus` is equivalent, so I took the liberty to replace all usages of `context: TerminalKeybindingContexts.terminalActive` in this class.

vince-fugnitto (approved):
LGTM :+1: 

I can confirm that the context-menu is correctly rendered on both the browser and electron, and the commands work as expected.

PR #12338, opened 2023-03-22
Guard against identical ID's in plugin tree view

### LINKED ISSUE #12337: New PluginTree ID generation algorithm generates identical ID's
### Bug Description:

More details [here](https://github.com/eclipse-theia/theia/pull/12120/files#r1145151410)

cc: @tsmaeder 

### Steps to Reproduce:

1. Set up a pair of TS files like this:

image

3. Do a reference search for the variable `a`
4. First, note that not all references are found - in this screenshot, only two console logs are shown, because some of them have the same ID, so the tree only shows one.
5. Repeat the reference search.
6. The application will freeze and eventually crash, trapped in a loop in `TopDownTreeIterator` because the nodes with the same ID's end up being their own siblings, so the iteration is infinite.

### Additional Information

- Operating System: RHEL
- Theia Version: 1ca8131154e762c8924beb2f1052542f7c30da10

colin-grant-work:
#### What it does

Fixes #12337 mostly by bringing in VSCode's approach to the same problem.

#### How to test

1. Follow procedure in #12337 
2. All references should be rendered separately, and application should not crash.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/tree/tree-views.ts
+++ b/packages/plugin-ext/src/plugin/tree/tree-views.ts
@@ -23,6 +23,7 @@ import {
 // TODO: extract `@theia/util` for event, disposable, cancellation and common types
 // don't use @theia/core directly from plugin host
 import { Emitter } from '@theia/core/lib/common/event';
+import { basename } from '@theia/core/lib/common/paths';
 import { Disposable, DisposableCollection } from '@theia/core/lib/common/disposable';
 import { DataTransfer, DataTransferItem, Disposable as PluginDisposable, ThemeIcon } from '../types-impl';
 import { Plugin, PLUGIN_RPC_CONTEXT, TreeViewsExt, TreeViewsMain, TreeViewItem, TreeViewRevealOptions, DataTransferFileDTO } from '../../common/plugin-api-rpc';
@@ -194,8 +195,6 @@ class TreeViewExtImpl<T> implements Disposable {
     private static readonly ID_COMPUTED = 'c';
     private static readonly ID_ITEM = 'i';
 
-    private nextItemId: 0;
-
     private readonly onDidExpandElementEmitter = new Emitter<TreeViewExpansionEvent<T>>();
     readonly onDidExpandElement = this.onDidExpandElementEmitter.event;
 
@@ -330,7 +329,7 @@ class TreeViewExtImpl<T> implements Disposable {
             // parent is inconsistent
             return undefined;
         }
-        const candidateId = this.buildTreeItemId(parentId, treeItem);
+        const candidateId = this.buildTreeItemId(parentId, treeItem, false);
         if (this.nodes.has(candidateId)) {
             return chain.concat(candidateId);
         }
@@ -361,18 +360,32 @@ class TreeViewExtImpl<T> implements Disposable {
         return idLabel;
     }
 
-    private buildTreeItemId(parentId: string, item: TreeItem): string {
-        // build tree id according to https://code.visualstudio.com/api/references/vscode-api#TreeItem
-        // note: the front end tree implementation cannot handle reparenting items, hence the id is set to the "path" of individual ids
-        let id = typeof item.id === 'string' ? item.id : this.getItemLabel(item);
-        if (id) {
-            // we use '' as the id of the root, we don't consider that a valid id
-            // since '' is falsy, we'll never get '' in this branch
-            id = TreeViewExtImpl.ID_ITEM + id;
-        } else {
-            id = TreeViewExtImpl.ID_COMPUTED + this.nextItemId++;
+    // Modeled on https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/extHostTreeViews.ts#L822
+    private buildTreeItemId(parentId: string, item: TreeItem, mustReturnNew: boolean): string {
+        if (item.id) {
+            return `${TreeViewExtImpl.ID_ITEM}-@-${parentId}-@-${item.id}`;
         }
-        return `${parentId}/${id}`;
+
+        const treeItemLabel = this.getItemLabel(item);
+        const prefix: string = `${TreeViewExtImpl.ID_COMPUTED}-@-${parentId || ''}-@-`;
+        let elementId = treeItemLabel ? treeItemLabel : item.resourceUri ? basename(item.resourceUri.fsPath) : '';
+        elementId = elementId.indexOf('/') !== -1 ? elementId.replace('/', '//') : elementId;
+        const childrenNodes = (this.nodes.get(parentId)?.children || []);
+
+        let id: string;
+        let counter = 0;
+        do {
+            id = `${prefix}/${counter}:${elementId}`;
+            if (!mustReturnNew || !this.nodes.has(id) || this.nodes.get(id) === item) {
+                // Return first if asked for or
+                // Return if handle does not exist or
+                // Return if handle is being reused
+                break;
+            }
+            counter++;
+        } while (counter <= childrenNodes.length);
+
+        return id;
     }
 
     async getChildren(parentId: string): Promise<TreeViewItem[] | undefined> {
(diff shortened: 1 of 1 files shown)


tsmaeder:
@colin-grant-work I'm a bit confuse, because the API documentation of [TreeItem](https://code.visualstudio.com/api/references/vscode-api#TreeItem) clearly states: 

> Optional id for the tree item that _has to be unique across tree_. The id is used to preserve the selection and expansion state of the tree item.

Is the "references view" built in extension not compliant with the published API?

packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -346,18 +345,32 @@ class TreeViewExtImpl<T> implements Disposable {
...
-        let id = typeof item.id === 'string' ? item.id : this.getItemLabel(item);
-        if (id) {
-            // we use '' as the id of the root, we don't consider that a valid id
-            // since '' is falsy, we'll never get '' in this branch
-            id = TreeViewExtImpl.ID_ITEM + id;
-        } else {
-            id = TreeViewExtImpl.ID_COMPUTED + this.nextItemId++;
+    // Modeled on https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/extHostTreeViews.ts#L822
+    private buildTreeItemId(parentId: string, item: TreeItem, mustReturnNew: boolean): string {
+        if (item.id) {
```
tsmaeder, line 350: We still need the logic to construct paths from ID's for all the reasons in https://github.com/eclipse-theia/theia/pull/12120/files. While the incremental tree update is running, a node might be in more than one place at the same time. The drag/drop example helps from the PR helps test this.

packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -346,18 +345,32 @@ class TreeViewExtImpl<T> implements Disposable {
...
-        return `${parentId}/${id}`;
+
+        const treeItemLabel = this.getItemLabel(item);
+        const prefix: string = parentId ? parentId : TreeViewExtImpl.ID_COMPUTED;
+        let elementId = treeItemLabel ? treeItemLabel : item.resourceUri ? basename(item.resourceUri.fsPath) : '';
+        elementId = elementId.indexOf('/') !== -1 ? elementId.replace('/', '//') : elementId;
+        const childrenNodes = (this.nodes.get(parentId)?.children || []);
+
+        let id: string;
+        let counter = 0;
```
tsmaeder, line 361: This will generate the same ids potentially different domain elements, because it's based on the position of an element in the parent. I think this is wrong: we're assuming sameness, when in fact, we don't know. A simple fix would be to have a global counter instead of starting at 0 here.
The API doc states

> If not provided, an id is generated using the tree item's label. Note that when labels change, ids will change and that selection and expansion state cannot be kept stable anymore.

Note the absence of "position" as a as a part of the computed ID.

packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -346,18 +345,32 @@ class TreeViewExtImpl<T> implements Disposable {
...
-        // note: the front end tree implementation cannot handle reparenting items, hence the id is set to the "path" of individual ids
-        let id = typeof item.id === 'string' ? item.id : this.getItemLabel(item);
-        if (id) {
-            // we use '' as the id of the root, we don't consider that a valid id
-            // since '' is falsy, we'll never get '' in this branch
-            id = TreeViewExtImpl.ID_ITEM + id;
-        } else {
-            id = TreeViewExtImpl.ID_COMPUTED + this.nextItemId++;
+    // Modeled on https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/extHostTreeViews.ts#L822
+    private buildTreeItemId(parentId: string, item: TreeItem, mustReturnNew: boolean): string {
```
tsmaeder, line 349: I'm not sure what `mustReturnNew` means and why it's true in one case, and false in the other.

packages/plugin-ext/src/plugin/tree/tree-views.ts
tsmaeder: I think the _really_ correct thing to do is to uniqify any two computed id's that end up the same and otherwise treat them as a true id.

colin-grant-work: If `mustReturnNew` is true, then it won't generate the same id's for different items. The current children can't be using all of the counter values up to the current length of the children array; either there's a gap in the middle, or the value at `.length` itself must be open.

packages/plugin-ext/src/plugin/tree/tree-views.ts
colin-grant-work: This logic does use the parent ID as part of the ID if present:

```
const prefix: string = parentId ? parentId : TreeViewExtImpl.ID_COMPUTED;
```

packages/plugin-ext/src/plugin/tree/tree-views.ts
tsmaeder: But what about the next time we fetch the children? We will generate the same id for a potentially different element, just because it's in the same place?

packages/plugin-ext/src/plugin/tree/tree-views.ts
tsmaeder: But why would it ever makes sense to do `mustReturnNew = false`? for multiple elements that have the same computed "id", it will just return the first one, no? Would that be correct?

packages/plugin-ext/src/plugin/tree/tree-views.ts
colin-grant-work: This code does seem to work with the example extension from your drag and drop PR. Do you have a concrete example where this would fail?

packages/plugin-ext/src/plugin/tree/tree-views.ts
colin-grant-work: Yes, I think that's a bit odd as well, since it seems non-deterministic, but the same is true in VSCode - in the parallel calls, the corresponding flag is true in one place and false in the other. Given the ID's are being computed, I suppose the idea is that really can't guarantee that we'll find the same one, so we'll take the first one that's as close as we can reasonably get, given the data available.

packages/plugin-ext/src/plugin/tree/tree-views.ts
colin-grant-work: Yes, that is possible. As I noted below, the promise of certainty with computed ID's is lower than with plugin-supplied ID's. There's no way around that. Since we depend on something that is not guaranteed to be unique, and indeed is demonstrably not unique in a number of cases - it is impossible to guarantee that we will always match the same item. But matching an item that is approximately the same may be better than matching no item at all. That seems to be the design choice that VSCode has made.

packages/plugin-ext/src/plugin/tree/tree-views.ts
tsmaeder: I was able to reproduce the problem from https://github.com/eclipse-theia/theia/issues/12111 eventually. The simple scenario from the bug does not reproduce the problem reliably, but randomly dragging elements around for a while eventually triggered the stack overflow. The problem seems to be timing-dependent.

tsmaeder: I'm thinking this might be a problem because we're using the same separator for the path separator and the prefix separator. Let's think about an element that has the id `/i/parent/child`. The tree item id will be `i/i/parent/child`. Wouldn't that be the same id we get when we compute the id of an element with id `child`, whose parent is an element with id `parent`? Because the computed id of the parent is `i/parent`, so the expression becomes `${'i'}/${'i/parent'}/'child'` = `i/i/parent/child`. 

Come to think of it, we might have to escape the path separator in given id's in order to prevent this "attack". Sorry to make seemingly trivial change so complicated 🤦

packages/plugin-ext/src/plugin/tree/tree-views.ts
tsmaeder: That whole "computed id" business is a disaster in the VS Code API, if you ask me, but I buy your argument.

tsmaeder:
The current version fixes the problem with D&D reparenting.

tsmaeder (approved):
Fine to merge from my side: I still think we could construct an example where unescaped weird id values could wreak havoc, but that situation is no worse than before, so not an obstacle to merging the PR.

PR #12347, opened 2023-03-23
Check renderer schema when updating preference renderers

### LINKED ISSUE #12345: Preference node rendered incorrectly after schema change
### Bug Description:

After https://github.com/eclipse-theia/theia/pull/11678, the `PreferenceEditorWidget` maintains a record of the schema it believes each node is being rendered with.

https://github.com/eclipse-theia/theia/blob/656cd9bbf4f267e15c68e2bf0cc09c76e04c4987/packages/preferences/src/browser/views/preference-editor-widget.ts#L172-L187

However, the rendering of the node depends not on that schema, but on data stored only on the node:

https://github.com/eclipse-theia/theia/blob/656cd9bbf4f267e15c68e2bf0cc09c76e04c4987/packages/preferences/src/browser/views/preference-editor-widget.ts#L141

Specifically its `preference` field:

https://github.com/eclipse-theia/theia/blob/656cd9bbf4f267e15c68e2bf0cc09c76e04c4987/packages/preferences/src/browser/util/preference-types.ts#L67

Under certain conditions, those two sources of truth, the map stored in the `PreferenceEditorWidget` and the schema found on the node itself can come into conflict.

### Steps to Reproduce:

1. Implement a contribution that overrides the schema of a particular preference. Ours looks like this because we were working with a plugin that had specific expectations about the `editor.inlayHints.enabled` preference that reflected an older state of Monaco / VSCode.
```typescript
@postConstruct()
    protected init(): void {
        this.schemaProvider.getCombinedSchema().properties['editor.inlayHints.enabled'] = {
            'type': 'boolean',
            'default': true,
            'defaultValue': true,
            'description': 'Enables inlay hints in editors.',
            'overridable': true,
            'restricted': false,
            'scope': PreferenceScope.Folder,
        };
        this.schemaProvider['onDidPreferenceSchemaChangedEmitter'].fire();
    }
```
2. Open the preference view.
3. Observe that the node for the overridden preference is rendered as though its original schema were still applicable.

### Additional Information

- Operating System: RHEL
- Theia Version: 656cd9bbf4f267e15c68e2bf0cc09c76e04c4987

colin-grant-work:
#### What it does

Fixes #12345

#### How to test

1. Follow #11678

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/browser/views/preference-editor-widget.ts
+++ b/packages/preferences/src/browser/views/preference-editor-widget.ts
@@ -17,6 +17,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { postConstruct, injectable, inject } from '@theia/core/shared/inversify';
 import throttle = require('@theia/core/shared/lodash.throttle');
+import * as deepEqual from 'fast-deep-equal';
 import {
     PreferenceService,
     CompositeTreeNode,
@@ -35,7 +36,6 @@ import { Preference } from '../util/preference-types';
 import { COMMONLY_USED_SECTION_PREFIX } from '../util/preference-tree-generator';
 import { PreferencesScopeTabBar } from './preference-scope-tabbar-widget';
 import { PreferenceNodeRendererCreatorRegistry } from './components/preference-node-renderer-creator';
-import stableJsonStringify = require('fast-json-stable-stringify');
 
 export interface PreferencesEditorState {
     firstVisibleChildID: string,
@@ -127,7 +127,7 @@ export class PreferencesEditorWidget extends BaseWidget implements StatefulWidge
     protected handleSchemaChange(isFiltered: boolean): void {
         for (const [id, renderer, collection] of this.allRenderers()) {
             const node = this.model.getNode(renderer.nodeId);
-            if (!node || (Preference.LeafNode.is(node) && this.hasSchemaChanged(node))) {
+            if (!node || (Preference.LeafNode.is(node) && this.hasSchemaChanged(renderer, node))) {
                 renderer.dispose();
                 collection.delete(id);
             }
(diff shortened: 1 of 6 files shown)


packages/preferences/src/browser/views/preference-editor-widget.ts
```diff
@@ -169,21 +166,8 @@ export class PreferencesEditorWidget extends BaseWidget implements StatefulWidge
...
-        const schemaProperty = this.schemaProvider.getSchemaProperty(leafNode.preferenceId);
-        if (schemaProperty) {
-            const key = stableJsonStringify(schemaProperty);
-            this.preferenceDataKeys.set(leafNode.id, key);
-            return key;
-        } else {
-            return undefined;
-        }
+    protected hasSchemaChanged(renderer: GeneralPreferenceNodeRenderer, node: Preference.LeafNode): boolean {
+        return !deepEqual(renderer.schema, node.preference.data);
```
msujew, line 170: I don't think this `deepEqual` call ever returns `false`, since `renderer.schema` and `node.preference.data` always have the same object pointer. That's the reason why I needed to cache the stringified object in the first place.

msujew (commented):
I noticed a regression for installing theme extensions. Due to the issue outlined in the comment below, the preference renderer isn't updated and continues to show the old values.

packages/preferences/src/browser/views/preference-editor-widget.ts
colin-grant-work: Hm... that's a fair point. In the case I was most interested in, I was putting a new object in place, so the old version and the new version weren't strictly equal, but you're right, if only the internals are changed, that wouldn't be the case.

colin-grant-work: On the other hand, that's really in the hands of the updater. By default, the schema provider just assigns the object passed into to `updateSchemaProperty` to the appropriate key, so if a new object is passed in, then there won't be referential equality:

```
updateSchemaProperty(key: string, property: PreferenceDataProperty): void {
        this.updateSchemaProps(key, property);
        this.fireDidPreferenceSchemaChanged();
    }

    protected updateSchemaProps(key: string, property: PreferenceDataProperty): void {
        this.combinedSchema.properties[key] = property;

        switch (property.scope) {
            case PreferenceScope.Folder:
                this.folderSchema.properties[key] = property;
            // Fall through. isValidInScope implies that User ⊃ Workspace ⊃ Folder,
            // so anything we add to folder should be added to workspace, but not vice versa.
            case PreferenceScope.Workspace:
                this.workspaceSchema.properties[key] = property;
                break;
        }
    }
```

colin-grant-work: Pushed a change that adjusts the existing uses of `updateSchemaProperty` to use new objects, and added a comment on the method to indicate that that's the recommended usage. What do you think of that approach?

PR #12353, opened 2023-03-27
Fix memory leak in plugin tree views

### LINKED ISSUE #10404: Memory leak in plugin tree views
### Bug Description:
Theia always fetches whole tree in plugin tree views when change event is fired. (It calls tree refresh without argument and it causes complete refresh).
In tree-views.ts Theia calls clearChildren to dispose previous commands and remove children before new child elements are fetched. 
When full refresh is called it begins with root node (with empty id) and old nodes are not properly disposed/removed. If tree is huge and is constantly updating it consumes several gb of memory and sometimes even crashes. I found out that old commands are not disposed and still kept in command registry (and maybe somewhere else). Same tree works fine in VSCode.

### Steps to Reproduce:

1. Install vscode extension that has huge tree view and is constanlty updating.
2. Open it
3. Open chrome inspector (memory tab).
4. Start tree updates and wait for some time.
5. In chrome inspector memory consumption will increase (from 35mb to 600-900mb in my case)

### Additional Information

- Operating System:
- Theia Version:

pisv:
Fixes #10404.

#### What it does
Fixes a bug in `TreeViewExtImpl<T>.getChildren(parentId: string)`, where newly created top-level children were added to the old root node instead of the new one. This caused a memory leak by retaining an implicit reference to the old root node from the new one and failing to properly track and, hence, dispose the top-level children.

The fix ensures that old nodes get disposed and become subject to garbage collection by reusing the single root node instead of creating a new one on full refresh.

See https://github.com/eclipse-theia/theia/issues/10404#issuecomment-1481293316 for more information.

#### How to test
Verify that plugin tree views still work as expected, and there is no memory leak on full refresh now.

For example, open Theia on the [vscode-extension-samples/helloworld-sample](https://github.com/microsoft/vscode-extension-samples/tree/main/helloworld-sample) workspace, and use the `Node Dependencies` view contributed by the [vscode-extension-samples/tree-view-sample](https://github.com/microsoft/vscode-extension-samples/tree/main/tree-view-sample):

* Take a heap snapshot using Chrome DevTools and verify that there is exactly six `Dependency` objects.
* Refresh the view and take another heap snapshot. Verify that there is still exactly six `Dependency` objects.

#### Review checklist

#### Reminder for reviewers


tsmaeder (approved):
Looks good to me.

tsmaeder:
I was able to verify that the number of nodes doesn't grow upon refresh.

PR #12367, opened 2023-03-31
Editor: restore `highlightModifiedTabs` functionality

### LINKED ISSUE #12185: `workbench.editor.highlightModifiedTabs` has not effect
### Bug Description:

Setting `workbench.editor.highlightModifiedTabs` to `true` does not do anything. The [logic still exists](https://github.com/eclipse-theia/theia/blob/6f536911866fd8de800d9590039bfa0ffc449903/packages/core/src/browser/common-frontend-contribution.ts#L466-L471) in Theia, but the styles have been removed. Breaking change: https://github.com/eclipse-theia/theia/pull/10908/

### Steps to Reproduce:

1. Open Theia (browser or electron), set auto-save to off and `workbench.editor.highlightModifiedTabs` to `true`,
2. Touch a file. You can see the dirty modifier, but the extra top border is missing from the dirty tab. It works from VS Code.

Electron:

https://user-images.githubusercontent.com/1405703/219052938-976dafbd-1388-4995-9e97-75b896c29870.mp4

Browser:

https://user-images.githubusercontent.com/1405703/219053042-2e269ad2-87a8-427b-bdee-41454a778725.mp4

VS Code:

https://user-images.githubusercontent.com/1405703/219053145-de2fd8ee-2d97-4be4-a326-1a0e6b95aaae.mp4

### Additional Information

- Operating System: macOS
- Theia Version: 6f5369118

FernandoAscencio:
#### What it does
Closes #12185

Use of box-shadow parameter causes several issues when several box-shadow parameters apply to a single element.
Specially since one parameter is used to manage two independent visual cues (`borderTop` and `borderBottom`)
There are situations when the top border will have no assigned color and this will cause the entire `box-shadow` parameter to not render. Meaning that a missing color for `borderTop` causes an active `borderBottom` to disappear and vice-versa. To add to the issue, the parameter will overwrite other applicable `box-shadow`, so you can end up with other stylings not showing when they should.

There are two possible solutions:
- First one involves a complicated system that checks if the colors for top and bottom exist and then builds a `box-shadow` parameter out of the results and finally adding it to the style sheet.
- The second is changing `box-shadow` for `border-top` and `border-bottom` decoupling their dependency on one another.
For editor tabs, there is no real difference between using `box-shadow` and `border`. Thus, this commit changes the use of `box-shadow` to `border`.

This commit also fixes a tangential issue with High Contrast themes and how borders did not properly render when `workbench.editor.highlightModifiedTabs` was active.

#### How to test
Similarly to #12185
1. Open Theia (browser or electron), set auto-save to off and workbench.editor.highlightModifiedTabs to true
2. Select a file, modify it and see the top border appear.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/common-styling-participants.ts
+++ b/packages/core/src/browser/common-styling-participants.ts
@@ -194,15 +194,49 @@ export class TabbarStylingParticipant implements StylingParticipant {
             `);
         }
         const tabActiveBackground = theme.getColor('tab.activeBackground');
-        const tabActiveBorderTop = theme.getColor('tab.activeBorderTop') || (highContrast && contrastBorder) || 'transparent';
+        const tabActiveBorderTop = theme.getColor('tab.activeBorderTop');
+        const tabUnfocusedActiveBorderTop = theme.getColor('tab.unfocusedActiveBorderTop');
         const tabActiveBorder = theme.getColor('tab.activeBorder') || (highContrast && contrastBorder) || 'transparent';
+        const tabUnfocusedActiveBorder = theme.getColor('tab.unfocusedActiveBorder') || (highContrast && contrastBorder) || 'transparent';
         collector.addRule(`
             #theia-main-content-panel .p-TabBar .p-TabBar-tab.p-mod-current {
                 color: var(--theia-tab-activeForeground);
-                ${tabActiveBackground && `background: ${tabActiveBackground};`}
-                box-shadow: 0 1px 0 ${tabActiveBorderTop} inset, 0 -1px 0 ${tabActiveBorder} inset;
+                ${tabActiveBackground ? `background: ${tabActiveBackground};` : ''}
+                ${tabActiveBorderTop ? `border-top: 1px solid ${tabActiveBorderTop};` : ''}
+                border-bottom: 1px solid ${tabActiveBorder};
+            }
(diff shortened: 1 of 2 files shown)


msujew (commented):
> For editor tabs, there is no real difference between using box-shadow and border.

Unfortunately there is: While `box-shadow` has absolutely no influence on the layout of DOM elements, `border` does influence the layouting of child elements. This is visible in this PR when quickly switching between modified and unmodified document state. I've visualized the location of the tab label with a red line in the following:

image

image

Note the way that the top border moves the content of the tab bar down by 1 pixel. I think a better box shadow solution is necessary (or more css to offset the border).

FernandoAscencio:
@msujew 
Thanks for the feedback.
I have solved the issue with a bit more CSS. However, it is restricted to one specific border thickness.  Both for the normal top border highlight and the `highlightModifiedTabs` border highlight are restricted to their current thickness with this solution. As the `tab.activeBorder` and `tab.unfocusedActiveBorder` are both used to add the bottom border in High Contrast themes (as box-shadow` originally and as `border` here), I doubt their thickness will change. I am not sure we care much about the thickness change in the top border when we want to highlight modified tabs.

If you are interested I can do another PR with the solution involving `box-shadow` I previously did.

PR #12377, opened 2023-04-04
monaco-quick-input-service: set context key "inQuickOpen"

marcdumais-work:
#### What it does

Context Key ""inQuickOpen"" exists in vscode and its value (true or false) reflects whether the Quick Open, AKA Quick Input, is currently open/visible or not.

This context key can be useful in Theia too, e.g. it can be used in the "when" clause of commands registered by vscode extensions, or set programmatically whenever it's useful to know the state of the Quick Open UI.

#### How to test

I have provided a test in the browser suite. Confirm that CI passes.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-quick-input-service.ts
+++ b/packages/monaco/src/browser/monaco-quick-input-service.ts
@@ -30,6 +30,7 @@ import { MonacoResolvedKeybinding } from './monaco-resolved-keybinding';
 import { IQuickAccessController } from '@theia/monaco-editor-core/esm/vs/platform/quickinput/common/quickAccess';
 import { QuickAccessController } from '@theia/monaco-editor-core/esm/vs/platform/quickinput/browser/quickAccess';
 import { ContextKeyService as VSCodeContextKeyService } from '@theia/monaco-editor-core/esm/vs/platform/contextkey/browser/contextKeyService';
+import { IContextKey } from '@theia/monaco-editor-core/esm/vs/platform/contextkey/common/contextkey';
 import { IListOptions, List } from '@theia/monaco-editor-core/esm/vs/base/browser/ui/list/listWidget';
 import * as monaco from '@theia/monaco-editor-core';
 import { ResolvedKeybinding } from '@theia/monaco-editor-core/esm/vs/base/common/keybindings';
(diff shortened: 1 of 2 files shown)


packages/monaco/src/browser/monaco-quick-input-service.ts
```diff
@@ -100,7 +102,7 @@ export class MonacoQuickInputImplementation implements IQuickInputService {
 
     setContextKey(key: string | undefined): void {
         if (key) {
-            this.contextKeyService.createKey<string>(key, undefined);
+            this.contextKey = this.contextKeyService.createKey<string>(key, undefined);
         }
     }
```
vince-fugnitto, line 107: I believe this method is meant for callers, and it may be problematic for us to set the `contextKey` variable especially since we later rely on it for the `inQuickOpen`.

packages/monaco/src/browser/monaco-quick-input-service.ts
```diff
@@ -177,6 +179,9 @@ export class MonacoQuickInputImplementation implements IQuickInputService {
 
     private initController(): void {
         this.controller = new QuickInputController(this.getOptions());
+        this.setContextKey('inQuickOpen');
+        this.onShow(() => { this.contextKey.set(true); });
+        this.onHide(() => { this.contextKey.set(false); });
         this.updateLayout();
     }
```
vince-fugnitto, line 186: In addition to the previous comment, I think we can simplify and use the following which is present in other areas of the framework, where we register listeners in the `postConstruct`:

```diff
+ protected inQuickOpen: ContextKey<boolean>;

 @postConstruct()
 protected init(): void {
     this.initContainer();
     this.initController();
 
+    this.inQuickOpen = this.contextKeyService.createKey<boolean>('inQuickOpen', false);
+    this.onShow(() => { this.inQuickOpen.set(true); });
+    this.onHide(() => { this.inQuickOpen.set(false); })
```

packages/monaco/src/browser/monaco-quick-input-service.ts
vince-fugnitto: For reference VS Code does not keep a single key stored in this case, but a map of them:
- https://github.com/microsoft/vscode/blob/34d69db22caab3d00c6752bc7a8681239bc0c5d2/src/vs/platform/quickinput/browser/quickInputService.ts#L56
- https://github.com/microsoft/vscode/blob/34d69db22caab3d00c6752bc7a8681239bc0c5d2/src/vs/platform/quickinput/browser/quickInputService.ts#L121-L147

marcdumais-work:
Closing in favor of https://github.com/eclipse-theia/theia/pull/12427 - thanks @FernandoAscencio !

Issue #12407, opened 2023-04-13
Problem view is slow when setting markers of multiple files


hyy215:
### Bug Description:
Problem view is slow when setting markers of multiple files at the same time.
|number of files|time|
|---|---|
|500|4639ms|
|1000|17180ms|
|2000|69509ms|
|10000|too long|

### Steps to Reproduce:

1.Register a command to set markers of multiple files.
```ts
const command: Command = {
    id: 'set-markers',
    label: 'Set Markers'
};

commands.registerCommand(command, {
    execute: () => {
        for (let i = 0; i < 500; i++) {
            const fileUri = new URI(Uri.file(`test-${i}`));
            this.problemManager.setMarkers(fileUri, `test-${i}`, [{
                message: `test-${i}`,
                range: {
                    start: { line: i, character: 0 },
                    end: { line: i, character: 5 }
                },
                severity: DiagnosticSeverity.Error
            }]);
        }
    }
});
```
2. Excute the command.
3. View chrome devtools performance panel.
Inserting node to marker tree,  updating problem decoration, and setting makers for a monaco model take up too much time.
![500](https://user-images.githubusercontent.com/34715914/231662116-52ae7c9c-2cf4-45df-ac7b-a4537e70faef.gif)
![7bbd7e08f49e2b8faa6196321c77f1d](https://user-images.githubusercontent.com/34715914/231662123-366bf8fa-38ed-4d4e-a912-d38bfda89d36.jpg)
![9708b469779cf17779c67348363c5d7](https://user-images.githubusercontent.com/34715914/231662128-b2b36343-8b9e-4f09-9ac1-b125998c248f.jpg)

### Additional Information

- Operating System: win10(cpu i7-8700K, memory 16G)
- Theia Version: master [c335538](https://github.com/eclipse-theia/theia/commit/c33553806b7dc3d17d4a820627104280bc84ab3a)

PR #12408, opened 2023-04-13
Improve problem view performance

hyy215:
#### What it does

Fixed [#12407](https://github.com/eclipse-theia/theia/issues/12407)
1. Delay inserting nodes to marker tree. Collect and sort the nodes, and insert to the root.
If insert N nodes one by one, it maybe compare 1 + 2 + ... +  (N - 2) + (N - 1) + N times.
2. Creates a debounced function that delays updating decoration.
3. Setting markers for a model only if the monaco model exists.

#### How to test

Check that the bugs descriped in [#12407](https://github.com/eclipse-theia/theia/issues/12407). 
- Operating System: win10(cpu i7-8700K, memory 16G)

|number of files|before|now|
|---|---|---|
|500|4639ms|85ms|
|1000|17180ms|122ms|
|2000|69509ms|217ms|
|10000|too long|1160ms|

#### Review checklist

#### Reminder for reviewers


packages/markers/src/browser/problem/problem-decorations-provider.ts
```diff
@@ -35,13 +36,15 @@ export class ProblemDecorationsProvider implements DecorationsProvider {
...
-            this.currentUris = newUris;
-        });
+        this.problemManager.onDidChangeMarkers(() => this.fireDidDecorationsChanged());
     }
 
+    private fireDidDecorationsChanged = debounce(() => {
+        const newUris = Array.from(this.problemManager.getUris(), stringified => new URI(stringified));
+        this.onDidChangeEmitter.fire(newUris.concat(this.currentUris));
+        this.currentUris = newUris;
+    }, 10);
```
msujew, line 46: Suggestion: A delay of 10ms is fairly short, and might not yield the expected improvements on a poor network connection. We should increase this delay to 50ms, which is in line with most other delays that we have in the frontend.

packages/markers/src/browser/problem/problem-tree-model.ts
```diff
@@ -77,12 +79,21 @@ export class ProblemTree extends MarkerTree<Diagnostic> {
...
+        ProblemCompositeTreeNode.addChildren(this.root as MarkerRootNode, this._markers);
+
+        for (const { node, markers } of this._markers.values()) {
+            const children = this.getMarkerNodes(node, markers);
+            node.numberOfMarkers = markers.length;
+            this.setChildren(node, children);
+        }
+
+        this._markers.clear();
+    }, 10);
```
msujew, line 96: Suggestion: We should use 50ms, see comment above.

packages/markers/src/browser/problem/problem-composite-tree-node.ts
```diff
@@ -33,33 +33,46 @@ export namespace ProblemCompositeTreeNode {
         parent.severity = maxSeverity;
     };
 
-    export function addChild(parent: CompositeTreeNode, child: MarkerInfoNode, markers: Marker<Diagnostic>[]): CompositeTreeNode {
-        ProblemCompositeTreeNode.setSeverity(child, markers);
+    export function addChildren(parent: CompositeTreeNode, insertChildren: Map<string, { node: MarkerInfoNode, markers: Marker<Diagnostic>[] }>): void {
```
msujew, line 36: Question: What is the significance of storing the children in a map, if we only access the values of the map in this function? IMO we should just use a plain array in that case.

hyy215: Array is slower than Map in some cases. But in this case, they take about the same amount of time. I will use array.

hyy215: > Question: What is the significance of storing the children in a map, if we only access the values of the map in this function? IMO we should just use a plain array in that case.

Array is slower than Map in some cases. But in this case, they take about the same amount of time. I will use array.

msujew (approved):
Thank you for the contribution, looks good to me 👍

PR #12418, opened 2023-04-14
In MonacoEditorModel, use value for snapshot

### LINKED ISSUE #12406: Save As has a 64Kib size limit
### Bug Description:
Normal save works fine. The `File > Save As...` produces a copy that cuts off after 65536 bytes.

image

### Steps to Reproduce:

1. Open yarn.lock.
2. Use File > Save As..., and give the file a suitable name
3. Check the file sizes

### Additional Information

- Operating System: RHEL7
- Theia Version: master branch as of today. I haven't checked how far back it goes.

colin-grant-work:
#### What it does

Fixes #12406 by changing the snapshotting logic in `MonacoEditorModel`

It turns out that the brand of `ITextModel` that we use in Monaco has this size restriction on its snapshots:

https://github.com/microsoft/vscode/blob/7bd1368261677fa160c28ff043c444d905331ee7/src/vs/editor/common/model/textModel.ts#L155-L157

That means that in general, we want to use the value rather than delegating to that implementation of `createSnapshot`.

#### How to test

1. Create a large file (> 64kb)
2. Use save as to create a copy of that file.
3. Check that the new file has the same size and content as the original.
4. Create a large untitled file.
5. Use save (as) to save it.
6. Ensure that the new file has the same size and content as its untitled source.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-editor-model.ts
+++ b/packages/monaco/src/browser/monaco-editor-model.ts
@@ -550,7 +550,7 @@ export class MonacoEditorModel implements IResolvedTextEditorModel, TextEditorDo
         }
 
         const contentLength = this.model.getValueLength();
-        const content = this.model.createSnapshot() || this.model.getValue();
+        const content = this.model.getValue();
         try {
             const encoding = this.getEncoding();
             const version = this.resourceVersion;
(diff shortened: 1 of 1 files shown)


vince-fugnitto (approved):
LGTM 👍 
I confirmed that the size after performing a `save as...` is the same as the original.

PR #12420, opened 2023-04-14
Enable workspace file extension customization

msujew:
#### What it does

As of now, only `.theia-workspace` and `.code-workspace` files are supported when selecting/creating workspace files. This change allows to easily customize the workspace file extension and also to exclude `.code-workspace` files.

#### How to test

1. Change the `WorkspaceFileService#getWorkspaceFileTypes` method to return a different array.
2. Assert that services that interact with workspace files adapt as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/workspace/src/common/workspace-file-service.ts
+++ b/packages/workspace/src/common/workspace-file-service.ts
@@ -1,0 +1,63 @@
+// *****************************************************************************
+// Copyright (C) 2023 TypeFox and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { URI } from '@theia/core';
+import { injectable } from '@theia/core/shared/inversify';
+import { FileStat } from '@theia/filesystem/lib/common/files';
+
+export interface WorkspaceFileType {
+    extension: string
+    name: string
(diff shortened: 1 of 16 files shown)


msujew:
@tsmaeder The reasoning behind this PR isn't to change the workspace file format, but to allow adopters to specifiy their own file extensions. Right now, it's hard coded to be either `.theia-workspace` or `.code-workspace`. End-users of adopter IDEs usually don't care that their IDE is built on top of Theia and they expect everything to be branded towards the actual IDE.

See also [this question](https://community.theia-ide.org/t/customize-workspace-file-ending/2230) we had on the discourse forum.

tsmaeder (commented):
Looked through the code and it seems O.K. I'm just wondering whether it would make more sense to model this as "a list of extensions, one of which is the default" instead of "the default extension plus possibly VS Code".

packages/workspace/src/common/workspace-file-service.ts
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * Example: We should not try to read the contents of an .exe file.
+     */
+    isWorkspaceFile(candidate: FileStat | URI): boolean {
+        const uri = FileStat.is(candidate) ? candidate.resource : candidate;
+        const extensions = this.getWorkspaceFileTypes().map(type => `.${type.extension}`);
+        return extensions.includes(uri.path.ext);
+    }
+
+    getWorkspaceFileTypes(): WorkspaceFileType[] {
```
tsmaeder, line 46: The fact that the first returned file type is used as the default file type is not apparent, here, an adopter would have to look at the usages to discover that fact. I would either document the facts that 
1. You need to return at least one file type
2. The first file type is considered the "default file type".

or else add a property/method `defaultFileType` or `defaultFileTypeIndex` in order to encapsulate the convention in this class.

tsmaeder (commented):
Some people are never satisfied ;-) Just one more suggestion.

packages/workspace/src/common/workspace-file-service.ts
msujew: Makes sense, thanks for the input :)

tsmaeder (approved):
Looks good to me now.

paul-marechal:
I cancelled an old pending review and GH decided to publish it... Don't mind me.

Issue #10788, opened 2022-02-23
change `postConstruct` functions from asynchronous to synchronous


paul-marechal:
A problem in our code base is that some of those `postConstruct` functions return promises. Inversify won't wait on them, as the instance creation step assumes everything happens synchronously. This means that more often than not, by the time our asynchronously-post-constructed services are injected, dependents might try to call into uninitialized properties...

We can handle asynchronous processes in post-constructors, but it needs to be done with more care.

PR #12425, opened 2023-04-17
Update inversify to fix TypeScript 5.0 errors

msujew:
#### What it does

With TypeScript 5.0, parameter checking for decorators has been made a bit stricter, which seems to clash with the type definitions of `inversify@^5.0.0`. [`inversify@^6.0.0`](https://github.com/microsoft/TypeScript/issues/52435#issuecomment-1405522045) seems to resolve this. This is also related to https://github.com/eclipse-theia/generator-theia-extension/issues/164.

#### How to test

The CI should be green.

#### Review checklist

#### Reminder for reviewers


msujew:
I've noticed a few issues with this PR. I'll put it on hold for now.

msujew:
To explain the issues in more detail: We cannot inject promises into services with inversify 6. This isn't so bad, as we can easily work around this. However, services are now not allowed to contain promise fields at construction time. The following class can no longer be injected due to inversify incorrectly identifiying it as an "async class":

```ts
class InjectableService {
  constructor() {
    this.deferred = new Deferred(); // <-- This contains a promise, which makes this class only injectable in an async context.
  }
}
```

See also https://github.com/inversify/InversifyJS/issues/1482 for more context.

paul-marechal:
Your code snippet wouldn't actually cause any issue with `inversify@6`. But there are definitely breaking changes when _injecting_  promises. If the binding for an injection is a promise then Inversify will require that you resolve your component asynchronously.

Note that I encountered the same issue and opened a PR on Inversify: https://github.com/inversify/InversifyJS/pull/1461

The reason I closed it is because it really feels like a DI anti-pattern to inject promises. When practising dependency injection you want to inject abstractions. Injecting a promise to an abstraction leads to problems such as handling delayed initialization of classes depending on promises that will eventually resolve. So Inversify went the route of "injecting a promise means we need to delay the initialization of your component until it resolves to the actual dependency that actually should be injected".

The quick fix for that specific issue would be to convert all bindings to promises into Inversify's providers:

```ts
// before:
bind(OurPromiseApi).toConstantValue(thePromise);

// after:
bind(OurPromiseApi).toProvider(ctx => async () => thePromise);
```

Then we'll need to replace injection sites like `this.ourPromiseApi.then()` with `this.ourPromiseApi().then()`.

msujew:
@paul-marechal Yeah, you're right. I misinterpreted the error message that I got. The issue was with the async `@postConstruct` that we perform on a lot of services, not with the containing deferred promise.

paul-marechal:
Ah true the async post constructors are also a problem when using `inversify@6`.

Related: https://github.com/eclipse-theia/theia/issues/10788

msujew:
@paul-marechal This is ready for a review now. Note that I had to [change something in the rpc-proxy mechanism](https://github.com/eclipse-theia/theia/blob/63de67510d1cf0df047ef935d9d43acf52879634/packages/core/src/common/messaging/proxy-factory.ts#L244-L247), since Inversify thought that every proxy that we use as a dependency is in fact a `Promise` object, since it has a `then` function. 

I'll add all the breaking changes after this gets an initial review. There are quite a few breaking changes in this PR, and I want a review on the general approach before documenting them all.

packages/core/src/browser/keyboard/browser-keyboard-layout-provider.ts
```diff
@@ -59,7 +59,11 @@ export class BrowserKeyboardLayoutProvider implements KeyboardLayoutProvider, Ke
     }
 
     @postConstruct()
-    protected async initialize(): Promise<void> {
+    protected initialize(): void {
+        this.doInit();
+    }
```
paul-marechal, line 64: Just for the sake of being consistent but [you already renamed a previous post constructor to `init`](https://github.com/eclipse-theia/theia/pull/12425/files#diff-7786689c4e73bd3c218a39af4560a369240cb1f336e1d4c968c9bffd8d378aa7R199) so I think we could do it here too.

packages/core/src/common/performance/stopwatch.ts
```diff
@@ -50,7 +50,7 @@ export abstract class Stopwatch {
     @inject(ILogger)
     protected readonly logger: ILogger;
 
-    protected constructor(protected readonly defaultLogOptions: LogOptions) {
+    constructor(protected readonly defaultLogOptions: LogOptions) {
```
paul-marechal, line 53: It seems to make sense to remove the `protected` keyword but I didn't see where else it might have caused issues in this PR? I am not asking to get rid of this change, just trying to understand what I missed.

msujew: `protected constructor` does some weird stuff to the constructor return type, which the `@injectable` annotation cannot deal with. See:

image

msujew:
@paul-marechal I've updated the migration and will now go on vacation. For any outstanding comments, please provide the changes yourself. I trust you :)

paul-marechal:
Just caught async dep issues on Electron and managed to fix them.

Please have one last look and merge if you don't find any other regressions/issues.

PR #12429, opened 2023-04-18
plugin: Support LogOutputChannel

### LINKED ISSUE #12017: [vscode] Support LogOutputChannel
see https://code.visualstudio.com/api/references/vscode-api#LogOutputChannel
since vscode 1.74

This requires also 
- support for [LogLevel](https://code.visualstudio.com/api/references/vscode-api#LogLevel) 
- support in [namespace/env](https://code.visualstudio.com/api/references/vscode-api#env): loglevel & onDidChangeLogLevel

ndoschek:
#### What it does

- Support LogOutputChannel
- Support LogLevel
- Support in namespace/env: logLevel & onDidChangeLogLevel

Resolves #12017

##### Remark:
This PR contributes the basic VS Code API support for LogOutputChannels, to fully support all aspects of a LogOutputViewChannel as in VS Code some additional features will be needed:
  - In VSCode there is support to set the "Developer log level" in the application which also configures the log level for LogOutputviewChannels. Currently the log level for LogOutputChannels is set to the default log level, i.e. info.
  - VSCode also writes the logs to a separate file per extension and offers a command and tool in the output view to directly open the extensions's log file.
 
I suggest to extract those additional features into separate follow-up issues if you agree.

[Edit 2023-04-21]: Extracted to follow-up issue #12440

#### How to test

To test, I created a simple extension that creates a LogOutputChannel and contributes commands to send log messages in different levels:
[Edit 2023-04-21: updated VSIX test extension] [vscode-extension-log-output-channel-12017-0.0.1.zip](https://github.com/eclipse-theia/theia/files/11294198/vscode-extension-log-output-channel-12017-0.0.1.zip)

1. Download and install the test extension `vscode-extension-log-output-channel-12017-0.0.1.vsix` in your Theia test workspace (via `Extensions` view - top menu bar `More actions...` - `Install from VSIX`).
2. On activation of the extension, the log output channel `LogOutputChannel-Test-12017` is shown in the output view and the initial message (`LogLevel set to: Info`) should be shown.
3. Execute the commands (`Log a <loglevel> message via LogOutputChannel (with args)` / `Throw an ERROR and log msg/stack via LogOutputChannel` / `Test replace LogOutputChannel` / `Append text content (as new line) to LogOutputChannel`) via the command palette and check for the messages in the output channel.
Remark: Currently the loglevel is set to default INFO log level.

#### Review checklist

#### Reminder for reviewers


ndoschek:
We will do a first internal review cycle and mark this one as ready for review afterwards.

packages/plugin/src/theia.d.ts
```diff
@@ -5250,6 +5353,15 @@ export module '@theia/plugin' {
          */
         export function createOutputChannel(name: string): OutputChannel;
 
+        /**
+         * Creates a new {@link OutputChannel output channel} with the given name.
+         * If options are given, creates a new {@link OutputChannel output channel} with the given name.
```
kittaakos, line 5358: This does not make much sense here. `options` is mandatory, and it will create a `LogOutputChannel` instance and not an `OutputChannel`. What about copying the docs from Code:

> Creates a new [log output channel](https://code.visualstudio.com/api/references/vscode-api#LogOutputChannel) with the given name.

packages/plugin-ext/src/common/plugin-api-rpc.ts
```diff
@@ -286,7 +286,8 @@ export interface TerminalServiceExt {
     getEnvironmentVariableCollection(extensionIdentifier: string): theia.EnvironmentVariableCollection;
 }
 export interface OutputChannelRegistryExt {
-    createOutputChannel(name: string, pluginInfo: PluginInfo): theia.OutputChannel
+    createOutputChannel(name: string, pluginInfo: PluginInfo): theia.OutputChannel,
+    createOutputChannel(name: string, pluginInfo: PluginInfo, options?: { log: true }): theia.LogOutputChannel
```
kittaakos, line 290: `options` is not optional here.

kittaakos (commented):
It works great 👍  I even verified with an `Error` after modifying your test VSIX:


```
2023-04-20T07:52:54.937Z [Error] Error: Logging an ERROR message via LogOutputChannel
    at /Users/a.kitta/dev/git/theia/plugins/vscode-extension-log-output-channel-12017-0.0.1/extension/out/extension.js:40:34
    at /Users/a.kitta/dev/git/theia/packages/plugin-ext/lib/plugin/command-registry.js:61:59
    at CommandRegistryImpl.executeLocalCommand (/Users/a.kitta/dev/git/theia/packages/plugin-ext/lib/plugin/command-registry.js:108:20)
    at CommandRegistryImpl.$executeCommand (/Users/a.kitta/dev/git/theia/packages/plugin-ext/lib/plugin/command-registry.js:70:25)
    at /Users/a.kitta/dev/git/theia/packages/plugin-ext/lib/common/proxy-handler.js:91:71
    at process.processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async RpcProtocol.handleRequest (/Users/a.kitta/dev/git/theia/packages/core/lib/common/message-rpc/rpc-protocol.js:167:28)
```

Regarding the API, when creating the `LogOutputChannel`, `options` must be defined, and `log` must be `true` to be in sync with Code API. What do you think?

packages/plugin-ext/src/common/plugin-api-rpc.ts
tsmaeder: This is an internal interface, so we can make it whatever we want ;-). But you're right, it probably shouldn't be.

packages/plugin/src/theia-proposed.d.ts
```diff
@@ -168,35 +168,6 @@ export module '@theia/plugin' {
         color?: ThemeColor;
     }
 
-    // #region LogLevel: https://github.com/microsoft/vscode/issues/85992
-
-    /**
-     * The severity level of a log message
-     */
-    export enum LogLevel {
-        Trace = 1,
```
tsmaeder, line 177: It's only "proposed", but we might want to add a breaking change notice in case an adopter is using this API.

packages/plugin-ext/src/plugin/output-channel-registry.ts
```diff
@@ -13,27 +13,39 @@
...
+    createOutputChannel(name: string, pluginInfo: PluginInfo): theia.OutputChannel;
+    createOutputChannel(name: string, pluginInfo: PluginInfo, options?: { log: true; }): theia.LogOutputChannel;
+    createOutputChannel(name: string, pluginInfo: PluginInfo, options?: { log: true; }): theia.OutputChannel | theia.LogOutputChannel {
         name = name.trim();
         if (!name) {
             throw new Error('illegal argument \'name\'. must not be falsy');
-        } else {
-            return new OutputChannelImpl(name, this.proxy, pluginInfo);
         }
+        const isLogOutput = options && typeof options === 'object' && options.log;
```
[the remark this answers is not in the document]
tsmaeder: There is also `isObject<T>(...)` from types.ts

packages/plugin-ext/src/plugin/output-channel/logoutput-channel.ts
```diff
@@ -0,0 +1,80 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { AbstractMessageLogger, DEFAULT_LOG_LEVEL, LogLevel } from '@theia/monaco-editor-core/esm/vs/platform/log/common/log';
+import * as theia from '@theia/plugin';
+
+import { OutputChannelRegistryMain, PluginInfo } from '../../common/plugin-api-rpc';
+import { toLogLevel } from '../type-converters';
+
+export class LogOutputChannelImpl extends AbstractMessageLogger implements theia.LogOutputChannel {
```
tsmaeder, line 23: I believe the proper filename according to conventions would be `log-output-channel.ts` or else it would have to be `LogoutputChannel`.

packages/plugin-ext/src/plugin/output-channel/logoutput-channel.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    private _disposed: boolean = false;
+    get disposed(): boolean { return this._disposed; }
+
+    override onDidChangeLogLevel: theia.Event<theia.LogLevel>;
+
+    constructor(readonly name: string, protected proxy: OutputChannelRegistryMain, protected readonly pluginInfo: PluginInfo) {
```
tsmaeder, line 30: is the `proxy` not readonly?

packages/plugin-ext/src/plugin/type-converters.ts
```diff
@@ -1385,3 +1386,16 @@ export namespace DataTransfer {
         return dataTransfer;
     }
 }
+
+export function toLogLevel(logLevel: MonacoLogLevel): theia.LogLevel {
+    switch (logLevel) {
+        case MonacoLogLevel.Trace: return types.LogLevel.Trace;
```
tsmaeder, line 1392: One more argument for using `OutputChannelImpl` as a base class: we wouldn't need this conversion.

tsmaeder (changes requested):
I don't think using the logger implementation from VS Code is a good idea in this case. Feel free to convince me, though :-)

packages/plugin-ext/src/plugin/output-channel/logoutput-channel.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    appendLine(value: string): void {
+        this.append(value + '\n');
+    }
+
+    replace(value: string): void {
+        this.info(value);
+        this.proxy.$append(this.name, value, this.pluginInfo);
+    }
```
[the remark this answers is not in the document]
ndoschek: Thanks for catching that! It basically should replace the channel's content with the given text value.
With the updated implementation, we reuse the `OutputChannelImpl` and can reuse the logic, which clears the content first and then appends the given text value.
Also, I updated the test extension, which also offers a command now to call the replace function.

packages/plugin-ext/src/plugin/output-channel/logoutput-channel.ts
ndoschek: Thanks for raising this Thomas! I was a bit biased by vscode's implementation I guess.
I had a look and it makes a lot of sense to extend the `OutputChannelImpl`. However, I mimicked a bit of logic from the formerly used Logger. I marked a block of code as a copy, could you have another look if this is okay this way? Thanks!

packages/plugin-ext/src/plugin/output-channel/logoutput-channel.ts
ndoschek: Good catch, thank you, I agree and also think it should be readonly. However, it wasn't readonly in the `OutputChannelImpl` either, hence I updated it for both channels.

ndoschek:
Thank you both for your reviews!
I updated my implementation and also updated the test extension in the PR description (it now offers a few more commands like replace and error throwing)

Regarding the additional log file functionality in VS Code (see my remark in the PR description), is it okay if I add a follow-up for this?

packages/plugin-ext/src/plugin/output-channel/log-output-channel.ts
```diff
@@ -0,0 +1,132 @@
...
+                const array = Array.prototype.slice.call(arguments) as unknown[];
+                array[0] = message.stack;
+                this.log(LogLevel.Error, this.format(array));
+            } else {
+                this.log(LogLevel.Error, this.format([message, ...args]));
+            }
+        }
+    }
+
+    private format(args: any): string {
```
tsmaeder, line 113: Isn't this just doing this?
```
args.map(arg => stringify the arg here).join(' ');
```

packages/plugin-ext/src/plugin/output-channel/output-channel-item.ts
```diff
@@ -20,7 +20,7 @@ export class OutputChannelImpl implements theia.OutputChannel {
 
     private disposed: boolean;
 
-    constructor(readonly name: string, private proxy: OutputChannelRegistryMain, private readonly pluginInfo: PluginInfo) {
+    constructor(readonly name: string, protected readonly proxy: OutputChannelRegistryMain, protected readonly pluginInfo: PluginInfo) {
```
tsmaeder, line 23: This is a breaking change, as well, I think.

packages/plugin-ext/src/plugin/output-channel-registry.ts
```diff
@@ -13,27 +13,40 @@
...
+import { PLUGIN_RPC_CONTEXT as Ext, OutputChannelRegistryExt, OutputChannelRegistryMain, PluginInfo } from '../common/plugin-api-rpc';
+import { RPCProtocol } from '../common/rpc-protocol';
+import { isObject } from '../common/types';
+import { LogOutputChannelImpl } from './output-channel/log-output-channel';
 import { OutputChannelImpl } from './output-channel/output-channel-item';
 
 export class OutputChannelRegistryExtImpl implements OutputChannelRegistryExt {
 
-    proxy: OutputChannelRegistryMain;
+    private proxy: OutputChannelRegistryMain;
```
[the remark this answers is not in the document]
ndoschek: I'm not sure why I changed that to private in the first commit, I reverted that change again.

packages/plugin-ext/src/plugin/output-channel/log-output-channel.ts
```diff
@@ -0,0 +1,132 @@
...
+import { OutputChannelImpl } from './output-channel-item';
+import { LogLevel } from '../types-impl';
+
+export class LogOutputChannelImpl extends OutputChannelImpl implements theia.LogOutputChannel {
+
+    readonly onDidChangeLogLevelEmitter: Emitter<theia.LogLevel> = new Emitter<theia.LogLevel>();
+    readonly onDidChangeLogLevel: theia.Event<theia.LogLevel> = this.onDidChangeLogLevelEmitter.event;
+    public logLevel: theia.LogLevel;
+
+    constructor(override readonly name: string, protected override readonly proxy: OutputChannelRegistryMain, protected override readonly pluginInfo: PluginInfo) {
```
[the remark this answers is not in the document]
ndoschek: It is also readonly in the superclass, so I guess this is okay this way

packages/plugin-ext/src/plugin/output-channel/log-output-channel.ts
```diff
@@ -0,0 +1,132 @@
...
+            this.onDidChangeLogLevelEmitter.fire(this.logLevel);
+        }
+    }
+
+    getLogLevel(): theia.LogLevel {
+        return this.logLevel;
+    }
+
+    override append(value: string): void {
+        this.info(value);
```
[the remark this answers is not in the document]
ndoschek: Yes it does not explicitely say that, but if we do use the basic append mechanism, the output would look like this:

```
2023-04-21T08:13:28.630Z [Warning] Logging a WARNING message via LogOutputChannel
2023-04-21T08:13:33.981Z [Error] Logging an ERROR message via LogOutputChannel
Append text content as new line to LogOutputChannel
2023-04-21T08:13:41.973Z [Warning] Logging a WARNING message via LogOutputChannel
Append text content to LogOutputChannel2023-04-21T08:13:49.676Z [Warning] Logging a WARNING message via LogOutputChannel
```

Using `.info()` ensures that all logged output is formatted consistently.
VS Code has implemented it the same way, which I would prefer as well as the log output is overall more consistent.
But if you insist, we can also change it to the simple append mechanism.

packages/plugin-ext/src/plugin/output-channel/log-output-channel.ts
```diff
@@ -0,0 +1,132 @@
...
+    /*---------------------------------------------------------------------------------------------
+     *  Copyright (c) Microsoft Corporation. All rights reserved.
+     *  Licensed under the MIT License. See License.txt in the project root for license information.
+     *--------------------------------------------------------------------------------------------*/
+    private checkLogLevel(level: theia.LogLevel): boolean {
+        return this.logLevel <= level;
+    }
+
+    trace(message: string, ...args: any[]): void {
+        if (this.checkLogLevel(LogLevel.Trace)) {
```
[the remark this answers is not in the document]
ndoschek: That's true, thanks for catching that! I changed it to your suggestion.

packages/plugin-ext/src/plugin/output-channel/log-output-channel.ts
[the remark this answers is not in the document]
tsmaeder: But why would you redeclare a variable that is already present in the superclass and visible just fine in the subclass? It's unnecessary for all three constructor params, IMO.

packages/plugin-ext/src/plugin/output-channel/log-output-channel.ts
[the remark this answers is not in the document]
tsmaeder: Hmh...that's odd, why have `append` if it's exactly like `info`? But since this is VS Code API, compatibility trumps common sense. I'd leave it like it is.

packages/plugin-ext/src/plugin/output-channel/output-channel-item.ts
```diff
@@ -65,7 +65,7 @@ export class OutputChannelImpl implements theia.OutputChannel {
         this.proxy.$close(this.name);
     }
 
-    private validate(): void {
+    protected validate(): void {
```
tsmaeder, line 68: This one is breaking, too, as a subclass might already have this method.

tsmaeder (commented):
Some nitpicks left.

tsmaeder:
@ndoschek I just have two nitpicks left, then we're good to go. Works as advertised.

packages/plugin-ext/src/plugin/output-channel/log-output-channel.ts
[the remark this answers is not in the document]
ndoschek: Yes basically I agree, but as I had to write a custom constructor (to initalize the log level) I have no other choice, as Typescript requires them to be overridden due to the same variable names, see `This parameter property must have an 'override' modifier because it overrides a member in base class 'OutputChannelImpl'.ts(4115)`

packages/plugin-ext/src/plugin/output-channel/output-channel-item.ts
ndoschek: Yes thanks, I already added it to the Changelog breaking changes with the last commit.

ndoschek:
Thank you @tsmaeder for your review! I left my replies in the open threads.

packages/plugin-ext/src/plugin/output-channel/log-output-channel.ts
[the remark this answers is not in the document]
tsmaeder: This works for me:
```
constructor(name: string, proxy: OutputChannelRegistryMain, pluginInfo: PluginInfo) {
        super(name, proxy, pluginInfo);
        this.setLogLevel(LogLevel.Info);
    }
```

The "readonly" is a property of the implicitly created field, not the constructor parameter.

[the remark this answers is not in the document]
ndoschek: Thanks  I was not aware of that, that works great! I pushed another update for this.

tsmaeder (approved):
Looks good to me, now.

Issue #4199, opened 2019-01-31
[vscode] support activation events


dehru:
Hello, I've got a vscode extension running in theia version "next" (  "version": "0.4.0-next.e2f3ec66" ).  But there's a slight difference in when the activation() lifecycle method gets called.  

The vscode extension was built with "vscode": "1.1.28", and has this info in it's package.json
```
        "activationEvents": [
		"onCommand:extension.execute-command"
	],
	"main": "./out/extension.js",
	"contributes": {
		"commands": [{
			"command": "extension.execute-command",
			"title": "Hello Execute Command"
		}]
	},
```

And a very simple extension.ts that looks like this.
```
import * as vscode from 'vscode';

export function activate(context: vscode.ExtensionContext) {
	console.log('activate called');
	vscode.commands.executeCommand('setContext', 'thisContext', true);
}

export function deactivate() {}
```

When running in vscode, the activate method is not called until the command is actually triggered through the vscode command palette.

When running inside theia, the activate method is called as soon as the browser is connected / opened.  This makes debugging the activate method a little more challenging.  And could lead to different behavior with vscode extensions.

akosyakov:
Activation events are not supported yet, VS Code extensions are activated eagerly for now.

dehru:
I'm looking for suggestions for how to I might hook into a theia lifecycle events, to allow me to get around some issues I'm having with eager activation of vscode extensions.  

In my project, I need to show or hide certain commands in the CommandPalette, or in the right click context menus, depending on what the context is.

Is there a way that I can use events/context to show / hide things in the command palette or other menus dynamically?

Thank you for any input you might have.

akosyakov:
Hit it again trying to debug LS process of vetur VS Code extensions. It's really hard since debug configuration rely on lazy loading of extensions :(

akosyakov:
For rust extension, it is necessary to fix it, otherwise it starts contributing UI elements regardless whether rust project opened or not.

akosyakov:
- [ ] `onWebviewPanel` https://github.com/microsoft/vscode/blob/893bf4ea1f8c158125746c3fe959ef4e23883282/src/vs/workbench/api/browser/mainThreadWebview.ts#L72
- [ ] `onView` https://github.com/microsoft/vscode/blob/893bf4ea1f8c158125746c3fe959ef4e23883282/src/vs/workbench/browser/parts/views/customView.ts#L499
[...]

PR #12431, opened 2023-04-18
Support 'onTaskType' activation event

alvsan09:
#### What it does

Provides support for `onTaskType` activation event 

see https://github.com/eclipse-theia/theia/issues/4199#issuecomment-507253188

This brings the following benefits:

1) Activates specific plugins rather than all task provider plugins. This is achieved by presenting the 'provided' task types to the user and upon selection activate only the provider(s) for that specific selected task type.
2) The above avoids startup delays, and reduces the memory footprint on the extension host by only loading the task provider plugins which are selected.
3) Enables the use of plugins which support `onTaskType` e.g. 'Gulp', 'Grunt', 'Jake'.

NOTE: The above apply to extensions that have stopped using activation by the command `workbench.action.tasks.runTask` in favor of `onTaskType` activation. For backward compatibility the activation
by the command mentioned  above is still supported.

Additionally, Selecting the option to configure tasks (e.g., Terminal/Configure tasks...) requires and triggers  activation
of all tasks providers.

The solution introduces two pick levels for 'provided' tasks,one for the selection of the task type and the second one to resolve and show the corresponding tasks to the selected type which the user can then pick from.

The following scenarios are impacted:

1) Terminal/Run Task (for a specific task type)
        a) Long resolution (e.g., many provided tasks)
        b) quick resolution (e.g., small number of provided tasks)
2) Terminal/Configure Tasks...
3) Command palette Quick Access (e.g. ?Task)

#### How to test

Prerequisites:
1. Install Gulp CLI (Used by the Gulp builtin extension)

> npm install gulp-cli -g

2. Install Jake CLI (Optionally, Used by the Jake builtin extension)

> npm install -g jake

3. Activate `Auto Detect` preference for `Gulp` to `on`
From the: 
Command palette -> type 'open user settings' -> click and then ->  Navigate to (Extensions -> Gulp) and change 'Auto Detect' to 'on'

4. Activate `Auto Detect` preference for `Jake` to `on`
From the: 
Command palette -> type 'open user settings' -> click and then ->  Navigate to (Extensions -> Jake) and change 'Auto Detect' to 'on' 

5. Use a clone of 'vscode' as the project to test `theia` with
>  gh repo clone microsoft/vscode
*  edit `.vscode/settings.json` to delete the line:
	"gulp.autoDetect": "off"
* edit `.vscode/tasks.json` to remove the only "gulp" task
   this to prevent early activation of gulp.
* Copy a valid Jakefile.js under the root folder.
     Example test content:

```js
desc('Default Task');
task('default', function () {
 console.log('Hello, Jake!');
});

desc('Test Task');
task('test', function () {
  console.log('This is a test task.');
});
```

Steps to test: 
1. Debug `theia` and open the cloned `vscode` repository mentioned above.
2. Under the debugging IDE open the `DEBUG CONSOLE` using the filter `load`
    * Make sure that neither of the following extensions supporting `onTaskType` activation has been loaded/activated i.e. `gulp`, `jake`, `grunt`
3. On the `theia` instance run a `gulp` task i.e. Terminal / Run Task... => Select `gulp` => wait for it to populate the plugin provided tasks on a second quick pick and select a task to run e.g. the `vscode-api-test`
    * On the `DEBUG CONSOLE` observe that `gulp` was the only loaded / activated plugin.
4. If Jake was installed, test the same as above but Run a `Jake` task.
6. Run `Terminal/Configure tasks...
    * Configuring tasks need to activate all tasks providers, therefore `grunt` must have been activated with this action.
8. Run tasks via `Quick Access` i.e., From the command palette, type `?` and then select `task`
    * Make sure that the two quick pick levels work for provided tasks e.g. `gulp`, `jake` (First selecting type, secondly selecting a provided task).
9. Stop Debugging and start it again to repeat the options above in different order e.g. Running `Configure Task...` first and then again running `Quick Access` first.

See test example video:
https://user-images.githubusercontent.com/76971376/232908518-580d56ce-fea2-4eac-b6ac-b018fbb33d20.mp4

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/type-converters.ts
+++ b/packages/plugin-ext/src/plugin/type-converters.ts
@@ -835,10 +835,12 @@ export function fromTask(task: theia.Task): TaskDto | undefined {
     if ('detail' in task) {
         taskDto.detail = task.detail;
     }
-    if (typeof task.scope === 'object') {
-        taskDto.scope = task.scope.uri.toString();
-    } else if (typeof task.scope === 'number') {
+    if (typeof task.scope === 'number') {
         taskDto.scope = task.scope;
+    } else if (task.scope !== undefined) {
+        taskDto.scope = task.scope.uri.toString();
+    } else {
+        taskDto.scope = types.TaskScope.Workspace;
     }
 
     if (task.presentationOptions) {


packages/task/src/browser/quick-open-task.ts
```diff
@@ -16,29 +16,41 @@
...
 
+export type TaskEntry = QuickPickItemOrSeparator | QuickPickValue<string>;
+export namespace TaskEntry {
+    export function isQuickPickValue(item: QuickPickItemOrSeparator | QuickPickValue<String>): item is QuickPickValue<string> {
+        return 'value' in item && typeof item.value === 'string';
+    }
+}
+
+const CHOOSE_TASK = 'Choose task to open';
+const NO_TASK_TO_RUN = 'No task to run found. Configure Tasks...';
```
msujew, line 47: Missing localization (`nls.localize`/`nls.localizeByDefault`). Also, we might want to export these values.

packages/task/src/browser/task-service.ts
```diff
@@ -403,11 +403,14 @@ export class TaskService implements TaskConfigurationClient {
...
-     * Returns an array of the task configurations which are provided by the extensions.
+     * Returns an array that contains the task configurations provided by the task providers for the specified task type.
      * @param token  The cache token for the user interaction in progress
+     * @param type The task type (filter) associated to the returning TaskConfigurations
+     *
+     * '*' indicates all tasks regardless of the type
      */
-    getProvidedTasks(token: number): Promise<TaskConfiguration[]> {
-        return this.providedTaskConfigurations.getTasks(token);
+    async getProvidedTasks(token: number, type?: string): Promise<TaskConfiguration[]> {
```
msujew, line 412: Minor: I don't think it's necessary to have this method async.
```suggestion
    getProvidedTasks(token: number, type?: string): Promise<TaskConfiguration[]> {
```

alvsan09: Both comments applied !! 
Thanks Mark!

packages/plugin-ext/src/plugin/type-converters.ts
```diff
@@ -835,10 +835,14 @@ export function fromTask(task: theia.Task): TaskDto | undefined {
...
-    } else if (typeof task.scope === 'number') {
-        taskDto.scope = task.scope;
+    if (task.scope) {
+        if (typeof task.scope === 'number') {
+            taskDto.scope = task.scope;
+        } else  {
+            taskDto.scope = task.scope.uri.toString();
+        }
+    } else {
+        taskDto.scope = types.TaskScope.Workspace;
```
vince-fugnitto, line 845: minor: we can avoid the nested loops

```suggestion
    if (typeof task.scope === 'number') {
        taskDto.scope = task.scope;
    } else if (task.scope !== undefined) {
        taskDto.scope = task.scope.uri.toString();
    } else {
        taskDto.scope = types.TaskScope.Workspace;
    }
```

vince-fugnitto (commented):
The functionality looks quite good, I only have some minor comments :+1: 

---

I noticed some improvements that can be made in follow-up pull-requests:
- display a `+ Configure Task` entry when no tasks are configured (above `contributed` items)
- add `Show All Tasks...` entry at the bottom of the list
- improve ordering of entries in a multiroot (sort by root then tasks)
- add `Go back ↩` functionality

packages/plugin-ext/src/plugin/type-converters.ts
alvsan09: Thanks @vince-fugnitto, 
I have addressed your comments  and are now implemented in the latest commits, 
I have raised the following ticket to follow up the implementation of the proposed improvements: 
https://github.com/eclipse-theia/theia/issues/12455

vince-fugnitto (approved):
LGTM :+1: 
I confirmed that:
- [x] contributed tasks are now displayed 
- [x] contributed tasks work well, and show progress when loading (ex: gulp)
- [x] plugins support `onTaskType` properly
- [x] the quick-access menu works for tasks
- [x] `configure task...` works well, and displays the loaded tasks (ex: gulp)

PR #12433, opened 2023-04-19
Fix unhandled promise rejection during git operations

erezmus:
#### What it does
This fixes an issue we've seen happening on KSC when there's a conflict while pulling git changes from a remote.

It seems that when exception is thrown from the dugite-git library, an unhandled promise rejection is caught by the 'unhandledRejection' handler which rethrows the error and ends up in the 'uncaughtException' handler, which caused the container to crash on kubernetes. On a local docker compose setup, the container doesn't crash.

#### How to test
* Clone a repository in theia
* In another clone of the same repository, add some extra commits
* Back in theia, change one of the files from previous step
* Pull the changes, it should give you an error message of the conflict
* In the logs, the following should not be visible:

```
2023-04-19T13:35:48.732Z root ERROR Uncaught Exception:  GitError: Your local changes to the following files would be overwritten
```

#### Review checklist

#### Reminder for reviewers


tsmaeder:
@erezmus wondering why this would crash the container on Kubernetes. I have a hard time to imagine why that would be the case.

erezmus:
@tsmaeder not sure why it happened in our cluster but after applying that patch in our theia fork, it fixed the problem :shrug:

packages/git/src/node/git-repository-manager.ts
```diff
@@ -30,7 +30,7 @@ export class GitRepositoryManager {
 
     run<T>(repository: Repository, op: () => Promise<T>): Promise<T> {
         const result = op();
-        result.then(() => this.sync(repository));
+        result.then(() => this.sync(repository)).catch(e => console.log(e));
```
msujew, line 33: Only the `sync` is actually unhandled, whereas the `op()` call is handled at a higher level. If we use a catch-all mechanism here, we potentially swallow errors that should be thrown and caught on the caller site. We should only call `catch` on the `sync` promise:
```suggestion
        result.then(() => this.sync(repository).catch(e => console.log(e)));
```

msujew:
@erezmus Are you still interested in contributing the fix? I have an outstanding comment that I would like to see addressed before we can merge this.

If there's no further feedback, I would close the PR in the next few days.

erezmus:
@msujew i'll try to have a look today or tomorrow, otherwise next week

erezmus:
@msujew I've had a test both on theia and ksc with your adopted change, looks good

msujew (approved):
Awesome, looks good to me as well 👍

Issue #12458, opened 2023-04-26
The default welcome view of the plugin is not displayed


helloworldmygithub:
### Steps to Reproduce:

1.  Run start theia 
2.  Download the installation package in the vscode plugin market
     eg: https://marketplace.visualstudio.com/_apis/public/gallery/publishers/rohinivsenthil/vsextensions/postcode/1.3.9/vspackage
4.  Install from VSIX
5.  The plugin doesn' have a welcom view to create request, in vsocde it works 
![image](https://user-images.githubusercontent.com/37649395/234500987-1bfe2c69-4f64-43ee-a3d7-fbfab2f21c59.png)

![image](https://user-images.githubusercontent.com/37649395/234498015-09b41161-bbca-4c2a-b659-72967e8e991f.png)

### Additional Information

- Operating System:
- Theia Version:

helloworldmygithub:
I am trying to solve this problem
 https://github.com/eclipse-theia/theia/pull/12466

helloworldmygithub:
I have submitted the code and it can solve the problem, but perhaps someone else has a better solution

msujew:
I'll close this issue in favor of https://github.com/eclipse-theia/theia/issues/9361.

PR #12466, opened 2023-04-27
fix: plugin doesn't show welcome view (#12458)

helloworldmygithub:
#### What it does

Closes: #12458 
The pull-request checks the view data provider after the start of a plugin to show the plugin's welcome view.

#### How to test

1. start theia
2. download the plugin installation package postcode 1.3.9）
 3. install the plugin using `isntall from vsix...`

_Before_:
image

_After_:
image

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/hosted/browser/hosted-plugin.ts
+++ b/packages/plugin-ext/src/hosted/browser/hosted-plugin.ts
@@ -303,7 +303,7 @@ export class HostedPluginSupport {
         }
         await this.startPlugins(contributionsByHost, toDisconnect);
 
-        // check data provider after start plugin to show welcom view
+        // check data provider after plugin start to show welcome view
         this.viewRegistry.checkViewDataProvider();
 
         this.deferredDidStart.resolve();
--- a/packages/plugin-ext/src/hosted/browser/hosted-plugin.ts
+++ b/packages/plugin-ext/src/hosted/browser/hosted-plugin.ts
@@ -303,6 +303,9 @@ export class HostedPluginSupport {
         }
         await this.startPlugins(contributionsByHost, toDisconnect);
 
+        // check data provider after start plugin to show welcom view
+        this.viewRegistry.checkViewDataProvider();
+
         this.deferredDidStart.resolve();
 
         this.restoreWebviews();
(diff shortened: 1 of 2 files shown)


packages/plugin-ext/src/hosted/browser/hosted-plugin.ts
```diff
@@ -303,6 +303,9 @@ export class HostedPluginSupport {
         }
         await this.startPlugins(contributionsByHost, toDisconnect);
 
+        // check data provider after start plugin to show welcom view
```
vince-fugnitto, line 306: minor: 
```suggestion
        // check data provider after plugin start to show welcome view
```

packages/plugin-ext/src/main/browser/view/plugin-view-registry.ts
```diff
@@ -780,19 +780,48 @@ export class PluginViewRegistry implements FrontendApplicationContribution {
...
+                            }));
+                            const unsubscribe = () => toDisposeOnDidExpandView.dispose();
+                            view.disposed.connect(unsubscribe);
+                            toDisposeOnDidExpandView.push(Disposable.create(() => view.disposed.disconnect(unsubscribe)));
+                        }
+                    }
+                });
+            }
+        }
+    }
```
vince-fugnitto, line 805: This code duplicates a lot of what `registerViewDataProvider` already does, I don't think it is the proper solution as is.

helloworldmygithub: May I ask if the meaning here is that I should try to reuse code? Or should we solve the problem in a different way

vince-fugnitto: Sorry, I mean that we should not have duplication, and it looks like the current solution duplicates a lot of what `registerViewDataProvider` already does without any clear reason as to why it does. We should find a solution which does not duplicate the code, either by seeing what can be reused once we better understand the problem, or a completely different solution if necessary.

helloworldmygithub: Okay, I understand. In Theia，widgets should be opened only when view data providers are registered, but some vscode plugins only have some welcomeViews and do not register a viewDataProvider. So I want to check if the plugin view where the welcome view is located has been registered a viewDataProvider after the plugin is launched, and if not, give a default

helloworldmygithub: Thank you for your previous suggestion, but I can't find a simpler solution. I will create a new PR to remove the commit messages and wait for other results

PR #12470, opened 2023-04-27
styling: TreeWidget Highlighted nodes full display

FernandoAscencio:
#### What it does
There is a display discrepancy on selected nodes between Theia and VS Code, as seen below:
##### VS Code
VSCode

##### Theia
MasterDisplay

This commit fixes that discrepancy in display. As seen below:
##### This Commit
BranchDisplay

##### Notes
I have added some extra code handling to keep the current display of Theia. However, as seen in VS Code, the left margin/padding for the Search-In-Workspace and Source-Control trees is no different from the Explorer and Reference trees. As such, if the VS Code tree alignment is preferable, it would also reduce the number of lines added.

#### How to test
1. Run example Theia
2. Do a search and modify a file
3. Check that the selection (both $\textcolor{blue}{\textsf{on focus}}$  and $\textcolor{gray}{\textsf{out of focus}}$) highlights the entire row instead of leaving a space.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/search-in-workspace/src/browser/styles/index.css
+++ b/packages/search-in-workspace/src/browser/styles/index.css
@@ -23,7 +23,7 @@
 }
 
 .t-siw-search-container {
-  padding: 5px;
+  padding: 0px 1px;
   display: flex;
   flex-direction: column;
   height: 100%;
@@ -52,8 +52,7 @@
 }
 
 .t-siw-search-container .searchHeader {
-  width: 100%;
-  margin-bottom: 10px;
+  padding: 5px 5px 15px 2px;
 }
 
 .t-siw-search-container .searchHeader .controls.button-container {
@@ -206,7 +205,6 @@
 
 .t-siw-search-container .resultContainer {
   height: 100%;
-  margin-left: 13px;
 }
 
 .t-siw-search-container .result {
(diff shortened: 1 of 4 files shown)


msujew (approved):
Looks good to me. I couldn't see any regressions and the changes fix the issues at hand 👍

packages/core/src/browser/tree/tree-widget.tsx
```diff
@@ -1420,7 +1420,7 @@ export class TreeWidget extends ReactWidget implements StatefulWidget {
         return this.labelProvider.getLongName(node);
     }
     protected getDepthPadding(depth: number): number {
-        return depth * this.props.leftPadding;
+        return (depth + 1) * this.props.leftPadding;
```
vince-fugnitto, line 1423: @FernandoAscencio do you mind explaining this change a bit further, I did not see anything in the pull-request description or commit message. Why do we want to increase the depth padding globally for all trees?

FernandoAscencio: It has been a while. After a review and some testing I found no difference in display. The `+1` will return to the `file-tree-widget` override.

packages/search-in-workspace/src/browser/styles/index.css
```diff
@@ -52,8 +52,7 @@
 }
 
 .t-siw-search-container .searchHeader {
-  width: 100%;
-  margin-bottom: 10px;
+    padding: 5px 5px 15px 2px;
```
vince-fugnitto, line 55: minor: mind fixing the formatting, the padding is inconsistent with the rest of the file.

vince-fugnitto (commented):
Overall the changes look much better when selecting or highlighting nodes in trees.
I only have a minor comment.

Issue #446, opened 2017-09-01
Support dynamic menu items in electron


kittaakos:
Currently, the menu items are built at application startup and have their static nature. That means, `enabled`, `visible` properties are not updated dynamically.

Also, handler enablement is calculated when the application starts.

For now on, I enable everything (#210), just like VSCode and move the command > handler lookup into the `click(()` function.

See: https://github.com/electron/electron/issues/528

kittaakos:
https://github.com/electron/electron/issues/846

kittaakos:
https://github.com/Microsoft/vscode/blob/9c27b11bde339d58a8da0e3f6d5f88f7a4fcafa7/src/vs/code/electron-main/menus.ts#L195

svenefftinge:
I think https://github.com/electron/electron/issues/528#issuecomment-267576538 could be a viable workaround until they support updates in the rendering process.

kittaakos:
I have noticed that too, but when do you fire the `ipcRenderer.sendSync('rebuild-menu');`? Maybe, I am just missing something and this should be the workaround.

svenefftinge:
you are right. Is there a callback on menu show?

kittaakos:
I was looking for something like that but did not find anything useful.

kittaakos:
https://github.com/electron/electron/issues/8999

vince-fugnitto:
@kittaakos I noticed that vscode has the option of having custom menus (much like our browser), would this be a viable option to update our menus to use our browser menus instead so we have more control and customizability over them ?

kittaakos:
This should be possible to do with the current electron version we use.

`node_modules/electron/electron.d.ts`:
```ts
    /**
     * Emitted when a popup is closed either manually or with menu.closePopup().
     */
    on(event: 'menu-will-close', listener: (event: Event) => void): this;
    once(event: 'menu-will-close', listener: (event: Event) => void): this;
    addListener(event: 'menu-will-close', listener: (event: Event) => void): this;
    removeListener(event: 'menu-will-close', listener: (event: Event) => void): this;
    /**
     * Emitted when menu.popup() is called.
     */
    on(event: 'menu-will-show', listener: (event: Event) => void): this;
    once(event: 'menu-will-show', listener: (event: Event) => void): this;
    addListener(event: 'menu-will-show', listener: (event: Event) => void): this;
    removeListener(event: 'menu-will-show', listener: (event: Event) => void): this;
```

[...]

PR #12483, opened 2023-05-04
Add support for "enablement" property in plugin command contributions

### LINKED ISSUE #12426: Implement `enablement` Property in Command Contributions
### Feature Description:
VS Code supports a property called `enablement` on command contributions (see https://code.visualstudio.com/api/extension-guides/command#enablement-of-commands). The property consists of a when-clause which controls whether the command is rendered as enabled or disable, but does not affect command visibility in the UI.

tsmaeder:
#### What it does
Reads the `enablement` property for commands contributed in a plugin `package.json` and updates toolbar items when necessary.

Fixes #12426

#### How to test
1. Install the attached plugin
2. Open the view "Test View Drag And Drop"
3. There are two toolbar actions: the left one prints the selction to the console and the right one changes the enablement of the other action
4. Click on the second action
5. Observe: the toolbar item becomes grayed-out/normal
6. Observe: you cannot execute the action if the item appears disabled and vice-versa.
7. Click on the "more" menu in the view and make the enablement in the menu is correct
8. Open the context menu on an item in the view and make sure the enablement is correct
9. Repeat the above tests in electron-native and electron-custom mode
10. Repeat the above tests in browser mode
11. Check main-menu behavior: in electron-custom and browser mode, enablement/disablement should work, in electron-native mode, main menu items should always be enabled (see  https://github.com/eclipse-theia/theia/issues/446). The "Selection" menu is a good candidate to check the behavior.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
+++ b/packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
@@ -172,7 +172,11 @@ export class TabBarToolbar extends ReactWidget {
     }
 
     protected isEnabled(item: AnyToolbarItem): boolean {
-        return (!!item.command && this.commandIsEnabled(item.command) && this.evaluateWhenClause(item.when)) || (!!item.menuPath && !item.command);
+        if (!!item.command) {
+            return this.commandIsEnabled(item.command) && this.evaluateWhenClause(item.when);
+        } else {
+            return !!item.menuPath;
+        }
     }
 
     protected getToolbarItemClassNames(item: AnyToolbarItem): string[] {
--- a/packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
+++ b/packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
@@ -43,6 +43,7 @@ export class TabBarToolbar extends ReactWidget {
     protected more = new Map<string, TabBarToolbarItem>();
 
     protected contextKeyListener: Disposable | undefined;
+    protected toDisposeOnUpdateItems: DisposableCollection = new DisposableCollection();
 
     @inject(CommandRegistry) protected readonly commands: CommandRegistry;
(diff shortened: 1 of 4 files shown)


tsmaeder:
Need to check something before review.

tsmaeder:
Here's the test plugin:
[custom-view-samples-0.0.1.zip](https://github.com/eclipse-theia/theia/files/11403863/custom-view-samples-0.0.1.zip)

And here's the source:


[tree-view-sample.zip](https://github.com/eclipse-theia/theia/files/11403876/tree-view-sample.zip)

packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
```diff
@@ -162,10 +171,14 @@ export class TabBarToolbar extends ReactWidget {
         </div>;
     }
 
+    protected isEnabled(item: AnyToolbarItem): boolean {
+        return (!!item.command && this.commandIsEnabled(item.command) && this.evaluateWhenClause(item.when)) || (!!item.menuPath && !item.command);
```
planger, line 175: nitpick: maybe it helps readability if we extract two variables with helpful names for each of the two parts of the conditions.

planger (approved):
Thank you Thomas! Works as advertised in browser, electron native and electron custom mode.
Code looks reasonable to me. I just added a minor nitpick inline.

tsmaeder:
@planger I changed the `isEnabled()` logic. Have a quick look?

planger (approved):
Thank you, much better imho!

vince-fugnitto:
@tsmaeder @planger we may have to revert this change, it breaks context-menus in electron. 
For example: https://github.com/eclipse-theia/theia/issues/12548.

tsmaeder:
@vince-fugnitto since action enablement works just fine for other context menus (file explorer or the general context menu on the editor), this might just be a bug in the action enablement that is _revealed_ since we now actually honor the enabled state.

PR #8783, opened 2020-11-26
Fix tree view reveal() behavior

amiramw:
Signed-off-by: Amiram Wingarten <amiram.wingarten@sap.com>

#### What it does

Various fixes to support tree view reveal in several unsupported scenarios and align it with vscode.

- reveal called without focus now opens the view container (left pane) and view (section), without focus
- reveal of a node that its ancestors are collapsed now expand all parents and reveals it
- reveal of a node that was not cached by backend or not fetched by client (for example when showing the tree for the first time) now works
- reveal of a tree element that is not identical (`===`) to cached node now works

#### How to test

- Clone https://github.com/amiramw/vscode-tree-expand
- `yarn && vsce package` to create the vsix file
- Load the vsix in theia
- Test view tree will show in explorer view a tree with nodes **a ➡️ b ➡️ c**

##### Scenarios

First time:
1. Open theia from chrome incognito (like first time).
2. F1 -> Reveal c New Object or F1 -> Reveal c Same Object
3. Test view section is opened, a and b are expanded and c is selected, before this PR nothing happened

From other view container:
1. Go to debug pane.
2. F1 -> Reveal c New Object or F1 -> Reveal c Same Object
3. Explorer is opened. Test view section is opened, a and b are expanded and c is selected, before this PR nothing happened.

From other view container:
1. Go to debug pane.
2. F1 -> Reveal c New Object or F1 -> Reveal c Same Object
3. Explorer is opened. Test view section is opened, a and b are expanded and c is selected, before this PR nothing happened.

With non identical element:
1. Open test view 
2. Expand the tree to show c and then collapse it
3. F1 -> Reveal c New Object or F1 -> Reveal c Same Object
4. Tree is expanded and c is selected. Before this PR only "Reveal c Same Object" worked

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #12489, opened 2023-05-05
Fix TreeView.reveal() behavior. 

### LINKED ISSUE #12488: TreeView.reveal() is broken
### Steps to Reproduce:

1. Install the attached plugin
2. Open the "Test View Drag and Drop" view
3. Select "Say hello" from the view toolbar menu
4. Observe: nothing happens. It's supposed to reveal the item "aaaa".

### Additional Information

- Operating System: Windows
- Theia Version: 1.37 master

tsmaeder:
#### What it does
Fixes the `TreeView.reveal` behavior. The behavior was broken when we moved to a new scheme for generating `TreeNode` ids from `TreeItem`s, In addition, the reveal flags (select, focus, reveal) were not passed to the front end.

Fixes #12488

#### How to test
Make sure the failure case from the related issue now works. In addition, I used the procedure from https://github.com/eclipse-theia/theia/pull/8783 to guard against regressions.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/tree/tree-views.ts
+++ b/packages/plugin-ext/src/plugin/tree/tree-views.ts
@@ -247,11 +247,14 @@ class TreeViewExtImpl<T> implements Disposable {
 
     async reveal(element: T, options?: Partial<TreeViewRevealOptions>): Promise<void> {
         await this.pendingRefresh;
+        const select = !(options?.select === false); // default to true
+        const focus = !!options?.focus;
+        const expand = typeof options?.expand === 'undefined' ? false : options!.expand!;
 
         const elementParentChain = await this.calculateRevealParentChain(element);
         if (elementParentChain) {
             return this.proxy.$reveal(this.treeViewId, elementParentChain, {
-                select: true, focus: false, expand: false, ...options
+                select: select, focus: focus, expand: expand, ...options
             });
         }
     }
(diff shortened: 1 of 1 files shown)


packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -247,11 +247,14 @@ class TreeViewExtImpl<T> implements Disposable {
 
     async reveal(element: T, options?: Partial<TreeViewRevealOptions>): Promise<void> {
         await this.pendingRefresh;
+        const select = !(options?.select === false); // default to true
```
vince-fugnitto, line 250: suggestion:

```suggestion
        const select = options?.select !== false; // default to true
```

`true` → `true`
`false` → `false`
`undefined` → `true`

packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -247,11 +247,14 @@ class TreeViewExtImpl<T> implements Disposable {
...
+        const focus = !!options?.focus;
+        const expand = typeof options?.expand === 'undefined' ? false : options!.expand!;
```
vince-fugnitto, line 252: minor: the trailing `!` does not seem necessary?

```suggestion
        const expand = typeof options?.expand === 'undefined' ? false : options!.expand;
```

packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -247,11 +247,14 @@ class TreeViewExtImpl<T> implements Disposable {
...
 
         const elementParentChain = await this.calculateRevealParentChain(element);
         if (elementParentChain) {
             return this.proxy.$reveal(this.treeViewId, elementParentChain, {
-                select: true, focus: false, expand: false, ...options
+                select: select, focus: focus, expand: expand, ...options
```
vince-fugnitto, line 257: suggestion: can be omitted

```suggestion
                select, focus, expand, ...options
```

vince-fugnitto (approved):
I confirmed that the plugin works as expected :+1:

Issue #12513, opened 2023-05-10
TreeViewOptions.showCollapseAll has no effect


tsmaeder:
### Bug Description:
While we do have the `showCollapseAll` field in `TreeViewOptions`, we never show the "Collapse All" toolbar item on the view. It's simply not implemented.

### Steps to Reproduce:

1. Install the extension from https://github.com/eclipse-theia/theia/issues/12488
2. Open the "Test View Drag and Drop"
3. Observe: there is no "collapse all" toolbar item in the view toolbar.

### Additional Information

- Operating System: Windows
- Theia Version: Theia 1.37 master

PR #12514, opened 2023-05-10
Show "Collapse All" command in tree view toolbar. 

tsmaeder:

#### What it does

Closes https://github.com/eclipse-theia/theia/issues/12513.

Adds a "Collapse All" toolbar item in tree views contributed by plugins if the plugin requests it in the options (`showCollapseAll`). This PR introduces the concept of `DynamicWidget`. It allows to update the toolbar after the contents of a view widget has been create lazily.

#### How to test
Use the procedure from the related issue and make sure the icon appears. You can also use the built-in "NPM SCRIPTS" view as a second test case.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/view-container.ts
+++ b/packages/core/src/browser/view-container.ts
@@ -77,6 +77,20 @@ export namespace BadgeWidget {
 }
 
 /**
+ * A widget that may change it's internal structure dynamically. Current use is for
+ * updating the toolbar when a contributed view is contructed "lazily"
+ */
+export interface DynamicWidget {
+    onChanged: CommonEvent<void>;
+}
+
+export namespace DynamicWidget {
+    export function is(arg: unknown): arg is DynamicWidget {
+        return isObject(arg) && 'onChanged' in arg;
+    }
+}
+
+/**
  * A view container holds an arbitrary number of widgets inside a split panel.
  * Each widget is wrapped in a _part_ that displays the widget title and toolbar
  * and allows to collapse / expand the widget content.
@@ -968,6 +982,10 @@ export class ViewContainerPart extends BaseWidget {
         if (BadgeWidget.is(this.wrapped)) {
             this.wrapped.onDidChangeBadge(() => this.onDidChangeBadgeEmitter.fire(), undefined, this.toDispose);
             this.wrapped.onDidChangeBadgeTooltip(() => this.onDidChangeBadgeTooltipEmitter.fire(), undefined, this.toDispose);
+        }
+
+        if (DynamicWidget.is(this.wrapped)) {
+            this.wrapped.onChanged(() => this.toolbar.updateTarget(this.wrapped));
         }
 
         const { header, body, disposable } = this.createContent();
--- a/packages/core/src/browser/view-container.ts
+++ b/packages/core/src/browser/view-container.ts
@@ -80,13 +80,13 @@ export namespace BadgeWidget {
  * A widget that may change it's internal structure dynamically. Current use is for
  * updating the toolbar when a contributed view is contructed "lazily"
  */
-export interface DynamicWidget {
-    onChanged: CommonEvent<void>;
+export interface DynamicToolbarWidget {
+    onDidChangeToolbarItems: CommonEvent<void>;
 }
 
-export namespace DynamicWidget {
-    export function is(arg: unknown): arg is DynamicWidget {
-        return isObject(arg) && 'onChanged' in arg;
+export namespace DynamicToolbarWidget {
+    export function is(arg: unknown): arg is DynamicToolbarWidget {
+        return isObject(arg) && 'onDidChangeToolbarItems' in arg;
     }
 }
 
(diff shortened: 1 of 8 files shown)


packages/plugin-ext/src/main/browser/plugin-frontend-contribution.ts
```diff
@@ -15,21 +15,56 @@
...
+
     registerCommands(commands: CommandRegistry): void {
         commands.registerCommand(OpenUriCommandHandler.COMMAND_METADATA, {
             execute: (arg: URI) => this.openUriCommandHandler.execute(arg),
             isVisible: () => false
         });
+        commands.registerCommand(PluginApiFrontendContribution.COLLAPSE_ALL_COMMAND, {
+            execute: (widget: Widget) => {
+                if (widget instanceof PluginViewWidget && widget.widgets[0] instanceof TreeViewWidget) {
+                    const model = widget.widgets[0].model;
```
martin-fleck-at, line 45: Do you think it might be valuable to have that 'collapseTree' functionality generically in the model or is it specific to the tree that is usually provided through the plugins? To me, it looks quite generic but I know that roots might be handled differently depending on the expected tree structure.

packages/plugin-ext/src/main/browser/plugin-frontend-contribution.ts
```diff
@@ -15,21 +15,56 @@
...
 @injectable()
-export class PluginApiFrontendContribution implements CommandContribution {
+export class PluginApiFrontendContribution implements CommandContribution, TabBarToolbarContribution {
 
     @inject(OpenUriCommandHandler)
     protected readonly openUriCommandHandler: OpenUriCommandHandler;
 
+    static readonly COLLAPSE_ALL_COMMAND = Command.toDefaultLocalizedCommand({
+        id: 'treeviews.collapseAll',
+        iconClass: codicon('collapse-all')
```
martin-fleck-at, line 34:
```suggestion
        iconClass: codicon('collapse-all'),
        label: nls.localizeByDefault('Collapse All')
```

Otherwise we don't get any hover info on the toolbar button.

packages/plugin-ext/src/main/browser/plugin-frontend-contribution.ts
```diff
@@ -15,21 +15,56 @@
...
+                    }
+                }
+            },
+            isVisible: (widget: Widget) => widget instanceof PluginViewWidget && widget.widgets[0] instanceof TreeViewWidget && widget.widgets[0].showCollapseAll
+        });
+
     }
 
+    registerToolbarItems(registry: TabBarToolbarRegistry): void {
+
```
martin-fleck-at, line 61: nit: Unnecessary empty line?

packages/core/src/browser/view-container.ts
```diff
@@ -76,6 +76,20 @@ export namespace BadgeWidget {
     }
 }
 
+/**
+ * A widget that may change it's internal structure dynamically. Current use is for
+ * updating the toolbar when a contributed view is contructed "lazily"
+ */
+export interface DynamicWidget {
+    onChanged: CommonEvent<void>;
```
martin-fleck-at, line 84: I really like the idea of having a dynamic widget that is able to trigger an update on it's own toolbar but I was wondering about some things:
- Isn't it conceptually very close to the `onDidChange` event on the `ConditionalToolbarItem`, is there any way to re-use that or is it because of the indirection through the ViewContainerPart?
- The name is quite generic even through it has a very specific use case (at the moment), do we really want to rely on such a generic name that might be inadvertently already implemented on a widget? Also I believe we should follow the `onDidX` naming scheme here, according to the [Coding Guidelines](https://github.com/eclipse-theia/theia/wiki/Coding-Guidelines).

martin-fleck-at (changes requested):
Hi @tsmaeder,

Thank you very much for that change! I tested it with a single-root and a multi-root tree and both times the behavior was as expected. The flag is also correctly considered (i.e., if missing of `false` the button is not shown) so everything is working very nicely.

I do have some minor comments from which at least the command label one should be addressed. The others are more general thoughts and questions.

packages/core/src/browser/view-container.ts
tsmaeder: The answer to both, IMO, is no. What's changing here is not a single toolbar item, but the set of toolbar items that apply for this view. The concrete case I'm covering is that the "real" widget that forms the context for action enablement is created lazily and we need to update the toolbar when that happens. However, that lazy creation is happening inside the view widget, and is an implementation detail of the view widget which I don't want to expose. 
I know the terminology is vague, but intentionally so. Once have more cases and understand the issue better, a better naming may become clear.

packages/plugin-ext/src/main/browser/plugin-frontend-contribution.ts
tsmaeder: Yes, this sounds quite generic, but I don't see that refactoring as a necessary part of this PR.

tsmaeder:
Hmh...I noticed that the shown actions are not correct when starting up with the view open in a view container with only one view. This is the case where the `ToolbarAwareTabBar` is showing the toolbar instead of the `ViewContainerPart`. The problem is that widget responsible for contributing to the toolbar is created lazily and we're not re-updating the toolbar at the right time in this case. I've tried to find a solution, but to no avail, so far.

tsmaeder:
@martin-fleck-at I addressed the comments and fixed the behavior when a single view is present in a view container: you can test this by opening the "Test View Drag and Drop" and then refreshing the browser.

packages/plugin-ext/src/main/browser/plugin-frontend-contribution.ts
martin-fleck-at: Understood.

martin-fleck-at (approved):
I re-tested the code and everything seems to be working now fine. Thank you @tsmaeder!

PR #12545, opened 2023-05-18
[ci][3PP License Check] Detect when dash-licenses exits with an internal error

marcdumais-work:
#### What it does

This commit adapts the `check_3pp_licenses.js` script to handle new "127" exit code from dash-licenses. This follows us noticing that it was possible for the script to not detect that `dash-licenses` had encountered an internal error, not related to the 3PPs licenses, that voided the run's results. We will now detect such cases and appropriately exit with an error code, so that if this happens in CI, the job will fail.

See dash-licenses issue:
https://github.com/eclipse/dash-licenses/issues/236

and Pull Request:
https://github.com/eclipse/dash-licenses/pull/237

#### How to test

Check that you can successfully still run `yarn license:check` locally. I guess it would be possible to use a mock `dash-licenses` that exits with code `127` could be used to test the change, but it may not be necessary.

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
LGTM :+1: 
I did a run where errors are expected from dash and they correctly were reported (https://github.com/vince-fugnitto/theia/pull/18).

PR #12551, opened 2023-05-19
replace deprecated substr methods

vladarama:
#### What it does

This PR addresses the use of deprecated "substr" methods by replacing them with the recommended "substring" method. This ensures compatibility with the latest browser versions.

#### How to test

Since most of the arguments in the substr methods in Theia were positive, there are 4 possible cases for the replacement of substr:

**Case 1: a single positive argument**
Ensure that str.substr(startIndex) is replaced by str.substring(startIndex) 

**Case 2: a single negative argument**
Ensure that str.substr(startIndex) is replaced by str.substring(str.length - startIndex) 

**Case 3: two arguments that are both positive**
Ensure that str.substr(startIndex, length) is replaced by str.substring(startIndex, length + startIndex) 

**Case 4: the first argument is 0**
Ensure that str.substr(0, length) is replaced by str.substring(0, length) 

These four cases should cover the replacement of every instance of substr with the newer substring method. 
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-package/src/npm-registry.ts
+++ b/dev-packages/application-package/src/npm-registry.ts
@@ -147,7 +147,7 @@ export class NpmRegistry {
     protected async doView(name: string): Promise<ViewResult> {
         let url = this.props.registry;
         if (name[0] === '@') {
-            url += '@' + encodeURIComponent(name.substr(1));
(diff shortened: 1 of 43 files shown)


paul-marechal (approved):
LGTM, thanks for the cleanup

PR #12559, opened 2023-05-24
dialog: fix button order of `ConfirmSaveDialog`

### LINKED ISSUE #12246: dialog: improve user experience of `ConfirmSaveDialog`
#### Bug Description:

The user-experience of https://github.com/eclipse-theia/theia/pull/12166 should be improved as we currently have the following problems for end-users:

- the "don't save" button is marked as a `main` (primary) button - only one button should be main
- the ordering of the buttons seems incorrect - should be `cancel` > `don't save` > `save all` if we follow our conventions

Screen Shot 2023-03-01 at 12 18 10 PM

#### Steps to Reproduce:

1. start the application
2. confirm that "auto save" is disabled
3. create a new untitled file and make it dirty
4. attempt to close the app and notice the dialog

vladarama:
#### What it does

This PR fixes #12246 by making changes to the UI as per the defined conventions. It reorders the dialog buttons in the `ConfirmSaveDialog` to follow `Cancel` > `Don't save` > `Save all` sequence. This PR also makes sure that only one primary button is highlighted, which is the `Save all` button.

#### How to test

1. Remove AutoSave: `File` -> `AutoSave`
2. Create a new untitled file and make it dirty or make changes to an existing file
3. Attempt to close the app or close the workspace
4. Notice the dialog and test all three options ( `Cancel` > `Don't save` > `Save all`)
5. Make sure that all three options work on both browser and electron

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/dialogs.ts
+++ b/packages/core/src/browser/dialogs.ts
@@ -199,17 +199,18 @@ export abstract class AbstractDialog<T> extends BaseWidget {
     }
 
     protected appendCloseButton(text: string = Dialog.CANCEL): HTMLButtonElement {
-        this.closeButton = this.createButton(text);
-        this.controlPanel.appendChild(this.closeButton);
-        this.closeButton.classList.add('secondary');
-        return this.closeButton;
+        return this.closeButton = this.appendButton(text, false);
     }
 
     protected appendAcceptButton(text: string = Dialog.OK): HTMLButtonElement {
-        this.acceptButton = this.createButton(text);
-        this.controlPanel.appendChild(this.acceptButton);
-        this.acceptButton.classList.add('main');
-        return this.acceptButton;
+        return this.acceptButton = this.appendButton(text, true);
+    }
+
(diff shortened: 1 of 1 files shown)


msujew (commented):
I'm not really sure that this is the correct approach to this issue. First off, we have some regressions with this:

image

I believe a refactoring of the `ConfirmSaveDialog` is in order. I'm not perfectly sure that the dialog should be in charge of actually executing the change. Instead, I would like to see it being decoupled from `ConfirmDialog` and returning some sort of enum:

```ts
enum ConfirmSaveResult {
  Cancel,
  Save,
  DontSave
}
```

That way we can actually save outside of the dialog and also make it easier to order the buttons, without needing to touch the `ConfirmDialog` at all.

msujew (approved):
I think I led you down the wrong path. I've refactored this a bit more to reduce some of the duplication. Looks good to me now 👍

vladarama:
Thanks for the changes Mark, I realize how much I overcomplicated some stuff.

PR #12567, opened 2023-05-26
Properly set the value of the select component in `DebugBreakpointWidget`

### LINKED ISSUE #12547: Setting Hit Count for a breakpoint sometimes sets Expression instead
### Steps to Reproduce:

1. Using Theia Electron (or Browser) Example on the master branch,
add a source breakpoint in a JavaScript (or TypeScript) file.

2. Edit the breakpoint and set a Hit Count.

3. Edit the breakpoint again and remove the Hit Count.

4. Edit the breakpoint once again and set a Hit Count.

5. Note that the breakpoint has an Expression set to the value entered for the Hit Count (which can be seen in a tooltip for the breakpoint, and also in a debugger).

| Observed | Expected |
| -------- | -------- |
| Screenshot |  Screenshot |

pisv:
#### What it does

Fixes #12547 by ensuring that the value of the select component of the `DebugBreakpointWidget` is set to `this.context` in the `render` method.

#### How to test

Use the steps described in #12547 to verify that the issue is fixed. Check that there are no regressions.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/editor/debug-breakpoint-widget.tsx
+++ b/packages/debug/src/browser/editor/debug-breakpoint-widget.tsx
@@ -92,6 +92,8 @@ export class DebugBreakpointWidget implements Disposable {
         }
     }
 
+    private readonly selectComponentRef = React.createRef<SelectComponent>();
+
     @postConstruct()
     protected init(): void {
         this.doInit();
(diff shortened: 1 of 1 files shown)


pisv:
Although this fix seems to work fine and is based on the approach that is in line with other similar places like `DebugConfigurationSelect` and `PreferenceSelectInputRenderer`, I'd like to note that the issue *might* also be fixed in a more general way by ensuring that the `SelectComponent` itself resets its selection in case the `defaultValue` property changes. This could be done by overriding its `componentDidUpdate` method:

```typescript
    override componentDidUpdate(prevProps: Readonly<SelectComponentProps>): void {
        if (this.props.defaultValue !== prevProps.defaultValue && this.value !== this.props.defaultValue) {
            this.value = this.props.defaultValue;
        }
    }

```

However, the alternative fix seems less safe to me.

vince-fugnitto (approved):
LGTM :+1: I can confirm the bug on master, and the changes work as expected.

PR #12593, opened 2023-06-05
Rework sidepanel behavior on tab overflow

### LINKED ISSUE #12416: Align scroll behavior for side panel views with VS Code
As a user, I want to conveniently work with the left toolbar also if it contains many views that go beyond what can be shown in the available space.
Currently Theia shows a scroll bar, if the number of views in the left toolbar go beyond what can be shown in the available space, whereas the scrollbar only becomes visible when users hover over the toolbar. This is not very intuitive.

A behavior similar to VS Code would be more user friendly. Instead of displaying a scroll bar a new icon showing three dots should appear as the last item in the toolbar if there are more views than can currently fit in.

On clicking the three dots, a menu showing all views that aren’t visible shall be displayed.

The toolbar with the three dots at the end should always show the following:

- Views that fit on the screen
- The currently active view
- The three dots

If the currently active view would not fit at the moment, it becomes the last item after the views that still fit and before the three dots. If the currently active view fits on the list of visible views, it remains at the original position and is shown as active.

In our opinion this is major UX improvement compared to the way it is currently handled and should be implemented as the default behavior. However, if the community disagrees we can also aim at making this an opt-in feature that is configurable via preferences.

tortmayr:
#### What it does

Reworks the behavior of the sidepanels if there are  more tabs than they can fit (overflow). Instead of the counter intuitive on-hover scroll bar we introduce behavior similar to VS Code: 
Overflowing tabs will simply be hidden or 'merged' into one menu button at the end of the tabBar. Clicking the button allows revealing of a currently hidden view. Special handling is implemented to ensure that the currently selected view never gets hidden away. When a view that is currently hidden is revelead (i.e. by opening via command) it will be moved to the visible part of the tabbar.

As outline in #12416 this feature is currently intended to become the default behavior and should replace the scroll approach. If there are any objects we could also make in an opt-in feature (toggleable via settings).

Fixes #12416

#### How to test

Feature can be tested by populating the sidebars with a lot of views and/or resizing the window. Following cases should be tested:
- If a view/tab no longer fits, the additional view menu appears. Clicking on the view menu opens up a context menu with commands to reveal the hidden views.
- Clicking an entry of the "addtional views" menu properly reveals the corresponding view and the context menu is updated accordingly after wards
- Closing a view should reorder the tabbar and bring back one of the hidden tabs
- Resizing the window should properly add/remove tabs from the bar and add them to the view menu accordingly
- The currently selected tab should never get hidden when resizing.
- Revealing a hidden view (e.g. via command) properly moves the view in question to the visible part of the bar
- Drag & drop behavior should work as before

And probably a lot of other corner cases that I didn't think about 😉 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/additional-views-menu-widget.tsx
+++ b/packages/core/src/browser/shell/additional-views-menu-widget.tsx
@@ -11,7 +11,7 @@
 // with the GNU Classpath Exception which is available at
 // https://www.gnu.org/software/classpath/license.html.
 //
-// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
 import { inject, injectable } from '../../../shared/inversify';
--- a/packages/core/src/browser/shell/additional-views-menu-widget.tsx
+++ b/packages/core/src/browser/shell/additional-views-menu-widget.tsx
@@ -1,0 +1,71 @@
+// *****************************************************************************
+// Copyright (C) 2023 STMicroelectronics and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { inject, injectable } from '../../../shared/inversify';
+import { Command, CommandRegistry, Disposable, MenuModelRegistry, MenuPath, nls } from '../../common';
+import { Title, Widget, codicon } from '../widgets';
+import { SidebarMenuWidget } from './sidebar-menu-widget';
+import { SideTabBar } from './tab-bars';
+
+export const AdditionalViewsMenuWidgetFactory = Symbol('AdditionalViewsMenuWidgetFactory');
+export type AdditionalViewsMenuWidgetFactory = (side: 'left' | 'right') => AdditionalViewsMenuWidget;
+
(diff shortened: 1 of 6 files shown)


packages/core/src/browser/shell/tab-bars.ts
```diff
@@ -586,7 +595,7 @@ export class ScrollableTabBar extends TabBar<Widget> {
     protected scrollBar?: PerfectScrollbar;
 
     private scrollBarFactory: () => PerfectScrollbar;
-    private pendingReveal?: Promise<void>;
+    protected pendingReveal?: Promise<void>;
```
msujew, line 598: Minor: I'm wondering why we have any `private` fields at all here. Usually, we just have everything `protected`, especially in the `core` package.

msujew (commented):
The change looks fairly good. I only have a single point of critique: When decreasing the size of the window so that two widgets are hidden, increasing the size afterwards only shows one again, while the other one keeps being hidden behind the kebab menu:


https://github.com/eclipse-theia/theia/assets/4377073/10be97ce-8dd2-4ebe-883d-9cb7c9f52506

I would expect the last two entries to reappear at the same time.

packages/core/src/browser/shell/tab-bars.ts
tortmayr: Good point. I changed the visibility of private fields in `tab-bars.ts` class to protected.

tortmayr:
Thanks for the review @msujew. I have rebased and updated the PR.
The ellipsis menu should now behave as expected.

packages/core/src/browser/shell/additional-views-menu-widget.tsx
```diff
@@ -0,0 +1,71 @@
...
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
```
msujew, line 14: This line in the file header is failing the lint CI.

msujew (approved):
Thanks, looks good to me 👍 Only one remark, see below.

packages/core/src/browser/shell/additional-views-menu-widget.tsx
tortmayr: Good catch. Should be fixed now

PR #12596, opened 2023-06-07
cli: Add `check:theia-extensions` command

### LINKED ISSUE #12572: cli: add check for detecting multiple versions of the same theia extension
It seems that a Theia extension could not exist in 2 different versions at runtime as one would override the other in the Dependency Injection context, leading to unexpected behavior during runtime (or is it overridden during the Theia CLI build?). In our case, we develop several Theia extensions and we systematically have to check that they are all on a single version from the product's dependency graph. On one hand, we have a tool similar to the [dependency check utility](https://github.com/eclipse-theia/theia/pull/11483) which require us to know in advance all Theia extensions in our dependency graph, on the other hand, the [Theia CLI filters the dependency graph to get all Theia extensions during the build](https://github.com/eclipse-theia/theia/blob/master/dev-packages/application-package/src/extension-package-collector.ts#L21).

### Feature Description:
Would it be possible for the Theia CLI to automatically detect if a Theia extension with 2 different versions is present in the dependency graph during the build? using an option like "-ensure-theia-extensions-version-uniqueness" would fail the build if this is the case.

sgraband:
#### What it does

The goal of this PR is to improve the existing `theia check:dependencies` script introduced in [this PR](https://github.com/eclipse-theia/theia/pull/11483). The existing script returns a list of all mismatching version numbers for all dependencies. While introducing two versions of the same dependency can always possibly introduce issues, this will be much more common for Theia extensions, than for other dependencies like `loadash` or `rimraf`. Thats why:

This allows to check for duplicate versions of theia extensions only. 
Added the option `only-theia-extensions` to `theia check:dependencies`. 
Theia extensions are identified via their `theiaExtensions` field in the `package.json`. 
The check is running recursively as long as it finds Theia extensions.
The recursion was added to also display cases where more than 2 versions of a dependency might extist.
The advantage of this approach is that a shorter, but more important, list is returned.

Fixes #12572.

#### How to test

- Check out this PR and build with `yarn`
- Go to `dev-packages/cli` and run `yarn link`
- Checkout [this example](https://github.com/eclipsesource/theia/tree/cliExampleApplication) (or create any other example you would like) and run `yarn link "@theia/cli"` in its root
- Run `yarn` to see the output (this will run `theia check:theia-extensions` on postinstall)

The provided example has a few issues built-in:
- The `test-extension` requires a `@theia/core` version of `1.37.2` while the application depends on `1.38.0`
- To test nested (and 3rd-party)  dependencies the following dependencies where added
    - `test-extension` > `test-extension-3` > `@eclipse-emfcloud/modelserver-theia:0.8.0-theia-cr03`
    - `test-extension2` > `test-extension-4` > `@eclipse-emfcloud/modelserver-theia:0.8.0-theia-cr01`
    - `@eclipse-emfcloud/modelserver-theia:0.8.0-theia-cr01` depends on version `1.37.2` of `@theia/(core|filesystem|process|variable-resolver|workpace)`

So the result of `theia check:theia-extensions` should find the following issues:
- Different versions for `@theia/(filesystem|process|variable-resolver|workpace)` from the application and `@eclipse-emfcloud/modelserver-theia`
- Different versions for `@theia/core` from the application, `@eclipse-emfcloud/modelserver-theia` and `test-extension`
- Different versions for `@eclipse-emfcloud/modelserver-theia` from `test-extension-3` and `test-extension-4`

Note that one of the versions will always be hoisted to the root `node_modules` version

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ## v1.40.0 - 
 
 -  Show command shortcuts in toolbar item tooltips. #12660 (https://github.com/eclipse-theia/theia/pull/12660) - Contributed on behalf of STMicroelectronics
+- [cli] added `check:theia-extensions` which checks the uniqueness of Theia extension versions [#12596](https://github.com/eclipse-theia/theia/pull/12596) - Contributed on behalf of STMicroelectronics
 
 <a name="breaking_changes_1.40.0">[Breaking Changes:](#breaking_changes_1.40.0)</a>
 
--- a/dev-packages/cli/src/check-dependencies.ts
+++ b/dev-packages/cli/src/check-dependencies.ts
@@ -32,6 +32,7 @@ interface CheckDependenciesOptions {
     skipHoisted: boolean,
     skipUniqueness: boolean,
     skipSingleTheiaVersion: boolean,
+    onlyTheiaExtensions: boolean,
     suppress: boolean
 }
 
@@ -46,7 +47,9 @@ interface Package {
     /** Whether the package is hoisted or not, i.e., whether it is contained in the root `node_modules`. */
     hoisted: boolean,
     /** Workspace location in which the package was found. */
-    dependent: string | undefined
+    dependent: string | undefined,
+    /** Whether the package is a Theia extension or not */
+    isTheiaExtension?: boolean,
 }
 
 /** Issue found with a specific package. */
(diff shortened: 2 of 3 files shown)


sgraband:
@tsmaeder Would you have time to review this sometime this week or next week? Or should we find someone else to review this?

tsmaeder:
Sorry, this fell off the wagon. I'll make some time tomorrow. That said, could you resolve the conflicts?

tsmaeder (approved):
The code looks fine and seems to do what it promises.

PR #12597, opened 2023-06-07
Track visible viewlet

### LINKED ISSUE #12589: Implement `activeViewlet` and Related Context Keys
### Feature Description:
Some when-clause context keys are unimplemented in Theia (see https://code.visualstudio.com/api/references/when-clause-contexts#visible-view-container-when-clause-context). 

We have definitions of the context keys in `view-context-key-service`, and the values are referenced in some places, but there seems to be no place where the value is actually set.

tsmaeder:
#### What it does
Track active top level widget and set the relevant context keys if the visible widget is either well known or a contributed view container.
Fixes #12589

#### How to test

1. Add the following snippet to `plugin-frontend-contribution.ts`:

```
    @inject(ViewContextKeyService)
    private readonly viewContextKeys: ViewContextKeyService;

        commands.registerCommand({
            id: 'debug context keys',
            label: 'Debug context keys'
        }, {
            execute: () => {
                console.log(`activeViewlet: ${this.viewContextKeys.activeViewlet.get()}`);
                console.log(`activePanel: ${this.viewContextKeys.activePanel.get()}`);
                console.log(`activeAuxiliary: ${this.viewContextKeys.activeAuxiliary.get()}`);
            }
        });
        ...
```
2. Activate various panels, remove and open views, making sure each time that the view context keys are as expected.
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/application-shell.ts
+++ b/packages/core/src/browser/shell/application-shell.ts
@@ -1153,9 +1153,11 @@ export class ApplicationShell extends Widget {
                     tabBar.revealTab(index);
                 }
             }
-            const panel = this.getAreaPanelFor(newValue);
-            if (panel instanceof TheiaDockPanel) {
-                panel.markAsCurrent(newValue.title);
+            const widget = this.toTrackedStack(newValue.id).pop();
+            const panel = this.findPanel(widget);
+            if (panel) {
+                // if widget was undefined, we wouldn't have gotten a panel back before
+                panel.markAsCurrent(widget!.title);
             }
             // Add checks to ensure that the 'sash' for left panel is displayed correctly
             if (newValue.node.className === 'p-Widget theia-view-container p-DockPanel-widget') {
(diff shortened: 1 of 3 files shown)


msujew (commented):
I've noticed that terminal widgets aren't tracked correctly by this change (they also weren't tracked by the old implementation). You will likely need to handle the terminal case in a special manner, since terminal widgets don't use the `TERMINAL_WIDGET_FACTORY_ID` as their `id`. For example, the first terminal the application spawns has `id: 'terminal-0'`, with every subsequent terminal just incrementing that number.

Everything else looks fine to me though 👍

tsmaeder:
>I've noticed that terminal widgets aren't tracked correctly by this change
That's on purpose: VS Code has a single terminal view with multiple terminals running inside it. Theia has multiple terminal views. As a consequence "the terminal view is active" does not really make sense in Theia.

msujew:
> That's on purpose: VS Code has a single terminal view with multiple terminals running inside it. Theia has multiple terminal views. As a consequence "the terminal view is active" does not really make sense in Theia.

Ok, I can agree with that. However, the code itself doesn't really say that, and is more in line with my expectation. Maybe add a comment in there explaining that the current terminal widget will never hit that mapping and should be revised when a consolidated terminal widget like in vscode has been introduced to the framework?

msujew (approved):
Alright, looks good to me 👍

Issue #12390, opened 2023-04-11
Support SSH Remote Development


msujew:
### Feature Description:

This feature has been available as a vscode extension for quite some time. See also [Developing on Remote Machines using SSH](https://code.visualstudio.com/docs/remote/ssh). The plan is to move the installed application backend to a remote system and connect to a server running on that system.

PR #12599, opened 2023-06-08
[remote] Support download of native dependencies

jonah-iden:
#### What it does
This is a part of the Remote SSH feature #12390.
This PR is supposed to solve the problem of needing different native dependencies when copying the backend to a remote machine with a different OS or cpu-architecture.
As discussed in the theia-dev meeting on 6.6.23 these nativ dependencies should be included as assets in the respective theia release.

#### How to test
Its kind of hard to test since the rest of the feature is not yet integrated or used by anything.

The Github Actions workflow file is also deployed [here](https://github.com/jonah-iden/theia-native-dependencies/actions/workflows/node.js.yml) for testing.
For testing the download contributions, it's probably easiest to create a small `BackendApplicationContribution` in the `remote` package and execute the `NativeDependencyDownloadService` `downloadDependencies` in `onStart()` in there like this:
```Typescript
@injectable()
export class TestBackendContrib implements BackendApplicationContribution {
    @inject(DependencyDownloadService) service: DependencyDownloadService;
    onStart(): void {
        this.service.downloadDependencies(`${process.platform}-${process.arch}`);
    }
}
```
 
After that at least for rigrep (all other dependencies are not yet deployed so yet so it kind of download the website). Through debugging you could see that they are downloaded. or maybe writing the to the file system
#### Review checklist

#### Reminder for reviewers


packages/core/src/node/dependency-download.ts
```diff
@@ -0,0 +1,60 @@
...
+import { MaybePromise, URI } from 'src/common';
+
+/**
+ * contribution used for downloading prebuild nativ dependency when connecting to a remote machine with a different system
+ */
+export interface DependencyDownloadContribution {
+    httpOptions?: RequestOptions;
+    skipUnzip?: boolean
+
+    getDownloadUrl(remoteOS: string, theiaVersion: string | undefined): MaybePromise<string>;
```
msujew, line 27: Suggestion: Not every native dependency will be placed into the same directory. For example, the `spawn-helper` from `node-pty` will go into `lib/build/Release`, whereas `pty.node` will be placed in `lib/backend/native`.

Therefore, we need to include more instructions on how to place the downloaded dependencies in the resulting build. Something like:

```ts
export interface FileDependencyResult {
    targetFile: string;
    /** Some files needs to be made executable on the target system */
    chmod?: number;
}

export interface FileDependencyDownload {
    /** Defaults to `true` */
    unzip?: boolean;
    file: FileDependencyResult
    downloadHandler: string | Buffer | () => Promise<Buffer | string>
}

// Directories are expected to be in a zipped format anyway
// We always unzip them and call `files` on each contained file
export interface DirectoryDependencyDownload {
    files: (path: string) => FileDependencyResult;
    downloadHandler: string | Buffer | () => Promise<Buffer | string>
}
```

Also, we can replace `httpOptions` and the arguments for this function with something better:

```ts
export interface DownloadOptions {
    remoteOS: string;
    theiaVersion: string;
    /** These are the `NodeRequestOptions` for the `NodeRequestService` */
    download: (params: string | NodeRequestOptions) => Promise<Buffer>
}
```
```suggestion
    getDownloadUrl(options: DownloadOptions): MaybePromise<FileDependencyDownload | DirectoryDependencyDownload>;
```

packages/core/src/node/dependency-download.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import { RequestOptions } from 'https';
+import { MaybePromise, URI } from 'src/common';
+
+/**
+ * contribution used for downloading prebuild nativ dependency when connecting to a remote machine with a different system
+ */
+export interface DependencyDownloadContribution {
+    httpOptions?: RequestOptions;
+    skipUnzip?: boolean
```
msujew, line 25: We can remove these two properties, see the comment above.

packages/core/src/node/dependency-download.ts
jonah-iden: i meant this feature for just downloading the dependencies. Placing the dependencies correctly is probably something the actual copying step should do. So the copying step would get the folder where the dependencies are and could then put them into the correct location

msujew: I agree. However, the SSH copy mechanism still needs information on where to put these files. I would like to see a pipeline, in which this information doesn't need to be duplicated. I.e. there's only a single contribution for each native dependency and all other systems (like SSH copy or WSL copy, etc.) can just reuse the information from the contributions. We should use the aforementioned `FileDependencyResult` for this.

msujew:
We've decided to integrate this change directly into https://github.com/eclipse-theia/theia/pull/12618. I'll close this PR in favor of the other one.

PR #12608, opened 2023-06-12
Update ripgrep arguments for file search

### LINKED ISSUE #12607: Intermittent file search failure in Electron Forge app
### Bug Description:

After bundling a Windows Theia app using Electron Forge, opening the file-search (`Ctrl+P`) exhibits two issues:

1. The file search doesn't work
2. It instead spawns multiple, empty Windows terminals

Note that this isn't 100% reproducable, but it happens almost every time with the build of the application I'm working on.

Opening the `.exe` file using a terminal does not reproduce issue. There are no indicators on why this happens, neither logs in the frontend nor backend.

### Steps to Reproduce:

1. Bundle a Theia app using Electron Forge on Windows
2. Start the `.exe` and load a workspace
3. Open the file search (`Ctrl+P`) and type something

### Additional Information

- Theia Version: 1.38.0

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/12607

Updates our used arguments and child process spawn options to be aligned with [vscode](https://github.com/eclipse-theia/theia/issues/12607). In particular, the issue exhibited in #12607 was the result of the added `stdio` options property.

Removing the `stdio` field from the options fixed #12607 locally. Also updates the `private` methods to be `protected`.

#### How to test

1. Build the applications for different OS for Browser and Electron.
2. Perform file search, it should work as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/file-search/src/node/file-search-service-impl.ts
+++ b/packages/file-search/src/node/file-search-service-impl.ts
@@ -129,7 +129,7 @@ export class FileSearchServiceImpl implements FileSearchService {
         return [...exactMatches, ...fuzzyMatches].slice(0, opts.limit);
     }
 
-    private doFind(rootUri: URI, options: FileSearchService.BaseOptions, accept: (fileUri: string) => void, token: CancellationToken): Promise<void> {
+    protected doFind(rootUri: URI, options: FileSearchService.BaseOptions, accept: (fileUri: string) => void, token: CancellationToken): Promise<void> {
         return new Promise((resolve, reject) => {
             const cwd = FileUri.fsPath(rootUri);
             const args = this.getSearchArgs(options);
@@ -159,7 +159,7 @@ export class FileSearchServiceImpl implements FileSearchService {
         });
     }
 
-    private getSearchArgs(options: FileSearchService.BaseOptions): string[] {
+    protected getSearchArgs(options: FileSearchService.BaseOptions): string[] {
         const args = ['--files', '--hidden', '--case-sensitive', '--no-require-git', '--no-config'];
         if (options.includePatterns) {
             for (const includePattern of options.includePatterns) {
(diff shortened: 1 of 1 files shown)


giraffesyo:
@msujew @vince-fugnitto 
We noticed a performance regression after these ripgrep changes were introduced, and had to downgrade to the version just before this. This is on linux, inside a container. (Redhat UBI specifically). What we saw was many `rg` processes in the process list, which led us to try and figure out where they were coming from. We tracked it down to theia and then to this change. For now we're just sticking to v1.38.0 which doesn't have this issue.

msujew:
@giraffesyo You can try to override the service in question in your adopter application and try to see which arguments/process spawn options lead to the issue. Since I'm unaware of anyone else having that issue, it might be a very specific set of environment+arguments+process spawn options that lead to this issue.

giraffesyo:
> @giraffesyo You can try to override the service in question in your adopter application and try to see which arguments/process spawn options lead to the issue. Since I'm unaware of anyone else having that issue, it might be a very specific set of environment+arguments+process spawn options that lead to this issue.

Awesome thanks for the very quick reply. That's a great idea, we will try these suggestions out!

giraffesyo:
@msujew Our original assumptions of it being ripgrep were incorrect. There was also the `@theia/git` extension added, which depends on this package: https://github.com/Axosoft/find-git-repositories. That package seems to iterate through all directories in the workspace, to an infinite depth.

We have a mounted sftp using rclone, and the directory structure can be very large on the other side. When putting rclone into verbose mode it showed folders even 10 levels deep being hit before the entire container was OOM killed. After uninstalling the @theia/git extension all is well.

Worth noting, disabling the watchers with the setting 
```json
 {
"files.watcherExclude": {
    "**/**": true,
  }
}
```

also does the trick, but this has the unwanted side-effect that files created don't show up without refreshing the explorer pane.

PR #12618, opened 2023-06-13
Support SSH remote development feature

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/12390

This change adds the functionality to connect to remote backends using SSH. It adds infrastructure to support other types of connections as well (i.e. WSL or Devcontainers), but implements the functionality only for SSH connections.

Note that this change supports only the Electron version of Theia.

#### How to test

Requires an SSH capable system to which the connection can be established. Note that due to https://github.com/eclipse-theia/theia/issues/12782, this requires a fairly updated system with `glibc >= 2.33` (on Linux at least). Also due to a limitation on GitHub actions, it requires a x64 machine. I've simply used a Gitpod workspace.

1. Start the application and run the `SSH: Connect to Host...` command (or click on the remote status bar entry in the bottom left of the application). Enter the remote system address.
2. The local system should connect to the remote system successfully (or display a message on failure).
3. The remote system can be used as expected, all functionality works as it should.

#### Follow-Ups

* Add more target architectures. Currently, only x64 machines are supported by the feature, meaning no arm64 (and therefore no Mac M1 or M2).
* The backend doesn't copy the plugins from the original user to the remote host yet. This will come in a separate PR.
* The security constraints for electron token validation have been slightly relaxed. I plan to reimplement that feature as a more generalized security token for both browser and electron.

#### Review checklist

#### Reminder for reviewers


msujew:
@kkistm I've updated the project readme. It explains the whole remote connection lifecycle in detail 👍

packages/core/src/node/remote/app-native-dependency-contribution.ts
```diff
@@ -0,0 +1,38 @@
...
+import { injectable } from 'inversify';
+import { RemoteNativeDependencyContribution, DownloadOptions, DependencyDownload, RemotePlatform } from './remote-native-dependency-contribution';
+
+@injectable()
+export class AppNativeDependencyContribution implements RemoteNativeDependencyContribution {
+
+    // TODO: Points for testing purposes to a non-theia repo
+    // Should be replaced with:
+    // 'https://github.com/eclipse-theia/theia/releases/download'
+    appDownloadUrlBase = 'https://github.com/msujew/theia/releases/download';
```
marcdumais-work, line 26: I get a 404 on this URL. I do see the native dependencies if I navigate one level up, to "releases".

marcdumais-work: image

marcdumais-work: Nevermind - I see this is just the beginning, to build a valid URL (as reflected by the name).

packages/process/src/node/process-backend-module.ts
```diff
@@ -64,4 +64,5 @@ export default new ContainerModule(bind => {
     bind(MultiRingBuffer).toSelf().inTransientScope();
     /* 1MB size, TODO should be a user preference. */
     bind(MultiRingBufferOptions).toConstantValue({ size: 1048576 });
+
```
marcdumais-work, line 67: left over?

packages/core/src/node/remote/app-native-dependency-contribution.ts
marcdumais-work: NM

marcdumais-work:
> Also due to a limitation on GitHub actions, it requires a x86_64 machine.

I think it's reasonable for a first iteration. 

Among others, there are a couple flavour of `arm`, that would eventually be interesting to cover. I think it's possible since the `VSCodium`project does it successfully, apparently using docker containers and QEMU.

marcdumais-work:
Using the Electron example, I got the exception below, a few seconds after using `SSH: Connnect to host` to connect to a remote computer. 

Then I retried, and it worked. So maybe this is intermittent. I think I'll clear the remote folder and see if I can reproduce.

```
2023-08-18T14:37:24.732Z root ERROR Request establishConnection failed with error: Failed to unzip: 
gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
 Error: Failed to unzip: 
gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now

    at RemoteSetupService.unzipRemote (/home/myuser/theia/examples/electron/lib/backend/main.js:22505:19)
    at async RemoteSetupService.setup (/home/myuser/theia/examples/electron/lib/backend/main.js:22391:13)
    at async RemoteSSHConnectionProviderImpl.establishConnection (/home/myuser/theia/examples/electron/lib/backend/main.js:22775:13)
    at async RpcProxyFactory.onRequest (/home/myuser/theia/examples/electron/lib/backend/packages_core_lib_common_index_js.js:4009:24)
    at async RpcProtocol.handleRequest (/home/myuser/theia/examples/electron/lib/backend/packages_core_lib_common_index_js.js:3448:28)
```

marcdumais-work:
I got the same exception, starting from scratch. It took a few more attempts to get it to work this time. 

IDE pop-up, that goes with the exception: 
image

marcdumais-work:
I have noticed that I can't start a new terminal. I think I see the terminal widget but it disappears very fast. There's a warning each time I try: _"ColorManager.ts:222 Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true. See: https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently"_

Peek 2023-08-18 11-26

msujew:
@marcdumais-work Thanks for looking into this! I was aware of the issue that the server continues running on the remote, but the others are new to me. Due to vacation, I'll come back to this in 2 weeks.

marcdumais-work:
@msujew One potential source of issues is that I used a slightly older desktop machine, as the remote to connect-to. Both older hardware (Intel i7 from ~8-10 years ago) and OS: Ubuntu 20.04 (vs 22.04 on my laptop). It could be that, now that my desktop can start/use the file-watchers (thank you for fixing the GLIBC thing!) and proceed to successfully start the remote backend app, there may still remain some more subtle compatibility issues. 

Such incompatibilities would not be necessarily a defect in the context of this PR. A follow-up could e.g. setup in CI the necessary to build more varied  OS-specific native dependencies archive packages, and in a case like mine,  use the X86-64 package that's specific to Ubuntu 20.04.

marcdumais-work:
Another relatively easy strategy could be used, taking advantage of general better backward compatibility vs forward. i.e. software built on OS version `x`, will generally work too on version `x+1`  and sometimes a few more. The reverse however, if often _not true_, even just one version apart. 

e.g. `x` = `Ubuntu 20.04` and `x+2` = `Ubuntu 22.04`.

So, in our case, that would mean targeting the older supported OS version, when we generate the native dependencies archive packages, as much as possible, if we do not do it already. So, ATM, we would use Ubuntu 20.04 instead of a newer version (assuming we do support that version).

msujew:
@marcdumais-work The issue was that I used `ubuntu-latest` when I initially generated the native dependencies for the GitHub release that this PR uses. Since `ubuntu-latest` is `22.04`, that was the only supported release.

marcdumais-work:
@msujew I see, thanks for the info. When you have time, could the package be deleted and re-created using the newer code? Or maybe it already has? Alternatively, I tried to use Gitpod as remote machine, like you mentioned doing, but could not immediately figure how to do that.

msujew:
> When you have time, could the package be deleted and re-created using the newer code?

I've already done that. I've tested it with an ubuntu-20.04 machine, which worked as expected. Something doesn't work for you?

> Alternatively, I tried to use Gitpod as remote machine, like you mentioned doing, but could not immediately figure how to do that.

Gitpod offers to connect to any running workspace via SSH. You just need to [register a key](https://www.gitpod.io/docs/configure/user-settings/ssh) and then open the dashboard and select `Open with SSH`. Copy the address in the field into the quick input.

msujew:
@marcdumais-work I've fixed the headless server issue. Processes are now terminated when the ssh connection closes.

I believe I was also able to fix the tar issue you were experiencing. It's due incompatible sftp servers on the target machine. See also [here](https://github.com/theophilusx/ssh2-sftp-client#issues-with-fastput-and-fastget-methods). Please let me know whether that works better for you :)

vince-fugnitto (commented):
@msujew I believe that Marc will be unable to complete the review for the time being so he asked me to review in his place, I am however busy with other work but I'll try to review as soon as I can. 

Do you mind describing the procedure to test the use-case with Gitpod? I have added an `ssh` key but I'm running into issues where it cannot complete downloading and installing Node.JS on the remote:

image

msujew:
@vince-fugnitto Thanks for looking into this! Once the initial handshake passes, the ssh connection should be established and no further configuration work should be necessary. So that seems to work well. I've just pushed a commit that makes the node.js setup a bit more robust, but I'm confused by it got stuck in the first place. Can you also check whether Theia creates something in the `~/.theia-example-1.41.0-remote` folder on the remote system?

msujew:
@tsmaeder and I had an offline discussion about this topic. The feature is now using a more reasonable networking architecture now. Please take a look :)

packages/remote/src/electron-node/setup/remote-copy-service.ts
```diff
@@ -0,0 +1,114 @@
...
+        if (this.initialized) {
+            return this.copyRegistry.getFiles();
+        }
+        await Promise.all(this.copyContributions.getContributions()
+            .map(copyContribution => copyContribution.copy(this.copyRegistry)));
+        this.initialized = true;
+        return this.copyRegistry.getFiles();
+    }
+
+    protected async loadNativeDependencies(remotePlatform: RemotePlatform, tempDir: string): Promise<RemoteFile[]> {
```
tsmaeder, line 94: So we're downloading the native dependencies to the local machine and add it to the zip to be sent. I guess that makes sense, since we're not sure the remote has access via http, right?

packages/remote/src/electron-node/setup/remote-node-setup-service.ts
```diff
@@ -0,0 +1,89 @@
...
+        const fileName = this.getNodeFileName(platform);
+        const downloadPath = this.getDownloadPath(fileName);
+        const zipPath = RemotePlatform.joinPath(platform, targetPath, fileName);
+        const download = this.scriptService.downloadFile(platform, downloadPath, zipPath);
+        const unzip = this.scriptService.unzip(zipPath, targetPath);
+        return this.scriptService.joinScript(platform, download, unzip);
+    }
+
+    protected getDownloadPath(fileName: string): string {
+        return `https://nodejs.org/dist/v${REMOTE_NODE_VERSION}/${fileName}`;
```
tsmaeder, line 86: I think how we get this uri should be configurable: think of companies like Ericsson that have environments without access to the public internet. Also for performance and uptime reasons, they might prefer a local copy.

packages/remote/src/electron-node/setup/remote-setup-service.ts
```diff
@@ -0,0 +1,163 @@
...
+            } else if (result.stdout.includes('Linux')) {
+                return 'linux';
+            } else if (result.stdout.includes('Darwin')) {
+                return 'darwin';
+            }
+        }
+        throw new Error('Failed to identify remote system: ' + result.stdout + '\n' + result.stderr);
+    }
+
+    protected async getRemoteHomeDirectory(connection: RemoteConnection, platform: RemotePlatform): Promise<string> {
```
tsmaeder, line 117: All of these methods dispatch on the platform. Why not abstract the differences in a platform-specific "scripts" objects that we instantiate when we detect the platform? (StrategyObject pattern)

packages/remote/src/electron-node/ssh/remote-ssh-connection-provider.ts
```diff
@@ -0,0 +1,355 @@
...
+                    });
+                }
+                if (!await fs.pathExists(identityKey.filename)) {
+                    // Try next identity file
+                    return callback(NEXT_AUTH);
+                }
+
+                const keyBuffer = await fs.promises.readFile(identityKey.filename);
+                let result = ssh2.utils.parseKey(keyBuffer); // First try without passphrase
+                if (result instanceof Error && result.message === 'Encrypted private OpenSSH key detected, but no passphrase given') {
```
tsmaeder, line 157: How stable is this error message?

tsmaeder:
@msujew if I understand it correctly, the original back-end remains after pointing a window to a new back end as a manager of the remote connections and to take over again when a windows disconnects from the remote, right? If we had multiple theia-based products, they would have separate back ends. Would we clash in some way in that case?

One thing I didn't quite understand is how stuff works with resources from webviews? How are the http requests for resources rewritten to fetch stuff from the remote?

packages/core/src/common/quick-pick-service.ts
```diff
@@ -272,6 +272,7 @@ export interface QuickPickOptions<T extends QuickPickItemOrSeparator> {
     onDidTriggerItemButton?: (ItemButtonEvent: QuickPickItemButtonContext<T>) => void
 }
 
+export const quickInputServicePath = '/services/quickInput';
```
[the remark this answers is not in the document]
msujew: Not really. Currently, only the quick-pick service has a backend binding, but I need a quick input service binding as well for the ssh connection provider (to get the entered ssh password).

packages/core/src/browser/remote-service.ts
```diff
@@ -0,0 +1,30 @@
...
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { injectable } from 'inversify';
+
+export const RemoteService = Symbol('RemoteService');
+
+export interface RemoteService {
+    isConnected(): boolean;
```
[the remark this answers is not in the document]
msujew: It is used in the `electron-file-dialog-service.ts` to find out what kind of dialog to display.

packages/core/src/electron-browser/keyboard/electron-keyboard-module.ts
```diff
@@ -21,7 +21,7 @@ import { ElectronKeyboardLayoutChangeNotifier } from './electron-keyboard-layout
 
 export default new ContainerModule((bind, unbind, isBound, rebind) => {
     bind(KeyboardLayoutProvider).toDynamicValue(ctx =>
-        WebSocketConnectionProvider.createProxy<KeyboardLayoutProvider>(ctx.container, keyboardPath)
+        WebSocketConnectionProvider.createLocalProxy<KeyboardLayoutProvider>(ctx.container, keyboardPath)
```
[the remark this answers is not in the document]
msujew: Right, but without trying to navigate older architectural issues in this PR, using the `createLocalProxy` method seems to achieve exactly that. It creates a proxy to the service instance of the local backend instance. It doesn't access the remote backend for that info.

packages/core/src/electron-browser/messaging/electron-local-ws-connection-provider.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    protected override createEndpoint(path: string): Endpoint {
+        const localPort = getLocalPort();
+        if (!localPort) {
+            throw new Error('This should only be called in case there is a local port specified!');
+        }
+        const endpoint = new Endpoint({
+            path
+        }, {
+            host: `localhost:${localPort}`,
```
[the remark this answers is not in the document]
msujew: I believe it's more resilient against ipv6 issues. Since Node 18, addresses are assigned via ipv6 by default (as far as I know). Using `localhost` will resolve against the correct ipv6 address.

packages/core/src/browser/messaging/ws-connection-provider.ts
```diff
@@ -48,23 +50,28 @@ export class WebSocketConnectionProvider extends AbstractConnectionProvider<WebS
         return container.get(WebSocketConnectionProvider).createProxy<T>(path, arg);
     }
 
-    protected readonly socket: Socket;
+    static createLocalProxy<T extends object>(container: interfaces.Container, path: string, arg?: object): RpcProxy<T> {
+        return container.get<WebSocketConnectionProvider>(LocalWebSocketConnectionProvider).createProxy<T>(path, arg);
+    }
+
+    static createDualProxy(container: interfaces.Container, path: string, arg?: object): void {
```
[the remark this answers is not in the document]
msujew: The important part is that the `arg` becomes the endpoint for calls from the local backend. Using the normal `createProxy` only establishes a proxy from the remote backend to the local frontend.

I wanted to add some documentation to this method, but thought the naming would be fine. Do you have a recommendation on naming?

packages/core/src/electron-node/token/electron-token-validator.ts
```diff
@@ -76,8 +79,13 @@ export class ElectronTokenValidator implements WsRequestValidatorContribution {
...
      * Returns the token to compare to when authorizing requests.
      */
-    protected getToken(): ElectronSecurityToken {
-        return JSON.parse(process.env[ElectronSecurityToken]!);
+    protected getToken(): ElectronSecurityToken | undefined {
+        const token = process.env[ElectronSecurityToken];
+        if (token) {
+            return JSON.parse(token);
+        } else {
+            return undefined;
```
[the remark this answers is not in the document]
msujew: Sure, I'll add a comment to the code. Note that I plan to rewrite the electron security token validation anyway - in favor of a general security token that can be used for both electron and browser contexts. However, this is part of a follow up.

packages/core/src/browser/messaging/ws-connection-provider.ts
```diff
@@ -48,23 +48,15 @@ export class WebSocketConnectionProvider extends AbstractConnectionProvider<WebS
...
-                this.reconnectChannelOpeners = [];
-            }
-            this.socket.on('disconnect', () => this.fireSocketDidClose());
-            this.socket.on('message', () => this.onIncomingMessageActivityEmitter.fire(undefined));
-            this.fireSocketDidOpen();
-        });
-        this.socket.connect();
+    }
+
+    @postConstruct()
```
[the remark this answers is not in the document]
msujew: I've reset those changes for now - I don't need them any longer. We can probably tackle this issue in another PR.

packages/core/src/node/remote/app-native-dependency-contribution.ts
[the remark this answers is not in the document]
msujew: Most Theia apps don't add further dependencies that require native bindings/native compilation. If they need it, they are encouraged to create a new `RemoteNativeDependencyContribution` instead of modifying the existing `AppNativeDependencyContribution`.

packages/core/src/node/remote/remote-copy-contribution.ts
```diff
@@ -0,0 +1,85 @@
...
+@injectable()
+export class RemoteCopyRegistry {
+
+    @inject(ApplicationPackage)
+    protected readonly applicationPackage: ApplicationPackage;
+
+    protected readonly files: RemoteFile[] = [];
+
+    getFiles(): RemoteFile[] {
+        return this.files.slice();
```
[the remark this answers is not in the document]
msujew: We could, but is there any real benefit? If people want to modify the objects in the array, they probably know what they're doing.

packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
```diff
@@ -35,10 +36,14 @@ import { OpenDialogOptions, SaveDialogOptions } from '../../electron-common/elec
 export class ElectronFileDialogService extends DefaultFileDialogService {
 
     @inject(MessageService) protected readonly messageService: MessageService;
+    @inject(RemoteService) protected readonly remoteService: RemoteService;
 
     override async showOpenDialog(props: OpenFileDialogProps & { canSelectMany: true }, folder?: FileStat): Promise<MaybeArray<URI> | undefined>;
     override async showOpenDialog(props: OpenFileDialogProps, folder?: FileStat): Promise<URI | undefined>;
     override async showOpenDialog(props: OpenFileDialogProps, folder?: FileStat): Promise<MaybeArray<URI> | undefined> {
+        if (this.remoteService.isConnected()) {
```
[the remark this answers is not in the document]
msujew: I've changed the dialog so that it no longer needs to know whether it runs on a remote server. I'm not sure that is the best solution, but yours is imo out of scope.

packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
```diff
@@ -35,10 +36,14 @@ import { OpenDialogOptions, SaveDialogOptions } from '../../electron-common/elec
 export class ElectronFileDialogService extends DefaultFileDialogService {
```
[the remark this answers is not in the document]
msujew: I agree, but out of scope for this PR.

packages/remote/src/electron-browser/remote-frontend-contribution.ts
```diff
@@ -0,0 +1,145 @@
...
+        });
+        commands.registerCommand(RemoteCommands.REMOTE_DISCONNECT, {
+            execute: () => this.disconnectRemote()
+        });
+    }
+
+    protected disconnectRemote(): void {
+        const port = new URLSearchParams(location.search).get('localPort');
+        if (port) {
+            this.windowService.reload({
```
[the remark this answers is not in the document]
msujew: Tunnel and remote backend are closed and terminated, respectively.

packages/remote/src/electron-node/setup/remote-copy-service.ts
msujew: Exactly, neither http ingress nor egress is garanteed.

packages/remote/src/electron-node/setup/remote-node-setup-service.ts
msujew: For now adopters can just override the variable - making it entirely user configurable is a bit difficult, as the exact download location depends on the remote platform and architecture.

packages/remote/src/electron-node/ssh/remote-ssh-connection-provider.ts
msujew: Good question - I've made this a bit more flexible only listening on the `no passphrase given` info, which seems to be consistent across all ssh client implementations.

msujew:
> If I understand it correctly, the original back-end remains after pointing a window to a new back end as a manager of the remote connections and to take over again when a windows disconnects from the remote, right?

We need the local backend for 3 reasons:
1. Manage the remote connection on the local system (i.e. ssh handling, proxying, etc.)
2. Allow other windows to work with the local system (we only use a single backend instance for multiple local windows on desktop)
3. Local-only services such as the remote status or keyboard services. These services need a connection to the local backend - even when connected to a remote backend.

> If we had multiple theia-based products, they would have separate back ends. Would we clash in some way in that case?

We shouldn't clash, all potentially clashable resources are:
1. Local ports - randomly assigned from one of the free ports
2. Remote ports - randomly assigned from one of the free ports
3. Theia backends - addressable by application name + version

The only clash could happen if two distinct Theia apps use the same name and version. Then the remote extension would incorrectly assume that the remote system already has a remote backend available to start up. However, I think this is quite unlikely.

> One thing I didn't quite understand is how stuff works with resources from webviews? How are the http requests for resources rewritten to fetch stuff from the remote?

The local backend listening on `localhost:port` accepts any request coming in from any of its subdomains - i.e. `{{guid}}.localhost:port`. It then redirects the request to the remote system via ssh port forwarding on `localhost:remotePort`, losing the subdomain info. That the subdomain info is missing is of no consequence for the remote system, since it can serve the files just as well via the localhost directly. The important part is that the frontend thinks it accesses a subdomain so it doesn't expose cookies and other security relevant information to other webviews.

packages/remote/src/electron-node/setup/remote-node-setup-service.ts
tsmaeder: Having a hard-coded URL just makes me cringe. If the URL ever changes, we'll have to fix old versions of Theia. Wouldn't a simple template string we read from configuration (preferences?) take care of most use cases?

packages/remote/src/electron-node/ssh/remote-ssh-connection-provider.ts
tsmaeder: But does it make sense to even check the error message? Can't we just try _with_ the passphrase if it doesn't work without in any case?

packages/core/src/electron-browser/keyboard/electron-keyboard-module.ts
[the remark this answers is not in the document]
tsmaeder: If it was the only instance where we use a local proxy, we could have gotten rid of `createLocalProxy`, but it's not.

packages/core/src/electron-browser/messaging/electron-local-ws-connection-provider.ts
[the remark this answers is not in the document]
tsmaeder: Not an expert, just making sure you thought about it.

packages/remote/src/electron-node/setup/remote-node-setup-service.ts
```diff
@@ -0,0 +1,111 @@
...
+        const dirName = `node-v${REMOTE_NODE_VERSION}-${platformId}-${platform.arch}`;
+        return dirName;
+    }
+
+    protected validatePlatform(platform: RemotePlatform): void {
+        if (platform.os === OS.Type.Windows && !platform.arch.match(/^x(64|86)$/)) {
+            this.throwPlatformError(platform, 'x64 and x86');
+        } else if (platform.os === OS.Type.Linux && !platform.arch.match(/^(x64|armv7l|arm64)$/)) {
+            this.throwPlatformError(platform, 'x64, armv7l and arm64');
+        } else if (platform.os === OS.Type.Linux && !platform.arch.match(/^(x64|arm64)$/)) {
```
tsmaeder, line 61: "Mac"?

packages/remote/src/electron-node/setup/remote-node-setup-service.ts
msujew: Alright, the URL is now configurable 👍

packages/core/src/electron-node/hosting/electron-ws-origin-validator.ts
```diff
@@ -15,13 +15,21 @@
...
 
 @injectable()
 export class ElectronWsOriginValidator implements WsRequestValidatorContribution {
 
+    @inject(BackendRemoteService)
+    protected readonly backendRemoteService: BackendRemoteService;
+
     allowWsUpgrade(request: http.IncomingMessage): boolean {
+        // If we are running as a remote server, requests will come from an http endpoint
+        if (this.backendRemoteService.isRemoteServer()) {
```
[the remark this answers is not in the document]
msujew: > Would it make sense to reify the back end as an object we can ask questions?

Sure, though I would do that in a separate PR. Likely the same one where I also fix the electron security token issue.

tsmaeder:
@msujew I'd add "make it work with ARM" to the follow-ups. Otherwise, we'll have a lot of disgruntled mac users ;-)

tsmaeder:
@msujew when I connect to my Windows localhost, I get an error after entering my password:

"Could not open SSH connection to remote. Failed to create directory: 'New-Item' is not recognized as an internal or external command, operable program or batch file."

tsmaeder:
I've tested that I can connect from a windows host to windows/linux and from a linux host to windows/linux. Works well withing the constraints detailed in the description. Plugins that are installed on the local machine (as opposed to built-in) are picked up by the remote back-end. That is something we might want to look up when solving the "transport local plugins" problem.

PR #12620, opened 2023-06-14
fix for windows execute command in terminal

jonah-iden:
#### What it does
This fixes an issue that on windows commands where only pasted in the terminal and needed the user to press Enter again instead of being exeuted directly like in other Operating Systems

#### How to test
I tested it with the [codeLLDB](https://open-vsx.org/extension/vadimcn/vscode-lldb) extension:
1. Create a basic rust or c/c++ project
2. create a codeLLDB debug configuration for launching the project.
3. run the debug configuration.
4. Debugging should start directly instead of displaying the terminal with the command pasted in there.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/os.ts
+++ b/packages/core/src/common/os.ts
@@ -29,6 +29,8 @@ function is(userAgent: string, platform: string): boolean {
 export const isWindows = is('Windows', 'win32');
 export const isOSX = is('Mac', 'darwin');
 
+export const EOL = isWindows ? '\r\n' : '\n';
+
 export type CMD = [string, string[]];
 export function cmd(command: string, ...args: string[]): CMD {
     return [
@@ -65,7 +67,8 @@ export namespace OS {
     export const backend = {
         type,
         isWindows,
-        isOSX
+        isOSX,
+        EOL
     };
 
 }
--- a/packages/terminal/src/browser/terminal-widget-impl.ts
+++ b/packages/terminal/src/browser/terminal-widget-impl.ts
@@ -706,7 +706,7 @@ export class TerminalWidgetImpl extends TerminalWidget implements StatefulWidget
     }
 
     async executeCommand(commandOptions: CommandLineOptions): Promise<void> {
-        this.sendText(this.shellCommandBuilder.buildCommand(await this.processInfo, commandOptions) + '\n');
+        this.sendText(this.shellCommandBuilder.buildCommand(await this.processInfo, commandOptions) + (OS.backend.isWindows ? '\r\n' : '\n'));
     }
 
     scrollLineUp(): void {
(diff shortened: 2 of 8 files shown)


packages/terminal/src/browser/terminal-widget-impl.ts
```diff
@@ -706,7 +706,7 @@ export class TerminalWidgetImpl extends TerminalWidget implements StatefulWidget
     }
 
     async executeCommand(commandOptions: CommandLineOptions): Promise<void> {
-        this.sendText(this.shellCommandBuilder.buildCommand(await this.processInfo, commandOptions) + '\n');
+        this.sendText(this.shellCommandBuilder.buildCommand(await this.processInfo, commandOptions) + (OS.backend.isWindows ? '\r\n' : '\n'));
```
msujew, line 709: If I remember correctly, we use this pattern often throughout the application. It might make sense to add `OS.EOL` and `OS.backend.EOL` fields that can be used instead of the ternary operator. That should reduce the probability of these kind of errors.

msujew (commented):
The fix looks good, though I would like to see a minor refactoring regarding the use of EOL tokens, see below.

packages/terminal/src/browser/terminal-widget-impl.ts
jonah-iden: sounds like a good idea

packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -34,11 +34,12 @@ import { WorkspaceCompareHandler } from './workspace-compare-handler';
 import { FileDownloadCommands } from '@theia/filesystem/lib/browser/download/file-download-command-contribution';
 import { FileSystemCommands } from '@theia/filesystem/lib/browser/filesystem-frontend-contribution';
 import { WorkspaceInputDialog } from './workspace-input-dialog';
-import { Emitter, Event, isWindows, OS } from '@theia/core/lib/common';
+import { Emitter, Event, OS } from '@theia/core/lib/common';
 import { FileService } from '@theia/filesystem/lib/browser/file-service';
 import { FileStat } from '@theia/filesystem/lib/common/files';
 import { nls } from '@theia/core/lib/common/nls';
 import { ClipboardService } from '@theia/core/lib/browser/clipboard-service';
+import { EOL } from 'os';
```
msujew, line 42: You cannot import from `os` (Node.js dependency)

jonah-iden: ups. Good catch. used the wrong import

packages/core/src/common/os.ts
```diff
@@ -65,7 +67,8 @@ export namespace OS {
     export const backend = {
         type,
         isWindows,
-        isOSX
+        isOSX,
+        EOL
```
msujew, line 71: This property needs to be overriden [here](https://github.com/eclipse-theia/theia/blob/fbb02df763b115c12e2d68babf33aac3a781b449/packages/core/src/browser/preloader.ts#L50-L58) as well. Otherwise it will stay as the frontend EOL, which isn't the actual backend EOL.

msujew (approved):
Looks good to me, thanks 👍

PR #12604, opened 2023-06-09
ui/ux: styling vsx and source control input fields

vladarama:
#### What it does

Closes: #10855 
This PR provides better styling for various input fields when the view is limited. This includes the vsx-registry and the source control views. These changes allow the placeholder overflow to be `hidden` and replaced by an `ellipsis` to make the styling cleaner and more consistent.

Demo:
VSX Styling Demo

#### How to test

- Start Theia, open the `Extensions` or the `Source Control` views
- Resize the view, making it smaller
- The input field should be responsive, with the text overflow hidden by an ellipsis

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #12622, opened 2023-06-14
ui/ux: fix bug induced by PR #12604

vladarama:
#### What it does

This PR fixes a bug that was induced in #12604 where the styling for input fields would not be displayed properly. An example would be the input field appearing when renaming a file. This was caused by `overflow: hidden` being applied on all text-inputs. Now, the `overflow: hidden` will only be applied to the `vsx-extension-search-bar` which requires it.

Before:
image

After:
image

#### How to test

1. Start Theia, open the `Extensions` or the `Source Control` views
2. Resize the view, making it smaller
3. The input field should be responsive, with the text overflow hidden by an ellipsis
4. Try renaming a file, the input field should be styled correctly

#### Review checklist

#### Reminder for reviewers


vince-fugnitto (approved):
I can confirm the bug on master, and that this change successfully fixes it 👍

PR #12585, opened 2023-05-31
ui/ux: styling input validation

vladarama:
#### What it does

This PR fixes #12243 by providing visual feedback for input validation that is consistent with the input validation in the `search-in-workspace` and the `preferences` views. It adds this input validation for the `SingleTextInputDialog` to create better feedback for the user.

#### How to test

1. Try causing an error when renaming a file
2. Try causing an error when creating a file from the workspace using `New File`
3. Try causing an error when creating a new folder from the workspace using `New Folder`
4. Try causing an error when editing key bindings using `File` -> `Preferences` -> `Keyboard Shortcuts`
5. The error styling should be consistent with the one in `Preferences` and `search-in-workspace`

Preferences Error Validation: 
MicrosoftTeams-image

Easy errors to test include: creating a file/folder with no name or a duplicate name, renaming a file to the same name, creating empty key bindings ...

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #12642, opened 2023-06-21
dialog: adding maxWidth to `SingleTextInputDialog`

vladarama:
#### What it does

This PR adds a maxWidth attribute to `SingleTextInputDialogs`  in order to fix a styling issue that would occur when the input validation error message would get too long. This would cause the dialog to extend indefinitely. This fix adds a max-width of 400 px to the dialogs to make their styling more user-friendly. The modified dialogs include: `New File`, `New Folder`, `Rename` and `EditKeybinding` dialogs.

Before:
no maxwidth

After:
maxwidth

#### How to test

1. Open Theia application
2. Try causing errors while renaming files, creating new files/folders, modifying keybindings
3. The dialog should have a `max-width: 400px` css style
4. Long error messages should not extend the dialog indefinitely

Easy errors to test include: creating a file/folder with no name or a duplicate name, 
renaming a file to the same name, creating empty key bindings ...

Should be merged after #12585 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/dialogs.ts
+++ b/packages/core/src/browser/dialogs.ts
@@ -166,8 +166,10 @@ export abstract class AbstractDialog<T> extends BaseWidget {
         container.classList.add('dialogBlock');
         if (props.maxWidth === undefined) {
             container.setAttribute('style', 'max-width: none');
-        } else {
+        } else if (props.maxWidth < 400) {
             container.setAttribute('style', `max-width: ${props.maxWidth}px; min-width: 0px`);
+        } else {
+            container.setAttribute('style', `max-width: ${props.maxWidth}px`);
         }
         this.node.appendChild(container);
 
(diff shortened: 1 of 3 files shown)


packages/core/src/browser/dialogs.ts
```diff
@@ -166,8 +166,10 @@ export abstract class AbstractDialog<T> extends BaseWidget {
         container.classList.add('dialogBlock');
         if (props.maxWidth === undefined) {
             container.setAttribute('style', 'max-width: none');
-        } else {
+        } else if (props.maxWidth < 400) {
             container.setAttribute('style', `max-width: ${props.maxWidth}px; min-width: 0px`);
+        } else {
+            container.setAttribute('style', `max-width: ${props.maxWidth}px`);
```
msujew, line 172: Question: Why would we want to remove the `min-width` style in case `max-width` is `>= 400`?

vladarama: The dialogs already have a minimum-width of 400 px from here:
From my understanding, the only case where we would need to override this `min-width: 400px` and set it to `0px` would be when we have a `max-width < 400`. In the other cases, we should keep the value from `.dialogBlock` and avoid overriding it which is why I removed the `min-width` style. Please tell me if I am wrong.
https://github.com/eclipse-theia/theia/blob/803efc994ef71352d36d4dfdc6fb87183cb9b6b7/packages/core/src/browser/style/dialog.css#L32-L40

msujew: I see, I didn't know about that. Makes sense now 👍

msujew (approved):
Alright, looks good to me 👍

PR #12648, opened 2023-06-22
Add preference for visualPreview on hover

### LINKED ISSUE #12646: Provide visual preview for widgets
### Feature Description:
With #12350 an enhanced preview option was introduced that displays the title and the caption of a view, when hovering over a tab. To get even more information about the view as a user a visual preview of the current state of the view could be displayed.

To not break the behavior introduced in #12350, the `window.tabbar.enhancedPreview` setting could be extended to be an enum preference with three options:
- classic: Display a simple preview about the view, containing the name.
- enhanced: Display an enhanced preview containing the name and a caption.
- visual: Display the enhanced preview together with a visual preview of the view.

Furthermore, the preview should not be displayed on active views, because the information is already visible on the screen.

sgraband:
#### What it does

Made window.tabbar.enhancedPreview an enum preference with these options:
- classic: a simple unstyled preview, containing the name
- enhanced: a styled preview containing title and caption (#12350)
- visual: the enhanced preview +  visual preview of the view Extended the hover service so that it supports the visual preview. 

Added the `PreviewableWidget` interface.
Widgets, implementing this interface, can be previewed once they were loaded. The loaded flag is set to true when a widget is set to active. Widgets implementing the interface can specify how the preview should look like. The default is simply the node of the widget.
Webviews are currently not previewable, as they raise some challenges. For example, scripts in the webviews would need to be handled/blocked. Therefore, an approach for webviews should be tackled in a follow up.

Fixes #12646

#### How to test

1. Check out this PR an open Theia
2. Set the `window.tabbar.enhancedPreview` preference to `visual`
3. Observer, that on horizontal tab bars when hovering a tab of a view, that is currently not visible, but was loaded before, a visual preview is rendered in the hover feedback.
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/tab-bars.ts
+++ b/packages/core/src/browser/shell/tab-bars.ts
@@ -535,11 +535,14 @@ export class TabBarRenderer extends TabBar.Renderer {
                         // Use the current visible editor as a fallback if not available
                         const height: number = visualPreview.style.height === '' ? this.tabBar.currentTitle!.owner.node.offsetHeight : parseFloat(visualPreview.style.height);
                         const width: number = visualPreview.style.width === '' ? this.tabBar.currentTitle!.owner.node.offsetWidth : parseFloat(visualPreview.style.width);
+                        const desiredRatio = 9 / 16;
+                        const desiredHeight = desiredWidth * desiredRatio;
                         const ratio = height / width;
                         visualPreviewDiv.style.width = `${desiredWidth}px`;
-                        visualPreviewDiv.style.height = `${desiredWidth * ratio}px`;
+                        visualPreviewDiv.style.height = `${desiredHeight}px`;
 
-                        const scale = desiredWidth / width;
+                        // If the view is wider than the desiredRatio scale the width and crop the height. If the view is longer its the other way around.
+                        const scale = ratio < desiredRatio ? (desiredHeight / height) : (desiredWidth / width);
                         visualPreview.style.transform = `scale(${scale},${scale})`;
                         visualPreview.style.removeProperty('top');
                         visualPreview.style.removeProperty('left');
--- a/packages/core/src/browser/shell/tab-bars.ts
+++ b/packages/core/src/browser/shell/tab-bars.ts
@@ -37,6 +37,7 @@ import { HoverService } from '../hover-service';
 import { Root, createRoot } from 'react-dom/client';
 import { SelectComponent } from '../widgets/select-component';
 import { createElement } from 'react';
+import { PreviewableWidget } from '../widgets/previewable-widget';
 
 /** The class name added to hidden content nodes, which are required to render vertical side bars. */
 const HIDDEN_CONTENT_CLASS = 'theia-TabBar-hidden-content';
@@ -165,7 +166,9 @@ export class TabBarRenderer extends TabBar.Renderer {
             ? nls.localizeByDefault('Unpin')
             : nls.localizeByDefault('Close');
 
-        const hover = this.tabBar && (this.tabBar.orientation === 'horizontal' && !this.corePreferences?.['window.tabbar.enhancedPreview']) ? { title: title.caption } : {
+        const hover = this.tabBar && (this.tabBar.orientation === 'horizontal' && this.corePreferences?.['window.tabbar.enhancedPreview'] === 'classic')
+            ? { title: title.caption }
+            : {
             onmouseenter: this.handleMouseEnterEvent
         };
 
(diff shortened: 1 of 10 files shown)


sgraband:
@vince-fugnitto Could you review this? Since you reviewed the previous PR (#12350)?

(I will resolve the Changelog conflict together with any review feedback ;))

packages/core/src/browser/widgets/previewable-widget.ts
```diff
@@ -0,0 +1,30 @@
...
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+import { isFunction, isObject } from '../../common';
```
vince-fugnitto, line 16: minor: please include a newline.

vince-fugnitto (commented):
Thank you for the contribution, I've noticed a couple of things:

- the feature does not work well with different layouts, and often hides the visual hover
- the styling of the visual preview can be enhanced to have a more fixed width so it is consistent for different tabs and for different nested files (at the moment the size is quite big and hides a lot of information as opposed to the ones present in the browser, and depending where a file is located will differ in size giving an odd user-experience)

image

image

sgraband:
@vince-fugnitto I added three commits:
[ea5397a](https://github.com/eclipse-theia/theia/pull/12648/commits/ea5397aadd9d180d3e62878f25cfaa6ab1ad8314) addresses your minor feedback in the comments.
[39be6ce](https://github.com/eclipse-theia/theia/pull/12648/commits/39be6ce05247aa49e96ab97fd91f4f0ca3f1bf11) fixes the aspect ratio of the preview windows to 16:9. This should help to unify the hover experience between views.
[b30b06f](https://github.com/eclipse-theia/theia/pull/12648/commits/b30b06f70867da15b6b8237804f4f847aa21f0bf) fixes a small bug with the title and caption so that they acknowledge the max-width of the hover.

I am not sure, if the width problem should (or needs to) be solved with this PR. The way it is now, adaptors can set the [`--theia-hover-max-width`](https://github.com/eclipse-theia/theia/blob/dab0aade1d0de8a1394975e249db636070b4b9b8/packages/core/src/browser/style/hover-service.css#L20) css variable to the maximum width they like for the hover. The visual preview takes this max-width into account. And with the latest commit, as mentioned before, the caption and title will also acknowledge the max-width. So this is already customizable, but maybe could be improved in the future via a preference for example.

Could you take a look at the changes and tell me what you think?

sgraband:
@vince-fugnitto would you have time to take another look this week? I will be out of office next week, so it would be great to have some feedback this week. TIA

packages/core/src/browser/style/hover-service.css
```diff
@@ -17,7 +17,7 @@
 /* Adapted from https://github.com/microsoft/vscode/blob/7d9b1c37f8e5ae3772782ba3b09d827eb3fdd833/src/vs/workbench/services/hover/browser/hoverService.ts */
 
 :root {
-  --theia-hover-max-width: 500px;
+  --theia-hover-width: 300px;
```
vince-fugnitto, line 20: I believe this is another breaking change that adopters might depend on?
I don't think there is harm in keeping both variables (one for width and one for max-width).

packages/core/src/browser/style/hover-service.css
```diff
@@ -27,7 +27,7 @@
   background-color: var(--theia-editorHoverWidget-background);
   border: 1px solid var(--theia-editorHoverWidget-border);
   padding: var(--theia-ui-padding);
-  max-width: var(--theia-hover-max-width);
+  width: var(--theia-hover-width);
```
vince-fugnitto, line 30: This change actually breaks other hovers we have in the framework (ex: statusbar).
Can we instead only style the enhanced preview with the proper width and not update the more generic `theia-hover`:


https://github.com/eclipse-theia/theia/assets/40359487/bd82c119-3285-426f-b061-43ca94dd37c3

sgraband: Good catch! I reworked the last commit, so that both variables are there, but the `--theia-hover-preview-width` variable is only used by the preview hovers.

packages/core/src/browser/shell/tab-bars.ts
```diff
@@ -510,6 +513,58 @@ export class TabBarRenderer extends TabBar.Renderer {
...
+        const widget = title.owner;
+        // Check that the widget is not currently shown, is a PreviewableWidget and it was already loaded before
+        if (this.tabBar && this.tabBar.currentTitle !== title && PreviewableWidget.isPreviewable(widget)) {
+            const html = document.getElementById(widget.id);
+            if (html) {
+                const previewNode: Node | undefined = widget.getPreviewNode();
+                if (previewNode) {
+                    const clonedNode = previewNode.cloneNode(true);
+                    const visualPreviewDiv = document.createElement('div');
+                    visualPreviewDiv.classList.add('enhanced-preview-div');
```
msujew, line 526: Minor: Given that this is already a `div`, I think we should use a less technical name for the css class.
```suggestion
                    visualPreviewDiv.classList.add('enhanced-preview-container');
```

msujew (approved):
Seems to work pretty well. I noticed some unexpected behavior when the widget has been stretched across the screen, see [here](https://github.com/eclipse-theia/theia/assets/4377073/229d0e3a-7fb5-448e-bdc3-8df40ce2430b). However, I believe that's due to the general approach of re-rendering the widget in another HTML element, and cannot be easily addressed. It might also just be me, I've been hoping that the preview shows more than what is rendered in the widget itself.

Anyway, aside from that I just have a minor comment. Otherwise this looks pretty good 👍

PR #12658, opened 2023-06-28
cli: use `includeAllVersions` when resolving packs

vince-fugnitto:
#### What it does

The commit re-introduces the `includeAllVersions` flag when querying for extension-packs in order to find a compatible version of a declared plugin. Before the change we were not properly iterating the versions of a plugin which resulted in a plugin failing to download if the latest version was not compatible to our version range (ex: `ms-vscode.js-debug`). 

The change means that we we will properly iterate the versions of a plugin until we hit a version which is compatible with the framework (with respect to our declared supported API range).

#### How to test

1. confirm that CI for the `yarn download:plugins` script is successful

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/cli/src/download-plugins.ts
+++ b/dev-packages/cli/src/download-plugins.ts
@@ -129,7 +129,7 @@ export default async function downloadPlugins(ovsxClient: OVSXClient, requestSer
             await parallelOrSequence(Array.from(ids, id => async () => {
                 try {
                     await rateLimiter.removeTokens(1);
-                    const { extensions } = await ovsxClient.query({ extensionId: id });
+                    const { extensions } = await ovsxClient.query({ extensionId: id, includeAllVersions: true });
                     const extension = apiFilter.getLatestCompatibleExtension(extensions);
                     const version = extension?.version;
                     const downloadUrl = extension?.files.download;
--- a/packages/vsx-registry/src/browser/vsx-extensions-model.ts
+++ b/packages/vsx-registry/src/browser/vsx-extensions-model.ts
@@ -315,7 +315,7 @@ export class VSXExtensionsModel {
             const client = await this.clientProvider();
             let data: VSXExtensionRaw | undefined;
             if (version === undefined) {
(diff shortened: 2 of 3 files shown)


paul-marechal (approved):
LGTM, thanks!

vince-fugnitto:
@msujew thank you for catching it, I managed to fix the following three bugs:
- the resolution of plugins in packs compared to out supported api version
- the download of plugins at runtime
- the ability to open an extension information (vsx-editor) from the extensions-view

msujew (commented):
Works better, but installing the current `German Language Pack` from the vsx extension view still fails for some reason. It does work through the `Configure Display Language` command though, so I'm not sure what's wrong there.

vince-fugnitto:
> Can you confirm Vince?

@msujew I can confirm as well, the query we use https://open-vsx.org/api/-/query?extensionId=ms-ceintl.vscode-language-pack-de&includeAllVersions=true does not seem to produce any results oddly enough (even when I tried on their [swagger ui](https://open-vsx.org/swagger-ui/index.html#/registry-api/getQuery) page.

msujew (approved):
> I can confirm as well, the query we use https://open-vsx.org/api/-/query?extensionId=ms-ceintl.vscode-language-pack-de&includeAllVersions=true does not seem to produce any results oddly enough

I see. In that case we're not really responsible for the issue, as it appears to be a regression from open-vsx. The main issue that this change covers is fixed anyway, so I'll approve as well.

vince-fugnitto:
@msujew actually, it might just be the fact the open-vsx is case-sensitive:

- https://open-vsx.org/api/-/query?extensionId=MS-CEINTL.vscode-language-pack-de&includeAllVersions=true

The following would produce results.

msujew:
@vince-fugnitto Oh well, that seems to be new... Any chance you can quickly get the case sensitivity issue fixed in this PR as well?

msujew:
@vince-fugnitto Sounds good. Though it makes sense to merge this fix now to get it into today's release.

PR #12677, opened 2023-07-03
DRAFT: collaboration concept

msujew:
#### What it does

DO NOT MERGE THIS!

This PR provides a platform for discussion on the proposed collaboration feature. It will be followed up by a PR that actually implements this concept.

Please comment if you have any critique or questions with the approach outlined in the document. I'm happy to discuss, revise and extend the concept.

### DIFF
--- a/collaboration-concept.md
+++ b/collaboration-concept.md
@@ -1,0 +1,103 @@
+# Collaboration Concept
+
+## Microsoft VSCode Live Share Extension
+
+The official Microsoft [Live Share Extension](https://visualstudio.microsoft.com/services/live-share/) implements collaboration support for the following features:
+
+- Editor Syncing (Text + Client markers for each user)
+- Host Syncing (i.e. other users follow the host into their currently selected file + position)
+- Terminal Syncing (if enabled). Both input+output is synced
+- Debug Syncing (if enabled). Both users and host can step through the code as if it were on one machine. Breakpoints are synced as well.
+- Plugin Syncing (i.e. external user has access to all installed plugins/extensions of the host user)
+- Chat Functionality
+- Mechanism for making the API available to other extensions, see [vsls](https://www.npmjs.com/package/vsls)
+
+Similar to the tunnel feature, the connection between host and other users is accomplished using hole-punching.
+I.e. a single central server serves as an intermediary to exchange messages. No messages are being exchanged directly between users. 
+
+## Requirements for Theia Collaboration Feature
+
+### Technical Requirements
+
+The protocol/implementation that this new live share feature is supposed to be using therefore has a few different tasks for different widgets/features
+
+1. Document syncing akin to yjs monaco binding + Awareness information exchange (i.e. broadcast information to all users)
+2. Virtual File System coming from the host user
+3. Frontend message proxying for debug sync (i.e. debug launch or monaco language service requests need to be proxied to the plugin host and the results send back)
+4. Terminal input + output syncing requires a yjs like approach for input and a broadcast for output
+5. Broadcast required for Chat functionality
+
+### Messaging System Architecture
+
+For the technical requirements above, we effectively need a two different kind of message types:
+
+1. Broadcasting actions to all users:
+    1. Document entry events (yjs shared data types sync; monaco editor + terminal input)
+    2. Awareness information broadcast, i.e. current cursor position in monaco/custom editor
+    3. Chat functionality
+    4. Output of host terminal
+2. Proxy message to host (with result or without):
+    1. Debug launch (notification)
+    2. File System information
+    3. Monaco language services (i.e. LSP requests)
+
+The following diagram shows the Session creation/entry negotiations. In all cases where the User/Host communicate directly in the diagram, the intermediary server provides the messaging exchange infrastructure (using a hole-punching mechanism). The host is always the user that provides the workspace
+
+```mermaid
+sequenceDiagram
+    participant Host
+    participant I as Intermediary
+    participant User
+    Host->>I: Create Collab Session
+    I->>Host: Return with Session ID
+    Host-->>User: Sends Session ID via chat
+    User->>Host: Request Session Entry
+    Host->>User: Allow Session Entry (with permissions)
+    User->>Host: Download Workspace Info
+```
+
+Diagram outlining the broadcasting mechanism. We skip the Intermediary here, since it only proxies messages between users. Also note that we don’t need a host, since in this message scenario, all users have “equal rights”. This example uses the notification for file changes, but can be used for broadcasts for debugging or terminal output
+
+```mermaid
+sequenceDiagram
+    participant U1 as User A
+    participant U2 as User B
+    participant U3 as User C
+  
+    Note over U1: Modifies a file
+	
+    par U1 to U2
+        U1-)U2: Send file notification event
+    and U1 to U2
+        U1-)U3: Send file notification event
+    end
+    par 
+        Note over U2: Updates modified file
+        Note over U3: Updates modified file
+    end 
+```
+
+Proxy host messaging example:
+
+```mermaid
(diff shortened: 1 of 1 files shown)


collaboration-concept.md
```diff
@@ -0,0 +1,137 @@
+# Collaboration Concept
+
```
sdirix, line 2: I would start with a short motivational and overview, e.g.

"Collaboration is an often asked for feature in Theia. In this document we analyze the collaboration feature of VS Code, specify technical requirements and architecture for Theia and give an overview over the API concept."

collaboration-concept.md
```diff
@@ -0,0 +1,137 @@
...
+- Editor Syncing (Text + Client markers for each user)
+- Host Syncing (i.e. other users follow the host into their currently selected file + position)
+- Terminal Syncing (if enabled). Both input+output is synced
+- Debug Syncing (if enabled). Both users and host can step through the code as if it were on one machine. Breakpoints are synced as well.
+- Plugin Syncing (i.e. external user has access to all installed plugins/extensions of the host user)
+- Chat Functionality
+- Mechanism for making the API available to other extensions, see [vsls](https://www.npmjs.com/package/vsls)
+
+Similar to the tunnel feature, the connection between host and other users is accomplished using hole-punching.
+I.e. a single central server serves as an intermediary to exchange messages. No messages are being exchanged directly between users.
```
sdirix, line 16: According to the docs they do send messages directly between host and users if the network allows it, see https://learn.microsoft.com/en-us/visualstudio/liveshare/reference/connectivity#connection-modes. However every connection is authenticated via the cloud, so it's never fully private.

collaboration-concept.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Technical Requirements
+
+The protocol/implementation that this new live share feature is supposed to be using therefore has a few different tasks for different widgets/features
+
+1. Document syncing akin to yjs monaco binding + Awareness information exchange (i.e. broadcast information to all users)
+2. Virtual File System coming from the host user
+3. Frontend message proxying for debug sync (i.e. debug launch or monaco language service requests need to be proxied to the plugin host and the results send back)
+4. Terminal input + output syncing requires a yjs like approach for input and a broadcast for output
+5. Broadcast required for Chat functionality
```
sdirix, line 28: Should we also list other potential requirements like:
 - privacy (e.g. allow to configure an own message server without going through one central public instance)
 - encryption
 - authentication

sdirix (commented):
Thanks for the creation of this document! It gives a good first overview!

collaboration-concept.md
```diff
@@ -0,0 +1,137 @@
+# Collaboration Concept
+
+## Microsoft VSCode Live Share Extension
```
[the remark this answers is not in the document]
msujew: As far as I know, APIs cannot be licensed, only the implementation. If we were to offer a drop-in replacement for Microsoft's `vsls` that uses our custom collaboration feature implementation, we shouldn't infringe on any license agreements (fair use). See also [Oracle v. Google](https://en.wikipedia.org/wiki/Google_LLC_v._Oracle_America,_Inc.).

msujew:
> Did I get that correctly that your example for the API concept is code that is expected to live in a Theia frontend extension?

Yes.

> any frontend extension can listen to and send requests, notifications, and broadcast messages

Depending on permissions (which I only mentioned in [here](https://github.com/eclipse-theia/theia/blob/93e9e802c37846d6232e0a76b6c6c7bb09403726/collaboration-concept.md?plain=1#L69)). The frontend should reject any impermissible action.

> the frontend extension can then decide whether to directly process those messages or whether to forward this to the plugin host via main/ext communication?

Right 👍 

> there is a generic forwarder to the plugin host to support VS Code extensions that use the vsls API? If yes, how do we determine when to forward and when we don't, or do we forward all messages and the vsls API clients would decide whether to react?

Good question, haven't put too much thought into that yet. First thought would be to have some sort of message wrapping, but not too sure. This is probably one of the last features that I plan to implement anyway.

collaboration-concept.md
```diff
@@ -0,0 +1,151 @@
...
+- Terminal Syncing (if enabled). Both input+output is synced
+- Debug Syncing (if enabled). Both users and host can step through the code as if it were on one machine. Breakpoints are synced as well.
+- Plugin Syncing (i.e. external user has access to all installed plugins/extensions of the host user)
+- Chat Functionality
+- Mechanism for making the API available to other extensions, see [vsls](https://www.npmjs.com/package/vsls)
+
+Similar to the tunnel feature, the connection between host and other users can be accomplished using hole-punching.
+I.e. a single central server serves as an intermediary to exchange messages. No messages are being exchanged directly between users.
+If possible - for example, if the users are on the same network - the extension will try to connect directly to the host,
+see [here](https://learn.microsoft.com/en-us/visualstudio/liveshare/reference/connectivity#connection-modes).
```
paul-marechal, line 21: I don't understand the "i.e." following the mention of "hole punching" as this term means establishing a direct connection through each peer's NAT by "punching holes" with the help of an intermediary server to know where to hit. I've also successfully managed to do hole punching in the past using UDP while TCP was more finicky (still worked). The implication means we'd need a UDP transport for direct connections? But that's only if you truly want to do hole punching over the internet.

collaboration-concept.md
```diff
@@ -0,0 +1,151 @@
...
+
+For the messages architecture above, we need a coordination or messaging server. The server serves exactly two purposes:
+
+1. Opening up the host system to the world. The host and the coordination server negotiate a new collab session. This can be as simple as an HTTP call to this server. The coordination server then creates a virtual room and assigns the host that originated the call as its host.
+2. Communication between collaborators within a session. For example, a broadcast notification needs to be send to the coordination server first, which then relays the message to all other participants.
+
+Assuming the server is built specifically for Theia, we can reuse some of Theia’s dependencies for this, i.e. socket-io for bidirectional, high-performance messaging on top of HTTP.
+
+
+## API Concept
```
paul-marechal, line 120: What kind of API are we talking about here? Theia Extension API? VS Code Extension API?

Should we inject the `vsls` extension the same way we do with the `vscode` API so that extensions relying on this package can work in Theia?

Should we take care to properly implement the groundwork feature in the framework before exposing it through the plugin API? (By contrast with webviews which are baked-in the `plugin-ext` package at the moment.)

collaboration-concept.md
msujew: Hole punching might not be the correct term here, right. In the end, all that we need to accomplish is that users can communicate with each other through some means, be it some intermediary server or whatever.

paul-marechal: Fine if your goal is to abstract the API from how the communication might happen. Although properly laying down the expected modes of communication would help in designing said API: P2P vs Server-centric, how to handle auth/identity/permissions, etc...

If the goal is to be highly generic then the act of creating sessions must be defined at least partially by the embedding layer for the target IDE:
1. Define the communication implementation (a.k.a "driver"? i.e. connect to some server and request a new room)
2. Said driver may implement auth (i.e. prompt the user for credentials in the IDE? defer to some OAuth service?)

collaboration-concept.md
paul-marechal: Then is it fair to assume that this library must support at least what `vsls` supports? That would make for a good starting point for designing the API.

msujew:
For anyone interested: I'm currently working on a PoC for this, the communication infrastructure is working as designed in this document and I've got workspace sharing and editor sync already in a working state:


https://github.com/eclipse-theia/theia/assets/4377073/f3d84326-9994-49a5-a3d2-19cefe5109b5

msujew:
Thanks everyone for the valuable feedback. This has been superseeded by https://github.com/eclipse-theia/theia/pull/13309.

PR #12690, opened 2023-07-10
core: enable portable-mode

vladarama:
## Portable Mode in Theia

Creating a `data` folder at the root of a Theia application will make it portable. So all of the app data and user data will get written to the `data` folder instead of their default locations.

The `.theia`folder is moved from the user’s home directory (`%userprofile%` or `~/`) to a folder called `user-data` inside the `data` folder. `user-data` contains all of the settings, preferences and extensions.

In addition, all of the application data which is stored by default inside a `Theia App Name (Theia Electron Example)`folder in (`%appdata%` or `~/.config` ) is moved to a folder called `app-data` inside the `data` folder. `app-data` contains all of the application’s cache and layout information.

Please refer to the following PR to further understand how the Portable mode is implemented in Theia Blueprint:
https://github.com/eclipse-theia/theia-blueprint/pull/276

#### How to test

The following changes should make the `user-data` portable (settings, preferences, recent workspace ...).
It also makes the `app-data` portable so all of the electron cache and layout information will be stored in `data/app-data` instead of the default location.

You can test that the portable mode is functional using the Electron Example Application. 

- Run `yarn` and `yarn electron build` 
- Create a `data` folder inside `/theia/examples/electron`. 
- Run `yarn electron start`.

All of the `user-data` which is stored by default inside the user's directory will be stored inside of `/theia/examples/electron/data/user-data`. In addition, all of the `app-data` which is stored by default in `~/.config/Theia Electron Example or %appdata%/Theia Electron Example` will be stored inside of `/theia/examples/electron/data/app-data`.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-main/electron-main-application.ts
+++ b/packages/core/src/electron-main/electron-main-application.ts
@@ -19,7 +19,8 @@ import { screen, app, BrowserWindow, WebContents, Event as ElectronEvent, Browse
 import * as path from 'path';
 import { Argv } from 'yargs';
 import { AddressInfo } from 'net';
-import { promises as fs, existsSync, mkdirSync } from 'fs';
+import { promises as fs } from 'fs';
+import { pathExists, mkdir } from 'fs-extra';
 import { fork, ForkOptions } from 'child_process';
 import { DefaultTheme, FrontendApplicationConfig } from '@theia/application-package/lib/application-props';
 import URI from '../common/uri';
@@ -171,11 +172,11 @@ export class ElectronMainApplication {
     @inject(TheiaElectronWindowFactory)
     protected readonly windowFactory: TheiaElectronWindowFactory;
 
-    protected isPortable = this.makePortable();
+    protected isPortable: boolean;
 
-    protected readonly electronStore = new Storage<{
+    protected electronStore: Storage<{
         windowstate?: TheiaBrowserWindowOptions
-    }>();
+    }>;
 
     protected readonly _backendPort = new Deferred<number>();
     readonly backendPort = this._backendPort.promise;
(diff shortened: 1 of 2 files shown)


vladarama (commented):
@msujew 
Thank you for the review. I have changed all of the fs methods to their asynchronous version which makes more sense. However, I am running into a few issues I would like your thoughts on. Even after calling `await this.makePortable()` at the beginning of the start method in `electron-main-application.ts`, the `Local State` file gets written to the `appdata` under `Theia Electron Example` instead of being written to the `data` folder. Everything else seems to work fine, so I assume we are running into a race condition where `Theia Electron Example` gets created and written to before making the electron application portable.  I will keep digging into it, but do you have any idea why this might be happening ?

msujew:
@vladarama I see, I wasn't aware of that. I assume that the `electron.app` fires it's `ready` event while we await the `fs.pathExists` results, which then automatically starts writing stuff. Using sync calls of course prevents this event from firing until we're done.

I'm fine with changing the `isPortable` stuff back to sync functions. I was mostly concerned about the non-cached sync calls in `env-variables-server.ts`. This could've possibly led to performance regressions.

JonasHelming:
@vladarama This is ready for a re-review, correct?

vladarama:
@JonasHelming 
Yes, this PR should be ready for another review.

msujew (approved):
Thanks looks good to me 👍

XeonG:
Does this even work.. .it's a bit of mission to even try it..

why if if the 'data' folder is created, does it then need me to go messing about copy other folders which in turn require the user to go and install thiea and then run the app to create those default file folders.. etc, its all madness..   

should just be an installer with a portable checkbox and then anywhere that is installed, is a portable version, all self contained.

I don't mind copy files and folders from an existing setup if I wanted the replicate the same setup, it in portable manner but this is crazy.. and it doesn't even work from what I've tried...


running this with that 'data' folder and copy the mentioned existing files and folders.. and it was still just saving/loading settings from the default user and not to the data folder I created.
explorer_U9OakO8Auk
related ticket https://github.com/eclipse-theia/theia/discussions/14508#discussioncomment-11376564

PR #12691, opened 2023-07-10
terminal: bump `xterm` and addons

vince-fugnitto:
#### What it does

The pull-request bumps `xterm` and related addons to their latest versions to benefit from additional improvements and bug fixes.

#### How to test

1. start the example application(s)
2. spawn a terminal - confirm it works as expected
3. confirm that resizing and searching within the terminal works as expected
4. confirm that `split terminal` works as expected
5. confirm that updating terminal related preferences takes effect

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -232,6 +232,8 @@
 
 - [deps] bumped supported Node.js version from 16.x to >=18, you may need to update your environments [#12711](https://github.com/eclipse-theia/theia/pull/12711)
 - [preferences] removed the `welcome.alwaysShowWelcomePage` preference in favor of `workbench.startupEditor` [#12813](https://github.com/eclipse-theia/theia/pull/12813)
+- [terminal] deprecated `terminal.integrated.rendererType` preference [#12691](https://github.com/eclipse-theia/theia/pull/12691)
+- [terminal] removed protected method `TerminalWidgetImpl.getTerminalRendererType` [#12691](https://github.com/eclipse-theia/theia/pull/12691)
 
 ## v1.40.0 - 07/27/2023
 
--- a/packages/terminal/package.json
+++ b/packages/terminal/package.json
@@ -9,9 +9,9 @@
     "@theia/process": "1.46.0",
     "@theia/variable-resolver": "1.46.0",
     "@theia/workspace": "1.46.0",
-    "xterm": "^4.16.0",
-    "xterm-addon-fit": "^0.5.0",
-    "xterm-addon-search": "^0.8.2"
+    "xterm": "^5.2.1",
+    "xterm-addon-fit": "^0.7.0",
+    "xterm-addon-search": "^0.12.0"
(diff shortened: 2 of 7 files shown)


msujew (commented):
The change works well on Chromium based browsers, but breaks on FireFox, since we do a bit of monkey-patching:

https://github.com/eclipse-theia/theia/blob/abc45782eb7f9b132a2680eec3df540e40fac74a/packages/terminal/src/browser/terminal-widget-impl.ts#L644-L661

`_onIntersectionChange` doesn't seem to be a method in the `renderService` anymore.

msujew:
I was able to get it to work again by replacing the existing code with this:

```ts
const renderService: any = (this.term as any)._core._renderService;
const originalFunc: (entry: IntersectionObserverEntry) => void = renderService._handleIntersectionChange.bind(renderService);
const replacement = function (entry: IntersectionObserverEntry): void {
    if (entry.target.ownerDocument !== document) {
        // in Firefox, the intersection observer always reports the widget as non-intersecting if the dom element
        // is in a different document from when the IntersectionObserver started observing. Since we know
        // that the widget is always "visible" when in a secondary window, so we refresh the rows ourselves
        renderService._pausedResizeTask.flush();
        renderService.refreshRows(0, renderService._rowCount - 1);
    } else {
        originalFunc(entry);
    }
};

renderService._handleIntersectionChange = replacement;
```

vince-fugnitto:
> I was able to get it to work again by replacing the existing code with this:

@msujew good catch and thank you for the suggestion! I'll add you as a co-author whenever we merge :)

msujew:
> Was there a reason you changed more than just the method name?

Yep, just using the correct method fixed the error in the console, but terminals still experienced a render issue. Any terminal didn't re-render anymore after popping it out to a secondary window.

tsmaeder:
I think this is worth investigating: if you look at the code in RendererService.ts, I think the only time the `_isPaused` flag is set is in the  `_handleIntersectionChange` method, which means `_needsFullRefresh` is never reset to `false`. We're effectively turning off incremental paints, just at a glance. That doesn't sound like the right thing to do.

tsmaeder (dismissed):
Putting a block on this one so nobody accidentally merges it: I believe we have a performance problem with incremental updates after the library update.

tsmaeder:
@msujew can you remember what the exact rendering issue was on firefox? For me, there are rendering issues when I resize the external window, but they are the same on chrome and on firefox. 
@vince-fugnitto mind if I take over this one?

vince-fugnitto:
> @vince-fugnitto mind if I take over this one?

Sure, go ahead!

tsmaeder:
@msujew I fixed the tests. Could you give this one another review?

msujew (approved):
Yep, perfect. Works good now. No sluggishness on Firefox anymore.

PR #12699, opened 2023-07-11
Stop publishing next

paul-marechal:
Related https://github.com/eclipse-theia/theia/discussions/12519


paul-marechal:
We can revert this in the future if it ends up causing trouble. But first we need to try to stop publishing on every commit.

vince-fugnitto (approved):
LGTM :+1: 
The changes make sense and it should help both `npm` (not to publish so many unused versions), and our CI (not to publish `yarn docs` for unreleased versions).

PR #12710, opened 2023-07-13
bugfix/12709 Added localization for some phrases

### LINKED ISSUE #12709: Canceled debug request error is not localized
### Steps to Reproduce:

https://github.com/eclipse-theia/theia/blob/e52fc5b296d5d5d0f586d3fffe94e50fa795afc8/packages/debug/src/browser/debug-session-connection.ts#L156

https://github.com/eclipse-theia/theia/blob/e52fc5b296d5d5d0f586d3fffe94e50fa795afc8/packages/debug/src/browser/debug-session-connection.ts#L200

https://github.com/eclipse-theia/theia/blob/e52fc5b296d5d5d0f586d3fffe94e50fa795afc8/packages/debug/src/browser/debug-session-connection.ts#L208

### Additional Information

- Operating System: Any
- Theia Version: The latest

Zebsterpasha:
#### What it does

Closes #12709 

#### How to test

Just check the keys and add localized phrases)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-session-connection.ts
+++ b/packages/debug/src/browser/debug-session-connection.ts
@@ -153,7 +153,7 @@ export class DebugSessionConnection implements Disposable {
 
     protected checkDisposed(): void {
         if (this.disposed) {
-            throw new Error(nls.localize('theia/debug/isDisposed', 'the debug session connection is disposed, id: {0}', this.sessionId));
+            throw new Error(nls.localize('theia/debug/connectionClosed', 'Debug connection unexpectedly closed'));
         }
     }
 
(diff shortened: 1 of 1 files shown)


Zebsterpasha:
How can I restart the testing? I think, this is not my mistake, it should fix by itself.

msujew (commented):
Sorry for the late feedback.

I'm not sure we should actually localize the error messages - especially since they have quite a lot of technical information that is useless to the actual user (such as a `sessionId` or `requestId`). I would much rather like to see a generic error message related to the failing debug connection rather than just printing the error as it comes from the debug connection.

Zebsterpasha:
> I would much rather like to see a generic error message related to the failing debug connection rather than just printing the error as it comes from the debug connection.

Okay, what text should I use in this situation? I just added localization and didn’t think about the meaning then)

msujew:
In case any errors related to the connection appears, the frontend should probably just print "Debug connection unexpectedly closed".

msujew (commented):
That's not quite what I meant. We should keep the original errors (including original error messages), but print a generic message (i.e. "Debug connection unexpectedly closed") where we catch those errors. That way we still have the option to log the underlying error reason, while still giving the user a reasonable (and translatable) message.

msujew:
@Zebsterpasha Do you intend to continue with this change? I would close the PR in a few days if there's no further comment.

JonasHelming:
@msujew Unfortunately, I think we should close this.

Issue #11220, opened 2022-05-30
future: support Node 18


paul-marechal:
While dropping support for Node 12 in https://github.com/eclipse-theia/theia/pull/11210 I tried adding testing for Node 18 but failed.

It seems like there's a couple breakages in Node 18 that we will need to adapt to once we decide to support it:

- It needs Ubuntu >= 20.04 to run because it uses a newer version of glibc.
- `net.Server.listen(port, address, ...)` now defaults to an IPv6 localhost address when unspecified.
- ~~The `macaddress` npm package used by `plugin-ext` breaks on Node 18.~~
- We need to update the `macaddress` package for it to work on Node 18.
- More?

Alexander-Taran:
pull request to fix `macaddress` package
https://github.com/scravy/node-macaddress/pull/40

Alexander-Taran:
Macaddress published fixes version to nomjs

anavarre:
Any update on Node.js 18 support? It's been a year since this issue was raised and with the upcoming [EOL of Node.js 16 in September](https://nodejs.org/en/blog/announcements/nodejs16-eol) I'd like to make sure that we have a secure option then. Thanks!

shriacquia:
Any update on Node.js 18 support?

paul-marechal:
Gave it a quick try and it seems trivial to bump to node 18.x so I opened a PR for it.

Please try the PR and tell us if you find any regressions that weren't caught by our CI.

vince-fugnitto:
Please note that we generally want to follow the same runtime present in VS Code for compatibility with extensions.
They declare that versions higher than 17 have not yet been tested, and still use 16 during CI:

- https://github.com/microsoft/vscode/blob/406618d71adb41af931ca787e178b797a475eb6c/build/npm/preinstall.js#L12-L18
- https://github.com/microsoft/vscode/blob/406618d71adb41af931ca787e178b797a475eb6c/.github/workflows/ci.yml#L26-L28

paul-marechal:
We use [Electron 23](https://github.com/eclipse-theia/theia/blob/dab0aade1d0de8a1394975e249db636070b4b9b8/packages/electron/package.json#L15) which [embarks Node 18 already](https://www.electronjs.org/blog/electron-23-0).

danepowell:
Since this seems to be blocked on vscode, note that vscode will move to Node 18 as part of Electron 25: https://github.com/microsoft/vscode/issues/184021

[...]

PR #12711, opened 2023-07-13
Move from node `16` to node `>=18`

paul-marechal:
#### What it does

Update our CI/CD to run using Node.js 18.x and 20.x.

Update `macaddress` to include the fix for Node.js 18.x.

Closes https://github.com/eclipse-theia/theia/issues/11220

#### How to test

Everything should run as usual? CI/CD should be green.

#### Review checklist

#### Reminder for reviewers


paul-marechal:
OK there's something weird going on with the CI runners and I won't have much time to troubleshoot this soon.

I don't get any of the errors the Windows workflow has, even when starting from a fresh clone.

The `/bin/sh: 1: theia: not found` errors on Ubuntu/Mac are also odd as they imply that `yarn` doesn't do a good job of exposing npm bins to `package.json` scripts? I'd need someone to try on their own Ubuntu machine to check...

Help is welcome here.

rahulgupta-acquia:
@paul-marechal  Could you please rebase this repo, this repo has some conflicts that needs to fixed for testing.

rahulgupta-acquia:
@paul-marechal could you please suggest a way to test this build in my local ( MAC ) system,
 
I followed the below steps for setup https://github.com/eclipse-theia/theia/blob/master/doc/Developing.md#quick-start 
```
git clone https://github.com/eclipse-theia/theia \
    && cd theia \
    && yarn \
    && yarn download:plugins \
    && yarn browser build \
    && yarn browser start
```

but after cloning and yarn install, it throws the 
```
 $ yarn download:plugins
command not found: theia
```

paul-marechal:
@rahulgupta-acquia well the fact that it doesn't work is the issue I was asking help with... But looking at the logs again I might have figured it out: see latest commit.

rahulgupta-acquia:
> @rahulgupta-acquia well the fact that it doesn't work is the issue I was asking help with... But looking at the logs again I might have figured it out: see latest commit.

yes, i checked `yarn add global node-gyp` is required for further steps.

paul-marechal:
@rahulgupta-acquia you mean that it worked for you when using `yarn global add node-gyp` before doing `yarn install`?

rahulgupta-acquia:
> @rahulgupta-acquia you mean that it worked for you when using `yarn global add node-gyp` before doing `yarn install`?

No, i am facing issue at the time of `yarn download;plugin`.

I think my issue is not related to this PR.

rahulgupta-acquia:
@paul-marechal Do you have any document or steps to test this PR in Local system (MAC)?

paul-marechal:
@rahulgupta-acquia I don't have any more steps than what you found here: https://github.com/eclipse-theia/theia/blob/master/doc/Developing.md#quick-start

rahulgupta-acquia:
I need to build IDE using this PR.

paul-marechal:
@rahulgupta-acquia I'd like for other contributors to check for regressions before proceeding but yes!

kittaakos:
Thanks for the update. Could you please adjust the developer readme? Thanks

https://github.com/eclipse-theia/theia/blob/c8fb5adfa598fd158d4f11783ccb923af449b12f/doc/Developing.md?plain=1#L55

paul-marechal:
IP v6 addresses were being used over IP v4 but I found a way to get back to the previous behaviour using [`dns.setDefaultResultOrder('ipv4first')`](https://nodejs.org/docs/latest-v18.x/api/dns.html#dnssetdefaultresultorderorder)

rahulgupta-acquia (commented):
Reviewed the code changes.

rahulgupta-acquia:
@paul-marechal Could you please rebase this repo. 
Thanks

rahulgupta-acquia:
@paul-marechal code rebase needed. 

@msujew @tsmaeder @vince-fugnitto  Request for review.

.github/workflows/ci-cd.yml
```diff
@@ -51,7 +52,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-2019, ubuntu-latest, macos-11]
-        node: [16.x]
+        node: [18.x, 20.x]
```
vince-fugnitto, line 55: I believe we should keep `16.x` in the matrix (perhaps `[16.x, 18.x]`) to test for backwards compatibility given that builtins and extensions in VS Code will be using `Node 16x` APIs.

If we attempt to build VS Code with a newer Node version we get the [warning](https://github.com/microsoft/vscode/blob/be900a4b816a6cf83f3413f99bf6dcdd447ad7de/build/npm/preinstall.js#L16-L18):

```
vscode ±|main|→ node -v
v18.17.0

vscode ±|main|→ yarn
yarn install v1.22.19
$ node build/npm/preinstall.js
*** Warning: Versions of node.js >= 17 have not been tested.

[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
...
```

And their CI makes use of Node 16:

https://github.com/microsoft/vscode/blob/be900a4b816a6cf83f3413f99bf6dcdd447ad7de/.github/workflows/ci.yml#L116-L118

msujew (commented):
We need to upgrade the node version [here](https://github.com/eclipse-theia/theia/blob/1e16b0899e5bf857b147a8bcd2f2c2da548dd18b/.gitpod.dockerfile#L20) as well. I also agree with Vince, that as long VSCode extensions are using Node 16, that we should use 16 in our testing matrix as well.

Otherwise this looks pretty good to me 👍

.github/workflows/ci-cd.yml
paul-marechal: Done.

msujew (approved):
The changes look good to me 👍 

@vince-fugnitto Do we want to include this in the upcoming release?

vince-fugnitto:
> @vince-fugnitto Do we want to include this in the upcoming release?

Sounds good to me!

package.json
```diff
@@ -4,17 +4,17 @@
   "version": "0.0.0",
   "engines": {
     "yarn": ">=1.7.0 <2",
-    "node": ">=16.14.0 <17"
+    "node": ">=16"
```
jcortell68, line 7: Curious: the Developing.md was updated to say developers should use >= 18.17 and < 21. But this line says >= 16. Why that discrepancy?

msujew: While we generally recommend the version range that you've mentioned, there is no technical reason for limiting it to that range. Though we know that < 16.0.0 doesn't work due to native dependencies. 18.17.0 is specifically the recommended minimum version because that's the version that our Electron version uses.

paul-marechal: We recommend a certain version range in our documentation as we try to make sure it works for said range.

This `engines.node` field will have Yarn stop the build if you use anything outside the defined range, and I might have left it lax so that it's more convenient when switching between node versions for development purposes.

<details>
  <summary>Most other NPM projects -including our deps- are also pretty lax with this range (click to expand).</summary> 

  - https://github.com/websockets/ws/blob/d8dd4852b81982fc0a6d633673968dff90985000/package.json#L32
  - https://github.com/socketio/socket.io/blob/9a2a83fdd42faa840d4f11fd223349e5d8e4d52c/package.json#L92
  - https://github.com/jprichardson/node-fs-extra/blob/a277cbbdf630424482b1e7418749ccaac0db98fe/package.json#L6
  - https://github.com/RogerDeng/drivelist/blob/6b79bf74a40e19bc69a561b978067807060192c4/package.json#L36
  - https://github.com/facebook/react/blob/faa65778b7c277e0020b8bb1fc172e11a863c0b7/packages/react/package.json#L38
  - https://github.com/Axosoft/nsfw/blob/13a84d97c0d778535bbc3baebcc6dc1216a26325/package.json#L22
  - https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/package.json#L113
  - etc
</details>

But we could move the lower bound along the minimum recommended version.

At the very minimum we need to update this range based on the kind of constructs we use from NodeJS: we can't use new Node 18 syntax and claim that we require anything below Node 18 to run our code base. Although syntax-wise we're dependent on our TypeScript compilation target: No matter what TS syntax we use `tsc` will "dumb it down" for us in the output JS based on our configs. [We're currently targeting ES2019](https://github.com/eclipse-theia/theia/blob/dcfff8326ed4272710e51665a52782a2a2ba2a5e/configs/base.tsconfig.json#L19) and it seems like [Node 16 supports it very well](https://node.green/#ES2019).

paul-marechal: In the long run it's in everyone's best interest to always use the latest Node LTS, but some might still lag behind so the range can accommodate for it somewhat.

Now I remember: Had I updated the range to `>=18` for that release then all users of the framework would have had build issues because they most likely were still running on Node 16 at that point. So instead of breaking everyone on release I kept it that way so people can update when they're ready. It's not like the project will stop working on Node 16 right away :)

Issue #12632, opened 2023-06-21
[vscode][proposed] dropMetadata proposed API updates for 1.79


rschnekenbu:
dropMetadata proposed API has evolved between 1.78 and 1.79:
- DocumentDropEdit now has 2 additional properties: id and priority
- DocumentDropEditProvider does not have required id property

This proposed API is used by several builtins (ipynb & markdown-language-feature)

PR #12125, opened 2023-01-30
Support `DocumentDropEditProvider` (#12025)

jonah-iden:
#### What it does
resolves #12025
this implements the vscode extension api for 
- `DocumentDropEdit `
- `DocumentDropEditProvider` 
- as well as the corresponding registration method in the `language` namespace

also implments DTOs including converters and ability to resolve the file Data for those for #11776.

#### How to test
here is a small test extension: 
[documentdropedit-test-0.0.1.zip](https://github.com/eclipse-theia/theia/files/10536306/documentdropedit-test-0.0.1.zip)

or you can build it here:
https://github.com/jonah-iden/documentdropedit-test

1. install extension
2. execute "Register dropEdit provider"
3. drop one or multiple fils into editor
4. should see messages with name of each file and content of one file should have been added to current cursor position in the editor

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #12736, opened 2023-07-20
[vscode] Add proposed dropMetadata API

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/12632

This PR adds the groundworks for supporting the proposed API. Note that while the API is supported on a type level, our current monaco version isn't able to actually make use of the changes. I've annotated all places in the code with `@monaco-uplift` so that any contributor can improve on this behavior once we have performed the uplift to `^1.79.0`.

#### How to test

There isn't really anything to test, no behavior should have changed. Confirm that the changes in https://github.com/eclipse-theia/theia/pull/12125 still work as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
+++ b/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
@@ -329,6 +329,10 @@ export interface DocumentDropEdit {
     additionalEdit?: WorkspaceEdit;
 }
 
+export interface DocumentDropEditProviderMetadata {
+    readonly dropMimeTypes: readonly string[];
+}
+
 export interface DataTransferFileDTO {
     readonly name: string;
     readonly uri?: UriComponents;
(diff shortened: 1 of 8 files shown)


tsmaeder:
@msujew this seems to work fine. The only strange thing when I test according to the instruction in https://github.com/eclipse-theia/theia/pull/12125 is this: I get both a new editor with the file I'm dropping _and_ I get the file contents inserted into the editor I dropped into. Is this the desired behavior?

msujew:
@tsmaeder I assume that this is an untested combination of https://github.com/eclipse-theia/theia/pull/12065 and https://github.com/eclipse-theia/theia/pull/12125. Since both PRs were submitted at similar times, I assume that the latter one didn't include the former yet, so I didn't notice the strange behavior during testing. And no, this isn't desired behavior. I believe that vscode has a special mode (activated via `shift`) to change between file dropping and resolving via the `DocumentDropEditProvider`. We don't seem to do that.

tsmaeder:
I've opened https://github.com/eclipse-theia/theia/issues/12748

tsmaeder (approved):
Since the observed weird behavior cannot possibly be due to the changes in this PR, this looks good to me.

PR #12753, opened 2023-07-25
Fix bug in `TheiaDialog` page object

tortmayr:
#### What it does

Fixes a  bug with the `TheiaDialog.waitUntilMainButtonIsEnabled` method not working as expected. 
Calling the method resulted in an infinite timeout because `this` was used inside of the `pageFunction`. The `pageFunction` is executed inside the browser where  the `TheiaDialog` page objectis not available and therefore it was `undefined`.
Call `waitUntilMainButtonIsEnabled` in `TheiaExplorView.renameNode` to ensure that the method is covered by the test case execution.

#### How to test

This change is directly tested by the playwright test suite.
#### Review checklist

#### Reminder for reviewers


planger (approved):

Issue #12713, opened 2023-07-14
A VS Code extension contributing a configurationDefault doesn't seem to work


jcortell68:
### Bug Description:
According to the Theia codebase, a VS Code extension should be able to contribute a configuration default via the `contribute.configurationDefaults` contribution point. I tried it and it seems to not have any effect. The same contribution point declaration in VS Code works as expected.

```
git clone -b configuration https://github.com/jcortell68/theiasandbox
cd theiasandbox
yarn
yarn start:electron
```

Go to File > Preferences > Settings. The first setting you see is **Files: Auto Save**. Set it to "off", if it's not already set to that. Then hover the mouse to the left of the setting to make the gear icon appear. Click it and choose. **Reset Setting**. Notice it stays on "off", despite the VS Code extension having declared:

```
    "configurationDefaults": {
      "files.autoSave": "onWindowChange"
    }
```

You can verify that the VS Code extension has been found by finding and invoking the **Hello World** command in the command palette (CTRL-SHIFT-P)

### Additional Information

- Operating System: Windows
- Theia Version: 1.39

Issue #10812, opened 2022-02-28
Handle more `files.autoSave` / `editor.autoSave` options


colin-grant-work:
### Feature Description:

The VSCode preference `files.autoSave` (currently `editor.autoSave` in Theia) has four options: `off`, `afterDelay`, `onFocusChange`, `onWindowChange`. Theia currently only has `off` and `on`. For compatibility, we should adjust the name and handle all cases correctly.

PR #4296, opened 2019-02-08
support language specific preferences

akosyakov:
fix #1106 - language specific preferences
fix #3540 - `configurationDefault` vscode contribution point in order to provide default language specific preferences

lang_preferences

Issue #3540, opened 2018-11-19
[plug-in] Implement 'configurationDefault' contribution point


akurinnoy:
[contributes.configurationDefaults](https://code.visualstudio.com/api/references/contribution-points#contributes.configurationDefaults)

Configuration example:
```
"contributes": {
    "configurationDefaults": {
        "[markdown]": {
            "editor.wordWrap": "on",
            "editor.quickSuggestions": false
        }
    }
}
```

akosyakov:
@akurinnoy Are you working on it?

akurinnoy:
@akosyakov unfortunately, we cannot take it in following sprint, but, we could do it in next sprint in 3 weeks, I think

akosyakov:
@akurinnoy Would it be alright if i look at it now?

akurinnoy:
@akosyakov Sure! Should I re-assign the issue?

PR #12758, opened 2023-07-26
A VS Code extension configurationDefault does not work #12713

jfaltermeier:
#### What it does

Fixes: #12713

* plugin contribution handler will now also register configurationDefault for non language related configuration changes
* extend preference-contribution#doSetSchema to allow modifying default values for existing schemas when contributed via a special schema id
* also update select options in preference-select-input, since the default value may change now, which requires the SelectOptions to recreate

The solution is based on the existing implementation for providing language specific editor settings:
https://github.com/eclipse-theia/theia/issues/3540
https://github.com/eclipse-theia/theia/pull/4296
In the existing solution we are using a special `defaultOverrides` PreferenceSchema that gets registered. I've adapted the registration code to also work for existing preference keys when provided via this special schema. 

I am open for suggestions to implement this in a different way, however we should probably refactor the language specific editor overrides then as well. 

#### How to test
See the bug report for a reproducer: https://github.com/eclipse-theia/theia/issues/12713
Check the preference's (default) value when installing/uninstalling the extension. 
Check the preference UI (default highlight in drop down) and test resetting to the default value

Please be aware of this bug during testing with the vsix from the reproducer: https://github.com/eclipse-theia/theia/issues/10812

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/browser/views/components/preference-select-input.ts
+++ b/packages/preferences/src/browser/views/components/preference-select-input.ts
@@ -29,14 +29,15 @@ import { escapeInvisibleChars } from '@theia/core/lib/common/strings';
 export class PreferenceSelectInputRenderer extends PreferenceLeafNodeRenderer<JSONValue, HTMLDivElement> {
 
     protected readonly selectComponent = React.createRef<SelectComponent>();
-    protected readonly selectOptions: SelectOption[] = [];
+
+    protected selectOptions: SelectOption[] = [];
 
     protected get enumValues(): JSONValue[] {
         return this.preferenceNode.preference.data.enum!;
     }
 
     protected updateSelectOptions(): void {
-        this.selectOptions.splice(0);
+        const updatedSelectOptions: SelectOption[] = [];
         const values = this.enumValues;
         const preferenceData = this.preferenceNode.preference.data;
         const defaultValue = preferenceData.default;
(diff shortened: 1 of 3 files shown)


packages/preferences/src/browser/views/components/preference-select-input.ts
```diff
@@ -29,13 +29,14 @@ import { escapeInvisibleChars } from '@theia/core/lib/common/strings';
...
+    protected readonly selectOptions: SelectOption[] = [];
 
     protected get enumValues(): JSONValue[] {
         return this.preferenceNode.preference.data.enum!;
     }
 
-    protected get selectOptions(): SelectOption[] {
-        const items: SelectOption[] = [];
+    protected updateSelectOptions(): void {
+        this.selectOptions.splice(0);
```
msujew, line 39: Suggestion: This feels kind of hacky to me. Maybe just make the `selectOptions` non readonly and assign to it directly in `updateSelectOptions`? It's `protected` anyway.

msujew (approved):
Looks pretty good to me. I've used [this extension](https://github.com/eclipse-theia/theia/files/12229841/configurationdefault-0.0.1.zip) for testing, which changes the `editor.fontSize` configuration for all editors to `20` and for markdown editors to `30`. 👍 

* The defaults are correctly applied in the preferences.
* Changing to the original default value highlights the preference as changed and resetting it changes it back to the config default override.
* Deinstalling the plugin and reloading the app correctly resets the preference defaults.
* Changing the preferences without installing a plugin that changes the defaults still works as expected.

I have a minor suggestion, see below.

PR #12799, opened 2023-08-04
vscode: independent editor/title/run menu

### LINKED ISSUE #12687: VS Code menu contributions to "editor/title/run" seem to behave no different than to "editor/title"
### Bug Description:
I see that support for menu insertion point "editor/title/run" was added to Theia in recent months. It works...but not really how it works in VS Code. 

In VS Code, if you have two or more contributions to "editor/title/run", a "Run or Debug..." submenu item is added to the editor title area. It appears as a regular menu item in the editor title toolbar (as icon or text), with a carat next to it for opening the submenu.

image

In Theia, the contributions to "editor/title/run" seem to appear exactly as if they were contributed to "editor/title". I.e., there is no grouping of them into a "Run or Debug..." submenu with a caret, etc.

I don't know if this is a known limitation of the current implementation. Perhaps the point was to get the contributed menu items into the UI in the simplest form possible, but not provide the full behavior of VS Code. However, if this is an unintended behavior/gap, I'll provide reproducibility steps.

cdamus:
#### What it does

In VS Code, contributions to the `"editor/title/run"` menu contribution point are not added to the ellipsis menu but to a dedicated item on the toolbar. This commit makes Theia do the same, except that unlike VS Code, a pop-up menu is always presented, even if there is only one action available.

In my opinion, the inconsistency of the VS Code implementation in not showing the Run/Debug pop-up menu but just the one contributed action on the toolbar, which looks quite different to the Run/Debug pop-up menu action, is confusing. If an action is contributed to an extension point that is a menu of Run/Debug actions, then that should show a menu of Run/Debug actions even if there is only one item on the menu.

_If it really is necessary to emulate this detail of the VS Code behaviour, then that should be a follow-up issue to discuss with the user community._

Note also that this PR uses the `debug-alt` icon from the VS Code `codicons` set as the best available alternative from that set to the icon that VS Code uses for its "Run/Debug" menu, which is not in the `codicons` set. If somebody would like to contribute a better license-compatible open-source icon for the purpose, please do!

Fixes #12687

#### How to test

1. Clone the `menu_bug_3` of the OP's sample repository used for other recent bug reports: [Theia Sandbox](/jcortell68/theiasandbox).
2. Change the menu contribution in the `myext` extension to the `"editor/title/run"` point (it should be checked out as `"editor/lineNumber/context"` for another issue report) and remove the `"group"` property that isn't needed for this use case.
3. Because there probably wouldn't be any other actions on the "more" menu, I recommend adding the `"myext.goodbyeWorld"` command from this extension to the `"editor/title"` menu so that we can see that this menu exists and doesn't get contributions intended for the `"editor/title/run"` menu.
4. Build the `myext` extension and link it into your `plugins/` directory.
5. Launch Theia and open some editor (e.g., TS or JSON file).
6. Observe that the "Run/Debug" menu shows up something like the screen recording below, with the "Hello World" action from the extension. Additionally, hovering the pointer over it shows a "Run or Debug..." tooltip, similar to the "More Items..." tooltip on the `...` "more" menu.
7. Verify that the "Hello World" action is _not_ in the `...` menu.

Also worth verifying are some related use cases:

- check that a Theia instance without any contributions to the `"editor/title/run"` menu does not show that menu in the toolbar.
- also that in the case there are contributions to this menu, but all of them are suppressed by their `"when"` clauses not matching the current context, then in this case also the menu does not appear in the toolbar.

CleanShot 2023-08-04 at 13 38 43

#### Review checklist

#### Reminder for reviewers


jcortell68:
> If an action is contributed to an extension point that is a menu of Run/Debug actions, then that should show a menu of Run/Debug actions even if there is only one item on the menu.

@cdamus I agree. IMO, doing things better than VS Code in this case is better than emulating VS Code.

packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar-registry.ts
```diff
@@ -103,9 +103,10 @@ export class TabBarToolbarRegistry implements FrontendApplicationContribution {
         }
         const result: Array<TabBarToolbarItem | ReactTabBarToolbarItem> = [];
         for (const item of this.items.values()) {
-            const visible = TabBarToolbarItem.is(item)
+            const visible = TabBarToolbarItem.is(item) && item.command
```
martin-fleck-at, line 106: I think this overall condition is flawed as I can see the menu also on views, not only editors. I believe the problem is that we do not evaluate the conditions on menu toolbar items (only if they have content) but we do register our menu in `menus-contribution-handler.ts` with a visibility condition. Since this already a rather complex condition, I'd suggest to pull that out into something that is more readable and easier to overwrite, e.g.,:


```
    protected isVisible(item: TabBarToolbarItem | ReactTabBarToolbarItem, widget: Widget): boolean {
        if (AnyToolbarItem.isConditional(item)) {
            if (item.isVisible && !item.isVisible(widget)) {
                return false;
            }
            if (item.when && !this.contextKeyService.match(item.when, widget.node)) {
                return false;
            }
        }
        if (TabBarToolbarItem.is(item) && item.command && !this.commandRegistry.isVisible(item.command, widget)) {
            return false;
        }
        if (MenuToolbarItem.is(item) && !this.hasMenuContent(item, widget)) {
            return false;
        }
        return true;
    }
```

packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar.tsx
```diff
@@ -285,6 +287,57 @@ export class TabBarToolbar extends ReactWidget {
...
+     * Renders a toolbar item that is a menu, presenting it as a button with a little
+     * chevron decoration that pops up a floating menu when clicked.
+     *
+     * @param item a toolbar item that is a menu item
+     * @returns the rendered toolbar item
+     */
+    protected renderMenuItem(item: TabBarToolbarItem & MenuToolbarItem): React.ReactNode {
+        const icon = typeof item.icon === 'function' ? item.icon() : item.icon ?? 'ellipsis';
+        return <div key={item.id} className={TabBarToolbar.Styles.TAB_BAR_TOOLBAR_ITEM + ' enabled menu'}>
+            <div id={item.id} className={codicon(icon, true)} onClick={this.showPopupMenu.bind(this, item.menuPath)}
```
martin-fleck-at, line 300: Could you please add the `onClick` listener to the parent? It feels a bit odd that you cannot open the menu by clicking the chevron.

cdamus: The bigger target area for clicking is important for accessibility in any case!

packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar-registry.ts
cdamus: A welcome refactoring suggestion.

martin-fleck-at (approved):
I re-tested the changes and everything looks great to me now, thank you very much Christian! As soon as the CI is happy we can merge this.

cdamus:
Commit 0e6186fa621488aa84724513c2192b8658d107ee adds the missing return type annotation found by ESLint.

Issue #12724, opened 2023-07-18
Problems View fails to show new problems until you alter the horizontal size of the app window


jcortell68:
### Bug Description:
I was playing around with defining a problem matcher in a VS Code extension. In Theia, my problem matcher correctly parsed the task output, and the problems view shows that there is an error via the title decorator, but the view itself doesn't show the error. Only after altering the width of the Theia app's frame window, does the problems view show the problem entry.

This is what the Problems view looks like after running the task. Note the annotation showing that there's one problem, but the view is telling me there are no problems:
![image](https://github.com/eclipse-theia/theia/assets/109228283/c2a4a3f1-9965-49ba-b90c-ec3252efdc2e)

Here's what happens when I change the width of the frame window:
![image](https://github.com/eclipse-theia/theia/assets/109228283/c5ece63b-da5d-4e37-899c-a431ddabaeec)

Oddly enough, changing the height of the window doesn't cause the problems view to show the proper content. Only changing the width does.

Finally, there are two more bugs at play, which may be considered separate issues. Let me know if you'd like me to create dedicated tickets for them
1. Note the garbage preceding the filename (somefile.abc) in the second screenshot above.
2. An exception is thrown (appears in the node terminal) when running the task
```
2023-07-18T18:23:59.174Z root ERROR Request resize failed with error: Cannot resize a pty that has already exited Error: Cannot resize a pty that has already exited
    at WindowsPtyAgent.resize (C:\j.cortell\git\theiasandbox\node_modules\node-pty\lib\windowsPtyAgent.js:141:23)
    at WindowsTerminal.<anonymous> (C:\j.cortell\git\theiasandbox\node_modules\node-pty\lib\windowsTerminal.js:132:26)
    at WindowsTerminal._defer (C:\j.cortell\git\theiasandbox\node_modules\node-pty\lib\windowsTerminal.js:157:24)
    at WindowsTerminal.resize (C:\j.cortell\git\theiasandbox\node_modules\node-pty\lib\windowsTerminal.js:131:14)
    at TaskTerminalProcess.resize (C:\j.cortell\git\theiasandbox\node_modules\@theia\process\lib\node\terminal-process.js:158:23)
    at ShellTerminalServer.resize (C:\j.cortell\git\theiasandbox\node_modules\@theia\terminal\lib\node\base-terminal-server.js:112:18)
[...]

PR #12802, opened 2023-08-07
Properly re-render problem widget and fix problem matching

martin-fleck-at:
#### What it does
- Ensure we re-render the tree model if the problem widget is activated
- Update Ansi stripping to catch Windows 'clear screen' and other codes
- Consistently use Theia URI in problem matcher protocol

Fixes https://github.com/eclipse-theia/theia/issues/12724

#### How to test
- Follow instructions in bug report or checkout https://github.com/eclipse-theia/theia/tree/martin-fleck-at/issue-12724-test
- Run Electron App (either through launch config or command line)
- Create a task in a workspace with the following content:
```
{
   "version": "2.0.0",
   "tasks": [
      {
         "label": "echo",
         "type": "shell",
         "command": "@echo somefile.abc:1:4: warning: the roof is on fire ",
         "problemMatcher" : ["$myproblemmatcher"]
      }
   ]
}
```
- Run Task 'echo' and check the Problems view:

image
(Windows)

image
(File path on Hover)

image
(Linux)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/markers/src/browser/problem/problem-widget.tsx
+++ b/packages/markers/src/browser/problem/problem-widget.tsx
@@ -21,7 +21,7 @@ import { ProblemTreeModel } from './problem-tree-model';
 import { MarkerInfoNode, MarkerNode, MarkerRootNode } from '../marker-tree';
 import {
     TreeWidget, TreeProps, ContextMenuRenderer, TreeNode, NodeProps, TreeModel,
-    ApplicationShell, Navigatable, ExpandableTreeNode, SelectableTreeNode, TREE_NODE_INFO_CLASS, codicon
+    ApplicationShell, Navigatable, ExpandableTreeNode, SelectableTreeNode, TREE_NODE_INFO_CLASS, codicon, Message
 } from '@theia/core/lib/browser';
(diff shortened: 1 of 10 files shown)


jfaltermeier (commented):
Thank you! I've tested on Windows and Linux. 
There is no need to resize the problem view anymore. Also the encoding issue in the filename is fixed. 

However the bug report also mentions `ERROR Request resize failed with error: Cannot resize a pty that has already exited Error: Cannot resize a pty that has already exited`. 
I am still seeing this on Windows.

martin-fleck-at:
@jfaltermeier Thanks! I'll look into this and see if there is something I can do.

martin-fleck-at:
@jfaltermeier That was actually quite a messy one but I think I found a good solution based on how VS Code handles the terminals in Windows. Could you please have another look?

jfaltermeier (approved):
Thanks, lgtm now!

martin-fleck-at:
@tsmaeder @msujew Since you guys also worked on the terminal, do you wanna have another look at this? Since the change is not that small, another pair of eyes might provide additional insights.

Issue #12725, opened 2023-07-18
contributes.problemPatterns is not working


jcortell68:
While the [compatibility status](https://eclipse-theia.github.io/vscode-theia-comparator/status.html) doesn't call out `problemPatterns` explicitly (it calls out `problemMatchers`, which is related), [the Theia code](https://github.com/eclipse-theia/theia/blob/41c6d9f6f15caf9b2059b6a03c91699d31d80029/packages/plugin-ext/src/main/browser/plugin-contribution-handler.ts#L338) does seem to indicate it's supported.

Yet, it doesn't seem to work. I created a small VS Code extension example that works fine inside Theia when I embed the pattern expression directly into the `problemMatchers` declaration. However, when I extract the expression into a `problemPatterms` and then use that from the `problemMatchers`, that does _not_ work.

The same VS Code extension works in VS Code, which indicates a bug in Theia's support for `problemPatterns`

### Steps to Reproduce:

```
git clone -b problem_patterns https://github.com/jcortell68/theiasandbox
cd theiasandbox
yarn
yarn start:electron
```
Next, create a task in the launched Theia app as follows (contents of .theia/tasks.json)

```
{
   "version": "2.0.0",
   "tasks": [
      {
         "label": "echo",
         "type": "shell",
         "command": "@echo somefile.abc:1:4: warning: the roof is on fire ",
         "problemMatcher" : ["$myproblemmatcher"]
      }
   ]
}
```

Finally, launch the task and you'll see that the Problems View doesn't end up with an entry. You can easily see things working with the expression embedded in the `problemMatchers` by getting a different git branch

```
git checkout problem_matchers
```

### Additional Information

- Operating System: Windows
- Theia Version: 1.39

PR #12805, opened 2023-08-07
Ensure contributed problem patterns are found correctly

martin-fleck-at:
#### What it does
- Resolve pattern names correctly by removing leading '$'
- Introduce utility function for dealing with variable names
- Ensure 'ready' promises are already available with object creation

Fixes https://github.com/eclipse-theia/theia/issues/12725

#### How to test
- Follow instructions in bug report or checkout https://github.com/eclipse-theia/theia/tree/martin-fleck-at/issue-12725-test
- Run Electron App (either through launch config or command line)
- Create a task in a workspace with the following content:
```
{
   "version": "2.0.0",
   "tasks": [
      {
         "label": "echo",
         "type": "shell",
         "command": "echo somefile.abc:1:4: warning: the roof is on fire ",
         "problemMatcher" : ["$myproblemmatcher"]
      }
   ]
}
```
- Run Task 'echo' and check the Problems view:
image

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/task/src/common/task-util.ts
+++ b/packages/task/src/common/task-util.ts
@@ -14,10 +14,30 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
+/**
+ * Converts the given standard name to a variable name starting with '$' if not already present.
+ *
+ * Variable names are used, for instance, to reference problem matchers, within task configurations.
+ *
+ * @param name standard name
+ * @returns variable name with leading '$' if not already present.
+ *
+ * @see {@link fromVariableName} for the reverse conversion.
+ */
 export function asVariableName(name: string): string {
     return name.startsWith('$') ? name : `$${name}`;
 }
 
+/**
+ * Converts a given variable name to a standard name, effectively removing a leading '$' if present.
+ *
+ * Standard names are used, for instance, in registries to store variable objects
+ *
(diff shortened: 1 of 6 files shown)


packages/task/src/common/task-util.ts
```diff
@@ -0,0 +1,23 @@
...
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
```
jfaltermeier, line 15: Should we add some tsdocs to these new util methods?

jfaltermeier (commented):
Thank you, looks good to me and works as expected. 
I think we should add some docs to `asVariableName` and `fromVariableName` though, so that users know what happens when they pass variables/non-variables to each of the methods.

packages/task/src/common/task-util.ts
martin-fleck-at: Good point! I pushed an update with some comments.

jfaltermeier (approved):

Issue #12755, opened 2023-07-25
vscode: implement support for `file/newFile` menu contributions


vince-fugnitto:
#### Feature Description:

The goal is to add support for the `file/newFile` menu path which extension authors can reference when declaring their [menu contributions](https://code.visualstudio.com/api/references/contribution-points#contributes.menus).

> The New File item in the File menu and Get Started page - `file/newFile`

#### Additional Information

The `vscode.ipynb` for reference makes use of the menu path:

```
2023-07-25T18:49:43.379Z root WARN Failed to register a menu item for plugin vscode.ipynb contributed to file/newFile {command: 'ipynb.newUntitledIpynb', submenu: undefined, alt: undefined, group: 'notebook', when: undefined} Error: Could not find submenu with id file/newFile
    at MenuModelRegistry.getMenuNode (http://localhost:3000/bundle.js:147792:23)
    at http://localhost:3000/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:5826:66
    at Array.forEach (<anonymous>)
    at MenusContributionPointHandler.handle (http://localhost:3000/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:5820:37)
    at http://localhost:3000/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:6563:71
    at pushContribution (http://localhost:3000/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:6458:32)
    at PluginContributionHandler.handleContributions (http://localhost:3000/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:6563:9)
    at HostedPluginSupport.loadContributions (http://localhost:3000/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:577:61)
    at HostedPluginSupport.doLoad (http://localhost:3000/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:488:42)
    at async ProgressService.withProgress (http://localhost:3000/bundle.js:151138:20)
```

jonah-iden:
Hi, could someone assign me, i'll probably try to implement this later today

PR #12814, opened 2023-08-08
Submenu contribution to editor/title and view/title not working #12706

### LINKED ISSUE #12706: VS Code submenu contribution to editor/title and view/title not working correctly
### Bug Description:
I was testing support for contributing a submenu via a VS Code extension. It works as expected in all the locations I tried but two:
* "editor/title"
* "view/title"

The buggy behavior is identical in both. The following will refer to only "editor/title" for simplicity, though.

The submenu appears in "invisible" fashion. The submenu's icon (a pen) should appear in the editor title toolbar, as shown here in an equivalent VS Code sample:
image

In Theia, the submenu does in fact end up in the editor title toolbar, but without its icon (and without its label). And so it's effectively invisible. I.e., you can't see that it's there. If you hover the mouse over that area, you will see a tooltip pop up with the submenu's name. This hints that your mouse is over something you can't see. And sure enough, if you left click there, you will end up seeing the submenu.

image

But even with this "workaround" you still end up with undesirable behavior. Instead of being presented with the three submenu items (Larry, Curly and Moe), you see the submenu title and icon, and you have to click yet again to get to the submenu items.

This all works intuitively in VS Code. In Theia, the implementation is off.

Note that this might seem similar to issue 12686, since it's related to the same menu insertion point ("editor/title") and it's also involves an "invisible" menu contribution. But this is a different case and I don't think fixing one will fix the other. In that other ticket, the problem is that if a menu item lacks an icon, it's shown invisibly instead of with its title. That does not apply here. Here, the submenu is given an icon (in package.json) but it does not appear in the UI. And then as mentioned above, even if it did, there's a second issue at hand...the additional click that would be necessary to get to the submenu items.

Finally, note that the example below contributes the submenu to other locations as well and the behavior seems to be good there. I.e., if you right click in the editor body, you see this

image

### Steps to Reproduce:

```
git clone -b menu_bug_5 https://github.com/jcortell68/theiasandbox
cd theiasandbox
yarn
yarn start:electron
```

In the launched Theia, initiate the Hello World command via the Command Palette, to ensure the VS Code extension is activated and the Tree View is populated. Not technically needed to see the problem reported here.

### Additional Information

- Operating System: Windows
- Theia Version: 1.39

jfaltermeier:
#### What it does

The menu registered at the registry is correctly registered (Context menu > Sub Menu > Child Entries). 

In this case the renderer should skip rendering the sub menu however. 
The sub menu is represented by the toolbar button and its tooltip already. 
So we just want to see the child entries in the context menu. 

This PR introduces a helper method to adjust an existing menu-model to be rendered without a single root-submenu. Renderers may use this helper as required.  

* introduce option in RenderContextMenuOptions to ask renderer to not render a menu with single submenu. Instead render only the children
* add helper method to menu-model-registry to remove the parent nodes from a menu node as described above
* adjust renderers and callers accordingly
* fix icons in toolbar

fixes #12706

Peek 2023-08-08 15-31

#### How to test
See the reproducer on #12706

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/menu/menu-model-registry.ts
+++ b/packages/core/src/common/menu/menu-model-registry.ts
@@ -269,7 +269,7 @@ export class MenuModelRegistry {
         }
         let nonEmptyNode = undefined;
         for (const child of fullMenuModel.children) {
-            if (!this.isEmpty(child.children!)) {
+            if (!this.isEmpty(child.children || [])) {
                 if (nonEmptyNode === undefined) {
                     nonEmptyNode = child;
                 } else {
@@ -285,16 +285,11 @@ export class MenuModelRegistry {
         return CompoundMenuNode.is(nonEmptyNode) ? nonEmptyNode : fullMenuModel;
     }
 
-    private allChildrenCompound(children: ReadonlyArray<MenuNode>): boolean {
-        for (const child of children) {
-            if (!CompoundMenuNode.is(child)) {
-                return false;
-            }
-        }
-        return true;
+    protected allChildrenCompound(children: ReadonlyArray<MenuNode>): boolean {
+        return children.every(CompoundMenuNode.is);
     }
 
-    private isEmpty(children: ReadonlyArray<MenuNode>): boolean {
+    protected isEmpty(children: ReadonlyArray<MenuNode>): boolean {
         if (children.length === 0) {
             return true;
         }
@@ -302,7 +297,7 @@ export class MenuModelRegistry {
             return false;
         }
         for (const child of children) {
-            if (!this.isEmpty(child.children!)) {
+            if (!this.isEmpty(child.children || [])) {
                 return false;
             }
         }
--- a/packages/core/src/common/menu/menu-model-registry.ts
+++ b/packages/core/src/common/menu/menu-model-registry.ts
@@ -278,6 +278,10 @@ export class MenuModelRegistry {
             }
         }
 
+        if (CompoundMenuNode.is(nonEmptyNode) && nonEmptyNode.children.length === 1 && CompoundMenuNode.is(nonEmptyNode.children[0])) {
+            nonEmptyNode = nonEmptyNode.children[0];
+        }
+
         return CompoundMenuNode.is(nonEmptyNode) ? nonEmptyNode : fullMenuModel;
     }
 
--- a/packages/core/src/common/menu/menu-model-registry.ts
+++ b/packages/core/src/common/menu/menu-model-registry.ts
@@ -255,6 +255,57 @@ export class MenuModelRegistry {
     }
 
     /**
+     * Checks the given menu model whether it will show a menu with a single submenu.
+     *
+     * @param fullMenuModel the menu model to analyze
+     * @param menuPath the menu's path
+     * @returns if the menu will show a single submenu this returns a menu that will show the child elements of the submenu,
+     * otherwise the given `fullMenuModel` is return
+     */
+    removeSingleRootNode(fullMenuModel: MutableCompoundMenuNode, menuPath: MenuPath): CompoundMenuNode {
+        // check whether all children are compound menus and that there is only one child that has further children
+        if (!this.allChildrenCompound(fullMenuModel.children)) {
+            return fullMenuModel;
+        }
+        let nonEmptyNode = undefined;
+        for (const child of fullMenuModel.children) {
+            if (!this.isEmpty(child.children!)) {
+                if (nonEmptyNode === undefined) {
+                    nonEmptyNode = child;
+                } else {
+                    return fullMenuModel;
+                }
+            }
+        }
+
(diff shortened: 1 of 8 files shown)


JonasHelming:
@jonah-iden You mentioned yesterday that you miss some menus, is this maybe related?

jonah-iden:
@JonasHelming thanks for the info, but the not notebook specific menus im missing are:
`file/newFile` from btoh the python and ipynb extension and `editor/content` which only the python extension is missing.
All other missing menus are notebook specific, so i have to add them myself

JonasHelming:
https://github.com/eclipse-theia/theia/issues/12755

martin-fleck-at (changes requested):
Code looks good to me in general and as far as I can see, all the menus are showing their content as they do in VS Code, even if the location within the context menu sometimes varies a bit. So great work, Johannes!

The only issue I could detect was with the icon. For some reason, it does not show up for me. I looked a bit into this and the image resource is properly hosted and loaded so it seems more of a CSS thing. The CSS is applied in `plugin-shared-style.ts` and I noticed that if I change the `display` property from the default `inline` to something else, the icon shows up. However, I'm not sure if this might have some side effects on other icons. Could you please have another look?

jfaltermeier:
> Code looks good to me in general and as far as I can see, all the menus are showing their content as they do in VS Code, even if the location within the context menu sometimes varies a bit. So great work, Johannes!
> 
> The only issue I could detect was with the icon. For some reason, it does not show up for me. I looked a bit into this and the image resource is properly hosted and loaded so it seems more of a CSS thing. The CSS is applied in `plugin-shared-style.ts` and I noticed that if I change the `display` property from the default `inline` to something else, the icon shows up. However, I'm not sure if this might have some side effects on other icons. Could you please have another look?

Thanks, I can reproduce it when I merge the current master into my branch. I think you mentioned that you've rebased before testing. Looks like https://github.com/eclipse-theia/theia/pull/12827 causes the issue for my previous fix. I'll have a look.

jfaltermeier:
I've added a dedicated css entry that sets `display` to `flex`. With this it seems to work. 

I also noticed an issue with the `removeSingleRootNode` method while retesting. It did not detect submenus that are not rendered (because they have no children) as something that can be skipped as well. This was reproducible immediately after installing the test vsix. 
Please have a look at the implementation again.

packages/core/src/common/menu/menu-model-registry.ts
```diff
@@ -254,6 +254,61 @@ export class MenuModelRegistry {
...
+        }
+
+        if (CompoundMenuNode.is(nonEmptyNode) && nonEmptyNode.children.length === 1 && CompoundMenuNode.is(nonEmptyNode.children[0])) {
+            nonEmptyNode = nonEmptyNode.children[0];
+        }
+
+        return CompoundMenuNode.is(nonEmptyNode) ? nonEmptyNode : fullMenuModel;
+    }
+
+    private allChildrenCompound(children: ReadonlyArray<MenuNode>): boolean {
```
martin-fleck-at, line 288: I'd prefer this method to be protected.

packages/core/src/common/menu/menu-model-registry.ts
```diff
@@ -254,6 +254,61 @@ export class MenuModelRegistry {
...
+        for (const child of children) {
+            if (!CompoundMenuNode.is(child)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private isEmpty(children: ReadonlyArray<MenuNode>): boolean {
```
martin-fleck-at, line 297: I'd prefer this method to be protected.

packages/core/src/common/menu/menu-model-registry.ts
```diff
@@ -254,6 +254,61 @@ export class MenuModelRegistry {
...
+        if (children.length === 0) {
+            return true;
+        }
+        if (!this.allChildrenCompound(children)) {
+            return false;
+        }
+        for (const child of children) {
+            if (!this.isEmpty(child.children!)) {
```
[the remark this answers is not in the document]
jfaltermeier: We did a check that all children are CompoundMenuNodes before, but I can add it

jfaltermeier:
> Hello Johannes, thank you for the quick update! Everything looks good to me now, I just have a minor nitpick regarding private methods in classes that may be re-used by adopters. Please let me know what you think.

Thanks for the review. I've made the methods protected, although I would prefer them to be private. I understand why you prefer it the other way, but those methods are only existing, because we use them as helper methods from one the the actual API methods to make the code nicer. If this implementation has to change at some point, we still have the burden of having those additional API methods that can't be removed/adapted without a major API break. So I'm not a fan.

martin-fleck-at (approved):
Thank you Johannes, everything looks good to me now!

Issue #10874, opened 2022-03-11
Inline breakpoint appears when deleting the last line code


granicayv:
Steps to reproduce:
1) Create new file (for example, .py or .txt)
2) Write something on the first line
3) Put a breakpoint on the first line

<img width="229" alt="screenshot 2022-03-11 в 13 48 26" src="https://user-images.githubusercontent.com/40887281/157854277-973bc40a-1fe9-4897-b66a-6b6224951024.png">

4) Delete all code from this line

Result: Inline breakpoint appears when deleting the last symbol in the line.

<img width="225" alt="screenshot 2022-03-11 в 13 48 37" src="https://user-images.githubusercontent.com/40887281/157854287-27574335-fdce-4d01-b16a-fcc3e2564de7.png">

After that when user write something in this line, inline breakpoint moves after the text.

<img width="268" alt="screenshot 2022-03-11 в 13 48 44" src="https://user-images.githubusercontent.com/40887281/157854286-52b82cda-fddb-48d6-b8bc-c7057c11eb86.png">

Theia Version: 1.21.0
Operation System: mac os 11.2.3

vatsal-uppal-1997:
Hi @vince-fugnitto,

I've raised the following PR to resolve this issue:

https://github.com/eclipse-theia/theia/pull/12832

Can you please review it once?

PR #12832, opened 2023-08-14
Fix Inline breakpoint appears when deleting the last line code [#10874]

vatsal-uppal-1997:
#### What it does

Fixes: #10874 

This commit resolves the issue where inline breakpoints were getting added when removing the last line of code in a particular line.

#### How to test

Follow the steps as mentioned in #10874 
#### Review checklist

#### Reminder for reviewers


msujew (approved):
Looks good to me!

* [x] I can no longer reproduce the incorrectly set column-breakpoint with this change
* [x] No regression in the actual debugging process found

PR #12838, opened 2023-08-17
[vscode] Support EnvironmentVariableCollection description #12696

### LINKED ISSUE #12696: [vscode] Support EnvironmentVariableCollection description
EnvironnementVariableCollection now has a new description field.

https://code.visualstudio.com/updates/v1_80#_environmentvariablecollectiondescription

EnvironnementVariableDescription

jfaltermeier:
#### What it does
The first commit adds the new `description` field to `EnvironmentVariableCollection` . This will be enough to be compatible with VSCode 1.80.

Further commits add ui that shows the descriptions to the user.
VSCode does it like this: https://code.visualstudio.com/updates/v1_80#_environmentvariablecollectiondescription
As far as I know we did not have a similar ui yet, so I've started to implement one. The results looks like this atm:

Peek 2023-08-24 09-44

This is an enhanced preview for the terminal widget that shows the descriptions on hover, similar to VSCode. I've also included PID and Command Line like in VSCode.

Fixes #12696

Signed-off-by: Johannes Faltermeier <jfaltermeier@eclipsesource.com>

#### How to test
I've prepared three extensions that set either an undefined description, a string description, or a markdown description. 

Sources: 
[env-description-sources.zip](https://github.com/eclipse-theia/theia/files/12367528/env-description-sources.zip)

Extensions:
[env-description-extensions.zip](https://github.com/eclipse-theia/theia/files/12367534/env-description-extensions.zip)

#### Review checklist

#### Reminder for reviewers


tsmaeder:
@jfaltermeier it's not clear to me how I can test this: (having no familiarity with EnvVar collections. Do I just open a terminal?

jfaltermeier:
> @jfaltermeier it's not clear to me how I can test this: (having no familiarity with EnvVar collections. Do I just open a terminal?

@tsmaeder VSCode Extension may change the env variables in a new terminal. E.g. the builtin git extension will set `GIT_ASKPASS`:

```
echo $GIT_ASKPASS
/usr/share/code/resources/app/extensions/git/dist/askpass.sh
```

This functionality was existing in Theia already. So I haven't tested this part much.
The added VSCode API allows to describe the changes made to the environment to the users. 

In the three test extensions I attached I am changing `FOO`, `BAR`, and `FOOBAR`. 

So you may start with some default values and look at them in the terminal before installing the extensions. 

```
# in terminal used for launching theia:
export FOO=hello
export BAR=world
export FOOBAR=123

# yarn electron start
# open terminal in started Theia and run

echo $FOO
echo $BAR
echo $FOOBAR
```

Then install the extensions, close and reopen a terminal. This should not show changed values:

```
$ echo $FOO
echo $BAR
echo $FOOBAR
markdown
string
undefined
```

The new toolbar item shows the description strings on hover, one being a markdown string, one a regular string, and one being a missing undefined description. 

Please note that I pushed one additional commit, because I noticed to toolbar item was showing in too many locations.

packages/plugin-ext/src/plugin/terminal-ext.ts
```diff
@@ -313,7 +315,18 @@ export class TerminalServiceExtImpl implements TerminalServiceExt {
 
     private syncEnvironmentVariableCollection(extensionIdentifier: string, collection: EnvironmentVariableCollection): void {
         const serialized = [...collection.map.entries()];
-        this.proxy.$setEnvironmentVariableCollection(extensionIdentifier, collection.persistent, serialized.length === 0 ? undefined : serialized);
+        this.proxy.$setEnvironmentVariableCollection(extensionIdentifier, collection.persistent, serialized.length === 0 ? undefined : serialized,
+            this.descriptionToDTO(collection.description));
+    }
+
+    private descriptionToDTO(value: string | theia.MarkdownString | undefined): string | MarkdownStringDTO | undefined {
```
tsmaeder, line 322: Does it make sense to move this to the Converter? Are there other locations that do the same conversion?

packages/terminal/src/browser/base/terminal-widget.ts
```diff
@@ -58,6 +59,9 @@ export abstract class TerminalWidget extends BaseWidget {
      */
     abstract processInfo: Promise<TerminalProcessInfo>;
 
+    /** The extensions contributing to the environment of this terminal */
+    abstract contributingExtensions: Promise<Map<string, string | MarkdownString | undefined>>;
```
tsmaeder, line 63: I think the name is quite non-descriptive here: what is being held in this map?

packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -250,7 +263,7 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
             this.storageService.getData<string>(ENVIRONMENT_VARIABLE_COLLECTIONS_KEY).then(data => {
                 if (data) {
                     const collectionsJson: SerializableExtensionEnvironmentVariableCollection[] = JSON.parse(data);
-                    collectionsJson.forEach(c => this.shellTerminalServer.setCollection(c.extensionIdentifier, true, c.collection));
+                    collectionsJson.forEach(c => this.shellTerminalServer.setCollection(c.extensionIdentifier, true, c.collection, undefined));
```
tsmaeder, line 266: If we make the description part of the env var collection, we can serialize it at the same time, right?

packages/terminal/src/browser/terminal-info-toolbar-item.tsx
```diff
@@ -0,0 +1,113 @@
...
+            <div
+                id={this.id}
+                className='codicon codicon-terminal-bash action-label'
+                onMouseEnter={e => this.onMouseEnter(e, toDispose, widget)}
+                onMouseLeave={e => this.onMouseLeave(toDispose)}
+            ></div>
+        );
+    }
+
+    protected async onMouseEnter(
```
tsmaeder, line 49: It seems odd that we have to implement our own "show hover on mouse-over" support. Dont' we have this somewhere else already?

packages/terminal/src/browser/terminal-info-toolbar-item.tsx
```diff
@@ -0,0 +1,113 @@
...
+        event: React.MouseEvent<HTMLElement, MouseEvent>, toDispose: DisposableCollection, currentTerminal?: Widget
+    ): Promise<void> {
+        const currentTarget = event.currentTarget;
+        if (currentTerminal instanceof TerminalWidget) {
+            const extensions = await currentTerminal.contributingExtensions;
```
tsmaeder, line 54: Wouldn't it make more sense to let the TerminalWidget render (and cache) a Markdown string that we can just render with the markdow renderer instead of manipulating divs ourselves here?

packages/terminal/src/browser/terminal-info-toolbar-item.tsx
```diff
@@ -0,0 +1,113 @@
...
+            this.hoverService.requestHover({
+                content: mainDiv,
+                target: currentTarget,
+                position: 'right',
+            });
+        }
+    }
+
+    protected async onMouseLeave(toDispose: DisposableCollection): Promise<void> {
+        toDispose.dispose();
```
tsmaeder, line 111: `toDispose` is too generic. How about `hoverResources.dispose()`

jfaltermeier:
> I don't think I like the toolbar button as a UI here: in VS Code, I can press the button and it produces a drop-down menu. Toolbars are for actions, not pure hover targets: as a user, I would think something is broken if I have a toolbar button with no function. In VS Code, as soon as you have more than one terminal, you get a "tab bar" with an entry for each open terminal. Since we open a new tab for each terminal, I think showing the descriptions as a hover on the tab would be more logical. I think with the "enhancecd preview" support for tabs we have, this should be possible. WDYT?

Thanks for the review, I will have a look at the comments. 

Regarding the Preview:
I had a look at the enhanced preview initially and imo the problem is that it just uses the widget title's label and caption, which are both strings: https://github.com/eclipse-theia/theia/blob/628d6d22a38956a7e08497cb64f8f53d5b3c47c1/packages/core/src/browser/shell/tab-bars.ts#L500 
The env var collection's description may be a markdownstring however. Is there a way to transform this to a regular string while keeping the formatting?

We could probably try the visualPreview, but this seems hacky as we are showing something different that the visual preview

tsmaeder:
Why not just do the same thing as for the visual preview? Test whether the Widget supports rendering an "enhanced" title hover and if not fall back to the current code? We could then implement the protocol for the TerminalWidget.

packages/plugin-ext/src/common/plugin-api-rpc.ts
```diff
@@ -404,7 +404,8 @@ export interface TerminalServiceMain {
      */
     $disposeByTerminalId(id: number, waitOnExit?: boolean | string): void;
 
-    $setEnvironmentVariableCollection(extensionIdentifier: string, persistent: boolean, collection: SerializableEnvironmentVariableCollection | undefined): void;
+    $setEnvironmentVariableCollection(extensionIdentifier: string, persistent: boolean, collection: SerializableEnvironmentVariableCollection | undefined,
```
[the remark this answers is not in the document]
jfaltermeier: I am not sure if I understand this, because there is no `SerializableEnvironmentCollection`. Are you referring to `SerializableEnvironmentVariableCollection`. This type is also used used in `SerializableExtensionEnvironmentVariableCollection`, which is actually persisted using the storageService. So I think changing this interface would break reading existing data. 

But I guess we could add the `description` to `SerializableExtensionEnvironmentVariableCollection` and use `$setEnvironmentVariableCollection(persistent: boolean, collection: SerializableExtensionEnvironmentVariableCollection): void;` as a signature, if this does not count as an API break.

packages/terminal/src/browser/terminal-info-toolbar-item.tsx
jfaltermeier: Not that I am aware of. I think the reusable functionality is all in the `HoverService`. However we have to call `hoverService.requestHover` from somewhere. Here we have drawn the `div` over which the hover will appear ourself and the `onMouseEnter` only creates the custom hover ui and then calls the Hover Service. I don't think we could omit much of this.

packages/terminal/src/browser/terminal-info-toolbar-item.tsx
jfaltermeier: When adapting the enhanced preview this definitely makes sense. So I will change this.

packages/plugin-ext/src/common/plugin-api-rpc.ts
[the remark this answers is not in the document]
jfaltermeier: I've went with `$setEnvironmentVariableCollection(persistent: boolean, collection: SerializableExtensionEnvironmentVariableCollection): void` in the updated PR. Please let me know what you think.

packages/terminal/src/browser/base/terminal-widget.ts
jfaltermeier: Went with `envVarCollectionDescriptionsByExtension`

packages/terminal/src/browser/terminal-frontend-contribution.ts
jfaltermeier: I've added the description to `SerializableExtensionEnvironmentVariableCollection` now.

packages/terminal/src/browser/terminal-info-toolbar-item.tsx
jfaltermeier: Code removed because of usage of enhanced preview

jfaltermeier:
I've addressed the comments in some follow up commits. The enhanced preview looks like this atm:
Peek 2023-08-24 09-44

packages/core/src/browser/widgets/enhanced-preview-widget.ts
```diff
@@ -0,0 +1,27 @@
...
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { isFunction, isObject } from '../../common';
+
+export interface EnhancedPreviewWidget {
+    getEnhancedPreviewNode(): Node | undefined;
```
tsmaeder, line 20: Can this be any node or does it have to be a HTMLELement``?

tsmaeder (commented):
Just one more question.

packages/core/src/browser/widgets/enhanced-preview-widget.ts
jfaltermeier: We use it here: https://github.com/eclipse-theia/theia/blob/bab13b9505905c40757c17bdfb7e95932d72b30d/packages/core/src/browser/shell/tab-bars.ts#L512C15-L512C15
`appendChild` accepts any `Node`: https://developer.mozilla.org/en-US/docs/Web/API/Node/appendChild

tsmaeder: I know that the typing works out, the question is: does it make sense to ever return a node that's not a HTMLElement?

jfaltermeier: `document.createTextNode()` is not an `HTMLElement` as far as I know. I don't see why we should restrict this. 
Moreover it is consistent with the visual previews interface which uses `Node | undefined` as well.

tsmaeder (approved):

Issue #12840, opened 2023-08-18
module load redirection logic in @theia/plugin-ext-vscode misbehaves in the presence of a symlink


jcortell68:
This happens only in the development environment of someone writing a VS Code extension for Theia. I don't think it happens in a deployed installation of the app, since I don't think there's a symlink at play there. The Theia documentation recommends the use of a `plugin` symlink in the monorepo. See the recommendation [here](https://theia-ide.org/docs/authoring_vscode_extensions/#developing-vs-code-extensions-in-a-theia-project).

### Bug Description:
The problem is in the super simple implementation of the local utility function `findPlugin` [here](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext-vscode/src/node/plugin-vscode-init.ts#L78)

That module keeps a cache of API objects--one for each VS Code extension the Theia app loads. The idea is that if VS Code extension XYZ has 50 modules which each do, e.g., 

```
import * as vscode from 'vscode';
```
all 50 imports will end up getting the same (cached) module exports object.

The problem is that `findPlugin` is too simple and doesn't consider the duplicity that can happen when symlinks are involved. The `Plugin.pluginFolder` uses the symlink; however the `parent` object passed into the `module._load()` function does not use the symlink. 

By example:
```
Plugin.pluginFolder=C:\git\mytheiaapp\plugins\myext
parent.filename=C:\git\mytheiaapp\myext\dist\extension.js
```
Note that `plugins` is not present in the latter, thus `findPlugin` returns `undefined`.

I fear there might be functional problems that results from this. Consider that the API factory takes a `Plugin` object. And if we look at [the factory implementation](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/plugin-context.ts#L290), we see the plugin argument is used to create many parts of the API. When the match described above fails due to the symlink, the extension is given a default singleton API object object that is created using `emptyPlugin`, thus lacking context that may be necessary for the vscode API implementation to work correctly. And in fact, I wonder if that's not the root cause of https://github.com/eclipse-theia/theia/issues/12828

### Steps to Reproduce:
```
git clone -b vscode_extension https://github.com/jcortell68/theiasandbox
cd theiasandbox
yarn
yarn start:electron
```
Invoke the Hello World command contribution.

In the stdout, you'll see:

```
[...]

PR #12841, opened 2023-08-18
Fix symlinks should be resolved when initializing plugins

vatsal-uppal-1997:
#### What it does

Fixes: #12840 

#### How to test

Follow the steps mentioned in #12840 to reproduce the issue

To test the fix I did a quick and dirty approach where I compiled "plugin-vscode-init.ts" in my fix repo and replaced that file in the issue repo's node_modules. The process would like as follows:

- In the fix repo do a "yarn build" inside the "plugin-ext-vscode" directory
- Copy the "plugin-ext-vscode\lib\node\plugin-vscode-init.js" and paste it in the bug repositories node_modules "node_modules\\@theia\plugin-ext-vscode\lib\node"

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext-vscode/src/node/plugin-vscode-init.ts
+++ b/packages/plugin-ext-vscode/src/node/plugin-vscode-init.ts
@@ -19,7 +19,6 @@
 import * as theia from '@theia/plugin';
 import { BackendInitializationFn, PluginAPIFactory, Plugin, emptyPlugin } from '@theia/plugin-ext';
 import { VSCODE_DEFAULT_API_VERSION } from '../common/plugin-vscode-types';
-import { realpathSync } from 'fs';
 
 process.env['VSCODE_PID'] = process.env['THEIA_PARENT_PID'];
 
@@ -31,7 +30,7 @@ let pluginApiFactory: PluginAPIFactory;
 
 export const doInitialization: BackendInitializationFn = (apiFactory: PluginAPIFactory, plugin: Plugin) => {
     pluginsApiImpl.set(plugin.model.id, createVSCodeAPI(apiFactory, plugin));
-    plugins.push({ ...plugin, pluginFolder: realpathSync(plugin.pluginFolder) });
+    plugins.push(plugin);
     pluginApiFactory = apiFactory;
 
     if (!isLoadOverride) {
--- a/packages/plugin-ext-vscode/src/node/plugin-vscode-init.ts
+++ b/packages/plugin-ext-vscode/src/node/plugin-vscode-init.ts
@@ -19,6 +19,7 @@
 import * as theia from '@theia/plugin';
 import { BackendInitializationFn, PluginAPIFactory, Plugin, emptyPlugin } from '@theia/plugin-ext';
 import { VSCODE_DEFAULT_API_VERSION } from '../common/plugin-vscode-types';
+import { realpathSync } from 'fs';
 
 process.env['VSCODE_PID'] = process.env['THEIA_PARENT_PID'];
 
@@ -30,7 +31,7 @@ let pluginApiFactory: PluginAPIFactory;
 
 export const doInitialization: BackendInitializationFn = (apiFactory: PluginAPIFactory, plugin: Plugin) => {
     pluginsApiImpl.set(plugin.model.id, createVSCodeAPI(apiFactory, plugin));
-    plugins.push(plugin);
+    plugins.push({ ...plugin, pluginFolder: realpathSync(plugin.pluginFolder) });
     pluginApiFactory = apiFactory;
 
     if (!isLoadOverride) {
--- a/packages/plugin-ext/src/hosted/node/scanners/backend-init-theia.ts
+++ b/packages/plugin-ext/src/hosted/node/scanners/backend-init-theia.ts
@@ -17,7 +17,6 @@
 import * as theia from '@theia/plugin';
 import { BackendInitializationFn } from '../../../common/plugin-protocol';
 import { PluginAPIFactory, Plugin, emptyPlugin } from '../../../common/plugin-api-rpc';
-import { realpathSync } from 'fs';
 
 const pluginsApiImpl = new Map<string, typeof theia>();
 const plugins = new Array<Plugin>();
@@ -30,7 +29,7 @@ export const doInitialization: BackendInitializationFn = (apiFactory: PluginAPIF
     const apiImpl = apiFactory(plugin);
     pluginsApiImpl.set(plugin.model.id, apiImpl);
 
-    plugins.push({ ...plugin, pluginFolder: realpathSync(plugin.pluginFolder) });
+    plugins.push(plugin);
     pluginApiFactory = apiFactory;
 
     if (!isLoadOverride) {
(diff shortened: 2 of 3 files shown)


martin-fleck-at (changes requested):
Hi @vatsal-uppal-1997, thank you very much for that change! I had a look at it and it works as expected given the example from the bug report.

My only concern is that I'm thinking about applying this fix a bit earlier as otherwise we need to apply it to all our custom API handling, i.e., we should at least also apply it to `backend-init-theia.ts` where the same mechanism is applied as in `plugin-vscode-init.ts`.

Just to give a bit of context: We have different plugin resolvers (`PluginDeployerResolver`) that can add plugins to a resolution context (`PluginDeployerResolverContext`) where they are transformed to plugin entries (`PluginDeployerEntry`) that are further modified by the file and directory handlers (`PluginDeployerDirectoryHandler` and `PluginDeployerFileHandler`). Those handlers in particular can still modify the path of the entry. Afterwards, we read the manifest from each plugin using the final path which ends up in the `Plugin` that you are adapting. I believe another good point to do apply this fix would be during the translation from the `pluginPath` to the `PluginPackage`in the `HostedPluginReader.readPackage` method. I did some initial tests and it all looks good to me. @tsmaeder Do you have an opinion on this?

tsmaeder:
@martin-fleck-at the existence of better fix should not prevent us from merging a PR in general. The questions we ask should be: "does this make the code base better" and "does this prevent us from doing anything in the future". Why not merge it and then move it to a different place in a follow-up?

packages/plugin-ext-vscode/src/node/plugin-vscode-init.ts
```diff
@@ -19,6 +19,7 @@
 import * as theia from '@theia/plugin';
 import { BackendInitializationFn, PluginAPIFactory, Plugin, emptyPlugin } from '@theia/plugin-ext';
 import { VSCODE_DEFAULT_API_VERSION } from '../common/plugin-vscode-types';
+import { realpathSync } from 'fs';
```
kittaakos, line 22: Please do not use blocking fs call.

kittaakos (changes requested):

msujew:
@vatsal-uppal-1997 Do you plan on getting back to this to integrate the review comments? Otherwise, I'll close this PR in a few days.

vatsal-uppal-1997:
> @vatsal-uppal-1997 Do you plan on getting back to this to integrate the review comments? Otherwise, I'll close this PR in a few days.

Hi @msujew,

Sorry it has been busy for me for quite a while, I'll try to look into this PR this weekend. Thanks for being patient!

packages/plugin-ext-vscode/src/node/plugin-vscode-init.ts
vatsal-uppal-1997: The `doInitialization` function where `realpathSync` is being called is a synchronous function, as such we would want it to error out in case `realpathSync` fails for some reason. An async implementation can possibly lead to a situation where `realpathSync` will fail silently whereas the call to `doInitialization` would succeed

msujew: I believe the sync vs async issue should be resolved together with the issue @martin-fleck-at brought up.

vatsal-uppal-1997:
Hi @martin-fleck-at,

I've made the changes as per your suggestions. Requesting you to please review them to see if they're satisfactory.

(++ @msujew )

packages/plugin-ext-vscode/src/node/plugin-vscode-init.ts
vatsal-uppal-1997: @msujew Thanks for your comment! I've made the changes as suggested by @martin-fleck-at and am using sync calls for the `doInitialization` methods and async one for the `readPackage` method. Hope that resolves this

packages/plugin-ext/src/hosted/node/scanners/backend-init-theia.ts
```diff
@@ -29,7 +30,7 @@ export const doInitialization: BackendInitializationFn = (apiFactory: PluginAPIF
     const apiImpl = apiFactory(plugin);
     pluginsApiImpl.set(plugin.model.id, apiImpl);
 
-    plugins.push(plugin);
+    plugins.push({ ...plugin, pluginFolder: realpathSync(plugin.pluginFolder) });
```
msujew, line 33: Suggestion: I don't believe this is necessary anymore, given that we already store the resolved plugin path in the `plugin.pluginFolder`.

packages/plugin-ext-vscode/src/node/plugin-vscode-init.ts
```diff
@@ -30,7 +31,7 @@ let pluginApiFactory: PluginAPIFactory;
 
 export const doInitialization: BackendInitializationFn = (apiFactory: PluginAPIFactory, plugin: Plugin) => {
     pluginsApiImpl.set(plugin.model.id, createVSCodeAPI(apiFactory, plugin));
-    plugins.push(plugin);
+    plugins.push({ ...plugin, pluginFolder: realpathSync(plugin.pluginFolder) });
```
msujew, line 34: Suggestion: I don't believe this is necessary anymore, given that we already store the resolved plugin path in the `plugin.pluginFolder`.

vatsal-uppal-1997:
Hi @msujew,

Thanks for your suggestions! I've removed the unnecessary changes from the files you pointed out.

I also tested the change and it seems that I only needed to modify the `readPackage` method of `HostedPluginReader`

martin-fleck-at (approved):
@vatsal-uppal-1997 Thank you for the update! The change looks good to me now and the problem seems to be resolved. If everyone else is also happy with this, we can merge it.

msujew (approved):
Looks good to me as well 👍

kittaakos:
Can somebody please explain why the https://github.com/eclipse-theia/theia/pull/12841#discussion_r1341981869 was marked as resolved, but eventually, the changes got into the `master` with a sync FS call?

martin-fleck-at:
@kittaakos I'm not sure I follow, the merged commit (https://github.com/eclipse-theia/theia/commit/525b3b11ebc42adc986acb0e5c1b33513e5e4ad5) uses an async realpath call to properly set the `packagePath` which is why this is then no longer needed later on.

vatsal-uppal-1997:
Hi @kittaakos,

I had removed the sync calls based on your and @msujew comment in this [commit](https://github.com/eclipse-theia/theia/commit/d0ed4e741b4fee18425b11ebaceeb848bf775465)

kittaakos:
> @kittaakos I'm not sure I follow, the merged commit ([525b3b1](https://github.com/eclipse-theia/theia/commit/525b3b11ebc42adc986acb0e5c1b33513e5e4ad5)) uses an async realpath call to properly set the `packagePath` which is why this is then no longer needed later on.



> Hi @kittaakos,
> 
> I had removed the sync calls based on your and @msujew comments in this [commit](https://github.com/eclipse-theia/theia/commit/d0ed4e741b4fee18425b11ebaceeb848bf775465)

You're both great!  Thanks! I confirm I was wrong; the sync call was not part of the merged PR.


----

I checked the PR today; I did not see the sync call. I happily put my 👍 to https://github.com/eclipse-theia/theia/pull/12841#pullrequestreview-1652247819. Then I rechecked the PR, and I saw the sync call. I was either blind (most likely), or a cached state had loaded in my browser.

Issue #12852, opened 2023-08-22
Run Theia without a backend


sdirix:
![image](https://github.com/eclipse-theia/theia/assets/8998368/a7044401-2bf5-4da5-bf86-67c2f970a768)

# tl;dr;
We would like to enable a new use case for Theia applications: Support for "browser-only" Theia applications which can be executed without a backend.

Our vision is to support Theia "browser-only" applications as a first class citizen, next to the "browser" and "electron" targets. For this we adapted the Theia CLI tooling to support the new `browser-only` target and `frontendOnly` Theia extensions. Using this tooling the feature can be supported in a minimally invasive way, requiring none (or little to none) changes to existing Theia based applications.

# Check it out yourself

You can find our PR here. Every review is appreciated!

Because it's so easy to host static sites, we deployed three example applications for you to check out:
 - [Minimal browser-only application](https://minimal--browseronly-theia.netlify.app/)
 - [Including all Theia packages](https://all--browseronly-theia.netlify.app/)
 - [POC for VS Code web extension support](https://plugins--browseronly-theia.netlify.app/)
   - You can execute a `hello world` command implemented by a Theia plugin
   - By opening a Typescript file you can see the Typescript VS Code web extension at work
   - You can clone Github repositories via the `File > Import repository` action. Note that this uses the open proxy [cors.isomorphic-git.org](https://cors.isomorphic-git.org/) so it's pretty slow. This proxy is freely available so please be considerate with its usage.
   - **Note that this POC is not optimized, so it takes a while to start.**

# Motivation
Support of browser-only use cases is a value adding feature for the whole Theia ecosystem. While the existing Theia framework offers large flexibility and therefore supports a wide array of use cases, it requires the deployment and execution of a Node based backend. This often comes with a lot of operational effort and additional costs.

There are many use cases in which not the full feature set of Theia based applications are required. For example, demo and quick start applications often come with hard coded examples anyway, do not require persistent storage and can work without terminal support.

This feature is requested from time to time in the community too:
 - https://github.com/eclipse-theia/theia/issues/8368
 - https://github.com/eclipse-theia/theia/issues/6732
 - https://github.com/eclipse-theia/theia/issues/6560
 - https://community.theia-ide.org/t/does-theia-work-without-any-backend/46
 - https://community.theia-ide.org/t/would-it-be-possible-to-create-a-backend-less-theia/1545
[...]

PR #12853, opened 2023-08-22
Support 'browser-only' Theia

sdirix:
image

Adds tooling, adapts the current code base and provides an example application for the new 'browser-only' Theia application target. This target generates a Theia frontend application which can run without a backend, transforming the Theia application to a static site.

Implements #12852

#### What it does
Adapts tooling to:
- support new 'browser-only' application target
- support new 'frontendOnly' and 'frontendOnlyPreload' module declaration
- use http-server when starting 'browser-only' applications

Replaces backend services with browser-only variants
- implementation of BrowserFS-based filesystem
- implementation of 'ServiceConnectionProvider' which timeouts for all requests, enabling loading all Theia packages

Adds a browser-only example application and api-samples showcasing the customization of the new BrowserFS-based filesystem.

#### How to test

You can also find a hosted variant [here](https://all--browseronly-theia.netlify.app/).

Note that you can build the 'browser-only' Theia even without building the Node native parts via:
```
yarn install --ignore-scripts
yarn browser-only prepare:no-native
yarn browser-only build
```

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-manager/src/application-package-manager.ts
+++ b/dev-packages/application-manager/src/application-package-manager.ts
@@ -119,8 +119,30 @@ export class ApplicationPackageManager {
     start(args: string[] = []): cp.ChildProcess {
         if (this.pck.isElectron()) {
             return this.startElectron(args);
+        } else if (this.pck.isBrowserOnly()) {
+            return this.startBrowserOnly(args);
         }
         return this.startBrowser(args);
+    }
+
+    startBrowserOnly(args: string[]): cp.ChildProcess {
+        const { command, mainArgs, options } = this.adjustBrowserOnlyArgs(args);
+        return this.__process.spawnBin(command, mainArgs, options);
+    }
+
+    adjustBrowserOnlyArgs(args: string[]): Readonly<{ command: string, mainArgs: string[]; options: cp.SpawnOptions }> {
+        let { mainArgs, options } = this.adjustArgs(args);
+
+        // first parameter: path to generated frontend
+        // second parameter: disable cache to support watching
+        mainArgs = ['lib/frontend', '-c-1', ...mainArgs];
+
+        const portIndex = mainArgs.findIndex(v => v.startsWith('--port'));
+        if (portIndex === -1) {
+            mainArgs.push('--port=3000');
+        }
+
+        return { command: 'http-server', mainArgs, options };
     }
 
     startElectron(args: string[]): cp.ChildProcess {
--- a/dev-packages/cli/src/theia.ts
+++ b/dev-packages/cli/src/theia.ts
@@ -94,12 +94,12 @@ function rebuildCommand(command: string, target: ApplicationProps.Target): yargs
 }
 
 function defineCommonOptions<T>(cli: yargs.Argv<T>): yargs.Argv<T & {
-    appTarget?: 'browser' | 'electron'
+    appTarget?: 'browser' | 'electron' | 'browser-only'
 }> {
     return cli
         .option('app-target', {
             description: 'The target application type. Overrides `theia.target` in the application\'s package.json',
-            choices: ['browser', 'electron'] as const,
+            choices: ['browser', 'electron', 'browser-only'] as const,
         });
 }
 
--- a/package.json
+++ b/package.json
@@ -62,8 +62,9 @@
   "scripts": {
     "all": "yarn -s install && yarn -s lint && yarn -s build",
     "browser": "yarn -s --cwd examples/browser",
+    "browser-only": "yarn -s --cwd examples/browser-only",
     "build": "yarn -s compile && yarn -s build:examples",
-    "build:examples": "yarn browser build && yarn electron build",
+    "build:examples": "yarn browser build && yarn electron build && yarn browser-only build",
     "clean": "yarn -s rebuild:clean && yarn -s lint:clean && node scripts/run-reverse-topo.js yarn -s clean",
     "compile": "echo Compiling TypeScript sources... && yarn -s compile:clean && yarn -s compile:tsc",
     "compile:clean": "ts-clean dev-packages/* packages/*",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 - [plugin] stub multiDocumentHighlightProvider proposed API [#13248](https://github.com/eclipse-theia/theia/pull/13248) - contributed on behalf of STMicroelectronics
 - [terminal] update terminalQuickFixProvider proposed API according to vscode 1.85 version [#13240](https://github.com/eclipse-theia/theia/pull/13240) - contributed on behalf of STMicroelectronics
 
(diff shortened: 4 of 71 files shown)


package.json
```diff
@@ -62,14 +62,15 @@
...
-    "build:examples": "yarn browser build && yarn electron build",
+    "build:examples": "yarn browser build && yarn electron build && yarn browser-only build",
     "clean": "yarn -s rebuild:clean && yarn -s lint:clean && node scripts/run-reverse-topo.js yarn -s clean",
     "compile": "echo Compiling TypeScript sources... && yarn -s compile:clean && yarn -s compile:tsc",
     "compile:clean": "ts-clean dev-packages/* packages/*",
     "compile:references": "node scripts/compile-references.js",
     "compile:tsc": "tsc -b",
     "docs": "rimraf gh-pages/docs/next && typedoc --tsconfig tsconfig.json --options configs/typedoc.json .",
     "download:plugins": "theia download:plugins",
+    "browser-only": "yarn -s --cwd examples/browser-only",
```
msujew, line 73: Minor: We should move this after the `browser` script. I almost overlooked it during testing.

dev-packages/application-manager/src/application-package-manager.ts
```diff
@@ -117,10 +117,33 @@ export class ApplicationPackageManager {
...
+
+        const portIndex = mainArgs.findIndex(v => v.startsWith('--port'));
+        if (portIndex === -1) {
+            mainArgs.push('--port=3000');
+        }
+        const configIndex = mainArgs.findIndex(v => v.startsWith('--config') || v.startsWith('-c'));
+        if (configIndex === -1) {
+            mainArgs.push('--config=webpack.config.js');
+        }
+        return { command: 'webpack', mainArgs, options };
```
msujew, line 144: Discussion: We don't use `webpack serve` for any other configuration. I'm not totally against it, I just didn't expect to see it here. I would've expected a lighter solution via something like `http-server`. Especially since we have `watch` and `build` scripts in our `browser-only` app, which shouldn't be needed with `webpack serve`, right?

Note that I've noticed this in particular, because `webpack serve` returns `Invalid Host header` when trying to access the website using Gitpod for some reason.

msujew (commented):
This works really well. Good job 👍 

I have a few remarks and questions on some of the changes here though.

dev-packages/cli/src/theia.ts
```diff
@@ -180,6 +180,7 @@ async function theiaCli(): Promise<void> {
         .command(rebuildCommand('rebuild', target))
         .command(rebuildCommand('rebuild:browser', 'browser'))
         .command(rebuildCommand('rebuild:electron', 'electron'))
+        .command(rebuildCommand('rebuild:browser-only', 'browser-only'))
```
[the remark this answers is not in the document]
sdirix: You are right, actually we don't need to build them at all as I even indicated in the PR description ;)

dev-packages/application-manager/src/application-package-manager.ts
sdirix: You are right. `http-server` is much more lightweight and should serve fine here!

sdirix:
We're currently working on updating the branch on the latest state of `master` as well as incorporating the feedback of @msujew. I'll ping you once it's ready to be reviewed again.

packages/core/src/browser/messaging/ws-connection-provider.ts
```diff
@@ -31,8 +31,14 @@ export interface WebSocketOptions {
...
 
+export const WebSocketConnectionProvider = Symbol('WebsocketConnectionProvider');
+export interface WebSocketConnectionProvider extends ConnectionProvider<WebSocketOptions> {
+    onSocketDidOpen: Event<void>
+    onSocketDidClose: Event<void>
+};
+
 @injectable()
-export class WebSocketConnectionProvider extends AbstractConnectionProvider<WebSocketOptions> {
+export class WebSocketConnectionProviderImpl extends AbstractConnectionProvider<WebSocketOptions> implements WebSocketConnectionProvider {
```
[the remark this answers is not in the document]
sdirix: I removed this breaking change completely. It's possible to accomplish our goal without it too.

sdirix:
Hi @msujew,
Thanks again for your review! I updated the PR to the latest state of Theia and integrated all your review comments.

The license check complains about `browserfs` although it's MIT licensed. Do you know why this is the case and how we can resolve the issue?

msujew:
@sdirix I had to trigger the license harvesting manually on `clearlydefined.io`. It's now correctly recognized 👍

JonasHelming:
@msujew could we maybe get this in for January?

paulalexwilson:
The landscape of theia plugin types and deployment modes is feeling somewhat complex from a relative outsider looking in:

- Theia Extensions
- Theia Plugins (deprecated?)
- VS Code Extensions
- Headless (Daemon?) Plugins
- FrontendOnly Mode

It would be great if this PR also added some statement around how this interacts with existing plugin/extension types and what limitations exist.

msujew (approved):
Great, looks good to me. There is one minor issue that I've noticed in the browser-only file system implementation:

1. Create an untitled file
2. Save it into the workspace
3. The navigator doesn't update

I think this is minor enough so we can go ahead to merge this anyway.

msujew:
@paulalexwilson I don't think that the documentation change makes sense in the scope of this PR. Feel free to create a separate issue though, and someone might pick it up. Note that in general, the browser-only Theia version is supposed to eventually support all features of Theia that are possible without a dedicated backend (which is basically everything except for running a terminal). See https://github.com/eclipse-theia/theia/issues/12852 for open tasks and further discussion.

sdirix:
I rebased on latest `master` to get rid off the conflict in the Changelog. No further changes were done.

erlmachinedev:
@sdirix Thanks for the valuable contribution 👍 I just tested the feature from 56cf3dbe76693735bfef8b8ac0ec424c38567931 but I see no syntax highlighting for `Lua`, `json`, etc..

Probably the default language setup should work (even without extensions). The file icon type is fetched properly and indicated at the explorer tab but the Monaco editor seems doesn't  highlight syntax yet.


Screen Shot 2024-01-21 at 10 23 35 PM
Screen Shot 2024-01-21 at 10 24 38 PM

The same issue is present on the demo app https://minimal--browseronly-theia.netlify.app/#

P.S. If there is a need to attach dedicated issue (supported by reproduction steps), please let me know.

msujew:
@erlmachinedev Given that all syntax highlighting is currently exclusively contributed by vscode extensions, this is expected behavior. This will work as usual once extension support will be added. I think we won't support dedicated syntax highlighting for the browser-only mode, as it's just as much work as adding real support for vscode extensions.

sdirix:
Hi @erlmachinedev,
Thanks for your interest in "browser-only" Theia. This feature is currently in a MVP state: It can already be used to build applications, however there are many missing features compared to regular Theia. I updated the original issue (#12852) with the current state and a list of missing features.

PR #12861, opened 2023-08-24
Codicon color and URI support to `TerminalOptions`

### LINKED ISSUE #12074: vscode: Missing partial icon and color API support for ExtensionTerminalOptions and TerminalOptions
With #11504, the added API was partially stub. Currently, only a ThemeIcon is supported, while the vscode API provides also support for URI and {dark: URI, liught: URI}. Also the color icon support has been stubbed, as the current implementation of the title widget as terminal title does not let an easy customization of the icon color.

FernandoAscencio:
#### What it does
Closes #12074

- This commit adds URI and { light, dark } support for terminal icons.
- This commit adds ColorTheme support for ThemeIcons
- Support for `ExtensionTerminalOptions#iconPath`

#### How to test
1. Open Theia
2. Install from VSIX [the test extension](https://github.com/eclipse-theia/theia/files/12434150/terminalcreationtests-0.0.1.zip)
3. Check the result follow VSCode test results seen below:

VSCode test Results
Dark
TerminalIconTests
Light
TerminalIconTestsLight

###### Wishlist or Missing
- Manage color change for URI icons
  - Possible implementation using CSS filter property with possible issues with older browsers.
- Manage other color formats aside from default `terminal.ansi**color**`

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin/src/theia.d.ts
+++ b/packages/plugin/src/theia.d.ts
@@ -3140,7 +3140,7 @@ export module '@theia/plugin' {
         /**
          * The icon path or {@link ThemeIcon} for the terminal.
          */
-        iconPath?: ThemeIcon;
+        iconPath?: string | ThemeIcon | Uri | { light: Uri | string, dark: Uri | string };
 
         /**
          * The icon {@link ThemeColor} for the terminal.
@@ -3261,7 +3261,7 @@ export module '@theia/plugin' {
         /**
          * The icon path or {@link ThemeIcon} for the terminal.
          */
-        iconPath?: ThemeIcon;
+        iconPath?: string | ThemeIcon | Uri | { light: Uri | string, dark: Uri | string };
 
         /**
          * The icon {@link ThemeColor} for the terminal.
--- a/packages/terminal/src/browser/base/terminal-widget.ts
+++ b/packages/terminal/src/browser/base/terminal-widget.ts
@@ -178,9 +178,9 @@ export interface TerminalWidgetOptions {
     readonly title?: string;
 
     /**
-     * icon class
+     * icon class with or without color modifier
      */
-    readonly iconClass?: string;
+    readonly iconClass?: string | {icon: string, color: string};
 
     /**
      * Path to the executable shell. For example: `/bin/bash`, `bash`, `sh`.
--- a/packages/terminal/src/browser/terminal-widget-impl.ts
+++ b/packages/terminal/src/browser/terminal-widget-impl.ts
@@ -19,7 +19,7 @@ import { FitAddon } from 'xterm-addon-fit';
 import { inject, injectable, named, postConstruct } from '@theia/core/shared/inversify';
 import { ContributionProvider, Disposable, Event, Emitter, ILogger, DisposableCollection, Channel, OS } from '@theia/core';
 import {
-    Widget, Message, WebSocketConnectionProvider, StatefulWidget, isFirefox, MessageLoop, KeyCode, codicon, ExtractableWidget, ContextMenuRenderer
+    Widget, Message, WebSocketConnectionProvider, StatefulWidget, isFirefox, MessageLoop, KeyCode, ExtractableWidget, ContextMenuRenderer
 } from '@theia/core/lib/browser';
 import { isOSX } from '@theia/core/lib/common';
 import { WorkspaceService } from '@theia/workspace/lib/browser';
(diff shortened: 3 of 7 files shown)


packages/plugin/src/theia.d.ts
```diff
@@ -3140,7 +3140,7 @@ export module '@theia/plugin' {
         /**
          * The icon path or {@link ThemeIcon} for the terminal.
          */
-        iconPath?: ThemeIcon;
+        iconPath?: string | ThemeIcon | Uri | { light: Uri | string, dark: Uri | string };
```
msujew, line 3143: Suggestion: In vscode, both of these `iconPath` definitions are defined as:

```ts
iconPath?: Uri | { light: Uri; dark: Uri } | ThemeIcon;
```

Currently, both of them are marked as `partial` by the comparator report.

packages/terminal/src/browser/base/terminal-widget.ts
```diff
@@ -178,9 +178,9 @@ export interface TerminalWidgetOptions {
     readonly title?: string;
 
     /**
-     * icon class
+     * icon class with or without color modifier
      */
-    readonly iconClass?: string;
+    readonly iconClass?: string | {icon: string, color: string};
```
msujew, line 183: Suggestion: Since this is a breaking change anyway, would it make sense to type this differently? This would make dealing with this value much more straightforward. We should also preferably export the underlying type.

```ts
export interface TerminalIcon {
    icon: string;
    color?: string;
}
```

```suggestion
    readonly iconClass?: TerminalIcon;
```

msujew (commented):
Functionality wise this looks pretty good to me. I have a few suggestions/improvements to the code, see below.

JonasHelming:
@FernandoAscencio Do you plan to address the comments from Mark?

vince-fugnitto:
> @FernandoAscencio Do you plan to address the comments from Mark?

As an FYI, Fernando's internship ended and he's back to being a full-time student. Someone I assume would need to pick up the changes, and address the comments.

FernandoAscencio:
@msujew 
Those are good suggestions. Unfortunately, at the moment I am unable to address the changes and test the results. Unfortunately, I am swamped with schoolwork and no longer employed at Ericsson. So my time to work on this project has been push to the back.

@JonasHelming 
I apologize for the late response. As Vince mentioned, I am no longer employed and thus my involvement in this project has been pushed to low priority below university projects.

packages/plugin/src/theia.d.ts
FernandoAscencio: Does this mean that the implementation in this PR is a partial implementation over what is done in VS Code?

I was having trouble implementing the vscode definition for `iconPath` and change the icon addressing from URI to hosted address. Originally, `iconPath` uses `ThemeIcon` only for this reason. `ThemeIcon` are essentially codicons and thus the source of the icon is in the application itself. Using Uri means that the source for the icon can be references "externally". In this case the browser version does not like it referencing a specific file in the computer for security reasons. So the reference address has to be of the form "hostedPlugins/plugin/address". This is different on how URI handles addresses. So at the end using a URI to pass the relative address ends causing errors.

I constrained myself to finish this PR in a couple days before the end of my internship. The current solution is not ideal, and I would prefer a better one to be used, but implementing one would have taken more time.

There are some paths that can be taken to do this:
1. Changing how URI works and are defined. This would also streamline a bit of the process in Themables for the git plugin. (Another PR I worked on).
2. Change how the icon addressing information is passed so that you can pass a "hostedPlugin" address instead of just URI or ThemeIcons.

packages/terminal/src/browser/base/terminal-widget.ts
FernandoAscencio: I like this and it should be an easy enough implementation.

packages/terminal/src/browser/terminal-widget-impl.ts
```diff
@@ -322,6 +332,28 @@ export class TerminalWidgetImpl extends TerminalWidget implements StatefulWidget
         return value === 'line' ? 'bar' : value;
     }
 
+    private isTerminalAnsiColor(color: string): boolean {
```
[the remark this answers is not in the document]
FernandoAscencio: I agree, should be easy enough.

JonasHelming:
@vladarama Maybe you want to take a look at this? (@vince-fugnitto FYI)

dhuebner:
@JonasHelming @vince-fugnitto 
An old and a nice PR.
I could look into it if nobody want to...
I just wonder how to take the changes over to keep the original author information.

JonasHelming:
Great! If the final commit references both of you as authors, that would be fine.

dhuebner:
@JonasHelming 
Please assign me to https://github.com/eclipse-theia/theia/issues/12074 then.

JonasHelming:
@dhuebner Sure, could you comment on the ticket, please, otherwise I cannot assign it to you (GitHub doesn't let me)

PR #12897, opened 2023-09-12
Open initial electron window early and reuse it later

planger:
#### What it does

This avoids having a rather long pause from Theia start until something is visibly coming up. In my initial experiment, I observe the empty window showing up ~1.5s earlier than usual.

This behavior can be controlled via the application config in the application's package.json. For instance, to turn off, use:

```
  "theia": {
    "target": "electron",
    "frontend": {
      "config": {
        "applicationName": "Theia Electron Example",
        "electron": {
          "showWindowEarly": false
        }
      }
    }
  }
```

To mitigate flickering, we set a background color of the window, which shows before any page is loaded, according to the expected background color of the Theia app. Also, we store the background color of the current color theme to be able to show the correct background color in the initial window.

In future changes, we may consider loading a very simple static web page mimicking the basic workbench structure before we switch to the actual application and in the Theia app don't show a spinner but a loading indicator on top of the basic workbench structure. With that, the application rather builds up visually instead of flickering between nothing, the spinner and the actual app.

##### Without this change

https://github.com/eclipse-theia/theia/assets/588090/289051fa-9862-4651-ba0c-52faaa447b11

##### With this change

https://github.com/eclipse-theia/theia/assets/588090/91f9d7b8-05c9-4e10-ba03-63ff662d1bcf

#### How to test

Build the example electron app with `"showWindowEarly": false` and with `"showWindowEarly": true` to compare the differences.

In addition, it is worth testing the following aspects:

* Window size is correctly restored after opening the electron app
* Window layout is correctly restored after opening the electron app
* Opening a new window from the app works as before

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-browser/menu/electron-menu-contribution.ts
+++ b/packages/core/src/electron-browser/menu/electron-menu-contribution.ts
@@ -30,6 +30,8 @@ import { WindowTitleService } from '../../browser/window/window-title-service';
 
 import '../../../src/electron-browser/menu/electron-menu-style.css';
 import { MenuDto } from '../../electron-common/electron-api';
+import { ThemeService } from '../../browser/theming';
+import { ThemeChangeEvent } from '../../common/theme';
 
 export namespace ElectronCommands {
     export const TOGGLE_DEVELOPER_TOOLS = Command.toDefaultLocalizedCommand({
@@ -88,6 +90,9 @@ export class ElectronMenuContribution extends BrowserMenuBarContribution impleme
     @inject(WindowService)
     protected readonly windowService: WindowService;
 
+    @inject(ThemeService)
+    protected readonly themeService: ThemeService;
+
     @inject(CustomTitleWidgetFactory)
     protected readonly customTitleWidgetFactory: CustomTitleWidgetFactory;
 
@@ -123,6 +128,9 @@ export class ElectronMenuContribution extends BrowserMenuBarContribution impleme
             this.handleToggleMaximized();
         });
         this.attachMenuBarVisibilityListener();
+        this.themeService.onDidColorThemeChange(e => {
+            this.handleThemeChange(e);
+        });
     }
 
     protected attachWindowFocusListener(app: FrontendApplication): void {
@@ -412,6 +420,11 @@ export class ElectronMenuContribution extends BrowserMenuBarContribution impleme
         } else {
             this.shell.topPanel.hide();
         }
+    }
+
+    protected handleThemeChange(e: ThemeChangeEvent): void {
+        const backgroundColor = window.getComputedStyle(document.body).backgroundColor;
+        window.electronTheiaCore.setBackgroundColor(backgroundColor);
     }
 
 }
--- a/packages/core/src/electron-main/electron-main-application.ts
+++ b/packages/core/src/electron-main/electron-main-application.ts
@@ -15,13 +15,13 @@
 // *****************************************************************************
 
 import { inject, injectable, named } from 'inversify';
-import { screen, app, BrowserWindow, WebContents, Event as ElectronEvent, BrowserWindowConstructorOptions, nativeImage } from '../../electron-shared/electron';
+import { screen, app, BrowserWindow, WebContents, Event as ElectronEvent, BrowserWindowConstructorOptions, nativeImage, nativeTheme } from '../../electron-shared/electron';
 import * as path from 'path';
 import { Argv } from 'yargs';
 import { AddressInfo } from 'net';
 import { promises as fs } from 'fs';
 import { fork, ForkOptions } from 'child_process';
-import { FrontendApplicationConfig } from '@theia/application-package/lib/application-props';
+import { DefaultTheme, FrontendApplicationConfig } from '@theia/application-package/lib/application-props';
 import URI from '../common/uri';
 import { FileUri } from '../node/file-uri';
 import { Deferred } from '../common/promise-util';
@@ -180,9 +180,12 @@ export class ElectronMainApplication {
 
     protected _config: FrontendApplicationConfig | undefined;
     protected useNativeWindowFrame: boolean = true;
+    protected customBackgroundColor?: string;
     protected didUseNativeWindowFrameOnStart = new Map<number, boolean>();
     protected windows = new Map<number, TheiaElectronWindow>();
     protected restarting = false;
+
+    protected initialWindow?: BrowserWindow;
 
     get config(): FrontendApplicationConfig {
         if (!this._config) {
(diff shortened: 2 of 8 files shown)


kittaakos:
> This change aims to avoid having a rather long pause from Theia start until something is visibly coming up and thus intents to improve the perceived performance. Also it aims to prevent users from wondering whether the application is actually starting, as they don't have any feedback for several seconds when opening Theia.

This is a great start, but it would also be great to fix the problem instead of showing a fake window. Also, how will this work when multiple windows have to be restored when the app starts?

According to my profiling, a lot of time is spent with the backend starting and loading the plugin contributions and the languages, especially when there are many, and parsing the JSON files takes time.

It would be great to move the plugin loading to a worker. This is what it takes currently:


https://github.com/eclipse-theia/theia/assets/1405703/9386fb1b-882f-47df-867a-a4779827ac67

My initial tweak to overcome it is to avoid `awaiting` when reading the contributions: https://github.com/kittaakos/arduino-ide/commit/b78800571348c27a36536b3201bffc8f776aebe5. The backend starts earlier, and the electron window opens earlier.

packages/core/src/electron-main/electron-main-application.ts
```diff
@@ -242,6 +245,16 @@ export class ElectronMainApplication {
         return this.didUseNativeWindowFrameOnStart.get(webContents.id) ? 'native' : 'custom';
     }
 
+    protected showInitialWindow(): void {
+        if (!process.env.SKIP_INITIAL_WINDOW) {
```
kittaakos, line 249: How can one configure this when there is a final bundled app?

tsmaeder:
@kittaakos @sdirix I think we're talking about different things here: one is the _perception_ of performance and the other is effective performance. Both contribute to user satisfaction. Things like putting up a splash screen do not make loading faster, but users still want to see something is happening. Ideally, we would show some loading progress on the "early window".

JonasHelming:
I agree to all said above. @kittaakos It is worth mentioning that we are currently investigating the start up performance in general, this PR is just one piece. It seems you are working in the same area. I believe it would make sense to align a bit to share ideas and avoid duplicate work. Are you interested in this? If so, maybe you join the next dev call and maybe we can schedule a break out session there for interested parties?

kittaakos:
> Are you interested in this?

Sure. Thank you! I saved the date of the next dev call.

planger:
For references, this is the epic to track all current streams of work around start-up performance improvements: https://github.com/eclipse-theia/theia/issues/12924

packages/core/src/electron-main/electron-main-application.ts
kittaakos: Please add the property to the electron frontend app config to make this configurable downstream.

https://github.com/eclipse-theia/theia/blob/f038673dc228485b34e8762686525d8e5f258afb/dev-packages/application-package/src/application-props.ts#L37-L45

packages/core/src/electron-main/electron-main-application.ts
```diff
@@ -303,6 +316,7 @@ export class ElectronMainApplication {
         return {
             show: false,
             title: this.config.applicationName,
+            backgroundColor: nativeTheme.shouldUseDarkColors ? '#1E1E1E' : '#FFFFFF',
```
kittaakos, line 319: A similar dispatch is already at:
 https://github.com/eclipse-theia/theia/blob/f038673dc228485b34e8762686525d8e5f258afb/packages/core/src/browser/preload/theme-preload-contribution.ts#L28

Could you single-source them somewhere here? Thanks
 https://github.com/eclipse-theia/theia/blob/f038673dc228485b34e8762686525d8e5f258afb/dev-packages/application-package/src/application-props.ts#L48-L63

planger:
@kittaakos Thank you very much for the helpful review! I've incorporated the changes you suggested and also enhanced the change so that we store the user's background color in the electron store to match the theme color with the initial window's background color. Unfortunately, I had to rebase to make the changes you suggested. Sorry if that makes a re-review more cumbersome.

packages/core/src/electron-browser/menu/electron-menu-contribution.ts
```diff
@@ -30,6 +30,8 @@ import { WindowTitleService } from '../../browser/window/window-title-service';
 
 import '../../../src/electron-browser/menu/electron-menu-style.css';
 import { MenuDto } from '../../electron-common/electron-api';
+import { ThemeService } from '../../browser/theming';
+import { ThemeChangeEvent } from 'src/common/theme';
```
kittaakos, line 34:
```suggestion
import { ThemeChangeEvent } from '../../common/theme';
```

kittaakos (approved):
I verified 1d630a6ef, and it worked for me. Thank you!

Why is it required to set the window background?

packages/core/src/electron-browser/menu/electron-menu-contribution.ts
planger: Ah nice catch, thank you!

planger:
> Why is it required to set the window background?

Quite some time (500ms to 1000ms) is needed until the frontend actually starts rendering and during that time the window is just empty showing the default system background. Imho it looks faster and better to already show the empty window with the background color that matches the expected background color until the actual app is loaded to avoid flickering.

kittaakos:
> Quite some time (500ms to 1000ms) is needed until the frontend actually starts rendering and during that time the window is just empty showing the default system background. Imho it looks faster and better to already show the empty window with the background color that matches the expected background color until the actual app is loaded

I understand why it is set in the electron main. Here:

https://github.com/eclipsesource/theia/blob/bd18c66d1020897215a6b856c9435c917d880442/packages/core/src/electron-main/electron-main-application.ts#L329

But how is opening and reusing the first window requires to set the background on theme change:

https://github.com/eclipsesource/theia/blob/bd18c66d1020897215a6b856c9435c917d880442/packages/core/src/electron-browser/menu/electron-menu-contribution.ts#L131-L133

Feel free to merge it though. I only want to understand what's the catch here. Thanks

planger:
> But how is opening and reusing the first window requires to set the background on theme change:
> 
> https://github.com/eclipsesource/theia/blob/bd18c66d1020897215a6b856c9435c917d880442/packages/core/src/electron-browser/menu/electron-menu-contribution.ts#L131-L133

This is to match the background color if the user switched to a different theme. Whenever the user switches theme, we notify the Electron window and it stores the background color into the electron store, so it can be restored on next start correctly.
You can test this by switching color theme to something with a completely different background color and observe how the correct background color is shown even though the page hasn't been loaded yet.

Or am I misunderstanding the question?

kittaakos:
> Or am I misunderstanding the question?

No, you're right. Thanks for the explanation

planger:
@kittaakos Great, thanks! Can you please re-approve the rebase, if you agree to merge? Thank you!

JonasHelming (approved):

PR #12923, opened 2023-09-19
dialogs - Allow multiple selection on Open...

### LINKED ISSUE #12905: Allow multiple selection on Open... dialog
Currently, multi selection is not supported for the `Open...` dialog. To open multiple files or multiple folders, the dialog has to be opened several times.
It would be nice to be able to select multiple files or multiple folders at once. In case of heterogeneous selection, i.e several files and folders at once, specific action to each kind should be performed (open file or open new workspace on folder)

rschnekenbu:
#### What it does

fixes #12905 - Allow multiple selection on Open...

- allow multiple selection in doOpen dialog 
- handle an array or a single URI or undefined when closing dialog

#### How to test

When running Theia in browser, select the Open... dialog:
- It is now possible to open multiple files at once. 
- It is also possible to select multiple folders. In that case, all folders will be open as folders in a new workspace (as done in Electron).
- In case of heterogeneous selection, all files will be opened in current workspace, and the folders will be opened together as a new workspace.

12905-browser

Electron has 2 different menus to select either Folder or Files:
- This patch updates the `Open file...` selection dialog to also be able to select multiple items. If the selection has several files and a folder selected in the dialog, the open button of the dialog won't react. It is supposed to browse the folder if there is only one folder selected. 
- The `open folder` command already supports multiple folder selection, creating a new workspace containing all folders. No updates were done to this command.

#### Follow-ups

No Follow-up was identified during this task development. 

#### Review checklist

#### Reminder for reviewers

- [X] As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

### DIFF
--- a/packages/workspace/src/browser/workspace-frontend-contribution.ts
+++ b/packages/workspace/src/browser/workspace-frontend-contribution.ts
@@ -262,50 +262,76 @@ export class WorkspaceFrontendContribution implements CommandContribution, Keybi
      * This is the generic `Open` method. Opens files and directories too. Resolves to the opened URI.
      * Except when you are on either Windows or Linux `AND` running in electron. If so, it opens a file.
      */
-    protected async doOpen(): Promise<URI | undefined> {
+    protected async doOpen(): Promise<URI[] | undefined> {
         if (!isOSX && this.isElectron()) {
             return this.doOpenFile();
         }
         const [rootStat] = await this.workspaceService.roots;
-        const destinationUri = await this.fileDialogService.showOpenDialog({
+        let selectedUris = await this.fileDialogService.showOpenDialog({
             title: WorkspaceCommands.OPEN.dialogLabel,
             canSelectFolders: true,
-            canSelectFiles: true
+            canSelectFiles: true,
+            canSelectMany: true
         }, rootStat);
-        if (destinationUri && this.getCurrentWorkspaceUri()?.toString() !== destinationUri.toString()) {
-            const destination = await this.fileService.resolve(destinationUri);
-            if (destination.isDirectory) {
-                this.workspaceService.open(destinationUri);
-            } else {
-                await open(this.openerService, destinationUri);
+        if (selectedUris) {
+            if (!Array.isArray(selectedUris)) {
+                selectedUris = [selectedUris];
             }
-            return destinationUri;
+            const folders: URI[] = [];
+            //  Only open files then open all folders in a new workspace, as done with Electron see doOpenFolder.
+            for (const uri of selectedUris) {
+                const destination = await this.fileService.resolve(uri);
+                if (destination.isDirectory) {
+                    if (this.getCurrentWorkspaceUri()?.toString() !== uri.toString()) {
+                        folders.push(uri);
+                    }
+                } else {
+                    await open(this.openerService, uri);
+                }
+            }
+            if (folders.length > 0) {
+                const openableURI = await this.getOpenableWorkspaceUri(folders);
+                if (openableURI && (!this.workspaceService.workspace || !openableURI.isEqual(this.workspaceService.workspace.resource))) {
+                    this.workspaceService.open(openableURI);
+                }
+            }
+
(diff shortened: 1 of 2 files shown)


packages/workspace/src/browser/workspace-frontend-contribution.ts
```diff
@@ -262,28 +262,60 @@ export class WorkspaceFrontendContribution implements CommandContribution, Keybi
      * This is the generic `Open` method. Opens files and directories too. Resolves to the opened URI.
      * Except when you are on either Windows or Linux `AND` running in electron. If so, it opens a file.
      */
-    protected async doOpen(): Promise<URI | undefined> {
+    protected async doOpen(): Promise<MaybeArray<URI> | undefined> {
         if (!isOSX && this.isElectron()) {
             return this.doOpenFile();
         }
         const [rootStat] = await this.workspaceService.roots;
         const destinationUri = await this.fileDialogService.showOpenDialog({
```
tsmaeder, line 270: `destinationUri` is no longer a good name for what might be multiple URI's. `selectedUris` comes to mind.

packages/workspace/src/browser/workspace-frontend-contribution.ts
```diff
@@ -262,28 +262,60 @@ export class WorkspaceFrontendContribution implements CommandContribution, Keybi
...
-            canSelectFiles: true
+            canSelectFiles: true,
+            canSelectMany: true
         }, rootStat);
-        if (destinationUri && this.getCurrentWorkspaceUri()?.toString() !== destinationUri.toString()) {
-            const destination = await this.fileService.resolve(destinationUri);
-            if (destination.isDirectory) {
-                this.workspaceService.open(destinationUri);
+        if (destinationUri) {
+            if (Array.isArray(destinationUri)) {
```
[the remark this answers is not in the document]
tsmaeder: On a general note, I think APIs that return `T | T[]` don't make sense in many cases: if you care about the "single element" case, you can still check for `length === 1` and if you don't care, you don't have to handle the case.

packages/workspace/src/browser/workspace-frontend-contribution.ts
```diff
@@ -262,28 +262,60 @@ export class WorkspaceFrontendContribution implements CommandContribution, Keybi
...
+                // only open files then open all folders in a new workspace, as done with Electron see doOpenFolder
+                const folders: MaybeArray<URI> = [];
+                for (const uri of destinationUri) {
+                    if (this.getCurrentWorkspaceUri()?.toString() !== uri.toString()) {
```
tsmaeder, line 281: This check is only necessary in the 'folder' case. If one of the selected file is a workspace file, we still want to open it as a file, not a workspace. Performance is not an issue here, since we're doing this a few times max per user interaction.

packages/workspace/src/browser/workspace-frontend-contribution.ts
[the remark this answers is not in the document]
rschnekenbu: Thanks for the hint to simplify algorithm! Indeed, that's better to handle all cases at once. 
I made explicit in the signature of the doOpen... methods that an array of elements will be returned, getting rid of the MaybeArray. I kept however the signature of the showOpenDialog service with this MaybeArray, to avoid compatibility issues.

packages/workspace/src/browser/workspace-frontend-contribution.ts
rschnekenbu: Changed to selectedUris and selectedFilesUris.

packages/workspace/src/browser/workspace-frontend-contribution.ts
rschnekenbu: Check has been moved to the folder case only. It is now possible to open a workspace file as a simple file.

tsmaeder (approved):
Code looks good to me, now

PR #12936, opened 2023-09-25
Improve frontend startup performance

sdirix:
The frontend awaits 'initialize', 'configure' and 'onStart' for all frontend contributions. It's therefore very important that no expensive code is run there.

By not awaiting expensive operations in

 - DebugFrontendApplicationContribution.onStart (~300ms)
 - EditorNavigationContribution.onStart (~36-400ms)
 - TerminalFrontendContribution.onStart (~100-300ms)

the reported startup time without using plugins is reduced by ~400-1000ms which is an improvement of ~20-40%.

#### What it does

It removes some `await`ed instructions in the critical execution paths which are not really necessary to be awaited. The remaining `await` instruction in the `common-frontend-contribution` which takes about ~200ms of time to be executed was not removed as this might have some side effects in theory, although in practice I was not able to observe any.

---

Note: I also tried to run all `initialize`, `configure` and `onStart` phases each in "parallel", using `Promise.all`. However when combined with the remaining changes this actually lead to a slowdown, probably because the Javascript thread is already fully utilized.

#### How to test

Verify that Theia behaves as before. Please have a look at the code paths to see whether something was missed.

You can start Theia with `logLevel=debug` for more extensive logging and verify the reported startup times. For proper logging in `production` mode, you need to disable class name substitution. This can be achieved by adding the following to the `webpack.config.js`:

```javascript
const TerserPlugin = require('terser-webpack-plugin');
configs[0].optimization = {
    minimizer: [
        new TerserPlugin({
            terserOptions: {
                keep_classnames: true
            }
        })
    ]
};
```

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ## v1.42.0
 
 - [core] fixed logger level propagation when log config file changes at runtime [#12566](https://github.com/eclipse-theia/theia/pull/12566) - Contributed on behalf of STMicroelectronics
+- [core] improve frontend startup time [#12936](https://github.com/eclipse-theia/theia/pull/12936) - Contributed on behalf of STMicroelectronics
 - [vscode] stub TestController invalidateTestResults [#12944](https://github.com/eclipse-theia/theia/pull/12944) - Contributed by STMicroelectronics
 - [vscode] support iconPath in QuickPickItem [#12945](https://github.com/eclipse-theia/theia/pull/12945) - Contributed by STMicroelectronics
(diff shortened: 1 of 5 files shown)


msujew (approved):
The changes look good to me. I cannot identify any regressions with the approach and the startup time of frontend contributions is indeed reduced.

Though not as much as advertised, but the initial startup time wasn't as long anyway. Reduced the `FrontendApplicationContribution#onStart` time from ~500ms to ~300ms for me.

sdirix:
I measured the times only on my own machine ;)

msujew:
@sdirix I wasn't fast enough to press the merge button. Please rebase again :)

PR #12590, opened 2023-06-01
Add inversify lifecycle to frontend preload script

msujew:
#### What it does

Currently, the browser preload scripts has 3 tasks:
* Load the translations for the selected locale
* Load the backend OS to correctly display stuff like line endings
* Set the background of the application based on the last selected theme

However, these tasks are hardcoded and adopters cannot add new scripts that are supposed to be executed during the preload phase of the frontend application. This change adds an inversify lifecycle to the aforementioned frontend preload scripts and a new contribution point to the `package.json` called `frontendPreload`. This allows to define new `PreloadContribution` items.

It also changes the way communication within the preload script is being performed. Instead of using `fetch`, we use the normal rpc messaging system via `WebSocketConnectionProvider`.

As a drive by, also:
Closes https://github.com/eclipse-theia/theia/issues/12670
Closes https://github.com/eclipse-theia/theia/issues/6675

#### How to test

1. Confirm that the tasks outlined above still work as expected.
2. Remove the `frontendPreload` from `@theia/core/package.json`. Assert that the preload module is removed from the generated `index.js` file.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #12950, opened 2023-09-27
fix: restore src-gen frontend production behavior

sdirix:
Restores the usage of 'require' in production mode when loading included modules.

At the moment, both in development as well as in production mode, the frontend will be generated using the 'import' function. Webpack will use this opportunity for code splitting, leading to the generation of many small bundles. As they are sequentially loaded, this has a negative effect on the startup speed of the frontend.

By using 'require' instead, webpack will produce a single bundle, which is faster to transmit without additional roundtrips. This behavior was used for production mode in all previous Theia versions and is now restored.

#### What it does

Use `require` for importing frontend modules in `production` mode, use `import` in the remaining modes (mainly development).

#### How to test

Trigger src-gen for the included example applications. Trigger with `--mode development` as well as `--mode production`.

#### Review checklist

#### Reminder for reviewers

#### Context

The require generation was removed with https://github.com/eclipse-theia/theia/pull/12590

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - [core] fixed logger level propagation when log config file changes at runtime [#12566](https://github.com/eclipse-theia/theia/pull/12566) - Contributed on behalf of STMicroelectronics
 - [core] improve frontend startup time [#12936](https://github.com/eclipse-theia/theia/pull/12936) - Contributed on behalf of STMicroelectronics
+- [dev-packages] restore src-gen frontend production behavior [12950](https://github.com/eclipse-theia/theia/pull/12950) - Contributed on behalf of STMicroelectronics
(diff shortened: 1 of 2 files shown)


msujew (commented):
Does it make sense to replace all `import` statements with `require`? I.e. even in development build? Is there a benefit of using `import` at all in this context?

Note that we should definitely replace the `import` call at line `100`.

sdirix:
> Note that we should definitely replace the import call at line 100.

Ah I missed that! Good catch! Will update the PR.

>Does it make sense to replace all import statements with require? I.e. even in development build? Is there a benefit of using import at all in this context?

Good questions, I assumed that the webpack build can be a bit faster, including in watch mode, if all modules are bundled separately. However I did not check whether that's really the case.

msujew (approved):
Looks good to me 👍

Issue #12962, opened 2023-09-30
`Cannot GET /bundle.js` with browser production build


perrinjerome:
### Bug Description:

I have been using a theia browser application very similar to examples/browser, where unlike examples/browser which uses `theia build` in development mode (see below), I just used the `theia build` with default mode, which is production.

https://github.com/eclipse-theia/theia/blob/bb7cdbbb46dc14cc9db093756243bc94571bb1c6/examples/browser/package.json#L65

It was working fine with theia 1.35.0 but when I tried to update to 1.42.0 I have an error when opening http://localhost:3000/ with the browser because `/bundle.js` serves an error page `Cannot GET /bundle.js`.

### Steps to Reproduce:

After cloning theia repository:

```sh
yarn install
cd examples/browser/
yarn compile
npx theia clean  # make sure we are clean from previous builds
npx theia rebuild
npx theia build  # ( without --mode=production )
npx theia start
```
```sh
# test the application, this prints 200 on success, 404 on error
curl -s -o /dev/null -w "%{http_code}" http://localhost:3000/bundle.js
```

### Additional Information

I have bisected that this comes from 45a0953d7eed19cd311840281f17999d268ba7cf which seems unrelated, but maybe triggers something with webpack in production mode. I have no idea why this would happen, but when reverting part of this commit by applying this patch on top of master, `bundle.js` is generated again:

```patch
diff --git a/packages/core/src/browser/preload/theme-preload-contribution.ts b/packages/core/src/browser/preload/theme-preload-contribution.ts
index a8df24634..c275f72ec 100644
--- a/packages/core/src/browser/preload/theme-preload-contribution.ts
+++ b/packages/core/src/browser/preload/theme-preload-contribution.ts
@@ -17,14 +17,14 @@
 import { PreloadContribution } from './preloader';
 import { DEFAULT_BACKGROUND_COLOR_STORAGE_KEY } from '../frontend-application-config-provider';
 import { injectable } from 'inversify';
-import { DefaultTheme } from '@theia/application-package';
+// import { DefaultTheme } from '@theia/application-package';
 
 @injectable()
 export class ThemePreloadContribution implements PreloadContribution {
 
     initialize(): void {
         const dark = window.matchMedia?.('(prefers-color-scheme: dark)').matches;
-        const value = window.localStorage.getItem(DEFAULT_BACKGROUND_COLOR_STORAGE_KEY) || DefaultTheme.defaultBackgroundColor(dark);
[...]

PR #12964, opened 2023-09-30
Fix preload application package import

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/12962

Fixes the import to use one that doesn't rely on node.js dependencies.

#### How to test

Confirm that the preload functionality behaves as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/preload/theme-preload-contribution.ts
+++ b/packages/core/src/browser/preload/theme-preload-contribution.ts
@@ -17,7 +17,7 @@
 import { PreloadContribution } from './preloader';
 import { DEFAULT_BACKGROUND_COLOR_STORAGE_KEY } from '../frontend-application-config-provider';
 import { injectable } from 'inversify';
-import { DefaultTheme } from '@theia/application-package';
+import { DefaultTheme } from '@theia/application-package/lib/application-props';
 
 @injectable()
 export class ThemePreloadContribution implements PreloadContribution {


msujew:
@vince-fugnitto I imagine we need a `1.42.1` with this fix, the current `1.42.0` version isn't working well.

planger (approved):
Thank you for this fix! I can confirm that preload and production build works again for browser with this fix.

I'm sorry for having introduced this issue with https://github.com/eclipse-theia/theia/pull/12897, I didn't expect to break the preload / production build with this package level import.

To reveal such an issue in the future early, I'd suggest to run a very simple playwright test on the production mode on every change, which just checks if the application loads correctly. It'd be very easy to set up and I'd be happy to open a PR for that. WDYT?

msujew:
> To reveal such an issue in the future early, I'd suggest to run a very simple playwright test on the production mode on every change, which just checks if the application loads correctly. It'd be very easy to set up and I'd be happy to open a PR for that. WDYT?

I wanted to propose something similar. It's not like I had tests for the preload stuff, but I was still suprised that our tests didn't catch the issue. I'm all in for that 👍

planger:
> All good, I know how it feels to have broken something accidentally ;)

Thank you! :-)

> I wanted to propose something similar. It's not like I had tests for the preload stuff, but I was still suprised that our tests didn't catch the issue. I'm all in for that 👍

Great, I'll open a PR in the next 1-2 days!
It'd be good to at least have a very simple smoke test for the production build, which currently doesn't seem to be verified at all.

planger:
For future references: I've opened https://github.com/eclipse-theia/theia/pull/12965 to add a smoke test for the production build

vince-fugnitto:
> @vince-fugnitto I imagine we need a `1.42.1` with this fix, the current `1.42.0` version isn't working well.

Sounds good, we can mention it during the dev meeting (tomorrow) and perform the release.

PR #12965, opened 2023-09-30
Add smoke test for production build

planger:
#### What it does

Currently we don't seem to have any automated test that verifies whether the production build works as expected.
For a recent issue and its root cause, which remained unrevealed by the CI, see https://github.com/eclipse-theia/theia/issues/12962.

With this PR, we add a new Github workflow, which builds the example browser app in production mode and runs a simple smoke test. This smoke test just opens the app and awaits the loading spinner to disappear and expects the workbench to be visible within a certain timeframe.

#### How to test

The check should be green after https://github.com/eclipse-theia/theia/pull/12964 is merged.

#### Follow-ups

None

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.github/workflows/production-smoke-test.yml
+++ b/.github/workflows/production-smoke-test.yml
@@ -28,10 +28,10 @@ jobs:
           node-version: "18.x"
           registry-url: "https://registry.npmjs.org"
 
-      - name: Use Python 3.x
+      - name: Use Python 3.11
         uses: actions/setup-python@v4
         with:
-          python-version: "3.x"
+          python-version: "3.11"
 
       - name: Build Browser Example Application for Production
         shell: bash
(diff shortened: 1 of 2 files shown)


planger:
@msujew Sorry for the ping, do you have time to review this change? If not, let me know. Thank you!

.github/workflows/production-smoke-test.yml
```diff
@@ -0,0 +1,54 @@
...
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - master
+  schedule:
+    - cron: "0 5 * * *" # Runs every day at 5am: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule
```
msujew, line 12: Question: Do we need the schedule, if it's run on every PR and every push to master?

msujew (approved):
Looks pretty good 👍

.github/workflows/production-smoke-test.yml
planger: Yeah, you're probably right. I'll remove it.

PR #12971, opened 2023-10-02
12970 Fixes hover widget jumping

### LINKED ISSUE #12970: Debug hover widget jumping after pointer
### Bug Description:
Debug hover widget is moving after the pointer with some delay. It interferes to move to the widget. Widget should not move after pointer if pointer does not leave the variable

### Steps to Reproduce:

1. Start debugging
2. Stop on breakpoint
3. Hover the pointer above any variable
4. Move the pointer side-to-side above this variable
5. Debug hover widget is moving after the pointer with some delay

### Additional Information

- Operating System: any
- Theia Version: the latest

Zebsterpasha:
#### What it does
Fixes #12970

#### How to test
1. Start debugging
2. Stop on breakpoint
3. Hover the pointer above any variable
4. Move the pointer side-to-side
5. Debug hover widget is not moving after the pointer

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/editor/debug-hover-widget.ts
+++ b/packages/debug/src/browser/editor/debug-hover-widget.ts
@@ -203,12 +203,12 @@ export class DebugHoverWidget extends SourceTreeWidget implements monaco.editor.
             this.hide();
             return;
         }
-        const lineContent = this.editor
+        const expressionLineContent = this.editor
             .getControl()
             .getModel()!
             .getLineContent(this.options.selection.startLineNumber);
         const startColumn =
-            lineContent.indexOf(
+            expressionLineContent.indexOf(
                 matchingExpression,
                 this.options.selection.startColumn - matchingExpression.length
             ) + 1;
--- a/packages/debug/src/browser/editor/debug-hover-widget.ts
+++ b/packages/debug/src/browser/editor/debug-hover-widget.ts
@@ -203,6 +203,22 @@ export class DebugHoverWidget extends SourceTreeWidget implements monaco.editor.
             this.hide();
             return;
         }
+        const lineContent = this.editor
+            .getControl()
+            .getModel()!
+            .getLineContent(this.options.selection.startLineNumber);
+        const startColumn =
+            lineContent.indexOf(
+                matchingExpression,
+                this.options.selection.startColumn - matchingExpression.length
+            ) + 1;
+        const endColumn = startColumn + matchingExpression.length;
+        this.options.selection = new monaco.Range(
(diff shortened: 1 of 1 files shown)


Zebsterpasha:
@planger Okey, I'll think about moving code to the method ) I've just fixed wrong behavior, showed in the first video)

planger:
> @planger Okey, I'll think about moving code to the method ) I've just fixed wrong behavior, showed in the first video)

Thank you for looking into this again! However, now your change doesn't seem to have an impact for my test with the mock debug extension anymore, as the condition on line 180 is true, so your code is never executed. Or am I missing something?
What is your test scenario?

Can you please take another look? Thank you very much!

Zebsterpasha:
@planger Yeah, this is correct. In your situation my code, probably, should not work. Mock Debug Extension has its own ExpressionProvider 
https://github.com/microsoft/vscode-mock-debug/blob/b8d3d3a436e94f73ca193bf0bfa7c5c416a8aa8c/src/activateMockDebug.ts#L104
and this provider choose expression to highlight. In your first video we can see some jumps, the reason is that Mock Provider choose one expression to highlight and one Range (it could be bigger than expression), but we overrided the Range and made it less.

My main scenario - we don't have any ExpressionProvider, just pure Editor and model, but I have tested the latest version of MockDebug, DebugWidget is not jumping after pointer. 
I think, ExpressionProvider should manage range, but not we. 
Based on debug-hover-widget source code it was meant to be.

planger (approved):
@Zebsterpasha Ok thanks, makes sense! I don't have objections to your changes and they already improve the situation in the context of the scenario you describe. So I'll go ahead and approve this PR to avoid offloading more on to your fix.

However, I think at least mid-term we should expand this work to improve the situation for all scenarios of a debug hovers. At least for me with the MockDebugExtension (with and without your changes), the debug hover behavior is not ideal.

1. It follows the pointer if moved above an expression
2. It is not always nicely rendered vertically (above or below the expression but sometimes covers the expression)
3. It is sometimes placed very weirdly

https://github.com/eclipse-theia/theia/assets/588090/8c51e199-421f-4059-92b5-8ec7308566b6

@Zebsterpasha If you have the chance to additionally investigate the scenarios above for all debug hover cases, as you are already familiar with this code, we'd be very grateful! If you can't do that, we can merge this change and I'll open a new ticket for the future improvements.

Thanks again for your valuable contribution!

planger:
Alright, I've opened https://github.com/eclipse-theia/theia/issues/12994

PR #12976, opened 2023-10-04
Fix URL path for output renderer scripts

### LINKED ISSUE #12966: Notebook renderer basepath should be relative
### Bug Description:
The rendering of notebook output is not working in our app, as renderer scripts are imported from the root path whereas our app is mounted elsewhere. Any way this can be taken into account?

https://github.com/eclipse-theia/theia/blob/ab4627e4ffe15aec236793f98daab534c4e77f40/packages/plugin-ext/src/main/browser/plugin-contribution-handler.ts#L416-L422C10

### Steps to Reproduce:

1. Access a Theia app through a reverse proxy rewriting the url path to anything other than the root 
3. Open e.g. a Jupyter notebook (the appropriate extension should ofc be installed)
4. Attempt to render outputs via cell execution
5. Observe 404 requests for renderer script

image

### Additional Information

- Operating System: `5.10.186-179.751.amzn2.x86_64 GNU/Linux`
- Theia Version: 1.41

jonah-iden:
#### What it does

Fixes #12966
Fixes an issue with notebook api downloading output renderer scripts from base path instead of deployed path

#### How to test

Easiest for me was using nginx.
1. run Theia locally
2. Add following config to the server configuration of nginx
```
location /theia/ {
            rewrite ^/theia(/.*)$ $1 break;
            proxy_pass http://localhost:3000/;
            proxy_http_version 1.1;
            proxy_set_header Host $host;
        }
```
3. open theia on http://localhost/theia
4. open a notebook, execute a cell and see the output renderer works.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/notebooks/renderers/output-webview-internal.ts
+++ b/packages/plugin-ext/src/main/browser/notebooks/renderers/output-webview-internal.ts
@@ -160,8 +160,8 @@ export async function outputWebviewPreload(ctx: PreloadContext): Promise<void> {
             if (this.rendererApi) {
                 return this.rendererApi;
             }
-
-            const rendererModule = await __import(this.data.entrypoint.uri) as { activate: rendererApi.ActivationFunction };
+            const baseUri = window.location.href.replace(/\/webview\/index\.html.*/, '');
+            const rendererModule = await __import(`${baseUri}/${this.data.entrypoint.uri}`) as { activate: rendererApi.ActivationFunction };
             this.rendererApi = await rendererModule.activate(this.createRendererContext());
             return this.rendererApi;
         }
--- a/packages/plugin-ext/src/main/browser/plugin-contribution-handler.ts
+++ b/packages/plugin-ext/src/main/browser/plugin-contribution-handler.ts
@@ -23,7 +23,7 @@ import { PluginViewRegistry } from './view/plugin-view-registry';
 import { PluginCustomEditorRegistry } from './custom-editors/plugin-custom-editor-registry';
 import {
     PluginContribution, IndentationRules, FoldingRules, ScopeMap, DeployedPlugin,
-    GrammarsContribution, EnterAction, OnEnterRule, RegExpOptions, getPluginId
+    GrammarsContribution, EnterAction, OnEnterRule, RegExpOptions, PluginPackage
 } from '../../common';
 import {
     DefaultUriLabelProviderContribution,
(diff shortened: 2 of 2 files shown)


bvenreply:
Hello @jonah-iden, we tested this solution but it's not working for us, we're getting this error now:

image

It seems the browser import function is being finicky about relative module paths. Prefixing a `./` seems to fix it:

image

jonah-iden:
Hi, sorry for the late reply. 
My solution worked for me and was also used in other parts of theia. I'll try to take a look later today.

bvenreply:
> Hi, sorry for the late reply. My solution worked for me and was also used in other parts of theia. Maybe there is a bigger problem in theia then and i did not test thoroughly enough. I'll try to take a look later today.

No problem. Thank you @jonah-iden. Btw another possibility to consider is that I messed something up on my end...

jonah-iden:
@bvenreply, i was able to reproduce your error and improved my previous solution. For me this works great. could you test it again if you find the time?

bvenreply:
@jonah-iden I tested the new changes, I confirm it's working. Thanks!

JonasHelming:
@msujew We discussed yesterday that your approval is enough for this one to be merged

packages/plugin-ext/src/main/browser/notebooks/renderers/output-webview-internal.ts
```diff
@@ -160,8 +160,8 @@ export async function outputWebviewPreload(ctx: PreloadContext): Promise<void> {
             if (this.rendererApi) {
                 return this.rendererApi;
             }
-
-            const rendererModule = await __import(this.data.entrypoint.uri) as { activate: rendererApi.ActivationFunction };
+            const baseUri = window.location.href.replace(/\/webview\/index\.html.*/, '');
```
msujew, line 163: The `/webview/` part used here is part of an implementation detail of the [`WebviewEnvironment`](https://github.com/eclipse-theia/theia/blob/f1056b3c567e8cc9d3807850164d4f48967f28a6/packages/plugin-ext/src/main/browser/webview/webview-environment.ts#L55-L61), making it fragile in case adopters decide to change the path.

Instead I'd advocate for a different approach. Why don't we pass the full `URI` to the `data.entrypoint.uri` property? Currently, we only pass the path of the renderer, but we should probably pass the full URI.

jonah-iden: but if we pass the full path wouldn't we get cross origin issues?

jonah-iden: I just confirmed we get cross origin errors if we pass in the full uri. 
What we could do would be to pass the full path and then just use the host and port from the webview url

msujew: I guess, yeah we do...

Ok, I'll approve for now, and we can get back to this in case someone actually has trouble with this approach. I assume no one actually manipulates the `/webview` path anyway.

msujew (approved):
Alright, works well. I'm not 100% sure this is the best approach to this, but it works and we can easily change it in case there are any issues.

PR #12980, opened 2023-10-08
Fix #12546: UI not updated upon editing breakpoint condition

### LINKED ISSUE #12546: UI not updated upon editing breakpoint condition
### Steps to Reproduce:

1. Using Theia Electron (or Browser) Example on the master branch,
add a source breakpoint in a JavaScript (or TypeScript) file.

2. Edit the breakpoint and set a condition (Expression or Hit Count).

3. Note that UI (e.g. the breakpoint's icon) is not updated after the condition is set.

| Observed | Expected |
|--------- | -------- |
| Screenshot | Screenshot 2023-05-18 at 19 42 07 |

### Additional Information

Debugging revealed that no `SourceBreakpointsChangeEvent` is fired from the `BreakpointManager.setMarkers` method upon setting breakpoint condition (step 2).

See also #8851.

pisv:
#### What it does

Fixes #12546, which is a regression introduced by #12183, by ensuring that `BreakpointManager.setMarkers` fires a `SourceBreakpointsChangeEvent` when `oldMarker === newMarker`, as there is no way to actually detect a change in this case.

#### How to test

- Verify that #12546 is actually fixed by following the steps given in the issue description.

- Also verify that #11878, which was fixed by #12183, is still fixed after this patch is applied, and there are no other regressions in breakpoint management in general.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/breakpoint/breakpoint-manager.ts
+++ b/packages/debug/src/browser/breakpoint/breakpoint-manager.ts
@@ -75,7 +75,9 @@ export class BreakpointManager extends MarkerManager<SourceBreakpoint> {
                 added.push(newMarker);
             } else {
                 // We emit all existing markers as 'changed', but we only fire an event if something really did change.
-                didChangeMarkers ||= !!added.length || !deepEqual(oldMarker, newMarker);
+                // We also fire an event if oldMarker === newMarker, as we cannot actually detect a change in this case
+                // (https://github.com/eclipse-theia/theia/issues/12546).
+                didChangeMarkers ||= !!added.length || oldMarker === newMarker || !deepEqual(oldMarker, newMarker);
                 changed.push(newMarker);
             }
         }


tsmaeder:
I don't know if this is the right fix: if we look at `DebugSourceBreakpoint.updateOrigins()`, we see that the breakpoint objects are updated in place. But that does not make sense if we have a model where we have change detection in the BreakpointManager. Shouldn't we try and clean up the update process so that the change detection works correctly?

tsmaeder:
> The contract of the MarkerManager.setMarkers method, which the BreakpointManager overrides, states:
>>    replaces the current markers for the given uri and owner with the given data

Yes, but isn't `DebugSourceBreakpoint.updateOrigins()` violating that contract by reaching into the internal structure of the `SourceBreakpoint`s? To me it seems like there is no clear separation of responsibility here. Since `SourceBreakpoint` has not change notification facility, some other instance has to take care of appropriately sending out change notification. Which component owns the SourceBreakpoint.raw field?

pisv:
Perhaps, but I would not consider it as directly relevant to the proposed patch.

Basically, as a Theia user, I would just like the issue #12546, which was reported back in May, to be fixed.

Since it has not been fixed so far, I proposed a patch after due consideration of the underlying issue, which does fix a major part of the breakage introduced by #12183, which caused the issue in the first place.

Given that, I'd like to leave any further optimisations in the design or implementation of the related code, which are of course possible, to the esteemed committers or other contributors.

It seems fair, doesn't it?

pisv:
OK. The change to `BreakpointManager.setMarkers` from #12183 would make sense in its current form if and only if all the instances of the `SourceBreakpoint` were (deeply) immutable. That probably means that all of the instances of any subclasses of the `BaseBreakpoint` would need to be made immutable too, and also the `ExceptionBreakpoint`, for the sake of the design symmetry. Of course, it would be an even more radical breaking change. But it does look like a better design.

I could try to make the change myself, but since it would be a big breaking change and a significant amount of work, I'd like to be sure upfront that we are really on the same page here, and such a change could actually be accepted.

tsmaeder:
<sigh>So I guess we'll have to live with your fix for the moment, even though, IMO, it's a workaround for the general bogusness surrounding this topic.

tsmaeder (approved):

tsmaeder:
I've filed https://github.com/eclipse-theia/theia/issues/13052

tsmaeder:
@pisv could you rebase and push to trigger running a the tests against the latest version of master?

pisv:
> So I guess we'll have to live with your fix for the moment, even though, IMO, it's a workaround for the general bogusness surrounding this topic.

Thank you for your review and approval, Thomas.

I agree with you regarding "the general bogusness surrounding this topic" (to put it mildly).

I still think though that the only clean solution in the presence of change detection in `BreakpointManager.setMarkers` would be to make the `SourceBreakpoint` immutable. This would allow us to keep the general contract of `MarkerManager.setMarkers` and avoid any unexpected behavior. Even without the change detection, it would be a better design IMO.

Having said that, I can understand that such a breaking change could be unacceptable.

> could you rebase and push to trigger running a the tests against the latest version of master?

I have rebased and pushed the branch, but it looks like an intermittent build failure has occurred now.

PR #12981, opened 2023-10-09
Remove baseUrl from tsconfigs

tortmayr:
#### What it does

According to the [tsconfig doc](https://www.typescriptlang.org/tsconfig#baseUrl): 

This feature was designed for use in conjunction with AMD module loaders in the browser, and is not recommended in any other context. As of TypeScript 4.1, baseUrl is no longer required to be set when using paths. 

In addition, this options does not work well with auto import resolving of Typescript-LSPs (i.e. Theia/VS Code). Auto import might resolves to a statement starting with `src` ( e.g. `  import { ChannelMultiplexer } from 'src/common/message-rpc/channel';`) which does not work well with inversify and can result in runtime errors.

Most of the packages don't use it anyways. This PR removes the remaining usages and adapts `src` imports.

#### How to test

Building & running the example application should still work as before

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/playwright/src/tests/theia-output-view.test.ts
+++ b/examples/playwright/src/tests/theia-output-view.test.ts
@@ -15,7 +15,7 @@
 // *****************************************************************************
 
 import { expect } from '@playwright/test';
-import { TheiaOutputViewChannel } from 'src/theia-output-channel';
+import { TheiaOutputViewChannel } from '../theia-output-channel';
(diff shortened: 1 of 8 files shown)


martin-fleck-at (approved):
Looks good to me and I couldn't find any other places where the `baseUrl` is defined. CI is also happy so let's get this merged! Thank you, Tobias!

Issue #12982, opened 2023-10-09
ApplicationShellOptions are not used on ApplicationShell


martin-fleck-at:
### Bug Description:
When binding custom `ApplicationShellOptions`, they are not properly applied in the `ApplicationShell` due to a missing assignment.
 
### Steps to Reproduce:

1. Use a custom re-binding in your application, e.g., `rebind(ApplicationShellOptions).toConstantValue(<ApplicationShell.Options>{ leftPanel: { initialSizeRatio: 0.5 } });` or replace the binding in Theia with something that has values (https://github.com/eclipse-theia/theia/blob/master/packages/core/src/browser/frontend-application-module.ts#L163)
2. Ensure your workspace layout is reset to default.
3. Expand the left sidebar and see that the wrong ratio is used.

### Additional Information

- Operating System: any
- Theia Version: master

PR #12983, opened 2023-10-09
Ensure ApplicationShellOptions are properly used for shell init

martin-fleck-at:
#### What it does
Ensure ApplicationShellOptions are properly used for shell init
Fixes https://github.com/eclipse-theia/theia/issues/12982

#### How to test

1. Use a custom re-binding in your application, e.g., `rebind(ApplicationShellOptions).toConstantValue(<ApplicationShell.Options>{ leftPanel: { initialSizeRatio: 0.5 } });` or replace the binding in Theia with something that has values (https://github.com/eclipse-theia/theia/blob/master/packages/core/src/browser/frontend-application-module.ts#L163)
2. Ensure your workspace layout is reset to default.
3. Expand the left sidebar and see that the right ratio is used.

#### Follow-ups

None.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/application-shell.ts
+++ b/packages/core/src/browser/shell/application-shell.ts
@@ -272,6 +272,22 @@ export class ApplicationShell extends Widget {
         @inject(SecondaryWindowHandler) protected readonly secondaryWindowHandler: SecondaryWindowHandler,
     ) {
         super(options as Widget.IOptions);
+
+        // Merge the user-defined application options with the default options
+        this.options = {
+            bottomPanel: {
+                ...ApplicationShell.DEFAULT_OPTIONS.bottomPanel,
+                ...options?.bottomPanel || {}
+            },
+            leftPanel: {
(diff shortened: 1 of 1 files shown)


packages/core/src/browser/shell/application-shell.ts
```diff
@@ -307,15 +307,15 @@ export class ApplicationShell extends Widget {
         this.options = {
```
tortmayr, line 307: IMO we could resolve the options directly in the constructor. There is no need to this in postConstruct because we don't rely on any injected services. This would also remove the need to expose the partial options as protected property.

martin-fleck-at: Thank you for the having a look! Good point, I updated the PR with this simplification.

tortmayr (approved):
Thanks for the quick fix. Change works great and custom shell options are now properly picked up and used for shell init. 👍🏼

PR #12984, opened 2023-10-09
Implement EnvironmentVariableMutatorOptions API.

### LINKED ISSUE #12941: [vscode] EnvironmentVariableMutator can now be configured
Vscode 1.82 has improved the support for EnvironmentVariableMutator, with options to configure when the Mutator is applied

See https://code.visualstudio.com/updates/v1_82#_configure-when-a-environmentvariablemutator-is-applied

EnvironmentVariable-Mutator

tsmaeder:
#### What it does
Implements options when mutating `EnvironmentVariableCollection`s. Since Theia does not support shell integration scripts, one of the options is stubbed. See the related issue for details.

 Fixes #12941

#### How to test

Install the attached extension. It exercises the various cases like so:

```
    context.environmentVariableCollection.append('MUTATED_DEFAULT', 'a string');
    context.environmentVariableCollection.append('MUTATED_TRUE', 'a string', { applyAtProcessCreation: true });
    context.environmentVariableCollection.append('MUTATED_FALSE', 'a string', { applyAtProcessCreation: false });
    context.environmentVariableCollection.append('MUTATED_INT_TRUE', 'a string', { applyAtShellIntegration: true, applyAtProcessCreation: false });
```` 
After startup, open a terminal. My expectation is that THE `MUTATED_DEFAULT` and `MUTATED_TRUE` are set in the terminal shell.

#### Follow-ups
If we ever decide to implement shell integration scripts, we need to honor the corresponding flag.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/terminal-ext.ts
+++ b/packages/plugin-ext/src/plugin/terminal-ext.ts
@@ -372,16 +372,16 @@ export class EnvironmentVariableCollection implements theia.EnvironmentVariableC
         return this.map.size;
     }
 
-    replace(variable: string, value: string): void {
-        this._setIfDiffers(variable, { value, type: EnvironmentVariableMutatorType.Replace });
+    replace(variable: string, value: string, options?: theia.EnvironmentVariableMutatorOptions): void {
+        this._setIfDiffers(variable, { value, type: EnvironmentVariableMutatorType.Replace, options: options ?? { applyAtProcessCreation: true } });
     }
 
-    append(variable: string, value: string): void {
-        this._setIfDiffers(variable, { value, type: EnvironmentVariableMutatorType.Append });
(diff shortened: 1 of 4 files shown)


tsmaeder:
Here's the extension:

[envvartest-0.0.1.zip](https://github.com/eclipse-theia/theia/files/12848038/envvartest-0.0.1.zip)

and here's the source:

[envvartest.zip](https://github.com/eclipse-theia/theia/files/12848050/envvartest.zip)

martin-fleck-at (approved):
Thank you @tsmaeder! I checked the code with your extension and everything works as expected. Values are properly handled based on their mutator options, meaning that in our case if you have `applyAtProcessCreation` set to false explicitly, they will not be available and `applyAtShellIntegration` is only stubbed/ignored.

Issue #12977, opened 2023-10-06
 Preloading Localization Occurs After Partial Internal Content Localization, Resulting in Localization Failure


likesubject:
### Bug Description:

When preloading localization is performed during the `load` process, it causes a failure in localizing the commands within `CommonCommands`. This issue arises because the first import statement in the code snippet already attempts to localize the commands within `CommonCommands` before the localization is fully loaded.

```javascript
const { messagingFrontendModule } = require('@theia/core/lib/electron-browser/messaging/electron-messaging-frontend-module');
const container = new Container();
container.load(messagingFrontendModule);
```

This sequence of code triggers the localization process, but since the localization hasn't completed yet, it prevents proper localization for the commands within `CommonCommands`.

To resolve this issue, it is necessary to ensure that the localization process completes before attempting to localize any content, including the commands within `CommonCommands`.

### Steps to Reproduce:

1.When setting the language to Chinese, I noticed that some translations are not working properly.

### Additional Information

- Operating System: windows10 electron
- Theia Version: v1.42.1

msujew:
I'll look into this once I'm back from vacation.

PR #12993, opened 2023-10-12
Fix localization import order

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/12977

Fixes two localization based issues, see the first two commits for more info. The issue came from the fact that we now import a bit more code during our preload script and accidentally imported code that was supposed to be translated during its import time - effectively removing the ability to translate those files.

The same thing happened for monaco, where we import code from `@theia/monaco-editor-core` before being able to localize the monaco texts. This code has been moved to the correct package.

#### How to test

1. Open Theia (Electron) and change the locale
2. Assert that common commands and monaco commands are localized

#### Review checklist

#### Reminder for reviewers


planger (approved):
Thank you, this change looks fine to me!

I can confirm that on `master` several translations don't work in a Theia Electron app anymore and with this change they work again just fine. :+1: Also, the move of the language registration to the Monaco extension makes more sense now in general.

Thank you for this fix @msujew!

Issue #12821, opened 2023-08-10
Cannot "Connect to Github"


tsmaeder:
The very useful Gitlens extensions offers to "Connect to Github" in order to directly jump to the relevant PR on github from an editor hover. This simply does not work.

# To Reproduce
1. Install gitlens
2. Open an editor on a file in a github repo, for example theia itself.
3. When you hover at the end of a line, you get a hover that shows you information about the commit that last changed this line. 
4. Scroll to the bottom of the hover
5. Observe: there is a link that says "Connect to Github".
6. Click the link
7. Observe: nothing happens (not even messages in the browser log)
My expectation would be that I would be taken through github authentication and then back to see links to PR's in the hover.

vince-fugnitto:
@tsmaeder it perhaps relies on the [`vscode.github`](https://open-vsx.org/extension/vscode/github) and [`vscode.github-authentication`](https://open-vsx.org/extension/vscode/github-authentication) builtin plugins which are [not included](https://github.com/eclipse-theia/theia-blueprint/blob/bf31e6c3659c3672bd7b0494b1a827f00a6933d1/applications/electron/package.json#L145-L146) in blueprint?

tsmaeder:
That is entirely possible. Any reason why we're _not_ including all the built-ins from VS Code?

vince-fugnitto:
> That is entirely possible. Any reason why we're _not_ including all the built-ins from VS Code?

At least for the [main repo](https://github.com/eclipse-theia/theia/blob/f75c9a244aac878c1a6582346b0c150938480d00/package.json#L110-L113) they are excluded since they [rely](https://github.com/microsoft/vscode/blob/6c683c327ca8656de700552a519927d9c7316226/extensions/github/package.json#L19-L21) on the `git` builtins. It's possible that it is something that was copied over to blueprint from the main repo.

marcdumais-work:
On vscode, an external browser is opened opwards GitHub, when the "Allow" button is pressed. 

![image](https://github.com/eclipse-theia/theia/assets/25749063/f0a108c0-29e6-4d37-a71a-b60228ea0b58)

I think this operation fails on Theia because we have a stub for `proposed` "ExternalUriOpener":

https://github.com/eclipse-theia/theia/blob/master/packages/plugin/src/theia.proposed.externalUriOpener.d.ts

AlexandraBuzila:
@JonasHelming I can investigate this

JonasHelming:
[...]

PR #12997, opened 2023-10-12
Support string argument in `vscode.open`

### LINKED ISSUE #12996: Allow `vscode.open` command to accept string URI
In vscode the [`vscode.open`](https://github.com/microsoft/vscode/blob/9d32835bd7b2facd133f63287c10910b294a26d8/src/vs/workbench/api/common/extHostApiCommands.ts#L398-L409) command - often used by extensions - allows to pass a string as the URI parameter.

Using such an extension in Theia will result in an error when the command is executed since it only allows to pass a URI.

msujew:
#### What it does

Closes #12996

#### How to test

1. Install the [CodeGPT Chat](https://open-vsx.org/extension/DanielSanMedium/dscodegpt) extension
2. Open the webview and click on the "Connect to CodeGPT Plus" button
3. Theia should open the URL provided in an external browser window

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
+++ b/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
@@ -229,7 +229,10 @@ export class PluginVscodeCommandsContribution implements CommandContribution {
     registerCommands(commands: CommandRegistry): void {
         commands.registerCommand(VscodeCommands.OPEN, {
             isVisible: () => false,
-            execute: async (resource: URI, columnOrOptions?: ViewColumn | TextDocumentShowOptions) => {
+            execute: async (resource: URI | string, columnOrOptions?: ViewColumn | TextDocumentShowOptions) => {
+                if (typeof resource === 'string') {
+                    resource = URI.parse(resource);
+                }
                 try {
                     await this.openWith(VscodeCommands.OPEN.id, resource, columnOrOptions);
                 } catch (error) {


packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
```diff
@@ -229,7 +229,10 @@ export class PluginVscodeCommandsContribution implements CommandContribution {
     registerCommands(commands: CommandRegistry): void {
         commands.registerCommand(VscodeCommands.OPEN, {
             isVisible: () => false,
-            execute: async (resource: URI, columnOrOptions?: ViewColumn | TextDocumentShowOptions) => {
+            execute: async (resource: URI | string, columnOrOptions?: ViewColumn | TextDocumentShowOptions) => {
+                if (typeof resource === 'string') {
+                    resource = URI.parse(resource);
```
JonasHelming, line 234: What happens, if the String does not parse into a URL? In case a URI is not openable, we have logging below, does this alos work for a malformed URI passed in as a String?

JonasHelming:
This might be related to https://github.com/eclipse-theia/theia/issues/12821 @AlexandraBuzila

packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
msujew: [`URI.parse`](https://github.com/microsoft/vscode-uri/blob/53e4ca6263f2e4ddc35f5360c62bc1b1d30f27dd/src/uri.ts#L266-L279) never throws, the URI will just be unusable. So we should be good in that regard. AFAIK vscode also doesn't validate the input when parsing from a string.

JonasHelming (approved):

Issue #13013, opened 2023-10-20
Support VS Code Default Language Icons


martin-fleck-at:
### Feature Description:

VS Code supports the definition of default language icons: https://code.visualstudio.com/api/extension-guides/file-icon-theme#language-default-icons that can be specified as part of the general language contribution. Such an icon should be used in the explorer and show up in the tab of the active editor. 

Theia currently only has some hard-coded default icons.

Themes are still allowed to override those language icons.

This has also been partially requested in https://github.com/eclipse-theia/theia/issues/3751

PR #13014, opened 2023-10-20
Support VS Code Default Language Icons

martin-fleck-at:
#### What it does

Add support for default language icons similar to VS Code:
- Extend language service with necessary methods
- Create CSS rules for contributed language icons
- Use language icons if theme sets 'showLanguageModeIcons'
- Ensure custom theme can override language icons

Fixes https://github.com/eclipse-theia/theia/issues/13013

#### How to test

- Provide a plugin with a language contribution that adds a custom default language icon, ideally a "known" language like python
- Install [Material Icon Theme](https://open-vsx.org/extension/PKief/material-icon-theme) which already provides custom icons for the same language.
- Switch file icon themes to see that everything is working as expected:
  - `None`: no icons shown
  - `File Icons (Theia)`: uses our default language icon
  - `Material Icon Theme`: uses icon from the theme

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-languages.ts
+++ b/packages/monaco/src/browser/monaco-languages.ts
@@ -25,6 +25,10 @@ import { Language, LanguageService } from '@theia/core/lib/browser/language-serv
 import { MonacoMarkerCollection } from './monaco-marker-collection';
 import { ProtocolToMonacoConverter } from './protocol-to-monaco-converter';
 import * as monaco from '@theia/monaco-editor-core';
+import { FileStat } from '@theia/filesystem/lib/common/files';
+import { FileStatNode } from '@theia/filesystem/lib/browser';
+import { ILanguageService } from '@theia/monaco-editor-core/esm/vs/editor/common/languages/language';
+import { StandaloneServices } from '@theia/monaco-editor-core/esm/vs/editor/standalone/browser/standaloneServices';
 
 export interface WorkspaceSymbolProvider {
     provideWorkspaceSymbols(params: WorkspaceSymbolParams, token: CancellationToken): MaybePromise<SymbolInformation[] | undefined>;
@@ -37,6 +41,7 @@ export class MonacoLanguages implements LanguageService {
     readonly workspaceSymbolProviders: WorkspaceSymbolProvider[] = [];
 
     protected readonly markers = new Map<string, MonacoMarkerCollection>();
+    protected readonly icons = new Map<string, string>();
 
     @inject(ProblemManager) protected readonly problemManager: ProblemManager;
     @inject(ProtocolToMonacoConverter) protected readonly p2m: ProtocolToMonacoConverter;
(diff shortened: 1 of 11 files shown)


packages/monaco/src/browser/monaco-languages.ts
```diff
@@ -81,10 +86,61 @@ export class MonacoLanguages implements LanguageService {
         return this.mergeLanguages(monaco.languages.getLanguages().filter(language => language.id === languageId)).get(languageId);
     }
 
+    detectLanguage(obj: unknown): Language | undefined {
```
msujew, line 89: Minor: We should probably return an explicit `undefined` at the end of the method.

packages/monaco/src/browser/monaco-languages.ts
```diff
@@ -81,10 +86,61 @@ export class MonacoLanguages implements LanguageService {
...
+        }
+        if (FileStat.is(obj)) {
+            return this.detectLanguageByURI(obj.resource);
+        }
+        if (FileStatNode.is(obj)) {
+            return this.detectLanguageByURI(obj.uri);
+        }
+    }
+
+    protected detectLanguageByIdOrName(obj: string): Language | undefined {
```
msujew, line 107: Minor: Explicit return

packages/monaco/src/browser/monaco-languages.ts
```diff
@@ -81,10 +86,61 @@ export class MonacoLanguages implements LanguageService {
...
+            return languageById;
+        }
+
+        const languageId = this.getLanguageIdByLanguageName(obj);
+        if (languageId) {
+            return this.getLanguage(languageId);
+        }
+    }
+
+    protected detectLanguageByURI(uri: URI): Language | undefined {
```
msujew, line 119: Minor: Explicit return

packages/plugin-ext/src/common/plugin-protocol.ts
```diff
@@ -300,6 +300,7 @@ export interface PluginPackageLanguageContribution {
     aliases?: string[];
     mimetypes?: string[];
     configuration?: string;
+    icon: IconUrl;
```
msujew, line 303: Suggestion: I don't think this property is mandatory. Missing a `?`?

packages/plugin-ext/src/main/browser/plugin-icon-theme-service.ts
```diff
@@ -387,7 +409,8 @@ export class PluginIconTheme extends PluginIconThemeDefinition implements IconTh
     protected collectSelectors(
         associations: RecursivePartial<PluginIconsAssociation>,
         accept: (definitionId: string, ...icons: string[]) => void
-    ): void {
+    ): { hasSpecificFileIcons: boolean; coveredLanguages: { [languageId: string]: boolean } } {
+        const iconInfo = { hasSpecificFileIcons: false, coveredLanguages: {} as { [languageId: string]: boolean } };
```
msujew, line 413: Suggestion: We should probably have a more explicit type for this
```suggestion
    ): PluginLanguageIconInfo {
        const iconInfo: PluginLanguageIconInfo  = {
            hasSpecificFileIcons: false, 
            coveredLanguages: {}
        };
```

msujew (commented):
> Provide a plugin with a language contribution that adds a custom default language icon, ideally a "known" language like python

Can you provide such a plugin? As far as I can tell, the builtin plugins don't contribute such an icon to any language.

packages/plugin-ext/src/main/browser/plugin-icon-theme-service.ts
martin-fleck-at: Good point, I fixed it.

packages/plugin-ext/src/common/plugin-protocol.ts
martin-fleck-at: You are absolutely right! Thank you for catching that!

packages/monaco/src/browser/monaco-languages.ts
martin-fleck-at: Done.

packages/monaco/src/browser/monaco-languages.ts
martin-fleck-at: Done.

martin-fleck-at:
@msujew Thank you for having a look! I pushed an example to https://github.com/eclipse-theia/theia/tree/issues/13013_sample or you can download the zipped [vsix](https://github.com/eclipse-theia/theia/files/13071878/plugin-a-1.42.0.zip).

language-icon

tsmaeder:
@msujew are you reviewing this one?

martin-fleck-at:
I rebased the branch onto main as there was a conflict in `scanner-theia.ts`.

msujew (approved):
Thanks for the example plugin. This change works well for me 👍 

I've encountered a very minor issue during testing. After installing the extension (but before changing the color/icon theme), any open editor keeps the old icon (while it correctly updates in the navigator). After changing to another theme, everything is updated as expected. This shouldn't stop this change from getting merged though.

martin-fleck-at:
@msujew Very good catch! I added a commit that should fix that problem. It was just adding an event so I thought it would be faster than to create a follow up issue ;-)

msujew (approved):
Awesome, thanks for the update @martin-fleck-at 👍 

The new changes look good as well.

packages/plugin-ext/src/main/browser/plugin-shared-style.ts
```diff
@@ -98,30 +103,44 @@ export class PluginSharedStyle {
...
-            iconClass,
-            dispose: () => toDispose.dispose()
-        };
+        let iconClass = 'plugin-icon-' + this.iconSequence++;
+        if (type === 'icon') {
+            toDispose.push(this.insertRule('.' + iconClass + '::before', theme => `
+                    content: "";
+                    background-position: 2px;
+                    width: ${size}'px';
+                    height: ${size}'px';
```
bvenreply, line 129: Hello @martin-fleck-at, what is the intent behind the addition of these single quotes? I don't think this is valid syntax, these properties are just being discarded from the css stylesheet at the moment.

I'm taking a look in the context of #13014 btw.

Issue #13004, opened 2023-10-17
Language pack localizations are not applied for launch config attributes


prusakov-a1:
### Bug Description:
It was noticed that some localization messages from VS code language pack were applied and some are not. Initially it was discovered in custom IDE based on `Theia v.1.39.0` and language pack `MS-CEINTL.vscode-language-pack-ru v.1.78.0` with next configuration settings:

```
"theiaPlugins": {
        "MS-CEINTL.vscode-language-pack-ru": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-ru/1.78.0/file/MS-CEINTL.vscode-language-pack-ru-1.78.0.vsix"
}
```

Here we can see that the tooltip for `configuration` attribute is localized properly and have expected Russian localization:
![screenshot1_ru](https://github.com/eclipse-theia/theia/assets/148201588/558ff08a-b41f-4edf-91a6-aac272ff8d71)

But if we move mouse cursor over `name` attribute then the tooltip will have default English localization:
![screenshot2_ru](https://github.com/eclipse-theia/theia/assets/148201588/bf0ad20d-734d-4e89-81ac-46ef15d2f952)
The same behavior was reproduced with latest `Theia v.1.42.1` and latest language pack `MS-CEINTL.vscode-language-pack-ru v.1.83.0`.

### Steps to Reproduce:

1. Enable VS Code localization by adding configuration for any language, e.g.:
```
"theiaPlugins": {
        "MS-CEINTL.vscode-language-pack-ru": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-ru/1.78.0/file/MS-CEINTL.vscode-language-pack-ru-1.78.0.vsix"
}
```
2. Call `Command pallet` by `Ctl+Shift+P`.
3. Select `configure Display Language` command.
4. Select previously installed localization from the list.
5. Go to `Debug`.
6. Add new configuration via drop down menu.
7. Move mouse over `configurations`, `name` attributes and compare locales.

### Additional Information

- Operating System: Windows 11
- Theia Version: 1.39.0, 1.42.1

PR #13017, opened 2023-10-21
Correctly localize debug schema attributes

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/13004

Changes the behavior of the debug contributions so that the configurations are filled on the frontend instead of the backend. That allows access to the user locale to translate any description values.

Also eliminates a lot of dead/duplicated code in `vscode-debug-adapter-contribution`. The file appears to be unused and I don't expect any adopters to make use of the file.

#### How to test

1. Install a language pack and switch to the new language
2. Hover over the "builtin" attributes of the debug launch configs, such as `type` or `name`.
3. They should be correctly localized to the current language.

#### Review checklist

#### Reminder for reviewers


planger (approved):
Thanks @msujew! Works fine and code looks good to me! Also thanks for the cleanup of the dead code.

PR #13020, opened 2023-10-23
Add support for provideDocumentRangesFormattingEdits. 

### LINKED ISSUE #12939: [vscode] Support DocumentRangeFormattingEditProvider provideDocumentRangesFormattingEdits
New API has been added to DocumentRangeFormattingEditProvider in vscode 1.82.0.

DocumentRangeFormattingEditProvider

See https://code.visualstudio.com/api/references/vscode-api#DocumentRangeFormattingEditProvider
See also https://code.visualstudio.com/updates/v1_82#_support-for-batch-range-formatting

tsmaeder:
#### What it does
Adds the optional function `Support DocumentRangeFormattingEditProvider.provideDocumentRangesFormattingEdits(...)` to the VS Code API. Since our current version of the Monaco editor does not support this function yet, the change is entirely non-functional. We'll have to provide an implementation once we've uplifted the Monaco implementation.

Fixes #12939

#### How to test
There is nothing to test, really. In order to make sure nothing is broken, I selected a range in a json file and invoked "Format Selection", which worked fine.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin/src/theia.d.ts
+++ b/packages/plugin/src/theia.d.ts
@@ -10263,6 +10263,26 @@ export module '@theia/plugin' {
             options: FormattingOptions,
             token: CancellationToken | undefined
         ): ProviderResult<TextEdit[] | undefined>;
+
+        /**
+         * Provide formatting edits for multiple ranges in a document.
+         *
+         * This function is optional but allows a formatter to perform faster when formatting only modified ranges or when
+         * formatting a large number of selections.
+         *
+         * The given ranges are hints and providers can decide to format a smaller
+         * or larger range. Often this is done by adjusting the start and end
+         * of the range to full syntax nodes.
+         *
+         * @param document The document in which the command was invoked.
+         * @param ranges The ranges which should be formatted.
+         * @param options Options controlling formatting.
+         * @param token A cancellation token.
+         * @returns A set of text edits or a thenable that resolves to such. The lack of a result can be
+         * signaled by returning `undefined`, `null`, or an empty array.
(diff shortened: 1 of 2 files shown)


martin-fleck-at (approved):
LGTM!

And thanks for fixing the typo in the CHANGELOG ;-)

packages/plugin/src/theia.d.ts
```diff
@@ -10263,6 +10263,26 @@ export module '@theia/plugin' {
...
+         * The given ranges are hints and providers can decide to format a smaller
+         * or larger range. Often this is done by adjusting the start and end
+         * of the range to full syntax nodes.
+         *
+         * @param document The document in which the command was invoked.
+         * @param ranges The ranges which should be formatted.
+         * @param options Options controlling formatting.
+         * @param token A cancellation token.
+         * @returns A set of text edits or a thenable that resolves to such. The lack of a result can be
+         * signaled by returning `undefined`, `null`, or an empty array.
```
martin-fleck-at, line 10282: Since it is only stubbed, I don't think it really matters but `undefined` or `null` are not really supported by the signature, are they?

tsmaeder: It _would_ matter if we want to compile plugins against `theia.d.ts`, but `ProviderResult<T>` is defined as:

``` 
export type ProviderResult<T> = T | undefined | null | Thenable<T | undefined | null>;
```

martin-fleck-at: Ah, yes, I knew I was missing something obvious, thanks!

Issue #13042, opened 2023-11-01
Electron: Dialogs are not modal by default


zole40:
### Bug Description:
There was a similar problem more than 1 year ago, which has been fixed.. https://github.com/eclipse-theia/theia/issues/10768
But there is a regression in later versions, and the dialogs are not modal by default but should be by this discussion. https://github.com/eclipse-theia/theia/discussions/13037#discussioncomment-7436537

### Steps to Reproduce:

1. Go to the File menu, and select the Open File... menu entry. A dialog to select the file should open.
2. Click outside the dialog, e.g. in the main window
3. The dialog goes into the background and a new dialog can be opened via the Open File... command, while the first one is still open.

### Additional Information

- Operating System: Win 10.
- Theia Version: 1.43.0

bryanchenmchp:
Hi @zole40, could you close this issue to indicate that it's been resolved?

PR #13043, opened 2023-11-01
Set electron dialogs to modal by default

zole40:
#### What it does

I found regression in related to the electron dialogs https://github.com/eclipse-theia/theia/issues/13042
#### How to test
1. Build the electron example
2. Run electron example 
3. Go to the file menu and click on the open folder

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
+++ b/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts
@@ -107,7 +107,7 @@ export class ElectronFileDialogService extends DefaultFileDialogService {
         result.title = props.title;
         result.buttonLabel = props.openLabel;
         result.maxWidth = props.maxWidth;
-        result.modal = props.modal;
+        result.modal = props.modal ?? true;
         result.openFiles = props.canSelectFiles;
         result.openFolders = props.canSelectFolders;
         result.selectMany = props.canSelectMany;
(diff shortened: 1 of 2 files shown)


AlexandraBuzila (commented):
Hi @zole40! Thank you for the fix, it works as expected :+1: 

However, I think it would be better if the `modal` prop was already initialized in the `toOpenDialogOptions` and `toSaveDialogOptions` from the [ElectronFileDialogService](https://github.com/eclipse-theia/theia/blob/f33547ef9e2521b5a7e3507badaaf34b35da5d70/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts#L102-L157), together with the other dialog options.

zole40:
> Hi @zole40! Thank you for the fix, it works as expected 👍
> 
> However, I think it would be better if the `modal` prop was already initialized in the `toOpenDialogOptions` and `toSaveDialogOptions` from the [ElectronFileDialogService](https://github.com/eclipse-theia/theia/blob/f33547ef9e2521b5a7e3507badaaf34b35da5d70/packages/filesystem/src/electron-browser/file-dialog/electron-file-dialog-service.ts#L102-L157), together with the other dialog options.

Should we set a default value if it is undefined?
For example: `result.modal = props.modal ?? true;`

AlexandraBuzila:
> Should we set a default value if it is undefined? For example: `result.modal = props.modal ?? true;`

Yes, exactly. That would ensure that the options already have the default value for the `modal` prop when it is checked in `electron-api-main.ts`

AlexandraBuzila (approved):
Thank you, it looks good to me! 

@JonasHelming could you have a look as well, please?

tsmaeder (approved):

PR #13057, opened 2023-11-06
fix: do not reject unknown schemes in WindowStateExt.asExternalUri

### LINKED ISSUE #12821: Cannot "Connect to Github"

AlexandraBuzila:
#### What it does

Fixes #12821

After testing the same workflow of the Gitlens extension in VS Code, I think URLs with unknown schemes were rejected prematurely in Theia.
When the `Connect to Github` action is triggered, the extension invokes two methods of the `WindowStateExtImpl`: first [asExternalUri](https://github.com/eclipse-theia/theia/blob/f33547ef9e2521b5a7e3507badaaf34b35da5d70/packages/plugin-ext/src/plugin/window-state.ts#L56) and at a later point [openUri](https://github.com/eclipse-theia/theia/blob/f33547ef9e2521b5a7e3507badaaf34b35da5d70/packages/plugin-ext/src/plugin/window-state.ts#L52).

The parameter passed to the `asExternalUri` method is an URI of the form `theia://...` and the existing code rejects it. 
In VS Code, the URI has an application specific schema as well (not http or https), but the code does not reject it. This allows the Github authentication to process the request and later on call the `openUri` method with a valid URI.

I modified the code and it now handles URIs similar to VS Code. This fixes the initial problem, which was reported for the gitlens extension, but I would love some feedback as I can't tell if this is generally safe for all extensions.

#### How to test
1. Install gitlens. I had to remove the vscode.git and vscode.github extensions from the [excluded plugins list](https://github.com/eclipse-theia/theia/blob/f33547ef9e2521b5a7e3507badaaf34b35da5d70/package.json#L114C4-L117C36) and removed the '@theia/git' package to make this work with the Theia examples.
2. Open an editor on a file in a github repo, for example theia itself.
3. When you hover at the end of a line, you get a hover that shows you information about the commit that last changed this line.
4. Scroll to the bottom of the hover
5. Observe: there is a link that says "Connect to Github".
6. Click the link
7. There's a popup that says "You have not yet finished authorizing this extension ...". Click 'Yes'.
8. Another popup shows up - click "Copy & Continue to Github".
9. A Github link should open in the browser.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/window-state.ts
+++ b/packages/plugin-ext/src/plugin/window-state.ts
@@ -19,7 +19,6 @@ import { WindowState } from '@theia/plugin';
 import { WindowStateExt, WindowMain, PLUGIN_RPC_CONTEXT } from '../common/plugin-api-rpc';
 import { Event, Emitter } from '@theia/core/lib/common/event';
 import { RPCProtocol } from '../common/rpc-protocol';
-import { Schemes } from '../common/uri-components';
 
 export class WindowStateExtImpl implements WindowStateExt {
 
@@ -50,15 +49,16 @@ export class WindowStateExtImpl implements WindowStateExt {
     }
 
     openUri(uri: URI): Promise<boolean> {
+        if (!uri.scheme.trim().length) {
+            throw new Error('Invalid scheme - cannot be empty');
+        }
+
         return this.proxy.$openUri(uri);
     }
 
     async asExternalUri(target: URI): Promise<URI> {
         if (!target.scheme.trim().length) {
             throw new Error('Invalid scheme - cannot be empty');
-        }
-        if (Schemes.http !== target.scheme && Schemes.https !== target.scheme) {
-            throw new Error(`Invalid scheme '${target.scheme}'`);
         }
 
         const uri = await this.proxy.$asExternalUri(target);


packages/plugin-ext/src/plugin/window-state.ts
```diff
@@ -50,16 +50,20 @@ export class WindowStateExtImpl implements WindowStateExt {
     }
 
     openUri(uri: URI): Promise<boolean> {
+        if (!uri.scheme.trim().length) {
+            throw new Error('Invalid scheme - cannot be empty');
+        }
+
+        if (Schemes.http !== uri.scheme && Schemes.https !== uri.scheme) {
```
tsmaeder, line 57: T.b.h, I don't think it makes sense to check the scheme at all here: see the definition of `openExternal` in https://code.visualstudio.com/api/references/vscode-api#env. It's explicitly mentioned that an URI could be of the form `mailto://...`.

packages/plugin-ext/src/plugin/window-state.ts
```diff
@@ -50,16 +50,20 @@ export class WindowStateExtImpl implements WindowStateExt {
...
+            throw new Error(`Invalid scheme '${uri.scheme}'`);
+        }
         return this.proxy.$openUri(uri);
     }
 
     async asExternalUri(target: URI): Promise<URI> {
         if (!target.scheme.trim().length) {
             throw new Error('Invalid scheme - cannot be empty');
         }
-        if (Schemes.http !== target.scheme && Schemes.https !== target.scheme) {
```
tsmaeder, line 60: If I read the documentation at https://code.visualstudio.com/api/references/vscode-api#env correctly, `asExternalUri` treats `http`, `https` and the value of `vscode.env.uriScheme` (I suspect it's `theia`) specially and converts anything else into a workspace uri. I don't think that is what's happening on the main side. Removing the check is correct, IMO, though.

tsmaeder (changes requested):

packages/plugin-ext/src/plugin/window-state.ts
AlexandraBuzila: Thank you for the review, Thomas! Good point, I removed the check.

JonasHelming:
@AlexandraBuzila Is this ready for a re-review?

tsmaeder:
The "Connect to Github..." workflow still does not work for me: when I invoke it, nothing happens and I get a dialog saying
```
Sie haben die Autorisierung dieser Erweiterung für die Verwendung von GitHub noch nicht abgeschlossen. Möchten Sie eine andere Methode ausprobieren? (Gerätecode)
```

I would expect to be redirected to the github authentication page.

AlexandraBuzila:
> I would expect to be redirected to the github authentication page.

Does the link open in the browser if you confirm that dialog?

I see now that the workflow in vscode looks different: you are asked to log in, while in theia you are asked to use another login method. I can check to see where the difference is coming from.

JonasHelming:
@tsmaeder Opne question above :-)

tsmaeder (approved):
After retesting, the login seems to work now.

planger:
I'm very much looking forward to having this up and running.

So I tried with today's master, removed the git and github related extensions from `theiaPluginsExcludeIds` and tried the Github view as well as exactly the steps from this PR with Gitlens. When trying to connect to Github, I just received this though:
image

No browser window was opened. Do you have any suggestion @AlexandraBuzila? Thank you!

AndKe:
Joining in as I too am very interested in this feature and I stop at the same issue as the post above.

planger:
@AlexandraBuzila @tsmaeder Do you have feedback on the issue mentioned above? Should we re-open this issue?

tsmaeder:
@AndKe how exactly is this failing for you? What's your setup for running Theia?

tsmaeder:
@AndKe arduinoIDE is still on release 1.41.0. The fix is not in that version.

AndKe:
@tsmaeder are you saying that is should work in >= 1.41 ? / and where can I get that version?

per1234:
@AndKe you must have forgotten the details of the [discussion](https://forum.arduino.cc/t/arduino-in-vscode-without-the-redundant-c-bs/1244744/13) we had about this over on Arduino Forum. The "arduino-ide_2.3.3-snapshot-de5b91f_Linux_64bit.AppImage" you are using is a beta tester build of Arduino IDE with Theia version 1.46.1 (https://github.com/arduino/arduino-ide/commit/de5b91f480b919a9e6feb2acf14958e77da191d8):

https://github.com/arduino/arduino-ide/actions/runs/8596568336#artifacts

tsmaeder:
@AndKe as the PR was merged in Dec., it should be in 1.46.1. 

> I started with copying my working plugins from vscode:

This is not a good idea: when we download plugins, we select a version of the built-ins that works with that particular Theia version.

As for the gitlens example from the testing instructions, I don't even get the "Connect to Github" link anymore. Reopening.

Issue #12917, opened 2023-09-18
`Run build task` is failing to collect tasks


VGRSTM:
### Bug Description:
`Task: Run Build Task` command is failing to collect build & test tasks out of the ones `Task: Run Task` is able to
PS: VSCode is able to

### Steps to Reproduce:
1. Just open a  Theia instance
2. Create a build type task running `Task: Configure Tasks ...` electing MSBuild as example. Save promoted tasks.json file.
2. Get a look at either `Task: Run Task` command or rely on `Terminal > Run Task` menu collected tasks. Just created `build` task is promoted.
3. Get a look at either `Task: Run Build Task` command or rely on `Terminal > Run Build Task` menu. No task is promoted .. `build` was expected one at least.

### Additional Information

- Operating System: All
- Theia Version: 1.41.0 (and older ones)

planger:
Hi @VGRSTM,
thank you very much for your bug report. I was starting to review #12918 and first tried to reproduced this bug on `master` and `1.41.0`, but was unable to do that. Below is what I get in both versions:

https://github.com/eclipse-theia/theia/assets/588090/af28861a-3203-4c6c-bfa4-d2c925755d64

Am I doing something wrong?

VGRSTM:
@planger you're right I'm not able neither right now to reproduce issue as described  (1.41.0 tag) ??? I'm ending with same as you.

What I'm able to reproduce (and this is a behavior delta to VSCode regarding build task collection) is editing generated task you've got moving from `"group": "build"` to `"group": {"kind": "build", "isDefault": false}`. 
My guess is for one reason or another at time I've reported this issue I was getting this `group` object syntax.

planger:
Right, I can confirm two differences with respect to VS Code:
* If there is just one build task `"group": "build"` VS Code directly runs it without asking the user. Theia shows this task and the user has to confirm.
* Theia considers a task with `"group": {"kind": "build","isDefault": false}` to be invalid, while VS Code just treats it like a task with `"group": "build"`.

Is this the same you are seeing, or did you refer to another issue I didn't see so far?

VGRSTM:
I'm seeing the same. My issue is behavior delta especially the one about invalidating `"group": {"kind": "build","isDefault": false}`. 
[...]

PR #13075, opened 2023-11-13
fix(task): Align task group validation with VS Code

planger:
#### What it does

In VS Code, the following is considered correct in a `tasks.json`:
```
"group": {"kind": "build","isDefault": false}
```

In Theia, however, this is considered an error as the enumeration in the JSON Schema restricts valid values to `isDefault: true`. This also prevents such tasks to be listed correctly when e.g. executing the command "Run Build Task".

Fixes https://github.com/eclipse-theia/theia/issues/12917

#### How to test

* Create a `tasks.json` with the following content:
```json
{
    "version": "2.0.0",
    "tasks": [
        {
            "label": "echo",
            "type": "shell",
            "command": "echo Hello",
            "group": {
                "kind": "build",
                "isDefault": true
            }
        }
    ]
}
```
* Observe how it is considered valid and automatically executes when the user executes "Run Build Task"
* Change the task to below and observe how it is still considered valid and shows up as a suggestion when executing "Run Build Task".
```json
        {
            "label": "echo",
            "type": "shell",
            "command": "echo Hello",
            "group": {
                "kind": "build",
                "isDefault": false
            }
        }
```

#### Follow-ups

There is another difference to the behavior of VS Code:
If there is only one task in the group, it'll be considered the default task and executed automatically. In Theia, the user is still prompted which task to execute even though there is just one. Not sure if we want to change that though.

#### Review checklist

#### Reminder for reviewers


msujew (approved):
Thanks, that works way better and makes plainly more sense 👍

PR #13028, opened 2023-10-26
Update nls metadata for VSCode API 1.82.0

msujew:
#### What it does

Updates the nls metadata for the latest API bump https://github.com/eclipse-theia/theia/pull/13025.

#### How to test

CI is green, no translation warnings are displayed on application startup.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #13078, opened 2023-11-16
Improve default translation using preferred keys

msujew:
#### What it does

Related to a comment in https://github.com/eclipse-theia/theia/pull/13028.

Adds support for entering preferred translation keys for the nls.metadata.json translation process. This issue in particular is seen because there are ~5 different entries that translate to `File` in the metadata file. The last one is translated differently compared all others. This isn't an issue usually, since all entries that use the same english value translate to the same value in any other language.

For this case, we now have a set of "preferred keys" that have priority over other keys. This currently only seems to be necessary for the `File` entry, but contributors can any keys in there to improve localization.

#### How to test

1. Change the locale of the application to something other than English
2. The "File" menu entry should be translated to whatever "File" in your selected language is. It shouldn't show the translation for "local file".

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/nls.ts
+++ b/packages/core/src/common/nls.ts
@@ -81,6 +81,10 @@ interface NlsInfo {
 
 class LocalizationKeyProvider {
 
+    private preferredKeys = new Set([
+        // We only want the `File` translation used in the menu
+        'vscode/fileActions.contribution/filesCategory'
+    ]);
     private data = this.buildData();
 
     get(defaultValue: string): string | undefined {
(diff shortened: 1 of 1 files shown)


martin-fleck-at (approved):
Change looks good to me and definitely solves the problem with the 'File' menu entry having the wrong translation. So we can definitely merge this, thank you Mark!

Just to make it very clear: The idea here is to have Theia maintain this list of preferred keys, is that right? Cause there is no way as an application to contribute to this list or override/extend the LocalizationKeyProvider.

msujew:
> The idea here is to have Theia maintain this list of preferred keys, is that right? Cause there is no way as an application to contribute to this list or override/extend the LocalizationKeyProvider.

Yes, adopters shouldn't really use `nls.localizeByDefault` (which is the only consumer of this API) anyway. Adopters should provide their own translations. We just do it because we would need to provide 1000+ translations, which are contributed by language packs anyway.

martin-fleck-at:
@msujew Great, thank you for the clarification!

PR #13079, opened 2023-11-17
fix: allow all listener to be executed for AsyncEmitter

Maddobun:
#### What it does
With AsyncEmitters, if a listener callback does not call `event.waitUntil`, other listeners that are registered after this one will fail to be called due to early loop termination.

#### How to test
Register two separate callbacks without `waitUntil` for `FileService.onDidRunUserOperation` or any other `AsyncEmitter.event` and fire an event. Verify that both callbacks are called correctly.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/event.ts
+++ b/packages/core/src/common/event.ts
@@ -456,7 +456,7 @@ export class AsyncEmitter<T extends WaitUntilEvent> extends Emitter<T> {
                 delete (asyncEvent as any)['waitUntil'];
             }
             if (!waitables.length) {
-                return;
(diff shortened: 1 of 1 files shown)


tortmayr (approved):
Thanks! Looks good to me.
AsyncEmitters now work as expected and all listener callback are invoked when an event is fired even if some of them don't use `Event.waitUntil` in the callback.
Nice catch!

JonasHelming:
@tortmayr Thank you for the reivew, do you want to merge?

PR #13095, opened 2023-11-23
Incorrect "Unsupported activation event" error in stdout #12953

### LINKED ISSUE #12953: Theia can incorrectly generate "Unsupported activation event" error in stdout
### Bug Description:
A Theia product that introduces new plugin API can also introduce new [activation events](https://code.visualstudio.com/api/references/activation-events)

Introducing a new activation is quite straightforward. You really just call `HostedPluginSupport.activateByEvent(activationEvent: string)` from within your Theia extension to effectively "generate the event". Theia's implementation of that function simply activates any plugins that have that string declared in their package.json as an `activationEvent`. So really, it's just a meeting of the mind between your product-specific Theia extension and the developers writing plugins for your product. If you two are on the same page, then it all just works., You don't need your Theia extension to somehow register the new activation event with the plugin mechanism. The activation event is ultimately just an opaque string to the plugin mechanism.

But...Theia tries to warn folks installing VS Code extensions when an extension declares an `activationEvent` that the Theia platform itself doesn't support (doesn't generate). And to do that, it has to have knowledge of what activation events it supports. That static list is harcoded into Theia [here](https://github.com/eclipse-theia/theia/blob/990908be18b5045280d2fa12c2be12dc3c7bdb82/packages/plugin-ext/src/plugin/plugin-manager.ts#L82)

Problem is, it doesn't consider that a product developer might be introducing new activation events. And so while it will infact support any and all foreign activation events, it will end up warning the user via stdout that it _doesn't_ support them. And that's obviously not right. We need some way for a Theia extension to tell the aforementioned code: "hey, here are some additional activation event that I support", so that it doesn't generate that error msg in stdout.

### Steps to Reproduce:
```
git clone -b activation_event_gap https://github.com/jcortell68/theiasandbox
cd theiasandbox
yarn
yarn start:electron
```
Note the error message in stdout
```
root ERROR [hosted-plugin: <pid>] Unsupported activation events: onSayHello, please open an issue: https://github.com/eclipse-theia/theia/issues/new
```
Also note that the activation event works quite well despite that misguided message :-)

jfaltermeier:
#### What it does
In this PR we are adding a mechanism to add additional known activation events to avoid "Unsupported activation event" messages in the log. 

In my initial implementation the information is taken from an environment variable. 
For the browser application use case the environment may be set when building the container image for the app.
For the electron application use case a custom `main` script may be used, that sets the desired environment variable value and then hands over to the generated `electron-main.js`. 
Also the user/application may influence the environment variable further if required. 

Any additional plugin installed by the user may bring in new warnings, so I was not sure if a more programmatic approach, where the activation events have to be specified at compile/build time, is working for all use cases. 
A user preference might work, but it seems a bit weird to me to have this in the settings. 
So this is my reasoning for the initial environment variable approach. 

If you would prefer a different approach to specify the events, let me know and I will adjust the PR. 

Fixes #12953

#### How to test
Run Electron Example with downloaded plugins

```
yarn && yarn download:plugins && yarn electron build && clear && yarn electron start
```

With the default plugins you should get the following warnings:

```
Unsupported activation events: onProfile, onProfile:github, please open an issue: https://github.com/eclipse-theia/theia/issues/new
Unsupported activation events: onOpenExternalUri:http, onOpenExternalUri:https, please open an issue: https://github.com/eclipse-theia/theia/issues/new
Unsupported activation events: onWalkthrough:nodejsWelcome, please open an issue: https://github.com/eclipse-theia/theia/issues/new
```

Now run the same same application with the `ADDITIONAL_ACTIVATION_EVENTS` env variable. 

```
export ADDITIONAL_ACTIVATION_EVENTS=onProfile,onOpenExternalUri,onWalkthrough && clear && yarn electron start
```

This will remove the warnings/errors.

#### Follow-ups

-

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/plugin-manager.ts
+++ b/packages/plugin-ext/src/plugin/plugin-manager.ts
@@ -79,30 +79,6 @@ class ActivatedPlugin {
 
 export class PluginManagerExtImpl implements PluginManagerExt, PluginManager {
 
-    static BUILTIN_ACTIVATION_EVENTS = new Set([
-        '*',
-        'onLanguage',
-        'onCommand',
-        'onDebug',
-        'onDebugInitialConfigurations',
-        'onDebugResolve',
-        'onDebugAdapterProtocolTracker',
-        'onDebugDynamicConfigurations',
-        'onTaskType',
-        'workspaceContains',
-        'onView',
-        'onUri',
-        'onTerminalProfile',
-        'onWebviewPanel',
-        'onFileSystem',
-        'onCustomEditor',
-        'onStartupFinished',
-        'onAuthenticationRequest',
-        'onNotebook',
-        'onNotebookSerializer'
-    ]);
-    static ADDITIONAL_ACTIVATION_EVENTS_ENV = 'ADDITIONAL_ACTIVATION_EVENTS';
-
     private configStorage: ConfigStorage | undefined;
     private readonly registry = new Map<string, Plugin>();
     private readonly activations = new Map<string, (() => Promise<void>)[] | undefined>();
@@ -222,11 +198,7 @@ export class PluginManagerExtImpl implements PluginManagerExt, PluginManager {
         this.webview.init(params.webview);
         this.jsonValidation = params.jsonValidation;
 
-        this.supportedActivationEvents = new Set(PluginManagerExtImpl.BUILTIN_ACTIVATION_EVENTS);
-        const additionalActivationEvents = await this.envExt.getEnvVariable(PluginManagerExtImpl.ADDITIONAL_ACTIVATION_EVENTS_ENV);
-        if (additionalActivationEvents) {
-            additionalActivationEvents.split(',').forEach(event => this.supportedActivationEvents.add(event));
-        }
+        this.supportedActivationEvents = new Set(params.supportedActivationEvents ?? []);
     }
 
     async $start(params: PluginManagerStartParams): Promise<void> {
--- a/packages/plugin-ext/src/plugin/plugin-manager.ts
+++ b/packages/plugin-ext/src/plugin/plugin-manager.ts
@@ -79,7 +79,7 @@ class ActivatedPlugin {
 
 export class PluginManagerExtImpl implements PluginManagerExt, PluginManager {
 
-    static SUPPORTED_ACTIVATION_EVENTS = new Set([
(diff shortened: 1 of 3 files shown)


packages/plugin-ext/src/plugin/plugin-manager.ts
```diff
@@ -219,6 +221,12 @@ export class PluginManagerExtImpl implements PluginManagerExt, PluginManager {
 
         this.webview.init(params.webview);
         this.jsonValidation = params.jsonValidation;
+
+        this.supportedActivationEvents = new Set(PluginManagerExtImpl.BUILTIN_ACTIVATION_EVENTS);
+        const additionalActivationEvents = await this.envExt.getEnvVariable(PluginManagerExtImpl.ADDITIONAL_ACTIVATION_EVENTS_ENV);
```
martin-fleck-at, line 226: I wonder whether the additional activation events shouldn't be part of the `PluginManagerInitializeParams` used in the $init call to initialize the plugin manager. It already contains the some other setup data and is also called before the $start method with the `PluginManagerStartParams` where we get the first set of activation events. I think that would also give us a more fine-grained option to handle how those events are obtained and maybe even provide different events for different plugin managers. For now I'm fine if they are still coming from environment variables, just gathered in the `hosted-plugin.ts` during initialization of the manager and added as a dedicated, optional parameter.

What do you think?

martin-fleck-at (commented):
Thank you very much Johannes! Change looks good to me in general, I'm just wondering whether we should be so focused on the environment variable approach (see inline comment),

packages/plugin-ext/src/plugin/plugin-manager.ts
jfaltermeier: Sounds good, please have another look at the changes. 
I've moved the known events + reading the env var to `hosted-plugin.ts` and extended the `PluginManagerInitializeParams` to include the supported events.

martin-fleck-at (approved):
Great, changes look good to me, thank you Johannes!

jcortell68:
Thanks, guys. @martin-fleck-at I wonder if the new-and-improved custom plugin API markdown file shouldn't mention this environment variable. Otherwise, app developers will have to debug the error message to find out about the env variable.

martin-fleck-at:
@jcortell68 That is a very good point! I'll open another PR to add some information on that.

PR #13101, opened 2023-11-27
fix: set `display: block` on plugin icon pseudo-element (#13096)

### LINKED ISSUE #13096: Custom icons are not displayed in plugin view
### Bug Description:

Custom icons are not displayed in plugin views. It seems that (some?) custom icons are displayed as `inline` elements by default, thus ignoring the `width` property.

image

### Steps to Reproduce:

The most mainstream vscode extension I could reproduce the issue with was [ms-kubernetes-tools](https://open-vsx.org/extension/ms-kubernetes-tools/vscode-kubernetes-tools). If you have no Kubernetes clusters, you can have one displayed using this dummy kubeconfig:
```yaml
apiVersion: v1
kind: Config
current-context: dev
preferences: {}
clusters:
  - cluster:
      server: https://fake-k8s.com.local
    name: dev
users:
  - name: dev
    user:
      token: dummy
contexts:
  - name: dev
    context:
      cluster: dev
      user: dev
```
Just save it somewhere, then add it with the command *Kubernetes: Set Kubeconfig*.

Then just open the Kubernetes plugin view and note the missing icon to the left of the cluster name, as compared to vscode.

### Additional Information

- Theia Version: `1.43.1`
- Browser: Chromium version `119.0.6045.159`

I believe the issue was introduced with [this change](https://github.com/eclipse-theia/theia/pull/12827/files#diff-1d4f8a4206f04c217432c91be8bf49d10df51618f61d0f53279403b29dae27b9L113-R114), in this PR: eclipse-theia/theia/pull/12827 . I tested setting the display mode explicitly to something other than `inline`, the icons were displayed correctly in that case (see image above).

bvenreply:
#### What it does

Fixes #13096

Fix the icon pseudo-element being displayed as `inline` by default. This meant that the `width` and `height` properties were being ignored and the icons displayed with 0 width.

Also fix css syntax for `width` and `height` properties.

#### How to test

The most mainstream vscode extension I could reproduce the issue with was [ms-kubernetes-tools](https://open-vsx.org/extension/ms-kubernetes-tools/vscode-kubernetes-tools). If you have no Kubernetes clusters, you can have one displayed using this dummy kubeconfig:
```yaml
apiVersion: v1
kind: Config
current-context: dev
preferences: {}
clusters:
  - cluster:
      server: https://fake-k8s.com.local
    name: dev
users:
  - name: dev
    user:
      token: dummy
contexts:
  - name: dev
    context:
      cluster: dev
      user: dev
```
Just save it somewhere, then add it with the command *Kubernetes: Set Kubeconfig*.

Then just open the Kubernetes plugin view and note the no longer missing icon to the left of the cluster name.

I also found the issue reproduced in the [gitlens plugin](https://open-vsx.org/extension/eamodio/gitlens) list of commits, here's a before/after comparison:

image

To reproduce this you can just install gitlens and open any git repository.

All this works on the Theia browser example.

#### Follow-ups

None that I can see.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/plugin-shared-style.ts
+++ b/packages/plugin-ext/src/main/browser/plugin-shared-style.ts
@@ -125,8 +125,9 @@ export class PluginSharedStyle {
             toDispose.push(this.insertRule('.' + iconClass + '::before', theme => `
                     content: "";
                     background-position: 2px;
-                    width: ${size}'px';
-                    height: ${size}'px';
+                    display: block;
+                    width: ${size}px;
+                    height: ${size}px;
                     background: center no-repeat url("${theme.type === 'light' ? lightIconUrl : darkIconUrl}");
                     background-size: ${size}px;
                 `));


planger (approved):

PR #13103, opened 2023-11-27
Avoid class-name based contribution filtering

tortmayr:
#### What it does
Update example to avoid class name based contribution filtering. Class names are minified in production which breaks the contribution filter.

Use instance of check instead. 
Also mention this in the `FilterContribution` JS-Doc

#### How to test
Open command palette and verify that the command "example_command.filtered" is still filtered out and not available.
 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/contribution-filter/contribution-filter.ts
+++ b/packages/core/src/common/contribution-filter/contribution-filter.ts
@@ -50,6 +50,15 @@ export const FilterContribution = Symbol('FilterContribution');
 export interface FilterContribution {
     /**
      * Use the registry to register your contribution filters.
+     * * Note that filtering contributions based on their class (constructor) name is discouraged.
+     * Class names are minified in production builds and therefore not reliable.
+     * Use instance of checks or direct constructor comparison instead:
+     *
+     * ```ts
+     * registry.addFilters('*', [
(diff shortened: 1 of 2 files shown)


packages/core/src/common/contribution-filter/contribution-filter.ts
```diff
@@ -50,6 +50,9 @@ export const FilterContribution = Symbol('FilterContribution');
 export interface FilterContribution {
     /**
      * Use the registry to register your contribution filters.
+     * Note that filtering contributions based on their class (constructor) name is discouraged.
+     * Class names are minified in production builds and therefore not reliable.
+     * Use instance of checks or direct constructor comparison instead.
```
msujew, line 55: It might make sense to add an example code snippet here:
```suggestion
     * Note that filtering contributions based on their class (constructor) name is discouraged.
     * Class names are minified in production builds and therefore not reliable.
     * Use instance of checks or direct constructor comparison instead:
     * 
     * ```ts
     * registry.addFilters('*', [
     *     contrib => !(contrib instanceof SampleFilteredCommandContribution)
     * ]);
     * ```
```

msujew (approved):
LGTM 👍

Issue #13065, opened 2023-11-06
[vscode] bump API compatibility version to 1.83.0


rschnekenbu:
When all implementation tasks of #13062 have been addressed, the compatibility version of the VS code API can be increased to 1.83.0.

PR #13105, opened 2023-11-27
vscode support of editor.indentSize

### LINKED ISSUE #13058: [vscode] Support TextEditorOptions#indentSize added with 1.83
new API has been added to TextEditorOptions with 1.83:

```
/**
 * The number of spaces to insert when {@link TextEditorOptions.insertSpaces insertSpaces} is true.
 *
 * When getting a text editor's options, this property will always be a number (resolved).
 * When setting a text editor's options, this property is optional and it can be a number or `"tabSize"`.
 */
indentSize?: number | string;
```

TextEditorOptions_indentSize

rschnekenbu:
#### What it does

Fix missing API TextEditorOptions.indentSize made public on vscode 1.83 (was a proposed API before). 
The behavior is available, but the preference requires an uplift of the monaco version. It is not yet available in the generated preferences. 

Fixes #13058

#### How to test 
The test can be done on a text file, using "Indent using spaces" and "Indent using tabs" commands. The behavior should be the same as before. Only the internals have changed. 
The preferences test has to wait the monaco uplift.

The following provided extension can also show the value of identSize when the value has changed with an information message:
- zip: [indent-sample-0.0.1.zip](https://github.com/eclipse-theia/theia/files/13347091/indent-sample-0.0.1.zip)
- src: [indent-sample-src-0.01.zip](https://github.com/eclipse-theia/theia/files/13347093/indent-sample-src-0.01.zip)

#### Follow-ups

Preferences need to be supported once a monaco uplift has been performed.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-text-model-service.ts
+++ b/packages/monaco/src/browser/monaco-text-model-service.ts
@@ -127,12 +127,15 @@ export class MonacoTextModelService implements ITextModelService {
 
     protected readonly modelOptions: { [name: string]: (keyof ITextModelUpdateOptions | undefined) } = {
         'editor.tabSize': 'tabSize',
-        'editor.insertSpaces': 'insertSpaces'
+        'editor.insertSpaces': 'insertSpaces',
+        'editor.indentSize': 'indentSize'
     };
 
     protected toModelOption(editorPreference: EditorPreferenceChange['preferenceName']): keyof ITextModelUpdateOptions | undefined {
         switch (editorPreference) {
             case 'editor.tabSize': return 'tabSize';
+            // @monaco-uplift: uncomment this line once 'editor.indentSize' preference is available
+            //  case 'editor.indentSize': return 'indentSize';
             case 'editor.insertSpaces': return 'insertSpaces';
             case 'editor.bracketPairColorization.enabled':
             case 'editor.bracketPairColorization.independentColorPoolPerBracketType':
@@ -170,6 +173,8 @@ export class MonacoTextModelService implements ITextModelService {
         const overrideIdentifier = typeof arg === 'string' ? undefined : arg.languageId;
         return {
             tabSize: this.editorPreferences.get({ preferenceName: 'editor.tabSize', overrideIdentifier }, undefined, uri),
+            // @monaco-uplift: when available, switch to 'editor.indentSize' preference.
+            indentSize: this.editorPreferences.get({ preferenceName: 'editor.tabSize', overrideIdentifier }, undefined, uri),
             insertSpaces: this.editorPreferences.get({ preferenceName: 'editor.insertSpaces', overrideIdentifier }, undefined, uri),
             bracketColorizationOptions: {
                 enabled: this.editorPreferences.get({ preferenceName: 'editor.bracketPairColorization.enabled', overrideIdentifier }, undefined, uri),
--- a/packages/plugin-ext/src/plugin/text-editor.ts
+++ b/packages/plugin-ext/src/plugin/text-editor.ts
@@ -277,7 +277,8 @@ export class TextEditorExt implements theia.TextEditor {
 }
 
 export class TextEditorOptionsExt implements theia.TextEditorOptions {
-    private _tabSize?: number;
+    private _tabSize: number;
+    private _indentSize: number | 'tabSize';
     private _insertSpace: boolean;
     private _cursorStyle: TextEditorCursorStyle;
     private _lineNumbers: TextEditorLineNumbersStyle;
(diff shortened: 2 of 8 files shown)


packages/monaco/src/browser/monaco-text-model-service.ts
```diff
@@ -127,12 +127,15 @@ export class MonacoTextModelService implements ITextModelService {
...
-        'editor.insertSpaces': 'insertSpaces'
+        'editor.insertSpaces': 'insertSpaces',
+        'editor.indentSize': 'tabSize'
     };
 
     protected toModelOption(editorPreference: EditorPreferenceChange['preferenceName']): keyof ITextModelUpdateOptions | undefined {
         switch (editorPreference) {
             case 'editor.tabSize': return 'tabSize';
+            // monaco-uplift: uncomment this line once 'editor.indentSize' preference is available
+            //  case 'editor.indentSize': return 'tabSize';
```
msujew, line 138: Suggestion: We use `@` before the `monaco-uplift`. Without it, this line might be missed. In line 176 as well.

packages/monaco/src/browser/monaco-text-model-service.ts
```diff
@@ -127,12 +127,15 @@ export class MonacoTextModelService implements ITextModelService {
 
     protected readonly modelOptions: { [name: string]: (keyof ITextModelUpdateOptions | undefined) } = {
         'editor.tabSize': 'tabSize',
-        'editor.insertSpaces': 'insertSpaces'
+        'editor.insertSpaces': 'insertSpaces',
+        'editor.indentSize': 'tabSize'
```
msujew, line 131: Question: Is it correct that both set the `tabSize` property of the monaco text model?

packages/monaco/src/browser/monaco-text-model-service.ts
rschnekenbu: Fixed with new commit for both lines

packages/monaco/src/browser/monaco-text-model-service.ts
rschnekenbu: Good catch! I switched indeed to 'indentSize'.

rschnekenbu:
Thanks a lot for your review, @msujew ! I updated the code with your comments, this can be reviewed again.

packages/monaco/src/browser/monaco-text-model-service.ts
```diff
@@ -127,12 +127,15 @@ export class MonacoTextModelService implements ITextModelService {
...
-        'editor.insertSpaces': 'insertSpaces'
+        'editor.insertSpaces': 'insertSpaces',
+        'editor.indentSize': 'indentSize'
     };
 
     protected toModelOption(editorPreference: EditorPreferenceChange['preferenceName']): keyof ITextModelUpdateOptions | undefined {
         switch (editorPreference) {
             case 'editor.tabSize': return 'tabSize';
+            // @monaco-uplift: uncomment this line once 'editor.indentSize' preference is available
+            //  case 'editor.indentSize': return 'tabSize';
```
msujew, line 138: Minor: Given that the previous question resulted in `indentSize`, it should be here the same as well.

```suggestion
            //  case 'editor.indentSize': return 'indentSize';
```

packages/plugin-ext/src/plugin/text-editor.ts
```diff
@@ -334,6 +336,48 @@ export class TextEditorOptionsExt implements theia.TextEditorOptions {
...
+            return; // ignore invalid values
+        }
+
+        if (typeof indentSize === 'number') {
+            if (this._indentSize === indentSize) {
+                return;
+            }
+            this._indentSize = indentSize;
+        } else if (val === 'tabSize') {
+            this._indentSize = this._tabSize!;
```
msujew, line 355: Suggestion: I believe this might yield incorrect values in case the `tabSize` is changed after `indentSize` has been set:

```ts
ext.tabSize = 2;
ext.indentSize = 'tabSize';
ext.tabSize = 4;
console.log(ext.indentSize) // prints '2'
```

It should store the `'tabSize'` value and return the current `tabSize` value from the `indentSize` getter.

packages/monaco/src/browser/monaco-text-model-service.ts
rschnekenbu: Fixed with new commit.

packages/plugin-ext/src/plugin/text-editor.ts
rschnekenbu: Indeed, this should be a better long term solution. Fixed in new version

rschnekenbu:
@msujew, hanks again for your review! 
All your comments were addressed. I also rebased on latest master.

msujew (approved):
Thanks, looks good to me 👍

rschnekenbu:
@msujew, can you merge it please? I can then create the PR for #13065, as this is the last remaining task for the vscode compatibility task.

PR #13115, opened 2023-11-28
fix(notebooks): fix race condition in view registration (#13106)

### LINKED ISSUE #13106: Notebooks are not rendered correctly when opened for the first time in a page load
### Bug Description:

Jupyter notebooks are not rendered correctly when __first__ opened in a page load. A completely empty document is shown instead. Closing and reopening the same document works as expected, and so does opening another notebook entirely. After a refresh, the issue reoccurs *unless* there is an open notebook that was recovered from the previous page.
### Steps to Reproduce:

This can be reproduced on the Theia browser example:

-  Install the necessary vscode plugins:
	- `ms-python.python` (https://open-vsx.org/extension/ms-python/python)
	- `ms-toolsai.jupyter` (https://open-vsx.org/extension/ms-toolsai/jupyter)
	- vscode builtins (https://open-vsx.org/extension/eclipse-theia/builtin-extension-pack)

- Create a new file with extension `.ipynb`
- Open the file from the explorer
- Observe the totally empty editor
image

### Additional Information

I believe the problem here is that this promise never resolves:
https://github.com/eclipse-theia/theia/blob/59613f98e6161fe18e7a35160297c25d1bdf0173/packages/notebook/src/browser/notebook-editor-widget.tsx#L111

There is some logic to handle providers being resolved after they're needed:
https://github.com/eclipse-theia/theia/blob/59613f98e6161fe18e7a35160297c25d1bdf0173/packages/notebook/src/browser/service/notebook-service.ts#L134-L155

But it seems the callback `onDidRegisterNotebookSerializer` is registered *after* the first event is fired here:

https://github.com/eclipse-theia/theia/blob/59613f98e6161fe18e7a35160297c25d1bdf0173/packages/notebook/src/browser/service/notebook-service.ts#L90-L102

In fact, after checking in a production build, we see the `waitForData` promise is rejected after the 20s timer runs out:

image

- Theia Version: `1.43.1`
- Operating system: `Debian 6.1.55-1 (2023-09-29) x86_64 GNU/Linux`
- Browser: reproduces on recent Chromium, Firefox, and electron as well.

bvenreply:
#### What it does

Fixes #13106

Fixes a race condition that occurred when event handlers for `onDidRegisterNotebookSerializer` were registered *after* said event was fired. This prevented the correct rendering of a notebook the first time one was opened during the lifetime of a page.

~To solve this, we're storing the events that are fired before any event handler is registered and replay them for the first such handler __only__.~ *Review comments highlighted this was not necessary and that there was a simpler solution*

#### How to test

-  Install the necessary vscode plugins:
	- `ms-python.python` (https://open-vsx.org/extension/ms-python/python)
	- `ms-toolsai.jupyter` (https://open-vsx.org/extension/ms-toolsai/jupyter)
	- vscode builtins (https://open-vsx.org/extension/eclipse-theia/builtin-extension-pack)

- Create a new file with extension `.ipynb`
- Open the file from the explorer
- Observe the notebook UI is shown immediately, as opposed to an empty editor before the patch.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/service/notebook-service.ts
+++ b/packages/notebook/src/browser/service/notebook-service.ts
@@ -135,22 +135,31 @@ export class NotebookService implements Disposable {
         if (this.notebookProviders.has(type)) {
             return this.notebookProviders.get(type);
         }
-        await Promise.all(this.willUseNotebookSerializerEmitter.fire(type));
         const deferred = new Deferred<NotebookProviderInfo | undefined>();
         // 20 seconds of timeout
         const timeoutDuration = 20_000;
-        const disposable = this.onDidRegisterNotebookSerializer(viewType => {
+
+        // Must declare these variables where they can be captured by the closure
+        let disposable: Disposable;
+        // eslint-disable-next-line
+        let timeout: ReturnType<typeof setTimeout>;
+
+        // eslint-disable-next-line
+        disposable = this.onDidRegisterNotebookSerializer(viewType => {
             if (viewType === type) {
                 clearTimeout(timeout);
                 disposable.dispose();
                 deferred.resolve(this.notebookProviders.get(type));
             }
         });
-        const timeout = setTimeout(() => {
+        timeout = setTimeout(() => {
             clearTimeout(timeout);
             disposable.dispose();
-            deferred.reject();
+            deferred.reject(new Error(`Timed out while waiting for notebook serializer for type ${type} to be registered`));
         }, timeoutDuration);
+
+        await Promise.all(this.willUseNotebookSerializerEmitter.fire(type));
+
         return deferred.promise;
     }
 


msujew (commented):
This fix looks a bit overengineered to me. The only thing that is actually incorrect is the order in which we wait for things in `waitForNotebookProvider`. We can fix the issue by just reordering what we're doing in there:

```ts
protected async waitForNotebookProvider(type: string): Promise<NotebookProviderInfo | undefined> {
    if (this.notebookProviders.has(type)) {
        return this.notebookProviders.get(type);
    }

    const deferred = new Deferred<NotebookProviderInfo | undefined>();
    // 20 seconds of timeout
    const timeoutDuration = 20_000;

    const disposable = this.onDidRegisterNotebookSerializer(viewType => {
        if (viewType === type) {
            clearTimeout(timeout);
            disposable.dispose();
            deferred.resolve(this.notebookProviders.get(type));
        }
    });
    const timeout = setTimeout(() => {
        clearTimeout(timeout);
        disposable.dispose();
        deferred.reject(new Error(`Timed out while waiting for notebook serializer for type ${type} to be registered`));
    }, timeoutDuration);

    await Promise.all(this.willUseNotebookSerializerEmitter.fire(type));
    return deferred.promise;
}
```

The main issue seems to be that the provider is registered during the `await Promise.all(this.willUseNotebookSerializerEmitter.fire(type));` call. We basically have a blind spot there. If we register the callback before that, we don't need the replay mechanism.

bvenreply:
@msujew I can give that a try. Though those closure capture still need to be fixed.

bvenreply:
@msujew You were right, that approach also works and it's a lot cleaner. Thank you. I force-pushed a new commit.

msujew (approved):
Alright, looks good to me 👍

PR #13116, opened 2023-11-29
fix: clear error message when group is undefined

huzhixiao:
#### What it does

clear error message when group is undefined
image

#### How to test

import @theia/toolbar on pakcage.json
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/toolbar/src/browser/toolbar.tsx
+++ b/packages/toolbar/src/browser/toolbar.tsx
@@ -143,7 +143,7 @@ export class ToolbarImpl extends TabBarToolbar {
 
     protected renderGroupsInColumn(groups: ToolbarItem[][], alignment: ToolbarAlignment): React.ReactNode[] {
         const nodes: React.ReactNode[] = [];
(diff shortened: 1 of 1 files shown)


packages/toolbar/src/browser/toolbar.tsx
```diff
@@ -143,7 +143,7 @@ export class ToolbarImpl extends TabBarToolbar {
 
     protected renderGroupsInColumn(groups: ToolbarItem[][], alignment: ToolbarAlignment): React.ReactNode[] {
         const nodes: React.ReactNode[] = [];
-        groups.forEach((group, groupIndex) => {
+        groups?.forEach((group, groupIndex) => {
```
msujew, line 146: Suggestion: I don't think that this is the correct fix. Looking at the method definition, we shouldn't even be allowed to pass `undefined` in here. The caller site of this method should take care of not passing in `undefined`.

msujew (commented):
Hey @huzhixiao, thank you for your contribution.

Issue #13062, opened 2023-11-06
[vscode] API gap: Theia master vs. VS Code 1.83.0 


rschnekenbu:
This bug tracks the differences between current master Theia VS Code API and VS Code 1.83.0:

Public APIs
- [x] #12943
- [x] #13058 
- [x] #13059
- [x] #13060 
- [x] #13061 

Proposed API evolutions:
- [x] #13064

Application management
- [x] #13065 

Repo
- [x] #13066

rschnekenbu:
All sub-tasks done, closing this issue. Thanks for all contributions!

PR #13118, opened 2023-11-29
Bump API version to 1.83.1

rschnekenbu:
#### What it does

Increase compatibility version to 1.83.0 as implementation tasks from task #13062  are closed.
Fixes https://github.com/eclipse-theia/theia/issues/13065

#### How to test
Make sure the system starts and built-ins work as expected as described in https://github.com/eclipse-theia/vscode-builtin-extensions/wiki/Produce-and-Publish-the-set-of-vscode-built-in-extensions#local-testing

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-package/src/api.ts
+++ b/dev-packages/application-package/src/api.ts
@@ -18,4 +18,4 @@
  * The default supported API version the framework supports.
  * The version should be in the format `x.y.z`.
  */
-export const DEFAULT_SUPPORTED_API_VERSION = '1.82.0';
+export const DEFAULT_SUPPORTED_API_VERSION = '1.83.1';


rschnekenbu:
I tested with the 1.83.0 versions of the builtins-extension. 
Testing done:
- I did not notice any additional issue or error log. 
- editing of Typescript / json files was working, 
- the git views and actions I tested were also working.

tsmaeder (approved):

tsmaeder (commented):
Shouldn't this be 1.83.1?

tsmaeder:
https://eclipse-theia.github.io/vscode-theia-comparator/status.html mentions 1.83.1

rschnekenbu:
I usually think that we are compatible with a major/minor version for an API, and the bug fixes versions are not supposed to interfere with any API compatibility. 
In our case, the version is compared when trying to install new extensions on open-vsix. The extensions provided on vsix may be compatible with 1.83.0 at the time they were released. If we set the API to 1.83.1, we loose these extensions in the view by default, even if they are supposed to be compatible. And I would not expect developers to update their extension compat version for any new patch release of vscode.
For the report you mention, we can basically set what we want in the parameters when generated, 1.83.0 or 1.83.1. I would not expect any differences in the generated report. Same as before, the API was compatible with 1.82.0, where the report indicates 1.82.3

tsmaeder:
> If we set the API to 1.83.1, we loose these extensions in the view by default, 

That's not how it works, IMO: any extensions that requires 1.83.0 will work with 1.83.1, but not the other way around. In particular, the built-ins for 1.83.1 (which is the release 1.83.x version) will not work with 1.83.0 per default.

see https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/11676 We've got IP-approval for the 1.83.1 version. IMO, there is not api change between 1.83.0 and 1.83.1.

tsmaeder:
Oh, wow! If I decided to release the 1.85.0 version of the built-ins, it would just pick that version, even though our api version is 1.83.1? That sounds like a bug! How does that work with non-theia clients of open-vsx? Anyway, the official VS Code version for the 1.83 release is 1.83.1: https://code.visualstudio.com/updates/v1_83 I still think it's the correct version to set. It would still work if I release the 1.83.1 built-ins, right?

rschnekenbu:
I suspect the code work as you say. I may have a check when I will have more time to do so. 
Let's go with the 1.83.1 version then!

msujew:
> Currently, osvx api filter compares the ranges of version for a given extension and looks for the one(s) who is/are greater than the supported API version.

No, it doesn't do that. It's just an unintuitive argument ordering. This line:

https://github.com/eclipse-theia/theia/blob/f85528844c3108f1b4998a3153cab23bb6648c6f/dev-packages/ovsx-client/src/ovsx-api-filter.ts#L48

Ensures that our version of the supported VSCode API is greater or equal the version that is declared in the extension's package.json engine entry. We actually perform a `lte` comparison on these two values:

https://github.com/eclipse-theia/theia/blob/f85528844c3108f1b4998a3153cab23bb6648c6f/dev-packages/ovsx-client/src/ovsx-api-filter.ts#L85

rschnekenbu:
Thanks for clarification, @msujew! 
That may need a small update of the naming and documentation then, as the doc says also "greater or Equal to" 
https://github.com/eclipse-theia/theia/blob/f85528844c3108f1b4998a3153cab23bb6648c6f/dev-packages/ovsx-client/src/ovsx-api-filter.ts#L77
I would be happy to handle this update if useful.

msujew:
@rschnekenbu Now I'm confused as well. That looks pretty weird.

rschnekenbu:
Issue created: #13123. Can one of you assign it to me?

tsmaeder (approved):

PR #13134, opened 2023-12-01
Don't await plugin deployment in backend process

tortmayr:
#### What it does
- Don't await `initialize` method in `PluginDeployerContribution` sync to  continue with backend loading while plugins are deployed
- Fix wrong performance measurements of `resolvePlugins` & `deployPlugins` &  in `PluginDeployerImpl`
- Improve performance logging in `HostedPluginSupport` by only logging relevant measurements of `sync`/`load` and `start` plugins  (i.e. if the plugin count is 0 just stop the measurement but don`t log)
- Update `I18nPreloadContribution` to ensure that we only load the localizations from the back end if the current locale is not equal to the default locale. This fixes an issue where we unnecessarily queried the localization server in the preload step if the locale is explicitly set to "en". (This happens for instance if you change the display language and then switch back to english).

#### How to test
Everything should work as before. In the example apps the deploy plugin step does not take very long because most of the
default plugins are rather minimal and can deployed very fast.
To see a the real benefit you could add a timeout to the `PluginDeployerImpl` to simulate a longer plugin deploy process:
```ts
    protected async doStart(): Promise<void> {
            await new Promise(resolve => setTimeout(resolve, 3500));
            ...
    }
  ```
  
  On the current master the backend process will block and not start the frontend loading until the `pluginDeployer` is finished.
  With this change the fronted is started and loaded right away. Once the plugins are deployed the frontend will be notified and
  starts the plugins.
  

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/preload/i18n-preload-contribution.ts
+++ b/packages/core/src/browser/preload/i18n-preload-contribution.ts
@@ -33,7 +33,7 @@ export class I18nPreloadContribution implements PreloadContribution {
                 locale: defaultLocale
             });
         }
-        if (nls.locale) {
+        if (nls.locale && nls.locale !== nls.defaultLocale) {
             const localization = await this.localizationServer.loadLocalization(nls.locale);
             if (localization.languagePack) {
                 nls.localization = localization;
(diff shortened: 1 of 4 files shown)


msujew (approved):
Looks good to me 👍

Issue #12959, opened 2023-09-28
When backend-to-backend plugins are supported, we'll need the plugin-host to be launched on backend startup


jcortell68:
### Feature Description:
Today, a plugin-host is forked only when a frontend attaches to the backend. @martin-fleck-at confirmed this recently. This is fine given that today's plugins only interact with the browser. I.e., all functions in the VS Code extension API are serviced by Theia extension code running in the frontend. Thus there is no point in loading plugins if there is no frontend.

Issue #12946 asks for backend-to-backend plugin support. Once that's available, we'll want to take things a step further. There might be backend functionality in an app that we want to run in a headless manner--either via a command-line interface, or via a server/REST model. Any backend-to-backend plugins need to be able to participate in that functionality, which means plugins need to be loaded and activated in that headless Theia environment.

msujew:
I'm sceptical how this is supposed to work and the general usefulness of the feature compared to it's potential complexity. Generally, we have three classes of vscode plugins:

1. Written for VSCode, but work in Theia as well (virtually everything that's on the Open VSX marketplace).
2. Written for Theia but using the VSCode API; they work just as well in VSCode (the common approach taken by Theia product developers trying to add language support or Debuggers to their app).
3. Written for Theia, using product specific APIs.

While 1. and 2. don't get any benefit from this feature, since they were written against VSCode's API with VSCode design and behavior in mind (which also doesn't spawn a plugin host on backend start), 3. could just have been written as a regular Theia extension and work just as well (if not better). Not to mention that you effectively need to mock a frontend just for the plugin host to start - which sounds pretty hacky in my opinion.

jcortell68:
@msujew We are developing a product that needs to support vendors, third parties and even savy users extending the application in product-specific ways. Furthermore, this needs to be doable from components that can be deployed and installed in the field.

msujew:
I get the business requirement/purpose behind this feature proposal - I just don't think it fits into the scope of the Theia open source repo. My main gripe with this feature is that it requires to mock a frontend on the backend which is just not architecturally sound. @tsmaeder I'd be interested in hearing your take on this, since you're always going on about architectural soundness ;)

[...]

Issue #12946, opened 2023-09-26
Need ability to add custom plugin API namespace that's directly connected to a backend implementation


jcortell68:
### Feature Description:
The following documentation describes how a Theia extension can extend the plugin API with new API namespaces.
   https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md

And it works great.  But...it's limited in that the new API can't be **directly** supported by the Theia extensions' backend. The API connects the plugin running in the plugin host node process to the Theia extension's frontend. And if the API ultimately is about driving logic in the frontend, then that works out perfectly. BTW, this configuration is the only one shown in the diagram [here](https://github.com/eclipse-theia/theia/blob/master/doc/Plugin-API.md).

But imagine the Theia extension having a backend service that it wants to expose to plugins. It should be able to define an API namespace that connects the plugin code to the backend service in the Theia extension. Today, you can't make that direct connection. You would have to go through an intermediary in the frontend. That is inefficient on multiple fronts.

Basically, a Theia extension's backend should be able to bind a `MainPluginApiProvider` and have it work in the same way it does when the frontend binds that. But `MainPluginApiProvider` is only injected by Theia frontend code (in `@theia/plugin-ext`).

Here is an example of a Theia app that contributes a working API namespace wired to the frontend, but then fails to contribute API namespace from the backend. See commit message and comments in the code. In particular, look at the diff between the head commit and the previous one.

```
git clone -b api_namespace_backend https://github.com/jcortell68/theiasandbox
cd theiasandbox
yarn
yarn start:electron
```
What are you looking for? Notice the following doesn't show up in stdout when you launch the app.
```
HelloWorldBakendMainPluginApiProvider.initialize() called
```
That absence proves that the Theia extension has no chance of contributing an API namespace implemented in its backend. You can see that the normal (wired to frontend) API namespace works fine by invoking the `Hello World` command that the extension contributes. The message string you see was obtained by exercising the custom plugin API.

tsmaeder:
@jcortell68  Please correct me if I'm paraphrasing this issue incorrectly, but you're asking for a way to extend the Theia back-end at install/run time, as opposed to the compile-time extensibility we have now, correct?
[...]

PR #13138, opened 2023-12-01
Implement "headless plugins" in a new plugin host outside of frontend connections

cdamus:
#### What it does

Implement a new kind of plugin that I call "headless plugin" that is hosted in a child Node process like the current backend plugins _except_ that this host is not started on a per-frontend-connection basis but exists independently of all such connections.
Thus it is "headless" as it will never know a Theia frontend.

This implements the functionality requested in #12946 and, because it makes no sense not to, #12959.

There are a few points to note about this implementation:

- I am not wedded to the term "headless plugin." I welcome suggestions for a better or more descriptive name, but I would like to avoid the term "backend" because we already have backend plugins that are the backend of a frontend connection, whereas headless plugins have no frontend. And the resulting similarity of names in the code will just be too confusing
- several of the existing interfaces and classes in the `plugin-ext` package are refactored to pull up common functionality to reuse for headless plugins with the aim of promoting maintainability of both plugin stacks over the long term. I have endeavoured to maintain API compatibility for Theia extenders, but please do keep a sharp eye out for compatibility issues
- ~a `@theia/headless-plugin` API declaration is provided that is essentially a greatly stripped-down variant of `@theia/plugin`, providing only capabilities that are available outside of frontend connections, and probably even less than that (see below for some possible follow-up in that area)~
- it is intended by this PR that the support for "headless plugins" be optional. A Theia application could be built without the `headless-plugin` and `headless-plugin-ext` packages and work exactly as it does today. If the Theia team would rather just fold support for headless plugins into the existing `plugin-ext` package (e.g. because that would eliminate the need for some of the refactorings done in this PR, actually simplifying the code and possibly making it more correct) then I would be happy to do that
- the layout of the source tree in the new `headless-plugin-ext` extension tries to match the organization of corresponding code in the `plugin-ext` package, to make it easier to understand the similarities and differences. However, as there is no frontend to headless plugins, of course there are no `browser` directories so some things (such as `XyzMain` API) are organized differently, usually in the `node/` folders
- refactoring in the Terminal Service Ext API merged after this PR was initially implemented required some late, and quick, adaptation that may not be ideal, such as the `MinimalTerminalServiceExt` type used by the `AbstractPluginManagerExt` class to declare only what it actually needs to do its work

#### How to test

This PR includes new examples demonstrating the use case that motivated issue #12946 that can be used for testing. These are:

- `examples/api-provider-sample/`: a new example Theia extension that demonstrates how to provide a custom API object for plugins to use. This is important because the original use case motivating the headless plugins feature is one in which those plugins exist to plug into such application-specific APIs. The API provides a "Greeting of the Day" service with custom functions to call, classes to create, and call-backs for a plugin to register for events
- `sample-plugins/sample-namespace/plugin-gotd`: an example dual headless/backend plugin that uses the greeting-of-the-day API of the other new example ~and also illustrates simple usage of the API provided by Theia, itself, to headless plugins~. It further demonstrates how Theia plugins may provide both backend and headless entry-points to run in both plugin hosts

To test the intended use case of this feature, simply copy the `plugin-gotd-v1.44.0.tgz` tarball from the sample plugin into your `plugins/` folder and start up the Browser Backend. You should see the new headless plugin host start up, load this plugin, and output from it on the console using `[GOTD]` tags to help you locate them in the output. Then hit the frontend in your browser to see that the connection-scoped backend plugin host does its usual thing, and that the GOTD plugin's backend entry-point runs in this context, printing messages tagged with `[GOTD-BE]` to distinguish them from messages printed by the headless plugin.

Then, of course, test the same behaviour in Electron as much as makes sense.

Apart from that, please go as crazy as you like in regression testing the usual plugin capabilities of Theia.

#### Follow-ups

- this PR does not yet include updates to user guides and other public documentation about the Theia plugin architecture. I will be happy to make such updates in this PR once we have agreed (if we do agree) to merge the implementation. Or I will be equally happy to raise a follow-up issue to tackle the docs
- ~some~ no attempt has been made to support ~a few~ any core Theia services in the headless plugin host that should be available generally, outside of frontend connections. ~I don't really know how to test most of that so there may be problems found by early adopters that we need to fix~
- ~also there may be other core Theia services that should be exposed in the new `@theia/headless-plugin` API that currently are not, contribution points and activation events to support in the package manifest that currently are not implemented. I should like to leave it to adopters to identify real-world needs in these areas that we can add later~
- this PR does not provide any default API (analogous to the `theia` or `vscode` API namespace) to plugins but relies on extensions and other plugins to provide APIs, as demonstrated by the new example GotD plugin and the API that it uses provided by the new example extension. It is an area of future consideration whether to provide some default API namespace and what capabilities it should offer, as use cases are discovered

#### Review checklist

#### Reminder for reviewers

----

### Updated 5 Dec 2023

This description is amended to rebrand the new kind of plugin as "headless plugin" (was "base plugin") and to adapt the test instructions for the now dual nature of the example GotD plugin, which previously supported only headless mode but now also backend (of a frontend connection) mode.

### Updated 11 Dec 2023

This description is amended to reflect the withdrawal of the default API namespace that was provided to headless plugins.


jcortell68:
> I am not wedded to the term "base plugin." 

I agree. `backend plugin` would be terribly confusing. Just putting this out there after starting to read your comments. How about `headless plugin`? Something better might come to mind, but that's what I have on my first hour after ten days away from work.

jcortell68:
> If the Theia team would rather just fold support for base plugins into the existing plugin-ext package (e.g. because that would eliminate the need for some of the refactorings done in this PR, actually simplifying the code and possibly making it more correct) then I would be happy to do that

We definitely would not want that given how many Theia front-end things `plugin-ext` currently pulls in. Someone should be able to at some point develop a completely headless Theia app that supports these new types of plugins, and with minimal bloat. That just wouldn't be possible if both plugin stacks were in the same package.

cdamus:
> We definitely would not want that given how many Theia front-end things `plugin-ext` currently pulls in. Someone should be able to at some point develop a completely headless Theia app that supports these new types of plugins, and with minimal bloat. That just wouldn't be possible if both plugin stacks were in the same package.

Currently the `@theia/base-plugin-ext` package has a dependency on `@theia/plugin-ext` because it reuses a lot of the latter's code. This is on the theory that Theia's greatest value proposition is in the support for a UI frontend connected to that backend so there's little incentive to support an absolutely headless Theia application (why not just use Node.js in that case?). The "base plugins" should work, however, in a scenario like CLI invocation of a Theia application that then doesn't need to bring up any front-end for that brief interaction.

To factor out the code from `plugin-ext` that `base-plugin-ext` requires would be feasible, perhaps introducing yet another new extension `@theia/common-plugin-ext` but per the above I'm not really convinced that the additional complexity is worth it.

jcortell68:
> so there's little incentive to support an absolutely headless Theia application (why not just use Node.js in that case?). 

That's a good point. It's unlikely someone intending to write a CLI application would come to Theia. With the introduction of this new plugins stack, I suppose it makes the scenario more likely, though. I.e., if I wanted to write a CLI application in TypeScript and support third parties extending the application using a robust plugin mechanism that firewalls the main application from the plugins...then maybe??? But it's a stretch, I'll admit. I don't think the likelihood warrants further work at this point. We can cross that bridge if and when we get there.

cdamus:
Commit b79be18442e04e35448e9308ad39ea645966dbaa updates the PR to

- change "base plugin" terminology to "headless plugin"
- support dual-entry-point backend plugins for the usual backend-of-a-frontend-connection host and also the headless plugin host process. This lets a single Theia plugin package provide both headless and backend plugins

The PR description is updated accordingly.

tsmaeder:
I don't think "headless" is a good term to describe this kind of plugin because that is not the relevant distinction: I would argue most plugins _would_ hook into some kind of UI. The distinction is scope: the back end plugins are shared among all front-ends. And "back end" is the right term here, since they are scoped to what we call the "back end" (as opposed to the Electron main process, for example). As with dependency injection, there is really no conceptual reason to not consider runtime extensibility for all parts of our system, including the Electron main process.

tsmaeder:
I've looked at the PR on a very high level, and here are a couple of suggestions.

1. Split the `plugin-ext` package into `plugin-ext`, containing the mechanism to contribute and run plugin namespaces and plugin-ext-theia, which contributes the theia namespace. There is way too much code serving different purposes in this packages and any cleanup and separation of concerns would be a boon. It would also solve the "we're getting too much stuff in the back-end plugins" problem that @jcortell68 mentioned.
2. While we're at it: why not use dependency injection from the start in the plugin host? It will facilitate sharing ext-services between different api namespaces: that's one roadblock to making vscode and theia namespaces contributed instead of built-in.
3. Let's not have a default namespace: all back-end plugin namespaces should be contributed
4. Lets not contribute any namespaces by default: we currently have no idea what users might want
5. If we _do_ decide to contribute any API in the back end, it should not be derived in any way from theia.d.ts: for one, the theia namespace is tightly bound to the vscode namespace, so we're not free to evolve it. Secondly, the requirements for for back end plugins might be very different from what the vscode API (of which theia.d.ts is basically a copy) might need. Let's not restrict ourselves.

jcortell68:
> I don't think "headless" is a good term to describe this kind of plugin because that is not the relevant distinction: I would argue most plugins _would_ hook into some kind of UI. The distinction is scope: the back end plugins are shared among all front-ends. And "back end" is the right term here, since they are scoped to what we call the "back end" (as opposed to the Electron main process, for example). As with dependency injection, there is really no conceptual reason to not consider runtime extensibility for all parts of our system, including the Electron main process.

I respectfully disagree. I think it is precisely the relevant distinction. There are currently two types of plugins in Theia (and VS Code). One runs in the backend, the other in the frontend, but _both are wired to the frontend_. They drive functionality in the UI (the "head"). The new type of plugin this PR introduces not only lives in the backend but interacts only with the backend (of Theia extensions). Thus "headless" actually seems like a very suitable term. As one of the existing types of plugins today runs in the backend, I think calling this new thing a "backend plugin" would be pretty confusing.

> since they are scoped to what we call the "back end" (as opposed to the Electron main process, for example).

The Electron main process is really two things running side-by-side in the same process: Node.js and the Chromium browser. I.e., it is both the frontend and the backend.  Thus I'm not really understanding your point here.

cdamus:
Commit 15f3886004aa6d740f80093c8b656c851ddf9bb4 removes the `@theia/headless-plugin` extension defining a default "theia" API namespace for headless plugins. It therefore also removes a lot of code supporting the implementation, initialization, and module-loading of that API.

It did require refactoring a few dependencies on concrete Ext implementations in the common plugin manager code to use instead internal interfaces that could be stubbed with no-ops for the headless plugin manager.

The GotD example plugin is updated to not attempt to use the API that no longer is available. Otherwise, everything seems to run as before (headless and VS Code plugins).

sample-plugins/sample-namespace/plugin-gotd/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "private": true,
+  "name": "plugin-gotd",
+  "version": "1.44.0",
+  "main": "extension.js",
```
jcortell68, line 5: This `main` field is extraneous, right? Relevant in a traditional plugin (VS Code externsion), but not in this case.

cdamus: Yeah, I forgot to delete that when I reworked the plugin as a dual headless/backend plugin. It would also work for a headless plugin using the `theiaHeadless` engine but that's not the case, here.

cdamus:
> I wonder if we shouldn't be using `activate/deactivate` for the headless plugin entry points instead of `start/stop`.

For whatever reason, `start` and `stop` are the function names expected by Theia plugin using the `theiaPlugin` engine. As our new headless plugins have no analogue in VS Code, it seemed appropriate to use the Theia-aligned names without some very compelling reason to diverge from them. The fact that internally in the plugin infrastructure there's a "starting" of plugins (which honestly doesn't really start anything) is not exposed in any way to plugin authors, so I don't see that this should be a point of confusion.

jcortell68:
> is not exposed in any way to plugin authors

It is exposed in the electron app's stdout. 

It's yet another unfortunate discrepancy that is likely to cause confusion (another is the "extension" vs "plugin" naming thing). I believe most plugin development for Theia apps will be of the type VS Code extensions and headless plugins. I'm told people aren't really writing `theaPluigin` types of plugins, so the `start/stop` choice there is mostly irrelevant. There is no need to try to stay consistent with a non-existent practice. I recommend we at least make the plugin mechanism support both. If we want to give precedent to `start/stop` in the implementation, that's fine. The documentation should point folks in the direction of `activate/deactivate`, though. IMO, it's more important to stay consistent with an active practice than an inactive one.

cdamus:
Commit a551b36ecc41b0ce8ac400ae6209ec0f080f97a4 adds support for the "headless" entry point in a plugin that otherwise is an ordinary VS Code plugin. And because of this the expected start/stop lifecycle function names are changed to `activate` and `deactivate` respectively. This means that

- the VS Code plugin scanner is enhanced to pick up the `theiaPlugin.headless` entrypoint in the manifest if it happens to be specified
- the headless plugin scanner is updated to adapt Theia plugins by renaming their start/stop lifecycle function names
- the GotD example plugin is changed
    - to be a VS Code plugin, using the `vscode` API in its normal backend entrypoint script, now renamed `extension.js` as is tradition for VS Code plugins
    - in the `headless.js` script to rename its exported lifecycle functions

sample-plugins/sample-namespace/plugin-gotd/package.json
cdamus: Ah, the `"main"` entrpoint is back now that this example plugin is a VS Code plugin. 😉

jcortell68:
> Commit [a551b36](https://github.com/eclipse-theia/theia/commit/a551b36ecc41b0ce8ac400ae6209ec0f080f97a4) adds support for the "headless" entry point in a plugin that otherwise is an ordinary VS Code plugin...

Thank you!

jcortell68:
@cdamus I'm still working with your initial commit, as I had to rebase your changes onto Theia 1.43.0 and that was a bit tricky. Bringing in any further changes might take a little time. So far, you haven't changed anything fundamental in the subsequent commits so I'll keep trucking as-is for now.

It appears that the headless plugin isn't activated unless I have
```
"activationEvents": ["*"],
```

which is unexpected. My understanding is that there is no support yet for activation events, so I assume the plugin should be loaded unconditionally at this point.

jcortell68:
Or maybe it's not that there is no support for activation events and that there just aren't any activation events to trigger off of. So you have to basically request to be loaded unconditionally with "*". If that's the case, then we'd for sure want to use a unique package.json element specifically for headless plugin--maybe `headlessActivationEvents`

cdamus:
> Or maybe it's not that there is no support for activation events and that there just aren't any activation events to trigger off of. So you have to basically request to be loaded unconditionally with "*". If that's the case, then we'd for sure want to use a unique package.json element specifically for headless plugin--maybe `headlessActivationEvents`

I had thought about this. There are no activation events specified by the framework, but I expect that applications will naturally find their own suite of events according to what their extension points are that headless plugins will supply. So I thought that applications could read the `"activationsEvents"` from the plugin manifests and manage activation themselves, which would mean that the `"*"` event still has meaning and in its absence a plugin should not be activated by the framework but that responsibility would be delegated to the application.

Does that make sense?

jcortell68:
> Does that make sense?
Yes. It does.

> BTW, is there a problem with adding application-specific activation events that VS Code doesn't understand to the `"activationEvents"` object? I'd rather not reduplicate everything in the package manifest schema for headless plugins.
That's all fine and well when  plugins specify actual activation events. We're sort of counting on their being no name collision on the activation events between the two types of plugins. Not ideal, but that will probably be fine. The real issue comes when one type of plugin wants to be activated by default ("*"). Under the current design/implementation, a "*" dictates the behavior for both types of plugins, which we really don't want. I might have a plugin package with a very lightweight headless plugin but a beast of a VS Code extension. If I need the headless plugin to be loaded by default, that means an unnecessary heavy penalty on the frontend.

One approach is to have a distinct string to represent "*" for headless. But that starts looking ugly. With the approach I recommended (`headlessActivationEvents`), we kill both of those birds with one stone. And I get it...that's not winning any beauty pageants either, but I think it is more future proof and arguably more intuitive.

packages/plugin-ext/src/common/plugin-ext-api-contribution.ts
```diff
@@ -19,33 +19,66 @@ import { interfaces } from '@theia/core/shared/inversify';
...
-     * Provide API description
+     * Provide API description.
      */
     provideApi(): ExtPluginApi;
 }
 
 /**
- * Plugin API extension description.
- * This interface describes scripts for both plugin runtimes: frontend(WebWorker) and backend(NodeJs)
+ * Provider for backend extension API description.
```
jcortell68, line 32: Hm. I must be confused. Historically, we've had just `ExtPluginApiProvider` which was used for both traditional plugins and web extensions. Now we have

- ExtPluginApiProvider
- ExtPluginBackendApiProvider
- ExtPluginFrontendApiProvider

It seems the third of these is unused and is maybe cruft from the early stages of implementing headless-plugin support. If that's right, then I also think `ExtPluginBackendApiProvider` should probably be named `ExtHeadlessPluginApiProvider`

jcortell68: Nevermind. I think I might be confused. But maybe some additional documentation of these varying providers would help clarify the situation.

examples/api-provider-sample/src/node/gotd-backend-module.ts
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+import { ContainerModule } from '@theia/core/shared/inversify';
+import { ExtPluginApiProvider } from '@theia/plugin-ext';
+import { ExtPluginGotdApiProvider } from './ext-plugin-gotd-api-provider';
+import { MainPluginApiProvider } from '@theia/plugin-ext/lib/common/plugin-ext-api-contribution';
+import { GotdMainPluginApiProvider } from './gotd-main-plugin-provider';
+
+export default new ContainerModule(bind => {
+    bind(Symbol.for(ExtPluginApiProvider)).to(ExtPluginGotdApiProvider).inSingletonScope();
```
jcortell68, line 23: How do we specify here that this provider should be providing an API object only in the headless plugin-host and not the traditional plugin-host? In other words, if this `api-provider-sample` Theia Extension wanted to contribute namespace ABC to traditional plugins (VS Code Extensions) and namespace DEF to headless plugins, is that not possible?

jcortell68: Disregard this comment. As I dove further into having a **single** Theia Extension provide different API namespaces for traditional plugins vs headless one, I see that it's not really feasible. So, I think the moral of the story here is: in that situation, the Theia application developer should have two Theia Extensions--one that provides the traditional API, one that provides the headless API.

packages/headless-plugin-ext/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@theia/headless-plugin-ext",
+  "version": "1.44.0",
+  "description": "Theia - Backend-only Plugin Extension",
+  "main": "lib/common/index.js",
```
jcortell68, line 5: There is no `index` module or ambient module in `src/common`. This looks like maybe a stowaway from `@theia/plugin-ext`

cdamus:
Commit f8fec654d32ae0e8335351d8bbf58941e368fd27 proposes a solution for applications to define and use their own suite of activation events to trigger plugin activation according to their needs. This includes:

- adding a `"headless"` property to the plugin `package.json` that — so far — has only an optional `"activationEvents"` property of type `string[]`
- taking contributions of string arrays of supported activation events names on a `SupportedHeadlessActivationEvents` Inversify service-identifier, which lets the application define the headless activation events that it needs the plugin manager in the isolated host process to recognize
- self-contribution of the `"*"` and `"onStartupFinished"` activation events that are applicable in general even to headless plugins
- adaptation of the GotD example plugin to declare headless activation events in the new `"headless"` property

For now, only simple activation event strings are supported. Parameterized events (with `"<event-name>:<data>"`) structure would require delegation to some kind of activation handlers contributed by the application. That can be a future undertaking if and when it is needed.

examples/api-provider-sample/src/node/gotd-backend-module.ts
jcortell68: I’ve experimented with this PR and I believe I've confirmed my suspicions. There is no scoping of API providers to a specific plugin-host (normal vs headless). A contributed API provider is instantiated and exercised in both types of plugin-hosts. This seems problematic to me. It means headless plugin can do an import of a normal (non-headless) custom API module and the corresponding API provider running in the headless plugin-host will give it to him--even if it makes no sense. The headless plugin will end up with an API object whose calls will never complete because the Ext object has a proxy to something that has no Main on the other end. 


But “wait”, you say. “Why is a headless plugin trying to interact with the UI? That’s not what headless plugins are for.” Absolutely. The current implementation in this PR is not responsible for that nonsense. It is, however, responsible for allowing the nonsense to get too far. The import of the normal API object from a headless plugin should fail but it doesn’t.


There is a case where this actually would be desirable, though. Technically speaking, an API object need not interact with anything on the mainland. An app developer could theoretically provide an API object whose implementation can be carried out entirely in the plugin-host–i.e., one that does not make use of RPC. E.g., a Math API object, as silly as that would be. In that case, there would be an argument for an API provider to allow itself to be available for both types of plugins: normal and headless one. Should we support that? If so…sure; that’s fine. But we still need the scoping to prevent it in the 99% of situations, where it’s not going to make sense.

packages/headless-plugin-ext/package.json
cdamus: This is fixed in commit f100932d6c68cd309a3ba08adcd82531012bb72c.

cdamus:
Commit f100932d6c68cd309a3ba08adcd82531012bb72c adds a distinction of headless initialization script and backend initialization script in the Ext API providers. This lets an API provider decide whether to publish its API to backend plugins, headless plugins, frontend plugins, or any combination of the three. The Greeting-of-the-Day API provider example is updated to publish to both headless and (VS Code) backend plugins.

examples/api-provider-sample/src/node/gotd-backend-module.ts
cdamus: The distinction of which plugin host an API provider may publish its API should now be possible with commit f100932d6c68cd309a3ba08adcd82531012bb72c.

jcortell68:
@cdamus I ran into something that isn't working for me and it's something that your example custom API implementation doesn't try to do. Could you please try to do an InversifyJS injection in `GreetingMainImpl`. Try to inject anything that has been bound into the application backend by another package (maybe even the same package, but I've not tried that, as it's not what I need in my app). 

I've tried this with implicit injection as well as explicitly using the `container` object that is passed to `GreetingMainImpl`'s constructor. Neither works for me, with the latter throwing the exception
```
Error: No matching bindings found for serviceIdentifier: XYZ
```

examples/api-provider-sample/README.md
```diff
@@ -27,7 +27,7 @@ The source code is laid out in the `src/` tree as follows:
 
 - `gotd.d.ts` — the TypeScript definition of the `gotd` API object that plugins import to interact with the "Greeting of the Day" service
 - `plugin/` — the API initialization script and the implementation of the API objects (`GreetingExt` and similar interfaces).
-  All code in this directory runs exclusively in the separate plugin-host Node process, isolated from the main Theia process.
+  All code in this directory runs exclusively in the separate plugin-host Node process, isolated from the main Theia process, together with either headless plugins or the backend of VS Code plugins.
```
jcortell68, line 30: `in the separate plugin-host Node process`

I wonder if the use of `the` will mislead the reader into thinking there is only one plugin-host Node process. Using `a` may equally mislead the reader in thinking that the code will run in a Node process dedicated to that code. Of course, neither is true.

- there is a single plugin-host process for all headless plugins running in an app
- there is a plugin-host process for each frontend that attaches to the app's backend. Each one houses all traditional VS Code plugins, servicing their respective frontend only.

How to succinctly express that here...oof. I leave that up to you. 😅

packages/headless-plugin-ext/src/common/headless-plugin-ext-api-contribution.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { PluginManager } from '@theia/plugin-ext';
+import { RPCProtocol } from '@theia/plugin-ext/lib/common/rpc-protocol';
+
+export * from '@theia/plugin-ext';
+
+declare module '@theia/plugin-ext' {
+/**
+ * Plugin API extension description.
+ * This interface describes scripts for all three plugin runtimes: frontend(WebWorker), backend(NodeJs), and headless(NodeJs)
```
jcortell68, line 25: While this comment is true, the reader is likely going to scratch their head. It's not obvious that what's described here is the result of [interface merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces). The other properties of the interface come from a different module, specifically pulled in via line 20. I would add a comment here to help the reader make that connection.

jcortell68:
Just in time for my holiday vacation, I've completed my test drive of this PR, using the Theia application I'm working on (not using the example API provider and plugin provided in this PR). 

I recently moved up to the latest commit (5) so as to verify the latest state of the PR. Everything is working as per my expectations. I have added custom headless plugin API into my app and I have a headless plugin successfully exercising that API. The servicing of that API (the Main object) is interacting well with my application's backend. I've also been able to make one plugin package contribute both types of plugins: normal and headless. The headless-scoped activationEvents is working as expected, and the plugin's entry points are as expected for my scenario (activate/deactivate). 

So, this is good to go from my perspective.

BTW, I think the various possibilities for how plugins can be declared in `package.json` will need some careful documentation, as it's not really straightforward, especially now with the headless type of plugin in the mix. Various elements of package.json come into play in non-intuitive ways: `main`, `engine` and `theiaPlugin`. Web Extensions only adds more possibilities. In all, there's eleven possibilities for what a plugin package can contribute. Spelling out recommended values for `main`, `engine` and `theiaPlugin` in all the scenarios would go a long way in helping plugin developers navigate these murky waters. 

vscode plugin
headless plugin
web extension

vscode plugin + headless plugin
vscode plugin + web extension
headless plugin + web extension

vscode plugin + headless plugin + web extension

Plus four with `theia plugin` replacing `vscode plugin`. Even though theiaPlugin's aren't really a thing at this point, the code does support them, so they should be covered.

cdamus:
> I've looked at the PR on a very high level, and here are a couple of suggestions.

Thanks, Thomas, for this initial feedback.

> 1. Split the `plugin-ext` package into `plugin-ext`, containing the mechanism to contribute and run plugin namespaces and plugin-ext-theia, which contributes the theia namespace. There is way too much code serving different purposes in this packages and any cleanup and separation of concerns would be a boon. It would also solve the "we're getting too much stuff in the back-end plugins" problem that @jcortell68 mentioned.

Before I start on this, it will be important to settle the naming question (see below). Then I can factor out the `theia` API namespace from `plugin-ext` into `plugin-ext-theia` and rename the current `headless-plugin-ext` package as `plugin-ext-<agreed-name>` only once.

> 2. While we're at it: why not use dependency injection from the start in the plugin host? It will facilitate sharing ext-services between different api namespaces: that's one roadblock to making vscode and theia namespaces contributed instead of built-in.

How shall the DI container in the plugin host process be configured? Should the Theia app generator generate a new script to load `ContainerModule`s referenced by a new `package.json` property (e.g. `pluginHostExtensions`) in similar fashion to the `Container` setup currently generated from the `theiaExtensions` property? And then presumably each object in the `pluginHostExtensions` array would have properties `backend`, `frontend`, and `headless` (to be renamed) to target injections to each of these plugin hosts.

> 3. Let's not have a default namespace: all back-end plugin namespaces should be contributed
> 4. Lets not contribute any namespaces by default: we currently have no idea what users might want

It isn't clear to me what is the distinction between these items. At any rate, I think they are addressed by the removal of the `theiaHeadless` API namespace in commit 99b5002833b8b0c9c10031074bf9e30237970338.

> 5. If we _do_ decide to contribute any API in the back end, it should not be derived in any way from theia.d.ts: for one, the theia namespace is tightly bound to the vscode namespace, so we're not free to evolve it. Secondly, the requirements for for back end plugins might be very different from what the vscode API (of which theia.d.ts is basically a copy) might need. Let's not restrict ourselves.

This is effectively mooted by commit 99b5002833b8b0c9c10031074bf9e30237970338 that retracts the default API namespace for these plugins.

@tsmaeder and @jcortell68 my suggestions for an alternative to the term "headless" in "headless plugin" are, in order of my preference:

1. "shared" — as remarked already, the behaviours provided by these plugins are shared by all frontend connections as they are instantiated in one host process in the backend that exports its API objects to the world
2. "backendShared" — like the previous but makes it clear that these plugins are resident in the Theia backend, only it's more of a mouthful
3. "singleton" or "backendSingleton" — this emphasizes the single instantiation of these plugins instead of their shared-ness, which is implied by their singleton-ness

Any better ideas? Really trying to avoid just calling them "backend" plugins because we already have those and the keyword "backend" is already all over the package.jsons.

jcortell68:
> @tsmaeder and @jcortell68 my suggestions for an alternative to the term "headless" in "headless plugin" are, in order of my preference:
> 
> 1. "shared" — as remarked already, the behaviours provided by these plugins are shared by all frontend connections as they are instantiated in one host process in the backend that exports its API objects to the world
> 2. "backendShared" — like the previous but makes it clear that these plugins are resident in the Theia backend, only it's more of a mouthful
> 3. "singleton" or "backendSingleton" — this emphasizes the single instantiation of these plugins instead of their shared-ness, which is implied by their singleton-ness
> 
> Any better ideas? Really trying to avoid just calling them "backend" plugins because we already have those and the keyword "backend" is already all over the package.jsons.

I think whatever term we come up with _must_ be concise. That for me eliminates `backendShared` and `backendSingleton`. `shared` and `singleton` are too generic/ambiguous, IMO--even if they are part of the explanation as to what the new plugin type represents. Furthermore, the term we pick should not only be relevant, but it should be _obvious and intuitive_--as much as possible, anyway. The only term I've seen that comes close to that is `headless`, and I'm not just saying that because I suggested it. I'll happily vote for something else that equally meets the criteria I mentioned--concise and intuitive. Those are my two cents, anyway. I'm not a maintainer of the project, so those two cents won't go very far, but there you have them all the same. Whatever we do, let's please not call them "backend" plugins, for exactly the reason you mentioned. It will cause a lot of confusion.

cdamus:
Commit dab5684a1882188cffc43b3f5deb17ee593fb223 rebases the branch onto the latest `master` to resolve conflicts, including:

- new version numbers across the board
- intervening changes in the plugin host framework
- accounting for the new asynchronous execution of plugin deployment during the backend application start-up sequence

cdamus:
> How shall the DI container in the plugin host process be configured? Should the Theia app generator generate a new script to load `ContainerModule`s referenced by a new `package.json` property (e.g. `pluginHostExtensions`) in similar fashion to the `Container` setup currently generated from the `theiaExtensions` property? And then presumably each object in the `pluginHostExtensions` array would have properties `backend`, `frontend`, and `headless` (to be renamed) to target injections to each of these plugin hosts.

Discussed off-line, it's clear that, in fact, we should be able to accomplish this using the plugin module loading already in place. As we need to load a module to access the start/stop functions and API factory of the plugin already, this can be expanded to access an Inversify container-module to configure our host's DI container.

packages/plugin-ext/src/main/node/handlers/plugin-theia-directory-handler.ts
```diff
@@ -42,15 +42,20 @@ export class PluginTheiaDirectoryHandler implements PluginDeployerDirectoryHandl
 
     async accept(resolvedPlugin: PluginDeployerEntry): Promise<boolean> {
 
-        console.debug('PluginTheiaDirectoryHandler: accepting plugin with path', resolvedPlugin.path());
+        console.debug(`${this.constructor.name}: accepting plugin with path ${resolvedPlugin.path()}`);
```
jfaltermeier, line 45: I believe `${this.constructor.name}` might not work after a production build, but it's a debug message anyway, so this might not be too important

cdamus: Ah, interesting point. I think probably it's trying too hard anyways. I'll just go with the generic name. Thanks!

examples/api-provider-sample/src/node/ext-plugin-gotd-api-provider.ts
```diff
@@ -0,0 +1,30 @@
...
+// *****************************************************************************
+import * as path from 'path';
+import { injectable } from '@theia/core/shared/inversify';
+import { ExtPluginApi, ExtPluginApiProvider } from '@theia/plugin-ext-headless';
+
+@injectable()
+export class ExtPluginGotdApiProvider implements ExtPluginApiProvider {
+    provideApi(): ExtPluginApi {
+        // We can support both backend plugins and headless plugins
+        const universalInitPath = path.join(__dirname, '../backend/gotd-api-init');
```
jfaltermeier, line 24: I was a bit confused about this path since within this package we only have `../plugin/gotd-api-init`. 
However this init path is relative to the built application as far as I could see. Could we adjust the comment to make this more clear?

packages/plugin-ext-headless/src/hosted/node/plugin-ext-headless-hosted-module.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    bind(BackendApplicationContribution).toDynamicValue(({container}) => {
+        let hostedPluginSupport: HeadlessHostedPluginSupport | undefined;
+
+        return {
+            onStart(): MaybePromise<void> {
+                // Create a child container to isolate the Headless Plugin hosting stack
+                // from all connection-scoped frontend/backend plugin hosts and
+                // also to avoid leaking it into the global container scope
+                const basePluginsContainer = container.createChild();
```
martin-fleck-at, line 62: nit: Any reason this is called "base" plugins container?

packages/plugin-ext-headless/src/hosted/node/plugin-host-headless-rpc.ts
```diff
@@ -0,0 +1,94 @@
...
+        return {
+            envExt: this.envExt,
+            localizationExt: this.localizationExt
+        };
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    protected createAPIFactory(_extInterfaces: HeadlessExtInterfaces): undefined {
+        // As yet there is no default API namespace for backend plugins to access the Theia framework
+        return undefined;
```
martin-fleck-at, line 71: I wonder if there is anything better that we can return instead of undefined? I understand that the generic argument is also undefined but it feels a but odd, maybe wrongly so.

packages/plugin-ext-headless/src/hosted/node/plugin-host-headless-rpc.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    protected createExtInterfaces(): HeadlessExtInterfaces {
+        return {
+            envExt: this.envExt,
+            localizationExt: this.localizationExt
+        };
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
martin-fleck-at, line 68:
```suggestion
```

packages/plugin-ext-headless/src/hosted/node/plugin-host-headless-rpc.ts
```diff
@@ -0,0 +1,94 @@
...
+    protected createAPIFactory(_extInterfaces: HeadlessExtInterfaces): undefined {
+        // As yet there is no default API namespace for backend plugins to access the Theia framework
+        return undefined;
+    }
+
+    protected override getBackendPluginPath(pluginModel: PluginModel): string | undefined {
+        return pluginModel.entryPoint.headless;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
martin-fleck-at, line 78:
```suggestion
```

packages/plugin-ext-headless/src/hosted/node/plugin-host-headless.ts
```diff
@@ -0,0 +1,137 @@
...
+        //         pluginApiCache.dispose();
+        //     }
+        // });
+    }
+});
+container.bind(RPCProtocol).toConstantValue(rpc);
+
+container.bind(AbstractPluginHostRPC).toService(HeadlessPluginHostRPC);
+container.bind(HeadlessPluginHostRPC).toSelf().inSingletonScope();
+container.bind(EnvExtImpl).to(EnvNodeExtImpl).inSingletonScope();
```
martin-fleck-at, line 134: Is there any particular reason we bind the environment extension and the localization extension against an actual implementation but only use dummies for the others, see  `HeadlessPluginManagerExtImpl`? Do we need those two for some special reason as opposed to the others? Could they also be

packages/plugin-ext-headless/src/hosted/node/plugin-host-headless-rpc.ts
```diff
@@ -0,0 +1,94 @@
...
+    protected readonly localizationExt: LocalizationExtImpl;
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    constructor(@inject(RPCProtocol) _rpc: RPCProtocol) {
+        super(
+            _rpc, 'HEADLESSPLUGIN_HOST', undefined,
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            (pluginHost: PluginHost, extInterfaces: HeadlessExtInterfaces, rpc: any) => {
+                const { envExt, localizationExt } = extInterfaces;
+                return new HeadlessPluginManagerExtImpl(pluginHost,
```
martin-fleck-at, line 50: Since we already have dependency injection, I wonder whether we should really hardcode the construction here. Maybe we can introduce a factory function or something similar so it could be overridden? Or is that not possible anyway since there is not much we can actually override at this point? Just wondering.

packages/plugin-ext-headless/src/main/node/env-main.ts
```diff
@@ -0,0 +1,44 @@
...
+// *****************************************************************************
+
+import { interfaces } from '@theia/core/shared/inversify';
+import { EnvVariablesServer } from '@theia/core/lib/common/env-variables';
+import { RPCProtocol } from '@theia/plugin-ext/lib/common/rpc-protocol';
+import { EnvMain } from '@theia/plugin-ext/lib/common/plugin-api-rpc';
+import { isWindows, isOSX } from '@theia/core';
+import { OperatingSystem } from '@theia/plugin-ext/lib/plugin/types-impl';
+
+export class EnvMainImpl implements EnvMain {
```
martin-fleck-at, line 24: This seems to be a 1:1 copy from plugin-ext, could we not simply re-use their implementation or at least inherit from it?

packages/plugin-ext-headless/src/main/node/message-registry-main.ts
```diff
@@ -0,0 +1,49 @@
...
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { interfaces } from '@theia/core/shared/inversify';
+import { MessageService } from '@theia/core/lib/common/message-service';
+import { MessageRegistryMain, MainMessageType, MainMessageOptions, MainMessageItem } from '@theia/plugin-ext/lib/common/plugin-api-rpc';
+
+export class MessageRegistryMainImpl implements MessageRegistryMain {
```
martin-fleck-at, line 21: This also seems to be a 1:1 copy from the plugin-ext? Could we not simply extend their implementation? I know they are in browser/node respectively but we could move them to common. I see that they do have some dependencies to specific types that we do not use for headless so I'll leave it up to you whether you want to have another look at it.

packages/plugin-ext-headless/src/main/node/plugin-ext-headless-main-module.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    //
+    // Main API dependencies
+    //
+
+    bind(MessageService).toSelf().inSingletonScope();
+    bind(MessageClient).toSelf().inSingletonScope(); // Just logs to console
+    bind(ProgressService).toSelf().inSingletonScope();
+    bind(ProgressClient).toDynamicValue(ctx => ({
+            async showProgress(_progressId: string, message: ProgressMessage, cancellationToken: CancellationToken): Promise<string | undefined> {
```
martin-fleck-at, line 41: Should we create a class for this? We do not actually seem to need the dynamic context anyway and it would make it easier to rebind the progress-client.

packages/plugin-ext/src/hosted/node/hosted-plugin-process.ts
```diff
@@ -55,7 +55,7 @@ export class HostedPluginProcess implements ServerPluginRunner {
     @named(PluginHostEnvironmentVariable)
     protected readonly pluginHostEnvironmentVariables: ContributionProvider<PluginHostEnvironmentVariable>;
 
-    @inject(MessageService)
+    @inject(MessageService) @optional()
```
martin-fleck-at, line 58: I wonder whether we should not simply bind a `MessageService` in the headless plugin case so that messages are properly logged and not completely lost. Simply writing out to the console like the default `MessageClient` and `MessageService` should be enough, I think.

packages/plugin-ext/src/hosted/node/plugin-host.ts
```diff
@@ -104,5 +120,35 @@ process.on('message', async (message: string) => {
...
 
-const pluginHostRPC = new PluginHostRPC(rpc);
-pluginHostRPC.initialize();
+const container = new Container();
+container.bind(PluginContainerModuleLoader).toConstantValue((module: ContainerModule) => {
+    container.load(module);
+    const internalModule = module as InternalPluginContainerModule;
+    const pluginApiCache = internalModule.initializeApi?.(container);
+    if (pluginApiCache) {
+        // TODO
```
martin-fleck-at, line 129: See my other comment on the TODO, what do we need to do? What is the impact? Should we track that in a separate issue?

packages/plugin-ext/src/plugin/preference-registry.ts
```diff
@@ -86,8 +88,8 @@ export class PreferenceRegistryExtImpl implements PreferenceRegistryExt {
     readonly onDidChangeConfiguration: Event<theia.ConfigurationChangeEvent> = this._onDidChangeConfiguration.event;
 
     constructor(
-        rpc: RPCProtocol,
-        private readonly workspace: WorkspaceExtImpl
+        @inject(RPCProtocol) rpc: RPCProtocol,
+        @inject(WorkspaceExtImpl) private readonly workspace: WorkspaceExtImpl
```
martin-fleck-at, line 92: nit: Should that be protected for convenience?

martin-fleck-at (changes requested):
Hi Christian, thank you very, very much for that change! I had a look at the code and will do some runtime testing a little bit later. From a first glance everything looks good and I believe this feature will add a lot of value to Theia, especially with the introduction of Inversify on the extension host side. My comments are mostly regarding understanding and extensibility. As soon as everything is reviewed and agreed upon, we should consider whether the documentation properly explains the possibilities and pit falls. In any case, great work!

examples/api-provider-sample/src/node/ext-plugin-gotd-api-provider.ts
cdamus: Good point!

examples/api-provider-sample/src/plugin/gotd-api-init.ts
```diff
@@ -0,0 +1,98 @@
...
+import { inject, injectable, postConstruct } from '@theia/core/shared/inversify';
+import { RPCProtocol } from '@theia/plugin-ext/lib/common/rpc-protocol';
+import { Plugin } from '@theia/plugin-ext/lib/common/plugin-api-rpc';
+import type * as gotd from '../gotd';
+import { GreetingKind, GreetingExt, MAIN_RPC_CONTEXT } from '../common/plugin-api-rpc';
+import { GreetingExtImpl } from './greeting-ext-impl';
+import { Disposable, DisposableCollection } from '@theia/core';
+import { ApiFactory, PluginContainerModule } from '@theia/plugin-ext/lib/plugin/node/plugin-container-module';
+
+// Creates the custom API object and dishes it out when a plugin's module
```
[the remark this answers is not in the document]
cdamus: I'll do better than that and not use such informal idioms. This was taken from the OP's attachment to the issue, and I don't know if it that was derived from some earlier source.

examples/api-provider-sample/src/common/plugin-api-rpc.ts
```diff
@@ -0,0 +1,70 @@
+// *****************************************************************************
+// Copyright (C) 2023 EclipseSource and others.
```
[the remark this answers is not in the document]
cdamus: Yes. This began last year but has not been published, yet. Will update all!

packages/plugin-ext-headless/src/hosted/node/scanners/scanner-theia-headless.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        // Same as in VS Code
+        lifecycle.startMethod = 'activate';
+        lifecycle.stopMethod = 'deactivate';
+
+        return result;
+    }
+
+    protected adaptContributesForHeadless(plugin: DeployedPlugin): DeployedPlugin['contributes'] {
+        // We don't yet support and contribution points in headless plugins
```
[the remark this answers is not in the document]
cdamus: Yes, this and the activation events will be candidates for follow-up issues.

packages/plugin-ext-headless/src/hosted/node/scanners/scanner-theia-headless.ts
```diff
@@ -0,0 +1,69 @@
...
+    adaptForHeadless(plugin: DeployedPlugin): DeployedPlugin {
+        return {
+            type: plugin.type,
+            metadata: this.adaptMetadataForHeadless(plugin),
+            contributes: this.adaptContributesForHeadless(plugin)
+        };
+    }
+
+    protected adaptMetadataForHeadless(plugin: DeployedPlugin): DeployedPlugin['metadata'] {
+        const result = deepClone(plugin.metadata);
```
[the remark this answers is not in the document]
cdamus: Yes, it is precisely a matter of cloning the plugin and modifying the DNA of one clone (the headless one — ick, not liking where this metaphor is going) to behave differently. I'll explain in a comment.

packages/plugin-ext-headless/src/hosted/node/plugin-ext-headless-hosted-module.ts
cdamus: Oops, missed rename from the first edition of this PR!

packages/plugin-ext-headless/src/hosted/node/plugin-host-headless-rpc.ts
cdamus: I think perhaps I don't understand the question. We have no API object that we can provide by default, like the `theia` or `vscode` object. I think `null` would better signal that there is not supposed to be an API object rather than it just not yet being available and initialized. Is that what you mean?

packages/plugin-ext-headless/src/hosted/node/plugin-host-headless.ts
cdamus: Seems you didn't finish your thought, but I can answer in part, which is that my understanding is that the LocalizationExt is used for localization of strings in the package manifest, which I thought we would need. The `EnvExtImpl` I think maybe we don't need any longer as it was only used to provide some minimal capabilities of the default API object (such as getting environment variables) but we've since removed that default API. So I shall try excising the EnvExtImpl. Not sure about the localization.

packages/plugin-ext-headless/src/main/node/env-main.ts
cdamus: I think we can do better and delete it; see my response to an earlier comment 😀

packages/plugin-ext/src/plugin/workspace.ts
```diff
@@ -63,9 +65,9 @@ export class WorkspaceExtImpl implements WorkspaceExt {
 
     private canonicalUriProviders = new Map<string, theia.CanonicalUriProvider>();
 
-    constructor(rpc: RPCProtocol,
-        private editorsAndDocuments: EditorsAndDocumentsExtImpl,
-        private messageService: MessageRegistryExt) {
+    constructor(@inject(RPCProtocol) rpc: RPCProtocol,
+        @inject(EditorsAndDocumentsExtImpl) private editorsAndDocuments: EditorsAndDocumentsExtImpl,
```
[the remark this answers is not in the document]
cdamus: Protected does seem to be what the Theia project defaults to, so yeah. Also, I wonder whether we should make the breaking change of removing the constructor parameters altogether as now they're injected and change them to the preferred property injection. What's the process for making such a decision on deliberate API breakage?

packages/plugin-ext-headless/src/hosted/node/plugin-host-headless.ts
martin-fleck-at: Ah, you are right! Thanks for the explanation. Localization of the package manifest may absolutely make sense! Would be great to document that.

And I believe how that should have finished is (but now I'm not sure that makes sense anymore):
> Do we need those two for some special reason as opposed to the others? Could they also be
... injected using the same mechanism the user would be using (PluginContainerModule? something else?).

cdamus: > ... injected using the same mechanism the user would be using (PluginContainerModule? something else?).

Yes, I definitely want to think about that. Makes perfect sense!

examples/api-provider-sample/src/node/greeting-main-impl.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    private greeterData: Record<string, GreeterData> = {};
+
+    constructor(@inject(RPCProtocol) rpc: RPCProtocol) {
+        this.proxy = rpc.getProxy(MAIN_RPC_CONTEXT.GREETING_EXT);
+    }
+
+    async $createGreeter(): Promise<GreeterData> {
+        const result: GreeterData = {
+            uuid: UUID.uuid4(),
```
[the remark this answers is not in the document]
cdamus: Thanks for the tip! Great not to have to use a Phosphor import, here 😀

packages/plugin-ext/src/plugin/node/plugin-container-module.ts
```diff
@@ -0,0 +1,138 @@
...
+        });
+        return result;
+    }
+});
+
+export type InternalPluginContainerModule = ContainerModule & {
+    initializeApi?: (container: Container) => PluginApiCache<object>;
+};
+
+class PluginApiCache<T extends object> {
```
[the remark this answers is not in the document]
cdamus: Well, the exported `InternalPluginContainerModule` type actually exposes this type, so it should be exported regardless!

packages/plugin-ext-headless/src/hosted/node/scanners/scanner-theia-headless.ts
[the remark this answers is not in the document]
cdamus: BTW, both of these are mentioned as follow-up items in [the PR description](#issue-2021319998).

packages/plugin-ext/src/plugin/workspace.ts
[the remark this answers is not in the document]
martin-fleck-at: I think we can make the change to property injection as this is the [preferred](https://github.com/eclipse-theia/theia/blob/master/doc/coding-guidelines.md#property-injection) way in Theia.

packages/plugin-ext-headless/src/main/node/env-main.ts
cdamus: Actually, now that we are using DI in the plugin hosts, third-party extensions will be able to inject the `EnvExtImpl` into their own ext-classes, so I need to retain it and the main-side implementation as this will be a way to get environment variables.

packages/plugin-ext/src/hosted/node/hosted-plugin-process.ts
cdamus: The goofy part is that we already have a `MessageService` bound. So this doesn't need to be optional.

packages/plugin-dev/src/node/hosted-plugin-reader.ts
```diff
@@ -42,7 +42,7 @@ export class HostedPluginReader implements BackendApplicationContribution {
             const hostedPlugin = new PluginDeployerEntryImpl('Hosted Plugin', pluginPath!, pluginPath);
             hostedPlugin.storeValue('isUnderDevelopment', true);
             const hostedMetadata = await this.hostedPlugin.promise;
-            if (hostedMetadata!.model.entryPoint && hostedMetadata!.model.entryPoint.backend) {
+            if (hostedMetadata!.model.entryPoint && (hostedMetadata!.model.entryPoint.backend || hostedMetadata!.model.entryPoint.headless)) {
```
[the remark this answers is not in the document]
cdamus: I prefer not to expose non-backend-capable plugins to any code that may be out there that asks this API for backend plugins. They won't know that there's a new kind of plugin that they need to start filtering out.

cdamus:
Thanks @jfaltermeier and @martin-fleck-at for your reviews! I think I have addressed most, if not all, of your comments with commit 8eaf8fe4284a93b728caa0efe55530a474fbb35b. Overall I think this really helped to clean up and rationalize the code!

examples/api-provider-sample/src/plugin/gotd-api-init.ts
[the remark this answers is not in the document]
martin-fleck-at: Much better, thank you! ;-)

packages/plugin-dev/src/node/hosted-plugin-reader.ts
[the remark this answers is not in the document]
martin-fleck-at: Fine with me.

packages/plugin-ext-headless/src/hosted/node/scanners/scanner-theia-headless.ts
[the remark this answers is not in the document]
martin-fleck-at: Fantastic, thank you!

packages/plugin-ext-headless/src/hosted/node/plugin-host-headless.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+});
+
+const container = new Container();
+container.bind(PluginContainerModuleLoader).toConstantValue((module: ContainerModule) => {
+    container.load(module);
+    const internalModule = module as InternalPluginContainerModule;
+    const pluginApiCache = internalModule.initializeApi?.(container);
+    if (pluginApiCache) {
+        // TODO
```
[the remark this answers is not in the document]
martin-fleck-at: Great, thank you!

packages/plugin-ext-headless/src/main/node/env-main.ts
martin-fleck-at: Noted.

packages/plugin-ext-headless/src/main/node/message-registry-main.ts
martin-fleck-at: I saw that you refactored the code nicely into a `BasicMessageRegistryMainImpl`, works for me.

packages/plugin-ext/src/hosted/node/hosted-plugin-process.ts
martin-fleck-at: Yay! :D

jcortell68:
@cdamus Not sure if I'm confused, but at some point, this PR had a number of commits that reflected the evolving nature of the change. Today I see just one commit. Did you collapse all the commits into one and force push your branch?

examples/api-provider-sample/src/plugin/gotd-api-init.ts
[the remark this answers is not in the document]
jcortell68: To "dish out" is a folksy way to say "to dispense". I sometimes resort to such informalities to bring a little levity to the code--humanize it a bit. I've seen codebases that read like a standup comedy routine and that's unpleasant. But a sprinkling of clues that we're not robots doesn't seem like such a bad thing...to me, anyway. Of course, this isn't my code so not suggesting it should stay. Just chiming in with the history of how it got there.

jfaltermeier (approved):

martin-fleck-at (approved):
Thank you very much for the detailed explanation and the updated PR. Everything looks good to me now from a code and runtime perspective. Great work! Since this is such a large PR, we'll ask another reviewer to have a look before merging it.

msujew (approved):
This looks alright to me. I can confirm that the feature works as expected and introduces no regressions in the plugins I've tested.

cdamus:
Commit cffddc2a0e07e10788bd79f938c4cf3a3f52616f and successors are rebased to resolve the merge conflict by just deleting the changelog entry as it will be generated automatically.

jcortell68:
Celebration and congratulations are in order!

Issue #13067, opened 2023-11-07
Documentation for adding new plugin api namespace needs updating


jcortell68:
[This markdown](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md) explains how a Theia app developer can add a new plugin API namespace. I followed it a while back and it's been working great.

However, recently, our Theia app was adjusted to consume the backend in webpacked form (something that became possible in Theia 1.40.1), and that broke the custom API namespace. Specifically, the way `backendInitPath` is set is now invalid:

```
@injectable()
export class FooPluginApiProvider implements ExtPluginApiProvider {
    provideApi(): ExtPluginApi {
        return {
            frontendExtApi: {
                initPath: 'my-frontend-api-init',
                initFunction: 'fooInitializationFunction',
                initVariable: 'foo_global_variable'
            },
            backendInitPath: path.join(__dirname, 'my-backend-api-init')
        };
    }
}
```

So, this ticket involves a question and an ask:

Question: how should I set `backendInitPath` so that it works in the backend-webpacked environment
Ask: would be nice for the markdown to be updated with this information so others don't trip on this

jcortell68:
So, with a little help from my friends, turns out the code above doesn't need adjusting. What does need adjusting is the build configuration. Specifically, in the electron app's `webpack.config.fs`, you need an entry to tell webpack to treat `implementation.js` with care:

```
const nodeConfig = require('./gen-webpack.node.config.js');
...
if (nodeConfig.config.entry) {
    nodeConfig.config.entry['implementation'] = {
        import: require.resolve('@myapp/mypackage/lib/node/implementation'),
        library: { type: 'commonjs2' }
    };
}
```

What that effectively does is puts a modified version of the transpiled `implementation.js` into the backend webpack files as its own file. It's modified in that it's webpacked. This way, when Theia dynamically imports the module, it's found.

jcortell68:
I'll leave the ticket open as I do think it would be helpful for the markdown documentation to provide this additional insight. If nothing else, maybe include a warning that the module specified for the API need build-time special handling in the case of bundling webpacked frontend/backend, and then provide a link to this ticket for the details?

PR #13153, opened 2023-12-05
Improve documentation on the addition of Plugin API in the plugin host

martin-fleck-at:
#### What it does

Improves documentation on the API extensions in plugin hosts.

Fixes https://github.com/eclipse-theia/theia/issues/13067

#### How to test

Read, compare, try :-)

#### Follow-ups

No follow-ups.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
+++ b/packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
@@ -16,6 +16,7 @@ However, both APIs can be provided by implementing and binding an `ExtPluginApiP
 ## Declare your plugin API provider
 
 The plugin API provider is executed on the respective plugin host to add your custom API object and namespaces.
+Add `@theia/plugin-ext` as a dependency in your `package.json`
 
 Example Foo Plugin API provider:
 
@@ -50,6 +51,8 @@ Example `foo.d.ts`:
 
 ```typescript
 declare module '@bar/foo' {
+    export class Foo { }
+
     export namespace fooBar {
         export function getFoo(): Promise<Foo>;
     }
@@ -61,7 +64,7 @@ declare module '@bar/foo' {
 In our example, we aim to provide a new API object for the backend.
 Theia expects that the `backendInitPath` that we specified in our API provider is a function called `provideApi` that follows the `ExtPluginApiBackendInitializationFn` signature.
 
-Example `foo-init.ts`:
+Example `node/foo-init.ts`:
 
 ```typescript
 import * as fooBarAPI from '@bar/foo';
@@ -132,7 +135,7 @@ We create a dedicated API object for each individual plugin as part of the modul
 Each API object is returned as part of the module loading process if a script imports `@bar/foo` and should therefore match the API definition that we provided in the `*.d.ts` file.
 Multiple imports will not lead to the creation of multiple API objects as we cache it in our custom `overrideInternalLoad` function.
 
-Example `foo-init.ts` (continued):
+Example `node/foo-init.ts` (continued):
 
 ```typescript
 export function createAPIFactory(rpc: RPCProtocol): ApiFactory {
@@ -175,19 +178,19 @@ export interface FooExt {
 }
 
 // Plugin host will obtain a proxy using these IDs, main application will register an implementation for it.
-export const PLUGIN_RPC_CONTEXT = {
+export const FOO_PLUGIN_RPC_CONTEXT = {
     FOO_MAIN: createProxyIdentifier<FooMain>('FooMain')
 };
 
 // Main application will obtain a proxy using these IDs, plugin host will register an implementation for it.
-export const MAIN_RPC_CONTEXT = {
+export const FOO_MAIN_RPC_CONTEXT = {
     FOO_EXT: createProxyIdentifier<FooExt>('FooExt')
 };
 ```
 
 On the plugin host side we can register our implementation and retrieve the proxy as part of our `createAPIFactory` implementation:
 
-Example `foo-ext.ts`:
+Example `plugin/foo-ext.ts`:
 
 ```typescript
 export class FooExtImpl implements FooExt {
@@ -195,8 +198,8 @@ export class FooExtImpl implements FooExt {
     private proxy: FooMain;
 
     constructor(rpc: RPCProtocol) {
-        rpc.set(MAIN_RPC_CONTEXT.FOO_EXT, this); // register ourselves
-        this.proxy = rpc.getProxy(PLUGIN_RPC_CONTEXT.FOO_MAIN); // retrieve proxy
+        rpc.set(FOO_MAIN_RPC_CONTEXT.FOO_EXT, this); // register ourselves
+        this.proxy = rpc.getProxy(FOO_PLUGIN_RPC_CONTEXT.FOO_MAIN); // retrieve proxy
     }
 
     getFooImpl(): Promise<Foo> {
@@ -207,7 +210,7 @@ export class FooExtImpl implements FooExt {
 
 On the main side we need to implement the counterpart of the ExtPluginApiProvider, the `MainPluginApiProvider`, and expose it in a browser frontend module:
 
-Example `foo-main.ts`:
+Example `main/browser/foo-main.ts`:
 
 ```typescript
 @injectable()
@@ -217,7 +220,7 @@ export class FooMainImpl implements FooMain {
 
     init(rpc: RPCProtocol) {
         // We would use this if we had a need to call back into the plugin-host/plugin
-        this.proxy = rpc.getProxy(MAIN_RPC_CONTEXT.FOO_EXT);
+        this.proxy = rpc.getProxy(FOO_MAIN_RPC_CONTEXT.FOO_EXT);
     }
 
     async $getFooImpl(): Promise<Foo> {
@@ -235,7 +238,7 @@ export class FooMainPluginApiProvider implements MainPluginApiProvider {
         // create a new FooMainImpl as it is not bound as singleton
         const fooMainImpl = container.get(FooMainImpl);
         fooMainImpl.init(rpc);
-        rpc.set(PLUGIN_RPC_CONTEXT.FOO_MAIN, fooMainImpl);
+        rpc.set(FOO_PLUGIN_RPC_CONTEXT.FOO_MAIN, fooMainImpl);
     }
 }
 
@@ -260,6 +263,7 @@ foo.fooBar.getFoo();
 ## Packaging
 
 When bundling our application with the generated `gen-webpack.node.config.js` we need to make sure that our initialization function is bundled as a `commonjs2` library so it can be dynamically loaded.
+Adjust the `webpack.config.js` accordingly:
 
 ```typescript
 const configs = require('./gen-webpack.config.js');
--- a/packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
+++ b/packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
@@ -1,17 +1,21 @@
-# How to add a new plugin API namespace
+# How to add new custom plugin API
 
-This document describes how to add new plugin API namespace in the plugin host.
-Depending on the plugin host we can either provide a frontend or backend API extension:
+As a Theia developer, you might want to make your app extensible by plugins in ways that are unique to your application.
+That will require API that goes beyond what's in the VS Code Extension API and the Theia plugin API.
+You can do that by implementing a Theia extension that creates and exposes an API object within the plugin host.
+The API object can be imported by your plugins and exposes one or more API namespaces.
+
+Depending on the plugin host we can either provide a frontend or backend plugin API:
 
 - In the backend plugin host that runs in the Node environment in a separate process, we adapt the module loading to return a custom API object instead of loading a module with a particular name.
 - In the frontend plugin host that runs in the browser environment via a web worker, we import the API scripts and put it in the global context.
 
-In this document we focus on the implementation of a backend plugin API.
+In this document we focus on the implementation of a custom backend plugin API.
 However, both APIs can be provided by implementing and binding an `ExtPluginApiProvider` which should be packaged as a Theia extension.
 
 ## Declare your plugin API provider
 
-The plugin API provider is executed on the respective plugin host to add your custom API namespace.
+The plugin API provider is executed on the respective plugin host to add your custom API object and namespaces.
 
 Example Foo Plugin API provider:
 
@@ -149,7 +153,7 @@ export function createAPIFactory(rpc: RPCProtocol): ApiFactory {
 In the example above the API object creates a local object that will fulfill the API contract.
 The implementation details are hidden by the object and it could be a local implementation that only lives inside the plugin host but it could also be an implementation that uses the `RPCProtocol` to communicate with the main application to trigger changes, register functionality or retrieve information.
 
-### Implementing Main-Ext communication
+### Implement Main-Ext communication
 
 In this document, we will only highlight the individual parts needed to establish the communication between the main application and the external plugin host.
 For a more elaborate example of an API that communicates with the main application, please have a look at the definition of the [Theia Plugin API](https://github.com/eclipse-theia/theia/blob/master/doc/Plugin-API.md).
--- a/packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
+++ b/packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
@@ -1,26 +1,23 @@
-# This document describes how to add new plugin api namespace
+# How to add a new plugin API namespace
 
-New Plugin API namespace should be packaged as Theia extension
+This document describes how to add new plugin API namespace in the plugin host.
+Depending on the plugin host we can either provide a frontend or backend API extension:
 
-## Provide your API or namespace
+- In the backend plugin host that runs in the Node environment in a separate process, we adapt the module loading to return a custom API object instead of loading a module with a particular name.
+- In the frontend plugin host that runs in the browser environment via a web worker, we import the API scripts and put it in the global context.
 
-This API developed in the way that you provide your API as separate npm package.
-In that package you can declare your api.
-Example `foo.d.ts`:
+In this document we focus on the implementation of a backend plugin API.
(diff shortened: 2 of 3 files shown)


jcortell68:
@martin-fleck-at This is looking good. I notice you haven't added anyone as a reviewer, so maybe this isn't ready for review. But I'll add this comment while I have it on my mind.

One gap that existed before and persists in your draft is that the guide doesn't touch on the need to have a TypeScript file that declares the interfaces and defines constants to refer to the RPC proxies for both side of the interface. In a rescent related PR, that would be the file 
```
examples/api-provider-sample/src/common/plugin-api-rpc.ts
```
in [this recent PR](https://github.com/eclipse-theia/theia/pull/13138)

packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
```diff
@@ -1,22 +1,19 @@
...
-            export function getFoo(): Foo;
-        }
-    }
-```
+## Declare your plugin API provider
+
+The plugin API provider is executed on the respective plugin host to add your custom API namespace.
 
-## Declare `ExtPluginApiProvider` implementation
+Example Foo Plugin API provider:
```
jcortell68, line 16: I recommend we name this `FooExtPluginApiProvider`. There is a `MainPluginApiProvider` that needs to be implemented as well.

jcortell68:
Another gap that persists is that there is no mention of the need to implement and contribute (via DI) a `MainPluginApiProvider`

packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
```diff
@@ -90,23 +122,59 @@ function findPlugin(filePath: string): Plugin | undefined {
 }
 ```
 
-## Next you need to implement `createAPIFactory` factory function
+## Implement your API object
```
jcortell68, line 125: The `findPlugin()` implementation at line 120 (can't add a comment there) doesn't work when the plugin is found via a symlink, which is typically the scenario during development. See [12840](https://github.com/eclipse-theia/theia/issues/12840) for a more detailed explanation of the problem and how to address it.

martin-fleck-at: I don't think that this problem exists anymore as the issue that you referenced was actually closed and solved outside the external plugin API by ensuring that the `manifest.packagePath` already has a properly resolved path.

packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
martin-fleck-at: Good idea, I'll change it.

jcortell68:
> I think that the reason for the missing `Main` part is that in theory you could provide an API object that just does some local work without the need to go to the main side. 

Oh wow. I never thought of that. Indeed it could. That API would be entirely serviced in the plugin-host. Very interesting. But yes, I would imagine that's more of an academic use case than one likely to actually happen. So yes, please...let's get the "likely" scenario documented here and maybe add a note that it's technically optional, but likely to be needed.

packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
jcortell68: Ahhh. You know, I somehow missed all the activity after the initial commit/fix was proposed in that pull request--the one that relied on `fs.realpathSync`! I had no idea the solution was reworked and the fix moved elsewhere in two subsequent commits. I still have the original fix in my product...lol. Guess I can remove that!

packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
```diff
@@ -90,23 +122,153 @@ function findPlugin(filePath: string): Plugin | undefined {
...
+    $getFooImpl(): Promise<Foo>;
+}
+
+export interface FooExt {
+    // placeholder for callbacks for the main application to the extension
+}
+
+// Plugin host will obtain a proxy using these IDs, main application will register an implementation for it.
+export const PLUGIN_RPC_CONTEXT = {
+    FOO_MAIN: <ProxyIdentifier<FooMain>>createProxyIdentifier<FooMain>('FooMain')
```
jcortell68, line 175: Very nice job on this new section!

If I'm not mistaken, I think the `<ProxyIdentifier<FooMain>>` can be omitted.  Same idea on line 180.

packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
```diff
@@ -90,23 +122,153 @@ function findPlugin(filePath: string): Plugin | undefined {
...
-        return <typeof fooApi>{
+        return <typeof fooBarAPI>{
             fooBar : FooBar
         };
     }
+}
+```
+
+In the example above the API object creates a local object that will fulfill the API contract.
+The implementation details are hidden by the object and it could be a local implementation that only lives inside the plugin host but it could also be an implementation that uses the `RPCProtocol` to communicate with the main application to retrieve additional information.
```
jcortell68, line 150: "to retrieve additional information" may give the wrong idea to the reader. Often, an API triggers changes or drives functionality on the main side of things. Yes, sometimes it just gets information, but that's probably more the exception than the rule. The wording here should cover both cases.

packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
martin-fleck-at: Absolutely correct, thanks!

packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
martin-fleck-at: Yeah, I get that. Good catch! I adapted the sentence to cover more cases and give a better idea of what could be possible.

packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
```diff
@@ -90,23 +122,153 @@ function findPlugin(filePath: string): Plugin | undefined {
...
+        return new Foo();
+    }
+}
+
+@injectable()
+export class FooMainPluginApiProvider implements MainPluginApiProvider {
+    @inject(MessageService) protected messageService: MessageService;
+
+    initialize(rpc: RPCProtocol, container: interfaces.Container): void {
+        this.messageService.info('We were called from an extension!');
```
jcortell68, line 227: Should this info message be here or at line 218?...with the message being:
`We were called from the plugin-host at the behest of the plugin`

martin-fleck-at: I like that idea and so I added _another_ one at line 218. I just wanted to have this message here as well to showcase that we are part of the dependency injection graph so we can simply inject whatever we need. But that is true for both classes so I added just another message ;-)

packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
```diff
@@ -90,23 +122,153 @@ function findPlugin(filePath: string): Plugin | undefined {
...
+
+Example `foo-main.ts`:
+
+```typescript
+@injectable()
+export class FooMainImpl implements FooMain {
+    protected proxy: FooExt;
+
+    init(rpc: RPCProtocol) {
+        this.proxy = rpc.getProxy(MAIN_RPC_CONTEXT.FOO_EXT);
```
jcortell68, line 215: This example understandably doesn't provide an example of the main side of things calling back into the plugin-host. But it would be helpful to have a comment here, something along the lines of
```
# We would use this if we had a need to call back into the plugin-host/plugin.
this.proxy = rpc.getProxy(MAIN_RPC_CONTEXT.FOO_EXT);
```

martin-fleck-at: Good point, I added the comment you suggested.

packages/plugin-ext/doc/how-to-add-new-plugin-namespace.md
jcortell68: With your change to the message string (in the responding commit), this message location makes sense to me, too.

jcortell68:
I've completed the detailed review. Again, this is great stuff. I just have one final comment on the title (`How to add a new plugin API namespace`) since it's something that I initially scratched my head on a bit. My first reaction when I came across it was...Is this telling me how to add a new API namespace to the VS Code Extension API (if I'm working on compatibility gaps), or likewise, if I'm trying to upstream a Theia enhancement and want to add something to the Theia plugin API (less likely). This guide is definitely not about those things, but that's not clear from the title.

The problem here is that this isn't just about API _namespaces_. This article is meant for Theia app developers who want to introduce custom API for use by a Theia plugin. Plugin API is made up of numerous namespaces, but they are scoped by API objects. I.e. there is an API object that specifically provides both VS Code Extension API and Theia plugin API, and that API object is contributed by Theia itself. As an app developer, I'm not looking to add API namespaces to that API object. I want to develop a Theia extension that exposes a **custom plugin API object** (that has custom plugin API namespaces). Again, the current title of this guide doesn't really make that clear. As such, I recommend renaming it to something like `How to add custom plugin API to your Theia application`. And then add some paragraph that explains the usecase. Something along the lines of

`As a Theia app developer, you might want make your app extensible by plugins in ways that are unique to your app. That will require API that goes beyond what's in the VS Code Extension API and the Theia plugin API. You can do that by implementing a Theia extension that creates and exposes an API object to plugins. The object will have one or more API namespaces. A plugin will access the API object via an app-specific JS module`

The above suggestion is just that--suggested wording. It's the messaging that's important.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -0,0 +1,281 @@
...
+
+## Define your API
+
+To ease the usage of your API, it should be developed as separate npm package that can be easily imported without any additional dependencies, cf, the VS Code API or the Theia Plugin API.
+
+Example `foo.d.ts`:
+
+```typescript
+declare module '@bar/foo' {
+    export namespace fooBar {
```
jfaltermeier, line 53:
```suggestion

    export class Foo { }

    export namespace fooBar {
```

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -0,0 +1,281 @@
...
+    }
+}
+```
+
+## Implement your plugin API provider
+
+In our example, we aim to provide a new API object for the backend.
+Theia expects that the `backendInitPath` that we specified in our API provider is a function called `provideApi` that follows the `ExtPluginApiBackendInitializationFn` signature.
+
+Example `foo-init.ts`:
```
jfaltermeier, line 64:
```suggestion
Example `node/foo-init.ts`:
```

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
jfaltermeier: As far as I could see the `Foo` object is no created anywhere else

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
jfaltermeier: I think it might make sense to include suitable parent directories for the sample files

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -0,0 +1,281 @@
...
+}
+```
+
+## Implement your API object
+
+We create a dedicated API object for each individual plugin as part of the module loading process.
+Each API object is returned as part of the module loading process if a script imports `@bar/foo` and should therefore match the API definition that we provided in the `*.d.ts` file.
+Multiple imports will not lead to the creation of multiple API objects as we cache it in our custom `overrideInternalLoad` function.
+
+Example `foo-init.ts` (continued):
```
jfaltermeier, line 135:
```suggestion
Example `node/foo-init.ts` (continued):
```

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -0,0 +1,281 @@
...
+
+```typescript
+export function createAPIFactory(rpc: RPCProtocol): ApiFactory {
```
jfaltermeier, line 138:
```suggestion
export function createAPIFactory(rpc: RPCProtocol): (plugin: Plugin) => typeof fooBarAPI {
```

jfaltermeier: I am not sure where `ApiFactory` is coming from

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -0,0 +1,281 @@
...
+
+// Main application will obtain a proxy using these IDs, plugin host will register an implementation for it.
+export const MAIN_RPC_CONTEXT = {
+    FOO_EXT: createProxyIdentifier<FooExt>('FooExt')
+};
+```
+
+On the plugin host side we can register our implementation and retrieve the proxy as part of our `createAPIFactory` implementation:
+
+Example `foo-ext.ts`:
```
jfaltermeier, line 190:
```suggestion
Example `plugin/foo-ext.ts`:
```

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -0,0 +1,281 @@
...
+
+    getFooImpl(): Promise<Foo> {
+        return this.proxy.$getFooImpl();
+    }
+}
+```
+
+On the main side we need to implement the counterpart of the ExtPluginApiProvider, the `MainPluginApiProvider`, and expose it in a browser frontend module:
+
+Example `foo-main.ts`:
```
jfaltermeier, line 210:
```suggestion
Example `main/browser/foo-main.ts`:
```

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -0,0 +1,281 @@
...
+
+```typescript
+import * as foo from '@bar/foo';
+
+foo.fooBar.getFoo();
+```
+
+## Packaging
+
+When bundling our application with the generated `gen-webpack.node.config.js` we need to make sure that our initialization function is bundled as a `commonjs2` library so it can be dynamically loaded.
```
jfaltermeier, line 262:
```suggestion
When bundling our application with the generated `gen-webpack.node.config.js` we need to make sure that our initialization function is bundled as a `commonjs2` library so it can be dynamically loaded. Adjust `webpack.config.js` accordingly:
```

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -0,0 +1,281 @@
...
+
+- In the backend plugin host that runs in the Node environment in a separate process, we adapt the module loading to return a custom API object instead of loading a module with a particular name.
+- In the frontend plugin host that runs in the browser environment via a web worker, we import the API scripts and put it in the global context.
+
+In this document we focus on the implementation of a custom backend plugin API.
+However, both APIs can be provided by implementing and binding an `ExtPluginApiProvider` which should be packaged as a Theia extension.
+
+## Declare your plugin API provider
+
+The plugin API provider is executed on the respective plugin host to add your custom API object and namespaces.
```
jfaltermeier, line 18:
```suggestion
The plugin API provider is executed on the respective plugin host to add your custom API object and namespaces. Add `@theia/plugin-ext` as a dependency in your `package.json`
```

jcortell68 (approved):
Looks great to me!

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -0,0 +1,281 @@
...
+export interface FooMain {
+    $getFooImpl(): Promise<Foo>;
+}
+
+export interface FooExt {
+    // placeholder for callbacks for the main application to the extension
+}
+
+// Plugin host will obtain a proxy using these IDs, main application will register an implementation for it.
+export const PLUGIN_RPC_CONTEXT = {
```
jfaltermeier, line 178: Maybe we should use different names than `PLUGIN_RPC_CONTEXT` and `MAIN_RPC_CONTEXT`, because those are exported from `@theia/plugin-ext'` as well. Users might pick the wrong import then.

jfaltermeier (commented):
Thanks, looks very good already. 
I've added a few comments/suggestions on the code examples that I noticed while following the docs.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
jfaltermeier: My bad, I just saw where it is defined ;-)

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
martin-fleck-at: Sure, I added it as a placeholder, thank you for catching that!

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
martin-fleck-at: I love that idea, I added them as you suggested.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
martin-fleck-at: I added a `FOO_` prefix to make it even more obvious, thank you!

jcortell68: Ah, the age-old dilemma of when to scope-qualify symbols not because they're needed by the code but to make it easier on humans navigating the code. Some would argue the longer names clutter up the source (FOO is short enough to make that a weak argument in this case). Ultimately, it's very subjective, and as with most such discussions, I try to stay far awau...except of course to highlight the dilemma itself. That part is fun and non-judgmental 😆

jfaltermeier (approved):

Issue #13107, opened 2023-11-27
WebView of the same vscode plugin not loading in second electron instances


eneufeld:
### Bug Description:
Running Theia in Electron (e.g as Blueprint) twice and trying to open the same WebView in both instances fails. The first instances opens the WebView the second one fails with:

```
Uncaught (in promise) DOMException: Failed to register a ServiceWorker: The document is in an invalid state.
```
The error is thrown in:
`host.js`
`navigator.serviceWorker.register("service-worker.js").then((async a=>{`

### Steps to Reproduce:

1. Install the Draw.io Integration extension
2. start a second theia blueprint instance
3. Run the "Draw.io: New Draw.io Diagram" Command and leave the diagram view open
4. switch to second Theia instance
5. Run the "Draw.io: New Draw.io Diagram" Command again and see a view opening but no content rendered, see the logged error in the Developers Tools

### Additional Information

- Operating System: Linux Ubuntu 22.04
- Theia Version: 1.43.0 Beta

tsmaeder:
Just to be sure: "a second instance" is a whole new process, not just a new window, right?

msujew:
Since this only happens when starting the application multiple times, this is related to https://github.com/eclipse-theia/theia/issues/10890.

tsmaeder:
From all I'm reading I get that running multiple process on the same data directory (containing indexdb and other files) is not supported in electron. Note that VS Code does not seem to support this either: if you do `code` on the command line a second time, you do not get a node `code.exe` process.
Of course, we could create a separate data area for each electron main process instance, but then sharing and persisting stuff per user, for example via the `localStorage` would not work anymore. I would be interested in knowing the use case we're trying to cover by starting multiple processes.

msujew:
@tsmaeder I believe https://github.com/eclipse-theia/theia/issues/10890 explains one of thise use cases. I.e. starting two Theia instances with a different set of env variables. IMO a weird use case, but I didn't focus on the issue any further, so nothing came out of it.

tsmaeder:
If the idea is just to let the user open more windows via `theia ./myWorkspace` the `secondInstance` even on the app has all the info we need. If we really need to run two copies, we'd have to have some support for using a different data area (cli or a profile mechanism like Firefox has), but running twice on the same data are is not going to fly, IMO.

msujew:
[...]

PR #13155, opened 2023-12-05
Add a cli parameter '--userDataArea` to set the Electron 'userData' path

tsmaeder:
This PR allows to set a non-default Electron `userData` path in order to allow two copies of Theia without concurrency issues related to shared `localStorage` etc. 

Part of #13107

#### What it does
Adds the relevant cli parameter.

#### How to test
Run two instances of Theia with different user data areas and make sure webviews work in both instances and the there are no concurrency-related issues show up in the log.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-main/electron-main-application.ts
+++ b/packages/core/src/electron-main/electron-main-application.ts
@@ -57,7 +57,7 @@ export interface ElectronMainCommandOptions {
     readonly cwd: string;
 
     /**
-     * If the app is launched for the first time, `secondInstance` is false. 
+     * If the app is launched for the first time, `secondInstance` is false.
      * If the app is already running but user relaunches it, `secondInstance` is true.
      */
     readonly secondInstance: boolean;
--- a/packages/core/src/electron-main/electron-main-application.ts
+++ b/packages/core/src/electron-main/electron-main-application.ts
@@ -55,6 +55,11 @@ export interface ElectronMainCommandOptions {
     readonly file?: string;
 
     readonly cwd: string;
+
+    /**
+     * If the app is launched for the first time, `secondInstance` is false. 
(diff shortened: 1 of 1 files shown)


packages/core/src/electron-main/electron-main-application.ts
```diff
@@ -54,19 +54,8 @@ export interface ElectronMainCommandOptions {
...
- *
- * This kind of event is triggered in two different contexts:
- *  1. The app is launched for the first time, `secondInstance` is false.
- *  2. The app is already running but user relaunches it, `secondInstance` is true.
- */
-export interface ElectronMainExecutionParams {
-    readonly secondInstance: boolean;
-    readonly argv: string[];
     readonly cwd: string;
+    readonly secondInstance: boolean;
```
planger, line 58: I think we should transfer the type doc comments for `secondInstance`, as it may not be obvious what that means.

packages/core/src/electron-main/electron-main-application.ts
```diff
@@ -212,21 +201,38 @@ export class ElectronMainApplication {
...
-        await this.launch({
-            secondInstance: false,
-            argv: args,
-            cwd: process.cwd()
-        });
+        const argv = this.processArgv.getProcessArgvWithoutBin(process.argv);
+        createYargs(argv, process.cwd())
+            .command('$0 [file]', false,
+                cmd => cmd
+                    .option('userDataArea', {
```
planger, line 208: I suggest using `electronUserData` to clarify that this is about the Electron user data, and not to confuse with the config location (`.theia` by default or defined by `$THEIA_CONFIG_DIR`). WDYT?

planger (changes requested):
Thank you! Code makes sense and works fine. I just have two rather minor comments inline.

planger (approved):

Issue #13114, opened 2023-11-28
File Explorer displays wrong indent guide with compact folders


bryanchenmchp:
### Bug Description:

When compact folders are displayed in the file explorer, the highlighted tree indentation guide is misaligned.

![image](https://github.com/eclipse-theia/theia/assets/146364035/e5073de2-24d0-47a7-b998-4ab3b064cd72) ![image](https://github.com/eclipse-theia/theia/assets/146364035/a2e4ef63-942a-4958-943f-02725d03f6a1) 
![image](https://github.com/eclipse-theia/theia/assets/146364035/f619b596-946c-405d-bc75-10b025d93ce6)

### Steps to Reproduce:

1. Open any directory that includes within it a compact folder (single folder inside another folder)
2. Focusing a file or directory node causes the guide lines to be offset

### Additional Information

- Operating System: Windows 11
- Theia Version: 1.42

PR #13162, opened 2023-12-10
Fix compressed navigator indent highlight

msujew:
#### What it does

Fixes https://github.com/eclipse-theia/theia/issues/13114

The idea for this fix is fairly straighforward: During the rendering process, simply ignore all tree nodes that don't contribute a separate indentation to the navigator. In the compressed tree, only the head node of the compression chain actually contributes an indentation marker, so we need to check that in the new `shouldRenderIndent` method.

#### How to test

1. Create a workspace similar to what's outlined in #13114 
2. Assert that the navigator indent highlight is being correctly calculated

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/tree/tree-widget.tsx
+++ b/packages/core/src/browser/tree/tree-widget.tsx
@@ -894,20 +894,31 @@ export class TreeWidget extends ReactWidget implements StatefulWidget {
         let current: TreeNode | undefined = node;
         let depth = props.depth;
         while (current && depth) {
-            const classNames: string[] = [TREE_NODE_INDENT_GUIDE_CLASS];
-            if (this.needsActiveIndentGuideline(current)) {
-                classNames.push('active');
-            } else {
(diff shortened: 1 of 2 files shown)


packages/core/src/browser/tree/tree-widget.tsx
```diff
@@ -894,22 +894,33 @@ export class TreeWidget extends ReactWidget implements StatefulWidget {
...
-            }} />);
             current = current.parent;
-            depth--;
         }
         return indentDivs;
     }
 
+    /**
+     * Determines whether an indentation div should be rendered for the specified tree node.
+     * If there are multiple tree node inside of a single rendered row,
```
martin-fleck-at, line 917:
```suggestion
     * If there are multiple tree nodes inside of a single rendered row,
```

nit

martin-fleck-at (changes requested):
Change looks good to me and works well, thank you very much @msujew! I just have a very nit-picky issue in one of the docs.

martin-fleck-at (approved):
LGTM!

PR #13171, opened 2023-12-12
Fix notebook kernel selection

### LINKED ISSUE #13167: Connecting to an existing Jupyter Kernel not possible
### Bug Description:

After installing the Python and Jupyter notebook extensions, users cannot use the `Existing Jupyter Kernel` quick pick entry for selecting the kernel that's used to run the notebook. Every time it is clicked, an error in the plugin host occurs that is thrown on [this line](https://github.com/eclipse-theia/theia/blob/469bd744a02e5dcdbcbc338af178d410dd909616/packages/plugin-ext/src/plugin/command-registry.ts#L209).

### Steps to Reproduce:

1. Install all the necessary Jupyter plugins
2. Try to select the `Existing Jupyter Kernel` to run a cell.

### Additional Information

- Operating System: Ubuntu
- Theia Version: 1.44.0

msujew:
#### What it does

Closes #13167

Fixes quite a few issues which prevented the notebook kernel selection from working correctly. The most major issue was related to the fact that the safe command wasn't transmitted correctly (it was missing its arguments). See changes in the `command.ts` and `command-registry.ts` file.

There where also a few issues related to the way we communicated open/focused editors to the plugin host. This was fixed in `notebook-editor-widget-service.ts` and `notebook-documents-and-editors-main.ts`.

#### How to test

1. Install the Jupyter notebook related extensions.
2. Select a different notebook kernel. All options should work as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/notebook-editor-widget.tsx
+++ b/packages/notebook/src/browser/notebook-editor-widget.tsx
@@ -28,6 +28,7 @@ import { inject, injectable, interfaces, postConstruct } from '@theia/core/share
 import { Emitter } from '@theia/core/shared/vscode-languageserver-protocol';
 import { NotebookEditorWidgetService } from './service/notebook-editor-widget-service';
 import { NotebookMainToolbarRenderer } from './view/notebook-main-toolbar';
+import { Deferred } from '@theia/core/lib/common/promise-util';
 
 export const NotebookEditorWidgetContainerFactory = Symbol('NotebookEditorWidgetContainerFactory');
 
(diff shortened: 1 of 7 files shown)


packages/core/src/common/command.ts
```diff
@@ -46,6 +46,13 @@ export interface Command {
      */
     category?: string;
     originalCategory?: string;
+    /**
+     * Default arguments for the command.
+     *
+     * Arguments specified in the command invocation are appended to this array.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    arguments?: any[];
```
msujew, line 55: @tsmaeder What are your thoughts on this change? I'm not sure where these changes are supposed to go otherwise.

jonah-iden (approved):
Code looks good to me and everything seems to work fine. 
Tested with different notebook kernels (python installations, venv and conda) and [jupyter_server](https://github.com/jupyter-server/jupyter_server)

the extra options in the cell toolbars contributed by jupyter are not yet working. But i don't think thats in the scope of this PR

packages/core/src/common/command.ts
jonah-iden: to me this seems to be just an additional feature for commands in general so why should this be a problem?

msujew: FYI, I improved on this. Changes to the command/command registry are no longer necessary.

jonah-iden: ah ok nice. Looks good

Issue #13145, opened 2023-12-04
[vscode] new contextual menu for TestMessages: testing/message/context and testing/message/content


rschnekenbu:
vscode 1.84 introduced 2 new menu contributions points:
- testing/message/context
- testing/message/content

see https://code.visualstudio.com/updates/v1_84#_finalized-testmessagecontextvalue-api

rschnekenbu:
This task will be partially done in the context of #13176. As indicated in #13176, the `testing/message/content` makes currently no sense given  current testing UI implementation.

rschnekenbu:
This one will be closed, and an addition will be done to the declarative API issue here: #13051

PR #13176, opened 2023-12-14
[vscode] Support TestMessage#contextValue

### LINKED ISSUE #13144: [vscode] Support TestMessage#contextValue optional property
new API has been finalized for TestMessage with 1.84:

TestMessage_contextValue

links:
https://code.visualstudio.com/updates/v1_84#_finalized-testmessagecontextvalue-api

rschnekenbu:
#### What it does

Provides support for 1.84 vscode API `TestMessage#contextValue` optional property. This also adds one introduced menu using this API (`testing/message/context`). See https://code.visualstudio.com/updates/v1_84#_finalized-testmessagecontextvalue-api  
The other menu is not relevant currently, as the test API implementation is not as advanced as the vscode one (no diff editor with actual and expected compared for example).

Fixes #13144

#### How to test 
The test can be done installing the following extension:
- zip: 
[test-extension-sample-0.0.1.zip](https://github.com/eclipse-theia/theia/files/13670849/test-extension-sample-0.0.1.zip)
- src: 
[test-extension-sample-0.0.1-src.zip](https://github.com/eclipse-theia/theia/files/13670853/test-extension-sample-0.0.1-src.zip)

This adds a test controller that required the creation of a test run profile before being used (_Command palette_ > _Create Test Run Profile_). Once the test has run, the results are shown. On the test results, there is a new context menu action that displays a message information with the arguments passed to this command. The expected arguments are described in the comment of TestMessage#contextValue (see theia.d.ts - line to be added once commited)

issue-13144

Note that since the `testing/message/content` extension menu is not yet implemented, the extension will show a warning at startup.

#### Follow-ups

Support of the `testing/message/content` extension menu. This is tracked under #13145

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/common/test-types.ts
+++ b/packages/plugin-ext/src/common/test-types.ts
@@ -84,6 +84,7 @@ export interface TestMessageDTO {
     readonly actual?: string;
     readonly location?: Location;
     readonly message: string | MarkdownString;
+    readonly contextValue?: string;
 }
 
 export interface TestItemDTO {
@@ -131,3 +132,22 @@ export namespace TestItemReference {
     }
 }
 
+export interface TestMessageArg {
+    testItemReference: TestItemReference | undefined,
+    testMessage: TestMessageDTO
+}
+
+export namespace TestMessageArg {
+    export function is(arg: unknown): arg is TestMessageArg {
+        return isObject<TestMessageArg>(arg)
+            && isObject<TestMessageDTO>(arg.testMessage)
+            && (MarkdownString.is(arg.testMessage.message) || typeof arg.testMessage.message === 'string');
+    }
+
+    export function create(testItemReference: TestItemReference | undefined, testMessageDTO: TestMessageDTO): TestMessageArg {
+        return {
+            testItemReference: testItemReference,
+            testMessage: testMessageDTO
+        };
+    }
+}
--- a/packages/plugin-ext/src/main/browser/menus/plugin-menu-command-adapter.ts
+++ b/packages/plugin-ext/src/main/browser/menus/plugin-menu-command-adapter.ts
@@ -23,10 +23,13 @@ import { ScmRepository } from '@theia/scm/lib/browser/scm-repository';
 import { ScmService } from '@theia/scm/lib/browser/scm-service';
 import { TimelineItem } from '@theia/timeline/lib/common/timeline-model';
 import { ScmCommandArg, TimelineCommandArg, TreeViewItemReference } from '../../../common';
+import { TestItemReference, TestMessageArg } from '../../../common/test-types';
 import { PluginScmProvider, PluginScmResource, PluginScmResourceGroup } from '../scm-main';
 import { TreeViewWidget } from '../view/tree-view-widget';
 import { CodeEditorWidgetUtil, codeToTheiaMappings, ContributionPoint } from './vscode-theia-menu-mappings';
 import { TAB_BAR_TOOLBAR_CONTEXT_MENU } from '@theia/core/lib/browser/shell/tab-bar-toolbar';
+import { TestItem, TestMessage } from '@theia/test/lib/browser/test-service';
+import { fromLocation } from '../hierarchy/hierarchy-types-converters';
 
 export type ArgumentAdapter = (...args: unknown[]) => unknown[];
 
@@ -79,6 +82,7 @@ export class PluginMenuCommandAdapter implements MenuCommandAdapter {
     @postConstruct()
     protected init(): void {
         const toCommentArgs: ArgumentAdapter = (...args) => this.toCommentArgs(...args);
+        const toTestMessageArgs: ArgumentAdapter = (...args) => this.toTestMessageArgs(...args);
         const firstArgOnly: ArgumentAdapter = (...args) => [args[0]];
         const noArgs: ArgumentAdapter = () => [];
         const toScmArgs: ArgumentAdapter = (...args) => this.toScmArgs(...args);
@@ -100,6 +104,7 @@ export class PluginMenuCommandAdapter implements MenuCommandAdapter {
             ['scm/resourceGroup/context', toScmArgs],
             ['scm/resourceState/context', toScmArgs],
             ['scm/title', () => [this.toScmArg(this.scmService.selectedRepository)]],
+            ['testing/message/context', toTestMessageArgs],
             ['timeline/item/context', (...args) => this.toTimelineArgs(...args)],
             ['view/item/context', (...args) => this.toTreeArgs(...args)],
             ['view/title', noArgs],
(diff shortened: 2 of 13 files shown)


rschnekenbu:
@tsmaeder, as you worked on the test API, could you have a look to this one?

packages/test/src/browser/view/test-run-widget.tsx
```diff
@@ -216,6 +218,12 @@ export class TestRunTreeWidget extends TreeWidget {
                     onDidAddTestOutput: Event.map(node.parent.run.onDidChangeTestOutput, evt => evt.filter(item => item[0] === node.item).map(item => item[1]))
                 };
                 this.uiModel.selectedTestState = node.parent.run.getTestState(node.item);
+                if (this.uiModel.selectedTestState && TestFailure.is(this.uiModel.selectedTestState.state)) {
```
tsmaeder, line 221: Shouldn't this code live in the `uiModel` as it's the owner of the `selectedTestState`?

tsmaeder (commented):
Just two remarks.

packages/plugin-ext/src/common/test-types.ts
```diff
@@ -131,3 +132,23 @@ export namespace TestItemReference {
     }
 }
 
+export interface TestMessageArg {
+    typeTag: '$type_test_message_arg',
```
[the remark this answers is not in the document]
rschnekenbu: That was easier to detect the kind of message, as we send an array of 2 elements with one optional and the other with almost everything optional except 1 property. But this is definitely not mandatory. 
That can be removed and we can have the following code for TestMessageArg.is():
```suggestion
export interface TestMessageArg {
    testItemReference: TestItemReference | undefined,
    testMessage: TestMessageDTO
}

export namespace TestMessageArg {
    export function is(arg: unknown): arg is TestMessageArg {
        return isObject<TestMessageArg>(arg)
            && isObject<TestMessageDTO>(arg.testMessage)
            && typeof arg.testMessage.message === 'string';
    }
```

packages/test/src/browser/view/test-run-widget.tsx
rschnekenbu: Thanks for noticing! Indeed, that should happen in the uiModel. I will update the code and move this part of the code on the selectedTestState mutator.

packages/plugin-ext/src/common/test-types.ts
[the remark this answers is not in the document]
tsmaeder: I was more thinking of making the argument transformer method recursive: we are already detecting if it's an array, but the assumption is that the contents can only be TestItemReferences

tsmaeder:
When testing, I never get anything in the info popup for the message context except: `Message Context with args: [object Object]`. In instance, I got a stack overflow:
```
4rpc-message-encoder.ts:151 Uncaught (in promise) Error: Error during encoding: 'Maximum call stack size exceeded'
    at MsgPackMessageEncoder.encode (rpc-message-encoder.ts:151:23)
    at MsgPackMessageEncoder.request (rpc-message-encoder.ts:137:14)
    at RpcProtocol.sendRequest (rpc-protocol.ts:161:22)
    at proxy-handler.ts:74:45
```
This happened when doing this:

1. left-click "sample-child 1" in the Test Explorer view
2. **right-click** "sample-child 1" in the Test Run view and select "Message Context" from the context menu

Not sure whether that is due to the test plugin or the PR code.

packages/plugin-ext/src/plugin/tests.ts
```diff
@@ -335,6 +336,8 @@ export class TestingExtImpl implements TestingExt {
                     return this.toTestItem(arg);
                 } else if (Array.isArray(arg)) {
                     return arg.map(param => TestItemReference.is(param) ? this.toTestItem(param) : param);
```
tsmaeder, line 338: The argument processor is invoked for each argument we pass to the command, so if we pass the arguments `[testItemReference, testMessageDto]`, we just need to add code to transform TestMessageDto (if necessary).

packages/plugin-ext/src/main/browser/menus/plugin-menu-command-adapter.ts
```diff
@@ -230,6 +235,30 @@ export class PluginMenuCommandAdapter implements MenuCommandAdapter {
...
+        if (testMessage) {
+            const testItemReference = (testItem && testItem.controller) ? TestItemReference.create(testItem.controller.id, testItem.path) : undefined;
+            const testMessageDTO = {
+                message: testMessage.message,
+                actual: testMessage.actual,
+                expected: testMessage.expected,
+                contextValue: testMessage.contextValue,
+                location: testMessage.location ? fromLocation(testMessage.location) : undefined
+            };
+            return [TestMessageArg.create(testItemReference, testMessageDTO)];
```
tsmaeder, line 257: We can just return `[testItemReference, testMessageDTO]` and get rid of the TestMessageArgs namespace entirely. Just keep the code to handle TestMessageDTO

tsmaeder (changes requested):

packages/plugin-ext/src/main/browser/menus/plugin-menu-command-adapter.ts
rschnekenbu: We have to create a specific object for the commands on test message according to theia.d.ts:
The command will be called with an object containing:
         * - `test`: the {@link TestItem} the message is associated with, *if* it
         *    is still present in the {@link TestController.items} collection.
         * - `message`: the {@link TestMessage} instance.
So we still have to modify the argument processor on TestExtImpl to identify the 2 elements testItem and testMessage and create again this object. I thought it was easier to implement like it is currently. In the case we only pass the array  [testItemReference, testMessageDTO], we can't just map the elements of the array with toTestItem or toTestMessage, but we also need to return this new object. Or I may be missing an other processor that could do this on its own when invoking these specific commands.

rschnekenbu:
> When testing, I never get anything in the info popup for the message context except: `Message Context with args: [object Object]`. In instance, I got a stack overflow:
> 
> ```
> 4rpc-message-encoder.ts:151 Uncaught (in promise) Error: Error during encoding: 'Maximum call stack size exceeded'
>     at MsgPackMessageEncoder.encode (rpc-message-encoder.ts:151:23)
>     at MsgPackMessageEncoder.request (rpc-message-encoder.ts:137:14)
>     at RpcProtocol.sendRequest (rpc-protocol.ts:161:22)
>     at proxy-handler.ts:74:45
> ```
> 
> This happened when doing this:
> 
> 1. left-click "sample-child 1" in the Test Explorer view
> 2. **right-click** "sample-child 1" in the Test Run view and select "Message Context" from the context menu
> 
> Not sure whether that is due to the test plugin or the PR code.

I tried again, using the extension provided in the PR and with latest version, I do not get these stack error messages on sample-child1. I tested on both browser and electron version. I see the difference with your message information text, I will push a new version of the extension with pretty printed objects.

I can see however a call stack error while right clicking on the Test Run root object. I will fix that.

rschnekenbu:
I updated the PR with more checks on arguments processing. Element tree in Test Run widget can also be `TestRun`, which was not caught in argument processor and forwarded as such without cleaning.

Here is the version of the test extension with stringified argument:
zip: 
[test-extension-sample-0.0.3.zip](https://github.com/eclipse-theia/theia/files/13715118/test-extension-sample-0.0.3.zip)
src:
[test-extension-sample-0.0.3-src.zip](https://github.com/eclipse-theia/theia/files/13715119/test-extension-sample-0.0.3-src.zip)

packages/plugin-ext/src/main/browser/menus/plugin-menu-command-adapter.ts
tsmaeder: You are right, I did not notice that in the doc. How weird 🤷

rschnekenbu: Yes, I totally agree with the weird arguments :)

tsmaeder:
With the latest plugin, I always get "Message Context with args: undefined" when I click the context menu. What gives?

rschnekenbu:
> With the latest plugin, I always get "Message Context with args: undefined" when I click the context menu. What gives?

Sorry for that, @tsmaeder! I am pretty sure I had check thoroughly the changes :/ In certain cases, TestMessage.is() was returning false despite the element passed being one. I updated the PR so this should not happen anymore. 
In the case the test items are not in failure or the selected element in the tree is not a TestItem, the command shall return  the kind of message that you have, as the expected arguments cannot be computed. In fact, TestMessage is only available in case of test failures. I assume there it is up to the extension to check the enablement of the command.

tsmaeder (approved):
I think your original code was correct as far as the typing goes: the "location" field in TestMessage should probably be optional. But that's another problem. Lgtm now.

rschnekenbu:
Thanks a lot, @tsmaeder! I will rebase and resolve conflicts then.

JonasHelming:
@tsmaeder Is this approved?

PR #13178, opened 2023-12-15
fix: revise vscode extension directories

### LINKED ISSUE #12757: Plugin cache prevents vscode extension install
I got into a state where I cannot install the gitlens extension. Here's what I do:

1. I go to the plugins view and search for "gitlens". The exension appears in the results
2. I press the "install" button on the search result
3. Nothing happens: the extension does not activate. Restarting does not help either

xai:
#### What it does

This patch eliminates the use of the temporary directories `vscode-unpacked` and `vscode-copied` for installing and deploying vscode extensions. The file-handler and directory handler for vscode extensions have been adjusted accordingly:

1. A temporary directory is now only used for downloading extensions from the registry. This temporary directory is now user-specific and resides within the configdir (i.e., per default in the user's home: <userhome>/.theia/tmp/) to avoid clashes and permission issues on multi-user operating systems that share temporary directories, such as Linux or BSDs. Having this temporary directory in a location that is configurable by the user also seems the more sensible approach when extensions are considered confidential data.
2. `$configDir/deployedPlugins` replaces our volatile `/tmp/vscode-copied` deployment directory. Having a more permanent way of handling installed extensions should improve startup time and reduce issues with multiple instances running in parallel.
3. The local file resolver unpacks the vsix file from the temp dir into `$configDir/deployedPlugins/<extension-id>`.
4. The simplified directory handler loads unpacked extensions directly from `$configDir/deployedPlugins/<extension-id>`.
5. We use `$configDir/extensions` as a location for the user to drop vsix files that will be installed to the deployment location automatically on startup. We do not manage or remove files within `$configDir/extensions`.

Overall, this should improve the stability on systems with shared temp dir locations and reduce the startup of the first application start after a reboot.

Fixes #12757

#### How to test

Installing from registry

1. Start Theia
2. Use the vsix registry to install an extension
3. Verify that the extension is loaded and its functionality is enabled.
4. Verify that the extension is permanently stored unpacked in $configDir/deployedPlugins/<extension-id>.
5. Try starting Theia again to verify the extension gets correctly loaded
6. Remove the extension (this should indicate that a reload of the window is required), and verify that the extension is no longer active
7. Verify that $configDir/extensions/<extension-id> has been removed

Installing a local vsix extensions

1. Start Theia
2. Select `Install from VSIX` in the triple-dot menu of the extensions view
3. Browse to a local .vsix file and install ut
4. Verify that the extension is permanently stored unpacked in $configDir/deployedPlugins/<extension-id>.
5. Remove the extension (this should indicate that a reload of the window is required), and verify that the extension is no longer active
6. Verify that $configDir/extensions/<extension-id> has been removed

Store a vsix file in the drop-in location

1. Put a .vsix file into $configDir/extensions
2. Start Theia
3. Verify that the extension is permanently stored unpacked in $configDir/deployedPlugins/<extension-id>.
4. Removing the extension will cause it to be reinstalled until you remove it from the drop-in location

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext-vscode/src/common/plugin-vscode-environment.ts
+++ b/packages/plugin-ext-vscode/src/common/plugin-vscode-environment.ts
@@ -24,13 +24,36 @@ export class PluginVSCodeEnvironment {
     @inject(EnvVariablesServer)
     protected readonly environments: EnvVariablesServer;
 
-    protected _extensionsDirUri: URI | undefined;
-    async getExtensionsDirUri(): Promise<URI> {
-        if (!this._extensionsDirUri) {
+    protected _userExtensionsDirUri: URI | undefined;
+    protected _deployedPluginsUri: URI | undefined;
+    protected _tmpDirUri: URI | undefined;
+
+    async getUserExtensionsDirUri(): Promise<URI> {
+        if (!this._userExtensionsDirUri) {
             const configDir = new URI(await this.environments.getConfigDirUri());
-            this._extensionsDirUri = configDir.resolve('extensions');
+            this._userExtensionsDirUri = configDir.resolve('extensions');
         }
-        return this._extensionsDirUri;
+        return this._userExtensionsDirUri;
     }
 
+    async getDeploymentDirUri(): Promise<URI> {
+        if (!this._deployedPluginsUri) {
+            const configDir = new URI(await this.environments.getConfigDirUri());
+            this._deployedPluginsUri = configDir.resolve('deployedPlugins');
+        }
+        return this._deployedPluginsUri;
+    }
+
+    async getTempDirUri(prefix?: string): Promise<URI> {
+        if (!this._tmpDirUri) {
+            const configDir: URI = new URI(await this.environments.getConfigDirUri());
+            this._tmpDirUri = configDir.resolve('tmp');
+        }
+
+        if (prefix) {
+            return this._tmpDirUri.resolve(prefix);
+        }
+
+        return this._tmpDirUri;
+    }
 }
--- a/packages/plugin-ext-vscode/src/node/local-vsix-file-plugin-deployer-resolver.ts
+++ b/packages/plugin-ext-vscode/src/node/local-vsix-file-plugin-deployer-resolver.ts
@@ -14,18 +14,18 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import * as fs from '@theia/core/shared/fs-extra';
 import * as path from 'path';
 import { inject, injectable } from '@theia/core/shared/inversify';
-import { FileUri } from '@theia/core/lib/node';
 import { PluginDeployerResolverContext } from '@theia/plugin-ext';
 import { LocalPluginDeployerResolver } from '@theia/plugin-ext/lib/main/node/resolvers/local-plugin-deployer-resolver';
 import { PluginVSCodeEnvironment } from '../common/plugin-vscode-environment';
 import { isVSCodePluginFile } from './plugin-vscode-file-handler';
+import { existsInDeploymentDir, unpackToDeploymentDir } from './plugin-vscode-utils';
 
 @injectable()
 export class LocalVSIXFilePluginDeployerResolver extends LocalPluginDeployerResolver {
     static LOCAL_FILE = 'local-file';
+    static FILE_EXTENSION = '.vsix';
 
     @inject(PluginVSCodeEnvironment) protected readonly environment: PluginVSCodeEnvironment;
 
@@ -38,28 +38,14 @@ export class LocalVSIXFilePluginDeployerResolver extends LocalPluginDeployerReso
     }
 
     async resolveFromLocalPath(pluginResolverContext: PluginDeployerResolverContext, localPath: string): Promise<void> {
-        const fileName = path.basename(localPath);
-        const pathInUserExtensionsDirectory = await this.ensureDiscoverability(localPath);
-        pluginResolverContext.addPlugin(fileName, pathInUserExtensionsDirectory);
-    }
+        const extensionId = path.basename(localPath, LocalVSIXFilePluginDeployerResolver.FILE_EXTENSION);
 
-    /**
-     * Ensures that a user-installed plugin file is transferred to the user extension folder.
-     */
-    protected async ensureDiscoverability(localPath: string): Promise<string> {
-        const userExtensionsDir = await this.environment.getExtensionsDirUri();
-        if (!userExtensionsDir.isEqualOrParent(FileUri.create(localPath))) {
-            try {
-                const newPath = FileUri.fsPath(userExtensionsDir.resolve(path.basename(localPath)));
-                await fs.mkdirp(FileUri.fsPath(userExtensionsDir));
-                await new Promise<void>((resolve, reject) => {
-                    fs.copyFile(localPath, newPath, error => error ? reject(error) : resolve());
-                });
-                return newPath;
-            } catch (e) {
-                console.warn(`Problem copying plugin at ${localPath}:`, e);
-            }
+        if (await existsInDeploymentDir(this.environment, extensionId)) {
+            console.log(`[${pluginResolverContext.getOriginId()}]: Target dir already exists in plugin deployment dir`);
+            return;
         }
-        return localPath;
+
+        const extensionDeploymentDir = await unpackToDeploymentDir(this.environment, localPath, extensionId);
+        pluginResolverContext.addPlugin(extensionId, extensionDeploymentDir);
     }
 }
--- a/packages/plugin-ext-vscode/src/node/plugin-vscode-deployer-participant.ts
+++ b/packages/plugin-ext-vscode/src/node/plugin-vscode-deployer-participant.ts
@@ -15,8 +15,11 @@
 // *****************************************************************************
 
 import { injectable, inject } from '@theia/core/shared/inversify';
+import * as fs from '@theia/core/shared/fs-extra';
+import { FileUri } from '@theia/core/lib/node';
 import { PluginVSCodeEnvironment } from '../common/plugin-vscode-environment';
 import { PluginDeployerParticipant, PluginDeployerStartContext } from '@theia/plugin-ext/lib/common/plugin-protocol';
+import { LocalVSIXFilePluginDeployerResolver } from './local-vsix-file-plugin-deployer-resolver';
 
 @injectable()
 export class PluginVSCodeDeployerParticipant implements PluginDeployerParticipant {
@@ -25,8 +28,21 @@ export class PluginVSCodeDeployerParticipant implements PluginDeployerParticipan
     protected readonly environments: PluginVSCodeEnvironment;
 
     async onWillStart(context: PluginDeployerStartContext): Promise<void> {
-        const extensionsDirUri = await this.environments.getExtensionsDirUri();
-        context.userEntries.push(extensionsDirUri.withScheme('local-dir').toString());
-    }
+        const extensionDeploymentDirUri = await this.environments.getDeploymentDirUri();
+        context.userEntries.push(extensionDeploymentDirUri.withScheme('local-dir').toString());
 
+        const userExtensionDirUri = await this.environments.getUserExtensionsDirUri();
+        const userExtensionDirPath = FileUri.fsPath(userExtensionDirUri);
+
+        if (await fs.pathExists(userExtensionDirPath)) {
+            const files = await fs.readdir(userExtensionDirPath);
+            for (const file of files) {
+                if (file.endsWith(LocalVSIXFilePluginDeployerResolver.FILE_EXTENSION)) {
+                    const extensionUri = userExtensionDirUri.resolve(file).withScheme('local-file').toString();
+                    console.log(`found drop-in extension "${extensionUri}"`);
+                    context.userEntries.push(extensionUri);
+                }
+            }
+        }
+    }
 }
--- a/packages/plugin-ext-vscode/src/node/plugin-vscode-file-handler.ts
+++ b/packages/plugin-ext-vscode/src/node/plugin-vscode-file-handler.ts
@@ -14,32 +14,20 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { PluginDeployerFileHandler, PluginDeployerEntry, PluginDeployerFileHandlerContext, PluginType } from '@theia/plugin-ext';
-import * as fs from '@theia/core/shared/fs-extra';
-import * as path from 'path';
+import { PluginDeployerFileHandler, PluginDeployerEntry, PluginDeployerFileHandlerContext } from '@theia/plugin-ext';
 import * as filenamify from 'filenamify';
-import type { URI } from '@theia/core';
 import { inject, injectable } from '@theia/core/shared/inversify';
-import { Deferred } from '@theia/core/lib/common/promise-util';
-import { getTempDirPathAsync } from '@theia/plugin-ext/lib/main/node/temp-dir-util';
+import * as fs from '@theia/core/shared/fs-extra';
+import { FileUri } from '@theia/core/lib/node';
 import { PluginVSCodeEnvironment } from '../common/plugin-vscode-environment';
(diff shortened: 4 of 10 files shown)


tsmaeder:
I have a couple of general comments:

1. We should not download extensions to a shared directory. I have heard of use cases where even the existence of certain extensions may be a trade secret. Isn't there a user-specific tmp on linux?
2. We should never move or delete files that are managed by the user. So if the user puts a *.vsix into `.theia/extensions`, we should not move nor delete it from there.
3. If we download *.vsix files as a side-effect of an installation, we should clean up those files once we no longer need them. Relying on the OS or user to clean up files created by software is bad practice, IMO.
4. I think it would be a good idea to separate "drop in locations" and "deployment locations". Any given directory should be one or the other. If an application does not have the ability to install additional plugins, we should never deal with drop-in locations. Concretely, this would mean that we have `.theia/extensions` as a drop-in location where the user can put extensions and `.theia/deployedPlugins` where stuff is extracted to.

xai:
Thanks, @tsmaeder, for the quick response!

>     1. We should not download extensions to a shared directory. I have heard of use cases where even the existence of certain extensions may be a trade secret. Isn't there a user-specific tmp on linux?

I don't think there is a consistent user-specific temp dir in Linux. According to `man 3 tmpfile`:
``` 
The  standard  does  not  specify  the  directory  that  tmpfile()  will  use.  Glibc will try the path prefix P_tmpdir defined in <stdio.h>, and if that fails the directory /tmp.
```
The shared /tmp is often also on a different filesystem/mountpoint than the user home. In several scenarios, the user home will be better secured (e.g., encrypted with user-specific credentials, mounted after kerberos auth, ...) than the shared temp dir location.

In our use case, the download directory is meant to be user-specific (that's why I encoded the username into it), but sure, the default permissions will most likely be world-readable (755) if we do not change them. So I see two options:
1. Make sure the permission of /tmp/vscode-download-$username is 700
2. Download to a path within the user home, such as .theia/tmp/download

I think if we want to treat extensions as potential secrets, then the second approach is more sensible, considering the user home to be a much safer place than /tmp. Of course, this not being a temp dir that the OS handles, we have to clean it up ourselves.
So what do you think about `<userhome>/.theia/tmp/download-<randomuuid/processid>`?

>     2. We should never move or delete files that are managed by the user. So if the user puts a *.vsix into `.theia/extensions`, we should not move nor delete it from there.

ACK.

>     3. If we download *.vsix files as a side-effect of an installation, we should clean up those files once we no longer need them. Relying on the OS or user to clean up files created by software is bad practice, IMO.

That makes sense, especially if we decide to use a location within the user's home. We have to be careful here with regard to robustness in case of multiple Theia processes (run by the same user). So, cleaning up after successful unpacking is easy enough. What do we do if the application gets killed hard by the user or the OOM killer and our cleanup is not finished? Should we try to clean at the application start? Then we have to make sure not to remove stuff that another running Theia process is currently using.

>     4. I think it would be a good idea to separate "drop in locations" and "deployment locations". Any given directory should be one or the other. If an application does not have the ability to install additional plugins, we should never deal with drop-in locations. Concretely, this would mean that we have `.theia/extensions` as a drop-in location where the user can put extensions and `.theia/deployedPlugins` where stuff is extracted to.

Makes sense; this is backward-compatible and avoids unwanted interactions between our unpacking/deployment and the user. I'll revise the code accordingly.

xai:
I revised my PR to use

* `~/.theia/deployedPlugins/` as a permanent location for installed extensions. This replaces the volatile `/tmp/vscode-copied`
* `~/.theia/extensions/` as a drop-in location that the user manages (we do not touch anything there) and that only serves as a source of extensions that will be installed on startup
* `~/.theia/tmp/` as a replacement for the remaining shared temp directories. It can be used to download .vsix files from the registry (they do not belong in ~/.theia/extensions`, as we no longer manage this directory).

packages/plugin-ext-vscode/src/node/plugin-vscode-decompress.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import * as decompress from 'decompress';
+import * as path from 'path';
+import * as fs from '@theia/core/shared/fs-extra';
+
+export async function decompressExtension(sourcePath: string, destPath: string): Promise<boolean> {
+    try {
+        await decompress(sourcePath, destPath);
+        if (sourcePath.endsWith('.tgz')) {
+            const extensionPath = path.join(destPath, 'package');
```
tsmaeder, line 25: A comment to explain this path mangling would be nice

packages/plugin-ext-vscode/src/node/plugin-vscode-deployer-participant.ts
```diff
@@ -25,8 +27,21 @@ export class PluginVSCodeDeployerParticipant implements PluginDeployerParticipan
     protected readonly environments: PluginVSCodeEnvironment;
 
     async onWillStart(context: PluginDeployerStartContext): Promise<void> {
-        const extensionsDirUri = await this.environments.getExtensionsDirUri();
-        context.userEntries.push(extensionsDirUri.withScheme('local-dir').toString());
+        const extensionDeploymentDirUri = await this.environments.getDeploymentDirUri();
+        context.userEntries.push(extensionDeploymentDirUri.withScheme('local-dir').toString());
```
tsmaeder, line 31: Is this releated to the `--plugins=local-dir:../../plugins` cli parameter? Since we separated "drop-in" directories from "deployment" directories, what does the cli-parameter now mean?

packages/plugin-ext-vscode/src/node/plugin-vscode-decompress.ts
```diff
@@ -0,0 +1,36 @@
...
+            const vscodeNodeModulesPath = path.join(extensionPath, 'vscode_node_modules.zip');
+            if (await fs.pathExists(vscodeNodeModulesPath)) {
+                await decompress(vscodeNodeModulesPath, path.join(extensionPath, 'node_modules'));
+            }
+        }
+        return true;
+    } catch (error) {
+        console.error(`Failed to decompress ${sourcePath} to ${destPath}: ${error}`);
```
tsmaeder, line 33: Why are we not just letting the exception propagate?

packages/plugin-ext-vscode/src/node/plugin-vscode-file-handler.ts
```diff
@@ -50,34 +39,45 @@ export class PluginVsCodeFileHandler implements PluginDeployerFileHandler {
...
+            }
+
+            if (!await decompressExtension(context.pluginEntry().path(), extensionDeploymentDir)) {
+                console.error(`[${id}]: decompression failed`);
+                return;
+            }
+
+            console.log(`[${id}]: decompressed`);
+        } catch (error) {
+            console.error(`[${id}]: error while decompressing`, error);
```
tsmaeder, line 65: Why not let the error propagate? How is the invoker going to tell sometihing failed?

packages/plugin-ext-vscode/src/node/plugin-vscode-file-handler.ts
```diff
@@ -50,34 +39,45 @@ export class PluginVsCodeFileHandler implements PluginDeployerFileHandler {
...
-            if (await fs.pathExists(vscodeNodeModulesPath)) {
-                await context.unzip(vscodeNodeModulesPath, path.join(extensionPath, 'node_modules'));
-            }
+    protected async getExtensionDeploymentDir(pluginId: string): Promise<string> {
+        try {
+            const deployedPluginsDirUri = await this.environment.getDeploymentDirUri();
+            const extensionDeploymentDirPath = FileUri.fsPath(deployedPluginsDirUri.resolve(pluginId));
+            return extensionDeploymentDirPath;
+        } catch (error) {
+            console.error('Error resolving extension directory:', error);
```
tsmaeder, line 79: If the error is unhandled, we'll get a stack trace. If it isn't shouldn't the handler print the error?

packages/vsx-registry/src/node/vsx-extension-resolver.ts
```diff
@@ -74,16 +75,24 @@ export class VSXExtensionResolver implements PluginDeployerResolver {
                 return;
             }
         }
-        const downloadPath = (await this.environment.getExtensionsDirUri()).path.fsPath();
-        await fs.ensureDir(downloadPath);
-        const extensionPath = path.resolve(downloadPath, path.basename(downloadUrl));
-        console.log(`[${resolvedId}]: trying to download from "${downloadUrl}"...`, 'to path', downloadPath);
-        if (!await this.download(downloadUrl, extensionPath)) {
+        const downloadDir = await this.getTempDir();
+        await fs.ensureDir(downloadDir);
```
tsmaeder, line 79: Do we clean this up somewhere?

tsmaeder (changes requested):
Mostly comments, but I think we should fix the "move" logic.

tsmaeder:
Some notes from testing:

1. Dropping the "vscode-drawio" plugin into the "extensions" folder inside the `~/.theia` folder blocks the UI at startup for 20 seconds with a white screen. When we install extensions via the UI, the extensions can startup after being installed, so there's no reason to block while we wait for the plugins to install. 
This happens on **every** startup, not just the first. 
3. After startup, we end up with an "uninstall" button for the extesnsion in the plugins view. When you click it, the extension is uninstalled. However, the extension is reinstalled the next time we start Theia. Maybe we should consider extensions installed via "drop-in" folder as "built-in" and the way to remove them would be to remove the *.vsix from the drop-in folder?

I can kind of live with the second behavior, but Nr. 1 is a must-fix, IMO.

packages/plugin-ext-vscode/src/node/plugin-vscode-decompress.ts
xai: I just moved this part of code over there from https://github.com/eclipse-theia/theia/blob/2d30b295c1158f401d63591eb67b46e93cd6f318/packages/plugin-ext-vscode/src/node/plugin-vscode-file-handler.ts#L76

I had no idea for which exact use case we needed this, but I did not want to create regressions, so I kept it. Comes from [this commit](https://github.com/eclipse-theia/theia/commit/ccaa5d84c5718ce66e887b7bed620366fc859c11) originally. It seems that it unpacks an additional `vscode_node_modules.zip` file in case the source file was a .tgz. I can add a comment that says so, but I do not know the exact use case.

packages/plugin-ext-vscode/src/node/local-vsix-file-plugin-deployer-resolver.ts
```diff
@@ -38,28 +40,70 @@ export class LocalVSIXFilePluginDeployerResolver extends LocalPluginDeployerReso
...
+    protected async getPluginId(extensionDir: string): Promise<string> {
+        const packageJsonPath = path.join(extensionDir, 'extension', 'package.json');
+        const packageJson = await fs.readJson(packageJsonPath);
+        return `${packageJson.publisher}.${packageJson.name}-${packageJson.version}`;
+    }
+
+    protected async getExtensionDeploymentDir(pluginId: string): Promise<string> {
+        try {
+            const deployedPluginsDirUri = await this.environment.getDeploymentDirUri();
+            const normalizedExtensionId = filenamify(pluginId, { replacement: '_' });
```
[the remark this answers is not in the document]
xai: "Should not happen" ™, as the `pluginId` that is passed in there is parsed from the package.json after unpacking. This is btw. the reason why the application is starting slow when drop-in vsix files are used currently. We always unpack to parse the plugin-id from the package.json. This is a problem ...

packages/plugin-ext-vscode/src/node/plugin-vscode-deployer-participant.ts
xai: I don't think this part of the code ever cared about `--plugins`:
https://github.com/eclipse-theia/theia/blob/2d30b295c1158f401d63591eb67b46e93cd6f318/packages/plugin-ext-vscode/src/common/plugin-vscode-environment.ts#L31
Maybe this is a difference between theia plugins and vscode extensions? For vscode extensions, it looks already hard-coded to me.

packages/plugin-ext-vscode/src/node/plugin-vscode-decompress.ts
xai: Sure, stupid mistake.

packages/plugin-ext-vscode/src/common/plugin-vscode-environment.ts
```diff
@@ -24,13 +24,32 @@ export class PluginVSCodeEnvironment {
...
+            this._deployedPluginsUri = configDir.resolve('deployedPlugins');
+        }
+        return this._deployedPluginsUri;
+    }
+
+    async getTempDirUri(prefix?: string): Promise<URI> {
+        if (!this._tmpDirUri) {
+            const configDir: URI = new URI(await this.environments.getConfigDirUri());
+            const processId: string = process.pid.toString();
+            this._tmpDirUri = configDir.resolve('tmp').resolve(`${prefix ?? 'tmp-'}-${processId}`);
```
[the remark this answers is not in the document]
xai: If I understood this correctly, this will use the operating system's temp dir mechanism, that we considered unsafer than a location within the user's conf dir.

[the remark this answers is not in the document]
xai: I'll try.

packages/vsx-registry/src/node/vsx-extension-resolver.ts
xai: Now we do ;)
I adjusted the `PluginVsCodeFileHandler` to cleanup `pluginEntry().originalPath()` if it is a temp dir.

xai:
>     1. Dropping the "vscode-drawio" plugin into the "extensions" folder inside the `~/.theia` folder blocks the UI at startup for 20 seconds with a white screen. When we install extensions via the UI, the extensions can startup after being installed, so there's no reason to block while we wait for the plugins to install.
>        This happens on **every** startup, not just the first.

It should be fixed now. I reverted back to using the filename of a dropped vsix file as extension-id instead of *always* unpacking and parsing package.json. Would be nicer to have the correct extension id, but we can deal with this in a future PR, I'd be happy to look into this.


>     2. After startup, we end up with an "uninstall" button for the extesnsion in the plugins view. When you click it, the extension is uninstalled. However, the extension is reinstalled the next time we start Theia. Maybe we should consider extensions installed via "drop-in" folder as "built-in" and the way to remove them would be to remove the *.vsix from the drop-in folder?

Treating them as "built-in" would be in line with the user being the only one who manages them. This would probably also fit well with having a lookup mechanism for extensions because then we could track the source of deployed extensions (e.g., downloaded, dropped-in, installed-from-file) and treat them accordingly. 

> I can kind of live with the second behavior, but Nr. 1 is a must-fix, IMO.

Nr. 1 should be done now. Nr. 2 is definitely a good candidate for a follow-up ticket.

packages/plugin-ext-vscode/src/node/local-vsix-file-plugin-deployer-resolver.ts
```diff
@@ -38,28 +40,70 @@ export class LocalVSIXFilePluginDeployerResolver extends LocalPluginDeployerReso
...
+        const extensionDeploymentDir = await this.getExtensionDeploymentDir(pluginId);
+        if (await fs.pathExists(extensionDeploymentDir)) {
+            console.log(`[${pluginResolverContext.getOriginId()}]: Target dir already exists in plugin deployment dir: "${extensionDeploymentDir}"`);
+            return;
+        }
+        console.log(`[${pluginResolverContext.getOriginId()}]: moving to extension dir "${extensionDeploymentDir}"...`);
+        await fs.move(tempUnpackDir, extensionDeploymentDir, { overwrite: true });
+
+        // if temp dir is empty, remove it
+        if (fs.readdirSync(tempDir).length === 0) {
```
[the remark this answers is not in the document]
xai: Yes, that was my thought here. I removed all Sync operations now and hope that the error handling is sufficient.

tsmaeder (commented):
Code looks good to me now, but I still get a 20s pause with a blank window the first time I drop the draw.io vsix into the `.theia/extensions` folder.

tsmaeder (approved):

tsmaeder:
@xai we have a conflict.

xai:
> @xai we have a conflict.

Resolved.

PR #13179, opened 2023-12-15
Make tree widget indentation configurable

AlexandraBuzila:
#### What it does
Make the indentation of the tree widget configurable via preference.

#### How to test
1. Open the settings view from File → Preferences → Settings
2. Search for `workbench.tree.indent`
3.  Change the `Workbench › Tree: Indent` preference and notice how the indentation of the tree widget, e.g. in the workspace navigator view, changes accordingly

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/tree/tree-preference.ts
+++ b/packages/core/src/browser/tree/tree-preference.ts
@@ -1,0 +1,50 @@
+// *****************************************************************************
+// Copyright (C) 2024 STMicroelectronics and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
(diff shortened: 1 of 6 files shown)


msujew (commented):
I don't think this change is working as expected/as it should.

1. Indent pinning doesn't work, see below.
2. Increasing the indent also increases the indent of the root tree items. This isn't the case in vscode. In vscode, the root tree items always stay in the same spot. This indicates that we still need the `leftPadding` property that tells us how much the root tree items are padded, while adding a `indentPadding` that tells us how much each indented tree item needs to be padded.

By fixing 2. you should also be able to fix 1. as a drive by.

AlexandraBuzila:
Thank you for the review, @msujew 

I updated the code and restored the `leftPadding` prop: it now controls the padding for the tree root, while the new preference only affects the indentation of the nodes below it.

msujew (commented):
Looks mostly good to me, just one more issue see below. This should be good to go afterwards.

packages/core/src/browser/tree/tree-preference.ts
```diff
@@ -0,0 +1,49 @@
...
+import { PreferenceProxy, PreferenceContribution, PreferenceSchema } from '../preferences';
+import { PreferenceProxyFactory } from '../preferences/injectable-preference-proxy';
+
+export const PREFERENCE_NAME_TREE_INDENT = 'workbench.tree.indent';
+
+export const treePreferencesSchema: PreferenceSchema = {
+    type: 'object',
+    properties: {
+        [PREFERENCE_NAME_TREE_INDENT]: {
+            description: 'Controls tree indentation in pixels',
```
msujew, line 27: Minor: This is also missing its localization.
```suggestion
            description: nls.localizeByDefault('Controls tree indentation in pixels'),
```

msujew (approved):
Alright, looks good to me 👍

PR #13180, opened 2023-12-18
[RPC] Delay sending messages when there are handlers running initialization

### LINKED ISSUE #13158: 'unknown document' error is thrown without mentioning what document is unknown
### Bug Description:
There are two places in [document.ts](https://github.com/eclipse-theia/theia/blob/273c7e269015c3aa8328f364a1475a96dcd625c6/packages/plugin-ext/src/plugin/documents.ts#L161) where an `Error` is thrown saying a document is unknown, but the code isn't including the document URI string in the error. It sure would be helpful if it did.

### Steps to Reproduce:
None needed. Self evident. The error condition isn't what's being reported here; I'm pointing out a deficiency in how the error is reported.

### LINKED ISSUE #13172: RPC ClientProxyHandler lazy init may lead to problems when proxies have dependencies on each other
### Bug Description:
Currently this bug is visible as this error message on the browser console of the electron example application:

```
Uncaught (in promise) Error: unknown document
    at DocumentsExtImpl.$acceptDirtyStateChanged (~/git/theia/examples/electron/lib/backend/plugin-host.js:5333:19)
    at ~/git/theia/examples/electron/lib/backend/packages_plugin-ext_lib_common_plugin-api-rpc_js.js:2060:71
    at process.processTicksAndRejections (VM33 task_queues:95:5)
    at async RpcProtocol.handleRequest (~/git/theia/examples/electron/lib/backend/packages_core_lib_common_index_js-node_modules_vscode-languageserver-types_lib_umd_sync_recursive.js:3576:28)
```

### Steps to Reproduce:

1. Build and run electron example with downloaded plugins `yarn && yarn build && yarn download:plugins && yarn electron start`
2. Open Debugger Tools in Electron and check the console
3. You should see a similar `unknown document` exception

### Additional Information

At the location where the error is thrown we have this code:
```
const data = this.editorsAndDocuments.getDocument(uriString);
if (!data) {
    throw new Error('unknown document');
}
```
Here `DocumentsExtImpl` accesses information from `EditorsAndDocumentsExtImpl` (so we have a dependency between those two). 

The error chain is triggered like this (as far as I could see):

The `electron-menu-contribution` calls `this.preferenceService.set('window.titleBarStyle', this.titleBarStyle, PreferenceScope.User);`

This will open `user-storage:/user/settings.json`, write to it, and close it again via the `preference-transaction-manager`. This is happening very fast. 

On the browser side `DocumentsMainImpl` and `EditorsAndDocumentsMain` are both reacting to this and forward to their proxies in the expected order:

`EditorsAndDocumentsMain` calls
`this.proxy.$acceptEditorsAndDocumentsDelta(deltaExt);`
because there is a new document (`user-storage:/user/settings.json`). 

`DocumentsMainImpl` calls
`this.proxy.$acceptDirtyStateChanged(m.textEditorModel.uri, m.dirty);`
because `user-storage:/user/settings.json` was touched. 

`EditorsAndDocumentsMain` calls
`this.proxy.$acceptEditorsAndDocumentsDelta(deltaExt);`
again, because the document is closed again. 

When we debug the other side of the communication (`DocumentsExtImpl` and `EditorsAndDocumentsExtImpl`) we can see that `$acceptDirtyStateChanged` arrives after the second `$acceptEditorsAndDocumentsDelta` call, which leads to above error. The document is already removed from `editorsAndDocuments`.

I think the issue is that the `ClientProxyHandler` implementation has lazy initialization code: https://github.com/eclipse-theia/theia/blob/469bd744a02e5dcdbcbc338af178d410dd909616/packages/plugin-ext/src/common/proxy-handler.ts#L50

While the `ClientProxyHandler` for editors-and-documents is already initialized (because it has been called before already) the documents handler is initialized on the first `acceptEditorsAndDocumentsDelta` call. 
This allows the second call to `acceptEditorsAndDocumentsDelta` to "overtake" the first one, leading to the exception. 

Hacking the `ClientProxyHandler` to call `initializeRpc()` in its constructor (i.e. basically removing the lazy init) removes the exception, but I guess this is not the desired solution. 

So I think we need a mechanism to sync the initialization of `ClientProxyHandler`s when they have dependencies on each other. 
This could e.g. be 
* a init listener mechanism + a way to manually initialize handlers , or
* to provide information about dependencies to the framework, so that the framework can init all of them

I've played a bit with the listener mechanism and could open a PR if desired to kick off a discussion about the best solution. 

- Operating System: Linux
- Theia Version: 1.44

jfaltermeier:
#### What it does

With this PR `ClientProxyHandler`s will report their init state. When there is a running initialization, we delay sending messages on all handlers until this initialization is done. 
This should prevent cases where messages sent later via an already initialized handler can overtake earlier messages that are waiting for their handler's initialization to finish. 

This will help avoiding race condition like in #13172

Fixes #13158
Fixes #13172

Commits should be squashed when closing the PR

#### How to test

* Check if the bug described in #13172 does not happen any more. 
* Might have an impact on performance with messages being delayed. Does this cause issue anywhere?

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/common/proxy-handler.ts
+++ b/packages/plugin-ext/src/common/proxy-handler.ts
@@ -78,14 +78,18 @@ export class ClientProxyHandler<T extends object> implements ProxyHandler<T> {
         const isNotify = this.isNotification(name);
         return (...args: any[]) => {
             const method = name.toString();
-            return this.proxySynchronizer.pendingProxyInitializations().then(() => this.rpcDeferred.promise.then(async (connection: RpcProtocol) => {
+            return this.sendWhenNoInit(async (connection: RpcProtocol) => {
                 if (isNotify) {
                     connection.sendNotification(method, args);
                 } else {
                     return await connection.sendRequest(method, args) as Promise<any>;
                 }
-            }));
+            });
         };
+    }
+
+    private sendWhenNoInit(send: (connection: RpcProtocol) => Promise<any>): Promise<any> {
+        return this.proxySynchronizer.pendingProxyInitializations().then(() => this.rpcDeferred.promise.then(send));
     }
 
     /**
--- a/packages/plugin-ext/src/common/proxy-handler.ts
+++ b/packages/plugin-ext/src/common/proxy-handler.ts
@@ -25,17 +25,16 @@ export interface RpcHandlerOptions {
 }
 export interface ProxyHandlerOptions extends RpcHandlerOptions {
     channelProvider: () => Promise<Channel>,
-    initCallback: InitializationCallback,
+    proxySynchronizer: ProxySynchronizer,
 }
 
 export interface InvocationHandlerOptions extends RpcHandlerOptions {
     target: any
 }
 
-export interface InitializationCallback {
-    reportInit(id: string): void
-    reportInitDone(id: string): void
-    checkInit(): Promise<void>
+export interface ProxySynchronizer {
+    startProxyInitialization(id: string, init: Promise<void>): void
+    pendingProxyInitializations(): Promise<void>
 }
 
 /**
@@ -48,7 +47,7 @@ export class ClientProxyHandler<T extends object> implements ProxyHandler<T> {
 
     readonly id: string;
     private readonly channelProvider: () => Promise<Channel>;
-    private readonly initCallback: InitializationCallback;
+    private readonly proxySynchronizer: ProxySynchronizer;
     private readonly encoder: RpcMessageEncoder;
     private readonly decoder: RpcMessageDecoder;
 
@@ -59,13 +58,12 @@ export class ClientProxyHandler<T extends object> implements ProxyHandler<T> {
     private initializeRpc(): void {
         // we need to set the flag to true before waiting for the channel provider. Otherwise `get` might
         // get called again and we'll try to open a channel more than once
-        this.initCallback.reportInit(this.id);
+        this.proxySynchronizer.startProxyInitialization(this.id, this.rpcDeferred.promise.then(() => { }));
         this.isRpcInitialized = true;
         const clientOptions: RpcProtocolOptions = { encoder: this.encoder, decoder: this.decoder, mode: 'clientOnly' };
         this.channelProvider().then(channel => {
             const rpc = new RpcProtocol(channel, undefined, clientOptions);
             this.rpcDeferred.resolve(rpc);
-            this.initCallback.reportInitDone(this.id);
         });
     }
 
@@ -80,7 +78,7 @@ export class ClientProxyHandler<T extends object> implements ProxyHandler<T> {
         const isNotify = this.isNotification(name);
         return (...args: any[]) => {
             const method = name.toString();
-            return this.initCallback.checkInit().then(() => this.rpcDeferred.promise.then(async (connection: RpcProtocol) => {
+            return this.proxySynchronizer.pendingProxyInitializations().then(() => this.rpcDeferred.promise.then(async (connection: RpcProtocol) => {
                 if (isNotify) {
                     connection.sendNotification(method, args);
                 } else {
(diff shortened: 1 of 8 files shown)


jfaltermeier:
@martin-fleck-at Could you please have another look? I've added an init callback that delays already initialized handlers from sending their messages immediately while an init is going on. As far as I could see, this should sync the messages/events as expected.

martin-fleck-at:
@jfaltermeier Sure, I'll have another look at it tomorrow.

packages/plugin-ext/src/common/proxy-handler.ts
```diff
@@ -50,11 +59,13 @@ export class ClientProxyHandler<T extends object> implements ProxyHandler<T> {
     private initializeRpc(): void {
         // we need to set the flag to true before waiting for the channel provider. Otherwise `get` might
         // get called again and we'll try to open a channel more than once
+        this.initCallback.reportInit(this.id);
         this.isRpcInitialized = true;
         const clientOptions: RpcProtocolOptions = { encoder: this.encoder, decoder: this.decoder, mode: 'clientOnly' };
         this.channelProvider().then(channel => {
             const rpc = new RpcProtocol(channel, undefined, clientOptions);
             this.rpcDeferred.resolve(rpc);
+            this.initCallback.reportInitDone(this.id);
```
martin-fleck-at, line 68: I'm wondering whether we could re-use the rpcDeferred in our synchronization class, something like `initializeProxy(id, deferred)` and have the implementation resolve it and remove it from the queue when deferred is resolved. Just a thought though.

martin-fleck-at (changes requested):
Change looks good to me and I think it should properly sync messages during initialization phase. However, I do have a few challenges with the naming as I believe it makes it hard to understand what is happening. Please see inline comments for that.

packages/plugin-ext/src/common/rpc-protocol.ts
```diff
@@ -149,6 +154,39 @@ export class RPCProtocolImpl implements RPCProtocol {
     }
 }
 
+export class InitializationCallbackImpl implements ProxySynchronizer {
```
martin-fleck-at, line 157: Could we maybe calls this class `DefaultProxySynchronizer` or `ProxySynchronizerImpl` or something similar instead of `InitializationCallbackImpl` since I think the name `callback` is not that great in that context. That also applies to the property name then.

packages/plugin-ext/src/common/proxy-handler.ts
```diff
@@ -25,11 +25,18 @@ export interface RpcHandlerOptions {
...
     channelProvider: () => Promise<Channel>,
+    proxySynchronizer: ProxySynchronizer,
 }
 
 export interface InvocationHandlerOptions extends RpcHandlerOptions {
     target: any
 }
+
+export interface ProxySynchronizer {
+    startProxyInitialization(id: string, init: Promise<void>): void
```
martin-fleck-at, line 36: Should we ease our lives and simply take a `Promise<RpcProtocol>`? Then we do not have to do an empty promise-chaining later.

packages/plugin-ext/src/common/proxy-handler.ts
```diff
@@ -69,13 +78,13 @@ export class ClientProxyHandler<T extends object> implements ProxyHandler<T> {
         const isNotify = this.isNotification(name);
         return (...args: any[]) => {
             const method = name.toString();
-            return this.rpcDeferred.promise.then(async (connection: RpcProtocol) => {
+            return this.proxySynchronizer.pendingProxyInitializations().then(() => this.rpcDeferred.promise.then(async (connection: RpcProtocol) => {
```
martin-fleck-at, line 81: Should we make this into an async function to make the code look a little bit nicer? Doesn't have to be done, just a thought.

martin-fleck-at (changes requested):

packages/plugin-ext/src/common/rpc-protocol.ts
jfaltermeier: Sorry, missed this one during the rename

packages/plugin-ext/src/common/proxy-handler.ts
jfaltermeier: I would rather keep it this way since the `RpcProtocol` is private in the `ClientProxyHandler` and I don't see a benefit in handing it out.

martin-fleck-at (approved):
LGTM!

packages/plugin-ext/src/common/proxy-handler.ts
martin-fleck-at: Fine with me.

PR #13196, opened 2023-12-20
fix: debug configuration provider replaces different provider

arekzaluski:
#### What it does

This PR solves the problem when multiple plugins try to register debug configuration provider under the same handle number. Currently the old provider gets replaced. After this change the new number will be assigned.

Similar guard exists already in `registerDebugAdapterContribution`

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/debug/plugin-debug-service.ts
+++ b/packages/plugin-ext/src/main/browser/debug/plugin-debug-service.ts
@@ -97,6 +97,13 @@ export class PluginDebugService implements DebugService {
     }, 100);
 
     registerDebugConfigurationProvider(provider: PluginDebugConfigurationProvider): Disposable {
+        if (this.configurationProviders.has(provider.handle)) {
+            const configuration = this.configurationProviders.get(provider.handle);
+            if (configuration && configuration.type !== provider.type) {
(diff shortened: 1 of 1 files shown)


packages/plugin-ext/src/main/browser/debug/plugin-debug-service.ts
```diff
@@ -97,6 +97,13 @@ export class PluginDebugService implements DebugService {
     }, 100);
 
     registerDebugConfigurationProvider(provider: PluginDebugConfigurationProvider): Disposable {
+        if (this.configurationProviders.has(provider.handle)) {
+            const configuration = this.configurationProviders.get(provider.handle);
+            if (configuration && configuration.type !== provider.type) {
+                console.warn(`Different debug configuration provider with type '${configuration.type}' already registered.`);
+                provider.handle = this.configurationProviders.size;
+            }
+        }
```
arekzaluski, line 106: Different way of solving it is the approach from `registerDebugAdapterContribution`. Instead of assigning a new handle number we can return `Disposable.NULL`

thegecko (approved):

planger (approved):

JonasHelming:
@thegecko If you merge now, it is the todays release :-)

Issue #13161, opened 2023-12-07
FileService.readFile reads whole file instead of only the requested part (inefficient in general, even fails for >2GB files)


dfriederich:
When using FileService.readFile to only read a part of a file ({ length: 2 } option), the API still reads in the full file and then only returns the requested bytes.
This did cause in my usage to detect the file type that it failed for files > 2GB unnecessarily. 
But even for files < 2GB, silently reading in much more than was requested is quite inefficient. The handling is according tot he source comment, and optimization. Maybe it should not happen if a length is requested?

Not a big problem for me, I'm not using as workaround FileService.readFileStream
`// Workaround for FileService.readFile to just read the file unbuffered as optimization.
// This breaks to read the first 2 bytes of a > 2GB file as in that code flow it reads the whole monster and then cuts
// afterwards. The workaround bypasses that code and reads the file as a stream.

async function readFileBuffered(fileService: FileService, resource: URI, options?: ReadFileOptions): Promise<FileContent> {
    const stream = await fileService.readFileStream(resource, options);
    return {
        ...stream,
        value: await BinaryBufferReadableStream.toBuffer(stream.value)
    };
}
`

but through still worth reporting as issue to possibly improve.

PR #13197, opened 2023-12-20
Only read unbuffered when we read the whole file

martin-fleck-at:
#### What it does

We only want to use unbuffered reading when we read the whole file, otherwise the optimization we gain through it seems to be lost.

Please note that there is a difference between `readFile` which returns a binary buffer and `read` which is used to read textual content. Furthermore, limits are applied automatically on the `read` method so even if you only want to read 10 characters it will fail if the file size is larger than the limit (`files.maxFileSizeMB` preference). 

Fixes https://github.com/eclipse-theia/theia/issues/13161

#### How to test

- Open any file with the `FileService` and provide a specific length to read, see bug report
- You can use https://github.com/eclipse-theia/theia/tree/issues/13161-test which simply reads a length of 10 with `read` and `readFile` using the `SampleCommand2`.
- Both should work without issues, even if you choose a large enough file as I set the limit option of the `read` method to the highest, safe integer.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/browser/file-service.ts
+++ b/packages/filesystem/src/browser/file-service.ts
@@ -694,12 +694,7 @@ export class FileService {
     async read(resource: URI, options?: ReadTextFileOptions): Promise<TextFileContent> {
         const [bufferStream, decoder] = await this.doRead(resource, {
             ...options,
-            // optimization: since we know that the caller does not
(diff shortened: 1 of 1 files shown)


tortmayr (approved):
Changes look good to me.The Test Scenario now works as expected.

Issue #13094, opened 2023-11-23
Theia creates 2 web-sockets if we override the WebSocketConnectionProvider class


safisa:
### Bug Description:
Hi,
It could be difficult to explain the issue I am facing on the latest Theia (1.43.1), but I will try:

I have a class that extends the WebSocketConnectionProvider:
```
export class CustomWebSocketConnectionProvider extends WebSocketConnectionProvider {
    protected override createWebSocket(url: string): Socket {
        ...
    }
}
```

And here is the binding code:
```
bind(CustomWebSocketConnectionProvider).toSelf().inSingletonScope();
rebind(WebSocketConnectionProvider).toService(CustomWebSocketConnectionProvider);
```

It was working correctly until I upgraded to 1.43.1 (from 1.40). I see now in the dev-tools on the network tab that there are two web sockets, and only one is actually working (before that there was only one).

If I debug the code from dev-tools and put a breakpoint inside the createWebSocket method in both classes (CustomWebSocketConnectionProvider and the base WebSocketConnectionProvider class), and reload the page, the break point breaks in both of them! this should not be the case, since I override this class and rebind it to the new one.

I think this issue is caused by the Remote SSH PR from 1.43: https://github.com/eclipse-theia/theia/pull/12618

### Steps to Reproduce:

1. create a new class that extends the WebSocketConnectionProvider
2. it can be an empty class, no need to override anything
3. add the binding as above
4. reload and check the network tab with WS filtering
5. you will see 2 WS
6. also you can override the createWebSocket, you can just call the super once, and put the breakpoint as described above..

### Additional Information

- Operating System: Mac
- Theia Version: 1.43.1

msujew:
@safisa You probably need to rebind the local websocket connection provider then as well:

```ts
rebind(LocalWebSocketConnectionProvider).toService(CustomWebSocketConnectionProvider);
```

See also [our own messaging module](https://github.com/eclipse-theia/theia/blob/master/packages/core/src/browser/messaging/messaging-frontend-module.ts).

safisa:
@msujew I already tried this but didn't help, same issue.

safisa:
@msujew I think it is related to the PR from 1.42 and not the one above: [12590](https://github.com/eclipse-theia/theia/pull/12590)

But, I don't know how to make my custom binding work at the preload level.
I tried to move my binding from my frontend module to a new preload module like this:

src/browser/preload/custom-preload-module.ts:
```
[...]

PR #13199, opened 2023-12-20
Allow to rebind messaging services in preload

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/13094

When trying to rebind the messaging services in the preload step, you were only able to bind them for a child container of the original container. That eventually led to multiple messaging services being created, which leads to unexpected issues, such as multiple websocket connections (where only one is actually needed).

This change binds the preload services directly on the global frontend container. This way, adopters can rebind messaging services already in the preload stage, preventing multiple instances of those services from appearing.

#### How to test

You will need to rebind the messaging services. Create a new module and perform the binding like this:

```ts
import { ContainerModule } from 'inversify';
import { LocalConnectionProvider, RemoteConnectionProvider, ServiceConnectionProvider } from '../messaging/service-connection-provider';
import { ConnectionSource } from '../messaging/connection-source';
import { WebSocketConnectionSource } from '../messaging/ws-connection-source';

const backendServiceProvider = Symbol('backendServiceProvider');

export default new ContainerModule((bind, _, __, rebind) => {
    bind(backendServiceProvider).toDynamicValue(ctx => {
        bind(CustomServiceConnectionProvider).toSelf().inSingletonScope();
        const container = ctx.container.createChild();
        container.bind(ConnectionSource).toService(WebSocketConnectionSource);
        return container.get(CustomServiceConnectionProvider);
    }).inSingletonScope();
    rebind(LocalConnectionProvider).toService(backendServiceProvider);
    rebind(RemoteConnectionProvider).toService(backendServiceProvider);
});

class CustomServiceConnectionProvider extends ServiceConnectionProvider {

    override init(): void {
        console.log('Constructing custom service connection');
        super.init();
    }
}

```

Afterwards put this in an additional `frontendPreload` entry in the package.json, build and start the browser app. At frontend startup, the logging above should be printed only **once** and the frontend preload should perform without errors.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-manager/src/generator/frontend-generator.ts
+++ b/dev-packages/application-manager/src/generator/frontend-generator.ts
@@ -92,9 +92,7 @@ function load(container, jsModule) {
         .then(containerModule => container.load(containerModule.default));
 }
 
-async function preload(parent) {
-    const container = new Container();
-    container.parent = parent;
+async function preload(container) {
     try {
 ${Array.from(frontendPreloadModules.values(), jsModulePath => `\
         await load(container, ${this.importOrRequire()}('${jsModulePath}'));`).join(EOL)}


tsmaeder (approved):
Makes sense.

PR #13201, opened 2023-12-20
Set initial size and position of secondary windows

vladarama:
#### What it does
This PR implements the initial size and positions of secondary windows. It adds a preference called `Secondary Window Placement` allowing the user to choose the placement of the secondary windows between:
- `originalSize` (default): same size as the original widget.
- `halfSize`: half the size of the running Theia application
- `fullSize`: the same size as the running Theia application.

In addition, another preference (`alwaysOnTop`) is introduced that allows the user to decide if they want the secondary window to always stay above all other windows, including those of different applications.

#### How to test

1.  Open Theia electron or browser application.
2.  Install the `drawio` extension.
3.  Modify the `Secondary Window Placement` preference.
4.  Create a new `test.drawio` file or create a new terminal and open it as a secondary window.
5.  The secondary window' size and placement should match the chosen preference.
6.  Repeat steps 3-5 with the other preferences.

#### Follow-ups

The secondary window popup might not line up perfectly with the original widget since the API used to get the window dimensions has inconsistencies across different browsers and operating systems.

#### Review checklist

#### Reminder for reviewers


packages/core/src/browser/window/default-secondary-window-service.ts
```diff
@@ -124,6 +135,48 @@ export class DefaultSecondaryWindowService implements SecondaryWindowService {
...
+                width = widget.node.clientWidth;
+                left = window.screenLeft + clientBounds.x;
+                top = window.screenTop + (window.outerHeight - window.innerHeight) + offsetY;
+                if (environment.electron.is()) {
+                    top = window.screenTop + clientBounds.y;
+                }
+                break;
+            }
+            case 'halfSize': {
+                height = window.innerHeight - (window.outerHeight - window.innerHeight);
```
jfaltermeier, line 157: From the preference name I would have expected the height to be half the size as well. 
Should we adjust the preference name/height computation/preference description?

packages/core/src/browser/window/default-secondary-window-service.ts
```diff
@@ -100,7 +105,13 @@ export class DefaultSecondaryWindowService implements SecondaryWindowService {
     }
 
     protected doCreateSecondaryWindow(widget: ExtractableWidget, shell: ApplicationShell): Window | undefined {
-        const newWindow = window.open(DefaultSecondaryWindowService.SECONDARY_WINDOW_URL, this.nextWindowId(), 'popup') ?? undefined;
+        let options;
+        const [height, width, left, top] = this.findSecondaryWindowCoordinates(widget);
+        options = `popup=1,width=${width},height=${height},left=${left},top=${top}`;
+        if (this.preferenceService.get('window.secondaryWindowAlwaysOnTop')) {
+            options += ',alwaysOnTop=true';
```
jfaltermeier, line 112: `alwaysOnTop` makes the window to be on top of all windows, at least on Ubuntu with the electron application. 
When I move a different application to this screen, e.g. the browser, the popup will also be on top of the browser. I would have expected the popup to behave like a modal dialog without disabling interaction with the main application. 

Is this true on your OS as well? Can this be changed? If not maybe the default should be `false` instead. What do you think?

In the browser application this does not seems to be supported as it opens a new browser window for me.

packages/core/src/browser/window/default-secondary-window-service.ts
```diff
@@ -124,6 +135,48 @@ export class DefaultSecondaryWindowService implements SecondaryWindowService {
         return newWindow;
     }
 
+    protected findSecondaryWindowCoordinates(widget: ExtractableWidget): (number | undefined)[] {
+        const clientBounds = widget.node.getBoundingClientRect();
+        const preference = this.preferenceService.get('window.secondaryWindowPlacement');
+
+        let height; let width; let left; let top;
+        const offsetY = 20;
```
jfaltermeier, line 143: Could you explain where the 20 is coming from? Maybe as a comment in the code

jfaltermeier (commented):
Thanks for the PR. I've tested on Ubuntu with the electron and the browser application and it works very well overall. I've added a few comments where the implementation did not match what I expected as a user.

packages/core/src/browser/window/default-secondary-window-service.ts
vladarama: Good point, I changed it to `halfWidth`. I expect users to use this preference like a split-screen mode.

packages/core/src/browser/window/default-secondary-window-service.ts
vladarama: I agree, I set the default to `false` and updated the preference description to better explain the behavior.

packages/core/src/browser/window/default-secondary-window-service.ts
vladarama: Its an offset to avoid the browser window title bar. I use it to better align the secondary-windows. 
I also added a comment for clarification :+1:

vladarama (commented):
@jfaltermeier 
Thanks for the review,
I have addressed your comments and the implementation should now match the user's expectations :rocket:

jfaltermeier (approved):
Thanks, lgtm!

PR #13217, opened 2023-12-28
Upgrade Monaco dependency to 1.83.1

### LINKED ISSUE #12679: monaco: upgrade `@theia/monaco-editor-core`
#### Feature Description:

The issue tracks the upgrade of `@theia/monaco-editor-core` which pulls a newer version of the monaco-editor built from [theia-ide/vscode](https://github.com/theia-ide/vscode). The current version of monaco matches `1.72.3` while our supported API is currently at `1.78.0` and soon to be `1.79.0`.

I propose we use the latest tagged release of VS Code and benefit from monaco updates, UI/UX improvements and increased compatibility.

#### Steps

- [ ] release a new version of `@theia/monaco-editor-core`
- [ ] adapt the new editor version to the framework
- [ ] extensive testing and open an upstream pull-request

#### Documentation

Information related to performing the monaco uplift:
- https://github.com/eclipse-theia/theia/tree/master/packages/monaco#monaco-uplifts

tsmaeder:
#### What it does
Upgrades our Monaco dependency to 1.83.1

Fixes #12679

Various changes were necessary, in particular:

1. There is a new Implementation of the diff editor
2. The `setContext` command has been renamed to '_setContext`.
3. The typing for the URI class has changed and we need to do some casting: https://github.com/microsoft/vscode/issues/190584
4. Our overriding of VS Code services in `StandaloneServices` has not been working for a while, but VS Code instantiating services eagerly instead of lazily has revealed the problem
5. VS Code adds many icons to the registry in parts of VS Code that are not used in Theia. Instead of hacking the Theme icon id field, we now just register those icons
6. Instead of adding our own styling for contributed icons, we now properly use the VS Code icon registry.

#### How to test
A lot of parts of Theia may be affected by this change and need testing, in particular:

1. The monaco editor
2. quick input
3. debugger
4. Built-in icons and themes

I'm opening this PR now in order to facilitate code review, but we still need

- [x] an IP check issue to be approved: https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/12577

#### Follow-ups
- [ ] support editor.indentSize: https://github.com/eclipse-theia/theia/pull/13105
- [ ] stubbed VS Code API:
  - provideDocumentRangesFormattingEdits
  - registerMultiDocumentHighlightProvider
- [ ] check implementation notes in `languages-main.ts`
- [ ] IQuickinputOptions.linkOpenerDelegate: unimplementd feature

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,18 @@
 
 - [Previous Changelogs](https://github.com/eclipse-theia/theia/tree/master/doc/changelogs/)
 
-<!-- ## not yet released
+## v1.47.0 not yet released
 
 - [component] add here
 
 <a name="breaking_changes_not_yet_released">[Breaking Changes:](#breaking_changes_not_yet_released)</a>
-
-- [component] add here -->
+- [monaco] Upgrade Monaco dependency to 1.83.1 [#13217](https://github.com/eclipse-theia/theia/pull/13217)- contributed on behalf of STMicroelectronics\
+  There are a couple of breaking changes that come with this monaco update
+  - Moved `ThemaIcon` and `ThemeColor` to the common folder
+  - Minor typing adjustments in QuickPickService: in parti
+  - FileUploadService: moved id field from data transfer item to the corresponding file info
+  - The way we instantiate monaco services has changed completely: if you touch monaco services in your code, please read the description in the 
+    file comment in `monaco-init.ts`.
 
 ## v1.46.0 - 01/25/2024
 - [plugin] Add prefix to contributed view container ids [#13362](https://github.com/eclipse-theia/theia/pull/13362) - contributed on behalf of STMicroelectronics
--- a/dev-packages/application-manager/src/generator/frontend-generator.ts
+++ b/dev-packages/application-manager/src/generator/frontend-generator.ts
@@ -108,7 +108,7 @@ ${Array.from(frontendPreloadModules.values(), jsModulePath => `\
 }
 
 module.exports = (async () => {
-    const { messagingFrontendModule } = require('@theia/core/lib/${!this.pck.isElectron()
+    const { messagingFrontendModule } = require('@theia/core/lib/${this.pck.isBrowser()
                 ? 'browser/messaging/messaging-frontend-module'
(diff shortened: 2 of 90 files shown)


tsmaeder:
Note that the failure in test are due to the way nodejs handle ES6 modules: here's what I fount out:

1. The @theia/monaco-editor-core module is loaded as a CJS module due to the fact that the package.json does not contain a `type: module` line

3. With this uplift, we _do_, for the first time have actual ES6 syntax in the monaco files: 
```@theia/preferences: D:\a\theia\theia\node_modules\@theia\monaco-editor-core\esm\vs\base\common\marked\marked.js:3084
@theia/preferences: export var Lexer = (__marked_exports.Lexer || exports.Lexer);
@theia/preferences: ^^^^^^
@theia/preferences: SyntaxError: Unexpected token 'export'
@theia/preferences:     at compileFunction (<anonymous>)
```
   This is the only place actual `import` or `export` statements are used.

4. When I tried to convert the monaco package to a real module, I got 
```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: <....monaco module>
```
This is because typescript compiles import to use "require". So we can't really mix & match ESM and CJS modules, it seems.

tsmaeder:
Maybe we can hack the module in question to remove the exports, but what about the next time some ES6-syntax appears in VS Code?

msujew (commented):
FYI, I've noticed that the change breaks notebook support, see:

image

tsmaeder:
@msujew the performance hit vs. VS Code seems to be mostly due to the patched `IConfigurationService` we set up in `monaco-frontend-module.ts`. At least, if I just return a `StandaloneConfigurationService`, opening `theia.d.ts` performs similarly to VS Code. If there is worse performance in this branch I assume it's because with the new service override implementation, the patched service is used everywhere in Monaco (for example in editor contributions) where it might not have been before.

tsmaeder:
@msujew I believe the performance problem should be fixed with the latest commit. Can you give it a go?

msujew (approved):
Looks pretty good to me. I couldn't find any further regressions and the performance is back to the before-change level.

PR #13225, opened 2024-01-03
Use more stable hostname for webviews #13092

### LINKED ISSUE #13092: vscode: incorrect Web API localStorage and sessionStorage behavior in Theia webview
### Bug Description:
Using `window.localStorage` and `window.sessionStorage` from a vscode extension Webview in Theia doesn't behave as expected--not compared to how it works in VS Code.

Data stored in `sessionStorage` survives a number of things in VS Code:

- the view being hidden and reshown
- the view being closed and reopened
- the app being refreshed (with CTRL-R)
- It does *not* survive an app restart.

Data stored in `localStorage` survives all those things plus it survives an app restart. I.e., the data is stored to disk.

In Theia, neither storage behaves how it does in VS Code. Both forms only survive the view being hidden and reshown.
 
### Steps to Reproduce:
```
git clone -b webview_storage https://github.com/jcortell68/theiasandbox
cd theiasandbox
yarn
yarn start:electron
```
1. View > Open View > My Webview
2. Click `Increment Counters` a few times
3. Note that both counters shown are the same value
4. Click on the Problems view (or another view in the Pane) then bring My Webview back to the forefrount.
5. Click 'Increment Counters`. 
6. Note that both counters increased by one from the value you last saw. That's good. Expected and correct.
7. Now close the view. Right click on the tab and hit 'Close'
8. Reopen the view with View > Open View > My Webview
9. Click `Increment Counters` and notice that both counters got reset. They now have a value of 1. This is the bug.

In VS Code, the counters would not have reset. In fact, even if you CTRL-R (refresh the VS Code app), the counters don't reset.
If you close VS Code altogether, relaunch it, bring up the view and click `Increment Counters`, you would see that `sessionStorage` was reset, which is expected. `localStorage` is not reset as that data is apparently being saved to disk. And that is why there are two forms of storage. One of them stores to the filesystem; the other just to RAM.

You can see this in vscode as follows:
```
git clone -b webview_storage https://github.com/jcortell68/vscodeextfeatures
cd vscodeextfeatures
yarn
code .
```

Launch the app from vscode and repeat the steps described above for Theia.

### Additional Information

- Operating System: Windows
- Theia Version: 1.43.1

jfaltermeier:
#### What it does

* add optional `viewId` to `WebviewWidgetIdentifier`, which may be used to create stable hostnames for webviews
* use this `viewId` to store a uuid which will be used as part of the stable hostname

For VSCode the WebView iFrame src url looks like this:
```
vscode-webview://0a7iho5kovpr5p4ctn5fvcns9h4du51vr2fb4iachonmmqsceaes/fake.html?id=ad9dff6c-9d20-4aec-b8b1-edfeecc19d07 (initial open)
vscode-webview://0a7iho5kovpr5p4ctn5fvcns9h4du51vr2fb4iachonmmqsceaes/fake.html?id=faec60d8-1f93-4e66-94b7-931081a8a4ba (close and reopen; restart of application)
```

In Theia, the pattern was like this:
```
http://401ed951-431e-4866-9a42-b033d95521d4.webview.localhost:33391/webview/fake.html?id=401ed951-431e-4866-9a42-b033d95521d4 (initial open)
http://837d0c96-f06d-4bad-91ba-efec327876be.webview.localhost:33391/webview/fake.html?id=837d0c96-f06d-4bad-91ba-efec327876be (close and reopen)
http://6444a35c-161b-4205-96b4-a5aee01c0f9e.webview.localhost:37883/webview/fake.html?id=6444a35c-161b-4205-96b4-a5aee01c0f9e (restart of application)
```

The changing id was used as part of the hostname. Also the port changes between restarts of the electron app. 

With this change the src url will look more like this:
```
http://33544e87-f49c-45eb-b7f7-3387c39fc913.webview.localhost:33391/webview/fake.html?id=401ed951-431e-4866-9a42-b033d95521d4 (initial open)
http://33544e87-f49c-45eb-b7f7-3387c39fc913.webview.localhost:33391/webview/fake.html?id=837d0c96-f06d-4bad-91ba-efec327876be (close and reopen)
http://33544e87-f49c-45eb-b7f7-3387c39fc913.webview.localhost:37883/webview/fake.html?id=6444a35c-161b-4205-96b4-a5aee01c0f9e (restart of application)
```

In order to fix the port in electron, which is 0 by default so any non-used port, adopers may use the --port option when starting Theia. I think 0 should stay the default.

Fixes #13092

#### How to test

Webviews should still work as before. 
When it comes to testing the storage you may use the reproducer from https://github.com/eclipse-theia/theia/issues/13092 and start the electron example either with 
* `yarn electron start --port=3124`
* a custom electron main script, e.g. 
```
process.argv.push('--port=3124')
require('../lib/backend/electron-main.js');
```

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/view/plugin-view-registry.ts
+++ b/packages/plugin-ext/src/main/browser/view/plugin-view-registry.ts
@@ -18,7 +18,7 @@ import { injectable, inject, postConstruct, optional } from '@theia/core/shared/
 import {
     ApplicationShell, ViewContainer as ViewContainerWidget, WidgetManager, QuickViewService,
     ViewContainerIdentifier, ViewContainerTitleOptions, Widget, FrontendApplicationContribution,
-    StatefulWidget, CommonMenus, TreeViewWelcomeWidget, ViewContainerPart, BaseWidget
+    StatefulWidget, CommonMenus, TreeViewWelcomeWidget, ViewContainerPart, BaseWidget,
 } from '@theia/core/lib/browser';
 import { ViewContainer, View, ViewWelcome, PluginViewType } from '../../../common';
 import { PluginSharedStyle } from '../plugin-shared-style';
@@ -424,7 +424,10 @@ export class PluginViewRegistry implements FrontendApplicationContribution {
 
     protected async createNewWebviewView(viewId: string): Promise<WebviewView> {
         const webview = await this.widgetManager.getOrCreateWidget<WebviewWidget>(
-            WebviewWidget.FACTORY_ID, <WebviewWidgetIdentifier>{ id: v4() });
+            WebviewWidget.FACTORY_ID, <WebviewWidgetIdentifier>{
+                id: v4(),
+                viewId,
+            });
         webview.setContentOptions({ allowScripts: true });
 
         let _description: string | undefined;
@@ -893,7 +896,7 @@ export class PluginViewRegistry implements FrontendApplicationContribution {
             const webviewView = await this.createNewWebviewView(viewId);
             webviewId = webviewView.webview.identifier.id;
         }
-        const webviewWidget = this.widgetManager.getWidget(WebviewWidget.FACTORY_ID, <WebviewWidgetIdentifier>{ id: webviewId });
+        const webviewWidget = this.widgetManager.getWidget(WebviewWidget.FACTORY_ID, <WebviewWidgetIdentifier>{ id: webviewId, viewId });
         return webviewWidget;
     }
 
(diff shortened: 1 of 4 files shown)


martin-fleck-at (approved):
Change looks good to me, thank you very much Johannes! 

I can see that with a fixed port both the session storage and the local storage behave as in VS Code. Without a fixed port, only the session storage behaves as expected and the local storage is cleared. While I understand the technical reasons behind it, I am wondering whether there is a way to also abstract from the port number using a custom scheme as VS Code does or whether this is something we do not even want. In any case, this contribution already provides enough value in itself so it can definitely be merged.

jfaltermeier:
Thanks. I haven't investigated if we could introduce a custom scheme, we if required at some point we can come back to it. 
For the browser version we may want to keep the current scheme though.

Issue #13051, opened 2023-11-02
VS Code declarative API compatibility


tortmayr:
Due to recent efforts Theia's plugin API is mostly compatible with the programmatic VS Code API.
However, looking at the declarative API (i.e. contribution points in the package.json) there are still many contribution points
which are not (or only partly) supported.

I had a more detailed look on the menu contribution points (https://code.visualstudio.com/api/references/contribution-points#contributes.menus) and did a quick reference check in Theia repo do identify which contributions points are currently not supported:

- [ ] comments/commentThread/title: Only referenced in ts-doc, no actual implementation
- [ ]  extension/context: not supported (verified with ms.python extension)
- [ ] interactive/toolbar: not mentioned/supported
- [ ] interactive/cell/title: not mentioned/supported
- [ ] timeline/title:  not mentioned/supported
- [ ] scm/change/title: not mentioned/supported
- [ ] scm/sourceControl: not mentioned/supported
- [ ] terminal/context: not mentioned/supported
- [ ] terminal/title/context: not mentioned/supported
- [ ] testing/item/gutter: not mentioned/supported
- [ ] touchbar: not supported (only mentioned in i18n) [mac only feature]
- [ ] webview/context: not supported (tested with small custom extension)

tortmayr:
@rschnekenbu @planger

umar-shabbir:
I checked VS code walkthrough and it is not supported in Theia.
https://community.theia-ide.org/t/vs-code-walkthrough-contribution-point-in-theia/3312

rschnekenbu:
As a follow up of #13145, we are missing also 'testing/message/content' introduced with vscode 1.84. This menu contribution is not yet relevant given the current test UI.

jonah-iden:
Some of these contribution points can't be implemented because the associated feature is missing in theia
- timeline/title: needs the timeline feature
- scm/sourceControl: needs multi repository support
- scm/change/title: needs review comment support i think. See [here](https://code.visualstudio.com/updates/v1_18#_scm-new-menu-scmchangetitle)
- interactive/toolbar and interactive/cell/title: Should be somewhere in the notebook support, but i havn't been able to find them in vscode yet.

rschnekenbu:
_vscode.git_ extension contributes to a currently proposed menu extension, `editor/content`. 
This menu is not tracked by this issue obviously, as it is a proposed one. But it may be required for the adoption of the _vscode.git_ extension. 
[...]

PR #13226, opened 2024-01-03
Add a few vscode menu contribution points

jonah-iden:
#### What it does

implements parts of #13051
specificly:
- extension/context
- terminal/context
- terminal/title/context

since Theias terminals are there own tabs, the last contribution is implemented when right clicking on the tab title.
The others should be at the same place they are in vscode

#### How to test

[Here](https://github.com/jonah-iden/contribution-point-test) is a test extension (inlucluding vsix) which adds a menu entry to all of theias missing contribution points.

`extension/context` can be found when clicking the cog wheel (or right clicking) on an extension in the Extension widget
`terminal/context` can be found when right clicking inside a terminal
`terminal/title/context` can be found when right clicking the tab title of a terminal

im not too happy with having the menu path `['extensions_context_menu', '3_contribution']` duplicated. If you have an idea where to place the constant for this i'd be happy to implement it.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/tab-bars.ts
+++ b/packages/core/src/browser/shell/tab-bars.ts
@@ -648,7 +648,7 @@ export class TabBarRenderer extends TabBar.Renderer {
                 this.selectionService.selection = NavigatableWidget.is(widget) ? { uri: widget.getResourceUri() } : widget;
             }
 
-            const contextKeyServiceOverly = this.contextKeyService?.createOverlay([['tabContextWidgetId', widget?.id]]);
+            const contextKeyServiceOverly = this.contextKeyService?.createOverlay([['isTerminalTab', widget && '_terminalId' in widget]]);
             this.contextMenuRenderer.render({
                 menuPath: this.contextMenuPath!,
                 anchor: event,
--- a/packages/core/src/browser/shell/tab-bars.ts
+++ b/packages/core/src/browser/shell/tab-bars.ts
@@ -39,6 +39,7 @@ import { SelectComponent } from '../widgets/select-component';
 import { createElement } from 'react';
 import { PreviewableWidget } from '../widgets/previewable-widget';
 import { EnhancedPreviewWidget } from '../widgets/enhanced-preview-widget';
+import { ContextKeyService } from '../context-key-service';
 
 /** The class name added to hidden content nodes, which are required to render vertical side bars. */
 const HIDDEN_CONTENT_CLASS = 'theia-TabBar-hidden-content';
@@ -102,7 +103,8 @@ export class TabBarRenderer extends TabBar.Renderer {
         protected readonly selectionService?: SelectionService,
         protected readonly commandService?: CommandService,
         protected readonly corePreferences?: CorePreferences,
-        protected readonly hoverService?: HoverService
+        protected readonly hoverService?: HoverService,
+        protected readonly contextKeyService?: ContextKeyService,
     ) {
         super();
         if (this.decoratorService) {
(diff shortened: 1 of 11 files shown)


rschnekenbu:
I would be happy to help for review here

packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -731,6 +735,15 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
...
             commandId: TerminalCommands.KILL_TERMINAL.id
         });
+
+        menus.registerSubmenu(TerminalMenus.TERMINAL_CONTRIBUTIONS, '', {
+            role: CompoundMenuNodeRole.Group
+        });
+
+        menus.registerSubmenu(TerminalMenus.TERMINAL_TITLE_CONTRIBUTIONS, '', {
+            role: CompoundMenuNodeRole.Group,
+            when: 'tabContextWidgetId =~ /^terminal-[0-9]+$/'
```
rschnekenbu, line 745: This regex might be too restrictive. I have the contributed menu on the title when this is a SHELL terminal, but I do not have it for the Javascript Debug Terminal. In VS Code, I have the menu contribution for both of them.

packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
```diff
@@ -143,6 +143,9 @@ export class VSXExtensionsContribution extends AbstractViewContribution<VSXExten
             commandId: VSXExtensionsCommands.INSTALL_ANOTHER_VERSION.id,
             label: nls.localizeByDefault('Install Another Version...'),
         });
+        menus.registerSubmenu(VSXExtensionsContextMenu.CONTRIBUTION, '', {
```
rschnekenbu, line 146: The example menu is shown for me on an extension, but when I try to execute the action, I have a failure with rpc message. There are 2 arguments passed for me to the handler (position and VSXExtension selected). I think the latter one would need to be removed, or at least processed before being passed to RPC. 

Message:
```
Uncaught EncodingError Error: Error during encoding: 'Unknown type: symbol'
    at encode (/home/remi/Git/theia/packages/core/src/common/message-rpc/rpc-message-encoder.ts:151:23)
    at request (/home/remi/Git/theia/packages/core/src/common/message-rpc/rpc-message-encoder.ts:137:14)
    at sendRequest (/home/remi/Git/theia/packages/core/src/common/message-rpc/rpc-protocol.ts:161:22)
    at <anonymous> (/home/remi/Git/theia/packages/plugin-ext/src/common/proxy-handler.ts:76:45)
    --- await ---
    at delegate (/home/remi/Git/theia/packages/core/src/common/menu/menu-adapter.ts:69:43)
```

rschnekenbu (changes requested):
Thanks for your contribution, @jonah-iden!
I have some issues with the sample extension provided on vsx extensions and terminal titles, could you have a look?

packages/terminal/src/browser/terminal-frontend-contribution.ts
jonah-iden: should be fixed now. Though if you have a better idea i'd be happy. Setting the `isTerminalTab` by looking if there is a field in a widget is probably not the nicest solution. I just can't think of anything else

packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
jonah-iden: Fixed and the command also correctly return the extension id as its first arg

packages/core/src/browser/shell/tab-bars.ts
```diff
@@ -614,10 +616,12 @@ export class TabBarRenderer extends TabBar.Renderer {
                 this.selectionService.selection = NavigatableWidget.is(widget) ? { uri: widget.getResourceUri() } : widget;
             }
 
+            const contextKeyServiceOverly = this.contextKeyService?.createOverlay([['isTerminalTab', widget && '_terminalId' in widget]]);
```
rschnekenbu, line 619: I would rather use _terminalId_ property from the interface rather than the one from the implementation. 
That is usually what we do with the is() methods. I do not see better way to identify if the widget is a terminal widget.

rschnekenbu, line 619: nitpick: there is a typo here with a missing 'a' => contextKeyServiceOverl**a**y

rschnekenbu (commented):
That looks good to me, thanks for the contribution. I only have 2 minors comments, but otherwise, provided extension point work as expected.

rschnekenbu (approved):
Looks good to me now! Thanks @jonah-iden!

Issue #13224, opened 2024-01-02
Plugin API markdown file needs updating


jcortell68:
### Bug Description:

[The section on passing objects across RPC](https://github.com/eclipse-theia/theia/blob/master/doc/Plugin-API.md#complex-objects-and-rpc) is imprecise and apparently incomplete at this point. See the following discussion between myself and @cdamus 

![image](https://github.com/eclipse-theia/theia/assets/109228283/f7bbf952-4578-4140-9f98-782de63b9dc3)

PR #13238, opened 2024-01-04
Improve documentation on passing objects across RPC

martin-fleck-at:
#### What it does

Improves documentation.
Fixes https://github.com/eclipse-theia/theia/issues/13224

#### How to test

- Read and see if it is better than before ;-)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/doc/Plugin-API.md
+++ b/doc/Plugin-API.md
@@ -66,12 +66,17 @@ To communicate with each other, the implementation of each side of the API - `Ma
 The proxy is based on the interface of the other side: `Main` implementation has a proxy of the `Ext` interface and vice versa.
 The implementations do not have explicit dependencies to each other.
 
-Communication via RPC only supports transferring plain JSON objects: Only pure DTO objects without any functions can be transmitted.
-Consequently, objects with functions need to be cached and references to such objects need to be transmitted as handles (ids) that are resolved to the original object later on to invoke functions.
+### Encoding and Decoding RPC Messages
 
-For instance, in [LanguagesExtImpl](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/languages.ts)#registerCodeActionsProvider a new code action provider is cached on the `Ext` side and then registered on the `Main` side via its handle.
-When the code action provider’s methods are later invoked on the `Main` side (e.g. in [LanguagesMainImpl](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/main/browser/languages-main.ts)#provideCodeActions), it calls the `Ext` side with this handle.
-The `Ext` side then gets the cached object, executes appropriate functions and returns the results back to the `Main` side (e.g. in [LanguagesExtImpl](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/languages.ts)#$provideCodeActions).
+The communication between each side of the API is goverend by proxies that use a `RpcProtocol` on a given channel to transmit RPC messages such as requests and notifications.
+In Theia, the encoding and decoding process of RPC messages can be customized through dedicated `RpcMessageEncoder` and `RpcMessageDecoder` classes that can be provided when a new RpcProtocol is created.
+The `RpcMessageEncoder` writes RPC messages to a buffer whereas the `RpcMessageDecoder` parses a binary message from a buffer into a RPC message.
+
+By default, Theia uses an encoder and decoder based on [msgpackr](https://www.npmjs.com/package/msgpackr) that already properly handles many built-in types as well as arrays and maps.
+We can separately extend the encoding and decoding of our own classes by installing extensions using the `MsgPackExtensionManager` singleton, accessible from both ends of the channel.
+Examples of this can be found in the [extension for Errors](https://github.com/eclipse-theia/theia/blob/72421be24d0461f811a39324579913e91056d7c4/packages/core/src/common/message-rpc/rpc-message-encoder.ts#L171) and the [extension for URI, Range and other classes](https://github.com/eclipse-theia/theia/blob/72421be24d0461f811a39324579913e91056d7c4/packages/plugin-ext/src/common/rpc-protocol.ts#L223).
+Please note that it is the responsibility of the extension provider to use a free tag between 1 and 100.
+We call these extensions in the `index.ts` file to ensure they are available early on.
 
 ## Adding new API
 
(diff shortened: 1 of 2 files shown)


doc/Plugin-API.md
```diff
@@ -147,27 +152,30 @@ They each create the proxy to the other side in their constructors by using the
...
 ### Complex objects and RPC
 
-Only pure DTO objects without any functions or references to other objects can be transmitted via RPC.
-This often makes it impossible to just transfer objects provided by a plugin directly via RPC.
-In this case a DTO interface is necessary.
-These are defined [plugin-ext/src/common/plugin-api-rpc.ts](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/common/plugin-api-rpc.ts).
-Utility functions to convert between DTO and API types on the `Ext` side are usually added to [plugin-ext/src/plugin/type-converters.ts](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/type-converters.ts).
-Thus, this is also a good starting point to look for conversion utilities for existing types.
+When [encoding and decoding RPC messages](#encoding-and-decoding-rpc-messages) pure DTO objects can always safely be transmitted.
+For custom classes, we can also adapt the encoding and decoding through extensions.
```
jcortell68, line 156: see note above regarding `class`

doc/Plugin-API.md
```diff
@@ -147,27 +152,30 @@ They each create the proxy to the other side in their constructors by using the
...
-The `Ext` side then gets the cached object, executes appropriate functions and returns the results back to the `Main` side (e.g. in [LanguageExtImpl](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/languages.ts)#$provideCodeActions).
+The `Ext` side then gets the cached object, executes appropriate functions and returns the results back to the `Main` side (e.g. in [LanguagesExtImpl](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/languages.ts)#$provideCodeActions).
 
 Another example to browse are the [TaskExtImpl](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/tasks/tasks.ts) and [TaskMainImpl](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/main/browser/tasks-main.ts) classes.
 
+As we can see from the language and the task example, in Theia we have a complete API protocol that we need provide for plugins that may not be serializable by default.
```
jcortell68, line 169: This sentence needs some massaging, IMO. 

As part of that, I would say that the API involves transferring data back and forth that is housed in non-serializable objects, due to the presence of functions in those objects and/or references to other such objects. The objects need to be "dumbed down" to DOT types.

doc/Plugin-API.md
```diff
@@ -147,27 +152,30 @@ They each create the proxy to the other side in their constructors by using the
...
+In such cases, it has also proven useful to explicitly define a layer of DTO types that are then converted to their API counterpart on the other side.
```
jcortell68, line 170: I guess this is the part that has me a bit baffled (still), and why I didn't feel comfortable drafting these improvements. Above we mention the use of MsgPackExtensionManager for marshalling custom non-DTO objects, but here we point out this other technique of _explicitly_ defining DTO counterpart types and converting between them. As a reader, I'm wondering...why? Is it historical (predates MsgPackExtensionManager)? Is there some benefit to doing it explicitly this way in some cases? Is it _necessary_ to do it this way in some cases? Which approach should I take?

doc/Plugin-API.md
```diff
@@ -66,12 +66,17 @@ To communicate with each other, the implementation of each side of the API - `Ma
...
-The `Ext` side then gets the cached object, executes appropriate functions and returns the results back to the `Main` side (e.g. in [LanguagesExtImpl](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/languages.ts)#$provideCodeActions).
+The communication between each side of the API is goverend by proxies that use a `RpcProtocol` on a given channel to transmit RPC messages such as requests and notifications.
+In Theia, the encoding and decoding process of RPC messages can be customized through dedicated `RpcMessageEncoder` and `RpcMessageDecoder` classes that can be provided when a new RpcProtocol is created.
+The `RpcMessageEncoder` writes RPC messages to a buffer whereas the `RpcMessageDecoder` parses a binary message from a buffer into a RPC message.
+
+By default, Theia uses an encoder and decoder based on [msgpackr](https://www.npmjs.com/package/msgpackr) that already properly handles many built-in types as well as arrays and maps.
```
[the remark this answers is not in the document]
martin-fleck-at: Yeah, good point. They truly count as JavaScript built-in objects and are handled by [msgpackr natively](https://github.com/kriszyp/msgpackr/blob/master/pack.js#L419) as well. I'll changed the sentence as you suggested.

doc/Plugin-API.md
```diff
@@ -66,12 +66,17 @@ To communicate with each other, the implementation of each side of the API - `Ma
...
+We can separately extend the encoding and decoding of our own classes by installing extensions using the `MsgPackExtensionManager` singleton, accessible from both ends of the channel.
```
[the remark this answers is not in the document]
martin-fleck-at: I used the word class here because that is what is used as key by msgpackr when adding an extension, see https://github.com/kriszyp/msgpackr?tab=readme-ov-file#custom-extensions. We probably could also use the word `type` here as what you really need to provide is a constructor (function) so that we can make an `instanceof` check (see https://github.com/kriszyp/msgpackr/blob/master/pack.js#L444).

With regard to your comment below, I'll try to discuss with other people what the best option is. I don't want to change it to "non-DTO object" if we actually would suggest to "replace" the usage of DTO objects with a custom serialization approach.

doc/Plugin-API.md
martin-fleck-at: You are right, I believe in theory you could implement all DTOs through the msgpackr extension mechanism and avoid the explicit conversion from domain objects to DTOs and vice versa. I think there are probably several reasons that this is not the case for Theia API:
- msgpackr was only added later and we just added custom serializations for the classes that had that already
- API compliance: VS Code is also using [DTOs](https://github.com/microsoft/vscode/blob/cc7a0865e407e8b5eeaaf0b5cbda6f06f8a99d8b/src/vs/workbench/api/common/extHost.protocol.ts) so it is easier to align with their codebase if we keep the same/a similar system
- msgpackr only allows to register extensions globally so they leak into all connections whether they are related to the custom API or not, see also https://github.com/kriszyp/msgpackr/issues/93. So every single message that is sent may need to be checked against all registered extensions which may impact performance.
- msgpackr only supports [up to 100 extensions](https://github.com/kriszyp/msgpackr?tab=readme-ov-file#custom-extensions) (may not be a big problem for most APIs).

So I think it only makes sense to use msgpackr extensions for very common classes but use DTOs for the actual API. @tsmaeder What do you think about this?

doc/Plugin-API.md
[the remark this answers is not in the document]
jcortell68: Ah, ok. Makes sense to me.

doc/Plugin-API.md
jcortell68: It seems good to me...as long as it's in the markdown file and not just this PR comment :-)

I hadn't commented on your response since it looked like you were looking for confirmation from @tsmaeder . If you got that confirmation somehow, then yes...please include that insight in the markdown.

tsmaeder: Sorry, I kind of missed that one. The main problem with this, IMO, is that we don't know the context in which an object is used in a msgpackr extension. We would have to detect the correct type of object purely from the shape of the object. With 10s or 100s of different object, this seems difficult to maintain correctly whereas if we do the conversion manually, we _know_ the type of object we need to convert and the conversion code is close to its usage. Also, we get instances of interfaces from user code sometimes which we need to send to the front end: these objects may contain extra fields which fool our proposed type detection algorithm into thinking it's the wrong type of object. Or a user might implement an interface with a private field (`_myPrivateField`) and a getter (`myPublicGetter`). You can't detect this from shape of the object if you don't know the type.

martin-fleck-at: Thank you for your take on that @tsmaeder! I pushed an update that now should better explain when we use the msgpackr extensions and when we use custom DTOs.

martin-fleck-at:
@tsmaeder @jcortell68 I pushed an updated version of the documentation. It would be really great if you could have a very quick read through it to see if there is still something unclear or missing from your point of view. Thank you!

jcortell68:
> @tsmaeder @jcortell68 I pushed an updated version of the documentation. It would be really great if you could have a very quick read through it to see if there is still something unclear or missing from your point of view. Thank you!

The new text/description looks perfect to me. Thank you! I know this was a bit of a pain to pin down, but if it was difficult for us to do that, imagine the Theia newbie working on API trying to figure that out on his own. This will help a lot.

jcortell68 (approved):

jcortell68:
> @jcortell68 @tsmaeder If the documentation looks good to you, please approve the PR so I can merge it, thank you!

Ah. Didn't think my approval was technically needed since I'm not a maintainer/comitter in the project. Maybe Theia works differently than other open source projects I've worked with. Done.

martin-fleck-at:
@jcortell68 It is technically not needed but I like to get a clean state on the PR before merging it, also for history-preserving reasons ;-)

jfaltermeier (approved):

PR #13239, opened 2024-01-04
fixed multiple problems with the notebook output rendering

### LINKED ISSUE #13223: Notebooks: extra rendering of images
### Bug Description:

Images shown in a notebook using the `display` IPython builtin are rendered twice if they're above a certain size and  `display` is called more than once. That is, calling `display` twice will result in three images being rendered:

image

### Steps to Reproduce:

This can be reproduced in the Theia browser example.

- Install JN depencies: extensions for Python, Jupyter and VSCode builtins
- Install the required Python packages with `pip install ipykernel Pillow`. You may want to use a virtual environment
- Create a JN (a file with extension `.ipynb`) and open it
- Paste this example code:
```python
from PIL import Image

img_1 = Image.new("RGBA", (1600, 200), color="red")
img_2 = Image.new("RGBA", (1600, 200), color="blue")

display(img_1)
display(img_2)

```
- Run the notebook. Make sure the notebook is running in the environment where you installed the above packages

Observe the second image is sometimes rendered twice, once before the first image. The probability somehow depends on the size of the image, with the above size it should be around 50%, with larger images it approaches 100%.

In VSCode we only ever see the correct output, i.e. two images being rendered (one per `display` call):
image

### Additional Information

- Operating System: `Debian 6.1.67-1 (2023-12-12) x86_64 GNU/Linux`
- Theia Version: `master`
- Python version: 3.10
- `Pillow` version: `10.2.0`
- `IPython` version: `8.19.0`

jonah-iden:
#### What it does

fixes #13223

This fixes a few issues with the notebook output rendering and the associated webview. 
Also improves performance a bit because the webview is reused when the cell is executed again.

#### How to test
Create a new notebook or take an existing one. Just try to execute some cells.

Also see #13223 for specific test code. This should now be rendered correctly without extra outputs.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/view/notebook-code-cell-view.tsx
+++ b/packages/notebook/src/browser/view/notebook-code-cell-view.tsx
@@ -149,7 +149,8 @@ interface NotebookCellOutputProps {
 
 export class NotebookCodeCellOutputs extends React.Component<NotebookCellOutputProps> {
 
-    protected outputsWebview: CellOutputWebview | Promise<CellOutputWebview> | undefined;
+    protected outputsWebview: CellOutputWebview | undefined;
+    protected outputsWebviewPromise: Promise<CellOutputWebview> | undefined;
 
     protected toDispose = new DisposableCollection();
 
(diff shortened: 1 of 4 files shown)


packages/notebook/src/browser/view/notebook-code-cell-view.tsx
```diff
@@ -148,37 +149,52 @@ interface NotebookCellOutputProps {
 
 export class NotebookCodeCellOutputs extends React.Component<NotebookCellOutputProps> {
 
-    protected outputsWebview: CellOutputWebview | undefined;
+    protected outputsWebview: CellOutputWebview | Promise<CellOutputWebview> | undefined;
```
msujew, line 152: Suggestion: I feel like this leads to some anti-patterns. See in line 165:

```ts
this.outputsWebview = outputWebviewFactory(cell).then(webview => {
    this.outputsWebview = webview;
    ...
```
We should split this into a `outputsWebview` and `outputsWebviewPromise`.

msujew (commented):
Great change. I have a suggestion for the promise handling in the renderer, but otherwise this looks pretty good.

packages/notebook/src/browser/view/notebook-code-cell-view.tsx
```diff
@@ -149,36 +150,53 @@ interface NotebookCellOutputProps {
...
     }
 
+    override async componentWillUnmount(): Promise<void> {
+        this.toDispose.dispose();
+        (await this.outputsWebviewPromise)?.dispose();
+    }
+
     override render(): React.ReactNode {
-        return this.outputsWebview ?
+        return this.outputsWebview && !isThenable(this.outputsWebview) ?
```
msujew, line 199: Checking for the promise is no longer necessary

```suggestion
        return this.outputsWebview ?
```

msujew (approved):
Looks good to me 👍

PR #13242, opened 2024-01-05
fix: update message for missing Electron main entries

sdirix:
The warning message for missing Electron main entries still refered to `src-gen/frontend`. However since Theia 1.39.0 it should be either `src-gen/backend` or as now indicated `lib/backend`.

#### What it does

Update the warning message emitted by the Theia CLI when starting Theia Electron whose `package.json` does not contain a `main` entry.

#### How to test

Remove the `main` entry of the example Electron application and start it.

#### Review checklist

#### Reminder for reviewers


msujew (approved):
Thanks, that makes more sense 👍

Ebaome (approved):

PR #13254, opened 2024-01-10
Memory Leak when using MessageService#showProgress on Backend #13253

### LINKED ISSUE #13253: Memory Leak when using MessageService#showProgress on Backend
### Bug Description:

There is a memory leak when using e.g. `MessageService#showProgress` on the backend when the `Progress` is updated many times. 

In `RpcProtocol` new event listeners are added via `cancellationToken?.onCancellationRequested(() => this.sendCancel(id))` on each update, but those are not removed again. 

This will result in these kind of warnings on the backend log:

```
2024-01-10T11:24:18.341Z root WARN Possible Emitter memory leak detected. 176 listeners added. Use event.maxListeners to increase the limit (175). MOST frequent listener (1):
2024-01-10T11:24:18.341Z root WARN     at _event.Object.assign.maxListeners (/home/user/git/theia/examples/electron/lib/backend/packages_core_lib_common_index_js-node_modules_vscode-languageserver-types_lib_umd_sync_recursive.js:1430:54)
    at RpcProtocol.sendRequest (/home/user/git/theia/examples/electron/lib/backend/packages_core_lib_common_index_js-node_modules_vscode-languageserver-types_lib_umd_sync_recursive.js:3546:101)
    at /home/user/git/theia/examples/electron/lib/backend/packages_core_lib_common_index_js-node_modules_vscode-languageserver-types_lib_umd_sync_recursive.js:4236:58
    at new Promise (<anonymous>)
    at /home/user/git/theia/examples/electron/lib/backend/packages_core_lib_common_index_js-node_modules_vscode-languageserver-types_lib_umd_sync_recursive.js:4229:64
```

### Steps to Reproduce:

1. Execute this code on the backend:

```
    @inject(MessageService)
    protected readonly messages: MessageService;

    async init(): Promise<void> {
        const progress = await this.messages.showProgress({ text: 'Progressing...', options: { cancelable: false } });
        for (let i = 1; i <= 1000; i++) {
            progress.report({ message: undefined, work: { done: i, total: 1000 } });
            await wait(50); // import { wait } from '@theia/core/lib/common/promise-util';
        }
    }
```

(e.g. temporarily add it to `packages/git/src/node/init/git-init.ts` at the start of the `init()` method, which already uses the `MessageService`)

2. Let this run for some time and check that there are multiple logs about the leak. 

### Additional Information

- Theia Version: 1.45.0

jfaltermeier:
#### What it does

* dispose cancellation event listeners in `RpcProtocol`

Fixes #13253

#### How to test

See reproducer in #13253

#### Follow-ups

/

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/message-rpc/rpc-protocol.ts
+++ b/packages/core/src/common/message-rpc/rpc-protocol.ts
@@ -46,6 +46,31 @@ export interface RpcProtocolOptions {
 }
 
 /**
+ * Wrapper for a {@link Disposable} that is not available immediately.
+ */
+export class MaybeDisposable {
+
+    private disposed = false;
+    private disposable: Disposable | undefined = undefined;
+
+    setDisposable(disposable: Disposable): void {
+        if (this.disposed) {
+            disposable.dispose();
+        } else {
+            this.disposable = disposable;
+        }
+    }
+
+    dispose(): void {
+        this.disposed = true;
+        if (this.disposable) {
+            this.disposable.dispose();
+            this.disposable = undefined;
+        }
+    }
+}
+
+/**
  * Establish a RPC protocol on top of a given channel. By default the rpc protocol is bi-directional, meaning it is possible to send
  * requests and notifications to the remote side (i.e. acts as client) as well as receiving requests and notifications from the remote side (i.e. acts as a server).
  * Clients can get a promise for a remote request result that will be either resolved or
@@ -57,6 +82,7 @@ export class RpcProtocol {
     static readonly CANCELLATION_TOKEN_KEY = 'add.cancellation.token';
 
     protected readonly pendingRequests: Map<number, Deferred<any>> = new Map();
+    protected readonly pendingRequestCancellationEventListeners: Map<number, MaybeDisposable> = new Map();
 
     protected nextMessageId: number = 0;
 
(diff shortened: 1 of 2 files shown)


packages/core/src/common/message-rpc/rpc-protocol.ts
```diff
@@ -45,6 +45,31 @@ export interface RpcProtocolOptions {
     mode?: 'default' | 'clientOnly' | 'serverOnly'
 }
 
+/**
+ * Wrapper for a {@link Disposable} that is not available immediately.
+ */
+export class MaybeDisposable {
```
martin-fleck-at, line 51: I really like the idea of this class and I believe we can even add it to the common `disposable.ts`. From my understanding it is a wrapper that you mark as disposed and either the already given disposable or any later set disposable is then also disposed, making sure that whatever is given to the wrapper has the correct disposed state.

So I think that this class should actually also implement `Disposable` so it actually acts more like a wrapper fulfilling the same interface. Also, from the name alone it is not very clear what this class does, so maybe something like `DelayedDisposable` or actually `WrappedDisposable` or `DisposableWrapper` might be better. I believe we could even see this as a special case of the `DisposableCollection` with an adapted `push` that does not add things to the collection once dispose was called but instead simply disposes anything that is added to it, so maybe we could also adapt it to that. That would also solve the problem of people calling `setDisposable` multiple times without knowing whether we already wrapped a disposable or not.

However, I'll leave it up to you, I just think we should share it in a more common API and fix the naming a bit.

martin-fleck-at (changes requested):
Change looks good to me and works great, thank you Johannes! I just have a minor issue with the naming but then we can merge this.

packages/core/src/common/message-rpc/rpc-protocol.ts
jfaltermeier: I'll move it to `disposable.ts`. I am not sure about `DisposableCollection` though. If something like this is needed with a collection I would rather expect people placing the wrapper inside a disposable collection, which is possible now with implementing `Disposable`.

martin-fleck-at: Sure, I'm fine with that.

martin-fleck-at (approved):
Changes look good to me now, thank you very much Johannes!

Issue #12874, opened 2023-08-29
nls: add missing `removeSession` localization when signing out


vince-fugnitto:
#### Feature Description:

There is a missing signing out localization for the `removeSession` method:

https://github.com/eclipse-theia/theia/blob/b3e8d425155eb92bb8a68773b2b7bb2f9584183b/packages/plugin-ext/src/main/browser/authentication-main.ts#L362-L365

We can benefit from the default localization:

```diff
removeSession(sessionId: string): Thenable<void> {
    return this.proxy.$removeSession(this.id, sessionId)
-       .then(() => { this.messageService.info('Successfully signed out.'); });
+       .then(() => this.messageService.info(nls.localizeByDefault('Successfully signed out.')));
}
```

VanshSutariya:
I'd like to work on this issue , can you assign me this issue or I just raised the PR

PR #13257, opened 2024-01-10
Fix Localization for the removeSession method (#12874)

dhuebner:
Add a missing signing out localization for the removeSession method as described in  #12874


msujew (approved):
Thanks, looks good to me 👍

dhuebner:
@msujew 
Could you also close https://github.com/eclipse-theia/theia/issues/12874 please?

PR #13261, opened 2024-01-11
SelectComponent does not render droplist correctly in dialog (#12895)

### LINKED ISSUE #12895: SelectComponent does not render droplist correctly when embedded in a dialog
### Bug Description:
I am using the SelectComponent component in a dialog, the droplist content is hidden behind the dialog area.

### Steps to Reproduce:

1. Create a dialog with a SelectComponent component
2. Click on the arrow to expand the droplist option

### Additional Information

The issue is that the dialog overlay has an z-index of 5000 and the select-component-container does not have any z-index set.
By providing these two CSS rules to the select-component-container element, the droplist option is show correctly within the dialog and the setting page.

```CSS
.select-component-container {
    position: fixed; /* required to set z-index */
    z-index: 6000;   /* dialog overlay has a z-index of 5000 */
}
```

- Theia Version: 1.40

dhuebner:
#### What it does
Fixes #12895

#### How to test
I've added a new "Sample React Dialog" as there were any samples that are using react. Open it using the "Sample React Dialog" command.

Before:
Bildschirmfoto 2024-01-11 um 10 57 45

After:

Bildschirmfoto 2024-01-11 um 10 56 27

#### Follow-ups
I can remove the sample dialog again if you don't want to have it.
Also, I don't  know why, but the *.css file was auto formatted with a probably different rules as it was initially created.

### DIFF
--- a/examples/api-samples/src/browser/menu/sample-menu-contribution.ts
+++ b/examples/api-samples/src/browser/menu/sample-menu-contribution.ts
@@ -14,12 +14,16 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { ConfirmDialog, QuickInputService } from '@theia/core/lib/browser';
+import { ConfirmDialog, Dialog, QuickInputService } from '@theia/core/lib/browser';
+import { ReactDialog } from '@theia/core/lib/browser/dialogs/react-dialog';
+import { SelectComponent } from '@theia/core/lib/browser/widgets/select-component';
 import {
     Command, CommandContribution, CommandRegistry, MAIN_MENU_BAR,
     MenuContribution, MenuModelRegistry, MenuNode, MessageService, SubMenuOptions
 } from '@theia/core/lib/common';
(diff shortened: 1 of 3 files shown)


examples/api-samples/src/browser/menu/sample-menu-contribution.ts
```diff
@@ -50,6 +54,10 @@ const SampleQuickInputCommand: Command = {
     category: 'Quick Input',
     label: 'Test Positive Integer'
 };
+const SampleReactDialog: Command = {
+    id: 'sample-command-react-dialog',
+    label: 'Sample React Dialog'
+};
```
msujew, line 60: Minor: Since this is specifically to test the select component, we should give this a better name
```suggestion
const SampleSelectDialog: Command = {
    id: 'sample-command-select-dialog',
    label: 'Sample Select Component Dialog'
};
```

msujew (approved):
Thanks, the change looks good to me 👍 

One minor issue, see below. Afterwards, this is good to go.

dhuebner:
@msujew 
Thank you Mark. Do you know why the git tests are failing?

msujew:
@dhuebner You need to adjust the references to the old `SampleReactDialog`, which was renamed.

PR #12836, opened 2023-08-16
Implement VS Code tree view checkbox API.

tsmaeder:
#### What it does
Implements the ability to have check boxes on tree views in plugins

Fixes #12695

Contributed on behalf of STMicroelectronics

#### How to test
Install the attached VS Code extension. It contributes a a view called "Checkbox API Test". Open the view if necessary. Play around with checking/unchecking the checkboxes. The system automatically checks/unchecks parent and child checkboxes: see the VS Code doc linked in the issue for details.
The extension also contributes a preference "Manage Checkbox State Manually". When this preference is on, the automatic checking/unchecking is disabled. The extensions writes to the log, when it is notified of checkbox state changes. There is also a context menu action to check/uncheck the tree items programmatically. 

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #13271, opened 2024-01-15
Add keybindings to toggle the tree checkbox

### LINKED ISSUE #13263: Can't Toggle Checkbox on Tree Node using Keyboard
### Feature Description:
While the tree implementation is very easy to traverse with just keyboard inputs, I do not seem to be able to toggle checkboxes attached to tree nodes with the keyboard.  I'm suggesting some kind of key input to toggle the checkbox of all currently selected tree nodes.

haydar-metin:
#### What it does

Allows the user to toggle the checkbox on a tree node using the keyboard.

Fixes #13263

#### How to test

1. Download the VSIX 
[checkbox-keyboard-toggle-0.0.1.zip](https://github.com/eclipse-theia/theia/files/13936129/checkbox-keyboard-toggle-0.0.1.zip)
https://github.com/haydar-metin/theia-playground/tree/issues/13263 
2. In the explorer view you should have a new tree view 
image
3. With `space` you can toggle the checkbox

The current behavior of collapsed entires is defined here:
https://github.com/eclipse-theia/theia/blob/master/packages/plugin/src/theia.d.ts#L6055

`If the tree item is collapsed by default (meaning that the children haven't yet been fetched) then child checkboxes will not be updated.`

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/tree/tree-widget.tsx
+++ b/packages/core/src/browser/tree/tree-widget.tsx
@@ -1227,12 +1227,15 @@ export class TreeWidget extends ReactWidget implements StatefulWidget {
 
     /**
      * Handle the `space key` keyboard event.
-     * - By default should be similar to a single-click action.
+     * - If the element has a checkbox, it will be toggled.
+     * - Otherwise, it should be similar to a single-click action.
      * @param event the `space key` keyboard event.
      */
     protected handleSpace(event: KeyboardEvent): void {
         const { focusedNode } = this.focusService;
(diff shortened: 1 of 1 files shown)


martin-fleck-at (approved):
Thank you very much for that change @haydar-metin! Everything looks good to me. 

I also tested this with a tree widget (as opposed to tree view widget coming from the VS Code API) and I noticed that the children/parent toggle management does not exist there. Probably that is on purpose since the option is only available for view widgets (see https://github.com/eclipse-theia/theia/pull/12836) but I just wanted to note that down somewhere.

PR #13281, opened 2024-01-17
Resize terminal process after resizing widget

### LINKED ISSUE #13174: Terminal starts with incorrect size
### Steps to Reproduce:

1. Open a terminal
2. Try to write a very long line
3. Notice that it wraps at 80 characters even though the terminal is much wider
4. You can also run `tput cols` and `tput lines` to see the size
5. Resize (just wiggle slightly)
6. Try to write a very long line
7. Notice that it now wraps correctly

Note that in addition to the size being incorrect there is also a strange wrapping behavior that occurs, where it overwrite the same line instead of wrapping to the next line. This behavior is also fixed after a resize.

After some debugging I found the root problem to be that during start of the terminal widget, `resizeTerminal` is called together with `resizeTerminalProcess`. But `resizeTerminal` is the debounced `doResizeTerminal`. So `resizeTerminalProcess` actually runs first and before the actual size is updated.

I managed to fix the problem for me using the following:

```ts
@injectable()
export class TerminalWidgetImplFix extends TerminalWidgetImpl {
    protected doResizeTerminal(): void {
        super.doResizeTerminal();
        this.resizeTerminalProcess();
    }
}
```

and

```ts
rebind(TerminalWidget).to(TerminalWidgetImplFix);
```

I haven't tried if this problem also occurs on master, but I quickly looked at the code changes and didn't see anything that would fix it.

### Additional Information

- Operating System: Linux/Browser
- Theia Version: 1.43.1

Lythenas:
#### What it does

Fixes #13174

Resize the terminal process in the debounced `doResizeTerminal` (after the terminal widget has the new size) instead of directly in the `onUpdateRequest`.

Contributed on behalf of Elektrobit Automotive GmbH.

#### How to test

1. Start theia, open it in a browser
    - It's a bit inconsistent to reproduce, since it is a race condition. I found that setting a breakpoint on `packages/terminal/src/browser/terminal-widget-impl.ts:604` (i.e. in `onUpdateRequest`) and stepping through always reproduces the issue.
2. Open a terminal
3. Run `tput cols` and `tput lines`, see that it is the incorrect size (defaults to 80 and 24, instead of the real size of the terminal widget after it is fix into the view)
4. Try to write a very long line, see that it wraps as 80 characters even though there is still space left
5. Resize the terminal (just wiggling is fine)
6. Do step 3 and 4 again and see that it now shows the correct size and behaves correctly

With this PR step 3 and 4 should behave correctly. I.e. the size of the terminal is correct after opening, without having to resize.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/terminal/src/browser/terminal-widget-impl.ts
+++ b/packages/terminal/src/browser/terminal-widget-impl.ts
@@ -611,8 +611,6 @@ export class TerminalWidgetImpl extends TerminalWidget implements StatefulWidget
         if (this.needsResize) {
             this.resizeTerminal();
             this.needsResize = false;
-
-            this.resizeTerminalProcess();
         }
     }
 
@@ -797,6 +795,8 @@ export class TerminalWidgetImpl extends TerminalWidget implements StatefulWidget
         const cols = geo.cols;
         const rows = geo.rows - 1; // subtract one row for margin
         this.term.resize(cols, rows);
+
+        this.resizeTerminalProcess();
     }
 
     protected resizeTerminalProcess(): void {


msujew (approved):
Thanks a lot for the fix @Lythenas 👍

Issue #13269, opened 2024-01-12
Install vsix files from the explorer view


dhuebner:
### Feature Description:
Right now there is an "Install from VSIX..." menu item in the "Extensions" view.
It would be handy to have a new "Install Extension VSIX" menu item also in the explorer context menu. This entry should only be active when `*.vsix` file is selected. 

<img width="638" alt="Bildschirmfoto 2024-01-12 um 15 40 20" src="https://github.com/eclipse-theia/theia/assets/454322/6492ba3b-2cc6-45a2-9d7b-69b6c0911ed6">

dhuebner:
I could prepare a PR if the community is interested in having this feature.

msujew:
@dhuebner I think that makes sense 👍

dhuebner:
@msujew 
This feature request can be closed as fixed with #13291

PR #13291, opened 2024-01-19
Install vsix files from the explorer view (#13269)

dhuebner:
#### What it does

Allows installing VSIX extensions from the file explorer using context menu.

- Contributes a new command that uses selection service to get the *.vsix file instead of opening a dialog
- Adds a `@theia/navigator`  dependency to get the navigator menu ID (existing dependencies like `@theia/plugin-ext` already depends on the navigator package) 
- Refresh 'Installed' tree entries when extension view becomes visible, otherwise one will see updated extension count (badge), but not the new installed entry.  
- Other fixes: Added explicit `canSelectFiles: true` to dialog options, otherwise one can't select files in the electron app. Could be a problem with default handling in `ElectronFileDialogService.toOpenDialogOptions()`

#### How to test

Add a `*.vsix` to your workspace. Use the context menu to execute `Install Extension VSIX` action. You should see the 'Successfully installed' message. Go to the "Extensions" view and check the "Installed" list contains the new installed extension.

#### Follow-ups

When testing extension installation I noticed that when installing an already installed extension a `Failed to install` error is reported. This is because local file deployer rejects installing existing extensions, but because the installing routine can only check the number of installed extension (in this case it's `0`) a misleading error is reported. 
I think we need to change the installation routine return value to something more meaningful than just a number. With more information we can then to produce a better user feedback.  See ` PluginDeployerImpl.deployPlugins()`.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/vsx-registry/package.json
+++ b/packages/vsx-registry/package.json
@@ -5,6 +5,7 @@
   "dependencies": {
     "@theia/core": "1.46.0",
     "@theia/filesystem": "1.46.0",
+    "@theia/navigator": "1.46.0",
     "@theia/ovsx-client": "1.46.0",
     "@theia/plugin-ext": "1.46.0",
     "@theia/plugin-ext-vscode": "1.46.0",
--- a/packages/vsx-registry/src/browser/vsx-extension-commands.ts
+++ b/packages/vsx-registry/src/browser/vsx-extension-commands.ts
@@ -36,6 +36,11 @@ export namespace VSXExtensionsCommands {
         label: nls.localizeByDefault('Install from VSIX') + '...',
         dialogLabel: nls.localizeByDefault('Install from VSIX')
     };
+    export const INSTALL_VSIX_FILE: Command = Command.toDefaultLocalizedCommand({
+        id: 'vsxExtensions.installVSIX',
+        label: 'Install Extension VSIX',
+        category: EXTENSIONS_CATEGORY,
+    });
     export const INSTALL_ANOTHER_VERSION: Command = {
         id: 'vsxExtensions.installAnotherVersion'
     };
--- a/packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
+++ b/packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
@@ -14,29 +14,32 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { DateTime } from 'luxon';
-import { injectable, inject, postConstruct } from '@theia/core/shared/inversify';
-import debounce = require('@theia/core/shared/lodash.debounce');
-import { Command, CommandRegistry } from '@theia/core/lib/common/command';
-import { AbstractViewContribution } from '@theia/core/lib/browser/shell/view-contribution';
-import { VSXExtensionsViewContainer } from './vsx-extensions-view-container';
-import { VSXExtensionsModel } from './vsx-extensions-model';
+import { CommonMenus, LabelProvider, PreferenceService, QuickInputService, QuickPickItem } from '@theia/core/lib/browser';
+import { ClipboardService } from '@theia/core/lib/browser/clipboard-service';
 import { ColorContribution } from '@theia/core/lib/browser/color-application-contribution';
 import { ColorRegistry } from '@theia/core/lib/browser/color-registry';
-import { Color } from '@theia/core/lib/common/color';
 import { FrontendApplication } from '@theia/core/lib/browser/frontend-application';
 import { FrontendApplicationContribution } from '@theia/core/lib/browser/frontend-application-contribution';
-import { MenuModelRegistry, MessageService, nls } from '@theia/core/lib/common';
+import { AbstractViewContribution } from '@theia/core/lib/browser/shell/view-contribution';
+import { MenuModelRegistry, MessageService, SelectionService, nls } from '@theia/core/lib/common';
+import { Color } from '@theia/core/lib/common/color';
+import { Command, CommandRegistry } from '@theia/core/lib/common/command';
+import URI from '@theia/core/lib/common/uri';
+import { UriAwareCommandHandler } from '@theia/core/lib/common/uri-command-handler';
+import { inject, injectable, postConstruct } from '@theia/core/shared/inversify';
 import { FileDialogService, OpenFileDialogProps } from '@theia/filesystem/lib/browser';
-import { LabelProvider, PreferenceService, QuickPickItem, QuickInputService, CommonMenus } from '@theia/core/lib/browser';
+import { NAVIGATOR_CONTEXT_MENU } from '@theia/navigator/lib/browser/navigator-contribution';
+import { OVSXApiFilter, VSXExtensionRaw } from '@theia/ovsx-client';
 import { VscodeCommands } from '@theia/plugin-ext-vscode/lib/browser/plugin-vscode-commands-contribution';
-import { VSXExtensionsContextMenu, VSXExtension } from './vsx-extension';
-import { ClipboardService } from '@theia/core/lib/browser/clipboard-service';
-import { BUILTIN_QUERY, INSTALLED_QUERY, RECOMMENDED_QUERY } from './vsx-extensions-search-model';
-import { IGNORE_RECOMMENDATIONS_ID } from './recommended-extensions/recommended-extensions-preference-contribution';
-import { VSXExtensionsCommands } from './vsx-extension-commands';
-import { VSXExtensionRaw, OVSXApiFilter } from '@theia/ovsx-client';
+import { DateTime } from 'luxon';
 import { OVSXClientProvider } from '../common/ovsx-client-provider';
+import { IGNORE_RECOMMENDATIONS_ID } from './recommended-extensions/recommended-extensions-preference-contribution';
+import { VSXExtension, VSXExtensionsContextMenu } from './vsx-extension';
+import { VSXExtensionsCommands } from './vsx-extension-commands';
+import { VSXExtensionsModel } from './vsx-extensions-model';
+import { BUILTIN_QUERY, INSTALLED_QUERY, RECOMMENDED_QUERY } from './vsx-extensions-search-model';
+import { VSXExtensionsViewContainer } from './vsx-extensions-view-container';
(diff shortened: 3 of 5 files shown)


packages/vsx-registry/package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@theia/core": "1.45.0",
     "@theia/filesystem": "1.45.0",
+    "@theia/navigator": "1.45.0",
```
martin-fleck-at, line 8: Conflict cause we are now at `1.46.0`

packages/vsx-registry/src/browser/vsx-extension-commands.ts
```diff
@@ -36,6 +36,13 @@ export namespace VSXExtensionsCommands {
         label: nls.localizeByDefault('Install from VSIX') + '...',
         dialogLabel: nls.localizeByDefault('Install from VSIX')
     };
+    export const INSTALL_VSIX: Command = {
```
martin-fleck-at, line 39: nit: I think we could use `Command.toDefaultLocalizedCommand` here.

martin-fleck-at, line 39: I'm worried that the const name may make it easy to confuse this with `INSTALL_FROM_VSIX` but I understand it is hard to find a name without changing the existing ones. What do you think of `INSTALL_VSIX_FILE`?

Similar to the label, I'd prefer `Install VSIX extension` or just `Install Extension` with the `...` since it is a longer running operation but maybe the dots can be skipped (not sure if we only use them when a dialog opens).

packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
```diff
@@ -199,25 +214,33 @@ export class VSXExtensionsContribution extends AbstractViewContribution<VSXExten
...
             } else {
                 this.messageService.error(nls.localize('theia/vsx-registry/invalidVSIX', 'The selected file is not a valid "*.vsix" plugin.'));
             }
         }
     }
 
+    /**
+     * Installs a local extension file.
+     */
+    protected async installVSIX(fileURI: URI): Promise<void> {
```
martin-fleck-at, line 233: I think we should also move the guarding into this method, just to avoid any obvious errors:

```
if (extensionUri.path.ext === '.vsix') {
   ...
}
} else {
   this.messageService.error(nls.localize('theia/vsx-registry/invalidVSIX', 'The selected file is not a valid "*.vsix" plugin.'));
}
```

What do you think?

packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
```diff
@@ -95,6 +99,12 @@ export class VSXExtensionsContribution extends AbstractViewContribution<VSXExten
             execute: () => this.installFromVSIX()
         });
 
+        commands.registerCommand(VSXExtensionsCommands.INSTALL_VSIX,
+            UriAwareCommandHandler.MonoSelect(this.selectionService, {
+                execute: fileURI => this.installVSIX(fileURI)
+            }
```
martin-fleck-at, line 105: nit: formatting

packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
```diff
@@ -95,6 +99,12 @@ export class VSXExtensionsContribution extends AbstractViewContribution<VSXExten
             execute: () => this.installFromVSIX()
         });
 
+        commands.registerCommand(VSXExtensionsCommands.INSTALL_VSIX,
+            UriAwareCommandHandler.MonoSelect(this.selectionService, {
+                execute: fileURI => this.installVSIX(fileURI)
```
martin-fleck-at, line 104: Since we already know this will only work on *.vsix files, we should add the `isEnabled` (and maybe `isVisible`) condition here or was there a particular reason you only restricted that on the menu action?

packages/vsx-registry/src/browser/vsx-extensions-widget.tsx
```diff
@@ -153,4 +153,13 @@ export class VSXExtensionsWidget extends SourceTreeWidget implements BadgeWidget
         }
         return super.renderTree(model);
     }
+
+    protected override onAfterShow(msg: Message): void {
+        super.onAfterShow(msg);
+        if (this.options.id === VSXExtensionsSourceOptions.INSTALLED) {
```
martin-fleck-at, line 159: This works when closing and re-opening the view. Unfortunately it seems that we do not have a notification system for when extensions are deployed, at least none that I can see and implementing one is probably out of scope here.

martin-fleck-at (changes requested):
Thank you very much @dhuebner! I think this is a really nice contribution and it works well! I do have a few minor comments regarding naming and visibility, I'd be interested to get your take on that before we merge this change.

packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
dhuebner: As  this method can be used only programmatically and *.vsix is just a zip file I didn't add a file extension guard here. Docu comment says 'Installs a local extension file.'  so theoretically it can be used to install from a Zip file with any extension.

packages/vsx-registry/src/browser/vsx-extension-commands.ts
dhuebner: `INSTALL_VSIX_FILE` might be a better choice, I like it. 
Regarding the label, I tend to use the same as vs-code uses, so we can use the existing translation. 
And yes, `...` are normally for dialogs where an additional user action is required. Here it is not the case.

packages/vsx-registry/package.json
dhuebner: Thanks! I probably need to rebase...

packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
dhuebner: Or you mean someone can execute it from the command palette? This is a good point, I need to check it.

packages/vsx-registry/src/browser/vsx-extension-commands.ts
dhuebner: Not sure if we should, I need to check it. Thanks!

martin-fleck-at: If there is already a label for that in VS Code then let's go with that! I must have overlooked it, thank you for clarifying!

packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
martin-fleck-at: Didn't think about that but you are right, if no other code is testing the extension a simple zip might also work. Our other actions do restrict this though so if we keep it like that (at which I'm not against), please add a short comment explaining why we do not do the check here.

packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
martin-fleck-at: Yes. If this is only supposed to be called programmatically, I think you can hide the command by not providing a label and only provide the label in the menu action. Again, if that is the purpose please add a short comment because it is definitely non-obvious ;-)

packages/vsx-registry/src/browser/vsx-extension-commands.ts
dhuebner: I see, `toDefaultLocalizedCommand` does the job for me. Thanks!

packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
dhuebner: This is, unfortunately, the formatted code. I can fix it, but next auto format will probably reformat it again...

martin-fleck-at: Oh, then never mind! I thought it just looked off but it's definitely better to keep the automated way ;-)

packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
dhuebner: I was thinking about using the command from the command palette and I like the idea.
I've added `isEnabled`  to the command registration, so now one can install vsix files from the explorer view without using a mouse. I know there are "keyboard first" fans out there and they will like it, I hope.

packages/vsx-registry/src/browser/vsx-extensions-widget.tsx
dhuebner: > we do not have a notification system for when extensions are deployed

Yeah, I also found nothing there.
An other point I mentioned in the description is that, it would be nice to have a possibility to check if an extension is already installed. And what is even more important, to have a detailed return value when attempting to install an extension. Right now it is just a number of extensions being installed and if it is `0` an error is reported.

packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
dhuebner: After rewriting the code it formats better now.

dhuebner:
@martin-fleck-at 
Thank you very much for your review, it was really helpful!
I changed all the related places and think the PR is now even better :)

packages/vsx-registry/src/browser/vsx-extensions-widget.tsx
martin-fleck-at: I fully agree, there is definitely room for improvement on the exposure of extension management. I remember there being a discussion somewhere about being notified properly for uninstallations but I can't find it right now. Do you mind logging a Feature Request for the extension of that API?

dhuebner:
@martin-fleck-at 
Are here some remaining changes to be made?
Do you mind to look into my other PR #13344 as well? It is also kind of files related...

packages/vsx-registry/src/browser/vsx-extensions-widget.tsx
dhuebner: Sure, will do.

martin-fleck-at (approved):
Thank you very much for the update! Everything looks good to me now!

martin-fleck-at:
> @martin-fleck-at Are here some remaining changes to be made? Do you mind to look into my other PR #13344 as well? It is also kind of files related...

Sure, I can have a look.

PR #13299, opened 2024-01-21
Update plugin API docs for headless plugins and Inversify DI

cdamus:
#### What it does

Update the plugin API documentation to cover

- headless plugins (new since #12946 was implemented)
- Inversify DI in the plugin host (included in the implementation of headless plugin support)
- simplified ext API initialization and API factory using new PluginContainerModule capability (included in the implementation of DI in the plugin host)

#### How to test

Read and critique the updated documentation pages at

- [Plugin API doc](https://github.com/cdamus/theia/blob/issues/13290-plugin-docs/doc/Plugin-API.md)
- [How to add custom API doc](https://github.com/cdamus/theia/blob/issues/13290-plugin-docs/packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md)

#### Follow-ups

No further documentation changes are identified at this time.

#### Review checklist

#### Reminder for reviewers


doc/Plugin-API.md
```diff
@@ -6,12 +6,16 @@ In the following, we focus on the mechanics of Theia plugins and Theia’s compa
 This documentation aims to support developers extending Theia’s plugin API to either enhance the extensibility of Theia via plugins and/or increase Theia’s coverage of the VS Code Extension API – and with that the number of VS Code extensions that can be used in Theia.
 
 Theia plugins, as well as VS Code extensions, can be installed and removed from a Theia installation at runtime and may extend many different capabilities of Theia, such as theming, language support, debuggers, tree views, etc., via a clearly defined API.
+A special category of "headless" plugins addresses more application-specific extension points, not intended to extend Theia's out-of-the-box capabilities.
+For the most part, the discussion in this document of the plugin API does not apply to these plugins but only to VS Code compatible plugins.
```
martin-fleck-at, line 10: Could we maybe exchange that with a more specific sentence that highlights the similarities or differences to the other plugins. It might be enough to state that that the Theia plugin API aims to be a superset VS Code extension API and is therefore more concerned with the user facing aspects of the application, whereas headless plugins are meant as a hook for application-specific API that only runs in the backend. We could even consider having a picture explaining how the Theia/VS Code plugins differ from the headless plugins if that helps.

I just feel that as such a vague statement makes a user doubt the completeness of this documentation a bit - and for no reason as you did a great job in extending the the different sections!

doc/Plugin-API.md
```diff
@@ -167,14 +176,45 @@ The `Ext` side then gets the cached object, executes appropriate functions and r
...
 
 New classes and other types such as enums are usually implemented in [plugin-ext/src/plugin/types-impl.ts](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/types-impl.ts).
 They can be added there and then we can add them to the API object created in the API factory.
 
+## Headless Plugins
+
+The majority of plugin use cases are for extension of the Theia user experience via the VS Code compatible API provided by Theia.
+These plugins use either the `vscode` API object if they use the `"vscode"` engine type in their package manifests or else the `theia` object if they use the `"theiaPlugin"` engine.
+The lifecycle of these kinds of plugins is bound to _frontend connections_: for each connected frontend, the Theia backend spawns a plugin host host process dedicated to it in which these plugins are loaded and activated (as applicable to their declared activation events).
+In the plugin host for a frontend connection these plugins have access to Theia services as discussed above, isolated from Theia itself and from all of the other instances of the same plugin running in plugin hosts for all other frontend connections.
```
martin-fleck-at, line 192: > isolated from Theia itself

I'm not sure I fully understand what you mean by that? Do you mean the Theia backend?

doc/Plugin-API.md
```diff
@@ -167,14 +176,45 @@ The `Ext` side then gets the cached object, executes appropriate functions and r
...
+
+Headless plugins, by contrast, are quite different in most respects:
+
+- they are encapsulated in a single plugin host process in the backend, separate from all frontend-connection plugin hosts.
```
martin-fleck-at, line 196:
```suggestion
- They are encapsulated in a single plugin host process in the backend, separate from all frontend-connection plugin hosts.
```

Should we start with a capital letter if we have a complete sentence? I'd also make sure that each sentense then ends with a dot.

doc/Plugin-API.md
```diff
@@ -167,14 +176,45 @@ The `Ext` side then gets the cached object, executes appropriate functions and r
...
+This host is spun up only if there are any headless plugins to run in it
```
martin-fleck-at, line 197:
```suggestion
This host is spun up only if there are any headless plugins to run in it.
```

See comment above and if you agree, please apply to all items.

doc/Plugin-API.md
```diff
@@ -167,14 +176,45 @@ The `Ext` side then gets the cached object, executes appropriate functions and r
...
+- Theia does not export any default API object, analogous to `vscode` or `theia` for other plugins, as Theia itself defines no use cases for headless plugins.
+Such use cases are entirely defined by the Theia-based application's requirements
```
martin-fleck-at, line 199: Might be interesting to add one sentence about how one can actually define such an API (are there interfaces I need to implement?).

doc/Plugin-API.md
```diff
@@ -167,14 +176,45 @@ The `Ext` side then gets the cached object, executes appropriate functions and r
...
+- Theia does not support any contribution points for headless plugins not any non-trivial activation events (only `'*'` and `'onStartupFinished'`).
+This is a corollary of the use cases being entirely application-specific: the application needs to define its own contribution points and activation events
```
martin-fleck-at, line 201: ... through what? How can I define contribution points and activation events?

doc/Plugin-API.md
```diff
@@ -167,14 +176,45 @@ The `Ext` side then gets the cached object, executes appropriate functions and r
...
+Alternatively, a VS Code or Theia plugin that extends the frontend user experience may also contribute a headless entrypoint for a headless deployment by identifying such entrypoint script in the `"headless"` property of the `"theiaPlugin"` object in its package manifest in addition to the `"main"` entrypoint (for VS Code plugins) or the `"theiaPlugin.backend"` entrypoint (for Theia plugins).
+
+The only API namespaces that are available to headless plugins are those custom APIs that are contributed by the application's custom build-time extensions or by the activation of other headless plugins.
+For details of how to contribute custom API, see the [pertinent documentation](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md).
+
+## Dependency Injection
+
+Both the `Main` and the `Ext` sides of the plugin API are configured using [InversifyJS](https://inversify.io) dependency injection.
+On the `Main` side, the usual Theia container modules are used to bind implementations of the API objects.
```
martin-fleck-at, line 214: Could you please add the specific symbol for that binding?

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -5,18 +5,24 @@ That will require API that goes beyond what's in the VS Code Extension API and t
 You can do that by implementing a Theia extension that creates and exposes an API object within the plugin host.
 The API object can be imported by your plugins and exposes one or more API namespaces.
 
-Depending on the plugin host we can either provide a frontend or backend plugin API:
+Depending on the plugin host we can either provide a frontend or backend plugin API, or an API for headless plugins that extend backend services that are not tied to any frontend connection:
```
martin-fleck-at, line 8: Is "extend" really the correct term here? A plugin may extend backend services or it may simply use them to perform whatever it needs to do. Maybe a more neutral term such as "has access to" or something like that may be more fitting?

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -5,18 +5,24 @@ That will require API that goes beyond what's in the VS Code Extension API and t
...
 
 - In the backend plugin host that runs in the Node environment in a separate process, we adapt the module loading to return a custom API object instead of loading a module with a particular name.
+There is a distinct plugin host for each connected Theia frontend.
 - In the frontend plugin host that runs in the browser environment via a web worker, we import the API scripts and put it in the global context.
+There is a distinct plugin host for each connected Theia frontend.
+- In the headless plugin host that also runs in the Node environment in a separate process, we similarly adapt the module loading mechanism.
+There is only one plugin host for headless plugins, if there are any headless plugins deployed.
```
martin-fleck-at, line 15: Does that also imply that headless plugins cannot be installed during runtime but only during startup/deployment time?

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -30,7 +36,9 @@ export class FooExtPluginApiProvider implements ExtPluginApiProvider {
                 initFunction: 'fooInitializationFunction',
                 initVariable: 'foo_global_variable'
             },
-            backendInitPath: path.join(__dirname, 'foo-init')
+            backendInitPath: path.join(__dirname, 'foo-init'),
+            // (Optional) provide the same API to headless plugins, too (or a different/subset API)
```
martin-fleck-at, line 40: Not sure I want to have the `(Optional)` here as it seems that otherwise you need to provide the other two when in fact all of them are actually optional.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -62,94 +70,70 @@ declare module '@bar/foo' {
 ## Implement your plugin API provider
 
 In our example, we aim to provide a new API object for the backend.
-Theia expects that the `backendInitPath` that we specified in our API provider is a function called `provideApi` that follows the `ExtPluginApiBackendInitializationFn` signature.
+Theia expects that the `backendInitPath` that we specified in our API provider exports an [InversifyJS](https://inversify.io) `ContainerModule` under the name `containerModule`.
+This container-module configures the Inversify `Container` in the plugin host for creation of our API object.
+It also implements for us the customization
```
martin-fleck-at, line 75: It seems this sentence was cut off?

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -62,94 +70,70 @@ declare module '@bar/foo' {
...
+For export of the API to headless plugins, use the `headlessInitPath` with a `provideApi` function of `ExtPluginApiHeadlessInitializationFn` type.
```
martin-fleck-at, line 76: Not sure if we need the type info here since we do not show the same level of detail for the other plugin types, especially not the backendInitPath which is the one we actually showcase.

martin-fleck-at (commented):
Thank you very much for that great update, Christian! I have a few comments and may also ask other people to chime. Really grateful you also fixed some of the typos ;-)

JonasHelming:
Should this maybe also be mentioned here: https://theia-ide.org/docs/extensions/

martin-fleck-at:
@JonasHelming Very good point! I fully agree that we should also mention in on the website doc!

doc/Plugin-API.md
cdamus: There are different styles for interpretation of bullet lists as complete sentences or not. I'll try harder to follow the established pattern in these docs!

doc/Plugin-API.md
cdamus: Which specific symbol? There are loads of symbols that can be bound. What I'm trying to get at with this statement is that support for Inversify DI on the Main side is exactly like all of the other DI in Theia, using ContainerModules registered in package manifests and loaded at start-up by a code-generated script. Maybe I should just state it in those terms in the doc 😀

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
cdamus: I suppose a plugin could be deployed that is just activated on start-up and then only calls into Theia's APIs, never actually "extending" some service. But even that is an extension (qua expansion) of Theia's run-time behaviour. This is an _extension_ mechanism, after all. 😝 

I'll try to revise for clarity.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
cdamus: No, if they are deployed later, then the plugin-host will start at that time. Will clarify.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
cdamus: One too many times on the ⌘Z, methinks.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
cdamus: This is a relic from before my second pass that rewrote the example in terms of the new `ContainerModule` mechanism that supersedes the `provideApi` function!

cdamus:
Thanks @martin-fleck-at for the thorough review. Very helpful! I've pushed commit 7063d6e7b09b3003e7d18b9c072bfe0f350841d8 that I hope should address your comments. Updates to the website will require a separate PR on that repository.

cdamus:
> Should this maybe also be mentioned here: https://theia-ide.org/docs/extensions/

Done: https://github.com/eclipse-theia/theia-website/pull/502

jcortell68:
@cdamus Great job with this new documentation. With the new plugin type in the mix, this stuff is even trickier to explain than it was before. You did so as best as it could have been done, IMO.

doc/Plugin-API.md
```diff
@@ -167,14 +187,50 @@ The `Ext` side then gets the cached object, executes appropriate functions and r
...
+In the plugin host for a frontend connection these plugins have access to Theia services as discussed above, isolated from the main Theia backend process and from all of the other instances of the same plugin running in plugin hosts for all other frontend connections.
+
+Headless plugins, by contrast, are quite different in most respects:
+
+- They are encapsulated in a single plugin host process in the backend, separate from all frontend-connection plugin hosts.
+This host is spun up only if there are any headless plugins to run in it.
+- Theia does not export any default API object, analogous to `vscode` or `theia` for other plugins, as Theia itself defines no use cases for headless plugins.
+Such use cases are entirely defined by the Theia-based application's requirements and reflected in its custom APIs defined [as described in this how-to document][custom-api-howto].
+- Theia does not support any contribution points for headless plugins not any non-trivial activation events (only `'*'` and `'onStartupFinished'`).
```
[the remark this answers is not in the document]
cdamus: `sed -E s/not/nor/`

doc/Plugin-API.md
```diff
@@ -167,14 +187,50 @@ The `Ext` side then gets the cached object, executes appropriate functions and r
...
+
+Thus, headless plugins are best suited to the contribution of third-party extensions of a Theia application's custom backend services, where those service are shared by all connected frontends or serve some kind of headless scenario like a CLI.
+
+A headless plugin may be restricted to only the headless deployment, in which case it may make this explicit by declaring the `"theiaHeadlessPlugin"` engine in its package manifest.
+Alternatively, a VS Code or Theia plugin that extends the frontend user experience may also contribute a headless entrypoint for a headless deployment by identifying such entrypoint script in the `"headless"` property of the `"theiaPlugin"` object in its package manifest in addition to the `"main"` entrypoint (for VS Code plugins) or the `"theiaPlugin.backend"` entrypoint (for Theia plugins).
+
+The only API namespaces that are available to headless plugins are those custom APIs that are contributed by the application's custom build-time extensions or by the activation of other headless plugins.
```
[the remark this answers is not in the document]
cdamus: There's also the return result of the `activate()` function that publishes an API to the other plugins in the host. I can try to clarify this.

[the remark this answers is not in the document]
jcortell68: Ah, yes. I'd forgotten about that! But yes. that might need to be spelled out more given I think most people don't know about that feature of `activate()`.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -62,94 +72,69 @@ declare module '@bar/foo' {
...
-        }
+const FooBarApiFactory = Symbol('FooBarApiFactory');
+type FooBarApiFactory = ApiFactory<FooBarApi>;
 
-        return defaultApi;
-    };
-}
+// Retrieved by Theia to configure the Inversify DI container when the plugin is initialized.
+// This is called when the plugin-host process is forked.
+export const containerModule = PluginContainerModule.create(({ bind, bindApiFactory }) => {
```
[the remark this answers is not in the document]
cdamus: The new "Sample API Provider" Theia-extension uses this mechanism and I adapted its (working) code to fit the Foo example here, so it should be OK. 😀

cdamus:
Thanks @jcortell68 for the review! Commit 2ed73f38b280ecf69310a7df9b89873830ef86a7 should be better than ever.

jcortell68 (approved):
Looks great. Thanks for the adjustments!

cdamus:
Quick-fix action test failures on Ubuntu Linux are the same as occurred in the commit(s) on which this PR is based, so not a regression in this PR.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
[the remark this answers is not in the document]
martin-fleck-at: Full trust ;-)

martin-fleck-at (approved):
@cdamus Thank you very much for the enhanced documentation! I think that adds great value, especially with new images and further clarifications.

PR #13304, opened 2024-01-24
Hide cmd terminal while debugging

mulinJiao:
#### What it does

Hide cmd terminal while debugging

#### How to test

* Use [cpptools](https://github.com/microsoft/vscode-cpptools) plug-in to debug c/c++ code
* `startDebugAdapter` will pull up an empty cmd terminal
image

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/node/debug/plugin-node-debug-adapter-creator.ts
+++ b/packages/plugin-ext/src/plugin/node/debug/plugin-node-debug-adapter-creator.ts
@@ -113,6 +113,10 @@ export class NodeDebugAdapterCreator extends PluginDebugAdapterCreator {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         const options: any = { stdio: ['pipe', 'pipe', 2] };
 
+        if (isWindows) {
+            options.windowsHide = true;
+        }
+
         if (executable.options) {
             options.cwd = executable.options.cwd;
 


packages/plugin-ext/src/plugin/node/debug/plugin-node-debug-adapter-creator.ts
```diff
@@ -113,6 +113,10 @@ export class NodeDebugAdapterCreator extends PluginDebugAdapterCreator {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         const options: any = { stdio: ['pipe', 'pipe', 2] };
 
+        if (isWindows) {
+            Object.assign(options, { windowsHide: true });
```
msujew, line 117: Suggestion: Since `options` is of type `any` already, you should just be able to directly set the field.
```suggestion
            options.windowsHide = true;
```

mulinJiao: done

msujew:
@mulinJiao I just saw that you haven't signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php) yet. Please sign it with the same email that you used to create the commit.

mulinJiao:
> @mulinJiao I just saw that you haven't signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php) yet. Please sign it with the same email that you used to create the commit. Afterwards you might be required to force-push the commit to get the check to trigger correctly.

Done. Please take a look.

msujew:
Requesting @tsmaeder on this, since you're more knowledgable on the debugging stuff. I'm a bit torn on this issue:
* VSCode doesn't seem to need these arguments to debug successfully
* I'm not sure whether we unintentionally can break other stuff. As far as I know this will prevent all terminal windows from opening up during a debug session. This seems reasonable, but I'm not totally sure about this.
* The extension in question is not supported by Theia due to its licensing terms (Microsoft proprietary license).

This also acts as an FYI to @mulinJiao: The extension in question cannot be legally installed in Theia, due to Microsofts license terms.

planger:
Apologies for the delayed reply.

I'm a bit torn on this issue. On one hand, given that a debug adapter serves as a backend service usually out of sight for the user, it might seem justifiable for Theia to automatically close command windows. However, I concur with @tsmaeder that the decision to close a window should ideally rest with the debug adapter itself, and it feels somewhat overreaching for Theia to unilaterally shut down a command window initiated by the debug adapter.

Considering we haven't encountered this issue with other debug adapters, and the specific adapter in question isn't compatible with Theia anyway, my inclination is to maintain the status quo. This way, we avoid setting a precedent where Theia potentially overrides the matters of a debug adapter.

tsmaeder:
I'm closing this one. Feel free to revive in case there are new arguments for it.

PR #13306, opened 2024-01-24
Implement previously stubbed window#registerURIhandler (#13169)

### LINKED ISSUE #13169: registerUriHandler is not implemented
### Feature Description:
The function [registerUriHandler](https://github.com/eclipse-theia/theia/blob/469bd744a02e5dcdbcbc338af178d410dd909616/packages/plugin-ext/src/plugin/plugin-context.ts#L571) is not implemented in theia.
It is defined in the [VS code API](https://code.visualstudio.com/api/references/vscode-api#window.registerUriHandler).

This is needed to handle callbacks to theia through URIs, for example when creating an extension using GitHub OAuth.

rschnekenbu:
#### What it does
This PR implements the stubbed API window#registerUriHandler(). This was not marked properly as '@stubbed', but the implementation returned nothing.
A detailed explanation on how this is supposed to be used in extensions is located here:  https://github.com/microsoft/vscode/blob/1.85.1/src/vscode-dts/vscode.d.ts#L10146.

Fixes #13169

#### How to test 

1. install the following extension on theia browser example, main branch: 
- vsix archive: [open-external-0.1.0.zip](https://github.com/eclipse-theia/theia/files/14036122/open-external-0.1.0.zip)
- src: https://github.com/rschnekenbu/open-external
2. trigger the "Open External: Open external URI", then select "uri handler" menu. Nothing will happened
13169-before-patch

3. Update to this PR
4. The same sequence of actions will display a notification, as implemented by the uri handler provided by the test extension.
13169-after-patch

#### Follow-ups

- [x] https://github.com/eclipse-theia/theia/issues/13960
- [ ] https://github.com/eclipse-theia/theia-blueprint/issues/378

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/uri-main.ts
+++ b/packages/plugin-ext/src/main/browser/uri-main.ts
@@ -1,0 +1,72 @@
+// *****************************************************************************
+// Copyright (C) 2024 STMicroelectronics.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { Disposable, URI } from '@theia/core';
+import { MAIN_RPC_CONTEXT, UriExt, UriMain } from '../../common';
+import { RPCProtocol } from '../../common/rpc-protocol';
+import { interfaces } from '@theia/core/shared/inversify';
+import { OpenHandler, OpenerOptions, OpenerService } from '@theia/core/lib/browser';
+import { FrontendApplicationConfigProvider } from '@theia/core/lib/browser/frontend-application-config-provider';
+import { HostedPluginSupport } from '../../hosted/browser/hosted-plugin';
+
+export class UriMainImpl implements UriMain, Disposable {
+    private readonly proxy: UriExt;
+    private handlers = new Set<string>();
+    private readonly openerService: OpenerService;
+    private readonly pluginSupport: HostedPluginSupport;
+    private readonly openHandler: OpenHandler;
+
+    constructor(rpc: RPCProtocol, container: interfaces.Container) {
+        this.proxy = rpc.getProxy(MAIN_RPC_CONTEXT.URI_EXT);
+        this.openerService = container.get(OpenerService);
+        this.pluginSupport = container.get(HostedPluginSupport);
+
+        this.openHandler = {
+            id: 'theia-plugin-open-handler',
+            canHandle: async (uri: URI, options?: OpenerOptions | undefined): Promise<number> => {
+                if (uri.scheme !== FrontendApplicationConfigProvider.get().electron.uriScheme) {
+                    return 0;
+                }
+                await this.pluginSupport.activateByUri(uri.scheme, uri.authority);
+                if (this.handlers.has(uri.authority)) {
+                    return 500;
+                }
+                return 0;
+            },
+            open: async (uri: URI, options?: OpenerOptions | undefined): Promise<undefined> => {
+                if (!this.handlers.has(uri.authority)) {
+                    throw new Error(`No plugin to handle this uri: : '${uri}'`);
+                }
+                this.proxy.$handleExternalUri(uri.toComponents());
+            }
+        };
+
+        this.openerService.addHandler?.(this.openHandler);
+    }
+
+    dispose(): void {
+        this.openerService.removeHandler?.(this.openHandler);
+        this.handlers.clear();
+    }
+
+    async $registerUriHandler(pluginId: string, extensionDisplayName: string): Promise<void> {
+        this.handlers.add(pluginId);
+    }
+
+    async $unregisterUriHandler(pluginId: string): Promise<void> {
+        this.handlers.delete(pluginId);
+    }
+}
--- a/packages/plugin-ext/src/plugin/uri-ext.ts
+++ b/packages/plugin-ext/src/plugin/uri-ext.ts
@@ -1,0 +1,60 @@
+// *****************************************************************************
+// Copyright (C) 2024 STMicroelectronics.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import * as theia from '@theia/plugin';
+import {
+    UriExt,
+    PLUGIN_RPC_CONTEXT, PluginInfo, UriMain
+} from '../common/plugin-api-rpc';
+import { RPCProtocol } from '../common/rpc-protocol';
+import { Disposable, URI } from './types-impl';
+import { UriComponents } from '../common/uri-components';
+
+export class UriExtImpl implements UriExt {
+
+    private handlers = new Map<string, theia.UriHandler>();
+
+    private readonly proxy: UriMain;
+
+    constructor(readonly rpc: RPCProtocol) {
+        this.proxy = rpc.getProxy(PLUGIN_RPC_CONTEXT.URI_MAIN);
+        console.log(this.proxy);
+    }
+
+    registerUriHandler(handler: theia.UriHandler, plugin: PluginInfo): theia.Disposable {
+        const pluginId = plugin.id;
+        if (this.handlers.has(pluginId)) {
+            throw new Error(`URI handler already registered for plugin ${pluginId}`);
+        }
+
+        this.handlers.set(pluginId, handler);
+        this.proxy.$registerUriHandler(pluginId, plugin.displayName || plugin.name);
+
+        return new Disposable(() => {
+            this.proxy.$unregisterUriHandler(pluginId);
+            this.handlers.delete(pluginId);
+        });
+    }
+
+    $handleExternalUri(uri: UriComponents): Promise<void> {
+        const handler = this.handlers.get(uri.authority);
+        if (!handler) {
+            return Promise.resolve();
+        }
+        handler.handleUri(URI.revive(uri));
+        return Promise.resolve();
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,14 @@
 
 - [Previous Changelogs](https://github.com/eclipse-theia/theia/tree/master/doc/changelogs/)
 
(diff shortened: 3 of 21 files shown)


rschnekenbu:
@tsmaeder, would you have time to review this one?

JonasHelming:
@tsmaeder This would be good to have for the release

packages/core/src/browser/extension-open-handler.ts
```diff
@@ -0,0 +1,63 @@
...
+import { MaybePromise, URI } from '../common';
+import { OpenHandler, OpenerOptions } from './opener-service';
+
+export interface UriHandler {
+    canHandleURI(uri: URI): boolean;
+    handleUri(uri: URI): Promise<boolean>;
+}
+
+@injectable()
+export class ExtensionOpenHandler implements OpenHandler {
```
tsmaeder, line 27: Hmh...this is an extension mechanism to the `OpenerService`, which itself already has an extension mechanism. IMO, it would make more sense to make OpenerService properly extensible.

packages/core/src/browser/extension-open-handler.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    readonly id = 'extensionsURIHandlers';
+
+    private providers = new Map<string, UriHandler>();
```
tsmaeder, line 31: If it stores handlers, it should be called `handlers` ideally `handlersByAuthority`.

packages/core/src/browser/extension-open-handler.ts
```diff
@@ -0,0 +1,63 @@
...
+        return 0;
+    }
+
+    open(uri: URI, options?: OpenerOptions | undefined): MaybePromise<object | undefined> {
+        const authority = uri.authority;
+        const provider = this.providers.get(authority);
+        if (provider) {
+            return provider.handleUri(uri);
+        }
+        return Promise.reject(`Impossible to handle ${uri}.`);
```
tsmaeder, line 52: We should reject with an error. This should not happen if `canHandle` returns true.

packages/core/src/browser/extension-open-handler.ts
```diff
@@ -0,0 +1,63 @@
...
+            return 500;
+        }
+        return 0;
+    }
+
+    open(uri: URI, options?: OpenerOptions | undefined): MaybePromise<object | undefined> {
+        const authority = uri.authority;
+        const provider = this.providers.get(authority);
+        if (provider) {
+            return provider.handleUri(uri);
```
tsmaeder, line 50: The documentation of `OpenHandler` requires an opened widget or `undefined` as the return value, not a boolean.

packages/core/src/browser/extension-open-handler.ts
```diff
@@ -0,0 +1,63 @@
...
+        }
+        return Promise.reject(`Impossible to handle ${uri}.`);
+    }
+
+    public registerHandler(extensionId: string, handler: UriHandler): void {
+        this.providers.set(extensionId, handler);
```
tsmaeder, line 56: You can't talk about "extensionId" in the core package: plugin support might not be included in any particular Theia application. This handling needs to go into the `UriMain` service.

packages/plugin-ext/src/main/browser/uri-main.ts
```diff
@@ -0,0 +1,76 @@
...
+        const extensionId = this.handlers.get(handle);
+        if (extensionId) {
+            this.handlers.delete(handle);
+            this.extensionOpenHandler.unregisterHandler(extensionId);
+        }
+    }
+
+}
+
+class ExtensionUriHandler implements UriHandler {
```
tsmaeder, line 56: We must not talk about "Extensions" when we mean plugins. Extensions are NPM modules that extend Theia at compile time. This is confusing enough without mixing the terms in our source code.

packages/plugin-ext/src/plugin/uri-ext.ts
```diff
@@ -0,0 +1,72 @@
...
+            this.handles.delete(extensionId);
+            this.handlers.delete(handle);
+        });
+    }
+
+    $handleExternalUri(handle: number, uri: UriComponents): Promise<void> {
+        const handler = this.handlers.get(handle);
+
+        if (!handler) {
+            return Promise.resolve(undefined);
```
tsmaeder, line 61: I think we can remove the `undefined`.

packages/plugin-ext/src/plugin/uri-ext.ts
```diff
@@ -0,0 +1,72 @@
...
+        }
+        try {
+            handler.handleUri(URI.revive(uri));
+        } catch (err) {
+            console.log(`error while handling external uri: ${uri}`);
```
tsmaeder, line 66: Why not let the error propagate?

tsmaeder (changes requested):

tsmaeder:
@rschnekenbu what I don't understand is where we hook up theia to be the system-wide URI handler for some uri's. Is this already hooked up somewhere?

packages/plugin-ext/src/main/browser/plugin-uri-handler-service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export const PluginUriHandlerService = Symbol('PluginUriHandlerService');
+
+export interface PluginUriHandlerService {
+    registerUriHandler(id: string, uriHandler: UriHandler): Disposable;
+    unregisterUriHandler?(id: string): void;
+}
+
+@injectable()
+export class DefaultPluginUriHandlerService implements PluginUriHandlerService {
```
tsmaeder, line 34: It's not clear to me why we need this class. Can't this code just live in the `UriMainImpl` class?

jfaltermeier:
I would like to build the release soon, what is the status here?

rschnekenbu:
The PR is under review again, not sure when it will be accepted.

packages/plugin-ext/src/main/browser/uri-main.ts
```diff
@@ -0,0 +1,82 @@
...
+        this.proxy = rpc.getProxy(MAIN_RPC_CONTEXT.URI_EXT);
+        this.openerService = container.get(OpenerService);
+    }
+
+    dispose(): void {
+        this.handlers.forEach(handler => this.openerService.removeHandler?.(handler));
+        this.handlers.clear();
+    }
+
+    async $registerUriHandler(handle: number, extensionId: string, extensionDisplayName: string): Promise<void> {
```
tsmaeder, line 39: "plugin"! It's probably just easiest to do a "find" on this PR page to find all occurrences of "extension".

packages/plugin-ext/src/main/browser/uri-main.ts
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        if (uri.authority === this.pluginId) {
+            return 500;
+        }
+        return 0;
+    }
+
+    async open(uri: URI, options?: OpenerOptions | undefined): Promise<undefined> {
+        if (this.canHandle(uri) === 0) {
```
tsmaeder, line 76: I don't think this check is necessary: if we do not return an O.K.-value from `canHandle()`, this method should never be called.

packages/plugin-ext/src/main/browser/uri-main.ts
```diff
@@ -0,0 +1,82 @@
...
+        this.handlers.forEach(handler => this.openerService.removeHandler?.(handler));
+        this.handlers.clear();
+    }
+
+    async $registerUriHandler(handle: number, extensionId: string, extensionDisplayName: string): Promise<void> {
+        const extensionUriHandler = new PluginOpenHandler(this.proxy, handle, extensionId, extensionDisplayName);
+        if (this.openerService.addHandler?.(extensionUriHandler)) {
+            this.handlers.set(handle, extensionUriHandler);
+        }
+        return Promise.resolve();
```
tsmaeder, line 44: Do we need this?

packages/plugin-ext/src/main/browser/uri-main.ts
```diff
@@ -0,0 +1,82 @@
...
+            return 500;
+        }
+        return 0;
+    }
+
+    async open(uri: URI, options?: OpenerOptions | undefined): Promise<undefined> {
+        if (this.canHandle(uri) === 0) {
+            throw new Error(`Extension ${this.pluginName} is not supposed to handle this URI: ${uri}`);
+        }
+        await Promise.resolve(this.proxy.$handleExternalUri(this.handle, uri.toComponents()));
```
tsmaeder, line 79: You can just write `await this.proxy.handle....` as the last line. No need to return a promise from an async method.

packages/plugin-ext/src/plugin/uri-ext.ts
```diff
@@ -0,0 +1,65 @@
...
+    PLUGIN_RPC_CONTEXT, PluginInfo, UriMain
+} from '../common/plugin-api-rpc';
+import { RPCProtocol } from '../common/rpc-protocol';
+import { Disposable, URI } from './types-impl';
+import { UriComponents } from '../common/uri-components';
+
+export class UriExtImpl implements UriExt {
+
+    private static handle = 0;
+    private handles = new Set<string>();
```
tsmaeder, line 29: This set does not store handles, it stores the ids of the extensions that have already registered a handler. `registeredExtensions` maybe?

packages/plugin-ext/src/plugin/uri-ext.ts
```diff
@@ -0,0 +1,65 @@
...
+        return new Disposable(() => {
+            this.proxy.$unregisterUriHandler(handle);
+            this.handles.delete(extensionId);
+            this.handlers.delete(handle);
+        });
+    }
+
+    $handleExternalUri(handle: number, uri: UriComponents): Promise<void> {
+        const handler = this.handlers.get(handle);
+        if (!handler) {
```
tsmaeder, line 59: Shouldn't this throw an error? It sounds like this should never happen. In general, it's not a good idea to fail silently.

packages/plugin-ext/src/plugin/uri-ext.ts
```diff
@@ -0,0 +1,65 @@
...
+            return Promise.resolve();
+        }
+        handler.handleUri(URI.revive(uri));
+        return Promise.resolve();
```
tsmaeder, line 63: This line is not necessary.

tsmaeder (changes requested):
Still a bunch of "extensions" left.

tsmaeder:
@rschnekenbu when I open the "run" dialog from the Windows start menu and type "theia://ryanleonhardt.open-external" Windows tells me there is not handler in the system for the "theia" scheme. Isn't this what the "sytem-wide" url handler (https://code.visualstudio.com/api/references/vscode-api#window) is supposed to do? When I open a "vscode://foo.bar" uri, VS Code handles it.

tsmaeder:
See also https://github.com/microsoft/vscode/blob/c72447e8d8aaa7497c9a4bd68bc4301584b92beb/src/vs/platform/url/electron-main/electronUrlListener.ts#L20

tsmaeder:
@msujew since @rschnekenbu and @tsmaeder both worked on this, we're look for a third party to review.

packages/plugin-ext/src/main/browser/uri-main.ts
```diff
@@ -0,0 +1,72 @@
...
+                if (this.handlers.has(uri.authority)) {
+                    return 500;
+                }
+                return 0;
+            },
+            open: async (uri: URI, options?: OpenerOptions | undefined): Promise<undefined> => {
+                if (!this.handlers.has(uri.authority)) {
+
+                }
+                this.proxy.$handleExternalUri(uri.toComponents());
```
msujew, line 53: What is the empty `if` branch supposed to do?

msujew (commented):
I didn't have time to test the changes yet, but the code looks mostly good. Just one issue see below.

I'll perform a final review tomorrow once I have tested everything.

msujew (approved):
Alright, did some testing, things look good to me. Please fix the issue I've commented on above and feel free to merge afterwards.

PR #13322, opened 2024-01-26
Replace Discourse with Github Discussions

### LINKED ISSUE #13321: Update Readme and publishing doc to Github Discussions (instead of discourse)
Discourse is deprecated, we should update links

JonasHelming:
fixed #13321


jfaltermeier (approved):
Thanks, lgtm!

PR #13324, opened 2024-01-26
Add comment-documentation for several key Theia utility classes

jcortell68:
#### What it does

Adds comment-documentation for several key Theia utility classes
* DisposableCollection
* AbstractReferenceCollection
* ReferenceCollection
* SyncReferenceCollection

Lack of documentation, especially for non-trivial but pervasively used types is, IMO, a hindrance to the widespread adoption of the platform. Imagine STL or any other major library expecting developers to learn and understand its various facilities by reading the code line-by-line. It significantly increases the learning curve and makes the platform less approachable.

Where I am having to do such line-by-line studying of the code to understand these essential utilties, I'll be making contributions to hopefully bring some relief. However, I do think this technical debt should be tackled in a more comprehensive and intentional manner now that Theia has presumably reached some sort of critical mass.

#### How to test

Someone intimate with these utility types should read the new documentation and check for accuracy and understandability.

#### Follow-ups

N/A

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/disposable.spec.ts
+++ b/packages/core/src/common/disposable.spec.ts
@@ -37,34 +37,28 @@ describe('Disposables', () => {
         const collection = new DisposableCollection();
         collection.onDispose(onDispose);
 
-        // DisposableCollection doesn't provide direct access to its array,
-        // so we have to defeat TypeScript to properly test pruning.
-        function collectionSize(): number {
-            /* eslint-disable  @typescript-eslint/no-explicit-any */
-            return (<any>collection).disposables.length;
-        }
         const disposable1 = Disposable.create(elementDispose);
         collection.push(disposable1);
-        expect(collectionSize()).equal(1);
+        expect(collection['disposables']).to.have.lengthOf(1);
 
         const disposable2 = Disposable.create(elementDispose);
         collection.push(disposable2);
-        expect(collectionSize()).equal(2);
+        expect(collection['disposables']).to.have.lengthOf(2);
 
         disposable1.dispose();
-        expect(collectionSize()).equal(1);
+        expect(collection['disposables']).to.have.lengthOf(1);
         expect(onDispose).to.have.not.been.called();
         expect(collection.disposed).is.false;
 
         // Test that calling dispose on an already disposed element doesn't
         // alter the collection state
         disposable1.dispose();
-        expect(collectionSize()).equal(1);
+        expect(collection['disposables']).to.have.lengthOf(1);
         expect(onDispose).to.have.not.been.called();
         expect(collection.disposed).is.false;
 
         disposable2.dispose();
-        expect(collectionSize()).equal(0);
+        expect(collection['disposables']).to.be.empty;
         expect(collection.disposed).is.true;
         expect(onDispose).to.have.been.called.once;
     });
--- a/packages/core/src/common/disposable.spec.ts
+++ b/packages/core/src/common/disposable.spec.ts
@@ -14,8 +14,11 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { expect } from 'chai';
+import { expect, spy, use } from 'chai';
 import { DisposableCollection, Disposable } from './disposable';
+import * as spies from 'chai-spies';
+
+use(spies);
 
 describe('Disposables', () => {
     it('Is safe to use Disposable.NULL', () => {
@@ -25,5 +28,73 @@ describe('Disposables', () => {
         collectionA.dispose();
         expect(collectionA.disposed, 'A should be disposed after being disposed.').to.be.true;
         expect(collectionB.disposed, 'B should not be disposed because A was disposed.').to.be.false;
+    });
+
+    it('Collection is auto-pruned when an element is disposed', () => {
+        const onDispose = spy(() => { });
+        const elementDispose = () => { };
+
+        const collection = new DisposableCollection();
+        collection.onDispose(onDispose);
+
+        // DisposableCollection doesn't provide direct access to its array,
+        // so we have to defeat TypeScript to properly test pruning.
+        function collectionSize(): number {
+            /* eslint-disable  @typescript-eslint/no-explicit-any */
+            return (<any>collection).disposables.length;
+        }
+        const disposable1 = Disposable.create(elementDispose);
+        collection.push(disposable1);
+        expect(collectionSize()).equal(1);
+
+        const disposable2 = Disposable.create(elementDispose);
+        collection.push(disposable2);
+        expect(collectionSize()).equal(2);
+
+        disposable1.dispose();
+        expect(collectionSize()).equal(1);
+        expect(onDispose).to.have.not.been.called();
+        expect(collection.disposed).is.false;
+
+        // Test that calling dispose on an already disposed element doesn't
+        // alter the collection state
+        disposable1.dispose();
+        expect(collectionSize()).equal(1);
+        expect(onDispose).to.have.not.been.called();
+        expect(collection.disposed).is.false;
+
+        disposable2.dispose();
+        expect(collectionSize()).equal(0);
+        expect(collection.disposed).is.true;
+        expect(onDispose).to.have.been.called.once;
+    });
+
+    it('onDispose is only called once on actual disposal of elements', () => {
+        const onDispose = spy(() => { });
+        const elementDispose = spy(() => { });
+
+        const collection = new DisposableCollection();
+        collection.onDispose(onDispose);
+
+        // if the collection is empty 'onDispose' is not called
+        collection.dispose();
+        expect(onDispose).to.not.have.been.called();
+
+        // 'onDispose' is called because we actually dispose an element
+        collection.push(Disposable.create(elementDispose));
+        collection.dispose();
+        expect(elementDispose).to.have.been.called.once;
+        expect(onDispose).to.have.been.called.once;
+
+        // if the collection is empty 'onDispose' is not called and no further element is disposed
+        collection.dispose();
+        expect(elementDispose).to.have.been.called.once;
+        expect(onDispose).to.have.been.called.once;
+
+        // 'onDispose' is not called again even if we actually dispose an element
+        collection.push(Disposable.create(elementDispose));
+        collection.dispose();
+        expect(elementDispose).to.have.been.called.twice;
+        expect(onDispose).to.have.been.called.once;
     });
 });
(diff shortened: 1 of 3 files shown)


packages/core/src/common/disposable.ts
```diff
@@ -47,6 +47,29 @@ Object.defineProperty(Disposable, 'NULL', {
...
 });
 
+/**
+ * Utility for tracking a collection of Disposable objects, specifically for
+ * the sake of finding out when all items in the collection have been disposed.
+ *
+ * ```
+ * const dc = new DisposableCollection(myDisposables);
+ * dc.onDispose(() => {
+ *    console.log('All elements in the collection have been disposed');
```
martin-fleck-at, line 57: I'd say it is a bit risky to claim this as the main goal of this class since the `onDispose` element is actually only called once, even if you keep re-using the collection.

packages/core/src/common/disposable.ts
```diff
@@ -47,6 +47,29 @@ Object.defineProperty(Disposable, 'NULL', {
...
+ * });
+ * ```
+ *
+ * You can add any number of Disposables to the collection at any time. You
+ * cannot query the contents of the collection. An element is automatically
+ * pruned from the collection when it is disposed.
+ *
+ * The collection will notify only once. It will continue to function
```
martin-fleck-at, line 65: I'd say this is the main point of this doc besides the general purpose, because I believe that is the most unexpected part of the collections behavior.

packages/core/src/common/disposable.ts
```diff
@@ -47,6 +47,29 @@ Object.defineProperty(Disposable, 'NULL', {
...
+ * in so far as accepting new Disposables and pruning them when they
+ * are disposed, but such activity will never result in another
+ * notification.
+ *
+ * Calling dispose() on the collection disposes all its elements. A notification
+ * is sent (if no notification was previously sent).
```
martin-fleck-at, line 71:
```suggestion
 * is sent the first time such a call is made and elements were actually disposed.
```

suggestion.

packages/core/src/common/disposable.ts
```diff
@@ -47,6 +47,29 @@ Object.defineProperty(Disposable, 'NULL', {
...
+ */
```
martin-fleck-at, line 72: Since we are already touching this, could I maybe ask you to add a test case to `disposable.spec.ts` that also covers the behavior in more detail?

```
import { expect, spy, use } from 'chai';
import { DisposableCollection, Disposable } from './disposable';
import * as spies from 'chai-spies';

use(spies);

describe('Disposables', () => {
    [...]
    it('onDispose is only called once on actual disposal of elements', () => {
        const onDispose = spy(() => { });
        const elementDispose = spy(() => { });

        const collection = new DisposableCollection();
        collection.onDispose(onDispose);

        // if the collection is empty 'onDispose' is not called
        collection.dispose();
        expect(onDispose).to.not.have.been.called();

        // 'onDispose' is called because we actually dispose an element
        collection.push(Disposable.create(elementDispose));
        collection.dispose();
        expect(elementDispose).to.have.been.called.once;
        expect(onDispose).to.have.been.called.once;

        // if the collection is empty 'onDispose' is not called and no further element is disposed
        collection.dispose();
        expect(elementDispose).to.have.been.called.once;
        expect(onDispose).to.have.been.called.once;

        // 'onDispose' is not called again even if we actually dispose an element
        collection.push(Disposable.create(elementDispose));
        collection.dispose();
        expect(elementDispose).to.have.been.called.twice;
        expect(onDispose).to.have.been.called.once;
    });
    [...]
```

martin-fleck-at (changes requested):
Thank you very much for the additional documentation, John! I do have some remarks regarding the `DisposableCollection` but other than that, this looks correct to me.

packages/core/src/common/disposable.ts
```diff
@@ -47,6 +47,29 @@ Object.defineProperty(Disposable, 'NULL', {
     }
 });
 
+/**
+ * Utility for tracking a collection of Disposable objects, specifically for
+ * the sake of finding out when all items in the collection have been disposed.
```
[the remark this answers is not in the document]
jcortell68: Mabe I'm overlooking something. I'm a newbie to these classes, but...

1. The value you mention seems rudimentary. I can just as easily do this
```
const disposables: Disposable[] = [];
// add disposables to array
disposables.forEach(disposable => disposable.dispose());
disposables.length = 0;
```

In other words, with two short lines of code, I can get exactly the functionality you mention. Sure, it's more convenient to just call `dispose()` on a collection object, but surely that convenience is not the key value-add of a class that's 75 lines long. The real magic that I see the utility providing is how it injects itself in the dispose() of its members so that it can find out when all items are disposed and then fire off an event. This lets it be used in other classes to, for example, provide reference counting in a non-intrusive way. I really think that's the special sauce here. But please let me know if you think I'm still off.

2. 
> However, even after calling dispose I can still continue filling the collection with more disposables which may again be disposed.

That's not different than a regular array; see my example above. Oh, and note that the notification mechanism is non-functional once the collection count drops to zero for the first time. So, you really couln't continue using it the way you say and still get the full (and what I'd say is the real) benefit of the utility.

packages/core/src/common/disposable.ts
jcortell68: See my comment above. That it lets you continue using the collection after reaching length zero is an oddity for sure. But I stand by the claim. See my argument above and let me know if you're still not convinced.

packages/core/src/common/disposable.ts
jcortell68: That's a good point. The collection auto-pruning itself when its elements are disposed is indeed a key value-add, even if you're not making use of the notification feature. OK. Let me revise this and see if we can come up with something we both like :-)

packages/core/src/common/disposable.ts
jcortell68: You bet

packages/core/src/common/disposable.ts
jcortell68: @martin-fleck-at I've revamped the description of `DisposableCollection`. I think the new version should address your concerns, and is, frankly, far better thanks to your feedback.

jcortell68:
@martin-fleck-at Sorry for the churn.  All done now and ready for re-review when you get a chance.

packages/core/src/common/disposable.ts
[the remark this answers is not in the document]
martin-fleck-at: I do agree that most things could also be done with an array and maybe I'm off by the real value of the collection. In any case, I think the way you improved the documentation now is way more neutral and highlights all the benefits so that adopters simply can make their own call. Thank you very much for that!

packages/core/src/common/disposable.ts
martin-fleck-at: See answer above ;-) And yes, I agree it is an oddity and I'm not sure why it was done this way but to have it documented is definitely the right move, especially since it may not be intuitive.

packages/core/src/common/disposable.ts
martin-fleck-at: Well, thank you for adapting it so quickly and I agree it reads very clear now.

packages/core/src/common/disposable.ts
martin-fleck-at: ... and you even added a test case of your own, thank you!

packages/core/src/common/disposable.ts
```diff
@@ -47,6 +47,34 @@ Object.defineProperty(Disposable, 'NULL', {
...
+/**
+ * Utility for tracking a collection of Disposable objects.
+ *
+ * This utility provides a number of benefits over just using an array of
+ * Disposables:
+ *
+ * - the collection is auto-pruned when an element it contains is disposed by
+ * any code that has a reference to it
+ * - you can register to be notified when all elements in the collection have
+ * been disposed[1]
```
martin-fleck-at, line 59:
```suggestion
 * been disposed [1]
```

Not sure cause usually we do not have a space when we actually have superscript but having no space here seems a bit odd to me.

packages/core/src/common/disposable.spec.ts
```diff
@@ -26,4 +29,72 @@ describe('Disposables', () => {
...
+    it('Collection is auto-pruned when an element is disposed', () => {
+        const onDispose = spy(() => { });
+        const elementDispose = () => { };
+
+        const collection = new DisposableCollection();
+        collection.onDispose(onDispose);
+
+        // DisposableCollection doesn't provide direct access to its array,
+        // so we have to defeat TypeScript to properly test pruning.
+        function collectionSize(): number {
```
martin-fleck-at, line 42: I'm not the biggest fan of this construct and I believe we could replace this by simply accessing the `disposables` and utilizing the [lenghtOf](https://www.chaijs.com/api/bdd/#method_lengthof) method from chai.

```
expect(collection['disposables']).to.have.lengthOf(1);
expect(collection['disposables']).to.be.empty;
```

martin-fleck-at (changes requested):
Thank you for the quick update, John! I think we are almost there, I'd just like some cleanup of the new test case.

packages/core/src/common/disposable.spec.ts
jcortell68: Oh gosh. I'm still such a TypeScript newbie in many ways. I had no idea that accessing a property with bracket notation circumvents the type's accessibility attributes!

jcortell68:
@martin-fleck-at I think I took care of all your feedback items. Let me know if I missed something

martin-fleck-at (approved):
Thank you for the quick update, everything looks good to me now!

packages/core/src/common/disposable.spec.ts
martin-fleck-at: Haha, yes, we can profiting from the freedom of JavaScript and indexed access!

jcortell68: Well, it's funny because at first I thought backet access would give you an `any` and give you full freedom from there on out. But not so! 
```
class SomeThing {
    foo() : void {}
    private bar() : void {}
}
class SomeClass {
    private obj = new SomeThing();
}
const sc = new SomeClass();
sc.obj.foo();      // error; obj is private
sc['obj'].foo();   // ok; bracket notation to the rescue
sc['jjj'].foo();   // error property jjj does not existin in SomeClass
sc['obj'].bar();   // error; compiler doesn't treat sc['obj'] as any.
                   // bracket notation circumvents accessibility
                   // traits but that's about it, I think
````

martin-fleck-at: You are right, index access does not imply an implicit `any`. This is definitely an advantage as you keep your type-safety for compiler checks (e.g., someone renames the field would result in a compile error instead of a runtime error). So in your example you would still be able to do `sc['obj']['bar']();` to call the private method of a private object and get all the typing information.

PR #13342, opened 2024-02-01
Fix rending of quickpick buttons

### LINKED ISSUE #13076: QuickPick buttons do not render
### Bug Description:

QuickPick buttons do not render in Theia. They render as expected in VS Code.

### Steps to Reproduce:

1. Create a QuickPick with a button.
2. Await to be able to keep QuickPick shown.
3. Show the QuickPick.

Example:

```ts
const quickPick = vscode.window.createQuickPick();
quickPick.buttons = [vscode.QuickInputButtons.Back];

await new Promise<void>(resolve => {
    quickPick.onDidAccept(() => {
        resolve();
    });

    quickPick.show();
});
```

### Additional Information

- Operating System: Windows 10 Enterprise
- Theia Version: 1.43.0

QuickPick with Back button in VS Code:

image

QuickPick with Back button in Theia:

image

tortmayr:
#### What it does

Ensure that  the Theia specific wrapper for the `MonacoQuickPickItem` properly forwards assignments of the "buttons" property to the wrapped item.

More context:
The buttons property is assigned in the [QuickOpenMainImpl](https://github.com/eclipse-theia/theia/blob/b15b639fa3b64a9518efdf3e82ea9098915ba145/packages/plugin-ext/src/main/browser/quick-open-main.ts#L272). This works perfectly fine when creating a `InputBox` via the quickopen service but for the normal `QuickPick`
this assignment gets lost because the actual monaco quickpick item is wrapped and the buttons property assignment is not forwarded.

Fixes #13076

#### How to test

Follow the steps to reproduced described in https://github.com/eclipse-theia/theia/issues/13076#issuecomment-1896065081
(vsix for testing is provided).
After installling the plugins., execute the `Hello World` command. Notice that buttons are now properly rendered (while not being rendered at all on master)
#### Review checklist

#### Reminder for reviewers


packages/monaco/src/browser/monaco-quick-input-service.ts
```diff
@@ -580,6 +580,14 @@ class MonacoQuickPick<T extends QuickPickItem> extends MonacoQuickInput implemen
         return this.wrapped.items.map(item => QuickPickSeparator.is(item) ? item : item.item);
     }
 
+    get buttons(): ReadonlyArray<IQuickInputButton> {
```
tsmaeder, line 583: The typing here is not right: there is no "buttons" property on `QuickPick`. Returning monaco-specific types here is not appropriate. `QuickPick` is a wrapper around `IQuickPick`.

tsmaeder (changes requested):

packages/monaco/src/browser/monaco-quick-input-service.ts
tortmayr: So how should we handle this?
The current implementation in the [quick-open-main](https://github.com/eclipse-theia/theia/blob/b15b639fa3b64a9518efdf3e82ea9098915ba145/packages/plugin-ext/src/main/browser/quick-open-main.ts#L272) does not do any type checks at all. It simply iterates through the "params" (of type `TransferQuickPick`) list and assigns them as property of the `input` element.

So basically we would have to make sure that all properties of the `TransferQuickPick` are present on the wrapper and properly forwared to the underyling monaco specific element.

On a sidenote: There is also a wrapper for `MonacoQuickInput` which seems currently unused. That`s why the buttons works when using `QuickInput` because in that case we use the monaco type directly.
We should probably also wrap that correctly ore remove this class altogether

tsmaeder: Add the buttons types `readonly QuickInputButton[]` to the `QuickPick` interface and fix the typing by appropriately casting/converting.
Which wraper is unused? Can you explain, please?

JonasHelming:
@tortmayr

packages/monaco/src/browser/monaco-quick-input-service.ts
tortmayr: @tsmaeder 

Unfortunatley the `QuickInputButton` and Monacos `IQuickInputButton` interface are not compatible.
They use different URI types for the `iconPath`:
 - QuickInputButton: import from 'vscode-uri';
-  IQuickInputButton: import  from '@theia/monaco-editor/lib/vs/base/common/uri`

They difference is the `UriComponent` definition which has partial optional properties in the monaco variant
```
export interface UriComponents {
	scheme: string;
	authority?: string;
	path?: string;
	query?: string;
	fragment?: string;
}
``` 

but has no optional properties in the vscode-uri variant:
export interface UriComponents {
    scheme: string;
    authority: string;
    path: string;
    query: string;
    fragment: string;
}
```

tsmaeder: Can't you just cast the relevant structures? The values are not changed by the cast anyway.

tortmayr: 👍🏼 done

tsmaeder (approved):

PR #13344, opened 2024-02-02
Provide "New File" default implementation (#13303)

### LINKED ISSUE #13303: New File doesn't accept a filename
### Bug Description:
Selecting **New File...** from the File menu brings up a selection box prompting the user to "Select file type or enter file name." However, entering a filename doesn't do anything. The box disappears and no new editor is opened.

### Additional Information

- Operating System: MacOS Sonoma, Chrome browser
- Theia Version: 1.45

Extensions loaded:

```@theia/bulk-edit 1.45.0
@theia/callhierarchy 1.45.0
@theia/console 1.45.0
@theia/core 1.45.0
@theia/debug 1.45.0
@theia/editor 1.45.0
@theia/editor-preview 1.45.0
@theia/file-search 1.45.0
@theia/filesystem 1.45.0
@theia/getting-started 1.45.0
@theia/git 1.45.0
@theia/keymaps 1.45.0
@theia/markers 1.45.0
@theia/messages 1.45.0
@theia/metrics 1.45.0
@theia/mini-browser 1.45.0
@theia/monaco 1.45.0
@theia/navigator 1.45.0
@theia/notebook 1.45.0
@theia/outline-view 1.45.0
@theia/output 1.45.0
@theia/plugin-dev 1.45.0
@theia/plugin-ext 1.45.0
@theia/plugin-ext-vscode 1.45.0
@theia/plugin-metrics 1.45.0
@theia/preferences 1.45.0
@theia/preview 1.45.0
@theia/process 1.45.0
@theia/property-view 1.45.0
@theia/scm 1.45.0
@theia/scm-extra 1.45.0
@theia/search-in-workspace 1.45.0
@theia/task 1.45.0
@theia/terminal 1.45.0
@theia/timeline 1.45.0
@theia/toolbar 1.45.0
@theia/typehierarchy 1.45.0
@theia/userstorage 1.45.0
@theia/variable-resolver 1.45.0
@theia/vsx-registry 1.45.0
@theia/workspace 1.45.0
```

dhuebner:
#### What it does

Fixes #13303 

Adds a new dynamic `Create New File` item to the  create file picker. The new item will appear as soon as the user types something in the input field. Accepting the item will execute a `file.newFile` command handler and bypass the file name to it.

#### How to test

- In the browser app select `New File...`item from the `File` menu.
- Start typing. A new item `Create New File` should become visible
- Selecting this item will open a "save file" dialog with a file name pre-set. The parent directory should be the current user working directory or a workspace folder if selected.  

Bildschirmfoto 2024-02-02 um 12 43 46

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/common-frontend-contribution.ts
+++ b/packages/core/src/browser/common-frontend-contribution.ts
@@ -280,6 +280,15 @@ export namespace CommonCommands {
         category: VIEW_CATEGORY,
         label: 'Toggle Menu Bar'
     });
+    /**
+     * Command Parameters:
+     * - `fileName`: string
+     * - `directory`: URI
+     */
+    export const NEW_FILE = Command.toDefaultLocalizedCommand({
+        id: 'workbench.action.files.newFile',
+        category: FILE_CATEGORY
+    });
     export const NEW_UNTITLED_TEXT_FILE = Command.toDefaultLocalizedCommand({
         id: 'workbench.action.files.newUntitledTextFile',
         category: FILE_CATEGORY,
@@ -1424,6 +1433,7 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
         const items: QuickPickItemOrSeparator[] = [
             {
                 label: nls.localizeByDefault('New Text File'),
+                description: nls.localizeByDefault('Built-in'),
                 execute: async () => this.commandRegistry.executeCommand(CommonCommands.NEW_UNTITLED_TEXT_FILE.id)
             },
             ...newFileContributions.children
@@ -1446,10 +1456,43 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
 
                 })
         ];
+
+        const CREATE_NEW_FILE_ITEM_ID = 'create-new-file';
+        const hasNewFileHandler = this.commandRegistry.getActiveHandler(CommonCommands.NEW_FILE.id) !== undefined;
+        // Create a "Create New File" item only if there is a NEW_FILE command handler.
+        const createNewFileItem: QuickPickItem & { value?: string } | undefined = hasNewFileHandler ? {
+            id: CREATE_NEW_FILE_ITEM_ID,
+            label: nls.localizeByDefault('Create New File ({0})'),
+            description: nls.localizeByDefault('Built-in'),
+            execute: async () => {
+                if (createNewFileItem?.value) {
+                    const parent = await this.workingDirProvider.getUserWorkingDir();
+                    // Exec NEW_FILE command with the file name and parent dir as arguments
+                    return this.commandRegistry.executeCommand(CommonCommands.NEW_FILE.id, createNewFileItem.value, parent);
+                }
+            }
+        } : undefined;
+
         this.quickInputService.showQuickPick(items, {
             title: nls.localizeByDefault('New File...'),
             placeholder: nls.localizeByDefault('Select File Type or Enter File Name...'),
-            canSelectMany: false
+            canSelectMany: false,
+            onDidChangeValue: picker => {
+                if (createNewFileItem === undefined) {
+                    return;
+                }
+                // Dynamically show or hide the "Create New File" item based on the input value.
+                if (picker.value) {
+                    createNewFileItem.alwaysShow = true;
+                    createNewFileItem.value = picker.value;
+                    createNewFileItem.label = nls.localizeByDefault('Create New File ({0})', picker.value);
+                    picker.items = [...items, createNewFileItem];
+                } else {
+                    createNewFileItem.alwaysShow = false;
+                    createNewFileItem.value = undefined;
+                    picker.items = items.filter(item => item !== createNewFileItem);
+                }
+            }
         });
     }
 
--- a/packages/filesystem/src/browser/filesystem-frontend-contribution.ts
+++ b/packages/filesystem/src/browser/filesystem-frontend-contribution.ts
@@ -14,27 +14,36 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { injectable, inject } from '@theia/core/shared/inversify';
-import URI from '@theia/core/lib/common/uri';
-import { environment } from '@theia/core/shared/@theia/application-package/lib/environment';
-import { MaybePromise, SelectionService, isCancelled, Emitter } from '@theia/core/lib/common';
-import { Command, CommandContribution, CommandRegistry } from '@theia/core/lib/common/command';
+import { nls } from '@theia/core';
 import {
-    FrontendApplicationContribution, ApplicationShell,
-    NavigatableWidget, NavigatableWidgetOptions,
-    Saveable, WidgetManager, StatefulWidget, FrontendApplication, ExpandableTreeNode,
-    CorePreferences,
+    ApplicationShell,
     CommonCommands,
+    CorePreferences,
+    ExpandableTreeNode,
+    FrontendApplication,
+    FrontendApplicationContribution,
+    NavigatableWidget, NavigatableWidgetOptions,
+    OpenerService,
+    Saveable,
+    StatefulWidget,
+    WidgetManager,
+    open
 } from '@theia/core/lib/browser';
 import { MimeService } from '@theia/core/lib/browser/mime-service';
 import { TreeWidgetSelection } from '@theia/core/lib/browser/tree/tree-widget-selection';
-import { FileSystemPreferences } from './filesystem-preferences';
-import { FileSelection } from './file-selection';
-import { FileUploadService, FileUploadResult } from './file-upload-service';
-import { FileService, UserFileOperationEvent } from './file-service';
-import { FileChangesEvent, FileChangeType, FileOperation } from '../common/files';
+import { Emitter, MaybePromise, SelectionService, isCancelled } from '@theia/core/lib/common';
+import { Command, CommandContribution, CommandRegistry } from '@theia/core/lib/common/command';
 import { Deferred } from '@theia/core/lib/common/promise-util';
-import { nls } from '@theia/core';
+import URI from '@theia/core/lib/common/uri';
+import { environment } from '@theia/core/shared/@theia/application-package/lib/environment';
+import { inject, injectable } from '@theia/core/shared/inversify';
+import { UserWorkingDirectoryProvider } from '@theia/core/lib/browser/user-working-directory-provider';
+import { FileChangeType, FileChangesEvent, FileOperation } from '../common/files';
+import { FileDialogService, SaveFileDialogProps } from './file-dialog';
+import { FileSelection } from './file-selection';
+import { FileService, UserFileOperationEvent } from './file-service';
+import { FileUploadResult, FileUploadService } from './file-upload-service';
+import { FileSystemPreferences } from './filesystem-preferences';
 
 export namespace FileSystemCommands {
 
@@ -77,6 +86,15 @@ export class FileSystemFrontendContribution implements FrontendApplicationContri
 
     @inject(FileService)
     protected readonly fileService: FileService;
+
+    @inject(FileDialogService)
+    protected readonly fileDialogService: FileDialogService;
+
+    @inject(OpenerService)
+    protected readonly openerService: OpenerService;
+
+    @inject(UserWorkingDirectoryProvider)
+    protected readonly workingDirectory: UserWorkingDirectoryProvider;
 
     protected onDidChangeEditorFileEmitter = new Emitter<{ editor: NavigatableWidget, type: FileChangeType }>();
     readonly onDidChangeEditorFile = this.onDidChangeEditorFileEmitter.event;
@@ -134,6 +152,11 @@ export class FileSystemFrontendContribution implements FrontendApplicationContri
                 }
             }
         });
+        commands.registerCommand(CommonCommands.NEW_FILE, {
+            execute: (...args: unknown[]) => {
+                this.handleNewFileCommand(args);
+            }
+        });
     }
 
     protected canUpload({ fileStat }: FileSelection): boolean {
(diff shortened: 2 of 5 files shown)


packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -225,14 +223,36 @@ export class WorkspaceCommandContribution implements CommandContribution {
...
+                            this.fireCreateNewFile({ parent: parentUri, uri: fileUri });
+                            open(this.openerService, fileUri);
+                        }
+                    };
+
+                    // If a name was provided, check if it is valid, create a file and return.
+                    // Otherwise open a dialog to get the name.
+                    if (args !== undefined && args.length > 0 && args[0]) {
+                        newFileName = args[0];
+                        if (await this.validateFileName(newFileName, parent, true) === '') {
```
martin-fleck-at, line 246: We should double check whether `newFileName` is a string here because when we trigger that command from the toolbar in the File Navigator then actually the first command will be the `FileNavigatorWidget` and we will get an exception when we try to call the command from the toolbar.

martin-fleck-at (changes requested):
Hi Dennis, thank you very much for those changes, I think it definitely fills a gap in a very common functionality.

I added some comments but I also have a core question: Is there any particular reason we enforce this to be the same as the 'New File' command we have in the workspace? I feel like a separation may reflect the difference in functionality a little bit better:
- The Workspace 'New File' command is the one that you can trigger from the File Navigator toolbar and the one where we can actually create files in a known root folder.
- The Common 'New File' command is one that can be triggered from the command palette even if there is no open workspace and that will ask you where to save the file. Currently calling the New File command from the menu when no workspace is open will simply not show the option to create a file which is a shame cause basically all the pieces are already in place.

We can combine the code, of course, but I think aligning with VS Code might be better at the moment, i.e., always asking where to save the file and if we are in a workspace context and we do know about a selected URI we can propose this as the parent but we should always show the dialog. This ensures that the code always works properly.

I hope it is clear what I mean, if not, let me know and I'll try my best to explain better.

dhuebner:
Thank you for your comments! 

> * The Workspace 'New File' command is the one that you can trigger from the File Navigator toolbar and the one where we can actually create files in a known root folder.
> * The Common 'New File' command is one that can be triggered from the command palette even if there is no open workspace and that will ask you where to save the file. Currently calling the New File command from the menu when no workspace is open will simply not show the option to create a file which is a shame cause basically all the pieces are already in place.


Yes, I was thinking about this as well. In my PR Draft I tried to explain it in the PR description and offered some option how to implement it.  So I'm really glad to have somebody to discuss it now!

In fact the 'New File...' command in vs-code is the one from the welcome page and just opens a system 'Save' dialog. This behavior is really confusing when you already have a workspace open IMHO. I also think that a lot of people will tend to use the menu entry to create a new workspace file, especially in case where the Explorer view is closed or hidden. See also the user expectation [here](https://github.com/eclipse-theia/theia/issues/13303#issuecomment-1925612704).  

I think the best would be:
- when we don't have a workspace open - open a "Save" system dialog (just like vs-code does)  
- if we do have workspace open - do what the PR currently does

This leads me to some other questions:
1. Should we keep the "New File..." menu in the `core` package, or move it to `workspace` package?  (Currently you can not save the new created untitled files when in browser mode) 
2. How to open a new "Save" system dialog? :)  

FDYT?

martin-fleck-at:
> Yes, I was thinking about this as well. In my PR Draft I tried to explain it in the PR description and offered some option how to implement it. So I'm really glad to have somebody to discuss it now!
> 
> In fact the 'New File...' command in vs-code is the one from the welcome page and just opens a system 'Save' dialog. This behavior is really confusing when you already have a workspace open IMHO. I also think that a lot of people will tend to use the menu entry to create a new workspace file, especially in case where the Explorer view is closed or hidden. See also the user expectation [here](https://github.com/eclipse-theia/theia/issues/13303#issuecomment-1925612704).
> 
> I think the best would be:
> 
> * when we don't have a workspace open - open a "Save" system dialog (just like vs-code does)
> * if we do have workspace open - do what the PR currently does
> 
> This leads me to some other questions:
> 
> 1. Should we keep the "New File..." menu in the `core` package, or move it to `workspace` package?  (Currently you can not save the new created untitled files when in browser mode)
> 2. How to open a new "Save" system dialog? :)
> 
> FDYT?

So in VS Code it seems they not only trigger it from the Welcome Page but they also have a 'New File...' menu item in the 'File' menu, exactly the same as we do. And that item is available no matter whether you have a workspace open or not (i.e., just having a new empty window) by always opening a save dialog after the user provided some input. So I believe we should at least take that as a baseline of things that need to work. You can open a save dialog through the `FileDialogService.showSaveDialog` which is implemented for both the browser and the Electron context. I believe you can even provide the input the user has given in the quickbox as suggested default name.

I believe most of the "confusion" comes from the fact that we are mixing two different actions under the same name:

- A common `NEW_FILE` command that creates a new file independent from any workspace. That is currently used in the File menu of VS Code and Theia and basically provides a list of quick inputs through the `CommonMenus.FILE_NEW_CONTRIBUTIONS` in Theia. Extensions may add additional mechanisms, e.g., the Python extension. When we look at how they do it (even if a workspace is closed and even in the browser context), we can see that there is definitely no assumption of any kind of workspace but simply providing a save dialog.

- A workspace command that can create a new file based on the workspace context (usually provided through the toolbar in the File Explorer or the context menu of the file explorer). Here we can take the shorter way.

I believe we should not mix them and even if we were to have the information in case 1, since other contributions may not utilize that, I believe we should simply align with VS Code for now. However, that is just my opinion and others may disagree so if you want to consider the context (in case we have a workspace context), we can definitely involve other people in this discussion.

packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -225,14 +223,36 @@ export class WorkspaceCommandContribution implements CommandContribution {
 
     registerCommands(registry: CommandRegistry): void {
         this.registerOpenWith(registry);
-        registry.registerCommand(WorkspaceCommands.NEW_FILE, this.newWorkspaceRootUriAwareCommandHandler({
-            execute: uri => this.getDirectory(uri).then(parent => {
+        registry.registerCommand(CommonCommands.NEW_FILE, this.newWorkspaceRootUriAwareCommandHandler({
```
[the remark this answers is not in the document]
dhuebner: Obsolete as this are now two different commands.

packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -241,15 +261,7 @@ export class WorkspaceCommandContribution implements CommandContribution {
...
-
-                    dialog.open().then(async name => {
-                        if (name) {
-                            const fileUri = parentUri.resolve(name);
-                            await this.fileService.create(fileUri);
-                            this.fireCreateNewFile({ parent: parentUri, uri: fileUri });
-                            open(this.openerService, fileUri);
-                        }
-                    });
+                    dialog.open().then(doCreateFile);
```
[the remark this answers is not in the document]
dhuebner: I rewrote the implementation with `await` syntax.

dhuebner:
@martin-fleck-at 
> I believe we should simply align with VS Code for now.

It is also my opinion. I changed the implementation and hope now all the use cases work the same way as the do  in vs-code.  I also added `New File...` widget to the Welcome page. 


Bildschirmfoto 2024-03-01 um 13 38 24

planger:
> The test `../../src/tests/theia-main-menu.test.ts:95:9 › Theia Main Menu › open about dialog using menu` is failing in GH actions, but works fine locally. Is it flaky?

Not that I'm aware of. I'm monitoring the playwright tests and didn't see this one in particular failing before. There is however a certain amount of flakiness in our Playwright tests in general (< 1 % though). So I'll add this one to my list of "suspects" and observe whether I see this failing again.

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -1446,10 +1451,43 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
...
+        // Create a "Create New File" item only if there is a NEW_FILE command handler.
+        const createNewFileItem: QuickPickItem & { value?: string } | undefined = hasNewFileHandler ? {
+            id: CREATE_NEW_FILE_ITEM_ID,
+            label: nls.localizeByDefault('Create New File ({0})'),
+            description: nls.localizeByDefault('Built-in'),
+            execute: async () => {
+                if (createNewFileItem?.value) {
+                    const parent = await this.workingDirProvider.getUserWorkingDir();
+                    // Exec NEW_FILE command with the file name as additional argument
+                    return this.commandRegistry.executeCommand(CommonCommands.NEW_FILE.id, parent, createNewFileItem.value);
```
martin-fleck-at, line 1466: nit: I think that when it comes to parameters for this command the fileName is the more important and the mandatory one whereas the parent directory is more of an improved user experience so the fileName should come first and the potential parent should be second, what do you think?

packages/filesystem/src/browser/filesystem-frontend-contribution.ts
```diff
@@ -134,6 +148,22 @@ export class FileSystemFrontendContribution implements FrontendApplicationContri
...
         });
+        commands.registerCommand(CommonCommands.NEW_FILE, {
+            execute: async (...args: unknown[]) => {
+                let newFileName = undefined;
+                if (args !== undefined && typeof args[1] === 'string') {
+                    newFileName = args[1];
+                }
+                const title = nls.localizeByDefault('Create File');
+                const props: SaveFileDialogProps = { title, saveLabel: title, inputValue: newFileName };
+                const selectedFile = this.getSelection(...args)?.fileStat ?? await this.getFolderFromArgs(...args);
```
martin-fleck-at, line 159: I really like that we are trying to give our best to provide a good parent directory. However, I'm a bit concerned about the order. I believe that when a directory is explicitly given as argument to the command that it should have precedence. So maybe we should inject the `UserWorkingDirectoryProvider` here as well and then do have the following order: explicitly given argument > selection > user directory fallback.

What do you think?

packages/filesystem/src/browser/filesystem-frontend-contribution.ts
```diff
@@ -160,6 +190,19 @@ export class FileSystemFrontendContribution implements FrontendApplicationContri
         return this.toSelection(args[0]) ?? (Array.isArray(selection) ? selection.find(FileSelection.is) : this.toSelection(selection));
     };
 
+    /**
+     * Retrieves the folder from the arguments.
+     * @param args The arguments passed to a command.
+     * @returns A promise that resolves to a FileStat representing the folder, or undefined if the folder cannot be found.
+     */
+    protected async getFolderFromArgs(...args: unknown[]): Promise<FileStat | undefined> {
```
martin-fleck-at, line 198: nit: I'd think this method is very specific to our handling of the argument (first argument of an array must be a URI which is then converted) so I'd move it back into the calling method.

nit: Maybe we should use `directory` instead of folder since this seems the common terminology.

packages/filesystem/src/browser/filesystem-frontend-contribution.ts
```diff
@@ -134,6 +148,22 @@ export class FileSystemFrontendContribution implements FrontendApplicationContri
                 }
             }
         });
+        commands.registerCommand(CommonCommands.NEW_FILE, {
+            execute: async (...args: unknown[]) => {
+                let newFileName = undefined;
+                if (args !== undefined && typeof args[1] === 'string') {
```
martin-fleck-at, line 154: nit: The order in which we are processing the arguments seems arbitrary. I think resolving all arguments first might be a nice idea, e.g., index 0 before 1 etc. to get more clarity on the method.

packages/filesystem/src/browser/filesystem-frontend-contribution.ts
```diff
@@ -134,6 +148,22 @@ export class FileSystemFrontendContribution implements FrontendApplicationContri
                 }
             }
         });
+        commands.registerCommand(CommonCommands.NEW_FILE, {
+            execute: async (...args: unknown[]) => {
+                let newFileName = undefined;
```
martin-fleck-at, line 153: As someone who regularly needs to override methods, I'd really appreciate it if the execution of this command would be in it's own protected method. I know I'm being picky here but I think it makes also for nicer code.

packages/getting-started/src/browser/getting-started-widget.tsx
```diff
@@ -224,6 +234,7 @@ export class GettingStartedWidget extends ReactWidget {
 
         return <div className='gs-section'>
             <h3 className='gs-section-header'><i className={codicon('folder-opened')}></i>{nls.localizeByDefault('Open')}</h3>
```
martin-fleck-at, line 236: I wonder whether we should also call this section 'Start' instead of 'Open' as it is in VS Code as this command does not really reflect opening anything. What do you think?

martin-fleck-at (changes requested):
@dhuebner I really, really like the new set of changes! Now everything is very clear cut and the usability is also fantastic, so thank you very much!

I do have some minor comments on the code which you may want to address but the functionality is definitely there.

JonasHelming:
@dhuebner Do you plan to adress the comments? Would be great to have this in the upcoming release.

dhuebner:
@JonasHelming 
Yes, sure. I was sick last week, will apply the suggestions this week.

packages/getting-started/src/browser/getting-started-widget.tsx
dhuebner: I think it is a good idea.

packages/filesystem/src/browser/filesystem-frontend-contribution.ts
dhuebner: Moved to separate protected method

packages/filesystem/src/browser/filesystem-frontend-contribution.ts
dhuebner: Sure, I will change the args order anyway.

packages/filesystem/src/browser/filesystem-frontend-contribution.ts
dhuebner: I moved  the code into the new `handleNewFileCommand` function

packages/filesystem/src/browser/filesystem-frontend-contribution.ts
dhuebner: As the selection is checked inside the `UserWorkingDirectoryProvider` the order is:  
explicitly given argument  > user directory fallback.

Makes it simpler, I like it.

packages/core/src/browser/common-frontend-contribution.ts
dhuebner: Just to explain my motivation:  There is a `getSelection()` function in `filesystem-frontend-contribution.ts` which is used in some other handlers. The first argument is a current selected file URI. First I was about to re-use this function and then implemented an own and just kept the order. Another motivation is, that a file name is not enough, one also needs to know where to create this file, so the parent is kind of mandatory too.
What I'm really missing is a command parameter definition to expand the `...args` parameter in something meaningful. Like e.g. in vs-code API or many other command APIs or UI frameworks.

However, I will change the order if it feels more natural.

packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -88,11 +88,13 @@ export namespace WorkspaceCommands {
         category: WORKSPACE_CATEGORY,
         label: 'Close Workspace'
     });
+
```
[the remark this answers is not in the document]
dhuebner: Sure, to reduce the noise.

dhuebner:
@martin-fleck-at 
I've applied all the suggested changes, please have a look.

martin-fleck-at:
@dhuebner Thank you for the update, I'll have a look at this tomorrow!

martin-fleck-at (approved):
@dhuebner Fantastic, thank you very much for the great contribution! Everything looks good to me now!

PR #13350, opened 2024-02-06
(fix) bundle-size: reuse ts helper functions from tslib instead of in…

### LINKED ISSUE #13345: Shrink bundle by 500Kb with "importHelpers": true typescript setting
Typescript emits helper functions into produced js modules depending on features used.
It also has a compiler option  "importHelpers": true 
Which emits something like `import {decorate} from "tslib"` or require equivalent depending on module format.
Size of tslib module is ~16kb

On large codebases source code of emitted helpers per module tends to grow large and it is duplicated code all over the transpiled codebase.

For example: __decorate helper
```js
var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
    var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc, d;
    if (typeof Reflect === "object" && typeof Reflect.decorate === "function") r = Reflect.decorate(decorators, target, key, desc);
    else for (var i = decorators.length - 1; i >= 0; i--) if (d = decorators[i]) r = (c < 3 ? d(r) : c > 3 ? d(target, key, r) : d(target, key)) || r;
    return c > 3 && r && Object.defineProperty(target, key, r), r;
};
```
is 563 bytes.
In Theia packages we use to build our IDE it is encountered 952 times which gives us roughly 500kb of minified bundle size for only the decorate helper `copy-pasted` all over.
`@theia` scoped packages contribute 872 of those copies.

### Feature Description:
Please consider taking a dependency on tslib for theia packages and changing `importHelpers` compiler option to `true`
One decorate function is plenty enough for the whole bundle.

xcariba:
#### What it does

Fixes #13345

Reuses ts helper functions from tslib instead of inlining

#### How to test

Build master, check its size, build this branch, size will be smaller

#### Follow-ups

none

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/bulk-edit/package.json
+++ b/packages/bulk-edit/package.json
@@ -8,7 +8,8 @@
     "@theia/filesystem": "1.46.0",
     "@theia/monaco": "1.46.0",
     "@theia/monaco-editor-core": "1.83.101",
-    "@theia/workspace": "1.46.0"
+    "@theia/workspace": "1.46.0",
+    "tslib": "^2.6.2"
   },
   "publishConfig": {
     "access": "public"
--- a/configs/base.tsconfig.json
+++ b/configs/base.tsconfig.json
@@ -12,6 +12,7 @@
     "strictNullChecks": true,
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
+    "importHelpers": true,
     "downlevelIteration": true,
     "resolveJsonModule": true,
     "module": "CommonJS",
--- a/dev-packages/application-manager/package.json
+++ b/dev-packages/application-manager/package.json
@@ -59,6 +59,7 @@
     "source-map-support": "^0.5.19",
     "string-replace-loader": "^3.1.0",
     "style-loader": "^2.0.0",
+    "tslib": "^2.6.2",
     "umd-compat-loader": "^2.1.2",
     "webpack": "^5.76.0",
     "webpack-cli": "4.7.0",
--- a/dev-packages/application-package/package.json
+++ b/dev-packages/application-package/package.json
@@ -39,6 +39,7 @@
     "nano": "^10.1.3",
     "resolve-package-path": "^4.0.3",
     "semver": "^7.5.4",
+    "tslib": "^2.6.2",
     "write-json-file": "^2.2.0"
   },
   "devDependencies": {
--- a/dev-packages/cli/package.json
+++ b/dev-packages/cli/package.json
@@ -52,6 +52,7 @@
     "puppeteer-core": "19.7.2",
     "puppeteer-to-istanbul": "1.4.0",
     "temp": "^0.9.1",
+    "tslib": "^2.6.2",
     "yargs": "^15.3.1"
   },
   "devDependencies": {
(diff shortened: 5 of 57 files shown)


packages/bulk-edit/package.json
```diff
@@ -8,7 +8,8 @@
     "@theia/filesystem": "1.46.0",
     "@theia/monaco": "1.46.0",
     "@theia/monaco-editor-core": "1.72.3",
-    "@theia/workspace": "1.46.0"
+    "@theia/workspace": "1.46.0",
+    "tslib": "^2.6.2"
```
msujew, line 12: We don't need this for every package. It should be enough as long as `@theia/application-package` depends on `tslib`. This only adds maintenance effort, as we need to keep track of this dependency for every package.

msujew (commented):
Thanks for the contribution. The CLI actually also generates code that depends on `tslib`. Any project that just pulls `@theia/cli` without pulling `@theia/core` will break with this change. So you need to add the dependency there as well.

sdirix:
> Alternatively, we could think about only using this in a few selected packages that are guaranteed to get pulled like `@theia/core` and `@theia/application-package`.

Personally I would prefer having a clean dependency in each package. The additional effort is really minimal, because if we need to update all versions then a simple search-replace will do.

I'm not sure whether it's of practical relevance but not declaring the dependency in each package could lead to weird errors. For example a "random" `tslib` version being resolved for the ones which don't declare, potentially leading to errors.

msujew:
@sdirix Yeah, I can see that. I think the maintenance overhead is negligible and it probably makes sense to prevent weird dependency resolving errors.

Alexander-Taran:
@sdirix @msujew 
having it in dependencies is relevant only for cases when a package will get bundled or executed outside of scope where @theia/cli is pulled.

In case of /packages/ we want to ensure that tslib is present in node_modules at the time of bunlding the app.
Do dev-packages have a story being pulled on their own?

sdirix:
Hi @Alexander-Taran,

The error case described above is also relevant for the dev-packages. Dev packages like `@theia/application-manager` and `@theia/application-package` are consumed by `@theia/cli` and are therefore part of the `node_modules` of all downstream projects.

If only `@theia/cli` has a dependency to `tslib`, then only for the `@theia/cli` code itself it's guaranteed to consume the correct version of `tslib`. Once the code execution enters `@theia/application-manager` an incompatible `tslib` coming in from the downstream project might be consumed leading to errors.

Even if that would not be a use case: All published packages should have the correct set of dependencies as we can never predict what consumers will do with them.

Alexander-Taran:
@sdirix that's exactly what I meant in my comment to my original PR. That we (our team) don't know weather dev-packages need this dependency. And if so - which ones do.

msujew:
@Alexander-Taran Once you start setting the `importHelper` property to `true`, every package that is affected by this setting should have the `tslib` dependency. Otherwise, one might into dependency resolution errors (either during bundling or runtime).

Alexander-Taran:
@msujew @sdirix we've updated dev-packages also. (-:

Alexander-Taran:
💤

msujew (approved):
Looks good to me. Can you rebase so we can merge this?

Alexander-Taran:
> Looks good to me. Can you rebase so we can merge this?

@xcariba is in process

PR #13358, opened 2024-02-06
Minor improvements/fixes to custom plugin API doc

jcortell68:
#### What it does
Corrects/improves a few things in documentation and in one of the example Theia Extensions.

Found these by going through the steps in the recently updated documentation reflecting the new-and-improved mechanics for contributing custom plugin API.

#### How to test

* Read the updated doc and check for accuracy.
* Build the gotd example Theia extensions and ensure it builds and works

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/api-provider-sample/src/plugin/gotd-api-init.ts
+++ b/examples/api-provider-sample/src/plugin/gotd-api-init.ts
@@ -21,7 +21,7 @@ import type * as gotd from '../gotd';
 import { GreetingKind, GreetingExt, MAIN_RPC_CONTEXT } from '../common/plugin-api-rpc';
 import { GreetingExtImpl } from './greeting-ext-impl';
 import { Disposable, DisposableCollection } from '@theia/core';
-import { ApiFactory, PluginContainerModule } from '@theia/plugin-ext/lib/plugin/node/plugin-container-module';
+import { PluginContainerModule } from '@theia/plugin-ext/lib/plugin/node/plugin-container-module';
 
 // This script is responsible for creating and returning the extension's
 // custom API object when a plugin's module imports it. Keep in mind that
@@ -31,7 +31,6 @@ import { ApiFactory, PluginContainerModule } from '@theia/plugin-ext/lib/plugin/
 
 type Gotd = typeof gotd;
 const GotdApiFactory = Symbol('GotdApiFactory');
-type GotdApiFactory = ApiFactory<Gotd>;
 
 // Retrieved by Theia to configure the Inversify DI container when the plugin is initialized.
 // This is called when the plugin-host process is forked.
--- a/packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
+++ b/packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
@@ -89,9 +89,9 @@ import { FooExtImpl } from './foo-ext-impl';
 import * as fooBarAPI from '@bar/foo';
 
 type FooBarApi = typeof fooBarAPI;
+type Foo = FooBarApi['Foo'];
 
 const FooBarApiFactory = Symbol('FooBarApiFactory');
-type FooBarApiFactory = ApiFactory<FooBarApi>;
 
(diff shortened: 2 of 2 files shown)


examples/api-provider-sample/src/plugin/gotd-api-init.ts
```diff
@@ -31,7 +31,6 @@ import { ApiFactory, PluginContainerModule } from '@theia/plugin-ext/lib/plugin/
 
 type Gotd = typeof gotd;
 const GotdApiFactory = Symbol('GotdApiFactory');
-type GotdApiFactory = ApiFactory<Gotd>;
```
jcortell68, line 34: Explanation: this type seems extraneous.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -89,7 +89,6 @@ import { FooExtImpl } from './foo-ext-impl';
 import * as fooBarAPI from '@bar/foo';
 
 type FooBarApi = typeof fooBarAPI;
-type Foo = FooBarApi['Foo'];
```
cdamus, line 92: This type is used at line 132.

examples/api-provider-sample/src/plugin/gotd-api-init.ts
cdamus: So it is! There's no problem of unused name reported by tsc/eslint because the const symbol of the same name is referenced.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
```diff
@@ -89,7 +89,6 @@ import { FooExtImpl } from './foo-ext-impl';
...
 
 const FooBarApiFactory = Symbol('FooBarApiFactory');
 type FooBarApiFactory = ApiFactory<FooBarApi>;
```
cdamus, line 94: Similarly to the GotD example, this type definition is actually unused. We can remove it and the `ApiFactory` import as in GotD.

cdamus (commented):
Thanks for looking after this developer guide material! I do have a couple of comments in-line.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
jcortell68: Indeed. I apparently got myself confused there. Will fix

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
jcortell68: Above you pointed out it's used at line 132, and that does appear to be the case. So am I wrong about being wrong?

jcortell68: I guess part of the confusion is that both snippets are actually in the same file, and thus the contents of the file are spread out in two disjointed parts of the documentation.

jcortell68: So, where I got confused is that I mis-associated your comment here. My eye went to the pink-highlighted line (92) instead of the line directly above your comment (95). I should probably get more coffee. All is good. I'll fix the one issue

jcortell68:
OK. Hopefully it's all straight now. Thanks for the review.

packages/plugin-ext/doc/how-to-add-new-custom-plugin-api.md
cdamus: And we also need to remove the now-unused import of `ApiFactory` from the Foo example in the MD as we did for the GotD example in "real code".

jcortell68: Indeed. Done

cdamus (approved):
Great! Thanks, @jcortell68 .

martin-fleck-at (approved):
Thank you for double checking this and improving this example! Looks good to me!

Issue #13349, opened 2024-02-06
Theia's file service delete function fails to erase empty folders without recursive option


dfriederich:
When erasing just an empty folder with
app/node_modules/@theia/filesystem/src/browser/file-service.ts
's delete method, 

### Bug Description:

// folderPath points to empty folder, running in the front end
const folderUri = URI.fromFilePath(folderPath);
await fileService.delete(folderUri, { recursive: false })

this fails on Linux with an error that unlink is not for directories.
The issue is that the file-service is using in the end
packages/filesystem/src/node/disk-file-system-provider.ts's doDelete method:

```
>    protected async doDelete(filePath: string, opts: FileDeleteOptions): Promise<void> {
>        if (!opts.useTrash) {
>            if (opts.recursive) {
>                await this.rimraf(filePath);
>           } else {
>               await promisify(unlink)(filePath);
>            }
>        } else {
>            await trash(filePath);
>        }
>    }

```

When not using the trash support, and without recursive, then this maps to the nodejs unlink.
but unlink is not deleting directories :-(.
From its docu:
`     * `fs.unlink()` will not work on a directory, empty or otherwise. To remove a
     * directory, use {@link rmdir}.
`

### Steps to Reproduce:

1. use the front end file-service to remove an empty directory without the recursive option, without using the trash-can support.

- Operating System:
Linux (Ubuntu 22.04)
- Theia Version:
1.46.0

dfriederich:
I did create a pull request with a proposed fix.
https://github.com/eclipse-theia/theia/pull/13361/files

PR #13361, opened 2024-02-07
Fix non recursive folder delete using file system

dfriederich:
Fixes https://github.com/eclipse-theia/theia/issues/13349

Deleting a folder without recursive enabled failed as the used nodejs API unlink does not support folder deletion. Fixing it by first checking if it is a folder, and if so using rmdir instead.

Added a simple unit test, which also exposed that
the folder deletion also fails when using trash support, but that is a different issue.

#### What it does

Handle the deletion of a folder using the file system api.
Previously this did only work when asking to delete the folders content as well, now with recursive: false,
it does delete the folder if it is empty.

#### How to test

I did add a unit test, which fails without the fix.

#### Follow-ups

I did notice that when using the trash support, the deletion of a folder with content (recursive: true) did fail.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/node/disk-file-system-provider.ts
+++ b/packages/filesystem/src/node/disk-file-system-provider.ts
@@ -509,7 +509,12 @@ export class DiskFileSystemProvider implements Disposable,
             if (opts.recursive) {
                 await this.rimraf(filePath);
             } else {
-                await promisify(unlink)(filePath);
+                const stat = await promisify(lstat)(filePath);
+                if (stat.isDirectory() && !stat.isSymbolicLink()) {
+                    await promisify(rmdir)(filePath);
+                } else {
+                    await promisify(unlink)(filePath);
+                }
             }
         } else {
             await trash(filePath);
--- a/packages/filesystem/src/node/disk-file-system-provider.spec.ts
+++ b/packages/filesystem/src/node/disk-file-system-provider.spec.ts
@@ -26,7 +26,7 @@ import { promises as fs } from 'fs';
 import { join } from 'path';
 import * as temp from 'temp';
 import { v4 } from 'uuid';
-import { FilePermission, FileSystemProviderError, FileSystemProviderErrorCode } from '../common/files';
+import { FilePermission, FileSystemProviderCapabilities, FileSystemProviderError, FileSystemProviderErrorCode } from '../common/files';
 import { DiskFileSystemProvider } from './disk-file-system-provider';
 import { bindFileSystemWatcherServer } from './filesystem-backend-module';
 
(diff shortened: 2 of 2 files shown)


JonasHelming:
@dfriederich Hi! If you mention "fixed #13349" in the description, an automatic link to the issue is created.

dfriederich:
> Fixes #13349

You want me to change the description? Or I see you manually added the reference in this PR, was that sufficient?
If I change it, should I change the one commit and force push, or create a separate commit and then we squash the commit message when/if merging? Or can we just edit the commit message when merging?
Sorry, don't know how github is setup for theia.

JonasHelming:
Sorry for the confusion. What I meant is that if you mention "Fixes #xyz" in the description of the pull request (no necessarily the commit), Github will automatically link it on the issue. When I saw this PR I remembered that there was a corresponding issue. I searched for it and saw that you manually created a link over there so I thought, this information (auto linking) might be useful for you. We do not have strict rules about this, but traceability in both directions (PR<=>issue) is always good. Also, github will then automatically close the issue when this PR is merged.
Mentioning fixes #xyz is also good practice IMHO, but again we don't enforce this.

I believe Mark added the "fixes" tag in the beginning of the pull request description in the meantime.
I hope this makes sense! :-)

dfriederich:
@JonasHelming Thanks for the tip  :-).

packages/filesystem/src/node/disk-file-system-provider.ts
```diff
@@ -509,7 +509,12 @@ export class DiskFileSystemProvider implements Disposable,
             if (opts.recursive) {
                 await this.rimraf(filePath);
             } else {
-                await promisify(unlink)(filePath);
+                const stat = await promisify(lstat)(filePath);
+                if (stat.isDirectory() && !stat.isSymbolicLink()) {
```
jfaltermeier, line 513: I am testing on Ubuntu 22.04 and it looks very good. I noticed one weird thing however. 

Given that `/tmp/foo/Untitled` is a directory and a symbolic link created with `ln -s`. 

`lsstat` has different results depending on whether the path ends with a / or without. 
 
with a trailing `/` `stat.isDirectory()` is `true` while `stat.isSymbolicLink()` is `false`. 
without a trailing `/` `stat.isDirectory()` is `false` while `stat.isSymbolicLink()` is `true`. 

This means that
```
                const folderUri = URI.fromFilePath('/tmp/foo/Untitled/');
                this.fileService.delete(folderUri, { recursive: false });
```
fails, while
```
                const folderUri = URI.fromFilePath('/tmp/foo/Untitled');
                this.fileService.delete(folderUri, { recursive: false });
```
is working. 

Maybe we have to normalize the file path by removing trailing `/` first?

Do you see this behavior on your machine as well?

dfriederich: Yes, I see the same behavior, but I actually also have Ubuntu 22.04 :-).

if symlink.to.test.folder is a symlink to a folder, I would think that 'rm symlink.to.test.folder/'
asks to delete the folder the symlink links to, whereas 'rm symlink.to.test.folder' asks to remove the symbolic link itself. But for example when experimenting with bash:

`d.friederich@xdfriederich:.../packages/filesystem$rmdir /tmp/tt/symlink.to.test.folder/
rmdir: failed to remove '/tmp/tt/symlink.to.test.folder/': Not a directory
d.friederich@xdfriederich:.../packages/filesystem$rm /tmp/tt/symlink.to.test.folder/
rm: cannot remove '/tmp/tt/symlink.to.test.folder/': Is a directory
d.friederich@xdfriederich:.../packages/filesystem$stat /tmp/tt/symlink.to.test.folder/
  File: /tmp/tt/symlink.to.test.folder/
  Size: 4096      	Blocks: 8          IO Block: 4096   directory
Device: 10302h/66306d	Inode: 25834198    Links: 2
Access: (0755/drwxr-xr-x)  Uid: (503046/d.friederich)   Gid: (103210/d.friederich_g)
Access: 2024-02-08 13:42:01.011807685 -0600
Modify: 2024-02-08 13:42:01.011807685 -0600
Change: 2024-02-08 13:42:01.011807685 -0600
 Birth: 2024-02-08 13:42:01.011807685 -0600
`

so for stat and rm, symlink.to.test.folder/ is referencing a directory, but for rmdir it is not a directory.
So a bit inconsistent. I don't think we should drop the trailing / in this particular case, IMHO asking to delete symlink.to.test.folder/ is not the same as asking to delete symlink.to.test.folder
The current behavior is following what the command line tools do, it fails. In this particular case, both rm and rmdir are failing.

 (The ubuntu rm actually has an option to make it allow to remove an empty directory -d, but in this case it fails because it is not a directory. so rm is inconsistent within itself :-)
 `d.friederich@xdfriederich:.../packages/filesystem$rm /tmp/tt/symlink.to.test.folder/
rm: cannot remove '/tmp/tt/symlink.to.test.folder/': Is a directory
d.friederich@xdfriederich:.../packages/filesystem$rm -d /tmp/tt/symlink.to.test.folder/
rm: cannot remove '/tmp/tt/symlink.to.test.folder/': Not a directory
`
)

dfriederich: So I think the behavior of failing to remove symlink.to.test.folder/ is OK.

jfaltermeier (approved):

PR #13364, opened 2024-02-07
Fix race condition in notebook kernel association

msujew:
#### What it does

There's a race condition in the notebook kernel association code. Whenever you open a notebook, Theia will automatically try to associate a kernel to it. However, if this notification arrives in the plugin host before the notebook open notification arrives, it will not be able to associate the kernel to the notebook - this effectively disables all cell execution for the specific notebook.

#### How to test

The issue can be fairly reliably reproduced by:

1. Open a notebook and select a kernel for it
2. Just reload the application, with the notebook being open

Previously, notebook cells weren't executing afterwards. This change should alleviate that issue.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/notebook/notebooks.ts
+++ b/packages/plugin-ext/src/plugin/notebook/notebooks.ts
@@ -312,6 +312,26 @@ export class NotebooksExtImpl implements NotebooksExt {
         return result;
     }
 
+    waitForNotebookDocument(uri: TheiaURI, duration = 2000): Promise<NotebookDocument> {
+        const existing = this.getNotebookDocument(uri, true);
+        if (existing) {
+            return Promise.resolve(existing);
+        }
+        return new Promise<NotebookDocument>((resolve, reject) => {
+            const listener = this.onDidOpenNotebookDocument(event => {
+                if (event.uri.toString() === uri.toString()) {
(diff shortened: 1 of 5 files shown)


jonah-iden:
Thanks for the fix, i'll take a look tomorrow

packages/plugin-ext/src/plugin/notebook/notebooks.ts
```diff
@@ -312,6 +312,26 @@ export class NotebooksExtImpl implements NotebooksExt {
         return result;
     }
 
+    waitForNotebookDocument(uri: TheiaURI, duration = 2000): Promise<NotebookDocument> {
```
jonah-iden, line 315: should this timeout maybe be configurable somewhere?

jonah-iden (approved):
Works great and looks good to me 👍

packages/plugin-ext/src/plugin/notebook/notebooks.ts
msujew: I don't think that's necessary. The race condition is usually in a fairly small time frame, on local host it's roughly ~10ms. Having a larger timeout should rarely be an issue, while 2000ms is probably enough for most if not all use cases.

PR #13366, opened 2024-02-07
Update notebook execution timer

msujew:
#### What it does

Does essentially as the title says. In vscode, the execution time display is updated every 100ms to show that the execution is still running. We just do the same.

#### How to test

1. Open a notebook and select a kernel
2. Start a long running execution, see:
```py
import time
print('1')

time.sleep(2)

print('2')

time.sleep(2)

print('3')
```
3. The timer should behave as expected

#### Review checklist

#### Reminder for reviewers


jonah-iden (commented):
out of some reason its starting the counter at 2 seconds when executing a cell 

https://github.com/eclipse-theia/theia/assets/34068281/97aad53f-2b31-4d0e-aaeb-d7058b6113a7

so in this example the timer ends 2 seconds after the actual execution time

jonah-iden (approved):
works like a charm now

PR #13367, opened 2024-02-07
fix: configure tasks from open task quick pick

### LINKED ISSUE #13086: Cannot configure provided tasks
### Bug Description:

In a Theia application that has VS Code extensions providing tasks (e.g. for npm, grunt, etc.), they are correctly shown when executing *Run Task...*. However, when I select a provided task folder (e.g. npm), I cannot click on the cogwheel to configure them in my `tasks.json`.

### Steps to Reproduce:

1. Open Blueprint (or any other Theia application with VS Code extension that contribute provided tasks)
2. Open a workspace where a provided task is discoverable (e.g. a folder with an npm module, like a Theia application)
3. Select *npm*
4. Try to click on the cogwheel

https://github.com/eclipse-theia/theia/assets/588090/3bfd4ca3-1c27-4ae5-bda8-ae80960b2956

What happens: nothing
What should happen: the provided task is copied over to tasks.json for further customization.

### Additional Information

- Operating System: Ubuntu Linux
- Theia Version: 1.43.0

AlexandraBuzila:
#### What it does

The gear icon in the task quick pick should open the task configuration.

Fixes #13086

#### How to test

1. Open the task quick pick from _>Task: Run Task..._
2. For any configured task, press the gear icon  → the `tasks.json` configuration file should open

Or:
1. Open the task quick pick from _>Task: Run Task..._
2. Select a task category from the dialog, e.g. `npm`
4. In the list of tasks, press the gear icon → the `tasks.json` configuration file should open

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/task/src/browser/quick-open-task.ts
+++ b/packages/task/src/browser/quick-open-task.ts
@@ -182,6 +182,7 @@ export class QuickOpenTask implements QuickAccessProvider {
         picker.matchOnDescription = true;
         picker.ignoreFocusOut = false;
         picker.items = this.items;
+        picker.onDidTriggerItemButton(({ item }) => this.onDidTriggerGearIcon(item));
 
         const firstLevelTask = await this.doPickerFirstLevel(picker);
 
@@ -225,7 +226,10 @@ export class QuickOpenTask implements QuickAccessProvider {
             execute: () => this.showMultiLevelQuickPick(true)
         }));
 
-        this.quickInputService?.showQuickPick(providedTasksItems, { placeholder: CHOOSE_TASK });
+        this.quickInputService?.showQuickPick(providedTasksItems, {
+            placeholder: CHOOSE_TASK,
+            onDidTriggerItemButton: ({ item }) => this.onDidTriggerGearIcon(item)
+        });
     }
 
     attach(): void {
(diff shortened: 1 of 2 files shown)


packages/task/src/browser/quick-open-task.ts
```diff
@@ -182,6 +182,7 @@ export class QuickOpenTask implements QuickAccessProvider {
         picker.matchOnDescription = true;
         picker.ignoreFocusOut = false;
         picker.items = this.items;
+        picker.onDidTriggerItemButton(({ item }) => this.onDidTriggerGearIcon(item as QuickPickItem));
```
tsmaeder, line 185: It seems odd that we have to case here, but not below. Any idea why?

AlexandraBuzila: The two methods have different signatures :smile: 
Here the item can be a separator as well, even if it won't happen (separators  won't have buttons that will trigger this method). However, to make this nicer I updated the code and added a check here, instead.

tsmaeder: That was the question: is it true that one time the button item can be a separator as well or is that a mistake in the typing of "picker.onDidTriggerItemButton()"?

AlexandraBuzila: Hi @tsmaeder! No, the item passed to the button cannot be a separator. The separator only has a label and no buttons. I restricted the type in the interface for this method. Ideally, we would have a dedicated type for a separator picker, that does not have this method at all, but I think that refactoring is out of scope for this PR.

tsmaeder (approved):
:lgtm:

PR #13369, opened 2024-02-08
Support plugin copying for remote feature

msujew:
#### What it does

This change adds support for copying plugins from the host system to the remote system. Works in two ways:

1. Builtin plugins are transmitted directly through SSH
2. User installed plugins are installed via the new `--install-plugins` argument. This new runtime CLI argument automatically installs the specified extensions from open-vsx.

#### How to test

0. (optional) install some plugins from open-vsx.
1. Use the SSH remote feature to connect to a remote system.
2. Both builtin and manually installed plugins should have been transmitted to the remote system. You can open the vsx-registry view to confirm this.

#### Review checklist

#### Reminder for reviewers


jonah-iden (commented):
Some things i noticed while testing. 
- It seems to not be able to install all extensions
There were two extensions missing which were install from VSX previously for testing: `zaaack.markdown-editor` and `ms-toolsai.jupyter-keymap`
- of the builtin extensions only 26 of 83 were installed. Even basic things like css and dart language basics
- of course extensions installed through `install from vsix` were not installed. That would probably take additional significant effort right?

jonah-iden:
ok i take that back with the not installed extensions it seems it just did't update correctly. Maybe it would be nice to have a installing plugins progress message

jonah-iden:
It's nice that launch args can be added via a contribution point, but the way the `--install-plugins` option is added currently, its kind of hard to modify it from other theia packages. Like for example the dev-container support should later on add the ability to install specific extensions in the container.

tsmaeder:
@msujew have you considered:

1. Extensions that are installed from a *.vsix file: the may not be available on open-vsx
2. Maybe the remote ide does not have the same access to the internet my local machine had when I installed the *.vsix

That considered, wouldn't it make more sense to ship _all_ local plugins to the remote ide?

msujew:
> That considered, wouldn't it make more sense to ship all local plugins to the remote ide?

I've thought about that, but the main issue here lies with platform specific plugins: If I'm running windows, I will run the native rust-analyzer plugin binaries. Shipping those to the remote ide will stop the plugin from working.

tsmaeder:
@jonah-iden what's the current state of this one?

jonah-iden:
@tsmaeder to be honest im not totally sure. If this is not urgently needed i would wait for mark to get back from vacation

msujew:
@jonah-iden I just rebased on top of Thomas' changes regarding the `--install-plugin` cli argument. I kindly ask for a retest :)

> of course extensions installed through install from vsix were not installed. That would probably take additional significant effort right?

Right. It's possible in theory (we can just repackage the unpacked plugin) but I would implement that once it is actually required. 

> of the builtin extensions only 26 of 83 were installed. Even basic things like css and dart language basics

I haven't been able to reproduce this. In all of my tests, all 85 builtins were successfully copied to the remote.

jonah-iden (approved):
Works great. All my installed extensions were copied perfectly fine. 
Code looks good to me

PR #13380, opened 2024-02-12
electron: allow accessing the metrics endpoint for performance analysis

xai:
#### What it does

By default, when running Theia in Electron, all endpoints are protected by the ElectronTokenValidator.
This patch allows accessing the '/metrics' endpoint without a token, which enables us to collect metrics for performance analysis.

For this, ElectronTokenValidator is extended to allow access to the metrics endpoint. All other endpoints are still protected.

#### How to test

* Build and start the Electron example. To avoid getting a random free port on runtime, use `yarn electron start --port=3000`.
* Access the '/metrics' endpoint, e.g., by calling `curl http://localhost:3000/metrics`.
* Verify that metrics are returned.
* Access a different endpoint, e.g. `http://localhost:3000/foo`.
* Verify that the request is denied.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/metrics/src/electron-node/electron-metrics-backend-module.ts
+++ b/packages/metrics/src/electron-node/electron-metrics-backend-module.ts
@@ -1,0 +1,24 @@
+// *****************************************************************************
+// Copyright (C) 2023 STMicroelectronics and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { ContainerModule } from '@theia/core/shared/inversify';
+import { MetricsElectronTokenValidator } from './electron-token-validator';
+import { ElectronTokenValidator } from '@theia/core/lib/electron-node/token/electron-token-validator';
+
+export default new ContainerModule((bind, unbind, isBound, rebind) => {
+    bind(MetricsElectronTokenValidator).toSelf().inSingletonScope();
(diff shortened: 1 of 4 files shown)


packages/metrics/src/electron-node/electron-metrics-backend-module.ts
```diff
@@ -0,0 +1,41 @@
...
+import { ElectronTokenValidator } from '@theia/core/lib/electron-node/token/electron-token-validator';
+import { BackendApplicationContribution } from '@theia/core/lib/node';
+import { ExtensionMetricsContribution, MetricsBackendApplicationContribution, MetricsContribution, NodeMetricsContribution } from '../node';
+import { ConnectionHandler, RpcConnectionHandler, bindContributionProvider } from '@theia/core';
+import { measurementNotificationServicePath } from '../common';
+import { MeasurementMetricsBackendContribution } from '../node/measurement-metrics-contribution';
+
+export default new ContainerModule((bind, unbind, isBound, rebind) => {
+    bind(MetricsElectronTokenValidator).toSelf().inSingletonScope();
+    rebind(ElectronTokenValidator).to(MetricsElectronTokenValidator);
```
sdirix, line 23: The only difference to the regular backend is the electron token validator rebinding. I would like to suggest moving this into a separate `ContainerModule` and in the `package.json`'s `theiaExtensions` add a new entry only containing the `backendElectron` entry.

This way we could avoid all of the duplicated binding code as the new Container would only contain the electron token validation specifics.

sdirix (commented):
This works, however we should think about whether we really want to expose the `metrics` endpoint by default in Electron. I think it would be more reasonable to only do the rebinding in case we are in debug or development mode. We already only collect startup metrics in development mode, so we could align with that.

Rebinding the `ElectronTokenValidator` will also often not have an effect in downstream projects. It's a common pattern to rebind it to allow for custom exposed APIs, so they will overwrite this one again. We should probably have a contribution point within the `ElectronTokenValidator` to allow contributing whitelists. Of course this is out of scope for this PR, but we should create a ticket for that.

packages/metrics/src/electron-node/electron-metrics-backend-module.ts
sdirix: Already resolved :+1:

xai:
Thanks, @sdirix and @tortmayr! I will revise my PR accordingly.

tortmayr (approved):
Looks good to me and works as expected.
One minor regarding imports but otherwise everything works.

JonasHelming:
@tortmayr Can you merge this?

PR #13388, opened 2024-02-14
[vscode] Add support for TestRunProfile onDidChangeProfile event

### LINKED ISSUE #13354: [vscode] TestRunProfile now fires an event when isDefault is changed with 1.86
`TestRunProfile` can now provide a `onDidChangeDefault` event when `isDefault` is changed.

see https://github.com/microsoft/vscode/blob/e5447598324c57ac814c150148e7e4c830e0f9ee/src/vscode-dts/vscode.d.ts#L17156

rschnekenbu:
#### What it does

This pull request add the support of API introduced with VS Code 1.86 on TestRunProfile, the onDidChangeDefault event.
It also adds the capability of changing default profile for a given test controller and a given kind of test run. The _isDefault_ value is now updated as soon as the command "Select Default Test Profiles..." is used in the tool.

Fixes #13354

#### How to test

Install the following extension, mainly inspired from https://github.com/KevinClapperton/Test-API-functional-test
- src: 
[Test-API-functional-test-0.0.6-src.zip](https://github.com/eclipse-theia/theia/files/14281002/Test-API-functional-test-0.0.6-src.zip)

- zip vsix:
[test-api-functional-test-0.0.6.zip](https://github.com/eclipse-theia/theia/files/14281003/test-api-functional-test-0.0.6.zip)

This extension will add a test controller and a command to create a test run profile, install a bunch of test and create a test run. 
1. When extension is installed, the "Test API: Run all tests" command can be invoked
2. It will create 2 run profiles, and some tests. A couple of message windows will pop up.
3. Then it is possible to switch between default profiles using the "Select Default Test Profiles..." in the command palette. 
4. The app should notify that _isDefault_ has changed for the profiles impacted by the choice, i.e. the profiles from same controller that have the right kind as selected in the first pick dialog. Only one profile per controller and kind can be activated at once, on the contrary of VS code implementation. This latter implementation is a bug IMO, as only one profile will be chosen to run the tests, so having several default profiles is not relevant. 

testing-default-profile

Note: the set `isDefault` value is not kept in settings, so user may need to configure it each time the app is started. This behavior is similar to the one on VS Code. 

#### Follow-ups

There are no follow ups known currently.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/common/plugin-api-rpc.ts
+++ b/packages/plugin-ext/src/common/plugin-api-rpc.ts
@@ -2186,6 +2186,9 @@ export interface TestingExt {
     /** Configures a test run config. */
     $onConfigureRunProfile(controllerId: string, profileId: string): void;
 
+    /** Sets the default on a given run profile */
+    $onDidChangeDefault(controllerId: string, profileId: string, isDefault: boolean): void;
+
     $onRunControllerTests(reqs: TestRunRequestDTO[]): void;
 
     /** Asks the controller to refresh its tests */
--- a/packages/plugin-ext/src/plugin/tests.ts
+++ b/packages/plugin-ext/src/plugin/tests.ts
@@ -417,6 +417,14 @@ export class TestingExtImpl implements TestingExt {
     }
 
     /** @inheritdoc */
+    $onDidChangeDefault(controllerId: string, profileId: string, isDefault: boolean): void {
+        const profile = this.controllersById.get(controllerId)?.getProfile(profileId);
+        if (profile) {
+            profile.doSetDefault(isDefault);
+        }
+    }
+
+    /** @inheritdoc */
     async $refreshTests(controllerId: string, token: CancellationToken): Promise<void> {
         await this.withController(controllerId).refreshHandler?.(token);
     }
@@ -483,8 +491,29 @@ export class TestRunProfile implements theia.TestRunProfile {
     @observableProperty('notifyPropertyChange')
     label: string;
 
-    @observableProperty('notifyPropertyChange')
-    isDefault: boolean;
+    _isDefault: boolean;
+
+    get isDefault(): boolean {
+        return this._isDefault;
+    }
+
+    set isDefault(isDefault: boolean) {
+        if (this.doSetDefault(isDefault)) {
+            this.proxy.$updateTestRunProfile(this.controllerId, this.profileId, { isDefault: isDefault });
+        }
+    }
+
+    doSetDefault(isDefault: boolean): boolean {
+       if (this._isDefault !== isDefault) {
+            this._isDefault = isDefault;
+            this.onDidChangeDefaultEmitter.fire(isDefault);
+            return true;
+        }
+        return false;
+    }
+
+    private onDidChangeDefaultEmitter = new Emitter<boolean>();
+    onDidChangeDefault = this.onDidChangeDefaultEmitter.event;
 
     @observableProperty('notifyTagChange')
     tag: theia.TestTag | undefined;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ## v1.47.0 not yet released
 
 - [plugin] Add command to install plugins from the command line [#13406](https://github.com/eclipse-theia/theia/issues/13406) - contributed on behalf of STMicroelectronics
+- [testing] support TestRunProfile onDidChangeDefault introduced in VS Code 1.86.0 [#13388](https://github.com/eclipse-theia/theia/pull/13388) - contributed on behalf of STMicroelectronics
 
 <a name="breaking_changes_not_yet_released">[Breaking Changes:](#breaking_changes_not_yet_released)</a>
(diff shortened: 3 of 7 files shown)


packages/plugin-ext/src/common/plugin-api-rpc.ts
```diff
@@ -2173,6 +2173,9 @@ export interface TestingExt {
     /** Configures a test run config. */
     $onConfigureRunProfile(controllerId: string, profileId: string): void;
 
+    /** Sets the default on a given run profile */
+    $onChangeDefault(controllerId: string, profileId: string, isDefault: boolean): void;
```
tsmaeder, line 2177: Not really correct according to our naming conventions ("onDidChange..."), but 1. it's consistent with the rest of the interface and 2. I invented it 🤦 .

packages/plugin-ext/src/plugin/tests.ts
```diff
@@ -483,9 +491,18 @@ export class TestRunProfile implements theia.TestRunProfile {
...
     label: string;
 
-    @observableProperty('notifyPropertyChange')
+    @observableProperty('notifyIsDefaultChange')
     isDefault: boolean;
 
+    private onDidChangeDefaultEmitter = new Emitter<boolean>();
+    onDidChangeDefault = this.onDidChangeDefaultEmitter.event;
+
+    protected notifyIsDefaultChange(_property: keyof TestRunProfileDTO, isDefault: boolean): void {
```
tsmaeder, line 500: I suspect this will generate an extra call to the front end:

1. Front set "isDefault"
2. TestingExt.$onChangeDefault() is called
3. notifyDefaultChanged is invoked and sends `$updateTestRunProfile()`

I don't think the "observableProperty" mechanism works for this: we need to distinguish where a change is coming from.

rschnekenbu: I updated to avoid this circular call. Only one call shall be made now.

rschnekenbu:
Hi @tsmaeder, thanks for your comments. I will solve conflicts and rebase to master once the review is done.

packages/plugin-ext/src/plugin/tests.ts
```diff
@@ -483,8 +491,30 @@ export class TestRunProfile implements theia.TestRunProfile {
...
+    }
+
+    set isDefault(isDefault: boolean) {
+        if(this.doSetDefault(isDefault)) {
+            this.proxy.$updateTestRunProfile(this.controllerId, this.profileId, { isDefault: isDefault });
+        }
+    }
+
+    doSetDefault(isDefault: boolean): boolean {
+        let change = false;
```
tsmaeder, line 507: Why not just 
```
if (this._isDefault !== isDefault) {
            this._isDefault = isDefault;
            this.onDidChangeDefaultEmitter.fire(isDefault);
            return false;
        }
        return false;

tsmaeder:
@rschnekenbu I'm not seeing the test controller after I install the attached extension. Also, does the ""Select default test Profile" se t the `isDefault` property on the profile _from the extension_?

rschnekenbu:
@tsmaeder, the extension is only activated when you run the command 'Test API: Run all tests'. Do you have any error on the console? There should be an entry in the log that indicates the extension was successful to create the controller ( _SUCCESS TestController created: AllTestAPITestController_ )
The _isDefault_ is not set from the extension.  The extension reacts on change, but the change is changed from main.

packages/plugin-ext/src/plugin/tests.ts
rschnekenbu: Good one! Some remnants of old checkstyles rules :/ That will be fixed.

tsmaeder:
@rschnekenbu I am having trouble testing this change: could you please give concrete steps how I change the default profile and how I can verify that a default profile is set? Secondly, we really _should_ have a command that sets the default profile from the plugin side: otherwise this code path can't be verified.

rschnekenbu:
@tsmaeder, here is an updated version of the extension with a timeout function that sets the first profile to default after a small delay (~3sec). This will test the extension side of setting the default value.
- src:
[test-api-functional-test-0.0.8-src.zip](https://github.com/eclipse-theia/theia/files/14431647/test-api-functional-test-0.0.8-src.zip)

- zip vsix: 
[test-api-functional-test-0.0.8.zip](https://github.com/eclipse-theia/theia/files/14431637/test-api-functional-test-0.0.8.zip)

For the steps:
- Once the extension is installed, trigger from the command palette 'Test API: Run all Test'. This will create the test controller, test profiles, test items and test runs.
- Once the command was run, you should have the AllTestAPITestController in the test explorer.
- You should also get 2 notifications, one that will say that the testRunProfile could be created, and the second one that indicates that the profile.isDefault property was changed from the extension.
- To change the default values, there is an entry on contextual menu of the TestItems "Select Default Test Profiles...". You can then select 'Run' or 'Debug' kind for the test profiles to configure. The list of test profiles with the right kind will show up. Any default test profile will have (default) append to their label. You can select the one you want to set as default.

tsmaeder (approved):

Issue #13384, opened 2024-02-13
Add trigger for automatic (next) releases


msujew:
While it's theoretically possible to consume Theia directly from source (`master` branch) to test new features/bug fixes in adopter projects, I would like to have an option to trigger an automatic `next` release to npm. We used to have something like this on every commit, but it became quite cumbersome, which is why we've disabled that.

We can also reuse that workflow to perform the actual release process after tagging/releasing a specific commit. We use this in Langium as well, see [here](https://github.com/eclipse-langium/langium/blob/main/.github/workflows/publish.yml).

msujew:
@jfaltermeier Since you're our current release engineer, what is your opinion on that? It would probably make your work a bit easier :)

jfaltermeier:
I agree that having the release build as a workflow would be good. This would remove any possible user errors, like wrong node versions or something like that. 
Also having npm provenance would be nice. 

Regarding the `next` builds. I like having `next` builds available, but as far as I remember we stopped with this because we did not want to spam the npm registry with all those versions nearly nobody was using. 
Having e.g. a weekly next versions might be a good compromise for allowing end users to test early but not spamming on every commit. 

I was also experimenting with with a dockerized verdaccio registry, allowing to build Theia locally and get it published to a local registry with just a single docker command. I plan to continue on this. This might then be a good starting point for users that want to test a specific Theia commit.

PR #13399, opened 2024-02-19
Automated release publishing

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/13384

Adds a GitHub action that is triggered on every GitHub release. It automatically publishes the code that is associated with the release. We have a very similar mechanism in [Langium](https://github.com/eclipse-langium/langium/blob/main/.github/workflows/publish.yml). Uses the `NPM_CONFIG_PROVENANCE` env variable to provide provenance. That seems to work with Lerna, see https://github.com/lerna/lerna/issues/3657.

#### How to test

We will see during the next release whether this works as expected. But since the `Publish Next` GitHub Action already works as expected, I don't see too much room for error.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -4,8 +4,16 @@ permissions:
   id-token: write
 
 on:
-  release:
-    types: [published]
+  workflow_dispatch:
+    inputs:
+      release_type:
+        description: 'Release Type'
+        required: true
+        default: 'minor'
+        type: choice
+        options:
+          - 'minor'
+          - 'patch'
 
 jobs:
   publish:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,45 @@
...
+    types: [published]
+
+jobs:
+  publish:
+    name: Perform Publishing
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
jfaltermeier, line 17: Just for my understanding. The `release` trigger will make sure that the tagged commit from the release will be checked out here? (Probably via `GITHUB_SHA` env? https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release)

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,45 @@
...
+          yarn global add node-gyp
+          yarn --skip-integrity-check --network-timeout 100000
+        env:
+          NODE_OPTIONS: --max_old_space_size=4096
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # https://github.com/microsoft/vscode-ripgrep/issues/9
+
+      - name: Publish NPM
+        shell: bash
+        run: |
+          yarn publish:latest
```
jfaltermeier, line 42: Last time I ran the release `yarn publish:latest` was an interactive command. 
I had to select the incremented version, which is then written to the package.json files. I think we might have to pass this via an argument to the workflow?
Also `packages/core/README.md` is update and the package.json changes have to be committed: 
https://github.com/eclipse-theia/theia/blob/master/doc/Publishing.md#release

I think the translation workflow opens PRs with changes atm (https://github.com/eclipse-theia/theia/blob/master/.github/workflows/translation.yml), so the required changes might be similar

msujew: I see, thanks. I've updated the PR to make this into a workflow dispatch:
1. Allows to select minor/patch version for release.
2. After the publishing is done, it will automatically create a PR from the changes.

jfaltermeier (approved):
Looks good to me, thanks. 
I've tried to test `yarn publish:latest -- patch` and `yarn publish:latest -- minor` with a local verdaccio registry a bit and results looked good.

msujew:
Just to make sure, @jfaltermeier do you plan to try out the new workflow to publish the new version on Thursday? If that has proven to work, I'll update the documentation to reflect the modified publishing process.

jfaltermeier:
Yes I plan to try it out.

jfaltermeier:
@msujew Ok, looks like the publishing failed before all packages have been published (at least I think so): https://github.com/eclipse-theia/theia/actions/runs/8096588531/job/22125663621

I can't see any useful error message unfortunately. 

I will try to publish the missing pieces

jfaltermeier:
```
Successfully published:
 - @theia/plugin-metrics@1.47.0
 - @theia/remote@1.47.0
lerna success published 2 packages
Done in 85.58s.
```

I think we should at least update the workflow to always create a PR with the changes.

msujew:
@jfaltermeier Hm, weird. I would've expected some error message at least. But that's good to know, a publishing process error shouldn't stop the PR from being created.

Issue #13085, opened 2023-11-21
Support to capture terminal contents


kavitha-subr:
### Feature Description:

As a developer, I would like the ability to capture the contents of the terminal so that I can identify errors and warnings for further action.
Currently, the only possible way to get terminal contents is to spawn a new terminal and have an event listener listen to the output data. I would like to have this feature without having to create a new terminal.

kavitha-subr:
Yup, this looks great!

tsmaeder:
I've made a branch at https://github.com/tsmaeder/theia/tree/13085_terminal_observer with the necessary changes. Here's the source for a plugin using the API: 

[terminalobserver.zip](https://github.com/eclipse-theia/theia/files/13460830/terminalobserver.zip)

tsmaeder:
To test, you'll have to build & run theia from the branch. To build the plugin, you will have to yarn link the `@theia/plugin` package. In order to do so, you have to 

1. go into the `<theia>/packages/plugin` folder and type `yarn link`
2. go in the root folder of the example plugin source and type 'yarn link @theia/plugin`

Then you can just build the plugin and copy the resulting *.theia file into the theia plugins folder.

kavitha-subr:
Thank you so much! I'll verify if everything works as expected.

PR #13402, opened 2024-02-20
Add terminal observer API

tsmaeder:
#### What it does
This is a draft PR for feedback on https://github.com/eclipse-theia/theia/issues/13085

#### How to test
The attached theia plugin adds two commands: "Register Observer" and "Remove Observer". The first one allows to register a terminal observer with a given name, regular expression to match and a number of lines to match. The second command unregisters obeservers registered before by name.

The registered observers write the matched strings to the Theia back end log.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/terminal/src/browser/terminal-widget-impl.ts
+++ b/packages/terminal/src/browser/terminal-widget-impl.ts
@@ -29,7 +29,8 @@ import { IBaseTerminalServer, TerminalProcessInfo, TerminalExitReason } from '..
 import { TerminalWatcher } from '../common/terminal-watcher';
 import {
     TerminalWidgetOptions, TerminalWidget, TerminalDimensions, TerminalExitStatus, TerminalLocationOptions,
-    TerminalLocation
+    TerminalLocation,
+    TerminalBuffer
 } from './base/terminal-widget';
 import { Deferred } from '@theia/core/lib/common/promise-util';
 import { TerminalPreferences } from './terminal-preferences';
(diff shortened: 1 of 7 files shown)


tsmaeder:
Here's the plugins

[terminalobserver.zip](https://github.com/eclipse-theia/theia/files/14436401/terminalobserver.zip)

and here's the source:

[terminalobserver.zip](https://github.com/eclipse-theia/theia/files/14436415/terminalobserver.zip)

packages/terminal/src/browser/terminal-widget-impl.ts
```diff
@@ -60,6 +61,23 @@ export interface TerminalContribution {
     onCreate(term: TerminalWidgetImpl): void;
 }
 
+class TerminalBufferImpl implements TerminalBuffer {
```
jonah-iden, line 64: suggestion: Maybe move this and the interface to its own file. The files they are in are allready pretty long

jonah-iden (approved):
All in all seems to work great and code looks good to me.

just something i found a little weird but i don't know if its a problem. Following console output
```
gitpod /workspace/theia (13085_terminal_observer) $ testx
bash: testx: command not found
gitpod /workspace/theia (13085_terminal_observer) $ 
```
having an observer with `.*testx.*` for one line only prints
```
["gitpod /workspace/theia (13085_terminal_observer) $ testx                                                                     "]```
But with two lines it prints
```
```
["gitpod /workspace/theia (13085_terminal_observer) $ testx                                                                     "]
["bash: testx: command not found                                                                                                "]
["bash: testx: command not found                                                                                                "]
```
I would have expected for one line it would print the command not found once and for two lines maybe twice because one for when the output is first printed and one when the input prompt is back again since it scans or none for 1 line and once for two lines

tsmaeder:
> All in all seems to work great and code looks good to me.
> 
> just something i found a little weird but i don't know if its a problem. Following console output
> 
> ```
> gitpod /workspace/theia (13085_terminal_observer) $ testx
> bash: testx: command not found
> gitpod /workspace/theia (13085_terminal_observer) $ 
> ```
> 
> having an observer with `.*testx.*` for one line only prints
> 
> ```
> ["gitpod /workspace/theia (13085_terminal_observer) $ testx                                                                     "]```
> ```
> 
> But with two lines it prints
> 
> ```
> ["gitpod /workspace/theia (13085_terminal_observer) $ testx                                                                     "]
> ["bash: testx: command not found                                                                                                "]
> ["bash: testx: command not found                                                                                                "]
> ```
> 
> I would have expected for one line it would print the `command not found` once and for two lines maybe twice because one for when the output is first printed and one when the input prompt is back again or none for 1 line and once for 2 lines

No, I think this is correct: the code is matching against the last n lines each time text is added to the terminal. If you type "testx", the matching will be done basically on every character you type. Since the previous line already matches `.*testx.*`, the listener will be triggered multiple times as you type.

packages/terminal/src/browser/terminal-widget-impl.ts
jonah-iden: sure, makes sense

Issue #13374, opened 2024-02-10
Webview Inconsistency in acquireVsCodeApi injection behavior.


dineug:
### Bug Description:

Webview Inconsistency in acquireVsCodeApi injection behavior.
Consistency in behavior for compatibility.

```js
// vscode
globalThis.acquireVsCodeApi = (function() {
    // ...
})();
delete window.parent;
delete window.top;
delete window.frameElement;
```

```js
// theia
const acquireVsCodeApi = (function() {
    // ...
})();
const acquireTheiaApi = acquireVsCodeApi;
delete window.parent;
delete window.top;
delete window.frameElement;    
```

### Steps to Reproduce:

```js
// vscode
globalThis.acquireVsCodeApi(); // OK
window.acquireVsCodeApi(); // OK
acquireVsCodeApi(); // OK
```
```js
// theia
globalThis.acquireVsCodeApi(); // NO
window.acquireVsCodeApi(); // NO
acquireVsCodeApi(); // OK
```

### Additional Information

- Operating System:
- Theia Version: 1.46.100
- Vscode Version: 1.86.1

tsmaeder:
@dineug  why would you assume that the function "acquireVSCodeApi()` is available on the `window` object? I have found neither documentation nor an example that implies that. Is there a concrete case where this is a problem?

msujew:
@tsmaeder I would argue that even though it doesn't make a lot of sense, we should align to whatever weird behavior vscode has, just to keep plugin compatibility.

dineug:
@tsmaeder 
It seems like some plugins are being used because they work in VSCode. I also used them, but recently made changes.

https://github.com/excalidraw/excalidraw-vscode/blob/aaeb6c3d7eef780bd150f04793dee38270c3b53a/webview/src/vscode.ts#L15

https://github.com/dineug/erd-editor/commit/864d596429a4e4896516de301c37196885dee841

PR #13411, opened 2024-02-22
Make `acquireVsCodeApi` available on global objects

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/13374

Aligns our implementation of the `acquireVsCodeApi` method to vscode, which is also available on the `globalThis` and `window` objects.

#### How to test

1. Open a webview and hit a breakpoint within
2. Run these scripts in the browsers debugger:
```js
globalThis.acquireVsCodeApi()
window.acquireVsCodeApi()
acquireVsCodeApi()
```
3. All of them should return the vscode API object.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/webview/pre/main.js
+++ b/packages/plugin-ext/src/main/browser/webview/pre/main.js
@@ -146,7 +146,7 @@
      */
     function getDefaultScript(state) {
         return `
-const acquireVsCodeApi = (function() {
+globalThis.acquireVsCodeApi = (function() {
     const originalPostMessage = window.parent.postMessage.bind(window.parent);
     const originalConsole = {...console};
     const targetOrigin = '*';
@@ -198,7 +198,7 @@ const acquireVsCodeApi = (function() {
         });
     };
 })();
-const acquireTheiaApi = acquireVsCodeApi;
+globalThis.acquireTheiaApi = acquireVsCodeApi;
 delete window.parent;
 delete window.top;
 delete window.frameElement;        


packages/plugin-ext/src/main/browser/webview/pre/main.js
```diff
@@ -198,7 +198,7 @@ const acquireVsCodeApi = (function() {
         });
     };
 })();
-const acquireTheiaApi = acquireVsCodeApi;
+globalThis.acquireTheiaApi = acquireVsCodeApi;
```
tsmaeder, line 201: Isn't this a no-op?

msujew: No, see reproduction steps of https://github.com/eclipse-theia/theia/issues/13374

tsmaeder: But we're already assigning to globalThis.actquireVsCodeApi further up in this file and here we're reading from the same variable (implicitly `acquireVSCodeApi === globalThis.acquireVsCodeApi) and assigning it a second time to `globalThis.acquireVsCodeApi`. What am I missing?

msujew: Oh, that's what you mean. One is `acquireVsCodeApi` and the other one is `acquireTheiaApi`.

tsmaeder: If you're relying on me to read text with attention to detail, you're doomed ;-)

tsmaeder (approved):

Issue #13353, opened 2024-02-06
[vscode] workspace registerFileSystemProvider options can explain why FS is readonly with 1.86 


rschnekenbu:
With vscode 1.86, workspace.registerFileSystemProvider() readonly option can now be a markdown string to explain why the file system is readonly. 

see https://github.com/microsoft/vscode/blob/e5447598324c57ac814c150148e7e4c830e0f9ee/src/vscode-dts/vscode.d.ts#L13663

rschnekenbu:
On looking to that issue, I stumbled upon #8167. I do agree with the issue of relative / absolute uri transformation which is done even for a folder using another file system provider. Workspace file in VS Code keeps the URI as it is, where we can only have uris with the 'file:/' scheme in ours.

msujew:
@rschnekenbu You don't need to implement this feature from scratch, see https://github.com/eclipse-theia/theia/pull/13403. It should lay most of the groundwork for the this issue.

rschnekenbu:
Indeed, I was implementing similar parts. I rebased my work on your changes. 
I can then also participate to your PR easily.

Issue #8167, opened 2020-07-14
[vscode] Running example FileSystemProvider plugin has different behavior


Glamhoth:
### Bug Description:
When trying to use example [FileSystemProvider plugin](https://github.com/microsoft/vscode-extension-samples/tree/master/fsprovider-sample) the (expected) behavior on Code[1] differs from the behavior on Theia[2].

[1] Running plugin command `Setup workspace` with or without opened workspace creates new and empty directory with name that was passed to `vscode.workspace.updateWorkspaceFolders` in `extension.ts`.
![image](https://user-images.githubusercontent.com/5027115/87424531-61a67f80-c5dc-11ea-90e4-23ef8aafbf4c.png)

[2] Running plugin command `Setup workspace` without opened workspace throws an error `Failed to update workspace folders: Error: There is not active workspace`. If any folder was already opened in the workspace, the same plugin command opens new directory, but instead of named, empty one it's system's root.
![image](https://user-images.githubusercontent.com/5027115/87424743-c82b9d80-c5dc-11ea-96f1-d1f1a6a6a1cc.png)

### Steps to Reproduce:

1. Clone `microsoft/vscode-extension-samples`
2. Add `fsprovider-sample` plugin to Theia
3. Run Theia
4. Execute `Setup workspace` command from the command paletter

### Additional Information

- Operating System:
Ubuntu 18.04 if running on Docker image
Arch Linux, kernel 5.7.7 if running on host
- Theia Version:
Tried both release `1.3.0` and compiling `akosyakov/integrate-vscode-api-7269` branch.

akosyakov:
Thought it is related to fs APIs, but it is related to workspace APIs. I think it is known bug that we don't respect workspace names yet.

Glamhoth:
How does workspace support for custom scheme/filesystem provider look like? In the example plugin a `memfs` scheme is registered, but even if I try to manually edit the workspace file it's not recognized.

akosyakov:
I don't think it is related to fs providers, it is about:
```
vscode.workspace.updateWorkspaceFolders(0, 0, { uri: vscode.Uri.parse('memfs:/'), name: "MemFS - Sample" });
```

We ignore currently `MemFS - Sample` all together. I've started working on it once, but did not have time to finish: https://github.com/eclipse-theia/theia/pull/7222

Glamhoth:
But this plugin registers provider
[...]

PR #13414, opened 2024-02-22
Better Support for ReadOnly message on editors

rschnekenbu:
#### What it does

This Pull Request continues the work initiated with #13403 and it adds the support of the introduced API in VS Code as described in #13353.  This API addition is the workspace.registerFileSystemProvider() with readonly option being also potentially a Markdown String and not only a boolean. 

Also, it completes the file system API sample with an action to add and remove a read-only message. 
Finally, it improves toggling the readOnly value for a given editor by retrieving the real status of the file, and not only the status of the file system provider. The file can be readonly because of specific permissions for example.

Note: it is currently hard to test the VS Code API as indicated there: https://github.com/eclipse-theia/theia/issues/13353#issuecomment-1946067274. I used the FS provider vscode  sample, but I had same issue. I do not understand why URI has to be transformed to relative and absolute. Removing the lines as indicated in the comment https://github.com/eclipse-theia/theia/issues/8167#issuecomment-842317679 and avoiding some absolute translation works for me, but there are probably reasons why this was done initially.  

#### How to test

1. In a standard workspace, create 2 text files, and make one only writable, the second with only read permissions.
2. Use the various commands from API sample for File system: 
- Add a File System ReadonlyMessage for readonly
- Remove File System ReadonlyMessage for readonly
- Toggle File System Readonly 

3. Check if the lock symbol appears in the title, and also what is displayed in the editor when trying to edit when readonly, for both files. For example, when toggling readonly to false and no readonly message, the file with only read permissions shall still not be editable.

ReadOnlyMessage

The behavior is similar to the one on VS Code: the readOnlyMessage has priority over the capability. 

#### Follow-ups

Check the API thanks to this extension when the workspace update is working with other file systems:
- zip vsix: 
[vscode-memfs-0.0.8.zip](https://github.com/eclipse-theia/theia/files/14377115/vscode-memfs-0.0.8.zip)

- src: 
[vscode-memfs-0.0.8-src.zip](https://github.com/eclipse-theia/theia/files/14377113/vscode-memfs-0.0.8-src.zip)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/browser/file-resource.ts
+++ b/packages/filesystem/src/browser/file-resource.ts
@@ -41,7 +41,7 @@ export namespace FileResourceVersion {
 }
 
 export interface FileResourceOptions {
-    isReadonly: boolean
+    readOnly: boolean | MarkdownString
     shouldOverwrite: () => Promise<boolean>
     shouldOpenAsText: (error: string) => Promise<boolean>
 }
@@ -65,8 +65,8 @@ export class FileResource implements Resource {
     get encoding(): string | undefined {
         return this._version?.encoding;
     }
-    get readOnly(): boolean {
-        return this.options.isReadonly || this.fileService.hasCapability(this.uri, FileSystemProviderCapabilities.Readonly);
+    get readOnly(): boolean | MarkdownString {
+        return this.options.readOnly;
     }
 
     constructor(
@@ -92,11 +92,30 @@ export class FileResource implements Resource {
             console.error(e);
         }
         this.updateSavingContentChanges();
-        this.toDispose.push(this.fileService.onDidChangeFileSystemProviderCapabilities(e => {
+        this.toDispose.push(this.fileService.onDidChangeFileSystemProviderCapabilities(async e => {
             if (e.scheme === this.uri.scheme) {
-                this.updateSavingContentChanges();
+                this.updateReadOnly();
             }
         }));
+        this.fileService.onDidChangeFileSystemProviderReadOnlyMessage(async e => {
+            if (e.scheme === this.uri.scheme) {
+                this.updateReadOnly();
+            }
+        });
+    }
+
+    protected async updateReadOnly(): Promise<void> {
+        const oldReadOnly = this.options.readOnly;
+        const readOnlyMessage = this.fileService.getReadOnlyMessage(this.uri);
+        if (readOnlyMessage) {
+            this.options.readOnly = readOnlyMessage;
+        } else {
+            this.options.readOnly = this.fileService.hasCapability(this.uri, FileSystemProviderCapabilities.Readonly);
+        }
+        if (this.options.readOnly !== oldReadOnly) {
+            this.updateSavingContentChanges();
+            this.onDidChangeReadOnlyEmitter.fire(this.options.readOnly);
+        }
     }
 
     dispose(): void {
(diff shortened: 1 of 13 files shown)


packages/filesystem/src/browser/file-resource.ts
```diff
@@ -92,11 +92,30 @@ export class FileResource implements Resource {
...
+        });
+    }
+
+    protected async updateReadOnly(): Promise<void> {
+        const oldReadOnly = this.options.readOnly;
+        const readOnlyMessage = this.fileService.readOnlyMessage(this.uri);
+        if (readOnlyMessage) {
+            this.options.readOnly = readOnlyMessage;
+        } else {
+            this.options.readOnly = await this.fileService.isReadOnly(this.uri) || this.fileService.hasCapability(this.uri, FileSystemProviderCapabilities.Readonly);
```
msujew, line 113: Suggestion: I'm not sure whether we should actually call the `fileService.isReadOnly()` method here. While this has the benefit of reflecting the current readonly state of the file we're reading (and potentially making it readonly/writable based on the actual file system permissions of the file), it is only updated when the fs capability changes. This feels pretty inconsequential, as the change isn't immediately reflected in the editor - only after we change the capabilities (which is pretty non-trivial from a user perspective). I would remove the `isReadOnly` call here IMO.

Unrelated to that, I have another suggestion that would improve the need to call `isReadOnly` in certain situations:
```suggestion
            this.options.readOnly = this.fileService.hasCapability(this.uri, FileSystemProviderCapabilities.Readonly) || await this.fileService.isReadOnly(this.uri);
```

packages/filesystem/src/browser/file-resource.ts
```diff
@@ -65,8 +65,8 @@ export class FileResource implements Resource {
     get encoding(): string | undefined {
         return this._version?.encoding;
     }
-    get readOnly(): boolean {
-        return this.options.isReadonly || this.fileService.hasCapability(this.uri, FileSystemProviderCapabilities.Readonly);
+    get readOnly(): boolean | MarkdownString {
+        return this.options.readOnly;
```
msujew, line 69: This introduces a regression: Opening a file in a readonly file system marks the file incorrectly as writable, since the file options are decoupled from the capabilities of the file system when creating the initial `FileResource`. See also:

https://github.com/eclipse-theia/theia/blob/31632178a156904eb478259c8bb4dabee597b54d/packages/filesystem/src/browser/file-resource.ts#L350-L354

msujew (commented):
I'm not sure why, but there's a regression: Running the `Toggle File System Readonly` command and opening a new file opens it as writable, whereas it should be opened as readonly.

packages/filesystem/src/common/files.ts
```diff
@@ -765,6 +766,18 @@ export function hasUpdateCapability(provider: FileSystemProvider): provider is F
...
 
+export interface ReadOnlyMessageFileSystemProvider {
+    readOnlyMessage: MarkdownString | undefined;
+    readonly onDidChangeReadOnlyMessage: Event<MarkdownString | undefined>;
+}
+
+export namespace ReadOnlyMessageFileSystemProvider {
+    export function is(arg: unknown): arg is ReadOnlyMessageFileSystemProvider {
+        return isObject<ReadOnlyMessageFileSystemProvider>(arg)
+        && 'readOnlyMessage' in arg;
```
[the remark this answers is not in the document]
rschnekenbu: @msujew, I was unsure for that one. The interface describes _readOnlyMessage_ as being a property, but implementation is a getter / setter couple. Is this still valid to test for it being a function?

packages/filesystem/src/browser/file-resource.ts
rschnekenbu: I updated the way readOnly was setup. This regression should now be removed.

rschnekenbu:
@msujew, can you approve this PR as https://github.com/eclipse-theia/theia/pull/13414#discussion_r1502706998 was resolved with https://github.com/eclipse-theia/theia/pull/13414/commits/76ac7d90ceefc5e7acd3dab59b52f00a78c7488a please?
That will be the last remaining task for VS code compat 1.86.

msujew (approved):
Thanks, looks good to me 👍

PR #13425, opened 2024-02-27
Allow updates of CompositeMenuNode properties (#12948)

### LINKED ISSUE #12948: Options of `CompositeMenuNode` are not actually updated in `updateOptions`
### Bug Description:

In my use case, I want to reuse the context menu to configure an element.
This is done via a toggleable menu item.
To improve usability, I want to propagate the configuration state to its containing parent submenu by adding an iconClass.
I noticed, that the [`updateOptions` function in the `CompositeMenuNode`](https://github.com/eclipse-theia/theia/blob/ba3722b04ff91eb6a4af6a571c9e263c77cdd8b5/packages/core/src/common/menu/composite-menu-node.ts#L64) only updates a node's options via the nullish coalescing operator.
As a result, my submenus are not correctly updated if the configuration changes, as the icon option is never reset, even if the updated option object would set the iconClass explicitly to undefined.

I suggest to actually update the options as indicated in the function name, not only if the option values are nullish, do you agree?

ndoschek:
#### What it does

This PR refines the `updateOptions` method in the `CompositeMenuNode` class.
Previously, the method would only update the properties if they were undefined to begin with (due to the the nullish coalescing operator). With the updated implementation, `updateOptions` now allows properties to be updated.
(Please also see my use case description in the linked issue description)

Allow updates of `CompositeMenuNode` properties
- refactor `updateOptions` function to allow updates
- add unit test cases for `updateOptions`

Fixes #12948

#### How to test

The change is covered by unit tests since there's no existing use case in the example application to demonstrate the updated behavior so far.
These tests confirm the method's functionality across various scenarios, including handling partial updates and dealing with undefined values.

#### Follow-ups

--

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/menu/composite-menu-node.spec.ts
+++ b/packages/core/src/common/menu/composite-menu-node.spec.ts
@@ -1,0 +1,67 @@
+// *****************************************************************************
+// Copyright (C) 2024 EclipseSource and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+import { expect } from 'chai';
+import { CompositeMenuNode } from './composite-menu-node';
+import { CompoundMenuNodeRole } from './menu-types';
+
+describe('composite-menu-node', () => {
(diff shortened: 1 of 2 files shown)


packages/core/src/common/menu/composite-menu-node.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+import * as chai from 'chai';
+import { CompositeMenuNode } from './composite-menu-node';
+import { CompoundMenuNodeRole } from './menu-types';
+
+const expect = chai.expect;
```
jonah-iden, line 20: minor: why not just do `import { expect } from 'chai';` like all the other tests do?

jonah-iden (approved):
Other than the on minor comment this looks god to me 👍

ndoschek:
Thank you for the swift review, @jonah-iden! I've updated the imports in the test file and revised my commit accordingly.

PR #13435, opened 2024-02-29
plugin-ext-vscode: ignore vsix files in local-plugins dir

### LINKED ISSUE #13222: Builtin VSIX plugin is added to the installed list instead of the builtin list
### Bug Description:
Hi,

I have some vscode plugins (in a VSIX file format) in the plugins folder (where all builtin files exist). After upgrading to the latest Theia 1.45 those plugins are part of the installed view instead of the built-in view! They should be treated as builtin as before since they statically exist in the package plugins folder.

### Steps to Reproduce:

1. add any vsix file to the plugins folder
2. run the browser example
3. check and see that this plugin is under the installed list in the extensions view/panel.

### Additional Information

- Operating System: Mac
- Theia Version: Theia 1.45

xai:
#### What it does

In the local-plugins directory (i.e., the one passed to theia via the --plugins argument), we expect to find and deploy unpacked extensions.

This patch fixes an unexpected behavior where theia would also pick up and deploy .vsix files from the local-plugins directory into the deployedPlugins directory, where they will be treated as user-installed extensions on the next start of theia.

Instead, we now ignore .vsix files in the local-plugins directory and print a warning to the console.

Fixes #13222

#### How to test

1. Add any vsix file to the plugins folder (choose one that is not already in ~/.theia/deployedPlugins/)
2. Start Theia
3. Verify that the extension is NOT loaded in Theia (neither as installed nor as builtin)
4. Verify that the extension was NOT unpacked into ~/.theia/deployedPlugins and that it is NOT loaded in Theia.
5. Unpack the extension in the plugins folder
6. Start Theia
7. Verify that the extension is loaded as builtin in Theia
8. Verify that the extension was NOT unpacked into ~/.theia/deployedPlugins

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/node/plugin-deployer-impl.ts
+++ b/packages/plugin-ext/src/main/node/plugin-deployer-impl.ts
@@ -225,8 +225,20 @@ export class PluginDeployerImpl implements PluginDeployer {
     }
 
     protected async resolveAndHandle(id: string, type: PluginType, options?: PluginDeployOptions): Promise<PluginDeployerEntry[]> {
-        const entries = await this.resolvePlugin(id, type, options);
-        await this.applyFileHandlers(entries);
+        let entries = await this.resolvePlugin(id, type, options);
+        if (type === PluginType.User) {
+            await this.applyFileHandlers(entries);
+        } else {
+            const filteredEntries: PluginDeployerEntry[] = [];
+            for (const entry of entries) {
+                if (await entry.isFile()) {
+                    this.logger.warn(`Only user plugins will be handled by file handlers, please unpack the plugin '${entry.id()}' manually.`);
+                } else {
+                    filteredEntries.push(entry);
+                }
+            }
+            entries = filteredEntries;
+        }
         await this.applyDirectoryFileHandlers(entries);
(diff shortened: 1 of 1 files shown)


tsmaeder:
I'm not sure about the approach here: what this PR does is exclude exactly files of type `*.vsix` from exactly the location given via the `--plugins` CLI parameter. I'm not really sure what we consider "built-ins" if I'm looking at the related issue. Looking at the code, I would think that any `PluginDeployerEntry` that is of type `System` should be considered a built-in plugin. Wouldn't it make more sense in that case to simply not apply the file handlers in `PluginDeployerImpl.applyFileHandlers`?

tsmaeder:
@ping?

xai:
Yes, you are right, this would be the better approach instead of a very specific exception. I'll adjust the PR.

xai:
@tsmaeder pong

tsmaeder:
While testing, I found https://github.com/eclipse-theia/theia/issues/13638, but it happens also in master, so unrelated.

tsmaeder (approved):

PR #13483, opened 2024-03-12
Fix overflow behavior of sidebars

### LINKED ISSUE #13482: Overflow behavior for sidepanels not working correctly
### Bug Description:
For certain corner cases the tab overflow behavior for sidebars introduced with #12593 is not working correctly:

- The overflow behavior does not work as expected if only two tabs are in the tabbar (and none are hidden via '...' menu)
- The last visible element never gets added to the `...` menu even if there is not enough space to show it. 
Overflow computation does not consider potential custom styling. Eg. if additional margin between icons are added in an adopter project the overflow calculation will be off.

tortmayr:
#### What it does

- Ensure that last visible element also gets added to `...` overflow menu if there is not enough space
  (Normally this only happens if there is also a top menu configured for the side bar. Otherwise application specific
  min height restrictions typically prevent resizing that far)
- Remove special behavior for only one overflowing element
  -  This behavior would always add a second element to the `overflow` menu if only one element is currently overflowing. This did not work for the case where only two tabs where open in the first place. In addition, it unnecessarily hide a tab in some cases even if there was enough space to render it. This is inline with the behavior of VS Code.
- Also consider potential tab margins in overflow computation

Fixes #13482

#### How to test

__Precondition__:
To fully test this you should have ta at least one top menu in the sidebar. Otherwise application specific minimum size settings might prevent you from reszing the window to a point where all tabs are hidden.

I did this by adding the following code to the `configure` method of the `CommonFrontendContribution`

```ts
app.shell.leftPanelHandler.addTopMenu({
            id: 'settings-menu1',
            iconClass: 'codicon codicon-settings-gear',
            title: nls.localizeByDefault(CommonCommands.MANAGE_CATEGORY),
            menuPath: MANAGE_MENU,
            order: 1,
        });
app.shell.leftPanelHandler.addTopMenu({
            id: 'settings-menu2',
            iconClass: 'codicon codicon-settings-gear',
            title: nls.localizeByDefault(CommonCommands.MANAGE_CATEGORY),
            menuPath: MANAGE_MENU,
            order: 1,
        });
```
The test the overflow behavior by resizing the window.
In particular you should ensure that the previously bugged corner cases are now working as expected
1. Overflow with only two open tabs.
On master the overflow behavior does not work as expected. Instead of showing  the 
`...` menu the view icons just flicker.

2. Ensure that the last not-hidden tab (i.e. the first tab overall) also gets hidden if there is not enough space
3. Ensure that overflow is working correctly if margins are added via custom styling
  You can do this in the dev tools by adding he following rule to the `inspector stylesheet`
  ```css
    .p-TabBar-tab{
      margin-top:15px;
      margin-bottom:15px;
  }
  ```
  If the `inspector-stylesheet` does not show up via ctrl+p you might have to force the creation by navigating to the
  `elements` tabs and applying a new style via the `toolbox` in the styles section:
  
image

After that the stylesheet should be present.

See also this short demo video for reference:

https://github.com/eclipse-theia/theia/assets/2311075/84a68c6b-cc76-4200-a7db-fce40afdf013

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/side-panel-handler.ts
+++ b/packages/core/src/browser/shell/side-panel-handler.ts
@@ -653,7 +653,7 @@ export class SidePanelHandler {
     }
 
     protected onTabsOverflowChanged(sender: SideTabBar, event: { titles: Title<Widget>[], startIndex: number }): void {
-        if (event.startIndex >= 0 && event.startIndex <= sender.currentIndex) {
+        if (event.startIndex > 0 && event.startIndex <= sender.currentIndex) {
             sender.revealTab(sender.currentIndex);
         } else {
             this.additionalViewsMenu.updateAdditionalViews(sender, event);
--- a/packages/core/src/browser/shell/side-panel-handler.ts
+++ b/packages/core/src/browser/shell/side-panel-handler.ts
@@ -211,6 +211,7 @@ export class SidePanelHandler {
     protected createAdditionalViewsWidget(): AdditionalViewsMenuWidget {
         const widget = this.additionalViewsMenuFactory(this.side);
         widget.addClass('theia-sidebar-menu');
+        widget.addClass('theia-additional-views-menu');
         return widget;
     }
 
--- a/packages/core/src/browser/shell/tab-bars.ts
+++ b/packages/core/src/browser/shell/tab-bars.ts
@@ -1212,10 +1212,12 @@ export class SideTabBar extends ScrollableTabBar {
                 let overflowStartIndex = -1;
                 for (let i = 0; i < n; i++) {
                     const hiddenTab = hiddenContent.children[i];
-                    // Extract tab padding from the computed style
+                    // Extract tab padding and margin from the computed style
                     const tabStyle = window.getComputedStyle(hiddenTab);
-                    const paddingTop = parseFloat(tabStyle.paddingTop!);
-                    const paddingBottom = parseFloat(tabStyle.paddingBottom!);
+                    const paddingTop = tabStyle.paddingTop ? parseFloat(tabStyle.paddingTop) : 0;
+                    const paddingBottom = tabStyle.paddingBottom ? parseFloat(tabStyle.paddingBottom) : 0;
+                    const marginTop = tabStyle.marginTop ? parseFloat(tabStyle.marginTop) : 0;
+                    const marginBottom = tabStyle.marginBottom ? parseFloat(tabStyle.marginBottom) : 0;
                     const rd: Partial<SideBarRenderData> = {
                         paddingTop,
                         paddingBottom
@@ -1231,28 +1233,20 @@ export class SideTabBar extends ScrollableTabBar {
                     if (iconElements.length === 1) {
                         const icon = iconElements[0];
                         rd.iconSize = { width: icon.clientWidth, height: icon.clientHeight };
-                        actualWidth += icon.clientHeight + paddingTop + paddingBottom + this.tabRowGap;
+                        actualWidth += icon.clientHeight + paddingTop + paddingBottom + marginTop;
 
-                        if (actualWidth > availableWidth && i !== 0) {
+                        if (actualWidth > availableWidth) {
                             rd.visible = false;
                             if (overflowStartIndex === -1) {
                                 overflowStartIndex = i;
                             }
                         }
+                        // Add the tab row gap to the actual width after the visibility check
+                        actualWidth += marginBottom + this.tabRowGap;
                         renderData[i] = rd;
                     }
                 }
 
-                // Special handling if only one element is overflowing.
-                if (overflowStartIndex === n - 1 && renderData[overflowStartIndex]) {
-                    if (!this.tabsOverflowData) {
-                        overflowStartIndex--;
-                        renderData[overflowStartIndex].visible = false;
-                    } else {
-                        renderData[overflowStartIndex].visible = true;
-                        overflowStartIndex = -1;
-                    }
-                }
                 // Render into the visible node
                 this.renderTabs(this.contentNode, renderData);
                 this.computeOverflowingTabsData(overflowStartIndex);
--- a/packages/core/src/browser/shell/tab-bars.ts
+++ b/packages/core/src/browser/shell/tab-bars.ts
@@ -1084,8 +1084,6 @@ export class SideTabBar extends ScrollableTabBar {
         startIndex: number
     };
 
-    protected _rowGap: number;
-
     constructor(options?: TabBar.IOptions<Widget> & PerfectScrollbar.Options) {
         super(options);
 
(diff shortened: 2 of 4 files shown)


tsmaeder:
@tortmayr I still don't know how to reproduce the problem (on Windows). Is this a Linux-only thing?

tortmayr:
> ```ts
> app.shell.leftPanelHandler.addTopMenu({
>             id: 'settings-menu2',
>             iconClass: 'codicon codicon-settings-gear',
>             title: nls.localizeByDefault(CommonCommands.MANAGE_CATEGORY),
>             menuPath: MANAGE_MENU,
>             order: 1,
>         });
> ```

I have not tested it under Windows, but the issue was originally reported by an adopters that uses Windows so IMO it should be a cross-platform issue.

I have recorded videos with more detailled steps on how to reproduce the bug on master, on how the fixed version based on this PR looks like:
Master:

https://github.com/eclipse-theia/theia/assets/2311075/493215f6-aaed-4e50-9e24-ab41f27bb54c


Fixed:



https://github.com/eclipse-theia/theia/assets/2311075/a7ae44d4-2341-4cea-b432-8a37421db6cb

tsmaeder:
@tortmayr how do you get the two gear icons at the top? That's not happening for me.

tortmayr:
It's explained in the `How to test` section:

To fully test this you should add (at least one) top menu to the sidebar. Otherwise application specific minimum size settings might prevent you from reszing the window to a point where all tabs are hidden.

I did this by adding the following code to the `configure` method of the `CommonFrontendContribution`

```ts
app.shell.leftPanelHandler.addTopMenu({
            id: 'settings-menu1',
            iconClass: 'codicon codicon-settings-gear',
            title: nls.localizeByDefault(CommonCommands.MANAGE_CATEGORY),
            menuPath: MANAGE_MENU,
            order: 1,
        });
app.shell.leftPanelHandler.addTopMenu({
            id: 'settings-menu2',
            iconClass: 'codicon codicon-settings-gear',
            title: nls.localizeByDefault(CommonCommands.MANAGE_CATEGORY),
            menuPath: MANAGE_MENU,
            order: 1,
        });
    
   ```

**Comment by tsmaeder** [comment]
@tortmayr generally, I much prefer written "steps to reproduce" over a video. In a video, it's hard to discern what you mean and to follow the steps. Also, I can't copy/paste text out of a video, so I might make mistakes following the video. You have even written the text, why not just paste it here?

**Comment by tortmayr** [comment]
> @tortmayr generally, I much prefer written "steps to reproduce" over a video. In a video, it's hard to discern what you mean and to follow the steps. Also, I can't copy/paste text out of a video, so I might make mistakes following the video. You have even written the text, why not just paste it here?

Most of this was already written down in the `How to test` section of the PR description.
Nevertheless, I have augmented the PR description and should now cover all steps to reproduce.

>@tortmayr when I do ctrl-p=>inspector-stylesheet, nothing happens in my Theia/Electron. How can I test this?

If the inspector-stylesheet does not show up via ctrl+p you might have to force the creation by navigating to the
elements tabs and applying a new style via the toolbox in the styles section (does not matter to which element):

image

After that the stylesheet should be present.

File: packages/core/src/browser/shell/tab-bars.ts [hunk 1]
```diff
@@ -1231,28 +1233,20 @@ export class SideTabBar extends ScrollableTabBar {
                     if (iconElements.length === 1) {
                         const icon = iconElements[0];
                         rd.iconSize = { width: icon.clientWidth, height: icon.clientHeight };
-                        actualWidth += icon.clientHeight + paddingTop + paddingBottom + this.tabRowGap;
```
tsmaeder, line 1234: Do we not care about the `tabRowGap` anymore? Wouldn't it be relevant?

packages/core/src/browser/shell/tab-bars.ts
```diff
@@ -1212,10 +1212,12 @@ export class SideTabBar extends ScrollableTabBar {
                 let overflowStartIndex = -1;
                 for (let i = 0; i < n; i++) {
                     const hiddenTab = hiddenContent.children[i];
-                    // Extract tab padding from the computed style
+                    // Extract tab padding and margin from the computed style
                     const tabStyle = window.getComputedStyle(hiddenTab);
-                    const paddingTop = parseFloat(tabStyle.paddingTop!);
-                    const paddingBottom = parseFloat(tabStyle.paddingBottom!);
+                    const paddingTop = tabStyle.paddingTop ? parseFloat(tabStyle.paddingTop) : 0;
```
tsmaeder, line 1217: Was this broken before? In what way?

packages/core/src/browser/shell/side-panel-handler.ts
```diff
@@ -653,7 +653,7 @@ export class SidePanelHandler {
     }
 
     protected onTabsOverflowChanged(sender: SideTabBar, event: { titles: Title<Widget>[], startIndex: number }): void {
-        if (event.startIndex >= 0 && event.startIndex <= sender.currentIndex) {
+        if (event.startIndex > 0 && event.startIndex <= sender.currentIndex) {
```
tsmaeder, line 656: This the "first element not added to the menu case, right?"

packages/core/src/browser/shell/tab-bars.ts
```diff
@@ -1231,28 +1233,20 @@ export class SideTabBar extends ScrollableTabBar {
...
 
-                        if (actualWidth > availableWidth && i !== 0) {
+                        if (actualWidth > availableWidth) {
                             rd.visible = false;
                             if (overflowStartIndex === -1) {
                                 overflowStartIndex = i;
                             }
                         }
+                        // Add the tab row gap to the actual width after the visibility check
+                        actualWidth += marginBottom + this.tabRowGap;
```
tsmaeder, line 1245: Why does `paddingBottom` get added before the visibility check, but "marginBottom" afterwards?

tsmaeder:
I found what I believe a new problem: here's what's happening:

1. Open exactly two views in the sidebar; the code is patched to have the two gear icons at the top
2. Reduce the height of the window until the lower view icon starts to disappear
3. Observe: the first (upper) view icon disappears as well _after_ the first one disappears

I'm not sure whether I still move the mouse up after the first one disappears or whether the icon disappears without moving another pixel. It's hard to tell.

tsmaeder:
@tortmayr  here's a video showing the problem:


https://github.com/eclipse-theia/theia/assets/13163770/c9b7f66c-c9ac-40e1-8ff7-c9a6caaf9da6

tsmaeder:
The other problem seem fixed with this PR.

tortmayr:
@tsmaeder Thanks for reviewing, I will look into the bug report.

tortmayr:
@tsmaeder I finally got time to work on this again. Sorry for the long turnaround.

Regarding the problem you encountered:
This is technically not a bug. If only the last tab is overflowing, we hide the tab and render the additional views widget instead.
The additional views widget has  the same size as the previously hidden tab. So in the end we end up with almost the same remaining space as before. For instance lets assume the available height is 300px, the tab height is 48 and the rowgap is 4.
If we now resize to 299px, we hide the last tab and render the additional views menu instead. So this almost nullfies the height gain as we gain 52 px by hiding the tab (height + rowgap) and then lose 48px by rendering the additional menu. So in practice when resizing you have a 4px range in which only the last tab gets hidden, anything higher than that and the second to last tab is hidden as well.

Since this behavior is inline with what VS Code does I opted for not changing it:
https://github.com/eclipse-theia/theia/assets/2311075/dde08670-f990-44a7-acec-750fe502e342

However, I don't have a strong opinion on that. If you think its better to adjust this and e.g. always hide the last two tabs if we enter overflowing mode I'm happy to adjust this.

In general, I reworked the entire hidden tabs calculation and moved it into the `computeOverflowingTabsData` method.
At this point the actual tab bar is already rendered and we can use the actual size/position information to compute hidden tabs.
This remove the need to manually compute available and actual size in `updateTabs` using the information from the hidden tab content node. And we no longer have to fiddle around with extracting margin,borderWidth etc. from the computed styles.
In general this approach is more robust and can handle arbitrary custom styling.

In addition, I encountered a bug with the `AdditionalViewMenuWidget`. Both the widget for the left and right tab bar used the same menu path. As a consequence hidden tabs from one widget leaked into the widget of the other side.
E.g.  In the left side bar ViewA and ViewB are hidden, in the right side bar ViewC is hidden.
Then both menu widgets would offer all views as options (ViewA, ViewB, ViewC).

After my change this works now correctly and the left menu widget only has View A and B and the right only has View C.

JonasHelming:
@tsmaeder Ping

tortmayr:
@tsmaeder Thanks for the review.
I have addressed your comments and we now directly inject the menu path into the `AdditionalViewsMenuWidget`

tsmaeder (approved):

PR #13488, opened 2024-03-13
notebook open optimizations

jonah-iden:
#### What it does

One larger performance issue for notebooks was the creation of the `MonacoEditorModels` used by the cell editors. The problem was that they needed to be created when the notebook is first loaded to be available to the extensions but where created one by one which required firing a network request for each single `MonacoEditorModel`. For a notebook with 100 cells this took up to 500ms.

This PR changes that so that the documents are created seperately between backend and frontend. 
In the backend they are created when the document for the notebook itself is created. In the frontend they are created afterwards in `createNotebookModel` in the `notebook-service.ts` 

Im not sure how nice the solution with rebinding the `MonacoTextModelService` is. Would love some feedback on this

#### How to test

Since its about performance, best way to test is to check if working with notebooks is still working.
open, execute cell, change cell, execute again

#### Review checklist

#### Reminder for reviewers


packages/plugin-ext/src/main/browser/notebooks/notebook-documents-main.ts
```diff
@@ -14,28 +14,33 @@
...
 
-    private readonly notebookModelResolverService: NotebookModelResolverService;
+    protected readonly notebookModelResolverService: NotebookModelResolverService;
+
+    protected notebookMonacoTextModelService: NotebookMonacoTextModelService;
 
     constructor(
         rpc: RPCProtocol,
-        container: interfaces.Container
+        container: interfaces.Container,
```
jbicker, line 43: Minor: The `,` is a left-over, right?

jbicker (approved):
Works fine! Thanks!

jonah-iden:
would love to have a second review on this PR. especially a second opinion on the rebinding of the `monaco-text-model-service`

packages/notebook/src/browser/service/notebook-monaco-text-model-service.ts
```diff
@@ -0,0 +1,45 @@
...
+import { MonacoTextModelService } from '@theia/monaco/lib/browser/monaco-text-model-service';
+import { MonacoEditorModel } from '@theia/monaco/lib/browser/monaco-editor-model';
+import { NotebookModel } from '../view-model/notebook-model';
+
+/**
+ * special service for creating monaco textmodels for notebook cells.
+ * Its for optimization purposes since there is alot of overhead otherwise with calling the backend to create a document for each cell and other smaller things.
+ */
+@injectable()
+export class NotebookMonacoTextModelService extends MonacoTextModelService {
```
tsmaeder, line 28: Can you give a reason why this class extends (and replaces via rebind) the global ITextModelService? The functionality here does not change the behavior of the extended class in any way, as far as I can see.

tsmaeder (dismissed):

tsmaeder:
> would love to have a second review on this PR. especially a second opinion on the rebinding of the `monaco-text-model-service`

You asked for it ;-) 🤷

packages/notebook/src/browser/service/notebook-monaco-text-model-service.ts
tsmaeder: What I mean is this: the "document" as far as the VS Code API is concerned seems to be the main notebook document. Are these "mini-documents" not just local, ephemeral structures that the back-end does not need to know about?

jonah-iden: thanks alot for taking the time. Im not super happy with this solution but don't really know how to make it better.
To eleborate a bit more on this. The jupyter extension seem to require the documents in the backend to read and change cell content as well as to get the code to execute. The source paramter is only changed when saving and reading the document initially. (actually i just found two possible problems there). 
So as described above i need the functionality from the MonacoTextModelService to create the documents on fronted side but without it calling the plugin api to create the documents on backend side because i do that when i create the notebook document in the backend. If i would not do that it would make a call for each cell which takes quite some time. Calling the plugin API is done through a listener being called from the reference collection thats why i have a seperate reference collection for these notebook cell models. So this was the best way i could think of to reuse the functionality. If I would create a second i think there would be some issues with notifying about changes

tsmaeder: Fair enough that you need the cell documents in the back end, but still, this class neither calls nor overrides anything from the base class. Why is it not a separate, unrelated class?

packages/notebook/src/browser/service/notebook-monaco-text-model-service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    protected readonly cellmodels = new ReferenceCollection<string, MonacoEditorModel>(
+        uri => this.loadModel(new URI(uri))
```
tsmaeder, line 31: Ah, I see, this is the call.

jonah-iden: exactly

packages/notebook/src/browser/service/notebook-monaco-text-model-service.ts
jonah-iden: th `loadModel()` call

tsmaeder (commented):
There are two things that bother me about this PR:

1. We are overriding a service in the framework itself. This should not be necessary: we should be able refactor the framework until we have the desired configurability. What if a second module in Theia would do the same? That module and the notebook support could not work in parallel (both trying to impose their implementation).
2. Whether we replicate text documents between front-end and back-end is purely a matter concerning the plugin system. The logic should be confined to the plugin-ext package, IMO. 

What we're trying to achieve, in the end, is to optimize the special case where some documents exist implicitly, so no need to replicate them explicitly. This idea is valid. However, we also have other browser-side documents that also are not replicated to the extension side: for example the editors for breakpoint conditions (not sure how that is achieved, though). But if this is a capability multiple clients of the editor service require, we should build a corresponding extension point into the service instead of overriding it.

jonah-iden:
Thanks alot for the feedback! I'll if i can find a nicer solution with contribution point

jonah-iden:
thinking more about this contribution points are a bit overkill for this aren't they? I just added a `createUnmanagedModel` method to the `monaco-text-model-service`. That way the service can easily be reused

tsmaeder:
>thinking more about this contribution points are a bit overkill for this aren't they? 

A "contribution point" for me is a way to add/change behavior in the framework without replacing the component in the framework. Not a technical term.

PR #13491, opened 2024-03-14
Remove deprecated actions and Playwright API usage

planger:
#### What it does

* Replace usage of deprecated `GabrielBB/xvfb-action@v1`
* Replace usage of deprecated `<obj>.type(...)`

#### How to test

Ensure workflows and Playwright tests still work as before.

#### Follow-ups

We should update https://github.com/eclipse-theia/theia-playwright-template and https://github.com/eclipse-theia/theia-e2e-test-suite accordingly to avoid using the deprecated GH action.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.github/workflows/performance-tests.yml
+++ b/.github/workflows/performance-tests.yml
@@ -20,10 +20,10 @@ jobs:
           node-version: "18.x"
           registry-url: "https://registry.npmjs.org"
 
-      - name: Use Python 3.x
+      - name: Use Python 3.11
         uses: actions/setup-python@v4
         with:
-          python-version: "3.x"
+          python-version: '3.11'
 
       - name: Build
         shell: bash
@@ -36,14 +36,12 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # https://github.com/microsoft/vscode-ripgrep/issues/9
 
       - name: Performance (browser)
-        uses: coactions/setup-xvfb@v1
-        with:
-          run: yarn performance:startup:browser
+        shell: bash
+        run: yarn performance:startup:browser
 
       - name: Performance (Electron)
-        uses: coactions/setup-xvfb@v1
-        with:
-          run: yarn performance:startup:electron
+        shell: bash
+        run: xvfb-run yarn performance:startup:electron
 
       - name: Analyze performance results
         uses: benchmark-action/github-action-benchmark@v1
--- a/.github/workflows/performance-tests.yml
+++ b/.github/workflows/performance-tests.yml
@@ -36,12 +36,12 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # https://github.com/microsoft/vscode-ripgrep/issues/9
 
       - name: Performance (browser)
-        uses: GabrielBB/xvfb-action@v1
+        uses: coactions/setup-xvfb@v1
         with:
           run: yarn performance:startup:browser
 
       - name: Performance (Electron)
-        uses: GabrielBB/xvfb-action@v1
+        uses: coactions/setup-xvfb@v1
         with:
           run: yarn performance:startup:electron
 
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -49,6 +49,5 @@ jobs:
           yarn --cwd examples/playwright build
 
       - name: Test (playwright)
-        uses: GabrielBB/xvfb-action@v1
-        with:
-          run: yarn --cwd examples/playwright ui-tests-ci
+        shell: bash
+        run: yarn --cwd examples/playwright ui-tests-ci
--- a/.github/workflows/production-smoke-test.yml
+++ b/.github/workflows/production-smoke-test.yml
@@ -47,6 +47,5 @@ jobs:
           yarn --cwd examples/playwright build
 
       - name: Run Smoke Test (examples/playwright/src/tests/theia-app)
-        uses: GabrielBB/xvfb-action@v1
-        with:
-          run: yarn test:playwright theia-app
+        shell: bash
+        run: yarn test:playwright theia-app
--- a/examples/playwright/src/theia-quick-command-palette.ts
+++ b/examples/playwright/src/theia-quick-command-palette.ts
@@ -64,12 +64,10 @@ export class TheiaQuickCommandPalette extends TheiaPageObject {
             this.open();
         }
         const input = this.page.locator(`${this.selector} .monaco-inputbox .input`);
-        if (input != null) {
-            await input.focus();
-            await input.pressSequentially(value, { delay: USER_KEY_TYPING_DELAY });
-            if (confirm) {
-                await this.page.keyboard.press('Enter');
-            }
+        await input.focus();
+        await input.pressSequentially(value, { delay: USER_KEY_TYPING_DELAY });
+        if (confirm) {
+            await this.page.keyboard.press('Enter');
         }
     }
 
--- a/examples/playwright/src/theia-quick-command-palette.ts
+++ b/examples/playwright/src/theia-quick-command-palette.ts
@@ -63,10 +63,10 @@ export class TheiaQuickCommandPalette extends TheiaPageObject {
         if (!await this.isOpen()) {
             this.open();
         }
-        const input = await this.page.waitForSelector(`${this.selector} .monaco-inputbox .input`);
+        const input = this.page.locator(`${this.selector} .monaco-inputbox .input`);
         if (input != null) {
             await input.focus();
-            await input.type(value, { delay: USER_KEY_TYPING_DELAY });
+            await input.pressSequentially(value, { delay: USER_KEY_TYPING_DELAY });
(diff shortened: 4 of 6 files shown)


.github/workflows/performance-tests.yml
```diff
@@ -36,14 +36,12 @@ jobs:
...
-        with:
-          run: yarn performance:startup:browser
+        shell: bash
+        run: yarn performance:startup:browser
 
       - name: Performance (Electron)
-        uses: GabrielBB/xvfb-action@v1
-        with:
-          run: yarn performance:startup:electron
+        shell: bash
```
planger, line 43: @xai Do you think omitting xvfb will be a problem for Electron? If yes, we'd need to switch to https://github.com/coactions/setup-xvfb as https://github.com/marketplace/actions/gabrielbb-xvfb-action is deprecated. Thanks!

xai: We need xvfb because we cannot run Electron headless. So yes, we should switch to the maintained action.

planger: Ok, no problem! I updated this PR. Can you please test if this works now? Thank you!

xai: I think we need to test by running the workflow, right?

xai (approved):
Thank you very much, @planger, for spotting the unmaintained action we are consuming and for fixing our API calls!

Tested fine locally, i.e., the electron tests are still flaky, but no more flaky than before.

I agree that in the Electron case, we need to replace the unmaintained [xvbf-action](https://github.com/GabrielBB/xvfb-action) which had its last real commit in 2021-11.

The [replacement](https://github.com/coactions/setup-xvfb) used in this PR matches the recommendation in the discontinued xvbf-action and it seems to be actively maintained. It also seems to be used frequently (1/10 of the usage that the discontinued action has, which sounds reasonable to me). In the end, by executing the action, we end up executing this file: https://github.com/coactions/setup-xvfb/blob/main/dist/index.js

I very briefly scanned what it does and I have to mention that it includes odd-looking stuff such as [this](https://github.com/coactions/setup-xvfb/blob/6b00cf1889f4e1d5a48635647013c0508128ee1a/dist/index.js#L3454), which is explained with performance optimization. I did not spot any obvious red flags or outright harmful code, but keep in mind that this is a 4K lines javascript file and I looked at it for 3 minutes, which is not much, especially when considering that I am trying to limit my coffee input this week (read: I feel dumb and slow).

I assume that we need to establish a process for vetting third-party actions anyway at some point, and this workflow currently does not even seem to be executed by schedule or dispatch, so I will approve.

marcdumais-work:
> The [replacement](https://github.com/coactions/setup-xvfb) used in this PR matches the recommendation in the discontinued xvbf-action and it seems to be actively maintained. It also seems to be used frequently (1/10 of the usage that the discontinued action has, which sounds reasonable to me). In the end, by executing the action, we end up executing this file: https://github.com/coactions/setup-xvfb/blob/main/dist/index.js

Good catch @xai - indeed it's hard to know for sure that this code is not doing anything we would not want it to. I looked at their issue tracker and noticed [this issue](https://github.com/coactions/setup-xvfb/issues/21), that claims that `xvfb` is now part of the Ubuntu 22.04 GitHub runner image, and that in consequence using the action is not longer needed, for that OS. It seems that we only use the action on Ubuntu, so maybe we can forego using that action and still be able to run Electron tests?

.github/workflows/performance-tests.yml
```diff
@@ -36,12 +36,12 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # https://github.com/microsoft/vscode-ripgrep/issues/9
 
       - name: Performance (browser)
-        uses: GabrielBB/xvfb-action@v1
+        uses: coactions/setup-xvfb@v1
```
marcdumais-work, line 39: for the browser, `xvfb` should not be needed? At least when I run this test suite locally, I see no window being created.

.github/workflows/performance-tests.yml
```diff
@@ -36,12 +36,12 @@ jobs:
...
         with:
           run: yarn performance:startup:browser
 
       - name: Performance (Electron)
-        uses: GabrielBB/xvfb-action@v1
+        uses: coactions/setup-xvfb@v1
```
marcdumais-work, line 44: I suggest we give it a try without this action - as I commented on the PR, it would seems `ubuntu-latest` comes with `xvfb` pre-installed.

.github/workflows/performance-tests.yml
```diff
@@ -36,12 +36,12 @@ jobs:
...
         with:
           run: yarn performance:startup:electron
```
marcdumais-work, line 46: ATM this test suite does not seem to work - however it's the same on master, so it's not a problem caused by this PR.

I added trace "analyseTrace(): profilePath ..." - it seems that the trace files, that we attempt to parse using `JSON.parse()`, are binary files not json...

```
$ yarn -s electron rebuild && cd scripts/performance && node electron-performance.js --name 'Electron Frontend Startup' --folder electron --runs 10
native node modules are already rebuilt for electron
analyseTrace(): profilePath: /home/user/theia/scripts/performance/profiles/electron/1.json
[Performance][Electron Frontend Startup][1] Largest Contentful Paint (LCP) failed to obtain a measurement: Unexpected token o in JSON at position 1
Failed to obtain a measurement. The most likely cause is that the performance trace file was incomplete because the script did not wait long enough for "Largest Contentful Paint (LCP)".
SyntaxError: Unexpected token o in JSON at position 1
    at JSON.parse (<anonymous>)
    at analyzeTrace (/home/user/theia/scripts/performance/common-performance.js:151:26)
    at measure (/home/user/theia/scripts/performance/common-performance.js:86:26)
analyseTrace(): profilePath: /home/user/theia/scripts/performance/profiles/electron/2.json
[Performance][Electron Frontend Startup][2] Largest Contentful Paint (LCP) failed to obtain a measurement: Unexpected token o in JSON at position 1
Failed to obtain a measurement. The most likely cause is that the performance trace file was incomplete because the script did not wait long enough for "Largest Contentful Paint (LCP)".
SyntaxError: Unexpected token o in JSON at position 1
    at JSON.parse (<anonymous>)
    at analyzeTrace (/home/user/theia/scripts/performance/common-performance.js:151:26)
    at measure (/home/user/theia/scripts/performance/common-performance.js:86:26)

[...]
```

.github/workflows/performance-tests.yml
marcdumais-work: I think these are the same errors today in repo theia-playwright-template, that went away after pinning Python to 3.11:
https://github.com/eclipse-theia/theia/actions/runs/8286508515/job/22676662833

.github/workflows/performance-tests.yml
xai: I agree, for browser we do not need xvfb.

xai:
> I looked at their issue tracker and noticed [this issue](https://github.com/coactions/setup-xvfb/issues/21), that claims that `xvfb` is now part of the Ubuntu 22.04 GitHub runner image, and that in consequence using the action is not longer needed, for that OS. It seems that we only use the action on Ubuntu, so maybe we can forego using that action and still be able to run Electron tests?

Wow, nice! Thanks for checking this! This is really great news! If we have the xvfb-run binary in the image, we can save ourselves a lot of trouble.

.github/workflows/performance-tests.yml
planger: Thank you very much for looking into this!
I get the same error (in this PR and on master). I think this workflow file has not been used for a long time, also I don't see any runs of this workflow on GH. I just updated it to avoid leaving the deprecated action there.

.github/workflows/performance-tests.yml
planger: Right, I think so too. I just kept it in as it was there before. But you're right, let's remove it.

.github/workflows/performance-tests.yml
planger: I updated the PR to use `xvfb-run` directly. Thank you for pointing this out!

planger:
Thank you @xai and @marcdumais-work for your thorough review. I've updated the `performance-tests.yml` accordingly. However, as you also say, it doesn't work anyway (also on master) and I just updated it to get rid of the deprecated action and close the security risk, once someone picks up the work on this workflow again.

marcdumais-work:
>  I just updated it to get rid of the deprecated action and close the security risk, once someone picks up the work on this workflow again.

I think that's fair. I was curious to confirm whether the image-built-on `xvfb` would work - I think we can trigger a manual run once the PR is merged, to confirm that it's at least present.

marcdumais-work (approved):
LGTM - thanks @planger !

marcdumais-work:
>  I think we can trigger a manual run once the PR is merged, to confirm that it's at least present.

Confirmed, I think: we get the same parsing error I noticed locally, that we briefly discussed : 

https://github.com/eclipse-theia/theia/actions/runs/8296738478/job/22706349277#step:7:20

Issue #13559, opened 2024-04-03
Wrong Window Revealed when Navigating in Editor


tsmaeder:
### Bug Description:
When navigating to a symbol in the editor (`ctrl-click`) with multiple Theia windows open, the wrong window is brought to the front.

### Steps to Reproduce:

1. Open a workspace with some typescript code in Theia Electron
2. Open a second workspace from the File->Open Folder menu
3. In the first window, `ctrl-click` on the usage of a symbol to navigate to the declaration of the symbol
4. Observe: the second window is brought to the front.

### Additional Information

- Operating System: Windows
- Theia Version: 1.48.x master

PR #13561, opened 2024-04-03
Fix window revealing when navigating with multiple windows

tsmaeder:
#### What it does

Makes sure the correct window is brought to the top when navigating between editors.

Fixes: #13559

#### How to test
1. Follow the scenario from the linked issue and make sure the correct windows is revealed when navigating in both windows
2. In each Theia window from scenario No. 1, open a file in a secondary window and make sure we can navigate from both the secondary window and the main window and vice-versa.

#### Review checklist

#### Reminder for reviewers


planger (approved):
Thanks, looks good to me! :+1: Confirmed that https://github.com/eclipse-theia/theia/pull/13561 applies in Ubuntu Linux and that this PR fixes the bug.

PR #13577, opened 2024-04-09
Improve notebook markdown cell rendering

msujew:
#### What it does

Improves the markdown rendering logic of markdown cells by rendering the HTML element directly into the markdown cell without using `innerHTML` manipulation.

This allows to keep the mouse based events that the markdown renderer attaches to the HTML elements. Elements such as links are broken when attempting to use `innerHTML` manipulation.

#### How to test

1. Create a new markdown cell and set the content:
```md
## Heading 2

### Heading 3

This is a paragraph of text.

- Item 1
- Item 2
- Item 3

**Bold text**

*Italic text*

[Link to Google](https://www.google.com)
```
2. Click on the link and ensure that it opens in a new browser tab.
3. Create a new empty markdown cell and finish editing. Double clicking the cell should enter edit mode.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/view/notebook-markdown-cell-view.tsx
+++ b/packages/notebook/src/browser/view/notebook-markdown-cell-view.tsx
@@ -61,18 +61,23 @@ function MarkdownCell({ markdownRenderer, monacoServices, cell, notebookModel, n
         return () => listener.dispose();
     }, [editMode]);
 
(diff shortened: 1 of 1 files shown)


jonah-iden (approved):
nice solotion. Works great! Also seems to be more performant according to some articles. Maybe we should do the same with the cell editor

PR #13581, opened 2024-04-09
Hide empty plugin view containers from user

msujew:
#### What it does

For some reason, the jupyter notebook extension registers an empty view container. See [here](https://github.com/microsoft/vscode-jupyter/blob/7d8d3274703ce1ceb46a04c9feb106e3d791a70e/package.json#L2014-L2020). Theia shows the empty container in the `Views` menu and even allows the user to open it (even though it is completely empty):

image

However, this doesn't matter in vscode, as empty view containers (i.e. view containers which never have a view registered to them) are just not visible to the user in any way.

This change aligns our UX to vscode and only registers the appropriate commands/menu entries once a view has been added to the view container.

#### How to test

1. Install the Jupyter plugin
2. Only one `Jupyter` entry should be visible in the `Views` menu
3. Only one `View: Toggle Jupyter` command should be visible
4. All other plugin view containers should work as expected

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/playwright/src/tests/theia-quick-command.test.ts
+++ b/examples/playwright/src/tests/theia-quick-command.test.ts
@@ -61,8 +61,8 @@ test.describe('Theia Quick Command', () => {
     });
 
     test('should trigger \'Toggle Explorer View\' command after typing', async () => {
-        await quickCommand.type('Toggle Explorer');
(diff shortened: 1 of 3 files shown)


jonah-iden (approved):
works as described and looks good to me

Issue #13140, opened 2023-12-02
Remote SSH not working


jlahoda:
I wanted to try the remote ssh in Theia, as introduced by this:
https://github.com/eclipse-theia/theia/pull/12618

But, it does not see to work (for me). I am on a Linux machine, trying to connect to the same machine. I tried the same (user) account and a different account, did not notice any difference.

What I did:
- the following checked-out commit 764d9baf205eab01b1262db2c5821816f446aeeb for me:
- `$ git clone https://github.com/eclipse-theia/theia-blueprint`
- `$ cd theia-blueprint`
- //since the blueprint does not seem to include the remote plugin, I added it:
- apply patch:
```
diff --git a/applications/electron/package.json b/applications/electron/package.json
index 8e19bfe..9f8b6f2 100644
--- a/applications/electron/package.json
+++ b/applications/electron/package.json
@@ -74,6 +74,7 @@
     "@theia/preferences": "1.43.0",
     "@theia/process": "1.43.0",
     "@theia/property-view": "1.43.0",
+    "@theia/remote": "1.43.0",
     "@theia/secondary-window": "1.43.0",
     "@theia/scm": "1.43.0",
     "@theia/search-in-workspace": "1.43.0",
```
- `$ yarn && yarn build && yarn download:plugins && yarn electron package`
- `$ ./applications/electron/dist/TheiaBlueprint.AppImage`
- clicked in the left bottom corner, selected "Connet to Host", typed `ide@localhost`. I have a key without a passphrase (specifically for this experiment), that can login to the account without password.
- Theia connects, and starts to download of Node.js and the backend.
- it then fails with:
```
Could not open SSH connection to remote. Could not start remote system: node:internal/modules/cjs/loader:1080 throw err; ^ Error: Cannot find module '/home/ide/.blueprint-electron-app-1.43.0-remote/lib/backend/main.js' [90m at Module._resolveFilename (node:internal/modules/cjs/loader:1077:15)[39m [90m at Module._load (node:internal/modules/cjs/loader:922:27)[39m [90m at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)[39m [90m at node:internal/main/run_main_module:23:47[39m { code: [32m'MODULE_NOT_FOUND'[39m, requireStack: [] } Node.js v18.17.0
```

If I look at the content of .blueprint-electron-app-1.43.0-remote/lib/backend:
```
[...]

PR #13584, opened 2024-04-10
Fix remote support in packaged apps

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/13140

Currently. the `ApplicationPackage` service is bound this way:

```ts
bind(ApplicationPackage).toDynamicValue(({ container }) => {
  const { projectPath } = container.get(BackendApplicationCliContribution);
  return new ApplicationPackage({ projectPath });
}).inSingletonScope();
```

However, the `projectPath` remains undefined, until the CLI app lifecycle has completed. This means that the `ApplicationPackage` may already be injected into other services with its `projectPath` undefined. This seems to be the case in https://github.com/eclipse-theia/theia-blueprint/pull/350 and in https://github.com/eclipse-theia/theia/issues/13140.

This change changes the way the `projectPath` is set. Instead of waiting until the CLI lifecycle completes, we set the `BackendApplicationPath` as soon as possible (at js load time). It didn't really make a lot of sense for the `projectPath` to be a CLI setting anyway, given that we use it to find our own code.

#### How to test

1. Use the SSH remote feature and assert that it works as expected
2. Open a webview from a plugin and assert that it loads and works as expected

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-manager/src/generator/backend-generator.ts
+++ b/dev-packages/application-manager/src/generator/backend-generator.ts
@@ -52,10 +52,12 @@ if (process.env.LC_ALL) {
 }
 process.env.LC_NUMERIC = 'C';
 
+const { resolve } = require('path');
+const theiaAppProjectPath = resolve(__dirname, '..', '..');
+process.env.THEIA_APP_PROJECT_PATH = theiaAppProjectPath;
 const { default: electronMainApplicationModule } = require('@theia/core/lib/electron-main/electron-main-application-module');
 const { ElectronMainApplication, ElectronMainApplicationGlobals } = require('@theia/core/lib/electron-main/electron-main-application');
 const { Container } = require('inversify');
-const { resolve } = require('path');
 const { app } = require('electron');
 
 const config = ${this.prettyStringify(this.pck.props.frontend.config)};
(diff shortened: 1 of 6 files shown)


msujew:
cc @jfaltermeier This should fix https://github.com/eclipse-theia/theia-blueprint/pull/350. Can you test it?

jfaltermeier:
@msujew Yes I can test it tomorrow.

jfaltermeier:
Hi, so I've built a local next version based on your branch and consumed it in Blueprint. 
I then ran into an error message "Server error while downloading native dependency from", because there is no github release matching my next version, from where it would get the native parts. 
I then more or less hardcoded this URL to the last Github Release and built a new local next version. 

With this I ran into the same issue as before (https://github.com/eclipse-theia/theia-blueprint/pull/350#issuecomment-2034718037)

But I don't know how the remote package works exactly, so maybe with above "fix"/hack it is not expected to work. 
Do I have to rebuild those native parts as well? Do the versions have to match for some copying to work (e.g. I can see that the `.theia-ide-electron-app-1.48.300-remote` directory exists in the container including the native parts. This has the version number of the last release and not my local next version)?

msujew:
@jfaltermeier Thanks for testing. Looking closer at the error, I believe that I misunderstood the source of the issue. Sorry for that :/

I'll try to debug this myself later to see what goes wrong.

msujew:
@jfaltermeier Alright, found the issue. When the application is packaged, the `path.relative` call returned bogus data for some reason. In fact, the mapping using `path.relative` wasn't necessary at all, since `glob` already returns all file paths in relative form. See latest commit on this PR. I've applied the same fix to the theia-blueprint repo and was successfully able to run the dev-container and remote SSH feature.

Note that it still required the other change for me, since otherwise the `projectPath` wasn't correctly computed.

jfaltermeier (approved):
Looks good to me, thank you!
I was also able to get a dev-container running in Theia Blueprint as well:
ksnip_20240417-142232

PR #13585, opened 2024-04-10
Use notebook URI as context for toolbar commands

msujew:
#### What it does

Fixes the `Restart Kernel` command of the jupyter notebook extension. It expects a [URI as its argument](https://github.com/microsoft/vscode-jupyter/blob/efe0f5c797c0d0aa0108388a9705dd2f09bf6805/src/notebooks/notebookCommandListener.ts#L72). We used to pass in the whole notebook model.

#### How to test

1. Open a Jupyter notebook file
2. Run the `Restart Kernel` command from the main toolbar
3. The kernel should restart (see notification) and then work as expected

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/contributions/notebook-actions-contribution.ts
+++ b/packages/notebook/src/browser/contributions/notebook-actions-contribution.ts
@@ -14,7 +14,7 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { Command, CommandContribution, CommandHandler, CommandRegistry, CompoundMenuNodeRole, MenuContribution, MenuModelRegistry, nls } from '@theia/core';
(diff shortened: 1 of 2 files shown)


jonah-iden:
This is actually what the ArgumentProcessor in notebooks.ts was for. Wouldn't that be a nicer solution than or does that not work because it needs the uri and not the document

msujew:
> This is actually what the ArgumentProcessor in notebooks.ts/notebooks-main.ts was for. Wouldn't that be a nicer solution than or does that not work because it needs the uri and not the document

@jonah-iden Well, the ArgumentProcessor correctly converted the `NotebookModel` into a `theia.NotebookDocument` on the plugin host. However, this isn't the type that's expected by commands registered to the `interactive/toolbar` menu. The command [here](https://github.com/microsoft/vscode-jupyter/blob/efe0f5c797c0d0aa0108388a9705dd2f09bf6805/src/notebooks/notebookCommandListener.ts#L72) expects a `vscode.Uri`, not `vscode.NotebookDocument`.

jonah-iden (approved):
ok then its as i thought. Approved

PR #13590, opened 2024-04-11
Generate Extension Info in server application

### LINKED ISSUE #13481: List of Theia extensions is missing in AboutDialog in packaged Theia Electron based applications
### Bug Description:

In packaged Theia Electron based applications the default about dialog does not show the list of contained Theia extensions.

*Note that the [Theia IDE](https://theia-ide.org/#theiaidedownload) uses a customized about dialog and therefore the problem is not visually apparent.*

### Steps to Reproduce:

1. Build the Theia Browser example app
2. Delete the `node_modules` directory (it's not required as the backend is bundled by default)
3. Run the example app (`node examples/browser/lib/backend/main.js`)
4. Open the about dialog

image

You can also reproduce this with the Theia Electron example app, however you need to keep all Electron related dependencies in `node_modules` as they are not bundled with Theia.

The issue typically occurs in packaged Theia Electron based applications as publishers no longer package `node_modules` since the backend bundling was introduced.

In development environments and in Docker images the `node_modules` are usually present, thereby the error does not manifest then. However in theory the `node_modules` could also not be present, as shown in the reproduction steps.

### Analysis

* [The dialog requests the information from the `AppServer`](https://github.com/eclipse-theia/theia/blob/9ff0cedff1d591b0eb4be97a05f6d992789d0a24/packages/core/src/browser/about-dialog.tsx#L64)
* [The `AppServer` receives the information from the injected `ApplicationPackage`](https://github.com/eclipse-theia/theia/blob/9ff0cedff1d591b0eb4be97a05f6d992789d0a24/packages/core/src/node/application-server.ts#L29)
* [The `ApplicationPackage` is created by the `backend-application-module`](https://github.com/eclipse-theia/theia/blob/9ff0cedff1d591b0eb4be97a05f6d992789d0a24/packages/core/src/node/backend-application-module.ts#L104-L108)
* [The `ApplicationPackage` uses the `ExtensionPackageCollector`](https://github.com/eclipse-theia/theia/blob/9ff0cedff1d591b0eb4be97a05f6d992789d0a24/dev-packages/application-package/src/application-package.ts#L106-L116)
* [The `ExtensionPackageCollector` walks through the fileystem and tries to collect `package.json` files](https://github.com/eclipse-theia/theia/blob/9ff0cedff1d591b0eb4be97a05f6d992789d0a24/dev-packages/application-package/src/extension-package-collector.ts)

As the extension information comes through the collector, it fails if the packages are not available, i.e. in the case the backend is bundled and the `node_modules` removed. Notably the `ApplicationPackage` as well as the `ExtensionPackageCollector` are actually coming from the `dev-packages`, so it can be argued they should not be used at all in Theia itself.

### Potential solution

The Theia extension information is already collected during the build process. So instead of re-discovering this information with the tools of `dev-packages` at runtime, the information could be included in src-gen instead.

rschnekenbu:
#### What it does

Generate information about extension in server application rather than retrieving them in the about dialog at runtime. This was leading to missing information when an electron app was packaged. 

Fixes #13481

#### How to test

- Build the Theia Browser example app
- Delete the node_modules directory (it's not required as the backend is bundled by default)
- Run the example app (`node examples/browser/lib/backend/main.js`)
- Open the about dialog
- There should be the list of installed extensions.

The generation for Theia electron example app should also work. It is harder to test (from Stefan's comment on issue:  _you need to keep all Electron related dependencies in node_modules as they are not bundled with Theia._)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-manager/src/generator/backend-generator.ts
+++ b/dev-packages/application-manager/src/generator/backend-generator.ts
@@ -185,6 +185,8 @@ const main = require('@theia/core/lib/node/main');
 
 BackendApplicationConfigProvider.set(${this.prettyStringify(this.pck.props.backend.config)});
 
+globalThis.extensionInfo = ${this.prettyStringify(this.pck.extensionPackages.map(({ name, version }) => ({ name, version }))) };
+
 const serverModule = require('./server');
 const serverAddress = main.start(serverModule());
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,11 @@
 
 ## not yet released
 
+- [application-manager] Generate Extension Info in server application to avoid empty About Dialog [#13590](https://github.com/eclipse-theia/theia/pull/13590) - contributed on behalf of STMicroelectronics
 - [scm] added support for dirty diff peek view [#13104](https://github.com/eclipse-theia/theia/pull/13104)
 
 <a name="breaking_changes_not_yet_released">[Breaking Changes:](#breaking_changes_not_yet_released)</a>
+
 - [scm] revised some of the dirty diff related types [#13104](https://github.com/eclipse-theia/theia/pull/13104)
   - replaced `DirtyDiff.added/removed/modified` with `changes`, which provides more detailed information about the changes
   - changed the semantics of `LineRange` to represent a range that spans up to but not including the `end` line (previously, it included the `end` line)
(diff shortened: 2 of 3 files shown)


dev-packages/application-manager/src/generator/backend-generator.ts
```diff
@@ -192,6 +192,8 @@ serverAddress.then((addressInfo) => {
 });
 
 globalThis.serverAddress = serverAddress;
+globalThis.extensionInfo = ${this.prettyStringify(this.pck.extensionPackages.map(({ name, version }) => ({ name, version }))) };
```
msujew, line 195: Suggestion: Since the information is statically available, (unlike the `serverAddress`), we should set the `extensionInfo` value before importing from `./server`, as adopters might want to access this information during load time.

msujew (approved):
Looks pretty good, just a minor suggestion, see below.

Issue #13535, opened 2024-03-27
vscode.env.openExternal opens a given file in Theia rather than in an external app


chroberino:
### Bug Description:
I guess `vscode.env.openExternal` in Theia should do the same as in VS Code, i.e. use the OS' default application for opening a given URI. I was trying to use this call in a VS Code extension and observed that in Theia files are always be opened within Theia (while in VS Code they open in external apps).

### Steps to Reproduce:
1. Create a VS Code extension:
   - In `package.json` of your VS Code extension create a command like
            ```{
                "command": "myextension.open-in-external-app",
                "title": "Open in External App"
            }```
   - Implement this command by invoking `vscode.env.openExternal`, aiming to open the currently chosen file in its default app (as configured in the OS).

2. When using your VS Code extension in Theia select a file in the Explorer and pick the "Open in External App" command from the command palette
3. Unexpectedly, the file will be opened within Theia, not in an external app.

As an easy alternative (without any coding efforts) you could also reproduce this issue with the [Open in External App](https://open-vsx.org/extension/YuTengjing/open-in-external-app) extension: If you don't give it any specific configuration, it will utilize `vscode.env.openExternal` striving to open a given file URI in its default application. However, the result is the same as described above, the file will not be opened in an external app but within Theia.

<!-- Please provide any additional information available. -->
<!-- Additional information can be in the form of logs, screenshots, screencasts. -->

### Additional Information

- Operating System: Windows 10
- Theia Version:
  - I have tried with the follwing two:
    - Download of ready-made Eclipse Theia IDE 1.47.100 (https://theia-ide.org/#theiaidedownload)
    - Self-built Electron app from Eclipse Theia 1.45.0 sources

dhuebner:
@chroberino 
Are you also missing the "Open in External App" file browser menu contribution after installing  Open in External App extension? 
I see the command in the palette, but not the menu entry...

dhuebner:
Right now `theia.env.openExternal()` is called when the  `vscode.env.openExternal`  is triggered.  This calls `windowStateExt.openUri(uri)` and its current implementation just redirects to the `opener-service.ts` without providing any further information except the URI. The only handler that matches is the EditorPreview so the file is opened in Theia. I also didn't found a registered handler that can call external applications. In electron case such a handler should probably call `shell.openPath(fsPath)` to redirect to the OS editors.

dhuebner:
Maybe we could also offer `Open With... -> System Editor` in Theia?

[...]

PR #13601, opened 2024-04-12
Add a new built-in handler to open files with system application

dhuebner:
#### What it does

Adds a new Command and an OpenWithService handler to open files using system application.

Bildschirmfoto 2024-04-12 um 15 06 50

#### How to test

-  Select a file in navigator. Call "Open With System Editor" command from the command palette. 
- ...or, from the Navigator menu. Select "Open With...".  Select the new "System Editor" entry and press ENTER. 

In both cases the file should be opened with your system default editor. 

#### Follow-ups

This new command can be called to handle `vscode.env.openExternal` request, see #13535 .

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/navigator/src/electron-browser/electron-navigator-menu-contribution.ts
+++ b/packages/navigator/src/electron-browser/electron-navigator-menu-contribution.ts
@@ -14,17 +14,19 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { Command, CommandContribution, CommandRegistry, MenuContribution, MenuModelRegistry, SelectionService } from '@theia/core';
-import { CommonCommands, KeybindingContribution, KeybindingRegistry } from '@theia/core/lib/browser';
+import { Command, CommandContribution, CommandRegistry, MenuContribution, MenuModelRegistry, SelectionService, URI } from '@theia/core';
+import { CommonCommands, KeybindingContribution, KeybindingRegistry, OpenWithService } from '@theia/core/lib/browser';
 import { WidgetManager } from '@theia/core/lib/browser/widget-manager';
-import { inject, injectable } from '@theia/core/shared/inversify';
-import { FileStatNode } from '@theia/filesystem/lib/browser';
-import { FileNavigatorWidget, FILE_NAVIGATOR_ID } from '../browser';
-import { NavigatorContextMenu, SHELL_TABBAR_CONTEXT_REVEAL } from '../browser/navigator-contribution';
-import { isWindows, isOSX } from '@theia/core/lib/common/os';
-import { WorkspaceService } from '@theia/workspace/lib/browser';
+import { nls } from '@theia/core/lib/common';
+import { FileUri } from '@theia/core/lib/common/file-uri';
+import { isOSX, isWindows } from '@theia/core/lib/common/os';
 import { UriAwareCommandHandler } from '@theia/core/lib/common/uri-command-handler';
 import '@theia/core/lib/electron-common/electron-api';
+import { inject, injectable } from '@theia/core/shared/inversify';
+import { FileStatNode } from '@theia/filesystem/lib/browser';
+import { WorkspaceService } from '@theia/workspace/lib/browser';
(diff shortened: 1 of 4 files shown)


packages/navigator/src/electron-browser/electron-navigator-menu-contribution.ts
```diff
@@ -54,6 +64,28 @@ export class ElectronNavigatorMenuContribution implements MenuContribution, Comm
...
+            canHandle: uri => (uri.scheme === 'file') ? 10 : 0,
+            open: uri => {
+                this.openWithSystemApplication(uri);
+                return {};
+            }
+        });
+    }
+
+    protected openWithSystemApplication(uri: URI): void {
+        window.electronTheiaCore.openWithSystemApp(uri['codeUri'].fsPath);
```
tsmaeder, line 88: Use `FileUri.fsPath` instead, please. It handles special cases.

tsmaeder: It does not make any difference in this case (electron will be the same OS between browser and back-end), but in general, we should not do any OS-specific file manipulation on the browser side.

packages/navigator/src/electron-browser/electron-navigator-menu-contribution.ts
```diff
@@ -54,6 +64,28 @@ export class ElectronNavigatorMenuContribution implements MenuContribution, Comm
             isEnabled: uri => !!this.workspaceService.getWorkspaceRootUri(uri),
             isVisible: uri => !!this.workspaceService.getWorkspaceRootUri(uri),
         }));
+        commands.registerCommand(OPEN_WITH_SYSTEM_APP, UriAwareCommandHandler.MonoSelect(this.selectionService, {
+            execute: async uri => {
+                this.openWithSystemApplication(uri);
+            },
+            isEnabled: uri => !!this.workspaceService.getWorkspaceRootUri(uri),
```
tsmaeder, line 71: Why do we care whether the file is in the workspace? As long as it's a file uri, why not try to open it?

packages/navigator/src/electron-browser/electron-navigator-menu-contribution.ts
msujew: I would recommend `Uri#path#fsPath()` instead of `FileUri.fsPath`, since it deals better with the case where backend and frontend are running on different OS.

tsmaeder: We have ~100 usages of `FileUri.fsPath()`. If it's not the right thing, we have a problem. Should FileUri.fsPath forward to `Uri`?

msujew: `FileUri.fsPath()` used to be exclusively used on the backend (I believe you moved it from `node` to `common` recently), where only available OS was the backend OS. Changing it to forward to `Uri#path#fsPath` would probably be a good idea though.

packages/navigator/src/electron-browser/electron-navigator-menu-contribution.ts
dhuebner: I thought, because right now it is only possible to call "Open with..." from the navigator it would make sense to restrict the command to WS files. But yes, you are right, why not to allow for all URIs? I will change it.

packages/navigator/src/electron-browser/electron-navigator-menu-contribution.ts
dhuebner: Good, I will use `FileUri.fsPath()` and in other PR somebody need to redirect from `FileUri.fsPath()` to `Uri#path#fsPath`. Do we have an issue for that already?

tsmaeder: Nope

dhuebner: Okay, I will create one.
Otherwise, all suggestions are applied, please have a look.

dhuebner: Opened #13620

tsmaeder (approved):

PR #13606, opened 2024-04-15
Fix notebook model/cell disposal

msujew:
#### What it does

Previously, when deleting a cell or a whole notebook, the deletion wasn't propagated to the documents-ext handler. Meaning that there was still a virtual document available in the plugin host for non-existing cell. This later on led to caching issues, for example accidentally executing outdated code cells.

Furthermore, closed notebooks didn't dispose their underlying model when the editor was closed. The model was always available in memory.

This change fixes the notebook document removal event and also simply deletes all cell documents from the plugin host.

#### How to test

1. Set a breakpoint in [here](https://github.com/eclipse-theia/theia/blob/fd5be308a0da5c1e9f660c3afa86ebe60fc5bf8e/packages/plugin-ext/src/plugin/editors-and-documents.ts#L53).
2. Open a notebook document and delete a cell.
3. Assert that the delta that is available in the debugger is a document deletion.
4. Close the notebook document.
5. All cells should now be removed from the documents.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/notebook-editor-widget.tsx
+++ b/packages/notebook/src/browser/notebook-editor-widget.tsx
@@ -258,6 +258,7 @@ export class NotebookEditorWidget extends ReactWidget implements Navigatable, Sa
         this.onDidReceiveKernelMessageEmitter.dispose();
         this.onPostRendererMessageEmitter.dispose();
         this.viewportService.dispose();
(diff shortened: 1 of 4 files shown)


jonah-iden (approved):
Thanks for the fix. Seems to work fine

PR #13611, opened 2024-04-17
Fix github authentication built-in for electron case

### LINKED ISSUE #13599: Authentication UI is not showing Sessions restored from State
### Bug Description:
When an `AuthenticationProvider` restores sessions from stored state (e.g. secrets storage), no `changed` even may be sent. In this case, the front end is never notified. We should notify the front end of any initial sessions when a authentication provider is registered.

### Steps to Reproduce:

1. Log in with some authentication provider
2. Restart Theia
3. Observe the session is active, but you cannot see ti when you click on the little manikin icon at the bottom of the left sidebar: when you click it, it says "You are not logged into any accounts".

Unfortunately, I have no ready scenario to reproduce this: I noticed it with my github login, which you cannot reproduce with the current `github-autentication` extension (https://github.com/eclipse-theia/theia/issues/12821)

### Additional Information

Fix in authentication-ext.ts would look like this:

```
    registerAuthenticationProvider(id: string, label: string, provider: theia.AuthenticationProvider, options?: theia.AuthenticationProviderOptions): theia.Disposable {
        if (this.authenticationProviders.get(id)) {
            throw new Error(`An authentication provider with id '${id}' is already registered.`);
        }

        this.authenticationProviders.set(id, provider);
        provider.getSessions().then(sessions => {
            if (sessions.length > 0) {
                this.proxy.$onDidChangeSessions(id, {
                    added: sessions,
                    removed: [],
                    changed: []
                });
            }
        });
        const listener = provider.onDidChangeSessions(e => {
            this.proxy.$onDidChangeSessions(id, e);
        });

        this.proxy.$registerAuthenticationProvider(id, label, !!options?.supportsMultipleAccounts);

        return new Disposable(() => {
            listener.dispose();
            this.authenticationProviders.delete(id);
            this.proxy.$unregisterAuthenticationProvider(id);
        });
    }
```

- Operating System:
- Theia Version:

tsmaeder:
#### What it does
Add dummy command to fix github authentication built-int flows. Also includes code to consider sessions which are not created, but restored from storage at registration time. 

When deploying Theia as a web-app, one would normally use a web application flow (https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#web-application-flow). However, this will not work, because the VSCode extension explicitly restricts the domains for which that flow can be used. Also, we obviously do not know the "client secret" of the VS Code Github ouath app. 

Fixes #13599
Partial fix for #12821

Contributed on behalof of STMicroelectronics
#### How to test
1. Install the github-authentication and gitlens plugins
2. The gitlens plugin will ask for "workspace trust": say "yes"
3. Click the little manikin icon on the bottom of the left side bar.
4. Observe if you are logged into github, the menu will display a menu allowing you to "sign out"
5. If logged in, sign out, for some reason, you'll have to restart Theia now
6. Open a file that is under git control
7. Observe: gitlens will display text after the end of the current line. If you hover, you'll get a link "Connect to Github"
8. Click that link
9. Follow the instructions to sign into github
10. Observe: when you now hover over the gitlens text, you'll get a link to the relevant PR for the line, instead of "Connect to Github"

#### Follow-ups
Make "github" authentication work for the web app case.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
+++ b/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
@@ -82,6 +82,12 @@ import { CodeEditorWidgetUtil } from '@theia/plugin-ext/lib/main/browser/menus/v
 import { OutlineViewContribution } from '@theia/outline-view/lib/browser/outline-view-contribution';
 
 export namespace VscodeCommands {
+
+    export const GET_CODE_EXCHANGE_ENDPOINTS: Command = {
+        id: 'workbench.getCodeExchangeProxyEndpoints' // this command is used in the github auth built-in
+        // see: https://github.com/microsoft/vscode/blob/191be39e5ac872e03f9d79cc859d9917f40ad935/extensions/github-authentication/src/githubServer.ts#L60
+    };
+
     export const OPEN: Command = {
         id: 'vscode.open'
     };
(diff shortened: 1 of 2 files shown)


martin-fleck-at (approved):
Thank you very much, Thomas! I can confirm the the workflow now works (even though it seems needlessly complicated but that is a different issue entirely ;-) ) with the Electron app and (still) fails with the Browser app. So definitely an improvement.

Issue #13613, opened 2024-04-17
Build fails on Windows on latest Node versions


msujew:
### Bug Description:

Due to the [latest LTS update of Node v18/v20](https://nodejs.org/en/blog/release/v18.20.2), running `cp.spawn` without setting the options to `{ shell: true }` simply yields an error on Windows. You can see this [here](https://github.com/eclipse-theia/theia/issues/13603#issuecomment-2061363524). 

### Steps to Reproduce:

1. Try to build Theia on Windows using an updated LTS version of Node
2. Assert that the build fails due to `spawn EINVAL` errors

### Additional Information

- Operating System: Windows 11

tsmaeder:
@msujew any idea _why_ this breaks our build and how to fix it?

msujew:
@tsmaeder Yes, the `node-pty` version we're using is using a `cp.spawn` call that doesn't have the `{ shell: true }` options on Windows. This is basically illegal now :)

We had a few of those as well, some even in our own build processes. See https://github.com/eclipse-theia/theia/pull/13614 for an extensive list of changes.

tsmaeder:
IMO, even though the node version update is a security update, I don't think building with older node versions is a problem for Theia: the CVE allows for arbitrary commands to be executed as the back-end user by using spawn parameters. However, the user can do this anyway: he just has to open a terminal in the IDE. The only vector I can see where the user is unaware of the commands being executed is covered by workspace trust and that's an entirely different problem. What do you think @msujew ?

msujew:
@tsmaeder Same thoughts. The only issue for our adopters right now is that it's not very obvious why builds are consistently failing on GitHub Actions and on their PCs (in case they installed the latest node.js update). From a security standpoint, there is little to worry about for adopters of Theia.

tsmaeder:
This is still one we need to fix: broken builds with up-to date version of node LTS is likely to break downstream builds. This build will be a community build, so doubly so 🤷

PR #13614, opened 2024-04-17
Fix spawn calls for node LTS versions

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/13613

1. Updates `node-pty` to `1.0.0`, as our currently used version uses a `spawn` call that leads to errors on current LTS versions of node
2. Updates our `child_process#spawn` calls to use the `{ shell: true }` options. This circumvents the changes for the LTS versions.

#### How to test

Confirm that the CI is green **on Windows**

Otherwise, perform a manual smoke test on Windows with focus on the terminal functionality.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-manager/src/application-process.ts
+++ b/dev-packages/application-manager/src/application-process.ts
@@ -32,11 +32,17 @@ export class ApplicationProcess {
     ) { }
 
     spawn(command: string, args?: string[], options?: cp.SpawnOptions): cp.ChildProcess {
-        return cp.spawn(command, args || [], Object.assign({}, this.defaultOptions, options));
+        return cp.spawn(command, args || [], Object.assign({}, this.defaultOptions, {
+            ...options,
+            shell: true
+        }));
     }
 
     fork(modulePath: string, args?: string[], options?: cp.ForkOptions): cp.ChildProcess {
-        return cp.fork(modulePath, args, Object.assign({}, this.defaultOptions, options));
+        return cp.fork(modulePath, args, Object.assign({}, this.defaultOptions, {
+            ...options,
+            shell: true
+        }));
     }
 
     canRun(command: string): boolean {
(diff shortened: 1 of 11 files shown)


tsmaeder:
@msujew I guess this one's not ready for review with the PR checks failing?

msujew:
@tsmaeder Nope, there seem to be some pretty major changes in node-pty (on Unix, it still works as expected on Windows) in more current versions related to the error behavior. I.e. spawning a process on a non-existing file doesn't throw anymore but instead just prints to the console and exits with exit code `1`. I'm currently working on consolidating those changes with our expectations.

msujew:
@tsmaeder I found an earlier version of `node-pty` that doesn't have the erroneous install script, while still using the old (non-breaking) behavior. This seems to satisfy all our tests (except for the Playwright tests, which is probably alright for now). Asking nicely for a review so we can have this in tomorrows release :)

planger:
@msujew Sure, I'm just quite busy the next few days. So I suggest to just skip the failing tests for now and please create an issue for re-enabling them with a fix. I'm happy to look into it next week. Thank you very much!

tsmaeder (commented):
Did you find a description of the fix that explains what adding "shell: true" is the right workaund? I haven't been able to find a description of what nodejs _actually_ changed.

msujew:
> Did you find a description of the fix that explains what adding "shell: true" is the right workaund? I haven't been able to find a description of what nodejs actually changed.

Not really. The changelog just states that they fixed an issue when not using `shell: true`, see [here](https://nodejs.org/en/blog/release/v18.20.2#notable-changes). The consequence is that by not using `shell: true` the `spawn` call will just fail with `EINVAL`. See also other projects that suffer from this issue like https://github.com/forcedotcom/cli/issues/2843.

tsmaeder:
I think this one is a good description: https://github.com/nodejs/node/issues/52681#issuecomment-2076426887. I guess always using "shell: true" won't hurt, usually. As for when we construct the process execution ourselves, we could differentiate between scrips and executables.

tsmaeder (approved):

msujew:
@planger False alarm. It wasn't a false positive, but I actually broke terminals for Unix. I've fixed it again and the tests are green now, so no need for action on your side :)

PR #13615, opened 2024-04-17
working notebook cell language select

jonah-iden:
#### What it does

This allows clicking the language on the bottom right of notebook cells and change the cell language. 
For this it moves some of the editor-command code to select a code language to its own service

Also fixes a small but pretty substantial problem with notebook cells and not being able to create new lines when pressing enter

#### How to test

Open a notebook. 
Create a multi line cell check that enter works correctly
Click on the language tag in the lower right corner of the cell and change the language

#### Review checklist

#### Reminder for reviewers


msujew (commented):
It seems like the language is getting persisted incorrectly. After changing a piece of `Python` code to `Go`, the syntax hightlighting is correctly adjusted:

image

After closing and reopening the notebook, it still says `Go`, but the highlighting has reverted to Python:

image

jonah-iden:
@msujew  it works fine for me at least for all cells except the first. But the reason for that seems to be a bigger one. Seems like the preferred cell language is not set correctly by the extensions. Havn't figured out why yet

jonah-iden:
@msujew forget that i found the culprit. Should work now. At least it does for me

msujew (approved):
Great, works for me perfectly as well now 👍

PR #13625, opened 2024-04-19
chore: refer to commit sha of actions instead of tags

### LINKED ISSUE #13624: Replace GH action tags (e.g. @v3) with commit hashes
In our workflows, we refer to Github actions using tags, such as `actions/checkout@v3`.

The problem here is that tags are mutable and may be changed to point to another commit at any time.
This opens a vector for supply chain attacks when using third-party actions.

A typical solution is to replace all such tags in workflow files with explicit commit hashes, which are immutable.

xai:
#### What it does

Instead of using tags to refer to a version of an action (e.g. `actions/checkout@v3`), we now use its commit sha.

The problem with tags is that they are mutable and can be changed to point to a different commit. This opens a vector for supply chain attacks.

Resolves #13624

#### How to test

Please verify that all tags have been resolved correctly and that no references to tags have been missed.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.github/workflows/translation.yml
+++ b/.github/workflows/translation.yml
@@ -53,7 +53,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@38e0b6e68b4c852a5500a94740f0e535e0d7ba54 # v4.2.4
+        uses: peter-evans/create-pull-request@6d6857d36972b65feb161a90e484f2984215f83e # v6.0.5
         with:
           commiter: ${{ github.actor }} <${{ fromJson(steps.actor_user_data.outputs.data).email }}>
           author: ${{ github.actor }} <${{ fromJson(steps.actor_user_data.outputs.data).email }}>
--- a/.github/workflows/translation.yml
+++ b/.github/workflows/translation.yml
@@ -10,16 +10,16 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
 
       - name: Use Node.js 18.x
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@1a4442cacd436585916779262731d5b162bc6ec7 # v3.8.2
         with:
           node-version: 18.x
           registry-url: "https://registry.npmjs.org"
 
       - name: Use Python 3.x
-        uses: actions/setup-python@v4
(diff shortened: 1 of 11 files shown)


xai:
> There are a few SHAs missing in our (rarely used) `publish-gh-pages.yml` file. See [here](https://github.com/eclipse-theia/theia/blob/06dda6944d966bc2e92c341b127f51ab8d4cf34b/.github/workflows/publish-gh-pages.yml#L47-L64).

Thanks for catching this! I adjusted my regex and found some more.

One case is particularly interesting: Consuming the action `nick-invision/retry` causes a redirect:
```bash
$ curl -sI 'https://github.com/nick-invision/retry'
HTTP/2 301
server: GitHub.com
date: Tue, 30 Apr 2024 10:03:11 GMT
content-type: text/html; charset=utf-8
vary: X-PJAX, X-PJAX-Container, Turbo-Visit, Turbo-Frame, Accept-Encoding, Accept, X-Requested-With
location: https://github.com/nick-fields/retry
[...]
```
I assume that the username has been renamed from nick-invision to nick-fields.
This is the exact thing that becomes dangerous when using tags. Someone could claim the old user name and decide to publish an action of the same name (but maybe with malicious content). I updated the reference in our yml accordingly to use the new user name.

.github/workflows/translation.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@v4
+        uses: peter-evans/create-pull-request@38e0b6e68b4c852a5500a94740f0e535e0d7ba54 # v4.2.4
```
jfaltermeier, line 56: We are using `peter-evans/create-pull-request@6d6857d36972b65feb161a90e484f2984215f83e # v6.0.5` in a different workflow. Can we align those to use the same versions?

jfaltermeier (commented):
LGTM, I have one comment about potentially bumping the version of one action. 
I've verified that all SHAs exist (and also point to a version tag)

.github/workflows/translation.yml
xai: Nice catch!

I updated to v6 in the translation workflow in a follow-up commit.
Now we should be using consistent versions across all workflows for each action.

jfaltermeier (approved):
LGTM, I think the test failures are unrelated, since they fail on other PRs as well at the moment.

jfaltermeier:
I think a rebase should fix the test failures now

xai:
> I think a rebase should fix the test failures now

Let's see :)
Just rebased and force-pushed.

Issue #13549, opened 2024-04-02
[vscode] Support for TestCoverage API


rschnekenbu:
proposed API for **TestCoverage** was finalized and added in VS code 1.88:

in **TestRunProfile** 
```
/**
 * An extension-provided function that provides detailed statement and
 * function-level coverage for a file. The editor will call this when more
 * detail is needed for a file, such as when it's opened in an editor or
 * expanded in the **Test Coverage** view.
 *
 * The {@link FileCoverage} object passed to this function is the same instance
 * emitted on {@link TestRun.addCoverage} calls associated with this profile.
 */
loadDetailedCoverage?: (testRun: TestRun, fileCoverage: FileCoverage, token: CancellationToken) => Thenable<FileCoverageDetail[]>;
```

in **TestRun**:
```
/**
 * Adds coverage for a file in the run.
 */
 addCoverage(fileCoverage: FileCoverage): void;

/**
 * An event fired when the editor is no longer interested in data
 * associated with the test run.
 */
onDidDispose: Event<void>;
```

Added Interfaces:
```
	/**
	 * A class that contains information about a covered resource. A count can
	 * be give for lines, branches, and declarations in a file.
	 */
	export class TestCoverageCount {
		/**
		 * Number of items covered in the file.
		 */
		covered: number;
		/**
		 * Total number of covered items in the file.
		 */
		total: number;

		/**
		 * @param covered Value for {@link TestCoverageCount.covered}
		 * @param total Value for {@link TestCoverageCount.total}
		 */
		constructor(covered: number, total: number);
	}

	/**
	 * Contains coverage metadata for a file.
	 */
	export class FileCoverage {
		/**
		 * File URI.
		 */
		readonly uri: Uri;

		/**
		 * Statement coverage information. If the reporter does not provide statement
		 * coverage information, this can instead be used to represent line coverage.
		 */
		statementCoverage: TestCoverageCount;

		/**
		 * Branch coverage information.
		 */
		branchCoverage?: TestCoverageCount;

		/**
		 * Declaration coverage information. Depending on the reporter and
		 * language, this may be types such as functions, methods, or namespaces.
		 */
		declarationCoverage?: TestCoverageCount;

		/**
		 * Creates a {@link FileCoverage} instance with counts filled in from
		 * the coverage details.
		 * @param uri Covered file URI
		 * @param detailed Detailed coverage information
		 */
[...]

Issue #12010, opened 2022-12-20
[vscode] Implement stubbed VS Code API


planger:
In this issue, we track the VS Code API that we've decided to stub and implement later:

- [ ] #8395
- [ ] #10669

PR #13631, opened 2024-04-22
Stub TestCoverage API

rschnekenbu:
#### What it does
Stub the new API introduced with vscode 1.88 about TestCoverage. This was previously a proposed API.

Fixes https://github.com/eclipse-theia/theia/issues/13549

#### How to test

I did not have a testing plugin yet, as usual plugins used for testing are not implementing the coverage API. For example, Playwright does not yet implement code coverage as seen here: https://github.com/microsoft/playwright/issues/30298. 

#### Follow-ups

Implement TestCoverage API. This will be tracked as a task of https://github.com/eclipse-theia/theia/issues/12010.

#### Review checklist

#### Reminder for reviewers


packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -3330,6 +3330,77 @@ export class TestMessage implements theia.TestMessage {
...
+export class BranchCoverage implements theia.BranchCoverage {
+    constructor(public executed: number | boolean, public location?: Position | Range, public label?: string) { }
+}
+
+@es5ClassCompat
+export class DeclarationCoverage implements theia.DeclarationCoverage {
+    constructor(public name: string, public executed: number | boolean, public location: Position | Range) { }
+}
+@es5ClassCompat
+export class TestCoverageCount implements theia.TestCoverageCount {
```
tsmaeder, line 3398: `theia.TestCoverageCount` is not an interface. Why the "implements"?

packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -3330,6 +3330,77 @@ export class TestMessage implements theia.TestMessage {
...
+    ) { }
+}
+
+@es5ClassCompat
+export class StatementCoverage implements theia.StatementCoverage {
+    constructor(public executed: number | boolean, public location: Position | Range, public branches: BranchCoverage[] = []) { }
+}
+
+export namespace StatementCoverage {
+    export function is(coverage: unknown): coverage is StatementCoverage {
```
tsmaeder, line 3384: Wouldn't `instanceof` work just as well, since this is a class?

tsmaeder:
@rschnekenbu we should at least have a test plugin that exercises the stubbed code: extensions should at least not crash when they call the API.

tsmaeder (commented):
The `FileCoverage` API is weird in that you can assign to the public fields, but changes to those fields are not handled. Seems to be that same in VS Code, though. Seems undercooked for a GA API on their side, IMO.

packages/plugin-ext/src/plugin/types-impl.ts
rschnekenbu: Good catch, I removed this namespace addition in favor of a instanceof.

packages/plugin-ext/src/plugin/types-impl.ts
rschnekenbu: I removed this declaration, that indeed makes no sense.

rschnekenbu:
I addressed the review comments in a following patch. 
Here is a testing extension that adds coverage support for the test runs and also provides the optional method TestRunProfile.loadCoverageDetail for our TestRunProfile.
src: [test-extension-sample-0.1.0-src.zip](https://github.com/eclipse-theia/theia/files/15092078/test-extension-sample-0.1.0-src.zip)
extension: 
[test-extension-sample-0.1.0.zip](https://github.com/eclipse-theia/theia/files/15092084/test-extension-sample-0.1.0.zip)

To run the extension:
1. Start theia browser example on a workspace and install the provided extension
2. Use the command to create a test run profile 
3. Then in the test run view, run the tests. They should run as usual, even if a coverage is added. No errors should be logged or displayed. 

Here is the behavior on vscode 1.88:
vscode-13549

And on theia browser example:
theia-13549

tsmaeder (approved):

PR #13640, opened 2024-04-25
fixes clear cell outputs command and improves toolbar creation

jonah-iden:
#### What it does

This fixes the `clear cell output` and `change presentation` from the cell sidebar context menu

#### How to test

Open or create a notebook including a cell with multiple output representations.
For example the following (requires PIL installed in the kernel)
```python
 from PIL import Image

img_1 = Image.new("RGBA", (1600, 200), color="red")

display(img_1)
```
Check that changing the ouptut presentation works.
Check that clearing the output works.

#### Follow-ups

There is currently no support for multiple outputs yet. So if you have two outputs both are affected by the actions.
Also it seems there is still a bug with going back to images after changing to text

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/contributions/notebook-cell-actions-contribution.ts
+++ b/packages/notebook/src/browser/contributions/notebook-cell-actions-contribution.ts
@@ -324,7 +324,7 @@ export class NotebookCellActionContribution implements MenuContribution, Command
             }
         });
         commands.registerCommand(NotebookCellCommands.CLEAR_OUTPUTS_COMMAND, this.editableCellCommandHandler(
-            (notebook, cell) => notebook.applyEdits([{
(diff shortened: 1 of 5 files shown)


msujew (commented):
This works partially: When initially rendering the output, changing the presentation doesn't work on the first attempt. Only after clearing the output and then rerunning the cell does it work as expected. Also, changing from image to text and then back to image simply removes (?) the output. Only after running the cell again, I get to see the image output. See the attached video:



https://github.com/eclipse-theia/theia/assets/4377073/50ce395f-1b9c-4f89-b3e7-7f117af5c420

msujew (approved):
Alright, in this case, this looks good to me 👍

Issue #13642, opened 2024-04-25
Properly separate back-end and headless plugins


tsmaeder:
### Feature Description:
Currently, we deploy back end and headless plugins the same way. This can lead to issues like https://github.com/eclipse-theia/theia/issues/13638. There are a couple of questions open for me:

1. Can plugins be loaded both as headless and back end plugins? Does that ever make sense?
2. Since plugins _can_ be run both as back end and front end plugins, does it make sense to have two different deployment paths for those? 
3. For each plugin, we need to decide where a plugin should be run. For extensions that can live in different places, there needs to be a policy. For FE/BE plugins, the scope of the decision is the front end, for headless plugins, the scope of the decision is the back end process.
4. Currently, we decide based on the "entry point" where a plugins runs. However, this is not optimal: many plugins (i.e. those providing only syntax) may have no entry points, but they could probably run in the front end no problem.

JonasHelming:
@cdamus @martin-fleck-at

cdamus:
On item 1 I can say that it is a deliberate decision to let plug-in providers offer headless and backend plug-ins in the same package for convenience and clarity of distribution, installation, and update. I'm working on a project that foresees taking advantage of this single-sourcing of both kinds of plug-ins: headless plug-ins contributing application-specific APIs and backend plugins contributing UI that in part delegates to other plug-ins that contribute extensions on those custom APIs.

On item 4 this is one reason why a plugin can optionally use the `"theia-headless"` value for the `"engine"` property, if it otherwise wouldn't provide a headless entry-point script. I'm not sure myself how useful this is as it introduces issues in VSIX packaging, I think.

Issue #13641, opened 2024-04-25
Don't wait for plugin deployment at startup


tsmaeder:
### Feature Description:
Currently, we wait for the initial "deployment" of plugins before the front end finishes initializing. This is not necessary, as we can handle the case where plugins are added after startup. The delay can be considerable in the case where we really unzip some large plugins like in https://github.com/eclipse-theia/theia/issues/13638

Issue #13638, opened 2024-04-24
Drop-in VS Code extension does not show up as Installed.


tsmaeder:
### Bug Description:
When I install a particular VS Code extension it does not show up as installed (even though it is uncompressed to the "deployedPlugins"-folder.

### Steps to Reproduce:

1. Have gitlens installed from open-vsix
2. Download the [spring-boot extension](https://open-vsx.org/extension/vmware/vscode-spring-boot) and put it into `~/.theia/extensions`
3. Start Theia Electron
4. Observe: the command line output shows that the extension is being decompressed into `~/theia/deployedPlugins`
5. Observe: eventually, gitlens and other installed extensions show up in the "extensions" view in the "installed" section
6. Observe: the spring boot extension never shows up
7. Observe: there are no commands starting with "Spring" available from the command pallette
8. Restart Theia electron (not reload, restart the whole process).
9. Observe: the spring-boot extensions is available and shows up in the extensions view.

### Additional Information

- Operating System: Windows
- Theia Version: 1.48 master

tsmaeder:
Note that this is not happening in the browser application, only electron.

tsmaeder:
I tried the same scenario with only the "github-authentication" plugin from microsoft, and that works. This might simply be a timing issue.

tsmaeder:
Other difference: the spring-boot extension has redhat.java as a dependency.

tsmaeder:
Turns out in `browser/hosted-plugin.ts#acceptPlugin()`, sometimes returns false for a bunch of plugins.

tsmaeder:
I finally got to the bottom of this: the problem came with the introduction of back-end plugins: there is a `Deferred` in `hosted-plugin-deployer-handler.ts` called `backendPluginsMetadataDeferred`. Since headless plugins are deployed using `deployBackendPlugins()`, the deferred gets resolved when deploying back and plugins. However, the deployment is done in parallel in `plugin-deployer-impl.ts#deployPlugins()`. Since there are no headless plugins, this "deployment" is very fast and the deferred gets resolved quite early. In the normal case, when no actualy unzipping of large plugins takes place, the list of plugins we get in the front end for loading contributions is complete. But when it takes longer, the list of plugins is incomplete: in fact, in my debug case it only includes about 1/3 of the actually installed plugins. The short term fix is to run the deployments in sequence and do the back end plugins first.

The long term fix for this would be twofold:

[...]

PR #13643, opened 2024-04-25
Reorder back-end plugin deployment to prevent early promise resolution

tsmaeder:
#### What it does
Upon startup, we the front end indirectly waits for the initial deployment of plugins before loading their contributions. Because of parallel execution of back end/headless deployment, the corresponding promise was resolved too early and the list of plugins was incomplete.

Workaround for #13638

#### How to test
Repeat the "how to reproduce" from the linked issue. The installed plugins should show up reliably in the UI.

#### Follow-ups
https://github.com/eclipse-theia/theia/issues/13641
https://github.com/eclipse-theia/theia/issues/13642

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/node/plugin-deployer-impl.ts
+++ b/packages/plugin-ext/src/main/node/plugin-deployer-impl.ts
@@ -298,13 +298,12 @@ export class PluginDeployerImpl implements PluginDeployer {
         const pluginPaths = [...acceptedBackendPlugins, ...acceptedHeadlessPlugins].map(pluginEntry => pluginEntry.path());
         this.logger.debug('local path to deploy on remote instance', pluginPaths);
 
-        const deployments = await Promise.all([
-            // headless plugins are deployed like backend plugins
-            this.pluginDeployerHandler.deployBackendPlugins(acceptedHeadlessPlugins),
-            // start the backend plugins
-            this.pluginDeployerHandler.deployBackendPlugins(acceptedBackendPlugins),
-            this.pluginDeployerHandler.deployFrontendPlugins(acceptedFrontendPlugins)
-        ]);
+        const deployments = [];
+        // start the backend plugins
+        deployments.push(await this.pluginDeployerHandler.deployBackendPlugins(acceptedBackendPlugins));
+        // headless plugins are deployed like backend plugins
+        deployments.push(await this.pluginDeployerHandler.deployBackendPlugins(acceptedHeadlessPlugins));
+        deployments.push(await this.pluginDeployerHandler.deployFrontendPlugins(acceptedFrontendPlugins));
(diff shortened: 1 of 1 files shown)


tsmaeder:
@cdamus might be a good idea if you had a quick look.

packages/plugin-ext/src/main/node/plugin-deployer-impl.ts
```diff
@@ -298,13 +298,12 @@ export class PluginDeployerImpl implements PluginDeployer {
...
-            // start the backend plugins
-            this.pluginDeployerHandler.deployBackendPlugins(acceptedBackendPlugins),
-            this.pluginDeployerHandler.deployFrontendPlugins(acceptedFrontendPlugins)
-        ]);
+        const deployments = [];
+        // start the backend plugins
+        deployments.push(await this.pluginDeployerHandler.deployBackendPlugins(acceptedBackendPlugins));
+        // headless plugins are deployed like backend plugins
+        deployments.push(await this.pluginDeployerHandler.deployBackendPlugins(acceptedHeadlessPlugins));
+        deployments.push(await this.pluginDeployerHandler.deployFrontendPlugins(acceptedFrontendPlugins));
```
msujew, line 306: Minor: Since we only count the number of deployments anyway, wouldn't something like this be more reasonable? And then simply return with `deployments.length`.

```suggestion
        const deployments = [];
        // start the backend plugins
        deployments.push(...await this.pluginDeployerHandler.deployBackendPlugins(acceptedBackendPlugins));
        // headless plugins are deployed like backend plugins
        deployments.push(...await this.pluginDeployerHandler.deployBackendPlugins(acceptedHeadlessPlugins));
        deployments.push(...await this.pluginDeployerHandler.deployFrontendPlugins(acceptedFrontendPlugins));
```

msujew (approved):
Great, thanks 👍 

I assume that this should also fix that @jonah-iden was experiencing when testing https://github.com/eclipse-theia/theia/pull/13369.

tsmaeder:
@msujew 

> of the builtin extensions only 26 of 83 were installed. Even basic things like css and dart language basics

For me it's typically 27, but yeah, that looks like the same problem.

packages/plugin-ext/src/main/node/plugin-deployer-impl.ts
tsmaeder: The `deploy<X>Plugins()` calls return a count or undefined, not the actual plugins and I really don't want to invest time in what is basically a workaround for a larger problem.

msujew: Alright, I see. Ok

PR #13645, opened 2024-04-25
Align to vscode notebook commands

jonah-iden:
this makes more keybindings available.
Also implements to notbookOutputInputFocused context key

#### What it does

It aligns some commands with the ids from vscode so that shortcuts contributed by jupyter should work. 
Also implements the `notebookOutputInputFocused` context key which most of these shortcuts are using

#### How to test

open a notebook and create a cell like:
```python
import ipywidgets as widgets
widgets.Text()
```

The following shortcuts from jupyter should now work: `l`, `o`, `c`, `v`, `x`
Also when focusing the input fied in the cell output none of the shoudl activate

#### Review checklist

#### Reminder for reviewers


msujew (approved):
Looks pretty good to me. The keybindings work well 👍 

Collapsing the output and reopening it still doesn't work, but I guess that's out of scope for this PR.

PR #13648, opened 2024-04-26
Fixed storing of the notebook-outlineview state data

jonah-iden:
#### What it does

This fixes the notebook outline view so it does not break the layout storing on reload.
Previously this lead to the layout-restorer throwing a circular structure error.

#### How to test

Open a notebook and check the outline view is working.
Reload the page and confirm the layout is correctly restored

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/contributions/notebook-outline-contribution.ts
+++ b/packages/notebook/src/browser/contributions/notebook-outline-contribution.ts
@@ -21,16 +21,20 @@ import { OutlineViewService } from '@theia/outline-view/lib/browser/outline-view
 import { NotebookModel } from '../view-model/notebook-model';
 import { OutlineSymbolInformationNode } from '@theia/outline-view/lib/browser/outline-view-widget';
 import { NotebookEditorWidget } from '../notebook-editor-widget';
-import { DisposableCollection, URI } from '@theia/core';
+import { DisposableCollection, isObject, URI } from '@theia/core';
 import { CellKind, CellUri } from '../../common';
 import { NotebookService } from '../service/notebook-service';
 export interface NotebookCellOutlineNode extends OutlineSymbolInformationNode {
-    uri: string;
+    uri: URI;
 }
 
 export namespace NotebookCellOutlineNode {
     export function is(element: object): element is NotebookCellOutlineNode {
-        return TreeNode.is(element) && OutlineSymbolInformationNode.is(element) && 'uri' in element && (element as NotebookCellOutlineNode).uri.startsWith(CellUri.cellUriScheme);
+        return TreeNode.is(element)
+            && OutlineSymbolInformationNode.is(element)
+            && isObject<NotebookCellOutlineNode>(element)
+            && typeof element.uri === 'object'
+            && (element as NotebookCellOutlineNode).uri.scheme === CellUri.cellUriScheme;
     }
 }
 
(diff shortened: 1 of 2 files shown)


packages/notebook/src/browser/contributions/notebook-outline-contribution.ts
```diff
@@ -21,18 +21,16 @@ import { OutlineViewService } from '@theia/outline-view/lib/browser/outline-view
...
 export interface NotebookCellOutlineNode extends OutlineSymbolInformationNode {
-    notebookCell: NotebookCellModel;
-    uri: URI;
+    uri: string;
 }
 
 export namespace NotebookCellOutlineNode {
     export function is(element: object): element is NotebookCellOutlineNode {
-        return TreeNode.is(element) && OutlineSymbolInformationNode.is(element) && 'notebookCell' in element;
+        return TreeNode.is(element) && OutlineSymbolInformationNode.is(element) && 'uri' in element && (element as NotebookCellOutlineNode).uri.startsWith(CellUri.cellUriScheme);
```
msujew, line 33: Suggestion: Since this method is executed on arbitrary `OutlineSymbolInformationNode` instances (many of which use `uri: URI`), this code leads to a lot of errors. You should be more conservative about your type checking here:
```suggestion
        return TreeNode.is(element)
            && OutlineSymbolInformationNode.is(element)
            && isObject<NotebookCellOutlineNode>(element)
            && typeof element.uri === 'string'
            && element.uri.startsWith(CellUri.cellUriScheme);
```

packages/notebook/src/browser/contributions/notebook-outline-contribution.ts
```diff
@@ -21,18 +21,16 @@ import { OutlineViewService } from '@theia/outline-view/lib/browser/outline-view
...
 import { OutlineSymbolInformationNode } from '@theia/outline-view/lib/browser/outline-view-widget';
 import { NotebookEditorWidget } from '../notebook-editor-widget';
-import { NotebookCellModel } from '../view-model/notebook-cell-model';
 import { DisposableCollection, URI } from '@theia/core';
 import { CellKind, CellUri } from '../../common';
 import { NotebookService } from '../service/notebook-service';
 export interface NotebookCellOutlineNode extends OutlineSymbolInformationNode {
-    notebookCell: NotebookCellModel;
-    uri: URI;
+    uri: string;
```
msujew, line 28: Suggestion: Independent of the comment above, we should continue using `URI` here. The `OutlineViewWidget` inflates nodes back using the `URI` object. See [here](https://github.com/eclipse-theia/theia/blob/9c9456b8f7d32f9788d60f83377f109f1743aaf5/packages/outline-view/src/browser/outline-view-widget.tsx#L208). This should make the rest of the code better to maintain as well.

msujew (commented):
Works pretty well, but found my console spammed with errors (and the tests fail due to that as well). See below for root causes/solutions.

packages/notebook/src/browser/contributions/notebook-outline-contribution.ts
jonah-iden: ah i didn't know that. I saw some error because of a path element so i assumed it was because of the uri as well. Should be better now

packages/notebook/src/browser/contributions/notebook-outline-contribution.ts
```diff
@@ -21,18 +21,20 @@ import { OutlineViewService } from '@theia/outline-view/lib/browser/outline-view
...
 }
 
 export namespace NotebookCellOutlineNode {
     export function is(element: object): element is NotebookCellOutlineNode {
-        return TreeNode.is(element) && OutlineSymbolInformationNode.is(element) && 'notebookCell' in element;
+        return TreeNode.is(element)
+            && OutlineSymbolInformationNode.is(element)
+            && isObject<NotebookCellOutlineNode>(element)
+            && typeof element.uri === 'object'
+            && (element as NotebookCellOutlineNode).uri.scheme === CellUri.cellUriScheme;
```
msujew, line 37: Minor: The casting is already done by `isObject`
```suggestion
            && element.uri instanceof URI
            && element.uri.scheme === CellUri.cellUriScheme;
```

msujew (approved):
Looks good to me, thanks 👍

PR #13656, opened 2024-04-29
Select next notebook cell on first or last line of editor

jonah-iden:
#### What it does

This replicates the behaviour of vscode when having the cursor on the first or last line in a cell editor and pressing `up`/`down` it will select the next/previous cell as if it was one text editor.

#### How to test

Open or create a notebook with multiple cells. Focus an editor and move up down through the cells with the arrow key

#### Follow-ups

there is not really a difference between edit and command mode currently in theias notebook editor. So when getting to a markdown cell and then moving to a code cell, it will not focus the editor again

#### Review checklist

#### Reminder for reviewers


msujew (approved):
Looks good to me 👍

Issue #13657, opened 2024-04-29
Terminal not spawning on Ubuntu


jfaltermeier:
### Bug Description:

Currently when I try to open a Terminal (e.g. via the application menu), it stays empty and I get errors on the console. 

Might be related to https://github.com/eclipse-theia/theia/issues/13613
@msujew  

So this might only affect Ubuntu/Linux and potentially Mac. 

### Steps to Reproduce:

1. Launch Theia Electron or Browser Example
2. `Menu Bar -> Terminal -> New Terminal`
3. Terminal stays empty, this appears on the log for me

```
2024-04-29T10:23:26.923Z terminal ERROR Error: posix_spawn failed: No such file or directory
    at new UnixTerminal (/home/johannes/Git/theia/examples/browser/lib/backend/vendors-node_modules_drivelist_build_Release_drivelist_node-node_modules_stroncium_procfs_lib-adc4cb.js:47870:24)
    at spawn (/home/johannes/Git/theia/examples/browser/lib/backend/vendors-node_modules_drivelist_build_Release_drivelist_node-node_modules_stroncium_procfs_lib-adc4cb.js:47519:12)
    at ShellProcess.createPseudoTerminal (/home/johannes/Git/theia/examples/browser/lib/backend/main.js:23946:47)
    at startTerminal (/home/johannes/Git/theia/examples/browser/lib/backend/main.js:23902:29)
    at new TerminalProcess (/home/johannes/Git/theia/examples/browser/lib/backend/main.js:23933:43)
    at new ShellProcess (/home/johannes/Git/theia/examples/browser/lib/backend/main.js:27806:9)
    at createInstanceWithInjections (/home/johannes/Git/theia/examples/browser/lib/backend/vendors-node_modules_fuzzy_lib_fuzzy_js-node_modules_inversify_es_inversify_js-node_modules_i-0dedd3.js:2701:20)
    at _createInstance (/home/johannes/Git/theia/examples/browser/lib/backend/vendors-node_modules_fuzzy_lib_fuzzy_js-node_modules_inversify_es_inversify_js-node_modules_i-0dedd3.js:2691:22)
    at resolveInstance (/home/johannes/Git/theia/examples/browser/lib/backend/vendors-node_modules_fuzzy_lib_fuzzy_js-node_modules_inversify_es_inversify_js-node_modules_i-0dedd3.js:2786:18)
[...]

PR #13658, opened 2024-04-29
Fix unix terminals for bundled apps

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/13657

While updating the `node-pty` version in https://github.com/eclipse-theia/theia/pull/13614, I've removed code to copy the `spawn-helper` binary to the target directory, since the binary did no longer exist in `v1.0.0` of the library. However, after I downgraded to `v0.11.0.beta24`, I forgot to add the copy call again. This breaks unix terminal support in all bundled apps.

This change adds the copy code back.

#### How to test

Tests should be green (especially the playwright test) and terminal should work on unix platforms.

#### Review checklist

#### Reminder for reviewers


jfaltermeier (approved):
Thanks, this fixes the issue for me

PR #13663, opened 2024-04-29
Use responsive design for the main notebook toolbar

jonah-iden:
#### What it does

When the notebook editor widget in not wide enough to contain all main toolbar items it will now add a `...` menu containing all overflowing items

#### How to test

Open a notebook and adjust the widget size. You should see groups being added and removed again from the context menu. 

#### Follow-ups

Currently only the top level groups are added to context menu instead of singular `MenuAction`s like in vscode

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/view/notebook-main-toolbar.tsx
+++ b/packages/notebook/src/browser/view/notebook-main-toolbar.tsx
@@ -129,8 +129,8 @@ export class NotebookMainToolbar extends React.Component<NotebookMainToolbarProp
         if (this.gapElement && this.gapElement.getBoundingClientRect().width < NotebookMainToolbar.MIN_FREE_AREA && this.state.numberOfHiddenItems < numberOfMenuItems) {
             this.setState({ ...this.state, numberOfHiddenItems: this.state.numberOfHiddenItems + 1 });
             this.lastGapElementWidth = this.gapElement.getBoundingClientRect().width;
-        } else if (this.gapElement && this.gapElement.getBoundingClientRect().width > this.lastGapElementWidth) {
-            this.setState({ ...this.state, numberOfHiddenItems: this.state.numberOfHiddenItems <= 1 ? 0 : this.state.numberOfHiddenItems - 1 });
+        } else if (this.gapElement && this.gapElement.getBoundingClientRect().width > this.lastGapElementWidth && this.state.numberOfHiddenItems > 0) {
+            this.setState({ ...this.state, numberOfHiddenItems: 0 });
             this.lastGapElementWidth = this.gapElement.getBoundingClientRect().width;
         }
     }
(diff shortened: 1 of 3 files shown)


packages/notebook/src/browser/view/notebook-main-toolbar.tsx
```diff
@@ -97,18 +114,64 @@ export class NotebookMainToolbar extends React.Component<NotebookMainToolbarProp
...
             </div>
-        </div>;
+        </div >;
+    }
+
+    protected gapElementChanged(element: HTMLDivElement | null): void {
+        this.gapElement = element ?? undefined;
+        if (this.gapElement) {
+            this.lastGapElementWidth = this.gapElement.getBoundingClientRect().width;
+            new ResizeObserver(() => this.calculateItemsToHide()).observe(this.gapElement);
```
msujew, line 173: Suggestion: This code leaks the `ResizeObserver`. Every time the gap element is changed, another resize observer is added, which leads to a lot of calls to `calculateItemsToHide`. After changing the size of the window a few times, I already get more than 50+ calls to `calculateItemsToHide` by just maximizing the window.

The component should have a single resize observer that uses `observe` and `unobserve` as expected by the API.

msujew (commented):
Works pretty well in general, but we have a reference/memory leak due to `ResizeObserver` in there.

packages/notebook/src/browser/view/notebook-main-toolbar.tsx
jonah-iden: true should be fixed now

msujew (commented):
That's better, but I've noticed that with this change there is a minor regression: When resizing the window to be "too small" for anything to be displayed in the widget, the widget just renders as empty:



https://github.com/eclipse-theia/theia/assets/4377073/d3e7c069-4c9e-486e-bb3a-3b24097f9dc3



I'm also getting this error in the console:

```
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```

jonah-iden:
Oh overlooked that. good chatch im gonna take another look

jonah-iden:
@mark not rendering when too small and the maximum update deapth should be fixed now.

msujew:
@jonah-iden Linting fails in the CI.

msujew (commented):
With the newest change it looks like the amount of displayed groups can only increase/decrease by one. However, this isn't the expected behavior when minimizing/maximizing the application:


https://github.com/eclipse-theia/theia/assets/4377073/a2f12b73-ee59-439e-8603-2f1f503d0131

jonah-iden:
@msujew are you sure you build it correctly? 

https://github.com/eclipse-theia/theia/assets/34068281/62e32ed0-d056-4d0f-8449-9f4e76e8e9ae

For me it looks like this

msujew:
@jonah-iden slowly resizing the app works for me as well. The issue comes up when the resizing is happening in a single discrete step, like when maximizing the window. In this case [this code](https://github.com/eclipse-theia/theia/blob/16279ca286ee65e3acdfc687230980e169efd6c8/packages/notebook/src/browser/view/notebook-main-toolbar.tsx#L130) will only unhide a single group in the toolbar.

jonah-iden:
oh ok

msujew (approved):
Thank you, everything works well now 👍

Issue #13662, opened 2024-04-29
SocketWriteBuffer repeats errors many times


CamilleLetavernier:
### Bug Description:

When using reloadOnReconnect, and the client disconnects, we keep a cache for each active socket channel, with a 100kB buffer. Once this buffer is full, an error is thrown:

`Max disconnected buffer size exceeded by adding 50186 bytes`

Then, the same error is repeated whenever a message is posted to this same channel, which results in a potentially heavy log spam. Moreover, the classes involved use `protected` methods, but are not actually exported, making it difficult to fix in specific applications (or to increase buffer size, etc).

I think we need two steps to improve this:

- only log the error once per SocketWriteBuffer instance
- open the API a bit, so this can be customized in specific apps

### Steps to Reproduce:

Note: the bug can be difficult to reproduce in Theia, because you need to hit the 100kB buffer limit with a single service, before the disconnected client session expires and the buffer is closed. This means that, with the default settings, you have 30 seconds to fill 100kB with a single service, which probably won't happen under normal conditions. This could however easily happen e.g. during a large build, which will generate many file system events.

1. Increase the frontendConnectionTimeout to a higher value, so the buffer isn't cleared too quickly (e.g. `1800000` for 30 minutes). This gives more time to reproduce the issue
2. Load the Theia application, then refresh the tab in your browser. This will create two frontend sessions: one for the initial page, and one for the refreshed page. The initial session will use a `SocketWriteBuffer` and fill it with messages; the refreshed session will work as expected.
3. Copy-paste deep folder hierarchies in the Explorer multiple times. This should generate enough messages to fill the buffer quickly with `/services/remote-filesystem` change notifications. 
4. Observe the backend logs: `Max disconnected buffer size exceeded by adding ... bytes` will be logged many times, and logged again for every update after that

### Additional Information

- Operating System: any
- Theia Version: >= 1.45 (reconnect behavior), tested on 1.45.1 (custom app) and 1.48.0 (Theia example browser app)

PR #13665, opened 2024-04-29
SocketWriteBuffer: only throw buffer error once per channel

CamilleLetavernier:
refs #13662

#### What it does

- add a flag in SocketWriteBuffer to only throw errors once. 
- export ReconnectableSocketChannel to make it easier to override buffer behavior in custom applications

I'm not really satisfied with this patch, as I was hoping we could only block the "logging" part of the issue. But since the `Max disconnected buffer size exceeded` is handled pretty late, it's hard to prevent multiple logs while still throwing the exception; so I'm only throwing the exception once, and then silently ignoring following errors, which is a bit awkward. I'm not quite sure how to properly solve this issue.

Regarding the exported types, I would be tempted to also change the visibility of some fields in both classes (`ReconnectableSocketChannel` and `SocketWriteBuffer`) from private to protected; as without this, actually overriding the behavior is pretty difficult. Thoughts?

#### How to test

See https://github.com/eclipse-theia/theia/issues/13662

#### Follow-ups

If we could detect that the client "closed" or "refreshed" the tab (as opposed to "lost connection"), we could probably immediately clean up the session. Even if the client reopens the tab, it will happen with a different frontendId, so it will not rely on the disconnected socket buffer. However, this is probably a more complex task.

#### Review checklist

#### Reminder for reviewers


msujew (approved):
LGTM 👍

CamilleLetavernier:
Hi Mark,

Thanks for the review! I've discussed in more details with @tsmaeder , and it turns out this approach probably isn't the right way to handle the issue. Hiding further errors may lead to inconsistencies between the backend and frontend.

For now, the only solution is to use shorter timeouts (short enough that we won't saturate the buffer), and use reloadOnReconnect to fully refresh the frontend after a longer disconnect.

So I think we should simply cancel this PR

Issue #12863, opened 2023-08-25
Improve notebook UI/UX


msujew:
After merging #12442, there are still quite a few issues to be tackled related to the UI/UX of the notebook support:

- [ ] `files.autoSave` does not seem to be respected by the notebook editor widget/model.
- [ ] Closing a dirty notebook editor widget does not prompt to save the widget.
- [x] Switching file-icon themes is not respected by the notebook
- [x] Improve notebook scrolling:

https://github.com/eclipse-theia/theia/assets/40359487/dbf87893-bb3c-4664-8016-041a04f89ac3

- [x] Improve toolbar items for the notebook https://github.com/eclipse-theia/theia/pull/13334
  - Items are not rendered similarly to the rest of the application, rounded border
  - The tooltip for the items do not display their full name similarly to VS Code
- [x] "Select Another Kernel" should allow to go back in the quick-input prompt

Feel free to use this as a general tracker, and add more issues here.

jbicker:
[Improved](https://github.com/eclipse-theia/theia/pull/13334) the toolbar items. 
Will look into the scrollbar issue now.

Issue #12844, opened 2023-08-21
Save confirmation popup does not save on custom binary editor


safisa:
### Bug Description:
Hi,
We are using vscode extension with custom editors, and some are based on a binary data model (CustomEditorProvider). In the latest releases there is a regression in the save confirmation popup when pressing save on a dirty/modified editor.
I have found that the PR caused this issue is back to Theia 1.33 - PR https://github.com/eclipse-theia/theia/pull/11261
Once going back one commit before this change, the save works!

Note that using the custom text model (CustomTextEditorProvider, eg. example.cscratch) is working.
 

### Steps to Reproduce:

1. use the famous cat-customs-0.0.2.vsix extension sample
2. open the example.pawDraw (binary format)
3. do some changes
4. see that the tab/file is in dirt mode
5. close the tab, and a save confirmation will popup
6. press save
7. reopen the file
8. file is still the same (save didn't work)

### Additional Information

- Operating System: mac/win
- Theia Version: 1.33-1.40

Issue #8722, opened 2020-11-06
Theia runs save on app startup if `"editor.autoSave": "on"` is configured in the settings.json


kittaakos:
When I have the following `settings.json` in my workspace and the app starts, a save event will be fired from the editor commands:

`settings.json`:
```json
{
  "editor.autoSave": "on"
}
```

https://github.com/eclipse-theia/theia/blob/d66074711a6e67136ba22dbf0792acf70a813bc8/packages/editor/src/browser/editor-command.ts#L171-L175

Theia incorrectly assumes there was a preference change, but there wasn't. Auto-save is enabled by default, so I would expect the exact same behavior when I have an empty `settings.json` or one with `"editor.autoSave": "on"`.

### Steps to Reproduce:

1.
2.
3.

### Additional Information

- Operating System:
- Theia Version:

kittaakos:
This [PR](https://github.com/eclipse-theia/theia/pull/8163) causes the undesired side effect.

kittaakos:
Nice, the default auto-save preference is `off`:

https://github.com/eclipse-theia/theia/blob/d66074711a6e67136ba22dbf0792acf70a813bc8/packages/editor/src/browser/editor-preferences.ts#L1246

But editor command ignores this and assumes auto-save if on when not set:

https://github.com/eclipse-theia/theia/blob/d66074711a6e67136ba22dbf0792acf70a813bc8/packages/editor/src/browser/editor-command.ts#L333

kittaakos:
Here is a workaround for the time being:

`your-module.ts`:
```ts
bind(EditorCommandContribution).toSelf().inSingletonScope();
rebind(TheiaEditorCommandContribution).toService(EditorCommandContribution);
```

`your-editor-commands.ts`:
```ts
import { injectable, postConstruct } from 'inversify';
import { EditorCommandContribution as TheiaEditorCommandContribution } from '@theia/editor/lib/browser/editor-command';

@injectable()
export class EditorCommandContribution extends TheiaEditorCommandContribution {

    @postConstruct()
    protected init(): void {
        // Workaround for https://github.com/eclipse-theia/theia/issues/8722.
        this.editorPreferences.onPreferenceChanged(({ preferenceName, newValue, oldValue }) => {
            if (preferenceName === 'editor.autoSave') {
                const autoSaveWasOnBeforeChange = !oldValue || oldValue === 'on';
                const autoSaveIsOnAfterChange = newValue === 'on';
[...]

Issue #7480, opened 2020-04-01
Unexpected Value in Preference Change Notification


tsmaeder:
When a preference value is an object, the value in effect should be constructed by merging the objects from different preference scopes. So when I listen for preference changes on an object-valued preference, I would expect the the "newValue" of the preference change to be the new, merged object value. However, the "newValue" is whatever was passed to "setPreference()".

tsmaeder:
Here's a test to illustrate the issue. 

```typescript
    it('should report a change when object preferences are set in more general scope', () => {
        prefSchema.setSchema({
            properties: {
                'test.object': {
                    type: 'object',
                    scope: 'resource'
                }
            }
        });
        let isOk = false;

        prefService.onPreferenceChanged(event => {
            if (deepEql(event.newValue, {
                first: 'zoz',
                second: 'bar'
            })) {
                isOk = true;
            }
        });

        const userProvider = getProvider(PreferenceScope.User);
        const workspaceProvider = getProvider(PreferenceScope.Workspace);
        userProvider.setPreference('test.object', { first: 'foo' });
        workspaceProvider.setPreference('test.object', { second: 'bar' });
        userProvider.setPreference('test.object', { first: 'zoz' });
        assert.ok(isOk);
    });
```

The debugger shows that a change is notified with new Value being `{ first: 'zoz' }`

vince-fugnitto:
I'm trying to understand, do you mean you expect a combination (merge) of the user and workspace preference value? I thought that the more precise scope (workspace in this case) would be the value which takes precedence and no merging happens.

tsmaeder:
Not for object-valued preferences, see https://code.visualstudio.com/api/references/vscode-api#WorkspaceConfiguration

tsmaeder:
@vince-fugnitto thx for adding the labels.

vince-fugnitto:
> Not for object-valued preferences, see https://code.visualstudio.com/api/references/vscode-api#WorkspaceConfiguration

I see, thank you for clarifying, merging can happen for preferences such as `files.exclude`.

akosyakov:
@tsmaeder Is it about vscode compatibility or only about PreferenceService?

I think we should add compatibility tests for change events as we did for other PreferenceService APIs:
- create a VS Code extension API tests and test under different conditions
- migrate such tests to Theia integration tests (or run VS Code extension API tests against Theia)

tsmaeder:
[...]

PR #13683, opened 2024-05-05
Refactor auto save mechanism

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/12863
Closes https://github.com/eclipse-theia/theia/issues/12844
Closes https://github.com/eclipse-theia/theia/issues/8722

Enhances the `SaveResourceService` to perform all the auto save preference work. Previously, all editor classes had to do their own handling of the `afterDelay` auto save functionality. This has been centralized into one service. Therefore, custom editors and notebook editors now behave exactly the same as monaco text editors in regards to auto saves.

This also fixes a long standing issue, where `onFocusChange` and `onWindowChange` have been handled as if `afterDelay` was selected as the auto save mode. These two values are now properly handled.

Note that `onWindowChange` doesn't work well with custom editors. This is due to the usage of `iframe` and the inability to look into an iframe. It still works, but is a bit too eager with them.

#### How to test

Basically, confirm that the auto save functionality behaves as expected with all different values of the enum preference. This should be tested for all 3 kinds of editors (text, notebook, custom editor).

Also test that closing unsaved editors prompts (or automatically saves) the user whether or not to save those files.

#### Follow-ups

We might want to find a way to fix `onWindowChange` for custom editors as well based on mouse position event information.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/save-resource-service.ts
+++ b/packages/core/src/browser/save-resource-service.ts
@@ -14,15 +14,153 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
  ********************************************************************************/
 
-import { inject, injectable } from 'inversify';
-import { MessageService, UNTITLED_SCHEME, URI } from '../common';
+import type { ApplicationShell } from './shell';
+import { injectable } from 'inversify';
+import { UNTITLED_SCHEME, URI, Disposable, DisposableCollection } from '../common';
 import { Navigatable, NavigatableWidget } from './navigatable-types';
-import { Saveable, SaveableSource, SaveOptions } from './saveable';
+import { AutoSaveMode, Saveable, SaveableSource, SaveOptions, SaveReason } from './saveable';
 import { Widget } from './widgets';
+import { FrontendApplicationContribution } from './frontend-application-contribution';
+import { FrontendApplication } from './frontend-application';
+import throttle = require('lodash.throttle');
 
 @injectable()
-export class SaveResourceService {
-    @inject(MessageService) protected readonly messageService: MessageService;
+export class SaveResourceService implements FrontendApplicationContribution {
+
+    protected saveThrottles = new Map<Saveable, AutoSaveThrottle>();
+    protected saveMode: AutoSaveMode = 'off';
+    protected saveDelay = 1000;
+    protected shell: ApplicationShell;
+
+    get autoSave(): AutoSaveMode {
+        return this.saveMode;
+    }
+
+    set autoSave(value: AutoSaveMode) {
+        this.updateAutoSaveMode(value);
+    }
+
+    get autoSaveDelay(): number {
+        return this.saveDelay;
+    }
+
+    set autoSaveDelay(value: number) {
+        this.updateAutoSaveDelay(value);
+    }
+
+    onDidInitializeLayout(app: FrontendApplication): void {
+        this.shell = app.shell;
+        // Register restored editors first
+        for (const widget of this.shell.widgets) {
+            const saveable = Saveable.get(widget);
+            if (saveable) {
+                this.registerSaveable(widget, saveable);
+            }
+        }
+        this.shell.onDidAddWidget(e => {
+            const saveable = Saveable.get(e);
+            if (saveable) {
+                this.registerSaveable(e, saveable);
+            }
+        });
+        this.shell.onDidChangeCurrentWidget(e => {
+            if (this.saveMode === 'onFocusChange') {
+                const widget = e.oldValue;
+                const saveable = Saveable.get(widget);
+                if (saveable && widget && this.shouldAutoSave(widget, saveable)) {
+                    saveable.save({
+                        saveReason: SaveReason.FocusChange
+                    });
+                }
+            }
+        });
+        this.shell.onDidRemoveWidget(e => {
+            const saveable = Saveable.get(e);
+            if (saveable) {
+                this.saveThrottles.get(saveable)?.dispose();
+                this.saveThrottles.delete(saveable);
+            }
+        });
+    }
+
+    protected updateAutoSaveMode(mode: AutoSaveMode): void {
+        this.saveMode = mode;
+        for (const saveThrottle of this.saveThrottles.values()) {
+            saveThrottle.autoSave = mode;
+        }
+        if (mode === 'onFocusChange') {
+            // If the new mode is onFocusChange, we need to save all dirty documents that are not focused
+            const widgets = this.shell.widgets;
+            for (const widget of widgets) {
+                const saveable = Saveable.get(widget);
+                if (saveable && widget !== this.shell.currentWidget && this.shouldAutoSave(widget, saveable)) {
+                    saveable.save({
+                        saveReason: SaveReason.FocusChange
+                    });
+                }
+            }
+        }
+    }
+
+    protected updateAutoSaveDelay(delay: number): void {
+        this.saveDelay = delay;
+        for (const saveThrottle of this.saveThrottles.values()) {
+            saveThrottle.autoSaveDelay = delay;
+        }
+    }
+
+    registerSaveable(widget: Widget, saveable: Saveable): Disposable {
+        const saveThrottle = new AutoSaveThrottle(
+            saveable,
+            () => {
+                if (this.saveMode === 'afterDelay' && this.shouldAutoSave(widget, saveable)) {
+                    saveable.save({
+                        saveReason: SaveReason.AfterDelay
+                    });
+                }
+            },
+            this.addBlurListener(widget, saveable)
+        );
+        saveThrottle.autoSave = this.saveMode;
+        saveThrottle.autoSaveDelay = this.saveDelay;
+        this.saveThrottles.set(saveable, saveThrottle);
+        return saveThrottle;
+    }
+
+    protected addBlurListener(widget: Widget, saveable: Saveable): Disposable {
+        const document = widget.node.ownerDocument;
+        const listener = (() => {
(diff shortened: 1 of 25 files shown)


packages/core/src/browser/saveable.ts
```diff
@@ -209,6 +230,7 @@ export namespace Saveable {
 
     export function apply(
         widget: Widget,
+        isAutoSaveEnabled?: () => boolean,
```
tsmaeder, line 233: If we're trying to centralize the `save` handling, do we still need the whole dance of `Saveable.apply()`, etc.? Why do we retain the option of using a different implementation of `isAutosaveEnabled`?

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -1233,7 +1233,8 @@ export class ApplicationShell extends Widget {
         this.checkActivation(widget);
         Saveable.apply(
             widget,
-            () => this.widgets.filter((maybeSaveable): maybeSaveable is Widget & SaveableSource => !!Saveable.get(maybeSaveable)),
+            () => this.saveResourceService.autoSave !== 'off',
+            () => this.widgets.filter(maybeSaveable => !!Saveable.get(maybeSaveable)),
```
tsmaeder, line 1237: We now have code in  `save-resource-service.ts` to track saveable widgets. Why do we still need this parameter?

packages/core/src/browser/save-resource-service.ts
```diff
@@ -14,15 +14,153 @@
...
+import { AutoSaveMode, Saveable, SaveableSource, SaveOptions, SaveReason } from './saveable';
 import { Widget } from './widgets';
+import { FrontendApplicationContribution } from './frontend-application-contribution';
+import { FrontendApplication } from './frontend-application';
+import throttle = require('lodash.throttle');
 
 @injectable()
-export class SaveResourceService {
-    @inject(MessageService) protected readonly messageService: MessageService;
+export class SaveResourceService implements FrontendApplicationContribution {
```
tsmaeder, line 28: This service saves `Saveable` instances, not resources. If we _do_ have  breaking changes anyway, I would rename it.

packages/core/src/browser/save-resource-service.ts
```diff
@@ -58,3 +196,66 @@ export class SaveResourceService {
         return Promise.reject('Unsupported: The base SaveResourceService does not support saveAs action.');
     }
 }
+
+export class AutoSaveThrottle implements Disposable {
+
+    private _saveable: Saveable;
+    private _cb: () => void;
```
tsmaeder, line 203: `callback` would be immediately more understandable.

packages/monaco/src/browser/monaco-editor-model.ts
```diff
@@ -364,7 +366,7 @@ export class MonacoEditorModel implements IResolvedTextEditorModel, TextEditorDo
     }
 
     save(options?: SaveOptions): Promise<void> {
-        return this.scheduleSave(TextDocumentSaveReason.Manual, undefined, undefined, options);
+        return this.scheduleSave(options?.saveReason ?? TextDocumentSaveReason.Manual, undefined, undefined, options);
```
tsmaeder, line 369: I'm not quite sure why we differentiate the different "save" operations by reason.

packages/notebook/src/browser/view-model/notebook-model.ts
```diff
@@ -205,21 +213,24 @@ export class NotebookModel implements Saveable, Disposable {
...
-        }
+        this.dirty = this.dirtyCells.length > 0;
     }
 
     setData(data: NotebookData): void {
         // Replace all cells in the model
         this.replaceCells(0, this.cells.length, data.cells, false);
         this.metadata = data.metadata;
-        this.dirty = false;
+        this.dirty = true;
```
tsmaeder, line 223: What's behind this change?

packages/filesystem/src/browser/filesystem-save-resource-service.ts
```diff
@@ -25,9 +25,14 @@ import { FileDialogService } from './file-dialog';
 @injectable()
 export class FilesystemSaveResourceService extends SaveResourceService {
```
tsmaeder, line 26: Does our API allow to ever save to anything else but the file system? If not, why to we have this subclass?

tsmaeder (commented):
First reading of the code.

packages/notebook/src/browser/view-model/notebook-model.ts
msujew: It just aligns our behavior with VSCode. VSCode also sets notebooks as dirty if their data is overwritten programmatically.

packages/editor/src/browser/editor-command.ts
```diff
@@ -242,9 +241,15 @@ export class EditorCommandContribution implements CommandContribution {
 
     @postConstruct()
     protected init(): void {
-        this.editorPreferences.onPreferenceChanged(e => {
-            if (e.preferenceName === 'files.autoSave' && e.newValue !== 'off') {
-                this.shell.saveAll();
+        this.preferencesService.ready.then(() => {
```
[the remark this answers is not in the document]
msujew: Could be, but then `SaveResourceService` (from `@theia/core`) has an indirect dependency on `@theia/editor`, since the `files.autoSave` preference is defined in there. Are you in favor of that?

packages/core/src/browser/save-resource-service.ts
```diff
@@ -58,3 +196,66 @@ export class SaveResourceService {
...
+            saveable.onContentChanged(() => {
+                this.throttledSave();
+            }),
+            saveable.onDirtyChanged(() => {
+                this.throttledSave();
+            })
+        );
+    }
+
+    protected throttledSave(): void {
```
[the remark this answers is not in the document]
msujew: > However, autosave is for making sure I don't lose my work

You don't lose your work with `autoSave === on`, as closing the tab/window automatically saves any changes when that's enabled. This is also the convention used by VSCode, since the editor just saving between two keystrokes is bad UX.

packages/filesystem/src/browser/filesystem-save-resource-service.ts
msujew: Since the `SaveResourceService` lives in the `@theia/core` package, it doesn't have access to services such as `FileService`. In general, the `Saveable` API is disk-agnostic, but the `Save As` functionality isn't.

packages/core/src/browser/save-resource-service.ts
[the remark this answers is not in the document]
tsmaeder: The question is what you want to guard against: if you want to guard against the IDE crashing, the dialog will never be shown. If you don't guard against crashes, what is autosave for?

tsmaeder:
I've found a weird little bug:

1. Turn autosave to "off".
2. Open an editor
3. Types something
4. Close the editor
5. Observe: you get a dialog asking you to save the dirty editor

Works as expected.

1. Open the settings, set set autosave to "after delay"
7. Open an editor
8. Type something in the editor
9. Observe: after a delay, the "dirty" indicator in the editor tab disappears
10. Open the "File" menu: there is a checked menu item called "Autosave". Select the item
11. Open the "File" menu again: observe that the "Autosave" item is unchecked
12. Type in the editor.
13. Observe: the "dirty" indicator never goes away.
14. Close the editor
15. Observe: there is no dialog asking me to save and the changes I make are saved to disk automatically. My expectation is that I am asked to save the dirty editor upon closing.

Note that I can have the settings open and observe the "autosave" setting change to "off" when I uncheck the menu item, but I only get the "save"-dialog back once I manually edit the setting in the settings editor.

packages/core/src/browser/save-resource-service.ts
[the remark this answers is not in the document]
msujew: I don't know, why don't you ask the user that set their auto save delay to 2 minutes? ;)

IMO this way of implementing the feature is correct, especially since VSCode and Theia have been doing it the same way. I don't see a reason to deviate from that due to edge cases.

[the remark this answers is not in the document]
tsmaeder: If it's not to guard against crashes, I just don't get what "autosave" is supposed to be for. But I guess it's "works as intended".

packages/filesystem/src/browser/filesystem-save-resource-service.ts
tsmaeder: To me it seems that the partition of responsiblities does not really fit reality, leading to stuff like this and for example the empty "save as" implementation in core. But I do realize this is not what you're trying to do in this PR.

packages/notebook/src/browser/view-model/notebook-model.ts
tsmaeder: Hmh...looking at the call sites, this seems to be called from `revert`, for example. If it's a fix, maybe the fix needs to be at certain call sites, not here? What are the user-level scenarios that invoke this method?

msujew: Probably, yeah, that's right. The caller site that I thought about would be [here](https://github.com/eclipse-theia/theia/blob/44b4f27c63028c1bdc7dcd9b565d666428af6f97/packages/plugin-ext/src/main/browser/notebooks/notebook-documents-main.ts#L165), but that's not the only one. I'll adjust this :)

tsmaeder:
> @tsmaeder Thanks for the repro steps. I believe I ran into a variation of #7480 when using `PreferenceChange#newValue`, which was undefined after the auto save preference was updated. I'm now using the computed value instead, which seems to work well.

From the linked issue : 

> It's just a matter of time until someone thinks: "I'll just use the new value" and introduces an error. 

😎 🤣

tsmaeder (approved):

Issue #13410, opened 2024-02-22
Feature Request: Configurable Splash Screen for Theia Electron


sdirix:
Splash screens are an often implemented feature as they can improve user experience while also offering a branding opportunity.

In case a splash screen is implemented for a Theia based application, it naturally makes sense to show it until the frontend is fully loaded, i.e. until the loading spinner vanishes. For this, framework support is useful.

Suggested requirements:

- The splash screen support is off by default and must be explicitly configured
- Typical splash screen use cases should be configurable via the package.json
- The splash screen should allow content customization similar to the `preloadContent`
- The splash screen should be displayed as early as possible during startup, with the frontend application loading invisible in the background.
- The splash screen should be hidden once the frontend app is fully loaded, revealing the frontend at the same time

The splash screen configuration could also support the following parameters:
- `minDuration: number`: Minimum duration to show the splash screen, in milliseconds. Defaults to `undefined`, indicating this feature is disabled.
- `maxDuration: number`: Maximum duration to show the splash screen, in milliseconds. Defaults to `undefined`, indicating this feature is disabled.
- `width: number`: Width of the splash screen in pixels. Default value `640`.
- `height: number`: Height of the splash screen in pixels. Default value `480`.

msujew:
@kittaakos Is this still interesting for you guys over at Arduino?

tsmaeder:
We now have a splash screen.

PR #13699, opened 2024-05-07
Make browserWindow of splashScreen transparent

sgraband:
#### What it does
Currently an image with a transparent background is rendered with a white background. 
image_480
This looks weird and is not the expected behavior when a transparent image is used. 
With this change the browserWindow is set to transparent, so that the background is not rendered. 
If adoptors want a white background they can simply use an image with a white background (or any other color).

Part of #13410

#### How to test

Start the example electron application and observe that the splash screen image no longer has a white background.

#### Review checklist

#### Reminder for reviewers


sdirix (approved):
Looks good to me!

JonasHelming (approved):

PR #13701, opened 2024-05-08
Stop execution when deleting cell

jonah-iden:
#### What it does

This stops execution of cells when deleting them. Through this the kernel won't break and cells will executable normally after that.

#### How to test

Create or open a notebook with at least 2 cells.
have a longer running cell like:
```python
import time
for i in range(1,100): 
    print('gffff' + str(i))
    time.sleep(0.1)
```

execute the cell and delete it while it executes. After that the other cells should still be executable 

#### Review checklist

#### Reminder for reviewers


msujew (approved):
Great, looks good to me 👍

PR #13708, opened 2024-05-13
fix: do not override user-defined THEIA_CONFIG_DIR

### LINKED ISSUE #13700: Failed to customize the data folder
### Bug Description:
Prior to version 1.44, the ConfigDirUri variable could be customized via process.env.THEIA_CONFIG_DIR. This meant users could set process.env.THEIA_CONFIG_DIR within their application to specify a custom data directory. However, after the introduction of the logic in the app directory in version 1.44, process.env.THEIA_CONFIG_DIR is being re-assigned, which seems to override previous settings. I understand that there should be a priority order where:

1. User-defined settings take precedence,
2. Followed by application-specific settings if defined,
3. Lastly, defaulting to the standard THEIA directory if neither is provided.

Here is the logical code for handling ConfigDirUri, which I understand should not be specified directly if process.env.THEIA_CONFIG_DIR has a value
```
protected async createConfigDirUri(): Promise<string> {
      const dataFolderPath = join(BackendApplicationPath, 'data');
      const userDataPath = join(dataFolderPath, 'user-data');
      const dataFolderExists = this.pathExistenceCache[dataFolderPath] ??= await pathExists(dataFolderPath);
      if (dataFolderExists) {
          const userDataExists = this.pathExistenceCache[userDataPath] ??= await pathExists(userDataPath);
          if (userDataExists) {
              process.env.THEIA_CONFIG_DIR = userDataPath;
          } else {
              await mkdir(userDataPath);
              process.env.THEIA_CONFIG_DIR = userDataPath;
              this.pathExistenceCache[userDataPath] = true;
          }
      } else {
          process.env.THEIA_CONFIG_DIR = join(homedir(), '.theia');
      }
      return FileUri.create(process.env.THEIA_CONFIG_DIR).toString();
  }
```

### Steps to Reproduce:
if process.env.THEIA_CONFIG_DIR is specified as a directory, the latest cache file address is not matched

### Additional Information

- Operating System:
- Theia Version: 1.44 ～1.49

xai:
#### What it does

The current behavior of Theia is to ignore a user-defined `THEIA_CONFIG_DIR` environment variable, which is an unexpected regression from a user's point of view.

This patch ensures that a user-defined value has precedence over the `BackendApplicationPath/data/user-data` location and the default location `~/.theia`.

Fixes #13700

#### How to test

1. Set the `THEIA_CONFIG_DIR` environment variable to a non-default value:
   `export THEIA_CONFIG_DIR=~/theia-test-data-dir` 
2. Start Theia (e.g., `yarn browser start`)
3. Confirm that you see a line similar to the following in stdout, indicating that the value you provided in step 1 is used as configuration directory:
```
Configuration directory URI: 'file:///home/xai/theia-test-data-dir'
```
4. Confirm that this directory has been created on your filesystem (if it has not already existed before)
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/node/env-variables/env-variables-server.ts
+++ b/packages/core/src/node/env-variables/env-variables-server.ts
@@ -46,6 +46,11 @@ export class EnvVariablesServerImpl implements EnvVariablesServer {
     }
 
     protected async createConfigDirUri(): Promise<string> {
+        if (process.env.THEIA_CONFIG_DIR) {
+            // this has been explicitly set by the user, so we do not override its value
+            return FileUri.create(process.env.THEIA_CONFIG_DIR).toString();
+        }
+
         const dataFolderPath = join(BackendApplicationPath, 'data');
         const userDataPath = join(dataFolderPath, 'user-data');
         const dataFolderExists = this.pathExistenceCache[dataFolderPath] ??= await pathExists(dataFolderPath);


msujew (approved):
Looks good to me 👍

PR #13713, opened 2024-05-16
doc: improve WebSocketConnectionProvider deprecation message

eneufeld:
The current deprecation message leads to confusion as the referenced class is not exported and not usable as an injection symbol. The message is improved and the new ServiceConnectionProvider is exported.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/messaging/index.ts
+++ b/packages/core/src/browser/messaging/index.ts
@@ -15,3 +15,4 @@
 // *****************************************************************************
 
 export * from './ws-connection-provider';
+export * from './service-connection-provider';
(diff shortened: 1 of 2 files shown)


JonasHelming:
@tortmayr Why do we actually a different symbol compared to the type?

tortmayr:
I don't know the exact reasoning for that, we would have to ask @tsmaeder.

However, I have now migrated a couple of Theia projects to the latest version and I agree that having a separate type and symbol is quite cumbersome. But I'm not sure if there is an easy way to fix this.

tsmaeder:
Does https://github.com/eclipse-theia/theia/pull/13805 explain why? It's because of remote back ends.

msujew:
To explain it in short: In any Electron app, there can be two instances of a `ServiceConnectionProvider`, one for the local backend and one for the remote one. To differentiate them, we need two symbols.

PR #13715, opened 2024-05-16
notebook: Select the Last Cell when deleting selected last cell

jonah-iden:
#### What it does

This aligns more to vscodes notebook behaviour that when deleting the last cells where one of them is selected, then the last cell in the notebook is selected. Previously no cell was selected after that.

#### How to test

Open a notebook with at least two cells. Select the last cell and delete it. The previous cell should now be selected

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/view-model/notebook-model.ts
+++ b/packages/notebook/src/browser/view-model/notebook-model.ts
@@ -314,7 +314,7 @@ export class NotebookModel implements Saveable, Disposable {
 
             // if selected cell is affected update it because it can potentially have been replaced
             if (cell === this.selectedCell) {
-                this.setSelectedCell(this.cells.length > cellIndex ? this.cells[cellIndex] : this.cells[this.cells.length - 1]);
+                this.setSelectedCell(this.cells[Math.min(cellIndex, this.cells.length - 1)]);
             }
         }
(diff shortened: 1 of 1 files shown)


packages/notebook/src/browser/view-model/notebook-model.ts
```diff
@@ -314,7 +314,7 @@ export class NotebookModel implements Saveable, Disposable {
 
             // if selected cell is affected update it because it can potentially have been replaced
             if (cell === this.selectedCell) {
-                this.setSelectedCell(this.cells[cellIndex]);
+                this.setSelectedCell(this.cells.length > cellIndex ? this.cells[cellIndex] : this.cells[this.cells.length - 1]);
```
msujew, line 317: Suggestion: Maybe something like this?
```suggestion
                this.setSelectedCell(this.cells[Math.min(Math.max(cellIndex, 0), this.cells.length - 1)]);
```

jonah-iden: why force it to be at least 0? accessing an array with index -1 just returns undefined which in this case is expected to happen

msujew (approved):
LGTM 👍

PR #13718, opened 2024-05-16
[vscode] Add WindowState active in plugin API

### LINKED ISSUE #13692: [vscode] Support WindowState active boolean introduced in VS Code 1.89
API Interface WindowState has a new boolean property active:

```
/**
 * Whether the window has been interacted with recently. This will change
 * immediately on activity, or after a short time of user inactivity.
 */
readonly active: boolean;
```

This was implemented as a proposed API in VS Code, but not supported in previous versions of Theia.

rschnekenbu:
#### What it does

Implements the new VS Code API WindowState Active

Fixes #13692 

#### How to test 
 
This can be tested with the following extension:
- src: [window-activestate-sample-0.0.1-src.zip](https://github.com/eclipse-theia/theia/files/15337523/window-activestate-sample-0.0.1-src.zip)
- zip vsix: [window-activestate-sample-0.0.1.zip](https://github.com/eclipse-theia/theia/files/15337526/window-activestate-sample-0.0.1.zip)

This extension will display a notification has soon as window state is changed, focus or active. To test, you can get or lose focus with any combination of idle time to check activity state. User is considered as inactive after 20sec of doing nothing, and this timer will be reset has soon as a mouse button is clicked / key stroke or touch gesture.

#### Follow-ups

None

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/window-activity-tracker.ts
+++ b/packages/plugin-ext/src/main/browser/window-activity-tracker.ts
@@ -1,0 +1,96 @@
+// *****************************************************************************
+// Copyright (C) 2024 STMicroelectronics and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { Disposable, Emitter, Event } from '@theia/core';
+
+const CHECK_INACTIVITY_LIMIT = 30;
+const CHECK_INACTIVITY_INTERVAL = 1000;
+
+const eventListenerOptions: AddEventListenerOptions = {
+    passive: true,
+    capture: true
+};
+export class WindowActivityTracker implements Disposable {
+
+    private inactivityCounter = 0; // number of times inactivity was checked since last reset
+    private readonly inactivityLimit = CHECK_INACTIVITY_LIMIT; // number of inactivity checks done before sending inactive signal
+    private readonly checkInactivityInterval = CHECK_INACTIVITY_INTERVAL; // check interval in milliseconds
+    private interval: NodeJS.Timeout | undefined;
+
+    protected readonly onDidChangeActiveStateEmitter = new Emitter<boolean>();
+    private _activeState: boolean = true;
+
+    constructor(readonly win: Window) {
+        this.initializeListeners(this.win);
+    }
+
+    get onDidChangeActiveState(): Event<boolean> {
+        return this.onDidChangeActiveStateEmitter.event;
+    }
+
+    private set activeState(newState: boolean) {
+        if (this._activeState !== newState) {
+            this._activeState = newState;
+            this.onDidChangeActiveStateEmitter.fire(this._activeState);
+        }
+    }
+
+    private initializeListeners(win: Window): void {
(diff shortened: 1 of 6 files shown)


packages/plugin-ext/src/main/browser/window-activity-tracker.ts
```diff
@@ -0,0 +1,95 @@
...
+const CHECK_INACTIVITY_LIMIT = 2;
+const CHECK_INACTIVITY_INTERVAL = 10000;
+
+const eventListenerOptions: AddEventListenerOptions = {
+    passive: true,
+    capture: true
+};
+export class WindowActivityTracker implements Disposable {
+
+    private inactivityTime = 0; // number of times inactivity was checked since last reset
```
msujew, line 27: Suggestion: Isn't this a counter, not a timespan?

packages/plugin-ext/src/main/browser/window-activity-tracker.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    // Reset inactivity time
+    private resetInactivity = (): void => {
+        this.inactivityTime = 0;
+        if (!this.interval) {
+            // it was not active. Set as active and restart tracking inactivity
+            this.activeState = true;
+            this.startTracking();
+        }
+    };
```
msujew, line 73: Suggestion: I'm not sure I agree with the logic here. While this resets the inactivity counter, it doesn't restart the interval. I.e. in case the user performs an action just before the interval rolls over (and thereby increases the counter by 1), the inactivity will already be registered after 10 seconds, instead of the intended 20 seconds.

I'd personally reduce the interval duration to 1 second and require 20 ticks instead of 2. This reduces the issue above to a timespan of <1s, which is effectively unnoticeable.

msujew (commented):
Works generally pretty well, but I some suggestions on the logic & code, see below.

rschnekenbu:
Thanks for your review, @msujew ! 
Good catch for the logic of the reset. I changed the inactivity check interval to 1sec and extended the number of checks to 30.

packages/plugin-ext/src/main/browser/window-activity-tracker.ts
rschnekenbu: adressed in new commit. Changed accordingly to your suggestion.

packages/plugin-ext/src/main/browser/window-activity-tracker.ts
rschnekenbu: renamed in new commit

packages/plugin-ext/src/main/browser/window-state-main.ts
```diff
@@ -46,14 +47,22 @@ export class WindowStateMain implements WindowMain, Disposable {
         const fireDidBlur = () => this.onFocusChanged(false);
         window.addEventListener('blur', fireDidBlur);
         this.toDispose.push(Disposable.create(() => window.removeEventListener('blur', fireDidBlur)));
+
+       const tracker = new WindowActivityTracker(window);
+       this.toDispose.push(tracker.onDidChangeActiveState(isActive => this.onActiveStateChanged(isActive)));
+       this.toDispose.push(Disposable.create(() => tracker.dispose()));
```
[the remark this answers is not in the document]
rschnekenbu: Creating a disposable to dispose a disposable :face_with_head_bandage:  
Fixed with new commit

packages/plugin-ext/src/main/browser/window-activity-tracker.ts
```diff
@@ -0,0 +1,95 @@
+// Copyright (C) 2022 STMicroelectronics and others.
```
msujew, line 1: Minor:
```suggestion
// *****************************************************************************
// Copyright (C) 2024 STMicroelectronics and others.
```

msujew (approved):
LGTM 👍

packages/plugin-ext/src/main/browser/window-activity-tracker.ts
msujew: @rschnekenbu This is still missing the `// *****...` part.

Issue #13719, opened 2024-05-16
Dynamically registered menus are ignored by the menu renderer


msujew:
### Bug Description:

Discussed in https://github.com/eclipse-theia/theia/discussions/13716.

Registering a menu with a delay prevents that menu from showing up in the main application menu.

The issue is that the main menu is only recreated during startup and when the menu bar visibility changes. It doesn't listen on any other events.

### Steps to Reproduce:

1. Add a delay to the `registerMenus` method of any `MenuContribution`.
2. Notice that the newly registered menu doesn't appear correctly
3. Adjust the `window.menuBarVisibility` setting.
4. The newly registered menu is now visible.

### Additional Information

- Operating System: Windows
- Theia Version: 1.49.0

baoqing1124:
`        setTimeout(() => {
            menus.registerSubmenu(subSubMenuPath, 'Sample sub menu', { order: '2' });
            menus.registerMenuAction(subSubMenuPath, {
                commandId: SampleCommand2.id,
                order: '3'
            });
            const visibility = this.preferenceService.get('window.menuBarVisibility');
            this.preferenceService.updateValue('window.menuBarVisibility', visibility === 'classic' ? 'visible' : 'classic')
        }, 5000);`
@msujew  Thanks for your help, I changed `window.menuBarVisibility` like this and it worked for me

PR #13720, opened 2024-05-17
Support dynamic menu contributions

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/13719

Adds a new `onDidChange` event to the `MenuModelRegistry`. Any changes to the menus (i.e. registering, unregistering, disposal) fire this event now. Adds a listener to this event in the menu factories to ensure that the menus are correctly recreated.

#### How to test

1. Adjust some of the `sample-menu-contribution.ts` file:
```ts
setTimeout(() => {
  const disposable = menus.registerMenuAction(subSubMenuPath, {
    commandId: SampleCommand2.id,
    order: '3'
  });
  setTimeout(() => disposable.dispose(), 8000);
}, 8000);
```
2. After starting the app, open the `Sample Menu > Sample sub menu`
3. It should only show 2 entries
4. After a few seconds the menu should refresh showing an additional entry
5. A few seconds later the entry should disappear again.

#### Follow-ups

I've noticed that the menu flashes when the model is changed (i.e. any open main sub menus are closed). That's something that could maybe be addressed in a separate PR. A smart diffing of the menu bar HTML element would probably be appropriate here.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/api-tests/src/menus.spec.js
+++ b/examples/api-tests/src/menus.spec.js
@@ -44,8 +44,8 @@ describe('Menus', function () {
 
     const container = window.theia.container;
     const shell = container.get(ApplicationShell);
+    /** @type {BrowserMenuBarContribution} */
     const menuBarContribution = container.get(BrowserMenuBarContribution);
-    const menuBar = /** @type {import('@theia/core/lib/browser/menu/browser-menu-plugin').MenuBarWidget} */ (menuBarContribution.menuBar);
     const pluginService = container.get(HostedPluginSupport);
     const menus = container.get(MenuModelRegistry);
     const commands = container.get(CommandRegistry);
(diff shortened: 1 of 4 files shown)


msujew:
cc @kittaakos this is maybe interesting for you.

rschnekenbu:
@msujew, on my machine (Linux), I tried both electron and browser version (also both on original branch and branch rebased on latest master). 
On browser version, the menu disappears when it is refreshing. If I open it again, I have the right number of commands depending on the timeout. I understand the menu should refresh and content is added/remove, but it should stay visible?
On electron version, the menu stays (there is a kind of small flash on the top menu, as it was deselected), but the content is not updated. If I close it and reopen it, then the content is correct.
I can of course run some screen recording to show you what is happening on my machine for both version if you want.

rschnekenbu:
> @rschnekenbu I've mentioned this in the follow-ups of the PR:
> 
> > I've noticed that the menu flashes when the model is changed (i.e. any open main sub menus are closed). That's something that could maybe be addressed in a separate PR. A smart diffing of the menu bar HTML element would probably be appropriate here.
Indeed, that makes sense. This could be indeed part of a second PR. 

Other than these rendering issues, I am OK with this PR. I will approve it.

rschnekenbu (approved):
Looks good with of course the current known limitations :) Thanks @msujew!

PR #13729, opened 2024-05-24
Delegate showing help to the back end process.

### LINKED ISSUE #13727: Bogus help message from Theia Electron Main
### Bug Description:
We're not getting the proper help when invoking the `electron-main.js` script, for example by invoking `TheiaIde.exe` on Windows

### Steps to Reproduce:

1. Install Theia IDE
2. Invoke `TheiaIDE.exe --help` on the command line
3. Observe: You get this help message:
```
TheiaIDE.exe [file]

Positionals:
  file                                                                  [string]

Options:
  --help              Show help                                        [boolean]
  --version           Show version number                              [boolean]
  --electronUserData  The area where the electron main process puts its data
                                                                        [string]
```
My expectation would be that I get the full list of cli options for theia.

### Additional Information

- Operating System:
- Theia Version:

tsmaeder:
#### What it does
Fixes #13727

The approach is to add non-functional cli contributions for any arguments that concern only the electron-main process. Currently, we have a bit of confusion between the `theia` cli tool, which is mostly a development-time tool (`theia build`, etc.) and the actual Theia application executable (`TheiaIde.exe`). It would probably make sense to evolve the architecture in a way that all possible command line options are available to all "customer-facing" cli tools and can be selectively forwarded to the back-end, plugin-process and any other processes making up Theia. 

#### How to test
The closest equivalent to what the `TheiaIDE.exe` does is the "Launch electron back end" launch configuration. To test, you can add parameters after the initial `.` parameter (which is a parameter to Electron). One can also use `npx theia start <...params>` on the command line in the electron or browser example.
In these way, test that the help commands output reasonable advice and that start parameters are properly handled.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -153,6 +153,9 @@
       "frontendElectron": "lib/electron-browser/window/electron-window-module"
     },
     {
+      "backendElectron": "lib/electron-node/cli/electron-backend-cli-module"
+    },
+    {
       "frontend": "lib/browser/keyboard/browser-keyboard-module",
       "frontendElectron": "lib/electron-browser/keyboard/electron-keyboard-module",
       "backendElectron": "lib/electron-node/keyboard/electron-backend-keyboard-module"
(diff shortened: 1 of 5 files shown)


msujew (approved):
Looks good to me 👍 

I'm getting a bit of superfluous output before the `--help` command, but similarly to the server case, there's probably no easy way around this:

```
Showing splash screen
SplashScreen options { content: 'resources/theia-logo.svg', height: 90 }
```

PR #13741, opened 2024-05-29
Fixed focus loss of the notebook editor widget when bluring a cell editor

jonah-iden:
#### What it does

When you had a cell editor focused but executed a command which changed the selection, further commands would not work because the cell editor blurred itself and so the widget itself was not focused anymore.

#### How to test

Best to test is with the shift+enter

1. Open a notebook with at least two code cells.
2. Focus the text of the first cell
3. shift+enter
4. cell should execute and the next cell should be selected (without focusing the text)
5. another shift+enter
6. next cell should be selected or created 

Step 5 was broken because widget focus was lost previous to this fix

#### Review checklist

#### Reminder for reviewers


msujew (approved):
Thanks, LGTM 👍

PR #13742, opened 2024-05-29
Notebooks: Aligned scroll into view behaviour with vscode

jonah-iden:
#### What it does

As in vscode clicking a cell to select it or deleting a cell and changing the selection that way should not move the newly selected cell into view

#### How to test

In a notebook test:
1. clicking on a cell slightly outside of the view. Should not scroll
2. deleting a cell. Should not scroll
3. inserting new cell: Should scroll
4. arrow keys. With and without text focus. Should scroll

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/service/notebook-context-manager.ts
+++ b/packages/notebook/src/browser/service/notebook-context-manager.ts
@@ -88,6 +88,7 @@ export class NotebookContextManager {
         // Cell Selection realted keys
         this.scopedStore.setContext(NOTEBOOK_CELL_FOCUSED, !!widget.model?.selectedCell);
         widget.model?.onDidChangeSelectedCell(e => {
+            this.selectedCellChanged(e.cell);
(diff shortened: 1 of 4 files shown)


msujew (commented):
This doesn't work as expected for me, see:


https://github.com/eclipse-theia/theia/assets/4377073/01bc9b9e-5660-4950-855d-b53ffbd024b0

Clicking on/deleting a cell still scrolls the (potentially non-existing) cell fully into view.

jonah-iden:
seems i forgot two places two disable scrolling. Thanks for finding! Should be fixed now

msujew (approved):
Thanks, looks good to me 👍 

Something worth mentioning (but can be fixed in a separate PR) would be that focusing and then moving the cursor via the arrow keys scrolls inside of the notebook in VSCode. As the cursor needs to be kept in view. This doesn't work in Theia.

PR #13751, opened 2024-06-03
feat: add support for files.insertFinalNewline during formatOnSave

### LINKED ISSUE #13475: Formatter removes final newline on package.json
### Bug Description:
The Theia workspace contains a setting to turn on "format on save" for the workspace. This is problematic, since the formatter does not work as intended: if I edit a package.json file, the formatter will remove the final newline. This leads to problems on UNIX-derived platforms.

### Steps to Reproduce:

1. Edit a package.json in the Theia source
2. Save
3. Observe: the formatter removes the final newline.

### Additional Information

- Operating System: Windows
- Theia Version: 1.47.0

xai:
#### What it does

When the `files.insertFinalNewline` property is set to true, formatOnSave() will ensure that the file ends with a newline character instead of always removing a final empty line.

Fixes #13475 

#### How to test

1. Open a workspace and set `"editor.formatOnSave"` to `true` in the workspace settings.
2. Edit a file that has a final empty line and save.
3. Verify that the empty line has been removed.
4. Set `"files.insertFinalNewline"` to `true` in the workspace settings.
5. Verify that final empty lines are kept or added when saving.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/browser/filesystem-preferences.ts
+++ b/packages/filesystem/src/browser/filesystem-preferences.ts
@@ -95,6 +95,12 @@ export const filesystemPreferenceSchema: PreferenceSchema = {
             description: nls.localizeByDefault('When enabled, will trim trailing whitespace when saving a file.'),
             scope: 'language-overridable'
         },
+        'files.insertFinalNewline': {
+            type: 'boolean',
+            default: false,
+            description: nls.localizeByDefault('When enabled, insert a final new line at the end of the file when saving it.'),
+            scope: 'language-overridable'
+        },
         'files.maxConcurrentUploads': {
             type: 'integer',
             default: 1,
@@ -116,6 +122,7 @@ export interface FileSystemConfiguration {
     'files.participants.timeout': number
     'files.maxFileSizeMB': number
     'files.trimTrailingWhitespace': boolean
+    'files.insertFinalNewline': boolean
     'files.maxConcurrentUploads': number
 }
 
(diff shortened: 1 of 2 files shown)


JonasHelming:
@xai What is the state of this?

xai:
> @xai What is the state of this?

As I commented in the related issue, I feel that the added protected function might be living in the wrong place. But it works, and I do not know what a better file/package would be for that function, so I am just going to remove the draft state now and hope for input from a reviewer.

xai:
@tsmaeder alerted me that the property seems to be ignored in certain cases:
- It works as intended without the built-in extensions (i.e., omit `yarn download:plugins` while building or clear plugins/ directory).
- With all built-in extensions, the behavior of an extension seems to override the new files.insertFinalNewline setting and also the format-on-save setting.

Merging as is does not make sense, because it will not work in most cases (i.e., all built-in extensions are present). So, I am going to debug into this to find out which extension we are having this feature interaction with.

xai:
FYI: I am retesting this right now ...

xai:
Just rebased against master, above [comment](https://github.com/eclipse-theia/theia/pull/13751#issuecomment-2443928883) still applies.

xai:
So, what is the expected behavior if a generic Theia workspace setting such as `files.insertFinalNewline` and extension settings (e.g., in editorconfig, or any language-specific vscode extension) mismatch? From a user perspective, I would probably assume that the more specific setting (i.e., typically the extension) takes precedence. Any opinion, @tsmaeder?

tsmaeder (approved):

PR #13757, opened 2024-06-04
Fix editors theme change and widget not attached error

jonah-iden:
#### What it does

There was a problem when closing an editor to the right of the current or the most right one and then changing the theme, the editors would not update correctly. Was probably introduced by the secondary window changes. This should fix it.

Also fixes the small issue of `widget not attached` errors from the webviews when closing a notebook editor with outputs

#### How to test

Theming:
1. open at least two editors, the right one should be a standalone one. Left can be either notebook or standalone.
2. close the right editor.
3. change the theme to the opposite theme (dark/light)
4. see that it now updates correctly (previously the editor would stay on the previous color)

Widget not attached:
1. just open a notebook with outputs
2. close it
3. there should be no widget not attached errors in the console

#### Review checklist

#### Reminder for reviewers


msujew (approved):
Thanks, looks good to me 👍

PR #13768, opened 2024-06-05
Avoid changing context key view when changing selection

### LINKED ISSUE #13375: NPM Support builtin extension: Issue in the Refresh toolbar command
### Bug Description:
Hi,

There is an issue in the builtin vscode NPM support extension, where the refresh command could be added to other tree components like the file explorer and the extensions list.

### Steps to Reproduce:

1. have a project with some package.json files
2. open the npm view
3. expand a package.json item to see its scripts
4. collapse it back
5. you will see that a new refresh icon is added to the navigator tree, also to the whole explorer view, and to the extensions view...
6. if you do inspect on this icon, you will see that it comes from the npm.refresh command

### Additional Information

- Operating System: Mac
- Theia Version: 1.45.1

rschnekenbu:
#### What it does

Closes #13375

In a tree view provided by an extension, refreshing the tree node selection was changing the context key view. This was changing the view in the root context. So when the actions in toolbar were refreshed, the "when" rule was always returning `true` for any view, not only the concerned one.
Avoid changing the context key view should not be done on selection refresh in this case. 

#### How to test

1. Use the procedure described in 13375
2. Prior to the patch, when selecting elements in the NPM scripts view, the refresh icon was showing on every view.
3. With the patch, the refresh action does not show on every view.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/view/tree-views-main.ts
+++ b/packages/plugin-ext/src/main/browser/view/tree-views-main.ts
@@ -24,7 +24,6 @@ import {
     CompositeTreeNode,
     WidgetManager
 } from '@theia/core/lib/browser';
-import { ViewContextKeyService } from './view-context-key-service';
 import { Disposable, DisposableCollection } from '@theia/core';
 import { TreeViewWidget, TreeViewNode, PluginTreeModel, TreeViewWidgetOptions } from './tree-view-widget';
 import { PluginViewWidget } from './plugin-view-widget';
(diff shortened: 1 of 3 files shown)


packages/plugin-ext/src/main/browser/view/tree-views-main.ts
```diff
@@ -197,7 +194,6 @@ export class TreeViewsMainImpl implements TreeViewsMain, Disposable {
         }));
 
         this.toDispose.push(treeViewWidget.model.onSelectionChanged(event => {
-            this.contextKeys.view.set(treeViewId);
```
tsmaeder, line 200: Would it make sense to remove the `view` context key? It never makes sense to set it in the global context, as I understand it.

rschnekenbu: I removed this context key, as it indeed makes no sense on a global context. _focusedView_ makes sense here, but not _view_.

rschnekenbu:
Pushed a new version that removes view from the view context key, as it makes no sense in a global context. 
It solves also usual conflicts on changelog file.

tsmaeder (approved):

PR #13773, opened 2024-06-05
Added notebook output options and tag preference search

jonah-iden:
#### What it does

This adds multiple new notebook output preferences.
Also adds the ability to filter by tags in the settings so that the settings link in truncated notebook outputs work.

#### How to test

Open or create a notebook with a cell with large output like 
```python
for i in range(1,100):
    print('a')
```
after running the cell at the bottom you should see a `settings` link. Clicking it should open the settings and show the new preferences. Playing around with them (except see follow ups) should update the output.

#### Follow-ups

the following preferences are currently not live updated but require clearing the output and running a cell again:
- `outputLineLimit`
- `outputScrolling`
- `outputWordWrap`

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/service/notebook-options.ts
+++ b/packages/notebook/src/browser/service/notebook-options.ts
@@ -1,0 +1,123 @@
+
+// *****************************************************************************
+// Copyright (C) 2024 TypeFox and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { inject, injectable, postConstruct } from '@theia/core/shared/inversify';
+import { PreferenceService } from '@theia/core/lib/browser';
+import { Emitter } from '@theia/core';
+import { NotebookPreferences } from '../contributions/notebook-preferences';
+import { EditorPreferences } from '@theia/editor/lib/browser';
+
+const notebookOutputOptionsRelevantPreferences = [
+    'editor.fontSize',
+    'editor.fontFamily',
+    NotebookPreferences.NOTEBOOK_LINE_NUMBERS,
+    NotebookPreferences.OUTPUT_LINE_HEIGHT,
+    NotebookPreferences.OUTPUT_FONT_SIZE,
+    NotebookPreferences.OUTPUT_FONT_FAMILY,
+    NotebookPreferences.OUTPUT_SCROLLING,
+    NotebookPreferences.OUTPUT_WORD_WRAP,
+    NotebookPreferences.OUTPUT_LINE_LIMIT
+];
+
+export interface NotebookOutputOptions {
+    // readonly outputNodePadding: number;
+    // readonly outputNodeLeftPadding: number;
+    // readonly previewNodePadding: number;
+    // readonly markdownLeftMargin: number;
+    // readonly leftMargin: number;
+    // readonly rightMargin: number;
+    // readonly runGutter: number;
+    // readonly dragAndDropEnabled: boolean;
+    readonly fontSize: number;
+    readonly outputFontSize?: number;
+    readonly fontFamily: string;
+    readonly outputFontFamily?: string;
+    // readonly markupFontSize: number;
+    // readonly markdownLineHeight: number;
+    readonly outputLineHeight: number;
+    readonly outputScrolling: boolean;
+    readonly outputWordWrap: boolean;
+    readonly outputLineLimit: number;
+    // readonly outputLinkifyFilePaths: boolean;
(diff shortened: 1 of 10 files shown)


packages/notebook/src/browser/service/notebook-options.ts
```diff
@@ -0,0 +1,123 @@
...
+        return {
+            fontSize,
+            outputFontSize: this.preferenceService.get<number>(NotebookPreferences.OUTPUT_FONT_SIZE),
+            fontFamily: this.preferenceService.get<string>('editor.fontFamily')!,
+            outputFontFamily: this.preferenceService.get<string>(NotebookPreferences.OUTPUT_FONT_FAMILY),
+            outputLineHeight: this.computeOutputLineHeight(outputLineHeight, outputFontSize ?? fontSize),
+            outputScrolling: this.preferenceService.get<boolean>(NotebookPreferences.OUTPUT_SCROLLING)!,
+            outputWordWrap: this.preferenceService.get<boolean>(NotebookPreferences.OUTPUT_WORD_WRAP)!,
+            outputLineLimit: this.preferenceService.get<number>(NotebookPreferences.OUTPUT_LINE_LIMIT)!
+        };
```
msujew, line 96: Suggestion: Using the nullable coercion (`!`) operator for the `preferenceService#get` call is not ideal. Use the second parameter of `get` to specify the default value in case the preference value is invalid.

packages/plugin-ext/src/main/browser/notebooks/renderers/webview-communication.ts
```diff
@@ -59,7 +59,20 @@ export interface PreloadMessage {
     readonly resources: string[];
 }
 
-export type ToWebviewMessage = UpdateRenderersMessage | OutputChangedMessage | ChangePreferredMimetypeMessage | CustomRendererMessage | KernelMessage | PreloadMessage;
+export interface notebookStylesMessage {
+    readonly type: 'notebookStyles';
+    styles: {
+        [key: string]: string;
+    }
```
msujew, line 66: Minor:
```suggestion
    readonly styles: Record<string, string>
```

packages/notebook/src/browser/service/notebook-options.ts
```diff
@@ -0,0 +1,123 @@
...
+    protected readonly preferenceService: PreferenceService;
+
+    @inject(EditorPreferences)
+    protected readonly editorPreferences: EditorPreferences;
+
+    protected outputOptionsChangedEmitter = new Emitter<NotebookOutputOptions>();
+    onDidChangeOutputOptions = this.outputOptionsChangedEmitter.event;
+
+    @postConstruct()
+    init(): void {
```
msujew, line 73: Minor:
```suggestion
    protected init(): void {
```

packages/notebook/src/browser/service/notebook-options.ts
```diff
@@ -0,0 +1,123 @@
...
+            fontFamily: this.preferenceService.get<string>('editor.fontFamily')!,
+            outputFontFamily: this.preferenceService.get<string>(NotebookPreferences.OUTPUT_FONT_FAMILY),
+            outputLineHeight: this.computeOutputLineHeight(outputLineHeight, outputFontSize ?? fontSize),
+            outputScrolling: this.preferenceService.get<boolean>(NotebookPreferences.OUTPUT_SCROLLING)!,
+            outputWordWrap: this.preferenceService.get<boolean>(NotebookPreferences.OUTPUT_WORD_WRAP)!,
+            outputLineLimit: this.preferenceService.get<number>(NotebookPreferences.OUTPUT_LINE_LIMIT)!
+        };
+    }
+
+    private computeOutputLineHeight(lineHeight: number, outputFontSize: number): number {
```
msujew, line 99: Minor:
```suggestion
    protected computeOutputLineHeight(lineHeight: number, outputFontSize: number): number {
```

msujew (commented):
With the tag search in place, the search has become very restricted. When placing a tag in the preference search bar, searching now **only** works for that tag. For example, in vscode, I can search for `@tag:notebookLayout cell` and get all layout related settings for cells. In this PR, it doesn't show anything, since the search attempts to look for a `notebookLayout cell` tag, which doesn't exist.

As far as I can tell, tags act as filters, not as search parameters. I.e. tags should only be used to restrict the search results. Meaning that the search content should still be respected and only filtered by any existing tags.

packages/notebook/src/browser/service/notebook-options.ts
jonah-iden: But these specific settings with nullable coercion allready have default values. So they should be always set. Should i rather write this as `this.preferenceService.get<boolean>(NotebookPreferences.OUTPUT_WORD_WRAP, notebookPreferenceSchema.properties[NotebookPreferences.OUTPUT_FONT_FAMILY].default as string)`?

msujew: One could argue that adopters could remove some of these preferences, thereby deleting the default value. I find this scenario to be highly unlikely though. I think we can keep it as it is for now.

jonah-iden: ok should be a little better now

msujew: Yes, perfect, thanks!

jonah-iden:
seartch should be improved now to work very simliar to vscode

msujew (commented):
With the default settings, the text output looks pretty squished:

image

Can you reproduce? I haven't changed any of the output layout settings.

jonah-iden:
@msujew thanks, good catch. should be fixed now

msujew (approved):
Looks great, thanks 👍

Issue #13777, opened 2024-06-06
[vscode] Implementation of the Chat and Language Model APIs


rschnekenbu:
As these interfaces have been stubbed with #13756, this issue tracks the implementation of these APIs.

chbo123456:
Hi, any news on this issue? Will this be planned in the near future? Would be awesome to have an implemention. =)

rschnekenbu:
Hi @chbo123456, we are currently in the process of implementing AI within Theia (see https://github.com/eclipse-theia/theia/issues/14119 and https://github.com/eclipse-theia/theia/pull/14048). 
Once we have done good progress on this integration, we will improve the support of the VS Code API.

chbo123456:
Thanks @rschnekenbu , I checked out the new theiaai features, quite awesome! Does this mean that don't really need this plugin anymore, if I want to use the  "Codex" Model (aka github copilot) or did I missunderstand the new abilities and "only" standard openai/ollama models are usable here?

In the end, I don't really need the vscode API, as long as I could make use of github copilot, that's why I'm asking.

msujew:
> In the end, I don't really need the vscode API, as long as I could make use of github copilot, that's why I'm asking.

GitHub Copilot isn't really supported by Theia. The only way to download the extension is via the [official VS Code marketplace](https://marketplace.visualstudio.com/items?itemName=GitHub.copilot), which prohibits installation of any downloaded extension into any app that isn't officially distributed by Microsoft (i.e. Desktop VS Code or GitHub CodeSpaces). Even though it might be technically possible to support the extension in Theia, we won't do so due to the Terms of Service/licensing issues mentioned above.

> Does this mean that I don't really need e.g. the official github copilot plugin anymore, if I want to use the "Codex" Model (aka github copilot) or did I missunderstand the new abilities and "only" standard openai/ollama models are usable here?

The Theia AI integration doesn't limit you to any specific models - we support any model provider that supports either the OpenAI API (i.e. VLLM) or Ollama. Those can be hosted by OpenAI themselves, or any other party that offers a OpenAI compatible API (so essentially everyone, you can even spin this up locally).

AFAIK this issue here in particular is about contributing the VS Code API to support extensions that use the [language model API](https://code.visualstudio.com/api/extension-guides/language-model).

PR #11817, opened 2022-10-28
plugin-ext: cleanup `theia.d.ts` for `extensions`

vince-fugnitto:
#### What it does

Related: #11815 

The commit cleans up the `theia.d.ts` with the following changes:
- moves `plugin` namespace to `theia-extra.d.ts` since it is different than `vscode.d.ts`.
- adds `extension` namespace to `theia.d.ts` for our compatibility report.

#### How to test

The changes should work like before, since `extensions` was already supported but the code now organizes it in the proper location, and the report now passes for the API:

<details>

<summary>Compatibility Report</summary>

<br />

image

</details>

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

Signed-off-by: vince-fugnitto <vincent.fugnitto@ericsson.com>

PR #13778, opened 2024-06-06
[vscode] Stub Chat and Language Model API

### LINKED ISSUE #13756: [vscode] Implement Chat and Language Model initial APIs
VS Code 1.90 introduced the extension support for Chat (ChatParticipant) and Language Models 
- see https://code.visualstudio.com/api/extension-guides/language-model
- see https://code.visualstudio.com/api/extension-guides/chat
- see https://github.com/microsoft/vscode/blob/5d6671dacb9d6a582b9354ea317211a8e2b2f918/src/vscode-dts/vscode.d.ts#L19190
- see https://github.com/microsoft/vscode/blob/5d6671dacb9d6a582b9354ea317211a8e2b2f918/src/vscode-dts/vscode.d.ts#L18580

rschnekenbu:
#### What it does

Stub the API for chat and Language Models introduced in VS Code 1.90. 
Closes #13756

Status of the report when comparing with vscode 1.90: [status.zip](https://github.com/user-attachments/files/15650003/status.zip)

#### How to test

1. Start the theia browser example on master and install the following extension (derived from vscode sample extension chat-sample):
- src: [chat-sample-0.0.2-src.zip](https://github.com/user-attachments/files/15649765/chat-sample-0.0.2-src.zip)
- zip vsix: [chat-sample-0.2.0.zip](https://github.com/user-attachments/files/15649798/chat-sample-0.2.0.zip)

2. On extension activation, some error messages are displayed in the console.
3. With the PR, the API does not show any error on extension startup. extension context and calls to chat/lm namespaces do not throw any error.

#### Follow-ups
Real implementation of the API is tracked here: #13777 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/plugin-context.ts
+++ b/packages/plugin-ext/src/plugin/plugin-context.ts
@@ -19,7 +19,7 @@
 
 import type * as theia from '@theia/plugin';
 import { CommandRegistryImpl } from './command-registry';
-import { Emitter } from '@theia/core/lib/common/event';
+import { Emitter, Event } from '@theia/core/lib/common/event';
 import { CancellationError, CancellationToken, CancellationTokenSource } from '@theia/core/lib/common/cancellation';
 import { QuickOpenExtImpl } from './quick-open';
 import {
@@ -211,7 +211,19 @@ import {
     DeclarationCoverage,
     FileCoverage,
     StatementCoverage,
-    TestCoverageCount
+    TestCoverageCount,
+    ChatRequestTurn,
+    ChatResponseTurn,
+    ChatResponseAnchorPart,
+    ChatResponseCommandButtonPart,
+    ChatResponseFileTreePart,
+    ChatResponseMarkdownPart,
+    ChatResponseProgressPart,
+    ChatResponseReferencePart,
+    ChatResultFeedbackKind,
+    LanguageModelChatMessage,
+    LanguageModelChatMessageRole,
+    LanguageModelError
 } from './types-impl';
 import { AuthenticationExtImpl } from './authentication-ext';
 import { SymbolKind } from '../common/plugin-api-rpc-model';
@@ -1223,7 +1235,25 @@ export function createAPIFactory(
             /** @stubbed MappedEditsProvider */
             registerMappedEditsProvider(documentSelector: theia.DocumentSelector, provider: theia.MappedEditsProvider): Disposable {
                 return Disposable.NULL;
+            },
+            /** @stubbed ChatRequestHandler */
+            createChatParticipant(id: string, handler: theia.ChatRequestHandler): theia.ChatParticipant {
+                return {
+                    id,
+                    requestHandler: (request: theia.ChatRequest, context: theia.ChatContext, response: theia.ChatResponseStream, token: CancellationToken) => { },
+                    dispose() {},
+                    onDidReceiveFeedback: (listener, thisArgs?, disposables?) => Event.None(listener, thisArgs, disposables)
+                };
             }
+        };
+
+        const lm: typeof theia.lm = {
+            /** @stubbed LanguageModelChat */
+            selectChatModels(selector?: theia.LanguageModelChatSelector): Thenable<theia.LanguageModelChat[]> {
+                return Promise.resolve([]);
+            },
+            /** @stubbed LanguageModelChat */
+            onDidChangeChatModels: (listener, thisArgs?, disposables?) => Event.None(listener, thisArgs, disposables)
         };
 
         return <typeof theia>{
@@ -1244,6 +1274,7 @@ export function createAPIFactory(
             notebooks,
             l10n,
             tests,
+            lm,
             // Types
             StatusBarAlignment: StatusBarAlignment,
             Disposable: Disposable,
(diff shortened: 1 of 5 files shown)


packages/plugin-ext/src/plugin/plugin-context.ts
```diff
@@ -1223,9 +1235,27 @@ export function createAPIFactory(
             /** @stubbed MappedEditsProvider */
             registerMappedEditsProvider(documentSelector: theia.DocumentSelector, provider: theia.MappedEditsProvider): Disposable {
                 return Disposable.NULL;
+            },
+            /** @stubbed ChatRequestHandler */
+            createChatParticipant(id: string, handler: theia.ChatRequestHandler): theia.ChatParticipant {
+                return {
+                    id,
+                    requestHandler: (request: theia.ChatRequest, context: theia.ChatContext, response: theia.ChatResponseStream, token: CancellationToken) => { },
```
tsmaeder, line 1243: Why not just assign the passed-in request handler instead of a dummy?

packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -3846,3 +3846,99 @@ export class TerminalQuickFixOpener {
...
+        this.participant = participant;
+        this.references = references;
+    };
+}
+
+export class ChatResponseTurn {
+    readonly command?: string;
+
+    private constructor(public readonly response: ReadonlyArray<theia.ChatResponseMarkdownPart | theia.ChatResponseFileTreePart | theia.ChatResponseAnchorPart
+        | theia.ChatResponseCommandButtonPart>, public readonly result: theia.ChatResult, public readonly participant: string) { }
```
tsmaeder, line 3867: `public` is not necessary, IMO. Goes for all similar occurrences.

packages/plugin-ext/src/plugin/types-impl.ts
```diff
@@ -3846,3 +3846,99 @@ export class TerminalQuickFixOpener {
...
+        return new LanguageModelChatMessage(LanguageModelChatMessageRole.Assistant, content, name);
+    }
+
+    constructor(public role: LanguageModelChatMessageRole, public content: string, public name?: string) { }
+}
+
+export class LanguageModelError extends Error {
+
+    static NoPermissions(message?: string): LanguageModelError {
+        return new LanguageModelError(message, LanguageModelError.NotFound.name);
```
tsmaeder, line 3925: This seems strange when reading (https://code.visualstudio.com/api/references/vscode-api#LanguageModelError). What should the `Error.name` and `code` properties be when we construct a `LanguageModelError` with the `NoPermissions` static method?` Right now, it would be `code === NotFound` and `name === `LanguageModelError`, right? What am I missing?

tsmaeder (commented):
A couple of comments.

packages/plugin/src/theia-extra.d.ts
```diff
@@ -291,6 +291,13 @@ export module '@theia/plugin' {
          * see - workspace.fs for how to read and write files and folders from an uri.
          */
         readonly logUri: Uri;
+
+        /**
+         * An object that keeps information about how this extension can use language models.
+         *
+         * @see {@link LanguageModelChat.sendRequest}
+         */
+        readonly languageModelAccessInformation: LanguageModelAccessInformation;
```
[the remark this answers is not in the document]
rschnekenbu: All the code in ExtensionContext (theia.d.ts) and PluginContext (theia-extra.d.ts) is duplicated. The code in ExtensionContext is there for the API report and that theia.d.ts is as similar as possible to vscode version.The PluginContext is the implementation used by Theia. See https://github.com/eclipse-theia/theia/pull/11817

packages/plugin-ext/src/plugin/types-impl.ts
rschnekenbu: I fixed this part. The LanguageModelError NoPermissions/NotFound/Blocked should now have the correct code.

BTW, due to Typescript version used, the 'cause' property used in VS code in the [constructor](https://github.com/microsoft/vscode/blob/ca8cb6f637fc5ac820b581eb4115bc3917ba7548/src/vs/workbench/api/common/extHostTypes.ts#L4568) is not available yet (es2022). Is there a way to track these issues, as we do for the monaco uplift annotations? should we introduce a typescript-uplift annotation?

packages/plugin-ext/src/plugin/types-impl.ts
rschnekenbu: This should be addressed in the new commit.

packages/plugin-ext/src/plugin/plugin-context.ts
rschnekenbu: Done in new commit

packages/plugin-ext/src/plugin/types-impl.ts
tsmaeder: I doubt this is an issue of the typescript version: we are using 5.4 at the moment, which is fairly new. I figure the problem is with the compilation target, right?

rschnekenbu: yes, the problem is the target

tsmaeder (approved):

PR #13782, opened 2024-06-07
Block local navigation and open new windows externally in electron

### LINKED ISSUE #13592: Open external links in user browser for Theia electron app
Since https://github.com/eclipse-theia/theia/pull/11048, the window open handler for electron application is set to be always a secondary window. 
For all external links, we have to rely on the opener-service to delegate properly to an external application with the 'http'  open handler (https://github.com/eclipse-theia/theia/blob/master/packages/core/src/browser/http-open-handler.ts).

I would rather expect the Theia electron to open any external links by default with an external browser, and only manually code the opening of a secondary window when required. 

This may however cause some issues with security (https://www.electronjs.org/docs/latest/tutorial/security#13-disable-or-limit-navigation)

tsmaeder:
#### What it does

Fixes #13592

Allow to open https/http links externally and ask the user for all other protocols.

#### How to test

I changed the `renderHeader` method in `about-dialog.tsx` to return a couple of naked `<a href=...` links:

`` 
        return <>
            <h3>{detailsLabel}</h3>
            <div className='about-details'>
                {applicationInfo && <p>{`${versionLabel}: ${applicationInfo.version}`}</p>}
                <p>{`${defaultApiLabel}: ${DEFAULT_SUPPORTED_API_VERSION}`}</p>
                <p>
                    <a href={ compatibilityUrl }>
                    { compatibilityLabel }
                    </a>
                </p>
                <p>
                    <a href="https://www.ethz.ch" target='_blank'>
                    ETH Zürich
                    </a>
                </p>
                <p>
                    <a href= "mailto:t.s.maeder@gmail.com" target='_blank'>
                    Send Mail
                    </a>
                </p>
            </div>
        </>;
```

The first one is a local navigation and is blocked. The second is a https link and is opened in the system browser. The third one provokes a dialog asking the user whether to open the link externally.


#### Follow-ups

<!-- Please list potential follow-up work, including known issues, possible future work, identified technical debt, and potentially introduced technical debt. If the PR introduces technical debt, specify the reason why this is acceptable. Please create tickets and link them here. Please use the label "technical debt" for new issues when it applies. -->

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-main/electron-main-application.ts
+++ b/packages/core/src/electron-main/electron-main-application.ts
@@ -15,7 +15,8 @@
 // *****************************************************************************
 
 import { inject, injectable, named } from 'inversify';
-import { screen, app, BrowserWindow, WebContents, Event as ElectronEvent, BrowserWindowConstructorOptions, nativeImage, nativeTheme } from '../../electron-shared/electron';
+import { screen, app, BrowserWindow, WebContents, Event as ElectronEvent, BrowserWindowConstructorOptions, nativeImage,
+    nativeTheme, shell, dialog } from '../../electron-shared/electron';
 import * as path from 'path';
 import { Argv } from 'yargs';
 import { AddressInfo } from 'net';
@@ -31,7 +32,7 @@ import { ContributionProvider } from '../common/contribution-provider';
 import { ElectronSecurityTokenService } from './electron-security-token-service';
 import { ElectronSecurityToken } from '../electron-common/electron-token';
 import Storage = require('electron-store');
-import { CancellationTokenSource, Disposable, DisposableCollection, isOSX, isWindows } from '../common';
+import { CancellationTokenSource, Disposable, DisposableCollection, Path, isOSX, isWindows, nls } from '../common';
 import { DEFAULT_WINDOW_HASH, WindowSearchParams } from '../common/window';
 import { TheiaBrowserWindowOptions, TheiaElectronWindow, TheiaElectronWindowFactory } from './theia-electron-window';
 import { ElectronMainApplicationGlobals } from './electron-main-constants';
@@ -144,6 +145,9 @@ export namespace ElectronMainProcessArgv {
         };
     }
 }
+
+const secondaryWindowHTML = FileUri.create(path.resolve('./lib/frontend/secondary-window.html')).toString();
+console.error(`secondary url: ${secondaryWindowHTML}`);
 
 @injectable()
 export class ElectronMainApplication {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/core/src/electron-main/electron-main-application.ts [hunk 1]
```diff
@@ -145,6 +146,9 @@ export namespace ElectronMainProcessArgv {
     }
 }
 
+const secondaryWindowHTML = FileUri.create(path.resolve('./lib/frontend/secondary-window.html')).toString();
+console.error(`secondary url: ${secondaryWindowHTML}`);
```
msujew, line 150: Leftover from testing?

packages/core/src/electron-main/electron-main-application.ts
```diff
@@ -731,6 +710,60 @@ export class ElectronMainApplication {
...
+                return {
+                    action: 'allow',
+                    overrideBrowserWindowOptions: options,
+                };
+            } else {
+                const uri: URI = new URI(details.url);
+                let okToOpen = uri.scheme === 'https' || uri.scheme === 'http';
+                if (!okToOpen) {
+                    const button = dialog.showMessageBoxSync(BrowserWindow.fromWebContents(webContents)!, {
+                        message: nls.localize('theia/core/openLink.message', 'Open link\n\n{0}\n\nin the system handler?', details.url),
```
msujew, line 749: Note: Only the frontend actually loads the localizations. The Electron main app isn't really capable of translating anything (yet). I should maybe look into how we could accomplish this. In the meantime, everything will be displayed in English.

msujew (commented):
Looks pretty good, just a minor issue, see below.

packages/core/src/electron-main/electron-main-application.ts
tsmaeder: Hmh...that sucks. An alternative would be to delegate showing the dialog to the front end, but that would require some back-and-forth again. Hmh....

msujew: It's fine for me, I just wanted to mention it, since I was a bit confused by the import of `nls` in the electron main thread.

packages/core/src/electron-main/electron-main-application.ts
tsmaeder: Yes.

packages/core/src/electron-main/electron-main-application.ts
tsmaeder: Come to think of it: why is `nls` in the `common` folder if it's only available on the browser side?

msujew: See https://github.com/eclipse-theia/theia/pull/10153#issuecomment-927616277 :)

msujew: It's mainly due to stuff like this:

https://github.com/eclipse-theia/theia/blob/d79b7101adef9db71e74c13744a3fde4f79a343c/packages/core/src/common/command.ts#L62-L80

The code is in `common` (and is actually only called from `browser`) but requires the nls localization for it to work. We should refactor this sometime.

msujew (approved):
LGTM 👍

PR #13791, opened 2024-06-10
Improved notebook cell drag images

jonah-iden:
#### What it does

This improves the rendering of drag images for notebook cells and aligns to more how they look in vscode.

#### How to test

Open notebook and try dragging a cell. 
Code and mardown cells have different drag image renderers

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/view/notebook-cell-list-view.tsx
+++ b/packages/notebook/src/browser/view/notebook-cell-list-view.tsx
@@ -96,6 +96,7 @@ export class NotebookCellListView extends React.Component<CellListProps, Noteboo
                                 this.props.notebookModel.setSelectedCell(cell, false);
                             }}
                             onDragStart={e => this.onDragStart(e, index, cell)}
+                            onDragEnd={e => NotebookCellListView.dragGhost?.remove()}
                             onDragOver={e => this.onDragOver(e, cell)}
                             onDrop={e => this.onDrop(e, index)}
                             draggable={true}
@@ -137,9 +138,6 @@ export class NotebookCellListView extends React.Component<CellListProps, Noteboo
             return;
         }
 
-        if (NotebookCellListView.dragGhost) {
-            NotebookCellListView.dragGhost.remove();
-        }
         NotebookCellListView.dragGhost = document.createElement('div');
         NotebookCellListView.dragGhost.classList.add('theia-notebook-drag-ghost-image');
         NotebookCellListView.dragGhost.appendChild(this.props.renderers.get(cell.cellKind)?.renderDragImage(cell) ?? document.createElement('div'));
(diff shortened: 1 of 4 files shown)


packages/notebook/src/browser/view/notebook-code-cell-view.tsx
```diff
@@ -102,7 +107,20 @@ export class NotebookCodeCellRenderer implements CellRenderer {
...
+        dragImage.style.height = '100px';
+        dragImage.style.display = 'flex';
+
+        const fakeRunButton = document.createElement('span');
+        fakeRunButton.className = `${codicon('play')} theia-notebook-cell-status-item`;
+        dragImage.appendChild(fakeRunButton);
+
+        const fakeEditor = document.createElement('div');
+        dragImage.appendChild(fakeEditor);
+        const firstLine = new MarkdownString(`\`\`\`${cell.language}\n${cell.source.split('\n')[0]}\n\`\`\``, { supportHtml: true, isTrusted: false });
```
msujew, line 120: Suggestion: Using 100 px makes this perfect to show the first 5 lines. IMO this looks way better:
image
```suggestion
        const lines = cell.source.split('\n').slice(0, 5).join('\n');
        const markdownCode = new MarkdownString(`\`\`\`${cell.language}\n${lines}\n\`\`\``, { supportHtml: true, isTrusted: false });
```

packages/notebook/src/browser/view/notebook-cell-list-view.tsx
```diff
@@ -137,14 +137,14 @@ export class NotebookCellListView extends React.Component<CellListProps, Noteboo
             return;
         }
 
-        if (this.dragGhost) {
-            this.dragGhost.remove();
+        if (NotebookCellListView.dragGhost) {
+            NotebookCellListView.dragGhost.remove();
```
msujew, line 141: Minor:
```suggestion
        NotebookCellListView.dragGhost?.remove();
```

msujew (commented):
I'm wondering whether we should call `NotebookCellListView.dragGhost?.remove()` on `onDrop` as well. Otherwise the element stays in the `body` indefinitely, right?

Aside from that, looks pretty great. Good idea using the markdown renderer to produce "pseudo" editors during dragging.

msujew (commented):
I've managed to break the markdown rendering by including ```` ``` ```` in the first line of the cell source:

image

I think it would be alright to just run `replaceAll('```', '')` on the `lines` value. It probably doesn't matter if it had any significance in the input text.

Aside from that, I've noticed that scrolling in the notebook doesn't work while dragging a cell, but that's unrelated to the PR.

jonah-iden:
@msujew thanks, good catch. fixed it the way you suggested for. Not super happy with that though since in python iirc you use ``` to indicate multi line strings.

The scrolling seems to be general theia problem. Same thing happens in the file explorer and other places as soon as you drag anything. If there is not already an issue for that we should probably reate one

msujew:
@jonah-iden AFAIK python uses `"""`. Anyway, I found a way to correctly escape everything and pushed it to your branch.

msujew (approved):
Looks good to me, thanks 👍

PR #13799, opened 2024-06-11
Fix type definition of `TheiaAppFactory`

tortmayr:
#### What it does

Ensure that the `TheiaAppFactory` type does  match the constructor function of TheiaApp. `InitialWorkspace` is required in the TheiaApp constructor butwas optional in `TheiaAppFactory` This ensures that (sub)classes of TheiaApp can properly be used as app factory without type casting

#### How to test

Using the TheiaAppLoader with a `TheiApp` (sub)class as factory parameter should now work and not have a ts error:
```ts
// Mock playwright args
cons args= {} as TheiaPlaywrightTestConfig & PlaywrightWorkerArgs

// Simle test 
const app= TheiaAppLoader.load({playwright,browser}, undefined, TheiaApp)

// test with subclass
class SpecialTheiaApp extends TheiaApp{
}
const app:SpecialTheiaApp= TheiaAppLoader.load({playwright,browser}, undefined, SpecialTheiaApp)

On master this would throw a compile error and a cast like `SpecialTheiaApp as TheiaAppFactory<SpecialApp>` would required
```
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/playwright/src/theia-app-loader.ts
+++ b/examples/playwright/src/theia-app-loader.ts
@@ -22,7 +22,7 @@ import { TheiaWorkspace } from './theia-workspace';
 import { OSUtil } from './util';
 
 export interface TheiaAppFactory<T extends TheiaApp> {
-    new(page: Page, initialWorkspace?: TheiaWorkspace, isElectron?: boolean): T;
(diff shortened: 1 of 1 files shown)


planger (approved):

PR #13805, opened 2024-06-12
Update doc comments on service-connection-provider.ts

tsmaeder:
#### What it does

Add some documentation to make the usage of the local/remote service id's more understandable.

#### How to test

n/a

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/messaging/service-connection-provider.ts
+++ b/packages/core/src/browser/messaging/service-connection-provider.ts
@@ -20,7 +20,13 @@ import { ChannelMultiplexer } from '../../common/message-rpc/channel';
 import { Deferred } from '../../common/promise-util';
 import { ConnectionSource } from './connection-source';
 
+/**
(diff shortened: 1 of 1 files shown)


tsmaeder:
@eneufeld

msujew (commented):
Looks good, two points of improvement on this:

1. You mention SSH a bunch, but this isn't the only way to use a remote system (dev container being another one).
2. I think we can avoid further confusion by specifying that, if the user isn't using a remote system (i.e. normal Electron/Browser system), both `LocalConnectionProvider` and `RemoteConnectionProvider` point towards the same system.

tsmaeder:
> You mention SSH a bunch, but this isn't the only way to use a remote system

The idea is not to give a complete description of the system here, only an example.

msujew (approved):
Thanks, it's definitely an improvement over the previous doc 👍

Issue #13679, opened 2024-05-03
TabGroup.viewColumn and Editor.viewColumn are out of sync for the same editor.


tsmaeder:
### Bug Description:
When starting up typescript-language-features 1.881 the context key `typescript.isManagedFile` is never set. Analyzing the problem led to finding out that the extension fails to find the active typescritpt editor due to this comparison failing for the single open editor: 
```
editor.viewColumn === tab.group.viewColumn;
```

`editor.viewColumn` is one and `tab.group.ViewColumn` is 2, but there is only one editor open.

### Steps to Reproduce:
n/a

### Additional Information

- Operating System:  Windows 11
- Theia Version: Theia 1.49 master

tsmaeder:
Remove the workaround from https://github.com/eclipse-theia/theia/pull/13673 once we have a fix here.

tsmaeder:
Turns out this problem leads to pretty much all commands from `typescript-language-features` being disabled, even stuff that is not context dependent like "restart tsserver".

tsmaeder:
https://github.com/microsoft/vscode/blob/main/extensions/typescript-language-features/src/ui/activeJsTsEditorTracker.ts#L64

tsmaeder:
@msujew I'm thinking the right thing to do here would to separate state computation for editors and documents completely. We can track documents quite easily and we don't need to use diffing to compute a delta. A second service should track editors and their state, including the tab groups, etc. Currently, we have `tabs-main.ts`, `documents-main.ts`, `editors-and-document-main.ts`, `text-editors-main` which are listening to each other in order to update the back end properly. I think consolidating these services might be simpler. WDYT?

tsmaeder:
Here's a list of data that the plugin api gives access that is pertinent to this issue:

1. `notebookDocuments/textDocuments`: 
the list of documents "known to the editor", whatever that means. This collection might also contain documents which are used internally by the front end but not visible in an editor.
2. `openTextDocument/NotebookDocument`. I would expect opened document to be reflected in 1. 
3. TextEditor/NotebookEditor
These reference a document. I would expect the document to be reflected in 1. They also reference a `ViewColumn`
4. `TabGroup`s. 
[...]

PR #13812, opened 2024-06-13
Fix tab group API event order

msujew:
#### What it does

Works around (but not actually closes) https://github.com/eclipse-theia/theia/issues/13679

This fixes the tab group API event order and ensures that the tab groups are always up-to-date on the plugin host by delaying the triggering of other events like `onDidChangeActiveEditor`.

This also sets all columns of groups to `0`. This should at least work around https://github.com/eclipse-theia/theia/issues/13679. See follow ups for more details on this.

#### How to test

1. Assert that the CI is green (TypeScript tests pass)
2. Open the App and open a TypeScript file
3. The TypeScript commands should be available
4. This should also work after switching between tabs and tab groups.

#### Follow-ups

The implementation of the `ViewColumnService` hasn't been kept up-to-date with the current requirements of Theia or VS Code. For example, it isn't aware of the position of tab groups (only of tabs) or of floating editor windows. And even for those tabs, it yields very much unexpected values. This needs an entire refactoring using something similar like VS Code's `EditorGroupService` implementation.

I'll probably get to this in a separate PR, but that's a larger issue so I'm not sure when I'll be able to do that.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/tabs/tabs-main.ts
+++ b/packages/plugin-ext/src/main/browser/tabs/tabs-main.ts
@@ -25,7 +25,6 @@ import { toUriComponents } from '../hierarchy/hierarchy-types-converters';
 import { TerminalWidget } from '@theia/terminal/lib/browser/base/terminal-widget';
 import { DisposableCollection } from '@theia/core';
 import { NotebookEditorWidget } from '@theia/notebook/lib/browser';
-import { ViewColumnService } from '@theia/core/lib/browser/shell/view-column-service';
 import { Deferred } from '@theia/core/lib/common/promise-util';
 
 interface TabInfo {
@@ -50,7 +49,6 @@ export class TabsMainImpl implements TabsMain, Disposable {
     private currentActiveGroup: TabGroupDto;
 
     private tabGroupChanged: boolean = false;
-    private viewColumnService: ViewColumnService;
 
     private readonly defaultTabGroup: TabGroupDto = {
         groupId: 0,
@@ -66,7 +64,6 @@ export class TabsMainImpl implements TabsMain, Disposable {
         this.proxy = rpc.getProxy(MAIN_RPC_CONTEXT.TABS_EXT);
 
         this.applicationShell = container.get(ApplicationShell);
-        this.viewColumnService = container.get(ViewColumnService);
         this.createTabsModel();
 
         const tabBars = this.applicationShell.mainPanel.tabBars();
@@ -82,7 +79,6 @@ export class TabsMainImpl implements TabsMain, Disposable {
         );
 
         this.connectToSignal(this.toDisposeOnDestroy, this.applicationShell.mainPanel.widgetAdded, (mainPanel, widget) => {
-            this.viewColumnService.updateViewColumns();
             if (this.tabGroupChanged || this.tabGroupModel.size === 0) {
                 this.tabGroupChanged = false;
                 this.createTabsModel();
@@ -276,7 +272,6 @@ export class TabsMainImpl implements TabsMain, Disposable {
     }
 
     private onTabTitleChanged(title: Title<Widget>): void {
-        this.viewColumnService.updateViewColumns();
         const tabInfo = this.getOrRebuildModel(this.tabInfoLookup, title);
         if (!tabInfo) {
             return;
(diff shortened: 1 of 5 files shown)


packages/plugin-ext/src/main/browser/tabs/tabs-main.ts
```diff
@@ -110,6 +113,22 @@ export class TabsMainImpl implements TabsMain, Disposable {
         });
     }
 
+    waitForWidget(widget: Widget): Promise<void> {
+        const deferred = new Deferred<void>();
+        this.waitQueue.set(widget, deferred);
+
+        const timeout = setTimeout(() => {
+            deferred.resolve(); // resolve to unblock the event
```
tsmaeder, line 121: How are we running into this case?

packages/plugin-ext/src/main/browser/tabs/tabs-main.ts
```diff
@@ -247,9 +272,11 @@ export class TabsMainImpl implements TabsMain, Disposable {
             tabDto,
             groupId: group.groupId
         });
+        this.waitQueue.get(args.title.owner)?.resolve();
```
tsmaeder, line 275: Wouldn't it make sense to remove the wait queue deferred immediately?

tsmaeder:
If I understand correctly, the idea of this workaround is to delay notifications through the editor service until the tabs main service has had time to update the tab info, right?
The problem seems to arise because we state (editor state, tab groups) that needs to be update "atomically" in order to not present an inconsistent picture to clients. However, since we're using independent widget listeners to trigger updates to various components, we can't sequence the actions properly.
To me it sounds as if we needed to introduce a single coordinator object that listens to the widget changes and then orchestrates the proper sequence of updates. Once the local state is consistent again, we can replicate the state to the back end/plugin host. Your thoughts?

msujew:
@tsmaeder My thoughts exactly, see the follow-ups section of the PR. There would only be a single service that would fire events in the expected order while also correctly keeping track of the tabbars/viewcolumns.

packages/plugin-ext/src/main/browser/tabs/tabs-main.ts
msujew: For example, if the widget has been added (somehow) to a non-main-panel tabbar. We probably still want to trigger the event on the plugin host, even if the tab API is not aware of the widget.

packages/plugin-ext/src/main/browser/tabs/tabs-main.ts
```diff
@@ -80,6 +82,7 @@ export class TabsMainImpl implements TabsMain, Disposable {
         );
 
         this.connectToSignal(this.toDisposeOnDestroy, this.applicationShell.mainPanel.widgetAdded, (mainPanel, widget) => {
+            this.viewColumnService.updateViewColumns();
```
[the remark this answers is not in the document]
msujew: Oops, that's a leftover from a previous attempt to fix this.

packages/plugin-ext/src/main/browser/tabs/tabs-main.ts
msujew: We do, see 

https://github.com/eclipse-theia/theia/blob/c727f6d56a3b945b2d058a252cb88a2344368b55/packages/plugin-ext/src/main/browser/tabs/tabs-main.ts#L124-L127

tsmaeder: I've seen that, but is the "then" guaranteed to run in the same tick?

msujew: No, but does it make any real difference?

tsmaeder: The only scenario I can see is that someone might invoke `waitForWidget` in the mean time, setting a new `deferred` into the `waitForQueue` and then the resolve would remove the second deferred. This would lead to no deferred in the map, the second deferred would get resolved after a second anyway, due to the timeout.

tsmaeder (approved):

packages/plugin-ext/src/main/browser/tabs/tabs-main.ts
msujew: I guess, it's unlikely, but I've adjusted the code 👍

PR #13830, opened 2024-06-20
improved shown keybindings in context menu

jonah-iden:
#### What it does

This improves the visible keybindings in context menus by not showing keybindings that are inactive in the current context.

#### How to test

For example add a keybinding like this to some keybinding contribution
```
 {
        command: CommonCommands.UNDO.id,
        keybinding: 'z',
        when: 'notebookEditorFocused',
}
```

previously this was the always shown as `Z` in the context menu regardless of if a notebook editor was focused.

Now this should still normally be `ctrl/cmd+z` but when a notebook editor is focused it should show as `Z`

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/menu/browser-menu-plugin.ts
+++ b/packages/core/src/browser/menu/browser-menu-plugin.ts
@@ -470,7 +470,9 @@ export class MenuCommandRegistry extends PhosphorCommandRegistry {
         const bindings = keybindingRegistry.getKeybindingsForCommand(id);
         // Only consider the first active keybinding.
         if (bindings.length) {
-            const binding = bindings.find(b => !b.when || this.services.contextKeyService.match(b.when)) ?? bindings[0];
+            const binding = bindings.length > 1 ?
+                bindings.find(b => !b.when || this.services.contextKeyService.match(b.when)) ?? bindings[0] :
+                bindings[0];
             const keys = keybindingRegistry.acceleratorFor(binding, ' ', true);
             this.addKeyBinding({
                 command: id,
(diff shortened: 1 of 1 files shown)


packages/core/src/browser/menu/browser-menu-plugin.ts
```diff
@@ -468,9 +468,11 @@ export class MenuCommandRegistry extends PhosphorCommandRegistry {
         });
 
         const bindings = keybindingRegistry.getKeybindingsForCommand(id);
-        // Only consider the first keybinding.
+        // Only consider the first active keybinding.
         if (bindings.length) {
-            const binding = bindings[0];
+            const binding = bindings.length > 1 ?
+                bindings.find(b => !b.when || this.services.contextKeyService.match(b.when)) ?? bindings[0] :
+                bindings[0];
```
msujew, line 475: Suggestion: Do we really need the conditional? Given that we always fall back to `bindings[0]` anyway?
```suggestion
            const binding = bindings.find(b => !b.when || this.services.contextKeyService.match(b.when)) ?? bindings[0]
```

msujew (approved):
LGTM 👍

packages/core/src/browser/menu/browser-menu-plugin.ts
jonah-iden: @msujew it is a small performance improvement. Most keybindings only have one binding. with the conditional it does not have to evaluate the when clause for every single keybinding

msujew: Ok, I can see that 👍

PR #13831, opened 2024-06-20
Switch single instance on per default.

### LINKED ISSUE #10890: Use Electron's single instance lock by default
_Originally inspired by [this thread](https://community.theia-ide.org/t/chinese-language-pack-extension-not-work-in-second-instance/2302)_

By default, the electron backend does not use the `singleInstance` lock. Consequently, clicking on the app shortcut starts another Theia process, which __doesn't have access to the default `index.db` file__, since each backend app locks their instance of the `index.db`. It will therefore create its own, with all application wide settings reseted (layout, locale, etc.).

I propose to set `singleInstance` to true by default and refactor the `onSecondInstance` method to simply start a new workspace with the supplied arguments.

tsmaeder:
#### What it does
Turn on the `singleInstance` application config on in the default case. Invoking theia a second time will open a new window in the first instance when `singleInstance === true`. If a workspace directory is given as the first position argument, this workspace is opened, otherwise an empty window is opened.

Fixes #10890

#### How to test
Invoke theia via `theia electron start` multiple times and with or without parameters. Make sure an independent instance is started when `singleInstance` is off. I use `theia electron start` and start the electron back end in the debugger to verify the behavior.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-main/electron-main-application.ts
+++ b/packages/core/src/electron-main/electron-main-application.ts
@@ -721,14 +721,19 @@ export class ElectronMainApplication {
     }
 
     protected async onSecondInstance(event: ElectronEvent, argv: string[], cwd: string): Promise<void> {
-        const electronWindows = BrowserWindow.getAllWindows();
-        if (electronWindows.length > 0) {
-            const electronWindow = electronWindows[0];
-            if (electronWindow.isMinimized()) {
-                electronWindow.restore();
-            }
-            electronWindow.focus();
-        }
+        createYargs(this.processArgv.getProcessArgvWithoutBin(argv), process.cwd())
+        .help(false)
+        .command('$0 [file]', false,
+            cmd => cmd
+                .positional('file', { type: 'string' }),
+                async args => {
+                    this.handleMainCommand({
+                        file: args.file,
+                        cwd: process.cwd(),
+                        secondInstance: true
+                    });
+                },
+            ).parse();
     }
 
     protected onWindowAllClosed(event: ElectronEvent): void {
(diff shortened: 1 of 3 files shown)


tsmaeder:
@sdirix have a look.

packages/core/src/electron-main/electron-main-application.ts
```diff
@@ -721,14 +721,19 @@ export class ElectronMainApplication {
...
+            cmd => cmd
+                .positional('file', { type: 'string' }),
+                async args => {
+                    this.handleMainCommand({
+                        file: args.file,
+                        cwd: process.cwd(),
+                        secondInstance: true
+                    });
+                },
+            ).parse();
```
msujew, line 736: Formatting: Applying the TypeScript formatter will autoindent the lines.
```suggestion
        createYargs(this.processArgv.getProcessArgvWithoutBin(argv), process.cwd())
            .help(false)
            .command('$0 [file]', false,
                cmd => cmd
                    .positional('file', { type: 'string' }),
                async args => {
                    this.handleMainCommand({
                        file: args.file,
                        cwd: process.cwd(),
                        secondInstance: true
                    });
                },
            ).parse();
```

msujew (approved):
I'm not sure whether this should be done as part as this PR or in a separate PR:

In VS Code, when attempting to open a workspace via `code <path>`, it will first check whether a window with that workspace is already opened and bring that into focus. I'm wondering whether we should do the same thing.

Anyways, given that this PR accomplishes what it's supposed to, I'll approve. Please fix the minor formatting issue (see below) before merging.

PR #13833, opened 2024-06-20
Serialize saveables to disk for "Save As"

msujew:
#### What it does

Adds a new optional `serialize` method to all `Saveable` objects. This method will be used in case the saveable is not directly string-serializable (such as notebooks) for the "Save As" action.

#### How to test

1. Open a `ipynb` file, and edit it
2. Save it to a new file using `Save As...`
3. The new file should contain all old changes + new changes
4. The old file should be correctly reverted

Please also confirm that the "Save As" operation still works for normal text editors

#### Review checklist

#### Reminder for reviewers


jonah-iden (approved):
Works great both for notebooks and text editors. Other webview based editors like drawio or bpmn do not have the option to be "saved as" so i couldn't test it with something else.

PR #13834, opened 2024-06-20
Fix automated releases

msujew:
#### What it does

Updates our GitHub actions to retry in case they hit network related errors like rate limiting. This was also used in the old `next` publishing action:

https://github.com/eclipse-theia/theia/blob/85f5e928ddfa650b5ba58a510a7db9532a392e7a/.github/workflows/ci-cd.yml#L164-L174

Also removes the `--canary` option. That should hopefully fix the issue of unexpected minor versions (we currently publish all `next` releases as `1.46.0-next`).

#### How to test

Nothing to test, we need to see whether it will work now :)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -45,10 +45,16 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # https://github.com/microsoft/vscode-ripgrep/issues/9
 
       - name: Publish NPM
-        shell: bash
-        run: |
-          yarn publish:latest -- ${{ inputs.release_type }}
-          yarn publish:check
+        # We need to retry the publish command in case of network errors.
+        # Those errors are usually due to rate limiting on the NPM registry.
+        uses: nick-invision/retry@7152eba30c6575329ac0576536151aca5a72780e # v3.0.0
+        with:
+          timeout_minutes: 5
+          retry_wait_seconds: 30
+          max_attempts: 3
+          retry_on: error
+          command: yarn publish:latest -- ${{ inputs.release_type }} && yarn publish:check
+          on_retry_command: git reset --hard
         env:
           NPM_CONFIG_PROVENANCE: "true" # enable provenance check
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-release.yml
```diff
@@ -45,10 +45,16 @@ jobs:
...
-          yarn publish:check
+        # We need to retry the publish command in case of network errors.
+        # Those errors are usually due to rate limiting on the NPM registry.
+        uses: nick-invision/retry@7152eba30c6575329ac0576536151aca5a72780e # v3.0.0
+        with:
+          timeout_minutes: 5
+          retry_wait_seconds: 30
+          max_attempts: 3
+          retry_on: error
+          command: yarn publish:latest -- ${{ inputs.release_type }} && yarn publish:check
```
jfaltermeier, line 56: The publishing guide recommends using a specific command in case of publishing errors:
https://github.com/eclipse-theia/theia/blob/master/doc/Publishing.md#failures-during-publishing
`npx lerna publish from-package --no-git-reset --no-git-tag-version --no-push`

I'm not familiar with the reasoning behind this choice. Maybe @vince-fugnitto could provide some insight?

vince-fugnitto: @jfaltermeier in the past we often encountered NPM socket errors which meant that only a subset of extensions were successfully published. In this state we cannot re-perform the publishing as it would fail with a message that the extension was already published. In order to only publish the extensions which were not previously published we manually ran the steps referenced in [failure-during-publishing](https://github.com/eclipse-theia/theia/blob/master/doc/Publishing.md#failures-during-publishing) to continue with the release.

It's possible that the proposed workflow will fail when retrying if we are in a state where we partially published extensions.

msujew: Since we run `git reset --hard` in between retries, running `lerna publish from-package` is not necessary. Just repeating the `publish:latest` command should work as expected. I don't think we need to adjust something.

Also, having two separate commands in the CI (for normal publishing + publishing following an error) would make this more complicated to maintain imo.

PR #13839, opened 2024-06-24
[vscode] Support TestRunRequest.preserveFocus

### LINKED ISSUE #13759: [vscode] Implement TestRunRequest preserveFocus API
TestRunRequest has a new boolean Property, preserveFocus, that drives the behavior of the Test Results view

rschnekenbu:
#### What it does

Add a progress service for tests to support activation of views
Support vscode API to listen for test run request preserveFocus:
- when preserve focus is true, nothing had to be done
- if preserveFocus is false, then the Test Results view shall be activated.

fixes #13759

#### How to test

1. Start the theia browser example on master and install the following extension (derived from vscode sample extension test-provider-sample):
- src: [test-provider-sample-0.0.1-src.zip](https://github.com/user-attachments/files/15952608/test-provider-sample-0.0.1-src.zip)
- zip vsix: [test-provider-sample-0.0.1.zip](https://github.com/user-attachments/files/15952610/test-provider-sample-0.0.1.zip)

2. When starting extensions from Test UI, nothing should happen (as _preserveFocus_ is always true in this case)
3. With the PR and the extensions, 2 additional commands are available with the markdown tests: Test Sample: Trigger Tests (preserveFocus: **true/false**). With the one with true, nothing shall happen also. The Test Result view shall activate when using the command with the false preserve Focus. 

#### Follow-ups
The current implementation may be extended later based on preferences and test results status. In VS Code, when tests are started or failing, the Test Result view or the Test Explorer view can be activated based on user preferences (see https://github.com/microsoft/vscode/blob/f9238dd94a8b54dc3adb274927c7871e7c215300/src/vs/workbench/contrib/testing/browser/testingProgressUiService.ts#L49)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/common/plugin-api-rpc.ts
+++ b/packages/plugin-ext/src/common/plugin-api-rpc.ts
@@ -2265,7 +2265,7 @@ export interface TestingMain {
 
     // Test runs
 
-    $notifyTestRunCreated(controllerId: string, run: TestRunDTO): void;
+    $notifyTestRunCreated(controllerId: string, run: TestRunDTO, preserveFocus: boolean): void;
     $notifyTestStateChanged(controllerId: string, runId: string, stateChanges: TestStateChangeDTO[], outputChanges: TestOutputDTO[]): void;
     $notifyTestRunEnded(controllerId: string, runId: string): void;
 }
--- a/packages/plugin-ext/src/main/browser/test-main.ts
+++ b/packages/plugin-ext/src/main/browser/test-main.ts
@@ -18,6 +18,7 @@ import { SimpleObservableCollection, TreeCollection, observableProperty } from '
 import {
     TestController, TestItem, TestOutputItem, TestRun, TestRunProfile, TestService, TestState, TestStateChangedEvent
 } from '@theia/test/lib/browser/test-service';
+import { TestExecutionProgressService } from '@theia/test/lib/browser/test-execution-progress-service';
 import { AccumulatingTreeDeltaEmitter, CollectionDelta, DeltaKind, TreeDelta, TreeDeltaBuilder } from '@theia/test/lib/common/tree-delta';
 import { Emitter, Location, Range } from '@theia/core/shared/vscode-languageserver-protocol';
 import { Range as PluginRange, Location as PluginLocation } from '../../common/plugin-api-rpc-model';
@@ -245,13 +246,14 @@ class TestRunProfileImpl implements TestRunProfile {
         this.proxy.$onConfigureRunProfile(this.controllerId, this.id);
     }
 
-    run(name: string, included: TestItem[], excluded: TestItem[]): void {
+    run(name: string, included: TestItem[], excluded: TestItem[], preserveFocus: boolean): void {
         this.proxy.$onRunControllerTests([{
             controllerId: this.controllerId,
             name,
             profileId: this.id,
             includedTests: included.map(item => itemToPath(item)),
-            excludedTests: excluded.map(item => itemToPath(item))
+            excludedTests: excluded.map(item => itemToPath(item)),
+            preserveFocus
         }]);
     }
 }
@@ -551,12 +553,14 @@ class TestControllerImpl implements TestController {
 
 export class TestingMainImpl implements TestingMain {
     private testService: TestService;
+    private testExecutionProgressService: TestExecutionProgressService;
     private controllerRegistrations = new Map<string, [TestControllerImpl, Disposable]>();
     private proxy: TestingExt;
     canRefresh: boolean;
 
     constructor(rpc: RPCProtocol, container: interfaces.Container, commandRegistry: CommandRegistryMainImpl) {
         this.testService = container.get(TestService);
+        this.testExecutionProgressService = container.get(TestExecutionProgressService);
         this.proxy = rpc.getProxy(MAIN_RPC_CONTEXT.TESTING_EXT);
         commandRegistry.registerArgumentProcessor({
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -614,7 +618,8 @@ export class TestingMainImpl implements TestingMain {
     $removeTestRunProfile(controllerId: string, profileId: string): void {
         this.withController(controllerId).removeProfile(profileId);
     }
-    $notifyTestRunCreated(controllerId: string, run: TestRunDTO): void {
+    $notifyTestRunCreated(controllerId: string, run: TestRunDTO, preserveFocus: boolean): void {
+        this.testExecutionProgressService.onTestRunRequested(preserveFocus);
         this.withController(controllerId).addRun(run.id, run.name, run.isRunning);
     }
     $notifyTestStateChanged(controllerId: string, runId: string, stateChanges: TestStateChangeDTO[], outputChanges: TestOutputDTO[]): void {
--- a/packages/plugin-ext/src/plugin/tests.ts
+++ b/packages/plugin-ext/src/plugin/tests.ts
@@ -135,7 +135,7 @@ export class TestControllerImpl implements theia.TestController {
     }
 
     createTestRun(request: theia.TestRunRequest, name?: string, persist: boolean = true): theia.TestRun {
-        return this.testRunStarted(request, name || '', persist, true);
+       return this.testRunStarted(request, name || '', persist, true);
     }
 
     dispose() {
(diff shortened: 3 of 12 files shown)


packages/plugin-ext/src/common/plugin-api-rpc.ts
rschnekenbu, line 1: Not really sure about naming here. Maybe $notifyTestRunRequested may be more aligned with the other methods?

packages/test/src/browser/view/test-execution-progress-service.ts
```diff
@@ -0,0 +1,54 @@
...
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { inject, injectable } from '@theia/core/shared/inversify';
```
tsmaeder, line 17: Not sure why we need this class. Why isn't this handled in `TestService`?

tsmaeder: Or alternatively, don't go through the test `TestService` but notify this service directly from `test-main.ts`.

packages/plugin-ext/src/plugin/tests.ts
```diff
@@ -197,7 +198,7 @@ export class TestControllerImpl implements theia.TestController {
             .filter(isDefined);
 
         const request = new TestRunRequest(
-            includeTests, excludeTests, profile, false // don't support continuous run yet
+            includeTests, excludeTests, profile, false /* don't support continuous run yet */, preserveFocus
         );
 
         const run = this.testRunStarted(request, name, false, false);
```
tsmaeder, line 204: We are not notifying the UI with `$notifyOnTestRunRequest`. Why not?

packages/plugin/src/theia.d.ts
```diff
@@ -16242,13 +16242,22 @@ export module '@theia/plugin' {
          */
         readonly continuous?: boolean;
 
+        /**
+         * Controls how test Test Results view is focused.  If true, the editor
+         * will keep the maintain the user's focus. If false, the editor will
+         * prefer to move focus into the Test Results view, although
+         * this may be configured by users.
```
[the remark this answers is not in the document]
tsmaeder: Adding the preference shouldn't be hard, right?

packages/plugin-ext/src/main/browser/test-main.ts
```diff
@@ -624,4 +625,8 @@ export class TestingMainImpl implements TestingMain {
     $notifyTestRunEnded(controllerId: string, runId: string): void {
         this.withController(controllerId).withRun(runId).ended();
     }
+
+    $notifyOnTestRunRequest(preserveFocus: boolean): void {
```
tsmaeder, line 629: Can't we use `$notifyTestRunCreated` instead of adding a new notification? A test run starts either at creation or when it's "running" flage changes, right?

tsmaeder (commented):
First reading of the code.

packages/test/src/browser/view/test-execution-progress-service.ts
rschnekenbu: I had an issue with the architecture there: the test views depend on the TestService, so I could not have the TestService to depend on them.
But yes, I can notify it from test-main.ts. I will propose a new version with TestExecutionProgressService directly used in TestingMain.

packages/plugin/src/theia.d.ts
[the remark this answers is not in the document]
rschnekenbu: The preference has been added, with a bonus additional option: Open Test Explorer on TestRun start.

VS Code has a fourth preference: openOnFail, but i would prefer to focus on other APIs for the end of this month.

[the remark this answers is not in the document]
rschnekenbu: Default is of course _neverOpen_, I also do not like switching view without explicitly requiring it by hand.

packages/test/src/browser/view/test-execution-progress-service.ts
rschnekenbu: The new commit has shifted a bit the way it is architectured, as you suggested. TestingMain is directly invoking the progress service and not channeling events through main test service anymore.

packages/plugin-ext/src/plugin/tests.ts
rschnekenbu: The `$notifyOnTestRunRequest` has been replaced by the `$notifyOnTestRunCreated`. It should now be always notified, regardless where the TestRun is started.

packages/plugin-ext/src/main/browser/test-main.ts
rschnekenbu: done on new commit.

packages/test/src/browser/test-execution-progress-service.ts
```diff
@@ -0,0 +1,69 @@
...
+export interface TestExecutionProgressService {
+    onTestRunRequested(preserveFocus: boolean): Promise<void>;
+}
+
+export const TestExecutionProgressService = Symbol('TestExecutionProgressService');
+
+@injectable()
+export class DefaultTestExecutionProgressService implements TestExecutionProgressService {
+
+    @inject(WidgetManager)
```
tsmaeder, line 33: This seems unused.

packages/test/src/browser/test-execution-progress-service.ts
```diff
@@ -0,0 +1,69 @@
...
+    protected readonly widgetManager: WidgetManager;
+
+    @inject(ApplicationShell)
+    protected readonly shell: ApplicationShell;
```
tsmaeder, line 37: unused

packages/test/src/browser/test-execution-progress-service.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    @inject(TestResultViewContribution)
+    protected readonly testResultView: TestResultViewContribution;
+
+    @inject(TestViewContribution)
+    protected readonly testView: TestViewContribution;
+
+    @inject(TestService)
```
tsmaeder, line 45: seems unused.

tsmaeder:
@rschnekenbu I have a couple of remarks from testing:

1. The preferences are in the "Extensions/Testing" section, but IMO should be in "Features/Testing".
2. When I close the test explorer view, it is not opened when I run the tests with "preserveFocus: false".
3. Not sure what the distinction is between "openOnStart" and "openExplorerOnStart" :shrug:

rschnekenbu:
I moved the preferences to right place. That is now under Features section.
I removed the openWithTextExplorer, as the option (copied from vs code) is not really obvious given the comment on the d.ts file and in the description of the preference. Now, the 2 options are 'do nothing' or 'open Test Results View'. I kept however the enum choice if we want to expand later the choices to other views.
I also removed unused code left from previous implementation.

rschnekenbu:
A small detail about the test of this feature: Running the tests from the UI should preserve the focus in any case, and not open any views. Only triggering from the extension, with the 'preserveFocus:false' version should open the view if the preference is set for this.

tsmaeder:
@rschnekenbu for me revealing the test explorer does not work in this scenario:

1. Set the preference to `openOnTestStart`.
2. Make sure the file explorer view is visible
3. Invoke `Trigger Test (preserveFocus: false)
4. Observe: the test explorer view is not revealed.

🤷

rschnekenbu:
> @rschnekenbu for me revealing the test explorer does not work in this scenario:
> 
> 1. Set the preference to `openOnTestStart`.
> 2. Make sure the file explorer view is visible
> 3. Invoke `Trigger Test (preserveFocus: false)
> 4. Observe: the test explorer view is not revealed.
> 
> 🤷

The Test Results View shall open, not the Test Explorer. Is that not the case?

tsmaeder:
> The Test Results View shall open, not the Test Explorer. Is that not the case?

No, the file explorer still is visible.

tsmaeder:
Cleared this up: The "Test Result View" is the one that shows up in the lower right corner where the terminals, etc. are.

tsmaeder (approved):

PR #13854, opened 2024-06-27
Use relative paths for ctx.importScripts()

### LINKED ISSUE #13813: Correctly load frontend plugins when deployed under a path
### Bug Description:

Currently in [worker-main](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/hosted/browser/worker/worker-main.ts) [importScripts](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts) is used to load the [plugin-vscode-init-fe.ts](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext-vscode/src/node/context/plugin-vscode-init-fe.ts). However, this does not work for instances, where Theia is not hosted at the base-url (or origin), like the `try.theia-cloud.io` deployment, where each instance of Theia is hosted under `ws.theia-cloud.io/<pod-id>`.

After a quick investigation the problem seems to be that `importScripts` is called with `/context/<file-name>`, which is a absolute path. Hence it resolves to only `base-url/context/<file-name>`, whereas we would need `base-url/<pod-id>/context/<file-name>`. This could be fixed by simply calling `importScripts` with `./context/<file-name>`. I already tested this and it seems to work.

Does anyone have any concerns about that change? It should not change the behavior for other use cases, but fix the issue for that specifc scenario.

I can gladly contribute this.

### Steps to Reproduce:

1. Use a hosted Theia instance that is running under a path (e.g. `https://try.theia-cloud.io/` - path meaning, that the pods are running under `ws.theia-cloud.io/<pod-id>` and not just `ws.theia-cloud.io`)
2. Install a frontend-only vscode extension
3. Observe that the console throws and error that the `plugin-vscode-init-fe.js` fails to load.

### Additional Information

- Operating System:
- Theia Version:

sgraband:
#### What it does

Before we tried to load absolute paths e.g. `/context/`. This is a problem, when working in deployed use cases, where Theia might be hosted under a path (e.g. try.theia-cloud.io). Because then the loaded path will be from the root of the page and not, like wanted, from the current path. Simply changing it to relative paths (e.g. `./context/`) solves this issue.

Fixes #13813

#### How to test

This will be hard to test, as you would need a deployment where Theia is hosted under a path (e.g. try.theia-cloud.io) and this fix is applied. As far as i am aware there is no easy way to run theia under a path locally.
However, it could be tested, that this does not break something for the example applications.

To test simply install a frontend vscode extension and observe that the console throws no error and that the files are loaded from the correct location.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/hosted/browser/worker/worker-main.ts
+++ b/packages/plugin-ext/src/hosted/browser/worker/worker-main.ts
@@ -45,7 +45,7 @@ const pluginsModulesNames = new Map<string, Plugin>();
 const scripts = new Set<string>();
 
 function initialize(contextPath: string, pluginMetadata: PluginMetadata): void {
-    const path = '/context/' + contextPath;
+    const path = './context/' + contextPath;
 
     if (!scripts.has(path)) {
         ctx.importScripts(path);
@@ -70,7 +70,7 @@ pluginManager.setPluginHost({
                         ctx.frontendModuleName = plugin.lifecycle.frontendModuleName;
                     }
 
-                    ctx.importScripts('/hostedPlugin/' + getPluginId(plugin.model) + '/' + plugin.pluginPath);
+                    ctx.importScripts('./hostedPlugin/' + getPluginId(plugin.model) + '/' + plugin.pluginPath);
                 }
             }
 


msujew (approved):
Looks good to me. I couldn't test everything, but the existing behavior is still correct, so I assume this still works as expected when Theia is deployed under a path.

PR #13900, opened 2024-07-09
[vscode] Implement stubbed API for activeStackFrame API

### LINKED ISSUE #13846: [vscode] Implement stubbed DebugStackFrame and DebugThread API
The API has been  stubbed in the debug namespace for:
```
export const activeStackItem: DebugThread | DebugStackFrame | undefined;
export const onDidChangeActiveStackItem: Event<DebugThread | DebugStackFrame | undefined>;
```
This task should implement the stubbed implementation.

rschnekenbu:
#### What it does
Implement previously stubbed API for activeStackFrame API (DebugStackFrame & DebugThread)
Fixes #13846

#### How to test
1. Install the following extension, that adds the listener on the active Stack item:
- src: [debug-extension-sample-0.0.2-src.zip](https://github.com/user-attachments/files/16140775/debug-extension-sample-0.0.2-src.zip)
- zip vsix: [debug-extension-sample-0.0.2.zip](https://github.com/user-attachments/files/16140777/debug-extension-sample-0.0.2.zip)
2. Start debugging a project (I ran 2 simple hello world examples in the video)
3. Switching between stack frames in the Call Stack or to a DebugThread in the Threads view. Information message shall be displayed in the notification area, with the current stack item information. 

https://github.com/eclipse-theia/theia/assets/3964263/d36e75d7-9be3-4a2e-8f32-6ff59536ab55

We don't have exactly the same interface as in VS Code, so we get more easily the empty stack item message. That should not impact users, as selecting a thread or a stack frame still throws the active stack item change event.
  
#### Follow-ups
none

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-session.tsx
+++ b/packages/debug/src/browser/debug-session.tsx
@@ -78,6 +78,11 @@ export class DebugSession implements CompositeTreeElement {
         return this.onDidFocusStackFrameEmitter.event;
     }
 
+    protected readonly onDidFocusThreadEmitter = new Emitter<DebugThread | undefined>();
+    get onDidFocusThread(): Event<DebugThread | undefined> {
+        return this.onDidFocusThreadEmitter.event;
+    }
+
     protected readonly onDidChangeBreakpointsEmitter = new Emitter<URI>();
     readonly onDidChangeBreakpoints: Event<URI> = this.onDidChangeBreakpointsEmitter.event;
     protected fireDidChangeBreakpoints(uri: URI): void {
@@ -254,8 +259,12 @@ export class DebugSession implements CompositeTreeElement {
         return this._currentThread;
     }
     set currentThread(thread: DebugThread | undefined) {
+        if (this._currentThread?.id === thread?.id) {
+            return;
+        }
         this.toDisposeOnCurrentThread.dispose();
         this._currentThread = thread;
+        this.onDidFocusThreadEmitter.fire(thread);
         this.fireDidChange();
         if (thread) {
             this.toDisposeOnCurrentThread.push(thread.onDidChanged(() => this.fireDidChange()));
--- a/packages/debug/src/browser/model/debug-thread.tsx
+++ b/packages/debug/src/browser/model/debug-thread.tsx
@@ -56,11 +56,18 @@ export class DebugThread extends DebugThreadData implements TreeElement {
         return this.session.id + ':' + this.raw.id;
     }
 
+    get threadId(): number {
+        return this.raw.id;
+    }
+
     protected _currentFrame: DebugStackFrame | undefined;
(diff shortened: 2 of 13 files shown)


rschnekenbu:
It seems the 3 Macos tests can't start. That should not block the review however.

packages/plugin-ext/src/plugin/debug/debug-ext.ts
```diff
@@ -262,6 +269,37 @@ export class DebugExtImpl implements DebugExt {
         });
     }
 
+    set activeStackItem(stackItem: theia.DebugStackFrame | theia.DebugThread | undefined) {
```
tsmaeder, line 272: I think we should check whether the active item actually changed before sending notifications.

packages/debug/src/browser/debug-session.tsx
```diff
@@ -254,8 +259,12 @@ export class DebugSession implements CompositeTreeElement {
         return this._currentThread;
     }
     set currentThread(thread: DebugThread | undefined) {
+        if (this._currentThread === thread) {
```
tsmaeder, line 262: I comparing instances the right thing or should we look at the id?

packages/debug/src/browser/model/debug-thread.tsx
```diff
@@ -56,11 +56,18 @@ export class DebugThread extends DebugThreadData implements TreeElement {
...
+    get threadId(): number {
+        return this.raw.id;
+    }
+
     protected _currentFrame: DebugStackFrame | undefined;
     get currentFrame(): DebugStackFrame | undefined {
         return this._currentFrame;
     }
     set currentFrame(frame: DebugStackFrame | undefined) {
+        if (this._currentFrame === frame) {
```
tsmaeder, line 68: Again, is there only ever one copy of the same frame?

tsmaeder (commented):
This seems to be one of the weaker API definitions from the VS Code team. What does "focused" even mean?  Some small comments.

packages/debug/src/browser/model/debug-thread.tsx
rschnekenbu: The code should update the existing one instead of creating new ones (see https://github.com/eclipse-theia/theia/blob/ade6c8de0fe4c223660fff67d1400d723ce89d3f/packages/debug/src/browser/model/debug-thread.tsx#L202). But comparing identifiers may be more robust indeed.

I pushed a new commit to address the issue.

rschnekenbu:
> This seems to be one of the weaker API definitions from the VS Code team. What does "focused" even mean? Some small comments.

Yes, a surprising API. In this case, 'focus' means what is current selection in the UI from what I can see from VS Code :shrug:

tsmaeder (approved):

Issue #13904, opened 2024-07-09
Plugin install fail from open-vsx.org


DigitLib:
### Bug Description:
Using Extensions tab don't install from open-vsx.org

### Steps to Reproduce:

1. Open Thei IDE
2. Go to Extensions tab
3. Install Astro plugin

### Additional Information

- Operating System: Linux CachyOS (Arch based)
- Theia Version: 1.51.0

None of extensions are installed when use this tab. Tried Continue and Astro. Shows installing but nothing happened.
CachyOS added recently installation in their repo.
Tried to installed appimage (official) and in installed app from repo.

JonasHelming:
I just tried, both extensions install for me without any issues. Can you describe exactly how you get to the "Extension tab", it is supposed to be the "Extension view" opened e.g. via the menue "View" => "Extensions"

DigitLib:
I use regular tab on the left, also tried with View -> Extensions. 
Checked the INFO in terminal and see that old versions are used 
Continue is  v0.8.43 but get in Extensions 0.1.3

For Astro latest is v2.11.0 but get v0.26.2 and can not install because can not find a file

INFO [[object Object]]: trying to resolve latest version...
INFO [astro-build.astro-vscode]: resolved to 'astro-build.astro-vscode-0.26.2'
INFO [astro-build.astro-vscode-0.26.2]: trying to download from "https://open-vsx.org/api/astro-build/astro-vscode/0.26.2/file/astro-build.astro-vscode-0.26.2.vsix"... to path /home/andrax/.theia-ide/tmp/vscode-download
INFO [astro-build.astro-vscode-0.26.2]: not found

Thank you.

msujew:
I can reproduce and found the issue, seems like the documentation is wrong... (note the `>=` and `<=` symbols)

![image](https://github.com/eclipse-theia/theia/assets/4377073/b06f25f9-7fb9-4809-a6da-9d61e31233e9)

DigitLib:
@msujew Yes!

PR #13907, opened 2024-07-10
Fix plugin version comparison

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/13904

I'm confused by how this wasn't caught earlier. Even though our method was supposed to return `gte`, it did run `lte` internally. I've simply adjusted the called method to match the name and documentation. See also:

image

#### How to test

Try installing the [Astro](https://open-vsx.org/extension/astro-build/astro-vscode) VS Code extension via the vsx-registry view. It should work as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/ovsx-client/src/ovsx-api-filter.ts
+++ b/dev-packages/ovsx-client/src/ovsx-api-filter.ts
@@ -127,7 +127,7 @@ export class OVSXApiFilterImpl implements OVSXApiFilter {
         if (!versionA || !versionB) {
             return false;
         }
-        return semver.lte(versionA, versionB);
+        return semver.gte(versionA, versionB);
     }
 
     protected supportedVscodeApiSatisfies(vscodeApiRange: string): boolean {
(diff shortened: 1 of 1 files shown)


rschnekenbu:
I remember discussing that point when reviewing https://github.com/eclipse-theia/theia/pull/13118.  I'll have a double check then.

rschnekenbu (changes requested):
I think we will have an issue for builtins if we fix this bug.
See https://github.com/eclipse-theia/theia/pull/13118#issuecomment-1831863457. With the existing implementation, we check the builtins version is lower than our API compatibility version. 

I would suggest to change the order of the parameters here: https://github.com/eclipse-theia/theia/blob/1491cd1673830e2971760ed8444d8a921f31b91c/dev-packages/ovsx-client/src/ovsx-api-filter.ts#L111. If not, we will finally install builtins with a version greater than our API compatibility.

rschnekenbu (approved):
Looks good to me! I could install astro (latest version) and also I could have Github builtin 1.88.1 installed.

PR #13916, opened 2024-07-11
Allow opening changes for files associated with custom editors

### LINKED ISSUE #9079: [Diff] Images should open in preview
The default open handler of binary images (gif, png, jpg, jpeg) in the diff pane should display images as a preview when double clicking them.

### Feature Description:

image

### LINKED ISSUE #10787: custom-editor: restore editor as part of layout

pisv:
#### What it does

Allows the user to visually compare two versions of a file that is associated with a custom editor. This is done by opening both versions of the file simultaneously in a side-by-side editor, with each version being opened in a separate custom editor contained in the side-by-side editor.

For example, this feature can be used to help spot the differences between two versions of an image file associated with the custom editor contributed by the VS Code `media-preview` plugin, as illustrated with the following screenshot:

Screenshot

More screenshots illustrating this feature being used with other custom editors:

Screenshot

Screenshot

This PR contains a series of individual commits that work together to make it possible for the `CustomEditorOpener` to open a diff-uri in a side-by-side editor containing the corresponding `CustomEditor`s. While the main functionality is contributed by the latest commit, the earlier commits provide various fixes and enhancements required to make it work.

Here is the sequence of commits:

- Fix loading of webview resources that depend on query params
    
    Some resource url (notably git) use the query part of the url to store additional information. The query part of the url was incorrectly being dropped while attempting to load these resources inside of webviews.

- Fix `Error: Unknown Webview` messages in the log

- Fix design and implementation issues surrounding `CustomEditorOpener`
    
    This commit ensures that the promise returned by `CustomEditorOpener.open` will only resolve to a properly initialized and opened `CustomEditorWidget`. In particular, it ensures that the widget is opened according to the specified `WidgetOpenerOptions`, including `widgetOptions.ref` and `mode`.
    
    Essentially, it revises the work done in #9671 and #10580.

- Restore custom editors as part of layout
    
    Fixes an incorrect assumption that a custom editor cannot be restored if no `WebviewPanelSerializer` is registered for its view type. (Actually, custom editors are created and restored using a custom editor provider.)
    
    Also, ensures that `CustomEditorWidget.modelRef` satisfies the shape for the `CustomEditorWidget` defined in `editor.ts` and cannot return `undefined`. (However, `CustomEditorWidget.modelRef.object` can be `undefined` until the custom editor is resolved.)
    
    Fixes #10787

- Fix a race condition when file system provider is activated
    
    When file system provider is activated, wait until it is registered.
    
- git: add support for custom editors
    
    * Uses `OpenerService` instead of `EditorManager` to open editors
    
    * Contributes a `FileSystemProvider` for git-resources
    
    * Fixes an issue with getting blob contents

- custom editor: open a diff-uri in a side-by-side editor
    
    `CustomEditorOpener` is now able to open a diff-uri in a side-by-side editor, which contains the corresponding `CustomEditor`s.
    
    Fixes #9079

#### How to test

There are various configurations that can be used for testing this feature:

* Windows vs. Linux vs. macOS

* Browser vs. Electron

* `vscode.git` vs. (now deprecated) `@theia/git`

This feature can be tested with any custom editor.

A basic test scenario involves making a change to a file associated with a custom editor (e.g., `.jpg`, `.png`, `.pawDraw`, or `.cscratch`) and then opening changes in a side-by-side editor by clicking on the file in the Source Control view.

Since this PR also fixes an issue with custom editors not being restored as part of layout, it is worth checking out that custom editors are now correctly restored, both standalone and as part of side-by-side editors.

Also, it is important to verify that there are no regressions, especially that standalone custom editors can still be opened and work as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/saveable.ts
+++ b/packages/core/src/browser/saveable.ts
@@ -21,7 +21,7 @@ import { MaybePromise } from '../common/types';
 import { Key } from './keyboard/keys';
 import { AbstractDialog } from './dialogs';
 import { nls } from '../common/nls';
-import { DisposableCollection, isObject } from '../common';
+import { Disposable, DisposableCollection, isObject } from '../common';
 import { BinaryBuffer } from '../common/buffer';
 
 export type AutoSaveMode = 'off' | 'afterDelay' | 'onFocusChange' | 'onWindowChange';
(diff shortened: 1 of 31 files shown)


pisv:
All checks are now green, and this PR is eagerly waiting for a review :-)

msujew:
FYI, I'll likely take a look at this, but don't have time for it before the upcoming release. Probably next week.

pisv:
@msujew Good to know! Thank you. I'd really appreciate your review.

msujew:
Sorry for the delay, but unexpected deadlines came up. Will have time to review starting next week :)

pisv:
No worries :) That'd be great! Thanks!

pisv:
(I've resolved a merge conflict and retested everything in preparation to the upcoming review.)

pisv:
@msujew Thank you very much for the thorough review!

> I've noticed one annoying issue/regression...

In general, a custom editor model cannot be shared between multiple editor instances, unlike a text editor model.

Therefore, the current implementation of the custom editor opener (on the master branch) forbids opening another instance of the same custom editor:

https://github.com/eclipse-theia/theia/blob/d34c0b3a8db8483da5b2bf25176e495110b25e52/packages/plugin-ext/src/main/browser/custom-editors/custom-editor-opener.tsx#L64-L91

Thus, in general, it would not be possible to open both the normal editor and the diff editor on the same custom editor model simultaneously in separate tabs, and there is no other way to open the diff editor on the same custom editor  model than to replace the existing normal editor.

And if the user tries to open the normal editor when the diff editor is already open on the same custom editor model, we could either replace or just activate the existing diff editor. I have chosen the latter alternative (activating the existing diff editor rather than replacing it with the normal editor), since the right side of the diff editor already contains a fully-functional custom editor.

> The left editor currently allows modifications

I noticed that it didn't always seem to be the case. For example, for `pawDraw` files, the left editor does not allow modifications. So, probably, it also depends on the implementation of a specific custom editor.

> sometimes the model stays dirty even though it was already saved

I noticed this issue as well, but thought it would be better to tackle it in a separate PR, as you have suggested.

Thanks again for your review!

PR #13936, opened 2024-07-17
Introduce environment variable to override connexion settings

### LINKED ISSUE #13800: Allow to override connection settings
In Theia, the desired connection settings are usually configured via the package.json, see for example [here where they are configured for the Theia IDE](https://github.com/eclipse-theia/theia-blueprint/blob/4580073933642b6cffb9e7aa90ba382330ab5551/applications/electron/package.json#L44-L45).

It's good practice to run e2e tests against the final built product. However, depending on how the tests are implemented, they might need a different set of connection settings than the actual delivered product itself. Therefore it would be nice to make them overridable, e.g. via environment variables. The e2e tests could then run against the same delivery, just with different settings.

rschnekenbu:
#### What it does
Adds support for a `FRONTEND_CONNECTION_TIMEOUT` environment variable to override configuration setting for the backend.

fixes #13800

#### How to test
1. Start default browser example (`yarn browser start` from terminal). Open a browser window and access `localhost:3000`
2. Close this browser window and check the settings value for frontend connexion timeout on the log in debug console. This should be some entry like `setting close timeout for id fb80451d-10da-416b-b656-062c78ba408a to 3000`
3. Set the environment variable to another value, as for example `export FRONTEND_CONNECTION_TIMEOUT=10000`
4. restart the default browser example and close the browser again. this time, the close timeout log should be the one that you set.
5. Some other tests can be made with `-1`, closing directly the connection or with something else than a number. It should then turn back to default config, e.g. 3000.

#### Follow-ups

none

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -152,6 +152,10 @@ Where `root` is the name of the logger and `INFO` is the log level. These are op
   - If possible, you should set this environment variable:
     - When not set, Theia will allow any origin to access the WebSocket services.
     - When set, Theia will only allow the origins defined in this environment variable.
+- `FRONTEND_CONNECTION_TIMEOUT`
+  - The duration in milliseconds during which the backend keeps the connexion contexts for the frontend to reconnect.
+  - This duration defaults to '0' if not set.
+  - If set to negative number, the backend will never close the connection.    
 
 ## Additional Information
 
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -153,7 +153,7 @@ Where `root` is the name of the logger and `INFO` is the log level. These are op
     - When not set, Theia will allow any origin to access the WebSocket services.
     - When set, Theia will only allow the origins defined in this environment variable.
 - `FRONTEND_CONNECTION_TIMEOUT`
-  - The duration in milliseconds during which the backend keeps the connexion contexts for the frontend to reconnect.
+  - The duration in milliseconds during which the backend keeps the connection contexts for the frontend to reconnect.
   - This duration defaults to '0' if not set.
   - If set to negative number, the backend will never close the connection.    
 
--- a/packages/core/README_TEMPLATE.md
+++ b/packages/core/README_TEMPLATE.md
@@ -121,6 +121,10 @@ Where `root` is the name of the logger and `INFO` is the log level. These are op
   - If possible, you should set this environment variable:
     - When not set, Theia will allow any origin to access the WebSocket services.
     - When set, Theia will only allow the origins defined in this environment variable.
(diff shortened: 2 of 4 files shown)


packages/core/README_TEMPLATE.md
```diff
@@ -121,6 +121,10 @@ Where `root` is the name of the logger and `INFO` is the log level. These are op
   - If possible, you should set this environment variable:
     - When not set, Theia will allow any origin to access the WebSocket services.
     - When set, Theia will only allow the origins defined in this environment variable.
+- `FRONTEND_CONNECTION_TIMEOUT`
+  - The duration in milliseconds during which the backend keeps the connexion contexts for the frontend to reconnect.
```
planger, line 125:
```suggestion
  - The duration in milliseconds during which the backend keeps the connection contexts for the frontend to reconnect.
```

packages/core/README.md
```diff
@@ -152,6 +152,10 @@ Where `root` is the name of the logger and `INFO` is the log level. These are op
   - If possible, you should set this environment variable:
     - When not set, Theia will allow any origin to access the WebSocket services.
     - When set, Theia will only allow the origins defined in this environment variable.
+- `FRONTEND_CONNECTION_TIMEOUT`
+  - The duration in milliseconds during which the backend keeps the connexion contexts for the frontend to reconnect.
```
planger, line 156:
```suggestion
  - The duration in milliseconds during which the backend keeps the connection contexts for the frontend to reconnect.
```

packages/core/src/node/messaging/websocket-frontend-connection-service.ts
```diff
@@ -124,6 +125,16 @@ export class WebsocketFrontendConnectionService implements FrontendConnectionSer
...
     }
+
+    private frontendConnectionTimeout(): number {
+        const envValue = Number(process.env['FRONTEND_CONNECTION_TIMEOUT']);
+        if (!isNaN(envValue)) {
+            return envValue;
+        }
+
+        const configValue = BackendApplicationConfigProvider.get().frontendConnectionTimeout;
+        return configValue;
```
planger, line 136:
```suggestion
        return BackendApplicationConfigProvider.get().frontendConnectionTimeout;
```

This is just a suggestion, ignore if you prefer the explicit variable, but I feel it doesn't add any additional clarity.

planger (approved):
Thanks, looks good to me and seems to work fine! :+1: Just a few minor inline suggestions.

rschnekenbu:
Thanks for this very quick review and suggestions, @planger! I still don't understand why we have different spelling for "connection" in French and English :D

PR #13939, opened 2024-07-18
Add DebugSessionOptions.testRun

### LINKED ISSUE #13872: [vscode] Support DebugSessionOptions#testRun introduced in VS Code 1.91
A new property was introduced on VS Code 1.91 on DebugSessionOptions: 
` /**
  * Signals to the editor that the debug session was started from a test run
  * request. This is used to link the lifecycle of the debug session and
  * test run in UI actions.
  */
  testRun?: TestRun;`

tsmaeder:
#### What it does
Adds the `DebugSessionOptions.testRun` field. The behavioral change associated with this is that a debug session that is started as part of test run will be terminated when the test run ends.

Fixes #13872

#### How to test
The attached plugin contributes a test controller with a single test. 
1. In the test explorer, click the "debug" icon on the line of the test
2. Observe: a test run is started and a nodejs program associated with test is started in the debugger
3. Select the test run in the test run view and choose "cancel test run" from the context menu
4. Observe: the debug session is terminated.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-session.tsx
+++ b/packages/debug/src/browser/debug-session.tsx
@@ -33,7 +33,7 @@ import { DebugSourceBreakpoint } from './model/debug-source-breakpoint';
 import debounce = require('p-debounce');
 import URI from '@theia/core/lib/common/uri';
 import { BreakpointManager } from './breakpoint/breakpoint-manager';
-import { DebugConfigurationSessionOptions, InternalDebugSessionOptions } from './debug-session-options';
+import { DebugConfigurationSessionOptions, InternalDebugSessionOptions, TestRunReference } from './debug-session-options';
 import { DebugConfiguration, DebugConsoleMode } from '../common/debug-common';
 import { SourceBreakpoint, ExceptionBreakpoint } from './breakpoint/breakpoint-marker';
 import { TerminalWidgetOptions, TerminalWidget } from '@theia/terminal/lib/browser/base/terminal-widget';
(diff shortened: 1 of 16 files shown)


tsmaeder:
Here's the extension

[testcontroller-0.0.1.zip](https://github.com/user-attachments/files/16281029/testcontroller-0.0.1.zip)

And here's the source code:


[testcontroller-src.zip](https://github.com/user-attachments/files/16281051/testcontroller-src.zip)

packages/debug/src/browser/debug-session.tsx
```diff
@@ -93,6 +95,9 @@ export class DebugSession implements CompositeTreeElement {
         readonly id: string,
         readonly options: DebugConfigurationSessionOptions,
         readonly parentSession: DebugSession | undefined,
+        testService: TestService,
```
rschnekenbu, line 98: I'm curious about consistency here (beginner question): Why not putting this as a private - or protected - readonly property instead of leaving it as a parameter? I see it is only use in constructor, so that may be the reason? Or to avoid the `override` from the `PluginDebugSession`?

packages/test/src/browser/view/test-run-view-contribution.ts
```diff
@@ -62,7 +62,10 @@ export class TestRunViewContribution extends AbstractViewContribution<TestRunTre
     override registerCommands(commands: CommandRegistry): void {
         super.registerCommands(commands);
         commands.registerCommand(TestViewCommands.CANCEL_RUN, {
-            isEnabled: t => TestRun.is(t) && t.isRunning,
+            isEnabled: t => {
+                console.log('is enabled');
```
rschnekenbu, line 66: Probably a leftover from debugging?

rschnekenbu (approved):
Looks good to me, it works as expected with the test extension.
Only a question and a minor change (console.log)

packages/debug/src/browser/debug-session.tsx
tsmaeder: Yes, if something is only used in the constructor, no need to add state to the class. State is bad!

PR #13903, opened 2024-07-09
upgrade ws to 8.17.1 & jsdom to 24.1.0

rschnekenbu:
#### What it does
Fixes #13848

This is a simple update on dependent library, known to have a risk of DoS. I tested the tool, no changes in behavior, and I also di not notice any changes in the logs.

#### How to test
Since this updates websocket lib, build and open the tool. Observe the logs to make sure we don't get any new entries.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #13942, opened 2024-07-22
Bug fix: "core.saveAll" command shall only save dirty widgets

### LINKED ISSUE #13923: Save All command blindly modifies all open files
### Bug Description:

The "core.saveAll" command saves all open files to disk, despite them not having been changed, changing the "last modified" timestamp. In our case, this is unnecessarily causing the GCC compiler to pick these files up again in its next run.

I guess the correct behavior would be to check for each file's isDirty flag and refrain from touching it if the isDirty flag is false.

### Steps to Reproduce:

1. Open any file and aside to it, the Properties view
2. Note the "Last modified" timestamp in the Properties view
3. Run the command "File: Save All"
4. Check the "Last modified" timestamp again in the Properties view (make sure to switch to another file and back to update the view first, it won't update automatically)

### Additional Information

- Operating System: Observed on OSX & Linux
- Theia Version: 1.51.0

KR155E:
#### What it does

Fixes #13923 

Makes the "core.saveAll" command check each file's isDirty flag instead of just blindly saving all open files.

#### How to test

1.  Open any file and aside to it, the Properties view
2. Note the "Last modified" timestamp in the Properties view
3. Run the command "File: Save All"
4. Check the "Last modified" timestamp again in the Properties view (make sure to switch to another file and back to update the view first, it won't update automatically)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/application-shell.ts
+++ b/packages/core/src/browser/shell/application-shell.ts
@@ -2066,7 +2066,7 @@ export class ApplicationShell extends Widget {
      */
     async saveAll(options?: SaveOptions): Promise<void> {
         for (const widget of this.widgets) {
-            if (this.saveableService.canSaveNotSaveAs(widget)) {
+            if (Saveable.isDirty(widget) && this.saveableService.canSaveNotSaveAs(widget)) {
                 await this.saveableService.save(widget, options);
             }
         }


martin-fleck-at (approved):
Change looks good to me and works as expected, thank you very much for your contribution @KR155E!

Note that the CI has some issues with Node version < 18 but this is unrelated to this PR and is caused by https://github.com/eclipse-theia/theia/pull/13903.

PR #8437, opened 2020-08-27
16502 plugin startup

tsmaeder:
#### What it does
This is a fix for https://github.com/eclipse-theia/theia/issues/8429. There are a couple of changes and refactorings going on here:

* In order to make sure that startup of plugin hosts happens in a well defined sequence, plugin hosts announce that they are starting by sending `willStartPluginHost`. Once they have finished, they send `didStartPluginHost` to the `HostedPluginClient` on the browser side.
* ✅ `MessageConnection` and related code always sends strings around. It is the responsibility of clients to `stringify` and `parse` objects
* ✅ `RPCMultiplexer` is now an implementation detail of `RPCProtocol`. 
* ✅ Sending messages between plugin hosts and clients now requires an explicit `pluginHostId`. So far, there was an implicit protocol based on `pluginId`, which just took the first plugin deployed to a plugin host to do routing. 
* ✅ Rename `HostedPluginServer/Client` in the plugin-dev package to `PluginDevServer/Client`.  Before, the wrong client interface was used for the plugin dev server, showing that the naming was confusing before.
* Turned `HostedPluginRunner` into just another `ServerPluginRunner`, removing the special case.

#### How to test
Install various plugins and check what happens to them 
* upon starting up for the first time
* kill the back end and start it again: plugins should come back correctly
* Refresh the browser window: plugins should be shut down and restarted.

Note: front-end plugins are currently borked and therefore could not be tested (https://github.com/eclipse-theia/theia/issues/6601)

#### Review checklist

- [x] as an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- as a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #13980, opened 2024-08-01
Don't use ChannelMultiplexer in RPCProtocol

### LINKED ISSUE #13960: Cannot Register URI Handler in Activation Event
### Bug Description:
In the context of https://github.com/eclipse-theia/theia/pull/13306, I have implemented the `onUri` activation event. The logical thing to do in response to this event is to register a uri handler. However, this does not work because the activation finishes for the front end before the uri handler registration call is invoked. 

### Steps to Reproduce:

1. Install the extension from the referenced PR
2. Add the "onUri" activation event to its package.json
3. Run Theia
4. Open the uri `theia://tylerleonhardt.open-external` in the OS (in the browser, for eample)
5. Observe: there is no message being shown by Theia
6. Open the same uri again
7. Observe: you get a message saying the uri has been handled by the plugin

### Additional Information

- Operating System: Windows
- Theia Version: https://github.com/eclipse-theia/theia/pull/13306

tsmaeder:
#### What it does
The use of a ChannelMultiplexer in `RPCProtocol` was causing the sending of request from the plugin host to be delayed until a sub-channel could be opened. This would lead to the reply for the request that caused the "callback" to the front end to be sent before the request from the plugin host.
This PR removes the channel multiplexer from the `RPCProtocol` and instead just sends the proxy id as an additional parameter. 

Fixes #13960

#### How to test

Since this PR fixes a problem discovered in https://github.com/eclipse-theia/theia/pull/13306, we'll need that on to be merged in order to test that the issue is actually fixed. What we _can_ test right now, though, is that the change does not cause any trouble for the operation of plugins. This means using plugins and observing the logs, etc.

#### Follow-ups
The startup sequence for plugins hosts is still not entirely deterministic. Some of the ideas in https://github.com/eclipse-theia/theia/pull/8437 would still be worth reviving in new PR, even though we don't run into any timing problems, currently.

#### Review checklist

#### Reminder for reviewers


tsmaeder:
@msujew this one would be good to get merged soon to get the code exercised a bit before the next release. @tortmayr is on vacation until the next week.

msujew (approved):
Looks good to me. I can confirm that extension features still work as expected (tested with TypeScript & Jupyter). Less code is always better :)

PR #13982, opened 2024-08-01
Support find widget in notebooks

msujew:
#### What it does

Adds support for a monaco-like find widget to the notebook widget. 

#### How to test

The search widget can be brought up by pressing `Ctrl+F` inside of any area of the notebook widget. Having the Jupyter Keymaps extension installed also allows to bring it up by just pressing `F` outside of an editor area (while the notebook widget is focused).

Multiple test scenarios:
* In general: Test that the widget behaves similarly to the existing find widget for text editors.
* Select text in a monaco editor and press `Ctrl+F`. The selected text should be placed as the initial search value of the find widget.
* Search for text in code. The code should be highlighted (with the current selection getting a special highlighting) and jumping to different matches should work.
* Search using match case, word boundaries and regex should work as expected.
* Search for text in rendered markdown cells. The text should be highlighted directly in the HTML. Jumping should work as well.
* Replacing code in the code editors should work as expected.
* Replacing code in the rendered markdown cells should not be possible (button greyed out).
* Upon changing the editing mode of a markdown cell, the search should update and highlight the now rendered code editor instead (also in reverse).

#### Follow-ups

Known issues:
* Cannot find in webviews
* Cannot replace on rendered markdown
* There is no global undo for the global replace yet. Each cell change needs to be undone individually.
* Performance for huge rendered markdown is quite bad

The current API is pretty suboptimal (especially for finding matches inside of rendered markdown cells). I plan to get back to this later to revise the API and extract a common `FindWidget` that can be used by any widget.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dependency-check-baseline.json
+++ b/dependency-check-baseline.json
@@ -1,3 +1,3 @@
 {
-  "npm/npmjs/@types/qs/6.9.11": "Pending https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/13990"
+  "npm/npmjs/-/advanced-mark.js/2.6.0": "Manually approved"
 }
--- a/packages/core/src/browser/browser.ts
+++ b/packages/core/src/browser/browser.ts
@@ -18,7 +18,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { environment } from '../common';
+import { Disposable, environment } from '../common';
 
 const userAgent = typeof navigator !== 'undefined' ? navigator.userAgent : '';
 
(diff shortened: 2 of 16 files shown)


jbicker (changes requested):
- [x] After step by step replacement => 'NaN of [number]' appeared in find widget, then 'find' is not working at all anymore. 
FandR-NaN
Reproduce: find and replace all occurrences step by step until there is finally 'No Results' in counter field. Change to the find input again and begin to type. 'NaN of [number of occurrences]' appears and highlighting does not work anymore.
- [x] If the last occurrence has been replaced (if available, for example if "great" is replaced with "super great"), it must jump back to the first occurrence. At the moment the search remains at the last occurrence.
- [x] pasting a search term from clipboard into find input does not work
- [x] When playing around with markdown: if markdown is rendered it sometimes doesn't find occurrences when using match case, word boundaries and/or regex, (unfortunately I can't find a reliable way to reproduce it but maybe it has something to do with the currently pretty suboptimal API you've mentioned)
- [x] - [ ] CMD-Option-F resp. Strg-Alt-F does not work (should open 'find and replace')

msujew:
Thanks @jbicker, I believe I fixed every discovered issue except for one:

> CMD-Option-F resp. Strg-Alt-F does not work (should open 'find and replace')

This one isn't supported in VS Code (in notebooks) and I'm not sure adding a keybinding for this is a good idea. It might accidentally break some keybindings that I'm not aware of.

jbicker (approved):
LGTM. Everything works as expected.

PR #14016, opened 2024-08-06
Fix notebook output scrolling and text rendering

msujew:
#### What it does

Fixes two issues related to scrolling in notebooks:

1. When scrolling down while the cursor hovers over output text, the scrolling is no longer interrupted.
2. When rendering too much text, the "scrollable element" is actually scrollable and not just a very long text element.

#### How to test

1. Create a python notebook file and create this cell:
```py
for i in range(1,100): 
    print('g' + str(i))
```
2. Run the cell. Hover over some text within the generated output and scroll down. The content within the notebook should successfully scroll down.
3. Click on the `scrollable element` link in the bottom of the rendered text. It should create a scrollable element that is similar in size to the original rendered text.

This PR also fixes a few minor rendering issues related to missing CSS in output webviews.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/style/index.css
+++ b/packages/core/src/browser/style/index.css
@@ -239,7 +239,8 @@ blockquote {
     -o-user-select: none;
 }
 
-:focus {
+/* Since an iframe has its own focus tracking, we don't show focus on iframes */
+:focus:not(iframe) {
     outline-width: 1px;
     outline-style: solid;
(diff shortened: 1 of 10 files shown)


jbicker (approved):
Works like a charm!

jbicker (approved):

Issue #14026, opened 2024-08-08
Make Theia work on NodeJS 22.x


tsmaeder:
### Feature Description:
Currently, Theia does not build against Node 22.x, there are build errors when building native packages, for example nswf. 

prerequisites: 

- [ ] https://github.com/eclipse-theia/theia/issues/14015

JonasHelming:
@tsmaeder Is this done now that the prerequisite is done?

tsmaeder:
After an update of node-pty to the latest beta version (1.1.0-beta27), Theia builds. Since node-pty is used for our whole process support, this needs serious testing, though.

PR #14027, opened 2024-08-08
Drop support for Node 16.x

### LINKED ISSUE #13956: Drop support of Node 16
According to https://nodejs.org/en/about/releases/, 
- Node 16 is not maintained anymore.
- The next active release for Node (22) is planned for October this year

We should then drop the support from Node 16 in our builds (https://github.com/eclipse-theia/theia/blob/master/doc/runtime-policy.md).

Node 16 support blocked the update for jsdom for example (see https://github.com/eclipse-theia/theia/issues/13943).

tsmaeder:
#### What it does
Fixes #13956

We're dropping support for Node 16 targets because it's out of maintenance and we'd like to update some packages to versions that are incompatible with Node 16.

Node that support for Node 22 is not working yet due to some incompatible packages we need to replace first.

#### How to test
Code needs to build on all platforms and do a general "smoke" test of programming functionality.

#### Follow-ups
- [ ] https://github.com/eclipse-theia/theia/issues/14026

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -20,10 +20,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
 
-      - name: Use Node.js 18.x
+      - name: Use Node.js 20.x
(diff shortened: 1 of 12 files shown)


msujew (approved):
LGTM 👍

PR #14030, opened 2024-08-09
Fix 429 errors on OVSX requests

msujew:
#### What it does

After https://github.com/eclipse-theia/theia/pull/13825 we started having a few 429 errors (and adopters as well) when downloading plugins from extension packs. 

This change correctly propagates the OVSX rate limit through all layers of the application including the runtime. This can be seen in the `ovsx-http-client.ts` file changes. This should ensure that users never hit 429 errors from OVSX. If they do, they can use a lower rate limit via a CLI option.

#### How to test

I've upped the rate limit in the CI - the CI should run successfully.

Also, simply running the `theia download:plugins` CLI command should run without errors.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/ovsx-client/src/ovsx-http-client.ts
+++ b/dev-packages/ovsx-client/src/ovsx-http-client.ts
@@ -47,15 +47,27 @@ export class OVSXHttpClient implements OVSXClient {
     }
 
     protected async requestJson<R>(url: string): Promise<R> {
-        await this.rateLimiter.removeTokens(1);
-        const context = await this.requestService.request({
-            url,
-            headers: { 'Accept': 'application/json' }
-        });
-        if (context.res.statusCode === 429) {
-            console.warn('OVSX rate limit exceeded. Consider reducing the rate limit.');
+        const attempts = 5;
+        for (let i = 0; i < attempts; i++) {
+            // Use 1, 2, 4, 8, 16 tokens for each attempt
+            const tokenCount = Math.pow(2, i);
+            await this.rateLimiter.removeTokens(tokenCount);
+            console.log('Sending request at: ' + Date.now());
+            const context = await this.requestService.request({
+                url,
+                headers: { 'Accept': 'application/json' }
+            });
+            if (context.res.statusCode === 429) {
+                // If there are still more attempts left, retry the request with a higher token count
+                if (i < attempts - 1) {
+                    continue;
+                } else {
+                    console.warn('OVSX rate limit exceeded. Consider reducing the rate limit.');
+                }
+            }
+            return RequestContext.asJson<R>(context);
         }
-        return RequestContext.asJson<R>(context);
+        throw new Error('Failed to fetch data from OVSX.');
     }
 
     protected buildUrl(url: string, query?: object): string {
--- a/dev-packages/ovsx-client/src/ovsx-http-client.ts
+++ b/dev-packages/ovsx-client/src/ovsx-http-client.ts
@@ -52,17 +52,15 @@ export class OVSXHttpClient implements OVSXClient {
             // Use 1, 2, 4, 8, 16 tokens for each attempt
             const tokenCount = Math.pow(2, i);
             await this.rateLimiter.removeTokens(tokenCount);
-            console.log('Sending request at: ' + Date.now());
             const context = await this.requestService.request({
                 url,
                 headers: { 'Accept': 'application/json' }
             });
             if (context.res.statusCode === 429) {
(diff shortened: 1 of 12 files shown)


tsmaeder:
Could you explain a bit how #13825 lead to the 429 errors? After all, we did not rate limit before and we didn't after. Did we just increase the number of requests?

msujew:
> Did we just increase the number of requests?

@tsmaeder Yes, exactly :) After #13825, we perform 2 requests for every `query` request - one with and one without the `targetPlatform` parameter. So even if the outer loop performs rate limiting, we still hit the ovsx servers twice as hard and therefore get 429 responses in return.

dev-packages/ovsx-client/src/ovsx-http-client.ts
```diff
@@ -43,10 +47,20 @@ export class OVSXHttpClient implements OVSXClient {
     }
 
     protected async requestJson<R>(url: string): Promise<R> {
-        return RequestContext.asJson<R>(await this.requestService.request({
-            url,
-            headers: { 'Accept': 'application/json' }
-        }));
+        for (let i = 0; i < 10; i++) {
```
tsmaeder, line 50: I'm not sure about the logic here: repeating the request when we're already running into the rate limit seems counterproductive. Wouldn't the right answer be to abort the task and for the client of this API to adjust the rate limit?

dev-packages/ovsx-client/src/ovsx-http-client.ts
```diff
@@ -43,10 +47,20 @@ export class OVSXHttpClient implements OVSXClient {
...
+            await this.rateLimiter.removeTokens(1);
+            const context = await this.requestService.request({
+                url,
+                headers: { 'Accept': 'application/json' }
+            });
+            if (context.res.statusCode === 429) {
```
tsmaeder, line 56: Shouldn't we at least output some diagnostic in this case? Can we determine something from the response, like "retry-after" or similar?

dev-packages/ovsx-client/src/ovsx-http-client.ts
msujew: Yes, I've removed the loop and added a warning/log for this.

dev-packages/ovsx-client/src/ovsx-http-client.ts
msujew: That'd be great. Unfortunately, Open-VSX isn't really helpful:
```ts
{
  server: 'nginx',
  date: 'Mon, 12 Aug 2024 18:49:43 GMT',
  'content-type': 'application/json',
  'content-length': '35',
  connection: 'close',
  vary: 'Accept-Encoding',
  'x-rate-limit-retry-after-seconds': '0', // <-- ???
  'access-control-expose-headers': 'X-Rate-Limit-Retry-After-Seconds, X-Rate-Limit-Remaining',
  'access-control-allow-origin': '*'
}
```

dev-packages/ovsx-client/src/ovsx-http-client.ts
msujew: `RequestContext.asJson` already throws an error in that case. All non `2xx` codes will be thrown as errors in there.

dev-packages/ovsx-client/src/ovsx-http-client.ts
tsmaeder: 😞

tsmaeder:
I just ran `yarn download:plugins` with the PR branch and I'm getting three 429 errors. Either we're still not rate-limiting properly or rate-limiting is broken on the open-vsx side. 🤷‍♂️

msujew:
@tsmaeder I've added exponential backoff to our requests. I think this should work around the issues of open-vsx reasonably well.

dev-packages/ovsx-client/src/ovsx-http-client.ts
```diff
@@ -47,15 +47,27 @@ export class OVSXHttpClient implements OVSXClient {
...
+            await this.rateLimiter.removeTokens(tokenCount);
+            console.log('Sending request at: ' + Date.now());
+            const context = await this.requestService.request({
+                url,
+                headers: { 'Accept': 'application/json' }
+            });
+            if (context.res.statusCode === 429) {
+                // If there are still more attempts left, retry the request with a higher token count
+                if (i < attempts - 1) {
+                    continue;
```
tsmaeder, line 63: As a last nitpick, can we get a warning if we ran into 429's even if we eventually succeed? IMO it's important info also for open-vsx.

tsmaeder (approved):

PR #14031, opened 2024-08-09
Creating new untitled notebook doesn't work

dhuebner:
#### What it does

Right now, when the new notebook is created with `File -> New File... -> New Jupyter Notebook` the opened editor is not usable. Many errors are logged saying that a cell document is not loaded.
This PR fixes the issue.

#### How to test

Create a new untitled Notebook using  `File -> New File... -> New Jupyter Notebook`.
Switch the cell language to Python. Executing Python code (e.g. `print("Test me")`)  should work, outputs should be visible. 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/notebooks/notebook-documents-main.ts
+++ b/packages/plugin-ext/src/main/browser/notebooks/notebook-documents-main.ts
@@ -21,7 +21,7 @@ import { NotebookModelResolverService } from '@theia/notebook/lib/browser';
 import { NotebookModel } from '@theia/notebook/lib/browser/view-model/notebook-model';
 import { NotebookCellsChangeType } from '@theia/notebook/lib/common';
 import { NotebookMonacoTextModelService } from '@theia/notebook/lib/browser/service/notebook-monaco-text-model-service';
-import { MAIN_RPC_CONTEXT, NotebookCellsChangedEventDto, NotebookDataDto, NotebookDocumentsExt, NotebookDocumentsMain } from '../../../common';
+import { MAIN_RPC_CONTEXT, NotebookCellsChangedEventDto, NotebookDataDto, NotebookDocumentsExt, NotebookDocumentsMain, NotebookRawContentEventDto } from '../../../common';
 import { RPCProtocol } from '../../../common/rpc-protocol';
 import { NotebookDto } from './notebook-dto';
 import { MonacoEditorModel } from '@theia/monaco/lib/browser/monaco-editor-model';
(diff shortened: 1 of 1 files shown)


msujew (changes requested):
I'm not sure this change actually fixes anything - it breaks the support for supplying custom `content` to any newly created notebook and instead inserts a plaintext cell. See also https://github.com/eclipse-theia/theia/pull/13696.

Instead, we should figure out why the errors are logged in the first place.

dhuebner:
@msujew 
> I'm not sure this change actually fixes anything

The `File -> New File... -> New Jupyter Notebook` doesn't work currently, the new created editor is not usable at all. With the fix provided it is usable and you can change the cell type to what ever you want. 

> Instead, we should figure out why the errors are logged in the first place.

Errors are logged because the editor creates a new cell, even thought the model doesn't contain one, and the Monaco editor inside the cell tries to open a cell document that is not created due to a missing cell in the model. 

> it breaks the support for supplying custom content to any newly created notebook

I need to have a look here #3696 maybe adding a default content supplier does the trick.

msujew:
@dhuebner I see. I can reproduce the errors, but it's definitely not a real fix for the issue. We should ensure that the cells are correctly created on the plugin host (the errors are coming from there).

dhuebner:
@msujew 
The problem is that the changes applied from the `options.content` in `$tryCreateNotebook()` are not properly propagated to the proxy. So the additionally added cells are missing  in the notebook plugin document.
The comment saying `// apply content changes... slightly HACKY -> this triggers a change event` assumes the change events a forwarded to the proxy, probably similar as in inside the `handleNotebooksAdded()`, but nobody handles the change events and I don't know how it was working before.
One way to solve this is to manually send related change events to the proxy, or register the new created notebook by calling `handleNotebooksAdded()`. Calling `handleNotebooksAdded()` seems to be strange to me as it is called when a notebook is actually opened in the editor, but we are inside the `tryCreateNotebook()`. The new diff contains the first solution by sending corresponding change events manually.

msujew (approved):
Alright, looks good to me. Thanks :)

msujew:
@dhuebner Note that the linter is failing and I cannot merge this PR.

dhuebner:
@msujew 
I fixed missing semicolons. Eslint in vs-code doesn't work for me, but I know was working before. Output says: 
`Failed to load plugin '@theia' declared in 'dev-packages/private-eslint-plugin/.eslintrc.js » ../../configs/build.eslintrc.json » ./base.eslintrc.json': Unexpected end of JSON input`

Issue #14034, opened 2024-08-12
Preference/Setting issue when using the Redhat Java plugin


safisa:
### Bug Description:
Hi,
In latest Theia there is a regression in the preference/setting view when the Redhat Java plugin is installed, all the Java settings were shown under one main category under Extenions--> Java, but now all the java sub-categories are rendered under the Extenions parent, making it difficult to know that those categories belong to the Java extension.

Maybe this was caused after this PR: https://github.com/eclipse-theia/theia/pull/13819

### Steps to Reproduce:

1. install Redhat java plugin (1.33)
2. open setting
3. you will see many new categories under the Extensions category, instead of adding only one Java category as before.

### Additional Information

- Operating System: Mac
- Theia Version: 1.52

tsmaeder:
Invoking @msujew for comment.

msujew:
Can reproduce, looking at it right now :)

PR #14036, opened 2024-08-12
Fix preference tree for plugins

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/14034

I've figured out three configurations in which plugin preferences can appear:
1. Single array of preferences (no title) - generate a single entry for the plugin in the tree view
2. Single array of preferences (with a title) - generate a single entry for the plugin with the given title
3. Nested array of preferences - generate a parent entry for the plugin with sub entries for each array of preferences.

#### How to test

1. Install an extension that fits into the 3rd category. For example, the RedHat Java extension.
2. Open the settings view and expand the `Extensions` category
3. Assert that the tree nodes are rendered as expected.

It'd be best to compare the tree to whatever is rendered in VS Code.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/browser/views/preference-tree-widget.tsx
+++ b/packages/preferences/src/browser/views/preference-tree-widget.tsx
@@ -24,6 +24,7 @@ import {
 } from '@theia/core/lib/browser';
 import React = require('@theia/core/shared/react');
 import { PreferenceTreeModel, PreferenceTreeNodeRow, PreferenceTreeNodeProps } from '../preference-tree-model';
+import { Preference } from '../util/preference-types';
 
 @injectable()
 export class PreferencesTreeWidget extends TreeWidget {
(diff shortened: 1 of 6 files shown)


packages/preferences/src/browser/views/preference-tree-widget.tsx
```diff
@@ -50,13 +51,28 @@ export class PreferencesTreeWidget extends TreeWidget {
...
         this.updateScrollToRow();
     }
 
+    protected isVisibleNode(row: PreferenceTreeNodeRow): boolean {
+        const node = row.node;
+        if (Preference.TreeNode.isTopLevel(node)) {
+            return true;
+        }
+        let parent = node.parent;
+        while (parent) {
```
tsmaeder, line 67: Why not just do it recursively? 
```
if (isTopLevelNode(node)) {
   return true;
}
return isVisible(node.parent) && !isCollapsed(node.parent);
```

tsmaeder (commented):
So if I understand correctly, the policy is this:
- if an extensions contributes multiple property section, they are shown under a heading consisting of the extensions display name
- if an extension only show one extension, that section (with it's contributed title) is shown under "Extensions".

msujew:
>  if an extensions contributes multiple property section, they are shown under a heading consisting of the extensions display name.  if an extension only show one extension, that section (with it's contributed title) is shown under "Extensions".

Exactly, yes :)

tsmaeder (approved):

Issue #14045, opened 2024-08-14
Editor status bar items stay when switching to a custom editor


jonah-iden:
### Bug Description:

![grafik](https://github.com/user-attachments/assets/d6982054-125c-4c46-af39-d7de95ed9b15)

When opening a custom editor like a webview or the notebook editor, the status bar items stay even though the editor they belong to is not visible anymore. In VSCode it disappears when the editor is not visible anymore.

### Steps to Reproduce:

1. Open some text file. The status bar should contain information about cursor position, encoding etc
2. Open a custom editor like a notebook or webview based (i also tested with the https://open-vsx.org/extension/redhat/vscode-extension-bpmn-editor)
3. The text editor is not visible anymore, but the status bar items still stay

### Additional Information

- Operating System: Windows/Firefox
- Theia Version: 1.52.0

dhuebner:
See also  #14190

PR #14046, opened 2024-08-14
Add notebook selected cell status bar item and center selected cell command

jonah-iden:
#### What it does

Adds "Cell X of Y" status bar text to indicate the current selected cell. Clicking it should center the selected cell in the viewport.
Only difference to vscode is that currently in theia it is possible to not have a cell selected, so the statu bar item shows cell 0 in that case.

#### How to test

1. Open a notebook
2. See the status bar item showing the current selected cell
3. clicking the item shoudl center the selected cell
4. closing the notebook should remove the status bar item

#### Follow-ups

there seems to still be a small bug with text editors and their status bar items. See https://github.com/eclipse-theia/theia/issues/14045

#### Review checklist

#### Reminder for reviewers


msujew (commented):
I've noticed that when opening a notebook and not selecting a cell, the status bar shows `Cell of <number>`. I think we should align to VS Code wrt to the selection behavior. Simply select the first cell by default, and don't allow to "unselect" a cell, meaning that there should always be a selected cell.

msujew (approved):
Looks good to me now, thanks 👍

Issue #14051, opened 2024-08-15
Tab Maximize issue on custom editors


safisa:
### Bug Description:
Hi,
I think there is an issue in tab Maximize feature on custom editors. When using the maximized tab on a dirty tab, it will open it in dirty mode (modified state), but the issue is that if the user closes the maximized modified tab, and in the save modified file confirmation popup, select not to save, once the file is opened again it will open as modified !!!

### Steps to Reproduce:

1. open a custom editor file (e.g cat plugin)
2. do some changes
3. maximize the tab
4. see that it is still a dirty file (bullet on the tab)
5. close that tab
6. in the save confirmation popup, choose not to save
7. open the same file again, you will see that it is still in a modified state! it will keep like this until you save the file!

### Additional Information

- Operating System: Mac
- Theia Version: 1.52

JonasHelming:
@dhuebner If I remember correctly, did you review or do a change in this area recently?

dhuebner:
@JonasHelming 
Not sure I did something in the maximize editor workflow, but I can have a look.

JonasHelming:
:-) Not with maximizing, but with the save behavior or am I mistaken? Anyways, thank you for having a look!

msujew:
@JonasHelming I've recently refactored the saving mechanism in https://github.com/eclipse-theia/theia/pull/13683 :)

msujew:
See https://github.com/eclipse-theia/theia/pull/14062 for a fix.

PR #14062, opened 2024-08-21
Correctly revert saveable on widget close

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/14051

Fixes a bug where we always passed `false` as the `revert` flag when closing a widget. Now correctly checks whether the same saveable exists somewhere else: If it exists, do not revert. If it is the last instance of the saveable, do revert.

#### How to test

1. Open a text file and modify it.
2. Split the text file (i.e. create a second widget for it).
3. Close one of the editors - the widget should just close
4. Close the other editor and click "Don't save".
5. Open the file again - assert that the changes have been reverted as expected.

#### Review checklist

#### Reminder for reviewers


tortmayr (approved):
LGTM 👍🏼 
After testing I can confirm that the revert now works as expected.

PR #14065, opened 2024-08-22
Download json schema catalog at build-time instead of run-time

### LINKED ISSUE #11881: [core] Consider getting the `schemastore.azurewebsites.net/api/json/catalog.json` at build time
### Feature Description:

Whenever I start Theia from the sources (electron, browser) or TheiaBluePrint (1.31.1 (1.31.1.85)), getting the default preference schema fails. The fetched JSON size has increased during the years, and getting the schema never completes in 500ms.

From the sources:
```
logger-protocol.ts:113 2022-11-18T13:42:37.774Z root ERROR DefaultJsonSchemaContribution.registerSchemas is taking more than 500.0 ms, new schemas are ignored.
```

From the bundled electron app:
```
logger-protocol.ts:113 2022-11-18T13:48:50.330Z root ERROR h.registerSchemas is taking more than 500.0 ms, new schemas are ignored.
```

Either increase the timeout or get the schema at build time. The current implementation makes no sense. Note that getting the schema at build time has one drawback, the most recent schema won't be fetched on app load, so it could be outdated. In one of the downstream projects, we get the schema at build time, which works well.

tsmaeder:
#### What it does
This fixes the problem where we run into timeout at front-end start time because downloading the json schema catalog takes longer than expected. We now download the catalog as part of the core package build and just `require()` the json file at run time.

Fixes #11881

#### How to test
Build Theia from source and make sure that there are no errors at startup. Also check that json files that are on the schemastore.org catalog have proper code assist according to their schema.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 
 ## Unreleased
+- [core] Download json schema catalog at build-time - [#14065](https://github.com/eclipse-theia/theia/pull/14065/) - Contributed on behalf of STMicroelectronics
 
 <a name="breaking_changes_1.53.0">[Breaking Changes:](#breaking_changes_1.53.0)</a>
 - [dependencies] Updated electron to version 30.1.2 - [#14041](https://github.com/eclipse-theia/theia/pull/14041) - Contributed on behalf of STMicroelectronics
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -197,6 +197,7 @@
     "src"
   ],
   "scripts": {
+    "prebuild": "yarn download:json-schema",
     "build": "theiaext build",
     "clean": "theiaext clean",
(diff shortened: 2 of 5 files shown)


tsmaeder:
@msujew I seem to have a "works on my machine" problem here. Maybe hold off on the review until I have fixed it.

tsmaeder:
@msujew turns out our CI build script does not actually invoke `build` on the individual packages 🤷. Should be fine now.

msujew (approved):
Looks good to me and I can confirm that the feature still works as expected 👍

tsmaeder:
@msujew I added a flag that overwrites the file if you already have a `catalog.json` downloaded. If you think this is bogus, we can always revert the PR.

PR #14098, opened 2024-08-30
Playwright API for Notebooks

dhuebner:
#### What it does

Adds Playwright page objects to interact with a notebook Editor and its cells.
Following actions are supported: 
- Open a Notebook Editor
- Select a Kernel and read selected kernel name. 
- Adding a new Code/Markdown Cell
- Read Cell mode (e.g. python, markdown etc.)
- Getting cell editor text and adding text to the cell editor
- Execute a cell
- Read cell output

All of the above listed functionalities are tested in [theia-notebook-editor.test.ts](https://github.com/eclipse-theia/theia/compare/master...dhuebner:theia:dhuebner/notebook-ui-tests?expand=1#diff-1d3cea4eca8b8d6310270ee81d60b40e12ec3df2b876ad5fe2b49a5bd72b5459) playwright test.

#### How to test

Run `theia-notebook-editor.test.ts` mentioned above.

#### Follow-ups

Planned is to add other handy features like e.g. reading cell execution count and interaction with cell's toolbar.  

#### Review checklist

#### Reminder for reviewers


dhuebner:
The failing CI tests might indicates that a kernel installation is missing. I can check the Python setup for the corresponding GH action. Or maybe some expert could help faster?

examples/browser/package.json
```diff
@@ -19,6 +19,15 @@
...
   },
+  "theiaPluginsDir": "../../plugins",
+  "theiaPlugins": {
+    "eclipse-theia.builtin-extension-pack": "https://open-vsx.org/api/eclipse-theia/builtin-extension-pack/1.88.1/file/eclipse-theia.builtin-extension-pack-1.88.1.vsix",
+    "ms-toolsai.jupyter": "https://open-vsx.org/api/ms-toolsai/jupyter/2024.6.0/file/ms-toolsai.jupyter-2024.6.0.vsix",
+    "ms-python.python": "https://open-vsx.org/api/ms-python/python/2024.12.3/file/ms-python.python-2024.12.3.vsix"
+  },
+  "theiaPluginsExcludeIds": [
+    "ms-python.vscode-pylance"
+  ],
```
msujew, line 30: Suggestion: These should be put into the root package.json, which already includes the builtin extension pack.

dhuebner: @msujew Sure I can move the two additional ones to the root. First I wanted to put them into the root, but I was not sure whether this change will affect the other non-playwright related jobs. And maybe people have some concerns having this extensions in the dev environment.

msujew: It's IMO pretty unexpected to see the same plugin pack multiple times in the same repo. In that case, we shouldn't add `eclipse-theia.builtin-extension-pack` in this package.json, but use the `download:plugins` script from the root in addition to the one defined in the browser package.

dhuebner: Moved to the root package.json

dhuebner:
Rebased on master and squashed commits.

msujew (commented):
This looks pretty good to me. However, I would like to get a review from someone more familiar with Playwright before merging this @JonasHelming do you think someone from ES could spare some time to review this?

examples/playwright/src/tests/theia-notebook-editor.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    test.afterEach(async () => {
+        if (editor) {
+            await editor.closeWithoutSave();
+        }
+    });
+
+    test('kernels are installed', async () => {
+        editor = await app.openEditor('sample.ipynb', TheiaNotebookEditor);
+        await editor.tabLocator().click(); // force activate
```
planger, line 46: I think it might be better to move that into the `TheiaNotebookEditor` page object, if this is really necessary. Otherwise every test would have to be aware that this needs to be called before subsequent interactions.

examples/playwright/src/theia-notebook-cell.ts
```diff
@@ -0,0 +1,212 @@
...
+    async isCodeCell(): Promise<boolean> {
+        const classAttribute = await this.mode();
+        return classAttribute !== 'markdown';
+    }
+
+    /**
+     * @returns The mode of the cell, e.g. 'python', 'markdown', etc.
+     */
+    async mode(): Promise<string> {
+        this.locator.waitFor({ state: 'visible' });
```
planger, line 55:
```suggestion
        await this.locator.waitFor({ state: 'visible' });
```

examples/playwright/src/theia-notebook-cell.ts
```diff
@@ -0,0 +1,212 @@
...
+        if (linesCount === undefined) {
+            return undefined;
+        }
+        for (let line = 1; line <= linesCount; line++) {
+            const lineText = await this.monacoEditor.textContentOfLineByLineNumber(line);
+            if (lineText === undefined) {
+                break;
+            }
+            lines.push(lineText);
+        }
```
planger, line 82: Would it make sense to move this into the `TheiaMonacoEditor`? I guess this functionality could be useful in general for the text editor and would then be easier to test and maintain, if it is at a single place.

examples/playwright/src/theia-notebook-cell.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * Adds text to the editor of the cell.
+     * @param text  The text to add to the editor.
+     * @param lineNumber  The line number where to add the text. Default is 1.
+     */
+    async addEditorText(text: string, lineNumber: number = 1): Promise<void> {
+        const line = await this.editor.lineByLineNumber(lineNumber);
+        await line?.click();
+        await this.page.keyboard.type(text);
```
planger, line 94: As above, I think it could make sense to move this behavior into the `TheiaMonacoEditor`.

examples/playwright/src/theia-notebook-editor.ts
```diff
@@ -0,0 +1,164 @@
...
+        }
+        return this.notebookToolbar().locator.locator('#kernel-text').innerText();
+    }
+
+    /**
+     *  Allows to select a kernel using toolbar item.
+     * @param kernelName  The name of the kernel to select.
+     */
+    async selectKernel(kernelName: string): Promise<void> {
+        // TODO should we use NotebookCommands.SELECT_KERNEL_COMMAND.id? Need a dependency...
```
planger, line 75: Good question, I think I tend to avoid this dependency but rather introduce a namespace with constants here in the Playwright page object library.

examples/playwright/src/theia-notebook-editor.ts
```diff
@@ -0,0 +1,164 @@
...
+        const widget = await this.page.waitForSelector(qInput.selector, { timeout: 5000 });
+        if (widget && !await qInput.isOpen()) {
+            throw new Error('Failed to trigger kernel selection');
+        }
+        await qInput.type(kernelName, true);
+        await qInput.hide();
+    }
+
+    async availableKernels(): Promise<string[]> {
+        // TODO should we use NotebookCommands.SELECT_KERNEL_COMMAND.id? Need a dependency...
```
planger, line 87: As above, I'd have a slight preference to introduce a namespace with constants in this page object file.

examples/playwright/src/theia-quick-command-palette.ts
```diff
@@ -78,4 +78,14 @@ export class TheiaQuickCommandPalette extends TheiaPageObject {
         }
         return command.$('.monaco-list-row.focused .monaco-highlighted-label');
     }
+
+    async visibleItems(): Promise<ElementHandle<SVGElement | HTMLElement>[]> {
```
planger, line 82: I think, we should add a dedicated test for that new method in `examples/playwright/src/tests/theia-quick-command.test.ts`.

packages/notebook/src/browser/view/notebook-main-toolbar.tsx
```diff
@@ -152,17 +152,17 @@ export class NotebookMainToolbar extends React.Component<NotebookMainToolbarProp
 
     override render(): React.ReactNode {
         const menuItems = this.getMenuItems();
-        return <div className='theia-notebook-main-toolbar'>
+        return <div className='theia-notebook-main-toolbar' id='notebook-main-toolbar'>
```
planger, line 155: Is it save to use a static `id` here? Couldn't there be more than one toolbars in the DOM, eventually leading to non-unique IDs in the DOM? If yes, it might be better to use `data-<xyz>` attributes and locate them in the context of a parent?

planger (commented):
Thank you very much! This looks great. :+1: 
I just added a few inline uncritical comments.

examples/playwright/configs/playwright.ci.config.ts
```diff
@@ -21,6 +21,7 @@ const ciConfig: PlaywrightTestConfig = {
     ...baseConfig,
     workers: 1,
     retries: 2,
+    timeout: 30 * 1000,
```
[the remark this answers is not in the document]
dhuebner: I think we can remove it again.

examples/playwright/src/tests/theia-notebook-editor.test.ts
dhuebner: It was an attempt to fix flakiness on first run. I didn't found out why, but right after the start and opening an editor the focus jumps to the browser address field. Now I think it was because the "Workspace trust" message dialog was popping-up. It should no longer be necessary as the check is disabled in the example project package.json. I will remove this lines, let's see if the tests are passing in CI.

examples/playwright/src/theia-notebook-cell.ts
dhuebner: Nice catch!

examples/playwright/src/theia-quick-command-palette.ts
dhuebner: Test is added

packages/notebook/src/browser/view/notebook-main-toolbar.tsx
dhuebner: I think it is okay to use a static `id` for the **main** toolbar. Should be only one, right? In case there are other sibling toolbars with the same `id` it can be handled as a list. For react one will need to adjust the `key` property.

msujew: There is one toolbar **per editor**, meaning that it's not unique. I.e. if you have two notebook editors side by side, it will no longer be unique.

examples/playwright/configs/playwright.ci.config.ts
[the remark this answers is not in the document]
dhuebner: Now, I know why it is explicitly listed. CI config extends the `baseConfig` where the time out is 60sec which is way too much IMHO.

packages/notebook/src/browser/view/notebook-main-toolbar.tsx
dhuebner: They will be unique if using `locator` API instead of `selector`, what I do in notebook editor page object.  Locator gives you the Notebook editor element under test, hence all subsequent queries are applied on the editor element and not the whole page.

dhuebner:
I have no idea why it started to fail now.
test-failed-1

dhuebner:
Tests are running now. The cause was that a wrong Python 3.10 version was picked by default, where it should be Python 3.11. Fixed it now by requesting 3.11.* explicitly. 

@planger 
Thanks for your detailed feedback!
I've applied all of your suggestions except the one with the `id` for the main toolbar. I think  it is not necessary and having a dynamic `id` will make the querying by ID inside the test more complex. Could you please review the current state?

dhuebner:
Btw. the playwright reports are now created also on successful job run and are stored in job's Artifacts for two days. See for example: https://github.com/eclipse-theia/theia/actions/runs/10943260523?pr=14098

planger (approved):
Thanks a lot for the revision. Looks great to me!

dhuebner:
@JonasHelming @msujew 
Then this PR can be merged. I'm looking forward to see it in the upcoming `.next` release.

msujew:
@dhuebner @planger FYI I would wait to merge this after https://github.com/eclipse-theia/theia/pull/14234 has been merged. Just so we can fix unanticipated breakage in the playwright tests in this PR already.

msujew:
@dhuebner Do you mind rebasing? https://github.com/eclipse-theia/theia/pull/14234 has been merged.

dhuebner:
@msujew 
PR is ready to merge now. I've added an additional test to cover the new "split cell" action.

msujew:
@dhuebner Great, thank you!

PR #14125, opened 2024-09-04
Switch from `browserfs` to `zenfs`

robertjndw:
#### What it does

This PR builds on top of the work that was done in PR #12853.
As `browserfs` was deprecated in March 24 (see [original repo](https://github.com/jvilk/BrowserFS?tab=readme-ov-file)), it is beneficial to switch to the actively maintained fork `zenfs` (see [here](https://github.com/zen-fs)).

In addition to the active maintenance, the fork also introduces some new features and an improved API (e.g. promise based instead of callback). The new features will simplify the implementation of the missing parts in the [`BrowserFSFileSystemProvider`](packages/filesystem/src/browser-only/browserfs-filesystem-provider.ts)

#### How to test

Similarly to the previously mentioned PR #12853.

#### Follow-ups

The unimplemented methods of the old PR still exist in this one and will be implemented in a future PR.

#### Review checklist

#### Reminder for reviewers


packages/filesystem/src/browser-only/browserfs-filesystem-initialization.ts
```diff
@@ -14,48 +14,29 @@
...
 export const BrowserFSInitialization = Symbol('BrowserFSInitialization');
 export interface BrowserFSInitialization {
-    createMountableFileSystem(): Promise<MountableFileSystem>
-    initializeFS: (fs: FSModule, provider: BrowserFSFileSystemProvider) => Promise<void>;
+    createMountableFileSystem(): Promise<void>
+    initializeFS: (provider: BrowserFSFileSystemProvider) => Promise<void>;
 }
 
 @injectable()
 export class DefaultBrowserFSInitialization implements BrowserFSInitialization {
```
sdirix, line 29: We should rename these symbols, interfaces, classes etc.

packages/filesystem/package.json
```diff
@@ -10,7 +10,8 @@
     "@types/tar-fs": "^1.16.1",
     "async-mutex": "^0.3.1",
     "body-parser": "^1.18.3",
-    "browserfs": "^1.4.3",
+    "@zenfs/core": "^0.16.4",
+    "@zenfs/dom": "^0.2.14",
```
sdirix, line 14: This change (and the suggested renames) need to be listed as breaking changes in the Changelog

examples/api-samples/src/browser-only/filesystem/example-filesystem-initialization.ts
```diff
@@ -19,15 +19,15 @@ import { inject, injectable, interfaces } from '@theia/core/shared/inversify';
...
+import { fs } from '@zenfs/core';
 
 @injectable()
 export class ExampleBrowserFSInitialization extends DefaultBrowserFSInitialization {
 
     @inject(EncodingService)
     protected encodingService: EncodingService;
 
-    override async initializeFS(fs: FSModule, provider: BrowserFSFileSystemProvider): Promise<void> {
+    override async initializeFS(provider: BrowserFSFileSystemProvider): Promise<void> {
```
sdirix, line 30: I did not look into zenfs. Is this the expected usage? Seems weird to get rid of the filesystem as a parameter as it seems to lose flexibility.

sdirix (commented):
Thanks for the work! I ran it locally and ZenFS worked as expected for me.

I have some minor comments, please have a look

sdirix:
I had a closer look at `@zenfs/core`. I think it should be discussed whether we *really* want to switch to this framework or rather check whether we should integrate a more popular alternative down the line.

examples/api-samples/src/browser-only/filesystem/example-filesystem-initialization.ts
james-pre: @sdirix,

> Is this the expected usage? 

Yes, using the `fs` import/export is correct.

>  Seems weird to get rid of the filesystem as a parameter as it seems to lose flexibility.

This has simplified the internal API a lot since ZenFS doesn't need to worry about handling instances of the emulation layer. ZenFS' [`configure`](https://zenfs.dev/core/functions/configure.html) allows a lot of flexibility. 

Additionally, if you'd like to continue using dependency injection, you can still do so. For example:

```ts
import type { fs } from '@zenfs/core';

export type FSModule = typeof fs;

export class ExampleFSInitialization {
	// ...
	initializeFS(fs: FSModule) // ...
	// ...
}
```

However, I would not recommend this since it adds needed complexity. It may be easier to allow the user to pass in the configuration object instead.

james-pre:
> I had a closer look at `@zenfs/core`. I think it should be discussed whether we really want to switch to this framework or rather check whether we should integrate a more popular alternative down the line.

At the time of writing, `@zenfs/core` has 4948 weekly downloads compared to `browserfs` with 5818. That is a difference of only 16%.

james-pre:
@sdirix,

I don't mean to pressure the team, but ZenFS is now more popular than BrowserFS (due to [`@dsnp/parquetjs`](https://www.npmjs.com/package/@dsnp/parquetjs), which has ~36000 weekly downloads), with ~8800 weekly downloads to BrowserFS' ~5900 at the time of writing. Additionally, `@zenfs/core` 1.0.0 will be released in a couple of days, so the API will be guaranteed to be stable.

@robertjndw You may want to update the version of `@zenfs/core` to 1.0.0 once it comes out. I also plan on updating the other packages (e.g. `@zenfs/dom`) so the referenced core version is `^1.0.0`.

sdirix:
@james-pre Thank you for your work on `zenfs`.

We are currently considering several options, including a switch to `zenfs` or leveraging [OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system) directly. We'll keep you informed as we finalize the decision.

msujew:
I would assume we can close this PR with the addition of https://github.com/eclipse-theia/theia/pull/14263?

sdirix:
> I would assume we can close this PR with the addition of #14263?

Yes, removing the dependency altogether is the better approach for a framework like Theia. Any user who would like to integrate `zenfs` can still do so by implementing their own `FileSystemProvider`.

Issue #14138, opened 2024-09-05
Support the `workbench.editorAssociations` preference


msujew:
### Feature Description:

The `workbench.editorAssociations` preference in VS Code allows to override the default editor for a specific file type. It can be triggered from the UI by using the `Open With...` command:

![image](https://github.com/user-attachments/assets/d2b7cd0b-c896-4090-8f5e-1ed9641cfd35)

PR #14139, opened 2024-09-05
Support `workbench.editorAssociations` preference

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/14138

Extends the `OpenWithService` to allow to select the default editor for a specific file type. This decision is then stored in the `workbench.editorAssociations` preference.

Later on, in the respective open handlers, we use this preference to figure out whether the current handler is the "default handler". If it is, we return an absurdly high number for `canHandle` to ensure that the handler is the one with the highest priority.

#### How to test

For the following tests, I would recommend testing the [hex editor extension](https://open-vsx.org/extension/ms-vscode/hexeditor), notebooks and normal text files.

1. Run the `Open With...` command on a file.
2. Select the `Configure default editor` option in the quick pick
3. Select a new default.
4. Clicking on the file should open it every time with the newly selected default.
5. Using `Open With...` should also show that it is the default now. The ordering within the quick pick should not have changed.
6. The open handler is correctly registered in the `workbench.editorAssociations` preference in the user settings.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/core-preferences.ts
+++ b/packages/core/src/browser/core-preferences.ts
@@ -281,6 +281,15 @@ export const corePreferenceSchema: PreferenceSchema = {
             default: 200,
             minimum: 10,
             description: nls.localize('theia/core/tabDefaultSize', 'Specifies the default size for tabs.')
+        },
+        'workbench.editorAssociations': {
+            type: 'object',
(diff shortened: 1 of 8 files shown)


jonah-iden (approved):
Works great and LGTM 👍

Issue #13582, opened 2024-04-09
New File dialog doesn't accept Enter for default


profbbrown:
### Bug Description:
This is in reference to new feature #13303 that was added in 1.48, in which the New File menu item now allows a filename to be entered.

After entering the filename, a dialog box is shown to prompt the user to select a folder in which to create the file. The dialog doesn't allow Enter (or Return) to be pressed to accept the default.

### Steps to Reproduce:

1. Select File > New File.
2. Enter a filename, press Enter or Return.
3. On the next dialog box, the user's home directory should be highlighted and the Create File button should be also be highlighted indicating that pressing Enter or Return would select it. But nothing happens when that key is pressed.

### Additional Information

- Operating System: MacOS Sonoma
- Theia Version: 1.48

JonasHelming:
@dhuebner FYI

dhuebner:
@profbbrown 
Thanks for reporting! I will check it.

dhuebner:
@profbbrown 
Are you using the official Theia IDE Desktop app? If yes, which folder is shown in the system file dialog? Is it the same folder that is shown when clicking `File -> Open...` ,  or `CMD + O`? 

If you are using a browser app, then this is the standard behavior to preselect/focus the file tree and not the "accept" button in the file dialog.
I'm not sure what is the best default here. On the one hand it is unlikely that the user want to create a new file in the user folder, hence will probably first use the navigation tree to navigate to the project (or what ever) folder. On the other hand for the fast file creation in the user folder it would probably be faster to just hit ENTER  instead of using TAB key three times and then ENTER.

profbbrown:
Thanks for the response. I appreciate the thought that went into the choices that were made.

I expected Enter to work as the default because the Create File button was highlighted. In most other apps, a highlighted button will be the one that is "clicked" when Enter is pressed. Perhaps the answer is to not highlight the button so there's no confusion from a UX perspective.

tsmaeder:
I think it's important here to distinguish between keyboard focus an the default button of a dialog. The default button is the one that is "highlighted", but the keyboard focus (at least on windows) is indicated by a dotted line. 
T.b.h, I think using a "save as" dialog does not make sense with the UI we have: the user has already selected a file name, the only thing left to chose is the folder it's going to be in. So a "select folder" dialog would be the appropriate choice here.
Alternatively, add an option "Save as" to the initial drop-down menu instead of allowing the user to enter a file name, then the dialog would make sense.
[...]

PR #14146, opened 2024-09-06
New File dialog doesn't accept Enter for default

dhuebner:
#### What it does

Allow to accept the File Dialog by pressing `Enter` and canceling by pressing `Escape`. 

See #13582 

#### How to test

In `theia-main-menu.test.ts` the "Create file via New File menu and accept" test uses `Enter` to accept the File Dialog an save a file. 

Or follow the description in  #13582 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/playwright/src/tests/theia-main-menu.test.ts
+++ b/examples/playwright/src/tests/theia-main-menu.test.ts
@@ -110,18 +110,17 @@ test.describe('Theia Main Menu', () => {
         expect(await fileDialog.isVisible()).toBe(false);
     });
 
-    test('Create file via New File menu and cancel', async () => {
-        const openFileEntry = 'New File...';
-        await (await menuBar.openMenu('File')).clickMenuItem(openFileEntry);
+    test('Create file via New File menu and accept', async () => {
+        await (await menuBar.openMenu('File')).clickMenuItem('New File...');
         const quickPick = app.page.getByPlaceholder('Select File Type or Enter');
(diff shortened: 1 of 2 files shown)


tsmaeder:
@dhuebner is this a problem on Mac only? I can accept the file dialog with enter on Windows.

dhuebner:
@tsmaeder 
No, I don't think it is Mac specific. The problem was, that the tree widget eats the Enter and ESC key stroke events so that the dialog can not be closed using keyboard. One need to click on the corresponding buttons or use TAB to jump to them and then hit Enter.

dhuebner:
@tsmaeder 
Now, I see. You need to use the menu "File" -> "New File...". Then save the untitled file and you will get this dialog.

tsmaeder:
That's what I'm doing: in the video, I just press "Enter" when the "Save as"-dialog opens. What am I missing?


https://github.com/user-attachments/assets/a82dd52a-ba23-42a4-b56b-f0b35b6733fe

dhuebner:
@tsmaeder 
The problem only occurs when running in the browser. In the electron version the system dialog is used and works fine.
Sorry for the confusion, I need to better explain how to test.

tsmaeder (approved):
Seems to do the trick. I'm still not convinced we should not handle the keyboard focus differently, but that's a larger discussion.

dhuebner:
@tsmaeder 
Thank you for the review!

PR #14147, opened 2024-09-06
Handle only the user workspace security settings

dhuebner:
#### What it does

Prevents setting workspace security settings in the workspace scope.

#### How to test

Open settings, switch to Workspace scope tab. `Security` should not be visible.
 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/workspace/src/browser/workspace-trust-service.ts
+++ b/packages/workspace/src/browser/workspace-trust-service.ts
@@ -15,7 +15,7 @@
 // *****************************************************************************
 
 import { ConfirmDialog, Dialog, PreferenceChange, StorageService } from '@theia/core/lib/browser';
-import { PreferenceService } from '@theia/core/lib/browser/preferences/preference-service';
+import { PreferenceScope, PreferenceService } from '@theia/core/lib/browser/preferences/preference-service';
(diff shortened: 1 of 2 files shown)


packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -115,17 +115,19 @@ export class WorkspaceTrustService {
     }
 
     protected async handlePreferenceChange(change: PreferenceChange): Promise<void> {
-        if (change.preferenceName === WORKSPACE_TRUST_STARTUP_PROMPT && change.newValue !== WorkspaceTrustPrompt.ONCE) {
-            this.storage.setData(STORAGE_TRUSTED, undefined);
-        }
+        if (change.scope === PreferenceScope.User) {
```
msujew, line 118: Suggestion: This change isn't necessary. The change event will not trigger if a `User` scoped preference is changed in the `Workspace` scope.

dhuebner: It's the second line of defense :)

msujew: Ok, fine with me.

msujew (approved):
Alright, works well for me 👍

PR #14149, opened 2024-09-09
Add handling of multiple accounts to authentication api

### LINKED ISSUE #14110: [vscode] Additions to authentication namespace and session management with several accounts
The authentication APIs now have more control when handling multiple accounts: addition of AuthenticationSessionAccountInformation, used in several places, getSession update, get Accounts in authentification namespace, etc.
 
See https://github.com/microsoft/vscode-docs/blob/vnext/release-notes/v1_93.md#authentication-account-api

tsmaeder:
#### What it does
Adds "account" parameter to `get/createSession()` calls. This allows to handle multiple accounts with the same scopes for the same authentication provider. This may be a breaking change for users and implementers of `AuthenticationService`. Also removed two deprecated methods. 

Fixes #14110

#### How to test
The attached extensions contributes four commands:

1. "Log into AuthTest". 
Allows to log in with the Auth Test provider. Either into a preexisting session or with a new user name. The "default session" setting gets cleared.
2. "Use Any Auth Test Login"
Asks the system for a auth session while clearing the preferred session preference
3. "Use Preferred Auth Test Login"
Same as above, but don't clear the preferred session preference
4. "Use Named Auth Test Login"
Same as above, but you can select a specific user name the system will ask for

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 <!-- ## Unreleased
 
 <a name="breaking_changes_1.54.0">[Breaking Changes:](#breaking_changes_1.54.0)</a> -->
+- [core] Updated AuthenticationService to handle multiple accounts per provider [#14149](https://github.com/eclipse-theia/theia/pull/14149) - Contributed on behalf of STMicroelectronics
 
 ## 1.53.0 - 08/29/2024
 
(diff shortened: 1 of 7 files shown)


tsmaeder:
Here's the extensions
[auth-test-0.0.1.zip](https://github.com/user-attachments/files/16927384/auth-test-0.0.1.zip)
And here's the source:
[auth-test-src.zip](https://github.com/user-attachments/files/16927394/auth-test-src.zip)

PR #14181, opened 2024-09-16
feat(ai): Add toolbar actions for chat nodes

planger:
#### What it does

Introduces a `ChatNodeToolbarActionContribution` which allows to add icons to the chat nodes in the Chat UI on hover. Clicking these icons will trigger a command that gets the respective chat node as argument.

Possible use cases are actions for reporting the content, giving feedback, or perform other actions within the tool based on the response.

#### How to test

Add a `ChatNodeToolbarActionContribution` to a frontend module:
```ts
bind(ChatNodeToolbarActionContribution).toDynamicValue(context => ({
 getToolbarActions: (args: RequestNode | ResponseNode) => {
     if (isResponseNode(args)) {
         return [{
             commandId: 'core.about',
             icon: 'codicon codicon-feedback',
             tooltip: 'Show about dialog on response nodes'
         }];
     } else {
         return [];
     }
 }
```

Then trigger a chat and check whether the action appears on hover.

#### Follow-ups

N/A

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/api-samples/src/browser/chat/chat-node-toolbar-action-contribution.ts
+++ b/examples/api-samples/src/browser/chat/chat-node-toolbar-action-contribution.ts
@@ -1,0 +1,41 @@
+// *****************************************************************************
+// Copyright (C) 2024 STMicroelectronics and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import {
+    ChatNodeToolbarActionContribution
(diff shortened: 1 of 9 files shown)


sdirix (approved):
Works for me!

I tested with multiple contributions including different priorities and the toolbar was rendered as expected.

I wonder whether we should add an example to the official Theia examples. Maybe with the hello world command instead of about.

planger:
@sdirix Thanks for the review!

Good idea regarding the example. I added a chat node action contribution. I used the about command, however, as I didn't find a better suited command defined in the examples. Please let me know what you think.

examples/api-samples/src/browser/chat/chat-node-toolbar-action-contribution.ts
```diff
@@ -0,0 +1,41 @@
...
+import { interfaces } from '@theia/core/shared/inversify';
+
+export function bindChatNodeToolbarActionContribution(bind: interfaces.Bind): void {
+    bind(ChatNodeToolbarActionContribution).toDynamicValue(context => ({
+        getToolbarActions: (args: RequestNode | ResponseNode) => {
+            if (isResponseNode(args)) {
+                return [{
+                    commandId: 'core.about',
+                    icon: 'codicon codicon-feedback',
+                    tooltip: 'Show about dialog'
```
sdirix, line 34: I would like to suggest the following, this makes the example intent much more clear
```suggestion
                    commandId: 'sample-command',
                    icon: 'codicon codicon-feedback',
                    tooltip: 'Example command'
```

JonasHelming (approved):

PR #14182, opened 2024-09-16
Consitently name agents and add tags

### LINKED ISSUE #14180: [Theia AI] Consitently name agents
Use short names (comment in code that it must be unique)
Ids are hidden from the user for now (comment in code that it might be removed and should be the same as ID for now)
In the AI Configuration View, we suffix Chat Agents with ""Chat"
Introduce "Tags" for additional properties (such as "Chat" for now)

JonasHelming:
fixed #14180

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
+++ b/packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
@@ -73,7 +73,9 @@ export class AIAgentConfigurationWidget extends ReactWidget {
             <div className='configuration-agents-list preferences-tree-widget theia-TreeContainer' style={{ width: '25%' }}>
                 <ul>
                     {this.agentService.getAllAgents().map(agent =>
-                        <li key={agent.id} className='theia-TreeNode theia-CompositeTreeNode theia-ExpandableTreeNode' onClick={() => this.setActiveAgent(agent)}>{agent.name}</li>
+                        <li key={agent.id} className='theia-TreeNode theia-CompositeTreeNode theia-ExpandableTreeNode' onClick={() => this.setActiveAgent(agent)}>
+                            {this.renderAgentName(agent)}
+                        </li>
                     )}
                 </ul>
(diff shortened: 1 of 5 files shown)


packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
```diff
@@ -83,6 +85,12 @@ export class AIAgentConfigurationWidget extends ReactWidget {
         </div>;
     }
 
+    private renderAgentName(agent: Agent): string {
+        // Check if the agent has tags, and join them with a comma separator if they exist
+        const tagsSuffix = agent.tags?.length ? ` (${agent.tags.join(', ')})` : '';
+        return `${agent.name}${tagsSuffix}`;
+    }
```
planger, line 92: We may want to render those tags more nicely, but this is optional.

```suggestion
    private renderAgentName(agent: Agent): React.ReactNode {
        const tagsSuffix = agent.tags?.length ? <span>{agent.tags.map(tag => <span className='agent-tag'>{tag}</span>)}</span> : '';
        return <span>{agent.name} {tagsSuffix}</span>;
    }
```

image

To make them look as above, we need the following CSS in `packages/ai-core/src/browser/style/index.css`:
```css
  padding: calc(var(--theia-ui-padding) * 2 / 3);
  padding-top: 0px;
  padding-bottom: 0px;
  border-radius: calc(var(--theia-ui-padding) * 2 / 3);
  background: hsla(0, 0%, 68%, 0.31);
```

planger (approved):

I have just a suggestion regarding the styling of the tags below. Imho this would look a bit nicer.

packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
JonasHelming: looks very nice, thank you! applied in new commit

PR #14186, opened 2024-09-17
Leak fix in tabs-main.ts

dfriederich:
The disconnect from the signal was not passing the this reference as it was used in the connect, hence the disconnect failed to find the handler.

When developing a theia based app we run into leaks due to TabsMainImpl keeping our custom control
in memory even when it was closed. The leak was through the controls title, through TabsMainImpl.
This PR fixes the main issue, which was leaking every control for us.

However the TabsMainImpl still is not optimal as it only releases references when it rebuilds its model,
and that does not happen when closing a tab.

### DIFF
--- a/packages/plugin-ext/src/main/browser/tabs/tabs-main.ts
+++ b/packages/plugin-ext/src/main/browser/tabs/tabs-main.ts
@@ -231,9 +231,13 @@ export class TabsMainImpl implements TabsMain, Disposable {
         return { kind: TabInputKind.UnknownInput };
     }
 
-    protected connectToSignal<T>(disposableList: DisposableCollection, signal: { connect(listener: T, context: unknown): void, disconnect(listener: T): void }, listener: T): void {
+    protected connectToSignal<T>(disposableList: DisposableCollection,
+        signal: {
+            connect(listener: T, context: unknown): void,
+            disconnect(listener: T, context: unknown): void
+        }, listener: T): void {
         signal.connect(listener, this);
-        disposableList.push(Disposable.create(() => signal.disconnect(listener)));
+        disposableList.push(Disposable.create(() => signal.disconnect(listener, this)));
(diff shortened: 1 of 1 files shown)


msujew (approved):
Thanks looks good to me 👍

> However the TabsMainImpl still is not optimal as it only releases references when it rebuilds its model, and that does not happen when closing a tab.

I plan to refactor the code that handles the tab API and the whole editor manager eventually anyway. This will likely be fixed by that change.

dfriederich:
>I plan to refactor the code that handles the tab API and the whole editor manager eventually anyway. 
> This will likely be fixed by that change.

The leak is quite unfortunate IMHO, wonder how bad it would be to recompute the model for any tab removal?
would at least fix the leak for me. I will create an issue to track that it currently leaks.

martin-fleck-at:
I'm also interested in this, so thank you @dfriederich for creating https://github.com/eclipse-theia/theia/issues/14203.
@msujew Do you have any approximate timeline on this?

martin-fleck-at:
> I plan to refactor the code that handles the tab API and the whole editor manager eventually anyway. This will likely be fixed by that change.

Is there any issue to track this or do you have any timeline in mind? I'm just wondering because I've seen some other issues on the tabs where the sometimes we end up with a '-1' index in the tabs (because it can't be found in the model) and other issues that results in an 'invalid tab' error message and I am curious what the refactoring of that might look like.

msujew:
@martin-fleck-at Thanks for asking! There is currently no issue for this but I can create one next week. I currently have very little time to work on this, so any help is very much appreciated. My current idea is to use something very similar to VS Code's `EditorGroupService` which can then be used as a source for the VS Code tab API. Currently, everything is based on the application shell, which isn't a good source of data for the tab API, as it doesn't keep track of columns.

PR #14190, opened 2024-09-18
Aligned active text and notebook editor more towards vscode

jonah-iden:
#### What it does

this aligns the vscode API's `activeTextEditor` and `activeNotebookEditor` more to how it is in Vscode
- `activeNotebookEditor` should always be undefined when a text editor is selcted.
- `activeNotebookEditor` should be set when a notebook editor is and `activeTextEditor` should be undefined as long as no cell is selected
-  `activeTextEditor` should be referencing the cell when a notebook is open and a cell is in edit mode

#### How to test

you can find and download a test extension [here](https://github.com/jonah-iden/theia-notebook-selection-test)

Open the outputs tab and select the `Notebook Selection Test` channel. Here you should see all editor change events.

#### Follow-ups

The endstate should be correct but the amount of events are a few more than in vscode. This requires a larger refactoring of the `editor-manager.ts` to fully fix. 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/editors-and-documents-main.ts
+++ b/packages/plugin-ext/src/main/browser/editors-and-documents-main.ts
@@ -34,6 +34,8 @@ import { DisposableCollection, Emitter, URI } from '@theia/core';
 import { EditorManager, EditorWidget } from '@theia/editor/lib/browser';
 import { SaveableService } from '@theia/core/lib/browser/saveable-service';
 import { TabsMainImpl } from './tabs/tabs-main';
+import { NotebookCellEditorService } from '@theia/notebook/lib/browser';
+import { SimpleMonacoEditor } from '@theia/monaco/lib/browser/simple-monaco-editor';
 
 export class EditorsAndDocumentsMain implements Disposable {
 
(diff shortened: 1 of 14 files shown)


packages/plugin-ext/src/main/browser/editors-and-documents-main.ts
```diff
@@ -384,7 +394,7 @@ class EditorAndDocumentState {
 
 class EditorSnapshot {
     readonly id: string;
-    constructor(readonly editor: MonacoEditor) {
+    constructor(readonly editor: MonacoEditor | SimpleMonacoEditor) {
```
msujew, line 397: Suggestion: We could simply use the underlying type of `getControl()` (which is `ICodeEditor`). Using the union everywhere seems to lead to a lot of code duplication.

msujew (commented):
From a functionality standpoint, this works pretty well. However, I think we can cleanup the code a lot, see comment below.

packages/plugin-ext/src/main/browser/editors-and-documents-main.ts
jonah-iden: that does not work because both of them are inheriting from different ICodeEditor interfaces which both use an enum with the same name but different implementation somewhere deep inside. So the getControl of both of them is not compatible

jonah-iden: or maybe not. Maybe my IDE is just confused out of some weird reason

jonah-iden: SimpleMonacoEditor uses the ICodeEditor from editorBrowser.ts and the MonacoEditor uses it from the monac editor api directly

msujew: Oh, ok I see. Makes sense.

msujew (approved):
It's not perfect, but good enough and way better than previously. I think we can continue refactoring this later 👍

Issue #14015, opened 2024-08-06
Drop nsfw support 


rschnekenbu:
NSFW (node-sentinel-file-watcher) is one of our dependencies and was not updated for the last 2 years.

In the context of #13957, there may be some evolution to do on the FileSystemWatcher mechanism, it could be a good point to evaluate changing this library to a newer one. In short about these evolutions:
- [file-system-event-service-ext-impl.ts ](https://github.com/eclipse-theia/theia/blob/267ed206e43d090a6392b158519a2e9f0cb9dccc/packages/plugin-ext/src/plugin/file-system-event-service-ext-impl.ts#L21C6-L21C7) may follow evolutions as indicated in its comment
- the proposed API used by typescript require API and behavior evolution on our current framework. 

As a note: VS Code is now using parcel (https://github.com/parcel-bundler/watcher).

msujew:
For the necessary changes, see https://github.com/eclipse-theia/theia/pull/12784 :)

PR #14194, opened 2024-09-18
Replace `nsfw` with `@parcel/watcher`

msujew:
#### What it does

Continuation of https://github.com/eclipse-theia/theia/pull/12784

Closes https://github.com/eclipse-theia/theia/issues/14015

Replaces our `nsfw` dependency with `@parcel/watcher` to align with vscode.

#### How to test

Test file watching related capabilities and confirm that there are no regressions:
* Editing files on the file system should be reflected in the editor
* Editing `settings.json` should be reflected in changes in the preferences
* Editing `launch.json` should be reflected in the debug view

#### Review checklist

#### Reminder for reviewers


msujew:
@tsmaeder I'll be on vacation for a week starting Friday. In case you want to change anything in the PR, please feel free :)

tsmaeder:
There is a basic scenario that seems to be broken with the new watcher: 

1. Open a workspace with the theia source
2. Do `yarn` to populate the node_modules folder
3. Expand some the folders inside the node_modules folder
4. In the Windows explorer, select all folders inside the node_modules folder
5. Press delete
6. Observe: Windows moves the folders to the trash
7. Observe: the changes inside the node_modules folder are not picked up inside Theia file explorer

In master, the deletion of the modules is reflected in the Theia file navigator

msujew:
@safisa It might fix the issue, yes.

msujew:
@planger Could you have a look at the failing playwright tests? The errors are not reproducible on Windows and the tests don't run for me on Ubuntu (Gitpod) for some reason.

tsmaeder:
I've looked into the failures of the Playwright tests. They only fail on Linux. IMO, this is a bug in the "compressed tree" and is probably unrelated to the change at hand. So, with the resources folder from the playwright test, you can do the following scenario:

1. delete "nestedFolder2"
2. Observe: there is no change visible and a `console.log` in the rendering of the compressed rows is not hit.
3. Select the row in question
4. Observe: `nestedFolder2` disappears from the row.

The same events seem to be sent on Linux and Windows, so my suspicion is that this is a reordering of async  blocks du to the different timings on the respective platforms.

planger:
Thanks for looking into this @tsmaeder! This makes sense, looking at the logs briefly.

We can take a look at the failing test in more detail, but I also suspect it might be an actual UI issue on Linux. To avoid having failing tests, I would suggest skipping this failing test for now and open a new issue.

tsmaeder (approved):

PR #14195, opened 2024-09-18
Makes the cell editor border grey when not focused

dhuebner:
#### What it does

Currently if a code cell is selected  (Blue line on the lefthand side)  but the cell editor itself has no focus - the editor has a focused (blue) border. This change makes the editor outline/border grey is the cell editor doesn't have focus. 
 

#### How to test

Open a notebook and select a cell editor. The editor should have a focus (blue) border. Press `ESC` - it turns off the edit mode and editor lose focus. The editor border should become grey.
 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/style/index.css
+++ b/packages/notebook/src/browser/style/index.css
@@ -114,7 +114,8 @@
   margin: 0px 16px 0px 10px;
 }
 
-.theia-notebook-cell.focused .theia-notebook-cell-editor-container {
+/* Only mark an editor cell focused if the editor has focus */
+.theia-notebook-cell-editor-container:has(.monaco-editor.focused) {
   outline-color: var(--theia-notebook-focusedEditorBorder);
 }
 


msujew (commented):
While this fixes the issue with the editor, the cell sidebar still has the blue highlight color:

image

In VS Code, the cell side bar marker becomes "inactive":

image

dhuebner:
> While this fixes the issue with the editor, the cell sidebar still has the blue highlight color:

It is the intended behavior. Wenn pressing `ESC` only the editor should be turned into non-edit state. The cell itself should remain selected.
 
> In VS Code, the cell side bar marker becomes "inactive":

Not in my vs-code. Pressing `ESC` turns off the editor focus, pressing `ENTER` activates the editor again. The side bar shows selected state. 

What you maybe see, is when mouse-click somewhere outside a cell in the editor, the last selected **cell** is still active (marked focused). This was before this change and is not a part of the fix.

msujew (approved):
Ah, I see. Thanks for the explanation. Then everything works as expected 👍

PR #14196, opened 2024-09-18
feat(ai): Make response parsing extensible

planger:
#### What it does

Turns the response parsing method of the chat agents into a more flexible algorithm that can work with multiple response content matchers. Each response content matcher has a start and an end regexp to define a match, as well as a `contentFactory` function that is invoked to turn the matched content into a specific `ChatResponseContent` object.

This way, clients can add and combine multiple response content matchers to process LLM responses and turn them into custom response parts, while still keeping the generically useful ones, like the matcher for code blocks, etc.

Additionally, the parsing method has a fallback content factory that will be applied to all unmatched parts. By default, this is the existing markdown content part, but this can be customized if needed.

Both, the response content matchers and the fallback content factory are configurable via DI.

#### How to test

Ensure normal behavior is still working as expected. That is, we should test the processing and rendering of plain chat responses that contain an arbitrary mixture of markdown parts, code parts, and tool call parts.

In addition, one could test to customize the `DefaultResponseContentFactory` or register additional `ResponseContentMatcherProvider`.

```ts
    // Change the default response part
    rebind(DefaultResponseContentMatcherProvider).to(YourDefaultPartCustomization);
    // Add additional matchers
    bind(ResponseContentMatcherProvider).to(YourAdditionalResponseContentMatcherProvider);
    // Change default matchers (code)
    rebind(DefaultResponseContentFactory).to(YourCustomization);
```

Since the parsing is a bit of a dense code, I added unit tests for it.

#### Follow-ups

I tried to simplify the parsing. However, there is one aspect, which I found a bit confusing but is not changed by this PR:
The fall back to markdown, handling of tool calls, or parsing into specific parts is a bit spread across the code
* 2x in [`parse()`](https://github.com/eclipse-theia/theia/blob/4a0228b362801af796c8596317b735860fc9e427/packages/ai-chat/src/common/chat-agents.ts#L369)
* 3x in [`addContentsToResponse`](https://github.com/eclipse-theia/theia/blob/4a0228b362801af796c8596317b735860fc9e427/packages/ai-chat/src/common/chat-agents.ts#L282C30-L282C51), where we currently need to ensure, we prevent parsing, if the previous content was a tool call, which is also somewhat hidden.

Maybe there is an opportunity to further simplify this and parse this all in one place and one go.

Also it was unclear to me why we don't parse for specific content parts in case we have a [`isLanguageModelTextResponse` === true](https://github.com/eclipse-theia/theia/blob/4a0228b362801af796c8596317b735860fc9e427/packages/ai-chat/src/common/chat-agents.ts#L282C30-L282C51) case. This is actually changed so that it now uses the `parseContents` function to obtain the content parts instead of creating a plain markdown part either way, which feels more consistent.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-agents.ts
+++ b/packages/ai-chat/src/common/chat-agents.ts
@@ -371,7 +371,7 @@ export abstract class AbstractStreamParsingChatAgent extends AbstractChatAgent {
         const content = token.content;
         // eslint-disable-next-line no-null/no-null
         if (content !== undefined && content !== null) {
-            return this.defaultContentFactory?.create(content);
+            return this.defaultContentFactory.create(content);
         }
         const toolCalls = token.tool_calls;
         if (toolCalls !== undefined) {
--- a/packages/ai-chat/src/common/chat-agents.ts
+++ b/packages/ai-chat/src/common/chat-agents.ts
@@ -25,6 +25,7 @@ import {
     LanguageModel,
     LanguageModelRequirement,
     LanguageModelResponse,
+    LanguageModelStreamResponse,
     PromptService,
     ResolvedPromptTemplate,
     ToolRequest,
@@ -37,19 +38,20 @@ import {
     LanguageModelStreamResponsePart,
     MessageActor,
 } from '@theia/ai-core/lib/common';
-import { CancellationToken, CancellationTokenSource, ILogger, isArray } from '@theia/core';
-import { inject, injectable } from '@theia/core/shared/inversify';
+import { CancellationToken, CancellationTokenSource, ContributionProvider, ILogger, isArray } from '@theia/core';
+import { inject, injectable, named, postConstruct } from '@theia/core/shared/inversify';
 import { ChatAgentService } from './chat-agent-service';
 import {
     ChatModel,
     ChatRequestModel,
     ChatRequestModelImpl,
     ChatResponseContent,
-    CodeChatResponseContentImpl,
     ErrorChatResponseContentImpl,
     MarkdownChatResponseContentImpl,
     ToolCallChatResponseContentImpl
 } from './chat-model';
+import { findEarliestMatch, parseContents } from './parse-contents';
+import { DefaultResponseContentFactory, ResponseContentMatcher, ResponseContentMatcherProvider } from './response-content-matcher';
 
 /**
  * A conversation consists of a sequence of ChatMessages.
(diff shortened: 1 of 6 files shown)


packages/ai-chat/src/common/chat-model.ts
```diff
@@ -601,6 +601,10 @@ class ChatResponseImpl implements ChatResponse {
         return this._content;
     }
 
+    addContents(contents: ChatResponseContent[]): void {
+        contents.forEach(c => this.addContent(c));
```
eneufeld, line 605: does it make sense to take this shortcut?
each add call will fire an change event which will lead to many change events which we maybe don't want?

packages/ai-chat/src/common/chat-agents.ts
```diff
@@ -366,19 +338,48 @@ export abstract class AbstractStreamParsingChatAgent extends AbstractChatAgent {
...
+            }
+        }
+    }
+
+    protected parse(token: LanguageModelStreamResponsePart, previousContent: ChatResponseContent[]): ChatResponseContent | ChatResponseContent[] {
         const content = token.content;
         // eslint-disable-next-line no-null/no-null
         if (content !== undefined && content !== null) {
-            return new MarkdownChatResponseContentImpl(content);
+            return this.defaultContentFactory?.create(content);
```
eneufeld, line 374: why is the factory optional here and non optional in line 382?

eneufeld (commented):
I like the tests.
Maybe we could improve the readability a bit as the findEarliestMatch and parseContents methods are hard to read.
I will take a second look maybe I find something to suggest

packages/ai-chat/src/common/chat-model.ts
planger: Good question, I was mainly trying to avoid the use of this when I had to add it a second time. The first time was already here:
https://github.com/eclipse-theia/theia/pull/14196/files/9a7013848910ae57dcb77e30cde9a3410bc95d0a#diff-b2169b1bd3b36aa51d71862eb11d8ca6ea57f214898af862f0cf6777656778bdL301

I assume, sending just one event would be fine. WDYT?

packages/ai-chat/src/common/chat-agents.ts
planger: Oh, this is actually an oversight. It was optional first, but I made it mandatory with a default along the way. So we can omit this `?`.

packages/ai-chat/src/common/parse-contents.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    return result;
+}
+
+export function findEarliestMatch(contentMatchers: ResponseContentMatcher[], text: string): Match | undefined {
+    let earliestMatch: { matcher: ResponseContentMatcher, index: number, content: string } | undefined;
+    for (const matcher of contentMatchers) {
+        const startMatch = matcher.start.exec(text);
+        if (startMatch) {
```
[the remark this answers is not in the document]
planger: Sure, if you prefer, we can invert the condition and use a `continue`. Not sure what's better though. I'm not a big fan of using `continue`, but I'm ok using it here, especially in the beginning as a guard of the iteration, if you find it easier to read.

planger:
@eneufeld Thank you for the review. I've addressed your comments. Can you please re-check and re-test? Thank you!

eneufeld (approved):
Works for me, thank you

JonasHelming (approved):
LGTM!

PR #12277, opened 2023-03-09
[plugin system] Add stubbed support for proposed extensions.allAcross ExtensionHosts and related APIs

marcdumais-work:
#### What it does

I believe that we do not support multiple extension hosts yet, so this is effectively adding a stubbed version of these proposed vscode APIs, for the benefit of extensions that use them.

At the moment, built-in "vscode.html-language-features" (v1.70.2) does, and will not activate correctly without this.

Fixes #12276
#### How to test

Here's the extension that I built locally (it's not published yet):
[html-language-features-1.70.2.zip](https://github.com/eclipse-theia/theia/files/10943856/html-language-features-1.70.2.zip)
- unzip the extension and copy it to folder `plugins`. Remove folder `vscode.html-language-features`  if it exists.
- start the example application and open a `.html` file
- [ ] confirm that the extension activate correctly and seems to work
- [ ] ~~confirm that the [CQ](https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/7350) for the copied content is approved~~ [1]

[1 ] : no IP check necessary in this case. See: https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/7350#note_1090755

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #14199, opened 2024-09-19
[vscode] Support evolution on proposed API extensionAny

### LINKED ISSUE #14115: [vscode] Evolution of the extensionsAny.d.ts interface
This interface is implemented here: https://github.com/eclipse-theia/theia/blob/754eb84caecc8e318def10c6b96e4d11ba7081b9/packages/plugin-ext/src/plugin/plugin-context.ts#L884
Update to the proposed interface file and implementation code may necessary:

Change: Added a new overload for the getExtension function with the signature `getExtension<T = any>(extensionId: string, includeDifferentExtensionHosts: true): Extension<T | undefined> | undefined. `

This proposed extension is used by the html language features extension.

rschnekenbu:
#### What it does

Proposed API extensionsAny will now have a new signature for getExtension:
https://github.com/microsoft/vscode/blob/38c31bc77e0dd6ae88a4e9cc93428cc27a56ba40/src/vscode-dts/vscode.proposed.extensionsAny.d.ts#L31

This Pull Request adds this to the similar interface in Theia file, and also update getExtension from plugin-context to align with this new signature. 
Since https://github.com/eclipse-theia/theia/pull/12277, the `includeDifferentExtensionHosts` parameter is ignored, as there is only one host in Theia. So there should be no other change to support this evolution.

fixes #14115

#### How to test

Use the builtins and assume they still work with this evolution. The html-language-feature is the only builtin relying on the extensionsAny proposed API.

#### Follow-ups

None

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin/src/theia.proposed.extensionsAny.d.ts
+++ b/packages/plugin/src/theia.proposed.extensionsAny.d.ts
@@ -41,9 +41,13 @@ export module '@theia/plugin' {
          * @param extensionId An extension identifier.
          * @param includeDifferentExtensionHosts Include extensions from different extension host
          * @return An extension or `undefined`.
+         *
+         * *Note* In Theia, includeDifferentExtensionHosts will always be set to false, as we only support one host currently.
          */
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         export function getExtension<T = any>(extensionId: string, includeDifferentExtensionHosts: boolean): Extension<T> | undefined;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        export function getExtension<T = any>(extensionId: string, includeDifferentExtensionHosts: true): Extension<T | undefined> | undefined;
 
         /**
          * All extensions across all extension hosts.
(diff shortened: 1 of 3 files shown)


packages/plugin/src/theia.proposed.extensionsAny.d.ts
```diff
@@ -41,9 +41,13 @@ export module '@theia/plugin' {
          * @param extensionId An extension identifier.
          * @param includeDifferentExtensionHosts Include extensions from different extension host
          * @return An extension or `undefined`.
+         *
+         * *Note* In Theia, includeDifferentExtensionHosts will always be set to false, as we only support one host currently.
```
tsmaeder, line 45: That is actually not true: we support at least two different extension hosts: front-end and back-end. Back in the Che days, there was even support for remote-calling the extension interface via RPC.

tsmaeder (approved):

tsmaeder:
@sdirix since you're Mr. "browser-only-Theia", would it make sense to do a follow-up to this PR that actually implements the functionality for multiple extensions hosts?

PR #14211, opened 2024-09-23
fix markdown request renderer

### LINKED ISSUE #14208: [Theia AI] HTML is not escaped in the Chat View
If you use HTML in the users message, it is not escaped, e.g. use

`Theia IDE Screenshot`

in the message. This will send "Theia IDE Screenshot" only to the underlying LLM.

eneufeld:
#### What it does

Currently requests in the ai chat view use a markdown renderer which parses html input and tries to render this. This leads to unexpected behavior when using html in the query.
By deactivating html support there, an alternative is needed as the markdown renderer just renders an empty paragraph if html is now passed to it. So to render the actual text, a span is created and filled with the text content.

fixes #14208
#### How to test

Use a string like `Theia IDE Screenshot` in you query.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-response-renderer/markdown-part-renderer.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-response-renderer/markdown-part-renderer.tsx
@@ -25,6 +25,7 @@ import { ReactNode, useEffect, useRef } from '@theia/core/shared/react';
 import * as React from '@theia/core/shared/react';
 import * as markdownit from '@theia/core/shared/markdown-it';
 import * as DOMPurify from '@theia/core/shared/dompurify';
+import { MarkdownString } from '@theia/core/lib/common/markdown-rendering';
 
 @injectable()
 export class MarkdownPartRenderer implements ChatResponsePartRenderer<MarkdownChatResponseContent | InformationalChatResponseContent> {
@@ -52,7 +53,7 @@ export class MarkdownPartRenderer implements ChatResponsePartRenderer<MarkdownCh
 }
 
 const MarkdownRender = ({ response }: { response: MarkdownChatResponseContent | InformationalChatResponseContent }) => {
-    const ref = useMarkdownRendering(response.content.value);
+    const ref = useMarkdownRendering(response.content);
 
     return <div ref={ref}></div>;
 };
@@ -67,13 +68,14 @@ const MarkdownRender = ({ response }: { response: MarkdownChatResponseContent | 
  * @param markdown the string to render as markdown
  * @returns the ref to use in an element to render the markdown
  */
-export const useMarkdownRendering = (markdown: string) => {
+export const useMarkdownRendering = (markdown: string | MarkdownString) => {
     // eslint-disable-next-line no-null/no-null
     const ref = useRef<HTMLDivElement | null>(null);
     useEffect(() => {
         const markdownIt = markdownit();
         const host = document.createElement('div');
-        const html = markdownIt.render(markdown);
+        const markdownString = typeof markdown === 'string' ? markdown : markdown.value;
+        const html = markdownIt.render(markdownString);
         host.innerHTML = DOMPurify.sanitize(html, {
             ALLOW_UNKNOWN_PROTOCOLS: true // DOMPurify usually strips non http(s) links from hrefs
         });
--- a/packages/ai-chat-ui/src/browser/chat-response-renderer/markdown-part-renderer.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-response-renderer/markdown-part-renderer.tsx
@@ -71,10 +71,10 @@ const MarkdownRender = ({ response }: { response: MarkdownChatResponseContent | 
 export const useMarkdownRendering = (markdown: string | MarkdownString) => {
     // eslint-disable-next-line no-null/no-null
     const ref = useRef<HTMLDivElement | null>(null);
+    const markdownString = typeof markdown === 'string' ? markdown : markdown.value;
     useEffect(() => {
         const markdownIt = markdownit();
         const host = document.createElement('div');
-        const markdownString = typeof markdown === 'string' ? markdown : markdown.value;
         const html = markdownIt.render(markdownString);
         host.innerHTML = DOMPurify.sanitize(html, {
             ALLOW_UNKNOWN_PROTOCOLS: true // DOMPurify usually strips non http(s) links from hrefs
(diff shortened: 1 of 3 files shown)


packages/ai-chat-ui/src/browser/chat-response-renderer/markdown-part-renderer.tsx
```diff
@@ -38,34 +38,51 @@ export class MarkdownPartRenderer implements ChatResponsePartRenderer<MarkdownCh
...
+        //     const html = this.markdownIt.render(response.content);
+        //     host.innerHTML = DOMPurify.sanitize(html, {
+        //         ALLOW_UNKNOWN_PROTOCOLS: true // DOMPurify usually strips non http(s) links from hrefs
+        //     });
+        //     while (ref?.current?.firstChild) {
+        //         ref.current.removeChild(ref.current.firstChild);
+        //     }
+
+        //     ref?.current?.appendChild(host);
+        // }, [response.content]);
```
sdirix, line 56: should be removed

packages/ai-chat-ui/src/browser/chat-tree-view/chat-view-tree-widget.tsx
```diff
@@ -389,6 +377,27 @@ export class ChatViewTreeWidget extends TreeWidget {
...
+        const html = markdownIt.render(text);
+        host.innerHTML = DOMPurify.sanitize(html, {
+            ALLOW_UNKNOWN_PROTOCOLS: true // DOMPurify usually strips non http(s) links from hrefs
+        });
+        while (ref?.current?.firstChild) {
+            ref.current.removeChild(ref.current.firstChild);
+        }
+
+        ref?.current?.appendChild(host);
+    }, [node.request]);
```
sdirix, line 396: This is almost the same content as in the response renderer. I would like to suggest a custom hook `useMarkDownRendering(markDownText)` so we can reuse this logic in both.

The hook could then also be documented on why it exists and why it uses markdownit directly.

packages/ai-chat/src/common/chat-model.ts
```diff
@@ -411,35 +410,35 @@ export class TextChatResponseContentImpl implements TextChatResponseContent {
 
 export class MarkdownChatResponseContentImpl implements MarkdownChatResponseContent {
     kind: 'markdownContent' = 'markdownContent';
-    protected _content: MarkdownStringImpl = new MarkdownStringImpl();
+    protected _content: string;
```
sdirix, line 413: shouldn't we keep the markdown string content? depending on its attributes we could then hand over different options to markdownit. Thereby we could allow html embeddings if someone wants to have that.

packages/ai-chat-ui/src/browser/chat-response-renderer/markdown-part-renderer.tsx
```diff
@@ -38,34 +38,53 @@ export class MarkdownPartRenderer implements ChatResponsePartRenderer<MarkdownCh
...
+
+        // return <div ref={ref}></div>;
+        return <MarkdownRender response={response} />;
     }
 
 }
 
-export const MarkdownWrapper = (props: { data: MarkdownString, renderCallback: (md: MarkdownString) => HTMLElement }) => {
+const MarkdownRender = ({ response }: { response: MarkdownChatResponseContent | InformationalChatResponseContent }) => {
     // eslint-disable-next-line no-null/no-null
```
sdirix, line 56:
```suggestion
```

packages/ai-chat-ui/src/browser/chat-response-renderer/markdown-part-renderer.tsx
```diff
@@ -38,34 +38,53 @@ export class MarkdownPartRenderer implements ChatResponsePartRenderer<MarkdownCh
...
+ * overriden by theia with a monaco version. This monaco version strips all html
+ * tags from the markdown with empty content.
+ * This leads to unexpected behavior when rendering markdown with html tags.
+ *
+ * @param markdown the string to render as markdown
+ * @returns the ref to use in an element to render the markdown
+ */
+export const useMarkdownRendering = (markdown: string) => {
+    // eslint-disable-next-line no-null/no-null
+    const ref = useRef<HTMLDivElement | null>(null);
```
sdirix, line 74: Is there a benefit of using `null` over `undefined`?

packages/ai-chat-ui/src/browser/chat-response-renderer/markdown-part-renderer.tsx
```diff
@@ -38,34 +38,53 @@ export class MarkdownPartRenderer implements ChatResponsePartRenderer<MarkdownCh
...
     render(response: MarkdownChatResponseContent | InformationalChatResponseContent): ReactNode {
         // TODO let the user configure whether they want to see informational content
         if (InformationalChatResponseContent.is(response)) {
             // null is valid in React
             // eslint-disable-next-line no-null/no-null
             return null;
         }
-        return <MarkdownWrapper data={response.content} renderCallback={this.renderMarkdown.bind(this)}></MarkdownWrapper>;
+
+        // return <div ref={ref}></div>;
```
sdirix, line 49:
```suggestion
```

packages/ai-chat-ui/src/browser/chat-response-renderer/markdown-part-renderer.tsx
```diff
@@ -38,34 +38,53 @@ export class MarkdownPartRenderer implements ChatResponsePartRenderer<MarkdownCh
...
+ * @param markdown the string to render as markdown
+ * @returns the ref to use in an element to render the markdown
+ */
+export const useMarkdownRendering = (markdown: string) => {
+    // eslint-disable-next-line no-null/no-null
+    const ref = useRef<HTMLDivElement | null>(null);
+    useEffect(() => {
+        const markdownIt = markdownit();
+        const host = document.createElement('div');
+        const html = markdownIt.render(markdown);
```
sdirix, line 78:
```suggestion
        // TODO: accept MarkdownString objects and set options accordingly
        const html = markdownIt.render(markdown);
```

sdirix (commented):
Works for me!

I tested requests like:
```
Hi <span> there </span>
```

And I executed prompts like:
```
Paste me an html template. Do not wrap in code block
```

These requests and answers were shown correctly :+1: 
---
I only have minor comments

packages/ai-chat-ui/src/browser/chat-response-renderer/markdown-part-renderer.tsx
eneufeld: the ref paramerter wants null, can't be used with undefined

sdirix (approved):
Works great

PR #14213, opened 2024-09-23
Wrap api objects returned to clients in a proxy

### LINKED ISSUE #13522: Vscode extension LaTeX Workshop's Viewer does not open in Theia IDE in browser
### Bug Description:

The [LaTeX Workshop](https://marketplace.visualstudio.com/items?itemName=James-Yu.latex-workshop) vscode extension's Viewer does not work on [Theia IDE](https://github.com/eclipse-theia/theia-blueprint) in the browser.

(I considered opening this as a Discussion but decided on an Issue/Bug because this involves a vscode extension working in vscode but specifically not working on Theia.  I downloaded the extension on vscode locally and the extension's Viewer did open, so it seems that the problem has to do with Theia.  Moreover, while [the LaTeX Workshop vscode extension does not officially support Eclipse Theia](https://github.com/James-Yu/LaTeX-Workshop/wiki/FAQ#eclipse-theia-is-not-supported), Theia does aim to support vscode extensions. So I can still expect that this extension should fully work on Theia, including on Theia IDE &mdash; correct?)

### Steps to Reproduce:

1. Run [Theia IDE](https://github.com/eclipse-theia/theia-blueprint) in the browser.
2. Install the [LaTeX Workshop](https://marketplace.visualstudio.com/items?itemName=James-Yu.latex-workshop) vscode extension (and refresh the browser page).
3. Navigate in the Explorer to folder containing a latex file and its corresponding compiled pdf.  Open the latex file in the Editor.
4. Click on the extension's View LaTeX PDF file button </img> at the top-right of the editor (or select this action via the command palette). (Or, select any of the extension's other actions beginning with "View LaTeX PDF" from the command palette &mdash; none of them worked for me.)
7. Expect to see the extension's Viewer open, and observe that it does not open.  (Compare to vscode where it does open.)

### Additional Information

This question is related to https://github.com/eclipse-theia/theia/discussions/7988, but that discussion was never answered (and in any case seems outdated, perhaps even dating to before Theia implemented vscode extensions at runtime?), so I would like to restart this discussion. Moreover, that is not the exact issue that I have. For me the LaTeX Workshop's Viewer is not even opening.

### System Information

- Operating System that Theia IDE is running on: Ubuntu 24.04
- Operating System that Theia IDE is accessed via: outdated Chrome Version 96.0.4664.110 on outdated macOS 10.15.7
- Theia Version: Version 1.46.1 (VS Code API Version: 1.85.1)

tsmaeder:
#### What it does
This PR wraps class-derived objects in the VS Code API inside a proxy. Additionally, any functions that are defined in the class prototype are bound to the object in question. This guards against the misuse of object methods as pure functions. As an additional bonus, clients no longer have access to the internals of the API objects like the main/ext proxies. 

Fixes #13522

#### How to test
This one is hard to test comprehensively as it covers a large code surface. The best way is to work with a wide variety of plugins to verify that we have not missed any subtle interactions with the API object implementation. 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/plugin-context.ts
+++ b/packages/plugin-ext/src/plugin/plugin-context.ts
@@ -278,6 +278,21 @@ import { NotebookEditorsExtImpl } from './notebook/notebook-editors';
 import { TestingExtImpl } from './tests';
 import { UriExtImpl } from './uri-ext';
 
+export function createAPIObject<T extends Object>(rawObject: T): T {
+    return new Proxy({}, {
+        get(target, p, receiver) {
+            const isOwnProperty = !!Object.getOwnPropertyDescriptor(rawObject, p);
+            const val = Reflect.get(rawObject, p);
+            if (!isOwnProperty && typeof val === 'function') {
+                // bind functions that are inherited from the prototype to the object itself. 
+                // This should handle the case of events.
+                return val.bind(rawObject);
+            }
+            return val;
+        },
+    }) as T;
+}
+
 export function createAPIFactory(
     rpc: RPCProtocol,
     pluginManager: PluginManager,
@@ -492,7 +507,8 @@ export function createAPIFactory(
                 return quickOpenExt.showQuickPick(plugin, items, options, token);
             },
             createQuickPick<T extends theia.QuickPickItem>(): theia.QuickPick<T> {
-                return quickOpenExt.createQuickPick(plugin);
+
+                return createAPIObject(quickOpenExt.createQuickPick(plugin));
             },
             showWorkspaceFolderPick(options?: theia.WorkspaceFolderPickOptions): PromiseLike<theia.WorkspaceFolder | undefined> {
                 return workspaceExt.pickWorkspaceFolder(options);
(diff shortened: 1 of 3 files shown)


tsmaeder:
`yarn browser test` works on Windows, but not Linux. I was able to reproduce the failures in a VM. Investigating....

packages/plugin-ext/src/plugin/plugin-context.ts
```diff
@@ -278,6 +278,21 @@ import { NotebookEditorsExtImpl } from './notebook/notebook-editors';
...
+            const isOwnProperty = !!Object.getOwnPropertyDescriptor(rawObject, p);
+            const val = Reflect.get(rawObject, p);
+            if (!isOwnProperty && typeof val === 'function') {
+                // bind functions that are inherited from the prototype to the object itself.
+                // This should handle the case of events.
+                return val.bind(rawObject);
+            }
+            return val;
+        },
+    }) as T;
```
msujew, line 293: Suggestion: Using the `rawObject` as the target of the proxy seems to fix the tests. I assume some code (either our own or some plugin code) relies on more than just the `get` functionality of the proxy. Stuff like `has` (used for the `in` operator) should be implemented as well. 
```suggestion
    return new Proxy(rawObject, {
        get(target, p) {
            const isOwnProperty = !!Object.getOwnPropertyDescriptor(target, p);
            const val = Reflect.get(target, p);
            if (!isOwnProperty && typeof val === 'function') {
                // bind functions that are inherited from the prototype to the object itself.
                // This should handle the case of events.
                return val.bind(target);
            }
            return val;
        },
    }) as T;
```

tsmaeder: The idea was to make sure users of the proxy did not have access to the wrapped object. I believe it's impossible to access the target object when you have a proxy object, but I'm not 100% sure. Do you know more about this topic?

msujew: As far as I'm aware, the JS API does not expose the target of a `Proxy` object to outsiders.

msujew (approved):
Great, looks good to me and I can confirm that the Latex Workshop plugin now works as expected in Theia! 🎉

PR #14219, opened 2024-09-25
Fix window maximization when using splash screen

### LINKED ISSUE #14218: [electron] Splash screen behavior is inconsistent with regards to starting maximized/unmaximized
### Bug Description:
When a splash screen is configured by default the `preventAutomaticShow` option is set to true.
This means the main window should remain invisible until the layout is fully initialized.
This works as expected when the Theia electron app starts the main Window unmaximized.
However, when the app is started with `isMaximized`=true then the main Window will still show early.
Resulting in an overlay of the preload window and the splash screen

Expected behavior:
Using a splashscreen, the main window should remain invisible until the splash screen is closed independent of whether the window is maximized or not

### Steps to Reproduce:

1. Build the electron example app
2. Start the app 
3. If the window starts not maximized:
    (State. Splash screen is shown, main window is invisible until splash screen is removed)
    3.1 Maximize the window
    3.2 Close the app
    3.3 Restart the app and continue with 4.
4. If the window is maximized
   State: Splash screen is shown. Main window becomes immediately visible and is visible alongside splashscreen

Here is demonstration of the Bug using Theia IDE under Linux:

https://github.com/user-attachments/assets/0e70a6dd-57cf-42c6-a3bb-6a894076d368

### Additional Information
Reproducible with following OS/versions:
- Operating System: Ubunut 22.04, Windows 11
- Theia Version: Theia IDE 1.52, Current Theia master

tortmayr:
#### What it does

Avoid eager restore of maximized state in `TheiaElectronWindow.restoreMaximizeState`
Calling `window.maximize()` implicitly also makes the window visible.
Therefore we have to check if the window is already visibible before restoring t he maximizeState.
If the window is not yet visible yet we restore the state once the window is shown.

This fixes and issue were  both the splashscreen and the preload window are be visible at the same time.

Fixes #14218
#### How to test

1. Build electron example app
2. Start electron example app
3. Maximize window and close app
4. Start app again

Only the splash screen should be shown, main window should remain invisible until the splash screen is hidden.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-main/theia-electron-window.ts
+++ b/packages/core/src/electron-main/theia-electron-window.ts
@@ -186,10 +186,18 @@ export class TheiaElectronWindow {
     }
 
     protected restoreMaximizedState(): void {
-        if (this.options.isMaximized) {
-            this._window.maximize();
+        const restore = () => {
+            if (this.options.isMaximized) {
+                this._window.maximize();
+            } else {
+                this._window.unmaximize();
+            }
+        };
+
+        if (this._window.isVisible()) {
+            restore();
         } else {
-            this._window.unmaximize();
+            this._window.once('show', () => restore());
         }
     }
 


packages/core/src/electron-main/theia-electron-window.ts
```diff
@@ -196,6 +196,9 @@ export class TheiaElectronWindow {
     protected trackApplicationState(): void {
         this.toDispose.push(TheiaRendererAPI.onApplicationStateChanged(this.window.webContents, state => {
             this.applicationState = state;
+            if (state === 'ready' && this.options.preventAutomaticShow) {
+                this.restoreMaximizedState();
+            }
```
sdirix, line 201: I'm not sure that here is the best place to put the maximize logic. I agree that this fixes the issue, however this depends on the fact that we know that the main application will call `show` once the application state is `'ready'`.

What do you think of the following:
- If a Window is immediately shown (i.e. `this.options.show === true || typeof this.options.show === 'undefined'`), we directly call the `restoreMaximizedState`
- If a Window is not immediately shown we 
  - return a `Proxy` of `BrowserWindow`
  - The proxy overrides the `show` of `BrowserWindow`
  - On the first time `show` is called, we also implicitly call `restoreMaximizedState`

tortmayr: IMO using a proxy approach here is kind of an overkill. Maybe it would be better to move the maximization logic into the main application entirely. Here we know exactly when the main window will be shown and can invoke the maximization at this point if nencessary.
It could look like this:

```diff
diff
diff --git a/packages/core/src/electron-main/electron-main-application.ts b/packages/core/src/electron-main/electron-main-application.ts
index b464c916a84..13f7fe127b7 100644
--- a/packages/core/src/electron-main/electron-main-application.ts
+++ b/packages/core/src/electron-main/electron-main-application.ts
@@ -334,19 +334,22 @@ export class ElectronMainApplication {
                 });
                 if (this.isShowSplashScreen()) {
                     console.log('Showing splash screen');
-                    this.configureAndShowSplashScreen(this.initialWindow);
+                    this.configureAndShowSplashScreen(this.initialWindow, options.isMaximized);
                 }

                 // Show main window early if windows shall be shown early and splash screen is not configured
                 if (this.isShowWindowEarly() && !this.isShowSplashScreen()) {
                     console.log('Showing main window early');
                     this.initialWindow.show();
+                    if (options.isMaximized) {
+                        this.initialWindow.maximize();
+                    }
                 }
             });
         }
     }

-    protected async configureAndShowSplashScreen(mainWindow: BrowserWindow): Promise<BrowserWindow> {
+    protected async configureAndShowSplashScreen(mainWindow: BrowserWindow, isMaximized?: boolean): Promise<BrowserWindow> {
         const splashScreenOptions = this.getSplashScreenOptions()!;
         console.debug('SplashScreen options', splashScreenOptions);

@@ -382,6 +385,9 @@ export class ElectronMainApplication {
             cancelTokenSource.cancel();
             if (!mainWindow.isVisible()) {
                 mainWindow.show();
+                if (isMaximized) {
+                    mainWindow.maximize();
+                }
             }
             splashScreenWindow.close();
         };
diff --git a/packages/core/src/electron-main/theia-electron-window.ts b/packages/core/src/electron-main/theia-electron-window.ts
index 720865f8f59..4faa5852fb7 100644
--- a/packages/core/src/electron-main/theia-electron-window.ts
+++ b/packages/core/src/electron-main/theia-electron-window.ts
@@ -86,7 +86,6 @@ export class TheiaElectronWindow {
         if (!this.options.preventAutomaticShow) {
             this.attachReadyToShow();
         }
-        this.restoreMaximizedState();
         this.attachCloseListeners();
         this.trackApplicationState();
         this.attachReloadListener();
@@ -185,14 +184,6 @@ export class TheiaElectronWindow {
         return TheiaRendererAPI.requestClose(this.window.webContents, reason);
     }

-    protected restoreMaximizedState(): void {
-        if (this.options.isMaximized) {
-            this._window.maximize();
-        } else {
-            this._window.unmaximize();
-        }
-    }
-
     protected trackApplicationState(): void {
         this.toDispose.push(TheiaRendererAPI.onApplicationStateChanged(this.window.webContents, state => {
             this.applicationState = state;
  ```
WDYT?

Sidenote: options.show is currently set to false by default in Theia and not considered at all in the initialization logic.
I think we should keep it that way.

Comment (sdirix) [reply packages/core/src/electron-main/theia-electron-window.ts:201]: Fine by me in general. We just need to make sure that when the window is shown automatically, without the main application calling `show` itself, we also restore the maximized state.

**Comment by tsmaeder** [comment]
@sdirix @tortmayr what's the state on this one?

**Comment by sdirix** [comment]
> @sdirix @tortmayr what's the state on this one?

@tortmayr Can you update the PR as discussed?

**Comment by tortmayr** [comment]
Sorry, for the long roundtrip. I was busy with other projects and didn't find the time to work on this PR.
I have now pushed an update.
In the end I found a pretty simple solution for this issue.
In the `restoreMaximizeState` method we simply have to check if the window is already visible.
If so restore the state directly. 
If not we have listen for the next `show` event on the window and restore the state then.

**Review by sdirix — APPROVED** [review]
Works for me! Great final solution.

File: packages/core/src/electron-main/theia-electron-window.ts [hunk 2]
```diff
@@ -186,10 +186,18 @@ export class TheiaElectronWindow {
...
+            } else {
+                this._window.unmaximize();
+            }
+        };
+
+        if (this._window.isVisible()) {
+            restore();
         } else {
-            this._window.unmaximize();
+            this._window.once('show', () => restore());
```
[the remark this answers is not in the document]
tortmayr: We still `unmaximize`. The git diff just looks confusing here.
The restore function checks the isMaximized options and then calls either `maximize` or `unmaximize`

PR #14247, opened 2024-10-03
feat(ai): Update OpenAI models supporting structured output

planger:
#### What it does

Updates the list of OpenAI models supporting structured output. As `gpt-4o` now points to 'gpt-4o-2024-08-06', it now also support structured output. In addition, I added 'gpt-4o-latest', which always points to the latest release and thus supports structured output too.

#### How to test

Configure `gpt-4o` to be used for the terminal agent and test the terminal agent.
You can verify if it uses structured output by setting a breakpoint in `ai-terminal-agent.ts` at line 182.

#### Follow-ups

None

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-openai/src/node/openai-language-model.ts
+++ b/packages/ai-openai/src/node/openai-language-model.ts
@@ -134,9 +134,12 @@ export class OpenAiModel implements LanguageModel {
     }
 
     protected supportsStructuredOutput(): boolean {
-        // currently only the lastest 4o and 4o-mini models support structured output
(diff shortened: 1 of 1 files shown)


JonasHelming (approved):
LGTM!

Maybe we want to add "latest" to the default list, too?

planger:
Good point, but actually `-latest` doesn't seem to be available via the completion API. So it probably doesn't make sense to include it in the structured output check too. I removed it.

JonasHelming:
approval still valid :-)

PR #14260, opened 2024-10-07
Optimize showing recent workspaces

msujew:
#### What it does

I've recently noticed that attempting to execute the "Open recent workspace" command takes a few seconds on Windows. This change optimizes quite a lot of the old code to improve the performance and also fix a few issues regarding incorrectly displayed paths. In order to get the most performance out of it, I had to remove the pretty expensive `FileService#resolve` calls which were used to differentiate between file and directory workspaces.

#### How to test

1. Run the "Open recent workspace" command.
2. Assert that it effectively instantly opens the quick pick for the workspaces
3. Assert that there are no regressions.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/workspace/src/browser/quick-open-workspace.ts
+++ b/packages/workspace/src/browser/quick-open-workspace.ts
@@ -18,10 +18,7 @@ import { injectable, inject, optional } from '@theia/core/shared/inversify';
 import { QuickPickItem, LabelProvider, QuickInputService, QuickInputButton, QuickPickSeparator } from '@theia/core/lib/browser';
 import { EnvVariablesServer } from '@theia/core/lib/common/env-variables';
 import { WorkspaceService } from './workspace-service';
-import { WorkspacePreferences } from './workspace-preferences';
 import URI from '@theia/core/lib/common/uri';
-import { FileService } from '@theia/filesystem/lib/browser/file-service';
-import { FileStat } from '@theia/filesystem/lib/common/files';
 import { nls, Path } from '@theia/core/lib/common';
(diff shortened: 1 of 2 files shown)


packages/workspace/src/browser/quick-open-workspace.ts
```diff
@@ -48,45 +42,34 @@ export class QuickOpenWorkspace {
     };
 
     async open(workspaces: string[]): Promise<void> {
-        this.items = [];
-        const [homeDirUri] = await Promise.all([
-            this.envServer.getHomeDirUri(),
-            this.workspaceService.getUntitledWorkspace()
-        ]);
-        const home = new URI(homeDirUri).path.toString();
-        await this.preferences.ready;
```
sdirix, line 57: Do you know why this was awaited and why it's no longer necessary?

sdirix (approved):
Works for me. However I only tested the Browser variant on Linux.

packages/workspace/src/browser/quick-open-workspace.ts
msujew: Back in https://github.com/eclipse-theia/theia/commit/543b119a32652f662968411370979ea002a1736a#diff-52cd48457988091d685e5eb18c42cab1159c2a17740bd035d5cc1dd9ccd07464, it seems like the preferences were actually used. However, the usage was removed in https://github.com/eclipse-theia/theia/commit/3ac912faf25c298c4af4578311562d23dde257c0#diff-52cd48457988091d685e5eb18c42cab1159c2a17740bd035d5cc1dd9ccd07464, without cleaning up call to `ready`.

Issue #13520, opened 2024-03-23
BrowserFS Deprecated


dr-vortex:
Hi there, I'm a maintainer of BrowserFS. I've been going though the [dependents](https://www.npmjs.com/package/browserfs?activeTab=dependents) of BrowserFS.

Please see https://github.com/browser-fs/NOTICE

Would you be interested in updating your project to use the updated fork, [ZenFS](https://github.com/zen-fs/core)?

JonasHelming:
@sdirix FYI

james-pre:
@sdirix @msujew @JonasHelming 

Any updates on this?

james-pre:
@JonasHelming @sdirix @msujew,

 Any updates on this? It has been a few months. Please let me know if you have any questions or concerns. I may even be willing to work on a PR for this due to the size of this project.

msujew:
@james-pre We have a PR in review for this over at https://github.com/eclipse-theia/theia/pull/14125.

sdirix:
Hi @james-pre, thanks for the notice and your efforts on `zenfs`.

We evaluated the available options and decided to switch to an [OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system) based approach. OPFS is browser native and offers all required functionality for a basic file system support within the Browser. While this is a bit more effort to implement compared to switching to a Node-fs-alternative like `zenfs`, it has the benefit of eliminating this dependency from the base Theia framework.

Any Theia-based application which wants to leverage `zenfs` can do so by implementing their own `FileSystemProvider`.

The switch to OPFS is implemented in https://github.com/eclipse-theia/theia/pull/14263

PR #14263, opened 2024-10-07
Switch from browserfs to OPFS

robertjndw:
#### What it does

This PR builds on top of the work that was done in PR https://github.com/eclipse-theia/theia/pull/12853.
As browserfs was deprecated in March 24 (see [original repo](https://github.com/jvilk/BrowserFS?tab=readme-ov-file)), it is beneficial to switch to an alternative. 
As discussed in #14125, the direct successor of browserfs is zenfs. However, during the research I also stumbled across [OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system) which is a new native Browser API which covers a lot of the features necessary for this use case.

#### How to test

Similarly to the previously mentioned PR https://github.com/eclipse-theia/theia/pull/12853.

For debugging purposes of the OPFS, I can really recommend this extension for the Chrome DevTools: https://github.com/tomayac/opfs-explorer

#### Follow-ups

The watch functionality is currently missing. Since OPFS does not provide a native API for this, it has to be manually tracked and implemented

#### Review checklist

#### Reminder for reviewers


sdirix (changes requested):
Thanks for the work :heart: ! For me it mostly works ;)

Please have a look at my comments and make sure that all automated CI checks go through.

james-pre:
Could a maintainer please link this to #13520 so #13520 will be auto-closed on merge? Thanks!

packages/filesystem/src/browser-only/opfs-filesystem-initialization.ts
```diff
@@ -0,0 +1,36 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import type { OPFSFileSystemProvider } from './opfs-filesystem-provider';
+import { injectable } from '@theia/core/shared/inversify';
+
+export const OPFSInitialization = Symbol('OPFSInitialization');
+export interface OPFSInitialization {
+    createMountableFileSystem(): Promise<FileSystemDirectoryHandle>
+    initializeFS: (fs: FileSystemDirectoryHandle, provider: OPFSFileSystemProvider) => Promise<void>;
```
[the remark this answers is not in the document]
robertjndw: I see your point about simplifying the initialization into one method now that we're exclusively using OPFS. However, I think keeping the two methods still makes sense in this context. 
One focuses on returning the root directory, while the other is responsible for preparing the file system, which is especially useful when we want to create files and folders in advance for the example. What are your thoughts on this approach?

packages/filesystem/src/browser-only/opfs-filesystem-provider.ts
```diff
@@ -0,0 +1,286 @@
...
+        } catch (error) {
+            throw this.toFileSystemProviderError(error);
+        } finally {
+            this.onDidChangeFileEmitter.fire([{ resource, type: FileChangeType.DELETED }]);
+        }
+    }
+    async rename(from: URI, to: URI, opts: FileOverwriteOptions): Promise<void> {
+        await this.initialized;
+        try {
+            const content = await this.readFile(from);
```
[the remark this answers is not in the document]
robertjndw: Thanks for catching that bug! I missed it in the initial implementation, but it should be fixed now 👍🏼

robertjndw:
@sdirix Can you maybe take a look at the CI/CD error? It seems that with the addition of `dom.asynciterable` to the tsconfig something broke 🤔

sdirix:
> @sdirix Can you maybe take a look at the CI/CD error? It seems that with the addition of `dom.asynciterable` to the tsconfig something broke 🤔

---

>@theia/ffmpeg: $ eslint --cache=true --no-error-on-unmatched-pattern=true "{src,test}/**/*.{ts,tsx}"
@theia/re-exports: 
@theia/re-exports: /home/runner/work/theia/theia/dev-packages/private-re-exports/src/bin-package-re-exports-from-package.ts
@theia/re-exports:   0:0  error  Parsing error: Invalid value for lib provided: dom.asynciterable

---

Seems like the Typescript Eslint integration does not like the `dom.asynciterable` lib. I'm not sure why this is the case as the Typescript we are using is definitely compatible to it.

Anyway, you can override the lib settings in `configs/base.eslintrc.json` and there specify the `parserOptions.lib` to the previous values, i.e.
```json
    "lib": [
      "ES2019",
      "ES2020.Promise",
      "DOM"
    ]
```

That should fix the issue

james-pre:
LGTM! @robertjndw any updates on this?

packages/filesystem/src/browser-only/opfs-filesystem-provider.ts
```diff
@@ -0,0 +1,345 @@
...
+                try {
+                    // Determine the type of the entry (file or directory)
+                    if (handle.kind === 'file') {
+                        result.push([name, FileType.File]);
+                    } else if (handle.kind === 'directory') {
+                        result.push([name, FileType.Directory]);
+                    }
+                } catch (error) {
+                    // Ignore errors for individual entries, log them and continue
+                    // console.error(error);
```
sdirix, line 134: Are any actual errors thrown here? What does this try-catch protect against?

examples/api-samples/src/browser-only/filesystem/example-filesystem-initialization.ts
```diff
@@ -17,35 +17,36 @@
...
-            }
-            if (!fs.existsSync('/home/workspace2')) {
-                await provider.mkdir(new URI('/home/workspace2'));
-                await provider.writeFile(new URI('/home/workspace2/my-file.json'), this.encodingService.encode('{ foo: true }').buffer, { create: true, overwrite: false });
-                await provider.writeFile(new URI('/home/workspace2/my-file2.json'), this.encodingService.encode('{ bar: false }').buffer, { create: true, overwrite: false });
-            }
+            await provider.readdir(new URI('/home/workspace'));
         } catch (e) {
-            console.error('An error occurred while initializing the demo workspaces', e);
+            console.error('An error occurred while reading the demo workspaces', e);
```
sdirix, line 34: We should check whether the error is a file not found error. Only then we should create files. Also we should not log the error as this is an expected error for any first time user.

Only other errors than the expected ones should be logged.

sdirix: Even better would be to offer some kind of "exists" method so that we don't need to go through try-catch here.

packages/filesystem/src/browser-only/opfs-filesystem-provider.ts
robertjndw: You're right. The type system should protect us against any other type than file and directory

examples/api-samples/src/browser-only/filesystem/example-filesystem-initialization.ts
robertjndw: I agree and `exists` method would definitely make sense. Unfortunately, it is not yet supported by the OPFS, but we can implement a workaround until then 👍🏼

tsmaeder:
@sdirix what's the state on this one?

sdirix:
@tsmaeder I have to review the latest changes

sdirix (approved):
Looks good to me! Great work!

PR #14277, opened 2024-10-09
blur cell on shift+enter, only update cell height when changed

jonah-iden:
#### What it does

Small fix for notebook shift+enter keybinding. When having an editor focused it should blur the previously selected, if a new cell is created that should be focused, but if switching to an existing cell there should be no editor focus.
Also small fix so that the height changed event only fired if the height acutally changed

#### How to test

Open a notebook.
Edit a cell and use shift+enter to execute it and focus the next one or create a next one.
The next cell should be selected and have text focus

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/view-model/notebook-cell-model.ts
+++ b/packages/notebook/src/browser/view-model/notebook-cell-model.ts
@@ -274,8 +274,10 @@ export class NotebookCellModel implements NotebookCell, Disposable {
     }
 
     set cellHeight(height: number) {
-        this.onDidCellHeightChangeEmitter.fire(height);
-        this._cellheight = height;
+        if (height !== this._cellheight) {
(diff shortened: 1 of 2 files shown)


dhuebner (changes requested):
Works good except in case of the last cell. A new cell is created, but didn't grab the editor focus.

dhuebner:
@jonah-iden 
Could you please also check the behavior of `ALT+Enter` ?

Btw. the markdown cell works a bit different in both cases Shift+Enter  and Alt+Enter.

jonah-iden:
sure thanks for testing

jonah-iden:
ups i just realized i remembered vscode behaviour wrong... 
I'll rework it so its aligned

dhuebner (approved):
Works as expected, thanks!

PR #14281, opened 2024-10-10
feat: add support for llamafile as ai model provider

eneufeld:
#### What it does

This adds support to add llms run from llamafiles.

- The configured llamafiles are stored in the preferences
- The llamafile output is send to a new output channel <mymodel>-llamafile
- As a user you have to provide:
  - a custom modelname
  - the file uri to the llamafile
  - the port to run the llama.cpp server on

The current implementation does not support tools/functions so that agents like the workspace-agent don't work well with models provided via llamafiles.

#### How to test

Download a llamafile and follow the instructions on how to make it runnable from here: https://github.com/Mozilla-Ocho/llamafile?tab=readme-ov-file#quickstart
In Theia open the preferences, go to llamafile in the ai section and add a new entry.
You need to provide a name, uri (the file uri to the llamafile, e.g file:///home/me/Downloads/my.llamafile) and a port (e.g 30000).
Adopt one agent to use the new model, e.g the universal agent.
Then Call the "Start Llamafile" command.
See that there is a new output channel (my local model-llamafile).
Say "hi" in the chat and see your cpu/gpu working overtime to produce an answer ;-) .

#### Review checklist

#### Reminder for reviewers


packages/ai-llamafile/src/browser/llamafile-frontend-application-contribution.ts
```diff
@@ -0,0 +1,63 @@
...
+            const llamafiles = this.preferenceService.get<LlamafileEntry[]>(PREFERENCE_LLAMAFILE, []);
+
+            const models = llamafiles.map(llamafile =>
+                LlamafileLanguageModel.createNewLlamafileLanguageModel(llamafile.name, llamafile.uri, llamafile.port, this.llamafileServerManager));
+            this.languageModelRegistry.addLanguageModels(models);
+
+            this.preferenceService.onPreferenceChanged(event => {
+                if (event.preferenceName === PREFERENCE_LLAMAFILE) {
+                    // old models in the preference
+                    const oldModels = new Set((event.oldValue as LlamafileEntry[]).map(v => v.name));
```
sdirix, line 47: `oldValue` will be `undefined` when the preference was not set before. In that case the `new Set` will actually result in `undefined` too and `.map` will throw an error.

packages/ai-llamafile/src/browser/llamafile-frontend-application-contribution.ts
```diff
@@ -0,0 +1,63 @@
...
+                    const newModels = (event.newValue as LlamafileEntry[]).reduce((acc, v) => { acc.set(v.name, v); return acc; }, new Map<string, LlamafileEntry>());
+
+                    const modelsToRemove = [...oldModels.values()].filter(model => !newModels.has(model));
+                    const modelDescriptionsToAdd = [...newModels.values()].filter(model => !oldModels.has(model.name));
+
+                    this.languageModelRegistry.removeLanguageModels(modelsToRemove);
+
+                    const modelsToAdd = modelDescriptionsToAdd.map(llamafile =>
+                        LlamafileLanguageModel.createNewLlamafileLanguageModel(llamafile.name, llamafile.uri, llamafile.port, this.llamafileServerManager));
+                    this.languageModelRegistry.addLanguageModels(modelsToAdd);
```
sdirix, line 58: We directly add the language models in the frontend here. We should not do that as this will not work in a Cloud deployment, e.g. in "Try Online" without opening up additional ports. As the ports are even configurable, these is even harder.

Instead we should manage the language model in the backend as we do with ollama and openai.

packages/ai-llamafile/src/browser/llamafile-preferences.ts
```diff
@@ -0,0 +1,63 @@
...
+export const PREFERENCE_LLAMAFILE = 'ai-features.llamafile.llamafiles';
+
+export const aiLlamafilePreferencesSchema: PreferenceSchema = {
+    type: 'object',
+    properties: {
+        [PREFERENCE_LLAMAFILE]: {
+            title: AI_LLAMAFILE_PREFERENCES_TITLE,
+            markdownDescription: '❗ This setting allows you to add llamafiles.\
+            \n\
+            You need to provide a user friendly name, the path to the llamafile and the port to use.\
```
sdirix, line 30:
```suggestion
            You need to provide a user friendly `name`, the file `uri` to the llamafile and the `port` to use.\
```

packages/ai-llamafile/src/common/llamafile-language-model.ts
```diff
@@ -0,0 +1,131 @@
...
+            // const data = await response.json();
+            // if (data && data.content) {
+            //     return {
+            //         text: data.content
+            //     };
+            // } else {
+            //     return {
+            //         text: 'No content field found in the response.'
+            //     };
+            // }
```
sdirix, line 118: can be removed?

packages/ai-llamafile/src/common/llamafile-language-model.ts
```diff
@@ -0,0 +1,131 @@
...
+            console.error('Error:', error);
+            return {
+                text: `Error: ${error}`
+            };
+        }
+    }
+
+    static createNewLlamafileLanguageModel(name: string, uri: string, port: number, serverManager: LlamafileServerManager): LlamafileLanguageModel {
+        return new LlamafileLanguageModel(name, uri, port, serverManager);
+    }
```
sdirix, line 129: can we remove this static util? does not seem to be useful for anything

packages/ai-llamafile/src/common/llamafile-server-manager.ts
```diff
@@ -0,0 +1,29 @@
...
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+export const LlamafileServerManager = Symbol('LlamafileServerManager');
+
+export const LlamafileServerManagerPath = '/services/llamafileservermanager';
+
+export interface LlamafileServerManager {
```
sdirix, line 20: This could be adapted to be the `LlamafileManager` and also handle the language models on the backend.

sdirix (changes requested):
Works for me!

My major suggestion is to move the language model into the backend, see my comments. Also make sure that all CI checks pass. Thanks!

sdirix (changes requested):
Two more things:

Note: We should also add the new extension to the Electron example

packages/ai-llamafile/src/browser/llamafile-command-contribution.ts
```diff
@@ -0,0 +1,91 @@
...
+                        return;
+                    }
+                    const options = llamaFiles.map(llamaFile => ({ label: llamaFile }));
+                    const result = await this.quickInputService.showQuickPick(options);
+                    if (result === undefined) {
+                        return;
+                    }
+                    this.llamafileManager.stopServer(result.label);
+                } catch (error) {
+                    console.error('Something went wrong during the llamafile stop.', error);
```
sdirix, line 86: We should also here use the messageService

packages/ai-llamafile/src/node/llamafile-manager-impl.ts
```diff
@@ -0,0 +1,92 @@
...
+    @inject(LanguageModelRegistry)
+    protected languageModelRegistry: LanguageModelRegistry;
+
+    private processMap: Map<string, ChildProcessWithoutNullStreams> = new Map();
+    private client: LlamafileServerManagerClient;
+
+    addLanguageModels(llamaFiles: LlamafileEntry[]): void {
+        const models = llamaFiles.map(llamafile =>
+            new LlamafileLanguageModel(llamafile.name, llamafile.uri, llamafile.port));
+        this.languageModelRegistry.addLanguageModels(models);
```
sdirix, line 36: This unconditionally adds language models to the registry. However this method can be called mutliple times, e.g. when there are mutliple frontends open. We should not add duplicates to the registry.

packages/ai-llamafile/src/browser/llamafile-frontend-application-contribution.ts
```diff
@@ -0,0 +1,53 @@
...
+                    // old models in the preference
+                    const oldModels = new Set(((event.oldValue ?? []) as LlamafileEntry[]).map(v => v.name));
+                    // new models in the preference as map to if a model was added and be able to create the llm and register it
+                    const newModels = (event.newValue as LlamafileEntry[]).reduce((acc, v) => { acc.set(v.name, v); return acc; }, new Map<string, LlamafileEntry>());
+
+                    const modelsToRemove = [...oldModels.values()].filter(model => !newModels.has(model));
+                    this.llamafileManager.removeLanguageModels(modelsToRemove);
+
+                    const modelDescriptionsToAdd = [...newModels.values()].filter(model => !oldModels.has(model.name));
+                    this.llamafileManager.addLanguageModels(modelDescriptionsToAdd);
```
sdirix, line 48: - This fails when the user edits the preferences, e.g. I'm editing the preferences and did not yet enter a port. A change is already triggered and the language model with an invalid port created. After I finish the editing with a valid port, the language model will never be updated with it as we always filter here. The preference will be broken until a full backend restart and/or a change to the name.
 - We should not even create a language model if the entry in the settings is incomplete, e.g. when the port is missing.

sdirix (changes requested):
I found some more usability issues as I was testing the latest state. Please have a look at the open ai language model implementation which is able to handle these cases (i.e. duplicate handling and update of language models after the fact)

sdirix (approved):

PR #14282, opened 2024-10-10
Start menu handler ids at one, not zero

### LINKED ISSUE #14279: First Menu Action not Working with Emtpy Main Menu
### Bug Description:
An adopter has a product that does not have a main menu bar originally. When opening the "manage" menu (the gear icon at the bottom of the left side bar), the first menu item ("Settings" in their case) does not have any effect the first time the menu is opened. When reopening the menu, everything works fine.

### Additional Information

- Operating System: Linux
- Theia Version: 1.48.3 Linux

tsmaeder:
#### What it does
In order to handle native menu invocations on the browser side, we allocate handler ids when that we send over to the electron-main side. On the electron-main side, we check for the handler id with `if (menu.handlerId)` before invoking the menu action. Since the handler ids start at 0, the very first menu action created could never be invoked.

Fixes #14279

#### How to test
This can't really be tested in our example applications. But I think I can convince a reviewer that the change at least does no harm: we just need to make sure menus and context menus still work as advertised.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-browser/preload.ts
+++ b/packages/core/src/electron-browser/preload.ts
@@ -35,8 +35,8 @@ const { ipcRenderer, contextBridge } = require('electron');
 
 // a map of menuId => map<handler id => handler>
 const commandHandlers = new Map<number, Map<number, () => void>>();
-let nextHandlerId = 0;
-const mainMenuId = 0;
+let nextHandlerId = 1;
+const mainMenuId = 1;
 let nextMenuId = mainMenuId + 1;
 
 let openUrlHandler: ((url: string) => Promise<boolean>) | undefined;


msujew (approved):
Looks good, however, regarding:

> On the electron-main side, we check for the handler id with if (menu.handlerId) before invoking the menu action. 

Does it make sense to rewrite this to `if (typeof menu.handlerId === 'number')`? We also do this in the plugin system to ensure that `0` is a valid value.

tsmaeder:
> Looks good, however, regarding:
> 
> > On the electron-main side, we check for the handler id with if (menu.handlerId) before invoking the menu action.
> 
> Does it make sense to rewrite this to `if (typeof menu.handlerId === 'number')`? We also do this in the plugin system to ensure that `0` is a valid value.

While it is the better check in this case, I think it's better to fix this problem at the source, i.e. the generation of the id's in general. Here's why: the client does not have to care whether the id is a number or a string or undefined: the empty string might well be the in indication of "no id" and is often not considered a valid id. It is hard to make sure that we do the check correctly for the type of id in all places, but if we ensure valid id values are never falsy, it just doesn't matter how we check for it. It's more robust as a policy, IMO.

PR #14295, opened 2024-10-11
Set menu bar less often at startup

### LINKED ISSUE #14280: Calling `setMenu()` More than 400 Times at Startup
### Bug Description:
While debugging a problem, I noticed we invoke `ElectronAPI.setMenu()` over 400 times at startup. It would be good to reduce that.

### Steps to Reproduce:
Put a log statement in the core `preload.ts#setMenu()` implementation. 

### Additional Information

- Operating System: Windows
- Theia Version: 1.54.x master

tsmaeder:
#### What it does
Debugging some code, I found out we're setting the main menu bar about 400 times at startup. This change reduces the number to around 10.

Fixes #14280

#### How to test
Make sure the main menu bar is set correctly, including checkmarks (example "View->Toggle Minimap").
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-browser/menu/electron-main-menu-factory.ts
+++ b/packages/core/src/electron-browser/menu/electron-main-menu-factory.ts
@@ -74,42 +74,45 @@ export type ElectronMenuItemRole = ('undo' | 'redo' | 'cut' | 'copy' | 'paste' |
 @injectable()
 export class ElectronMainMenuFactory extends BrowserMainMenuFactory {
 
-    protected _menu?: MenuDto[];
-    protected _toggledCommands: Set<string> = new Set();
+    protected menu?: MenuDto[];
+    protected toggledCommands: Set<string> = new Set();
 
     @inject(PreferenceService)
     protected preferencesService: PreferenceService;
 
+    setMenuBar = debounce(() => this.doSetMenuBar());
+
     @postConstruct()
     postConstruct(): void {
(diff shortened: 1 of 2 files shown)


packages/core/src/electron-browser/menu/electron-main-menu-factory.ts
```diff
@@ -74,57 +74,58 @@ export type ElectronMenuItemRole = ('undo' | 'redo' | 'cut' | 'copy' | 'paste' |
...
 
-    protected _menu?: MenuDto[];
-    protected _toggledCommands: Set<string> = new Set();
+    protected menu?: MenuDto[];
+    protected toggledCommands: Set<string> = new Set();
 
     @inject(PreferenceService)
     protected preferencesService: PreferenceService;
 
+    setMenuBar = debounce(() => this.doSetMenuBar());
```
msujew, line 83: Minor: Having an explicit delay value here would be good. I never know, what the default for this is.
```suggestion
    setMenuBar = debounce(() => this.doSetMenuBar(), 10);
```

msujew (approved):
Works great, thanks!

packages/core/src/electron-browser/menu/electron-main-menu-factory.ts
tsmaeder: @msujew I upped the value to 100, since updating the main menu 100 times per second seems a bit excessive ;-)

Issue #13796, opened 2024-06-11
VSX Registry: workbench.extensions.uninstallExtension is not implemented


safisa:
### Bug Description:
Hi,
The vscode builtin command workbench.extensions.uninstallExtension is not implemented in Theia

### Additional Information

- Operating System: Mac
- Theia Version: 1.50

msujew:
Resolved by https://github.com/eclipse-theia/theia/issues/15600.

Issue #13795, opened 2024-06-11
VSX Registry: add version support to the workbench.extensions.installExtension command


safisa:
### Bug Description:
Hi,
The vscode builtin command workbench.extensions.installExtension supports specifying the extension version as part of the extension id: extId@version
In Theia this pattern is not supported.

See this vscode PR:
https://github.com/microsoft/vscode/pull/136314

### Additional Information

- Operating System: Mac
- Theia Version: 1.50

msujew:
Resolved by https://github.com/eclipse-theia/theia/pull/14298.

PR #14298, opened 2024-10-11
Support version for installExtension, support uninstall cmd

dhuebner:
#### What it does

Allows to use version when executing `workbench.extensions.installExtension` command in format `<extensionId>@<version>`  #13795
Adds support for `workbench.extensions.uninstallExtension` command #13796

#### How to test

Run uninstall on some installed extension. `extension.uninstall` now uses the new `workbench.extensions.uninstallExtension`  under the hood. 

Install command `workbench.extensions.installExtension` is currently not used in Theia with a `string` parameter, one need to create a separate command.
 
#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
+++ b/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
@@ -52,7 +52,7 @@ import { DiffService } from '@theia/workspace/lib/browser/diff-service';
 import { inject, injectable, optional } from '@theia/core/shared/inversify';
 import { Position } from '@theia/plugin-ext/lib/common/plugin-api-rpc';
 import { URI } from '@theia/core/shared/vscode-uri';
-import { PluginServer } from '@theia/plugin-ext/lib/common/plugin-protocol';
+import { PluginDeployOptions, PluginIdentifiers, PluginServer } from '@theia/plugin-ext/lib/common/plugin-protocol';
 import { TerminalFrontendContribution } from '@theia/terminal/lib/browser/terminal-frontend-contribution';
 import { QuickOpenWorkspace } from '@theia/workspace/lib/browser/quick-open-workspace';
 import { TerminalService } from '@theia/terminal/lib/browser/base/terminal-service';
@@ -108,6 +108,10 @@ export namespace VscodeCommands {
 
     export const INSTALL_EXTENSION_FROM_ID_OR_URI: Command = {
         id: 'workbench.extensions.installExtension'
+    };
+
+    export const UNINSTALL_EXTENSION: Command = {
+        id: 'workbench.extensions.uninstallExtension'
     };
 
     // see https://github.com/microsoft/vscode/blob/2fc07b811f760549dab9be9d2bedd06c51dfcb9a/src/vs/workbench/contrib/extensions/common/extensions.ts#L246
@@ -378,7 +382,14 @@ export class PluginVscodeCommandsContribution implements CommandContribution {
         commands.registerCommand(VscodeCommands.INSTALL_EXTENSION_FROM_ID_OR_URI, {
             execute: async (vsixUriOrExtensionId: TheiaURI | UriComponents | string) => {
                 if (typeof vsixUriOrExtensionId === 'string') {
-                    await this.pluginServer.install(VSCodeExtensionUri.fromId(vsixUriOrExtensionId).toString());
+                    let extensionId = vsixUriOrExtensionId;
+                    let opts: PluginDeployOptions | undefined;
+                    const versionedId = PluginIdentifiers.idAndVersionFromVersionedId(vsixUriOrExtensionId);
+                    if (versionedId) {
+                        extensionId = versionedId.id;
+                        opts = { version: versionedId.version, ignoreOtherVersions: true };
+                    }
+                    await this.pluginServer.install(VSCodeExtensionUri.fromId(extensionId).toString(), undefined, opts);
                 } else {
                     await this.deployPlugin(vsixUriOrExtensionId);
                 }
@@ -393,6 +404,18 @@ export class PluginVscodeCommandsContribution implements CommandContribution {
                 } else {
                     await this.deployPlugin(uris);
                 }
+            }
+        });
+        commands.registerCommand(VscodeCommands.UNINSTALL_EXTENSION, {
+            execute: async (id: string) => {
+                if (!id) {
+                    throw new Error(nls.localizeByDefault('Extension id required.'));
+                }
+                const idAndVersion = PluginIdentifiers.idAndVersionFromVersionedId(id);
+                if (!idAndVersion) {
+                    throw new Error(`Invalid extension id: ${id}\nExpected format: <publisher>.<name>@<version>.`);
+                }
+                await this.pluginServer.uninstall(PluginIdentifiers.idAndVersionToVersionedId(idAndVersion));
             }
         });
         commands.registerCommand({ id: 'workbench.action.files.save', }, {


packages/plugin-ext/src/common/plugin-identifiers.ts
```diff
@@ -81,4 +81,31 @@ export namespace PluginIdentifiers {
         }
         return { id: probablyId.slice(0, endOfName) as UnversionedId, version: probablyId.slice(endOfName + 1) };
     }
+
+    const EXTENSION_IDENTIFIER_WITH_VERSION_REGEX = /^([^.]+\..+)@((prerelease)|(\d+\.\d+\.\d+(-.*)?))$/;
```
tsmaeder, line 85: This is unnecessarily restrictve, IMO: all we need to know here is what is before the '@' and what is after the '@'. Why introduce the added complexity? Also, regexes should always have a comment that describes what they are _supposed_ to detect, Because no-one understands Regexes at first sight. If the id does not match the expression, you end up with the exact same result as before.

packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
```diff
@@ -370,16 +375,36 @@ export class PluginVscodeCommandsContribution implements CommandContribution {
...
                     await this.pluginServer.deploy(`local-file:${uriPath}`);
                 }
             }
         });
+        commands.registerCommand(VscodeCommands.UNINSTALL_EXTENSION, {
+            execute: async (id: string) => {
+                if (!id) {
+                    throw new Error(nls.localizeByDefault('Extension id required.'));
+                }
+                if (!PluginIdentifiers.isVersionedId(id)) {
```
tsmaeder, line 400: I don't think this makes sense from a user's perspective: while technically, we can have multiple versions of the same plugin installed, that's not how it's presented to the user: for them, you either have a plugin installed, or not, independent of the version, so we should probably pass a naked id here.

packages/vsx-registry/src/browser/vsx-extension.tsx
```diff
@@ -324,8 +325,10 @@ export class VSXExtension implements VSXExtensionData, TreeElement {
             if (plugin) {
                 await this.progressService.withProgress(
                     nls.localizeByDefault('Uninstalling {0}...', this.id), 'extensions',
-                    () => this.pluginServer.uninstall(PluginIdentifiers.componentsToVersionedId(plugin.metadata.model))
-                );
+                    async () => {
```
tsmaeder, line 328: Why go through the command here? Why stringify/destringify the id?

packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
```diff
@@ -28,60 +30,59 @@ import {
 } from '@theia/core/lib/browser';
 import { ContextKeyService } from '@theia/core/lib/browser/context-key-service';
 import { ApplicationShellMouseTracker } from '@theia/core/lib/browser/shell/application-shell-mouse-tracker';
+import { SelectableTreeNode } from '@theia/core/lib/browser/tree/tree-selection';
```
tsmaeder, line 33: It's unfortunate that all the imports in this file are rewriten. It just adds noise to the PR.

tsmaeder:
@dhuebner still some open questions on this one.

packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
dhuebner: `PluginServer.uninstall()` expects `PluginIdentifiers.VersionedId` parameter which is typed as `${string}.${string}@${string}` that why it the code above makes sense to me. 

> that's not how it's presented to the user: for them, you either have a plugin installed, or not

The `UNINSTALL_EXTENSION` command is not presented to the user, but is used in `theia/packages/vsx-registry/src/browser/vsx-extension.tsx` where the versioned extension id is given. 
I would prefer to keep require versioned id here.

packages/plugin-ext/src/common/plugin-identifiers.ts
dhuebner: This is the same RegEx that is also used in vs-code, I thought it is a good idea to do the same.
I will add a comment to describe what is expected by this RegEx.

packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts
dhuebner: I also hate that! But somebody need to start applying  the proper imports sort :) 
Okay, I can revert the change.

packages/vsx-registry/src/browser/vsx-extension.tsx
dhuebner: > Why go through the command here? Why stringify/destringify the id?

The implementation is moved to a command to resolve #13796 , now calling the new command to reduce code duplicates.

dhuebner:
@tsmaeder 
Please find my answers above. 
I removed the organize imports changes and added comments to the RegEx. Also re-based on current master.

JonasHelming:
@tsmaeder Ping

tsmaeder (changes requested):
Hi Dennis, sorry for the delay, the last two releases have been somewhat crazy. Unfortunately, I still think this PR needs a bit of work.
Also, I don't really have a way to test the PR: when plugin API is involved and we don't have a known extension that exposes a plugin, we're making a little extension that exercises the API, in this case why not one that has two commands to install/uninstall with the user inputting a versioned id string?

packages/plugin-ext/src/common/plugin-identifiers.ts
dhuebner: I removed the RegEx now

packages/vsx-registry/src/browser/vsx-extension.tsx
dhuebner: Well this PR is here for too long now. I've replaced the command call with direct `pluginServer.uninstall` call.

dhuebner:
> Hi Dennis, sorry for the delay, the last two releases have been somewhat crazy. 

No problem! Same here, a lot of TODOs :) 

> Unfortunately, I still think this PR needs a bit of work. 

I've applied all the changes you suggested.

> Also, I don't really have a way to test the PR: when plugin API is involved and we don't have a known extension that exposes a plugin, we're making a little extension that exercises the API, in this case why not one that has two commands to install/uninstall with the user inputting a versioned id string?

The simplest way to test is to add labels to the commands: 
```ts
    export const INSTALL_FROM_VSIX: Command = {
        id: 'workbench.extensions.installExtension',
        label: nls.localizeByDefault('Install Extension')
    };

    export const UNINSTALL_EXTENSION: Command = {
        id: 'workbench.extensions.uninstallExtension',
        label: nls.localizeByDefault('Uninstall Extension')
    };
``` 

And then defaults for `id` parameters:
```ts 

commands.registerCommand(VscodeCommands.INSTALL_FROM_VSIX, {
            execute: async (vsixUriOrExtensionId: TheiaURI | UriComponents | string = 'langium.langium-vscode') => {
```

```ts
commands.registerCommand(VscodeCommands.UNINSTALL_EXTENSION, {
            execute: async (id: string = 'langium.langium-vscode@3.3.1') => {

```

dhuebner:
Btw. "Install Extension" user command would be a nice feature for Theia. Like shown [here](https://github.com/ChristianKohler/NpmIntellisense?tab=readme-ov-file#installation) in Installation section.

PR #14299, opened 2024-10-11
feat: support custom keys for custom Open AI models

### LINKED ISSUE #14288: API key passthrough missing for custom OpenAI-compatible models
### Bug Description:

Theia 1.54.0 claims to support custom OpenAI-compatible models including ["in the cloud"](https://github.com/eclipse-theia/theia/pull/14172#:~:text=running%20locally%20or-,in%20the%20Cloud,-as%20long%20as) but it really only supports non-api-key-protected models, as it does not seem to pass through a custom-model API key environment variable from the AI features settings.  So there seems to be no straightforward/non-hacky way to connect an API key requiring cloud-hosted OpenAI-compatible AI model (e.g. Perplexity Pro API).

And there is not even a straightforwardly-hacky way of doing it, without customizing theia code (e.g. by utilizing the official OpenAI API environment variable for non-official/OpenAI-compatible APIs), as the current implementation outright drops the api key for a hardcoded 'no-key' string when calling a custom model; here: https://github.com/eclipse-theia/theia/blob/c7fb4f525a3e4ededf600fef1ca71cd9fdaaeca0/packages/ai-openai/src/node/openai-language-model.ts#L188

### Steps to Reproduce:

1. Enable AI features in 1.54.0
2. Configure a custom OpenAI-compatible model in the settings.json &mdash; try to add an API key for this custom model in `"ai-features.openAiCustom.customOpenAiModels"` but...
3. Observe that the custom model will not work if it requires an API key

### Additional Information

- Operating System:
- Theia Version: 1.54

sdirix:
#### What it does

The configuration for custom OpenAI models now allows specifying a unique 'apiKey' for each model, or reusing the global OpenAI API key.

fixes #14288

#### How to test

- Add a custom Open AI model (does not need to be reachable for testing)
- Debug the "initializeOpenAi" method in the Open AI language models
   - should use `no-key` if no `apiKey` is specified
   - should use the specified key in `apiKey` if it exists
   - should use the global key (preference or environment variable) if `apiKey` is set to `true`

Note that during my testing not every nested change of the custom model preference resulted in a preference notification. To be safe you can reload the whole window which will trigger a "force-read" of the preferences. I think this issue is unrelated to the changes here.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-openai/README.md
+++ b/packages/ai-openai/README.md
@@ -14,7 +14,25 @@
 
 The `@theia/ai-openai` integrates OpenAI's models with Theia AI.
 The OpenAI API key and the models to use can be configured via preferences.
-Alternatively the OpenAI API key can also be handed in via an environment variable.
+Alternatively the OpenAI API key can also be handed in via the `OPENAI_API_KEY` variable.
+
+### Custom models
+
+The extension also supports OpenAI compatible models hosted on different end points.
+You can configure the end points via the `ai-features.openAiCustom.customOpenAiModels` preference:
+
+```ts
+{
+    model: string
+    url: string
+    id?: string
+    apiKey?: string | true
+}
+```
+
+- `model` and `url` are mandatory attributes, indicating the end point and model to use
+- `id` is an optional attribute which is used in the UI to refer to this configuration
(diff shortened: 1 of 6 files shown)


planger (approved):
Thanks, looks good to me! :+1: Works as advertised. I just tested it via debugging through the code, but not with an actual custom OpenAI compatible model though.

PR #14301, opened 2024-10-11
feat: add support for custom agents

eneufeld:
#### What it does

This adds support for custom agents which are basically a custom system prompt with additional metadata (id, name and description). This features allows to add very specific agents without coding. All features, like variable and functions are supported.

#### How to test

In the AI Configuration UI press the "Add Agent" button.
In the newly created yaml file add:
```yaml
- id: my_agent
  name: My Agent
  description: This is my custom agent which acts as a skater and uses skater terms. It can help with the current workspace.
  prompt: |
    You are a skater and use skater terms. Always use the skater perspective.

    You can use these functions to access the workspace:
    - ~{getWorkspaceFileList}
    - ~{getFileContent}. Never shorten the file paths when using this function
```

See that you can access the new agent via `@My Agent` in the chat or if you simply ask for support from a skayer this agent should also be used.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
+++ b/packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
@@ -87,6 +87,7 @@ export class AIAgentConfigurationWidget extends ReactWidget {
 
         this.aiSettingsService.onDidChange(() => this.update());
         this.aiConfigurationSelectionService.onDidAgentChange(() => this.update());
+        this.agentService.onDidChangeAgents(() => this.update());
         this.update();
     }
 
@@ -100,6 +101,7 @@ export class AIAgentConfigurationWidget extends ReactWidget {
                         </li>
                     )}
                 </ul>
+                <button style={{ marginLeft: 0 }} className='theia-button main' onClick={() => this.addCustomAgent()}>Add Custom Agent</button>
             </div>
             <div className='configuration-agent-panel preferences-editor-widget'>
                 {this.renderAgentDetails()}
@@ -203,6 +205,10 @@ export class AIAgentConfigurationWidget extends ReactWidget {
 
     protected showVariableConfigurationTab(): void {
         this.aiConfigurationSelectionService.selectConfigurationTab(AIVariableConfigurationWidget.ID);
+    }
+
+    protected addCustomAgent(): void {
+        this.promptCustomizationService.openCustomAgentYaml();
     }
 
     protected setActiveAgent(agent: Agent): void {
--- a/packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
+++ b/packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
@@ -101,7 +101,9 @@ export class AIAgentConfigurationWidget extends ReactWidget {
                         </li>
                     )}
                 </ul>
-                <button style={{ marginLeft: 0 }} className='theia-button main' onClick={() => this.addCustomAgent()}>Add Custom Agent</button>
+                <div className='configuration-agents-add'>
+                    <button style={{ marginLeft: 0 }} className='theia-button main' onClick={() => this.addCustomAgent()}>Add Custom Agent</button>
+                </div>
             </div>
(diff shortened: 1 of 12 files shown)


JonasHelming:
Great feature, I found a few issues:

- [ ] It used an arbitrary model as a default for me (in my case llama3.1)
- [ ] The system prompt is not used at all (neither the default nor a customized one, I cannot see it in the output)
- [ ] We should fill the yaml with a default structure on creation (My Agent, this is the agent description, this is the default system prompt) and then a comment underneath "Add more custom agents here"
- [ ] The add button looks a bit weird, maybe just have a little "+" and with more space or another location

eneufeld:
- the arbitrary model is the default behavior for all agents. I would not tackle that as part of this pr
- system prompt I have to retest, I was sure it worked, maybe the logging doesn't work
- I like the suggestion with the pre-filled yaml
- I will check the UI

JonasHelming (changes requested):
see comment above

packages/ai-core/src/browser/frontend-prompt-customization-service.ts
```diff
@@ -17,13 +17,19 @@
...
 import { FileChangesEvent } from '@theia/filesystem/lib/common/files';
 import { AICorePreferences, PREFERENCE_NAME_PROMPT_TEMPLATES } from './ai-core-preferences';
 import { AgentService } from '../common/agent-service';
 import { EnvVariablesServer } from '@theia/core/lib/common/env-variables';
+import { load } from 'js-yaml';
+
+const templateEntry = `-id: my_agent
+name: My Agent
+description: This is an example agent. Please adapt the properties to fit your needs.
+prompt: You are an example agent. Be nice and helpful to the user.`;
```
sdirix, line 32: This template is invalid and breaks the whole functionality. The resulting template looks like this:

```yaml
-id: my_agent
name: My Agent
description: This is an example agent. Please adapt the properties to fit your needs.
prompt: You are an example agent. Be nice and helpful to the user.
```

This is not an array but an object with 4 parameters, including `-id`.

---

As we are using `js-yaml` we should use its functionality to generate the yaml from a json object and not do any string manipulation on our own. See https://github.com/nodeca/js-yaml?tab=readme-ov-file#dump-object---options-

packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
```diff
@@ -100,6 +101,9 @@ export class AIAgentConfigurationWidget extends ReactWidget {
                         </li>
                     )}
                 </ul>
+                <div className='configuration-agents-add'>
+                    <button style={{ marginLeft: 0 }} className='theia-button main' onClick={() => this.addCustomAgent()}>Add Custom Agent</button>
+                </div>
```
sdirix, line 106: If there are already custom agents then this should be called  "Edit Custom Agents" as it does no longer add one

sdirix (changes requested):
It woks for me! However, there are some bugs and ux issues:

- The list of custom agents is not updated on template-folder-preference changes.
- It should be possible to set a default or recommended language model for the custom agent within the yaml. At the moment it just choses some random one which on my machine is a non-existent ollama one. The property should be part of the template and be set to chat gpt 4o.

sdirix:
To reproduce use the following yaml
```
- id: my_agent
  name: My Agent
  description: This is an example agent. Please adapt the properties to fit your needs.
  prompt: You are an example agent. Be nice and helpful to the user.
  defaultLLM: openai/gpt-4o
- id: my_agent2
  name: My Agent2
  description: This is an example agent. Please adapt the properties to fit your needs.
  prompt: You are an example agent. Be nice and helpful to the user.
  defaultLLM: openai/gpt-4o
- id: my_agent3
  name: My Agent3
  description: This is an example agent. Please adapt the properties to fit your needs.
  prompt: You are an example agent. Be nice and helpful to the user.
  defaultLLM: openai/gpt-4o
```

Then set the id of the second agent to the same as the first one and save

sdirix (approved):
Works for me. We should add a follow up ticket for just "editing" custom agents.

JonasHelming (approved):

PR #14320, opened 2024-10-15
Migrate from PhosphorJS to Lumino

sdirix:
#### What it does

Migrate from PhosphorJS to Lumino

[PhosphorJS](https://phosphorjs.github.io/) is no longer actively maintained [for a few years now](https://github.com/phosphorjs/phosphor). [Lumino](https://github.com/jupyterlab/lumino) is an active fork, maintained by the [JupyterLab](https://github.com/jupyterlab/jupyterlab) community.

While PhosphorJS proved reliable for a long time, the cracks are starting to show. For example [we need to patch PhosphorJS](https://github.com/eclipse-theia/theia/tree/c7fb4f525a3e4ededf600fef1ca71cd9fdaaeca0/dev-packages/cli/patches) to support secondary windows.

The JupyterLab community [is welcoming and is open to accept contributions](https://discourse.jupyter.org/t/adopting-lumino/23901).

#### Current state

The PR is split into separate logical commits to make the reviewing and maintenance work easier in case the PR takes longer to complete.

This PR is a draft PR as not all known issues are yet fixed.

Open issues are:

- [ ] Tab bars currently have double the width then they should have
- [ ] The "..." bar actions do not work
- [ ] The preferences tree does not react on selections

If anyone is interested in solving one or more of these issues, contact me or open a PR against the [`lumino-migration` branch in the EclipseSource fork](https://github.com/eclipsesource/theia/tree/lumino-migration).

Note that Lumino introduced a code change in which empty menus can't be opened. However Theia uses an approach in which menus are filled only shortly before they are shown. Therefore this new behavior is for now patched in this draft PR. However this means the patching is now ALWAYS required and not only when wanting to support secondary windows.

We might want to think about contributing a change to Lumino first before requiring this for all downstream projects.

In any case this is of course a huge breaking change.

#### How to test

Build and test the application

#### Review checklist

#### Reminder for reviewers


packages/core/src/browser/menu/browser-menu-plugin.ts
```diff
@@ -31,6 +31,11 @@ import { ApplicationShell } from '../shell';
 import { CorePreferences } from '../core-preferences';
 import { PreferenceService } from '../preferences/preference-service';
 
+/**
+ * Enable/Disable the new focus adaptations
+ */
+const ENABLE_FOCUS_ADAPTATIONS = true;
```
tsmaeder, line 37: Not sure why we need this: the new behavior either correct or it's not.

sdirix: We don't need it. I marked that commit as wip as I'm not yet satisfied with the changes and I wanted to mark with what I did in the code. However I wanted to push so that others, like you, could take a look if necessary.

Before merging I will definitely remove that constant again.

sdirix: I removed it again

tsmaeder:
> selecting a title menu (which opens the sub menu popup) and clicking again in the title menu brings the clicked on menu into a focus state. Clicking any other menu item in the title menu next, will still open that focused menu.

@sdirix I don't understand what this does: can you give a more concrete example what you click on in what sequence?

sdirix:
> > selecting a title menu (which opens the sub menu popup) and clicking again in the title menu brings the clicked on menu into a focus state. Clicking any other menu item in the title menu next, will still open that focused menu.
> 
> @sdirix I don't understand what this does: can you give a more concrete example what you click on in what sequence?

After the merge of the latest `master` state, the issue is now less severe: The focus state is still there, however it "just" prevents opening the menu and does no longer open the wrong one. I'll update the description in the overview text.

### Focus State Issue

The issue only occurs in the non-native menu bar, i.e. in the Theia Browser variant or when disabling the native menu in the Electron case:

FocusStateMenu

- Clicking on a menu and closing it again still focuses the menu item
- The next click does nothing besides losing the focus

This is how it works in the current Theia

TheiaMenuCorrect


### File Explorer Interaction

This is already broken in current `master`, however it's worse on the current lumino branch because of the issue above:

- Click a file explorer item
- Click on the menu -> Nothing happens
- Click on the menu -> Nothing happens
- Click on the menu -> Only now does it open

In the current `master` we need two clicks to open the menu. In the lumino branch we need three clicks.

tsmaeder:
It turns out the focus change causes the view part titles to change, which in turn refreshes the view container menu. This updates the visiblity toggle commands for that container. This in turn causes BrowserMainMenuFactory.showMenuBar() to invoked which replaces the whole main menu. This is happening while the menu is being shown, so the menu becomes empty. On the second try, the focused widget probably does not change anymore and the menu is shown O.K. When I remove the listener for the title change, the main menu bar works just as expected.

tsmaeder:
Note that we need update the main menu when the menu registry changes, menus might have been unregistered and the main menu bar has no "onBeforeShow" obviously, as it is always visible. Dynamic menus for the main menu bar seem more and more a strange concept to me. Also putting local menu items like visibility of the views in a view part container into a global menu registry seems odd to me. Maybe we need a completely different approach here.

sdirix:
@tsmaeder Thanks for the fixes! I tested it a bit and it's significantly better than before. There is one behavioral change though: The title bar menus can no longer be closed by clicking them again which seems a bit weird.

sdirix:
@msujew ping ;)

sdirix:
I merged in the latest `master` to solve the merge conflicts, no other changes done

msujew (commented):
I was able to find two regressions:

1. The menu CSS seems to have changed apparently. This affects all main menu and context menu rendering. Menu entries now have a focus outline. The outline is also visible in case the element is not `:hover`:
image
image
Current design for reference (with rounded corners):
image
2. Clicking on a top level menu element (such as `File`) brings it up correctly, but clicking the menu button again does nothing. I would expect it to close the menu again. This is the case in VS Code and current Theia.

Aside from that, this looks very good. I think after fixing those two issues, we should be good to go.

tsmaeder:
@msujew thanks for the review.

> Clicking on a top level menu element (such as File) brings it up correctly, but clicking the menu button again does nothing. I would expect it to close the menu again. This is the case in VS Code and current Theia.

I'm not sure this is an actual bug: we used to have a bug that would close the menu on focus change which was fixed. It might be that this is actually the correct behaviour for Lumino menus. We should check. Anyway: I don't think we should block the merge for that one: you can simple hit "escape" or click anywhere outside the menu to make it disappear.

sdirix:
> @msujew thanks for the review.
> 
> > Clicking on a top level menu element (such as File) brings it up correctly, but clicking the menu button again does nothing. I would expect it to close the menu again. This is the case in VS Code and current Theia.
> 
> I'm not sure this is an actual bug: we used to have a bug that would close the menu on focus change which was fixed. It might be that this is actually the correct behaviour for Lumino menus. We should check. Anyway: I don't think we should block the merge for that one: you can simple hit "escape" or click anywhere outside the menu to make it disappear.

Personally I would also prefer the menus to close again, however this is not a regression of this PR. The current Theia `master` already has this behavior in the Browser variant as well as when using title bar style "custom" instead of native in the Electron variant, including the Theia IDE.

tsmaeder:
> The current Theia master already has this behavior in the Browser variant as well as when using title bar style "custom"

Now I'm confused: in Theia IDE, the menu closes on second click in the **native** menus and stays open in the **custom** menus

sdirix:
> > The current Theia master already has this behavior in the Browser variant as well as when using title bar style "custom"
> 
> Now I'm confused: in Theia IDE, the menu closes on second click in the **native** menus and stays open in the **custom** menus

Yes, that's because Electron native menus behave that way, and our changes on master regarding (PhosporJS) menus behave the other way. We should align them again in a follow up at some point.

sdirix:
I pushed two commits addressing your review issues. As discussed above, the title menu's (non-native) behavior already changed on `master` and is therefore not a regression by this PR.

sdirix:
I overlooked that we already had some CSS customization in menu.css. I moved the focus customization there as that is a better fit than index.css. No functional or CSS changes.

msujew:
> The current Theia master already has this behavior in the Browser variant as well as when using title bar style "custom" instead of native in the Electron variant, including the Theia IDE.

I can confirm. I was testing using the `1.58.100` build of the Theia IDE, but the `1.59.100` build actually also has this issue.

package-lock.json
```diff
@@ -368,6 +368,10 @@
         "node": ">= 12"
       }
     },
+    "dev-packages/private-test-setup": {
+      "name": "@theia/test-setup",
+      "version": "1.56.0"
```
msujew, line 373: This is outdated. Please run `npm i` :)

msujew (approved):
Alright, thank you for the swift reply. I could find no more regressions from `master`, and the code looks good to me 👍

package-lock.json
sdirix: good catch!

sdirix:
@tsmaeder Can you do or trigger the prerelease?

PR #14349, opened 2024-10-23
Show error message when uploading fails

dhuebner:
#### What it does

Shows an error dialog when file upload fails, e.g. if no space left on target device. Before this commit upload progress dialog was just silently closed with no error reported to the user. This cause a bad user experience as the uploaded file is in most cases just corrupt.

#### How to test

Test using the browser version.

To test the issue one needs to simulate a file upload failure. The easiest way I found, is creating a small partition on an USB device and add the mounted device to the Theia workspace.  Now try to upload a file that is larger than the available space on the device to the mounted folder . You should see an error message popping up with a corresponding note.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/browser/file-upload-service.ts
+++ b/packages/filesystem/src/browser/file-upload-service.ts
@@ -254,6 +254,7 @@ export class FileUploadService {
         } catch (error) {
             uploadSemaphore.cancel();
             if (!isCancelled(error)) {
+                this.messageService.error(nls.localize('theia/filesystem/uploadFailed', 'An error occurred while uploading a file. {0}', error.message));
                 throw error;
             }
         }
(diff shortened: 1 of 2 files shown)


packages/filesystem/src/browser/file-upload-service.ts
```diff
@@ -254,6 +254,7 @@ export class FileUploadService {
         } catch (error) {
             uploadSemaphore.cancel();
             if (!isCancelled(error)) {
+                this.messageService.error(`An error occurred while uploading a file. ${error.toString()}`);
```
jonah-iden, line 257: maybe add localization for this message just with `nls.localize('key', 'message {0}', error)`

jonah-iden (approved):
looks good to me. Just maybe add the localization

packages/filesystem/src/browser/file-upload-service.ts
dhuebner: Good idea

PR #14354, opened 2024-10-25
fix: properly initialize custom agents

sdirix:
#### What it does

Prompt templates of newly added custom agents were not stored within the PromptService. As a consequence their prompt was ignored when invoked in Chat.

This issue is now fixed by restructuring the prompt storing mechanism. Prompts are now implicitly stored and removed based on all currently registered agents.

##### Use Case - Initial Prompt
- Add a new custom agent
- Adjust its prompt within the yaml file
- Verify that the prompt is actually used by invoking the agent in the chat

##### Use Case - Modified Prompt
- Open the custom agents yaml file (manually or by adding a new agent)
- Modify the prompt of an agent (whose prompt is not overriden)
- Verify that the prompt is actually used by invoking the agent in the chat

##### Use Case - Customized Prompt
- Edit the prompt of a custom agent via the "Edit" button, i.e. not via the yaml file
- Verify that the new prompt is actually used by invoking the agent in the chat
- Reset the prompt
- Verify that the initial prompt of the yaml file is used by invoking the agent in the chat

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/browser/ai-configuration/template-settings-renderer.tsx
+++ b/packages/ai-core/src/browser/ai-configuration/template-settings-renderer.tsx
@@ -25,7 +25,7 @@ export interface TemplateSettingProps {
 
 export const TemplateRenderer: React.FC<TemplateSettingProps> = ({ agentId, template, promptCustomizationService }) => {
     const openTemplate = React.useCallback(async () => {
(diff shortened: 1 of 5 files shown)


planger (approved):
Thanks, looks good to me and works as advertised! :+1:

Issue #14203, opened 2024-09-19
The TabsMainImpl class leaks references until group changes


dfriederich:
The TabsMainImpl keeps a list of dispose operations which are 'only' run when a tab group changes.
Say if a tab contains 2 elements, and one of them is closed, then that element still remains references, and hence not garbage collected. Only when both elements are removed and a new one is added, then the group is considered changed and the elements are disposed.

https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/main/browser/tabs/tabs-main.ts#L127
is the code recomputing the model, and in this process the disposables collected in disposableTabBarListeners are disposed.
But the code should dispose no longer needed elements as soon as their tab is removed.

tsmaeder:
@martin-fleck-at so is this one a duplicate that can be closed?

dfriederich:
The #14186 fix was fixing that every control was leaked. But this is about that the existing code is only releasing its references in specific circumstances, for example when the first tab gets created, but not when a tab gets closed.

PR #14359, opened 2024-10-28
Properly dispose tab/title-based resources on tab close

martin-fleck-at:
#### What it does

The complete tabs model is only disposed and re-created whenever we we need to initialize it (no groups are available), we detect a broken state (we cannot find a tab that should be present) or when the overall tab groups have changed. This re-creation disposes the listeners we have on the current tab bars. However, in this group of listeners we also have some listeners that are only related to a particular tab/widget/title and they do not get properly disposed when a single tab is closed. The goal of this PR is to fix this.

Fixes https://github.com/eclipse-theia/theia/issues/14203

#### How to test

- Use the Chrome Developer Tools to make Memory snapshots and observe the behavior of closing and opening tabs over time.
- Some resources are cleaned but you can observe that after closing all tabs after having several open, we have memory X and after opening a single tab again the memory is less than X because we triggered the re-creation of the tabs model.
- In the new version the memory size should naturally follow the number of open tabs. 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/tabs/tabs-main.ts
+++ b/packages/plugin-ext/src/main/browser/tabs/tabs-main.ts
@@ -43,6 +43,7 @@ export class TabsMainImpl implements TabsMain, Disposable {
     private applicationShell: ApplicationShell;
 
     private disposableTabBarListeners: DisposableCollection = new DisposableCollection();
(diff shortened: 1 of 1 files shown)


martin-fleck-at:
@dfriederich FYI

msujew (approved):
Thanks, looks good to me 👍

PR #14363, opened 2024-10-28
chore(ai): avoid line wrap in code response parts

planger:
#### What it does

Line wrapping lead to cut off code. Instead, we now never wrap and show a horizontal bar. Moreover, this change ensures that mouse wheel events aren't caught be the code parts but always continue scrolling the chat view.

#### How to test

Get an LLM to print some code parts, e.g. the Universal chat agent with the message *Give me an example for a Typescript interface*. Once the code part is visible resize the chat widget in all directions and observe that it never cuts off the code. Also test whether scrolling of the chat widget is smooth, even if you hover above a code part.

#### Follow-ups

None

#### Review checklist

#### Reminder for reviewers


JonasHelming (approved):
Works well for me!

PR #14377, opened 2024-10-30
Keep closable state through pin/unpin

### LINKED ISSUE #14370: Unpinning an unclosable widget makes it closable
### Bug Description:

By pinning and unpinning a tab containing an unclosable widget, it becomes closable. The reason lies in the unpin function making the widget closable no matter what in https://github.com/eclipse-theia/theia/blob/f29c3d04241b32791462c7027938abeb887f8140/packages/core/src/browser/widgets/widget.ts#L373

tsmaeder:
#### What it does
Fixes #14370

#### How to test
Use the "Sample Unclosable View" and the File explorer to make sure views that are pinned/unpinned return to the same "closable" state after pinning/unpinning.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/widgets/widget.ts
+++ b/packages/core/src/browser/widgets/widget.ts
@@ -396,7 +396,6 @@ export function unpin(title: Title<Widget>): void {
     title.className = title.className.replace(PINNED_CLASS, '').trim();
 }
 
-
 export function isLocked(title: Title<Widget>): boolean {
     return title.className.includes(LOCKED_CLASS);
 }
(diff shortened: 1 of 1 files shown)


packages/core/src/browser/widgets/widget.ts
```diff
@@ -364,23 +365,37 @@ function waitForVisible(widget: Widget, visible: boolean, attached?: boolean): P
     });
 }
 
+const pinnedTitles = new Map<Title<Widget>, [boolean, Slot<Widget, void>]>();
```
sgraband, line 368:
```suggestion
const pinnedTitles = new Map<Title<Widget>, PinnedProperties>();
```

Can we maybe create a type for the values of the map? This way the values in the array are better described and the code in the unpin method becomes more readable imho. I am not sure about the name, but something along the lines of `PinnedProperties` could be a good fit.
WDYT?

tsmaeder: There already is a type for the values in the map, it's a tuple of boolean and a Slot. Explicitly naming that type might make sense if the type was used elsewhere for documentatoin, but it's all private and happening in a couple of lines, so I don't think it adds any value.

PR #14401, opened 2024-11-05
fix upper most context menu item sometimes not clickable in electron

jonah-iden:
#### What it does

In electron when the context menu overlaps the center of the top window toolbar as seen here
grafik
The overlapping element is not clickable. 
This PR fixes this issue similar to how vscode seems to fix it, by not allowing the context menu to overlap the window toolbar.

#### How to test

Open a file, open the context menu somewhere in the middle so that it would normally start at the top of the application. 
The upper most entry should still be clickable and it should start below the window toolbar.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-browser/menu/electron-context-menu-renderer.ts
+++ b/packages/core/src/electron-browser/menu/electron-context-menu-renderer.ts
@@ -23,7 +23,7 @@ import {
 import { ElectronMainMenuFactory } from './electron-main-menu-factory';
 import { ContextMenuContext } from '../../browser/menu/context-menu-context';
 import { MenuPath, MenuContribution, MenuModelRegistry } from '../../common';
-import { BrowserContextMenuRenderer } from '../../browser/menu/browser-context-menu-renderer';
+import { BrowserContextMenuAccess, BrowserContextMenuRenderer } from '../../browser/menu/browser-context-menu-renderer';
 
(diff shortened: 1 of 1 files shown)


packages/core/src/electron-browser/menu/electron-context-menu-renderer.ts
```diff
@@ -115,7 +115,14 @@ export class ElectronContextMenuRenderer extends BrowserContextMenuRenderer {
             this.context.resetAltPressed();
             return new ElectronContextMenuAccess(menuHandle);
         } else {
-            return super.doRender(options);
+            const menuAccess = super.doRender(options);
+            const node = (menuAccess as BrowserContextMenuAccess).menu.node;
+            // ensure the context menu is not displayed outside of the main area
+            if (node.style.top && parseInt(node.style.top.substring(0, node.style.top.length - 2)) < 32) {
```
msujew, line 121: Suggestion: It might make sense to dynamically retrieve the height of the top panel by doing something like this:

```ts
const height = document.getElementById('theia-top-panel')?.clientHeight ?? 0;
```

This is a bit more resilient against possible changes to the CSS in the future (or by adopters).

jonah-iden: good Idea, thanks!

msujew (approved):
Thanks, LGTM now 👍

PR #14412, opened 2024-11-06
Add support for Hugging Face

### LINKED ISSUE #14411: [Theia AI] Support for Hugging Face

JonasHelming:
fixed #14411

#### What it does

Adds support for using Hugging Face as an inference provider. This was possible before by using the OpenAI API, but some models require custom parameters

#### How to test

Create a HF Account and an API Key. Copy any model in the settings and select it to be used in the AI configuration view.
For simplicity, select a model that supports server less so that inference is for free.
For StarCoder code completion you can use:

```
The language is {{language}}.
<fim_prefix>{{textUntilCurrentPosition}}<fim_suffix>{{textAfterCurrentPosition}}<fim_middle>
```

#### Follow-ups

- We should think about prompt management for agents so that you can define different prompt depending on the connected model.
- We might want to harmonize the LLM Providers we have and extract common behavior soon

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-hugging-face/src/node/huggingface-language-model.ts
+++ b/packages/ai-hugging-face/src/node/huggingface-language-model.ts
@@ -1,0 +1,126 @@
+// *****************************************************************************
+// Copyright (C) 2024 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import {
+    LanguageModel,
+    LanguageModelRequest,
+    LanguageModelRequestMessage,
+    LanguageModelResponse,
+    LanguageModelStreamResponsePart,
+    LanguageModelTextResponse,
+    MessageActor
+} from '@theia/ai-core';
+import { CancellationToken } from '@theia/core';
+import { HfInference } from '@huggingface/inference';
+
+export const HuggingFaceModelIdentifier = Symbol('HuggingFaceModelIdentifier');
+
+function toHuggingFacePrompt(messages: LanguageModelRequestMessage[]): string {
+    if (messages.length === 1) {
+        return messages[0].query;
+    }
+    return messages.map(message => `${toRoleLabel(message.actor)}: ${message.query}`).join('\n');
+}
+
+function toRoleLabel(actor: MessageActor): string {
+    switch (actor) {
+        case 'user':
+            return 'User';
+        case 'ai':
+            return 'Assistant';
+        case 'system':
+            return 'System';
+        default:
+            return '';
+    }
+}
+
+export class HuggingFaceModel implements LanguageModel {
+    private hfInference: HfInference;
+
+    /**
+     * @param id the unique id for this language model. It will be used to identify the model in the UI.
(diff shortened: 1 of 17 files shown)


packages/ai-hugging-face/src/node/huggingface-language-model.ts
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * @param id the unique id for this language model. It will be used to identify the model in the UI.
+     * @param model the model id as it is used by the Hugging Face API
+     * @param apiKey function to retrieve the API key for Hugging Face
+     */
+    constructor(public readonly id: string, public model: string, public apiKey: () => string | undefined) {
+        const token = this.apiKey();
+        if (!token) {
+            throw new Error('Please provide a Hugging Face API token.');
+        }
```
sdirix, line 63: We should not throw errors for expected code paths. This is even the default code path as the api key will be `undefined` by default. Therefore this will show up as an error for all users in their logs.

If we don't want to be able to configure hugging face models without keys, then the `huggingface-language-model-manager` should not even create them.

Note that if the user deletes the `apiKey` in their preferences, you will still use it as the `hfInference` is only created once here. In fact the whole dynamic `apiKey` provider is not utilized in this implementation as it's only called once.

I would like to suggest to either make the `apiKey` static here and making sure that the models are removed/recreated if the `apiKey` changes or refactor them to being able to handle non existent api keys at request time, like in the open ai language model implementation.

packages/ai-hugging-face/src/node/huggingface-language-models-manager-impl.ts
```diff
@@ -0,0 +1,59 @@
...
+            }
+        }
+    }
+
+    removeLanguageModels(...modelIds: string[]): void {
+        this.languageModelRegistry.removeLanguageModels(modelIds);
+    }
+
+    setApiKey(apiKey: string | undefined): void {
+        this._apiKey = apiKey || undefined;
```
sdirix, line 57: Currently language models are not created when the `apiKey` is not set. When the user then sets the `apiKey` the language models are not created and they can't select them in the configuration view.

This is different to the open ai models which are created in any case, no matter whether the key exist. The behavior here should either be aligned to the open ai models behavior, or setting the `apiKey` afterwards also needs to create the models.

sdirix (changes requested):
In principle it seems to work. I was able to communicate with starcoder hosted at huggingface.

- There is an issue with the language model management which I described in detail in the comments
- I used the suggested
  ```
  The language is {{language}}.
  <fim_prefix>{{textUntilCurrentPosition}}<fim_suffix>{{textAfterCurrentPosition}}<fim_middle>
   ```
   as the code-completion prompt but only got very bad completions back :shrug:

JonasHelming:
@sdirix I changed the behavior so that API Key changes are respected (used the APIkeyProvider). It is now similar to the openAI provider. If no API key is set, the models are visible, but an error is thrown. This is catched by the chat and shown to the user (just like with the open AI provider)

packages/ai-hugging-face/src/node/huggingface-language-model.ts
```diff
@@ -0,0 +1,129 @@
...
+                return_full_text: false,
+                do_sample: true,
+                stop: ['<|endoftext|>']
+            }
+        });
+
+        const asyncIterator = {
+            async *[Symbol.asyncIterator](): AsyncIterator<LanguageModelStreamResponsePart> {
+                for await (const chunk of stream) {
+                    const content = chunk.token.text.replace(/<\|endoftext\|>/g, '');
```
sdirix, line 106: Will this work for most/all hugging face models or only for starcoder? If this is starcoder specific then we should check for that model before setting the parameters / replacing the content.

sdirix (approved):
Works for me with the suggested starcoder model. I have one more question regarding the code.

packages/ai-hugging-face/src/node/huggingface-language-model.ts
JonasHelming: I found that almost all HuggingFace Model return stop charakters. And that it is most robust to explcitly specify it and then replace it. Good test case is meta-llama/Llama-3.2-3B-Instruct which is always "warm"

PR #14414, opened 2024-11-07
Support bundling Theia with esbuild

msujew:
#### What it does

This PR adds support for bundling Theia with esbuild. Bundling Theia with webpack is still possible, however we might want to deprecate this in the future due to the maintenance effort if we go through with this PR.

This change essentially just adds three esbuild build contexts to the webpack-generator (now renamed bundler-generator) that are invoked via the command line.

I needed to do a few changes to our CSS, because esbuild didn't want to compile them. This mostly includes moving `~` based imports of other CSS files into code and moving the `@import` statements to the top of the file. Additionally, esbuild doesn't want to bundle the CSS into the JS, but creates a separate `bundle.css` file. This is now consumed via the `index.html`.

The new bundler is **way faster** than webpack. I've observed a tenfold improvement, reducing the build time from 20 seconds to only 2 seconds.

#### How to test

Test everything. I.e. assert that every feature works in every configuration (browser, browser-only, electron). This specifically includes:

- plugin support
- webview support
- notebook support
- search-in-workspace support
- terminal support
- build/watch modes

#### Follow-ups

As of now, this change is missing support for:
- compressed artifacts
- code splitting/chunking - ESBuild does not support that

#### Review checklist

#### Reminder for reviewers


sdirix (commented):
Nice initiative! I'm on Ubuntu 22.04. Building is much faster :+1:

I only did some quick test by starting the Browser and Electron apps. This is what I saw:

- The terminal does not start (browser and Electron app)
```
2024-11-08T15:54:15.608Z terminal ERROR TypeError: pty.fork is not a function
    at new UnixTerminal2 (/theia/examples/browser/lib/backend/main.js:270250:26)
    at spawn (/theia/examples/browser/lib/backend/main.js:270444:14)
    at ShellProcess.createPseudoTerminal (/theia/examples/browser/lib/backend/main.js:270934:47)
    at startTerminal (/theia/examples/browser/lib/backend/main.js:270900:25)
    at new TerminalProcess (/theia/examples/browser/lib/backend/main.js:270921:43)
    at new ShellProcess (/theia/examples/browser/lib/backend/main.js:271503:9)
    at createInstanceWithInjections (/theia/examples/browser/lib/backend/main.js:25307:18)
    at _createInstance (/theia/examples/browser/lib/backend/main.js:25298:16)
    at resolveInstance (/theia/examples/browser/lib/backend/main.js:25390:16)
    at _getResolvedFromBinding (/theia/examples/browser/lib/backend/main.js:25698:20)
2024-11-08T15:54:15.608Z root ERROR Error: pty process did not start correctly
    at ShellProcess.checkTerminal (/theia/examples/browser/lib/backend/main.js:271012:17)
    at get pid [as pid] (/theia/examples/browser/lib/backend/main.js:270965:14)
    at /theia/examples/browser/lib/backend/main.js:271732:51
    at /theia/examples/browser/lib/backend/main.js:31322:69
    at CallbackList.invoke (/theia/examples/browser/lib/backend/main.js:31328:23)
    at _Emitter.fire (/theia/examples/browser/lib/backend/main.js:31438:34)
    at ShellProcess.emitOnError (/theia/examples/browser/lib/backend/main.js:269086:27)
    at process.processTicksAndRejections (node:internal/process/task_queues:81:21)
```

- Starting Electron I get errors like these
```
TypeError: this._keymapping.onDidChangeKeyboardLayout is not a function
    at NativeBinding.onDidChangeKeyboardLayout (/theia/examples/electron/lib/backend/electron-main.js:71800:26)
    
  2024-11-08T16:10:17.975Z root ERROR TypeError: this._keymapping.getKeyMap is not a function
    at NativeBinding.getKeyMap (/theia/examples/electron/lib/backend/main.js:129581:33)

2024-11-08T16:10:17.975Z root ERROR TypeError: this._keymapping.getCurrentKeyboardLayout is not a function
```

msujew:
@sdirix Thank you for taking a look. I actually fell into a esbuild trap, that I didn't notice before. Using `'.node': 'file'` doesn't actually load the file - it just replaces the import with a file path (which is pretty useless, to be honest). Anyway, I performed some changes to the esbuild plugin and the `.node` files should be loaded correctly now.

msujew:
For some reason, the call to `rg` seems to fail during the runtime (it finds the file, but the calls result in error code `1`). I'll not entirely sure why. I'll take a look at at that during the week.

msujew:
@nimo23 AFAIK Vite stopped using esbuild as part of their pipeline and completely switched to Rollup, see also their [documentation](https://vite.dev/guide/why#why-not-bundle-with-esbuild). They say themselves that Vite is slower than esbuild for that reason.

I know that Vite has a few benefits (and is easier to deal with in general), but Theia adopters don't need to configure too much anyway, as the cli already does most of the heavy lifting. Do you have a specific use case in mind that is difficult to implement using esbuild?

msujew:
@nimo23 Unfortunately, it's not that easy. Webpack is already capable of doing hot reloading, but we're not making use of that either - mainly because we're serving the frontend from our own backend and not using a dev server for that. Whether we use Vite, webpack or esbuild doesn't matter for that - hot reloading is not really supported in Theia right now.

msujew:
I'm converting this into a draft. While I could finish this PR, I would actually much rather wait on the new [rolldown](https://rolldown.rs/) bundler that likely makes more sense to use.

msujew:
@sdirix Sorry for the long break/pause on this. I believe this should be ready for a proper review now.

@ndoschek I renamed one of our packages in this PR (it's now simply `@theia/bundle-plugin`) - mainly because `native-webpack-plugin` neither fit our need (it did more than just handling the native dependencies, like monaco for example) and it doesn't properly reflect that an ESBuild plugin is included in there as well. I hope that's alright? The `@theia/native-webpack-plugin` on npm should be deprecated once this is released.

pisv:
@msujew Thanks a lot for this work, building is much faster now!

As a quick feedback, I've had a successful clean build of the PR branch on macOS locally, but I'm seeing a major loss of functionality in Diff Editor, 3-way Merge Editor, and Dirty Diff Peek View, which look completely broken in both Electron and Browser. I have not checked it on other operating systems.

msujew:
@pisv Thanks for the hint and testing my changes! The normal monaco editor worked as expected, but I haven't actually looked at the other editors. I'll have a look at that. The playwright tests are also pointing in the direction that some things aren't working as expected :)

msujew:
@pisv I cannot actually reproduce the issues you're experiencing. All editors look fine to me:

grafik

Have you tried deleting the `lib` directory of the `browser/electron` apps before building? Perhaps something is messing with the output. I was able to fix the other tests - I'm not sure how ESBuild was able to break them, but they should work again now.

msujew:
As for the license check, I've filed [#27012](https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/27012) and [#27013](https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/27013). Hopefully, this should be resolved soon.

Edit: It is resolved now.

pisv:
> @pisv I cannot actually reproduce the issues you're experiencing. All editors look fine to me

This time I've made a fresh clone of the Theia repository to build the PR branch. (Previously, I just did `npm run clean` and `git clean -xfd` before the usual build steps.)

Now everything that was previously broken works as expected. I'm really sorry for the false alarm! It looks like the clean build was not really clean :/

msujew:
All good, no worries. Thanks for taking another look! Glad to hear things are working well with the new build system :)

ndoschek (commented):
Thank you for this great improvement Mark!
I tested this a bit locally and the build works great, I am really impressed by the speed improvement! 🚀

Renaming the @theia/webpack-plugin package is fine, thanks! I cam take care of the deprecation on npm then later.

I have one question: Since preferESBuild() falls back to webpack when an existing webpack.config.js is present: should we keep webpack in theia-ide for now (e.g. for the electron product see here: https://github.com/eclipse-theia/theia-ide/blob/master/applications/electron/webpack.config.js), or would it be a good opportunity to switch early and get broader testing coverage for the esbuild path?

msujew:
>  or would it be a good opportunity to switch early and get broader testing coverage for the esbuild path?

I'd say we should switch early - from what I've gathered during the dev meetings, we will soon deprecate the webpack support, which means that we need to switch anyway. Getting more projects onto esbuild earlier, should give us more confidence in deprecating the old stuff.

ndoschek:
Thanks! I've created an issue on the theia-ide repo to track the switch (see eclipse-theia/theia-ide#694)

Can you please also add this to the changelog and migration guide so downstream adopters know about the esbuild option and the upcoming webpack deprecation? TIA!

msujew:
I've added it to the migration guide - as for the changelog: There are no breaking changes in here (I think).

dev-packages/bundle-plugin/src/esbuild-plugin.ts
```diff
@@ -0,0 +1,335 @@
+// *****************************************************************************
+// Copyright (C) 2024 TypeFox and others.
```
sdirix: Should we use 2026?

examples/electron/esbuild.mjs
```diff
@@ -0,0 +1,34 @@
...
+} else {
+    try {
+        await browserContext.rebuild();
+        await browserContext.dispose();
+        await nodeContext.rebuild();
+        await nodeContext.dispose();
+        await electronContext.rebuild();
+        await electronContext.dispose();
+    } catch (err) {
+        process.exit(1);
```
sdirix, line 32: Should we log the error?

examples/browser/esbuild.mjs
```diff
@@ -0,0 +1,29 @@
...
+        nodeContext.watch(),
+    ]);
+} else {
+    try {
+        await browserContext.rebuild();
+        await browserContext.dispose();
+        await nodeContext.rebuild();
+        await nodeContext.dispose();
+    } catch (err) {
+        process.exit(1);
```
sdirix, line 27: Should we log the errror?

examples/browser-only/esbuild.mjs
```diff
@@ -0,0 +1,19 @@
...
+const browserContext = await esbuild.context(browserOptions);
+
+if (watch) {
+    await browserContext.watch();
+} else {
+    try {
+        await browserContext.rebuild();
+        await browserContext.dispose();
+    } catch (err) {
+        process.exit(1);
```
sdirix, line 17: Should we log the error?

ndoschek:
Hey @msujew, thanks again for working on this!
Unfortunately the recent release introduced some conflicts on your branch - would you mind doing another rebase when you have a minute? TIA

msujew:
@ndoschek Thanks for the hint. I've updated the PR and also added an entry to the breaking changes!

package-lock.json
```diff
@@ -13782,6 +14682,109 @@
       "integrity": "sha512-HJDGx5daxeIvxdBxvG2cb9g4tEvwIk3i8+nhX0yGrYmZUzbkdg8QbDevheDB8gd0//uPj4c1EQua8Q+MViT0/w==",
       "license": "MIT"
     },
+    "node_modules/esbuild": {
```
ndoschek, line 14203: It looks like we have the esbuild dependency duplicated now.
esbuild@^0.28.0 conflicts with the peer range of esbuild-plugins-node-modules-polyfill.
npm installs two copies of esbuild: 0.24.2 hoisted to root (to satisfy the polyfill peer) and 0.28.0 nested under application-manager.
Can we either pin to ^0.27.0 so both declarations agree, or move to a polyfill plugin variant that supports 0.28.x?

doc/Migration.md
```diff
@@ -9,6 +9,12 @@ Please see the latest version (`master`) for the most up-to-date information. Pl
 
 ### General
 
+_ESBuild_:
+
+In addition to `webpack`, Theia is now also supporting [`ESBuild`](https://esbuild.github.io/) for bundling the application (frontend+backend). We will soon deprecate and then remove the `webpack` bundling option. Adopters can already use the ESBuild based bundler simply by deleting their `webpack.config.js`, which will automatically generate an `esbuild.mjs` file upon the next build.
```
ndoschek, line 14: nit: since the guidance is "delete webpack.config.js and an esbuild.mjs is generated", it would be worth adding a one-line pointer on where custom esbuild.mjs hooks fit in (e.g. which plugins from @theia/bundle-plugin are intended to be kept) so adopters with custom webpack tweaks know what the ESBuild equivalents look like.

ndoschek (commented):
Thank you for the update, Mark!
LGTM overall, I just noticed a duplication of the esbuild dependency in the package-lock (please see inline).
Then we probably also need to run the ip check again as well.
Could you have another look when you have a minute, TIA!

doc/Migration.md
msujew: I'm not sure I can follow this. The new build script boils down to:

```ts
import { browserOptions, watch } from './gen-esbuild.browser.mjs';
import { nodeOptions } from './gen-esbuild.node.mjs';
import esbuild from 'esbuild';

const browserContext = await esbuild.context(browserOptions);
const nodeContext = await esbuild.context(nodeOptions);
```

Everything required by Theia itself is already embededded inside of the `options` objects. And as for any customizations - I don't think there's any good recommendation for adopters aside from linking the esbuild documentation.

packages/core/src/browser/style/index.css
```diff
@@ -182,15 +207,18 @@ blockquote {
 
 .theia-preload::after {
     animation: 1s theia-preload-rotate infinite;
-  color: #777; /* color works on both light and dark themes */
-  content: "\eb19"; /* codicon-load */
+    color: #777;
+    /* color works on both light and dark themes */
+    content: "\eb19";
+    /* codicon-load */
```
sdirix: Nit: Seems weird to have the comment _after_ the property

packages/core/src/browser/style/index.css
```diff
@@ -230,12 +258,18 @@ blockquote {
...
+    -webkit-user-select: none;
+    /* Safari */
+    -khtml-user-select: none;
+    /* Konqueror HTML */
+    -moz-user-select: none;
+    /* Firefox */
+    -ms-user-select: none;
+    /* Internet Explorer/Edge */
+    user-select: none;
+    /* Non-prefixed version, currently
```
sdirix: Nit: Same here

packages/core/src/browser/style/index.css
```diff
@@ -160,6 +184,7 @@ blockquote {
     right: 0;
     bottom: 0;
     background: var(--theia-editor-background);
+    height: 100%;
```
sdirix, line 191: Where is this CSS change coming from?

packages/core/src/browser/style/index.css
```diff
@@ -34,17 +62,13 @@
...
-  --theia-ui-font-size2: calc(
-    var(--theia-ui-font-size1) * var(--theia-ui-font-scale-factor)
-  );
-  --theia-ui-font-size3: calc(
-    var(--theia-ui-font-size2) * var(--theia-ui-font-scale-factor)
-  );
-  --theia-ui-icon-font-size: 14px; /* Ensures px perfect FontAwesome icons */
+    --theia-ui-font-size0: calc(var(--theia-ui-font-size1) / var(--theia-ui-font-scale-factor));
+    --theia-ui-font-size1: 13px;
+    /* Base font size */
```
sdirix: Nit: Comment _after_ property

packages/core/src/browser/style/select-component.css
```diff
@@ -15,8 +15,6 @@
  ********************************************************************************/
 
 .theia-select-component-container {
-  /* required to set z-index */
-  position: fixed;
```
sdirix, line 19: Why is this removed?

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
```diff
@@ -27,9 +27,8 @@ export class PluginWorker {
 
     constructor() {
         this.worker = new Worker(new URL('./worker/worker-main',
-            // @ts-expect-error (TS1343)
             // We compile to CommonJS but `import.meta` is still available in the browser
-            import.meta.url));
+            ''));
```
sdirix: Is this a breaking change?

packages/plugin-ext/src/main/browser/webview/pre/fake.html
```diff
@@ -1,6 +1,8 @@
 <!DOCTYPE html>
 <html lang="en">
 
+Test
+
```
sdirix: Should be removed?

dev-packages/bundle-plugin/src/esbuild-plugin.ts
```diff
@@ -0,0 +1,335 @@
...
+    const cases = [];
+
+    for (const [module, node] of Object.entries(bindings)) {
+        cases.push(`${' '.repeat(8)}case '${module}': return require('${node}');`);
+    }
+
+    return `
+module.exports = function (jsModule) {
+    switch (jsModule) {
+${cases.join('/')}
```
sdirix: `cases.join('/')` produces invalid JavaScript when there is more than one native binding. The webpack version uses `os.EOL` [for this](https://github.com/eclipse-theia/theia/blob/a8b7d4add9a8254cc7877e82ddcfe33c430b5e95/dev-packages/native-webpack-plugin/src/native-webpack-plugin.ts#L209).

dev-packages/bundle-plugin/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "@theia/native-webpack-plugin",
+  "name": "@theia/bundle-plugin",
   "version": "1.70.0",
```
sdirix: `esbuild-plugin.ts` imports `resolve-package-path`, but it's not listed in `dependencies` here. It happens to be available as a transitive dep from `@theia/application-package`, but that's fragile. Should be an explicit dependency.

dev-packages/application-manager/src/generator/bundler-generator.ts
```diff
@@ -583,6 +632,253 @@ module.exports = {
...
+    '.jpg': 'dataurl',
+    '.png': 'dataurl',
+    '.gif': 'dataurl',
+    '.wasm': 'dataurl',
+    '.plist': 'dataurl',
+    '.node': 'file'
+};
+
+/**
+ * {@type {import('esbuild').BuildOptions}}
```
sdirix: `{@type ...}` is the inline JSDoc tag syntax, which isn't valid here. This should be `@type {import('esbuild').BuildOptions}` (without the outer braces). Same issue on the node and electron config methods below.

dev-packages/application-manager/src/generator/bundler-generator.ts
```diff
@@ -583,6 +632,253 @@ module.exports = {
...
+function join(...parts) {
+    return path.join(...parts).replace(/\\\\/g, '/');
+}
+
+const { mode, staticCompression, watch } = yargs.option('mode', {
+    description: "Mode to use",
+    choices: ["development", "production"],
+    default: "production"
+}).option('static-compression', {
+    description: 'Controls whether to enable compression of static artifacts. TODO: Not implemented yet.',
```
sdirix: This `TODO` will show up in every generated esbuild config. Maybe remove and track as an issue instead?

dev-packages/application-manager/src/generator/bundler-generator.ts
```diff
@@ -18,24 +18,58 @@ import * as paths from 'path';
...
+            if (await this.shouldGenerateUserWebpackConfig()) {
+                await this.write(this.configPath, this.compileUserWebpackConfig());
+            }
         }
-        if (await this.shouldGenerateUserWebpackConfig()) {
-            await this.write(this.configPath, this.compileUserWebpackConfig());
+    }
+
+    async preferESBuild(): Promise<boolean> {
+        // If a esbuild file already exists, prefer esbuild
```
sdirix: Nit: "If a esbuild" should be "If an esbuild".

dev-packages/application-manager/src/generator/bundler-generator.ts
```diff
@@ -583,6 +632,253 @@ module.exports = {
     nativePlugin,
     ignoredResources
 };
+`;
+    }
+
+    compileESBuildBrowserConfig(): string {
```
sdirix: Should these be `protected` like their webpack counterparts?

doc/Migration.md
ndoschek: Ah fair enough, with the options fully encapsulated there's nothing extra to document. Leaving as-is is fine by me :+1:

dev-packages/application-manager/src/generator/bundler-generator.ts
```diff
@@ -583,6 +632,253 @@ module.exports = {
...
+                    // copy webview files to lib folder
+                    from: join(resolvePackagePath('@theia/plugin-ext', __dirname), '..', 'src', 'main', 'browser', 'webview', 'pre', '*'),
+                    to: join(__dirname, 'lib', 'webview', 'pre')
+                }`)}${this.ifPackage('@theia/plugin-ext-vscode', `,
+                {
+                    // copy frontend plugin host files
+                    from: join(resolvePackagePath('@theia/plugin-ext-vscode', __dirname), '..', 'lib', 'node', 'context', 'plugin-vscode-init-fe.js'),
+                    to: join(__dirname, 'lib', 'frontend', 'context')
+                }`)}
+            ]
```
sdirix, line 721: The webpack browser config copies `@theia/terminal`'s shell integration scripts to `lib/backend/shell-integrations`, but the esbuild configs don't copy them at all. Terminal shell integration will therefore silently fail in the esbuild build because the injector won't find the scripts.

sdirix (commented):
Still works for me and the speed is amazing. I left some comments.

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
msujew: Honestly, I have no clue what this even does. Is the `worker-main` file ever compiled by Webpack/Esbuild? Esbuild just complained that `import.meta.url` is unavailable here, and inserts `""` by default.

packages/core/src/browser/style/select-component.css
msujew: It seems like this was no longer necessary and broke the select-component when compiling with ESBuild. The select component works as expected for both ESBuild and Webpack when removing this line of CSS.

packages/core/src/browser/style/index.css
msujew: No idea why, but without it, the applicationshell is only at like 20% height.

examples/electron/esbuild.mjs
msujew: The default options contain a plugin for logging all errors. Logging the `err` should not be necessary or will lead to duplicate output. I'll remove the `(err)` part.

msujew:
@sdirix Thanks for the review! I believe I was able to fix everything except for https://github.com/eclipse-theia/theia/pull/14414#discussion_r3086373553 - Is there a straightforward way to test this?

dev-packages/application-manager/src/application-package-manager.ts
```diff
@@ -73,12 +73,14 @@ export class ApplicationPackageManager {
...
         await Promise.all([
             this.remove(this.pck.lib()),
             this.remove(this.pck.srcGen()),
-            this.remove(webpackGenerator.genConfigPath),
-            this.remove(webpackGenerator.genNodeConfigPath)
+            this.remove(bundlerGenerator.genConfigPath),
+            this.remove(bundlerGenerator.genNodeConfigPath),
+            this.remove(bundlerGenerator.genESBuildBrowserPath),
+            this.remove(bundlerGenerator.genESBuildNodePath),
         ]);
```
sdirix, line 85: `clean()` removes the browser and node esbuild configs but not the electron one. `genESBuildElectronPath` should be added here too, otherwise `theia clean` leaves a stale `gen-esbuild.electron.mjs` behind for electron apps.

packages/core/src/browser/frontend-application-module.ts
```diff
@@ -15,9 +15,11 @@
 // *****************************************************************************
 
 import '../../src/browser/style/index.css';
-require('../../src/browser/style/materialcolors.css').use();
+import '../../src/browser/style/materialcolors.css';
```
sdirix, line 18: On master this was `require('...materialcolors.css').use()` which activates webpack's lazy singleton style tag, see [here](https://github.com/eclipse-theia/theia/blob/d93a0bb1abcecbe4da7d25a0250918080375bd7d/dev-packages/application-manager/src/generator/webpack-generator.ts#L140-L154). Switching to a bare `import` means `.use()` is never called when building with webpack, so the theme CSS style tag may never be injected. Does theming still work correctly on the webpack path, I did not test it.

packages/plugin-ext/src/hosted/browser/plugin-worker.ts
```diff
@@ -26,10 +26,7 @@ export class PluginWorker {
     public readonly rpc: RPCProtocol;
 
     constructor() {
-        this.worker = new Worker(new URL('./worker/worker-main',
-            // @ts-expect-error (TS1343)
-            // We compile to CommonJS but `import.meta` is still available in the browser
-            import.meta.url));
+        this.worker = new Worker(new URL('./worker/worker-main'));
```
sdirix: I checked the documentation: https://docs.webpack.js.org/guides/web-workers

Webpack 5 recognizes the `new URL('./worker/worker-main', import.meta.url)` pattern and emits `worker-main` as a separate chunk. Without the second argument, webpack will NOT split the worker out.

This is required for browser plugins to work.

Maybe we should add it back for webpack?

~You can test it via the "Simple Browser" built-in plugin by executing "Simple Browser: Show". It works in latest Theia but fails on this branch. Note that you need to test a page without a proper CSP as all the regular pages fail as they don't like to be embedded. I tested it with a website served via localhost.~

~Edit: Actually it works for me on the branch, I entered a wrong URL when testing :sweat_smile:. Would need to check whether it still works on Webpack build too.~

I realized that non of the built-ins are pure browser extensions and therefore all of them run in the backend. I pushed a real example of a browser-only VS Code extension (i.e. web extension) with which you can reproduce the issue. It in fact fails on this branch while it works in the latest Theia.

dev-packages/application-manager/src/generator/frontend-generator.ts
```diff
@@ -66,6 +66,7 @@ export class FrontendGenerator extends AbstractGenerator {
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta name="apple-mobile-web-app-capable" content="yes">
+  <link rel="stylesheet" href="./bundle.css">
```
sdirix: This `<link>` is emitted for all builds, but webpack inlines CSS into JS via `style-loader`, so no `bundle.css` file exists in the output. The browser will get a 404 for this file on webpack builds. Maybe we should emit this based on bundler type?

sdirix:
I realized that non of the built-ins are pure browser extensions and therefore all of them run in the backend. I pushed a real example of a browser-only VS Code extension (i.e. web extension) with which you can reproduce the issue.

For testing you need to copy the `sample-browser-plugin` from `sample-plugins` to `plugins`. It fails on this branch and it works on `master`.

msujew:
> For testing you need to copy the sample-browser-plugin from sample-plugins to plugins. It fails on this branch and it works on master.

Thanks, I'll take a look at that tomorrow 👍

packages/core/src/browser/frontend-application-module.ts
msujew: It didn't, but that's only because `materialcolors.css` was explicitly excluded from the default css loader, and had a separate loader - which I've now removed. I'm wondering why this is in place at all actually. Any idea?

There is no comparable feature like the style-loader `lazySingletonStyleTag` in ESBuild AFAIK. It now works as expected for both esbuild and webpack.

msujew:
@sdirix Thanks a lot for the input. I was able to work around this bundler behavior by adding an explicit bundling instruction for the `worker-main` for both bundlers. I didn't like to rely on bundler magic there. Using `import.meta.url` broke the esbuild build (it built with a warning, and crashed during runtime).

ndoschek:
Hi again @msujew,
the release caused some conflicts on your branch - could you do another rebase when you have a minute? TIA

dev-packages/application-manager/src/generator/bundler-generator.ts
```diff
@@ -583,6 +616,261 @@ module.exports = {
...
+    protected compileESBuildElectronConfig(): string {
+        return `/**
+ * Don't touch this file. It will be regenerated by theia build.
+ * To customize the build process, change ./esbuild.mjs
+ */
+import { problemMatcherPlugin } from '@theia/bundle-plugin';
+import { watch, loader, minify, sourcemap } from './gen-esbuild.browser.mjs';
+
+/**
+ * {@type {import('esbuild').BuildOptions}}
```
sdirix: The previous `{@type ...}` JSDoc fix wasn't applied here. Should be `@type {import('esbuild').BuildOptions}` (no surrounding braces).

dev-packages/application-manager/src/generator/bundler-generator.ts
```diff
@@ -583,6 +616,261 @@ module.exports = {
...
+                    // copy shell integration scripts
+                    from: join(resolvePackagePath('@theia/terminal', __dirname), '..', 'src', 'node', 'shell-integrations', '**', '*'),
+                    to: join(__dirname, 'lib', 'backend', 'shell-integrations')
+                },`)}
+            ]
+        })
+    ]
+};
+`;
+    };
```
sdirix: Stray semicolon after the method body.

doc/Migration.md
sdirix, line 14: Should we also mention the `@theia/native-webpack-plugin` -> `@theia/bundle-plugin` rename here? We do in the changelog.

sdirix (approved):
Works for me including the web extension. _At least in the Browser environment. In Electron environement it does not run, but it also does not on `master`, so it's not a regression_

Thanks for the great work!

ndoschek:
Thanks again for all the great work on this!

Just FYI: I've taken care of the publishing side:

- Published `@theia/bundle-plugin` initially to npm so future canary/release publishes will work (see https://www.npmjs.com/package/@theia/bundle-plugin)
- Deprecated `@theia/native-webpack-plugin` with a pointer to the new package name (https://www.npmjs.com/package/@theia/native-webpack-plugin)

We'll continue as agreed with the switch to esbuild on the Theia IDE side.

PR #14419, opened 2024-11-08
Fix active notebook editor staying active as long as the editor is active

jonah-iden:
#### What it does

This fixes the activeTextEditor should stay as the active cell as long as the associated notebook editor is also active.
This is important for extensions with a chat window that want to access the current selected code while the chat input is active

#### How to test

You can find a test extension including vsix [here](https://github.com/jonah-iden/theia-notebook-selection-test)

You can use the `Notebook-Test: get current selection` to print out the current notebook selection or use the output channel to see the change events. 

Even when for example selecting the workspace search or debug side panels, the selection inside the cell should still be available in the extension.

#### Review checklist

#### Reminder for reviewers


msujew (approved):
Good solution, thanks 👍

PR #14424, opened 2024-11-08
Show arguments in tool call response renderer

### LINKED ISSUE #14423: [Theia AI] Tool calls do not show arguments in the response renderer

JonasHelming:
fixed #14423

#### What it does

Shows a expandable ">" on tool functions in the hat if they have arguments

#### How to test

run a tool call and look at the chat. E.g.: "How to build this project".
Use a long-running to see the arguments while progress is shown, e.g. "how many files are in the workspace" on a big project such as theia

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
@@ -50,7 +50,7 @@ export class ToolCallPartRenderer implements ChatResponsePartRenderer<ToolCallCh
     }
 
     protected renderCollapsibleArguments(args: string | undefined): ReactNode {
-        if (!args || !args.trim() || args.trim() === "{}") {
+        if (!args || !args.trim() || args.trim() === '{}') {
             return undefined;
         }
 
--- a/packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
@@ -29,17 +29,45 @@ export class ToolCallPartRenderer implements ChatResponsePartRenderer<ToolCallCh
         }
         return -1;
     }
-    render(response: ToolCallChatResponseContent): ReactNode {
-        return <h4 className='theia-toolCall'>
-            {response.finished ?
-                <details>
-                    <summary>Ran {response.name}</summary>
-                    <pre>{this.tryPrettyPrintJson(response)}</pre>
-                </details>
-                : <span><Spinner /> Running [{response.name}]</span>
-            }
-        </h4>;
 
+    renderCollapsibleArguments(args: string | undefined): ReactNode {
+        return (
+            args && (
+                <details style={{ display: 'inline' }}>
+                    <summary
+                        style={{
+                            display: 'inline',
+                            cursor: 'pointer',
+                            textDecoration: 'underline',
+                        }}>
+                        &gt;
+                    </summary>
+                    <span>{JSON.stringify(args, undefined, 2)}</span>
+                </details>
+            )
+        );
+    }
+
+    render(response: ToolCallChatResponseContent): ReactNode {
+        return (
+            <h4 className='theia-toolCall'>
+                {response.finished ? (
+                    <details>
+                        <summary>Ran [{response.name}
+                            {response.arguments && <>({this.renderCollapsibleArguments(response.arguments)})</>}
+                            ]</summary>
+                        <pre>{this.tryPrettyPrintJson(response)}</pre>
+                        {this.renderCollapsibleArguments(response.arguments)}
+                    </details>
+                ) : (
+                    <span>
+                        <Spinner /> Running [{response.name}
+                        {response.arguments && <>({this.renderCollapsibleArguments(response.arguments)})</>}
+                        ]
+                    </span>
+                )}
+            </h4>
+        );
     }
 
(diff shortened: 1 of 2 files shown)


packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
```diff
@@ -29,17 +29,45 @@ export class ToolCallPartRenderer implements ChatResponsePartRenderer<ToolCallCh
...
+
+    renderCollapsibleArguments(args: string | undefined): ReactNode {
+        return (
+            args && (
+                <details style={{ display: 'inline' }}>
+                    <summary
+                        style={{
+                            display: 'inline',
+                            cursor: 'pointer',
+                            textDecoration: 'underline',
```
planger, line 41: We should probably assign a descriptively named class and add these CSS settings to the CSS file for this particularly classed `summary` element.

packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
```diff
@@ -29,17 +29,45 @@ export class ToolCallPartRenderer implements ChatResponsePartRenderer<ToolCallCh
...
-        return <h4 className='theia-toolCall'>
-            {response.finished ?
-                <details>
-                    <summary>Ran {response.name}</summary>
-                    <pre>{this.tryPrettyPrintJson(response)}</pre>
+
+    renderCollapsibleArguments(args: string | undefined): ReactNode {
+        return (
+            args && (
+                <details style={{ display: 'inline' }}>
```
planger, line 36: We should probably assign a descriptively named class and add these CSS settings to the CSS file for this particularly classed `details` element.

packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
```diff
@@ -29,17 +29,45 @@ export class ToolCallPartRenderer implements ChatResponsePartRenderer<ToolCallCh
...
+    render(response: ToolCallChatResponseContent): ReactNode {
+        return (
+            <h4 className='theia-toolCall'>
+                {response.finished ? (
+                    <details>
+                        <summary>Ran [{response.name}
+                            {response.arguments && <>({this.renderCollapsibleArguments(response.arguments)})</>}
+                            ]</summary>
+                        <pre>{this.tryPrettyPrintJson(response)}</pre>
+                        {this.renderCollapsibleArguments(response.arguments)}
```
planger, line 60:
```suggestion
```

Is this duplication of the arguments below the tool call result on purpose? I found it surprising.

planger (changes requested):
Great idea to show the arguments! Thank you!

However, there are a few things I observed, which I'd like to clarify:

* The arguments are shown again beneath the tool call result (see line 60). Is this on purpose? I found it a bit surprising:
image

* I think it'd be more clear that `>` is an expandable if we rather use a colored icon instead of the plain `>`, similar to the triangle we use for the entire tool call. The mouse pointer config doesn't really help a lot, because we already set this to `pointer`.

image

* The JSON of the arguments is not really nicely formatted for me.

image

* Just a matter of taste, and I don't have a strong opinion on this aspect, but I find those surrounding `[` ... `]` a bit superfluous. Imho there should be as few clutter as possible and it should be as close as possible to normal language without extra technical characters. Only if people are interested in looking into details, those technical details should appear. Which leads me also to the suggestion to show the arguments just as a section of the tool call result if the user unfolded them and not with the rather technical `(>)`? Background is also that this is likely to be used in non-IDE products.

planger:
Sorry, one more thing: It would probably be clearer if we move the `renderCollapsibleArguments` method below the method that calls (`render`) it and make it `protected`.

JonasHelming:
i decided to leav the "()" in. It looks very weird without them. In a non-technical context, I believe you would need a custom renderer anyways TBH.
A good simple test case is "Show me the package.json" it will have one call without and one with a parameter.

JonasHelming:
@planger Ping :-)

planger (changes requested):
Thanks for the revision. However, for me this is now how it looks like (see line break after ...):

image

Also, if you don't put this in a `pre` element, the pretty printing of the JSON gets lost (see [here](https://github.com/eclipse-theia/theia/pull/14424/files#diff-56b8517f4348e0f66a8ff2f2c9c5c43c5e4c7673d3c9a7ee2910d7a98bbfa17dR60)).

JonasHelming:
> Thanks for the revision. However, for me this is now how it looks like (see line break after ...):
> 
> image

This drives me mad. Somehow, I am almost certain, this was not the case when I committed this, and I literally do not find the reason for this. I will continue to search.

> Also, if you don't put this in a `pre` element, the pretty printing of the JSON gets lost (see [here](https://github.com/eclipse-theia/theia/pull/14424/files#diff-56b8517f4348e0f66a8ff2f2c9c5c43c5e4c7673d3c9a7ee2910d7a98bbfa17dR60)).

This was actually on purpose. I you add <pre> then the argument is shown as mutliple lines. However, I am fine with this, too.

planger:
@JonasHelming Thanks!

> This was actually on purpose. I you add pre then the argument is shown as multiple lines. However, I am fine with this, too.

Ah ok, I was just wondering, because we stringify the JSON with 2 spaces (`undefined, 2`), so I expected the intention was to show it formatted across multiple lines. I'm fine with it either way.

JonasHelming:
@planger  Looks wonderful now, thank you so much!

planger (approved):

Issue #14361, opened 2024-10-28
AI Workspace agent is slow and inaccurate


dannaf:
### Bug Description:

The Theia AI Workspace Agent is slow and inaccurate on a basic 'smoke test', when asked a simple question: "how many files are in this repo?" while open in a workspace containing the theia repo.

### Steps to Reproduce:

1. Launch theia 1.54+ and configure the experimental AI features with an (official) OpenAI api key.  Configure the AI agents to be all disabled except the Workspace agent which is set to use `openai/gpt-4o-mini`.
2. Git clone `https://github.com/eclipse-theia/theia` and open it in theia as a workspace.
3. Ask the AI Chat "how many files are in this repo?"
4. Observe the results: ![image](https://github.com/user-attachments/assets/b802ff31-7b9d-4309-8ff3-4059f114654c)
5. Observe that it takes a very long time for it to answer such a simple question (approx. 2 minutes for me)
6. Observe that the answer is incorrect.  It responded with "There are a total of 386 files in this repository."  But presumably this is not correct.  And also it is inconsistent with the intermediary list that it used, below, when it ran getWorkspaceFileList, which by pasting into python as `files = <paste>` and counting with `len(files)` says there are 3757 entries in this list.

<details><summary>Partial view of what the list of files used as intermediary by Workspace Agent looked like (GitHub prevented the entire list due to a 65536 character limit on comment length)</summary>

```
[...]

Issue #14119, opened 2024-09-03
Theia AI Epic


JonasHelming:
This epic provides an overview of upcoming improvements and features for Theia AI and the AI integration in the Theia IDE:

# High prio
- [x] #14133 
- [x] When closing and reopening the chat, experimental warning is shown again	(150)
- [x] #14128
- [x] #14185
- [x] #14180
- [x] #14161
- [x] #14167
- [x] #14856
- [x] #14205
- [x] #14174
- [x] #14164
- [x] #14201
- [x] #14208
- [ ] #14209
- [x] #14236
- [x] #14768
# Medium prio
- [x] #14613
- [x] #14512
- [x] #14485
- [x] #14469
- [x] #14471
- [x] #14464
- [x] #14460
- [x] #14423
- [x] #14411
- [x] #14392
- [x] Invalid Response Format error	(148)
- [x] #14183
- [x] Error is thrown regarding templates (134)
- [x] Error is thrown in backend (extensions) in regards to ChatView (135)
- [x] Responses are recorded multiple times (136)
- [x] #14252
- [x] #14245
- [x] #14322
- [x] Consider limiting the scope of tool functions (107)
- [x] #14286
- [x] Chat view conflicts with vim extension in Theia IDE (133)
- [x] Response Part Renderer for `code-diff` (20)
- [x] #14313
- [x] #14312
- [x] User context aware chat agent (159)
- [x] Terminal Chat Provider (129)
- [x] Response Part Renderer to execute terminal commands (52)
- [x] Edit user prompt (102)
- [x] Prompt debugging support (103)
- [x] Offer a way to clear history (139)
- [x] Description of Agents in the Agent Configuration is cut (123)
- [x] Explicitly support multi-turn sessions (162)
- [x] Support attaching arbitrary blobs of data to responses (163)
- [ ] https://github.com/eclipse-theia/theia/issues/14324
- [x] #14329
- [x] #14448
- [x] Workspace functions ignore multiple workspaces
- [x] #14441
- [ ] #14452
- [ ] https://github.com/eclipse-theia/theia/issues/14543#event-15452436624

JonasHelming:
All open issues transfered to here: #15068

PR #14426, opened 2024-11-09
Improve Workspace agent functions and prompt

JonasHelming:
related to #14361

#### What it does

- Add new function GET_WORKSPACE_DIRECTORY_STRUCTURE_FUNCTION
- Refactored GET_WORKSPACE_FILE_LIST_FUNCTION to only return the files in one directory
- Improved function error handling
- Improved Workspace prompt

#### How to test

Play with the workspace agent, e.g. let it search for a file or for an information.
Or let it create code based on a template, e.g.

"I want to integrate Anthropic as a LLM provider in the Theia IDE. There is already an integration of OpenAI that you can use as a template. It can be found in my workspace under ai-openai. there is another example for huggingface under ai-hugging-face Look at all files in these two examples and then generate all necessary files for me to add a new package "ai-anthropic" that uses the anthropic Typescript API (@anthropic-ai/sdk). Add "claude-3-5-sonnet-20241022" as an example model"

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-workspace-agent/src/browser/functions.ts
+++ b/packages/ai-workspace-agent/src/browser/functions.ts
@@ -19,11 +19,61 @@ import { inject, injectable } from '@theia/core/shared/inversify';
 import { FileService } from '@theia/filesystem/lib/browser/file-service';
 import { FileStat } from '@theia/filesystem/lib/common/files';
 import { WorkspaceService } from '@theia/workspace/lib/browser';
-import { FILE_CONTENT_FUNCTION_ID, GET_WORKSPACE_FILE_LIST_FUNCTION_ID } from '../common/functions';
+import { FILE_CONTENT_FUNCTION_ID, GET_WORKSPACE_DIRECTORY_STRUCTURE_FUNCTION_ID, GET_WORKSPACE_FILE_LIST_FUNCTION_ID } from '../common/functions';
 
-/**
- * A Function that can read the contents of a File from the Workspace.
- */
+function shouldExclude(stat: FileStat): boolean {
+    const excludedFolders = ['node_modules', 'lib'];
+    return stat.resource.path.base.startsWith('.') || excludedFolders.includes(stat.resource.path.base);
+}
+
+@injectable()
+export class GetWorkspaceDirectoryStructure implements ToolProvider {
+    static ID = GET_WORKSPACE_DIRECTORY_STRUCTURE_FUNCTION_ID;
+
+    getTool(): ToolRequest {
+        return {
+            id: GetWorkspaceDirectoryStructure.ID,
+            name: GetWorkspaceDirectoryStructure.ID,
+            description: 'Retrieve the complete directory structure of the workspace, listing only directories (no file contents). This structure excludes specific directories, such as node_modules and hidden files, ensuring paths are within workspace boundaries.',
+            handler: () => this.getDirectoryStructure()
+        };
+    }
+
+    @inject(WorkspaceService)
+    protected workspaceService: WorkspaceService;
+
+    @inject(FileService)
+    protected readonly fileService: FileService;
+
+    private async getDirectoryStructure(): Promise<string[]> {
+        const wsRoots = await this.workspaceService.roots;
+
+        if (wsRoots.length === 0) {
+            throw new Error('Workspace root not found');
+        }
+
+        const workspaceRootUri = wsRoots[0].resource;
+
+        return this.buildDirectoryStructure(workspaceRootUri);
+    }
+
+    private async buildDirectoryStructure(uri: URI, prefix: string = ''): Promise<string[]> {
+        const stat = await this.fileService.resolve(uri);
+        const result: string[] = [];
+
+        if (stat && stat.isDirectory && stat.children) {
+            for (const child of stat.children) {
+                if (!child.isDirectory || shouldExclude(child)) continue;
+                const path = `${prefix}${child.resource.path.base}/`;
+                result.push(path);
+                result.push(...await this.buildDirectoryStructure(child.resource, `${path}`));
+            }
+        }
+
+        return result;
+    }
+}
+
 @injectable()
 export class FileContentFunction implements ToolProvider {
     static ID = FILE_CONTENT_FUNCTION_ID;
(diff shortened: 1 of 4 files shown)


packages/ai-workspace-agent/src/browser/functions.ts
```diff
@@ -19,11 +19,62 @@ import { inject, injectable } from '@theia/core/shared/inversify';
 import { FileService } from '@theia/filesystem/lib/browser/file-service';
 import { FileStat } from '@theia/filesystem/lib/common/files';
 import { WorkspaceService } from '@theia/workspace/lib/browser';
-import { FILE_CONTENT_FUNCTION_ID, GET_WORKSPACE_FILE_LIST_FUNCTION_ID } from '../common/functions';
+import { FILE_CONTENT_FUNCTION_ID, GET_WORKSPACE_DIRECTORY_STRUCTURE_FUNCTION_ID, GET_WORKSPACE_FILE_LIST_FUNCTION_ID } from '../common/functions';
+
+function shouldExclude(stat: FileStat): boolean {
+    const excludedFolders = ['node_modules', 'lib'];
```
planger, line 25: I know this was there before, but it'd be better to at least put the function as a method of the tool, so adopters can at least customize this hard-coded list of folders to be excluded.
Ideally it should even be an injectable service, maybe with a default implementation that either provides a commonly useful list or looks into the gitignore.

packages/ai-workspace-agent/src/browser/functions.ts
```diff
@@ -19,11 +19,62 @@ import { inject, injectable } from '@theia/core/shared/inversify';
...
+    protected readonly fileService: FileService;
+
+    private async getDirectoryStructure(): Promise<string[]> {
+        const wsRoots = await this.workspaceService.roots;
+
+        if (wsRoots.length === 0) {
+            throw new Error('Workspace root not found');
+        }
+
+        const workspaceRootUri = wsRoots[0].resource;
```
planger, line 56: With just looking into the first workspace root, we don't really support multi-root workspaces. I think it'd be good to try supporting multi-root workspaces here. We could try to just make this explicit to the LLM:
```
{
  "<path-of-workspace-root[0]>": [ <file1>, ... ],
  "<path-of-workspace-root[1]>": [ <file1>, ... ],
}
```

And then in the `FileContentFunction` we'd need the <path-of-workspace-root> as a parameter or request absolute paths again? Maybe that would work?

packages/ai-workspace-agent/src/browser/functions.ts
```diff
@@ -19,11 +19,62 @@ import { inject, injectable } from '@theia/core/shared/inversify';
...
+    protected workspaceService: WorkspaceService;
+
+    @inject(FileService)
+    protected readonly fileService: FileService;
+
+    private async getDirectoryStructure(): Promise<string[]> {
+        const wsRoots = await this.workspaceService.roots;
+
+        if (wsRoots.length === 0) {
+            throw new Error('Workspace root not found');
```
planger, line 53:
```suggestion
            throw new Error('No workspace has been opened yet');
```

If we give this error to the LLM, it would probably give better answers to the user if no workspace folder(s) have been opened.

packages/ai-workspace-agent/src/browser/functions.ts
```diff
@@ -61,15 +114,36 @@ export class FileContentFunction implements ToolProvider {
...
+        }
+
+        const workspaceRootUri = wsRoots[0].resource;
+
+        const targetUri = workspaceRootUri.resolve(file);
+
+        if (!targetUri.toString().startsWith(workspaceRootUri.toString())) {
+            throw new Error('Access outside of the workspace is not allowed');
+        }
```
planger, line 129: This code is pretty redundant across multiple tools (besides the lack of multi-root workspace support). Maybe we should extract that into a common place.

packages/ai-workspace-agent/src/browser/functions.ts
```diff
@@ -90,45 +177,51 @@ export class GetWorkspaceFileList implements ToolProvider {
...
+        if (wsRoots.length === 0) {
+            throw new Error('Workspace root not found');
+        }
+
+        const workspaceRootUri = wsRoots[0].resource;
+        const targetUri = path ? workspaceRootUri.resolve(path) : workspaceRootUri;
+
+        if (!targetUri.toString().startsWith(workspaceRootUri.toString())) {
+            throw new Error('Access outside of the workspace is not allowed');
+        }
```
planger, line 192: See comments above

planger (changes requested):
Thank you for improving the workspace agent! If this agent is optimized, there is huge potential for being super useful!

I played around with your suggestion, but found a few cases, where it actually performs worse because of directing the response to just directories and omitting important files on the root level. So I'm not sure if this fits all use cases, unless we guide the LLM more in the system prompt to consider the root contents with a higher priority.

### Before

It read the README or the package.json to give a concise and correct answer:

image

### After

It actually looked into `src/index.ts` and guessed that this maybe is built with `npm`. It was right here, but just by accident (this may well have been a yarn or pnpm project).

image

### Why?

As can be seen, it will never "see" the readme or the package.json which both are crucial to answer important questions with the new approach, while it directly went to the readme or the package.json in the previous approach.

packages/ai-workspace-agent/src/browser/functions.ts
JonasHelming: Maybe two settings:
consider .gitignore: boolean
ignore directories: String[]
plus an injectable service?

planger: Yeah, that would be great plus extra points :-)
To me it'd be important that platform adopters can easily customize the filter list via injection and that there are reasonable defaults for Theia IDE. Configuration options for the Theia IDE user is then extra nice on top.

JonasHelming: OK, did the services and added user settings as a follow-up (https://github.com/eclipse-theia/theia/issues/14119)

packages/ai-workspace-agent/src/browser/functions.ts
JonasHelming: I would prefer making "mutiple workspaces support" a follow-up, added it here: https://github.com/eclipse-theia/theia/issues/14119

JonasHelming:
Changes:
- Adapt prompt to make "root" files more important
- Refactored code to remove duplication
- Make WorkspaceUtils a service including "shouldIgnore"
- Pass "Workspace not found" error to the LLM

planger (approved):
Great, thanks for the revision! Looks good to me and the previous use cases now work much better again.

Just one optional minor:
I'd rename `WorkspaceUtils` into something more specific to the workspace agent, e.g. `WorkspaceAgentScope` or something like that. Since this is a global key, a generic name may more likely clash or confuse at some point.

JonasHelming:
@planger Great point, I renamed it to WorkspaceFunctionScope

PR #14432, opened 2024-11-11
fix(ai): adjust chat input height dynamically

planger:
#### What it does

So far the height of the chat input field only was increased with Ctrl+Enter, but not when the lines broke due to their length, or when the view was resized.

With this change, we adapt the height of the input also during normal typing when the content wraps into a new line. Moreover, we add a short transition when adjusting the height to make it look smoother.

#### How to test
[output2.webm](https://github.com/user-attachments/assets/1ba967fc-f87a-4485-88b1-f26b03349a99)

Type long text into the AI Chat input and observe how the height is adjusted up to a maximum height. Also resize the chat window until the line wrapping is affected, and type Ctrl+Enter.

#### Follow-ups

None

#### Review checklist

#### Reminder for reviewers


sdirix (approved):
Works great for me. Good work! :heart: 

I just have a minor nitpick.

Issue #14431, opened 2024-11-11
samples generated by `yo` can not be used nor built


cbries:
### Bug Description:
The samples generated by `yo` can not be used nor built.
Upgrade of `yo` has been done via: 
`npm update -g yo generator-theia-extension`

### Steps to Reproduce:

1. open PowerShell 7 as Administrator
2. go to `c:\git\tmp-theia\helloWorldIssue`
3. call `yo theia-extension`
4. select "Hello World" with <enter>
5. take default project name "hello-world"

First issue, few errors:
```
PS C:\git\tmp-theia\helloWorldIssue> yo theia-extension
? The extension's type Hello World
? The extension's name hello-world
HelloWorld
   create package.json
   create lerna.json
   create .gitignore
   create README.md
   create .vscode\launch.json
   create hello-world\package.json
   create hello-world\tsconfig.json
   create hello-world\src\browser\hello-world-frontend-module.ts
   create hello-world\src\browser\hello-world-contribution.ts
   create hello-world\README.md
   create browser-app\package.json
   create electron-app\package.json
Installing dependencies
node:events:497
      throw er; // Unhandled 'error' event
      ^

Error: spawn yarn ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:286:19)
    at onErrorNT (node:internal/child_process:484:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
Emitted 'error' event on ChildProcess instance at:
    at ChildProcess._handle.onexit (node:internal/child_process:292:12)
    at onErrorNT (node:internal/child_process:484:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'spawn yarn',
  path: 'yarn',
  spawnargs: []
}

Node.js v20.18.0
```

6. enter `yarn install`
Result:
```
$ lerna run prepare
lerna info version 2.4.0
hello-world: > hello-world@0.0.0 prepare
hello-world: > yarn run clean && yarn run build
hello-world: yarn run v1.22.22
hello-world: warning package.json: No license field
hello-world: $ rimraf lib
hello-world: Done in 0.35s.
hello-world: yarn run v1.22.22
hello-world: warning package.json: No license field
hello-world: $ tsc
hello-world: Done in 2.56s.
lerna success run Ran npm script 'prepare' in packages:
lerna success - hello-world
Done in 129.55s.
```

[...]

PR #14435, opened 2024-11-12
Update Inversify to latest

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/14431.

The latest minor version of Inversify (6.1.0) introduced some behavior changes with regards to injected base classes. Inversify validates that the constructor arguments are in some way satisfied (by asserting constructor function length). Since we sometimes use an inheritance pattern that breaks this assertion, this leads to runtime failures in newer versions of Inversify.

This change adds `@unmanaged()` decorations to all constructors that aren't injected. Also adds a few more type constraints during compile time where Inversify got more strict in its typing.

#### How to test

Assert that the CI is green, and the application builds and runs without any start error.

#### Follow-ups

I couldn't really find *all* places where we're using the above mentioned pattern, as I couldn't find a way to easily search for it.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -51,7 +51,7 @@
     "http-proxy-agent": "^5.0.0",
     "https-proxy-agent": "^5.0.0",
     "iconv-lite": "^0.6.0",
-    "inversify": "^6.0.1",
+    "inversify": "^6.1.3",
     "jschardet": "^2.1.1",
     "keytar": "7.2.0",
     "lodash.debounce": "^4.0.8",
--- a/dependency-check-baseline.json
+++ b/dependency-check-baseline.json
@@ -1,3 +1,7 @@
 {
-  "npm/npmjs/-/advanced-mark.js/2.6.0": "Manually approved"
+  "npm/npmjs/-/advanced-mark.js/2.6.0": "Manually approved",
+  "npm/npmjs/-/inversify/6.1.3": "Manually approved",
+  "npm/npmjs/@inversifyjs/common/1.3.2": "Manually approved",
+  "npm/npmjs/@inversifyjs/core/1.3.3": "Manually approved",
+  "npm/npmjs/@inversifyjs/reflect-metadata-utils/0.2.2": "Manually approved"
 }
--- a/examples/api-samples/src/common/vsx/sample-ovsx-client-factory.ts
+++ b/examples/api-samples/src/common/vsx/sample-ovsx-client-factory.ts
@@ -22,9 +22,9 @@ export function rebindOVSXClientFactory(rebind: interfaces.Rebind): void {
     // rebind the OVSX client factory so that we can replace patterns like "${self}" in the configs:
     rebind(OVSXUrlResolver)
         .toDynamicValue(ctx => {
-            const appInfo = ctx.container.get(SampleAppInfo);
+            const appInfo = ctx.container.get<SampleAppInfo>(SampleAppInfo);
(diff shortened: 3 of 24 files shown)


packages/core/package.json
```diff
@@ -51,7 +51,7 @@
     "http-proxy-agent": "^5.0.0",
     "https-proxy-agent": "^5.0.0",
     "iconv-lite": "^0.6.0",
-    "inversify": "^6.0.1",
```
sdirix, line 54: Inversify `6.1.3` was just released, containing the build fix. So we should already indicate that we want to use that.

msujew: Yeah, good idea.

sdirix (commented):
I was able to build and start the browser application and did not encounter any error just by clicking a bit around and trying some features.

I then checked with the regex `@injectable(.|\n)*?constructor` to find all places where an `@unmanaged` might be needed.

I found the following `@injectable` classes which either have constructor parameters which are neither injected nor are declared unmanaged or execute a `super` call to a parent class without such declarations for their parameters.

- [AbstractDialog](https://github.com/eclipse-theia/theia/blob/1900a805a877e7bee9d87e228036060b8d91b233/packages/core/src/browser/dialogs.ts#L154-L155)
- [AbstractTreeDecoratorService](https://github.com/eclipse-theia/theia/blob/1900a805a877e7bee9d87e228036060b8d91b233/packages/core/src/browser/tree/tree-decorator.ts#L173)
- [BaseWidget](https://github.com/eclipse-theia/theia/blob/1900a805a877e7bee9d87e228036060b8d91b233/packages/core/src/browser/widgets/widget.ts#L120-L122)
- [BlameDecorator](https://github.com/eclipse-theia/theia/blob/1900a805a877e7bee9d87e228036060b8d91b233/packages/git/src/browser/blame/blame-decorator.ts#L31)
- [MarkerTree](https://github.com/eclipse-theia/theia/blob/1900a805a877e7bee9d87e228036060b8d91b233/packages/markers/src/browser/marker-tree.ts#L35-L36)
- [HostedPlugin](https://github.com/eclipse-theia/theia/blob/1900a805a877e7bee9d87e228036060b8d91b233/packages/plugin-ext/src/hosted/common/hosted-plugin.ts#L106)
- [ProxyPluginDeployerEntry](https://github.com/eclipse-theia/theia/blob/1900a805a877e7bee9d87e228036060b8d91b233/packages/plugin-ext/src/main/node/plugin-deployer-proxy-entry-impl.ts#L28)
- [AbstractHostedPluginSupport](https://github.com/eclipse-theia/theia/blob/1900a805a877e7bee9d87e228036060b8d91b233/packages/plugin-ext/src/hosted/common/hosted-plugin.ts#L106)
- [Process](https://github.com/eclipse-theia/theia/blob/1900a805a877e7bee9d87e228036060b8d91b233/packages/process/src/node/process.ts#L89-L92)
- [Task](https://github.com/eclipse-theia/theia/blob/1900a805a877e7bee9d87e228036060b8d91b233/packages/task/src/node/task.ts#L46-L48)
- [DefaultTerminalProfileService](https://github.com/eclipse-theia/theia/blob/1900a805a877e7bee9d87e228036060b8d91b233/packages/terminal/src/browser/terminal-profile-service.ts#L98)

sdirix (approved):
Works for me! Thanks for the work!

tsmaeder:
What's keeping this PR from being merged?

PR #14443, opened 2024-11-13
fix: AI history view crashing on first use

sdirix:
#### What it does

The AI history view sometimes crashed on first use when the user tried to select an agent. This was caused by an empty agent list bound in the initial UI.

Refactors the history view by making sure that the history view is only instantiated once it is shown instead of early in application start. This avoids binding a still empty list of agent contributions to the UI.

Also adds fallback content in case there are no registered agents as the used 'SelectComponent' crashes with empty options.

Also moves the history sort change tracking outside of the widget. Previously this was broken as only the initial widget was tracked.

##### Crash fix

1. Start the example application
2. Open the chat history view (should not have been open before)
3. Try to select an agent

Observe that there is no longer a crash of the view

##### Sorting fix

1. Start the example application
2. Open the chat history view
3. Close the chat history view
4. Open the chat history view
5. Click chat history sorting button a view times

Observe that the sorting button is updated on every click and keeps working

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-history/src/browser/ai-history-contribution.ts
+++ b/packages/ai-history/src/browser/ai-history-contribution.ts
@@ -15,7 +15,7 @@
 // *****************************************************************************
 import { FrontendApplication, codicon } from '@theia/core/lib/browser';
 import { AIViewContribution } from '@theia/ai-core/lib/browser';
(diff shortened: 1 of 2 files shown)


planger (approved):
Tested it and could not reproduce the original issues, thank you!

PR #14449, opened 2024-11-13
Add preference to ignore files in workspace functions

### LINKED ISSUE #14448: [Theia AI] User setting for ignore files in Workspace functions (including considering .gitignore)

JonasHelming:
fixed #14448

#### What it does

adds two settings to specify fles an directories to be ignored by the workspace functions:
- gitignore : Will respect the gitignore file
- User defined list of files or folders

#### How to test

Ignore files in both settings. Ask the workspace agent "is the a file/folder XYZ in the workspace". Specifically change the .gitignore at runtime to test file watching.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-workspace-agent/src/browser/functions.ts
+++ b/packages/ai-workspace-agent/src/browser/functions.ts
@@ -87,11 +87,11 @@ export class WorkspaceFunctionScope {
         return false;
     }
 
-    private isUserExcluded(fileName: string, userExcludePatterns: string[]): boolean {
+    protected isUserExcluded(fileName: string, userExcludePatterns: string[]): boolean {
         return userExcludePatterns.some(pattern => new Minimatch(pattern, { dot: true }).match(fileName));
     }
 
-    private async isGitIgnored(stat: FileStat, workspaceRoot: URI): Promise<boolean> {
+    protected async isGitIgnored(stat: FileStat, workspaceRoot: URI): Promise<boolean> {
         await this.initializeGitignoreWatcher(workspaceRoot);
 
         const gitignoreUri = workspaceRoot.resolve(this.GITIGNORE_FILE_NAME);
--- a/packages/ai-workspace-agent/src/browser/functions.ts
+++ b/packages/ai-workspace-agent/src/browser/functions.ts
@@ -24,6 +24,7 @@ import ignore from 'ignore';
 import { Minimatch } from 'minimatch';
 import { PreferenceService } from '@theia/core/lib/browser';
 import { CONSIDER_GITIGNORE_PREF, USER_EXCLUDE_PATTERN_PREF } from './workspace-preferences';
+
 @injectable()
 export class WorkspaceFunctionScope {
     protected readonly GITIGNORE_FILE_NAME = '.gitignore';
--- a/packages/ai-workspace-agent/src/browser/functions.ts
+++ b/packages/ai-workspace-agent/src/browser/functions.ts
@@ -98,13 +98,18 @@ export class WorkspaceFunctionScope {
 
         try {
             const fileStat = await this.fileService.resolve(gitignoreUri);
-            if (fileStat && !fileStat.isDirectory) {
+            if (fileStat) {
                 if (!this.gitignoreMatcher) {
                     const gitignoreContent = await this.fileService.read(gitignoreUri);
                     this.gitignoreMatcher = ignore().add(gitignoreContent.value);
                 }
                 const relativePath = workspaceRoot.relative(stat.resource);
-                if (relativePath && this.gitignoreMatcher.ignores(relativePath.toString())) { return true; }
+                if (relativePath) {
+                    const relativePathStr = relativePath.toString() + (stat.isDirectory ? '/' : '');
+                    if (this.gitignoreMatcher.ignores(relativePathStr)) {
+                        return true;
+                    }
+                }
             }
         } catch {
             // If .gitignore does not exist or cannot be read, continue without error
--- a/packages/ai-workspace-agent/src/browser/functions.ts
+++ b/packages/ai-workspace-agent/src/browser/functions.ts
@@ -20,11 +20,25 @@ import { FileService } from '@theia/filesystem/lib/browser/file-service';
 import { FileStat } from '@theia/filesystem/lib/common/files';
 import { WorkspaceService } from '@theia/workspace/lib/browser';
 import { FILE_CONTENT_FUNCTION_ID, GET_WORKSPACE_DIRECTORY_STRUCTURE_FUNCTION_ID, GET_WORKSPACE_FILE_LIST_FUNCTION_ID } from '../common/functions';
-
+import ignore from 'ignore';
+import { Minimatch } from 'minimatch';
+import { PreferenceService } from '@theia/core/lib/browser';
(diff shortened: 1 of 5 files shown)


packages/ai-workspace-agent/src/browser/functions.ts
```diff
@@ -39,15 +53,59 @@ export class WorkspaceFunctionScope {
...
+        this.fileService.onDidFilesChange(async event => {
+            if (event.contains(gitignoreUri)) {
+                this.gitignoreMatcher = undefined;
+            }
+        });
+
+        this.gitignoreWatcherInitialized = true;
+    }
+
+    async shouldExclude(stat: FileStat): Promise<boolean> {
```
planger, line 74: Nit pick: This method now got a bit overloaded, imho, because it does three things:
1. get preferences
2. check for exclusion of first pref (user excludes)
3. check for exclusion based on gitignore

Maybe we should extract 2 and 3 into separate methods:
```ts
async shouldExclude(stat: FileStat): Promise<boolean> {
    const shouldConsiderGitIgnore = this.preferences.get(CONSIDER_GITIGNORE_PREF, false);
    const userExcludePatterns = this.preferences.get<string[]>(USER_EXCLUDES_PREF, []);

    if (this.isUserExcluded(stat.resource.path.base, userExcludePatterns)) {
        return true;
    }

    if (shouldConsiderGitIgnore && await this.isGitIgnored(stat)) {
        return true;
    }

    return false;
}
```

It could be even better, if we prepare for multi-root support and hand the workspace root into those methods (`isUserExcluded` and `isGitIngored`) as parameters, then it'll be easier to support multi-root at some point.

packages/ai-workspace-agent/src/browser/functions.ts
```diff
@@ -39,15 +53,59 @@ export class WorkspaceFunctionScope {
...
+        const considerGitIgnore = this.preferences.get(CONSIDER_GITIGNORE_PREF, false);
+        const userExcludes = this.preferences.get<string[]>(USER_EXCLUDES_PREF, []);
+
+        const fileName = stat.resource.path.base;
+
+        for (const pattern of userExcludes) {
+            const matcher = new Minimatch(pattern, { dot: true });
+            if (matcher.match(fileName)) { return true; }
+        }
```
planger, line 83: Nitpick: Maybe the following is more idiomatic:
```ts
userExcludes.some(pattern => new Minimatch(pattern, { dot: true }).match(fileName));
```

packages/ai-workspace-agent/src/browser/functions.ts
```diff
@@ -39,15 +53,59 @@ export class WorkspaceFunctionScope {
...
+                    if (!this.gitignoreMatcher) {
+                        const gitignoreContent = await this.fileService.read(gitignoreUri);
+                        this.gitignoreMatcher = ignore().add(gitignoreContent.value);
+                    }
+                    const relativePath = workspaceRoot.relative(stat.resource);
+                    if (relativePath && this.gitignoreMatcher.ignores(relativePath.toString())) {
+                        return true;
+                    }
+                }
+            } catch (error) {
```
planger, line 103:
```suggestion
            } catch {
```

We don't need the error as a variable in that case.

planger (changes requested):
Very nice, thank you! The "User Excludes" seem to work very well.

However, in my tests the Git Ignore exclusion doesn't seem to work for the `getWorskpaceFileList` function:

image

A few minors:
* The title "Worskpace-function" stands out a bit, because it is the only pref title that uses a "-". In general, I think the labels are not very nice if they are camel-cased. I created a separate ticket on that (#14452)
* I think we should name the preference about user excludes more from the perspective of the user, e.g. "Excluded File Patterns" is better?

JonasHelming:
@planger I tried to adress all comments.
However, about the described Bug that .gitignore is not respected, I failed to reproduce this. Can you point me to the repo where you tried this?

packages/ai-workspace-agent/src/browser/functions.ts
```diff
@@ -20,12 +20,26 @@ import { FileService } from '@theia/filesystem/lib/browser/file-service';
 import { FileStat } from '@theia/filesystem/lib/common/files';
 import { WorkspaceService } from '@theia/workspace/lib/browser';
 import { FILE_CONTENT_FUNCTION_ID, GET_WORKSPACE_DIRECTORY_STRUCTURE_FUNCTION_ID, GET_WORKSPACE_FILE_LIST_FUNCTION_ID } from '../common/functions';
-
+import ignore from 'ignore';
+import { Minimatch } from 'minimatch';
+import { PreferenceService } from '@theia/core/lib/browser';
+import { CONSIDER_GITIGNORE_PREF, USER_EXCLUDE_PATTERN_PREF } from './workspace-preferences';
 @injectable()
```
planger, line 27:
```suggestion

@injectable()
```

packages/ai-workspace-agent/src/browser/functions.ts
```diff
@@ -39,16 +53,65 @@ export class WorkspaceFunctionScope {
...
+                }
+                const relativePath = workspaceRoot.relative(stat.resource);
+                if (relativePath && this.gitignoreMatcher.ignores(relativePath.toString())) { return true; }
+            }
+        } catch {
+            // If .gitignore does not exist or cannot be read, continue without error
+        }
+
+        return false;
+    }
```
planger, line 113:
```suggestion
    private async isGitIgnored(stat: FileStat, workspaceRoot: URI): Promise<boolean> {
        await this.initializeGitignoreWatcher(workspaceRoot);

        const gitignoreUri = workspaceRoot.resolve(this.GITIGNORE_FILE_NAME);

        try {
            const fileStat = await this.fileService.resolve(gitignoreUri);
            if (fileStat) {
                if (!this.gitignoreMatcher) {
                    const gitignoreContent = await this.fileService.read(gitignoreUri);
                    this.gitignoreMatcher = ignore().add(gitignoreContent.value);
                }
                const relativePath = workspaceRoot.relative(stat.resource);
                if (relativePath) {
                    const relativePathStr = relativePath.toString() + (stat.isDirectory ? '/' : '');
                    if (this.gitignoreMatcher.ignores(relativePathStr)) {
                        return true;
                    }
                }
            }
        } catch {
            // If .gitignore does not exist or cannot be read, continue without error
        }

        return false;
    }
```

I've debugged into the issue I had and understand now what's wrong:
1. we never filter directories at all based on the gitignore
2. the `ignoreMatcher` doesn't return `true` if in the gitignore it says `dist/` but the `relativePath` is `dist`

I adapted to code to imho work more as we would expect, but I'll leave it up to you if we want to filter directories. I think that we do, because otherwise the LLM would only see the directory but never a file it contains. This is ok for git (because it only deals with files anyway), but for the LLM it seems more correct to already remove the files.

Note that e.g. `dist` won't be filtered if the gitignore contains `dist/*`, but I'd argue with `dist/*` the user explicitly refers to contents of `dist` but not `dist` directly, while `dist/` is kind of the default everyone uses and it means, I want to ignore `dist`... but this is starting to get philosophical :-)

planger (approved):
Thanks for the revision! Looks much cleaner now imho!

Please see two more suggestions inline, also regarding the bug I observed. Eitherway, I'm fine with the current state, even though I think my suggestions make sense.

Issue #14455, opened 2024-11-15
Refactor `BaseWidget` lifecycle


tsmaeder:
### Feature Description:
In `BaseWidget.onCloseRequested()` we invoke the `dispose()` method in the widget. I believe this is wrong, due to the argument in https://github.com/eclipse-theia/theia/issues/12093#issuecomment-2476046868: 

> when a client invokes ApplicationShell.addWidget(), it can pass widgets which are not derived from BaseWidget. When we later call ApplicationShell.closeWidget() on that widget, we should invoke dispose() on the widget after calling close() on it, since we do not know that close() invokes dispose(). Currently, any widget that does not dispose itself upon close() will go undisposed and might leak resources. However, calling dispose() on a BaseWidget after it is close will double-dispose and throw an error.

The clients managing the widget lifecycle should be reponsible for disposing of the widgets, not the widget itself.

PR #14456, opened 2024-11-15
Do not dispose dialogs on close

### LINKED ISSUE #12093: Invalid lifecycle for ReactDialog
### Bug Description:

In https://github.com/eclipse-theia/theia/issues/11669, a fix was provided to support re-opening a ReactDialog that had been closed. However, trying to reopen/reuse a widget that has been previously disposed violates the widget API, which may cause further issues. Especially, BaseWidget.dispose() can only be invoked once, so the reused/revived widget can no longer be disposed after this.

As a consequence, dialogs based on ReactDialog now have an inconsistent lifecycle, which causes errors:

```
ERROR Warning: You are calling ReactDOMClient.createRoot() on a container that has already been passed to createRoot() before. Instead, call root.render() on the existing root instead if you want to update it.
```

My understanding is that we shouldn't be trying to reuse closed (disposed) Dialogs in the first place; instead, a new instance of the dialog should be created after the previous one was closed/disposed.

### Steps to Reproduce:

1. In the Theia Browser Example, open the AboutDialog
2. Close it using Esc key or X button: Dialog is properly disposed
3. Reopen it: Dialog is "successfully" reused, but is still flagged as disposed
4. Close it again: Dispose is not properly invoked, because the widget is already disposed
5. Reopen it once more: A React error is thrown, because we're trying to mount the dialog again, even though it wasn't successfully unmounted in the previous step

```
ERROR Warning: You are calling ReactDOMClient.createRoot() on a container that has already been passed to createRoot() before. Instead, call root.render() on the existing root instead if you want to update it.
```

### Additional Information

- Operating System: Any
- Theia Version: Tested on 1.33.0, but probably any version >= 1.30.0

tsmaeder:
#### What it does
Fixes #12093

The about dialog in Theia is reopened and must no be disposed on close. This PR overrides the `onCloseRequest` method from `BaseWidget` to not invoke `dispose()`. In order to minimize impact for clients of the API, I've added an optional parameter to the `open()` method that controls whether the dialog is disposed when it is closed. 
This change might be breaking in the case where clients reuse dialog instances and do not rely on `open()` to show the dialog. In that case, they will have to dispose the dialog themselves now. That should be a very rare case, though.

#### How to test
- Open and close the about dialog at least three times. Make sure there are not messages and exceptions in the browser log.
- Test behavior of both dismissing and accepting dialogs, for example
   - ConfirmationDialog: I used the "Clear editor history" commnand
   - Add a function breapoint in the "breakpoints" view
   - Make an editor dirty and close it: there is a dialog asking whether to save
Ensure these scenarios work correctly in the browser and Electron case.

#### Follow-ups
- [ ] https://github.com/eclipse-theia/theia/issues/14455

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - [Previous Changelogs](https://github.com/eclipse-theia/theia/tree/master/doc/changelogs/)
 
 <!-- ## 1.56.0 - not yet released -->
+<a name="breaking_changes_1.56.0">[Breaking Changes:](#breaking_changes_1.56.0)</a>
+- [core] Do not dispose dialogs on close - [#14456](https://github.com/eclipse-theia/theia/pull/14456) - Contributed on behalf of STMicroelectronics 
 
 ## 1.55.0 - 10/31/2024
 
--- a/packages/core/src/browser/common-frontend-contribution.ts
+++ b/packages/core/src/browser/common-frontend-contribution.ts
@@ -1199,7 +1199,7 @@ export class CommonFrontendContribution implements FrontendApplicationContributi
     }
 
     protected async openAbout(): Promise<void> {
-        this.aboutDialog.open();
+        this.aboutDialog.open(false);
     }
 
     protected shouldPreventClose = false;
--- a/packages/core/src/browser/dialogs.ts
+++ b/packages/core/src/browser/dialogs.ts
@@ -258,13 +258,15 @@ export abstract class AbstractDialog<T> extends BaseWidget {
         }
     }
 
-    open(): Promise<T | undefined> {
+    open(disposeOnResolve: boolean = true): Promise<T | undefined> {
         if (this.resolve) {
             return Promise.reject(new Error('The dialog is already opened.'));
         }
         this.activeElement = this.node.ownerDocument.activeElement as HTMLElement;
         return new Promise<T | undefined>((resolve, reject) => {
-            this.resolve = resolve;
+            this.resolve = value => {
+                resolve(value);
+            };
             this.reject = reject;
             this.toDisposeOnDetach.push(Disposable.create(() => {
                 this.resolve = undefined;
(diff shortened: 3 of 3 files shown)


sdirix (changes requested):
Thanks for the changes.

I tested all mentioned scenarios in the browser. There I could not see any errors.

I also tested all mentioned scenarios in Electron. There I was able to reproduce the following error when opening the about dialog multiple times: `ERROR Warning: You are calling ReactDOMClient.createRoot() on a container that has already been passed to createRoot() before. Instead, call root.render() on the existing root instead if you want to update it.` The other test scenarios did not produce any errors.

tsmaeder:
@sdirix I just opened and closed the AboutDialog 5 times in a row with now messages in the log. Where's the difference?

sdirix:
> @sdirix I just opened and closed (`Esc`) the AboutDialog in Electron 5 times in a row with now messages in the log. Where's the difference?

I retested and could no longer reproduce the issue. Not sure what I did different than the first time. As it could be a mistake on my side, I don't see this as a blocker no more.

---

A question before we merge the current state: Do you think it's really worth it to overwrite the `dispose` behavior of `ReactDialog` to satisfy this one dialog which does not behave as the others?

As an alternative we could adapt the `AboutDialog` to not be a singleton and then call the `widgetManager.getOrCreateWidget` whenever we want to show it. This would have the benefit that the `AboutDialog` is not preconstructed/injected when it's never even shown.

tsmaeder:
It's not about that one dialog: by having that dialog, we're communicating that it's O.K. to reopen dialogs. The documented contract for `Widget` says the same thing. For clients not reopening the dialog nothing changes, since they will get the auto-dispose behaviour from the `show()` method. Clients reusing dialogs are already broken by the double-dispose issue. What changes is that we give clients the opportunity to correctly implement dispose if they reuse dialogs.

sdirix (approved):

PR #14465, opened 2024-11-16
All three brackets for variables

### LINKED ISSUE #14464: [Theia AI] Allows three brackets for variables in prompt templates

JonasHelming:
fixed #14464

#### What it does

Allows to use three brackets for variables in prompts. No mixed formats are allows for consistency

#### How to test

There are automated test.
To manually test, 
- open the code completion prompt and a code file. 
- Add one bracket to {{textUntilCurrentPosition}}, see how it fails. 
- Add the second bracket to complete {{{textUntilCurrentPosition}}} and see how it works again. 
- Check that the variable is still listed in the AI configuration view then.
- Check that in the prompt editor, three brackets are valid, but mixtures are an error

#### Review checklist

#### Reminder for reviewers


packages/ai-core/data/prompttemplate.tmLanguage.json
```diff
@@ -49,4 +88,4 @@
   "fileTypes": [
     ".prompttemplate"
   ]
-}
+}
```
planger, line 91:
```suggestion
}

```

packages/ai-core/src/common/prompt-service.ts
```diff
@@ -181,13 +181,18 @@ export class PromptServiceImpl implements PromptService {
...
+    }
+
     async getPrompt(id: string, args?: { [key: string]: unknown }): Promise<ResolvedPromptTemplate | undefined> {
         const prompt = this.getRawPrompt(id);
         if (prompt === undefined) {
             return undefined;
         }
 
-        const matches = [...prompt.template.matchAll(PROMPT_VARIABLE_REGEX)];
+        const matches = this.matchVariables(prompt.template);
```
planger, line 195: Is there a reason why we extract this call of `matchVariablesRegEx()` into a method `this.matchVariables()`? Is this to simplify overriding it by adopters? Not sure this is a very common use case though.

I personally would keep this function call direct and not extract it into a separate method in this class (similar to how we do it in `agent-configuration-widget.tsx`.

If you like to keep the method (`this.matchVariables()`), I wouldn't make it `public` but rather `protected`; if we really want to keep the method at all.

planger: Should we also put the other invocation of `prompt.template.matchAll` into a function for consistency?
See `[...prompt.template.matchAll(PROMPT_FUNCTION_REGEX)]` on line 215.

planger (changes requested):
Thanks, interpreting the variables seem to work fine!

However, the syntax highlighting doesn't work for me:
image

Also I have a few comments inline above.

packages/ai-core/src/common/prompt-service.spec.ts
planger, line 1: Sorry, this message somehow got lost, readding it:

Nice, thanks for adding additional tests also covering the negative cases now.

I think we may want to avoid duplicating the three tests for the two- and three-bracket cases. This will reduce maintenance effort in the future. I think the following should work (but haven't tested it):

```ts
['{{', '{{{'].forEach((openingBrackets, index) => {
    const closingBrackets = openingBrackets.replace(/{/g, '}'); // Replace each `{` with `}`
    it(`should retrieve raw prompt by id (${index + 2} brackets)`, () => {
        const rawPrompt = promptService.getRawPrompt('8');
        expect(rawPrompt?.template).to.equal(`Hello, ${openingBrackets}name${closingBrackets}`);
    });
});
```

JonasHelming:
> However, the syntax highlighting doesn't work for me

Should be fixed

packages/ai-core/src/common/prompt-service.spec.ts
JonasHelming: I am no sure about this proposal. I agree that the duplication on the tests might require maintenance (if we ever change this). However, I am not a huge fan of making tests with loops instead of clearly writing them out. I think if the test ever break, it will be much easier to understand the test cases if they are simple.

planger (approved):
Thanks for the revision, works fine now! :+1:

PR #14470, opened 2024-11-17
Allows comments in prompt templates

### LINKED ISSUE #14469: [Theia AI] Allow comments in prompt templates

JonasHelming:
This should be reviewed first: https://github.com/eclipse-theia/theia/pull/14465

fixed #14469 

#### What it does

Allows comments in prompt templates
- Comments must start in the first line only, but can then be multile lines
- Comments are {{!-- comment --}}

#### How to test
There are automated tests.

Manual test:
Add comments and observe no effect on:
- LLM requests
- Variable detection in the AI configuration view

Check comments are highlighted correctly in the prompt editor:
image

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/data/prompttemplate.tmLanguage.json
+++ b/packages/ai-core/data/prompttemplate.tmLanguage.json
@@ -21,7 +21,7 @@
     },
     {
       "name": "comment.block.prompttemplate",
-      "begin": "^[\\t ]*{{!--",
+      "begin": "\\A{{!--",
       "beginCaptures": {
         "0": {
           "name": "punctuation.definition.comment.begin"
--- a/packages/ai-core/data/prompttemplate.tmLanguage.json
+++ b/packages/ai-core/data/prompttemplate.tmLanguage.json
@@ -20,6 +20,22 @@
       }
     },
     {
+      "name": "comment.block.prompttemplate",
+      "begin": "^[\\t ]*{{!--",
+      "beginCaptures": {
+        "0": {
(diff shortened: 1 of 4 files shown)


planger (commented):
Nice, works pretty well! I don't approve it yet, because it is based on another PR that hasn't been approved yet and I want to avoid accidentally merging the other commit.

Code looks good! I just have a few general thoughts that me might want to address:

image

Line 7 above:
The syntax highlighting shows the content of line 7 as comment, whereas in reality it isn't. This might be confusing.

Line 6 above:
It doesn't show the comment-like string as a comment, however, due to the similar begin and end delimiters to variables, it highlights this false comment as a variable. This again might be a bit confusing. Maybe it'd be better to choose a different sequence of characters for comments, to avoid this overlap and false highlighting.

JonasHelming:
I will check this once https://github.com/eclipse-theia/theia/pull/14465 is done, as it needs to be tested in combination

JonasHelming:
> Line 6 above:
> It doesn't show the comment-like string as a comment, however, due to the similar begin and end delimiters to variables, it highlights this false comment as a variable. This again might be a bit confusing. Maybe it'd be better to choose a different sequence of characters for comments, to avoid this overlap and false highlighting.

This is actually expected. This was a trade off. {{!-- is a very common pattern in template engines. Also we take less characters away for regular usage in prompts. So if {{!-- occurs somewhere not in the first line, it is indeed treated as a variable.

The other issue should be fixed now

planger (approved):

PR #14487, opened 2024-11-20
Support prompt variants

### LINKED ISSUE #14485: [Theia AI] Support prompt variants

JonasHelming:
fixed #14485

#### What it does

Introduce the capability for prompt variants, i.e. variants of a main prompt to e.g. support specific LLMs.
There is always a main prompt (default) that can have multiple variants.
The currently selected variant is a setting.

- Extended prompt service to manage variants
- Extended AgentSettings to store if a variant for a main template is selected.
- Prompt Variants can be edited as well
- Add an empty prompt for the universal model (useful to talk to small models or just "diretcly" to an LLM)

#### How to test

- Switch the prompt template for the universal agent and check in History view that the system prompt is adapted
- Restart to check the the setting is persistent
- Modify the variant
- etc.

NOTE: There is an unrelated bug in the prompt watching. I can sometimes get into a state, where I change a prompt template, but the prompt customization service is not notified anymore. This usually happens during demos, but I have no reproducible path yet.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
+++ b/packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
@@ -137,14 +137,26 @@ export class AIAgentConfigurationWidget extends ReactWidget {
                     Enable Agent
                 </label>
             </div>
-            <div className='ai-templates'>
-                {agent.promptTemplates?.map(template =>
-                    <TemplateRenderer
-                        key={agent?.id + '.' + template.id}
-                        agentId={agent.id}
-                        template={template}
-                        promptCustomizationService={this.promptCustomizationService} />)}
+            <div className="settings-section-subcategory-title ai-settings-section-subcategory-title">
+                Prompt Templates
             </div>
+            <div className='ai-templates'>
+                {agent.promptTemplates
+                    ?.filter(template => !template.variantOf)
+                    .map(template => (
+                        <div key={agent.id + '.' + template.id}>
+                            <TemplateRenderer
+                                key={agent?.id + '.' + template.id}
+                                agentId={agent.id}
+                                template={template}
+                                promptService={this.promptService}
+                                aiSettingsService={this.aiSettingsService}
+                                promptCustomizationService={this.promptCustomizationService}
+                            />
+                        </div>
+                    ))}
+            </div>
+
             <div className='ai-lm-requirements'>
                 <LanguageModelRenderer
                     agent={agent}
--- a/packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
+++ b/packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
@@ -140,22 +140,28 @@ export class AIAgentConfigurationWidget extends ReactWidget {
             <div className="settings-section-subcategory-title ai-settings-section-subcategory-title">
                 Prompt Templates
             </div>
-            <div className='ai-templates'>
-                {agent.promptTemplates
-                    ?.filter(template => !template.variantOf)
-                    .map(template => (
-                        <div key={agent.id + '.' + template.id}>
-                            <TemplateRenderer
-                                key={agent?.id + '.' + template.id}
-                                agentId={agent.id}
-                                template={template}
-                                promptService={this.promptService}
-                                aiSettingsService={this.aiSettingsService}
-                                promptCustomizationService={this.promptCustomizationService}
-                            />
-                        </div>
-                    ))}
+            <div className="ai-templates">
+                {(() => {
+                    const defaultTemplates = agent.promptTemplates?.filter(template => !template.variantOf) || [];
+                    return defaultTemplates.length > 0 ? (
+                        defaultTemplates.map(template => (
+                            <div key={agent.id + '.' + template.id}>
+                                <TemplateRenderer
+                                    key={agent.id + '.' + template.id}
+                                    agentId={agent.id}
+                                    template={template}
+                                    promptService={this.promptService}
+                                    aiSettingsService={this.aiSettingsService}
+                                    promptCustomizationService={this.promptCustomizationService}
+                                />
+                            </div>
+                        ))
+                    ) : (
+                        <div>No default template available</div>
+                    );
+                })()}
             </div>
+
 
             <div className='ai-lm-requirements'>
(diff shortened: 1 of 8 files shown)


planger:
Thanks, this enables a nice use case!

I'd be for option (b) or (c) -- tendency towards (b). I think the agent shouldn't really be aware of that at all when obtaining the prompt.

I tend to prefer keeping it as an agent setting too:
As far as I understand, the use case is a user switches to another LLM for an agent and as a consequence selects a variant of the prompt template that is optimized for the prompt. Both of these settings are currently scoped to the agent (at least practically).

The confusing thing is, that anyone can request this prompt template, and they would obtain the "agent-scoped" variant of the prompt when resolving it too. This is because prompt templates are per se indepenent of agents, but we only use it in an agent-specific way so far. However, I guess this is a rather theoretical problem, I can't see a lot of valid use cases where another agent than the agent who contributed the prompt template would consume it?
It feels like we should discourage that and rather say: create a copy of the template for your agent. WDYT?

packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
```diff
@@ -137,14 +137,26 @@ export class AIAgentConfigurationWidget extends ReactWidget {
...
+            </div>
             <div className='ai-templates'>
-                {agent.promptTemplates?.map(template =>
-                    <TemplateRenderer
-                        key={agent?.id + '.' + template.id}
-                        agentId={agent.id}
-                        template={template}
-                        promptCustomizationService={this.promptCustomizationService} />)}
+                {agent.promptTemplates
+                    ?.filter(template => !template.variantOf)
```
eneufeld, line 145: what if the agent does not provide a template without a variant? shouldn't we handle this? eg show 'no default template available'?

packages/ai-core/src/browser/ai-configuration/template-settings-renderer.tsx
```diff
@@ -14,26 +14,103 @@
...
+    return (
+        <div className="template-renderer">
+            <div className="settings-section-title template-header">
+                <strong>{template.id}</strong>
+            </div>
+            <div className="template-controls">
+                {variantIds.length > 1 && (
+                    <>
+                        <label htmlFor={`variant-selector-${template.id}`} className="template-select-label">
+                            Variant:
```
eneufeld, line 91: should this be the 'Selected Variant'?

packages/ai-core/src/common/agent-service.ts
```diff
@@ -99,7 +99,7 @@ export class AgentServiceImpl implements AgentService {
     registerAgent(agent: Agent): void {
         this._agents.push(agent);
         agent.promptTemplates.forEach(
-            template => this.promptService.storePrompt(template.id, template.template)
+            template => this.promptService.storePromptTemplate(template)
```
eneufeld, line 102: do we need the old method?

packages/ai-core/src/common/prompt-service.ts
```diff
@@ -68,6 +74,11 @@ export interface PromptService {
      * @param prompt the prompt template to store
      */
     storePrompt(id: string, prompt: string): void;
```
eneufeld, line 76: as above, do we need this anymore?

packages/ai-core/src/browser/ai-configuration/template-settings-renderer.tsx
```diff
@@ -14,26 +14,103 @@
...
+
+    const handleVariantChange = async (event: React.ChangeEvent<HTMLSelectElement>) => {
+        const newVariant = event.target.value;
+        setSelectedVariant(newVariant);
+
+        const agentSettings = await aiSettingsService.getAgentSettings(agentId);
+        const selectedVariants = agentSettings?.selectedVariants || {};
+
+        const updatedVariants = { ...selectedVariants };
+        if (newVariant === DEFAULT_VARIANT) {
```
eneufeld, line 60: why do we need to handle that case special?

eneufeld (commented):
I like the new feature, I left some small remarks

eneufeld:
Tested it, works great.

packages/ai-core/src/common/agent-service.ts
JonasHelming: This would break the API, I don't know whether people are already using it. For conveniance, I would probably leave it.

packages/ai-core/src/common/prompt-service.ts
JonasHelming: This would break the API, I don't know whether people are already using it. For conveniance, I would probably leave it.

packages/ai-core/src/browser/ai-configuration/template-settings-renderer.tsx
JonasHelming: I dont understand the question. If the default template is selected for a template ID, we remove the entry from the selected variants?

packages/ai-core/src/common/agent-service.ts
JonasHelming: Hmm, it is just a simpler version for all manditory values, I would keep it.

packages/ai-core/src/browser/ai-configuration/agent-configuration-widget.tsx
JonasHelming: image
Added

JonasHelming: Great suggestions, see above

JonasHelming:
Adressed all comments (except the last question)

packages/ai-core/src/browser/ai-configuration/template-settings-renderer.tsx
eneufeld: yeah i see that, why not just store that the user selected the default variant?
I'm fine also with the current solution was just wondering, whether there is a special reason to do so

JonasHelming: Hmm, I find it cleaner to only store if variants are actually selected.

sdirix (approved):
LGTM! Thanks!

As discussed, preferably a single method for storing templates would simplify the interface.

PR #14509, opened 2024-11-24
Allow adding variant via prompttemplate files

### LINKED ISSUE #14507: [Theia AI] Allow user to add custom prompt variants

JonasHelming:
fixed #14507

#### What it does

Allows users to add custom variants to existing main prompts via adding .prompttemplate files.
If the file name of a .prompttemplate file starts with the id of an existing main prompt, the custom prompt file is treated as a variant.

#### How to test

- Create a new .prompttemplate file which starts with the id of an existing main prompt template.
- Browse to the respective agent and select the variant
- Edit the variant
- No other prompt feature should be influenced

#### Follow-ups

we might allow users to create variants in the UI

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/browser/ai-configuration/template-settings-renderer.tsx
+++ b/packages/ai-core/src/browser/ai-configuration/template-settings-renderer.tsx
@@ -34,20 +34,19 @@ export const TemplateRenderer: React.FC<TemplateRendererProps> = ({
     promptService,
     aiSettingsService,
 }) => {
-    const [variantIds, setVariantIds] = React.useState<string[]>([]);
+    const variantIds = [DEFAULT_VARIANT, ...promptService.getVariantIds(template.id)];
     const [selectedVariant, setSelectedVariant] = React.useState<string>(DEFAULT_VARIANT);
 
-    React.useEffect(() => {
+    React.useMemo(() => {
         (async () => {
-            const variants = promptService.getVariantIds(template.id);
-            setVariantIds([DEFAULT_VARIANT, ...variants]);
-
             const agentSettings = await aiSettingsService.getAgentSettings(agentId);
             const currentVariant =
                 agentSettings?.selectedVariants?.[template.id] || DEFAULT_VARIANT;
             setSelectedVariant(currentVariant);
         })();
-    }, [template.id, promptService, aiSettingsService, agentId]);
+    }, [template.id, aiSettingsService, agentId]);
+
+    const isInvalidVariant = !variantIds.includes(selectedVariant);
 
     const handleVariantChange = async (event: React.ChangeEvent<HTMLSelectElement>) => {
         const newVariant = event.target.value;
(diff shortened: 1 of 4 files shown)


sdirix (commented):
Works for me!

I tested editing the workspace agent template. Creating a new template and setting it in the AI configuration works fine.

There are some glitches when deleting this file:
 - Create this custom variant template
 - Set it in the UI for an agent
 - Send a request for the agent, observe it is correctly set as system prompt

If the file is now deleted, the following happens:
- It's still shown in the AI configuration UI until the selection is switched back and forth
- Although it's gone, clicking on "Edit" will still open an editor for the no longer existing file
- When sending a request to the agent, it will use NO system prompt

What is likely going on is that, although deleted, Theia AI still tries to use the deleted prompt variant. In the UI this leads to the variant no longer appearing in the dropdown, but still being selected. At request time, the prompt can't be found and then no prompt is used.

This use case should be handled cleaner:
 - The UI should be updated immediately
 - The Edit button should always open the correct file
 - Requests should fallback to their regular prompt

JonasHelming:
Great catch, thank you!

New commit should fix the describe issue:

- We update the selection box on file changes
- We show to the user if a variant is no longer existing and disallow editing it
- If a variant is not available, but configured, an empty system prompt is expected, though

sdirix (changes requested):
There is one more problematic flow: 

- When there are multiple variants, the flow works as expected.
- However when there is no variant left, then the UI disables all buttons, the dropdown vanishes, there is no error message and the user can't recover from the state without re-adding a variant

image


Steps to reproduce:
- Open AI configuration view
- Select workspace agent
- Click "edit", edit the template and use "File > Save As" to save it as a variant
- Select the newly added variant in the UI
- Delete the prompt template again
- Now the UI is in the broken state and does not recover without reloading or readding a variant

Some more thing:
 - Deleting the variant can also be accomplished via the "Reset" button.
 - For these custom variants it would be nice if the button would show "Delete" instead as they are then gone in contrast to the "built-in" variants

Besides this edge case it seems to work great now :+1:

packages/ai-core/src/browser/ai-configuration/template-settings-renderer.tsx
```diff
@@ -34,20 +34,19 @@ export const TemplateRenderer: React.FC<TemplateRendererProps> = ({
     promptService,
     aiSettingsService,
 }) => {
-    const [variantIds, setVariantIds] = React.useState<string[]>([]);
+    const variantIds = [DEFAULT_VARIANT, ...promptService.getVariantIds(template.id)];
     const [selectedVariant, setSelectedVariant] = React.useState<string>(DEFAULT_VARIANT);
 
-    React.useEffect(() => {
+    React.useMemo(() => {
```
sdirix, line 40: `useMemo` is definitely incorrect. What is executed here is an effect, not the calculation of a value. We should for example also never call a `setHook` within `useMemo`

JonasHelming:
Great catch, fixed the describe workflow!
I will create a follow up for "delete".

packages/ai-core/src/browser/ai-configuration/template-settings-renderer.tsx
JonasHelming: If I use effect, the update does not work.

sdirix: For me it works perfectly fine with `useEffect` :confused:

sdirix (approved):
Works great for me. Thanks!

Issue #14137, opened 2024-09-05
Dev containers do not have built-in plugins


sgraband:
### Bug Description:

Currently dev containers are started without the built-in plugins of the "parent" application (e.g. Theia IDE).
This leads to the application behaving differently than outside of the dev container, for example the syntax highlighting is missing.
I do not think, that the creator of the `devcontainer.json` file should be responsible for specifying the built-in plugins, as one would expect that they are "part of the application".

### Steps to Reproduce:

1. Open a dev container
2. Observe that there are no built-in plugins displayed in the extension view
3. Check that plugins are not installed (e.g. settings page, no syntax highlighting, etc..)

### Additional Information

- Operating System: Ubuntu 22.04
- Theia Version: Theia IDE 1.53.100

msujew:
Note that this works in the Theia Dev Electron app:

![image](https://github.com/user-attachments/assets/3b8d7a4a-1122-4d8d-832f-261d8d8b7131)

@sgraband What application have you been using to replicate this? With the latest TheiaIDE, this doesn't work.

sgraband:
Thanks for the fast reply. I replicated it with Theia IDE 1.53.100 as well(also updated the ticket).

I will take a detailed look at the PR you linked in a minute, but would you expect this to work automatically with any product or are there some requirements on the products side to configure this/install the plugins in a certain way?

msujew:
It might be that the `localDir` call [here](https://github.com/eclipse-theia/theia/blob/41f8c72ad14a9afcdde7e17ce5440ec2630e2f8e/packages/plugin-ext/src/main/node/plugin-remote-copy-contribution.ts#L29) behaves somehow unexpected in a packaged environment.

jonah-iden:
@sgraband Could you provide us with some more information on how your plugins setup is? Do you set the --plugins option or the THEIA_DEFAULT_PLUGINS env variable? Are you installing some plugins manually before connecting with the container? Is it a pacakges electron distribution with an asar file?

jonah-iden:
@msujew ok thanks will try it out

sgraband:
@jonah-iden. Yes i was reporting this straight from the Theia IDE. Thanks for looking into it.

PR #14530, opened 2024-11-26
DevContainers: Support THEIA_DEFAULT_PLUGINS

jonah-iden:
#### What it does

Fixes https://github.com/eclipse-theia/theia/issues/14137

Adds support for copying and using plugins defined by local-dir through THEIA_DEFAULT_PLUGINS env variable

#### How to test

remove the "--plugins=local-dir:../../plugins" fro your start target, 
set THEIA_DEFAULT_PLUGINS environment variable to "--plugins=local-dir:../../plugins". 
reopen in any dev container. See that the plugins should be correctly applied

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/node/plugin-remote-cli-contribution.ts
+++ b/packages/plugin-ext/src/main/node/plugin-remote-cli-contribution.ts
@@ -27,10 +27,11 @@ export class PluginRemoteCliContribution implements RemoteCliContribution {
 
     enhanceArgs(context: RemoteCliContext): MaybePromise<string[]> {
         const pluginsFolder = this.pluginCliContribution.localDir();
-        if (!pluginsFolder) {
-            return [];
-        } else {
+        const defaultPlugins = process.env.THEIA_DEFAULT_PLUGINS;
+        if (pluginsFolder || defaultPlugins) {
             return ['--plugins=local-dir:./plugins'];
         }
+        return [];
+
     }
 }
--- a/packages/plugin-ext/src/main/node/plugin-remote-copy-contribution.ts
+++ b/packages/plugin-ext/src/main/node/plugin-remote-copy-contribution.ts
@@ -27,10 +27,13 @@ export class PluginRemoteCopyContribution implements RemoteCopyContribution {
 
     async copy(registry: RemoteCopyRegistry): Promise<void> {
         const localDir = this.pluginCliContribution.localDir();
-        if (localDir) {
-            const fsPath = FileUri.fsPath(localDir);
-            await registry.directory(fsPath, 'plugins');
-        }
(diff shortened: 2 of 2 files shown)


sgraband (commented):
@jonah-iden thanks for looking into this. Could you maybe explain a little bit more in detail on why you went with this approach?
As far as i can tell (but i am not too familiar with the code), we missed the plugin directories provided via THEIA_DEFAULT_PLUGINS before this change, right? Then this addition definately makes sense.
However, when an adoptor only provides the plugins via the cli and not via the env it will still not work, right? Because that was the original issue described in #14137.

jonah-iden:
@sgraband could i have misunderstood the issue? 
I thought the issue was that TheiaIDE did not correctly copy the plugins. That was because TheiaIDE is not using the commandline argument but instead is using the environment variable.
Using the command line variable still works fine and copies the plugins correctly as far as i tested. At least using theia ide with `--plugins=local-dir... ` worked fine as well before this fix

sgraband:
I actually thought that the Theia IDE sets the plugins via the cli param, and that the cli param did not work in a bundled environment. However i quickly checked and you are right. I will create a local build of the Theia IDE with your fix and check if it works :+1: Sorry for the confusion!

sgraband:
I just tested it and i am afraid it still does not work in a packaged environment (local Theia IDE build with your changes).
The built-ins still do not show up correctly in the dev-container.
The main issue is, is that this is not really debuggable in the packaged application (at least there is no way that i know of).

jonah-iden:
Thanks for testing. Do you know if in you test installation everything is bundled together into an asar archive?

sgraband:
That should not be the case: https://github.com/eclipse-theia/theia-ide/blob/d2daf785498829593c46940f86720bb44cc3a284/applications/electron/electron-builder.yml#L10

jonah-iden:
ok, are you sure you build the theiaIde correctly. I just copied my `lib/backend?` folder build from the branch into my theia-ide installation resources/app/lib (renamed the old one). And it works fine for me with the copy
grafik

sgraband:
I made a full local build of the theia ide following this: https://github.com/eclipse-theia/theia-ide/blob/jf/local-theia-build/local-theia-build.md. I can try it out again later on, but i am pretty sure i did not make a mistake.

Also note, that i am on ubuntu, maybe this makes a difference somehow?

jonah-iden:
thanks i'll try to do a local build as well

jonah-iden:
Sadly i can not get the local build to work even in wsl. Probably becuase docker is still running on windows and the dockerfile is build for unix.

sgraband (approved):
I just retested it with packaging the example electron app and everything works as intended. I think this is good to go. Should any issues arise when updating the Theia IDE to the next version i would investigate again on why this happens. I will create a reminder for myself.

Thanks for the improvement :tada:

PR #14533, opened 2024-11-26
Cancel pending hover on mouse exit

### LINKED ISSUE #14532: Hover service renders the hover pop-up too eagerly
### Bug Description:

As soon as the mouse pointer enters some element that requests a hover pop-up, the `HoverService` sets a timer that _will_ fire and show the pop-up, regardless of what the pointer does next. Even passing quickly over some element on the way to another UI element that the user wants to interact with will cause the hover to appear and then, because the pointer is already elsewhere in the application, disappear again quickly. The resulting flash of UI is distracting.

### Steps to Reproduce:

1. Pass the mouse pointer into one of the items in the status bar and then quickly away to something else in the workbench.
2. See the status bar hover pop up despite that it is now well removed from the scene of mouse pointer interactions.

A screen recording is attached, below.

### Additional Information

- Operating System: macOS Sequoia 15.1.1
- Theia Version: 1.55.0, tip of main branch as of 26 Nov 2024

CleanShot 2024-11-26 at 11 17 11

cdamus:
#### What it does

If the mouse exits before the HoverService has rendered its pop-up, cancel the pending presentation of that pop-up.

Fixes #14532.

#### How to test

Pass the mouse pointer quickly over items in the status bar that provide hover pop-ups. See that those pop-ups do not appear when the mouse pointer has already gone elsewhere. Verify that the pop-up behaves as before when the mouse pointer lingers on the target element. Both cases as captured here in a screen recording:

CleanShot 2024-11-26 at 11 32 32

#### Follow-ups

None.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/hover-service.ts
+++ b/packages/core/src/browser/hover-service.ts
@@ -25,6 +25,10 @@ import '../../src/browser/style/hover-service.css';
 
 export type HoverPosition = 'left' | 'right' | 'top' | 'bottom';
 
+// Threshold, in milliseconds, over which a mouse movement is not considered
+// quick enough as to be ignored
+const quickMouseThresholdMillis = 200;
+
 export namespace HoverPosition {
     export function invertIfNecessary(position: HoverPosition, target: DOMRect, host: DOMRect, totalWidth: number, totalHeight: number): HoverPosition {
         if (position === 'left') {
@@ -100,11 +104,13 @@ export class HoverService {
         if (request.target !== this.hoverTarget) {
             this.cancelHover();
             this.pendingTimeout = disposableTimeout(() => this.renderHover(request), this.getHoverDelay());
+            this.hoverTarget = request.target;
+            this.listenForMouseOut();
         }
     }
 
     protected getHoverDelay(): number {
-        return Date.now() - this.lastHidHover < 200
+        return Date.now() - this.lastHidHover < quickMouseThresholdMillis
             ? 0
             : this.preferences.get('workbench.hover.delay', isOSX ? 1500 : 500);
     }
@@ -116,7 +122,6 @@ export class HoverService {
         if (cssClasses) {
             host.classList.add(...cssClasses);
         }
-        this.hoverTarget = target;
         if (content instanceof HTMLElement) {
             host.appendChild(content);
             firstChild = content;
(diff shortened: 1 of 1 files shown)


sdirix (approved):
I like this change!

I found one edge/case issue: When switching fast and often between two hoverable elements, I sometimes can reproduce that no more tooltip appears although I'm hovering an element. Once the mouse leaves and enters again the tooltip will reappear.

tooltips

The behavior is better than the status quo, so I would be fine with a merge.

sdirix (changes requested):
Actually I noticed one more issue:

Previously it was possible to be quick enough to enter the tooltip with the mouse to copy its contents. This is no longer possible.

- Is this a worthwhile feature?
- Do we want to keep it?

In this Gif we can see the current state of `master` with the
- weird delay of tooltips showing although the element is no longer hovered
- the ability to enter and copy the tooltip content

EnterToolTips

sdirix:
@tsmaeder any preference on your side for today's release and/or followups?

tsmaeder:
I don't think this is a critical problem, so let's not force it in if it's not ready.

cdamus:
> I don't think this is a critical problem, so let's not force it in if it's not ready.

I'm certainly not in such a hurry, so I wouldn't ask for it to make this release 😀 

> Previously it was possible to be quick enough to enter the tooltip with the mouse to copy its contents. This is no longer possible.
> 
> * Is this a worthwhile feature?
> * Do we want to keep it?

This is something that I actually do quite a lot in other tools (some don't support selection in tool-tips, others do). I think I should be able to preserve this capability with a judicious sprinkling of debounces or similar. I'll give it a go.

cdamus:
Commit ecf3c2d117885c03dda31ba8d4da9ead2f071ea7 restores mouse pointer access to the hover pop-up itself and also fixes other glitches.

cdamus:
@sdirix would you like to take another look?

cdamus:
Commit 56cfe5ddcc93a8b39cb0bcdbabbbbac99cb45c83 takes a much simpler approach to fix both the problem of quick mouse fly-overs popping up the hover widget and also the difficulty of mousing over to the hover pop-up once it's shown.

msujew (approved):
Looks pretty good to me 👍 I can confirm that this fixes the issue without introducing side-effects.

sdirix (approved):
Very nice! The code is simple and this is a great improvement for the usability in Theia. Great work :+1:

PR #14535, opened 2024-11-26
Add support for users to specify custom request settings, model and optionally provider specific

JonasHelming:
#### What it does

Add support for users to specify custom request settings, model and optionally provider specific.
The reason for making them provider specific is that providers have different options and sometimes even different names for the same (see below)

#### How to test

Add the settings below and adapt it to models you have.
Qwen/Qwen2.5-Coder-32B-Instruct is always "warm" on serverless Hugginface
Starcoder2.3B can be downloaded here: https://huggingface.co/Mozilla/starcoder2-llamafile/tree/main
gemma2 can be directly downloaded with ollama (ollama serve and ollama run gemma2)

Two good test cases:
- set the max_length to something small and observe the model to stop
- set a stop token and ask the model to say it :-)

```
"ai-features.modelSettings.requestSettings": [
        {
            "modelId": "Qwen/Qwen2.5-Coder-32B-Instruct",
            "requestSettings": {
                "max_new_tokens": 2048,
                "stop": [
                    "<|im_end|>",
                ]
            },
            "providerId": "huggingface"
        },
        {
            "modelId": "gemma2",
            "requestSettings": {
                "num_predict": 1024,
                "stop": [
                    "<|endoftext|>"
                ],
            },
            "providerId": "ollama"
        },
        {
            "modelId": "StarCoder2.3B",
            "requestSettings": {
                "n_predict": 200,
                "stream": true,
                "stop": [
                    "<file_sep>",
                    "<|endoftext|>"
                ],
                "cache_prompt": true,
            },
            "providerId": "llamafile"
        },
        {
            "modelId": "gpt-4o-2024-05-13",
            "requestSettings": {
                "max_tokens": 10,
                "stop": [
                    "<|im_end|>"
                ]
            },
            "providerId": "openai"
        },
    ]
```

#### Follow-ups

This should be the last thing we add to the provider layer before we refactor it all together:
- Make the settings more concistent
- Remove a lot of code duplication

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-llamafile/src/browser/llamafile-frontend-application-contribution.ts
+++ b/packages/ai-llamafile/src/browser/llamafile-frontend-application-contribution.ts
@@ -18,7 +18,9 @@ import { FrontendApplicationContribution, PreferenceService } from '@theia/core/
 import { inject, injectable } from '@theia/core/shared/inversify';
 import { LlamafileEntry, LlamafileManager } from '../common/llamafile-manager';
 import { PREFERENCE_LLAMAFILE } from './llamafile-preferences';
+import { PREFERENCE_NAME_REQUEST_SETTINGS, RequestSetting } from '@theia/ai-core/lib/browser/ai-core-preferences';
 
+const LLAMAFILE_ID = 'llamafile';
 @injectable()
 export class LlamafileFrontendApplicationContribution implements FrontendApplicationContribution {
 
@@ -33,27 +35,63 @@ export class LlamafileFrontendApplicationContribution implements FrontendApplica
     onStart(): void {
         this.preferenceService.ready.then(() => {
             const llamafiles = this.preferenceService.get<LlamafileEntry[]>(PREFERENCE_LLAMAFILE, []);
-            this.llamafileManager.addLanguageModels(llamafiles);
-            llamafiles.forEach(model => this._knownLlamaFiles.set(model.name, model));
+            const requestSettings = this.preferenceService.get<RequestSetting[]>(PREFERENCE_NAME_REQUEST_SETTINGS, []);
+            const modelsWithSettings = llamafiles.map(model => this.applyRequestSettingsToLlamaFile(model, requestSettings));
+
+            this.llamafileManager.addLanguageModels(modelsWithSettings);
+            modelsWithSettings.forEach(model => this._knownLlamaFiles.set(model.name, model));
 
             this.preferenceService.onPreferenceChanged(event => {
                 if (event.preferenceName === PREFERENCE_LLAMAFILE) {
-                    // only new models which are actual LLamaFileEntries
-                    const newModels = event.newValue.filter((llamafileEntry: unknown) => LlamafileEntry.is(llamafileEntry)) as LlamafileEntry[];
-
-                    const llamafilesToAdd = newModels.filter(llamafile =>
-                        !this._knownLlamaFiles.has(llamafile.name) || !LlamafileEntry.equals(this._knownLlamaFiles.get(llamafile.name)!, llamafile));
-
-                    const llamafileIdsToRemove = [...this._knownLlamaFiles.values()].filter(llamafile =>
-                        !newModels.find(a => LlamafileEntry.equals(a, llamafile))).map(a => a.name);
-
-                    this.llamafileManager.removeLanguageModels(llamafileIdsToRemove);
-                    llamafileIdsToRemove.forEach(model => this._knownLlamaFiles.delete(model));
-
-                    this.llamafileManager.addLanguageModels(llamafilesToAdd);
-                    llamafilesToAdd.forEach(model => this._knownLlamaFiles.set(model.name, model));
+                    this.handleLlamaFilePreferenceChange(event.newValue as LlamafileEntry[]);
+                } else if (event.preferenceName === PREFERENCE_NAME_REQUEST_SETTINGS) {
+                    this.handleRequestSettingsChange(event.newValue as RequestSetting[]);
                 }
             });
         });
+    }
+
+    protected handleLlamaFilePreferenceChange(newModels: LlamafileEntry[]): void {
+        const requestSettings = this.preferenceService.get<RequestSetting[]>(PREFERENCE_NAME_REQUEST_SETTINGS, []);
+
+        // Models to add or update
+        const llamafilesToAdd = newModels.filter(llamafile =>
+            !this._knownLlamaFiles.has(llamafile.name) ||
+            !LlamafileEntry.equals(this._knownLlamaFiles.get(llamafile.name)!, llamafile));
+
+        // Models to remove
+        const llamafileIdsToRemove = [...this._knownLlamaFiles.values()].filter(llamafile =>
+            !newModels.find(newModel => LlamafileEntry.equals(newModel, llamafile)))
+            .map(llamafile => llamafile.name);
+
+        // Update the manager
+        this.llamafileManager.removeLanguageModels(llamafileIdsToRemove);
+        llamafileIdsToRemove.forEach(id => this._knownLlamaFiles.delete(id));
+
+        const modelsWithSettings = llamafilesToAdd.map(model => this.applyRequestSettingsToLlamaFile(model, requestSettings));
+        this.llamafileManager.addLanguageModels(modelsWithSettings);
+        modelsWithSettings.forEach(model => this._knownLlamaFiles.set(model.name, model));
+    }
+
+    protected handleRequestSettingsChange(newSettings: RequestSetting[]): void {
+        const llamafiles = this.preferenceService.get<LlamafileEntry[]>(PREFERENCE_LLAMAFILE, []);
+        const modelsWithSettings = llamafiles.map(model => this.applyRequestSettingsToLlamaFile(model, newSettings));
+
+        this.llamafileManager.addLanguageModels(modelsWithSettings);
+        modelsWithSettings.forEach(model => this._knownLlamaFiles.set(model.name, model));
+    }
+
+    protected applyRequestSettingsToLlamaFile(model: LlamafileEntry, requestSettings: RequestSetting[]): LlamafileEntry {
+        const matchingSettings = requestSettings.filter(
+            setting => (!setting.providerId || setting.providerId === LLAMAFILE_ID) && setting.modelId === model.name
+        );
+        if (matchingSettings.length > 1) {
+            console.warn(`Multiple entries found for model "${model.name}". Using the first match.`);
+        }
+
+        return {
+            ...model,
+            defaultRequestSettings: matchingSettings[0]?.requestSettings
+        };
     }
 }
--- a/packages/ai-llamafile/src/browser/llamafile-frontend-application-contribution.ts
+++ b/packages/ai-llamafile/src/browser/llamafile-frontend-application-contribution.ts
@@ -20,7 +20,7 @@ import { LlamafileEntry, LlamafileManager } from '../common/llamafile-manager';
 import { PREFERENCE_LLAMAFILE } from './llamafile-preferences';
 import { PREFERENCE_NAME_REQUEST_SETTINGS, RequestSetting } from '@theia/ai-core/lib/browser/ai-core-preferences';
 
-const LLAMAFILE_ID = 'llamafile';
+const LLAMAFILE_PROVIDER_ID = 'llamafile';
 @injectable()
 export class LlamafileFrontendApplicationContribution implements FrontendApplicationContribution {
 
@@ -83,7 +83,7 @@ export class LlamafileFrontendApplicationContribution implements FrontendApplica
 
     protected applyRequestSettingsToLlamaFile(model: LlamafileEntry, requestSettings: RequestSetting[]): LlamafileEntry {
         const matchingSettings = requestSettings.filter(
-            setting => (!setting.providerId || setting.providerId === LLAMAFILE_ID) && setting.modelId === model.name
+            setting => (!setting.providerId || setting.providerId === LLAMAFILE_PROVIDER_ID) && setting.modelId === model.name
         );
         if (matchingSettings.length > 1) {
             console.warn(`Multiple entries found for model "${model.name}". Using the first match.`);
--- a/packages/ai-llamafile/src/browser/llamafile-frontend-application-contribution.ts
+++ b/packages/ai-llamafile/src/browser/llamafile-frontend-application-contribution.ts
@@ -16,7 +16,7 @@
 
 import { FrontendApplicationContribution, PreferenceService } from '@theia/core/lib/browser';
 import { inject, injectable } from '@theia/core/shared/inversify';
-import { LlamafileEntry, LlamafileManager } from '../common/llamafile-manager';
+import { LlamafileManager, LlamafileModelDescription } from '../common/llamafile-manager';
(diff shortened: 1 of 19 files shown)


JonasHelming:
@dhuebner Could you check the Ollama adaptations please?

JonasHelming:
See this for docu: https://github.com/eclipse-theia/theia-website/pull/662

dhuebner:
@JonasHelming 
Sure, will do.

dhuebner:
@JonasHelming 
Ollama works as expected!

I have a question to another topic. Would it be possible to mark a `LanguageModelRequest` with an explicit flag that tells if it's a chat or a completion request? Having this information model implementors could decide if they should use stream or text response.

packages/ai-ollama/src/node/ollama-language-model.ts
```diff
@@ -37,30 +37,41 @@ export class OllamaModel implements LanguageModel {
...
         const ollama = this.initializeOllama();
 
         if (request.response_format?.type === 'json_schema') {
             return this.handleStructuredOutputRequest(ollama, request);
         }
         const response = await ollama.chat({
-            ...this.DEFAULT_REQUEST_SETTINGS,
             model: this.model,
+            ...this.DEFAULT_REQUEST_SETTINGS,
+            ...settings,
```
sdirix, line 71: This is now a shallow merge of options and therefore behaves differently than the other language models, e.g. huggingface. I think we should either always shallow merge or always exclusively use the handed over ones or default ones.

packages/ai-ollama/src/node/ollama-language-model.ts
```diff
@@ -77,12 +88,14 @@ export class OllamaModel implements LanguageModel {
     }
 
     protected async handleStructuredOutputRequest(ollama: Ollama, request: LanguageModelRequest): Promise<LanguageModelParsedResponse> {
+        const settings = this.getSettings(request);
```
sdirix, line 91: This merges the other way around, the speficic request settings will be overwritten by the default settings

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
```diff
@@ -38,72 +40,119 @@ export class OpenAiFrontendApplicationContribution implements FrontendApplicatio
             this.manager.setApiKey(apiKey);
 
             const models = this.preferenceService.get<string[]>(MODELS_PREF, []);
-            this.manager.createOrUpdateLanguageModels(...models.map(createOpenAIModelDescription));
+            const requestSettings = this.preferenceService.get<RequestSetting[]>(PREFERENCE_NAME_REQUEST_SETTINGS, []);
+            this.manager.createOrUpdateLanguageModels(...models.map(modelId => this.createOpenAIModelDescription(modelId, requestSettings)));
             this.prevModels = [...models];
 
             const customModels = this.preferenceService.get<Partial<OpenAiModelDescription>[]>(CUSTOM_ENDPOINTS_PREF, []);
```
sdirix, line 47: This type is no longer correct as it was extended by the `defaultRequestSettings` which are in fact not stored in this preference.

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
```diff
@@ -38,72 +40,119 @@ export class OpenAiFrontendApplicationContribution implements FrontendApplicatio
...
-                        !oldModels.some(model =>
-                            model.id === newModel.id &&
-                            model.model === newModel.model &&
-                            model.url === newModel.url &&
-                            model.apiKey === newModel.apiKey &&
-                            model.enableStreaming === newModel.enableStreaming));
-
-                    this.manager.removeLanguageModels(...modelsToRemove.map(model => model.id));
-                    this.manager.createOrUpdateLanguageModels(...modelsToAddOrUpdate);
+                    this.handleCustomModelChanges(event.newValue as Partial<OpenAiModelDescription>[]);
```
sdirix, line 57: Same issue with the type here. It's no longer correct as it was extended by the `defaultRequestSettings` which are not stored in this preference. We should either adapt the preference schema or the type.

JonasHelming:
> @sdirix
> I have a question to another topic. Would it be possible to mark a `LanguageModelRequest` with an explicit flag that tells if it's a chat or a completion request? Having this information model implementors could decide if they should use stream or text response.

@dhuebner  I also already thought about this, would you mind creating a new ticket and mark me there?

packages/ai-ollama/src/node/ollama-language-model.ts
```diff
@@ -77,12 +88,14 @@ export class OllamaModel implements LanguageModel {
...
         const result = await ollama.chat({
+            ...settings,
             ...this.DEFAULT_REQUEST_SETTINGS,
             model: this.model,
             messages: request.messages.map(this.toOllamaMessage),
             format: 'json',
-            ...request.settings
+            stream: false,
```
sdirix, line 98: Do we never stream on purpose here?

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
```diff
@@ -18,6 +18,9 @@ import { FrontendApplicationContribution, PreferenceService } from '@theia/core/
 import { inject, injectable } from '@theia/core/shared/inversify';
 import { OpenAiLanguageModelsManager, OpenAiModelDescription } from '../common';
 import { API_KEY_PREF, CUSTOM_ENDPOINTS_PREF, MODELS_PREF } from './openai-preferences';
+import { PREFERENCE_NAME_REQUEST_SETTINGS, RequestSetting } from '@theia/ai-core/lib/browser/ai-core-preferences';
+
+const OPENAI_ID = 'openai';
```
sdirix, line 23:
```suggestion
export const OPENAI_PROVIDER_ID = 'openai';
```

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
```diff
@@ -38,72 +40,119 @@ export class OpenAiFrontendApplicationContribution implements FrontendApplicatio
...
+        const newModels = this.createCustomModelDescriptionsFromPreferences(newCustomModels);
+
+        const modelsToRemove = oldModels.filter(model => !newModels.some(newModel => newModel.id === model.id));
+        const modelsToAddOrUpdate = newModels.filter(newModel =>
+            !oldModels.some(model =>
+                model.id === newModel.id &&
+                model.model === newModel.model &&
+                model.url === newModel.url &&
+                model.apiKey === newModel.apiKey &&
+                model.defaultRequestSettings === newModel.defaultRequestSettings &&
```
sdirix, line 89: As the request settings are of type object, this will likely always be `false` in case they are not `undefined` in the open ai preference.

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
```diff
@@ -38,72 +40,119 @@ export class OpenAiFrontendApplicationContribution implements FrontendApplicatio
...
+                model.enableStreaming === newModel.enableStreaming));
+
+        this.manager.removeLanguageModels(...modelsToRemove.map(model => model.id));
+        this.manager.createOrUpdateLanguageModels(...modelsToAddOrUpdate);
```
sdirix, line 93: At this point of time the `requestSettings` were never mixed into the `modelsToAddOrUpdate` and will therefore be lost.

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
```diff
@@ -38,72 +40,119 @@ export class OpenAiFrontendApplicationContribution implements FrontendApplicatio
...
+
+            return [
+                ...acc,
+                {
+                    id: pref.id && typeof pref.id === 'string' ? pref.id : pref.model,
+                    model: pref.model,
+                    url: pref.url,
+                    apiKey: typeof pref.apiKey === 'string' || pref.apiKey === true ? pref.apiKey : undefined,
+                    enableStreaming: pref.enableStreaming ?? true,
+                    defaultRequestSettings: modelRequestSetting?.requestSettings ?? pref.defaultRequestSettings
```
sdirix, line 136: With this we now allow the defaultRequestSettings to also be defined within the custom open ai preferences. Is this intentional? If yes, then we also should add it to the preferences schema and documentation. Note that in the Llamafile preferences we do not support that and forcefully ignore such preferences in case the user did define them.

packages/ai-openai/src/node/openai-language-models-manager-impl.ts
```diff
@@ -45,23 +45,27 @@ export class OpenAiLanguageModelsManagerImpl implements OpenAiLanguageModelsMana
...
-                    console.info(`Open AI: skip creating model ${modelDescription.id} because it already exists`);
+                    console.warn(`OpenAI: model ${modelDescription.id} is not an OpenAI model`);
                     continue;
                 }
-                model.url = modelDescription.url;
                 model.model = modelDescription.model;
                 model.enableStreaming = modelDescription.enableStreaming;
+                model.url = modelDescription.url;
                 model.apiKey = apiKeyProvider;
+                model.defaultRequestSettings = modelDescription.defaultRequestSettings; // Update default settings
```
sdirix, line 58:
```suggestion
                model.defaultRequestSettings = modelDescription.defaultRequestSettings;
```

sdirix (changes requested):
Thanks for the work :heart: ! I found some inconsistencies which should be fixed before we merge.

packages/ai-ollama/src/node/ollama-language-model.ts
dhuebner: @sdirix 
Ollama setting are inside the `options` property and the default don't provide them.

JonasHelming: @dhuebner I agree @sdirix If you also agree, please resolve the conversation.

packages/ai-ollama/src/node/ollama-language-model.ts
JonasHelming: That was the behavior before. We have add this explicitly, otherwise we have a type error because the function of the ollama API is overloaded. It does make sense to not stream structured output.

packages/ai-ollama/src/node/ollama-language-model.ts
sdirix: You're right, I overlooked that part. Minor: I would like to suggest to improve the naming here and for example set `options` here directly instead of hiding that behind a similarly named `settings`.

packages/ai-ollama/src/node/ollama-language-model.ts
sdirix: If you like you can comment in the code why this property was set

JonasHelming: @dhuebner Can you confirm that this is intended?

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
JonasHelming: I remove the ability

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
JonasHelming: I will remove this check, as we do not allow to specify them directly in custom models anymore.

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
JonasHelming: I don't get this comment, they should be added in createCustomModelDescriptionsFromPreferences?

packages/ai-llamafile/src/common/llamafile-manager.ts
```diff
@@ -34,17 +34,34 @@ export interface LlamafileEntry {
...
 }
 
 export namespace LlamafileEntry {
     export function equals(a: LlamafileEntry, b: LlamafileEntry): boolean {
-        return a.name === b.name && a.uri === b.uri && a.port === b.port;
+        return (
+            a.name === b.name &&
+            a.uri === b.uri &&
+            a.port === b.port &&
+            JSON.stringify(a.defaultRequestSettings) === JSON.stringify(b.defaultRequestSettings)
```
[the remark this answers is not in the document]
JonasHelming: I think this will be removed then anyway, as it should not be part of the settings entry (see below)

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
sdirix: The `createCustomModelDescriptionsFromPreferences` is called without the requestSettings handed over, therefore they are not retrieved and matched

JonasHelming: see line 126, but I will test again

sdirix: In line 126 the `requestSettings` handed over to `createCustomModelDescriptionsFromPreferences` is used. However in that method `requestSettings` is an optional parameter with a default value of an empty array. In line 79 and 80, we call `createCustomModelDescriptionsFromPreferences` without handing over the `requestSettings` parameter. Therefore `requestSettings` will be empty and all models will be updated with an `undefined` `defaultRequestSettings`

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
```diff
@@ -38,72 +40,118 @@ export class OpenAiFrontendApplicationContribution implements FrontendApplicatio
...
+            model: modelId,
+            apiKey: true,
+            enableStreaming: !openAIModelsWithDisabledStreaming.includes(modelId),
+            defaultRequestSettings: modelRequestSetting?.requestSettings
+        };
+    }
+
+    protected createCustomModelDescriptionsFromPreferences(
+        preferences: Partial<OpenAiModelDescription>[],
+        requestSettings: RequestSetting[] = []
```
sdirix, line 118:
```suggestion
        requestSettings: RequestSetting[]
```

packages/ai-llamafile/src/browser/llamafile-frontend-application-contribution.ts
```diff
@@ -33,27 +35,63 @@ export class LlamafileFrontendApplicationContribution implements FrontendApplica
     onStart(): void {
         this.preferenceService.ready.then(() => {
             const llamafiles = this.preferenceService.get<LlamafileEntry[]>(PREFERENCE_LLAMAFILE, []);
```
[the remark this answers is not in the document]
JonasHelming: Introduced LlamafileModelDescriptors

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
JonasHelming: Fixed

JonasHelming:
Thank you for the great review. I tried to address all comments (in individual commits) and tested all providers again with the final state.

packages/ai-ollama/src/node/ollama-language-model.ts
dhuebner: @JonasHelming 
I never received structured requests when using Theia AI so I can't really check if it works or not. But I think streams are useful mostly in chat context.

sdirix (approved):
Works for me. I tested with all model integrations we offer.

Thanks for the great work :heart:

PR #14537, opened 2024-11-26
Show progress while calculating AI code completion

### LINKED ISSUE #14536: [Theia AI] Show progress while calculating the code completion

JonasHelming:
fixed #14536

#### What it does

Shows progress in the status bar while AI code completion is running.

#### How to test

See the progress while completing code. If you cancel via moving the cursor in the code editor, progress bar should also disapear.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-code-completion/src/common/code-completion-agent.ts
+++ b/packages/ai-code-completion/src/common/code-completion-agent.ts
@@ -18,7 +18,7 @@ import {
     Agent, AgentSpecificVariables, CommunicationRecordingService, getTextOfResponse,
     LanguageModelRegistry, LanguageModelRequest, LanguageModelRequirement, PromptService, PromptTemplate
 } from '@theia/ai-core/lib/common';
-import { generateUuid, ILogger } from '@theia/core';
+import { generateUuid, ILogger, ProgressService } from '@theia/core';
 import { inject, injectable, named } from '@theia/core/shared/inversify';
 import * as monaco from '@theia/monaco-editor-core';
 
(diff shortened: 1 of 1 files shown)


packages/ai-code-completion/src/common/code-completion-agent.ts
```diff
@@ -36,82 +36,89 @@ export class CodeCompletionAgentImpl implements CodeCompletionAgent {
         context: monaco.languages.InlineCompletionContext,
         token: monaco.CancellationToken
     ): Promise<monaco.languages.InlineCompletions | undefined> {
-        const languageModel =
-            await this.languageModelRegistry.selectLanguageModel({
-                agent: this.id,
-                ...this.languageModelRequirements[0],
+        const progress = await this.progressService.showProgress(
```
martin-fleck-at, line 39: I was wondering why we are not using the `withProgress` but just noticed that this method always provides something cancellable and I guess this is not something we want in this case.

I'm also wondering whether it makes sense to report progress on the different steps (create prompt, send request, formulate response). However, while adding transparency, it might be too fine-grained.

packages/ai-code-completion/src/common/code-completion-agent.ts
```diff
@@ -36,82 +36,89 @@ export class CodeCompletionAgentImpl implements CodeCompletionAgent {
...
+            { text: 'Calculating AI code completion...', options: { location: 'window' } }
+        );
+        try {
+            const languageModel =
```
martin-fleck-at, line 43: nit: For the sake of easier extensibility, I wonder whether it makes sense to split some of the steps and the whole process out into a separate method that can then be overwritten. However, this is definitely not something that needs to be part of this PR if considered at all.

martin-fleck-at (approved):
Change looks good to me and works as advertised, thank you very much Jonas! I do have a few comments but nothing important that should stop this PR from being merged.

packages/ai-code-completion/src/common/code-completion-agent.ts
JonasHelming: Everything except the LLM request is <1ms, you will never see it atm

packages/ai-code-completion/src/common/code-completion-agent.ts
JonasHelming: We should do this, if we adapt the code completion to more LLMs.

PR #14539, opened 2024-11-27
Allow specifying max lines used for AI code completion context

JonasHelming:
Review https://github.com/eclipse-theia/theia/pull/14537 first, please

#### What it does

Adds a new user setting allowing to specify max lines used for AI code completion context

#### How to test

- Change the setting to other values than 0
- Open the AI History View => Code Completion
- Trigger code completion (SHIFT+SPACE) and check the context

#### Follow-ups

This could be more sophisticated, e.g. with percentages for prefix and suffix or by respecting the AST.
#### Review checklist

#### Reminder for reviewers


packages/ai-code-completion/src/browser/ai-code-completion-preference.ts
```diff
@@ -39,6 +40,14 @@ export const AICodeCompletionPreferencesSchema: PreferenceSchema = {
                 type: 'string'
             },
             default: []
+        },
+        [PREF_AI_INLINE_COMPLETION_MAX_CONTEXT_LINES]: {
+            title: 'Maximum Context Lines',
+            type: 'number',
+            description: 'The maximum number of lines shared between the prefix and suffix for AI code completion.\
```
martin-fleck-at, line 47: Are the terms `prefix` and `suffix` something the user can understand? These are agent-specific variables that may not be obvious to the user so I wonder whether we shouldn't replace that with something like:
`The maximum number of lines distributed among the lines before and after the cursor position.`

packages/ai-code-completion/src/browser/ai-code-completion-preference.ts
```diff
@@ -39,6 +40,14 @@ export const AICodeCompletionPreferencesSchema: PreferenceSchema = {
...
+            Set this to 0 to use the full file as context without any line limit.',
```
martin-fleck-at, line 48: This makes me a bit sad, because now there is no way to actually just have a single line, i.e., if I set 1 I get the cursor line + one above. I think it may make more sense to make `-1` the "whole document" setting and `0` the "no prefix/suffix" setting, what do you think?

packages/ai-code-completion/src/browser/ai-code-completion-preference.ts
```diff
@@ -39,6 +40,14 @@ export const AICodeCompletionPreferencesSchema: PreferenceSchema = {
                 type: 'string'
             },
             default: []
+        },
+        [PREF_AI_INLINE_COMPLETION_MAX_CONTEXT_LINES]: {
+            title: 'Maximum Context Lines',
```
martin-fleck-at, line 45: Just thinking out loud: Is that a sensible setting or would it also make sense to have two settings: Number of lines above the cursor position AND number of lines below the cursor position. That way, a user would have much more fine-grained control instead of finding a number that is then evenly distributed with +1 on the "above side" on an odd number.

martin-fleck-at (commented):
The change looks good to me and works as expected. Thank you very much, Jonas! However, I do have some questions regarding the setting as a whole that I'd like to clarify before we merge this.

packages/ai-code-completion/src/browser/ai-code-completion-preference.ts
JonasHelming: We could, but I found a more common thing is maxLines and then be to add a percentage for the prefix as a second settings (which we can do as a follow up)

packages/ai-code-completion/src/browser/ai-code-completion-preference.ts
JonasHelming: Greta suggestion, applied! thank you!

packages/ai-code-completion/src/browser/ai-code-completion-preference.ts
JonasHelming: Great point, adapted description!

PR #14548, opened 2024-11-28
Remote: still use settings and configuration from the local user dir

### LINKED ISSUE #14291: User settings not applied in dev containers
### Bug Description:
In development containers, user settings are not applied, and only the default settings are used. This can lead to an altered user experience, potentially without the user noticing, as their custom configurations are not respected in the container environment.

### Steps to Reproduce:

1. Open a development container.
2. Check the user settings.
3. Observe that only the default settings are applied instead of the user's configured settings.

### Additional Information

- Theia Version: 1.54.0

jonah-iden:
#### What it does

This makes it so that when connected to a remote system or dev-container, configuration from the user dir like tasks, settings, ai-prompts, etc. are still used instead of the configurations located on the remote system/container.

Fixes #14291

#### How to test

Connect to a remote system and see you theme, settings etc are still all set to what they were before connecting.

#### Review checklist

#### Reminder for reviewers


packages/remote/src/electron-browser/remote-frontend-module.ts
```diff
@@ -65,4 +70,19 @@ export default new ContainerModule((bind, _, __, rebind) => {
...
+            ServiceConnectionProvider.createLocalProxy<EnvVariablesServer>(ctx.container, envVariablesPath) :
+            ctx.container.get(EnvVariablesServer));
+    bind(LocalRemoteFileSystemProvider).toSelf().inSingletonScope();
+    rebind(UserStorageContribution).to(RemoteUserStorageContribution);
+
 });
+
+function isRemoteConnection(): boolean {
+    return !!new URLSearchParams(window.location.search).get('localPort');
+}
```
sgraband, line 88:
```suggestion
export function isRemoteConnection(): boolean {
    return !!new URLSearchParams(window.location.search).get('localPort');
}
```

Should we maybe export this, to make it easier for adoptors to rebind?
Also is there a way to fetch the 'localPort' string from somewhere and not hard code it here?

packages/remote/src/electron-browser/remote-frontend-module.ts
```diff
@@ -65,4 +70,19 @@ export default new ContainerModule((bind, _, __, rebind) => {
...
+    bind(LocalRemoteFileSytemServer).toDynamicValue(ctx =>
+        isRemoteConnection() ?
+            ServiceConnectionProvider.createLocalProxy(ctx.container, remoteFileSystemPath, new RemoteFileSystemProxyFactory()) :
+            ctx.container.get(RemoteFileSystemServer));
+    bind(LocalEnvVariablesServer).toDynamicValue(ctx =>
+        isRemoteConnection() ?
+            ServiceConnectionProvider.createLocalProxy<EnvVariablesServer>(ctx.container, envVariablesPath) :
+            ctx.container.get(EnvVariablesServer));
+    bind(LocalRemoteFileSystemProvider).toSelf().inSingletonScope();
+    rebind(UserStorageContribution).to(RemoteUserStorageContribution);
```
sgraband, line 82: It's not really a breaking change, but if adoptors already rebound the `UserStorageContribution` somewhere they will probably not notice this change and not consume this fix. Do you have an idea on how we could document this? Or do you think this is uneccessary?

sgraband (commented):
Thank you very much for the fix. Works really well and my settings are now correctly applied. It does however not fix the issue with the terminal on my end (but this should not be a blocker for this).

Also i added 2 minor comments, where i would like to have your input.

packages/remote/src/electron-browser/remote-frontend-module.ts
jonah-iden: is changed it to use the isRemote flag from the browser.ts. Also unified all calls to get the "port" and "localPort" url search query params

jonah-iden:
are you sure this does not work with the terminal. Adding and removing this in the settings works absolutely fine for me
```
    "terminal.integrated.defaultProfile.linux": "bash",
    "terminal.integrated.profiles.linux": {
        "bash": {
            "path": "/bin/bash"
        },
    },
```
Or just switching the defualtProfile.linux between bash and SHELL works fine as well

sgraband (approved):
> are you sure this does not work with the terminal. Adding and removing this in the settings works absolutely fine for me

With the options you provided it works, however i am talking about the case that the user does not have this setting set (as i do for example). In VSCode i then still get a `bash` terminal by default in the devcontainer. In theia it is just a plain shell. Is there a default value that we could set differently?

However, if not a really quick fix, this is not part of this ticket anymore. For the promised functionality i tested it and everything works great! Thanks a lot :tada:

jfaltermeier:
Just an FYI, this doesn’t seem to work on the Theia IDE Preview for 1.57 at the moment. We haven’t investigated yet: https://github.com/eclipse-theia/theia/discussions/14643#discussioncomment-11602921

PR #14553, opened 2024-11-28
Updating publishing guide and set markdown formatter

sgraband:
#### What it does

This PR updates the publishing guide to reflect the current state and my learnings from the last releases.
When working with the publishing guide (and the changelog for that matter) i noticed that saving always produced changes on my end.
Therefore i went ahead and added a markdownlinter extension and set it as the formatter (and formatted all files).
If this is unwanted, or we want another solution, let me know, i will gladly look into it.
However, with the current state i am not quite sure, why we do not get a lot of formatting changes from different setups that developers have with their markdown formatter configuration.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


.vscode/settings.json
```diff
@@ -51,6 +51,9 @@
     "editor.tabSize": 2,
     "editor.defaultFormatter": "vscode.json-language-features",
   },
+  "[markdown]": {
```
jfaltermeier, line 54: We also have this file checked in: https://github.com/eclipse-theia/theia/blob/master/.theia/settings.json.
I’m using the Theia IDE and was asked which Markdown formatter to use. When I selected the DavidAnson one, it was written to .theia/settings.
Should we update both this file and the Theia settings, or check if we can unify them to use only the VS Code one?

jfaltermeier:
I can reproduce that running `yarn` causes changes in core/README.md and electron/README.md

doc/Publishing.md
```diff
@@ -1,152 +1,215 @@
...
+  > The Eclipse Theia v{version} release is happening today, starting around {time} CET.
+  > If you have any nearly-complete PRs that need to be included, let us know in this thread before we begin.
+  >
+  > We’ll post updates when the release starts and once it’s done.
+  > Please avoid merging any PRs during the release until we confirm it's over.
 
-The localization (`nls`) updates should be performed before a release ([example](https://github.com/eclipse-theia/theia/pull/12665)).
-To help we have an [automatic translation](https://github.com/eclipse-theia/theia/actions/workflows/translation.yml) workflow which can be triggered.
-Note that due to the required CI check (`lint`) we will need for force-push the branch that the bot creates to properly trigger CI.
+- Refer to [this example](https://github.com/eclipse-theia/theia/discussions/13314) for guidance.
```
jfaltermeier, line 57: We should probably use a more recent example with the updated wording: https://github.com/eclipse-theia/theia/discussions/14547

jfaltermeier (commented):
I only have one more comment, then it looks good to me.

doc/Publishing.md
sgraband: Good point. I updated the link

sgraband:
@jfaltermeier Could you take another look at the slight adjustments i added in the last commit? It includes better descriptions for patch releases. TIA

doc/Publishing.md
```diff
@@ -106,51 +106,95 @@ Format:
...
+  - Expiration: `7 days`.
+  - Under Packages and scopes:
+    - Permissions: `Read and write`.
+    - Select `Only select packages and scopes`
+    - Select packages and scopes: `@theia`.
+
+- Set the token:
+
+  ```bash
+  npm set "//registry.npmjs.org/:_authToken=${TOKEN}"
```
jfaltermeier, line 139: I usually do this with a leading space. On many shells, this prevents the command from being added to the shell's history.

sgraband: Good point. Added a commit for this :+1:

jfaltermeier (approved):

Issue #14307, opened 2024-10-14
Dev Container uses Wrong Window


tsmaeder:
### Bug Description:
When I have multiple windows open, clicking on the "><" icon on the bottom left of the the first opened window and invoking the command "Reopen in Container" Uses that last opened window instead of the window I clicked on.

### Steps to Reproduce:

1. Open a Theia window with a `.devcontainer` folder
2. Open another Theia window on a workspace that does not have the `.devcontainer` folder
3. In the first window, try starting the devcontainer as described above
4. Observe nothing happens. When I open the browser log, it says "Error: ENOENT: no such file or directory, scandir 'C:\Users\thomas\code\theia\theia2\.devcontainer'"

Further experimentation reveals that I have to restart all of Theia IDE to get rid of the error message once it happens

### Additional Information

- Operating System: Windows 11
- Theia Version: 1.54.0

tsmaeder:
@sgraband

JonasHelming:
@jonah-iden

PR #14557, opened 2024-11-28
Dev Container uses the right workspace now

jbicker:
#### What it does

When there are more than one workspaces open the config json is picked from the right workspace. 
Solves #14307

#### How to test

See #14307 and try to reproduce the issue. Should not be possible anymore. :-)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/dev-container/src/electron-node/dev-container-file-service.ts
+++ b/packages/dev-container/src/electron-node/dev-container-file-service.ts
@@ -38,12 +38,7 @@ export class DevContainerFileService {
         return configuration;
     }
 
-    async getAvailableFiles(): Promise<DevContainerFile[]> {
-        const workspace = await this.workspaceServer.getMostRecentlyUsedWorkspace();
-        if (!workspace) {
-            return [];
-        }
-
+    async getAvailableFiles(workspace: string): Promise<DevContainerFile[]> {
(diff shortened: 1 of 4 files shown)


jonah-iden:
The linting task still fails

jonah-iden (approved):
For documentation purposes: There is no support for multi root workspaces yet. But thats something for another task
Works like a charm, and code looks good.

packages/dev-container/src/electron-node/dev-container-file-service.ts
```diff
@@ -54,7 +49,7 @@ export class DevContainerFileService {
     }
 
     protected async searchForDevontainerJsonFiles(directory: string, depth: number): Promise<string[]> {
-        if (depth < 0) {
+        if (depth < 0 || !fs.existsSync(directory)) {
```
msujew, line 52: Please don't use sync file system code in the backend - it blocks the whole server process and is bad practice.

jfaltermeier:
Just an FYI, this doesn’t seem to work on the Theia IDE Preview for 1.57 at the moment. We haven’t investigated yet: https://github.com/eclipse-theia/theia/discussions/14643#discussioncomment-11602921

Issue #14292, opened 2024-10-11
Improve handling of dev container workspaces


sgraband:
### Bug Description:
There are no designated dev-container entries (e.g., `[dev-container] <workspaceName>`) in the recent workspaces list that would allow for directly reopening a workspace as a development container. Currently, reopening a dev container requires opening the folder as a normal workspace first, then manually reopening it as a dev container, which could be streamlined into a single click.

Additionally, when closing Theia with a development container as the last active workspace, the expected behavior is for Theia to reopen in the same dev container upon restart. However, Theia currently reopens the workspace in its normal state, requiring manual reopening as a dev container, adding unnecessary steps to the workflow.

Furthermore, there is no clear indication in the workspace name (displayed at the top of the application) to show that a workspace is opened as a dev container. The only visual cue is in the status bar, which can be easily overlooked.

Improving the handling of development container workspaces by:
- [ ] Adding designated entries in the recent workspaces list for direct reopening as dev containers.
- [ ] Ensuring Theia reopens in the last active dev container automatically after restart.
- [ ] Providing clearer visual indicators (such as in the workspace name) when a workspace is opened as a dev container.

### Steps to Reproduce:
1. Open a workspace as a dev container.
2. Observe that the workspace name does not indicate that it is a dev container.
3. Close Theia.
4. Reopen Theia.
5. Notice that the workspace is no longer in the dev container.
6. Check the recent workspaces list and observe that there is no entry to directly reopen the workspace as a dev container.

### Additional Information:
- Operating System: Ubuntu 22.04
- Theia Version: Theia IDE v1.54.0

PR #14567, opened 2024-12-02
Devcontainer workspaces openable through recent workspaces

jonah-iden:
#### What it does

This adds devcontainer workspaces to the recent workspaces so they can be opened from there. 
Also adds the ability to add custom workspace types through contributions to theia in general

Maybe the `workspaceServer` should be a `localProxy` so that it still shows local workspaces even when connected to a remote? Any comments regarding this would be appreciated.

Implements #14292

#### How to test

Open a dev-container. After disconnecting again the workspace should appear in recent workspaces as `workspace name  devcontainer name [Dev Container]`. Selecting it should directly open the devcontainer.

#### Follow-ups

Since devcontainers is using the WorkspaceStorageService to store the last container info, its different based on what workspace is open. So if no workspace is open it will create a new container instead of reusing the initially created one.
We need some other place to store this info workspace independent.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/dev-container/src/electron-browser/container-connection-contribution.ts
+++ b/packages/dev-container/src/electron-browser/container-connection-contribution.ts
@@ -92,7 +92,7 @@ export class ContainerConnectionContribution extends AbstractRemoteRegistryContr
         };
         const files = await this.connectionProvider.getDevContainerFiles(uri.path.toString());
         const devcontainerFile = files.find(file => file.path === containerFilePath);
-        return `${uri.path.base} @ ${devcontainerFile?.name} [Dev Container]`;
+        return `${uri.path.base} [Dev Container: ${devcontainerFile?.name}]`;
     }
 
     async openInContainer(): Promise<void> {
--- a/packages/dev-container/src/electron-browser/container-connection-contribution.ts
+++ b/packages/dev-container/src/electron-browser/container-connection-contribution.ts
@@ -16,12 +16,14 @@
 
 import { inject, injectable } from '@theia/core/shared/inversify';
 import { AbstractRemoteRegistryContribution, RemoteRegistry } from '@theia/remote/lib/electron-browser/remote-registry-contribution';
-import { LastContainerInfo, RemoteContainerConnectionProvider } from '../electron-common/remote-container-connection-provider';
+import { DevContainerFile, LastContainerInfo, RemoteContainerConnectionProvider } from '../electron-common/remote-container-connection-provider';
 import { RemotePreferences } from '@theia/remote/lib/electron-browser/remote-preferences';
 import { WorkspaceStorageService } from '@theia/workspace/lib/browser/workspace-storage-service';
-import { Command, QuickInputService } from '@theia/core';
-import { WorkspaceService } from '@theia/workspace/lib/browser/workspace-service';
+import { Command, MaybePromise, QuickInputService, URI } from '@theia/core';
+import { WorkspaceInput, WorkspaceOpenHandlerContribution, WorkspaceService } from '@theia/workspace/lib/browser/workspace-service';
 import { ContainerOutputProvider } from './container-output-provider';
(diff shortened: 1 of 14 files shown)


sgraband:
> Maybe the workspaceServer should be a localProxy so that it still shows local workspaces even when connected to a remote? Any comments regarding this would be appreciated.

I believe that would be a great addition. This helps a lot with not having to change your workflows when working in a devconainer imho. I am however not sure what the effort is for this. But if it is a bigger effort we should definately track this in an issue :+1:

jonah-iden:
nah thats one code line change. I'm just not sure if this could create any other issues. But i don't think so. 
i guess lets just try it out

jonah-iden:
ok iguess its a little more work, since opening local workspaces would require disconnecting the remote

jonah-iden:
i think it's better to do the local-workspace server in another task. Its quite a bit more work than i first thought. Also on a conceptual level since in theory you can have multiple workspaceses inside of the container and outside on the local system.

packages/dev-container/src/electron-browser/container-connection-contribution.ts
```diff
@@ -59,46 +64,85 @@ export class ContainerConnectionContribution extends AbstractRemoteRegistryContr
...
+    }
+
+    async getWorkspaceLabel(uri: URI): Promise<string | undefined> {
+        const containerFilePath = new URLSearchParams(uri.query).get(DEV_CONTAINER_PATH_QUERY);
+        if (!containerFilePath) {
+            return;
+        };
+        const files = await this.connectionProvider.getDevContainerFiles(uri.path.toString());
+        const devcontainerFile = files.find(file => file.path === containerFilePath);
+        return `${uri.path.base} @ ${devcontainerFile?.name} [Dev Container]`;
```
sgraband, line 95: I personally find this way of representing the information a bit to clutered. I would prefer something like in VSCode (e.g. `{uri.path.base} [Dev Container: ${devcontainerFile?.name}]`.
But this is of course only my personal feeling. WDYT?

Also would it be possible to render the output of the `getWorkspaceLabel()` method in the title bar of the application? Currently this does not display the indicator that it is a dev container in the title for me.

sgraband (commented):
Thanks for the change @jonah-iden! Great improvement and works great! I just have one inline comment and a general remark:

Currently, when the last workspace is a dev container and i open Theia again it will not bring up the dev container, but an empty workspace. Is there something we can do about that?

sgraband:
> i think it's better to do the local-workspace server in another task. Its quite a bit more work than i first thought. Also on a conceptual level since in theory you can have multiple workspaceses inside of the container and outside on the local system.


Yes i agree, could you create a ticket for this with your finding or should i take this over?

jonah-iden:
sure, i'll create a ticket

packages/dev-container/src/electron-browser/container-connection-contribution.ts
jonah-iden: This is exactly how its displayed for me in vscode. So i just adapted that. But im totally open to any other formats

Regarding your second point thats something Jan is doing in this [PR](https://github.com/eclipse-theia/theia/pull/14571#event-15506041995)

jonah-iden:
created an issue [here](https://github.com/eclipse-theia/theia/issues/14574)

sgraband (approved):
Thanks @jonah-iden! I think this label is much more readable :tada:

sgraband (commented):
> Currently, when the last workspace is a dev container and i open Theia again it will not bring up the dev container, but an empty workspace. Is there something we can do about that?

I forgot about this. Do you think this is easily fixable or should we create a follow up for this as well?

jonah-iden:
its kind of part of the other issue i created recently which would allow opening local workspaces from when connected to a remote or container. At least its not super simple to do but im already looking into the new issue so i can take a look here as well

PR #14573, opened 2024-12-03
basics to allow for hidden cells

jonah-iden:
#### What it does

Adds basic functionality that can be used by adopters to add logic for hidden cells. Lays the foundation for adding support for colapsing cell regions.

#### How to test

Hard to test since it does not really add anything new. Best way is probably to modify the `getVisibleCells`method in the `NotebookModel` to check that everything sitll works for the visible cells

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/notebooks/renderers/cell-output-webview.tsx
+++ b/packages/plugin-ext/src/main/browser/notebooks/renderers/cell-output-webview.tsx
@@ -354,21 +354,27 @@ export class CellOutputWebviewImpl implements CellOutputWebview, Disposable {
             this.webviewWidget.show();
         }
 
+        const visibleCells = this.notebook.getVisibleCells();
+
         const updateOutputMessage: OutputChangedMessage = {
             type: 'outputChanged',
-            changes: updates.map(update => ({
-                cellHandle: update.cellHandle,
-                newOutputs: update.newOutputs.map(output => ({
-                    id: output.outputId,
-                    items: output.outputs.map(item => ({ mime: item.mime, data: item.data.buffer })),
-                    metadata: output.metadata
-                })),
-                start: update.start,
-                deleteCount: update.deleteCount
-            }))
+            changes: updates
+                .filter(update => visibleCells.some(cell => cell.handle === update.cellHandle))
+                .map(update => ({
+                    cellHandle: update.cellHandle,
+                    newOutputs: update.newOutputs.map(output => ({
+                        id: output.outputId,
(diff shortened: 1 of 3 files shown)


packages/plugin-ext/src/main/browser/notebooks/renderers/cell-output-webview.tsx
```diff
@@ -354,21 +354,27 @@ export class CellOutputWebviewImpl implements CellOutputWebview, Disposable {
...
-                newOutputs: update.newOutputs.map(output => ({
-                    id: output.outputId,
-                    items: output.outputs.map(item => ({ mime: item.mime, data: item.data.buffer })),
-                    metadata: output.metadata
-                })),
-                start: update.start,
-                deleteCount: update.deleteCount
-            }))
+            changes: updates
+                .filter(update => visibleCells.some(cell => cell.handle === update.cellHandle))
```
msujew, line 362: Suggestion: This leads to `n * m` runtime behavior. Can we build something this instead and use `has`?

```ts
const visibleCellHandles = new Set(this.notebook.getVisibleCells().map(cell => cell.handle));
```

packages/plugin-ext/src/main/browser/notebooks/renderers/cell-output-webview.tsx
```diff
@@ -382,11 +388,12 @@ export class CellOutputWebviewImpl implements CellOutputWebview, Disposable {
                     toIndex: event.newIdx + i,
                 } as CellsMoved)));
             } else if (event.kind === NotebookCellsChangeType.ModelChange) {
+                const visibleCells = this.notebook.getVisibleCells();
                 changes.push(...event.changes.map(change => ({
                     type: 'cellsSpliced',
                     start: change.start,
                     deleteCount: change.deleteCount,
-                    newCells: change.newItems.map(cell => cell.handle)
+                    newCells: change.newItems.filter(cell => visibleCells.includes(cell as NotebookCellModel)).map(cell => cell.handle)
```
msujew, line 396: Suggestion: `n * m` behavior again, please use a `Set` instead.

packages/plugin-ext/src/main/browser/notebooks/renderers/cell-output-webview.tsx
jonah-iden: not sure if thats really that worth it, since in 90% of cases this will probably only be a single update. But i changed it. Your probably right that it could improve performance for a few cases

PR #14608, opened 2024-12-10
chore(ai): enable chat input on chat widget activate

planger:
#### What it does

We get warnings in the console from the chat widget as it doesn't accept focus. More importantly, it is cumbersome having to click the input field to start typing.

This change therefore focuses the input field when the chat widget is activated. Also it shows the placeholder text until a text is entered, which is a nicer touch than before.

#### How to test

Active, hide and reactive the chat window and observe how the input receives focus and you can immediately start typing.
Check if the placeholder shows and disappears nicely when you enter or delete text in the chat input.

#### Follow-ups

None

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
@@ -22,6 +22,7 @@ import * as React from '@theia/core/shared/react';
 import { MonacoEditorProvider } from '@theia/monaco/lib/browser/monaco-editor-provider';
 import { CHAT_VIEW_LANGUAGE_EXTENSION } from './chat-view-language-contribution';
 import { IMouseEvent } from '@theia/monaco-editor-core';
+import { Deferred } from '@theia/core/lib/common/promise-util';
 
 type Query = (query: string) => Promise<void>;
 type Cancel = (requestModel: ChatRequestModel) => void;
@@ -44,6 +45,7 @@ export class AIChatInputWidget extends ReactWidget {
     protected readonly contextMenuRenderer: ContextMenuRenderer;
 
     protected editorRef: MonacoEditor | undefined = undefined;
+    private editorReady = new Deferred<void>();
 
     protected isEnabled = false;
 
@@ -70,9 +72,11 @@ export class AIChatInputWidget extends ReactWidget {
 
     protected override onActivateRequest(msg: Message): void {
         super.onActivateRequest(msg);
-        if (this.editorRef) {
-            this.editorRef.focus();
-        }
+        this.editorReady.promise.then(() => {
+            if (this.editorRef) {
+                this.editorRef.focus();
+            }
+        });
     }
 
     protected getChatAgents(): ChatAgent[] {
@@ -90,7 +94,10 @@ export class AIChatInputWidget extends ReactWidget {
                 untitledResourceResolver={this.untitledResourceResolver}
                 contextMenuCallback={this.handleContextMenu.bind(this)}
                 isEnabled={this.isEnabled}
-                setEditorRef={editor => { this.editorRef = editor; }}
+                setEditorRef={editor => {
+                    this.editorRef = editor;
+                    this.editorReady.resolve();
+                }}
             />
         );
     }
@@ -201,14 +208,11 @@ const ChatInput: React.FunctionComponent<ChatInputProperties> = (props: ChatInpu
     React.useEffect(() => {
         createInputElement();
         return () => {
+            props.setEditorRef(undefined);
             if (editorRef.current) {
                 editorRef.current.dispose();
             }
         };
-    }, []);
-
-    React.useEffect(() => () => {
-        props.setEditorRef(undefined);
     }, []);
 
     React.useEffect(() => {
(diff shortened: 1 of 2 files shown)


packages/ai-chat-ui/src/browser/chat-input-widget.tsx
```diff
@@ -65,9 +67,12 @@ export class AIChatInputWidget extends ReactWidget {
         this.title.closable = false;
         this.update();
     }
+
     protected override onActivateRequest(msg: Message): void {
         super.onActivateRequest(msg);
-        this.node.focus({ preventScroll: true });
+        if (this.editorRef) {
+            this.editorRef.focus();
+        }
```
sdirix, line 75: On first rendering of the view, the `editorRef` does not exist yet. So the input will not get focus. I think we need to work with a Promise which resolves once the editor is available.

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
```diff
@@ -85,6 +90,7 @@ export class AIChatInputWidget extends ReactWidget {
                 untitledResourceResolver={this.untitledResourceResolver}
                 contextMenuCallback={this.handleContextMenu.bind(this)}
                 isEnabled={this.isEnabled}
+                setEditorRef={editor => { this.editorRef = editor; }}
```
sdirix, line 93: If we keep the `ref` solution, then we should reuse the React ref concept instead repeating it.

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
```diff
@@ -196,6 +207,10 @@ const ChatInput: React.FunctionComponent<ChatInputProperties> = (props: ChatInpu
         };
     }, []);
 
+    React.useEffect(() => () => {
+        props.setEditorRef(undefined);
+    }, []);
```
sdirix, line 212: We could place this into the previous `useEffect`, no need to create a new one

sdirix (changes requested):
It works when the chat view already exists. However on first creation it does not work.

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
planger: Right! I'm happy to move it into the existing one. I don't know if it causes overhead, but sometimes I feel adding one `useEffect` per concern is easier to understand. But you know more React than I do, so I'm happy to move it.

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
planger: Is there an alternative solution that you'd prefer?

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
planger: Hm, I can't reproduce the issue. If I have the view closed, refresh and then open it, the focus is correctly obtained.
If I have the view open and then refresh, the input receives focus, but then something else steals it in Theia, which I guess is fine, I don't know if there is a dedicated focus priority.

sdirix: InitialOpeningChat

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
sdirix: In general I agree, however here for me both of these `useEffect` have the same topic: life cycle management, i.e. what to do when the component is mounted/unmounted and therefore belong together.

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
planger: Thanks for demonstrating the case! This is quite curious, because for me this works on a clean build of the exact branch commit:

https://github.com/user-attachments/assets/77f71eee-81e4-4735-b582-0da0798de5a9

Anyway, I understand why this may not work and will try to propose a fix.

planger: I added a new commit ([1862f9b](https://github.com/eclipse-theia/theia/pull/14608/commits/1862f9ba4434c8c8d7c207d90c6ff727fc88798e)), can you please check if this now works for you? Thank you!

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
```diff
@@ -185,9 +202,11 @@ const ChatInput: React.FunctionComponent<ChatInputProperties> = (props: ChatInpu
         );
 
         editorRef.current = editor;
+        props.setEditorRef(editor);
     };
 
     React.useEffect(() => {
+        props.setEditorRef(undefined);
```
sdirix, line 209: This must be moved into the returned function

sdirix (changes requested):
Works for me :+1: , however there is an issue in the code

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
planger: Nice catch, sorry I mismoved it in my previous commit!

sdirix (approved):
Works for me. Thanks!

PR #14614, opened 2024-12-11
Support Anthropic as an LLM provider

### LINKED ISSUE #14613: [Theia AI] Support Anthropic as an LLM provider

JonasHelming:
fixed #14613

#### What it does

Adds support for Anthropic as an LLM provider.

#### How to test

Use Anthropic models.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/browser/package.json
+++ b/examples/browser/package.json
@@ -22,6 +22,7 @@
   },
   "theiaPluginsDir": "../../plugins",
   "dependencies": {
+    "@theia/ai-anthropic": "1.56.0",
     "@theia/ai-chat": "1.56.0",
     "@theia/ai-chat-ui": "1.56.0",
     "@theia/ai-code-completion": "1.56.0",
--- a/examples/browser/tsconfig.json
+++ b/examples/browser/tsconfig.json
@@ -9,6 +9,9 @@
       "path": "../../dev-packages/cli"
     },
     {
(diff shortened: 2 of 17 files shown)


JonasHelming:
@sdirix Tool functions do not work yet. That is probably in the way how they are parsed from the stream, but I have no clue how that works for openai TBH, maybe you can have a quick look at this.

JonasHelming:
3PP Check was approved, should succeed with the next rebuild

sdirix (commented):
Anthropic works for me, I tested with streaming enabled and disabled.

I also had a look at the tools. The issue is that in our OpenAI language models we directly hand over real Javascript functions into the API which are used by the OpenAI client to resolve the tools, see [here](https://github.com/eclipse-theia/theia/blob/79d55445839d5b37e54f8a8999e552c3be6eb755/packages/ai-openai/src/node/openai-language-model.ts#L221).

Anthropic seemingly does not support this but instead expects a new message from the "user", see the [first example here](https://docs.anthropic.com/en/docs/build-with-claude/tool-use#tool-use-examples).

Suggestion: For the first implementation we skip the tool support and add this in a followup. Of course feel free to add it here too if you want to see it supported.

sdirix (approved):
Works for me!

PR #14618, opened 2024-12-12
Fix rendering of output of cells added with already existing output

jonah-iden:
#### What it does

This is a fix so that outputs of cells are rendered even when a new cell is added that already has outputs. This can for example happen in the case of undo.

#### How to test

- Open a notebook with at least one cell
- Execute it
- Delete the cell
- undo
- see that the output should be rendered correctly

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/notebooks/renderers/output-webview-internal.ts
+++ b/packages/plugin-ext/src/main/browser/notebooks/renderers/output-webview-internal.ts
@@ -613,6 +613,7 @@ export async function outputWebviewPreload(ctx: PreloadContext): Promise<void> {
             }
 
             cell.clearOutputs(cellChange.start, cellChange.deleteCount);
+
             for (const outputData of cellChange.newOutputs ?? []) {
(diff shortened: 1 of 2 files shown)


packages/plugin-ext/src/main/browser/notebooks/renderers/output-webview-internal.ts
```diff
@@ -613,7 +613,7 @@ export async function outputWebviewPreload(ctx: PreloadContext): Promise<void> {
             }
 
             cell.clearOutputs(cellChange.start, cellChange.deleteCount);
-
+            console.log(cellChange.newOutputs);
```
msujew, line 616: Leftovers from testing?

jonah-iden: oh yeah. Thanks!

msujew (approved):

PR #14621, opened 2024-12-13
fix cell height when updating output

jonah-iden:
#### What it does

This fixes an issue that the cell height was not updated in the output webview after rendering an output. This lead to overlap of output with other cells

#### How to test

- Open a notebook 
- create a few cells and execute them
- validate that changing the cell height (by for example ading a line) the distance  between cell and output does not change anymore.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


dhuebner (approved):
Looks good to me. Thanks!
The bogus behavior is no longer reproducible.

Issue #14285, opened 2024-10-10
AI Workspace agent freezes


dannaf:
### Bug Description:

The Workspace AI agent 'generates' without displaying any response, perhaps only when querying the workspace.

As the screenshot below shows, my AI api key etc. is set up correctly and the AI agent is connected in the chat, given that the test message went through and a response was quickly provided.  But when I asked even a simple question about the workspace, it got stuck and never displayed any result. (I eventually gave up and pressed the stop button &mdash; also to make sure it is not billing me for an infinite loop that it is not even displaying the results of, just in case.)

### Steps to Reproduce:

1. Enable the experimental AI features in 1.54.0 and connect an OpenAI API key in the settings.
2. I then disabled all the agents except the Workspace agent in the AI configuration (maybe I still had the code-completion and terminal agents, but definitely disabled the other ones like command and universal).
3. Type a simple test message that does not query regarding anything about the workspace ("Test" as shown below, or "can you hear me", etc.). Observe response from the AI agent to confirm that the API key is connected.
4. Then query something about the actual workspace and observe the AI agent get stuck, supposedly 'generating'.

![image](https://github.com/user-attachments/assets/2a5f92d2-3a94-49a4-92c9-4f48f567db1b)

### Additional Information

The Theia log showed the following message:

```
2024-10-10T11:35:53.610Z root ERROR [hosted-plugin: 263] There is no document for untitled:/Untitled-1aichatviewlanguage
```

- Operating System: Oracle Linux 2022 (for hosting theia/theia-blueprint; accessed via Windows 11 via the browser)
- Theia Version: 1.54.0 (via Firefox 131.0)

JonasHelming:
@dannaf : Thank you for the report. Was this a rather big workspace by any chance?

dannaf:
Not at all

JonasHelming:
Just to make sure, you can reproduce this all the time?

JonasHelming:
I just tried the exact same query, it works for me. However, the question "How many lines of code" will probably force it to send every file, which might take a while. However, it is is doing this, you should see this already:

![image](https://github.com/user-attachments/assets/a4f6b67d-a324-4422-8ed5-6fe4277969f8)

dannaf:
We've made some progress.  I now saw this:

![image](https://github.com/user-attachments/assets/29d6842a-8f4c-4b07-b47f-108e5e24fd38)

[...]

PR #14623, opened 2024-12-13
Ollama LLM provider tools support

### LINKED ISSUE #14610: [Theia AI] Ollama LLM provider does not support tools
Also see here: https://github.com/eclipse-theia/theia/issues/14285

### Additional Information

- Operating System: any
- Theia Version: 1.56.x

dhuebner:
#### What it does

Resolves #14610

Adds Tools handling for Ollama LMs

#### How to test

Set an Ollama model as workspace agent and ask questions about the workspace. For  example:
`@Workspace How many files are in my workspace? `

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-ollama/src/node/ollama-language-model.ts
+++ b/packages/ai-ollama/src/node/ollama-language-model.ts
@@ -20,7 +20,9 @@ import {
     LanguageModelRequest,
     LanguageModelRequestMessage,
     LanguageModelResponse,
+    LanguageModelStreamResponse,
     LanguageModelStreamResponsePart,
+    ToolCall,
     ToolRequest
 } from '@theia/ai-core';
 import { CancellationToken } from '@theia/core';
@@ -31,7 +33,9 @@ export const OllamaModelIdentifier = Symbol('OllamaModelIdentifier');
 export class OllamaModel implements LanguageModel {
 
     protected readonly DEFAULT_REQUEST_SETTINGS: Partial<Omit<ChatRequest, 'stream' | 'model'>> = {
-        keep_alive: '15m'
+        keep_alive: '15m',
+        // options see: https://github.com/ollama/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values
+        options: {}
     };
 
     readonly providerId = 'ollama';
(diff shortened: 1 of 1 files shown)


planger:
@dhuebner Thank you for the PR! Which models did you use for testing?

dhuebner:
@planger 
llama3.1 as WS agent and llama3.2 as orchestrator, why?

planger:
@dhuebner thanks, no particular reason, just out of curiosity and to know how to best test the PR. Thank you!

JonasHelming:
Great feature addition!

I tried to test this with LLama3.1 but I was not really successful (see below).
Also I noticed:
- You do not stream at all in the case of tools present? Is this intentional? For the user, they just see "generating" for quite a while and nothing happens
- Because of not using streaming, we do not show the function calls in the chat as for the open AI provider. I think this is a really nice feature showing the user what happens in a transparent way.

see 
image


image

dhuebner:
@JonasHelming 

> You do not stream at all in the case of tools present? Is this intentional? For the user, they just see "generating" for quite a while and nothing happens

The reason is that Ollama doesn't seem to support tools with streaming. The document says:
`tools: tools for the model to use if supported. Requires stream to be set to false`

There is also no possibility to listen to e.g. `onToolsCall()`, means we will need to buffer the stream until we recognize if it is a `tools_call` and than present it to the user. 

> This was my test with Ollama:

I tried the questions mentioned in #14285 by @navr32 , that worked okay. I will test with more complex questions, although I don't know how to make it perform better with just calling the API provided...

JonasHelming:
@dhuebner OK that makes sense. However I think we should still add the tool call to the response so that is its rendered in the chat.

dhuebner:
@JonasHelming 
> However I think we should still add the tool call to the response so that is its rendered in the chat.

How should it look like? I can send only one response during a request, or is there a special API to achieve this?

planger:
I think at the moment we cannot really represent tool calls generically in the non-streaming case.

In the non-streaming case, we can only return a `LanguageModelTextResponse`:
```ts
export interface LanguageModelTextResponse {
    text: string;
}
```

In the streaming-case we use `LanguageModelStreamResponsePart` which can contain tool calls:
```ts
export interface LanguageModelStreamResponsePart {
    content?: string | null;
    tool_calls?: ToolCall[];
}
```

These are then translated in the method `AbstractStreamParsingChatAgent.parse()` to `ToolCallChatResponseContentImpl`, which will show in the UI.

So I think we may need to extend our `LanguageModelResponse` model to be able to capture tool calls also for non-streaming responses and then process them in the `AbstractTextToModelParsingChatAgent` to translate them into `ToolCallChatResponseContentImpl`, so they show up in the UI.

dhuebner:
@planger 
Good idea!
But for now as we do not have this API, what is still missing in this PR?

sdirix:
Just an idea: Can't we map non-stream requests to streamed ones rather easily? We just pretend that there is a stream and once we have the answer of the LLM we send it as one blob. This way we could reuse the tools?

dhuebner:
@sdirix 
Okay, we can do that

dhuebner:
@sdirix @JonasHelming 
 Now returning a fake stream response with tool-call data attached. The current state looks like this:

Bildschirmfoto 2025-01-10 um 14 46 45


Please consider to make `ToolCall.id` mandatory. It cost me some time to figure out that without an `id`  multiple tool-call message parts are merged into one  in a wired way.

sdirix (approved):
Works for me! I tested with the Workspace Agent and was able to retrieve my workspace files via the tool function, although it took forever to do so on my machine.

PR #14644, opened 2024-12-17
refactor: centralize cancellation token in chat model

planger:
#### What it does

So far it is the responsibility of the chat agent to create the cancellation token and react to the chat model being canceled to update the token that is passed to the language model request. This is very indirect and forces sub agents that may contribute to providing an answer to do the same over and over again for their language model requests.

Instead one chat request/response should have one cancellation token that is being created automatically. The UI can then invoke cancel on the chat request/response and all agents that are involved can reuse the cancellation token for their requests.

#### How to test

Make sure that canceling a chat request/response still works.

#### Follow-ups

None

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-view-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-view-widget.tsx
@@ -14,7 +14,7 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 import { CommandService, deepClone, Emitter, Event, MessageService } from '@theia/core';
-import { ChatRequest, ChatRequestModel, ChatRequestModelImpl, ChatService, ChatSession } from '@theia/ai-chat';
+import { ChatRequest, ChatRequestModel, ChatService, ChatSession } from '@theia/ai-chat';
 import { BaseWidget, codicon, ExtractableWidget, Message, PanelLayout, PreferenceService, StatefulWidget } from '@theia/core/lib/browser';
 import { nls } from '@theia/core/lib/common/nls';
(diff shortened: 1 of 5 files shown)


packages/ai-chat-ui/src/browser/chat-view-widget.tsx
```diff
@@ -176,9 +176,9 @@ export class ChatViewWidget extends BaseWidget implements ExtractableWidget, Sta
     }
 
     protected onCancel(requestModel: ChatRequestModel): void {
-        // TODO we should pass a cancellation token with the request (or retrieve one from the request invocation) so we can cleanly cancel here
-        // For now we cancel manually via casting
-        (requestModel as ChatRequestModelImpl).response.cancel();
+        if (requestModel instanceof ChatRequestModelImpl) {
+            requestModel.cancel();
+        }
```
sdirix, line 181: As we're already cleaning up the code:
We should avoid this cast wherever possible, ideally everywhere. Having to use the cast means that there is something not properly cleaned up in the code base. Either:
- the `chat-view-widget` should have access to the `ChatRequestModelImpl` directly (without casting), OR
- the cancel should be triggered differently, e.g. via the token or a callback from the invocation

Generally: Should have the `chat-view-widget` access to the `ChatRequestModelImpl`? Only mutators should have access.

planger: Right, avoiding those casts should indeed be the goal!
I've moved the cancellation now into the chat service, as the widget already interacts with it to send the request and the chat service already deals with the instantiation and its concrete impl classes.
WDYT?

sdirix (approved):
LGTM!

PR #14649, opened 2024-12-18
Use cellHandle instead of index when talking to output webview

dhuebner:
#### What it does

Now using cellHandle instead of index when "talking" to the output webview. The reason is that in some cases (cell deletion) there is no way to identify a visible cell to the corresponding start index as the notebook no longer contain this cell. 

Doing so, we also align `cellsSpliced` to existing `cellMoved` webview communication channel, that already uses handles instead of ids.   
 
#### How to test

Open a notebook with many cells that has output.  Delete some cell in the middle. Check that there are any old/overlapping outputs. 

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/view-model/notebook-model.ts
+++ b/packages/notebook/src/browser/view-model/notebook-model.ts
@@ -394,7 +394,12 @@ export class NotebookModel implements Saveable, Disposable {
         });
         this.addCellOutputListeners(cells);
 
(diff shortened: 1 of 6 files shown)


dhuebner:
@JonasHelming 
How to create a `next` release from this PR? This is an important fix...

jonah-iden:
@dhuebner for a next release you just go to actions in this tab. Select the publish next release action and then run the run workflow button. There you can select this branch. 
It could be that it goes red but still publishes all important packages

jbicker (approved):
LGTM

Issue #14067, opened 2024-08-23
File streaming broken with socket.io polling


ramymagdy:
### Bug Description:

We have to run Theia through a proxy that doesn't support the Connection "Upgrade"  for websocket. So instead we need to rely on socket.io long polling.

With the socket.io connection polling the File Explorer act's randomly and no files are opened in the editor ( regardless of the file type )
I have to go randomly clicking through files   till I get one of the opening, 

![theia](https://github.com/user-attachments/assets/6c1f80f6-0465-4b5c-8cde-86a4bd7a50f5)

### Steps to Reproduce:

Apply the following patch

```
diff --git a/packages/core/src/browser/messaging/ws-connection-source.ts b/packages/core/src/browser/messaging/ws-connection-source.ts
index 24d7d7920..5d4a60a37 100644
--- a/packages/core/src/browser/messaging/ws-connection-source.ts
+++ b/packages/core/src/browser/messaging/ws-connection-source.ts
@@ -202,6 +202,7 @@ export class WebSocketConnectionSource implements ConnectionSource {
     protected createWebSocket(url: string): Socket {
         return io(url, {
             path: this.createSocketIoPath(url),
+           transports: ["polling"],
             reconnection: true,
             reconnectionDelay: 1000,
             reconnectionDelayMax: 10000,
diff --git a/packages/core/src/node/messaging/websocket-endpoint.ts b/packages/core/src/node/messaging/websocket-endpoint.ts
index 187603c77..f98e50ce7 100644
--- a/packages/core/src/node/messaging/websocket-endpoint.ts
+++ b/packages/core/src/node/messaging/websocket-endpoint.ts
@@ -42,6 +42,7 @@ export class WebsocketEndpoint implements BackendApplicationContribution {
 
     onStart(server: http.Server | https.Server): void {
         const socketServer = new Server(server, {
+           transports: ["polling"],
             pingInterval: this.checkAliveTimeout,
             pingTimeout: this.checkAliveTimeout * 2,
             maxHttpBufferSize: this.maxHttpBufferSize

```

Or else you may use the following haproxy configurations that will just deny any request with "Upgrade" header.

```

global
	log /dev/log	local0
	log /dev/log	local1 notice
	chroot /var/lib/haproxy
	stats socket /run/haproxy/admin.sock mode 660 level admin expose-fd listeners
	stats timeout 30s
	user haproxy
	group haproxy
	daemon

	# Default SSL material locations
	ca-base /etc/ssl/certs
[...]

PR #14659, opened 2024-12-20
Fix file data streaming for http polling

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/14067

When using socket.io's long polling mechanism, sending messages to the frontend in the backend while responding to a request from the frontend might result in messages arriving in an unexpected order. In particular, it might result in those notifications arriving before the request response arrives.

Usually this is no issue, since we rarely depend on the exact order of messages. However, in the file data streaming, we expect the response before any of the notifications arrive. This seems to be the case with websockets, but not with long polling.

This change makes the frontend in charge of the sequence id handle instead of the backend. This entirely removes the possibility of a race condition without impacting the behavior of the file streaming feature.

#### How to test

1. Follow the reproduction steps from https://github.com/eclipse-theia/theia/issues/14067.
2. Ensure that files can be successfully opened.

#### Follow-ups

We might want to look into any potential other places where we depend on notifications + a response for a backend RPC call. I don't think there are any others, but I wouldn't guarantee it.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/common/remote-file-system-provider.ts
+++ b/packages/filesystem/src/common/remote-file-system-provider.ts
@@ -46,7 +46,7 @@ export interface RemoteFileSystemServer extends RpcServer<RemoteFileSystemClient
     open(resource: string, opts: FileOpenOptions): Promise<number>;
     close(fd: number): Promise<void>;
     read(fd: number, pos: number, length: number): Promise<{ bytes: Uint8Array; bytesRead: number; }>;
-    readFileStream(resource: string, opts: FileReadStreamOptions, token: CancellationToken): Promise<number>;
+    readFileStream(resource: string, handle: number, opts: FileReadStreamOptions, token: CancellationToken): Promise<void>;
     readFile(resource: string): Promise<Uint8Array>;
     write(fd: number, pos: number, data: Uint8Array, offset: number, length: number): Promise<number>;
     writeFile(resource: string, content: Uint8Array, opts: FileWriteOptions): Promise<void>;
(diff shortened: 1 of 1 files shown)


tsmaeder:
@msujew I have to disagree that message ordering is not important, see for example: https://github.com/eclipse-theia/theia/issues/13960. This is exactly the kind of problem that is totally baffling and seems to appear at randoem. Can we fix the problem in the polling socket implementation so that it respects message ordering?

msujew:
@tsmaeder I've changed the code so that the race condition observed in the initial post is no longer possible. The frontend is now in charge of the file handle instead of receiving the file handle from the backend.

msujew:
@tsmaeder I've read up on this a bit [socket.io guarantees message ordering on both websocket and http fallback](https://socket.io/docs/v4/delivery-guarantees). I would generally assume that this works as intended. However, what we've been doing is a bit of a hack anyway: We attempt to respond with a handler id, that we already use in the events, without making any guarantees about the order. This doesn't seem to be a problem in websocket contexts, but likely due to some specific implementation details. I see no reason why this also shouldn't fail on websockets.

The new implementation fixes this issue by specifiying the handler on the frontend level, something that is way cleaner anyway and something that we do throughout the rest of the codebase as well.

tsmaeder:
Ah, I get it:

1. Front end requests stream read
2. back end starts servicing request
3. back end allocates handle for stream
4. back end starts reading the file, generating events sent to front end
5. back end sends hande as reply to original request.

Just wanted to make sure we don't overlook some deeper problem here.

tsmaeder:
The only difference I can see for polling is that websocket probably uses some native event mechanism, whereas the polling likely goes through a timeout. That might reorder the execution releative to promises.

tsmaeder:
Never mind, I was able to reproduce to for some files in the browser version.

tsmaeder (approved):

PR #14668, opened 2024-12-23
Fix override of default key bindings

### LINKED ISSUE #14667: Cannot override default key bindings
### Bug Description:

It is not possible to override default key bindings, the original command will still be triggered, even if the user assigns the command to another key binding.

### Steps to Reproduce:

1. open keybindings and search for "trigger inline completion"
2. Rebind to another key combination
3. open a code editor and observe that the key combination still triggers inline completion (and does not add a special space)

### Additional Information

- Operating System:
- Theia Version: 1.57

JonasHelming:
fixed #14667

#### What it does

fixes #14667
If a key binding is overridden by the user, an entry like this is added to the keymaps.json to indicate that the original keybinding is not valid anymore (with a '-' in the beginning of the command id:
```
    {
        "command": "-editor.action.inlineSuggest.trigger",
        "keybinding": "Shift+Space",
        "when": "!editorReadonly && editorTextFocus"
    }
```
The code in 'matchKeybinding' (see only change in this PR) records this and makes the key binding invalid also in upper contexts. However, it first checks "isEnabled" which in 'isEnabledInScope' checks 'this.commandRegistry.isEnabled(binding.command, binding.args)'
This is always false for invalid commands and the '-' makes commands invalid. As a consequence, 'isUsable' is not checked anymore, and the isUsable list is not correctly filled.
Therefore, turning the checks around, i.e. first checking isUsable fixes the issues, as overriden commands are now correctly remembers.

#### How to test

see #14667

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/keybinding.ts
+++ b/packages/core/src/browser/keybinding.ts
@@ -622,13 +622,13 @@ export class KeybindingRegistry {
     matchKeybinding(keySequence: KeySequence, event?: KeyboardEvent): KeybindingRegistry.Match {
         let disabled: Set<string> | undefined;
         const isEnabled = (binding: ScopedKeybinding) => {
-            if (event && !this.isEnabled(binding, event)) {
-                return false;
-            }
             const { command, context, when, keybinding } = binding;
             if (!this.isUsable(binding)) {
                 disabled = disabled || new Set<string>();
                 disabled.add(JSON.stringify({ command: command.substring(1), context, when, keybinding }));
+                return false;
+            }
+            if (event && !this.isEnabled(binding, event)) {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


tsmaeder:
I'm not sure I understand the bug here: when I go to the keyboard shortcuts editor and I press "edit" on the "Copy" binding with no "when" clause, I can easily rebind that to "ctrl-i". What's different in your example?

JonasHelming:
After rebinding, the original default binding will still be active. Press Shift Space and it will still trigger the inline suggestion (in addition to Ctrl+I which will also work)

tsmaeder (approved):

PR #14676, opened 2024-12-27
Refactor menu nodes

### LINKED ISSUE #14217: Refactor MenuNode to get rid of MenuExecutor, MenuCommandAdapterRegistry and MenuCommandExecutor
### Feature Description:
Currently menu nodes (whether they be sub-menus, groups, separators or actions) are pure data. users of the `MenuNode` interfaces must implement all concerns like whether a menu item is visible in a context or how to adapt menu command parameters (for example, converting Widget instances to a view id to send to a plugin) themselves. In order to prevent breaking layers, we have introduced a registry infrastructure (`MenuCommandAdapterRegistry` ) that allows in particular the VS Code contribution handlers to influence the execution and visibility of items in toolbars and menus. 
By making menu nodes active and polymorphous (for example, replacing the `command` field with a `run` method), we could simplify the menu item handling and move the complexity to where it is needed and hide it from the rest of the system.

tsmaeder:
#### What it does
Fixes #14217

Makes menu nodes active object that can decide on visibility, enablement, etc. themselves. This should simplify handling of menus and toolbar items because a lot of special-case code can be moved to polymorphous implementations instead of being handled at each client site. This also allows the special handling of parameter transformations for plugin-contributed actions (toolbar/menu) to be confined to the plugin engine implementation.

#### How to test
Test that menu functionality still works, in particular actions contributed from plugins. That includes menus and toolbars everywhere for browser and electron cases.

#### Follow-ups
One strange thing is with the menu/toolbar support is that the `MenuRegistry` and associated classes are located in the "common" part of the core packages. If we decide that this is no longer that case, we could move the code to "browser". However, this would introduce another huge batch of changes and therefore has been left out of this PR.
There are also a couple of renamings that would make the naming more consistent and logical that have been left out of the PR to keep the size of the PR manageable.

#### Breaking changes
This PR introduces breaking changes to be documented after a first round of reviews.

#### Attribution

#### Review checklist

#### Reminder for reviewers


dhuebner:
@tsmaeder 
Linting and Playwright-tests are failing. I will fix the tests and add a commit. 
Also `flat` menu functionality is not working (see screenshot) 

Bildschirmfoto 2025-01-03 um 11 57 25

Notebook cell actions are broken too.

tsmaeder:
@dhuebner What are you trying to achieve through CompoundMenuNodeRole.flat?

packages/toolbar/src/browser/toolbar.tsx
```diff
@@ -234,23 +232,11 @@ export class ToolbarImpl extends TabBarToolbar {
...
+    protected renderItemWithDraggableWrapper(item: TabBarToolbarItem, position: ToolbarItemPosition): React.ReactNode {
         const stringifiedPosition = JSON.stringify(position);
         let toolbarItemClassNames = '';
-        let renderBody: React.ReactNode;
+        const renderBody = item.render(this);
 
-        if (!ReactTabBarToolbarItem.is(item)) {
-            toolbarItemClassNames = TabBarToolbar.Styles.TAB_BAR_TOOLBAR_ITEM;
-            if (this.evaluateWhenClause(item.when)) {
-                toolbarItemClassNames += ' enabled';
```
dhuebner, line 245: Without adding 'enabled',  toolbar items are rendered as disabled

tsmaeder: That is the responsibility of the toolbar item.

tsmaeder:
@dhuebner rigth now, I'm mostly looking for functional and architectural feedback. I can fix the tests once the dust settles.

dhuebner:
@tsmaeder 
> What are you trying to achieve through CompoundMenuNodeRole.flat?

The previous effect was that if you click on three dots toolbar item in notebook cell  the children of the `more` sub-menu were rendered directly (see code before).  Now an empty composite menu is rendered and the children only if you click on it.

packages/toolbar/src/browser/toolbar.tsx
dhuebner: Then we are missing some implementation, cause the test is failing because `workbench.action.showCommands` toolbar item is not enabled.

dhuebner:
@tsmaeder 
I've pushed fixes for Notebook and Preferences tests. The toolbar test is still failing because of missing `enabled` css class, see comments above.
Overall it seems to be a huge impact change that might break adopters.

packages/test/src/browser/view/test-view-contribution.ts
```diff
@@ -274,6 +274,14 @@ export class TestViewContribution extends AbstractViewContribution<TestTreeWidge
...
             order: 'aaaaa'
         });
+
+        menus.registerSubmenu([...PLUGIN_TEST_VIEW_TITLE_MENU, TestViewCommands.RUN_ALL_TESTS.id], '', undefined, undefined, undefined, {
+            'testing.profile.context.group': 'run'
+        });
+
+        menus.registerSubmenu([...PLUGIN_TEST_VIEW_TITLE_MENU, TestViewCommands.DEBUG_ALL_TESTS.id], '', undefined, undefined, undefined, {
+            'testing.profile.context.group': 'debug'
+        });
```
msujew, line 284: Having `undefined, undefined, undefined` here is a good point to support my argument in favor of objects :)

msujew (commented):
Looks generally really good. I believe we can improve on the API. Feel free to request my review again once the tests are fixed and I'll do a more in-depth code review.

packages/core/src/common/menu/menu-model-registry.ts
```diff
@@ -297,78 +289,40 @@ export class MenuModelRegistry {
...
      */
-    removeSingleRootNode(fullMenuModel: MutableCompoundMenuNode, menuPath: MenuPath): CompoundMenuNode {
-        // check whether all children are compound menus and that there is only one child that has further children
-        if (!this.allChildrenCompound(fullMenuModel.children)) {
-            return fullMenuModel;
-        }
-        let nonEmptyNode = undefined;
+    static removeSingleRootNode(fullMenuModel: CompoundMenuNode): CompoundMenuNode {
+
+        let singleChild = undefined;
```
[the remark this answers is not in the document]
tsmaeder: Yet the compiler can tell that the method returns the correct type. What good is type inference if explicitly type all variable anyway?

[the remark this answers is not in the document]
msujew: Nvm, I guess you're right. TypeScript seems to have become smart enough to infer this correctly.

JonasHelming:
@tsmaeder What is the current state of this PR?

tsmaeder:
There are a couple of issues that need to be addressed before merging. Since it's a major refactoring, it's not a good idea to merge before the release anyway.

packages/core/src/electron-browser/menu/electron-main-menu-factory.ts
```diff
@@ -71,11 +66,28 @@ export type ElectronMenuItemRole = ('undo' | 'redo' | 'cut' | 'copy' | 'paste' |
     'selectNextTab' | 'selectPreviousTab' | 'mergeAllWindows' | 'clearRecentDocuments' |
     'moveTabToNewWindow' | 'windowMenu');
 
+function traverseMenuDto(items: MenuDto[], callback: (item: MenuDto) => void): void {
```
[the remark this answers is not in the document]
tsmaeder: I don't think there is a use-case for the traverse functions to be polymorphic because they are simple algorithms. If anything should be overridable it should be what happens when a preference changes. But that is tightly coupled to the mechanisms in this class, so I don't think overriding parts of it makes much sense.

tsmaeder:
@msujew ping if we want to merge it this release cycle ;-)

doc/Migration.md
```diff
@@ -59,6 +59,14 @@ For example:
 }
 ```
 
+### v1.59.0
```
msujew, line 62: Minor: Updated for 1.61.0?

packages/notebook/src/browser/service/notebook-context-manager.ts
```diff
@@ -43,9 +43,6 @@ export class NotebookContextManager {
 
     protected readonly toDispose = new DisposableCollection();
 
-    protected readonly onDidChangeContextEmitter = new Emitter<ContextKeyChangeEvent>();
-    readonly onDidChangeContext = this.onDidChangeContextEmitter.event;
```
msujew, line 47: Question: (why) are these no longer necessary?

msujew (approved):
Looks pretty good to me. Just minor comments, see below.

PR #14681, opened 2024-12-29
Consistently handle OpenAI models not supporting system messages instead of using dedicated O1 chat agent

planger:
#### What it does

* Consistently handle OpenAI LLMs that don't support system messages in OpenAI model provider
* Remove O1-specific chat agent as it isn't needed anymore
* Update to latest OpenAI library
* Change "system" to "developer" message role (https://cdn.openai.com/spec/model-spec-2024-05-08.html#follow-the-chain-of-command)
* Add 'o1-mini' model

Fixes https://github.com/eclipse-theia/theia/issues/14686

#### How to test

* Test OpenAI models still work as before, including their system message.
* Test selecting o1-models for normal chat agents to confirm they now work (except they don't support streaming and thus no tool calling), but there is no need for a separate agent anymore.

#### Follow-ups

None

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


JonasHelming:
Changes look great, thank you!
I would explicitly mention that the o1 agent is removed and because of this, I would be in favor of removing the "chore" and adding an issue for this PR to document this changes, as it is potentially breaking for the user.

JonasHelming (approved):

JonasHelming:
One more thing: I think the checkmarks are set wrong, it does not introduce breaking changes 8the changelog is not updated), but I guess you have reviewed it :-)

Issue #13946, opened 2024-07-24
"vscode.TreeItemCollapsible" Never Changes


rroessler:
### Bug Description:

In VSC extensions, whenever a refresh is requested by a `vscode.TreeDataProvider`, trees are refreshed to an "initialized" collapsible state instead of maintaining their current collapsible state.

This is an issue for tree-views that show real-time data (eg: custom debugger panels) as refreshes will force all nodes to snap back to their original collapsible state.

### Steps to Reproduce:

This issue can be reproduced by implementing a simple tree-view provider that refreshes every 5-seconds. If the root-node is clicked, anytime the interval fires the root-node will reset to the original collapsed state as opposed to adhering to the current expanded state.

```typescript
import vscode from 'vscode';

class Provider implements vscode.TreeDataProvider {
    private m_root = new vscode.TreeItem('Root', vscode.TreeItemCollapsibleState.Collapsed);
    private m_children = Array.from(Array(5), (_, ii) => new vscode.TreeItem(`Node: ${ii}`));

    readonly emitter = new vscode.EventEmitter<vscode.TreeItem | vscode.TreeItem[] | undefined | null | void>();
    readonly onDidChangeTreeData = this.emitter.event.bind(this.emitter);

    getTreeItem(element: vscode.TreeItem): vscode.TreeItem {
        return element;
    }

    getChildren(element?: vscode.TreeItem | undefined): vscode.ProviderResult<vscode.TreeItem[]> {
        if (typeof element === 'undefined') return [this.m_root];
        return element === this.m_root ? this.m_children : [];
    }
}

export function activate(context: vscode.ExtensionContext) {
    const treeDataProvider = new Provider();
    context.subscriptions.push(vscode.window.createTreeView('exampleTreeProvider', { treeDataProvider }));

    // set an interval to trigger refreshes every 5 seconds
    const interval = setInterval(() => treeDataProvider.emitter.fire(), 5000);
    context.subscriptions.push(() => clearInterval(interval));
}
```

rroessler:
Alongside this, triggering refreshes via "onDidChangeTreeData" causes the current scroll-state to reset to an initial position.

tsmaeder:
@rroessler do you know what the behavior in VS Code is?

rroessler:
[...]

PR #14697, opened 2025-01-06
Only call refresh on given elements given by plugin

### LINKED ISSUE #14390: In TheiaIDE the "Plugins" items are not unwrap in a multi-module MAVEN JAVA project
### Bug Description:
In TheiaIDE the "Plugins" items are not unwrap in a multi-module MAVEN JAVA project
Section MAVEN has items
```
    Parent-project
    Lifecycle
    Plugins
    Dependencies
    Favorites
    Profiles
    Sub-project
        Lifecycle
        Plugins
        Dependencies
        Favorites
        Profiles
```
If you try to open "Plugins", all items collapse to the root. Item "Plugins" not work in Parent-project and Sub-project. Other items unwrap normally.

### Steps to Reproduce:

1. Create a multi-module maven project
2. Try to expand the Plugins item

### Additional Information

- Operating System: macos Sequoia 15.1 (it didn't work on the previous version either)
- Theia Version: 1.54.0.165

tsmaeder:
#### What it does
In the scenario from the issue, the extensions calls `onDidChangeTreeData?: [Event]<void | T | T[]>` on some elements. However, Theia disregards the elements to be refreshed and always refreshes from the root. Since the `TreeItem` instances furnished by the extensions seem to be new each time, we cannot associate the refreshed items with cached tree nodes and therefore we lose the expansion state: for us, it looks like we have a completely new tree that starts out collapsed.
This PR implements the refresh of only a given elemement. This prevents the refresh from the root and thus fixes the problem.

Fixes #14390

#### How to test
Make sure other contributed views still work the same as before, for example views from gitlens, reference, etc. 

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers


msujew:
@tsmaeder Does this also fix https://github.com/eclipse-theia/theia/issues/13946?

packages/plugin-ext/src/main/browser/view/tree-view-widget.tsx
```diff
@@ -428,6 +428,18 @@ export class PluginTreeModel extends TreeModelImpl {
 
 @injectable()
 export class TreeViewWidget extends TreeViewWelcomeWidget {
+    async refresh(items: string[] | undefined): Promise<void> {
```
msujew, line 431: Minor: It might make more sense to use `?` here, as I would much rather call `tree.refresh()` than `tree.refresh(undefined)`.
```suggestion
    async refresh(items?: string[]): Promise<void> {
```

packages/plugin-ext/src/plugin/tree/tree-views.ts
```diff
@@ -252,8 +252,20 @@ class TreeViewExtImpl<T> implements Disposable {
...
+            if (!Array.isArray(elements)) {
+                elements = [elements];
+            }
+            for (const element of elements) {
+                for (const node of this.nodes.values()) {
+                    if (node.value === element) {
+                        ids.push(node.id);
+                    }
+                }
+            }
```
msujew, line 267: Suggestion: I'm not sure how bad this can get in reality, but this is `O(n * m)` behavior. We can cut this down to `O(n + m)`.
```suggestion
            const set = new Set<T>();
            for (const element of elements) {
                set.add(element);
            }
            for (const node of this.nodes.values()) {
                if (set.has(node.value)) {
                    ids.push(node.id);
                }
            }
```

msujew (approved):
Looks good to me. I've tried it out with the maven plugin, and everything seems to work perfectly. No regressions as far as I could tell as well.

JonasHelming:
@tsmaeder Can we get this merged?

tsmaeder:
@JonasHelming I want to apply the reasonable suggestions from @msujew, first, though. Currently busy with other stuff.

PR #14713, opened 2025-01-10
fix: improve cancel logic in openAi model

eneufeld:
#### What it does

The cancel of a request was not working correctly. With the changes the cancelToken is better taken into account.

#### How to test

Try using the inline completion of the IDE AI Feature. This leads to a very slow responsiveness of the ui as the underlying language model was continuing creating responses.
With the fix this is not the case anymore for the OpenAI Model.

#### Follow-ups

Improve cancel handling of all other language models.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


sdirix (changes requested):
I did not notice any performance issues with this PR.

However something in the cancel/error flow does not work properly yet. In case a cancel occurs, the console logs the error two times:

>root ERROR Uncaught Exception:  Error: Canceled
>root ERROR Error: Canceled

I think we should:
- check for cancel errors and not log them. Maybe leave a log message on the `DEBUG` log level
- make sure the errors are caught and handled appropriately

At the moment, once the error occurs, the code completion progress message `Calculating AI code completion...` is shown forever with the progress never canceling. Likely because the error was not caught.

packages/ai-core/src/browser/frontend-language-model-registry.ts
```diff
@@ -286,6 +292,14 @@ export class FrontendLanguageModelRegistryImpl
         throw new Error(`Could not find a tool for ${toolId}!`);
     }
 
+    error(id: string, error: Error): void {
+        if (!this.streams.has(id)) {
+            throw new Error('Somehow we got a callback for a non existing stream!');
+        }
+        const streamState = this.streams.get(id)!;
+        streamState.reject?.(error);
+    }
```
sdirix, line 301: In the `send` callback which is also called from the backend we construct the stream state if it does not exist yet. I would suggest to do the same here.

If the first feedback of a stream is an error it could happen that the error arrives here first before the corresponding stream state was created as we don't control the async flows in the backend. We don't want to throw an error in that case but just handle it like all the other cases too.

packages/ai-core/src/browser/frontend-language-model-registry.ts
```diff
@@ -286,6 +292,14 @@ export class FrontendLanguageModelRegistryImpl
         throw new Error(`Could not find a tool for ${toolId}!`);
     }
 
+    error(id: string, error: Error): void {
```
sdirix, line 295: please add the same comment here as for `send` and `toolCall`

sdirix (commented):
Works fine! I only have one small suggestion for an edge case

sdirix (approved):
Thanks!

PR #14716, opened 2025-01-13
Feature - Chat Agent Pinning

atahankilc:
#### What it does

An "agent pinning" feature has been introduced to enhance the user experience in consecutive agent usage within the chat interface.

- Automatic Pinning:
If a chat agent is mentioned in a prompt and there is no prior pinned agent, the mentioned agent is automatically pinned. This eliminates the need for the user to repeatedly mention the agent in subsequent prompts.

- Handling New Mentions:
If a different agent is mentioned while another agent is pinned, the pinned agent remains unchanged. The newly mentioned agent is used only for the specific prompt where it is mentioned, without affecting the pinned agent.

- Manual Control:
Users can manually unpin the agent through the chat interface if desired.

- New Session with Pinned Agent:
A new chat session can now be initiated directly with a pinned agent.

#### How to test

- Mention an agent in chat interface
- Execute AI_CHAT_NEW_CHAT_WINDOW_WITH_PINNED_AGENT_COMMAND with passing desired chat agent

#### Follow-ups

As part of this contribution, We encountered an issue with the added command `ai-chat-ui.new-chat-with-pinned-agent`. The command is expected to take a single argument of type ChatAgent | undefined. However, the function call fails when the argument is passed directly in the expected form. To make the command work, We had to pass an argument array instead, where the first item in the array is undefined, followed by the actual ChatAgent value.

This behavior seems to deviate from the intended design. While this workaround resolves the issue for now, the underlying cause still needs to be investigated. The error message encountered when not using this approach is:

`Error: The command 'ai-chat-ui.new-chat-with-pinned-agent' cannot be executed. There are no active handlers available for the command.`

We are including this as a follow-up issue to address later and investigate why the command handler does not behave as expected when a single argument is passed directly.

#### Attribution

This feature was developed by @atahankilc, @ZidongWang123, @TANG839, and @mihaela21k as part of the AI Coding Exercises project in the Practical Course Software Engineering for Business Information Systems at Technical University of Munich.

#### Review checklist

#### Reminder for reviewers


JonasHelming:
@atahankilc Could you fix the linting:
@theia/ai-chat-ui: /home/runner/work/theia/theia/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
@theia/ai-chat-ui:   284:1  error  Trailing spaces not allowed  no-trailing-spaces

packages/ai-chat-ui/src/browser/style/index.css
planger, line 1: There are a few unnecessary reformattings in this file, e.g.

```diff
- div:last-child > .theia-ChatNode {
+ div:last-child>.theia-ChatNode {
```

and added lines.

Please remove them as they add unnecessary noise and potential for conflicts.

packages/ai-chat/src/common/chat-service.ts
```diff
@@ -160,8 +163,14 @@ export class ChatServiceImpl implements ChatService {
         session.title = request.text;
 
         const parsedRequest = this.chatRequestParser.parseChatRequest(request, session.model.location);
+        let agent = this.getAgent(parsedRequest);
+
+        if (!session.pinnedAgent && agent && agent.id !== OrchestratorChatAgentId) {
+            session.pinnedAgent = agent;
+        } else if (session.pinnedAgent && this.getMentionedAgent(parsedRequest) === undefined) {
+            agent = session.pinnedAgent;
+        }
```
planger, line 172: I find it a bit concerning that we add a dependency to a specific agent from the chat service. It is not super critical as it is "just" the ID, but in general the chat service should not need to be aware of any specific agents. Also the agents may be moved to another package at some point.

Is there a way to avoid that?
Couldn't we rather check if it is the "default agent" in the first if condition:
If we don't have a pinned agent yet and the agent is not the default agent, then we pin the agent.
Alternatively, we could use the parsed request and check if an agent is explicitly mentioned and then have the condition: if there is an agent mentioned in the parsed request and we don't have a pinned agent yet, then we pin it.

planger (changes requested):
Very nice, thank you for your contribution! I only have one main concern, the rest of my comments inline are rather minor.

packages/ai-chat/src/common/chat-service.ts
atahankilc: Thank you for your review. If statement is changed to check default agent rather than orchestrator.

planger:
As the UI of the chat input has changed quite a bit since this PR started, I'd suggest to integrate the pinned agent as sketched in the following screenshot:

image

We have room below the text input for additional controls and information. For instance, we use a + icon to add context in [another PR](https://github.com/eclipse-theia/theia/pull/14787). I think it'd be great to use this room to put the agent pinning UI there as well, in the form of a @ button to e.g. pin or unpin an agent, and showing the currently pinned agent, if any, next to it.

Buttons shown at this place are currently controlled here in the code:
https://github.com/eclipse-theia/theia/blob/master/packages/ai-chat-ui/src/browser/chat-input-widget.tsx#L322

We may need to extend this a bit to allow also showing an optional text left to the button.

This avoids also any interference with the changesets that may be shown above the chat input:

image

What do you think?

atahankilc:
Hey @planger. Thank you for your guidance. I tried to refactor the pinning UI and related items. I've attached the updated UI to this comment. There's one thing I couldn't manage to sort out and wanted to get your opinion on. Our implementation does not include agent pinning from the input widget component, so I tried to implement it in a way that the "@" icon is displayed by default, which would trigger agent completion pop-up when clicked. However, I got a bit tangled up in the process. As a result, the option is currently only shown if there is a pinned agent, and clicking it only allows unpinning. With this said, is there anything I could do to trigger agent suggestions? And if not, should I implement a separate UI for selecting an agent?

Screenshot 2025-02-03 at 03 10 15

packages/ai-chat-ui/src/browser/style/index.css
```diff
@@ -280,6 +280,23 @@ div:last-child > .theia-ChatNode {
...
+  padding-right: 10px;
+  padding-bottom: 11px;
+  display: flex;
+  flex-direction: row;
+  align-items: start;
+  align-self: flex-end;
+  gap: 10px;
+  border: var(--theia-border-width) solid var(--theia-dropdown-border);
+  border-radius: 4px;
+}
```
planger, line 298: I think this isn't needed anymore, right?

packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
```diff
@@ -79,6 +79,12 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
             isEnabled: widget => this.withWidget(widget, () => true),
             isVisible: widget => this.withWidget(widget, () => true),
         });
+        registry.registerCommand(AI_CHAT_NEW_CHAT_WINDOW_WITH_PINNED_AGENT_COMMAND, {
+            // TODO - not working if function arg is set to type ChatAgent | undefined ?
+            execute: (...args: unknown[]) => this.chatService.createSession(ChatAgentLocation.Panel, {focus: true}, args[1] as ChatAgent | undefined),
+            isEnabled: widget => this.withWidget(widget, () => true),
+            isVisible: widget => this.withWidget(widget, () => true),
+        });
```
planger, line 87: Currently this command is never invoked anywhere

packages/ai-chat-ui/src/browser/chat-view-commands.ts
```diff
@@ -39,6 +39,11 @@ export const AI_CHAT_NEW_CHAT_WINDOW_COMMAND: Command = {
     iconClass: codicon('add')
 };
 
+export const AI_CHAT_NEW_CHAT_WINDOW_WITH_PINNED_AGENT_COMMAND: Command = {
+    id: 'ai-chat-ui.new-chat-with-pinned-agent',
+    iconClass: codicon('add')
+};
+
```
planger, line 46: Not really used at the moment.

packages/ai-chat-ui/src/browser/chat-view-widget.tsx
```diff
@@ -113,10 +115,12 @@ export class ChatViewWidget extends BaseWidget implements ExtractableWidget, Sta
         this.toDispose.push(
             this.chatService.onActiveSessionChanged(event => {
                 const session = event.sessionId ? this.chatService.getSession(event.sessionId) : this.chatService.createSession();
+
```
planger, line 118: Unnecessary empty line.

packages/ai-chat/src/common/chat-service.ts
```diff
@@ -127,12 +128,13 @@ export class ChatServiceImpl implements ChatService {
...
 
-    createSession(location = ChatAgentLocation.Panel, options?: SessionOptions): ChatSession {
+    createSession(location = ChatAgentLocation.Panel, options?: SessionOptions, pinnedAgent?: ChatAgent): ChatSession {
         const model = new ChatModelImpl(location);
         const session: ChatSessionInternal = {
             id: model.id,
             model,
-            isActive: true
+            isActive: true,
+            pinnedAgent: pinnedAgent
```
planger, line 137:
```suggestion
            pinnedAgent
```

planger (commented):
Thank you very much for this iteration! I think this already feels pretty nice. I'm not sure we really need an explicit pin action though. I think we have two options for pinning agents:

1. Show a context menu on click and offer "Unpin" if pinned, and "Pin > <list-of-chat-agents>".
2. Trigger the auto completion for agents as you mentioned.

I personally prefer option 2. To trigger the code completion, I think a handler like this should work (if pinnedAgent === undefined):

```ts
handler: () => {
                    if (editorRef.current) {
                        editorRef.current.getControl().getModel()?.applyEdits([{
                            range: {
                                startLineNumber: 1,
                                startColumn: 1,
                                endLineNumber: 1,
                                endColumn: 1
                            },
                            text: '@ ',
                        }]);
                        editorRef.current.getControl().setPosition({ lineNumber: 1, column: 2 });
                        editorRef.current.getControl().getAction('editor.action.triggerSuggest')?.run();
                    }
                }

planger:
Oh, and one more comment. I think we should ensure to show the at-icon only if more than one chat agent is registered -- and make it configurable similar to the `showContext`:
https://github.com/eclipse-theia/theia/blob/master/packages/ai-chat-ui/src/browser/ai-chat-ui-frontend-module.ts#L64

Because we do have adopters who only provide one chat agent and don't provide the option for users to switch between chat agents.

packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
atahankilc: For the extension we are developing for our course, we are using this command to directly open a chat widget with an agent pinned. If you don't have such a functionality expectation, I can remove it.

atahankilc:
> Oh, and one more comment. I think we should ensure to show the at-icon only if more than one chat agent is registered -- and make it configurable similar to the `showContext`: https://github.com/eclipse-theia/theia/blob/master/packages/ai-chat-ui/src/browser/ai-chat-ui-frontend-module.ts#L64
> 
> Because we do have adopters who only provide one chat agent and don't provide the option for users to switch between chat agents.

I have added a setting option under ai features to dis/enable this pinning functionality and also made it configurable similar to the showContext, however, I couldn't directly link the preference in the settings to the frontend module. If you find it appropriate, I can write a class/function similar to AIActivationService to enable or disable pinning button in chat widget. This way, I believe I can retrieve the number of currently registered agents from the ChatAgentService or reflect the setting value by getting it from PreferenceService.

packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
planger: Yeah, I don't think we need this in Theia at the moment.

packages/ai-chat-ui/src/browser/chat-view-commands.ts
planger: For now, I'd remove this.

planger (commented):
Thank you very much for the revision! I think we are almost there. Can you please remove the comment and rebase it, there are a few conflicts. But then this is imho ready.

Thank you very much for this nice contribution!

atahankilc:
Thanks a lot for your feedbacks. I have removed the code parts you mentioned in your comments and rebased branch. I noticed that DefaultChatAgentId binding was removed from ai-chat-frontend-module.ts, and if PinChatAgent also needs to be removed, I can remove it.. It should be all set for another review. Let me know if anything else needs tweaking.

Appreciate it!

atahankilc:
I have noticed a bug in chat input widget and wanted to share with you:

The height of the chat input box kept increasing continuously, and it seemed to be fixed when I set the height in the theia-ChatInput-Editor style to a fixed value.

https://github.com/user-attachments/assets/c64efcda-90ab-4169-ad85-443d8728f447

planger:
@atahankilc Thanks, yes there seems to be a rather severe bug on `master` in the chat input at the moment. I'll wait for it to be resolved and then review this PR rebased on the hopefully soon fixed `master`. Otherwise it is hard to verify the final behavior.

Also, I expect a few conflicts with https://github.com/eclipse-theia/theia/pull/14787. So I may opt to merge the other one first and then I'll take over rebasing this one and resolving their conflicts, because this change is smaller, it is probably easier this way.

Code looks already good so far! Thank you!

planger (approved):
I cherry-picked https://github.com/eclipse-theia/theia/pull/14942 and retested this. I think it works great now, thank you very much for your consistent work on this nice feature!

We could already merge this PR now and rebase https://github.com/eclipse-theia/theia/pull/14787 --- I expect several conflicts though. Or we do it the other way around, which might be easier. I'll give it a try locally.

planger:
Conflicts are straight forward to resolve, so I'll rebase https://github.com/eclipse-theia/theia/pull/14787

Issue #14714, opened 2025-01-11
Cannot create projects with swift extension


suvnoque:
### Bug Description:
Cannot create new project with the swift extension

### Steps to Reproduce:

1. Install Swift extension
2. Run command "Swift: Create New Project..."
3. Error message appears

### Additional Information
![image](https://github.com/user-attachments/assets/b1192234-51b8-4580-99bc-54982f511c48)

- Operating System: Windows 10 Home
- Theia Version: 1.57.100

JonasHelming:
@tsmaeder

tsmaeder:
That looks like the extension is trying to load something from VS Code's `node-modules` folder. We don't have such a folder since we are webpack-bundling our back end. It looks like the code in question is https://github.com/swiftlang/vscode-swift/blob/main/src/utilities/native.ts There's also https://github.com/microsoft/node-pty/issues/582 and the discussion in https://github.com/microsoft/vscode/issues/658. T.b.h, I don't really know what we could do except trying to reproduce the exact environment VS Code provides, but even the MS devs warn that this environment is not stable.

tsmaeder:
I think @msujew knows a bit about platform-specific extensions. Maybe he has an idea?

sdirix:
We have a similar issue for another extension: https://github.com/eclipse-theia/theia/discussions/14638

i.e. some extensions expect some binaries in specific places in the delivery

msujew:
Note that we had the same discussion @tsmaeder in https://github.com/eclipse-theia/theia/issues/13779. We definitely could reproduce the VS Code environment (sans the exact version of node-pty). It shouldn't be too much work, but it's rather a question of whether we want to support that.

tsmaeder:
I am just wary of setting up another regular task where we have to track the node_modules folder in VS code. Maybe we just should set up some "common library" of node modules that we make available in the plugin host. The trouble is then if folks rely on behavior of a particular version (e.g. node-pty has _many_ preview versions). The same will happen to extensions as VS code updates their versions of packaged node modules, though.

sdirix:
I think a low effort approach to place some of our dependencies/binaries in the same location as VS Code does, makes sense. Just to be a bit more compatible in a pragmatic fashion.

I completely [agree with @tsmaeder](https://github.com/eclipse-theia/theia/issues/13779#issuecomment-2162287986) that we should not try to replicate VS Code's environment 100% as this will require a lot of work and is not encouraged anyway.

[...]

Issue #13779, opened 2024-06-06
Google Cloud extension fails to load.


kineticsquid:
### Bug Description:

I use the Google Cloud extension on VS. I tried it on the latest Theia release and it fails on loading:

<img width="527" alt="Screenshot 2024-06-06 at 12 59 26 PM" src="https://github.com/eclipse-theia/theia/assets/681707/2a251707-7dbb-454d-8739-56eeab8d06bd">

I confirmed the latest release of the extension on my VS installation and then tried various combinations of un-installing, re-instalilng and restarting Theia, but no luck.

### Additional Information

- Operating System: Mac OSX 14.4.1
- Theia Version: Theia IDE Version 1.49.101 (1.49.101.151)

msujew:
It's likely that the extension is just using an unbundled `require('node-pty')` call and expects that vscode just makes the library available in the extension host. There is a bit of discussion surrounding this practice and it's discouraged by the maintainer of node-pty (who is a major vscode contributor), see https://github.com/microsoft/node-pty/issues/582.

Anyway, an obvious fix to this would be to expose the `node-pty` library in our plugin host. As far as I know, vscode is doing this by bundling their app in a smart way, while we just bundle everything, which makes plain node.js `require` calls fail.

kineticsquid:
@msujew Thanks for the explanation. To make sure I understand, this is a change you would make to expose the library?

msujew:
@kineticsquid We could do that, yes. It's not ideal, since it's discouraged, but it's not like VSCode actually punishes extension developers that just side-load vscode's owns dependencies.

msujew:
@tsmaeder Your opinion on this? VS Code exposes its own native dependencies to its extensions. Doing the same would require us to keep our versions of these dependencies up to date with VS Code.

tsmaeder:
@kineticsquid do you mean this one? https://marketplace.visualstudio.com/items?itemName=GoogleCloudTools.cloudcode

@msujew @kineticsquid I cannot express in polite conversation just what I think of the practice of depending on the implementation details of your host platform if said platform goes out of its way to provide a well-defined API for interacting with that platform. 

As for aligning our dependencies with VS Code: that way lies madness! The Theia platform is implemented differently from VS Code and therefore necessarily has different dependencies and dependency versions. If we start doing this, we would also make the integration work of our adopters infinitely more difficult. So as far as I am concerned, it's a hard no!

[...]

PR #14720, opened 2025-01-13
Provide node-pty package for plugins

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/14714
Closes https://github.com/eclipse-theia/theia/issues/13779

Adds a further override of the `require` function to the plugin host. In case any `node-pty` package is required, it will simply return the node-pty package that is packaged with Theia.

#### How to test

Run the steps in https://github.com/eclipse-theia/theia/issues/14714 and see that you don't get the `require` error.

#### Follow-ups

We could add further dependencies in there.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/hosted/node/plugin-require-override.ts
+++ b/packages/plugin-ext/src/hosted/node/plugin-require-override.ts
@@ -35,11 +35,19 @@ export function overridePluginDependencies(): void {
     const node_module = require('module');
     const original = node_module._load;
     node_module._load = function (request: string): unknown {
-        for (const filter of overrides) {
-            if (request === filter.package || request.endsWith(`node_modules/${filter.package}`) || request.endsWith(`node_modules\\${filter.package}`)) {
-                return filter.module;
+        try {
+            // Attempt to load the original module
+            // In some cases VS Code extensions will come with their own `node_modules` folder
+            return original.apply(this, arguments);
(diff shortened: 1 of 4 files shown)


packages/plugin-ext/src/hosted/node/plugin-require-override.ts
```diff
@@ -0,0 +1,45 @@
...
+ * See also:
+ * https://github.com/eclipse-theia/theia/issues/14714
+ * https://github.com/eclipse-theia/theia/issues/13779
+ */
+export function overridePluginDependencies(): void {
+    const node_module = require('module');
+    const original = node_module._load;
+    node_module._load = function (request: string): unknown {
+        for (const filter of overrides) {
+            if (request === filter.package || request.endsWith(`node_modules/${filter.package}`) || request.endsWith(`node_modules\\${filter.package}`)) {
```
tsmaeder, line 39: So if someone actually went to the trouble of packaging their own node-pty with the plugins, would our version override the packaged version? Is this what we want or should we first check if module in question can be loaded normally?

msujew: I guess, yes. It probably makes sense to check for it first, yes.

tsmaeder (approved):

PR #14722, opened 2025-01-13
fix: Support Azure OpenAI

### LINKED ISSUE #14711: [Theia AI] Support Azure OpenAI
### Feature Description:

We host an OpenAI gpt-4o model in Azure and wanted to connect Theia AI via **OpenAI Compatible Models** as described here https://theia-ide.org/docs/user_ai/#openai-compatible-models-eg-via-vllm

```json
{
    "ai-features.AiEnable.enableAI": true,
    "ai-features.openAiCustom.customOpenAiModels": [
        {
            "model": "gpt-4o",
            "url": "https://<custom>.openai.azure.com/openai/deployments/<my_deployment>/chat/completions?api-version=2024-08-01-preview",
            "id": "my_model",
            "apiKey": "<my_api_key>"
        }
    ],
        "Universal": {
            "languageModelRequirements": [
                {
                    "purpose": "chat",
                    "identifier": "my_model"
                }
            ]
        },
        "Orchestrator": {
            "languageModelRequirements": [
                {
                    "purpose": "agent-selection",
                    "identifier": "my_model"
                }
            ]
        }
    }
}
```

This doesn't work. I always get

```
{"error":{"code":"404","message": "Resource not found"}}
```

After searching around for the cause, we noticed that the issue is related to OpenAI on Azure. This is shortly described in [openai-node GitHub](https://github.com/openai/openai-node?tab=readme-ov-file#microsoft-azure-openai)

We created a small node project to verify this. When using the `OpenAI` class in our small reproducer, we also get the `404` error when trying to access the Azure OpenAI resource. Switching to the `AzureOpenAI` class in the reproducer, the issue is gone and we get a correct response.

While looking into the Theia sources, I think we narrowed the issue down to [OpenAiModel](https://github.com/eclipse-theia/theia/blob/master/packages/ai-openai/src/node/openai-language-model.ts).

Either there needs to be some configuration to specify that Azure OpenAI is tried to access, and based on that use either `OpenAI` or `AzureOpenAI`. But that also means that the result needs to be parsed differently as far as we saw. Another option would be a new _ai-azureopenai_ module. But I am not sure if this would be too much of copied code.

fipro78:
* Add configuration of OpenAI API version
* Add configuration of supportsDeveloperMessage
* If OpenAI API version is configured, use AzureOpenAI

Fixes #14711

#### What it does

It fixes #14711 and adds support for accessing an OpenAI model deployed on Azure.
To make this work, the `AzureOpenAI` class needs to be used, as described in the [openai-node docs](https://github.com/openai/openai-node?tab=readme-ov-file#microsoft-azure-openai).

To build up the request with `AzureOpenAI` you need to pass a `apiVersion`. This PR therefore adds the option to configure the `apiVersion` in the settings. If no `apiVersion` is configured, the default `OpenAI` object is used for initialization.

I also noticed that it can happen, that the OpenAI model version deployed on Azure might not yet support the developer role. Therefore I also added the option to configure whether the developer role is supported or not, which defaults to `true`.

#### How to test

The reviewer would need to have an OpenAI model available in Azure and then configure a custom openai model like this:

```json
{
  "ai-features.AiEnable.enableAI": true,
  "ai-features.openAiCustom.customOpenAiModels": [
    {
      "model": "gpt4o",
      "url": "<AZURE_OPENAI_API_BASE_URL>",
      "id": "azure-deployment",
      "apiKey": "<AZURE_OPENAI_API_KEY>",
      "apiVersion": "<AZURE_OPENAI_API_VERSION>",
      "supportsDeveloperMessage": false
    }
  ],
    "Universal": {
      "languageModelRequirements": [
        {
          "purpose": "chat",
          "identifier": "azure-deployment"
        }
      ]
    },
    "Orchestrator": {
      "languageModelRequirements": [
        {
          "purpose": "agent-selection",
          "identifier": "azure-deployment"
        }
      ]
    }
  }
}
```

Of course a non Azure model deployment should still work as before.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-openai/README.md
+++ b/packages/ai-openai/README.md
@@ -27,12 +27,66 @@ You can configure the end points via the `ai-features.openAiCustom.customOpenAiM
     url: string
     id?: string
     apiKey?: string | true
+    apiVersion?: string | true
+    supportsDeveloperMessage?: boolean
+    enableStreaming?: boolean
 }
 ```
 
 - `model` and `url` are mandatory attributes, indicating the end point and model to use
 - `id` is an optional attribute which is used in the UI to refer to this configuration
 - `apiKey` is either the key to access the API served at the given URL or `true` to use the global OpenAI API key. If not given 'no-key' will be used.
+- `apiVersion` is either the api version to access the API served at the given URL in Azure or `true` to use the global OpenAI API version.
+- `supportsDeveloperMessage` is a flag that indicates whether the model supports the `developer` role or not. `true` by default.
+- `enableStreaming` is a flag that indicates whether the streaming API shall be used or not. `true` by default.
+
+### Azure OpenAI
+
+To use a custom OpenAI model hosted on Azure, the `AzureOpenAI` class needs to be used, as described in the 
+[openai-node docs](https://github.com/openai/openai-node?tab=readme-ov-file#microsoft-azure-openai).
+
+Requests to an OpenAI model hosted on Azure need an `apiVersion`. To configure a custom OpenAI model in Theia you therefore need to configure the `apiVersion` with the end point.
+Note that if you don't configure an `apiVersion`, the default `OpenAI` object is used for initialization and a connection to an Azure hosted OpenAI model will fail.
+
+An OpenAI model version deployed on Azure might not support the `developer` role. In that case it is possible to configure whether the `developer` role is supported or not via the 
+`supportsDeveloperMessage` option, which defaults to `true`.
+
+The following snippet shows a possible configuration to access an OpenAI model hosted on Azure. The `AZURE_OPENAI_API_BASE_URL` needs to be given without the `/chat/completions` 
+path and without the `api-version` parameter, e.g. _`https://<my_prefix>.openai.azure.com/openai/deployments/<my_deployment>`_
+
+```json
+{
+  "ai-features.AiEnable.enableAI": true,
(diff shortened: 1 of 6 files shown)


planger (approved):
Thank you very much @fipro78!

Code looks good to me and existing behavior still seems to work fine. I can't test with Azure though, as I don't have an account at hand at the moment.

Can you please also add a short hint on how to configure/use this to the README.md of this package (like you did in the description of this PR)?

fipro78:
@planger 
Thanks for the review. I updated the readme.

JonasHelming:
@planger I guess we can merge this then?

packages/ai-openai/README.md
```diff
@@ -27,12 +27,66 @@ You can configure the end points via the `ai-features.openAiCustom.customOpenAiM
     url: string
     id?: string
     apiKey?: string | true
+    apiVersion?: string | true
+    supportsDeveloperMessage: boolean
+    enableStreaming: boolean
```
sdirix, line 32: If these are optional then we should mark them with question marks
```suggestion
    supportsDeveloperMessage?: boolean
    enableStreaming?: boolean
```

sdirix (approved):
Code looks good to me! I don't have an Azure key, so I did not test the new functionality however I did verify that the regular Open AI models continue working.

PR #14736, opened 2025-01-16
Allow to auto-start MCP servers on frontend start-up

### LINKED ISSUE #14692: [Theia AI] Allow to auto-start MCP servers on start-up

JonasHelming:
fixed #14692

#### What it does

Introduce a new optional property "autostart" for MCP servers to start them automatically when a frontend is started.

#### How to test

- Set autostart to true for a server
- Restart the backend
- Check that the server is started already and functions are available

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-mcp/src/browser/mcp-frontend-application-contribution.ts
+++ b/packages/ai-mcp/src/browser/mcp-frontend-application-contribution.ts
@@ -13,6 +13,7 @@
 //
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
+
 import { FrontendApplicationContribution, PreferenceProvider, PreferenceService } from '@theia/core/lib/browser';
 import { inject, injectable } from '@theia/core/shared/inversify';
 import { MCPServerDescription, MCPServerManager } from '../common';
@@ -24,6 +25,7 @@ interface MCPServersPreferenceValue {
     command: string;
     args?: string[];
     env?: { [key: string]: string };
+    autostart?: boolean;
 };
 
 interface MCPServersPreference {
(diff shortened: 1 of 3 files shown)


packages/ai-mcp/src/browser/mcp-frontend-application-contribution.ts
```diff
@@ -83,6 +87,18 @@ export class McpFrontendApplicationContribution implements FrontendApplicationCo
...
+        const serverNames = servers.keys();
+        const startedServers = await this.frontendMCPService.getStartedServers();
+        for (const name of serverNames) {
+            const serverDesc = servers.get(name);
+            if (serverDesc && serverDesc.autostart) {
+                if (!startedServers.includes(name)) {
+                    await this.frontendMCPService.startServer(name);
+                }
+            }
+        }
```
sdirix, line 100: Minor: Seems weird to specifically iterate over all keys but to retrieve all their associated values anyway. We could just iterate over all entries in one go.

packages/ai-mcp/src/browser/mcp-frontend-application-contribution.ts
```diff
@@ -83,6 +87,18 @@ export class McpFrontendApplicationContribution implements FrontendApplicationCo
         });
         this.frontendMCPService.registerToolsForAllStartedServers();
     }
+    protected async autoStartServers(servers: Map<string, MCPServerDescription>): Promise<void> {
```
sdirix, line 90: Nitpick: We usually place a newline between functions

packages/ai-mcp/src/browser/mcp-frontend-application-contribution.ts
```diff
@@ -72,8 +75,9 @@ export class McpFrontendApplicationContribution implements FrontendApplicationCo
                 MCP_SERVERS_PREF,
                 {}
             ));
-            this.syncServers(servers);
             this.prevServers = this.convertToMap(servers);
+            this.syncServers(this.prevServers);
+            this.autoStartServers(this.prevServers);
```
sdirix, line 80: Some usability questions:
- This will only auto start servers when the frontend is started. Should we also auto start servers which are added to the preferences with `autostart: true` or when their `autostart` value is added or changed?
- If a frontend is started with a server set to `autostart: false` (or non existing), should we then stop the server in case it is already running?

sdirix (commented):
Works as expected!

I added some usability questions and minor comments.

packages/ai-mcp/src/browser/mcp-frontend-application-contribution.ts
JonasHelming: 1.: I decided to not start them on preference change (this is documented in the settings). Reasoning was that if you start a server manually, you get the message with the functions. If you set up a new server, you probably want to start it manually once. If you have a strong feeling to change this behavior, its fine, but I found it a bit more logical this way.
2.: I decided no, because this would break other frontends that still use this server. Also, "removing auto start", does not imply "stop it" to me. You can still manually stop them, though.

sdirix (approved):
Thanks!

PR #14748, opened 2025-01-20
Fix Console Scrolling

parisa-mchp:
#### What it does

Fixes autoscrolling functionality in debug console. How this should work, is if the console is scrolled to the bottom, then when new content is output, it should be shown automatically, but if the user scrolls up, the view should stay put.

#### How to test

This functionality can be tested with any debuggable project. For example I used the [cpp-debug-workspace](https://github.com/eclipse-theia/theia-cpp-extensions/tree/master/examples/cpp-debug-workspace), which after installing [cdt-gdb-vscode](https://open-vsx.org/extension/eclipse-cdt/cdt-gdb-vscode), can be debugged following the instructions in the README.md file. To test, place a breakpoint within the main loop i.e. `a.cpp:35`. Allow the debugger to continue a few times. Each time it does, any new output to the debug console should be shown automatically. Then, scroll up a few lines and continue a few more times. Now the view should remain fixed.

#### Explanation

Here's how it worked previously:

When `ConsoleContentWidget.model.onChanged()` is fired, this triggers 3 notable events: `ConsoleWidget.revealLastOutputIfNeeded`, `TreeWidget.updateScrollToRow`, and `TreeWidget.updateRows` usually in that order. `revealLastOutputIfNeeded` uses the selection service to focus the newly added node. Next, `updateScrollToRow` attempts to find the row corresponding the the focused node, and fails to find it and thus doesn't scroll. Finally, after that happens `updateRows` creates the corresponding row. 

Furthermore, the logic to determine when the console should be scrolling or not uses the `onScrollUp`, and `onScrollYReachEnd` methods, which don't appear fire for tree widgets (they work for other widgets). I believe this is because it's not the root widget container that is scrolling, but one of it's child elements, but wasn't able to figure this out.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/console/src/browser/console-content-widget.tsx
+++ b/packages/console/src/browser/console-content-widget.tsx
@@ -14,11 +14,10 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { Message } from '@theia/core/shared/@phosphor/messaging';
 import { interfaces, Container, injectable } from '@theia/core/shared/inversify';
 import { MenuPath } from '@theia/core';
 import { TreeProps } from '@theia/core/lib/browser/tree';
-import { SourceTreeWidget, TreeSourceNode, TreeElementNode } from '@theia/core/lib/browser/source-tree';
+import { SourceTreeWidget, TreeElementNode } from '@theia/core/lib/browser/source-tree';
 import { ConsoleItem } from './console-session';
 import { Severity } from '@theia/core/lib/common/severity';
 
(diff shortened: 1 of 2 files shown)


parisa-mchp:
Would it be good to also remove the `onScrollReachEndEmitter` and `onScrollUpEmitter`? Those appear to have been added specifically for this use case.

JonasHelming:
@msujew Could you have a look at this, please?

msujew (approved):
Looks good to me. There is a bit of jumpyness when trying to expand an object in the debug console, but it's definitely better than previously.

msujew:
@parisa-mchp If you want to create a follow-up PR for this, feel free to do so. I'll merge this for now, since this is definitely a major improvement on the existing behavior.

Issue #14749, opened 2025-01-20
Add Support for AI Agent-Driven Change Proposals in AI Chat


planger:
Enable AI agents to propose a set of changes directly within the AI chat interface. The proposed changes should support:  

1. **Generic Editable Elements**:  
   - Changes can target files or custom concepts.  
   - A generic interface should allow for specific editable elements to be implemented, with customizable labels, icons, and actions.  

2. **User Interaction with Changes**:  
   - Display all elements with suggested changes, allowing users to open, review diffs, and decide whether to apply or reject changes.  
   - Support for change types such as *add*, *delete*, and *modify* for each element.  

Proposing changes from an AI agent solely through chat responses can quickly become unwieldy, especially when multiple changes span across several elements. Managing such proposals inline with chat responses can lead to confusion, as it becomes challenging to track the relationship between changes, navigate across modified elements, or grasp the overall context of the suggestions. Furthermore, refining the proposed changes often requires iterative feedback and adjustment, making it essential to have a dedicated, persistent, and structured view of the current state of suggestions. By decoupling these "current suggested changes" from specific chat responses and presenting them in a centralized location, users gain a clearer and more organized overview of all proposed changes, enabling better navigation, review, and decision-making.

PR #14750, opened 2025-01-20
feat: change set support in chat input and chat model

planger:
#### What it does

* Add the concept of a change set to chat model and input UI
* Add an implementation of change set elements for files
* Add an agent for testing: @ChangeSet

Other fixes in Chat Input UI:
* The inProgress state of the chat input was actually unsafely managed. This change addresses the proper management of the inProgress state.
* The positioning, e.g. of the placeholder is now more adaptive.

As the change set feature directly relates to another feature (context, work in progress), this change also already prepares for those changes in the chat UI:
* Prepare chat input for adding context to requests
* Add context in the form of variables to chat model

Fixes https://github.com/eclipse-theia/theia/issues/14749

#### How to test

To demonstrate the usage and simplify testing, this PR includes a test agent named `@ChangeSet`. This statically sets a somewhat random change set into the chat model, with one new, one modified, and one deleted file change set element.

https://github.com/user-attachments/assets/fd4347ec-4ed4-4808-ac41-3e4b4b0735c2

#### Follow-ups

Next steps include:

- [x] Hook up the [Coder Agent](https://github.com/eclipse-theia/theia/blob/master/packages/ai-workspace-agent/src/browser/coder-agent.ts) to allow for testing this change set support in more breadth.
- [ ] Allow to clear a changeset (all changes a single changes)
- [ ] When using with the Coder agent, the progress could be shown in a better way. We should add the change earlier while the LLM is still streaming. We might also open the file and show some line animation there.
* The [Coder Agent](https://github.com/eclipse-theia/theia/blob/master/packages/ai-workspace-agent/src/browser/coder-agent.ts) should be switched to use the monaco model, if it exists, to capture potentially dirty states -- otherwise the AI sees a different version than the user.
* Improve the diff editor we use here: currently there seems to be no simple way to show a diff between the workspace (editable) and a suggested modified version, while allowing to "accept" changes from the suggested version to the workspace version. Therefore, we take a somewhat ugly shortcut in this PR to show a diff between the suggested changes as readonly original version left and the editable workspace version on the right. This works, however, the colors are wrong (red for additions) and the labels are wrong (reject, even though it is actually an accept).
* Potentially optimize the `ChangeSetFileElement` how it applies and reverts changes, taking the monaco text model into account. I aimed at reusing the functionality to modify the monaco model if an editor is open without saving, and saving to disk directly if no editor is opened. But this should be tested for user experience and consistency.

#### Breaking changes

#### Attribution

None

#### Review checklist

#### Reminder for reviewers


examples/api-samples/src/browser/chat/change-set-chat-agent-contribution.ts
```diff
@@ -0,0 +1,132 @@
...
+                    chatSessionId
+                })
+            );
+        }
+        request.session.setChangeSet(changeSet);
+
+        request.response.response.addContent(new MarkdownChatResponseContentImpl(
+            'I have created a change set for you. You can now review and apply it.'
+        ));
+        request.response.complete();
```
JonasHelming, line 125: I wonder when we should update the change set in a real scenario. Simple solution would be when a response is there, but if more then one file is changed, it looks too late from a user POV. Unfortunately, the agent currently does not know which functions are called, therefore, we have no good trigger.

JonasHelming:
Very nice, looks really great. I believe this is a very powerful feature!

Couple of high level comments/questions:

- What is our story to close a changeset or to clear it?

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -0,0 +1,224 @@
...
+            this.monacoWorkspace.applyBackgroundEdit(document, [{
+                range: document.textEditorModel.getFullModelRange(),
+                text
+            }], true);
+        } else {
+            await this.fileService.write(this.uri, text);
+        }
+    }
+
+    async reject(): Promise<void> {
```
JonasHelming, line 209: 'reject' is rather 'undo' or 'discard', two comments:
- I would rename this action and only show it, if the change has already been applied
- I believe we need something to remove a change from a changeset without applying it.

packages/ai-chat/src/common/chat-model.ts
```diff
@@ -420,21 +512,66 @@ export class ChatModelImpl implements ChatModel {
...
+    constructor(public readonly title: string, elements: ChangeSetElement[] = []) {
+        this.addElements(elements);
+    }
+
+    getElements(): ChangeSetElement[] {
+        return this._elements;
+    }
+
+    addElement(element: ChangeSetElement): void {
+        this.addElements([element]);
```
JonasHelming, line 530: Why is no notification triggered in this case?

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -0,0 +1,224 @@
...
+    chatSessionId: string;
+    /**
+     * The state of the file after the changes have been applied.
+     * If `undefined`, there is no change.
+     */
+    targetState?: string;
+};
+
+@injectable()
+export class ChangeSetFileElement implements ChangeSetElement {
```
JonasHelming, line 46: I am not 100% sure if I like the mixture between behavior and state in this class. It relies heavily on DI and Theia services and seems hard to test. Maybe we should extract the behavior to a separate interface?

JonasHelming (commented):
Just a first highlevel look for now

packages/ai-chat/src/browser/change-set-file-resource.ts
JonasHelming, line 1: Is this used somewhere?

packages/ai-chat/src/common/chat-model.ts
eneufeld: this just delegates to the addElements method below, which triggers the notification

packages/ai-chat/src/browser/change-set-file-resource.ts
planger: Yes, the resource resolver is registered here:
https://github.com/eclipse-theia/theia/pull/14750/files/0817dcbf334ccfe0a430798ffcbce1dbcce9baa9#diff-444a8731310f8ef30095775299ab7f86f4c392bdd312ac603b9b795f9b3759d4R103

This is needed to obtain the suggested content of a file via the Theia URIs resolution, enabling to reuse the DiffUris in Theia.

packages/ai-chat/src/browser/change-set-file-element.ts
JonasHelming: My suggestion would be more to extract the behavior into a separate interface (delegation), I am not 100% sure myself, though.

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
```diff
@@ -215,15 +222,40 @@ const ChatInput: React.FunctionComponent<ChatInputProperties> = (props: ChatInpu
...
+        };
+        const listener = lastRequest?.response.onDidChange(onResponseUpdate);
         return () => listener?.dispose();
     }, [lastRequest]);
 
+    // track change set updates
+    React.useEffect(() => {
+        const onChatModelUpdate = (event: ChatChangeEvent) => {
+            if (ChatChangeEvent.isChangeSetEvent(event)) {
+                forceUpdate();
```
sdirix, line 252: It seems the `forceUpdate` is used to update the UI for changes in the changeSet. A bit more elegant would be to maintain a "view model" abstraction in `useState` for the `ChangeSetBox` and update that one whenever needed. Then we would not need a `forceUpdate`

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
```diff
@@ -215,15 +222,40 @@ const ChatInput: React.FunctionComponent<ChatInputProperties> = (props: ChatInpu
...
+    React.useEffect(() => {
+        if (lastRequest) {
+            setInProgress(ChatRequestModel.isInProgress(lastRequest));
+        }
+        const onResponseUpdate = () => {
+            setInProgress(ChatRequestModel.isInProgress(lastRequest));
+        };
+        const listener = lastRequest?.response.onDidChange(onResponseUpdate);
         return () => listener?.dispose();
     }, [lastRequest]);
```
sdirix, line 246: This was not introduced by your PR but as I am just noticing it: We should never build `useEffect` based transformations for `state` which is under our control. At best this introduces unnecessary rerenderings and confusing state updates, at worst this ends up in endless rerender cycles.

As `lastRequest` is maintained by us, we should perform any side effects we want to see at the same time as `lastRrequest` is changed, i.e. by the code calling the setter.

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
```diff
@@ -268,30 +300,143 @@ const ChatInput: React.FunctionComponent<ChatInputProperties> = (props: ChatInpu
...
 
+    const leftOptions = React.useMemo<Option[]>(() => (
+        props.showContext ? [{
+            title: 'Attach elements to context',
+            handler: () => { /* TODO */ },
+            className: 'codicon-add'
+        }] : []
+    ), [props.showContext]);
+
+    const rightOptions = React.useMemo<Option[]>(() => (
```
sdirix, line 311: At the moment the memos are not worth it. The only thing we save are the construction of the arrays, however the `ChatInputOptions` will be rerendered anyway.

Only if we would `memo` the whole `ChatInputOptions`, then the `useMemo`s are needed. Not sure whether all these complications are worth it, I would imagine that the `ChatInputOptions` are cheap to render.

JonasHelming:
It is now integrated with Coder, two example prompts to get all three changes types:

@Coder Move the readme file to a sub direcotry called /readme

@Coder Implement the interface MCPServerManager in the same file (packages/ai-mcp/src/common/mcp-server-manager.ts). Don't ask me anything, jut do it.

examples/api-samples/src/browser/chat/change-set-chat-agent-contribution.ts
planger: As discussed, the agent can at any time update the chat model, and after https://github.com/eclipse-theia/theia/pull/14751 also tool functions can update it.

packages/ai-chat/src/browser/change-set-file-element.ts
planger: I see, so I would propose to add a `remove` action which is always visible and show a `discard` if the change has already been applied. WDYT?

JonasHelming: yes, that makes sense. I would also add a "clear" action next to accept to clear the complete changeset

planger:
@JonasHelming @sdirix 

I've addressed the review feedback and added delete functionality.

@sdirix I've reintroduce the `latestRequest` ref to have proper value in functions. Maybe there is a better solution `latestRequest` was always `undefined` when used in the function `onDeleteChangeSet` and `onDeleteChangeSetElement`.

JonasHelming:
Looks great!

I found one additional issue, either to fix now or as a follow up:
- Discard does not work for me after applying a change (accept a change => discard => nothing happens)

Once we merge this, we should create a follow up epic with all the open topics => I can do this.

JonasHelming (approved):

JonasHelming (approved):

PR #14751, opened 2025-01-21
feat: allow passing context to tool calls

planger:
#### What it does

With this change we enable passing optional context to tool calls.

For tool calls in the context of a chat, we pass the chat request model. This way tool call handlers can retrieve additional information as a context or update the chat model directly. Note that this can also be used to receive the cancellation token via `request.response.cancellationToken` to be able to cancel long-running tool calls if the user hits cancel in the chat.

#### How to test

Create a tool call that receives and processes the optional second argument of its `handler` function.
E.g.

```ts
@injectable()
export class FileContentFunction implements ToolProvider {
    static ID = FILE_CONTENT_FUNCTION_ID;

    getTool(): ToolRequest {
        return {
            id: FileContentFunction.ID,
            name: FileContentFunction.ID,
            description: `The relative path to the target file within the workspace. This path is resolved from the workspace root, and only files within the workspace boundaries
             are accessible. Attempting to access paths outside the workspace will result in an error.`,
            parameters: {
                type: 'object',
                properties: {
                    file: {
                        type: 'string',
                        description: `Return the content of a specified file within the workspace. The file path must be provided relative to the workspace root. Only files within
                         workspace boundaries are accessible; attempting to access files outside the workspace will return an error.`,
                    }
                },
                required: ['file']
            },
            handler: (arg_string: string, ctx: ChatRequestModelImpl) => {
                // if it is a handler that should work in the context of a chat and outside too, it can receive `unknown`
                // and use an instanceof check or ChatRequestModel.is() to switch between cases
                const file = this.parseArg(arg_string);
                return this.getFileContent(file);
            }
        };
    }
...
```

#### Follow-ups

None

#### Breaking changes

#### Attribution

None

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/browser/ai-chat-frontend-module.ts
+++ b/packages/ai-chat/src/browser/ai-chat-frontend-module.ts
@@ -37,10 +37,13 @@ import { aiChatPreferences } from './ai-chat-preferences';
 import { AICustomAgentsFrontendApplicationContribution } from './custom-agent-frontend-application-contribution';
 import { FrontendChatServiceImpl } from './frontend-chat-service';
 import { CustomAgentFactory } from './custom-agent-factory';
+import { ChatToolRequestService } from '../common/chat-tool-request-service';
 
 export default new ContainerModule(bind => {
     bindContributionProvider(bind, Agent);
     bindContributionProvider(bind, ChatAgent);
+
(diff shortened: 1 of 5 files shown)


JonasHelming (approved):
Works well, thank you!

PR #14756, opened 2025-01-22
Support dragging files in browser

msujew:
#### What it does

Adds a feature that wasn't originally included in https://github.com/eclipse-theia/theia/pull/12065. We currently support dragging files from the OS explorer to the main area of Theia, which results in opening the exact file (via its original URI) that was dropped into the application.

This won't work in browser based Theia apps for obvious reasons. Instead, this PRs adjusts the behavior for non-Electron apps. We now simply read the file name and text data of the dropped file into an `untitled` resource that we then attempt to open. Reusing `untitled` resources has a few benefits, mainly that it automatically handles disposal in case the editor is closed and also that it has special behavior when it comes to saving the file (it will automatically open the "Save As..." dialog).

#### How to test

1. Open the browser example app
2. Drag & drop a text (!) file into the main area. Dropping a non-text file (like an image) into the editor will usually result in unexpected behavior. However, it seems like `vscode.dev` also struggles with that, so it's at least as good as VS Code.
3. The file should open, contain the expected file name and otherwise behave as any other `untitled` resource.

#### Follow-ups

We could potentially also support non-text files later on.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/application-shell.ts
+++ b/packages/core/src/browser/shell/application-shell.ts
@@ -22,7 +22,7 @@ import {
 } from '@phosphor/widgets';
 import { Message } from '@phosphor/messaging';
 import { IDragEvent } from '@phosphor/dragdrop';
-import { RecursivePartial, Event as CommonEvent, DisposableCollection, Disposable, environment, isObject } from '../../common';
+import { RecursivePartial, Event as CommonEvent, DisposableCollection, Disposable, environment, isObject, UntitledResourceResolver, UNTITLED_SCHEME } from '../../common';
 import { animationFrame } from '../browser';
 import { Saveable, SaveableWidget, SaveOptions } from '../saveable';
 import { StatusBarImpl, StatusBarEntry, StatusBarAlignment } from '../status-bar/status-bar';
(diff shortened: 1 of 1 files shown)


tsmaeder:
@msujew I have a strange behavior: when I drag a file into the editor area, it opens in a dirty text editor, as expected. But when I click on the close box, I only get two options: "Save" and "Cancel". I should have "Don't save", as well.

msujew:
@tsmaeder Cannot reproduce :/


https://github.com/user-attachments/assets/b1e6ee07-f838-49f6-83a0-9bafdae81c91

msujew:
The `Don't save` button isn't even conditional; it should always appear on the dialog. Perhaps something in your build broke for some reason?

https://github.com/eclipse-theia/theia/blob/c8ae828511e82f654031df5468a2ba2be3f41347/packages/core/src/browser/saveable.ts#L372

tsmaeder:
Here's what I do...looks like the untitled resource causes a "save as..." dialog. This is firefox on windows.


https://github.com/user-attachments/assets/c8177446-9006-4b07-9bc5-934847e32758

tsmaeder (approved):
I had "autosave" on for some reason 🤷

PR #14757, opened 2025-01-22
Add task related context keys

msujew:
#### What it does

Addresses an issue found in https://github.com/eclipse-theia/theia/discussions/14725#discussioncomment-11894424.

Adds a few new task related context keys that are in use by the Python extension. See [here](https://github.com/microsoft/vscode-python/blob/f0962971ffe3c46184cb941907ddef0cce3aff77/package.json#L1468).

#### How to test

1. Open a python file that simply contains a `print("hello world")` script.
2. Assert that the `editor/title/run` menu contains all expected entries:
image
3. Assert that these entries work as expected.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/task/src/browser/task-context-key-service.ts
+++ b/packages/task/src/browser/task-context-key-service.ts
@@ -1,0 +1,72 @@
+// *****************************************************************************
+// Copyright (C) 2024 TypeFox and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { injectable, inject, postConstruct } from '@theia/core/shared/inversify';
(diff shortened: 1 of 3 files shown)


packages/task/src/browser/task-service.ts
```diff
@@ -193,6 +194,9 @@ export class TaskService implements TaskConfigurationClient {
     @inject(TaskTerminalWidgetManager)
     protected readonly taskTerminalWidgetManager: TaskTerminalWidgetManager;
 
+    @inject(TaskContextKeyService)
+    protected readonly taskContextKeyService: TaskContextKeyService;
```
tsmaeder, line 198: I suspect this is to force the creation of an instance?

packages/task/src/browser/task-context-key-service.ts
```diff
@@ -0,0 +1,72 @@
...
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { injectable, inject, postConstruct } from '@theia/core/shared/inversify';
+import { ContextKeyService, ContextKey } from '@theia/core/lib/browser/context-key-service';
+import { ApplicationServer } from '@theia/core/lib/common/application-protocol';
+
+@injectable()
+export class TaskContextKeyService {
```
tsmaeder, line 22: I guess these keys come from VS Code, right? Would it make sense to have a reference to the origin?

tsmaeder (approved):

packages/task/src/browser/task-service.ts
msujew: Yes, we don't really use those currently, and otherwise the service does not get created :/

packages/task/src/browser/task-context-key-service.ts
msujew: I agree, I've added a reference 👍

Issue #14143, opened 2024-09-06
Theia LLM - First iteration


sdirix:
## Overview

This is a ticket to manage the direct follow up of the initial Theia AI support contribution (#14048)

### Architecture

These tasks / evaluations challenge major parts of the contributed Theia AI architecture

- The initial contribution adds an own AI variable service. Theia already has a generic variable service. [It must be evaluated whether they should be fully merged instead of just integrated as they are right now](https://github.com/eclipse-theia/theia/pull/14048#discussion_r1734322460).
- The AI configuration view partially competes with the preferences. It must be evaluated whether they can be (partially) merged and how the new role of the configuration view then looks like.

### Major impact

- The original contribution customizes the `EditorManager` in the `AIEditorManager` and rebinds it. [This is problematic for downstream projects which also rebind the `EditorManager`](https://github.com/eclipse-theia/theia/pull/14048#discussion_r1734227614). The editor manager enhancements should be merged into the base implementation instead. 
- The agent inheritance tree is quite large already. It must be checked whether it can be simplified, [including using composition patterns over inheritance](https://github.com/eclipse-theia/theia/pull/14048/files#r1736144466). 
- [It must be discussed whether we want to adapt the ai code completion items or remove them](https://github.com/eclipse-theia/theia/pull/14048/files#r1742286819). We will definitively keep the inline suggestions.  
- User edited prompt templates are stored in separate configurable directories. [For this a already layered mechanism can be used](https://github.com/eclipse-theia/theia/pull/14048#discussion_r1741016957).

### Best practices

- Rename and refactor the language model delegation mechanic. Some of the used classes/services can be merged and they should [follow the Theia conventions for naming](https://github.com/eclipse-theia/theia/pull/14048/files#r1740906884).
- Introduce i18n for all hard coded strings
- A user message was shown by customizing the `MonacoEditorProvider` in the same way as the `EditorManager`. This was removed from the contribution. However it would be great to get the customization back in. This must be evaluated.
- The contribution introduces a new offset interface. [It should be checked whether it can be replaced with an already existing concept in Theia](https://github.com/eclipse-theia/theia/pull/14048/files#r1736491182).

### Minor scale improvements

- Prompts implementation [can be improved](https://github.com/eclipse-theia/theia/pull/14048#discussion_r1740966401)
- Some of the new variables are contributed one at a time in separate contributions. [They should be merged](https://github.com/eclipse-theia/theia/pull/14048/files#r1741024476).
[...]

Issue #14701, opened 2025-01-07
[Theia AI] MCP does not work after page reload in the browser


safisa:
### Bug Description:
Hi,
I have installed some MCP servers which worked as expected in my custom agent, but after reloading the page, the agent doesn't recognize the tools of the MCP server.

### Steps to Reproduce:

1. define an MCP server, e.g.: 
``` 
"filesystem": {
            "command": "npx",
            "args": [
                "-y",
                "@modelcontextprotocol/server-filesystem",
                "a-local-folder"
            ]
        }
```
2. start this MCP server
3. add a custom agent, and define one of the MCP tools in the custom agent system prompt.
4. test the agent and let it use the tool, it should work
5. reload the page
6. try step 4, it will not work

### Additional Information

- Operating System: Mac
- Theia Version: Theia 1.57.1 (latest)
-

sdirix:
This error makes sense as the tools are registered directly within the command, see [here](https://github.com/eclipse-theia/theia/blob/9edbf2c9ef8d30379dbe092b3850e814672346d3/packages/ai-mcp/src/browser/mcp-command-contribution.ts#L104-L106).

As the mcp servers run on the backend, we should add a mechanism to recover/register already managed servers. This would then resolve this ticket.

PR #14760, opened 2025-01-23
fix: manage AI bindings separately per connection

sdirix:
#### What it does

Moves all AI backend bindings into ConnectionContainerModules. Therefore these are now scoped per connection (i.e. frontend) instead of globally for all connections.

The global bindings lead to a lot of interference when multiple connections were open. In fact AI streaming did only work for the latest connection.

Related to https://github.com/eclipse-theia/theia/issues/14143

#### How to test

- Open multiple tabs/windows of Theia and verify that Theia AI works in all of them.
- Verify that AI features are connection scoped, e.g. use workspace preferences for AI models, open two different workspaces at the same time and observe that the expected models show up in the corresponding windows

#### Follow-ups

Related to https://github.com/eclipse-theia/theia/issues/14143 this prepares the code base to simplify / streamline a lot of the delegation mechanics.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-anthropic/src/node/anthropic-backend-module.ts
+++ b/packages/ai-anthropic/src/node/anthropic-backend-module.ts
@@ -18,11 +18,17 @@ import { ContainerModule } from '@theia/core/shared/inversify';
 import { ANTHROPIC_LANGUAGE_MODELS_MANAGER_PATH, AnthropicLanguageModelsManager } from '../common/anthropic-language-models-manager';
 import { ConnectionHandler, RpcConnectionHandler } from '@theia/core';
 import { AnthropicLanguageModelsManagerImpl } from './anthropic-language-models-manager-impl';
+import { ConnectionContainerModule } from '@theia/core/lib/node/messaging/connection-container-module';
 
-export default new ContainerModule(bind => {
+// We use a connection module to handle AI services separately for each frontend.
+const anthropicConnectionModule = ConnectionContainerModule.create(({ bind, bindBackendService, bindFrontendService }) => {
     bind(AnthropicLanguageModelsManagerImpl).toSelf().inSingletonScope();
     bind(AnthropicLanguageModelsManager).toService(AnthropicLanguageModelsManagerImpl);
     bind(ConnectionHandler).toDynamicValue(ctx =>
         new RpcConnectionHandler(ANTHROPIC_LANGUAGE_MODELS_MANAGER_PATH, () => ctx.container.get(AnthropicLanguageModelsManager))
     ).inSingletonScope();
+});
+
+export default new ContainerModule(bind => {
+    bind(ConnectionContainerModule).toConstantValue(anthropicConnectionModule);
 });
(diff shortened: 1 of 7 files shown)


sdirix:
@JonasHelming Note that each frontend then has their own (auto-started) mcp servers, they are no longer globally managed.

JonasHelming:
@sdirix Thank you. I consider this to be a good thing, right? Or do you see any issues upcoming.

JonasHelming (approved):
Thank you that resolves a lot of issues!

JonasHelming:
Good point. I am actually torn. On the one hand, we have the auto-start now. On the other hand, we received this report before:
https://github.com/eclipse-theia/theia/issues/14701
Is there any reason, why you wanted to scope the MCP servers per connection in the first place?

sdirix:
I think it's really a use case point of view:
 - If you use one window per workspace and have different (auto-start) workspace settings for mcp servers for each of them, you probably also want to have them managed per window instead
 - If there is a stateful or expensive mcp server, you might want to have it shared instead

JonasHelming:
OK, I thought about it. As there are servers with a state, e.g. via roots, I believe we should scope them. A use case is for example that we transfer the current workspace as a root to the server, which would conflict if several clients use the same instance.
Especially now that we have auto start.

sdirix:
Hi @kittaakos, you can rebind them as before, it just has to happen in the connection container.

So your code changes from this:

```ts
import { OpenAiLanguageModelsManagerImpl } from '@theia/ai-openai/lib/node/openai-language-models-manager-impl';
import { ContainerModule } from '@theia/core/shared/inversify';
import { MyManagerImpl } from './my-openai-language-models-manager';

export default new ContainerModule((bind, _unbind, _isBound, rebind) => {
    rebind(OpenAiLanguageModelsManagerImpl).to(MyManagerImpl).inSingletonScope();
});
```

to

```ts
import { OpenAiLanguageModelsManagerImpl } from '@theia/ai-openai/lib/node/openai-language-models-manager-impl';
import { ContainerModule } from '@theia/core/shared/inversify';
import { MyManagerImpl } from './my-openai-language-models-manager;
import { ConnectionContainerModule } from '@theia/core/lib/node/messaging/connection-container-module';

const connectionModule = ConnectionContainerModule.create(({ rebind }) => {
    rebind(OpenAiLanguageModelsManagerImpl).to(MyManagerImpl).inSingletonScope();
});

export default new ContainerModule((bind) => {
    bind(ConnectionContainerModule).toConstantValue(openAiConnectionModule);
});
```

sdirix:
>The factory symbol is unused:
> https://github.com/eclipse-theia/theia/blob/f29a72129c64ea865e6468195dcc06d8139094f5/packages/ai-openai/src/node/openai-backend-module.ts#L24

This is an oversight and should be removed. Thanks for pointing it out.

kittaakos:
> So your code changes from this:

Thank you so much for the prompt support 🙏

PR #14764, opened 2025-01-23
Fix running playwright test with npm and electron.

### LINKED ISSUE #14763: Playwright test don't work for electron with npm build
### Bug Description:
In https://github.com/eclipse-theia/theia-e2e-test-suite/pull/44, our performance tests don't work for Electron. This is due to the electron package being hosted to a different location than with yarn. 

### Steps to Reproduce:
1. go to examples/playwright
2. run 'npm run ui-tests-electron'
3. observe: none of the tests work.

### Additional Information

- Operating System: Windows
- Theia Version: 1.57 master

tsmaeder:
#### What it does
This PR launches the electron playwright tests with the installed electron version. It also introduces an environment variable `THEIA_NO_SPLASH` to prevent showing a splash screen (which breaks the tests).

Fixes #14763

#### How to test
Run the playwright tests with electron. Note that we don't expect all of the tests to pass for electron.

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/playwright/src/theia-app-loader.ts
+++ b/examples/playwright/src/theia-app-loader.ts
@@ -101,7 +101,7 @@ namespace TheiaElectronAppLoader {
         const appPath = electronConfig.electronAppPath!;
         const pluginsPath = electronConfig.pluginsPath;
         const launchOptions = electronConfig.launchOptions ?? {
-            additionalArgs: ['--no-cluster'],
+            additionalArgs: ['--no-sandbox', '--no-cluster'],
             electronAppPath: appPath,
             pluginsPath: pluginsPath
         };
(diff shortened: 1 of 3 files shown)


tsmaeder:
@planger does the splash screen show up? Do you have the plugins downloaded?

planger:
Yes, I've downloaded the plugins.

I confirmed that the splash screen doesn't show up in I run `npm run start:electron` and the app starts.  But when executing the UI tests with electron, no application seems to come up whatsoever.

tsmaeder:
When copying the parameters from playwright electron launcher, the launch fails because of the well known sandbox problem. Adding `--no-sandbox` fixes the problem when launching electron on the command line. Still doesn't help with the tests, though. But I suspect launching on a recent linux distro has never worked at all without `-no-sandbox`.

tsmaeder:
It seems the problem I still have is with nvm: the ouput from launching the electron script says `/usr/bin/env 'node': no such file or directory`. The same happens when I do `sudo /usr/bin/env node`. nvm only puts nodejs on the path of the current terminal session. It seems the way playwright launches the session does not include the current node installation in the path and therefore `node_modules/.bin/electron` fails to launch.

planger:
I can confirm that Electron UI tests now work on Ubuntu 22.04. Thank you! :partying_face:

planger (approved):
Thank you, works fine on Linux!

PR #14769, opened 2025-01-24
fix: improve performance in AI request logging

sdirix:
#### What it does

Instead of appending AI requests line by line to the output channel, we now append them in a single step. Since appending to a channel is a relatively expensive operation, this change significantly improves performance for large requests.

This is particularly noticeable when AI inline code completions are enabled for large files, as many requests with many lines may be sent during typing.

##### Performance Analysis when typing in average speed for 5 seconds in a large file (1000 lines) *without* this PR

image
image

- We do not spend any time being idle
- 92% of time is spent in `doAppend` of the output channel

##### Performance Analysis when typing in average speed for 5 seconds in a large file (1000 lines) *with* this PR

image
image
image

- A lot is going on but we also were idle 40% of the time
- Only 3% of the time is spent in `doAppend`
- Most time is spent in React and animations

##### Performance Analysis when typing in average speed for 5 seconds in a large file (1000 lines) without inline completions

image
image

- Barely any difference compared to enabled completions with this PR

#### How to test

Enable automatic inline completions and feel the performance difference. Especially noticeable in large files

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-code-completion/src/browser/code-completion-agent.ts
+++ b/packages/ai-code-completion/src/browser/code-completion-agent.ts
@@ -114,6 +114,9 @@ export class CodeCompletionAgentImpl implements CodeCompletionAgent {
             const requestId = generateUuid();
             const request: LanguageModelRequest = {
                 messages: [{ type: 'text', actor: 'user', query: prompt }],
+                settings: {
+                    stream: false
+                }
             };
             if (token.isCancellationRequested) {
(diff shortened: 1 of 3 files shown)


eneufeld (approved):
Looks great!

JonasHelming (approved):

PR #14777, opened 2025-01-27
Allow filerting backticks in AI code completion

### LINKED ISSUE #14461: [Theia AI] Setting to automatically remove backticks from LLM response
Some LLMs, e.g. smaller llama are reluctant to return code without backticks. Add a setting to automatically remove the backticks from the LLM response

JonasHelming:
fixed #14461

#### What it does

- Add a new prompt variant "-next" to code completion agents
- Add a settings to filter backticks from code completion

#### How to test

Use the next variant for code completion (report if it performs better)
Check that backticks are removed (compare with history view where backicks are not removed)

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-code-completion/src/browser/code-completion-postprocessor.ts
+++ b/packages/ai-code-completion/src/browser/code-completion-postprocessor.ts
@@ -1,0 +1,48 @@
+// *****************************************************************************
+// Copyright (C) 2024 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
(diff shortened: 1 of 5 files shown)


packages/ai-code-completion/src/browser/code-completion-postprocessor.ts
```diff
@@ -0,0 +1,48 @@
...
+            return this.stripBackticks(text);
+        }
+        return text;
+    }
+
+    public stripBackticks(text: string): string {
+        if (text.startsWith('```')) {
+            // Remove the first backticks and any language identifier
+            const startRemoved = text.slice(3).replace(/^\w*\n/, '');
+            const secondBacktickIndex = startRemoved.indexOf('```');
```
eneufeld, line 43: shouldn't we use lastIndexOf ?
This would better catch this case:
` ```\nFoo```Bar```FooBar``` `
I would expect the result to be: ` Foo```Bar```FooBar `
but in your code I assume it would be: `Foo`

eneufeld (commented):
Looks good and works well. 
I left a remark about the algorithm though

eneufeld (approved):
Thank you!

sdirix (approved):

PR #14786, opened 2025-01-28
Improve behavior of AI Changeset diff editor

### LINKED ISSUE #14785: Improve Changeset Diff Editor Experience
### Feature Description:
The experience when opening the a diff from a AI-proposed change set is confusing. It would be nice if we could at least get the proposed change on the right side.

tsmaeder:
#### What it does
Moves the diff to the right side and applies the changes to the underlying file upon "save" in the diff editor. If an editor is open at the time, it will be updated and will retain the same dirty state as before.

partially fixes #14785

#### How to test
Create an AI change set (for example by invoking @ChangeSet in the AI chat) and play around with the diff editor, "accept", etc.

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dependency-check-baseline.json
+++ b/dependency-check-baseline.json
@@ -1,3 +1,2 @@
 {
-  "npm/npmjs/-/unicode-property-aliases-ecmascript/2.1.0": "Manually approved"
 }
--- a/packages/ai-chat/src/browser/change-set-file-resource.ts
+++ b/packages/ai-chat/src/browser/change-set-file-resource.ts
@@ -64,7 +64,7 @@ export class ChangeSetFileResourceResolver implements ResourceResolver {
             initiallyDirty: true,
             readContents: async () => element.targetState ?? '',
             saveContents: async (content: string, options?: ResourceSaveOptions): Promise<void> => {
-                element.accept();
+                element.accept(content);
             },
             dispose: () => { }
         };
--- a/packages/ai-chat/src/browser/change-set-file-resource.ts
+++ b/packages/ai-chat/src/browser/change-set-file-resource.ts
@@ -14,7 +14,7 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { Resource, ResourceResolver, URI } from '@theia/core';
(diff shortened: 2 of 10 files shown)


packages/ai-chat/src/browser/change-set-file-resource.ts
```diff
@@ -60,8 +60,12 @@ export class ChangeSetFileResourceResolver implements ResourceResolver {
 
         return {
             uri,
-            readOnly: true,
+            readOnly: false,
+            initiallyDirty: true,
             readContents: async () => element.targetState ?? '',
+            saveContents: async (content: string, options?: ResourceSaveOptions): Promise<void> => {
+                element.accept();
```
planger, line 67: I think we should still save the `content` and not just plain accept, because the user may have edited something in the diff editor and then expects that the save stores the file with their edit and not just the original suggestion of the AI agent.
Setting the element to state "accept" is a nice touch though as it is a great reference for the user what they already reviewed and saved.

planger (changes requested):
Excellent work, thank you! Feels pretty good now.
I just found two issues while testing and reviewing the changes:

* When the user hits save (Ctrl-S) on a new file, Theia asks where to store this file, even though it should actually already be clear where to put this file.
* When the user modifies the content in the diff editor, the changes are lost. I think this may be easy to fix, if we just store the `content` -- see inline comment.

WDYT?

dependency-check-baseline.json
```diff
@@ -1,3 +1,2 @@
 {
-  "npm/npmjs/-/unicode-property-aliases-ecmascript/2.1.0": "Manually approved"
```
planger, line 2: This change seems unrelated

planger (approved):
Thanks for the revision! This is certainly a great improvement to before, so I'm ok with merging it and addressing the remaining issues in follow-ups.

Issues I found:
* When opening new files (click on the added file item), the label is some random id instead of the file name. Save works fine though.
* When opening the diff editor on new files and saving, it still asks me where to save the file. I'd have expected that it would just create and save the current content to the defined URI location.

dependency-check-baseline.json
tsmaeder: Yes, it's not needed anymore

PR #14789, opened 2025-01-29
Notebook: fixed small race condition for outputs

jonah-iden:
#### What it does

This fixes a small race condition that could happen when the rendering of the outputs widget was triggered before the `WebviewWidget`. This led to the outputs not being rendered at all

#### How to test

The only way i was able to reproduce this once was when i had an notebook with outputs open and then did a window reload.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


packages/plugin-ext/src/main/browser/notebooks/renderers/cell-output-webview.tsx
```diff
@@ -242,6 +243,7 @@ export class CellOutputWebviewImpl implements CellOutputWebview, Disposable {
     protected element?: HTMLDivElement; // React.createRef<HTMLDivElement>();
 
     protected webviewWidget: WebviewWidget;
+    protected webiviewWidgetInitialized = new Deferred();
```
msujew, line 246: Typo:
```suggestion
    protected webviewWidgetInitialized = new Deferred();
```

msujew (approved):

PR #14792, opened 2025-01-29
Override change elements on additional changes

### LINKED ISSUE #14791: [Theia AI] Override existing change elements per file
To avoid confusion and inconsistent states, we should override a change element if changes for the same file are proposed later.

JonasHelming:
fixed #14791

#### What it does

Coder functions use new addOrReplace function in changeset to override existing changes for a file.

#### How to test

generate two changes for a file and check that only the second change is applied e.g.:

1. '@Coder Remove the prefix "MCP" from all user message in packages/ai-mcp/src/browser/mcp-command-contribution.ts'
2. '@Coder Sorry I meant replace "MCP" with "PCM"'

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-model.ts
+++ b/packages/ai-chat/src/common/chat-model.ts
@@ -571,13 +571,10 @@ export class ChangeSetImpl implements ChangeSet {
     }
 
     addOrReplaceElement(element: ChangeSetElement): void {
-        const index = this._elements.findIndex(e => e.uri.toString() === element.uri.toString());
-        if (index >= 0) {
-            this._elements[index] = element;
-        } else {
-            this._elements.push(element);
+        if (!this.replaceElement(element)) {
+            this.addElement(element);
+            this.notifyChange();
         }
-        this.notifyChange();
     }
 
     removeElement(index: number): void {
--- a/packages/ai-chat/src/common/chat-model.ts
+++ b/packages/ai-chat/src/common/chat-model.ts
@@ -570,6 +570,16 @@ export class ChangeSetImpl implements ChangeSet {
         return true;
     }
 
+    addOrReplaceElement(element: ChangeSetElement): void {
+        const index = this._elements.findIndex(e => e.uri.toString() === element.uri.toString());
+        if (index >= 0) {
+            this._elements[index] = element;
+        } else {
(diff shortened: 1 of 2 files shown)


packages/ai-chat/src/common/chat-model.ts
```diff
@@ -570,6 +570,16 @@ export class ChangeSetImpl implements ChangeSet {
...
 
+    addOrReplaceElement(element: ChangeSetElement): void {
+        const index = this._elements.findIndex(e => e.uri.toString() === element.uri.toString());
+        if (index >= 0) {
+            this._elements[index] = element;
+        } else {
+            this._elements.push(element);
+        }
+        this.notifyChange();
+    }
```
planger, line 581: We may want to call `replaceElement()` and only if it returns `false`, call `addElement()` to avoid the code duplication. The only disadvantage would be the duplicate event.

Not tested but would look something like below:

```suggestion
    addOrReplaceElement(element: ChangeSetElement): void {
        if (!this.replaceElement(element)) {
          this.addElement(element);
        }
    }
```

planger (approved):
Looks good, but I think we can simplify the method, if we are fine with duplicate events. Otherwise, it looks good to me.

packages/ai-chat/src/common/chat-model.ts
JonasHelming: @planger Good suggestion, I think we can avoid the duplicate event, see latest commit

packages/ai-chat/src/common/chat-model.ts
```diff
@@ -570,6 +570,13 @@ export class ChangeSetImpl implements ChangeSet {
         return true;
     }
 
+    addOrReplaceElement(element: ChangeSetElement): void {
+        if (!this.replaceElement(element)) {
+            this.addElement(element);
+            this.notifyChange();
```
planger, line 576: I think we shouldn't notify again because addElement already notifies.

JonasHelming: True, I was totally confused about your previous comment that your change introduces a duplicate notification :-)

planger (approved):
Thanks, looks good!

PR #14794, opened 2025-01-30
Fix translation update script

jfaltermeier:
#### What it does

Fixes this error: https://github.com/eclipse-theia/theia/actions/runs/13048443071/job/36403212934#step:7:20

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/localization-manager/src/deepl-api.ts
+++ b/dev-packages/localization-manager/src/deepl-api.ts
@@ -15,10 +15,16 @@
 // *****************************************************************************
 
 import * as bent from 'bent';
+import { RateLimiter } from 'limiter';
 
 const post = bent('POST', 'json', 200);
 // 50 is the maximum amount of translations per request
 const deeplLimit = 50;
+const rateLimiter = new RateLimiter({
+    tokensPerInterval: 10,
+    interval: 'second',
+    fireImmediately: true
+});
 
 export async function deepl(
     parameters: DeeplParameters
@@ -30,12 +36,11 @@ export async function deepl(
     while (textArray.length > 0) {
         textChunks.push(textArray.splice(0, deeplLimit));
     }
-    const responses: DeeplResponse[] = await Promise.all(textChunks.map(chunk => {
+    const responses: DeeplResponse[] = await Promise.all(textChunks.map(async chunk => {
         const parameterCopy: DeeplParameters = { ...parameters, text: chunk };
-        return post(`https://${sub_domain}.deepl.com/v2/translate`, Buffer.from(toFormData(parameterCopy)), {
-            'Content-Type': 'application/x-www-form-urlencoded',
-            'User-Agent': 'Theia-Localization-Manager'
-        });
+        const url = `https://${sub_domain}.deepl.com/v2/translate`;
+        const buffer = Buffer.from(toFormData(parameterCopy));
+        return postWithRetry(url, buffer);
     }));
     const mergedResponse: DeeplResponse = { translations: [] };
     for (const response of responses) {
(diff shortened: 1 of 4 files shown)


jfaltermeier:
Triggered a test run here: https://github.com/eclipse-theia/theia/actions/runs/13048824155/job/36404384953#step:7:146
Now fails with "Too many requests", which should be unrelated to the script itself. 
I'll retrigger

tsmaeder:
Alternatively, we could add a "translate" script to the root package.json.

jfaltermeier:
Getting "Too many requests" again. I don’t think there’s anything I can do about it. Deepl status looks fine: https://www.deeplstatus.com/
Maybe we’ve hit our limit? I’m not sure what plan or API token we’re using for the translation service.
@msujew Do you know?

msujew:
@jfaltermeier My API token almost hasn't been used this month. It seems like we're hitting some rate limiting?

msujew:
@jfaltermeier I pushed a commit to this PR that should resolve the issue. It seems like DeepL has implemented rather strict rate limiting (without saying what it is - the documentation just says to "try again after a delay").

jfaltermeier:
Thanks, with these commits I get: https://github.com/eclipse-theia/theia/actions/runs/13050402257/job/36409193760#step:7:151

```
Could not translate into language "RU" Error: Requested tokens 11 exceeds maximum tokens per interval 10
    at RateLimiter.removeTokens (/home/runner/work/theia/theia/node_modules/limiter/dist/cjs/RateLimiter.js:42:19)
    at postWithRetry (/home/runner/work/theia/theia/dev-packages/localization-manager/lib/deepl-api.js:55:27)
    at postWithRetry (/home/runner/work/theia/theia/dev-packages/localization-manager/lib/deepl-api.js:64:20)
    at async Promise.all (index 0)
    at async LocalizationManager.deepl [as localizationFn] (/home/runner/work/theia/theia/dev-packages/localization-manager/lib/deepl-api.js:37:23)
    at async LocalizationManager.translateLanguage (/home/runner/work/theia/theia/dev-packages/localization-manager/lib/localization-manager.js:88:45)
    at async Promise.all (index 8)
    at async LocalizationManager.localize (/home/runner/work/theia/theia/dev-packages/localization-manager/lib/localization-manager.js:63:25)
    at async Object.handler (/home/runner/work/theia/theia/dev-packages/cli/lib/theia.js:399:29)
 ```

File: dev-packages/localization-manager/src/deepl-api.ts [hunk 1]
```diff
@@ -47,6 +52,22 @@ export async function deepl(
     return mergedResponse;
 }
 
+async function postWithRetry(url: string, buffer: Buffer, attempt: number): Promise<DeeplResponse> {
+    try {
+        await rateLimiter.removeTokens(attempt);
```
jfaltermeier, line 57: Could this error (https://github.com/eclipse-theia/theia/actions/runs/13050402257/job/36409193760#step:7:151) be related to this line? Shouldn't we be removing only one token per attempt?

msujew: Yes, but we should use a backoff algorithm, so we don't run into this every 100ms.

jfaltermeier: Ah, I see. So the error is just because we tried to remove more tokens than there were, even if full, and not because it got negative. Now it works: https://github.com/eclipse-theia/theia/pull/14796

tsmaeder (approved):
The code looks reasonable to me. Does it work with the deepl  API, though?

msujew:
@tsmaeder This fix has resulted in the PR https://github.com/eclipse-theia/theia/pull/14796. This seems to work as expected :)

PR #14827, opened 2025-02-04
improve integration between VariableRegistry and AI VariableService

### LINKED ISSUE #14817: Streamline AI Variable Service
The initial contribution adds an own AI variable service. Theia already has a generic variable service. It must be evaluated whether they should be fully merged instead of just integrated as they are right now.

CamilleLetavernier:
- make TheiaVariableContribution more easily to customize/override

fixes #14817

#### What it does

- Bind TheiaVariableContribution to self, to make it easier to override

Note: the ticket was about integrating the AI Variable Service with the Theia VariableRegistry (ideally reusing the VariableRegistry directly for AI). However, there are some key differences between the two:

- AI Variable Service is part of Common, whereas VariableRegistry only exists in the Frontend
- AI Variable Service is slightly more customizable, as Variable Resolvers are separate from the Variable Definition, and can be overridden more easily
- VariableRegistry supports more types (including structured arrays/objects); AI Variable Service only supports strings

Moreover, the AI Variable Service already includes a Contribution that delegates to the VariableRegistry in the browser, so no additional work is required here. I made the TheiaVariableContribution a bit easier to replaced (by binding it to self), but actual use cases for customization are unclear at this point, so I don't know how relevant this is.

#### How to test

- Start Theia with AI Chat view, and check that Theia Variables still work as expected (e.g. #lineNumber should appear in auto-completion)

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/browser/theia-variable-contribution.ts
+++ b/packages/ai-core/src/browser/theia-variable-contribution.ts
@@ -18,6 +18,9 @@ import { inject, injectable } from '@theia/core/shared/inversify';
 import { VariableRegistry, VariableResolverService } from '@theia/variable-resolver/lib/browser';
 import { AIVariableContribution, AIVariableResolver, AIVariableService, AIVariableResolutionRequest, AIVariableContext, ResolvedAIVariable } from '../common';
 
+/**
+ * Integrates the Theia VariableRegistry with the Theia AI VariableService when running in the browser.
+ */
 @injectable()
 export class TheiaVariableContribution implements AIVariableContribution, AIVariableResolver {
     @inject(VariableResolverService)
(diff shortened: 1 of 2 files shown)


packages/ai-core/src/browser/theia-variable-contribution.ts
```diff
@@ -18,6 +18,9 @@ import { inject, injectable } from '@theia/core/shared/inversify';
 import { VariableRegistry, VariableResolverService } from '@theia/variable-resolver/lib/browser';
 import { AIVariableContribution, AIVariableResolver, AIVariableService, AIVariableResolutionRequest, AIVariableContext, ResolvedAIVariable } from '../common';
 
+/**
+ * Integrates the Theia VariableRegistry with the Theia AI VariableService when running in the browser.
```
sdirix, line 22:
```suggestion
 * Integrates the Theia VariableRegistry with the Theia AI VariableService
```

sdirix (approved):

Issue #14826, opened 2025-02-04
monaco hover popup hidden by the toolbar


kittaakos:
### Bug Description:

When the toolbar is activated using the shortcut <kbd>Alt/⌥</kbd> + <kbd>T</kbd>, the layout may become incorrect, causing the monaco editor's hover popup to be hidden. This can make it challenging to read the information displayed in the hover.

### Steps to Reproduce:

1. Open `theia` in Theia and navigate to the `packages/property-view/src/browser/resource-property-view/index.ts` file (this is just an example; you can use any other file that has code elements triggering a hover in Monaco).
2. Activate the toolbar by pressing <kbd>Alt/⌥</kbd>+<kbd>T</kbd>.
3. Hover over a code element with validation errors in the monaco editor to trigger the hover popup.
4. Notice that the hover popup is cut off due to the layout of the toolbar.
5. Disable the toolbar with <kbd>Alt/⌥</kbd>+<kbd>T</kbd>, inspect the popup is not hidden.

https://github.com/user-attachments/assets/0325d39c-a8f7-436d-9546-458a9e29c88c

Currently, it can be difficult to read when the toolbar is visible, depending on the space between the top panel and the popup. This issue arises because the popup's position is calculated relative to the `theia-top-panel` instead of the `main-toolbar`, leading to improper placement.

### Expected behavior:

See how the popup's rect is correctly adjusted to the `theia-top-panel` when there is no toolbar.

https://github.com/user-attachments/assets/cc12f693-0d3e-48f6-9713-4417a3b3100e

### Additional Information

- Operating System:
- Theia Version: 1.58.0

PR #14836, opened 2025-02-04
fix: `ContentHoverWidget` respects Theia styles

kittaakos:
#### What it does

Customize the default `ContentHoverWidget` behavior to ensure it uses Theia styles when calculating available space for the hover widget.

VS Code uses 30 pixels for the top height and 24 pixels for the bottom height, but Theia uses 32 pixels for the top and 22 for the bottom.

Closes eclipse-theia/theia#14826

This PR fixes the available space calculation for the content hover widget in the monaco editor. As described in #14826, the Theia top and bottom height (menu + status bar) are different in VS Code. In some cases, the popup either was on the menu or was hidden under the Theia toolbar. This PR fixes both cases.

#### How to test

Please, as shown in the issue, try to make the editor hover either overlay the menu or hide below the toolbar when active. Let me know if you still find glitches.

In-action:

https://github.com/user-attachments/assets/e5cd92a0-b3ad-4893-a68a-dac444d52ed2

https://github.com/user-attachments/assets/e9b117ef-8ff1-4548-a4cd-18fd67c42dc7

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-init.ts
+++ b/packages/monaco/src/browser/monaco-init.ts
@@ -50,56 +50,9 @@ import { MonacoQuickInputImplementation } from './monaco-quick-input-service';
 import { IQuickInputService } from '@theia/monaco-editor-core/esm/vs/platform/quickinput/common/quickInput';
 import { IStandaloneThemeService } from '@theia/monaco-editor-core/esm/vs/editor/standalone/common/standaloneTheme';
 import { MonacoStandaloneThemeService } from './monaco-standalone-theme-service';
-import { ContentHoverWidget } from '@theia/monaco-editor-core/esm/vs/editor/contrib/hover/browser/contentHoverWidget';
-import { IPosition } from '@theia/monaco-editor-core/esm/vs/editor/common/core/position';
+import { createContentWidgetPatcher } from './content-hover-widget-patcher';
 
-// https://github.com/microsoft/vscode/blob/1430e1845cbf5ec29a2fc265f12c7fb5c3d685c3/src/vs/editor/contrib/hover/browser/resizableContentWidget.ts#L13-L14
-const VSCODE_TOP_HEIGHT = 30;
-const VSCODE_BOTTOM_HEIGHT = 24;
-
-// VS Code uses 30 pixel for top height, and 24 pixels for bottom height, but Theia uses 32 pixel for the top and 22 for the bottom.
-// https://github.com/eclipse-theia/theia/blob/b752ea690bdc4e7c5d9ab98a138504ead05be0d1/packages/core/src/browser/style/menus.css#L22
-// https://github.com/eclipse-theia/theia/blob/b752ea690bdc4e7c5d9ab98a138504ead05be0d1/packages/core/src/browser/style/status-bar.css#L18
-// https://github.com/eclipse-theia/theia/issues/14826
-function patchContentHoverWidget(topPanelHeight = 32): { setActualTopHeightForContentHoverWidget: (value: number) => void } {
-    let _actualTopHeight = topPanelHeight;
-    function getTopHeightDiff(): number {
-        return _actualTopHeight - VSCODE_TOP_HEIGHT;
-    }
-
-    const actualBottomHeight = 22; // Theia's status bar height
-    const bottomHeightDiff = actualBottomHeight - VSCODE_BOTTOM_HEIGHT;
-
-    const originalAvailableVerticalSpaceAbove = ContentHoverWidget.prototype['_availableVerticalSpaceAbove'];
-    ContentHoverWidget.prototype['_availableVerticalSpaceAbove'] = function (position: IPosition): number | undefined {
-        const value = originalAvailableVerticalSpaceAbove.call(this, position);
-        // The original implementation deducts the height of the top panel from the total available space.
-        // https://github.com/microsoft/vscode/blob/1430e1845cbf5ec29a2fc265f12c7fb5c3d685c3/src/vs/editor/contrib/hover/browser/resizableContentWidget.ts#L71
-        // However, in Theia, the top panel has generally different size (especially when the toolbar is visible).
-        // This additional height must be further subtracted from the computed height for accurate positioning.
-        const topHeightDiff = getTopHeightDiff();
-        return typeof value === 'number' ? value - topHeightDiff : undefined;
-    };
-
-    const originalAvailableVerticalSpaceBelow = ContentHoverWidget.prototype['_availableVerticalSpaceBelow'];
-    ContentHoverWidget.prototype['_availableVerticalSpaceBelow'] = function (position: IPosition): number | undefined {
-        const value = originalAvailableVerticalSpaceBelow.call(this, position);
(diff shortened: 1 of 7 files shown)


packages/toolbar/src/browser/application-shell-with-toolbar-override.ts
```diff
@@ -60,12 +61,14 @@ export class ApplicationShellWithToolbarOverride extends ApplicationShell {
...
+        const show = doShowToolbarFromPreference && !isShellMaximized;
+        if (show) {
             this.toolbar.show();
-            return true;
+        } else {
+            this.toolbar.hide();
         }
-        this.toolbar.hide();
-        return false;
+        setActualTopHeightForContentHoverWidget(show ? 32 * 2 : 32);
```
martin-fleck-at, line 70: Could you elaborate where the 32 comes from? In the browser I see that the top panel has 32px but the toolbar only has a height of 30px. Overall I'm a bit unsure about those hard-coded numbers. Wouldn't it be better to use something like `this.topPanel.node.getBoundingClientRect().height` and `this.toolbar.node.getBoundingClientRect().height` instead?

packages/monaco/src/browser/monaco-init.ts
```diff
@@ -50,6 +50,41 @@ import { MonacoQuickInputImplementation } from './monaco-quick-input-service';
...
+    const originalAvailableVerticalSpaceAbove = ContentHoverWidget.prototype['_availableVerticalSpaceAbove'];
+    ContentHoverWidget.prototype['_availableVerticalSpaceAbove'] = function (position: IPosition): number | undefined {
+        const value = originalAvailableVerticalSpaceAbove.call(this, position);
+        return typeof value === 'number' ? value - topHeightDiff() : undefined;
+    };
+
+    const originalAvailableVerticalSpaceBelow = ContentHoverWidget.prototype['_availableVerticalSpaceBelow'];
+    ContentHoverWidget.prototype['_availableVerticalSpaceBelow'] = function (position: IPosition): number | undefined {
+        const value = originalAvailableVerticalSpaceBelow.call(this, position);
+        return typeof value === 'number' ? value + 2 : undefined;
```
martin-fleck-at, line 77: Where is this 2 coming from? I assume it is the bottomHeightDiff as it steams from the 24 - 22? Shouldn't we also hand that in somehow from the outside like the top height? Also, I'm not too happy about the hard-coded 22, shouldn't we much rather determine this through `statusBar.node.getBoundingClientRect().height`?

martin-fleck-at (commented):
The change looks good to me overall and seems to work well, thank you Akos! I'm just a bit worried about the hard-coded numbers and the difference between handling the top diff vs the bottom diff. Maybe you could elaborate on that a bit more.

packages/toolbar/src/browser/application-shell-with-toolbar-override.ts
kittaakos: > Could you elaborate where the 32 comes from? In the browser I see that the top panel has 32px but the toolbar only has a height of 30px.

Thank you! I updated the PR. No ~32~ (which was incorrect) magic number, but the height is calculated from the toolbar + topPanel's bound rects.

I resolve this thread; please comment and reopen it if you disagree.

packages/monaco/src/browser/monaco-init.ts
kittaakos: > Where is this 2 coming from? I assume it is the bottomHeightDiff as it steams from the 24 - 22?

Correct.

> Shouldn't we also hand that in somehow from the outside like the top height?

Theia can do this, but the bottom height doesn’t change. Why do you think it's a good idea to set this from the outside, for example, through the status bar (from the core package)? This would also introduce a back dependency.

> shouldn't we much rather determine this through statusBar.node.getBoundingClientRect().height?

Sure, Theia should calculate it from the status bar node, but this code cannot access the status bar instance without a query selector. I’m open to calculating it once and storing the client height, but I don't see dependency injection (DI) as a feasible option here.

packages/monaco/src/browser/monaco-init.ts
```diff
@@ -50,6 +50,56 @@ import { MonacoQuickInputImplementation } from './monaco-quick-input-service';
...
+// https://github.com/eclipse-theia/theia/blob/b752ea690bdc4e7c5d9ab98a138504ead05be0d1/packages/core/src/browser/style/menus.css#L22
+// https://github.com/eclipse-theia/theia/blob/b752ea690bdc4e7c5d9ab98a138504ead05be0d1/packages/core/src/browser/style/status-bar.css#L18
+// https://github.com/eclipse-theia/theia/issues/14826
+function patchContentHoverWidget(topPanelHeight = 32): { setActualTopHeightForContentHoverWidget: (value: number) => void } {
+    let _actualTopHeight = topPanelHeight;
+    function getTopHeightDiff(): number {
+        return _actualTopHeight - VSCODE_TOP_HEIGHT;
+    }
+
+    const actualBottomHeight = 22; // Theia's status bar height
```
kittaakos, line 70: maybe we can move this to an arg with default value export the function, and if downstream projects redefine the status bar height, they can reuse this code as-is without further prototype patching

packages/monaco/src/browser/monaco-init.ts
martin-fleck-at: I think the bottom height can actually change quite easily. Theia exposes the `--theia-statusBar-height` CSS variable so adopters can very easily define another height for the status bar which should be considered correctly.

However, I do agree that dependency injection is definitely not a feasible option here. Much rather, it may make more sense to add an argument to `setActualTopHeightForContentHoverWidget` so that we can set the top AND the bottom height from the application shell where we do already have access to the status bar, the top panel, the tool bar, etc. In that case the method should probably drop the `Top` in the name. So basically I'm arguing that we should handle the bottom height the same way we handle the top height, what do you think?

packages/monaco/src/browser/monaco-init.ts
martin-fleck-at: You mean like we do for `actualTopHeight`? That seems reasonable.

kittaakos: Please see the new API proposal.

packages/monaco/src/browser/monaco-init.ts
kittaakos: It's implemented as requested. Thanks for the suggestion.

kittaakos:
In-action with https://github.com/eclipse-theia/theia/pull/14836/commits/c1980286fd9bb39c4b28f326dc2997bfbab5897b:


https://github.com/user-attachments/assets/1ed5080b-77ac-4b7d-acef-ee3be1d1599c

https://github.com/user-attachments/assets/b06f121f-a0db-430a-b10a-dfabe3f577ca

kittaakos:
> Do you maybe want to add your company or personal name to that?

If it's optional to leave it blank for now, I would choose to do so; I’m not quite sure what should or could go there at this time. Thank you!

martin-fleck-at (approved):
LGTM!

Issue #14843, opened 2025-02-05
[dev] `npm run watch:browser` rebuilds `ripgrep.js` and `bindings.js` in an infinite loop


kittaakos:
### Bug Description:

This is a solely dev problem. Maybe my dev steps are the problem, but I have noticed when I run Theia in watch mode, webpack constantly rebuilds some backend modules.

```
[bundle] assets by status 17.6 MiB [cached] 27 assets
[bundle] cached modules 15.1 MiB (javascript) 33.3 KiB (runtime) [cached] 2434 modules
[bundle] javascript modules 385 bytes
[bundle]   ./lib/backend/native-webpack-plugin/ripgrep.js 131 bytes [built]
[bundle]   ./lib/backend/native-webpack-plugin/bindings.js 254 bytes [built]
[bundle]   
[bundle] webpack 5.97.1 compiled successfully in 208 ms
```

<details><summary>Click to see the longer version</summary>
<p>

```
[bundle]   
[bundle] webpack 5.97.1 compiled successfully in 207 ms
[bundle] 
[bundle] assets by status 17.6 MiB [cached] 27 assets
[bundle] cached modules 15.1 MiB (javascript) 33.3 KiB (runtime) [cached] 2434 modules
[bundle] javascript modules 385 bytes
[bundle]   ./lib/backend/native-webpack-plugin/ripgrep.js 131 bytes [built]
[bundle]   ./lib/backend/native-webpack-plugin/bindings.js 254 bytes [built]
[bundle]   
[bundle] webpack 5.97.1 compiled successfully in 216 ms
[bundle] 
[bundle] assets by status 17.6 MiB [cached] 27 assets
[bundle] cached modules 15.1 MiB (javascript) 33.3 KiB (runtime) [cached] 2434 modules
[bundle] javascript modules 385 bytes
[bundle]   ./lib/backend/native-webpack-plugin/ripgrep.js 131 bytes [built]
[bundle]   ./lib/backend/native-webpack-plugin/bindings.js 254 bytes [built]
[bundle]   
[bundle] webpack 5.97.1 compiled successfully in 202 ms
[bundle] 
[bundle] assets by status 17.6 MiB [cached] 27 assets
[bundle] cached modules 15.1 MiB (javascript) 33.3 KiB (runtime) [cached] 2434 modules
[bundle] javascript modules 385 bytes
[bundle]   ./lib/backend/native-webpack-plugin/ripgrep.js 131 bytes [built]
[...]

PR #14844, opened 2025-02-05
Fix webpack watching

msujew:
#### What it does

Fixes https://github.com/eclipse-theia/theia/issues/14843

It seems like with a recent webpack upgrade, the watching behavior changed. This change only writes changes to the replacement files in case the file content differs from disk. Otherwise, the contents stay as they are.

#### How to test

Run `npm run watch:browser`. Assert that it doesn't run into an infinite loop.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/native-webpack-plugin/src/native-webpack-plugin.ts
+++ b/dev-packages/native-webpack-plugin/src/native-webpack-plugin.ts
@@ -58,15 +58,7 @@ export class NativeWebpackPlugin {
         let ripgrepIssuer: string | undefined;
         compiler.hooks.initialize.tap(NativeWebpackPlugin.name, async () => {
             const directory = path.resolve(compiler.outputPath, 'native-webpack-plugin');
-            if (fs.existsSync(directory)) {
-                await fs.promises.rm(directory, {
-                    force: true,
-                    recursive: true
-                });
-            }
-            await fs.promises.mkdir(directory, {
-                recursive: true
-            });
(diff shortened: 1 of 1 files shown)


dev-packages/native-webpack-plugin/src/native-webpack-plugin.ts
```diff
@@ -58,15 +58,11 @@ export class NativeWebpackPlugin {
...
         compiler.hooks.initialize.tap(NativeWebpackPlugin.name, async () => {
             const directory = path.resolve(compiler.outputPath, 'native-webpack-plugin');
-            if (fs.existsSync(directory)) {
-                await fs.promises.rm(directory, {
-                    force: true,
+            if (!fs.existsSync(directory)) {
+                await fs.promises.mkdir(directory, {
                     recursive: true
                 });
             }
```
kittaakos, line 65:
```suggestion
            await fs.promises.mkdir(directory, { recursive: true });
```

What do you think? Isn't it the same and easier to read? It's better not to rely on sync checks and then dispatch some logic.

kittaakos (approved):
It's working great, thanks for the quick fix 👍

dev-packages/native-webpack-plugin/src/native-webpack-plugin.ts
msujew: You're right. I forgot that `mkdir` is a no-op in case the directory already exists :)

jbicker (approved):
Thank you!

Issue #8309, opened 2020-08-05
upgrade to latest oniguruma


akosyakov:
It should be possible since we support Node.js 12 now.

PR #14848, opened 2025-02-05
chore: update `vscode-oniguruma`+`vscode-textmate`

kittaakos:
#### What it does

This PR updates the dependencies to `vscode-oniguruma@2.0.1` and `vscode-textmate@9.2.0`.

Closes eclipse-theia/theia#8309

#### How to test

I verified a few languages with the browser and the electron examples.

I did not find any breaking likes in the changelogs:
 - https://github.com/microsoft/vscode-oniguruma/compare/v1.6.2...v2.0.1
 - https://github.com/microsoft/vscode-textmate/compare/v9.1.0...v9.2.0 (there is no `v9.0.0` tag!)

https://github.com/user-attachments/assets/d3e2488e-ded9-458f-881e-b44fec951af0

https://github.com/user-attachments/assets/70d8db16-e56f-4921-9ec8-049559ae207a

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/plugin-contribution-handler.ts
+++ b/packages/plugin-ext/src/main/browser/plugin-contribution-handler.ts
@@ -16,7 +16,6 @@
 
 import { injectable, inject, named } from '@theia/core/shared/inversify';
 import { ITokenTypeMap, IEmbeddedLanguagesMap } from 'vscode-textmate';
-import { StandardTokenType } from 'vscode-textmate/release/encodedTokenAttributes';
 import { TextmateRegistry, getEncodedLanguageId, MonacoTextmateService, GrammarDefinition } from '@theia/monaco/lib/browser/textmate';
 import { MenusContributionPointHandler } from './menus/menus-contribution-handler';
 import { PluginViewRegistry } from './view/plugin-view-registry';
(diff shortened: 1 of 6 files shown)


packages/plugin-ext/src/main/browser/plugin-contribution-handler.ts
```diff
@@ -55,6 +54,14 @@ import { ContextKeyService } from '@theia/core/lib/browser/context-key-service';
...
 import { ThemeIcon } from '@theia/monaco-editor-core/esm/vs/base/common/themables';
 
+// The enum export is missing from `vscode-textmate@9.2.0`
+const enum StandardTokenType {
+    Other = 0,
+    Comment = 1,
+    String = 2,
+    RegEx = 3
+}
+
```
sdirix, line 64: The export still exists, see [here](https://github.com/microsoft/vscode-textmate/blob/f0f2a44b4c055b0e0fa3869e5de338efda5b7cf1/src/encodedTokenAttributes.ts#L159-L165).

The change to this file should be reverted.

sdirix (changes requested):
Works for me! I didn't see any regression in Browser or Electron.

However one adaption in this PR seems unnecessary and potentially dangerous if it gets out of sync with the lib.

packages/plugin-ext/src/main/browser/plugin-contribution-handler.ts
kittaakos: > Edit: My mistake. The export exists but it's not part of the release build.
> 
> However it's also defined here `@theia/monaco-editor-core/esm/vs/editor/common/encodedTokenAttributes`, so we might want to use that one? Not sure. What do you think?

I can provide an update, but it does not matter; they're the same. Please let me know.

I think it's a bug in the vscode-textmate API and could be the side-effect of an API extractor in https://github.com/microsoft/vscode-textmate/commit/62fcd6fd4b150a570bded80b6de9bc74bff3a1b1.

But I am just guessing what has caused the API break; they did not create a tag:

> * [microsoft/vscode-textmate@v9.1.0...v9.2.0](https://github.com/microsoft/vscode-textmate/compare/v9.1.0...v9.2.0) (there is no `v9.0.0` tag!)

sdirix (approved):

msujew (approved):
Works for me as well 👍

Issue #14858, opened 2025-02-06
Log config file watching does not longer work


msujew:
### Bug Description:

I've recently noticed that the log config file watching feature no longer works. I probably accidentally broke it in https://github.com/eclipse-theia/theia/pull/14194.

### Steps to Reproduce:

1. Instruct Theia to use a `log-config` file via the CLI
2. Use the `defaultLevel: error` in the file initially
3. Change the `defaultLevel` to `trace`
4. No trace related logs get reported.

### Additional Information

- Operating System: Windows
- Theia Version: 1.58.1

PR #14861, opened 2025-02-06
Streamline logging API

msujew:
#### What it does

Fixes https://github.com/eclipse-theia/theia/issues/14858

This change fixes a bunch of issues that have been bugging me about the logging API. Due to the previous way of how the binding was done for the `@named` decorator, creating a dedicated logger was quite a lot of work (at least compared to the new binding logic). Now, named loggers are simply created on the fly.

Also adds a new `log-file` argument to the CLI. This isn't strictly necessary, as operating systems offer ways of streaming the output of a executable into files, but I find this to be pretty useful anyways, even if just for discoverability.

#### How to test

1. Ensure that named loggers still work as expected and they are correctly generated on the fly.
2. Use the new `log-file` CLI argument to see whether this works as expected.
3. Ensure that https://github.com/eclipse-theia/theia/issues/14858 has been fixed by following the repro instructions.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/node/console-logger-server.ts
+++ b/packages/core/src/node/console-logger-server.ts
@@ -20,7 +20,7 @@ import { LogLevelCliContribution } from './logger-cli-contribution';
 import { ILoggerServer, ILoggerClient, ConsoleLogger, rootLoggerName } from '../common/logger-protocol';
 import { format } from 'util';
 import { EOL } from 'os';
-import * as fs from 'fs/promises';
+import * as fs from 'fs';
 
 @injectable()
 export class ConsoleLoggerServer implements ILoggerServer {
@@ -33,7 +33,7 @@ export class ConsoleLoggerServer implements ILoggerServer {
     @inject(LogLevelCliContribution)
     protected cli: LogLevelCliContribution;
 
-    protected failedLogFileWrite = false;
+    protected logFileStream?: fs.WriteStream;
 
     @postConstruct()
     protected init(): void {
@@ -65,22 +65,21 @@ export class ConsoleLoggerServer implements ILoggerServer {
         const configuredLogLevel = await this.getLogLevel(name);
         if (logLevel >= configuredLogLevel) {
             const fullMessage = ConsoleLogger.log(name, logLevel, message, params);
-            await this.logToFile(fullMessage, params);
+            this.logToFile(fullMessage, params);
         }
     }
 
-    protected async logToFile(message: string, params: any[]): Promise<void> {
-        if (this.cli.logFile) {
-            try {
-                const formatted = format(message, ...params) + EOL;
-                await fs.appendFile(this.cli.logFile, formatted);
-            } catch (error) {
-                // Avoid spamming the console with errors in case the log file is not writable
-                if (!this.failedLogFileWrite) {
-                    console.error(`Failed to write to log file: ${error}`);
-                    this.failedLogFileWrite = true;
-                }
-            }
+    protected logToFile(message: string, params: any[]): void {
+        if (this.cli.logFile && !this.logFileStream) {
+            this.logFileStream = fs.createWriteStream(this.cli.logFile, { flags: 'a' });
+            // Only log errors once to avoid spamming the console
+            this.logFileStream.once('error', error => {
+                console.error(`Error writing to log file ${this.cli.logFile}`, error);
+            });
+        }
+        if (this.logFileStream) {
+            const formatted = format(message, ...params) + EOL;
+            this.logFileStream.write(formatted);
         }
     }
 
--- a/packages/core/src/node/console-logger-server.ts
+++ b/packages/core/src/node/console-logger-server.ts
@@ -18,17 +18,22 @@ import { inject, injectable, postConstruct } from 'inversify';
 import { LoggerWatcher } from '../common/logger-watcher';
 import { LogLevelCliContribution } from './logger-cli-contribution';
 import { ILoggerServer, ILoggerClient, ConsoleLogger, rootLoggerName } from '../common/logger-protocol';
+import { format } from 'util';
+import { EOL } from 'os';
+import * as fs from 'fs/promises';
 
 @injectable()
 export class ConsoleLoggerServer implements ILoggerServer {
 
-    protected client: ILoggerClient | undefined = undefined;
+    protected client?: ILoggerClient;
 
     @inject(LoggerWatcher)
     protected watcher: LoggerWatcher;
 
     @inject(LogLevelCliContribution)
     protected cli: LogLevelCliContribution;
+
+    protected failedLogFileWrite = false;
 
     @postConstruct()
     protected init(): void {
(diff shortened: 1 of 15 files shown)


packages/core/src/node/console-logger-server.ts
```diff
@@ -59,7 +64,23 @@ export class ConsoleLoggerServer implements ILoggerServer {
...
+            const fullMessage = ConsoleLogger.log(name, logLevel, message, params);
+            await this.logToFile(fullMessage, params);
+        }
+    }
+
+    protected async logToFile(message: string, params: any[]): Promise<void> {
+        if (this.cli.logFile) {
+            try {
+                const formatted = format(message, ...params) + EOL;
+                await fs.appendFile(this.cli.logFile, formatted);
```
kittaakos, line 76: It may be overkill, but Theia could create the write stream once with the append flag and push the messages into it instead of always writing the file. I have not tried it, but it may be more efficient.

```js
const s = fs.createWriteStream(this.cli.logFile, { flags: 'a' });
// later
if (!this.failedLogFileWrite) s.push(formatted);
```

msujew: I agree, using a file stream is probably the better solution. I've adjusted the code to use a file stream.

packages/core/src/node/console-logger-server.ts
```diff
@@ -59,7 +64,22 @@ export class ConsoleLoggerServer implements ILoggerServer {
     async log(name: string, logLevel: number, message: string, params: any[]): Promise<void> {
         const configuredLogLevel = await this.getLogLevel(name);
         if (logLevel >= configuredLogLevel) {
-            ConsoleLogger.log(name, logLevel, message, params);
+            const fullMessage = ConsoleLogger.log(name, logLevel, message, params);
+            this.logToFile(fullMessage, params);
```
kittaakos, line 68: ConsoleLogger.log provides the formatter message. Why must the params be passed down and reformat with utils again?

msujew: `ConsoleLogger.log` doesn't include the parameters into the message, but passes them into `console.<method>`. `util.format` will then create a human readable string from message + parameters.

packages/core/src/node/logger-cli-contribution.ts
```diff
@@ -87,22 +99,36 @@ export class LogLevelCliContribution implements CliContribution {
                 console.error(`Error reading log config file ${filename}: ${e}`);
             }
         }
+
+        if (args['log-file'] !== undefined) {
+            let filename: string = args['log-file'] as string;
+            try {
+                filename = path.resolve(filename);
+                await fs.writeFile(filename, '');
```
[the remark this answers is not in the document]
msujew: I'm not sure. I personally think that it should truncate the file, but there might be other opinions. @tsmaeder Care to chime in?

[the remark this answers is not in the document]
tsmaeder: I see three alternatives:
1. We append to the file.
That would mix different runs of Theia in the log file. I don't think that is what I would want.
2. We truncate the file
3. If the file already exists, we rename it to something else.
If we go with this option, it would probably make sense whenever we write a log file, not just the when we an explicit parameter.

@kittaakos why would you not truncate the file? What do you propose instead?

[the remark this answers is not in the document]
kittaakos: > why would you not truncate the file? 

To avoid unintentionally destroying the log file contents by starting the server again.

> What do you propose instead?

All three of the above options are acceptable.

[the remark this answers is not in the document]
sdirix: Generally speaking I'm fine with all three approaches, however personally I would go with the rename approach.

When I'm passing the log-file parameter to Theia then explicitly because I want to keep the logs. Overwriting seems counterproductive to me in that case. I would rename the old file to something like `<old-name><creation-time-stamp>.old`, e.g. `theia.log.2025-02-07T12-34-56.old` or something similar. It's just too inconvenient to change the parameter with every invocation.

packages/core/src/node/logger-cli-contribution.ts
```diff
@@ -87,22 +99,46 @@ export class LogLevelCliContribution implements CliContribution {
                 console.error(`Error reading log config file ${filename}: ${e}`);
             }
         }
+
+        if (args['log-file'] !== undefined) {
+            let filename: string = args['log-file'] as string;
```
kittaakos, line 104:
```suggestion
            let filename = args['log-file'] as string;
```

packages/core/src/node/logger-cli-contribution.ts
```diff
@@ -87,22 +99,46 @@ export class LogLevelCliContribution implements CliContribution {
...
+            try {
+                filename = path.resolve(filename);
+                try {
+                    const stat = await fs.stat(filename);
+                    if (stat && stat.isFile()) {
+                        // Rename the previous log file to avoid overwriting it
+                        const oldFilename = `${filename}.${stat.ctime.toISOString().replace(/:/g, '-')}.old`;
+                        await fs.rename(filename, oldFilename);
+                    }
+                } catch (err) {
```
kittaakos, line 114:
```suggestion
                } catch {
```

packages/core/src/node/console-logger-server.ts
```diff
@@ -59,7 +64,22 @@ export class ConsoleLoggerServer implements ILoggerServer {
...
+        }
+    }
+
+    protected logToFile(message: string, params: any[]): void {
+        if (this.cli.logFile && !this.logFileStream) {
+            this.logFileStream = fs.createWriteStream(this.cli.logFile, { flags: 'a' });
+            // Only log errors once to avoid spamming the console
+            this.logFileStream.once('error', error => {
+                console.error(`Error writing to log file ${this.cli.logFile}`, error);
```
kittaakos, line 77: If the log file creation fails for whatever reason, both `this.cli.logFile` and `this.logFileStream` will be falsy, and the error message about the log file write failure will never be visible.

Screenshot 2025-02-11 at 08 23 04

_I made the log file creation fail by providing an invalid filename (`--log-file=&&&\\//.xxx`), but the file creation could fail due to another error, for example, `EPERM`._

kittaakos (approved):
I have tried it, and it worked very well for me. Thanks

packages/core/src/node/console-logger-server.ts
kittaakos: Closing as invalid. It's literally two entries above that the log file creation has failed.

sdirix (commented):
Works for me! 

I can confirm that the log config works again, the log-file and log-file renaming works too. Also logging works in general ;)

Note that this is a breaking change because if adopters followed the Theia conventions for creating their loggers, they will now face errors like this
>Uncaught Exception:  Error: Ambiguous match found for serviceIdentifier: Symbol(ILogger)
Error: Ambiguous match found for serviceIdentifier: Symbol(ILogger)

So we should mention that in the changelog I think

msujew:
@sdirix Thank you for the hint. I didn't think about that. I added a breaking changelog entry for that :)

sdirix (approved):

yduuuuuuuu:
Hello! Does it work for the package as well ? Can we get the log file if we run the .exe ?

msujew:
> Hello! Does it work for the package as well ? Is there any way for us to get the log file if we run the .exe ?

@yduuuuuuuu Yes, you just need to start the `.exe` with the `--log-file=./log.txt` CLI argument.

Issue #7917, opened 2020-05-28
[debug] Same debug/launch config can be started in parallel


kittaakos:
### Bug Description:

VS Code does not allow starting the "same" thing twice:

![Screen Shot 2020-05-28 at 15 25 02](https://user-images.githubusercontent.com/1405703/83147151-911a2f00-a0f7-11ea-8b8d-a995528a6831.jpg)

Theia allows launching the same module in parallel:

![Screen Shot 2020-05-28 at 15 26 29](https://user-images.githubusercontent.com/1405703/83147229-b0b15780-a0f7-11ea-99a7-82255cd05934.jpg)

Is it a bug or a feature? I would have expected the same behavior.

### Steps to Reproduce:

1. Create a `index.js`, set a breakpoint in the module and start `Launch with Node.js`
2. Start `Launch with Node.js` again, you can. VS Code does not allow it.
3.

### Additional Information

- Operating System:
- Theia Version:

PR #14862, opened 2025-02-06
fix: warn user before starting the same debug session multiple times

kittaakos:
#### What it does

When a debug session is initiated by the user, and the `suppressMultipleSessionWarning` is not set to `true` in the launch config, prompt the user before starting a new debug session with the same label.

- Introduce an optional `suppressMultipleSessionWarning` prop to the debug config.
- Add an optional `startedByUser` prop to the session options.
- Add `getLabel` to `DebugSession`

Closes eclipse-theia/theia#7917

#### How to test

Try to start the "same" debug config twice; Theia will warn you unless the `suppressMultipleSessionWarning` is explicitly set to `true`.

https://github.com/user-attachments/assets/66706c66-44de-434e-a85c-353212bf00ac

https://github.com/user-attachments/assets/13cf1540-c39e-471b-8685-6a35964218b2

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


packages/debug/src/browser/debug-session-options.ts
```diff
@@ -32,21 +32,22 @@ export class DebugCompoundRoot {
...
+    controllerId: string;
+    runId: string;
 }
 
 export interface DebugSessionOptionsBase {
-    workspaceFolderUri?: string,
-    testRun?: TestRunReference
+    workspaceFolderUri?: string;
+    testRun?: TestRunReference;
+    startedByUser?: boolean;
```
sdirix, line 42: The option `startedByUser` does not indicate the consequence of this option. For example without looking at the code it's not clear whether this means that the user already confirmed a "force" start.

Should we rename this to `checkForMultipleSessions` or `warnIfMultipleSessions`? Then it's clear what will happen if it's set to `true`.

packages/debug/src/common/debug-configuration.ts
```diff
@@ -84,6 +84,9 @@ export interface DebugConfiguration {
 
     /** When true, the debug viewlet will not be automatically revealed for this session. */
     suppressDebugView?: boolean;
+
+    /** Disable the warning when trying to start the same debug configuration more than once. `true` by default. */
+    suppressMultipleSessionWarning?: boolean;
```
sdirix, line 89: Note that the current PR behaves as if `false` is default, i.e. it will warn when the option is not given

sdirix (changes requested):
Works for me! Thanks for the improvement.

I left some comments.

packages/debug/src/browser/debug-session-label-provider.ts
```diff
@@ -0,0 +1,44 @@
+// *****************************************************************************
+// Copyright (C) 2025 and others.
```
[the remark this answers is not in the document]
kittaakos: Thanks for keeping an eye on it. No, I do not. Same reasoning: https://github.com/eclipse-theia/theia/pull/14836#issuecomment-2637406454.

packages/plugin-ext/src/main/browser/debug/plugin-debug-service.ts
```diff
@@ -360,6 +360,11 @@ export class PluginDebugService implements DebugService {
                 default: 'openOnFirstSessionStart',
                 description: nls.localizeByDefault('Controls when the internal Debug Console should open.')
             };
+            properties['suppressMultipleSessionWarning'] = {
+                type: 'boolean',
+                description: nls.localizeByDefault('Disable the warning when trying to start the same debug configuration more than once.'),
+                default: true
```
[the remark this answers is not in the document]
kittaakos: > I think this should be `false` by default. It's much more likely the user confused the continue button with the start debug button than that they actually want to run a debug configuration twice.
> 
> Note that the current PR behaves as if `false` is default, i.e. it will warn when the option is not given

Very nice. I appreciate you taking the time to think it through. I was also tricked by this. Actually, it is [verbatim copied from VS Code](https://github.com/microsoft/vscode/blob/853c9a6ccfc07063bf986155ac763476415d7122/src/vs/workbench/contrib/debug/browser/debugAdapterManager.ts#L203).

As a user, the main reason for including this property in your launch configuration is to disable the default behavior by setting it to `true`. It wouldn't make sense to set it to any value other than `true`, as doing so would effectively be a no-op. Therefore, when a user wants to use this prop, it should always be set to `true`. The content assistance in the launch editor will provide this default value. I hope this clarifies things.

I have tested the UX myself during the development, and there is a good reason to keep it `true`. You might also want to try it out. What do you think?

packages/debug/src/browser/debug-session-label-provider.ts
```diff
@@ -0,0 +1,44 @@
...
+        let { name, workspaceFolderUri } = params;
+        const includeRoot = this.workspaceService.tryGetRoots.length > 1;
+        if (includeRoot && workspaceFolderUri) {
+            const uri = new URI(workspaceFolderUri);
+            const path = uri.path;
+            const basenameOrAuthority = path.name || uri.authority;
+            name += ` (${basenameOrAuthority})`;
+        }
+        return name;
+    }
```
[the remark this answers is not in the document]
kittaakos: > Instead of copying the VS Code behavior, we could reuse our own.

It would be best if the Theia implementation worked as Code, and it would be easier to keep things compatible. What do you think? I can update and reuse the label provider in Theia here, but I prefer keeping the VS Code behavior.

> We already have this logic in place in the UI, see [here](https://github.com/eclipse-theia/theia/blob/9a20727743172a4efa46334f4136a1242b160848/packages/debug/src/browser/view/debug-configuration-select.tsx#L258-L268).

Yes, there is a similar one.

I bet there was a good reason they picked the authority as the qualifier if there is no basename of path, perhaps a case we have not even considered.

packages/debug/src/browser/debug-session-options.ts
kittaakos: > For example without looking at the code it's not clear whether this means that the user already confirmed a "force" start.

Looking at the VS Code sources and seeing where they are set to true gave me quite a good understanding of how this should behave. 

> Should we rename

No, I prefer to keep it the way it is named in the VS Code. Every time there is a compatibility issue in Theia, we will look into VS Code. Naming things the same will make keeping things running and compatible easier.

sdirix: I did not check the VS code sources. If it's named the same there, then I'm fine with it :+1:

packages/debug/src/common/debug-configuration.ts
kittaakos: Fixed via https://github.com/eclipse-theia/theia/pull/14862/commits/df5ec4dd2ac57a5e7c4a6940210b02a8c0278509.

packages/debug/src/browser/view/debug-view-model.ts
```diff
@@ -136,12 +137,12 @@ export class DebugViewModel implements Disposable {
...
     }
 
-    async start(): Promise<void> {
+    async start(options: Partial<Pick<DebugSessionOptionsBase, 'startedByUser'>> = {}): Promise<void> {
         const { session } = this;
         if (!session) {
             return;
         }
-        const newSession = await this.manager.start(session.options);
+        const newSession = await this.manager.start(Object.assign(session.options, options));
```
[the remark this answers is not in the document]
kittaakos: Fixed via https://github.com/eclipse-theia/theia/pull/14862/commits/f0e27c7e0e0c10da2b3b4bd8f652a13ee93b8b6b.

packages/debug/src/browser/debug-session-label-provider.ts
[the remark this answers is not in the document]
kittaakos: https://github.com/eclipse-theia/theia/pull/14862/commits/957c4c32d1616287569d0099bc65f37d1c569da0

packages/debug/src/browser/debug-session-label-provider.ts
```diff
@@ -0,0 +1,46 @@
...
+@injectable()
+export class DebugSessionLabelProvider {
+
+    @inject(WorkspaceService)
+    protected readonly workspaceService: WorkspaceService;
+
+    // https://github.com/microsoft/vscode/blob/907518a25c6d6b9467cbcc57132c6adb7e7396b0/src/vs/workbench/contrib/debug/browser/debugSession.ts#L253-L256
+    getLabel(
+        params: Pick<DebugSessionOptions, 'name' | 'workspaceFolderUri'>,
+        includeRoot = this.workspaceService.tryGetRoots.length > 1
```
sdirix, line 35: This is always false, we need to invoke the function ;)
In the view this was not noticeable as it hands over `includeRoot` itself, but the multiple-check did not differentiate between launch configurations from different roots.
```suggestion
        includeRoot = this.workspaceService.tryGetRoots().length > 1
```

packages/debug/src/browser/debug-session-label-provider.ts
```diff
@@ -0,0 +1,46 @@
...
+import { inject, injectable } from '@theia/core/shared/inversify';
+import { WorkspaceService } from '@theia/workspace/lib/browser';
+import { type DebugSessionOptions } from './debug-session-options';
+
+/**
+ * Provides a label for the debug session without the need to create the session.
+ * Debug session labels are used to check if sessions are the "same".
+ */
+@injectable()
+export class DebugSessionLabelProvider {
```
sdirix, line 27: We should call this `DebugSessionConfigurationLabelProvider` to better express its functionality. I noticed that each `DebugSession` actually already has a `LabelProvider` attached with it, so this could be misleading overall.

sdirix (changes requested):
Thanks for the refactoring. I found an issue: The current code does not differentiate between launch configurations of different roots, i.e. a launch configuration with the same name from a different root will show a warning immediately. More details in the comments.

packages/debug/src/browser/debug-session-label-provider.ts
kittaakos: https://github.com/eclipse-theia/theia/pull/14862/commits/142cd4feda368ea7e647736d87ee808f43636119

Issue #11955, opened 2022-12-05
debug: `watch expressions` are not updated properly


vince-fugnitto:
#### Bug Description:

There is an issue with `watch expressions` not being updated in the debug-view which causes a difference in behavior with vscode and user expectations.

_For example:_

If we add a `watch expression` with no debug session present, we correctly get information that the expression is not available.

![image](https://user-images.githubusercontent.com/40359487/205692932-18a7abb7-5f08-4def-ad43-2c9869d80a24.png)

However, if we start a debug session and hit the breakpoint (making the variable `value` assigned), the watch expression is not updated:

![image](https://user-images.githubusercontent.com/40359487/205692875-4408eba5-fea2-46c8-a764-cecc831a8df1.png)

If we edit the watch expression with the same name (`value`) then it gets updated:

![image](https://user-images.githubusercontent.com/40359487/205693985-b1336a79-e5a8-4ece-aab5-630010273943.png)

#### Steps to Reproduce:

1. start the application using theia as a workspace
2. open the `cli.spec.ts` file and set a breakpoint at line 38 (after `const value`)
3. open the debug-view and add a watch expression for `value`
4. the watch expression should include the message "not available"
5. start the debug session with "run mocha tests"
6. the breakpoint should be hit, but the watch expression is not updated 
7. right-click the `value` watch expression, select "edit expression" and press "enter" (keeping `value` as a name)
8. the watch expression will be properly updated

kittaakos:
This is because the very first 'evaluate' happens too early, and the VS Code debug JS extension does not even reject that request: https://github.com/microsoft/vscode-js-debug/blob/d7730405d64e81eb364f76ea38efb7c93a5f877d/src/dap/connection.ts#L172

Since there is a queue of promises to update the watch expression in Theia, and the first neither rejects nor resolves, then the next updates will also wait forever:

https://github.com/eclipse-theia/theia/blob/e7619c59d8f1da42d9c298c274945b6aef4919eb/packages/debug/src/browser/view/debug-view-model.ts#L218-L228

If the 'evaluate' promises are not queued, the console error can be seen in the backend log, but there is no rejection:

```
2025-02-07T14:13:38.292Z root INFO REFRESHING WATCH EXPRESSIONS 5
2025-02-07T14:13:38.293Z root INFO EVALUATING WATCH EXPRESSION 5 nestedObject
2025-02-07T14:13:38.293Z root INFO REFRESHING WATCH EXPRESSIONS 5
[...]

PR #14874, opened 2025-02-07
fix: no watch evaluation if no current stack frame

kittaakos:
#### What it does

The `vscode-js-debug` does not reject when the `evaluate` request arrives too early to the debug adapter and the evaluate handler is not yet registered. See the details at https://github.com/eclipse-theia/theia/issues/11955#issuecomment-2643161927.

Ref: https://github.com/microsoft/vscode-js-debug/blob/d7730405d64e81eb364f76ea38efb7c93a5f877d/src/dap/connection.ts#L172

If the first evaluate request is never rejected or resolved, none of the following evaluate requests will be, as they are queued, making the Theia UI never update.

This PR changes how Theia evaluates the expressions: no evaluation happens when there is no current stack frame.
_(I debugged into VS Code, and they do not evaluate when there is no current stack frame. I think this makes sense, though I might have overlooked something.)_

When there is no active debug session or there is no current stack frame, use <i>not available</i>, just like VS Code does.

Closes eclipse-theia/theia#11955

Theia:

https://github.com/user-attachments/assets/6df170aa-0783-4e0a-9292-1a74a3ec7b3a

VS Code:

https://github.com/user-attachments/assets/e98f0f60-f61b-40a1-85c3-dbbb421cc869

#### How to test

Follow the steps from #11955.

#### Follow-ups

~I wonder why the vscode-js-debug sends the initialized when it cannot yet accept requests?~
 - ~https://github.com/microsoft/vscode-js-debug/blob/d7730405d64e81eb364f76ea38efb7c93a5f877d/src/dap/connection.ts#L42-L48~
 - ~https://github.com/microsoft/vscode-js-debug/blob/d7730405d64e81eb364f76ea38efb7c93a5f877d/src/dap/connection.ts#L184-L187~

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/view/debug-view-model.ts
+++ b/packages/debug/src/browser/view/debug-view-model.ts
@@ -220,7 +220,7 @@ export class DebugViewModel implements Disposable {
         this.refreshWatchExpressionsQueue = this.refreshWatchExpressionsQueue.then(async () => {
             try {
                 for (const watchExpression of this.watchExpressions) {
-                    await Promise.all([
+                    await Promise.race([
                         watchExpression.evaluate(),
                         // For example vscode-js-debug does not error when evaluating watch expressions "too early".
                         // https://github.com/eclipse-theia/theia/issues/11955#issuecomment-2643161927
--- a/packages/debug/src/browser/view/debug-view-model.ts
+++ b/packages/debug/src/browser/view/debug-view-model.ts
@@ -27,7 +27,6 @@ import { DebugWatchExpression } from './debug-watch-expression';
 import { DebugWatchManager } from '../debug-watch-manager';
 import { DebugFunctionBreakpoint } from '../model/debug-function-breakpoint';
 import { DebugInstructionBreakpoint } from '../model/debug-instruction-breakpoint';
-import { timeoutReject } from '@theia/core/lib/common/promise-util';
 
 @injectable()
 export class DebugViewModel implements Disposable {
@@ -219,14 +218,7 @@ export class DebugViewModel implements Disposable {
     protected refreshWatchExpressions = debounce(() => {
         this.refreshWatchExpressionsQueue = this.refreshWatchExpressionsQueue.then(async () => {
             try {
-                for (const watchExpression of this.watchExpressions) {
-                    await Promise.race([
-                        watchExpression.evaluate(),
-                        // For example vscode-js-debug does not error when evaluating watch expressions "too early".
-                        // https://github.com/eclipse-theia/theia/issues/11955#issuecomment-2643161927
-                        timeoutReject(1_000, 'Failed to refresh watch expressions: timeout'),
-                    ]);
-                }
+                await Promise.all(Array.from(this.watchExpressions).map(expr => expr.evaluate()));
             } catch (e) {
                 console.error('Failed to refresh watch expressions: ', e);
             }
--- a/packages/debug/src/browser/view/debug-view-model.ts
+++ b/packages/debug/src/browser/view/debug-view-model.ts
@@ -27,6 +27,7 @@ import { DebugWatchExpression } from './debug-watch-expression';
 import { DebugWatchManager } from '../debug-watch-manager';
 import { DebugFunctionBreakpoint } from '../model/debug-function-breakpoint';
 import { DebugInstructionBreakpoint } from '../model/debug-instruction-breakpoint';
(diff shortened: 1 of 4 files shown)


packages/debug/src/browser/view/debug-view-model.ts
```diff
@@ -219,7 +220,12 @@ export class DebugViewModel implements Disposable {
         this.refreshWatchExpressionsQueue = this.refreshWatchExpressionsQueue.then(async () => {
             try {
                 for (const watchExpression of this.watchExpressions) {
-                    await watchExpression.evaluate();
+                    await Promise.race([
+                        watchExpression.evaluate(),
+                        // For example vscode-js-debug does not error when evaluating watch expressions "too early".
+                        // https://github.com/eclipse-theia/theia/issues/11955#issuecomment-2643161927
+                        timeoutReject(1_000, 'Failed to refresh watch expressions: timeout'),
+                    ]);
```
msujew, line 228: Question: This creates a new `timeoutReject` promise for every individual watch expression. I'm wondering whether it makes sense to collect them and have a single `timeoutReject` for all of them? Something like this:

```ts
try {
    await Promise.race([
        Promise.all(this.watchExpressions.map(expr => expr.evaluate())),
        // For example vscode-js-debug does not error when evaluating watch expressions "too early".
        // https://github.com/eclipse-theia/theia/issues/11955#issuecomment-2643161927
        timeoutReject(1_000, 'Failed to refresh watch expressions: timeout')
    ]);
} catch (e) {
   ....
}
```

Or is it important to wait for each watch expressions individually?

packages/debug/src/browser/view/debug-view-model.ts
```diff
@@ -219,7 +220,12 @@ export class DebugViewModel implements Disposable {
         this.refreshWatchExpressionsQueue = this.refreshWatchExpressionsQueue.then(async () => {
             try {
                 for (const watchExpression of this.watchExpressions) {
-                    await watchExpression.evaluate();
+                    await Promise.race([
+                        watchExpression.evaluate(),
+                        // For example vscode-js-debug does not error when evaluating watch expressions "too early".
+                        // https://github.com/eclipse-theia/theia/issues/11955#issuecomment-2643161927
+                        timeoutReject(1_000, 'Failed to refresh watch expressions: timeout'),
```
msujew, line 227: Suggestion: I haven't tried it out, but with this error message the log will print `Failed to refresh watch expressions: Failed to refresh watch expressions: timeout`, right?
```suggestion
                        timeoutReject(1_000, 'timeout'),
```

packages/debug/src/browser/view/debug-view-model.ts
kittaakos: I like the idea of evaluating all expressions at once. However, the one-second timeout could be too low on a non-high-end dev machine. Even if Theia increases the timeout to, let's say, 3 seconds, it can be a problem in a slower environment.

I will put the PR back to draft and spend some time investigating whether Theia can defer evaluating the expressions and avoid using any timeouts.

kittaakos: Evaluating the expressions at once has been implemented at https://github.com/eclipse-theia/theia/pull/14874/commits/005e772ef5461bdb3d8dbd6eefc2b6327bdceca6#diff-36b6b1bf068f9c3e7bf28c4c23f3a6d3dc9cefd4d259b667f7443dc3f76c2a9cL222-R221.

packages/debug/src/browser/view/debug-view-model.ts
kittaakos: Thanks! I did not know this


The next proposal tries to solve this issue without timeouts. I am resolving this thread.

kittaakos:
@msujew, thanks for the review. I have another proposal without timeouts; it's open for discussion.

msujew (approved):
Works great, thank you :)

msujew:
@kittaakos Do you want to squash the commits yourself and create an appropriate commit description? Right now, it still references the timeout as the solution.

kittaakos:
> Do you want to squash the commits yourself and create an appropriate commit description? Right now, it still references the timeout as the solution.

If you could squash+merge it with the GH UI and give the commit a proper title, it would be great. I've already updated the PR title. Thank you

Otherwise, I can do it.

PR #14889, opened 2025-02-10
Use DiffEditor DiffNavigator to navigate diffs

### LINKED ISSUE #14888: Enable change navigation in diff editors
Currently, diff editors opened either by the Git extension or through AI interactions etc. don't show the arrows visible in VSCode, and navigating between changes in the file is not possible.

colin-grant-work:
#### What it does

Fixes #14888 by adding diff navigation commands to the toolbar of diff editors and using the existing diff navigator tool in `MonacoDiffEditors` to navigate among changes.

#### How to test

1. Open a diff editor with some changes.
> Common ways to do this: modify a file and then open it from the source control view; use the 'select for comparison...compare with selected' commands in the file tree; using the AI chat features, ask the `@Coder` agent to make modifications.
2. You should see up and down arrows on the toolbar, and clicking them should take to the next / previous change in the file, as appropriate.
3. If you open a diff editor with no differences, the arrows should appear, but they should be greyed out.

> @ reviewers, there was relevant work in #13104, but since that was mostly related to `@theia/git`, I'm not sure the machinery implemented there is relevant anymore. The code here just skips that machinery in the case that I believe I know how to handle better, but possibly it's just no longer necessary. In particular, I'm not sure how we would have a diff in an editor with a `file` scheme.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/scm/src/browser/scm-contribution.ts
+++ b/packages/scm/src/browser/scm-contribution.ts
@@ -43,8 +43,9 @@ import { ScmCommand } from './scm-provider';
 import { ScmDecorationsService } from '../browser/decorations/scm-decorations-service';
 import { nls } from '@theia/core/lib/common/nls';
 import { isHighContrast } from '@theia/core/lib/common/theme';
-import { EditorMainMenu } from '@theia/editor/lib/browser';
+import { EditorMainMenu, EditorWidget } from '@theia/editor/lib/browser';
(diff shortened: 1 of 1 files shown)


rschnekenbu (approved):
LGTM, @colin-grant-work. Thanks for contribution!

Issue #14768, opened 2025-01-24
[Theia AI] Coder and ChangeSet Epic


JonasHelming:

colin-grant-work:
> [Theia AI] Check diff approach | Are we uing the right diff editor? It does not allow typical features such as navigating between the different changes

This is true of diff editors opened by e.g. the Git extension for changes to the working tree, as well. We likely need to implement the relevant commands for all diff editors, and we'll get them on AI diff editors for free.

colin-grant-work:
> [Theia AI] Close diff editors of accepted or cleared changes | they are still open in dirty state which is very confusing
>
> [Theia AI] Diff editor asks for file location for new files | When saving a diff editor for a new file, it should not open the “save” file system dialog. And even if it does, then not on the root system location
> 
> [Theia AI] Diff editor behaves weird with auto save on | changes are directly applied

Fundamentally, what's the UX we want to get to re: the pending changes. The flow chart I see looks like something like:
Agent proposes changes, and we display them. Then...
User can open individual files
Interact and save --> counts as acceptance of state of editor at the time. Should remove that file from proposed changes(?)
Interact and close without saving --> neutral, no action re: change set node.
User can click accept on node --> If file not open, enact suggestion. If changes open, is that accepting the current state or the original suggestion?
User can click reject on node --> If file not open, just remove suggestion. If changes open and dirty --> revert to state before suggestions? Keep dirty editor available in case the user meant that they didn't want the suggestions as proposed, but maybe they've made progress in the dirty editor.
User can click Accept on change set representation --> delegate to individual acceptance actions.
User leaves suggestions around for a long time while they work on the relevant files in other ways (maybe even other chats) --> how to handle staleness if they come back?

In any case, it seems clear that autosaving in this context shouldn't count as 'normal' saving: we shouldn't count that as the user having actually accepted the changes, however acceptance is handled in our case. This is one argument against our current dirty default.

JonasHelming:
> > [Theia AI] Close diff editors of accepted or cleared changes | they are still open in dirty state which is very confusing
> > [Theia AI] Diff editor asks for file location for new files | When saving a diff editor for a new file, it should not open the “save” file system dialog. And even if it does, then not on the root system location
> > [Theia AI] Diff editor behaves weird with auto save on | changes are directly applied
> 
> Fundamentally, what's the UX we want to get to re: the pending changes. The flow chart I see looks like something like: Agent proposes changes, and we display them. Then...
> 
[...]

PR #14901, opened 2025-02-11
Update editors when ChangeElement added or replaced

### LINKED ISSUE #14900: [Theia AI] Diff editor does not update to new changes
When a diff editor is still open for a file and a new change for the same file was generated, then clicking on the change set diff will just focus the editor, although it does not contain the changes. I need to close the old editor and open it again via the new changeset to see the diff.

colin-grant-work:
#### What it does

Fixes #14900 by updating any open editors referring to a given change proposal when the change proposal is updated.

> I've put some thoughts together re: this problem [here](https://github.com/eclipse-theia/theia/issues/14768#issuecomment-2651484965). This PR implements the Monaco editor update solution proposed there. I have a branch that uses the resource-based approach, but that has the problem that we consider the editor 'dirty.' Would be happy to discuss implementing any of the solutions proposed there, or any others. This one isn't my favorite, because it feels like a weird place to assign responsibility for updating the editors, but it does get around the issue of dirty editors most directly.

#### How to test

1. Open a chat and use a prompt like
> `@Coder` Remove the prefix "MCP" from all user message in packages/ai-mcp/src/browser/mcp-command-contribution.ts
2. The agent should propose a change set. Open a diff for the file affected.
3. Prompt the agent again with something like.
> `@Coder` Actually, let's change the original MCP to CMP instead.
4. The agent should suggest a new set of changes, and the content of the right side of the diff should be updated accordingly.

> As I mention over on the issue, this means that any changes the user had made get wiped out. That's probably ok in a case like this, because the user chose to go back to the chat rather than use what had been proposed. On the other hand, if we chose to go with a solution like the Cline extension's, then the file would have been modified in place, and any changes the user made to the suggestions would be picked up by the agent in its second go around. On the other hand, it wouldn't see any MCP's to change to CMP, so it would fail on that count in this case. Lots of tradeoffs when trying to work intelligently with multiple potential versions of the same text 🤷 .

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/browser/change-set-file-element.ts
+++ b/packages/ai-chat/src/browser/change-set-file-element.ts
@@ -19,6 +19,7 @@ import { inject, injectable, postConstruct } from '@theia/core/shared/inversify'
 import { ChangeSetElement, ChangeSetImpl } from '../common';
 import { createChangeSetFileUri } from './change-set-file-resource';
 import { ChangeSetFileService } from './change-set-file-service';
+import { MonacoWorkspace } from '@theia/monaco/lib/browser/monaco-workspace';
 
 export const ChangeSetFileElementFactory = Symbol('ChangeSetFileElementFactory');
 export type ChangeSetFileElementFactory = (elementProps: ChangeSetElementArgs) => ChangeSetFileElement;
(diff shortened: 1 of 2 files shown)


JonasHelming:
@eneufeld Could you review this, please?

colin-grant-work:
Converting this to a draft for now, because I think I can do better as part of a [larger reconfiguration](https://github.com/eclipse-theia/theia/issues/14768#issuecomment-2652089017) of the AI diff editor UX.

colin-grant-work:
Superseded by https://github.com/eclipse-theia/theia/pull/14910

PR #14906, opened 2025-02-11
Fix tool calling string in messages

### LINKED ISSUE #14905: [Theia AI] Coder switched to attempt tool calls in text like this "Tool Call:"

JonasHelming:
fixed #14905

#### What it does

- Removes the string "Tool Call:..." from AI messages that were introduced by the ToolCallResponsContent.asString
- Make the Creation of ToolCallResponsContent adaptable by:
  - Moving it into an overridable method, so you can adapt it per agents
  - Retrieving the default in ChatAgent via DI, so you can globally override it
- Add a 'asDisplayString' (currently not yet used) to ToolCallResponsContent

#### How to test

Try the Coder agent in a longer conversation and see that it does not switch to attempting calling tools by outputting text like:
"Tool Call:"

see that copy still works in the chat and that the function call is mentioned in the history view.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-model.ts
+++ b/packages/ai-chat/src/common/chat-model.ts
@@ -964,14 +964,14 @@ class ChatResponseImpl implements ChatResponse {
     }
 
     protected _updateResponseRepresentation(): void {
-        this._responseRepresentation = this.responseRepresentationsToString(this._content, false);
-        this._responseRepresentationForDisplay = this.responseRepresentationsToString(this.content, true);
+        this._responseRepresentation = this.responseRepresentationsToString(this._content, 'asString');
+        this._responseRepresentationForDisplay = this.responseRepresentationsToString(this.content, 'asDisplayString');
     }
 
-    protected responseRepresentationsToString(content: ChatResponseContent[], forDisplay: boolean): string {
+    protected responseRepresentationsToString(content: ChatResponseContent[], collect: 'asString' | 'asDisplayString'): string {
         return content
             .map(responseContent => {
-                if (forDisplay) {
+                if (collect === 'asDisplayString') {
                     if (ChatResponseContent.hasDisplayString(responseContent)) {
                         return responseContent.asDisplayString();
                     }
--- a/packages/ai-chat/src/common/chat-model.ts
+++ b/packages/ai-chat/src/common/chat-model.ts
@@ -171,6 +171,7 @@ export interface ChatResponseContent {
      * representation of the response.
      */
     asString?(): string | undefined;
+    asDisplayString?(): string | undefined;
     merge?(nextChatResponseContent: ChatResponseContent): boolean;
 }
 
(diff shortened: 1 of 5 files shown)


JonasHelming:
@sdirix You suggested to add a 'asDisplayText' instead to not loose the copy feature in the chat. I have added the function. Do you have a hint on which caller to adapt to make this used for copy paste?

sdirix:
They are located here: https://github.com/eclipse-theia/theia/blob/3aed840f718d9a35e70c17f62c7cc154099f07c6/packages/ai-chat-ui/src/browser/chat-view-contribution.ts#L47-L93

packages/ai-chat/src/common/chat-model.ts
```diff
@@ -948,8 +964,18 @@ class ChatResponseImpl implements ChatResponse {
     }
 
     protected _updateResponseRepresentation(): void {
-        this._responseRepresentation = this._content
+        this._responseRepresentation = this.responseRepresentationsToString(this._content, false);
+        this._responseRepresentationForDisplay = this.responseRepresentationsToString(this.content, true);
+    }
+
+    protected responseRepresentationsToString(content: ChatResponseContent[], forDisplay: boolean): string {
```
sdirix, line 971: Nitpick: boolean parameters are always hard to parse and understand. Here it's easier as the caller and definition are side by side, even then one needs to mentally parse what happens in each call.

This could be refactored using a string, e.g.
```
  this._responseRepresentation = this.responseRepresentationsToString(this._content, 'asString');
  this._responseRepresentationForDisplay = this.responseRepresentationsToString(this.content, 'asDisplayString');

protected responseRepresentationsToString(content: ChatResponseContent[], collect: 'asString' | 'asDisplayString' {
```

packages/ai-chat/src/common/chat-model.ts
```diff
@@ -954,13 +988,17 @@ class ChatResponseImpl implements ChatResponse {
                 );
                 return undefined;
             })
-            .filter(text => text !== undefined)
+            .filter(text => (text !== undefined || text === ''))
```
sdirix, line 991: This change does nothing as `''` is already `!== undefined`, you probably wanted to do this to filter empty string?

```suggestion
            .filter(text => (text !== undefined && text !== ''))
```

At the moment we have a lot of whitespace where the tool calls were in the past:

```
 "query": "I'll help you check the contents of your workspace using the getWorkspaceFileList function for the root directory.
                
                
                
                
                
                Here are the files and directories in your workspace root:
                
                Files:
                - CHANGELOG.md
                ...
```

With the suggested change this changes to:


```
      "query": "I'll help you check the contents of your workspace using the getWorkspaceFileList function for the root directory.
                
                
                
                Here are the main files and directories in your workspace root:
```

Which is still a lot but at least it's less

sdirix (commented):
Works for me, however I found a code issue

sdirix (approved):

PR #14909, opened 2025-02-12
Empty hidden editors

### LINKED ISSUE #14880: Symbols in Monaco stop to be recognized and cannot be navigated to
### Steps to Reproduce:

1. Code in Theia and navigate to the definition of a symbol using ctrl+click (ideally defined outside the current file)
2. use the back navigation to go back to the file you came from
3. try to navigate to the symbol again

Unfortunately this is not 100% guaranteed to lead to the issue. You have sometime to navigate to multiple different symbols and go back and forth. After some time symbols cannot be navigated to anymore.

### Additional Information

- Operating System: Ubuntu 22.04.5
- Theia Version: 1.58.100

tsmaeder:
#### What it does
Monaco editors don't work correctly when they are hidden by setting `display: none`. This PR unsets the editor model from the editor when it is hidden and restores it when the editor is revealed again.

Fixes #14880

#### How to test
Test the scenario from the related issue. Have an eye out for an missing state in the editor or focus issues. This needs to be tested on Linux to make sure the issue is fixed as it did not happen on Windows.

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/widgets/widget.ts
+++ b/packages/core/src/browser/widgets/widget.ts
@@ -225,14 +225,18 @@ export class BaseWidget extends Widget implements PreviewableWidget {
     override setFlag(flag: Widget.Flag): void {
         super.setFlag(flag);
         if (flag === Widget.Flag.IsVisible) {
-            this.onDidChangeVisibilityEmitter.fire(this.isVisible);
+            this.handleVisiblityChanged(this.isVisible)
         }
+    }
+
+    protected handleVisiblityChanged(isNowVisible: boolean): void {
+        this.onDidChangeVisibilityEmitter.fire(isNowVisible);
     }
 
     override clearFlag(flag: Widget.Flag): void {
         super.clearFlag(flag);
         if (flag === Widget.Flag.IsVisible) {
-            this.onDidChangeVisibilityEmitter.fire(this.isVisible);
+            this.handleVisiblityChanged(this.isVisible);
         }
     }
 }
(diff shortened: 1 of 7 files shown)


tsmaeder:
@sdirix @msujew would be great if I could get some timely love for this one as it's the last one scheduled for the IDE patch release.

sdirix (changes requested):
This workaround fixes the main issue on my machine. However it seems to introduce other behavior changes.

See this video for changes between the latest Theia IDE and the Electron Example of this branch:

https://github.com/user-attachments/assets/e1d1f014-d5db-4cab-95ef-1451029ce9f9

The current Theia IDE behavior is already not that great:
 - The editors are opened on the correct line on first click, however the top overlay swallows the selected text. Still the editors are "almost" correct
 - On second click the correct line is shown

On this branch:
- More often than not, the editor just opens on the last line of the file which is bad behavior
- The second click then works fine

As this breaks just clicking on file results for me in the IDE, I would not recommend merging in this state

sdirix (approved):
Edit: I tested the latest `master` and there the same last-line issue appears. So it's not a regression of this branch, but something else on `master`.

As the main issue is fixed, I'll approve

Edit2: Maybe it's a me issue :shrug:, I went back several Theia versions on `master` and still encounter this issue when running the example apps.

msujew (approved):
I can confirm that this works as expected and does not introduce any regressions (I specifically checked the notebook integration as well, which simply behaves as usual).

PR #14910, opened 2025-02-12
Improve AI diff UX

colin-grant-work:
- Updates editors when new suggestions proposed
 - Separates accept and saving actions to avoid unnecessary prompts
 - Avoids use of untitled editors to avoid prompts for them

#### What it does

Fixes parts of #14768. In particular:
 - Eliminates autosaving for AI chat generated editors. Here, I chose to pipe a new `autosavable` field from the `Resource` interface to the `Saveable` interface and then not perform autosaves if the `Saveable` says it doesn't want to be autosaved. @msujew, you've recently worked on the autosaving infrastructure, so I'd be grateful to get your thoughts. Another alternative I considered was to pipe the `SaveReason` through to the `Resource`'s save options so that some resources could just reject automatic saves, but that feels less clean than having the `Saveable` declare itself exempt from that system.
 - Switches to using query-based in-memory resources for the left side of chat diffs. That makes them read-only, which I think makes sense: the user's focus is (or should be) on the right side, where the AI has suggested things. This avoids some synching issues, but arguably introduces others. `@Reviewers`, please see what you think of the way the diff behaves when you save the right side. It will still show the diff relative to original state, without showing that the underlying left-side file has actually been updated. I can fix that by using mutable in-memory resources, but then the lifecycle gets a little trickier.
 - Separates the save cycle of the resource from the `apply` cycle of the ChangeSetFileElement. This allows us to avoid hanging dirty state or prompts to save when the user believes they _have_ saved / applied the changes.
 - Changes the names of methods on `ChangeSetElements` from `accept` and `discard` to `apply` and `revert` and removes the `'discarded'` state option. The practical reason for doing so was that we were inconsistent in our handling of 'reverted' changes: we let them be reapplied on the element level, but we didn't reactivate the global `Activate` button on the ChangeSet level. Now if the user reverts a change, it returns to `'pending'` state: I don't think we should stand in the way of the user having third thoughts. Semantically, `discard` also sounded more like `delete` than it behaved. I'm open to discussion about this point.
 - Speaking of `delete`, I added optional `dispose` to the `ChangeSetElement` interface to be called when the element is deleted. This lets me do cleanup like closing diff editors when a change element is deleted. I have some misgivings about the lifecycle still: we currently replace elements in the change set, but we don't want everything associated with a given proposal (e.g. diff editors)  to die when it gets replaced.
 - And finally (I think), I adapted my code from https://github.com/eclipse-theia/theia/pull/14901 to update an editor when a new suggestion is made for the same file in the same session.

#### How to test

1. Fairly thorough testing of all the interactions you've gotten used to re: LLM suggestions.
2. Turn on autosave and observe that your AI editors aren't affected.
3. Open a diff editor and close it in a dirty state -> you should be prompted to save. I thought about getting rid of this if the user hasn't actually interacted with the editor, since then the state is saved as part of the LLM interaction. Let me know your thoughts.
4. Open a diff editor and save it. The corresponding node should be marked as accepted in the UI.
5. Open a diff and accept it in the chat UI. The diff should close without a prompt re: saving.
6. Have the LLM generate a new file. No matter what you do, you should never be prompted for the path at which to save something.
7. Open a diff editor and then delete the corresponding node. The editor should close. It should prompt you whether you want to save. Same question as re: (3).
8. Open a diff editor, then ask the agent to do something else in the same file (see instructions for https://github.com/eclipse-theia/theia/pull/14901, e.g.). The editor should be updated with the new content.

#### Breaking changes

- [x] This PR introduces breaking changes and requires careful review.
- [ ]  If yes, the breaking changes section in the [changelog](https://github.com/eclipse-theia/theia/blob/master/CHANGELOG.md) has been updated before merge.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
@@ -426,8 +426,8 @@ const buildChangeSetUI = (changeSet: ChangeSet, labelProvider: LabelProvider, on
         name: element.name ?? labelProvider.getName(element.uri),
         additionalInfo: element.additionalInfo ?? labelProvider.getDetails(element.uri),
         openChange: element?.openChange?.bind(element),
-        accept: element.state !== 'applied' ? element?.accept?.bind(element) : undefined,
-        discard: element.state === 'applied' ? element?.discard?.bind(element) : undefined,
+        accept: element.state !== 'applied' ? element?.apply?.bind(element) : undefined,
+        discard: element.state === 'applied' ? element?.revert?.bind(element) : undefined,
         delete: () => onDeleteChangeSetElement(changeSet.getElements().indexOf(element))
     }))
 });
@@ -557,7 +557,7 @@ const ChatInputOptions: React.FunctionComponent<ChatInputOptionsProps> = ({ left
 );
 
 function acceptAllPendingElements(changeSet: ChangeSet): void {
-    acceptablePendingElements(changeSet).forEach(e => e.accept!());
+    acceptablePendingElements(changeSet).forEach(e => e.apply!());
 }
 
 function hasPendingElementsToAccept(changeSet: ChangeSet): boolean | undefined {
@@ -565,7 +565,7 @@ function hasPendingElementsToAccept(changeSet: ChangeSet): boolean | undefined {
 }
 
 function acceptablePendingElements(changeSet: ChangeSet): ChangeSetElement[] {
-    return changeSet.getElements().filter(e => e.accept && (e.state === undefined || e.state === 'pending'));
+    return changeSet.getElements().filter(e => e.apply && (e.state === undefined || e.state === 'pending'));
 }
 
 function getLatestRequest(chatModel: ChatModel): ChatRequestModel | undefined {
--- a/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
@@ -417,7 +417,7 @@ const noPropagation = (handler: () => void) => (e: React.MouseEvent) => {
 const buildChangeSetUI = (changeSet: ChangeSet, labelProvider: LabelProvider, onDeleteChangeSet: () => void, onDeleteChangeSetElement: (index: number) => void): ChangeSetUI => ({
     title: changeSet.title,
     disabled: !hasPendingElementsToAccept(changeSet),
-    acceptAllPendingElements: () => acceptAllPendingElements(changeSet),
+    applyAllPendingElements: () => applyAllPendingElements(changeSet),
     delete: () => onDeleteChangeSet(),
     elements: changeSet.getElements().map(element => ({
         open: element?.open?.bind(element),
@@ -426,8 +426,8 @@ const buildChangeSetUI = (changeSet: ChangeSet, labelProvider: LabelProvider, on
         name: element.name ?? labelProvider.getName(element.uri),
         additionalInfo: element.additionalInfo ?? labelProvider.getDetails(element.uri),
         openChange: element?.openChange?.bind(element),
-        accept: element.state !== 'applied' ? element?.apply?.bind(element) : undefined,
-        discard: element.state === 'applied' ? element?.revert?.bind(element) : undefined,
+        apply: element.state !== 'applied' ? element?.apply?.bind(element) : undefined,
+        revert: element.state === 'applied' || element.state === 'stale' ? element?.revert?.bind(element) : undefined,
         delete: () => onDeleteChangeSetElement(changeSet.getElements().indexOf(element))
     }))
 });
@@ -439,15 +439,15 @@ interface ChangeSetUIElement {
     additionalInfo: string;
     open?: () => void;
     openChange?: () => void;
-    accept?: () => void;
-    discard?: () => void;
+    apply?: () => void;
+    revert?: () => void;
     delete: () => void;
 }
 
 interface ChangeSetUI {
     title: string;
     disabled: boolean;
-    acceptAllPendingElements: () => void;
+    applyAllPendingElements: () => void;
     delete: () => void;
     elements: ChangeSetUIElement[];
 }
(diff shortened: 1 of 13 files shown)


JonasHelming:
Question: The PR description says the change log was updated (due to breaking changes), but i do not see the update in the PR?

colin-grant-work:
> Question: The PR description says the change log was updated (due to breaking changes), but i do not see the update in the PR?

Fair point. I'd say that some of the breakage is optional, and I'd prefer to update the changelog once I know which bits have survived review (or been added because of it). But I can write an entry now and then try to keep it up to date if anything changes, if you prefer.

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -107,31 +111,36 @@ export class ChangeSetFileElement implements ChangeSetElement {
...
+            this.readOnlyUri,
             this.changedUri
         );
     }
 
-    async accept(contents?: string): Promise<void> {
-        this.state = 'applied';
-        if (this.type === 'delete') {
+    async apply(contents?: string): Promise<void> {
+        if (await this.changeSetFileService.trySave(this.changedUri)) { /** Continue */ } else if (this.type === 'delete') {
```
planger, line 120: Optional nitpick: Just a matter of taste, but I'm not loving of the /** Continue */ but would prefer negating the condition and either duplicate the closeDiffsFor()-call or by nesting the if-statement in the if-statement of the negated condition.

planger (commented):
Awesome, this improves the current state already immensely!

### Asking to Save

> 3. Open a diff editor and close it in a dirty state -> you should be prompted to save. I thought about getting rid of this if the user hasn't actually interacted with the editor, since then the state is saved as part of the LLM interaction. Let me know your thoughts.

> 7. Open a diff editor and then delete the corresponding node. The editor should close. It should prompt you whether you want to save. Same question as re: (3).

I personally would prefer, if we don't ask the user whether they would like to save, if the right side of the suggestion-diff-editor hasn't been changed by the user. Feels unexpected. But its tricky and I can happily live with both.

### Accept vs Apply, etc.

I like the new naming much better than what we had before! Thanks!
Shouldn't we also propagate that to the UI too? The labels of the button and hovers still use the term accept. I think it'll be better to keep the naming in the code and UI consistent, and change "Accept" to "Apply" and "Undo" to "Revert" and "Delete" to "Discard"?

### Dirty State When Applied

If the state of a suggestion is "Applied" and we open the diff editor, shouldn't we avoid the dirty state in the diff editor? It is already applied and the dirty state suggests that it is not, in my view.

### The Left Side of the Suggestion Diff Editor

Currently the left side is the state of the time where the change set element has been created and remains in this state.

Shouldn't we keep the left side of the diff editor always up to date with what's currently in the file system or Monaco workspace? If I made changes in between, and hit save in the diff editor, I wouldn't even see that I'm overwriting my changes. If we would show the actual current state left, I'd at least see that as a diff.

On the other hand, it currently reflects what the LLM was using as its basis; also, if the user applied the suggestions, the diff editor now still shows the actual original suggestions. Makes sense too.

However, overall, it feels more logical to me, if the left side is always the current workspace state. This would also mean that if I applied the suggestions, there are no further AI suggestions to review -- so the diff editor becomes clean (no changes).

In conclusion, I lean towards syncing the left side with the current Monaco workspace state (or file system state if not open). To me the benefits outweigh the single advantage that I still can see what the LLM originally suggested. But to the user this is not such a relevant information, once they already reviewed and accepted the changes. And they can always revert in the change set UI.

planger:
Oh, one more observation:
It looks like with the [`@ChangeSet` agent](https://github.com/eclipse-theia/theia/blob/master/examples/api-samples/src/browser/chat/change-set-chat-agent-contribution.ts), the diff editor content isn't replaced (even if it is the same file), while it works with the `@Coder`. I think, this is because the `@ChangeSet` agent always creates a new change set, while the `ChangeSetFileResourceResolver` listens to the specific change set for updates.

We should probably change the `@ChangeSet` to ensure the example conforms to best practice.

colin-grant-work:
@planger, thanks for your review. I've pushed a new version that reorganizes things considerably, pushing more of the responsibility back to the element level, but implementing a resource that makes it feasible for a sequence of change elements to handle state reasonably, on the assumption that we don't have two file change elements for the same file active in the same session at the same time, which I take to be a reasonable assumption, apart from brief transitions.

NB: Most of the utilities I've implemented for the change resources are enhancements on the base `InMemoryResources` utilities. I reimplemented them in large part because as far as I could tell, disposing of the reference resources in the current `InMemoryResources` didn't actually do anything, so we leak memory. That seems buggy, but it's also been part of the `InMemoryResources` system for quite some time, and 'fixing' the bug might break others who depend on the current behavior.

## Asking to Save

I've set the system up so that the right side of the editors is not initially dirty. A manual save will still set the change to applied.

## Terminology

I've applied the new terms on the frontend as well. I've also added a `stale` state that is set if the file changes and its contents aren't equal to either the original or target state of the change. In that case, reverting or applying will have the normal effect, but the user will be prompted to confirm that they want to perform the action, and the diff will be opened automatically for them to consider. Maybe we wouldn't want to allow reversions in that state - it would just take the user back to the state they were in when they requested the suggestion from the LLM.

## The Left Side of the Suggestion Diff Editor

The left side now reflects the state of the file _on disk_ unsaved changes in an editor won't appear. This, together with the `stale` state, seems like a reasonable balance of presenting the file as it is, and the file as the LLM saw it when it made its suggestion.

## The `@ChangeSet` example agent should now behave correctly, since the content management has been pushed back down to the to the element level, and that agent does create file change elements.

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
```diff
@@ -460,10 +460,10 @@ const ChangeSetBox: React.FunctionComponent<{ changeSet: ChangeSetUI }> = ({ cha
                 <button
                     className='theia-button'
                     disabled={changeSet.disabled}
-                    title={nls.localize('theia/ai/chat-ui/acceptAll', 'Accept all pending changes')}
-                    onClick={() => changeSet.acceptAllPendingElements()}
+                    title={nls.localize('theia/ai/chat-ui/acceptAll', 'Apply all pending suggestions')}
+                    onClick={() => changeSet.applyAllPendingElements()}
                 >
-                    Accept
+                    {nls.localizeByDefault('Copy All')}
```
planger, line 466:
```suggestion
                    {nls.localizeByDefault('Apply All')}
```

planger (approved):
Thanks for the fast fix! Works great now and code looks good.
I just think we should rename "Copy All" to "Apply All" in the UI (see inline suggestion).

Thanks again for this excellent UX improvement!

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
colin-grant-work: Seems to have been changed by a `lint --fix` invocation - switched to the more verbose non-default form.

PR #14917, opened 2025-02-13
Deprioritize file resource resolver to avoid resolution delays

### LINKED ISSUE #14916: [Theia AI] Diff Opening Slow
The first time opening the diff editor from a changeset can take a long time without any visual feedback. Subsequent openings are quicker.

It turns out that this exposes a conflict of expectations. On the one hand, in some places, we expect that every scheme we open should be served by full `FileSystemProvider` implementation. In particular, the delay in this case is being caused by this `waitForEvent` call:

https://github.com/eclipse-theia/theia/blob/9a20727743172a4efa46334f4136a1242b160848/packages/plugin-ext/src/hosted/browser/hosted-plugin.ts#L456-L463

Because the `FileService` is waiting on that before trying to resolve the content of the ai chat scheme URI:

https://github.com/eclipse-theia/theia/blob/9a20727743172a4efa46334f4136a1242b160848/packages/filesystem/src/browser/file-service.ts#L404-L417

But internally, for a number of schemes we just provide resource resolvers rather than full file system providers, so we actually run into rejections for a number of internal schemes.

Although the problem as described has been around for a while - I think since the implementation of the VSCode file system API - the reason we're seeing it now even though the same wait fails for e.g. `untitled` scheme URI's is that it only occurs if a resource provider for a given scheme occurs _after_ the `FileResourceResolver`, so that the resource resolution system first tries the `FileResourceResolver` and then hits the 3-second delay before that resolver throws an error indicating that it won't handle the URI after all. Most of our internal resource resolvers are registered in `core`, so they precede the resolver from `filesystem`, but the AI resolver is registered later.

There are a few ways I can see getting around this:
 - Change the mechanism by which the resource resolution system decides whether to try a given resolver. The resource resolution system currently relies on resource resolvers throwing an error, and most resolvers handle only one scheme and immediately throw an error if the URI doesn't have that scheme. The file resolver is hooked into the plugin system and dynamic registrations, so it takes time to figure out whether it can handle a given URI. Instead, we could have a system where resolvers can declare whether they will, won't, or might (with priority) handle a given scheme. This is somewhat breaking, but if we filter down to resolvers that don't have the relevant method + resolvers that declare they might handle a scheme and then follow the current procedure, downstream shouldn't be affected much.
- Have the FileSystem check for the existence of a resource resolver for a given scheme before it goes to its dynamic resolution. This would mean that we prioritize internal resource resolution over anything a plugin might contribute, but for the most part, that's true already anyway. (This maybe doesn't really work without _also_ implementing the changes above - right now resource resolvers don't actually declare what schemes they'll handle, and that's why we have this problem in the first place.)
- A slightly heavier version of that would be to wrap internal resource resolver registrations in lightweight file system provider implementations automatically so that the file system just finds that they actually satisfy its expectations.
- Demote the file resolver, since we know that it has the property of holding the system up; if we want to do it without hardcoding, we could add a priority declaration on resource resolvers and just make sure the file resolver has a very / infinitely low priority.

colin-grant-work:
#### What it does

Fixes #14916 by deprioritizing the `FileResourceResolver`. See discussion on the issue - this seems like the least intrusive solution.

#### How to test

1. Reload the window (if you previously had an AI chat diff open, the application will attempt to resolve its URI at startup to restore the layout; it will fail and the editor won't be restored, but the attempt will mean you wouldn't have seen the slow diff load)
2. Open a diff from an AI chat.
3. Observe that the diff comes up quickly without a ca. 3 delay.
4. Open regular files / editors of other kinds.
5. Observe that they open without undue delay, as well.

#### Follow-ups

Although it's used in a few places, the system of trying various options to find the one that doesn't throw an error is a bit fragile, especially in an asynch context.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/resource.ts
+++ b/packages/core/src/common/resource.ts
@@ -14,7 +14,7 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { injectable, inject, named } from 'inversify';
+import { injectable, inject, named, postConstruct } from 'inversify';
 import { TextDocumentContentChangeEvent } from 'vscode-languageserver-protocol';
 import URI from '../common/uri';
 import { ContributionProvider } from './contribution-provider';
@@ -179,6 +179,11 @@ export namespace ResourceError {
 export const ResourceResolver = Symbol('ResourceResolver');
 export interface ResourceResolver {
     /**
+     * Resolvers will be ordered by descending priority.
+     * Default: 0
+     */
+    priority?: number;
+    /**
      * Reject if a resource cannot be provided.
      */
     resolve(uri: URI): MaybePromise<Resource>;
@@ -194,6 +199,11 @@ export class DefaultResourceProvider {
         @inject(ContributionProvider) @named(ResourceResolver)
         protected readonly resolversProvider: ContributionProvider<ResourceResolver>
     ) { }
+
+    @postConstruct()
+    init(): void {
+        this.resolversProvider.getContributions().sort((a, b) => (b.priority ?? 0) - (a.priority ?? 0));
+    }
 
     /**
      * Reject if a resource cannot be provided.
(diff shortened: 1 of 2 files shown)


planger (approved):
What a neat change and elegant fix! At least I don't see any negative side effects, and the delay for the file-change-set-element-diff-editor is gone.

Thank you!

PR #14925, opened 2025-02-14
Fix recent workspace tracking for devcontainer workspaces

sgraband:
#### What it does
Ensure devcontainer workspaces are correctly added to recentWorkspaces. 
Previously, incorrect URI handling caused invalid workspace entries. 
Now, the workspace path is passed instead of the full URI. 
Also the setMostRecentlyUsedWorkspace is now called with the correct workspace. 
Renamed the ContainerConnectionOptions.workspaceUri to workspacePath for consistency.

#### How to test

1. Open a devcontainer from a workspace
2. Notice that is has been added to the recentWorkspaces
3. Open some other workspaces
4. Open the same devcontainer again
5. Observe that it is being opened and now at the top of the recentWorkspaces again

(Same steps for reproducing the issue on current master, except there the entry is not moved to the top and another weird entry is added)

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/dev-container/src/electron-browser/container-connection-contribution.ts
+++ b/packages/dev-container/src/electron-browser/container-connection-contribution.ts
@@ -82,7 +82,7 @@ export class ContainerConnectionContribution extends AbstractRemoteRegistryContr
             throw new Error(`Devcontainer file at ${filePath} not found in workspace`);
         }
 
-        return this.doOpenInContainer(devcontainerFile, uri.toString());
+        return this.doOpenInContainer(devcontainerFile, uri.path.toString());
     }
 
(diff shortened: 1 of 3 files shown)


jonah-iden (approved):
Didn't even know this was broken. Thanks alot for the fix. Seems to work fine. 

There is only still the issue for me that opening a recent workspace creates a new container instead of reusing the existing one. I thought this was working once but i think that has nothing to do with these changes

sgraband:
Thank you for the fast review! Do we have a ticket for the issue you have metioned?

jonah-iden:
i don't think so

sgraband:
Can you reproduce this everytime? For me this sometimes works and sometimes does not.

PR #14930, opened 2025-02-14
Add Scan OSS as action on Change Set level

### LINKED ISSUE #14915: [Theia AI] Run ScanOSS on changeset
It should run only on new code (even in case the full file was rewritten). The result should be shown in the changeset on the file. It should only be active if ScanOSS pref is set to automatic or manual.

colin-grant-work:
Also adds a service for contributing arbitrary actions

#### What it does

Fixes #14915 by implementing a contribution to run ScanOSS on change set level.

#### How to test

1. Get an AI agent to suggest some changes so you have a changeset to work with.
2. If your `ai-features.SCANOSS.mode` preference is set to `off` (the default), you should see nothing related to this PR.
3. If that preference is set to `manual`, you should see a button to run ScanOSS on the diff; the button should also appear if you change the preference from `off` -> `manual` with the changeset available.
> Things could get odd here. We've implemented this widget so that change events are handled pretty deep in the React. If those React components aren't mounted at the time of the change event, I'm not sure that they'd be handled at all; the same goes for the current change handlers.
4. If you click that button, or if you change the preference to `automatic`, a scan should be run.
5. If the scan passes (you can force this by generating a change that only deletes content - as we only scan new material, this is automatically counted as safe), you should see a 'no matches' message, and additional clicks on the 'button' will do nothing (should have the default pointer, to indicate this).
6. If the scan produces matches (you can force this by having the agent replace the content of one file with the content of another - the new content will match Theia, which is EPL licensed), you should see a message to that effect, and clicking the button should open a report.
7. If the scan errors, you should see a message to that effect, and clicking the button will rerun the scan.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


packages/ai-chat-ui/src/browser/chat-input-widget.tsx
```diff
@@ -524,32 +567,26 @@ interface ChangeSetUIElement {
...
-                    title={nls.localize('theia/ai/chat-ui/acceptAll', 'Accept all pending changes')}
-                    onClick={() => changeSet.acceptAllPendingElements()}
-                >
-                    Accept
-                </button>
-                <span className='codicon codicon-close action' title={nls.localize('theia/ai/chat-ui/deleteChangeSet', 'Delete Change Set')} onClick={() => changeSet.delete()} />
-            </div>
-        </div>
+                {actions.map(action => <div key={action.id} className='theia-changeSet-Action'>{action.render(changeSet)}</div>)}
+                <span className='codicon codicon-close action' title='Delete Change Set' onClick={() => deleteChangeSet()} />
```
sdirix, line 584: The title to be translated

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
```diff
@@ -584,9 +621,9 @@ const ChangeSetBox: React.FunctionComponent<{ changeSet: ChangeSetUI }> = ({ cha
                     </li>
                 ))}
             </ul>
-        </div>
-    </div>
-);
+        </div >
+    </div >
```
sdirix, line 625: These are weird formatting changes. I'm surprised the linter accepts them

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
```diff
@@ -524,32 +567,26 @@ interface ChangeSetUIElement {
...
-                >
-                    Accept
-                </button>
-                <span className='codicon codicon-close action' title={nls.localize('theia/ai/chat-ui/deleteChangeSet', 'Delete Change Set')} onClick={() => changeSet.delete()} />
-            </div>
-        </div>
+                {actions.map(action => <div key={action.id} className='theia-changeSet-Action'>{action.render(changeSet)}</div>)}
+                <span className='codicon codicon-close action' title='Delete Change Set' onClick={() => deleteChangeSet()} />
+            </div >
+        </div >
```
sdirix, line 586: Same formatting issues here

packages/ai-scanoss/src/browser/ai-scanoss-code-scan-action.tsx
```diff
@@ -119,15 +118,12 @@ const ScanOSSIntegration = React.memo((props: {
...
+        if (scanOSSResult === 'pending') {
             return;
-        }
-        if (!scanResult || scanResult.type === 'error') {
-            scanResult = await scanCode();
-        }
-        if (scanResult && scanResult.type === 'match') {
-            const dialog = new ScanOSSDialog(scanResult);
+        } else if (!scanOSSResult || scanOSSResult.type === 'error') {
+            scanCode();
```
sdirix, line 124: This is a behavioral change. Now I need to click twice to see the results if they did not exist before. Please restore the previous behavior.

packages/ai-scanoss/src/browser/ai-scanoss-code-scan-action.tsx
```diff
@@ -144,37 +140,35 @@ const ScanOSSIntegration = React.memo((props: {
...
 });
 
 export class ScanOSSDialog extends ReactDialog<void> {
     protected readonly okButton: HTMLButtonElement;
 
-    constructor(protected result: ScanOSSResultMatch) {
+    constructor(protected results: ScanOSSResultMatch[]) {
         super({
-            title: 'SCANOSS Results',
+            title: 'ScanOSS Results',
```
sdirix, line 171: The translations in this file were seemingly missed before. We should add those translations.

sdirix (changes requested):
Thanks for the changes! It works for me!

Some improvement suggestions

## General issues

 - The CI builds fail because not all auto-generated build adjustments are part of the commit(s). Please add those
 - All user facing strings need to be translated as we merged https://github.com/eclipse-theia/theia/pull/14857

## Styling

 - If a match was found, ideally the whole button should change color so it's easily noticable, and/or we should use the same "warning" highlight we do on the code snippet button for consistency: At the moment we show a checkmark if all is fine. In the warning case we should show the warning sign instead.
 - Currently we apply the same half-transparent background color as for the code snippet button in the chat view. This looks off for two reasons:
   - The button is blue, so the yellow half-transparent highlight looks off (the code snippet background is usually white or dark)
   - The button is a square, so it looks weird (the code snippet button is round)
 - We need to check all three highlight colors (red, green, yellow)

image

image

## Behavior with automatic scanning

- It seems that changes to the change set are not leading to a rescan. To reproduce: 
  - Ask `@Coder log the outputs in packages/preferences/src/browser/views/components/preference-string-input.ts`
  - Then in the same thread: `@Coder Move packages/preferences/src/browser/views/components/preference-select-input.ts one directoy up`.
  - Observe that no rescan is scheduled and the previous "No matches found" continues to be shown
- Even worse: It seems this is even true within the same change set. Try `@Coder modify the fallback value in packages/preferences/src/browser/views/components/preference-string-input.ts to 'foobar'. Also move packages/preferences/src/browser/views/components/preference-select-input.ts one directory up` and it will end up with "No matches found" because of the order it produces the changes.

## General behavior

- If there are multiple matches, the UI lists the results but does not really show which match corresponds to which change set entry
- In the error case, we show much more information in the code snippet action 
image
- The regular scanning button for code snippets now needs to be clicked twice in manual mode

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
colin-grant-work: This appears to be acceptable according to [the spec](https://dev.w3.org/html5/spec-LC/syntax.html#end-tags) for HTML tags, and the rest of the code base has a mix, admittedly with a majority without space. I don't _think_ I've configured anything specific to this case, so this is likely to be the default behavior of the TS formatter on JSX?

packages/ai-scanoss/src/browser/ai-scanoss-code-scan-action.tsx
colin-grant-work: Hm... this was deliberate. In general, I think opening a modal as a result of an asynchronous action is not a great experience for the user: if the data isn't available instantly, then the user isn't in control of when the dialog actually appears, meaning they either have to sit and wait, or it can interrupt them if they move on to another task.

sdirix: ScanOSS is very fast (it's basically a hash table lookup), so in the majority of cases the result is there immediately and the user needs to perform a "double click" to open the dialog. The only case this "fixes" is when the user is on Electron and has poor internet reception.

Suggestion: Let's measure the elapsed time until the response is received and if it's below some threshold (e.g. 500ms) then let's open the modal immediately without requiring a further click. This should cover the majority of typical usage scenarios.

Only if it takes longer we don't open the dialog automatically. The user is informed about this as the button also does not switch to a result state and therefore indicates that the result is not there yet.

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
sdirix: Overall it's a nitpick as it does not really matter. So feel free to keep it.

It's just strange to me that we change these two `div` elements here specifically (they are not even new ones), and for example not the `ul` or `li` elements. As this is very arbitrary and not even consistent within the same JSX block I would like to suggest removing this formatting change.

colin-grant-work:
@sdirix, thanks for the review. I've pushed a new version in which:

 - hopefully committed all of the automatic changes required.
 - added NLS calls for hopefully all user-facing strings that didn't previously have them
 - styled the status icons in the button for matched and clean states to be warning and success colors, respectively.
 - added listeners to update the state of the button when the changeset changes, and added some cache logic so that we don't go to ScanOSS if a particular file hasn't changed.
 - added an additional heading level in the report to indicate which file a given result came from.

sdirix (changes requested):
Thanks for the work!

## Cache failing

I found edge cases in which the new cache fails:

1. Use manual scanning
2. Execute `@Coder log the outputs in packages/preferences/src/browser/views/components/preference-string-input.ts`
3. Run Scan action -> Should not find a match
4. In the same thread execute: `@Coder Move packages/preferences/src/browser/views/components/preference-select-input.ts one directoy up`
5. Run Scan action -> Should find a match
6. Delete the create and delete changeset entries, leaving the original output logging change
7. Rerun the Scan action
8. It still reports a match which should not be the case (as it's the same scenario as step 3)

Another example

1. Use manual scanning
2. Execute `@Coder place this function     protected createInteractable(parent: HTMLElement): void {
        this.updateSelectOptions();
        const interactable = document.createElement('div');
        const selectComponent = React.createElement(SelectComponent, {
            options: this.selectOptions,
            defaultValue: this.getDataValue(),
            onChange: (_, index) => this.handleUserInteraction(index),
            ref: this.selectComponent
        });
        this.interactable = interactable;
        const root = createRoot(interactable);
        root.render(selectComponent);
        parent.appendChild(interactable);
    } into packages/preferences/src/browser/views/components/preference-string-input.ts`
3. Run Scan action -> Should find a match
4. Apply all changes
5. Execute `@Coder Replace the copyright year in packages/preferences/src/browser/views/components/preference-string-input.ts with 2025`
6. Run Scan action -> Should not find a match but it does

## Hover

The "No Matches" still reacts on hover which makes it seem clickable

HoverNoMatchesFound

## Error case

Thanks for adding the notification in an error case. However in case there should be an error, for example the service is offline, then we will get a separate error notification with *every* executed automatic scan. That's very annoying and will lead the user to disable the feature.

I think in case of an automated scan, we should not show the error notification except for when the user did actually run the action manually afterwards by clicking on it.

## Translation

~[This translation is missing](https://github.com/eclipse-theia/theia/pull/14930#discussion_r1960096068).~ Added in https://github.com/eclipse-theia/theia/pull/14930/commits/3efafdb2975fb3e9fec9455aa715a20f8ad1dffd

sdirix (approved):
Thanks for the amazing work!

PR #14943, opened 2025-02-17
Improve error handling on optional dependencies

msujew:
#### What it does

Currently, whenever adopters build a Theia app that exclusively uses the Theia browser build, the `@theia/application-package` package prints an error about `@theia/electron` missing due to it being declared as an optional peer dependency. However, the current implementation does not take this optionality into account and prints an error even though it is not strictly required.

This change takes the `peerDependenciesMeta` property into account to print a better message for unresolved peer dependencies.

Note that this error message has confused **a lot of adopters** in the past.

#### How to test

Cannot be properly tested in this repo. In theory you will need to build a Theia browser app and then simply build the app. It should not log an error about being unable to find `@theia/electron`.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-package/src/extension-package-collector.ts
+++ b/dev-packages/application-package/src/extension-package-collector.ts
@@ -21,7 +21,7 @@ import { ExtensionPackage, ExtensionPackageOptions, RawExtensionPackage } from '
 export class ExtensionPackageCollector {
 
     protected readonly sorted: ExtensionPackage[] = [];
-    protected readonly visited = new Map<string, boolean>();
+    protected readonly visited = new Set<string>();
 
     constructor(
         protected readonly extensionPackageFactory: (raw: PublishedNodePackage, options?: ExtensionPackageOptions) => ExtensionPackage,
@@ -40,7 +40,8 @@ export class ExtensionPackageCollector {
             ...Object.entries(pck.dependencies ?? {}),
             ...Object.entries(pck.peerDependencies ?? {})
         ]) {
-            this.collectPackage(packagePath, dependency, versionRange!);
+            const optional = pck.peerDependenciesMeta?.[dependency]?.optional || false;
+            this.collectPackage(packagePath, dependency, versionRange!, optional);
         }
     }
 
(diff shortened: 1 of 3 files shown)


dev-packages/application-package/src/extension-package-collector.ts
```diff
@@ -52,17 +53,21 @@ export class ExtensionPackageCollector {
...
+        this.visited.add(name);
 
         let packagePath: string | undefined;
         try {
             packagePath = this.resolveModule(parentPackagePath, name);
-        } catch {
-            console.debug(`Failed to resolve module: ${name}`);
+        } catch (err) {
+            if (optional) {
+                console.log(`Could not resolve optional peer dependency '${name}'. Skipping...`);
```
tsmaeder, line 67: Do we not still want the error message? Maybe the error is not that the package is missing (expected), but something unrelated?

dev-packages/application-package/src/npm-registry.ts
```diff
@@ -49,6 +53,7 @@ export interface NodePackage {
     keywords?: string[];
     dependencies?: Dependencies;
     peerDependencies?: Dependencies;
+    peerDependenciesMeta?: PeerDependenciesMeta;
```
tsmaeder, line 56: I guess this just describes what we get from npm, right?

dev-packages/application-package/src/extension-package-collector.ts
msujew: I'm not sure what the unrelated errors might be, since the only thing we actually do in the `try` block is to resolve the dependency to its `package.json` file. But I guess checking for `ENOENT` is an option.

dev-packages/application-package/src/npm-registry.ts
msujew: Not npm, but the entry in the `package.json`. But yes, it's just part of the `package.json` structure.

tsmaeder (approved):

JonasHelming:
@msujew  Can this be merged?

Issue #14842, opened 2025-02-05
Error: There are no document for untitled:/Untitled-1aichatviewlanguage


kittaakos:
### Bug Description:

When I start Theia from the sources, I see this error twice in the console:

```
root ERROR Error: There are no document for untitled:/Untitled-1aichatviewlanguage
    at CodeActionAdapter.provideCodeAction (/Users/kittaakos/dev/git/theia/examples/browser/lib/backend/packages_core_shared_reflect-metadata_index_js-packages_plugin-ext_lib_hosted_node_plugin-hos-48197a.js:8188:35)
    at /Users/kittaakos/dev/git/theia/examples/browser/lib/backend/packages_core_shared_reflect-metadata_index_js-packages_plugin-ext_lib_hosted_node_plugin-hos-48197a.js:7705:93
    at LanguagesExtImpl.withAdapter (/Users/kittaakos/dev/git/theia/examples/browser/lib/backend/packages_core_shared_reflect-metadata_index_js-packages_plugin-ext_lib_hosted_node_plugin-hos-48197a.js:7448:20)
    at LanguagesExtImpl.$provideCodeActions (/Users/kittaakos/dev/git/theia/examples/browser/lib/backend/packages_core_shared_reflect-metadata_index_js-packages_plugin-ext_lib_hosted_node_plugin-hos-48197a.js:7705:21)
    at RPCProtocolImpl.handleRequest (/Users/kittaakos/dev/git/theia/examples/browser/lib/backend/packages_plugin-ext_lib_common_plugin-api-rpc_js.js:2081:31)
    at RpcProtocol.requestHandler (/Users/kittaakos/dev/git/theia/examples/browser/lib/backend/packages_plugin-ext_lib_common_plugin-api-rpc_js.js:2063:96)
    at RpcProtocol.handleRequest (/Users/kittaakos/dev/git/theia/examples/browser/lib/backend/packages_core_lib_common_index_js-node_modules_vscode-languageserver-types_lib_umd_sync_recursive.js:3625:39)
    at RpcProtocol.handleMessage (/Users/kittaakos/dev/git/theia/examples/browser/lib/backend/packages_core_lib_common_index_js-node_modules_vscode-languageserver-types_lib_umd_sync_recursive.js:3512:26)
    at /Users/kittaakos/dev/git/theia/examples/browser/lib/backend/packages_core_lib_common_index_js-node_modules_vscode-languageserver-types_lib_umd_sync_recursive.js:3498:66
[...]

Issue #7671, opened 2020-04-27
There is no document for breakpointinput:vs.editor.ICodeEditor%3A1


akosyakov:
It seems we provide documents for inline editors like breakpoint expressions editor to the plugin host process. Not sure we should do it.

```
Error: There is no document for breakpointinput:vs.editor.ICodeEditor%3A1
    at CodeLensAdapter.provideCodeLenses (/theia/node_modules/@theia/plugin-ext/lib/plugin/languages/lens.js:70)
    at /theia/node_modules/@theia/plugin-ext/lib/plugin/languages.js:372
    at LanguagesExtImpl.<anonymous> (/theia/node_modules/@theia/plugin-ext/lib/plugin/languages.js:153)
    at step (/theia/node_modules/@theia/plugin-ext/lib/plugin/languages.js:47)
    at Object.next (/theia/node_modules/@theia/plugin-ext/lib/plugin/languages.js:28)
    at /theia/node_modules/@theia/plugin-ext/lib/plugin/languages.js:22
    at new Promise (<anonymous>)
    at __awaiter (/theia/node_modules/@theia/plugin-ext/lib/plugin/languages.js:18)
    at LanguagesExtImpl.withAdapter (/theia/node_modules/@theia/plugin-ext/lib/plugin/languages.js:145)
    at LanguagesExtImpl.$provideCodeLenses (/theia/node_modules/@theia/plugin-ext/lib/plugin/languages.js:372)
    at editor.main.js:30
```

jili544:
hello   I also encountered the same problem, how can I solve it?

vince-fugnitto:
> hello I also encountered the same problem, how can I solve it?

@jili544 thank you for the response, do you mind providing details on how you reproduced the issue (steps)?

tsmaeder:
To reproduce, open any file (for example, a package.json) while the debug view is open.

colin-grant-work:
> To reproduce, open any file (for example, a package.json) while the debug view is open.

Or even when the debug view isn't open. I seem to get it opening any source code file at all. Closing the debug view entirely doesn't stop the error.

FinnChen:
To reproduce this problem, follow the steps below.
step 1: install the `GitLens` VS Code extension
step 2: open any file

```
Error: There is no document for breakpointinput:vs.editor.ICodeEditor%3A26
[...]

Issue #14924, opened 2025-02-14
Chat Input is unusable


sdirix:
### Bug Description:

The chat input grows indefinitely and is unusable

![Image](https://github.com/user-attachments/assets/d46bde6e-3538-4f57-b85f-6d5b0c7ae579)

### Steps to Reproduce:

1. Open the Chat View

### Additional Information

Seems to be caused by https://github.com/eclipse-theia/theia/pull/14909

- Operating System: Ubuntu 22.04
- Theia Version: latest master 9944d0e477870308362e0a0e2bd074cc80651afb

tsmaeder:
To me, it seems that the `ChatInput` component adds a resize listener to the editor and in that listener, it sets the size of the editor again. This leads to an infinite loop if the size we're setting does not converge on a constant value.

tsmaeder:
Although: I can't get a breakpoint in the updating code to ever hit. Maybe something else is going on here.

tsmaeder:
The problem does not exist on Firefox.

tsmaeder:
Correction: after rebuilding with `npm install`+ `npm run build` of current master, I cannot reproduce the problem. Very strange.

sdirix:
Might be useful hint: When reverting 72ca8ed8559d6309e374d74f42f631e0f57e915c, then `updateEditorHeight` is hit. So for some reason we seem to never enter the code with 72ca8ed8559d6309e374d74f42f631e0f57e915c

PR #14957, opened 2025-02-18
Fix monaco model reference creation

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/7671
Closes https://github.com/eclipse-theia/theia/issues/14842

As outlined in https://github.com/eclipse-theia/theia/issues/14842, the culprit for a lot of these issues seem to be that we don't actually create real `MonacoEditorModel` references for our inline editors. These models weren't synced with the plugin host and so we encountered a ton of errors within the plugin host which were then simply propagated and logged in the frontend.

This fixes the issue by ensuring that we actually create a `Resource` for each `MonacoEditorModel` that we attempt to create. This ensures that the models are correctly send to the plugin host.

Additionally adjusts how we create editor models for the notebook to ensure that they are all placed in one single reference list. Previously we had 2 lists, which then led to clashes inside of monaco.

Also fixes a very minor issue related to completion and language support in notebook editors.

#### How to test

1. Open a JS file and add a conditional breakpoint.
2. Open the AI chat widget and type something.
3. Ensure that no errors about unknown models/files is being logged
4. Open a notebook (`.ipynb`) file and scroll through it.
5. Ensure that no errors about duplicate monaco models are being logged.

Please also ensure that https://github.com/eclipse-theia/theia/issues/14924 remains fixed. I couldn't reproduce it (but I also only tested on Windows)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
@@ -14,7 +14,7 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 import { ChangeSet, ChangeSetElement, ChatAgent, ChatChangeEvent, ChatModel, ChatRequestModel } from '@theia/ai-chat';
-import { Disposable, nls, UntitledResourceResolver } from '@theia/core';
+import { Disposable, InMemoryResources, URI, nls } from '@theia/core';
 import { ContextMenuRenderer, LabelProvider, Message, ReactWidget } from '@theia/core/lib/browser';
 import { Deferred } from '@theia/core/lib/common/promise-util';
 import { inject, injectable, optional, postConstruct } from '@theia/core/shared/inversify';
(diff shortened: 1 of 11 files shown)


msujew:
cc @sdirix Can you check whether this does not reintroduce https://github.com/eclipse-theia/theia/issues/14924?

msujew:
> Edit: Looking at the code, it seems you did exactly that with the MonacoEditorModelFilter, right?

Yes, the filter will prevent the model from being propagated to to plugin host. However, in order to prevent the errors from being logged, we currently **want to send the models to the plugin host**. A different fix for this would include 2 parts:
1. Create filters for those specific models
2. Check whether a model reference exists for a URI before calling the plugin host (i.e. when calling `provideCodeActions`).

tsmaeder:
@sdirix the monaco `StandaloneCodeEditor` can be largely configured. See IEditorOptions in the VS Code codebase. In particular code lens can be turned off. Not sure we're surfacing this configuration when creating inline editors, but it feels like a cleaner way to turn off editor features.

jonah-iden (approved):
Looks like a nice solution to me. 

Im just getting a few different errors sometimes when opening notebooks like: 

`Assertion failed (document 'vscode-notebook-cell:/c%3A/Typefox/Open_Source/theia_test_workspace/notebooks/1-cell-test.ipynb#W3sZmlsZQ%3D%3D' doesn't exist` or
`A resource provider for 'vscode-notebook-cell:/c%3A/Typefox/Open_Source/theia_test_workspace/notebooks/1-cell-test.ipynb#W0sZmlsZQ%3D%3D`

But thats probably out of scope for this PR?

sdirix:
I can confirm that https://github.com/eclipse-theia/theia/issues/14924 is still fixed!

msujew:
> But thats probably out of scope for this PR?

@jonah-iden I can't really reproduce those error logs. I would assume they're either unrelated or a consequence of existing errors that are only uncovered through this change. I would indeed try to fix them in a separate PR.

Issue #14879, opened 2025-02-10
Introduce "Active Element" concept


tsmaeder:
### Feature Description:
Right now, many decisions, in particular menu enablement, are based on the focused element. If we talk of "focused element", one needs to understand that "focus" always means "keyboard focus". This leads to problems like this one: https://github.com/eclipse-theia/theia/issues/14002. Without going into detail: the problem there is that the menu bar accepts the focus (it needs, to, otherwise it would not be navigable by keyboard and therefore not accessible). This leads to the menu being rebuilt. We also have some workarounds in the code because Theia needs the focus on the "right" element to compute the enabled and visible elements. This code is brittle and prone to break when changing anything (like making dynamic menu contributions actually work).

I propose the we introduce the concept of "active element" per window and base action enablement upon this concept (i.e. the context keys get set when the active element changes). The Eclipse IDE has the similar concept of "WorkbenchPart", which can be either a view or an editor. In Theia, we'd have to analyze which elements constitute the parts, probably editors and view parts?

A workbench part generally becomes the active part through different routes:
1. An UI element inside the part accepts focus
2. The Window in which the part resides becomes the active window (for example because the use opens the window) and the part was the last or default active part for this window
3. The part is opened
4. The active part is changed through some keybinding or programmatic means (think "the debugger stopped on a breakpoint and now the debugger should become the active part")

When the keyboard focus moves out of the active part, the part remains the active part until another part becomes the active part. So things like the toolbar and menus can never be the active part and influcence action enablement.

### Benefits
We could get rid of the hacks in our menus and streamline the action enablement code. I just spent two days analyzing an issue in our menu because we're overriding stuff from lumino/phosphor in complicate ways.

sdirix:
@tsmaeder Thanks for opening the issue.

Let me emphasize that the current focus-based system is inherently broken. As focus is a specific concept in the Browser with all kind of associated behavior, reusing it for the "activeElement" management leads to:
 - a lot of [hacky code](https://github.com/eclipse-theia/theia/blob/8b3ea1cbfbfd93e621fc1bff14d24d400656748a/packages/core/src/browser/menu/browser-menu-plugin.ts#L266-L283) which easily breaks
 - weird observable user issues like elements gaining/losing focus without user interaction
 - usability issues as we have a lot of focus manipulation, breaking keyboard navigation, screen readers etc.

PR #14959, opened 2025-02-18
Fix problems related with menu bar updates on focus change.

### LINKED ISSUE #14002: Menu bar does not open on first click when certain sidebar elements have focus
### Bug Description:
When certain sidebar elements have focus, clicking on the menu bar does not open the selected menu group, but appears to 'weaken' the focus on the element, thus another click will open the menu group.

Clicking again on the menu group to close it returns focus to the original element and so the bug will repeat if attempting to reopen the menu.

For an example of 'weakening focus': when the search input box has focus, the first click on the menu bar will remove the "(up/down for history)" text from the input box, but the insertion caret will still be flashing.

### Steps to Reproduce:

1. Click inside an affected sidebar panel or on it's icon in the side panel selector
2. Click on the menu bar

Here is the results of my experimentation on whether different sidebar elements exhibit the bug behaviour (Yes means affected):
Explorer
- Explorer heading: No
- Open Editors sub-heading: No
- Open Editors box: No
- Project Folder sub-heading: No
- Project Folder box: Yes
Search
- Search heading: No
- Search input box: Yes (and each toggle box - Match case, etc)
- Toggle replace button: Yes
- Replace input box: Yes
- Search results box: Yes
Source Control
- Source Control heading: No
- Rest of box: Yes
- (Note that due to https://github.com/eclipse-theia/theia/issues/14001 I cannot install any SCM extensions, thus I cannot try any other elements)
Debug
- No elements appear to be affected
Extensions
- Extensions heading: No
- Search input box: Yes
- Recommended sub-heading: No
- Recommended box: Yes
- Installed sub-heading: No
- Installed box: Yes
Testing
- Testing heading: No
- Test explorer sub-heading: No
- Test explorer box: Yes
- Test runs sub-heading: No
- Test runs box : No

### Additional Information

- Operating System: Windows 10 (Build 19045.4651)
- Theia Version: 1.51.0

tsmaeder:
#### What it does
This PR introduces an event object describing changes to the menu registry more precisely. The code that manages the menu bar then only rebuilds itself when the change actually concerns the main menu bar. This prevents the menu bar from being replaced when the keyboard focus changes.

Fixes #14002

#### How to test
Make sure you set the "title bar style" setting to "custom" (This issue does not affect the native menus). Observe that the Sample menu appears dynamically and is correctly displayed and works. Make sure the scenario from the linked issue is fixed. Test electron and browser cases. 

#### Follow-ups
This is a workaround that fixes the problem at hand. To attack the root problem, we would have to solve https://github.com/eclipse-theia/theia/issues/14879

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/menu/browser-menu-plugin.ts
+++ b/packages/core/src/browser/menu/browser-menu-plugin.ts
@@ -15,18 +15,20 @@
 // *****************************************************************************
 
 import { injectable, inject } from 'inversify';
-import { MenuBar, Menu as MenuWidget, Widget } from '@phosphor/widgets';
+import { Menu, MenuBar, Menu as MenuWidget, Widget } from '@phosphor/widgets';
 import { CommandRegistry as PhosphorCommandRegistry } from '@phosphor/commands';
+import { ElementExt } from '@phosphor/domutils';
 import {
     CommandRegistry, environment, DisposableCollection, Disposable,
-    MenuModelRegistry, MAIN_MENU_BAR, MenuPath, MenuNode, MenuCommandExecutor, CompoundMenuNode, CompoundMenuNodeRole, CommandMenuNode
+    MenuModelRegistry, MAIN_MENU_BAR, MenuPath, MenuNode, MenuCommandExecutor, CompoundMenuNode, CompoundMenuNodeRole, CommandMenuNode,
+    ArrayUtils
 } from '../../common';
 import { KeybindingRegistry } from '../keybinding';
 import { FrontendApplication } from '../frontend-application';
 import { FrontendApplicationContribution } from '../frontend-application-contribution';
 import { ContextKeyService, ContextMatcher } from '../context-key-service';
 import { ContextMenuContext } from './context-menu-context';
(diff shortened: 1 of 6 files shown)


packages/core/src/browser/menu/browser-menu-plugin.ts
```diff
@@ -156,6 +160,10 @@ export class BrowserMainMenuFactory implements MenuWidgetFactory {
 
 }
 
+export function isMenuElement(element: HTMLElement | null): boolean {
+    return !!element && element.className.includes('lm-Menu');
+}
```
msujew, line 165: Question: Is this required? It does not seem to be in use and also uses Lumino stuff.

msujew (approved):
Looks good to me 👍

packages/core/src/browser/menu/browser-menu-plugin.ts
tsmaeder: Good catch, I ported this commit from the lumino-PR.

PR #14969, opened 2025-02-18
fix(chat): move variable drop handler registration to the frontend

planger:
#### What it does

In #14787 we've introduced a DOM type dependency to the common package, which prevents it from being properly used on the backend. This change moves the parts that depend on the DOM to the frontend.

One potentially ugly aspect of this change might be that the contributions on the frontend can expect an extended interface of the variable service (`FrontendVariableContribution` hands in a `FrontendVariableService` as a parameter instead of the `AIVariableService`) without explicitly knowing it. I've added a comment to the `AIVariableContribution` interface, but I'm not sure this is enough or whether we should make that explicit and have two separate contribution points?

In general, I think it is good practice that we keep as much code as possible independent of the frontend and in common, however, at the same time this adds complexity in this case which isn't really justified by current or expected use. So maybe just moving this service to the frontend entirely is the cleaner solution.

#### How to test

Ensure the functionality added by #14787 still works.

#### Follow-ups

None

#### Breaking changes

#### Attribution

None

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
@@ -23,7 +23,8 @@ import { IMouseEvent } from '@theia/monaco-editor-core';
 import { MonacoEditor } from '@theia/monaco/lib/browser/monaco-editor';
 import { MonacoEditorProvider } from '@theia/monaco/lib/browser/monaco-editor-provider';
 import { CHAT_VIEW_LANGUAGE_EXTENSION } from './chat-view-language-contribution';
-import { AIVariableResolutionRequest, AIVariableService } from '@theia/ai-core';
+import { AIVariableResolutionRequest } from '@theia/ai-core';
+import { FrontendVariableService } from '@theia/ai-core/lib/browser';
 import { ContextVariablePicker } from './context-variable-picker';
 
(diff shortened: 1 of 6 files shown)


colin-grant-work (approved):
Changes look good to me.

> whether we should make that explicit and have two separate contribution points?

In principle, I'd probably be in favor of this approach so that we can define the appropriate methods exclusively on the frontend and have an explicit statement there that something like `registerDropHandler` will receive `FrontendVariableService` but `registerVariables` might receive the 'normal' version, but I don't have especially strong feelings. The fact is that once a service gotten a hold of some registry in a `register` function, they can do whatever they want, and perform whatever checks they want, so I'm not sure we're actually preventing any mischief.

sdirix:
>In general, I think it is good practice that we keep as much code as possible independent of the frontend and in common, however, at the same time this adds complexity in this case which isn't really justified by current or expected use. So maybe just moving this service to the frontend entirely is the cleaner solution.

I think it's worth it. There are certainly adopters who would like to use Theia AI within the backend, headlessly or within test cases. Keeping the core services free from browser specifics helps to achieve that.

planger:
Thanks for the feedback @colin-grant-work and @sdirix! Let's go with this current approach then for now. Thanks!

PR #14971, opened 2025-02-18
Add `contextSummary` variable to AI system

### LINKED ISSUE #14895: [Theia AI] Integrate context feature with Coder
This is based on https://github.com/eclipse-theia/theia/pull/14787

Coder should actively take context elements manually added by the user into account.

## Background

The context is a set of resolved variables in the chat request model. Context elements can be added by the user either by mentioning them in the chat input or by just attaching them to the request, without explicitly mentioning them. The rationale behind adding context elements is to tell the agent: take those elements into account in your reply. How agents actually use them is up to their implementation.

## Goal of this ticket

The Coder agent should consider the context elements by mentioning them in a list as part of the system message. Coder also includes tool functions that provide the LLM with workspace access. Thus, Coder does not need to process or attach the entire context, but rather mention to the LLM that the user has specified some additional context elements (files at the moment), to guide the LLM in reading them whenever needed on demand.

Therefore, the system message of Coder should be enhanced with a prompt variable at the end of the prompt that can resolve to a stringified overview of the context, such as:

```
## Additional context given
The following files have been provided for additional context (some of them are referred to above)
- filepath: description (if any)
- filePath2: description
```

### Step 1

Enable the usage of variables that require chat session access in the system message.

As we want to add a variable that reads the context from the chat request, we need to provide the variable resolver with the chat model context. This works for variables added by users to normal user messages, because the chat service [provides the chat model / session as context](https://github.com/eclipse-theia/theia/blob/master/packages/ai-chat/src/common/chat-service.ts#L198) for resolving the mentioned variables.

However, this is currently not the case for variables system messages. The abstract chat agent [calls `getSystemMessageDescription()`](https://github.com/eclipse-theia/theia/blob/master/packages/ai-chat/src/common/chat-agents.ts#L164) without providing the chat request model and then just plainly adds the returned system message text. Concrete chat agents will usually use the [prompt service to get the system message](https://github.com/eclipse-theia/theia/blob/master/packages/ai-ide-agents/src/browser/coder-agent.ts#L45), so normal variables will be resolved, however, the chat model will not be provided as a context in this case, which is why variables that want to read the chat request's context, won't work

So we need to:
- Add the session as an optional parameter to getSystemMessageDescription in agents (to pass this to the agent)
- Add an optional parameter to getPrompt to pass the respective context to the prompt service
- use this parameter then if available in the prompt service to resolve variables like they are resolved in the chat
The optional parameter should be used by all existing chat agents.

To harmonize passing the chat context for variables between the chat service and the chat agents, we may want to pull this out into a separate service altogether. But this is optional as you see fit.

### Step 2

Implement a variable called "contextSummary" that resolves to the text described above reading from the context.

### Step 3

Add the "contextSummary" variable at the end of the Coder system message

### Step 4

Add a new tool function (implementation and to the Coder system message) that allows the LLM to add interesting files (with a short description) to the context itself, to help finding the correct files in future requests.

## Notes

At the moment context elements don't have a description (https://github.com/eclipse-theia/theia/pull/14787). If this is needed, it'll need to be added there. Adding this field is easy. The more critical question is how to let the user define and edit this description. Maybe showing a pencil icon that opens a quick input as a first step?

### LINKED ISSUE #14945: [Theia AI] Optimize / resolve file context variable for Universal and Workspace
based on https://github.com/eclipse-theia/theia/issues/14895

For the Workspace agent we do the same integration as with Coder. (use the same function and variable)
For universal, as we do not have functions, we add a list of all files with their content via a new separate variable.

For Coder, we should also resolve the changeset in its current state

colin-grant-work:
- Adds context to end of Coder agent's prompt
- Adds a function to allow agents to supplement the context

#### What it does

Fixes #14895, Closes #14945

image

#### How to test

1. Add a file to the context following the instructions in #14787
2. Then ask Coder about something in the context. The agent should know what you're talking about.
3. Ask Coder to add something to the context. It should succeed in doing so, and the file should pop up in the UI.

#### Follow-ups

1. I notice that files in the context are getting labeled with a leading slash, making them appear absolute. We could address that, if we think it's confusing.
2. I also notice that if you drag+drop a file from the explorer tree onto the AI input widget, the text input gets populated with the file information:

image

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/change-set-variable.ts
+++ b/packages/ai-chat/src/common/change-set-variable.ts
@@ -33,7 +33,7 @@
 import { MaybePromise, nls } from '@theia/core';
 import { injectable } from '@theia/core/shared/inversify';
 import { AIVariable, ResolvedAIVariable, AIVariableContribution, AIVariableResolver, AIVariableService, AIVariableResolutionRequest, AIVariableContext } from '@theia/ai-core';
-import { ChatSessionContext } from './chat-service';
+import { ChatSessionContext } from './chat-agents';
 
 export const CHANGE_SET_SUMMARY_VARIABLE: AIVariable = {
     id: 'changeSetSummary',
@@ -54,10 +54,10 @@ export class ChangeSetVariableContribution implements AIVariableContribution, AI
     }
 
     async resolve(request: AIVariableResolutionRequest, context: AIVariableContext): Promise<ResolvedAIVariable | undefined> {
-        if (!ChatSessionContext.is(context) || request.variable.name !== CHANGE_SET_SUMMARY_VARIABLE.name || !context.session.model.changeSet?.getElements().length) { return undefined; }
+        if (!ChatSessionContext.is(context) || request.variable.name !== CHANGE_SET_SUMMARY_VARIABLE.name || !context.model.changeSet?.getElements().length) { return undefined; }
         return {
             variable: CHANGE_SET_SUMMARY_VARIABLE,
-            value: context.session.model.changeSet.getElements().map(element => `- file: ${element.uri.toString()}, status: ${element.state}`).join('\n-')
+            value: context.model.changeSet.getElements().map(element => `- file: ${element.uri.toString()}, status: ${element.state}`).join('\n-')
         };
     }
 }
--- a/packages/ai-chat/src/common/change-set-variable.ts
+++ b/packages/ai-chat/src/common/change-set-variable.ts
@@ -1,0 +1,64 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { MaybePromise, nls } from '@theia/core';
+import { injectable } from '@theia/core/shared/inversify';
+import { AIVariable, ResolvedAIVariable, AIVariableContribution, AIVariableResolver, AIVariableService, AIVariableResolutionRequest, AIVariableContext } from '@theia/ai-core';
+import { ChatSessionContext } from './chat-service';
+
+export const CHANGE_SET_SUMMARY_VARIABLE: AIVariable = {
+    id: 'changeSetSummary',
+    description: nls.localize('theia/ai/core/changeSetSummaryVariable/description', 'Provides a summary of the files in a change set and their contents.'),
+
+    name: 'changeSetSummary',
+};
+
+@injectable()
+export class ContextSummaryVariableContribution implements AIVariableContribution, AIVariableResolver {
+
+    registerVariables(service: AIVariableService): void {
+        service.registerResolver(CHANGE_SET_SUMMARY_VARIABLE, this);
+    }
+
+    canResolve(request: AIVariableResolutionRequest, context: AIVariableContext): MaybePromise<number> {
+        return request.variable.name === CHANGE_SET_SUMMARY_VARIABLE.name ? 50 : 0;
+    }
+
+    async resolve(request: AIVariableResolutionRequest, context: AIVariableContext): Promise<ResolvedAIVariable | undefined> {
+        if (!ChatSessionContext.is(context) || request.variable.name !== CHANGE_SET_SUMMARY_VARIABLE.name || !context.session.changeSet?.getElements().length) { return undefined; }
+        return {
+            variable: CHANGE_SET_SUMMARY_VARIABLE,
+            value: context.session.changeSet.getElements().map(element => `- file: ${element.uri.toString()}, status: ${element.state}`).join('\n-')
+        };
+    }
+}
+
--- a/packages/ai-chat/src/common/change-set-variable.ts
+++ b/packages/ai-chat/src/common/change-set-variable.ts
@@ -43,7 +43,7 @@ export const CHANGE_SET_SUMMARY_VARIABLE: AIVariable = {
 };
 
 @injectable()
-export class ContextSummaryVariableContribution implements AIVariableContribution, AIVariableResolver {
+export class ChangeSetVariableContribution implements AIVariableContribution, AIVariableResolver {
 
     registerVariables(service: AIVariableService): void {
         service.registerResolver(CHANGE_SET_SUMMARY_VARIABLE, this);
@@ -54,10 +54,10 @@ export class ContextSummaryVariableContribution implements AIVariableContributio
     }
 
     async resolve(request: AIVariableResolutionRequest, context: AIVariableContext): Promise<ResolvedAIVariable | undefined> {
-        if (!ChatSessionContext.is(context) || request.variable.name !== CHANGE_SET_SUMMARY_VARIABLE.name || !context.session.changeSet?.getElements().length) { return undefined; }
+        if (!ChatSessionContext.is(context) || request.variable.name !== CHANGE_SET_SUMMARY_VARIABLE.name || !context.session.model.changeSet?.getElements().length) { return undefined; }
         return {
             variable: CHANGE_SET_SUMMARY_VARIABLE,
-            value: context.session.changeSet.getElements().map(element => `- file: ${element.uri.toString()}, status: ${element.state}`).join('\n-')
+            value: context.session.model.changeSet.getElements().map(element => `- file: ${element.uri.toString()}, status: ${element.state}`).join('\n-')
         };
     }
 }
--- a/packages/ai-chat/src/common/change-set-variable.ts
+++ b/packages/ai-chat/src/common/change-set-variable.ts
@@ -31,9 +31,10 @@
 // *****************************************************************************
 
 import { MaybePromise, nls } from '@theia/core';
-import { injectable } from '@theia/core/shared/inversify';
+import { inject, injectable } from '@theia/core/shared/inversify';
 import { AIVariable, ResolvedAIVariable, AIVariableContribution, AIVariableResolver, AIVariableService, AIVariableResolutionRequest, AIVariableContext } from '@theia/ai-core';
 import { ChatSessionContext } from './chat-agents';
+import { WorkspaceService } from '@theia/workspace/lib/browser';
 
 export const CHANGE_SET_SUMMARY_VARIABLE: AIVariable = {
     id: 'changeSetSummary',
@@ -44,6 +45,8 @@ export const CHANGE_SET_SUMMARY_VARIABLE: AIVariable = {
 
 @injectable()
 export class ChangeSetVariableContribution implements AIVariableContribution, AIVariableResolver {
+    @inject(WorkspaceService)
+    protected readonly workspaceService: WorkspaceService;
 
     registerVariables(service: AIVariableService): void {
         service.registerResolver(CHANGE_SET_SUMMARY_VARIABLE, this);
@@ -55,9 +58,12 @@ export class ChangeSetVariableContribution implements AIVariableContribution, AI
 
     async resolve(request: AIVariableResolutionRequest, context: AIVariableContext): Promise<ResolvedAIVariable | undefined> {
         if (!ChatSessionContext.is(context) || request.variable.name !== CHANGE_SET_SUMMARY_VARIABLE.name || !context.model.changeSet?.getElements().length) { return undefined; }
+        const entries = await Promise.all(
+            context.model.changeSet.getElements().map(async element => `- file: ${await this.workspaceService.getWorkspaceRelativePath(element.uri)}, status: ${element.state}`)
+        );
         return {
             variable: CHANGE_SET_SUMMARY_VARIABLE,
-            value: context.model.changeSet.getElements().map(element => `- file: ${element.uri.toString()}, status: ${element.state}`).join('\n-')
+            value: entries.join('\n')
         };
     }
 }
--- a/packages/ai-chat/src/common/chat-model.ts
+++ b/packages/ai-chat/src/common/chat-model.ts
@@ -24,7 +24,7 @@ import { MarkdownString, MarkdownStringImpl } from '@theia/core/lib/common/markd
 import { Position } from '@theia/core/shared/vscode-languageserver-protocol';
 import { ChatAgentLocation } from './chat-agents';
 import { ParsedChatRequest } from './parsed-chat-request';
-import { ResolvedAIContextVariable } from '@theia/ai-core';
+import { AIVariableResolutionRequest, ResolvedAIContextVariable } from '@theia/ai-core';
 
 /**********************
  * INTERFACES AND TYPE GUARDS
@@ -33,6 +33,8 @@ import { ResolvedAIContextVariable } from '@theia/ai-core';
 export type ChatChangeEvent =
     | ChatAddRequestEvent
     | ChatAddResponseEvent
+    | ChatAddVariableEvent
+    | ChatRemoveVariableEvent
     | ChatRemoveRequestEvent
     | ChatSetChangeSetEvent
     | ChatUpdateChangeSetEvent
@@ -64,6 +66,14 @@ export interface ChatRemoveChangeSetEvent {
     changeSet: ChangeSet;
 }
 
+export interface ChatAddVariableEvent {
+    kind: 'addVariable';
+}
+
+export interface ChatRemoveVariableEvent {
+    kind: 'removeVariable';
+}
+
 export namespace ChatChangeEvent {
     export function isChangeSetEvent(event: ChatChangeEvent): event is ChatSetChangeSetEvent | ChatUpdateChangeSetEvent | ChatRemoveChangeSetEvent {
         return event.kind === 'setChangeSet' || event.kind === 'removeChangeSet' || event.kind === 'updateChangeSet';
@@ -84,6 +94,7 @@ export interface ChatModel {
     readonly id: string;
     readonly location: ChatAgentLocation;
     readonly changeSet?: ChangeSet;
+    readonly context: ChatContextManager;
     getRequests(): ChatRequestModel[];
     isEmpty(): boolean;
 }
@@ -93,6 +104,14 @@ export interface ChangeSet extends Disposable {
     readonly title: string;
     getElements(): ChangeSetElement[];
     dispose(): void;
+}
+
+export interface ChatContextManager {
+    onDidChange: Event<ChatAddVariableEvent | ChatRemoveVariableEvent>;
+    getVariables(): readonly AIVariableResolutionRequest[]
+    addVariables(...variables: AIVariableResolutionRequest[]): void;
+    deleteVariables(...indices: number[]): void;
+    clear(): void;
 }
 
 export interface ChangeSetElement {
@@ -478,11 +497,13 @@ export class MutableChatModel implements ChatModel, Disposable {
     protected _requests: MutableChatRequestModel[];
     protected _id: string;
     protected _changeSet?: ChangeSetImpl;
+    protected readonly _contextManager = new ChatContextManagerImpl();
 
     constructor(public readonly location = ChatAgentLocation.Panel) {
         // TODO accept serialized data as a parameter to restore a previously saved ChatModel
         this._requests = [];
         this._id = generateUuid();
+        this._contextManager.onDidChange(e => this._onDidChangeEmitter.fire(e));
     }
 
     getRequests(): MutableChatRequestModel[] {
@@ -499,6 +520,10 @@ export class MutableChatModel implements ChatModel, Disposable {
 
     get changeSet(): ChangeSetImpl | undefined {
         return this._changeSet;
+    }
+
+    get context(): ChatContextManager {
+        return this._contextManager;
     }
 
     setChangeSet(changeSet: ChangeSetImpl | undefined): void {
@@ -612,6 +637,51 @@ export class ChangeSetImpl implements ChangeSet {
     dispose(): void {
         this._onDidChangeEmitter.dispose();
         this._elements.forEach(element => element.dispose?.());
+    }
+}
+
+export class ChatContextManagerImpl implements ChatContextManager {
+    protected readonly variables = new Array<AIVariableResolutionRequest>();
+    protected readonly onDidChangeEmitter = new Emitter<ChatAddVariableEvent | ChatRemoveVariableEvent>();
+    get onDidChange(): Event<ChatAddVariableEvent | ChatRemoveVariableEvent> {
+        return this.onDidChangeEmitter.event;
+    }
+
+    getVariables(): readonly AIVariableResolutionRequest[] {
+        const result = this.variables.slice();
+        Object.freeze(result);
+        return result;
+    }
+
+    addVariables(...variables: AIVariableResolutionRequest[]): void {
+        let modified = false;
+        variables.forEach(variable => {
+            if (this.variables.some(existing => existing.variable.id === variable.variable.id && existing.arg === variable.arg)) {
+                return;
+            }
+            this.variables.push(variable);
+            modified = true;
+        });
+        if (modified) {
+            this.onDidChangeEmitter.fire({ kind: 'addVariable' });
+        }
+    }
+
+    deleteVariables(...indices: number[]): void {
+        const toDelete = indices.filter(candidate => candidate <= this.variables.length).sort((left, right) => right - left);
+        if (toDelete.length) {
+            toDelete.forEach(index => {
+                this.variables.splice(index, 1);
+            });
+            this.onDidChangeEmitter.fire({ kind: 'removeVariable' });
+        }
+    }
+
+    clear(): void {
+        if (this.variables.length) {
+            this.variables.length = 0;
+            this.onDidChangeEmitter.fire({ kind: 'removeVariable' });
+        }
     }
 }
 
--- a/packages/ai-chat/src/common/context-details-variable.ts
+++ b/packages/ai-chat/src/common/context-details-variable.ts
@@ -19,28 +19,43 @@ import { injectable } from '@theia/core/shared/inversify';
 import { AIVariable, ResolvedAIVariable, AIVariableContribution, AIVariableResolver, AIVariableService, AIVariableResolutionRequest, AIVariableContext } from '@theia/ai-core';
 import { ChatSessionContext } from './chat-service';
 
-export const CONTEXT_SUMMARY_VARIABLE: AIVariable = {
-    id: 'contextSummary',
-    description: nls.localize('theia/ai/core/contextSummaryVariable/description', 'Describes files in the context for a given session.'),
-    name: 'contextSummary',
+export const CONTEXT_DETAILS_VARIABLE: AIVariable = {
+    id: 'contextDetails',
+    description: nls.localize('theia/ai/core/contextDetailsVariable/description', 'Provides full text values and descriptions for all context variables.'),
+    name: 'contextDetails',
 };
 
 @injectable()
 export class ContextSummaryVariableContribution implements AIVariableContribution, AIVariableResolver {
+    protected variableService: AIVariableService | undefined;
+
     registerVariables(service: AIVariableService): void {
-        service.registerResolver(CONTEXT_SUMMARY_VARIABLE, this);
+        this.variableService = service;
+        service.registerResolver(CONTEXT_DETAILS_VARIABLE, this);
     }
 
     canResolve(request: AIVariableResolutionRequest, context: AIVariableContext): MaybePromise<number> {
-        return request.variable.name === CONTEXT_SUMMARY_VARIABLE.name ? 50 : 0;
+        return request.variable.name === CONTEXT_DETAILS_VARIABLE.name ? 50 : 0;
     }
 
     async resolve(request: AIVariableResolutionRequest, context: AIVariableContext): Promise<ResolvedAIVariable | undefined> {
-        if (!ChatSessionContext.is(context) || request.variable.name !== CONTEXT_SUMMARY_VARIABLE.name) { return undefined; }
+        if (!ChatSessionContext.is(context) || request.variable.name !== CONTEXT_DETAILS_VARIABLE.name || !this.variableService) { return undefined; }
         return {
-            variable: CONTEXT_SUMMARY_VARIABLE,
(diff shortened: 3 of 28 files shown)


JonasHelming:
Great, thank you for this integration!! The whole context story really significantly improves the work with Coder.

I did not look at the code yet, but tested a few things. My observations so far:

# Case 1: Context added in chat
This worked before and the additional context summary does nothing bad (as expected). 

# Case 2: Context in the chat + some files only in the context
When I add additional files without any further information, and it was not related to the task, the LLM did not even look at it. When I provided an additional file that was actually relevant for the task, the LLM looked at it, it probably guessed that from the file name and the context. Looks definitely good enough to merge this to me, but we have to evaluate in practise.

# Case 3: Context only in the context, not in the chat
It looked at the file(s) most of the time. Only if the prompt does not imply any implementation task, it did not. Example:
"Do not display a message to the user anymore about the tools" with context mcp-command-contribution.ts => It said: OK I won't :-D

So from case 2 and 3 we identify a **Potential issue 1**  that the LLM does not look at some files only in the context, but the user expects it to. I found only one prompt though, where this was really an issue. We should evaluate this in practice. If it is a problem, we can augment the prompt in iterations (during internal test) , e.g.:

- "Look at them if you need them for your task"
- "Always look at them to understand your task"

So issue 1 is not a blocker for this PR.


# Case 4: Agent builds the context by itself

**Issue 2**: If a file is added to the context multiple times, it appears there multiple times (@planger I believe this is an issue we should solve in the context layer, WDYT?)

**Issue 3** With my prompts, the LLM rather rarely "remembered" a file by itself.

I tried two options how to fix this:

Option A: Add the context add function to a sub bullet in the Context Retrieval section "If you find a file interesting for the task, remember it" => The issue here was that the LLM tended to add irrelevant files, but this needs to be evaluated.

Option B: Add the context function to a sub bullet in the apply change area "- **Remember changed file**: Use using~{context_addFile} to remember all files you change and files that were needed to create the changes." => Seemed to have better results than A as it does that in a phase, were it implements.

Option C: We add all modified files to the context in the modify functions ourselves (did not try). This very likely makes sense anyways.

Option A and B can be tried "on the go" as they are prompt modifications. Option C we have to implement.

My current option would be to go for Option C as a default. 
Reasoning:
- Our current agents do not tend to look at "other relevant files" by itself, e.g. for interface declarations. So the user has to typically provide these themselves and the LLM might add irrelevant things.
- We have the context add function in place and can evaluate it more in practice with a prompt variant.

All described issues IMHO do not block this PR, we can handle them in follow-ups

Thanks again for this great addition!

planger:
> Issue 2: If a file is added to the context multiple times, it appears there multiple times (@planger I believe this is an issue we should solve in the context layer, WDYT?)

Yes, this can be considered a bug. We shouldn't add the same element twice. I'll put it on my list.

JonasHelming:
A pretty 'satisfying' user prompt for me to test this was the following, works well with 40 and o3-mini. it is actually the follow-up of this PR, so the PR follow-ups itself :-)

---
@Coder Add a new variable to   #file:packages/ai-chat/src/common/context-summary-variable.ts that fully resolves the elements in the context including their content. Look at   #file:packages/ai-ide-agents/src/browser/context-functions.ts on how to resolve the context. Now add this variable to the prompt of the universal agent (which is define in   #file:packages/ai-ide-agents/src/common/universal-chat-agent.ts). similar to how it is used in the prompt of the coder agent.
---

Note that you should type the file variables out, as they are not added to the context when pasting (the prompt of course still works, but for the test this is important)

I literately adapted the system prompt to the one below. This worked pretty nicely, but it did not add files to the context by itself, so I still believe we should for for option C for now.

---
You are an AI assistant integrated into Theia IDE, designed to assist software developers with code tasks. You can interact with the code base and suggest changes.

## Context Retrieval
Use the following functions to interact with the workspace files if you require context:
- **~{getWorkspaceDirectoryStructure}**: Returns the complete directory structure.
- **~{getWorkspaceFileList}**: Lists files and directories in a specific directory.
- **~{getFileContent}**: Retrieves the content of a specific file.

## Propose Code Changes

  - **Always Retrieve Current Content**: Use getFileContent to get the latest content of the target file.
  - **Remember changed file**: Use ~{context_addFile} to remember all files you change and files that were needed to create the changes.
  - **Change Content**: Use ~{changeSet_writeChangeToFile} or ~{changeSet_replaceContentInFile} to suggest file changes to the user.  Only select and call one function per file.
  
## Additional Context

The following files have been provided for additional context. Some of them may also be referred to above. Always look at the relevant files to understand your task using getFileContent
{{contextSummary}}
---

JonasHelming:
Thought about it a bit more and I believe instead of option C, we should add a resolved version of the changeset (paths and state) to the prompt with another variable. This is now very easy to do, as we have the infrastructure in place. reasoning: the previously changed files are always interesting. I can add this in a follow-up PR.

Then we can go for option A with this PR and optimized from there.

@planger WDYT?

planger (commented):
Thank you very much! Code looks great to me and seems to work fine.

My main concern is how we make the context summary as interpretable as possible for the LLM and give the context variable provider more control over it (see my inline comment).

> Thought about it a bit more and I believe instead of option C, we should add a resolved version of the changeset (paths and state) to the prompt with another variable. This is now very easy to do, as we have the infrastructure in place. reasoning: the previously changed files are always interesting. I can add this in a follow-up PR.
> 
> Then we can go for option A with this PR and optimized from there.

I would agree. I think for now keeping those separated is simpler and more flexible. If we observe that it confuses the LLMs (which might be the case), we can reconsider. But we'll need more evaluation to confirm this even is a problem we would need to solve. :-)

JonasHelming:
Let's go with this prompt then: https://github.com/eclipse-theia/theia/pull/14971#issuecomment-2668098393

packages/ai-chat/src/common/context-summary-variable.ts
```diff
@@ -0,0 +1,46 @@
...
+// *****************************************************************************
+
+import { MaybePromise, nls } from '@theia/core';
+import { injectable } from '@theia/core/shared/inversify';
+import { AIVariable, ResolvedAIVariable, AIVariableContribution, AIVariableResolver, AIVariableService, AIVariableResolutionRequest, AIVariableContext } from '@theia/ai-core';
+import { ChatSessionContext } from './chat-service';
+
+export const CONTEXT_SUMMARY_VARIABLE: AIVariable = {
+    id: 'contextSummary',
+    description: nls.localize('theia/ai/core/todayVariable/description', 'Does something for today'),
```
planger, line 24: I think this is a leftover from the previous implementation. :-)

JonasHelming:
for #14945  we should directly add a second function to this PR, something like the follewing (not compiling semi-pseudo code)

```
export const CONTEXT_RESOLVE_VARIABLE: AIVariable = {
    id: 'contextResolve',
    description: nls.localize('theia/ai/core/contextResolve/description', 'Resolves the full content of context elements'),
    name: 'contextResolve',
};

@injectable()
export class ContextResolvedVariableContribution implements AIVariableContribution, AIVariableResolver {
    registerVariables(service: AIVariableService): void {
        service.registerResolver(CONTEXT_RESOLVE_VARIABLE, this);
    }

    canResolve(request: AIVariableResolutionRequest, context: AIVariableContext): MaybePromise<number> {
        if (request.variable.name === CONTEXT_RESOLVE_VARIABLE.name) {
            return 50;
        }
        return 0;
    }

    async resolve(request: AIVariableResolutionRequest, context: AIVariableContext): Promise<ResolvedAIVariable | undefined> {
        if (!ChatSessionContext.is(context) || request.variable.name !== CONTEXT_RESOLVE_VARIABLE.name) { return undefined; }

        return {
            variable: CONTEXT_RESOLVE_VARIABLE,
            value: (() => {
                context.variables.map((contextElement: any) => {
                    const variable = context.session.context.getVariables().find(v => v.id === contextElement);
                    return variable ? `Content-ID: ${variable.arg}\n + Content: ${variable.contextValue}` : 'Variable not found.'
                }
                
            })()
        }
    };
}
```

JonasHelming:
@colin-grant-work Please test whether declining a changeste element and clearing the changeset works on this branch, I believe it did not when I tested, it works on master. But I am not sure

packages/ai-chat/src/common/chat-model.ts
```diff
@@ -615,6 +640,42 @@ export class ChangeSetImpl implements ChangeSet {
     }
 }
 
+export class ChatContextManagerImpl implements ChatContextManager {
```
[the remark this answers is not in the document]
colin-grant-work: In this case, I was following the example of `ChangeSetImpl`, since that's the closest analogy as a small utility class to help the chat session, but I'm open to suggestions of more descriptive names. We have a few `Default...` classes, which might apply here.

packages/ai-chat/src/common/context-summary-variable.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    canResolve(request: AIVariableResolutionRequest, context: AIVariableContext): MaybePromise<number> {
+        return request.variable.name === CONTEXT_SUMMARY_VARIABLE.name ? 50 : 0;
+    }
+
+    async resolve(request: AIVariableResolutionRequest, context: AIVariableContext): Promise<ResolvedAIVariable | undefined> {
+        if (!ChatSessionContext.is(context) || request.variable.name !== CONTEXT_SUMMARY_VARIABLE.name) { return undefined; }
+        return {
+            variable: CONTEXT_SUMMARY_VARIABLE,
+            value: context.session.context.getVariables().filter(variable => !!variable.arg).map(variable => `- ${variable.arg}`).join('\n')
```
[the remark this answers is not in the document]
colin-grant-work: @planger, thanks for the suggestion. While I like the idea, and, I've basically implemented it for the new `contextDetails` variable for use in agents other than Coder, it has the downside that the `value` field isn't available unless the variable has been resolved, and I don't think we have access to resolved variables at this point - after all, we're busy resolving one ourselves. Or is there somewhere where resolved values are cached (and would it be safe to use that cache)?

We could use the static `description` value associated with the variable itself, something like the file variable's `'Resolves the contents of a file'`. That's not really directed at the LLM, but we could add a field that would be.

colin-grant-work:
@planger, @JonasHelming, I've pushed a new version of the code that addresses some of the comments above, but there is likely still some work to do re: formatting things as intelligibly as possible.

 - I have added two new variables, one for change sets (added to Coder's system prompt), and one for context details, added to the universal agent's prompt.
 - Re: context details, @planger proposed adding various details of resolved variables to the `contextSummary` variable. Since that variable is intended for use with Coder, which can resolve some details on its own, that seems perhaps overkill, but I have attempted to resolve the variables in context for the `contextDetails` variable. Let me know what you think about that approach, but also let me know if there's some way to access resolved variables while resolving other variables; then we could use them as we see fit in both cases, and not resolve them ourselves, as I've done for `contextDetails`.
 - Since there's a lot of data for `contextDetails`, I've just JSON stringified what's available - maybe there's a better way? Maybe I should do that for the change set, as well?

@JonasHelming, I have fixed a bug re: showing the deletion of change sets, so that should be working. On the other hand, I'm still not seeing the agent respond to changes in context during the session, while it seemed like in your interactions, the agent was resolving changes to context later in the session, and not just for the initial interaction.

planger (commented):
Thank you very much for this iteration! I think this goes into the right direction.

I didn't consider the variable dependency issue before (sorry for that), but I feel this is a valid general use case. So I suggest to explicitly support dependent variables.

We need to ensure that we
* Don't inject the variable service into variable resolvers (cyclic DI)
* Aim at taking care of cyclic variable dependencies
* Avoid duplicate resolution

We might be able to address those constraints by handing in a `resolveDependency` callback to resolvers and enhance the variable resolution method.

### Extend the variable resolver

```ts
export interface AIVariableResolverWithVariableDependencies extends AIVariableResolver {
    resolve(
        request: AIVariableResolutionRequest,
        context: AIVariableContext,
        resolveDependency: (req: AIVariableResolutionRequest) => Promise<ResolvedAIVariable | undefined>
    ): Promise<ResolvedAIVariable | undefined>;
}

function isResolverWithDependencies(resolver: AIVariableResolver | undefined): resolver is AIVariableResolverWithVariableDependencies {
    return resolver !== undefined && resolver.resolve.length >= 3;
}
```

### Enhance the resolveVariable method of the variable service

```ts
interface CacheEntry {
    promise: Promise<ResolvedAIVariable | undefined>;
    inProgress: boolean;
}

export class DefaultAIVariableService implements AIVariableService {
    // ... existing properties and methods

    async resolveVariable(
        request: AIVariableArg,
        context: AIVariableContext,
        cache: Map<string, CacheEntry> = new Map()
    ): Promise<ResolvedAIVariable | undefined> {
        const variableName = typeof request === 'string'
            ? request
            : typeof request.variable === 'string'
                ? request.variable
                : request.variable.name;

        if (cache.has(variableName)) {
            const entry = cache.get(variableName)!;
            if (entry.inProgress) {
                this.logger.warn(`Cycle detected for variable: ${variableName}. Skipping resolution.`);
                return undefined;
            }
            return entry.promise;
        }

        const entry: CacheEntry = { promise: Promise.resolve(undefined), inProgress: true };
        cache.set(variableName, entry);

        const promise = (async () => {
            const variable = this.getVariable(variableName);
            if (!variable) {
                return undefined;
            }
            const arg = typeof request === 'string' ? undefined : request.arg;
            const resolver = await this.getResolver(variableName, arg, context);
            let resolved: ResolvedAIVariable | undefined;
            if (isResolverWithDependencies(resolver)) {
                resolved = await resolver.resolve(
                    { variable, arg },
                    context,
                    async (depRequest: AIVariableResolutionRequest) =>
                        this.resolveVariable(depRequest, context, cache)
                );
            } else {
                resolved = await resolver?.resolve({ variable, arg }, context);
            }
            return resolved ? { ...resolved, arg } : undefined;
        })();

        entry.promise = promise;
        promise.finally(() => {
            entry.inProgress = false;
        });

        return promise;
    }
}
```

### Summary

- **AIVariableResolverWithVariableDependencies** lets a resolver access dependencies through a callback during its resolution.
- The updated `resolveVariable` method uses a promise-based cache with an `inProgress` flag for cycle detection. This way, if a variable is already being resolved, it’s skipped—avoiding cycles.
- With this approach, resolved variables are accessible during the resolution of dependent ones. This could be leveraged for both contextSummary (letting Coder resolve details) and contextDetails, reducing redundant resolution.

What do you think about this approach?

Maybe I'm over-complicating this, but I feel like letting variables access other variables might be a useful thing in general, and if it is, we should explicitly support it.

If you or others disagree that we should support that, I'm happily follow your lead and think your approach is fine, given it is rather local and puts the responsibility of using this approach carefully into the hands of a specific variable resolver.

packages/ai-chat/src/common/chat-model.ts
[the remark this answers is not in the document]
planger: Yeah, you are right, I'm happy with either naming and follow what you think feels best, given that we already have a few naming patterns in place here.

packages/ai-chat/src/common/change-set-variable.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export const CHANGE_SET_SUMMARY_VARIABLE: AIVariable = {
+    id: 'changeSetSummary',
+    description: nls.localize('theia/ai/core/changeSetSummaryVariable/description', 'Provides a summary of the files in a change set and their contents.'),
+
+    name: 'changeSetSummary',
+};
+
+@injectable()
+export class ContextSummaryVariableContribution implements AIVariableContribution, AIVariableResolver {
```
planger, line 46: I think we should name this `ChangeSetSummaryVariableContribution` right?

packages/ai-chat/src/common/context-details-variable.ts
```diff
@@ -0,0 +1,61 @@
...
+            value: await this.resolveAllContextVariables(context)
+        };
+    }
+
+    protected async resolveAllContextVariables(context: ChatSessionContext): Promise<string> {
+        const values = await Promise.all(context.session.context.getVariables().map(variable => this.variableService?.resolveVariable(variable, context)));
+        const data = values.filter((candidate): candidate is ResolvedAIVariable => !!candidate)
+            .map(resolved => ({
+                variableKind: resolved.variable.name,
+                variableKindDescription: resolved.variable.description,
```
planger, line 54: You are probably right to remove this.

packages/ai-chat/src/common/context-summary-variable.ts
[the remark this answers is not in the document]
planger: Please see my suggestion on the variable dependencies.

Regarding the description, I totally agree. This was a misconception on my end. The description is not targeted at the LLM and therefore is probably in most cases not a great fit.

packages/ai-chat/src/common/change-set-variable.ts
planger: Currently this variable is targeted at files, even though the context may contain arbitrary context variables, like symbols or even domain-specific things.

If we want to keep it that way, we should also reflect that in the naming. It is already reflected in the description.

In general, I tend to believe we can generalize that, because the only file specific thing is currently the description and the word "file" in line 60. Both can be avoided by using the `contextSummary` field or the variable name.

colin-grant-work:
@planger, following our discussion, my latest push mostly cleans up the existing implementation.

 - I've corrected the naming of and bound the ContextDetails, ContextSummary, and ChangeSet variable contributions.
 - I've  removed the context summary variable's focus on files and exposed the context value resolution function to the both the Architect and Coder agents so that they can context items generically.
 - I've exposed the ContextDetails 
 - I've updated the various places we pass context into variable resolvers to consistently use the `ChatSessionContext` interface where available.

There is one outstanding question from https://github.com/eclipse-theia/theia/issues/14945, namely whether we want to provide more details than the file names and status and provide e.g. the full text of a given suggestion.

Re: follow ups, I have not implemented your suggestions re: explicit support for dependent variables. In the event, it wasn't necessary because we already pre-resolve the context, with which the variables in this PR are concerned. In my view, we shouldn't eagerly resolve the context, because we don't know that it will actually be used, but we should implement some kind of variable caching (perhaps on the request level?) and better support for variable dependencies, since if we don't eagerly resolve context variables, then the context summary will definitely fall into that category.

You'll also see that I've changed the passing of the context a bit: rather than having it passed in as an argument to the chat service, I've located it on the ChatModel. That makes it easier to pass around as part of variable resolution - e.g. when handling system messages - but if you believe that it's important that the context be separated from the model, I'm open to discussion.

JonasHelming:
Note: we just decided in a direct discussion to split the generic context summary variable provided by the framework and the one used by Coder and Architect. the reasoning is that we want to make it maximum simple for Coder and Architect and as the can already access files, they just need the path. Furthermore, as they are part of the IDE, we have under control, which additional future context element types they have to deal with and can handle this explicitly. Having the generic variable in the framework provides value for adopters of Theia AI, as they can directly use this as a good default.

packages/ai-ide/src/common/coder-replace-prompt-template.ts
```diff
@@ -36,14 +39,26 @@ Use the following functions to interact with the workspace files if you require
...
-  - **Always Retrieve Current Content**: Use ${FILE_CONTENT_FUNCTION_ID} to get the latest content of the target file.
-  - **Change Content**: Use ~{changeSet_writeChangeToFile}${withSearchAndReplace ? ' or ~{changeSet_replaceContentInFile}' : ''} to suggest file changes to the user.\
-  ${withSearchAndReplace ? 'Only select and call one function per file.' : ''}`,
+- **Always Retrieve Current Content**: Use ${FILE_CONTENT_FUNCTION_ID} to get the latest content of the target file.
+- **Change Content**: Use ~{changeSet_writeChangeToFile}${withSearchAndReplace ? ' or ~{changeSet_replaceContentInFile}' : ''} to suggest file changes to the user.\
+  ${withSearchAndReplace ? 'Only select and call one function per file.' : ''}
+  
+## Additional Context
+
+The following files have been provided for additional context. Some of them may also be referred to above.\
```
JonasHelming, line 54:
```suggestion
The following files have been provided for additional context. Some of them may also be referred to above. \
```

packages/ai-ide/src/common/coder-replace-prompt-template.ts
```diff
@@ -36,14 +39,26 @@ Use the following functions to interact with the workspace files if you require
...
 
 ## Propose Code Changes
 To propose code changes or any file changes to the user, never print code or new file content in your response.
 
 Instead, for each file you want to propose changes for:
-  - **Always Retrieve Current Content**: Use ${FILE_CONTENT_FUNCTION_ID} to get the latest content of the target file.
-  - **Change Content**: Use ~{changeSet_writeChangeToFile}${withSearchAndReplace ? ' or ~{changeSet_replaceContentInFile}' : ''} to suggest file changes to the user.\
-  ${withSearchAndReplace ? 'Only select and call one function per file.' : ''}`,
+- **Always Retrieve Current Content**: Use ${FILE_CONTENT_FUNCTION_ID} to get the latest content of the target file.
+- **Change Content**: Use ~{changeSet_writeChangeToFile}${withSearchAndReplace ? ' or ~{changeSet_replaceContentInFile}' : ''} to suggest file changes to the user.\
```
JonasHelming, line 49:
```suggestion
- **Change Content**: Use ~{changeSet_writeChangeToFile}${withSearchAndReplace ? ' or ~{changeSet_replaceContentInFile}' : ''} to propose file changes to the user.\
```

packages/ai-ide/src/common/coder-replace-prompt-template.ts
```diff
@@ -36,14 +39,26 @@ Use the following functions to interact with the workspace files if you require
...
+  ${withSearchAndReplace ? 'Only select and call one function per file.' : ''}
+  
+## Additional Context
+
+The following files have been provided for additional context. Some of them may also be referred to above.\
+Always look at the relevant files to understand your task using getFileContent
+{{${CONTEXT_FILES_VARIABLE_ID}}}
+
+## Previously Proposed Changes
+Changes have been proposed for the following files. Some suggestions may have been accepted, while others may still be pending.
```
JonasHelming, line 59:
```suggestion
You have previously proposed changes for the following files. Some suggestions may have been accepted by the user, while others may still be pending.
```

packages/ai-ide/src/common/architect-prompt-template.ts
```diff
@@ -38,5 +39,10 @@ Use the following functions to interact with the workspace files as needed:
 2. **Confirm Paths**: Always verify paths by listing directories or files as you navigate. Avoid assumptions based on user input alone.
 3. **Navigate Step-by-Step**: Move into subdirectories only as needed, confirming each directory level.
 
+## Additional Context
+
+The following files have been provided for additional context. Some of them may also be referred to above.\
```
JonasHelming, line 44:
```suggestion
The following files have been provided for additional context. Some of them may also be referred to above. \
```

JonasHelming (commented):
I tested Coder, Architect and Universal and did not find any case, where this does not work well so far. I added some minor suggestions.
I will let Philip apporve this, if the framework part is fine.

Great work, thank you

packages/ai-chat/src/common/context-details-variable.ts
```diff
@@ -0,0 +1,53 @@
...
+import { MaybePromise, nls } from '@theia/core';
+import { injectable } from '@theia/core/shared/inversify';
+import { AIVariable, ResolvedAIVariable, AIVariableContribution, AIVariableResolver, AIVariableService, AIVariableResolutionRequest, AIVariableContext } from '@theia/ai-core';
+import { dataToJsonCodeBlock } from './chat-string-utils';
+import { ChatSessionContext } from './chat-agents';
+import { CHAT_CONTEXT_DETAILS_VARIABLE_ID } from './context-variables';
+
+export const CONTEXT_DETAILS_VARIABLE: AIVariable = {
+    id: CHAT_CONTEXT_DETAILS_VARIABLE_ID,
+    description: nls.localize('theia/ai/core/contextDetailsVariable/description', 'Provides full text values and descriptions for all context variables.'),
```
planger, line 26:
```suggestion
    description: nls.localize('theia/ai/core/contextDetailsVariable/description', 'Provides full text values and descriptions for all context elements.'),
```

packages/ai-ide/src/common/coder-replace-prompt-template.ts
```diff
@@ -36,14 +39,26 @@ Use the following functions to interact with the workspace files if you require
...
-  ${withSearchAndReplace ? 'Only select and call one function per file.' : ''}`,
+- **Always Retrieve Current Content**: Use ${FILE_CONTENT_FUNCTION_ID} to get the latest content of the target file.
+- **Change Content**: Use ~{changeSet_writeChangeToFile}${withSearchAndReplace ? ' or ~{changeSet_replaceContentInFile}' : ''} to propose file changes to the user.\
+  ${withSearchAndReplace ? 'Only select and call one function per file.' : ''}
+  
+## Additional Context
+
+The following files have been provided for additional context. Some of them may also be referred to above. \
+Always look at the relevant files to understand your task using getFileContent
+{{${CONTEXT_FILES_VARIABLE_ID}}}
```
planger, line 56: Probably not a problem, but we imply in the text above that we always have files, even though there might not be any context files, which will lead to an empty value here. We might want to change the phrasing a bit (like below) and use "if any, the files below...", but I'm not sure if that would impact the quality of the LLM response. Anyway, this can be evaluated in practice and changed later on.

packages/ai-ide/src/common/coder-replace-prompt-template.ts
```diff
@@ -36,14 +39,26 @@ Use the following functions to interact with the workspace files if you require
...
+
+## Previously Proposed Changes
+You have previously proposed changes for the following files. Some suggestions may have been accepted by the user, while others may still be pending.
```
planger, line 59: Same as above, we imply that there always is a changeset, even though there might not be any. But as above, this can be evaluated in practice and changed later on.

planger (approved):
Thank you very much @colin-grant-work for this excellent work and for pushing through all of the discussions we had on this PR!

This looks great and works very well. I didn't test in detail as @JonasHelming already did, but what I tested seemed pretty good.

I don't have any comment that should block merging this change, just a few details that may well be first evaluated in practice or that can be done in a follow-up. Besides my inline comments, I'd suggest to rename and move:
* `ChangeSetVariableContribution`
* `ContextSummaryVariableContribution`
Both are somewhat file-specific at the moment, which is fine as discussed---we can generalize them later, but then we should probably move them to `ai-ide` and rename them. This would then also be in line with e.g. the `ContextFilesVariableContribution`.

Even that however, can be done in a follow-up after the pre-release as it doesn't impact the behavior at all. So I'm fine to merge this and we'll create follow-ups.

packages/ai-ide/src/common/architect-prompt-template.ts
```diff
@@ -38,5 +39,10 @@ Use the following functions to interact with the workspace files as needed:
 2. **Confirm Paths**: Always verify paths by listing directories or files as you navigate. Avoid assumptions based on user input alone.
 3. **Navigate Step-by-Step**: Move into subdirectories only as needed, confirming each directory level.
 
+## Additional Context
+
+The following files have been provided for additional context. Some of them may also be referred to above. \
+Always look at the relevant files to understand your task using getFileContent
```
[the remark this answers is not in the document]
JonasHelming: The function is already referenced above. Did you test this change that mentioning the function twice does not introduce a regression? otherwise I can do.

packages/ai-ide/src/common/coder-replace-prompt-template.ts
```diff
@@ -36,14 +39,26 @@ Use the following functions to interact with the workspace files if you require
...
-  - **Change Content**: Use ~{changeSet_writeChangeToFile}${withSearchAndReplace ? ' or ~{changeSet_replaceContentInFile}' : ''} to suggest file changes to the user.\
-  ${withSearchAndReplace ? 'Only select and call one function per file.' : ''}`,
+- **Always Retrieve Current Content**: Use ${FILE_CONTENT_FUNCTION_ID} to get the latest content of the target file.
+- **Change Content**: Use ~{changeSet_writeChangeToFile}${withSearchAndReplace ? ' or ~{changeSet_replaceContentInFile}' : ''} to propose file changes to the user.\
+  ${withSearchAndReplace ? 'Only select and call one function per file.' : ''}
+  
+## Additional Context
+
+The following files have been provided for additional context. Some of them may also be referred to above. \
+Always look at the relevant files to understand your task using getFileContent
```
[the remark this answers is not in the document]
JonasHelming: same question

packages/ai-ide/src/common/coder-replace-prompt-template.ts
```diff
@@ -36,14 +39,26 @@ Use the following functions to interact with the workspace files if you require
...
-  - **Always Retrieve Current Content**: Use ${FILE_CONTENT_FUNCTION_ID} to get the latest content of the target file.
-  - **Change Content**: Use ~{changeSet_writeChangeToFile}${withSearchAndReplace ? ' or ~{changeSet_replaceContentInFile}' : ''} to suggest file changes to the user.\
-  ${withSearchAndReplace ? 'Only select and call one function per file.' : ''}`,
+- **Always Retrieve Current Content**: Use ${FILE_CONTENT_FUNCTION_ID} to get the latest content of the target file.
+- **Change Content**: Use ~{changeSet_writeChangeToFile}${withSearchAndReplace ? ' or ~{changeSet_replaceContentInFile}' : ''} to propose file changes to the user.\
+  ${withSearchAndReplace ? 'Only select and call one function per file.' : ''}
+  
+## Additional Context
+
+The following files have been provided for additional context. Some of them may also be referred to above. \
```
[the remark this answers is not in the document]
JonasHelming: same comment as above

packages/ai-ide/src/common/coder-replace-prompt-template.ts
JonasHelming: AFAIK, {{${CONTEXT_FILES_VARIABLE_ID}}} resolves to "no files" or something similar if undefned, so should be fine

packages/ai-ide/src/common/coder-replace-prompt-template.ts
JonasHelming: agreed, did not have issues with this so far

packages/ai-ide/src/common/architect-prompt-template.ts
```diff
@@ -38,5 +39,10 @@ Use the following functions to interact with the workspace files as needed:
 2. **Confirm Paths**: Always verify paths by listing directories or files as you navigate. Avoid assumptions based on user input alone.
 3. **Navigate Step-by-Step**: Move into subdirectories only as needed, confirming each directory level.
 
+## Additional Context
+
+The following files have been provided for additional context. Some of them may also be referred to above. \
```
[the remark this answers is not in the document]
planger: Yes, that's even better

packages/ai-ide/src/common/architect-prompt-template.ts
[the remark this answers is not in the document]
planger: Good point! I didn't test this on this PR, but already referenced functions multiple times before in other cases without a problem. A few things have changed in the meantime, so it is probably worth re-testing. If it really is causing an issue, we should definitely track a bug report on that :-)

packages/ai-ide/src/common/coder-replace-prompt-template.ts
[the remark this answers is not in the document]
planger: https://github.com/eclipse-theia/theia/pull/14971#discussion_r1965064048

packages/ai-ide/src/common/coder-replace-prompt-template.ts
planger: No, it returns empty string then:
https://github.com/eclipse-theia/theia/pull/14971/files#diff-0a238071acebd16667ceb34ff1da44e3704ad463f083e9bc47383371cec1bf34R43

JonasHelming: I think I leave this for now to avoid "if". My assumption is that an empty list is easier to comprehend than another "if" in the sentence

packages/ai-ide/src/common/architect-prompt-template.ts
JonasHelming, line 44:
```suggestion
The following files have been provided for additional context. Some of them may also be referred to by the user. \
```

packages/ai-ide/src/common/coder-replace-prompt-template.ts
JonasHelming, line 54:
```suggestion
The following files have been provided for additional context. Some of them may also be referred to by the user. \
```

JonasHelming (approved):
Tested again with the minor updates, no issues found

PR #14981, opened 2025-02-19
Fix next build

sgraband:
#### What it does

Fixed the next build by updating the latest tag to be an annotated tag. 
Adjusted the documentation to make this clearer.
Also fixed some versions that were left behind in the `lerna.json` an `@examples/playwright`. 
Improved the Publishing guide for the native dependency steps.
Removed options from publish commands that are already covered by the lerna config.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


package.json
```diff
@@ -75,9 +75,9 @@
     "lint:clean": "rimraf .eslintcache",
     "preinstall": "node-gyp install",
     "postinstall": "theia-patch && npm run -s compute-references && lerna run afterInstall",
-    "publish:latest": "lerna publish --exact --yes --no-push",
```
jfaltermeier, line 78: Just to confirm, since we want to have two commits, e.g., v1.58.3 and core: update re-exports for 1.58.3:
We can remove this since it's already part of lerna.json, right?
Though, maybe it still makes sense to keep `--no-push` explicitly to avoid accidentally ending up with a single pushed commit/pushed commits at all. Wdyt?

sgraband: It should be covered by the lerna.json yes. I am not sure, but i would say, that we should trust that the lerna config works, because if it does not something else is broken, and it should be noticed during the review anyway.
WDYT?

jfaltermeier: Fine for me :+1:

jfaltermeier (approved):

PR #14982, opened 2025-02-19
Make context element mandatory when showing a context menu.

### LINKED ISSUE #14960: Terminal Context Menu on wrong Window
### Bug Description:
When moving a terminal to a secondary window and doing a right-click inside the terminal, the context menu opens on the main window instead of the secondary windows. 

### Steps to Reproduce:

1. Open a terminal
2. Extract the terminal to a secondary window
3. Right-click inside the terminal
4. Observe: the context menu opens in a position relative to the main window, not at the mouse location.

### Additional Information

- Operating System:  Windows
- Theia Version: 1.58.100

tsmaeder:
#### What it does
This PR makes the context menu argument when showing a context menu mandatory. This is necessary to attach the context menu to the correct window when extracting widgets to secondary window. If now context element is available at a call site, clients can just pass `window.document.body` as the context.

Fixes #14960

#### How to test
Check that context menus everywhere are shown in the correct location and have the correct content for the location of the menu. This also includes "more" menus on toolbars, etc. (i.e. all menus other than the main menu bar. Where available, check that the menus work in secondary windows.

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/context-menu-renderer.ts
+++ b/packages/core/src/browser/context-menu-renderer.ts
@@ -110,10 +110,10 @@ export interface RenderContextMenuOptions {
      */
     includeAnchorArg?: boolean;
     /**
-     * A DOM context to use when evaluating any `when` clauses
-     * of menu items registered for this item.
+     * A DOM context for the menu to be shown
+     * Will be used to attach the menu to a window and to evaluate enablement ("when"-clauses)
      */
-    context?: HTMLElement;
+    context: HTMLElement;
     contextKeyService?: ContextMatcher;
     onHide?: () => void;
     /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ## 1.59.0
 
 <a name="breaking_changes_1.59.0">[Breaking Changes:](#breaking_changes_1.59.0)</a>
-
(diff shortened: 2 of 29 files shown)


packages/core/src/browser/context-menu-renderer.ts
```diff
@@ -110,10 +110,10 @@ export interface RenderContextMenuOptions {
      */
     includeAnchorArg?: boolean;
     /**
-     * A DOM context to use when evaluating any `when` clauses
-     * of menu items registered for this item.
+     * A DOM context for the menu to be shown
+     * Will be used to attach the menu to a window and to evaluate enablement ("when"-clauses)
      */
-    context?: HTMLElement;
```
jonah-iden, line 116: If `window.document.body` is a valid working context, why not just keep this element optional and make `window.document.body` the default value?

sdirix: If it's optional it's easy to forget and the secondary window usage could break. The caller should always indicate where they want the context menu to appear and then only fallback to `document.body` if they don't have a better choice

jonah-iden: ok thanks for the explanation. Makes sense to me

sdirix (approved):
works for me! thanks!

Issue #13798, opened 2024-06-11
Theia app is not loading while minimized at startup


rschnekenbu:
When minimizing the window of a Theia based electron application at startup, the loading of the application seems suspended until the window is restored. It can be expected here that the window is still loading despite being minimized.

tsmaeder:
I tried to reproduce this, but could not. Is there a particular sequence of actions to make it happen?

rschnekenbu:
I could reproduce on my side on Linux, using Theia IDE. I start the application from the command line.
In some conditions, I get the splash screen and the main window showing at the same time. If I minimize the main window, the log in the console is stopped, and there won't be anything showing in the main window. The splash screen remains on the screen.

In some other conditions, I don't have the main window showing, only splash screen is displayed. But I can't figure out yet what are these conditions.

rschnekenbu:
Sample video from stalled start 
https://github.com/eclipse-theia/theia/assets/3964263/a362a6f9-da51-4cf3-8ae5-2612ddde45b4

tsmaeder:
> it only actually loads the frontend when I switch to the browser tab.

Yes, but it _does_ load it, so why wouldn't it in the electron case? On my side, I can't reproduce this on Windows, even if I turn off the splash screen configuration.

tsmaeder:
@rschnekenbu "this video can't be played because the file is corrupt" 🤷

rschnekenbu:
@tsmaeder, I can open the video on chrome (and edge), but not from firefox. On firefox, I get the same corrupted file error :shrug:

tsmaeder:
Google reveals it's well known compatibility problems with ffmpeg encoding and Firefox. Why can't we just all get along?

tsmaeder:
@rschnekenbu while I can use chrome for this instance, it would be nice if you fixed your setup so folks are not forced to use Chrome.
Also: If I look at the video, the UI seems to come up alright at the end. What's the problem?

rschnekenbu:
Definitely, I will check for other software options for recording video, convert the video or switch to another format. I found this issue about Kazam mp4 encoding: https://support.mozilla.org/mk/questions/1379414. 

For the video, until I restore the window, the app is not loading. The log also does not show any progress. As soon as it is displayed again, the loading process seems active, new events are logged, etc. 
[...]

PR #15005, opened 2025-02-21
Ensure correct node expansion with short trailing sections

colin-grant-work:
#### What it does

Fixes a bug where clicking on a preference at the end of the list with few items would result in the _previous_ section appearing expanded in the list.

I also switched the default section for preferences to `features`. The older default `extensions` was mainly a workaround to avoid implementing a new, plugin-aware preference service in the plugin system. Since a method has been added specifically for registering preferences from plugins, it no longer makes sense as the default. I'm happy to discuss the implications of the change, because it could affect downstream applications and where users expect to find preferences.

> NB: Only the second commit is the fix to the scroll bug. The first commit is optional, and motivated by the considerations above.

#### How to test

1. Start the application without many extensions (e.g. no user extensions, and move the development `plugins` folder out of the repo temporarily)
2. Open the preferences view and click on the bottom 'Extensions' category.
3. That section should remain expanded and highlighted, even though the first item appearing in the UI may not belong to it.
> If you apply only the first commit from this PR and try the same thing, the previous section, AI Features, will be expanded (unless you're on a very, very short monitor).

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


packages/preferences/src/browser/views/preference-editor-widget.ts
```diff
@@ -314,18 +317,26 @@ export class PreferencesEditorWidget extends BaseWidget implements StatefulWidge
...
             }
         }
     }
 
+    /** Ensures that we don't set the model's selection while attempting to scroll in reaction to a model selection change. */
+    protected scrollWithoutModelUpdate(node?: HTMLElement): void {
+        if (!node) { return; }
+        this.shouldUpdateModelSelection = false;
+        node.scrollIntoView();
+        requestAnimationFrame(() => this.shouldUpdateModelSelection = true);
```
tsmaeder, line 337: Two comments here: Why are we delaying this here? If it's for the code to execute after some promises we create in the meantime, that's not reliable because execution order of promises vs requestAnimationFrame() is not guarantueed.
Second: are we sure this code is not invoked before the Window is visible? Because otherwise, this might delay loading, see https://github.com/eclipse-theia/theia/issues/13798

colin-grant-work: It appears that `setTimeout` does not reliably push the reset past the running of the handler. This particular `requestAnimationFrame` currently doesn't run at startup, as we don't restore model selection independently.

packages/preferences/src/browser/util/preference-tree-generator.ts
```diff
@@ -148,7 +148,7 @@ export class PreferenceTreeGenerator {
         if (!property.owner) {
             return;
         }
-        const groupID = this.defaultTopLevelCategory;
+        const groupID = 'extensions';
```
martin-fleck-at, line 151: I'm not really sure what this change entails and if someone would ever change the `defaultTopLevelCategory` but if someone this, this change makes it a bit harder for them to get their previous behavior back. Would it at least make sense to add this as an optional argument with default value to the method signature, so that it can be more easily overwritten?

packages/preferences/src/browser/util/preference-tree-generator.ts
```diff
@@ -43,7 +43,7 @@ export class PreferenceTreeGenerator {
 
     protected readonly onSchemaChangedEmitter = new Emitter<CompositeTreeNode>();
     readonly onSchemaChanged = this.onSchemaChangedEmitter.event;
-    protected readonly defaultTopLevelCategory = 'extensions';
+    protected readonly defaultTopLevelCategory = 'features';
```
martin-fleck-at, line 46: I'm not really sure what impact this change has, could you elaborate just a bit more. As far as I can tell, it really is just a visual change that moves some of the preferences from 'Extension' to 'Features'. I'm not sure if this is more in line with what VS Code does but to me it makes sense. However, I want to make sure I'm not missing anything.

martin-fleck-at (commented):
Overall the changes look good to me and seem to work well, thank you very much, Colin! I do have some open questions but they may turn out to be moot.

colin-grant-work:
> Do I understand that correctly that if we were to run this at startup, we would run into a problem or is it just delaying the visibility of the window?

Just delaying the startup behavior, in this case, I think only of this widget, as nothing depends on the widget having scrolled (or not) anywhere in particular.

Re: your other comments, they have to do with the changes in the preference tree generator, which are part of the first commit and can be removed. I believe they're 'correct,' but they can be taken up separately from the fix for the scrolling issue.

martin-fleck-at (approved):
Alright, thank you very much, Colin! I'm happy with the changes then but would like to give @tsmaeder a chance to chime in as he raised the timing issues in the first place.

tsmaeder:
@martin-fleck-at I just randomly came across this PR, saw the "requestAnimationFrame()" and given recent discussions wanted to makes sure we're not calling the code from our startup sequence (for aforementioned reasons). 
As for the change in general: in the preferences widget, we're updating the selection upon changing scroll state and we're updating the scroll state based on the selection. That is tough to get right. 
At the root. the problem is this: we cannot distinguish between changes due to user interaction and changes due to programmatic changes. The PR tries to "cut the loop" with a state variable. The asynchronous nature of DOM updates requires us to reset the variable "later". I believe the PR improves the behaviour and have no objections to merging it. However, similar problems with scroll tracking also plague our tree implementation and I firmly believe we'll have to find a clean solution for it.

Issue #12452, opened 2023-04-24
Missing menus in CommentSample


rschnekenbu:
Some buttons generated by commands are missing in the comment sample since Theia release 1.28. This can be observed when the menu refactoring #11290 was applied. Before that refactoring, for example on theia 1.27, the buttons were displayed without issues and working. 

This was discovered while working on PR #12007 and the issue #12231.    
From what I could observe:
- Before refactoring, the buttons could be seen on the comment sample editor. The process to get the dialog is documented in this [CommentThread state PR](https://github.com/eclipsesource/theia/pull/77)

![theia-1 27-comment](https://user-images.githubusercontent.com/3964263/234121266-a74d6938-002a-4956-af70-c7a89873cedc.png)

- After refactoring, e.g. on Theia 1.28+, almost no buttons can be seen using same extension.

![theia-master-comment](https://user-images.githubusercontent.com/3964263/234121614-0bc68f4c-5587-4e8d-bec2-d7ac1c9480af.png)

- The buttons are added back when the `when` and `group` entries are removed from the package.json of the extension (see https://github.com/microsoft/vscode-extension-samples/blob/4bef4529beb758da66d00f0ea708730267fd64fe/comment-sample/package.json#L104-L123). This may be an hint on the underlying issue. When the buttons are visible as the `when` and `group` entries are removed, the command can still not run. I could not trace what the main cause is for this failure.

rschnekenbu:
On 1.58, the UI is also broken in the way that the comment editor is too small, the button area is not visible. Navigation to the buttons is sometimes required, using Tab key for example.

PR #15015, opened 2025-02-24
[vscode] introduction of the CommentingRange type

### LINKED ISSUE #14941: [vscode] introduction of the CommentingRange type
`CommentingRanges` have been introduced; with some changes to `commentThread` and `CommentingRangeProvider`

rschnekenbu:
#### What it does

Introduce the support for Commenting Ranges and capability to create comments on file rather than on a specific range.

also it fixes the creation of comments, with arguments passed to the command being the one expected by argument processor

fix #14941

It is interesting to note that commenting range is supported, but the API do not let the extensions create Thread with not range (see https://github.com/microsoft/vscode/blob/c5549bc00e5f2e17773e9cffa694996fff318772/src/vscode-dts/vscode.d.ts#L17413). So this is hard to test and debug, even on vscode itself... For now, theia code support the ranges to be undefined, but it is not possible to test at runtime. Weird!

#### How to test

Steps to test:
1. Install this extension to test the feature on comparing markdown or typescript files (commenting ruler is available on Theia on diff comparator):
- src:[comment-sample-0.0.4-src.zip](https://github.com/user-attachments/files/18940465/comment-sample-0.0.4-src.zip)
- zip vsix: [vscode-comment-api-example-0.0.4.zip](https://github.com/user-attachments/files/18940456/vscode-comment-api-example-0.0.4.zip)
Using this extension, you get 2 comment providers: 
- the markdown (*.md files) comment controller provide 10 lines range every 20 lines and also allow file comments, 
- the typescript (*.ts files) allow commenting on every line. 

2. Open the comparison between 2 files (for example, 2 typescript files or 2 markdown files using **_compare with each other_** menu

3. You can see in the diff editor the comment gutter (grey bar) on the side of the text editors, you can create a comment while clicking on the gutter. Notice that the md files have a gutter that is starting every 20 lines for a length of 10 lines. ts files have a gutter for the whole file length.

comment-sample-extension-ranges

The UI seems currently broken on comments, I commented on a followup issue for comments action and UI: https://github.com/eclipse-theia/theia/issues/12452.  (css are bad, actions do no support the 'when' condition, etc.). But this is not relevant for this API implementation

#### Follow-ups

Improvements on Comment support: 

Improvements on buttons and UI: #12452

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
+++ b/packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
@@ -73,7 +73,7 @@ export class CommentThreadWidget extends BaseWidget {
         this.toDispose.push(this.commentGlyphWidget = new CommentGlyphWidget(editor));
         this.toDispose.push(this._commentThread.onDidChangeCollapsibleState(state => {
             if (state === CommentThreadCollapsibleState.Expanded && !this.isExpanded) {
-                const lineNumber = this._commentThread.range.startLineNumber;
+                const lineNumber = this._commentThread.range?.startLineNumber ?? 0;
 
                 this.display({ afterLineNumber: lineNumber, afterColumn: 1, heightInLines: 2 });
                 return;
@@ -256,8 +256,8 @@ export class CommentThreadWidget extends BaseWidget {
         const frameThickness = Math.round(lineHeight / 9) * 2;
         const body = this.zoneWidget.containerNode.getElementsByClassName('body')[0];
 
-        const computedLinesNumber = Math.ceil((headHeight + body.clientHeight + arrowHeight + frameThickness + 8 /** margin bottom to avoid margin collapse */) / lineHeight);
-        this.zoneWidget.show({ afterLineNumber: this._commentThread.range.startLineNumber, heightInLines: computedLinesNumber });
+        const computedLinesNumber = Math.ceil((headHeight + body?.clientHeight + arrowHeight + frameThickness + 8 /** margin bottom to avoid margin collapse */) / lineHeight);
+        this.zoneWidget.show({ afterLineNumber: this._commentThread.range?.startLineNumber ?? 0, heightInLines: computedLinesNumber });
     }
 
     protected render(): void {
@@ -588,7 +588,8 @@ export class CommentEditContainer extends React.Component<CommentEditContainer.P
                 {menus.getMenu(COMMENT_CONTEXT).children.map((node, index) => {
                     const onClick = () => {
                         commands.executeCommand(node.id, {
-                            thread: commentThread,
+                            commentControlHandle: commentThread.controllerHandle,
+                            commentThreadHandle: commentThread.commentThreadHandle,
                             commentUniqueId: comment.uniqueIdInThread,
                             text: this.inputRef.current ? this.inputRef.current.value : ''
                         });
@@ -623,7 +624,8 @@ export class CommentsInlineAction extends React.Component<CommentsInlineAction.P
                 title={node.label}
                 onClick={() => {
                     commands.executeCommand(node.id, {
-                        thread: commentThread,
+                        commentControlHandle: commentThread.controllerHandle,
+                        commentThreadHandle: commentThread.commentThreadHandle,
                         commentUniqueId
                     });
                 }} />
@@ -651,8 +653,10 @@ export class CommentActions extends React.Component<CommentActions.Props> {
                     commands={commands}
                     node={node}
                     onClick={() => {
+                        console.log('CommentActions.onClick: ' + commentThread);
                         commands.executeCommand(node.id, {
-                            thread: commentThread,
+                            commentControlHandle: commentThread.controllerHandle,
+                            commentThreadHandle: commentThread.commentThreadHandle,
                             text: getInput()
                         });
                         clearInput();
--- a/packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
+++ b/packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
@@ -256,7 +256,8 @@ export class CommentThreadWidget extends BaseWidget {
         const frameThickness = Math.round(lineHeight / 9) * 2;
         const body = this.zoneWidget.containerNode.getElementsByClassName('body')[0];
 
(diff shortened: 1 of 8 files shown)


tsmaeder:
@rschnekenbu not sure how to test this: is the commenting UI available in Theia before this PR? It's not clear to me how I can invoke the commenting functionality from the video 🤷 Side note: is this an animated gif? If so, that is very inconvenient for Firefox users, because I don't get the video controls: I have to watch it from the start each time I want to go back.

rschnekenbu:
@tsmaeder , the UI is existing (and was not working) prior to this PR. To test it, you have to compare 2 different files (in my case, a markdown file and a copy of it) using the compare with each other menu when the 2 are selected. In this case, the comment gutter (a grey bar) is shown in the Diff Editor.
The area covered by the gutter is the one given by the comment providers in the extension. In the case of MD files, 10 lines can be commented, 10 not, and the same again. For typescript, any line can be commented.
For the gif, that is the most portable solution over the OS and the browsers but indeed with no controls.

tsmaeder:
I'm not seeing the "comment gutter" 🤷 
image

rschnekenbu:
Do you have any error in the log when the app is supposed to activate? 
When you see it and your mouse is on top of the gutter, a [+] grey icon is showing on top of the gutter.

tsmaeder:
Now I did a "Compare with each other" again. No error in the log this time, but still not gutter.

rschnekenbu:
I will test again on my side, but I never see such an issue. I do not get why the contextKeyService shall be disposed.

rschnekenbu:
@tsmaeder  I added some steps to clarify the testing.

tsmaeder:
What I was missing is that you need to install the extension in order to make the UI show up 🤷

tsmaeder:
@rschnekenbu I get the "+" signs in the gutter as well on the left side of the left editor in the diff for removed sections. However, when I click on the plus, nothing happens (no error log either).

packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
```diff
@@ -256,8 +256,8 @@ export class CommentThreadWidget extends BaseWidget {
         const frameThickness = Math.round(lineHeight / 9) * 2;
         const body = this.zoneWidget.containerNode.getElementsByClassName('body')[0];
 
-        const computedLinesNumber = Math.ceil((headHeight + body.clientHeight + arrowHeight + frameThickness + 8 /** margin bottom to avoid margin collapse */) / lineHeight);
-        this.zoneWidget.show({ afterLineNumber: this._commentThread.range.startLineNumber, heightInLines: computedLinesNumber });
+        const computedLinesNumber = Math.ceil((headHeight + body?.clientHeight + arrowHeight + frameThickness + 8 /** margin bottom to avoid margin collapse */) / lineHeight);
```
tsmaeder, line 259: Probably needs body?.clientHeight ?? 0, no?

packages/plugin-ext/src/plugin/comments.ts
```diff
@@ -172,11 +177,30 @@ export class CommentsExtImpl implements CommentsExt {
 
         const documentData = this._documents.getDocumentData(URI.revive(uriComponents));
         if (documentData) {
-            const ranges: theia.Range[] | undefined | null = await commentController.commentingRangeProvider!.provideCommentingRanges(documentData.document, token);
-            if (ranges) {
-                return ranges.map(x => fromRange(x));
+            const commentingRanges:
+                | theia.Range[]
```
tsmaeder, line 181: Can we not let type inference determine the type here?

packages/plugin-ext/src/plugin/comments.ts
```diff
@@ -172,11 +177,30 @@ export class CommentsExtImpl implements CommentsExt {
...
+                    ranges: commentingRanges.map(x => fromRange(x)),
+                    fileComments: false
+                };
+            } else if (commentingRanges) {
+                return {
+                    ranges: commentingRanges.ranges?.map(x => fromRange(x)) || [],
+                    fileComments: commentingRanges.enableFileComments
+                };
+            } else {
+                return commentingRanges ?? undefined;
```
tsmaeder, line 200: We already now that `!commentingRanges` here, why not just return `undefined`?

rschnekenbu: indeed, being null or undefined, we don't care as we want undefined only. Removed in next commit.

packages/plugin-ext/src/main/browser/comments/comment-thread-widget.tsx
rschnekenbu: changed in next commit

packages/plugin-ext/src/plugin/comments.ts
```diff
@@ -221,7 +245,8 @@ export class ExtHostCommentThread implements theia.CommentThread, theia.Disposab
     readonly onDidUpdateCommentThread = this._onDidUpdateCommentThread.event;
 
     set range(range: theia.Range) {
-        if (!range.isEqual(this._range)) {
+        if ( ((range === undefined) !== (this._range === undefined))
```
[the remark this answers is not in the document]
rschnekenbu: This is now possible to change to an undefined range. Evolution should come later with the ability to create comments without range from the API.

rschnekenbu:
@tsmaeder  could you finally get the comment dialog to open ?

tsmaeder:
> @rschnekenbu I get the "+" signs in the gutter as well on the left side of the left editor in the diff for removed sections. However, when I click on the plus, nothing happens (no error log either).

@rschnekenbu ping?

rschnekenbu:
@tsmaeder, indeed, the left editor can't have comments. That does not make sense for the set of changes I have introduced, probably some other defect existing prior to this change. But in any case, I will have a look and check why left editor does not support commenting.

tsmaeder:
If it's not a regression, I don't have a problem merging this one, but I can't really test that because the plugin does not work without the PR.

rschnekenbu:
The onMouseDownEvent in [comments-contribution.ts](https://github.com/eclipse-theia/theia/blob/ec91680d8dcfb7b84f457084a103689919c03c41/packages/plugin-ext/src/main/browser/comments/comments-contribution.ts#L127) is not send when clicking on the left gutter. This seems an issue with the current implementation of the comments contribution and not from the API evolution.

tsmaeder (approved):

PR #15018, opened 2025-02-24
fix(chat): improve variable autocompletion

planger:
#### What it does

Improve auto-completion in the AI chat in the following ways:
* Append ':' to variable completions if they have a mandatory argument; this also avoids the editor from submitting if the user already entered the full text of the variable and hits enter, even though arguments need to be provided
* Fix manually triggered auto-completion for variables, functions, etc. as we didn't properly test for already typed characters after the trigger character before; this lead to no suggestions being given if the user typed ~a then escape to close auto-completion and then retriggering it via Ctrl-space

#### How to test

Play around with the auto-completion, completing variables, functions, arguments, etc.

#### Follow-ups

None

#### Breaking changes

#### Attribution

None

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
+++ b/packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
@@ -71,17 +71,15 @@ export class ChatViewLanguageContribution implements FrontendApplicationContribu
     }
 
     getCompletionRange(model: monaco.editor.ITextModel, position: monaco.Position, triggerCharacter: string): monaco.Range | undefined {
-        // Check if the character before the current position is the trigger character
+        const wordInfo = model.getWordUntilPosition(position);
         const lineContent = model.getLineContent(position.lineNumber);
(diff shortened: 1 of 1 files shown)


JonasHelming (approved):
This works very nicely! One question: Is it inteded that you get a different behavior:

- if you type "#file:" you get the list of file in place
- If you type "#file" => enter => the file picker is opened
?

planger:
Yes, kind of. If we complete the variable, the argument picker is invoked, similar to if you select the + and then select the respective variable.
If you just type ':' it is the normal completion kicking in.

Whether this is good is debatable. I modeled this after how it is in VS Code.

JonasHelming:
> Yes, kind of. If we complete the variable, the argument picker is invoked, similar to if you select the + and then select the respective variable. If you just type ':' it is the normal completion kicking in.
> 
> Whether this is good is debatable. I modeled this after how it is in VS Code.

I am fine with this! Let's see what users think, thank you!

PR #15025, opened 2025-02-24
Add resource management to MCP services

eneufeld:
#### What it does

- Implemented resource registration and management in MCPFrontendService.
- Expanded MCPServer and MCPServerManager interfaces to handle resources.
- Updated MCPServerManagerImpl to support resource retrieval.
- Included resource handling capabilities in MCPServer.
#### How to test

1. add the everything mcp server to the list of mcp servers
```
"everything": {
      "command": "npx",
      "args": [
        "-y",
        "@modelcontextprotocol/server-everything"
      ]
    }
```
2. start the server
3. in the chat see that there are new example variables: Resource_1, Resource_2 etc
4. use them in a request and check in the Output view that the variable was resolved

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-mcp/src/browser/mcp-frontend-service.ts
+++ b/packages/ai-mcp/src/browser/mcp-frontend-service.ts
@@ -15,7 +15,8 @@
 // *****************************************************************************
 import { injectable, inject } from '@theia/core/shared/inversify';
 import { MCPServer, MCPServerManager } from '../common/mcp-server-manager';
-import { ToolInvocationRegistry, ToolRequest } from '@theia/ai-core';
+import { AIVariableService, ToolInvocationRegistry, ToolRequest } from '@theia/ai-core';
+import { DisposableCollection } from '@theia/core';
 
 @injectable()
 export class MCPFrontendService {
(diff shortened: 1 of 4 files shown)


JonasHelming:
Two questions are open to me:

1. Variable definition
Is it really a good idea to make all MCP resources root variables. This might be many. Would it make more sense to define a variable for mcp resources with an argument like we do for files OR even a variable per mcp server (also with an argument). This way it would be more transparent for the user where these resources are coming from and easier to find them. This is particularly the case for the IDE use case, where users configure their MCP servers. In a custom tool, the tool builder might decide to handle this differently, so I believe this part should be adaptable (e.g. via DI)
2.  Context
Should all MCP resources be standard variables or contetx variables. Or should be allow both options, similar to the discussion we have here: https://github.com/eclipse-theia/theia/pull/15196#pullrequestreview-2688333937

@planger FYI

planger:
I'd agree that we should probably group MCP-contributed variables in a prefix. Imho this doesn't have to be `mcp` but can be the identifier of the `mcp` server, but it should not be a root variable, also to avoid naming-conflicts and to add clarity what the context of this variable is. As the `mcp` prefix is also used for tool functions, I guess we should be consistent with this naming scheme.

Regarding variable vs context variable... difficult to say. I guess this depends on the MCP server. I'm not sure if the majority of MCP servers already evolved a common practice on how to use resources for which types of values, or if they ever would consolidate reliably. So the option of having both seems compelling. However, in a first step, I'd add them as normal variables and observe the ecosystem of MCP servers until we have more clarity.

JonasHelming:
1. OK, the go with the same prefix as for the functions (I would even reuse the code then) e.g. "mcp_git:resourcename"
2. Fine, let do it as a regular variable first, but lets create a follow up for the context variable option.

ndoschek:
Hi @eneufeld, just checking in on this PR. The backend parts (resource methods on MCPServer/MCPServerManager) have since been implemented in #15842 afaict.
What's still missing is exposing MCP resources as chat variables on the frontend side. Are you still planning to continue with that part?

eneufeld:
Not really if the support for resources was implemented, let's close this.

Issue #14899, opened 2025-02-11
[Theia AI] Support prompt fragments


JonasHelming:
Prompt fragments allow to define parts of prompts in a reusable way. You can embed these prompts then via a variable in the chat or in our prompt templates. This also allows to have a set of prompts available in the chat without always defining a custom agent.

Step 1: Allow referencing prompt fragments

The PromptCustomizationService should already load any file in the prompt service. When we add a file there which does not have the id of an agent, we assume it is a resuable prompt.
We now need a variable "#prompt:${promptId}" that resolves to the prompt with the specified id. We need to check the resolvement logic in the prompt service so that variables and function occuring in injected prompt fragments are resolved , too. same applies for chat messages.
With this, the user can add a prompt fragment by just placing a file e.g. myprompt.prompttemplate in the prompt templates directory and then use it via "#prompt:myprompt". "myPrompt" gets auto completed.

Step 2: Add UI

Add a new tab to the AI configuration view showing a list of all loaded prompt templates allowing to:
- Edit
- Add Variant (of an existing one)
- Add prompt fragment
- Reset/delete (if they are resettable, e.g. they have been created by an agent, otherwise delete)
- Filter to agent prompts or prompt fragments
- Open the directory
- Sort

sdirix:
[VS Code also introduced prompt management and reusable prompts](https://code.visualstudio.com/docs/copilot/copilot-customization#_reusable-prompt-files-experimental). We should evaluate whether we want to align Theia's prompt management.

JonasHelming:
I had a look. It is pretty similar, but I like our UI integration a lot better for now.

JonasHelming:
We could allow to place prompt variables in the context, though so they get part of the system prompt. @planger WDYT?

planger:
The VS Code concept for prompt templates is pretty similar indeed. I only see two real differences:
* They only allow attaching prompt files to the context, while we foresee to reference them directly in the prompt via a variable, which would then be replaced with the actual prompt text.
* They allow making prompt templates part of the workspace (`.github/prompts`), which imho is a nice addition to simplify sharing prompt templates.

TL;DR:
* I'm in favor of keeping our prompt variables as they are for now (so users can put them in their user prompt to be replaced with the prompt template text). This gives them more control where exactly to put the prompt template content, if they want to.
* I second the idea of marking prompt template variables as context variables, so they could also just be referenced in the context (is almost for free -- see below, and may enable some extra use cases)
[...]

PR #15026, opened 2025-02-25
Add variable completion for {{}} syntax in prompttemplate

eneufeld:
#### What it does

Add variable autocomplete support for `{{}}` and `{{{}}}` syntax in prompttemplate

fixes: #15202

#### How to test

Open the prompt template editor, eg for custom editors and try to use a variable. Using `#{` you get an autocompletion.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/browser/prompttemplate-contribution.ts
+++ b/packages/ai-core/src/browser/prompttemplate-contribution.ts
@@ -110,6 +110,11 @@ export class PromptTemplateContribution implements LanguageGrammarDefinitionCont
             triggerCharacters: ['{'],
             provideCompletionItems: (model, position, _context, _token): ProviderResult<monaco.languages.CompletionList> => this.provideVariableCompletions(model, position),
         });
+        monaco.languages.registerCompletionItemProvider(PROMPT_TEMPLATE_LANGUAGE_ID, {
+            // Monaco only supports single character trigger characters
+            triggerCharacters: ['{', ':'],
+            provideCompletionItems: (model, position, _context, _token): ProviderResult<monaco.languages.CompletionList> => this.provideVariableWithArgCompletions(model, position),
+        });
 
         const textmateGrammar = require('../../data/prompttemplate.tmLanguage.json');
         const grammarDefinitionProvider: GrammarDefinitionProvider = {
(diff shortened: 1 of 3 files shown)


eneufeld:
I retested and the auto-completion works but the variables are not resolved. Furthermore this does not work in the default prompt for custom agents in the yml.
We need to align this with https://github.com/eclipse-theia/theia/issues/14899 and finally also decide whether we will show ids or names for functions, variables and agents

packages/ai-core/src/browser/prompttemplate-contribution.ts
```diff
@@ -122,6 +135,19 @@ export class PromptTemplateContribution implements LanguageGrammarDefinitionCont
...
     }
 
+    provideVariableCompletions(model: monaco.editor.ITextModel, position: monaco.Position): ProviderResult<monaco.languages.CompletionList> {
+        return this.getSuggestions(
+            model,
+            position,
+            '#{',
+            this.variableService.getVariables(),
+            monaco.languages.CompletionItemKind.Variable,
+            variable => variable.id,
```
planger, line 145:
```suggestion
            variable => variable.args?.some(arg => !arg.isOptional) ? variable.name + PromptText.VARIABLE_SEPARATOR_CHAR : variable.name,

```

In [`chat-view-language-contribution.ts`](https://github.com/eclipsesource/theia/blob/f1562a2187c0cb644d851352ddcc119aed31c86f/packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts#L137-L138) we actually complete to the `variable.name` and not the id. I think we should make this consistent and use `name` here too.

Even if we don't provide a completion for arguments here (which might be a separate thing to do at some point), it might be worth to still complete the `:` if it has mandatory arguments, to at least indicate that this variable requires an argument.

planger (commented):
Thank you Eugen, looks good to me! Just one inline comment.

JonasHelming:
@planger I think the first essential question here is whether we want to support #{ as a syntax for variables in prompts?

planger:
> @planger I think the first essential question here is whether we want to support #{ as a syntax for variables in prompts?

You mean in user prompts (entered into the chat input)?

Or do you mean why we use #{ instead of just { in prompt templates?

I didn't want to open up this topic on this PR, as it goes beyond the scope, but I was rather wondering if it is still a good idea to have a different syntax between prompt templates and user prompts. It doesn't seem to add value but prevents us from sharing completion logic and highlighting as is. So we might consider harmonizing the syntax. In that case though, I'd stick with the syntax we use in the chat input. The only extra syntax in prompt templates would be {{ }} for agent-specific variables.

If we don't harmonize the user prompt and prompt template syntax, I don't have a strong opinion whether we use #{ or just {{ for variables.

planger (approved):

planger:
https://github.com/eclipse-theia/theia/issues/15202 also mentions variable argument auto-completion. So we should probably leave https://github.com/eclipse-theia/theia/issues/15202 open or create a follow-up to add variable argument auto-completion, as you prefer. Just mentioning this here, because merging this PR will close #15202.

eneufeld:
@planger I'll take a look as part of this.

eneufeld:
@planger  please take a second look, I'm not 100% happy. But with the syntax using `{{` we run into quite some issues we don't have in the chat input.
I suggest to clean everything up as part of the alignment of the chatinput and prompttemplate syntaxes

planger (changes requested):
Thanks for the revision!

The suggestions work fine now, however, I think there is something wrong with the insertion range. For me it duplicates the `file:` in the inserted text. Also the suggestions don't seem to update when I enter text after `file:`, it'll only suggest the recent files. Only if I press Esc and then Ctrl-space again, it'll bring up the suggestions that match the current text.

eneufeld:
@planger i can't reproduce the issue with the duplication of the `file:`, maybe I do something different

For the suggested files, this is the same behavior as in the chat input. we basically collect all possible arguments for the variable and then the filter needs to work, and it doesn't do a good job.

planger:
> @planger i can't reproduce the issue with the duplication of the `file:`, maybe I do something different

https://github.com/user-attachments/assets/97d496aa-98c7-458f-8afe-22a9a25c8c81

> For the suggested files, this is the same behavior as in the chat input. we basically collect all possible arguments for the variable and then the filter needs to work, and it doesn't do a good job.

Yes, you are right, this isn't great in either of the two use cases. Just wanted to mention it that it isn't great.

eneufeld:
@planger same happens in the chat input.
should we open a follow up?

planger:
I don't see this happening in the chat input on `master`.

https://github.com/user-attachments/assets/e0c149f8-5e98-4bd0-bbf7-081b45e5790a

eneufeld:
> I don't see this happening in the chat input on `master`.
> 
>  chat-input-auto-complete-working-fine.mp4

good catch, a missing `()` was the culprit

planger (approved):
Thanks for the fix, looks good to me! :+1:

PR #15030, opened 2025-02-25
Fix: Prevent autocompletion from triggering incorrectly within words

### LINKED ISSUE #15028: [Theia AI] Hashtags always autocomplete in the chat
They should only autocomplete when it’s the first character of the line or if a space precedes them, e.g. I can’t type “foo#bar” without triggering autocomplete on variables.

eneufeld:
#### What it does

This commit fixes an issue where autocompletion would incorrectly trigger when typing a variable character '#' in the middle of words (e.g., "Hello#").

Changes:
- Removed '#' from trigger characters in the variable argument completion provider
- Improved boundary detection in the variable argument picker to prevent triggering when '#' appears within words
- Added better context validation to ensure completion only happens in appropriate cases
- Fixed the implementation to maintain correct functionality for legitimate variable references like "#file:"
- Refactored code to remove code duplication

The fix ensures that autocompletion is only triggered in intended contexts while preserving all existing functionality. Fixed #15028

#### How to test

In the chat Input view test different cases:
- starting with `~`, `@` or `#` should trigger the autocompletion
- continuing after a variable selection via `#file` with a `:` should trigger the argument selection
- typing `Foo~`, `Foo@` or `Foo#` should not trigger the autocompletion
- typing `Foo ~`, `Foo @` or `Foo #` should trigger the autocompletion

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
+++ b/packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
@@ -92,7 +92,7 @@ export class ChatViewLanguageContribution implements FrontendApplicationContribu
             triggerCharacter: PromptText.FUNCTION_CHAR,
             getItems: () => this.toolInvocationRegistry.getAllFunctions(),
             kind: monaco.languages.CompletionItemKind.Function,
-            getId: tool => `{tool.id} `,
+            getId: tool => `${tool.id} `,
             getName: tool => tool.name,
             getDescription: tool => tool.description ?? ''
         });
@@ -119,15 +119,22 @@ export class ChatViewLanguageContribution implements FrontendApplicationContribu
         // one to the left, and -1 for 0-based index
         const characterBeforeCurrentWord = lineContent[wordInfo.startColumn - 1 - 1];
 
-        // Check if we're in the middle of a word, not at the beginning
-        if (wordInfo.startColumn > 1 && characterBeforeCurrentWord !== triggerCharacter) {
+        if (characterBeforeCurrentWord !== triggerCharacter) {
             return undefined;
         }
 
-        // Check if the trigger character is actually part of another word
-        const wordAtTriggerPosition = model.getWordAtPosition({ lineNumber: position.lineNumber, column: wordInfo.startColumn - 1 });
-        if (wordAtTriggerPosition && wordAtTriggerPosition.word.length > 1) {
-            return undefined;
+        // we are not at the beginning of the line
+        if (wordInfo.startColumn > 2) {
+            const charBeforeTrigger = model.getValueInRange({
+                startLineNumber: position.lineNumber,
+                startColumn: wordInfo.startColumn - 2,
+                endLineNumber: position.lineNumber,
+                endColumn: wordInfo.startColumn - 1
+            });
+            // If the character before the trigger is not whitespace, don't provide completions
+            if (!/\s/.test(charBeforeTrigger)) {
+                return undefined;
+            }
         }
 
         return new monaco.Range(
@@ -222,32 +229,15 @@ export class ChatViewLanguageContribution implements FrontendApplicationContribu
             return;
         }
 
-        // Get the current line's content and the word at cursor
-        const lineContent = model.getLineContent(position.lineNumber);
+        // // Get the word at cursor
         const wordInfo = model.getWordUntilPosition(position);
-
-        // Only trigger if current word starts with #, followed by actual text
-        if (!wordInfo.word.startsWith(PromptText.VARIABLE_CHAR) || wordInfo.word.length <= 1) {
-            return;
-        }
-
-        // Ensure the # isn't part of another word (like "Hello#")
-        // by checking if there's a word character before the # character
-        if (wordInfo.startColumn > 1) {
-            // Convert to 0-based index and account for the # at start of word
-            const indexBeforeHash = wordInfo.startColumn - 2;
-            if (indexBeforeHash >= 0 && /\w/.test(lineContent[indexBeforeHash])) {
-                // We're in a case like "Hello#variable" - don't trigger
-                return;
-            }
-        }
 
         // account for the variable separator character if present
         let endOfWordPosition = position.column;
-        let insertTextPrefix = PromptText.VARIABLE_SEPARATOR_CHAR;
-        if (this.getCharacterBeforePosition(model, position) === PromptText.VARIABLE_SEPARATOR_CHAR) {
+        if (wordInfo.word === '' && this.getCharacterBeforePosition(model, position) === PromptText.VARIABLE_SEPARATOR_CHAR) {
             endOfWordPosition = position.column - 1;
-            insertTextPrefix = '';
+        } else {
+            return;
         }
 
         const variableName = model.getWordAtPosition({ ...position, column: endOfWordPosition })?.word;
@@ -267,7 +257,7 @@ export class ChatViewLanguageContribution implements FrontendApplicationContribu
 
         inputEditor.executeEdits('variable-argument-picker', [{
             range: new monaco.Range(position.lineNumber, position.column, position.lineNumber, position.column),
-            text: insertTextPrefix + arg
+            text: arg
         }]);
 
         await this.addContextVariable(variableName, arg);
--- a/packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
+++ b/packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
@@ -31,6 +31,16 @@ const VARIABLE_RESOLUTION_CONTEXT = { context: 'chat-input-autocomplete' };
 const VARIABLE_ARGUMENT_PICKER_COMMAND = 'trigger-variable-argument-picker';
 const VARIABLE_ADD_CONTEXT_COMMAND = 'add-context-variable';
 
+interface CompletionSource<T> {
+    triggerCharacter: string;
+    getItems: () => T[];
+    kind: monaco.languages.CompletionItemKind;
+    getId: (item: T) => string;
+    getName: (item: T) => string;
+    getDescription: (item: T) => string;
+    command?: monaco.languages.Command;
+}
+
 @injectable()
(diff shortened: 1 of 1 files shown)


packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
```diff
@@ -31,6 +31,17 @@ const VARIABLE_RESOLUTION_CONTEXT = { context: 'chat-input-autocomplete' };
 const VARIABLE_ARGUMENT_PICKER_COMMAND = 'trigger-variable-argument-picker';
 const VARIABLE_ADD_CONTEXT_COMMAND = 'add-context-variable';
 
+// Define a consistent interface for completion sources
```
planger, line 34:
```suggestion
```

Imho this comment doesn't add a lot of information

packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
```diff
@@ -49,34 +60,77 @@ export class ChatViewLanguageContribution implements FrontendApplicationContribu
     onStart(_app: FrontendApplication): MaybePromise<void> {
         monaco.languages.register({ id: CHAT_VIEW_LANGUAGE_ID, extensions: [CHAT_VIEW_LANGUAGE_EXTENSION] });
 
-        monaco.languages.registerCompletionItemProvider(CHAT_VIEW_LANGUAGE_ID, {
-            triggerCharacters: [PromptText.AGENT_CHAR],
-            provideCompletionItems: (model, position, _context, _token): ProviderResult<monaco.languages.CompletionList> => this.provideAgentCompletions(model, position),
+        // Register all completion providers
```
planger, line 63:
```suggestion
```

Imho this comment doesn't add a lot of information

packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
```diff
@@ -49,34 +60,77 @@ export class ChatViewLanguageContribution implements FrontendApplicationContribu
...
+        this.registerCompletionProviders();
+
+        // Register editor commands
+        monaco.editor.registerCommand(VARIABLE_ARGUMENT_PICKER_COMMAND, this.triggerVariableArgumentPicker.bind(this));
+        monaco.editor.registerCommand(VARIABLE_ADD_CONTEXT_COMMAND, (_, ...args) => args.length > 1 ? this.addContextVariable(args[0], args[1]) : undefined);
+    }
+
+    protected registerCompletionProviders(): void {
+        // Register standard completion providers
```
planger, line 72:
```suggestion
```

Imho this comment doesn't add a lot of information

packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
```diff
@@ -49,34 +60,77 @@ export class ChatViewLanguageContribution implements FrontendApplicationContribu
...
+        this.registerStandardCompletionProvider({
+            triggerCharacter: PromptText.AGENT_CHAR,
+            getItems: () => this.agentService.getAgents(), // Use function to fetch the latest items
```
planger, line 75:
```suggestion
            getItems: () => this.agentService.getAgents(),
```

Imho this comment doesn't add a lot of information

packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
```diff
@@ -49,34 +60,77 @@ export class ChatViewLanguageContribution implements FrontendApplicationContribu
...
+            getName: agent => agent.name,
+            getDescription: agent => agent.description
         });
-        monaco.languages.registerCompletionItemProvider(CHAT_VIEW_LANGUAGE_ID, {
-            triggerCharacters: [PromptText.VARIABLE_CHAR],
-            provideCompletionItems: (model, position, _context, _token): ProviderResult<monaco.languages.CompletionList> => this.provideVariableCompletions(model, position),
+
+        this.registerStandardCompletionProvider({
+            triggerCharacter: PromptText.VARIABLE_CHAR,
+            getItems: () => this.variableService.getVariables(), // Use function to fetch the latest items
```
planger, line 84:
```suggestion
            getItems: () => this.variableService.getVariables(),
```

packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
```diff
@@ -49,34 +60,77 @@ export class ChatViewLanguageContribution implements FrontendApplicationContribu
...
+                id: VARIABLE_ARGUMENT_PICKER_COMMAND,
+            }
         });
-        monaco.languages.registerCompletionItemProvider(CHAT_VIEW_LANGUAGE_ID, {
-            triggerCharacters: [PromptText.VARIABLE_CHAR, PromptText.VARIABLE_SEPARATOR_CHAR],
-            provideCompletionItems: (model, position, _context, _token): ProviderResult<monaco.languages.CompletionList> => this.provideVariableWithArgCompletions(model, position),
+
+        this.registerStandardCompletionProvider({
+            triggerCharacter: PromptText.FUNCTION_CHAR,
+            getItems: () => this.toolInvocationRegistry.getAllFunctions(), // Use function to fetch the latest items
```
planger, line 97:
```suggestion
            getItems: () => this.toolInvocationRegistry.getAllFunctions(),
```

packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
```diff
@@ -49,34 +60,77 @@ export class ChatViewLanguageContribution implements FrontendApplicationContribu
...
+            kind: monaco.languages.CompletionItemKind.Function,
+            getId: tool => tool.id,
+            getName: tool => tool.name,
+            getDescription: tool => tool.description ?? ''
         });
+
+        // Register the variable argument completion provider (special case)
```
planger, line 104:
```suggestion
```

Imho this comment is rather confusing out of context

planger (changes requested):
Thanks for this improvement! It indeed is a good idea to avoid triggering a code completion if the trigger character is not the beginning of the word.

However, there are a few cases that are not covered / broken now as far as I tested:
1. We should also skip requesting completions for `provideVariableWithArgCompletions`. Currently the file suggestions still show up if the user enters `abc#re` and then types Ctr+Space.
2. It looks like there are no direct argument suggestions anymore from the file provider via `provideVariableWithArgCompletions` after typing `#`. On `master` it worked to trigger the completion with `#` and then directly start typing a file name (at least after character 3-4). Note that it still works when canceling the suggestions and retrigger them manually with Ctrl+space.
3. The argument picker isn't triggered anymore after completing a variable with arguments, e.g. `#file`. On `master` this would insert `#file:` and trigger the argument picker via the quick input.

As a side note: The implementation of the completion providers in this class was always a bit brittle a. I had to fix a few bugs here lately. This was mainly because the implementation of the completion providers relied a lot on the trigger characters instead of focusing on the current text at the position. The user can trigger an autocompletion at any time manually and it should always be complete, not just if the autocompletion is triggered by the trigger character. I aimed at improving the situation a bit with https://github.com/eclipse-theia/theia/pull/15018, but I feel that it still is somewhat brittle and inconsistent (triggered suggestions vs manual Ctrl+space). Probably because the computed range doesn't always fit with the suggestions.

eneufeld:
@planger  thank you for the review and the improvement suggestions. I updated the code and the `provideVariableWithArgCompletions` should now also work as expected

packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts
```diff
@@ -164,33 +206,42 @@ export class ChatViewLanguageContribution implements FrontendApplicationContribu
...
         if (!model || !position) {
             return;
         }
 
+        // Get the current line's content and the word at cursor
+        const lineContent = model.getLineContent(position.lineNumber);
+        const wordInfo = model.getWordUntilPosition(position);
+
+        // Only trigger if current word starts with #, followed by actual text
+        if (!wordInfo.word.startsWith(PromptText.VARIABLE_CHAR) || wordInfo.word.length <= 1) {
```
planger, line 230: This prevents the regression no 3 I [mentioned before](https://github.com/eclipse-theia/theia/pull/15030#pullrequestreview-2644624680). The word is actually empty, because we end with a `:` if the variable has mandatory arguments. So we would actually need to check if the character at the position is `PromptText.VARIABLE_SEPARATOR_CHAR` (see code below on line 246 etc.).

planger (changes requested):
Thanks for the revision! Looks much better. The regression no 3 I mentioned in my previous comment still persists. I quickly debugged the code and marked the causing line. We need to account for the `:` that we add to variables with mandatory arguments.

Thank you!

eneufeld:
> Thank you, now case 3 works again. However, I'm afraid the manual auto-completion after the `:` is know broken. Try typing #file: and then type Ctrl-space.

I can't reproduce your error, or I don't understand it:
AutoComplete

So when I type #fil I can select the #file entry which completes it to #file: and then triggers the selection at the top. I can abort it using escape key and then trigger the autocompletion using ctrl+space and get the autocompletion in place again.

planger:
> > Thank you, now case 3 works again. However, I'm afraid the manual auto-completion after the `:` is know broken. Try typing #file: and then type Ctrl-space.
> 
> I can't reproduce your error, or I don't understand it:
> AutoComplete
> 
> So when I type #fil I can select the #file entry which completes it to #file: and then triggers the selection at the top. I can abort it using escape key and then trigger the autocompletion using ctrl+space and get the autocompletion in place again.


Right, the behavior I'm referring to happens if you just type (no auto-completion) `#file:`. This should trigger a suggestion on the `:` which shows the possible arguments of the file variable.

On master this works but seemed to stop working in this PR. Let me know if you're having troubles to reproduce this.

eneufeld:
I still can't reproduce it:
AutoComplete_2

planger (approved):
I think your latest force-push actually fixed it. I tested it and now it looks like all use cases work fine.

PR #15047, opened 2025-02-25
[vscode] ShellExecution update with undefined command

rschnekenbu:
#### What it does

Adds in the API the possibility to have an undefined command or args for the shellexecution. This was already supported by the  implementation but not clearly defined in the API. 

#### How to test

1. Install the following extension: 
- src: [vscode-math-evaluator-extension-0.0.1-src.zip](https://github.com/user-attachments/files/18968576/vscode-math-evaluator-extension-0.0.1-src.zip)
- zipped vsix: [vscode-math-evaluator-extension-0.0.1.zip](https://github.com/user-attachments/files/18968578/vscode-math-evaluator-extension-0.0.1.zip)
2. Open a new markdown file and add this line to it: `1+1`. 
3. The extension will create 2 tasks with 2 different shell execution, 1 with a command line (and then command and args undefined), the second with the command+arg, and then commandLine undefined.
4. Change active editor to refresh the tasks if necessary (the tasks are recomputed on active editor change)
5. 2 messages will be displayed, one for each tasks describing the content of the shell execution for each task. 
6. Ensure that the command+ args or commandLine are undefined given the task
7. Tasks can be run to check if ShellExecution handles nicely the 2 cases using run task... command => Math => the task you want to check. 
 
https://github.com/user-attachments/assets/8ab5fba3-0ba9-459f-ad3d-c436637e7958

#### Follow-ups

No follow ups

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin/src/theia.d.ts
+++ b/packages/plugin/src/theia.d.ts
@@ -3436,10 +3436,17 @@ export module '@theia/plugin' {
         /**
          * The exit code reported by the shell.
          *
-         * Note that `undefined` means the shell either did not report an exit  code (ie. the shell
-         * integration script is misbehaving) or the shell reported a command started before the command
-         * finished (eg. a sub-shell was opened). Generally this should not happen, depending on the use
-         * case, it may be best to treat this as a failure.
+         * When this is `undefined` it can mean several things:
+         *
+         * - The shell either did not report an exit  code (ie. the shell integration script is
+         *   misbehaving)
+         * - The shell reported a command started before the command finished (eg. a sub-shell was
+         *   opened).
+         * - The user canceled the command via ctrl+c.
+         * - The user pressed enter when there was no input.
+         *
+         * Generally this should not happen. Depending on the use case, it may be best to treat this
+         * as a failure.
          *
          * @example
          * const execution = shellIntegration.executeCommand({
@@ -3449,15 +3456,14 @@ export module '@theia/plugin' {
          * window.onDidEndTerminalShellExecution(event => {
          *   if (event.execution === execution) {
          *     if (event.exitCode === undefined) {
-         *      console.log('Command finished but exit code is unknown');
+         *       console.log('Command finished but exit code is unknown');
          *     } else if (event.exitCode === 0) {
-         *      console.log('Command succeeded');
+         *       console.log('Command succeeded');
          *     } else {
-         *      console.log('Command failed');
+         *       console.log('Command failed');
          *     }
          *   }
          * });
-         * @stubbed
          */
         readonly exitCode: number | undefined;
     }
@@ -13186,7 +13192,7 @@ export module '@theia/plugin' {
         /**
          * The shell command line. Is `undefined` if created with a command and arguments.
          */
-        commandLine?: string;
+        commandLine: string | undefined;
 
         /**
          * The shell options used when the command line is executed in a shell.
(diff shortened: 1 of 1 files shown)


packages/plugin/src/theia.d.ts
```diff
@@ -3436,10 +3436,17 @@ export module '@theia/plugin' {
         /**
          * The exit code reported by the shell.
          *
-         * Note that `undefined` means the shell either did not report an exit  code (ie. the shell
-         * integration script is misbehaving) or the shell reported a command started before the command
-         * finished (eg. a sub-shell was opened). Generally this should not happen, depending on the use
-         * case, it may be best to treat this as a failure.
+         * When this is `undefined` it can mean several things:
```
tsmaeder, line 3439: These are just drive-by fixes, right?

packages/plugin/src/theia.d.ts
```diff
@@ -13197,12 +13203,12 @@ export module '@theia/plugin' {
         /**
          * The shell command. Is `undefined` if created with a full command line.
          */
-        command?: string | ShellQuotedString;
+        command: string | ShellQuotedString | undefined;
```
tsmaeder, line 13206: If I read this change correctly, it does not really change the usage of the class in any way, right? it probably would force literals to include a value for "command" if it was an interface, but not for a class.
That said, if I look at the implementation of `ShellExecution` in `types-impl.ts`, I see this:
```
set commandLine(value: string) {
        if (typeof value !== 'string') {
            throw illegalArgument('commandLine');
        }
```

So it throws an exception if I assign `undefined`. But that's not consistent with the tying. What's the reason?

packages/plugin/src/theia.d.ts
rschnekenbu: Yes, the comments were changed in equivalent file in vscode, but these are only better explanation on the API behavior.

packages/plugin/src/theia.d.ts
rschnekenbu: Basically, for the shell execution, we are supposed to have either commandLine set to a valid string, or the command and eventually args. 
The check in the types-impl is redundant with the one in the constructor. I would also not expect the shell execution to be changed at runtime. In case we still want to change at runtime, this may be done using an atomic change between command line and command + args, to ensure only one group is defined at a time.

tsmaeder: I have a `command` value, shouldn't we be checking that we don't have a `commandLine` value? Because we're not supposed to have both, right? Anyway, why aren't we respecting the interface and complaining at execution time that the user set something unusable. 
>  I would also not expect the shell execution to be changed at runtime.

But the API allows it, so people are going to do it 🤷 This check might break exensions that do 
``` 
execution.command  = undefined;
execution.commandLine = 'ls -ls`;
```
which would work perfectly fine.

rschnekenbu: So if I understand you, we should allow `undefined` in the setCommand, setCommandLine (and setArgs), remove the checks done on the values, and allow the getters to return also `undefined`? 
Otherwise, we would have issues with the first line 
`execution.command  = undefined;`
both commandLine and command would be undefined at the same time.

tsmaeder: The typing says you can assign undefined to  `command`, so if the implementation is not allowing that, there needs to be a reasons for that. If we try to not allow both `command` and `commandLine`, we would have to have a different check. It seems the checks are coming from the `extHostImplementation.ts` in VS Code. But those checks would still allow for both `command` and `commandLine` to be set. So we're bug-compatible with VS Code in this respect, but it is still wrong and it still bugs me 🤷

tsmaeder (approved):
This is a typing-only change that should not affect clients and the implementation, albeit bogus, is compatible with VS Code and has been like that for years.

rschnekenbu:
Thanks @tsmaeder, I created https://github.com/eclipse-theia/theia/issues/15062 as a follow up to tackle the implementation issues and not this API update.

PR #15064, opened 2025-02-26
fix: quick input hover initialization

### LINKED ISSUE #15042: "bad scroll event" occurring in Command Palette in Theia if the application does not include the "@theia/ai-chat-ui" module
### Bug Description:

When starting to develop a fresh extension using the yeoman code generator, the default electron-app and browser-app folders each contain a `package.json` listing a minimum Theia environment as dependencies.

But when I build an run these applications, and in the fresh, empty Theia application, I invoke the Command Palette (using CMD+SHIFT+P), the Command Palette is not cleanly formatted, but looks distorted.
Additionally, on the console, this error is logged:

```text
ERROR Got bad scroll event: {
  inSmoothScrolling: false,
  oldWidth: 0,
  oldScrollWidth: 0,
  oldScrollLeft: 0,
  width: 0,
  scrollWidth: 0,
  scrollLeft: 0,
  oldHeight: 0,
  oldScrollHeight: 0,
  oldScrollTop: 0,
  height: 226,
  scrollHeight: 1980,
  scrollTop: 0,
  widthChanged: false,
  scrollWidthChanged: false,
  scrollLeftChanged: false,
  heightChanged: true,
  scrollHeightChanged: true,
  scrollTopChanged: false
}
```

I have started to debug this by including the full set of dependencies from Theia IDE and one by one stripping them.
For some reason, the problem seems to be that the "@theia/ai-chat-ui" dependency is missing, because, when I add it to the minimal application, the Command Palette works as expected; but when I remove it again, the Command Palette is distorted again.

Maybe there is an implicit dependency somewhere that causes this? (Or it is a transitive dependency of the _ai-chat-ui_ module ...?). 

In any case, the yeoman generator should be fixed to include the necessary dependency (or better yet, if it is really the missing _ai-chat-ui_ module that causes this, I would expect the minimal Theia to even work fine without any AI module installed...).

### Steps to Reproduce:

1. Start in an empty directory and invoke `yo theia-extension`
2. Create the HelloWorld extension
3. Start `yarn && yarn build:electron && yarn start:electron`
4. In the launched Theia application, bring up the Command Palette using CMD+SHIFT+P

### Additional Information

- Operating System: MacOS 15.3 (24D60) (M1 Max / arm64)
- Theia Version: 1.58.3
- generator-theia-extension@0.1.40

sdirix:
#### What it does

Makes sure the global Monaco BaseLayerHoverDelegate is properly initialized with Monaco's HoverService before opening a quick input.

This fixes rendering issues caused by uncaught errors when opening a quick input, for example the command palette, before any Monaco editor was initialized. This scenario was easily reproducible in minimal Theia applications, not leveraging all offered Theia extensions.

fixes #15042

#### How to test

- Adapt the browser example application and remove
  - all AI packages
  - all plugin packages
  - <details>
    <summary>You can use this list for convenience: </summary>

    ```
        "@theia/console": "1.58.0",
        "@theia/core": "1.58.0",
        "@theia/debug": "1.58.0",
        "@theia/dev-container": "1.58.0",
        "@theia/editor": "1.58.0",
        "@theia/editor-preview": "1.58.0",
        "@theia/file-search": "1.58.0",
        "@theia/filesystem": "1.58.0",
        "@theia/keymaps": "1.58.0",
        "@theia/markers": "1.58.0",
        "@theia/messages": "1.58.0",
        "@theia/monaco": "1.58.0",
        "@theia/navigator": "1.58.0",
        "@theia/outline-view": "1.58.0",
        "@theia/output": "1.58.0",
        "@theia/preview": "1.58.0",
        "@theia/process": "1.58.0",
        "@theia/property-view": "1.58.0",
        "@theia/remote": "1.58.0",
        "@theia/scanoss": "1.58.0",
        "@theia/scm": "1.58.0",
        "@theia/scm-extra": "1.58.0",
        "@theia/search-in-workspace": "1.58.0",
        "@theia/secondary-window": "1.58.0",
        "@theia/task": "1.58.0",
        "@theia/terminal": "1.58.0",
        "@theia/test": "1.58.0",
        "@theia/timeline": "1.58.0",
        "@theia/toolbar": "1.58.0",
        "@theia/typehierarchy": "1.58.0",
        "@theia/variable-resolver": "1.58.0",
        "@theia/workspace": "1.58.0"
    ```

    </details>
- Open the browser example on a workspace you did not use before, so that there are no "recent' entries in your quick input
- Open the command palette `Ctrl + Shift + P`
- Observe that the palette works as expected

Without the fix, the palette's rendering will be broken and errors thrown in the console

---

As an alternative reproducer you can use the Theia generator, there the problem immediately appears.

#### Follow-ups

- This fixes the issue for the quick input. Previously I discussed with @tsmaeder to move this code to `monaco-init`. However as this change is less invasive, I put it into the quick input for now.
- This fix should be backported to the community release

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


tsmaeder:
@sdirix I think I like the solution in `monaco-init.ts` better, here's why: if another service uses the hover delegate singleton, it might not be fixed with the current fix. You're saying the current solution is less intrusive, but I don't see how that is really the case: we're not really using any monaco services before monaco-init, (the overrides would not work if we do). It's easily testable whether the hover service works at that time. And we're overriding the singleton as soon as we're opening an editor anyway.
Since this is not relevant for Theia IDE, I would marge this PR after the Theia IDE. If it needs to be backported for adopters, we should back-port it to the community release anyway, not 1..59

sdirix:
> @sdirix I think I like the solution in `monaco-init.ts` better, here's why: if another service uses the hover delegate singleton, it might not be fixed with the current fix. You're saying the current solution is less intrusive, but I don't see how that is really the case: we're not really using any monaco services before monaco-init, (the overrides would not work if we do). It's easily testable whether the construction of the hover service works at that time. And we're overriding the singleton as soon as we're opening an editor anyway.

Okay I will try this too

> Since this is not relevant for Theia IDE, I would marge this PR after the Theia IDE release. If it needs to be backported for adopters, we should back-port it to the community release anyway, not 1..59

If we don't include it in `1.59` then every user of the yeoman Theia generator will continue to experience the issue as well as every adopter using a minimal Theia code base consuming `1.58` or `1.59`

sdirix:
@tsmaeder updated the fix as discussed

Issue #15110, opened 2025-02-26
Watch for changes fails due too-large workspace


cdamus:
### Bug Description:

In a Git workspace that is an order of magnitude smaller than Theia, when I run a `yarn` build, often (but not always), Theia IDE will notify me that it has had to stop watching for changes in the workspace because my workspace is too big. And it includes a link to instructions to work around the issue. That links to VS Code user documentation at visualstudio.com, not to theia-ide.org which is the website for the software I'm using.

### Steps to Reproduce:

1. Edit text in some TypeScript files.
2. Run a `yarn` build.
3. See the notification shown below (or repeat from step 1 until you see it).

### Additional Information

- Operating System: macOS Sequoia 15.3.1
- Theia Version: 1.58.300.206

![Image](https://github.com/user-attachments/assets/47e4c7d0-3b6c-45d2-8fd4-f4535d899792)

PR #15111, opened 2025-03-04
Don't stop watching folder because of dropped events.

tsmaeder:
Theia stops watching directories when an error is passed to the event handling function. This includes the case where the watcher has dropped some events due to performance reasons. This PR does not stop watching the directory anymore in this case.

Mitigates #15110

I call this a mitigation, becuase on Arm Macs, it is not expected that we run into a performance issue for the scenarios mentioned in the issue.

#### How to test
Follow the development scenarios mentioned in the issue on a Arm Mac.

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/filesystem/src/node/parcel-watcher/parcel-filesystem-service.ts
+++ b/packages/filesystem/src/node/parcel-watcher/parcel-filesystem-service.ts
@@ -242,7 +242,7 @@ export class ParcelWatcher {
         }
         return subscribe(fsPath, (err, events) => {
             if (err) {
-                if (err.message && err.message.indexOf('File system must be re-scanned') !== -1) {
+                if (err.message && err.message.includes('File system must be re-scanned')) {
                     console.log(`FS Events were dropped on watcher ${fsp}`);
                 } else {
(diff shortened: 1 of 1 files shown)


packages/filesystem/src/node/parcel-watcher/parcel-filesystem-service.ts
```diff
@@ -242,12 +242,18 @@ export class ParcelWatcher {
         }
         return subscribe(fsPath, (err, events) => {
             if (err) {
-                console.error(`Watcher service error on "${fsPath}":`, err);
-                this._dispose();
-                this.fireError();
-                return;
+                if (err.message && err.message.indexOf('File system must be re-scanned') !== -1) {
```
msujew, line 245: Suggestion: `includes` has been a thing for a few years now :)

```suggestion
                if (err.message && err.message.includes('File system must be re-scanned')) {
```

msujew (approved):
Makes sense, LGTM 👍

Issue #15115, opened 2025-03-05
Make it easier to find chat sessions by properly naming them


planger:
Finding chat sessions is currently very difficult as we are recoding the last request as a name. It'd be better to use a language model to generate a descriptive name for the chat session and also record the last interaction date, to have proper sorting and a chronological relation to the chat sessions.

Moreover, users should be enabled to manually specify a name for the chat session, which will be used instead of the generated name.

PR #15116, opened 2025-03-05
Planger/naming-chat-sessions

planger:
#### What it does

Finding chat sessions is currently very difficult as we are recoding the last request as a name. It'd be better to use a language model to generate a name for the chat session and also record the last interaction date, to have proper sorting and a chronological relation to the chat sessions.

Moreover, this PR allows users to manually specify a name for the chat session, which will be used instead of the generated name.

Fixes https://github.com/eclipse-theia/theia/issues/15115

#### How to test

Create a few chat sessions and invoke the chat switcher.
Try creating the chat message in a few different languages to check if the name follows this language.

Also switch locale and check if the last interaction time indicator in the quick picks follows the locale.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


JonasHelming:
@sdirix Ping

packages/ai-chat/src/common/chat-session-naming-service.ts
```diff
@@ -0,0 +1,122 @@
...
+        'If the chat message contains @<some-name>, this means that the user addressed a specific chat agent.' +
+        'Use the same language for the name as used in the provided conversation, if in doubt default to English.' +
+        'Below we also provide the already existing other conversation names, make sure your suggestion for a name is unique with respect to the existing ones.' +
+        'IMPORTANT: Your answer MUST ONLY CONTAIN THE PROPOSED NAME and must not be preceded or succeeded with any other text.' +
+        '\n\nOther session names:\n{{listOfSessionNames}}' +
+        '\n\nConversation:\n{{conversation}}',
+};
+
+@injectable()
+export class ChatSessionNamingService implements Agent {
```
sdirix, line 46: Is there a specific reason to name this `ChatSessionNamingService`? I would have expected `ChatSessionNamingAgent` similar to the other agents.

packages/ai-chat/src/common/chat-service.ts
```diff
@@ -127,6 +130,9 @@ export class ChatServiceImpl implements ChatService {
     @inject(FallbackChatAgentId) @optional()
     protected fallbackChatAgentId: FallbackChatAgentId | undefined;
 
+    @inject(ChatSessionNamingService) @optional()
+    protected chatSessionNamingService: ChatSessionNamingService | undefined;
```
sdirix, line 134: We should not directly inject the agent but go through the registry. That way, the user settings regarding enabling/disabling this agent will be taken into account.

packages/ai-chat/src/common/chat-service.ts
```diff
@@ -255,6 +260,28 @@ export class ChatServiceImpl implements ChatService {
...
+            const namingService = this.chatSessionNamingService;
+            let didGenerateName = false;
+            request.response.onDidChange(() => {
+                if (request.response.isComplete && !didGenerateName) {
+                    namingService.generateChatSessionName(session, otherSessionNames).then(name => {
+                        session.title = name;
+                        didGenerateName = true;
+                    }).catch(error => this.logger.error('Failed to generate chat session name', error));
+                }
+            });
```
sdirix, line 281: This does not take into account when the user manually edited the name while the response was not complete, i.e. the user name will be overwritten.

packages/ai-chat/src/common/chat-session-naming-service.ts
```diff
@@ -0,0 +1,122 @@
...
+import { ChatSession } from './chat-service';
+import { generateUuid } from '@theia/core';
+
+const CHAT_SESSION_NAMING_PROMPT = {
+    id: 'chat-session-naming-service-prompt',
+    template: '{{!-- Made improvements or adaptations to this prompt template? We\'d love for you to share it with the community! Contribute back here: ' +
+        'https://github.com/eclipse-theia/theia/discussions/new?category=prompt-template-contribution --}}\n\n' +
+        'Provide a short and descriptive name for the given AI chat conversation of an AI-powered tool based on the conversation below.' +
+        'The purpose of the name is for users to recognize the chat conversation easily in a list of conversations.' +
+        'If the chat message contains @<some-name>, this means that the user addressed a specific chat agent.' +
```
sdirix, line 37: Does this help in your experience?
```suggestion
        'If the chat message contains @<some-name>, this means that the user addressed a specific chat agent.' +
```

packages/ai-chat/src/common/chat-session-naming-service.ts
```diff
@@ -0,0 +1,122 @@
...
+        'Use the same language for the name as used in the provided conversation, if in doubt default to English.' +
+        'Below we also provide the already existing other conversation names, make sure your suggestion for a name is unique with respect to the existing ones.' +
+        'IMPORTANT: Your answer MUST ONLY CONTAIN THE PROPOSED NAME and must not be preceded or succeeded with any other text.' +
```
sdirix, line 40: Should we enforce a structured response?

packages/ai-chat/src/common/chat-session-naming-service.ts
```diff
@@ -0,0 +1,122 @@
...
+        const requestId = generateUuid();
+        this.recordingService.recordRequest({
+            agentId: this.id,
+            sessionId,
+            requestId,
+            ...request
+        });
+
+        const result = await lm.request(request);
+        const response = await getTextOfResponse(result);
```
sdirix, line 110: Should we do some sanity checks, i.e. check for markers, length etc?

sdirix (changes requested):
Works for me, however I have some comments.

packages/ai-chat/src/common/chat-session-naming-service.ts
planger: Hm, good question. I felt it is rather a service that happens to be implemented as an agent.
Probably the cleanest approach is to introduce a service, that uses an agent. This also allows adopters to rebind the service with a non-LLM-based naming approach.

packages/ai-chat/src/common/chat-session-naming-service.ts
planger: I'd prefer that, yes. But I refrained from using a structured response to not limit the LLMs that can be used for this agent (as not all support structured responses). So the benefit of the structured response seemed to weigh less than the gained flexibility in this case. I didn't really encounter issues without structured response, at least with 4o and 4o-mini.

packages/ai-chat/src/common/chat-session-naming-service.ts
planger: Good question, I didn't encounter issues, but I guess, you're right, we should remove at least line breaks and limit to a certain number of characters.

packages/ai-chat/src/common/chat-session-naming-service.ts
planger: I did some vibe testing and it indeed doesn't seem to have a great impact actually. I'll remove it. But I added better line breaks and mention that chat conversation names should start with an upper case.

packages/ai-chat/src/common/chat-session-naming-service.ts
```diff
@@ -0,0 +1,122 @@
...
+@injectable()
+export class ChatSessionNamingService implements Agent {
+    id = 'chat-session-naming-service';
+    name = 'Chat Session Naming Service';
+    description = 'Service to generate chat session names';
+    variables = [];
+    promptTemplates: PromptTemplate[] = [CHAT_SESSION_NAMING_PROMPT];
+    languageModelRequirements: LanguageModelRequirement[] = [{
+        purpose: 'chat-session-naming',
+        identifier: 'openai/gpt-4o',
```
planger, line 54: Actually I think we can switch to gpt-4o-mini for this task to save some time and cost.

planger:
@sdirix ping... thank you! :-)

packages/ai-chat/src/common/chat-session-naming-service.ts
```diff
@@ -0,0 +1,135 @@
...
+        }
+        return chatSessionNamingAgent.generateChatSessionName(chatSession, otherNames);
+    }
+}
+
+@injectable()
+export class ChatSessionNamingAgent implements Agent {
+    static ID = 'chat-session-naming-service';
+    id = ChatSessionNamingAgent.ID;
+    name = 'Chat Session Naming Service';
```
sdirix, line 62:
```suggestion
    name = 'Chat Session Naming';
```

packages/ai-chat/src/common/chat-service.ts
```diff
@@ -241,6 +247,30 @@ export class ChatServiceImpl implements ChatService {
...
+                if (request.response.isComplete && !didGenerateName) {
+                    namingService.generateChatSessionName(session, otherSessionNames).then(name => {
+                        if (name && session.title === requestText) {
+                            session.title = name;
+                        }
+                        didGenerateName = true;
+                    }).catch(error => this.logger.error('Failed to generate chat session name', error));
+                }
+            });
+        }
```
sdirix, line 271: It's easy to rename the chat when the response is not yet completed. In that case we override the manually set name again.

sdirix (approved):
Works great for me.

I have two minor comments. Feel free to tackle them or create follow ups.

packages/ai-chat/src/common/chat-session-naming-service.ts
```diff
@@ -0,0 +1,135 @@
...
+    LanguageModelRequirement,
+    PromptService,
+    PromptTemplate
+} from '@theia/ai-core';
+import { inject, injectable } from '@theia/core/shared/inversify';
+import { ChatSession } from './chat-service';
+import { generateUuid } from '@theia/core';
+
+const CHAT_SESSION_NAMING_PROMPT = {
+    id: 'chat-session-naming-service-prompt',
```
planger, line 33:
```suggestion
    id: 'chat-session-naming-prompt',
```

packages/ai-chat/src/common/chat-session-naming-service.ts
```diff
@@ -0,0 +1,135 @@
...
+        if (!(chatSessionNamingAgent instanceof ChatSessionNamingAgent)) {
+            return undefined;
+        }
+        return chatSessionNamingAgent.generateChatSessionName(chatSession, otherNames);
+    }
+}
+
+@injectable()
+export class ChatSessionNamingAgent implements Agent {
+    static ID = 'chat-session-naming-service';
```
planger, line 60:
```suggestion
    static ID = 'chat-session-naming-agent';
```

packages/ai-chat/src/common/chat-session-naming-service.ts
```diff
@@ -0,0 +1,135 @@
...
+    id = ChatSessionNamingAgent.ID;
+    name = 'Chat Session Naming';
+    description = 'Service to generate chat session names';
```
planger, line 63:
```suggestion
    description = 'Agent to generate chat session names';
```

PR #14695, opened 2025-01-06
Fix z-index overlay issue in dock panels

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/14290

Phosphor.js automatically sets the `z-index: 0` css property on all dock panel items and children of split panels. However, this prevents the monaco editor popups from actually popping up on top of other panel items, therefore being partially hidden.

I've tested the fix suggested in the issue and found no further bugs with it. Completion, hover and diagnostics all seem to work as expected without any unexpected side effects.

#### How to test

See issue https://github.com/eclipse-theia/theia/issues/14290 for reproduction steps. Ensure that the popups are fully visible.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #15154, opened 2025-03-08
fix: clipping issue in debug hover (#15086)

### LINKED ISSUE #15086: Debug hover is clipped
### Bug Description:

When debugging with Theia IDE, I hover over a variable to show its properties.
If the hover does not fit in the editor, it is clipped at the editor's borders.
(whereas, e.g., code assist hovers in the editor are not clipped)

Image

### Steps to Reproduce:

1. Debug a typescript application within Theia IDE
2. Place and hit a breakpoint
3. Make the editor part small enough so the hover does not fit within its borders
4. Hover over an object 

### Additional Information

- Operating System: MacOS 15.3
- Theia Version: Theia IDE 1.58.100

xpomul:
#### What it does

Fixes #15086 by adding a z-index property to the `theia-debug-hover` CSS class, as proposed in that issue.
The root cause was related to PR https://github.com/eclipse-theia/theia/pull/14695 which has addressed overlay issues in the dock panels.

Notes: 
1. I did a quick check of how z-index is used in Theia's css files but there seems no greater logic and z-indexes seem to be assigned on a "whatever works" basis. I chose 1000 because the Debug Hover should be on top of other things, but maybe not the Notification Panel (which has a z-index of 1111).
2. When applying the change and saving, the workspace formatter settings introduce several format changes. Obviously the affected CSS file is not in a proper format according to the Theia guidelines. I chose to not include the formatting changes into this PR to keep it small and simple.

#### How to test

Simplest way of testing:
- In the Debug Panel, check the "Caught Exceptions" breakpoint
- Launch "Start Browser Backend" and wait for the process to suspend
- Hover in the source editor (possibly resize the window in a way that the hover overlaps with docked panels)

See #15086 for a screenshot.

#### Follow-ups

See notes in the "What it does" section. The file `packages/debug/src/browser/style/index.css` should be reformatted according to the Theia project settings. (But most likely this is not the only file that is affected...?)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/style/index.css
+++ b/packages/debug/src/browser/style/index.css
@@ -366,6 +366,7 @@
   border: 1px solid var(--theia-editorHoverWidget-border);
   background: var(--theia-editorHoverWidget-background);
   color: var(--theia-editorHoverWidget-foreground);
+  z-index: 1000;
 }
 
 .theia-debug-hover.complex-value {


sgraband (approved):
Thank you for the fix! LGTM :+1: 1000 seems like a reasonable value imho.

PR #15161, opened 2025-03-11
Ensure word-wrap respected when diff editors side by side

### LINKED ISSUE #15105: [Theia AI] Diff editor line wrapping
The diff editor does not use line wrapping on the left side, but uses line wrapping on the right hand side. This makes the diff editor hard to use for text documents. I did not see a setting to switch line wrapping for the diff editor. This applies for the change set diff editor, but also the regular triggered from the git view.

colin-grant-work:
#### What it does

Fixes #15105
See also https://github.com/microsoft/vscode/issues/217386#issuecomment-2711750462

Adds code to ensure that the `wordWrapOverride2` setting is explicitly cleared when a diff editor is large enough for side-by-side rendering.

#### How to test

1. Set the `wordWrap` preference to `true` and open a diff editor.
2. Any lines that need to wrap should wrap on both sides.
3. Resize the editor horizontally so that it converts to inline diff rendering. (By default, at 900 pixels)
4. It should convert smoothly to inline mode.
5. Resize it so that it is large enough for side by side rendering.
6. Word wrap should work on both sides of the editor again.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-diff-editor.ts
+++ b/packages/monaco/src/browser/monaco-diff-editor.ts
@@ -31,6 +31,8 @@ import { IInstantiationService } from '@theia/monaco-editor-core/esm/vs/platform
 import { ContextKeyValue, IContextKey } from '@theia/monaco-editor-core/esm/vs/platform/contextkey/common/contextkey';
 import { IDisposable } from '@theia/monaco-editor-core/esm/vs/base/common/lifecycle';
 import { ICommandHandler } from '@theia/monaco-editor-core/esm/vs/platform/commands/common/commands';
+import { EditorContextKeys } from '@theia/monaco-editor-core/esm/vs/editor/common/editorContextKeys';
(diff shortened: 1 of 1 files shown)


sdirix (approved):
Works for me!

Note that there is a related minor issue which I found: When a diff editor is open and we switch the word wrap preference, then the editor reacts accordingly, however the diff markers are then gone afterwards.

If it makes sense you could fix the issue here too or we create a follow up ticket.

colin-grant-work:
>  When a diff editor is open and we switch the word wrap preference, then the editor reacts accordingly, however the diff markers are then gone afterwards.

It doesn't appear to have to do with toggling the word wrap preference, but with hiding the editor. If you use Alt + Z to toggle the preference, or put the preference view in a different tab bar and then open the diff editor and then use the preference view to change the preference, the editor reacts correctly, and if you hide the editor in any way and then show it again without doing anything else, the diff markers are gone. I suspect that this is has to do with @tsmaeder's recent changes to set and unset the model when an editor is hidden, so I'll create a follow up ticket.

Issue #15165, opened 2025-03-11
Two-segment preferences not rendered if first segment is group


colin-grant-work:
### Steps to Reproduce:

1. Add code like this to any preference contribution (I added it in `packages/ai-ide/src/browser/workspace-preferences.ts`
```typescript
        'ai-features.agent': {
            type: 'object',
            properties: {
                none: {type: 'null'},
            },
            required: [],
            markdownDescription: nls.localize('theia/ai/ide/agent-description',
                'Additional settings for AI agents can be configured using the [AI Configuration View]({0}).',
               'command:aiConfiguration:open'
            )
        },
```
2. Start the application, open preferences, navigate to the AI Features section.
3. See that the `agent` preference is rendered as a header, but with no children:

![Image](https://github.com/user-attachments/assets/225506d9-70b0-4d4c-9107-2b80361c0fe5)

The issue comes from [this change](https://github.com/eclipse-theia/theia/pull/13819/files#r1989564719). Because `ai-features` is a group, we don't enter the first `if` block - we have now consumed the first segment of the labels `['ai-features', 'agent']` - and we enter the `else if` block. Previously, we would not have entered the `else if` block because the length of the label was 2, meaning that after consuming the first segment, the second segment was the leaf label, not a subgroup. @msujew, what was the motivation for that change in the context of your refactor?

### Additional Information

- Operating System: Mac OS
- Theia Version: dc1e0a2e7a10001165a0758a850900c1ab6bdbdd

colin-grant-work:
It also looks as though intermediate segments aren't split from camelCase -> Camel Case. E.g. 

```typescript
        'ai-features.promptTemplates.details': {
            type: 'null',
            markdownDescription: nls.localize('theia/ai/ide/prompt-template-description',
                'Additional AI prompt template settings can be configured using the [AI Configuration View]({0}).',
                'command:aiConfiguration:open'
            )
        },
```

creates an intermediate label `PromptTemplates` not `Prompt Templates`.

msujew:
> Previously, we would not have entered the else if block because the length of the label was 2, meaning that after consuming the first segment, the second segment was the leaf label, not a subgroup. @msujew, what was the motivation for that change in the context of your refactor?

Not sure I remember that quite correctly - it just happened to work well with the preferences that were available at that time in the Theia framework. The AI features came later. Feel free to adjust it back to 2, in case it doesn't lead to any regressions.

[...]

PR #15166, opened 2025-03-11
Add dummy preference descriptions to open AI config widget

### LINKED ISSUE #15160: [Theia AI] Add reference to AI configuration view in the settings
Add a settings entry "Agent" (under AI Features) that explains that agent and prompt settings are conducted in the AI configuration view and add a link to open it. Add a similar entry under the category "PromptTemplates" (as prompts are configured per agent). And finally, the same under "model settings" (as you can select the LLM to be used per agent)

colin-grant-work:
#### What it does

Fixes #15160 by adding preferences whose descriptions redirect users to the config widget to enable them to perform more advanced AI configuration.

Fixes https://github.com/eclipse-theia/theia/issues/15165 by generating section names differently depending on presence or absence of layout item and by using the preference tree label provider's logic for label generation.

#### How to test

1. Open the settings and look for settings under 'AI Features ... Agents', 'AI Features ... Models', and 'AI Features ... Prompt Templates'
2. You will find a dummy preference with a link to open the configuration widget.
3. Click the link, and the config widget should open.

The third segment (`details`) is to work around https://github.com/eclipse-theia/theia/issues/15165

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/browser/views/components/preference-null-input.ts
+++ b/packages/preferences/src/browser/views/components/preference-null-input.ts
@@ -1,0 +1,52 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { injectable, interfaces } from '@theia/core/shared/inversify';
+import { PreferenceLeafNodeRenderer, PreferenceNodeRenderer } from './preference-node-renderer';
+import { Preference } from '../../util/preference-types';
+import { PreferenceLeafNodeRendererContribution } from './preference-node-renderer-creator';
+
+@injectable()
+/** For rendering preference items for which the only interesting feature is the description */
+export class PreferenceNullInputRenderer extends PreferenceLeafNodeRenderer<null, HTMLElement> {
+    protected override createInteractable(container: HTMLElement): void {
+        const span = document.createElement('span');
+        this.interactable = span;
+        container.appendChild(span);
+    }
+
+    protected override getFallbackValue(): null {
+        // eslint-disable-next-line no-null/no-null
+        return null;
+    }
+
+    protected override doHandleValueChange(): void { }
+}
+
+@injectable()
+export class PreferenceNullRendererContribution extends PreferenceLeafNodeRendererContribution {
+    static ID = 'preference-null-renderer';
+    id = PreferenceNullRendererContribution.ID;
+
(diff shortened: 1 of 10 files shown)


sdirix (commented):
It fulfills the purpose but I'm not so happy about the result as the "Edit in settings.json" is wildly misleading.

image

We might need to register a custom preference render to avoid producing it.

And if we're doing that, we could instead have a custom renderer for these sections and within there render the hint. This would allow us to avoid declaring a non-existing preference.

Generally speaking, supporting a description text for sections would be nice but might be out of scope for this issue.

colin-grant-work:
> Generally speaking, supporting a description text for sections would be nice but might be out of scope for this issue.

I had the same thought. One easy workaround would be a special renderer for preferences with declared type `null` that just doesn't have any interactable at all.

colin-grant-work:
I've pushed a commit that adds a renderer for preferences whose only declared type is `null` and confirmed that that includes only the three that I've added here so that only the message appears.

packages/preferences/src/browser/views/components/preference-null-input.ts
```diff
@@ -0,0 +1,52 @@
+// *****************************************************************************
+// Copyright (C) 2025 Eclipse GmbH and others.
```
sdirix, line 2:
```suggestion
// Copyright (C) 2025 EclipseSource GmbH and others.
```

sdirix (commented):
I'm a bit torn because it works, however in the end it's still a hack. 

It's a hack because we have many unintended side effects:
 - we register preferences although there are none
 - we offer these preferences in the JSON Editor for autocomplete etc.
 - we offer UI to copy the preferences name, id, etc. although they are useless
 - they influence search in the preferences UI

Of course the overall impact is minimal, and I also don't see any collateral as I can't think of a real world use case for preferences of type null.

---

Overall it would be much nicer to enrich sections to allow optional descriptions. then we could add this text to these sections but they would not be registered.

--

However I don't want to block the merge if we want to tackle this in a follow up then

sdirix (changes requested):
We should not get rid of the AI subsections:

Screenshot from 2025-03-18 14-45-31

image

I think they are necessary for quick access and as an overview over the AI features.

colin-grant-work:
> We should not get rid of the AI subsections:

Quite right, looks like leaving things in a skeletal state in https://github.com/eclipse-theia/theia/pull/14048 created a number of issues in the preference system. I'll fill out the children that should have been added then.

sdirix:
If we remove `settings: ['ai-features.*']` from the preference layout again, then the sections are back. Or do you prefer a manual approach?

colin-grant-work:
I'll do both: #13819 assumes that everything should be in the layout provider, and so broke sections that didn't manually specify all their children, because two-segment sections aren't rendered and intermediate labels aren't split correctly (`'CodeCompletion'` instead of `Code Completion`). 

Manual specification has the advantage of better NLS support, but now I'm frustrated enough with the problems that I'll go and fix the automatic generation more thoroughly, as well, rather than plastering over it with the `['ai-features.*']` rule - shouldn't take long.

sdirix (approved):
Works for me. Thanks! :+1: 

---

Coming back to the original topic of this PR: Using pseudo-preferences to achieve UI changes feels like we're working around the system rather than with it. This approach brings some minor but notable side effects that aren't ideal from a technical standpoint.

That said, the user experience benefits are significant enough that I'll approve this PR. I do think we should consider a cleaner solution in the future that accomplishes the same UI improvements without repurposing our preference system in ways it wasn't designed for.

sdirix (approved):

PR #15185, opened 2025-03-13
chore: use shared packages in src-gen

sdirix:
#### What it does

Src-gen is executed in the environment controlled by adopters. In the generated code dependencies are imported. If these dependencies are not located where webpack expects them, the adopter will experience build errors. This is not a common case, but it does happen in practice.

The generated code is made more robust by reusing the shared core dependencies where applicable. As it's common to declare '@theia/core' as a dependency within the adopter package, this more commonly results in a working build.

#### How to test

Make sure example applications continue working

#### Follow-ups

This PR only adjusts imports which are already exported by `@theia/core`. This already improves the issue in real world use cases in which these packages were the affected ones in build errors.

For a more complete solution we should:
 - Reexport all packages in `@theia/core` which are used in src-gen, **or**,
 - Instruct adopters that this issue can occur with more esoteric setups and advise on how to resolve that, for example by redeclaring the non-shared-dependencies in the adopter package or making sure that the affected dependencies are hoisted.

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/application-manager/src/generator/backend-generator.ts
+++ b/dev-packages/application-manager/src/generator/backend-generator.ts
@@ -123,8 +123,8 @@ if ('ELECTRON_RUN_AS_NODE' in process.env) {
 
 const path = require('path');
 process.env.THEIA_APP_PROJECT_PATH = path.resolve(__dirname, '..', '..')
-const express = require('express');
-const { Container } = require('inversify');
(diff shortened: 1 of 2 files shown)


rschnekenbu (approved):
LGTM!

Issue #15178, opened 2025-03-12
Diff markers lost when diff editor hidden


colin-grant-work:
When a diff editor is hidden and shown again, the diff highlights disappear from the modified (right) side. It may have to do with @tsmaeder's recent changes to set and unset the model when an editor is hidden.

## To reproduce

1. Open a diff editor (e.g. make a change and then use the SCM view, or compare two files)
2. Hide the diff editor (e.g. open another editor, or the settings widget)
3. Reveal the editor.
4. Observe that there are no diff highlights on the right side.

A possibly related issue is that if a diff editor is hidden when changes affecting its display are made - e.g. toggling word wrapping - an error is thrown that points to this Monaco code:

https://github.com/microsoft/vscode/blob/c2261b02faf97df586d23f8b7d2eb75e67bd00ed/src/vs/editor/browser/widget/diffEditor/components/diffEditorViewZones/diffEditorViewZones.ts#L601

indicating that the non-null assertion has failed: `_getViewModel()` has returned `null`.

JonasHelming:
Thank you for reporting this, could you have a look at a fix?

colin-grant-work:
The full stack is below. The part of interest to us is the entry from our code at 

```
updateMonacoDiffEditorOptions (monaco-editor-provider.ts:317)
```

which kicks off updates on the Monaco side, which through various layers of observables ends up wanting to calculate how the diff regions match up.

```
getAdditionalLineHeights (diffEditorViewZones.ts:601)
computeRangeAlignment (diffEditorViewZones.ts:466)
(anonymous) (diffEditorViewZones.ts:96)
_recomputeIfNeeded (derived.ts:301)
get (derived.ts:272)
reportChanges (base.ts:188)
get (derived.ts:257)
reportChanges (base.ts:188)
get (derived.ts:257)
reportChanges (base.ts:188)
endUpdate (autorun.ts:265)
endUpdate (derived.ts:374)
endUpdate (derived.ts:374)
endUpdate (derived.ts:374)
finish (base.ts:371)
set (base.ts:444)
(anonymous) (diffEditorViewZones.ts:72)
doRun (async.ts:1064)
onTimeout (async.ts:1059)
setTimeout (async)
schedule (async.ts:1031)
(anonymous) (diffEditorViewZones.ts:78)
_deliver (event.ts:1243)
_deliverQueue (event.ts:1254)
fire (event.ts:1278)
(anonymous) (codeEditorWidget.ts:283)
[...]

PR #15189, opened 2025-03-13
Handle visibility changes for diff editors

colin-grant-work:
#### What it does

Fixes https://github.com/eclipse-theia/theia/issues/15178 by adding code to handle both sides of a diff editor when handling visibility change events.

#### How to test

1. Similar to https://github.com/eclipse-theia/theia/pull/14909, but with special attention to diff markers.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-diff-editor.ts
+++ b/packages/monaco/src/browser/monaco-diff-editor.ts
@@ -42,6 +42,9 @@ export namespace MonacoDiffEditor {
 export class MonacoDiffEditor extends MonacoEditor {
     protected _diffEditor: IStandaloneDiffEditor;
     protected _diffNavigator: DiffNavigator;
+    protected savedDiffState: monaco.editor.IDiffEditorViewState | null;
+    protected originalTextModel: monaco.editor.ITextModel;
+    protected modifiedTextModel: monaco.editor.ITextModel;
 
     constructor(
         uri: URI,
@@ -55,6 +58,8 @@ export class MonacoDiffEditor extends MonacoEditor {
         parentEditor?: MonacoEditor
     ) {
         super(uri, modifiedModel, node, services, options, override, parentEditor);
+        this.originalTextModel = originalModel.textEditorModel;
+        this.modifiedTextModel = modifiedModel.textEditorModel;
         this.documents.add(originalModel);
         const original = originalModel.textEditorModel;
         const modified = modifiedModel.textEditorModel;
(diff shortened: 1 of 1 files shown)


packages/monaco/src/browser/monaco-diff-editor.ts
```diff
@@ -123,6 +128,22 @@ export class MonacoDiffEditor extends MonacoEditor {
     override shouldDisplayDirtyDiff(): boolean {
         return false;
     }
+
+    override handleVisibilityChanged(nowVisible: boolean): void {
+        if (nowVisible) {
```
tsmaeder, line 133: What's the reasoning for not calling `super()` somewhere?

packages/monaco/src/browser/monaco-diff-editor.ts
```diff
@@ -123,6 +128,22 @@ export class MonacoDiffEditor extends MonacoEditor {
...
+            this.diffEditor.setModel({original: this.originalTextModel, modified: this.modifiedTextModel});
+            this.diffEditor.restoreViewState(this.savedDiffState);
+            this.diffEditor.focus();
+        } else {
+            const originalModel = this.diffEditor.getOriginalEditor().getModel();
+            if (originalModel) {this.originalTextModel = originalModel; }
```
tsmaeder, line 139: Personally, I find the one-line `if`s very unreadable, I much prefer 
```typescript
if (originalModel) {
   this.originalTextModel = originalModel; 
}
```

but the coding guide does not say anything about that, so 🤷.

tsmaeder (commented):
The code appears to work.

packages/monaco/src/browser/monaco-diff-editor.ts
colin-grant-work: The `super` implementation operates on `this.editor` and `this.model`, and we could likely make that line up with one side or the other of the diff editor, but the only thing we actually have in common is extracting the model and assigning it to a field on the class because the types of the `setModel` call and the save/restore view state calls are different, so the `super` method doesn't do much work that we need to do, and its restoration / storage of one side likely isn't something that we want to do, since it didn't work to preserve the diff markers on the side that it was operating on before.

packages/monaco/src/browser/monaco-diff-editor.ts
colin-grant-work: I like one-liners, but I'm not that invested - three lines it is.

tsmaeder (approved):

tsmaeder:
@colin-grant-work just FYI, a while back, the community settled on leaving additional commits ("addressed review comments") separate until merge time. It just makes it easier to see the delta. Not that it matters much in a small PR like this one.

colin-grant-work:
Since I think I was one of the advocates for that + squash and merge, I should probably be better about actually doing it, as well.

PR #14985, opened 2025-02-19
[TheiaAi] Support referencing prompt fragments via variable

lucas-koehler:
#### What it does

Prompt fragments allow to define parts of prompts in a reusable way. You can embed these prompts then via a variable in the chat or in our prompt templates. This also allows to have a set of prompts available in the chat without always defining a custom agent.

This PR defines a Theia Ai variable `prompt` with an argument containing the prompt fragment's id. The variable is fully resolved including variables and function calls in the fragment.

This also adds auto completion for available custom and builtin prompt fragments to the chat UI.

The variable is defined as a context variable meaning it is added to the chat context similar to referenced files. This was discussed here: https://github.com/eclipse-theia/theia/issues/14899#issuecomment-2655942388

Note that agents need to be adapted to use this from the context as explained here: https://github.com/eclipse-theia/theia/pull/14787#issuecomment-2663510081
This is not part of this PR.

[prompt-template-autocomplete.webm](https://github.com/user-attachments/assets/58ccaa80-2d27-4d11-a3ce-b03319f6096c)

Part of #14899 

#### How to test

- Configure setting `Ai-features › Prompt Templates: Prompt Templates Folder` to point to a folder of your choosing
- Create a `.prompttemplate` file in the specified folder.
- Define a prompt in the file. Ideally, it uses a variable to test that this is resolved. E.g.
  ```
  Convert the timestamp to a human readable date: {{today:inUnixSeconds}}.
  ```
- Open the AI Chat and verify that the autocompletion works (see video in `What id does`)
- Send the request and make sure that your prompt fragment is used. You can use the AI Agent History to see what was sent to the LLM

#### Follow-ups

Step 2 of #14899 implementing settings UI and editing for prompt fragments

#### Breaking changes

- [ ] This PR introduces breaking changes and requires careful review. If yes, the breaking changes section in the [changelog](https://github.com/eclipse-theia/theia/blob/master/CHANGELOG.md) has been updated.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #15196, opened 2025-03-14
Support prompt fragments and variable dependencies with recursive resolving

### LINKED ISSUE #14899: [Theia AI] Support prompt fragments

### LINKED ISSUE #15000: Support AI variable dependencies explicitly
In the course of #14971, we identified the potential need of AI variables that depend on other variables. The concrete example use case was a *chat context summary variable* that lists all variables of the context; to be able to summarize a context variable, it needs to be resolved first. Other use cases may involve reusing core variables, like workspace root, environment, etc.

Currently, resolving a variable in the course of another variable resolution is not explicitly supported; the variable provider can only capture a reference to the variable service when it is asked to register the variable itself, and then use this variable. It cannot get the variable service injected directly, as this would cause cyclic dependency injection. Also we may run into cyclic variable resolutions.

As a potential solution we discussed the approach below.

### Extend the variable resolver

```ts
export interface AIVariableResolverWithVariableDependencies extends AIVariableResolver {
    resolve(
        request: AIVariableResolutionRequest,
        context: AIVariableContext,
        resolveDependency: (req: AIVariableResolutionRequest) => Promise<ResolvedAIVariable | undefined>
    ): Promise<ResolvedAIVariable | undefined>;
}

function isResolverWithDependencies(resolver: AIVariableResolver | undefined): resolver is AIVariableResolverWithVariableDependencies {
    return resolver !== undefined && resolver.resolve.length >= 3;
}
```

### Enhance the resolveVariable method of the variable service

```ts
interface CacheEntry {
    promise: Promise<ResolvedAIVariable | undefined>;
    inProgress: boolean;
}

export class DefaultAIVariableService implements AIVariableService {
    // ... existing properties and methods

    async resolveVariable(
        request: AIVariableArg,
        context: AIVariableContext,
        cache: Map<string, CacheEntry> = new Map()
    ): Promise<ResolvedAIVariable | undefined> {
        const variableName = typeof request === 'string'
            ? request
            : typeof request.variable === 'string'
                ? request.variable
                : request.variable.name;

        if (cache.has(variableName)) {
            const entry = cache.get(variableName)!;
            if (entry.inProgress) {
                this.logger.warn(`Cycle detected for variable: ${variableName}. Skipping resolution.`);
                return undefined;
            }
            return entry.promise;
        }

        const entry: CacheEntry = { promise: Promise.resolve(undefined), inProgress: true };
        cache.set(variableName, entry);

        const promise = (async () => {
            const variable = this.getVariable(variableName);
            if (!variable) {
                return undefined;
            }
            const arg = typeof request === 'string' ? undefined : request.arg;
            const resolver = await this.getResolver(variableName, arg, context);
            let resolved: ResolvedAIVariable | undefined;
            if (isResolverWithDependencies(resolver)) {
                resolved = await resolver.resolve(
                    { variable, arg },
                    context,
                    async (depRequest: AIVariableResolutionRequest) =>
                        this.resolveVariable(depRequest, context, cache)
                );
            } else {
                resolved = await resolver?.resolve({ variable, arg }, context);
            }
            return resolved ? { ...resolved, arg } : undefined;
        })();

        entry.promise = promise;
        promise.finally(() => {
            entry.inProgress = false;
        });

        return promise;
    }
}
```

### Summary

- **AIVariableResolverWithVariableDependencies** lets a resolver access dependencies through a callback during its resolution.
- The updated `resolveVariable` method uses a promise-based cache with an `inProgress` flag for cycle detection. This way, if a variable is already being resolved, it’s skipped—avoiding cycles.
- With this approach, resolved variables are accessible during the resolution of dependent ones.

lucas-koehler:
#### What it does

fixes #14899
fixes #15000 
supersedes #14985

Summary: Support custom prompt fragments via variable `prompt` with prompt id as argument. Prompt fragments may reference each other recursively while the variable service ensures safe cycle handling.

---

Prompt fragments allow to define parts of prompts in a reusable way. You can embed these prompts then via a variable in the chat or in our prompt templates. This also allows to have a set of prompts available in the chat without always defining a custom agent.

This PR defines a Theia Ai variable `prompt` with an argument containing the prompt fragment's id. Corresponding auto completion for available custom and builtin prompt fragments to the chat UI is implemented, too.

Proper usage of prompt fragments - which can reference other variables and prompt fragments - requires two additional features implemented in this PR:

1. Resolving of tool functions in resolved variable text in the prompt service and the chat request parser. Both have been extended to provide this.
2. Recursive resolving of variables including cycle detection. Prompt fragments may compose other prompt fragments. Thus, they need to be resolved recursively. Furthermore, cycles are detected and resolving aborted for the cycle to a void crashing Theia in an endless loop in case users configure a cycle.

[prompt-template-autocomplete.webm](https://github.com/user-attachments/assets/a9829115-567d-4ac1-aeb2-97a5813403c3)

#### How to test

- Configure setting `Ai-features › Prompt Templates: Prompt Templates Folder` to point to a folder of your choosing
- Create one or multiple `.prompttemplate` files in the specified folder.
  - For recursion testing, you need at least two referencing each other
  - Find example prompt templates here: [prompttemplates.zip](https://github.com/user-attachments/files/19251429/prompttemplates.zip). Add them to you prompt templates folder
- Open the AI chat and test using templates there (see video above)
  - for instance use `#prompt:golang` and tell it to generate mergesort. You'll see it generates it in go
  - Use agent `@Universal` which usually does not have access to the workspace. Ask it to explain the contents of a file (use relative path) and use `#prompt:workspace-functions`. Now the agent should read the file. Try again without the variable and see that it will not read the file
- Edit the prompt editor for an agent (e.g. Universal again). Reference prompt templates there via `{{prompt:workspace-functions}}`. Observe that you Universal agent now can access files even when you don't specify the variable in the chat

#### Follow-ups

- https://github.com/eclipse-theia/theia/issues/15202
- https://github.com/eclipse-theia/theia/issues/15226

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@
 
 - [core] fixed version `@types/express` to `^4.17.21` and `@types/express-serve-static-core` to `5.0.4`. This might be required for adopters as well if they run into typing issues. [#15147](https://github.com/eclipse-theia/theia/pull/15147)
 
+<a name="breaking_changes_1.60.0">[Breaking Changes:](#breaking_changes_1.60.0)</a>
+
+- [ai-chat] `ParsedChatRequest.variables` is now `ResolvedAIVariable[]` instead of a `Map<string, AIVariable>` [#15196](https://github.com/eclipse-theia/theia/pull/15196)
+- [ai-chat] `ChatRequestParser.parseChatRequest` is now asynchronous and expects an additional `ChatContext` parameter [#15196](https://github.com/eclipse-theia/theia/pull/15196)
+
 ## 1.59.0 - 02/27/2025
 
 - [ai] added claude sonnet 3.7 to default models [#15023](https://github.com/eclipse-theia/theia/pull/15023)
--- a/packages/ai-chat/src/common/chat-request-parser.spec.ts
+++ b/packages/ai-chat/src/common/chat-request-parser.spec.ts
@@ -18,32 +18,42 @@ import * as sinon from 'sinon';
 import { ChatAgentServiceImpl } from './chat-agent-service';
 import { ChatRequestParserImpl } from './chat-request-parser';
 import { ChatAgentLocation } from './chat-agents';
-import { ChatRequest } from './chat-model';
+import { ChatContext, ChatRequest } from './chat-model';
 import { expect } from 'chai';
-import { DefaultAIVariableService, ToolInvocationRegistry, ToolInvocationRegistryImpl } from '@theia/ai-core';
+import { AIVariable, DefaultAIVariableService, ResolvedAIVariable, ToolInvocationRegistryImpl, ToolRequest } from '@theia/ai-core';
+import { ILogger, Logger } from '@theia/core';
+import { ParsedChatRequestTextPart, ParsedChatRequestVariablePart } from './parsed-chat-request';
 
 describe('ChatRequestParserImpl', () => {
     const chatAgentService = sinon.createStubInstance(ChatAgentServiceImpl);
     const variableService = sinon.createStubInstance(DefaultAIVariableService);
-    const toolInvocationRegistry: ToolInvocationRegistry = sinon.createStubInstance(ToolInvocationRegistryImpl);
-    const parser = new ChatRequestParserImpl(chatAgentService, variableService, toolInvocationRegistry);
+    const toolInvocationRegistry = sinon.createStubInstance(ToolInvocationRegistryImpl);
+    const logger: ILogger = sinon.createStubInstance(Logger);
+    const parser = new ChatRequestParserImpl(chatAgentService, variableService, toolInvocationRegistry, logger);
 
-    it('parses simple text', () => {
+    beforeEach(() => {
+        // Reset our stubs before each test
+        sinon.reset();
+    });
+
+    it('parses simple text', async () => {
         const req: ChatRequest = {
             text: 'What is the best pizza topping?'
         };
-        const result = parser.parseChatRequest(req, ChatAgentLocation.Panel);
+        const context: ChatContext = { variables: [] };
+        const result = await parser.parseChatRequest(req, ChatAgentLocation.Panel, context);
         expect(result.parts).to.deep.contain({
             text: 'What is the best pizza topping?',
             range: { start: 0, endExclusive: 31 }
         });
     });
 
-    it('parses text with variable name', () => {
+    it('parses text with variable name', async () => {
         const req: ChatRequest = {
             text: 'What is the #best pizza topping?'
         };
-        const result = parser.parseChatRequest(req, ChatAgentLocation.Panel);
+        const context: ChatContext = { variables: [] };
+        const result = await parser.parseChatRequest(req, ChatAgentLocation.Panel, context);
         expect(result).to.deep.contain({
             parts: [{
                 text: 'What is the ',
@@ -59,11 +69,12 @@ describe('ChatRequestParserImpl', () => {
         });
     });
 
-    it('parses text with variable name with argument', () => {
+    it('parses text with variable name with argument', async () => {
         const req: ChatRequest = {
             text: 'What is the #best:by-poll pizza topping?'
         };
-        const result = parser.parseChatRequest(req, ChatAgentLocation.Panel);
+        const context: ChatContext = { variables: [] };
+        const result = await parser.parseChatRequest(req, ChatAgentLocation.Panel, context);
         expect(result).to.deep.contain({
             parts: [{
                 text: 'What is the ',
@@ -79,11 +90,12 @@ describe('ChatRequestParserImpl', () => {
         });
     });
 
-    it('parses text with variable name with numeric argument', () => {
+    it('parses text with variable name with numeric argument', async () => {
         const req: ChatRequest = {
             text: '#size-class:2'
         };
-        const result = parser.parseChatRequest(req, ChatAgentLocation.Panel);
+        const context: ChatContext = { variables: [] };
+        const result = await parser.parseChatRequest(req, ChatAgentLocation.Panel, context);
         expect(result.parts[0]).to.contain(
             {
                 variableName: 'size-class',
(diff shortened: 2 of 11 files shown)


JonasHelming (commented):
This works very well for me, thanks!

@planger Could you have a look at the implementation and architecture, please?

Three comments:

1. Message vs System prompt
The prompt is not added to the context as indicated in the video. From my POV this is expected. At the moment, adding a prompt to a chat is basically like pasting a String in a message. If you want to add something to the system prompt, you need to edit the agent prompt or create a custom agent. The question is whether we also want to allow editing the system prompt dynamically with variables. This could be done by adding a second variable that works exactly like the current one, but returns a context variable with a contextValue. @planger WDYT, is this a valuable use case? If so, I would not necessarily add it to this PR, but create a follow-up. However, if we want to add this later, we should now think about the naming.
Maybe
#prompt for the message
#prompt-system for adding it to system
or
#prompt-msg | #prompt-inline
#prompt-system | #prompt-global

2. Auto-completion in the chat
The auto completion is a bit weird if you auto-complete the word "#prompt" as it is now auto completed to "#prompt:". You have to press CTRL+SPACE again then to get auto completion for the ID. For the file variable, the quick picker is used in this case. @planger What would you expect for this variable to happen?

3. Auto-completion in the prompt editor
We should create a follow-up ticket for this please.

lucas-koehler:
Hi @JonasHelming ,
thanks for the feedback!
1. Yes, the prompt fragment is not added to the context. This is intentional as discussed. The video was done before the addition to the context was removed again. Sorry for the confustion
2. - (waiting for comment from @planger 
3. I created issue https://github.com/eclipse-theia/theia/issues/15202 for this

planger (commented):
Thank you for this excellent work @lucas-koehler!

The architecture and code looks great to me. It might be worth letting someone else weigh in on this PR on code-level too, as it is a pretty substantial change.

> 2\. Auto-completion in the chat
>     The auto completion is a bit weird if you auto-complete the word "#prompt" as it is now auto completed to "#prompt:". You have to press CTRL+SPACE again then to get auto completion for the ID. For the file variable, the quick picker is used in this case. @planger What would you expect for this variable to happen?

Yes, I'd agree it would be nicer to continue the suggestions after the user completed `#prompt:` --- either via quick picks or via code completion items. If we prefer code completion items over quick picks, the prompt completion item would need to provide the 'editor.action.triggerSuggest' command in the completion`s `command` property, similar to [here](https://github.com/eclipse-theia/theia/blob/9620ac2d651efb221f86b5b69aebd36e5b47824c/packages/ai-chat-ui/src/browser/chat-view-language-contribution.ts#L85).

This would then trigger the suggestions, equivalent to the user typing Ctrl-space after the `:`.

planger:
> 1. Message vs System prompt
>    The prompt is not added to the context as indicated in the video. From my POV this is expected. At the moment, adding a prompt to a chat is basically like pasting a String in a message. If you want to add something to the system prompt, you need to edit the agent prompt or create a custom agent. The question is whether we also want to allow editing the system prompt dynamically with variables. This could be done by adding a second variable that works exactly like the current one, but returns a context variable with a contextValue. @planger WDYT, is this a valuable use case? If so, I would not necessarily add it to this PR, but create a follow-up. However, if we want to add this later, we should now think about the naming.

To me, the current behavior is as expected. Allowing to add content to the system message via prompt templates inserted into the user input doesn't feel like a much needed use case to me at the moment.

JonasHelming:
@lucas-koehler OK, then we have calrity. I would be in favor of not useing the quick pick item, but keeping the auto completion in the chat input window, but fix the above described ":" case.

lucas-koehler:
https://github.com/eclipse-theia/theia/pull/15196/commits/135e3a61d80c34fd4153f8e86e224b0d6a618808 fixes the wording suggestion by @planger and automatically triggers argument completion after entering the prompt variable.

@planger I used the variable argument picker to trigger the argument autocompletion. However, I only trigger your suggested command there and do not actually pick an argument myself. Please have a brief look whether this solution is okay for you. Using the `command` property of a completion does not work that easily here because the base variable names are provided by the default completion provider that already registers said argument picker command.

[prompt-template-autocomplete-new.webm](https://github.com/user-attachments/assets/9ab05d30-240b-4ae6-99e7-d12889e647cb)

eneufeld (approved):
Code looks good to me.

JonasHelming (approved):
Wonderful! @planger If you could check Lucas comment, we might have a wrap!

planger (approved):
Thank you @lucas-koehler! Looks good to me and works well.

I've created https://github.com/eclipse-theia/theia/issues/15226 to track the removal for the rather awkward workaround to trigger editor suggestions instead of a picker. I think we should make this a more simple option for variable providers.

JonasHelming:
@lucas-koehler Let's merge!

lucas-koehler:
> Thank you @lucas-koehler! Looks good to me and works well.
> 
> I've created #15226 to track the removal for the rather awkward workaround to trigger editor suggestions instead of a picker. I think we should make this a more simple option for variable providers.

Thank you ! That is definitely a good idea to improve eventually. For future reference, I also added it to the list of follow ups in the PR description

PR #15198, opened 2025-03-15
Set correct defaults for max tokens for anthropic models

### LINKED ISSUE #15195: [Theia AI] Increase Anthropics default max token
@eneufeld  Do you remember a reason why we set it to 4096 "only" per default?

JonasHelming:
fixed #15195

#### What it does

- Set max token to the correct default per known anthropic model (see https://docs.anthropic.com/en/docs/about-claude/models/all-models)
- Use 4096 as a default (This is the minimum of all know models and assumed to be OK for future models as well)

#### How to test

Test all existing Anthropic models, set a break point in /packages/ai-anthropic/src/node/anthropic-language-model.ts to check the corretc values are used. There must also be not error. An error occurs if a value is set that is higher than possible for the model.

Check that you can override the  settings with customr equest settings, e.g. the following settings should lead to an error:

    {
        "modelId": "claude-3-7-sonnet-latest",
        "requestSettings": {
            "max_tokens": 64001
        },
        "providerId": "anthropic"
    },

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-anthropic/src/browser/anthropic-frontend-application-contribution.ts
+++ b/packages/ai-anthropic/src/browser/anthropic-frontend-application-contribution.ts
@@ -22,6 +22,14 @@ import { PREFERENCE_NAME_REQUEST_SETTINGS, RequestSetting } from '@theia/ai-core
 
 const ANTHROPIC_PROVIDER_ID = 'anthropic';
 
+// Model-specific maxTokens values
+const DEFAULT_MODEL_MAX_TOKENS: Record<string, number> = {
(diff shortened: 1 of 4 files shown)


eneufeld (commented):
I don't think it is a good idea to handle maxTokens with a complete own way.
Why not just update the DEFAULT_MAX_TOKENS variables?

eneufeld (approved):
As we want to have mandatory parameters separated let's go ahead here

Issue #13901, opened 2024-07-09
Implement proper enablement management in plugin extension


rschnekenbu:
Currently, Theia does not support enabling or disabling extensions, like for example the builtins, or extensions installed by the user. 

Several issues are already created in the backlog, so I opened this one to gather the requirements we may have for such feature.

- #7558
- #13282
- #13876 
- #13897 

A basic start would be to implement the core service for extension management (maybe in the plugin extension) and then use these different issues as input on how to present or drive this service from a user point of view (adding buttons, commands and/or any other ways to configure like profiles or a file white listing or black listing extension in workspace or user space, etc.)

PR #15205, opened 2025-03-17
Add the ability to disable installed plugins

tsmaeder:
#### What it does
This PR lets you disable installed and built-in plugins. The problem is more complicated than one would expect because plugin "installation" and "deployment" are not really separate concerns. Therefore we recompute what is considered an "installed" plugin from the available data in the front end. The PR also contains two renames where the class and file names were not following our usual naming patterns. I've added a simple key-value store for the back-end because preferences are not available in the back end.

Step 1 of #13901

#### How to test
Install, disable, enable, uninstall, etc. both built-ins, recommended and general plugins and make sure the display of the extensions is as expected.

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers


rschnekenbu (commented):
The Pull Request looks very good currently!
I only have a couple of comments on the code itself (usage of the implementation rather than interface).

Also, when I disable a builtin extension, it is shown at the bottom of the list when the window is reloaded. Is this an expected effect? I would assume it should stay in place and only display the 'disabled' status. For example, disabling Git (built-in) puts it at the last place of the builtins

Finally, if I disable an extension, reload the window and then install another version of the extension, I have an error in the console and the reload window button is always shown. That is the only corner case where I could get the updated plugin manager to have an issue otherwise.

packages/plugin-dev/src/node/hosted-plugin-reader.ts
```diff
@@ -30,8 +30,8 @@ export class HostedPluginReader implements BackendApplicationContribution {
 
     private readonly hostedPlugin = new Deferred<PluginMetadata | undefined>();
 
-    @inject(HostedPluginDeployerHandler)
-    protected deployerHandler: HostedPluginDeployerHandler;
+    @inject(PluginDeployerHandlerImpl)
+    protected deployerHandler: PluginDeployerHandlerImpl;
```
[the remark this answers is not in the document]
rschnekenbu: There are a couple of other places in the code where similar comment can be done.

rschnekenbu:
Also, you mentioned in the breaking changes that you "removed the ability of `HostedPluginProcess` to add extra deployed plugins". Was that breaking change necessary for your implementation? I am not familiar with this feature and how it is used by community

packages/vsx-registry/src/browser/vsx-extension.tsx
```diff
@@ -435,32 +491,24 @@ export abstract class AbstractVSXExtensionComponent<Props extends AbstractVSXExt
...
-                <div className="codicon codicon-settings-gear action" onClick={this.manage}></div>
-            </div>;
-        }
-        return <button className="theia-button prominent action" onClick={this.install}>{nls.localizeByDefault('Install')}</button>;
+            {
+                !builtin && ((installed && !uninstalled) ?
+                    <button className="theia-button action" onClick={this.uninstall}>{nls.localizeByDefault('Uninstall')}</button> :
+                    <button className="theia-button prominent action" onClick={this.install}>{nls.localizeByDefault('Install')}</button>)
+            }
+            <div className="codicon codicon-settings-gear action" tabIndex={tabIndex} onClick={this.manage}></div>;
```
msujew, line 510: Bug: There is an additional `;` which gets rendered into the HTML:

image

```suggestion
            <div className="codicon codicon-settings-gear action" tabIndex={tabIndex} onClick={this.manage}></div>
```

packages/vsx-registry/src/browser/vsx-extension.tsx
```diff
@@ -517,6 +565,7 @@ export class VSXExtensionComponent<Props extends VSXExtensionComponent.Props = V
                     </div>
                 </div>
                 <div className='noWrapInfo theia-vsx-extension-description'>{description}</div>
+                {disabled && <div className='disabled'>{nls.localizeByDefault('disabled')}</div>}
```
msujew, line 568: Suggestion: Instead of adding a small `disabled` marker, can we align this to VS Code instead? They just reduce the opacity of the text and icon to indicate that an extension is disabled:

image

msujew (commented):
The code looks good, I have a few complaints about the UI :)

JonasHelming:
@tsmaeder We should mere this in time now

packages/vsx-registry/src/browser/vsx-extension.tsx
tsmaeder: I've thought about this, but I don't think like the UI metaphor VS Code uses: IMO, grayed out style should be used for inactive UI elements, but you can still interact with the item. Grayed out might also mean it's not applicable to the current environment, for example. I don't think it's super clear as a metaphor.

msujew: I see. In that case I would just put it behind the name of the author? Anyway, having the `Disabled` indicator in a new line messes with the layout.

tsmaeder:
@msujew I change the "disabled" notification and also noticed and fixed the problem that extensions were not properly sorted initially.

packages/vsx-registry/src/browser/vsx-extension.tsx
tsmaeder: I put the "disabled" indication on the first line, since the last line has the "reload window", etc. buttons and therefore does not have a lot of space.

msujew (approved):
Thank you, looks good to me 👍

PR #15210, opened 2025-03-18
Fix handling of additionalEventTypes in TreeWidget.addKeyListener

### LINKED ISSUE #15209: `Widget.addKeyListener` additionalEventTypes handling broken
### Steps to Reproduce:

1. Pass additional event types to the `addKeyListener` method of the `TreeWidget`
2. In the handler, refer to a field of what is expected to be a `KeyboardEvent`
3. Get an error that you can't read that field of `undefined`.

### Additional Information

The problem is here:

https://github.com/eclipse-theia/theia/blob/3ab5674c9cbb85b2cf3e33d4fe10f28ca29d7d00/packages/core/src/browser/widgets/widget.ts#L303-L307

`type` is a string; `type['keydown']` is virtually guaranteed to be `undefined`, but the result is then passed into the handler in place of the original event.

- Operating System:
- Theia Version:

colin-grant-work:
#### What it does

Fixes #15209 by passing the original event to the handler and casting the event type as a keyboard event type rather than casting the type as any and reading a field off of it.

#### How to test

1. Pass additional event types to the `addKeyListener` method of the `TreeWidget`
2. In the handler, refer to a field of what is expected to be a `KeyboardEvent`
3. See that your handler behaves as you'd expect.

#### Breaking changes

⚠️  In my view, the current implementation is clearly buggy in two respects: ⚠️ 
 - For events other than `keydown`, it calls the handler with `undefined` rather than the original event. This breaks the type contract of the signature: the `action` is allowed to expect a `KeyboardEvent`.
 - For events other than `keydown`, it calls the handler even if the intended predicate re: keys pressed is not satisfied.

The latter behavior may be intentional, and could be relied on by downstream implementations, since for events other than `keydown`, the behavior of modifiers in particular may be less predictable, depending on order of release, etc. The branch I've pushed changes both behaviors, but we could focus on the former if we believe that changing the latter could disrupt downstream implementations.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/widgets/widget.ts
+++ b/packages/core/src/browser/widgets/widget.ts
@@ -290,7 +290,7 @@ export function addKeyListener<K extends keyof HTMLElementEventMap>(
             return (actual: KeyCode) => KeysOrKeyCodes.toKeyCodes(keysOrKeyCodes).some(k => k.equals(actual));
         }
     })();
-    toDispose.push(addEventListener(element, 'keydown', e => {
+    const wrappedHandler = (e: KeyboardEvent) => {
         const kc = KeyCode.createKeyCode(e);
         if (keyCodePredicate(kc)) {
             const result = action(e);
@@ -299,17 +299,10 @@ export function addKeyListener<K extends keyof HTMLElementEventMap>(
                 e.preventDefault();
             }
         }
-    }));
+    };
+    toDispose.push(addEventListener(element, 'keydown', wrappedHandler));
     for (const type of additionalEventTypes) {
-        toDispose.push(addEventListener(element, type, e => {
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            const event = (type as any)['keydown'];
-            const result = action(event);
-            if (typeof result !== 'boolean' || result) {
-                e.stopPropagation();
-                e.preventDefault();
-            }
-        }));
+        toDispose.push(addEventListener(element, type as 'keydown', wrappedHandler));
     }
     return toDispose;
 }


martin-fleck-at (commented):
Thank you very much @colin-grant-work! I had a look at the change and I'm not sure if this is not actually breaking, as you yourself has suggested. I added the following testing code to the `tree-widget.ts`

```
        this.addKeyListener(this.node, Key.ESCAPE, event => {
            console.log(event.code);
        }, 'contextmenu');
```

On main, this currently executes but as you said for the context menu call, the event is actually undefined causing problems. For the new implementation, the handler is not called at all. It's a bit difficult to guess how much downstream adopters have relied on their handlers being called but I can imagine this being very hard to spot even if we mark it as a breaking change.

What do you think about having an overloaded declaration of `addKeyListener` without the additionalEventTypes where we the action will receive the `KeyboardEvent` (which I assume is the main use case) and another overload with additionalEventTypes where we only provide `UIEvent` (or a proper super interface)? This would break the code where adopters accessed properties from the event that we cannot guarantee but leave most cases intact, I hope.

colin-grant-work:
@martin-fleck-at, thanks for your comments. Looking around the repo, it looks like the main (ab)use of the extra event types is to handle a click along with `Enter`, so it definitely is expected that the handler will be called even when the key predicate is not satisfied. Given that, I've pushed a change in the typing that I think is the minimum to let callers know what they can really expect: now depending on the extra event types passed, the type of the event passed into the handlers will be widened, but if no extra events are passed, it will be restricted to `KeyboardEvent`s.

packages/core/src/browser/widgets/widget.ts
```diff
@@ -302,9 +307,7 @@ export function addKeyListener<K extends keyof HTMLElementEventMap>(
     }));
     for (const type of additionalEventTypes) {
```
martin-fleck-at, line 308: thought: Should we filter 'keydown' from the additionalEventTypes so adopters may not register the same handler twice like
```
        this.addKeyListener(this.node, Key.ESCAPE, event => {
            console.log(event.ctrlKey);
        }, 'keydown');
```

But that would also change the previous behavior so probably we should just keep the current behavior.

martin-fleck-at (approved):
Thank you very much for the quick turnaround and I really like the improvement, great work! I think that greatly improves our support.

Issue #15139, opened 2025-03-06
Prerequisites need to be updated


fipro78:
The [documentation](https://github.com/eclipse-theia/theia/blob/master/doc/Developing.md) says that the prerequisites are **Node.js >= 18.17.0 and < 21**

But trying to run a build created with the latest version of Theia fails with the following message:

```
error node-abi@4.1.0: the engine "node" is incompatible with this module. expected version ">=22.12.0".
```

So it seems a transitive dependency introduced an update in the required Node.js version. 
This should be updated in the docs.

msujew:
@fipro78 I don't think the prerequisites should be updated, as our Electron version still requires us to use 18.17.0 (as that's the node version that comes with our Electron version). Instead, we should probably improve our version constraint for `node-abi`.

msujew:
@fipro78 On second look, how do you even get `node-abi` to that version? We only depend on `electron-rebuild@3.2.9`, but that uses `node-abi": "^3.0.0"`, so version `4.1.0` should not be retrieved by Theia. Can you run `yarn why node-abi` or `npm list node-abi`?

msujew:
Huh, apparently we're using `*` as a version for `node-abi`:

https://github.com/eclipse-theia/theia/blob/afde78f081ef1725a6b48f08425548209f7fd609/dev-packages/application-manager/package.json#L52

I would assume it should reuse the version that is available from `electron-rebuild`, but it does not seem to do that. We should stick to a specific version. cc @tsmaeder.

tsmaeder:
Maybe I'm naive, but I don't find any uses of the `node-abi` package in @theia/application-manager. Why do we have that dependency?

msujew:
Maybe a leftover from some previous work? Anyway, we should probably remove it, if it's not directly in use.

JonasHelming:
This comment might give a hint: https://github.com/eclipse-theia/generator-theia-extension/pull/241#issuecomment-2743081507

PR #15212, opened 2025-03-18
fix: Pin node-abi version to 3.x due to Node.js 22+ requirement in 4.x

fanyipin:
## Description

This PR addresses an issue with the `node-abi` package version. The current version 4.x of `node-abi` requires Node.js version 22 or higher, which is not compatible with our current environment.

To resolve this issue, we are pinning the `node-abi` version to 3.x in the `package.json` file. This ensures compatibility with our existing Node.js versions and avoids potential disruptions.

## Changes

- Updated `package.json` to pin `node-abi` version to `3.x`.

## Motivation and Context

The recent update to `node-abi` version 4.x introduced a requirement for Node.js 22 or higher. Our current project setup uses Node.js versions lower than 22, leading to compatibility issues. By pinning the version to 3.x, we maintain compatibility with our current environment and prevent any disruptions caused by the new version requirement.

## Testing

- Verified that the application builds and runs successfully with the updated `node-abi` version.
- Ensured all existing tests pass without errors.


tsmaeder:
@fanyipin is this related to https://github.com/eclipse-theia/theia/issues/15139?

fanyipin:
> @fanyipin is this related to #15139?

yes, I encountered the same issue

fanyipin:
> Also: why aren't we just removing that package instead of pinning it? Theia builds & runs just fine without it.

I apologize for not paying attention to how the node-abi package is referenced in the project. After reviewing the source code, I realized that the node-abi package is not actually required for application-manager.

fanyipin:
I reviewed the history and found that this [commit](https://github.com/eclipse-theia/theia/commit/82f152d48038c7429c3db4db0cf52fef468d1ca9) is related。Currently, it appears that both node-abi and @types/node-abi can be removed.

tsmaeder:
> I apologize for not paying attention to how the node-abi package is referenced in the project

No need to apologize, that's why we do reviews. Would you update the PR if you agree it's the right fix, though?

fanyipin:
I have removed the unused node-abi from application-manager, and the reference to @theia/cli in playwright also needs to be updated to the current version of the monorepo.

fanyipin:
> @fanyipin Do you mind running `npm i`? The `package-lock.json` needs to be updated as well.

ok, the package-lock.json is updated

tsmaeder (approved):
lgtm.

PR #15219, opened 2025-03-18
[vscode] Support SourceControlResourceGroup optional contextValue

### LINKED ISSUE #15132: [vscode] SourceControlResourceGroup#contextValue added
Since vscode 1.98, SourceControlResourceGroup#contextValue optional string property was added.

rschnekenbu:
#### What it does
Introduces the support of the optional property contextValue in SourceControlResourceGroup API. This was introduced with VS Code 1.98 and is used to filter out some menus on the resource groups.

fix #15132

#### How to test

Install the following extension on a browser or electron theia example:
- src: [scm-resourcegroup-actions-0.0.1-src.zip](https://github.com/user-attachments/files/19316482/scm-resourcegroup-actions-0.0.1-src.zip)
- zip vsix: [scm-resourcegroup-actions-0.0.1.zip](https://github.com/user-attachments/files/19316487/scm-resourcegroup-actions-0.0.1.zip)

It contains a new Software Configuration Manager, based on file names in the workspace (files with name starting with "A_" are considered as added, files names starting with "M_" are modified and "D_" are deleted.

The extension provides also a set of menus to be opened or filtered on resource groups added / deleted / modified:
- one action should be always visible ("on any file")
- one specific action should be visible only on deleted or modified ("On Deleted" and "On Modified")
- the action "On Added" has the test "scmResourceGroupState == added", but the "added" resource group does not have a context value, so we can test the `undefined` case
- the action "Should not be displayed" should never be visible, as the state does not exist on any of the groups

https://github.com/user-attachments/assets/504e41c1-0cf7-4d7b-b2c4-450810e031cd

#### Follow-ups

None

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/common/plugin-api-rpc.ts
+++ b/packages/plugin-ext/src/common/plugin-api-rpc.ts
@@ -1047,6 +1047,7 @@ export interface SourceControlProviderFeatures {
 
 export interface SourceControlGroupFeatures {
     hideWhenEmpty: boolean | undefined;
+    contextValue: string | undefined;
 }
 
 export interface ScmRawResource {
--- a/packages/plugin-ext/src/main/browser/scm-main.ts
+++ b/packages/plugin-ext/src/main/browser/scm-main.ts
@@ -53,6 +53,8 @@ export class PluginScmResourceGroup implements ScmResourceGroup {
 
     get hideWhenEmpty(): boolean { return !!this.features.hideWhenEmpty; }
 
+    get contextValue(): string | undefined { return this.features.contextValue; }
+
     private readonly onDidChangeEmitter = new Emitter<void>();
     readonly onDidChange: Event<void> = this.onDidChangeEmitter.event;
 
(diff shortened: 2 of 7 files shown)


tsmaeder:
Guess one needs to reload the scm view, but then the "git" stuff is gone. Can we only have one scm provider per window?

rschnekenbu:
@tsmaeder, you can have only one repository active at the same time, either the SCM test or the git one. you can select the active one by switching it on the status bar. 

repository-selection

rschnekenbu:
@tsmaeder, fyi, on a base theia, the content of the SCM resource groups will be empty. I get some errors due to this line:
                    handle, sourceUri, letter: (r as any).letter, colorId: (r as any).**color**.id, icons,
color can be undefined. so looking for id can lead to issues. I patched that in this PR

tsmaeder:
@rschnekenbu here's what it looks like for me on electron/windows with a Theia repo open. 

image

tsmaeder (approved):

PR #15233, opened 2025-03-20
feat(mcp-server): add support for SSE transport and enhance server co…

kaikongbj:
This pull request introduces several changes to improve the functionality and flexibility of the MCP server. The most important changes include adding support for SSE transport in the MCP server.

Enhancements to MCP server:

* Added support for SSE transport by introducing a new `TransportType` and updating the `MCPServer` class to handle different transport types (`stdio` and `sse`). This includes modifications to the constructor, `start` method, and `update` method to accommodate the new transport type and SSE URL. [[1]](diffhunk://#diff-baac4f4f84bbd7dd539d2a99f2489830f03476a2a59f924c507724caa7d2850fR17-R47) [[2]](diffhunk://#diff-baac4f4f84bbd7dd539d2a99f2489830f03476a2a59f924c507724caa7d2850fL42-R73) [[3]](diffhunk://#diff-baac4f4f84bbd7dd539d2a99f2489830f03476a2a59f924c507724caa7d2850fL52-R91) [[4]](diffhunk://#diff-baac4f4f84bbd7dd539d2a99f2489830f03476a2a59f924c507724caa7d2850fL97-R142)


JonasHelming:
Thank you for the contribution!

Could you add information on how to test this? I.e. how to set up a (ideally standard node-based) MCP server this way.
Does it also make sense to allow the user to configure such servers or should this be an "API-only" feature?

packages/ai-mcp/src/node/mcp-server.ts
```diff
@@ -13,8 +13,20 @@
...
+import {Client} from '@modelcontextprotocol/sdk/client/index.js';
+
+export type TransportType = 'stdio' | 'sse';
+
+export interface MCPServerConfig {
+    name: string;
+    command?: string;
+    args?: string[];
+    env?: { [key: string]: string };
+    transportType?: TransportType;
```
JonasHelming, line 27: Having the transport Type and the sseUrl as separate variables is maybe not optimal, if you use SSE as a transport type, the url is mandatory. As we currently support two transport types, I would either
1. Only provide the seeURL and if it is set we use SSE
2. Include the transaport specific parameters (env and URL) into a transportType object (with two versions)
The second options seems cleaner to me, as env is ignored for SSE. This way, the API makes it very clear what parameters you can provide

packages/ai-mcp/src/node/mcp-server.ts
```diff
@@ -39,22 +55,40 @@ export class MCPServer {
...
-            ...(this.env || {})
-        };
-        const transport = new StdioClientTransport({
-            command: this.command,
-            args: this.args,
-            env: mergedEnv,
-        });
-        transport.onerror = error => {
+
+        console.log(`Starting server "${this.name}" with transport: ${this.transportType}`);
```
JonasHelming, line 59: We have two messages now instead of one. I would prefer having only one, i.e. augment the specific ones below with a common string from here.

JonasHelming (changes requested):
I did not test this yet, waiting for instructions, left a few minor comments, but this looks pretty good already, thanks!

JonasHelming:
@kaikongbj In order to merge this, we would need you to sign the ECA: https://api.eclipse.org/git/eca/status/gh/eclipse-theia/theia/15233

PR #15236, opened 2025-03-20
Gh 15200

JonasHelming:
#### What it does

Introduces three new settings:

- Workspace relative Directories to load prompt templates from
- File extensions to consider as prompt templates (in the workspace directories and the global user prompt directory)
- A list of workspace relative paths to files to consider as prompt templates

All these prompt templates are then available via #prompt:filename (without file extension)

In case of conflicting ids, we prefer workspace files => workspace directories => global directory
In case of conflicting ids within workspace directories, we log a warning and ignore the secondly loaded template

#### How to test

- Create and delete prompt templates in all three possible locations (and add them in the settings)
- Hint: Add "say "directory "" or "say "from global"" as content. This way the universal agent will respond e.g. with "from global" if you just send a message #prompt:file. Alternatively you can watch the output view to verify the right result
- Check that conflicts are resolved correctly and that changing, deleting, adding files and settings updates correctly
- Use #prompt in the chat and in the prompt editor of an agent
- Validate the existing prompt editing for regular agents and the variant management

#### Follow-ups

We might add #prompt:AI-README.md as a default to some agents, e.g. Coder or Architect and point a files settings to some default location.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


packages/ai-core/src/browser/frontend-prompt-customization-service.ts
```diff
@@ -154,23 +460,29 @@ export class FrontendPromptCustomizationServiceImpl implements PromptCustomizati
...
-        if (filename.endsWith(suffix)) {
-            return filename.slice(0, -suffix.length);
+        for (const ext of this.templateExtensions) {
+            if (filename.endsWith(ext)) {
+                return filename.slice(0, -ext.length);
+            }
+        }
+        // If no matching extension found, try the default one
+        if (filename.endsWith(PROMPT_TEMPLATE_EXTENSION)) {
+            return filename.slice(0, -PROMPT_TEMPLATE_EXTENSION.length);
```
sdirix, line 470: It's inconsistent to hard code this value here, but not for all other cases. We should either hard code the `.prompttemplate` extension for all use cases, or remove this code here.

packages/ai-ide/src/browser/workspace-preferences.ts
```diff
@@ -38,6 +41,39 @@ export const WorkspacePreferencesSchema: PreferenceSchema = {
...
+            default: [],
+            items: {
+                type: 'string'
+            }
+        },
+        [PROMPT_TEMPLATE_WORKSPACE_EXTENSIONS_PREF]: {
+            type: 'array',
+            title: nls.localize('theia/ai/promptTemplates/extensions/title', 'Workspace-specific Prompt Template File Extensions'),
+            description: nls.localize('theia/ai/promptTemplates/extensions/description',
+                'List of file extensions in workspace-specific locations that are considered as prompt templates. When templates with the same ID exist in multiple locations, \
```
sdirix, line 60: The priority explanation is unrelated to the extensions preference and should rather be mentioned in the other 3 preferences.

sdirix, line 60: It's mainly used for file extensions but our code actually does not care at all that these are file extensions. We just do an `endsWith` check. I like the code as it's more generic as just forcing extensions, however technically the preference description is too specific.

packages/ai-core/src/browser/frontend-prompt-customization-service.ts
```diff
@@ -72,55 +104,259 @@ export class FrontendPromptCustomizationServiceImpl implements PromptCustomizati
...
+     * @param sourceUri The URI of the source file (used to distinguish updates from conflicts)
+     * @param loadedTemplates The map to track all loaded templates
+     */
+    protected addTemplate(
+        templatesMap: Map<string, TemplateEntry>,
+        id: string,
+        content: string,
+        priority: number,
+        sourceUri: string,
+        loadedTemplates: Map<string, TemplateEntry> = this.allLoadedTemplates
```
sdirix, line 150: This should not be an optional parameter. If we refer to `this.allLoadedTemplates` while loading, we might run into race conditions (interleaving updates) again.

packages/ai-core/src/browser/frontend-prompt-customization-service.ts
```diff
@@ -72,55 +104,259 @@ export class FrontendPromptCustomizationServiceImpl implements PromptCustomizati
...
+
+    /**
+     * Removes a template from the templates map based on the source URI and fires change event.
+     * Also checks for any lower-priority templates with the same ID that might need to be loaded.
+     * @param sourceUri The URI of the source file being removed
+     * @param loadedTemplates The map of all loaded templates
+     */
+    protected removeTemplateBySourceUri(
+        sourceUri: string,
+        loadedTemplates: Map<string, TemplateEntry> = this.allLoadedTemplates
```
sdirix, line 191: Same here, referring to the global is not alloweed

packages/ai-core/src/browser/frontend-prompt-customization-service.ts
```diff
@@ -72,55 +104,259 @@ export class FrontendPromptCustomizationServiceImpl implements PromptCustomizati
...
-        this.toDispose.push(this.fileService.watch(templateURI, { recursive: true, excludes: [] }));
+        this.toDispose.push(this.fileService.onDidFilesChange(async (event: FileChangesEvent) => {
+
+            // Handle deleted files
+            for (const deletedFile of event.getDeleted()) {
+                const fileUriString = deletedFile.resource.toString();
+                const fileInfo = this.watchedFiles.get(fileUriString);
+
+                if (fileInfo) {
+                    this.removeTemplateBySourceUri(fileUriString, this.allLoadedTemplates);
```
sdirix, line 253: Again referring to the global one

packages/ai-core/src/browser/frontend-prompt-customization-service.ts
```diff
@@ -72,55 +104,259 @@ export class FrontendPromptCustomizationServiceImpl implements PromptCustomizati
...
+
+                if (fileInfo) {
+                    const fileContent = await this.fileService.read(fileInfo.uri);
+                    this.addTemplate(
+                        this.effectiveTemplates,
+                        fileInfo.templateId,
+                        fileContent.value,
+                        TemplatePriority.TEMPLATE_FILE,
+                        fileUriString,
+                        this.allLoadedTemplates
```
sdirix, line 272: Again referring to the global one

packages/ai-core/src/browser/frontend-prompt-customization-service.ts
```diff
@@ -72,55 +104,259 @@ export class FrontendPromptCustomizationServiceImpl implements PromptCustomizati
...
```
sdirix, line 291: Again referring to the global one

packages/ai-core/src/browser/frontend-prompt-customization-service.ts
```diff
@@ -72,55 +104,259 @@ export class FrontendPromptCustomizationServiceImpl implements PromptCustomizati
...
                 this.onDidChangeCustomAgentsEmitter.fire();
             }
+
             // check deleted templates
             for (const deletedFile of event.getDeleted()) {
                 if (this.trackedTemplateURIs.has(deletedFile.resource.toString())) {
                     this.trackedTemplateURIs.delete(deletedFile.resource.toString());
                     const templateId = this.removePromptTemplateSuffix(deletedFile.resource.path.name);
-                    _templates.delete(templateId);
+                    this.removeTemplateBySourceUri(deletedFile.resource.toString(), this.allLoadedTemplates);
```
sdirix, line 328: Again referring to the global one

packages/ai-core/src/browser/frontend-prompt-customization-service.ts
```diff
@@ -72,55 +104,259 @@ export class FrontendPromptCustomizationServiceImpl implements PromptCustomizati
...
                 }
             }
+
             // check updated templates
             for (const updatedFile of event.getUpdated()) {
                 if (this.trackedTemplateURIs.has(updatedFile.resource.toString())) {
                     const fileContent = await this.fileService.read(updatedFile.resource);
                     const templateId = this.removePromptTemplateSuffix(updatedFile.resource.path.name);
-                    _templates.set(templateId, fileContent.value);
+                    this.addTemplate(this.effectiveTemplates, templateId, fileContent.value, priority, updatedFile.resource.toString(), this.allLoadedTemplates);
```
sdirix, line 338: Again referring to the global one

packages/ai-core/src/browser/frontend-prompt-customization-service.ts
```diff
@@ -72,55 +104,259 @@ export class FrontendPromptCustomizationServiceImpl implements PromptCustomizati
...
             // check new templates
             for (const addedFile of event.getAdded()) {
-                if (addedFile.resource.parent.toString() === templateURI.toString() && addedFile.resource.path.ext === PROMPT_TEMPLATE_EXTENSION) {
+                if (addedFile.resource.parent.toString() === dirURI.toString() &&
+                    this.isPromptTemplateExtension(addedFile.resource.path.ext)) {
                     this.trackedTemplateURIs.add(addedFile.resource.toString());
                     const fileContent = await this.fileService.read(addedFile.resource);
                     const templateId = this.removePromptTemplateSuffix(addedFile.resource.path.name);
-                    _templates.set(templateId, fileContent.value);
+                    this.addTemplate(this.effectiveTemplates, templateId, fileContent.value, priority, addedFile.resource.toString(), this.allLoadedTemplates);
```
sdirix, line 350: Again referring to the global one

packages/ai-ide/src/browser/template-preference-contribution.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    @inject(WorkspaceService)
+    protected readonly workspaceService: WorkspaceService;
+
+    onStart(): void {
+        Promise.all([this.preferenceService.ready, this.workspaceService.ready]).then(() => {
+            // Set initial template directories, extensions, and files from preferences
+            this.updateTemplateDirectories();
+            this.updateTemplateFileExtensions();
+            this.updateTemplateFiles();
```
sdirix, line 45: All three calls will each lead to a full rescan of all files, disposing and recreation of file watchers etc. We should think about an (re)-initialize call in the service to update all 3 data points (actually 4, including the preference which is currently managed by `ai-core-preferences`) at the same time to avoid this. Same for the workspace root changes below.

packages/ai-ide/src/browser/template-preference-contribution.ts
```diff
@@ -0,0 +1,118 @@
...
+import {
+    PROMPT_TEMPLATE_WORKSPACE_DIRECTORIES_PREF,
+    PROMPT_TEMPLATE_WORKSPACE_EXTENSIONS_PREF,
+    PROMPT_TEMPLATE_WORKSPACE_FILES_PREF
+} from './workspace-preferences';
+import { WorkspaceService } from '@theia/workspace/lib/browser';
+import { Path } from '@theia/core';
+
+@injectable()
+export class TemplatePreferenceContribution implements FrontendApplicationContribution {
```
sdirix, line 29: This contribution should also manage the other template customization directory preference which is currently part of the `ai-core-preferences`.

sdirix (changes requested):
It works for me! :heart: 

Please have a look at my comments

## UX improvements

With the new workspace related directories, we are able to override existing prompts and existing prompt customizations. However it's very intransparent that this is happening. The AI configuration view should indicate this and ideally even allow to open an editor for the currently "active" prompt file. Can of course be handled in a follow up.

When I want to override the default prompt of an agent, it's difficult to determine which key I need to use as they are only shown as `default` in the UI. Only by editing them, I can see their key in the customization file name. We should show the key of the prompts in the configuration view and just add a (default) suffix for the default ones.  Can of course be handled in a follow up.

packages/ai-ide/src/browser/template-preference-contribution.ts
sdirix: Ah I didn't realize that this was only ide-specific. Is there a reason for this?

JonasHelming: I would not put it into a core package, as it enables the user to configure these directories. Adopters might not want this in other tools. So I claim the preferences are tool/IDE specific

packages/ai-core/src/browser/frontend-prompt-customization-service.ts
JonasHelming: I made sure now, that the default .prompttemplate is always in this list.

packages/ai-ide/src/browser/workspace-preferences.ts
```diff
@@ -38,6 +41,39 @@ export const WorkspacePreferencesSchema: PreferenceSchema = {
...
+                have higher priority than global templates but lower priority than specific template files.'),
+            default: [],
+            items: {
+                type: 'string'
+            }
+        },
+        [PROMPT_TEMPLATE_WORKSPACE_EXTENSIONS_PREF]: {
+            type: 'array',
+            title: nls.localize('theia/ai/promptTemplates/extensions/title', 'Workspace-specific Prompt Template File Extensions'),
+            description: nls.localize('theia/ai/promptTemplates/extensions/description',
```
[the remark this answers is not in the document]
JonasHelming: Agreed, I make sure the .prompttemplate is always there as a default in the frontend-prompt-customization-service and renamed this preference

packages/ai-ide/src/browser/template-preference-contribution.ts
```diff
@@ -0,0 +1,118 @@
...
+            return uri.path.toString();
+        });
+
+        // Update the template directories in the customization service
+        await this.customizationService.updateTemplateDirectories(absoluteDirectories);
+    }
+
+    protected async updateTemplateFileExtensions(): Promise<void> {
+        // Get the template file extensions from preferences
+        const extensions = this.preferenceService.get<string[]>(PROMPT_TEMPLATE_WORKSPACE_EXTENSIONS_PREF, ['.prompttemplate']);
```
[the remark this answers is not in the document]
JonasHelming: This is removed now, as .prompttemplate is now a default in the frontend-prompt-customization-service

packages/ai-ide/src/browser/workspace-preferences.ts
JonasHelming: I don't think so, we check for the extension:

```
 /**
     * Checks if the given file extension is registered as a prompt template extension
     * @param extension The file extension including the leading dot (e.g., '.prompttemplate')
     * @returns True if the extension is registered as a prompt template extension
     */
    protected isPromptTemplateExtension(extension: string): boolean {
        return this.templateExtensions.has(extension);
    }
```

And the callers:

`this.isPromptTemplateExtension(fileURI.path.ext)`

JonasHelming: adapted

packages/ai-ide/src/browser/template-preference-contribution.ts
JonasHelming: This makes sense. Basically I would add an initialize method to the frontend-prompt-customization-service where I can hand in all three parameters at once and only update once after this.
Does this make sense to you?

As the ai-core-preference comes from a different contribution, I am not sure, whether we can avoid this duplication, though?

packages/ai-ide/src/browser/workspace-preferences.ts
```diff
@@ -38,6 +41,39 @@ export const WorkspacePreferencesSchema: PreferenceSchema = {
...
                 type: 'string'
             }
+        },
+        [PROMPT_TEMPLATE_WORKSPACE_DIRECTORIES_PREF]: {
+            type: 'array',
+            title: nls.localize('theia/ai/promptTemplates/directories/title', 'Workspace-specific Prompt Template Directories'),
+            description: nls.localize('theia/ai/promptTemplates/directories/description',
+                'List of relative paths indicating folders in the current workspace to be scanned for WORKSPACE specific prompt templates. Templates from workspace directories \
+                have higher priority than global templates but lower priority than specific template files.'),
+            default: [],
```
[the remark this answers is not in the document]
JonasHelming: Yes, this makes sense. I would do it in a follow-up though. I would like to have this out for a field test before we add (potentially) arbitrary files the user is not aware of. Added a follow-up here: https://github.com/eclipse-theia/theia/issues/15068

packages/ai-ide/src/browser/workspace-preferences.ts
```diff
@@ -38,6 +41,39 @@ export const WorkspacePreferencesSchema: PreferenceSchema = {
...
+                type: 'string'
+            }
+        },
+        [PROMPT_TEMPLATE_WORKSPACE_EXTENSIONS_PREF]: {
+            type: 'array',
+            title: nls.localize('theia/ai/promptTemplates/extensions/title', 'Workspace-specific Prompt Template File Extensions'),
+            description: nls.localize('theia/ai/promptTemplates/extensions/description',
+                'List of file extensions in workspace-specific locations that are considered as prompt templates. When templates with the same ID exist in multiple locations, \
+                conflicts are resolved by priority: specific template files (highest) > workspace directories > global directories (lowest).'),
+            default: ['.prompttemplates'],
```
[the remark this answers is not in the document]
JonasHelming: Yes, this makes sense. I would do it in a follow-up though. I would like to have this out for a field test before we add (potentially) arbitrary files the user is not aware of. Added a follow-up here: https://github.com/eclipse-theia/theia/issues/15068

JonasHelming:
> With the new workspace related directories, we are able to override existing prompts and existing prompt customizations. However it's very intransparent that this is happening. The AI configuration view should indicate this and ideally even allow to open an editor for the currently "active" prompt file. Can of course be handled in a follow up.
> 
> When I want to override the default prompt of an agent, it's difficult to determine which key I need to use as they are only shown as `default` in the UI. Only by editing them, I can see their key in the customization file name. We should show the key of the prompts in the configuration view and just add a (default) suffix for the default ones. Can of course be handled in a follow up.

Very valid points, added a follow-up: https://github.com/eclipse-theia/theia/issues/15319

JonasHelming:
@sdirix Thank you for the great review and feedback! I only have one open question above. then I can adress the last review comment and retest everything once more.

packages/ai-ide/src/browser/template-preference-contribution.ts
sdirix: You're right, I think it becomes too clunky to avoid the duplication there.

But also it would be weird to have a `initialize` method which only takes 3 of the 4 parameters.

Suggestion: Declare a single `update` method in the service which takes an object as a parameter, with an optional entry for the properties to update. Then it can set all of the handed over parameters at once, performing the internal update also once. This makes the service interface nicer.

Of course we still have the duplicated effort then, with the update once called with one parameter and then again with 3 parameters, but I think we can live with that for now.

packages/ai-ide/src/browser/workspace-preferences.ts
sdirix: I overlooked that, I just saw the `endsWith` in the file extension removal. :+1:

packages/ai-ide/src/browser/template-preference-contribution.ts
JonasHelming: I have now introduced a update function for the three external preferences. Did you mean to also include the 4th (internal) preference? If so, we should extract this preference from the service and set it from a frontendApplicationContribution in ai-core. Please quickly confirm if you like this or if we should keep the preference for the global directory as is (for now) @sdirix

sdirix (approved):
Works for me! Thanks

PR #15238, opened 2025-03-20
Ensure valid editor constructed before attempting to use editor services

### LINKED ISSUE #15102: In-line Git diffs stop working after the first interaction
### Bug Description:

After the first time clicking on a Git diff marker in the gutter to show it (maybe even revert it), any further clicking on similar markers has no effect. No other Git diffs are shown.

### Steps to Reproduce:

1. Edit text in some TypeScript files in a Git repository.
2. Click on a difference marker in the gutter to show the difference.
3. Revert or not, it doesn't matter. It's enough to bring up the difference viewer pop-up.
4. Now click on any other difference marker.
5. Observe that nothing happens.

### Additional Information

- Operating System: macOS Sequoia 15.3.1
- Theia Version: 1.58.300.206

colin-grant-work:
#### What it does

Fixes #15102.

The bug is a bit of an odd one. Basically, when a `StandaloneCodeEditor` is constructed, it [sets](https://github.com/microsoft/vscode/blob/9f4e94e38442c0ca13b8ab4a94cd1e9948c44180/src/vs/editor/standalone/browser/standaloneCodeEditor.ts#L296) a [global value](https://github.com/microsoft/vscode/blob/9f4e94e38442c0ca13b8ab4a94cd1e9948c44180/src/vs/base/browser/ui/hover/hoverDelegateFactory.ts#L19-L22) for a hover delegate factory. That is then used ([indirectly](https://github.com/microsoft/vscode/blob/9f4e94e38442c0ca13b8ab4a94cd1e9948c44180/src/vs/base/browser/ui/hover/hoverDelegateFactory.ts#L33-L38)) by the [action bar](https://github.com/microsoft/vscode/blob/9f4e94e38442c0ca13b8ab4a94cd1e9948c44180/src/vs/base/browser/ui/actionbar/actionbar.ts#L121) in  [peek widgets](https://github.com/microsoft/vscode/blob/9f4e94e38442c0ca13b8ab4a94cd1e9948c44180/src/vs/editor/contrib/peekView/browser/peekView.ts#L199) if no hover delegate is specified in the options.

In the code for the `DirtyDiffPeekView`, the header for the view was getting constructed before the editor for a given view had been constructed, meaning that if the last thing to happen was the disposal of an editor with its local instantiation service, the attempt to create a hover delegate used the global value that referred to a now defunct instantiation service.

This PR therefore adjusts the lifecycle of the peek view to construct the editor before the rest of the code to build the view is run, ensuring that there is a valid instantiation service available to construct the hover delegate.

#### How to test

1. Make uncommitted changes to a file.
2. Open that file and scroll to your changes.
3. Click the marginal marker indicating the changed lines.
4. A diff peek view should appear.
5. Close it.
6. Open another by clicking again.
7. It should also open, you should see no errors in the console, and other Monaco services (quick picks, etc.) should also still work.

> In addition to the original issue reported in #15102, I occasionally observed issues with other Monaco services, but they were much less consistently reproducible than the reported issue.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-editor-peek-view-widget.ts
+++ b/packages/monaco/src/browser/monaco-editor-peek-view-widget.ts
@@ -90,6 +90,14 @@ export class MonacoEditorPeekViewWidget {
                 return this._actionbarWidget;
             }
 
+            fillContainer(container: HTMLElement): void {
+                super._fillContainer(container);
+            }
+
+            protected override _fillContainer(container: HTMLElement): void {
+                that.fillContainer(container);
+            }
+
             fillHead(container: HTMLElement, noCloseAction?: boolean): void {
                 super._fillHead(container, noCloseAction);
             }
@@ -136,6 +144,26 @@ export class MonacoEditorPeekViewWidget {
 
             protected override revealRange(range: monaco.Range, isLastLine: boolean): void {
                 that.doRevealRange(that.editor['m2p'].asRange(range), isLastLine);
+            }
+
+            getBodyElement(): HTMLDivElement | undefined {
+                return this._bodyElement;
+            }
+
+            setBodyElement(element: HTMLDivElement | undefined): void {
+                this._bodyElement = element;
+            }
+
+            getHeadElement(): HTMLDivElement | undefined {
+                return this._headElement;
+            }
+
+            setHeadElement(element: HTMLDivElement | undefined): void {
+                this._headElement = element;
+            }
+
+            override setCssClass(className: string, classToReplace?: string | undefined): void {
+                super.setCssClass(className, classToReplace);
             }
         }(
             editor.getControl() as unknown as ICodeEditor,
@@ -185,6 +213,10 @@ export class MonacoEditorPeekViewWidget {
         return action;
     }
 
+    protected fillContainer(container: HTMLElement): void {
+        this.delegate.fillContainer(container)
+    }
+
     protected fillHead(container: HTMLElement, noCloseAction?: boolean): void {
         this.delegate.fillHead(container, noCloseAction);
     }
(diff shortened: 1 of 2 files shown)


sdirix (approved):
I can confirm that it works much better now. Before this PR, the diff preview often did not even open. With the PR there are no errors in the console and the preview opens consistently.

---

However there is still one issue left. Feel free to tackle it here or to create a follow up:

1. Open an editor
2. Modify a line of code
3. Click as fast as possible on the marker to open the diff preview
4. The preview will either not open or it will open and close again immediately.

When waiting for a second after step 2, the issue does not occur, so there is some timing issue going on.

---

Still it's much better than before

colin-grant-work:
I'll take a look at the failure of opening immediately after a change is made; if I can find the cause, I'll fix it here.

colin-grant-work:
@sdirix, I believe I've corrected the issue you observed. It seems we had code that would close the editor whenever the diff what changed, and if you clicked quickly enough, the signal for (part of) your change might come after your click. I have adjusted the behavior to be closer to the way it seems VSCode organizes things: the editor is closed if you execute an action other than navigation successfully, but not on every change.

sdirix:
The described error case now works better than before. It's also fun to continue typing with the diff preview live updating. I like this variant better than before and it's also more close to VS Code behavior.

However this exposes a new usability issue: When a diff preview is open and the user edits somewhere else in the file, producing a new diff location, then the diff editor is not updating. Also the new location will not open a diff preview until the previous one is closed.

To reproduce:
1. Create two "diffs" in a file
2. Open diff preview for one of them
3. Notice that the preview shows `1/2 of 2 changes`
4. While keeping the preview open, edit somewhere else in the file
5. Notice that the preview still shows `1/2 of 2 changes`
6. Try to open the new diff in the diff preview. Observe that it doesn't work
7. Close the diff preview, now the new location works and shows `1/2/3 of 3 changes`.

Note that this works in VS Code.

sdirix (changes requested):
With the latest changes, opening a diff editor, editing further text and opening a new diff preview is possible.

However there are some other issues:
1. Edit text and open a diff preview
2. Keep diff preview open and edit further text
3. Observe that the existing diff preview does not update the number of changes, e.g. if before it showed 1 of 3 changes, then it will still show that although a new change is present in the file
4. Observe that clicking" show next change" and "show previous change" does not close and open the offer diff previews, instead it shows the change in the existing diff preview which is unintuitive
5. When reverting the code, for example via undo, open diff editors stay open and need to be manually closed.

colin-grant-work:
I was able to reproduce the failure of Monaco services on this branch with the right sequence of editor openings and closings, so it looks like there are still some cases where a disposed instantiation service can leak - I will take a look at the details.

sdirix (approved):
Very nice. I was not able to provoke the errors and all the diff use cases work great now. Thanks!

PR #15252, opened 2025-03-22
Add shortcut for adding the current file to the AI chat context

JonasHelming:
#### What it does

Adds a shortcut "_f" to the variable currentRelativeFilePath

#### How to test

Use #_f in the AI chat and see that it behaves the same as #currentRelativeFilePath

I am open to debate on the format, but I researched a bit and this was the best option I could come up with.
I think a prefix for shortcuts makes sense and things like "%" or "$" or "." are either not allowed in variable names or look really weird in combination with "#"

#### Follow-ups

we might want to create other shortcuts, e.g. for the selection?

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


planger (approved):
Works fine and code looks good! I think the `_` prefix is fine. Thank you!

PR #15257, opened 2025-03-24
add workarounds for gemini openai usage

### LINKED ISSUE #15247: [Theia AI] Support Gemini API
This might not be necessary, as Google supports the OpenAI API. I tested it with adding, e.g.:

    {
        "model": "gemini-2.0-pro-exp-02-05",
        "url": "https://generativelanguage.googleapis.com/v1beta/openai/",
        "apiKey": "key",
        "enableStreaming": true,
        "developerMessageSettings": "developer"
    }

and i was able to chat.

However, tool calls do not work. The tool gets called but then it hangs forever. @eneufeld Do you have any idea?
You can create an API key with free tokens (rate limited) here: https://aistudio.google.com/

eneufeld:
#### What it does

Add some code to handle incorrect/unexpected behavior of gemini when using the openai endpoint.
Update OpenAI client Library

fixes #15247

#### How to test

See the linked issue on how to test.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21368,36 +21368,6 @@
         "yjs": "^13.0.0"
       }
     },
-    "node_modules/openai": {
-      "version": "4.86.1",
-      "resolved": "https://registry.npmjs.org/openai/-/openai-4.86.1.tgz",
-      "integrity": "sha512-x3iCLyaC3yegFVZaxOmrYJjitKxZ9hpVbLi+ZlT5UHuHTMlEQEbKXkGOM78z9qm2T5GF+XRUZCP2/aV4UPFPJQ==",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@types/node": "^18.11.18",
-        "@types/node-fetch": "^2.6.4",
-        "abort-controller": "^3.0.0",
-        "agentkeepalive": "^4.2.1",
-        "form-data-encoder": "1.7.2",
-        "formdata-node": "^4.3.2",
-        "node-fetch": "^2.6.7"
-      },
-      "bin": {
-        "openai": "bin/cli"
-      },
-      "peerDependencies": {
-        "ws": "^8.18.0",
-        "zod": "^3.23.8"
-      },
-      "peerDependenciesMeta": {
-        "ws": {
-          "optional": true
-        },
-        "zod": {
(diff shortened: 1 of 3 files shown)


JonasHelming:
Great, I will test this later? There are build failures, though.
And: Do you think it would make sense to only activate these work arounds based on a model setting? Similarily like we did for the Azure Compatibility with the developerMessage setting? https://theia-ide.org/docs/user_ai/#openai-compatible-models-eg-via-vllm

eneufeld:
I'll check the build.
To enable the workaround based on some property would require to introduce a new property. we could check for model name or something, but i'm not sure.

JonasHelming (approved):
Works for me, the work arounds should have no effect if the provider behaves correctly, so I think we can add it without any property
I also tested OpenAI native without any regressions.

JonasHelming (commented):
Found another issue unfortunately, but I cannot reproduce it now.

image

JonasHelming:
Works fine for me now. I think this is good enough to kick off a field test (before officially anouncing it). @sdirix Are you fine with the work arounds?

sdirix:
I'm fine with the changes if they don't break for non-Gemini cases. Also would be interesting whether this breaks in case the Gemini side gets fixed on the server?

The code is pretty adventurous, e.g.
```ts
const toolCall = (choice.message.tool_calls as any).undefined;
```

Maybe it's worth it to hide this between a Gemini workaround toggle?

---

If we can point to a Gemini issue somewhere then it would be nice to add this to the comment in the code.

JonasHelming:
@sdirix There was a discussion here: https://github.com/eclipse-theia/theia/issues/15247
So you would be in favor of introdcuing a property? Eugen's point was that due to the leading if statements, the work arounds would not even apply if everything is returned correctly. Introducing a property is almost a no-op, too. We need a judgement call

sdirix:
If the if cases only apply for invalid cases then that's fine and no additional property is required.

eneufeld:
@JonasHelming  should we still merge this although we now have a dedicated google provider?
I would suggest to cancel this.

JonasHelming:
yes, agreed, cancel

PR #15261, opened 2025-03-24
Add 'plus' key support

KR155E:
#### What it does

Despite being mentioned in the keymaps package's readme, the "plus" key is unavailable for key bindings. This PR adds the necessary const to support it on a German QWERTZ keyboard.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/keys.ts
+++ b/packages/core/src/common/keys.ts
@@ -609,6 +609,7 @@ export namespace Key {
     export const F24: Key = { code: 'F24', keyCode: 135, easyString: 'f24' };
 
     export const NUM_LOCK: Key = { code: 'NumLock', keyCode: 144, easyString: 'numlock' };
+    export const PLUS: Key = { code: 'Plus', keyCode: 171, easyString: 'plus' };
     export const SEMICOLON: Key = { code: 'Semicolon', keyCode: 186, easyString: ';' };
     export const EQUAL: Key = { code: 'Equal', keyCode: 187, easyString: '=' };
     export const COMMA: Key = { code: 'Comma', keyCode: 188, easyString: ',' };


msujew (commented):
@KR155E Can you add a `How to test` section? I've tried binding a command to `Alt+Plus`, but that didn't seem to work (I'm on QWERTZ).

KR155E:
I am sorry, I have been unable to build Theia for a while and don't have the time to dive deeper into the problem. What I did was to hack the key constant into the bundled Theia (1.57.0) in my downstream IDE, then backport that constant. It did work in my working environment when I tried assigning simply "plus" as the keybinding for a custom command.

JonasHelming:
@KR155E Hi, do you find the time to get back to this? Otherwise, we would archive this PR for now.

KR155E:
@JonasHelming @msujew Sorry for the delay. I was able to check this PR today and remapped "Zoom In" to "ctrlcmd+plus", which worked just fine.

ndoschek:
Thanks for the contribution @KR155E! I tested this on both QWERTY and QWERTZ and found that it only works in Electron, not in the browser (on either layout, including Firefox).

The issue is that `code: 'Plus'` is not a standard [W3C UI Events KeyboardEvent code](https://www.w3.org/TR/uievents-code/#key-alphanumeric-writing-system), so browser keyboard events never produce it. The `KeyboardLayoutService` maps `+` to `{ key: Key.EQUAL, shift: true }` ([keyboard-layout-service.ts:260](https://github.com/eclipse-theia/theia/blob/master/packages/core/src/browser/keyboard/keyboard-layout-service.ts#L260)), so keybindings using `plus` never match in the browser. VS Code also does not support `plus` as a key, since `+` is already reachable via `shift+=` (QWERTY) or layout-mapped physical keys.

Thanks again for taking the time to look into this! Feel free to reopen if you disagree.

PR #15142, opened 2025-03-06
Improve plugin package localization

msujew:
#### What it does

Closes https://github.com/eclipse-theia/theia/issues/15093

Prevents the issue from appearing by no longer using a regex at all.

#### How to test

See reproduction steps of https://github.com/eclipse-theia/theia/issues/15093.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #15268, opened 2025-03-24
Prevent plugin localization errors

msujew:
#### What it does

After https://github.com/eclipse-theia/theia/pull/15142, the backend became to eager trying to localize values in the package.json. This led to errors in the output console during the translation process. This PR fixes these errors.

#### How to test

1. Current Theia prints errors such as: `2025-03-24T21:59:06.485Z root ERROR Failed to localize plugin 'vscode.r'. TypeError: Cannot read properties of undefined (reading 'replace')`
2. After applying this patch, these errors should no longer appear.

#### Review checklist

#### Reminder for reviewers


sdirix (approved):
Works for me

msujew:
> It didn't remove the checks which we are now adding.

Correct, but the Regexp previously used was actually more strict than just checking for `%` and the end and start of the string. In the new code, `%` will result in errors, as that just attempts to translate an empty string, which usually doesn't work, as it wasn't meant to be translated in the first place.

sdirix:
Ah I see, thanks for the explanation

PR #15276, opened 2025-03-25
Support for devcontainer workspaceMount & workspaceFolder

jonah-iden:
#### What it does

This adds support for `workspaceMount` and `workspaceFolder` options for devcontainer.json files

#### How to test

Create a `devcontainer.json` with content like this:

```
{
    "image": "mcr.microsoft.com/devcontainers/typescript-node:0-18",
    "forwardPorts": [3000],
    "workspaceFolder": "/workspace/testfolder/someTestfolderInWorkspace",
    "workspaceMount": "dst=/workspace/testfolder,type=bind",
    "containerEnv": {
        "TESTVAR": "testValue"
    }
}
```
optionally you can also add source and readonly options to the `workspaceMount` property.

this should then put the workspace at the specified location and open the folder defined by `workspaceFolder`
#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/dev-container/src/electron-node/devcontainer-contributions/workspace-creation-contribution.ts
+++ b/packages/dev-container/src/electron-node/devcontainer-contributions/workspace-creation-contribution.ts
@@ -1,0 +1,65 @@
+// *****************************************************************************
+// Copyright (C) 2024 Typefox and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { injectable } from '@theia/core/shared/inversify';
+import { ContainerCreationContribution } from '../docker-container-service';
+import { MaybePromise } from '@theia/core';
+import { ContainerCreateOptions } from 'dockerode';
+import { ContainerOutputProvider } from '../../electron-common/container-output-provider';
+import { DevContainerConfiguration } from '../devcontainer-file';
+import * as Docker from 'dockerode';
+
+@injectable()
+export class WorkspaceCreationContribution implements ContainerCreationContribution {
+    handleContainerCreation(createOptions: ContainerCreateOptions,
+        containerConfig: DevContainerConfiguration, api: Docker,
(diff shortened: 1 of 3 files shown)


packages/dev-container/src/electron-node/devcontainer-contributions/workspace-creation-contribution.ts
```diff
@@ -0,0 +1,65 @@
...
+
+        if (containerConfig.workspaceMount && containerConfig.workspaceFolder && createOptions.HostConfig?.Mounts) {
+            createOptions.HostConfig.Mounts[0] = {
+                ...createOptions.HostConfig.Mounts[0],
+                ...this.parseWorkspaceMount(containerConfig.workspaceMount as string)
+            };
+        }
+    }
+
+    protected parseWorkspaceMount(workspaceMount: string): Partial<Docker.MountSettings> {
```
sgraband: I believe this parsing logic duplicates what already exists in `MountsContribution.parseMountString`. Could we consolidate these into a shared utility function? The existing `MountsContribution` already handles mount string parsing and could be extended to support the additional options needed here.

See the existing implementation: https://github.com/eclipse-theia/theia/blob/d1bde5ecb778d50bd928050841068b4b8ededccb/packages/dev-container/src/electron-node/devcontainer-contributions/main-container-creation-contributions.ts#L142-L148

packages/dev-container/src/electron-node/devcontainer-contributions/workspace-creation-contribution.ts
```diff
@@ -0,0 +1,65 @@
...
+            } else if (key === 'readonly' || key === 'ro') {
+                mountSetting.ReadOnly = true;
+            } else if (key === 'bind-propagation') {
+                mountSetting.BindOptions = {
+                    Propagation: value as Docker.MountPropagation
+                };
+            }
+        }
+
+        return mountSetting;
```
sgraband: Could you add validation here? If `Source`, `Target`, or `Type` are missing from the parsed mount settings, the Docker API will fail. I think we should either throw a clear error message or provide sensible defaults.

packages/dev-container/src/electron-node/devcontainer-contributions/workspace-creation-contribution.ts
```diff
@@ -0,0 +1,65 @@
...
+import { DevContainerConfiguration } from '../devcontainer-file';
+import * as Docker from 'dockerode';
+
+@injectable()
+export class WorkspaceCreationContribution implements ContainerCreationContribution {
+    handleContainerCreation(createOptions: ContainerCreateOptions,
+        containerConfig: DevContainerConfiguration, api: Docker,
+        outputProvider?: ContainerOutputProvider | undefined): MaybePromise<void> {
+
+        if (containerConfig.workspaceMount && containerConfig.workspaceFolder && createOptions.HostConfig?.Mounts) {
```
sgraband: I believe the condition here is too strict. According to the devcontainer spec, `workspaceFolder` should work independently of `workspaceMount`. If only `workspaceFolder` is specified, it should still be used to set the workspace path (without modifying mounts). Could you split this into two separate checks?

PR #15287, opened 2025-03-25
Fix SiW height mismatch

### LINKED ISSUE #15271: Multiple scrollbars on search widget
### Bug Description:

The search widget displays two scrollbars: one appears on hover even when there is nothing to scroll, spanning the full height, and a second scrollbar appears when there is scrollable content but only stretches the content part. Possibly related to #14586.

### Steps to Reproduce:

1. Open the search view.
2. Observe that a scrollbar appears on hover even if there's no scrollable content (Try to scroll if it does not show up immediately).
3. Search for a letter (e.g., "a") to generate a large number of search hits.
4. Notice that a second scrollbar appears for the scrollable content.

Image

### Additional Information

- Theia Version: Theia IDE Preview 1.60.0 and Theia master

colin-grant-work:
#### What it does

Fixes #15271 by taking the SiW scrollbar out of the layout.
> Perhaps we should make this behavior the default, since in general, I don't think we want the progress bars affecting the layout of surrounding elements. But maybe there are better ways of handling the problem: we could just modify the height of the affected item to be 100% - 2px.

#### How to test

1. Open the Search in Workspace widget.
2. There shouldn't be an extra scrollbar when the widget isn't populated.
3. Perform a search.
4. If the search takes some time, you should see a progress bar.
5. When results are populated, there should be a scrollbar for the results (if necessary), but not for the rest of the widget.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/search-in-workspace/src/browser/styles/index.css
+++ b/packages/search-in-workspace/src/browser/styles/index.css
@@ -16,17 +16,12 @@
 
 #search-in-workspace {
   height: 100%;
-  position: relative;
+  display: flex;
+  flex-flow: column nowrap;
 }
 
 #search-in-workspace .theia-TreeContainer.empty {
   overflow: hidden;
-}
-
-#search-in-workspace .theia-progress-bar-container {
-    position: absolute;
-    top: 0;
-    left: 0;
 }
 
 .t-siw-search-container {
@@ -35,6 +30,7 @@
   flex-direction: column;
   height: 100%;
   box-sizing: border-box;
+  flex: 1 1 auto;
 }
 
 .t-siw-search-container .theia-ExpansionToggle {
--- a/packages/search-in-workspace/src/browser/styles/index.css
+++ b/packages/search-in-workspace/src/browser/styles/index.css
@@ -16,10 +16,17 @@
 
 #search-in-workspace {
   height: 100%;
+  position: relative;
 }
 
 #search-in-workspace .theia-TreeContainer.empty {
   overflow: hidden;
+}
+
+#search-in-workspace .theia-progress-bar-container {
+    position: absolute;
+    top: 0;
+    left: 0;
 }
 
 .t-siw-search-container {


tsmaeder:
> Perhaps we should make this behavior the default, since...

Just wondering where the quoted text in the description comes from. Also: why does this change fix the problem? Is it the right fix? What mechanism is causing the problem and why isn't it appearing elsewhere?

colin-grant-work:
@tsmaeder

> > Perhaps we should make this behavior the default, since...
> 
> Just wondering where the quoted text in the description comes from. Also: why does this change fix the problem? Is it the right fix? What mechanism is causing the problem and why isn't it appearing elsewhere?

I use that styling for commentary as well as quotation, so that's me commenting on the solution.

Re: what mechanism is causing the problem, I posted [a comment](https://github.com/eclipse-theia/theia/issues/15271#issuecomment-2752093841) on the issue explaining what's pushing the main widget out of the bounds of its container.

Re: why we don't see it elsewhere, [in at least one case](https://github.com/eclipse-theia/theia/blob/bfc328a790bcf92d99014142e7af6b54ab478123/packages/toolbar/src/browser/style/toolbar.css#L32-L35), we don't see it because we've used the same fix as I'm proposing here.

In other cases, styling like flexbox takes care of making sure that everything fits in its container; that would also work here - CSS is very flexible :-). Let me know if you have a preferred mechanism.

packages/search-in-workspace/src/browser/styles/index.css
```diff
@@ -16,12 +16,19 @@
...
 #search-in-workspace {
   height: 100%;
+  position: relative;
 }
 
 #search-in-workspace .theia-TreeContainer.empty {
   overflow: hidden;
 }
 
+#search-in-workspace .theia-progress-bar-container {
```
tsmaeder, line 26: I really don't understand why fixing the progress bar container to position 0,0 is the right fix here: If I understand correctly, the problem is that revealing the progress bar makes the `node` of the widget 2px longer and it no longer fits inside the container, thus causing the scroll bar. If we never want a scroll bar, shouldn't we layout the widget in a way that fits it's contents inside the bounds of the widget?

colin-grant-work: > shouldn't we layout the widget in a way that fits it's contents inside the bounds of the widget?

Yes, we should, and there are various ways of doing that.
 - Make the progress bar take up no space in the layout. That's what this PR currently does with position absolute. Whereas before the overflowing div had a height equal to its container, but started 2 px lower, thereby overflowing by 2px, now it has the same height and starts at the top, because the progress bar doesn't take up space.
 - Use styles like flex box to resize all children to sizes that fit in the container, even if we stick an extra 2px in there with the progress bar.
 - Use a calc value like `height: calc(100% - 2px)` to accommodate the progress bar.

No doubt any number of alternatives - CSS can do crazy things.

Do you have a preferred approach? The current approach does work - does ensure that, with the current styles, there is no overflow - but I'm happy to use a different technique if you think it would be better.

tsmaeder (approved):

PR #15288, opened 2025-03-26
Make parameters in ToolRequest mandatory

### LINKED ISSUE #14663: [Theia AI] Tool Calling with OpenAI API Bug/Improvement: Ensure each function has a clear schema
### Handling of functions with no parameters in Theia
Right now the function schema passed to OpenAI APIs and other AI interfaces for `getWorkspaceDirectoryStructure` does not contain a `parameter` field. According to standard practice, all function calls by the language models must include a argument object which could contain no argument fields if the arguments are empty. However, it's still necessary to provide the correct parameter object schema in the tool calling request in order to trigger the correct generation behavior in LLMs. E.g. if the `parameter` field doesn't exist, many models won't be able to generate correct arguments and lead to a chain of problems.

### Examples in Theia
Right now the schema for `getWorkspaceDirectoryStructure` is
```json
{
  "type": "function",
  "function": {
    "name": "getWorkspaceDirectoryStructure",
    "description": "Retrieve the complete directory structure of the workspace, listing only directories (no file contents). This structure excludes specific directories,\n            such as node_modules and hidden files, ensuring paths are within workspace boundaries."
  }
}
```

This should be amended to something like
```json
{
  "type": "function",
  "function": {
    "name": "getWorkspaceDirectoryStructure",
    "description": "Retrieve the complete directory structure of the workspace, listing only directories (no file contents). This structure excludes specific directories,\n            such as node_modules and hidden files, ensuring paths are within workspace boundaries."
  },
  "parameters": {
    "type": "object",
    "properties": {}
  }
}
```

### A few notes on tool calling templates in Theia

Another problem worth noting is the inclusion of predefined function calls in chat templates. Currently it's impossible to control the exact description of each function in the prompt, and you must include the variables for them to be added to the call. This is quite inflexible, in the case for example when the user wants to provide custom introductions/usage prompts to these functions.

eneufeld:
#### What it does

Make the parameters property of ToolRequest interface mandatory.
Most LLMs expect the parameters section of a ToolRequest to be defined.
OpenAI is there a bit more flexible but Anthropic isn't.
An Anthropic specific workaround could be removed now.

Fixes #14663

#### How to test
Test especially tools that do not have parameters, eg `getWorkspaceDirectoryStructure` in OpenAi and Anthropic.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-anthropic/src/node/anthropic-language-model.ts
+++ b/packages/ai-anthropic/src/node/anthropic-language-model.ts
@@ -28,11 +28,6 @@ import { Anthropic } from '@anthropic-ai/sdk';
 import { MessageParam } from '@anthropic-ai/sdk/resources';
 
 export const DEFAULT_MAX_TOKENS = 4096;
-const EMPTY_INPUT_SCHEMA = {
-    type: 'object',
-    properties: {},
-    required: []
-} as const;
 
 interface ToolCallback {
     readonly name: string;
@@ -254,7 +249,7 @@ export class AnthropicModel implements LanguageModel {
         return request.tools?.map(tool => ({
             name: tool.name,
             description: tool.description,
-            input_schema: tool.parameters ?? EMPTY_INPUT_SCHEMA
+            input_schema: tool.parameters
         } as Anthropic.Messages.Tool));
     }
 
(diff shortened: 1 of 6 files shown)


sdirix (approved):
Works for me

PR #15315, opened 2025-03-27
Explicitly set scrolling element to avoid document scroll

### LINKED ISSUE #15248: Setting: expand category/section with long sub-categories list shrinks the whole Theia page app length --> no main menu
### Bug Description:
Hi,
This is a strange issue, on the latest Theia build (master).

### Steps to Reproduce:

1. open the browser app
2. open the setting panel
3. expand the extensions category
4. the whole page length shrinks --> the main menu is not visible
5. click on a different view than the selected view (explorer) in the left panel, e.g. search, the page length gets back to its normal and the main menu is visible again
6. expand the AI Feature in the setting, it will shrink again...

### Additional Information

- Operating System: Mac - Chrome
- Theia Version: Latest master

colin-grant-work:
#### What it does

Fixes #15248

Calling `scrollIntoView()` on a node in the preference view caused the `document` to scroll part of the way. This PR explicitly instructs the proference widget's scroll container to execute the scroll, instead, preventing any leakage to the `document` element.

> NB: I do not know why this behavior changed, although it seems reasonable to guess that it is related to the Lumino migration. Previously, the `scrollIntoView` call affected only the preferences widget, and the document was not scrolled. It may be worth investigating if Lumino is explicitly making the `document` element scrollable in a way that older Phosphor did not, or if it handles scrolling within widgets differently so that we can find a solution that would protect adopters.

#### How to test

1. Open the browser app (or Electron app - it showed the same behavior)
2. Open the setting panel
3. Expand the extensions category
4. The extensions section should be scrolled into view, but the application apart from the scroll

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/browser/views/preference-editor-widget.ts
+++ b/packages/preferences/src/browser/views/preference-editor-widget.ts
@@ -242,7 +242,7 @@ export class PreferencesEditorWidget extends BaseWidget implements StatefulWidge
                 const { id, collection } = this.analyzeIDAndGetRendererGroup(nodeIDToScrollTo);
                 const renderer = collection.get(id);
                 if (renderer?.visible) {
-                    renderer.node.scrollIntoView();
+                    this.scrollContainer.scrollTo(0, renderer.node.offsetHeight);
                     return;
                 }
             }
--- a/packages/preferences/src/browser/views/preference-editor-widget.ts
+++ b/packages/preferences/src/browser/views/preference-editor-widget.ts
@@ -333,7 +333,7 @@ export class PreferencesEditorWidget extends BaseWidget implements StatefulWidge
     protected scrollWithoutModelUpdate(node?: HTMLElement): void {
         if (!node) { return; }
         this.shouldUpdateModelSelection = false;
-        node.scrollIntoView();
+        this.scrollContainer.scrollTo(0, node.offsetTop);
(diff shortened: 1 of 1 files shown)


packages/preferences/src/browser/views/preference-editor-widget.ts
```diff
@@ -333,7 +333,7 @@ export class PreferencesEditorWidget extends BaseWidget implements StatefulWidge
     protected scrollWithoutModelUpdate(node?: HTMLElement): void {
         if (!node) { return; }
         this.shouldUpdateModelSelection = false;
-        node.scrollIntoView();
+        this.scrollContainer.scrollTo(0, node.offsetTop);
```
tsmaeder, line 336: This fixes the problem when I expand the node when the preference editor is open. However, I can still reproduce the problem when I reload the page with the pref editor open and node expanded. I can fix that case if I change the other instance of `scrollIntoView()` in this file (line 245) to 
```
this.scrollContainer.scrollTo(0, renderer.node.offsetTop);
```

tsmaeder (changes requested):
See comment.

packages/preferences/src/browser/views/preference-editor-widget.ts
colin-grant-work: Good catch 👍

tsmaeder (approved):

PR #15334, opened 2025-03-30
add native gemini provider

eneufeld:
#### What it does

Alternative fix for #15247

This provides a native provider for GeminiAI.

#### How to test

Switch the model to use in your agent config.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-model.ts
+++ b/packages/ai-chat/src/common/chat-model.ts
@@ -1,5 +1,5 @@
 // *****************************************************************************
-// Copyright (C) 2024 EclipseSource GmbH.
+// Copyright (C) 2024-2025 EclipseSource GmbH.
 //
 // This program and the accompanying materials are made available under the
 // terms of the Eclipse Public License v. 2.0 which is available at
@@ -1032,7 +1032,8 @@ export class ToolCallChatResponseContentImpl implements ToolCallChatResponseCont
             actor: 'user',
             type: 'tool_result',
             tool_use_id: this.id ?? '',
-            content: this.result
+            content: this.result,
+            name: this.name ?? ''
         }];
     }
 }
--- a/packages/ai-chat/src/common/chat-model.ts
+++ b/packages/ai-chat/src/common/chat-model.ts
@@ -1,5 +1,5 @@
 // *****************************************************************************
-// Copyright (C) 2024-2025 EclipseSource GmbH.
+// Copyright (C) 2024 EclipseSource GmbH.
 //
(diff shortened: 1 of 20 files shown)


JonasHelming (commented):
Very very cool! Looking forward to test it as soon as I have an API key (on it). Maybe we can get this still into the current release!

Two high level comments before I can test:

1. Should we really call it "Gemini" (which I believe is one model familly) or rather "Google AI"?
2. Does this provider consider custom request settings (global, chat specific)?

JonasHelming:
Ia m running the license review to get rid of the license warning

eneufeld:
> Very very cool! Looking forward to test it as soon as I have an API key (on it). Maybe we can get this still into the current release!
> 
> Two high level comments before I can test:
> 
> 1. Should we really call it "Gemini" (which I believe is one model familly) or rather "Google AI"?
> 2. Does this provider consider custom request settings (global, chat specific)?

hmm, good point with the name, probably google would be better.

The provider should behave just as any other provider and allow to set settings globally or per chat.

Unfortunately many features don't yet work. Basically you can change temperature. 
Here is the full list: https://googleapis.github.io/js-genai/interfaces/types.GenerateContentConfig.html

packages/ai-google/src/common/google-language-models-manager.ts
```diff
@@ -0,0 +1,46 @@
+// *****************************************************************************
+// Copyright (C) 2025EclipseSource GmbH.
```
JonasHelming, line 2:
```suggestion
// Copyright (C) 2025 EclipseSource GmbH.
```

packages/ai-chat/src/common/chat-model.ts
```diff
@@ -1,5 +1,5 @@
 // *****************************************************************************
-// Copyright (C) 2024 EclipseSource GmbH.
+// Copyright (C) 2024-2025 EclipseSource GmbH.
```
JonasHelming, line 2:
```suggestion
// Copyright (C) 2024 EclipseSource GmbH.
```

JonasHelming (approved):
I tested this, it works very well. I did not investigate the code in detail. I will mark this a "experimental" in the documentation. But let's get this into the release, so we can test it in the field.

JonasHelming:
I have pushed on the two open IP tickets.

JonasHelming:
@eneufeld The license reviews are approved, so after a rebuild, the license check should go green and we can merge this

PR #15339, opened 2025-03-31
Fix ScanOSS button/icon in chat ui

### LINKED ISSUE #15294: [Theia AI] The ScanOSS button/icon does not align with the UI
Change Set: The ScanOSS button seems too large, especially since it is just an information that everything is ok. We should also consider moving the information down to the single files (e.g. as an overlay of the file icon)

And the `Apply All` button should be on the right side.

Image

Image

+ The ScanOSS icon above code blocks (generated) seems larger and doesn’t have the same foreground color as the other icons

haydar-metin:
#### What it does

Closes #15294 

## Code Editor
Default:

image

No match:

image

Match:

image

## Change Set

Default:

Screenshot from 2025-04-24 16-22-40

No Match (Readonly div):

image

Match (Clickable Button): 

image

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


sdirix:
The green and yellow icons are very dark. Should we (re)use (almost) the same color as for the checkmark and warning symbol?

haydar-metin:
IMO I think having a gray tone (default color, but lighter/darker) would be better. Otherwise, we would have green on green / yellow on yellow. What to you think?

planger:
Thank you @haydar-metin! This already looks much better.

Are the borders around the language and icons above the code editor just for debugging purposes or are they actually there? I think we should remove them for the actual application.

Also I'd put the ScanOSS button to the left-most position.

planger:
> Are the borders around the language and icons above the code editor just for debugging purposes or are they actually there? I think we should remove them for the actual application.

I double-checked. This only appears in the Light Modern and Dark Modern theme. I proposed a fix for that:
https://github.com/eclipse-theia/theia/pull/15351

haydar-metin:
@planger @sdirix I updated the colors and changed the order. See updated images.

packages/ai-scanoss/src/browser/style/index.css
```diff
@@ -1,169 +1,199 @@
...
+.icon-container.scanoss-icon::before {
+  width: 16px;
+  height: 16px;
+  mask-size: 16px;
+}
+
+.icon-container.scanoss-icon.clean,
+.icon-container.scanoss-icon.match,
+.icon-container.scanoss-icon.error {
+  margin-right: 6px;
```
planger, line 69:
```suggestion
  margin-right: 12px;
```

planger: image

planger (approved):
Thank you! Looks good to me. I'd just increase the margin-right of the scanoss icon to have a consistent gap between the icons (see inline).

image

PR #15344, opened 2025-03-31
AI Chat Context UI: Collapse label from right & no printed &nbsp;

colin-grant-work:
#### What it does

This PR addresses three minor issues in the styling of chat items:
 - Delegates to the label provider system for the 'additional information' label on context files. The current implementation more or less recapitulates the existing `WorkspaceUriLabelProviderContribution` logic, but by delegating directly to the `ChangeSetFileService`, it makes it more difficult to override labeling for specific URI types.
 - Wraps the components of the context element label so that the 'additional information' is clipped before the 'name':

Before:  
image

Note that both additional info and the name are clipped, though the name less so.
 
 After:  
  
image

Note that the extra path is clipped with `...`, but the name is not. This is more consistent with the behavior of e.g. tree elements.

- Removes the replacement of leading and trailing whitespace with a non-breaking space. The code indicated a desire to retain such whitespace, but there was misbehavior:
  - Trailing whitespace was preserved, but a literal `&nbsp` was also printed.
  - Line-initial whitespace can be meaningful in markdown, e.g. it can be a substitute for triple ticks to start a code block. By substituting it, we were negating that interpretation.

Before:

image

After:

image

I'm open to discussion on both points, if any reviewer would like to make a case for the old behavior.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
@@ -719,12 +719,14 @@ const ChatContext: React.FunctionComponent<ChatContextUI> = ({ context }) => (
             {context.map((element, index) => (
                 <li key={index} className="theia-ChatInput-ChatContext-Element" title={element.details} onClick={() => element.open?.()}>
                     <div className={`theia-ChatInput-ChatContext-Icon ${element.iconClass}`} />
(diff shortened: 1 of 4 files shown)


colin-grant-work:
> What do you think?

Sounds good to me.

planger (approved):

PR #15366, opened 2025-04-01
Avoid double registration of sticky scroll toggle

colin-grant-work:
#### What it does

At the moment, both the `editor` and the `monaco` packages register the command `editor.action.toggleStickyScroll` leading to a warning in the console upon the second registration.

This PR favors the `editor` package's implementation and suppresses the `monaco` package's registration. That seems reasonable in this case because the primary action in both the `editor` package and in [VSCode's implementation](https://github.com/microsoft/vscode/blob/0e333c267aa2ea042aa3128cbdaaf930dce66326/src/vs/editor/contrib/stickyScroll/browser/stickyScrollActions.ts#L44-L52) is to update the preferences, which can be done without direct access to a given editor.

Alternatively, we could register the command and the `View` menu item in the `editor` package and then defer to Monaco to register the handler. 

#### How to test

1. The command should be accessible from the `View` menu and from the command palette, and it should work in both locations.
2. There should be no warnings about double registration of the command.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-command.ts
+++ b/packages/monaco/src/browser/monaco-command.ts
@@ -46,6 +46,7 @@ export namespace MonacoCommands {
 
     export const EXCLUDE_ACTIONS = new Set([
         'editor.action.quickCommand',
+        'editor.action.toggleStickyScroll', // Handled by `editor` package.
         'undo',
         'redo'
     ]);


packages/monaco/src/browser/monaco-command.ts
```diff
@@ -46,6 +46,7 @@ export namespace MonacoCommands {
 
     export const EXCLUDE_ACTIONS = new Set([
         'editor.action.quickCommand',
+        'editor.action.toggleStickyScroll', // Handled by `editor` package.
```
tsmaeder, line 49: Would it be an option to remove our implementation of the command and go with the monaco implementation? It seems like this command is tightly bound to the monaco editor.

colin-grant-work: I agree that sticky scroll isn't really a 'generic' editor capability. On the other hand, we do register the sticky scroll preference in the `editor` package, and as I say in the description, the main effect of both implementations is a change in that preference, so from that perspective, it has a natural home there.

tsmaeder (approved):

PR #15373, opened 2025-04-02
Make icons right-aligned.

### LINKED ISSUE #15296: Toolbar actions are sometimes not right-aligned
### Bug Description:

Toolbar actions for "Problems view" and "Output" are not right-aligned. More views might be affected, so a general fix should be found.

Image

Image

### Additional Information

- Operating System: ubuntu 24.04
- Theia Version: latest master (https://github.com/eclipse-theia/theia/commit/240245010314e4605ab0fa3e893fe3b0155163f6)

tsmaeder:
#### What it does
fixes #15296

#### How to test
Make sure tab-bar toolbars are always right-aligned, for example in editors and views like the problems view.  Also make sure breadcrumbs still show up correctly. 
The main change is that we're' explicitly hiding (`display: none`) the breadcrumbs row when not active. 
I also restructured the code a bit, removing (IMO) unnecessary complications. 

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/tab-bars.ts
+++ b/packages/core/src/browser/shell/tab-bars.ts
@@ -697,9 +697,8 @@ export interface TabBarPrivateMethods {
  */
 export class ScrollableTabBar extends TabBar<Widget> {
 
-    protected scrollBar?: PerfectScrollbar;
+    protected scrollBar: PerfectScrollbar | undefined;
 
-    protected scrollBarFactory: () => PerfectScrollbar;
     protected pendingReveal?: Promise<void>;
     protected isMouseOver = false;
     protected needsRecompute = false;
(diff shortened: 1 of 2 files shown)


sdirix (approved):
I like that you got rid of so much "rewire" code! That's great! :rocket: 

I first thought I found a issue: When moving views next to each other, we need to focus them for the actions to show up:

image

image

Personally I think this is weird behavior, as for example the sidebar views still show their actions, even when not focused.

However I noticed that VS Code behaves the same, so maybe that's just a limitation there.

Therefore I approve this PR :+1:

packages/core/src/browser/shell/tab-bars.ts
```diff
@@ -712,11 +711,27 @@ export class ScrollableTabBar extends TabBar<Widget> {
     protected openTabsContainer: HTMLDivElement;
     protected openTabsRoot: Root;
 
-    constructor(options?: TabBar.IOptions<Widget> & PerfectScrollbar.Options, dynamicTabOptions?: ScrollableTabBar.Options) {
+    constructor(options?: TabBar.IOptions<Widget>, protected readonly scrollbarOptions?: PerfectScrollbar.Options, dynamicTabOptions?: ScrollableTabBar.Options) {
```
colin-grant-work, line 714: I agree with this change, but it is technically breaking. Is it worth the breakage?

tsmaeder: Yes, I think it is: the code has a better separation of concerns and since clients will have to change their code, it's preferable that the get a compile time error over keeping the code type compatible but breaking at runtime. How to fix the problem is very straightforward.

colin-grant-work: In that case, don't forget to add it to the changelog :-).

Alexander-Taran:
@sdirix 
rewireDOM
was marked as protected
so removing it was a breaking change.

we had it overridden in  our custom toolbar that does not work now  after bumping theia version )-:

sdirix:
> @sdirix rewireDOM was marked as protected so removing it was a breaking change.
> 
> we had it overridden in our custom toolbar that does not work now after bumping theia version )-:

All our methods are protected. We can not keep 100% backwards compatibility and never remove a protected method as this would mean that our codebase is practically frozen.

Having said that, we could have marked this PR as "breaking changes" so that is showed up more prominently in the changelog. However you would still face the same issue.

In any case, in your subclass you can restore the old behavior and/or call your overridden `rewireDOM` in the constructor, like we did before the changes.

Alexander-Taran:
> Having said that, we could have marked this PR as "breaking changes" so that is showed up more prominently in the changelog. However you would still face the same issue.

That is in fact what ppl  look at when they bump versions.
And we missed it during upgrade.

That was actually the point of my comment.
These kind of things should end up in breaking changes section.

As a side note: it would have been strange to leave an empty rewireDOM in the code and still call it from constructor.
We would have noticed the break faster though.

PR #15376, opened 2025-04-02
Do not re-export the configuration folder as env variable

sgraband:
#### What it does
Before this the configurationFolder (which includes the user settings for example) was exported as a env variable THEIA_CONFIG_DIR. This is problematic, as we also allow users to define a configurationFolder via this variable, and use it whenever provided for a new session.
Therefore, if you debug another application from a Theia tool (e.g. the Theia IDE) you will always get the tools configurationFolder, which is probably not intended. As this results in settings being changed in the started application to also be applied in the root too (e.g. colorTheme).

#### How to test
1. Start the Theia example application from a Theia application (Theia IDE, example application)
2. Run `echo $THEIA_CONFIG_DIR`
3. Notice that before this change you would see the root applications directory and with these changes you will get no output.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/node/env-variables/env-variables-server.ts
+++ b/packages/core/src/node/env-variables/env-variables-server.ts
@@ -55,19 +55,20 @@ export class EnvVariablesServerImpl implements EnvVariablesServer {
         const dataFolderPath = join(BackendApplicationPath, 'data');
         const userDataPath = join(dataFolderPath, 'user-data');
         const dataFolderExists = this.pathExistenceCache[dataFolderPath] ??= await pathExists(dataFolderPath);
+        let theiaConfigDir: string;
         if (dataFolderExists) {
             const userDataExists = this.pathExistenceCache[userDataPath] ??= await pathExists(userDataPath);
             if (userDataExists) {
-                process.env.THEIA_CONFIG_DIR = userDataPath;
+                theiaConfigDir = userDataPath;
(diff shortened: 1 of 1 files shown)


packages/core/src/node/env-variables/env-variables-server.ts
```diff
@@ -55,19 +55,20 @@ export class EnvVariablesServerImpl implements EnvVariablesServer {
         const dataFolderPath = join(BackendApplicationPath, 'data');
         const userDataPath = join(dataFolderPath, 'user-data');
         const dataFolderExists = this.pathExistenceCache[dataFolderPath] ??= await pathExists(dataFolderPath);
+        let theiaConfigDir: string;
         if (dataFolderExists) {
             const userDataExists = this.pathExistenceCache[userDataPath] ??= await pathExists(userDataPath);
             if (userDataExists) {
-                process.env.THEIA_CONFIG_DIR = userDataPath;
+                theiaConfigDir = userDataPath;
```
eneufeld, line 62: I would move this assignment one level up, so eg to line 60 or even directly assign the theiaConfigDir to it and just override for the else case in line 69

eneufeld (approved):
Looks good to me, I would take the chance to refactor the code a bit though

eneufeld (approved):

PR #15378, opened 2025-04-02
Add initial support for tracking token usage

eneufeld:
#### What it does

This adds logic to anthropic, openai and google providers to report token usage.
Furthermore it adds a new tab to the AiConfiguration view which shows the token usage per model.
Currently the usage is not persisted over application restart.

#### How to test

Open the AiConfigurationView and trigger some llm request.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/common/token-usage-service.ts
+++ b/packages/ai-core/src/common/token-usage-service.ts
@@ -1,0 +1,56 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { TokenUsageServiceClient } from './protocol';
+
+export const TokenUsageService = Symbol('TokenUsageService');
+
+export interface TokenUsage {
+    /** The input token count */
+    inputTokens: number;
+    /** The output token count */
+    outputTokens: number;
+    /** The model identifier */
+    model: string;
+    /** The timestamp of when the tokens were used */
+    timestamp: Date;
+    /** Request identifier */
+    requestId: string;
+}
+
+export interface TokenUsageParams {
+    /** The input token count */
+    inputTokens: number;
+    /** The output token count */
+    outputTokens: number;
+    /** Request identifier */
+    requestId: string;
+}
+
+export interface TokenUsageService {
+    /**
+     * Records token usage for a model interaction.
+     *
+     * @param model The identifier of the model that was used
+     * @param params Object containing token usage information
+     * @returns A promise that resolves when the token usage has been recorded
+     */
+    recordTokenUsage(model: string, params: TokenUsageParams): Promise<void>;
+
+    getTokenUsages(): Promise<TokenUsage[]>;
+
+    setClient(tokenUsageClient: TokenUsageServiceClient): void;
+}
--- a/packages/ai-core/src/node/token-usage-service-impl.ts
+++ b/packages/ai-core/src/node/token-usage-service-impl.ts
@@ -1,0 +1,65 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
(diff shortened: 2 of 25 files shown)


packages/ai-core/src/common/token-usage-service.ts
```diff
@@ -0,0 +1,46 @@
...
+export interface TokenUsageService {
+    /**
+     * Records token usage for a model interaction.
+     *
+     * @param inputTokens The number of input tokens used
+     * @param outputTokens The number of output tokens generated
+     * @param model The identifier of the model that was used
+     * @returns A promise that resolves when the token usage has been recorded
+     */
+    recordTokenUsage(inputTokens: number, outputTokens: number, model: string): Promise<void>;
```
planger, line 41: Nitpick: To me, it seems more logical to put the model id as a first parameter and the second parameter is an object with input and output tokens. This way we can more easily extend what we track besides input and output tokens (thinking in the direction of image and audio).

packages/ai-core/src/node/token-usage-service-impl.ts
```diff
@@ -0,0 +1,67 @@
...
+    private client: TokenUsageServiceClient | undefined;
+
+    /**
+     * Sets the client to notify about token usage changes
+     */
+    setClient(client: TokenUsageServiceClient | undefined): void {
+        this.client = client;
+    }
+
+    // Store token usage records
```
planger, line 32:
```suggestion
```

packages/ai-ide/src/browser/ai-configuration/token-usage-configuration-widget.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+// Using the interface from the token usage service
+
+@injectable()
+export class AITokenUsageConfigurationWidget extends ReactWidget {
+
+    static readonly ID = 'ai-token-usage-configuration-container-widget';
+    static readonly LABEL = nls.localize('theia/ai/tokenUsage/label', 'Token Usage');
+
+    // Data will be fetched from the service
```
planger, line 31:
```suggestion
```

packages/ai-ide/src/browser/ai-configuration/token-usage-configuration-widget.tsx
```diff
@@ -0,0 +1,153 @@
...
+    @inject(TokenUsageFrontendService)
+    protected readonly tokenUsageService: TokenUsageFrontendService;
+
+    @postConstruct()
+    protected init(): void {
+        this.id = AITokenUsageConfigurationWidget.ID;
+        this.title.label = AITokenUsageConfigurationWidget.LABEL;
+        this.title.closable = false;
+
+        // Load initial data
```
planger, line 46:
```suggestion
```

packages/ai-ide/src/browser/ai-configuration/token-usage-configuration-widget.tsx
```diff
@@ -0,0 +1,153 @@
...
+        this.refreshData();
+
+        // Subscribe to token usage updates
```
planger, line 49:
```suggestion
```

packages/ai-ide/src/browser/style/index.css
```diff
@@ -348,3 +348,104 @@
...
 }
+
+/* Token Usage Configuration Styles */
+.token-usage-configuration-container {
+  padding: 16px;
+}
+
+.token-usage-configuration-title {
+  margin: 0 0 16px 0;
+  border-bottom: 1px solid var(--theia-panelTitle-activeBorder);
```
planger, line 359: I think we shouldn't use the active border here. This color is somewhat reserved for components that have focus and are active. Here we use this color all the time.

packages/ai-ide/src/browser/style/index.css
```diff
@@ -348,3 +348,104 @@
...
+  background-color: var(--theia-button-hoverBackground);
+}
+
+.token-usage-refresh-button i {
+  margin-right: 6px;
+}
+
+.token-usage-table-container {
+  overflow-x: auto;
+  border: 1px solid var(--theia-panelTitle-activeBorder);
```
planger, line 392: I think we shouldn't use the active border here. This color is somewhat reserved for components that have focus and are active. Here we use this color all the time.

packages/ai-ide/src/browser/style/index.css
```diff
@@ -348,3 +348,104 @@
...
+.token-usage-table {
+  width: 100%;
+  border-collapse: collapse;
+}
+
+.token-usage-header th {
+  text-align: left;
+  padding: 10px;
+  background-color: var(--theia-editor-background);
+  border-bottom: 1px solid var(--theia-panelTitle-activeBorder);
```
planger, line 406: I think we shouldn't use the active border here. This color is somewhat reserved for components that have focus and are active. Here we use this color all the time.

planger (commented):
Very nice! Tested with OpenAI and Anthropic models.

### Tracking Request IDs

I think it might be worth include the request id in addition to the model id, when tracking the token usage. This would allow us to show the cost of specific requests later on.

### Token Usage View

The Token Usage view uses the active border color, even if it isn't active or has focus. I suggest to use a normal border color instead. As the table is not really getting focus, it should probably never use the focus colors.

Should we use `date-fns` to show the date relative to now (e.g. 1 week ago) and the raw date only on hover? This would feel more natural than showing the raw date. See for instance [the recent chat session](https://github.com/eclipse-theia/theia/blob/c7ac3edbabefc9379a9720c31e1511774c510f5f/packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts#L153).

planger (approved):
Very nice! Thanks for the revision. Looks good to me!

Issue #15386, opened 2025-04-03
Enable showing custom UI while certain LLM responses are streamed


planger:
### Problem
The current implementation of LLM [response content matchers](https://github.com/eclipse-theia/theia/blob/fcb684aa9bd71665002f65878baf973d5ca26f54/packages/ai-chat/src/common/response-content-matcher.ts#L46) only processes content when both start and end delimiters are present. This creates a suboptimal user experience when streaming responses from AI models:

- While content is streaming, users see raw, unprocessed text (like code blocks, technical instructions, etc.) rendered as markdown
- For structured content that will be processed by the agent into a specific UI eventually, this shows technical content not intended for the user
- Users see text "transform" into the custom UI when the complete content is received, creating a jarring visual experience

### Proposed Solution
Enhance the content matching system to optionally handle incomplete content matches during streaming:

- Add `incompleteContentFactory` to content matchers to create appropriate UI representations for partial content
- Allow matchers to display temporary content parts, such as progress indicators or placeholder components, while waiting for complete content
- Maintain a complete text buffer for streamed responses to properly handle incremental parsing
- Provide a renderer for displaying progress state to be used for certain use cases

PR #15387, opened 2025-04-03
feat(chat): enhance streaming content parsing for incomplete response content matches

planger:
#### What it does

- Enables `ResponseContentMatcher` to provide a `incompleteContentFactory` which will be used if the start delimiter matches, but no end delimiter hasn't been matched yet.
- Refactors `addStreamResponse` method to maintain a complete text buffer for streamed responses, enabling the handling of incomplete matches.
- Updated `parseContents` function to differentiate between complete and incomplete matches, allowing for better content management during streaming.
- Added unit tests for new functionality in `parse-contents-stream.spec.ts` and updated existing tests to utilize custom matchers for code content parsing.
- Introduces `ProgressChatResponseContent` interface and its implementation to support progress updates in chat responses via incomplete matches

There is some downside to using this approach:
As we eagerly match now, the `start` regex has to be pretty strict as we otherwise will show the content part where we shouldn't. So if the start delimiter is somewhat ambiguous it might be better to not use this approach for those cases. This is why this is optional and controlled by the specific response content matcher -- it can choose to provide an `incompleteContentFactory` or not.

Fixes https://github.com/eclipse-theia/theia/issues/15386

#### How to test

See how code content parts are now immediately matched and rendered:

https://github.com/user-attachments/assets/ffa022ce-7515-4ab6-abc9-ab37d53ecc2b

See how the `AskAndContinue` agent now shows a progress part instead of surfacing the structured `<question>` response to the user:

https://github.com/user-attachments/assets/cd9d8915-28c0-4a48-8d4b-e4d5ed9d619b

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-agents.ts
+++ b/packages/ai-chat/src/common/chat-agents.ts
@@ -60,7 +60,7 @@ import {
     ChatRequestModel,
     ThinkingChatResponseContentImpl
 } from './chat-model';
-import { findFirstMatch, parseContents } from './parse-contents';
+import { parseContents } from './parse-contents';
 import { DefaultResponseContentFactory, ResponseContentMatcher, ResponseContentMatcherProvider } from './response-content-matcher';
 import { ChatToolRequest, ChatToolRequestService } from './chat-tool-request-service';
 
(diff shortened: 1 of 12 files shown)


packages/ai-chat/src/common/chat-agents.ts
```diff
@@ -389,28 +389,38 @@ export abstract class AbstractStreamParsingChatAgent extends AbstractChatAgent {
...
+
+            // If it's a text token, append to our buffer
+            if (isTextResponsePart(token) && token.content) {
+                completeTextBuffer += token.content;
+
+                // Parse the entire buffer to get the updated content including incomplete matches
+                const parsedContents = this.parseContents(completeTextBuffer, request);
+
+                // Get all content before our marker
+                const contentBeforeMarker = request.response.response.content.slice(0, markerIndex);
```
eneufeld, line 412: I think the index needs to be recalculated in the else cases

eneufeld (changes requested):
The new code removes tool calls from the result.
Just ask coder to suggest improvements.

Yes the result is there but getFileContent and changeSet is not part of the chat in the end

packages/ai-chat/src/common/chat-agents.ts
planger: Great catch, I missed that! I reworked the algorithm. To keep it simple, we now only parse "uninterrupted" text. That means if there is a tool call in between, we won't retry parsing the text altogether again, but just keep what we've collected and parsed before.

Do you think this is a fair assumption? Thank you!

eneufeld:
Code looks better so far.
Can you check the parsing for code snippets?
This is how it now looks for me:
image
This was the prompt:
`@Universal create a ts example of an interface a class and a main code`

planger:
@eneufeld Do you mean the missing syntax highlighting?

I can't reproduce this, unfortunately. I did a clean build and this is how it looks like:

https://github.com/user-attachments/assets/1cef6373-c67e-472e-93b1-fe5a3eb1330e

Did you have not downloaded the plugins (VS Code extensions), by any chance?

eneufeld (approved):
I didn't download the plugins before. Works fine. Nice addition thank you!

PR #15389, opened 2025-04-03
Use SimpleMonacoEditor for most inline editors

### LINKED ISSUE #15218: Isolate embedded/standalone Monaco Editors
### Overview

We use Monaco Editors not only for file editing, but also inline like in the chat input, console input, and breakpoint condition editing.

At the moment these inline editors are not decoupled from the global Monaco services. As a consequence there are unintended side effects, for example the popular [vim extension](https://open-vsx.org/extension/vscodevim/vim) interferes with these embedded editors, making it practically unusable. 

### Suggestion

 - Offer a way to create isolated inline editors, which behave as intended by the caller, without extensions interfering with it
 - Make sure to still offer a way for certain extensions to extend the behavior on a whitelist principle, for example code completion should work within the breakpoint condition editor
 - Switch all inline editors to this new functionality, whitelisting what is necessary for each use case

Optional goal:
- Also offer the same mechanism for all non-inline editors

### References

See https://github.com/eclipse-theia/theia/pull/14957 for related changes.

colin-grant-work:
#### What it does

Closes #15218

Adds a new function `createSimpleInline` to the `MonacoEditorProvider` that returns an instance of `SimpleMonacoEditor` that is more customizable than our standard `MonacoEditor`, allowing customization of both services, which the `MonacoEditor` also does, and editor contributions, which the `MonacoEditor` does not, and 'simplicity', which the `MonacoEditor` also doesn't. Control of contributions is the more direct way to control the presence or absence, rather than behavior, of certain features: remove the inlay hints contribution, and the editor won't have inlay hints, etc.

See discussion on issue re: current design.

#### How to test

1. Three current inline editor use cases have been replaced: breakpoint widgets, AI chat input, and AI chat code snippet rendering.
2. Confirm that each of the three behaves as expected:
  - breakpoint widgets should have code completion for the current context. (But NB that the code completion [isn't currently](https://github.com/eclipse-theia/theia/issues/15218#issuecomment-2775733958) as sophisticated as in the editor itself.
  - AI chat input should have code completion for chat variables, etc.
  - AI chat code snippets should still be colored according to the language in which they're written.

Note that currently, all three editors are instantiated with a full complement of both services and editor contributions - the primary difference in behavior is from setting `isSimpleWidget` to true; see discussion [here](https://github.com/eclipse-theia/theia/issues/15218#issuecomment-2773845226). 

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/api-samples/src/browser/chat/ask-and-continue-chat-agent-contribution.ts
+++ b/examples/api-samples/src/browser/chat/ask-and-continue-chat-agent-contribution.ts
@@ -36,7 +36,7 @@ export function bindAskAndContinueChatAgentContribution(bind: interfaces.Bind): 
 const systemPrompt: PromptTemplate = {
     id: 'askAndContinue-system',
     template: `
-You are an agent demonstrating on how to generate questions and continuing the conversation based on the user's answers.
+You are an agent demonstrating how to generate questions and continue the conversation based on the user's answers.
 
 First answer the user's question or continue their story.
 Then come up with an interesting question and 2-3 answers which will be presented to the user as multiple choice.
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "license:check": "node scripts/check_3pp_licenses.js",
     "license:check:review": "node scripts/check_3pp_licenses.js --review",
     "lint": "lerna run lint",
+    "lint:fix": "lerna run lint -- --fix",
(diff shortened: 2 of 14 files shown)


sdirix (approved):
Looks good to me! I did not notice any regression.

I tested with the `vim` extension, and it worked although I had breakpoint editor and chat view open! Nice!

PR #15427, opened 2025-04-08
Agent prompt suggestions & chat summary

colin-grant-work:
#### What it does

This PR adds a number of capabilities to the AI chat system discussed in #15094:
- The ability for particular agents to contribute suggestions to the chat, which are rendered above the input.
> At the moment, these are a static part of the agent's declaration. They could be implemented as a push system (agent fires event when suggestions changed) or pull system (agent can implement a method that is passed, e.g. a `ChatSession` and returns suggestions).
- A new context variable type: session summaries, along with facilities to generate summaries, resolve summary variables, and resolve a set of summaries as part of the system prompt of a given agent.
- The ability to open editors to view the values of context variables.
> I'm not 100% sold on this feature. For one thing, it's a bit redundant for files. For another, at the moment, the content would be basically static and reflect the value of the variable at the time that it was resolved (the time at which the editor was opened). It might be possible to use the resolution context to provide updates when the variable service resolves a variable - e.g. if the same session is involved - but the way things are declared at the moment, it's hard to use the resolution context in a generic way.

Closes #15094 

To do:

 - [x] 2.1: I would introduce this in the default renderer, i.e. actions in the text plus optionally additional actions that are shown under "more" in a popup. Follow-up created: https://github.com/eclipse-theia/theia/issues/15068
 - [x] show progress when creating the summary (or create it earlier).
 - [x] I think I'm watching for file changes but not doing anything when they happen - implement updates if files are deleted / modified on disk.
 - [ ] Maybe hook up response rendering system instead of current (almost) direct markdown?

#### How to test

1. Have a conversation with an agent.
2. Start a new chat.
> If your first chat was with Coder, it should give you two suggestions about starting a new chat, or starting a new chat with a summary of the current chat. If you choose the latter, it should automatically add a context variable.
3. You can add a summary of the first chat to the current chat in a couple of ways:
  - it can be added to the context using the `+` in the bottom left of the input. Choose 'Session Summary' and then the session you're interested in.
  - it can be added to a message using `#session-summary:<id>` - you should get auto-completion.
4. If you add a session summary, it should be included in the message to the LLM - check the history.
> There's bit of a UX hiccup here. It takes several seconds to get the AI-generated summary for the first time (or to generate a new one if the session to be summarized adds messages). Only once that interaction is complete do we initiate the request that required the summary. Perhaps we should show the user message in the UI before we've parsed all of its variables, and then show the progress message until we start getting content back for the main request?

1. Separately, if you add any kind of variable to the context, you should be able to click on the element in the context display and see the resolved value of that variable, e.g. the summary it provides of a given chat session.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
+++ b/packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
@@ -292,9 +292,10 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
     protected async summarizeActiveSession(): Promise<string | undefined> {
         const activeSession = this.chatService.getActiveSession();
         if (!activeSession) { return; }
-        await this.taskContextService.summarize(activeSession).catch(err => {
+        return this.taskContextService.summarize(activeSession).catch(err => {
             console.warn('Error while summarizing session:', err);
             this.messageService.error('Unable to summarize current session. Please confirm that the summary agent is not disabled.');
+            return undefined;
         });
     }
 }
--- a/packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
+++ b/packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
@@ -19,10 +19,7 @@ import { CommandRegistry, Emitter, isOSX, nls, QuickInputButton, QuickInputServi
 import { Widget } from '@theia/core/lib/browser';
 import {
     AI_CHAT_NEW_CHAT_WINDOW_COMMAND,
-    AI_CHAT_NEW_WITH_TASK_CONTEXT,
-    AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION,
     AI_CHAT_SHOW_CHATS_COMMAND,
-    AI_CHAT_SUMMARIZE_CURRENT_SESSION,
     ChatCommands
 } from './chat-view-commands';
 import { ChatAgentLocation, ChatService } from '@theia/ai-chat';
@@ -99,7 +96,7 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
             execute: () => this.openView().then(() => this.chatService.createSession(ChatAgentLocation.Panel, { focus: true })),
             isVisible: widget => this.withWidget(widget, () => true),
         });
-        registry.registerCommand(AI_CHAT_NEW_WITH_TASK_CONTEXT, {
+        registry.registerCommand(ChatCommands.AI_CHAT_NEW_WITH_TASK_CONTEXT, {
             execute: async () => {
                 const activeSession = this.chatService.getActiveSession();
                 const id = await this.summarizeActiveSession();
@@ -110,7 +107,7 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
             },
             isVisible: () => false
         });
-        registry.registerCommand(AI_CHAT_SUMMARIZE_CURRENT_SESSION, {
+        registry.registerCommand(ChatCommands.AI_CHAT_SUMMARIZE_CURRENT_SESSION, {
             execute: async () => this.summarizeActiveSession(),
             isVisible: widget => {
                 if (widget && !this.withWidget(widget)) { return false; }
@@ -126,7 +123,7 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
                     && !this.taskContextService.hasSummary(activeSession);
             }
         });
-        registry.registerCommand(AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION, {
+        registry.registerCommand(ChatCommands.AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION, {
             execute: async () => {
                 const id = await this.summarizeActiveSession();
                 if (!id) { return; }
@@ -192,13 +189,13 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
         this.taskContextService.onDidChange(() => sessionSummarizibilityChangedEmitter.fire());
         this.chatService.onSessionEvent(event => event.type === 'activeChange' && sessionSummarizibilityChangedEmitter.fire());
         registry.registerItem({
-            id: 'chat-view.' + AI_CHAT_SUMMARIZE_CURRENT_SESSION.id,
-            command: AI_CHAT_SUMMARIZE_CURRENT_SESSION.id,
+            id: 'chat-view.' + ChatCommands.AI_CHAT_SUMMARIZE_CURRENT_SESSION.id,
+            command: ChatCommands.AI_CHAT_SUMMARIZE_CURRENT_SESSION.id,
             onDidChange: sessionSummarizibilityChangedEmitter.event
         });
         registry.registerItem({
-            id: 'chat-view.' + AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION.id,
-            command: AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION.id,
+            id: 'chat-view.' + ChatCommands.AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION.id,
+            command: ChatCommands.AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION.id,
             onDidChange: sessionSummarizibilityChangedEmitter.event
         });
     }
--- a/packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
+++ b/packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
@@ -15,7 +15,7 @@
 // *****************************************************************************
 
 import { inject, injectable } from '@theia/core/shared/inversify';
-import { CommandRegistry, Emitter, isOSX, nls, QuickInputButton, QuickInputService, QuickPickItem } from '@theia/core';
+import { CommandRegistry, Emitter, isOSX, MessageService, nls, QuickInputButton, QuickInputService, QuickPickItem } from '@theia/core';
 import { Widget } from '@theia/core/lib/browser';
 import {
     AI_CHAT_NEW_CHAT_WINDOW_COMMAND,
@@ -48,6 +48,8 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
     protected readonly quickInputService: QuickInputService;
     @inject(TaskContextService)
     protected readonly taskContextService: TaskContextService;
+    @inject(MessageService)
+    protected readonly messageService: MessageService;
 
     protected static readonly RENAME_CHAT_BUTTON: QuickInputButton = {
         iconClass: 'codicon-edit',
@@ -290,7 +292,10 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
     protected async summarizeActiveSession(): Promise<string | undefined> {
         const activeSession = this.chatService.getActiveSession();
         if (!activeSession) { return; }
-        return await this.taskContextService.summarize(activeSession);
+        await this.taskContextService.summarize(activeSession).catch(err => {
+            console.warn('Error while summarizing session:', err);
+            this.messageService.error('Unable to summarize current session. Please confirm that the summary agent is not disabled.');
+        });
     }
 }
 
--- a/packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
+++ b/packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
@@ -17,7 +17,14 @@
 import { inject, injectable } from '@theia/core/shared/inversify';
 import { CommandRegistry, Emitter, isOSX, nls, QuickInputButton, QuickInputService, QuickPickItem } from '@theia/core';
 import { Widget } from '@theia/core/lib/browser';
-import { AI_CHAT_NEW_CHAT_WINDOW_COMMAND, AI_CHAT_NEW_WITH_TASK_CONTEXT, AI_CHAT_SHOW_CHATS_COMMAND, AI_CHAT_SUMMARIZE_CURRENT_SESSION, ChatCommands } from './chat-view-commands';
+import {
+    AI_CHAT_NEW_CHAT_WINDOW_COMMAND,
+    AI_CHAT_NEW_WITH_TASK_CONTEXT,
+    AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION,
+    AI_CHAT_SHOW_CHATS_COMMAND,
+    AI_CHAT_SUMMARIZE_CURRENT_SESSION,
+    ChatCommands
+} from './chat-view-commands';
 import { ChatAgentLocation, ChatService } from '@theia/ai-chat';
 import { AbstractViewContribution } from '@theia/core/lib/browser/shell/view-contribution';
 import { TabBarToolbarContribution, TabBarToolbarRegistry } from '@theia/core/lib/browser/shell/tab-bar-toolbar';
@@ -44,7 +51,6 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
     protected readonly quickInputService: QuickInputService;
     @inject(TaskContextService)
     protected readonly taskContextService: TaskContextService;
-
 
     protected static readonly RENAME_CHAT_BUTTON: QuickInputButton = {
         iconClass: 'codicon-edit',
@@ -109,9 +115,27 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
             isVisible: widget => {
                 if (widget && !this.withWidget(widget)) { return false; }
                 const activeSession = this.chatService.getActiveSession();
-                return !!activeSession?.model.getRequests().length
-                    && activeSession?.model.location === ChatAgentLocation.Panel
+                return activeSession?.model.location === ChatAgentLocation.Panel
                     && !this.taskContextService.hasSummary(activeSession);
+            },
+            isEnabled: widget => {
+                if (widget && !this.withWidget(widget)) { return false; }
+                const activeSession = this.chatService.getActiveSession();
+                return activeSession?.model.location === ChatAgentLocation.Panel
+                    && !activeSession.model.isEmpty()
+                    && !this.taskContextService.hasSummary(activeSession);
+            }
+        });
+        registry.registerCommand(AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION, {
+            execute: async () => {
+                const id = await this.summarizeActiveSession();
+                if (!id) { return; }
+                await this.taskContextService.open(id);
+            },
+            isVisible: widget => {
+                if (widget && !this.withWidget(widget)) { return false; }
+                const activeSession = this.chatService.getActiveSession();
+                return !!activeSession && this.taskContextService.hasSummary(activeSession);
             }
         });
         registry.registerCommand(AI_CHAT_SHOW_CHATS_COMMAND, {
@@ -170,6 +194,11 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
         registry.registerItem({
             id: 'chat-view.' + AI_CHAT_SUMMARIZE_CURRENT_SESSION.id,
             command: AI_CHAT_SUMMARIZE_CURRENT_SESSION.id,
+            onDidChange: sessionSummarizibilityChangedEmitter.event
+        });
+        registry.registerItem({
+            id: 'chat-view.' + AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION.id,
+            command: AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION.id,
             onDidChange: sessionSummarizibilityChangedEmitter.event
         });
     }
--- a/packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
+++ b/packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
@@ -15,9 +15,9 @@
 // *****************************************************************************
 
 import { inject, injectable } from '@theia/core/shared/inversify';
-import { CommandRegistry, isOSX, nls, QuickInputButton, QuickInputService, QuickPickItem } from '@theia/core';
+import { CommandRegistry, Emitter, isOSX, nls, QuickInputButton, QuickInputService, QuickPickItem } from '@theia/core';
 import { Widget } from '@theia/core/lib/browser';
-import { AI_CHAT_NEW_CHAT_WINDOW_COMMAND, AI_CHAT_SHOW_CHATS_COMMAND, ChatCommands } from './chat-view-commands';
+import { AI_CHAT_NEW_CHAT_WINDOW_COMMAND, AI_CHAT_NEW_WITH_TASK_CONTEXT, AI_CHAT_SHOW_CHATS_COMMAND, AI_CHAT_SUMMARIZE_CURRENT_SESSION, ChatCommands } from './chat-view-commands';
 import { ChatAgentLocation, ChatService } from '@theia/ai-chat';
 import { AbstractViewContribution } from '@theia/core/lib/browser/shell/view-contribution';
 import { TabBarToolbarContribution, TabBarToolbarRegistry } from '@theia/core/lib/browser/shell/tab-bar-toolbar';
@@ -30,6 +30,8 @@ import { AI_SHOW_SETTINGS_COMMAND } from '@theia/ai-core/lib/browser';
 import { OPEN_AI_HISTORY_VIEW } from '@theia/ai-history/lib/browser/ai-history-contribution';
 import { ChatNodeToolbarCommands } from './chat-node-toolbar-action-contribution';
 import { isEditableRequestNode, type EditableRequestNode } from './chat-tree-view';
+import { TASK_CONTEXT_VARIABLE } from '@theia/ai-chat/lib/browser/task-context-variable';
+import { TaskContextService } from '@theia/ai-chat/lib/browser/task-context-service';
 
 export const AI_CHAT_TOGGLE_COMMAND_ID = 'aiChat:toggle';
 
@@ -40,6 +42,9 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
     protected readonly chatService: ChatService;
     @inject(QuickInputService)
     protected readonly quickInputService: QuickInputService;
+    @inject(TaskContextService)
+    protected readonly taskContextService: TaskContextService;
+
 
     protected static readonly RENAME_CHAT_BUTTON: QuickInputButton = {
         iconClass: 'codicon-edit',
@@ -85,14 +90,34 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
             })
         });
         registry.registerCommand(AI_CHAT_NEW_CHAT_WINDOW_COMMAND, {
-            execute: () => this.chatService.createSession(ChatAgentLocation.Panel, { focus: true }),
-            isEnabled: widget => this.withWidget(widget, () => true),
+            execute: () => this.openView().then(() => this.chatService.createSession(ChatAgentLocation.Panel, { focus: true })),
             isVisible: widget => this.withWidget(widget, () => true),
+        });
+        registry.registerCommand(AI_CHAT_NEW_WITH_TASK_CONTEXT, {
+            execute: async () => {
+                const activeSession = this.chatService.getActiveSession();
+                const id = await this.summarizeActiveSession();
+                if (!id || !activeSession) { return; }
+                const newSession = this.chatService.createSession(ChatAgentLocation.Panel, { focus: true }, activeSession.pinnedAgent);
+                const summaryVariable = { variable: TASK_CONTEXT_VARIABLE, arg: id };
+                newSession.model.context.addVariables(summaryVariable);
+            },
+            isVisible: () => false
+        });
+        registry.registerCommand(AI_CHAT_SUMMARIZE_CURRENT_SESSION, {
+            execute: async () => this.summarizeActiveSession(),
+            isVisible: widget => {
+                if (widget && !this.withWidget(widget)) { return false; }
+                const activeSession = this.chatService.getActiveSession();
+                return !!activeSession?.model.getRequests().length
+                    && activeSession?.model.location === ChatAgentLocation.Panel
+                    && !this.taskContextService.hasSummary(activeSession);
+            }
         });
         registry.registerCommand(AI_CHAT_SHOW_CHATS_COMMAND, {
             execute: () => this.selectChat(),
-            isEnabled: widget => this.withWidget(widget, () => true) && this.chatService.getSessions().length > 1,
-            isVisible: widget => this.withWidget(widget, () => true)
+            isEnabled: widget => this.withWidget(widget) && this.chatService.getSessions().length > 1,
+            isVisible: widget => this.withWidget(widget)
         });
         registry.registerCommand(ChatNodeToolbarCommands.EDIT, {
             isEnabled: node => isEditableRequestNode(node) && !node.request.isEditing,
@@ -138,6 +163,14 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
             group: 'ai-settings',
             priority: 1,
             isVisible: widget => this.withWidget(widget),
+        });
+        const sessionSummarizibilityChangedEmitter = new Emitter<void>();
+        this.taskContextService.onDidChange(() => sessionSummarizibilityChangedEmitter.fire());
+        this.chatService.onSessionEvent(event => event.type === 'activeChange' && sessionSummarizibilityChangedEmitter.fire());
+        registry.registerItem({
+            id: 'chat-view.' + AI_CHAT_SUMMARIZE_CURRENT_SESSION.id,
+            command: AI_CHAT_SUMMARIZE_CURRENT_SESSION.id,
+            onDidChange: sessionSummarizibilityChangedEmitter.event
         });
     }
 
@@ -226,6 +259,12 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
 
     canExtractChatView(chatView: ChatViewWidget): boolean {
         return !chatView.secondaryWindow;
+    }
+
+    protected async summarizeActiveSession(): Promise<string | undefined> {
+        const activeSession = this.chatService.getActiveSession();
+        if (!activeSession) { return; }
+        return await this.taskContextService.summarize(activeSession);
     }
 }
 
--- a/packages/ai-chat-ui/src/browser/chat-view-commands.ts
+++ b/packages/ai-chat-ui/src/browser/chat-view-commands.ts
@@ -38,27 +38,29 @@ export namespace ChatCommands {
         category: CHAT_CATEGORY,
         iconClass: codicon('bracket')
     }, 'Set Session Settings', CHAT_CATEGORY_KEY);
+
+    export const AI_CHAT_NEW_WITH_TASK_CONTEXT: Command = {
+        id: 'ai-chat.new-with-task-context',
+    };
+
+    export const AI_CHAT_SUMMARIZE_CURRENT_SESSION = Command.toLocalizedCommand({
+        id: 'ai-chat-summary-current-session',
+        iconClass: codicon('go-to-editing-session'),
+        label: 'Summarize Current Session',
+        category: CHAT_CATEGORY
+    }, undefined, CHAT_CATEGORY_KEY);
+
+    export const AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION = Command.toLocalizedCommand({
+        id: 'ai-chat-open-current-session-summary',
+        iconClass: codicon('note'),
+        label: 'Open Current Session Summary',
+        category: CHAT_CATEGORY
+    }, undefined, CHAT_CATEGORY_KEY);
 }
 
 export const AI_CHAT_NEW_CHAT_WINDOW_COMMAND: Command = {
     id: 'ai-chat-ui.new-chat',
     iconClass: codicon('add')
-};
-
-export const AI_CHAT_NEW_WITH_TASK_CONTEXT: Command = {
-    id: 'ai-chat.new-with-task-context',
-};
-
-export const AI_CHAT_SUMMARIZE_CURRENT_SESSION: Command = {
-    id: 'ai-chat-summary-current-session',
-    iconClass: codicon('go-to-editing-session'),
-    label: 'Summarize Current Session'
-};
-
-export const AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION: Command = {
-    id: 'ai-chat-open-current-session-summary',
-    iconClass: codicon('note'),
-    label: 'Open Current Session Summary'
 };
 
 export const AI_CHAT_SHOW_CHATS_COMMAND: Command = {
--- a/packages/ai-chat-ui/src/browser/chat-view-commands.ts
+++ b/packages/ai-chat-ui/src/browser/chat-view-commands.ts
@@ -45,6 +45,15 @@ export const AI_CHAT_NEW_CHAT_WINDOW_COMMAND: Command = {
     iconClass: codicon('add')
 };
 
+export const AI_CHAT_NEW_WITH_TASK_CONTEXT: Command = {
+    id: 'ai-chat.new-with-task-context',
+};
+
+export const AI_CHAT_SUMMARIZE_CURRENT_SESSION: Command = {
+    id: 'ai-chat-summary-current-session',
+    iconClass: codicon('note')
+};
+
 export const AI_CHAT_SHOW_CHATS_COMMAND: Command = {
     id: 'ai-chat-ui.show-chats',
     iconClass: codicon('history')
--- a/packages/ai-chat-ui/src/browser/chat-view-commands.ts
+++ b/packages/ai-chat-ui/src/browser/chat-view-commands.ts
@@ -51,7 +51,14 @@ export const AI_CHAT_NEW_WITH_TASK_CONTEXT: Command = {
 
 export const AI_CHAT_SUMMARIZE_CURRENT_SESSION: Command = {
     id: 'ai-chat-summary-current-session',
-    iconClass: codicon('note')
+    iconClass: codicon('go-to-editing-session'),
+    label: 'Summarize Current Session'
+};
+
+export const AI_CHAT_OPEN_SUMMARY_FOR_CURRENT_SESSION: Command = {
+    id: 'ai-chat-open-current-session-summary',
+    iconClass: codicon('note'),
+    label: 'Open Current Session Summary'
 };
 
 export const AI_CHAT_SHOW_CHATS_COMMAND: Command = {
--- a/packages/ai-chat/src/browser/task-context-service.ts
+++ b/packages/ai-chat/src/browser/task-context-service.ts
@@ -19,7 +19,7 @@ import { MaybePromise, ProgressService, URI, generateUuid, Event } from '@theia/
 import { ChatAgent, ChatAgentLocation, ChatService, ChatSession, MutableChatModel, MutableChatRequestModel, ParsedChatRequestTextPart } from '../common';
 import { ChatSessionSummaryAgent } from '../common/chat-session-summary-agent';
 import { Deferred } from '@theia/core/lib/common/promise-util';
-import { PromptService } from '@theia/ai-core';
+import { AgentService, PromptService } from '@theia/ai-core';
 import { CHAT_SESSION_SUMMARY_PROMPT } from '../common/chat-session-summary-agent-prompt';
 
 export interface SummaryMetadata {
@@ -49,7 +49,7 @@ export class TaskContextService {
     protected pendingSummaries = new Map<string, Promise<Summary>>();
 
     @inject(ChatService) protected readonly chatService: ChatService;
-    @inject(ChatSessionSummaryAgent) protected readonly summaryAgent: ChatSessionSummaryAgent;
+    @inject(AgentService) protected readonly agentService: AgentService;
     @inject(PromptService) protected readonly promptService: PromptService;
     @inject(TaskContextStorageService) protected readonly storageService: TaskContextStorageService;
     @inject(ProgressService) protected readonly progressService: ProgressService;
@@ -104,7 +104,13 @@ export class TaskContextService {
         }
     }
 
-    protected async getLlmSummary(session: ChatSession, promptId: string = CHAT_SESSION_SUMMARY_PROMPT.id, agent: ChatAgent = this.summaryAgent): Promise<string> {
+    protected async getLlmSummary(session: ChatSession, promptId: string = CHAT_SESSION_SUMMARY_PROMPT.id, agent?: ChatAgent): Promise<string> {
+        agent = agent || this.agentService.getAgents().find<ChatAgent>((candidate): candidate is ChatAgent =>
+            'invoke' in candidate
+            && typeof candidate.invoke === 'function'
+            && candidate.id === ChatSessionSummaryAgent.ID
+        );
+        if (!agent) { throw new Error('Unable to identify agent for summary.'); }
         const model = new MutableChatModel(ChatAgentLocation.Panel);
         const prompt = await this.promptService.getPrompt(promptId || CHAT_SESSION_SUMMARY_PROMPT.id, undefined, { model: session.model });
         if (!prompt) { return ''; }
--- a/packages/ai-chat/src/browser/task-context-service.ts
+++ b/packages/ai-chat/src/browser/task-context-service.ts
@@ -1,0 +1,138 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { inject, injectable } from '@theia/core/shared/inversify';
+import { MaybePromise, ProgressService, URI, generateUuid, Event } from '@theia/core';
+import { ChatAgent, ChatAgentLocation, ChatService, ChatSession, MutableChatModel, MutableChatRequestModel, ParsedChatRequestTextPart } from '../common';
+import { ChatSessionSummaryAgent } from '../common/chat-session-summary-agent';
+import { Deferred } from '@theia/core/lib/common/promise-util';
+import { PromptService } from '@theia/ai-core';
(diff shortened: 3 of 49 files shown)


eneufeld:
I tested this. The ui looks nice and I think this is a great improvement.

This was my test:
- initial prompt: `@Coder suggets improvement for #currentRelativeFilePath `
- After the answer (which contained reading the file and a changeset) I then started a new chat with a summary (the load took a bit without a progress what was concerning) 
- Looking at the context, it did not contain any of the file references
- Then I prompted: `how do we continue?` and it searched the workspace to find the file from the first prompt but only based on the class name it had in the summary.

Based on this I would suggest:
- resolve variables in the summary
- show progress when creating the summary

JonasHelming:
Quick tested this, looks cool, but i want to dig into it more. We should collect all feedback before iterating on it, but thanks Colin for preparing this!

packages/ai-ide/src/browser/coder-agent.ts
```diff
@@ -30,6 +32,8 @@ export class CoderAgent extends AbstractStreamParsingChatAgent {
         identifier: 'openai/gpt-4o',
     }];
     protected defaultLanguageModelPurpose: string = 'chat';
+    readonly suggestions = [new MarkdownStringImpl(`Keep chats short and focused. [Start a new chat](command:${AI_CHAT_NEW_CHAT_WINDOW_COMMAND.id}) for a new task`
```
JonasHelming, line 35: Currently these are static. 
Question 1: To be able to also use this UI mechanism, we believe this mechanism should be dynamic. Do we want to allow the agent to push suggestions or do we pull from the chat, e.g. when a message is completed (Jonas thinks: push via the chat session)

Question 2: Currently, we use markdown String, which restricts us to commands. This might be weird, if we want to do something agent specific. Do we want to also allow real call backs, i.e. handing in functions here?

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
```diff
@@ -0,0 +1,115 @@
...
+            range,
+            insertText: `${prefix}${id}`,
+            detail: id,
+            filterText: userInput,
+        }));
+    }
+
+    protected getItems(): QuickPickItemOrSeparator[] {
+        return [
+            ...(this.summaries.size ? [{ type: 'separator', label: 'Recent Summaries' }] satisfies QuickPickSeparator[] : []),
```
JonasHelming, line 75: Question 3:
Instead of holding the items here, we discussed to introduce a "Task Context Service" that is responsible for creating and managing task summaries. By default it holds task contexts in memory, but you can set a directory. If a directory is set, the task contexts are persisted as .md files there.
In the IDE package, we introduce a user setting with the default .prompts/task-contexts which sets the directory, so in the IDE tasks are persisted by default. Fine?

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
```diff
@@ -0,0 +1,115 @@
...
+        return request.variable.id === SESSION_SUMMARY_VARIABLE.id ? 10000 : -5;
+    }
+
+    async resolve(request: AIVariableResolutionRequest, context: AIVariableContext): Promise<ResolvedAIContextVariable | undefined> {
+        if (request.variable.id !== SESSION_SUMMARY_VARIABLE.id || !request.arg) { return; }
+        const existingSession = this.chatService.getSession(request.arg);
+        const existingSummary = this.summaries.get(request.arg);
+        const newSummaryPossibleAndNecessary = !!existingSession && (!existingSummary || existingSummary.length !== existingSession.model.getRequests().length);
+        try {
+            const value = newSummaryPossibleAndNecessary ? await this.summarizeSession(existingSession) : existingSummary?.summary;
```
JonasHelming, line 98: This would call the TaskContextService instead

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
```diff
@@ -0,0 +1,115 @@
...
+            return value ? { ...request, value, contextValue: value } : undefined;
+        } catch (err) {
+            console.warn('Error retrieving chat session summary for session', request.arg, err);
+            return;
+        }
+    }
+
+    protected async summarizeSession(session: ChatSession): Promise<string> {
```
JonasHelming, line 106: Question 4: We discussed to create the summaries not on the fly, but at that moment, when the summary is created. For the current use case, this would mean if the user creates a summary from a chat and continues the summarized chat afterwards, the summary would not be automatically updated, but I would not expect this. fine?

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
```diff
@@ -0,0 +1,115 @@
...
+        const summary = await this.summaryAgent.generateChatSessionSummary(session);
```
JonasHelming, line 107: Question 5: We discussed adding a function summarizeSessionToTaskContext (with a better name maybe) to the TaskContextService. As parameters, it would get the chat session and optionally a prompt. The reason for the prompt is that there might be different focusses in the summarization, e.g.:

Summarize what has been done from coder for a follow up
Summarize the plan for Coder from the Architect

Both would probably have slightly different prompts.

packages/ai-chat/src/common/chat-session-summary-agent.ts
```diff
@@ -0,0 +1,118 @@
...
+    LanguageModelRequirement,
+    PromptService,
+    PromptTemplate,
+    UserRequest
+} from '@theia/ai-core';
+import { inject, injectable } from '@theia/core/shared/inversify';
+import { ChatSession } from './chat-service';
+import { generateUuid } from '@theia/core';
+
+const CHAT_SESSION_SUMMARY_PROMPT = {
```
JonasHelming, line 32: Question 6: Should this be an agent? I think it is nice as it gets all the infrastucture. The only confusing thing might be if we allow use case specific prompts for the summarization. We could treat the prompt of this agent as the default though and mention in the doc part that this might be overriden by specific calls.

packages/ai-ide/src/browser/coder-agent.ts
JonasHelming: Question 2.1: We should maybe add a button as well that opens something for showing even more actions, as there might be many suggetions, such as "update project-info" or common prompts. WDYT?

planger (commented):
First of all, I just want to say how much I like these powerful features added here—really great work!

A couple of thoughts from my side:

**Suggestions Handling**: I feel the suggestions are conceptually quite close to the chat session’s change set in their lifecycle, how they are created and updated, and also how they are shown to the user (as a dedicated area near the input, in reference to a chat session). So I feel like it might be better to manage these suggestions in a similar manner, i.e. directly within the chat session itself. That way, the agent or tools could update them dynamically based on the current context—like proposing follow-up questions or suggesting the user start a new task once the session only after it reached a certain length. That flexibility could really open up a lot of additional use cases.

**UI Representation**: I agree with Jonas—it’d be great to support going beyond markdown and support richer UI in suggestions area. That said, this raises some interesting questions. If we keep suggestions with the chat session and have them filled by agents or tools, it might feel awkward to directly create React nodes. So far, agents rather produce chat response nodes, and those are transferred to the UI using dedicated renderers. Should we stick with that model or would that be a misuse of the chat response infrastructure we already have?

**TaskContextService**: I also like the idea of pulling out the chat task management logic and persistence into a separate service and the summarization into a specific agent. That abstraction would make creating summaries and managing task contexts accessible not only through variables but also through a more direct API to agents and potentially other consumers. This feels like a good move for configurability of the summarization (LLMs and prompts) and reusability of reading or creating tasks.

Thanks again for the great work!

packages/ai-ide/src/browser/coder-agent.ts
JonasHelming: Answers for now:
1.: Do it like the change set (push to the chat session)
2.: Go with an extended MarkdownString impl for now, tha also allows call backs to the agent. We should however ideally mke the rendering more flexible in the future, i.e. that you can register a render for a type of "SuggestionContent" to also allow other UI here in the future
2.1: I would introduce this in the default renderer, i.e. actions in the text plus optionally additional actions that are shown under "more" in a popup.

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
JonasHelming: 4. fine

packages/ai-chat/src/common/chat-session-summary-agent.ts
JonasHelming: 6.: fine

packages/ai-ide/src/browser/coder-agent.ts
colin-grant-work: Done apart from 2.1 - added as todo in description.

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
colin-grant-work: Done.

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
colin-grant-work: No updates are expected, though if the summary is stored as a file, the user could go and modify it.

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
colin-grant-work: The TaskContextService accepts arguments for promptId's and agents. At the moment, it's assumed that the prompts will be known to the prompt service, but we could instead accept fully formed messages and not worry whether they came from prompts, and leave that to the caller?

packages/ai-chat/src/common/chat-session-summary-agent.ts
colin-grant-work: This is basically how it is now working: the summary agent is a minimal extension of the abstract agent implementation (useful for e.g. checking history), but it is possible to use other agents.

colin-grant-work:
> Should we stick with that model or would that be a misuse of the chat response infrastructure we already have?

I think that that is the closest analogue we have if we decide we want to allow LLM's to contribute suggestions of their own, rather than hardcoding them on specific agents. Then it wouldn't be an abuse of the system, but just the system: if the LLM is generating them, then they are just responses :-). For now, though, I haven't provided a mechanism for the LLM's to suggest things, but only for agents to provide suggestions. How they get them is left open, though.

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
JonasHelming: @colin-grant-work I think the summaries are still created only when the variable is resolved. I would create them when the user clicks the button. Is there any reason to only create them lazily? The lazy creation also introduces some weirdness: If I summarize a chat and I have a folder for persistence is configured, nothing happens until I either enter something in the new chat or click on the summary variable to see its content. This weird behavior would be gone if we summarize directly on clicking the action.

packages/ai-ide/src/common/coder-replace-prompt-template.ts
```diff
@@ -51,7 +51,11 @@ Instead, for each file you want to propose changes for:
 ${withSearchAndReplace ? ' If ~{changeSet_replaceContentInFile} continously fails use ~{changeSet_writeChangeToFile}. Calling a function on a file will override previous \
 function calls on the same file, so you need exactly one successful call with all proposed changes per changed file. The changes will be presented as a applicable diff to \
 the user in any case.' : ''}
-  
+
+## Previous Interactions and Background
```
JonasHelming, line 55: 1. I think this is too specific, because we can also use these new variables for new tasks
2. We should not put this heading here, but inside of the variable resolvement. If there is not context from this variable, the heading should also not appear.

packages/ai-ide/src/browser/context-session-summary-variable.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    async resolve(
+        request: AIVariableResolutionRequest,
+        context: AIVariableContext,
+        resolveDependency?: (variable: AIVariableArg) => Promise<ResolvedAIVariable | undefined>
+    ): Promise<ResolvedAIVariable | undefined> {
+        if (!resolveDependency || !ChatSessionContext.is(context) || request.variable.name !== CONTEXT_SESSION_MEMORY_VARIABLE.name) { return undefined; }
+        const allSummaryRequests = context.model.context.getVariables().filter(candidate => candidate.variable.id === SESSION_SUMMARY_VARIABLE.id);
+        const allSummaries = await Promise.all(allSummaryRequests.map(summaryRequest => resolveDependency(summaryRequest).then(resolved => resolved?.value)));
+        const value = allSummaries.map((content, index) => `# Context Memory ${index + 1}\n\n${content}`).join('\n\n');
```
JonasHelming, line 57: I would use the heading "Task Context" and then resolve the task contexts. Also this whole thing should resolve to an empty String in case no task contexts have been added to the session

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
```diff
@@ -0,0 +1,93 @@
...
+    description: 'Resolves to a summary of the session with the given ID.',
+    name: 'session-summary',
+    label: 'Session Summary',
+    iconClasses: codiconArray('clippy'),
+    isContextVariable: true,
+    args: [{ name: 'session-id', description: 'The ID of the session to summarize.' }]
+};
+
+@injectable()
+export class SessionSumaryVariableContribution implements FrontendVariableContribution, AIVariableResolver {
```
JonasHelming, line 37: I would rename the variable to "TaskContext" (and all related things such as the label provider)

JonasHelming: specifically also the variable the user types "#taskContext"

JonasHelming:
You can click now on any context variable, e.g. also instances of the file variable. However, for the file variable, it opens an empty read only editor for me. This is unexpected, it should open the underlying file instead. 
As this behavior is not really part of this feature, I am also perfectly fine with restricting the click to the new taskContext variable, only.

packages/ai-chat/src/browser/ai-chat-preferences.ts
```diff
@@ -38,6 +39,14 @@ If no Default Agent is configured, Theia´s defaults will be applied.'),
 You can manually unpin or switch agents anytime.'),
             default: true,
             title: AI_CORE_PREFERENCES_TITLE,
+        },
```
JonasHelming, line 42: We should move this pref to workspace-preferences.ts in the ide package. It only makes sense if tool have a workspace. Furthermore, I would move them under the "promptTemplates" category. They are related to prompt fragments.

JonasHelming:
Is there a reason why we make the editor for tastContexts read only for the case that there is no underlying file? Could we easily allow the user to modify it in memory instead? This can for sure be a follow-up, but we should capture it.

JonasHelming:
If persistence for taskContexts is set-up, the in memory version is still kept, so you have two versions of the same taskContex then.

- Create a task context (via Coder suggestion click)
- Resolve it (e.g. via saying something in a new chat
- type "#session-summary:" => You see two version of the tastContext, the file and the in-memory one

=> If persistence is active, we should fully get rid of the in memory ones I think

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
colin-grant-work: Yes, also added as a todo.

packages/ai-ide/src/browser/coder-agent.ts
```diff
@@ -38,5 +41,25 @@ export class CoderAgent extends AbstractStreamParsingChatAgent {
...
+    override async invoke(request: MutableChatRequestModel): Promise<void> {
+        await super.invoke(request);
+        this.suggest(request);
+    }
+    async suggest(context: ChatSession | ChatRequestModel): Promise<void> {
+        const model = ChatRequestModel.is(context) ? context.session : context.model;
+        const session = this.chatService.getSessions().find(candidate => candidate.model.id === model.id);
+        if (!(model instanceof MutableChatModel) || !session) { return; }
+        if (model.isEmpty()) {
+            model.setSuggestions([
```
JonasHelming, line 53: How do you test this, i.e. get to this state? If I write any message, the model is not empty anymore, correct?

packages/ai-ide/src/browser/coder-agent.ts
```diff
@@ -38,5 +41,25 @@ export class CoderAgent extends AbstractStreamParsingChatAgent {
...
     override functions = [GET_WORKSPACE_DIRECTORY_STRUCTURE_FUNCTION_ID, GET_WORKSPACE_FILE_LIST_FUNCTION_ID, FILE_CONTENT_FUNCTION_ID, WriteChangeToFileProvider.ID];
     protected override systemPromptId: string | undefined = CODER_REPLACE_PROMPT_TEMPLATE_ID;
-
+    override async invoke(request: MutableChatRequestModel): Promise<void> {
+        await super.invoke(request);
+        this.suggest(request);
+    }
+    async suggest(context: ChatSession | ChatRequestModel): Promise<void> {
+        const model = ChatRequestModel.is(context) ? context.session : context.model;
+        const session = this.chatService.getSessions().find(candidate => candidate.model.id === model.id);
```
JonasHelming, line 50: @planger Is this the expected way to get sessions now?

colin-grant-work:
> When I click on a taskContext variable in the chat which points to an underlying file, it opens a read only editor. This is unexpected, it should open the underlying file.

We can do this, but it would require different services. At the moment, we create one kind of resource and then populate that using the variable resolution system, which returns text. If we want to open the underlying file, then we have to add to the variable resolver interface the possibility of opening a variable so that resolver itself can decide whether to use a variable-specific resource or a standard URI. Would you like that as part of this PR?

packages/ai-ide/src/browser/coder-agent.ts
```diff
@@ -38,5 +41,25 @@ export class CoderAgent extends AbstractStreamParsingChatAgent {
...
+        if (model.isEmpty()) {
+            model.setSuggestions([
+                {
+                    kind: 'callback',
+                    callback: () => this.chatService.sendRequest(session.id, { text: '@Coder please look at #_f and fix any problems.' }),
+                    content: '[Fix problems](_callback) in the current file.'
+                },
+            ]);
+        } else {
+            model.setSuggestions([new MarkdownStringImpl(`Keep chats short and focused. [Start a new chat](command:${AI_CHAT_NEW_CHAT_WINDOW_COMMAND.id}) for a new task`
```
JonasHelming, line 61: @planger Should suggestions be part of the session or the model?

packages/ai-chat/src/browser/ai-chat-preferences.ts
colin-grant-work: The `ai-chat` package already depends on the `workspace` package, so there's no need to move this, at least from the perspective of dependency management. I'm willing to move it, but I don't really like having different pieces of the same functionality spread out if they don't have to be.

packages/ai-chat-ui/src/browser/chat-view-commands.ts
```diff
@@ -45,6 +45,10 @@ export const AI_CHAT_NEW_CHAT_WINDOW_COMMAND: Command = {
     iconClass: codicon('add')
 };
 
+export const AI_CHAT_NEW_WITH_MEMORY: Command = {
+    id: 'ai-chat.new-with-memory',
```
JonasHelming, line 49:
```suggestion
    id: 'ai-chat.new-with-task-context',
```

packages/ai-ide/src/browser/coder-agent.ts
JonasHelming: I have created 2.1 as a follow up in https://github.com/eclipse-theia/theia/issues/15068 so we can close it here.

JonasHelming:
> > When I click on a taskContext variable in the chat which points to an underlying file, it opens a read only editor. This is unexpected, it should open the underlying file.
> 
> We can do this, but it would require different services. At the moment, we create one kind of resource and then populate that using the variable resolution system, which returns text. The editor is readonly (and not hooked up to language services, etc.) because the variable resolution system doesn't know anything except that it can get a string. If we want to open the underlying file, then we have to add to the variable resolver interface the possibility of opening a variable so that resolver itself can decide whether to use a variable-specific resource or a standard URI. Would you like that as part of this PR?
> 
> > Is there a reason why we make the editor for tastContexts read only for the case that there is no underlying file? Could we easily allow the user to modify it in memory instead? This can for sure be a follow-up, but we should capture it.
> 
> The answer is the same for this: we're not opening the file, we're retrieving the value of a variable.

I see! @planger WDYT about this? I think we will want this anyways, also for the file variable, right?

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
JonasHelming: I see, that was this "(or create it earlier)" :-)

packages/ai-chat/src/browser/session-summary-variable-contribution.ts
JonasHelming: I like this a lot for now!

packages/ai-chat/src/common/chat-session-summary-agent.ts
JonasHelming: Yeah, I already love it as I can see/edit the prompt and see the history

packages/ai-ide/src/browser/coder-agent.ts
colin-grant-work: Placed there for now because it's currently where change sets and context also live.

packages/ai-chat/src/browser/ai-chat-preferences.ts
JonasHelming: It's not about dependency management. It is more about keeping the core clean from IDEish or more specific things. And having a workspace relative directory for taskContexts is something that probably many tool would not want.

For this reason, I would turn around the flow. The Task-Context-Service has a "setter" for a location. The IDE package can set this to a workspace related directory if the pref is set. So the task context service interface is clean and the task context service does not know the user pref.

Adopters can then choose to set a static directory for example or do a custom pref.

packages/ai-ide/src/common/coder-replace-prompt-template.ts
JonasHelming: Tested it a bit, we should put this at the end of the prompt (plus the comments above)

packages/ai-ide/src/browser/coder-agent.ts
JonasHelming: We can definitily also move this to a follow up, i.e. the ability to show suggestions early

packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
```diff
@@ -83,10 +84,21 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
...
         });
+        registry.registerCommand(AI_CHAT_NEW_WITH_MEMORY, {
+            execute: () => {
+                const activeSessions = this.chatService.getSessions().filter(candidate => candidate.isActive);
+                if (activeSessions.length !== 1) {
+                    return;
+                }
+                const activeSession = activeSessions[0];
+                const newSession = this.chatService.createSession(ChatAgentLocation.Panel, { focus: true }, activeSession.pinnedAgent);
+                newSession.model.context.addVariables({ variable: SESSION_SUMMARY_VARIABLE, arg: activeSession.id });
```
JonasHelming, line 98: I just tried out to adopt this feature in another agent which leads again to the issue that the actual context creation happens inside of the resolvement. I would make the variable "stupid". Caller have to create the context themselve using the service. Otherwise you can for example as an agent creating a context not influence the prompt, although the task service would allow to do this. So the variable would only get the id of the task context and this task context must exist, otherwise it will just not resolve.

I would expect this here to be like:
id = task-context-service.summarizeAndStore(activeSession)
newSession.model.context.addVariables({ variable: SESSION_SUMMARY_VARIABLE, arg: id });

Does this make sense?

colin-grant-work:
> => If persistence is active, we should fully get rid of the in memory ones I think

With the files, we have competing demands:
 - The filename should give some idea of the content.
 - For files derived from sessions, it should still be possible to identify the session.
 - The filename should be reasonably formatted (not too long, e.g.).

The ID is the only guaranteed unique identifier of the session but isn't a legible filename, and the session label is a nice basis for a filename, but it isn't guaranteed to be unique and it may be very long and not a very nice filename. The compromise that I've aimed for is to keep a reference to the sessionId and full label as frontmatter in the summary file and use in-memory storage to make sure that we don't have to read every file on disk to figure out if we already have a summary for a given session and a formatted filename for storage.

To address the problems you saw with the same session appearing multiple times, I've improved the matching and filtering logic, and also filtered out the current session.

packages/ai-chat/src/browser/ai-chat-preferences.ts
colin-grant-work: I have a little bit of an aversion to setters - harder to figure out who's setting what when - so I've added an additional interface for `TaskContextStorageService` with the default implementation being the in-memory version and then an additional preference+file-based version in the `ai-ide` package. How does that sound?

JonasHelming: that sounds very good to me!

packages/ai-ide/src/browser/task-background-summary-variable.ts
```diff
@@ -0,0 +1,67 @@
...
+export class TaskContextSummaryVariableContribution implements AIVariableContribution, AIVariableResolverWithVariableDependencies {
+    registerVariables(service: AIVariableService): void {
+        service.registerResolver(TASK_CONTEXT_SUMMARY_VARIABLE, this);
+    }
+
+    canResolve(request: AIVariableResolutionRequest, context: AIVariableContext): MaybePromise<number> {
+        return request.variable.name === TASK_CONTEXT_SUMMARY_VARIABLE.name ? 50 : 0;
+    }
+
+    async resolve(
```
JonasHelming, line 52: If there is no summary, this still resolves to {{taskContextSummary}}, we should explicitly resolve to empty String like done here: packages/ai-core/src/common/prompt-variable-contribution.ts

packages/ai-ide/src/browser/task-context-file-storage-service.ts
```diff
@@ -0,0 +1,174 @@
...
+            { dispose: () => this.clearInMemoryStorage() },
+        );
+        await this.cacheNewTasks(newStorage);
+    }
+
+    protected async handleChanges(changes: FileChange[]): Promise<void> {
+        await Promise.all(changes.map(change => {
+            switch (change.type) {
+                case FileChangeType.DELETED: return this.deleteFileReference(change.resource);
+                case FileChangeType.ADDED:
```
JonasHelming, line 76: Can we implement this case? If the user manually adds a file to the directory, we should consider it.

JonasHelming:
I tested this a bit. The core use case works very well, great work. I found the following issues:

1. If I summarize a session and then in the new session zype "#taskSummary:" I get the created summary twice in auto-completion, so there is still some duplication happening
2. If the user adds an empty file to the task context directory, this is not picked up (because "summary" is empty/undefined). While you could argue about the sense of empty summaries, I would still pick empty files up.
3. Only one manual added file is picked up (the first). So if you add multiple files yourselve to the task context directory, you can only use the first.

packages/ai-ide/src/browser/task-background-summary-variable.ts
colin-grant-work: Hm... given that `undefined` is a legitimate return value according to the interface, it seems to me that it ought to be treated the same as an `''` for the purposes of printing the prompt - is there any circumstance in which printing the `{{}}` template value in place makes sense when the return is `undefined`?

packages/ai-ide/src/browser/task-context-file-storage-service.ts
colin-grant-work: It falls through and is handled the same as modification: the file is read and the summary added / updated.

colin-grant-work:
> Only one manual added file is picked up (the first). So if you add multiple files yourselve to the task context directory, you can only use the first.

Hm... sounds like a problem with file watching, since we certainly don't dispose of the watcher after the first addition. I'll see whether the events are coming through.

packages/ai-ide/src/browser/task-background-summary-variable.ts
JonasHelming: Yes, for variables, where we allow "empty" I would return ''. For variables which should be resolved (and by default) we return {{}} to ease debugging (you can see that the variable has not been resolved. This is maybe not perfect, but this is how I would go for now (see the example I posted where also "empty" is a valid value.

JonasHelming:
> > Only one manual added file is picked up (the first). So if you add multiple files yourselve to the task context directory, you can only use the first.
> 
> Hm... sounds like a problem with file watching, since we certainly don't dispose of the watcher after the first addition. I'll see whether the events are coming through.

You can even override the first one by adding one that is "before" in alphabetical order. Also after restart there is only one, so it is likely not the watching I believe.

colin-grant-work:
> You can even override the first one by adding one that is "before" in alphabetical order. Also after restart there is only one, so it is likely not the watching I believe.

Yes, turned out to be careless sessionID comparison so that any summary with no session Id was replacing any other with none.

colin-grant-work:
> If I summarize a session and then in the new session zype "#taskSummary:" I get the created summary twice in auto-completion, so there is still some duplication happening.

I'm not able to reproduce this - can you post a screen capture?

packages/ai-ide/src/browser/coder-agent.ts
planger: I think in this case we don't even need to retrieve the actual chat session, because `model.id` should always be equivalent to the `ChatSession.id` (see chat-serivce.ts:187).

In general, we do have a distinction there:

  * The `ChatModel` encompasses all information on a conversation (also referred to as session in the chat model, e.g. `ChatRequestModel.session`).
  * The Chat Service wraps each `ChatModel` into another object, called `ChatSession`, which also adds more UI-related state relevant for the chat service, such as whether the session is currently active in the chat view, the currently pinned agent, last interaction, etc.

But the IDs should always be the same.

packages/ai-chat/src/common/chat-agents.ts
```diff
@@ -126,6 +127,7 @@ export interface ChatAgent extends Agent {
     locations: ChatAgentLocation[];
     iconClass?: string;
     invoke(request: MutableChatRequestModel, chatAgentService?: ChatAgentService): Promise<void>;
+    suggest?(context: ChatRequestModel | ChatSession): Promise<void>;
```
JonasHelming, line 130:
```suggestion
```
Any reason this needs to be in the interface?

colin-grant-work: Without that, we can't e.g. call `suggest` on the pinned agent when a session is created with one:

```
packages/ai-chat/src/common/chat-service.ts(193,40): error TS2339: Property 'suggest' does not exist on type 'ChatAgent'.
```

In general, if we want it known that making suggestions may be a capability of agents but isn't an obligatory one, we have to put it in the interface so that we can call it even if we don't know exactly which agent we're dealing with.

packages/ai-ide/src/browser/coder-agent.ts
colin-grant-work: > because model.id should always be equivalent to the ChatSession.id (see chat-serivce.ts:187).
> ... But the IDs should always be the same.

This is (currently) true (for our implementation of ChatSessions), but not guaranteed by the interfaces / API, so I'd prefer to keep the structure that matches ID's on elements that are supposed to be the same.

The alternative is the `getSession` API, which does the same iteration, so this isn't much less efficient:

https://github.com/eclipse-theia/theia/blob/505c885de592efcb0c86e2b8ff6b270d091d9553/packages/ai-chat/src/common/chat-service.ts#L180-L182

Alternatively, we could a reference to the session on the model, or API on the chat service to retrieve a session given a model?

packages/ai-chat/src/common/chat-agents.ts
JonasHelming: But where do we currently call it from outside of the agent?

JonasHelming: We can add this in the future, if need I think

JonasHelming:
> > > If I summarize a session and then in the new session zype "#taskSummary:" I get the created summary twice in auto-completion, so there is still some duplication happening.
> > 
> > 
> > I'm not able to reproduce this - can you post a screen capture?
> 
> reproduce [ reproduce ](https://private-user-images.githubusercontent.com/1052615/436129339-8f28895b-16cf-4361-8093-95e110613a14.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDU0MTM4MjQsIm5iZiI6MTc0NTQxMzUyNCwicGF0aCI6Ii8xMDUyNjE1LzQzNjEyOTMzOS04ZjI4ODk1Yi0xNmNmLTQzNjEtODA5My05NWUxMTA2MTNhMTQuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDQyMyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA0MjNUMTMwNTI0WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ZWE1OTFkOTRlOGIxYTAwMTcyYzM2ODgwYjY4NjQxOTc3Y2JiYzg4YmY4ZjNlODVmNjNjZTJiMDhkNDg5ZTA3OSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.SuCG0SyCo8emfJhVyalpPNFjkfIv5m9_54OWn_KAjcw) [ ](https://private-user-images.githubusercontent.com/1052615/436129339-8f28895b-16cf-4361-8093-95e110613a14.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDU0MTM4MjQsIm5iZiI6MTc0NTQxMzUyNCwicGF0aCI6Ii8xMDUyNjE1LzQzNjEyOTMzOS04ZjI4ODk1Yi0xNmNmLTQzNjEtODA5My05NWUxMTA2MTNhMTQuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDQyMyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA0MjNUMTMwNTI0WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ZWE1OTFkOTRlOGIxYTAwMTcyYzM2ODgwYjY4NjQxOTc3Y2JiYzg4YmY4ZjNlODVmNjNjZTJiMDhkNDg5ZTA3OSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.SuCG0SyCo8emfJhVyalpPNFjkfIv5m9_54OWn_KAjcw)

I cannot reproduce this any longer so nevermind.

JonasHelming:
> I tested this a bit. The core use case works very well, great work. I found the following issues:
> 
> 1. If I summarize a session and then in the new session zype "#taskSummary:" I get the created summary twice in auto-completion, so there is still some duplication happening
> 2. If the user adds an empty file to the task context directory, this is not picked up (because "summary" is empty/undefined). While you could argue about the sense of empty summaries, I would still pick empty files up.
> 3. Only one manual added file is picked up (the first). So if you add multiple files yourselve to the task context directory, you can only use the first.

All these issues are solved now!

packages/ai-chat/src/browser/task-context-variable.ts
```diff
@@ -0,0 +1,28 @@
...
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { AIVariable } from '@theia/ai-core';
+import { codiconArray } from '@theia/core/lib/browser';
+
+export const TASK_CONTEXT_VARIABLE: AIVariable = {
+    id: 'taskContext',
+    description: 'Provides background information on task planning, particularly summaries of chat sessions.',
```
JonasHelming, line 22:
```suggestion
    description: 'Provides context information for a task, e.g. the plan for completing a task or a summary of a previous sessions',
```

packages/ai-chat/src/browser/task-context-variable-label-provider.ts
```diff
@@ -0,0 +1,67 @@
+// *****************************************************************************
+// Copyright (C) 2025 Eclipse GmbH and others.
```
JonasHelming, line 2:
```suggestion
// Copyright (C) 2025 EclipseSource GmbH and others.
```

packages/ai-chat/src/common/chat-session-summary-agent.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const CHAT_SESSION_SUMMARY_PROMPT = {
+    id: 'chat-session-summary-prompt',
+    template: '{{!-- Made improvements or adaptations to this prompt template? We\'d love for you to share it with the community! Contribute back here: ' +
+        'https://github.com/eclipse-theia/theia/discussions/new?category=prompt-template-contribution --}}\n\n' +
+        'You are a chat agent for summarizing AI agent chat sessions for later use. \
+Review the conversation above and generate a concise summary that captures every crucial detail, \
+including all requirements, decisions, and pending tasks. \
+Ensure that the summary is sufficiently comprehensive to allow seamless continuation of the workflow. The summary will primarily be used by other AI agents, so tailor your \
+response for use by AI agents.',
```
JonasHelming, line 32:
```suggestion
response for use by AI agents.
Also consider the system message.
Make sure you include all necessary context information and use unique references (such as URIs, file paths, etc.).
If the conversation was about a task, describe the state of the task, i.e. what has been completed and what is open.
If a changeset has been used in the session, describe the state of the suggested changes.',
```

packages/ai-chat/src/common/chat-session-summary-agent.ts
```diff
@@ -0,0 +1,51 @@
...
+// *****************************************************************************
+
+import {
+    LanguageModelRequirement,
+    PromptTemplate
+} from '@theia/ai-core';
+import { injectable } from '@theia/core/shared/inversify';
+import { AbstractStreamParsingChatAgent, ChatAgent } from './chat-agents';
+
+export const CHAT_SESSION_SUMMARY_PROMPT = {
```
JonasHelming, line 24: Please see my suggestion below. Also, please move this to a separate file under MIT (see other agents) and also use the new comment (including MIT reference)

packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
```diff
@@ -211,6 +243,20 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
     canExtractChatView(chatView: ChatViewWidget): boolean {
         return !chatView.secondaryWindow;
     }
+
+    protected getActiveSession(): ChatSession | undefined {
```
JonasHelming, line 247: Should we just add a getActiveSession to the ChatService?

packages/core/src/browser/icon-theme-service.ts
```diff
@@ -95,7 +95,7 @@ export class IconThemeService {
         return this._iconThemes.get(id);
     }
 
-    @inject(NoneIconTheme) protected readonly noneIconTheme: NoneIconTheme;
+    @inject(NoneIconTheme) protected readonly noneIconTheme: NoneIconTheme; // @CJG - something here!
```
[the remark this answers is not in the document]
colin-grant-work: Investigations for other bugs sneaking in. :-/

JonasHelming:
This works very nicely, thanks!

Just found one more functional issue:

If I use the TaskContext variables without "#taskContext" like shown below, they do not resolve (i cannot click them and they resolve to undefined). The first in the video does not work, the second does.

reproduce2

packages/core/src/common/resource.ts
colin-grant-work, line 1: The changes to `MutableResource` and `ReferenceMutableResource` here are aimed at updating them with some new fields from `Resource`, `autosaveable`, `readonly`, `initiallyDirty`, and making those fields and the saving behavior of the `MutableResource` a bit more configurable. I had implemented them separate in the `ChangeSetResourceResolver` in part to avoid disturbing these existing implementations, but then I wanted to use similar features in several places and thought it might then be worth adding these features to the base implementations. I believe that I've set things up so that existing callers using the pre-existing API will get the same behavior, but modifying these classes runs afoul of the open-closed principle.

So, @tsmaeder or @msujew, what do you think of modifying these classes vs. introducing something like an `InMemoryResources2` with separate implementations to protect adopters from any breakage if they've extended these classes?

packages/ai-chat/src/common/chat-session-summary-agent-prompt.ts
```diff
@@ -0,0 +1,27 @@
...
+// See LICENSE-MIT.txt in the project root for license information.
+// https://opensource.org/license/mit.
+//
+// SPDX-License-Identifier: MIT
+
+import { CHANGE_SET_SUMMARY_VARIABLE_ID } from './context-variables';
+
+export const CHAT_SESSION_SUMMARY_PROMPT = {
+    id: 'chat-session-summary-prompt',
+    template: `{{!-- Made improvements or adaptations to this prompt template? We\'d love for you to share it with the community! Contribute back here: ' +
```
JonasHelming, line 15:
```suggestion
    template: `{{!-- !-- This prompt is licensed under the MIT License (https://opensource.org/license/mit).
Made improvements or adaptations to this prompt template? We\'d love for you to share it with the community! Contribute back here: ' +
```

packages/ai-chat/src/common/chat-agents.ts
JonasHelming: So What I meant is, I would not expose this function now, because we do not need it anywhere

packages/ai-chat/src/common/chat-session-summary-agent-prompt.ts
colin-grant-work: It looks like we haven't done this - either the move to an MIT-licensed licensed or the addition of the MIT license notice to the prompt - with the prompt for chat session naming, which is what I based this one off of. Do we want to do either of those things?

packages/core/src/common/resource.ts
tsmaeder: This feels like we're importing a whole lot of complexity into a very simple service. We have the `Resource` interface to abstract over different behaviours of resources. If we want to share implementation, that's a different story, but here most of the implementation seems to be used only in the AI packages, so my clear preference would be to keep that complexity out of the previous implementation.

packages/core/src/common/resource.ts
```diff
@@ -224,29 +224,54 @@ export class DefaultResourceProvider {
 
 }
 
+export type ResourceInitializationOptions = Pick<Resource, 'autosaveable' | 'initiallyDirty' | 'readOnly'>
```
tsmaeder, line 227: This is hard to understand as opposed to a simple interface type.

packages/core/src/common/resource.ts
```diff
@@ -224,29 +224,54 @@ export class DefaultResourceProvider {
...
+    & { contents?: string | Promise<string>, onSave?: Resource['saveContents'] };
+
 export class MutableResource implements Resource {
-    protected contents: string = '';
+    protected contents: string | Promise<string>;
+
+    constructor(readonly uri: URI, protected options?: ResourceInitializationOptions) { }
```
tsmaeder, line 233: This makes all of the fields in this class effectively mutable.

packages/core/src/common/resource.ts
colin-grant-work: @tsmaeder, I would view it another way. The previous implementation offered limited access to the resource interface: the resources couldn't be initially dirty, couldn't set readonly, couldn't be autosaveable or not. That was mostly because some of those fields weren't available when those resource types were written, and adding them seems to be broadly useful.

That means that the main innovations here, apart from more getters on the `ReferenceMutableResource` are:
- the ability to update those fields post-construction by passing in new options. 
- ability to separate content update (equivalent to say, updating when a resource changes on disk) from saving (in response to a user action).
- the ability to customize what 'saving' means for a given resource.

It's true, the saving part in particular is mostly to be used in the AI packages, but the functionality isn't AI-specific. In one case, I want to retain a record of the value after save even if the editor is closed, and the current system doesn't allow that with a reasonable lifecycle - I'd have to keep a reference to the resource itself forever - and in another case, I need to change content on disk in another location when the user chooses to save. Neither of those is necessarily bound to AI use cases, that just happens to be where I've been working, so it's where I've felt that these capabilities would be useful.

packages/core/src/common/resource.ts
colin-grant-work: I consider that an advantage, given that it does have `Mutable` in its name.

packages/core/src/common/resource.ts
colin-grant-work: It's also guaranteed to maintain the correct type if anything changes in `Resource`. For example, `readonly` had changed from `boolean` to `boolean | MarkdownString` since the original declaration was written, and the declaration hadn't kept up.

sdirix (commented):
Thanks for the great work. All comments from mine besides the regression mentioned below could be handled in follow ups if the time until release runs out.

---

Architecturally I understand that each agent can add own suggestions. However it's a bit weird to me that such a generic suggestion like we have now for the Coder is not handled generically too.

Could be a good follow up.

---

**Blocker regression**: The edit chat button UI is broken because the input is not shown anymore. 
image

**Minor**: The pointer is shown over the whole text instead of only the actions: 
image

packages/ai-core/src/common/ai-variable-resource.ts
```diff
@@ -0,0 +1,86 @@
...
+@injectable()
+export class AIVariableResourceResolver {
+    @inject(InMemoryResources) protected readonly inMemoryResources: InMemoryResources;
+
+    @postConstruct()
+    protected init(): void {
+        this.inMemoryResources.onWillDispose(resource => this.cache.delete(resource.uri.toString()));
+    }
+
+    protected readonly cache = new Map<string, [Resource, AIVariableContext]>();
```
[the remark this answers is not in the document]
colin-grant-work: I think that would only work if we expected to get the very same context back; the idea with `deepEqual` is that we can tell if the context is the same even if it's not the very same object.

packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
```diff
@@ -139,6 +164,14 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
...
             isVisible: widget => this.withWidget(widget),
         });
+        const sessionSummarizibilityChangedEmitter = new Emitter<void>();
+        this.taskContextService.onDidChange(() => sessionSummarizibilityChangedEmitter.fire());
+        this.chatService.onSessionEvent(event => event.type === 'activeChange' && sessionSummarizibilityChangedEmitter.fire());
+        registry.registerItem({
+            id: 'chat-view.' + AI_CHAT_SUMMARIZE_CURRENT_SESSION.id,
+            command: AI_CHAT_SUMMARIZE_CURRENT_SESSION.id,
+            onDidChange: sessionSummarizibilityChangedEmitter.event
+        });
```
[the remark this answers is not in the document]
colin-grant-work: Added a label that should fix this.

packages/ai-chat-ui/src/browser/chat-view-commands.ts
```diff
@@ -45,6 +45,15 @@ export const AI_CHAT_NEW_CHAT_WINDOW_COMMAND: Command = {
     iconClass: codicon('add')
 };
 
+export const AI_CHAT_NEW_WITH_TASK_CONTEXT: Command = {
```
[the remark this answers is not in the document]
colin-grant-work: Good question. I came in at the bottom of the file and started adding commands in the style I found there, which happened to be outside the namespace. It looks like the breakout from the namespace started with the commit that added the initial LLM support:

image

packages/ai-chat-ui/src/browser/ai-chat-ui-contribution.ts
```diff
@@ -85,14 +90,34 @@ export class AIChatContribution extends AbstractViewContribution<ChatViewWidget>
...
+        registry.registerCommand(AI_CHAT_SUMMARIZE_CURRENT_SESSION, {
+            execute: async () => this.summarizeActiveSession(),
+            isVisible: widget => {
+                if (widget && !this.withWidget(widget)) { return false; }
+                const activeSession = this.chatService.getActiveSession();
+                return !!activeSession?.model.getRequests().length
+                    && activeSession?.model.location === ChatAgentLocation.Panel
+                    && !this.taskContextService.hasSummary(activeSession);
+            }
```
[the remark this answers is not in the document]
colin-grant-work: Added a command to open a summary if available.

packages/ai-chat/src/common/chat-session-summary-agent.ts
```diff
@@ -0,0 +1,41 @@
...
+import {
+    LanguageModelRequirement,
+    PromptTemplate
+} from '@theia/ai-core';
+import { injectable } from '@theia/core/shared/inversify';
+import { AbstractStreamParsingChatAgent, ChatAgent } from './chat-agents';
+import { CHAT_SESSION_SUMMARY_PROMPT } from './chat-session-summary-agent-prompt';
+
+@injectable()
+export class ChatSessionSummaryAgent extends AbstractStreamParsingChatAgent implements ChatAgent {
```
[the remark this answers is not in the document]
colin-grant-work: Removed tag.

packages/ai-chat/src/browser/task-context-service.ts
```diff
@@ -0,0 +1,138 @@
...
+    open(identifier: string): Promise<void>;
+}
+
+@injectable()
+export class TaskContextService {
+
+    protected pendingSummaries = new Map<string, Promise<Summary>>();
+
+    @inject(ChatService) protected readonly chatService: ChatService;
+    @inject(ChatSessionSummaryAgent) protected readonly summaryAgent: ChatSessionSummaryAgent;
```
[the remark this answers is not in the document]
colin-grant-work: Now retrieving from agent service.

colin-grant-work:
> Blocker regression: The edit chat button UI is broken because the input is not shown anymore.

Fixed the initialization of the new `AIChatTreeInputWidget` so that the error causing the apparent disappearance wouldn't occur.

> Minor: The pointer is shown over the whole text instead of only the actions:

Tweaked the check here so that it wouldn't appear on non-callback suggestions.

packages/ai-chat/src/common/chat-session-summary-agent.ts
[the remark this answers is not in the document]
sdirix: I think we should not even implement the `ChatAgent` interface

sdirix (approved):
Works for me! Thanks for the work!

**Potential follow up:**
The agent can now be successfully disabled, however the UI does not react.

- If the user tries to trigger a summary, they do not get an error shown, so they don't know what is going wrong (maybe they forgot that they disabled the agent)
- In case the agent is disabled, we could hide the "start new chat with summary" text in case the summary does not exist yet (as it will not be creatable)
- In case the agent is disabled, we could hide the "summary chat" action in the toolbar (but keep the "show summary one)

JonasHelming (approved):

Issue #14348, opened 2024-10-23
Cannot Open Terminal on Windows Remote


parisa-mchp:
### Bug Description:

On linux, after building the theia electron application, and connecting to a windows server over SSH, terminal widgets can be created, but cannot be interacted with.

### Steps to Reproduce:

1. On linux, build and launch the theia electron app using the following commands (taken from [here](https://github.com/eclipse-theia/theia/blob/master/doc/Developing.md))

```sh
git clone https://github.com/eclipse-theia/theia \
    && cd theia \
    && yarn \
    && yarn download:plugins \
    && yarn browser build \
    && yarn browser start
```

2. Using the command `SSH: Connect Current Window to Host...` connect to a windows server.
3. Once connected, use `Terminal: Create New Integrated Terminal from a Profile` and select the `cmd` profile.
4. A new terminal should be created, but it's output should be blank. Furthermore clicking on the terminal doesn't allow you to type in it.

![{86FC3A9A-9C91-40AF-8064-CFA993DC29D1}](https://github.com/user-attachments/assets/6cceb9cd-4584-4db7-914b-96c0b88c1c1e)

### Additional Information

Although building the IDE locally causes this issue, it works fine if using the [prepackaged appimage](https://www.eclipse.org/downloads/download.php?file=/theia/ide/latest/linux/TheiaIDE.AppImage&r=1). I wanted to test building the appimage locally to see if this addresses the issue, but couldn't figure out how to do this. It's mentioned in the [documentation ](https://theia-ide.org/docs/blueprint_documentation) that the `yarn electron package` command does this, but looking at `example/electron/package.json` this command doesn't seem to exist. More information on this would be appreciated as well.

- Operating System: Ubuntu 22.04.2 LTS (WSL)
- Theia Version: 1.54.0

tsmaeder:
@msujew ?

msujew:
Hey @parisa-mchp,

> It's mentioned in the [documentation ](https://theia-ide.org/docs/blueprint_documentation)that the yarn electron package command does this, but looking at example/electron/package.json this command doesn't seem to exist.

You're looking in the wrong repo. See [here](https://github.com/eclipse-theia/theia-ide/blob/1fc2fc267893586d146581dcd0b979a89cbd5018/applications/electron/package.json#L146)

[...]

PR #15443, opened 2025-04-10
Fix ssh terminals

parisa-mchp:
#### What it does

This PR addresses an [issue](https://github.com/eclipse-theia/theia/issues/14348#issuecomment-2500722794) where from a linux client connecting to a windows server with by ssh, you cannot open terminals. This is because the node-pty module that we rely on to create terminals is compiled in a platform specific way, so when the node backend is copied onto the server, an incorrect version of node-pty is provided. For some reason, this does not cause issues making the reverse connection, from a windows client to a linux server.

To fix this, we do a webpack exclude so that the node-pty module is resolved at run-time instead of at compile time, and as part of the server setup we install node-pty so a correct version will be available for the server's platform. We don't exclude node-pty when not connected to a server. Otherwise, additional care would need to be taken when creating a theia extension to ensure any generated appimages include the node-pty module, and that a remote, which would then have two copies of the module, resolve the correct one.

#### How to test

To reproduce the issue, connect from a linux client to a windows server, and attempt to open an ide terminal. Any other client-server combination should still work. I have tested linux-linux, linux-windows, windows-linux, and windows-windows, but have don't have access to any systems running macos. When connecting to a server twice with different builds, make sure to delete the `.theia-example-electron-1.60.0-remote` folder.

#### Follow-ups

The current implementation is somewhat complex. Suggestions would be appreciated, but at the moment I don't see a way to simplify it.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/dev-packages/native-webpack-plugin/src/native-webpack-plugin.ts
+++ b/dev-packages/native-webpack-plugin/src/native-webpack-plugin.ts
@@ -25,6 +25,7 @@ const REQUIRE_VSCODE_WINDOWS_CA_CERTS = '@vscode/windows-ca-certs';
 const REQUIRE_BINDINGS = 'bindings';
 const REQUIRE_KEYMAPPING = './build/Release/keymapping';
 const REQUIRE_PARCEL_WATCHER = './build/Release/watcher.node';
+const REQUIRE_NODE_PTY_CONPTY = '../build/Release/conpty.node';
 
 export interface NativeWebpackPluginOptions {
     out: string;
@@ -71,6 +72,11 @@ export class NativeWebpackPlugin {
                 [REQUIRE_VSCODE_WINDOWS_CA_CERTS]: windowsCaCertsFile,
                 [REQUIRE_PARCEL_WATCHER]: issuer => Promise.resolve(findNativeWatcherFile(issuer))
             };
+            if (process.platform !== 'win32') {
+                // The expected conpty.node file is not available on non-windows platforms during build.
+                // We need to provide a stub that will be replaced by the real file at runtime.
+                replacements[REQUIRE_NODE_PTY_CONPTY] = () => buildFile(directory, 'conpty.js', conhostWindowsReplacement());
+            }
         });
         compiler.hooks.normalModuleFactory.tap(
             NativeWebpackPlugin.name,
@@ -208,3 +214,7 @@ ${cases.join(os.EOL)}
     throw new Error(\`unhandled module: "\${jsModule}"\`);
 }`.trim();
 };
+
+const conhostWindowsReplacement = (nativePath: string = '.'): string => `
+module.exports = __non_webpack_require__('${nativePath}/native/conpty.node');
+`;


packages/remote/src/electron-node/setup/remote-setup-service.ts
```diff
@@ -91,18 +91,36 @@ export class RemoteSetupService {
...
+            report('Installing node modules');
+            let prefix;
+            if (platform.os === OS.Type.Windows) {
+                // We might to switch to PowerShell beforehand on Windows
+                prefix = this.scriptService.exec(platform);
+            } else {
+                // ensure the node binary is on path so the node modules are compiled for the correct version
+                prefix = `export PATH=${nodeBinDir}:$PATH;`;
+            }
+            await connection.exec(`${prefix} cd ${libDir}; ${npmExecutable} init -y; ${npmExecutable} install node-pty tslib`);
```
msujew, line 111: This might not work if the remote machine doesn't have a "normal" network connection. I.e. outbound HTTP traffic might be blocked to some degree.

msujew (commented):
I'm not sure this is a good solution to the issue. Is there something wrong with the binaries we're uploading? Can you try using the code from `master` and just replace the binaries for node-pty with ones that come with the installation of `node-pty` on Windows?

parisa-mchp:
> Can you try using the code from master and just replace the binaries for node-pty with ones that come with the installation of node-pty on Windows?

This is a little bit tricky if you look at the way `windowsPtyAgent` gets compiled it looks like this:

```javascript
var WindowsPtyAgent = /** @class */ (function () {
    function WindowsPtyAgent(file, args, env, cwd, cols, rows, debug, _useConpty, _useConptyDll, conptyInheritCursor) {
        var _this = this;
        if (_useConptyDll === void 0) { _useConptyDll = false; }
        if (conptyInheritCursor === void 0) { conptyInheritCursor = false; }
        this._useConpty = _useConpty;
        this._useConptyDll = _useConptyDll;
        this._pid = 0;
        this._innerPid = 0;
        if (this._useConpty === undefined || this._useConpty === true) {
            this._useConpty = this._getWindowsBuildNumber() >= 18309;
        }
        if (this._useConpty) {
            if (!conptyNative) {
                try {
                    conptyNative = __webpack_require__(Object(function webpackMissingModule() { var e = new Error("Cannot find module '../build/Release/conpty.node'"); e.code = 'MODULE_NOT_FOUND'; throw e; }()));
                }
```

That line at the bottom being the problematic one. If we add the `conpty.node` post hoc, it won't use it because at compile time the error is hard coded. To circumvent this, I tried manually copying conpty into the module between the `npm install` and `npm run build:electron` steps, but the above file was still compiled the same way by webpack. I wasn't sure why this is, I suspect it has either to do with the fact that the module is already built when we run `npm install`, or as a result of some sort of dead code elimination step that assumes the module isn't need because we're compiling on linux.

All of this is to say, I suspect what you've suggested is doable, but I'm not sure how to.

msujew:
> This is a little bit tricky if you look at the way windowsPtyAgent gets compiled it looks like this

Ah, thanks a lot, that explains it well! Alright, I think that we probably can get a simpler solution by performing a bit of bundler magic in Webpack. I'll prepare something next week 👍

parisa-mchp:
> I think that we probably can get a simpler solution by performing a bit of bundler magic in Webpack.

Where's this at? Would you like me to look into it implementing the idea you described?

msujew:
@parisa-mchp I actually created a WIP over at https://github.com/eclipse-theia/theia/compare/master...msujew/fix-remote-windows-terminals but didn't have time to get back to it (I also currently don't have a setup in which I can easily test this). The change might work out of the box, but it shows the approach how I would fix this. If you have some time to dig into it, please go ahead and continue on my changes :)

parisa-mchp:
@msujew Thanks so much for providing this. I don't know webpack too well so I would have had a hard time doing this part. From here I think I can get something working.

msujew:
@parisa-mchp For some reason, the ECA validation doesn't work correctly with my email in the commit. Feel free to remove my attribution from the commit 👍

parisa-mchp:
Okay, right now everything seems to be working on my end. As before, I don't have a way to test macos, but every combination of windows and linux I tried seemed okay.

As for the test failures, I think there's some kind of outage in one of the servers the plugins are downloaded from. It was working fine earlier today for me, but the `download:plugins` step doesn't seem to be working.

msujew:
@parisa-mchp It looks like the Eclipse foundation service infrastructure faces a major outage right now. That's the reason for the failing pipelines. Nothing to worry about :)

msujew (approved):
Thank you! I'll merge this as soon as the Eclipse foundation outage is fixed (one of the checks is a required check unfortunately).

JonasHelming:
@msujew Should be good to retrigger now!

msujew:
@parisa-mchp Do you mind force pushing to the PR? The GH actions seem to be stuck.

Issue #15444, opened 2025-04-11
There is no way to switch the locale language back to the default one.


Zebsterpasha:
### Bug Description:
When trying to switch the localization language (Configure Display Language) back to the default language, some of the inscriptions remain in the previous language.

### Steps to Reproduce:
1. Launch IDE
2. Press F1
3. Configure Display Language to something other than default
4. IDE will reboot
5. Make sure `getNlsConfig()` returns the correct language ( https://github.com/eclipse-theia/theia/blob/b7670041a232df0868e51bd8622bec758ea98afe/packages/plugin-ext/src/hosted/node/hosted-plugin-process.ts#L181 )
6. Switch the language to the default language (e.g. English)
7. `getNlsConfig()` will return the previous language (breakpoint from point 5)

Probably error is here: https://github.com/eclipse-theia/theia/blob/b7670041a232df0868e51bd8622bec758ea98afe/packages/core/src/browser/preload/i18n-preload-contribution.ts#L42
This condition is OK for the first run, but it is not suitable for switching language back to default.
It would be better to return the condition as it was:  `if (nls.locale)`

### Additional Information

- Operating System: any
- Theia Version: 1.59

Zebsterpasha:
@msujew My suggested fix works. If this option is acceptable, I will make a PR. Otherwise, we can suggest contributer @tortmayr to improve his solution.

msujew:
I think it makes sense. The additional check was there to prevent the additional call to the backend, but it's actually required if the set the language back to english.

PR #15445, opened 2025-04-11
Fixes switching the localization language back to the default #15444 

Zebsterpasha:
#### What it does
Fixes switching the localization language back to the default #15444 

#### How to test
1) Launch IDE
2) Press F1
3) Configure Display Language to something other than default
4) IDE will reboot
5) Make sure getNlsConfig() returns the correct language 
https://github.com/eclipse-theia/theia/blob/b7670041a232df0868e51bd8622bec758ea98afe/packages/plugin-ext/src/hosted/node/hosted-plugin-process.ts#L181
6) Switch the language to the default language (e.g. English)
7) getNlsConfig() will return the default language (breakpoint from point 5). Before the fix, it returned the previous language instead of the default language.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/preload/i18n-preload-contribution.ts
+++ b/packages/core/src/browser/preload/i18n-preload-contribution.ts
@@ -39,7 +39,7 @@ export class I18nPreloadContribution implements PreloadContribution {
             });
         }
         let locale = nls.locale ?? nls.defaultLocale;
-        if (nls.locale && nls.locale !== nls.defaultLocale) {
+        if (nls.locale) {
             const localization = await this.localizationServer.loadLocalization(locale);
             if (localization.languagePack) {
                 nls.localization = localization;


martin-fleck-at (commented):
Hi @Zebsterpasha,

Thank you for your PR! It seems that this change was introduced as part of https://github.com/eclipse-theia/theia/pull/13134 as an optimization:

> Update I18nPreloadContribution to ensure that we only load the localizations from the back end if the current locale is not equal to the default locale. This fixes an issue where we unnecessarily queried the localization server in the preload step if the locale is explicitly set to "en". (This happens for instance if you change the display language and then switch back to english).

However, it seems that due to that we are missing the `setCurrentLanguage` call that the server is making in the backend when we load the localizations:
https://github.com/eclipse-theia/theia/blob/ccc9f3e205e261e2d53ba0e1549ccb0568c6f4db/packages/core/src/node/i18n/localization-server.ts#L49

@tortmayr Do you remember if there was any issue with loading localizations for the default language in particular? It seems to me that loading them should be quite fast anyways and I don't see why we would never want to load for the default locale as other localizations may have been added. Unless there is a good reason for this, I'm inclined to accept this PR (as soon as the CI is happy as well).

tortmayr:
> @tortmayr Do you remember if there was any issue with loading localizations for the default language in particular? It seems to me that loading them should be quite fast anyways and I don't see why we would never want to load for the default locale as other localizations may have been added. Unless there is a good reason for this, I'm inclined to accept this PR (as soon as the CI is happy as well).

I don't recall exactly. I think it gave a slight performance boost when using the default locale but nothing to significant.
Anyways, since there are issues with it we should accept this PR.
If we encounter significant performance degradation (witch I don't expect) we can handle this in a follow-up.

Issue #14450, opened 2024-11-14
Intermittent Playwright test failures in `theia-notebook-editor.test`


planger:
I'm monitoring the stability of the Playwright tests and observed that there are a number of intermittent, recurring test failures at the `theia-notebook-editor.test.ts`.

```
Error:   1) ../../src/tests/theia-notebook-editor.test.ts:98:9 › Theia Notebook Editor interaction › should split cell 
    Error: elementHandle.waitForElementState: Element is not attached to the DOM
    Call log:
      -   waiting for element to be visible


      111 |         */
      112 |         const line = await cell.editor.lineByLineNumber(1);
    > 113 |         await line?.waitForElementState('visible');
          |                     ^
      114 |         await line?.click();
      115 |         await line?.press('ArrowRight');
      116 |

        at /home/runner/work/theia/theia/examples/playwright/src/tests/theia-notebook-editor.test.ts:113:21
```

* https://github.com/eclipse-theia/theia/actions/runs/11793428938/job/32848940686
* https://github.com/eclipse-theia/theia/actions/runs/11720136001/job/32644823097
* https://github.com/eclipse-theia/theia/actions/runs/11589018656/job/32263710648

It is not alarmingly frequent, but it might be good to look into it anyway to avoid false-positives.

planger:
@dhuebner Would you be able to see if we can avoid these failures? Thanks a lot in advance!

dhuebner:
@planger 
Sure, I will take a look. Thank you for reporting!

dhuebner:
@planger 
I was not able to reproduce it locally :( 
I've prepared a PR #14457 that removes the `visible` check and adds some logging to see the line Element state. If the test fails again, I hope I can  narrow down the cause. In the test reports it is obvious that all cell-editor lines are visible when the test was failing. See attached report. 

[playwright-test-results-3.zip](https://github.com/user-attachments/files/17776966/playwright-test-results-3.zip)

planger:
It looks like we have a few additional flaky Notebook editor tests:
```
    src/tests/theia-notebook-editor.test.ts:87:9 › Theia Notebook Editor interaction › should execute all cells 
    src/tests/theia-notebook-editor.test.ts:70:9 › Theia Notebook Editor interaction › should add a new code cell 
    src/tests/theia-notebook-editor.test.ts:231:9 › Theia Notebook Cell interaction › Check x/c/v works 
```

See for instance:
[...]

PR #15446, opened 2025-04-11
Improve playwright tests reliability

dhuebner:
#### What it does

Currently playwright tests sometimes failing when accessing the MonacoEditor lines. In this PR the MonacoEditor page object was refactored to use Locator API instead of selectors. 

See #14450

#### How to test

Tests should not fail sporadically. 

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/playwright/src/tests/theia-notebook-editor.test.ts
+++ b/examples/playwright/src/tests/theia-notebook-editor.test.ts
@@ -112,7 +112,7 @@ test.describe('Theia Notebook Editor interaction', () => {
         print("Line-1")
         <|>print("Line-2")
         */
-        const line = await cell.editor.line(1);
+        const line = await cell.editor.monacoEditor.line(1);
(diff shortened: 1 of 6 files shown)


planger (approved):
Thank you very much! Looks good to me. Let's observe if that improves the stability. :crossed_fingers:

planger:
@dhuebner I'm afraid, this change didn't really seem to help on a first glance, what do you think?
https://github.com/eclipse-theia/theia/actions/runs/15036406059
Thank you in advance!

dhuebner:
@planger 
This seems to be something different/new.
One case is Python kernel selection, could be that a new version was installed. Here we need a more lax selector that picks just one if nothing matches. 

  
test-failed-2

And the other one is "auto-save" that didn't worked, so the editor can not be closed after successful test. I need to investigate this further:
test-failed-2

I will re-open the initial issue #14450 with new information added.

PR #15451, opened 2025-04-12
Correctly return tool input when creating the message

### LINKED ISSUE #15455: [Theia AI] Conversation shows Error "400" after tool call with empty arguments
The following error appears in the chat:

400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.2.content.0.tool_use.input: Field required"}}

Likely after at least one tool call was made with no argument.
A reliable way to reproduce this is:
- Open an empty workspace
- Ask Architect "Look at my workspace"
- type "Hi" as a next message

@eneufeld I would consider this a blocker warranting a hotfix release or do you know any work around?

See: https://github.com/eclipse-theia/theia/pull/15451
for a fix

eneufeld:
fixed #15455

#### What it does

For tool calls which had no arguments, the tool input was returned as empty string which does not work for some models.

#### How to test

Use Sonnet 3.7 as your llm. Let it search the workspace for the readme.md:
`@Coder find the location of the readme.md`
After it finishes send any follow up message. eg `Hi`. 
Without the fix this breaks.
#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-model.ts
+++ b/packages/ai-chat/src/common/chat-model.ts
@@ -1043,7 +1043,7 @@ export class ToolCallChatResponseContentImpl implements ToolCallChatResponseCont
             actor: 'ai',
             type: 'tool_use',
             id: this.id ?? '',
-            input: this.arguments && this.arguments.length !== 0 ? JSON.parse(this.arguments) : undefined,
+            input: this.arguments && this.arguments.length !== 0 ? JSON.parse(this.arguments) : {},
             name: this.name ?? ''
         }, {
             actor: 'user',
--- a/packages/ai-chat/src/common/chat-model.ts
+++ b/packages/ai-chat/src/common/chat-model.ts
@@ -1043,7 +1043,7 @@ export class ToolCallChatResponseContentImpl implements ToolCallChatResponseCont
             actor: 'ai',
             type: 'tool_use',
             id: this.id ?? '',
-            input: (this.arguments && JSON.parse(this.arguments)) ?? undefined,
+            input: this.arguments && this.arguments.length !== 0 ? JSON.parse(this.arguments) : undefined,
             name: this.name ?? ''
         }, {
             actor: 'user',


JonasHelming:
@eneufeld Thank you for the PR. I can unfortunately, still reproduce this, e.g. with an empty workspace:

@Architect Look at the workspace

and then

"hi"

eneufeld:
More stable message to use:
@Universal  ~getWorkspaceDirectoryStructure  look at the workspace

So it is important that a function that has no parameters is called

JonasHelming:
Great, issue is gone for me! @planger Could you give a second approval to the code change? I would suggest to make a hotfix release with this change only.

packages/ai-chat/src/common/chat-model.ts
```diff
@@ -1043,7 +1043,7 @@ export class ToolCallChatResponseContentImpl implements ToolCallChatResponseCont
             actor: 'ai',
             type: 'tool_use',
             id: this.id ?? '',
-            input: (this.arguments && JSON.parse(this.arguments)) ?? undefined,
+            input: this.arguments && this.arguments.length !== 0 ? JSON.parse(this.arguments) : {},
```
planger, line 1046: I didn't really test this but I don't like this pattern, because it is easily overlooked that if `this.arguments` is falsy `input` will be falsy, even though the ternary operator suggests that it should be a `{}`, if `this.arguments` has a length of !== 0.

Did you mean?
`(this.arguments && this.arguments.length !== 0) ? JSON.parse(this.arguments) : {},`

This would say: If we have arguments then parse them, otherwise return {}
This would make sense to me.

But currently it says:
if we have arguments and a length !== 0 then parse them, otherwise:
  if arguments is falsy return false
     else
        return {}

eneufeld: I don't think the brackets make a difference here, but I will add them.

planger: Ok great, I think for clarity this makes a lot of sense. Thank you!

planger (approved):

PR #15452, opened 2025-04-12
feat(hover-service) make hover-service use platform showPopover()

Alexander-Taran:
#### What it does

Prevents any z-index problems for hovers/

It makes HoverService use web platform `[popover='hint']`.
Which it turn puts them on "top-layer"
Popover API is baseline available now and some of us use it.
https://developer.mozilla.org/en-US/docs/Web/API/Popover_API

HoverService popovers had no chance of being on top of "true" popovers.
This PR addresses just that.
Screenshot of browser-only example with this change in place:

image

Bonus: It also allows to close such popovers with "Escape"

#### How to test

HoverService popovers should be visible and on top of everything else.

#### Follow-ups

Someday when options for showPopover() will be baseline available it would be possible to position them declaratively to target. So code that calculates position could be thrown away.
It is available in chromium based browsers already.
image

#### Attribution

on behalf of web standards and performance 

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/style/hover-service.css
+++ b/packages/core/src/browser/style/hover-service.css
@@ -31,7 +31,7 @@
   max-width: var(--theia-hover-max-width);
 }
 
-/* user agent style-sheet has overflow:hidden, so pointer is not visible but scroll is */
+/* overwrite potentially different default user agent styles */
 .theia-hover[popover] {
   inset: unset;
   overflow: visible;
--- a/packages/core/src/browser/style/hover-service.css
+++ b/packages/core/src/browser/style/hover-service.css
@@ -29,7 +29,12 @@
   border: 1px solid var(--theia-editorHoverWidget-border);
   padding: var(--theia-ui-padding);
   max-width: var(--theia-hover-max-width);
-  z-index: 1000;
+}
+
+/* user agent style-sheet has overflow:hidden, so pointer is not visible but scroll is */
+.theia-hover[popover] {
+  inset: unset;
+  overflow: visible;
 }
 
 .theia-hover .hover-row:not(:first-child):not(:empty) {
--- a/packages/core/src/browser/hover-service.ts
+++ b/packages/core/src/browser/hover-service.ts
@@ -92,6 +92,7 @@ export class HoverService {
             this._hoverHost = document.createElement('div');
             this._hoverHost.classList.add(HoverService.hostClassName);
             this._hoverHost.style.position = 'absolute';
+            this._hoverHost.setAttribute('popover', 'hint');
         }
         return this._hoverHost;
     }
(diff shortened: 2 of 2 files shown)


packages/core/src/browser/style/hover-service.css
```diff
@@ -29,7 +29,12 @@
   border: 1px solid var(--theia-editorHoverWidget-border);
   padding: var(--theia-ui-padding);
   max-width: var(--theia-hover-max-width);
-  z-index: 1000;
+}
+
+/* user agent style-sheet has overflow:hidden, so pointer is not visible but scroll is */
```
martin-fleck-at, line 34: nit: For me on Chrome, the user agent stylesheet has `overflow:auto` and not `overflow:hidden`, maybe we should change the comment to simply say we want to overwrite the default user agent stylesheet.

packages/core/src/browser/style/hover-service.css
```diff
@@ -29,7 +29,12 @@
   border: 1px solid var(--theia-editorHoverWidget-border);
   padding: var(--theia-ui-padding);
   max-width: var(--theia-hover-max-width);
-  z-index: 1000;
```
martin-fleck-at, line 32: Since not all browser fully support the new API, does it make sense to still keep this or what is the argument for removing it. (Though if we keep it, extracting it into it's own variable similar to `--theia-hover-max-width` would be nice)

martin-fleck-at (changes requested):
This PR looks good to me, just want to clarify some things. Thank you @Alexander-Taran!

packages/core/src/browser/style/hover-service.css
martin-fleck-at: I didn't encounter any issues myself but officially, the `hint` value that we are using does not seem to be fully supported by all browsers:
https://developer.mozilla.org/en-US/docs/Web/API/Popover_API#api.htmlelement.popover. So I thought keeping that as a fallback might be useful.

Alexander-Taran: Value "hint" is kind of forward proofing. Unknown value is treated as auto AFAIK.
And in Theia code we still manage hiding previous hovers ourselves.
So technically there is no difference. They all still go to top layer and z-index is not needed.

martin-fleck-at: Since I also didn't encounter any issues, I'm fine with removing it. Thank you!

packages/core/src/browser/style/hover-service.css
martin-fleck-at:
```suggestion
/* ensure overflow is visible as user agent styles might have different defaults */
```

martin-fleck-at (approved):
Everything looks good to me now, thank you very much @Alexander-Taran!

PR #15463, opened 2025-04-14
Resolve unresolved variables to empty string

### LINKED ISSUE #15453: [Theia AI] Handling of unresolved variables
Open Question: If project-info or other variables such as context summary in agent prompts do not exist, the variable remains unresolved like this "{{prompt:project-info}}"
We might consider to change this behavior:

We generally remove unresolved variables (and functions) => might be an issue for debugging
We remove prompt fragment only that cannot be resolved (i.e. return "")
We introduce a syntax like {{prompt:project-info}}? to make things that are optional

@planger WDYT?

JonasHelming:
fixed #15453

#### What it does

Return an empty String instead of the variable name for the following cases:
- Unresolved prompt fragments
- Empty changesets

#### How to test

Ask coder something and at the end of the system prompt that neither {{changeSetSummary}} nor {{prompt:project-info}} appears.
Generate a change set and add a project-info.prompttemplate to a prompt directory. Observe that the two variables are resolved then.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/common/prompt-variable-contribution.ts
+++ b/packages/ai-core/src/common/prompt-variable-contribution.ts
@@ -81,7 +81,12 @@ export class PromptVariableContribution implements AIVariableContribution, AIVar
                 }
             }
         }
-        return undefined;
+        console.debug(`Could not resolve prompt variable '${request.variable.name}' with arg '${request.arg}'. Returning empty string.`);
+        return {
+            variable: request.variable,
+            value: '',
+            allResolvedDependencies: []
+        };
     }
 
     protected async triggerArgumentPicker(): Promise<string | undefined> {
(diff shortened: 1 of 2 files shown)


packages/ai-core/src/common/prompt-variable-contribution.ts
```diff
@@ -81,7 +81,12 @@ export class PromptVariableContribution implements AIVariableContribution, AIVar
                 }
             }
         }
-        return undefined;
+        console.debug(`Could not resolve prompt variable '${request.variable.name}' with arg '${request.arg}'. Returning empty string.`);
```
planger, line 84: I think it'd probably be better to use `ILogger` instead of plain console.

```ts
@inject(ILogger)
protected logger: ILogger;

...

this.logger.debug(`Could not resolve prompt variable '${request.variable.name}' with arg '${request.arg}'. Returning empty string.`);
```

planger (approved):
Looks good and works fine! Just one comment regarding `ILogger` vs console. Thank you!

Issue #15395, opened 2025-04-04
[Theia AI] Default Model Management


JonasHelming:
# Context:
There is a proposal to change the default LLM in the Theia IDE to Sonnet-3-7. While conceptually beneficial, such a change could break setups for users who have never altered their model configuration and have not provided an Anthropic API key.

# Proposed Short-Term Solution
Implement fallback logic directly in the Language Model Service, based on the presence of API keys (from user preferences or environment variables). The model selection would follow this priority:

Anthropic (Sonnet-3.7), OpenAI 4o, Google Gemini pro 2.5

If none are available → fall back to a neutral option (possibly still Sonnet or a bundled fallback).

# 🧩 Proposed Long-Term Architecture
Introduce a structured “Default Model” abstraction, which does not currently exist. Key aspects of this design:

Define a central mechanism for mapping logical default types/aliases (e.g., "default chat") to actual LLMs.

Introduce standard default model types for agent requirements:

"Default" for general use.

"Default Chat" for conversational agents.

"Default Coding"

"Default Code Completion" for completion tasks.

Migrate agents without specific model requirements to use "Default" or the corresponding type.

Determine the concrete default model based on available API keys using the above logic.

Allow users to configure all default model preferences in one centralized place, eliminating the need to adjust each agent individually. Maybe allow users to define custom aliases.

lucas-koehler:
Small follow up improvements to do after #15782 :

- Render Agents using an alias in alias view (@sgraband has a WIP implementation)
- Show resolved model for an alias in alias config view and in agent config view
- Display in chain the ones that are ready and the one that is chosen (only if applicable)
- Better display that a model is not ready just disabled for now
- Configure aliases as default models for agents
- Allow to add aliases
- Define fallbackchains for aliases

sgraband:
This is closed by #15782. Follow up opened here: https://github.com/eclipse-theia/theia/issues/16089

PR #15465, opened 2025-04-15
add gpt-4.1 models to the list of default models

eneufeld:
#### What it does

Add the new GPT 4.1 models to the list of default models to choose from.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-openai/src/browser/openai-preferences.ts
+++ b/packages/ai-openai/src/browser/openai-preferences.ts
@@ -36,7 +36,7 @@ on the machine running Theia. Use the environment variable `OPENAI_API_KEY` to s
             type: 'array',
             description: nls.localize('theia/ai/openai/models/description', 'Official OpenAI models to use'),
             title: AI_CORE_PREFERENCES_TITLE,
-            default: ['gpt-4o', 'gpt-4o-2024-11-20', 'gpt-4o-2024-08-06', 'gpt-4o-mini', 'o1', 'o1-mini', 'o3-mini', 'gpt-4.5-preview'],
+            default: ['gpt-4.1', 'gpt-4.1-nano', 'gpt-4.1-mini', 'gpt-4o', 'gpt-4o-2024-11-20', 'gpt-4o-2024-08-06', 'gpt-4o-mini', 'o1', 'o1-mini', 'o3-mini', 'gpt-4.5-preview'],
             items: {
                 type: 'string'
             }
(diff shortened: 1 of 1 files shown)


packages/ai-openai/src/browser/openai-preferences.ts
```diff
@@ -36,7 +36,19 @@ on the machine running Theia. Use the environment variable `OPENAI_API_KEY` to s
             type: 'array',
             description: nls.localize('theia/ai/openai/models/description', 'Official OpenAI models to use'),
             title: AI_CORE_PREFERENCES_TITLE,
-            default: ['gpt-4o', 'gpt-4o-2024-11-20', 'gpt-4o-2024-08-06', 'gpt-4o-mini', 'o1', 'o1-mini', 'o3-mini', 'gpt-4.5-preview'],
+            default: [
```
JonasHelming, line 39: I would not yet remove o3-mini before we tested o4-mini in practise.

packages/ai-openai/src/browser/openai-preferences.ts
```diff
@@ -36,7 +36,19 @@ on the machine running Theia. Use the environment variable `OPENAI_API_KEY` to s
...
+                'gpt-4.1',
```
JonasHelming, line 40: I believe, if we put this on place one, it will be the global default (instead of gpt-4o). I think we should maybe not do this before having tested it a bit in practice. And we should introduce: https://github.com/eclipse-theia/theia/issues/15395
soon to make the default management easier.

JonasHelming (changes requested):
Thanks, I have two minor comments!

packages/ai-openai/src/browser/openai-preferences.ts
eneufeld: I'll move it to second place and yes the management ticket is important

packages/ai-openai/src/browser/openai-preferences.ts
```diff
@@ -36,7 +36,20 @@ on the machine running Theia. Use the environment variable `OPENAI_API_KEY` to s
...
+                'gpt-4.1-nano',
+                'gpt-4.1-mini',
+                'gpt-4o-2024-11-20',
+                'gpt-4o-2024-08-06',
+                'gpt-4o-mini',
+                'o3',
+                'o3-mini',
+                'o4-mini',
+                'o4-mini-high',
+                'gpt-4.5-preview'
```
JonasHelming, line 51:
```suggestion
```

JonasHelming (approved):
LGTM!

PR #15478, opened 2025-04-16
Improve handling of logging with proxies

colin-grant-work:
#### What it does

`packr` will attempt to use `toJSON` to serialize an object if sees that `toJSON` is defined. That means that if an object is logged that includes a reference to an RPC proxy, `packr` will attempt to call `toJSON`, and a type error will be logged if the remote target does not implement `toJSON` (even if it does, `packr` isn't expecting a `Promise`!). This PR returns `undefined` when `toJSON` is requested from a proxy to prevent `packr` from attempting to use that method to serialize the object.

#### How to test

1. Attempt to log an object with a reference to a proxy (e.g. the `FileService`)
2. Observe that the log appears successfully on the frontend.
3. Observe that no error is thrown indicating that `this.target[method] is not a function`
> The log won't appear on the backend because the object couldn't be serialized, but that is the usual behavior when logging anything that doesn't serialize cleanly.

#### Breaking changes

> [!NOTE]
> Since `toJSON` is JS API, proxies generally shouldn't implement it, and I don't see any cases where it is included in an interface in the repo. But an adopter could have defined that method in something they proxy, and it would now fail.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/messaging/proxy-factory.ts
+++ b/packages/core/src/common/messaging/proxy-factory.ts
@@ -237,6 +237,10 @@ export class RpcProxyFactory<T extends object> implements ProxyHandler<T> {
             // Prevent inversify from identifying this proxy as a promise object.
             return undefined;
         }
+        if (p === 'toJSON') {
+            // Prevent packr from attempting to serialize proxies with `toJSON`.
+            return undefined;
+        }
         const isNotify = this.isNotification(p);
         return (...args: any[]) => {
             const method = p.toString();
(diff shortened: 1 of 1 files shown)


packages/core/src/common/messaging/proxy-factory.ts
```diff
@@ -274,7 +278,7 @@ export class RpcProxyFactory<T extends object> implements ProxyHandler<T> {
         return p.toString().startsWith('notify') || p.toString().startsWith('on');
     }
 
-    protected serializeError(e: any): any {
+    protected serializeError(e: any, method?: string): any {
```
martin-fleck-at, line 281: I do like that improvement if there are actual cases where we get that parameter but I'm not seeing it. Do you have a case for that?

martin-fleck-at (changes requested):
Changes look good to me and can almost be merged, thank you Colin! I just need some explanation for that additional parameter in the error serialization.

martin-fleck-at (approved):
LGTM!

PR #15479, opened 2025-04-17
Introduce editable chat requests

### LINKED ISSUE #15295: [Theia AI] Edit user messages / branch off conversations
Allow the user to edit the request message in the chat UI.

## UI

Show an edit icon next to the request and allow the user to provide a new input.
Afterward, the user should still be able to switch between versions.

## Code

Internally, we want to build a tree for the session model, where each change creates a new subtree. This will allow branching off the conversation.

haydar-metin:
#### What it does

Closes #15295 
@planger 

- Edit icon for edit
image

- UI after editing (same input as the chat input below)
image

- Supports same functionality
image

- Prev/Next possible
image

#### How to test

See images above.

Please also test the following
- Enabling the editing functionality, then scrolling around / switching to other chats

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-tree-view/chat-view-tree-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-tree-view/chat-view-tree-widget.tsx
@@ -418,23 +418,23 @@ export class ChatViewTreeWidget extends TreeWidget {
             variableService={this.variableService}
             openerService={this.openerService}
             provideChatInputWidget={() => {
-                const request = node.request;
-                if (EditableChatRequestModel.isEditing(request)) {
-                    let widget = this.chatInputs.get(node.id);
+                const editableNode = node;
+                if (isEditableRequestNode(editableNode)) {
+                    let widget = this.chatInputs.get(editableNode.id);
                     if (!widget) {
                         widget = this.inputWidgetFactory({
-                            request: node,
-                            initialValue: node.request.message.request.text,
+                            node: editableNode,
+                            initialValue: editableNode.request.message.request.text,
                             onQuery: async query => {
-                                request.submitEdit({ text: query });
+                                editableNode.request.submitEdit({ text: query });
                             }
                         });
 
-                        this.chatInputs.set(node.id, widget);
+                        this.chatInputs.set(editableNode.id, widget);
 
                         widget.disposed.connect(() => {
-                            this.chatInputs.delete(node.id);
-                            request.cancelEdit();
+                            this.chatInputs.delete(editableNode.id);
+                            editableNode.request.cancelEdit();
                         });
                     }
 
@@ -529,8 +529,7 @@ const WidgetContainer: React.FC<WidgetContainerProps> = ({ widget }) => {
     React.useEffect(() =>
         () => {
             setTimeout(() => {
-                // Delay clean up to allow react to finish its rendering
-                // cycle
+                // Delay clean up to allow react to finish its rendering cycle
                 widget.clearFlag(Widget.Flag.IsAttached);
                 widget.dispose();
             });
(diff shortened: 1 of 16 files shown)


planger:
@haydar-metin Looks awesome on a first glance :yay:! Can you please resolve the conflict? I'll have a detailed look at this PR in the afternoon.

packages/ai-chat-ui/src/browser/chat-tree-view/chat-view-tree-widget.tsx
```diff
@@ -397,50 +502,127 @@ export class ChatViewTreeWidget extends TreeWidget {
...
+            Widget.attach(widget, containerRef.current);
+        }
+    }, [containerRef.current]);
+
+    // Clean up
+    React.useEffect(() =>
+        () => {
+            setTimeout(() => {
+                // Delay clean up to allow react to finish its rendering
+                // cycle
```
planger, line 533:
```suggestion
                // Delay clean up to allow react to finish its rendering cycle
```

planger (changes requested):
Exceptional work! This is a really great feature and well implemented and already works almost perfectly. Thank you!

Below are a few things that I noticed that we should fix before merging.

### Context Variables

Probably the most complex one to solve, but also imho the most important one is that we only maintain one state of the context variables across all inputs (main input and editable request node inputs). This has a few side effects:

* When I modify a request, it'll always show the context that I currently have attached to the main input, but not the one that was originally attached to the request I'm editing. This may also leads to wrong context if I re-submit a request.
* I can't seem to change the context variables anymore for already submitted requests, even though the UI indicates that I can and did. But the request actually does't reflect this.

https://github.com/user-attachments/assets/f455d1a4-18a5-4abc-9817-d9a8351294a8

### Styling

* I'd suggest to remove left and right margin from `.theia-ChatInput-Editor-Box`, if it is in a `.theia-RequestNode`. This will give the input editor more space and better fits the width of the main input editor.
* There is an inadvertent line-break between agent label / variable label and the subsequent text (see above). I believe this happens because we introduced an intermediate div and this breaks the `.theia-RequestNode > p div` in `packages/ai-chat-ui/src/browser/style/index.css:138`.

### UX

* I suggest to cancel the editing if the user pushes `Esc`
* I think adding a context menu item for editing might be great too.

haydar-metin:
@planger Thanks for the feedback! The context menu handling and the other topics should be now fixed. If there exists any other functionality similar to the context, that I didn't see, then just tell me please!

planger (changes requested):
Thank you very much! This is almost perfect. I missed one more thing: The change set is similar to the context and thus requires special handling too.

Currently it shows up in the edit-message input editor.

image

Tracking the state of the change set per editable request --- similar to how you now do it for context elements --- might be a bit overly complex, what do you think? Also I think it is more comprehensible to users, if there is just one state of the change set per session (and not per editable message). So I tend to suggest hiding it from edit-message input editor and reset the change set entirely if a message is edited, at least for now. We may want to consider storing the state for each branch at some point. That is:
* We reset the change set when the message is edited and submitted
* But we store the change set that has been reset for this new branch so that when going back, we see the change set that was active before.

What do you think @JonasHelming?

Otherwise, this is imho working very very well! Maybe it'd be good to get another pair of eyes on the chat model changes by @eneufeld or @sdirix. To me the code looks good though!

Thank you!

JonasHelming:
@planger I agree to the proposal above where I was tagged :-)

haydar-metin:
@planger the last commit hides and resets the change set after adding a new edited request.

planger (approved):
Thank you for this powerful feature @haydar-metin!
I've tracked a potential future improvement (storing and restoring the change set state for each branch) in https://github.com/eclipse-theia/theia/issues/15513

PR #15481, opened 2025-04-17
Clarify error for multiple occurences during replace

JonasHelming:
#### What it does

Clarifies the error message when multiple occurrences are found, but the flag "multiple" is not set

#### How to test

The following prompt should work with replace (typically it tries it without mutliple first)

@Coder Rename the function toOpenAIMessage in packages/ai-openai/src/node/openai-language-model.ts to convertToOpenAIMessage

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


colin-grant-work (approved):
Changes look reasonable and the behavior is improved.

colin-grant-work (changes requested):
Though the code is good, it looks like the message was referred to in some tests, so those will have to be adjusted.

colin-grant-work (approved):
Tests fixed; merge if you're happy with the changes.

Issue #15498, opened 2025-04-22
Debug decorations for current stack frame are lost after switching editor tabs


pisv:
### Bug Description:

Debug decorations indicating the current stack frame are lost after switching editor tabs.

(Although it has already been mentioned in #15497, this is a separate and important issue that merits its own tracking.)

### Steps to Reproduce:

1. Set a breakpoint. When the breakpoint is hit, the current stack frame is indicated by special decorations in the editor:

<img width="1243" alt="Image" src="https://github.com/user-attachments/assets/b55025cb-9231-4a0c-b8c1-4d725ee9b667" />

2. Switch to another editor. Switch back to the first. The editor decorations related to the current stack frame are now lost:

<img width="1243" alt="Image" src="https://github.com/user-attachments/assets/a39c0dd7-6505-4578-bdca-69c7cfb62e80" />

### Additional Information

This looks like another regression from https://github.com/eclipse-theia/theia/pull/14909. See also https://github.com/eclipse-theia/theia/issues/15426.

- Operating System: macOS 13.7.5
- Theia Version: the master branch

Issue #15497, opened 2025-04-22
Debug exception widget looks broken after switching editor tabs


pisv:
### Bug Description:

An open debug exception widget looks broken after switching editor tabs.

### Steps to Reproduce:

0. Write a simple node.js program that throws an exception.
1. Set an exception breakpoint. Debug exception widget appears when the exception is thrown:

<img width="1243" alt="Image" src="https://github.com/user-attachments/assets/5eb4947e-4f9f-4327-a47d-323df7a198de" />

2. Switch to another editor. Switch back to the first. Debug exception widget now looks broken. Also, many of the editor decorations are lost, including the decorations indicating the current stack frame in the editor.

<img width="1243" alt="Image" src="https://github.com/user-attachments/assets/d1303c6f-939f-4c15-bbb2-1b4b7fb7f82e" />

### Additional Information

This looks like another regression from https://github.com/eclipse-theia/theia/pull/14909. See also #15496 and #15426.

- Operating System: macOS 13.7.5
- Theia Version: the master branch

Issue #15496, opened 2025-04-22
Peek view for conditional breakpoint looks broken after switching editor tabs


pisv:
### Bug Description:

An open peek view for conditional breakpoint / logpoint looks completely broken after switching editor tabs.

### Steps to Reproduce:

1. Open a `.ts` file. Open a peek view for adding a conditional breakpoint:

<img width="1525" alt="Image" src="https://github.com/user-attachments/assets/a8810c82-b51c-4131-a832-20328368d76d" />

2. Open another editor. Switch back to the first. The open peek view looks completely broken:

<img width="1525" alt="Image" src="https://github.com/user-attachments/assets/22325459-437d-4dee-a1b9-96a4e15ca36e" />

### Additional Information

This looks like yet another regression from #14909. See also https://github.com/eclipse-theia/theia/issues/15426#issuecomment-2797706094.

- Operating System: macOS 13.7.5
- Theia Version: the master branch

PR #15500, opened 2025-04-22
Do not clear the editor model when the editor is hidden

### LINKED ISSUE #15426: Dirty-diff decorations lost when editor hidden
### Bug Description:

When an editor with dirty-diff decorations is hidden and shown again (e.g. when switching editor tabs), the dirty-diff decorations disappear.

### Steps to Reproduce:

1. Open a file with diff changes in a text editor. Verify that dirty-diff decorations are shown for the changes.

2. Open another editor. Switch back to the first. Note that there are no dirty-diff decorations now.

### Additional Information

The issue is present on the current master. It looks like a regression from #14909. I have checked that the issue is present in https://github.com/eclipse-theia/theia/commit/72ca8ed8559d6309e374d74f42f631e0f57e915c, while the parent commit does not have this issue.

pisv:
#### What it does

TL;DR This PR proposes to reconsider the change introduced by #14909.

So, #14909 has caused regressions all around Theia and probably even outside Theia (in Theia extensions, as opposed to VS Code extensions) because it significantly changed the way that hidden text editors had been behaving in Theia for years. And, unfortunately, in its nature it is also the worst kind of a breaking change possible, one that the compiler would never tell you about.

Setting the editor model to `null` when the editor is hidden detaches the model from the editor, which has the following effects (the list is not exhaustive):

- All decorations owned by the editor are removed from the model;

- All view zones, overlay widgets, etc. are removed;

- The `monaco-editor` node is removed from the DOM.

While the editor is hidden, calling `TextEditor.deltaDecorations` API has no effect now, since the model is detached from the editor. (I have not checked it yet, but it seems likely that other editor-related APIs in Theia might also behave differently in this state now.)

When the editor becomes visible again and the model is reattached to the editor, the `monaco-editor` node hierarchy is recreated from scratch and some editor components (e.g. overlay widgets) are recreated from the data stored in the model, while others such as decorations owned by the editor and view zones are not automatically recreated and would need to recreated manually now.

Of course, such a drastic change in well-established behavior has a vast potential for regressions, which has already materialised in practice: after two months and a couple of releases, their list is still growing.

This PR proposes to reconsider the change introduced by #14909. It reverts the effect of #14909 in a non-disruptive way, making as few changes as possible. It fixes #15426, #15496, #15497, #15498, and any other known or potential issues caused by #14909. The fix should be safe to apply, since any code that have already been adapted to the change in #14909 is expected to keep working, while any code that have not been adapted, and is currently broken, will function normally again. Of course, it does mean that #14880, which was fixed by #14909, will need to be fixed in some other way.

#### How to test

Verify that all known regressions from #14909 are now fixed, including those that have already been fixed by adapting the code to the change in #14909. To the best of my knowledge, the list of currently known regressions is:

- #14924
- #15038
- #15057
- #15178
- #15426
- #15496
- #15497
- #15498

There are also #15201 and #15234, which may be related, but have not been confirmed as regressions.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-diff-editor.ts
+++ b/packages/monaco/src/browser/monaco-diff-editor.ts
@@ -43,9 +43,6 @@ export namespace MonacoDiffEditor {
 export class MonacoDiffEditor extends MonacoEditor {
     protected _diffEditor: IStandaloneDiffEditor;
     protected _diffNavigator: DiffNavigator;
-    protected savedDiffState: monaco.editor.IDiffEditorViewState | null;
-    protected originalTextModel: monaco.editor.ITextModel;
-    protected modifiedTextModel: monaco.editor.ITextModel;
 
     constructor(
         uri: URI,
@@ -59,8 +56,6 @@ export class MonacoDiffEditor extends MonacoEditor {
         parentEditor?: MonacoEditor
     ) {
         super(uri, modifiedModel, node, services, options, override, parentEditor);
-        this.originalTextModel = originalModel.textEditorModel;
-        this.modifiedTextModel = modifiedModel.textEditorModel;
         this.documents.add(originalModel);
         const original = originalModel.textEditorModel;
         const modified = modifiedModel.textEditorModel;
@@ -132,26 +127,6 @@ export class MonacoDiffEditor extends MonacoEditor {
 
     override shouldDisplayDirtyDiff(): boolean {
         return false;
-    }
-
-    override handleVisibilityChanged(nowVisible: boolean): void {
-        if (nowVisible) {
-            this.diffEditor.setModel({ original: this.originalTextModel, modified: this.modifiedTextModel });
-            this.diffEditor.restoreViewState(this.savedDiffState);
-            this.diffEditor.focus();
-        } else {
-            const originalModel = this.diffEditor.getOriginalEditor().getModel();
-            if (originalModel) {
-                this.originalTextModel = originalModel;
-            }
-            const modifiedModel = this.diffEditor.getModifiedEditor().getModel();
-            if (modifiedModel) {
-                this.modifiedTextModel = modifiedModel;
-            }
-            this.savedDiffState = this.diffEditor.saveViewState();
-            // eslint-disable-next-line no-null/no-null
-            this.diffEditor.setModel(null);
-        }
     }
 }
 
--- a/packages/monaco/src/browser/monaco-editor.ts
+++ b/packages/monaco/src/browser/monaco-editor.ts
@@ -126,8 +126,6 @@ export class MonacoEditor extends MonacoEditorServices implements TextEditor {
     readonly onDidResize = this.onResizeEmitter.event;
 
     readonly documents = new Set<MonacoEditorModel>();
-    protected model: monaco.editor.ITextModel | null;
-    savedViewState: monaco.editor.ICodeEditorViewState | null;
 
     protected constructor(
         readonly uri: URI,
@@ -259,19 +257,6 @@ export class MonacoEditor extends MonacoEditorServices implements TextEditor {
     }
 
     handleVisibilityChanged(nowVisible: boolean): void {
-        if (nowVisible) {
-            if (this.model) {
-                this.editor.setModel(this.model);
-                this.editor.restoreViewState(this.savedViewState);
-                this.editor.focus();
-            }
-        } else {
-            this.model = this.editor.getModel();
-            this.savedViewState = this.editor.saveViewState();
-
-            // eslint-disable-next-line no-null/no-null
-            this.editor.setModel(null); // workaround for https://github.com/eclipse-theia/theia/issues/14880
-        }
     }
 
     getVisibleRanges(): Range[] {


tsmaeder:
@pisv so how do we fix the original issue https://github.com/eclipse-theia/theia/issues/14880? It breaks basic code navigation in the editor. On a more general level: VS Code hides editors which are not shown, so we must assume that more issues like #14880 will arise in the future.

pisv:
Personally, I've never encountered #14880 in practice, and from its description, it seems that it occurs only under *some* conditions on *Linux*. So, although I don't know how to fix it, I'm far more concerned about the havoc that #14909 is causing. I'd bet that not every regression even in Theia itself has been discovered yet, not speaking of external Theia extensions.

Regarding VS Code, it has an entirely different editor lifecycle than Theia. An editor instance in VS Code is reused if possible when switching editor tabs. It is just given another editor input with another model; i.e. it is not hidden, it is reinitialized from scratch. As far as I can tell, there is no way to replicate this behavior in Theia without a nuclear breaking change.

Having said that, it is entirely your choice. I've created this PR mostly as a place for discussion among Theia committers (and for the sake of convenience, in case it is accepted).

Basically, there are two ways forward:

- Find and fix every regression from #14909 both in Theia and in Theia adopters' extensions.

- Revert #14909 and try to find another way to fix #14880.

sdirix:
Thank you @pisv for this suggestion and for identifying the current issues with editor hiding.

I agree with @tsmaeder that we should generally aim to match VS Code's behavior to avoid potential issues. However, this argument is less compelling if what we're implementing isn't actually what VS Code does.

#14880 is a major regression and was internally reported by many Theia IDE Dogfooders. It's very easy to experience during typical development. Whether it's worse than all the other issues combined? :shrug: 

I agree with your suggested approaches. If we **can** find a reliable simple fix for #14880 without hiding the editors then this would be the best way forward *for now*. We can then discuss at a later point whether we want to reintroduce editor hiding.

If we **can't** find a reliable simple fix for #14880 then I'm not sure whether trading issues is worth it. We would then either need to:
 - Not only revert the editor hiding but also the Monaco upgrade #14737 to fix all issues. Then, to not fall behind, we would need a branch with the Monaco upgrade + Editor hiding + all fixes for it and then merge again later. This is a lot of effort.
 - Only do editor hiding on Linux if #14880 is considered worse than the other issues. However this opens Pandora's Box with very different behaviors on different platforms. So I would rather avoid it
 - Find and fix all issues with editor hiding

pisv:
Speaking from the perspective of a Theia adopter (Theia as a platform, and not just the Theia IDE), I'd like add one point, which seems important to me. #14909 is a serious breaking change in well-established behavior. If the decision is made to keep the current behavior, and find and fix all regressions, those regressions may also be lurking in external Theia extensions (as opposed to VS Code extensions). As such, this breaking change needs to be communicated to Theia adopters with a clear migration strategy.

tsmaeder:
>  If something in Monaco really started to make assumptions about the specifics of the editor lifecycle in the VS Code workbench

The API of the `ICodeEditor` underlying our `MonacoEditor` implementation has had the option to set the text model (including `null` for a long time, we just chose to ignore this fact and never set the model on the underlying implementation. VS Code has been doing that for a long time, though.

pisv:
I meant something entirely different. I was speaking about the underlying cause of the issue #14880, not about the specific fix in #14909. From your comment that "we might be able to patch monaco so that it works when being hidden", and also from @sdirix comment that "we would need to revert Monaco upgrade to fix it", I thought that something in Monaco started to make assumptions about the specifics of the editor lifecycle in the VS Code workbench, and that caused the issue in Theia. Does it make sense?

tsmaeder:
@pisv I cannot precisely point you to a commit, but from debugging, the monaco editor does not correctly update it's line width cache when it is hidden: in particular, it updates the line width to zero, and never recalculates it, leading to ctrl-click not working.  The details are in https://github.com/eclipse-theia/theia/issues/14880

pisv:
I spent some of my vacation time analyzing the issue #14880.

Using the reproduction steps in https://github.com/eclipse-theia/theia/issues/14880#issuecomment-2647820453, I was able to reproduce the issue in Theia down to version `1.46.0`, which uses Monaco `1.72.3`.

I have not tested the earlier versions, but it seems safe to say that the issue was lurking in Theia for a very long time, if not from the very beginning. It is also safe to say that the issue is *not* caused by any of the recent Monaco uplifts.

I could reproduce the issue in both Safari and Firefox, but not in Chrome or Electron (macOS).

It appears that the issue can be reproduced on all platforms for which [`canUseFastRenderedViewLine`](https://github.com/microsoft/vscode/blob/91fbdddc47bc9c09064bf7acf133d22631cbf083/src/vs/editor/browser/viewParts/viewLines/viewLine.ts#L23-L45) in `viewLine.ts` returns `false`.

It also looks like the issue can be fixed by one liner change in Monaco. The line [`556`](https://github.com/microsoft/vscode/blob/91fbdddc47bc9c09064bf7acf133d22631cbf083/src/vs/editor/browser/view.ts#L556
) in `view.ts` that currently reads

```ts
if (!this.domNode.domNode.isConnected) {
```

needs to be replaced with the following line:

```ts
if (!this.domNode.domNode.isConnected || !this.domNode.domNode.offsetParent) {
```

I have verified that the fix seems to work fine in Theia `1.58.1` (i.e. after the latest Monaco uplift, but before merging #14909) by patching Monaco locally.

Hope that helps. Thank you for your attention.

xai:
I also stumbled upon a regression in a [VSCode extension](https://github.com/VSCodeVim/Vim), where, among other issues, moving in normal mode via the `hjkl` keys is broken after opening a new editor window in a split view.

Removing [this line](https://github.com/eclipse-theia/theia/blob/f2464378531f34781ea0a4fa8afc2b142bd67466/packages/monaco/src/browser/monaco-editor.ts#L273) resolves this specific regression, but, as expected, reintroduces the #14880 issue, which is far more critical to the overall user base.

Unfortunately, it appears that the issue opened by @tsmaeder on https://github.com/microsoft/vscode/issues/249966 was closed as 'wontfix'.

pisv:
FWIW, I've created a PR https://github.com/microsoft/vscode/pull/251583, based on my findings in https://github.com/eclipse-theia/theia/pull/15500#issuecomment-2931415910.

JonasHelming:
@pisv What is the state of this. Should we continue with this or close the PR?

pisv:
@JonasHelming I did not push it any further. So, basically, it has been stalled. It is quite unfortunate, but, given how much time has already passed since the breakage occurred, it might probably make sense to close this PR.

JonasHelming:
OK, I am closing this now, please return to it anytime!

PR #15502, opened 2025-04-23
Use same autosave heuristic when closing widget

### LINKED ISSUE #15501: Untitled editors always saved on close when autosave is active
### Bug Description:

When autosave is active, untitled editors are exempted from it. When an untitled editor is closed, however, the logic to determine whether to save the widget or show a prompt to save the widget jumps straight to saving because `autosave` is on:

https://github.com/eclipse-theia/theia/blob/505c885de592efcb0c86e2b8ff6b270d091d9553/packages/core/src/browser/saveable-service.ts#L205-L237

### Steps to Reproduce:

1. Set the `files.autoSave` preference to anything but `off`
2. Open an untitled editor.
3. Write some content in the untitled editor, and then attempt to close it.
4. A file picker dialog will open to pick the saving location.
5. Cancel the save.
6. The widget will still be open.
7. Keep trying as long as your patience lasts.

### Additional Information

- Operating System: MacOS X Sequoia
- Theia Version: 505c885de592efcb0c86e2b8ff6b270d091d9553

colin-grant-work:
#### What it does

Fixes #15501 by using the same heuristic for autosaveability when closing a widget as when deciding whether to apply autosaving.

#### How to test

1. Set your autosave preference to something other than `off`
2. Create an untitled editor and populate it.
3. Attempt to close it.
4. You should be offered the 'Change will be lost' dialog (not straight to the save location dialog).
5. Choose any of the options, and you should get the expected behavior.
> [!Note]
> If you choose to save, the newly saved version will be opened, rather than both being closed. Perhaps that should be otherwise?

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/saveable-service.ts
+++ b/packages/core/src/browser/saveable-service.ts
@@ -222,7 +222,8 @@ export class SaveableService implements FrontendApplicationContribution {
         if (!Saveable.isDirty(widget)) {
             return false;
         }
-        if (this.autoSave !== 'off') {
+        const saveable = Saveable.get(widget);
+        if (this.autoSave !== 'off' && (!saveable || this.shouldAutoSave(widget, saveable))) {
             return true;
         }
         const notLastWithDocument = !Saveable.closingWidgetWouldLoseSaveable(widget, Array.from(this.saveThrottles.keys()));
--- a/packages/core/src/browser/saveable-service.ts
+++ b/packages/core/src/browser/saveable-service.ts
@@ -223,6 +223,10 @@ export class SaveableService implements FrontendApplicationContribution {
             return false;
         }
         const saveable = Saveable.get(widget);
+        if (!saveable) {
+            console.warn('Saveable.get returned undefined on a known saveable widget. This is unexpected.');
+        }
+        // Enter branch if saveable absent since we cannot check autosaveability more definitely.
         if (this.autoSave !== 'off' && (!saveable || this.shouldAutoSave(widget, saveable))) {
             return true;
         }


packages/core/src/browser/saveable-service.ts
```diff
@@ -222,7 +222,8 @@ export class SaveableService implements FrontendApplicationContribution {
         if (!Saveable.isDirty(widget)) {
             return false;
         }
-        if (this.autoSave !== 'off') {
+        const saveable = Saveable.get(widget);
+        if (this.autoSave !== 'off' && (!saveable || this.shouldAutoSave(widget, saveable))) {
```
sdirix, line 226: I don't know the specifics of `Saveable`. Looking at the code, what's weird to me is that we return `true` in `shouldSaveWidget` if there is no saveable... Shouldn't it look like this?

```suggestion
        const saveable = Saveable.get(widget);
        if (this.autoSave !== 'off' && saveable && this.shouldAutoSave(widget, saveable)) {
```

Note that the use case is still fixed with this suggestion, but I may overlook some other consequences

sdirix (commented):
It works for me, however I have a comment.

packages/core/src/browser/saveable-service.ts
colin-grant-work: `saveable` should always be defined or we would never get here - we wouldn't even be considering saving the widget if we hadn't discovered a saveable earlier in the process, but we don't pass it into this particular method - so the two are logically equivalent, so I'm happy to go either way.

sdirix: So maybe like this
```
        const saveable = Saveable.get(widget);
        if (!saveable) {
          throw new Error('saveable does not exist in widget which is checked for saving')
          
          //or

          // should never happen
          return false;

          //or

          console.warn('saveable does not exist in widget which is checked for saving');
          return false;
        }
          
        if (this.autoSave !== 'off' && this.shouldAutoSave(widget, saveable)) {
```

or

```
        const saveable = Saveable.get(widget);          
        // saveable must always exist here
        if (this.autoSave !== 'off' && this.shouldAutoSave(widget, saveable!)) {
```

colin-grant-work: The trouble with each of your proposals is that they will throw errors if something unexpected happens, and I wanted to avoid that outcome, because that amounts to an API breakage that is only detectable at runtime, which is particularly frustrating for adopters. If we are genuinely concerned about the case where `Saveable.get` returns either nothing or something other than the expected `Saveable`, then we should break the API in a way that is tranparent in the type system and pass the saveable on which this method is dependent into it.

In my view, it's fairly unlikely that we will get anything other than the expected `Saveable` from the `Saveable.get` call, so my original solution, which leaves adopters in the position (perhaps) having to fix the bug for themselves if they've done something that leads to the saveable being `undefined`, seemed like a reasonable compromise between  fixing the bug for the normal case and leaving the detectable abnormal case of `undefined` `Saveable` unaffected, but if you'd prefer that we be sure that we're checking the correct `Saveable`, I'd be happy to implement it accordingly.

sdirix: I get that it's unnecessary to change behavior for adopters, at the same time it's weird that we explicitly check for an "impossible" state and then arbitrarily return `true` in `shouldSaveWidget` for something we don't even expect to be saveable.

To be pragmatic, let's go with the original suggestion of yours but add a comment? e.g.

```suggestion
        const saveable = Saveable.get(widget);
        // saveable should always exist here. We enter the if branch if it doesn't to keep backward compatibility
        if (this.autoSave !== 'off' && (!saveable || this.shouldAutoSave(widget, saveable))) {
```

sdirix (approved):
Thank you!

PR #15514, opened 2025-04-24
[vscode] Support TerminalState shell property

### LINKED ISSUE #15433: [vscode] TerminalState#shell property added
A new property has been added to TerminalState
```
 /**
 * The detected shell type of the {@link Terminal}. This will be `undefined` when there is
 * not a clear signal as to what the shell is, or the shell is not supported yet. This
 * value should change to the shell type of a sub-shell when launched (for example, running
 * `bash` inside `zsh`).
 *
 * Note that the possible values are currently defined as any of the following:
 * 'bash', 'cmd', 'csh', 'fish', 'gitbash', 'julia', 'ksh', 'node', 'nu', 'pwsh', 'python',
 * 'sh', 'wsl', 'zsh'.
 */
readonly shell: string | undefined;
```

rschnekenbu:
#### What it does

Adds the support of the TerminalState Shell property. This property reflects the current shell type (bash, powershell, gitbash, node, etc.) currently used by the terminal.
Note that current vscode implementation let the tool know more than current implementation in Theia. For example, in current implementation, typing 'node' in as bash terminal will throw a state change with node rather than bash. This is currently not the case, due to more limited support of terminal in Theia (see for example https://github.com/microsoft/vscode/blob/6916ae0d31fcfb26c9d1ace42c933987ef30f3bc/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts#L1432 where the terminal process manager can update on the current cwd, child processes, etc.).

This would require a larger update of current terminal implementation, which is not planned yet.

fixes #15433

#### How to test

install the following extension, that will allow to create a new terminal with a terminal state listener. this listener will display information messages on state changed:
- zip vsix: [terminal-state-0.0.1.zip](https://github.com/user-attachments/files/19895417/terminal-state-0.0.1.zip)
- src: [terminal-state-0.0.1-src.zip](https://github.com/user-attachments/files/19895423/terminal-state-0.0.1-src.zip)

Once installed, on browser or electron, you can create a new terminal with the listener using the command `Test Terminal State API` 
This command will create a new terminal and display 2 messages (I kept the vscode behavior here). The first step is on terminal creation with undefined shell type and interactable being false. Then a second notification comes with the shell being defined (bash or whatever is the command used there). 
Then if you type in the terminal, the interactable state should changed to true (meaning that the user had some key strokes in the terminal).
  

#### Follow-ups

Implement the dynamic shell type change once terminal implementation lets the tool determine the current sub shell if any.

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers


tsmaeder:
@rschnekenbu when I invoke the command, I always end up with shell being "undefined" on Windows/Electron

image

rschnekenbu:
@tsmaeder do you know which one is supposed  to be created in your case?

tsmaeder:
> which one is supposed to be created in your case

Looks like `cmd.exe`

tsmaeder (approved):

PR #15519, opened 2025-04-26
Fix: Show variables and functions only from selected prompt variants

### LINKED ISSUE #15320: Do not include unused prompt variants in configuration overview
### Bug Description:

Agents can specify multiple variants of their prompts. They can be selected in the Configuration View. The non-selected prompts will not be used by the agent.

The configuration overview also shows which variables and functions are used by the agent, including the ones specified in the prompt (customizations).

At the moment the overview shows a union of all variables and functions of all prompt variant (customizations). However this is misleading as the ones of the non-selected variants will actually not be used.

The overview should be adapted to only show the variables and functions of the selected prompt (customization).

Image

### Additional Information

- Operating System: Ubuntu 24.04
- Theia Version: latest master (https://github.com/eclipse-theia/theia/commit/4877a0b03e4940c27206ed17c7221b7855e74d7e)

eneufeld:
#### What it does

Previously, the configuration overview showed a union of all variables and functions from all prompt variants, which was misleading as only the selected variant's variables and functions are actually used by the agent.

This change:
- Modifies the prompt template parser to only analyze selected variants
- Adds event handling to update the UI when prompt variants are changed
- Improves state management for asynchronous variant selection
- Adds loading states for better user feedback

The configuration overview now correctly displays only the variables and functions from the currently selected prompt variants, making the UI more accurate and reducing potential confusion for users.

Fixes #15320

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


sdirix (changes requested):
Does not properly work yet:

I modified the Coder default template a bit:

image

- License is showing up as variable
- Used variable is marked as `Undeclared`
- Added variables and functions do not show up

eneufeld:
> License is showing up as variable
This is solved on main
> Used variable is marked as Undeclared
This is a bug on main

Good catch with the other one. Pushing an update

package-lock.json
```diff
@@ -263,6 +263,7 @@
     "dev-packages/ffmpeg": {
       "name": "@theia/ffmpeg",
       "version": "1.61.0",
+      "hasInstallScript": true,
```
sdirix, line 266: should be removed

sdirix (changes requested):
Mostly works for me :+1: Even adding/removing/modifying files related to the workspace template preferences updates the AI configuration view.

The following does not yet work:
- Adding the variable `{{file:packages/file-search/test-resources/subdir1/sub2/foo.txt}}` to a prompt, let's it show up in the "agent-specific" variables, however I expected it to show in the global variables list.
- Changing the template folder preference `ai-features.promptTemplates.promptTemplatesFolder` does not update the AI configuration view

sdirix:
@eneufeld Shall I already review or rather wait until you fixed the test cases?

eneufeld:
@sdirix  I updated, build works again.

sdirix (changes requested):
@eneufeld I pushed a merge of `master` as it was conflicting. Sadly it no longer works (e.g. does not show the use of variables at all) and needs to be reworked as the prompt management was refactored in the mean time.

package-lock.json
sdirix, line 1: We should not have package-lock.json changes with this PR

sdirix (changes requested):
- :heavy_check_mark: No matter where the active prompts come from, the list is correctly updated
- :x: There is an issue when an agent already specifies a `variables` field, e.g. the Orchestrator agent:
  - :heavy_check_mark: when the variable is removed from the prompt, then it's still shown in the agent UI. This is expected as the agent declared the variable manually.
  - :x: however if the prompt specifies *additional* variables, then these are not listed. My expectation is that additional variables are listed there. Observe that if an agent specifies `functions`, then additional functions found in the prompts are still added to the list

eneufeld:
Fixed the issue. Problem was that orchestrator does not define further variants and the code handled that incorrectly.

PR #15538, opened 2025-04-30
Make monaco commands available through command palette for notebook cells

jonah-iden:
#### What it does

There was a problem with notebook cells, that when opening the command palette and trying to execute a monaco command like `Move Line Up`, it was not there because the cell editor would lose focus when the command palette was opened. 
In general, a normal Editor can either have text focus or be active, which means the user can not type but still execute commands affecting the text. previously only Standalone Editor Widgets could be the active editor. All others could only be focused. 

#### How to test

1. Open a notebook. 
2. focus a cell
3. Press `f1` to open the command palette
4. see move line up should be available.

It would also be nice to test if standalone editors still work fine. In my tests i didn't see any change in behaviour but just to be sure i did not overlook something.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -263,6 +263,7 @@
     "dev-packages/ffmpeg": {
       "name": "@theia/ffmpeg",
       "version": "1.61.0",
+      "hasInstallScript": true,
       "license": "EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0",
       "dependencies": {
         "@electron/get": "^2.0.0",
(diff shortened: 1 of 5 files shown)


dhuebner (commented):
The expected functionality works, code changes are good.
One thing I noticed is that if you open a new Text editor and switch back to notebook, the `getActiveCodeEditor()` in the `monaco-editor-service` returns the previously opened Text editor when checking for command enablement. Even thought the command execution is applied  in the notebook cell afterwards. This is strange, could you check this please?

dhuebner:
> The expected functionality works, code changes are good. One thing I noticed is that if you open a new Text editor and switch back to notebook, the `getActiveCodeEditor()` in the `monaco-editor-service` returns the previously opened Text editor when checking for command enablement. Even thought the command execution is applied in the notebook cell afterwards. This is strange, could you check this please?

Seems to be a debugging issue, debug logging worked well.

dhuebner (approved):
Works as expected, thanks!

PR #15570, opened 2025-05-08
Preserve change sets across chat branch changes

### LINKED ISSUE #15513: Store state of Change Set for each branch tip when editing messages
With https://github.com/eclipse-theia/theia/pull/15479, we introduced editing messages, which will branch off the edited message in the chat model. However, the change set is only stored once per chat model. Therefore, we decided to just reset the change set when branching off and keep only one state of the change set per chat model.

It would be better to store the state of the change set for each tip of a branch, so that when we switch branch, we restore the change set that was suggested in that branch. When branching off for an edited message, we should create a copy of the current change set, if there is any.

colin-grant-work:
#### What it does

Fixes  #15513, and fixes an issue where the same chat message couldn't be edited more than once.

Basically, storage of chat sets has been moved to requests rather than the session, and a system that aims to copy-on-write has been implemented, so that change sets from parents are used, but if a branch is created and then a change set is set or modified, a new change set copy is created and placed at the branch point.

> [!NOTE]
> Things are perhaps a bit more complicated here than they need to be because the change set API envisions callers creating change sets themselves and calling `setChangeSet` to place it on a given ChatModel and then having the ability to retrieve it and interact with it directly. The copy-on-write functionality would be obviated by placing all of the interactions with changesets behind the `ChatModel`, and other record keeping would be simplified, as well. I'd be more than happy - delighted - to refactor that to effect.

#### How to test

1. Engage in chats that generate change sets.
2. Branch off of those chats.
3. Engage in further chats that modify the change set.
4. Any changes made (including change set removal) after branching should not affect the 'original' change set, and switching to a different branch should restore that branch's change set.
> This could be taken further than at present. For example, if you switch to a different chat branch, and both the branch you switch away from and the branch you switch to have a change for the same file, if the editor is open for that changed file, it could be switched to the text from the current branch.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


planger (commented):
Thank you very much! This works great. I like how that the change set is now managed and positioned in the chat request.

I just found one potential bug (see video below):
Observe how the change set is copied from the second request, even though one file (`template.env`) was actually only added by the second request and not the first. I'd have expected that only the `README.md` changes (from the first request) remain in the change set after modifying the second request. Instead, the file change suggested by the second was copied too.

For subsequent request edits, this worked fine.

https://github.com/user-attachments/assets/f8ba28a0-c877-4cd7-804c-149be57ed72b

Other than that, this looks great!

Regarding your comment to further simplify things: If it is reasonable effort, I tend to think it is worth it, as we are planning to extend the change set capabilities. In particular, we are thinking about a "write-through" change set that directly writes into the file system, if the user allows it to. This way, we can more easily leverage other tools, such as linters and compilers that operate on the file system, to fix up some issues the LLM introduced in one go.

This would mean that we also would have to reset the state of the file system to the state it had at a certain request, if a user edits a message. We are not sure yet if it is worth building this functionality based on git instead, but either way, the simpler the change set handling with message editing is now, the easier it might be to build this functionality in the future, I suppose.

colin-grant-work:
> I just found one potential bug (see video below):
> Observe how the change set is copied from the second request, even though one file (template.env) was actually only added by the second request and not the first. I'd have expected that only the README.md changes (from the first request) remain in the change set after modifying the second request. Instead, the file change suggested by the second was copied too.

This intended / expected, but of course modifiable behavior. The logic in play here is this:
 - First request -> Change set created on first request.
 - Second request `template.env` -> Change set from first request used. Because there is no branching, the second request is viewed as transparent.
- Third request (branching to `TEMPLATE.env`) -> Full change set copied from first request, which includes the `template.env` modification.

The principle is that we avoid copying where unnecessary, but we could instead copy every time any modification is made, or create a new one for every addition and accumulate entries from active ancestors.

planger:
@colin-grant-work I see, thanks for the clear feedback!

To me it would feel more logical, if we'd copy the change set of the `n-1` request, if request `n` is modified, because we are essentially cutting of the conversation at request `n`, so all consequences of request `n` (answers and change set modifications) are removed for the new branch.

In my view, the change set of a request is similar to the response given to the request -- so a consequence and not part of it. So replacing the request, replaces the resulting response and the resulting modifications of the change set.

What do you think?

planger:
@colin-grant-work I wanted to follow-up with you, whether you'd agree with changing this PR according to my suggestion above? Please let me know if we can help with anything. Thanks!

planger (approved):
Excellent, thank you very much! Works now very nicely and feels intuitive to me.

planger (approved):

PR #15571, opened 2025-05-08
Add default context resolution for custom agents

JonasHelming:
#### What it does

- Add the default context details variable to the template of custom agents so that by default they resolve all files in the context

#### How to test

See this video:

default-context-custom-agents

#### Follow-ups

We might move the template entry to a contribution and outside of the core package.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/browser/frontend-prompt-customization-service.ts
+++ b/packages/ai-core/src/browser/frontend-prompt-customization-service.ts
@@ -30,7 +30,17 @@ const templateEntry = {
     id: 'my_agent',
     name: 'My Agent',
     description: 'This is an example agent. Please adapt the properties to fit your needs.',
-    prompt: 'You are an example agent. Be nice and helpful to the user.',
+    prompt: `{{!-- Note: The context section below will resolve all context elements (e.g. files) to their full content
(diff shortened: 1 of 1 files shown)


packages/ai-core/src/browser/frontend-prompt-customization-service.ts
```diff
@@ -30,7 +30,17 @@ const templateEntry = {
...
+in the system prompt. Context elements can be added by the user in the default chat view (e.g. via DnD or the "+" button).
+If you want a more fine-grained, on demand resolvement of context elements, you can also resolve files to their paths only
+and equip the agent with functions so that the LLM can retrieve files on demand. See the Coder Agent prompt for an example.--}}
+
+# Role
+You are an example agent. Be nice and helpful to the user.
+
+## Current Context
+Some files and other pieces of data may have been added by the user to the context of the chat. If any have, the details can be found below.
+{{contextSummary}}`,
```
eneufeld, line 43: don't we need the `contextDetails` variable?

JonasHelming: Good catch :-)

eneufeld (approved):

PR #15598, opened 2025-05-15
[vscode] Support CommentAuthorInformation in CommentThread canreply

### LINKED ISSUE #15558: [vscode] CommentThread.canReply can now be also CommentAuthorInformation
Since VS Code 1.100, CommentThread.canReply property can be either boolean or CommentAuthorInformation. It was only boolean previously. 
Default value remains `true`

rschnekenbu:
#### What it does

Adds the support of the CommentAuthorInformation in the canReply CommentThread property. This property was only boolean before, and still defaults to true currently. 

There were also some adaptation needed in the CommentThreadWidget widget. The canReply property drives the content of the reply text area, where only the text area is displayed when canReply is true, and the author information (icon and name) are displayed when canReply has some AuthorInformation. 

Finally, some adaptations were required since the widget did not work anymore with the refactoring on menu nodes (https://github.com/eclipse-theia/theia/pull/14676). Arguments were not initially available to test for visibility and enablement of the action, and the widget was failing.

fixes #15558

#### How to test

1. Install following extension:
- src:[vscode-comment-api-example-0.0.5-src.zip](https://github.com/user-attachments/files/20225896/vscode-comment-api-example-0.0.5-src.zip)
- vsix zipped: [vscode-comment-api-example-0.0.5.zip](https://github.com/user-attachments/files/20225903/vscode-comment-api-example-0.0.5.zip)
2. Have in the workspace 2 different markdown files, more specifically ones that have the 'md' extension
3. Compare both of them
4. The Comment widget comes, with several actions. The "Toggle Reply" action cycle between canReply is true and canReply has some author information. The "Disable CanReply" action sets the canReply to false, and then all actions are disabled on the widget. The reply button let you enter a new comment in the thread.

#### Follow-ups

When menus are the ones mapped from vscode, e.g. from this mapping:
https://github.com/eclipse-theia/theia/blob/23e0bec71cf0df5b2537332808c4642d8efd681d/packages/plugin-ext/src/main/browser/menus/vscode-theia-menu-mappings.ts#L76
It likely seems that the menu node is undefined if no contributions are made to the menu. 
That causes an error message from the menu registry. See
https://github.com/eclipse-theia/theia/blob/23e0bec71cf0df5b2537332808c4642d8efd681d/packages/core/src/common/menu/menu-model-registry.ts#L331
 
This may pollute the error log. I added some check there: 
https://github.com/eclipse-theia/theia/commit/6bad05bd37084c5c76cf9906a5d12d286f320be9#diff-9030451153f2a294811451cfe5660a9df850b425d6b4f5a342246e15dc5eb066R563
I don't think this would be a long term solution.  

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
+++ b/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
@@ -743,8 +743,8 @@ export interface CommentThread {
     onDidChangeState: TheiaEvent<CommentThreadState | undefined>;
     onDidChangeCollapsibleState: TheiaEvent<CommentThreadCollapsibleState | undefined>;
     isDisposed: boolean;
-    canReply: boolean;
-    onDidChangeCanReply: TheiaEvent<boolean>;
+    canReply: boolean | theia.CommentAuthorInformation;
+    onDidChangeCanReply: TheiaEvent<boolean | theia.CommentAuthorInformation>;
 }
 
 export interface CommentThreadChangedEventMain extends CommentThreadChangedEvent {
--- a/packages/plugin-ext/src/common/plugin-api-rpc.ts
+++ b/packages/plugin-ext/src/common/plugin-api-rpc.ts
@@ -2091,7 +2091,7 @@ export type CommentThreadChanges = Partial<{
     comments: Comment[],
     collapseState: CommentThreadCollapsibleState;
     state: CommentThreadState;
-    canReply: boolean;
+    canReply: boolean | theia.CommentAuthorInformation;
 }>;
 
 export interface CommentsMain {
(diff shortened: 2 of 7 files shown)


tsmaeder (approved):
Code looks good and seems to work as advertised.

rschnekenbu:
@tsmaeder, thanks a lot for the review!
Any thoughts about the follow up point? Is this something you are aware of? It seems like I have a couple of these messages now on vscode mapped menus (plugin_view/title or plugin_cm/title)

tsmaeder:
> @tsmaeder, thanks a lot for the review! Any thoughts about the follow up point? Is this something you are aware of? It seems like I have a couple of these messages now on vscode mapped menus (plugin_view/title or plugin_cm/title)

Not really sure: throwing an exception seems atypical for a "get" method. I guess the idea is that the menu paths are well known. As a workaround, one can use `getMenuNode()`, which may return undefined.

tsmaeder:
Come to think of it:, IMO the component displaying the menu should register it with the menu registry. The VS Code API effectively requires all components that the VS Code API contributes into, so for example the comments component, as well.

PR #15642, opened 2025-05-24
Improve File Change Set Functions

JonasHelming:
#### What it does

- Accumulate changes when using the replace function
- Add an optional reset parameter to the replace function
- Add a clear function to clear change for a file changeset element
- Add a getProposedFileState function

#### How to test

Use Coder with the default or the next prompt. Do incremental changes on a file without accepting them.
Ask it questions about a state that has not been applied.
Ask it to undo changes on a file.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


JonasHelming:
@colin-grant-work If possible, I would like to get this into the release on Wedneday. If you do not have the time for a review, please let me know.

packages/ai-ide/src/browser/file-changeset-functions.ts
```diff
@@ -194,12 +215,76 @@ export class ReplaceContentInFileFunctionHelper {
...
+
+    private findExistingChangeElement(changeSet: ChangeSet, fileUri: URI): ChangeSetFileElement | undefined {
+        const elements = changeSet.getElements();
+        for (const element of elements) {
+            if (element.uri && element.uri.toString() === fileUri.toString()) {
+                return element as ChangeSetFileElement;
+            }
+        }
+        return undefined;
+    }
```
colin-grant-work, line 235: Saves a bit of redundant checking below if we can pass in a potentially undefined changeSet here.

```suggestion
    private findExistingChangeElement(changeSet: ChangeSet | undefined, fileUri: URI): ChangeSetFileElement | undefined {
        return changeSet?.getElements().find((candidate): candidate is ChangeSetFileElement => candidate instanceof ChangeSetFileElement && candidate.uri.isEqual(fileUri));
    }
```

packages/ai-ide/src/browser/file-changeset-functions.ts
```diff
@@ -194,12 +215,76 @@ export class ReplaceContentInFileFunctionHelper {
...
+                return `No pending changes found for file ${path}.`;
+            }
+
+            const elements = ctx.session.changeSet.getElements();
+            const indicesToRemove: number[] = [];
+            elements.forEach((element, index) => {
+                if (element.uri && element.uri.toString() === fileUri.toString()) {
+                    indicesToRemove.push(index);
+                }
+            });
```
colin-grant-work, line 251: Here also, the optional operator can save us repeating the 'No pending changes' check:

```suggestion
            const indicesToRemove: number[] = [];
            ctx.session.changeSet?.getElements().forEach((element, index) => {
                if (element.uri && element.uri.toString() === fileUri.toString()) {
                    indicesToRemove.push(index);
                }
            });
```

Now we end up in the branch where the array has length 0.

But it may make sense to add deletion by URI to the ChangeSet as well (in https://github.com/eclipse-theia/theia/pull/15570 I've changed the `removeElements` method to operate on URI's rather than indices.)

packages/ai-ide/src/common/coder-replace-prompt-template.ts
```diff
@@ -167,11 +169,15 @@ Use the following function to retrieve a list of problems in a file if the user
 To propose code changes or any file changes to the user, never print code or new file content in your response.
 
 Instead, for each file you want to propose changes for:
-- **Always Retrieve Current Content**: Use ${FILE_CONTENT_FUNCTION_ID} to get the latest content of the target file.
-- **Change Content**: Use ~{changeSet_writeChangeToFile} or ~{changeSet_replaceContentInFile} to propose file changes to the user.\
-If ~{changeSet_replaceContentInFile} continously fails use ~{changeSet_writeChangeToFile}. Calling a function on a file will override previous \
-function calls on the same file, so you need exactly one successful call with all proposed changes per changed file. The changes will be presented as a applicable diff to \
-the user in any case.'
+- **Always Retrieve Current Content**: Use ${FILE_CONTENT_FUNCTION_ID} to get the original content of the target file.
+- **View Pending Changes**: Use ~{changeSet_getProposedFileState} to see the current proposed state of a file, including all pending changes.
```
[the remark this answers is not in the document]
JonasHelming: I thought about this quite a bit. If we combine it, the agent has no awareness of the changeset anymore, so I cannot for example look at the original state anymore. We have to test this in practise, but I came to the conclusion (during original tests) that it handles the functions appropriatly.

packages/ai-ide/src/browser/file-changeset-functions.ts
JonasHelming: Can you harmoize this one while merging then? i would also prefer to have the remove function on the changset

PR #15683, opened 2025-05-28
fix: temporary scroll lock when user scrolls up in AI Chat

### LINKED ISSUE #15049: [Theia AI] Bad UX when scrolling up while a tool and its spinner is running or while streaming the response...
### Bug Description:
Hi,

### Steps to Reproduce:

1. Use the Coder agent to apply some changes on a file, while it is calling its changeSet_writeChangeToFile for example (since it takes time), try to scroll up.
2.  Or let the LLM stream some long content, and try to scroll up.

I know that we can turn off the auto-scroll, but if not the UX of the scrolling is bad, it shakes all the chant view !!

### Additional Information

- Operating System: mac
- Theia Version: latest master

CamilleLetavernier:
- when user tries to scroll up in the Chat View, temporarily enable scroll lock, even if it is originally disabled
- restore the previous scroll lock state when scrolling back to the bottom

fixes #15049

#### What it does

This PR improves the UX for scrolling behavior in the AI Chat view, when the LLM is active:
- If scroll lock is enabled: no change
- If scroll lock is disabled: automatically enable it when scrolling up, then automatically disable it again when scrolling back down

#### How to test

- Open the AI Chat view
- Turn off scroll-lock
- Send a request to a verbose agent (e.g. "`@Universal` Tell me a long story")
  - Reducing the width and height of the Chat window can help to get more scrolling opportunities! :) 
- While the Agent is replying, scroll up: the window should no longer scroll
- Scroll back to the bottom: auto-scroll resumes

Note: if Scroll-lock is explicitly enabled, the new behavior doesn't trigger; auto-scroll will be disabled no matter what

#### Follow-ups

The main issue I had was related to how scroll behavior is handled in the base Widget hierarchy. TreeWidget can be virtualized (which it is for the AI Chat), and in that case, it will not use the scroll-related APIs defined by the base Widgets. For example, "getScrollContainer" doesn't work in the AI Chat view. I had to introduce new APIs to handle the Virtuoso scroll events, so we now have at least two sets of APIs in parallel (regular BaseWidget scroll API with PerfectScrollbar, and virtualized TreeWidget APIs with Virtuoso).

These inconsistencies might cause issues and confusion in the future (although this isn't something new - this PR makes it a bit worse)

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/tree/tree-widget.tsx
+++ b/packages/core/src/browser/tree/tree-widget.tsx
@@ -63,6 +63,11 @@ export const TREE_NODE_CAPTION_CLASS = 'theia-TreeNodeCaption';
 export const TREE_NODE_INDENT_GUIDE_CLASS = 'theia-tree-node-indent';
 
 /**
+ * Threshold in pixels to consider the view as being scrolled to the bottom
+ */
+export const SCROLL_BOTTOM_THRESHOLD = 30;
+
+/**
  * Tree scroll event data.
  */
 export interface TreeScrollEvent {
@@ -1198,7 +1203,6 @@ export class TreeWidget extends ReactWidget implements StatefulWidget {
             return scrollState?.isAtBottom ?? true;
         } else {
             // Fallback to DOM-based calculation for non-virtualized trees
-            const threshold = 30;
             const scrollContainer = this.node;
             const scrollHeight = scrollContainer.scrollHeight;
             const scrollTop = scrollContainer.scrollTop;
@@ -1208,7 +1212,7 @@ export class TreeWidget extends ReactWidget implements StatefulWidget {
                 return true;
             }
 
-            return scrollHeight - scrollTop - clientHeight <= threshold;
+            return scrollHeight - scrollTop - clientHeight <= SCROLL_BOTTOM_THRESHOLD;
         }
     }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/browser/tree/tree-widget.tsx
```diff
@@ -1592,6 +1653,16 @@ export namespace TreeWidget {
                         });
                     }
                 }}
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                onScroll={(e: any) => {
+                    const scrollTop = e.target.scrollTop;
+                    const scrollHeight = e.target.scrollHeight;
+                    const clientHeight = e.target.clientHeight;
+                    const isAtBottom = scrollHeight - scrollTop - clientHeight <= 30; // 30px threshold
```
planger: nit: I think it'd be more clear if we'd assign the 30 to a variable.
```suggestion
                    const thresholdInPx = 30;
                    const isAtBottom = scrollHeight - scrollTop - clientHeight <= thresholdInPx;
```

planger (approved):
Great, thank you! Definitely a huge improvement and code looks good to me. Also I couldn't see a side-effect due to the new API.

One thing I noticed:
If I scroll up during inference, then wait a bit, and then scroll further, I see a bit of a shift. It is not critical, but just a tiny glitch that I noticed. Shouldn't block merging this PR though.

PR #15686, opened 2025-05-29
Start debug instance without Yarn

colin-grant-work:
#### What it does

Before the switch to NPM, the extension host plugin setup for Electron used Yarn's wrapping of the `node_modules/.bin` binaries to invoke the `theia` command. With the switch to NPM, that no longer works, and extension host debugging doesn't successfully launch a new window in Electron. This PR adds a script to the Electron`package.json` to invoke `theia start` with no additional arguments and then uses it in the extension host debug startup process.

#### How to test

1. Attempt to debug a VSCode extension in Theia's electron app.
2. An extension host window should pop up and you should be able to debug your extension.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/examples/electron/package.json
+++ b/examples/electron/package.json
@@ -100,6 +100,7 @@
     "lint": "theiaext lint",
     "rebuild": "theia rebuild:electron --cacheRoot ../..",
     "start": "theia start --plugins=local-dir:../../plugins --ovsx-router-config=../ovsx-router-config.json",
+    "start:raw": "theia start",
     "start:debug": "npm run -s start --log-level=debug --remote-debugging-port=9222",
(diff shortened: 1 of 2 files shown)


tsmaeder:
@colin-grant-work while this works in the electron example, it does not help for Theia IDE, where npm is not available. @sgraband didn't you work on a fix for this once?

tsmaeder:
@colin-grant-work since `plugin-dev` is a package that can be consumed by adopters, anything using `npm` is not the right fix, IMO, since npm will not be available at run time. We really need to figure out the correct start incantation for starting the plugin host, like we do in the browser case.

colin-grant-work:
I was suspicious of this solution - and suspicious of its previous `yarn` iteration - I'll see whether I can track down a more generally applicable mechanism.

sgraband:
@colin-grant-work Here is a PR (i ran out of time and ideas back then) with some feedback of @tsmaeder: https://github.com/eclipsesource/theia/pull/96

Maybe this helps.

PR #15696, opened 2025-06-02
fix: support more than 10 tool calls via OpenAI SDK

sdirix:
#### What it does

We use the convenience `runTools` API of the OpenAI SDK to handle requests with tool calls. By default, the SDK stops after 10 chat completions, whereby each function call is counting as one. Thereby the SDK just stops generating in use cases with many tool requests (e.g. Coder with GPT4.1).

We raise the limit to 100 now by default. The option is also made configurable for adopters.

Also updates to the latest OpenAI SDK major version 5 and adjusts the code where necessary.

See https://github.com/eclipsesource/theia/issues/219 for more information

#### How to test

1. Configure `GPT4.1` for `@Coder`.
2. Send the following request:
```
@Coder Add a toolbar item to the widget in packages/ai-ide/src/browser/ai-configuration/token-usage-configuration-widget.tsx that triggers a newly defined command called "reset token count". In the handler, just log something on the console for now.
```

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-openai/src/node/openai-language-model.ts
+++ b/packages/ai-openai/src/node/openai-language-model.ts
@@ -33,9 +33,10 @@ import { RunnableToolFunctionWithoutParse } from 'openai/lib/RunnableFunction';
 import { ChatCompletionMessageParam } from 'openai/resources';
 import { StreamingAsyncIterator } from './openai-streaming-iterator';
 import { OPENAI_PROVIDER_ID } from '../common';
-import { FinalRequestOptions } from 'openai/core';
+import type { FinalRequestOptions } from 'openai/internal/request-options';
+import type { RunnerOptions } from 'openai/lib/AbstractChatCompletionRunner';
 
-class MistralFixedOpenAI extends OpenAI {
+export class MistralFixedOpenAI extends OpenAI {
     protected override async prepareOptions(options: FinalRequestOptions): Promise<void> {
         (options.body as { messages: Array<ChatCompletionMessageParam> }).messages.forEach(m => {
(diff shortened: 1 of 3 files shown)


packages/ai-openai/src/node/openai-language-model.ts
```diff
@@ -61,6 +62,16 @@ export type DeveloperMessageSettings = 'user' | 'system' | 'developer' | 'mergeW
...
 export class OpenAiModel implements LanguageModel {
 
+    /**
+     * The options for the OpenAI runner.
+     */
+    protected runnerOptions: RunnerOptions = {
+        // The maximum number of chat completions to return in a single request.
+        // Each function call counts as a chat completion.
+        // To support use cases with many function calls (e.g. @Coder), we set this to a high value.
+        maxChatCompletions: 100,
```
sdirix, line 72: Info for reviewers: I set this to 100. With the example prompt in the PR description, GPT4.1 triggered 40 tool calls, collecting 600k tokens. Do you think we should raise the limit even higher?

JonasHelming (approved):
LGTM

packages/ai-openai/src/node/openai-language-model.ts
JonasHelming: Let Go with 100 for now, we can evaluate with the preview whether that works for most cases

JonasHelming:
Question: Are there similar limits for the Google AI API?

sdirix:
> Question: Are there similar limits for the Google AI API?

I can check later today

PR #15713, opened 2025-06-03
AppTester Agent via Browser Automation

martin-fleck-at:
#### What it does

This pull request introduces a new `AppTesterChatAgent` to the AI IDE, enhancing its functionality by integrating Playwright-based application testing capabilities through an MCP server.
See also https://github.com/eclipsesource/theia/issues/196

#### How to test

- If you use the browser app of Theia, you can ask the AppTester agent to test the app, e.g., `@AppTester test the application at http://localhost:3000`.  It should use the Playwright MCP server to open a new window, browse to the location and perform some tests. Maybe you need to be more specific in what to test.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


packages/ai-ide/src/common/app-tester-chat-agent.ts
```diff
@@ -0,0 +1,123 @@
...
+         }
+
+         const availableServers = await this.mcpService.getServerNames();
+         if (!availableServers.includes(EXPECTED_MCP_SERVER_NAME)) {
+            await this.mcpService.addOrUpdateServer({
+               name: EXPECTED_MCP_SERVER_NAME,
+               command: 'npx',
+               args: ['-y', '@playwright/mcp@latest'],
+               env: {},
+            });
```
martin-fleck-at: we probably need to address how we want to make sure that the MCP server is running. It is a pre-requisite for this agent since otherwise it will just produce nonsense code but there should be a feedback loop with the user since we are installing and starting a new server. I'd love to have a good discussion about a proper way to achieve this or surface to the user that otherwise this agent is not usable.

haydar-metin (approved):
@martin-fleck-at thanks for the implementation, now in the last commit we install the MCP server only after asking the user and the servers will be terminated after closing the connection.

planger (approved):

PR #15720, opened 2025-06-04
fix: filter out error messages from the messages send to the llm

eneufeld:
#### What it does

The llm does not care why the previous run was halted. So remove error messages from the list of messages send to the llm.

#### How to test

Break the llm request e.g by providing invalid request settings ("max_tokens":"aaa")

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-agents.ts
+++ b/packages/ai-chat/src/common/chat-agents.ts
@@ -61,6 +61,7 @@ import {
     ToolCallChatResponseContentImpl,
     ChatRequestModel,
     ThinkingChatResponseContentImpl,
+    ErrorChatResponseContent,
 } from './chat-model';
 import { parseContents } from './parse-contents';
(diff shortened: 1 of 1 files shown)


packages/ai-chat/src/common/chat-agents.ts
```diff
@@ -259,7 +260,12 @@ export abstract class AbstractChatAgent implements ChatAgent {
                 text: text,
             });
             if (request.response.isComplete || includeResponseInProgress) {
-                const responseMessages: LanguageModelMessage[] = request.response.response.content.flatMap(c => {
+                const responseMessages: LanguageModelMessage[] = request.response.response.content.filter(c => {
+                    if (ErrorChatResponseContent.is(c)) {
+                        return false;
+                    }
+                    return true;
+                }).flatMap(c => {
```
sdirix: Minor: Should we write it like this: `c => !ErrorChatResponseContent.is(c)`

sdirix (approved):
Works for me although I would prefer less verbose code

PR #15724, opened 2025-06-04
Enact code actions when change set file created

colin-grant-work:
#### What it does

This PR ensures that code actions configured to run on save (via `editor.codeActionsOnSave`) are also executed when a changeset file element is created, not just when a file is saved in the editor. Previously, code actions such as formatting, fixing, or trailing whitespace removal would only be applied on explicit save operations. With this change, these actions are now consistently applied to changeset file elements at creation, ensuring that generated or modified content adheres to the same standards and automation as regular file saves.

Key changes:
- Refactored the handling of code actions on save to centralize logic in the Monaco code action save participant.
- Removed redundant or duplicate code action application logic from the changeset file element implementation.
- Cleaned up related code and removed unused utilities and services.

#### How to test

1. Configure code actions on save in your user or workspace settings, e.g.:
   ```json
   "editor.codeActionsOnSave": {
     "source.fixAll": true,
     "source.organizeImports": true
   }
   ```
2. Create or modify a changeset file element (e.g., through the AI chat or code review features).
3. Verify that the configured code actions (such as formatting, fixAll, or organize imports) are automatically applied to the changeset file element upon creation.
4. Compare the behavior to a regular file save to ensure consistency.
5. Check that no regressions occur in file save or changeset workflows.

#### Breaking changes

- [ ] This PR introduces breaking changes and requires careful review. If yes, the breaking changes section in the changelog has been updated.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed the review guidelines

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with the review guidelines


ndoschek:
@haydar-metin will take a look :eyes:

colin-grant-work:
I've updated the original PR to better handle the write through case, with the caveat that it now applies code actions / formatting to the value by using the file's Monaco model. This has the consequence that if the user had dirty changes in a Monaco editor and then accepts Coder's suggestions (automatically in the write through case, or manually), they will lose their dirty changes. This is motivated by the fact that some code actions may expect to be operating on the actual file to, e.g., resolve imports / references, and we can only satisfy that expectation if we allow those tools to actually act on a model of the underlying file.

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -213,8 +250,130 @@ export class ChangeSetFileElement implements ChangeSetElement {
...
+            }
+
+            // Trim trailing whitespace
+            const trimTrailingWhitespace = this.fileSystemPreferences.get({ preferenceName: 'files.trimTrailingWhitespace', overrideIdentifier: languageId }, undefined, uriStr);
+            if (trimTrailingWhitespace) {
+                const ttws = new TrimTrailingWhitespaceCommand(new Selection(1, 1, 1, 1), [], false);
+                CommandExecutor.executeCommands(model.textEditorModel, [], [ttws]);
+            }
+
+            // Insert final newline
```
sgraband:
```suggestion
```

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -213,8 +250,130 @@ export class ChangeSetFileElement implements ChangeSetElement {
...
+                await instantiation.invokeFunction(
+                    formatDocumentWithSelectedProvider,
+                    model.textEditorModel,
+                    FormattingMode.Explicit,
+                    { report(): void { } },
+                    CancellationToken.None, true
+                );
+            }
+
+            // Trim trailing whitespace
```
sgraband:
```suggestion
```

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -213,8 +250,130 @@ export class ChangeSetFileElement implements ChangeSetElement {
...
+
+        return this.targetState;
+    }
+
+    /**
+     * Applies formatting preferences like format on save, trim trailing whitespace, and insert final newline.
+     */
+    protected async applyFormatting(model: MonacoEditorModel, languageId: string, uriStr: string): Promise<void> {
+        try {
+            // Format on save
```
sgraband:
```suggestion
```

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -213,8 +250,130 @@ export class ChangeSetFileElement implements ChangeSetElement {
...
+
+            this._targetStateWithCodeActions = tempModel.object.textEditorModel.getValue();
+            if (this._changeResource?.contents === this.elementProps.targetState) {
+                this._changeResource?.update({ contents: this.targetState });
+            }
+        } catch (error) {
+            console.warn('Failed to apply code actions to target state:', error);
+            this._targetStateWithCodeActions = targetState;
+        } finally {
+            // Dispose the temporary model and resource if they were created
```
sgraband:
```suggestion
```

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -213,8 +250,130 @@ export class ChangeSetFileElement implements ChangeSetElement {
...
+        let tempModel: IReference<MonacoEditorModel> | undefined;
+        try {
+            // Create a temporary model to apply code actions
+            const tempUri = new URI(`untitled://changeset/${Date.now()}${this.uri.path.ext}`);
+            tempResource = this.inMemoryResources.add(tempUri, { contents: this.targetState });
+            tempModel = await this.monacoTextModelService.createModelReference(tempUri);
+            tempModel.object.suppressOpenEditorWhenDirty = true;
+            tempModel.object.textEditorModel.setValue(this.targetState);
+
+            // Get language and URI for preference lookup
```
sgraband:
```suggestion
```

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -213,8 +250,130 @@ export class ChangeSetFileElement implements ChangeSetElement {
...
+            const targetContent = contents ?? this.targetState;
+            model.textEditorModel.setValue(targetContent);
+
+            const languageId = model.languageId;
+            const uriStr = this.uri.toString();
+
+            await this.codeActionService.applyOnSaveCodeActions(model.textEditorModel, languageId, uriStr, CancellationToken.None);
+            await this.applyFormatting(model, languageId, uriStr);
+
+            // Save the model
```
sgraband:
```suggestion
```

sgraband (commented):
After a short disucssion we decided that it is fine that it overwrites the users unsaved changes, because they are taken into account by the LLM, therefore they should be reflected in the changeSet.

I mostly found some unneccessary comments that should be removed, could you do that and rebase @colin-grant-work. Functionality wise this already looks good to me :+1:

sgraband (approved):
Thanks you for the changes @colin-grant-work! LGTM and works great :+1:

PR #15731, opened 2025-06-04
feat: implement prompt caching for Anthropic

### LINKED ISSUE #15729: Support Prompt caching
Anthropic supports [prompt caching](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching). We should support this too.

sdirix:
#### What it does

Mark system prompts, tools and messages for caching in the Anthropic LanguageModel. For typical agent workflows like `@Coder` this reduces token costs by a large factor and reduces the pressure on the rate limit.

Currently caching is enabled by default for all Anthropic models.

Also updates token tracking functionality to be aware of token caching and updates the Anthropic SDK.

fixes #15729

#### How to test

Use Anthropic models for all kind of use cases. Check the token usage count.

Example run on my machine:
image

#### Follow-ups

- Make prompt caching configurable for different models and/or use cases

The current structure in Anthropic is the following:
- Cache write tokens are 1.25 times the base input tokens price
- Cache read tokens are 0.1 times the base input tokens price
- Cache hits are not deducted against your rate limit.

If we should have use cases in which caching is not useful, we now make them a bit more expensive

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/node/token-usage-service-impl.ts
+++ b/packages/ai-core/src/node/token-usage-service-impl.ts
@@ -41,6 +41,8 @@ export class TokenUsageServiceImpl implements TokenUsageService {
     async recordTokenUsage(model: string, params: TokenUsageParams): Promise<void> {
         const usage: TokenUsage = {
             inputTokens: params.inputTokens,
+            cachedInputTokens: params.cachedInputTokens,
+            readCachedInputTokens: params.readCachedInputTokens,
             outputTokens: params.outputTokens,
             model,
             timestamp: new Date(),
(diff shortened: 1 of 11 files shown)


packages/ai-core/src/node/token-usage-service-impl.ts
```diff
@@ -50,7 +52,23 @@ export class TokenUsageServiceImpl implements TokenUsageService {
...
+            logMessage += ` Input Tokens read from cache: ${params.readCachedInputTokens};`;
+        }
+
+        logMessage += ` Output Tokens: ${params.outputTokens}; Model: ${model};`;
+
+        if (params.requestId) {
+            logMessage += `; RequestId: ${params.requestId}`;
+        }
+
+        console.log(logMessage);
```
eneufeld: do we actually need this? i think this is a left over

eneufeld (approved):
I left one minor nitpick

packages/ai-core/src/node/token-usage-service-impl.ts
sdirix: No that's from the original code, the code looked like this before:
```
console.log(`Input Tokens: ${params.inputTokens}; Output Tokens: ${params.outputTokens}; Model: ${model}${params.requestId ? `; RequestId: ${params.requestId}` : ''}`);
```
or do you not want the log message at all? We could also demote it to `debug` level

eneufeld: I would either remove it or demote it, we have a view now ;-)

PR #15736, opened 2025-06-04
enable Agent-to-Agent Delegation via Tool Calls

CamilleLetavernier:
#### What it does

Add a new AI Tool Provider allowing an Agent to delegate a prompt to another Agent and get the result back. The result is also displayed in a (collapsed) section in the main conversation. The delegation supports the following features:

- Display the delegated chat in a collapsed section
- Bubble-up the change sets from the delegated chat to the current chat
- Support streaming mode in the delegated chat

#### How to test

The tool needs to be added to your system prompt. For example:

```md
## Delegation
You can delegate tasks to other agents with this tool:
- **~{delegateToAgent}**

Agents available for delegation are:
- `Architect`: to analyse the current code base and design new features
- `Coder`: to implement code
- `Universal`: for general purpose questions
```

Then prompt for delegation (For real-world use cases, the Agents might be smart enough to delegate on their own when they need to; but for testing it's easier to just explicitly request it). Example prompt:

`@Architect ask Universal why the sky is blue, then give me a summary`

image

#### Follow-ups

- Currently, we create a new session to handle the delegated prompt. This is shown as a regular Chat in the UI, with which the user can interact. We should probably hide it somehow.
- Scrolling could be improved (the parent chat only updates scroll when receiving new tokens; but in this case, the tokens are coming to the delegated chat. Parent UI is not aware of this). This is only an issue when the delegated chat section is expanded.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-response-renderer/delegation-response-renderer.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-response-renderer/delegation-response-renderer.tsx
@@ -1,0 +1,106 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+import { inject, injectable } from '@theia/core/shared/inversify';
+import { ChatRequestInvocation, ChatResponseContent, ChatResponseModel } from '@theia/ai-chat';
+import { ChatResponsePartRenderer } from '../chat-response-part-renderer';
+import * as React from '@theia/core/shared/react';
+import { DelegationResponseContent, isDelegationResponseContent } from '@theia/ai-chat/lib/browser/delegation-response-content';
+import { ResponseNode } from '../chat-tree-view';
+import { CompositeTreeNode } from '@theia/core/lib/browser';
+import { SubChatWidgetFactory } from '../chat-tree-view/sub-chat-widget';
+
+@injectable()
+export class DelegationResponseRenderer implements ChatResponsePartRenderer<DelegationResponseContent> {
+
+    @inject(SubChatWidgetFactory)
+    subChatWidgetFactory: SubChatWidgetFactory;
+
+    canHandle(response: ChatResponseContent): number {
+        if (isDelegationResponseContent(response)) {
+            return 10;
+        }
+        return -1;
+    }
+    render(response: DelegationResponseContent, parentNode: ResponseNode): React.ReactNode {
+        return this.renderExpandableNode(response, parentNode);
+    }
+
+    private renderExpandableNode(response: DelegationResponseContent, parentNode: ResponseNode): React.ReactNode {
+        return <DelegatedChat
+            response={response.response}
+            agentId={response.agentId}
+            prompt={response.prompt}
+            parentNode={parentNode}
+            subChatWidgetFactory={this.subChatWidgetFactory} />;
+    }
+}
+
+interface DelegatedChatProps {
+    response: ChatRequestInvocation;
+    agentId: string;
+    prompt: string;
+    parentNode: ResponseNode;
+    subChatWidgetFactory: SubChatWidgetFactory;
+}
+
+interface DelegatedChatState {
+    node?: ResponseNode;
+}
+
+class DelegatedChat extends React.Component<DelegatedChatProps, DelegatedChatState> {
+    private widget: ReturnType<SubChatWidgetFactory>;
+
+    constructor(props: DelegatedChatProps) {
+        super(props);
+        this.state = {
+            node: undefined
+        };
+        this.widget = props.subChatWidgetFactory();
+    }
+
+    override componentDidMount(): void {
+        this.props.response.responseCompleted.then(chatModel => {
+            this.setState({ node: mapResponseToNode(chatModel, this.props.parentNode) });
+        });
+    }
+
+    override render(): React.ReactNode {
+        const { agentId, prompt } = this.props;
+        return (
+            <div className="theia-toolCall">
+                <details className="delegation-response-details">
+                    <summary>
+                        <strong>Agent:</strong> {agentId}
+                    </summary>
+                    <div>
+                        <div><strong>Delegated prompt:</strong> {prompt}</div>
+                        <div className='delegation-response-placeholder'>
+                            {this.state.node && this.widget.renderChatResponse(this.state.node)}
+                        </div>
+                    </div>
+                </details>
+            </div>
+        );
+    }
+}
+
+function mapResponseToNode(response: ChatResponseModel, parentNode: ResponseNode): ResponseNode {
+    return {
+        id: response.id,
+        parent: parentNode as unknown as CompositeTreeNode,
+        response
+    };
+}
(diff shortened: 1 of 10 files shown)


packages/ai-chat-ui/src/browser/chat-response-renderer/delegation-response-renderer.tsx
```diff
@@ -0,0 +1,155 @@
...
+            this.forceUpdate();
+        }).catch(error => {
+            console.error('Error in delegated chat response completion:', error);
+            // Force update anyway to show any partial content or error state
+            this.forceUpdate();
+        });
+    }
+
+    override componentWillUnmount(): void {
+        // Clean up all disposables to prevent memory leaks
```
planger:
```suggestion
```

packages/ai-chat-ui/src/browser/chat-tree-view/sub-chat-widget.tsx
planger, line 1: `SubChatWidget` seems a bit misleading. I'd go with either something related to `DelegationResponseWidget`.

packages/ai-chat/src/browser/agent-delegation-tool.ts
```diff
@@ -0,0 +1,207 @@
...
+            name: AgentDelegationTool.ID,
+            description:
+                'Delegate a task or question to a specific AI agent. This tool allows you to route requests to specialized agents based on their capabilities.',
+            parameters: {
+                type: 'object',
+                properties: {
+                    agentName: {
+                        type: 'string',
+                        description:
+                            'The name/ID of the AI agent to delegate the task to. Available agents can be found using the chatAgents variable.',
```
planger:
```suggestion
                            'The name of the AI agent to delegate the task to.',
```

I don't think the reference to the `chatAgents` variable helps here as the agent can't resolve this variable autonomously right?

packages/ai-chat/src/browser/agent-delegation-tool.ts
```diff
@@ -0,0 +1,207 @@
...
+
+    @inject(ChatServiceFactory)
+    protected readonly getChatService: () => ChatService;
+
+    getTool(): ToolRequest {
+        return {
+            id: AgentDelegationTool.ID,
+            name: AgentDelegationTool.ID,
+            description:
+                'Delegate a task or question to a specific AI agent. This tool allows you to route requests to specialized agents based on their capabilities.',
```
planger:
```suggestion
                'Delegate a task or question to a specific AI agent. This tool allows you to submit requests to specialized agents based on their capabilities.',
```

packages/ai-chat/src/browser/agent-delegation-tool.ts
```diff
@@ -0,0 +1,207 @@
...
+        arg_string: string,
+        ctx: MutableChatRequestModel
+    ): Promise<string> {
+        try {
+            const args = JSON.parse(arg_string);
+            const { agentName, prompt } = args;
+
+            if (!agentName || !prompt) {
+                const errorMsg = 'Both agentName and prompt parameters are required.';
+                console.error(errorMsg, { agentName, prompt });
```
planger: Probably better to use `ILogger`.

packages/ai-chat/src/browser/agent-delegation-tool.ts
```diff
@@ -0,0 +1,207 @@
...
+            }
+
+            // Check if the specified agent exists
+            const agent = this.getChatAgentService().getAgent(agentName);
+            if (!agent) {
+                const availableAgents = this.getChatAgentService()
+                    .getAgents()
+                    .map(a => a.id);
+                const errorMsg = `Agent '${agentName}' not found or not enabled. Available agents: ${availableAgents.join(', ')}`;
+                console.error(errorMsg);
```
planger, line 91: ILogger might be better.

packages/ai-chat/src/browser/agent-delegation-tool.ts
```diff
@@ -0,0 +1,207 @@
...
+                    undefined,
+                    { focus: false },
+                    agent
+                );
+
+                // Immediately restore the original active session to avoid confusing the user
+                if (currentActiveSession) {
+                    chatService.setActiveSession(currentActiveSession.id, { focus: false });
+                }
+
```
planger, line 114: Wouldn't it work to create another method in the chatService, like `createBackgroundSession` which just creates the session, but not send events or managing it internally to work around that?

packages/ai-chat-ui/src/browser/chat-tree-view/sub-chat-widget.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    protected getChatResponsePartRenderer(content: ChatResponseContent, node: ResponseNode): React.ReactNode {
+        const renderer = this.chatResponsePartRenderers.getContributions().reduce<[number, ChatResponsePartRenderer<ChatResponseContent> | undefined]>(
+            (prev, current) => {
+                const prio = current.canHandle(content);
+                if (prio > prev[0]) {
+                    return [prio, current];
+                } return prev;
+            },
+            [-1, undefined])[1];
```
planger, line 81: I think extracting this function would help readability.

packages/ai-chat-ui/src/browser/chat-tree-view/sub-chat-widget.tsx
```diff
@@ -0,0 +1,101 @@
...
+        if (!renderer) {
+            console.error('No renderer found for content', content);
```
planger, line 83: `ILogger`?

packages/ai-chat-ui/src/browser/chat-response-renderer/delegation-response-renderer.tsx
```diff
@@ -0,0 +1,155 @@
...
+                    <div>
+                        <div><strong>Delegated prompt:</strong> {prompt}</div>
+                        <div className='delegation-response-placeholder'>
+                            {hasNode && this.state.node ? this.widget.renderChatResponse(this.state.node) :
+                                <div className="theia-ChatContentInProgress">Starting delegation...</div>
+                            }
+                        </div>
+                    </div>
+                </details>
+            </div>
```
planger: The default UI might be improved a bit:

* Show actual icons for indicating the status of the delegation
* Visually separate the delegation prompt and the result of the delegation
* Add a border around the delegation so that one doesn't loose track what is the delegation node and what is the main node when scrolling

planger (commented):
Very cool, this is imho a very powerful feature. And it already works quite well for a first iteration. I just found some minors and UI improvements that we should probably tackle before merging this.

Thank you very much!

packages/ai-chat/src/browser/agent-delegation-tool.ts
CamilleLetavernier: Theia encourages console over ILogger. From `coding-guidelines.md`:

```md
* [1.](#logging-use-console-log) Use `console` instead of `ILogger` for the root (top-level) logging.

```ts
// bad
@inject(ILogger)

this.logger.info(``);

// good
console.info(``)
```

packages/ai-chat-ui/src/browser/chat-tree-view/sub-chat-widget.tsx
CamilleLetavernier: SubChatWidget is a direct copy-paste from (a subset of) the current ChatWidget, so I'd rather keep it consistent.

packages/ai-chat/src/browser/agent-delegation-tool.ts
CamilleLetavernier: createSession expects a Location, which refers to one of the UI Elements in Theia (then the UI reacts to this); so we can't do that while relying on createSession().

While the session is a simple object, chatService only works on session it knows, so I don't think we can create/use  a session without creating it via the chatService (which in turn will necessarily move it to the UI).

We'd need a new mechanism to filter/hide sessions in some parts of the UI

packages/ai-chat/src/browser/agent-delegation-tool.ts
planger: Oh I missed that. Thanks!

CamilleLetavernier:
I've rebased the PR and solved conflicts

planger (approved):
Excellent, thank you!

I'd just add to the follow-ups:

> * Currently, we create a new session to handle the delegated prompt. This is shown as a regular Chat in the UI, with which the user can interact. We should probably hide it somehow.
> * Scrolling could be improved (the parent chat only updates scroll when receiving new tokens; but in this case, the tokens are coming to the delegated chat. Parent UI is not aware of this). This is only an issue when the delegated chat section is expanded.

* Improve styling: the outline is a bit off-set with the border, also it feels the padding is a bit too large here and there, and we use static values for font sizes, paddings, border radius, etc. where we probably should try to re-use Theia style variables instead.

Can you please create those follow-up issues once merged? Thank you!

@ndoschek FYI

PR #15743, opened 2025-06-05
Make Workspace Search function more efficient / use less tokens

jfaltermeier:
#### What it does

* The function now has a preference to control the max number of search results. If the number of results exceeds the limit, we no longer return an incomplete list but raise an error instead.
  * Very eager models might try to work around this by searching subdirectories or some other things
  * However, working with incomplete data is likely to cause other issues, because I think the LLM is not aware that it is incomplete?
* The default limit is now 30 instead of 50
  * We'll likely need to fine-tune this through testing. Lowering the limit might not be the most effective way to reduce token usage - improving prompts or tool descriptions to encourage subdirectory searches could be a better option
* Subdirectory search is now supported
* The search result format should consume fewer tokens, mainly by using relative paths and trimming excess whitespace

#### How to test

Some sample prompts

```
@Universal ~searchInWorkspace Find the string SEARCH_IN_WORKSPACE_FUNCTION_ID 

@Universal ~searchInWorkspace Find the string SEARCH_IN_WORKSPACE_FUNCTION_ID in the packages subdirectory

@Coder Find the string SEARCH_IN_WORKSPACE_FUNCTION_ID in all typescript files and update it to SEARCH_IN_WORKSPACE_FUNCTION_TOOL_ID
```

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-ide/src/browser/workspace-search-provider.ts
+++ b/packages/ai-ide/src/browser/workspace-search-provider.ts
@@ -17,11 +17,13 @@
 import { MutableChatRequestModel } from '@theia/ai-chat';
 import { ToolProvider, ToolRequest } from '@theia/ai-core';
 import { CancellationToken, URI } from '@theia/core';
+import { PreferenceService } from '@theia/core/lib/browser/preferences/preference-service';
 import { inject, injectable } from '@theia/core/shared/inversify';
 import { SearchInWorkspaceService, SearchInWorkspaceCallbacks } from '@theia/search-in-workspace/lib/browser/search-in-workspace-service';
 import { SearchInWorkspaceResult, SearchInWorkspaceOptions } from '@theia/search-in-workspace/lib/common/search-in-workspace-interface';
 import { SEARCH_IN_WORKSPACE_FUNCTION_ID } from '../common/workspace-functions';
 import { WorkspaceFunctionScope } from './workspace-functions';
+import { SEARCH_IN_WORKSPACE_MAX_RESULTS_PREF } from './workspace-preferences';
 
 @injectable()
 export class WorkspaceSearchProvider implements ToolProvider {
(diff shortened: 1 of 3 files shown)


packages/ai-ide/src/browser/workspace-search-provider.ts
```diff
@@ -111,14 +146,15 @@ export class WorkspaceSearchProvider implements ToolProvider {
                     useRegExp: args.useRegExp,
                     matchCase: false,
                     matchWholeWord: false,
-                    maxResults: this.MAX_RESULTS,
+                    maxResults: undefined, // return all results, we will fail if too many later on
```
eneufeld: let's limit this to our max +1?

packages/ai-ide/src/browser/workspace-search-provider.ts
```diff
@@ -111,14 +146,15 @@ export class WorkspaceSearchProvider implements ToolProvider {
...
                 };
 
                 if (args.fileExtensions && args.fileExtensions.length > 0) {
                     options.include = args.fileExtensions.map(ext => `**/*.${ext}`);
                 }
 
-                this.searchService.search(args.query, callbacks, options)
+                this.determineSearchRoots(args.subDirectoryPath)
```
eneufeld: can we use await?

packages/ai-ide/src/browser/workspace-search-provider.ts
```diff
@@ -143,21 +179,23 @@ export class WorkspaceSearchProvider implements ToolProvider {
...
-            const formattedResults = finalResults.map(r => {
-                const fileUri = new URI(r.fileUri);
-                const relativePath = workspaceRoot.relative(fileUri);
-                return {
-                    file: relativePath ? relativePath.toString() : r.fileUri,
-                    matches: r.matches.map(m => ({ line: m.line, text: m.lineText }))
-                };
-            });
+            if (finalResults.length >= maxResults) {
+                return JSON.stringify({
```
eneufeld, line 194: how about we ammend this to the results by wrapping results into a json which contains the message (with info) and a results field with the results?

eneufeld (approved):

PR #15749, opened 2025-06-05
Enable multiSelect for the DebugBreakpointsWidget

xpomul:
I found it impractical that I could only enable/disable/remove all breakpoints at once, or only one by one, but not select multiple ones and enable/disable/remove them in one operation. Therefore, I implemented this functionality.

#### What it does

* Enable multiSelect for breakpoints in the DebugBreakpointsWidget
* Add commands to remove/enable/disable the selected breakpoints

The commands are shown only if the selection is set accordingly:
* enable is only shown if at least one disabled breakpoint is selected
* disable is only shown if at least one enabled breakpoint is selected
* remove is only shown if more than one breakpoint is selected

I have also changed the visibility of the existing menu items (edit, remove single breakpoint), so menu items that only make sense for a single selected item are not visible in case multiple items are selected.

#### How to test

Open any source file(s) and place multiple breakpoints (also consider adding breakpoints of multiple types, such as line breakpoints, instruction breakpoints, function breakpoints).

Then go to the debug panel, open the breakpoints widget and select multiple breakpoints (using the usual modifier keys available for multi-selection of items).

In the context menu there should now be additional menu items (Remove/Enable/Disable Selected Breakpoints) that behave as if the same operation would be executed one by one for all the selected items.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-frontend-application-contribution.ts
+++ b/packages/debug/src/browser/debug-frontend-application-contribution.ts
@@ -804,18 +804,18 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
             isEnabled: () => this.breakpointManager.hasBreakpoints()
         });
         registry.registerCommand(DebugCommands.ENABLE_SELECTED_BREAKPOINTS, {
-            execute: () => this.getSelectedBreakpoints().forEach(breakpoint => breakpoint.setEnabled(true)),
-            isVisible: () => this.getSelectedBreakpoints().some(breakpoint => !breakpoint.enabled),
-            isEnabled: () => this.getSelectedBreakpoints().some(breakpoint => !breakpoint.enabled)
+            execute: () => this.selectedBreakpoints.forEach(breakpoint => breakpoint.setEnabled(true)),
+            isVisible: () => this.selectedBreakpoints.some(breakpoint => !breakpoint.enabled),
+            isEnabled: () => this.selectedBreakpoints.some(breakpoint => !breakpoint.enabled)
         });
         registry.registerCommand(DebugCommands.DISABLE_ALL_BREAKPOINTS, {
             execute: () => this.breakpointManager.enableAllBreakpoints(false),
             isEnabled: () => this.breakpointManager.hasBreakpoints()
         });
         registry.registerCommand(DebugCommands.DISABLE_SELECTED_BREAKPOINTS, {
-            execute: () => this.getSelectedBreakpoints().forEach(breakpoint => breakpoint.setEnabled(false)),
-            isVisible: () => this.getSelectedBreakpoints().some(breakpoint => breakpoint.enabled),
-            isEnabled: () => this.getSelectedBreakpoints().some(breakpoint => breakpoint.enabled)
+            execute: () => this.selectedBreakpoints.forEach(breakpoint => breakpoint.setEnabled(false)),
+            isVisible: () => this.selectedBreakpoints.some(breakpoint => breakpoint.enabled),
+            isEnabled: () => this.selectedBreakpoints.some(breakpoint => breakpoint.enabled)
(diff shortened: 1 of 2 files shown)


packages/debug/src/browser/debug-frontend-application-contribution.ts
```diff
@@ -1588,4 +1622,17 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
...
+     */
+    protected getSelectedBreakpoints(): DebugBreakpoint[] {
+        const breakpointsWidget = this.widgetManager.tryGetWidget(DebugBreakpointsWidget.FACTORY_ID) as DebugBreakpointsWidget;
+        if (breakpointsWidget) {
+            const selectedTreeNodes = breakpointsWidget.model.selectedNodes as TreeElementNode[];
+            return selectedTreeNodes.map(node => node.element as DebugBreakpoint);
+        } else {
+            return [];
+        }
+    }
```
martin-fleck-at:
```suggestion
    get selectedBreakpoints(): DebugBreakpoint[] {
        const { breakpoints } = this;
        return breakpoints && breakpoints.model.selectedNodes
            .filter(TreeElementNode.is)
            .map(node => node.element)
            .filter(element => element instanceof DebugBreakpoint) as DebugBreakpoint[] || [];
    }
```

Nit: I'd prefer it if this were implemented the same way all the other "selected" getters are implemented in this class and maybe even move it up to the other getters. We also should make sure to properly test the TreeElementNodes.

packages/debug/src/browser/debug-frontend-application-contribution.ts
```diff
@@ -797,8 +826,8 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
                     await selectedFunctionBreakpoint.open();
                 }
             },
-            isEnabled: () => !!this.selectedBreakpoint || !!this.selectedFunctionBreakpoint,
-            isVisible: () => !!this.selectedBreakpoint || !!this.selectedFunctionBreakpoint
+            isEnabled: () => this.getSelectedBreakpoints().length === 1 && (!!this.selectedBreakpoint || !!this.selectedFunctionBreakpoint),
+            isVisible: () => this.getSelectedBreakpoints().length === 1 && (!!this.selectedBreakpoint || !!this.selectedFunctionBreakpoint)
```
martin-fleck-at: This is also true for the other commands: Do we really need those additional checks? It seems that the original checks should still be enough, what do you think?

martin-fleck-at (changes requested):
Hi @xpomul, thank you for those changes, they surely will make lives a bit simpler! Everything works as expected but I do have a few questions and suggestions about the code. Could you take a look at them?

packages/debug/src/browser/debug-frontend-application-contribution.ts
xpomul: Thanks. Applied.

packages/debug/src/browser/debug-frontend-application-contribution.ts
xpomul: Well, the difference is whether the single-element-related menu items are shown in case of multi-selection.

Without the length check, the menu looks like this:
Bildschirmfoto 2025-06-17 um 18 23 27

and with the length check:

Bildschirmfoto 2025-06-17 um 18 29 52

My feeling is that showing single-element-related menu items for multi-selection (e.g. edit) is not intuitive, as it is not clear which breakpoint the user will actually be editing. So I have kept the length check for now.

Please let me know if you feel differently.

martin-fleck-at: Ah, you are right, my bad!

PR #15753, opened 2025-06-05
fix: remove obsolete menu workaround

### LINKED ISSUE #15694: Empty submenu in the Terminal context menu
### Bug Description:
Hi,
See the attached context menu screenshot, it contains an empty menu item (no sub-items) at the end.

Image

### Additional Information

- Operating System: Mac with Chrome
- Theia Version: Theia 1.62

sdirix:
#### What it does

We previously used a workaround in the browser-menu-plugin to support initially empty menus, which weren’t compatible with Lumino.

Since the menu refactoring in #14676, this workaround is no longer needed. In fact, it now causes issues like empty submenus appearing in context menus. Removing the workaround fixes this regression.

Fixes #15694

#### How to test

- Check the terminal context menu, it should no longer have an empty submenu
- Check the title bar, that it continues working

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/menu/browser-menu-plugin.ts
+++ b/packages/core/src/browser/menu/browser-menu-plugin.ts
@@ -311,13 +311,6 @@ export class DynamicMenuWidget extends MenuWidget {
         while (items[items.length - 1]?.type === 'separator') {
             items.pop();
         }
-        // Add at least one entry to avoid empty menus.
-        // This is needed as Lumino does all kind of checks whether a menu is empty and for example prevents activating it
-        // This item will be cleared once the menu is opened via the next update as we don't have empty main menus
-        // See https://github.com/jupyterlab/lumino/issues/729
-        if (items.length === 0) {
-            items.push({ type: 'separator' });
-        }
         for (const item of items) {
             parent.addItem(item);
         }


tsmaeder:
@sdirix 

> This is needed as Lumino does all kind of checks whether a menu is empty and for example prevents activating it

I'm a bit worried about the "all kinds of checks" if we want to put this in the patch release. Can you make an argument why removing the workaround is safe or how we can test this with good confidence?

sdirix:
Good question @tsmaeder!

The workaround was needed to bypass [this check](https://github.com/jupyterlab/lumino/blame/204a845414f5f55ba4d97f95fed3ed3c3265490d/packages/widgets/src/menubar.ts#L148-L151). Other empty menu checks in Lumino are for example [here](https://github.com/jupyterlab/lumino/blob/8a4c5f59ed53bdfa1cedb0eb5641ba705a176a63/packages/widgets/src/menubar.ts#L435) and [here](https://github.com/jupyterlab/lumino/blob/8a4c5f59ed53bdfa1cedb0eb5641ba705a176a63/packages/widgets/src/menubar.ts#L505).

You can follow this (closed) PR on how the workaround was developed https://github.com/jupyterlab/lumino/pull/733.

In short: Before the menu refactoring, Theia always started with a menu bar with empty menus. The items of the menus were only filled in, when the menu is about to be shown. With Lumino and the added checks linked above, this no longer worked. The menu never became active, was not clickable and the "about-to-show" was never called.

The workaround bypassed all the empty menu checks by always adding at least one item (a separator) into each menu. As we repopulated the menu then anyway, these pseudo items were automatically removed. This worked because we don't have empty top level menus in Theia.

With the menu refactoring our behavior in Theia changed. We no longer have initially empty menus. And the workaround code is now also called for context menus (which it was not before) breaking them by showing menus without items.

For testing I made sure the menu bar still worked after removing the workaround. I could not find a regresssion. So the menu refactoring let's us remove this somewhat ugly Lumino customization, which is great!

PR #15717, opened 2025-06-03
Add ai tools for write through file changes

lucas-koehler:
#### What it does

- Rename the existing change set tool functions and classes to align better with that they do
- Add new tool functions to directly write changes to disk
- Adapt Coder prompts to adapt to the renamed tool function classes

#### How to test

1. Ensure current writing to changesets still works, i.e. changes are proposed. This should be the default with the current prompt
2. Adapt the Coder prompt to use the new write functions. Changes should be written directly to disk with this

You can use this prompt:

<details>
<summary>coder prompt</summary>

```
{{!-- This prompt is licensed under the MIT License (https://opensource.org/license/mit).
Made improvements or adaptations to this prompt template? We'd love for you to share it with the community! Contribute back here:
https://github.com/eclipse-theia/theia/discussions/new?category=prompt-template-contribution --}}
You are an AI assistant integrated into Theia IDE, designed to assist software developers with code tasks. You can interact with the code base and suggest changes.

## Context Retrieval
Use the following functions to interact with the workspace files if you require context:
- **~{getWorkspaceDirectoryStructure}**
- **~{getWorkspaceFileList}**
- **~{getFileContent}**
- **~{searchInWorkspace}**

Remember file locations that are relevant for completing your tasks using **~{context_addFile}**
Only add files that are really relevant to look at later.

## File Validation
Use the following function to retrieve a list of problems in a file if the user requests fixes in a given file: **~{getFileDiagnostics}**

## Propose Code Changes
To propose code changes or any file changes to the user, never print code or new file content in your response.

Instead, for each file you want to propose changes for:
- **Always Retrieve Current Content**: Use getFileContent to get the original content of the target file.
- **View Pending Changes**: Use ~{getProposedFileState} to see the current proposed state of a file, including all pending changes.
- **Change Content**: Use one of these methods to propose changes:
  - ~{writeFileReplacements}: For targeted replacements of specific text sections. Multiple calls will merge changes unless you set the reset parameter to true.
  - ~{writeFileContent}: For complete file rewrites when you need to replace the entire content. 
  - If ~{writeFileReplacements} continuously fails use ~{suggestFileContent}.
  - ~{clearFileChanges}: To clear all pending changes for a file and start fresh.

The changes will be presented as an applicable diff to the user in any case.

## Tasks

The user might want you to execute some task. You can find tasks using ~{listTasks} and execute them using ~{runTask}.

## Additional Context

[...]

PR #15761, opened 2025-06-05
Add async initialization to ChangeSetFileElement

AlexandraBuzila:
#### What it does
Ensures proper initialization of originalContent before operations.
UI actions (open, apply, revert, etc.) could occur before originalContent was assigned, leading to invalid states for the change set element.

This was originally raised as a follow-up to https://github.com/eclipse-theia/theia/pull/15717

#### How to test
- Using the `coder-search-replace` prompt,  trigger the creation of a change set from the chat. Without opening the comparison view for the proposed changes, accept them. Then, use the revert button and observe that reverting works.

- Using `coder-rewrite` with the sample prompt below, trigger the creation of a change set which should automatically be applied. Observe that the state of the `Apply All` button is correct (disabled) and that the revert button is visible and works as expected

<details>
<summary>Sample Prompt</summary>

```

{{!-- This prompt is licensed under the MIT License (https://opensource.org/license/mit).
Made improvements or adaptations to this prompt template? We'd love for you to share it with the community! Contribute back here:
https://github.com/eclipse-theia/theia/discussions/new?category=prompt-template-contribution --}}
You are an AI assistant integrated into Theia IDE, designed to assist software developers with code tasks. You can interact with the code base and suggest changes.

## Context Retrieval
Use the following functions to interact with the workspace files if you require context:
- **~{getWorkspaceDirectoryStructure}**
- **~{getWorkspaceFileList}**
- **~{getFileContent}**
- **~{searchInWorkspace}**

Remember file locations that are relevant for completing your tasks using **~{context_addFile}**
Only add files that are really relevant to look at later.

## File Validation
Use the following function to retrieve a list of problems in a file if the user requests fixes in a given file: **~{getFileDiagnostics}**

## Propose Code Changes

- **Always Retrieve Current Content**: Use getFileContent to get the original content of the target file.
- **View Pending Changes**: Use ~{getProposedFileState} to see the current proposed state of a file, including all pending changes.
- **Change Content**: Use one of these methods to propose changes:
  - ~{writeFileReplacements}: For targeted replacements of specific text sections. Multiple calls will merge changes unless you set the reset parameter to true.
  - ~{writeFileContent}: For complete file rewrites when you need to replace the entire content. 
  - If ~{writeFileReplacements} continuously fails use ~{suggestFileContent}.
  - ~{clearFileChanges}: To clear all pending changes for a file and start fresh.

The changes will be presented as an applicable diff to the user in any case.

## Tasks

The user might want you to execute some task. You can find tasks using ~{listTasks} and execute them using ~{runTask}.

## Additional Context

The following files have been provided for additional context. Some of them may also be referred to by the user (e.g. "this file" or "the attachment"). Always look at the relevant files to understand your task using the function ~{getFileContent}
{{contextFiles}}

## Previously Proposed Changes
You have previously proposed changes for the following files. Some suggestions may have been accepted by the user, while others may still be pending.
{{changeSetSummary}}

{{prompt:project-info}}

{{taskContextSummary}}

```

</details>

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/browser/change-set-file-element.ts
+++ b/packages/ai-chat/src/browser/change-set-file-element.ts
@@ -102,9 +102,7 @@ export class ChangeSetFileElement implements ChangeSetElement {
      * This includes loading the original content from the file system.
      */
     async ensureInitialized(): Promise<void> {
-        if (this._initializationPromise) {
-            await this._initializationPromise;
-        }
+        await this._initializationPromise;
     }
 
     /**
@@ -128,6 +126,10 @@ export class ChangeSetFileElement implements ChangeSetElement {
     protected listenForOriginalFileChanges(): void {
         this.toDispose.push(this.fileService.onDidFilesChange(async event => {
             if (!event.contains(this.uri)) { return; }
+            if (!this._initialized && this._initializationPromise) {
+                // make sure we are initialized
+                await this._initializationPromise;
+            }
             // If we are applied, the tricky thing becomes the question what to revert to; otherwise, what to apply.
             const newContent = await this.changeSetFileService.read(this.uri).catch(() => '');
             this.readOnlyResource.update({ contents: newContent });
@@ -156,8 +158,8 @@ export class ChangeSetFileElement implements ChangeSetElement {
             this.toDispose.push(this._readOnlyResource);
 
             // If not yet initialized, update the resource once initialization completes
-            if (!this._initialized && this._initializationPromise) {
-                this._initializationPromise.then(() => {
+            if (!this._initialized) {
+                this._initializationPromise?.then(() => {
                     this._readOnlyResource?.update({ contents: this._originalContent ?? '' });
                 });
             }
@@ -271,14 +273,7 @@ export class ChangeSetFileElement implements ChangeSetElement {
 
     onShow(): void {
         // Ensure we have the latest state when showing
-        if (!this._initialized && this._initializationPromise) {
-            this._initializationPromise.then(() => {
-                // Update the change resource with the correct target state after initialization
-                this.changeResource.update({ contents: this.targetState, onSave: content => this.writeChanges(content) });
-            });
-        } else {
-            this.changeResource.update({ contents: this.targetState, onSave: content => this.writeChanges(content) });
-        }
+        this.changeResource.update({ contents: this.targetState, onSave: content => this.writeChanges(content) });
     }
 
     async revert(): Promise<void> {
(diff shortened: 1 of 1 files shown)


packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -85,12 +87,38 @@ export class ChangeSetFileElement implements ChangeSetElement {
...
+    }
+
+    /**
+     * Ensures that the element is fully initialized before proceeding.
+     * This includes loading the original content from the file system.
+     */
+    async ensureInitialized(): Promise<void> {
+        if (this._initializationPromise) {
+            await this._initializationPromise;
+        }
```
colin-grant-work: Would there be a reason for this not to be defined?

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -120,10 +148,19 @@ export class ChangeSetFileElement implements ChangeSetElement {
...
             this.toDispose.push(this._readOnlyResource);
-            this.obtainOriginalContent();
-            this.listenForOriginalFileChanges();
+
+            // If not yet initialized, update the resource once initialization completes
+            if (!this._initialized && this._initializationPromise) {
+                this._initializationPromise.then(() => {
+                    this._readOnlyResource?.update({ contents: this._originalContent ?? '' });
+                });
+            }
```
colin-grant-work: We will set the contents to the original value when we load the content without waiting for it here, won't we?

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -214,16 +270,25 @@ export class ChangeSetFileElement implements ChangeSetElement {
...
     onShow(): void {
-        this.changeResource.update({ contents: this.targetState, onSave: content => this.writeChanges(content) });
+        // Ensure we have the latest state when showing
+        if (!this._initialized && this._initializationPromise) {
+            this._initializationPromise.then(() => {
+                // Update the change resource with the correct target state after initialization
+                this.changeResource.update({ contents: this.targetState, onSave: content => this.writeChanges(content) });
+            });
+        } else {
+            this.changeResource.update({ contents: this.targetState, onSave: content => this.writeChanges(content) });
```
colin-grant-work: Do we need to wait for initialization to handle anything related to `changeResource`?

AlexandraBuzila: Unfortunately yes, because `writeChanges` triggers the file system listener from `listenForOriginalFileChanges`, which uses the original content to determine the state of the element. I moved the check directly to listener now, I think it makes more sense to have it there directly.

packages/ai-chat/src/browser/change-set-file-element.ts
AlexandraBuzila: Yes, but only if we were initialized before the readonly resource was accessed. I think this call here still makes sense.

colin-grant-work (approved):

Issue #15664, opened 2025-05-28
Toolbar console errors


sdirix:
### Bug Description:

During regular usage of Theia, the console is flooded with errors like this

```
2025-05-28T05:30:27.061Z root ERROR Error: not a compound menu node: ["plugin_editor/title"]
    at MenuModelRegistry.getMenu (http://localhost:3000/bundle.js:409298:19)
    at TabBarToolbarRegistry.visibleItems (http://localhost:3000/bundle.js:391756:48)
    at TabBarToolbar.updateTarget (http://localhost:3000/bundle.js:392020:62)
    at ToolbarAwareTabBar.updateToolbar (http://localhost:3000/bundle.js:393271:22)
    at ToolbarAwareTabBar.onUpdateRequest (http://localhost:3000/bundle.js:393263:14)
    at ToolbarAwareTabBar.processMessage (http://localhost:3000/bundle.js:11962:22)
    at invokeHandler (http://localhost:3000/bundle.js:9067:21)
    at sendMessage (http://localhost:3000/bundle.js:8821:13)
    at runMessageLoop (http://localhost:3000/bundle.js:9118:17)
    at http://localhost:3000/bundle.js:8795:42
```

After 5 minutes of usage I had 20+ of these entries in the Console

### Steps to Reproduce:

1. Build the Browser example application (I did not even download the vsx extensions)
2. Use the tool a bit, switching between tabs is sufficient to trigger this for me

I have the toolbar visible.

This is my toolbar config, in case it's relevant

```
{
    "items": {
        "left": [
            [
                {
                    "id": "textEditor.commands.go.back",
                    "command": "textEditor.commands.go.back",
                    "icon": "codicon codicon-arrow-left"
                },
                {
                    "id": "textEditor.commands.go.forward",
                    "command": "textEditor.commands.go.forward",
                    "icon": "codicon codicon-arrow-right"
                }
            ],
            [
                {
                    "id": "workbench.action.splitEditorRight",
                    "command": "workbench.action.splitEditor",
                    "icon": "codicon codicon-split-horizontal"
                }
            ]
        ],
        "center": [
            [
                {
                    "id": "theia-sample-toolbar-contribution",
                    "group": "contributed"
                }
            ]
        ],
        "right": [
            [
                {
                    "id": "workbench.action.showCommands",
                    "command": "workbench.action.showCommands",
                    "icon": "codicon codicon-terminal",
                    "tooltip": "Command Palette"
                },
                {
[...]

PR #15764, opened 2025-06-05
Do not attempt to update actions in non-existent menus

cdamus:
#### What it does

The MenuModelRegistry now throws on attempts to request the menu for an ID that is not registered. As at least one of the menus that the DirtyDiffWidget does not exist in typical installations of Theia IDE, this was causing the diffs not to appear. Because the widget doesn't need to update actions in menus that don't exist, it takes a more lenient approach.

Fixes http://github.com/eclipsesource/theia/issues/253

#### How to test

1. Launch Theia IDE with the panoply of default VSX plugins downloaded and installed.
2. Make a change in some file in Git such that the source editor shows diff markers in the gutter.
3. Click on a diff marker.
4. Observe that the quick diff appears in the Dirty Diff Widget in the editor.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/scm/src/browser/dirty-diff/dirty-diff-widget.ts
+++ b/packages/scm/src/browser/dirty-diff/dirty-diff-widget.ts
@@ -16,7 +16,7 @@
 
 import { inject, injectable, postConstruct } from '@theia/core/shared/inversify';
 import { Position, Range } from '@theia/core/shared/vscode-languageserver-protocol';
-import { CommandMenu, CompoundMenuNode, Disposable, Emitter, Event, MenuModelRegistry, MenuPath, URI, nls } from '@theia/core';
+import { CommandMenu, CompoundMenuNode, Disposable, Emitter, Event, ILogger, MenuModelRegistry, MenuPath, URI, nls } from '@theia/core';
 import { codicon } from '@theia/core/lib/browser';
 import { ContextKeyService } from '@theia/core/lib/browser/context-key-service';
 import { MonacoEditor } from '@theia/monaco/lib/browser/monaco-editor';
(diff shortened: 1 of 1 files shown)


colin-grant-work:
@tsmaeder, it looks like the issue is that nothing is registered for `PLUGIN_SCM_CHANGE_TITLE_MENU` so that comes back as undefined. Is it expected that we define the contribution point, but don't register it at all? I don't remember whether that means we'll do it when a plugin needs it, or whether we need to have registered the menu at least before plugins could use it.

cdamus:
> it looks like the issue is that nothing is registered for `PLUGIN_SCM_CHANGE_TITLE_MENU`

I found that also scm_change_title_menu (no plugin) also was absent. In any case, while that may well be the root cause of the problem, this code still needs to be defensive because currently a required behaviour (showing the quick diff) is borked by an unrelated concern (updating actions).

tsmaeder:
There is https://github.com/eclipse-theia/theia/issues/15664

tsmaeder:
@cdamus I dont' think being defensive at a single call site is the right thing to do, here's why: if we consider requesting a well known menu that does not exist a programming error, we should make sure we never do this by throwing an exception. That would require us to register all menus beforehand. The second option is that we do not consider the request a programming error. In that case, we should not throw an exception, but return `CompundMenuNode | undefined` from `getMenu()`. That would ensure that all call sites handle the case.

cdamus:
My concern is simply that the presentation of the diff widget is being impeded by the failure of an — imo — at best tenuously-related computation, that being the updating of some menus that may or may not exist.

It seems that these menus were never previously required to exist in order for quick diff to work, so that is what this PR ensures.

("defensive" is not an accurate term for this; I misstated that)

cdamus:
Obsolete by #15792.

PR #15765, opened 2025-06-05
Improve image support for tool calls

martin-fleck-at:
#### What it does

- Introduce ToolCallResult type for better of known MCP result types
- Add visual playwright MCP server to AppTester (can use screenshots)

#### How to test

- Use a prompt like `@AppTester Please take a screenshot of http://localhost:3000 and tell me what you see.`
- See that if the MCP server is calling `mcp_playwright_browser_screen_capture`
- Expand the call from `mcp_playwright_browser_screen_capture` after it finished to see the screenshot and the textual content provided by the MCP server
- See if the answer from the LLM matches your expectations. Currently Anthropic seems to support images the best, as other APIs do not provide a specific format for objects as results of tool calls.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
@@ -25,6 +25,7 @@ import { ToolConfirmation, ToolConfirmationState } from './tool-confirmation';
 import { ToolConfirmationManager, ToolConfirmationMode } from '@theia/ai-chat/lib/browser/chat-tool-preferences';
 import { ResponseNode } from '../chat-tree-view';
 import { useMarkdownRendering } from './markdown-part-renderer';
+import { ToolCallResult } from '@theia/ai-core';
 
 @injectable()
 export class ToolCallPartRenderer implements ChatResponsePartRenderer<ToolCallChatResponseContent> {
@@ -55,12 +56,12 @@ export class ToolCallPartRenderer implements ChatResponsePartRenderer<ToolCallCh
     }
 
     protected renderResult(response: ToolCallChatResponseContent): ReactNode {
-        if (!response.result) {
+        const result = this.tryParse(response.result);
+        if (!result) {
             return <pre>{JSON.stringify(response, undefined, 2)}</pre>;
         }
-        const result = response.result;
         if (typeof result === 'string') {
-            return <pre>{this.tryPrettyPrintJson(result)}</pre>;
+            return <pre>{JSON.stringify(result, undefined, 2)}</pre>;
         }
         if ('content' in result) {
             return <div className='theia-toolCall-response-content'>
@@ -73,7 +74,7 @@ export class ToolCallPartRenderer implements ChatResponsePartRenderer<ToolCallCh
                         }
                         case 'text': {
                             return <div className='theia-toolCall-text-result'>
-                                <MarkdownRender text={this.tryPrettyPrintJson(content.text)} openerService={this.openerService} />
+                                <MarkdownRender text={content.text} openerService={this.openerService} />
                             </div>;
                         }
                         case 'audio':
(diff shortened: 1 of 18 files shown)


packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
```diff
@@ -163,35 +192,47 @@ const ToolCallContent: React.FC<ToolCallContentProps> = ({ response, confirmatio
...
+            )}
 
             {/* Show confirmation UI when waiting for allow */}
             {confirmationState === 'waiting' && (
-                <ToolConfirmation
-                    response={response}
-                    onAllow={handleAllow}
-                    onDeny={handleDeny}
-                />
+                <span className='theia-toolCall-waiting'>
```
eneufeld, line 217: this css class is not defined, i suggest to remove this span

packages/ai-chat-ui/src/browser/style/index.css
```diff
@@ -821,14 +819,18 @@ div:last-child > .theia-ChatNode {
...
 
-.theia-tool-pending {
+.theia-toolCall-pending {
   color: var(--theia-descriptionForeground);
 }
 
-.theia-tool-denied {
+.theia-toolCall-denied {
   color: var(--theia-errorForeground);
 }
```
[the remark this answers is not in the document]
martin-fleck-at: I re-worked the styling in a follow up.

packages/ai-mcp/src/node/mcp-server.ts
```diff
@@ -143,7 +144,7 @@ export class MCPServer {
             name: toolName,
             arguments: args,
         };
-        return this.client.callTool(params);
+        return this.client.callTool(params) as Promise<CallToolResult>;
```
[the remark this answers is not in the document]
martin-fleck-at: I'll keep the cast as the result can be almost anything anyways.

packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
```diff
@@ -163,35 +192,47 @@ const ToolCallContent: React.FC<ToolCallContentProps> = ({ response, confirmatio
...
                     </details>
-                ) : (
-                    confirmationState === 'allowed' && (
-                        <span>
-                            <Spinner /> {nls.localizeByDefault('Running')} {response.name}
-                        </span>
-                    )
-                )}
-            </h4>
+                </h4>
```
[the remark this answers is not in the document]
martin-fleck-at: I'll adapt the structure a bit to make sure we have valid HTML.

martin-fleck-at:
What I would consider still to be open after this PR:
- Improve OpenAI support. Currently there is no real way to return an image as part of a tool result. Converting it to text does not really seem to yield any useful results. Maybe we could introduce image messages (which are supported) as part of the tool results.
- Improve Google/Gemini support. Similar problem of not a native support for images as part of the tool results.

packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
```diff
@@ -47,7 +52,52 @@ export class ToolCallPartRenderer implements ChatResponsePartRenderer<ToolCallCh
...
             chatId={chatId}
             renderCollapsibleArguments={this.renderCollapsibleArguments.bind(this)}
-            tryPrettyPrintJson={this.tryPrettyPrintJson.bind(this)} />;
+            responseRenderer={this.renderResult.bind(this)} />;
+    }
+
+    protected renderResult(response: ToolCallChatResponseContent): ReactNode {
+        const result = this.tryParse(response.result);
+        if (!result) {
+            return <pre>{JSON.stringify(response, undefined, 2)}</pre>;
```
eneufeld: if result is undefined then response was undefined in the first place. so stringifying it probably won't help much.

packages/ai-mcp/src/node/mcp-server.ts
eneufeld: the only thing i don't understand is, why ts complains in the first place. shouldn't the types match between the method return and this call?

packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
martin-fleck-at: I am using it in the CSS to make the font bold, similar to the other states that we have:
https://github.com/eclipse-theia/theia/pull/15765/files#diff-0ecd014bbc1720277b7d8e99100329ae021f624d91392c0d256c33942de8ec32R825

packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
martin-fleck-at: True, simply returning undefined would be the better option. I'll push a commit to fix that.

packages/ai-mcp/src/node/mcp-server.ts
martin-fleck-at: The call does match the `CompatibilityCallToolResult` type which is basically `CallToolResult` extended with backwards compatibility to a previous protocol version (2024-10-07). In our case, the `resultSchema` parameter ([second one of the callTool method](https://github.com/modelcontextprotocol/typescript-sdk/blob/1878143f1a8ffc3ce5f7acd4dc61cef67b589b4b/src/client/index.ts#L419-L421)) defaults to `CallToolResultSchema` anyways, so for us the result should match CallToolResult but it may not be a given if you change that parameter, I guess.

PR #15776, opened 2025-06-06
fix: do not duplicate task context files in memory

xai:
Fixes https://github.com/eclipsesource/theia/issues/257

#### What it does

Removes a redundant store() of new task summary files to the in-memory storage.

#### How to test

1. Start Theia with existing task contexts and verify if they are shown correctly in the task context quick pick list
2. Trigger creation of a task 
3. Open the task context selector via the "+" button in the AI chat
3. Verify that the newly created task summary appears only once in the quick pick list.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-ide/src/browser/task-context-file-storage-service.ts
+++ b/packages/ai-ide/src/browser/task-context-file-storage-service.ts
@@ -142,7 +142,6 @@ export class TaskContextFileStorageService implements TaskContextStorageService 
             summary.uri = uri;
             await this.fileService.writeFile(uri, BinaryBuffer.fromString(content));
         }
-        this.inMemoryStorage.store(summary);
         this.onDidChangeEmitter.fire();
     }
 


ndoschek:
@AlexandraBuzila could you please have a look, TIA

AlexandraBuzila:
@xai LGTM and works as expected, thanks! :tada: 
There's a conflict with the main branch, though.

colin-grant-work:
The idea behind using the in-memory store even when the file store was active was to avoid having to reread the disk every time a summary is activated. Would it be possible to identify the mechanism that is causing the apparent duplication - presumably use of different ID's for the same summary? - and ensure that that doesn't occur, but we still get the advantages of keeping the summaries in memory? The code in lines 128-131 is intended to prevent duplication.

Or @JonasHelming, are we ok with reading the disk for every request, in which case, we likely just want to remove the in-memory store completely when the disk-based store is active? The main downside there is that we won't have on hand a list of things for which we know summaries - we'd have to check before populating the quick pick, etc.

packages/ai-ide/src/browser/task-context-file-storage-service.ts
```diff
@@ -142,7 +142,6 @@ export class TaskContextFileStorageService implements TaskContextStorageService
             summary.uri = uri;
             await this.fileService.writeFile(uri, BinaryBuffer.fromString(content));
         }
-        this.inMemoryStorage.store(summary);
         this.onDidChangeEmitter.fire();
```
colin-grant-work: If we do remove this `store` call, then we should remove the emitter firing - it will fire when the file watchers pick up this summary.

xai:
> The idea behind using the in-memory store even when the file store was active was to avoid having to reread the disk every time a summary is activated. Would it be possible to identify the mechanism that is causing the apparent duplication - presumably use of different ID's for the same summary? - and ensure that that doesn't occur, but we still get the advantages of keeping the summaries in memory? The code in lines 128-131 is intended to prevent duplication.

ACK, this requires some debugging then!

xai:
@colin-grant-work Just pushed another try by extending the existing deduplication logic instead of removing the store().
Edit: doesn't work yet, please ignore the retriggered review-request :(

PR #15777, opened 2025-06-06
fix: pin the last mentioned agent, rather than the first one

CamilleLetavernier:
#### What it does

- When explicitly mentioning an Agent, pin it to the Chat (if the corresponding option is enabled)
  - This now happens even if another Agent is already pinned: the last agent mentioned becomes the pinned one

#### How to test

- Start a multi-turn conversation, mentioning multiple agents each turn. For example:

```
@Architect Hello there
```
```
How are you today?
```
Agent: Architect
```
@Coder Hi!
```
```
What's up?
```
Agent: Coder

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-service.ts
+++ b/packages/ai-chat/src/common/chat-service.ts
@@ -313,14 +313,39 @@ export class ChatServiceImpl implements ChatService {
 
     protected getAgent(parsedRequest: ParsedChatRequest, session: ChatSession): ChatAgent | undefined {
         let agent = this.initialAgentSelection(parsedRequest);
-        if (this.pinChatAgent === false) {
+        if (!this.isPinChatAgentEnabled()) {
             return agent;
         }
-        if (!session.pinnedAgent && agent && agent.id !== this.defaultChatAgentId?.id) {
+        
+        return this.applyPinnedAgentLogic(parsedRequest, session, agent);
+    }
+
+    /**
+     * Determines if chat agent pinning is enabled.
+     * Can be overridden by subclasses to provide different logic (e.g., using preferences).
+     */
+    protected isPinChatAgentEnabled(): boolean {
+        return this.pinChatAgent !== false;
+    }
+
+    /**
+     * Applies the pinned agent logic to determine which agent to use.
+     * This centralizes the pinned agent behavior to avoid duplication.
+     */
+    protected applyPinnedAgentLogic(parsedRequest: ParsedChatRequest, session: ChatSession, agent: ChatAgent | undefined): ChatAgent | undefined {
+        const mentionedAgent = this.getMentionedAgent(parsedRequest);
+        
+        if (mentionedAgent !== undefined && agent) {
+            // If an agent is explicitly mentioned, it becomes the new pinned agent
             session.pinnedAgent = agent;
-        } else if (session.pinnedAgent && this.getMentionedAgent(parsedRequest) === undefined) {
+        } else if (!session.pinnedAgent && agent && agent.id !== this.defaultChatAgentId?.id) {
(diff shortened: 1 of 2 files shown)


planger:
I guess it is a matter of taste, but this was kind of on purpose.
The assumption is that you start a conversation with the main agent in the beginning and then have side discussions with other agents one-shot. If you want to change, you can still manually unpin the agent.

If we always switch, as proposed by this PR, depending on the use case, one could argue that we eventually need more explicit mentions of the agent. But I'm not against re-discussing what's the best approach.

packages/ai-chat/src/common/chat-service.ts
```diff
@@ -313,14 +313,39 @@ export class ChatServiceImpl implements ChatService {
 
     protected getAgent(parsedRequest: ParsedChatRequest, session: ChatSession): ChatAgent | undefined {
         let agent = this.initialAgentSelection(parsedRequest);
-        if (this.pinChatAgent === false) {
+        if (!this.isPinChatAgentEnabled()) {
             return agent;
         }
-        if (!session.pinnedAgent && agent && agent.id !== this.defaultChatAgentId?.id) {
+        
+        return this.applyPinnedAgentLogic(parsedRequest, session, agent);
```
colin-grant-work: The `Logic` suffix is a bit odd.

packages/ai-chat/src/common/chat-service.ts
```diff
@@ -313,14 +313,39 @@ export class ChatServiceImpl implements ChatService {
...
+    }
+
+    /**
+     * Applies the pinned agent logic to determine which agent to use.
+     * This centralizes the pinned agent behavior to avoid duplication.
+     */
+    protected applyPinnedAgentLogic(parsedRequest: ParsedChatRequest, session: ChatSession, agent: ChatAgent | undefined): ChatAgent | undefined {
+        const mentionedAgent = this.getMentionedAgent(parsedRequest);
+        
+        if (mentionedAgent !== undefined && agent) {
```
colin-grant-work: This looks odd - it appears to be based on the assumption that the `agent` passed in will actually be the same as `mentionedAgent`, which it will in the current call chain, but it isn't guaranteed to be so.

ndoschek:
@planger Sounds good to me as well.
To balance both approaches, what if we add an “auto-switch to new agents” setting that’s on by default, but still lets you pin agents manually?

@CamilleLetavernier, could you please mark this PR as a draft until we settle on the final UX? Once we’ve agreed on the approach, we can finalize it.

planger:
I don't want to over-complicate things. I just wanted to share the original thinking behind the pinning as it is right now.
Essentially it is: Pinned agent is the default agent (activated if you don't mention one). If you mention an agent and there is a pinned agent, no effect. If you mention one, and there is no pinned agent, it'll become the new pinned agent.

If you find the pinning behavior counter intuitive, I'm completely fine if that's changed and leave this up to you. :-)

ndoschek:
Hi again, I quickly discussed this with Jonas today as well, and we would prefer the approach initially proposed.
You are right, introducing a setting might be a bit too much for now. We can still do this later if needed.

@CamilleLetavernier, could you please have a quick look at Colin's comments so we can get this merged? Thanks in advance!

CamilleLetavernier:
Thanks @colin-grant-work . I've updated the PR. While testing the different code paths, I also noticed further issues with pinning, so I opened a follow-up ticket: https://github.com/eclipse-theia/theia/issues/15853 (I think these are relatively minor, as they only affect cases where Pin Agent is disabled but we still try to pin Agents; but the current behavior is definitely confusing).

colin-grant-work (approved):
The new logic looks good to me. 👍

PR #15782, opened 2025-06-06
Add AI model aliases and default models

lucas-koehler:
#### What it does

Initial implementation of #15395

Introduces support for AI model aliases to facilitate updating many agents' used models at once.
An alias either points to a single model selected by the user or a chain of default/fallback models to consider in priority order.
For the fallback chain, the first ready model (e.g. has an API key configured) is used. For this the language model was extended with a status to be set by its manager.

- Introduces a language model alias registry for registering and retrieving model aliases and chains
- Adds support for chains of default model aliases (`code`, `universal`, `code completion`, `summarize`)
- Selects the first available model (with an API key) from the chain when resolving an alias; falls back to the first in the chain if none are ready
- Enables users to point an alias to use a specific model
- Extends language models to track readiness (e.g., whether an API key is set)
- Adds a new Model Aliases tab to the the AI configuration UI to:
  - Display alias chains
  - Select a mode for an alias

#### How to test

- Open AI configuration view -> agents configuration
- Select an agent (e.g. Coder)
- Open the language model dropdown. Observe:
  - there are 4 aliases at the top of the dropdown
  - models you don't have an API key are not selectable and show a tooltip when hovering over them
- Select the "default/code" alias
- Go to the new Model Aliases tab in the AI configuration view
- Have an anthropic key configured
- open a chat and request something
- observe: tokens for claude 3.7 are used
- remove api key for anthropic and have a openai key configured
- do another request
- observe: tokens for gpt 4.1 are used
- select a model explicitly for the default/code alias (e.g. gpt-4.1-nano)
- do another request and observe tokens for the selected model are used

#### Follow-ups

- ~~Render Agents using an alias in alias view (@sgraband has a WIP implementation)~~
- ~~Show resolved model for an alias in alias config view and in agent config  view~~
- ~~Display in chain the ones that are ready and the one that is chosen (only if applicable)~~
- Configure aliases as default models for agents
- Define fallbackchains for aliases
- Better display that a model is not ready just disabled for now
- Allow to add aliases

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-anthropic/src/node/anthropic-language-model.ts
+++ b/packages/ai-anthropic/src/node/anthropic-language-model.ts
@@ -27,7 +27,8 @@ import {
     UserRequest,
     ImageContent,
     ToolCallResult,
-    ImageMimeType
+    ImageMimeType,
+    LanguageModelStatus
 } from '@theia/ai-core';
 import { CancellationToken, isArray } from '@theia/core';
 import { Anthropic } from '@anthropic-ai/sdk';
@@ -164,6 +165,7 @@ export class AnthropicModel implements LanguageModel {
     constructor(
         public readonly id: string,
         public model: string,
+        public status: LanguageModelStatus,
         public enableStreaming: boolean,
         public useCaching: boolean,
         public apiKey: () => string | undefined,
--- a/packages/ai-core/src/browser/ai-core-preferences.ts
+++ b/packages/ai-core/src/browser/ai-core-preferences.ts
@@ -31,6 +31,8 @@ export const PREFERENCE_NAME_REQUEST_SETTINGS = 'ai-features.modelSettings.reque
 export const PREFERENCE_NAME_MAX_RETRIES = 'ai-features.modelSettings.maxRetries';
 export const PREFERENCE_NAME_DEFAULT_NOTIFICATION_TYPE = 'ai-features.notifications.default';
 
+export const LANGUAGE_MODEL_ALIASES_PREFERENCE = 'ai-features.languageModelAliases';
+
 export const aiCorePreferenceSchema: PreferenceSchema = {
     type: 'object',
     properties: {
@@ -148,6 +150,30 @@ export const aiCorePreferenceSchema: PreferenceSchema = {
             type: 'string',
             enum: [...NOTIFICATION_TYPES],
             default: NOTIFICATION_TYPE_OFF
+        },
+        [LANGUAGE_MODEL_ALIASES_PREFERENCE]: {
+            title: nls.localize('theia/ai/core/preference/languageModelAliases/title', 'Language Model Aliases'),
+            markdownDescription: nls.localize('theia/ai/core/preference/languageModelAliases/description', 'Configure models for each language model alias in the \
+[AI Configuration View]({0}). Alternatiely you can set the settings manually in the settings.json: \n\
+```\n\
+"default/code": {\n\
+  "selectedModel": "anthropic/claude-opus-4-20250514"\n\
+}\n\```',
+                'command:aiConfiguration:open'
+            ),
+            type: 'object',
+            additionalProperties: {
+                type: 'object',
+                properties: {
+                    selectedModel: {
+                        type: 'string',
+                        description: nls.localize('theia/ai/core/preference/languageModelAliases/selectedModel', 'The user-selected model for this alias.')
+                    }
+                },
+                required: ['selectedModel'],
+                additionalProperties: false
+            },
+            default: {},
         }
     }
 };
--- a/packages/ai-core/src/browser/frontend-language-model-alias-registry.ts
+++ b/packages/ai-core/src/browser/frontend-language-model-alias-registry.ts
@@ -1,0 +1,166 @@
+// *****************************************************************************
+// Copyright (C) 2024-2025 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { injectable, inject, postConstruct } from '@theia/core/shared/inversify';
+import { Emitter, Event } from '@theia/core';
+import { LanguageModelAlias, LanguageModelAliasRegistry } from '../common/language-model-alias';
+import { PreferenceScope, PreferenceService } from '@theia/core/lib/browser';
+import { LANGUAGE_MODEL_ALIASES_PREFERENCE } from './ai-core-preferences';
+import { Deferred } from '@theia/core/lib/common/promise-util';
+
+@injectable()
+export class DefaultLanguageModelAliasRegistry implements LanguageModelAliasRegistry {
+
+    protected aliases: LanguageModelAlias[] = [
+        {
+            id: 'default/code',
+            defaultModelIds: [
+                'anthropic/claude-3-7-sonnet-latest',
+                'openai/gpt-4.1',
+                'google/gemini-2.5-pro-exp-03-25'
+            ],
+            description: 'Optimized for code understanding and generation tasks.'
+        },
+        {
+            id: 'default/universal',
+            defaultModelIds: [
+                'anthropic/claude-3-7-sonnet-latest',
+                'openai/gpt-4.1',
+                'google/gemini-2.5-pro-exp-03-25'
+            ],
+            description: 'Well-balanced for both code and general language use.'
+        },
+        {
+            id: 'default/code-completion',
+            defaultModelIds: [
+                'anthropic/claude-3-7-sonnet-latest',
+                'openai/gpt-4.1',
+                'google/gemini-2.5-pro-exp-03-25'
+            ],
+            description: 'Best suited for code autocompletion scenarios.'
+        },
+        {
+            id: 'default/summarize',
+            defaultModelIds: [
+                'anthropic/claude-3-7-sonnet-latest',
+                'openai/gpt-4.1',
+                'google/gemini-2.5-pro-exp-03-25'
+            ],
+            description: 'Models prioritized for summarization and condensation of content.'
+        }
+    ];
+    protected readonly onDidChangeEmitter = new Emitter<void>();
+    readonly onDidChange: Event<void> = this.onDidChangeEmitter.event;
+
+    @inject(PreferenceService)
+    protected readonly preferenceService: PreferenceService;
+
+    protected readonly _ready = new Deferred<void>();
+    get ready(): Promise<void> {
+        return this._ready.promise;
+    }
+
+    @postConstruct()
+    protected init(): void {
+        this.preferenceService.ready.then(() => {
+            this.loadFromPreference();
+            this.preferenceService.onPreferenceChanged(ev => {
+                if (ev.preferenceName === LANGUAGE_MODEL_ALIASES_PREFERENCE) {
+                    this.loadFromPreference();
+                }
+            });
+            this._ready.resolve();
+        }, err => {
+            this._ready.reject(err);
+        });
+    }
+
+    addAlias(alias: LanguageModelAlias): void {
+        const idx = this.aliases.findIndex(a => a.id === alias.id);
+        if (idx !== -1) {
+            this.aliases[idx] = alias;
+        } else {
+            this.aliases.push(alias);
+        }
+        this.saveToPreference();
+        this.onDidChangeEmitter.fire();
+    }
+
+    removeAlias(id: string): void {
+        const idx = this.aliases.findIndex(a => a.id === id);
+        if (idx !== -1) {
+            this.aliases.splice(idx, 1);
+            this.saveToPreference();
+            this.onDidChangeEmitter.fire();
+        }
+    }
+
+    getAliases(): LanguageModelAlias[] {
+        return [...this.aliases];
+    }
+
+    resolveAlias(id: string): string[] | undefined {
+        const alias = this.aliases.find(a => a.id === id);
+        if (!alias) {
+            return undefined;
+        }
+        if (alias.selectedModelId) {
+            return [alias.selectedModelId];
+        }
+        return alias.defaultModelIds;
+    }
+
+    /**
+     * Set the selected model for the given alias id.
+     * Updates the alias' selectedModelId to the given modelId, persists, and fires onDidChange.
+     */
+    selectModelForAlias(aliasId: string, modelId: string): void {
+        const alias = this.aliases.find(a => a.id === aliasId);
+        if (alias) {
+            alias.selectedModelId = modelId;
+            this.saveToPreference();
+            this.onDidChangeEmitter.fire();
+        }
+    }
+
+    /**
+     * Load aliases from the persisted setting
+     */
+    protected loadFromPreference(): void {
+        const stored = this.preferenceService.get<{ [name: string]: { selectedModel: string } }>(LANGUAGE_MODEL_ALIASES_PREFERENCE) || {};
+        this.aliases.forEach(alias => {
+            if (stored[alias.id] && stored[alias.id].selectedModel) {
+                alias.selectedModelId = stored[alias.id].selectedModel;
+            } else {
+                delete alias.selectedModelId;
+            }
+        });
+    }
+
+    /**
+     * Persist the current aliases and their selected models to the setting
+     */
+    protected saveToPreference(): void {
+        const map: { [name: string]: { selectedModel: string } } = {};
+        for (const alias of this.aliases) {
+            if (alias.selectedModelId) {
+                map[alias.id] = { selectedModel: alias.selectedModelId };
+            }
+        }
+        this.preferenceService.set(LANGUAGE_MODEL_ALIASES_PREFERENCE, map, PreferenceScope.User);
+    }
+}
+
--- a/packages/ai-core/src/browser/frontend-language-model-alias-registry.ts
+++ b/packages/ai-core/src/browser/frontend-language-model-alias-registry.ts
@@ -37,8 +37,8 @@ export class DefaultLanguageModelAliasRegistry implements LanguageModelAliasRegi
         {
             id: 'default/universal',
             defaultModelIds: [
+                'openai/gpt-4o',
                 'anthropic/claude-3-7-sonnet-latest',
-                'openai/gpt-4.1',
                 'google/gemini-2.5-pro-exp-03-25'
             ],
             description: 'Well-balanced for both code and general language use.'
@@ -46,8 +46,8 @@ export class DefaultLanguageModelAliasRegistry implements LanguageModelAliasRegi
         {
             id: 'default/code-completion',
             defaultModelIds: [
-                'anthropic/claude-3-7-sonnet-latest',
                 'openai/gpt-4.1',
+                'anthropic/claude-3-7-sonnet-latest',
                 'google/gemini-2.5-pro-exp-03-25'
             ],
             description: 'Best suited for code autocompletion scenarios.'
@@ -55,8 +55,8 @@ export class DefaultLanguageModelAliasRegistry implements LanguageModelAliasRegi
         {
             id: 'default/summarize',
             defaultModelIds: [
-                'anthropic/claude-3-7-sonnet-latest',
                 'openai/gpt-4.1',
+                'anthropic/claude-3-7-sonnet-latest',
                 'google/gemini-2.5-pro-exp-03-25'
             ],
             description: 'Models prioritized for summarization and condensation of content.'
--- a/packages/ai-core/src/common/language-model-alias.ts
+++ b/packages/ai-core/src/common/language-model-alias.ts
@@ -1,0 +1,76 @@
+// *****************************************************************************
+// Copyright (C) 2024-2025 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { Event } from '@theia/core';
+
+/**
+ * Represents an alias for a language model, allowing fallback and selection.
+ */
+export interface LanguageModelAlias {
+    /**
+     * The unique identifier for the alias.
+     */
+    id: string;
+    /**
+     * The list of default model IDs to use if no selectedModelId is set.
+     * Ordered by priority. The first entry also serves as fallback.
+     */
+    defaultModelIds: string[];
+    /**
+     * A human-readable description of the alias.
+     */
+    description?: string;
+    /**
+     * The currently selected model ID, if any.
+     */
+    selectedModelId?: string;
+}
+
+export const LanguageModelAliasRegistry = Symbol('LanguageModelAliasRegistry');
+/**
+ * Registry for managing language model aliases.
+ */
+export interface LanguageModelAliasRegistry {
+    /**
+     * Promise that resolves when the registry is ready for use (preferences loaded).
+     */
+    ready: Promise<void>;
+
+    /**
+     * Event that is fired when the alias list changes.
+     */
+    onDidChange: Event<void>;
+    /**
+     * Add a new alias or update an existing one.
+     */
+    addAlias(alias: LanguageModelAlias): void;
+    /**
+     * Remove an alias by its id.
+     */
+    removeAlias(id: string): void;
+    /**
+     * Get all aliases.
+     */
+    getAliases(): LanguageModelAlias[];
+    /**
+     * Resolve an alias or model id to a prioritized list of model ids.
+     * If the id is not an alias, returns [id].
+     * If the alias exists and has a selectedModelId, returns [selectedModelId].
+     * If the alias exists and has no selectedModelId, returns defaultModelIds.
+     * If the alias does not exist, returns undefined.
+     */
+    resolveAlias(id: string): string[] | undefined;
+}
--- a/packages/ai-core/src/common/language-model.ts
+++ b/packages/ai-core/src/common/language-model.ts
@@ -295,12 +295,18 @@ export interface LanguageModelMetaData {
     readonly family?: string;
     readonly maxInputTokens?: number;
     readonly maxOutputTokens?: number;
+    readonly status: LanguageModelStatus;
 }
 
 export namespace LanguageModelMetaData {
     export function is(arg: unknown): arg is LanguageModelMetaData {
         return isObject(arg) && 'id' in arg;
     }
+}
+
+export interface LanguageModelStatus {
+    status: 'ready' | 'unavailable';
+    message?: string;
 }
 
 export interface LanguageModel extends LanguageModelMetaData {
@@ -331,6 +337,10 @@ export interface LanguageModelSelector extends VsCodeLanguageModelSelector {
 export type LanguageModelRequirement = Omit<LanguageModelSelector, 'agent'>;
 
 export const LanguageModelRegistry = Symbol('LanguageModelRegistry');
+
+/**
+ * Base interface for language model registries (frontend and backend).
+ */
 export interface LanguageModelRegistry {
     onChange: Event<{ models: LanguageModel[] }>;
     addLanguageModels(models: LanguageModel[]): void;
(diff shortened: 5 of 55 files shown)


packages/ai-anthropic/src/node/anthropic-language-model.ts
```diff
@@ -172,6 +174,10 @@ export class AnthropicModel implements LanguageModel {
         protected readonly tokenUsageService?: TokenUsageService
     ) { }
 
+    static calculateStatus(apiKey: string | undefined): LanguageModelStatus {
+        return apiKey ? { status: 'ready' } : { status: 'unavailable', message: 'No Anthropic API key set' };
+    }
+
```
sdirix: This is unused. The `AnthropicLanguageModelsManagerImpl` calculates the status itself.

packages/ai-core/src/browser/frontend-language-model-alias-registry.ts
```diff
@@ -0,0 +1,154 @@
...
+            this.saveToPreference();
+            this.onDidChangeEmitter.fire();
+        }
+    }
+
+    /**
+     * Load aliases from the persisted setting
+     */
+    protected loadFromPreference(): void {
+        const stored = this.preferenceService.get<{ [name: string]: { selectedModel: string } }>(LANGUAGE_MODEL_ALIASES_PREFERENCE) || {};
```
sdirix, line 143: We need to wait until the preferenceService is ready. Otherwise we will only load the defaults instead of user customized data if we are constructed too early. Similarly we should either convert all our methods to async or we add a Promise which callers can await to be sure to resolve the aliases correctly, similar to the ready promise of the preferenceService.

packages/ai-core/src/common/language-model.ts
```diff
@@ -339,6 +346,16 @@ export interface LanguageModelRegistry {
...
+    /**
+     * Patch a language model by id, updating only the provided fields and firing the change event.
+     * If the model does not exist, logs a warning and returns.
+     */
+    patchLanguageModel<T extends LanguageModel = LanguageModel>(id: string, patch: Partial<T>): Promise<void>;
+    /**
+     * Returns the first model with status "ready" for a given identifier, or the first found model if none are ready.
+     * If the identifier is an alias, finds the highest-priority available model from that alias.
+     */
+    getLanguageModelForIdentifier(identifier: string): Promise<LanguageModel | undefined>;
```
sdirix: The naming should be improved. Just looking at the interface, what is the difference between `getLanguageModel(id)` and `getLanguageModelForIdentifier(id)`. It's a bit clunky but maybe `getReadyLanguageModel(idOrAlias: string)`, or reuse the `select` naming as there is some "calculation", e.g. `selectReadyLanguageModel(idOrAlias)`.

packages/ai-core/src/common/language-model-alias.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Represents an alias for a language model, allowing fallback and selection.
+ */
+export interface LanguageModelAlias {
+    /**
+     * The unique identifier for the alias.
+     */
+    id: string;
+    /**
+     * The list of default model IDs to use if no selectedModelId is set.
+     * Ordered by priority: the first entry is the highest priority fallback.
```
sdirix:
```suggestion
     * Ordered by priority. The first entry also serves as fallback.
```

packages/ai-core/src/common/language-model.ts
```diff
@@ -414,6 +434,38 @@ export class DefaultLanguageModelRegistryImpl implements LanguageModelRegistry {
...
+                const model = await this.getLanguageModel(modelId);
+                if (model?.status.status === 'ready') {
+                    return model;
+                }
+            }
+
+            // If no ready model was found, return the first model referenced by the alias
+            if (modelIds.length > 0) {
+                return this.getLanguageModel(modelIds[0]);
+            }
```
sdirix: Personally I would return `undefined` here. I don't think that there is a great use case for using an "unusable" language model. If `undefined` is returned instead, I can at least react to that in the code and for example show an alternative UI. When we keep it like this, every caller will need to inspect the returned language model to see whether it's actually ready to be used.

packages/ai-ide/src/browser/ai-configuration/language-model-renderer.tsx
```diff
@@ -23,17 +24,20 @@ export interface LanguageModelSettingsProps {
...
-    { agent, languageModels, aiSettingsService, languageModelRegistry }) => {
+    { agent, languageModels, aiSettingsService, languageModelRegistry, languageModelAliases: aliases }) => {
 
     const findLanguageModelRequirement = async (purpose: string): Promise<LanguageModelRequirement | undefined> => {
         const requirementSetting = await aiSettingsService.getAgentSettings(agent.id);
         return requirementSetting?.languageModelRequirements?.find(e => e.purpose === purpose);
     };
 
     const [lmRequirementMap, setLmRequirementMap] = React.useState<Record<string, LanguageModelRequirement>>({});
+    // State to hold resolved model for each alias identifier
```
sdirix:
```suggestion
```

packages/ai-ide/src/browser/ai-configuration/language-model-renderer.tsx
```diff
@@ -54,6 +58,21 @@ export const LanguageModelRenderer: React.FC<LanguageModelSettingsProps> = (
...
+            await Promise.all(Object.values(lmRequirementMap).map(async requirements => {
+                const id = requirements.identifier;
+                if (id && aliases.some(a => a.id === id)) {
+                    newResolved[id] = await languageModelRegistry.getLanguageModelForIdentifier(id);
+                }
+            }));
+            setResolvedAliasModels(newResolved);
+        };
+        resolveAliases();
+    }, [JSON.stringify(lmRequirementMap), JSON.stringify(aliases)]);
```
sdirix: Why `JSON.stringify`? The `lmRequirementMap` and the `aliases` are always newly set. If we want to optimize here, then we should rather not set a new `lmRequirementMap` or hand over a new `aliases` when they do not contain any changes.

packages/ai-ide/src/browser/ai-configuration/language-model-renderer.tsx
```diff
@@ -54,6 +58,21 @@ export const LanguageModelRenderer: React.FC<LanguageModelSettingsProps> = (
         computeLmRequirementMap();
     }, []);
 
+    // Effect to resolve alias to model whenever requirements.identifier changes and is an alias
```
sdirix:
```suggestion
    // Effect to resolve alias to model whenever requirements.identifier or aliases change
```

packages/ai-ide/src/browser/ai-configuration/language-model-renderer.tsx
```diff
@@ -88,35 +107,73 @@ export const LanguageModelRenderer: React.FC<LanguageModelSettingsProps> = (
...
-                        </select>
-                    </>
-                    <hr />
-                </div>
-            </React.Fragment>
-        ))}
-
+        {Object.values(lmRequirementMap).map((requirements, index) => {
+            const isAlias = requirements.identifier && aliases.some(a => a.id === requirements.identifier);
+            const resolvedModel = isAlias ? resolvedAliasModels[requirements.identifier!] : undefined;
```
sdirix: Is the force access necessary? We already check in the previous line that the identifier exists.

packages/ai-ide/src/browser/ai-configuration/language-model-renderer.tsx
```diff
@@ -88,35 +107,73 @@ export const LanguageModelRenderer: React.FC<LanguageModelSettingsProps> = (
...
+                                    <span style={{ marginLeft: 8 }}>
+                                        {resolvedModel.name ?? resolvedModel.id}
+                                        {resolvedModel.status.status === 'ready' ? (
+                                            <span style={{ color: 'green', marginLeft: 6 }} title="Ready">✓</span>
+                                        ) : (
+                                            <span style={{ color: 'red', marginLeft: 6 }} title={resolvedModel.status.message || 'Not ready'}>✗</span>
+                                        )}
+                                    </span>
+                                ) : (
+                                    <span style={{ marginLeft: 8, color: 'var(--theia-descriptionForeground)' }}>
```
sdirix: There are a lot of inline styles.. We should move them to CSS so they can be customized.

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
```diff
@@ -0,0 +1,267 @@
...
+    @postConstruct()
+    protected init(): void {
+        this.id = ModelAliasesConfigurationWidget.ID;
+        this.title.label = ModelAliasesConfigurationWidget.LABEL;
+        this.title.closable = false;
+
+        this.loadAliases();
+        this.loadLanguageModels();
+        this.loadMatchingAgentIdsForAllAliases();
+        this.update();
```
sdirix: We should execute the `update` once all the loading is completed (maybe also once before like now)

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
```diff
@@ -0,0 +1,267 @@
...
+            ...alias,
+            selectedModelId: newModelId
+        };
+        this.languageModelAliasRegistry.addAlias(updatedAlias);
+    };
+
+    render(): React.ReactNode {
+        const selectedAliasId = this.aiConfigurationSelectionService.getSelectedAliasId();
+        const selectedAlias = this.aliases.find(alias => alias.id === selectedAliasId);
+        // this.matchingAgentIdsForAlias is now available for use in the UI if needed
```
sdirix:
```suggestion
```

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
```diff
@@ -0,0 +1,267 @@
...
+                        {nls.localize(
+                            'theia/ai/core/modelAliasesConfiguration/defaultsHierarchy',
+                            'When no model is explicitly selected, the first available default model will be used.'
+                        )}
+                    </div>
+                </div>
+                <div style={{ marginBottom: 10 }}>
+                    <label style={{ fontWeight: 600 }}>{nls.localize('theia/ai/core/modelAliasesConfiguration/evaluatesTo', 'Evaluates to')}:</label>
+                    {resolvedModel ? (
+                        <span style={{ marginLeft: 8 }}>
```
sdirix: Again we should externalize most of the CSS

packages/ai-openai/src/node/openai-language-models-manager-impl.ts
```diff
@@ -42,6 +42,12 @@ export class OpenAiLanguageModelsManagerImpl implements OpenAiLanguageModelsMana
         return this._apiVersion ?? process.env.OPENAI_API_VERSION;
     }
 
+    protected calculateStatus(effectiveApiKey: string | undefined): LanguageModelStatus {
+        return effectiveApiKey
+            ? { status: 'ready' }
+            : { status: 'unavailable', message: 'No OpenAI API key set' };
+    }
```
sdirix: Custom Open AI models are not necessarily key protected, so users don't need to set keys for them. They will now show up as unavailable.

sdirix (changes requested):
Does not work for me. When testing on latest `master`, Theia AI works as expected. When I start from this PR, then all models are gone.

There are a lot of errors in the console which are likely the cause of this:

```
ERROR Error: You are attempting to construct Symbol(AnthropicLanguageModelsManager) in a synchronous way but it has asynchronous dependencies.
ERROR Error: You are attempting to construct Symbol(LanguageModelRegistryFrontendDelegate) in a synchronous way but it has asynchronous dependencies.
[...]
ERROR Uncaught Exception:  Error: No matching bindings found for serviceIdentifier: Symbol(LanguageModelAliasRegistry)
Trying to resolve bindings for "BackendLanguageModelRegistry"
ERROR Error: No matching bindings found for serviceIdentifier: Symbol(LanguageModelAliasRegistry)
```

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
sgraband: Maybe i am missing something, but isn't this what we are currently doing?

packages/ai-openai/src/node/openai-language-models-manager-impl.ts
sgraband: Do you have an idea on how we could solve this for now?

sgraband:
@sdirix Thanks for the feedback. I incorporated most of it and also fixed the regression you reported. I added comments on two of your suggestions. Could you take another look?

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
sdirix: The `load` methods are async. So we just trigger them and already perform the `update` before they are completed.

packages/ai-openai/src/node/openai-language-models-manager-impl.ts
sdirix: If it's a custom open AI model we always return `ready`. Only for the official ones we return `unavailable` if no API key is set.

Bonus points to set a custom open AI model to `unavailable` in case we do a request and an authorization error is returned. But this can be done in a follow up.

packages/ai-openai/src/browser/openai-frontend-application-contribution.ts
```diff
@@ -108,7 +109,7 @@ export class OpenAiFrontendApplicationContribution implements FrontendApplicatio
         this.manager.createOrUpdateLanguageModels(...this.createCustomModelDescriptionsFromPreferences(customModels));
     }
 
-    protected createOpenAIModelDescription(modelId: string): OpenAiModelDescription {
+    protected createOpenAIModelDescription(modelId: string, customModel = false): OpenAiModelDescription {
```
sdirix: I don't think you need to track a `customModel` property. All OpenAI models with a `url` are custom. Without an url they are the official ones.

sgraband: Good point! Fixed this and rebased on master.

packages/ai-ide/src/browser/ai-configuration/language-model-renderer.tsx
```diff
@@ -88,35 +106,73 @@ export const LanguageModelRenderer: React.FC<LanguageModelSettingsProps> = (
...
+                                {aliases?.sort((a, b) => a.id.localeCompare(b.id)).map(alias => (
+                                    <option key={`alias/${alias.id}`} value={alias.id} className="ai-model-alias-option-bold">{`[alias] ${alias.id}`}</option>
+                                ))}
+                                {languageModels?.sort((a, b) => (a.name ?? a.id).localeCompare(b.name ?? b.id)).map(model => {
+                                    const isNotReady = model.status.status !== 'ready';
+                                    return (
+                                        <option
+                                            key={model.id}
+                                            value={model.id}
+                                            disabled={isNotReady}
```
sdirix: We should not disable "not ready" models in the selection. For example even when my llama-file is not yet running, I still want to be able to configure it for my agents. Maybe I am thinking about switching to the vercel provider when scrolling through the list, why am I forced to first enter an API key for them.

image

It "feels" like a bug when using the dropdown:
 - The "bold" styling is unusual for dropdown items and looks like a CSS styling bug
 - For the user it is not apparent at all why some models can be selected, and why some can not. Once the user figured out that apparently the "normal" models can't be selected, the user is stuck. The user does not know how they can make the models selectable.

Instead of disabling the items and using a bold font, I would like to suggest to decorate the select items with a checkmark or some "ready" marker, similar to the alias tab:

image

packages/ai-ide/src/browser/ai-configuration/language-model-renderer.tsx
```diff
@@ -88,35 +106,73 @@ export const LanguageModelRenderer: React.FC<LanguageModelSettingsProps> = (
...
-                </div>
-            </React.Fragment>
-        ))}
-
+        {Object.values(lmRequirementMap).map((requirements, index) => {
+            const isAlias = requirements.identifier && aliases.some(a => a.id === requirements.identifier);
+            const resolvedModel = isAlias ? resolvedAliasModels[requirements.identifier] : undefined;
+            return (
+                <React.Fragment key={index}>
+                    <div><strong>{nls.localize('theia/ai/core/languageModelRenderer/purpose', 'Purpose')}:</strong></div>
```
sdirix: When selecting an alias, the purpose becomes empty and there is no distance between `Purpose` and `LANGUAGE MODEL`

image

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
```diff
@@ -0,0 +1,272 @@
...
+                                        title={isNotReady && model.status.message ? model.status.message : undefined}
+                                    >
+                                        {model.name ?? model.id}
+                                    </option>
+                                );
+                            }
+                            )}
+                    </select>
+                </div>
+                <div className="ai-alias-detail-defaults">
```
sdirix: The default model information is not relevant when the user already selected a model and just adds visual noise.

Either this should be a rankable list of models and the user can just add/remove models in this list as they see fit without an "override" select, or at least we remove all the unnecessary information when the user used an override.

image

packages/ai-core/src/browser/ai-core-preferences.ts
```diff
@@ -148,6 +150,23 @@ export const aiCorePreferenceSchema: PreferenceSchema = {
             type: 'string',
             enum: [...NOTIFICATION_TYPES],
             default: NOTIFICATION_TYPE_OFF
+        },
+        [LANGUAGE_MODEL_ALIASES_PREFERENCE]: {
```
sdirix, line 154: This preference is hard to find as there is no section for it:

image

packages/ai-core/src/browser/ai-core-preferences.ts
```diff
@@ -148,6 +150,23 @@ export const aiCorePreferenceSchema: PreferenceSchema = {
...
+            title: nls.localize('theia/ai/core/preference/languageModelAliases/title', 'Language Model Aliases'),
+            description: nls.localize('theia/ai/core/preference/languageModelAliases/description', 'Store selected models for each language model alias.'),
```
sdirix: The "store" is not the important part

```suggestion
            description: nls.localize('theia/ai/core/preference/languageModelAliases/description', 'Configure models for each language model alias.'),
```

sdirix: Whenever a user is forced to edit the settings.json we explain in the description how the configuration should look like.

packages/ai-core/src/browser/ai-core-preferences.ts
```diff
@@ -148,6 +150,23 @@ export const aiCorePreferenceSchema: PreferenceSchema = {
...
+            type: 'object',
+            additionalProperties: {
+                type: 'object',
+                properties: {
+                    selectedModel: {
```
sdirix, line 168: Why can we only configure the `selectedModel`. I would also like to configure the `defaultModelIds` as well as the `description`. Then we can add arbitrary model aliases with our own ranked priority list and modify the priority list of the built-ins.

packages/ai-core/src/browser/frontend-language-model-alias-registry.ts
```diff
@@ -0,0 +1,166 @@
...
+            alias.selectedModelId = modelId;
+            this.saveToPreference();
+            this.onDidChangeEmitter.fire();
+        }
+    }
+
+    /**
+     * Load aliases from the persisted setting
+     */
+    protected loadFromPreference(): void {
```
sdirix, line 142: I don't see a reason to not allow to add more aliases and configure the defaults via the preferences too. Should not be many lines to add here.

packages/ai-core/src/browser/frontend-language-model-alias-registry.ts
```diff
@@ -0,0 +1,166 @@
...
+            return [alias.selectedModelId];
+        }
+        return alias.defaultModelIds;
+    }
+
+    /**
+     * Set the selected model for the given alias id.
+     * Updates the alias' selectedModelId to the given modelId, persists, and fires onDidChange.
+     */
+    selectModelForAlias(aliasId: string, modelId: string): void {
```
sdirix, line 130: Instead of "select" this should rather be named override or similar

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
sdirix: In case of the fallback, the UI can also be streamlined:

image

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
```diff
@@ -0,0 +1,272 @@
...
+                    </ol>
+                    <div className="ai-alias-defaults-hint">
+                        {nls.localize(
+                            'theia/ai/core/modelAliasesConfiguration/defaultsHierarchy',
+                            'When no model is explicitly selected, the first available default model will be used.'
+                        )}
+                    </div>
+                </div>
+                <div className="ai-alias-evaluates-to-container">
+                    <label className="ai-alias-evaluates-to-label">{nls.localize('theia/ai/core/modelAliasesConfiguration/evaluatesTo', 'Evaluates to')}:</label>
```
sdirix: The evaluates to is not reactive

image

It still shows that it's available although it no longer is. It's only updated once I rerender the tab, e.g. by clicking another agent and coming back

sdirix: I used `openai/o3` as the "selected"/"override" model. It still should show that it evaluates to that model, but with the error marker. Even in the case of the "priority order", it should also show an error marker next to the "No model ready" text.

image

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
sdirix: In case we add default model editing, please add a link to the preference which allows to modify the priority order list.

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
```diff
@@ -0,0 +1,272 @@
...
+        const agents = this.agentService.getAllAgents().filter(agent => agentIds.includes(agent.id));
+        const resolvedModel = this.resolvedModelForAlias.get(alias.id);
+        return (
+            <div>
+                <div className="settings-section-title settings-section-category-title ai-alias-detail-title">
+                    <span>{alias.id}</span>
+                </div>
+                {alias.description && <div className="ai-alias-detail-description">{alias.description}</div>}
+                <div className="ai-alias-detail-selected-model">
+                    <label>{nls.localize('theia/ai/core/modelAliasesConfiguration/selectedModelId', 'Selected Model')}: </label>
```
sdirix, line 190: The UI / Evaluation does not properly handle the case that the user selected a model previously which is no longer available. The UI shows as if the "priority list" is selected

image

Instead it should look like this

image

sdirix (changes requested):
In general the functionality works. However regarding UX and feature set there are quite a lot of things to be improved.

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
sgraband: Great suggestion. I changed this, but i kept the Evaluates to: as i feel like this is helpful in the priority order case. Objections?

packages/ai-core/src/browser/ai-core-preferences.ts
sgraband: I would prefer to do this in a follow up. (Alias creation and setting the default models)

packages/ai-core/src/browser/frontend-language-model-alias-registry.ts
sgraband: See above (follow up)

packages/ai-ide/src/browser/ai-configuration/language-model-renderer.tsx
sgraband: I am not to familiar with the purpose field. Would you expect the field to go away when an alias is selected or should it still be displayed (without the Language model field).

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
sdirix: Makes sense, but only showing it in the priority order case

packages/ai-ide/src/browser/ai-configuration/language-model-renderer.tsx
sdirix: Should still be shown. It's a bit weird in the current UI. The idea is that the agent asks for a specific kind of LLM, e.g. `chat` or `code` specialized. The user should then select their LLM accordingly.

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
sgraband: Will investigate the first comment, but i disagree (partly) with the second one:

An alias should only evaluate to a model if it is ready, otherwise it evaluates to undefined. Otherwise it is weird, that it returns undefined if no default model is ready, but still the model if i selected an unready model manually imho.
I envision the workflow to be something like this: User sees alias has no model ready -> Go to Model Alias view to investigate it. Because the proble could also be a misconfiguration of the alias and not that the model is unready.
Based on this and for consistency reasons (not having to handle whether or not the evaluation is based on a selected model or the default list, i would prefer to keep it this way). WDYT?

sdirix: That's fine with me :+1:

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
```diff
@@ -0,0 +1,273 @@
...
+                                </span>
+                            ) : (
+                                <span className="ai-alias-evaluates-to-unresolved">
+                                    {nls.localize('theia/ai/core/modelAliasesConfiguration/noResolvedModel', 'No model ready for this alias.')}
+                                </span>
+                            )}
+                        </div></>
+                }
+                <div className="ai-alias-detail-agents">
+                    <label>{nls.localize('theia/ai/core/modelAliasesConfiguration/agents', 'Agents using this Alias')}:</label>
```
sdirix, line 262: There is a weird linebreak:

Image

packages/ai-ide/src/browser/ai-configuration/language-model-renderer.tsx
```diff
@@ -88,35 +79,71 @@ export const LanguageModelRenderer: React.FC<LanguageModelSettingsProps> = (
...
+                        {isAlias && (
+                            <div className="ai-alias-evaluates-to-container">
+                                <label className="ai-alias-evaluates-to-label">{nls.localize('theia/ai/core/modelAliasesConfiguration/evaluatesTo', 'Evaluates to')}:</label>
+                                {resolvedModel ? (
+                                    <span className="ai-alias-evaluates-to-value">
+                                        {resolvedModel.name ?? resolvedModel.id}
+                                        {resolvedModel.status.status === 'ready' ? (
+                                            <span className="ai-model-status-ready" title="Ready">✓</span>
+                                        ) : (
+                                            <span className="ai-model-status-not-ready" title={resolvedModel.status.message || 'Not ready'}>✗</span>
```
sdirix, line 133: This case can never happen, right? It will always resolve to a ready model

packages/ai-ide/src/browser/ai-configuration/language-model-renderer.tsx
```diff
@@ -88,35 +79,71 @@ export const LanguageModelRenderer: React.FC<LanguageModelSettingsProps> = (
...
+                                        )}
+                                    </span>
+                                ) : (
+                                    <span className="ai-alias-evaluates-to-unresolved">
+                                        {nls.localize('theia/ai/core/modelAliasesConfiguration/noResolvedModel', 'No model ready for this alias.')}
+                                    </span>
```
sdirix: I would like this case to also show a red X

packages/ai-ide/src/browser/ai-configuration/model-aliases-configuration-widget.tsx
```diff
@@ -0,0 +1,273 @@
...
+                    {agents.length > 0 ? (
+                        <ul>
+                            {agents.map(agent => (
+                                <li key={agent.id}>
+                                    <span>{agent.name}</span>
+                                    <span className="ai-alias-agent-id">({agent.id})</span>
```
sdirix: I would only show the id if it's different than the name. As id === name for all our agents, it looks more like an error.

packages/ai-ide/src/browser/ai-configuration/language-model-renderer.tsx
sgraband: Applied the other changes. For this one: I would like to keep this in case something breaks in the service the user can actually see here that it resolves to a unready model.

sgraband:
@JonasHelming Could you go over the proposed aliases [here](https://github.com/eclipse-theia/theia/blob/fc13df6016fd1886eae196b301264900f5f8494a/packages/ai-core/src/browser/frontend-language-model-alias-registry.ts#L29) and adjust which ones should be available, which models they should reference and the description. And then change the default model of the agents to a suitable alias and just add this as an commit on this PR?

JonasHelming:
I would go:
 
        {
            id: 'default/code',
            defaultModelIds: [
                'anthropic/claude-3-7-sonnet-latest',
                'openai/gpt-4.1',
                'google/gemini-2.5-pro-exp-03-25'
            ],
            description: 'Optimized for code understanding and generation tasks.'
        },
        {
            id: 'default/universal',
            defaultModelIds: [
                'openai/gpt-4o',
                'anthropic/claude-3-7-sonnet-latest',
                'google/gemini-2.5-pro-exp-03-25'
            ],
            description: 'Well-balanced for both code and general language use.'
        },
        {
            id: 'default/code-completion',
            defaultModelIds: [
                'openai/gpt-4.1',
                'anthropic/claude-3-7-sonnet-latest',
                'google/gemini-2.5-pro-exp-03-25'
            ],
            description: 'Best suited for code autocompletion scenarios.'
        },
        {
            id: 'default/summarize',
            defaultModelIds: [
                'openai/gpt-4.1',
                'anthropic/claude-3-7-sonnet-latest',
                'google/gemini-2.5-pro-exp-03-25'
            ],
            description: 'Models prioritized for summarization and condensation of content.'
        }
    ];

sgraband:
@JonasHelming could you take a look at the assignment between agents and aliases? Best to just delete your `ai-features.agentSettings` block in the settings.json.

@sdirix Could you take a look at the changes of [this commit](https://github.com/eclipse-theia/theia/pull/15782/commits/cc13d537d65c605b22c236c56dd61972c5ab48b1)? This was needed to properly render the agents using this alias if the setting was not set.

sdirix (approved):
I deleted all my ai-feature preferences and the tool worked as expected

JonasHelming:
@sgraband The follow up section above mentions: "Configure aliases as default models for agents" => isn't this done?

sgraband:
Yes. Since i already created a [follow-up](https://github.com/eclipse-theia/theia/issues/16089). I did not update this further.

PR #15783, opened 2025-06-06
Implement a retry mechanism for Google GenAI calls

xpomul:
#### What it does
See https://github.com/eclipse-theia/theia/discussions/15718: 

Google imposes rate limits for its LLMs. Especially with lower tiers (including the free tier - see https://ai.google.dev/gemini-api/docs/rate-limits?hl=de ), it can quickly happen that an agent with tool calls (such as the Coder agent) hits the Requests Per Minute (RPM) rate limit and the agent execution terminates with an error.
Also, in longer conversations, it can happen that the LLM does not return proper JSON which leads to an error in the GenAI API client, or reports 500 Internal Server Errors occasionally.

To make the Google LanguageModel implementation in Theia more robust against these errors, a retry mechanism is now implemented that can resend the last request after a configurable delay in case of an error.

The retry mechanism can be configured using preferences:

* `maxRetriesOnErrors` configures the maximum number of retries per request after which to give up. Defaults to `3`. If smaller than `1`, then the retry logic is disabled.
* `retryDelayOnRateLimitError` configures the delay in seconds to wait in case of a rate limit error. Defaults to `60` (1 minute). If negative, then no retry is attempted and the error is propagated.
* `retryDelayOnOtherErrors` configures the delay in seconds to wait in case of any other error. Defaults to `-1` (disabled). If negative, then no retry is attempted and the error is propagated.

#### How to test

* Get a Google GenAI API key for the free tier.
* Set up Theia with that key and configure the Coder agent to use google/gemini-2.0-flash
* Give a complex instruction to the Coder agent (e.g. with the Theia codebase loaded, ask it something like

```
@Coder Change the Breakpoint widget so that multi-selection of breakpoints is possible.

This means that the TreeWidget of the Breakpoint view should be changed to support multiSelect. If multiple breakpoints are selected, then in the context menu, there should be menu items:

* Enable all selected
* Disable all selected
* Delete all selected

which should behave like the counterpart commands for single breakpoints just applied to multiple ones.
```

Without the change, you should see an message 429 Rate Limit Exceeded after 10 tool calls.
With the change, the conversation will stall for some time in the middle, but eventually continue after 60s of waiting.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-google/src/node/google-language-model.ts
+++ b/packages/ai-google/src/node/google-language-model.ts
@@ -29,26 +29,7 @@ import {
 import { CancellationToken } from '@theia/core';
 import { GoogleGenAI, FunctionCallingConfigMode, FunctionDeclaration, Content, Schema, Part, Modality } from '@google/genai';
 import { wait } from '@theia/core/lib/common/promise-util';
-
-/**
- * Retry Settings in case of errors (rate limit errors etc.)
- */
-export interface RetrySettings {
-    /* Maximum number of retries in case of errors. If smaller than 1, then the retry logic is disabled */
-    maxRetriesOnErrors: number;
-    /* Delay in seconds between retries in case of rate limit errors. See https://ai.google.dev/gemini-api/docs/rate-limits */
-    retryDelayOnRateLimitError: number;
-    /* Delay in seconds between retries in case of other errors (sometimes the Google GenAI reports errors such as incomplete JSON syntax returned from the model
-     * or 500 Internal Server Error). Setting this to -1 prevents retries in these cases. Otherwise a retry happens either immediately (when set to 0) or after
-     * this delay in seconds (if set to a positive number). */
-    retryDelayOnOtherErrors: number;
-}
-
-const DEFAULT_RETRY_SETTINGS: RetrySettings = {
-    maxRetriesOnErrors: 3,
-    retryDelayOnRateLimitError: 60,
-    retryDelayOnOtherErrors: -1
-};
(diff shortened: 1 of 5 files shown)


packages/ai-google/src/node/google-language-model.ts
```diff
@@ -127,7 +148,22 @@ export class GoogleModel implements LanguageModel {
     ) { }
 
     protected getSettings(request: LanguageModelRequest): Readonly<Record<string, unknown>> {
-        return request.settings ?? {};
+        const settings = request.settings ?? {};
+        // Filter out RetrySettings keys to avoid passing them to the API
+        const apiSettings = { ...settings };
+        delete apiSettings.maxRetriesOnErrors;
```
eneufeld: we usually pass these kind of settings using the constructor.
The advantage is that you can expose these settings via a preference (which I think should be done). This allows the user to influence the settings. Currently you have to know the syntax and add these model unrelated settings to the modelsettings object.

xpomul: Ok. Thanks. 
I have changed the settings into preferences and updated the PR description accordingly.

eneufeld:
@sdirix any opinion on this? I think the change is fine to be merged.

sdirix:
Conceptually this looks good to me, however I did not test it or review it in detail

sdirix (approved):
I was not able to test the retry logic as I was not able to hit the rate limits. However the provider continued to work, so it's definitely not worse than before.

PR #15800, opened 2025-06-10
feat(ai-core): handle toolCall errors gracefully instead of throwing

eneufeld:
#### What it does

Replace thrown exceptions in FrontendLanguageModelRegistryImpl.toolCall() with structured error objects to prevent crashes and improve error handling.

#### How to test

add a broken tool/break an existing tool or use coder long enough until the llm provides an invalid json for patching.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/browser/frontend-language-model-registry.spec.ts
+++ b/packages/ai-core/src/browser/frontend-language-model-registry.spec.ts
@@ -1,0 +1,262 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { enableJSDOM } from '@theia/core/lib/browser/test/jsdom';
+let disableJSDOM = enableJSDOM();
+import { FrontendApplicationConfigProvider } from '@theia/core/lib/browser/frontend-application-config-provider';
+FrontendApplicationConfigProvider.set({});
+
+import { expect } from 'chai';
+import {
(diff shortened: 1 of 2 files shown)


packages/ai-core/src/browser/frontend-language-model-registry.ts
```diff
@@ -283,14 +283,19 @@ export class FrontendLanguageModelRegistryImpl
...
         const request = this.requests.get(id)!;
         const tool = request.tools?.find(t => t.id === toolId);
         if (tool) {
-            return tool.handler(arg_string);
+            try {
+                return tool.handler(arg_string);
+            } catch (error) {
+                const errorMessage = error instanceof Error ? error.message : String(error);
+                return Promise.resolve({ error: true, message: `Error executing tool '${toolId}': ${errorMessage}` });
+            }
```
sdirix: This is not an async-method, therefore you can not reliably use try-catch here. Generally speaking it's best to stick with `then`/`catch` in non-async methods and with `try-catch` in async methods to avoid surprises.

I noticed this because I modified the `getFileContent` handler to always throw an error. That error was not caught here. To clean this up I recommend the following

```suggestion
            return tool.handler(arg_string).catch(error => {
                const errorMessage = error instanceof Error ? error.message : String(error);
                return Promise.resolve({ error: true, message: `Error executing tool '${toolId}': ${errorMessage}` });
            });
```

packages/ai-core/src/browser/frontend-language-model-registry.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+        if (tool) {
+            try {
+                return tool.handler(arg_string);
+            } catch (error) {
+                const errorMessage = error instanceof Error ? error.message : String(error);
+                return Promise.resolve({ error: true, message: `Error executing tool '${toolId}': ${errorMessage}` });
+            }
+        }
+        return Promise.resolve({ error: true, message: `Tool '${toolId}' not found in the available tools for this request.` });
+    }
```
sdirix: I'm not a huge fan of copy-pasting this method. This just points to the original class being too large / too complex to properly unit test. There are two approaches to fix this rather easily:

- We don't need any of the injected dependencies. So modify the original "requests" field to be protected and then create a subclass of `FrontendLanguageModelRegistryImpl` for testing. Then you can properly test the `toolCall` method, OR
- The logic in `toolCall` is largely independent of the the internals of `FrontendLanguageModelRegistryImpl` except for retrieving the original request. So all the "business-logic" of this function could be placed in a util next to the class, with the `toolCall` just resolving the request and then executing the util function. The tests could then just test the util function instead of the `toolCall` function.

PR #15815, opened 2025-06-11
Allow adding variants via files

### LINKED ISSUE #15794: Creating a Prompt Variant does not work
### Bug Description:

Following the instructions from the documentation to copy an existing file in the prompt templates directory (user or workspace) by changing the name after the ID, does not work. Prompt variants are not created (registered) in the IDE.

### Steps to Reproduce:

1. In the IDE click "edit" behind any prompt variant of an AI agent. The prompttemplate file is created either in the user prompt template directory or, if set, in the workspace prompt template directory.
2. Copy the file and change the name after the ID (ID is "architect" or "coder"). 
3. Look in the IDE to see a new variant for that AI agent (restart IDE). No new variant visible.

### Additional Information

- Operating System: Win 11
- Theia Version: 1.62.101
-
@JonasHelming Can you take a look at this?

JonasHelming:
fixed #15794

#### What it does

Re-enable adding prompt variants by placing files that start with a variant set id.

#### How to test

Place a file that starts with the id of a variant set in the global or workspace prompt directory and see that you can use (and modify) it as a variant.

#### Follow-ups

We might consider to load customizations and custom variants into memory as done with the built ins in case we run into memory issues.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/common/prompt-service.ts
+++ b/packages/ai-core/src/common/prompt-service.ts
@@ -726,7 +726,24 @@ export class PromptServiceImpl implements PromptService {
     }
 
     getVariantIds(fragmentId: string): string[] {
-        return this._promptVariantSetsMap.get(fragmentId) || [];
+        const builtInVariants = this._promptVariantSetsMap.get(fragmentId) || [];
+
+        // Check for custom variants from customization service
+        if (this.customizationService) {
+            const allCustomizedIds = this.customizationService.getCustomizedPromptFragmentIds();
+            // Find customizations that start with the variant set ID
+            // These are considered variants of this variant set
+            const customVariants = allCustomizedIds.filter(id =>
+                id !== fragmentId && id.startsWith(fragmentId)
+            );
+
+            if (customVariants.length > 0) {
+                // Combine built-in variants with custom variants, without modifying the internal state
+                return [...builtInVariants, ...customVariants];
+            }
+        }
+
+        return builtInVariants;
     }
 
     getDefaultVariantId(promptVariantSetId: string): string | undefined {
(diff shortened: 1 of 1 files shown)


JonasHelming:
@sgraband Could you have a look?

packages/ai-core/src/common/prompt-service.ts
```diff
@@ -726,15 +726,50 @@ export class PromptServiceImpl implements PromptService {
     }
 
     getVariantIds(fragmentId: string): string[] {
```
sgraband:
```suggestion
    getVariantIds(variantSetId: string): string[] {
```

It's not part of your code, but i think it makes sense to fix this for clarity.

packages/ai-core/src/common/prompt-service.ts
```diff
@@ -726,15 +726,50 @@ export class PromptServiceImpl implements PromptService {
...
-        return this._promptVariantSetsMap.get(fragmentId) || [];
+        const builtInVariants = this._promptVariantSetsMap.get(fragmentId) || [];
+
+        // Check for custom variants from customization service
+        if (this.customizationService) {
+            const allCustomizedIds = this.customizationService.getCustomizedPromptFragmentIds();
+            // Find customizations that start with the variant set ID
+            // These are considered variants of this variant set
+            const customVariants = allCustomizedIds.filter(id =>
+                id !== fragmentId && id.startsWith(fragmentId)
```
sgraband: Why do we filter out the variants with the same name as the variantSetId? To avoid confusion? Then we should also rename the built in variants that are named the same as their variant set. WDYT?

JonasHelming: the reason is that we have built in variants with the same id as the variant set id. if you customize them, the customization would also be treated as a variant.

Maybe we should change all ides to this scheme:

VarianSet ID: architect-system
Variants: architect-system-default, architect-system-next, etc. so that they start with the variantsetid by default?

If so I would do this in a separate PR

packages/ai-core/src/common/prompt-service.ts
```diff
@@ -725,16 +725,51 @@ export class PromptServiceImpl implements PromptService {
...
+    getVariantIds(variantSetId: string): string[] {
+        const builtInVariants = this._promptVariantSetsMap.get(fragmentId) || [];
+
+        // Check for custom variants from customization service
+        if (this.customizationService) {
+            const allCustomizedIds = this.customizationService.getCustomizedPromptFragmentIds();
+            // Find customizations that start with the variant set ID
+            // These are considered variants of this variant set
+            const customVariants = allCustomizedIds.filter(id =>
+                id !== fragmentId && id.startsWith(fragmentId)
```
sgraband:
```suggestion
                id !== variantSetId && id.startsWith(variantSetId)
```

packages/ai-core/src/common/prompt-service.ts
sgraband: Yes i like the naming scheme :+1: Seperate PR sounds good.

sgraband (approved):
LGTM :+1:

PR #15816, opened 2025-06-12
feat: theia-ai agent completion notification

xai:
The preference `ai-features.agentSettings.<agent-id>.completionNotification` can be used to configure notification settings for a theia-ai agent.

Notifications will only be shown if the Chat Widget does not have the focus.

The following notification settings are supported:
- 'off': no notification. This is the default setting to be compliant with existing behavior.
- 'on': tries 'os-notification' first, if that fails, it falls back to 'message'
- 'os-notification': Displays a notification in the operating system or the web browser, depending on whether electron is used or not
- 'message': uses the theia-message service
- 'blink': blinks the tab in the browser

Fixes eclipsesource/theia#223

#### How to test

1. Verify that no notification appears if you send a message to an agent
2. Setup agent completion notification, e.g., do something like this in your settings.json:
```[json]
{
        "Coder": {
             "completionNotification": "on" // or "os-notification", "message", or "blink"
        }
    }
}
```
3. Test the respective notification behavior by sending a message to an agent in the chat view

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


packages/ai-core/src/common/settings-service.ts
```diff
@@ -35,4 +35,9 @@ export interface AgentSettings {
      * If a main template is not present in this mapping, it means the main template is used.
      */
     selectedVariants?: Record<string, string>;
+    /**
+     * Configuration for completion notifications when the agent finishes a task.
+     * If undefined, defaults to 'on' which uses global notification preferences.
+     */
+    completionNotification?: import('./notification-types').NotificationType;
```
sgraband: Can we use a standard import here?

packages/ai-core/src/common/notification-types.ts
```diff
@@ -0,0 +1,40 @@
...
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+export const NOTIFICATION_TYPE_ON = 'on' as const;
+export const NOTIFICATION_TYPE_OFF = 'off' as const;
+export const NOTIFICATION_TYPE_OS_NOTIFICATION = 'os-notification' as const;
+export const NOTIFICATION_TYPE_MESSAGE = 'message' as const;
+export const NOTIFICATION_TYPE_BLINK = 'blink' as const;
```
sgraband: Never seen these casts before, are they necessary?

sgraband: Same below

packages/ai-core/src/common/notification-types.ts
```diff
@@ -0,0 +1,40 @@
...
+export const NOTIFICATION_CONCRETE_TYPES = [
+    NOTIFICATION_TYPE_OS_NOTIFICATION,
+    NOTIFICATION_TYPE_MESSAGE,
+    NOTIFICATION_TYPE_BLINK
+] as const;
+
+export const ALL_NOTIFICATION_TYPES = [
+    ...NOTIFICATION_CONTROL_TYPES,
+    ...NOTIFICATION_CONCRETE_TYPES
+] as const;
```
sgraband: Why is this split like this? Can't we just have one field with all types?

I am also unsure about the `on` option. AFAIK we have the possibility to set a fallback notification if we just use the two options we now use for on on default, which would make the on option obsolete and the user can decide for himself what he sets as the main option and as the fallback. Or am i missing something?

packages/ai-core/src/browser/agent-preferences.ts
```diff
@@ -26,15 +30,16 @@ export const AgentSettingsPreferenceSchema: PreferenceSchema = {
       type: 'object',
       title: nls.localize('theia/ai/agents/title', 'Agent Settings'),
       hidden: true,
-      markdownDescription: nls.localize('theia/ai/agents/mdDescription', 'Configure agent settings such as enabling or disabling specific agents, configuring prompts and \
-        selecting LLMs.'),
+      markdownDescription: nls.localize('theia/ai/agents/mdDescription',
```
sgraband: Are these formatting changes required (e.g. for the linter)?

packages/ai-core/src/browser/agent-preferences.ts
```diff
@@ -61,10 +68,25 @@ export const AgentSettingsPreferenceSchema: PreferenceSchema = {
...
+            enum: [...ALL_NOTIFICATION_TYPES],
+            title: nls.localize('theia/ai/agents/completionNotification/title', 'Completion Notification'),
+            markdownDescription: nls.localize('theia/ai/agents/completionNotification/mdDescription',
+              'Notification behavior when this agent completes a task.\n\
+                - `on`: Use the global default and fallback notification preferences\n\
+                - `os-notification`: Show OS/system notifications only\n\
+                - `blink`: Blink the window bar\n\
+                - `message`: Show notifications in the status bar\n\
+                - `off`: Disable notifications for this agent'),
+            default: NOTIFICATION_TYPE_ON
```
sgraband:
```suggestion
            default: NOTIFICATION_TYPE_OFF
```

Should be OFF right?

packages/ai-core/src/browser/window-blink-service.ts
```diff
@@ -0,0 +1,257 @@
...
+    success: boolean;
+    /** Error message if the blink failed */
+    error?: string;
+}
+
+/**
+ * Service for blinking/flashing the application window to get user attention.
+ */
+@injectable()
+export class WindowBlinkService {
```
sgraband, line 34: This does not work for the `window.titleBarStyle`: `custom` setting

packages/ai-core/src/browser/os-notification-service.ts
```diff
@@ -0,0 +1,297 @@
...
+            return false;
+        }
+
+        return !!(
+            (window as unknown as { electronAPI?: unknown }).electronAPI ||
+            (window as unknown as { electron?: unknown }).electron ||
+            (window as unknown as { require?: unknown }).require ||
+            (typeof navigator !== 'undefined' && navigator.userAgent && navigator.userAgent.toLowerCase().includes('electron'))
+        );
+    }
```
sgraband: I believe Theia already provides a way to check if you run in Electron.

packages/ai-core/src/browser/window-blink-service.ts
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        // Additional detection methods as fallback
+        return !!(
+            (window as unknown as { electronAPI?: unknown }).electronAPI ||
+            (window as unknown as { electron?: unknown }).electron ||
+            (window as unknown as { process?: { type?: string } }).process?.type ||
+            (typeof navigator !== 'undefined' && navigator.userAgent && navigator.userAgent.toLowerCase().includes('electron'))
+        );
+    }
```
sgraband: Same as in other class. If this is not sufficient i would at least at this to the core package and provide the functinality there and not calculate this in every class seperately.

sgraband (changes requested):
Thanks for the contribution. Works great from what i tested. However, i have some inline comments. Could you check those? Thanks in advance :slightly_smiling_face:

packages/ai-core/src/common/settings-service.ts
xai: Absolutely!

packages/ai-core/src/common/notification-types.ts
xai: Was necessary but I agree it is weird. So I changed all of that to enum instead.

packages/ai-core/src/common/notification-types.ts
xai: Great catch! Originally, `on` was `auto` until I renamed it. But of course, no need for it at all! Removed!

packages/ai-core/src/browser/agent-preferences.ts
xai: No, these were indeed accidental changes. I reverted them.

packages/ai-core/src/browser/agent-preferences.ts
xai: Yes, good catch!

packages/ai-core/src/browser/window-blink-service.ts
xai: I changed the implementation of this and in my window-manager (i3wm) it works. Could be a follow-up if the problem remains.

packages/ai-core/src/browser/os-notification-service.ts
xai: Nice! I adjusted the code accordingly.

xai:
@sgraband Thanks for reviewing! I addressed all comments and rebased against master again.

packages/ai-core/src/common/settings-service.ts
```diff
@@ -35,4 +36,9 @@ export interface AgentSettings {
      * If a main template is not present in this mapping, it means the main template is used.
      */
     selectedVariants?: Record<string, string>;
+    /**
+     * Configuration for completion notifications when the agent finishes a task.
+     * If undefined, defaults to 'on' which uses global notification preferences.
```
sgraband:
```suggestion
     * If undefined, defaults to 'off'.
```

packages/ai-core/src/browser/os-notification-service.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    /**
+     * Check if running in Electron environment
+     *
+     * @returns true if running in Electron, false otherwise
+     */
+    isElectronEnvironment(): boolean {
+        return this.isElectron;
+    }
+
```
sgraband:
```suggestion

```
This is unused

packages/ai-core/src/common/notification-types.ts
```diff
@@ -0,0 +1,36 @@
...
+export type NotificationType = NotificationControlType | ConcreteNotificationType;
+
+export const NOTIFICATION_TYPE_OFF = NotificationControlType.OFF;
+export const NOTIFICATION_TYPE_OS_NOTIFICATION = ConcreteNotificationType.OS_NOTIFICATION;
+export const NOTIFICATION_TYPE_MESSAGE = ConcreteNotificationType.MESSAGE;
+export const NOTIFICATION_TYPE_BLINK = ConcreteNotificationType.BLINK;
+
+export const NOTIFICATION_CONTROL_TYPES = Object.values(NotificationControlType);
+export const NOTIFICATION_CONCRETE_TYPES = Object.values(ConcreteNotificationType);
+export const ALL_NOTIFICATION_TYPES = [...NOTIFICATION_CONTROL_TYPES, ...NOTIFICATION_CONCRETE_TYPES];
```
sgraband:
```suggestion
export const NOTIFICATION_TYPE_OFF = 'off';
export const NOTIFICATION_TYPE_OS_NOTIFICATION = 'os-notification';
export const NOTIFICATION_TYPE_MESSAGE = 'message';
export const NOTIFICATION_TYPE_BLINK = 'blink';

export type NotificationType =
    | typeof NOTIFICATION_TYPE_OFF
    | typeof NOTIFICATION_TYPE_OS_NOTIFICATION
    | typeof NOTIFICATION_TYPE_MESSAGE
    | typeof NOTIFICATION_TYPE_BLINK;

export const ALL_NOTIFICATION_TYPES: NotificationType[] = [
    NOTIFICATION_TYPE_OFF,
    NOTIFICATION_TYPE_OS_NOTIFICATION,
    NOTIFICATION_TYPE_MESSAGE,
    NOTIFICATION_TYPE_BLINK
];
```

Thanks for changing this. Just two things. We should probably use a union instead of an enum and also i think there is not benefit of splitting the two types right?

packages/ai-core/src/browser/ai-core-preferences.ts
```diff
@@ -130,14 +140,38 @@ export const aiCorePreferenceSchema: PreferenceSchema = {
...
+            enum: [...ALL_NOTIFICATION_TYPES],
+            default: NOTIFICATION_TYPE_OS_NOTIFICATION
+        },
+        [PREFERENCE_NAME_FALLBACK_NOTIFICATION_TYPE]: {
+            title: nls.localize('theia/ai/core/fallbackNotification/title', 'Fallback Notification Type'),
+            markdownDescription: nls.localize('theia/ai/core/fallbackNotification/mdDescription',
+                'The backup notification method used when the primary notification type fails.\n\
+                This ensures users always receive some form of notification when possible.'),
+            type: 'string',
+            enum: [...NOTIFICATION_CONCRETE_TYPES, NotificationControlType.OFF],
```
sgraband:
```suggestion
            enum: [...ALL_NOTIFICATION_TYPES],
```

Should be the same

packages/ai-core/src/browser/ai-core-preferences.ts
```diff
@@ -130,14 +140,38 @@ export const aiCorePreferenceSchema: PreferenceSchema = {
...
+            title: nls.localize('theia/ai/core/defaultNotification/title', 'Default Notification Type'),
+            markdownDescription: nls.localize('theia/ai/core/defaultNotification/mdDescription',
+                'The primary notification method used when an AI agent completes a task.\n\
+                - `os-notification`: Show OS/system notifications\n\
+                - `message`: Show notifications in the status bar/message area\n\
+                - `blink`: Blink or highlight the UI\n\
+                - `off`: Disable all notifications'),
+            type: 'string',
+            enum: [...ALL_NOTIFICATION_TYPES],
+            default: NOTIFICATION_TYPE_OS_NOTIFICATION
```
sgraband: This is now never used. Sorry for the back and forth, but i believe we should readd the on/auto option that takes this value. And then you can still choose another primary notification for the agent directly. I would however take some time to make this clearer in the descriptions (Btw. the newline does not work for me in the settings view.).

So something like: `The default notification method used when an AI agent completes a task`
And in the agent setting we should write that on auto this is taken then from this setting.

Again sorry for the confusion.

packages/ai-core/src/browser/window-blink-service.ts
sgraband: It still does not work for me. Follow up would be fine for me :+1:

packages/ai-core/src/common/settings-service.ts
xai: ACK

packages/ai-core/src/browser/os-notification-service.ts
xai: ACK

packages/ai-core/src/common/notification-types.ts
xai: ACK. union instead of enum. I also removed the ALL_ prefix, so now we only have `NOTIFICATION_TYPES`.

packages/ai-core/src/browser/ai-core-preferences.ts
xai: ACK

packages/ai-core/src/browser/ai-core-preferences.ts
xai: ok, no problem! added it back.

xai:
I updated the PR again. As the blink service does not really work for you, i tried to simplify it as much as possible (still works in i3wm) and we can do a follow-up issue.

ndoschek:
Thank you @xai for this great contribution!
Tested the changes in both browser and Electron – works great 👍
Also works nicely when switching to another chat session in the meantime.
Also the blinking behavior is working well for me on Ubuntu in Chrome, Firefox, and Electron. It's a nice, subtle alternative to OS notifications!
Would also recommend testing on Windows and macOS before merging to ensure cross-platform consistency.
I will ask Christian and Thomas to have a quick look as well.

I have a few comments that could be handled in a follow-up:
- Add the agent-specific settings for notifications to the Agent settings the AI Configuration view
- Clicking a notification (Theia notificiation or OS notification) could ideally bring the user directly to the corresponding chat session
- The settings UX could be a bit clearer imho: they apply only to the "on" state of notifications, but there are also specific per-agent settings – might be worth simplifying at least in the description.

xai:
> Would also recommend testing on Windows and macOS before merging to ensure cross-platform consistency. I will ask Christian and Thomas to have a quick look as well.

Very good point! I only tested on Linux so far.

> I have a few comments that could be handled in a follow-up:
> 
>     * Add the agent-specific settings for notifications to the Agent settings the AI Configuration view
>     * Clicking a notification (Theia notificiation or OS notification) could ideally bring the user directly to the corresponding chat session
>     * The settings UX could be a bit clearer imho: they apply only to the "on" state of notifications, but there are also specific per-agent settings – might be worth simplifying at least in the description.

:+1:

packages/ai-core/src/common/notification-types.ts
cdamus: nit: I prefer this pattern for maintainability:

```ts
export const ALL_NOTIFICATION_TYPES = [
    NOTIFICATION_TYPE_OFF,
    NOTIFICATION_TYPE_OS_NOTIFICATION,
    NOTIFICATION_TYPE_MESSAGE,
    NOTIFICATION_TYPE_BLINK
] as const;
export type NotificationType = typeof ALL_NOTIFICATION_TYPES[number];
```

cdamus:
It looks like the Settings UI intended a bullet-list formatting that didn't quite work out?
CleanShot 2025-06-24 at 09 07 21@2x

cdamus:
On macOS Sequoia 15.5, I tried the following:

- ask Architect to tell me how the new notification system for completion of chat requests works
- switched to another application so the asynchronous notification would alert me to completion of the response
- nothing happened (see screenshot below)

As per the default, I had OS notifications on as the default mechanism and Theia messages for the fall-back.

I did not see in System Preferences for Notifications that either Electron or Theia requested permission to send notifications. I wasn't sure whether I should expect the OS notification to work or not as I don't know whether when running the Electron example from the development environment uses a Mac app bundle with the Info.plist that requests the Notification entitlement. That will be necessary for any downstream packages like Theia IDE to handle; perhaps they've already accounted for it? That's something for somebody to check before this will really work on macOS 😀 

My gut feeling is that, when running the Electron example in development mode, the notification should have "just worked" without the whole app bundle scaffolding and Info.plist, because it's development mode. But I don't know enough about macOS application development to say that with any confidence. In any case, when it didn't work, I would have expected the fall-back message to be employed, and it seems that it wasn't.

CleanShot 2025-06-24 at 09 11 20@2x

tsmaeder:
I tested this with Gemini: 

1. Open settings, the default value for "notification" is "os-notification", so I did not change anything
2. I asked the architect agent a question
3. The architect finished answering, but not OS notification happened.

tsmaeder:
Same result when I changed the setting to "message".

tsmaeder:
Not clear to me whether I need to set the notification preference for every agent or not. If it's per agent, why is there a "global" setting?

sgraband:
Yes you are right @tsmaeder! Currently you need to set the setting for each agent, which is not very nice for the user. I will discuss with Olaf and get back to you, once we need a re-review.

cdamus:
I didn't see any notification settings in the Agents Configuration UI. Did I miss it somewhere? Or did I have to edit this in the JSON as outlined in the PR description? Sorry, I didn't read that so literally. If this is going to be per-agent configurable then it must be in the Agents UI.

tsmaeder:
> I didn't see any notification settings in the Agents Configuration UI. Did I miss it somewhere?

It's in the regular settings under "AI features"

cdamus:
> > I didn't see any notification settings in the Agents Configuration UI. Did I miss it somewhere?
> 
> 
> 
> It's in the regular settings under "AI features"

That's the defaults. I was inquiring about the per-agent settings that I expected to find in the Agents tab of the AI Configuration view 😀 I didn't make that clear, sorry.

sgraband:
After a quick discussion we decided to:
- Remove the fallback setting/option: It does not detect a lot of failing instances and just adds complexity.
- Apply the default notification option from the settings to all agents that do not have another option specified (default for the option is `off`): This way if you change the setting from `off` to `messsage` for example all agents will be adapted. The user can then decide to dig into the settings.json and set the value for each agent differently (advanced feature).
- Check if we can move the default notification option into the agent settings and adapt the description to make it more clear.
- And maybe even add a dropdown in the agent view to select the specfic notification option for an agent (this is likely in a follow up)
- Additionally, once we have these changes we will retest on Windows, Mac and Ubuntu.

ndoschek:
I created the follow up: #15894
Please feel free to add any additional items in there thanks!

@xai could you add the follow up to the respective section in the PR description as well? TIA

xai:
> I created the follow up: #15894 Please feel free to add any additional items in there thanks!
> 
> @xai could you add the follow up to the respective section in the PR description as well? TIA

Done! Thanks for creating the follow-up.

xai:
> LGTM 👍 But needs a rebase

Done.

PR #15839, opened 2025-06-13
Streamline Ask AI commands and fix Send to AI Chat error

### LINKED ISSUE #15827: Send to AI Chat throws an error
### Bug Description:

I don't know if that feature is a work  in progress (since it's experimental) but when I use "Send to AI Chat" in a code editor it throws the following error in the console and does nothing.

```
ai-editor-command-contribution.ts:138 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'prompt')
    at Object.execute (ai-editor-command-contribution.ts:138:40)
    at MonacoCommandRegistry.execute (monaco-command-registry.ts:69:50)
    at Object.execute (monaco-command-registry.ts:59:40)
    at CommandRegistry.executeCommand (command.ts:341:42)
```

### Steps to Reproduce:

Image

### Additional Information

- Operating System: Windows
- Theia Version: latest master

ndoschek:
#### What it does

- `Send to AI` Chat command is not available via the command palette, but can still be triggered internally by other commands, as it expects a mandatory prompt argument
- Align `Ask AI` commands for editor and terminal for a more streamlined UX

Fixes #15827

#### How to test

- Command Palette: Confirm that `Send to AI` is not listed without a prompt argument.
- Editor & Terminal `Ask AI`: Invoke both commands (either via command palette, keybinding or context menu) and verify they trigger the correct actions without interfering.  
- Code Actions: Run `Fix with AI` or `Explain with AI` on an erroneous code snippet to ensure they still work as before.  

#### Follow-ups

--

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-editor/src/browser/ai-editor-command-contribution.ts
+++ b/packages/ai-editor/src/browser/ai-editor-command-contribution.ts
@@ -16,26 +16,26 @@
 
 import { ChatAgentLocation, ChatService } from '@theia/ai-chat';
 import { AICommandHandlerFactory, ENABLE_AI_CONTEXT_KEY } from '@theia/ai-core/lib/browser';
-import { MenuContribution, MenuModelRegistry } from '@theia/core';
-import { KeybindingContribution, KeybindingRegistry } from '@theia/core/lib/browser';
+import { isObject, MenuContribution, MenuModelRegistry } from '@theia/core';
+import { ApplicationShell, codicon, KeybindingContribution, KeybindingRegistry } from '@theia/core/lib/browser';
 import { Coordinate } from '@theia/core/lib/browser/context-menu-renderer';
 import { Command, CommandContribution, CommandRegistry } from '@theia/core/lib/common';
-import { nls } from '@theia/core/lib/common/nls';
 import { inject, injectable } from '@theia/core/shared/inversify';
-import { EditorContextMenu } from '@theia/editor/lib/browser';
+import { EditorContextMenu, EditorWidget } from '@theia/editor/lib/browser';
 import { MonacoCommandRegistry, MonacoEditorCommandHandler } from '@theia/monaco/lib/browser/monaco-command-registry';
 import { MonacoEditor } from '@theia/monaco/lib/browser/monaco-editor';
 import { AskAIInputWidget } from './ask-ai-input-widget';
 
 export namespace AI_EDITOR_COMMANDS {
-    export const AI_EDITOR_ASK_AI: Command = {
+    export const AI_EDITOR_ASK_AI: Command = Command.toLocalizedCommand({
         id: 'ai-editor.contextAction',
-        label: nls.localize('theia/ai-editor/contextMenu', 'Ask AI [Experimental]')
-    };
-    export const AI_EDITOR_SEND_TO_CHAT: Command = {
+        label: 'Ask AI',
+        iconClass: codicon('sparkle')
+    }, 'theia/ai-editor/contextMenu');
+    export const AI_EDITOR_SEND_TO_CHAT: Command = Command.toLocalizedCommand({
         id: 'ai-editor.sendToChat',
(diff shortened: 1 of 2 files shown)


packages/ai-editor/src/browser/ai-editor-command-contribution.ts
```diff
@@ -175,3 +186,11 @@ export class AiEditorCommandContribution implements CommandContribution, MenuCon
         });
     }
 }
+
+function containsPrompt(args: unknown[]): args is (unknown | { prompt: string })[] {
+    return args.filter(arg => isPromptArg(arg)).length > 0;
```
colin-grant-work: We can avoid looking through every element with `.some()`

```suggestion
    return args.some(arg => isPromptArg(arg));
```

packages/ai-editor/src/browser/ai-editor-command-contribution.ts
```diff
@@ -175,3 +186,11 @@ export class AiEditorCommandContribution implements CommandContribution, MenuCon
         });
     }
 }
+
+function containsPrompt(args: unknown[]): args is (unknown | { prompt: string })[] {
```
colin-grant-work: `Array<unknown | something>` doesn't narrow the type from `Array<unknown>`, since `unknown` itself includes all possible unions (except maybe `never`?). Is the type predicate important, or would it make sense to return just `boolean`?

packages/ai-editor/src/browser/ai-editor-command-contribution.ts
```diff
@@ -175,3 +186,11 @@ export class AiEditorCommandContribution implements CommandContribution, MenuCon
...
+    return args.filter(arg => isPromptArg(arg)).length > 0;
+}
+
+function isPromptArg(arg: unknown): arg is { prompt: string } {
+    return isObject(arg) && 'prompt' in arg;
```
colin-grant-work: Would it be worth adding a check that `arg.prompt` is in fact a `string`?

packages/ai-terminal/src/browser/ai-terminal-contribution.ts
```diff
@@ -68,7 +74,8 @@ export class AiTerminalCommandContribution implements CommandContribution, MenuC
                         this.terminalAgent
                     );
                 }
-            }
+            },
+            isEnabled: () => this.shell.currentWidget instanceof TerminalWidget
```
colin-grant-work: The `execute` method checks for `TerminalWidgetImpl` - should the two checks match?

colin-grant-work:
> Command Palette: Confirm that Send to AI is not listed without a prompt argument.

If the goal is to ensure that it isn't visible in the command palette, that appears to have been achieved. Were there any other conditions in which it was appearing inappropriately?

packages/ai-editor/src/browser/ai-editor-command-contribution.ts
ndoschek: Totally agree 👍. Returning a boolean makes perfect sense here

packages/ai-terminal/src/browser/ai-terminal-contribution.ts
ndoschek: Ah yes totally, I must've overlooked that.
Also on a second look I noticed, this widget should only appear for user created terminals, could you have a second look if you think this check is good enough?

ndoschek:
Thanks a lot for your review @colin-grant-work!

> If the goal is to ensure that it isn't visible in the command palette, that appears to have been achieved. Were there any other conditions in which it was appearing inappropriately?

I am not aware of any additional scenarios either, so this should be fine.

Also, your last comment actually made me realize that the AI terminal widget should only show up for user-created terminals (as the widget would execute terminal commands, which is not possible for e.g. task terminals).
I've created a separate issue to track the streamlining part, to keep the scope a bit cleaner.

I've updated the PR description accordingly. If you have some time, a second review would be very helpful. Thanks again!

PR #15413, opened 2025-04-07
feat(mcp): enhance server management with resource handling and trans…

kaikongbj:
This pull request includes several updates and improvements to the MCP (Model Context Protocol) integration within the Theia framework. The changes primarily focus on enhancing the MCP server management capabilities, updating dependencies, and improving error handling.

### Dependency Updates:
* Updated the dependency version for `@modelcontextprotocol/sdk` in `package.json` to use a caret (^) to allow for minor version updates.

### Error Handling Improvements:
* Modified various methods in `MCPCommandContribution` to use `await` with `messageService` calls for better error handling and user feedback. [[1]](diffhunk://#diff-009879f2a3bec62e8ce7ca69eb5f73593bd1f2c58df41753c6a3f7405026b92aL60-R60) [[2]](diffhunk://#diff-009879f2a3bec62e8ce7ca69eb5f73593bd1f2c58df41753c6a3f7405026b92aL82-R84) [[3]](diffhunk://#diff-009879f2a3bec62e8ce7ca69eb5f73593bd1f2c58df41753c6a3f7405026b92aL101-R101) [[4]](diffhunk://#diff-009879f2a3bec62e8ce7ca69eb5f73593bd1f2c58df41753c6a3f7405026b92aL111-R113)

### MCP Server Enhancements:
* Added new methods `listResources` and `readResource` to the `MCPServer` and `MCPServerManager` interfaces to support resource listing and reading functionalities. [[1]](diffhunk://#diff-94634d82fb21669f1730960267df75493d9c3808f5312f2b08bf98fd0466fa28R41-R43) [[2]](diffhunk://#diff-94634d82fb21669f1730960267df75493d9c3808f5312f2b08bf98fd0466fa28R59-R61) [[3]](diffhunk://#diff-d885fef5f5e1af1ff6da79d77fb3d375390e1bfca45cee420a0e2f3c7f98eea3R134-R149) [[4]](diffhunk://#diff-baac4f4f84bbd7dd539d2a99f2489830f03476a2a59f924c507724caa7d2850fR198-R206)
[...]

PR #15842, opened 2025-06-16
feat: enhance mcp server management

fipro78:
#### What it does

This PR is an enhancement to https://github.com/eclipse-theia/theia/pull/15413

It adds the ability to connect to a remote MCP server. Compared to the original PR, this PR supports SSE and StreamableHttp, which is the successor to the SSE implementation. The implementation first tries to access the remote MCP server via StreamableHttp, if that fails it tries to fallback to SSE. While you could argue that this could also be a configuration value instead of the fallback solution, the question is on the user side. Is a user of a remote MCP server aware of the communication protocol? We therefore decided for the fallback solution, to make the configuration more convenient.

We also changed the configuration. While the original PR used the `env` for the server configuration, we introduced new top level configurations and made `command` optional. So with this PR you can either create a local MCP server configuration with `command`, `args` and `env`, or you create a remove MCP server configuration with `serverUrl`, `serverAuthToken` and an optional `serverAuthTokenHeader` in case the server uses a non-default authorization token header name.

#### How to test

Configure a remote MCP server and try to use it. You could also try to start an MCP server locally and access it via server URL. But I haven't tested this scenario, as we have remote MCP servers in our setup to test against.

I found a simple free accessible MCP that does not require a token.

```json
    "ai-features.mcp.mcpServers": {
      "cloudflare" :{
        "serverUrl": "https://demo-day.mcp.cloudflare.com/sse"
      }
    },
```

As you can see from the URL, it uses SSE, so it can be used to verify the automatic fallback.

Once connected, paste the following to the AI chat to see that it works:

_Can you tell be something about the Cloudflare Demo Day. Use ~mcp_cloudflare_mcp_demo_day_info  to answer the question._

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-mcp/src/browser/mcp-command-contribution.ts
+++ b/packages/ai-mcp/src/browser/mcp-command-contribution.ts
@@ -57,7 +57,7 @@ export class MCPCommandContribution implements CommandContribution {
                 try {
                     const startedServers = await this.mcpFrontendService.getStartedServers();
                     if (!startedServers || startedServers.length === 0) {
-                        await this.messageService.error(nls.localize('theia/ai/mcp/error/noRunningServers', 'No MCP servers running.'));
+                        this.messageService.error(nls.localize('theia/ai/mcp/error/noRunningServers', 'No MCP servers running.'));
                         return;
                     }
                     const selection = await this.getMCPServerSelection(startedServers);
@@ -79,9 +79,9 @@ export class MCPCommandContribution implements CommandContribution {
                     const startableServers = servers.filter(server => !startedServers.includes(server));
                     if (!startableServers || startableServers.length === 0) {
                         if (startedServers && startedServers.length > 0) {
-                            await this.messageService.error(nls.localize('theia/ai/mcp/error/allServersRunning', 'All MCP servers are already running.'));
+                            this.messageService.error(nls.localize('theia/ai/mcp/error/allServersRunning', 'All MCP servers are already running.'));
                         } else {
-                            await this.messageService.error(nls.localize('theia/ai/mcp/error/noServersConfigured', 'No MCP servers configured.'));
+                            this.messageService.error(nls.localize('theia/ai/mcp/error/noServersConfigured', 'No MCP servers configured.'));
                         }
                         return;
                     }
@@ -99,7 +99,7 @@ export class MCPCommandContribution implements CommandContribution {
                             if (serverDescription.tools) {
                                 toolNames = serverDescription.tools.map(tool => tool.name).join(',');
                             }
-                            await this.messageService.info(
+                            this.messageService.info(
                                 nls.localize('theia/ai/mcp/info/serverStarted', 'MCP server "{0}" successfully started. Registered tools: {1}', selection, toolNames ||
                                     nls.localize('theia/ai/mcp/tool/noTools', 'No tools available.'))
                             );
@@ -109,9 +109,9 @@ export class MCPCommandContribution implements CommandContribution {
                             console.error('Error while starting MCP server:', serverDescription.error);
                         }
                     }
-                    await this.messageService.error(nls.localize('theia/ai/mcp/error/startFailed', 'An error occurred while starting the MCP server.'));
+                    this.messageService.error(nls.localize('theia/ai/mcp/error/startFailed', 'An error occurred while starting the MCP server.'));
                 } catch (error) {
-                    await this.messageService.error(nls.localize('theia/ai/mcp/error/startFailed', 'An error occurred while starting the MCP server.'));
+                    this.messageService.error(nls.localize('theia/ai/mcp/error/startFailed', 'An error occurred while starting the MCP server.'));
                     console.error('Error while starting MCP server:', error);
                 }
             }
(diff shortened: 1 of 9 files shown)


packages/ai-mcp/src/browser/mcp-command-contribution.ts
```diff
@@ -57,7 +57,7 @@ export class MCPCommandContribution implements CommandContribution {
                 try {
                     const startedServers = await this.mcpFrontendService.getStartedServers();
                     if (!startedServers || startedServers.length === 0) {
-                        this.messageService.error(nls.localize('theia/ai/mcp/error/noRunningServers', 'No MCP servers running.'));
+                        await this.messageService.error(nls.localize('theia/ai/mcp/error/noRunningServers', 'No MCP servers running.'));
```
sdirix: In Theia we usually don't `await` the `messageService`. It's "just" an informational message to the user, so there is no need to block execution here.

packages/ai-mcp/src/node/mcp-server.ts
```diff
@@ -52,7 +57,8 @@ export class MCPServer {
     }
 
     isRunnning(): boolean {
```
sdirix, line 48: I just noticed that there is a typo here. If it's fine for you I would recommend that we fix it with this PR.

packages/ai-mcp/src/browser/mcp-frontend-application-contribution.ts
```diff
@@ -22,10 +22,13 @@ import { JSONObject } from '@theia/core/shared/@lumino/coreutils';
...
 interface MCPServersPreferenceValue {
-    command: string;
+    command?: string;
     args?: string[];
     env?: { [key: string]: string };
     autostart?: boolean;
+    serverUrl?: string;
+    serverAuthToken?: string;
+    serverAuthTokenHeader?: string;
 };
```
sdirix: I think we should type this so that either `command` or `serverUrl` are required. Having all optional allows "empty" configurations which with this PR lead to runtime errors. We can do so with Typescript unions.

packages/ai-mcp/src/browser/mcp-frontend-application-contribution.ts
```diff
@@ -35,10 +38,13 @@ interface MCPServersPreference {
...
-            'command' in obj && typeof obj.command === 'string' &&
+            (!('command' in obj) || typeof obj.command === 'string') &&
             (!('args' in obj) || Array.isArray(obj.args) && obj.args.every(arg => typeof arg === 'string')) &&
             (!('env' in obj) || !!obj.env && typeof obj.env === 'object' && Object.values(obj.env).every(value => typeof value === 'string')) &&
-            (!('autostart' in obj) || typeof obj.autostart === 'boolean');
+            (!('autostart' in obj) || typeof obj.autostart === 'boolean') &&
+            (!('serverUrl' in obj) || typeof obj.serverUrl === 'string') &&
+            (!('serverAuthToken' in obj) || typeof obj.serverAuthToken === 'string') &&
+            (!('serverAuthTokenHeader' in obj) || typeof obj.serverAuthTokenHeader === 'string');
     }
```
sdirix: This type guard should be adapted accordingly. This type guard is even more important than the interface, as it prevents invalid preferences to be propagated to MCP servers. With the adapted type guard it's possible to create empty servers which then crash (as we don't provide a `command` to the server)

packages/ai-mcp/src/common/mcp-server-manager.ts
```diff
@@ -79,7 +86,7 @@ export interface MCPServerDescription {
     /**
      * The command to execute the MCP server.
      */
-    command: string;
+    command?: string;
```
sdirix: Minor: Similarly it would be nice to have a union of `MCPServerDescription`, e.g.
 - A `BaseMCPServerDescription` with all common fields,
 - A `LocalServerDescription` with commands, args, env etc.
 - A `RemoteServerDescription` with serverUrl, token, headers etc.
 - with  type`MCPServerDescription = LocalServerDescription | RemoteServerDescription`

This is only a nice to have.

packages/ai-mcp/src/node/mcp-server.ts
```diff
@@ -149,23 +217,56 @@ export class MCPServer {
...
     }
 
     update(description: MCPServerDescription): void {
         this.name = description.name;
-        this.command = description.command;
+        this.command = description.command ? description.command : '';
         this.args = description.args;
         this.env = description.env;
+
+        if (description.serverUrl) {
```
sdirix: Minor: Here we already distinguish between the server and the local case, would be nicer to express this via the types as mentioned before.

sdirix (approved):

fipro78:
@sdirix 
I was just about to push the requested changes when I saw that you applied them yourself. The next time I will try to react faster on review comments. After one week without any reaction I didn't thought that it might be that urgent. ;)

At least I can see that I found the same solution that you applied, so I understood correctly what you requested.. :)

ndoschek:
Hi @fipro78,
Thanks again for your contribution!
This one’s actually on me – I asked to include it in today’s release and Stefan kindly offered to integrate the comments. Sorry for the confusion, I’ll make sure to communicate more clearly next time :)

Cheers,
Nina

fipro78:
@ndoschek 
If today is a release and the change is integrated in that release, I am more than happy that @sdirix helped out here. 

Thank you very much for the response and the explanation. Makes things much clearer.

PR #15846, opened 2025-06-16
Fix bug that customizations are treated as variants

JonasHelming:
#### What it does

Fixes a bug that prompt customizations are treated as variants if they start with the same id as the variant set

#### How to test

- Create a customization of the architect default prompt (the prompts for architect start with architect-system, which is the variant set id)
- Observe that the customizations are shown as variants, so they appear twice in the list.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


JonasHelming:
@ndoschek FYI

sgraband (approved):
LGTM :+1: We should move this logic to the customization service in a follow up, because this contains assumptions about the customization service, which could be changed.

PR #15854, opened 2025-06-18
Remove requirement that agent has to be mentioned first

### LINKED ISSUE #15272: Mentioning and agent needs to be at the beginning of prompt
### Bug Description:

Selecting an agent using "@" in the chat view only works when the message starts with the agent. It should be possible to reference an agent at any point in the prompt. Currently, when a different agent is referenced elsewhere in the prompt, the orchestrator takes over the selection and does not pin the referenced agent.

### Steps to Reproduce:

1. Open the chat view.
2. Write an instruction intended for the orchestrator that also references a different agent using "@" somewhere in the middle of the message.
3. Observe that the orchestrator selects an agent for you and does not pin the referenced agent.

Image

### Additional Information

- Theia Version: Theia IDE Preview 1.60.0 and Theia master

sgraband:
#### What it does

From my testing and looking at the code this does not have any side effects. @sdirix do you have an idea why we originally added this? Or was this an oversight?

Fixes #15272

#### How to test

Write something like `#_f @Coder` in the chat and see that the Agent is correctly set.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-request-parser.ts
+++ b/packages/ai-chat/src/common/chat-request-parser.ts
@@ -228,18 +228,6 @@ export class ChatRequestParserImpl implements ChatRequestParser {
             return;
         }
 
-        // The agent must come first
-        if (
-            parts.some(
-                p =>
-                    (p instanceof ParsedChatRequestTextPart &&
-                        p.text.trim() !== '') ||
-                    !(p instanceof ParsedChatRequestAgentPart)
-            )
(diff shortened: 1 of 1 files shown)


sdirix:
IIRC, variable resolving was dependent on agents at some point, so that would be an explanation on why this was coded this way. However I would definitely prefer if we are able to mention the agent wherever.

PR #15855, opened 2025-06-18
Fix caching in code completion

### LINKED ISSUE #15831: Caching break code completion
### Steps to Reproduce:

1. Empty .ts file
2. Write // Say hello to each former President of the United States
3. Some code gets completed (do not accept)
4. Change the first line to // Say hello to each member of the beatles
5. Suggestion still greets the presidents

When changing the file a lot, e.g. also adding // Say hellow to each member of the Rolling Stones
you can get all three suggestions eventually, probably depending on where you trigger the completion
I guess the cache must be invalidated much more agressivly.

### Additional Information

- Operating System:
- Theia Version: master

@ndoschek

sgraband:
#### What it does
Previously the caching only took into consideration the current line. This is not enough as it does not update when a previous or following line is changed. Fixed this with this commmt.
To avoid large caches the prefix and suffix is hashed.

Fixes #15831

#### How to test
- Create a empty ts file
- Type something like `// Greet the beatles` and press enter
- Wait for the completion to show
- Go back to line 1 and change to `Greet AC/DC` and press enter
- Notice that even though line 2 is unchanged the code completion suggestion changes.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-code-completion/src/browser/code-completion-cache.ts
+++ b/packages/ai-code-completion/src/browser/code-completion-cache.ts
@@ -27,31 +27,44 @@ export class CodeCompletionCache {
     }
 
     /**
-     * Generate a unique cache key based on input parameters
+     * Generate a unique cache key for code completion based on the file path, cursor position, and the hashed context (prefix and suffix).
+     * The prefix and suffix are hashed to avoid storing large or sensitive content directly in the cache key.
+     *
      * @param filePath Path of the current file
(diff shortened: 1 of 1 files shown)


ndoschek (approved):
Thanks @sgraband, works like a charm! :+1: 
I tested several different scenarios and it works great!

Issue #15804, opened 2025-06-10
Menu disappears when Opening *.md in preview


tsmaeder:
### Steps to Reproduce:

1. Set "title bar style" to "custom"
2. Open some files, including a `*.md` file
3. In the editor toolbar, click the "open in preview" icon
4. Observe: the preview opens to the right side of the text editor
5. Observe the menu bar disappears. Sometimes, you need to grab the top of the window and resize slightly for the menu bar to disappear.

### Additional Information

- Operating System: Windows 11
- Theia Version: 1.62. master

tsmaeder:
Note that the menu bar is still present and AFAICT properly sized, however, the "scrollTop" value of the `html` element gets set to 29 pixels, thus scrolling the menu bar out of view.

rschnekenbu:
I can indeed reproduce with Theia 1.62.100, and on Ubuntu 24.04. This works however when not using _custom_ title bar style

tsmaeder:
Strangely, this seems to be fixed by #15672

jfaltermeier:
On Theia master I am currently seeing this with Electron and a non custom title bar style on Ubuntu 24.04:

![Image](https://github.com/user-attachments/assets/79bf3ba6-027b-46e6-b1e9-0a8c10e8fbce)

The whole application is moving up under the native menu bar. 

I am not sure if this is related, but I could imagine that with a custom style, the toolbar itself would also move up and end up outside the electron window. 

Edit: yes, so I can reproduce it with both native and custom styles. For me the whole application moves up in both cases.

jfaltermeier:
What I reported seems to be reproducible only when I start the Theia Example from within the Theia IDE. I cannot reproduce it when launching from an OS terminal or a different IDE.
So it's probably a different issue that just looks similar

tsmaeder:
And by "from within Theia IDE" do  you mean in the debugger or from a terminal view inside the IDE?

jfaltermeier:
> And by "from within Theia IDE" do you mean in the debugger or from a terminal view inside the IDE?

From the terminal view, I haven't tried the debug launch configs.

ndoschek:
I observed the same behavior on current master as Johannes describes: the application content shifts upwards and a visible gap at the bottom, below the statusbar, appears.
This happens when started via debug launch configs, from a terminal inside the Theia IDE, and also from the OS terminal.
This happens for both browser and Electron examples, and with both native and custom titlebar styles.

[...]

Issue #15594, opened 2025-05-14
Restore Secondary Window to Main Window


tsmaeder:
### Feature Description:
We have the ability to extract widgets to a secondary window, if the widget implementation supports it. We would like to be able to return the extracted widget to the main window again.

sdirix:
My 2 cents:

> how will we restore the widget? Is there a button that returns the widget to where it came from or do we drag it back? 

For now we should move the widget back when we close the secondary window. That's the same behavior as VS Code has for its extracted editors

> Some widgets are singletons, for example the Explorer view. While this case does not happen in our current implementations, what would happen if we "unextracted" a singleton widget, but the there is already a copy in the main window? Can this even happen or does the scope of the singleton include the secondary windows?

Singletons should be singletons over the main window and secondary windows, i.e. if the file explorer would be open in a secondary window, it can't also be open in the main window

> Can we still close widgets or should "close" just unextract the widget?

For now I would go with "close" just triggering the unextract

> Do we need additional protocol from the widget implementation to make this work?

I don't think we need a lot of enhancement for the implementation. The trickiest part is to reattach to the original location. If that location no longer exists we should probably just reattach to the default location.

tsmaeder:
> For now we should move the widget back when we close the secondary window.

I don't think that makes a sense unless we have a very lightweight process for unextracting the window (e.g. by pressing "escape"). Otherwise, closing the widget becomes a two-click process. That sounds tedious.

PR #15837, opened 2025-06-13
Update electron to version 36.4.0

tsmaeder:
- use webUtils.getPathForFile() in drag and drop
- upated supported node versions to v20-v24

Fixes #15672

#### How to test
Test general use of Theia, with a focus on native integration like native menus, window trim, fonts. The code changes in this PR are purely related to the behavior of drag and drop into the IDE. Also have an eye on browser and back end log to make sure we have no new errors.

#### Breaking changes

- [ ] This PR introduces breaking changes and requires careful review. If yes, the breaking changes section in the [changelog](https://github.com/eclipse-theia/theia/blob/master/CHANGELOG.md) has been updated.

#### Attribution
Contributed on behalf of STMicroelectronics

#### Review checklist

- [ ] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #15871, opened 2025-06-23
Restore Extracted Widget to Main Window

jfaltermeier:
~~Note: I often see issues similar to https://github.com/eclipse-theia/theia/issues/15804 when testing this. I am currently investigating, but I don't think this is introduced with my changes in this PR.~~
~~Edit: I can reproduce this without my changes as well. I've commented on the issue.~~

#### What it does

This PR shows my current results from investigating #15594.
The issue did not specify all requirements, so we may discuss them here.

Current changes in this PR:

* extends `TheiaDockPanel` with options to disable DnD and to intercept tab close
* introduces `SecondaryWindowDockPanelWidget` to support dock panel layouts in secondary windows with DnD disabled
* adds area tracking (`previousArea`) for extractable widgets to support moving them back to their original location
* on secondary window close, widgets are moved back
* on tab close in secondary window, the widget is closed
* restricts extract widget command visibility to widgets not already in a secondary window
  * reasoning for now is to avoid complex chains when moving widgets back — there might not be a hard technical reason, we could also do this in a follow-up
* allows opening additional widgets in a secondary window (e.g. markdown preview editor via toolbar)
  * when closed, these widgets will be moved to the area of the initial widget that opened the secondary window
  * this relies on specifying a reference widget in the opening options
    * if missing or invalid, fallback is to open the widget on the main area
  * exception: widgets implementing `ExtractableWidget` with `isExtractable=false` are never moved to secondary windows

Open questions:

* Is the logic for moving widgets back okay? Currently:
  * on window close: move back
  * on tab close: close
* Depending on that, do we want to keep the code that intercepts tab close? It is not used right now but might be useful for downstream consumers of Theia

What do you think about this approach?

Peek 2025-05-27 10-38

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-view-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-view-widget.tsx
@@ -64,6 +64,7 @@ export class ChatViewWidget extends BaseWidget implements ExtractableWidget, Sta
     protected _state: ChatViewWidget.State = { locked: false, temporaryLocked: false };
     protected readonly onStateChangedEmitter = new Emitter<ChatViewWidget.State>();
 
+    isExtractable = true;
     secondaryWindow: Window | undefined;
 
     constructor(
@@ -237,10 +238,6 @@ export class ChatViewWidget extends BaseWidget implements ExtractableWidget, Sta
 
     get isLocked(): boolean {
         return !!this.state.locked;
-    }
-
-    get isExtractable(): boolean {
-        return this.secondaryWindow === undefined;
     }
 
     addContext(variable: AIVariableResolutionRequest): void {


tsmaeder:
@jfaltermeier I cannot reproduce the "disappearing menu" issue anymore after updating Electron. You might want to rebase and recheck once I merge https://github.com/eclipse-theia/theia/pull/15837

jfaltermeier:
> @jfaltermeier I cannot reproduce the "disappearing menu" issue anymore after updating Electron. You might want to rebase and recheck once I merge #15837

I rebased and commented here: https://github.com/eclipse-theia/theia/issues/15804#issuecomment-2996841482
Note sure if I am the only one with that issue. 
However there is no blocker for this PR.

tsmaeder:
@jfaltermeier cool stuff! What I don't really get, though: do you want to merge this now or are you just looking for feedback?

sdirix:
> * Is the logic for moving widgets back okay? Currently:
>   
>   * on window close: move back
>   * on tab close: close
> * Depending on that, do we want to keep the code that intercepts tab close? It is not used right now but might be useful for downstream consumers of Theia

I would answer this with yes and yes.

jfaltermeier:
@tsmaeder The idea was to collect feedback and then merge once everything is addressed. It is not urgent from my POV to get this in and it does not have to go into tomorrow's release afaik. 

If we are fine with

* moving back the widget on window close,
* closing widgets on tab close, and
* keeping the code to intercept tab close,

then this can be reviewed and tested now already.

tsmaeder:
One thing I noticed is that you can still drag the tabs in a secondary windows, highlighting seemingly random targets in the main window. I guess we'd have to turn off starting the drag operation instead of the "detach" request

jfaltermeier:
> One thing I noticed is that you can still drag the tabs in a secondary windows, highlighting seemingly random targets in the main window. I guess we'd have to turn off starting the drag operation instead of the "detach" request

I pushed an update that should fix this. By changing only the detach, I hope to keep all other drag-related features, like reordering. That's why I am currently trying the fix just there.

packages/core/src/browser/secondary-window-handler.ts
```diff
@@ -138,25 +183,24 @@ export class SecondaryWindowHandler {
             this.onWillAddWidgetEmitter.fire([widget, newWindow]);
 
             widget.secondaryWindow = newWindow;
-            const rootWidget = new SecondaryWindowRootWidget();
+            widget.previousArea = this.applicationShell.getAreaFor(widget);
+            const rootWidget: SecondaryWindowRootWidget = new SecondaryWindowDockPanelWidget(this.dockPanelFactory, this.dockPanelRendererFactory, this.onTabCloseRequested,
+                newWindow);
+            rootWidget.defaultRestoreArea = widget.previousArea;
```
sdirix, line 195: I don't understand what `defaultRestoreArea` is used for. It is set to the same as the initial widget. Why is it even needed then? The restore functionality will also check the widget first, so it will always be the widget's `previousArea`.

If it's about restoring additional widgets which were opened by the first widget, couldn't we check the first widget's `previousArea`?

packages/core/src/browser/shell/application-shell.ts
```diff
@@ -88,14 +88,16 @@ export interface DockPanelRendererFactory {
 export class DockPanelRenderer implements DockLayout.IRenderer {
     readonly tabBarClasses: string[] = [];
 
+    document?: Document | ShadowRoot;
```
sdirix, line 95: I would like to see a comment for this, e.g.
```suggestion
    /**
     * In case of DockPanels rendered in secondary windows, will be set
     * to the document of that window
     */
    document?: Document | ShadowRoot;
```

packages/core/src/browser/secondary-window-handler.ts
jfaltermeier: Yes, this is for widgets that were opened by the first one.
Since the first widget can be closed, we can't rely on looking it up from the first widget.
This then raises the question why we are not just using the `defaultRestoreArea`. My answer would be that if we implement drag and drop between extracted widgets at some point, it is still valuable to remember each extracted widget's `previousArea`, because restoring to the secondary window's `defaultRestoreArea` might then lead to different results.

jfaltermeier:
Restoring should work now in both Electron and browser apps.
I changed the drag and drop styling a bit, it now stays red if the detach case was triggered. The problem is that all this handling is in the tabbar and private, so it's hard to access and hook into for nicer handling.

jfaltermeier:
Regarding the monaco editor, it looks like `MonacoStandaloneThemeService` is not invoked the second time it is extracted, so the theme styles are missing. I'll check why this happens.

The Markdown preview issue could partly be related to [https://github.com/eclipse-theia/theia/issues/15804](https://github.com/eclipse-theia/theia/issues/15804) but not completely.

jfaltermeier:
The monaco issue was caused by a leak where the widget was not removed from the extracted widget list when being restored, which prevented the style from being applied on further extractions.

The missing extract action should be fixed by resetting `secondaryWindow` and `previousArea` before calling `addWidget` and `activate` on the main shell again. The command handler will now see the updated properties already and the visibility check should be correct.

The preview editor issues were caused by the `preview-frontend-module` not being registered as a secondaryWindow Theia extension. The CSS was missing.

I also added an implementation for `getTabBarFor` for secondary windows and some code that allows opening the dev tools for secondary windows, which was useful for debugging.

sdirix (commented):
Works much better now.

The only remaining issue is that the "Move View to Secondary Window" action still goes missing once the AI Chat view is reattached.

jfaltermeier:
I updated the `isExtractable` property of the AI chat window to true, since the command handler will check if there is a secondary window set already. Without this change `ExtractableWidget.is` stays false forever, which prevented updating/resetting the properties on restore.

sdirix (approved):
Works for me! Thank you for the great feature

PR #15873, opened 2025-06-23
[vscode] Support encoding/decoding from workspace and while opening text documents

### LINKED ISSUE #15559: [vscode] TextEncoding support and workspace.encode/decode functions
Since VS Code 1.100, there are additional functions `openTextDocument()`, `encode()` and `decode()` on workspace namespace. There is also `encoding` property on TextDocument

rschnekenbu:
#### What it does

Support the workspace openTextDocument, encode and decode methods with additional encoding options.

fix #15559

#### How to test

The following extension can be used:
- src: [encoding-sample-src.zip](https://github.com/user-attachments/files/20866079/encoding-sample-src.zip)
- zipped vsix: [encoding-sample-1.0.0.zip](https://github.com/user-attachments/files/20866067/encoding-sample-1.0.0.zip)

This extension provides:
- provide an action in the explorer and palette to open a text editor with a chosen encoding
- provide a contextual menu action to  reopen document with encoding in the current active text editor
- provide several actions to test encoding and decoding actions. These actions basically make a call to the API and displays a webview based on the result. Not much to see here, except that the encoding service is properly invoked on encode or decode methods.

Here is also a sample workspace with a set of files with different encodings. This can be used as a test for the Encoding Sample: open with encoding action or reopen editor with encoding actions
[workspace-encoding-sample.zip](https://github.com/user-attachments/files/20866064/workspace-encoding-sample.zip)

https://github.com/user-attachments/assets/25832620-9600-4d5a-9613-61720879d7f4

#### Follow-ups

None known

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/workspace-main.ts
+++ b/packages/plugin-ext/src/main/browser/workspace-main.ts
@@ -32,6 +32,9 @@ import { SearchInWorkspaceService } from '@theia/search-in-workspace/lib/browser
 import { FileStat } from '@theia/filesystem/lib/common/files';
 import { MonacoQuickInputService } from '@theia/monaco/lib/browser/monaco-quick-input-service';
 import { RequestService } from '@theia/core/shared/@theia/request';
+import { UTF16be, UTF16le, UTF8, UTF8_with_bom } from '@theia/core/lib/common/encodings';
+import { EncodingRegistry } from '@theia/core/lib/browser/encoding-registry';
+import { PreferenceService } from '@theia/core/lib/browser/preferences/preference-service';
 
 export class WorkspaceMainImpl implements WorkspaceMain, Disposable {
 
@@ -59,7 +62,11 @@ export class WorkspaceMainImpl implements WorkspaceMain, Disposable {
 
     private workspaceTrustService: WorkspaceTrustService;
 
+    private encodingRegistry: EncodingRegistry;
+
     private fsPreferences: FileSystemPreferences;
+
+    private preferenceService: PreferenceService;
 
     protected readonly toDispose = new DisposableCollection();
 
@@ -79,7 +86,9 @@ export class WorkspaceMainImpl implements WorkspaceMain, Disposable {
         this.workspaceService = container.get(WorkspaceService);
         this.canonicalUriService = container.get(CanonicalUriService);
         this.workspaceTrustService = container.get(WorkspaceTrustService);
+        this.encodingRegistry = container.get(EncodingRegistry);
         this.fsPreferences = container.get(FileSystemPreferences);
+        this.preferenceService = container.get(PreferenceService);
 
         this.processWorkspaceFoldersChanged(this.workspaceService.tryGetRoots().map(root => root.resource.toString()));
         this.toDispose.push(this.workspaceService.onWorkspaceChanged(roots => {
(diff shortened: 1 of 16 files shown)


packages/plugin-ext/src/main/browser/workspace-main.ts
```diff
@@ -318,6 +327,58 @@ export class WorkspaceMainImpl implements WorkspaceMain, Disposable {
...
+    }
+
+    async $resolveEncoding(resource: UriComponents | undefined, options?: { encoding?: string }): Promise<{ encoding: string; hasBOM: boolean }> {
+        let encoding: string;
+        if (resource) {
+            encoding = await this.encodingRegistry.getEncodingForResource(URI.fromComponents(resource), options?.encoding);
+        } else {
+            encoding = options?.encoding ?? UTF8;
+        }
+        // see https://github.com/microsoft/vscode/blob/118f9ecd71a8f101b71ae19e3bf44802aa173209/src/vs/workbench/services/textfile/browser/textFileService.ts#L806
```
tsmaeder, line 351: Does this file contain enough copied code to warrant adding a copyright by MS notice?

packages/plugin-ext/src/main/browser/workspace-main.ts
```diff
@@ -318,6 +327,58 @@ export class WorkspaceMainImpl implements WorkspaceMain, Disposable {
...
+        const hasBOM = encoding === UTF16be || encoding === UTF16le || encoding === UTF8_with_bom;
+
+        return { encoding, hasBOM };
+    }
+
+    async $validateDetectedEncoding(uri: UriComponents | undefined, detectedEncoding: string | undefined, options: { encoding: string } | undefined): Promise<string> {
```
tsmaeder: When I read `validate...`, I expect a method that returns an error if the parameters are not valid. Why not `$getValidEncoding`?

packages/plugin-ext/src/plugin/document-data.ts
```diff
@@ -74,6 +76,10 @@ export class DocumentDataExt {
         ok(!this.disposed);
         this.languageId = langId;
     }
+    acceptEncoding(encoding: string): void {
```
tsmaeder, line 79: This is never called. Do we need it?

packages/plugin-ext/src/plugin/documents.ts
```diff
@@ -238,13 +238,14 @@ export class DocumentsExtImpl implements DocumentsExt {
         }
     }
 
-    async openDocument(uri: URI): Promise<DocumentDataExt | undefined> {
+    async openDocument(uri: URI, options?: { language?: string; content?: string; encoding?: string }): Promise<DocumentDataExt | undefined> {
+        // If we have the document cached and no encoding options are provided,
+        // we should just return current document
         const cached = this.editorsAndDocuments.getDocument(uri.toString());
-        if (cached) {
+        if (cached && !options?.encoding) {
```
tsmaeder: Why does passing an encoding invalidate the cache? Maybe we read the content with the same encoding before?

packages/plugin-ext/src/plugin/workspace.ts
```diff
@@ -80,6 +84,7 @@ export class WorkspaceExtImpl implements WorkspaceExt {
     initialize(): void {
         this.proxy = this.rpc.getProxy(Ext.WORKSPACE_MAIN);
         this.logger = new PluginLogger(this.rpc, 'workspace');
+        this.encodingService = new EncodingService();
```
tsmaeder: Is this not injectable?

tsmaeder (commented):
Code looks reasonable +/- a couple of questions.

tsmaeder (dismissed):

tsmaeder:
@rschnekenbu This seems to work O.K. for me, but could you try to get the tests to pass, please?

rschnekenbu:
@tsmaeder , I updated the PR according for your comments, thank for the review!
I do not get which tests I am asked to fix. The current failures do not seem to be caused by my contribution.

packages/plugin-ext/src/plugin/workspace.ts
rschnekenbu: I added some bindings for the EncodingService, and used the injection rather than creating  a new one.

packages/plugin-ext/src/plugin/documents.ts
rschnekenbu: Checked for the current encoding of the cached one. This may ease the reuse of cached version.

packages/plugin-ext/src/plugin/document-data.ts
rschnekenbu: It is now called when encoding is changed. Thanks for catching this piece of missing code!

packages/plugin-ext/src/main/browser/workspace-main.ts
rschnekenbu: I added a copyright + comment that indicates where some parts of the code comes from.

tsmaeder:
@rschnekenbu could you just quickly rebase on master and push? I'd like to get some data on whether the electron update improved the failing tests.

rschnekenbu:
Rebased on top of master

PR #15879, opened 2025-06-24
Add fall back to default prompts if selected does not exist

JonasHelming:
#### What it does

Prompt service return the default prompt if the selected variant does not exist.
This is important as we refactor prompt ids with the upcoming release and if users selected variants, we need a fall-back.

#### How to test

- Create a variant by creating a file that starts with the variant set id in the prompt template folder, e.g. "coder-prompt-fooba.md"
- Select this variant
- Delete the file
- Talk to the agent and observe the default prompt is used

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/common/prompt-service.ts
+++ b/packages/ai-core/src/common/prompt-service.ts
@@ -544,7 +544,7 @@ export class PromptServiceImpl implements PromptService {
             const variantIds = this.getVariantIds(fragmentId);
             if (!variantIds.includes(defaultVariantId)) {
                 this.logger.error(`Default variant '${defaultVariantId}' for prompt set '${fragmentId}' does not exist.`);
-                return undefined; // Default variant doesn't exist
+                return undefined;
             }
             return defaultVariantId;
         }
--- a/packages/ai-core/src/common/prompt-service.ts
+++ b/packages/ai-core/src/common/prompt-service.ts
@@ -14,7 +14,7 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 
-import { Event, Emitter, URI } from '@theia/core';
+import { Event, Emitter, URI, ILogger } from '@theia/core';
 import { inject, injectable, optional, postConstruct } from '@theia/core/shared/inversify';
 import { AIVariableArg, AIVariableContext, AIVariableService, createAIResolveVariableCache, ResolvedAIVariable } from './variable-service';
 import { ToolInvocationRegistry } from './tool-invocation-registry';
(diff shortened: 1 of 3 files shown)


packages/ai-core/src/common/prompt-service.ts
```diff
@@ -367,8 +367,22 @@ export interface PromptService {
      * @param promptVariantSetId The prompt variant set id
      * @returns The selected variant ID or the main ID if no variant is selected
      */
+    /**
```
sgraband: Comment above this needs to be removed.

packages/ai-core/src/common/prompt-service.ts
```diff
@@ -505,17 +521,49 @@ export class PromptServiceImpl implements PromptService {
...
+        // First try to get the selected variant from settings
+        const selectedVariantId = await this.getSelectedVariantId(fragmentId);
+
+        // Check if the selected variant actually exists
+        if (selectedVariantId) {
+            const variantIds = this.getVariantIds(fragmentId);
+            if (!variantIds.includes(selectedVariantId)) {
+                this.logger.warn(`Selected variant '${selectedVariantId}' for prompt set '${fragmentId}' does not exist. Falling back to default variant.`);
+            } else {
+                return selectedVariantId; // Selected variant exists, return it
```
sgraband:
```suggestion
                return selectedVariantId;
```

packages/ai-core/src/common/prompt-service.ts
```diff
@@ -505,17 +521,49 @@ export class PromptServiceImpl implements PromptService {
...
+            }
+        }
+
+        // Fall back to default variant
+        const defaultVariantId = this.getDefaultVariantId(fragmentId);
+        if (defaultVariantId) {
+            const variantIds = this.getVariantIds(fragmentId);
+            if (!variantIds.includes(defaultVariantId)) {
+                this.logger.error(`Default variant '${defaultVariantId}' for prompt set '${fragmentId}' does not exist.`);
+                return undefined; // Default variant doesn't exist
```
sgraband:
```suggestion
                return undefined;
```

packages/ai-core/src/common/prompt-service.ts
```diff
@@ -505,17 +521,49 @@ export class PromptServiceImpl implements PromptService {
                 }
             }
         }
-        return this.getDefaultVariantId(fragmentId);
+        return undefined;
+    }
+
+    async getEffectiveVariantId(fragmentId: string): Promise<string | undefined> {
+        // First try to get the selected variant from settings
```
sgraband:
```suggestion
```

JonasHelming:
> Works great! I just have a few inline comments as well as another suggestion: imho this needs to be reflected in the Agent & Prompt Fragments view. In the agent view it is probably sufficient to just adapt the `The selected variant is not longer available`, but for the prompt fragment view we should probably render the `selected` badge for the effective variant and not the selected one. WDYT?

I added the "The selected..." text. About the prompt fragment view:

I agree. I think we should just add a warning as a text: "The selected variant is not available, the default will be used". IMHO there is not other case. Anyways, I propose to handle this as a follow up due to the release schedule. If you are fine, could you create the follow-up please?

sgraband:
Follow up: https://github.com/eclipse-theia/theia/issues/15887

Issue #15229, opened 2025-03-19
search boxes in "search in workspace" resize to imaginary tooltip ignoring white-space: nowrap;


Alexander-Taran:
### Steps to Reproduce:

1. resize left panel to something small
2. focus one of search boxes in siw
3. observe textarea grow in height

### Additional Information
![Image](https://github.com/user-attachments/assets/73f457b9-2db8-46f6-865f-b3f191fab554)

![Image](https://github.com/user-attachments/assets/4a9e0424-8cc1-451e-90a4-b1e20cbcdd95)

```css
.theia-input::placeholder {
    color: var(--theia-input-placeholderForeground);
    overflow: hidden;
    text-overflow: ellipsis;
    white-space: nowrap;
}
```

Looks like react-textarea-autosize calculates height based on `content || placeholder`

Disabling nowrap in style makes visible sense. But still not what one would want to see.

sdirix:
Thanks for the report! Do you want to open a PR which improves the search in workspace with a more reasonable behavior?

Alexander-Taran:
@sdirix 
Can't say for sure.
I think "css only" with data-attribute syncing  to value is the best option. 
https://css-tricks.com/auto-growing-inputs-textareas/

So placeholder is not taken into account.

kachurun:
I fixed that. I'll make a PR in an hour

kachurun:
Before patch
https://github.com/user-attachments/assets/4c4daa19-7cab-4f84-abd7-f02b1f895da6

After patch

https://github.com/user-attachments/assets/84e1803e-771f-493e-848c-7c1d4f1abee2

PR #15881, opened 2025-06-25
Fix: bug with incorrect height calculation of the search field in "search-in-workspace" (#15229)

kachurun:
#### What it does
Fixes https://github.com/eclipse-theia/theia/issues/15229

Bug caused by the `react-textarea-autosize` component, which sets the height of the search input textarea based on the placeholder length and ignores the fact that this placeholder has a white-space: nowrap style. So the fix is pretty simple: just limit maxRows for this component while the field is empty and the placeholder is shown. I also added forceUpdate just to be sure that maxRows updates immediately (without it, the textarea size updates only after the search task is finished).

#### How to test
Check bug:
https://github.com/eclipse-theia/theia/issues/15229

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/search-in-workspace/src/browser/components/search-in-workspace-textarea.tsx
+++ b/packages/search-in-workspace/src/browser/components/search-in-workspace-textarea.tsx
@@ -80,6 +80,10 @@ export class SearchInWorkspaceTextArea extends React.Component<TextareaAttribute
             e.preventDefault();
         }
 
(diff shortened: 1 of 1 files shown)


ndoschek (approved):
Thanks @kachurun for the fix!  
Looks good to me and works perfectly (tested on both browser and Electron) 🙌

PR #15885, opened 2025-06-25
Refine inline Ask AI chat input appearance

ndoschek:
#### What it does

- Align styling to match the chat input in the main chat view, ensuring consistent appearance
- Do not show top arrow for a cleaner UI
- Adjust the zone widget layout dynamically based on the chat input height
- Display pinned agents to allow agent selection in the inline input

#### How to test

- Open the Ask AI inline chat widget and observe that the styling matches the chat input in the AI Chat View
- Observe that the editor zone widget is more in line with other of that kind, like the git inline quick diff or the conditional breakpoint editing
- If multiple lines are entered, the editor zone widget grows dynamically

#### Follow-ups

x

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
@@ -162,6 +162,7 @@ export class AIChatInputWidget extends ReactWidget {
         this.toDispose.push(this.aiActivationService.onDidChangeActiveStatus(() => {
             this.setEnabled(this.aiActivationService.isActive);
         }));
+        this.toDispose.push(this.onDidResizeEmitter);
         this.setEnabled(this.aiActivationService.isActive);
         this.update();
     }


packages/ai-chat-ui/src/browser/chat-input-widget.tsx
```diff
@@ -151,6 +151,9 @@ export class AIChatInputWidget extends ReactWidget {
         this.update();
     }
 
+    protected onDidResizeEmitter = new Emitter<void>();
```
rschnekenbu, line 154: should this be added to the toDispose collection in the init()?

rschnekenbu (changes requested):
Code looks good in general except my comment  on emitter.
Behavior is also good (tested on electron), changing theme, font size, zoom, etc.

packages/ai-chat-ui/src/browser/chat-input-widget.tsx
ndoschek: Great catch, thanks! I added it now!

rschnekenbu:
@ndoschek , that was a quick fix :+1: I retriggered the failed license check, but that should not be a problem for this PR. No change in dependencies was pushed.

PR #15890, opened 2025-06-26
fix store view ,fix #15201

zhangzejun123:
I think that the previous modification set the model to null after the editor was hidden, so nothing would be done in editor.saveViewState(). However, savedViewState was set when the editor was hidden, so we only need to read this value. Do you think this is okay?

PR #15898, opened 2025-06-27
fix:IDE cannot fully restore to the previous state after restarting.

### LINKED ISSUE #15201: The IDE cannot fully restore to the previous state after restarting.
### Bug Description:

During the code editing process, multiple tab pages are opened, and the mouse pointers are at different positions. After closing the Theia IDE and reopening it, only the tab page that is displayed by default can restore the pointer position correctly. The mouse pointers of other tab pages all return to the top. Compared with editors such as VS Code and PyCharm, which can restore the states of all tab pages correctly, this is determined to be a bug.

### Steps to Reproduce:

1.Open three tab pages: extensio.ts(25:33), util.d.ts(14:20), and wasi.d.ts(52:17).
2.Close the Theia IDE and reopen it.
3.Observe that except for the extension.ts tab page where the pointer position is correct, the mouse pointers of the other two tab pages are at the top.

### Additional Information

- Operating System: windows10
- Theia Version: 1.59.1

### Attachment： 

Image

zhangzejun123:
#### What it does
Fixes #15201
I think that the previous modification set the model to null after the editor was hidden, so nothing would be done in editor.saveViewState(). However, savedViewState was set when the editor was hidden, so we only need to read this value.So I modified the storeViewState and restoreViewState functions in monaco-editor

#### How to test

1.I compiled theia-ide application locally, and it was verified that this is effective and can solve the problem #15201
2.For example, the gif image operation in #15201 will get the expected result

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/monaco/src/browser/monaco-editor.ts
+++ b/packages/monaco/src/browser/monaco-editor.ts
@@ -611,11 +611,16 @@ export class MonacoEditor extends MonacoEditorServices implements TextEditor {
     }
 
     storeViewState(): object {
-        return this.editor.saveViewState()!;
+        const state = this.editor.saveViewState();
+        if (state) {
+            this.savedViewState = state;
+        }
+        return this.savedViewState!;
     }
 
     restoreViewState(state: monaco.editor.ICodeEditorViewState): void {
         this.editor.restoreViewState(state);
+        this.savedViewState = state;
     }
 
     /* `true` because it is derived from an URI during the instantiation */


zhangzejun123:
reopen from #15890

zhangzejun123:
> reopen from #15890

@tsmaeder  Hello, can you review this PR again? I saw that no one has reviewed this PR. In fact, after the modification, I ran it locally for a while and it confirmed that this PR is indeed valid. However, I need your advice on whether the code is compliant. Thank you.

tsmaeder (approved):

PR #15928, opened 2025-07-01
Fix stale tree selection

colin-grant-work:
#### What it does

Fixes an issue where global selection, and the behavior of commands that point to global selection, could become stale if the owner of the selection failed to update it.

#### How to test

The steps are fairly specific:

1. Open two files.
2. Focus the editor of the second file
3. Select something else in the explorer tree (due to a separate bug - fix hopefully forthcoming - I'd recommend a folder)
4. Click on the exit `x` for the currently focused editor _without focusing the editor_.
5. The tree selection will jump to the first file because its editor is now focused.
6. Right click and trigger the `Rename` or `Delete` commands.
7.

_master_ Observe that the command reflects your previous selection, not the file you clicked on. See video.
_this branch_ Observe that the command reflects the current selection, and the file you clicked on.

https://github.com/user-attachments/assets/05e74d69-9db8-46a2-8f00-016e9a128c15

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


packages/core/src/browser/tree/tree-widget.tsx
```diff
@@ -345,6 +345,10 @@ export class TreeWidget extends ReactWidget implements StatefulWidget {
         }));
     }
 
+    protected shouldUpdateGlobalSelection(): boolean {
```
martin-fleck-at, line 360: nit: The name suggests that this acts as a guard for the `updateGlobalSelection` method. But that is not as there is the `focusin` listener  that also updates the global selection but has a different condition. However, I'm also not sure what a better name for that could be, so maybe it is fine for now.

colin-grant-work: We could reorganize the code so that instead of checking where we currently check, we check in `updateGlobalSelection` and just call that more unconditionally, but that would affect any downstream implementation that assumed that `updateGlobalSelection` acts unconditionally and is only called when we think we want to do the work.

martin-fleck-at: Yeah, no, I agree, the `updateGlobalSelection` should just do what it says unconditionally. Could we maybe simple pull out the event handler implementations altogether to at least make them overridable more easily, e.g., `handleModelSelectionChange`, `handleFocusChange`, `handleFocusInChange` or something similar? That way, if someone is unhappy with the `focusIn` condition, it would be much easier to adapt.

packages/core/src/browser/tree/tree-widget.tsx
```diff
@@ -315,38 +316,51 @@ export class TreeWidget extends ReactWidget implements StatefulWidget {
...
+        this.toDispose.push(Disposable.create(this.handleGlobalSelectionOnDisposal.bind(this)));
+    }
+
+    protected handleGlobalSelectionOnModelSelectionChange(): void {
+        if (this.shouldUpdateGlobalSelection()) {
+            this.updateGlobalSelection();
+        }
+    }
+
+    protected handleGlobalSelecitonOnFocusServiceFocusChange(focus: SelectableTreeNode | undefined): void {
```
martin-fleck-at: nit:
```suggestion
    protected handleGlobalSelectionOnFocusServiceFocusChange(focus: SelectableTreeNode | undefined): void {
```

Alexander-Taran:
This PR did have a breaking change though.
Now global selection updates even when widget is not active.
And in some cases when an update in tree produces clearSelection call and then restores selection
property panel receives [] as selection when previously it did not, because tree widget did not contain activeElement

martin-fleck-at:
@Alexander-Taran Good catch, the behavior has indeed changed in this regard!

colin-grant-work:
> property panel receives [] as selection when previously it did not, because tree widget did not contain activeElement

@Alexander-Taran, it's true that the property panel will see selections it previously didn't, but I did a little bit of testing, and it looks like if you clear and then restore selection, the property panel sees both the clearance and the restoration, even if the source of selection isn't focused, so its state continues to reflect the selection. In your case, do you see cases where the panel fails to reflect the later restoration of selection?

PR #15936, opened 2025-07-02
15822: improve ChatUI auto-scroll / scroll-lock consistency

CamilleLetavernier:
#### What it does

- improve consistency of auto-scroll / scroll-lock in the Chat View
  - Always enable (temporary) scroll-lock when user scrolls up
  - Disable (temporary) scroll-lock when user scrolls back down
  - Ignore scroll-events related to content shrinking (layout updates, e.g. end of code block), to avoid scroll jitter
- add "scroll to bottom" button to the Chat View

This reduces scroll-jitter in all identified cases, and makes scrolling overall more fluid and more predictable

#### How to test

Auto-scroll must work as follows:

- if auto-scroll is enabled and agent sends new tokens, scroll to the bottom
- if the user scrolls up, temporarily disable auto-scroll (this is roughly equivalent to enabling scroll lock)
- if the user scrolls back to the bottom (or close to the bottom), restore auto-scroll behavior
- when the view is not scrolled to the bottom, show "Jump to latest message" (scroll to bottom) button

Note: if scroll-lock is disabled ("Turn Auto-scroll off"), the behavior should be unchanged: the chat view should never scroll without explicit user interaction.

#### Follow-ups

No specific follow-ups at this point, but we could consider adding more specific auto-scroll behaviors (e.g. "scroll to the beginning of the current AI Response" instead of "Always jump to the end", so user can comfortably read the current answer while it's still printing). This would significantly reduce the amount of auto-scroll events.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-tree-view/chat-view-tree-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-tree-view/chat-view-tree-widget.tsx
@@ -93,6 +93,7 @@ export interface ChatWelcomeMessageProvider {
 
 @injectable()
 export class ChatViewTreeWidget extends TreeWidget {
+
     static readonly ID = 'chat-tree-widget';
     static readonly CONTEXT_MENU = ['chat-tree-context-menu'];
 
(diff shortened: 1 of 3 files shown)


CamilleLetavernier:
Scroll example:

https://github.com/user-attachments/assets/c10cd74e-e24f-42a0-aeed-5009e851eeee

PR #15937, opened 2025-07-02
Use node 22 for remote containers

sgraband:
#### What it does
This is needed because otherwise executing something on the container will fail. Also make sure to report the error channel instead of the stdout channel if starting the container fails. This will require a recompiling of the native dependencies.

#### How to test
1. Start a devcontainer and see that it now works.

On master, starting a devcontainer will not work and fail with `Could not start remote system:`.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.gitpod.dockerfile
+++ b/.gitpod.dockerfile
@@ -17,7 +17,7 @@ RUN sudo apt-get update \
 
 ENV SHELL=/bin/bash
 
-ENV NODE_VERSION="18.17.0"
+ENV NODE_VERSION="22.15.1"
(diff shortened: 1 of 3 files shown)


ndoschek (approved):
Thanks @sgraband for this fix!
I tested the [glsp-website-source](https://github.com/eclipse-glsp/glsp-website-source/) repository to open in the Dev Container and it works again as expected! 👍

rschnekenbu (approved):
LGTM! I only looked at the code, feature was tested by @ndoschek

PR #15951, opened 2025-07-03
Update ToolProvider to handle cancellation

eneufeld:
#### What it does

- Updated ToolProviders to cancel execution
- Added tests

#### How to test

Let universal call a long running tool, eg delegate to agent.
Press the cancel button in the chat view. see that delegation stops.
Previously this continued.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/browser/agent-delegation-tool.ts
+++ b/packages/ai-chat/src/browser/agent-delegation-tool.ts
@@ -25,6 +25,7 @@ import {
     MutableChatRequestModel,
     MutableChatModel,
     ChatSession,
+    ChatRequestInvocation,
 } from '../common';
 import { DelegationResponseContent } from './delegation-response-content';
 
@@ -71,6 +72,10 @@ export class AgentDelegationTool implements ToolProvider {
         arg_string: string,
         ctx: MutableChatRequestModel
     ): Promise<string> {
+        if (ctx.response.cancellationToken.isCancellationRequested) {
+            return 'Operation cancelled by user';
+        }
+
         try {
             const args = JSON.parse(arg_string);
             const { agentId, prompt } = args;
@@ -125,13 +130,27 @@ export class AgentDelegationTool implements ToolProvider {
                 text: prompt,
             };
 
-            let response;
+            let response: ChatRequestInvocation | undefined;
             try {
+                if (ctx?.response?.cancellationToken?.isCancellationRequested) {
+                    return 'Operation cancelled by user';
+                }
+
                 const chatService = this.getChatService();
                 response = await chatService.sendRequest(
                     newSession.id,
                     chatRequest
                 );
+
+                if (ctx?.response?.cancellationToken) {
+                    ctx.response.cancellationToken.onCancellationRequested(
+                        async () => {
+                            if (response) {
+                                ((await response?.requestCompleted) as MutableChatRequestModel).cancel();
+                            }
+                        }
+                    );
+                }
             } catch (sendError) {
                 const errorMsg = `Failed to send request to agent '${agentId}': ${sendError instanceof Error ? sendError.message : sendError}`;
                 console.error(errorMsg, sendError);
@@ -157,6 +176,12 @@ export class AgentDelegationTool implements ToolProvider {
                     // Return the raw text to the top-level Agent, as a tool result
                     return stringResult;
                 } catch (completionError) {
+                    if (
+                        completionError.message &&
+                        completionError.message.includes('cancelled')
+                    ) {
+                        return 'Operation cancelled by user';
+                    }
                     const errorMsg = `Failed to complete response from agent '${agentId}': ${completionError instanceof Error ? completionError.message : completionError}`;
                     console.error(errorMsg, completionError);
                     return errorMsg;
--- a/packages/ai-ide/src/browser/context-functions.spec.ts
+++ b/packages/ai-ide/src/browser/context-functions.spec.ts
@@ -1,0 +1,102 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+import { enableJSDOM } from '@theia/core/lib/browser/test/jsdom';
+let disableJSDOM = enableJSDOM();
+import { FrontendApplicationConfigProvider } from '@theia/core/lib/browser/frontend-application-config-provider';
+FrontendApplicationConfigProvider.set({});
+import { expect } from 'chai';
+import { ListChatContext, ResolveChatContext, AddFileToChatContext } from './context-functions';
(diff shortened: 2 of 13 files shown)


packages/ai-chat/src/browser/agent-delegation-tool.ts
```diff
@@ -125,15 +143,34 @@ export class AgentDelegationTool implements ToolProvider {
                 text: prompt,
             };
 
-            let response;
+            // Properly type the response variable
```
ndoschek:
```suggestion
```

packages/ai-chat/src/browser/agent-delegation-tool.ts
```diff
@@ -125,15 +143,34 @@ export class AgentDelegationTool implements ToolProvider {
...
+            let response: ChatRequestInvocation | undefined;
             try {
+                // Check for cancellation again before sending request
```
ndoschek:
```suggestion
```

packages/ai-chat/src/browser/agent-delegation-tool.ts
```diff
@@ -125,15 +143,34 @@ export class AgentDelegationTool implements ToolProvider {
...
+                    return 'Operation cancelled by user';
+                }
+
                 const chatService = this.getChatService();
                 response = await chatService.sendRequest(
                     newSession.id,
                     chatRequest
                 );
+
+                // Register cancellation handler
```
ndoschek:
```suggestion
```

packages/ai-chat/src/browser/agent-delegation-tool.ts
```diff
@@ -125,15 +143,34 @@ export class AgentDelegationTool implements ToolProvider {
...
+                if (ctx?.response?.cancellationToken) {
+                    ctx.response.cancellationToken.onCancellationRequested(
+                        async () => {
+                            // Try to cancel the delegated request
```
ndoschek:
```suggestion
```

packages/ai-ide/src/browser/context-functions.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+import { FrontendApplicationConfigProvider } from '@theia/core/lib/browser/frontend-application-config-provider';
+FrontendApplicationConfigProvider.set({});
+import { expect } from 'chai';
+import { ListChatContext, ResolveChatContext, AddFileToChatContext } from './context-functions';
+import { CancellationTokenSource } from '@theia/core';
+import { ChatContextManager, MutableChatModel, MutableChatRequestModel, MutableChatResponseModel } from '@theia/ai-chat';
+import { fail } from 'assert';
+import { ResolvedAIContextVariable } from '@theia/ai-core';
+disableJSDOM();
+// Tests for context functions with cancellation handling
```
ndoschek:
```suggestion
```

ndoschek (commented):
Thanks, @eneufeld. I mostly tested this with the agent delegation feature, and it looks good to me. I would just suggest removing a few superfluous comments (see inline).

Besides, I noticed quite a few changes that are essentially only code formatting. I know the code-formatting configuration is sometimes off (I also opened #15959 today), but usually for TypeScript files the discrepancy to the existing code base isn't that large (at least in my case?).
This PR contains many pure formatting changes; if you don't mind, could you revert those if you find the time? TIA

eneufeld:
@ndoschek I updated the pr

packages/ai-ide/src/browser/context-functions.ts
```diff
@@ -16,6 +16,7 @@
 
 import { MutableChatRequestModel } from '@theia/ai-chat';
 import { ToolProvider, ToolRequest } from '@theia/ai-core';
+
```
ndoschek:
```suggestion
```

packages/ai-chat/src/browser/agent-delegation-tool.ts
```diff
@@ -71,6 +72,10 @@ export class AgentDelegationTool implements ToolProvider {
         arg_string: string,
         ctx: MutableChatRequestModel
     ): Promise<string> {
+        if (ctx?.response?.cancellationToken?.isCancellationRequested) {
```
ndoschek:
```suggestion
        if (ctx.response.isCanceled) {
```

ndoschek (commented):
Hi @eneufeld, thank you for the update!

I had another look, and I noticed, the check for cancellation could be simplified by using the `.isCanceled` getter from the response model, wdyt?

Also, if we cancel a run, the `AgentCompletionNotificationService` notifies us that the task was completed, which is not entirely correct IMO.
(I am using the `ai-features.notifications.default`: `message' setting).
We could show the completion notification only if it was not canceled by the user for example. But we can also extract a follow up for that one, that's fine by me.

PS: unfortunately I cannot resolve my own comments right now, but the previous comments are resolved from my POV.

packages/ai-chat/src/browser/agent-delegation-tool.ts
eneufeld: we can remove the optional chaining but we should not change to isCanceled, we never use it in the tools but rely on the cancellationToken state

eneufeld:
> Also, if we cancel a run, the `AgentCompletionNotificationService` notifies us that the task was completed, which is not entirely correct IMO. (I am using the `ai-features.notifications.default`: `message' setting). We could show the completion notification only if it was not canceled by the user for example. But we can also extract a follow up for that one, that's fine by me.

this is fixed here:
https://github.com/eclipse-theia/theia/pull/15956

packages/ai-ide/src/browser/workspace-search-provider.spec.ts
```diff
@@ -1,5 +1,5 @@
 // *****************************************************************************
-// Copyright (C) 2025 EclipseSource GmbH.
+// Copyright (C) 2024 EclipseSource GmbH.
```
sdirix: seems like an accidental change

packages/ai-ide/src/browser/workspace-task-provider.spec.ts
```diff
@@ -0,0 +1,125 @@
+// *****************************************************************************
+// Copyright (C) 2024 EclipseSource GmbH.
```
sdirix:
```suggestion
// Copyright (C) 2025 EclipseSource GmbH.
```

packages/ai-ide/src/browser/workspace-functions.ts
```diff
@@ -147,7 +147,17 @@ export class GetWorkspaceDirectoryStructure implements ToolProvider {
                 type: 'object',
                 properties: {}
             },
-            handler: () => this.getDirectoryStructure()
+            handler: (_: string, ctx: MutableChatRequestModel) => {
+                const cancellationToken =
```
sdirix: Sometimes we deeply check the access of the cancellationToken and sometimes we don't. Can you unify this?

JonasHelming:
@sdirix Ping

packages/ai-ide/src/browser/workspace-functions.ts
```diff
@@ -285,7 +332,11 @@ export class GetWorkspaceFileList implements ToolProvider {
     @inject(WorkspaceFunctionScope)
     protected workspaceScope: WorkspaceFunctionScope;
 
-    async getProjectFileList(path?: string): Promise<string[]> {
+    async getProjectFileList(path?: string, cancellationToken?: CancellationToken): Promise<string[]> {
+        if (cancellationToken?.isCancellationRequested) {
+            return ['Error: Operation cancelled by user'];
```
sdirix: This looks like a very weird return value.  What do we do for tools which only return a number?

Maybe the tools should be typed to optionally return error object / string or we should throw errors instead? What do you think?

sdirix (commented):
Works for me. Especially canceling the delegation is a great feature.

I left some comments. Please have a look.

eneufeld:
@sdirix I updated this

sdirix (approved):
Works for me!

PR #15952, opened 2025-07-03
Use the agent name that requested the tool call in the change set title

### LINKED ISSUE #15950: File change set functions always set the title to "by Coder" even if it comes from a different agent
### Bug Description:

The file change set functions like `suggestFileContent` in file-changeset-functions.ts always set the change set title to `'Changes proposed by Coder'` instead of using the current agent's name that called the tool.

### Steps to Reproduce:

1. Create a custom agent with a prompt like this which will suggest file content:
```
You are an agent that writes poem files for the user in a workspace.
Follow the instructions of the user to write a poem and suggest the file content using this tool: ~{suggestFileContent}
Poem files should always be at the root of the workspace and use the .txt file extension.
## Current Context
Some files and other pieces of data may have been added by the user to the context of the chat. If any have, the details can be found below.
{{contextDetails}}
```
2. Chat with the agent to suggest file content.
3. The change set title reads 'Changes proposed by Coder' instead of the current agent's name.

Image

### Additional Information

- Operating System: Windows
- Theia Version: 1.63

Hanksha:
#### What it does

This PR fixes the file change set functions so they set the change set title with the agent's name that requested the tool call.

Closes #15950 

#### How to test

- Chat with agents that uses file change set functions, 
- Prompt them so that they suggest file changes.
- Check that the change set title includes the current agent name.

feature

#### Attribution

Contributed on behalf of Lonti.com Pty Lt

#### Review checklist

#### Reminder for reviewers


Hanksha:
I don't know how to fix the error happening when the @theia/ai-ide tests are running:

```
> @theia/ai-ide:test
@theia/ai-ide: > @theia/ai-ide@1.63.0 test
@theia/ai-ide: > theiaext test
@theia/ai-ide: $ nyc mocha --config ../../configs/mocharc.yml "./lib/**/*.*spec.js"
@theia/ai-ide:  Exception during run: TypeError: Cannot read properties of undefined (reading 'VIEW_CATEGORY')
@theia/ai-ide:     at Object.<anonymous> (/Users/runner/work/theia/theia/packages/core/lib/browser/shell/shell-layout-restorer.js:55:61)
@theia/ai-ide:     at Module._compile (node:internal/modules/cjs/loader:1529:14)
@theia/ai-ide:     at Module.replacementCompile (/Users/runner/work/theia/theia/node_modules/append-transform/index.js:60:13)
@theia/ai-ide:     at Module._extensions..js (node:internal/modules/cjs/loader:1613:10)
@theia/ai-ide:     at Object.<anonymous> (/Users/runner/work/theia/theia/node_modules/append-transform/index.js:64:4)
```

I can reproduce it locally but I can't figure what in my changes is causing the error.

eneufeld (commented):
Thank you for the proposed fix.
I think there is a bigger issue with the changeset labeling. The ChangeSet assumes that there is only one provider of changes, which was true until the agent delegation was introduced.
Now we can have multiple different agents contribute changes to the changeset.
This leads us to your proposed fix with using a dedicated service which provides the title string.
This solution will break as soon as we need to support the resolution of multiple agents. 
I suggest to inline the ChatAgentService into the tools.
Each toolCall has a context which is of type MutableChatRequestModel and this class has an agentId property which we can use to resolve the agent name

Hanksha:
@eneufeld That was already a problem before the agent delegation was introduced. In the same chat session you can call multiple agents "manually" sequentially. The last agent calling the tool will override the change set title even if it only proposed one of the elements in the change set.

bug

My fix still works with agent delegation but has the same issue, the last delegated agent call overrides the title. So if a prompt ends up generating multiple delegate to agent tool call which modify the change set, it's the last one that will set the final title.

Do you propose that the change set title contains all the names of the agents that contributed to the change set? e.g. "Changes proposed by Coder, Agent2, Agent3"

eneufeld:
You're totally right, the problem was there even before the agent delegation.
I'm not sure what the best way to show the changes would be.
A simple enumeration of all agents that contributed changed could be a solution.
We could also introduce sections and group the changes by the agent that did them. This might lead to issues in the case where one file is changed by different agents though.

@planger @sdirix @ndoschek @JonasHelming ideas?

planger:
You're absolutely right. This came from the fact that only Coder was using the change set and we never generalized the change set title accordingly.

My two cents: I wouldn't append the "by <agent>" at all. To me the change set is a place that collects all changes from a chat session into a central place for the user. Imho it isn't important to show the agent there (or which agent contributed which change).

Agents should be able to set the change set title to something specific, if they want, and we have a generic default name. So an agent can check, is there already a title set or changes proposed, if no they can specify something specific if they want, otherwise they just leave the title as is.

WDYT?

sdirix:
I agree, with @planger, I don't think we need the `by` for now. If the delegation use case takes off, many agents are in play, and there is value in knowing which change came from which agent, then we can still introduce something like @eneufeld suggested, i.e. some marker for each change from which agent it originates.

Hanksha:
@planger @sdirix I think that's fine too. I could change this PR to return the default nls string "Changes proposed" and leave the provider interface in place for extensions that want to override the behavior?

Hanksha:
I've updated the changes accordingly.

JonasHelming:
@eneufeld Could you have another look please?

eneufeld (approved):
Sorry for the late reply.
Thank you for the update.
I don't think the separate provider is needed but I'm fine with it for now.
I think `'theia/ai-ide/fileChangeSetTitle'` would be better as a key as it would be consistent with what we would like to have.
But let's keep as is for now.

Issue #15970, opened 2025-07-08
Interactive elements in hover tooltips are not clickable


ndoschek:
### Bug Description:
Hover tooltips that contain interactive UI (like links and buttons) cannot be clicked: the click is not handled, the hover is canceled and the intended action never gets triggered.

e.g.,
<img width="531" height="190" alt="Image" src="https://github.com/user-attachments/assets/2588137e-6eab-4e82-9071-8515f8f234c6" />

### Steps to Reproduce:

1. Start Theia with an extension that adds a language-status bar item containing interactive hover content (e.g. the built-in JSON lanugage basics or the cSpell Code Spell Checker).
2. Open a *.json file
3. Hover the editor language status status-bar item `{}` to open its tooltip.
4. Attempt to click an interactive element inside the tooltip (such as the pin icon to add the JSON Schema validation to the statusbar).
==> The tooltip closes and nothing happens; the interactive element does not receive the click.

### Additional Information

Operating System: Ubuntu 24.04
Theia Version: Theia IDE 1.63.200 (but probably even earlier)

chroberino:
@ndoschek I have just realised that this issue is not fully solved with Theia 1.64.0.

I can confirm it solves the case you described in your repro steps above (`Show Schema` link in built-in JSON lanugage hover).

But it does not work for status bar tooltips from many other extensions, e.g. GitLens:
<img width="523" height="236" alt="Image" src="https://github.com/user-attachments/assets/c70b9cd8-3dd6-4fbe-9874-7f81a28adbb3" />
The links in this GitLens tooltip still have the faulty behaviour, i.e. clicking them closes the tooltip and nothing more is happening.

Maybe Markdown-syled tooltips as the one from Gitlens make a difference here?
https://github.com/gitkraken/vscode-gitlens/blob/441f3f398be2d3470cce2608b4ee0bc5cd6f4e1b/src/plus/launchpad/launchpadIndicator.ts#L251

PR #15971, opened 2025-07-08
fix: enable clicks inside interactive hover tooltips and open tooltip on status-bar click

ndoschek:
#### What it does

- fix: interactive elements in hover tooltips are not clickable 
    - Add interactive flag to HoverRequest to signal clickable content
    - Update hover mouse-event handling to route clicks to interactive elements instead of cancelling the tooltip
    
    Fixes GH-15970

- fix: open tooltip on status-bar click when no other handler is set 
    
    Clicking a status-bar item that only shows an interactive tooltip should open that tooltip.
    If the item lacks a bound command and onClick handler, open its hover on click.

#### How to test

1. Start Theia with an extension that adds a language-status bar item containing interactive hover content (e.g. the built-in JSON lanugage basics or also the cSpell Code Spell Checker).
2. Open a *.json file
3. Hover the editor language status status-bar item `{}` to open its tooltip.
    - image
    - Attempt to click an interactive element inside the tooltip (such as the pin icon to add the JSON Schema validation to the statusbar) → the click should be forwarded and the expected action executed
4. Click the status-bar item itself
   - If it does not define `command`/`onClick`, the tooltip should open on click (mirrors the hover behaviour). E.g., the item  `{}` mentioned above.
   - If a handler is present, current behaviour remains unchanged. E.g., click the 'Go to line' item image which triggers a command.

Regression check: other hover tooltips (such as simple title tooltips, tree node tooltips or also visual previews in the editor tab bar) and existing status-bar commands should behave as before.

#### Follow-ups

x

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/hover-service.ts
+++ b/packages/core/src/browser/hover-service.ts
@@ -71,6 +71,11 @@ export interface HoverRequest {
      * Function that takes the desired width and returns a HTMLElement to be rendered.
      */
     visualPreview?: (width: number) => HTMLElement | undefined;
+    /**
+     * Indicates if the hover contains interactive/clickable items.
+     * When true, the hover will register a click handler to allow interaction with elements in the hover area.
+     */
+    interactive?: boolean;
 }
 
 @injectable()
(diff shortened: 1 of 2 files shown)


sdirix (approved):
Works for me!

I found a regression, unrelated to the PR. In the past we were able to click into the hover and copy/paste the content. This is no longer possible. I will open a follow up ticket for this.

sdirix:
Opened https://github.com/eclipse-theia/theia/issues/15986

PR #15974, opened 2025-07-09
Display warning when selected variant does not exist

### LINKED ISSUE #15887: Display invalid selection in prompt fragments view
### Feature Description:
After #15879 if the user has an invalid variant selected for an agent the agent will fallback to its default variant. This should be made visible in the prompt fragments view, by showing a warning or error that the default variant is being used.

sgraband:
#### What it does
Show a warning in the prompt fragment view if selected variant is not valid. Properly update the AISettingsService when preferences change. Debounce onPromptsChange in prompt-service to decrease calls (300 ms).

Fixes #15887 

#### How to test
In the settings.json set the selectedVariant of an agent to a value that does not exist and observe that a warning is shown in the prompts fragment view.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/browser/ai-settings-service.ts
+++ b/packages/ai-core/src/browser/ai-settings-service.ts
@@ -15,7 +15,7 @@
 // *****************************************************************************
 import { DisposableCollection, Emitter, Event } from '@theia/core';
 import { PreferenceScope, PreferenceService } from '@theia/core/lib/browser';
-import { inject, injectable } from '@theia/core/shared/inversify';
+import { inject, injectable, postConstruct } from '@theia/core/shared/inversify';
 import { JSONObject } from '@theia/core/shared/@lumino/coreutils';
 import { AISettings, AISettingsService, AgentSettings } from '../common';
 
@@ -28,6 +28,17 @@ export class AISettingsServiceImpl implements AISettingsService {
 
     protected readonly onDidChangeEmitter = new Emitter<void>();
     onDidChange: Event<void> = this.onDidChangeEmitter.event;
+
+    @postConstruct()
+    protected init(): void {
+        this.toDispose.push(
+            this.preferenceService.onPreferenceChanged(event => {
+                if (event.preferenceName === AISettingsServiceImpl.PREFERENCE_NAME) {
+                    this.onDidChangeEmitter.fire();
+                }
+            })
+        );
+    }
 
     async updateAgentSettings(agent: string, agentSettings: Partial<AgentSettings>): Promise<void> {
         const settings = await this.getSettings();
--- a/packages/ai-core/src/common/prompt-service.ts
+++ b/packages/ai-core/src/common/prompt-service.ts
@@ -448,14 +448,30 @@ export class PromptServiceImpl implements PromptService {
     protected _onSelectedVariantChangeEmitter = new Emitter<{ promptVariantSetId: string, variantId: string }>();
     readonly onSelectedVariantChange = this._onSelectedVariantChangeEmitter.event;
 
+    protected promptChangeDebounceTimer?: NodeJS.Timeout;
+
+    protected fireOnPromptsChangeDebounced(): void {
+        if (this.promptChangeDebounceTimer) {
+            clearTimeout(this.promptChangeDebounceTimer);
+        }
+        this.promptChangeDebounceTimer = setTimeout(() => {
+            this._onPromptsChangeEmitter.fire();
+        }, 300);
+    }
+
     @postConstruct()
     protected init(): void {
         if (this.customizationService) {
             this.customizationService.onDidChangePromptFragmentCustomization(() => {
-                this._onPromptsChangeEmitter.fire();
+                this.fireOnPromptsChangeDebounced();
             });
             this.customizationService.onDidChangeCustomAgents(() => {
-                this._onPromptsChangeEmitter.fire();
+                this.fireOnPromptsChangeDebounced();
+            });
+        }
+        if (this.settingsService) {
+            this.settingsService.onDidChange(() => {
+                this.fireOnPromptsChangeDebounced();
             });
         }
     }
@@ -764,7 +780,7 @@ export class PromptServiceImpl implements PromptService {
             }
         }
 
-        this._onPromptsChangeEmitter.fire();
+        this.fireOnPromptsChangeDebounced();
     }
 
     getVariantIds(variantSetId: string): string[] {
@@ -839,7 +855,7 @@ export class PromptServiceImpl implements PromptService {
             this.addFragmentVariant(promptVariantSetId, promptFragment.id, isDefault);
         }
 
-        this._onPromptsChangeEmitter.fire();
+        this.fireOnPromptsChangeDebounced();
     }
 
     // ===== Variant Management Methods =====
--- a/packages/ai-ide/src/browser/ai-configuration/prompt-fragments-configuration-widget.tsx
+++ b/packages/ai-ide/src/browser/ai-configuration/prompt-fragments-configuration-widget.tsx
@@ -77,14 +77,19 @@ export class AIPromptFragmentsConfigurationWidget extends ReactWidget {
     protected availableAgents: Agent[] = [];
 
     /**
-     * Maps prompt variant set IDs to their currently selected variant IDs
+     * Maps prompt variant set IDs to their resolved variant IDs
      */
-    protected selectedVariantIds: Map<string, string | undefined> = new Map();
+    protected effectiveVariantIds: Map<string, string | undefined> = new Map();
 
     /**
      * Maps prompt variant set IDs to their default variant IDs
      */
     protected defaultVariantIds: Map<string, string | undefined> = new Map();
+
+    /**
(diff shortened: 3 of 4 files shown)


packages/ai-ide/src/browser/style/index.css
```diff
@@ -337,6 +337,32 @@
   font-style: italic;
 }
 
+.prompt-variant-warning {
+  color: var(--theia-warningForeground, #b89500);
+  background: var(--theia-warningBackground, #fffbe6);
+  border-left: 4px solid var(--theia-warningBorder, #b89500);
```
ndoschek: I'm not sure we need those "fallback" color values. Our color variables should be stable enough, and we usually don't use fallback values, so I would prefer to use the variables only.

packages/ai-ide/src/browser/ai-configuration/prompt-fragments-configuration-widget.tsx
```diff
@@ -421,6 +425,31 @@ export class AIPromptFragmentsConfigurationWidget extends ReactWidget {
 
         const relatedAgents = this.getAgentsUsingPromptVariantId(promptVariantSetId);
 
+        // Determine warning/error state
+        let variantSetMessage: React.ReactNode | undefined = undefined;
+        if (effectiveVariantId === undefined) {
+            // Error: effectiveId is undefined, so nothing works
+            variantSetMessage = (
+                <div className="prompt-variant-error">
```
ndoschek, line 433: I couldn't test this part, as I'm not sure how to remove a default prompt. Is that even possible?
But anyway I think it is good that handle it!

packages/ai-ide/src/browser/ai-configuration/template-settings-renderer.tsx
```diff
@@ -32,13 +32,16 @@ export const PromptVariantRenderer: React.FC<PromptVariantRendererProps> = ({
     const defaultVariantId = promptService.getDefaultVariantId(promptVariantSet.id);
     const [selectedVariant, setSelectedVariant] = React.useState<string>(defaultVariantId!);
 
+    promptService.onSelectedVariantChange(notification => {
+        if (notification.promptVariantSetId === promptVariantSet.id) {
+            setSelectedVariant(notification.variantId ?? defaultVariantId!);
+        }
+    });
```
sdirix: With every rerender we register a new listener here. The listener should be registered as part of a `useEffect` and disposed with the return function of the `useEffect`.

packages/ai-ide/src/browser/ai-configuration/template-settings-renderer.tsx
```diff
@@ -32,13 +32,16 @@ export const PromptVariantRenderer: React.FC<PromptVariantRendererProps> = ({
...
     React.useEffect(() => {
         (async () => {
-            const currentVariant =
-                await promptService.getSelectedVariantId(promptVariantSet.id);
-            if (currentVariant) {
-                setSelectedVariant(currentVariant);
-            }
+            const currentVariant = promptService.getSelectedVariantId(promptVariantSet.id);
+            setSelectedVariant(currentVariant ?? defaultVariantId!);
         })();
```
sdirix: This async wrapping is no longer needed as we only call sync methods.

packages/ai-core/src/common/prompt-service.ts
```diff
@@ -445,21 +449,80 @@ export class PromptServiceImpl implements PromptService {
...
+    protected toDispose = new DisposableCollection();
+
+    protected fireOnPromptsChangeDebounced(): void {
+        if (this.promptChangeDebounceTimer) {
+            clearTimeout(this.promptChangeDebounceTimer);
+        }
+        this.promptChangeDebounceTimer = setTimeout(() => {
+            this._onPromptsChangeEmitter.fire();
+        }, 300);
+    }
```
sdirix, line 466: On a first glance this looks like a code smell. It looks like we don't have sufficient control over our change emitter firing and this is a bandaid to fix the symptom not the cause.

Which are the use cases which we want to solve here? I added a log here and only ever saw the log being hit twice in a row (when changing a variant in the agent settings view). So is the double notification the only one to streamline or are there other use cases too?

packages/ai-core/src/browser/ai-settings-service.ts
```diff
@@ -29,6 +29,17 @@ export class AISettingsServiceImpl implements AISettingsService {
     protected readonly onDidChangeEmitter = new Emitter<void>();
     onDidChange: Event<void> = this.onDidChangeEmitter.event;
 
+    @postConstruct()
+    protected init(): void {
+        this.toDispose.push(
+            this.preferenceService.onPreferenceChanged(event => {
+                if (event.preferenceName === AISettingsServiceImpl.PREFERENCE_NAME) {
+                    this.onDidChangeEmitter.fire();
```
sdirix, line 41: We fire a change event whenever the preferences change, but we also fire a change event manually when we ourselves change the preferences in `updateAgentSettings`.

This is the cause of double notifications for the same change.

With this notification for every preference change we no longer need to fire the event in `updateAgentSettings`. To be fully clean we shoud try-catch the preference change in `updateAgentSettings` and fire a change event in case we catch an error (i.e. updating the preferences was unsuccessful), but not in the case where updating the preferences was successful.

packages/ai-ide/src/browser/ai-configuration/prompt-fragments-configuration-widget.tsx
```diff
@@ -421,6 +425,31 @@ export class AIPromptFragmentsConfigurationWidget extends ReactWidget {
...
+                    <span className="codicon codicon-error"></span>
+                    {nls.localize('theia/ai/core/promptFragmentsConfiguration/variantSetError',
```
sdirix, line 435: Just for my understanding: This error case should never happen right?

We ask the `promptService` for all prompt variants, then for each variant we ask the `promptService` again for the `effectiveId`. So it should never be `undefined` as we always have an `effectiveId` for all prompt variants returned by the `promptService`.

I wonder whether it would make sense that the `promptService` offers a method to return all promptVariants including their `effective`, `default` and `selected` ids instead of collecting them manually via this code at the top:

```
for (const promptVariantSetId of this.promptVariantsMap.keys()) {
    const effectiveId = this.promptService.getEffectiveVariantId(promptVariantSetId);
    const defaultId = this.promptService.getDefaultVariantId(promptVariantSetId);
    const selectedId = this.promptService.getSelectedVariantId(promptVariantSetId) ?? defaultId;
    this.userSelectedVariantIds.set(promptVariantSetId, selectedId);
    this.effectiveVariantIds.set(promptVariantSetId, effectiveId);
    this.defaultVariantIds.set(promptVariantSetId, defaultId);
}
```

If we can manage to guarantee that `effectiveId` always exist we could remove this error case handling, translation work etc.

sdirix (changes requested):
In testing everything works! However I left some comments and questions, please have a look.

packages/ai-core/src/common/prompt-service.ts
sgraband: The main reason to do this was the reload/startup use case, where a lot of changes will be pushed. We can also remove this again if you feel like that this is not worth the added complexity.

packages/ai-ide/src/browser/ai-configuration/prompt-fragments-configuration-widget.tsx
sgraband: Yes, currently this should not happen, but afaik we might plan to add functionality for the user to add variant set ids as well, where this then might become relevant, depending on the implentation of that functionality. But we can also remove it if you like, i would however not alter the way the data is retrieved from the prompt-service, as the service is already quite complex and the suggested fucntion would most likely only be for config views.

packages/ai-ide/src/browser/ai-configuration/template-settings-renderer.tsx
```diff
@@ -33,13 +33,17 @@ export const PromptVariantRenderer: React.FC<PromptVariantRendererProps> = ({
...
+
+        const disposable = promptService.onSelectedVariantChange(notification => {
+            if (notification.promptVariantSetId === promptVariantSet.id) {
+                setSelectedVariant(notification.variantId ?? defaultVariantId!);
             }
-        })();
+        });
+        return () => {
+            disposable.dispose();
+        };
```
sdirix: The `useEffect` dependencies are off.

- `promptService` is missing. We know it should be a singleton but no reason not to list it
- `agentId` is not used within the effect
- `defaultVariantId` is missing

packages/ai-ide/src/browser/ai-configuration/template-settings-renderer.tsx
```diff
@@ -33,13 +33,17 @@ export const PromptVariantRenderer: React.FC<PromptVariantRendererProps> = ({
...
     }, [promptVariantSet.id, agentId]);
```
sdirix:
```suggestion
    }, [promptVariantSet.id, promptService, defaultVariantId]);
```

sdirix (approved):

PR #15979, opened 2025-07-10
Fix cannot drop explorer view files on AI chat input on Windows

### LINKED ISSUE #15976: Dropping a file from the Explorer view into the AI chat view input breaks the UI on Windows
### Bug Description:

Dropping a file from the Explorer view into the AI chat view input breaks the UI on Windows.
This is because the selected tree nodes drag event data ids have the workspace root and the absolute file path concatenated with ":". e.g. `/c:/Users/Vivien/martini-designer-e2e-tests:/c:/Users/Vivien/martini-designer-e2e-tests/conf/application.properties`
 The drop handler then try to split the id using ":" to get the file path but on Windows it also contains the drives like "/c:/" which ends up creating an invalid workspace relative path. e.g. `/Users/Vivien/martini-designer-e2e-tests` Which then break the `ContextFileVariableLabelProvider` with the following error:

```
index.js:73 Uncaught Error: [UriError]: If a URI does not contain an authority component, then the path cannot begin with two slash characters ("//")
    at _validateUri (index.js:73:1)
    at _URI.URI (index.js:149:1)
    at new _URI (index.js:370:1)
    at __webpack_modules__.../../node_modules/vscode-uri/lib/esm/index.js.URI.from (index.js:308:1)
    at URI.withPath (uri.ts:135:32)
    at get parent (uri.ts:77:21)
    at WorkspaceUriLabelProviderContribution.getDetails (workspace-uri-contribution.ts:71:89)
    at LabelProvider.handleRequest (label-provider.ts:370:47)
    at LabelProvider.getDetails (label-provider.ts:365:21)
    at ContextFileVariableLabelProvider.getDetails (context-file-variable-label-provider.ts:52:35)
```

Image

### Steps to Reproduce:

1. Open the Explorer and AI Chat
2. Select a file and drop it into the chat input
3. The chat input disappears and errors are visible in the console. 

### Additional Information

- Operating System: Windows
- Theia Version: 1.63

I'm interested in contributing a PR to fix this.

Hanksha:
#### What it does

This PR better handles the file drops from the explorer view into the AI chat input so that it also works on Windows.
The drop data contains node ids which follow the format `wsRootPath:filePath`. So to resolve the file path, I find the workspace root path corresponding to this id and substract it. This also work with paths containing drive letters on Windows unlike the previous implementation which splitted the id by `:` .

Closes #15976 

#### How to test

1. Open the Explorer and AI Chat
2. Select a file and drop it into the chat input
3. The file should properly show as context on the chat input

I only tested it on Windows, if someone could test it on linux or macOS that would be nice.

feature

#### Attribution

Contributed on behalf of Lonti.com Pty Ltd.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/browser/file-chat-variable-contribution.ts
+++ b/packages/ai-chat/src/browser/file-chat-variable-contribution.ts
@@ -25,6 +25,7 @@ import { WorkspaceService } from '@theia/workspace/lib/browser';
 import { FileService } from '@theia/filesystem/lib/browser/file-service';
 import { VARIABLE_ADD_CONTEXT_COMMAND } from './ai-chat-frontend-contribution';
 import { IMAGE_CONTEXT_VARIABLE, ImageContextVariable } from '../common/image-context-variable';
+import { ApplicationShell } from '@theia/core/lib/browser';
 
 @injectable()
 export class FileChatVariableContribution implements FrontendVariableContribution {
@@ -207,30 +208,26 @@ export class FileChatVariableContribution implements FrontendVariableContributio
     }
 
     protected async handleDrop(event: DragEvent, _: AIVariableContext): Promise<AIVariableDropResult | undefined> {
-        const data = event.dataTransfer?.getData('selected-tree-nodes');
-        if (!data) {
+        if (!event.dataTransfer) {
+            return undefined;
+        }
+
+        const uris = ApplicationShell.getDraggedEditorUris(event.dataTransfer);
+        if (!uris.length) {
             return undefined;
         }
 
         try {
-            const nodes: string[] = JSON.parse(data);
             const variables: AIVariableResolutionRequest[] = [];
             const texts: string[] = [];
-
-            for (const node of nodes) {
-                const [, filePath] = node.split(':');
-                if (!filePath) {
-                    continue;
-                }
-
-                const uri = URI.fromFilePath(filePath);
+            for (const uri of uris) {
                 if (await this.fileService.exists(uri)) {
                     const wsRelativePath = await this.wsService.getWorkspaceRelativePath(uri);
                     const fileName = uri.displayName;
 
-                    if (this.isImageFile(filePath)) {
+                    if (this.isImageFile(wsRelativePath)) {
                         const base64Data = await this.fileToBase64(uri);
-                        const mimeType = this.getMimeTypeFromExtension(filePath);
+                        const mimeType = this.getMimeTypeFromExtension(wsRelativePath);
                         variables.push(ImageContextVariable.createRequest({
                             [ImageContextVariable.name]: fileName,
                             [ImageContextVariable.wsRelativePath]: wsRelativePath,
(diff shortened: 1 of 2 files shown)


packages/ai-chat/src/browser/file-chat-variable-contribution.ts
```diff
@@ -216,13 +216,17 @@ export class FileChatVariableContribution implements FrontendVariableContributio
...
+            const rootPaths = (await this.wsService.roots).map(root => root.resource.path.toString());
 
             for (const node of nodes) {
-                const [, filePath] = node.split(':');
-                if (!filePath) {
+                const rootPath = rootPaths.find(path => node.startsWith(path + ':'));
+
+                if (!rootPath) {
                     continue;
                 }
```
sdirix: Please add a comment on why we are doing this. When stumbling over it later on, developers will ask themselves why this is necessary.

sdirix (commented):
Works for me on Linux. Actually makes Linux also behave better in case there is a `:` somewhere in the path. Before this change, drag and dropping such files were not possible, now it is.

packages/ai-chat/src/browser/file-chat-variable-contribution.ts
Hanksha: Done.

packages/ai-chat/src/browser/file-chat-variable-contribution.ts
```diff
@@ -216,13 +216,21 @@ export class FileChatVariableContribution implements FrontendVariableContributio
...
+            const rootPaths = (await this.wsService.roots).map(root => root.resource.path.toString());
 
             for (const node of nodes) {
-                const [, filePath] = node.split(':');
-                if (!filePath) {
+                // The nodes are ids following the format "wsRootPath:filePath"
+                // First we need the workspace root path that the node id belongs to
+                // and then we can extract the file path from the node id.
+                // We cannot simply split by ':' because the root or file path might contain colons (e.g. drive letter in Windows).
+                const rootPath = rootPaths.find(path => node.startsWith(path + ':'));
```
tsmaeder: I don't think this is a good way to get the paths of the dragged files: it depends on the format of the ids of the file tree model. But that is an implementation detail of the file navigator which we should not rely upon. The right way, IMO would be to use `ApplicationShell.getDraggedEditorUris()`, which would also work for dragging files from contributed tree views. The method is currently protected, but could easily be made public.

sdirix: Good insight!

Hanksha: That makes more sense, I'll try that! Thanks

sdirix (approved):
Works for me on Linux

JonasHelming:
@tsmaeder ?

tsmaeder (approved):

JonasHelming:
@sdirix Can we merge this?

PR #15982, opened 2025-07-11
fix: DynamicMenuWidget Memory Leak  #15975

### LINKED ISSUE #15975: DynamicMenuWidget Memory Leak
Bug Description:
There is a memory leak in Theia caused by inconsistent usage of addEventListener and removeEventListener. A listener is added with the third parameter set to true (capture mode), but later removed without the same flag. This mismatch prevents the listener from being properly removed, causing retained references and increasing memory usage over time. The issue is located in the Theia framework itself, specifically in the file:
@theia/core/lib/browser/browser-menu-plugin.ts

Steps to Reproduce:
Right-click anywhere in the Theia interface to open the context menu.

Close the context menu.

Repeat this process multiple times (open and close the right-click menu).

Observe memory usage increasing over time.

Additional Information
Operating System: Windows 10

Theia Version: 1.60.2 (also reproducible in the latest version)

Affected File: @theia/core/lib/browser/browser-menu-plugin.ts

Image

Hbbbit:
What it does
Fixes a memory leak in @theia/core/lib/browser/browser-menu-plugin.ts caused by inconsistent usage of addEventListener and removeEventListener.
Previously, an event listener was registered with capture: true, but removed without specifying the same option.
This mismatch prevented the listener from being properly removed, leading to retained memory on repeated context menu interactions.

This change ensures that both addEventListener and removeEventListener use the same options, allowing proper cleanup and stable memory usage.

Fixes #15975 

How to test
Launch Theia and open the browser developer tools (e.g., Chrome DevTools).

Right-click anywhere in the Theia UI to open the context menu, then click elsewhere to close it.

Repeat the open/close operation ~10 times.

Take two memory snapshots before and after.
You should now see that no redundant event listeners remain and memory usage stays stable.

Without the fix, you would observe increasing retained nodes in memory.

Follow-ups
None currently.
Could consider writing automated regression tests if event listener state becomes more observable in future refactoring.

Breaking changes
 This PR does not introduce breaking changes.

Attribution
Contributed by Hbb.

### DIFF
--- a/packages/core/src/browser/menu/browser-menu-plugin.ts
+++ b/packages/core/src/browser/menu/browser-menu-plugin.ts
@@ -251,7 +251,7 @@ export class DynamicMenuWidget extends MenuWidget {
     }
 
     protected override onBeforeDetach(msg: Message): void {
-        this.node.ownerDocument.removeEventListener('pointerdown', this);
+        this.node.ownerDocument.removeEventListener('pointerdown', this, true);
         super.onAfterDetach(msg);
     }
 


sdirix:
Hi! Thanks for the contribution :heart: 

Please make sure to sign the [ECA](https://api.eclipse.org/git/eca/status/gh/eclipse-theia/theia/15982), otherwise this can't be merged. Thanks!

Hbbbit:
DONE

sdirix (approved):
I can confirm that the `pointerDown` event listeners are successfully cleaned up with this PR

PR #15987, opened 2025-07-11
Don't console.log on non-existent backend-settings.json file

Torbjorn-Svensson:
#### What it does

Avoid logging a stacktrace on startup when the ~/.theia/backend-settings.json file does not exist.

#### How to test

Start the product without having the ~/.theia/backend-settings.json file.
There should be no stacktrace logged during the startup.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/node/setting-service.ts
+++ b/packages/core/src/node/setting-service.ts
@@ -50,7 +50,9 @@ export class SettingServiceImpl implements SettingService {
                 });
                 this.values = JSON.parse(contents);
             } catch (e) {
-                console.log(e);
+                if (e.code !== 'ENOENT') {
+                    console.log(e);
+                }
             } finally {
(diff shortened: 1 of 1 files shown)


rschnekenbu (approved):
LGTM!

rschnekenbu (changes requested):
This now conflicts with some udpates done on master branch (see https://github.com/eclipse-theia/theia/blob/b79a25bea5deb94a3ce39a08a633d484945a71a1/packages/core/src/node/setting-service.ts#L54). Do you think that is OK with the current changes and your PR can be closed?

Torbjorn-Svensson:
Fine by me. I just noticed the stack trace on startup in our logs and stack traces usually means that something important goes wrong, but that was not the case here.

PR #16004, opened 2025-07-15
Use environment variable instead of wmic to get remote architecture

### LINKED ISSUE #15990: Allow non-admin ssh windows connections
Connecting to a server, on windows we use wmic to detect the remote architecture.

[remote-setup-service.ts](https://github.com/eclipse-theia/theia/blob/master/packages/remote/src/electron-node/setup/remote-setup-service.ts#L158)

```typescript
const wmicResult = await connection.exec('wmic OS get OSArchitecture');
if (wmicResult.stdout.includes('64-bit')) {
    arch = 'x64';
} else if (wmicResult.stdout.includes('32-bit')) {
    arch = 'x86';
}
```

However, when I try connecting to a non-administrator remote user, I get permission errors that appear like so in the IDE:

Image

Using instead use the variable [PROCESSOR_ARCHITECTURE](https://learn.microsoft.com/en-ca/windows/win32/winprog64/wow64-implementation-details?redirectedfrom=MSDN), I see no issues connecting.

```typescript
const wmicResult = await connection.exec("echo %PROCESSOR_ARCHITECTURE%");
if (wmicResult.stdout.includes("64")) {
    arch = "x64";
} else if (wmicResult.stdout.includes("x86")) {
    arch = "x86";
}
```

Could we use this, or another method instead of relying on wmic?

Tested on [58507bbe](https://github.com/eclipse-theia/theia/tree/58507bbedb95724735981f44f034e7036fa2f19e)
Host is on Ubuntu 22.04, and server is on Windows 11

parisa-mchp:
Signed-off-by: Parisa Betel Miri <parisa.betelmiri@microchip.com>

#### What it does

Fixes #15990. Remoting over ssh to a non-admin user on a windows server shows this error:

465409284-a7f237fc-e69b-4f74-959b-8a805666b9bc

Debugging reveals that it's a permission issue with the wmic command used to determine the remote architecture, which we can instead check using the environment variable: [PROCESSOR_ARCHITECTURE](https://learn.microsoft.com/en-ca/windows/win32/winprog64/wow64-implementation-details?redirectedfrom=MSDN). The command is prefixed `cmd /c` so that it will run in cmd even if the parent shell is powershell (as suggested by @msujew). Below is the output of the command run locally in powershell and cmd:

image

image

#### How to test

Use the ssh extension to connect to a non-admin user on a windows server. Some servers might be configured so that the current implementation doesn't produce an error, but this was not the case on either of the servers I tested. 

https://serverfault.com/a/28523

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/remote/src/electron-node/setup/remote-setup-service.ts
+++ b/packages/remote/src/electron-node/setup/remote-setup-service.ts
@@ -155,10 +155,10 @@ export class RemoteSetupService {
         }
         let arch: string | undefined;
         if (os === OS.Type.Windows) {
-            const wmicResult = await connection.exec('wmic OS get OSArchitecture');
-            if (wmicResult.stdout.includes('64-bit')) {
+            const processorArchitecture = await connection.exec('cmd /c echo %PROCESSOR_ARCHITECTURE%');
+            if (processorArchitecture.stdout.includes('64')) {
                 arch = 'x64';
-            } else if (wmicResult.stdout.includes('32-bit')) {
+            } else if (processorArchitecture.stdout.includes('x86')) {
                 arch = 'x86';
             }
         } else {


msujew (approved):
Thank you, makes sense 👍

PR #15671, opened 2025-05-28
Improve type safety of remote ssh config

msujew:
#### What it does

I've noticed that https://github.com/eclipse-theia/theia/pull/15499 introduced a bunch of rather unsafe code that could potentially prevent the remote SSH feature from working.

The PR also did a bunch of manual work to replace env variables - this was not necessary, since we can already do this via the `VariablesResolverService`.

#### How to test

1. Assert that the `remote.ssh.connectToConfigHost` command still works as expected (i.e. variable resolving, etc.).
2. Change the value of the `remote.ssh.configFile` preference to point to a non-existing file path, and assert that a notification is shown to the user.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #16010, opened 2025-07-16
Using empty config in absence of .ssh/config

### LINKED ISSUE #16006: Need `.ssh/config` to use SSH
### Bug Description:

If trying to use SSH when the `.ssh/config` file is absent produces an error here: [remote-ssh-connection-provider.ts](https://github.com/eclipse-theia/theia/blob/master/packages/remote/src/electron-node/ssh/remote-ssh-connection-provider.ts#L102). However, if you create an empty config file it works fine. Would it make sense to handle no config file the same as an empty config?

### Steps to Reproduce:

1. Delete or rename the `.ssh/config` file.
2. Attempt to use the SSH extension.
3. There should be an error like:

Image

What's curious also is that on windows the error looks like this:

Image

where there should be a slash between the username and `.ssh`. This issue seems to be purely visual however and doesn't affect the functionality.

### Additional Information

- Operating System: Ubuntu 22.04 and Windows 11
- Theia Version: [b79a25be](https://github.com/eclipse-theia/theia/tree/b79a25bea5deb94a3ce39a08a633d484945a71a1)

parisa-mchp:
#### What it does

Fixes #16006. If the user hasn't created an ssh config file and tries to start a connection, we proceed with an empty config object. It still show errors if the file exists, but there's an issue, such as a parsing or permission error as the expectation is that the user intended to use that file.

#### How to test

To reproduce the issue on master, delete or rename the `~/.ssh/config` file, and try to use the remote extension. It should produce an error like:

466545029-6c5aec6c-5d6f-41e1-a6ef-d46e539e4b85

This error should be absent on this branch.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


msujew:
I believe this was already fixed by https://github.com/eclipse-theia/theia/pull/15671 (which also addressed a few other issues wrt. to ssh config parsing). @parisa-mchp Can you confirm? You will either way need to rebase the changes.

parisa-mchp:
Yeah, tested out the latest change and it seems to have fixes this issue.

Issue #16048, opened 2025-07-25
[ai-mcp] add option to run a resolve operation on server start


fipro78:
### Feature Description:

If you contribute a mcp server programmatically via VS Code extension, you can provide a `resolveMcpServerDefinition()` function that allows to add additional information before the server start. This can be for example an authorization token that needs to be provided by the user. This can be seen for example in the [VS Code MCP developer guide](https://code.visualstudio.com/api/extension-guides/ai/mcp#2.-implement-the-provider).

Adding an extension that contributes such a `McpServerDefinitionProvider` will open the input box for the authorization token when the mcp server is started.

If you add that extension to Theia, the input box will show even before the shell is visible. The reason is that the resolve operation is called on loading the mcp server description and not on start of the server. 

The reason for this difference in behavior is that the Theia `MCPServerDescription` does not provide a way to add a resolve method similar to VS Code.

IMO it would be beneficial if a developer could add a resolve method when adding a `MCPServerDescription` programmatically. This would additionally give the opportunity to align the lifecycle behavior of a mcp server with VS Code.

PR #16049, opened 2025-07-25
16048: feat add option to run a resolve operation on mcp server start

fipro78:
#### What it does

This PR adds the ability to add a `resolve` method to a `MCPServerDescription` that gets executed on starting the mcp server.

It will also align the lifecycle behavior of a mcp server that is contributed programmatically via VS Code extension in Theia. (see #16048)

A `MCPServerDescription` with a `resolve` method needs to be added via `MCPFrontendNotificationService` and not directly via `MCPServerManager`, as the resolve operation will be executed on the frontend.

#### How to test

Contribute a `MCPServerDescription` that has a `resolve` method, that for example opens a input box before the server is started.

```typescript
const exampleServer: RemoteMCPServerDescription = {
	name: "ExampleMCP",
	serverUrl: "<example_url>",
	resolve: async (serverDescription) => {
		console.log('Resolving MCP server description:', serverDescription.name);

		// Prompt user for authentication token
		const authToken = await this.quickInputService.input({
			prompt: 'Enter authentication token for MCP server',
			password: true,
			value: 'serverAuthToken' in serverDescription ? serverDescription.serverAuthToken || '' : ''
		});

		if (authToken) {
			// Return updated server description with new token
			return {
				...serverDescription,
				serverAuthToken: authToken
			} as RemoteMCPServerDescription;
		}

		// If no token provided, return original description
		return serverDescription;
	}
};
this.frontendMCPService.addOrUpdateServer(exampleServer);
```

#### Follow-ups

I am actually not 100% sure about the flow of the server-client-server communication. That probably needs to be reviewed more carefully. I tested with a VS Code extension that contributes servers with a resolve operation, and with a Theia extension that programmatically contributes a mcp server as shown above. Both examples worked as intended in a Theia browser application. But of course I only tested the single user use case. Not sure if I am missing something in case of multi-user-setups.

#### Breaking changes

The possible client-server communication might be a breaking change. That needs to be verified closely. Not sure if it is actually a breaking change, as the methods are mostly private and the resolve method is optional.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-mcp/src/browser/mcp-frontend-module.ts
+++ b/packages/ai-mcp/src/browser/mcp-frontend-module.ts
@@ -20,15 +20,13 @@ import {
     MCPFrontendService,
     MCPServerManager,
     MCPServerManagerPath,
-    MCPFrontendNotificationService,
-    MCPServerManagerServerClient,
-    MCPServerManagerServer,
-    MCPServerManagerServerPath
+    MCPFrontendNotificationService
 } from '../common/mcp-server-manager';
 import { McpFrontendApplicationContribution } from './mcp-frontend-application-contribution';
 import { MCPFrontendServiceImpl } from './mcp-frontend-service';
 import { MCPFrontendNotificationServiceImpl } from './mcp-frontend-notification-service';
 import { MCPServerManagerServerClientImpl } from './mcp-server-manager-server-client';
+import { MCPServerManagerServer, MCPServerManagerServerClient, MCPServerManagerServerPath } from '../common/mcp-protocol';
 
 export default new ContainerModule(bind => {
     bind(FrontendApplicationContribution).to(McpFrontendApplicationContribution).inSingletonScope();
@@ -47,12 +45,19 @@ export default new ContainerModule(bind => {
         const connection = ctx.container.get<ServiceConnectionProvider>(RemoteConnectionProvider);
         const client = ctx.container.get<MCPFrontendNotificationService>(MCPFrontendNotificationService);
         const serverClient = ctx.container.get<MCPServerManagerServerClient>(MCPServerManagerServerClient);
-        const proxy = connection.createProxy<MCPServerManager>(MCPServerManagerPath, client);
+        const backendServerManager = connection.createProxy<MCPServerManager>(MCPServerManagerPath, client);
 
         // Listen to server updates to clean up removed servers
-        client.onDidUpdateMCPServers(() => serverClient.syncServerDescriptions(proxy));
+        client.onDidUpdateMCPServers(() =>
+            backendServerManager.getServerNames()
+                .then(names => serverClient.cleanServers(names))
+                .catch((error: unknown) => {
+                    console.error('Error cleaning server descriptions:', error);
+                }));
 
-        return new Proxy(proxy, {
+        // We proxy the MCPServerManager to override addOrUpdateServer and getServerDescription
+        // to handle the resolve functions via the MCPServerManagerServerClient.
+        return new Proxy(backendServerManager, {
             get(target: MCPServerManager, prop: PropertyKey, receiver: unknown): unknown {
                 // override addOrUpdateServer to store the original description in the MCPServerManagerServerClient
                 // to be used in resolveServerDescription if a resolve function is provided
@@ -60,6 +65,22 @@ export default new ContainerModule(bind => {
                     return async function (this: MCPServerManager, ...args: [serverDescription: Parameters<MCPServerManager['addOrUpdateServer']>[0]]): Promise<void> {
                         const updated = serverClient.addServerDescription(args[0]);
                         await mgrServer.addOrUpdateServer(updated);
+                    };
+                }
+                // override getServerDescription to mix in the resolve function from the client
+                if (prop === 'getServerDescription') {
+                    return async function (this: MCPServerManager, name: string): ReturnType<MCPServerManager['getServerDescription']> {
+                        const description = await Reflect.apply(target.getServerDescription, target, [name]);
+                        if (description) {
+                            const resolveFunction = serverClient.getResolveFunction(name);
+                            if (resolveFunction) {
+                                return {
+                                    ...description,
+                                    resolve: resolveFunction
+                                };
+                            }
+                        }
+                        return description;
                     };
                 }
                 return Reflect.get(target, prop, receiver);
--- a/packages/ai-mcp/src/browser/mcp-server-manager-server-client.ts
+++ b/packages/ai-mcp/src/browser/mcp-server-manager-server-client.ts
@@ -15,32 +15,30 @@
 // *****************************************************************************
 
 import { injectable } from '@theia/core/shared/inversify';
-import { MCPServerDescription, MCPServerManager, MCPServerManagerServerClient } from '../common';
+import { MCPServerDescription } from '../common';
 import { generateUuid } from '@theia/core/lib/common/uuid';
+import { cleanServerDescription, MCPServerDescriptionRCP, MCPServerManagerServerClient } from '../common/mcp-protocol';
+
+type StoredServerInfo = Pick<MCPServerDescription, 'name' | 'resolve'>;
 
 @injectable()
 export class MCPServerManagerServerClientImpl implements MCPServerManagerServerClient {
 
-    protected serverDescriptions: Map<string, MCPServerDescription> = new Map();
+    protected serverDescriptions: Map<string, StoredServerInfo> = new Map();
 
-    /**
-     * Adds a server description to the client. If the description contains a resolve function,
-     * a unique resolveId is generated and the original description is stored to be used
-     * in resolveServerDescription.
-     * @param description The server description to add.
-     * @returns The server description with a unique resolveId if a resolve function is provided
-     * or the given description if no resolve function is provided.
-     */
-    addServerDescription(description: MCPServerDescription): MCPServerDescription {
+    addServerDescription(description: MCPServerDescription): MCPServerDescriptionRCP {
         if (description.resolve) {
-            const serverDescription: MCPServerDescription = {
+            const serverDescription: MCPServerDescriptionRCP = {
                 ...description,
                 resolveId: generateUuid(),
             };
 
-            // store the original description to be used in resolveServerDescription
+            // store only the name and resolve function
             if (serverDescription.resolveId) {
-                this.serverDescriptions.set(serverDescription.resolveId, description);
+                this.serverDescriptions.set(serverDescription.resolveId, {
+                    name: description.name,
+                    resolve: description.resolve
+                });
             }
 
             return serverDescription;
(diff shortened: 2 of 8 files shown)


packages/ai-mcp/src/browser/mcp-frontend-application-contribution.ts
```diff
@@ -84,9 +84,14 @@ export class McpFrontendApplicationContribution implements FrontendApplicationCo
     @inject(MCPFrontendService)
     protected frontendMCPService: MCPFrontendService;
 
+    @inject(MCPFrontendNotificationService)
+    protected readonly mcpFrontendNotificationService: MCPFrontendNotificationService;
+
     protected prevServers: Map<string, MCPServerDescription> = new Map();
 
     onStart(): void {
+        this.mcpFrontendNotificationService.setFrontendMCPService(this.frontendMCPService);
```
sdirix: We usually don't use `FrontendApplicationContribution`s to wire dependency services manually. The architecture of the services should be restructured to allow the usual `@inject` injection.

packages/ai-mcp/src/node/mcp-server-manager-impl.ts
```diff
@@ -136,6 +142,11 @@ export class MCPServerManagerImpl implements MCPServerManager {
         this.clients.forEach(client => client.didUpdateMCPServers());
     }
 
+    private async resolve(description: MCPServerDescription): Promise<MCPServerDescription> {
+        const results = await Promise.all(this.clients.map(client => client.resolveServerDescription(description)));
+        return results[0];
+    }
```
sdirix: The current solution has multiple downsides:
 - The `MCPServerManager` is "burdened" with the knowledge on how to resolve the server descriptions. It would be nicer if it could just call the `resolve` method on the description
 - The current solution will not work if an adopter registers a description from within the backend. The `MCPServerManager` will still try to resolve via the frontend.
 - The current solution will not work if an adopter uses the `MCPServerManager` proxy in the frontend to register their MCP servers and they are not told why.

Of course `MCPServerManager` could be made aware of these things and only call the frontend if necessary.

---

However I would like to suggest a different approach:
 - The `MCPServerManager` just always calls `resolve` on the description (after checking that the description has a `resolve`
 - The frontend `MCPServerManager` is enhanced to specialize the `addOrUpdateServer` call. It should not directly forward to the `MCPServerManager` of the backend, but call a specialized service on the backend. That service can then call `addOrUpdateServer` of the `MCPServerManager`, handing over a resolve which will correctly delegate to the frontend specialization of the `MCPServerManager`.

The benefits are:
 - This would clearly separate the resolve-specialization needed because of the RCP gap from the `MCPServerManager`.
 - No matter which variant of `addOrUpdateServer` is called, all of them would work

sdirix (commented):
I had a look at the implementation and left a rather large suggestion for a refactoring. What do you think?

packages/ai-mcp/src/browser/mcp-frontend-application-contribution.ts
fipro78: I also actually didn't really like this one. But as I didn't really understand the services structure here, got "hen-egg" injection issues and found the same "pattern" in another place in Theia you referenced in a conversation, I thought it would be ok.
Anyhow, if you have an idea how to restructure the architecture of the services, please let me know. My knowledge in that area is currently still limited and I would like to learn.

packages/ai-mcp/src/node/mcp-server-manager-impl.ts
fipro78: That sounds all reasoning to me. But actually I am not really sure how to achieve this. Any more detailed hints are appreciated. It took me quite a while to get even to this point, as the RPC stuff in Theia is not yet really clear to me.

packages/ai-mcp/src/browser/mcp-frontend-application-contribution.ts
sdirix: In this case here you can inject the `MCPFrontendNotificationService` in the `MCPFrontendService` and then call `setFrontendMCPService` in a `@postConstruct()` of the `MCPFrontendService`. This way the relations are more explicit.

To generalize this even more it would be nicer to have a separate interface for resolving so that the `MCPFrontendNociticationService` does not even know that it gets the `MCPFrontendService` set.

However all of this is not necessary if we choose the other approach.

packages/ai-mcp/src/node/mcp-server-manager-impl.ts
sdirix: First, adjust the `MCPServerManager` to just call `resolve` on a description in case the `resolve` function is specified. Any MCP servers added programmatically by adopters within the backend should then already be resolvable.

Whenever there is a service which only exists for frontend handling, we usually name it "....Server". Therefore, implement a new class `MCPServerManagerServer` on the backend. It only has an `addOrUpdateServer` method which will be called from the frontend. It accepts a `MCPServerDescription` as a parameter like the other functions, however with one difference: Instead of a `resolve` method it expects a `resolveId: string`.

The `MCPServerManagerServer` injects the `MCPServerManager` and forwards the description from `addOrUpdateServer`. However it removes the `resolveId: string` field and instead adds a proper `resolve` method. To properly implement this, we need a client so we can call something on the frontend.

Therefore add a `MCPServerManagerServerClient`. It's a very thin class. Basically it is used for exactly the same purpose what in the current state of the PR the `MCPFrontendNotificationService` was extended to do: To resolve the description on the frontend.

This client can be used in the `MCPServerManagerServer` to implement the `resolve` function, using the handed over `resolveId`.

On the frontend side, it's probably the easiest to use a Javascript proxy on the MCPServerManager RCP proxy: Where the `MCPServerManager` RCP proxy is created, you can wrap it with a Javascript proxy modifying the `addOrUpdateServer` method. The original description is stored in an array, a unique `resolveId` is generated, and instead of forwarding the description to the original `MCPServerManager` RCP proxy, you forward the modified one to `MCPServerManagerServer`. Also you can retrieve the `MCPServerManagerServerClient` and register yourself as the resolver.

That should be it. It sounds worse than it is. Basically just move the frontend resolving from the `MCPFrontendNotificationService` to a separate one and adding a new backend service with a single method.

fipro78: @sdirix 
I think I have finally managed to implement the changes as requested. I hope it matches your recommendations. Please have a look and let me know if it is ok this way or if I misunderstood something. 

I tested it locally with a VS Code Extension that contributes a mcp server with a resolve method, and I added an example in the api-samples to verify that it also works for programmatically contributing a MCP server via Theia extension.

packages/ai-mcp/src/browser/mcp-server-manager-server-client.ts
```diff
@@ -0,0 +1,55 @@
...
+@injectable()
+export class MCPServerManagerServerClientImpl implements MCPServerManagerServerClient {
+
+    protected serverDescriptions: Map<string, MCPServerDescription> = new Map();
+
+    addServerDescription(description: MCPServerDescription): MCPServerDescription {
+        if (description.resolve) {
+            const serverDescription: MCPServerDescription = {
+                ...description,
+                resolveId: crypto.randomUUID(),
```
sdirix: You can use `generateUuid` from theia/core, i.e. `@theia/core/lib/common/uuid`

packages/ai-mcp/src/browser/mcp-frontend-module.ts
```diff
@@ -16,18 +16,48 @@
...
     bind(MCPServerManager).toDynamicValue(ctx => {
+        const mgrServer = ctx.container.get<MCPServerManagerServer>(MCPServerManagerServer);
         const connection = ctx.container.get<ServiceConnectionProvider>(RemoteConnectionProvider);
         const client = ctx.container.get<MCPFrontendNotificationService>(MCPFrontendNotificationService);
-        return connection.createProxy<MCPServerManager>(MCPServerManagerPath, client);
+        const serverClient = ctx.container.get<MCPServerManagerServerClient>(MCPServerManagerServerClient);
+        const proxy = connection.createProxy<MCPServerManager>(MCPServerManagerPath, client);
+        return new Proxy(proxy, {
+            get(target: MCPServerManager, prop: PropertyKey, receiver: unknown): unknown {
+                if (prop === 'addOrUpdateServer') {
```
sdirix, line 64: We should also override the `getServerDescription` to return the one which was added instead of the one which will come from the backend.

packages/ai-mcp/src/browser/mcp-server-manager-server-client.ts
```diff
@@ -0,0 +1,55 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { injectable } from '@theia/core/shared/inversify';
+import { MCPServerDescription, MCPServerManagerServerClient } from '../common';
+
+@injectable()
+export class MCPServerManagerServerClientImpl implements MCPServerManagerServerClient {
+
+    protected serverDescriptions: Map<string, MCPServerDescription> = new Map();
```
sdirix: At the moment we are just adding `serverDescriptions` but we are never removing them. We should inject the `MCPFrontendNotificationService` and listen to changes and then remove descriptions which no longer exist in the backend, e.g. by calling `getServerNames` and then removing all the ones which were not part of that list.

packages/ai-mcp/src/node/mcp-server-manager-impl.ts
```diff
@@ -59,6 +59,14 @@ export class MCPServerManagerImpl implements MCPServerManager {
         if (!server) {
             throw new Error(`MCP server "${serverName}" not found.`);
         }
+        const description = await server.getDescription();
+        if (description.resolve) {
+            const resolved = await description.resolve(description);
+            const isEqual = JSON.stringify(description) === JSON.stringify(resolved);
+            if (!isEqual) {
+                server.update(resolved);
+            }
```
sdirix, line 68: I think we can just always call `server.update` without this equals check? The update does not really do anything besides setting the description. What the `update` also does is to change the `status` of the server, but this rather seems to be an error anyway. I don't think we should change the status of the server just by updating the description.

packages/plugin-ext/src/main/browser/lm-main.ts
```diff
@@ -130,12 +128,21 @@ export class McpServerDefinitionRegistryMainImpl implements McpServerDefinitionR
...
                 headers: convertedHeaders,
-                autostart: false, // Extensions should manage their own server lifecycle
+                autostart: false,
+                async resolve(description: MCPServerDescription): Promise<MCPServerDescription> {
+                    const resolved = await self.$resolveServerDefinition(handle, definition);
+                    if (resolved) {
+                        return self.convertToMcpServerDescription(handle, resolved);
+                    }
+                    return description;
+                }
```
sdirix, line 142: Do we need to always return the `resolve` field? If there is no resolve on plugin side we could also just skip it, right?

packages/plugin-ext/src/main/browser/lm-main.ts
```diff
@@ -155,6 +162,13 @@ export class McpServerDefinitionRegistryMainImpl implements McpServerDefinitionR
             args: definition.args,
             env: convertedEnv,
             autostart: false, // Extensions should manage their own server lifecycle
+            async resolve(serverDescription: MCPServerDescription): Promise<MCPServerDescription> {
+                const resolved = await self.$resolveServerDefinition(handle, definition);
+                if (resolved) {
+                    return self.convertToMcpServerDescription(handle, resolved);
+                }
+                return serverDescription;
+            }
```
sdirix, line 171: same here

packages/ai-mcp/src/common/mcp-server-manager.ts
```diff
@@ -102,6 +115,22 @@ export interface BaseMCPServerDescription {
...
+     *
+     * @param description The current server description
+     * @returns A promise that resolves to the processed server description
+     */
+    resolve?: (description: MCPServerDescription) => Promise<MCPServerDescription>;
+
+    /**
+     * Optional id for the resolve function. Needed to be able to transfer a resolve function between client and server.
+     */
+    resolveId?: string;
```
sdirix: I would move this field into a special separate interface. Usually these RCP specifics are in some kind of `protocol` file, e.g. in this case `mcp-protocol.ts`

sdirix (changes requested):
Conceptually the refactored PR looks great!

I only have minor comments left

packages/plugin-ext/src/main/browser/lm-main.ts
fipro78: I think we need to. At least I would not know how to solve this, because at this place we do not know if there is a `resolveServerDefinition` function on the  `McpServerDefinitionProvider`. The programmatic registration of MCP servers is different in VS Code compared to Theia. In Theia you register each MCP server separately with a complete mcp server description. In VS Code you have a `McpServerDefinitionProvider` with two methods `provideMcpServerDefinitions` and `resolveMcpServerDefinition`.

packages/plugin-ext/src/main/browser/lm-main.ts
fipro78: we could add a check here, but would it be then consistent with the above where we can't perform the check? I am not sure to be honest. Does it hurt?

packages/ai-mcp/src/common/mcp-server-manager.ts
fipro78: Sorry but I don't get that one. Maybe I am too much of a Java developer, but that field is defined in `BaseMCPServerDescription` which is extended by `LocalMCPServerDescription` and `RemoteMCPServerDescription`. Now what should I do? Create a new interface in some protocol file that extends the local and the remote description interfaces? Or something in between? I really don't understand what I should do here. It doesn't make sense to me from the class inheritence I am used to from the Java world. ;)

packages/ai-mcp/src/node/mcp-server-manager-impl.ts
fipro78: You actually mention the issue if we remove the equals check. In that case we get the updated server description on start, but the status will be shown as disconnected. So the status shown in the user interface is inconsistent. As I am not sure if changing the status is an error or if it was done by intention for the local mcp servers.

packages/ai-mcp/src/browser/mcp-server-manager-server-client.ts
fipro78: OK

packages/ai-mcp/src/browser/mcp-frontend-module.ts
fipro78: Actually I only store the description on the client side if it has a resolve method that needs to be handled. But I will add an appropriate method that first checks for the description on the client, and if none is available check on the server side.

packages/ai-mcp/src/browser/mcp-server-manager-server-client.ts
fipro78: That one turned out to be more complicated, because if I want to use the `MCPServerManager` or any other service that allows to get the server names from the backend, I end up in a cyclic dependency issue when creating the proxy. I think I have found a solution, but I am not sure if this is the correct way of doing this.

packages/ai-mcp/src/browser/mcp-frontend-module.ts
fipro78: It turned out that this would introduce an even bigger change, because we would need to keep the backend and the frontend descriptions in sync somehow. Currently the description in the frontend is only stored to keep the resolve function, the other information, especially the states, are not of interest. If I change this to have a `getServerDescription` that returns the server description from the client, there is an inconsistency, which for example directly shows up in the MCP Server user interface. The action buttons for the servers are missing for those servers with resolve functions, as the status is tracked only in the backend. Do you really think it is necessary to have the getter in the client? That would mean to extend the sync not to only check for the server names, but actually sync all of the descriptions on changes.

packages/ai-mcp/src/browser/mcp-server-manager-server-client.ts
sdirix: I see what you mean. The NotificationService is nice to get the notification but it doesn't help to actually get the descriptions. I think the new way you solved it is good enough. I would not have handed in the full MCPServerManager for syning but just the information we need, but that's a nitpick and does not really require an additional change.

packages/ai-mcp/src/common/mcp-server-manager.ts
sdirix: I pushed my suggested changes to this PR. See `mcp-protocol` for the RCP type, extracting the `resolveId` from the regular server description

packages/ai-mcp/src/node/mcp-server-manager-impl.ts
sdirix: I agree, it's out of scope of this PR

packages/ai-mcp/src/browser/mcp-frontend-module.ts
sdirix: I improved the code by:
- proxied the `getServerDescription` by returning the description from the server manager, only mixing in the resolve function. 
- only store the resolve function in the frontend client, none of the other properties

sdirix (approved):
Thank you very much again for the contribution :heart: 

I applied my remaining comments with a separate commit. Please let me know if you agree with these changes. Thank you!

fipro78:
@sdirix 
Thanks for polishing the PR. Not sure if I understand everything completely, but it looks nice. I just tested it locally with a VS Code extension that contributes MCP servers using a resolve function, and with the api-samples example that is added with this PR. Both work as intended.

PR #16053, opened 2025-07-25
Update compile-references.js package is a reserved keyword

lemanschik:
The code did only work because it got bundled or maybe executed without use strict even when it is annotated with it how ever package is a reserved word so renamed it to pkg

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/scripts/compile-references.js
+++ b/scripts/compile-references.js
@@ -48,9 +48,9 @@ compileTypeScriptReferences().catch(error => {
  * This script main entry point.
  */
 async function compileTypeScriptReferences() {
-    await Promise.all(Object.values(PACKAGE_LIST).map(async package => {
-        const references = await getTypescriptReferences(package);
-        await configureTypeScriptReferences(package, references);
+    await Promise.all(Object.values(PACKAGE_LIST).map(async pkg => {
+        const references = await getTypescriptReferences(pkg);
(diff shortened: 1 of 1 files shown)


msujew (commented):
Hey @lemanschik, thanks for the contribution! You're right, it likely works because we don't run the script with `"use strict"`.

In order to accept your changes, please sign the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/eca/) with the same email that was used to create the commits.

lemanschik:
@msujew oh thats a problem i do not know what email address is used for that commits please simple apply that on your name i do not want to pull that patch-1 branch and run 

    git log --pretty=format:"%ae" -n 1

lemanschik:
in my fork i do clean rewrites anyway via plugins.
```js
  preserveEntrySignatures: true,
  plugins: [
    aliasesTheia,
    patchModulesPlugin({
      [path.resolve('src/theia/scripts/compile-references.js')]: (code, id) => {
        const modified = code.replace('async package =>','async pkg =>').replace('package => package.name','pkg => pkg.name')
        if (modified !== code) {
          console.log('patchModulesPlugin: patched => ',id);
        }
        return modified
      }
    }),
    ```

PR #16067, opened 2025-07-29
Enable specifying original state in change sets

planger:
I'm working on a use case where I have an external agent that I integrate as a Theia AI agent. This external agent modifies workspaces resources directly and only notifies our Theia agent after the change, but it'll store a backup of the original in the file system, which we can obtain.

I want to use this in order to track and show file changes in a change set, so the user has a central place where they see all applied changes. Therefore, I experimented with extending the change set file element with an `originalState` that is passed from the agent, instead of being read from the file system.

### DIFF
--- a/examples/api-samples/src/browser/chat/original-state-test-agent-contribution.ts
+++ b/examples/api-samples/src/browser/chat/original-state-test-agent-contribution.ts
@@ -1,0 +1,169 @@
+// *****************************************************************************
+// Copyright (C) 2024 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import {
+    AbstractStreamParsingChatAgent,
+    ChatAgent,
+    MutableChatRequestModel,
+    MarkdownChatResponseContentImpl,
+    SystemMessageDescription
+} from '@theia/ai-chat';
+import { ChangeSetFileElementFactory } from '@theia/ai-chat/lib/browser/change-set-file-element';
+import { Agent, LanguageModelRequirement } from '@theia/ai-core';
+import { inject, injectable, interfaces } from '@theia/core/shared/inversify';
+import { FileService } from '@theia/filesystem/lib/browser/file-service';
+import { WorkspaceService } from '@theia/workspace/lib/browser';
+
+export function bindOriginalStateTestAgentContribution(bind: interfaces.Bind): void {
+    bind(OriginalStateTestAgent).toSelf().inSingletonScope();
+    bind(Agent).toService(OriginalStateTestAgent);
+    bind(ChatAgent).toService(OriginalStateTestAgent);
+}
+
+/**
+ * This is a test agent demonstrating how to test originalState functionality in change sets.
+ * It creates change set elements with original content provided and tests sequential updates.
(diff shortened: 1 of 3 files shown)


examples/api-samples/src/browser/chat/original-state-test-agent-contribution.ts
```diff
@@ -0,0 +1,169 @@
...
+            return changeComment + content + `\n// Addition from change ${changeNumber}`;
+        }
+
+        // Insert the change comment at the beginning and add some content
+        return changeComment + content + `\n// This line was added by change ${changeNumber} at ${new Date().toISOString()}`;
+    }
+
+    private delay(ms: number): Promise<void> {
+        return new Promise(resolve => setTimeout(resolve, ms));
+    }
```
colin-grant-work: Nit: there's a `wait` utility in core's `promise-util.ts`.

examples/api-samples/src/browser/chat/original-state-test-agent-contribution.ts
```diff
@@ -0,0 +1,169 @@
...
+        // Add a comment at the beginning to show the change
+        const changeComment = `// Modified by Original State Test Agent - Change ${changeNumber}\n`;
+
+        if (content.length < 50) {
+            return changeComment + content + `\n// Addition from change ${changeNumber}`;
+        }
+
+        // Insert the change comment at the beginning and add some content
+        return changeComment + content + `\n// This line was added by change ${changeNumber} at ${new Date().toISOString()}`;
+    }
```
colin-grant-work: Is this difference in behavior depending on text length important? It looks like we sandwich the old content between our two comments either way?

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -135,6 +141,10 @@ export class ChangeSetFileElement implements ChangeSetElement {
     }
 
     protected async obtainOriginalContent(): Promise<void> {
+        if (this.elementProps.originalState) {
+            this._originalContent = this.elementProps.originalState;
+            return;
+        }
```
colin-grant-work: I would probably prefer modifying the line below to be `this._originalContent = this.elementProps.originalState ?? await ...`

That would ensure that we keep the same data flow regardless of source. (And see next comment)

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -175,7 +189,7 @@ export class ChangeSetFileElement implements ChangeSetElement {
             this._readOnlyResource.update({
                 autosaveable: false,
                 readOnly: true,
-                contents: this._originalContent ?? ''
+                contents: this.elementProps.originalState ?? this._originalContent ?? ''
```
colin-grant-work: If we consistently treat `elementProps.originalState` as parallel to the input we would otherwise get from the file system, we should be able to keep the old code, since we'll have assigned the right thing to `this._originalContent`.

packages/ai-chat/src/browser/change-set-file-element.ts
```diff
@@ -254,6 +271,9 @@ export class ChangeSetFileElement implements ChangeSetElement {
      * Ensures initialization is complete before returning the content.
      */
     async getOriginalContent(): Promise<string | undefined> {
+        if (this.elementProps.originalState) {
+            return this.elementProps.originalState;
+        }
```
colin-grant-work: I have mixed feelings about these additions. Here, we can probably omit the addition: initialization will finish instantly by default, but there isn't any harm in waiting for it to definitely be done, and then we'll be returning the right thing when we return `this._originalContent`. In the synchronous case above, there would be a fraction of a tick between construction and the resolution of the `obtainOriginalContent` promise when we could have a problem we didn't need to have, so returning early there seems reasonable.

planger: You are right, these additional "insurances" to return the original content are not necessary and make the code less comprehensible. I removed them.

packages/ai-chat/src/browser/change-set-file-element.ts
planger: Absolutely, these were just additional insurances which aren't necessary, as `_originalContent` will reliably return the correct content, depending on whether it has been initialized directly from `originalState` or from file system.

colin-grant-work (approved):

PR #16068, opened 2025-07-29
fix(chat): cancel on Espace in chat input

planger:
#### What it does

We lost the ability to cancel the current request with Escape in the chat input. With this change we restore this functionality.

#### How to test

Start a chat and push `Escape` while you are in the chat input. It should cancel the current request if one is currently in progress. If no request is in progress, it should do nothing.

#### Follow-ups

None

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
@@ -313,7 +313,10 @@ export class AIChatInputWidget extends ReactWidget {
     }
 
     protected onEscape(): void {
-        // No op
+        const currentRequest = this._branch?.items?.at(-1)?.element ?? this._chatModel.getRequests().at(-1);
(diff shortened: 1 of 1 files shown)


sdirix:
Do we really want to support this? For me this behavior would be unexpected.

planger:
It was working before and we also show it in the [tool tip](https://github.com/eclipse-theia/theia/blob/cc519497b96f1869a08cb1e8431a2fea6ca6d8bb/packages/ai-chat-ui/src/browser/chat-input-widget.tsx#L766) ("Cancel (Esc)").

I personally am in favor of it, but I'm open to any other opinion. If we decide to not like it, we should adapt the tool tip or use another shortcut.

sdirix:
I don't have any hard feelings, I was just surprised. If we even indicate it in the UI, then it should also work :)

sgraband (approved):
Looks good to me. :+1: Just a minor nitpick: The first statement is hard to read imho and could use splitting or a comment.

PR #16071, opened 2025-07-29
15853: draft manually pin agents

CamilleLetavernier:
#### What it does

- replace Monaco-auto-complete (indirect pinning) with explicit popup

Manual pinning is tricky because the current implementation is completely unrelated to pinning: it inserts the '@' symbol, then triggers Monaco auto-completion. Implicit assumptions:

- Monaco will show a list of Agent IDs to insert after '@'
- Explicitly mentioning an Agent will pin it (assuming auto-pin is enabled)

The only solution I see to get a proper "manual pin", independent from the auto-pin setting, is to replace the Monaco auto-completion with a custom Popup showing the list of agents, and then pinning the selection (much more explicit behavior with no external assumption). However this is a more complex solution (I will open a draft PR for this and we can decide if we want to proceed with this approach, and refine it).

#### How to test

- Press "@" in the AI Chat view
- The agent pinning/unpinning should work properly

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-input-widget.tsx
@@ -14,7 +14,7 @@
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
 import {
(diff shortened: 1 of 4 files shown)


ndoschek:
Closing this one for the moment, we can reopen/revisit if it comes up again.

PR #16078, opened 2025-07-29
Improve PreferenceWidget state restoration on start and scope change

### LINKED ISSUE #15915: Settings window: moving from user scope to workspace scope closes the selected setting
### Bug Description:
Hi,
In the latest Theia version (I am upgrading from 1.57 to 1.63), there is an issue in the settings window. When selecting a subcategory and moving from the user scope to the workspace scope, the selection is reset (also the categories on the left will collapse), and the view changes to the first setting item, resulting in a poor user experience.

### Steps to Reproduce:

1. Open the settings window
2. e.g. select the Debug subcategory under the Feature section
3. Move to the workspace scope
4. The categories on the left will collapse, and the selection will change to the first item.

https://github.com/user-attachments/assets/b1c48e4d-8000-4d30-90cd-5fabc8bb57b7

### Additional Information

- Operating System: Mac
- Theia Version: 1.63

colin-grant-work:
#### What it does

Fixes #15915 and improves state restoration on start.

#### How to test

1. Scroll to somewhere in the middle of the preferences.
2. Switch scopes (e.g. user <--> workspace)
3. You should remain scrolled to the same preference, or the top if the preference to which you were scrolled is not available in the new scope.
4. Restart the application.
5. The preferences view should scroll to the same area, and the tree on the left should reflect the preference shown.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/browser/preference-tree-model.ts
+++ b/packages/preferences/src/browser/preference-tree-model.ts
@@ -251,7 +251,8 @@ export class PreferenceTreeModel extends TreeModelImpl {
      */
     selectIfNotSelected(node: SelectableTreeNode): boolean {
         const currentlySelected = this.selectedNodes[0];
-        if (node !== currentlySelected) {
+        if (!node.selected || node !== currentlySelected) {
+            node.selected = true;
             this.selectNode(node);
             return true;
         }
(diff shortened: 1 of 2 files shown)


sdirix (approved):
Works for me!

I have two suggestions regarding the use case that a node is selected which does not exist in the other scope:
 - Personally I find it a bit jarring that we jump to the top. I would have liked to at least stay somewhat in the area of where I was before, e.g. jump to the nearest node which still exists
 - When this jump to the top happens and I do absolutely nothing besides switching back to the scope where I was coming from, I would have liked to be back at the place where I was coming from.

These don't need to be tackled here or at all, just my 2 cents.

sdirix (approved):
I like it. Thanks for the additional improvement.

PR #16080, opened 2025-07-30
[draft] support vscode extension debug

rschnekenbu:
#### What it does

Note: This PR is largely based on work done in the context of https://github.com/eclipsesource/theia/pull/96. 

It tries to get away from using non-working npm based commands to debug vscode extensions from Theia (see  https://github.com/eclipse-theia/theia-blueprint/issues/ 281 (remove the whitespace when doing the final PR))

#### How to test

1.  Create a packaged electron version either via electron-builder or by linking the packages into the [theia-ide](https://github.com/eclipse-theia/theia-blueprint). I used the following contribution to be able to build Theia IDE based on a local version of theia: https://github.com/eclipse-theia/theia-ide/blob/jf/local-theia-build/local-theia-build.md. 
2. Create a workspace with a [vscode extension with the yeoman generator](https://code.visualstudio.com/api/get-started/your-first-extension)
3. Open the workspace with the theia application
You will need to slightly adapt the launch.json, due to other bugs. Comment out the preLaunchTask line
4. Run the Run extension debug config.
This is the part where it fails. A second window opens, but as the initial tool does not detect the start of the second app (looking for 'Theia app listening' pattern in logs), the debug session fails. 

#### Follow-ups

I did not fully rework on the filtering mechanism as commented by Thomas on https://github.com/eclipsesource/theia/pull/96. The filtering is a kind of mess between args or undefined being returned instead of true/false for example.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-dev/src/node/hosted-instance-manager.ts
+++ b/packages/plugin-dev/src/node/hosted-instance-manager.ts
@@ -16,14 +16,15 @@
 
 import { RequestOptions, RequestService } from '@theia/core/shared/@theia/request';
 import { inject, injectable, named } from '@theia/core/shared/inversify';
+import { MessageService, isWindows } from '@theia/core';
 import * as cp from 'child_process';
 import * as fs from '@theia/core/shared/fs-extra';
 import * as net from 'net';
 import * as path from 'path';
+import * as os from 'os';
 import URI from '@theia/core/lib/common/uri';
 import { ContributionProvider } from '@theia/core/lib/common/contribution-provider';
 import { HostedPluginUriPostProcessor, HostedPluginUriPostProcessorSymbolName } from './hosted-plugin-uri-postprocessor';
-import { environment, isWindows } from '@theia/core';
 import { FileUri } from '@theia/core/lib/common/file-uri';
 import { LogType } from '@theia/plugin-ext/lib/common/types';
 import { HostedPluginSupport } from '@theia/plugin-ext/lib/hosted/node/hosted-plugin';
@@ -66,7 +67,7 @@ export interface HostedInstanceManager {
      * Terminates hosted plugin instance.
      * Throws error if instance is not running.
      */
-    terminate(): void;
+    terminate(): Promise<void>;
 
     /**
      * Returns uri where hosted instance is run.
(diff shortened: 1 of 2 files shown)


packages/plugin-dev/src/node/hosted-instance-manager.ts
```diff
@@ -239,29 +279,96 @@ export abstract class AbstractHostedInstanceManager implements HostedInstanceMan
...
+
+    protected async getStartCommand(port: number, debugConfig?: PluginDebugConfiguration): Promise<string[]> {
+        const processArguments = process.argv;
+
+        const command = processArguments.filter((arg, index, args) => {
+            // remove --port=X and --port X arguments if set
+            // according to process.argv documentation, the first argument is the path to the node executable
+            // and the second argument is the path to the script being executed
+            // second argument will be treated as the workspace location, so it should be ignored
+            if (index === 1 || arg.startsWith('--port') || args[index - 1] === '--port') {
```
tsmaeder, line 358: `filter()` expects a boolean return, unlike `map()`

rschnekenbu:
Quick update on this task:
- Thanks to an adapted version of https://github.com/eclipse-theia/theia-ide/blob/jf/local-theia-build/local-theia-build.md, it is possible to build quite easily (takes time however!) a version of theia IDE with a local version of Theia. Changes consist of updating this line and the following line https://github.com/eclipse-theia/theia-ide/blob/046f5a2c6fa5f51ebe33750cec51b97ee9b6714c/docker/local-theia-build/entrypoint#L44 with the npm equivalent commands since Theia now builds with npm (yarn => npm & yarn build => npm run build)
- The run extension task consist of compiling the extension with a pre-build task, and then the tool tries to trigger a new instance of theia with the HOSTED_PLUGIN to be tested. The debugging application [listens](https://github.com/eclipse-theia/theia/blob/remi/run-extension/packages/plugin-dev/src/node/hosted-instance-manager.ts#L413) to the [line](https://github.com/eclipse-theia/theia/blob/remi/run-extension/packages/core/src/node/backend-application.ts#L300) "Theia app listening" on the output of the debugged instance to know it is started and should connect the debugger. 
- Currently, I could not get the debugged instance to trigger a new backend and a new frontend, but only a new frontend. Thus the output never receives "Theia app listening[...]" output and as such, never connects the debbuger. Also, the hosted plugin is not present in the list of installed plugins.

According to the documentation, we should be able to spawn a new backend if we provide a new electron user data area, but this is not sufficient. I tried also providing a theia config dir using some environment variables on the new spawned process, with no success.

ndoschek:
Closing it for now and continue from here later when we revisit the topic.

Issue #15931, opened 2025-07-01
[Epic] Debugger Usability and Stability Problems


ndoschek:
### Summary:

Multiple issues currently impact the debugging experience in Theia (IDE).
This umbrella issue is intended to track, discuss, and coordinate ongoing investigation and improvement efforts. Individual sub-issues may exist or be created for specific root causes as identified.

### Description:

1. Breakpoints appear as not bound, although they are (see https://github.com/eclipse-theia/theia/issues/14882)
Breakpoints are displayed as "not bound" even though they are actually hit during the debugging session.
Breakpoint binding appears to be thread-dependent: when a breakpoint is hit in the frontend, other frontend breakpoints become visibly bound, and similarly for backend breakpoints. However, breakpoints are only shown as active when the main thread is selected, and since threads are collapsed by default, selecting a thread does not immediately update the breakpoint status.
The sub-issue is already tracking this specific behavior and contains more observations.

2. Debug View steals focus, especially at startup - fixed ✅ 
When starting a backend and attempting to interact with the explorer or other parts of the IDE, the debug view or debug console repeatedly steals focus, making it nearly impossible to access other UI elements (like the Explorer view) on the host during startup. This focus issue appears to stabilize after the backend has finished starting, but is significantly disrupting.

3. Debug session restarts after pausing too long at a breakpoint
If a user pauses at a breakpoint for an extended period of more than 60 seconds (e.g., to investigate variables), the debug session appears to reset: the output states "Waiting for the debugger to disconnect" and the program seems to be running again. Expanding the thread shows it is still paused at the previous location (as described in the breakpoint issue from item 1).
If the user continues execution from here, the application restarts.

4. If startup of a launch config fails, the error notification indicates to `Debug session initialization failed. See console for details.`. However, the launch config gets removed from the debug console's launch config picker (select component) and hence the output cannot be viewed by the user. Even when a launch is not running anymore, we should keep the recently started launch configs in the select component to be able to select it and view the output.

This umbrella issue is meant for discussion and tracking and more sub issues can be extracted of course anytime.

### Additional Information

- Operating System: all
- Theia Version: most recent

Issue #14698, opened 2025-01-06
Support VS Code's `Run to Cursor` command


msujew:
### Feature Description:

After implementing the `Jump to Cursor` command (see https://github.com/eclipse-theia/theia/issues/14500), I've noticed that we're also missing the `Run to Cursor/Line` command. See also [here](https://github.com/microsoft/vscode/blob/b7f437d2a1974e160f3ec00b9e72526d27bc4fb9/src/vs/workbench/contrib/debug/browser/debugEditorActions.ts#L320-L339) and [here](https://github.com/microsoft/vscode/blob/b7f437d2a1974e160f3ec00b9e72526d27bc4fb9/src/vs/workbench/contrib/debug/browser/debugService.ts#L1318-L1372) for more info on how it's implemented in VS Code.

pisv:
I'd be interested in working on this issue.

@ndoschek Can you assign it to me, please?

Issue #13728, opened 2024-05-23
Debugger doesn't show error of setVariable if it failed


EvilBeaver:
### Bug Description:

Debugger can have ability to set value of a variable. Debug Adapter Protocol has `setVariable` command for this.
But if evaluation of new value fails (for example, if user entered non-valid expression) theia ignores result and error message from Debug Adapter and only logs error to console. As result, user doesn't see any appropriate reaction on his action

### Steps to Reproduce:

1. Launch JS debugging
2. stop at some breakpoint
3. in Variables panel right-click and choose `Set value`
4. enter an invalid JS expression, for example some syntax error
5. Nothing happens in UI. Only in console.log we see message "ERROR setValue failed..."

### Additional Information

Expected behavior:  This is how it works in VSCode
![image](https://github.com/eclipse-theia/theia/assets/2657561/8228af4b-6a9c-447c-a343-d47128b88c07)

Code: https://github.com/eclipse-theia/theia/blob/master/packages/debug/src/browser/console/debug-console-items.tsx#L206

- Operating System: Windows 10
- Theia Version: 1.49.1

EvilBeaver:
Hi, guys, any chances of fixing this?

pisv:
I'd be interested in working on this issue.

@ndoschek Can you assign it to me, please?

PR #16154, opened 2025-08-10
Fix breakpoints being lost when editing source

### LINKED ISSUE #16153: Lost breakpoints when editing source
### Bug Description:

When two or more breakpoints are on the same line, only one of them survives an affecting edit of the source.

### Steps to Reproduce:

1. Set two or more breakpoints on the same line.

Image

2. Add a new line before that line.

Image

Observe that only one of the breakpoints remains.

### Additional Information

I'll submit a fix in a short while.

- Operating System: all
- Theia Version: 1.64

pisv:
#### What it does

Fixes #16153.

#### How to test

Verify that #16153 is now fixed using the steps to reproduce the issue.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/editor/debug-editor-model.ts
+++ b/packages/debug/src/browser/editor/debug-editor-model.ts
@@ -314,19 +314,24 @@ export class DebugEditorModel implements Disposable {
     }
     protected createBreakpoints(): SourceBreakpoint[] {
         const { uri } = this;
-        const lines = new Set<number>();
+        const positions = new Set<string>();
         const breakpoints: SourceBreakpoint[] = [];
         for (const decoration of this.breakpointDecorations) {
             const range = this.editor.getControl().getModel()?.getDecorationRange(decoration);
-            if (range && !lines.has(range.startLineNumber)) {
+            if (range) {
                 const line = range.startLineNumber;
                 const column = range.startColumn;
                 const oldBreakpoint = this.breakpointRanges.get(decoration)?.[1];
-                const isLineBreakpoint = oldBreakpoint?.raw.line !== undefined && oldBreakpoint?.raw.column === undefined;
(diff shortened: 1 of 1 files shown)


pisv:
@ndoschek Thank you for your thorough review! 👍

> I noticed another issue, but definitely not introduced by this change: when we have a line breakpoint alongside inline breakpoints and insert new lines above, the line breakpoint does not shift accordingly to the new line.

Good catch! I could reproduce the issue using the specific steps you have given, but could not reproduce it using the simple `hello.js` file from #16153 (as the screenshots show). That's why I didn't notice it at first.

Debugging reveals that it is a separate issue, which I think would be better to track and fix separately, so that, e.g., the commit could be reverted on its own if needed. In the meantime, I'll surely take a deeper look at this issue and try to address it in a separate PR.

> Just to note: what i described mainly affects the editor while not actively debugging. There are still some general issues during debugging, like with bound breakpoints where line updates can be laggy. We're tracking those under https://github.com/eclipse-theia/theia/issues/15931, as there are several more problems impacting the overall debug experience in case you want to have a look.

Indeed, it seems that there are still quite a few opportunities for improving the debugging experience in Theia. I'd be currently interested in working on #14698 and #13728, if help is wanted. Hopefully, I might be able to help with some of the other debug-related issues as well, although I cannot make a specific promise now.

Thanks!

ndoschek (approved):
Thank you @pisv for checking and confirming, that makes sense 👍  
I'll go ahead and track the line breakpoint issue separately (Edit: see issue #16177).

Great to hear you’re interested in the other debug-related tickets, any help there is much appreciated 🙌  
Just let me know if you'd like me to assign any of them to you (once you comment on the issue I can assign you). Thanks!

PR #16179, opened 2025-08-14
feat(chat): cancel incomplete requests on new request

planger:
#### What it does

If a user sends a new request to a session that currently as an incomplete request, we should always cancel those first before adding a new request. Otherwise the state becomes very unclear (what was the context of the new request while the past ones are still generating).

This also makes it more intuitive when submitting prompts while a previous request is waiting for input or generating, as we show the cancel button in the input.

#### How to test

* Start a longer chat request
* Without canceling, just enter another request into the chat input and hit enter

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-service.ts
+++ b/packages/ai-chat/src/common/chat-service.ts
@@ -233,6 +233,8 @@ export class ChatServiceImpl implements ChatService {
             return undefined;
         }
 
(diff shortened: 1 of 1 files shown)


CamilleLetavernier (approved):
Looks good to me :+1:

ndoschek (approved):
Thanks!

PR #16181, opened 2025-08-14
Fix: correctly handle plugin-icons during menu creation and use tooltip

### LINKED ISSUE #15884: `view/title` menu tooltips from VS Code extensions not shown in Theia 1.62.2
### Bug Description:
In the Theia IDE version 1.62.200, I noticed that tooltips for commands contributed via `view/title` menus (from VS Code extensions) are no longer displayed when hovering over the buttons in the view header.

This used to work in earlier versions of Theia (unfortunately I don't know the exact version where it broke, but I’m fairly confident this is a regression).

### Steps to Reproduce:
1. Install/Use a VS Code extension that contributes a `view/title` menu command, referring to a `command` contribution with a `title` property.
1. Open the contributed view in Theia IDE.
1. Hover over the button in the title bar.

Image

### Expected Behavior
The tooltip (i.e. the `title` value of the corresponding command used in `contributes.menus["view/title"]`) is shown when hovering the mouse over the button.

### Actual Behavior
No tooltip is shown.

### Additional Information

- Theia Version: 1.62.2 (Theia IDE 1.62.200)

eneufeld:
#### What it does

Fixes #15884

All icon classes were send through the `codicon` function which lead to classNames for which no css rule existed thus breaking plugin contributed icons. Also tooltip is used for instead of text to provide the title for menus.

#### How to test

Install a vscode extension that contributes an icon to a toolbar menu, eg Git Graph.
On main there is an empty space in the SCM View Toolbar which you can click. With the fix the Icon is shown again.
If you use the GitLens extension you will see that toolbar items now have a tooltip if you hover over them. The broken case is shown in the bug report.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar-menu-adapters.tsx
+++ b/packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar-menu-adapters.tsx
@@ -97,11 +97,11 @@ abstract class AbstractToolbarMenuWrapper {
     protected renderMenuItem(widget: Widget): React.ReactNode {
         const icon = this.icon || 'ellipsis';
         const contextMatcher: ContextMatcher = this.contextKeyService;
+        const className = `${icon} ${ACTION_ITEM}`;
         if (CompoundMenuNode.is(this.menuNode) && !this.menuNode.isEmpty(this.effectiveMenuPath, this.contextKeyService, widget.node)) {
-
             return <div key={this.id} className={TabBarToolbar.Styles.TAB_BAR_TOOLBAR_ITEM + ' enabled menu'}>
-                <div className={codicon(icon, true)}
-                    title={this.text}
+                <div className={className}
+                    title={this.tooltip || this.text}
                     onClick={e => this.executeCommand(e)}
                 />
                 <div className={ACTION_ITEM} onClick={event => this.showPopupMenu(widget, this.menuPath!, event, contextMatcher)} >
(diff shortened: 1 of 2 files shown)


packages/core/src/browser/shell/tab-bar-toolbar/tab-bar-toolbar-menu-adapters.tsx
```diff
@@ -97,11 +97,21 @@ abstract class AbstractToolbarMenuWrapper {
...
+            This class was created from scratch during a big refactoring.
+            Some logic seems to have been lost in that process, so this is a bit of a workaround now.
+            The provided icon here can come from different sources:
+            - somebody might have created the icon and set an already valid codicon icon
+            - the icon comes from a plugin contribution so it is named `plugin-icon-${n}` with n being an integer
+            - something else which needs to be wrapped
+            So in the first two cases we don't want to wrap our class name with the codicon function.
+            Furthermore the title of a menuItem is only shown on hover, so the tooltip fits better and is not set to undefined as text is
+        */
+        const className = icon.startsWith('codicon') || icon.startsWith('plugin-icon-') ? `${icon} ${ACTION_ITEM}` : codicon(icon, true);
```
tsmaeder: Wouldn't it make more sense to always treat the `icon` property as a class name and apply the `codicon()` function at the point where the rendered item is created?

msujew: I agree, that would make this much more robust.

eneufeld: I'll take a look whether I can find the location where the items are created.

eneufeld: I'll updated the code. I found only one place where the codicon call was missing

ndoschek (approved):
Thanks @eneufeld, the fix works great, I tested it with different extensions (e.g., GitLens, Continue, Git Graph, Roo Code) and tooltips work as expected now :+1: 

Regarding the SCM view toolbar icon you mentioned in the "How to test" section, I wasn’t able to verify that part, I couldn’t locate the icon or the empty space in the source control view. Still, the change looks solid and non-critical, so I’ll go ahead and approve it.

Issue #16107, opened 2025-08-01
OPFS doesn't work properly in Safari, it can read files but can't create or write files or folders


kachurun:
### Bug Description:

Safari doesn't support [FileSystemFileHandle->createWritable](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle/createWritable) for now.

However, there's a possible workaround: we can create a separate web worker and use createSyncAccessHandle() (available only in dedicated web workers).

I used this code as a proof of concept:

```
self.onmessage = async (event) => {
    const { type, payload } = event.data;

    if (type === 'write') {
        const root = await navigator.storage.getDirectory();
        const fileHandle = await root.getFileHandle(payload.path, { create: true });
        const accessHandle = await fileHandle.createSyncAccessHandle();

        const encoder = new TextEncoder();
        const data = encoder.encode(payload.contents);

        accessHandle.write(data, { at: 0 });
        accessHandle.flush();
        accessHandle.close();

        self.postMessage({ ok: true });
    }

    if (type === 'read') {
        const root = await navigator.storage.getDirectory();
        const fileHandle = await root.getFileHandle(payload.path);
        const accessHandle = await fileHandle.createSyncAccessHandle();

        const size = accessHandle.getSize();
        const buffer = new Uint8Array(size);
        accessHandle.read(buffer, { at: 0 });
        accessHandle.close();

        const text = new TextDecoder().decode(buffer);
        self.postMessage({ ok: true, text });
    }
}
```

And from the main thread:

```
const worker = new Worker('/opfs.worker.js');

worker.postMessage({
    type: 'write',
    payload: {
        path: 'notes.txt',
        contents: 'Hello Safari 🧃'
    }
});

worker.postMessage({
    type: 'read',
    payload: {
        path: 'notes.txt'
    }
});
```

### Additional Information

- Safari Version: 18.5
- Theia Version: 1.64.0

Related issue: https://github.com/eclipse-theia/theia/issues/12852

kachurun:
It's very likely I can help solve this if someone can provide info on how to bundle a worker script into the frontend directory, since this isn't a common case in Theia (the only web worker I see is editor.worker from Monaco).

msujew:
Hey @kachurun, you can add the webworker bundling instruction into the generated webpack config. If you want to add this feature into the framework, you'll need to adjust the webpack code generator.

kachurun:
@msujew Cool, thank you for the fast reply!

[...]

PR #14806, opened 2025-02-01
Browser-only file search

robertjndw:
#### What it does
Currently, browser-only file search was not possible due to the lack of `@vscode/ripgrep` in an environment without Node support. This PR implements file search using the browser-only FileSystemProvider and [`minimatch`](https://www.npmjs.com/package/minimatch).
This implementation was necessary because some VSCode API weren't working correctly for Web Extensions (e.g. `vscode.workspace.findFiles()`)

#### How to test
1. Open a in-browser Project multiple files
2. Hit `CMD + P` to open the file search
3. Enter a file name or a pattern

#### Follow-ups
The workspace search is currently not working for a similar problem as it uses `@vscode/ripgrep` too.

#### Breaking changes

- [ ] This PR introduces breaking changes and requires careful review. If yes, the breaking changes section in the [changelog](https://github.com/eclipse-theia/theia/blob/master/CHANGELOG.md) has been updated.

#### Review checklist

- [X] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #16187, opened 2025-08-18
[browser-only] FileSystem improvements

kachurun:
#### What it does
1. Fixed the issue with the browser-only Theia not working in Safari versions before 26 Beta (https://github.com/eclipse-theia/theia/issues/16107). It also generally stabilizes the OPFS API because I found many issues when trying to perform operations in parallel. I'm using navigator.locks internally to ensure there are no conflicts between operations.
2. Implemented file watching (https://github.com/eclipse-theia/theia/issues/12852). 
3. Implemented folder copying (https://github.com/eclipse-theia/theia/issues/12852).
4. Implemented files/folders download (https://github.com/eclipse-theia/theia/issues/12852).
5. Implemented files/folders upload (https://github.com/eclipse-theia/theia/issues/12852).
6. Updated and deduplicated minimatch to the latest version.

#### How to test

**All file operations should work** (including renaming and copying folders with content), even in Safari.

**File watching now works.** Theia can detect changes made in other tabs (but not from external tools like the OPFS Explorer Extension in the browser inspector):

Video: [Before](https://github.com/user-attachments/assets/0199ffbd-03b1-4648-86f4-bb4f93620d83) | [Now](https://github.com/user-attachments/assets/7920018d-f1e0-45c6-b067-ddf678be3acd)

**Files can be downloaded.** Folders are downloaded as TAR archives. The copy download link functionality is disabled. If the browser supports `showSaveFilePicker`, it will ask where to save the file before starting; otherwise, it falls back to downloading as a blob (which uses more memory). By default, file downloads exclude files in node_modules, dist, .git, and other folders.

**Files/Folders can be uploaded.** This is memory-efficient, so you can upload even very large files without issues.

#### Follow-ups

Fixed "Search In Files" and "Find In Files":
https://github.com/eclipse-theia/theia/pull/16231

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@
 - [ai-chat-ui] add chat input history with arrow key navigation [#16174](https://github.com/eclipse-theia/theia/pull/16174) - contributed on behalf of STMicroelectronics
 - [dependencies] increased minimum node version to 20.x [#16218](https://github.com/eclipse-theia/theia/pull/16218) - Contributed on behalf of STMicroelectronics
 - [workspace] fix: cannot open same folder once declined exit confirmation [#16238](https://github.com/eclipse-theia/theia/pull/16238)
-- [filesystem] browser-only: fixed Safari compatibility issues with OPFS API, implemented file watching, folder copying, file/folder download and upload capabilities with streaming support [#16187](https://github.com/eclipse-theia/theia/pull/16187)
 
 <a name="breaking_changes_1.65.0">[Breaking Changes:](#breaking_changes_1.65.0)</a>
 


JonasHelming:
@sdirix

package-lock.json
```diff
@@ -31545,12 +31597,91 @@
...
+      "integrity": "sha512-xRc4oEhT6eaBpU1XF7AjpOFD+xQmXNB5OVKwp4tqCuBpHLS/ZbBDrc07mYTDqVMg6PfxUjjNp85O6Cd2Z/5HWA==",
+      "license": "MIT",
+      "dependencies": {
+        "mime-db": "^1.54.0"
+      },
+      "engines": {
+        "node": ">= 0.6"
+      }
+    },
+    "packages/filesystem/node_modules/minimatch": {
```
sdirix: Please make sure that all hoistable packages are actually hoisted

packages/filesystem/src/browser-only/opfs-filesystem-provider.ts
```diff
@@ -1,5 +1,5 @@
 // *****************************************************************************
-// Copyright (C) 2024 EclipseSource and others.
+// Copyright (C) 2025 EclipseSource, Maksim Kachurin and others.
```
sdirix: In the Theia project we usually do not update the header copyright notice

packages/filesystem/src/browser-only/opfs-filesystem-provider.ts
```diff
@@ -222,125 +268,53 @@ export class OPFSFileSystemProvider implements FileSystemProviderWithFileReadWri
...
+        else {
+            changeType = FileChangeType.UPDATED;
         }
+        
+        this.onDidChangeFileEmitter.fire([{ resource, type: changeType }]);
     }
 }
 
-function toFileSystemProviderError(error: DOMException, is_dir?: boolean): FileSystemProviderError {
+function toFileSystemProviderError(error: any): FileSystemProviderError {
```
sdirix: If possible avoid `any`. If the type is not known, `unknown` is preferred

packages/filesystem/src/browser-only/opfs-filesystem-provider.ts
kachurun: Ok, I'll remove it. We're only adding copyright headers with our own names or companies to new files, right?

packages/filesystem/src/browser-only/opfs-filesystem-provider.ts
kachurun: Thats a good point, I probably missed that

kachurun:
> I will be able to take a look early September if this is sufficient for you. Please make sure that the lint check is successful and the package-lock properly updated.

Yes, it's good, but I guess it will be bigger in September. I'm going to improve other FS-related methods and capabilities for browser-only, and also fix Search in Files, which currently doesn't work at all in the browser-only build. Should I make all these changes in the same branch/PR (this one), or is it better to create multiple smaller PRs?

I prefer to continue working here and focus on general browser-only improvements. I don't think it will add many more changes.

packages/filesystem/src/browser-only/opfs-filesystem-provider.ts
sdirix: Yes :+1:

sdirix:
Generally speaking, smaller PRs are nicer to review, iterate and merge. However this PR is pretty small so adding a few more things until September is fine by me. Especially if the PRs would be conflicting.

Note that for file-search there is already a PR in progress here: https://github.com/eclipse-theia/theia/pull/14806

kachurun:
> Generally speaking, smaller PRs are nicer to review, iterate and merge. However this PR is pretty small so adding a few more things until September is fine by me. Especially if the PRs would be conflicting.

Yes, totally agree!

> Note that for file-search there is already a PR in progress here: #14806

Oh wow, but it's so stale. How can we push it forward? I'll do my best to get it done if needed. But it also looks like it searches files by path, not by content within the files.

package-lock.json
kachurun: Hmm, I checked the version twice, it's all good, and I also ran `npm dedupe`, so hopefully that helps.

dev-packages/request/src/node-request-service.ts
```diff
@@ -98,7 +98,7 @@ export class NodeRequestService implements RequestService {
                 path: endpoint.pathname + endpoint.search,
                 method: options.type || 'GET',
                 headers: options.headers,
-                agent: options.agent,
+                agent: options.agent as unknown as http.Agent,
```
kachurun: Weird, but I can't build Theia without this. Looks like the node types were updated or something.

packages/filesystem/src/browser-only/opfs-filesystem-initialization.ts
```diff
@@ -19,18 +19,21 @@ import { injectable } from '@theia/core/shared/inversify';
...
 
 @injectable()
 export class DefaultOPFSInitialization implements OPFSInitialization {
+    getBroadcastChannel(): BroadcastChannel {
+        return new BroadcastChannel('theia-opfs-events');
+    }
 
-    getRootDirectory(): Promise<FileSystemDirectoryHandle> {
-        return navigator.storage.getDirectory();
+    getRootDirectory(): Promise<string> | string {
```
kachurun, line 33: The FileSystemDirectoryHandle type is no longer supported. However, the current code that calls getRootDirectory still checks for this type and retrieves its name for backward compatibility. This approach isn't reliable because you can't get the full path from a FileSystemDirectoryHandle, only the folder's own name.

packages/filesystem/src/browser-only/opfs-filesystem-provider.ts
```diff
@@ -21,36 +21,76 @@ import {
...
+        this.onDidChangeFileEmitter
+    );
+
     @postConstruct()
     protected init(): void {
         const setup = async (): Promise<true> => {
-            this.directoryHandle = await this.initialization.getRootDirectory();
+            const rootHandler = await this.initialization.getRootDirectory();
+            // NOTE: FileSystemDirectoryHandle here for backward compatibility
+            const root = (typeof rootHandler === 'string') ? rootHandler : (rootHandler as FileSystemDirectoryHandle).name;
```
kachurun: I prefer to remove this mess with FileSystemDirectoryHandle if it's acceptable to break compatibility with the old format of getRootDirectory

kachurun:
@sdirix
Added OpenReadWriteClose, Update, and FileReadStream capabilities. Now the OPFS provider is very close to the DiskProvider.
Also, I fixed all issues from the review.
So I think the filesystem part is done and can be passed to testing. I'll test it first, I'm pretty sure it has a lot of bugs after the last changes.
But most probably any further improvements or bug fixes I will do in the OPFSFileSystem class within the opfs-worker package.

Next I'm planning to find a way to fix Search-In-Files (cmd-shift-f) and Find in files (https://github.com/eclipse-theia/theia/pull/14806). I'll do it in a separate branch.

kachurun:
However, I have an idea about the Home folder.
Right now, the OPFS worker treats the home folder `~/` as the OPFS root, which creates a problem – we end up with files like settings.json or toolbar.json in the root of our filesystem.

I can add an option to the OPFS worker to specify a special path for the home directory, like this:

```
new OPFSFileSystem({
  root: '/projects/',
  home: '/home/theia/',
  trash: '/tmp/trash/' // <- why not
})
```

I could also implement trash functionality. However, I don’t really understand how Theia restores files from trash – I mean, how to make it possible to restore a removed file with `cmd+z`.

ndoschek:
Hi @kachurun, thanks a lot for the updates!  
Just a quick heads-up: @sdirix is on vacation this week and will likely continue the review next week.

kachurun:
It's okay, I'm implementing a search in workspace right now

kachurun:
Following up – Search in workspace
https://github.com/eclipse-theia/theia/pull/16231

kachurun:
Added support for downloading files and folders

packages/filesystem/package.json
```diff
@@ -71,7 +75,8 @@
     "watch": "theiaext watch"
   },
   "devDependencies": {
-    "@theia/ext-scripts": "1.64.0"
+    "@theia/ext-scripts": "1.64.0",
+    "@types/tar-stream": "^3.1.4"
```
sdirix: These types should be added to the "dependencies" as the other type dependencies. Otherwise adopters might get type issues in their builds.

kachurun:
Added support for uploading files and folders

kachurun:
@sdirix @ndoschek
I think it’s ready for review and testing 🎉 All file management operations should be working now.

kachurun:
Guys?

sdirix:
Hi @kachurun, I will be able to review the PR this week

packages/core/src/common/stream.ts
```diff
@@ -716,3 +717,34 @@ export function transform<Original, Transformed>(stream: ReadableStreamEvents<Or
...
+            while (true) {
+                const { value, done } = await reader.read();
+                if (done) {break;}
+
+                if (value?.byteLength) {
+                    await ws.write(BinaryBuffer.wrap(value));
+                }
+            }
+            ws.end();
+        } catch (e: any) {
```
sdirix: Nit: Whenever possible we should avoid `any`. `unknown` is a better fit here

packages/filesystem/package.json
```diff
@@ -4,17 +4,20 @@
...
-    "@types/multer": "^1.4.7",
+    "@types/body-parser": "^1.19.6",
+    "@types/multer": "^2.0.0",
     "@types/tar-fs": "^1.16.1",
-    "async-mutex": "^0.3.1",
-    "body-parser": "^1.18.3",
-    "http-status-codes": "^1.3.0",
-    "minimatch": "^5.1.0",
+    "async-mutex": "^0.5.0",
+    "body-parser": "^2.2.0",
```
sdirix: Is there a specific reason to raise the version on all these packages? In Theia we usually only update dependencies when they are needed for the fix/feature at hand. For maintenance updates to keep with the latest states of our dependencies we use separate PRs.

packages/filesystem/src/browser-only/download/file-download-service.ts
```diff
@@ -0,0 +1,361 @@
...
+    '**/build/**',
+    '**/.cache/**',
+    '**/.DS_Store',
+    '**/Thumbs.db',
+    '**/desktop.ini',
+    '**/*.tmp',
+    '**/*.temp',
+    '**/.vscode/**',
+    '**/.idea/**',
+];
```
sdirix: Seems weird to hard code these exceptions. Personally I would be rather surprised to have some arbitrary files and directories not included in my downloads. The user is not aware of this list and can not configure it, so it's a surprise and might be discovered much later, when the data is already gone.

Note that the regular Theia browser variant seems to include all files in the Download

packages/filesystem/src/browser-only/download/file-download-service.ts
```diff
@@ -0,0 +1,361 @@
...
+        let isArchive: boolean = true;
+
+        if (stat.isDirectory) {
+            // Create tar archive for directory
+            data = await this.createDirectoryArchive(uri, stat.name, abortSignal);
+            isArchive = true;
+        } else {
+            // Download single file
+            const content = await this.fileService.readFile(uri);
+            data = new Blob([content.value.buffer], { type: 'application/octet-stream' });
```
sdirix: Nit: With a later Typescript version (`5.9.2`) an error is thrown here:
```
Type '"SharedArrayBuffer"' is not assignable to type '"ArrayBuffer"'.ts(2322)
```

The error does not occur with Typescript `5.4.5` used by the project. However once we update, this issue will need to be solved. Would be nice if this line could be adapted to play nice with a later Typescript too. 

Can you do this or is there a reason this will not be necessary, e.g. with the update of Typescript some other types should also be updated and the issue would be gone...

packages/filesystem/src/browser-only/download/file-download-service.ts
```diff
@@ -0,0 +1,361 @@
...
+                    return;
+                }
+                chunks.push(chunk);
+            });
+
+            tarPack.on('end', () => {
+                cleanup();
+                abortSignal.removeEventListener('abort', onAbort);
+                try {
+                    const blob = new Blob(chunks, { type: 'application/x-tar' });
```
sdirix: The same Typescript issue happens here when using a later Typescript

packages/filesystem/src/browser-only/opfs-filesystem-provider.ts
sdirix: It's fine to break backward compatibility here, we should just add a notice in the changelog describing the breaking change

packages/filesystem/src/browser-only/download/file-download-service.ts
```diff
@@ -0,0 +1,361 @@
...
+];
+
+@injectable()
+export class FileDownloadServiceImpl implements FileDownloadService {
+
+    @inject(FileService)
+    protected readonly fileService: FileService;
+
+    @inject(ILogger)
+    protected readonly logger: ILogger;
```
sdirix, line 36: We should use a named logger

packages/filesystem/src/common/upload/file-upload-constants.ts
```diff
@@ -0,0 +1,87 @@
...
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+/**
+ * System file patterns that should be filtered out during file uploads.
+ * These patterns cover common system files, temporary files, and development artifacts
+ * that are typically not needed in the workspace.
+ */
+export const SYSTEM_FILE_PATTERNS = [
```
sdirix: These are very opinionated defaults. I would much prefer allowing the upload of everything the user selected.

packages/plugin-ext/src/main/browser/dialogs-main.ts
```diff
@@ -20,7 +20,7 @@ import { OpenDialogOptionsMain, SaveDialogOptionsMain, DialogsMain, UploadDialog
 import { OpenFileDialogProps, SaveFileDialogProps, FileDialogService } from '@theia/filesystem/lib/browser';
 import { WorkspaceService } from '@theia/workspace/lib/browser';
 import URI from '@theia/core/lib/common/uri';
-import { FileUploadService } from '@theia/filesystem/lib/browser/file-upload-service';
+import { FileUploadService } from '@theia/filesystem/lib/common/upload/file-upload';
```
sdirix, line 23: Move of files should at least be generically mentioned in the changelog breaking changes

packages/filesystem/package.json
```diff
@@ -4,17 +4,20 @@
...
+    "http-status-codes": "^2.3.0",
+    "minimatch": "^10.0.3",
     "multer": "^2.0.1",
+    "opfs-worker": "^1.3.1",
```
sdirix:
```suggestion
    "opfs-worker": "1.3.1",
```

sdirix (changes requested):
Thanks for this great contribution!

The code is clear and understandable. I tested file creation, deletion, copying, moving, downloading and uploading and found only one minor issue:

The `tar` download does not show up in the Browser download history while the single file download does. In the regular Theia Browser variant both show up in the history.

---

This PR introduces a new dependency: `opfs-worker`. As part of the dependency review, I noticed that the library is both very new and currently authored and maintained by you. In general, and especially in light of recent supply chain incidents on npm, the Theia framework needs to be very cautious about which dependencies are introduced.

Even though `opfs-worker` is only required by the browser-only variant, it would still be downloaded by all Theia adopters as part of their builds. This means that, hypothetically, a compromised update such as one containing a `postInstall` script could affect all Theia adopters. To be clear, we do not assume you would ever intend such an attack, but the attack vector exists and we need to account for it.

We discussed this in the Theia developers call and agreed on the following process: We will pin the dependency to a vetted version when introducing it. Any update to `opfs-worker` will take place in a separate PR where we manually review and validate the new version. We will reevaluate this process after one year.

This approach is not based on mistrust toward you personally but is a general policy we will apply consistently for any new niche dependencies. Our goal is to protect the integrity of Theia while still enabling valuable contributions such as yours.

For this PR, I suggest pinning opfs-worker to version 1.3.1

We hope this approach works for you, and we truly appreciate your effort in contributing to Theia :heart:

kachurun:
Thank you for the review ♥️. I will address the points you raised as soon as possible

packages/filesystem/src/common/upload/file-upload-constants.ts
kachurun: Sure, I'll just remove it

packages/filesystem/src/browser-only/download/file-download-service.ts
kachurun: Done

packages/filesystem/src/browser-only/download/file-download-service.ts
kachurun: Oh yeah, SharedArrayBuffer and ArrayBuffer were split in 5.9
I think wrapping it in a Uint8Array will fix that

kachurun: No, it's not a good idea from an optimization perspective. It's better to just cast it as BlobPart to keep TypeScript happy.

packages/filesystem/package.json
kachurun: I can't remember why I did this, I'll just revert that

kachurun:
> This PR introduces a new dependency: opfs-worker. As part of the dependency review, I noticed that the library is both very new and currently authored and maintained by you. In general, and especially in light of recent supply chain incidents on npm, the Theia framework needs to be very cautious about which dependencies are introduced.

> For this PR, I suggest pinning opfs-worker to version 1.3.1

Oh yeah, that's totally understandable and the right thing to do. I created this lib specifically for Theia and my personal project, which uses it alongside the browser-only Theia, but it hasn't even been released into production yet, so yes, the lib itself is still young.

So I totally get your concerns, that's the right thing after all these attacks on npm packages.

kachurun:
> The tar download does not show up in the Browser download history while the single file download does. In the regular Theia Browser variant both show up in the history.

Yes, that's bad, however, I did it on purpose. In the normal version, you create a temp array on the server and then just download it via a link. But I can't do that, so I need to create the archive in memory, and depending on the filesystem size, it could crash the browser due to memory. So instead of creating a blob in memory, I open a file picker where the user chooses where to save the file. After that, I stream into this file, so it's memory efficient. This only works in Chromium, and files downloaded this way don't appear in the downloads and we cant control this behavior.
I'm using this technique for all archives, and it actually causes a bug with high memory usage when trying to download some enormous file (not folder).
I could do it slightly differently. I could define a maximum file size limit after which I will do it with filepicker, before that limit I just create it in memory and download as block so it appear in downloads. Let's say 50MB should be reasonable amount. So for small codebases, it will work as in server Theia, but if you try to download a really big file(s), it will show the picker first.

kachurun:
So I also found a critical issue. It seems to me it's not related to browser-only, but I could be wrong:

When I try to open a large file (or just a non-text file), it immediately calls the fs.open handler and prompts with a message like "Are you sure you want to open this?" If I press No, it fails somewhere inside Monaco because the expected file is unavailable. It also doesn't call fs.close after that, so the file remains busy and I can't do anything with it—like remove, rename, or even download it.

[Fixed here](https://github.com/eclipse-theia/theia/pull/16187/commits/056e9c1686258b65448d23f3835c675b12d8739f)

kachurun:
@sdirix 
All changes have been made. Please check / test / review it again, thank you so much!

CHANGELOG.md
```diff
@@ -9,13 +9,18 @@
...
+- [filesystem] `FileUploadService` moved from `@theia/filesystem/lib/browser/file-upload-service` to `@theia/filesystem/lib/common/upload/file-upload` and refactored to use symbol `FileUploadService` with separate `FileUploadServiceImpl`. Update import statements and service bindings accordingly. [#16187](https://github.com/eclipse-theia/theia/pull/16187)
+- [filesystem] `FileDownloadService` refactored: `file-download-data.ts` renamed to `file-download.ts` and `FileDownloadService` is now defined as a symbol with separate `FileDownloadServiceImpl`. `node-file-upload-service.ts` moved from `src/node/` to `src/node/upload/`. Update import statements and service bindings accordingly. [#16187](https://github.com/eclipse-theia/theia/pull/16187)
+- [filesystem] `OPFSInitialization.getRootDirectory()` now returns `Promise<string> | string` instead of `Promise<FileSystemDirectoryHandle>`. The `FileSystemDirectoryHandle` type is no longer supported. Update code that calls this method to expect a string path. [#16187](https://github.com/eclipse-theia/theia/pull/16187)
```
sdirix: Thanks for adding information here. However we only add one entry per PR with very dense content. Please align with other breaking changes of the past. If needed add more details in the migration document.

package-lock.json
```diff
@@ -31365,6 +31408,21 @@
       "integrity": "sha512-i+bkd9UYFis40RcnkW7XrOprCujXRAHg62IVh/Ah3G8MmNXpCGt1m0dTFhSdx/AVs8XEMbdOGRwdkR1Bcta8AA==",
       "license": "BSD-3-Clause"
     },
+    "packages/ai-ide/node_modules/minimatch": {
```
sdirix: We should avoid duplicating dependencies into sub-packages if not absolutely necessary

package-lock.json
```diff
@@ -31889,6 +31957,148 @@
...
+    "packages/filesystem/node_modules/media-typer": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/media-typer/-/media-typer-1.1.0.tgz",
+      "integrity": "sha512-aisnrDP4GNe06UcKFnV5bfMNPBUw4jsLGaWwWfnH3v02GnBuXX2MCVn5RbrWo0j3pczUilYblq7fQ7Nw2t5XKw==",
+      "license": "MIT",
+      "engines": {
+        "node": ">= 0.8"
+      }
+    },
+    "packages/filesystem/node_modules/mime-db": {
```
sdirix: All of these are not hoisted

package-lock.json
sdirix: In general I would expect almost no changes to the package-lock by this PR as we only modify `minimatch`, and add `opfs-worker` & `tar-stream` (and their dependencies)

sdirix (changes requested):
Thanks for the changes, I like the new download behavior!

Please see the comments and make sure that the linter completes successfully. Thanks!

kachurun:
> make sure that the linter completes successfully

I always forget about it. Actually, I have a question: normally ESLint just runs in my IDE and fixes files on save, but I can't get it to work in the Theia repo. Does it work for you, or do you always run the linter only from the npm script?

Another question: would it be worth setting up [lint-staged](https://github.com/lint-staged/lint-staged) to auto-fix errors on commit and prevent commits with lint/tsc errors?

Like now it is yielding a couple of useless spaces in one file, and it could be fixed in the commit.

package-lock.json
kachurun: Yes, it's weird. Okay, I will revert it to the master version, clean up node_modules, and install again. I hope that helps.

kachurun: Ok, it helps, but minimatch still isn't hoisted because npm thinks some random old version from another npm module is better. So I just added it as a devDependency in the root. Now it's deduped properly, but packages that require older minimatch versions (like mocha or lerna) will have their own copies.

tar-stream also can't be hoisted. It's used by a bunch of modules, and the hoisted version is 2.2.0. However, filesystem uses tar-fs and tar-stream, both depending on tar-stream@3. Since it's not required by any other Theia package, I think it's fine. Otherwise, we could move it as a devDependency to the root package.json (but I don't think we should).

kachurun:
@sdirix, it's probably all set. Please take a moment to check it again. Thank you so much!

CHANGELOG.md
```diff
@@ -9,13 +9,15 @@
 - [ai-chat-ui] add chat input history with arrow key navigation [#16174](https://github.com/eclipse-theia/theia/pull/16174) - contributed on behalf of STMicroelectronics
 - [dependencies] increased minimum node version to 20.x [#16218](https://github.com/eclipse-theia/theia/pull/16218) - Contributed on behalf of STMicroelectronics
 - [workspace] fix: cannot open same folder once declined exit confirmation [#16238](https://github.com/eclipse-theia/theia/pull/16238)
+- [filesystem] browser-only: fixed Safari compatibility issues with OPFS API, implemented file watching, folder copying, file/folder download and upload capabilities with streaming support [#16187](https://github.com/eclipse-theia/theia/pull/16187)
```
sdirix:
```suggestion
```

Issue #16193, opened 2025-08-19
Theia does not allow VSCode extensions to provide resources outside extension directory


ngfelixl:
### Feature Description:

Hi,

Theia does not allow that VSCode extensions retrieve file resources outside of the extension directory. Imagine there is a tree, and that tree creates it's icons dynamically in a tmp directory (which is not the extension directory or child of it), than Theia would not allow to display those icons.

There is actually a simple fix for this problem. Here is the endpoint for the http request.
https://github.com/eclipse-theia/theia/blob/1539178c0eb33545f42b72ab1c8df68b9b157b48/packages/plugin-ext/src/hosted/node/plugin-reader.ts#L52

You see, that it uses `express`s res.sendFile with the `root` option. This option however prevents to send files outside of it, so if the relative path is `../../someotherdir` it won't work.

If we change this to the following, it works fine.

```ts
const absolutePath = path.resolve(localPath, filePath);
res.sendFile(absolutePath, e => {/*...*/})
```

If you wish, I can create a PR. Do you see any security issues here?

msujew:
Hey @ngfelixl,

does this work as expected in VS Code? I'm all for it if it works like that in VS Code, but if it doesn't I assume it's due to some security concerns.

ngfelixl:
In VSCode it works also with folders outside of the extension directory.

msujew:
I see, then a PR would be appreciated, thank you :)

tsmaeder:
This change sounds scary, but you could already copy any file on the back end system inside the plugin directory and access it from there.

ngfelixl:
In our case it's not possible, because Theia is installed to _C:/Program Files/..._, which requires Admin privileges. The built-in extensions are then also in _C:/Program Files_ and therefore Theia can not modify the extensions directory without Admin privileges.

And I think somewhere in the VSCode docs it says the extensions folder should not be touched or used for tmp files. I can't find it right now, but I will add a link once I do.

I can modify our Theia app, that's not the problem, but I thought it might be in your interest, too.

ngfelixl:
The related PR is merged, probably this issue can be closed, too.

PR #16195, opened 2025-08-19
Allow VSCode extension to access files outside of the extension directory

ngfelixl:
#### What it does

It allows VSCode extensions to access files outside of the extension directory.

Partially addresses #16193

#### How to test

Create a vscode extension with a treeview. Use the `MyTreeDataProvider` (I tested it), and then you can verify the paths running through the modified code manually.

```ts
import { TreeDataProvider, TreeItem, window, Uri } from 'vscode';

export function activate() {
  const treeDataProvider = new MyTreeDataProvider();
  const treeView = window.createTreeView<TreeItem>('tree-view-id', {treeDataProvider});
}

export class MyTreeDataProvider implements TreeDataProvider<TreeItem> {
  getTreeItem(item: TreeItem): TreeItem {
    return item;
  }

  getChildren(item?: TreeItem): TreeItem[] {
    if (item) {
      return [];
    }

    // const iconUri = Uri.joinPath(context.extensionUri, 'pathtoicon');                     // <-- works already
    const iconUri = Uri.file(/* some string to a local file, e.g. c:/.../iconname.png */);   // <-- did not work (outside of extension directory, but works with this PR
    const treeItem = new TreeItem('Test Node');
    treeItem.iconPath = {
      light: iconUri,
      dark: iconUri,
    };
    return [treeItem];
  }
}

```

And register the view `tree-view-id` in the `package.json`. Use an absolute icon path to a file in the file system which is not inside the extension directory.

#### Follow-ups

\-

#### Breaking changes

It __should__ not introduce a breaking change. At least I do not notice it. Please double check.

#### Attribution

\-

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/hosted/node/plugin-reader.ts
+++ b/packages/plugin-ext/src/hosted/node/plugin-reader.ts
@@ -49,7 +49,8 @@ export class HostedPluginReader implements BackendApplicationContribution {
 
             const localPath = this.pluginsIdsFiles.get(pluginId);
             if (localPath) {
-                res.sendFile(filePath, { root: localPath }, e => {
+                const absolutePath = path.resolve(localPath, filePath);
+                res.sendFile(absolutePath, e => {
                     if (!e) {
                         // the file was found and successfully transferred
                         return;


colin-grant-work (changes requested):
As far as I can tell, these changes are not working for me (on MacOS). I believe that in addition to this code, there needs to be a modification to the handlers for icons in TreeItems. I would take a look at this code, as that seems to be where my nice absolute paths are being mangled into something that can't subsequently be retrieved.

https://github.com/eclipse-theia/theia/blob/50b1c4dca9c29049a74fddcc4ef8c3845114d1b1/packages/plugin-ext/src/plugin/tree/tree-views.ts#L462-L468

ngfelixl:
Unfortunately, I don't have any possibility to run this on Mac. How do we want to proceed here? Should I abandon this PR?

colin-grant-work:
> Unfortunately, I don't have any possibility to run this on Mac. How do we want to proceed here? Should I abandon this PR?

It would be great if we could work through more of the places where we mangle those paths. To be honest, I'm surprised it works with just this change on any platform: at first glance, it doesn't look like there's a lot platform-dependent code in this chain. But we can accept the PR as necessary and not regressing, if not sufficient to deliver the full functionality described everywhere.

Issue #3424, opened 2018-11-07
Run to line


tsmaeder:
Implement an action that runs to a line in the currently selected frame. Will fill in the details in this issue, soon.

colin-grant-work:
@tsmaeder, are you still interested in this feature / is it still missing?

pisv:
@colin-grant-work Can you please close this issue? I think it has been addressed in #16219, but was not closed automatically.

colin-grant-work:
Closed by https://github.com/eclipse-theia/theia/pull/16219

colin-grant-work:
@pisv, yes, unfortunately, you have to use one of the magic words before every issue `closes #..., closes #...`; GitHub doesn't understand `closes #..., #...`.

pisv:
@colin-grant-work Oh, good to know! Thank you 👍

PR #16219, opened 2025-08-26
Add `Run to Cursor` and `Run to Line` commands

### LINKED ISSUE #14698: Support VS Code's `Run to Cursor` command

pisv:
#### What it does

Closes #14698, #3424.

This PR adds the `Run to Cursor` and `Run to Line` commands, which continue execution of the current thread to the current cursor position and to the current line, respectively. Both commands can be accessed through the command palette and the editor's context menu. The `Run to Line` command can also be accessed through the editor's line number context menu. The commands are visible and enabled while the current thread is paused and there is the current text editor.

#### How to test

Use the [attached project](https://github.com/user-attachments/files/21990353/hello.zip), which contains files `hello.js` and `hi.js`, and a configuration for launching `hello.js` as a Node.JS program.

1. Set a breakpoint on line 1 in `hello.js` and launch the program.

2. After the breakpoint is hit, open the editor's line number context menu on line 5 and choose `Run to Line`.

   The program will sleep in a loop for 5s before stopping at line 5. This 5s delay can be used to verify that the `Run to Line` command did set a temporary breakpoint on line 5. This delay can also be used to verify that the temporary breakpoint is removed when the current session is paused or stopped (e.g. explicitly by the user) even before the temporary breakpoint is hit. Also, this delay can be used to verify that neither `Run to Line` nor `Run to Cursor` commands are available while the current thread is running.

3. After the program stops at line 5, verify that the temporary breakpoint has been removed. Open the editor's context menu at the very end of line 5 and choose `Run to Cursor`. Verify that a dialog is opened informing the user that the target position will need to be shifted to Ln 8, Col 5 and asking the user to confirm running to the shifted position. Click the `No` button in the dialog and verify that the temporary breakpoint set at 8:5 has been removed.

4. Open the editor's context menu at line 5, column 9 and choose `Run to Cursor`. The current thread should run to the current cursor position and the temporary breakpoint set at 5:9 should be removed once it is hit.

5. Open the `hi.js` file, which is not part of the launched program. Open the editor's context menu on line 1 and choose `Run to Line`. Verify that a temporary breakpoint has been set on line 1, but it is disabled (i.e. not verified). After about 2s, the temporary breakpoint should be removed and a warning message should appear informing the user that the command could not be executed successfully, because stopping the current thread would not be possible at the given location.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-frontend-application-contribution.ts
+++ b/packages/debug/src/browser/debug-frontend-application-contribution.ts
@@ -1324,7 +1324,7 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
             if (thread.stopped && thread === this.manager.currentThread) {
                 return true;
             }
-            console.warn('Cannot run to the given location. The current thread has changed or is not stopped.');
+            console.warn('Cannot run to the specified location. The current thread has changed or is not stopped.');
             return false;
         };
         if (!checkThread()) {
@@ -1346,7 +1346,7 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
             }
             if (!sessionBreakpoint || !sessionBreakpoint.installed || !sessionBreakpoint.verified) {
                 this.messageService.warn(nls.localize('theia/debug/cannotRunToThisLocation',
-                    'Cannot run the current thread to the given location. Stopping the current thread would not be possible at the given location.'
+                    'Could not run the current thread to the specified location.'
                 ));
                 return;
             }
@@ -1354,7 +1354,7 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
             if (rawBreakpoint.line !== line || (column && rawBreakpoint.column !== column)) {
                 const shouldRun = await new ConfirmDialog({
                     title: nls.localize('theia/debug/confirmRunToShiftedPosition_title',
-                        'Cannot run the current thread to exactly the given location'),
+                        'Cannot run the current thread to exactly the specified location'),
                     msg: nls.localize('theia/debug/confirmRunToShiftedPosition_msg',
                         'The target position will be shifted to Ln {0}, Col {1}. Run anyway?', rawBreakpoint.line, rawBreakpoint.column || 1),
                     ok: Dialog.YES,


pisv:
@ndoschek Can you please take a look at this one as well? :)

packages/debug/src/browser/debug-frontend-application-contribution.ts
```diff
@@ -1258,6 +1315,87 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
...
+            const rawBreakpoint = sessionBreakpoint.raw!; // an installed breakpoint always has the underlying raw breakpoint
+            if (rawBreakpoint.line !== line || (column && rawBreakpoint.column !== column)) {
+                const shouldRun = await new ConfirmDialog({
+                    title: nls.localize('theia/debug/confirmRunToShiftedPosition_title',
+                        'Cannot run the current thread to exactly the given location'),
+                    msg: nls.localize('theia/debug/confirmRunToShiftedPosition_msg',
+                        'The target position will be shifted to Ln {0}, Col {1}. Run anyway?', rawBreakpoint.line, rawBreakpoint.column || 1),
+                    ok: Dialog.YES,
+                    cancel: Dialog.NO
+                }).open();
```
colin-grant-work: Since breakpoint shifting is usually done silently, I would probably omit this dialog, as VSCode does.

packages/debug/src/browser/debug-frontend-application-contribution.ts
```diff
@@ -1258,6 +1315,87 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
...
+        };
+        try {
+            const sessionBreakpoint = await this.verifyBreakpoint(breakpoint, thread.session);
+            if (!checkThread()) {
+                return;
+            }
+            if (!sessionBreakpoint || !sessionBreakpoint.installed || !sessionBreakpoint.verified) {
+                this.messageService.warn(nls.localize('theia/debug/cannotRunToThisLocation',
+                    'Cannot run the current thread to the given location. Stopping the current thread would not be possible at the given location.'
+                ));
```
colin-grant-work: I do like the additional information provided by toasts in this implementation. For this one, I might say something a little more explicit about the technical reason like 'session did not accept/verify breakpoint.'

packages/debug/src/browser/debug-frontend-application-contribution.ts
pisv: If you don't mind, I'd like to keep this confirmation dialog. I think that in most cases, there would be no breakpoint shifting when using this feature, so the dialog should not appear often, but when breakpoint shifting does occur, it can often be quite surprising to the user, in which case this dialog might actually be valuable.

When the user sets a breakpoint manually, there is a strong visual indication for the breakpoint. So, when the breakpoint is shifted by the debugger, it is usually readily apparent to the user. In general, the user can move the breakpoint to another position if necessary before it is hit.

In contrast, when the user invokes the `Run to Cursor` command, the fact that a temporary breakpoint is set automatically is just an implementation detail, which the user might be completely unaware of. The execution of the current thread resumes without delay, so in general, there is no chance to the user to even see the temporary breakpoint, much less to move it to another position, since oftetimes it is hit and removed immediately.

That's why I think that it actually makes sense to show the confirmation dialog in this case, to avoid unpleasant surprises to the user about the actual target position of the `Run to Cursor` command.

packages/debug/src/browser/debug-frontend-application-contribution.ts
pisv: Actually, my first iteration on this message sounded like 'Cannot run the current thread to the given location. The breakpoint at the given location could not be verified by the current session.' :)

I reconsidered it later on, because it started to sound too technical to me. Also, setting a temporary breakpoint is just an implementation detail of this feature, which the user might even be unaware of, so I thought that it would probably be better to express the message in more general terms...

pisv:
@colin-grant-work Thanks a lot for the prompt and thorough review, and your approval! 👍

I really appreciate your comments, please see my replies inline.

packages/debug/src/browser/debug-frontend-application-contribution.ts
colin-grant-work: Hm... I see what you mean, but to me the trouble is that the current message makes a claim that we can't be sure is correct: we don't know why the breakpoint wasn't installed and verified in the time allotted; we only know that it wasn't.

packages/debug/src/browser/debug-frontend-application-contribution.ts
colin-grant-work: To me, this introduces an inconsistency. In general, it's a feature of the Debug Adapter Protocol that the adapter is free to interpret the user's suggested breakpoint position however it makes sense to the adapter. For example, in your testing instructions, you say to put a breakpoint on line 1 of `hello.js`, but that breakpoint is automatically and silently moved to line 2 (the first line with executable code) when the breakpoint is handled by the Node debug adapter. Similarly, if I put a real breakpoint at the end of line 5 using the 'insert inline breakpoint' command, that breakpoint is automatically shifted to line 8. The principle is that the user proposes, but the adapter disposes, and I don't see good reason to call it out specially in this case.

But since I agree that that behavior can be a bit jarring, I won't insist that we remove the dialog, but I do think that it will cause more friction that it saves.

pisv: > I won't insist that we remove the dialog

I'd really appreciate it! Thank you.

FWIW, the dialog can easily be removed if/when needed, e.g. based on the  actual user feedback...

> In general, it's a feature of the Debug Adapter Protocol that the adapter is free to interpret the user's suggested breakpoint position however it makes sense to the adapter. For example, in your testing instructions, you say to put a breakpoint on line 1 of hello.js, but that breakpoint is automatically and silently moved to line 2 (the first line with executable code) when the breakpoint is handled by the Node debug adapter. Similarly, if I put a real breakpoint at the end of line 5 using the 'insert inline breakpoint' command, that breakpoint is automatically shifted to line 8.

Yes, but there is a significant difference in UX between setting a real breakpoint and then resuming the current thread's execution manually, and what happens when the `Run to Cursor` command is invoked.

When the current thread is paused (which is a precondition for the `Run to Cursor` command), and the user sets a breakpoint that is then shifted by the debugger, the effect is immediately visible to the user, and the user can take action such as moving the breakpoint to another position before the execution is resumed.

In contrast, when the `Run to Cursor` command is invoked, the user will not in general even see the temporary breakpoint set by the command, since the current thread's execution is resumed right away and in most cases the breakpoint will be immediately hit and removed. If the temporary breakpoint is shifted by the debugger, there is no reliable way to make the user informed about that except for showing the confirmation dialog.

In practice, it might look really puzzling and even infuriating, when instead of hitting the current cursor position, the program execution silently skips past it and hits a seemingly unrelated position.

Let's consider a simple example:

```js
var a = 0;

for (let i = 0; i < 1000; i++) {
    a += i;
}

console.log(a);
```

If the user places the cursor at the end of the line after `a += i;` and invokes `Run to Cursor`, where would the program execution be paused?

packages/debug/src/browser/debug-frontend-application-contribution.ts
pisv: > we don't know why the breakpoint wasn't installed and verified in the time allotted; we only know that it wasn't.

Yes, and therefore we just say that 'Stopping the current thread would not be possible at the given location'. As mentioned above, the user might not even be aware of the temporary breakpoint, which is just an implementation detail that in most cases remains hidden to the user.

However, if that still doesn't sound convincing to you, I'd be happy to implement any specific wording you'd like. My strong opinion about it is weakly held :)

colin-grant-work: Sorry to be so picky about it. Let's go with:

```
'Could not run the current thread to the specified location.'
```

or

```
'Failed to run the current thread to the specified location.'
```

That removes the `would not be possible` language that to me sounds like we're making a general statement, rather than just reporting that we in fact failed to do it on this occasion for unknown reasons.

pisv: No worries. I've implemented the first variant that you have proposed. Perhaps a part of the problem is that English is not my native language, as you might have guessed :) Thanks again for your review.

PR #16221, opened 2025-08-27
16098: cancel tool execution confirm dialogs when request is cancelled

CamilleLetavernier:
#### What it does

- Cancel pending tool usage confirmation requests when chat request is canceled

#### How to test

- Enable tool usage confirmation in the AI Configuration view ("Tools" tab)
- Ask Coder to make changes to your workspace (e.g. `@Coder Suggest changes in #_f`)
- Once the "Confirm Tool Execution" dialog shows up, cancel the chat request
- Expected result: the "Confirm Tool Execution" dialog is closed, and shows: `Execution canceled: getFileContent — Chat request canceled`

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-response-renderer/tool-confirmation.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-response-renderer/tool-confirmation.tsx
@@ -22,7 +22,7 @@ import { ToolCallChatResponseContent } from '@theia/ai-chat/lib/common';
 /**
  * States the tool confirmation component can be in
  */
-export type ToolConfirmationState = 'waiting' | 'allowed' | 'denied';
+export type ToolConfirmationState = 'waiting' | 'allowed' | 'denied' | 'rejected';
 
 export interface ToolConfirmationProps {
     response: ToolCallChatResponseContent;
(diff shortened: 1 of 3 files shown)


planger:
Would it be possible to also show this icon if the user cancels the chat request with `Always Allow` tool confirmation mode too? I think this is an open issue as well and seems very related.

Steps:
* Set the Tool Confirmation Mode to `Always Allow`
* Start a chat request having the tool confirmation disabled
* Wait until a tool call is running (e.g. `@Coder write a short poem into abc.txt`)
* Cancel the chat
* Expected: the tool call shows the corresponding icon and not the spinner

Nit: Imho `cancelled` might be a better name for the `ToolConfirmationState` instead of `rejected` as `rejected` feels semantically too close to `denied`.

CamilleLetavernier:
> Would it be possible to also show this icon if the user cancels the chat request with Always Allow tool confirmation mode too? I think this is an open issue as well and seems very related.

:+1: I've pushed another commit for this use case

> Nit: Imho cancelled might be a better name for the ToolConfirmationState instead of rejected as rejected feels semantically too close to denied.

Yeah I wasn't too sure about existing use cases. The way it was implemented until now was that "rejected" was for error handling; cancel being only one special type of error. 

I don't think there is any path that can lead to actual runtime errors (so the only way to reject the promise is to cancel the tool confirmation); but I didn't want to change the existing semantics too much. So I'm still treating "canceled" as just one special case of error.

Still, explicitly mapping "rejected" state to "cancelled" feels a bit weird; so maybe we need a more explicit state?

Issue #16215, opened 2025-08-20
[ai-chat] Google AI: "Mixing content and parts" error after tool call


chriscross2:
### Bug Description:
For Google's Gemini 2.5 Pro and Flash models used in Coder or Architect, the API request error "Mixing Content and Parts is not supported, please group the parts into a the appropriate Content objects and specify the roles for them" is (sometimes) returned after a tool call, such as Ran getFileContent. The chat is stopped.
Never happened with Mistral, Anthropic, or OpenAI models.

### Steps to Reproduce:

1. Right click in any open file in the editor and "Ask AI"
2. e.g. "(at)Coder Format the second part of the file like the first part"

### Additional Information
The error has not happened with the system-architect-next variant, so far.
Maybe related (also Gemini): @Architect "Summarize Task for Coder" returns (sometimes) a markdown file error and file has just a few lines.

- Operating System: Windows 11
- Theia Version: 1.64.100

sgraband:
@ndoschek I also noted this and it basically prevents users from using the google models with coder. Could you plan this in? (I think this issue should be transferred to the Theia repo, but that would need to be checked).

CamilleLetavernier:
I couldn't reproduce this exact error, but got something similar:

<img width="704" height="180" alt="Image" src="https://github.com/user-attachments/assets/83ca3f48-ae73-4f19-81d7-a5518774fbd3" />

After investigation, it turns out we don't properly catch "Error tokens" from Gemini, and the displayed error message is then slightly misleading. After fixing the error message, I got some more helpful information:

<img width="585" height="256" alt="Image" src="https://github.com/user-attachments/assets/ddc8a48d-b960-4e98-9916-65b1bcea7101" />

Based on the Gemini API Docs, it seems that this might be an Internal Server error:

> MALFORMED_FUNCTION_CALL	The function call generated by the model is invalid.

This might be related:

https://discuss.ai.google.dev/t/malformed-function-call-finish-reason-happens-too-frequently-with-vertex-ai/93630/4

If that's the case, it seems that it's nothing we can fix on our side. In any case, this seems to match my own experience: the exact same prompt might work 4 times in a row, then fail on the 5th try.

PR #16228, opened 2025-08-28
16215: improve error reporting for Google GenAI connector

CamilleLetavernier:
- handle error-case for tokens with a 'finishReason'
  - This PR **does** improve Error reporting
  - This PR **does not** fix the actual error, which seems to be related to the Gemini model itself, rather than the Theia integration.

refs #16215 

#### What it does

- Improve error reporting when Gemini LLMs reply with an Error token (Token with "finishReason" other than "STOP").

https://ai.google.dev/api/generate-content#FinishReason

#### How to test

- Configure Coder with google/gemini-2.5-flash
  - Also reproduced with all gemini-2.5-* models
  - Not tested with older Gemini models (<= 2.0)
- Prompt to make changes to an arbitrary file (e.g. `@Coder Suggest improvements in #_f `)
- Retry multiple times: in 10 to 20% of cases, the model will fail after a tool call. The error message should reflect the finishReason:

**BEFORE**:

image

**AFTER**:

image

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30525,7 +30525,7 @@
       "version": "1.64.0",
       "license": "EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0",
       "dependencies": {
-        "@google/genai": "^1.3.0",
+        "@google/genai": "^1.16.0",
         "@theia/ai-core": "1.64.0",
         "@theia/core": "1.64.0"
       },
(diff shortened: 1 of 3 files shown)


sdirix (approved):
Works for me

PR #16229, opened 2025-08-28
16122: deduplicate tools before sending LLM request

### LINKED ISSUE #16122: Anthropic error: Tool names must be unique (when used in chat with some ~tools)
### Bug Description:
Hi,
When using an agent with Anthropic model (like Coder by default), with some of the built-in tools using the prefix ~ (getFileDiagnostics, getFileContent and others), I am getting this error: 

```
2025-08-04T05:39:48.475Z root ERROR Error in Anthropic streaming: Error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"tools: Tool names must be unique."}}
    at APIError.generate (/theia/examples/browser/lib/backend/vendors-node_modules_drivelist_build_Release_drivelist_node-dev-packages_application-package_-e8e608.js:9615:20)
    at Anthropic.makeStatusError (/theia/examples/browser/lib/backend/vendors-node_modules_drivelist_build_Release_drivelist_node-dev-packages_application-package_-e8e608.js:9090:32)
    at Anthropic.makeRequest (/theia/examples/browser/lib/backend/vendors-node_modules_drivelist_build_Release_drivelist_node-dev-packages_application-package_-e8e608.js:9241:30)
```

Note: It works with OpenAI models, but I didn't test it with Google models and other LLMs.

### Steps to Reproduce:

1. Choose an agent with Anthropic model (like Coder by default)
2. Put this prompt, for example, in the chat: @Coder ~ clearFileChanges
3. You will get the above error!

### Additional Information

- Operating System: Mac
- Theia Version: Theia 1.64/1.63

CamilleLetavernier:
- some LLMs (e.g. Anthropic models) are sensitive to duplicate tools
- always deduplicate tools before sending a request

Fixes #16122 

#### What it does

- Always deduplicate tools before sending a request to an LLM
- Most LLMs properly handle duplicate tools; but not e.g. Anthropic models
- With this change, all models are properly supported

#### How to test

- Set the coder model to Anthropic (e.g. Sonnet 4.0)
- Using the default coder prompt, specify a prompt with a tool already defined in system prompt, e.g: `@Coder using ~getFileContent, suggest improvements in #_f`
- The LLM should not throw an error

**Before:**

image

**After:**

image

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-agents.ts
+++ b/packages/ai-chat/src/common/chat-agents.ts
@@ -314,6 +314,22 @@ export abstract class AbstractChatAgent implements ChatAgent {
         return requestMessages;
     }
 
+    /**
+     * Deduplicate tools by name (falling back to id) while preserving the first occurrence and order.
+     */
+    protected deduplicateTools(toolRequests: ChatToolRequest[]): ChatToolRequest[] {
+        const seen = new Set<string>();
+        const deduped: ChatToolRequest[] = [];
+        for (const tool of toolRequests) {
+            const key = tool.name ?? tool.id;
+            if (!seen.has(key)) {
+                seen.add(key);
+                deduped.push(tool);
+            }
+        }
+        return deduped;
+    }
+
     protected async sendLlmRequest(
         request: MutableChatRequestModel,
         messages: LanguageModelMessage[],
@@ -322,7 +338,8 @@ export abstract class AbstractChatAgent implements ChatAgent {
     ): Promise<LanguageModelResponse> {
         const agentSettings = this.getLlmSettings();
         const settings = { ...agentSettings, ...request.session.settings };
-        const tools = toolRequests.length > 0 ? toolRequests : undefined;
+        const dedupedTools = this.deduplicateTools(toolRequests);
+        const tools = dedupedTools.length > 0 ? dedupedTools : undefined;
         return this.languageModelService.sendRequest(
             languageModel,
             {


packages/ai-chat/src/common/chat-agents.ts
```diff
@@ -314,6 +314,22 @@ export abstract class AbstractChatAgent implements ChatAgent {
...
     }
 
+    /**
+     * Deduplicate tools by name (falling back to id) while preserving the first occurrence and order.
+     */
+    protected deduplicateTools(toolRequests: ChatToolRequest[]): ChatToolRequest[] {
+        const seen = new Set<string>();
+        const deduped: ChatToolRequest[] = [];
+        for (const tool of toolRequests) {
+            const key = tool.name ?? tool.id;
```
colin-grant-work, line 324: Why prioritize `name` over `id` here? Is one or the other used as the authoritative identifier elsewhere? Would we be happy if the same tool occurred with different combinations of the two, i.e. what would we want to do with the following list:

```javascript
[{ name: 'Foo', id: 'foo' },
{ name: 'Foo', id: 'bar' },
{ name: 'Bar', id: 'foo' },
{name: 'Bar', id: 'bar' }}
```

The current code would give us 

```javascript
[{ name: 'Foo', id: 'foo' },
{ name: 'Bar', id: 'foo' }]
```

CamilleLetavernier: As far as I can tell, using an id that is different from the name would likely fail in multiple (maybe hard to detect) ways; not just with this PR.

In short: all LLMs rely on Name; all internal Theia code relies on ID.
Theia de-duplication in the ToolRegistry is based on ID, but LLMs don't know/use the ID; we only pass the name.

All Theia Tool Providers are implemented with name === id.

IMO  we should remove the ID field entirely and only use the name, as that's the standard LLM way. We don't need two fields. But that would be a different issue entirely.

colin-grant-work (approved):

Issue #16223, opened 2025-07-15
Commands contributed by vscode extensions may not work in Theia Toolbar


hansd56:
### Bug Description:

When creating a new command in the toolbar it gets now saved and the icon is displayed correctly.. Unfortunately, when clicking on the newly created icon the command does not execute. Only the default commands work.

Edit: This is happening for commands contributed by various VS Code extensions, e.g. GitLens, cSpell just to name a few.

### Steps to Reproduce:

1. Create a new command contributed by a VS Code extension with icon in the toolbar.
2. Click on the icon and nothing happens.

### Additional Information

- Operating System: MX Linux 23.6
- Theia Version: 1.63.200

JonasHelming:
I just retested this in the Theia IDE latest version using the "about" command and it worked for me. Could you specify more specific steps to reproduce, please? Also: Are you using the Theia IDE?

hansd56:
Hi, thanks for looking at it again. I use Theia IDE 1.63.200

I tried a few options and indeed some options from the command palette work and some do not.

For example I use the Theia IDE for embedded development. I often need a serial monitor. I have installed the extension https://open-vsx.org/extension/eclipse-cdt/serial-monitor. To open the serial monitor dialog I have to select Open Serial from the command palette. I thought I could make live easier by adding it to the toolbar. It saves everything correctly but when I click on the icon it does nothing.

Maybe you could install the extension and see if it works on your machine. I am using MX Linux a Debian clone.

JonasHelming:
Can you observe any pattern, e.g. commands contributed via VS Code extensions do not work?

hansd56:
I am not so sure why you implemented the feature. But for me as a user I would expect that I can pop in any command from the command palette and it should work. What else can you do with it? It would certainly be useful.

JonasHelming:
@ndoschek Can you add this to the backlog, please

ndoschek:
Thanks for the detailed report, @hansd56.
I can reproduce that when adding commands contributed by VS Code extensions.
We're currently working on improving VS Code extension compatibility, as there are still some known issues in this area.
~~This specific issue might also be related to #16082, as the console logs show similar errors.~~

Edit: turns out, it is more of a general plugin-ext issue:
If commands are triggered from plugin-contributed toolbar items (either main toolbar or also widget toolbars) and handover the widget as an argument, case serialization errors and prevent command execution.
I opened a PR for this now

Issue #15932, opened 2025-07-01
[Epic] Improve compatibility with VS Code extensions (e.g. GitLens)


ndoschek:
### Summary

Theia aims to support VS Code extensions, but some extensions rely on specific APIs or behaviors that are not yet implemented or fully compatible.

This umbrella issue serves to discuss, track, and raise additional observations for improving compatibility with such extensions.
GitLens is one concrete example where certain features do not work as expected, like the support of command links (`command:` URIs) used heavily in the GitLens UI (see sub issue).

Contributions of further examples, incompatibilities, or insights are welcome to guide ongoing improvements.

PR #16230, opened 2025-08-28
fix: add argument processor for plugin toolbar command arguments

ndoschek:
#### What it does

If commands are triggered from plugin-contributed toolbar items (main or widget toolbars) and the widget is passed as an argument, process the argument and return only the widget ID. This prevents serialization errors and allows the command to execute properly.

This not only fixes custom toolbar items that add extension-contributed commands, it also fixes widget toolbar actions in extension views (if a corresponding handler is registered).

Part of GH-15932
Resolves GH-16223

#### How to test

1. Add custom toolbar item to the main toolbar and select a command contributed by an extension (e.g. from GitLens or cSpell). This should be correctly executed now (if the handler is available though). Previously, this was not executed and a ` Error during encoding: 'Maximum call stack size exceeded'` was printed to console.
2. Test a few widget toolbar actions contributed by GitLens for example:
   - File annotations in an editor:
     image
   - Commit Graph Settings in the GitLens graph view:
     image
   - GitLens actions in the native Theia SCM View
     image

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/plugin-ext-argument-processor.ts
+++ b/packages/plugin-ext/src/main/browser/plugin-ext-argument-processor.ts
@@ -1,0 +1,39 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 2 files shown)


eneufeld (approved):
Works great. Thank you!

PR #16231, opened 2025-08-28
[browser-only] Search In Workspace / Find In Files

kachurun:
#### What it does

1. This PR fixes broken Search in workspace (Cmd + Shift + F) functionality in the browser-only environment.
2. This PR fixes broken Find in Files (Cmd + P) functionality in the browser-only environment.

#### How to test

**"Search in Workspace"** and **"Find in files"** should work as expected.
It also respect .gitignore and .ignore / .rgignore files so behaviour should be the same as with normal/electron version.

#### Follow-ups

Based on https://github.com/eclipse-theia/theia/pull/16187

#### Breaking changes

#### Attribution

Would be nice

#### Review checklist

#### Reminder for reviewers


kachurun:
I think this is ready to test (after https://github.com/eclipse-theia/theia/pull/16187)

kachurun:
@sdirix @ndoschek
I think it’s ready for review and testing 🎉
All search and find operations are working now.

packages/search-in-workspace/src/browser/search-in-workspace-widget.tsx
```diff
@@ -427,30 +427,29 @@ export class SearchInWorkspaceWidget extends BaseWidget implements StatefulWidge
         const searchOnType = this.searchInWorkspacePreferences['search.searchOnType'];
         if (searchOnType) {
             const delay = this.searchInWorkspacePreferences['search.searchOnTypeDebouncePeriod'] || 0;
+
             window.clearTimeout(this._searchTimeout);
-            this._searchTimeout = window.setTimeout(() => this.doSearch(e), delay);
+            this._searchTimeout = window.setTimeout(() => {
```
kachurun: This code fixes the bug causing a double search call when pressing Enter.

Also, I think it's a bit clearer to have this code here rather than in a separate doSearch, because there are many methods with very similar names like search, doSearch, performSearch, onKeydownSearch, and it's really unclear which one is called when.

packages/filesystem/src/browser-only/file-search.ts
```diff
@@ -0,0 +1,184 @@
+// *****************************************************************************
```
kachurun, line 1: This is here in packages/core because these helpers are used by both the file-search and search-in-workspace packages

packages/filesystem/src/browser-only/file-search.ts
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * Processes ignore files (.gitignore, .ignore, .rgignore) in a directory.
+ * @param dir - The directory URI to process
+ * @param read - Function to read the ignore file content
+ * @returns Array of processed ignore patterns
+ */
+export async function getIgnorePatterns(dir: URI, read: (uri: URI) => Promise<string>): Promise<string[]> {
+    const fromPath = dir.path.toString();
+    const ignoreFiles = await Promise.allSettled(
+        ['.gitignore', '.ignore', '.rgignore'].map(file => read(dir.resolve(file)))
```
kachurun, line 109: These are the files that ripgrep reads by default in the normal version

kachurun:
@sdirix @ndoschek 
Hi guys! Can you please take a look at this second part of the improvements for browser-only? It's a smaller update this time.

packages/search-in-workspace/src/browser/search-in-workspace-widget.tsx
sdirix: I understand the motivation but we should not change protected method names for minor stylistic reasons as this unnecessarily breaks adopters. Also it would be nice to be able to override `doSearch`.

Within Theia we often have the pattern that a `x` method calls a `doX` method so the flow is not surprising.

packages/search-in-workspace/src/browser/search-in-workspace-widget.tsx
```diff
@@ -427,30 +427,29 @@ export class SearchInWorkspaceWidget extends BaseWidget implements StatefulWidge
...
+                        this.searchTerm = searchValue;
+                        this.performSearch();
+                    }
+                }
+            }, delay);
         }
     };
 
-    protected readonly onKeyDownSearch = (e: React.KeyboardEvent) => {
+    protected readonly searchByEnterKey = (e: React.KeyboardEvent) => {
```
sdirix: `onKeyDownSearch` is also arguably better than `searchByEnterKey` because it's handed over as a `onKeyDown` callback.

packages/search-in-workspace/src/browser/search-in-workspace-result-tree-widget.tsx
```diff
@@ -646,17 +642,20 @@ export class SearchInWorkspaceResultTreeWidget extends TreeWidget {
             searchOptions.maxResults -= numberOfResults;
         }
 
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        let pendingRefreshTimeout: any;
+        let pendingRefreshTimeout: NodeJS.Timeout | undefined;
```
sdirix: This is browser code so we should not type it as a NodeJS Timeout

packages/filesystem/src/browser-only/file-search.ts
sdirix: Fells like `@theia/filesystem` could be the better place for this

sdirix (approved):

packages/search-in-workspace/src/browser/search-in-workspace-widget.tsx
kachurun: Ok got it 👍

kachurun: I see you reverted that change. Good, thank you!

sdirix:
@kachurun Thank you very much for your great improvements for the browser-only variant of Theia! :heart:

PR #16235, opened 2025-08-29
[debug] Show the error message when setting a variable fails

### LINKED ISSUE #13728: Debugger doesn't show error of setVariable if it failed

pisv:
#### What it does

Fixes #13728.

#### How to test

```js
let a = 1;
console.log(a);
```

2. Double-click on the local `a` variable in the `Debug` view.

3. Try entering an invalid value for the variable like `@` or `foo` and press the `OK` button in the dialog.

4. Verify that the `OK` button becomes disabled and the appropriate error message is shown to the user in the dialog.

5. Enter a correct value for the variable, e.g. `123` and press `OK`.

6. Verify that the value has been correctly applied to the variable by the debugger.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/console/debug-console-items.tsx
+++ b/packages/debug/src/browser/console/debug-console-items.tsx
@@ -18,7 +18,7 @@ import * as React from '@theia/core/shared/react';
 import { DebugProtocol } from '@vscode/debugprotocol/lib/debugProtocol';
 import { SingleTextInputDialog } from '@theia/core/lib/browser';
 import { ConsoleItem, CompositeConsoleItem } from '@theia/console/lib/browser/console-session';
-import { DebugSession } from '../debug-session';
+import { DebugSession, formatMessage } from '../debug-session';
 import { Severity } from '@theia/core/lib/common/severity';
 import * as monaco from '@theia/monaco-editor-core';
 import { nls } from '@theia/core';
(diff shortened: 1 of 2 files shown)


packages/debug/src/browser/console/debug-console-items.tsx
```diff
@@ -244,12 +240,28 @@ export class DebugVariable extends ExpressionContainer {
...
+                    try {
+                        await this.setValue(value);
+                    } catch (error) {
+                        console.error('setValue failed:', error);
+                        if (error.body?.error) {
+                            const errorMessage: DebugProtocol.Message = error.body.error;
+                            if (errorMessage.showUser) {
+                                return formatMessage(errorMessage.format, errorMessage.variables);
+                            }
+                        }
```
sdirix, line 258: If we catch an error we should probably always return an error message, even if we didn't get one from the error. At the moment we will end up returning `true` if there is an error with an unexpected type.

sdirix (changes requested):
Works for me, there is just a minor detail regarding the error flow

packages/debug/src/browser/console/debug-console-items.tsx
pisv: Currently, if setting a variable fails, the error is logged in any case. But it is only manifested to the user in the dialog if the `showUser` property is true. I think that this behaviour matches the spec. It also makes sense to me, since errors of other kinds (e.g. system errors) have generally nothing to do with the user input. So, why should we disable the OK button and keep the dialog open in case of any error?

sdirix: I agree, thanks for the explanation

sdirix (approved):

pisv:
@sdirix Thanks a lot for your review and approval! 👍

PR #16238, opened 2025-09-01
fix: cannot open same folder once declined exit confirmation

### LINKED ISSUE #16227: Cannot open same folder once declined exit confirmation
**Bug Description**  
Nothing happens when opening a folder after declining the exit‑confirmation dialog.

**Steps to Reproduce**  

1. Add the following settings (for example, in `settings.json`):  
   ```json
   {
       "workspace.preserveWindow": true,
       "application.confirmExit": "always"
   }
   ```
2. Open **Folder A**.  
3. Choose **File → Open Folder…** and select **Folder B**. The *“Do you really want to exit?”* confirmation dialog appears.  
4. Click **No**. The current workspace remains **Folder A**.  
5. Again choose **File → Open Folder…** and select **Folder B**. Nothing happens – the IDE does not switch to the new folder.

**Additional Information**  

- **Operating System:** Windows 10  
- **Theia version:** 1.64.1  
- **Reproduction GIF:**  

  Reproduction GIF

**Root‑cause hypothesis**  
The code that changes the workspace first updates the workspace URI and *then* runs the veto check. If the user vetoes the operation (by clicking **No**), the URI is not restored to its previous value. Consequently, Theia believes the requested folder is already open, so subsequent “open folder” commands are ignored.

**Suggested fix**  

**Rollback the workspace URI** when a veto occurs.  
   - Store the current workspace URI before attempting the change.  
   - Run the veto check *before* committing the new URI, or restore the saved URI if `veto` returns `true`.

pisv:
#### What it does

Fixes #16227.

#### How to test

Use the steps to reproduce the issue #16227 to verify that the issue is now fixed. Also, verify that there are no regressions around opening and closing workspaces in general.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,14 @@
 
 - [ai-chat-ui] add chat input history with arrow key navigation [#16174](https://github.com/eclipse-theia/theia/pull/16174) - contributed on behalf of STMicroelectronics
 - [dependencies] increased minimum node version to 20.x [#16218](https://github.com/eclipse-theia/theia/pull/16218) - Contributed on behalf of STMicroelectronics
+- [workspace] fix: cannot open same folder once declined exit confirmation [#16238](https://github.com/eclipse-theia/theia/pull/16238)
 
 <a name="breaking_changes_1.65.0">[Breaking Changes:](#breaking_changes_1.65.0)</a>
 
 - [ai-ide]`TaskContextFileStorageService.getStorageLocation` and `TaskContextFileStorageService.watchStorage` modified to run synchronously, with corresponding change of signature. [#16063](https://github.com/eclipse-theia/theia/pull/16063)
 - [core] Make Preferences available in the back end [#16017](https://github.com/eclipse-theia/theia/pull/16017) - Contributed on behalf of STMicroelectronics\
 This PR allows to user default and user preferences in the back-end process. See [Migration.md](./doc/Migration.md) for breaking API changes
+- [workspace] `WorkspaceService.reloadWindow` now requires the workspace path string parameter [#16238](https://github.com/eclipse-theia/theia/pull/16238)
 
 ## 1.64.0 - 7/31/2025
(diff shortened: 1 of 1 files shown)


packages/workspace/src/browser/workspace-service.ts
```diff
@@ -530,25 +526,20 @@ export class WorkspaceService implements FrontendApplicationContribution, Worksp
...
                 // Fall back to reloading the current window in case the browser has blocked the new window
-                this._workspace = uri;
-                this.logger.error(error.toString()).then(() => this.reloadWindow());
+                this.logger.error(error.toString()).then(() => this.reloadWindow(workspacePath));
             }
         }
     }
 
-    protected reloadWindow(options?: WorkspaceInput): void {
+    protected reloadWindow(workspacePath: string, options?: WorkspaceInput): void {
```
ndoschek, line 540: Just a small suggestion,  we could use a default value here to avoid a breaking change in the signature. Wdyt?

```suggestion
    protected reloadWindow(options?: WorkspaceInput, workspacePath = ''): void {
```

ndoschek (commented):
Hi @pisv, thanks for contributing this fix! Works great! :+1: 
I added a small suggestion inline, please have a look what you think, thanks!

packages/workspace/src/browser/workspace-service.ts
pisv: @ndoschek Many thanks for your review! 👍

Since it is just a protected method, hopefully the breaking change should not be very disruptive, so I have decided to align the signature of this method with its sibling, `openNewWindow`. It would work better in the long run IMHO.

ndoschek: Okay that's fine by me, it will mainly concern adopters, but as it is listed in the breaking changes, that's acceptable, thanks!

PR #16242, opened 2025-09-02
New contribution point to modify the shell layout data

dmm9:
#### What it does

Allows adopters to modify the shell layout data when it is stored/restored (e.g. to remove certain tabs, modify tab positions, pin/unpin,...).

One use case is when widgets have special initialization sequences, so that those widgets should be reopened on every application start. With this contribution, those tabs can be removed from the layout data when it is restored.

See discussion: [https://github.com/eclipse-theia/theia/discussions/16151](https://github.com/eclipse-theia/theia/discussions/16151)

#### How to test

Create a frontend contribution that implements `ShellLayoutTransformer`. Modify layout data.
Reopen theia and see thae implemented layout modifications.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/shell/shell-layout-restorer.ts
+++ b/packages/core/src/browser/shell/shell-layout-restorer.ts
@@ -114,11 +114,7 @@ export interface ApplicationShellLayoutMigration {
 export const ShellLayoutTransformer = Symbol('ShellLayoutTransformer');
 /**
  * This contribution point allows arbitrary modifications to the shell layout
- * data when it is stored/restored.
- *
- * The layout transformations are applied during restore operations. This way,
- * when there is an application update, new layout transformations take
- * effect from the first application run.
+ * data when it is restored.
  */
 export interface ShellLayoutTransformer {
     /**


dmm9:
@sdirix This should do the trick (as discussed [here](https://github.com/eclipse-theia/theia/discussions/16151)). But I haven't found a frontend contribution within the examples to test it. Which would be the easiest way to test it?

packages/core/src/browser/shell/shell-layout-restorer.ts
```diff
@@ -111,6 +111,23 @@ export interface ApplicationShellLayoutMigration {
...
+ * The layout transformations are applied during restore operations. This way,
+ * when there is an application update, new layout transformations take
+ * effect from the first application run.
+ */
+export interface ShellLayoutTransformer {
+    /**
+     * Modifies the shell layout data before it is restored.
+     * @param layoutData 
+     */
+    transformLayout(layoutData: ApplicationShell.LayoutData): void;
```
sdirix: I think we should have two methods:
 - `onStore`
 - `onRestore`
 
 And then call them accordingly

sdirix (commented):
>@sdirix This should do the trick (as discussed https://github.com/eclipse-theia/theia/discussions/16151). But I haven't found a frontend contribution within the examples to test it. Which would be the easiest way to test it?

In the [api-samples](https://github.com/eclipse-theia/theia/tree/45521278237db8ce1083823d423c96a9673cc21b/examples/api-samples/src/browser) you could add an example transformer which always filters the "Sample uncloseable view"

packages/core/src/browser/shell/shell-layout-restorer.ts
dmm9: I first implemented those two methods, but then realized it only adds confusion and no real value to the adopters.

As a quirk, having two methods would require declaring them optional - therefore loosing ts warnings for unimplemented interface - or declaring the two methods mandatory - therefore forcing adopters to create empty methods - or choosing one mandatory and one optional - therefore leaving the optional method kind of pointless.

Do you have in mind any use case where onStore would be needed at all?

examples/api-samples/src/browser/view/sample-unclosable-view-contribution.ts
```diff
@@ -74,6 +74,36 @@ export class SampleUnclosableViewContribution extends AbstractViewContribution<S
         }
         return false;
     }
+
+    transformLayout(layoutData: ApplicationShell.LayoutData): void {
```
sdirix: I would add a bit of TSDoc explaining what is implemented here, i.e. it is made sure that the SampleViewUnclosableView is never restored

packages/core/src/browser/shell/shell-layout-restorer.ts
sdirix: The adopter might not want to persist specific information.

It's unnecessarily restrictive to only transform layouts on restore if they could equally be modified on store. The choice of only doing it on restore seems arbitrary.

> As a quirk, having two methods would require declaring them optional - therefore loosing ts warnings for unimplemented interface - or declaring the two methods mandatory - therefore forcing adopters to create empty methods - or choosing one mandatory and one optional - therefore leaving the optional method kind of pointless.

I don't see an issue with having two optional methods. It would also be possible to enforce at least one method being present via Typescript unions, but I don't really see a benefit of doing that.

dmm9: It is not arbitrary, it is a carefully considered design decision. TLDR: to provide the expected behavior during app updates and keep integration efforts for adopters as simple as possible.

When transforming layout during store, then users will not see the changes after an application update if the application update includes changes in the layout tranformations, because the layout data was stored using the old method. Also adopters that are not aware of such details will be left wondering why their layout transformation changes did not take effect on the first run. This does not happen if the transformation is applied during restore.

On the other side, I can't imagine any use case for consuming the serialized layout data outside of the layout restorer. So persisting information that will later be changed doesn't seem to cause any issue. If this was the case, such modifications of the persisted data would already be in place.

Both cases, changes in layout transformations and consuming stored layout data, are really rare. But the case of consuming serialized layout data (excluding the layout restorer) seems far more unusual. 

Zooming out to the component, we are talking about the "shell layout restorer" component, whose responisibility is . We want to offer a possibility to **manipulate the layout before it is restored**. When exactly this happens is an implementation detail that is not relevant for adopters. And therefore, offering two methods to accomplish the same result seems bloated and it also requires adopters an additional effort to understand those implementation details.

To sum up, the reason for doing transformations only during restore is to provide the expected behavior during app updates and keep integration efforts for adopters as simple as possible.

What do you think now? Should we go for the single method or for two methods?

sdirix: Generally speaking, I think `ShellLayoutRestorer` is misleading as the service is responsible for both storing and restoring layout data. It's really more a `ShellLayoutPersister`.

I agree that `onStore` might not be used that much. So adding both might be overkill (for now). However, to keep our options open, what if we rename `transformLayout` to `onRestore`? Then we can easily add `onStore` later if there is a use case.

This gives us the simplicity you requested while keeping the door open for future extensibility. What do you think?

dmm9: This seems a good compromise. I will change to `transformLayoutOnRestore`, so that the method name is more self-explanatory on the contributions, and not confuse with the lifecycle events onStart, onStop,...
```typescript
@injectable()
export class SampleUnclosableViewContribution extends AbstractViewContribution<SampleViewUnclosableView> implements TabBarToolbarContribution, ShellLayoutTransformer {
    // ...
        // ...
    }

    // Makes sure the 'Sample Unclosable View' view is never restored after app restart.
    transformLayoutOnRestore(layoutData: ApplicationShell.LayoutData): void {
        this.pruneConfig(layoutData.mainPanel?.main);
    }
}
```

sdirix: great!

dmm9:
Changes done.

packages/core/src/browser/shell/shell-layout-restorer.ts
```diff
@@ -111,6 +111,23 @@ export interface ApplicationShellLayoutMigration {
...
 
+export const ShellLayoutTransformer = Symbol('ShellLayoutTransformer');
+/**
+ * This contribution point allows arbitrary modifications to the shell layout
+ * data when it is stored/restored.
+ *
+ * The layout transformations are applied during restore operations. This way,
+ * when there is an application update, new layout transformations take
+ * effect from the first application run.
+ */
```
sdirix:
```suggestion
/**
 * This contribution point allows arbitrary modifications to the shell layout
 * data when it is restored.
 */
```

Issue #15626, opened 2025-05-21
Authentication service does not react to session changes


dmm9:
### Bug Description:

In general, the `AuthenticationService` is quite outdated in comparison with VSCode, and has bugs and behaviors that were corrected in VSCode long ago.

1. When an `AuthenticationProvider` is registered, the `AuthenticationService` does not consume `onDidChangeSessions` event. See [current VSCode implementation](https://github.com/microsoft/vscode/blob/86efdcd2c1dbf07b85d507378854ad45cefcc3b4/src/vs/workbench/services/authentication/browser/authenticationService.ts#L205-L209) for the expected behavior.
  * PS Maybe this is also the cause for #13599
2. The method `AuthenticationProvider.hasSessions()` does not exist in VSCode. However, failing to implement it properly in a Theia application results in the Accounts widget not being displayed properly. Not related to the previous point, but it is a deviation from the VSCode API.

### Steps to Reproduce:

Create and register an `AuthenticationProvider`

### Additional Information

- Theia Version: master

dmm9:
@tsmaeder fyi

PR #16252, opened 2025-09-05
fix: authentication service not reacting to session changes

dmm9:
The authentication service now subscribes to changes in authentication provider sessions.

#### What it does

The Authentication service is not reacting to changes in the providers' sessions. Therefore, it shows no login info (option to sign out in the auth menu on the bottom left) when the user successfully logs in.

Fixes: #15626

#### Follow-ups

In general, the authentication service is outdated compared to that of vscode. Some follow-up work could be necessary to keep 100% compatibility.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/authentication-service.ts
+++ b/packages/core/src/browser/authentication-service.ts
@@ -156,6 +156,7 @@ export class AuthenticationServiceImpl implements AuthenticationService {
     private sessionMap = new Map<string, DisposableCollection>();
 
     protected authenticationProviders: Map<string, AuthenticationProvider> = new Map<string, AuthenticationProvider>();
+    protected authenticationProviderDisposables: Map<string, DisposableCollection> = new Map<string, DisposableCollection>();
 
     private readonly onDidRegisterAuthenticationProviderEmitter: Emitter<AuthenticationProviderInformation> = new Emitter<AuthenticationProviderInformation>();
     readonly onDidRegisterAuthenticationProvider: Event<AuthenticationProviderInformation> = this.onDidRegisterAuthenticationProviderEmitter.event;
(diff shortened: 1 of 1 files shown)


packages/core/src/browser/authentication-service.ts
```diff
@@ -260,6 +261,14 @@ export class AuthenticationServiceImpl implements AuthenticationService {
             throw new Error(`An authentication provider with id '${id}' is already registered.`);
         }
         this.authenticationProviders.set(id, authenticationProvider);
+
+        const disposables = new DisposableCollection();
+        disposables.push(authenticationProvider.onDidChangeSessions(e => this.updateSessions(authenticationProvider.id, e)));
+        if (Disposable.is(authenticationProvider)) {
+            disposables.push(authenticationProvider);
```
colin-grant-work: This is a little bit suspect. Here, this service takes ownership of the lifecycle of the registered authentication provider when nothing about its API indicates that that would be the case. I would prefer that we own the listener but not dispose the provider - we don't know what the caller's plans are for it.

dmm9: This change replicates the behavior in VSCode. Please refer to [this part of the VSCode Authentication Service](https://github.com/microsoft/vscode/blob/86efdcd2c1dbf07b85d507378854ad45cefcc3b4/src/vs/workbench/services/authentication/browser/authenticationService.ts#L205-L209) which is also linked in the Theia issue #15626.

colin-grant-work: The reason I say that the service should not dispose of the provider object is that we don't know that it is appropriate to dispose of the authentication provider when it is deregistered: it could be a multi-purpose implementation that is also performing other functions elsewhere, and disposing of it on deregistration as an authentication provider might hamstring it in those other functions. Only the caller is aware of the intended lifecycle of the implementation, so if we're missing disposal, it should be added on the caller's side, not in this service. The VSCode code is also incorrect from an API perspective, in my view, but VSCode doesn't have to be prepared for adopters downstream to do unexpected things: they control all callers.

So to be clear, the listener should indeed be disposed of by the authentication service - it is created by the service and relevant only as long as the provider is registered - but the provider itself should not be disposed of by the service, because the service doesn't know that the provider object's lifecycle is tied to its registration.

dmm9: I just removed those lines

PR #16265, opened 2025-09-09
fix: arrow up/down only trigger prompt history if visual top/bottom line

planger:
#### What it does

With the previous implementation we already switched to the previous prompt, if we pushed `up` in the top most line from the text model point of view, not taking into account that the top line may actually be wrapped, resulting in being visually NOT the top/bottom line.

With this PR, we check if we are visually in the top-most/bottom-most line before we trigger the prompt history.

#### How to test

To reproduce/test:
* Paste a couple of long lines into the chat input
* Navigate up/down with the cursor keys
* Observe that only if VISUALLY in the top/bottom line, the history navigation is triggered

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


planger:
Great suggestion, thank you! Please see 8dd8cd5.

planger:
Thanks again for the review @ndoschek!
I now more closely aligned the behavior of the chat input placeholder with the search placeholder, that is, only show the hint with the history if focused. Also I improved the styling to avoid shifting line-height.
Unfortunately the formatter now made changes across the file.
I guess we want the CSS to conform to the configured settings anyway, so I didn't clean those formatting changes.
Thanks!

PR #16282, opened 2025-09-12
fix: Correctly handle reasoning output of openai/gpt-oss-120b

### LINKED ISSUE #16186: [Theia AI] Groq does not work as a provider with Theia AI
It would be nice if Groq would work as a provider, as it is very fast and also supports interesting models such as GPT-OSS. However, my first try failed, seems to be an issue wih their OpenAI API implementation, but we should check.

### Steps to Reproduce:

1. Enter groq as a custom OpenAI Provider:
{
        "model": "openai/gpt-oss-120b",
        "url": "https://api.groq.com/openai/v1",
        "id": "groq/gpt-oss-120b",
        "apiKey": "key",
    },

2. Set the model for an agent
3. Do a prompt that triggers a tool call:

Image

### Additional Information

- Operating System: Linux
- Theia Version: 1.64

eneufeld:
#### What it does

- Handle reasoning output by outputing it as thinking but removing it otherwise
- improve tool description

fixes #16186

#### How to test

See initial issue

#### Follow-ups

The OpenAI ChatCompletion API should be replaced with the newer Response API.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-openai/src/node/openai-streaming-iterator.ts
+++ b/packages/ai-openai/src/node/openai-streaming-iterator.ts
@@ -83,6 +83,15 @@ export class StreamingAsyncIterator implements AsyncIterableIterator<LanguageMod
                     }
                 });
             }
+            // OpenAI can push out reasoning tokens, but can't handle it as part of messages
+            if (snapshot?.choices[0]?.message && Object.keys(snapshot.choices[0].message).includes('reasoning')) {
+                const reasoning = (snapshot.choices[0].message as { reasoning: string }).reasoning;
+                this.handleIncoming({ thought: reasoning, signature: '' });
+                // delete message parts which cannot be handled by openai
+                delete (snapshot.choices[0].message as { reasoning?: string }).reasoning;
+                delete (snapshot.choices[0].message as { channel?: string }).channel;
+                return;
+            }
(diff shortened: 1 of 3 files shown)


JonasHelming:
I can confirm I can use GPT-OSS now on Groq, @sdirix could you have another quick look, please? (no need to retest GPT-OSS)

packages/ai-openai/src/node/openai-streaming-iterator.ts
```diff
@@ -83,6 +83,14 @@ export class StreamingAsyncIterator implements AsyncIterableIterator<LanguageMod
                     }
                 });
             }
+            // OpenAI can push out reasoning tokens, but can't handle it as part of messages
+            if (Object.keys(snapshot.choices[0].message).includes('reasoning')) {
+                const reasoning = (snapshot.choices[0].message as { reasoning: string }).reasoning;
+                this.handleIncoming({ thought: reasoning, signature: '' });
+                delete (snapshot.choices[0].message as { reasoning?: string }).reasoning;
+                delete (snapshot.choices[0].message as { channel?: string }).channel;
```
sdirix, line 92: Why do we need to delete them here?

sdirix (changes requested):
Please check the openai test cases as they are failing now

PR #16289, opened 2025-09-15
Add `Evaluate in Debug Console` command

### LINKED ISSUE #10281: “Evaluate in debug console” command is missing
Hi,
I am using v1.17.2(beta) on macos

Vscode has a command “Evaluate in debug console” (see [here](https://github.com/microsoft/vscode/issues/98390)).
It seems like Theia does not have this command, so I basically need to copy and paste the code to the debug console and then evaluate it (I am using python).
This was also confirmed by @vince-fugnitto

pisv:
#### What it does

Closes #10281.

#### How to test

Suspend execution of a launched program (e.g. at a breakpoint), select an expression in the source editor you'd like to evaluate against the current stack frame, right-click on the selection and choose `Evaluate in Debug Console`. (If the selection is empty, it will evaluate the current line's trimmed content, just as in VS Code.)

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-frontend-application-contribution.ts
+++ b/packages/debug/src/browser/debug-frontend-application-contribution.ts
@@ -259,6 +259,11 @@ export namespace DebugCommands {
         id: 'editor.debug.action.showDebugHover',
         label: 'Debug: Show Hover'
     });
+    export const EVALUATE_IN_DEBUG_CONSOLE = Command.toDefaultLocalizedCommand({
+        id: 'editor.debug.action.selectionToRepl',
+        category: DEBUG_CATEGORY,
+        label: 'Evaluate in Debug Console'
+    });
     export const JUMP_TO_CURSOR = Command.toDefaultLocalizedCommand({
         id: 'editor.debug.action.jumpToCursor',
         category: DEBUG_CATEGORY,
@@ -660,6 +665,7 @@ export class DebugFrontendApplicationContribution extends AbstractViewContributi
 
         const DEBUG_EDITOR_CONTEXT_MENU_GROUP = [...EDITOR_CONTEXT_MENU, '2_debug'];
         registerMenuActions(DEBUG_EDITOR_CONTEXT_MENU_GROUP,
+            DebugCommands.EVALUATE_IN_DEBUG_CONSOLE,
             DebugCommands.JUMP_TO_CURSOR,
             DebugCommands.RUN_TO_CURSOR,
             DebugCommands.RUN_TO_LINE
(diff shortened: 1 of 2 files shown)


packages/debug/src/browser/debug-frontend-application-contribution.ts
```diff
@@ -259,6 +259,11 @@ export namespace DebugCommands {
         id: 'editor.debug.action.showDebugHover',
         label: 'Debug: Show Hover'
     });
+    export const EVALUATE_IN_DEBUG_CONSOLE = Command.toDefaultLocalizedCommand({
+        id: 'editor.debug.action.selectionToRepl',
```
ndoschek, line 263: minor: not sure about this command id, I would prefer a more descriptive/matching one, e.g.:
```suggestion
        id: 'editor.debug.action.evaluateInDebugConsole',
```

ndoschek (commented):
Thanks @pisv, great catch and thanks for the fix, it works perfectly and is a great improvement! :sparkles:

Just one small remark about the command id, please have a quick look inline, thanks!

packages/debug/src/browser/debug-frontend-application-contribution.ts
pisv: `editor.debug.action.selectionToRepl` is the id for the `Evaluate in Debug Console` command in VS Code.

Not sure whether it really matters to retain this id for compatibility with VS Code.

I do agree that `editor.debug.action.evaluateInDebugConsole` would be a better id.

ndoschek: Thanks @pisv for clarifying. Since this isn't a public VS Code API and we're using our own implementation, there's no need to stick with the original VS Code command id IMO.
If you don't mind, I’d prefer to go with the suggested `editor.debug.action.evaluateInDebugConsole` instead 👍

pisv: Sure thing! I'll commit your suggestion. Thank you for your comments! 👍

msujew:
FYI for issues such as https://github.com/eclipse-theia/theia/pull/16289#discussion_r2355343274, we have a special contribution that maps VSCode commands to Theia commands. See [here](https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext-vscode/src/browser/plugin-vscode-commands-contribution.ts). We should generally try to use the command names registered in VSCode, since plugins might use them, even if they are not part of the [public command API](https://code.visualstudio.com/api/references/commands). We have had issues in the past where plugins didn't work in Theia, because they were using "hidden" commands.

pisv:
> We should generally try to use the command names registered in VSCode, since plugins might use them, even if they are not part of the [public command API](https://code.visualstudio.com/api/references/commands). We have had issues in the past where plugins didn't work in Theia, because they were using "hidden" commands.

Well, in this case I think it would be better to just stick with the original VS Code id for this command, would not it?

ndoschek:
Thanks @msujew, did not think of that, but makes sense!
@pisv, sure, then please ignore my comment and you can drop my suggestion again, thanks!

pisv:
@ndoschek No worries! Done :-)

PR #16305, opened 2025-09-21
Enhance localization of the debug package

pisv:
#### What it does

This PR provides localization enhancements for the debug package.

#### How to test

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-frontend-application-contribution.ts
+++ b/packages/debug/src/browser/debug-frontend-application-contribution.ts
@@ -193,21 +193,21 @@ export namespace DebugCommands {
         category: DEBUG_CATEGORY,
         label: 'Add Function Breakpoint',
     });
-    export const ENABLE_SELECTED_BREAKPOINTS = Command.toDefaultLocalizedCommand({
+    export const ENABLE_SELECTED_BREAKPOINTS = Command.toLocalizedCommand({
         id: 'debug.breakpoint.enableSelected',
         category: DEBUG_CATEGORY,
         label: 'Enable Selected Breakpoints',
-    });
+    }, 'theia/debug/enableSelectedBreakpoints', DEBUG_CATEGORY_KEY);
     export const ENABLE_ALL_BREAKPOINTS = Command.toDefaultLocalizedCommand({
         id: 'debug.breakpoint.enableAll',
         category: DEBUG_CATEGORY,
         label: 'Enable All Breakpoints',
     });
-    export const DISABLE_SELECTED_BREAKPOINTS = Command.toDefaultLocalizedCommand({
+    export const DISABLE_SELECTED_BREAKPOINTS = Command.toLocalizedCommand({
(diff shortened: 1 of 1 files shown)


packages/debug/src/browser/debug-session.tsx
```diff
@@ -342,7 +342,7 @@ export class DebugSession implements CompositeTreeElement {
         try {
             const response = await this.connection.sendRequest('initialize', {
                 clientID: 'Theia',
-                clientName: 'Theia IDE',
+                clientName: nls.localize('theia/debug/TheiaIDE', 'Theia IDE'),
```
ndoschek, line 345: I'm not sure about this one. Overall, "Theia IDE" doesn't seem quite right, it might just be "Theia" as well.
But since it's an optional property, it's not that critical, and probably we could also just keep the original string?
@colin-grant-work, do you have any thoughts on this?

packages/debug/src/browser/debug-frontend-application-contribution.ts
ndoschek, line 1: I noticed one context menu entry in the context menu of breakpoints which is not correctly localized.

This should probably be a localized command instead of a default one then:
`export const DISABLE_SELECTED_BREAKPOINTS = Command.toLocalizedCommand({ ...`


image

pisv: Good catch, thanks! Indeed, there are no matches for `Enable Selected Breakpoints` and `Disable Selected Breakpoints` in VS Code, so they need to be localized in Theia itself. I'll fix it.

packages/debug/src/browser/debug-session.tsx
pisv: Not quite sure about it either, but the `clientName` property is documented as 'The human-readable name of the client using this adapter' in the debug protocol, so I think it should be localized. This will also enable adopters to override the default client name using a `TextReplacementContribution` (#14712).

UPD: Having said that, both `clientID` and `clientName` properties are indeed optional, so they might just be removed entirely as an alternative.

UPD: Another idea would be to use `FrontendApplicationConfigProvider.get().applicationName` as the client name. It will default to 'Eclipse Theia' then.

What do you think?

packages/debug/src/browser/debug-frontend-application-contribution.ts
pisv: Hopefully, this is fixed in https://github.com/eclipse-theia/theia/pull/16305/commits/bc3de3391e012ae35f57f236001985e742e8e958. Just to clarify: the commands will not be immediately translated to other languages, it will just enable such translation.

packages/debug/src/browser/debug-session.tsx
ndoschek: Thanks for the insights @pisv!  
Given the `TextReplacementContribution`, keeping the original string and localizing it sounds good then.
Let's go with the current approach 👍

PR #16306, opened 2025-09-22
fix React warning for duplicate createRoot calls on singleton ReactDi…

YueCHEN195:
…alogs

#### What it does
This PR fixes an issue where singleton ReactDialog components would trigger React warnings about duplicate createRoot() calls after being closed and reopened.
The isMounted flag was only set to true in the constructor.When a singleton dialog was closed, isMounted was set to false via the disposable cleanup.On subsequent openings (onUpdateRequest), the check if (!this.isMounted) would always be true because the flag was never reset.This caused createRoot() to be called again on the same DOM container (this.contentNode), violating React's rule that createRoot() should only be called once per container. This resulted in the warning: Warning: You are calling ReactDOMClient.createRoot() on a container that has already been passed to createRoot() before....
​The Fix:​​ Added this.isMounted = true; inside the if (!this.isMounted) block within the onUpdateRequest method.

image

#### How to test

ceate a singleton dialog and close and reopen it.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/dialogs/react-dialog.tsx
+++ b/packages/core/src/browser/dialogs/react-dialog.tsx
@@ -42,6 +42,7 @@ export abstract class ReactDialog<T> extends AbstractDialog<T> {
         super.onUpdateRequest(msg);
         if (!this.isMounted) {
             this.contentNodeRoot = createRoot(this.contentNode);
+            this.isMounted = true;
         }
         this.contentNodeRoot?.render(<>{this.render()}</>);
     }


colin-grant-work:
@YueCHEN195, for your PR to be merged, it's required that you sign the [Eclipse Foundation contributor agreement](https://www.eclipse.org/legal/eca/). Let us know when you've signed the agreement and we can rerun the check.

colin-grant-work:
If you've already signed the ECA, the validation may have failed due to an outage at the foundation, so we can also rerun the validation when it's back up.

YueCHEN195:
> If you've already signed the ECA, the validation may have failed due to an outage at the foundation, so we can also rerun the validation when it's back up.

I have already signed the ECA, but the license check still failed

sdirix:
I retriggered the ECA check and it completed successfully. The License check is regarding our dependencies. Sadly the tool we are using is a bit flaky so it fails often. However we can still merge this improvement as it obviously does not contain any changes to the dependencies.

PR #15103, opened 2025-03-03
Option to put close icons at the start of the tab label on macOS

cdamus:
#### What it does

Adds a new `window.tabCloseIconPlacement` preference for whether to present the Close (X) icon in tab titles on the left or the right of the tab.
~Defaults to the left on macOS platform in conformity with the OS's native tab controls.~
The tab bar rendering is updated to take the preference into account.

The preference is suppressed in the Settings view on non-Mac platforms as being largely irrelevant on those systems.

Fixes eclipse-theia/theia-ide#460

#### How to test

On a macOS system, build and launch the Theia example app. See that tabs in the main editor area are on the left side where they belong.

Open Settings and change the "Tab close icon placement" preference to "end". See the close icons move back to where they are presented on Linux and Windows platforms. Restore the default setting. See the close icons restored to the left side.

On Linux and Windows systems, verify that the "Tab close icon placement" preference does not appear in the Settings UI and that tab titles look as they ever did.

#### Follow-ups

None.

#### Breaking changes

- [ ] This PR introduces breaking changes and requires careful review. If yes, the breaking changes section in the [changelog](https://github.com/eclipse-theia/theia/blob/master/CHANGELOG.md) has been updated.

#### Attribution

None.

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

> [!NOTE]
> My testing is only on macOS platform as that is the platform that my system runs and the feature is meant to be disabled on other platforms. I'm depending on buddy tests from the Theia community to check for regressions on other platforms.

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

----

#### Update 17 June 2025

Changed the default close icon placement to the end on all platforms, including macOS, to maintain current behaviour.

----

CleanShot 2025-03-03 at 16 36 06

PR #16332, opened 2025-09-26
fix: do not report misleading 'invalid scope' warning for excluded preferences

CamilleLetavernier:
fixes eclipse-theia/theia/issues/16236

#### What it does

- do no lot a warning about "invalid scopes" for preferences that are excluded (e.g. because of unsupported OS), to avoid misleading users

#### How to test

- on MacOS: set `"window.titleBarStyle": "custom"` in the User Settings (Use text mode)
- on other OS: set `"window.tabCloseIconPlacement": "start"` in the User Settings (Use text mode)
- then check that this warning is **not** displayed: `WARN Preference window.titleBarStyle in user-storage:/user/settings.json can only be defined in scopes: Default, User.`

(Note: in both cases, as the property is not allowed on the specified OS, you should still see a warning in the Settings Editor, such as: `Property window.tabCloseIconPlacement is not allowed.`)

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/common/abstract-resource-preference-provider.ts
+++ b/packages/preferences/src/common/abstract-resource-preference-provider.ts
@@ -205,7 +205,7 @@ export abstract class AbstractResourcePreferenceProvider extends PreferenceProvi
             const oldValue = oldPrefs[prefName];
             const newValue = newPrefs[prefName];
             const schemaProperty = this.schemaProvider.getSchemaProperty(prefName);
-            if (schemaProperty) {
+            if (schemaProperty && schemaProperty.included) {
                 const scope = schemaProperty.scope;
                 // do not emit the change event if the change is made out of the defined preference scope
                 if (!this.schemaProvider.isValidInScope(prefName, this.getScope())) {


sdirix (approved):
The PR works as stated.

However I wonder why the `window.tabCloseIconPlacement` is excluded on Linux. It definitely has an effect on the UI

ndoschek:
@CamilleLetavernier is on vacation this week, but I guess we can wait until he's back next week.

CamilleLetavernier:
Yes, the setting indeed seems to work on Linux; although I don't know much about possible restrictions on other variants of the Linux desktop.

All I can say is that it was convenient to have this setting to test the current patch; as I could only find two Theia Preferences with the "exclude" constraint.

It could also be a matter of UI consistency: maybe MacOS is more likely to expect tab-icons at the start, whereas other platforms only expect it at the beginning; so the setting is more relevant to Mac users?

Either way, changing this setting is out of scope for this ticket, and I don't have enough information about why this setting is only included for Mac; so I wouldn't worry too much about it (here).

ndoschek:
The original PR (https://github.com/eclipse-theia/theia/pull/15103) introduced the `window.tabCloseIconPlacement` option specifically to align with macOS UI conventions. That’s likely why it was initially restricted to that platform.

I quickly checked and the setting would work fine on Linux as well, ~~and the default placement is already at the end, it makes sense to make it available for all platforms. I'll open a follow-up issue to track this.~~
**Edit:** a closer look on the PR shows that this was on purpose:
> The preference is suppressed in the Settings view on non-Mac platforms as being largely irrelevant on those systems.
So we leave it like that for now, in case someone wants that on other platforms, the restricition can just be lifted if needed.

<br/>

Thanks @CamilleLetavernier  for the original fix :smile:

Issue #15921, opened 2025-07-01
Consider using `fs.watch` for file watching


tsmaeder:
### Feature Description:
In node 19, recursive file watching was added to the nodejs file watching API. (see https://nodejs.org/docs/latest/api/fs.html#fswatchfilename-options-listener). We should consider using this API instead of the parcel watcher.

tsmaeder:
Depends on https://github.com/eclipse-theia/theia/issues/15857

JonasHelming:
I find this interesting, as I believe we still might have some issues with file watching. @ndoschek

Niklas-Dahlquist:
This upgrade would solve the issue experienced on Windows11, see this discussion on parcel watcher
https://github.com/parcel-bundler/watcher/issues/155

tsmaeder:
@Niklas-Dahlquist IMO the biggest advantage would be to have a general purpose, actively developed implementation. NSFW is not really active and the Parcel watcher is, well, for Parcel, not general consumption.

rschnekenbu:
I have the same findings here on my side, working on the popup window effect (https://github.com/eclipse-theia/theia/issues/14890). 

The current issue on Windows is that Watchman (one of the potential backend for parcel/watcher library) may not be found. In order to be found, The watchman backend tries to connect with https://github.com/parcel-bundler/watcher/blob/1ca032aa8339260a8a3bcf825c3a1a71e3e43542/src/watchman/WatchmanBackend.cc#L68. The getSockPath implementation currently relies on windows on this: `popen("watchman --output-encoding=bser get-sockname 2>/dev/null", "r");`  
This leads to that error message and the forementionned console window to popup. 

It is possible to set the backend to another default one on windows, which is slower, but that does not cause an error on startup and so no popup window.  As watchman is not found, we can safely assume that on windows, the slower backend is used in any case as a backup. So that should not impact performances. 

There were some long discussion on using parcel or other file watching library here: https://github.com/vitejs/vite/issues/13593

PR #16335, opened 2025-09-26
fix: avoid trying to use watchman for file system watcher

rschnekenbu:
#### What it does

**Avoid trying to select watchman for the various OS for File Watching.** 
By default, @parcel/watcher tries to use watchman as backend. For this, it tries to connect to the watchman service via socket. This opens a pop-up on windows, as there is an error message displayed when watchman is not found.
Via options, it is possible to select the backend and not let @parcel/watcher evaluate if watchman exists. Thus it avoids the popup window. 
As watchman is not expected to be deployed, this should not hurt current performances. It is still possible for a particular application to deploy specific options and choose again watchman, as the parcel options can be injected via a service.
references: 
@parcel/watcher priorities: https://github.com/parcel-bundler/watcher?tab=readme-ov-file#watching
@parcel/watcher code for backend selection: https://github.com/parcel-bundler/watcher/blob/8dfd99440b8666e81cc6865f68697ab272dfe83d/src/Backend.cc
 @parcel/watcher code for watchman backend detection: https://github.com/parcel-bundler/watcher/blob/8dfd99440b8666e81cc6865f68697ab272dfe83d/src/watchman/WatchmanBackend.cc#L49

#### How to test

Build Theia sample electron application on windows with the patch and try to start the electron app. The popup should not come.
The process monitor shall also not display any cmd.exe process with watchman command triggered.

#### Follow-ups

potential follow-up
#15921

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers


ndoschek (approved):
Thank you for the fix @rschnekenbu! 🙌 I tested both the Electron and browser versions. On my Windows VM, the popup window issue in Electron is resolved. I also did some quick file-watching tests, which worked well as far as I could judge.
I also tested on my Linux (Ubuntu 24) system, and everything still works as expected.

Waiting for a second look from Mark before merging though.

PR #16340, opened 2025-09-28
Fix `Add to Watch` command

pisv:
#### What it does

Currently, the `Add to Watch` command, which adds the variable selected in the Debug Variables widget to the list of watch expressions, is never visible, since it incorrectly assumes that the current widget can simultaneously be `DebugVariablesWidget` and `DebugWatchWidget`.

This PR fixes this issue and also aligns the implementation of this command with VS Code (for reference, see [variablesView.ts](https://github.com/microsoft/vscode/blob/887fc4d185b974b63f19867b5f7800263466dac0/src/vs/workbench/contrib/debug/browser/variablesView.ts#L804) in VS Code).

#### How to test

```js
let abc = { foo: 1, bar: '' };
console.log(abc);
```

2. When the breakpoint is hit, right-click on the `abc` variable in the Debug Variables widget, and choose `Add to Watch`.

Verify that the command has the same behavior as in VS Code.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/console/debug-console-items.tsx
+++ b/packages/debug/src/browser/console/debug-console-items.tsx
@@ -152,6 +152,9 @@ export class DebugVariable extends ExpressionContainer {
     get name(): string {
         return this.variable.name;
     }
+    get evaluateName(): string | undefined {
+        return this.variable.evaluateName;
+    }
     protected _type: string | undefined;
     get type(): string | undefined {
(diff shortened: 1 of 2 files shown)


ndoschek (approved):
Looks great, thanks for fixing this @pisv!
The command works as expected when adding variables to the watch list 🎉

PR #16350, opened 2025-09-29
Enhance localization of the ai-core package

pisv:
#### What it does

This PR provides localization enhancements for the ai-core package.

#### How to test

#### Follow-ups

If similar PRs would be welcome for other AI-related packages, I could try to follow up with them.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-core/src/browser/window-blink-service.ts
+++ b/packages/ai-core/src/browser/window-blink-service.ts
@@ -96,9 +96,9 @@ export class WindowBlinkService {
 
     private async blinkDocumentTitle(agentName?: string): Promise<void> {
         const originalTitle = document.title;
-        const alertTitle = agentName
-            ? nls.localize('theia/ai/core/blinkTitle/namedAgentCompleted', '🔔 Theia - Agent "{0}" Completed', agentName)
-            : nls.localize('theia/ai/core/blinkTitle/agentCompleted', '🔔 Theia - Agent Completed');
+        const alertTitle = '🔔 ' + (agentName
+            ? nls.localize('theia/ai/core/blinkTitle/namedAgentCompleted', 'Theia - Agent "{0}" Completed', agentName)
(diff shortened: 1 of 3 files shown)


packages/ai-core/src/browser/window-blink-service.ts
```diff
@@ -97,8 +97,8 @@ export class WindowBlinkService {
     private async blinkDocumentTitle(agentName?: string): Promise<void> {
         const originalTitle = document.title;
         const alertTitle = agentName
-            ? `🔔 Theia - Agent "${agentName}" Completed`
-            : '🔔 Theia - Agent Completed';
+            ? nls.localize('theia/ai/core/blinkTitle/namedAgentCompleted', '🔔 Theia - Agent "{0}" Completed', agentName)
```
ndoschek: I’d prefer if the emoji was placed outside the localized strings. For example:  

```ts
`🔔 nls.localize('theia/ai/core/blinkTitle/namedAgentCompleted', 'Theia - Agent "{0}" Completed', agentName)`
```

That way we avoid having emojis inside translations, since they are not guaranteed to be handled consistently across all languages (I did a quick regex search in the workspace using `\p{Extended_Pictographic}`).

I know a few strings already include emojis, mostly in the ai-* packages, but it would be good to align on this pattern. Could you please also update the `AI_CORE_PREFERENCES_TITLE` accordingly? Thanks! 🙏

pisv:
@ndoschek Thanks once again for your thorough review, much appreciated!

I've tried to address your comments in 9f469c002924b163946bccbf492a6416af4b10ad. 

I'm not quite sure whether the name in `newCustomAgentEntry` should be localized, since both the name and the prompt don't seem to be localized for other agents like Coder.

Please let me know if you think that the name needs to be localized in this instance.

Issue #16339, opened 2025-09-28
Settings - Can't reset String settings + other issues in display


safisa:
### Bug Description:
Hi,

There are several issues in the settings of the String type (string input):
1. The gear menu button does not open on the first time, only after adding some text to the input box.
2. The reset settings action does not work, with no error!
3. When moving to the workspace scope and then back to the user scope, it is then displayed as modified (vertical blue line), although it is not.

### Steps to Reproduce:

1. Open the browser app, for example, go to the ai-features.chat.defaultChatAgent setting
2. Check that the gear menu button does not open
3. Type something in the textbox, then open the gear menu button (it will open now), and reset the setting
4. It will do nothing

For the modified indicator issue:
1. go for an example to the ai-features.chat.defaultChatAgent setting (should be a key with no change before)
2. Change to the workspace scope, then back to the user scope.
3. You will see the setting as modified.

### Additional Information

- Operating System: Mac - chrome
- Theia Version: 1.65

ndoschek:
Thanks for reporting this, @safisa!
I can also reproduce it, it seems to behave weird for empty/non-set string properties, we'll look into it! 👀

rschnekenbu:
This likely seems a classical issue on comparing some default values being undefined and the text editor that returns '' when the value is empty on the string editor. At least this is the case for the modified aspect.

rschnekenbu:
see the code in https://github.com/eclipse-theia/theia/blob/1e2d4d5ef97386b09487f70e6a16495e5b204579/packages/preferences/src/browser/views/components/preference-string-input.ts#L40-L41 and https://github.com/eclipse-theia/theia/blob/1e2d4d5ef97386b09487f70e6a16495e5b204579/packages/preferences/src/browser/views/components/preference-node-renderer.ts#L294
It seems to be that the undefined case is not properly handled. The string based preferences are always marked as modified when undefined by default

PR #16356, opened 2025-10-01
fix: restore fallback default values for preferences

ndoschek:
#### What it does

Fixes GH-16339

Restore returning type-based default values when no explicit default is set, avoiding undefined default values

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


tsmaeder (commented):
I'm not sure this is the right fix: if we give a "default default" here, it's impossible to detect whether a property has a default or not. IMO, no default should be handled in the clients, as `undefined` cannot be a default value (only `null` allowed in json).

ndoschek:
Thanks @tsmaeder, good point! I have updated my changes, the handling is now on the client, which already had fallback values available in the node renderers, so we just reuse that instead.

ndoschek:
Thanks @sdirix, I could spot the problem and fixed it with an additional commit. ~~I fixed another issue with it: if a value was reset and there was a value in a more specific scope, the reset also did not work.~~
Edit: The mentioned issue was extracted into a follow-up as it was not introduced during the last release period.

I updated the testing section accordingly.

eneufeld:
LGTM, can't reproduce the issue describes in the issue anymore. Thank you

eneufeld (approved):

PR #16361, opened 2025-10-02
Retain session-bound metadata for disabled breakpoints

### LINKED ISSUE #16192: Breakpoint relocation by cdt-gdb-vscode (or GDB) causes incorrect UI updates in Theia
### Bug Description:
When using the `cdt-gdb-vscode` extension in Theia, setting a breakpoint on a non-executable source line causes GDB to automatically relocate the breakpoint to the next valid instruction (e.g. the next executable line). While this relocation is expected and correctly handled by the debug adapter (`cdt-gdb-adapter`), **Theia's UI becomes inconsistent**, continuing to associate the breakpoint with the originally requested line instead of the actual one used by GDB.

This issue **does not occur in VS Code** using the same extension.

For transparency, the original issue was first reported in the `cdt-gdb-adapter` repository and now identified to be Theia-specific: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/411

### Example
Given this C code:
```c
void MyProgram(void)
{
    long result = 0; 
    if (switch1) {
        counter1++;
    }
    // ...
}
```

If I click on the line `void MyProgram(void)` to add a breakpoint:

- GDB relocates the breakpoint to the next valid line: `long result = 0;`
- The UI correctly reflects this relocation and shows the breakpoint on `long result = 0;`

However, Theia seems to continue associating the breakpoint with the original line (`MyProgram(void)`), which leads to two issues.

### Problems Observed
1. **Disabling the breakpoint via the UI** (not deleting it, but unchecking it) suddently shows the original line (`MyProgram(void)`) as disabled breakpoint. (the first part of the screen video)
2. **Disabling the breakpoint sometimes removes it completely** instead of disabling it. This especially happens after adding a second breakpoint. (the second part of the screen video)

https://github.com/user-attachments/assets/f1202fdb-61fa-4da7-ad60-f3626947fcf4

### Steps to Reproduce:

1. Open a C project in Theia and launch a debug session using the `cdt-gdb-vscode` extension.
1. Set a breakpoint on a line that is not an executable instruction, e.g. a function declaration like `void MyProgram(void)`.
1. Observe that the breakpoint is relocated by GDB to the next valid line (e.g. `long result = 0;`) and shown correctly in the UI.
1. Now disable the breakpoint using the checkbox in the Breakpoints view.
1. Observe that:
   - The disabled marker appears on the original, non-executable line instead of the relocated line.
   - In some cases, disabling the breakpoint removes it entirely, especially if more breakpoints are present.

### Additional Information

- Theia Version: 1.64.1

colin-grant-work:
#### What it does

Fixes #16192 by avoiding the creation of a new `DebugSourceBreakpoint` instance if a given debug session already has an instance corresponding to a given breakpoint ID. This ensures that data about session-bound relocations are retained when a breakpoint is disabled, as it retains a reference to its last relocated position.

#### How to test

1. Open the project [linked](https://github.com/user-attachments/files/21990353/hello.zip) in #16219
2. Put a breakpoint on the first line of `hello.js`
3. Start a debug session.
4. The breakpoint should be relocated to line 2, and you should hit it.
5. Disable the breakpoint.
6. The breakpoint should continue to be shown on line 2 until the session ends.
7. When the session ends, it should appear back on line 1.

> [!NOTE]
> This PR contains 2 commits. The first commit is an elaborate exploratory commit containing the changes that would be required to centralize bookkeeping for breakpoints in the BreakpointManager and implement a system closer to VSCode's system where session data is attached to entities representing (more or less) user-input breakpoints. I think that system has certain advantages, but it's a very large change, and once I'd come to understand the current record-keeping system by refactoring it, I saw a much smaller change that would address the issue at hand, so the second commit reverts most changes and adds what I had found to be the minimum necessary change. Even if it's squashed away, I'd like to keep the first commit in the PR for reference in case it seems worth refactoring the breakpoint system more radically in the future. 

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


eneufeld (commented):
so I tested the minimalist approach, it worked great.
I tested the maximalist approach but it doesn't compile. there are build issues.
I'm fine merging the maximalist approach but it should compile I would claim.

colin-grant-work:
@eneufeld, I don't necessarily want to merge as two commits - I just want to have the commit here in the PR record so that I can return to it later. I'm fine merging squashed so that the effect is just the minimal version.

PR #16364, opened 2025-10-03
Implement asynchronous tooltips

### LINKED ISSUE #16362: Support `theia.proposed.statusBarItemTooltip`, introducing tooltip factories

colin-grant-work:
#### What it does

Closes #16362 by adding asynchronous provision to the status bar and hover systems.

#### How to test

1. Implement a VSCode plugin that runs code like the following (I added it to `sample-plugins/sample-namespace/plugin-a/extension.js` and symlinked that into the `plugins` folder. Then you can trigger the code by running the 'hello from plugin a' command):
```javascript
    const myStatusItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left);
    myStatusItem.text = 'See what happens!';
    myStatusItem.tooltip2 = async (token) => {
        await new Promise(resolve => setTimeout(resolve, 5000));
        if (token.isCancellationRequested) {
            throw new vscode.CancellationError();
        }
        return 'You waited long enough!';
    }
    myStatusItem.show();
```
2. The status item should appear.
3. You should be able to hover it.
4. Initially, a `Loading...` message will appear.
5. After a few seconds, the message `'You waited long enough!'` should appear.
6. If you move away from the item, it should cancel the hover, and it should not appear or interrupt another hover after the time has elapsed.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/status-bar-message-registry-main.ts
+++ b/packages/plugin-ext/src/main/browser/status-bar-message-registry-main.ts
@@ -68,6 +68,7 @@ export class StatusBarMessageRegistryMainImpl implements StatusBarMessageRegistr
             color: color && (this.colorRegistry.getCurrentColor(color) || color),
             // In contrast to color, the backgroundColor must be a theme color. Thus, do not hand in the plain string if it cannot be resolved.
             backgroundColor: backgroundColor && (this.colorRegistry.getCurrentColor(backgroundColor)),
+            // true is used as a serializable sentinel value to indicate that the tooltip can be retrieved asynchronously
             tooltip: tooltip === true ? (token: CancellationToken) => this.proxy.$getMessage(id, token) : tooltip,
             command,
             accessibilityInformation,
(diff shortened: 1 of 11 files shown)


packages/plugin-ext/src/main/browser/status-bar-message-registry-main.ts
```diff
@@ -64,7 +68,7 @@ export class StatusBarMessageRegistryMainImpl implements StatusBarMessageRegistr
             color: color && (this.colorRegistry.getCurrentColor(color) || color),
             // In contrast to color, the backgroundColor must be a theme color. Thus, do not hand in the plain string if it cannot be resolved.
             backgroundColor: backgroundColor && (this.colorRegistry.getCurrentColor(backgroundColor)),
-            tooltip,
+            tooltip: tooltip === true ? (token: CancellationToken) => this.proxy.$getMessage(id, token) : tooltip,
```
ndoschek, line 72: nit: Maybe add a comment here, as `tooltip === true` is not obvious on the first glance to everyone.

ndoschek (approved):
Thanks @colin-grant-work!
Code looks great to me, and testing with the minimal extension worked as expected!
The code looks great, and testing with the minimal extension worked as expected 👍  
This is a really nice improvement looking at the new features the upcoming git extension version will bring 👍

PR #16366, opened 2025-10-03
Add support for context menu in debug hover

pisv:
#### What it does

VS Code supports context menu in debug hover. This PR adds similar support to Theia.

Screenshot 2025-10-03 at 13 05 30

#### How to test

```js
let abc = { foo: 1, bar: '' };
console.log(abc);
```

2. Hover over the `abc` variable to display the debug hover.

3. Right-click on a child variable (e.g. `foo`) in the debug hover and choose a command in the context menu like `Copy Value`, `Copy as Expression`, or `Add to Watch`. Verify that the commands behave the same as in the Debug Variables widget.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/debug-frontend-application-contribution.ts
+++ b/packages/debug/src/browser/debug-frontend-application-contribution.ts
@@ -61,6 +61,7 @@ import { DebugToolBar } from './view/debug-toolbar-widget';
 import { ConsoleWidget } from '@theia/console/lib/browser/console-widget';
 import { ConsoleContentWidget } from '@theia/console/lib/browser/console-content-widget';
 import { ConsoleContextMenu } from '@theia/console/lib/browser/console-contribution';
+import { DebugHoverWidget } from './editor/debug-hover-widget';
 
 export namespace DebugMenus {
     export const DEBUG = [...MAIN_MENU_BAR, '6_debug'];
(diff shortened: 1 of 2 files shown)


chroberino:
@pisv , @ndoschek, sorry for posting into this already closed topic.
I just got aware of this very nice feature via the Theia 1.66.0 release notes.

When debugging a C program (using the cdt-gdb-vscode extension) it works fine for struct types, but it has no effect on simple/plain data types. For simple datatypes (like `int`, `long`, `float`, ...) the debug tooltip only displays the current value, but it does not offer any further interaction possibilities. Is this intended? Or would it be possible to enable this context menu also for simple datatypes?
image

pisv:
> would it be possible to enable this context menu also for simple datatypes?

I don't think so. The commands in this context menu are applicable only to variables. There are no variables to speak of in this case. It looks like VS Code also does not support it.

chroberino:
> The commands in this context menu are applicable only to variables. There are no variables to speak of in this case.

Sorry for being unclear in my message. I was definitely referring to _variables_ in a C program, e.g. of type `long` or `float`. Not to the datatypes themselves. For these variables, the debug tooltip only gives me their current value, but no "add to watch" command in a context menu:
image



> It looks like VS Code also does not support it.

That's true, the screenshot above has actually been made in VS Code. However, I don't see a reason why one would not want to have this convenient "add to watch" feature also for variables of simple data types.

chroberino:
> It looks like VS Code also does not support it.

I just found out that VS Code offers "Add to watch" via the context menu (when right-clicking on such a variable):
image

pisv:
> For these variables, the debug tooltip only gives me their current value, but no "add to watch" command in a context menu

That's it. To be able to "add to watch" through the context menu, you need to right-click on a *variable*. However, in this case the debug hover only displays the *value* of the variable. That's what I meant, sorry for being unclear.

pisv:
> > It looks like VS Code also does not support it.
> 
> I just found out that VS Code offers "Add to watch" via the context menu (when right-clicking on such a variable).

Yes, but it is not the context menu of the *debug hover*, is it? If you'd like to have "Add to watch" in the context menu of the *editor*, please create a separate enhancement request. It is certainly doable.

chroberino:
> Yes, but it is not the context menu of the _debug hover_, is it?

Correct, that's in the context menu that appears when right-clicking on the varibale in the editor.

> If you'd like to have "Add to watch" in the context menu of the _editor_, please create a separate enhancement request. It is certainly doable.

Will do, thank you for the quick clarification!

PR #16374, opened 2025-10-04
Add GitHub Agent

JonasHelming:
#### What it does

- Add a GitHub Agent

#### How to test

- Task the GitHub agent to do any GitHub related actions. It supports with settings up the required MCP server

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


packages/ai-ide/src/browser/github-repo-variable-contribution.ts
```diff
@@ -0,0 +1,121 @@
...
+        if (!repository) {
+            return { variable: request.variable, value: 'No GitHub repository is currently selected or detected.' };
+        }
+
+        try {
+            const remotes = await this.git.remote(repository, { verbose: true });
+
+            // Find GitHub remote (prefer 'origin', then any GitHub remote)
+            const githubRemote = remotes.find(remote =>
+                (remote.name === 'origin' && this.isGitHubRemote(remote.fetch)) ||
```
eneufeld: this does not prefer 'origin' in any way. if in the list of remotes origin is last and you find a different remote which is a GitHub remote first, it will be taken. If you want to prefer origin, you can:
1. only search by name and take it and only if this yields no result search for the rest -> 2 searches
2. use filter instead of find, and inside filter check if there is a remote with the name 'origin'
3. use reduce to create an object result, which contains the origin remote (if it is github) in one prop and one prop with the first github remote which is not 'origin'

eneufeld:
Except the issue I described above, this looks good. worked great for me.

packages/ai-ide/src/browser/github-repo-variable-contribution.ts
JonasHelming: Great catch, should be fixed now!

sgraband:
@JonasHelming This PR introduced a breaking change for adoptors and the Theia IDE. We no longer publish the `@theia/git` extension as it is deprecated. Therefore importing from `@theia/git:1.65.0` works in the example apps, but not in applications that consume the `@theia/ai-ide` package, as the last published version of `@theia/git`  was `1.60`.

We need to fix this before the release on Thursday.

JonasHelming:
see https://github.com/eclipse-theia/theia/pull/16465

Issue #16391, opened 2025-10-07
Reloading window does not (always) refresh plugins


colin-grant-work:
### Bug Description:

Certain parts of extension lifecycle are tied to the window, so that refreshing the window should refresh those lifecycles. This appears not to be working in some cases.

### Steps to Reproduce:

1. Install an extension (e.g. markdownlint)
2. Run the `Install specific version...` command from the extension's gear menu and install a different version.
3. A 'Reload Window' button should appear on the extension. Click it.
4. When the window reloads, observe that that extension still shows 'Reload Window' - i.e. reloading the window hasn't reset that particular piece of lifecycle machinery.

### Additional Information

- Operating System: MacOS Tahoe
- Theia Version: 276431de531a3e6cc55376733c8d4ca088dbc705 (1.65.1 dev)

PR #16392, opened 2025-10-07
Use unversioned ID's for disabled extensions

colin-grant-work:
#### What it does

Closes https://github.com/eclipse-theia/theia-ide/issues/579

Currently, the details of which plugins are disabled includes the version at which the plugin is disabled. This is unexpected: all versions of a given plugin should be disabled until the user decides otherwise. This PR switches to checking only the ID and not the version.

#### How to test

1. Install an extension and disable it.
2. Use the 'Install another version' command and install a different version.
3. Reload the window (or exit the application and restart: otherwise you'll observe https://github.com/eclipse-theia/theia/issues/16391)
4. You should see the version you installed in step 2, but it should remain disabled.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/vsx-registry/src/browser/vsx-extension.tsx
+++ b/packages/vsx-registry/src/browser/vsx-extension.tsx
@@ -374,7 +374,7 @@ export class VSXExtension implements VSXExtensionData, TreeElement {
             await this.runTask(nls.localize('vsx.disabling', 'Disabling'), async () => {
                 await this.progressService.withProgress(
                     nls.localize('vsx.disabling.extensions', 'Disabling {0}...', this.id), 'extensions',
-                    () => this.pluginServer.disablePlugin(PluginIdentifiers.idAndVersionToVersionedId({ id: (id as PluginIdentifiers.UnversionedId), version: installedVersion }))
+                    () => this.pluginServer.disablePlugin(id as PluginIdentifiers.UnversionedId)
                 );
             });
         }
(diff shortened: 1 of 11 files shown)


ndoschek (commented):
Thanks a lot, @colin-grant-work!  
I tested the scenario and it works great now! 👍  
It is also visible in the backend preferences that the extension ID is stored without its version, e.g., `{"installedPlugins.disabledPlugins":"[\"ms-azuretools.vscode-containers\"]"}`

However, I noticed a small side effect: in the extension search results, the "Reload Window" button always appears next to the "Install" button now. Could you have a  look at that please?

Image

colin-grant-work:
@ndoschek, thanks, good catch. I've pushed some code that handles that case.

packages/vsx-registry/src/browser/vsx-extension.tsx
```diff
@@ -494,7 +494,9 @@ export abstract class AbstractVSXExtensionComponent<Props extends AbstractVSXExt
         const { builtin, currentTask, disabled, uninstalled, installed, deployed } = this.props.extension;
         const isFocused = (host?.model.getFocusedNode() as TreeElementNode)?.element === this.props.extension;
         const tabIndex = (!host || isFocused) ? 0 : undefined;
-        const outOfSync = installed && (deployed ? (disabled || uninstalled) : !(disabled || uninstalled));
+        const inactive = disabled || uninstalled || !installed;
+        const outOfSync = (installed && uninstalled) || (deployed ? inactive : !inactive);
+        console.log('SENTINEL FOR THE INPUT TO OUT OF SYNC', this.props.extension.id, { installed, uninstalled, deployed, disabled, inactive, outOfSync });
```
ndoschek: Thanks, works like a charm now! ✨  However, this log is polluting the console quite a bit, can we remove that again?

PR #16425, opened 2025-10-13
Only set `window.titleBarStyle` preference if different from active value

### LINKED ISSUE #16418: Title bar style preference unnecessarily written to settings.json
### Bug Description:

Follow-up on #16236. When filing that issue, I wasn't sure how the `window.titleBarStyle` setting had gotten into my `settings.json` file: I'd had the machine for quite a while, so it was possible I'd written it there myself some time ago and forgotten it. However, I've just set up a completely new machine and then found that the preference has been written into my `settings.json` file with the default value without my having done anything, and without it in fact being meaningful on my platform.

### Steps to Reproduce:

0. (On MacOS)
1. Remove your user-scoped `settings.json` file, or just remove any setting for `window.titleBarStyle` from the file.
2. Start Theia (dev) Electron.
3. Observe that a new `setting.json` file appears (or the setting reappears in your current file) and is populated with `"window.titleBarStyle": "native"` even though that setting is not relevant on MacOS, and that is the default value for Electron in any case.

### Additional Information

- Operating System: MacOS Tahoe
- Theia Version: 276431de531a3e6cc55376733c8d4ca088dbc705 (1.65 dev)

colin-grant-work:
#### What it does

Fixes #16418 by checking whether writing the preference would be necessary.

#### How to test

1. Follow steps from #16418
2. Observe that the preference is not written to the file on MacOS.

> [!WARNING]
> Arguably, this 'fix' actually reveals a bug. Since the preference should not be included on MacOS, the `inspect` call should return `undefined`. At the moment, it doesn't: the preference system in fact has added the preference to its schema, with the appropriate default value of `native`. We should likely fix that, and then ensure that the preference isn't active at all in that case, and that all consumers of the value correctly handle its absence.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/electron-browser/menu/electron-menu-contribution.ts
+++ b/packages/core/src/electron-browser/menu/electron-menu-contribution.ts
@@ -164,6 +164,10 @@ export class ElectronMenuContribution extends BrowserMenuBarContribution impleme
                 if (!currentValueActive) {
                     this.preferenceService.set('window.titleBarStyle', this.titleBarStyle, PreferenceScope.User);
                 }
+                // Enable the change flag after initialization is complete.
+                // This ensures that user-initiated changes will trigger a restart,
+                // while the synchronization change above (if any) is ignored.
+                this.titleBarStyleChangeFlag = true;
             });
         });
 
@@ -178,7 +182,6 @@ export class ElectronMenuContribution extends BrowserMenuBarContribution impleme
                     window.electronTheiaCore.setTitleBarStyle(newTitleBarStyle);
                     this.handleRequiredRestart();
                 }
-                this.titleBarStyleChangeFlag = true;
             }
         });
     }


packages/core/src/electron-browser/menu/electron-menu-contribution.ts
```diff
@@ -156,7 +156,14 @@ export class ElectronMenuContribution extends BrowserMenuBarContribution impleme
             this.titleBarStyle = style;
             this.setMenu(app);
             this.preferenceService.ready.then(() => {
-                this.preferenceService.set('window.titleBarStyle', this.titleBarStyle, PreferenceScope.User);
+                const current = this. preferenceService.inspect('window.titleBarStyle');
```
msujew:
```suggestion
                const current = this.preferenceService.inspect('window.titleBarStyle');
```

colin-grant-work:
> It seems like my changes in https://github.com/eclipse-theia/theia/pull/11588 broke with the latest preferences refactoring?

Certainly seems that way. At a casual glance, it looks like there is some checking of the `included` property, but I'm not exactly sure where in the schema building lifecycle it's taking place - apparently not everywhere it should.

msujew:
FYI my comment was just to confirm that something is broken - I can't actually test the changes, since I don't have a MacOS system available to me right now.

colin-grant-work:
> FYI my comment was just to confirm that something is broken - I can't actually test the changes, since I don't have a MacOS system available to me right now.

@msujew The preference is written regardless of platform, so it is possible to test that part on any OS. Just delete it from your settings JSON, and then check that it doesn't reappear. The only difference on MacOS is that if it does get written, then your settings JSON complains that it shouldn't be there.

ndoschek:
@msujew would you have time to take another look? TIA!

colin-grant-work:
> I tested this on Ubuntu, and unfortunately it doesn't work for me: removing the line from the user settings file causes it to reappear after a reload.

@ndoschek The behavior you're seeing is incorrect, but not because the setting is being written to your `settings.json`. On Linux, the default value for the preference is `native`, so that if `custom` is the active value, it needs to be written to the `settings.json`. This PR only aims to avoid writing the preference if it would be redundant. What is strange about what you're seeing is that you're changing the preference value from `custom` -> `undefined`, which should cause the default value `native` to take effect, but it doesn't appear that it is taking effect. Or is it? It's a bit hard to tell, since the menu items aren't visible in the recording.

ndoschek:
Ah, I see, thanks for the explanation! You're absolutely right: testing with `native` and removing the line, it no longer reappears. :tada: 

Yes, it is a bit strange, but I think in the case where the value is `custom` and I remove the line, it gets restored via `window.electronTheiaCore.getTitleBarStyleAtStartup()`. We already receive a style there, and I assume it's the last one that was set?

Regarding the reset case: I'll be working on ticket #16645 this month, and I'll make sure to address this case then!

But I noticed one more thing:
With this change, when there is no entry for the title bar in the settings.json and I switch back to custom in the settings UI, the reload dialog is no longer shown immediately. The titleBarStyleChangeFlag remains false in this first case, and only if I switch back to native and then to custom again does the reload prompt appear.

<details>
<summary>Remove line and switch to custom</summary>

https://github.com/user-attachments/assets/75833c26-cdc1-422f-848b-e82e5a25c89b
</details>

And if I remove the flag entirely, I end up in an infinite reload loop 😅
I can take another look tomorrow and see if I can spot the problem, since it's not really testable on your end.

ndoschek:
Hi again @colin-grant-work!
I rebased this branch onto current master and added a small fix: the first switch of the title bar setting wasn't triggering a restart, only the second one did. Took a while to get back to this, but the upside is that the reset behavior now works thanks to your PR that was merged in the meantime :smile: (#16846).
Colin, could you quickly test on your end as well if this still behaves as you intended? TIA

colin-grant-work:
@ndoschek, thanks for the fix and for the testing. On my side, everything still looks good, though that's a limited test on MacOS ;-), but I indeed don't see the preference in my `settings.json`.

PR #16428, opened 2025-10-13
Suggest terminal command

Fangoling:
#### What it does

This PR adds a new tool function that clears the current terminal input line and inserts a command into the buffer of the last-used terminal widget without executing the command.

If there is no last-used terminal widget, the function creates a new terminal widget with the URI of the current workspace, then inserts the command there.

Contributed under the supervision of @JonasHelming  as part of the TUM Bachelor Thesis project *"Enhancing Terminal Usability in Modern IDEs through AI-Assisted Interaction"*.

#### How to test

1. Open the terminal in Theia.
2. Open the AI Chat view.
3. Prompt the AI agent with a query that would lead to a suggestion of a terminal command, using the ~suggestTerminalCommand tool function.

Example query:
```
@Universal What is the content of my current directory? ~suggestTerminalCommand
```

The suggested command should be inserted into the buffer of the current terminal without executing it.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-ide/src/browser/ai-terminal-functions.ts
+++ b/packages/ai-ide/src/browser/ai-terminal-functions.ts
@@ -75,7 +75,7 @@ export class SuggestTerminalCommand implements ToolProvider {
                 } else {
                     this.terminalService.open(activeTerminal, { mode: 'activate' });
                 }
-                var command: string;
+                let command: string;
                 try {
                     const { command: parsedCommand } = JSON.parse(args);
                     command = parsedCommand;


packages/ai-ide/src/browser/ai-terminal-functions.ts
```diff
@@ -0,0 +1,66 @@
...
+            name: SuggestTerminalCommand.ID,
+            description: `Proposes executing a command in the terminal. This tool will automatically write the command into the terminal buffer.\n
+            Execution of the command is up to the user.`,
+            parameters: {
+                type: 'object',
+                properties: {
+                    command: {
+                        type: 'string',
+                        description: `The content of the command to write to the terminal buffer.\n
+                        ALWAYS provide the COMPLETE intended content of the command, without any truncation or omissions.\n
```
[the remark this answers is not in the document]
Fangoling: I will keep the line breaks for consistency, since the description of the tool function in packages/ai-ide/src/browser/file-changeset-functions.ts also uses manual line breaks.

packages/ai-ide/src/browser/ai-terminal-functions.ts
```diff
@@ -0,0 +1,66 @@
+// *****************************************************************************
+// Copyright (C) 2024 EclipseSource GmbH.
```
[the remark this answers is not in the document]
Fangoling: fixed in [this commit](https://github.com/eclipse-theia/theia/pull/16428/commits/b13b0177407fb4b1d1971deb30d1ad5ccbda6d07)

packages/ai-ide/src/browser/ai-terminal-functions.ts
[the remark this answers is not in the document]
Fangoling: fixed in [this commit](https://github.com/eclipse-theia/theia/pull/16428/commits/ca95c9ec0503f5908b705c58c222d1317b02fefb)

packages/ai-ide/src/browser/ai-terminal-functions.ts
```diff
@@ -0,0 +1,66 @@
...
+            handler: async (args: string, ctx: MutableChatRequestModel): Promise<string> => {
+                if (ctx?.response?.cancellationToken?.isCancellationRequested) {
+                    return JSON.stringify({ error: 'Operation cancelled by user' });
+                }
+                if (!this.terminalService.currentTerminal) {
+                    return JSON.stringify({ error: 'No open terminal. The user has to have an open terminal for this function to work.' });
+                }
+                const { command } = JSON.parse(args);
+                // Clear current input line, then insert the command
+                this.terminalService.currentTerminal?.sendText('\x01\x0b');
```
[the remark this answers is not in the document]
Fangoling: fixed in [this commit](https://github.com/eclipse-theia/theia/pull/16428/commits/5eb059f18421d7425054db1ee591068c4b3419f5)

packages/ai-ide/src/browser/ai-terminal-functions.ts
```diff
@@ -0,0 +1,67 @@
...
+            handler: async (args: string, ctx: MutableChatRequestModel): Promise<string> => {
+                if (ctx?.response?.cancellationToken?.isCancellationRequested) {
+                    return JSON.stringify({ error: 'Operation cancelled by user' });
+                }
+                if (!this.terminalService.currentTerminal) {
+                    return JSON.stringify({ error: 'No open terminal. The user has to have an open terminal for this function to work.' });
+                }
+                const { command } = JSON.parse(args);
+                // Clear the current input line by sending Ctrl+A (move to start) and Ctrl+K (delete to end)
+                this.terminalService.currentTerminal?.sendText('\x01\x0b');
```
JonasHelming: Doesn't CRTL+U do the job?

JonasHelming: I wonder whether we should move this to TerminalWidgetImpl and expose a new function that is then called here.
@sdirix WDYT, the goal here is to clear the terminal buffer before inserting a new command.

JonasHelming: This only works if a terminal is focussed. Is this intened? We could consider to also use lastUsedTerminal, although as we clean the buffer, this might be unexpected. Maybe the safest way would be to even check whether there is a command there and/or open a new terminal?
I am fine with the current state for the PR, but this should be revaluated from an UX POV.

Fangoling: CTRL+U only deletes from the cursor until the start of the line.

So if the cursor is in the middle of the current line, it would only delete half of the content.

Fangoling: I tried this out when implementing this PR. The problem was that when creating a new buffer via the terminal service, I did not (yet) find a way to create a terminal in the current project path (Only an empty terminal buffer). But this is possible, and this makes sense.

For the current implementation, the function responds with an error message that the terminal should be open for this to work.

Fangoling:
> This works well, great job! Some comments above.
> 
> While testing, I think I discovered a related bug: If you use "@universal do something ~suggestTerminalCommand" and then "Do something else (which includes another terminal access"
> 
> The agent silently stops. It seems that it does not have access to the function any longer, which could make sense, however, it silently fails.
> 
> if I explicitly include ~suggestTerminalCommand in the second message, it works.
> 
> This is likely a generic issue not an issue with the new function, but could you report this one please

What would be the intended behaviour (instead of the current one)? 
Should the agent tell the user that it can only execute commands when the functions is explicitly mentioned in the call or should the agent always be able to execute the function once it was mentioned in the chat?

packages/ai-ide/src/browser/ai-terminal-functions.ts
Fangoling: I implemented the terminal creation now with the context (URI) of the current workspace.

packages/ai-ide/src/browser/ai-terminal-functions.ts
```diff
@@ -0,0 +1,95 @@
...
+                required: ['command']
+            },
+            handler: async (args: string, ctx: MutableChatRequestModel): Promise<string> => {
+                if (ctx?.response?.cancellationToken?.isCancellationRequested) {
+                    return JSON.stringify({ error: 'Operation cancelled by user' });
+                }
+                // Ensure that there is a workspace
+                const root = this.workspaceService.tryGetRoots()[0];
+                if (!root) {
+                    return 'Error executing tool \'suggestTerminalCommand\': No workspace is currently open.';
```
sdirix: All errors should also use the JSON.stringify object format like the previous one `return JSON.stringify({ error: 'Operation cancelled by user' });`

packages/ai-ide/src/browser/ai-terminal-functions.ts
```diff
@@ -0,0 +1,95 @@
...
+                let activeTerminal: TerminalWidget | undefined = this.terminalService.lastUsedTerminal;
+                if (!activeTerminal) {
+                    try {
+                        activeTerminal = await this.terminalService.newTerminal({
+                            cwd: root.resource.toString()
+                        });
+                        this.terminalService.open(activeTerminal, { mode: 'activate' });
+                        await activeTerminal.start();
+                        // Wait until the terminal prompt is emitted
+                        await waitForEvent(activeTerminal.onOutput, 1000);
```
sdirix: Minor: Personally I would use a longer timeout, e.g. 3-5 seconds to be on the safe side.

packages/ai-ide/src/browser/ai-terminal-functions.ts
```diff
@@ -0,0 +1,95 @@
...
+                    }
+                },
+                required: ['command']
+            },
+            handler: async (args: string, ctx: MutableChatRequestModel): Promise<string> => {
+                if (ctx?.response?.cancellationToken?.isCancellationRequested) {
+                    return JSON.stringify({ error: 'Operation cancelled by user' });
+                }
+                // Ensure that there is a workspace
+                const root = this.workspaceService.tryGetRoots()[0];
```
sdirix: If there is no workspace open we could open a terminal at user home. This is what the current widget does. Just do not hand over a cwd in that case.

packages/ai-ide/src/browser/ai-terminal-functions.ts
```diff
@@ -0,0 +1,95 @@
...
+                        });
+                        this.terminalService.open(activeTerminal, { mode: 'activate' });
+                        await activeTerminal.start();
+                        // Wait until the terminal prompt is emitted
+                        await waitForEvent(activeTerminal.onOutput, 1000);
+                    } catch (error) {
+                        return `Error executing tool 'suggestTerminalCommand': ${error}`;
+                    }
+                } else if (this.applicationShell.bottomPanel.isHidden) {
+                    this.applicationShell.bottomPanel.show();
```
sdirix: This will only make sure that the bottom panel is opened:
 - If the terminal is not in the bottom panel, then this is meaningless
 - If the terminal is not the focused view in the bottom panel, it will still not be the active one afterwards

You should execute the same as in the other branch, i.e.

```
this.terminalService.open(activeTerminal, { mode: 'activate' });
```

This will find and properly show the terminal, no matter where it is

packages/ai-ide/src/browser/ai-terminal-functions.ts
```diff
@@ -0,0 +1,95 @@
...
+                }
+                const { command } = JSON.parse(args);
```
sdirix: To be safe I would verify the LLM output

packages/ai-ide/src/browser/ai-terminal-functions.ts
```diff
@@ -0,0 +1,95 @@
...
+                if (ctx?.response?.cancellationToken?.isCancellationRequested) {
+                    return JSON.stringify({ error: 'Operation cancelled by user' });
+                }
+                // Ensure that there is a workspace
+                const root = this.workspaceService.tryGetRoots()[0];
+                if (!root) {
+                    return 'Error executing tool \'suggestTerminalCommand\': No workspace is currently open.';
+                }
+                let activeTerminal: TerminalWidget | undefined = this.terminalService.lastUsedTerminal;
+                if (!activeTerminal) {
```
sdirix: It should not happen but just to be safe I would also check whether the activeTerminal is disposed

packages/ai-ide/src/browser/ai-terminal-functions.ts
```diff
@@ -0,0 +1,105 @@
...
+                        await activeTerminal.start();
+                        // Wait until the terminal prompt is emitted
+                        await waitForEvent(activeTerminal.onOutput, 3000);
+                    } catch (error) {
+                        return JSON.stringify({ error: `Error executing tool 'suggestTerminalCommand': ${error}` });
+                    }
+                } else {
+                    this.terminalService.open(activeTerminal, { mode: 'activate' });
+                }
+                var command: string;
```
sdirix: We do not use `var`s in Theia, which is why the linter complains

sdirix (changes requested):
Works great for me and code generally looks good! Please make sure that the lint check completes successfully though!

packages/ai-ide/src/browser/ai-terminal-functions.ts
Fangoling: My bad, fixed it. 😃

packages/ai-ide/src/browser/ai-terminal-functions.ts
sdirix: I think it's fine for now. If we need something like this again we can expose it there

PR #16429, opened 2025-10-13
Fix #16412: if a resource contains both staged and unstaged changes, selection in the SCM panel was not reliable

### LINKED ISSUE #16412: Change node selection does not work if a file contains both staged and upstaged changes
### Bug Description:

In the Source Control panel with the GIT SCM Provider there can be a file listed both in the Staged Changes and the (unstaged) Changes group. If I click on the file in the unstaged change group, the correct Diff editor is opened, but the selection then switches to the file in the Staged Group.

### Steps to Reproduce:

**Test Setup Instructions:**  
1. Clone a repo with at least one file that can be staged and modified again (let's call it `file.txt`.  
2. Ensure Git is enabled in the workspace.  
3. Switch to the `Source Control: Git` panel.

**Test Cases / Test Steps:**  
1. Edit `file.txt`, save, and stage the changed `file.txt`.  Verify that `file.txt` appears in the `Staged Changes` group.
2. Edit `file.txt` again. Verify that `file.txt` appears both in the `Changes` and `Staged Changes` groups.
3. In the SCM panel, click the *Unstaged* `file.txt`.  

**Observed Behavior**
The diff editor (index vs working tree) is opened. (correct!)
The *staged* `file.txt` is selected in the SCM panel. (incorrect!)

**Expected Behavior**
The diff editor (index vs working tree) is opened.
The *unstaged* `file.txt` is selected in the SCM panel.

### Additional Information

- Operating System: MacOS 26, M1 Max (aarch64)
- Theia Version: 1.65.0

### Analysis

The bug happens because of the code in `scm-tree-widget.tsx`, l. 363ff.

```typescript
    selectNodeByUri(uri: URI): void {
       for (const group of this.model.groups) {
            const sourceUri = new URI(uri.path.toString());
            const id = `${group.id}:${sourceUri.toString()}`;
            const node = this.model.getNode(id);
            if (SelectableTreeNode.is(node)) {
                this.model.selectNode(node);
                return;
            }
     [...]
```

Basically, this just takes the path of the opened file/diff of the editor and checks every `ScmGroup` for a node with that path. 

Therefore, we somehow need a way to find the correct ScmGroup for the given URI, instead of taking the first one that contains the path.

The URI is actually built in `git-scm-provider.ts` in the method `getUriToOpen(change: GitFileChange): URI`.
There, the query part of the URI is built, based on the state of the `GitFileChange`. So we could map the query part to the ScmGroup in theory. But since the `ScmTreeWidet`/`ScmTreeModel` does not know about Git specifics, there is currently no way to do this.

Basically, the fix is to enhance the interface `ScmProvider` and provide a method `getScmGroupForUri()`, and implement this method in the `GitScmProvider` (and all other `ScmProvider` implementations -> breaking change.

Then, implement a corresponding method in `ScmTreeModel` that delegates to the internal `provider`.

xpomul:
#### What it does

Adds the necessary code to resolve the URI back to the ScmGroup.

Missing: handling of PluginScmProvider

#### How to test

Test Setup Instructions:

- Clone a repo with at least one file that can be staged and modified again (let's call it file.txt.
- Ensure Git is enabled in the workspace.
- Switch to the Source Control: Git panel.

Test Cases / Test Steps:

- Edit file.txt, save, and stage the changed file.txt. Verify that file.txt appears in the Staged Changes group.
- Edit file.txt again. Verify that file.txt appears both in the Changes and Staged Changes groups.
- In the SCM panel, click the Unstaged file.txt.
- The diff editor (index vs working tree) is opened.
- The unstaged file.txt is selected in the SCM panel.

#### Breaking changes

(added `getScmGroupIdForUri()` to ScmProvider interface, but the method is optional, so not actually breaking...)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/git/src/browser/git-scm-provider.ts
+++ b/packages/git/src/browser/git-scm-provider.ts
@@ -144,6 +144,21 @@ export class GitScmProvider implements ScmProvider {
     get groups(): ScmResourceGroup[] {
         return this.state.groups;
     }
+
+    /**
+     * Find the group that contains the given URI.
+     * @param uri The URI to find.
+     * @returns The resource group that contains the URI, or undefined if not found.
+     */
+    getScmGroupIdForUri(uri: URI): string | undefined {
+        if (uri.query === 'HEAD') {
+            return 'workingTree';
+        } else if (uri.query === 'index') {
+            return 'index';
+        }
+        return undefined;
+    }
+
     get stagedChanges(): GitFileChange[] {
         return this.state.stagedChanges;
     }
--- a/packages/scm/src/browser/scm-provider.ts
+++ b/packages/scm/src/browser/scm-provider.ts
@@ -1,4 +1,4 @@
-// *****************************************************************************
+// ***************************************************************************//
 // Copyright (C) 2019 Red Hat, Inc. and others.
 //
 // This program and the accompanying materials are made available under the
@@ -36,6 +36,13 @@ export interface ScmProvider extends Disposable {
     readonly onDidChangeCommitTemplate: Event<string>;
 
     readonly amendSupport?: ScmAmendSupport;
+
+    /**
+     * Returns the SCM resource group that contains the given URI, or undefined if not found.
+     * @param uri The URI to search for
+     * @returns The SCM resource group containing the URI, or undefined if not found
+     */
+    readonly getScmGroupIdForUri?: (uri: URI) => string | undefined;
 }
 
 export const ScmResourceGroup = Symbol('ScmResourceGroup');
--- a/packages/scm/src/browser/scm-tree-model.ts
+++ b/packages/scm/src/browser/scm-tree-model.ts
@@ -79,7 +79,6 @@ export namespace ScmFileChangeNode {
 
 @injectable()
 export abstract class ScmTreeModel extends TreeModelImpl {
-
     private _languageId: string | undefined;
 
     protected provider: ScmProvider | undefined;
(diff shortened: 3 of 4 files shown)


ndoschek:
Hi @xpomul,  
Sorry for the delay on this. I checked your state, and as @msujew mentioned in the ticket, the changes in packages/git should be reverted.  
Nevertheless, I think we can use the details of the passed URI argument to determine which group it should belong to. The `query` part indicates, for example, whether it has already been staged or not.  
I did a quick test with the attached snippet (which is just a POC :sweat_smile:), but only for the specific scenario you mentioned, since there are other groups as well (e.g., `merge`). Would be great if you could also test the other cases, like files with merge conflicts, to make sure the solution holds up across all scenarios.

image

colin-grant-work:
Looking at VSCode, I think there's a difference in behavior here. In VSCode, if you have diffs open of the same file from two states (staged and unstaged) and you switch focus between the diffs, the SCM tree does _not_ change selection, even if the currently selected item is not related to those two diffs. On the other hand, if you have the original file open and switch to that, VSCode will switch selection to the last item in the tree corresponding to that file, if the selection was on something that is not related to that file. I think that this is because VSCode [does not](https://github.com/microsoft/vscode/blob/4df8d67121ffe247d7920ecdae462acc6e4812a1/src/vs/workbench/contrib/scm/browser/scmViewPane.ts#L2520-L2565) coerce the URI to a `file` scheme URI, as we do [here](https://github.com/eclipse-theia/theia/blob/63fba369ec2341c800fc9ec31dc46c6a53c6a12c/packages/scm/src/browser/scm-tree-widget.tsx#L365). That means that only URI's that match the _source_ scheme (presumably mostly `file`) change the SCM tree's selection, so the `git` scheme URI's on diffs don't produce any change, so opening a diff with doesn't shift the SCM tree selection.

> Nevertheless, I think we can use the details of the passed URI argument to determine which group it should belong to. The query part indicates, for example, whether it has already been staged or not.

I don't think this is a safe procedure, because both the URI's used for diffs and the groups in the tree themselves are generated by extensions, so we can't assume anything about how the properties of a given URI do or don't map onto group affiliation.

ndoschek:
Hi @xpomul, I opened #17158 which aims to fix this issue (among other diff editor improvements). I think this draft can be closed, thanks for the initial investigation!

Issue #16314, opened 2025-09-24
Add provenance attestation to Theia's regular NPM releases


ndoschek:
### Feature Description:

Our weekly next release builds already include provenance attestation via the `NPM_CONFIG_PROVENANCE: "true"` flag in the `publish-next.yml` workflow.

See also @theia/core package for an example of published artifacts with existing and missing provenance statements: https://www.npmjs.com/package/@theia/core?activeTab=versions

However, regular releases do not yet use this feature yet. We should extend provenance attestation to regular releases to ensure consistency and enhance supply chain security.

Relevant configuration for regular releases already exists:
https://github.com/eclipse-theia/theia/blob/master/.github/workflows/publish-release.yml
I will check with the release team on whether this setup can be reused as-is or if any blockers exist.

Issue #16405, opened 2025-10-08
1.65.2 endgame plan


ndoschek:
### Feature Description:

This issue collects fixes to be backported for the 1.65.2 patch release.
Check off the item when the fix has been ported to the release/1.65.x branch.

- [ ] #16398

ndoschek:
https://github.com/eclipse-theia/theia/releases/tag/v1.65.2

PR #16433, opened 2025-10-14
ci: update publish-release workflow for release automation

ndoschek:
#### What it does

Resolves GH-16314

- add is_patch_to_previous input to support patching previous release branches, with input validation
- clarify release_type input documentation
- fetch full git history to support accurate versioning
- split publish steps: separate steps for latest and previous patch publishing
- submit a pull request with the package updates using the eclipse-theia-bot account

**_Some notes:_**
- I saw the initial attempts to this workflow (https://github.com/eclipse-theia/theia/pull/13399/)
- As we have a successful workflow to publish next versions, I would like to achieve that also for the release publishing.
- And we could benefit from the provenance attestation when publishing via the workflow (as we already have for the next builds)

#### How to test

- I did test runs: https://github.com/eclipse-theia/theia/actions/workflows/publish-release.yml
- [The first one](https://github.com/eclipse-theia/theia/actions/runs/18495400939) tests the input validation 
- [Final test run](https://github.com/eclipse-theia/theia/actions/runs/18502369171/job/52722352117)
   - planned to fail to avoid publishing packages)  
   - Resulting package update PR: #16438

#### Follow-ups

x

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -3,6 +3,7 @@ name: Publish Release
 permissions:
   id-token: write
   contents: write
+  pull-requests: write
 
 on:
   workflow_dispatch:
@@ -81,18 +82,20 @@ jobs:
         env:
           NPM_CONFIG_PROVENANCE: "true"
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
-      
-      - name: Commit and Push Changes
+
+      - name: Submit PR for package updates
         shell: bash
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           # Configure git
           git config --local user.email "eclipse-theia-bot@eclipse.org"
           git config --local user.name "eclipse-theia-bot"
-          
+
           # Get the version from lerna.json after publishing
           VERSION=$(node -p "require('./lerna.json').version")
           echo "Lerna Version: $VERSION"
-          
+
           # First commit: Update packages/core/README.md
           if git diff --name-only | grep -q "packages/core/README.md"; then
             git add packages/core/README.md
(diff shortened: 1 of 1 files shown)


ndoschek:
Hi @jfaltermeier,
As mentioned in the description, I would like to move forward with the publish automation, ideally, I could test this workflow already publish the next path release for 1.65.x (see #16405).
Would be great if you could have a look if you can spot anything critical (especially the part with the PR vs. commit I mentioned in  the description).
Thanks in advance!

.github/workflows/publish-release.yml
```diff
@@ -44,30 +62,59 @@ jobs:
...
-        uses: peter-evans/create-pull-request@6d6857d36972b65feb161a90e484f2984215f83e # v6.0.5
-        with:
-          commiter: ${{ github.actor }} <${{ fromJson(steps.actor_user_data.outputs.data).email }}>
-          author: ${{ github.actor }} <${{ fromJson(steps.actor_user_data.outputs.data).email }}>
-          branch: bot/package-update
-          title: Package update for version ${{ env.NEXT_VERSION_NUMBER }}
-          commit-message: Package update for version ${{ env.NEXT_VERSION_NUMBER }}
-          body: Automated package update for Theia version ${{ env.NEXT_VERSION_NUMBER }}. Triggered by @${{ github.actor }}.
+          NPM_CONFIG_PROVENANCE: "true"
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
```
msujew, line 84: As of yesterday, npm changed their [auth token policy](https://github.blog/changelog/2025-09-29-strengthening-npm-security-important-changes-to-authentication-and-token-management/). Instead of rotating the tokens every 90 days (which is very annoying, since we always need to create a helpdesk issue for that), shouldn't we just enable this repo as an OIDC token provider?

jfaltermeier:
Hi @ndoschek, first of all this looks very good and I believe it should work.

Regarding the commit vs PR approach:

Using
https://api.eclipse.org/git/eca/status/91f7ee68fc1931687ce38fbaf5d703f9/ui
for eclipse-theia-bot@eclipse.org 
does not return a user account, so I think the ECA validation will fail for these commits.

I understand that we will have to force-push some more changes anyway, but the first commit `core: update re-exports for v${VERSION}` is probably fine already.
I guess it would be easy to forget about this commit and leave it unchanged. 

With a PR, the ECA validation would at least downvote if anything is wrong.

I don’t know if it would also be possible to sign an ECA for a bot user, which might be the nicest solution.

.github/workflows/publish-release.yml
ndoschek: Thanks, I totally agree! As discussed in today's dev meeting, I've created a follow up: #16434

ndoschek:
Thanks @jfaltermeier, I hadn't considered the ECA validation, but that makes total sense and as we use a similar flow for ide-snap publishing I think this is a good solution!

I've updated the workflow to create a PR again.
(In the meantime, I accidentally tested the NPM publishing step with the provenance flag, seems to work fine: https://www.npmjs.com/package/@theia/core?activeTab=versions 😉)

I also did another test run for the workflow, and the resulting PR looks good from my point of view. This should already help streamline the release process a bit, and now we have NPM provenance enabled as well.

Here’s the test run and the corresponding PR:  
- Test run: https://github.com/eclipse-theia/theia/actions/runs/18502369171/job/52722352117 (planned to fail to avoid publishing packages)  
- Resulting test PR: #16438

image

jfaltermeier (approved):
Thanks, LGTM

PR #16440, opened 2025-10-15
fix: debounce updateInMemoryResources to reduce startup times

CamilleLetavernier:
fixes issues/16406

#### What it does

- debounce `this.schemaProvider.onDidChangeSchema(() => this.updateInMemoryResources());` as this is called a lot on startup, causing a significant increase in startup time

#### How to test

- start Theia and see how long it takes to get to the workbench
  - you can get the exact time from the console, e.g. `Start of 100 plugins: 2353.7 ms [Finished 12.227 s after frontend start]`
- compare between this branch and master

My results:
On master: Start of 100 plugins: 2353.7 ms [Finished 12.227 s after frontend start]
On this PR: Start of 100 plugins: 2998.5 ms [Finished 5.866 s after frontend start]

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/browser/preferences-json-schema-contribution.ts
+++ b/packages/preferences/src/browser/preferences-json-schema-contribution.ts
@@ -23,6 +23,7 @@ import { UserStorageUri } from '@theia/userstorage/lib/browser';
 import debounce = require('@theia/core/shared/lodash.debounce');
 
 const PREFERENCE_URI_PREFIX = 'vscode://schemas/settings/';
+const DEBOUNCED_UPDATE_DELAY = 200;
 
 @injectable()
 export class PreferencesJsonSchemaContribution implements JsonSchemaContribution {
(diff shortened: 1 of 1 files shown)


packages/preferences/src/browser/preferences-json-schema-contribution.ts
```diff
@@ -38,6 +39,8 @@ export class PreferencesJsonSchemaContribution implements JsonSchemaContribution
     @inject(WorkspaceService)
     protected readonly workspaceService: WorkspaceService;
 
+    protected readonly debouncedUpdateInMemoryResources = debounce(() => this.updateInMemoryResources(), 200);
```
lucas-koehler: I'd prefer to load the "magic" debounce number from a constant at the top of the file. WDYT?

CamilleLetavernier: I think it would make sense. However, note that the pattern was copied from existing Theia Code where the delay is almost always inline (unless it is configurable). Not sure how relevant this is, though. I like constants as well :)

lucas-koehler: Let's set a good example and use a constant then :D

Issue #14324, opened 2024-10-16
[Theia AI] Allow to configure proxy settings


sdirix:
Many users sit behind a proxy and therefore need to be able to configure their proxy settings.

- The [OpenAI Api allows to configure proxy settings](https://github.com/openai/openai-node/blob/3c326626c1035d43ee184cedc9084a7d399331df/README.md?plain=1#L584-L604). 
- Theia already offers [some proxy settings](https://github.com/eclipse-theia/theia/blob/4612d3daad9c98c4165d483fb8c835851a691b48/packages/core/src/browser/core-preferences.ts#L138-L168).

[This was reported by the community](https://github.com/eclipse-theia/theia/discussions/14316).

### Feature Description:

- Leverage Theia's proxy settings to configure the Open AI Api accordingly
- If possible, do the same for the other providers, i.e. Ollama and Llamafile

somera:
I wanted use Theia IDE with Ollamy. But I wan't work, because of the proxy.

Other environment without proxy work fine.

akingscote:
I've found this too! Theia behind a corporate proxy for AI services dosent seem to work. Without the proxy, its fine.

sdirix:
@akingscote Are you interested in contributing this feature?

akingscote:
@sdirix Im happy to try haha!

PR #16453, opened 2025-10-20
[Theia AI] Allow to configure proxy settings (#14324)

fipro78:
#### What it does

This PR adds the handling of proxy configuration when trying to access a LLM. It currently only adds proxy support for Anthropic and OpenAI as I did not find information about proxy configurations for the other used APIs (e.g. Ollama or Google).

Resolves GH-14324

#### How to test

Be behind a proxy, add a proxy configuration and test if you can access the LLM. It should not work without it.

#### Follow-ups

Need to check if there is proxy support for the other LLM provider APIs.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-openai/src/node/openai-language-models-manager-impl.ts
+++ b/packages/ai-openai/src/node/openai-language-models-manager-impl.ts
@@ -27,6 +27,7 @@ export class OpenAiLanguageModelsManagerImpl implements OpenAiLanguageModelsMana
 
     protected _apiKey: string | undefined;
     protected _apiVersion: string | undefined;
+    protected _proxyUrl: string | undefined;
 
     @inject(LanguageModelRegistry)
     protected readonly languageModelRegistry: LanguageModelRegistry;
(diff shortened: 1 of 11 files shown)


packages/ai-openai/src/node/openai-language-models-manager-impl.ts
```diff
@@ -75,6 +76,28 @@ export class OpenAiLanguageModelsManagerImpl implements OpenAiLanguageModelsMana
                 }
                 return undefined;
             };
+            const proxyUrlProvider = (url: string | undefined) => {
```
eneufeld, line 79: should we do the same logic in the anthropic case?

fipro78: In the current anthropic model implementation there is no url field. So it is not possible to implement the same logic there. And the default url that is used by anthropic is a https one if I got that right from their documentation. That is why the implementation differs.

fipro78:
> Looks good. Thank you. I think we should have the same logic for the proxyUrlProvider in anthropic and openai case though. Or is there a reason to not align them?

@eneufeld 
As mentioned in the response for the comment on the code change, it is not possible to align the logic for the proxyUrlProvider as the anthropic model does not provide a url that could be used.

PR #16461, opened 2025-10-20
feat: disable claude code if API key not set

planger:
#### What it does

Add optional `status` field to `ChatAgent`. This allows for instance the `@Orchestrator` and other consumers of `#agents` to consider potential error states of chat agents to drive the decision whether or not to delegate.

This is now used by `@ClaudeCode` chat agent which returns an error message in `status` if neither the `ANTHROPIC_API_KEY` environment variable nor the `ai-features.anthropic.AnthropicApiKey` preference is set.

#### How to test

Unset the environment variable and the API Key preference and observe the output of `#chatAgents`, as used for instance by the `@Orchestrator`.
Set the API Key preference and observe the value again.
Start Theia with the environment variable set, and observe the value again.

#### Follow-ups

We may also want to check whether Claude Code is installed, and reflect this in its `status`.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-agents-variable-contribution.ts
+++ b/packages/ai-chat/src/common/chat-agents-variable-contribution.ts
@@ -65,17 +65,18 @@ export class ChatAgentsVariableContribution implements AIVariableContribution, A
         const agents = this.agents.getAgents().map(agent => ({
             id: agent.id,
             name: agent.name,
-            description: agent.description
+            description: agent.description,
+            status: agent.status
         }));
         const value = agents.map(agent => prettyPrintInMd(agent)).join('\n');
         return { variable: CHAT_AGENTS_VARIABLE, value };
     }
 }
 
-function prettyPrintInMd(agent: { id: string; name: string; description: string; }): string {
+function prettyPrintInMd(agent: { id: string; name: string; description: string; status?: string | undefined }): string {
     return `- ${agent.id}
   - *ID*: ${agent.id}
   - *Name*: ${agent.name}
-  - *Description*: ${agent.description.replace(/\n/g, ' ')}`;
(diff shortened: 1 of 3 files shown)


JonasHelming:
As we discussed offline, I am not sure, whether this work around is making things better. As an example, if you know you want to talk to ClaudeCode, you will not find it in the chat view anymore now until you configured it. Also, if you set the env variable after the fact, it won't get active.

Maybe we should rather add an optional "getStatus" or "isready" function to agents which with the can determine their readyness. This could be provided in the variable "agents" so that the orchestrator knows the agent is not ready. As a very quick fix, this information could be even dynamcally added to the description.

sdirix:
This PR might no longer be required. I tested the PR, deleted my API key and noticed that Claude Code still worked as before. (Note that for testing I merged the latest `master` locally, bringing in the SDK move update from https://github.com/eclipse-theia/theia/pull/16500).

To double check I actually debugged into the `claude-code-service-impl.ts` to make sure I was not leaking my API key from somewhere. We definitely hand over `undefined` but still receive responses. It seems that the SDK reuses my terminal authentication. Is there a way to verify that?

sgraband:
I can confirm that i also can use the claude code agent without providing an api key in Theia. It seems like the globally configured credentials are then used.

sdirix:
I created https://github.com/eclipse-theia/theia/pull/16508 to handle this topic. I suggest to close this PR

ndoschek:
Closing this in favor of the mentioned successor

sdirix:
@ndoschek I discussed this with @planger again.

The overall concept of a chat agent status would still be nice. However for Claude Code we should check whether the SDK is installed to avoid delegating to Claude Code when it's not even runnable.

It would be nice if we still could get *some* solution in before release which disables delegation to Claude Code either
 - always disabled, or
 - only disabled when the Claude Code SDK is not available on the system

PR #16465, opened 2025-10-20
Removed gtheia/git dependency

### LINKED ISSUE #16460: Replace deprecated @theia/git functionality in Github Agent
### Bug Description:

The recently added GitHub Agent for theia-ai uses code of the deprecated `@theia/git` package (was deprecated with Theia v1.61.0).
We should find another way for the scm functionality needed in this new Agent

### Steps to Reproduce:

1. Remove the deprecated @theia/git dependency (e.g. https://github.com/eclipse-theia/theia/pull/16459) and build the project.

### Additional Information

- Operating System: all
- Theia Version: >= 18afb5a1afcf58485c71f6adf56a91ec87e891f0

JonasHelming:
fixed #16460

#### What it does

Removes the theia/git dependency of the Github agent
Uses simple-git on thr backend to retrieve the current git remote

fixd #16460 

#### How to test

Ask the Github agent: What is the current repo.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -439,6 +439,7 @@
         "@theia/ai-chat-ui": "1.65.0",
         "@theia/ai-code-completion": "1.65.0",
         "@theia/ai-core": "1.65.0",
+        "@theia/ai-mcp": "1.65.0",
         "@theia/ai-mcp-server": "1.65.0",
         "@theia/core": "1.65.0",
         "@theia/file-search": "1.65.0",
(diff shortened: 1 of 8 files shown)


JonasHelming:
@ndoschek FYI

eneufeld (approved):
Thank you for fixing this.
One thing to quickly discuss and maybe provide a decision for is why we use simple-git instead of dugite. Basically the question is can we rely on the system having git installed.

JonasHelming:
@msujew Would you say we can assume that Git is also installed on Windows machines? If for whatever reason it is not and this feature does not work, the world doesn't end, but would you consider this to be standard?

planger:
I think for the Theia IDE users, we can assume that git is installed and available on path, whether it is Windows, Linux or Mac users. It'd be nice to run a probe (e.g. git --version) on first startup and show a notification if it isn't available that some features won't work if git isn't installed.

For Theia platform adopters, I guess, environments become more diverse. But then again adopters can adjust whether or not they ship with a git binary. If we use simple-git and they'd ship a git binary, they'd also need to take care of telling simple-git where a portable git is available. Seems ok to me.

So in summary, I think I'd rely on simple-git.

JonasHelming:
@eneufeld I agree with Philip, fine if I merge this?

eneufeld:
yes sure, go ahead

PR #16468, opened 2025-10-21
16372: clicking on a file in AI context does not open file

CamilleLetavernier:
#### What it does

- In the AI Chat Input Widget, highlight files that have been added to the context, but are actually invalid (e.g. due to LLM hallucination)
- This makes it less confusing to the user, when they try to open a file but nothing happens, because the file doesn't actually exist

Note: as the LLM can only read/write files that are part of the workspace, this feature will mark existing files outside of the current workspace as "Invalid" (`File does not exist in the workspace`). However, users can still click these files and open them in Theia - they are only hidden to the LLM. I'm not sure we have a consistent way to handle this in Theia; so `ContextFileValidationService` is a bit complex to handle all the different cases. I tried to make it consistent with how the Context resolves file paths at the moment.

#### How to test

- You can force the LLM to add valid/invalid files with a prompt such as (adjust based on your actual workspace location and contents):

```
@Architect Add the following files to the context. Use the exact paths:
- src/index.tsx
- /home/user/workspace/src/index.tsx
- file:///home/user/workspace/src/index.tsx
- src/index.ts
- /etc/passwd
- file:///etc/passwd
```

Here's the result (in this example, index.tsx exists; index.ts doesn't exist; and other files are located outside the workspace)

image

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-ide/src/browser/context-file-validation-service-impl.spec.ts
+++ b/packages/ai-ide/src/browser/context-file-validation-service-impl.spec.ts
@@ -21,12 +21,13 @@ FrontendApplicationConfigProvider.set({});
 
 import { expect } from 'chai';
 import { Container } from '@theia/core/shared/inversify';
-import { URI } from '@theia/core';
+import { URI, PreferenceService } from '@theia/core';
 import { FileService } from '@theia/filesystem/lib/browser/file-service';
 import { WorkspaceService } from '@theia/workspace/lib/browser/workspace-service';
 import { FileStat } from '@theia/filesystem/lib/common/files';
 import { ContextFileValidationService, FileValidationState } from '@theia/ai-chat/lib/browser/context-file-validation-service';
 import { ContextFileValidationServiceImpl } from './context-file-validation-service-impl';
+import { WorkspaceFunctionScope } from './workspace-functions';
 
 disableJSDOM();
 
@@ -35,6 +36,7 @@ describe('ContextFileValidationService', () => {
     let validationService: ContextFileValidationService;
     let mockFileService: FileService;
     let mockWorkspaceService: WorkspaceService;
+    let mockPreferenceService: PreferenceService;
 
     const workspaceRoot = new URI('file:///home/user/workspace');
 
@@ -62,7 +64,7 @@ describe('ContextFileValidationService', () => {
         disableJSDOM();
     });
 
-    beforeEach(() => {
+    beforeEach(async () => {
         container = new Container();
 
         // Mock WorkspaceService
@@ -70,18 +72,20 @@ describe('ContextFileValidationService', () => {
             tryGetRoots: () => [{
                 resource: workspaceRoot,
                 isDirectory: true
-            } as FileStat]
+            } as FileStat],
+            roots: Promise.resolve([{
+                resource: workspaceRoot,
+                isDirectory: true
+            } as FileStat])
         } as unknown as WorkspaceService;
 
         // Mock FileService
         mockFileService = {
             exists: async (uri: URI) => {
-                // Normalize the URI to resolve parent directory references
                 const normalizedUri = uri.path.normalize();
                 const normalizedUriString = uri.withPath(normalizedUri).toString();
                 const uriString = uri.toString();
 
-                // Check both the original and normalized URI strings
                 const exists = (existingFiles.has(uriString) && existingFiles.get(uriString) === true) ||
                     (existingFiles.has(normalizedUriString) && existingFiles.get(normalizedUriString) === true);
                 return exists;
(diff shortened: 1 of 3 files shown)


packages/ai-ide/src/browser/context-file-validation-service-impl.spec.ts
```diff
@@ -0,0 +1,392 @@
...
+            expect(result).to.be.false;
+        });
+
+        it('should reject file:// URI when no workspace is open', async () => {
+            const result = await validationService.validateFile('file:///home/user/file.txt');
+            expect(result).to.be.false;
+        });
+    });
+
+    describe('validateFile with multiple workspace roots', () => {
```
colin-grant-work, line 274: This is tricky: I believe that our agents are still restricted to operating in the [first workspace root](https://github.com/eclipse-theia/theia/blob/8dd3f8054344e74260eb6cbbb18dca86589ef018/packages/ai-ide/src/browser/workspace-functions.ts#L53-L59), so I'm not sure it makes sense to indicate that files in other roots are ok.

CamilleLetavernier:
@colin-grant-work Thanks for the review! I'm looking at this again.

> I think that the duties of the new ContextFileValidationService partly overlap with those of the [WorkspaceFunctionScope](https://github.com/eclipse-theia/theia/blob/8dd3f8054344e74260eb6cbbb18dca86589ef018/packages/ai-ide/src/browser/workspace-functions.ts#L38) service. I'd prefer to keep the sources of truth consolidated.

Right, there is a lot of redundancy. However, I can't add a dependency to this class, because it's part of `ai-ide`, and we need to validate in `ai-chat` / `ai-chat-ui`. My understanding is that ai-ide is for Monaco-editor-related stuff, so I could move the WorkspaceFunctionsScope to `ai-chat` and reuse it from there (it doesn't contain anything specific to Monaco).

> I also think that this comes a little late in the process. I think if the LLM tries to add a file to context or otherwise refer to a file on a path where it can't be read, we should report the tool use as a failure to the LLM

:+1: I focused on the user side of the issue, but you're right, LLM should also be informed. It probably needs to be validated in multiple places:

- when LLM adds to context (tool call)
- when user adds to context (d&d) or tries to open from context
- when we pass the context to the LLM (prompt)

> This is tricky: I believe that our agents are still restricted to operating in the [first workspace root](https://github.com/eclipse-theia/theia/blob/8dd3f8054344e74260eb6cbbb18dca86589ef018/packages/ai-ide/src/browser/workspace-functions.ts#L53-L59), so I'm not sure it makes sense to indicate that files in other roots are ok.

I've played with this a bit, and we really have some inconsistency with multi-root workspace:

- the file **can** be clicked and open, even if it's part of a secondary workspace
- the file isn't accessible to LLMs, so even though it's a valid file from the user point of view, the LLM can't actually use it
- files with the same relative path under different workspace roots cannot be distinguished, because we always write the path relative to the workspace root to which it belongs (and then resolve it again against the first ws root)

So my current solution is to mark files as "warning" when they belong to a WS Root that isn't the first one, and explaining that the LLM won't be able to use it. Marking them as "Error: file does not exist" would be misleading.

This won't completely solve the inconsistency issue, but should make it a little clearer at least.

colin-grant-work:
> Right, there is a lot of redundancy. However, I can't add a dependency to this class, because it's part of ai-ide, and we need to validate in ai-chat / ai-chat-ui. My understanding is that ai-ide is for Monaco-editor-related stuff

Hm... I think my understanding had `ai-ide` responsible for end-user functionality as we implement with an eye to Theia IDE, whereas the other packages are concerned more with infrastructure enabling the general kinds of interactions with AI's that we think people will be interested. Then adopters could omit `ai-ide` if they want to build different specific UX without losing any of the capabilities that we implement in the core/chat/ui packages. So given that, I think the `WorkspaceFunctionScope` service is in `ai-ide` because it represents choices about how our functions work, rather than a general decision about how all functions must work.

That said, I don't think it changes much about your analysis :-). It does mean that likely the `WorkspaceFunctionScope` service should be responsible for deciding whether a given addition to context by our functions in `ai-ide` is valid or not. It may also mean that your approach of considering everything that can be found in the workspace valid is correct, even if an agent using one of our functions couldn't actually get to the file, because that (dis)functionality is downstream of the base context concept.

CamilleLetavernier:
I've rebased the PR, and reworked it quite a bit, with the following changes:

- use WorkspaceFunctionScope for consistency, and reduce code duplication
- move the validation service implementation to ai-ide to avoid circular dependency
- add additional case (warning) for files that exist in a secondary workspace (not usable by AI)
- update the tool call to validate files immediately when the AI tries to add them to the context, and report errors

Results:

image
image

And if files somehow end up in context anyway (e.g. after the user explicitly added them, or files were deleted after being added):

image

Issue #16450, opened 2025-10-17
Improve localization and maintainability for html-based ui messages


ndoschek:
### Feature Description:
In several places across the codebase, such as the [AI chat welcome message provider](https://github.com/eclipse-theia/theia/blob/master/packages/ai-ide/src/browser/ide-chat-welcome-message-provider.tsx) and the [Getting Started widget content](https://github.com/eclipse-theia/theia/blob/master/packages/getting-started/src/browser/getting-started-widget.tsx) localized messages are embedded directly into HTML via `nls.localize` on individual tags. This results in partial localization and makes the code harder to read and maintain.

We should investigate and refactor toward a more generic localization approach to ensure full localization of such UI sections. As part of this, we should also check the codebase for other occurrences of user‑facing strings embedded inside HTML tags.

pisv:
@ndoschek Thank you for taking care of creating this issue! 👍

I'm investigating a simple approach that could be used in such cases. The first results look rather promising, so I have submitted PR #16470 where this approach could be reviewed and discussed in detail.

PR #16470, opened 2025-10-21
feat: Provide Markdown and LocalizedMarkdown components

pisv:
#### What it does

Provides `Markdown` and `LocalizedMarkdown` components.

This PR provides localization enhancements for `IdeChatWelcomeMessageProvider`, which was not previously fully localized due to complexities described in #16450.

#### How to test

Localization enhancements aside, everything should look and behave exactly as before regarding welcome and disabled messages for the AI Chat.

#### Follow-ups

If the approach suggested in this PR is approved, it can also be applied to `getting-started-widget.tsx` and similar cases.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


eneufeld (changes requested):
Thank you for pushing for a fix here.
I have to reject the current solution though as we can't use 'dangerouslySetInnerHTML' here. This will open an attack vector for anybody by just providing a localization file.
I think we should collect and discuss possible solutions shortly on the original ticket.
A possible solution I could think of is using markdown here. We use a markdown renderer already in theia so we could reuse it here, too.

sdirix:
I agree with @eneufeld, using `dangerouslySetInnerHTML` is essentially a hack. I like the suggested markdown approach as this still enables localizable and rich content.

pisv:
OK. Thank you for your comments, @eneufeld and @sdirix.

Just a note that `dangerouslySetInnerHTML` is already used in Theia in a few  places like

https://github.com/eclipse-theia/theia/blob/6128173c7a303f558d1b9b9374206032583f53e0/packages/vsx-registry/src/browser/vsx-extension.tsx#L647

and

https://github.com/eclipse-theia/theia/blob/6128173c7a303f558d1b9b9374206032583f53e0/packages/messages/src/browser/notification-component.tsx#L84

and some other places.

Having said that, if you think that just calling `DOMPurify.sanitize` would not be enough from security point of view in this case, I could, of course, rewrite these messages in markdown and render the markdown to HTML as an additional step, but using `dangerouslySetInnerHTML` would still be required even then, as far as I can see.

sdirix:
Hi @pisv,

You're right, we already use `innerHtml` /  `dangerouslySetInnerHtml` in some places, and I agree that the security concern is likely already solved via the purification.

I brought this topic up in today's Theia developer call, and we discussed it in more detail. We concluded that we'd still prefer the Markdown approach, but for different reasons than originally stated:

- Restricting the content to Markdown will help ensure we always produce valid, well-formed HTML
- Whenever rich content needs to be provided in VS Code, it's done using Markdown (for example, in detailed preference descriptions), which we already support in Theia as well

We also briefly discussed translating each paragraph separately but came to the conclusion that this is not flexible enough.

So overall we think that Markdown is the better fit and we recommend to use it across the codebase. I will clarify this in the developer guidelines asap.

sdirix:
I created https://github.com/eclipse-theia/theia/pull/16501 for the developer guideline

pisv:
@sdirix Thanks a lot for the clarification and guidance. I'll try to rework this PR using the recommended approach.

pisv:
I have updated this PR in accordance with the recommended approach.

Some remarks:
- Since Markdown does not support styling (CSS), each styled section needs to be localized individually. Conversely, if some styling needs to be introduced to a localized Markdown string later on, you cannot help but break the string into separately styled pieces.
- Obviously, refactoring from React code to Markdown is not as straightforward as refactoring to plain HTML. In general, I feel like the Markdown-based approach does not mix as well with React as the HTML-based approach.
- Although the current result looks and feels almost as before, the anchors for commands (i.e. those with `command:` URLs) are now missing `role=button`.

To sum up, I think that the recommended approach is good enough and has its own advantages, but it definitely seems to me less flexible (for this specific case, at least) than the HTML-based approach proposed initially in this PR.

sdirix:
I'll take a look :+1:

pisv:
See also #16578, which attempts to localize the AI banner in the getting started widget using the recommended approach for localizing rich content.

packages/ai-ide/src/browser/ide-chat-welcome-message-provider.tsx
```diff
@@ -57,30 +60,63 @@ const TheiaIdeAiLogo = ({ width = 200, height = 200, className = '' }) =>
 @injectable()
 export class IdeChatWelcomeMessageProvider implements ChatWelcomeMessageProvider {
 
+    /**
+     * @deprecated not needed by this class anymore
+     */
```
sdirix: We should remove this field if we don't need it anymore. The same for the deprecated method.

packages/ai-ide/src/browser/ide-chat-welcome-message-provider.tsx
```diff
@@ -57,30 +60,63 @@ const TheiaIdeAiLogo = ({ width = 200, height = 200, className = '' }) =>
...
-                    Attach context:  use variables, like <em>#file</em>, <em>#_f</em> (current file), <em>#selectedText</em>{' '}
-                    or click <span className="codicon codicon-add" />.
-                </p>
-                <p>
-                    Lean more in the <a target='_blank' href="https://theia-ide.org/docs/user_ai/#chat">documentation</a>.
-                </p>
-            </div>
+            <div className="theia-WelcomeMessage-Content" ref={node => {
+                node?.replaceChildren(welcomeMessageRendered.element);
+            }} />
```
sdirix: If I am not mistaken then this will be called every time the component is rerendered because ref is given a function.

packages/ai-ide/src/browser/ide-chat-welcome-message-provider.tsx
```diff
@@ -57,30 +60,63 @@ const TheiaIdeAiLogo = ({ width = 200, height = 200, className = '' }) =>
...
+        const welcomeMessage = nls.localize('theia/ai/ide/chatWelcomeMessage', `
+# Ask the Theia IDE AI
+
+To talk to a specialized agent, simply start your message with *@* followed by the agent's name: *@{0}*, *@{1}*, *@{2}*, and more.
+
+Attach context: use variables, like *#{3}*, *#{4}* (current file), *#{5}* or click {6}.
+
+Lean more in the [documentation](https://theia-ide.org/docs/user_ai/#chat).
+`, 'Coder', 'Architect', 'Universal', 'file', '_f', 'selectedText', '<span class="codicon codicon-add"></span>');
+        const welcomeMessageRendered = this.markdownRenderer.render(new MarkdownStringImpl(welcomeMessage, { supportHtml: true }));
```
sdirix: The rendered elements have a dispose function which should be called tied to the React lifecycle

sdirix:
@pisv @eneufeld please check my suggested changes for this PR

sdirix (approved):

pisv:
@sdirix Many thanks for your review and great help in improving this PR!

Your suggested changes look really good to me. They definitely provide much better integration with React. (I must say though that I'm not an expert in React and cannot comment on the implementation details.)

Just a note that the coding guide's example of using the markdown renderer should probably be updated to return `MarkdownRenderResult` instead of `HTMLElement` to properly dispose the result. (It would also be great to mention the new React components for Markdown, of course.)

Thanks! 👍

sdirix:
Good point, I will update that in a separate PR then

Issue #16442, opened 2025-10-15
[vscode] align terminalCompletionProvider proposed API updated in vscode 1.105.0


ndoschek:
### Feature Description:

The stubbed proposed API `proposed.terminalCompletionProvider` has been updated with vscode 1.105.0

- Several documentation updates
- Function signature update for
   ```ts
   export function registerTerminalCompletionProvider<T extends TerminalCompletionItem>(id: string, provider: TerminalCompletionProvider<T>, ...triggerCharacters: string[]): Disposable;
   ```
- Interface update for
  ```ts
   export interface TerminalResourceRequestConfig {
		/**
		 * Show files as completion items.
		 */
		filesRequested?: boolean;
		/**
		 * Show folders as completion items.
		 */
		foldersRequested?: boolean;
		/**
		 * A {@link GlobPattern glob pattern} that controls which files suggest should surface.
		 */
		globPattern?: GlobPattern;
		/**
		 * The cwd from which to request resources.
		 */
		cwd: Uri;
   }
   ```

PR #16475, opened 2025-10-22
[vscode] API evolution (proposed) to 1.105.0

ndoschek:
#### What it does

- [vscode] chore: align module declarations for proposed.*.d.ts
   - We used mixed declarations of the proposed plugin modules, this commit aligns them to use export consistently. As they are imported in the theia.d.ts which is the main definition of @theia/plugin, we do not want to re declare them in each proposed type file. The typedoc documentation benefits from that as well, this way the @theia/plugin types are correctly listed under the theia.d.ts entry then.
- [vscode] align terminalCompletionProvider proposed API
  - several documentation updates
  - function signature update for registerTerminalCompletionProvider (stubbed)
  - interface update for TerminalCompletionItem, TerminalResourceRequestConfig TerminalCompletionProvider

Resolves https://github.com/eclipse-theia/theia/issues/16442

#### How to test

- The affected interfaces are stubbed at this time

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin/src/theia.proposed.terminalCompletionProvider.d.ts
+++ b/packages/plugin/src/theia.proposed.terminalCompletionProvider.d.ts
@@ -18,16 +18,28 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-// code copied and modified from https://github.com/microsoft/vscode/blob/1.96.2/src/vscode-dts/vscode.proposed.terminalCompletionProvider.d.ts
+// code copied and modified from https://github.com/microsoft/vscode/blob/1.105.0/src/vscode-dts/vscode.proposed.terminalCompletionProvider.d.ts
 
 export module '@theia/plugin' {
 
     // https://github.com/microsoft/vscode/issues/226562
 
+    /**
+     * A provider that supplies terminal completion items.
+     *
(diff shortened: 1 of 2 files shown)


packages/plugin/src/theia.proposed.terminalCompletionProvider.d.ts
```diff
@@ -20,14 +20,26 @@
  *--------------------------------------------------------------------------------------------*/
 // code copied and modified from https://github.com/microsoft/vscode/blob/1.96.2/src/vscode-dts/vscode.proposed.terminalCompletionProvider.d.ts
 
-declare module '@theia/plugin' {
+export module '@theia/plugin' {
```
colin-grant-work, line 23: When we modify a file like this, it would be helpful to update the 'copied and modified from' line (though a pain to enforce!) to make it clearer what version we're currently aiming at.

ndoschek: Good catch, thanks! Yes definitely, I updated the last commit!

ndoschek:
Thanks for the review and discussion @colin-grant-work!
The main motivation for the change was to align the mixed usage of `declare module` and `export module` across the proposed API files. While both approaches work functionally, `export` avoids re-declarations of `theia.d.ts` and results in cleaner [output from typedoc](https://eclipse-theia.github.io/theia/docs/next/modules/_theia_plugin.html), as it prevents the full plugin API from being duplicated across the proposed types. It is not a strong preference, but the consistency felt like a small improvement 🙂

Issue #16441, opened 2025-10-15
[vscode] API evolution (public and proposed) and menu extension points between 1.104 and 1.105


ndoschek:
### Sources

| | Details |
|:-|:-|
| Commits compared | 1.104.0: [f220831](https://github.com/microsoft/vscode/commit/f220831ea2d946c0dcb0f3eaa480eb435a2c1260) <br> 1.105.0: [03c265b](https://github.com/microsoft/vscode/commit/03c265b1adee71ac88f833e065f7bb956b60550a) |
| Public API diff command | `git diff --name-status f220831 03c265b -- src/vscode-dts/` |
| Menu diff command | `git diff --quiet f220831 03c265b -- src/vs/workbench/services/actions/common/menusExtensionPoint.ts` |
| Built‑in extensions scan (on commit 03c265b) | `git submodule foreach "grep -irn enabledApiProposals --include=package.json \| cut -d ':' -f 1 \| xargs jq --raw-output '{filename: input_filename, enabledApiProposals: .enabledApiProposals}'" > ./vscode-builtins-enabled-proposals-1.105.0-master.txt` |

#### Public API (`vscode.d.ts`)

- several minor documentation updates
- several minor readonly modifier updates
- Add new function to `SecretStorage`: `keys(): Thenable<string[]>`
- Add new interface `AuthenticationWwwAuthenticateRequest`
- Update a few function parameters in `namespace authentication`
- Add new interface `LanguageModelChatCapabilities` (extracted from prev inline declaration)

#### Proposed API evolutions

| File name | Change | Comment |
|:-|:-|:-|
| vscode.proposed.authenticationChallenges.d.ts | update | not used by builtins, ignored for now |
| vscode.proposed.chatParticipantAdditions.d.ts | update | not used by builtins, ignored for now |
| vscode.proposed.chatParticipantPrivate.d.ts | update | not used by builtins, ignored for now |
| vscode.proposed.chatProvider.d.ts | update | not used by builtins, ignored for now |
| vscode.proposed.chatSessionsProvider.d.ts | update | not used by builtins, ignored for now |
| vscode.proposed.devDeviceId.d.ts | added | not used by builtins, ignored for now |
| vscode.proposed.languageModelCapabilities.d.ts | update | not used by builtins, ignored for now |
[...]

PR #16476, opened 2025-10-22
[vscode] API evolution (public) to 1.105.0

ndoschek:
#### What it does

Part of GH-16441

- [vscode] chore: several minor readonly modifier updates in theia.d.ts 
- [vscode] extract type LanguageModelChatCapabilities 
- [vscode] implement new keys() function in SecretStorage 
   - extend SecretStorage interface in theia.d.ts
   - implement keys() function in credential services
   - add browser api test case and KeyStoreServiceImpl test cases
- [vscode] update getSession arguments in authentication namespace 
   - add AuthenticationWwwAuthenticateRequest interface
   - update getSession function arguments in authentication namespace to allow AuthenticationWwwAuthenticateRequests

#### How to test

- For the updates of the SecretStorage interface, I added api tests
- For the updates in the authentication namespace, it should be tested if the authentication for an extension (e.g. GitLens or GitHub Authentication `vscode.github-authentication`) still works. I couldn't find an extension on openvsx that already uses the new www request api.

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/authentication-service.ts
+++ b/packages/core/src/browser/authentication-service.ts
@@ -52,12 +52,27 @@ export interface AuthenticationProviderInformation {
     label: string;
 }
 
+export interface AuthenticationWwwAuthenticateRequest {
+    readonly wwwAuthenticate: string;
+    readonly fallbackScopes?: readonly string[];
+}
+
+export function isAuthenticationWwwAuthenticateRequest(obj: unknown): obj is AuthenticationWwwAuthenticateRequest {
+    return !!(obj
+        && typeof obj === 'object'
+        && 'wwwAuthenticate' in obj
+        && (typeof obj.wwwAuthenticate === 'string'));
+}
+
 /** Should match the definition from the theia/vscode types */
 export interface AuthenticationProviderAuthenticationSessionsChangeEvent {
     readonly added: readonly AuthenticationSession[] | undefined;
     readonly removed: readonly AuthenticationSession[] | undefined;
     readonly changed: readonly AuthenticationSession[] | undefined;
 }
+
+// OAuth2 spec prohibits space in a scope, so use that to join them.
+const SCOPESLIST_SEPARATOR = ' ';
 
 export interface SessionRequest {
     disposables: Disposable[];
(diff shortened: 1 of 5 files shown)


packages/core/src/browser/authentication-service.ts
```diff
@@ -102,15 +117,21 @@ export interface AuthenticationProvider {
      * @param account The optional account that you would like to get the session for
      * @returns A promise that resolves to an array of authentication sessions.
      */
-    getSessions(scopes: string[] | undefined, account?: AuthenticationSessionAccountInformation): Thenable<ReadonlyArray<AuthenticationSession>>;
+    getSessions(
```
colin-grant-work, line 119: The doc comment should be updated here since the parameters have changed.

packages/core/src/browser/authentication-service.ts
```diff
@@ -102,15 +117,21 @@ export interface AuthenticationProvider {
...
+    ): Thenable<ReadonlyArray<AuthenticationSession>>;
 
     /**
      * Prompts a user to login.
      * @param scopes A list of scopes, permissions, that the new session should be created with.
      * @param options The options for createing the session
      * @returns A promise that resolves to an authentication session.
      */
-    createSession(scopes: string[], options: AuthenticationProviderSessionOptions): Thenable<AuthenticationSession>;
+    createSession(
```
colin-grant-work, line 130: Here, too.

packages/core/src/common/key-store.ts
```diff
@@ -23,4 +23,5 @@ export interface KeyStoreService {
     deletePassword(service: string, account: string): Promise<boolean>;
     findPassword(service: string): Promise<string | undefined>;
     findCredentials(service: string): Promise<Array<{ account: string, password: string }>>;
+    keys(service: string): Promise<string[]>;
```
colin-grant-work, line 26: It's a bit odd to me that this interface and the `CredentialsProvider` in `packages/core/src/browser/credentials-service.ts` are so similar but not quite the same. I suppose optional methods are a bigger problem across the RPC boundary than they would be on either side of the divide.

packages/plugin/src/theia.d.ts
```diff
@@ -15256,47 +15285,59 @@ export module '@theia/plugin' {
...
-         * quickpick to select which account they would like to use.
+         * Get an authentication session matching the desired scopes or request. Rejects if a provider with providerId is not
+         * registered, or if the user does not consent to sharing authentication information with the extension. If there
+         * are multiple sessions with the same scopes, the user will be shown a quickpick to select which account they would like to use.
+         *
+         * Built-in auth providers include:
+         * * 'github' - For GitHub.com
+         * * 'microsoft' For both personal & organizational Microsoft accounts
+         * * (less common) 'github-enterprise' - for alternative GitHub hostings, GHE.com, GitHub Enterprise Server
+         * * (less common) 'microsoft-sovereign-cloud' - for alternative Microsoft clouds
```
colin-grant-work: Are these built-in for our purposes?

packages/core/src/common/key-store.ts
ndoschek: You're right, that is odd indeed. I was looking at the VS Code implementation in that case, but I think in our case it makes more sense to have it aligned, so I removed optional modifier in the `CredentialsProvider` to have the two interfaces aligned.

packages/plugin/src/theia.d.ts
ndoschek: Good catch, thanks, no I think not. It was more a copy paste mistake, so I removed those bits again and kept the original listing.

Issue #16367, opened 2025-10-03
Redhat XML plugin error (on 1.65): Activating plugin 'XML' failed: Cannot set properties of undefined (setting 'file')


safisa:
### Bug Description:
Hi,
When upgraded from Theia 1.64 to 1.65, I got this error on page load (as a notification pop-up):
Activating plugin 'XML' failed: Cannot set properties of undefined (setting 'file')

And in the browser console logs:

```
2025-10-03T10:22:23.378Z root ERROR [9a27bbe6-5a5e-43eb-b7ec-7e2f7a59b703][redhat.vscode-xml]: Failed to load 'configuration' contribution. Error: Property with id 'redhat.telemetry.enabled' already exists
    at PreferenceSchemaServiceImpl.addSchema (http://localhost:3001/bundle.js:407618:23)
    at http://localhost:3001/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:8274:87
    at pushContribution (http://localhost:3001/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:8265:32)
    at PluginContributionHandler.handleContributions (http://localhost:3001/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:8274:17)
    at HostedPluginSupport.handleContributions (http://localhost:3001/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:723:41)
    at HostedPluginSupport.loadContributions (http://localhost:3001/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:1476:41)
    at HostedPluginSupport.doLoad (http://localhost:3001/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:1361:42)
    at async ProgressService.withProgress (http://localhost:3001/bundle.js:408571:20)
    at async http://localhost:3001/packages_plugin-ext_lib_hosted_browser_hosted-plugin_js.js:1317:17

2025-10-03T10:22:25.151Z plugin-manager ERROR Activating plugin 'XML' failed: TypeError: Cannot set properties of undefined (setting 'file')
    at t.getXMLSettings (/Users/safi/dev/theia/plugins/vscode.xml-redhat/extension/dist/extension.js:1:345027)
    at /Users/safi/dev/theia/plugins/vscode.xml-redhat/extension/dist/extension.js:1:302041
    at /Users/safi/dev/theia/plugins/vscode.xml-redhat/extension/dist/extension.js:1:302668
[...]

PR #16481, opened 2025-10-23
Fix vscode plugin activation failed error when it uses an already existing key in its settings

safisa:
#### What it does

Fixes: https://github.com/eclipse-theia/theia/issues/16367

#### How to test

Follow the steps in the original issue.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


safisa:
Hi @ndoschek 
Can you assign this PR to a reviewer?

safisa:
Hi @msujew 

One example I have is this case:
The Red Hat XML plugin registers the same setting ID 'redhat.telemetry.enabled' as in the Red Hat Java plugin, so it fails.

I also encounter the same situation with our Theia extension, where I had to define the same setting ID as in our vscode plugin just for grouping into categories to work in the preferences UI.

PR #16484, opened 2025-10-23
feat: add codex as an agent

### LINKED ISSUE #16607: Add support for Codex agent
### Feature Description:
Adding support for using Codex (by Open AI) as an agent within the Theia AI Chat view.

eneufeld:
#### What it does

This allows to use codex as an agent in the Theia Chat similar to the claude-code agent.

Fixes #16607 

#### How to test

Install Codex globally: `npm i -g @openai/codex-sdk` .
Start the chat and if you have an OpenAI Key defined it will be used and you will find a new `@Codex` agent in the chat.

#### Follow-ups

Maybe add support to acccess the Theia tools.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-codex/README.md
+++ b/packages/ai-codex/README.md
@@ -1,0 +1,70 @@
+<div align='center'>
+
+<br />
+
+<img src='https://raw.githubusercontent.com/eclipse-theia/theia/master/logo/theia.svg?sanitize=true' alt='theia-ext-logo' width='100px' />
+
+<h2>ECLIPSE THEIA - OPENAI CODEX INTEGRATION</h2>
+
+<hr />
+
+</div>
+
+## Description
+
+The `@theia/ai-codex` integrates OpenAI's Codex agent into the Theia platform, providing natural language coding assistance through the chat interface.
+
+## Features
+
+- Natural language coding assistance via OpenAI Codex SDK
+- Thread-based conversation management
+- Streaming responses with real-time updates
+- Token usage tracking and cost monitoring
+- Structured output support (JSON Schema, Zod)
+- Integration with Theia's AI chat infrastructure
+
+## Prerequisites
+
+- Node.js 18 or higher
+- OpenAI API key (configured via Theia preferences)
+
+## Configuration
+
+### API Key
+
+Codex requires an OpenAI API key. You can configure it in several ways (in order of priority):
+
+1. **Codex-specific key** (highest priority):
+   - Preference: `ai-features.codex.apiKey`
+   - Use this if you want a separate API key for Codex
+
+2. **Shared OpenAI key** (fallback):
+   - Preference: `ai-features.openAiOfficial.openAiApiKey`
+   - Shared with other OpenAI integrations in Theia
+
+3. **Environment variable** (final fallback):
+
+   ```bash
+   export OPENAI_API_KEY=your-api-key-here
+   ```
+
+
+### Example
+
+If you already use OpenAI features in Theia, Codex will automatically use that API key. No additional configuration needed!
+
+## Additional Information
+
+- [Theia - GitHub](https://github.com/eclipse-theia/theia)
+- [Theia - Website](https://theia-ide.org/)
+- [OpenAI Codex SDK](https://github.com/openai/codex)
+
+## License
+
+- [Eclipse Public License 2.0](http://www.eclipse.org/legal/epl-2.0/)
+- [一 (Secondary) GNU General Public License, version 2 with the GNU Classpath Exception](https://projects.eclipse.org/license/secondary-gpl-2.0-cp)
+
+## Trademark
+
+"Theia" is a trademark of the Eclipse Foundation
+<https://www.eclipse.org/theia>
--- a/packages/ai-codex/src/browser/codex-chat-agent.spec.ts
+++ b/packages/ai-codex/src/browser/codex-chat-agent.spec.ts
@@ -1,0 +1,672 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { FrontendApplicationConfigProvider } from '@theia/core/lib/browser/frontend-application-config-provider';
+import { enableJSDOM } from '@theia/core/lib/browser/test/jsdom';
+let disableJSDOM = enableJSDOM();
+FrontendApplicationConfigProvider.set({});
+
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+import { Container } from '@theia/core/shared/inversify';
+import { TokenUsageService } from '@theia/ai-core';
+import { FileService } from '@theia/filesystem/lib/browser/file-service';
+import { WorkspaceService } from '@theia/workspace/lib/browser';
+import { URI } from '@theia/core/lib/common/uri';
+import { ChangeSetFileElementFactory } from '@theia/ai-chat/lib/browser/change-set-file-element';
+import { ChatAgentLocation, MarkdownChatResponseContentImpl, ThinkingChatResponseContentImpl, ErrorChatResponseContentImpl, MutableChatRequestModel } from '@theia/ai-chat';
+import { CodexFrontendService } from './codex-frontend-service';
+import { CodexChatAgent, CODEX_CHAT_AGENT_ID, CODEX_TOOL_CALLS_KEY, CODEX_INPUT_TOKENS_KEY, CODEX_OUTPUT_TOKENS_KEY } from './codex-chat-agent';
+
+import type {
+    CommandExecutionItem,
+    FileChangeItem,
+    McpToolCallItem,
+    WebSearchItem,
+    TodoListItem,
+    AgentMessageItem,
+    ItemCompletedEvent,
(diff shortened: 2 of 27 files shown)


packages/ai-codex/src/browser/renderers/command-execution-renderer.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { ReactNode } from '@theia/core/shared/react';
+import { CommandExecutionItem } from '../../common';
+import { CodexToolCallChatResponseContent } from '../codex-tool-call-content';
+import { nls } from '@theia/core';
+
+@injectable()
+export class CommandExecutionRenderer implements ChatResponsePartRenderer<CodexToolCallChatResponseContent> {
+    canHandle(response: ChatResponseContent): number {
+        return response.kind === 'toolCall' &&
+            (response as CodexToolCallChatResponseContent).name === 'command_execution'
```
planger, line 31: We should make this typesafe with an instanceof check or a marker property (see `ClaudeCodeToolCallChatResponseContent`).

packages/ai-codex/src/browser/renderers/command-execution-renderer.tsx
```diff
@@ -0,0 +1,107 @@
...
+                <span className={`codex-exit-code ${success ? 'success' : 'error'}`}>
+                    {nls.localize('theia/ai/codex/exitCode', 'Exit code: {0}', item.exit_code)}
+                </span>
+            </div>
+            {hasOutput && isExpanded && (
+                <pre className="codex-command-output">{item.aggregated_output}</pre>
+            )}
+        </div>
+    );
+};
```
planger, line 104: It might be worth reusing the `BashToolComponent` from Claude Code (ie extracting those components and reusing them in both).

packages/ai-codex/src/browser/style/codex-tool-renderers.css
```diff
@@ -0,0 +1,120 @@
+/* *****************************************************************************
```
planger, line 1: Instead of duplicating it, it might be worth extracting and reusing the CSS from Claude Code.

packages/ai-codex/src/browser/codex-chat-agent.spec.ts
planger, line 1: Please remove this test, this is pretty useless :-)

packages/ai-codex/src/browser/codex-chat-agent.ts
```diff
@@ -0,0 +1,330 @@
...
+    /**
+     * Handle started item events (commands starting).
+     * Creates a pending tool call that will be updated when the item completes.
+     * @param event The item started event
+     * @param request The chat request being processed
+     */
+    protected handleItemStarted(event: ItemStartedEvent, request: MutableChatRequestModel): void {
+        const item = event.item;
+
+        if (item.type === 'command_execution') {
```
planger: There are several tool call types missing, see full list at https://github.com/openai/codex/blob/main/sdk/typescript/src/items.ts

packages/ai-codex/src/browser/codex-chat-agent.ts
```diff
@@ -0,0 +1,330 @@
...
+        }
+    }
+
+    /**
+     * Handle completed item events (commands, messages).
+     * Updates the pending tool call with results or creates a new one if no pending call exists.
+     * @param event The item completed event
+     * @param request The chat request being processed
+     */
+    protected handleItemCompleted(event: ItemCompletedEvent, request: MutableChatRequestModel): void {
```
planger: this can probably made a bit more generic to more easily support the remaining item types (tool calls) from https://github.com/openai/codex/blob/main/sdk/typescript/src/items.ts

packages/ai-codex/src/browser/codex-chat-agent.ts
```diff
@@ -0,0 +1,330 @@
...
+        try {
+            // Extract prompt text, removing agent address if present
+            const agentAddress = `${PromptText.AGENT_CHAR}${CODEX_CHAT_AGENT_ID}`;
+            let prompt = request.request.text.trim();
+            if (prompt.startsWith(agentAddress)) {
+                prompt = prompt.replace(agentAddress, '').trim();
+            }
+
+            // Send request to Codex service
+            const streamResult = await this.codexService.send(
```
planger, line 110: Maybe I've missed it, but this doesn't seem to support sessions yet. At least I don't see how we specify here which session we continue with this call. I assume it always creates a new session in Codex.

In Claude Code every response has a session id and it the next call you specify which session you want to resume. This way we support multiple sessions, as well as editing (branching off) past messages.

packages/ai-codex/src/browser/codex-frontend-service.spec.ts
planger, line 1: I don't think this test is very useful. It essentially just tests the preference system of Theia.

packages/ai-codex/src/browser/codex-tool-call-content.spec.ts
planger, line 1: Not very useful imho

packages/ai-codex/src/common/codex-preferences.ts
```diff
@@ -0,0 +1,50 @@
...
+                'OpenAI API key for Codex. If not set, falls back to the shared OpenAI API key (`ai-features.openAiOfficial.openAiApiKey`). Can also be set via `OPENAI_API_KEY` environment variable.'),
+            title: AI_CORE_PREFERENCES_TITLE,
+        },
+        [CODEX_SDK_PATH_PREF]: {
+            type: 'string',
+            markdownDescription: nls.localize('theia/ai/codex/sdkPath/description',
+                'Custom path to Codex SDK installation. If not specified, the system will attempt to resolve the path automatically from the global npm installation.'),
+            title: AI_CORE_PREFERENCES_TITLE,
+        },
+        [CODEX_SANDBOX_MODE_PREF]: {
```
planger: For now this can probably stay a preference, but mid-term this should really be a session-setting instead of a user setting. But we don't have a UI or generic support for modes yet, so for now we can leave it like this, or we just hard-code workspace-write for now in the agent, and update it so support the other two modes once we have mode support in Theia AI.

packages/ai-codex/src/common/codex-service.spec.ts
planger, line 1: Not sure this is useful.

packages/ai-codex/src/node/codex-service-impl.ts
```diff
@@ -0,0 +1,174 @@
...
+        } else {
+            codexPath = this.resolveCodexPath();
+        }
+
+        const sdkPath = path.join(codexPath, 'dist', 'index.js');
+
+        // Check if file exists before importing
+        if (!existsSync(sdkPath)) {
+            throw new Error(nls.localize('theia/ai/codex/installationNotFoundAt', 'Codex SDK installation not found. ' +
+                'Please install with: `npm install -g @openai/codex-sdk` ' +
```
planger: I guess we mean `npm install -g @openai/codex` here

packages/ai-codex/src/browser/codex-chat-agent.ts
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    /**
+     * Store Codex session ID in the request data.
+     * @param request The chat request
+     * @param sessionId The session ID to store
+     */
+    protected setCodexSessionId(request: MutableChatRequestModel, sessionId: string): void {
+        request.addData(CODEX_SESSION_ID_KEY, sessionId);
+    }
```
planger: I don't think we ever read this session id as we are using the `request.session.id` directly.

packages/ai-codex/src/node/.gitkeep
planger, line 1: I think we can remove this file.

planger (commented):
Thank you very much for the iteration! This is already much better.

Besides one new comment inline (regarding the unneeded tracking of the session ID) and the comments from my last review (mainly regarding potential reuse with the existing UI from Claude code), I'd just found the following issues:

Image

* There is no line-break between some messages (e.g. "Summarizing final changes")
* It'd be great to integrate the Theia changeset when Codex applies file changes
* I'd be great to have a nicer UI for the file changes that better fit the other tool call renderings (see also Claude Code)
* I think having a UI for TODO would be nice (see Claude Code)
* Showing the thinking certainly also helps

planger:
FYI: https://github.com/eclipse-theia/theia/pull/16489 could be used to replace the sandbox mode setting with actual modes.

eneufeld:
I need to check the mode. Good catch.
The problem with changeset is that the API does not offer a diff so I can only show the whole file.
A reuse of components between claude-code and codex except for the collapsable renderer is not possible at the moment as the types are to different.
I check the ignore list, too.

planger:
> The problem with changeset is that the API does not offer a diff so I can only show the whole file.

I see. I'm wondering whether we should then track them in the change list at all, if we cannot show proper diffs. Otherwise it seems quite confusing imho.

Thank you!

eneufeld:
> > The problem with changeset is that the API does not offer a diff so I can only show the whole file.
> 
> I see. I'm wondering whether we should then track them in the change list at all, if we cannot show proper diffs. Otherwise it seems quite confusing imho.
> 
> Thank you!

I would leave them for now. At least you see what files changed. There are also quite some request in the codex repo to expose more information.

packages/ai-codex/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Sandbox Mode
+
+Codex operates in a sandbox mode that controls what file system operations are allowed. Configure this via the `ai-features.codex.sandboxMode` preference:
+
+- **`read-only`**: Safest mode, only read operations allowed
+- **`workspace-write`**: Recommended mode (default), allows writes within workspace
+- **`danger-full-access`**: Unrestricted access (use with extreme caution)
+
+For most use cases, the default `workspace-write` mode provides a good balance between safety and functionality.
```
planger:
```suggestion
```

planger (approved):
Nice, works great, thank you very much!

Image

planger:
There still seems to be test failures, other than that this is good to go imho! Thank you!

eneufeld:
test failures should be fixed with https://github.com/eclipse-theia/theia/pull/16522

ndoschek:
Unfortunately, this PR introduced a test that fails on windows. A follow-up issue has been created here: https://github.com/eclipse-theia/theia/issues/16619

PR #16486, opened 2025-10-23
feat: implement chat session persistence

### LINKED ISSUE #15031: [Theia AI] Persistence for chats

sdirix:
#### What it does

Adds automatic storage and restoration of chat sessions to disk, enabling users to access their chat history across application restarts.

Key Features:
- Automatic debounced session persistence as session is in progress
- File-based storage using workspace .theia/chatSessions/ directory
- Session limit of 25 (oldest auto-deleted when exceeded)
- "Show Chats..." dialog displays both active and persisted sessions
- Full chat tree persistence including message alternatives (branch edits)
- Changeset persistence with functional apply/revert/open operations
- Graceful handling of unknown response content types

Architecture:
- Clean separation between domain data and persistence metadata
- Each model serializes itself via toSerializable()
- Extensible deserializer registries for response and changeset content
- Constructor-based restoration with minimal external state manipulation
- Session index for fast metadata lookup

Implementation includes comprehensive test coverage with unit tests for serialization/deserialization, changeset restoration, session management, and content serializers. Also provides an api sample agent demonstrating custom response content.

Resolves #15031

https://github.com/user-attachments/assets/3cdc4dd5-ed90-4dbf-9faa-3df66a32cd9c

##### Basic persistence
  - Start Theia and create a chat session with some messages
  - Close Theia completely
  - Restart Theia and use "Show Chats..." command
  - Select the archived session to restore it
##### Session limits
  - Create more than 25 chat sessions
  - Verify oldest sessions are automatically deleted when the limit is exceeded
  - Delete sessions manually, they should no longer appear to be restored
##### Changeset persistence
  - Create a chat that generates file changesets
  - Close and restart Theia
  - Restore the session and verify the changesets can still be applied, reverted, and opened
##### Custom content types
  - Test with the new CustomContent sample agent (in api-samples)
  - Verify serializable custom content is restored correctly
  - Verify non-serializable content shows fallback message after restoration
##### Complex scenarios
 - Address different agents in different messages
 - Edit messages to generate chat hierarchy
 - In each chat hierarchy branch, invoke Coder to populate changesets differently
 - Apply some parts of changesets and don't apply others
 - Start a long running agent and close application at some point

#### Follow-ups

Potential follow ups:

- Preference for history storage amount
- Alternative history storage, e.g. in Browser storage or in Workspace
  - Could also be controlled via preference
- Serialize and restore ParsedRequest
- Show stored sessions more prominent in Chat View instead of only within "Show Chats"
- Store and restore detailed agent history for History View
- Restore suggested actions
- Handle custom ClaudeCode tool calls more nicely. They are shown as regular tool calls after restoration

#### Breaking changes

Only very minor breaking changes, e.g. deleting a session is now async. Not worth to report I think.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-content-deserializer.ts
+++ b/packages/ai-chat/src/common/chat-content-deserializer.ts
@@ -46,27 +46,144 @@ import { ContributionProvider, ILogger, MaybePromise } from '@theia/core';
 
 export const ChatContentDeserializer = Symbol('ChatContentDeserializer');
 
+/**
+ * A deserializer for a specific kind of chat response content.
+ *
+ * Deserializers are responsible for reconstructing `ChatResponseContent` instances
+ * from their serialized data representations. Each deserializer handles a single
+ * content type identified by its `kind` property.
+ *
+ * @template T The type of the data object that this deserializer can process.
+ *
+ * @example
+ * ```typescript
+ * const textDeserializer: ChatContentDeserializer<TextContentData> = {
+ *     kind: 'text',
+ *     deserialize: (data) => new TextChatResponseContentImpl(data.content)
+ * };
+ * ```
+ */
 export interface ChatContentDeserializer<T = unknown> {
+    /**
+     * The unique identifier for the content type this deserializer handles.
+     * This must match the `kind` property of the serialized content data.
+     */
     readonly kind: string;
+
+    /**
+     * Deserializes the given data into a `ChatResponseContent` instance.
+     *
+     * @param data The serialized data to deserialize. The structure depends on the content kind.
+     * @returns The deserialized content, or a Promise that resolves to the deserialized content.
+     */
     deserialize(data: T): MaybePromise<ChatResponseContent>;
 }
 
+/**
+ * Contribution point for registering chat content deserializers.
+ *
+ * Implement this interface to contribute custom deserializers for application-specific
+ * or extension-specific chat response content types. Multiple contributions can be
+ * registered, and all will be collected via the contribution provider pattern.
+ *
+ * @example
+ * ```typescript
+ * @injectable()
+ * export class MyDeserializerContribution implements ChatContentDeserializerContribution {
+ *     registerDeserializers(registry: ChatContentDeserializerRegistry): void {
+ *         registry.register({
+ *             kind: 'customContent',
+ *             deserialize: (data: CustomContentData) =>
+ *                 new CustomContentImpl(data.title, data.items)
+ *         });
+ *     }
+ * }
+ *
+ * // In your module:
+ * bind(ChatContentDeserializerContribution).to(MyDeserializerContribution).inSingletonScope();
+ * ```
+ *
+ * @see {@link ChatContentDeserializerRegistry} for the registry that collects deserializers
+ * @see {@link DefaultChatContentDeserializerContribution} for built-in content type deserializers
+ */
 export interface ChatContentDeserializerContribution {
+    /**
+     * Registers one or more deserializers with the provided registry.
+     *
+     * This method is called during the registry's initialization phase (at `@postConstruct()` time).
+     *
+     * @param registry The registry to register deserializers with
+     */
     registerDeserializers(registry: ChatContentDeserializerRegistry): void;
 }
 export const ChatContentDeserializerContribution = Symbol('ChatContentDeserializerContribution');
 
+/**
+ * Registry for chat content deserializers.
+ *
+ * This registry maintains a collection of deserializers for different content types
+ * and provides methods to register new deserializers and deserialize content data.
+ *
+ * @example
+ * ```typescript
+ * // Usage in a service:
+ * @inject(ChatContentDeserializerRegistry)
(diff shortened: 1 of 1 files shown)


packages/ai-chat/src/browser/chat-session-store-impl.ts
```diff
@@ -0,0 +1,317 @@
...
+        await this.fileService.writeFile(
+            indexFile,
+            BinaryBuffer.fromString(JSON.stringify(index, undefined, 2))
+        );
+    }
+
+    private reviveData(data: unknown): SerializedChatData {
+        // At the moment we only have a single version of the data, so just cast
+        return data as SerializedChatData;
+    }
```
planger: Maybe we should already put a defense mechanism here, to help future developers find the right place to integrate a migration mechanism with a log output:

```ts
private reviveData(data: unknown): SerializedChatData {
    const parsed = data as SerializedChatData;
    
    // Defensive check for unexpected future versions
    if (parsed.version && parsed.version > CHAT_DATA_VERSION) {
        this.logger.error(
            `Session data version ${parsed.version} is newer than supported ${CHAT_DATA_VERSION}. ` +
            'Data may not load correctly.'
        );
    }
    
    return parsed;
}
```

planger (approved):
Excellent work! Tested it and it works very well. Also architecturally it is designed in a very solid and extensible way. Thanks also for adding reasonable tests!

I just have optional comments, see below / inline.

### Default Deserializer?

I'm not sure, but is it worth adding a fallback default de-/serializer for the majority of simple cases? I'm torn myself, because it cannot handle the edge cases and would potentially hide certain issues at runtime. This could only be mitigated by writing warnings, but then again the default case would pollute the warn-output, so it'd only make sense to write out those warnings if we are in development mode.

Maybe as an alternative, we could add some helper functions to make it super simple to register a custom de/serializer for the simple case (no functions, no circular dependencies) with a default base for `CustomSerializableContent` and a function to register it.

### Minors
* I think it'd be great to add typedoc to the deserializer registry with a concrete example of writing a minimal deserializer, corresponding to the one from the API examples.
* Should we put a `validateSerializable` guard in the serialization to not silently omit functions and provide better error messages if someone misses to handle it properly?
* We use a couple of `private` methods. I'd prefer to avoid those and use `protected` instead as things makes it simpler for later rebinding and customization.

sdirix:
> ### Default Deserializer?
> I'm not sure, but is it worth adding a fallback default de-/serializer for the majority of simple cases? I'm torn myself, because it cannot handle the edge cases and would potentially hide certain issues at runtime. This could only be mitigated by writing warnings, but then again the default case would pollute the warn-output, so it'd only make sense to write out those warnings if we are in development mode.
> 
> Maybe as an alternative, we could add some helper functions to make it super simple to register a custom de/serializer for the simple case (no functions, no circular dependencies) with a default base for `CustomSerializableContent` and a function to register it

There is a default serialization and deserialization fallback already implemented:
 - If a response content does not have a `toSerialized` function, we will still save its kind and its current `asString()` representation (i.e. what is used for follow up requests) as `fallbackMessage`
 - If we don't have a response content deserializer registered for the `kind`, we instantiate an `UnknownResponseContent` which renders the `fallbackMessage` in the UI with a warning and will add the `fallbackMessage` into followup requests. Therefore users can continue with their chat session, even including the previous information, without having that renderer available.

You can test this with the included sample agent which will render a response content for which there is no serialization/deserialization included.

eneufeld:
I didn't test but the code looks good. Thank you for the great feature.

JonasHelming:
@sdirix I totally love this feature, thank you for implementing it! One question: Was t considered to persist the sessions per workspace? I usually want to reuse chats within a specific context.

sdirix:
> @sdirix I totally love this feature, thank you for implementing it! One question: Was t considered to persist the sessions per workspace? I usually want to reuse chats within a specific context.

@JonasHelming Not in this initial implementation. It is listed as one of the follow up topics

JonasHelming:
@planger WDYT about this

> > @sdirix I totally love this feature, thank you for implementing it! One question: Was t considered to persist the sessions per workspace? I usually want to reuse chats within a specific context.
> 
> @JonasHelming Not in this initial implementation. It is listed as one of the follow up topics

planger:
> @planger WDYT about this
> 
> > > @sdirix I totally love this feature, thank you for implementing it! One question: Was t considered to persist the sessions per workspace? I usually want to reuse chats within a specific context.
> > 
> > 
> > @JonasHelming Not in this initial implementation. It is listed as one of the follow up topics

I agree that it would make a lot of sense to keep sessions separated by workspace folder. It is also this way in Copilot and Claude Code. So I'd plan that in with mid-to-high priority for future releases; it makes a lot of sense to keep the history separated, but it isn't blocking any use case as it is right now.

JonasHelming:
OK @sdirix Could you derive a ticket and ping nina for prio, please?

sdirix:
@ndoschek See https://github.com/eclipse-theia/theia/issues/16565

Issue #16216, opened 2025-08-26
OutputWidget and MonacoEditor Memory Leak


Hbbbit:
### Bug Description:
When repeatedly switching groups in the outputWidget, Electron DevTools Memory snapshots show that new EditWidget instances are created repeatedly and are not released.

### Steps to Reproduce:
1. Open the outputWidget
2. Take a memory snapshot
3. Repeatedly switch between sub-tabs (e.g. Git and GitHub Copilot)
4. Take another memory snapshot and compare it with step 2

### Actual Result:
The number of EditWidget instances keeps increasing and memory is not released, which eventually leads to performance degradation and possible crash.

### Expected Result:
When switching groups, old EditWidget instances should be disposed and released, and memory usage should remain stable.

### Additional Information
Evidence: Screenshots will be attached below
<img width="794" height="596" alt="Image" src="https://github.com/user-attachments/assets/01a1c329-7895-4932-bd6e-77df47242cf7" />

Other Information: None
- Operating System: Operating System: Windows 10
- Theia Version: Theia Version: 1.63.2 (the issue still exists on the latest source code)

Hbbbit:
<img width="944" height="650" alt="Image" src="https://github.com/user-attachments/assets/2d0811e7-d03c-4695-a46c-0d75d400b130" />
After calling Disposable.push to establish a parent-child relationship, when the child calls dispose, it is removed from the parent’s disposables. As a result, when the parent later executes dispose, the child’s dispose will no longer be invoked.

ndoschek:
Thank you @Hbbbit for catching that! 👀 
Would you be interested in contributing a PR for this? We'd be happy to support you with reviews or feedback along the way as well

Hbbbit:
ok

Hbbbit:
<img width="1118" height="927" alt="Image" src="https://github.com/user-attachments/assets/8630ce27-5413-49dc-b7b5-17032dd60aae" />
When I quickly open and close an editor tab, by the time execution reaches this point, the widget’s isDisposed is already true.
As a result, the subsequent onTextEditorRemove is not executed, which prevents the corresponding monacoEditor instance from being properly disposed.

Should I open a new issue for this?

PR #16487, opened 2025-10-24
fix: OutputWidget and MonacoEditor Memory Leak #16216

Hbbbit:
#### What it does

This PR fixes two memory leak issues.
#16216

**Issue 1:**

* **File:** `packages/editor/src/browser/editor-widget.ts`
* **Cause:** When repeatedly switching groups in the `outputWidget`, Electron DevTools memory snapshots show that new `EditorWidget` instances are continuously created and not released.
  This occurs because the `toDisposeOnTabbarChange` cleanup logic becomes detached from the widget’s main `toDispose`, preventing proper disposal.
* **Fix:** Wraps `toDisposeOnTabbarChange` in a new `Disposable` and adds it to the widget’s `toDispose`. This ensures that invoking `toDisposeOnTabbarChange.dispose()` does not break the linkage with the widget’s disposal chain, enabling correct cleanup and preventing the accumulation of unreleased `EditorWidget` instances.

**Issue 2:**

* **File:** `packages/plugin-ext/src/main/browser/editors-and-documents-main.ts`
* **Cause:** When quickly opening and closing an editor tab, by the time execution reaches this point, the target widget’s `isDisposed` flag is already `true`.
  As a result, the subsequent `onTextEditorRemove` is never executed, and the associated `monacoEditor` instance remains undisposed, leading to retained memory.
* **Fix:** Adds an early exit in `onTextEditorAdd`: if the target widget is already disposed, the method terminates immediately to prevent stale editor references.

#### How to test

**For Issue 1:**

1. Open the `outputWidget`.
2. Take a memory snapshot in Electron DevTools.
3. Repeatedly switch between sub-tabs (e.g. **Git** and **GitHub Copilot**).
4. Take another memory snapshot and compare it with step 2.

**Expected result:**
No unreleased `EditorWidget` instances remain and memory usage stays stable.

**For Issue 2:**

1. Quickly open and close an editor tab.
2. Take memory snapshots before and after.

**Expected result:**
No unreleased `monacoEditor` instances remain and memory usage remains stable.

#### Follow-ups

None currently. Automated regression tests could be added in future if event or disposal states become more observable.

#### Breaking changes

#### Attribution

Contributed by **Hbb**.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/editors-and-documents-main.ts
+++ b/packages/plugin-ext/src/main/browser/editors-and-documents-main.ts
@@ -299,6 +299,9 @@ class EditorAndDocumentStateComputer implements Disposable {
     }
 
     private onTextEditorAdd(widget: EditorWidget): void {
+        if (widget.isDisposed) {
+            return;
+        }
         const editor = MonacoEditor.get(widget);
         if (!editor) {
             return;


ndoschek:
Hi @Hbbbit, thanks for the PR!
I've updated the PR description to use our standard PR template, as we have some automated steps that read certain sections (like the attribution section). Just a heads-up for next time, thanks! :)

packages/editor/src/browser/editor-widget.ts
```diff
@@ -45,6 +44,9 @@ export class EditorWidget extends BaseWidget implements SaveableSource, Navigata
                 unlock(this.title);
             }
         }));
+        this.toDispose.push(Disposable.create(() => {
+            this.toDisposeOnTabbarChange.dispose();
+        }));
```
colin-grant-work, line 49: Nice catch here - the fact that this behaves differently from the existing code is a bit of a footgun of the `DisposableCollection` system.

colin-grant-work (approved):

Issue #16490, opened 2025-10-24
Claude Code SDK moved to different package


sdirix:
### Bug Description:

The Claude Code SDK is no longer available via the `@anthropic-ai/claude-code` package. Instead it must be consumed from the new `@anthropic-ai/claude-agent-sdk` package.

See [here](https://docs.claude.com/en/docs/claude-code/sdk/migration-guide) for more information

### Steps to Reproduce:

1. Install latest Claude Code
2. Try to use the Claude Code agent
3. It complains because it can't find the sdk file

### Additional Information

- Theia Version: latest `master` as well latest Theia IDE

Oghenesuvwe-dev:
Hello there @sdirix I hope you don't mind having me work on this real quick?
I've been a subscriber with Eclipse / Theia

JonasHelming:
@planger FYI

sdirix:
Hi @Oghenesuvwe-dev,

Sure, please go ahead 👍  Note however that we are under time pressure because we want to fix this before the next release. Would you have time to open a PR today or tomorrow?

planger:
@Oghenesuvwe-dev I'm sorry, but I needed this to work for a demo tomorrow and looked into it already. See https://github.com/eclipse-theia/theia/pull/16500. I didn't mean to take any work away from you and I certainly would more than welcome anyone further improving the Claude Code integration. Thank you!

PR #16500, opened 2025-10-28
fix: migrate to claude agent sdk

planger:
Fixes https://github.com/eclipse-theia/theia/issues/16490

#### What it does

Migrates to Claude Agent SDK. This involved fixing a few issues:
* Path of the SDK for lookup and messages to the user
* Adapting how we specify the system message
* Specifying where to load settings from (also to make hooks work again)
* Specifying the cwd explicitly

#### How to test

Test Claude Code in general, and in particular:
* Whether it is aware of the IDE context (like active editor)
* Whether it knows locally configured slash commands (not just completion but also executing them)
* Whether it correctly shows diffs of changes it did (tests whether hooks actually worked)

#### Follow-ups

It might be worth looking through the Agent SDK now more in detail to find out if there is anything we can simplify now, e.g. omit the hooks to track file changes, etc. But nothing that should block merging this PR for now, imho.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


planger:
Thanks for the review, one issue that I observed is that editing messages doesn't resume at the specific message but just amends to the message in the overall session. I'll create a follow-up issue for that.

PR #16506, opened 2025-10-29
Preference UI: Subscribe directly to provider changes

### LINKED ISSUE #16383: Preference update in lower scope does not trigger change event
### Bug Description:

Changing a value or resetting a preference value in a lower scope (e.g. User) does not emit a preference change event if the preference is also defined in a more specific scope (e.g. Workspace). As a result, the UI is not updated accordingly, although the settings file reflects the change. Switching scopes manually causes the UI to update.

The issue lies in the preference change reconciliation logic, which does not detect the change as an effective change due to the preference still being defined in a higher scope.

See
https://github.com/eclipse-theia/theia/blob/29757d9a028266af922090aff909b88ad9c4a9cb/packages/core/src/common/preferences/preference-service.ts#L360-L366

Remark: On a first attempt, the according unit tests showed that a change in this behaviour might cause unnecessary change events we want to avoid.

- Follow up of https://github.com/eclipse-theia/theia/pull/16356
- Note: The behaviour existed already before the backend preference refactoring of #16017

### Steps to Reproduce:

1. Set a preference value in both User and Workspace scopes.
2. Reset the preference in the User scope.
3. Observe that the UI does not reflect the change, although the settings file is updated. On switching the scopes the UI is updated correctly.

### Expected behaviour

Resetting a preference in a lower scope should emit a change event and update the UI accordingly.

### Additional Information

- Operating System: all
- Theia Version: < current

colin-grant-work:
#### What it does

Fixes #16383 by having the preference UI subscribe directly to change events from `PreferenceProviders`, rather than to the filtered events through the `PreferenceService`

#### How to test

1. Set a preference value in both User and Workspace scopes.
2. Reset the preference in the User scope.
3. Observe that the UI _does_ reflect the change immediately.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/preferences/src/browser/views/preference-editor-widget.ts
+++ b/packages/preferences/src/browser/views/preference-editor-widget.ts
@@ -25,7 +25,7 @@ import {
     TopDownTreeIterator,
     ExpandableTreeNode,
 } from '@theia/core/lib/browser';
-import { PreferenceChanges, PreferenceSchemaService, PreferenceService, unreachable } from '@theia/core/lib/common';
+import { Disposable, DisposableCollection, PreferenceProviderDataChanges, PreferenceProviderProvider, PreferenceSchemaService, PreferenceService, unreachable } from '@theia/core';
 import { BaseWidget, DEFAULT_SCROLL_OPTIONS } from '@theia/core/lib/browser/widgets/widget';
 import { PreferenceTreeModel, PreferenceFilterChangeEvent, PreferenceFilterChangeSource } from '../preference-tree-model';
 import { PreferenceNodeRendererFactory, GeneralPreferenceNodeRenderer } from './components/preference-node-renderer';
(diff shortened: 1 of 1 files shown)


safisa:
Hi,

- What about changes done in the file system (on the user settings json file)?
- Check also the issue in this ticket https://github.com/eclipse-theia/theia/issues/16255 ?

colin-grant-work:
@safisa, 

- I'm not sure exactly what you're concerned about regarding changes done on the file system, but if the question is whether this will catch changes made to the settings file for a particular scope, then yes, this does ensure that those changes are picked up by the UI immediately. The providers for specific scopes watch those files and fire preference changes in response to file changes; this change ensures that those events fired by providers reach the UI and aren't lost in the filtering the `PreferenceService` does regarding scope specificity.
- #16255 is not in scope for this particular fix - here the goal is to ensure that the preference UI is updated regardless of the presence of a setting in a narrower scope.

ndoschek (approved):
Thanks @colin-grant-work, changes look great and it works as expected! 🎉 
I tested also with a single and multi root workspace and reset works in any case as expected and shows the updated values immediately!

PR #16507, opened 2025-10-29
"Split Editor" functionality for Notebooks

dhuebner:
#### What it does

Adds a possibility to split a notebook editor using "Split Editor" and other split commands.
To make it possible a new `SplitEditorContribution` was introduced and implemented for `EditorWidget` and `NotebookEditorWidget`.

#### How to test

Open a Notebook. Open command palette and execute "Split Editor" command. A new Notebook editor widget should appear next to existing one. Same should still work with plain text editor. 

#### Follow-ups

Need to check/fix cell selection behavior in Notebook when more than one editor is open for the same URI.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/notebook/src/browser/view-model/notebook-model.ts
+++ b/packages/notebook/src/browser/view-model/notebook-model.ts
@@ -266,9 +266,6 @@ export class NotebookModel implements Saveable, Disposable {
             cell.onDidChangeOutputs(() => {
                 this.dirty = true;
             });
-            // cell.onDidRequestCellEditChange(() => {
-            //     this.onContentChangedEmitter.fire();
-            // });
         }
     }
 
--- a/packages/notebook/src/browser/view-model/notebook-view-model.ts
+++ b/packages/notebook/src/browser/view-model/notebook-view-model.ts
@@ -43,7 +43,7 @@ export class NotebookViewModel implements Disposable {
     }
 
     // Cell handle to CellViewModel mapping
-    cellViewModels: Map<number, CellViewModel> = new Map();
+    readonly cellViewModels: Map<number, CellViewModel> = new Map();
 
     initDataModel(model: NotebookModel): void {
         model.onDidAddOrRemoveCell(e => {


jonah-iden:
what would be challenging woth changing the location of the selected would be: 
- The actions in the notebook-action-contribution. We would probably have to change the default args of all those commands.
- The notebook-cell-editor since giving it access to the notebook-editor-widget creates a circular dependency
The rest would actually be ok to do
- The aspect that when executing edits on the model it creatres selected cell changes but they should only happen on the currently focused editor

dhuebner:
@jonah-iden 
Thanks for your investigation. I was already in doubt that splitting feature will "just work".
Would it work if you put your additional changes on top of this branch?

jonah-iden:
yeah i will do that

jonah-iden:
Should be ready to review.
it might be worth looking into if we can remove the selected cell from the state of the `notebook-cell-list-view.tsx`

jonah-iden:
thanks for testing i will look into it

jonah-iden:
interestingly, i just found a bug in vscode thanks your test

jonah-iden:
@msujew should be fixed now

packages/notebook/src/browser/view-model/notebook-model.ts
```diff
@@ -271,21 +261,14 @@ export class NotebookModel implements Saveable, Disposable {
...
         for (const cell of cells) {
             cell.onDidChangeOutputs(() => {
                 this.dirty = true;
             });
-            cell.onDidRequestCellEditChange(() => {
-                this.onContentChangedEmitter.fire();
-            });
+            // cell.onDidRequestCellEditChange(() => {
+            //     this.onContentChangedEmitter.fire();
+            // });
```
msujew: Question: Can this be removed entirely?

packages/notebook/src/browser/view-model/notebook-view-model.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    selectedCell?: NotebookCellModel;
+    get selectedCellViewModel(): CellViewModel | undefined {
+        if (this.selectedCell) {
+            return this.cellViewModels.get(this.selectedCell.handle);
+        }
+    }
+
+    // Cell handle to CellViewModel mapping
+    cellViewModels: Map<number, CellViewModel> = new Map();
```
msujew: Suggestion: Make this readonly to prevent accidental setting of the map?

PR #16510, opened 2025-10-30
feat: add orchestrator agent exclusion list

sdirix:
#### What it does

Add preference to exclude specific agents from orchestrator delegation, with ClaudeCode excluded by default.

- Add 'ai-features.orchestrator.excludedAgents' preference (default: ['ClaudeCode'])
- Introduce agent-specific variable 'availableChatAgents' for filtered agent list
- Override getSystemMessageDescription to resolve custom variable with exclusion filtering
- Also filter during response parsing and fallback selection

#### How to test

1. Reset your orchestrator template if you did not do so already
2. Send a request to orchestrator
3. Check the AI history to see how the `availableChatAgents` variable was resolved

---

1. Use `chatAgents` instead of `availableChatAgents` in orchestrator prompt
2. Send a request to orchestrator, hinting Claude Code to be used
3. Verify that Orchestrator still uses the fallback instead

---

1. Play around with the new preference `ai-features.orchestrator.excludedAgents` to see whether it has the desired effect

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-ide/src/common/ai-ide-preferences.ts
+++ b/packages/ai-ide/src/common/ai-ide-preferences.ts
@@ -19,6 +19,7 @@ import { nls, PreferenceSchema } from '@theia/core';
 
 // We reuse the context key for the preference name
 export const PREFERENCE_NAME_ENABLE_AI = 'ai-features.AiEnable.enableAI';
+export const PREFERENCE_NAME_ORCHESTRATOR_EXCLUSION_LIST = 'ai-features.orchestrator.excludedAgents';
 
 export const aiIdePreferenceSchema: PreferenceSchema = {
     properties: {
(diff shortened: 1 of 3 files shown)


packages/ai-ide/src/common/ai-ide-preferences.ts
```diff
@@ -37,6 +38,17 @@ export const aiIdePreferenceSchema: PreferenceSchema = {
             LLM provider below. Also see [the documentation](https://theia-ide.org/docs/user_ai/)**.'),
             type: 'boolean',
             default: false,
+        },
+        [PREFERENCE_NAME_ORCHESTRATOR_EXCLUSION_LIST]: {
+            title: AI_CORE_PREFERENCES_TITLE,
+            markdownDescription: nls.localize('theia/ai/ide/orchestrator/excludedAgents/mdDescription',
+                'List of agent IDs that the orchestrator is not allowed to delegate to. ' +
+                'These agents will not be visible to the orchestrator when selecting an agent to handle a request.'),
```
ndoschek, line 46: Nit: We could add a note to clarify that those agents are still available, e.g. something like below, but feel free to skip
`Note: Excluded agents can still be selected directly using @agent mentions, if they are configured and available.`

sdirix: If it's fine with you I would keep it as it is. It's scoped in path, name and description to the "Orchestrator" use case.

PR #16513, opened 2025-10-30
Use versioned id's for (un)installed and deployed plugins

### LINKED ISSUE #16391: Reloading window does not (always) refresh plugins

colin-grant-work:
#### What it does

Fixes #16391 by ensuring that certain pieces of state are bound to specific plugin versions rather than unversioned ID's. #16391 occurred because, after switching versions and reloading, the frontend got the new version of the plugin, but when it checked for whether its current version was uninstalled, it got a false positive from the old version. With this PR, it gets only information about its version.

#### How to test

1. Install an extension.
2. Use the 'Install a specific version command' to install another version.
3. A 'Reload Window' button should appear on the extension.
4. Reload the window
5. The extension should show the newly installed version and should _not_ show the 'Reload Window' message any longer.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
+++ b/packages/vsx-registry/src/browser/vsx-extensions-contribution.ts
@@ -352,9 +352,11 @@ export class VSXExtensionsContribution extends AbstractViewContribution<VSXExten
 
     protected async showRecommendedToast(): Promise<void> {
         if (!this.preferenceService.get(IGNORE_RECOMMENDATIONS_ID, false)) {
-            const recommended = new Set([...this.model.recommended]);
-            for (const installed of this.model.installed) {
-                recommended.delete(installed);
+            const recommended = new Set<string>();
+            for (const recommendation of this.model.recommended) {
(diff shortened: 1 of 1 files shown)


ndoschek (approved):
Thanks for the quick update, still works great and also the recommended popup works again as expected!
Regarding the second point from above, I'll create the follow up as discussed offline.

LGTM 🎉

PR #16522, opened 2025-10-31
fix: moved CommonCommands to separate file

eneufeld:
#### What it does

With CommandComamnds being part of the packages/core/src/browser/common-frontend-contribution.ts file we get circular dependencies. This leads to issues especially when running tests.

The Commands are now moved to an own file. This is a breaking change for whoever does a deep import of the CommonCommands namespace.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/browser/common-frontend-contribution.ts
+++ b/packages/core/src/browser/common-frontend-contribution.ts
@@ -19,6 +19,8 @@
 import debounce = require('lodash.debounce');
 import { injectable, inject, optional } from 'inversify';
 import { MAIN_MENU_BAR, MANAGE_MENU, MenuContribution, MenuModelRegistry, ACCOUNTS_MENU, CompoundMenuNode, CommandMenu, Group, Submenu } from '../common/menu';
+import { CommonMenus } from './common-menus';
+export { CommonMenus };
 import { KeybindingContribution, KeybindingRegistry } from './keybinding';
 import { FrontendApplication } from './frontend-application';
 import { FrontendApplicationContribution, OnWillStopAction } from './frontend-application-contribution';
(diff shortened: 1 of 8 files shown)


sdirix:
@ndoschek I think it would be good to add circular dependency analysis to the (CI) builds in some way so we can avoid issues like this in the future. They always hit at inconvenient times as they often "work" but are brittle to changes and can be triggered with slightly different usage patterns

ndoschek:
Thank you @sdirix for raising this, that sounds good. I'll create a ticket and plan it in for this month to have this check implemented soon.

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -23,6 +23,8 @@ import { KeybindingContribution, KeybindingRegistry } from './keybinding';
 import { FrontendApplication } from './frontend-application';
 import { FrontendApplicationContribution, OnWillStopAction } from './frontend-application-contribution';
 import { CommandContribution, CommandRegistry, Command } from '../common/command';
+import { CommonCommands } from './common-commands';
+export { CommonCommands };
```
colin-grant-work, line 29: With this change, the breaking change notice can probably be removed from the changelog: external consumers shouldn't be affected.

packages/core/src/browser/common-frontend-contribution.ts
```diff
@@ -112,269 +114,6 @@ export namespace CommonMenus {
 
 }
 
-export namespace CommonCommands {
```
colin-grant-work, line 115: The `CommonMenus` namespace immediately above this is similar to `CommonCommands` in that it consists entirely of dependency-free constants and is imported in a few other places, raising the risk of circular imports. We could move it out to its own file, as well.

(It's also involved in at least one circular dependency)

colin-grant-work (approved):

PR #16537, opened 2025-11-04
fix: show breakpoints from all active sessions as verified

### LINKED ISSUE #14882: Breakpoints marked as not bound although they are
### Bug Description:

I can set frontend breakpoints and successfully debug them.
However when they are set before "Launch Browser Frontend" is executed, they are marked as not bound, although they work.

### Steps to Reproduce:

1. Set two [breakpoints `renderHeader` of `about-dialog.tsx`](https://github.com/eclipse-theia/theia/blob/8b3ea1cbfbfd93e621fc1bff14d24d400656748a/packages/core/src/browser/about-dialog.tsx#L69-L70)
2. Launch Browser Backend
3. Launch Browser Frontend
4. Observe that the breakpoints are hit multiple times
5. Observe that breakpoints are both marked as `not bound`
6. Open about dialog
7. Observe that the breakpoints are hit
8. Observe that second breakpoint is marked as bound while the first one still is not
9. Set an additional breakpoint
10. Observe that now all breakpoints are marked as bound

Image

### Expected behavior

Breakpoints are always marked as bound when they are bound

### Additional Information

- Theia Version: 1.58.100

eneufeld:
#### What it does

Previously only breakpoints from the current session showed as active in the Breakpoints widget. When running multiple sessions (e.g., frontend + backend), breakpoints verified by other sessions incorrectly appeared unverified.
The Threads widget in the debug view is now readonly.

fixes #14882

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ old behavior and the new view. Default is the old behavior. [#16604](https://git
 
 - [core] `CommonCommands` has been extracted from `common-frontend-contribution.ts` into its own file `common-commands.ts`. This only affects code using deep imports: imports of `CommonCommands` from `@theia/core/lib/browser/common-frontend-contribution` should be updated to use the standard barrel export `@theia/core/lib/browser` instead.
 - [core] `CommonMenus` has been extracted from `common-frontend-contribution.ts` into its own file `common-menus.ts`. This only affects code using deep imports: imports of `CommonMenus` from `@theia/core/lib/browser/common-frontend-contribution` should be updated to use the standard barrel export `@theia/core/lib/browser` instead.
+- [debug] `DebugSessionManager.getFunctionBreakpoints()`, `DebugSessionManager.getInstructionBreakpoints()`, and `DebugSessionManager.getBreakpoints()` no longer default to the current session when called without arguments. Callers that relied on the implicit default to `currentSession` must now pass `this.currentSession` explicitly. [#16537](https://github.com/eclipse-theia/theia/pull/16537)
 - [debug] refactored some of the debug model elements as part of [#16689](https://github.com/eclipse-theia/theia/pull/16689):
   - added required `id` parameter to `DebugStackFrame` and `DebugScope` constructors
   - changed type of keys in the `DebugThread._frames` map from `number` to `string`
--- a/packages/debug/src/browser/debug-session-manager.ts
+++ b/packages/debug/src/browser/debug-session-manager.ts
@@ -429,9 +429,15 @@ export class DebugSessionManager {
                 state = session.state;
                 if (state === DebugState.Stopped) {
                     this.onDidStopDebugSessionEmitter.fire(session);
+                    // Only switch to this session if a thread actually stopped (not just state change)
+                    if (session.currentThread && session.currentThread.stopped) {
+                        this.updateCurrentSession(session);
+                    }
                 }
             }
-            this.updateCurrentSession(session);
+            // Always fire change event to update views (threads, variables, etc.)
+            // The selection logic in widgets will handle not jumping to non-stopped threads
+            this.fireDidChange(session);
         });
         session.onDidChangeBreakpoints(uri => this.fireDidChangeBreakpoints({ session, uri }));
         session.on('terminated', async event => {
@@ -450,7 +456,14 @@ export class DebugSessionManager {
         });
 
         session.onDispose(() => this.cleanup(session));
-        session.start().then(() => this.onDidStartDebugSessionEmitter.fire(session)).catch(e => {
+        session.start().then(() => {
+            this.onDidStartDebugSessionEmitter.fire(session);
+            // Set as current session if no current session exists
+            // This ensures the UI shows the running session and buttons are enabled
+            if (!this.currentSession) {
+                this.updateCurrentSession(session);
+            }
+        }).catch(e => {
             session.stop(false, () => {
                 this.debug.terminateDebugSession(session.id);
             });
@@ -611,7 +624,7 @@ export class DebugSessionManager {
         return currentThread && currentThread.topFrame;
     }
 
-    getFunctionBreakpoints(session: DebugSession | undefined = this.currentSession): DebugFunctionBreakpoint[] {
+    getFunctionBreakpoints(session?: DebugSession): DebugFunctionBreakpoint[] {
         if (session && session.state > DebugState.Initializing) {
             return session.getFunctionBreakpoints();
         }
@@ -619,7 +632,7 @@ export class DebugSessionManager {
         return this.breakpoints.getFunctionBreakpoints().map(origin => new DebugFunctionBreakpoint(origin, { labelProvider, breakpoints, editorManager }));
     }
 
-    getInstructionBreakpoints(session = this.currentSession): DebugInstructionBreakpoint[] {
+    getInstructionBreakpoints(session?: DebugSession): DebugInstructionBreakpoint[] {
         if (session && session.state > DebugState.Initializing) {
             return session.getInstructionBreakpoints();
         }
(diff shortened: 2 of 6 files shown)


packages/debug/src/browser/debug-session-manager.ts
```diff
@@ -588,15 +588,15 @@ export class DebugSessionManager {
         return currentThread && currentThread.topFrame;
     }
 
-    getFunctionBreakpoints(session: DebugSession | undefined = this.currentSession): DebugFunctionBreakpoint[] {
+    getFunctionBreakpoints(session?: DebugSession): DebugFunctionBreakpoint[] {
```
colin-grant-work, line 627: The changes in this file are breaking, and they deserve special mention in the changelog, since they're not breaking in type but are potentially very breaking in behavior.

packages/debug/src/browser/view/debug-threads-widget.ts
```diff
@@ -66,46 +67,187 @@ export class DebugThreadsWidget extends SourceTreeWidget {
...
             return;
         }
         this.updatingSelection = true;
         try {
             const { currentThread } = this.viewModel;
             if (currentThread) {
-                const node = this.model.getNode(currentThread.id);
-                if (SelectableTreeNode.is(node)) {
-                    this.model.selectNode(node);
+                console.log('[DebugThreadsWidget] updateWidgetSelection - currentThread:', currentThread.id, 'name:', currentThread.raw.name);
```
colin-grant-work: Lot of leftover logs in this class - if desired, switch to `debug` or `trace`.

packages/debug/src/browser/view/debug-threads-widget.ts
```diff
@@ -66,46 +67,187 @@ export class DebugThreadsWidget extends SourceTreeWidget {
...
-            return;
+
+    /**
+     * Wait for a node to appear in the tree, recursively expanding all session nodes
+     */
+    protected async waitForNode(threadId: string): Promise<TreeNode | undefined> {
+        const maxAttempts = 10;
+        const delayMs = 50;
+
+        const sessionId = threadId.split(':')[0];
```
colin-grant-work: I believe that the `session` is available as a property of `DebugThread`, so above where we get the current thread, we should be able to get its parent session and target the expansion a little more precisely without relying on particulars of ID construction. We should also be able to expand only the session we know should house our thread rather than all of them.

packages/debug/src/browser/view/debug-threads-widget.ts
```diff
@@ -66,46 +67,114 @@ export class DebugThreadsWidget extends SourceTreeWidget {
...
+            const root = this.model.root;
+            if (CompositeTreeNode.is(root)) {
+                await this.expandAllNodes(root);
+
+                // After expanding, try again
+                node = this.model.getNode(threadId);
+                if (node) {
+                    return node;
+                }
+            }
```
colin-grant-work: I still think we should never have to fall back to this. The thread knows its session and sessions know their parents, if there are any, so we should always be able to find the single node in the tree that we would need to expand to reveal the thread we want to see.

CHANGELOG.md
```diff
@@ -90,6 +90,7 @@
 <a name="breaking_changes_1.66.0">[Breaking Changes:](#breaking_changes_1.66.0)</a>
 
 - [core] `Listener.await` has been renamed to `Listener.awaitAll` for better compatibility with esbuild. [#16398](https://github.com/eclipse-theia/theia/pull/16398) - Contributed on behalf of STMicroelectronics
+- [debug] `DebugSessionManager.getFunctionBreakpoints()`, `DebugSessionManager.getInstructionBreakpoints()`, and `DebugSessionManager.getBreakpoints()` no longer default to the current session when called without arguments. Callers that relied on the implicit default to `currentSession` must now pass `this.currentSession` explicitly. [#16537](https://github.com/eclipse-theia/theia/pull/16537)
```
ndoschek: minor: This should be moved to the current version above

ndoschek (commented):
Thank you @eneufeld, I am really looking forward to this fix!
I tested the current state and so far it worked for most of my test cases.

However, when I am using compound launch configs, it does not work that well unfortunately. In this case I can see the second session is paused but i cannot switch (as before) to this session, and i cannot continue the debugger at all.
If i run the two configs independently from each other, then it works fine though.

In this screenshot the frontend breakpoint is hit, but I cannot select the frontend session to continue:
Image

For the compound to start, I had to adapt them a bit, removing the `stopAll` and I removed the plugin host config too:

You can find them at the bottom of the `.vscode/launch.json`:
```
"compounds": [
    {
      "name": "Launch Electron Backend & Frontend",
      "configurations": [
        "Launch Electron Backend",
        "Attach to Electron Frontend"
      ],
    },
    {
      "name": "Launch Browser Backend & Frontend",
      "configurations": [
        "Launch Browser Backend",
        "Launch Browser Frontend"
      ],
    }
  ]
```

ndoschek:
Hi @eneufeld, thanks for the updates!
Works great already! :tada: 
I just noticed two small issues with stopping debug sessions:
- if there are multiple sessions/threads running and one of those is selected and you hit stop via the toolbar, the first one in the list is stopped regardless of the selection:

   https://github.com/user-attachments/assets/e215eff5-b9e8-43d9-8e86-086bf457837e

- if there are multiple sessions/threads running and stopping one via the context menu that is not the first in the list, leaves an entry although the thread was stopped:

   https://github.com/user-attachments/assets/ef9d7d8b-ad10-4550-8f73-2542fc8dbd85

<details>
<summary>Run Simple Node App launch config</summary>

in launch.json:

```json
    {
      "type": "node",
      "request": "launch",
      "name": "Run Simple Node App",
      "program": "${workspaceFolder}/app.js"
    },
```

```js
// app.js
let counter = 0;

console.log("Counter program running.");

setInterval(() => {
    counter++;
    console.log(`Counter: ${counter}`);
}, 1000);
```
</details>

colin-grant-work:
I doubt that it's the fault of this PR, but in testing this, I'm seeing a bug where the debug configuration selector simply never updates pretty frequently. Makes it hard to get the sessions I want to run for testing going.

colin-grant-work:
> I doubt that it's the fault of this PR, but in testing this, I'm seeing a bug where the debug configuration selector simply never updates pretty frequently. Makes it hard to get the sessions I want to run for testing going.

Actually, what seems to happen is that when I try to select a different configuration, something happens that runs all of my cores at 100% for several seconds, causing everything to slow down and be unresponsive, but then eventually, the selection does take effect. The performance profile shows some missed frames in that time, so it recognizes the lack of responsiveness, but it isn't the frontend code that's causing the problem. It looks like every time I change that selector, many search processes are kicked off, and if you try to change it several times, a bunch of redundant searches kick off at the same time.

image

colin-grant-work:
After waiting out the RipGrep searches, I find I have the following problem that I think _is_ related to this branch:

- Start a debug session hit a breakpoint.
- Start another session, hit a breakpoint in that session.
- Call stack updates to reflect second breakpoint.
- Attempt to get back to first session by clicking in 'Threads' view.
- Call stack view still shows second session.
- Reveal editor where breakpoint has been hit in first session.
- No active breakpoint decorations available.

Basically, I can't figure out how to get back to the first session.

eneufeld:
@colin-grant-work you can't at the moment. we need to reenable that user can actually switch the selected session using the thread view. 
I need to look into this

eneufeld:
@ndoschek @colin-grant-work I updated the pr. you should be able to switch between session again. so the fix now focuses on marking all breakpoints as active which are active in any session and auto focusing on the thread that hit a breakpoint

ndoschek (approved):
Thanks a lot, @eneufeld! I tested it again with different scenarios, compounds and debuggers, and it works great for me! 🎉  
Really appreciate your effort!

packages/debug/src/browser/debug-session-manager.ts
Roger-Gu: @eneufeld Is there any reason the default value is removed for function and instruction breakpoints as well? Seems like the raw.verified field would not be set for them in this way.

eneufeld: If I remember correctly this was replaced because we need to be able to check all kind of breakpoints independent of the session. But in the end it does not matter as https://github.com/eclipse-theia/theia/pull/17083 removes the methods all together

Issue #16527, opened 2025-11-01
New Terminal Always Opens in Last-Used Panel Instead of Current Panel


Fangoling:
### Bug Description:
When multiple bottom panels are open, clicking the add terminal button on any bottom panel will add a new terminal to the panel of the last-used terminal, rather than the one where the button was clicked.

### Steps to Reproduce:

1. Open a terminal in the bottom panel.
2. Click the split terminal button on the bottom panel to open a second bottom panel.
3. Click on the terminal in the left panel to focus it.
4. Click on the add terminal button on the right panel

#### Observed Behaviour
The terminal widget is added to the panel of the last-used terminal

#### Expected Behaviour
The terminal widget should be added to the same panel where the 'Add Terminal' button was clicked.

### Additional Information
Operating System: MacOS 26.0.1
Theia Version: 1.65

ndoschek:
Hi @Fangoling, thanks for the detailed report!
Just FYI: We're currently working on a feature to align terminal behavior more closely with VS Code’s terminal tabs (see https://github.com/eclipse-theia/theia/issues/12673). Once that's in place, the terminal layout and behavior should be much more intuitive and self-contained.

However, the original terminal behaviour will be kept and the tabs feature is optional
In case you are interested in contributing a fix, we will happy to review and support along the way!

Fangoling:
Hi @ndoschek , thanks for the info! I will see what I can do.

PR #16538, opened 2025-11-04
fix: new terminals will be created in the panel of the clicked button

Fangoling:
#### What it does

Previously, clicking the "Add Terminal" button in the bottom panel would always create a terminal widget in the bottom panel of the "current terminal".

Now, clicking the "Add Terminal" button in a bottom panel will create a terminal widget in the same bottom panel, even when the "current terminal" is in another bottom panel.

Fixes [#16527](https://github.com/eclipse-theia/theia/issues/16527)

#### How to test

1. Open a terminal in the bottom panel.
2. Click the split terminal button on the bottom panel to open a second bottom panel.
3. Click on the terminal in the left panel to focus it.
4. Click on the add terminal button on the right panel

#### Follow-ups
In the function `splitTerminal` (line 1003) of `terminal-frontend-contribution.ts`, the result of the `withWidget` call is checked using the `||` operator instead of an explicit `=== false` comparison.
Since `withWidget` returns a value of type `T | false`, using `=== false` would make the condition clearer and more type-safe.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/terminal/src/browser/terminal-frontend-contribution.ts
+++ b/packages/terminal/src/browser/terminal-frontend-contribution.ts
@@ -534,7 +534,12 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
 
     registerCommands(commands: CommandRegistry): void {
         commands.registerCommand(TerminalCommands.NEW, {
-            execute: () => this.openTerminal()
+            execute: (widget?: Widget) => {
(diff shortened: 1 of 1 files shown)


colin-grant-work (approved):
Thanks for the fix. The new interactions are much more intuitive. 👍

Issue #16334, opened 2025-09-26
Epic: Claude Code Integration


planger:
Based on the initial integration of Claude Code (see https://github.com/eclipse-theia/theia/pull/16273), this epic contains a list of ideas of what can be done next:
 
* Providing Claude Code with much more power and insight into Theia, e.g. by hooking Theia up as an MCP server with Claude Code to read tasks, obtain diagnostics, control launches and debug sessions, etc.
* Support plan/edit modes: Here I'd suggest to introduce a generic mode concept in Theia AI and use it for Claude Code too. We could use the new `chatInputReceivingAgent` to determine the agent and ask it whether it provides modes. If yes, we could show them in the Chat UI for selecting (toggling through e.g. with Shift-Tab like in Claude Code)
* Switch welcome message based on addressed chat agent by listening to `chatInputReceivingAgent` context key
* Support for images
* Better task rendering based on the agent delegation UI we have in Theia AI
* Cancel spinner in Todo tool renderer, if the chat request is canceled
* Maybe there are a few more message types / tools that require custom rendering
* Some more [preferences](https://docs.anthropic.com/en/docs/claude-code/cli-reference#cli-flags), like max-turns, allowed/disallowed tools
* Hooking up [MCP support of Claude Code](https://docs.anthropic.com/en/docs/claude-code/mcp) with the Theia MCP server registry
* Connecting Claude Code with the Agent History view (currently only tracked in Output channel)

PR #16541, opened 2025-11-05
feat: migrate Claude Code to Theia native slash commands and modes

planger:
#### What it does

This PR migrates Claude Code to use Theia's native slash command and mode infrastructure, replacing custom Monaco-based completion providers with platform-standard implementations.

1. **Native Slash Command Support**: Migrates from a custom Monaco completion provider to Theia's `PromptService` for registering slash commands
   - Static commands (clear, compact, config, init, memory, review, resume) are registered at startup
   - Dynamic commands are loaded from `.claude/commands/*.md` files with live file watching

2. **Dynamic Command Management**: Implements file system watching for `.claude/commands/` directory
   - Automatically registers new commands when `.md` files are added
   - Updates commands when files are modified
   - Removes commands when files are deleted
   - Handles workspace changes by clearing and re-registering commands

3. **Mode Support**: Adds three operational modes to the Claude Code chat agent
   - "Ask before edit" (default): Requests permission for file modifications
   - "Edit automatically": Automatically applies suggested changes
   - "Plan mode": Plans changes without executing them
   - Integrates with Theia's native `ChatAgent.modes` interface

#### How to test

**Testing Slash Commands:**
1. Open the chat view
2. Type `@ClaudeCode /` and verify all static commands appear (clear, compact, config, init, memory, review, resume)
3. Create a `.claude/commands/` directory in your workspace
4. Add a markdown file (e.g., `test.md`) with command description
5. Verify the command appears in the slash command list without restarting
6. Modify the file name and verify the command updates
7. Delete the file and verify the command is removed
8. Switch to a different workspace and verify commands update accordingly

1. Type `@ClaudeCode /` and verify all static commands *work* (clear, compact, config, init, resume, ...)
2. Try if custom commands work

**Testing Modes:**
1. Open the chat view and type `@ClaudeCode`
3. Locate the mode selector (should show three modes)
4. Test "Ask before edit" mode: Send a request that would modify files and verify permission prompts appear
5. Test "Edit automatically" mode: Send a request and verify changes are applied without prompting
6. Test "Plan mode" mode: Send a request and verify it plans without executing

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-claude-code/src/browser/claude-code-chat-agent.ts
+++ b/packages/ai-claude-code/src/browser/claude-code-chat-agent.ts
@@ -27,8 +27,8 @@ import {
 import { AI_CHAT_NEW_CHAT_WINDOW_COMMAND, AI_CHAT_SHOW_CHATS_COMMAND } from '@theia/ai-chat-ui/lib/browser/chat-view-commands';
 import { PromptText } from '@theia/ai-core/lib/common/prompt-text';
 import { AIVariableResolutionRequest, BasePromptFragment, PromptService, ResolvedPromptFragment, TokenUsageService } from '@theia/ai-core';
-import { CommandService, nls, SelectionService } from '@theia/core';
-import { inject, injectable } from '@theia/core/shared/inversify';
+import { CommandService, ILogger, nls, SelectionService } from '@theia/core';
+import { inject, injectable, named } from '@theia/core/shared/inversify';
 import { EditorManager } from '@theia/editor/lib/browser';
 import { FileService } from '@theia/filesystem/lib/browser/file-service';
 import { WorkspaceService } from '@theia/workspace/lib/browser';
@@ -193,6 +193,9 @@ export class ClaudeCodeChatAgent implements ChatAgent {
     @inject(TokenUsageService)
     protected readonly tokenUsageService: TokenUsageService;
 
+    @inject(ILogger) @named('claude-code')
+    protected readonly logger: ILogger;
+
     async invoke(request: MutableChatRequestModel, chatAgentService?: ChatAgentService): Promise<void> {
         this.warnIfDifferentAgentRequests(request);
 
@@ -254,7 +257,7 @@ export class ClaudeCodeChatAgent implements ChatAgent {
 
             return request.response.complete();
         } catch (error) {
-            console.error('Error handling chat interaction:', error);
+            this.logger.error('Error handling chat interaction:', error);
             request.response.response.addContent(new ErrorChatResponseContentImpl(error));
             request.response.error(error);
         } finally {
@@ -393,19 +396,11 @@ export class ClaudeCodeChatAgent implements ChatAgent {
         request.addData(CLAUDE_SESSION_ID_KEY, sessionId);
     }
 
+    protected readonly ALLOWED_MODES: PermissionMode[] = ['default', 'acceptEdits', 'plan', 'bypassPermissions'];
+
     protected getClaudePermissionMode(request: MutableChatRequestModel): PermissionMode {
         const modeId = request.request.modeId ?? 'default';
-        switch (modeId) {
-            case 'acceptEdits':
-                return 'acceptEdits';
-            case 'plan':
-                return 'plan';
-            case 'bypassPermissions':
-                return 'bypassPermissions';
-            case 'default':
-            default:
-                return 'default';
-        }
+        return this.ALLOWED_MODES.includes(modeId as PermissionMode) ? modeId as PermissionMode : 'default';
     }
 
     protected getClaudeModelName(request: MutableChatRequestModel): string | undefined {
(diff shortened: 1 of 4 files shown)


packages/ai-claude-code/src/browser/claude-code-chat-agent.ts
```diff
@@ -389,11 +394,18 @@ export class ClaudeCodeChatAgent implements ChatAgent {
...
+            case 'acceptEdits':
+                return 'acceptEdits';
+            case 'plan':
+                return 'plan';
+            case 'bypassPermissions':
+                return 'bypassPermissions';
+            case 'default':
+            default:
+                return 'default';
+        }
```
sdirix: This looks a bit clunky... should we instead have an array of allowed modes and check whether the `request.modeId` is contained in them?

packages/ai-claude-code/src/browser/claude-code-slash-commands-contribution.ts
```diff
@@ -65,95 +66,170 @@ export class ClaudeCodeSlashCommandsContribution implements FrontendApplicationC
...
+            const content = await this.fileService.read(fileUri);
+            this.promptService.addBuiltInPromptFragment({
+                id: this.getDynamicCommandId(commandName),
+                template: content.value,
+                isCommand: true,
+                commandName,
+                commandAgents: [CLAUDE_CHAT_AGENT_ID]
             });
+        } catch (error) {
+            console.error(`Failed to register dynamic command ${commandName}:`, error);
```
sdirix: We should mention claude code in this error message so it's easier to understand for readers what is going on. Bonus points for using a named logger instead.

packages/ai-claude-code/src/browser/claude-code-slash-commands-contribution.ts
```diff
@@ -65,95 +66,170 @@ export class ClaudeCodeSlashCommandsContribution implements FrontendApplicationC
...
+
+        for (const filename of files) {
+            const commandName = this.getCommandNameFromFilename(filename);
+            this.promptService.removePromptFragment(this.getDynamicCommandId(commandName));
+        }
+    }
+
+    protected getWorkspaceRoot(): URI | undefined {
+        const roots = this.workspaceService.tryGetRoots();
+        return roots.length > 0 ? roots[0].resource : undefined;
```
sdirix, line 214: We should create a follow up to properly support multi root workspaces

packages/ai-claude-code/src/browser/claude-code-chat-agent.ts
```diff
@@ -389,11 +394,18 @@ export class ClaudeCodeChatAgent implements ChatAgent {
...
-    protected setClaudePermissionMode(request: MutableChatRequestModel, permissionMode: PermissionMode): void {
-        request.addData(CLAUDE_PERMISSION_MODE_KEY, permissionMode);
+        const modeId = request.request.modeId ?? 'default';
+        switch (modeId) {
+            case 'acceptEdits':
+                return 'acceptEdits';
+            case 'plan':
+                return 'plan';
+            case 'bypassPermissions':
+                return 'bypassPermissions';
```
sdirix: The `bypassPermissions` is currently not on any code path, correct?

sdirix (approved):
The change works for me. Great work 👍 

I left some comments about the code. All of them are nitpicks and can be ignored, handled later or be applied to this PR.

I performed all suggested tests. Generally speaking they succeeded for me, but there are some rough edges:

### Changing modes

The Claude Code modes work, however changing the mode while a request is in process does not work. Even if the request is stopped because Claude Code asks for an approval, the new mode will not be respected by Claude Code until there is an actual new request from the user.

As we start in "Ask before edit" mode, the user will typically encounter an approval message pretty quick. Then they change the mode and click approve, still all further changes need to be manually improved

### ExitPlanTool

Claude Code has an ExitPlanTool it tries to use to exit plan mode (after approval from the user). This fails and the user has to change the mode manually afterwards to continue

Image

### "Ask before edit" does not populate the change set

The "Ask before edit" asks for edit approval before doing any edits. However the approved edits do not show up in the changeset. Therefore there is no convenient way of directly seeing the edits of what Claude Code did compared to the "Edit automatically" mode

### Custom commands request rendering

This "regression" is not caused by this PR. However it seems very weird to me that the Claude Code commands are rendered as `#prompt` in the request, e.g.

Image

### Mode preference?

Personally I don't like that I need to change the mode for Claude Code in _every_ new chat session. There should be _some_ way to avoid this, for example via a preference.

---

None of these are deal breakers, so we can merge as is. However please capture these remarks as follow ups in case you don't tackle them here.

packages/ai-claude-code/src/browser/claude-code-chat-agent.ts
planger: true

planger:
Thanks for the great review @sdirix!

I've addressed your inline comments and rebased and tracked all follow-ups in https://github.com/eclipse-theia/theia/issues/16334.

Issue #16415, opened 2025-10-09
Support `dockerComposeFile` in Devcontainer Spec


Aweptimum:
### Feature Description:
Moving here from https://github.com/eclipse-theia/theia/issues/15587

Currently, the devcontainer feature is missing support for developing using a `compose.yml` file. This is necessary for allowing the devcontainer extension to validate and provision containers in the case that your project has more than one service (such as a backend + containerized database).

Additionally, I believe this is an area where Theia can improve over VS Code's implementation. Currently, I do not believe VS Code has support for passing arguments to the underlying `docker compose up` CLI command. I'm not sure why there hasn't been widespread enough demand for the feature, but it would simply be a matter of parity with `runArgs` when using a Dockerfile.
Relevant issue: https://github.com/microsoft/vscode-remote-release/issues/6298

JonasHelming:
@msujew

msujew:
I would continue the game of tag! @jonah-iden

jonah-iden:
I would love to add the docker compose support, but currently don't really have time for it. Other projects are slowing down a bit though. So maybe soon

jonah-iden:
I created a PR for basic dev container support [here](https://github.com/eclipse-theia/theia/pull/16577)

As you requested I also added the new `composUpArgs` to the devcontainer.json (no autocomplete for it though) to add custom arguments to the docker compose up call. Can be used like `"composeUpArgs": ["--force-recreate"]`

Aweptimum:
You are a massive hero 🙏 I'll have to try this when I have free time

jonah-iden:
I would close issue since the PR has been merged

PR #16017, opened 2025-07-18
Make Preferences available in the back end

tsmaeder:
#### What it does

The PR makes preferences support available in the back end. Only default and user preferences can be accessed in the back end. The API has changed in the following ways:
- Many files have been moved from the `browser` folder to the `common` folder. Imports will have to be adapted
- `PreferenceSchemaProvider` has been replaced by two separate `PreferenceSchemaServiceImpl` (and corresponding interface) and `DefaultsPreferenceProvider` classes.
- Preference schema typing has been simplified: a preference schema is no longer extending IJSONSchema and typing has been adapted to strictly use Theia types (for example for scopes) and a straight-forward extension of standard IJSONSchema for properties. This means schemas from VS Code (contributed via extensions) must be converted to Theia format.
- `PrefenceSchemaService` separates between adding a schema and registering a default override for a property. Also, the service uses explicit override identifiers instead of encoding the override in the preference key. The schema service strictly distinguishes between preference schema and the derived JSON Schema for preference files. `JSONValue` is now used instead of `any` where applicable. Schema properties must be added before overrides are registered.
`PreferenceSchemaService` now has the concept of`validScopes`. In the back end, only`Default` and `User` can be used.
- `PreferenceContribution` now has a `initSchema()` method in addition to the declarative Schema contribution. It is used to register overrides.  

Fixes https://github.com/eclipse-theia/theia/issues/15635

#### How to test

The PR adds three commands: "Get Backend Preference", "Inspect Backend Preference" and "Set Backend Preference". To test front-end prefernces, use the preference UI.

#### Follow-ups
https://github.com/eclipse-theia/theia/issues/16046
https://github.com/eclipse-theia/theia/issues/16044

#### Breaking changes

- [x] This PR introduces breaking changes and requires careful review. If yes, the breaking changes section in the [changelog](https://github.com/eclipse-theia/theia/blob/master/CHANGELOG.md) has been updated.

#### Attribution
Contributed on behalf of STMicroelectronics

#### Review checklist

- [ ] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #16577, opened 2025-11-12
Basic devcontainer Docker Compose support

jonah-iden:
#### What it does

implements #16415

This adds a first basic support for docker compose based dev containers.
Also includes a small fix for the forward port widget not showing forwarded ports correctly which were configured through the devcontainer.json file.

#### How to test

these two test files need to be in the .devcontainer folder in your workspace:

devcontainer.json:
```
{
	"name": "Docker compose test",
	
    "dockerComposeFile": "docker-compose.yml",
    "service": "devcontainer",
    "customizations": {
        "vscode": {
          "extensions": ["streetsidesoftware.code-spell-checker"]
        }
      },
      "forwardPorts": [3000]
}
```

docker-compose.yml:
```
version: '3.8'
services:
  devcontainer:
    image: mcr.microsoft.com/devcontainers/base:ubuntu
    volumes:
      - ../..:/workspaces:cached
    network_mode: service:db
    command: sleep infinity

  db:
    image: postgres:latest
    restart: unless-stopped
    volumes:
      - postgres-data:/var/lib/postgresql/data
    environment:
      POSTGRES_PASSWORD: postgres
      POSTGRES_USER: postgres
      POSTGRES_DB: postgres

volumes:
  postgres-data:
```

#### Follow-ups

Not all properties from ther devcontainer.json schema are implemented yet. Specificly for compose the `runServices` and 
`overrideCommand` are still missing.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/dev-container/src/electron-node/remote-container-connection-provider.ts
+++ b/packages/dev-container/src/electron-node/remote-container-connection-provider.ts
@@ -25,7 +25,7 @@ import { RemoteConnectionService } from '@theia/remote/lib/electron-node/remote-
 import { RemoteProxyServerProvider } from '@theia/remote/lib/electron-node/remote-proxy-server-provider';
 import { Emitter, Event, generateUuid, MessageService, RpcServer, ILogger } from '@theia/core';
 import { Socket } from 'net';
-import { inject, injectable } from '@theia/core/shared/inversify';
+import { inject, injectable, named } from '@theia/core/shared/inversify';
 import * as Docker from 'dockerode';
 import { DockerContainerService } from './docker-container-service';
 import { Deferred } from '@theia/core/lib/common/promise-util';
@@ -212,7 +212,7 @@ interface ContainerTerminalSession {
 
 export class RemoteDockerContainerConnection implements RemoteConnection {
 
-    @inject(ILogger) @named("dev-container")
+    @inject(ILogger) @named('dev-container')
     protected readonly logger: ILogger;
 
     id: string;


jonah-iden:
@tsmaeder if you have some time could you take a look at [this commit](https://github.com/eclipse-theia/theia/pull/16577/commits/c4067a1970a3b8c03ac3fa9c2868b33f596ee29a). not sure if that fix is ok and the issue seems related to [this PR](https://github.com/eclipse-theia/theia/issues/16017) you created a while ago. 
I found a race condition (mostly happening for me when launching a dev container) between the `init` of the `WorkspacePreferenceProvider` and the first `updateRoots` call in the `WorkspaceService`.  
If `updateRoots` is called before `WorkspacePreferenceProvider` can register it's `onWorkspaceLocationChanged` listener, `ensureDelegateUpToDate` is never called and with that is `_ready` never resolved, leading to the `PreferenceServiceImpl` `_ready` also never being resolved. Then you can not open editors and the status bar never becomes visible.

tsmaeder:
@jonah-iden not sure what you mean: after `WorkspacePreferenceProvider.init()` concludes, the listeners are already registered. The only window for a race condition is IMO if the block inside the `then()` is executed and resolving ready executes other thenables. But wouldn't that be easily solved by registering the listeners before `this.workspace.ready()....`?

jonah-iden:
@tsmaeder ok thanks for taking a look. i can try and check if the registering before the then helps

jonah-iden:
@tsmaeder moving the listeners as you suggest did not work. The issue seems to be that the `WorkspaceService.init` postConstruct is called before the `WorkspacePreferenceProvider.init`. Normally `WorkspaceService.init` does not get to the firing the `onWorkspaceChangeEmitter` emitter in one tick so the `WorkspacePreferenceProvider.init` is called before that. But in case of devcontainers it seems it can happen that the workspace service is initialized in one tick so `WorkspacePreferenceProvider.init` is only executed after the `onWorkspaceChangeEmitter`has been fired

packages/dev-container/src/electron-node/remote-container-connection-provider.ts
```diff
@@ -295,6 +303,7 @@ export class RemoteDockerContainerConnection implements RemoteConnection {
             });
             const stdout = new PassThrough();
             stdout.on('data', (data: Buffer) => {
+                console.log('REMOTE STDOUT:', data.toString());
```
msujew: Question: Leftover from testing? If you want to keep this for debug purposes, you could a dedicated logger for that and use the `debug` channel to ensure that we don't pollute the backend logs.

packages/dev-container/src/electron-node/remote-container-connection-provider.ts
```diff
@@ -305,6 +314,7 @@ export class RemoteDockerContainerConnection implements RemoteConnection {
             });
             const stderr = new PassThrough();
             stderr.on('data', (data: Buffer) => {
+                console.error('REMOTE STDERR:', data.toString());
```
msujew: Same here.

msujew (approved):
Looks good to me, just some minor stuff, see below.

packages/dev-container/src/electron-node/remote-container-connection-provider.ts
```diff
@@ -212,6 +212,9 @@ interface ContainerTerminalSession {
 
 export class RemoteDockerContainerConnection implements RemoteConnection {
 
+    @inject(ILogger)
+    protected readonly logger: ILogger;
```
msujew: Please use a named logger

```suggestion
    @inject(ILogger) @named("dev-container")
    protected readonly logger: ILogger;
```

jonah-iden: thanks that's what i wasn't sure about if it was correct

packages/dev-container/src/electron-node/remote-container-connection-provider.ts
```diff
@@ -208,6 +212,9 @@ interface ContainerTerminalSession {
 
 export class RemoteDockerContainerConnection implements RemoteConnection {
 
+    @inject(ILogger) @named("dev-container")
```
msujew: The import for `named` is missing.

PR #16579, opened 2025-11-12
dev: improve error handling in catalog.json download script

sdirix:
#### What it does

Provide clear error message and workarounds when the catalog.json download fails. This file is downloaded during the build.

This is useful for adopters/contributors who run into issues when trying to build Theia.

#### How to test

- Make sure the download script works as before
- You can test the error case by modifying the script and setting a bogus URL
- You can test the workaround by using a bogus URL but the file already in place

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/scripts/download-catalog.js
+++ b/packages/core/scripts/download-catalog.js
@@ -57,6 +57,7 @@ Possible workarounds:
   3. Download the file manually and place it at:
      ${targetFile}
      Download from: ${url}
+     Adapt core npm scripts to skip automatic download.
 `;
     console.error(errorMessage);
     process.exit(1);


ndoschek (commented):
Looks like it is just missing an import for path (`ReferenceError: path is not defined`) TIA!

sdirix:
I accidentally removed it when I removed an actually unnecessary import. Thanks for the notice

ndoschek (commented):
Thanks @sdirix for the update, this is a great improvement!
The error messages when the download fails are now much more informative and provide clear instructions on how to resolve the issue.

Regarding the workaround with the manual download: this alone doesn't fully resolve it for me. Even if I place the file manually and retry, the download still fails and I can't get past this step then.
I checked and saw that there's an option for the downloader: `skipExistingFileName: true`. This would allow it to accept the manually downloaded file and proceed.
We could also consider logging something in this case, not sure if necessary though, wdyt?

sdirix:
I only adapted the workaround description to not change the behavior of the script.

ndoschek (approved):
Thanks, LGTM now 👍

PR #16597, opened 2025-11-16
Make next replacer function default

JonasHelming:
#### What it does

- Make the next replacer function default for suggest file changes
- Keep Coder next prompt, but make it "identical" to edit prompt for now
- Kepp old function as "_simple" variant for no, but mark it deprecated

#### How to test

Test Theia Coder or any agent using the replace function (such as project info)

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-ide/src/common/coder-replace-prompt-template.ts
+++ b/packages/ai-ide/src/common/coder-replace-prompt-template.ts
@@ -28,8 +28,7 @@ import {
     SUGGEST_FILE_REPLACEMENTS_ID,
     WRITE_FILE_REPLACEMENTS_ID,
     CLEAR_FILE_CHANGES_ID,
-    GET_PROPOSED_CHANGES_ID,
-    SUGGEST_FILE_REPLACEMENTS_NEXT_ID
+    GET_PROPOSED_CHANGES_ID
 } from './file-changeset-function-ids';
 
 export const CODER_SYSTEM_PROMPT_ID = 'coder-system';
@@ -161,7 +160,7 @@ You are an autonomous AI agent. Do not stop until:
     };
 }
 
-function getCoderEditPromptTemplate(suggestFileReplacementsId: string): string {
+function getCoderEditPromptTemplate(): string {
     return `{{!-- This prompt is licensed under the MIT License (https://opensource.org/license/mit).
 Made improvements or adaptations to this prompt template? We'd love for you to share it with the community! Contribute back here:
 https://github.com/eclipse-theia/theia/discussions/new?category=prompt-template-contribution --}}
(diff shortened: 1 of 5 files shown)


packages/ai-ide/src/common/coder-replace-prompt-template.ts
sdirix, line 1: Shouldn't the template `CODER_SIMPLE_EDIT_TEMPLATE_ID` use the tool with `SUGGEST_FILE_REPLACEMENTS_SIMPLE_ID`?

As of right now the simple template is almost the same as the regular/next template, just with two paragraphs less.

So if the simple template is meant for backwards compatibility for users for whom the new tools might not work, then it should also use the simple tools.

packages/ai-ide/src/common/coder-replace-prompt-template.ts
```diff
@@ -235,14 +234,14 @@ You have previously proposed changes for the following files. Some suggestions m
 export function getCoderPromptTemplateEdit(): BasePromptFragment {
     return {
         id: CODER_EDIT_TEMPLATE_ID,
-        template: getCoderEditPromptTemplate(SUGGEST_FILE_REPLACEMENTS_ID)
+        template: getCoderEditPromptTemplate()
     };
 }
-
+// Currently, the next template is identical to the regular edit prompt
```
sdirix, line 240: Do we keep the next template in case we want to do further experiments in the future? If yes that's fine with me. If it's just because of backward compatibility I would rather remove it.

sdirix (changes requested):
It works but I think the simple template is not yet fully correct, see my commet

packages/ai-ide/src/common/coder-replace-prompt-template.ts
JonasHelming: Yes, I would like to keep it for future experiments. People who selected "next" now, will automatically be part of the experiement then.

packages/ai-ide/src/common/coder-replace-prompt-template.ts
JonasHelming: The simple template was actually meant for providing a less complex prompt for less powerful LLMs. That is why it has less tools available. The new replacement function does not increase the complexity for the LLM, so I would argue that it should use the new functions.

What we could do though is remove the simple template alltogether, as I believe it is hardly used at all. If you agree, I would do this in a separate PR, though.

sdirix: This means the simple replace tools are unused now. Should we remove them?

JonasHelming: I marked them as deprecated in case an adopter wants to explicitly use them. I have a trigger to remove them in three months.

PR #16599, opened 2025-11-17
enableAgent/disableAgent: add async/await

safisa:
#### What it does

This PR adds async and await to the enableAgent and disableAgent methods in the AgentServiceImpl class, to allow multiple callers in a row to work as expected (using await in these methods).

#### How to test

Try to call this:

```
        await this.agentService.disableAgent("AppTester");
        await this.agentService.disableAgent("ProjectInfo");
```

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-ide/src/browser/ai-configuration/agent-configuration-widget.tsx
+++ b/packages/ai-ide/src/browser/ai-configuration/agent-configuration-widget.tsx
@@ -347,16 +347,16 @@ export class AIAgentConfigurationWidget extends ReactWidget {
         this.updateParsedPromptParts();
     }
 
-    private toggleAgentEnabled = () => {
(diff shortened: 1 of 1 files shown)


packages/ai-core/src/common/agent-service.ts
```diff
@@ -131,14 +131,14 @@ export class AgentServiceImpl implements AgentService {
         return this._agents;
     }
 
-    enableAgent(agentId: string): void {
+    async enableAgent(agentId: string): Promise<void> {
```
ndoschek, line 134: Could you also update the interface definitions higher up in the file so they reflect the new async signatures? TIA!

safisa: done

PR #16603, opened 2025-11-17
Improve typing related to `globalThis` object

msujew:
#### What it does

Follow up on https://github.com/eclipse-theia/theia/pull/16602. I've noticed that we rely quite a lot on `@ts-expect-error` when handling custom properties on the `globalThis` object. This change adds a dedicated type and a constant that uses this new type.

#### How to test

Ensure that the features related to the usage of the `globalThis` object still works expected. I.e. opening the "About" popup still results in the extensions showing up.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/node/backend-global.ts
+++ b/packages/core/src/node/backend-global.ts
@@ -16,6 +16,7 @@
 
 import type { AddressInfo } from 'net';
 import type { ExtensionInfo } from '../common/application-protocol';
+import { isPromise } from 'util/types';
 
 export interface BackendGlobal {
     serverAddress?: Promise<AddressInfo>;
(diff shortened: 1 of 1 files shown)


packages/core/src/node/backend-global.ts
```diff
@@ -0,0 +1,36 @@
...
+    extensionInfo: ExtensionInfo[];
+}
+
+/**
+ * The global object for the backend application.
+ * Used to store application-wide information.
+ *
+ * See {@link BackendGlobal} for more details.
+ */
+export const backendGlobal: BackendGlobal = globalThis as unknown as BackendGlobal;
```
sdirix, line 32: Should we check whether the global looks like we expect before doing this cast, i.e. check whether `serverAdress` exists and if yes, whether it has a then?

The other property is already checked and overridden/initialized when wrong

msujew: Good point, I've added an error console log in case the `serverAddress` is invalid. I didn't want to throw in top-level code.

sdirix (approved):
Works for me! Thanks for the changes 👍

PR #12726, opened 2023-07-18
Draft: Terminal Manager 

kenneth-marut-work:
#### What it does

Relevant to #12673

This is PR adds the package `terminal-manager`, which was built for our downstream IDE. It is heavily inspired by VSCode's split-panel terminal view + tree, but not an exact copy. The terminal manager is separate widget that supplements the existing terminals inside Theia but is not meant as a replacement (at the moment). The terminal-manager's purpose is to serve as a place to maintain and organize terminals to improve a user's way of working.

I'm hoping to hand off this feature to anyone that's interested as I will not have the bandwidth to bring this feature up to parity with VSCode. This hopefully will serve as a good jumping off point for some interesting terminal view improvements to Theia. Happy to answer any questions otherwise :)

# Description

terminal-manager

**How to test**
- Build and run this branch
- Start Theia and run the command "Open Terminal Manager"
  - The terminal manager should appear in the bottom panel and look like this:

image

- On start you are given defaults of one page, one group, and one terminal, you can add additional pages by hovering over the treewidget and clicking the icons on the tree nodes
- Use the new keybinding `alt+q` to maximize and minimize the terminal panel. This keybinding is meant to serve as a convenient way of switching between the normal "editor" view and a fullscreen "terminal" view.  (you can also use the new toolbar item to maximize/minimize)

image

- Try adding groups, pages, terminals to get something like this:

image

- Right click on tree nodes and try renaming nodes
- Click on different nodes, corresponding terminals should flash when their node is selected (and vice versa)
  - Pages should automatically update when their respective nodes are selected (and renamed)
- Try deleting groups, pages, terminals from the tree widget (you won't be allowed to delete the last page/group/terminal for its respective parent, does this behavior make sense?)

- Try showing and hiding the tree widget with this icon: 

image

- If you want the tree widget to appear on the left/right, set this preference;

image

- The terminal manager's layout and state is stored in local storage
  - layouts should be stored with their corresponding workspace so a layout created in Workspace A will be saved whenever workspace A is opened, new workspaces/windows should start with a fresh terminal manager

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#requesting-a-review)

#### Reminder for reviewers

- As a reviewer, I agree to behave in accordance with [the review guidelines](https://github.com/theia-ide/theia/blob/master/doc/pull-requests.md#reviewing)

PR #16604, opened 2025-11-17
Terminal manager for multiple terminals in one view

### LINKED ISSUE #12673: VSCode style terminal tabs
### Feature Description:
Currently when new terminal instances are opened in the same widget they are displayed as tabs along the top of that widget. My proposal is to have a side bar where the terminal instances are displayed, see the image for an example.

image

lucas-koehler:
#### What it does

Closes #12673

Builds on top of https://github.com/eclipse-theia/theia/pull/12726

Introduces a new extension `terminal-manager` providing a terminal manager widget that enables managing multiple terminals inside a single view.
It supports splitting vertically and horizontally and supports multiple pages.

The extension provides a setting `terminal.grouping.mode` to configure whether new terminals are created as separate views (as before) or inside the terminal manager. For now, the old behavior is the default.

[theia-terminal-manager.webm](https://github.com/user-attachments/assets/50f9eb8e-aec5-45a8-8e49-43992eba50ba)

#### How to test

- Got to settings and set the terminal > grouping > mode setting to `tree`.
- Create and use terminals
- Set the setting to `separate` and create a new terminal. It should be created as a separate view again

#### Follow-ups

- Streamline UX of tree view
  - possibly hide tree for one terminal only
  - possibly get rid of explicit page nodes in tree view to align with VSCode and reduce UI clutter
- Support creating new terminal for profile in the terminal manager (command `TerminalCommands.PROFILE_NEW`)
- Automatically name terminal tree nodes based on folder and/or running command or shell

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/terminal-manager/src/browser/terminal-manager-widget.ts
+++ b/packages/terminal-manager/src/browser/terminal-manager-widget.ts
@@ -46,7 +46,6 @@ import { ConfirmDialog } from '@theia/core/lib/browser/dialogs';
 export namespace TerminalManagerWidgetState {
     export interface BaseLayoutData<ID> {
         id: ID,
-        childLayouts: unknown[];
     }
     export interface TerminalWidgetLayoutData {
         widget: TerminalWidget | undefined;
@@ -65,8 +64,7 @@ export namespace TerminalManagerWidgetState {
         childLayouts: PageLayoutData[];
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    export const isLayoutData = (obj: any): obj is LayoutData => typeof obj === 'object' && !!obj && 'type' in obj && obj.type === 'terminal-manager';
+    export const isLayoutData = (obj: unknown): obj is LayoutData => typeof obj === 'object' && !!obj && 'type' in obj && obj.type === 'terminal-manager';
     export interface PanelRelativeSizes {
         terminal: number;
         tree: number;
@@ -130,13 +128,12 @@ export class TerminalManagerWidget extends BaseWidget implements StatefulWidget,
     }
 
     @postConstruct()
-    protected async init(): Promise<void> {
+    protected init(): void {
         this.title.iconClass = codicon('terminal-tmux');
         this.id = TerminalManagerWidget.ID;
         this.title.closable = true;
         this.title.label = TerminalManagerWidget.LABEL;
         this.node.tabIndex = 0;
-        await this.terminalManagerPreferences.ready;
         this.registerListeners();
         this.createPageAndTreeLayout();
     }
@@ -211,7 +208,7 @@ export class TerminalManagerWidget extends BaseWidget implements StatefulWidget,
         }
     }
 
-    protected createPageAndTreeLayout(relativeSizes?: TerminalManagerWidgetState.PanelRelativeSizes): void {
+    protected async createPageAndTreeLayout(relativeSizes?: TerminalManagerWidgetState.PanelRelativeSizes): Promise<void> {
         const layout = this.layout = new PanelLayout();
         this.pageAndTreeLayout = new SplitLayout({
             renderer: SplitPanel.defaultRenderer,
@@ -223,14 +220,15 @@ export class TerminalManagerWidget extends BaseWidget implements StatefulWidget,
         });
 
         layout.addWidget(this.panel);
-        this.resolveMainLayout(relativeSizes);
+        await this.resolveMainLayout(relativeSizes);
         this.update();
     }
 
-    protected resolveMainLayout(relativeSizes?: TerminalManagerWidgetState.PanelRelativeSizes): void {
+    protected async resolveMainLayout(relativeSizes?: TerminalManagerWidgetState.PanelRelativeSizes): Promise<void> {
         if (!this.pageAndTreeLayout) {
             return;
         }
+        await this.terminalManagerPreferences.ready;
         const treeViewLocation = this.terminalManagerPreferences.get('terminal.grouping.treeViewLocation');
         const widgetsInDesiredOrder = treeViewLocation === 'left' ? [this.treeWidget, this.terminalPanelWrapper] : [this.terminalPanelWrapper, this.treeWidget];
         widgetsInDesiredOrder.forEach((widget, index) => {


packages/terminal-manager/src/browser/terminal-manager-alert-dialog.tsx
```diff
@@ -0,0 +1,115 @@
...
+        }
+        this.accept();
+    }
+
+    get value(): string | 'close' {
+        return this.selectedValue;
+    }
+}
+
+export const bindGenericErrorDialogFactory = (bind: interfaces.Bind): void => {
```
colin-grant-work: From the name, it looks like this is more generic than specific to the terminal manager. Is it useful enough to be put in `core` to be made more widely available?

packages/terminal-manager/src/browser/terminal-manager-frontend-contribution.ts
```diff
@@ -0,0 +1,170 @@
...
+        this.commandRegistry = commands;
+        this.preferences.ready.then(() => {
+            if (this.preferences.get('terminal.tabs.display') !== 'manager') {
+                console.debug('Terminal tab style is not manager. Use separate terminals.');
+                return;
+            }
+            console.debug('Terminal tab style is manager. Override command handlers accordingly.');
+            this.registerHandlers(commands);
+        });
+    }
```
colin-grant-work: I believe the `CommandRegistry` can be `inject`ed. If so, then I think it would make sense not to make this contribution a `CommandContribution` but just a `FrontendApplicationContribution` with an `onStart` method that handles checking the preference and conditionally registering commands and setting up the preference change listener.

packages/terminal-manager/src/browser/terminal-manager-frontend-view-contribution.ts
```diff
@@ -0,0 +1,316 @@
...
+import { TerminalManagerCommands, TerminalManagerTreeTypes, TERMINAL_MANAGER_TREE_CONTEXT_MENU } from './terminal-manager-types';
+import { TerminalManagerWidget } from './terminal-manager-widget';
+import { TerminalManagerTreeWidget } from './terminal-manager-tree-widget';
+import { AlertDialogFactory } from './terminal-manager-alert-dialog';
+
+@injectable()
+export class TerminalManagerFrontendViewContribution extends AbstractViewContribution<TerminalManagerWidget>
+    implements TabBarToolbarContribution, KeybindingContribution {
+    protected onBottomPanelMaximizeDidChangeEmitter = new Emitter<void>();
+    protected onBottomPanelMaximizeDidChange = this.onBottomPanelMaximizeDidChangeEmitter.event;
```
colin-grant-work: I think the shell has an event like this, so if we have a reference to the shell (and I think `ViewContributions` do), then instead of maintaining our own listener and only firing it when this contribution maximizes the the bottom panel - potentially missing cases where other actors maximize the panel - we can use that event below instead:

```typescript
        const bottomPanelMaximizationChanged = Event.map(Event.filter(this.shell.onDidToggleMaximized, widget => widget === this.shell.bottomPanel), () => undefined);
        toolbar.registerItem({
            id: TerminalManagerCommands.MANAGER_MAXIMIZE_BOTTOM_PANEL_TOOLBAR.id,
            command: TerminalManagerCommands.MANAGER_MAXIMIZE_BOTTOM_PANEL_TOOLBAR.id,
            icon: codicon('chevron-up'),
            onDidChange: bottomPanelMaximizationChanged,
        });
        toolbar.registerItem({
            id: TerminalManagerCommands.MANAGER_MINIMIZE_BOTTOM_PANEL_TOOLBAR.id,
            command: TerminalManagerCommands.MANAGER_MINIMIZE_BOTTOM_PANEL_TOOLBAR.id,
            icon: codicon('chevron-down'),
            onDidChange: bottomPanelMaximizationChanged,
        });
```

(Personally, I'd skip the `Event.map` and just assert that `Event<Widget>` satisfies `Event<void>`, but that's a pet peeve for another day :-)).

packages/terminal-manager/src/browser/terminal-manager-preferences.ts
```diff
@@ -0,0 +1,51 @@
...
+        },
+        'terminal.tabs.display': {
+            'type': 'string',
+            'enum': ['separate', 'manager'],
+            'description': nls.localize('theia/terminalManager/tabsDisplay',
+                'Controls how terminal tabs are displayed. \'manager\' shows terminals in a singular tabbed terminal manager view,'
+                + '\'separate\' shows each terminal in its own view.'),
+            'default': 'separate',
+            'scope': PreferenceScope.User,
+        },
```
colin-grant-work: At a minimum:

```suggestion
        'terminal.tabs.display': {
            'type': 'string',
            'enum': ['separate', 'manager'],
            'description': nls.localize('theia/terminalManager/tabsDisplay',
                'Controls how terminal tabs are displayed. \'manager\' shows terminals in a singular tabbed terminal manager view, '
                + '\'separate\' shows each terminal in its own view.'),
            'default': 'separate',
            'scope': PreferenceScope.User,
        },
```

But the terminology here is a little odd, since if we're using the terminal manager, there aren't really terminal tabs at all.

Maybe something like

```suggestion
        'terminal.grouping': {
            'type': 'string',
            'enum': ['tabbed', 'tree'],
            'description': nls.localize('theia/terminalManager/tabsDisplay',
                'Controls how terminals are displayed. \'tree\' shows multiple terminals in a single view with a tree view for management,'
                + '\'tabbed\' shows each terminal in its own view in a separate tab.'),
            'default': 'tabbed',
            'scope': PreferenceScope.User,
        },
```

I also think there's no reason this couldn't be a workspace-scoped preference - its effects are limited to a specific frontend - but I have no strong feelings about that. I don't see why people would prefer different terminal management strategies for different workspaces, but workspace scope is a little more flexible, in case people surprise me.

packages/terminal-manager/src/browser/terminal-manager-tree-model.ts
```diff
@@ -0,0 +1,349 @@
...
+    protected onTerminalAddedToGroupEmitter = new Emitter<{
+        terminalId: TerminalManagerTreeTypes.TerminalKey,
+        groupId: TerminalManagerTreeTypes.GroupId,
+    }>();
+
+    readonly onTerminalAddedToGroup = this.onTerminalAddedToGroupEmitter.event;
+    protected onTerminalDeletedFromGroupEmitter = new Emitter<{
+        terminalId: TerminalManagerTreeTypes.TerminalKey,
+        groupId: TerminalManagerTreeTypes.GroupId,
+    }>();
```
colin-grant-work: We really aren't very consistent about it, but our coding guidelines [do say](https://github.com/eclipse-theia/theia/blob/master/doc/coding-guidelines.md#event_names) that events should follow the `on[Will|Did]VerbNoun?` pattern.

packages/terminal-manager/src/browser/terminal-manager-tree-model.ts
```diff
@@ -0,0 +1,349 @@
...
+        this.toDispose.push(this.selectionService.onSelectionChanged(selectionEvent => {
+            const selectedNode = selectionEvent.find(node => node.selected);
+            if (selectedNode) {
+                this.handleSelectionChanged(selectedNode);
+            }
+        }));
+        this.root = { id: 'root', parent: undefined, children: [], visible: false } as CompositeTreeNode;
+    }
+
+    protected getContext = () => this;
```
colin-grant-work: Unused?

packages/terminal-manager/src/browser/terminal-manager-tree-model.ts
```diff
@@ -0,0 +1,349 @@
...
+            name += pageLabel;
+        }
+        if (groupLabel) {
+            name += ` > ${groupLabel}`;
+        }
+        if (terminalLabel) {
+            name += ` > ${terminalLabel}`;
+        }
+        return name;
+    }
```
colin-grant-work: I don't see any uses of this method(?), and if it were used, it should probably go in a label provider.

packages/terminal-manager/src/browser/terminal-manager-widget.ts
```diff
@@ -0,0 +1,714 @@
...
+import { TerminalManagerPreferences } from './terminal-manager-preferences';
+import { TerminalManagerTreeTypes } from './terminal-manager-types';
+import { TerminalManagerTreeWidget } from './terminal-manager-tree-widget';
+import { AlertDialogFactory } from './terminal-manager-alert-dialog';
+
+export namespace TerminalManagerWidgetState {
+    export interface BaseLayoutData<ID> {
+        id: ID,
+        childLayouts: unknown[];
+    }
```
colin-grant-work, line 49: I think it would be better to either omit `childLayouts` or parametrize that value as well. It looks like we override it in every concrete case, so the `unknown` value never surfaces.

packages/terminal-manager/src/browser/terminal-manager-widget.ts
```diff
@@ -0,0 +1,714 @@
...
+    export interface PageLayoutData extends BaseLayoutData<TerminalManagerTreeTypes.PageId> {
+        childLayouts: TerminalGroupLayoutData[];
+        groupRelativeWidths: number[] | undefined;
+    }
+    export interface TerminalManagerLayoutData extends BaseLayoutData<'ParentPanel'> {
+        childLayouts: PageLayoutData[];
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    export const isLayoutData = (obj: any): obj is LayoutData => typeof obj === 'object' && !!obj && 'type' in obj && obj.type === 'terminal-manager';
```
colin-grant-work:
```suggestion
    export const isLayoutData = (obj: unknown): obj is LayoutData => typeof obj === 'object' && !!obj && 'type' in obj && obj.type === 'terminal-manager';
```

packages/terminal-manager/src/browser/terminal-manager-widget.ts
```diff
@@ -0,0 +1,714 @@
...
+        if (!this.pageAndTreeLayout) {
+            return;
+        }
+        const treeViewLocation = this.terminalManagerPreferences.get('terminal.tabs.treeViewLocation');
+        const widgetsInDesiredOrder = treeViewLocation === 'left' ? [this.treeWidget, this.terminalPanelWrapper] : [this.terminalPanelWrapper, this.treeWidget];
+        widgetsInDesiredOrder.forEach((widget, index) => {
+            this.pageAndTreeLayout?.insertWidget(index, widget);
+        });
+        this.setPanelSizes(relativeSizes);
+    }
```
colin-grant-work: It looks like the preference check here is the reason for waiting on the `.ready` in `init`. This operation isn't completely free, but it also isn't so expensive that rerunning it once or twice as preferences resolve would really hurt either. I don't have strong feelings, but asynch `init` still makes me nervous, even if we correctly have a `asyncGet` in the one place where Inversify currently handles this widget.

packages/terminal-manager/src/browser/terminal-manager-widget.ts
```diff
@@ -0,0 +1,714 @@
...
+        const terminalKey = TerminalManagerTreeTypes.generateTerminalKey(widget);
+        this.selectTerminalNode(terminalKey);
+    }
+
+    protected selectTerminalNode(terminalKey: TerminalManagerTreeTypes.TerminalKey): void {
+        const node = this.treeWidget.model.getNode(terminalKey);
+        if (node && TerminalManagerTreeTypes.isTerminalNode(node)) {
+            this.treeWidget.model.selectNode(node);
+        }
+    }
```
colin-grant-work: MInor: this could be moved to the tree mode litself, since both the finding and the selecting happen there.

packages/terminal/src/browser/terminal-frontend-contribution.ts
```diff
@@ -982,7 +982,7 @@ export class TerminalFrontendContribution implements FrontendApplicationContribu
         });
     }
 
-    protected async selectTerminalProfile(placeholder: string): Promise<[string, TerminalProfile] | undefined> {
+    async selectTerminalProfile(placeholder: string): Promise<[string, TerminalProfile] | undefined> {
```
colin-grant-work: It doesn't look like this is actually accessed outside this class?

lucas-koehler: Good catch! I initially exposed it to use for new terminals using profiles. As this is now a follow-up, I'll revert this.

packages/terminal-manager/src/browser/terminal-manager-alert-dialog.tsx
lucas-koehler: Good catch! Upon reviewing it again, I'll remove the dialog completely because the existing ConfirmDialog of the core package is sufficient.

packages/terminal-manager/src/browser/terminal-manager-frontend-view-contribution.ts
lucas-koehler: MAkes sense! Thanks!

packages/terminal-manager/src/browser/terminal-manager-preferences.ts
lucas-koehler: Makes sense, I changed the prefix for both preferences to `terminal.grouping`. This one to `terminal.grouping.mode`.
I agree on the workspace scope: There is no reason not allow users to configure it separately if they'd like to.

packages/terminal-manager/src/browser/terminal-manager-tree-model.ts
lucas-koehler: Good point! We should definitely try to adhere to this as much as possible!

packages/terminal-manager/src/browser/terminal-manager-widget.ts
lucas-koehler: I made init sync and `createPageAndTreeLayout` and `resolveMainLayout` async instead

lucas-koehler:
Hi @colin-grant-work , thanks a lot for the extensive review! I addressed all your comments. Please have another look :)

ndoschek (approved):
Thanks! I tested it a little as well, LGTM 🎉

PR #16612, opened 2025-11-18
Use PreferenceService.get to merge agent settings from different scopes

### LINKED ISSUE #16601: AI Agents Settings - it uses only the user scope - can't use the a default application setting to control it
### Bug Description:
Hi,
I want to be able to disable some agents using the preference settings by default, using the application setting like this:

```
"theia": {
    "target": "browser",
    "frontend": {
      "config": {
             "AppTester": {
                 "enable": false
              },
              "ProjectInfo": {
                 "enable": false
              }
          }
      }
    }
}
```

But the theia agents settings service uses the user scope for load and save, and does not respect the other scope settings.

### Additional Information

- Operating System: Mac
- Theia Version: 1.66

colin-grant-work:
#### What it does

Fixes #16601 by switching from `PreferenceService.inspect` to `PreferenceService.get` when retrieving AI Agent settings. Whereas the `inspect` call resulted in our returning the value from the narrowest defined scope and ignoring all other scopes, `get` merges the settings from defined scopes.

#### How to test

1. In user-level settings, define something like

```json
     "AppTester": {
         "enable": false
      }
}
```

2. In workspace level settings, define something like:

```json
     "ProjectInfo": {
         "enable": false
      }
}
```

3. Observe that both agents are disabled - i.e. the two declarations have been merged, whereas on `master` the workspace-level setting would have clobbered the user-level setting, resulting in AppTester unexpectedly being enabled.

4. Add `enable: true` to workspace scope for `AppTester`. AppTester should then be enabled, i.e. the workspace-level value should override the user-level value.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/types.ts
+++ b/packages/core/src/common/types.ts
@@ -35,7 +35,7 @@ export type RecursiveReadonly<T> = {
     ? RecursiveReadonly<I>[]
     : T[P] extends object ? RecursiveReadonly<T[P]>
     : T[P];
-}
+};
 
 export function isBoolean(value: unknown): value is boolean {
     return value === true || value === false;


colin-grant-work:
@ndoschek, good catch. Objects returned by `.get` are deep frozen and immutable, but we were directly modifying them. I have modified the declarations and added a breaking change notice indicating that those objects are immutable and must be copied before mutation.

colin-grant-work:
@ndoschek, it appears the problem was that the deep equal check in the `setValue` method was failing to handle the possibility of the default value being undefined, which it is in this case. Sorry I missed the error, but the changes should now be persisted successfully.

ndoschek:
Thanks for the update @colin-grant-work!
The agent settings are now written to the user settings again as exptected!
But unfortunately I did notice another issue now: if a workspace-scope agent setting is present, all user-scope agent settings end up being written into the workspace scope as well if there is an update. On following updates some entries (e.g. `languagemodel` requirements) get multiplied in the workspace scope then.

https://github.com/user-attachments/assets/cacd69d0-95ad-4e43-96cd-667d5344dc1f

colin-grant-work:
> all user-scope agent settings end up being written into the workspace scope as well if there is an update.

This is expected, but can be adjusted. Basically, the `setValue` method aims to make preference changes that take effect, so if it detects that there is a definition of a given setting in a given narrow scope (workspace, in this case), it will make its changes in that scope in order to ensure that its changes take effect and aren't overridden by settings coming from a narrower scope. In general, that's good behavior, and it will work in this case, but may be unexpected. For this particular preference, since we're dealing with an object-type whose values are merged, things are a bit complex. We could:
- Keep the `setValue` behavior in place, with settings set in workspace scope if there are any settings defined there. The user is still free to set things in user scope themselves, if desired. (This is my recommendation)
- Set things in user scope by default, as before. I don't think this is a good idea, since it could mean that settings don't take effect if they're overridden in workspace scope.
- Unmerge the settings and do our own version of `setValue` by agent rather than by outermost preference key. At the moment, the full settings object will get written to the narrowest scope because the whole object is under the same `agentSettings` key, but of course not every agent may have a setting in every scope If we know which agent we're modifying, we can identify the narrowest scope in which settings for that agent are currently present and update them only there.

ndoschek:
Thanks Colin, that makes perfect sense, thanks for explaining this! :+1: 
Option 1 sounds good to me as well!

Issue #16609, opened 2025-11-18
[vscode] API evolution (public and proposed) and menu extension points between 1.105 and 1.106


ndoschek:
### Sources

| | Details |
|:-|:-|
| Commits compared | 1.105.0: [03c265b](https://github.com/microsoft/vscode/commit/03c265b1adee71ac88f833e065f7bb956b60550a) <br> 1.106.1: [cb1933b](https://github.com/microsoft/vscode/commit/cb1933bbc38d329b3595673a600fab5c7368f0a7) |
| Public API diff command | `git diff --name-status 03c265b cb1933b -- src/vscode-dts/` |
| Menu diff command | `git diff --quiet 03c265b cb1933b -- src/vs/workbench/services/actions/common/menusExtensionPoint.ts` |
| Built‑in extensions scan (on commit cb1933b) | `git submodule foreach "grep -irn enabledApiProposals --include=package.json \| cut -d ':' -f 1 \| xargs jq --raw-output '{filename: input_filename, enabledApiProposals: .enabledApiProposals}'" > ./vscode-builtins-enabled-proposals-1.106.1-master.txt` |

#### Public API (`vscode.d.ts`)

- several minor documentation updates
- Add new event to `DocumentRangeSemanticTokensProvider`: `onDidChangeSemanticTokens?: Event<void>`
- Add new property `idToken` to `AuthenticationSession` interface
- Update a function parameters in `LanguageModelChatMessage`, `LanguageModelChatResponse`, `LanguageModelToolResultPart`
- Update type defintions `LanguageModelResponsePart`, `LanguageModelInputPart`
- Add new class `LanguageModelDataPart`

#### Proposed API evolutions

| File name | Change | Comment |
|:-|:-|:-|
| vscode.proposed.chatContextProvider.d.ts | add | not used by builtins, ignored for now |
| vscode.proposed.chatParticipantAdditions.d.ts | update | not used by builtins, ignored for now |
| vscode.proposed.chatParticipantPrivate.d.ts | update | not used by builtins, ignored for now |
| vscode.proposed.chatSessionsProvider.d.ts | update | not used by builtins, ignored for now |
| vscode.proposed.idToken.d.ts | update | not used by builtins, ignored for now |
| vscode.proposed.contribSecondarySidebar.d.ts | update | not used by builtins, ignored for now |
| vscode.proposed.dataChannels.d.ts | update | not used by builtins, ignored for now |
[...]

PR #16625, opened 2025-11-21
[vscode] API evolution (public) to 1.106.1

ndoschek:
#### What it does

Part of GH-16609

- [vscode] chore: minor doc updates in theia.d.ts 
- [vscode] add new idToken property in AuthenticationSession 
- [vscode] add onDidChangeEvent in DocumentRangeSemanticTokensProvider 
- [vscode] add type LanguageModelDataPart 

#### How to test

- the new property and event are optional
- the new type LanguageModelDataPart is stubbed

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/main/browser/languages-main.ts
+++ b/packages/plugin-ext/src/main/browser/languages-main.ts
@@ -1176,14 +1176,23 @@ export class LanguagesMainImpl implements LanguagesMain, Disposable {
         }
     }
 
-    $registerDocumentRangeSemanticTokensProvider(handle: number, pluginInfo: PluginInfo, selector: SerializedDocumentFilter[], legend: theia.SemanticTokensLegend): void {
+    $registerDocumentRangeSemanticTokensProvider(handle: number, pluginInfo: PluginInfo, selector: SerializedDocumentFilter[], legend: theia.SemanticTokensLegend,
+        eventHandle: number | undefined): void {
         const languageSelector = this.toLanguageSelector(selector);
-        const provider = this.createDocumentRangeSemanticTokensProvider(handle, legend);
+        let event: Event<void> | undefined = undefined;
+        if (typeof eventHandle === 'number') {
+            const emitter = new Emitter<void>();
+            this.register(eventHandle, emitter);
+            event = emitter.event;
+        }
+        const provider = this.createDocumentRangeSemanticTokensProvider(handle, legend, event);
         this.register(handle, (monaco.languages.registerDocumentRangeSemanticTokensProvider as RegistrationFunction<monaco.languages.DocumentRangeSemanticTokensProvider>)
             (languageSelector, provider));
     }
 
-    protected createDocumentRangeSemanticTokensProvider(handle: number, legend: theia.SemanticTokensLegend): monaco.languages.DocumentRangeSemanticTokensProvider {
+    protected createDocumentRangeSemanticTokensProvider(
+        handle: number, legend: theia.SemanticTokensLegend, _onDidChangeEvent?: Event<void>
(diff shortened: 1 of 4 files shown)


packages/plugin-ext/src/main/browser/languages-main.ts
```diff
@@ -1202,7 +1211,8 @@ export class LanguagesMainImpl implements LanguagesMain, Disposable {
                     };
                 }
                 throw new Error('Unexpected');
-            }
+            },
+            // onDidChange property is not yet available with the current monaco version, probably with the next monaco update then
```
colin-grant-work: Use the [`@monaco-uplift`](https://github.com/eclipse-theia/theia/blob/master/doc/coding-guidelines.md#to-do-tags) tag to increase the chances this will be noticed on the next uplift.


```suggestion
            // @monaco-uplift onDidChange property is not yet available with the current monaco version, probably with the next monaco update then
```

packages/plugin-ext/src/plugin/languages.ts
```diff
@@ -875,9 +875,17 @@ export class LanguagesExtImpl implements LanguagesExt {
...
         legend: theia.SemanticTokensLegend, pluginInfo: PluginInfo): theia.Disposable {
+        const eventHandle = (typeof provider.onDidChangeSemanticTokens === 'function' ? this.nextCallId() : undefined);
+
         const handle = this.addNewAdapter(new DocumentRangeSemanticTokensAdapter(this.documents, provider));
-        this.proxy.$registerDocumentRangeSemanticTokensProvider(handle, pluginInfo, this.transformDocumentSelector(selector), legend);
-        return this.createDisposable(handle);
+        this.proxy.$registerDocumentRangeSemanticTokensProvider(handle, pluginInfo, this.transformDocumentSelector(selector), legend, eventHandle);
+        let result = this.createDisposable(handle);
+        if (eventHandle) {
+            // eslint-disable-next-line no-unsanitized/method
```
colin-grant-work: Nit: I don't think this comment is necessary here or in the other place it occurs in this file.

The code could be rearranged to make the non-null assertion on `onDidChangeSemanticTokens` unnecessary. This is actually significant, because currently, the `if (eventHandle)` check would fail for the first provider with an event, because `eventHandle` would be `0`, assuming nothing else had generated one in the meantime. We should ensure that we're not doing truthiness checks on those values elsewhere, or start the call ID's at > 0.


One idea:
```typescript
    registerDocumentRangeSemanticTokensProvider(selector: theia.DocumentSelector, provider: theia.DocumentRangeSemanticTokensProvider,
        legend: theia.SemanticTokensLegend, pluginInfo: PluginInfo): theia.Disposable {
        const handle = this.addNewAdapter(new DocumentRangeSemanticTokensAdapter(this.documents, provider));
        const unregister = this.createDisposable(handle);
        if (typeof provider.onDidChangeSemanticTokens === 'function') {
            const eventHandle = this.nextCallId();
            const subscription = provider.onDidChangeSemanticTokens!(_ => this.proxy.$emitDocumentSemanticTokensEvent(handle));
            this.proxy.$registerDocumentRangeSemanticTokensProvider(handle, pluginInfo, this.transformDocumentSelector(selector), legend, eventHandle);
            return Disposable.from(unregister, subscription);
        }
        this.proxy.$registerDocumentRangeSemanticTokensProvider(handle, pluginInfo, this.transformDocumentSelector(selector), legend, undefined);
        return unregister
    }
```

ndoschek: Thanks, that sounds reasonable!
I updated this and the other occurence in `registerDocumentSemanticTokensProvider`.

Issue #16645, opened 2025-11-25
Preference change listeners should handle undefined values for newValue properly


ndoschek:
Follow-up of GH-16617 (and GH-15635)

Preference change listeners in the code base do not consistently handle the case that a preference change can return a `change.newValue` being `undefined`. This might lead to different issues, as seen previously in the welcome page and SCM view.

Expected:
Listeners should not rely solely on `change.newValue` but instead query the current effective preference using `preferenceService.getValue(...)` if the updated preference value is needed.

This issue is a follow-up to fixes already made in the welcome page and SCM view.

PR #16663, opened 2025-11-26
16466: clarify role of Orchestrator agent in Theia AI chat interactions

CamilleLetavernier:
#### What it does

- improve the Chat UX for first time Theia AI users (when Theia AI is not fully configured)

#### How to test

- start from a fresh workspace
  - alternatively: unset all your API Keys (Anthropic, OpenAI, Gemini) and Ollama Configs (models, ...)
  - unset the default agent setting
- start a new conversation in AI Chat
  - the welcome page should indicate a configuration error, indicating that Models are not configured
  - the page also indicates the error messages reported by each model (Typically "missing API Key")
  - there is no error for Ollama, because Ollama no longer provides any model out of the box (we check "Language Models" rather than "Providers").
- add an API Key (can be a mock key) for OpenAI, Google or Anthropic (or any other provider)
  - should also work when adding a model for Ollama
- the Welcome Page changes: it now requests the user to pick a default Agent
  - this can be done either via one of the recommendations (single click) or by navigating to the settings via the suggested link and entering a custom value

Screen 1: AI Features are disabled (Not change for this screen)

image

Screen 2: AI Models are not configured (New screen :star2: )

image

Screen 3: Default Agent has not been selected (New screen :star2: )

image

Screen 4: Everything is (probably) fine, happy chatting! (No change for this screen)

image

#### Follow-ups

- Remove Orchestrator agent and related setting
- Align styling of all different chat views
- Update the preference change listener for the model entries for example in the general follow https://github.com/eclipse-theia/theia/issues/16645

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-ide/src/browser/ide-chat-welcome-message-provider.tsx
+++ b/packages/ai-ide/src/browser/ide-chat-welcome-message-provider.tsx
@@ -88,6 +88,7 @@ export class IdeChatWelcomeMessageProvider implements ChatWelcomeMessageProvider
     protected _hasReadyModels = false;
     protected _modelRequirementBypassed = false;
     protected _defaultAgent = '';
+    protected modelConfig: { hasModels: boolean; errorMessages: string[] } | undefined;
 
     protected readonly onStateChangedEmitter = new Emitter<void>();
 
@@ -125,6 +126,10 @@ export class IdeChatWelcomeMessageProvider implements ChatWelcomeMessageProvider
                 this.notifyStateChanged();
             })
         );
+        this.analyzeModelConfiguration().then(config => {
+            this.modelConfig = config;
+            this.notifyStateChanged();
+        });
         this.preferenceService.ready.then(() => {
             const defaultAgentValue = this.preferenceService.get(DEFAULT_CHAT_AGENT_PREF, '');
             const bypassValue = this.preferenceService.get(BYPASS_MODEL_REQUIREMENT_PREF, false);
@@ -137,6 +142,7 @@ export class IdeChatWelcomeMessageProvider implements ChatWelcomeMessageProvider
     protected async checkLanguageModelStatus(): Promise<void> {
         const models = await this.languageModelRegistry.getLanguageModels();
         this._hasReadyModels = models.some(model => model.status.status === 'ready');
+        this.modelConfig = await this.analyzeModelConfiguration();
         this.notifyStateChanged();
     }
 
@@ -213,23 +219,15 @@ Lean more in the [documentation](https://theia-ide.org/docs/user_ai/#chat).
     }
 
     protected renderModelConfigurationScreen(): React.ReactNode {
-        const ErrorContent = () => {
-            const [config, setConfig] = React.useState<{ hasModels: boolean; errorMessages: string[] }>(
-                { hasModels: false, errorMessages: [] }
-            );
+        const config = this.modelConfig ?? { hasModels: false, errorMessages: [] };
+        const { hasModels, errorMessages } = config;
 
-            React.useEffect(() => {
-                this.analyzeModelConfiguration().then(setConfig);
-            }, []);
-
-            const { hasModels, errorMessages } = config;
-
-            if (!hasModels) {
-                return <>
-                    <div className="theia-WelcomeMessage-ErrorIcon">⚠️</div>
-                    <LocalizedMarkdown
-                        localizationKey="theia/ai/ide/noLanguageModelProviders"
-                        defaultMarkdown={`
+        if (!hasModels) {
+            return <div className={'theia-WelcomeMessage'} key="setup-state">
+                <div className="theia-WelcomeMessage-ErrorIcon">⚠️</div>
+                <LocalizedMarkdown
+                    localizationKey="theia/ai/ide/noLanguageModelProviders"
+                    defaultMarkdown={`
 ## No Language Model Providers Available
 
 No language model provider packages are installed in this IDE.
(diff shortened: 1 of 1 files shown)


CamilleLetavernier:
I've rebased the PR on the current master, which included a refactoring of the Welcome Page from regular React to (Localized)Markdown: [issues/16470](https://github.com/eclipse-theia/theia/pull/16470)

packages/ai-ide/src/browser/ide-chat-welcome-message-provider.tsx
```diff
@@ -82,6 +211,168 @@ Lean more in the [documentation](https://theia-ide.org/docs/user_ai/#chat).
...
+                            {nls.localize('theia/ai/ide/openSettings', 'Open AI Settings')}
+                        </button>
+                        <button
+                            className="theia-button secondary"
+                            onClick={() => this.setModelRequirementBypassed(true)}>
+                            {nls.localize('theia/ai/ide/continueAnyway', 'Continue Anyway')}
+                        </button>
+                    </div>
+                    <small className="theia-WelcomeMessage-Hint">
+                        {nls.localize('theia/ai/ide/bypassHint', 'Some agents like Claude Code don\'t require Theia Language Models')}
```
ndoschek: A little more info here would be nice, that this sets the preference Bypass Model Requirement.
Maybe a tooltip would also be good here.

packages/ai-ide/src/browser/ide-chat-welcome-message-provider.tsx
```diff
@@ -82,6 +211,168 @@ Lean more in the [documentation](https://theia-ide.org/docs/user_ai/#chat).
...
+                    </div>
+                )}
+                <LocalizedMarkdown
+                    localizationKey="theia/ai/ide/commonFixes"
+                    defaultMarkdown={`
+**Common fixes:**
+
+- Check that your API key is valid and not expired
+- Verify that custom servers (Ollama, LlamaFile) are running
+- Check your network connection and firewall settings
```
ndoschek: This should also be left-aligned. If the UL content is centered, it looks off.

packages/ai-ide/src/browser/ide-chat-welcome-message-provider.tsx
```diff
@@ -82,6 +211,168 @@ Lean more in the [documentation](https://theia-ide.org/docs/user_ai/#chat).
...
+                    <div className="theia-WelcomeMessage-AgentButtons">
+                        {recommendedAgents.map(agent => (
+                            <button
+                                key={agent.id}
+                                className="theia-WelcomeMessage-AgentButton"
+                                onClick={() => this.setDefaultAgent(agent.id)}
+                                title={agent.description}>
+                                <span className="theia-WelcomeMessage-AgentButton-Icon">@</span>
+                                <span className="theia-WelcomeMessage-AgentButton-Label">{agent.label}</span>
+                            </button>
```
ndoschek, line 352: This is a little big, would be nice if we could get the same button size as on the other screen (open ui settings) to have a little consistency between the views.

packages/ai-ide/src/browser/ide-chat-welcome-message-provider.tsx
```diff
@@ -82,6 +211,168 @@ Lean more in the [documentation](https://theia-ide.org/docs/user_ai/#chat).
...
+                </>
+            ) : (
+                <div className="theia-WelcomeMessage-NoRecommendedAgents">
+                    <p>{nls.localize('theia/ai/ide/noRecommendedAgents', 'No recommended agents are available.')}</p>
+                </div>
+            )}
+            <LocalizedMarkdown
+                localizationKey="theia/ai/ide/moreAgentsAvailable"
+                defaultMarkdown={recommendedAgents.length > 0
+                    ? 'More agents are available. Use @AgentName to try others or configure a different default in [preferences]({0}).'
```
ndoschek: This box looks a little off, we should have a least the same padding as the input box on the bottom.
In general it would be great if we could reuse another theia concept, like the classes used in the [AlertMessage](https://github.com/eclipse-theia/theia/blob/master/packages/core/src/browser/widgets/alert-message.tsx).
E.g. in the scm view we have a warning alert message, but in this case we could use info level, wdyt?

Image

packages/ai-ollama/src/common/ollama-preferences.ts
```diff
@@ -30,7 +30,7 @@ export const OllamaPreferencesSchema: PreferenceSchema = {
         [MODELS_PREF]: {
             type: 'array',
             title: AI_CORE_PREFERENCES_TITLE,
-            default: ['llama3', 'gemma2'],
+            default: [],
```
ndoschek, line 33: Was this intentional?

packages/ai-ide/src/browser/style/index.css
ndoschek, line 1: In general I would like to double check the font sizes, margins and so on and use theia css variables where possible, but as mentioned in my general comment, it would also be fine for me to rework the styling in a follow up together with the other views we have.

ndoschek:
@JonasHelming could you also take a look at the flow in general and let us know if you’re fine with it or have any additional feedback? TIA!

packages/ai-ollama/src/common/ollama-preferences.ts
CamilleLetavernier: Yes; I thought I explained it in the PR description, but actually I only added a short entry in the commit message:

> - unset the Ollama default models (not relevant defaults; also makes it impossible to distinguish whether ollama is actually configured)

The problem is that in order to detect configuration issues, we're testing the state of each language model ("ready" or not). For the cloud-based providers, this means that the model's provider should have an API Key (we don't check the validity; only whether it exists or not).

However, for Ollama, we don't have or need an API Key. In fact, the Provider doesn't need any configuration; so all models are considered "ready" by default. As long as one model is "ready", the AI Chat will not display any configuration issue. So, Ollama was creating a false-positive with its out-of-the-box configuration here.

Moreover, these 2 models don't make that much sense as default values. At best they can be helpful "hints", but Ollama doesn't come with any model out-of-the-box. So, I opted to remove them. Now, if users want to use Ollama, they have to explicitly configure it, by adding the models they want to use.

The alternative would be to ping the server (to see if it exists), and maybe list the available models, but I wanted to avoid any startup network activity (so we're not checking if the OpenAI/Anthropic API Keys are valid, and we're not testing if the Ollama server actually exists either).

JonasHelming:
Thanks, this is a nice improvement for initial users!

1. We still have an orchestrator related setting
2. I was not able to get the "error" page, eve though I removed all AI related settings, so I could not test this
3. Feedback to the screenshot above:

I think we should make the message feel less like an exception, as users will very likely run into it.

Suggestion:

# Please configure at least one language model provider

If you want to use OpenAI, Anthropic or GoogleAI hosted models, please enter an API key in the settings. (maybe link the provider names to their settings page)
If you want to use another provider such as Ollama, please configure it in the settings and adapt agents or a model alias to use your custom model.
Note: Some agents, such as Claude Code do not need a provider to be configure, just continue in this case.
See [the documentation](https://theia-ide.org/docs/user_ai/) for more details.

# Current Configuration State
message from providers

packages/ai-ollama/src/common/ollama-preferences.ts
ndoschek: Ah thanks for the explanation, that sounds very reasonable! It is fine by me then! :+1:

CamilleLetavernier:
I've rebased the PR to main and addressed most of the review comments. In the meantime, the AI Configuration view was rewritten quite significantly, and I believe it introduced a regression, as the Welcome Message Provider was no longer notified when agents were enabled/disabled. I've patched the agent-service accordingly (it's not directly related to this PR; but I need this notification mechanism to track configuration changes and properly refresh the Welcome message).

> We still have an orchestrator related setting

This PR only removed the default agent, but didn't change anything about the Orchestrator itself. The Agent (and its settings) still exist.

> I was not able to get the "error" page, eve though I removed all AI related settings, so I could not test this

The Config warning message will go away if at least one provider is "properly" configured; and some providers are fully (or almost fully) configured out of the box (e.g. Ollama with its default values). So you also need to remove e.g. all Ollama models to see this configuration screen.

> Feedback to the screenshot above:
> I think we should make the message feel less like an exception, as users will very likely run into it.

:+1:

Issue #16640, opened 2025-11-24
Gemini models hang, return malformed tool calls, or fail with 503 errors during AI task execution


JonasHelming:
When using Gemini models with Theia AI task generation, execution is unreliable. In addition to long stalls (potentially rate-limit related), there are several recurring errors:

- MALFORMED_FUNCTION_CALL
- “Mixing Content and Parts is not supported”
- 503 “The model is overloaded. Please try again later.”

These errors occur even for simple interactions like reading the root README.md. Other models do not show similar issues.

For more details also see PR #16636:
[https://github.com/eclipse-theia/theia/pull/16636](https://github.com/eclipse-theia/theia/pull/16636?utm_source=chatgpt.com)

Steps to reproduce:

- Use Gemini (e.g., Gemini 3) as the model in Theia AI.
- Run a simple workflow, e.g., “@Coder check my root README for improvements.”
- Observe intermittent failures, malformed tool calls, or long hangs.

Expected:
Model responds consistently and tool calls are well-formed. Should work at least for all models we have in the default list, most importantly Gemini 3

Actual:
Requests often hang, return malformed function calls, or fail with 503 overload errors.

Notes:

Might be related to Gemini API behavior or rate limits.

Needs investigation to confirm whether the issue is in Theia AI handling, Coder agent logic, or the model itself.

JonasHelming:
Note: Gemini is very unstable for me today in the chat interface, so this might partly be an issue on their side, but not sure at all.

MatthewKhouzam:
Can you please post a typical curl request? I have just updated locally, and I needed to remove the "content" field to get it to work.

https://ai.google.dev/gemini-api/docs#rest shows the API, `content` is part of the OpenAI API, `parts` is part of the google api. 

My experience is that https://ai.google.dev/gemini-api/docs/openai works sometimes. 

This one is more stable. https://ai.google.dev/gemini-api/docs/gemini-3?thinking=high

Notice that `content` and `contents` are different. 

I would suggest maybe having a router for all OpenAI api, and having a helper that if you say gemini, will use this url
`https://generativelanguage.googleapis.com/v1beta/openai/chat/completions` as a backup if querying
`https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-flash:generateContent` fails.

Issue #16642, opened 2025-11-24
Improve how to enable thinking mode


eneufeld:
### Feature Description:
Currently you have to enable thinking in a quite complex way:
```
ai-features.modelSettings.requestSettings": [
  {
    "scope": {
      "modelId": "claude-sonnet-4-5-20250929",
      "agentId": "theia_dev"
    },
    "requestSettings": {
      "max_tokens": 40000,
      "temperature": 1,
      "thinking": {
        "type": "enabled",
        "budget_tokens": 20000
      }
    }
  }
]
```

Ideally we could have a dropdown to select how many token should be used for thinking/reasoning and the dropdown being disabled for non thinking models.

JonasHelming:
We should also set some reasonable default I guess

PR #16664, opened 2025-11-26
update @google/genai dependency, add thoughtSignature support and fix content extraction

ndoschek:
#### What it does

Resolves GH-16640

Update @google/genai dependency to 1.30.0

Following the @google/genai 1.30.0 SDK update, this commit adds proper support for Gemini's thinking feature and improves response handling.

- Enable thinking mode via thinkingConfig.includeThoughts for streaming
- Preserve thoughtSignature when converting tool calls to maintain conversation context across tool use cycles
- Extract text content from response parts instead of using chunk.text to correctly distinguish between thinking and regular content
- Fix functionResponse format to match Gemini API requirements name + response object, without id field)
- Add data field to ToolCall/ToolCallChatResponseContent for passing provider-specific metadata like thoughtSignature
- Handle MALFORMED_FUNCTION_CALL as warning instead of error
- Wrap non-object tool results in { result: ... } for Gemini compliance

#### How to test

- I tested with gemini 2.5-pro and gemini 3; gemini 3 performing better in my tests
  - slow performance from time to time for both of the models
- I did small coder requests like asking Coder to rename a variable and also longer requests like Jonas' request to add a reset token button.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-google/src/node/google-language-model.ts
+++ b/packages/ai-google/src/node/google-language-model.ts
@@ -13,7 +13,6 @@
 //
 // SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
 // *****************************************************************************
-import { webcrypto as crypto } from 'node:crypto';
 import {
     LanguageModel,
     LanguageModelRequest,
@@ -32,13 +31,30 @@ import { CancellationToken } from '@theia/core';
 import { GoogleGenAI, FunctionCallingConfigMode, FunctionDeclaration, Content, Schema, Part, Modality, FunctionResponse, ToolConfig } from '@google/genai';
 import { wait } from '@theia/core/lib/common/promise-util';
 import { GoogleLanguageModelRetrySettings } from './google-language-models-manager-impl';
+import { UUID } from '@theia/core/shared/@lumino/coreutils';
 
 interface ToolCallback {
     readonly name: string;
     readonly id: string;
-    readonly index: number;
     args: string;
 }
+/**
+ * Converts a tool call result to the Gemini FunctionResponse format.
+ * Gemini requires response to be an object, not an array or primitive.
+ */
+function toFunctionResponse(content: ToolCallResult): FunctionResponse['response'] {
+    if (content === undefined) {
+        return {};
+    }
+    if (Array.isArray(content)) {
+        return { result: content };
+    }
+    if (typeof content === 'object') {
+        return content as FunctionResponse['response'];
+    }
+    return { result: content };
+}
+
 const convertMessageToPart = (message: LanguageModelMessage): Part[] | undefined => {
     if (LanguageModelMessage.isTextMessage(message) && message.text.length > 0) {
         return [{ text: message.text }];
@@ -46,13 +62,13 @@ const convertMessageToPart = (message: LanguageModelMessage): Part[] | undefined
         return [{
             functionCall: {
                 id: message.id, name: message.name, args: message.input as Record<string, unknown>
-            }
+            },
+            thoughtSignature: message.data?.thoughtSignature,
         }];
     } else if (LanguageModelMessage.isToolResultMessage(message)) {
-        return [{ functionResponse: { id: message.tool_use_id, name: message.name, response: { output: message.content } } }];
-
+        return [{ functionResponse: { name: message.name, response: toFunctionResponse(message.content) } }];
     } else if (LanguageModelMessage.isThinkingMessage(message)) {
-        return [{ thought: true }, { text: message.thinking }];
+        return [{ thought: true, text: message.thinking }];
     } else if (LanguageModelMessage.isImageMessage(message) && ImageContent.isBase64(message.image)) {
         return [{ inlineData: { data: message.image.base64data, mimeType: message.image.mimeType } }];
     }
@@ -184,6 +200,10 @@ export class GoogleModel implements LanguageModel {
                             functionDeclarations
                         }]
                     }),
+                    thinkingConfig: {
+                        // https://ai.google.dev/gemini-api/docs/thinking#summaries
+                        includeThoughts: true,
+                    },
                     temperature: 1,
                     ...settings
                 },
@@ -195,7 +215,7 @@ export class GoogleModel implements LanguageModel {
         const asyncIterator = {
             async *[Symbol.asyncIterator](): AsyncIterator<LanguageModelStreamResponsePart> {
                 const toolCallMap: { [key: string]: ToolCallback } = {};
-                let currentContent: Content | undefined = undefined;
+                const collectedParts: Part[] = [];
                 try {
                     for await (const chunk of stream) {
                         if (cancellationToken?.isCancellationRequested) {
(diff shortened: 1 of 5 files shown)


JonasHelming:
I did not get the above mentioned errors anymore, but it hangs for longer runs.

# Longer runs
I get these warnings:

2025-11-26T22:39:02.138Z root WARN there are non-text parts functionCall in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.

2025-11-26T22:38:56.893Z root WARN there are non-text parts functionCall,thoughtSignature in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.

Test case for long runs: "Add a button to my token usage view to reset the token count"

# Short runs
First message works, but on the second I get:
{"error":{"message":"{\n \"error\": {\n \"code\": 400,\n \"message\": \"Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:getWorkspaceFileList` , position 2. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.\",\n \"status\": \"INVALID_ARGUMENT\"\n }\n}\n","code":400,"status":"Bad Request"}}

Test case for short runs:
1. Add a new line to my readme "fooba"
2. Translate the readme to spanish (fails)

ndoschek:
Hi again @JonasHelming, thanks for testing the first attempt!
I've now added a few more changes and did some additional testing myself. It already works much better now (tested with Gemini 2.5-Pro and 3-Pro-Preview) for both shorter and longer runs.  
Still, it would be great if you could test it again as well.

In general, the Gemini models feel slower to me compared to others, but I haven't used them very often in the past, so I don’t have a solid basis for comparison.
Also, I would appreciate it if you could take a look at the code too if it seems reasonable to you.

JonasHelming:
I will test this afternoon.

First try got the error below (which might be unrelated to us)

{"error":{"message":"{\n \"error\": {\n \"code\": 500,\n \"message\": \"An internal error has occurred. Please retry or report in https://developers.generativeai.google/guide/troubleshooting\",\n \"status\": \"INTERNAL\"\n }\n}\n","code":500,"status":"Internal Server Error"}}

JonasHelming:
@Coder add a button to the token usage view to reset the token count

With this prompt it still hangs and then I get the two warnings in the console:

Please refer to the non text parts for a full response from model.
2025-11-27T11:14:22.962Z root WARN there are non-text parts functionCall,thoughtSignature in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.
2025-11-27T11:14:23.321Z root WARN there are non-text parts functionCall in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.

I also have the feeling it takes very long, maybe it retries or thinks too much?

The simple "@Coder Add a new lint to the readme "fooba" 

1. ran into an endless loop for me
2. Works sometimes
3. Adds the line 3 times and then removes it again

I somehow have the feeling that there is smething wrong with the tool response, for example this would explain why it applies the change three times (it does not get feedback)?

ndoschek:
Review request was opened and already approved for the updated dependency:

[main] INFO A review request was created https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/25016 .

sdirix:
I have the same issue as @JonasHelming. When I send a non-trivial request to Gemini 3 with `@Coder`, it's basically non-functional: It will perform some tool calls but then just stop responding after a while with these warnings in the console

```
WARN there are non-text parts functionCall,thoughtSignature in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.
WARN there are non-text parts functionCall,thoughtSignature in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.
WARN there are non-text parts functionCall,thoughtSignature in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.
WARN there are non-text parts functionCall,thoughtSignature in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.
WARN there are non-text parts functionCall,thoughtSignature in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.
WARN there are non-text parts functionCall in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.
WARN there are non-text parts functionCall in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.
WARN there are non-text parts functionCall in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.
WARN there are non-text parts functionCall,thoughtSignature in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.
WARN there are non-text parts functionCall in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.
WARN there are non-text parts functionCall in the response, returning concatenation of all text parts. Please refer to the non text parts for a full response from model.
```

When using Gemini 2, I easily run into errors like this:

```
ERROR Error handling chat interaction: Error: Mixing Content and Parts is not supported, please group the parts into a the appropriate Content objects and specify the roles for them
    at tContents (/home/user/git/theia/examples/browser/lib/backend/vendors-node_modules_drivelist_build_Release_drivelist_node-dev-packages_application-package_-79e056.js:10101:19)
    at generateContentParametersToMldev (/home/user/git/theia/examples/browser/lib/backend/vendors-node_modules_drivelist_build_Release_drivelist_node-dev-packages_application-package_-79e056.js:16173:31)
    at Models.generateContentStreamInternal (/home/user/git/theia/examples/browser/lib/backend/vendors-node_modules_drivelist_build_Release_drivelist_node-dev-packages_application-package_-79e056.js:20326:26)
    at Models.generateContentStream (/home/user/git/theia/examples/browser/lib/backend/vendors-node_modules_drivelist_build_Release_drivelist_node-dev-packages_application-package_-79e056.js:19854:35)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async GoogleModel.withRetry (/home/user/git/theia/examples/browser/lib/backend/main.js:7609:24)
    at async GoogleModel.handleStreamingRequest (/home/user/git/theia/examples/browser/lib/backend/main.js:7373:24)
    at async [Symbol.asyncIterator] (/home/user/git/theia/examples/browser/lib/backend/main.js:7519:51)
    at async [Symbol.asyncIterator] (/home/user/git/theia/examples/browser/lib/backend/main.js:7521:42)
    at async /home/user/git/theia/examples/browser/lib/backend/main.js:6954:34
```

So sadly I don't think it really improves on the current state

ndoschek:
Thanks for testing!
This is really weird that it works for me then, I never saw the mentioned errors during my testing.
I had another look what might be different and I used the default coder prompt variant (coder-system-edit) during my tests apparently. With the next variant I now also got such issues.
Which variant did you both use?

JonasHelming:
Next variant!

But the default on master is now using the next variant, we made the new functions default

ndoschek:
Ah I see, I didn't reset my variant in my runtime yet, that would explain why it worked for me.
I'll convert this to a draft again then, thanks again for testing!

sdirix:
I don't think it's really Coder prompt related. The Gemini 2 models seem rather lazy to me. I invoked `@Coder` with the instruction to generate an implementation plan for a large feature. It then proceeded to not look at the code base once and just executed a single tool call to propose the implementation plan changes. A single tool call usually works fine, so I had to explicitly call it out to produce more tool calls and thereby the errors.

ndoschek:
What I can observe also: if I change the `suggestFileReplacements` to `simpleSuggestFileReplacements` (also different from `suggestFileReplacements_Simple`) in the coder default prompt it works fine again for me (except for the warning sin the backend, they're still there).
I'll have a look if I can spot something in that area.

ndoschek:
Summary of my findings so far:

I made a small rework to integrate thought summaries (temporarily), merge part text fields, and enable thought summaries for more insights. Also with this changes the warnings about ` Mixing Content and Parts is not supported` are gone now. Also I removed most of the previous changes again.

I'm still running into several issues with my Gemini model implementation:
- Short tasks work mostly fine (like adding a new line to a README), but it’s still a bit slow.

- Changing `suggestFileReplacements` to `simpleSuggestFileReplacements` (which is different from `suggestFileReplacements_Simple`) affected the behavior with the previous state of my changes, although with my latest changes there seems to be no difference anymore.

- Follow-ups to tasks are not working well at all: the model repeats too many steps and needs a long time to complete. The thinking summaries in the logs show this clearly. I'm not sure whether to double-check or adjust the function call parts based on Google’s function-calling documentation (see https://ai.google.dev/gemini-api/docs/function-calling?example=meeting, https://ai.google.dev/gemini-api/docs/thought-signatures#function-calling)

- Currently thought signatures are not handled explicitely, and I’m unsure whether we need to. The documentation says "Note: If you use the official Google Gen AI SDKs and use the chat feature (or append the full model response object directly to history), thought signatures are handled automatically. You do not need to manually extract or manage them, or change your code." (https://ai.google.dev/gemini-api/docs/thought-signatures), but not sure if this is the case for us.

- Retries on error runs always fail with an error saying that the function call is missing a thought signature:
   - `Function call is missing a thought_signature in functionCall parts… may lead to degraded model performance.`
   - This happens consistently, so I suspect that I'm missing some kind of reset or reinitialization before retrying.

@eneufeld @sdirix 
Could you have a look please as well if you can spot anything either in the implementation and/or the documentation? TIA

PS: @JonasHelming I think there is no immediate need to retest now, we can ping you again if there is more progress.

ndoschek:
Hi again,
together we could fix out a few problems due to updates in the genai API.
It now worked fine for me with the mentioned test runs in the PR description.

@eneufeld, @JonasHelming, @sdirix
Would be great if you could do a test run and give some feedback, TIA!

eneufeld (approved):
Change looks good to me. I tested it with TheiaDev and TheiaDevCoder. This worked fine.

packages/ai-google/src/node/google-language-model.ts
```diff
@@ -195,69 +215,97 @@ export class GoogleModel implements LanguageModel {
...
+                                if (part.text) {
+                                    if (part.thought) {
+                                        yield { thought: part.text, signature: part.thoughtSignature ?? '' };
+                                    } else {
+                                        yield { content: part.text };
+                                    }
+                                } else if (part.functionCall) {
+                                    const functionCall = part.functionCall;
+                                    // Handle function calls from Gemini
+                                    const callId = functionCall.id ?? crypto.randomUUID().replace(/-/g, '');
```
sdirix: - Do we really need to add an own random uuid? If yes, then an explanation why would be useful.
- Note that we have a shared uuid library in Theia, so you don't need to use the global crypto object

packages/ai-google/src/node/google-language-model.ts
```diff
@@ -184,6 +200,10 @@ export class GoogleModel implements LanguageModel {
                             functionDeclarations
                         }]
                     }),
+                    thinkingConfig: {
+                        // https://ai.google.dev/gemini-api/docs/thinking#summaries
+                        includeThoughts: true,
+                    },
```
sdirix, line 206: Do we set thinking on "on" on purpose? It's a bit cumbersome to turn it off again
```
"ai-features.modelSettings.requestSettings": [
  {
    "scope": {
      "providerId": "google"
    },
    "requestSettings": {
      "thinkingConfig": {
        "includeThoughts": false
      }
    }
  }
]
```

On the other hand, turning it on would be equally as difficult, so I am fine either way

packages/ai-google/src/node/google-language-model.ts
```diff
@@ -195,69 +215,97 @@ export class GoogleModel implements LanguageModel {
...
                         if (finishReason) {
-                            currentContent = chunk.candidates?.[0].content;
                             switch (finishReason) {
                                 // 'STOP' is the only valid (non-error) finishReason
                                 // "Natural stop point of the model or provided stop sequence."
                                 case 'STOP':
                                     break;
+                                // MALFORMED_FUNCTION_CALL: The model produced a malformed function call.
+                                // Log warning but continue - there might still be usable text content.
+                                case 'MALFORMED_FUNCTION_CALL':
```
sdirix, line 233: fyi: I did not encounter this case during testing

packages/ai-google/src/node/google-language-model.ts
```diff
@@ -195,69 +215,97 @@ export class GoogleModel implements LanguageModel {
...
-                                    // Update to existing tool call
-                                    toolCall.args = functionCall.args ? JSON.stringify(functionCall.args) : '{}';
-                                    yield {
-                                        tool_calls: [{
-                                            function: {
-                                                arguments: toolCall.args
-                                            }
-                                        }]
-                                    };
+                        // Handle thinking and text content from parts
```
sdirix: Minor: This comment is not fully correct as we also handle function calls here

packages/ai-google/src/node/google-language-model.ts
```diff
@@ -195,69 +215,97 @@ export class GoogleModel implements LanguageModel {
...
+                                collectedParts.push(part);
+                                if (part.text) {
+                                    if (part.thought) {
+                                        yield { thought: part.text, signature: part.thoughtSignature ?? '' };
+                                    } else {
+                                        yield { content: part.text };
+                                    }
+                                } else if (part.functionCall) {
+                                    const functionCall = part.functionCall;
+                                    // Handle function calls from Gemini
```
sdirix: Nitpick: This comment can be removed as it does not add value
```suggestion
```

packages/ai-google/src/node/google-language-model.ts
```diff
@@ -305,25 +348,27 @@ export class GoogleModel implements LanguageModel {
...
                         }));
                         const responseMessage: Content = { role: 'user', parts: toolResponses };
 
-                        const messages = [...(toolMessages ?? [])];
-                        if (currentContent) {
-                            messages.push(currentContent);
-                        }
-                        messages.push(responseMessage);
+                        // Build the model's response content from collected parts
+                        // Include all non-thinking parts to maintain proper conversation context
```
sdirix: This comment does not really explain anything. What we do here, is to remove all message parts which have a `thought`. Please adapt the comment to explain why this is needed.

packages/ai-google/src/node/google-language-model.ts
```diff
@@ -393,7 +431,17 @@ export class GoogleModel implements LanguageModel {
         }));
 
         try {
-            const responseText = model.text;
+            let responseText = '';
+            // Handle text content from parts
```
sdirix: Minor: This comment does not add any value. Maybe something like: For function declarations we are always only interested in text parts?

packages/ai-google/src/node/google-language-model.ts
ndoschek: Yes, I mainly added it to get to know the model better, but I agree it might not be for everyone. I'll create a follow-up to make it configurable via the UI (see GH-16642)

ndoschek:
Thanks for the reviews!
@sdirix I addressed most of your comments, for the thinking config I will create a follow up and keep it on for now.

packages/ai-google/src/node/google-language-model.ts
```diff
@@ -32,27 +31,44 @@ import { CancellationToken } from '@theia/core';
 import { GoogleGenAI, FunctionCallingConfigMode, FunctionDeclaration, Content, Schema, Part, Modality, FunctionResponse, ToolConfig } from '@google/genai';
 import { wait } from '@theia/core/lib/common/promise-util';
 import { GoogleLanguageModelRetrySettings } from './google-language-models-manager-impl';
+import { UUID } from '@theia/core/shared/@lumino/coreutils';
```
sdirix, line 34: I meant `generateUuid` from `@theia/core` ;)

ndoschek: Ah I see, thanks! I take a note that for the follow up then :)

Issue #16666, opened 2025-11-26
add support for custom hosted Anthropic models


udit-329:
There is already support for using open AI models with custom endpoints in Theia AI. However, that functionality doesn't exist for Anthropic models. In theory, we should be able to use the vercel sdk but I found it to not work very well.

udit-329:
I am working on the changes required to support custom anthropic models on my fork, I can open a PR once I am ready.

JonasHelming:
Sounds great!

PR #16673, opened 2025-11-27
add support for custom anthropic models

udit-329:
#### What it does

Adds support for custom anthropic models. This works similar to how user can configure custom openAI models for situations such as when the model is hosted on a custom url.

Addresses [#16666](https://github.com/eclipse-theia/theia/issues/16666)

#### How to test

If a user has an Anthropic model that is accessed from a custom endpoint, they can add it in settings.json similar to how it can already be done for openAI models.

```
"ai-features.AnthropicCustom.customAnthropicModels": [
        {
            "model": "claude-sonnet-4",
            "url": "<custom-url>",
            "id": "<model-ID>",
            "apiKey": "<API-Key>"
        }
    ]
```

#### Breaking changes

#### Review checklist

#### Reminder for reviewers


packages/ai-anthropic/src/browser/anthropic-frontend-application-contribution.ts
```diff
@@ -1,5 +1,5 @@
 // *****************************************************************************
-// Copyright (C) 2024 EclipseSource GmbH.
+// Copyright (C) 2024, 2025 EclipseSource GmbH.
```
sdirix: We don't update the year in the Theia project and just leave the header alone once a file is created.

packages/ai-anthropic/src/common/anthropic-preferences.ts
```diff
@@ -19,6 +19,7 @@ import { nls, PreferenceSchema } from '@theia/core';
 
 export const API_KEY_PREF = 'ai-features.anthropic.AnthropicApiKey';
 export const MODELS_PREF = 'ai-features.anthropic.AnthropicModels';
+export const CUSTOM_ENDPOINTS_PREF = 'ai-features.AnthropicCustom.customAnthropicModels';
```
sdirix:
```suggestion
export const CUSTOM_ENDPOINTS_PREF = 'ai-features.anthropicCustom.customAnthropicModels';
```

packages/ai-anthropic/src/common/anthropic-preferences.ts
```diff
@@ -38,5 +39,61 @@ export const AnthropicPreferencesSchema: PreferenceSchema = {
                 type: 'string'
             }
         },
+        [CUSTOM_ENDPOINTS_PREF]: {
+            type: 'array',
+            title: AI_CORE_PREFERENCES_TITLE,
+            markdownDescription: nls.localize('theia/ai/anthropic/useResponseApi/mdDescription',
```
sdirix:
```suggestion
            markdownDescription: nls.localize('theia/ai/anthropic/customEndpoints/mdDescription',
```

packages/ai-anthropic/src/node/anthropic-language-models-manager-impl.ts
```diff
@@ -114,9 +116,13 @@ export class AnthropicLanguageModelsManagerImpl implements AnthropicLanguageMode
...
     /**
-     * Returns the status for a language model based on the presence of an API key.
+     * Returns the status for a language model based on the presence of an API key or custom url.
      */
-    protected getStatusForApiKey(effectiveApiKey: string | undefined): LanguageModelStatus {
+    protected calculateStatus(modelDescription: AnthropicModelDescription, effectiveApiKey: string | undefined): LanguageModelStatus {
+        // Always mark custom models (models with url) as ready for now as we do not know about API Key requirements
+        if (modelDescription.url) {
+            return { status: 'ready' };
+        }
```
sdirix, line 125: This change is not sufficient. Within the AnthropicModel we check for the api key and if it does not exist, the model will immediately throw an error before even trying to use the SDK, see [here](https://github.com/eclipse-theia/theia/blob/0492714559815512cd60fa071f8f2c03d152f460/packages/ai-anthropic/src/node/anthropic-language-model.ts#L434-L437).
If we allow apiKey-less models to report ready, then there must be a successful path for them.

Please check what is required for this. In the OpenAI SDK we always need to hand over a key, so for keyless models we hand over a "stub key". I don't know whether this is also required for the Anthropic SDK or whether it actually supports not handing over a key if the endpoint does not need one.

ndoschek:
Hi @udit-329, just a gentle ping, the review is done and there are a few points that would need your input before we can proceed. TIA!

packages/ai-anthropic/src/node/anthropic-language-models-manager-impl.ts
udit-329: Anthropic SDK also always requires an API Key or an auth token (both can either be passed directly or set as environment variables). I added a similar stub for anthropic model when no key is present, this ensures the SDK doesn't throw an error when it shoudn't.

sdirix (approved):
Works great for me. Thanks! ❤️

PR #16681, opened 2025-11-28
Add support for "lazy" debug variables

### LINKED ISSUE #16591: Support "lazy" flag from VariablePresentationHint of debug variable from DAP
### Feature Description:

DAP allows to set a [variable presentation hint](https://microsoft.github.io/debug-adapter-protocol/specification#Types_VariablePresentationHint) which contains `lazy` option. This option allows to request heavy expensive values in a separate request to debug adapter.

It seems Thiea does not support this feature and moreover, there's no any options to easily customize creation of variable items in debug panel, there're no any factories which could be overriden and used via injection.

### Proposal

Support `lazy` flag of VariablePresentationHint, use specific UI for such variables and allow to create separate requests for variable value with 'lazy' flag.

pisv:
#### What it does

Closes #16591.

As described in #16591, debug adapters can return "lazy" variables, such as getters, which will not have their real value retrieved until the user clicks a button. Currently, Theia ignores the lazy flag and does not provide special UI for such variables. They are presented in a generic way as "placeholder" variables that the user needs to expand to see the child "resolved" variable. Such indirection can make it difficult to grasp the effective data structure:

This PR implements special support for "lazy" debug variables similarly to how it is done in VS Code.

Instead of expanding such variables, the user will need to click on a special button (`codicon-eye`) on the right hand side of the variable name. The button will disappear after clicking it, and the variable will be resolved. If the result is a simple variable, its value will be shown in place of the initial ("placeholder") value of the lazy variable. If the result is a structured variable, its value will be shown in place of the initial ("placeholder") value and its children will be revealed by auto-expanding the lazy variable.

This feature is supported in all Debug views where a variable can appear (Variables, Watch, Debug Console, Debug Hover):

Also, this PR adds support for the `debug.autoExpandLazyVariables` preference (also present in VS Code), which controls whether "lazy" variables are automatically resolved by the debugger (turned off by default).

#### How to test

This can be tested e.g. using a simple Node.JS program:

```js
class Foo {
  get bar() {
    return new Bar();
  }
  get x() {
    return 1;
  }
}
class Bar {
  a = 'a';
  b = 'b';
}
const foo = new Foo();
console.log(foo);

```

1. Set a breakpoint at line `console.log(foo);` and launch the program.
2. Expand the `foo` variable in the Debug Variables view.
3. Click on the resolve (eye) button on the right hand side of the name of the `bar` variable inside `foo`. The variable should be automatically expanded to reveal its resolved structure.
4. Click on the resolve (eye) button on the right hand side of the name of the `x` variable inside `foo`. The resolved value of the variable should be shown.
5. Add the `foo` variable to Watch, and repeat steps 3-4.
6. Evaluate `foo` in Debug Console, and repeat steps 3-4.
7. Repeat steps 3-4 in Debug Hover for `foo`.
8. Turn on the `debug.autoExpandLazyVariables` preference and relaunch the program. Verify that `bar` and `x` variables are now automatically resolved inside `foo`, and no resolve (eye) button is shown for them in Variables, Watch, Debug Console, and Debug Hover.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/editor/debug-inline-value-decorator.ts
+++ b/packages/debug/src/browser/editor/debug-inline-value-decorator.ts
@@ -194,7 +194,7 @@ export class DebugInlineValueDecorator implements FrontendApplicationContributio
                                     }
                                     if (expr) {
                                         const expression = new ExpressionItem(expr, () => stackFrame.thread.session);
-                                        await expression.evaluate('watch');
+                                        await expression.evaluate('watch', false);
                                         if (expression.available) {
                                             text = this.formatInlineValue(expr, expression.value);
                                         }
--- a/packages/debug/src/common/debug-preferences.ts
+++ b/packages/debug/src/common/debug-preferences.ts
@@ -64,7 +64,17 @@ export const debugPreferencesSchema: PreferenceSchema = {
             description: nls.localizeByDefault('Show Source Code in Disassembly View.'),
             type: 'boolean',
             default: true,
-        }
+        },
+        'debug.autoExpandLazyVariables': {
+            type: 'string',
+            enum: ['on', 'off'],
+            default: 'off',
+            enumDescriptions: [
+                nls.localizeByDefault('Always automatically expand lazy variables.'),
+                nls.localizeByDefault('Never automatically expand lazy variables.')
+            ],
+            description: nls.localizeByDefault('Controls whether variables that are lazily resolved, such as getters, are automatically resolved and expanded by the debugger.')
+        },
     }
 };
 
@@ -76,6 +86,7 @@ export class DebugConfiguration {
     'debug.showInStatusBar': 'never' | 'always' | 'onFirstSessionStart';
     'debug.confirmOnExit': 'never' | 'always';
     'debug.disassemblyView.showSourceCode': boolean;
+    'debug.autoExpandLazyVariables': 'on' | 'off';
 }
 
 export const DebugPreferenceContribution = Symbol('DebugPreferenceContribution');
--- a/packages/debug/src/browser/console/debug-console-items.tsx
+++ b/packages/debug/src/browser/console/debug-console-items.tsx
@@ -16,7 +16,7 @@
 
 import * as React from '@theia/core/shared/react';
 import { DebugProtocol } from '@vscode/debugprotocol/lib/debugProtocol';
-import { SingleTextInputDialog } from '@theia/core/lib/browser';
+import { codicon, SingleTextInputDialog } from '@theia/core/lib/browser';
 import { ConsoleItem, CompositeConsoleItem } from '@theia/console/lib/browser/console-session';
 import { DebugSession, formatMessage } from '../debug-session';
 import { Severity } from '@theia/core/lib/common/severity';
@@ -38,6 +38,7 @@ export class ExpressionContainer implements CompositeConsoleItem {
     protected variablesReference: number;
     protected namedVariables: number | undefined;
     protected indexedVariables: number | undefined;
+    protected presentationHint: DebugProtocol.VariablePresentationHint | undefined;
     protected readonly startOfVariables: number;
 
     constructor(options: ExpressionContainer.Options) {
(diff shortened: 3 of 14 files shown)


packages/debug/src/common/debug-preferences.ts
```diff
@@ -64,7 +64,18 @@ export const debugPreferencesSchema: PreferenceSchema = {
...
+            type: 'string',
+            enum: ['auto', 'on', 'off'],
+            default: 'auto',
+            enumDescriptions: [
+                nls.localizeByDefault('When in screen reader optimized mode, automatically expand lazy variables.'),
+                nls.localizeByDefault('Always automatically expand lazy variables.'),
+                nls.localizeByDefault('Never automatically expand lazy variables.')
+            ],
+            description: nls.localizeByDefault('Controls whether variables that are lazily resolved, such as getters, are automatically resolved and expanded by the debugger.')
+        },
```
sgraband, line 77: As far as i understand it the auto option is currently not implemented, right? I would then prefer to not have it available at all and remove it for now.

packages/debug/src/browser/editor/debug-inline-value-decorator.ts
```diff
@@ -194,7 +194,7 @@ export class DebugInlineValueDecorator implements FrontendApplicationContributio
                                     }
                                     if (expr) {
                                         const expression = new ExpressionItem(expr, () => stackFrame.thread.session);
-                                        await expression.evaluate('watch');
+                                        await expression.evaluate('watch', false);
```
sgraband, line 197: Personally, i find this code confusing. I know what it does, but it is strange that this is the only place where we need to set the resolveLazy to false. I believe the code would actually be more intuitive if the default value for resolveLazy in the evaluate method would be false and we explicitly set it to true when needed. It is "more code" but more intuitive imho as the default value of the option is also off/false. WDYT?

pisv:
@sgraband Thanks a lot for your review and comments! 👍

> I feel like we need to highlight the eye button a bit more
> VSCode just make the font a bit brighter/darker (depending on the theme). Could you check if this is easily possible?

Good catch, thank you! I didn't notice it initially. Looks like `color: var(--theia-icon-foreground)` needs to be added to the styles of the lazy-button. I will fix that in a moment.

packages/debug/src/common/debug-preferences.ts
pisv: Yes, I also thought about it, and actually implemented it first with just the 'on' and 'off' enum values, but later decided to fully align it with VS Code, and added the 'auto' value as the default, although it could not be supported currently in Theia in the same way as it is in VS Code. I'm a bit concerned that if we remove it for now, we'll probably have to change the default value in the future, but if you think it is OK, I will do it.

packages/debug/src/browser/editor/debug-inline-value-decorator.ts
pisv: I see, but that's just how it works (just as in VS Code): we always resolve lazy *expressions* (not variables!), irrespective of the `debug.autoExpandLazyVariables` preference, except for this particular case of inline values.

packages/debug/src/common/debug-preferences.ts
sgraband: I do not see switching the default value as an issue, especially as we should use off as a default, which means that only users with the screen reader enabled, will notice a difference if we switch the default values to auto. And that should be an improvement, which is fine imho.

packages/debug/src/browser/editor/debug-inline-value-decorator.ts
sgraband: I see, but i was not talking about switching the behaviour of the code, just the switching the default behavior of the `evaluate` method to `resolveLazy = false ` and then adjusting all the calls to do the same as before, just that the ones that did not specify anything need to specify true and the ones that specified false need to provide nothing, if that makes sense. The behavior should not change.

pisv: But would not then the default value no longer be "default"? I mean that if we didn't have this case of inline values, we would not even need the `resolveLazy` parameter, since in all other cases, lazy expressions are resolved immediately.

If anything, I'd much prefer then to not have the default value for this parameter at all, although I find it a bit strange, TBH, as in VS Code they also have implemented this parameter with the same default logic.

sgraband: Conviced :slightly_smiling_face: If you look at it this way it makes sense. Disregard my comment.

pisv:
Done! :-) I tried to address review comments and also rebased this PR on master (there was a conflict with my other PR).

sgraband (approved):
Perfect thank you! LGTM :+1:

pisv:
@sgraband Great to hear that! Thank you! 👍

PR #16700, opened 2025-12-03
Adds breakpoint actions to debug view

sgraband:
#### What it does
Implements breakpoint actions (edit, remove) to the debug view, including data, function, instruction and source breakpoints. This provides users with a more intuitive way to manage breakpoints.

#### How to test
image

Try out the new buttons in the Breakpoints view and test if they work as expected.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/debug/src/browser/model/debug-data-breakpoint.tsx
+++ b/packages/debug/src/browser/model/debug-data-breakpoint.tsx
@@ -45,11 +45,13 @@ export class DebugDataBreakpoint extends DebugBreakpoint<DataBreakpoint> {
     }
 
     protected doRender(): React.ReactNode {
-        return <span className="line-info theia-data-breakpoint" title={this.origin.info.description}>
-            <span className="name">{this.origin.info.description}</span>
-            <span className="theia-TreeNodeInfo theia-access-type" >{this.getAccessType()}</span>
+        return <>
+            <span className="line-info theia-data-breakpoint" title={this.origin.info.description}>
+                <span className="name">{this.origin.info.description}</span>
+                <span className="theia-TreeNodeInfo theia-access-type" >{this.getAccessType()}</span>
+            </span>
             {this.renderActions()}
-        </span>;
+        </>;
     }
 
     protected renderActions(): React.ReactNode {
--- a/packages/debug/src/browser/breakpoint/breakpoint-manager.ts
+++ b/packages/debug/src/browser/breakpoint/breakpoint-manager.ts
@@ -329,7 +329,7 @@ export class BreakpointManager extends MarkerManager<SourceBreakpoint> {
 
     removeDataBreakpoint(id: string): void {
         const index = this.dataBreakpoints.findIndex(bp => bp.id === id);
-        if (~index) { return; }
+        if (index < 0) { return; }
         const removed = this.dataBreakpoints.splice(index);
         this.fireOnDidChangeMarkers(BreakpointManager.DATA_URI);
         this.onDidChangeDataBreakpointsEmitter.fire({ uri: BreakpointManager.DATA_URI, added: [], removed, changed: [] });


packages/debug/src/browser/model/debug-data-breakpoint.tsx
```diff
@@ -16,6 +16,8 @@
 
 import { nls } from '@theia/core';
 import * as React from '@theia/core/shared/react';
+// import { codicon, TREE_NODE_SEGMENT_GROW_CLASS, DISABLED_CLASS } from '@theia/core/lib/browser';
```
ndoschek:
```suggestion
```

packages/debug/src/browser/style/index.css
```diff
@@ -251,6 +254,28 @@
...
+.theia-debug-breakpoint-actions {
+  display: flex;
+  align-items: center;
+  opacity: 0;
+  transition: opacity 0.1s;
+}
+
+.theia-source-breakpoint:hover .theia-debug-breakpoint-actions {
+  opacity: 1;
+}
```
ndoschek: This way we do not reserve the space for the actions and could avoid this gap when the widget is narrow or the breakpoint name and description are very long:

Image

```suggestion
.theia-debug-breakpoint-actions {
  display: none;
}

.theia-source-breakpoint:hover .theia-debug-breakpoint-actions {
  display: flex;
  align-items: center;
}
```

packages/debug/src/browser/style/index.css
```diff
@@ -251,6 +254,28 @@
...
+.theia-debug-breakpoint-actions > .codicon {
+  margin-left: 4px;
+  cursor: pointer;
+  padding: 2px;
+  border-radius: 2px;
+}
+
+.theia-debug-breakpoint-actions > .codicon:hover {
+  background: var(--theia-toolbar-hoverBackground);
+}
```
ndoschek: As we already use the `action-label` class on the codicons we can drop these two extra rules here.

packages/debug/src/browser/debug-frontend-application-contribution.ts
```diff
@@ -65,386 +65,10 @@ import { DebugHoverWidget } from './editor/debug-hover-widget';
 import { DebugExpressionProvider } from './editor/debug-expression-provider';
 import { AddOrEditDataBreakpointAddress } from './breakpoint/debug-data-breakpoint-actions';
```
ndoschek:
```suggestion
```

ndoschek:
Thank you for the updates @sgraband! :+1: 

I just noticed one further issue (not sure if it's new though):
While the session is running, the edit action does not work and logs the following error, the delete action works though.
As soon as i terminate the session the edit also works again as expected.
```
debug-source-breakpoint.tsx:170 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'executeCommand')
    at DebugSourceBreakpoint.onEdit (debug-source-breakpoint.tsx:170:29)
```

packages/debug/src/browser/model/debug-data-breakpoint.tsx
```diff
@@ -47,9 +48,21 @@ export class DebugDataBreakpoint extends DebugBreakpoint<DataBreakpoint> {
         return <span className="line-info theia-data-breakpoint" title={this.origin.info.description}>
             <span className="name">{this.origin.info.description}</span>
             <span className="theia-TreeNodeInfo theia-access-type" >{this.getAccessType()}</span>
+            {this.renderActions()}
         </span>;
```
ndoschek: This is incorrectly nested, if we make sure the actions are not children of the span element, they will not jump to the next line:
```suggestion
        return <>
            <span className="line-info theia-data-breakpoint" title={this.origin.info.description}>
                <span className="name">{this.origin.info.description}</span>
                <span className="theia-TreeNodeInfo theia-access-type" >{this.getAccessType()}</span>
            </span>
            {this.renderActions()}
        </>;
```

ndoschek (changes requested):
Thanks for the updates @sgraband!
I just found a small issue with the data breakpoint, the remove does not work properly, but this could be fixed in the `debug/src/browser/breakpoint/breakpoint-manager.ts` I think, the bitwise operator is not the correct check here.
Image

Issue #16708, opened 2025-12-04
AI agent settings not persisted when settings.json is missing agentSettings object


ndoschek:
### Bug Description:

When attempting to set values under `agentSettings` in a fresh project where the `agentSettings` object does not yet exist in `settings.json`, the operation fails silently as the error is caught, only a warning is logged:
`2025-12-03T10:36:09.151Z root WARN Updating the preferences was unsuccessful: TypeError: Cannot use 'in' operator to search for 'Architect' in undefined`

Probably a regression from GH-16612

### Steps to Reproduce:

1. Make sure your usesr `settings.json` does not contain an `agentSettings` object.
2. Open the AI configuration view (Alt+a) and try to update any setting for an Agent, e.g. set a new language model.
3. Observe the UI does not report any error, but the warning is logged and the settings are not persisted in the user settings.json file.

### Additional Information

- Operating System: all
- Theia Version: > 9c02ac6823b731519d4ef56076577c325c48be84

PR #16709, opened 2025-12-04
fix: use undefined-safe deep equality check in preference updates

ndoschek:
#### What it does

Replace JSONExt.deepEqual with PreferenceUtils.deepEqual to ensure undefined inspect value can be handled, since undefined is not a valid JSON value, but preference values can frequently be undefined in practice.

PreferenceUtils.deepEqual properly handles undefined by checking for it explicitly before delegating to JSONExt.deepEqual.

Fixes GH-16708

#### How to test

Please follow the steps mentioned in the issue

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/preferences/preference-service.ts
+++ b/packages/core/src/common/preferences/preference-service.ts
@@ -16,7 +16,7 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { JSONExt, JSONValue } from '@lumino/coreutils';
+import { JSONValue } from '@lumino/coreutils';
 import { inject, injectable, postConstruct } from 'inversify';
 import { Disposable, DisposableCollection, Emitter, Event, deepFreeze, unreachable } from '../../common';
 import { Deferred } from '../../common/promise-util';
@@ -541,7 +541,7 @@ export class PreferenceServiceImpl implements PreferenceService {
                 || (
                     scopesToChange.length === 1
                     && scopesToChange[0] === PreferenceScope.User
-                    && inspection.defaultValue !== undefined && JSONExt.deepEqual(value, inspection.defaultValue)
+                    && PreferenceUtils.deepEqual(value, inspection.defaultValue)
                 );
             const effectiveValue = isDeletion ? undefined : value;
             await Promise.all(scopesToChange.map(scope => this.set(preferenceName, effectiveValue, scope, resourceUri)));
@@ -549,7 +549,7 @@ export class PreferenceServiceImpl implements PreferenceService {
     }
 
     protected getScopesToChange(inspection: PreferenceInspection<any>, intendedValue: any): PreferenceScope[] {
-        if (JSONExt.deepEqual(inspection.value, intendedValue)) {
+        if (PreferenceUtils.deepEqual(inspection.value, intendedValue)) {
             return [];
         }
 


ndoschek:
Hi @colin-grant-work, could you please have a look when you have some time if you see any issues with this change? TIA

packages/ai-core/src/browser/ai-settings-service.ts
```diff
@@ -45,9 +45,9 @@ export class AISettingsServiceImpl implements AISettingsService {
 
     async updateAgentSettings(agent: string, agentSettings: Partial<AgentSettings>): Promise<void> {
         const settings = await this.getSettings();
-        const toSet = { ...settings, [agent]: { ...settings[agent], ...agentSettings } };
+        const updatedSettings = { ...settings, [agent]: { ...settings[agent], ...agentSettings } };
         try {
-            await this.preferenceService.updateValue(AISettingsServiceImpl.PREFERENCE_NAME, toSet);
+            await this.preferenceService.set(AGENT_SETTINGS_PREF, updatedSettings, PreferenceScope.User);
```
colin-grant-work: If we always set them in user scope, then we will have trouble with settings from workspace scope in the following scenario:
- Disable an agent in workspace scope preferences.
- Open AI configuration and enable the agent.
- `enable: true` will be set in user-scope preferences, but the workspace scope `enable: false` should win --> Action in the AI configuration fails to take effect.

colin-grant-work: However, it looks like we _don't_ have a problem in this scenario, which means we _do_ have a bug in how settings are being combined.

colin-grant-work: Actually, it looks like the reason we don't have a problem in that scenario is that we have two sources of truth, which also isn't good. On the one hand, we have the preferences, which declare whether agents should be enabled or disabled, but on the other hand, we have the `AgentServiceImpl`, which keeps a separate record and does not respond to preference changes. So what it sees is the action to enable, and it sets the agent as enabled, but then if you ask the preferences, the preferences will say the agent should be disabled. Those states shouldn't be inconsistent.

ndoschek: Thanks for the insights, that is indeed a problem then.
As discussed offline, I fixed the origin of this immediate issue in the `updateValue` function itself and reverted the usage of the `set` function again.

packages/core/src/common/preferences/preference-service.ts
```diff
@@ -541,15 +541,15 @@ export class PreferenceServiceImpl implements PreferenceService {
                 || (
                     scopesToChange.length === 1
                     && scopesToChange[0] === PreferenceScope.User
-                    && inspection.defaultValue !== undefined && JSONExt.deepEqual(value, inspection.defaultValue)
+                    && inspection.defaultValue !== undefined && PreferenceUtils.deepEqual(value, inspection.defaultValue)
```
colin-grant-work: The undefined check here was just to get around the limitations of `JSONExt.deepEqual`

```suggestion
                    && PreferenceUtils.deepEqual(value, inspection.defaultValue)
```

ndoschek: Ah good catch, thanks!

Issue #16722, opened 2025-12-08
ai-chat-ui: Highlight referenced tools in users' chat messages


lucas-koehler:
### Feature Description:

In the AI Chat window, tool functions are not highlighted in any way after the message was sent. In contrast, agents and variables are highlighted and hovering over them shows their description.
The same behavior should be introduced for tool functions. This makes it easier for users to see referenced tool functions in sent messages and enables them to read the description for additional clarification.

The screenshot shows the current state of the `Universal` agent being highlighted while the `getFileContent` tool request is just rendered as plain text.

<img width="620" height="126" alt="Image" src="https://github.com/user-attachments/assets/76ff52ab-9ea1-4b20-be19-8f3174b0067a" />

PR #16736, opened 2025-12-10
Serialize and restore ParsedChatRequest in chat sessions

### LINKED ISSUE #16561: AI Persistence: Serialize and Restore ParsedRequest
### Description

Currently, the `ParsedRequest` data (which contains parsed agent references, file references, and other structured request metadata) is not serialized and restored. After restoration, sessions lose this structured information.

The [`message`](https://github.com/eclipse-theia/theia/blob/7946ca874b04e6249d883e9a586f193194df1365/packages/ai-chat/src/common/chat-model.ts#L268) field in `ChatRequestModel` holds a `ParsedChatRequest` but is not currently serialized in [`toSerializable()`](https://github.com/eclipse-theia/theia/blob/7946ca874b04e6249d883e9a586f193194df1365/packages/ai-chat/src/common/chat-model.ts#L1640-L1650).

### Current Behavior

When a session is restored, request text is preserved but the parsed structure is lost.

### Desired Behavior

- Serialize `ParsedRequest` along with request text in `MutableChatRequestModel.toSerializable()`
- Restore `ParsedRequest` when loading persisted sessions in the [constructor](https://github.com/eclipse-theia/theia/blob/7946ca874b04e6249d883e9a586f193194df1365/packages/ai-chat/src/common/chat-model.ts#L1469-L1491)
- Preserve parsed data as much as possible

### Technical Considerations

- Add `toSerializable()` method to `ParsedRequest` or relevant classes
- Add deserialization logic in `MutableChatRequestModel` constructor for `ParsedRequest` data

### Related

- Follow up from initial implementation #16486

lucas-koehler:
#### What it does

Fixes #16561

Implement serialization of ParsedChatRequest data to preserve structured information (agent references, variable references, tool requests) across session persistence and restoration.

Previously, only raw text was preserved after session restoration, causing loss of parsed structure data. This affected UI rendering and prevented proper display of agent names, variables, and tool references in restored chat sessions.

Main code changes:

- Add serialization interfaces for ParsedChatRequest components in chat-model-serialization.ts (SerializableParsedRequestPart, SerializableToolRequest, SerializableResolvedVariable, etc.)
- Implement toSerializable() methods on ParsedChatRequestTextPart, ParsedChatRequestVariablePart, ParsedChatRequestFunctionPart, and ParsedChatRequestAgentPart classes
- Update MutableChatRequestModel to serialize/deserialize parsedRequest data via toSerializable() and deserializeParsedRequest() methods
- Serialize tool functions as id only and restore from the registry. For this, MutableChatModel and MutableChatRequestModel now require the tool registry as constructor parameter
- Add comprehensive test coverage for serialization/deserialization of all part types, including edge cases and complex mixed requests

[16561-pr.webm](https://github.com/user-attachments/assets/5b95201f-f6a9-420d-9cfa-5d290fec321f)

#### How to test

- Use the AI chat and chat using an agent and variable reference
- Restart theia
- Restore the session
- Observe: The agent, variable and tool function are now highlighted and provide additional information on hover

~**NOTE:** While the serialization also supports functions/tool requests, highlighting them does not work yet in general. See #16722~

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/common/chat-model.ts
+++ b/packages/ai-chat/src/common/chat-model.ts
@@ -1601,7 +1601,7 @@ export class MutableChatRequestModel implements ChatRequestModel, EditableChatRe
                     if (partData.variableValue !== undefined) {
                         varPart.resolution = {
                             variable: {
-                                id: partData.variableId ?? 'restored',
+                                id: partData.variableId,
                                 name: partData.variableName,
                                 description: partData.variableDescription
                             },
--- a/packages/ai-chat/src/common/chat-model-serialization.ts
+++ b/packages/ai-chat/src/common/chat-model-serialization.ts
@@ -48,7 +48,7 @@ export interface SerializableTextPart extends SerializableParsedRequestPartBase 
 
 export interface SerializableVariablePart extends SerializableParsedRequestPartBase {
     kind: 'var';
-    variableId?: string;
+    variableId: string;
     variableName: string;
     variableDescription: string;
     variableArg?: string;
--- a/packages/ai-chat/src/common/parsed-chat-request.ts
+++ b/packages/ai-chat/src/common/parsed-chat-request.ts
@@ -103,7 +103,7 @@ export class ParsedChatRequestVariablePart implements ParsedChatRequestPart {
         return {
             kind: 'var',
             range: this.range,
-            variableId: this.resolution?.variable.id,
+            variableId: this.resolution?.variable.id ?? 'unresolved variable',
             variableName: this.variableName,
             variableArg: this.variableArg,
             variableValue: this.resolution?.value,


lucas-koehler:
@sdirix That is a very good point! Thanks for catching that. I refactored the code accordingly to restore the tool requests from the registry in the chat service.
In the current refactoring, the chat service uses the registry directly  as the deserializer registry is currently only meant for response content but tool requests are part of the request.
Is this okay with you or do you think we should extend the registry now to also support request content?

sdirix:
I will review at the latest next week.

packages/ai-chat/src/common/chat-model.ts
```diff
@@ -1522,32 +1542,144 @@ export class MutableChatRequestModel implements ChatRequestModel, EditableChatRe
...
+                        partData.range,
+                        partData.variableName,
+                        partData.variableArg
+                    );
+                    if (partData.variableValue !== undefined) {
+                        varPart.resolution = {
+                            variable: {
+                                id: partData.variableId ?? 'restored',
+                                name: partData.variableName,
+                                description: 'Restored from serialized session'
```
sdirix: Should we also store/restore the description?

packages/ai-chat/src/common/chat-model.ts
```diff
@@ -1522,32 +1542,144 @@ export class MutableChatRequestModel implements ChatRequestModel, EditableChatRe
...
+                case 'var': {
+                    const varPart = new ParsedChatRequestVariablePart(
+                        partData.range,
+                        partData.variableName,
+                        partData.variableArg
+                    );
+                    if (partData.variableValue !== undefined) {
+                        varPart.resolution = {
+                            variable: {
+                                id: partData.variableId ?? 'restored',
```
sdirix: I think if there is no id we should not restore it? And during storing we should make sure they have an id?

packages/ai-chat/src/common/chat-model.ts
```diff
@@ -1522,32 +1542,144 @@ export class MutableChatRequestModel implements ChatRequestModel, EditableChatRe
...
+                                description: 'Restored from serialized session'
+                            },
+                            arg: partData.variableArg,
+                            value: partData.variableValue
+                        };
+                    }
+                    return varPart;
+                }
+                case 'function':
+                    // Create placeholder - will be restored by ChatService
```
sdirix, line 1615: We're now using a placeholder approach. I wonder whether we should just construct the whole `ParsedChatRequest` in the service so we could skip the placeholder and create the "real" ones immediately. What do you think? Or do you prefer the placeholders?

packages/ai-chat/src/common/chat-model.ts
lucas-koehler: Yes, this can be quite easily done. I'll add it

packages/ai-chat/src/common/chat-model.ts
lucas-koehler: I'd use the placeholder approach as this would result in a bigger refactoring and we use the pattern of restoring stuff (e.g. changesets) in the service already. Thus, I'd not like to do it as part of this PR but would be open to this being refactored in general as a separate task.

packages/ai-chat/src/common/chat-model.ts
lucas-koehler: I think the problem lies earlier: variable ids are only available if the variable is resolved during serialization (see packages/ai-chat/src/common/parsed-chat-request.ts). Currently we serialize the variable in any case.
We could think about serializing unresolved variables as text parts but there could be value in knowing this is a variable even if it was not/could not be resolved at the time of serialization.

lucas-koehler: For now, I moved adding a fallback value of the id to the serialization function.

lucas-koehler:
@sdirix I added serialization of the description. I also removed the 2 changelog commits during rebase: They cancelled each other and caused conflicts.
I also moved the initialization of the variable id fallback value to the serialization function of a variable part.
For now I left serializing and deserializing variable parts without an id. IMO the core question is whether we want to serialize variable parts with an unresolved variable at all. IMO this can be a follow up issue. WDYT?

Issue #16660, opened 2025-11-26
Quick command panel not closing when pressing escape


dmm9:
### Bug Description:

Starting conditions: The notification panel is open (one or more notifications are currently shown). Then, open the quick command panel with F1.

When pressing escape, the quick command panel does not get closed even when it has the focus. Instead, the notifications panel gets closed.

The root cause seems to be a wrong precedence of global keydown events. By relying only on global key events, we are obviating the DOM hierarchy for events. Maybe installing a keydown event listener directly on the quick-pick-service component would be more appropriate.

### Steps to Reproduce:

1. Have some notification (toast) open
2. Open quick command palette (F1)
3. Press Escape

Expected: Command palette gets closed, notification panel stays open
Actual: Command palette stays open, notification panel gets closed

dmm9:
Actually, I've discovered this because some playright tests have started to fail after adding an extension that shows a message during initialization.

```ts
// examples/playright/src/theia-quick-command-palette.ts

export class TheiaQuickCommandPalette extends TheiaPageObject {
    ...
    async hide(): Promise<void> {
        await this.page.keyboard.press('Escape');
        await this.page.waitForSelector(this.selector, { state: 'hidden' });  /// <--- Playright tests failing here. Palette remains open...
    }
}
```

PR #16763, opened 2025-12-16
Improve keybinding lookup and prioritization

colin-grant-work:
#### What it does

https://github.com/eclipse-theia/theia/issues/16660 revealed a general weakness in the keybinding prioritization system: there was no way for it to prioritize DOM focus in cases where multiple keybindings were active according to globally set context keys.

#16668 provided an ad hoc fix for the quick input system. This PR provides a more general fix for the keybinding system by allowing the keybinding to prioritize keybindings that use context keys defined in contexts applied to the focused DOM element or its parents. It then applies such a local keybinding context to the QuickInput system to address #16660.

It also addresses a long-standing FIXME to improve keybinding lookup times by building a TernarySearchTree for lookup rather than iterating over all defined keybindings. 

#### How to test

1. #16660 should still be fixed - if you reveal notifications and focus a quick input and hit `Esc`, the quick input close command should be triggered rather than the notification hide command.
2. Other keybinding behavior should continue to work as expected.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/ternary-search-tree.spec.ts
+++ b/packages/core/src/common/ternary-search-tree.spec.ts
@@ -1,5 +1,5 @@
 // *****************************************************************************
-// Copyright (C) 2024 TypeFox and others.
+// Copyright (C) 2026 EclipseSource GmbH and others.
 //
 // This program and the accompanying materials are made available under the
 // terms of the Eclipse Public License v. 2.0 which is available at


packages/core/src/common/ternary-search-tree.spec.ts
```diff
@@ -0,0 +1,267 @@
+// *****************************************************************************
+// Copyright (C) 2024 TypeFox and others.
```
ndoschek: nit: is this copyright header correct?

ndoschek (approved):
Thanks! LGTM 👍

PR #16766, opened 2025-12-18
feat: preserve LSP Diagnostic.data in markers

YueCHEN195:
#### What it does

Implemented conversion for the data field in the diagnostic <-> marker lifecycle by updating convertDiagnosticToMarkerData and reviveMarker, enabling ProblemManager markers to carry the data payload. The data field is an opaque, server-defined payload used to attach metadata to a diagnostic and must be preserved by the client. This field is defined on the Diagnostic interface in the vscode-languageserver-protocol package as below:

image

#### How to test

1. Prepare an LSP server or mock that publishes diagnostics including a data property (LSP 3.16+).
2. Open a workspace/file that receives those diagnostics.
3. Verify that ProblemManager exposes markers whose data field is present and matches the published diagnostic data. 
4. Inspect markers via the UI or programmatic API that uses MarkerManager.findMarkers() and check marker.data.data contains the expected value.

image

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
+++ b/packages/plugin-ext/src/common/plugin-api-rpc-model.ts
@@ -210,6 +210,8 @@ export interface MarkerData {
     endColumn: number;
     relatedInformation?: RelatedInformation[];
     tags?: MarkerTag[];
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    data?: any;
 }
 
 export interface RelatedInformation {
--- a/packages/plugin-ext/src/main/browser/languages-main.ts
+++ b/packages/plugin-ext/src/main/browser/languages-main.ts
@@ -1283,7 +1283,8 @@ function reviveMarker(marker: MarkerData): vst.Diagnostic {
         range: reviveRange(marker.startLineNumber, marker.startColumn, marker.endLineNumber, marker.endColumn),
         message: marker.message,
         source: marker.source,
-        relatedInformation: undefined
+        relatedInformation: undefined,
+        data: marker.data
     };
 
     if (marker.relatedInformation) {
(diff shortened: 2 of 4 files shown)


packages/plugin/src/theia.d.ts
```diff
@@ -10343,6 +10343,11 @@ export module '@theia/plugin' {
          */
         tags?: DiagnosticTag[];
 
+        /**
+         * A data entry field that is preserved between diagnostics and code actions.
+         */
+        data?: LSPAny;
```
msujew: Suggestion: The [original implementation](https://github.com/microsoft/vscode/blob/900e7cb348ddf36932721b2bfa45f487c827eed1/src/vscode-dts/vscode.d.ts#L7080-L7143) does not contain this field. Adding it here will result in a mismatch between our interfaces and the one from VSCode, see also our [comparison report](https://eclipse-theia.github.io/vscode-theia-comparator/status.html).

If you want to add it (for typing reasons), please add this to the `theia-extra.d.ts` file.

packages/plugin-ext/src/common/plugin-api-rpc-model.ts
```diff
@@ -210,6 +211,7 @@ export interface MarkerData {
     endColumn: number;
     relatedInformation?: RelatedInformation[];
     tags?: MarkerTag[];
+    data?: LSPAny;
```
msujew: Suggestion: Since this value can stem from non-LSP sources, it doesn't really make sense to use `LSPAny` here. Simply use `any`.

msujew (commented):
Seems to work well 👍

packages/plugin-ext/src/common/plugin-api-rpc-model.ts
YueCHEN195: @msujew Thanks for your Suggestion. I've addressed your comments. Please take another look.

msujew (approved):
Thanks, looks good to me now 👍

PR #16771, opened 2025-12-18
feat: add injectable LoggerSanitizer to mask sensitive data in logs

ndoschek:
#### What it does

Introduce LoggerSanitizer service that automatically masks credentials in log messages to prevent sensitive data leakage (e.g., proxy URLs with username:password, api keys).

- Add LoggerSanitizer interface and DefaultLoggerSanitizer implementation
- Integrate sanitizer into Logger.format()
- Provide a base set of sanitization rules to mask any URL protocol with credentials, api keys and authtokens
- Make sanitizer injectable and optional
- Add unit test cases

#### How to test

- add a proxy env variable either on cli or via the launch config, e.g. `"HTTP_PROXY": "http://myusername:PASSWORD@proxy.example.com:8080"`
- start an example application with loglevel debug (e.g. via npm run start:debug or via the launch configs)
  - for additional test cases open the user settings file, close the applicaiton and inspect the 'fireDidChangeContent' logs on startup (contains api keys/authtokens if stored in user settings)
- optional: add `"--log-file=example.log"` to make log inspection easier
- verify that the url containing the logged credentials/api keys/auth tokens have been masked, e.g. `"HTTP_PROXY": "http://****:****@proxy.example.com:8080"`

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/logger-sanitizer.ts
+++ b/packages/core/src/common/logger-sanitizer.ts
@@ -75,10 +75,9 @@ export const DefaultSanitizationRules: SanitizationRule[] = [
         /**
          * Regex pattern to match URLs with credentials.
          * Matches any URL with format: protocol://user:pass@host[:port]
-         * Protocol is any sequence of alphanumeric characters followed by ://
-         * Username and password cannot contain : / or @ to avoid over-matching
+         * Capture groups: $1=protocol, $2=username, $3=password, $4=host (with optional port)
          */
-        pattern: /([a-z][a-z0-9+.-]*:\/\/)([^:/@]+):([^:/@]+)@([^/:@\s]+(?::\d+)?)/gi,
+        pattern: /([a-z][a-z0-9+.-]*:\/\/)([^:/@]+):([^:/@]+)@([^/:@\s]+(?::\d+)?)/giu,
         replacement: '$1****:****@$4'
     },
     {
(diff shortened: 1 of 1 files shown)


packages/core/src/common/logger.ts
```diff
@@ -319,10 +321,32 @@ export class Logger implements ILogger {
...
+                    return value;
+                } else if (Array.isArray(value)) {
+                    return value.map(item => this.format(item));
+                } else {
+                    try {
+                        const stringified = JSON.stringify(value);
+                        const sanitized = this.sanitize(stringified);
+                        return JSON.parse(sanitized);
+                    } catch {
+                        return value;
```
CamilleLetavernier, line 342: Should we have two distinct try/catch here, to return `sanitize` if we successfully parsed value, but failed to parse the sanitized value?

ndoschek: Good point, thanks! I added an inner try/catch to return the sanitized string if parsing fails, along with a test case to cover this scenario too.

packages/core/src/common/logger-sanitizer.ts
```diff
@@ -75,17 +75,17 @@ export const DefaultSanitizationRules: SanitizationRule[] = [
         /**
          * Regex pattern to match URLs with credentials.
          * Matches any URL with format: protocol://user:pass@host[:port]
-         * Protocol is any sequence of alphanumeric characters followed by ://
-         * Username and password cannot contain : / or @ to avoid over-matching
+         * Capture groups: $1=protocol, $2=username, $3=password, $4=host (with optional port)
          */
-        pattern: /([a-z][a-z0-9+.-]*:\/\/)([^:/@]+):([^:/@]+)@([^/:@\s]+(?::\d+)?)/gi,
+        pattern: /([a-z][a-z0-9+.-]+:\/\/)([^:/@]+):([^:/@]+)@([^/:@\s]+(?::\d+)?)/giu,
```
CamilleLetavernier: I think '*' is better in this case, as we already check [a-z] first (so a:// would be a valid protocol)

`[a-z][a-z0-9+.-]+`: one letter followed by at least one alphanumeric character (so at least 2 characters) :x: 
`[a-z][a-z0-9+.-]*`: one letter, optionally followed by alphanumeric characters (so at least 1 character) :heavy_check_mark:

Issue #16753, opened 2025-12-15
TypeScript language features are broken in browser-based IDE on node >= 21


pisv:
### Steps to Reproduce:

0. Use Node.JS version >= 21.

1. Build and run the browser example:

```sh
git clone https://github.com/eclipse-theia/theia \
    && cd theia \
    && npm install \
    && npm run build:browser \
    && npm run download:plugins \
    && npm run start:browser
```

Start your browser on <http://localhost:3000>.

2. Open the Theia repository root folder in the browser-based IDE. (Actually, any project in TypeScript will do, as long as it contains a valid `tsconfig.json`.)

3. Open a `.ts` source file, e.g. `preference-string-input.ts`.

Observe that:

* The TypeScript language features provided in the editor are limited to the current file, e.g. suggestions are only shown from the current file and `Go to Definition` can only jump to a symbol in the current file.

* The language status item in the status bar indicates that the current file is in [partial mode](https://code.visualstudio.com/docs/nodejs/working-with-javascript#_partial-intellisense-mode), in which project-wide IntelliSense is unavailable.

* The Output view for TypeScript shows that semantic TS Server is not being used, as only syntax TS Server has been started:

```
[Info] Starting TS Server
[Info] Using tsserver from: /.../theia/plugins/vscode.typescript-language-features/extension/deps/typescript/lib/tsserver.js
[Info] <syntax> Forking...
[Info] <syntax> Starting...
```

### Additional Information

Actually, I could not believe my own eyes when I saw it and spent most of the weekend trying to troubleshoot the issue, my first thought being that something should be wrong with my local setup, as I could not find any open issue for it and a problem of this kind could certainly not go unnoticed. Also, I could swear that it worked just fine a while ago. (When I was still using node 20.x, as I understand now.)

Finally, I replaced the downloaded version 1.95.3 of the `vscode.typescript-language-features` extension with a locally built version from the VS Code main branch, and it fixed the issue for me.

Debugging the extension then revealed that the issue is caused by [this code](https://github.com/microsoft/vscode/blob/51eba2dc05a01bcfd8467fbb85bcaf4521b4b03d/extensions/typescript-language-features/src/utils/platform.ts#L8-L10) in the older versions of `typescript-language-features`, which has been fixed in version `1.102.0` by [this commit](https://github.com/microsoft/vscode/commit/0140ab3f1b383871c5ecc910cc0eba166607d485).

[...]

Issue #16400, opened 2025-10-08
Upgrade Typescript to 5.9.x


ndoschek:
### Feature Description:
- Upgrade Typescript to 5.9.x.
- Update target to `ES2023` as we support >= Node 20 (see https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping#node-20)

Issue #16397, opened 2025-10-08
[vscode] support typescript extension version 1.104.0


ndoschek:
### Feature Description:

Add support for new features introduced in the updated built-in TypeScript extension (vscode-typescript-language-features) version 1.104.0.

Features to implement:
- [ ] Implement `editorHoverVerbosityLevel` proposed API to support configurable hover verbosity in the editor

Issue #16279, opened 2025-09-12
[vscode] update VS Code built-in extensions


ndoschek:
### Feature Description:

Update the built-in extensions to target version `1.104.x` to align with the ongoing API compatibility work and ensure consistency for the upcoming 1.66.x community release.

- Requires completion of GH-16134
- Requires publishing of built-in extension (https://github.com/eclipse-theia/vscode-builtin-extensions/issues/139) Edit: we decided for an intermediate solution to publish a vsx archive on GH
- Requires typescript extension support (https://github.com/eclipse-theia/theia/issues/16397)

ndoschek:
Credentials for publishing are on actively being worked on, however we push this to Theia v1.67.x now, to not risk anything right before the CR 1.66.x candidate

ndoschek:
Update other built-ins (e.g. [EditorConfig](https://github.com/eclipse-theia/theia/discussions/16575) and eslitn)

PR #16774, opened 2025-12-18
update VS Code built-in extensions to 1.104.0 and Typescript to 5.9.3

ndoschek:
#### What it does

- `update VS Code built-in extensions and plugin downloader`
  - Resolves GH-16279
  - update built-in extensions to version 1.104.0 and consume tar.gz archive from GitHub releases instead of open-vsx for now
  - add support for decompressing vsix bundles from tar.gz archives
  - ensure download:plugins command exits with proper exit codes
  - update theiaPlugins versions (EditorConfig, ESLint, Jupyter)
  - remove vscode.microsoft-authentication from excluded IDs (as it is not a standalone built-ins anymore)
  - remove ms-vscode.references-view from excluded IDs (it is now vscode.references.view and works without issues)
- `chore: upgrade typescript to ~5.9.3`
  - update base tsconfig target to ES2023 as we support >= Node 20
  - Resolves GH-16400
- `fix: disable ES2022 class field semantics useDefineForClassFields=false`
  - TypeScript 5+ with target ES2022+ changes how class fields are initialized by using Object.defineProperty semantics that can alter initialization order relative to super() calls. Several parts of our codebase rely on the older (ES2019-style) behavior where field assignments occur after the super() call, so this flag preserves compatibility for now. We can revisit this later and enable the standard behavior (useDefineForClassFields=true) once all affected classes are refactored to be spec-compliant.
- `fix: update code for new typescript version `
  - Main areas of update:
    - v regex flag is only available when targeting 'esnext' or later (we still use commonjs)
    - fix reported ts errors, add missing overrides
- `vscode: implement proposed.editorHoverVerbosityLevel API`
  - Resolves GH-16397
  -  add proposed API definitions for editorHoverVerbosityLevel to support verbose hover (e.g. for typescript uses this in most cases now)
  - update plugin HoverAdapter that provides/releases hover
  - Remark: increase/decrease level will most likely work then with updated monaco version (GH-16401)

Resolves GH-16279
Resolves GH-16400
Resolves GH-16397
Fixes GH-16753

<details>
<summary>More details about the switch to an archive instead of the extension pack:</summary>

Open VSX changed its ownership behavior, and now you need to claim a namespace to be able to publish an extension under it.
Our extension pack was already in the eclipse-theia namespace, but all the bundled extensions were under the vscode namespace. Unsurprisingly, the vscode namespace is claimed by Microsoft.
Previously, this wasn’t an issue, and we were still able to publish the extensions under the vscode namespace.
Also, the extension pack is an Open VSX concept.
As an intermediate solution, we discussed in the last dev meeting that we’ll temporarily skip the Open VSX publishing step, package the VSIX ourselves, and consume it with less effort for now, since the last built-in update was over a year ago.
We decided to postpone the final decision on whether we want to continue publishing on Open VSX and how to resolve the namespace problem.
We also considered changing the publisher to eclipse-theia, but this might lead to unforeseen issues.
For example, if other extensions depend on the built-ins, we might need some kind of hacky workaround to intercept and override the publisher prefix somewhere in Theia.
And since they were always intended to be "built-in", there's probably not too much benefit in having them listed in the store.

- https://github.com/eclipse-theia/theia/wiki/Dev-Meetings#2025-12-16
- https://github.com/eclipse-theia/vscode-builtin-extensions/pull/140
- https://github.com/eclipse-theia/vscode-builtin-extensions/releases/tag/1.104.0
</details>

#### How to test

- start with a clean repo (ideally run: `npm run clean && git clean -xdf`)
- `npm run all`
- `npm run download:plugins`
- **Test built-in extensions with browser and electron example applications:**
  - Typescript language support
  - other language support (like json, css etc)
  - git support (scm view, but not support all features fully yet, see also follow ups)
  - Start a sample task to check the line matcher still works as expected
    <details>
    
    <summary>e.g. add a .vscode/tasks.json file </summary>
    {
      "version": "2.0.0",
      "tasks": [
        {
          "label": "Lint",
          "type": "shell",
          "command": "npm run lint",
          "problemMatcher": ["$tsc"]
        }
      ]
    }
    </details>
    
  - Observe logs for new exceptions/errors/warnings

#### Follow-ups

- GH-16401
- GH-16773
- GH-16280
- Investigate currently excluded built-ins `vscode.github` and `vscode.github-authentication` (see also GH-16713)
- Revisit `useDefineForClassFields=false` topic

#### Breaking changes

#### Attribution

#### Review checklist

#### Reminder for reviewers


packages/monaco/src/browser/monaco-editor-content-menu.ts
```diff
@@ -70,42 +70,44 @@ export class MonacoEditorContentMenuContribution implements FrontendApplicationC
...
-            // eslint-disable-next-line @typescript-eslint/no-shadow
-            toDispose.push(ObservableUtils.autorunWithDisposables(({ toDispose }) => {
-                const firstMatch = firstMatchObservable.get();
-                if (firstMatch) {
-                    const button = new MonacoEditorOverlayButton(editor, firstMatch.label);
-                    toDispose.push(button);
-                    toDispose.push(button.onClick(() =>
-                        this.withContext(context, () => firstMatch.run(EDITOR_CONTENT_MENU, editorWidget))
-                    ));
+            if (menuNodes) {
```
pisv: [nitpick] Although this additional check will certainly not hurt anything, it does not look like it's really necessary, as `menuNodes` is `CommandMenu[]` and cannot be falsy here.

ndoschek:
Thanks @pisv for testing and reviewing, much appreciated! :tada: 
I'll take another look after the holiday break!

packages/ai-chat-ui/src/browser/chat-tree-view/chat-view-tree-input-widget.tsx
```diff
@@ -51,7 +51,7 @@ export class AIChatTreeInputWidget extends AIChatInputWidget {
     protected readonly args: AIChatTreeInputArgs;
 
     @inject(AIChatTreeInputConfiguration) @optional()
-    protected override readonly configuration: AIChatTreeInputConfiguration | undefined;
+    protected declare readonly configuration: AIChatTreeInputConfiguration | undefined;
```
colin-grant-work: You mention the switch from `override` to `declare` as an adjustment for the new TS version. What is the significance of the change, and are there particular conditions under which it's required? I tried reverting a few new `declare`s to `override`, and the application still built, so at least in some cases, it doesn't seem that the new TS objects to the older `override` style. Is it a future-facing change to prepare for switch to the `define` semantics?

The `declare` keyword seems a little out of place, because for injections, at least, we are in fact setting the value, not just declaring its presence for type purposes. My interpretation of the keyword may be a little narrow, but the new usage feels at least unaccustomed.

packages/monaco/src/browser/monaco-editor.ts
```diff
@@ -406,6 +404,14 @@ export class MonacoEditor extends MonacoEditorServices implements TextEditor {
         return this.onMouseDownEmitter.event;
     }
 
+    get onDidChangeReadOnly(): Event<boolean | MarkdownString> {
+        return this.document.onDidChangeReadOnly;
+    }
+
+    get onEncodingChanged(): Event<string> {
+        return this.document.onDidChangeEncoding;
+    }
```
colin-grant-work: There are a number of changes like this, of constructor-initialized field access -> getter. I imagine that falls under the heading of changes 'not to access fields before initialization.' I also tried reverting some of these and ran into no build or runtime problems. Is this also something that would be a problem if we switched to define-property semantics?

The reason I mention this is that if there's no build failure (and no lint rule) that detects these, new code very well may just bring in new instances, so if it's a problem, or something we want to make sure avoids becoming a problem, we may need some alternate means.

packages/plugin-ext/src/plugin/languages/hover.ts
```diff
@@ -39,20 +50,48 @@ export class HoverAdapter {
...
+
+        const convertedHover: Hover = Converter.fromHover(value);
+        const id = this._hoverCounter;
+        // Check if hover map has more than 10 elements and if yes, remove oldest from the map
+        if (this._hoverMap.size === HoverAdapter.HOVER_MAP_MAX_SIZE) {
+            const minimumId = Math.min(...this._hoverMap.keys());
+            this._hoverMap.delete(minimumId);
+        }
+        this._hoverMap.set(id, value);
+        this._hoverCounter += 1;
```
colin-grant-work: Probably better if the use of the ID and the incrementation of the counter are closer together so that no future code can insert control flow between them and create tricky conditional bugs.

packages/plugin-ext/src/plugin/type-converters.ts
```diff
@@ -424,10 +424,12 @@ function convertTags(tags: types.DiagnosticTag[] | undefined): types.MarkerTag[]
...
 }
 
-export function fromHover(hover: theia.Hover): model.Hover {
+export function fromHover(hover: theia.VerboseHover): model.Hover {
     return <model.Hover>{
         range: fromRange(hover.range),
-        contents: fromManyMarkdown(hover.contents)
+        contents: fromManyMarkdown(hover.contents),
+        canIncreaseVerbosity: hover.canIncreaseVerbosity,
+        canDecreaseVerbosity: hover.canDecreaseVerbosity,
```
colin-grant-work, line 433: Minor, old code: there doesn't appear to be any need for the cast here. Probably true in a number of places in this file.

packages/console/src/browser/console-session.ts
```diff
@@ -39,7 +39,7 @@ export abstract class ConsoleSession extends TreeSource {
     protected selectedSeverity?: Severity;
     protected readonly selectionEmitter: Emitter<void> = new Emitter<void>();
     readonly onSelectionChange = this.selectionEmitter.event;
-    override id: string;
+    declare id: string;
```
colin-grant-work: This one is just bad: without strict property initialization, I don't think TS is actually doing anything to help us enforce that descendants of this class actually have an ID, regardless of whether we use `override` or `declare`.

packages/task/src/node/task-abstract-line-matcher.ts
```diff
@@ -124,7 +124,7 @@ export abstract class AbstractLineMatcher {
                 if (trim) {
                     value = value.trim();
                 }
-                (data[property] as string) = value;
+                (data[property] as unknown as string) = value;
```
colin-grant-work: I think the real fix here is not the cast, but to create a type like this:

```
type ProblemDataStringKeys = keyof { [K in keyof ProblemData as string extends ProblemData[K] ? K : never]: unknown };
```

And then use that instead of `keyof ProblemData` in this method and `appendProperty`. Then we exclude `ProblemLocationKind` and the type of `data[property]` is correctly computed to be `string | undefined`, so we can safely assign a string to it.

packages/monaco/src/browser/monaco-editor-content-menu.ts
ndoschek: Thanks for catching this, this probably was just a leftover, I reverted this again!

packages/ai-chat-ui/src/browser/chat-tree-view/chat-view-tree-input-widget.tsx
ndoschek: Thanks for the insights, that makes sense 👍  

I double-checked as well and reverted the `declare` usages again. It looks like I made that change before enabling the relevant TS flag and didn't question it again afterwards. Given that the code still builds fine with `override` and there’s no clear benefit from switching in this case, I agree it’s better to keep the original approach and avoid unnecessary code changes.

packages/monaco/src/browser/monaco-editor.ts
ndoschek: Good point, thanks for calling this out 👍  
I also tried reverting some of these changes and couldn’t reproduce any build or runtime issues. Similar to the previous case, to keep the changes minimal, I've reverted them as well.

packages/plugin-ext/src/plugin/type-converters.ts
ndoschek: Agreed, I changed that cast here, I've created a task for myself to have a look at the entire type converter file at a later point.

ndoschek:
Thank you both for the detailed feedback, much appreciated!

> I also tested git support to some extent and found a regression in the code editor's dirty diff peek view, where the actions for staging/reverting the current change were previously visible in the peek view's action bar like this:

I assume this is linked to the monaco update, I'll take a note for the monaco upgrade followup to double check this then as well!

I've updated my commits accordingly. When you have a moment, could you please take another look? TIA!

pisv:
+1 :-)

Thanks!

PR #16776, opened 2025-12-19
feat: implement Preference for AI Chat Persisted Session Limit

### LINKED ISSUE #16560: AI Persistence: Preference for Persisted Session Limit
### Description

Add a user preference to control how many chat sessions are persisted before old sessions are automatically deleted.

Currently, the system has a hard-coded limit of 25 sessions defined in [`ChatSessionStoreImpl`](https://github.com/eclipse-theia/theia/blob/7946ca874b04e6249d883e9a586f193194df1365/packages/ai-chat/src/browser/chat-session-store-impl.ts#L29).

### Proposed Behavior

- Add preference `ai-features.chat.persistedSessionLimit` (default: 25)
- Use `-1` to indicate unlimited sessions
- Use `0` to not persist any sessions
- When limit is reduced, oldest sessions beyond the new limit are automatically removed on next save
- When limit is increased, sessions are retained up to the new limit

### Technical Considerations

- Add preference definition to [`ai-chat-preferences.ts`](https://github.com/eclipse-theia/theia/blob/7946ca874b04e6249d883e9a586f193194df1365/packages/ai-chat/src/common/ai-chat-preferences.ts)
- Modify [`ChatSessionStoreImpl`](https://github.com/eclipse-theia/theia/blob/7946ca874b04e6249d883e9a586f193194df1365/packages/ai-chat/src/browser/chat-session-store-impl.ts#L29) to inject and read from `PreferenceService`
- Update cleanup logic in [`trimSessions()`](https://github.com/eclipse-theia/theia/blob/7946ca874b04e6249d883e9a586f193194df1365/packages/ai-chat/src/browser/chat-session-store-impl.ts#L222-L237) to respect the preference value
- **Important:** Limit enforcement only happens during save operations, not when preference changes, to avoid accidental data loss

### Related

- Follow up from initial implementation #16486

CamilleLetavernier:
fixes #16560

#### What it does

- add a preference `AI Features > Chat > Persisted Session Limit` with values:
  - -1: unlimited persisted sessions
  - 0: no persisted sessions
  - n>0: n persisted sessions
- update the persisted sessions mechanism to take the preference into account and support the new 0 and -1 cases

#### How to test

- test all 3 values (-1, 0, and n>0) and make sure the behavior matches the expectations above
  - note: persisted sessions are the sessions saved to disk. They are different from in-memory sessions. Refresh your browser tab when you want to check what is currently persisted, as in-memory sessions have no limit
  - note: sessions are trimmed only when a new session is available. Changing the setting doesn't have an immediate effect; you need to start a new conversation in order to update the current persisted sessions

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat/src/browser/chat-session-store-impl.ts
+++ b/packages/ai-chat/src/browser/chat-session-store-impl.ts
@@ -18,15 +18,16 @@ import { inject, injectable, named } from '@theia/core/shared/inversify';
 import { FileService } from '@theia/filesystem/lib/browser/file-service';
 import { WorkspaceService } from '@theia/workspace/lib/browser';
 import { EnvVariablesServer } from '@theia/core/lib/common/env-variables';
+import { PreferenceService } from '@theia/core/lib/common';
 import { StorageService } from '@theia/core/lib/browser';
 import { URI } from '@theia/core';
 import { BinaryBuffer } from '@theia/core/lib/common/buffer';
 import { ILogger } from '@theia/core/lib/common/logger';
 import { ChatModel } from '../common/chat-model';
 import { ChatSessionIndex, ChatSessionStore, ChatModelWithMetadata, ChatSessionMetadata } from '../common/chat-session-store';
+import { PERSISTED_SESSION_LIMIT_PREF } from '../common/ai-chat-preferences';
 import { SerializedChatData, CHAT_DATA_VERSION } from '../common/chat-model-serialization';
 
-const MAX_SESSIONS = 25;
 const INDEX_FILE = 'index.json';
 
 @injectable()
@@ -45,6 +46,9 @@ export class ChatSessionStoreImpl implements ChatSessionStore {
 
     @inject(ILogger) @named('ChatSessionStore')
     protected readonly logger: ILogger;
+
+    @inject(PreferenceService)
+    protected readonly preferenceService: PreferenceService;
 
     protected storageRoot?: URI;
     protected indexCache?: ChatSessionIndex;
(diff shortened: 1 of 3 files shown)


packages/ai-chat/src/browser/chat-session-store-impl.ts
```diff
@@ -215,25 +219,63 @@ export class ChatSessionStoreImpl implements ChatSessionStore {
...
         for (const session of sessionsToDelete) {
-            await this.deleteSession(session.sessionId);
+            const root = await this.getStorageRoot();
+            const sessionFile = root.resolve(`${session.sessionId}.json`);
+            try {
+                await this.fileService.delete(sessionFile);
+            } catch (e) {
+                this.logger.debug('Failed to delete session file', { sessionId: session.sessionId, error: e });
+            }
+            delete index[session.sessionId];
```
ndoschek, line 275: Is there a reason why we don't use the deleteSession function anymore here? (and also for the maxSessions === 0 case above)?

CamilleLetavernier: Using the deleteSession in a loop didn't work - I noticed a bug before working on the preferences, where my store sessions count was > 25. Possibly because of the way to store promise is chained onto itself? But I didn't really investigate the issue.

There might be a way to merge the two behaviors (fix deleteSession so that it works reliably in a loop, and keep using it from trimSessions), but out of the box, it didn't give the expected results.

deleteSession also saves the index again on every iteration, but in trimSessions, we only save the index once (at the end).

ndoschek: Ah alright, thanks for the explanation, then let's keep it that way :+1:

ndoschek (approved):
Thanks! LGTM 👍

PR #16798, opened 2026-01-06
Add todo tool

JonasHelming:
#### What it does

Adds a TodoList feature to the AI chat system, allowing agents to track task progress during conversations.
The todo tool is currently not integrated in any agent, yet.

  - Introduces TodoList interface and TodoListImpl in @theia/ai-chat
  - Follows the per-request pattern (like ChangeSet) with ChatTreeTodoList aggregator for branch-aware state 
  - Adds todoWrite and todoRead tools in @theia/ai-ide
  - Persists todo state with chat sessions

#### How to test

  1. Add ~{todoWrite} and ~{todoRead} to an agent prompt and start a chat session 
  2. Ask the agent to perform a multi-step task and use the todo list
  3. Verify the agent can create and update todos via todoWrite
  4. Verify the agent can read current state via todoRead
  5. Test branching: edit an earlier message, verify the new branch has independent todo state
  6. Test persistence: reload the session, verify todos are restored

#### Follow-ups

Integrate in agents, specifically the Theia Coder agent mode

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-ide/src/browser/todo-functions.ts
+++ b/packages/ai-ide/src/browser/todo-functions.ts
@@ -97,15 +97,7 @@ export class TodoWriteToolProvider implements ToolProvider {
                     });
                 }
 
-                return JSON.stringify({
-                    success: true,
-                    items: addedItems,
-                    summary: {
-                        total: addedItems.length,
-                        pending: addedItems.filter(i => i.state === 'pending').length,
-                        completed: addedItems.filter(i => i.state === 'completed').length
-                    }
-                });
+                return 'Todo list updated';
             }
         };
     }


sdirix:
I will review at the latest next week.

packages/ai-ide/src/browser/todo-functions.ts
```diff
@@ -0,0 +1,151 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH.
```
sdirix: Nitpick
```suggestion
// Copyright (C) 2026 EclipseSource GmbH.
```

packages/ai-chat/src/common/todo-list.ts
```diff
@@ -0,0 +1,142 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH and others.
```
sdirix: Nitpick
```suggestion
// Copyright (C) 2026 EclipseSource GmbH and others.
```

packages/ai-chat/src/common/todo-list.ts
```diff
@@ -0,0 +1,142 @@
...
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { Disposable, Emitter, Event } from '@theia/core';
+import { generateUuid } from '@theia/core/lib/common/uuid';
+
+export type TodoItemState = 'pending' | 'completed';
```
sdirix, line 20: An inProgress state could be useful. Anthropic uses that too and I think it just makes sense to return to the user what the LLM/agent is currently working on. Anthropic makes sure that only one todo can be in progress, maybe we want to enforce that too.

packages/ai-chat/src/common/todo-list.ts
```diff
@@ -0,0 +1,142 @@
...
+export interface TodoList extends Disposable {
+    onDidChange: Event<TodoListChangeEvent>;
+    getItems(): TodoItem[];
+    getItem(id: string): TodoItem | undefined;
+    addItem(content: string, notes?: string): TodoItem;
+    updateItem(id: string, updates: { content?: string; state?: TodoItemState; notes?: string }): boolean;
+    removeItem(id: string): boolean;
+    clear(): void;
+    toSerializable(): SerializedTodoItem[];
+    restoreFromSerialized(items: SerializedTodoItem[]): void;
```
sdirix, line 50: I would have expected the serialized list to be an optional constructor argument similar to the chat-model design.

packages/ai-chat/src/common/todo-list.ts
```diff
@@ -0,0 +1,142 @@
...
+        this._items.clear();
+        for (const item of items) {
+            this._items.set(item.id, {
+                id: item.id,
+                content: item.content,
+                state: item.state,
+                notes: item.notes
+            });
+        }
+        this.notifyChange();
```
sdirix, line 131: Restoring is done before we hand the ChatModel to the outside world, so no need to trigger notifications

packages/ai-ide/src/browser/todo-functions.ts
```diff
@@ -0,0 +1,151 @@
...
+            },
+            handler: async (args: string, ctx: MutableChatRequestModel): Promise<string> => {
+                if (ctx?.response?.cancellationToken?.isCancellationRequested) {
+                    return JSON.stringify({ error: 'Operation cancelled by user' });
+                }
+
+                const { todos } = JSON.parse(args) as TodoWriteArgs;
+                const todoList = ctx.session.todoList;
+
+                todoList.clear();
```
sdirix, line 84: Clearing and then readding all todos one by one will cause many many notifications to be sent. Mostly for todos which have not changed at all.

packages/ai-ide/src/browser/todo-functions.ts
```diff
@@ -0,0 +1,151 @@
...
+
+                const addedItems: Array<{ id: string; content: string; state: TodoItemState }> = [];
+
+                for (const todo of todos) {
+                    const item = todoList.addItem(todo.content, todo.notes);
```
sdirix, line 89: By clearing and readding todos on every change we will generate new ids for every todo item on every operation. 

So if the item "Explore codebase" is already handled it will still receive a new uuid-id for every todo list operation. These ids are exposed to the LLM and pollute the context.

packages/ai-chat/src/common/chat-model.ts
```diff
@@ -1095,6 +1112,107 @@ export class ChatTreeChangeSet implements Omit<ChangeSet, 'onDidChange'> {
...
+        this.getMutableTodoList().clear();
+    }
+
+    toSerializable(): SerializedTodoItem[] {
+        return this.getCurrentTodoList()?.toSerializable() ?? [];
+    }
+
+    restoreFromSerialized(items: SerializedTodoItem[]): void {
+        this.getMutableTodoList().restoreFromSerialized(items);
+    }
```
sdirix, line 1157: This restoring is actually never called and therefore todo list are not restored.

packages/ai-chat/src/common/todo-list.ts
```diff
@@ -0,0 +1,142 @@
...
+// *****************************************************************************
+
+import { Disposable, Emitter, Event } from '@theia/core';
+import { generateUuid } from '@theia/core/lib/common/uuid';
+
+export type TodoItemState = 'pending' | 'completed';
+
+export interface TodoItem {
+    readonly id: string;
+    readonly content: string;
```
sdirix, line 24: Anthropic has an additional field `activeForm` which is used as the label for the ToDo Item when shown visually to the user. So the `content` has the imperative form, e.g. "Run tests", while the `activeForm` as "Running tests".
We could foresee a distinction between these two different use cases too.

packages/ai-chat/src/common/todo-list.ts
```diff
@@ -0,0 +1,142 @@
...
+    readonly state: TodoItemState;
+    readonly notes?: string;
```
sdirix, line 26: Instead of notes we could define a data object like for request, chat-model etc. Adopters could then place arbitrary information in the ToDoItem instead of just a single string. But we could also do that later once the need for this arises.

packages/ai-chat/src/common/todo-list.ts
```diff
@@ -0,0 +1,142 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { Disposable, Emitter, Event } from '@theia/core';
+import { generateUuid } from '@theia/core/lib/common/uuid';
+
+export type TodoItemState = 'pending' | 'completed';
+
+export interface TodoItem {
+    readonly id: string;
```
sdirix, line 23: I think I would get rid of the id. I don't see any real usage for this besides filling the context. When adding todos we can make sure that content is unique.

packages/ai-ide/src/browser/todo-functions.ts
```diff
@@ -0,0 +1,151 @@
...
+                }
+
+                return JSON.stringify({
+                    success: true,
+                    items: addedItems,
+                    summary: {
+                        total: addedItems.length,
+                        pending: addedItems.filter(i => i.state === 'pending').length,
+                        completed: addedItems.filter(i => i.state === 'completed').length
+                    }
```
sdirix: What we return here is "all the information the llm handed us over" + "ids" + "summary". For the LLM context this just duplicated the ToDo list (it's in the parameters and in the result) and added some minimal summary information.
I think we should just return an "ok" here. Or if we really want to drive the LLM, then a reminder to continue with the current item in progress or the next not started item.

packages/ai-ide/src/browser/todo-functions.ts
```diff
@@ -0,0 +1,151 @@
...
+                });
+            }
+        };
+    }
+}
+
+@injectable()
+export class TodoReadToolProvider implements ToolProvider {
```
sdirix, line 107: I don't really understand what this tool is used for. An LLM knows the list, as it did write it before. When we restore from a previous chat or continue with a new message, the LLM still has the list in context as it has the last todo list write in context.

Even if we switch agents or models, we still send the old tool calls with us, so they still have access to the current list.

The only use case I could think of is to access the actual Todo list after a destructive compact operation where we don't send the old messages but just a summary of them. But even then we could just make sure the summary contains the current Todo list.

---

When I tried this PR, the LLMs basically just read once at the beginning of the chat. Don't get any items returned and then just never call this again. Sometimes it never called it.

packages/ai-ide/src/browser/todo-functions.ts
```diff
@@ -0,0 +1,151 @@
...
+
+@injectable()
+export class TodoWriteToolProvider implements ToolProvider {
+    static ID = TODO_WRITE_FUNCTION_ID;
+
+    getTool(): ToolRequest {
+        return {
+            id: TodoWriteToolProvider.ID,
+            name: TodoWriteToolProvider.ID,
+            description: 'Manages a structured task list for the current chat session. ' +
```
sdirix, line 40: I think we need many more detailed instructions when and how to use the todo list. For me it often happened that the Todo list was created once with 5-6 items. The LLM then did some work and the next call was with the completion of ALL items.

This was likely a case of todos being too trivial to be even used for that use case to begin with.

---

I tried adding an "in-progress" state and told it to only work on items which are "in-progress" (without limiting it to one like Anthropic does). That worked well for me and it completed todos always in order without skipping entries (i.e. completing multiple at once)

packages/ai-ide/src/common/todo-functions.ts
```diff
@@ -0,0 +1,18 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH.
```
sdirix: Nitpick
```suggestion
// Copyright (C) 2026 EclipseSource GmbH.
```

sdirix (changes requested):
The todo list works in itself as the concept is sound, giving the LLM a tool to structure its output/planning.

However I don't think we need the concept of "reading todos" nor do we really need to track todos over the various chat model branches, in fact we should think about whether todos should be a chat-model concept at all.

A well designed "writeToDo" tool is already giving the LLM all the structure it needs, and the curent state of the ToDo list is tracked via the tool call history in the chat branches already.

As of the current state of the PR I don't see a reason to store ToDos additionally, next to the already existing tool calls, into the chat model. The only use cases I can think of:
 - Injecting ToDos from outside the LLM programmatically
 - Cross-Session work management
 - Automatic todo list restoring after session compacting (which is not a feature in the state of the PR and can be achieved in different means too)
 - Special UI integration in the Chat View

But the use cases are not part of this PR, so as of now, I don't see a reason to add it. For the UI integration we can add special rendering for the ToDoWrite tool calls themselves without needing to rely on special structures in the chat-model.

packages/ai-chat/src/common/todo-list.ts
JonasHelming: I was thinking about this and had it in the first version. Your argument that it enforces step by step completions convinces me, will add it again.

packages/ai-ide/src/browser/todo-functions.ts
JonasHelming: done

Issue #16705, opened 2025-12-03
Interruption breaks chat


eneufeld:
### Bug Description:

Sometimes if you stop a chat while it is executing a tool, the chat gets corrupted and is then broken. We need to be able to cleanup this and not break the chat.

### Steps to Reproduce:

I don't have clean reproducible steps at the moment.

### Additional Information

- Operating System: Ubuntu
- Theia Version: 1.66

rubrx:
Hey, I would like to work on this issue, can you assign this to me?

ndoschek:
Hi again @R-Kri, it looks like eneufeld already had a fix prepared and opened a PR in case you would like to have a look!

rubrx:
Thanks @ndoschek  for the update! Looks good, glad there’s already a fix in place.

PR #16806, opened 2026-01-08
fix(ai-core): do not break chat when interrupting tool calls

eneufeld:
#### What it does

When parsing arguments of a toolcall, handle invalid JSON. This way the code does not crash.

fix: #16705

#### How to test

Ask an agent to change a file. press cancel while the tool is still running.
Then send a new request to the llm .
On main this should lead to an error.
With this it should be fixed.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -278,6 +278,7 @@
     "dev-packages/ffmpeg": {
       "name": "@theia/ffmpeg",
       "version": "1.67.0",
+      "hasInstallScript": true,
       "license": "EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0",
       "dependencies": {
         "@electron/get": "^2.0.0",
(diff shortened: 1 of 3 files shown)


packages/ai-core/src/browser/frontend-language-model-registry.ts
```diff
@@ -324,7 +324,7 @@ export class FrontendLanguageModelRegistryImpl
             } catch (error) {
                 const errorMessage = error instanceof Error ? error.message : String(error);
                 return { error: true, message: `Error executing tool '${toolId}': ${errorMessage}` };
-            };
```
ndoschek, line 327: nit: While this is a valid cleanup, it is unrelated to the scope of this PR as it is in another package. But I leave it up to you to decived if you want to keep it in here.

PR #16809, opened 2026-01-09
feat(workspace): implement workspace trust dialog and management

### LINKED ISSUE #16653: Dialog prompt for workspace trust on folder open
### Feature Description:

When opening a new folder, Theia should prompt the user to decide whether the folder is trusted. This mirrors the behavior in VS Code, which warns users about the implications of trusting a folder, primarily that code from the folder might be executed locally. The decision should be persisted (e.g., in the user settings or the workspace storage) so that the trust status can be remembered across sessions.

See https://code.visualstudio.com/docs/editing/workspaces/workspace-trust

Expected behavior:
- A dialog should appear asking whether to trust the folder.
- If trusted, the decision should be saved for future sessions for that workspace.
- If not trusted, restricted mode should apply (behavior to be defined in broader Workspace Trust implementation).

Tasks:
- Implement dialog and persist user decision
  - see also initial light-weight workspace trust for plugins (https://github.com/eclipse-theia/theia/pull/10473)

eneufeld:
#### What it does

Adds workspace trust functionality:
- Trust dialog on folder open (respecting startupPrompt preference)
- Trusted folders list preference for global trust decisions
- Per-workspace trust storage for 'once' mode
- Restricted Mode banner when workspace is untrusted
- "Manage Workspace Trust" command to toggle trust state
- onDidChangeWorkspaceTrust event for plugins

Plugin API:
- Add onDidChangeWorkspaceTrust: Event<boolean> to workspace namespace
- Update workspace.ts and workspace-main.ts for RPC event flow

Preferences added:
- security.workspace.trust.trustedFolders (string array)

Closes #16653
#### How to test

Open a new workspace after enabling workspace trust checks.
See a dialog pop up. Test that trusted workspaces don't show dialog anymore. Untrusted workspaces show a message.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/plugin-ext/src/plugin/workspace.spec.ts
+++ b/packages/plugin-ext/src/plugin/workspace.spec.ts
@@ -16,6 +16,7 @@
 
 import * as chai from 'chai';
 import { Emitter, Event } from '@theia/core/lib/common/event';
+import { DisposableCollection } from '@theia/core/lib/common/disposable';
 
 const expect = chai.expect;
 
@@ -32,6 +33,11 @@ describe('WorkspaceExtImpl trust change logic', () => {
         readonly onDidGrantWorkspaceTrust: Event<void> = this.didGrantWorkspaceTrustEmitter.event;
         private didChangeWorkspaceTrustEmitter = new Emitter<boolean>();
         readonly onDidChangeWorkspaceTrust: Event<boolean> = this.didChangeWorkspaceTrustEmitter.event;
+
+        dispose(): void {
+            this.didGrantWorkspaceTrustEmitter.dispose();
+            this.didChangeWorkspaceTrustEmitter.dispose();
+        }
 
         get trusted(): boolean {
             return !!this._trusted;
@@ -56,9 +62,11 @@ describe('WorkspaceExtImpl trust change logic', () => {
     let handler: TrustChangeHandler;
     let grantEvents: number;
     let changeEvents: boolean[];
+    const disposables = new DisposableCollection();
 
     beforeEach(() => {
         handler = new TrustChangeHandler();
+        disposables.push(handler);
         grantEvents = 0;
         changeEvents = [];
 
@@ -68,6 +76,10 @@ describe('WorkspaceExtImpl trust change logic', () => {
         handler.onDidChangeWorkspaceTrust(trust => {
             changeEvents.push(trust);
         });
+    });
+
+    afterEach(() => {
+        disposables.dispose();
     });
 
     it('fires onDidChangeWorkspaceTrust when transitioning from undefined to true', () => {
--- a/packages/workspace/src/browser/workspace-frontend-contribution.ts
+++ b/packages/workspace/src/browser/workspace-frontend-contribution.ts
@@ -507,15 +507,16 @@ export class WorkspaceFrontendContribution implements CommandContribution, Keybi
         const currentTrust = await this.workspaceTrustService.getWorkspaceTrust();
         const trust = nls.localizeByDefault('Trust');
         const dontTrust = nls.localizeByDefault("Don't Trust");
+        const currentSuffix = `(${nls.localizeByDefault('Current')})`;
 
         const items: QuickPickItem[] = [
             {
                 label: trust,
-                description: currentTrust ? nls.localize('theia/workspace/currentTrustState', '(current)') : undefined
+                description: currentTrust ? currentSuffix : undefined
             },
             {
                 label: dontTrust,
-                description: !currentTrust ? nls.localize('theia/workspace/currentTrustState', '(current)') : undefined
+                description: !currentTrust ? currentSuffix : undefined
             }
         ];
 
--- a/packages/workspace/src/browser/workspace-trust-service.ts
+++ b/packages/workspace/src/browser/workspace-trust-service.ts
@@ -17,19 +17,21 @@
 import { ConfirmDialog, Dialog, StorageService } from '@theia/core/lib/browser';
 import { StatusBar, StatusBarAlignment } from '@theia/core/lib/browser/status-bar/status-bar';
 import { OS } from '@theia/core';
+import { DisposableCollection } from '@theia/core/lib/common/disposable';
 import { Emitter, Event } from '@theia/core/lib/common';
 import URI from '@theia/core/lib/common/uri';
-import { PreferenceChange, PreferenceScope, PreferenceService } from '@theia/core/lib/common/preferences';
+import { PreferenceChange, PreferenceSchemaService, PreferenceScope, PreferenceService } from '@theia/core/lib/common/preferences';
 import { MessageService } from '@theia/core/lib/common/message-service';
 import { nls } from '@theia/core/lib/common/nls';
 import { Deferred } from '@theia/core/lib/common/promise-util';
-import { inject, injectable, postConstruct } from '@theia/core/shared/inversify';
+import { inject, injectable, postConstruct, preDestroy } from '@theia/core/shared/inversify';
 import { WindowService } from '@theia/core/lib/browser/window/window-service';
 import {
     WorkspaceTrustPreferences, WORKSPACE_TRUST_EMPTY_WINDOW, WORKSPACE_TRUST_ENABLED, WORKSPACE_TRUST_STARTUP_PROMPT, WORKSPACE_TRUST_TRUSTED_FOLDERS, WorkspaceTrustPrompt
 } from '../common/workspace-trust-preferences';
 import { FrontendApplicationConfigProvider } from '@theia/core/lib/browser/frontend-application-config-provider';
 import { WorkspaceService } from './workspace-service';
+import { WorkspaceCommands } from './workspace-commands';
 import { ContextKeyService } from '@theia/core/lib/browser/context-key-service';
 
 const STORAGE_TRUSTED = 'trusted';
@@ -52,6 +54,9 @@ export class WorkspaceTrustService {
     @inject(WorkspaceTrustPreferences)
     protected readonly workspaceTrustPref: WorkspaceTrustPreferences;
 
+    @inject(PreferenceSchemaService)
+    protected readonly preferenceSchemaService: PreferenceSchemaService;
+
     @inject(WindowService)
     protected readonly windowService: WindowService;
 
@@ -63,9 +68,12 @@ export class WorkspaceTrustService {
 
     protected workspaceTrust = new Deferred<boolean>();
     protected currentTrust: boolean | undefined;
+    protected pendingTrustDialog: Deferred<boolean> | undefined;
 
     protected readonly onDidChangeWorkspaceTrustEmitter = new Emitter<boolean>();
     readonly onDidChangeWorkspaceTrust: Event<boolean> = this.onDidChangeWorkspaceTrustEmitter.event;
+
+    protected readonly toDispose = new DisposableCollection(this.onDidChangeWorkspaceTrustEmitter);
 
     @postConstruct()
     protected init(): void {
@@ -75,18 +83,30 @@ export class WorkspaceTrustService {
     protected async doInit(): Promise<void> {
         await this.workspaceService.ready;
         await this.workspaceTrustPref.ready;
+        await this.preferenceSchemaService.ready;
         await this.resolveWorkspaceTrust();
-        this.preferences.onPreferenceChanged(change => this.handlePreferenceChange(change));
-        this.workspaceService.onWorkspaceChanged(() => this.handleWorkspaceChanged());
+        this.toDispose.push(
+            this.preferences.onPreferenceChanged(change => this.handlePreferenceChange(change))
+        );
+        this.toDispose.push(
+            this.workspaceService.onWorkspaceChanged(() => this.handleWorkspaceChanged())
+        );
 
         // Show status bar item if starting in restricted mode
         const initialTrust = await this.getWorkspaceTrust();
         this.updateRestrictedModeIndicator(initialTrust);
 
         // React to trust changes
-        this.onDidChangeWorkspaceTrust(trust => {
-            this.updateRestrictedModeIndicator(trust);
-        });
+        this.toDispose.push(
+            this.onDidChangeWorkspaceTrust(trust => {
+                this.updateRestrictedModeIndicator(trust);
+            })
+        );
+    }
+
+    @preDestroy()
+    protected onStop(): void {
+        this.toDispose.dispose();
     }
 
     getWorkspaceTrust(): Promise<boolean> {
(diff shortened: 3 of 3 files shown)


.prompts/project-info.prompttemplate
```diff
@@ -59,6 +59,28 @@ Tests are located in the same directory as the components under test.
...
+2. Create a `PreferenceContribution` symbol and bind it
+3. Use `PreferenceProxy` for type-safe access to preference values
+
+**Example:** `packages/workspace/src/common/workspace-trust-preferences.ts`
+
+### Plugin API Extensions
+
+To extend the plugin API (VS Code compatible API):
+
+1. **Type definitions:** Add to `packages/plugin/src/theia.d.ts`
```
ndoschek, line 76: We should also mention the proposed APIs here.

packages/plugin-ext/src/plugin/workspace.spec.ts
```diff
@@ -0,0 +1,142 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource and others.
```
ndoschek:
```suggestion
// Copyright (C) 2026 EclipseSource and others.
```

packages/workspace/src/browser/workspace-trust-service.spec.ts
```diff
@@ -0,0 +1,72 @@
+// *****************************************************************************
+// Copyright (C) 2025 EclipseSource GmbH.
```
ndoschek:
```suggestion
// Copyright (C) 2026 EclipseSource GmbH.
```

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -95,11 +125,68 @@ export class WorkspaceTrustService {
...
+            }
+        }
+
+        // For ALWAYS mode or ONCE mode with no stored decision, show dialog
+        return this.showTrustPromptDialog();
+    }
+
+    protected async showTrustPromptDialog(): Promise<boolean> {
+        const trust = nls.localizeByDefault('Trust');
+        const dontTrust = nls.localizeByDefault("Don't Trust");
```
ndoschek: I would prefer these default strings as VS Code uses:
```suggestion
        const trust = nls.localizeByDefault('Yes, I trust the authors');
        const dontTrust = nls.localizeByDefault("No, I don't trust the authors");
```

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -95,11 +125,68 @@ export class WorkspaceTrustService {
...
+
+        const dialog = new ConfirmDialog({
+            title: nls.localize('theia/workspace/trustDialogTitle', 'Do you trust the authors of this folder?'),
```
ndoschek:
```suggestion
            title: nls.localizeByDefault('Do you trust the authors of the files in this folder?'),
```

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -95,11 +125,68 @@ export class WorkspaceTrustService {
...
+            msg: nls.localize('theia/workspace/trustDialogMessage',
+                'If you trust the authors of this folder, code inside may be executed. Only trust folders that you trust the contents of.'),
```
ndoschek: I had a look at VS Code's dialog, and I like that they also again show the path of the folder that should be trusted.
Overall it looks a little more urgent if the dialog is larger, maybe we can also stretch the text we have a bit more to raise the user's attention a little for this.

Image

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -95,11 +125,68 @@ export class WorkspaceTrustService {
...
+        const workspaceUri = this.workspaceService.workspace?.resource;
+        if (!workspaceUri) {
+            return false;
+        }
+        const trustedFolders = this.workspaceTrustPref[WORKSPACE_TRUST_TRUSTED_FOLDERS] || [];
+        const normalizedWorkspaceUri = this.normalizeUri(workspaceUri.toString());
+        return trustedFolders.some(folder => this.normalizeUri(folder) === normalizedWorkspaceUri);
+    }
+
+    protected normalizeUri(uriStr: string): string {
```
ndoschek: This feels a bit weird to me, shouldn't we have something already in place to handle URIs, e.g. in the URI class itself?
And if not, maybe we should extract it to make it reusable?

packages/workspace/src/browser/workspace-commands.ts
```diff
@@ -42,9 +42,10 @@ import { ClipboardService } from '@theia/core/lib/browser/clipboard-service';
 
 const validFilename: (arg: string) => boolean = require('valid-filename');
 
+export const WORKSPACE_CATEGORY = 'Workspaces';
```
ndoschek: Do we really need to export this? I cannot see any usage outside of the WorkspaceCommands.

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -64,6 +71,15 @@ export class WorkspaceTrustService {
         await this.workspaceService.ready;
         await this.resolveWorkspaceTrust();
         this.preferences.onPreferenceChanged(change => this.handlePreferenceChange(change));
+
+        // Show banner if starting in restricted mode
```
ndoschek: Banner is not the correct term here, I'd prefer to call it notification to avoid any confusions.

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -64,6 +71,15 @@ export class WorkspaceTrustService {
...
+        const initialTrust = await this.getWorkspaceTrust();
+        this.updateRestrictedModeBanner(initialTrust);
+
+        // React to trust changes
```
ndoschek, line 99:
```suggestion
```

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -95,11 +126,77 @@ export class WorkspaceTrustService {
...
+    protected async showTrustPromptDialog(): Promise<boolean> {
+        const trust = nls.localizeByDefault('Yes, I trust the authors');
+        const dontTrust = nls.localizeByDefault("No, I don't trust the authors");
+        const folderPath = this.workspaceService.workspace?.resource?.path?.toString() ?? '';
+
+        const dialog = new ConfirmDialog({
+            title: nls.localizeByDefault('Do you trust the authors of the files in this folder?'),
+            msg: nls.localize('theia/workspace/trustDialogMessage',
+                'If you trust the authors of this folder, code inside may be executed. Only trust folders that you trust the contents of.') +
+                (folderPath ? `\n\n${folderPath}` : ''),
```
ndoschek: Can we wrap this folderPath in a code element to separate it visually from the above text?

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -141,6 +247,22 @@ export class WorkspaceTrustService {
...
+            this.restrictedModeNotificationShown = true;
+        } else if (trusted) {
+            this.restrictedModeNotificationShown = false;
+        }
+    }
+
+    protected showRestrictedModeNotification(): void {
+        this.messageService.warn(
+            nls.localize('theia/workspace/restrictedModeBanner',
+                'This workspace is in Restricted Mode. Some features may be disabled. Use "Manage Workspace Trust" command to change trust settings.')
```
ndoschek: I think it would be good to not have a timeout for the restricted mode popup, so the user has to dismiss it again manually, wdyt?

```suggestion
            nls.localize('theia/workspace/restrictedModeBanner',
                'This workspace is in Restricted Mode. Some features may be disabled. Use "Manage Workspace Trust" command to change trust settings.'),
            { timeout: -1 }
```

ndoschek (commented):
Thanks for the update Eugen!
It works much better now, but I still found a few cases which are not working correctly, yet.

1. After hitting Yes, I trust the authors in the dialog, I still get the restricted mode popup which is incocrrect, as the trusted path has been saved to the settings alrleady.
Also the additional notification from before keeps showing after I trust the workspace. Not sure if we should keep this at all? wdyt? In either case we should clear this popup if we trusted the workspace.

Image

2. If i cancel the trust dialog, the restriced mode popup shows as expected. But then the "Manage Workspace Trust" does not work if I try to select trust, this does not update the settings accordingly.

3. The dialog does not show for an empty window  if the setting is disabled:

Image

Overall it looks good to me, it also works fine with the remaining settings 👍 
Would be great if you could have another look, TIA!

eneufeld:
@ndoschek about `The dialog does not show for an empty window if the setting is disabled:`
So my understanding is:
emptyWindow: true → empty window trusted
emptyWindow: false → empty window in restricted mode

If we have an empty window we don't have a workspace that we can trust or not trust.
I fixed the issue of the dialog not showing on folder switch though.

ndoschek:
Hi @eneufeld, thank you for the updates!

About the emptyWindow setting: Ah that totally makes sense, thanks!
However unfortunately I always get the restricted popup independent of the setting (and it also does not listen to the setting change), could you have another look there please?

And unfortunately, point 1. from my previous comment is still not resolved for me:
- After hitting Yes, I trust the authors in the dialog, I still get the restricted mode popup which is incorrect, as the trusted path has been saved to the settings correctly.
- About the additional popup, I am not sure if we should keep that, if i cancel the new dialog for example and then hit yes in the popup nothing happens. I would suggest to remove the additional popup and integrate the info in the new dialog, WDYT?

And as mentioned above, the restricted popup does not listen to any settings change, either for the emptyWindow above. Also in this case: dismiss the dialog, restricted popup appears, executing the command and trust the workspace => the trust setting was updated for this folder, but the restricted popup is still showing and has to be dismissed manually. I think we should clear the restricted popup when the settings change.

Would be great if you could have another look.
Thanks for you patience with this, but if we can get these things fixed we should be good to get this merged, thanks!

packages/workspace/src/browser/workspace-frontend-contribution.ts
```diff
@@ -495,6 +503,38 @@ export class WorkspaceFrontendContribution implements CommandContribution, Keybi
...
 
+    protected async manageWorkspaceTrust(): Promise<void> {
+        const currentTrust = await this.workspaceTrustService.getWorkspaceTrust();
+        const trust = nls.localizeByDefault('Trust');
+        const dontTrust = nls.localizeByDefault("Don't Trust");
+
+        const items: QuickPickItem[] = [
+            {
+                label: trust,
+                description: currentTrust ? nls.localize('theia/workspace/currentTrustState', '(current)') : undefined
```
ndoschek:
```suggestion
                const currentSuffix = `(${nls.localizeByDefault('current')})`,
                ...
                description: currentTrust ? currentSuffix : undefined
```

And also for the item below

packages/plugin-ext/src/plugin/workspace.spec.ts
ndoschek, line 1: The test file creates emitters but doesn't dispose of them after tests complete. While this may not cause issues in isolated tests, it's good practice to clean up.

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -141,6 +278,29 @@ export class WorkspaceTrustService {
...
+    }
+
+    protected showRestrictedModeStatusBarItem(): void {
+        this.statusBar.setElement(WORKSPACE_TRUST_STATUS_BAR_ID, {
+            text: '$(shield) ' + nls.localizeByDefault('Restricted Mode'),
+            alignment: StatusBarAlignment.LEFT,
+            priority: 5000,
+            tooltip: nls.localize('theia/workspace/restrictedModeTooltip',
+                'Running in Restricted Mode. Some features are disabled because this folder is not trusted. Click to manage trust settings.'),
+            command: 'workspace:manageTrust'
```
ndoschek:
```suggestion
            command: MANAGE_WORKSPACE_TRUST.id
```

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -76,30 +99,109 @@ export class WorkspaceTrustService {
...
+            return false;
+        }
+        const trustedFolders = this.workspaceTrustPref[WORKSPACE_TRUST_TRUSTED_FOLDERS] || [];
+        const caseSensitive = !OS.backend.isWindows;
+        return trustedFolders.some(folder => {
+            // Strip trailing slash from folder string before creating URI
+            const normalizedFolder = folder.endsWith('/') ? folder.slice(0, -1) : folder;
+            const folderUri = new URI(normalizedFolder);
+            return workspaceUri.isEqual(folderUri, caseSensitive);
+        });
```
ndoschek, line 240: Could we use URI's built-in normalization?

```suggestion
        return trustedFolders.some(folder => {
           try {
              const folderUri = new URI(folder).normalizePath();
              return workspaceUri.normalizePath().isEqual(folderUri, caseSensitive);
           } catch {
              return false; // Invalid URI in preferences
           }
  });
```

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -62,8 +74,19 @@ export class WorkspaceTrustService {
...
+        this.workspaceService.onWorkspaceChanged(() => this.handleWorkspaceChanged());
+
+        // Show status bar item if starting in restricted mode
+        const initialTrust = await this.getWorkspaceTrust();
+        this.updateRestrictedModeIndicator(initialTrust);
+
+        // React to trust changes
+        this.onDidChangeWorkspaceTrust(trust => {
+            this.updateRestrictedModeIndicator(trust);
+        });
```
ndoschek: This subscription should be ideally stored in a disposable collection, as we never dispose it., probably also for the other subscriptions above.

packages/workspace/src/browser/workspace-trust-service.ts
```diff
@@ -76,30 +119,125 @@ export class WorkspaceTrustService {
...
+            const folderPath = this.workspaceService.workspace?.resource?.path?.toString() ?? '';
+
+            const dialog = new ConfirmDialog({
+                title: nls.localizeByDefault('Do you trust the authors of the files in this folder?'),
+                msg: nls.localize('theia/workspace/trustDialogMessage',
+                    'If you trust the authors of this folder, code inside may be executed. Only trust folders that you trust the contents of.') +
+                    (folderPath ? `\n\n"${folderPath}"` : ''),
+                ok: trust,
+                cancel: dontTrust,
+            });
```
colin-grant-work, line 197: Although this PR brings the management of workspace trust closer to VSCode's expectations, it doesn't bring the behavior of workspace trust (disabling various subsystems) any closer, so changing from the old message, which acknowledged the limited implementation, to this, which suggests that the feature is fully implemented, does not give the user an accurate picture of the behavior of the feature in its current state.

ndoschek: You're absolutely right, thanks for pointing that out.
It is indeed just the first step, and our plan is to start addressing the actual enforcement with the MCP servers next (GH-16872).

However, I was planning to open a follow up to this PR anyway to improve the styling a bit, so I will adjust the message text to better reflect the current, limited scope of the feature then with this follow up 👍

PR #16815, opened 2026-01-10
Switch writeFileReplacement to V2 Replacer

JonasHelming:
#### What it does

Use the V2 Replacer for "writeFileReplacement" function. The new replacer was already used for the "suggestFileReplacement" for quite a while and works more efficiently.

#### How to test

Use Theia Coder agent mode and observe the file replacement function calls. They should generally work. It will be hard to identify a difference while testing probably.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-ide/src/browser/file-changeset-functions.spec.ts
+++ b/packages/ai-ide/src/browser/file-changeset-functions.spec.ts
@@ -27,6 +27,7 @@ import {
     SuggestFileReplacements,
     SuggestFileReplacements_Simple,
     WriteFileReplacements,
+    WriteFileReplacements_Simple,
     ClearFileChanges,
     GetProposedFileState,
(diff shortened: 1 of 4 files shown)


packages/ai-ide/src/browser/file-changeset-functions.spec.ts
```diff
@@ -174,6 +193,12 @@ describe('File Changeset Functions Cancellation Tests', () => {
         expect(jsonResponse.error).to.equal('Operation cancelled by user');
     });
 
+    it('WriteFileReplacements should have correct ID', () => {
```
ndoschek, line 196: Nit: this test does not add much value, since it effectively asserts the same static string constant twice, so I'd prefer removing it.

PR #16816, opened 2026-01-10
Fix refresh after manualy changing task context

JonasHelming:
#### What it does

Fix a refresh bug for task context. If you work with an AI created task context and change it manually, the agent still works with the old version after saving the file (see how to test)

#### How to test
Before:
- Create a task context with architect
- Change it manually (e.g. add a new step)
- Run Coder (F1 => Chat: Task Context: Initiate Session)
- Observe that the change is not reflected. You can also stop the agent and ask to print out the task context
After:
The above case works reflecting the change

#### Follow-ups

I removed the cache for task context content, which I thin is a pragmatic solution. However, this might point to an issue with the file watcher we should investigate potentially..

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-ide/src/browser/task-context-file-storage-service.ts
+++ b/packages/ai-ide/src/browser/task-context-file-storage-service.ts
@@ -174,7 +174,10 @@ export class TaskContextFileStorageService implements TaskContextStorageService 
             return cached;
         }
         // Read fresh content from disk
-        const content = await this.fileService.read(cached.uri).then(read => read.value).catch(() => undefined);
+        const content = await this.fileService.read(cached.uri).then(read => read.value).catch(reason => {
+            this.logger.error(`Failed to read file ${cached.uri}: ${reason}`);
+            return undefined;
+        });
         if (content === undefined) {
             return cached; // Fall back to cache if read fails
         }


packages/ai-ide/src/browser/task-context-file-storage-service.ts
```diff
@@ -168,8 +168,18 @@ export class TaskContextFileStorageService implements TaskContextStorageService
...
 
-    get(identifier: string): Summary | undefined {
-        return this.inMemoryStorage.get(identifier);
+    async get(identifier: string): Promise<Summary | undefined> {
+        const cached = this.inMemoryStorage.get(identifier);
+        if (!cached?.uri) {
+            return cached;
+        }
+        // Read fresh content from disk
+        const content = await this.fileService.read(cached.uri).then(read => read.value).catch(() => undefined);
```
ndoschek: minor: this catch looks a bit odd, and if we already catch an error, we should also log it imo.

e.g. something like

```suggestion
        const content = await this.fileService.read(cached.uri).then(read => read.value).catch(reason => {
            this.logger.error(`Failed to read file ${cached.uri}: ${reason}`);
            return undefined;
        });
```

ndoschek (approved):
Thanks for the quick update! LGTM 👍

PR #16826, opened 2026-01-13
Pass widget consistently to toolbar enablement handlers

colin-grant-work:
#### What it does

Previously, the declaration and behavior of enablement methods on toolbar items was inconsistent:
- isVisible required a widget argument.
- isEnabled declared an optional widget argument but did pass that on to the underlying `isEnabled` evaluation.
- isToggled declared no widget argument and did not pass any on to the underlying command `isToggled` evaluation.

This PR makes the behavior of those handlers more consistent: all require a widget argument, and all pass it along to the underlying evaluation, if there is one.

#### How to test

A couple of test commits are provided.
1. Open a workspace.
2. On 'Bad test commit' you should see this in the explorer. The search sparkle icon should appear but clicking it will _not_ trigger a change in toggle state, because its toggle state depends on the widget being passed to the toggle handler.

image

4. On 'Good test commit' you should see this in the explorer. The search sparkle icon should appear and clicking it should trigger a change in toggle state.

image

5. On 'Final state,' the command should appear in the 'More' context menu. (On `master`, it does not appear because the event anchor was mistakenly being passed to these items when rendered in a context menu)

image

#### Breaking changes

> [!NOTE]
> Whether the changes are breaking is debatable: the implementation has been changed to have required arguments in places where they were previously optional, but the interface it was declared as implementing already made those arguments required in the public-facing methods.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/navigator/src/browser/navigator-contribution.ts
+++ b/packages/navigator/src/browser/navigator-contribution.ts
@@ -30,13 +30,11 @@ import {
     TabBar,
     Title,
     SHELL_TABBAR_CONTEXT_MENU,
-    OpenWithService,
-    codicon
+    OpenWithService
 } from '@theia/core/lib/browser';
 import { FileDownloadCommands } from '@theia/filesystem/lib/browser/download/file-download-command-contribution';
 import {
     CommandRegistry,
-    Event,
     isOSX,
     MenuModelRegistry,
     MenuPath,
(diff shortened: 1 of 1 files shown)


packages/navigator/src/browser/navigator-contribution.ts
```diff
@@ -560,12 +558,10 @@ export class FileNavigatorContribution extends AbstractViewContribution<FileNavi
 
         // More (...) toolbar items.
         this.registerMoreToolbarItem({
-            icon: codicon('search-sparkle'),
             id: FileNavigatorCommands.TOGGLE_AUTO_REVEAL.id,
             command: FileNavigatorCommands.TOGGLE_AUTO_REVEAL.id,
-            onDidChange: Event.map(Event.filter(this.preferenceService.onPreferencesChanged, e => 'explorer.autoReveal' in e), () => undefined),
```
colin-grant-work: The reason this goes away in the final state is that when this rendered in a context menu, its various enablements are evaluated at render time; when it's on a toolbar, it needs an additional signal to tell the toolbar to reevaluate those states. But this signal is present in both good and bad tests, so it is not what makes the difference.

Issue #15959, opened 2025-07-04
Audit code-formatting configuration in theia repository


ndoschek:
### Feature Description:

Developers report inconsistent or missing code formatting configuration across different file types when developing in the theia repository in the Theia IDE, suggesting that our current setup might not cover the full spectrum of files in the Theia repository.

- Inventory all file types in the theia repository
- Confirm formatter coverage for each extension
- Update workspace settings, `.editorconfig`, and recommended extensions as needed
- Optional: Consider adding a CI check that fails if `npm run format` produces changes?

Please comment if you have any thoughts or concerns

sdirix:
Some additional thoughts:

We already have a GH action which does a lint check. Some projects report formatting issues as lint issues. We could think about doing the same for Theia or choose a different more explicit approach, e.g. the `npm run format` check. Both are fine to me. [These are the thoughts of the prettier team](https://prettier.io/docs/integrating-with-linters), if someone is interested.

PR #16829, opened 2026-01-14
feat(editor): add formatter status bar with smart configuration managment

sgraband:
#### What it does
Introduce a new FormatterService abstraction and enhanced status bar
integration for managing document formatters across the editor.

New FormatterService (editor-formatter-service.ts):
- Define FormatterService interface for formatter management
- Provide formatter status, availability, and configuration APIs
- Support preference scope detection (User/Workspace/Folder)
- Emit events when formatters change

Monaco implementation (monaco-formatter-service.ts):
- Implement FormatterService using Monaco's language features
- Extract formatter metadata from registered providers
- Integrate with Theia's preference system for persistence

Formatter status contribution (editor-formatter-status-contribution.ts):
- Extract formatter-specific UI logic to dedicated contribution
- Handle display info for 0, 1, and multiple formatter scenarios
- Show warning/error icons for misconfigured formatters
- Provide quick pick for formatter selection with scope handling

Status bar integration (editor-language-status-service.ts):
- Add support for pinning language status items to the status bar
- Add Configure and Info buttons for formatter items in the tooltip
- Track pinned items to properly update on editor changes

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/editor/src/browser/language-status/editor-formatter-status-contribution.ts
+++ b/packages/editor/src/browser/language-status/editor-formatter-status-contribution.ts
@@ -339,8 +339,8 @@ export class EditorFormatterStatusContribution {
         // No formatter configured
         return {
             icon: FormatterIcon.Info,
-            label: nls.localize('theia/editor/nFormatters', '{0} formatters', formatterCount),
-            tooltip: nls.localize('theia/editor/noDefaultConfigured', 'No default formatter configured'),
+            label: nls.localize('theia/editor/noDefaultConfiguredLabel', 'No default formatter configured'),
+            tooltip: nls.localize('theia/editor/noDefaultConfiguredTooltip', 'No default formatter configured ({0} formatters available)', formatterCount),
             hasConfigureAction: true,
             severity: Severity.Info
         };
--- a/packages/editor/src/browser/language-status/editor-language-status-service.ts
+++ b/packages/editor/src/browser/language-status/editor-language-status-service.ts
@@ -287,8 +287,7 @@ export class EditorLanguageStatusService {
         const itemContainer = document.createElement('div');
         itemContainer.classList.add('hover-language-status');
 
-        const alwaysShowIcon = this.isFormatterItem(item);
-        itemContainer.appendChild(this.createSeverityIndicator(item.severity, alwaysShowIcon));
+        itemContainer.appendChild(this.createSeverityIndicator(item.severity));
 
         const textContainer = document.createElement('div');
         textContainer.className = 'element';
--- a/packages/monaco/src/browser/monaco-formatter-service.ts
+++ b/packages/monaco/src/browser/monaco-formatter-service.ts
@@ -15,7 +15,8 @@
 // *****************************************************************************
 
 import { inject, injectable, postConstruct } from '@theia/core/shared/inversify';
-import { Emitter, Event, PreferenceLanguageOverrideService, PreferenceScope, PreferenceService } from '@theia/core';
+import { Emitter, Event, PreferenceInspection, PreferenceLanguageOverrideService, PreferenceScope, PreferenceService } from '@theia/core';
+import { WorkspaceService } from '@theia/workspace/lib/browser';
 import { FormatterInfo, FormatterService, FormatterSettingScope, FormatterStatus } from '@theia/editor/lib/browser/editor-formatter-service';
 import { TextEditor } from '@theia/editor/lib/browser';
 import { StandaloneServices } from '@theia/monaco-editor-core/esm/vs/editor/standalone/browser/standaloneServices';
@@ -53,6 +54,9 @@ export class MonacoFormatterService implements FormatterService {
 
     @inject(PreferenceLanguageOverrideService)
     protected readonly preferenceSchema: PreferenceLanguageOverrideService;
+
+    @inject(WorkspaceService)
+    protected readonly workspaceService: WorkspaceService;
 
     protected readonly onDidChangeFormattersEmitter = new Emitter<void>();
     readonly onDidChangeFormatters: Event<void> = this.onDidChangeFormattersEmitter.event;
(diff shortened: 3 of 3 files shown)


packages/monaco/src/browser/monaco-formatter-service.ts
```diff
@@ -0,0 +1,212 @@
...
+
+        const languageFeaturesService = StandaloneServices.get(ILanguageFeaturesService);
+        const documentFormatters = languageFeaturesService.documentFormattingEditProvider.ordered(model);
+        const rangeFormatters = languageFeaturesService.documentRangeFormattingEditProvider.ordered(model);
+
+        return this.extractUniqueFormatters([...documentFormatters, ...rangeFormatters]);
+    }
+
+    protected getEditorModel(editor: TextEditor): ITextModel | undefined {
+        const monacoEditor = MonacoEditor.get(this.editorManager.currentEditor);
```
ndoschek: The method accepts an editor parameter but ignores it, always using this.editorManager.currentEditor instead, shouldn't it use the parameter instead?
```suggestion
        const monacoEditor = MonacoEditor.get(editor);
```

packages/editor/src/browser/editor-formatter-service.ts
```diff
@@ -0,0 +1,84 @@
...
+import { TextEditor } from './editor';
+
+export const FormatterService = Symbol('FormatterService');
+
+export interface FormatterInfo {
+    id: string;
+    displayName: string;
+}
+
+export type FormatterSettingScope = 'user' | 'workspace' | 'auto' | 'none';
```
ndoschek: I am not sure about this, this does not list the folder scope.
I would suggest to extend FormatterSettingScope to include 'folder' to match the preference scopes. Or if not, at least document this here then.
Also, independently, I think this could use some documentation to explain the different formatting scopes.

packages/editor/src/browser/language-status/editor-language-status-service.ts
```diff
@@ -114,6 +157,246 @@ export class EditorLanguageStatusService {
...
+
+        const formatterId = selectedFormatter.id === '' ? undefined : selectedFormatter.id;
+        await this.formatterService.setDefaultFormatter(editor, formatterId, targetScope);
+    }
+
+    protected async showFormatterSelectionPick(
+        editor: TextEditor,
+        formatters: ReturnType<FormatterService['getAvailableFormatters']>
+    ): Promise<QuickPickItem | undefined> {
+        const status = this.formatterService!.getFormatterStatus(editor);
```
ndoschek: After the guard clause at lines 303-305, all subsequent uses of FormatterService and QuickInputService within that method scope are safe and don't need `!`, so I think we should remove `!` operators after the guard clauses.

packages/editor/src/browser/language-status/editor-language-status-service.ts
```diff
@@ -114,6 +157,246 @@ export class EditorLanguageStatusService {
...
+
+        if (currentScope === PreferenceScope.Workspace) {
+            return PreferenceScope.Workspace;
+        }
+
+        if (currentScope === PreferenceScope.User) {
+            return this.showScopeSelectionPick();
+        }
+
+        return this.showScopeSelectionPick();
```
ndoschek: This could be simplified:
```suggestion
      if (currentScope === PreferenceScope.Folder || currentScope === PreferenceScope.Workspace) {
          return currentScope;                                                                                                  
      }
      return this.showScopeSelectionPick();
```

packages/editor/src/browser/language-status/editor-language-status-service.ts
ndoschek, line 1: We could also consider extracting formatter-related functionality into a separate class for easier maintenance and better adherence to Single Responsibility Principle.

ndoschek (commented):
Hi @sgraband, thanks for this improvement!
Overall it looks solid to me already, but I noticed one issue and have a few suggestions, would be great if you can take another look, TIA!

1. I have one issue, when testing the formatter item in the Theia example app.
When I try to switch the formatter, e.g. to the markdown table formatter, which is not configured as default one, I get this console error:
```
preference-service.ts:455 Uncaught (in promise) Error: Unable to write to Folder Settings because no resource is provided.
    at PreferenceServiceImpl.set (preference-service.ts:455:19)
    at MonacoFormatterService.setDefaultFormatter (monaco-formatter-service.ts:184:38)
    at EditorLanguageStatusService.showFormatterQuickPick (editor-language-status-service.ts:328:37)
```

2. I would prefer if we integrate the formatter item in the languageStatusItem ( `{ }` ). This way the user could still pin the item if needed and we do not show it by default.

Image


Also, could you please update the PR template, e.g. add a few test instructions, also for later reference.

PS: As discussed offline, I found a few issues with general formatting and configuring the formatting overall, but I will address this separately (GH-15959) after this PR is done.

packages/editor/src/browser/language-status/editor-language-status-service.ts
```diff
@@ -236,21 +432,24 @@ export class EditorLanguageStatusService {
...
+        this.updateLanguageStatusItems(editor);
     }
 
     protected getSeverityIconClasses(severity: Severity): string {
         switch (severity) {
             case Severity.Error: return codicon('error');
-            case Severity.Warning: return codicon('info');
-            default: return codicon('check');
+            case Severity.Warning: return codicon('warning');
+            default: return codicon('info');
```
ndoschek, line 451: nit: I would prefer to not show an icon by default to align with the other entries. This way warning and error also become more noticable to the user.

packages/editor/src/browser/language-status/editor-formatter-status-contribution.ts
```diff
@@ -0,0 +1,420 @@
...
+                ),
+                hasConfigureAction: true,
+                severity: Severity.Info
+            };
+        }
+
+        // No formatter configured
+        return {
+            icon: FormatterIcon.Info,
+            label: nls.localize('theia/editor/nFormatters', '{0} formatters', formatterCount),
```
ndoschek: I think it would be nice to show "No default formatter configured" as in the info message.
Perhaps also "No default formatter configured ({0} formatters available)" if this is not too long for this label.
Maybe we should also use Severity.warning for this case? Not sure though.

packages/editor/src/browser/language-status/editor-language-status.css
```diff
@@ -81,11 +86,20 @@
 
   /* ADDED STYLES - NOT FROM VSCODE */
   text-decoration: none;
+  cursor: pointer;
+  padding: 2px 6px;
```
ndoschek, line 90: nit: we could reuse theia CSS variables here and for the margin below

PR #16844, opened 2026-01-19
fix: improve tool call parameters rendering

### LINKED ISSUE #16029: Improve ToolCall parameters rendering in Chat
### Current state

Currently the tool call parameters are hidden completely behind 3 dots, e.g. see here:

Image

 The dots can be expanded, however the UI then looks a bit clunky:

Image

### Desired state

It would be nice if the UI could show a condensed version of the parameters initially, e.g. similar to this:

Image

A slightly different color would help to make it more readable. Then when the user clicks on the parameter, they could see the full variant. Or we show the full version of the parameter just in a Hover/Tooltip.

eneufeld:
#### What it does

Change the rendering of the tool call parameters to:
- show no key if only one parameter is passed
- show `...` if there are multiple parameters passed
- improve collapse/expand rendering for parameters

fixes #16029

#### How to test

Use any agent and let it call a function which expects a parameter.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
+++ b/packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
@@ -144,7 +144,7 @@ export class ToolCallPartRenderer implements ChatResponsePartRenderer<ToolCallCh
             return;
         }
         const prettyArgs = this.prettyPrintArgs(response.arguments);
-        const markdownString = new MarkdownStringImpl(`**${response.name}**\n\`\`\`json\n${prettyArgs}\n\`\`\``);
+        const markdownString = new MarkdownStringImpl(`**${response.name}**\n`).appendCodeblock('json', prettyArgs);
         this.hoverService.requestHover({
             content: markdownString,
             target,
--- a/packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.spec.ts
+++ b/packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.spec.ts
@@ -36,9 +36,9 @@ describe('condenseArguments', () => {
         expect(condenseArguments('{invalid}')).to.equal('{invalid}');
     });
 
-    it('condenses single string parameter as key: value', () => {
+    it('condenses single string parameter as value only', () => {
         const result = condenseArguments('{"query": "search term"}');
-        expect(result).to.equal('query: search term');
+        expect(result).to.equal('search term');
     });
 
     it('shows all key-value pairs for multiple parameters', () => {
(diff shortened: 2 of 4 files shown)


sdirix:
I agree, nested collapsibles should be avoided. As the arguments are always objects we could also specialize the rendering of it a bit. So it looks less like a JSON and more like a list of attributes.

In many cases we could also just not render the `...` in the title but immediately render the most important part. So for the example of Nina, this could be
```
Ran findFilesbyPattern (**/*.md ...)
```

With the `...` indicating that this was not the full argument.

---

This can't be solved generically. Each tool needs to have something like "getArgumentsShortLabel" which then returns something like `{ label: string, hasMore: boolean}`. The default implementation for a tool which does not provide that (e.g. MCP) could be the implementation of the current PR.

This would improve readability by a lot, often not even requiring the user to expand to see what happened.

packages/ai-chat-ui/src/browser/chat-response-renderer/toolcall-part-renderer.tsx
```diff
@@ -115,17 +121,35 @@ export class ToolCallPartRenderer implements ChatResponsePartRenderer<ToolCallCh
...
-                <summary className="collapsible-arguments-summary">...</summary>
-                <span>{this.prettyPrintArgs(args)}</span>
-            </details>
-        );
+    protected showArgsTooltip(response: ToolCallChatResponseContent, target: HTMLElement | undefined): void {
+        if (!target || !response.arguments || !response.arguments.trim() || response.arguments.trim() === '{}') {
+            return;
+        }
+        const prettyArgs = this.prettyPrintArgs(response.arguments);
+        const markdownString = new MarkdownStringImpl(`**${response.name}**\n\`\`\`json\n${prettyArgs}\n\`\`\``);
```
ndoschek: This can be simplified:
```suggestion
        const markdownString = new MarkdownStringImpl(`**${response.name}**\n`).appendCodeblock('json', prettyArgs);
```

Also to preserve the indentations of the pretty json, we could add this css rule:
```css
.theia-hover div.rendered-markdown {
  white-space: pre;
}
```

JonasHelming:
@eneufeld @ndoschek 
This does not look good to me:

image

JonasHelming:
Also I am not sure about the hover in general. If the mouse is above the chat it sometimes "flows" under the mouse pointer and opens arbitrary hovers then. Finally, we lost the ability to see the actual content of a tool call completly.

Sorry for raising these issues, In general i like this improvement a lot and I think is was long outstanding!

eneufeld:
@JonasHelming who checks writeFile content ;-) 

Seriously, good catch.
Can you explain what you mean by `Finally, we lost the ability to see the actual content of a tool call completly.`
I will open a follow up, maybe you can explain there so we can fix.

eneufeld:
@JonasHelming Thank you for raising the issues, can you elaborate here: https://github.com/eclipse-theia/theia/issues/16988

PR #16846, opened 2026-01-19
Fix uses of `newValue`

### LINKED ISSUE #16645: Preference change listeners should handle undefined values for newValue properly

### LINKED ISSUE #16821: The Toggle Status Bar Visibility command cannot display the status bar properly
### Bug Description:
When clicking Toggel Status Bar Visibility in the menu bar, the status bar cannot be displayed properly

### Steps to Reproduce:

1.When clicking Toggel Status Bar Visibility for the first time, the status bar is hidden normally
2.When clicked again, the status bar cannot be displayed properly

### Additional Information
The reason for this issue is that when the status bar switches to display, the newValue obtained by onPreferenceChanging is undefined

The newValue returned by onPreferenceChanged is undefined, which is related to the following code

Image

- Operating System:Mac
- Theia Version:1.65.100

colin-grant-work:
#### How to test

Fixes #16645
Fixes #16821

Offers two alternatives:

- First commit tries to make `newValue` meaningful. Then it would be (kind of) safe to use. I think that would be nice, but it would be a somewhat large change to the API. We're very clear that you _shouldn't_ use `newValue` as new value (makes one wonder why we include it), and we do provide a `scope` in which the change occurred, and giving a different value might be misleading.
- Second commit reverts that
- Third commit goes through and replaces uses of `newValue` the result of getting the new value. This is the safest.

#### Breaking changes

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/workspace/src/browser/workspace-trust-service.spec.ts
+++ b/packages/workspace/src/browser/workspace-trust-service.spec.ts
@@ -180,6 +180,7 @@ describe('WorkspaceTrustService', () => {
             it('should update trust to true when emptyWindow setting changes to true for empty window', async () => {
                 service.setCurrentTrust(false);
                 workspaceServiceStub.workspace = undefined;
+                workspaceTrustPrefStub[WORKSPACE_TRUST_EMPTY_WINDOW] = true;
 
                 const change: PreferenceChange = {
                     preferenceName: WORKSPACE_TRUST_EMPTY_WINDOW,
@@ -228,6 +229,7 @@ describe('WorkspaceTrustService', () => {
             it('should not update trust when emptyWindow setting changes but trust already matches', async () => {
                 service.setCurrentTrust(true);
                 workspaceServiceStub.workspace = undefined;
+                workspaceTrustPrefStub[WORKSPACE_TRUST_EMPTY_WINDOW] = true;
 
                 const change: PreferenceChange = {
                     preferenceName: WORKSPACE_TRUST_EMPTY_WINDOW,


packages/ai-ollama/src/browser/ollama-frontend-application-contribution.ts
```diff
@@ -43,9 +43,9 @@ export class OllamaFrontendApplicationContribution implements FrontendApplicatio
 
             this.preferenceService.onPreferenceChanged(event => {
                 if (event.preferenceName === HOST_PREF) {
-                    this.manager.setHost(event.newValue as string);
+                    this.manager.setHost(this.preferenceService.get<string>(HOST_PREF) || undefined);
```
ndoschek:
```suggestion
                    this.manager.setHost(this.preferenceService.get<string>(HOST_PREF));
```

packages/core/src/browser/request/browser-request-service.ts
```diff
@@ -40,9 +40,9 @@ export abstract class AbstractBrowserRequestService implements RequestService {
         });
         this.preferenceService.onPreferencesChanged(e => {
             this.configurePromise.then(() => this.configure({
-                proxyUrl: e['http.proxy']?.newValue as string,
-                proxyAuthorization: e['http.proxyAuthorization']?.newValue as string,
-                strictSSL: e['http.proxyStrictSSL']?.newValue as boolean
+                proxyUrl: this.preferenceService.get('http.proxy') as string,
```
ndoschek: Should we also use get<string>(...) instead of the cast as we do for almost all other cases?
```suggestion
                proxyUrl: this.preferenceService.get<string>('http.proxy'),
```

packages/core/src/browser/menu/browser-menu-plugin.ts
ndoschek, line 1: For this setting there seems to be an issue (for both browser and electron though), when I change the setting to compact and back/reset to classic, the menu does not show again as expected, but only shows a ... button.
But this was not introduced with this PR, so I'm going to open a follow up for this.

packages/terminal-manager/src/browser/terminal-manager-frontend-contribution.ts
```diff
@@ -49,7 +49,7 @@ export class TerminalManagerFrontendContribution implements FrontendApplicationC
         this.preferenceService.ready.then(() => {
             this.preferenceService.onPreferenceChanged(change => {
                 if (change.preferenceName === 'terminal.grouping.mode') {
-                    this.handleTabsDisplayChange(change.newValue as string);
+                    this.handleTabsDisplayChange(this.preferences.get('terminal.grouping.mode') as string);
```
ndoschek: Same here:

```suggestion
                    this.handleTabsDisplayChange(this.preferences.get<string>('terminal.grouping.mode'));
```

examples/api-tests/src/preferences.spec.js
```diff
@@ -200,10 +200,10 @@ describe('Preferences', function () {
         const results = await Promise.allSettled(promises);
         const expectedValues = { [searchPref]: searchDebounce, [channelPref]: channelHistory, [hoverPref]: hoverDelay };
         const actualValues = { [searchPref]: preferenceService.get(searchPref), [channelPref]: preferenceService.get(channelPref), [hoverPref]: preferenceService.get(hoverPref), }
-        const eventValues = event && Object.keys(event).reduce((accu, key) => { accu[key] = event[key].newValue; return accu; }, {});
+        const eventKeys = event && Object.keys(event).sort();
         toDispose.dispose();
         assert(results.every(setting => setting.status === 'fulfilled'), 'All promises should have resolved rather than rejected.');
-        assert.deepEqual(actualValues, eventValues, 'The event should reflect the current state of the service.');
+        assert.deepEqual([channelPref, hoverPref, searchPref], eventKeys, 'The event should contain the changed preference names.');
```
ndoschek: One more thing I just noticed: we seem to be hitting issues with this equality check in the CI runs on Node 20. Maybe we could use a less strict comparison for the array?

packages/core/src/browser/request/browser-request-service.ts
```diff
@@ -40,9 +40,9 @@ export abstract class AbstractBrowserRequestService implements RequestService {
         });
         this.preferenceService.onPreferencesChanged(e => {
             this.configurePromise.then(() => this.configure({
-                proxyUrl: e['http.proxy']?.newValue as string,
-                proxyAuthorization: e['http.proxyAuthorization']?.newValue as string,
-                strictSSL: e['http.proxyStrictSSL']?.newValue as boolean
+                proxyUrl: this.preferenceService.get<string>('http.proxy'),
+                proxyAuthorization: this.preferenceService.get('http.proxyAuthorization') as string,
+                strictSSL: this.preferenceService.get('http.proxyStrictSSL') as boolean
```
ndoschek: nit: sorry I overlooked those two lines earlier. Also above in the lines 31-33 we could apply the same for these three preferences.

```suggestion
                proxyAuthorization: this.preferenceService.get<string>('http.proxyAuthorization'),
                strictSSL: this.preferenceService.get<boolean>('http.proxyStrictSSL')
```

PR #16847, opened 2026-01-19
Store persistent chat sessions in the workspace by default

cdamus:
#### What it does

Implements storage of chat session persistence — initially implemented with #16486 — in the workspace by default.

A new complex preference settings is added with three details:

- **Session Storage: Scope** — enum of values `'workspace'` and `'global'`, workspace being the default, determining whether chat session storage is per-workspace (and so _in_ the workspace root directory) or global (shared by all workspaces, as previously). It has two sub-preferences:
  - **Workspace Path** — a workspace-relative path in which to store chat sessions when storage scope is `'workspace'`. The default is `.theia/chatSessions`. Validation enforces a nonempty relative path within three workspace
  - **Global Path** — an absolute path in the filesystem in which to store chat sessions when storage scope is `'global'`. The default is `$HOME/.theia/chatSessions`. Validation enforces a nonempty absolute path

The object-valued preference setting is edited in a new custom preference renderer that organizes the path inputs behind a disclosure triangle. Each of these inputs has its own revert button to restore the default setting. The preference object persists only those of its properties that do not have their default value. This ensures that if a workspace overrides the preference, user-specific default paths are not persisted in version control repositories.

#### How to test

Play with the new preferences for chat session storage and see how they interact with the "Show Chats" command. You should be able to store a different set of chats in global storage versus workspace storage, and set up multiple stores of each kind via the path preferences. Whatever the preference settings, the Show Chats command should show what actually is in that store and it should not be available when there are no chats, neither in the in-memory history nor the store.

#### Follow-ups

- should there be general interest in it, the complex setting with a master preference and disclosable details could be extracted into a reusable renderer. It would probably only be useful though if it could itself reuse the rest of the renderer library for rendering its details, which would imply a considerable refactoring of the leaf renderer framework to support this sub-preference use case (prop of a preference object value) in addition to actual preferences

#### Breaking changes

I suppose that, technically, changing the default storage location of chat sessions breaks the behaviour previously established and will appear at first to be a case of data loss for users, although reversion to the global scope is available. It does not seem worth the effort to provide a tool to import the chat session store into the current workspace.

There is no meaningful _API breaking_ change.

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/ai-chat-ui/src/browser/session-storage-preference-renderer.ts
+++ b/packages/ai-chat-ui/src/browser/session-storage-preference-renderer.ts
@@ -34,6 +34,7 @@ import {
     SessionStorageValue
 } from '@theia/ai-chat/lib/common/ai-chat-preferences';
 import { SessionStorageDefaultsProvider } from '@theia/ai-chat/lib/browser/session-storage-defaults-provider';
+import { WorkspaceService } from '@theia/workspace/lib/browser/workspace-service';
 import debounce = require('@theia/core/shared/lodash.debounce');
 
 /**
@@ -94,6 +95,9 @@ export class SessionStoragePreferenceRenderer extends PreferenceLeafNodeRenderer
 
     @inject(SessionStorageDefaultsProvider)
     protected readonly defaultsProvider: SessionStorageDefaultsProvider;
+
+    @inject(WorkspaceService)
+    protected readonly workspaceService: WorkspaceService;
 
     protected scopeSelectRef = React.createRef<SelectComponent>();
     protected expandedSection: HTMLDivElement;
@@ -201,7 +205,9 @@ export class SessionStoragePreferenceRenderer extends PreferenceLeafNodeRenderer
     protected createGlobalPathDelegate(): InputRowDelegate {
         return this.createPathDelegate(
             () => this.defaultsProvider.getDefaultGlobalPath(),
-            scope => scope === 'global',
+            // Global path is active when scope is 'global' OR when no workspace is available
+            // (since ChatSessionStoreImpl falls back to global storage without a workspace)
+            scope => scope === 'global' || this.workspaceService.tryGetRoots().length === 0,
             value => value.globalPath,
             (currentValue, newInputValue) => ({ ...currentValue, globalPath: newInputValue }),
             (value: string) => {
@@ -352,6 +358,20 @@ export class SessionStoragePreferenceRenderer extends PreferenceLeafNodeRenderer
     }
 
     protected createInteractable(parent: HTMLElement): void {
+        const hasWorkspace = this.workspaceService.tryGetRoots().length > 0;
+
+        if (hasWorkspace) {
+            this.createFullInteractable(parent);
+        } else {
+            this.createSimplifiedInteractable(parent);
+        }
+    }
+
+    /**
+     * Create the full UI with scope dropdown, disclosure button, and expandable path settings.
+     * Used when a workspace is available.
+     */
+    protected createFullInteractable(parent: HTMLElement): void {
         // Initialize dynamic defaults asynchronously - will update UI when ready
         this.initDynamicDefaults();
 
@@ -419,7 +439,8 @@ export class SessionStoragePreferenceRenderer extends PreferenceLeafNodeRenderer
             SessionStorageValue.Labels.workspacePathDescription(),
             currentValue?.workspacePath ?? this.workspacePathDelegate.getDefaultValue(),
             !this.workspacePathDelegate.isActiveForScope(currentValue?.scope ?? 'workspace'),
-            this.defaultsProvider.getDefaultWorkspacePath()
+            this.defaultsProvider.getDefaultWorkspacePath(),
+            true
         );
         this.expandedSection.appendChild(workspacePathRow);
 
@@ -430,11 +451,46 @@ export class SessionStoragePreferenceRenderer extends PreferenceLeafNodeRenderer
             SessionStorageValue.Labels.globalPathDescription(),
             currentValue?.globalPath ?? this.globalPathDelegate.getDefaultValue(),
             !this.globalPathDelegate.isActiveForScope(currentValue?.scope ?? 'workspace'),
-            this.defaultsProvider.getDefaultGlobalPath()
+            this.defaultsProvider.getDefaultGlobalPath(),
+            true
         );
         this.expandedSection.appendChild(globalPathRow);
 
         container.appendChild(this.expandedSection);
+
+        parent.appendChild(container);
+    }
+
+    /**
+     * Create a simplified UI with just the global storage path input.
+     * Used when no workspace is available - scope falls back to 'global'.
+     * No additional label/description needed since the preference title "Session Storage" is sufficient.
+     */
+    protected createSimplifiedInteractable(parent: HTMLElement): void {
+        // Initialize dynamic defaults asynchronously
+        this.initDynamicDefaults();
+
+        const container = document.createElement('div');
+        container.classList.add('session-storage-preference-container', 'session-storage-no-workspace', 'pref-input');
+        this.interactable = container;
+
+        const input = document.createElement('input');
+        input.type = 'text';
+        input.classList.add('theia-input', 'session-storage-path-input');
+        input.value = this.getValueWithDefaults().globalPath;
+        input.placeholder = this.defaultsProvider.getDefaultGlobalPath();
+        input.oninput = () => this.globalPathDelegate.handleInputChange();
+        input.onblur = () => this.globalPathDelegate.handleInputBlur();
+
+        this.globalPathDelegate.setInput(input);
+
+        // Message element for validation errors
+        const messageElement = document.createElement('div');
+        messageElement.classList.add('session-storage-message');
+        this.globalPathDelegate.setMessageElement(messageElement);
+        this.globalPathDelegate.setRow(container);
+
+        container.appendChild(input);
 
         parent.appendChild(container);
     }
@@ -446,7 +502,8 @@ export class SessionStoragePreferenceRenderer extends PreferenceLeafNodeRenderer
         description: string,
         value: string,
         isInactive: boolean,
-        placeholder: string
+        placeholder: string,
+        showResetButton: boolean = true
     ): HTMLDivElement {
         const row = document.createElement('div');
         row.classList.add('session-storage-path-row');
@@ -476,14 +533,16 @@ export class SessionStoragePreferenceRenderer extends PreferenceLeafNodeRenderer
 
         inputContainer.appendChild(input);
 
-        // Reset button
-        const resetButton = document.createElement('button');
-        resetButton.classList.add('theia-button', 'session-storage-reset-button');
-        resetButton.title = SessionStorageValue.Labels.resetToDefault();
-        resetButton.innerHTML = `<i class="${codicon('discard')}"></i>`;
-        resetButton.disabled = value === delegate.getDefaultValue();
-        resetButton.onclick = () => delegate.handleReset();
-        inputContainer.appendChild(resetButton);
+        if (showResetButton) {
+            // Reset button
+            const resetButton = document.createElement('button');
+            resetButton.classList.add('theia-button', 'session-storage-reset-button');
+            resetButton.title = SessionStorageValue.Labels.resetToDefault();
+            resetButton.innerHTML = `<i class="${codicon('discard')}"></i>`;
+            resetButton.disabled = value === delegate.getDefaultValue();
+            resetButton.onclick = () => delegate.handleReset();
+            inputContainer.appendChild(resetButton);
+        }
 
         // Message element for errors and info (with severity class)
         const messageElement = document.createElement('div');
--- a/packages/ai-chat-ui/src/browser/style/session-storage-preference.css
+++ b/packages/ai-chat-ui/src/browser/style/session-storage-preference.css
@@ -27,10 +27,6 @@
     margin-bottom: var(--theia-ui-padding);
 }
 
-.theia-settings-container .session-storage-scope-select {
-    max-width: 320px;
-}
-
 .theia-settings-container .session-storage-disclosure-row {
     display: flex;
     align-items: center;
@@ -99,6 +95,11 @@
     max-width: 320px;
 }
 
+/* In simplified no-workspace view, the input fills the pref-input container */
+.theia-settings-container .session-storage-no-workspace > .session-storage-path-input {
+    width: 100%;
+}
+
 .theia-settings-container .session-storage-reset-button {
     padding: 2px 6px;
     min-width: auto;
@@ -110,7 +111,8 @@
 }
 
 /* Message styling for session storage path (errors and info) */
-.theia-settings-container .session-storage-path-row .session-storage-message {
+.theia-settings-container .session-storage-path-row .session-storage-message,
+.theia-settings-container .session-storage-no-workspace .session-storage-message {
     border-style: solid;
     border-width: 1px;
     max-width: 320px;
@@ -119,12 +121,14 @@
     margin-top: calc(var(--theia-ui-padding) / 2);
 }
 
-.theia-settings-container .session-storage-path-row .session-storage-message.error {
+.theia-settings-container .session-storage-path-row .session-storage-message.error,
+.theia-settings-container .session-storage-no-workspace .session-storage-message.error {
     border-color: var(--theia-inputValidation-errorBorder);
     background-color: var(--theia-inputValidation-errorBackground);
 }
 
-.theia-settings-container .session-storage-path-row .session-storage-message.info {
+.theia-settings-container .session-storage-path-row .session-storage-message.info,
+.theia-settings-container .session-storage-no-workspace .session-storage-message.info {
     border-color: var(--theia-inputValidation-infoBorder);
     background-color: var(--theia-inputValidation-infoBackground);
 }
--- a/packages/ai-chat/src/browser/ai-chat-preference-contribution.spec.ts
+++ b/packages/ai-chat/src/browser/ai-chat-preference-contribution.spec.ts
@@ -1,0 +1,103 @@
+// *****************************************************************************
+// Copyright (C) 2026 EclipseSource GmbH.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License v. 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0.
+//
+// This Source Code may also be made available under the following Secondary
+// Licenses when the conditions for such availability set forth in the Eclipse
+// Public License v. 2.0 are satisfied: GNU General Public License, version 2
+// with the GNU Classpath Exception which is available at
+// https://www.gnu.org/software/classpath/license.html.
+//
+// SPDX-License-Identifier: EPL-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0
+// *****************************************************************************
+
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+import { Container } from '@theia/core/shared/inversify';
+import { AIChatPreferenceContribution } from './ai-chat-preference-contribution';
+import { SessionStorageDefaultsProvider } from './session-storage-defaults-provider';
+import { PreferenceSchemaService } from '@theia/core/lib/common/preferences/preference-schema';
+import { SESSION_STORAGE_PREF, SessionStorageValue } from '../common/ai-chat-preferences';
+import { Disposable } from '@theia/core/lib/common/disposable';
+
+describe('AIChatPreferenceContribution', () => {
+    let sandbox: sinon.SinonSandbox;
+    let container: Container;
+    let contribution: AIChatPreferenceContribution;
+    let mockDefaultsProvider: {
+        initialize: sinon.SinonStub;
+        getDefaultValue: sinon.SinonStub;
+    };
+    let mockSchemaService: {
+        registerOverride: sinon.SinonStub;
+    };
+
+    const DEFAULT_GLOBAL_PATH = '/__test__/mock-config/chatSessions';
+
+    const DEFAULT_STORAGE_VALUE: SessionStorageValue = {
+        scope: 'workspace',
+        workspacePath: '.theia/chatSessions',
+        globalPath: DEFAULT_GLOBAL_PATH
+    };
+
+    beforeEach(() => {
+        sandbox = sinon.createSandbox();
+
+        container = new Container();
+
+        mockDefaultsProvider = {
+            initialize: sandbox.stub().resolves(),
+            getDefaultValue: sandbox.stub().returns(DEFAULT_STORAGE_VALUE)
+        };
+
+        mockSchemaService = {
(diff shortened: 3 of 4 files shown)


colin-grant-work:
General comments before I take a look at the code:

> This PR includes a very minimal implementation of an optional visibleWhen property on a preference schema that allows the preference to condition its visibility in the UI upon the value of another preference.

I'm likely not that keen on this feature. I understand that it has some value for avoiding confusion - you see the path setting for the storage scope you're using, and not the other one that would have no effect - but it introduces complexity into the preferences UI that isn't completely transparent to the user. I can very easily imagine some user seeing mention of one of the scope preferences, then opening up the preferences only to find that they don't see that particular scope's path preference, then reaching out to their IDE team (this used to be me, whence the vividness of this scenario in my mind) and then having them explain that to see _that_ setting, you first have to change _another_ setting. Of course the other scenario is likely as well - they reach out because they changed the wrong scope and didn't see the behavior they expected - but we could probably avoid that with some path validation: insist on absolute paths for the global one and relative for the workspace one.

> This is a tiny subset of the VS Code when-clause specification for preference settings (config.<preference-id>) supporting only equal-to or not-equal-to some value.

But if my arguments above aren't persuasive, I'd probably prefer just supporting any when clause, rather than restricting it to `config.` prefixed values. That will make it possibly even more opaque - that would be the responsibility of whoever chooses the `when` clause - but it will be easier to document and the code will be simpler and anticipate a likely feature request.

cdamus:
> > This PR includes a very minimal implementation of an optional visibleWhen property on a preference schema that allows the preference to condition its visibility in the UI upon the value of another preference.
> 
> I'm likely not that keen on this feature.

What I would prefer TBH would be possibly an even further departure from what is currently supported. I would like to see:

- both path settings at the same time
- but they are clearly _subordinate to_ the scope setting, so actually nested under it. By indentation and preferably with a disclosure triangle to reveal them
- (optionally) only the one pertinent to the current scope value is enabled, though both are visible

If we'd rather not make any changes to the preferences UI for this feature, then I would suggest an alternative:

- a single path setting
- it is always a relative path
- for global scope it is relative to `$HOME/`
- for workspace scope it is relative to the (first) workspace root
- the default is `.theia/chatSessions`

Overriding the path setting itself at workspace scope should give users all the flexibility they need.

colin-grant-work:
> What I would prefer TBH would be possibly an even further departure from what is currently supported. I would like to see

This would be possible in principle: the preference rendering system is extensible so that if we had a base preference `ai-features.chat.sessionStorage` with an object schema specifying the others, we could then assign a PreferenceRenderer to it that would look like what you're describing. That would be a bit ad-hoc, but I actually think it would be pretty useful, because the general idea of 'packaged' preferences that in some way relate to one another should be displayed as such isn't such a crazy idea.

The main entry point into that system is [preference-node-renderer-creator.ts](https://github.com/eclipse-theia/theia/blob/master/packages/preferences/src/browser/views/components/preference-node-renderer-creator.ts), if you're interested in going down that path.

cdamus:
> The main entry point into that system is [preference-node-renderer-creator.ts](https://github.com/eclipse-theia/theia/blob/master/packages/preferences/src/browser/views/components/preference-node-renderer-creator.ts), if you're interested in going down that path.

Cool, thanks for that! I'll give it a try. So hold off on review if you'd like 😀

cdamus:
Thanks @colin-grant-work for the pointers to custom preference renderers. I'm very much happier with the result of refactoring the UI in commit c6721045b3e1c883ca390c8475cbbdbc7043d647 for a complex preference setting. And I've reverted the changes in the preference framework for visible-when conditions.

packages/ai-chat-ui/src/browser/style/index.css
```diff
@@ -1,3 +1,5 @@
+@import url("./session-storage-preference.css");
```
colin-grant-work: More usually this would be imported by the TS that uses the styles.

package-lock.json
```diff
@@ -8789,7 +8789,6 @@
       "version": "12.2.3",
       "resolved": "https://registry.npmjs.org/@types/markdown-it/-/markdown-it-12.2.3.tgz",
       "integrity": "sha512-GKMHFfv3458yYy+v/N8gjufHO6MSZKCOXpZc5GXIWWy8uldwfmPn98vp81gZ5f9SVw8YYBctgfJ22a2d7AOMeQ==",
-      "dev": true,
```
colin-grant-work: @ndoschek, these are the kinds of changes I also see if I run `npm install` locally. The deletion of these lines is likely unintentional, but their addition was also recent and likely unintentional. Do we have a policy / desired target for this kind of metadata in the `package-lock.json`?

packages/ai-chat-ui/src/browser/session-storage-preference-renderer.ts
```diff
@@ -0,0 +1,426 @@
...
+                    resetButton.disabled = !isActive || isDefault;
+                }
+            },
+
+            syncFromPreference: (value: SessionStorageValue) => {
+                if (input) {
+                    if (document.activeElement !== input) {
+                        input.value = getPathFromValue(value);
+                    }
+                    input.disabled = !isActiveForScope(value.scope);
```
colin-grant-work: I still don't really like disabling an input because it's not the active scope. I'd prefer some kind of message or warning color but otherwise allowing the user to interact with it. That's definitely a personal preference, though, so I won't insist if other, savvier UX designers disagree.

packages/ai-chat-ui/src/browser/session-storage-preference-renderer.ts
```diff
@@ -0,0 +1,426 @@
...
+        // Disclosure triangle and expanded section
+        const disclosureRow = document.createElement('div');
+        disclosureRow.classList.add('session-storage-disclosure-row');
+
+        this.disclosureButton = document.createElement('span');
+        this.disclosureButton.classList.add('session-storage-disclosure-button');
+        this.disclosureButton.innerHTML = `<i class="${codicon('chevron-right')}"></i>`;
+        this.disclosureButton.onclick = () => this.toggleExpanded();
+        this.disclosureButton.setAttribute('role', 'button');
+        this.disclosureButton.setAttribute('tabindex', '0');
```
colin-grant-work, line 414: This is focusable but doesn't show the usual focus outline when focused.

packages/ai-chat/src/common/ai-chat-preferences.ts
```diff
@@ -15,12 +15,101 @@
...
+        export const workspacePathLabel = (): string =>
+            nls.localize('theia/ai/chat/sessionStorage/workspacePath', 'Workspace Path');
+        export const workspacePathDescription = (): string =>
+            nls.localize('theia/ai/chat/sessionStorage/workspacePath/description',
+                'Relative path within the workspace root where chat sessions are stored. If empty, then chats will not be persisted.');
+        export const globalPathLabel = (): string =>
+            nls.localize('theia/ai/chat/sessionStorage/globalPath', 'Global Path');
+        export const globalPathDescription = (): string =>
+            nls.localize('theia/ai/chat/sessionStorage/globalPath/description',
+                'Absolute filesystem path where chat sessions are stored globally. Leave empty to use the default location ($HOME/.theia/chatSessions).');
```
colin-grant-work: It's good to be explicit about what an empty value means in each case. I don't think it's ideal that they mean different things in different scopes.

I would probably prefer something like:
- Third 'scope' option is `None` to disable storage, rather than requiring Workspace + empty path to achieve that.
- Value is required (and validated as a path, ideally) --> empty disallowed

packages/ai-chat/src/common/ai-chat-preferences.ts
```diff
@@ -15,12 +15,101 @@
...
+    configurable: false,
+
+    get(): SessionStorageValue {
+        return {
+            scope: 'workspace',
+            workspacePath: '.theia/chatSessions',
+            globalPath: ''
+        };
+    },
+});
```
colin-grant-work: Why this style rather than something like `export const DEFAULT = Object.freeze({...})`?

packages/ai-chat/src/browser/chat-session-store-impl.ts
```diff
@@ -54,9 +58,33 @@ export class ChatSessionStoreImpl implements ChatSessionStore {
...
+            if (event.preferenceName === SESSION_STORAGE_PREF) {
+                this.logger.debug('Session storage preference changed: invalidating cache.', { preference: event.preferenceName });
+                this.invalidateStorageCache();
+            }
+        });
+
+        this.workspaceService.onWorkspaceLocationChanged(() => {
+            this.logger.debug('Workspace location changed: invalidating storage cache.');
+            this.invalidateStorageCache();
+        });
```
colin-grant-work: I don't think this is the right event for this context. This fires only when the file in which the workspace is stored changes, so it doesn't fire every time a new folder is added to or removed from an existing multi-root workspace, which could change where we store things. I think we want the `onWorkspaceChanged` event.

packages/ai-chat/src/browser/chat-session-store-impl.ts
```diff
@@ -201,6 +246,52 @@ export class ChatSessionStoreImpl implements ChatSessionStore {
...
+        const storageConfig = SessionStorageValue.create(storagePref);
+
+        if (storageConfig.scope === 'workspace') {
+            const workspaceRoot = await this.getWorkspaceRoot();
+            if (workspaceRoot) {
+                // Empty workspace path means persistence is disabled
+                if (!storageConfig.workspacePath.trim()) {
+                    this.logger.debug('Workspace storage path is empty: session persistence disabled.');
+                    return undefined;
+                }
```
colin-grant-work: This could be moved up to before we try to retrieve the workspaceRoot. (If we retain this behavior)

packages/ai-chat/src/browser/chat-session-store-impl.ts
```diff
@@ -201,6 +246,52 @@ export class ChatSessionStoreImpl implements ChatSessionStore {
...
+        this.logger.debug('Using default global storage path', { path: defaultPath.toString() });
+        return defaultPath;
+    }
+
+    protected async getWorkspaceRoot(): Promise<URI | undefined> {
+        const roots = await this.workspaceService.roots;
+        if (roots.length > 0) {
+            return roots[0].resource;
+        }
+        return undefined;
```
colin-grant-work: If we're watching for workspace changes (and it looks like we are), this doesn't really need to be the async version; you can use `this.workspaceService.tryGetRoots` to get the current list, and if the list changes, we'll update anyway.

packages/ai-chat/src/browser/chat-session-store-impl.ts
```diff
@@ -201,6 +246,52 @@ export class ChatSessionStoreImpl implements ChatSessionStore {
...
+        // Global storage mode (or fallback)
+        if (storageConfig.globalPath.trim()) {
+            // Custom absolute path specified
+            const resolvedPath = new URI(storageConfig.globalPath).withScheme('file');
+            this.logger.debug('Using custom global storage path', { path: resolvedPath.toString() });
+            return resolvedPath;
+        }
+
+        // Default global storage: $HOME/.theia/chatSessions
+        const configDir = await this.envServer.getConfigDirUri();
```
colin-grant-work: Since the config dir is downstream-configurable, we don't actually know that it will be `.theia`. This affects the logic in the preference description.

packages/ai-chat/src/common/ai-chat-preferences.ts
colin-grant-work: This message shouldn't mention `.theia` - that may not be the storage directory name for downstream applications. If we want to mention the directory, we'll have to make this asynchronous so that we can match it to the `configDirUri`.

(Also, more minorly, the format for describing the location ideally wouldn't be Bash/POSIX centric, but I don't really mind)

packages/ai-chat/src/browser/chat-session-store-impl.ts
```diff
@@ -54,9 +58,33 @@ export class ChatSessionStoreImpl implements ChatSessionStore {
...
     protected storePromise: Promise<void> = Promise.resolve();
 
+    @postConstruct()
+    protected init(): void {
+        this.preferenceService.onPreferenceChanged(event => {
+            if (event.preferenceName === SESSION_STORAGE_PREF) {
+                this.logger.debug('Session storage preference changed: invalidating cache.', { preference: event.preferenceName });
+                this.invalidateStorageCache();
+            }
+        });
```
colin-grant-work: How does this affect existing stored sessions? Could we have orphaned storage depending on when the user switches? Would that be bad, or better than having nothing?

packages/ai-chat-ui/src/browser/session-storage-preference-renderer.ts
cdamus: I didn't want it to show a focus outline because it's not a button. It's just a disclosure element. Should it then not be focusable?

packages/ai-chat/src/common/ai-chat-preferences.ts
cdamus: Empty wasn't really intended as disabling storage, it just didn't seem like it made sense to use it as a path because it would _be_ the (first) workspace root. And I didn't like using an empty value to mean the default path .theia/chatSessions. But then again the global path uses a default when empty because that default uses a magic $HOME value that otherwise wouldn't be interpreted in the input.

A None scope could work but we already have a 0 limit on the number of stored sessions denoting that.

I think probably the best approach is validation, disallowing the empty path.

I would like to do the same in the global path field, too, but then I'd have to make the ~/.theia/chatSessions value an explicit default, and that suggests that people can use a leading tilde for the home directory more generally, and maybe I should just do that in this field. It's easy enough to detect a leading tilde and not support other environment variables, not even $HOME.

So how's this for a plan?

- change the global path default to ~/.theia/chatSessions explicitly (bearing in mind that .theia is itself configurable)
- disallow empty input in both path fields
- expand a leading tilde path segment to the home directory in the global path
- add validation to the workspace path to require a relative path, which then necessarily bans a leading tilde segment, because it's required to be relative to the workspace. Indeed, it should actually be _within_ the workspace so .. segments that escape are banned, too (basically, we should validate that the resolved path is within the workspace)
- no more clumsy way to disable storage via these paths

packages/ai-chat-ui/src/browser/session-storage-preference-renderer.ts
cdamus: This is a common pattern in settings UI in Mac apps but I won't insist on it. Not sure how a message or warning would work, though, and I would avoid colour for accessibility reasons.

package-lock.json
cdamus: The lock file only changed because I had to add a dependency from the ai-chat package to the preferences package. I'm loath to manually edit a lock file.

packages/ai-chat/src/common/ai-chat-preferences.ts
cdamus: It lets one build upon it my selective replacement of props but I can freeze it instead and require a spread for that scenario.

packages/ai-chat/src/browser/chat-session-store-impl.ts
cdamus: I think for the most part people will be glad to leave their global storage behind because it really was only applicable if you only work on one project ever.

But We could detect that we're creating a new workspace storage location and, in that case, if there is global storage, as the user whether she would like those chats to be imported.

packages/ai-chat-ui/src/browser/session-storage-preference-renderer.ts
cdamus: For keyboard access I think best would be to make the whole setting focusable and let the arrow keys disclose/hide the subordinate settings as usual in a tree widget. Does that make sense?

packages/ai-chat-ui/src/browser/style/index.css
cdamus: The index is imported by the frontend module, which doesn't really use the styles. But as in this case they really are only intended to be used by the preference renderer, I can import them there. 👍

packages/ai-chat-ui/src/browser/session-storage-preference-renderer.ts
colin-grant-work: Hm... tricky.

> For keyboard access I think best would be to make the whole setting focusable and let the arrow keys disclose/hide the subordinate settings as usual in a tree widget. Does that make sense?

That does make sense, and I like the idea, but there are also some hiccups. I think the setting is focusable already (for better or worse), but it doesn't show focus the usual way, so relying on that without a better indicator may not be ideal.

packages/ai-chat/src/common/ai-chat-preferences.ts
colin-grant-work: > So how's this for a plan?

I like every point 👍

packages/ai-chat/src/browser/chat-session-store-impl.ts
cdamus: You know what, I don't think we should even ask. The global chat session storage is global. So it should be the seed for any new workspace-specific chat history. Then over time that workspace's history will evolve and push out the stuff that was global. This is what I would expect as a user.

cdamus:
Thanks @colin-grant-work for your thorough review and excellent comments. I've pushed changes that I think, overall, address all of your feedback and make this solution so much better.

  1. Address small review comments (ff335ea21d)                                                                                                                                                                                                       
    - Moves CSS import to where it's actually needed                                                                                                                                                                                                  
    - Freezes SessionStorageValue::DEFAULT and uses spread operator when modification is needed                                                                                                                                                       
    - Adds focus border styling on the disclosure triangle when focused                                                                                                                                                                               
  2. Seed a new workspace storage from global storage (f12145d8d9)                                                                                                                                                                                    
    - When initializing workspace storage for the first time, copies existing chat sessions from global storage to the new workspace storage location                                                                                                 
  3. Use application's configuration directory name (75f18699c4)                                                                                                                                                                                      
    - Removes hard-coded .theia/ directory name                                                                                                                                                                                                       
    - Dynamically uses the application's actual configuration directory name                                                                                                                                                                          
    - Removes references to .theia in user-facing text                                                                                                                                                                                                
  4. Rationalize default path settings (70a7e81d7b)                                                                                                                                                                                                   
    - Sets explicit default values for both workspace and global paths                                                                                                                                                                                
    - Prevents empty inputs                                                                                                                                                                                                                           
    - Adds validation: global path must be absolute, workspace path must be relative and cannot escape the workspace directory                                                                                                                        
  5. Do not disable any input fields (16f44d6d0b)                                                                                                                                                                                                     
    - Changed the preference UI to no longer disable input fields based on scope                                                                                                                                                                      
    - Instead shows an informational label explaining when a path is not being used                                                                                                                                                                   
  6. Fix handling of workspace root changes (2c6530070f)                                                                                                                                                                                              
    - Listens for general workspace changes to detect when the first workspace root changes                                                                                                                                                           
    - Invalidates the storage cache when this happens                                                                                                                                                                                                 
    - Does not set up the workspace listener at all when storage scope is global

cdamus:
Oh somebody changed the API of preference node renderers _today_. Clever 😉 must be coincidence. I've rescued my renderer in commit 26e892857744f059c60b53ba26fe6cb889099b43.

package-lock.json
cdamus: This comment was mooted by the merge. Somebody else already removed these "dev" entries on main branch.

cdamus:
And finally a loose end, sorry:

7. Use tryGetRoots() instead of the roots promise (b641fbab4090752f3f89e26243aaa19b164db582)
  - we will always retry on the event that announces the roots anyways when they eventually do show up.

packages/ai-chat/src/common/chat-session-store.ts
```diff
@@ -46,9 +46,16 @@ export interface ChatSessionStore {
...
+     * Get index of all stored sessions.
+     * Note: This may trigger storage initialization if not already initialized.
      */
     getSessionIndex(): Promise<ChatSessionIndex>;
+    /**
+     * Check if there are persisted sessions available
+     * This has the benefit of not initializing the storage on disk if it does not
+     * already exist.
+     */
+    hasPersistedSessions(): Promise<boolean>;
```
cdamus, line 58: A note to reviewers: This API is added to provide clients (in particular the Show Chats command, which availability is conditional on this) a means to query whether there are any persisted sessions _without triggering an initialization of the store_.

This is important because now that we have workspace storage and a preference to set the path of that storage (and the global storage, for that matter), the initialization of that storage on disk must be deferred until it actually needs to exist. Especially because workspace storage is seeded from the global store.

Without this new API, the Show Chats command's recalculation of its enablement would fire while the user is typing out a path in the settings UI and create a bunch of unwanted stores on partial paths. This could possibly be addressed by hysteresis in the UI but that would be awkward and inevitably leave gaps, and creating junk files on disk is an egregious bug.

Adding this API does require incidental changes in other files in this package, especially in tests, that otherwise would not have been changed for this feature.

packages/ai-chat/src/browser/chat-session-store-impl.ts
```diff
@@ -236,22 +242,113 @@ export class ChatSessionStoreImpl implements ChatSessionStore {
...
+    /**
+     * Initializes the storage directory on disk, creating it if necessary
+     * and seeding it from global storage for new workspace storage locations.
+     */
+    protected async initializeStorage(root: URI): Promise<void> {
+        const storageConfig = await this.getStorageConfig();
+        const workspaceRoot = await this.getWorkspaceRoot();
+        // Only consider it workspace storage if both the preference is workspace AND we have a workspace open.
+        // When no workspace is open, we fall back to global storage even if scope preference is 'workspace'.
+        const isActuallyWorkspaceStorage = storageConfig.scope === 'workspace' && workspaceRoot !== undefined;
```
colin-grant-work, line 319: This is a pretty niche case, since not having a workspace is pretty unusual - though maybe if Theia is our usual conduit to AI, it would be more common? - but the described behavior gave me pause. It's probably right, but conceivably a user means by 'workspace' 'don't touch my home directory!'

cdamus: I know of Theia-based applications that do not use the workspace feature, so that's what this was meant to handle. But perhaps in that case it would actually be better to:

- make global scope default
- suppress the scope preference and the workspace path preference in the UI

I think this should be straightforward to do because we already dynamically define the preference schema and have a custom renderer. What do you think? If it's appealing, should I append to this PR or create a follow-up?

cdamus: And then, of course, the history limit preference can be set to 0 to disable the persistence feature.

colin-grant-work: > make global scope default
> suppress the scope preference and the workspace path preference in the UI

This seems like a good idea. At the moment, going from no workspace -> workspace involves a reload of the frontend, so I don't think it will be that disruptive to swap out defaults at that point.

cdamus: Done in commit 498dd2f276d9b20ef341993f18103e09a1b6d73e.

CleanShot 2026-01-23 at 15 38 32@2x

PR #16850, opened 2026-01-20
fix(core): change 'No reply handler' from throw to console.warn

### LINKED ISSUE #16842: rpc-protocol: Change "No reply handler" from throw to console.warn
## Problem

In `packages/core/src/common/message-rpc/rpc-protocol.ts`, when a reply arrives for a request that no longer exists in `pendingRequests`, the code throws an error:

```typescript
handleReply(id: number, value: any): void {
    const replyHandler = this.pendingRequests.get(id);
    if (replyHandler) {
        this.pendingRequests.delete(id);
        replyHandler.resolve(value);
    } else {
        throw new Error(`No reply handler for reply with id: ${id}`);
    }
    // ...
}
```

This causes crashes when:
- A request times out or is cancelled, but the reply arrives later
- Network latency causes out-of-order message delivery
- Plugin host restarts while requests are pending

## Proposal

Change `throw new Error` to `console.warn` for both `handleReply` and `handleReplyErr`:

```typescript
} else {
    console.warn(`No reply handler for reply with id: ${id}`);
}
```

This is a non-critical condition - a late reply for a cancelled request should be silently ignored, not crash the application.

## Related

- #13136 proposes better tracing for debugging these cases
- This change would make that tracing more useful by not crashing before tracing can help

## Environment

- Theia version: 1.67.0
- Observed when using VS Code extensions in plugin-host

DrumRobot:
#### What it does

Changes `throw new Error` to `console.warn` in `handleReply` and `handleReplyErr` methods of `RpcProtocol`.

Late replies for cancelled or timed-out requests are non-critical conditions that should not crash the application. This can occur when:
- A request times out but the reply arrives later
- Network latency causes out-of-order message delivery
- Plugin host restarts while requests are pending

Fixes #16842

#### How to test

1. Start Theia with plugins that use RPC communication
2. Trigger scenarios where reply handlers may be missing (e.g., rapid plugin activation/deactivation)
3. Verify that warnings appear in console instead of crashes

#### Follow-ups

None identified.

#### Breaking changes

#### Attribution

Contributed on behalf of [es6kr](https://github.com/es6kr)

#### Review checklist

#### Reminder for reviewers

### DIFF
--- a/packages/core/src/common/message-rpc/rpc-protocol.ts
+++ b/packages/core/src/common/message-rpc/rpc-protocol.ts
@@ -132,7 +132,9 @@ export class RpcProtocol {
             this.pendingRequests.delete(id);
             replyHandler.resolve(value);
         } else {
-            throw new Error(`No reply handler for reply with id: ${id}`);
+            // Late replies for cancelled/timed-out requests are non-critical - just warn
+            console.warn(`No reply handler for reply with id: ${id}`);
+            return;
         }
         this.disposeCancellationEventListener(id);
     }
@@ -143,7 +145,9 @@ export class RpcProtocol {
             this.pendingRequests.delete(id);
             replyHandler.reject(error);
         } else {
-            throw new Error(`No reply handler for error reply with id: ${id}`);
+            // Late error replies for cancelled/timed-out requests are non-critical - just warn
+            console.warn(`No reply handler for error reply with id: ${id}`);
+            return;
         }
         this.disposeCancellationEventListener(id);
     }


ndoschek:
Hi @DrumRobot, thanks for your contribution, someone will take a look shortly.
Just one request: Please use our [PR template](https://github.com/eclipse-theia/theia/blob/master/.github/PULL_REQUEST_TEMPLATE.md) as we use this as part of our changelog generation. Thanks!

